PR #6345, opened 2026-08-19
feat(#6329): detect machine-generated source by content marker, stamp it, and rank it below authored code

cajasmota:
PR1 of #6329: **detection + flag + one consumer + visibility.** Additive and reversible — it changes no entity counts and drops no content. Body-dropping (`ProjectDeclarations`) is PR2, deliberately.

Closes #6329.

## Read this first: three claims in this issue's history were wrong, including two of mine

1. **The filing premise.** I argued skipped generated declarations inflated the `bug-extractor` rate. Measured: the skip patterns were never loaded, and an A/B produced a **0-edge delta**. The dead config was deleted in #6330.
2. **The design report's ~105 marker hits** on a corpus. The strict regex finds **4**; the other 101 were JPA `@GeneratedValue` matched by a case-insensitive unbounded probe — a defect in the measurement.
3. **"Zero false positives"**, in the first version of this PR. It rested on 17 hand-checked hits across two repos, which is real verification of each hit but far too small a sample to support "zero". An independent sweep found three systematic false-positive classes. See below.

Body-stripping was also assumed to be the ranking win and is not: `buildDocTerms` (`internal/mcp/scoring.go:268-375`) indexes `Name`, file stem (weight 1.5), path dirs, `docstring` and discriminators — **`Entity.Content` is not indexed at all.** The ranking win is the reduced entity count out of a generated file plus the explicit demotion here.

Surviving justification: **@dcastro-imp's codebase is 30.5% of files but 47.6% of lines machine-generated**, which is @manuel1358000's #6314 with a number attached.

## Detection: marker first, filenames second

@dcastro-imp found a third generated-code category in their own codebase after two passes of filename counting — WSDL/XSD partials matching neither `*.Designer.vb` nor `Reference.vb` but carrying `' <auto-generated>`. A content marker catches generators nobody enumerated; filename globs produce false *positives*, #6330 being the standing example.

**The marker scan is restricted to the file's first contiguous comment block.** Language preambles (`#!`, `<?php`, `<?xml`, BOM) and a lone `package`/`namespace` statement keep the block open; `/* … */` and `<!-- … -->` interiors are part of it.

That narrowing exists because the first version scanned the first 20 lines and matched hand-written prose. Independent sweep over the Go module cache, **same method before and after** — 11,740 candidates built with `grep -rlIE`, then run through the detector:

| | flagged |
|---|---|
| pre-fix | 8,544 |
| post-fix | **8,506** |

[...]

Issue #6338, opened 2026-08-18
doctor and status say nothing about files skipped for having no extractor, so an unsupported language looks identical to a healthy index


cajasmota:
S1 of #6327, split out because it is **independent of VB.NET** and should ship first. It fixes the blind spot for every unsupported language at once, not just the one that prompted it.

This is @dcastro-imp's second ask in #6321, and their description of finding the problem is the specification:

> `grafel doctor` reports the group as healthy and shows no quarantined directories or errors, so the gap isn't obvious unless you go check file-by-file — it just silently indexes nothing for the whole language.
>
> it'd help a lot if `grafel doctor`/`grafel status` could surface "N files skipped, language not supported" per extension, so this kind of silent gap is visible without having to manually cross-check `find <repo> -iname "*.vb" | wc -l` against the indexed entity count like I did here.

## The problem

A file in an unsupported language is dropped at classification. It is not misparsed and it does not error — no extractor is ever reached, nothing is quarantined, and `doctor` has nothing to report. Their 670 `.vb` files produced **zero entities** while the group reported healthy.

The failure mode is the one this project keeps finding: the tool is confidently silent about something it did not do.

## What to build

`grafel doctor` and `grafel status` should report, per group or repo, the files that were seen and skipped **because no extractor claimed the extension**, aggregated by extension:

```
Unsupported languages (no extractor):
  .vb      672 files   (VB.NET — not supported, see #6327)
  .pas      14 files
  .m         3 files
```

Requirements:

- **Aggregate by extension, not by file.** A per-file list is unusable at 672 files and would bury the signal.
- **Distinguish "no extractor for this extension" from other skip reasons.** Vendored, ignored, quarantined and binary files are already handled elsewhere and are not this. Conflating them recreates the noise problem.
- **Report zero as absent, not as a zero row.** A clean repo should print nothing here.
- **Threshold sensibly.** One stray `.pas` file is noise; 672 `.vb` files is the headline. Consider surfacing only extensions above a small count in `status`, with the full table in `doctor`.
- If the extension corresponds to a language we know about but have not implemented, **say so by name** — "VB.NET — not supported" is far more actionable than ".vb".

## Where it goes

Classification already computes the decision — `internal/classifier/classifier.go` returns a `ClassifyResult`, and the walk/extract path drops the file at `internal/daemon/extract/coordinator.go:665` (`bucketByLanguage`). The counting belongs where that decision is made; the reporting hangs off the existing `doctor`/`status` surfaces.

[...]

Issue #6212, opened 2026-08-06
Stamp the manifest from walk-time hashes, not a re-hash at commit time — the contract is currently true only when nothing writes during the window


cajasmota:
Split out of #6207's review, where the skew was measured rather than theorised.

## The defect

`commitManifest` hashes every file **at commit time**. The document was built from the bytes `Run`'s walk read. Anything written in between is stamped into the manifest but is not in the graph.

A probe injecting a working-tree write inside the `writeGraphGen` seam confirmed it: the manifest stamps **post-walk content**. Next pass, `FilterWithGit` hash-matches the file and classifies it UNCHANGED (`diff.go:411-413`), so the edit is invisible until the file is touched again or HEAD advances.

The realistic trigger is the watcher-driven fallback reindex — which fires *because* files are changing. A developer saving a file mid-reindex is the ordinary case, not an exotic one.

#6207 made this wider on both paths: the window now spans rename-detect, algo carry-forward, `sortDocumentForEmission` and all of `WriteGraphGen` — seconds on a real repo, where it used to be immediately post-extraction. It also made it newly reachable on the fallback, which previously wrote no manifest at all, so the stale stamp caught the edit. That was the correct trade at the time (one edit lost until its next touch is self-recovering; the stuck state #6207 fixed was not), but it is a trade, and this is the bill.

## The fix, and why this shape over the alternative

**Stamp from the hashes computed during the walk/extraction**, not from a re-hash at commit.

- It makes the manifest's contract — *these bytes are what the graph was built from* — true **by construction**, on every path, instead of true whenever no write happened to land in a multi-second window. That is the difference between an invariant and a probability.
- Extraction already reads every file, so the hash is close to free there.
- **It deletes the O(repo) SHA-256 sweep from `commitManifest` entirely**, which is a direct saving on the fallback path and pulls the same direction as #6206.

The alternative — skip files whose mtime moved after the walk — was considered and is worse. It is a heuristic on a clock, so it misses writes inside filesystem mtime granularity; when it *does* fire it converts a wrong-content entry into a missing one, which is safe in the over-reporting direction but means permanent churn until that file is next extracted; and it still needs a stat sweep, so it buys none of the performance.

The cost of the preferred shape is plumbing: threading the stamp out of extraction and into `commitManifest`. That is the honest trade — more wiring, in exchange for an invariant that can actually be asserted.

## Acceptance

[...]

Issue #6201, opened 2026-08-06
Incremental gives up 25x too early, and giving up costs 7% more than never trying: the reject path pays a full SHA-256 sweep first


cajasmota:
Measured under #6199. Two independent defects that compound: the incremental path gives up ~25× earlier than it should, and giving up costs more than never having tried.

## The tax

The `too-many-changed` reject at `internal/extractors/incremental.go:90-113` fires *after* the tree walk, the git diff, **a full SHA-256 sweep over every file in the repo** (`diff.UpdateManifest`) and a manifest save. Then it returns, and the caller runs a full reindex — which walks and hashes everything again.

On a 3003-file / 58.5k-entity fixture, `GOMAXPROCS=4`, N=14, medians:

```
200 changed files, as shipped:   696 ms (rejected) + 9,540 ms (full reindex) = 10,236 ms
200 changed files, full reindex alone:                                          9,540 ms
```

**A rejected pass costs 7% more than not having attempted one.** The 696 ms buys nothing: the manifest it writes is immediately superseded by the full reindex's own.

## The ceiling is set ~25× below the real crossover

The measured cost model is a fixed tail plus a linear term: **1451 ms + 6.26 ms per changed file** (slope from the 10→500 columns). Against a 9,540 ms full reindex, incremental stops winning at **≈1280 changed files**.

The shipped ceiling is **50** on a default branch and **20** on a feature branch (`incremental.go:90,95,105-113`).

```
 50 changed files, incremental:  1,920 ms   =  20% of a full reindex
200 changed files, incremental:  2,947 ms   =  31%
500 changed files, incremental:  5,276 ms   =  55%
```

Every one of those wins comfortably, and all but the first are currently rejected. And `FilterWithGit:402-415` inflates the count via basename-stem matching — a measured 1-file edit can present as ~40 — so the effective ceiling is lower still than 50.

## What to change

**1. Reject before paying.** Move the count check ahead of `diff.UpdateManifest` and the manifest save. The reject path should cost the walk and the diff and nothing else. This is worth ~220 ms of the 696 ms on its own and is pure subtraction — there is no case where the discarded manifest was useful, because the full reindex rewrites it.

Better still if it is cheap: hoist the check ahead of the **tree walk** too, if the changed-file count can be obtained from the git diff alone. That would take the reject path to near-zero. Check whether the walk is a prerequisite of the diff or merely sequenced before it.

[...]

PR #6349, opened 2026-08-19
measure(#6199): land the incremental phase-trace harness, with the coverage claim proved from the AST

### LINKED ISSUE #6199: Incremental indexing is not delta-proportional: every pass pays a full walk, a full-repo hash sweep, whole-graph passes and a full re-marshal — and none of it is measured
Incremental indexing re-parses only the changed files — but everything around that parse is O(repo) or O(graph), and runs on **every** pass, including the fallback pass and the zero-change no-op.

## The fixed tail

From `internal/extractors/incremental.go`, in `TryIncremental`'s own order. Only two phases are O(delta): the AST-hash gate (`:427-440`) and per-file extraction (`:613-901`). Everything below runs regardless of delta size.

| Phase | Site | Cost |
|---|---|---|
| Full gitignore-aware tree walk | `:279` → `walkSourceFiles` `:1356` | O(repo), before any delta is known |
| **Full-repo SHA-256 sweep** | `:1150`, `:376`, `:457`, `:420` → `diff.UpdateManifest` (`internal/indexer/diff/diff.go:209-230`) | **O(repo)** — hashes every walked file, on success, on both fallbacks, **and on the zero-change no-op** |
| Full graph materialisation | `:466` `graph.LoadGraphFromDir` | O(graph) heap |
| 4× linear prune scans | `:482-525`, `:917-925` | O(graph) |
| Scoped resolve | `:951` → `ResolveScoped:316-327` | O(graph) — indexes all surviving entities |
| Whole-graph flow recompute | `:1028` `RunFlowsIncremental` | O(graph); its gate (`internal/engine/flow_incremental.go:108-111`) returns true for **any** new non-flow entity, i.e. every real edit |
| Lib boundary / structural coupling | `:1075`, `:1086` | O(graph) |
| **Coverage reachability BFS** | `:1095` → `internal/coverage/enrich.go:80,84` | O(graph), unconditional, **plus a full copy of every entity and relationship** into a new slice |
| Canonical sort | `:1113`, and **again** at `internal/graph/fbwriter/writer.go:97` | O(N log N) whole graph, twice |
| **Full graph re-marshal + rewrite** | `:1119` `WriteGraphGen` | O(graph) — the entire graph re-serialised as a new generation |

So a 1-file delta and a 50-file delta cost nearly the same, and both cost a large fraction of a full reindex. Wall time is dominated by graph size, not delta size.

**The prior graph is still fully materialised on this path.** `:466` → `graph/load.go:129` → `loadFBDocument:349` → `loadFBDoc(headerOnly=false)`. A header-only path exists (`load.go:365`) but is gated behind `GRAFEL_SERVE_FROM_MMAP` and documented as never called by the full-Doc path; `MultiReader` backs segment-set loading through the *same* materialise core (`load.go:137-138`). In-tree corroboration: `cmd/grafel/rebuild_history.go:12` calls it *"the full-materialise loader: ~608 MB of heap."*

## None of it is measured

- The only perf test is `internal/extractors/incremental_perf_test.go`, behind `//go:build perf` (excluded from the default suite), on a **10-file fixture**, with a 3s budget. It disclaims itself at `:78-88`: *"on a 10-file fixture there is no measurable separation between 'incremental' and 'degenerated into a full reindex'… needs a fixture big enough for the two paths to separate — a few thousand files, not ten."*
- No `Benchmark*` function touches incremental.
- **`Result.Duration` is computed and thrown away.** `incremental.go:1164` sets it; `cmd/grafel/daemon.go:1251-1255` copies only `Done`, `FallbackReason`, `ChangedFiles`. The only persisted timing is `GraphStatsSidecar.ExtractMS`, which despite its name is `time.Since(t0)` for the *whole* pass.
- The `~25× faster` claim at `internal/extractor/extractor.go:136` and `~5s → ~200ms` at `incremental.go:6-9` have no in-tree backing.

There is no per-phase telemetry in production, so the central claim of the incremental path is currently unobservable.

## A one-file edit can trip a full reindex

The fallback ceiling is **20 changed files on a feature branch, 50 on the default branch** (`incremental.go:90,95`, resolved at `:105-113`).

`FilterWithGit:402-415` marks any file sharing a basename stem with a dirty file as also dirty — so a single edit routinely inflates to ~40 changed files (see `cmd/grafel/diff_reindex_multiset_test.go:124`). On a feature branch that exceeds the 20-file ceiling and takes the full path.

A `git pull` of a few commits, any rebase, or any branch switch exceeds it every time. The comment at `:307-309` states this as intent: *"a large advance correctly trips the too-many-changed full-reindex fallback."*

## Also worth noting

`incremental.go:20-21` claims *"The #2167 conservative default of 5 is still the hard floor when the env override is absent."* No floor of 5 exists in `effectiveLimit` (`:105-113`) or `EffectiveIncrementalMaxFiles` (`internal/extractor/extractor.go:156-166`); `GRAFEL_INCREMENTAL_MAX_FILES=1` is honoured verbatim.

## Why this matters now

It gates the watcher design in #6180. An activity-gated watcher fleet leaves cold repos unwatched and refreshes them at query time — which is only viable if a query-time refresh is delta-proportional. Today it is not, and the cold repos the gate leaves unwatched are precisely the ones whose deltas exceed the fallback ceiling.

1. **Measure first.** Per-phase timing on a fixture of a few thousand files at 1 / 10 / 50 / 200 changed files, so the fixed tail is a number rather than a list. `Result.Duration` should also stop being discarded.
2. Then decide which whole-graph phases can be delta-scoped, deferred, or dropped — the coverage BFS and the flow recompute are unconditional today and are the most obviously delta-gateable.
3. The full-repo SHA-256 sweep on a zero-change pass is pure waste and is probably the cheapest win.
4. Whether the prior-graph materialisation can use the header-only/mmap path that already exists.

cajasmota:
Lands the per-phase instrumentation for `TryIncremental` that was written for #6199 and left on a branch when that issue closed. 723 lines of harness, plus the two things it needed before it could land.

Closes #6199.

## Why it is worth landing rather than deleting

It is the only tooling we have for per-phase incremental timings, and we have re-derived those by hand more than once. `TryIncremental` becomes a thin wrapper whose only job is owning the trace across every return path — the original commit's reasoning being that there are 10+ fallback returns, and *"a trace that misses the fallbacks would miss precisely the passes that pay the fixed tail and then throw it away."*

## The coverage claim is now a proof, not an inspection

The whole justification is "own the trace across EVERY return path", and nothing asserted it. That is the same defect shape as #6345's seam, which claimed a single chokepoint and turned out to cover one pass of three — a coverage claim verified by reading rather than by an invariant.

`TestPhaseTrace_EveryReturnIsTraced` does not list the returns; a list rots exactly the way the call-site list did. It discharges the claim from the AST in two parts:

- **(a)** `tryIncremental` has exactly **one** call site in the module — established by walking every `.go` file and enumerating `CallExpr`s, not by enumeration in the test.
- **(b)** In `TryIncremental`, `tr.emit` is an unconditional top-level `ExprStmt`, and no statement at any nesting depth before it contains a `ReturnStmt`.

(a) ∧ (b) ⟹ every return in `tryIncremental` lands on `emit`, because there is no other way in or out. That covers the 15 returns present today and every one added later. The test logs the count and fails if it drops below 5, so it cannot quietly stop guarding anything.

`TestPhaseTrace_EmitsOnFallbackAndSuccess` is the vacuity guard rather than the coverage claim: it drives a real too-many-changed fallback and a real successful pass, and reads the JSONL back.

Mutants, both killed:

- **M1** — wrap the emit in `if res.Done`, dropping the trace on every fallback. Both tests die; the JSONL is never written.
- **M2** — add a second, untraced entry point calling the body directly. `must have exactly ONE call site … found 2`. This is the half that kills the #6345 rot mode specifically, and it was re-verified independently.

## Default log output is unchanged, and that was not a preference

The summary line was unconditional while only the JSONL sidecar was gated. The evidence is decisive: the zero-change no-op — the most frequent incremental pass on a live daemon — logs **nothing** today, since the only `Printf` in the `totalChanged == 0` branch belongs to the absent-graph guard. So the line was not one more entry on an already chatty path; it was a new line in a silent one, once per poll per repo.

`GRAFEL_PHASE_TRACE` now gates all three outputs. `1|true|yes|on` gives the greppable summary and no file, so asking for the cheap line does not require inventing a throwaway sink; any other value is still the JSONL path. `TestPhaseTrace_SilentByDefault` pins it, and mutant M3 (ungate) fails showing the leaked line.

## Rebase was non-trivial

7 conflicts, all in `incremental.go`, because #6201/#6212/#6338 rewrote the exact statements the trace wrapped:

- both too-many-changed rejects moved from `diff.UpdateManifest`+`SaveManifest` to `UpdateManifestScoped`+`SaveManifestAtCommit`, so the branch's `manifest-sha256-sweep` span was measuring a call that **no longer exists there** — re-pointed and renamed `manifest-scoped-restamp`;
- step 9 moved to `diff.ApplyStampsAndFailures` (#6212, stamp from bytes read rather than re-hash) → `manifest-apply-stamps`;
- `writeGraphGen`: kept main's local symbol over the branch's `fbwriter.WriteGraphGen`;
- the sidecar write now carries forward `UnsupportedExtensions` (#6338) — kept, span wrapped around the write only.

No phase ordering changed.

## Measured rather than asserted

- **Span cost**: 20 spans + `newPhaseTrace` = **1561 ns/op**, against a no-op pass of ~97 ms — about 0.002%. Leaving the spans always-compiled is the right call. (Throwaway benchmark, not committed.)
- **testdata inertness**: `go list ./...` yields no package under `internal/extractors/testdata/`, and `go vet ./...` is clean. Confirmed rather than assumed. Caveat: `incrdrive/main.go` calls `extractors.TryIncremental` but is never compiled, so it will rot silently if that signature changes.
- **`measure.sh` exported `GRAFEL_HOME` and `GRAFEL_DAEMON_ROOT` but not `HOME`** — a re-run could have written into the real `~/.grafel`. Fixed, and the sandbox root is now overridable. That is the exact trap that took a live daemon down earlier in this cycle: neither variable alone is isolation.

## Known gap

`sortedPhaseNames` is dead (declared a test/diagnostic helper, no callers). Flagged rather than deleted, since it is the original author's.

Refs #6201, #6212, #6338, #6345


cajasmota:
## Independent adversarial review: REQUEST-CHANGES

Recording it here because the headline claim was broken by counterexample, and I had verified that claim myself an hour earlier — against the one shape it was already proof against.

What held: the gating decision (verified by execution), the per-span rebase for every span but one, the perf claim (re-measured at 1687 ns/op against the stated 1561 — same order), and the AST check's soundness against methods, `go`/`defer` call sites, function values, `ReturnStmt` inside a `FuncLit`, and build-tag-excluded files.

### 1. The AST "proof" is not a proof

`findCallSites` (`incremental_phasetrace_6199_test.go:200-204`) iterates `f.Decls` and only descends into `*ast.FuncDecl`. A call site in a package-level `var` initializer is a `*ast.GenDecl` and is never inspected:

```go
var untracedSecondCallSite = func(ctx context.Context, repoPath, stateDir string, logger *log.Logger, cfg *extractor.ExtractorConfig) Result {
	return tryIncremental(ctx, repoPath, stateDir, logger, cfg, time.Now(), nil)
}
```

Compiles, is a real reachable call site, and the test passes with **EXIT=0**. Every return through it emits nothing, and `tr == nil` is swallowed by the nil-guards in `span`/`add`/`emit`. So claim (a) — "exactly one call site in the entire module, enumerated from the AST of every `.go` file, not from a list" — is false as written.

Worth being precise about how this got through: my own verification mutant used a `func` declaration, which is exactly the shape the walk already covers. Verifying a coverage claim with an instance of the covered case is not verification. Fix is `ast.Inspect` over the whole file.

### 2. `panic` bypasses `emit` and permanently leaks the heap-sampler goroutine

`emit` is a plain call at `incremental.go:404`, not deferred, and `stopHeapSampler` is reachable only from inside it (`incremental_phasetrace.go:225`). A panic unwinds past both.

Not hypothetical here: `internal/daemon/sched/scheduler.go:1777-1785` explicitly `recover()`s the incremental call, with a comment that an index can panic for reasons the fbwriter fail-soft does not catch. The daemon survives — and the sampler survives with it. Measured, after recover + 200 ms + two forced GCs:

```
goroutines before=2 after=3 (delta=1)
```

Each recovered panic while `GRAFEL_PHASE_TRACE` is set permanently leaks a goroutine holding a 5 ms ticker calling `runtime.ReadMemStats` — a **stop-the-world** operation. N leaks means a STW every 5ms/N, inside the harness whose entire purpose is producing trustworthy timings.

**And the invariant test forbids the fix.** `isEmitCall` (`:137-148`) requires a bare top-level `*ast.ExprStmt`; the panic-safe shape `defer func(){ tr.emit(res) }()` is a `DeferStmt` containing a `FuncLit` and fails with "has no UNCONDITIONAL top-level tr.emit". The chosen shape and panic-safety are mutually exclusive as written, so the invariant has to widen rather than the fix route around it.

### 3. `extract-changed-files` has an unreachable `end` on two paths

The span opens at `incremental.go:857` and `endExtract()` is at `:1224`, but two function-level `return fallback(t0, …)` sit inside the region — `:1026` (`extract-error`) and `:1058` (`no-tree-no-records`). On exactly the two fallbacks that pay for full extraction and then throw it away, the trace reports **no `extract-changed-files` phase at all**, and `accounted` silently underreports `total` by the dominant phase.

That contradicts `incremental_phasetrace.go:82-84` — "reported even when the pass falls back part-way, which is the point" — and it is the confident-wrong-number class. A structural audit is being added rather than the one fix.

### 4. Two surprising `GRAFEL_PHASE_TRACE` values, verified by execution

`0`, `false`, `off`, `no`, `OFF`, `" "` all yield `enabled=true` and are taken as **file paths** (`traceePath="false"`, `traceePath="off"`). An operator disabling tracing with `=0` turns it on and appends JSONL to a file literally named `0` in the daemon's cwd.

Separately, `=1` **does** start the heap sampler, though the field comment says "only active when tracing to a file" — and its output is written only into the JSONL, which boolean mode never writes. So it pays a stop-the-world `ReadMemStats` every 5 ms to compute a number it discards.

### 5. A false claim in a doc comment

`incremental_phasetrace.go:15-17` states "There is no allocation per pass beyond a slice of ~20 samples". Measured with tracing disabled: **22 allocs/op, 2064 B/op** — the trace struct, the 576-byte samples array, and one heap-allocated closure per span, since `span` returns a capturing `func()`. Immaterial to performance; a false statement this repo will cite later.

### 6. `measure.sh` — the fix stopped at the script body

Line 12 correctly exports `HOME` + `GRAFEL_HOME` + `GRAFEL_DAEMON_ROOT`. The copy-paste recipe at `:87-88` — what an operator actually runs — still exports only `GRAFEL_HOME` and `GRAFEL_DAEMON_ROOT`. The exact bug that was fixed in the body, left intact in the instructions directly above it.

Secondary: `WT` defaults to a path that is never created, and with `set -u` but no `set -e`/`pipefail`, a missing binary makes `time` fail, `awk` yield empty, and the loop write blank rows and print `done` — a measurement harness that reports success on a completely empty run.

Fixes in progress on the branch, with a killing mutant required for each.

cajasmota:
## Independent re-review: APPROVE on shipped behaviour, four gaps in the audit tests

All nine fixes hold under execution. The findings below are in the tests that verify the code, not in the code — but they are being fixed rather than filed, because a coverage test that passes under a mutant is the exact failure this PR exists to close.

### What was verified by execution

**The deferred emit is equivalent on every non-panic path.** Three real passes (incremental success, `too-many-changed` fallback, empty no-op) with a byte-level mutant restoring the old plain trailing call. Output character-for-character identical, including phase ordering:

```
REC[0] done=true  reason=""                              changed=1 walked=1 ents=3 rels=2 nphases=22
REC[1] done=false reason="too-many-changed files=2 limit=1" changed=0 walked=2 ents=0 rels=0 nphases=6
REC[2] done=true  reason=""                              changed=0 walked=0 ents=0 rels=0 nphases=7
```

The closure-captures-variables risk is not live: `res` and `returned` are each assigned exactly once, nothing writes either afterwards, and `res` is not a named return so the defer cannot alter what the caller sees. `TryIncremental` has one `defer`; `tryIncremental` has **zero** `defer` and zero `recover()`, so there is no LIFO interaction and `returned` cannot be observed mid-state. The panic still propagates — asserted at the caller, and the test is non-vacuous (reverting to the plain call fails it with "no record was written at all").

**Disabled-path numbers re-derive.** Claimed 1782 → 132 ns/op, 25 → 1 allocs. Measured: **1812/1896/2105 ns · 2256 B · 25 allocs** before, **125.5/126.4/125.9 ns · 144 B · 1 alloc** after. Byte and alloc counts exact. Enabled mode drops no span. The shared `noopSpanEnd` is race-clean under 8 concurrent passes × 3 modes with `-race`.

**The audit is not vacuous**: 30 spans, hard floor at 25, and the positive control (deleting `endExtract()` before the `extract-error` return) fails naming `extract-changed-files` at `incremental.go:1055:4`.

### The four gaps — all mutants compiled and left the tests PASSING

**1. `isEmitCall` accepts a conditionally-skipped emit.** The deferred branch rejects only a *top-level* `ReturnStmt` before the emit, so `if !returned { return }` nested inside is invisible. The bare branch uses `countReturns`, which does descend — so the doc comment's claim that both forms carry the same guarantee is not what the code checks.

**2. A discarded span call escapes both the compiler and the audit.** `tr.span("x")` as a bare expression statement is not counted at all. **This refutes something I stated publicly:** I said the never-closed-at-all case was already caught by the compiler. `declared and not used` fires only for `end := tr.span(...)`; `_ = tr.span(...)` is caught by the audit; the bare-call form is caught by neither. My claim was true of the two forms I had in mind and false as stated.

**3. Shadowing launders the escape.** An inner `endX := tr.span(...)` in a nested scope satisfies the outer span, because the scan matches on identifier name and stops at the first call it meets.

**4. `break`/`continue`/`goto` are not modelled.** Only `ReturnStmt` is collected. Latent rather than live — the one loop-adjacent span is opened outside its loop — but it is a real hole.

### And a scope gap worth more than any of the four

The span audit parses **only `incremental.go`**, while the call-site check scans the whole module. Move a phase into a helper file and its spans go unaudited, silently, with the test still green. Being fixed by widening or by asserting no `tr.span(` exists elsewhere, so that day fails loudly instead of quietly reducing coverage.

### Out of scope, with reason

`testdata/incrfixture/main.go` has the same uncompiled-rot exposure that got `incrdrive/main.go` deleted, but a materially smaller blast radius: `incrdrive` re-implemented production call sequences, so any refactor could rot it silently, whereas `incrfixture` imports stdlib only and is invoked by explicit path from `measure.sh`, which fails loudly the moment it breaks. Nothing in grafel can rot it. Left in place.

Issue #6356, opened 2026-08-19
nine languages route with no registered extractor: files are indexed, nothing is emitted, and the #6338 report cannot see them


cajasmota:
Found while reviewing #6355 (VB.NET S2). This is #6321's failure mode, live today, for seven languages, and unlike VB.NET nothing reports it.

## The finding

Enumerating every value in `extensionLanguageMap` against `extractors.Get`, **8 of 67 languages the classifier recognises have no registered extractor**:

| language | extension | `Skip` | `SkipReason` | Tier | `SupportedExtension` | appears in the #6338 report? |
|---|---|---|---|---|---|---|
| nginx | `.nginx` | false | — | 1 | **true** | **no** |
| objective_c | `.m` | false | — | 1 | **true** | **no** |
| perl | `.pm` | false | — | 1 | **true** | **no** |
| prisma | `.prisma` | false | — | 1 | **true** | **no** |
| protobuf | `.proto` | false | — | 1 | **true** | **no** |
| r | `.R` | false | — | 1 | **true** | **no** |
| toml | `.toml` | false | — | 1 | **true** | **no** |
| vbnet | `.vb` | true | `unsupported_language` | 0 | false | yes (#6355) |

The seven classify as `Skip=false, Tier=1`. They are bucketed by `bucketByLanguage` (`internal/daemon/extract/coordinator.go:673-681`), a subprocess is spawned per language, and they land on `internal/daemon/extract/subproc.go:336-338`:

```go
errors.Is(exErr, ErrNoExtractorForLanguage) → stats.Skipped++
```

Counted as skipped, no entities emitted, nothing said. `internal/extractors/incremental.go:961-964` is the same shape with a log line.

**That is exactly what @dcastro-imp reported in #6321** — files indexed, nothing produced, no explanation — for seven more languages. The only difference is that VB.NET now has a row in the `doctor`/`status` report and these do not.

## Why it was invisible

`SupportedExtension` answers `true` for all seven, because it asks "does the router have a name for this file?" rather than "will an extractor produce entities?". The #6338 report filters on that function, so these languages are excluded from the one surface built to make this class of gap visible.

## Three hand-maintained lists of one derivable fact

The fact is computable — `extractors.Get(lang)`. Instead it is currently spread across:

1. `unsupportedLanguageNames` (`internal/classifier/unsupported.go`) — the #6338 report allow-list.
2. `languagesAwaitingExtractor` (same file, added by #6355) — one entry, `vbnet`.
[...]

PR #6357, opened 2026-08-19
fix(#6356): .proto files were never extracted — the extractor registered under a token the classifier never emits

cajasmota:
Every `.proto` file grafel has ever indexed produced **zero entities**. A complete, tested protobuf extractor has existed the whole time; nothing could reach it.

Split out of #6356. Refs #6356.

## Two names, not one

The extractor registered itself under one token (`proto.go:40`, `extractor.Register("proto", …)`) while the classifier routes the extension to another (`classifier.go:435`, `".proto": "protobuf"`). `extractors.Get("protobuf")` finds nothing, so every `.proto` file is discovered, classified `Skip=false, Tier=1`, bucketed, handed to a subprocess, and dropped at `subproc.go:336` with `stats.Skipped++`.

**Fixing that alone still produces zero entities.** `subproc.go:314` passes the classifier token verbatim to `treesitter.Parse`, and `migratedLanguages` only knew `"proto"` — so `Parse("protobuf")` returns `ErrUnsupportedLanguage`, `TSTree` stays nil, and `Extract` returns `nil, nil` on its first line. One silent failure swapped for another. Both halves are fixed, and mutant M1b exists to keep them fixed together: with the registry corrected and the alias removed, the derived registry gate **passes** while the dispatch test fails.

The extractor stamps its own entities `Language: "protobuf"` in six places, so it agreed with the classifier and disagreed only with its own registration key — a typo, not a design decision.

## What turning it on revealed, and what this PR does about it

Entities are correct, hand-checked against source: `lease.proto` yields exactly its 3 messages, all 5 fields with right types, 1 import, exact line numbers. But the first version of this PR would have **injected 25 dangling relationship targets** into every graph containing a `.proto` file, which `resolve/refs.go` materialises as phantom nodes.

| | entities | relationships | dangling targets |
|---|---:|---:|---:|
| before this PR | 0 | 0 | 0 |
| registration fix alone | 226 | 317 | **25** |
| with both defect fixes | **239** | **305** | **0** |

- **Enum values now exist.** `enum E` emitted `CONTAINS` edges to `…:E#E_UNSPECIFIED` while no enum-value entity was ever built — the package doc claimed "enum → enum value" and described behaviour that did not exist. `buildEnum` now emits one `enum_value` entity per value, mirroring `buildField`, with `Properties["number"]` and real line numbers. Emitting rather than dropping the edges, because an enum's values are its only content.
- **Unresolvable cross-file `REFERENCES` are suppressed.** `namedTypeRefs` stripped the package qualifier and built the ref against the *importing* file, so `etcdserverpb.LeaseTimeToLiveRequest` targeted a path that cannot exist. Correct resolution needs a proto-package index over the import closure, which `Extract` cannot build one file at a time — so the edge is now absent rather than dangling, guarded by a test that fails if the filter also kills the same-file case.
- **The double language stamp is aligned.** `internal/custom/cpp/protobuf.go` stamped `Language="proto"` on entities from the same files, and `RunCustomExtractors` runs independently of `extractors.Extract` — so a single `.proto` file would have yielded entities under both stamps, and anything filtering on `protobuf` would have missed half.

The phantom check is an **invariant over every emitted structural edge** (4 fixtures, 24 edges, with a `checked == 0` guard), not spot checks, and `internal/resolve/proto_enum_value_6357_test.go` proves the other half at the layer that decides phantom-vs-node.

## Correction: the parse-failure diagnosis in the first version was wrong

I attributed three failing files to "gogoproto custom options". Probed directly:

| input | error ratio |
|---|---|
| `option (gogoproto.marshaler_all) = true;` | **0.0000** |
| `syntax = "proto2";` (2-line file) | 0.2222 |
| proto2 + `optional` / `required` | 0.2105 / 0.2632 |
| `edition = "2023";` | 0.2500 |

**gogoproto options parse perfectly.** The bundled grammar accepts **proto3 only** — `syntax = "proto2"` and `edition = "2023"` are hard rejections, and `record.proto`/`snap.proto` contain no gogoproto at all. The ratio pins at ~0.25 regardless of file size, because the whole body collapses into one ERROR node, so proto2 files are dropped **whole, every time**.

So the limitation is not "some options aren't covered" — **grafel's protobuf support is proto3-only**, which is far wider and differently shaped. Recorded in the package doc, and the same wrong claim was found stated affirmatively in `docs/c3-feature-impact-analysis.md:70` and corrected there too.

A second arithmetic correction from the same re-measurement: **4 of the 14 local `.proto` files are classifier-skipped as `vendor_testdata`**, so 10 reach the extractor and **3**, not 4, are gate-rejected.

**The local `.proto` population is 14 files across two trees and does not size the real impact** — these are not gRPC codebases, and the population that cares about protobuf is the one not represented here.

## Filed rather than fixed here

- **#6358** — every `oneof` field is silently dropped; the members nest one level deeper than `buildMessage` looks. A perfectly-parsing file yields 1 entity where the source has 3.
- **#6359** — rpc request/response edges are bare-name, attributed to the *file* rather than the rpc, duplicated when the types match, and can be literal `"?"`.
- **#6360** — the 10% syntax-error gate is a whole-file average, so a proto3 file with one typo parses at 0.0033 and emits a message entity with a silently missing field. Not protobuf-specific; the gate is shared.

## Also

The gate's own failure text said *"register one, or drop the row from `classifierRepresentativeInputs`"* — inviting a future engineer to delete the regression gate for this bug, with nothing detecting the deletion. Reworded.

And `classifier_test.go:790` had **documented this exact bug with the exact fix** in `knownMismatches`. The test that read it only called `t.Log`. The knowledge sat in the repo for as long as that list existed; nothing acted on it because nothing failed.

Coverage matrix does not move — `extractorNonLanguageFormats` keys on the directory name `proto`, unchanged, so protobuf stays in the "multi" bucket and #6332's gate passes.


cajasmota:
## Independent adversarial review: REQUEST-CHANGES — and my diagnosis of the parse failures was wrong

The registration fix is correct, minimal and well-guarded. What it turns **on** is the problem, and one of the findings is a correction to this PR body.

### Verified sound, independently

- **Rename blast radius is complete.** An exhaustive sweep across Go, YAML, JSON, docs, fixtures and generated files found **every live consumer already on `"protobuf"`** — the registry key and `Language()` were the two outliers. `treesitter.SupportedLanguages()` has only test callers, so the list/registry asymmetry is invisible to any gate; `Extractor.Language()` has exactly one caller (`custom_dispatch.go:247`) which only formats an error string.
- **Leaving the persisted identifiers alone is right.** `refs.go:62-76` splits stubs into 6 segments and resolution reads only indices 1, 4 and 5 (`refs.go:2318-2331`) — **the `<lang>` segment at index 3 is never consulted.** A graph written before this change resolves identically after it.
- **The replacement gate is equivalent for protobuf**, mutation-verified: reverting the registry key with the row present fails and names `protobuf`; deleting the row *and* reverting the key passes silently, so the row is exactly what gates.
- **Entities are genuinely correct**, hand-checked against source: `lease.proto` yields exactly its 3 messages, all 5 fields with right types (`int64`, `etcdserverpb.LeaseTimeToLiveRequest`), 1 import, exact line numbers. `helloworld.proto` likewise. This is not plausible-looking garbage.

### Correction: the parse-failure diagnosis in this PR is wrong

I attributed the three failing files to "gogoproto custom options" and `edition = "2023"`. Probed directly against the grammar:

| input | error ratio |
|---|---|
| `option (gogoproto.marshaler_all) = true;` | **0.0000** |
| `syntax = "proto2";` (2-line file) | **0.2222** |
| proto2 + `optional` / `required` | 0.2105 / 0.2632 |
| `edition = "2023";` | 0.2500 |
| `rpc R(stream A) returns (stream B)` | 0.0000 |

**gogoproto options parse perfectly.** The real cause is that the bundled grammar accepts **proto3 only** — `syntax = "proto2"` and `edition = "2023"` are hard rejections. `record.proto` and `snap.proto` are plain proto2 with no gogoproto in them at all. The ratio pins at ~0.25 regardless of file size, because the whole body collapses into one ERROR node, so proto2 files are dropped **whole, every time**, and never dilute under the gate.

So the limitation is not "some options aren't covered" — it is **grafel's protobuf support is proto3-only**, which is far wider and differently shaped. Being corrected in the branch. This is the second wrong diagnosis I have put in a PR body today.

### Blocking: turning the extractor on injects phantoms

Two defects mean this PR is what starts putting unresolvable targets into every graph containing a `.proto` file.

- **Enum values emit `CONTAINS` to entities that are never created.** `enum E` emits edges to `…:E#E_UNSPECIFIED`, but no enum-value entity is ever built — contrast `buildField`, which emits one per message field. Subtype census: `map[endpoint:1 enum:1 field:3 import:1 message:2 service:1]`, those 3 being message fields only. The package doc at `proto.go:17-18` claims "enum → enum value", describing behaviour that does not exist.
- **Cross-file `REFERENCES` point into the wrong file.** `namedTypeRefs` (`proto.go:530-535`) strips the package qualifier and `proto.go:318` builds the ref against `file.Path`, so `etcdserverpb.LeaseTimeToLiveRequest` — defined in an imported file — targets a path that cannot exist.

Both are being fixed here rather than filed, because injecting a known-dangling target is not an acceptable interim state.

Also fixed in-branch: `internal/custom/cpp/protobuf.go` stamps `Language="proto"` on entities from the same files, and `RunCustomExtractors` runs independently of `extractors.Extract` — so post-PR one `.proto` file would yield entities under **both** stamps and anything filtering on `protobuf` would miss half.

### Being filed separately

- **`oneof` fields are dropped entirely** — `buildMessage` accepts only direct `field`/`map_field` children of `message_body`, and `oneof` children sit one level deeper. Zero entities, zero edges. `oneof` is ubiquitous in proto3.
- **rpc→message edges are bare-name, file-attributed and duplicated** — `{FromID: file.Path, ToID: reqType, Kind: "IMPORTS"}`, so the single most valuable fact in a proto file (`Greeter.SayHello` uses `HelloRequest`) does not link; identical request/response types emit the edge twice; and a missing `message_or_enum_type` child yields edges with literal `ToID: "?"`.
- **The 10% syntax-error gate hides partial garbage** — a proto3 file with one broken message among many parses at 5-6%, under the gate, and emits an entity for the broken message plus a field whose **name is wrong** (`field:M0.string` instead of `M0.b`). Silently incorrect entities are worse than none. No corpus file lands in that band today, but that is 11 files of evidence, not a guarantee.

### One more, not blocking

The gate's own failure text says *"register one, or drop the row from `classifierRepresentativeInputs`"* — inviting a future engineer to delete the regression gate for this bug, with nothing detecting the deletion. Being reworded.

Issue #6329, opened 2026-08-18
the generated-code tri-state is worth building for ranking, but the skip it was meant to fix has never run


cajasmota:
Pulled out of the VB.NET epic (#6327), because it is not a VB.NET problem. It is a problem in **five languages we shipped long ago**, and fixing it generally is worth more than fixing it for one language that has not shipped yet.

## The problem

The classifier is **binary**: index a file, or skip it entirely (`internal/classifier/classifier.go:201,240` → `matchGlobSkips`, reason `"generated"` at `:835`). 31 languages carry a `skip_patterns.yaml`, and **five of them skip generated source that declares symbols**:

| Language | Skipped generated source |
|---|---|
| Go | `*.pb.go`, `wire_gen.go`, mock files |
| C# | `*.Designer.cs`, `*.g.cs` |
| Python | `*_pb2.py` |
| JS/TS | generated clients |
| protobuf | generated output |

These are not build artifacts. **`*.pb.go` declares the types your handlers construct and pass around.** We skip the file at classification, so those types never enter the graph — and every reference to them from hand-written code becomes an unresolved stub.

That is not hypothetical. It lands in the `bug-extractor` disposition, which @manuel1358000 measured at **21.29% of all edges** in #6313. Some fraction of that is us skipping the declarations and then attributing the failure to the extractor.

The same shape appears in the other direction. Keeping generated code whole means machine-generated volume dominates the index — `InitializeComponent` in a WinForms designer file is thousands of near-identical property assignments. That is @manuel1358000's *other* issue, #6314: `grafel_find` bm25 ranking unrelated-domain matches above the relevant file. Flooding the index with generated entities is precisely the input that makes ranking worse, for every user.

**So the binary choice is wrong in both directions.** Skip and references dangle; keep and ranking drowns.

## The fix: three states, not two

| State | Behaviour | For |
|---|---|---|
| `index` | full extraction (default) | hand-written source |
| **`declarations`** | **parse declarations so references resolve; skip bodies; stamp `generated: true`; downweight in ranking** | `*.pb.go`, `*.Designer.cs`, `*_pb2.py`, `*.g.cs`, `wire_gen.go`, generated API clients |
| `skip` | never read | `zig-cache/`, `vendor/`, minified bundles, real build artifacts |

The middle state is the whole feature. The invariant it encodes is the same in every language: **generated code declares symbols that hand-written code references.**

[...]

PR #6362, opened 2026-08-19
feat(#6327): S3 — VB.NET pre-pass and per-file declaration table

cajasmota:
S3 of the VB.NET epic #6327: the pre-pass and per-file declaration table that S4's parser will run on.

**No entity is emitted by this story and no extractor is registered.** S2's behaviour is untouched — `.vb` still classifies `Skip=true / SkipReason=unsupported_language`, so the reporter's 672 files keep their `doctor`/`status` row until an extractor actually exists.

Refs #6327, #6321.

## Why a hand-written pre-pass at all

There is no tree-sitter grammar for VB.NET, so everything downstream is hand-written and its accuracy is decided here. The governing fact, from the epic's research: **`(` is two-way overloaded — index versus call — and `InvocationExpression` and `IndexExpression` are the same production in Microsoft's grammar.** Nothing can disambiguate `Foo(1)` without knowing what `Foo` is, which is why the symbol table is mandatory rather than an optimisation.

## Location: `internal/vbnet/`, not `internal/extractors/vbnet/`

Verified rather than assumed. `vbnet` is not in `languageRoster`; it appears in `tools/coverage` only as a failure-message example and a synthetic fixture. Creating `internal/extractors/vbnet/` would trip #6332's gate (merged as `590ec11ce`) — and satisfying that gate would publish a VB.NET coverage row for a language that emits nothing. Precedent for language machinery living outside `internal/extractors/`: `internal/generated/` (#6329) and `internal/treesitter/`. Confirmed by running `go test ./tools/coverage/` with the new directory present.

Six files, 157 passing subtests.

## Implicit line continuation: 16 of 21 positions, and the cost is in the suite

The mechanism is bracket-depth carry (which subsumes trailing `(`, interior `,` and a lone leading `)`), plus a trailing-operator set, a trailing-keyword set, and an attributes-only-line rule.

The gap is a fact rather than a claim: `ImplicitRuleCoverage` is an exported table of **all 21** documented positions with samples, and `TestImplicitRuleCoverage` drives every entry through the joiner **in both directions** — an unhonoured rule must still *fail* to join. Implementing one forces the table to be updated, and the gap cannot quietly change size.

The five omissions are `Let`, `Skip`, `Take`, `Distinct`, `Select` — bare LINQ contextual keywords that are also ordinary identifiers (`q.Take` alone on a line is a complete statement, and `Select` is the tail of `End Select`). **All five occur only inside method bodies, so none can split a declaration.** The cost lands on CALLS recall in S5, not on the declaration table this story exists to build.

## The declaration table

Twelve symbol kinds, each justified by a `ClassifyParen` answer it changes: `Namespace`, `Type`, `TypeParameter`, `Method`, `Property`, `Field`, `Const`, `Local`, `Parameter`, `Event`, `EnumMember`, `ImportAlias`.

- method / event / property → **call**
- type / type-parameter → **call** (conversion or constructor, never index)
- array field / local / parameter → **index**
- non-array value → **`ParenUnknown`**, because Default-property access and delegate invocation are genuinely indistinguishable at this layer — that is the epic's confidence signal, not a gap
- const / enum member → neither

`Resolve` picks the innermost enclosing declaration, so a local shadows a same-named field and the answer differs by scope.

XML doc comments are classified as `CommentXMLDoc` and carried on `LogicalLine.Doc` rather than discarded — `'''` is the standard doc form in the WinForms code #6321 is about, and distinguishing it costs one comparison.

## Mutants — six applied, all killed

`""` escape removed; comment scan not string-aware (12 subtests); `FoldName` case-sensitive (**53 failures across all four test files** — independently re-verified); explicit `_` joining removed; `KindMethod` dropped from the call arm; parameters not recorded.

**One mutant I specified was wrong, and the agent said so rather than reporting it as passing.** I asked for a mutant that "stops treating `''` inside a string as an escaped quote". In VB.NET `''` is two apostrophes — the escape is `""`. That mutant survived at exit 0, because `"a '' b"` exercises a different rule (a `'` inside a literal does not start a comment) and **nothing in the suite tested `""` at all**. Two cases were added (`Dim s = "a""b"` masking as one span; a `(` between `""` escapes not raising bracket depth) and the mutant split in two. Both now die. Textbook fixture vacuity, in the brief rather than the code.

## Evidence

**The pre-pass is unverified against real VB.NET source.** No `.vb` file exists on this machine — established during S2. Every fixture is constructed from the language reference and the shapes reported in #6321. The mechanism is pinned; the distribution is not, and cannot be locally. **Nothing here claims recall or precision on the reporter's ~670 files.** The 16/21 figure counts language-reference *positions*, not occurrences in real code.

Known limitations, flagged for S4 rather than left to be rediscovered:
- `$"..."` interpolated-string interiors are scanned as ordinary literals — a `"` inside a `{}` hole would confuse the scanner.
- Accessor containers are named by the folded keyword (`...MainForm.set`), because a property opens no container.
- `Handles` / `Implements` / `Inherits` clauses are stripped and discarded, not recorded — S5 and S7 own those edges.

### DIFF
--- a/internal/vbnet/attributes.go
+++ b/internal/vbnet/attributes.go
@@ -1,0 +1,116 @@
+package vbnet
+
+import "strings"
+
+// SplitAttributes removes attribute groups from a logical line and returns
+// their bodies (without the angle brackets) in source order, plus the code
+// with those groups removed.
+//
+// Two positions are recognised, and only two, because they are the only two
+// where '<' cannot be a less-than operator:
+//
+//	<Serializable()> Public Class X      ' leading, may repeat
+//	Sub F(<Out> ByRef n As Integer)      ' immediately after '(' or ','
+//
+// Attribute bodies nest parentheses (<Attr(GetType(List(Of T)))>) and may
+// contain string literals holding '>' — both are tracked, so the terminating
+// '>' is the one at paren depth zero outside a literal.
+func SplitAttributes(code string) (attrs []string, rest string) {
+	var out strings.Builder
+	i := 0
+
+	// Leading groups: skip whitespace, then consume every adjacent <...>.
+	for {
+		j := i
+		for j < len(code) && (code[j] == ' ' || code[j] == '\t') {
+			j++
+		}
+		if j >= len(code) || code[j] != '<' {
+			break
+		}
+		end, ok := attributeEnd(code, j)
+		if !ok {
+			break
+		}
+		attrs = append(attrs, strings.TrimSpace(code[j+1:end]))
+		i = end + 1
+	}
+
+	// Inline groups: '<' that directly follows '(' or ',' (whitespace aside).
+	for i < len(code) {
+		c := code[i]
+		if c == '"' {
+			// Copy the literal verbatim; '<' inside it is data.
+			end := literalEnd(code, i)
+			out.WriteString(code[i:end])
+			i = end
+			continue
+		}
+		if c == '(' || c == ',' {
+			out.WriteByte(c)
+			i++
+			j := i
+			for j < len(code) && (code[j] == ' ' || code[j] == '\t') {
+				j++
+			}
+			for j < len(code) && code[j] == '<' {
+				end, ok := attributeEnd(code, j)
+				if !ok {
+					break
+				}
+				attrs = append(attrs, strings.TrimSpace(code[j+1:end]))
+				// Consume the group but not the whitespace around it, so that
+				// removing an attribute never joins two tokens together.
+				i = end + 1
+				j = i
+				for j < len(code) && (code[j] == ' ' || code[j] == '\t') {
+					j++
+				}
+			}
+			continue
+		}
+		out.WriteByte(c)
+		i++
+	}
+	return attrs, strings.TrimSpace(out.String())
+}
+
+// attributeEnd returns the index of the '>' closing the attribute group that
+// opens at code[start], which must be '<'.
+func attributeEnd(code string, start int) (int, bool) {
+	depth := 0
+	for i := start + 1; i < len(code); i++ {
+		switch code[i] {
+		case '"':
+			i = literalEnd(code, i) - 1
+		case '(', '[', '{':
+			depth++
+		case ')', ']', '}':
+			depth--
+		case '>':
+			if depth <= 0 {
+				return i, true
+			}
+		}
+	}
+	return 0, false
+}
+
+// literalEnd returns the index one past the string literal beginning at
+// code[start], which must be '"'. An unterminated literal ends at the line
+// end. "" inside the literal is an escaped quote.
+func literalEnd(code string, start int) int {
+	i := start + 1
+	for i < len(code) {
+		if code[i] != '"' {
+			i++
+			continue
+		}
+		if i+1 < len(code) && code[i+1] == '"' {
+			i += 2
+			continue
+		}
+		return i + 1
+	}
+	return len(code)
+}
--- a/internal/vbnet/attributes_6327_test.go
+++ b/internal/vbnet/attributes_6327_test.go
@@ -1,0 +1,99 @@
+package vbnet
+
+import (
+	"reflect"
+	"testing"
+)
+
+// #6327 S3 — attribute stripping.
+//
+// '<' is overloaded in VB.NET exactly as '(' is: it opens an attribute list
+// and it is the less-than operator. The negatives below are the rules that
+// keep `If a < b Then` from being eaten as an attribute.
+//
+// UNVERIFIED against real VB.NET source; see the package doc.
+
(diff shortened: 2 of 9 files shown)


cajasmota:
## Independent re-review: REQUEST-CHANGES — one blocker, and three guards that held

### The three continuation guards survived 22 constructed counterexamples

- **`option` head guard** — no counterexample found. `takeWord` peels only ident bytes, so `[Option]` → `""`, `.Option` → `""`, `Options` → `"options"`, `Optional` → `"optional"`. `Option` is reserved, so any legitimate identifier use must be bracketed, which the guard is already blind to.
- **`End` guard** — `precedingWord` does a real word-boundary walk, not a suffix check: `Backend`, `Append`, `x.TrimEnd` and `[End]` all still join. The single refusal needs a member literally named `End`, which is illegal unbracketed and works bracketed.
- **`&` type-character guard** — correct on every shape the doc names.

### BLOCKER — `readType` uses `LastIndexByte('(')`

```
"Private s As New StreamReader(New FileStream(p))"
  got  type="StreamReader(New FileStream(p))"
  want type="StreamReader"
```

It finds the **innermost** open paren rather than the one matching the trailing `)`, so `matchBracket` returns an index ≠ `len(s)-1`, the loop breaks, and the whole tail survives. `New SqlConnection(GetConnString())`, `New Timer(New TimeSpan(0,0,5))`, `New StreamReader(Path.Combine(a,b))` all defeat it. The previous round's fix only ever worked on flat argument lists — `New StreamReader("path")` passes solely because masking flattens the literal. Reaches `parseParams`, `parseDeclarators`, `parseProperty` and `parseAsClauseLocal`, all of which funnel through `readType`.

The deeper point: **the algorithm cannot actually distinguish `List(Of String)` (keep) from constructor arguments (strip).** The generics case works by the same accident that breaks the nested case. The fix needs to state the real rule — `(Of …)` is a type-argument list, anything else after a `New` type name is constructor arguments — rather than leave a discrimination to be inferred that isn't there.

### Should fix — `New T(n) {}` loses `IsArray`

```
"Private b As New Byte(255) {}"       got "Byte(255) {}"      array=false
"Private n As New Integer() {1,2,3}"  got "Integer() {1,2,3}" array=false
```

The standard VB.NET array-creation form. The garbage `TypeName` is the lesser problem: **`IsArray` is exactly what `ClassifyParen` uses to answer `ParenIndex` instead of a phantom `ParenCall`**, which is the failure mode #6327 exists to prevent. The doc comment claims initialisers are dropped; they are not.

### Should record — `ParenUnknown` over-suppresses two real shapes

The new check cannot suppress anything `Resolve` returned via its `best` path (it re-runs the same predicate), so all over-suppression is on the fallback. Two probes went from correct `ParenCall` to `unknown`:

- **Module member promotion** — a `Module`'s members are unqualified-visible file- and project-wide, and Modules are pervasive in the legacy VB this story targets.
- **Same-file inheritance** — `Class B Inherits A` with both in one file is ordinary VB.

`visibleFrom` is pure lexical containment and models neither. Safe direction, not a correctness regression — but it is the cost the fix was supposed to be measured against, and it is currently unrecorded.

### Should record — the `&` guard splits where it provably shouldn't

VB.NET forbids the `&` (Long) type character on a `Char` literal and on a fractional literal, so a glued `&` there **must** be concatenation. `"x"c&` and `1.5&` both split, because `isIdentByte` sees the `c` and the `5`. Conservative direction, but the doc asserts the case cannot exist.

### Equivalent-mutant claim: CONFIRMED

The `splitStatements` `:` fast-path guard was declared unkillable rather than given a fake test. Verified by differential — the guarded and unguarded implementations compared through `walkStatement`'s own normalisation over **200,000 random colon-free inputs: 0 diffs**. With no `:`, `splitTopLevel` returns exactly `[]string{TrimSpace(s)}`, and the only difference is whitespace that `walkStatement` trims on entry.

Declaring an equivalent mutant, with the reasoning, is the correct alternative to inventing an assertion that passes for unrelated reasons. Confirmed rather than suspected, and the in-code note should stay.

### A claim I should not have made

`internal/vbnet` grew 156 → 205 test entries, and I described that as improving independence. Re-derived: `FoldName` case-sensitive kills **90/205 (43.9%)**, up from 60/156 (38.5%). The new tests are slightly *more* coupled to it, not less. A high blast radius is partly inherent for the load-bearing primitive of a case-insensitive language — that is fair to say; an independence win is not.

Also unpinned: mutating `MaskStringLiterals` to the identity kills 11 entries and **not one is a `BuildTable` test**, though `BuildTable` masks every logical line and `strip.go` justifies masking precisely so that "a `(` or an identifier appearing inside a literal cannot be mistaken for code". That claim has no test at the layer it describes.

Issue #6347, opened 2026-08-19
four changes this cycle passed their fixtures and were wrong on a real repo — propose a real-tree measurement rule in AGENTS.md


cajasmota:
Four times in this cycle a change passed its own tests, passed review, and was **wrong in a way only a real repository could show**. Every one was caught late or by luck. They are the same defect, so they deserve one rule rather than four postmortems.

## The four

| # | what the tests exercised | what a real tree showed |
|---|---|---|
| **#6338** | hand-picked lists of source files | on a real 31,820-file monorepo the "Unsupported languages" report was **60 rows** — `.json 1938`, `.xml 1531`, `.ds_store 11` — and the reporter's actual `.vb 672` line was buried in it |
| **#6342** | the same fixtures | shipped green **over** that report |
| **#6332** | `SupportedLanguages` against a synthetic `t.TempDir()` tree | nothing binds it to the real `internal/extractors/`, so a new language silently renders as **"Vbnet"** via the title-case fallback |
| **#6345** | 17 hand-checked marker hits across two repos | the same detector over `~/go/pkg/mod` flagged **8,236 hand-written files** in three systematic classes |

A fifth, adjacent: **#6329**'s own filing argued a mechanism confidently and measured to a **0-edge delta** — the code path was never executed.

## Why fixtures fail here specifically

grafel's inputs are *repositories*, and the properties that matter are distributional: how many rows a report emits, what fraction of files carry a marker, which extensions actually appear, what a tree contains that nobody thought to put in a fixture. A hand-built fixture encodes what the author already expected. That is exactly the wrong instrument for finding what they did not.

Note that #6345 was **not** careless — the author hand-checked every one of 17 hits, which is real verification. The gap was between *"I verified every hit I found"* and *"I looked in enough places to find the hits that matter."* A fixture suite cannot close that gap no matter how carefully each case is checked, because the sample is the problem, not the checking.

## Proposed rule for `AGENTS.md` → "Tests + gates"

> **Any change whose behaviour depends on the shape or distribution of a repository must be measured against at least one real tree before it ships, and the measurement recorded in the PR body.** Fixtures pin the mechanism; the real tree establishes the distribution. Report a count, not an impression. If the change alters what users are *shown* — a report, a ranking, a coverage table — the measurement must be of the rendered output, not of the function that feeds it.

The last sentence is what #6338 and #6342 both missed: the underlying classifier was correct and well-tested; the **report built from it** was unusable, and no test looked at the report.

[...]

Issue #6134, opened 2026-08-04
TestGroupAlgoProgress_StopsWithThePass fails under load: test-only race, or the heartbeat really outliving its pass?


cajasmota:
`TestGroupAlgoProgress_StopsWithThePass` (`cmd/grafel`) fails under host load and passes in isolation. Surfaced twice during #6107 review work, on branches that touch none of the code involved.

```
heartbeat kept ticking after the pass ended (1 -> 2 lines)
```

`go test -count=1 -p 1 ./cmd/grafel/` → exit 1 at host load ~15-25.
`go test -count=3 -run '^TestGroupAlgoProgress_StopsWithThePass$' ./cmd/grafel/` → exit 0.

Reproduced independently on `fix/6107-peak-heap` @ `5edf51fd8` and confirmed present on `origin/main` @ `9851adafb` — this is not branch-specific.

## Why it matters beyond the noise

The assertion is that the heartbeat emits **no further lines after the pass ends**. Under load it emits one more. Two readings, and they have opposite severity:

1. **Test-only race** — the test samples the line count before the heartbeat goroutine has observed the stop signal. Harmless; the fix is to synchronise on shutdown rather than on a count.
2. **Real shutdown race in the product** — the heartbeat genuinely outlives the pass it reports on, and a slow enough machine makes it observable. That is the class #6047 was filed for (progress reported after the work it describes has ended), and it would mean the wizard can still print a heartbeat for a finished pass.

Which one it is has not been determined. The distinction is the whole issue — do not close this by adding a retry or a sleep to the test until reading 2 has been ruled out.

## Context

Found while verifying an unrelated branch, so it is unowned. Filed rather than fixed to keep it from being absorbed into a PR that has nothing to do with it — and because a flake that only appears above a load threshold is exactly the kind that gets re-diagnosed from scratch by the next person who trips it.

Refs #6047, #6108

cajasmota:
**A second test of the same shape, found the same way — plus a hygiene defect in it that is worth more than the flake.**

`TestShutdownAcceptBound` (`internal/daemon/shutdown_accept_bound_test.go:103`):

```
Run returned after 303.777ms (watchdog=300ms)
```

A 300 ms watchdog missed by **3.7 ms** under host load. Passes 3/3 in isolation. Surfaced on a branch whose only change in that package is a single struct-field assignment, so it is not branch-specific.

[...]

Issue #5277, opened 2026-06-16
SAFETY: 'grafel uninstall' stops the GLOBAL live daemon even under GRAFEL_TEST_REQUIRE_ISOLATED_HOME (launchd label not HOME-scoped)


cajasmota:
## What happened
A sandboxed audit run did an isolated-HOME `grafel uninstall --yes --purge --remove-binary` (HOME=/tmp/..., GRAFEL_TEST_REQUIRE_ISOLATED_HOME=1). The `--remove-binary`/`--purge` correctly hit only the /tmp install (sandbox install.json), BUT the uninstall's **stop-daemon** step did `launchctl bootout gui/$uid/com.grafel.daemon` on the **host-global launchd label** — taking down the developer's LIVE daemon AND unloading the job so KeepAlive didn't respawn it. MCP went down until a manual `launchctl bootstrap`.

## Root cause
The daemon lifecycle (launchd label `com.grafel.daemon`, and likely the Windows named-pipe `service` label) is **host-global, not scoped by HOME / GRAFEL_DAEMON_ROOT**. So `grafel uninstall` stops/unregisters the global daemon regardless of an isolated HOME. The `GRAFEL_TEST_REQUIRE_ISOLATED_HOME=1` guard (which fail-closes go-test access to the live daemon) does NOT gate the CLI uninstall's daemon-stop.

## Why it matters (real users, not just our audit)
Any user running `grafel uninstall` in a scratch/isolated HOME (e.g. testing, CI on a dev box, a second sandboxed install) will nuke their primary daemon. The isolation contract is silently broken for the daemon lifecycle.

## Fix options
1. When `GRAFEL_TEST_REQUIRE_ISOLATED_HOME=1` (or HOME != the daemon's registered HOME), `grafel uninstall` must NOT stop/unregister the global daemon — only tear down the isolated install's artifacts. Fail-closed: refuse to touch a daemon whose registered root != the current isolated root.
2. Scope the launchd label / service name by daemon root (e.g. `com.grafel.daemon.<roothash>`), mirroring the #5269 root-scoped named-pipe fix, so isolated installs manage their own service and can't bootout the primary.
3. At minimum, `uninstall` should only stop the daemon whose `install.json`/registered root matches the install being removed.

Prefer (2)+(3): root-scope the service identity (consistent with #5269) AND match-on-registered-root before stopping. P1 — it's a footgun that already caused a live outage.

PR #6364, opened 2026-08-19
fix(#6331): partial isolation silently targeted the real ~/.grafel — refuse on XOR, warn on both-with-real-HOME

### LINKED ISSUE #6331: GRAFEL_HOME does not isolate the daemon, so a sandboxed grafel index silently targets the real installation
Hit this for real while running a measurement: an agent set `GRAFEL_HOME` to a temp directory believing that isolated it, ran `grafel index`, and **dialled the developer's live daemon instead**. A subsequent `grafel start` went through launchd and stopped the running daemon.

No data was lost — the daemon was restarted and the registered group was intact — but the failure mode is bad: you believe you are working in a sandbox and you are operating on the real installation, with nothing telling you otherwise.

## The trap

`GRAFEL_HOME` is not the isolation variable. The daemon layout is rooted at **`GRAFEL_DAEMON_ROOT`** (`internal/daemon/paths.go:36`). With `GRAFEL_HOME` set and `GRAFEL_DAEMON_ROOT` unset, `client.Dial` → `daemon.DefaultLayout` resolves to the **real** socket at `~/.grafel/sockets/daemon.sock`.

The names actively mislead: `GRAFEL_HOME` reads like the one that would redirect everything, and it is the variable the repo's own docs and test helpers reach for most visibly.

## Suggested fix

Make the mismatch impossible to hit silently. Options, roughly in order of preference:

1. **Derive the daemon root from `GRAFEL_HOME` when `GRAFEL_DAEMON_ROOT` is unset.** Setting `GRAFEL_HOME` would then isolate what people already believe it isolates.
2. If they must stay independent, **refuse to dial the default socket when `GRAFEL_HOME` points somewhere other than the real home** — error with a message naming `GRAFEL_DAEMON_ROOT` explicitly.
3. At minimum, have `grafel index`/`start`/`stop` print which daemon root and socket they resolved to when `GRAFEL_HOME` is set to a non-default value.

## Two adjacent findings from the same incident

- **`grafel index` has no in-process fallback and no timeout** (`internal/cli/index.go:18`). It blocks forever on the daemon RPC if the daemon is wedged — observed sitting at 0% CPU for 10 minutes with no output under `--quiet`. A timeout, or at least a "waiting for daemon" line after a few seconds, would turn a hang into a diagnosis.
- The live daemon reported the corpus graph as unreadable: `graph.fb format version 5 is older than required version 6`. Any measurement taken against that group reads an incomplete graph, and nothing about the reported entity counts makes that obvious. Worth checking separately whether a stale-format graph should be a louder condition than it currently is — it silently changes what every query returns.

cajasmota:
`GRAFEL_DAEMON_ROOT` moves the socket, pidfile and logs. `GRAFEL_HOME` moves the store. **Neither alone is isolation**, and every guard that says so was test-only — so running the real binary with one of them set silently wrote to the live `~/.grafel`, with no warning and no failure.

That is how a live MCP daemon serving a user's session went down during this cycle: the operator briefed the wrong variable and the CLI had no opinion about it.

Closes #6331.

## The asymmetry, not a missing check

`GuardRealHome` / `GuardRealHomeMain` take a `*testing.T` or run from `TestMain`; `homescan.go` is a static scan over test files; `internal/install/daemon_guard.go` is production but scoped to **uninstall** (#5277). There was nothing on the `grafel index` path.

So the surface where a mistake is *cheap* — a test run you rerun — was hardened, and the surface where it is *expensive* was left open. The knowledge was never missing either: `internal/testsupport/guard_main.go:31-55` spells out the full incantation under #6134. It just was not reachable from where the mistake gets made.

## Refuse on XOR, warn on both-with-real-HOME

New `internal/envguard`, wired on the root's `PersistentPreRunE` (previously unclaimed — a test now fails if a subcommand defines one).

- **XOR (exactly one set) → refuse.** Never intentional; no caller anywhere does it on purpose.
- **Both set with a real `HOME` → warn, not refuse.** This is against the direction I was asked to lean, and the evidence is why: `scripts/phase-b-bench.sh:16-17` does exactly this deliberately, because it wants a real launchd context. The store and daemon plane are both redirected there; the residue is only `HOME`-derived config. Refusing would have broken a legitimate caller to prevent a non-problem.
- `GRAFEL_ALLOW_PARTIAL_ISOLATION=1` downgrades a refusal to a warning, never silences it.

Exempt: `help`, `completion`, `__complete*`, `statusline` (runs in the user's shell prompt), `version`. **Read-only commands are not exempt** — a partially-isolated `grafel status` reports the wrong store authoritatively, which is the same defect wearing a quieter hat. The pre-cobra entrypoints in `main.go` (`index-internal`, `group-algo`, `links-internal`, `xrepo-verify`, `selftest`) are deliberately outside the guard.

## Every production setter was checked, and one was broken

| setter | shape | verdict |
|---|---|---|
| launchd plist, systemd unit | `HOME` only | OK — installer unaffected |
| `defaultEngineChildCommand`, `sched/subprocess_runner.go` | `os.Environ()` verbatim | inherits the parent's verdict |
| `cmd/grafel/selftest.go:206-235` | HOME + both vars + XDG | fully isolated |
| `scripts/phase-b-bench.sh:16-17` | both set, real HOME, **deliberately** | the legitimate caller above |
| **`.github/workflows/windows-cgo-experiment.yml:64`** | **`GRAFEL_DAEMON_ROOT` only, on a real `grafel.exe daemon`** | **was broken — the #6134 shape in our own CI. Fixed to set both.** |

## Observed behaviour, real binary built from this tree

```
GRAFEL_HOME only, real HOME, `grafel list`   → exit 1: "refusing to run: PARTIALLY ISOLATED …
    the STORE is redirected but the DAEMON PLANE is not: this command will dial the REAL
    daemon socket (~/.grafel/sockets/daemon.sock)"
GRAFEL_DAEMON_ROOT only, `grafel status`     → exit 1, names GRAFEL_HOME and #6134
both set, real HOME                          → warning naming the real home, then exit 0
fully isolated                               → silent, exit 0        (the anti-blanket control)
escape hatch set                             → warning, proceeds, exit 0
`grafel help` under partial isolation        → exit 0 (exempt)
```

## The guard immediately found real instances in our own suite

**Ten `internal/cli` tests** drove the cobra root under `GRAFEL_HOME`-only and now fail closed. `setupRefTestEnv` (`ref_flag_test.go:36`) already carried a comment about a prior incident where that exact shape globbed the developer's **real 140 watcher plists**. All ten given a short daemon root — short because `<root>/sockets/daemon.sock` must clear the 104-byte `sun_path` cap.

## Mutants

- XOR branch → `VerdictOK`: 9 failures including `grafel list ran under GRAFEL_HOME-only isolation; the guard did not refuse`. Independently re-verified.
- **Blanket refuse**: `TestIsolateHomeProducesVerdictOK` and `TestGuardAllowsFullIsolation` fail — the anti-blanket control bites, so the fix cannot degrade into "nothing runs anywhere".
- Unwire `installIsolationGuard(root)`: `TestRootWiresIsolationGuard` fails.
- `RealUserHome` via `os.UserHomeDir`: fails with `"/Users/jorgecajas" -> "/var/folders/.../001"`, proving the both-set leg needs `user.Current()`, which ignores `$HOME`.

## Not verified, and out of scope

- **Windows and Linux are unverified** — `sameDir` case-folding and `user.Current()` were exercised on darwin only. CI is the gate.
- ~100 other `internal/cli` tests set `GRAFEL_HOME` only but call `runXxx` directly rather than through cobra, so they pass unchanged. They remain partially isolated and would dial the live socket if they ever reached the daemon path — a separate sweep.
- Adjacent findings in the issue body left alone: `grafel index` has no timeout on the daemon RPC (`internal/cli/index.go:18`), and the stale-graph-format condition is still silent.

Refs #6134, #5277, #6347.

### DIFF
--- a/.github/workflows/windows-cgo-experiment.yml
+++ b/.github/workflows/windows-cgo-experiment.yml
@@ -61,7 +61,13 @@ jobs:
       - name: smoke test — start daemon, hit healthz, stop
         shell: powershell
         run: |
+          # GRAFEL_DAEMON_ROOT *and* GRAFEL_HOME (#6331). GRAFEL_DAEMON_ROOT
+          # alone moves the socket/pid/log but leaves the store on the real
+          # home, and the daemon's startup tail relocates and prunes that store
+          # (#6134). internal/envguard refuses that shape, so this step would
+          # not start a daemon at all with only one of the two set.
           $env:GRAFEL_DAEMON_ROOT = "$env:RUNNER_TEMP\grafel-test"
+          $env:GRAFEL_HOME = "$env:RUNNER_TEMP\grafel-test"
           New-Item -ItemType Directory -Path $env:GRAFEL_DAEMON_ROOT -Force | Out-Null
           $proc = Start-Process -FilePath ".\grafel.exe" -ArgumentList "daemon" -PassThru -RedirectStandardOutput "daemon.log" -RedirectStandardError "daemon.err"
           Start-Sleep -Seconds 5
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -58,11 +58,17 @@ When your PR adds, modifies, or fixes a capability that's tracked in the coverag
 - Claims about numbers come from a real measurement — see "Evidence"
 
 ## Daemon discipline
-- If you spawn a daemon for testing, set `GRAFEL_DAEMON_ROOT=/tmp/arch-<task>` and stop it on exit
+- If you spawn a daemon for testing, isolate ALL THREE of `HOME`, `GRAFEL_HOME` and `GRAFEL_DAEMON_ROOT`, and stop it on exit. `GRAFEL_DAEMON_ROOT` alone is NOT isolation, and the CLI now refuses it (#6331):
+
+  ```sh
+  export HOME=$(mktemp -d)
+  export GRAFEL_HOME=$HOME/.grafel
+  export GRAFEL_DAEMON_ROOT=$HOME/.grafel
+  ```
 - Verify no PIDs survive with `ps aux | grep grafel`
 - Never `git stash` (concurrent worktree race; commit-checkpoint instead)
 - See `docs/adrs/0004-single-mcp-process-per-machine.md` for the daemon architecture
-- `GRAFEL_DAEMON_ROOT` isolates THREE things: the daemon socket, the registry, AND per-repo state (issue #745). When the env var is set, per-repo state lives at `$GRAFEL_DAEMON_ROOT/state/<sha256(abs_repo_path)[:16]>/` instead of `<repo>/.grafel/`. This means two parallel agents can index the SAME fixture without racing, and the fixture's own `.grafel/` is never touched. When the env var is unset, ADR-0007 co-located behavior is preserved. Helper: `internal/daemon.StateDirForRepo` / `GraphPathForRepo` — use it for every per-repo state read/write; never hardcode `<repo>/.grafel/<file>`.
+- `GRAFEL_DAEMON_ROOT` isolates the daemon socket (plus pidfile and logs) AND per-repo state (issue #745). It does **NOT** isolate the registry — this bullet used to claim it did, and so did ADR-0017; both were wrong. `registry.HomeDir` reads `GRAFEL_HOME` only, so `GRAFEL_DAEMON_ROOT` alone gives you a private daemon pointed at the LIVE store, whose startup tail then relocates and prunes it (#6134). See the 2026-08-19 amendment in `docs/adrs/0017-single-binary-daemon-architecture.md`. When the env var is set, per-repo state lives at `$GRAFEL_DAEMON_ROOT/state/<sha256(abs_repo_path)[:16]>/` instead of `<repo>/.grafel/`. This means two parallel agents can index the SAME fixture without racing, and the fixture's own `.grafel/` is never touched. When the env var is unset, ADR-0007 co-located behavior is preserved. Helper: `internal/daemon.StateDirForRepo` / `GraphPathForRepo` — use it for every per-repo state read/write; never hardcode `<repo>/.grafel/<file>`.
 
 ## Where things live
 - MCP server: `internal/mcp/`
--- a/docs/adrs/0017-single-binary-daemon-architecture.md
+++ b/docs/adrs/0017-single-binary-daemon-architecture.md
@@ -152,7 +152,7 @@ Users on any prior binary reinstall via `grafel install`. There is no compat shi
 
 ## Amendment (2026-05-20) — Per-repo state isolation under `GRAFEL_DAEMON_ROOT` (issue #745)
 
-When parallel coding agents each run an isolated daemon (different `GRAFEL_DAEMON_ROOT`s) against a shared fixture corpus, the socket and registry are isolated but the per-repo state directory — `<repo>/.grafel/` per ADR-0007 — is shared. Two daemons indexing the same fixture race on `graph.json` and corrupt each other's outputs.
+When parallel coding agents each run an isolated daemon (different `GRAFEL_DAEMON_ROOT`s) against a shared fixture corpus, the socket is isolated (this amendment originally said "the socket and registry" — that was wrong; see the 2026-08-19 amendment below) but the per-repo state directory — `<repo>/.grafel/` per ADR-0007 — is shared. Two daemons indexing the same fixture race on `graph.json` and corrupt each other's outputs.
 
 This amendment narrows the daemon's reach into the repo working tree:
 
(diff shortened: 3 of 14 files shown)


cajasmota:
## Independent adversarial review: REQUEST-CHANGES

All four suites green (`envguard`, `cli`, 18 `daemon` packages, `cmd/grafel`), the mutant-kill is real, and the design holds. Four findings.

### F1 — a whitespace-only value bypasses the guard, and a test pins the hole open

`envguard.go:110-119` applies `strings.TrimSpace` before testing `!= ""`. **No consumer does.** `internal/registry/registry.go:237` is `if override := os.Getenv("GRAFEL_HOME"); override != ""`, and `paths_unix.go:50` / `paths_windows.go:33` are the same shape. So with `GRAFEL_HOME="   "` the guard says *unset* while the registry says *set*:

```
$ GRAFEL_HOME=" " ./grafel list
No groups registered. Run `grafel wizard` to create one.
```

No refusal, no warning. The store resolved to a literal `"   "` directory — hence the empty registry — while `daemon.DefaultLayout` still pointed at the **live** `~/.grafel/sockets/daemon.sock`. That is byte-for-byte the #6331 incident shape. `GRAFEL_DAEMON_ROOT="   "` is the mirror-image #6134 shape.

It is asserted rather than overlooked: `envguard_6331_test.go:95-99` pins `GRAFEL_HOME: "   "` → `VerdictOK`.

**The guard's notion of "set" must be the consumers' notion of "set", or it is checking a different environment than the one the program runs in.**

### F2 — a no-op override is refused, and the message is false

Both partial legs are pure presence checks with no path awareness:

```
$ GRAFEL_HOME="$HOME/.grafel" ./grafel list ; echo rc=$?
refusing to run: PARTIALLY ISOLATED environment.
  GRAFEL_HOME="/Users/jorgecajas/.grafel" is set but GRAFEL_DAEMON_ROOT is NOT.
  Consequence: the STORE is redirected but the DAEMON PLANE is not ...
rc=1
```

Nothing is redirected. A refusal whose **diagnosis is wrong** is the worst kind, because the user cannot reason their way out of it — and `export GRAFEL_HOME="$HOME/.grafel"` in a shell profile produces exactly this. Setting both to the real defaults yields a spurious warning on every command's stderr for the same reason.

### F3 — the PR refuses a workflow our own ADR still documents as supported

`docs/adrs/0017-single-binary-daemon-architecture.md:153-170` says of `GRAFEL_DAEMON_ROOT` **alone**: *"the socket and registry are isolated"* … *"the single switch for the full three-dimensional isolation (socket + registry + per-repo state)."* `AGENTS.md:65` repeats it.

The ADR is **wrong about the registry** — `registry.HomeDir()` (`registry.go:236-245`) reads only `GRAFEL_HOME` and never `GRAFEL_DAEMON_ROOT` — so the refusal is justified. But shipping it without correcting the documents means an agent follows `AGENTS.md`, is refused, and finds the repo's own architecture record contradicting the error message.

This is the third instance in this cycle of the repo already containing the truth and nothing acting on it: `classifier_test.go:790` documented the protobuf name mismatch **with the exact fix** in a test that only called `t.Log` (#6357), and `setupRefTestEnv` carried a comment about a prior incident where partial isolation globbed the developer's real 140 watcher plists. Written-down knowledge that no gate enforces does not change behaviour.

### F4 — `--help` is refused on the seven `DisableFlagParsing` commands

Cobra's help short-circuit precedes `PersistentPreRunE` only when it parses flags. Seven commands set `DisableFlagParsing: true`, so help reaches the guard:

```
index --help  rc=1 REFUSED        status --help rc=0 | mcp --help rc=0 | help advanced rc=0
doctor        rc=1 REFUSED
```

`grafel index --help` printing a refusal instead of usage violates the exemption doc's own principle (`isolation_guard.go:17-19`) that introspection must never fail.

### Exemption list

`help`, `completion`, `__complete*` verified necessary and correct. `statusline` correctly exempt (and verified not strictly needed — the printed snippet resolves the path itself, so there is no per-keystroke binary invocation).

**`doctor` is missing and should warn rather than refuse** — it is the one command a user runs to diagnose exactly the state being complained about, and `internal/install/copy.go:240` says so in as many words. **`version` is dead weight**: no such subcommand exists; `--version` is a cobra flag handled before hooks.

### Cleared

Escape hatch parses strictly — `0`, `false`, `true`, unset all refuse correctly, with `true` pinned by a test. No repeat of #6349's `GRAFEL_PHASE_TRACE=0`-turns-it-on defect.

`PersistentPreRunE` wiring verified by execution: adding a hook to `install-hooks` fires `TestNoSubcommandShadowsTheGuard`, naming the offending command. Deep nesting resolves correctly.

Git hooks wrap grafel calls in `|| true`, so a refusal degrades to a silent no-op rather than blocking a push. The launchd plist sets only `HOME`, so no existing installation's daemon breaks. Short-daemon-root sizing is 82 bytes against the 104-byte `sun_path` cap.

**My path-comparison hypothesis was disconfirmed.** I predicted that was the defect class; the both-set leg can only return `VerdictOK` or `VerdictWarn` and **structurally cannot refuse**, so every miss there is under-warning in the safe direction. Unset `HOME`, trailing slashes, symlinks, `..` components and a failing `user.Current()` were all checked and none can manufacture a refusal.

PR #6297, opened 2026-08-12
fix(#6295): a Solidity contract had no outbound EXTENDS edge, because the `is` clause was anchored to the file

arthurgeron:
## Summary

The Solidity EXTENDS emitter passed the extractor's file argument as `FromID`
(`internal/extractors/solidity/extractor.go:250-255` at the parent commit). That
value is non-empty and non-hex, so `ReferencesEmbeddedWithAllowlist` rewrites it
(`internal/resolve/refs.go:5986-5989`), and `FileEntity` names the file entity
with the same path (`internal/extractor/extractor.go:326-333`, mechanism spelled
out in its own doc comment at 318-320). Every rewrite therefore landed on the
file component. The contract had no outbound EXTENDS edge at all, and where one
`.sol` file declared several contracts their base lists merged onto that one
node.

The fix removes `FromID`. An empty `FromID` is the convention the other
type-anchoring extractors already use, and both assembly paths substitute the
owning record's own entity id:

- `cmd/grafel/index.go:6063-6069` (full index) runs `if fromID == "" { fromID = id }`,
  with `id` the record's `graph.EntityID` from `index.go:5921`.
- `internal/extractors/incremental.go:2215-2219` (`relRecordToGraphRel`) does the
  same, and its doc comment says it exists to mirror the full path for exactly
  this case (#6094, #6098).

Neither is conditional. `CONTAINS` on the same record (`extractor.go:310-313`)
already depended on it, and the package doc already named the intended endpoint:
``EXTENDS edges (on the contract component)`` (`extractor.go:10`).

`IMPORTS` in the same extractor keeps passing `filePath`, which `refs.go:3658-3669`
treats as the deliberate cross-language convention for import edges (#120). An
import statement belongs to a file; an `is` clause belongs to a type.

## Closes / Refs

Closes #6295

## Testing

- `go test -count=1 ./internal/extractors/solidity/ ./internal/resolve/ ./internal/extractors/sresolver/`:
  ok, all three.
- `go vet ./internal/extractors/solidity/`: clean. `gofmt -l internal/extractors/solidity/`:
  no output.
- The new `TestSolidity_ExtendsSourcedOnTheContract` fails on the parent commit
  once per contract in the fixture, `Vault EXTENDS Ownable: FromID = "src/Vault.sol", want ""`
  and the same line for `Registry`.
- `go test ./... -count=1`: 225 packages ok, 5 fail. The failures are nine named
  tests across `cmd/grafel`, `internal/cli`, `internal/daemon`,
[...]

Issue #6295, opened 2026-08-12
solidity: the EXTENDS edge is anchored to the file component, so a contract has no outbound EXTENDS at all


arthurgeron:
I index Solidity with grafel and read the emitted graph from a downstream consumer. A consumer holding a contract entity cannot reach that contract's bases, because the contract has no outbound EXTENDS edge at all.

`internal/extractors/solidity/extractor.go:250-255` sets the EXTENDS source to the file path:

```go
for _, parent := range extends {
    edge := types.RelationshipRecord{
        FromID: filePath,
        ToID:   parent,
        Kind:   "EXTENDS",
    }
```

`filePath` is the extractor's file argument. Because it is a non-empty non-hex value, the embedded-relationship pass rewrites it (`internal/resolve/refs.go:5986-5989`), and `FileEntity` names the file entity with that same path (`internal/extractor/extractor.go:326-333`, `Subtype: "file"`), so the rewrite lands on the file component. The edge is anchored, just never to the contract. `internal/extractors/sresolver/scoped.go:472-474` does the same on the incremental path.

Running the in-process indexer over a three-file fixture, one contract per file:

```
EXTENDS From=8b9d688b7ff0b474 (SCOPE.Component subtype=file  src/implementation/Vault.sol)
             To=b98788ba1c0c1ad7
contract node d73dd19562bfbdc0 (SCOPE.Component subtype=contract Vault): no outbound EXTENDS
```

The package doc names the other endpoint: ``//   - `contract Foo is Bar, Baz` → EXTENDS edges (on the contract component)`` (`extractor.go:10`). CONTAINS on the same record omits `FromID` and so resolves to the contract (`extractor.go:311-314`).

Where one `.sol` file declares more than one contract, the file anchor also merges their base lists into one node's outbound edges. Verified on a two-contract fixture: `contract A is Base` and `contract B is Other` both emerge with the file entity as the source.

IMPORTS in the same extractor passes `filePath` too (`extractor.go:451`), and `refs.go:3658-3669` documents that as the deliberate cross-language convention for import edges (#120). An import statement belongs to a file, so that one is consistent. An `is` clause belongs to a type.

[...]

Issue #120, opened 2026-05-09
Java cross-file receiver binding (follow-up to #105)


cajasmota:
## Context

Follow-up to issue #105 (split out from PR #119, fix (B)). PR #119 landed a conservative Java-gated bare-name allowlist that classifies JDK stdlib exception classes plus high-frequency, low-collision-risk Spring Data JPA / BindingResult / Model / Pageable helpers as external-known. That is the bare-name *fallback* — fix (B).

This issue tracks the structural fix (A): Java cross-file receiver binding. After extraction strips the receiver from a method call, the resolver loses the type information and cannot match the bare method name to a real entity declared in another file. The fallback covers the ~10 % of bare names that are framework helpers; the remaining ~90 % are user-defined methods on user classes that need true cross-file receiver type tracking.

## Evidence

VERIFY-2 on java/spring-petclinic after PR #119:

```
bug-rate          : 60.09 %  (target ≤ 35 %)
external-known    :   165 → 285 (+70 from fallback)
bug-extractor     :  2549 still bug
  bare-other      :   ~728 / 798 still unresolved (~91 %)
  dotted-lower-head: 538 IMPORTS still unresolved
```

Top still-unresolved bare names from spring-petclinic are user-defined method calls on Spring/JPA receivers that the language gate alone cannot safely classify:

```
28  perform        (MockMvc.perform — but `perform` is a generic verb)
28  andExpect      (ResultActions.andExpect — generic)
26  status / view  (MockMvcResultMatchers — collide with user methods)
20  assertThat     (AssertJ — could go in fallback but high collision)
18  isOk           (StatusResultMatchers — generic boolean)
17  model          (extremely common variable/field name)
16  post / param   (HTTP verbs — collide with HttpMethod constants)
14  getName        (every entity has one — DEFERRED here)
12  setName / getId / setId  (DEFERRED here)
11  getPet         (USER method — only fix (A) can resolve)
10  save           (already in fallback under lang=\"java\")
…
```

The pattern: many of these are real *cross-file* user-method calls (`pet.getOwner()`, `owner.getPet()`, `clinicService.findOwnerById()`) that the extractor cannot resolve without tracking receiver types across files.

## Required work

1. **Receiver-type tracking** — during Java extraction, record the declared type of each call's receiver (local variable type from the surrounding method body, field type from the class declaration, or import-resolved class for static calls).
[...]

Issue #6361, opened 2026-08-19
three coverage claims today enumerated a subset of what they claimed — propose a derive-don't-list rule in AGENTS.md


cajasmota:
Sibling to #6347. That issue covers *fixtures vs real trees*; this one covers a different failure that showed up **three separate times today**, each caught only by an adversarial review, each in code specifically written to guarantee coverage.

## The pattern

A test or mechanism asserts a property over "everything of kind X", and the enumeration of X is narrower than X. The test passes. The claim reads as total. The gap is invisible because *the thing that would reveal it is the thing that was skipped*.

Three instances, all today, all in coverage machinery:

| # | claimed | actually enumerated | found by |
|---|---|---|---|
| **#6345** | "`safeExtract` is the one chokepoint — no extractor can forget a flag it is never shown" | Pass 1 only. Passes 2.5 and 3 emit entities on both production paths and never touch it | review |
| **#6349** | an AST proof that "`tryIncremental` has exactly one call site in the entire module" | only `*ast.FuncDecl` bodies were walked; a call site in a package-level `var` initialiser passed with exit 0 | review |
| **#6357** | a phantom invariant "over every emitted structural edge" | filtered on a `scope:` prefix; the `IMPORTS` family — which is exactly where the bare-name edges live — was skipped by a `continue` | review |

A fourth, adjacent: **#6355**'s `SupportedExtension` documented "an extractor will produce entities" while computing "the router names a language" — true for 1 of 67 languages, false for the other 8.

## Why it keeps happening here

grafel's correctness claims are mostly *coverage* claims — every entity from a generated file is flagged, every return path emits a trace, every edge resolves, every language is reported. Coverage claims are the ones where a partial enumeration is indistinguishable from a total one at the point of writing, because the author enumerates what they can think of and the test confirms exactly that set.

Note what did **not** catch any of these: green suites, `go vet`, code review by reading, or the author's own mutation testing. In #6345 and #6349 the author ran mutants — but generated them from the same mental model as the enumeration, so the mutants landed inside the covered subset. In #6349 **I** verified a killing mutant myself and it was a `func` declaration, the shape the walk already handled.

That is the transferable lesson: **a mutant drawn from the same understanding that produced the enumeration cannot test the enumeration.**

## Proposed rule for `AGENTS.md` → `## Evidence`

[...]

PR #6365, opened 2026-08-19
fix(#6298): verilog misanchored EXTENDS onto the file node and astro dangled its edges entirely

### LINKED ISSUE #6298: verilog and astro anchor type relationships to the file, the same shape @arthurgeron fixed for Solidity in #6297
#6295 (fixed by @arthurgeron in #6297, `45b1e2013`) was the Solidity instance of a shape that appears in at least two more extractors. The contributor named verilog themselves and deliberately kept it out of their diff; the astro case came out of the review sweep afterwards.

## The shape

An extractor emits a non-IMPORTS relationship with `FromID` set to the source file's path. That value is non-empty and non-hex, so `ReferencesEmbeddedWithAllowlist` (`internal/resolve/refs.go:5986-5989`) rewrites it. Where the package also emits a `FileEntity` (`internal/extractor/extractor.go:326-333`), the rewrite lands on the **file** component, so the edge leaves the file rather than the type that declares it — and several types in one file merge their edges onto that single node.

The fix is to leave `FromID` empty: both assembly paths substitute the owning record's own entity id, unconditionally — `cmd/grafel/index.go:6063-6069` and `internal/extractors/incremental.go:2215-2219` (`relRecordToGraphRel`).

`IMPORTS` edges are **not** part of this. `refs.go:3658-3669` documents file-anchored `FromID` as the deliberate cross-language convention for imports (#120). An import statement belongs to a file; a type relationship belongs to a type.

## 1. verilog — same shape, confirmed

`internal/extractors/verilog/extractor.go:411-415` appends `FromID: filePath` to a `SCOPE.Component` class record whose own `SourceFile` is that same path, in a package that emits a `FileEntity` at `:244`. Byte-for-byte the Solidity defect. Reported by @arthurgeron in #6297 and independently confirmed.

## 2. astro — probably worse, and unproven

`internal/extractors/astro/extractor.go:503` (`RENDERS`) and `:527` (`IMPLEMENTS`) pass `FromID: filePath`, but the owning record (`entities[0]`) is named for the component (e.g. `Header`), not the file path — **and astro emits no `FileEntity` at all**.

So the path resolves to neither the component's own id nor a file entity. Where Solidity and verilog *misanchor* the edge, astro may **dangle** it. That is a different and probably worse failure: a misanchored edge is wrong but present and traceable; a dangling one may vanish or point at nothing.

**This is a lead, not a finding.** Nobody has written a failing test for it. Establish the actual behaviour first — does the edge survive resolution, and if so where does it land? — before assuming it matches the pattern. It is possible astro's assembly compensates somewhere the sweep did not look.

## Ruled out — do not "fix" these

- `internal/extractors/vhdl/extractor.go:1026` (`USES`, file→tool): the owning record is genuinely a file-scoped component. File anchoring is correct here.
- `internal/extractors/graphql/graphql.go:157` (`CONTAINS`): the owning record is an explicit synthetic file-level container (`Name: filePath`), so this is a benign self-reference.
- Every `IMPORTS` edge across all extractors: intentional, per #120.

## Coverage gap worth noting

There is no `.sol` file anywhere in the repository and no golden or quality fixture covering Solidity, verilog or astro — which is why none of this was caught by the benchmark. Whoever takes this should consider whether a fixture is warranted, or state plainly that the extractor-seam test is the whole guard.

Refs #6295, #6297, #120, #6094, #6098

cajasmota:
The sweep @arthurgeron invited when they fixed the Solidity instance in #6297 and named verilog themselves without touching it. Their reasoning is the reference implementation here.

Closes #6298. Refs #6295, #6297, #120.

## verilog — confirmed, and the ids prove it

`internal/extractors/verilog/extractor.go` class `EXTENDS` no longer sets `FromID: filePath`. Measured on two classes in one `.sv` file, id-stamped and run through `ResolveImports` → `ReferencesEmbedded`:

```
before   base_test EXTENDS uvm_test      FromID = "e0cbdeeb860dcd8e"
         reg_test  EXTENDS uvm_reg_test  FromID = "e0cbdeeb860dcd8e"    ← one id, twice
after    both FromID = ""  → assembly stamps each class's own id (8f9227f6484340e6, …)
```

That shared id is the package's own `FileEntity`. Both classes' base lists were merging onto the file node.

## astro — measured, and it is worse than the issue predicted

The issue called astro a *lead*, not a finding, and asked for the behaviour to be established before assuming it matched. It does not match: astro **dangles** rather than misanchors.

```
owner="Header" RENDERS    FROM=<UNRESOLVED:src/components/Header.astro>  TO=Nav@src/components/Nav.astro
owner="Header" IMPLEMENTS FROM=<UNRESOLVED:src/components/Header.astro>  TO=Nav@src/components/Nav.astro
```

`UNRESOLVED` is literal — no entity carries that id. Astro emits no `FileEntity` at all, and both assembly paths substitute the owner id **only** when `FromID == ""`, so the raw path reaches the graph verbatim. ToIDs bound fine; the FromIDs did not. A misanchored edge is wrong but traceable; a dangling one points at nothing.

**Real-tree count** on this repo's own `site/` — 11 `.astro` files, 13 records: **10 of 10 `RENDERS`+`IMPLEMENTS` dangling before, 0 of 10 after.** No verilog real tree exists locally (0 `.sv`/`.v` here or in the corpora), so verilog's evidence is the fixture plus the measured id collision, stated as such in the test doc.

## The guard is the deliverable, and it found 25 sites rather than 3

`TestNoNewFileAnchoredTypeRelationships` parses every non-test `.go` under `internal/extractors` and flags any composite literal setting `FromID` to a file-path expression with a non-`IMPORTS` `Kind`. Allow-list keyed by `package:Kind` — line-drift-proof — with a stated reason per entry. It also fails on a **stale** allow-list entry that no longer matches, and on a scan matching nothing at all.

Mutants, all three fire:

```
verilog restored → verilog/extractor.go:425  key "verilog:EXTENDS"
astro restored   → astro/extractor.go:514,538  keys "astro:RENDERS", "astro:IMPLEMENTS"
nim (a fourth extractor, never touched) given a synthetic EXTENDS
                 → nim/nim.go:344  key "nim:EXTENDS"
```

The last one is the point: a scan that only knows the extractors already fixed is decorative.

Ten of the 25 sites are **correct** — the owning record genuinely is file-scoped (vhdl, graphql `CONTAINS`/`FEDERATES`, proto, hcl ×2, bicep, dockerfile, yaml/helm, verilog's tool `USES`). Each is allow-listed with its reason, verified by reading the owning record at the site rather than by pattern.

## svelte has astro's exact defect, and is deliberately not fixed here

No `extractor.FileEntity` anywhere in the package; `entities[0]` is the component named from the basename; `RENDERS`/`USES`/`NAVIGATES_TO`/`CONTAINS` all pass `file.Path`. Measured: `Card.svelte` rendering `<Button />` yields `FROM=<UNRESOLVED:src/lib/Card.svelte>`.

Only `RENDERS` was measured — the other three are the same construction on the same record and are marked as **inference, not finding**, in the allow-list. Eleven sites across a language this issue did not review deserve their own change and their own measurement, so they are allow-listed as `KNOWN OFFENDER (#6298)` and filed separately.

## Fixture decision, stated plainly

None added. `testdata/fixtures` has 27 language dirs with no solidity, verilog or astro entry, and there are **0 `.sol` files in the tree** — the issue's coverage-gap claim holds. For verilog the extractor-seam plus resolve-through test **is** the whole guard. For astro the checked-in `site/` tree is a genuine real-tree corpus, so a synthetic fixture would add nothing.

## Not touched, on purpose

`IMPORTS` edges in every extractor — file-anchored `FromID` is the documented cross-language convention (#120); an import belongs to a file, a type relationship belongs to a type. Also `vhdl`'s file→tool `USES` and `graphql`'s synthetic file-container `CONTAINS`, both ruled out in the issue and both re-verified at the site.

### DIFF
--- a/internal/extractors/astro/extractor.go
+++ b/internal/extractors/astro/extractor.go
@@ -188,7 +188,7 @@ func (e *Extractor) Extract(ctx context.Context, file extractor.FileInput) ([]ty
 	// ── 3. Template: child components (RENDERS) and islands (IMPLEMENTS) ─────
 	body := extractBody(src)
 	if body != "" {
-		renderRels, islandRels := extractTemplateRelationships(body, file.Path, componentName)
+		renderRels, islandRels := extractTemplateRelationships(body, componentName)
 		entities[0].Relationships = append(entities[0].Relationships, renderRels...)
 		entities[0].Relationships = append(entities[0].Relationships, islandRels...)
 
@@ -488,7 +488,19 @@ func extractServerBuildEntities(fm string, fmStartLine int, filePath, subtype st
 // extractTemplateRelationships scans the HTML body for:
 //   - PascalCase component tags → RENDERS edges (deduplicated)
 //   - client:* directives on those tags → IMPLEMENTS edges
-func extractTemplateRelationships(body, filePath, componentName string) (renders []types.RelationshipRecord, islands []types.RelationshipRecord) {
+//
+// Both kinds leave FromID EMPTY so the assembly loop stamps the owning record's
+// own entity id — entities[0], the .astro component. Passing the file path is
+// strictly worse here than in the Solidity/verilog form of this defect (#6295,
+// #6297): this package emits NO FileEntity, so the path matched no entity at
+// all. Measured on three .astro files through the production resolver
+// (ResolveImports → ReferencesEmbedded): every RENDERS/IMPLEMENTS edge reached
+// the graph with FromID = "src/components/Header.astro" verbatim — a DANGLING
+// endpoint, not a misanchored one. See TestAstro_TemplateRelsAnchoredOnComponent.
+//
+// IMPORTS (extractImports) still passes the file path: that is the documented
+// cross-language convention for import edges (#120).
+func extractTemplateRelationships(body, componentName string) (renders []types.RelationshipRecord, islands []types.RelationshipRecord) {
 	seenRenders := make(map[string]struct{})
 	seenIslands := make(map[string]struct{})
 
@@ -500,9 +512,8 @@ func extractTemplateRelationships(body, filePath, componentName string) (renders
 		if _, exists := seenRenders[name]; !exists {
 			seenRenders[name] = struct{}{}
 			renders = append(renders, types.RelationshipRecord{
-				FromID: filePath,
-				ToID:   name,
-				Kind:   "RENDERS",
+				ToID: name,
+				Kind: "RENDERS",
 				Properties: types.Props{
 					{K: "from_component", V: componentName},
 					{K: "line", V: fmt.Sprintf("%d", lineIdx+1)},
@@ -524,9 +535,8 @@ func extractTemplateRelationships(body, filePath, componentName string) (renders
 				seenIslands[islandKey] = struct{}{}
 				directive := islandAttrRE.FindString(tagSrc)
 				islands = append(islands, types.RelationshipRecord{
-					FromID: filePath,
-					ToID:   name,
-					Kind:   "IMPLEMENTS",
+					ToID: name,
+					Kind: "IMPLEMENTS",
 					Properties: types.Props{
 						{K: "framework_island", V: name},
 						{K: "host_component", V: componentName},
--- a/internal/extractors/astro/issue6298_anchoring_test.go
+++ b/internal/extractors/astro/issue6298_anchoring_test.go
@@ -1,0 +1,123 @@
+package astro
+
+import (
+	"maps"
+	"slices"
+	"testing"
+
+	"github.com/cajasmota/grafel/internal/graph"
+	"github.com/cajasmota/grafel/internal/resolve"
+	"github.com/cajasmota/grafel/internal/types"
+)
+
+// issue6298_anchoring_test.go — RENDERS and IMPLEMENTS must be anchored on the
+// .astro component that owns them, not on the file path.
+//
+// WHAT WAS MEASURED BEFORE THE FIX. The three .astro files below, extracted,
+// id-stamped and pushed through the production resolver pipeline
+// (ResolveImports → ReferencesEmbedded), produced:
+//
+//	owner="Header" kind=RENDERS    FROM=<UNRESOLVED:src/components/Header.astro> TO=Nav @src/components/Nav.astro
+//	owner="Header" kind=RENDERS    FROM=<UNRESOLVED:src/components/Header.astro> TO=Logo @src/components/Logo.astro
+//	owner="Header" kind=IMPLEMENTS FROM=<UNRESOLVED:src/components/Header.astro> TO=Nav @src/components/Nav.astro
+//
+// "UNRESOLVED" is literal: no entity in the set carries that ID. This package
+// emits no FileEntity, so unlike the Solidity (#6295) and verilog forms of the
+// same defect — where the rewrite at least landed on a real file component —
+// the path survived rewriting verbatim and reached graph assembly as a DANGLING
+// FromID. Both assembly paths (cmd/grafel/index.go and relRecordToGraphRel in
+// internal/extractors/incremental.go) substitute the owning record's own entity
+// id only when FromID is EMPTY, so a non-empty path is passed straight through.
+//
+// After the fix the same run gives FROM=<empty> on all three, which assembly
+// stamps with Header's id.
+//
+// REAL-TREE COUNT (AGENTS.md § Evidence). This repository's own Astro site,
+// site/ — 11 .astro files, 13 entity records — extracted and resolved the same
+// way:
+//
+//	before: RENDERS+IMPLEMENTS = 10, FromID empty = 0,  DANGLING = 10
+//	after:  RENDERS+IMPLEMENTS = 10, FromID empty = 10, DANGLING =  0
+//
+// Every one of the ten template edges on a real Astro tree was dangling. The
+// measurement harness was a throwaway; site/ is checked in, so it is
+// reproducible by walking it through astroExtract + BuildIndex/ReferencesEmbedded.
+// No equivalent verilog tree exists locally (0 .sv / .v files in this repo and
+// in ~/Projects/archigraph-corpora), so verilog's evidence is the two-class
+// fixture in the sibling test file and its measured pre-fix id collision.
+func TestAstro_TemplateRelsAnchoredOnComponent(t *testing.T) {
+	files := map[string]string{
+		"src/components/Header.astro": "---\nimport Nav from './Nav.astro';\n---\n<div><Nav client:load /><Logo /></div>\n",
+		"src/components/Nav.astro":    "---\n---\n<nav>hi</nav>\n",
+		"src/components/Logo.astro":   "---\n---\n<span>logo</span>\n",
+	}
+	var recs []types.EntityRecord
+	for _, p := range slices.Sorted(maps.Keys(files)) {
+		recs = append(recs, astroExtract(t, p, files[p])...)
+	}
+	for i := range recs {
+		if recs[i].Name == "" {
+			continue
+		}
+		recs[i].ID = graph.EntityID("issue6298", recs[i].Kind, recs[i].Name, recs[i].SourceFile)
+	}
+
+	// The premise of the dangle: nothing in this package's output is named for
+	// the file, so the file path can never resolve to an entity here. If astro
+	// ever starts emitting a FileEntity this assertion is the thing that should
+	// be revisited — but the fix below is correct either way.
+	for i := range recs {
+		if recs[i].Name == "src/components/Header.astro" {
+			t.Errorf("astro now emits a file-named entity (%s/%s); "+
+				"re-read the dangle rationale in this file's doc comment",
+				recs[i].Kind, recs[i].Subtype)
+		}
+	}
+
+	resolve.ResolveImports(recs, resolve.BuildImportTable(recs))
+	resolve.ReferencesEmbedded(recs, resolve.BuildIndex(recs))
+
+	var header *types.EntityRecord
+	for i := range recs {
+		if recs[i].Name == "Header" && recs[i].Kind == "SCOPE.Component" {
+			header = &recs[i]
+		}
+	}
+	if header == nil {
+		t.Fatal("no Header component entity")
+	}
+
+	byID := make(map[string]*types.EntityRecord, len(recs))
+	for i := range recs {
+		byID[recs[i].ID] = &recs[i]
+	}
+
+	var renders, implements []string
+	for _, r := range header.Relationships {
+		switch r.Kind {
+		case "RENDERS", "IMPLEMENTS":
+		default:
+			continue // IMPORTS keeps the file path on purpose (#120).
+		}
+		if r.FromID != "" {
+			t.Errorf("%s → %s: FromID = %q, want \"\" so assembly stamps Header's id",
+				r.Kind, r.ToID, r.FromID)
+		}
+		target := "<dangling:" + r.ToID + ">"
+		if e := byID[r.ToID]; e != nil {
+			target = e.Name + "@" + e.SourceFile
+		}
+		if r.Kind == "RENDERS" {
+			renders = append(renders, target)
+		} else {
+			implements = append(implements, target)
+		}
+	}
+	slices.Sort(renders)
+	if got, want := slices.Compact(renders), []string{"Logo@src/components/Logo.astro", "Nav@src/components/Nav.astro"}; !slices.Equal(got, want) {
+		t.Errorf("RENDERS targets = %v, want %v", got, want)
+	}
+	if got, want := implements, []string{"Nav@src/components/Nav.astro"}; !slices.Equal(got, want) {
+		t.Errorf("IMPLEMENTS targets = %v, want %v", got, want)
+	}
+}
--- a/internal/extractors/file_anchored_rels_guard_test.go
+++ b/internal/extractors/file_anchored_rels_guard_test.go
@@ -78,12 +78,20 @@
 //	   a recognised path spelling. `filepath.Join(dir, base)` is not seen.
 //	F. concatenation, BOUNDED — `filePath + ""`, `filePath + ".x"`: the path must
 //	   come FIRST and every appended operand must be a string literal. A literal
-//	   PREFIX (`"scope:ormmodel:" + filePath + "#" + name` — ormlink, haskell,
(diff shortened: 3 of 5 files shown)


cajasmota:
## Independent adversarial review: REQUEST-CHANGES — the guard is itself the pattern it was written to catch

The two code fixes are correct and non-vacuous. **The allow-list is where this PR is wrong**, and the scan's completeness claim is wrong in a way that already exists in-tree.

### Confirmed sound

`entities[0]` is unambiguously the right owner for astro — appended first, unconditionally, with the rels attached to it explicitly, so index arithmetic cannot matter. Extra frontmatter exports are appended after; nested components live in their own files. The svelte deferral is honest: all 10 attach points are `entities[0].Relationships = append(...)`, and svelte emits no `FileEntity`, so the three unmeasured kinds genuinely are the same construction on the same record. No hidden second defect. All three vacuity guards fire, including on a *narrowing* of the matcher.

### The allow-list blesses six real defects

The stated criterion is "the owning record is genuinely file-scoped". The **operative** criterion is stricter: a path-valued `FromID` only lands if the package emits a node carrying **that exact path** as `Name`/`QualifiedName`, or an `extractor.FileEntity`. `internal/resolve/refs.go` has no path→entity index, and `looksLikeSourceFilePath` (`refs.go:2729`) only routes to `DispositionDynamic` — it **suppresses the bug metric without resolving anything**. That is exactly why astro's dangle survived until now, and the allow-list inherited the same blind spot.

| entry | verdict |
|---|---|
| `vhdl:USES`, `verilog:USES` | OK — but the owning record is the **tool** component, not a file record. The stated reason is factually wrong, and the PR's new comment repeats it |
| `graphql:CONTAINS`, `yaml:OVERRIDES` | OK and functional |
| `graphql:FEDERATES` | **misanchored** — the site's own comment says the edge comes "from the extending stub" |
| `proto:CONTAINS` ×3 | **dangles** — two lines below, the sibling service→rpc edge correctly leaves `FromID` empty |
| `hcl:CONTAINS` ×2 | **dangles** for every non-root `.tf`; resolves only by accident for a root `main.tf` |
| `hcl:DEPENDS_ON` | **dangles**, and the wrong owner |
| `bicep:DEPENDS_ON` | **dangles unconditionally**, wrong owner |
| `dockerfile:USES` | **dangles** for any non-root Dockerfile |

Aggravating: `.bicep` and `Dockerfile` are absent from `sourceFileExtensions` (`refs.go:2752-2767`), so those two are classified `DispositionBugExtractor` — live bug-extractor counts, blessed by a test.

They are being re-labelled `KNOWN OFFENDER (#6298)` rather than fixed here. Fixing six languages in this PR would repeat the mistake #6298 exists to record: astro was *assumed* to match verilog's pattern and turned out to be a different, worse failure. Each needs its own measurement.

### The key granularity blesses new offenders in the same package

Mutant: append a `FromID: filePath` `USES` edge to verilog's **class** record — next to the `EXTENDS` edge this PR fixes. The guard **passes**. The scanner sees the site (`verilog/extractor.go:429 key=verilog:USES`) and discards it, because the key collides with the allow-listed *tool* edge.

So the newest instance of the exact bug, in the exact package being fixed, is silently blessed. Six of the 22 current sites already share the single key `svelte:USES`.

### A form the scan cannot see, live in-tree

`internal/extractors/yaml/helm.go:875` does `rel.FromID = file.Path` **after** construction, for kinds `INCLUDES` and `BINDS` — under keys that are not on the allow-list and never will be reported. It happens to resolve via the yaml `SCOPE.Document` anchor, so it is not itself a bug, but it proves the scan's shape assumption is already violated.

Nine probes were built; the scan caught **one**, and that one only incidentally. Invisible: post-construction assignment (live above), an unlisted local-variable name, a struct field, a helper call, `filepath.Join`, concatenation, a **positional** composite literal (no `KeyValueExpr` at all), and a loop whose `FromID` and `Kind` are set in different literals. Root causes: `filePathExprs` is a 10-entry closed spelling list, and the scan requires `FromID` and `Kind` in the *same* literal.

The header comment does disclose "WHAT IT CANNOT SEE", which is the right instinct — but it understates the list and does not mention that one of those forms exists today.

### What this says about the approach

I asked for this scan because a one-line fix in two extractors is not the valuable part — catching the *next* extractor is. That reasoning holds. But the scan as written makes a totality claim that is true only for the literal form the three known offenders happened to use, and its allow-list converts six unexamined defects into blessed behaviour. That is #6361's pattern applied to the guard rather than to the code, which is the more expensive place for it, because a defect blessed by a test is harder to find than one merely unguarded.

Fixes in progress: honest re-labelling, granularity that distinguishes sites within a package, the post-construction form, the two cheap structural blind spots, and a disclosure that matches reality.

cajasmota:
## Independent re-review: REQUEST-CHANGES — the disclosure vouches for a live instance of the defect

Everything this round introduced held up except one claim, and that claim is the problem.

### Verified, independently, by reading the emitting site *and* the node emission

Last round six of ten stated reasons were wrong, so none of the corrected ones were taken on trust. **No corrected reason was wrong this round.**

- `vhdl:buildVHDLToolEntities:USES` and `verilog:buildToolEntities:USES` — owner is the **tool** record; resolves because the package unconditionally prepends `extractor.FileEntity`, whose `Name` **is** `file.Path`. Binds via `byName` (`refs.go:1591`).
- `graphql:extractGraphQL:CONTAINS` and `markdown:Extract:CONTAINS` — self-references onto their own path-named owner (`Name: filePath` / `QualifiedName: docQName`).
- The yaml/helm group — `buildYAMLDocument` is prepended whenever entities exist, with `QualifiedName: file.Path`, so the anchor is guaranteed present at all four sites.

### The site-count mechanism survives hard attack

Removal → stale-entry failure. Over-count → `expects 2 site(s), found 1`. Function swap → both the orphaned old key and the unaccounted new key fail. It is recomputed from source and compared every run, in both directions, so unlike the hand-maintained numbers this cycle has been deleting, **it cannot silently rot.**

### BLOCKER — `swift` is misclassified, and the misclassification conceals a live defect

Form F lists five schemes excluded as *"a STRUCTURAL REF … a different, deliberate mechanism, not this bug"*, all characterised as the literal-prefix spelling `"scope:ormmodel:" + filePath + "#" + name`.

**swift is not that shape.** `swift/package.go:189` and `:242` are `FromID: filePath + "::" + d.name` — path **first**, then a literal, then a non-literal ident, invisible to the matcher for a different reason (`isFilePathExpr` requires every trailing operand to be a `BasicLit`). And it is not a structural ref:

- the owning record (`swift/package.go:164-177`) sets `Name: d.name` and **no `QualifiedName` at all** — the only two `::` occurrences in the file are these two `FromID`s;
- nothing in `internal/resolve` knows the `swift_package` `::` spelling;
- both edges hang off `rec.Relationships` — the target component itself — so `FromID` should be empty.

Dangling **and** misowned: the astro failure mode, in a package the disclosure explicitly vouches for as *not* an instance. A probe of `filePath + "::" + name` confirms the guard cannot see it.

The other four (ormlink, abibridge, haskell, markdown-prefix) genuinely take the prefix form and are correctly characterised.

This is a stricter failure than an incomplete guard. An unguarded defect is merely unfound; one that a comment certifies as fine is actively defended. Same escalation as last round's allow-list, one layer further out: **code → allow-list → prose about the allow-list.**

### Four disclosure gaps, on the block's own standard

The block's standard is that an incomplete disclosure is worse than a narrow scan, because a stated boundary is read as *the* boundary.

1. **The scan root is `internal/extractors/` only** — never stated. Producers outside it are invisible: `internal/patterns/openapi_extractor.go` (7 sites), `internal/engine/{http_endpoint_synthesis,spring_routes,django_routes,feature_flag_edges}.go`, `internal/resolve/bazel_overlay.go`. **None is raw-path-valued today, so there is no live miss** — but "a new extractor is covered the day it lands" reads as completeness, and a new producer in `internal/patterns` would not be.
2. **Form G's shape test is a substring match on the type's spelling** — `relationshipShaped` requires `exprString(t)` to contain `"relationship"`, so a positional literal of a local `type myRel struct{FromID, ToID, Kind string}` is invisible. Keyed literals are unaffected.
3. **A split-literal `IMPORTS` edge loses its exemption and is wrongly flagged** — a false-positive tax on the documented #120 convention, currently unnamed.
4. **The count pins cardinality, not identity** — replacing a blessed form-A site with a *different* offending construction in the same function left the count at 1 and the test **green**. And a `?` key with `count ≥ 2` would bless any pair of split literals; latent today, since the only `?` entry has count 1.

### Also

`testdata/` is not excluded from the walk — `internal/extractors/testdata/incrfixture/main.go` and `golang/testdata/issue4426/constants.go` are parsed as extractor source. Harmless now; a fixture containing a `FromID:` line would report under a `golang/testdata/issue4426` key, and a deliberately-malformed Go fixture would fail the guard for an unrelated reason.

Swift is being allow-listed as `KNOWN OFFENDER (#6298)` and folded into #6367 rather than fixed here — each language needs its own measurement, and astro is the standing proof that assuming a shared shape is how this goes wrong. **The wrong claim is the blocker; the missing fix is not.**

cajasmota:
## Round 3 verified — the blocker is fixed, and I ran the mutants myself

`ebcab3633`. Only the guard test changed; `swift/package.go` is byte-identical to `main` (`2c581d58…` both sides).

**The swift claim, checked independently of the fix agent.** Both sites are `FromID: filePath + "::" + d.name`; the owning record sets `Name: d.name` with no `QualifiedName`; those are the only two `"::"` occurrences in the package; `internal/resolve` has no `swiftpm`/`swift_package` scheme. Dangling **and** misowned — the astro mode, in a package the previous disclosure certified as *not* an instance.

**Mutants I ran (not the agent's table):**

| mutant | result |
|---|---|
| baseline, both guards | green, and `allowedFileAnchored` contains **zero** swift keys — so the main scan genuinely cannot see the shape, rather than blessing it |
| delete the swift entry | **FAIL** — unaccounted key, both sites named with file:line and form |
| fix swift's code (`FromID: ""` ×2) | **FAIL** — stale entry, with the accurate diagnosis leading the vacuity guard |
| **mine, drawn from outside the agent's model:** give *dockerfile* the hidden shape (`file.Path + "#" + stageRef`) | **FAIL** — and instructively, by **both** scans: the main scan reports its allow-list entry as stale, the new scan reports the new site as unaccounted |

That last one is the claim the PR actually exists to make. Fixing swift and astro is worth little; catching the *next* extractor that acquires the shape is the whole point, and a language that has never had this bug acquires it and is caught, under a key that names the function.

Every mutant was an in-place edit restored by `cp` from a shasum-verified backup — never `git checkout`. `./internal/extractors/...` 74 packages ok, EXIT=0; `gofmt -l` clean.

**The four disclosure gaps** — scan root, form G's substring shape test, the split-literal `IMPORTS` false positive, and cardinality-not-identity — are now stated in the block, each with the probe that shows it.

### What three rounds of this PR taught, which outlives the PR

1. the code was wrong (verilog misanchored, astro dangled);
2. the **allow-list** was wrong — six entries blessed real defects;
3. the **prose about the allow-list** was wrong — it certified a live defect as a deliberate mechanism.

Each layer was added to guard the one below it, and each was wrong in the same direction: over-confident about coverage. An unguarded defect is merely unfound; one a comment certifies as fine is actively defended. The fix is not more layers — it is that every claim of totality needs a mutant drawn from **outside the author's model of the space**, which is why the dockerfile probe above is the one I insisted on running myself.

swift joins #6367. Merging.

cajasmota:
## Round 4 — the fix agent corrected my figures, and one of them was the same error I was correcting

`a7da53c26`. Diff is **comment-only** — every changed line in the 157/41 is a `//` line, verified mechanically — so round 3's mutant results carry over unchanged.

### What I got wrong

I handed the agent measurements to work from. It re-derived them instead of trusting me:

| I said | measured |
|---|---|
| `internal/engine/` ~80 files with `FromID:` | **66** |
| `internal/custom/` ~40 | **31** |
| 25 `.FromID =` assignments outside the tree | **16** — 9 of the 25 are *inside* it |
| `internal/links/` is a producer | contributes **zero** |
| "exactly four" path-token FromIDs outside the tree | **twelve** |

That last one is the one worth naming. I missed two direct hits (`cmd/grafel/index.go:4688`, `internal/membench/fixture.go:154`) and six one-hop indirections — `internal/engine/plugin_system_edges.go:96` (`fromID := "File:" + filePath`) and five `internal/custom/*typegraph.go` sites. **Five of the six sit in `internal/custom/`** — the tree whose omission I had just flagged as the defect in claim 1.

So I reproduced the error inside the correction, while naming it. I confirmed all five corrections myself before merging.

My conclusion survives — none of the twelve is raw-path-valued (`"File:" + path`, `"Function:" + caller`, hashed `EntityID`s), so **there is no live miss**. But the warrant I offered was as under-searched as the one I was criticising, and the disclosure now records the enumeration *method* and its known incompleteness rather than presenting a list as exhaustive.

### markdown: not pinned, and it proved why

The agent tried adding a `markdown:Extract:CONTAINS` pin and showed it fails instantly as stale. I reproduced that:

```
knownInvisibleOffenders["markdown:Extract:CONTAINS"] matches no site any more —
either the code was fixed (delete the entry) or the matcher was narrowed
```

The value never appears *at* a FromID: it is stored on the heading records and read back as `parentQName` (:243) and `headingEntities[parentIdx].QualifiedName` (:290). Both matchers inspect the FromID expression only, and `QualifiedName` is not in `filePathFields` — so form F's bound is not what hides it. A pin there would have been a comment wearing a test's coat. It is documented as a third category instead, with its apparent benignness marked INFERRED, not measured.

The agent also declined to repeat my `cross/*` prefix-scheme list on my authority, resting its refutation of "exactly four" only on the 12 `Build*StructuralRef` sites it measured itself. That is the right instinct and I want it on the record.

### Verified before merge

Baseline green; markdown-pin mutant fails as above; comment-only diff confirmed mechanically; `./internal/extractors/...` 74 packages ok EXIT=0; `gofmt -l` clean. Restores by `cp` from a shasum-verified backup (`ff97d97…`), never `git checkout`.

### Six instances on one PR

code → allow-list → prose about the allow-list → the corrected prose (swift → markdown, same mis-attribution one item to the left) → the scan-root claim → **my own correction of it**.

The sixth is mine, and it is the useful one: it rules out "the author wasn't careful." I was hunting this exact pattern, in a PR about this pattern, and still shipped an under-searched enumeration as a fact. That is the argument for #6361 being a written rule with a mechanical check rather than a review reflex — reflexes are what just failed.

Issue #6369, opened 2026-08-19
resolve: BuildIndex should not index import placeholders by name — the marker exists in a dozen extractors and the resolver ignores it


cajasmota:
Split out of #6296 (reported by @arthurgeron). This is the half that actually kills the reported false bind. **It is gated on a measurement — see the falsifier at the bottom. Do not land it on reasoning.**

Depends on #6368 (which stamps the marker this issue teaches the resolver to honour).

## The change

`BuildIndex` skips entities with `Subtype == "import"` when populating `byName`, `byPackageComponent`, and the `byLocationKind[...].real` tier.

## Why the marker alone is not enough

`grep -c Subtype internal/resolve/refs.go` → **0**. The resolver reads no Subtype anywhere today. Stamping `Subtype:"import"` on solidity's placeholders and re-running leaves the false bind exactly as it was — measured:

```
M1 (variant 1 + Subtype=import): EXTENDS to=18b169aa -> RESOLVED[name=IVault file=src/implementation/Vault.sol]   unchanged
M2 (variant 2 + Subtype=import): EXTENDS to="IVault" -> UNRESOLVED-STUB                                            unchanged
```

The marker already exists in a dozen-plus extractors and buys nothing at resolution time. That is the whole reason #6296's remedy (b) is two issues.

## Measured effect

With placeholders withheld from `BuildIndex`, all three of @arthurgeron's variants go correct:

```
R1 (base in another directory):  EXTENDS to=fb9fc886 -> RESOLVED[name=IVault file=src/interface/IVault.sol]  rewritten=3 ambiguous=0 unmatched=0
R2 (same directory):             EXTENDS to=9f9b4110 -> RESOLVED[name=IVault file=src/IVault.sol]            rewritten=3 ambiguous=0 unmatched=0
R3 (two importers):              both Vault and Vault2 -> RESOLVED[file=src/interface/IVault.sol]            rewritten=5 ambiguous=0 unmatched=0
```

Before: variant 1 binds to the placeholder in the *deriving* file and reports `ambiguous=0` — a **confident** wrong answer, and a dead end for every hierarchy walk (`internal/mcp/mro.go:109,172` BFSes EXTENDS and terminates on a node with no outgoing edges). Variants 2 and 3 leave the bare string `"IVault"`.

## Why this is a class, not a language

[...]

Issue #6156, opened 2026-08-05
Full rebuild orphans the IMPORTS edge for every third-party import — the SCOPE.External node it should bind to exists in the same graph


cajasmota:
The FULL rebuild leaves the `IMPORTS` edge for a **third-party** import pointing at the hex id of an import placeholder that `PruneImportPlaceholders` has already removed — a dangling endpoint — even though the `SCOPE.External` entity that edge should bind to **exists in the same graph**. The incremental path (`extractors.TryIncremental`) binds the same edge correctly, so this is a full-path defect surfaced by full-vs-incremental parity, with the incremental answer being the right one.

## Measured

Fixture: the #6129 content-parity corpus with `import falcon` in the delta file (added by the #6150 fix). Both graphs read back from disk via `LoadGraphFromDir`.

```
FULL ENT ext:falcon  SCOPE.External|falcon|  sub=package
FULL REL 5b13f231a92947d1 -> 42dcd97a6da50584   (toLive=false)   <- cphandler_delta.py -IMPORTS-> ???

INC  ENT ext:falcon  SCOPE.External|falcon|  sub=package
INC  REL 5b13f231a92947d1 -> ext:falcon        (toLive=true)
```

`42dcd97a6da50584` is the pruned `SCOPE.Component subtype="import"` placeholder for `falcon`. The indexer's own log for the full run:

```
import-placeholder-prune: considered=3 pruned=3 rels_hoisted=1 rels_orphaned=2 kept=0 edge_toid_rewrites=0 ref_repoints=2
```

`rels_orphaned=2` is this.

## Why the #6131 repoint does not cover it

The #6131 block in `internal/resolve/imports.go` re-points an incoming edge at "the entity the pipeline had already resolved the same (importer file, source_module) import to". For a **first-party** import that target exists at prune time, which is why `import cperrs_static` is fine. For a **third-party** import there is nothing resolved yet: the `SCOPE.External` node is created later, by `external.Synthesize`, which runs *after* resolution and does not adopt endpoints left dangling by the prune. So the repoint has no candidate and the edge is orphaned.

## Blast radius

Every third-party import in every repo, on the default full-rebuild path — this is not a fixture artefact. The graph loses the file→package dependency edge that `grafel_expand` / lib-boundary / module aggregation read, and gains a dangling endpoint.

[...]

Issue #742, opened 2026-05-19
[m1][js/ts] Placeholder analog of #693 — JS/TS SCOPE.Component import-stub pruning


cajasmota:
## Goal
Apply the placeholder-pruning fix from Python (#693 / PR #715) and Java (#681 / PR #694) to JS/TS. Fixture re-measurement found 288 orphan SCOPE.Component entities on fixture-b, 103 on fixture-c, 31 on fixture-e — exact same shape as the Python placeholder issue.

Expected impact:
- fixture-b: 5.2% orphan rate → ~1-2% (extrapolated from Python's 19.8% → 1.8% on django-realworld)
- fixture-c: 3.8% → ~1%
- fixture-e: 10.1% → ~3%

## Pattern (per #693 / #715)

JS/TS `import { X } from "lib"` currently emits a SCOPE.Component placeholder for `lib` (or `X`). The placeholder dangles because real usage code references the `ext:lib` external entity, not the placeholder. Fix: lift IMPORTS relationships onto the per-file SCOPE.Component/file entity and don't emit the wrapper.

See PR #715 for the Python reference implementation.

## Acceptance
- fixture-b SCOPE.Component orphans drop to near-zero
- fixture-b orphan rate drops ≥-3pp
- fixture-c and fixture-e similar improvements
- Quality fixtures preserved
- Cross-language bug-rate parity for non-JS/TS corpora

## Beyond the minimum
Verify on other JS/TS corpora in the test set (express, nextjs-commerce, nestjs-starter, etc.). Same pattern likely present.

## Related
- #693 / #715 (Python analog — landed)
- #681 / #694 (Java analog — landed)

Issue #693, opened 2026-05-19
Python placeholder pruning — SCOPE.Component/module import leftovers dominate post-#689 orphans (analog of #681)


cajasmota:
## Context

Step 0 verification for #679 (cross-file class-field resolution) on top of #689 (`feat/django-field-orphan-recovery`). Built that branch, indexed three Python corpora under an isolated daemon root (`/tmp/arch-679`), and computed the residual orphan structure.

**Result: #679's hypothesized residual (cross-file `self.<attr>` references) is NOT the dominant orphan class. The dominant residual is `SCOPE.Component/module` import placeholders — the Python analog of #681 (Java placeholder pruning).**

## Measurement (post-#689 binary)

| Corpus           | Entities | Orphans | Orphan rate | `SCOPE.Component/module` | Share of orphans |
|------------------|---------:|--------:|------------:|-------------------------:|-----------------:|
| django-realworld |      532 |      91 |      17.10% |                       85 |          **93.4%** |
| flask-realworld  |      799 |     124 |      15.51% |                       98 |          **79.0%** |
| pandas           |   13,089 |   5,374 |      41.05% |                    4,995 |          **92.9%** |

(client-fixture-a was not available in the local corpora set — same as PR #689 noted.)

## Concrete examples (10 sampled from django-realworld)

All have `start_line=0`, `end_line=0`, `signature=null` — i.e. import placeholders, not in-source entities:

- `rest_framework.status`   (conduit/apps/articles/views.py)
- `django.db.models.deletion`   (conduit/apps/profiles/migrations/0001_initial.py)
- `rest_framework.serializers`   (conduit/apps/articles/serializers.py)
- `os`   (conduit/settings.py)
- `django.conf.urls.url`   (conduit/apps/profiles/urls.py)
- `rest_framework.views.APIView`   (conduit/apps/profiles/views.py)
- `.models.Profile`   (conduit/apps/profiles/serializers.py)   ← relative import
- `django.conf.urls.url`   (conduit/urls.py)
- `rest_framework.permissions.AllowAny`   (conduit/apps/authentication/views.py)
- `rest_framework.serializers`   (conduit/apps/articles/relations.py)

These are external (`rest_framework`, `django`, `os`) or relative-import (`.models.Profile`) `SCOPE.Component/module` entities emitted by the Python extractor's import handling that are never bound to anything because the target lives outside the corpus (external) or because the import-target resolution misses the file inside the corpus (relative import case).

## Why #679 is not the right fix

[...]

PR #6371, opened 2026-08-19
fix(#6368): solidity import placeholders collided by EntityID with the declarations they shadow

### LINKED ISSUE #6296: solidity: an import placeholder shares the base's name, so an EXTENDS target resolves to the placeholder or to nothing
I index a group of Solidity repositories with grafel and read the emitted graph from a downstream consumer. Where a base contract is reached through an import, the EXTENDS target endpoint does not reach the base's declaration, even when that declaration is indexed in the same group. This is about the target endpoint only, and is independent of which entity the edge starts at.

`buildImportEntities` emits a `SCOPE.Component` per distinct import path, named after the path basename minus `.sol`, sourced to the importing file (`internal/extractors/solidity/extractor.go:442-449`, name at 430-434, dedup at 423-426). Solidity convention is one declaration per file named after the file, so that placeholder is normally name-identical to the base it stands for.

The EXTENDS target is the bare base name (`extractor.go:253`), resolved by name. So for any base reached through an import, two entities carry that name: the real declaration, and the placeholder in the deriving file.

Measured with the in-process indexer, a base declared in the group and imported by one file in another directory:

```
LookupStatusHint("IVault", "EXTENDS") = "" status=2   (statusAmbiguous)
EXTENDS ToID resolved to b98788ba1c0c1ad7
  = SCOPE.Component  subtype=""  IVault  src/implementation/Vault.sol
```

The collision deletes the `byName` entry and marks the name ambiguous (`internal/resolve/refs.go:1583-1589`), so the name tier declines. `rewriteOneWithCaller`'s locality tiers then leave exactly one that can fire for EXTENDS: the `"EXTENDS", "IMPLEMENTS"` arm reads `byPackageComponent[callerPkgDir]` (`refs.go:3377-3382`). The caller's own directory is where the placeholder lives, so that tier resolves the edge to it. The same-file tier cannot help: it reads the `.real` bucket and skips `SCOPE.*` (`refs.go:3327-3329`), and a `.sol` file yields nothing but `SCOPE.*`. The trailing component fallback (`refs.go:3397-3407`) and the by-member probe (`refs.go:3272-3277`) are both gated on CALLS.

Two variants of the same collision, also measured:

- Declaration and deriving file in one directory: `byPackageComponent[dir]` is blanked by the collision sentinel (`refs.go:1552`), so the target stays the bare string `"IVault"`.
- Two or more files importing the base: two placeholders plus the declaration, no tier recovers, target stays a bare string.

Where the base needs no import, because it is declared in the same file, no placeholder exists and the edge resolves to the declaration. That is the shape that works.

For example, given `src/implementation/Vault.sol`:

```solidity
// SPDX-License-Identifier: MIT
pragma solidity ^0.8.20;

import {IVault} from "../interface/IVault.sol";

contract Vault is IVault {
    function total() external view returns (uint256) { return 0; }
}
```

and `src/interface/IVault.sol` declaring `interface IVault`, the extractor emits, alongside the `Vault` contract:

```
entity name="IVault" kind="SCOPE.Component" subtype="" file="src/implementation/Vault.sol" line=4
```

and after resolution the EXTENDS edge points at that entity rather than at the interface.

I am not sending a fix. Whether the placeholders should stop being `SCOPE.Component`, carry a marker `BuildIndex` honours, or be resolved through the import path before any bare-name lookup, is a decision about the resolution model rather than a local correction, and each option changes what IMPORTS edges point at.

### LINKED ISSUE #6368: solidity: import placeholders collide by EntityID with the declarations they shadow, because dedup keys on path and naming keys on basename
Split out of #6296 (reported by @arthurgeron). This is the **cheap half** and it stands alone — it does not fix the reported EXTENDS false bind, which is #6296's b2.

## The shape

`internal/extractors/solidity/extractor.go:441-456` emits a `SCOPE.Component` per distinct import path, named after the path basename minus `.sol` (`:428-433`), sourced to the *importing* file, deduped by path (`:421-425`).

Two keys disagree: the dedup key is the **path**, the entity name is the **basename**. `graph.EntityID` hashes `repo|kind|name|sourceFile` and nothing else. So:

```
variant 5 — a file declaring `interface IERC20` that also imports another IERC20.sol:
  id=539aa0e5  SCOPE.Component  IERC20  src/IERC20.sol  lines=3-3   <- placeholder (import line)
  id=539aa0e5  SCOPE.Component  IERC20  src/IERC20.sol  lines=5-7   <- the real interface

variant 6 — ./a/Token.sol and ./b/Token.sol imported by one file:
  id=c207d145  SCOPE.Component  Token   src/Main.sol    lines=3-3
  id=c207d145  SCOPE.Component  Token   src/Main.sol    lines=4-4
```

Two records, one ID, placeholder emitted first. Whichever write wins decides whether the graph's `IERC20` node spans the real declaration or is a one-line stub on an `import` statement — so `grafel_get_source` can hand back an import line for a real contract. **Silent.** `internal/resolve/imports.go:3354-3370` already documents this hazard class for Go's empty-Subtype import entities; solidity is a second instance.

## The fix

Stamp **both**:

- `Subtype: "import"` — routes the placeholder into `PruneImportPlaceholders`, which deletes both records and moots the collision.
- `Properties["module"] = importPath` — **not optional.** `imports.go:2887` and `:3384` read the module path from `r.Name` or `r.Properties["module"]`, and solidity puts the path only on the *relationship's* `source_module`. Omitting it silently disables the #642 relative-import rewrite and the #6156 `ext:` restore.

ID-neutral (Subtype is not in the hash). No test asserts the empty Subtype. Aligns with the existing convention in a dozen-plus extractors.

## What it fixes beyond the collisions

- orphan mis-bucketing (`internal/quality/audit/heuristics.go:30`)
- `internal/mcp/mro.go:559` — `isDefinitionEntity` returns true for a subtype-less `SCOPE.Component`, so MRO can select an import placeholder **as the class definition** (`effective_contract_tool.go:352`). Note `:565` `isClassEntity` keys on Kind alone and would still say yes — worth a look.
- `internal/dashboard/shape_tree.go:694-698` — `findClassEntityByName` explicitly accepts `Subtype == ""`, so a placeholder can be returned as the class shape ahead of the real declaration.
- `internal/enrichment/candidates.go:640-686` — placeholders are eligible LLM-enrichment candidates (only names ≤3 chars are rejected), so import basenames are sent to the docgen LLM to be described as if they were code.

And via the `module` property, two pre-existing defects:

- `internal/graph/cycles.go:70-77` requires both endpoints in `known`; solidity's IMPORTS `ToID` is a raw path with no entity, so **every solidity IMPORTS edge is dropped from cycle detection** today — solidity contributes zero import cycles.
- `internal/quality/audit/audit.go:508-527` `classifyImportToID` buckets any ToID starting `./`, `../`, `/` as a path-format bug, so **every solidity import scores as an import-hygiene bug**, depressing the score at `handlers_quality.go:361-370` and the fidelity metric at `docgen_repair_tools.go:151-166`.

## Watch out

`internal/substrate/solidity.go:27-90` is an **independent second implementation** of solidity import parsing, computing the same "last segment minus `.sol`" local name at `:61-66`. A change to `buildImportEntities` not mirrored there drifts the substrate's binding table away from the graph. Worth its own follow-up either way.

## Evidence bar

Per `AGENTS.md` `## Evidence`: report the variant-5 and variant-6 ID collisions as measured before/after, not inferred. A probe harness exists (extractor → `EntityID` → `BuildImportTable` → `ResolveImports` → `BuildIndex` → `ReferencesEmbedded`, in-process).

Refs #6296, #642, #6156, #693, #742, #681

cajasmota:
Fixes #6368. Derived from @arthurgeron's report in #6296.

**This PR was rewritten after review. An earlier revision stamped `Subtype:"import"` + `Properties["module"]`; that approach was measured, found to regress the daemon's default path, and abandoned. See the round-1 review below.**

## The defect

Solidity emitted a `SCOPE.Component` per import, named after the path basename, deduped by *path*. `graph.EntityID` hashes `repo|kind|name|sourceFile`, so those keys disagree and the placeholder collides with any same-named record in the same file — the declaration it shadows, or a second import of the same basename.

## Why the marker approach was wrong

`resolve.PruneImportPlaceholders` has exactly **one** non-test caller: `cmd/grafel/index.go:5904`, the CLI full-rebuild path. The daemon's default reindex is `internal/extractors.TryIncremental`, and no placeholder prune runs there. Records fold through `foldDuplicateEntity`, which gap-fills but never overrides — and since imports are appended before contracts, the placeholder is always the survivor.

So the placeholder's **empty** Subtype was being gap-filled from the declaration it shadowed. Stamping `Subtype:"import"` blocked that accidental repair:

```
base:       IERC20 subtype="interface" lines=3-3   (wrong span, right kind)
marker cut: IERC20 subtype="import"    lines=3-3   (wrong span, wrong kind)
```

A marker only helps where something checks it.

## What ships: the #742 pattern

**No placeholder entity is emitted.** The IMPORTS relationship hangs on the per-file `subtype="file"` carrier the extractor already emits at `extractor.go:124`. Established in-repo three times — JS/TS #742, Java #681/#694, Python #693/#715 — and path-independent: no marker, no prune dependency, no fold interaction. Both collisions vanish because the colliding record is never created.

#742's six invariants were verified for solidity rather than assumed; independently re-confirmed in review, including that `BuildImportTable` reads IMPORTS from **any** record's Relationships keyed on `rel.FromID` falling back to `r.SourceFile` (`imports.go:215-227`).

## Measured, both indexing paths

**Variant 5** — `src/IERC20.sol` declares `interface IERC20` and imports another `IERC20.sol`:

```
Path A (daemon default, convertExtractedRecords):
  base:       id=97d3753f subtype="interface" lines=3-3   folded away = 1
  marker cut: id=97d3753f subtype="import"    lines=3-3   folded away = 1
  AFTER:      id=97d3753f subtype="interface" lines=5-7   folded away = 0, collisions 0

Path B (CLI full rebuild, post-prune): 7 -> 6 entities, COLLISIONS: none
```

**Variant 6** — `./a/Token.sol` + `./b/Token.sol` in one file: no `Token` entity at all; both IMPORTS edges on the `src/Main.sol` carrier.

## It also fixes #6296's headline defect on Path B

Not claimed in the first draft, and it should be. Removing the placeholder clears the `refs.go:1583` ambiguity sentinel, so the name tier binds to the real declaration:

```
base 529594550:   EXTENDS Vault -> b98788ba1c0c1ad7  (the placeholder, subtype="" lines=3-3)
branch 90ed0e9c0: EXTENDS Vault -> a00845d3fc42ac3f  (interface IVault, src/interface/IVault.sol)
```

`b98788ba1c0c1ad7` is the literal entity ID @arthurgeron reported. Measured on Path B, on his exact fixture. The general resolution question — the locality tier landing on whatever placeholder is in scope — remains #6369.

## Honest limitation

#642's relative-import rewrite and #6156's `ext:` restore are both gated on a live `Subtype=="import"` record. Solidity's placeholder carried `Subtype: ""` on base, so **neither ever fired for solidity** — removing the record cannot regress what never ran. Solidity IMPORTS ToIDs remain raw specifiers, so cycle detection still drops them and import hygiene still flags them: **identical to base, neither improved nor regressed.** An earlier draft of this body claimed metric gains (cycles 0→1, hygiene 3/3→1/3); those came from the abandoned mechanism and are **not** in this diff. Recovering them is #6369.

An extractor-side attempt to resolve specifiers was implemented and **reverted**: it worked on variants 5 and 6 but mis-bound a third fixture, landing an IMPORTS edge on a function (`SCOPE.Operation B.pong`). A confident wrong bind is the failure class this issue exists to remove.

## Tests

Two round-1 tests were deleted with the mechanism they pinned (neither existed on `main`). Net additive: 5 Path B tests + 2 Path A tests, up from 4. `line_attribution_test.go` moved its assertions from the placeholder's StartLine onto the edge's `line` property (the convention at ocaml:509, zig:376, nim:450, svelte:708) — net gain, not loss.

The Path A gate **fails on base** with a self-explaining message, so the regression this PR was rewritten to fix cannot silently return.

## Verification

Mutants I ran myself: re-emit the placeholder → dies on both paths; **host the rels on the last entity while leaving `FromID` correct** → dies naming the *host*, so the test pins the carrier rather than a string. Independent review added a third (partial fix: placeholder re-emitted alongside the new edges) — kills 5 tests plus both Path A gates.

`gofmt` clean, `go vet` clean, 89 packages green. `tools/coverage validate` exits 0; regenerated docs are byte-identical to the committed ones.

**Quality ratchet:** cannot move, and that is measured rather than assumed — `find internal/quality/golden -name '*.sol'` → 0, `grep -rli solidity internal/quality/golden/` → 0.


cajasmota:
## Independent review: REQUEST-CHANGES — the marker regresses a real declaration on the daemon's default path

The implementation is clean and the tests are non-vacuous. **The approach #6368 prescribed is wrong**, which is my error in writing the issue, not the implementer's.

### The defect

`resolve.PruneImportPlaceholders` has exactly **one** non-test caller: `cmd/grafel/index.go:5904` — the CLI full-rebuild path (Path B).

The daemon's primary reindex is a different pipeline: `internal/extractors.TryIncremental` (`incremental.go:393`, called from `cmd/grafel/daemon.go:1234`), and incremental is **on by default** (`internal/extractor/extractor.go:39`, #5231). `cmd/grafel/index.go:6299-6301` states the split outright. **No import-placeholder prune runs on that path at all.**

There, records fold by EntityID through `foldDuplicateEntity` (`incremental.go:2058-2070`), gap-fill-never-override:

```go
if surv.Subtype == "" && dup.Subtype != "" { surv.Subtype = dup.Subtype }
```

Solidity appends import placeholders (`extractor.go:130`) **before** contracts (`:139`), so the placeholder is always the survivor. I verified both halves myself.

**Before** this PR the placeholder's empty Subtype was gap-filled from the shadowed declaration, so the node read `subtype="interface"` — wrong line span, right kind. **After** it, `Subtype:"import"` blocks the gap-fill and a real `interface IERC20` reads as an import on every daemon reindex:

```
BEFORE (529594550): survivor IERC20 subtype="interface" lines=3-3 id=97d3753f
AFTER  (b14b86985): survivor IERC20 subtype="import"    lines=3-3 id=97d3753f
```

Downstream: `dashboard/shape_tree.go:694-698` accepted that node and rejects `"import"`; `quality/audit/heuristics.go:30` now buckets a real declaration as `CauseImportPlaceholder`. It also opens a full-vs-incremental entity-set divergence — #6131's class.

This is the `PlaceholderKept > 0` case the PR body calls hypothetical. On Path A it is **100% of daemon reindexes**.

### The over-claim that follows

"The collision is removed, not resolved" and "the survivor is the declaration" are true and robustly true on Path B — `canMigrate` and the file-carrier gate are both structurally satisfied for solidity, so `PlaceholderKept` is 0 there. Both sentences are unqualified in the body, and both are false on Path A.

### Redirecting to the #742 pattern

Not stamping a marker that one indexing path checks. Instead: **do not emit the placeholder at all**, and hang the IMPORTS relationship on the per-file `subtype="file"` carrier solidity already emits at `extractor.go:124`.

That is the established in-repo answer for this exact shape, three times over — JS/TS #742, Java #681/#694, Python #693/#715 — and `internal/extractors/javascript/prune_import_placeholders.go:1-55` documents its six correctness invariants. It is path-independent: no marker, no prune dependency, no fold interaction, and both collisions vanish because the colliding record is never created.

### What I attacked that held

- No gate drops or alters the placeholder between emission and the prune on Path B. `foldFileComponentDuplicates` (the #6138 hazard, a real risk here since the placeholder's name matches the file stem in variant 5) runs at `index.go:5948` — **after** the prune. Order verified, not assumed.
- **No `Properties["module"]` key collision.** `EnsureModule` and the `singleModuleLabel` overwrite sit at `index.go:6005-6011`, 101 lines after the prune, and all three readers run before it. Real ambiguity, currently unreachable.
- The redundant `"kind"`/`"subtype"` Properties keys **are** the convention — `internal/extractor/extractor.go:326-340` sets exactly those two. Non-finding.
- Tests are non-vacuous. Both mutants re-run independently, matching the body exactly. The only wrong implementations that slip past are cosmetic.
- All three unfixed-consumer claims are accurate (`mro.go:549`, `:565`, and `enrichment/candidates.go` with 0 `Subtype` matches).
- The golden-set claim reproduces: 0 `.sol` files, 0 solidity mentions. The ratchet argument stands.

### One correction to the body

`@openzeppelin/contracts/...` has no leading `./` or `/`, so `classifyImportToID` buckets it as `ImportFormatOther`, not `PathString`. The residual is real; the stated bucket was wrong.

### Follow-up, not for this PR

Searched: `Subtype:\s*"import"`, `Subtype\s*=\s*"import"`, a named-constant form (0 hits), `"subtype": "import"` in Properties/Metadata maps, and `.Subtype =` post-hoc mutation, across all non-test `*.go`. Found 16 sites. Of those, four have `Name` that is **not** the module path and no `Properties["module"]`, so `imports.go`'s reader gets a short segment: **razor** `:248`, **vue** `:1361`, **markdown** `:479`, **graphql** `:402`. `cpp` mirrors the subtype into `Metadata` rather than `Properties` (`:1305`, `:1357`), so it reads as bare to any `Properties` consumer. The other ten put the full path in `Name` and are correct. Filing separately.

Issue #6374, opened 2026-08-19
parent: three unrelated causes leave http_endpoint_definition entities without an IMPLEMENTS edge (the 24.71% headline is not reproducible)


cajasmota:
Split out of #6313 (@manuel1358000). That issue's headline — 99.6% of `http_endpoint_definition` orphaned — turned out to be a metric artifact (folded into #6346). **This is the real defect the metric was drowning.**

## Measured

10 corpus repos through the real pipeline (extract → resolve → engine passes), 344 `http_endpoint_definition` entities:

```
truly isolated (no semantic edge either way):   0 / 344   (0.00%)
no IMPLEMENTS edge at all (handler unbound):   85 / 344  (24.71%)
```

Per repo, where it concentrates:

| repo | definitions | no IMPLEMENTS |
|---|---|---|
| django | 249 | **62** |
| express | 47 | **19** |
| flask | 21 | 4 |
| everything else (7 repos) | 27 | 0 |

The endpoint exists, it is contained by its file, and nothing binds it to the function that serves it. So "which handler serves `POST /api/orders`" has no answer for a quarter of endpoints in the two most common Python/JS server frameworks.

## Why it was invisible

`internal/feedback/report.go:139` counts orphans by **outgoing** edges only, and `http_endpoint_definition` is a sink by construction — the handler points into it. So these 85 were indistinguishable from the 259 that are correctly bound: all 344 looked orphaned at the reporter's scale. Fixing the metric (#6346) makes this visible; it does not fix it.

## Where to start

`internal/engine/http_endpoint_resolve.go` — `bridgeEndpointToHandler` (`:939-951`) is what emits `handler --IMPLEMENTS--> definition`. The question is which endpoints reach it and which do not, and why the miss rate is ~25% in django and ~40% in express while it is 0% in the seven smaller repos.

[...]

Issue #6313, opened 2026-08-14
http_endpoint_definition orphaned (99.6%) despite recognized HTTP frameworks


manuel1358000:
# grafel feedback report

Generated: 2026-08-14T16:59:52Z
grafel version: v0.2.2 (commit 6813ff5, built 2026-08-14T04:15:15Z)
Group profile: 15 language(s) (typescript, yaml, css, javascript, markdown, json, shell, astro, python, groovy, dockerfile, html, sql, env, jsonschema), 348000-349000 entities, 1264000-1265000 relationships
Confidence: 67% (31/46 sanity checks passed)

### Entities by language

| Language | Entity count (range) |
|---|---|
| astro | 100+ |
| css | 100+ |
| dockerfile | 21-100 |
| env | 21-100 |
| groovy | 100+ |
| html | 21-100 |
| javascript | 100+ |
| json | 100+ |
| jsonschema | 21-100 |
| markdown | 100+ |
| python | 100+ |
| shell | 100+ |
| sql | 21-100 |
| typescript | 100+ |
| yaml | 100+ |

### Entity kind distribution

| Kind | Language | Count (range) |
|---|---|---|
| SCOPE.Component | astro | 21-100 |
| SCOPE.Operation | astro
| SCOPE.Pattern | astro | 21-100 |
| SCOPE.Stylesheet | css
| Config | dockerfile | 21-100 |
| Dependency | dockerfile
| SCOPE.Component | dockerfile | 6-20 |
| SCOPE.Config | dockerfi
| SCOPE.Config | env | 21-100 |
| SCOPE.Component | groov
| SCOPE.Operation | groovy | 100+ |
| SCOPE.Component | html
| Config | javascript | 100+ |
| SCOPE.Component | javas
| SCOPE.Config | javascript | 100+ |
| SCOPE.Operation | javas
| SCOPE.Pattern | javascript | 21-100 |
| SCOPE.Plugin | javascri
| SCOPE.Schema | javascript | 100+ |
| SCOPE.Config | json | 1
| SCOPE.Schema | jsonschema | 21-100 |
| SCOPE.CodeBlock | markd
| SCOPE.Document | markdown | 100+ |
| SCOPE.Component | pytho
| SCOPE.Operation | python | 100+ |
| SCOPE.Pattern | python
| SCOPE.Component | shell | 100+ |
| SCOPE.Operation | shell
| SCOPE.Schema | sql | 21-100 |
| Config | typescript | 1
[...]

PR #6375, opened 2026-08-19
fix(#6346): the orphan gate could not fire below 100% and the metric counted only outgoing edges, so sinks were miscounted as orphans

### LINKED ISSUE #6346: the feedback orphan gate is inverted: it cannot fire below 100% (so #6313's 99.6% passed) and fires on terminal-by-design kinds
`runSanityChecks` (`internal/feedback/sanity.go:33-47`) has an orphan check that is wrong in **both** directions. It cannot fail on the defect it exists to catch, and it fails loudly on kinds that are correct by design.

```go
// 2. Orphan rate < 100% for all kinds with N >= 10.
for kind, ks := range r.OrphanByKind {
    if ks.Total < 10 { continue }
    passed := ks.OrphanPct < 100.0
```

## Direction 1 — the threshold is exactly 100%, so it can never fire on a real regression

A kind is only reported as failing when **every single entity** of that kind is orphaned. One resolved edge anywhere in the repo — including an accidental one — flips the kind to passing and it stays passing forever.

This is not hypothetical, and #6313 is the demonstration. @manuel1358000 reported `http_endpoint_definition` at a **99.6% orphan rate** on a codebase with recognized HTTP frameworks. That is as close to totally broken as a kind can get. `99.6 < 100.0`, so the gate says **PASS**, the confidence score is unaffected, and nothing in the report tells the user anything is wrong. The reporter found it by reading the orphan table themselves.

A gate whose failure condition is "literally 100%" is measuring the wrong thing: 100% orphan usually means a kind is *terminal by design* (direction 2), whereas 99.6% means a resolver is *broken*. The check currently flags the healthy case and passes the broken one — it is close to inverted.

## Direction 2 — terminal-by-design kinds fail it, correctly, forever

Some entity kinds are leaves: nothing in the model ever points at them, so 100% orphan is their correct steady state. On a real repo this check emitted **15 failures**, none of which was a defect. That is noise deducted straight from the confidence score, and it trains the reader to ignore the whole section — which is exactly what happened here, because #6313's genuine 99.6% would have been invisible among them even if it *had* fired.

So the two directions compound: the false failures make the true failures unreadable.

## What a correct check needs

1. **A threshold that can fire.** Some meaningful bound (e.g. 90%, or a per-kind expected range) rather than the degenerate 100%.
2. **A notion of which kinds are terminal by design**, so leaf kinds are excluded from the check rather than failing it every run. This is the part that needs a decision — an explicit list, a structural derivation from the schema (does any edge type target this kind at all?), or a per-kind expected-orphan band.
3. **`N >= 10` re-examined.** A kind with 9 entities and 9 orphans is currently invisible regardless.

My inclination is (2) derived structurally rather than listed: "is this kind the target of any edge type we can emit?" is answerable from the schema and does not rot the way a hand-maintained list does — and a hand-maintained list of exclusions is exactly the shape of dead config #6330 just deleted 31 files of.

Wants a decision on the threshold and on how terminal kinds are identified before anyone writes code. Filing it separately from #6313 because #6313 is a resolver defect in one kind, and this is the reason nobody was told about it.

Refs #6313, #6330

cajasmota:
Refs #6346 (direction 1 — see "What this does not fix" below). Also resolves the mechanism behind #6313 (@manuel1358000's 99.6% report).

`internal/feedback/` only — no engine or resolve changes.

## The defects

**The gate could not fire.** `sanity.go` was an inline `passed := ks.OrphanPct < 100.0`. A 99.6% orphan rate passed. Only a literally-perfect failure tripped it.

**The metric counted one direction.** Orphan was defined by *outgoing* edges only, but many kinds are sinks by construction — `http_endpoint_definition` is pointed *into* by its handler. So a correctly-wired endpoint counted as orphan, and the rate degraded with group size: the only outgoing kind a definition sources is `ENTRY_POINT_OF`, gated behind `MaxProcesses: 300`. Small repos fit inside the budget; a 348k-entity group exhausts it and the rate goes to ~100%. It was also framework-unfair — per `internal/graph/coverage.go:590-604` the handler↔definition link is emitted in both directions depending on framework, so a Spring group scored 0% and a Django group ~100% **for identical graph quality**.

## The fix

Orphan now means **no semantic edge in either direction** (CONTAINS/DECLARES still excluded, both ways). Terminality is **derived from observed semantic participation per kind** rather than a hardcoded list. A second check, `kind-carries-semantic-edges[kind]`, gates the 100% end — without it, a totally broken kind was routed to the terminal bucket, which nothing gated, and reported confidence 100%.

Deriving rather than listing was a requirement: a name list is the #6361 failure class. No edge→kind schema exists in `internal/types` to derive from, so participation is derived from the graph itself.

## Measured — 12 corpora, independently reproduced by review

| repo | confidence | sanity failures |
|---|---|---|
| nextjs-commerce | 75→100% | 3→0 |
| nestjs | 77→93% | 3→1 |
| express | 62→79% | 8→5 |
| django | 66→74% | 12→10 |
| flask | 63→79% | 9→6 |
| flask-realworld | 85→86% | 2→2 |
| spring-petclinic | 89→94% | 2→1 |
| laravel-routing | 82→91% | 2→1 |
| **total** | improves in 11 of 12 | **50→32** |

Like-for-like on the orphan-rate check alone: **47 → 15**; the new participation check accounts for 14 firings.

**An earlier revision of this body claimed 50→18.** That was measured with the 100%-end hole open and is withdrawn. The figures above were re-measured after it was closed and independently re-derived by a reviewer who re-indexed all 12 corpora.

**The gate now fires strictly below 100% on real data for the first time** — django `Route`, `Module`, `Test`; spring-petclinic `Config` 90.9%.

**The strongest single finding:** flask has **172 `Route` entities and not one semantic edge among them** — `@app.route` paths never bound to the function immediately below them (verified by sampling, e.g. `conduit/user/views.py:30`). Under the previous revision that kind was silently reclassified as terminal and reported healthy.

## What this does NOT fix — #6346 direction 2 remains open

All 14 `kind-carries-semantic-edges` failures are kinds that **already failed at base** under the old `== 100%` rule. Same repos, same kinds, new check name. Triaged: ~8 are unarguably terminal by design and will fail forever (markdown code fences, CSS selectors, HTML form fields, `requirements.txt`/`pyproject.toml`), 2 are unambiguous true positives (`Route` ×2), ~4 arguable.

**True-positive rate ≈ 14%.** That is exactly the "trains the reader to skip the section" failure mode direction 2 names. This PR therefore does **not** close #6346; a follow-up issue tracks distinguishing terminal-by-design from a dead resolver without reintroducing a name list.

## The #6374 signal stays visible

After the fix the `http_endpoint_definition` orphan count is the #6374 unbound-handler count — express 19, django 62, flask 4, 85 of 369 — reproduced independently row for row. The identity is **empirical, not structural**: it holds because the process-participating definitions happen to be a subset of the `IMPLEMENTS`-bound ones.

## Corrections to the issues

1. **#6313's "every one of the 344 is attached" holds only if `CONTAINS` counts as attachment** — the metric excludes it. Under semantic-only edges 85/369 (23.0%) have no semantic edge in either direction. Corrected on that issue.
2. **A "no edges at all" definition would have been a dead metric** — *isolated* measures 0.0% across the corpora, reproducing the structurally-impossible-arm failure #6346 itself complains about.
3. **#6346 direction 2 over-attributed the noise to terminal-by-design kinds.** Most false failures were *sinks* with inbound semantic edges. An earlier revision of this body said "only 4 kinds corpus-wide have genuinely zero participation"; measured, it is **10 kinds across 14 kind×repo firings**, and the omitted set includes `Route`. That claim is withdrawn.

Also: #6346's point 3 (`N >= 10`) is not a gate rule — `Generate` only publishes kinds at that floor, so the gate cannot see smaller kinds regardless.

## Granularity limitation, stated

Per-kind is coarser than the two per-entity rules deleted. Of the 9 newly-failing orphan-rate kinds, **7 contain zero field-subtype entities** so the old rules could not have covered them; **2 are genuine granularity loss** — express-realworld `SCOPE.Schema` (43 of 60 unwired are field leaves) and laravel-routing (109 of 112). Both sides pinned by `TestGenerate_MixedParticipationSchemaFlipsFieldLeaves`.

## Mutants

Author's: threshold→99.9; drop the inbound arm; disable derivation; always-terminal; neuter the participation check; revert the `render.go` legend. Coordinator's, drawn from outside that set: make CONTAINS count inbound so nothing is ever an orphan (killed by 4 tests including one naming the #6374 risk); and `passed := tks.OrphanCount <= tks.Total`, an off-by-one that silently restores the blind spot (killed by 3, including *"the report would tell a maintainer their unlanded resolver is perfect"*).

A pre-existing vacuous test and a tautology introduced in round 1 were both replaced with `Generate`-driven equivalents.

`gofmt` clean, `go vet` clean, 10 packages green. Restores by `cp` from shasum-verified backups, never `git checkout`.

### DIFF
--- a/internal/feedback/render.go
+++ b/internal/feedback/render.go
@@ -96,7 +96,7 @@ func Render(w io.Writer, r *Report) error {
 
 	// Section 2 — Orphan Rate
 	fmt.Fprintf(w, "## 2. Orphan Rate\n\n")
-	fmt.Fprintf(w, "An entity is orphan when it has no semantic edge in EITHER direction (CONTAINS/DECLARES excluded, in both directions).\n\n")
+	fmt.Fprintf(w, "An entity is orphan when it has no semantic edge in EITHER direction (CONTAINS/DECLARES excluded, in both directions). The table below counts only orphans of kinds that carry a semantic edge SOMEWHERE in this group; kinds where no entity does are listed separately under **Expected/terminal orphans**, so a kind reading 0 here may still be entirely unwired there.\n\n")
 	if len(r.OrphanByKind) == 0 {
 		fmt.Fprintf(w, "_No entity kind with >= 10 entities found._\n\n")
 	} else {
--- a/internal/feedback/render.go
+++ b/internal/feedback/render.go
@@ -123,11 +123,13 @@ func Render(w io.Writer, r *Report) error {
 		}
 		fmt.Fprintf(w, "\n")
 
-		// Expected/terminal orphans: container-terminal Components and
-		// field-leaf terminals are not defects — routed here instead of the
-		// defect table above so the raw signal is not silently dropped.
+		// Expected/terminal orphans: kinds with zero observed semantic
+		// participation anywhere in the group. Routed here instead of the
+		// defect table above so the raw signal is not silently dropped, and
+		// labelled with the ambiguity rather than asserted as healthy — see
+		// the derivation comment in report.go (#6346).
 		if len(r.OrphanTerminalByKind) > 0 {
-			fmt.Fprintf(w, "**Expected/terminal orphans** (container-terminal by design, not defects):\n\n")
+			fmt.Fprintf(w, "**Expected/terminal orphans** — no entity of these kinds carries a semantic edge in either direction anywhere in the group, so they are terminal by construction as far as the graph can show. Not counted as defects. (A total resolver regression on a kind would look identical; if one of these used to be wired, treat it as a defect.)\n\n")
 			fmt.Fprintf(w, "| Kind | Total | Terminal orphan | Terminal orphan %% |\n|---|---|---|---|\n")
 			for _, kind := range sortedKindStatsKeys(r.OrphanTerminalByKind) {
 				tks := r.OrphanTerminalByKind[kind]
--- a/internal/feedback/render.go
+++ b/internal/feedback/render.go
@@ -96,7 +96,7 @@ func Render(w io.Writer, r *Report) error {
 
 	// Section 2 — Orphan Rate
 	fmt.Fprintf(w, "## 2. Orphan Rate\n\n")
-	fmt.Fprintf(w, "An entity is orphan when it has no outgoing semantic edges (CONTAINS/DECLARES excluded).\n\n")
+	fmt.Fprintf(w, "An entity is orphan when it has no semantic edge in EITHER direction (CONTAINS/DECLARES excluded, in both directions).\n\n")
 	if len(r.OrphanByKind) == 0 {
 		fmt.Fprintf(w, "_No entity kind with >= 10 entities found._\n\n")
 	} else {
@@ -129,7 +129,7 @@ func Render(w io.Writer, r *Report) error {
 		// labelled with the ambiguity rather than asserted as healthy — see
 		// the derivation comment in report.go (#6346).
 		if len(r.OrphanTerminalByKind) > 0 {
-			fmt.Fprintf(w, "**Expected/terminal orphans** — no entity of these kinds carries a semantic edge in either direction anywhere in the group, so they are terminal by construction as far as the graph can show. Not counted as defects. (A total resolver regression on a kind would look identical; if one of these used to be wired, treat it as a defect.)\n\n")
+			fmt.Fprintf(w, "**Expected/terminal orphans** — no entity of these kinds carries a semantic edge in either direction anywhere in the group, so they are terminal by construction as far as the graph can show. Excluded from the orphan defect count above, but each one raises a `kind-carries-semantic-edges` sanity check for triage: a total resolver regression looks identical from the graph alone, so if one of these used to be wired, it is a defect.\n\n")
 			fmt.Fprintf(w, "| Kind | Total | Terminal orphan | Terminal orphan %% |\n|---|---|---|---|\n")
 			for _, kind := range sortedKindStatsKeys(r.OrphanTerminalByKind) {
 				tks := r.OrphanTerminalByKind[kind]
--- a/internal/feedback/render_orphan_definition_test.go
+++ b/internal/feedback/render_orphan_definition_test.go
@@ -1,0 +1,39 @@
+package feedback
+
+import (
+	"strings"
+	"testing"
+)
+
+// TestRender_OrphanDefinitionSentenceMatchesTheMetric pins the one sentence in
+// the shipped report that tells a human what the orphan number underneath it
+// means (#6346 review C1).
+//
+// It read "no outgoing semantic edges" while the metric became direction-aware,
+// which made every shipped report's own legend false. #6313 was found by a
+// reporter reading that exact table by hand, so this string is load-bearing:
+// nothing else in the report explains the number.
+func TestRender_OrphanDefinitionSentenceMatchesTheMetric(t *testing.T) {
+	r := &Report{
+		TotalEntities: 100,
+		GeneratedAt:   mustParseTime("2026-05-27T00:00:00Z"),
+		GroupName:     "test-group",
+		OrphanByKind:  map[string]KindStats{"SCOPE.Route": {Total: 20, OrphanCount: 4, OrphanPct: 20.0}},
+		FrameworkHits: map[string]int{},
+	}
+	var sb strings.Builder
+	if err := Render(&sb, r); err != nil {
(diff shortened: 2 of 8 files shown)


cajasmota:
## Round 2 — the 100% end is gated, and the headline number was withdrawn

`f65a75929`.

### The blocker, and why it mattered

Round 1's review found that deriving terminality opened a hole at exactly 100% breakage: a zero-participation kind was routed to the terminal bucket, which nothing gated, so a **totally** broken kind reported `OrphanPct: 0.0` and confidence 100% — while base `83f3c898a` caught that same input, because `passed := ks.OrphanPct < 100.0` fires only at exactly 100%. The PR was removing the one case the old gate could catch.

The case that made it a blocker rather than a curiosity: a wholly unwired group — entities with `CONTAINS` only — reported perfect health. That is the shape of a new language extractor emitting entities before its resolver lands, and **VB.NET is on this milestone**.

### Fixed with a check, not a comment

New **`kind-carries-semantic-edges[kind]`**: fails when no entity of a published kind carries a semantic edge in either direction anywhere in the group.

Chosen over the longitudinal option because that needs a prior report and would be silent on a first run — precisely the new-extractor case. Chosen over gating a combined defect+terminal rate because that conflates two different statements into one number. It will fire on genuinely terminal kinds; the note says so and asks for triage rather than asserting a defect. Fail rather than warn, justified in the code: a false failure costs one confidence point and a line of text, a false pass ships a broken extractor with a clean bill of health. The size floor is the report's existing `N >= 10` publication floor, not a new knob.

`TestRunSanityChecks_ContainerTerminalKindNoFalseFailure` now asserts both halves, so it can no longer pin the blind spot as intended behaviour.

### The headline number was wrong and is withdrawn

**50 → 18 was measured with the hole open. Re-measured with it closed: 50 → 32.** Like-for-like on the orphan-rate check alone: 47 → 15. The new check accounts for 14 firings. Confidence still improves in 11 of 12 repos (django 66→74%, express 62→79%, flask 63→79%, nextjs-commerce 75→100%).

The standout: **flask has 172 `Route` entities and not one semantic edge among them.** Under the previous commit that kind was silently reclassified as terminal and reported as healthy. It now fails.

### C3 triage came out partly against the author

Of the 9 kinds newly failing the orphan-rate gate, **7 are not granularity loss** — django `Config`/`Module`/`Route`/`Test`, flask `Module`, flask-realworld `Module`, spring-petclinic `Config` contain zero field-subtype entities, so the deleted per-entity rules could not have covered them. **2 are exactly the granularity loss the reviewer suspected** — express-realworld `SCOPE.Schema` (43 of 60 unwired are field leaves) and laravel-routing `SCOPE.Schema` (109 of 112). Both sides are now pinned by `TestGenerate_MixedParticipationSchemaFlipsFieldLeaves`, and the limitation is stated rather than all 9 being claimed as signal.

### Claims narrowed

- The #6374 identity is **empirical, not structural** — it holds because the process-participating definitions happen to be a subset of the `IMPLEMENTS`-bound ones. "Exactly" removed.
- The two conflicting measurements counted **different sets** (all checks vs the orphan check alone); both now carry their scope. The numerator 85 is identical across all three denominators — 344, 369, 360 — only the denominator moved.
- "0.0% isolated across ~85k entities" dropped as broader than its evidence.
- `render.go:99` corrected, and a test asserts the old wording appears **nowhere** in the rendered report.

### Verified before push

My own mutant, distinct from the four in round 1 and the two here — an off-by-one a refactor would plausibly introduce, `passed := tks.OrphanCount <= tks.Total`, which always passes and silently restores the blind spot:

```
--- FAIL: TestGenerate_TotalRegressionOnOneKindFailsTheGate
    a kind that lost EVERY semantic edge PASSED every sanity check — the 100% end of the gate is blind
    confidence = 100% for a group with a totally unwired kind
--- FAIL: TestGenerate_WhollyUnwiredGroupIsNotReportedHealthy
    a group with no semantic edge anywhere reports confidence 100% — the report would tell a maintainer their unlanded resolver is perfect
```

Three tests, each naming the defect. The tautology at `sanity_gate_test.go:104` was replaced with a `Generate`-driven equivalent that produces its counts from product code, and both unguarded loops gained `found` guards.

`gofmt` clean, `go vet` clean, 10 packages green. Restores by `cp` from shasum-verified backups, never `git checkout`.

PR #6307, opened 2026-08-13
fix(#6304): nothing ever re-listed a directory whose fsnotify listing was abandoned, so its files stayed unwatched for the life of the daemon

cajasmota:
# fix(watch): nothing ever re-listed a directory whose fsnotify listing was abandoned, so its files stayed unwatched for the life of the daemon

## The defect, reproduced rather than inferred

fsnotify v1.10.1's kqueue backend mimics inotify by taking a descriptor on every
*entry* of a directory grafel `Add()`s, and by re-listing that directory on each
`NOTE_WRITE` to pick up entries that appeared since (`dirChange`). That re-listing
gives up silently. On an `f.Info()` error, and on an `EACCES`/`EPERM`/`ENOENT`
from `sendCreateIfNew`, `dirChange` `return nil`s **in the middle of the listing**;
`readEvents` then discards even the errors it *does* return. Entries after the
failure point are never watched and never reported — no `Create`, no `Errors`
entry, no log line.

The issue reported this as inference from the vendored source plus a
directory-quantized CI deficit, with 52 failed attempts to reproduce. It is now
reproduced deterministically, and the reproduction is in the suite. One entry
with mode `0000` makes `sendCreateIfNew`'s `internalWatch` fail with `EACCES`
**every** time, and `os.ReadDir` returns entries sorted — so a blocker named
`0blocker.bin` takes the whole rest of the directory with it, on demand, with no
dependence on scheduling:

```
after creating 5 files in the watched directory : 0 events
after editing one of those files in place       : 0 events   <- the real defect
after re-watching the entries                   : 2 events   <- CHMOD + WRITE
```

`TestAbandonedListingLeavesEntriesUnwatchedAndUncharged` asserts that broken
state and keeps asserting it after this change, so it fails loudly the day
fsnotify stops abandoning listings.

## Which consequence is real

The issue named two. Only one of them is a defect:

- **Never watched — real, and the one that matters.** kqueue reports a content
  change on the *file's* descriptor; a directory's `NOTE_WRITE` does not fire for
  a write that leaves the directory's contents unchanged. An entry fsnotify never
  opened therefore produces nothing when it is edited, and grafel stops
  re-indexing it. Permanently: `dirChange` runs only on the next `NOTE_WRITE`, and
  the last write of a burst has no successor. Same failure shape as #6269,
  reached by a different route.
- **Never charged — not an independent defect.** Descriptors that were never
  opened were also never charged, so `used` is an *accurate* account of a wrongly
[...]

Issue #6373, opened 2026-08-19
internal/daemon tests race the scheduler: one deadline flaky-fails under load, another flaky-PASSES and gets weaker the more loaded the machine is


cajasmota:
Observed blocking an unrelated PR (#6371, which touches only the solidity extractor and coverage docs):

```
--- FAIL: TestRebuild_SplitMode_FireAndForget_ReturnsImmediately (1.03s)
    rebuild_wait_test.go:234: fire-and-forget Rebuild did not return immediately
FAIL	github.com/cajasmota/grafel/internal/daemon	49.047s
```

`internal/daemon/rebuild_wait_test.go:233` asserts the RPC returns within a hard **1-second wall-clock deadline**. It failed at **1.03 s** on a Windows runner — three hundredths over. A re-run is in flight; whichever way it lands, the deadline is the defect.

## Two distinct shapes, and the second is worse

The test wants to prove a *structural* property — that `Rebuild` does not block on the rebuild — and expresses it as a race against the scheduler. There are two variants of that mistake in this package, failing in opposite directions:

**Shape 1 — the deadline IS the assertion. Flaky FAIL under load.**

```go
// rebuild_wait_test.go:227-235
select {
case err := <-done:
    …
case <-time.After(1 * time.Second):
    t.Fatal("fire-and-forget Rebuild did not return immediately")
}
```

A slow runner fails a correct implementation. Loud, and it costs a merge — but it is at least visible.

**Shape 2 — the deadline proves a NEGATIVE. Flaky PASS under load, silently.**

```go
// rebuild_split_mode_test.go:85-90
select {
case group := <-rebuilt:
    t.Fatalf("split mode ON must NOT call s.rebuild directly, but got a call for group %q", group)
case <-time.After(300 * time.Millisecond):
    // Expected: no direct call.
}
```

Here a slow runner makes the test **more** likely to pass. If a regression reintroduced the direct call but the call landed at 310 ms, the test goes green. **This test gets weaker exactly when the machine is loaded** — the same silence class as #6321 (670 files → zero entities, `doctor` reports healthy) and pre-fix F#. Shape 1 costs an afternoon; shape 2 costs a regression nobody sees.

## Scope, stated as a search

**Searched:** `time.After(` across `--include='*_test.go'` under `internal/daemon/`. **Found: 133 occurrences.** Filtering to sub-2s deadlines (`N*time.Millisecond` with N ≤ 999, or `1*time.Second`): **12**.

[...]

Issue #6309, opened 2026-08-13
RemoveRepo calls the fsnotify backend while holding w.mu — a deadlock on Windows in shipped code


cajasmota:
Split out of #6307, which is being deferred. **This half is a bug in already-shipped grafel and is independent of the reconciler.**

## Mechanism

fsnotify's Windows backend serialises `Add`/`Remove` through its single I/O goroutine (`w.input <- in; return <-in.reply`), and *that same goroutine* is the only sender on `Events`/`Errors`, via a `sendEvent`/`sendError` that blocks until grafel receives.

So a backend `Add` or `Remove` **completes only while the loop goroutine is draining**. `RemoveRepo` calls the backend while holding `w.mu`; the loop needs `w.mu` to drain; neither can proceed.

kqueue and inotify perform this work in the caller's goroutine and depend on nothing — which is why this is Windows-only and why macOS and Linux CI have never shown it.

Observed as a closed cycle in a CI goroutine dump: `subscribeRepo` holding `w.mu` waiting on `in.reply`; fsnotify's I/O goroutine in `sendError`; the loop — the only receiver — parked on `w.mu`.

## Reachability

Any `RemoveRepo` that overlaps in-flight events: unregistering a repo, a group losing a repo, teardown while a watched tree is active. On Windows the daemon wedges; the watcher stops delivering events entirely until restart.

## Why it was invisible

The #6304 work is what surfaced it — enabling a background sweep in the ledger test suite made the overlap likely enough to hit CI. The defect itself predates all of today's changes.

## Scope

The fix moves the backend calls out of the critical section and adds a regression test that fails on macOS (via an `fsAdd`/`fsRemove` seam, the same technique #6287 used for its Windows deadlock) — so the property is enforced on every platform, not only where it deadlocks.

The reconciler half of #6307 is deferred and stays open as #6304.

Refs #6287, #6304, #6307

Issue #6308, opened 2026-08-13
TestEventChargesDuringTheWalkSurviveTheWalksBookkeeping is flaky on macOS at ~3/60 — it predates the reconciler and can redden the release gate


cajasmota:
Measured on `8768e5c42` — i.e. **before** #6304's reconciler exists, so this is not caused by it:

```
GOMAXPROCS=4 go test -count=60 -run TestEventChargesDuringTheWalkSurviveTheWalksBookkeeping ./internal/daemon/watch/
--- FAIL: TestEventChargesDuringTheWalkSurviveTheWalksBookkeeping (3.65s)
FAIL   250.298s
```

A separate 20-run pass at the same commit was clean, so the rate is roughly a few percent. It also fired once during an unrelated `-race` package run on the #6304 branch, which is how it was noticed.

## Why it matters now

The per-PR gate (#6291) and the tag gate both run this package. At ~3/60 per invocation, across three platforms and two runs per release, this is a realistic source of a red gate that has nothing to do with the change under test — the precise failure mode the gate exists to avoid, and the reason two consecutive matrix runs on the same commit disagreed earlier in this release.

## Note on how it was nearly misattributed

While stressing the ledger suite for #6304, this failure and a second one (`TestChurnReturnsTheLedgerToItsSubscriptionBaseline`, 2/150 at base, 0/150 on that branch) were initially attributed to the new reconciler, and that attribution got as far as a recommendation to ship the sweep disabled. Running the base control is what corrected it.

Worth recording as a method: when a change lands in a package and a test starts failing, **run the control at base before attributing it**. Two of today's conclusions turned on that.

## What a fix needs

The mechanism is not established. It should be, before anything is changed — a poll budget widened without understanding why it was insufficient is the same shrug this project has rejected twice today (see #6304, where `held 0 of 30 polls` actively disproved the timing story).

Refs #6268, #6291, #6304

cajasmota:
Widening this: it is **at least two tests**, both flaky at base, and together they are why the release matrix has been flapping.

## Measured at `8768e5c42` (base — no reconciler, none of today's #6304 work)

| Run | Result |
|---|---|
| `-race -count=3` | clean (137s) |
| `-race -count=6` | **FAIL** — `TestEventChargesDuringTheWalkSurviveTheWalksBookkeeping` (297s) |
| `-count=20` (single test, no race) | clean |
| `-count=60` (single test, no race) | **FAIL** |

[...]

PR #6380, opened 2026-08-19
fix(#6304): rebase-and-retest of the abandoned-listing reconciler onto current main

### LINKED ISSUE #6304: fsnotify silently abandons a directory listing, so grafel neither charges nor watches the remainder — permanently, and it can redden the release gate
Surfaced by a macOS matrix failure on `main` (`22f89e45c`, run 31631684602), which was fully green on the immediately preceding run of the same commit (31629011598).

```
--- FAIL: TestInterleavedDirectoryFillsAcrossReposAreNotDoubleCharged
    fdbudget_6268_test.go:907: every file recreated in place:
        ledger read 636 (held 0 of 30 polls), want a settled 660
```

**This is not the poll budget being too tight, and it is not "CI was slow".** Both of those were actively ruled out — see below.

## The deficit is directory-quantized

`base = 10` (5 per repo, measured), `filled = 10 + 50 × 13 = 660`, observed `636`. The deficit is **24 = 2 directories × 12 files** — two whole directories' file complements, not 24 independent misses.

## Mechanism, confirmed in the vendored source

fsnotify v1.10.1's kqueue backend, `dirChange`: on `f.Info()` returning ENOENT, and on `sendCreateIfNew` returning EACCES/EPERM/ENOENT, it returns **`nil`** and abandons the rest of that directory listing. `readEvents` then discards even the errors that *are* returned. The result is completely silent — no Create event, no entry on the Errors channel, no log line. That matches a CI log with no `add failed`, no EMFILE and no package-level ERROR.

`internal/daemon/watch/watcher.go`'s own header already documents this hazard as a known under-count direction.

## Why it cannot be waited out

Once the write burst ends, **nothing re-lists that directory**. The missing Creates are never delivered — not late, never. `held 0 of 30 polls` across the full 10-second deadline is the evidence: the ledger never reached 660 at any poll, so raising the deadline or `ledgerStableReads` changes nothing.

## The production consequence, which is the part that matters

This is not only a test problem. When fsnotify abandons a listing, grafel:

1. **never charges** those descriptors — the ledger understates what the process holds, which is the #6268 defect: grafel keeps subscribing past the real ceiling and walks toward EMFILE instead of refusing; and
2. **never watches** those files — edits under them do not trigger a re-index, silently, until something else re-lists the directory.

(2) is arguably worse than (1). It is the same class as #6269 (a repo that missed changes while nobody was watching), reached by a different route.

## What needs deciding

Three options, and this is a judgement call:

- **Make the product resilient**: detect an abandoned listing and re-list — e.g. re-trigger on the next `NOTE_WRITE` for that directory, or reconcile periodically. This fixes the real defect but is the largest change.
- **Make the test tolerate it**: assert a bound rather than an equality, or re-trigger a listing before asserting. This makes the gate reliable but *changes what the test measures* — and the exact-equality assertion is what caught two genuine ledger bugs (#6287's over-count, #6293's double release), so weakening it has a real cost.
- **Accept an intermittently red gate.** Not viable: it went red on one of two consecutive macOS runs of the same commit.

Doing nothing leaves the release gate able to fail for a reason unrelated to the change under test, which is the failure mode the whole gate exists to avoid.

## Reproduction

Not reproduced locally in 52 attempts (refusal test: 40 idle + 60 under 8 CPU hogs at `GOMAXPROCS=3`; interleaved: 12 under load + 40 at `GOMAXPROCS=1`; full package 5× under load). This box is 12-core Apple Silicon; the runner is 3-core and much slower. The mechanism is grounded in the vendored source and the quantized deficit rather than in a local reproduction — stated as inference, not observation.

Refs #6268, #6269, #6287, #6293

cajasmota:
# fix(#6304): nothing ever re-listed a directory whose fsnotify listing was abandoned

Closes #6304.

## What this PR is

**This is a rebase-and-retest of the closed #6307, not a new implementation.** The change is
`1389dcc10` replayed onto current `main` (`83f3c898a`). No reconciler logic was rewritten and
no attempt was made to "fix" the hang described below. The point of the PR is to put the branch
in front of Windows CI on a current base and see what happens.

The full defect analysis, the darwin reproduction and the design rationale are unchanged and live
in the commit message and in #6307's body.

## Why now

#6307 was closed unmerged on 2026-08-13 for three reasons. Two have expired:

1. `v0.2.2` was being held on a set of ready fixes — **`v0.2.2` shipped.**
2. It is ~1,700 lines in the most-churned package in the tree — **still true.** A reason for care,
   not a blocker.
3. It failed Windows CI twice: a genuine deadlock (fixed on the branch), then a **14-minute hang in
   `TestReconcileVisitsEveryDirectoryAcrossPasses` whose mechanism was never established** — this is
   the live blocker and the thing being tested here.

The branch was based on `8768e5c42` and was 19 commits behind `main`. Two of the commits it was
missing bear directly on that hang:

- **`66e260d9a` (#6310)** — *"RemoveRepo called fsnotify's backend under `w.mu`, which deadlocks on
  Windows."* This was split out of *this very PR* when it was deferred, and landed afterwards. The
  Windows run that hung for 14 minutes therefore ran **without** the Windows deadlock fix that this
  PR's own analysis produced.
- **`6813ff540` (#6311)** — *"the ledger-invariant helper sampled its two halves at two different
  instants, so a loaded machine manufactured a mismatch."* The hanging test's final assertion is
  `assertLedgerMatchesPerRepo`, i.e. exactly that helper, in its pre-fix form, on a loaded CI runner.

Neither is proof. The rebase discriminates: **hang gone → the mechanism was one of those two;
hang persists → the two likeliest causes are eliminated** and the search narrows, on a current
branch rather than a 19-commit-stale one.

## What the rebase resolved

Conflicts were all in `internal/daemon/watch/watcher.go` (five hunks). Two of them are places where
the branch carried its **own earlier version of a fix that has since landed on `main`**. In both,
`main`'s version is kept:

- **`RemoveRepo` locking (#6310 / `66e260d9a`).** `main` already collects the detached directories
  under `w.mu` and calls `unwatchAll` after releasing it. `main`'s structure and comments are kept;
  the only thing taken from the branch is the `detachDirLocked(d)` call, which additionally yields
  the per-entry watches `repairDir` opened inside `d` (`fs.Remove(dir)` does not reach them, because
  fsnotify skips user-added paths). The branch's competing formulation of the same fix was dropped.
- **The refusal unwind in `subscribeRepo`.** `main`'s bare `fs.Remove` loop before the lock was
  dropped in favour of the branch's `detachDirLocked` / `unwatchAll` pair, which is a superset of it
  (same "no backend call under `w.mu`" discipline, plus the repair's entry watches).
- **`Watcher.fsAdd`/`fsRemove` doc comment** — `main`'s wording, which cites #6309.

Two more resolutions, in the test files:

- **`assertLedgerMatchesPerRepo` (#6311 / `6813ff540`).** `main`'s single-acquisition version
  survives on the rebased branch; the branch carried no competing copy. This is deliberate — it is
  one of the two mechanisms under test.
- **`lockAssertingBackend` + `TestRemoveRepoDoesNotCallTheBackendUnderTheLock` were declared twice**
  after the rebase, once in the branch's `reconcile_6304_test.go` and once in `main`'s
  `removerepo_6309_test.go` (the file split out of this PR). `main`'s helper is kept — it is the
  same stand-in plus a record of what was called. The branch's duplicate helper was deleted; its
  version of the test, which exercises `RemoveRepo` *after* a repair has taken entry watches, is
  kept under the non-colliding name
  `TestRemoveRepoAfterRepairDoesNotCallTheBackendUnderTheLock`.

## Verification done locally

- `go build ./...` — clean.
- `go vet ./internal/daemon/...` — clean (this is what caught the duplicate test declaration;
  `go build` does not compile tests).
- `gofmt -l internal/daemon/watch/` — clean.

The Go suite was **not** run locally. The gate here is Windows CI, which cannot be reproduced on
this machine, and the toolchain was in use.

## A green run is not sufficient evidence to merge

Stated plainly, because it would be easy to read a green Windows job as the end of the question.
See **#6373**: `internal/daemon` tests contain wall-clock deadlines that fail under load in one
direction and **silently pass** in the other. A green run on a lightly-loaded runner is weaker
evidence in this package than it looks.

So this PR is asking CI a question, not asserting an answer. Merging should additionally require:

- the Windows job green on repeated runs, not one,
- a decision on whether the 14-minute hang's mechanism is now *explained* or merely *not currently
  reproducing*,
- and, per #6307's closing comment, this landing on its own early in a cycle rather than bundled —
  `internal/daemon/watch` produced six defects in a single day, and this is 590 new lines in it.

Refs #6307, #6310, #6311, #6306, #6308, #6373

### DIFF
--- a/internal/daemon/watch/fdbudget_6268_test.go
+++ b/internal/daemon/watch/fdbudget_6268_test.go
@@ -1139,7 +1139,21 @@ walked:
 func TestInterleavedDirectoryFillsAcrossReposAreNotDoubleCharged(t *testing.T) {
 	repoA := makePrunedTree(t)
 	repoB := makePrunedTree(t)
-	w := newBudgetedWatcherCfg(t, Config{FDBudget: 1_000_000, disableQuarantine: true})
+	// This test's arithmetic is the only one in this file that a real macOS run
+	// has been observed to miss: fsnotify's dirChange abandoned two directories'
+	// listings mid-burst, so 24 of the 660 descriptors were never opened, never
+	// charged and — the part that matters off CI — never watched again. Nothing
+	// re-listed a subscribed directory, so the reading stayed 24 short for the
+	// whole 10-second deadline (`held 0 of 30 polls`).
+	//
+	// The assertion is unchanged and still an exact equality: 660, not "at
+	// least". What changed is that the product now has a way to get there, at
+	// the production sweep cadence — this test sets no interval, exactly like
+	// every other test in the file (#6304).
+	w := newBudgetedWatcherCfg(t, Config{
+		FDBudget:          1_000_000,
+		disableQuarantine: true,
+	})
 	for _, r := range []string{repoA, repoB} {
 		if _, err := w.AddRepo(r); err != nil {
 			t.Fatalf("premise broken: AddRepo(%s): %v", r, err)
--- a/internal/daemon/watch/fdbudget_test.go
+++ b/internal/daemon/watch/fdbudget_test.go
@@ -168,15 +168,32 @@ func newBudgetedWatcher(t *testing.T, budget int) *Watcher {
 // timings are the same in both.
 func newBudgetedWatcherCfg(t *testing.T, cfg Config) *Watcher {
 	t.Helper()
+	w := newBudgetedWatcherCfgNoCleanup(t, cfg)
+	t.Cleanup(w.Stop)
+	return w
+}
+
+// newBudgetedWatcherCfgNoCleanup is newBudgetedWatcherCfg without the automatic
+// Stop, for a test that calls Stop itself and would otherwise park on stopOnce.
+func newBudgetedWatcherCfgNoCleanup(t *testing.T, cfg Config) *Watcher {
+	t.Helper()
 	cfg.Debounce = time.Hour
 	cfg.BulkThreshold = 10000
 	cfg.HeartbeatInterval = time.Hour
+	// The reconcile sweep is deliberately NOT pinned. Debounce, bulk and
+	// heartbeat are, because those three feed the SINK and a test asserting
+	// ledger arithmetic does not want reindex signalling in the way. The sweep
+	// touches the LEDGER, which is exactly what these tests measure — so every
+	// one of them runs it, at its production cadence, and any pass where the
+	// sweep is not ledger-neutral is a failure this suite is entitled to catch
+	// (#6304). Tests that drive it on demand call reconcileOnce directly, and
+	// those switch the automatic one off so the two do not race — see
+	// withheldEventsWatcher.
 	cfg.fdCost = kqueueCostModel // test the macOS arithmetic everywhere
 	w, err := NewWatcherConfig(cfg, func(string, bool) {}, nil)
 	if err != nil {
 		t.Fatalf("NewWatcherConfig: %v", err)
 	}
-	t.Cleanup(w.Stop)
 	return w
 }
 
--- a/internal/daemon/watch/reconcile.go
+++ b/internal/daemon/watch/reconcile.go
@@ -85,11 +85,14 @@ import (
 // abandonment is permanent, so waiting one sweep to confirm costs the repair
 // that matters nothing at all.
 //
-// WHY THE REPAIR RUNS ON THE LOOP GOROUTINE. The scan is cheap and side-effect
-// free, so it stays on the sweep goroutine; the repair is handed to the loop
-// over Watcher.repairCh. See the field for why. Empirically: with the repair on
-// the sweep goroutine the same interleaved-fill test misreads the ledger about
-// one run in eight, in both directions.
+// WHY THE REPAIR RUNS ON THE SWEEP GOROUTINE. Scan and repair both run on the
+// sweep goroutine — reconcileLoop calls reconcileOnce, which calls scanDir,
+// which calls repairDir inline. Keeping the repair off the loop goroutine is a
+// hard requirement, not a preference: it issues one fsAdd per entry, and on
+// Windows an Add only completes while the loop goroutine is draining Events, so
+// an Add issued FROM that goroutine never gets its reply. See the invariant on
+// Watcher.fsAdd, and the note on repairDir for what running off the loop costs
+// (a Create can land mid-repair) and how that is handled.
 //
 // WHY IT DOES NOT FIGHT THE QUARANTINE. A quarantined directory is skipped
 // outright. A directory that churns pathologically is one the tracker is
@@ -202,10 +205,11 @@ func (w *Watcher) reconcileOnce() (repaired int) {
 			return repaired
 		}
 		if acted >= maxRepairEntriesPerPass {
-			// Scanning is cheap and would finish the batch; ACTING is what puts
-			// work on the loop goroutine, so acting is what is bounded. The
-			// directories left over keep their confirmed deficit and are acted
-			// on next pass.
+			// Scanning is cheap and would finish the batch; ACTING is what
+			// issues the fsAdd calls, so acting is what is bounded. The
+			// directories left over keep their confirmed deficit; this batch is
+			// already popped, so they are acted on in a later cycle, once the
+			// rotation comes back round to them.
 			break
 		}
 		out := w.scanDir(dir, cost)
@@ -241,13 +245,13 @@ func (w *Watcher) nextReconcileBatch() []string {
 }
 
 // scanDir looks for a directory short of watched entries and, on a CONFIRMED
-// deficit, hands the repair to the loop goroutine. It reports whether entries
-// were actually recovered.
+// deficit, calls repairDir. It reports whether entries were actually recovered.
 //
-// This is the half that runs on the SWEEP goroutine, and it is deliberately the
-// half that does the reading: one os.ReadDir, two short critical sections, no
-// ledger mutation and no fsnotify call. The entry list it builds travels with
-// the request, so the loop goroutine never performs a directory read.
+// This is the reading half, and it does no more than read: one os.ReadDir, two
+// short critical sections, no ledger mutation and no fsnotify call. The entry
+// list it builds is handed to repairDir, which runs on this same SWEEP
+// goroutine — never on the loop goroutine, which must stay free to drain Events
+// (see repairDir).
 func (w *Watcher) scanDir(dir string, cost fdCostModel) scanOutcome {
 	w.mu.Lock()
 	repo, watched := w.dirToRepo[dir]
@@ -344,10 +348,10 @@ func (w *Watcher) scanDir(dir string, cost fdCostModel) scanOutcome {
 	return scanOutcome{repaired: res.charged > 0, entriesActed: len(entries)}
 }
 
-// maxRepairEntriesPerPass bounds the work one pass may put on the LOOP
-// goroutine, counted in the unit that work is actually done in — one fs.Add per
-// entry — rather than in directories, which says nothing about size. At 4096
-// register syscalls it is a few milliseconds of the drain per pass, against a
+// maxRepairEntriesPerPass bounds the work one pass does on the SWEEP goroutine,
+// counted in the unit that work is actually done in — one fs.Add per entry —
+// rather than in directories, which says nothing about size. At 4096
+// register syscalls it is a few milliseconds of the sweep per pass, against a
 // subscribeDirRecursive that the event path already runs inline with no bound at
 // all. Scanning is bounded separately by reconcileBatch; this bounds acting.
 //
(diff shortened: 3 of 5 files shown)


cajasmota:
## The hang reproduces — and the mechanism is now established

`test (windows-latest)` failed on the rebased branch:

```
panic: test timed out after 15m0s
	running tests:
		TestReconcileVisitsEveryDirectoryAcrossPasses (13m51s)
FAIL	github.com/cajasmota/grafel/internal/daemon/watch	900.040s
```

**Both candidate mechanisms are eliminated.** The rebase carries `66e260d9a` (#6310, the Windows `RemoveRepo` locking fix) and `6813ff540` (#6311, the ledger-invariant helper), and the hang is unchanged at 13m51s. That is what this PR was for: a clean negative result that narrows the search.

## What is actually stuck

Two goroutines, both parked 13 minutes:

```
goroutine 1991 [chan receive, 13 minutes]:
  fsnotify.(*readDirChangesW).AddWith(...)  backend_windows.go:145
  fsnotify.(*Watcher).Add(...)              fsnotify.go:341
  watch.(*Watcher).subscribeRepo.func2(...)
  watch.(*Watcher).subscribeRepo(...)
  watch.(*Watcher).AddRepo(...)
  watch.TestReconcileVisitsEveryDirectoryAcrossPasses(...)

  fsnotify.(*readDirChangesW).sendEvent(...)  backend_windows.go:76
  fsnotify.(*readDirChangesW).readEvents(...) backend_windows.go:642
```

and, earlier in the same log:

```
level=ERROR msg="watcher: error" pkg=watch err="fsnotify: queue or buffer overflow"
```

## The cause is a test harness, not the reconciler

`reconcile_6304_test.go:469` — `withheldEventsWatcher` installs **unbuffered** channels as the backend's Events/Errors and nothing ever drains them:

```go
ev := make(chan fsnotify.Event)
errs := make(chan error)
cfg.testEvents, cfg.testErrors = ev, errs
```

The test then writes a file into each of `reconcileBatch+40` = **296** directories.

This repo already documents the exact hazard, at `internal/daemon/watch/watcher.go:172-185`:

> fsnotify's Windows backend serialises Add and Remove through its single I/O goroutine — `w.input <- in; return <-in.reply` — and that same goroutine is the only sender on Events and Errors, via a sendEvent/sendError that blocks until grafel receives. **So an Add or Remove completes only if the loop goroutine is free to drain.**

That comment pins two ways to starve the drain: holding `w.mu` across an `Add`, and calling `Add` from the loop goroutine. **This is a third: a harness that installs channels nobody reads.** The I/O goroutine blocks in `sendEvent` on an unbuffered channel, so the next `fsAdd` from `AddRepo` never receives its reply. Circular wait.

Off Windows, kqueue and inotify perform `Add` in the caller's goroutine and depend on nothing — which is exactly why this is invisible locally and has only ever appeared on Windows CI, and why it went unexplained for a week.

**Status: the stacks and the overflow error are measured; the causal step is inferred from them plus the mechanism the repo documents at `watcher.go:172-185`. Not yet confirmed by a fix.**

## Consequences

- **The reconciler is not implicated.** `reconcile.go` is byte-identical to the parked version and appears nowhere in the stuck stacks. This is a defect in the test's own harness.
- **Not a production defect** on this evidence: the daemon drains Events on its loop goroutine; only this harness withholds them.
- The fix is to drain `ev`/`errs` for the lifetime of the test, or buffer them beyond the event volume. Whoever does it must first establish whether withholding is load-bearing for what these tests assert — they count `reconcileOnce()` return values, not events, which suggests it is not, but that needs checking rather than assuming.
- Worth considering: the invariant comment at `watcher.go:172-185` should gain this third variant, since two engineers have now hit the class and the comment enumerates only two ways in.

Not merging. Fix queued behind the VB.NET S4 parser, which currently holds the single Go-compiling slot.

cajasmota:
## Windows is green — the hang is fixed, and the mechanism is now confirmed rather than inferred

```
ok  github.com/cajasmota/grafel/internal/daemon/watch   115.020s
test (windows-latest): pass
```

**115 seconds, from a 15-minute timeout.** The package genuinely ran — it is in the log with a real duration, not skipped.

That closes a hang whose mechanism was recorded as "never established" when #6307 was deferred on 2026-08-13. The chain that got there:

1. **Rebase eliminated both standing hypotheses.** The branch was 19 commits stale and missing `66e260d9a` (#6310, the Windows `RemoveRepo` locking fix — split out of this very PR when it was deferred) and `6813ff540` (#6311, the ledger-invariant helper, which is the hanging test's final assertion). With both present the hang was unchanged at 13m51s. A clean negative result.
2. **The stacks named the real cause.** Two goroutines parked 13 minutes — one in `fsnotify.AddWith` under `AddRepo`, one in `sendEvent` — plus `err="fsnotify: queue or buffer overflow"`.
3. **`withheldEventsWatcher` installed unbuffered Events/Errors channels that nothing drained**, while the test wrote a file into each of 296 directories. fsnotify's Windows backend serialises `Add` through the same goroutine that blocks sending events, so `Add` never received its reply. Circular wait.

This was the **third** way to starve that drain. `watcher.go:172-185` documented two — holding `w.mu` across an `Add`, and calling `Add` from the loop goroutine — and now lists three, with `testEvents` named as the mechanism and this helper as the worked example.

### Two things established along the way

**Withholding was load-bearing, but only from the loop goroutine — which is not what deadlocks.** `testEvents` substitutes what the *loop* drains; the tests' premise is that no backend Create is ever charged to the fd ledger. Draining the backend's own channels into a discard sink preserves that premise byte-for-byte, so no assertion moved. Established by reading the wiring, not assumed.

**`stoppableWatcher` has the identical shape and is safe for a fragile reason** — its four tests never call `AddRepo`, so the backend holds an empty watch set and never parks in `sendEvent`. Safe by a property of its callers, not of the helper. The expanded invariant comment is what should catch the day someone adds an `AddRepo` there.

Drain rather than buffer, deliberately: the volume is one event per file per directory, so any fixed buffer is a size that goes silently wrong the next time a fixture grows.

### Not merging on this

Green Windows answers the question that deferred #6307. It does not answer the *other* stated reason — this is ~1,700 lines in the package that produced six defects in a single day, and the reconciler itself has not been reviewed in this cycle. That review is queued behind VB.NET S5 for the single Go-compiling slot.

Worth restating from #6373: this package's tests contain wall-clock deadlines that fail under load in one direction and **silently pass** in the other, so a green run here is weaker evidence than it looks. It is necessary, not sufficient.

cajasmota:
## Reconciler ↔ FD-ledger review: APPROVE, plus one comment that was actively dangerous

This is the review this PR still owed. It was scoped deliberately to **the fd-ledger interaction** rather than to all ~1,700 lines, because that is the defect class this package actually has: #6306 (ledger and per-repo tally moved in separate critical sections, so a refused subscription stranded a descriptor) and #6311 (the invariant helper sampled its two halves at two different instants). Both were fd-ledger bugs, both within a day.

**Nothing of that class is reintroduced.** Every mutation moves the ledger and `fdReserved[repo]` inside one `w.mu` section (`reconcile.go:437-450`, `:568-577`, `unreserve()` at `:453-471`). All six abandon paths either never reserved or call `unreserve()`; the `dirToRepo[dir] != repo` case correctly *skips* the release because `RemoveRepo` (`watcher.go:838-842`) already released the whole `fdReserved[abs]`. `assertLedgerMatchesPerRepo` still reads both halves under one acquisition, and the reconciler only respects it.

### The finding worth keeping

`reconcile.go:88-90` claimed the repair runs **on the loop goroutine**, delivered over `Watcher.repairCh`. `reconcile.go:384` claimed the opposite. `grep -rn "repairCh" internal/` returns exactly one hit — the comment asserting it. **The field never existed.** The real path is `go w.reconcileLoop()` (`watcher.go:466`) → `reconcileOnce` → `scanDir` → `repairDir`, all inline: the repair runs on the **sweep** goroutine.

That is not a documentation nit. It is precisely the property behind this PR's Windows deadlock — fsnotify's backend serialises `Add` through the same goroutine that blocks sending events, so an `Add` issued from that goroutine never gets its reply (`watcher.go:172-185`). A comment answering it both ways is a trap.

And it had already sprung one: **`:516` called its own guard "unreachable while the placement is correct"** — reasoning from the false premise. Because the repair really is off the loop, a Create *can* land mid-repair, so that branch is the live guard. The comment was inviting a future reader to delete working code.

Four sites carried the same false claim (`:88`, `:243`, `:347`, `:516`), not the two first spotted.

### One deletion, verified independently

The `charged > est` arm at `:570` is unreachable and is gone. I checked the argument myself rather than taking it:

- `added` is initialised empty and appended at most once per iteration of `for _, p := range entries`, so `len(added) <= len(entries)`.
- `est` and `charged` use the **same** `have`, and `have` is re-verified unchanged between them.
- The leg that completes it: `:413`'s `if len(entries) <= have { return settled }` means `len(entries) > have` strictly past that point, so `est` cannot go negative — without which `charged` (clamped at 0) *could* exceed a negative `est`.
- `perEntry` is 1 (kqueue) or 0 (inotify), never negative. At 1 the inequality holds; at 0 both sides are 0.

No behaviour change on either platform.

### Left alone deliberately

The `fdPreCharged` residue at `:564` — a one-descriptor under-count, self-healed by the next sweep, same acknowledged class as `subscribeDirRecursive`.

### Scope, stated plainly

This reviewed the ledger interaction. It did not review all 1,700 lines, and "reviewed" should not be read as more coverage than that.

Issue #6363, opened 2026-08-19
the VB.NET epic has zero .vb files to test against, and S3's review found three declaration-corrupting defects because of it


cajasmota:
The VB.NET epic (#6327) has no `.vb` file to test against. Not a shortage — **zero**, across `~/Projects/archigraph`, `~/Projects/archigraph-corpora`, `~/go/pkg/mod` and every other tree on the development machine. Every fixture in S3 (#6362) is constructed from the language reference.

S3's review shows what that costs, and the cost is not theoretical.

## Three defects that only real files would have surfaced

All three were found by an adversarial reviewer hand-constructing syntax, not by the 156-subtest suite:

- **`Option Explicit On` swallowed the following line.** `On` was in the continuation-keyword set for LINQ `Join … On`. `Option Strict On` / `Option Explicit On` / `Option Infer On` are the **first line of a large fraction of legacy VB files**, so `Public Class C` was joined into the header, `walkStatement` returned on an unrecognised head, and the class was never declared. Every member of the file landed at file scope.
- **`End With` swallowed the next statement**, desynchronising the container stack for the remainder of the file — methods nesting inside other methods, locals nesting inside those.
- **`32&`** (a `Long` type-character literal, pervasive in `Declare`/Win32 interop) swallowed the following declaration.

The reviewer's diagnosis of *why* these survived is the transferable part: **a blank line after `Option` masks the first one.** Hand-written fixtures are tidy. Real files are not. This is the same fixtures-versus-reality failure recorded in #6347, arriving in the one place where there is no reality available to check against.

## Why this compounds rather than stays flat

S3 is the foundation for S4 (parser), S5 (edges) and S7 (WinForms depth). A pre-pass defect does not stay in the pre-pass — it produces a wrong declaration table, and S4 then classifies `(` wrongly with confidence. S5 emits edges from that. The graph looks populated and is wrong, which is worse than empty.

We are about to write the largest hand-rolled parser in the repo against zero examples of the language.

## What is available

VB.NET is not obscure — there is a large body of public, permissively-licensed code:

- **`microsoft/VBLanguageDesign`**, the language design repo, plus the VB samples in `dotnet/samples`.
- **WinForms sample applications** and the `.Designer.vb` shapes that @dcastro-imp reported (`#Region`, `Inherits System.Windows.Forms.Form`, `Handles` wiring) — precisely the population VB.NET support targets.
[...]

PR #6381, opened 2026-08-19
feat(#6327): S4 — the VB.NET recursive-descent parser, measured on 302 real files

cajasmota:
S4 of the VB.NET epic. Refs #6327, #6321, #6362, #6363.

The recursive-descent parser: a containment tree with spans, and a reference pass that classifies every `name(` as call or index. **No entities, no relationships, no registry** — nothing outside `internal/vbnet` imports it. That is S5.

## Real-corpus parse rate: 300 / 302 (99.3%)

302 files, 148,308 lines, 88 `.Designer.vb`, across three open-source WinForms projects. Verified independently:

```
corpus_6327_test.go:95: 300/302 files parsed clean (99.3%)
--- PASS: TestCorpusParseRate
--- PASS: TestCorpusTreeAgreesWithTable
```

**First run was 230/302 (76.2%).** Every defect that closed the gap came from real files, not fixtures — which is the whole argument of #6363, now with a number on it:

| Cause | Diagnostics | Fix |
|---|---|---|
| Multi-line lambdas (`x = Sub(v) … End Sub`) closing the enclosing *method* | 155 + 8 | lambda frames |
| UTF-8 BOM: `Partial Class X` unrecognised, `Inherits` orphaned | 11 designer files, +16 `End Class`, +5 `End Namespace` | BOM strip in the joiner and in spans |
| Interpolated strings scanned **inverted** — hole as text, text as code | corrupted masking in 106 of 302 files | `scanInterpolated` |
| Enum member named `Custom` (also the Custom Event modifier) | 3 | enum recognised before modifier peel |
| `Function() As List(Of Job)` generic return on a multi-line lambda | 1, plus 2 regressions caught by re-measuring | `consumeTypeExpr` |

**Residual, characterised:** both remaining files share one cause — VB 14 allows a string literal *and* an interpolation hole to span physical lines, and the pre-pass is line-oriented. Both are **reported as diagnostics, not silently mis-parsed**, and `TestMultiLineLiteralIsUnsupported` asserts the failure direction so lifting the limit forces the test to change.

## The invariant that protects S5

`TestCorpusTreeAgreesWithTable` — 5,919 table types/methods, all present in the tree at the same scope path, 0 divergences. This is what stops S3's declaration table and S4's tree drifting into classifying a use site for a scope the tree never puts it in.

I verified it is load-bearing with a mutant of my own, outside the author's set: removing `NodeStructure` from `IsType()` (so structures stop owning a scope) fails `TestCorpusTreeAgreesWithTable`.

## Mutants — 11 by the author, all killed

Including: case-sensitive `FoldName` (kills 53 S3 subtests plus the ref classification), no BOM strip (corpus drops to 92.1% with `11 x inherits outside a type declaration`), no lambda frames (84.4%, `155 x End sub with no matching block`), `New` not treated as a call, declaration headers leaking into use sites, and qualified names resolved locally (`Other.Compute` must stay unknown, not bind).

The author also found **its own fixture vacuous**: the first `TestMultiLineLiteralIsUnsupported` used a tidy multi-line literal, which parses fine. It only failed once `End Function` was placed *inside* the literal, mirroring the real `Audio.vb`. Same fixture-vs-reality failure #6363 exists to record.

## Two findings from my review — for a follow-up, not blocking

1. **`corpusFiles` skips when the path is set but yields no files.** `t.Skipf("no .vb files under %s")` is right for an *unset* variable, but a set-and-wrong path is a configuration error, not an absent corpus — I hit it, and the run reported success. In CI that is a gate that silently does not run. Should be `t.Fatalf` when the variable is set.
2. **The corpus walk swallows walk errors** (`if err != nil { return nil }`), so a permission error mid-walk shrinks the denominator and the rate is computed over survivors — a `>= 0.99` floor over a smaller set can pass on partial data.

Both are the same class the parser itself is careful about: a measurement that can quietly not happen.

## For the S5 author

- **Anchor to the type, not the file.** `Inherits`/`Implements` are already on the type node, never on `Result.File`. Leave `FromID` empty so assembly stamps the owner — `internal/extractors/file_anchored_rels_guard_test.go` fails on a new file-anchored relationship (this defect has now appeared in five languages: #6295, #6298, #6365). Nothing here was pre-built against that guard.
- **Use `Ref.IsCall()`, don't re-derive it** — it composes the table's `ParenKind` with `New` and `StatementHead` and excludes intrinsics. `Ref.Intrinsic` must not produce a CALLS edge; one designer file has hundreds of `CType`.
- ~~**`Ref.Qualifier != ""` means `Kind` is deliberately unknown** (except `Me`/`MyClass`). That is the epic's confidence signal.~~ **WRONG — do not follow this. Superseded by review.** `qualifierBefore` returns `("", true)` when the qualifier is an expression or a `With` target, and the `qualified` bit is discarded before it reaches `Ref` — so `.OpenSubKey(k)` inside a `With` is byte-identical to a genuinely unqualified name. **Measured: 1,603 of 41,240 refs.** Reading `Qualifier == ""` as "unqualified, resolve against this file" produces exactly the confidently-wrong edge the epic exists to prevent. A `Qualified bool` is being added; wait for it. Also undocumented until the fix lands: **every `With`-block member invocation is silently dropped** (`.SetValue(...)` as a whole statement is a real call with `IsCall() == false`).
- `AddressOf Foo` is not recorded (no parens); S7's `Handles` wiring will want it. `Node.Handles` is already populated.

`Option Infer` still has **zero** corpus instances and is covered only by a constructed fixture — stated so nobody reads "measured on 302 real files" as covering it.

### DIFF
--- a/internal/vbnet/attributes.go
+++ b/internal/vbnet/attributes.go
@@ -16,7 +16,55 @@ import "strings"
 // contain string literals holding '>' — both are tracked, so the terminating
 // '>' is the one at paren depth zero outside a literal.
 func SplitAttributes(code string) (attrs []string, rest string) {
+	ranges := attributeRanges(code)
+	if len(ranges) == 0 {
+		return nil, strings.TrimSpace(code)
+	}
 	var out strings.Builder
+	prev := 0
+	for _, r := range ranges {
+		attrs = append(attrs, strings.TrimSpace(code[r.start+1:r.end]))
+		seg := code[prev:r.start]
+		// Drop the whitespace an inline group was separated from its '(' or
+		// ',' by, so removing `<Out>` from `F(a, <Out> ByRef n)` leaves one
+		// space rather than two. The space AFTER the group is kept, which is
+		// what stops the removal from ever joining two tokens.
+		if t := strings.TrimRight(seg, " \t"); t != seg &&
+			(strings.HasSuffix(t, "(") || strings.HasSuffix(t, ",")) {
+			seg = t
+		}
+		out.WriteString(seg)
+		prev = r.end + 1
+	}
+	out.WriteString(code[prev:])
+	return attrs, strings.TrimSpace(out.String())
+}
+
+// BlankAttributes replaces every attribute group with spaces, preserving byte
+// length so offsets into the result still index the input.
+//
+// SplitAttributes removes the groups, which shifts every later offset; the
+// parser needs them gone *and* needs a use site's offset to still map back to
+// the physical line it came from, which is what this preserves.
+func BlankAttributes(code string) string {
+	ranges := attributeRanges(code)
+	if len(ranges) == 0 {
+		return code
+	}
+	b := []byte(code)
+	for _, r := range ranges {
+		for i := r.start; i <= r.end && i < len(b); i++ {
+			b[i] = ' '
+		}
+	}
+	return string(b)
+}
+
+// attributeRanges returns the inclusive [start,end] byte ranges of every
+// attribute group in code, in source order. start indexes the '<' and end the
+// matching '>'.
+func attributeRanges(code string) []span {
+	var ranges []span
 	i := 0
 
 	// Leading groups: skip whitespace, then consume every adjacent <...>.
@@ -32,7 +80,7 @@ func SplitAttributes(code string) (attrs []string, rest string) {
 		if !ok {
 			break
 		}
-		attrs = append(attrs, strings.TrimSpace(code[j+1:end]))
+		ranges = append(ranges, span{j, end})
 		i = end + 1
 	}
 
@@ -40,14 +88,11 @@ func SplitAttributes(code string) (attrs []string, rest string) {
 	for i < len(code) {
 		c := code[i]
 		if c == '"' {
-			// Copy the literal verbatim; '<' inside it is data.
-			end := literalEnd(code, i)
-			out.WriteString(code[i:end])
-			i = end
+			// '<' inside a literal is data.
+			i = literalEnd(code, i)
 			continue
 		}
 		if c == '(' || c == ',' {
-			out.WriteByte(c)
 			i++
 			j := i
 			for j < len(code) && (code[j] == ' ' || code[j] == '\t') {
@@ -58,9 +103,7 @@ func SplitAttributes(code string) (attrs []string, rest string) {
 				if !ok {
 					break
 				}
-				attrs = append(attrs, strings.TrimSpace(code[j+1:end]))
-				// Consume the group but not the whitespace around it, so that
-				// removing an attribute never joins two tokens together.
+				ranges = append(ranges, span{j, end})
 				i = end + 1
 				j = i
 				for j < len(code) && (code[j] == ' ' || code[j] == '\t') {
@@ -69,10 +112,9 @@ func SplitAttributes(code string) (attrs []string, rest string) {
 			}
 			continue
 		}
-		out.WriteByte(c)
 		i++
 	}
-	return attrs, strings.TrimSpace(out.String())
+	return ranges
 }
 
 // attributeEnd returns the index of the '>' closing the attribute group that
--- a/internal/vbnet/continuation.go
+++ b/internal/vbnet/continuation.go
@@ -21,6 +21,32 @@ type LogicalLine struct {
 	// Line and EndLine are 1-based physical line numbers, inclusive.
 	Line    int
 	EndLine int
+	// Pieces records, for each physical-line fragment joined into Text, the
+	// byte offset in Text at which it starts and the physical line it came
+	// from. Without it a use site on the third physical line of a continued
+	// statement would be reported at the statement's first line, which is the
+	// span accuracy S4 needs and the 46 corpus files carrying ' _' would lose.
+	Pieces []LinePiece
+}
+
+// LinePiece maps one offset in LogicalLine.Text back to a physical line.
+type LinePiece struct {
+	// Offset is the byte offset in Text where the fragment starts.
+	Offset int
+	// Line is the 1-based physical line the fragment came from.
+	Line int
+}
+
+// LineAt returns the 1-based physical line holding the byte at offset in Text.
+func (l LogicalLine) LineAt(offset int) int {
+	line := l.Line
+	for _, p := range l.Pieces {
+		if p.Offset > offset {
+			break
+		}
+		line = p.Line
+	}
+	return line
 }
 
 // continuationKeywords are the keywords after which VB.NET 10+ permits an
@@ -59,10 +85,17 @@ const continuationTrailingBytes = ",({.&=+-*/\\^<"
 // comment or a literal can never drive joining), explicit '_' continuation,
 // implicit continuation, and attribute splitting.
 func JoinContinuations(src string) []LogicalLine {
+	// A UTF-8 byte-order mark is not whitespace, so leaving it in place makes
+	// the first statement of the file start with three bytes that are neither
+	// an identifier nor a keyword: `Public Class Crash` is never recognised
+	// and every member of the file lands at file scope. 12 of the 302 corpus
+	// files start with one, most of them .Designer.vb (#6363).
+	src = strings.TrimPrefix(src, "\ufeff")
 	raw := strings.Split(strings.ReplaceAll(src, "\r\n", "\n"), "\n")
 
 	var out []LogicalLine
 	var buf []string
+	var pieces []LinePiece
 	var doc, comments []string
 	var pendingDoc []string
 	depth := 0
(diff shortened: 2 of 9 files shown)


cajasmota:
## Independent review: REQUEST-CHANGES (narrow) — one API gap S5 is about to be built on, and four claims the measurements contradict

I asked the reviewer to find a file that parses **clean and wrong**, on the grounds that a parser's real failure mode is not a crash but a confident wrong tree that S5 emits edges from. It found that class, sized it, and otherwise the parser won.

### What held, measured

- **Span accuracy** — 5,892 of 5,896 type/method nodes end exactly on `End <keyword>`, and `src[StartByte:EndByte]` ends there too. The 4 misses are `End Function 'comment`, an artifact of the check rather than a defect. BOM and CRLF handling in `spanOf` correct.
- **`LogicalLine.Pieces` / `LineAt`** — the riskiest new mechanism. **0 of 41,240** refs fall outside their owner's span; **0 of 41,240** land on a physical line not containing the ref's name.
- **`BlankAttributes`** preserves byte length unconditionally, so `parse.go:529`'s raw slicing is safe. `SplitAttributes`' rewrite is behaviour-identical on the paths traced; the 238-entry S3 suite is green.
- **Paren classification** on every hard shape probed: shadowed locals (longest visible scope wins), `With` blocks, `Foo(1).Bar(`, default properties on a non-array value, late binding on `Object`, `My.`, arrays of delegates (`h(i)(x)` — the second paren correctly reads as grouping and produces no ref), 2,638 intrinsics excluded from `IsCall()`. **"`Kind` is never guessed from shape" is true.**
- **`scanInterpolated` vs the VB `$` type character** — 11 corpus lines have `$"` after an identifier byte; all 11 are inside ordinary literals consumed before the guard. No instance.

### BLOCKER — `Ref` discards the `qualified` bit, and this PR body told S5 to rely on it

`parenrefs.go:106` computes `qual, qualified := qualifierBefore(...)`. `qualified` feeds `Intrinsic`, the `parenKeywords` skip and `classify`, then is **thrown away** — only `Qualifier string` reaches the `Ref`. Verified by reading it.

`qualifierBefore`'s own doc says it returns `("", true)` when the qualifier is an expression or a `With` target. So `.OpenSubKey(k)` inside a `With`, and `CType(x, I).BeginInit()`, are byte-identical to a genuinely unqualified name the table could not resolve. **Measured: 1,603 of 41,240 refs.**

The instruction in this PR body — *"`Ref.Qualifier != ""` means `Kind` is deliberately unknown … the epic's confidence signal"* — **is wrong for those 1,603 sites**, and an S5 author following it would resolve `.Foo(` against file-local declarations. That is the confidently-wrong edge the epic exists to prevent. Nothing escapes today only because `StatementHead` is false for all of them: luck of composition, not a guard. **That instruction is mine, and it is the most consequential thing this review caught.**

Fix is an explicit `Qualified bool` so the distinction survives to S5. Related and undocumented: **every `With`-block member invocation is silently dropped** — `.SetValue(...)` as a whole statement is a real call with `IsCall() == false`. Defensible as a recall limit; not defensible as an undocumented one, given `With` is pervasive in the WinForms code #6321 is about.

### The clean-but-wrong class, found and sized

The multi-line-literal limitation touches **nine** files, not two. Seven parse **clean** — the literal interior is read as code with no diagnostic. Measured damage: **0 phantom nodes, 5 spurious refs, 0 of them `IsCall()`**. `ApplicationSettings.vb` survives only because none of its 47 literal lines happens to read as an `End`.

The two diagnosed files do share one cause and are diagnosed rather than mis-parsed — that half of the claim holds. But "the residual is two files" is #6361's shape, in a PR opened the same day that rule was written.

### Three more, all corrections to claims rather than to code

- **`parenKeywords`' stated justification is false.** It says they are *"keywords, so no declaration can shadow them"*; many are contextual keywords and legal identifiers, and the corpus declares them — 5 `Stop`, 3 `Skip`, 5 `Equals`, 8 properties named `Property`, 6 named `Error`. Measured cost of the whole set: **7 refs, 1 of them a call.** A wrong reason, not a bug.
- **The corpus gate has a third hole beyond the two in the body.** `corpusCleanFloor` is a rate with no assertion on `len(files)`, so `299/302 = 0.9901` already passes — **one file of free slack today** — and the report is `t.Log`, invisible without `-v`. The slack grows with the corpus: 200 more clean files buys 5 tolerated failures instead of 3, so the gate gets *weaker* as the corpus grows.
- **The agreement invariant is one-directional and kind-blind**, in the direction S5 emits from. Tree → table is unchecked: **1,755 of 18,505 tree nodes** have no table symbol at the same path (1,751 accessors, expected; 4 `Operator =`/`<>`). And `inTree[path]` is a bool, so **16 paths where the table says `type` and the tree says `delegate`** satisfy it. It is not vacuous — the `NodeStructure`/`IsType()` mutant kills it — but it is weaker than this body claimed.

Also: `strip.go:27-33` still says *"UNVERIFIED against real VB.NET source. No .vb file exists anywhere on the machine this was written on."* This PR is its refutation.

All sent back. The parse is honest; the API it hands S5 is not yet.

cajasmota:
## Round 3 — every number that ships in a durable comment was re-measured, and the audit found two more

`9992d0b68`. Round 2's review found the code sound and could not break it; the three findings were all numbers in source comments that did not survive re-measurement. All three are fixed, and auditing the *rest* of the comments found two more nobody had flagged.

### The three from review

**`parenrefs.go`** — the table reproduces exactly against a baseline of 41,748 use sites / 8,702 `IsCall`:

| deleted | extra use sites | extra `IsCall` |
|---|---|---|
| the five named (`and or not in is`) | +150 | 0 |
| the 11 operator entries | +299 | 0 |
| the 31 control-flow entries | +1,329 | **+264** |
| both | +1,628 | +264 |

154 corresponds to nothing. More importantly, *"all of them phantoms"* was **false for the half the sentence called load-bearing**: the control-flow half surfaces 264 sites `IsCall()` would promote to CALLS. The comment now says the two halves fail *differently*, which is the actual reason the set exists. `custom` dropped — measured, it is not in the map.

**`parse.go`** — the shipping figure was `1,600 of 41,242`, i.e. the wrong number in the field doc S5 is told to trust. Now **1,624 of 41,748, 0 satisfying `IsCall`**, with the counting method beside it.

**The interpolation-hole explanation is withdrawn.** `internal/vbnet` was materialised at the parent commit `19969d2b7` as a standalone module and re-counted: **41,748 / 8,702 / 300 clean — identical**. Round 2 changed nothing about the ref population and could not have. It was a round-1 harness artifact, and a mechanism that does not exist should not be recorded.

The self-inconsistency in the O1 measurement was real, and splitting it changes the number: **across the 7 clean files, 5 use sites, 0 `IsCall`.** The 20/1 figure was the whole 9-file window, and that 1 is real code in a *diagnosed* file. Both now appear, labelled — they answer different questions.

### Two more the audit found

- **"12 of the 302 files carry a BOM"** was neither the carrying count nor the delta. Measured: **232 of 302 carry one** (55 `.Designer.vb`), and deleting the strip costs **22 files** their clean parse (300 → 278). Both stated, because they answer different questions. I verified the 232/55 independently — and got it wrong twice first, once to `od` spacing and once to shell word-splitting on the corpus's paths-with-spaces. The `-print0` count agrees with the agent exactly.
- **"163 of the 302 files' 175 End-mismatch diagnostics"** conflated files with diagnostics, and 175 matched nothing. Measured: deleting `pushLambda` gives **255/302 (45 files)**, producing 155 `End sub`, 8 `End function`, 7 unclosed accessors/methods.

`strip.go`'s four-defect list is now priced uniformly by parse-rate delta: BOM 22, interpolation **0**, `Custom` 3, lambdas 45. **Keeping the 0 visible matters** — inverted masking corrupts silently across 106 of 302 files rather than failing anywhere, so parse rate is the wrong instrument for that defect and the comment now says so.

### Also extended rather than merely documented

Round 2's review noted that tree → table checked presence only, never kind. Rather than document the gap, `anyNodeKindAgrees` closes it. Measured before enforcing: 16,745 declarations, **0 mismatches**, so no exception list was needed. New mutant **M16** (emit a property as `NodeField`) dies with `tree says field, table says map[property:true]`.

M12, M15, M16 and both denominator mutants re-run against the changed tree and still die.

### Standing

Three review rounds. Round 1 found the dropped `qualified` bit — the one defect that would have poisoned S5, and my own instruction in this body was the thing propagating it. Round 2 found the code sound and the numbers wrong. Round 3 fixed the numbers and found two more by auditing the ones nobody asked about.

`AGENTS.md` gained `## Derive, don't list` on `main` mid-way through this PR (#6379). Its rule — *write how a number was measured next to the number, or don't write it* — is what rounds 2 and 3 were enforcing, and this PR is the first thing measured against it.

PR #6383, opened 2026-08-19
feat(#6327): S5 — the VB.NET extractor and its edges, measured on 302 real files

cajasmota:
S5 of the VB.NET epic. Refs #6327, #6321, #6362, #6363, #6381.

`internal/extractors/vbnet/` did not exist — S4 shipped the parser and nothing consumed it — so this includes the entity projection as well as the edges. It also unwinds S2's `languagesAwaitingExtractor`, because the moment an extractor registers for `vbnet` four protocol tests fail by design.

## Measured — 302 real `.vb` files, 13,479 records

```
CALLS       total=5059   resolved=3386   unresolved=1673
CONTAINS    total=13205  resolved=13205  unresolved=0
EXTENDS     total=289    resolved=153    unresolved=136
IMPLEMENTS  total=72     resolved=15     unresolved=57
IMPORTS     total=763    resolved=1      unresolved=762
dispositions: resolved 17523 / external-known 969 / bug-extractor 1617 / bug-resolver 42
bug rate 0.0823
```

Endpoints, not counts:

```
EXTENDS FROM SCOPE.Component/class "Wake.repeatingMachine" @WakeOnLAN/Wake.vb:122
        TO   SCOPE.Component/class "Machine" @WakeOnLAN/Machines/Machines.vb:28
```

Every unresolved EXTENDS/IMPLEMENTS target is a .NET Framework type except two, both pinned by name: `Profile` (two unrelated declarations in different repos) and `TaskDialogBaseForm` (a **partial class** split across `.vb`/`.Designer.vb` — S7's problem).

**The prediction I gave the agent was wrong, and that is the useful part.** I told it to expect ~36 of 37 inheritance targets to be external, from an earlier measurement on the reporter's tree. Measured here: **153 of 289 EXTENDS bind in-tree**, because StaxRip and display-drivers-uninstaller carry their own base hierarchies on top of WinForms. I framed it as a falsifiable check rather than a target, and it falsified.

## The mutant that matters most

Anchoring hierarchy edges to the file instead of the type — the #6295/#6298 defect, now five languages deep — **improves the naive numbers**:

```
resolved 17,523 → 17,884      bug rate 0.0823 → 0.0809
```

A reviewer watching aggregate metrics would read that as a small win. It is 361 edges landing on the wrong node. That is the entire argument for asserting on FROM/TO endpoints rather than on counts, and it is why #6369 exists.

I verified this mutant myself, and it took three attempts: the first was a syntax error and the second referenced an out-of-scope identifier — both compile failures, which read as kills and prove nothing. The third compiles and dies on `EXTENDS … carries FromID="MainForm"; it must be EMPTY so assembly stamps the owning TYPE's id`.

## Other mutants, all killed

- intrinsic → CALLS: `intrinsics must not produce CALLS`; corpus +1,123 CALLS, bug rate 0.1288.
- `Qualified && Qualifier==""` resolved locally: +1,013 CALLS **of which 174 resolve** — 174 confidently-wrong edges, exactly the class S4's `Qualified` bit exists to prevent.
- drop the `classifyDispositionLang` arm: 969 endpoints move to bug-extractor, 0.0823 → 0.1304.
- dead / missing allowlist entry: both directions fire.

## Withdrawn by its own A/B

The agent added `.vb` to `sourceFileExtensions`, measured a **0-endpoint delta**, and removed it — the IMPORTS FromID always resolves because the file carrier carries that path as its `Name`, so `looksLikeSourceFilePath` is never reached. #6329's lesson applied without being asked.

## Handoff

**S7 owns:** partial-class merge (`.Designer.vb`), `Handles`, `AddressOf`, With-block receivers, method-level `Implements IFoo.Bar`.

`languagesAwaitingExtractor` and `unsupportedTrackingIssues` are now empty; a seeded positive test keeps the two lookups that lost their only real case from going vacuous.

S4's five handoff points all held exactly. Two additions: `Node.Inherits`/`Implements` are `[]string` with no positions, so hierarchy edges carry the type's declaration line; and `IsCall()`'s composition holds on the corpus, but the `Qualified && Qualifier==""` refusal is asserted separately and its mutant shows it is load-bearing on its own.

All touched suites green; `go build ./...` and `go vet` clean.


cajasmota:
## `test` is failing, and it is this PR's own blast radius — not the registry commit

Two failures, both on ubuntu and windows, both present on `cb73d32ec` and `97037b734` as well as `066eb9a89`. **They predate the coverage-docs commit**, so the registry work did not cause them.

```
--- FAIL: TestIndexRecordsUnsupportedExtensions
    unsupported_sidecar_6338_test.go:74: sidecar unsupported_extensions:
         got  map[.pas:1]
         want map[.pas:1 .vb:3]

--- FAIL: TestBucketByLanguage_TalliesUnsupportedExtensions
    unsupported_6338_test.go:42: unsupported tally:
         got  map[.pas:1]
         want map[.pas:1 .vb:3]
```

Both #6338 tests use `.vb` files as their worked example of an extension grafel does **not** support. This PR registers a VB.NET extractor, so `.vb` is now supported and correctly stops being tallied as unsupported. The tests are asserting the old world.

The failure is therefore *correct behaviour reported as a regression* — the assertions need to move, not the code.

## The enumeration was short, again

The S5 handoff said:

> It also unwinds S2's `languagesAwaitingExtractor`, because the moment an extractor registers for `vbnet` four protocol tests fail by design.

Four were found and fixed. **There were six.** The two above use `.vb` incidentally — as fixture data for a test about something else entirely — so they do not mention `vbnet`, `languagesAwaitingExtractor`, or `unsupportedTrackingIssues`, and a search for those terms cannot surface them. They are reachable only by asking "what breaks when `.vb` becomes supported?" and deriving the set from `.vb` itself.

This is the same failure class as #6361, one merge after it became a written rule in `AGENTS.md` (`## Derive, don't list`): a claim of totality that holds only for the subset the author searched. Logged here because the rule's value is measured by whether it catches this, and it did not.

## The fix is not "delete `.vb` from the expectation"

Dropping `.vb` leaves `map[.pas:1]` — a single extension with a count of 1. The tests exist to prove the tally aggregates *multiple* extensions with *counts above one*, and `.vb:3` was carrying both of those properties on its own. Deleting it would keep the tests green while quietly removing what they test.

The fixtures should instead swap `.vb` for an extension that is still genuinely unsupported, preserving a 3-count and a two-extension map — and that replacement has to be verified against the live supported-extension set rather than assumed, or this recurs the next time a language ships.

PR #6388, opened 2026-08-19
fix(#6352): coverage validate ran in one path-filtered workflow, so 18,007 of 23,450 cites could break on a PR that never ran it

### LINKED ISSUE #6352: coverage validate runs in one filtered workflow, so 18,007 of 23,450 cites can break on a PR that never runs it
Found while reviewing #6351 (the coverage-matrix gate). That PR fixes one instance of this shape; the numbers say the remainder is four times larger.

## The exposure

`go run ./tools/coverage validate` checks that every `cites` path in `docs/coverage/registry.json` still exists. It runs in **`coverage-docs.yml` and nowhere else** — grepped across all 14 workflows. That workflow triggers only on `docs/coverage/**`, `tools/coverage/**`, and (after #6351) `internal/extractors/**`.

There are **23,450 cites spanning 22 top-level directories**:

| directory | cites | share | triggers a run? |
|---|---:|---:|---|
| `internal/substrate` | 6,404 | 27.3% | no |
| `internal/custom` | 4,272 | 18.2% | no |
| `internal/links` | 4,173 | 17.8% | no |
| **`internal/extractors`** | **3,449** | **14.7%** | **yes (#6351)** |
| `internal/engine` | 3,158 | 13.5% | no |
| `internal/mcp` | 813 | 3.5% | no |
| 16 others | 1,181 | 5.0% | no |

**#6351 closes 14.7%. The remaining 18,007 cites — 76.8% — sit behind paths that trigger no run.**

## Why it matters

Rename or delete a cited file under `internal/substrate/` and: `validate` breaks, the PR merges green because nothing ran, and the failure surfaces later as an unrelated red on whatever unrelated PR next touches `docs/coverage/`. The person who broke it is gone; the person who sees it did not cause it.

That is the same failure shape as #6332 — a gate that cannot fire on the change that breaks it — at four times the scale.

## Why a longer path list is the wrong fix

Adding 22 directories to the filter would make the filter approximate "the whole repository", at which point it is not a filter, it is a slower way of running always. It also rots: cite coverage moves as the codebase moves, and nothing would tell us when the list falls behind. #6351 deleted three dead classification entries and #6330 deleted 31 dead config files, both of which were hand-maintained lists that drifted.

## Options

1. **Move the `validate` step into the main test job.** It becomes an ordinary gate that always runs. Cost: whatever `validate` takes on every PR (it is a JSON scan plus stat calls — measure before assuming).
2. **Drop the path filter from `coverage-docs.yml` entirely**, so the whole job runs on every PR. Simpler; costs the `gen` + diff work too, which is heavier than `validate` alone.
3. **Split the workflow**: `validate` unfiltered, `gen`+diff filtered. Keeps the expensive half cheap and makes the cheap half trustworthy.

My inclination is 3, subject to measuring what `validate` actually costs — it is the only option that makes the guarantee unconditional without paying for `gen` on every PR. But 1 is nearly as good and simpler, and if `validate` turns out to cost seconds, simpler wins.

CI is free on this repo (public, unlimited standard runners), so the constraint is wall-clock on the PR, not minutes.

## Acceptance

Whatever is chosen: delete a cited file under a directory that is **not** in any current filter, and confirm CI goes red on that PR rather than on the next unrelated one.

Refs #6332, #6330, #6347, #6351

cajasmota:
Fixes #6352.

`go run ./tools/coverage validate` checks that all **23,450** `cites` paths in `registry.json` still exist. It ran in `coverage-docs.yml` and nowhere else, and that workflow is path-filtered — so **18,007 cites (76.8%)** sat behind paths that trigger no run. Delete a cited file under `internal/substrate/` and the PR merges green; the breakage surfaces later as an unrelated red on someone else's PR, long after the author has moved on.

## Measured before choosing

| step | duration |
|---|---|
| **Validate schema + cites** | **11.2s** |
| backfill --check | 0.29s |
| fmt --check | 0.32s |
| gen | 0.58s |
| docs diff | 0.15s |

From run 32282703633 on `main`. The 11.2s is almost entirely the one-off `go run` compile plus printing 10,171 pre-existing advisory warnings; the scan itself is sub-second, as the 0.29s warm-cache `backfill` step shows.

The issue offered three options and a rule: prefer the split, **unless `validate` costs only seconds, in which case simpler wins**. It costs seconds. So this is **option 1** — the step moves into `test.yml`'s `test` job, which already has an unfiltered `pull_request:` trigger, guarded `if: runner.os == 'Linux'` so it runs once rather than per-OS, beside the existing Linux-only `gofmt` / `go vet` steps. No new job, no new trigger, and **no longer path list** — the issue is explicit that a 22-directory filter is not a filter, it is a slower way of running always, and it rots.

## Copied, not moved — and that distinction is load-bearing

`coverage-docs.yml` also runs on `push: main`, a leg `test.yml` does not subscribe to. Deleting the step there would have silently dropped the post-merge validate — **the same class of bug this PR fixes**. So the step stays in both places. Cost of the duplication is ~11s on the minority of runs that clear the path filter.

Nothing else moved: `backfill --check`, `fmt --check`, `gen`, and the docs diff all stay path-filtered in `coverage-docs.yml`.

## Acceptance — verified in both directions

The issue's acceptance criterion is that deleting a cited file under a directory in no current filter must go red **on that PR**. I ran it as a two-sided experiment, because a one-sided pass would not distinguish "the fix works" from "the gate was always firing":

- **With this branch** + a deleted cited file under `internal/substrate/` → `validate` runs and fails on the PR.
- **With `main`** + the same deletion → `validate` never runs; the PR is green.

Results linked in a comment below. Both scratch branches were deleted without merging.

### DIFF
--- a/.github/workflows/coverage-docs.yml
+++ b/.github/workflows/coverage-docs.yml
@@ -29,6 +29,15 @@ jobs:
       - uses: actions/setup-go@v5
         with:
           go-version: '1.26'
+      # The AUTHORITATIVE, unconditional copy of this check now lives in
+      # test.yml, which has no path filter on `pull_request` — see #6352 for
+      # why the filter below could never fire on 76.8% of the cited paths.
+      # This copy is deliberately RETAINED rather than moved, because it is not
+      # redundant: this workflow also runs on `push: main`, a leg test.yml does
+      # not subscribe to (test.yml reaches `push` only via release.yml's
+      # workflow_call on a v* tag). Deleting it here would silently drop the
+      # post-merge validate. It costs ~11s on the minority of runs that clear
+      # the path filter.
       - name: Validate schema + cites
         run: go run ./tools/coverage validate
       - name: Guard against incomplete grouped records (#2971)
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -237,6 +237,36 @@ jobs:
             exit 1
           fi
 
+      # Coverage-registry cite check (#6352). `go run ./tools/coverage validate`
+      # confirms every `cites` path in docs/coverage/registry.json still exists.
+      # It used to live ONLY in coverage-docs.yml, which is path-filtered to
+      # docs/coverage/**, tools/coverage/** and internal/extractors/**. But the
+      # 23,450 cites span 22 top-level directories: only internal/extractors
+      # (14.7%) is behind a filter path, so 18,007 cites — 76.8%, including all
+      # of internal/substrate, internal/custom, internal/links and
+      # internal/engine — sat behind paths that triggered no run. Deleting or
+      # renaming a cited file under any of those merged green, and the breakage
+      # then surfaced as an unrelated red on whoever next touched
+      # docs/coverage/ (the #6332 failure shape, at four times the scale).
+      #
+      # A longer path list was rejected in #6352: 22 directories is an
+      # approximation of "the whole repo" that rots silently — #6351 deleted
+      # three dead entries and #6330 deleted 31 dead files from exactly this
+      # kind of hand-maintained list. Running it here instead makes the
+      # guarantee unconditional, because this workflow has NO path filter on
+      # `pull_request` (#6291), so it fires on the change that breaks it.
+      #
+      # It is cheap: measured at 11.2s on run 32282703633, and nearly all of
+      # that is the one-off `go run` compile plus printing ~10k pre-existing
+      # advisory warnings — the scan itself is sub-second. That is why #6352's
+      # "if validate costs seconds, the simpler option wins" clause applies.
+      #
+      # Linux only: it is a JSON scan plus file-existence stats, so it is
+      # platform-independent — same rationale as the gofmt/vet steps above.
+      - name: Validate coverage registry schema + cites (#6352)
+        if: runner.os == 'Linux'
+        run: go run ./tools/coverage validate
+
       - run: go build ./...
 
       # Purge BOTH the test-result cache AND the compiled-object cache.


cajasmota:
## Acceptance: run as a two-sided experiment, and both sides landed

A one-sided pass cannot distinguish *"the fix works"* from *"the gate was firing all along"*, and #6352's whole claim is that a PR breaking a cite **merges green today**. So both arms were run: the same change, with and without this branch.

The change is a **rename**, not a deletion — `internal/substrate/crystal.go` → `crystal_acceptance_renamed.go`. `internal/substrate` is 27.3% of all cites and sits in no path filter. A rename leaves the package compiling and breaks only the cite, which isolates the variable; deleting a `.go` file would have failed the job on compilation for an unrelated reason.

### Arm A — WITH this branch (#6389)

```
success   go vet ./...
success   gofmt
failure   Validate coverage registry schema + cites (#6352)
skipped   Run go build ./...
skipped   Test
```

Everything before the gate passes, the gate fails, everything after is skipped. **The gate fires on the change that breaks it.**

### Arm B — control, on `main` WITHOUT this branch (#6390)

```
success   go vet ./...
success   gofmt
success   Run go build ./...
success   Test
```

The validate step **is not in the job at all** — `gofmt` runs straight into `go build`. The identical rename, with a dangling cite in `registry.json`, and the job is **fully green**.

`coverage-docs.yml` triggered on **neither** arm, which is the point: the rename is outside its `docs/coverage/** | tools/coverage/** | internal/extractors/**` filter, so today nothing validates the cites on such a PR at all.

That is #6352's reported failure, reproduced rather than asserted: green PR, broken cite, and the red lands later on whoever next touches `docs/coverage/`.

### What this does and does not establish

**Established:** the gate now fires for `internal/substrate` (27.3%), and demonstrably did not before.

**Not established:** I verified one directory, not all 22. The mechanism is a path filter versus no path filter, so it should hold for every non-filtered directory — but that is an inference from how the trigger works, not 22 measurements. Recording it as such rather than claiming coverage I did not measure.

Both scratch branches (#6389, #6390) are deleted; neither was merged.

PR #6390, opened 2026-08-19
SCRATCH: #6352 acceptance arm B (control, no fix) — DO NOT MERGE

cajasmota:
SCRATCH — DO NOT MERGE. Acceptance arm B (control) for #6352 (PR #6388).

The SAME rename as arm A, but based on `main` WITHOUT the fix.

**Expected: the cite gate never runs, and this is GREEN despite a dangling cite** — which is precisely the bug #6352 reports.

Deleted immediately after the result is recorded. Refs #6352.


cajasmota:
Acceptance arm B (control) complete — fully green with a dangling cite, which is #6352's reported bug reproduced. Results recorded on #6388. Scratch branch deleted, never merged.

Issue #6387, opened 2026-08-19
the Strawberry fixture's include_router(prefix=) implies coverage that does not exist — the asserted path is hardcoded


cajasmota:
## What

`internal/engine/http_endpoint_synthesis_strawberry_3066_test.go:239` contains

```python
app.include_router(graphql_app, prefix="/graphql")
```

and asserts the endpoint `http:GRAPHQL:/graphql/Query/hello`.

The fixture reads as coverage of `include_router(prefix=)` folding. It is not. The Strawberry synthesizer hardcodes the path at `internal/engine/http_endpoint_synthesis.go:4638`:

```go
path := "/graphql/" + rootType + "/" + methodName
```

The `prefix=` argument is never parsed by any synthesizer — `include_router` appears in **zero** non-test Go files under `internal/` and `cmd/`, which I verified directly. So the `/graphql` in the assertion comes from the hardcoded literal, not from the fixture's prefix, and the test should pass with the prefix set to any other value — or removed.

## Why this matters more than one test

A fixture that is satisfiable without exercising the behaviour it appears to cover is worse than no fixture: it reports coverage that does not exist, and it makes the next person believe a code path is protected when nothing touches it. This project has been bitten by exactly this shape before — a corpus fixture that scored 25/25 against a content-free sentinel (#6277), and two fixtures with no `expected.json` that were never graded at all (#6273).

The concrete cost here: someone implementing `include_router` prefix folding would search for existing coverage, find this fixture, and reasonably conclude the mount-site form was already handled.

## What to do

1. **Confirm the vacuity empirically before changing anything** — change the fixture's `prefix=` to a different literal and confirm the assertion still passes. I derived this from reading the synthesizer, not from running it, and it should be a measurement rather than an inference.
2. Make the fixture honest. Either drop the misleading `prefix=` argument so it stops implying coverage it does not provide, or — if kept for realism — add a comment at the fixture stating the prefix is *not* what produces the asserted path.
3. Do **not** attempt to implement prefix folding here. That is #6385 and it is a cross-file problem.

The test's actual subject — that Strawberry emits `GRAPHQL` endpoints per root-type method — is worth keeping. Only the part that implies prefix handling is at fault.

## Credit

Found by @arthurgeron while investigating #6385, in code they had no reason to be auditing. They spotted that their own apparent counterexample was not one, and said so, which is what surfaced this.

Refs #6385

Issue #6277, opened 2026-08-08
The quality benchmark matches must-haves on name+kind only, so a placeholder can satisfy an expectation and score green


cajasmota:
The golden benchmark decides whether a change improved or regressed extraction by comparing **counts**. `scripts/quality/ratchet.py`'s own docstring already lists this as a known limitation. #6275 shows what it costs.

## The concrete case

`java-spring-mini` reports `entity_found 25/25` — a perfect score. The `SCOPE.Component User` satisfying that expectation is **not** the Java class node. It is ormlink's sentinel, documented at `internal/extractors/cross/ormlink/extractor.go:46-49` as "intentionally low-quality so it doesn't compete with the real class entity". It drops from `start_line 7` to `start_line 0` and holds no members; all six `User.*` CONTAINS edges migrated onto a different node (#6275).

So for a consumer asking "does `User` exist?" the answer is yes, and for "what does it contain?" the answer is nothing. **That is worse than a missing node**, because a missing node is visible.

The benchmark cannot see it. 25 expected, 25 found, green.

## Why this outranks any single extraction bug

This is the instrument we use to judge every other change to extraction. #6260 is being merged partly on the strength of "+20 must-have entities" — a number produced by this same counting. If a count can be satisfied by a placeholder, then every recall figure quoted anywhere in this repo carries an unknown false-positive rate, and no amount of care in the individual extractor fixes it.

It also compounds with #6273 (two fixtures never graded at all): one failure mode hides fixtures, the other hides substitutions within a graded fixture. Both present as green.

## What a fix needs

The expectation format already carries more than name and kind — `expected.json` entries have `source_file`, and #6275 shows `start_line` is the discriminator that would have caught this. Matching is currently exact on `Kind\x00Name` only (`internal/quality/diff.go:107-127`).

Options, none obviously right:

- **Match on identity, not name+kind** — include `source_file`, and optionally a line-presence check. Cheapest, and would have caught this case.
- **Reject known-sentinel subtypes from satisfying a must-have.** Note `grep -n 'SubtypeSentinel\|orm_model_sentinel' cmd/grafel/index.go internal/extractors/*.go` currently returns nothing, so no dedup rule knows that subtype exists — the sentinel's own doc comment claiming it "doesn't compete" is not enforced anywhere. That is arguably the deeper bug.
- **Assert structural expectations**, e.g. that a must-have entity claimed to contain members actually has edges. Most thorough, most work.

[...]

Issue #6273, opened 2026-08-08
Two golden fixtures have no expected.json, so they are never graded — and a never-graded fixture is indistinguishable from a passing one


cajasmota:
`internal/quality/golden/groovy-grails-mini/` and `internal/quality/golden/swift-swiftui-mini/` each contain a `src/` tree and **no `expected.json`**. Verified: 20 fixture directories, 18 `expected.json`.

`scripts/quality/run.sh:109` iterates `internal/quality/golden/*/` unconditionally. These two produce no report, and strict mode at `:243` only bumps `EXIT=2` on a fixture that *ran and missed*. So they are **never graded, and a never-graded fixture is indistinguishable from a passing one** at every surface that consumes the benchmark.

## Why this outranks the recall gaps themselves

This is the repo's worst-ranked failure shape — silent wrongness, not loud breakage. A fixture that is never evaluated looks exactly like a fixture with nothing wrong.

It also means **every denominator quoted from this benchmark is over 18 fixtures, not 20**, including numbers I have written into issue threads myself:

- #6231's "14 of 20 golden fixtures red" — the "of 20" is wrong, and my phrasing there ("2 don't index") was imprecise in a way that hid this: they *do* index, they are simply never evaluated.
- #6260's "22 of 26 missing must-have entities" and the 229/258 must-have figure — both computed across the 18 that produce reports.

None of those conclusions flip, but the framing overstates coverage, and anyone using these numbers as leverage is quoting a benchmark that silently skips 10% of its corpus.

## What a fix has to decide

Writing the two missing `expected.json` files is the obvious half. The other half is the actual defect: **the runner must not treat "produced no report" as "fine"**.

`run.sh` should fail, or at minimum report a distinct non-zero status, for a fixture directory that yields no report — otherwise the next fixture added without expectations repeats this silently. Whether that is a hard failure or a named "ungraded" category is a small policy call; either is better than the current silence.

Note `baseline.json` does record `expectations_missing: true` for both, so the information exists — nothing acts on it.

Refs #6231, #6260

PR #6411, opened 2026-08-20
fix(#6377): defer the participation gate to history when a kind has any, so terminal-by-design kinds stop firing every run

### LINKED ISSUE #6377: terminal-by-design kinds fail kind-carries-semantic-edges on every run forever, at a ~14% true-positive rate
#6346 direction 2 — "terminal-by-design kinds fail it, correctly, forever" — was **relocated by #6346's fix, not solved**. The old `orphan-rate-not-100pct[kind]` check fired on kinds with zero semantic participation; the new `kind-carries-semantic-edges[kind]` fires on exactly the same kinds. Same repos, same kinds, new name and a more honest note.

## Measured

12 corpus repos, **per group** — not aggregated. That distinction is load-bearing: merging the repos into one synthetic group lets a kind borrow participation from another repo (`Route` aggregates to 93.6%, `Test` 91.9%, `Config` 90.6%, all non-zero), which hid six of these kinds and produced a wrong claim in an earlier revision of #6375. The report runs per group, so the per-group count is the only one that means anything.

**14 firings over 10 distinct kinds** — `Config`, `Fixture`, `Route`, `SCOPE.CodeBlock`, `SCOPE.Config`, `SCOPE.Constraint`, `SCOPE.Stylesheet`, `SCOPE.UIComponent`, `Service`, `Test`.

Triaged by sampling the underlying sources:

- **~8 unarguably terminal by design** — markdown code fences, CSS selectors (`a.edit::before`), HTML `<input>` fields, `requirements.txt` / `pyproject.toml` / `.eslintrc.json`. These will fail every run forever, with no action available to the reader.
- **2 unambiguous true positives** — `Route` in two Flask groups: `@app.route` paths with `sem=0 struct=1`, never bound to the function directly beneath them.
- **~4 arguable** — `Test`, `Fixture`, `Service`, `Config`.

**True-positive rate ≈ 14%.** A repo is permanently penalised for containing CSS or markdown. That is the "trains the reader to ignore the whole section" failure mode direction 2 exists to name, intact.

## Why it was not fixed in #6346

Distinguishing a terminal-by-design kind from a dead resolver requires evidence the feedback package does not have. **The graph alone cannot do it** — both are zero participation, indistinguishable by construction.

Two remedies were considered and rejected there:

- A hand-maintained kind/subtype name list is exactly what #6346 ruled out, and is the #6361 failure class.
- There is no declarative edge→endpoint-kind schema to derive from. Searched `internal/types/{kinds,relationship,entity}.go` and `internal/graph/`; only free-standing string constants exist.

## The candidate that needs neither

**Longitudinal comparison**: keep prior reports and fail only when a kind that *used to* participate no longer does. That converts an unanswerable question ("is this kind supposed to have edges?") into an answerable one ("did this kind lose edges it had last time?"), and drops the false-positive class to zero for stable kinds — CSS and markdown never participated, so they never regress.

It needs report history under `~/.grafel/feedback/`, which does not exist yet. **That persistence is the actual unit of work here**, and it is worth scoping on its own: it would also make every other metric in the report trendable rather than a single-point reading.

Known limitation to design around: a kind that is broken on its *first* report has no history, so the longitudinal check is silent there — which is precisely the new-extractor case `kind-carries-semantic-edges` was added to catch. The two checks are complementary rather than substitutes, and whatever lands here should keep the 100%-end gate for first runs.

## Evidence bar

Per `AGENTS.md` `## Evidence`: report the false-positive count per group before and after, measured per group rather than aggregated, on the same 12 corpora. A total across repos will look better than the truth for the reason described above.

Refs #6346, #6375, #6313, #6361

cajasmota:
Fixes #6377.

`kind-carries-semantic-edges[kind]` fires on kinds that are **terminal by design** — markdown code fences, CSS selectors, HTML `<input>` fields, `requirements.txt` entries. Measured **per group** across 12 corpus repos: **14 firings over 10 distinct kinds, ~2 of them real.** A gate that is wrong 86% of the time on every run does not protect anything; it teaches readers to skip it.

## Read this first — what this does NOT do

**The 14 corpus firings still fire once.** All 14 are *first* reports for their group, and a first run has no history to compare against, so the unconditional gate is still the only thing that can speak.

What changes is the recurrence: **the cost is now paid once per kind instead of on every run forever.** That is a real improvement and it is smaller than "fixes the 14 firings" — stating it here because a reader who re-runs the corpus and still sees 14 would otherwise conclude the change did nothing.

## The premise this issue corrected about itself

An earlier revision claimed the required history "does not exist yet" and called building it the actual unit of work. That was wrong: `grafel feedback` already writes every report to `~/.grafel/feedback/<group>-<timestamp>.md` (`internal/cli/feedback.go:106`). So this is **read-side** work.

## How history is read

New `internal/feedback/history.go`:

- `parseKindParticipation` scans only Section 2 (`## 2. Orphan Rate`) and returns a **three-state** per-kind map — participated / observed-but-never / absent. The third state is what lets the gate tell "this kind is new" from "this kind stopped working".
- It keys on the orphan **percentage** (`< 100.0` = participated), not the table header, so **pre-#6313 reports remain usable history** — back then zero-participation kinds stayed in the main table at 100.0%.
- `loadKindParticipation` scopes by **filename prefix**, because the report body never contains the group name (only a `Group profile:` language summary).
- Participation is **OR-ed across all reports**, not taken from the latest. Comparing only against the most recent run would accept a two-runs-ago breakage as the new normal.
- A missing directory or unreadable file degrades to first-run behaviour and **never fails report generation** — history is an enhancement, not a dependency.

## The blind spot that had to be preserved

Check 2b defers to history **only when the kind was observed before**. `!observed` — a first report, or a kind a new extractor has just started emitting — falls through to the unchanged unconditional failure. That is exactly the new-extractor case `kind-carries-semantic-edges` exists to catch, and replacing the old check outright would have reintroduced it.

Regressions get their own distinct check, `kind-participation-not-regressed[kind]`, rather than overloading the existing name.

## Anonymisation verified, not assumed

Kind names survive: `NameHash` uses the kind string only to select a prefix via `kindFamily` and never emits it; `PathScrub` only ever sees file paths; and the real report at `~/.grafel/feedback/event-driven-ai-20260716T065826.md` shows `SCOPE.Stylesheet` and `http_endpoint_definition` verbatim beside hashed identifiers. `TestKindNamesSurviveAnonymisation` round-trips a rendered Report back through the parser, so a future scrubbing change fails loudly instead of silently breaking the comparison.

## Verification

`go test ./internal/feedback/` → 0, `go vet` on `internal/feedback` + `internal/cli` → 0, `gofmt -l` clean, narrowed `-run Feedback ./internal/cli/` → 0.

Mutant, compiled: `if everParticipated` → `if !everParticipated`. Killed **both** directional tests — `KindThatStoppedParticipatingFails` and `TerminalByDesignKindWithHistoryDoesNotFail`. Restored by `cp`, `shasum db805161…` matching.

All tests use temp directories; the real `~/.grafel/` was read-only and never written.

### DIFF
--- a/internal/cli/feedback.go
+++ b/internal/cli/feedback.go
@@ -97,17 +97,24 @@ func runFeedback(cmd *cobra.Command, groupName, outPath string, yes bool) error 
 	}
 
 	// 3. Resolve output path.
+	//
+	// historyDir is resolved independently of outPath: prior reports always
+	// live in <home>/feedback even when this run is written elsewhere via
+	// --out, and the longitudinal participation check (#6377) reads them from
+	// there. A home dir that cannot be resolved only disables that check.
+	var historyDir string
+	if home, err := registry.HomeDir(); err == nil {
+		historyDir = filepath.Join(home, "feedback")
+	}
 	if outPath == "" {
 		ts := time.Now().UTC().Format("20060102T150405")
-		home, err := registry.HomeDir()
-		if err != nil {
-			return fmt.Errorf("feedback: resolve home dir: %w", err)
+		if historyDir == "" {
+			return fmt.Errorf("feedback: resolve home dir")
 		}
-		outDir := filepath.Join(home, "feedback")
-		if err := os.MkdirAll(outDir, 0o755); err != nil {
+		if err := os.MkdirAll(historyDir, 0o755); err != nil {
 			return fmt.Errorf("feedback: create output dir: %w", err)
 		}
-		outPath = filepath.Join(outDir, groupName+"-"+ts+".md")
+		outPath = filepath.Join(historyDir, groupName+"-"+ts+".md")
 	}
 
 	// 4. Show confirmation prompt (skipped with --yes).
@@ -159,8 +166,9 @@ func runFeedback(cmd *cobra.Command, groupName, outPath string, yes bool) error 
 	// 6. Generate report.
 	fmt.Fprintln(w, "Generating report...")
 	report, err := feedback.Generate(context.Background(), docs, feedback.Opts{
-		GroupName: groupName,
-		Version:   version.String(),
+		GroupName:  groupName,
+		Version:    version.String(),
+		HistoryDir: historyDir,
 	})
 	if err != nil {
 		return fmt.Errorf("feedback: generate: %w", err)
--- a/internal/feedback/history.go
+++ b/internal/feedback/history.go
@@ -29,20 +29,67 @@ import (
 //
 //   - The group name appears only in the FILENAME. The report body carries a
 //     "Group profile:" line with language counts, never the group name (see
-//     Render). So history is scoped by filename prefix.
+//     Render). So history is scoped by an EXACT match on the filename stem
+//     before the timestamp — not a prefix: group "foo" does not read
+//     "foo-bar-<ts>.md".
 //
 //   - Kind names survive anonymisation VERBATIM. NameHash uses the kind only
 //     to select a short prefix and never emits it; PathScrub only ever sees
 //     file paths. Kind is the join key of this whole comparison, so
 //     TestKindNamesSurviveAnonymisation asserts the round-trip through Render
 //     rather than trusting that to stay true.
+//
+// ONE-WAY DOOR, stated because nothing else documents it: participation is
+// OR-ed across every stored report and never expires (loadKindParticipation).
+// That is right for the common case — a kind that broke two runs ago must not
+// launder its breakage into the new normal — but it means a kind that
+// legitimately and PERMANENTLY stops participating (its repo left the group,
+// its extractor was retired, the language was dropped) raises
+// `kind-participation-not-regressed[kind]` on every run from then on, with no
+// code change able to clear it. The only remedy is to delete this group's
+// stored reports, `~/.grafel/feedback/<group>-*.md`, which resets the group to
+// first-run behaviour. The check's own note says so, so the reader never has
+// to find this comment.
 
-// section2Heading and the row pattern below are matched against the rendered
+// section2Heading and the row patterns below are matched against the rendered
 // report, which is the only durable artifact — there is no machine-readable
 // sidecar. Parsing is deliberately confined to Section 2 so rows from the
 // kind×language table in Section 1 or the disposition table in Section 3
 // cannot be mistaken for kinds.
 const section2Heading = "## 2. Orphan Rate"
+
+// directionAwareMarker is the Section-2 preamble Render has emitted since
+// #6375 (c6e6e148c, 2026-08-20). It is the version gate for history.
+//
+// Before that commit an entity was orphan when it had no OUTGOING semantic
+// edge; today it is orphan only when it has no semantic edge in EITHER
+// direction. That is not a cosmetic rewording — it changes what the numbers
+// in Section 2 MEAN. A pure sink (SCOPE.ExceptionType, SCOPE.Package,
+// SCOPE.Plugin, SCOPE.Config: pointed at by THROWS/IMPORTS, sourcing nothing)
+// reads 100.0% under the old metric while participating fully. Read as
+// history, such a report records those kinds as "observed, never
+// participated", which is precisely the state that SILENCES check 2b — so a
+// dead THROWS or IMPORTS resolver would raise nothing at all. Both reports
+// currently on disk are of that vintage and misclassify 8 kinds that way.
+//
+// A silent gate is worse than no history, so pre-#6375 files are ignored
+// outright and the group falls back to first-run behaviour.
+//
+// The marker is the report's own definition sentence rather than the
+// "grafel version:" line: the version line exists in both vintages but carries
+// `git describe` strings (v0.1.8-4-g61c99101b, v0.1.7.4-29-gf1bdcce5a-dirty)
+// that no total order can be recovered from, while the definition sentence is
+// exactly what changed and is verifiable in the files on disk today.
+const directionAwareMarker = "no semantic edge in EITHER direction"
+
+// defectTableHeader and terminalTableHeader identify which of the two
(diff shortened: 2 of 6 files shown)


cajasmota:
## Independent review: REQUEST-CHANGES. The back-compat claim in the body above is wrong, and it fails in the dangerous direction.

**The blocker: this silently DISABLES the gate on the only real history that exists.**

The body claims that keying on the orphan **percentage** rather than the table header keeps old reports usable. It does not, because **#6375 (`c6e6e148c`, landed 2026-08-20 — one day before this PR) changed what the percentage means.** Old reports: orphan = "no *outgoing* semantic edges." New reports: "no semantic edge in *either* direction." A pure **sink** kind reads `100.0%` under the old metric while participating perfectly — that miscount is precisely the thing #6375 was written to fix.

So the parser reads pre-#6375 history and concludes those kinds have *never participated*. Confirmed against the two real reports on disk (both pre-#6375): **8 kinds** land in observed-but-never-participated, including the textbook sinks `SCOPE.ExceptionType`, `SCOPE.Package`, `SCOPE.Plugin`, and `SCOPE.Config`. With those kinds terminal today, **neither** `kind-carries-semantic-edges[…]` **nor** the new `kind-participation-not-regressed[…]` fires.

A dead THROWS or IMPORTS resolver on that group would be **completely silent**. I asked the reviewer specifically to check whether degradation silently passes a gate, because that is worse than firing — and this is that case, reached not by a malformed file but by ordinary history written last month. The fix version-gates history and ignores pre-#6375 files.

Also correcting `sanity.go:264`, which flatly asserts "this **is** a resolver regression, not a terminal-by-design kind" with no hedge — on evidence whose definition changed underneath it.

**Two more that matter:**

**The entire feature unplugs with every test still green.** Mutating `r.priorParticipation = prior` → `nil` survives *both* `./internal/feedback/` and `-run Feedback ./internal/cli/` at exit 0. Both directional tests set `priorParticipation` on a struct literal, so the only path production actually uses — `Opts.HistoryDir` → `Generate` → `runSanityChecks` — has **zero** coverage. Green meant nothing here.

**The Section-2 scoping assertion is decorative.** Deleting `if !inSection2 { continue }` survives. The existing leak check keys on `"Disposition"`/`"Language"`/`"resolved"` rows, which are 2- and 3-column and can never match the 4-column row regex with or without scoping — so the assertion cannot fail. Needs a genuine 4-column row planted outside Section 2. Same defect class as #6273, #6277, #6387, and the one found on #6412 an hour ago.

**And a design cost I did not state:** OR-ing participation across all reports means once a kind participates, it participates forever. A kind that legitimately and permanently stops — repo dropped from the group, extractor retired, language removed — makes `kind-participation-not-regressed[kind]` fire **every run forever with no action available to the reader**. That is verbatim the failure #6377 exists to kill, recreated under a new name. The only remedy is manually deleting the group's stored reports, which was documented nowhere. The OR semantics stay — they are right for the common case — but the cost and the remedy get written down, in the check's own message.

## Corrections to the body's prose

- I wrote "filename **prefix**" scoping. It is **exact-stem** match, which is *safer* than I described: group `foo` does **not** pull `foo-bar-<timestamp>.md`. Verified. The description was wrong in the pessimistic direction, but wrong.
- I attributed the report-format change to **#6313** throughout. It is **#6346/#6375**. #6313 is unrelated.
- `%.1f` rendering means 9999 orphans out of 10000 renders `100.0%`, so a kind that genuinely participated is recorded as never-participating and a later true regression is missed.

## Verified and unaffected

Degradation on malformed input fails **closed**, as intended: a garbage file and a `0o000` file both yield an empty map with no error, an unreadable directory returns an error, and `report.go` then leaves `priorParticipation` nil so the unconditional gate runs. Kinds named `Kind`, or containing `|`, or leading `-`, degrade to *absent* → unconditional gate.

The review also disproved a risk I had flagged: the `report.go` diff is purely additive, with **no** `kindTotals` or N ≥ 10 derivation change, so no existing gate shifts as a side effect.

cajasmota:
## Coordinator mutation check on `42ef63e9a`

I ran my own mutants rather than relying on the fix report's table, and picked the **new** logic — the terminal-table authority that came in as a bonus fix, since new code is the surface nobody has attacked yet.

The uniqueness assertion earned its keep immediately: my first anchor matched **two** sites, and a blind `sed` would have mutated the wrong one and produced a meaningless result.

| Mutant | `go vet` | Test exit | Result |
|---|---|---|---|
| `out[kind] = false` at the terminal-table row (`history.go:181`) → `_ = kind` | 0 | 0 | **SURVIVED** |
| `out[kind] = false` in the terminal override loop (`history.go:194`) → `_ = kind` | 0 | 0 | **SURVIVED** |
| **both sites together** | 0 | **1** | **KILLED** — 4 tests fail |

The two sites are **mutually redundant**: the terminal-table branch sets `out[kind] = false` inline *and* records the kind in `terminal`, which the override loop then re-zeroes. Either one alone is sufficient under the current render order, so neither is individually observable.

**The behaviour is genuinely pinned** — removing both fails `TestParseKindParticipation_TerminalTableOverridesDefectRow`, `..._ReadsBothSection2Tables`, `TestLoadKindParticipation_ScopedToGroupAndOred`, and `TestKindNamesSurviveAnonymisation`. So this is defense-in-depth, not a coverage hole, and I am not treating it as a blocker.

It is worth writing down, though, because the redundancy is **order-dependent**: the inline `false` only wins because `Render` emits the terminal table after the defect table. Flip that order and the inline assignment stops protecting anything, leaving the override loop as the sole guard — which no test currently distinguishes. Anyone reordering Section 2 should re-run these two single-site mutants.

All files restored by `cp`, `shasum cd968b4d…` verified identical to the committed source. No `git checkout --` was used.

Merging once CI is green — not before, and not on green alone.

Issue #6283, opened 2026-08-08
run.sh grades whatever binary is on disk and never rebuilds, so a verification pass can confirm a result that has nothing to do with the code under test


cajasmota:
`scripts/quality/run.sh:112-116` only builds the benchmark binary if it is missing:

```bash
if [[ ! -x "$BIN" ]]; then
```

An existing binary is never rebuilt, so the runner grades **whatever is on disk**, not the working tree.

## How it surfaced

Found by the adversarial review of #6277. The reviewer ran the full benchmark at HEAD, then again with `isPlaceholderAnchor` mutated to `false` — a change that provably moves three fixtures — and got **byte-identical reports** both times. The mutant looked dead. It was not; the binary simply predated both runs. They only noticed after running `go build` by hand.

## Why this is worse than an ordinary staleness bug

This is the instrument the project uses to decide whether a change improved or regressed extraction, and it is the third way that instrument can report success without having measured anything:

- **#6273** — two fixtures had no `expected.json`, so they were never graded and looked identical to passing ones. Fixed.
- **#6277** — matching was satisfiable by a placeholder, so a fixture scored 25/25 on a content-free stand-in. In review.
- **This** — the runner grades a stale binary, so a verification pass can "confirm" a result that has nothing to do with the code under test.

The failure mode is specifically dangerous for *mutation testing*, which this repo relies on as its main defence against vacuous tests. A surviving mutant is supposed to mean "no test covers this". Here it can silently mean "you graded yesterday's binary". Every mutant verified through `run.sh` rather than `go test` is suspect until re-run.

It also silently rewards the wrong habit: the faster your loop, the more likely the binary is stale, so the more carefully you iterate the less your results mean.

## What a fix needs to decide

Rebuilding unconditionally is the obvious move and costs a `go build` per run. The question is whether that is acceptable for a script that already takes minutes, or whether a staleness check is better — compare the binary's mtime against the newest `.go` file under `cmd/` and `internal/`, and refuse (not silently rebuild) when it is older, so the operator learns why.

Refusing is probably right, by the same argument that settled #6273: a named category nobody acts on is how these gaps persist, and #6273 chose a hard failure over a quieter signal for exactly this reason.

Worth checking at the same time whether any other script or CI job grades a prebuilt binary the same way.

Refs #6273, #6277

Issue #6150, opened 2026-08-05
Incremental drops 6 of 7 Pass-2.5 framework entities per file — every http_endpoint_definition and Route on the default daemon path


cajasmota:
The incremental path (Path A, `internal/extractors.TryIncremental`) drops **six of the seven** Pass-2.5 framework entities for every re-extracted file — including every `http_endpoint_definition`, every `Route`, and the `SCOPE.Process` flow node.

Measured end-to-end, real `dvFullRebuild` vs real `TryIncremental`, one changed Falcon file:

| full rebuild (10 entities) | incremental (4) |
|---|---|
| `Config\|App()` | — |
| `Controller\|ZzThings` | `Controller\|ZzThings` |
| `Module\|zzapi_delta` | `Module\|zzapi_delta` |
| `Route\|/zzthings` | — |
| `Route\|on_get` | — |
| `SCOPE.Component\|zzapi_delta.py` | `SCOPE.Component\|zzapi_delta.py` |
| `SCOPE.Operation\|ZzThings.on_get` | `SCOPE.Operation\|ZzThings.on_get` |
| `SCOPE.Process\|http:GET:/zzthings → …` | — |
| `Service\|app` | — |
| `http_endpoint_definition\|http:GET:/zzthings` | — |

## Why it matters

Path A is the **default** path for a watched repo. So on a Flask/Falcon/FastAPI project — precisely the repos the YAML rule sets exist for — editing a route file silently destroys that file's endpoint and flow entities until the next full rebuild.

Those entities are the substrate for `grafel_traces` and `grafel_cross_links`. An MCP caller asking "what happens when a user hits this endpoint" gets a truthful-looking answer computed over a graph that quietly lost the endpoint.

## Current state after #6148

#6148 makes Path A run the framework detector and recovers exactly **one** of the seven: the class-kind fold (`Controller|ZzThings`). It pays the **entire** `Detect` cost and then hands only the class-kind slice to the fold, discarding the rest.

So the shape today is "run the classifier, keep 14% of its output, pay 100% of its cost." The `Detect` result is already computed and in hand — consuming its non-class entities is the natural completion, not new work.

## Not a regression

This predates #6148 and predates the parity gate. It has never been visible because the gate's fixture contains no framework app object and no route decorator, so the entire Pass-2.5 surface is **structurally unreachable** by it.

[...]

PR #6412, opened 2026-08-20
feat(#6327): S7a — a VB.NET class split across Foo.vb and Foo.Designer.vb is now one Component

cajasmota:
S7a of the VB.NET epic. Refs #6327.

A class split across `Foo.vb` and `Foo.Designer.vb` surfaced as **two separate Components**. That is the most common shape in real WinForms code — 88 of the 302 corpus files are `.Designer.vb` — and it is what @dcastro-imp reported.

## Mechanism, and why the alternative was rejected

An **identity rewrite inside the extractor**, not a post-hoc join.

`EntityRecord.ID` is ignored — `entityRecordToGraphEntity` derives identity unconditionally (#6150) — so the only lever is `(Kind, Name, SourceFile)`. Kind and Name already agree between halves; `SourceFile` is the sole reason they diverge. A `Partial` type in `Foo.Designer.vb` is now emitted with `SourceFile = Foo.vb` (and its CONTAINS ref keyed on that path), so **the fold that already exists on both assembly paths** collapses them. Nothing outside `internal/extractors/vbnet/` changed.

**Rejected: a cross-file grouping pass at assembly level.** The incremental path re-extracts only *changed* files, so editing `Foo.Designer.vb` alone would hand that pass a one-member group and a different canonical anchor than a full index — a full-vs-incremental disagreement about entity **identity**, which is the #6150 defect class and worse than the duplication being fixed. The anchor must be a pure function of one file's path.

## Two guards, measured rather than assumed

- **The sibling must exist on disk.** Of 88 `.Designer.vb` corpus files, only **33** have a `.vb` sibling. Blind infix-stripping would have re-anchored the other 55 — `My Project/Settings.Designer.vb`, `Resources.Designer.vb`, `Strings.<culture>.Designer.vb` — onto paths that do not exist.
- **The type must carry `Partial`.**

The rewritten half's span is **zeroed**: `Foo.Designer.vb` sorts before `Foo.vb`, so the designer record usually wins the survivor slot, and `foldDuplicateEntity` fills a span only when it is zero. Zeroing makes the merged span order-independent.

## Corpus measurement — 302 real `.vb` files

| | distinct ids | names with >1 id | duplicate rows |
|---|---:|---:|---:|
| before | 1046 | 66 | 86 |
| after | **1012** | **33** | **52** |

**34 duplicate Components removed.**

**The residual 34 is mostly *correct* separation, not misses** — `MySettings` ×5 and `Resources` ×5 in WakeOnLAN, one per `.vbproj`. Only **3 groups (5 rows)** are genuinely unmerged partials: `MainForm` across four `MainForm_*.vb`, `SetupAPI`/`SetupAPI_Inf`, and `MyApplication` across `ApplicationEvents.vb` + `Application.Designer.vb`. Those need a `.vbproj` compilation-unit model grafel does not have — documented in `partial.go`, deliberately not attempted.

That distinction matters: reading "34 remaining" as "34 failures" would be wrong.

## A wrong measurement, caught and corrected

The first corpus measurement filtered on *all* parts carrying `Partial`. **VB.NET lets one part omit it** — `Foo.vb` is often plain `Public Class Foo`. That undercount reported the designer heuristic covering 2 of 13 rows and would have led to the opposite decision. The corrected, project-scoped measurement is **34 of 39**.

## Verification

Rebased onto `origin/main` = `1592e9a94` and re-verified there, since the original base was four commits stale and missing #6408's shared tree-sitter change.

The **`vbnet-mini` graded fixture is unmoved**: 20/20 entities, 23/23 relationships, 0 forbidden, exit 0 — built fresh, with the grader self-reporting `git HEAD 94a76f7261df` so it cannot be a stale binary (#6283). Both guards are inert on those three sources by construction: no `Partial`, no `.Designer.vb` sibling.

`go test ./internal/vbnet/ ./internal/extractors/vbnet/ ./internal/quality/` → 0. `go test ./internal/extractors/ -run FileAnchored` → 0 (structural refs moved, so that guard is load-bearing here). `go vet`, `gofmt -l` → clean.

Mutant, compiled (`go vet` = 0): `if true ||` prepended to `partialAnchor`'s early return. `TestPartialDesignerPairMergesToOneComponent` **failed** while **all four anti-over-merge tests still passed** — the mutant kills only the direction it should. Restored by `cp`, `shasum de758ed1…` matching.

### DIFF
--- a/internal/extractors/vbnet/extractor.go
+++ b/internal/extractors/vbnet/extractor.go
@@ -111,7 +111,10 @@ func extractVBNet(src, filePath, repoRoot string) []types.EntityRecord {
 	// The file carrier owns anything declared outside a type, so file-level and
 	// namespace-level call sites anchor there rather than being dropped.
 	appendCalls(&out[0], res.File)
-	emit(&out, res.File, filePath, repoRoot, 0, "")
+	// The sibling cache is per-extraction: it is never shared across files, so
+	// concurrent extraction needs no synchronisation, and a designer file
+	// declaring several partial types reads its sibling once.
+	emit(&out, res.File, filePath, repoRoot, 0, "", siblingCache{})
 	return out
 }
 
@@ -125,14 +128,14 @@ func extractVBNet(src, filePath, repoRoot string) []types.EntityRecord {
 // Emitting a namespace node per file would create one duplicate component per
 // file sharing a namespace, which in this corpus is 92 files for StaxRip.UI
 // alone.
-func emit(out *[]types.EntityRecord, n *vbnet.Node, filePath, repoRoot string, ownerIdx int, ns string) {
+func emit(out *[]types.EntityRecord, n *vbnet.Node, filePath, repoRoot string, ownerIdx int, ns string, sib siblingCache) {
 	for _, child := range n.Children {
 		if child.Kind == vbnet.NodeNamespace {
 			inner := child.Name
 			if ns != "" && inner != "" {
 				inner = ns + "." + inner
 			}
-			emit(out, child, filePath, repoRoot, ownerIdx, inner)
+			emit(out, child, filePath, repoRoot, ownerIdx, inner, sib)
 			continue
 		}
 
@@ -142,7 +145,7 @@ func emit(out *[]types.EntityRecord, n *vbnet.Node, filePath, repoRoot string, o
 			// entity of their own. Their call sites belong to the nearest
 			// declaration that does have one, which is what ownerIdx is.
 			appendCalls(&(*out)[ownerIdx], child)
-			emit(out, child, filePath, repoRoot, ownerIdx, ns)
+			emit(out, child, filePath, repoRoot, ownerIdx, ns, sib)
 			continue
 		}
 
@@ -150,9 +153,10 @@ func emit(out *[]types.EntityRecord, n *vbnet.Node, filePath, repoRoot string, o
 		// S7a (#6327) — a `Partial` type declared in a designer half is
 		// emitted under its `Foo.vb` sibling so both halves derive one
 		// graph.EntityID and the existing fold merges them. See partial.go for
-		// why the anchor is per-file, why the sibling must exist, and why the
-		// rewritten half's span is dropped rather than kept.
-		recFile, rewritten := partialAnchor(filePath, repoRoot, child)
+		// why the anchor is per-file, why the sibling must declare the same
+		// type in the same namespace, and why the rewritten half's span is
+		// dropped rather than kept.
+		recFile, rewritten := partialAnchor(filePath, repoRoot, ns, child, sib)
 		startLine, endLine := child.Span.StartLine, child.Span.EndLine
 		if rewritten {
 			startLine, endLine = 0, 0
@@ -191,7 +195,7 @@ func emit(out *[]types.EntityRecord, n *vbnet.Node, filePath, repoRoot string, o
 				},
 			})
 
-		emit(out, child, filePath, repoRoot, idx, ns)
+		emit(out, child, filePath, repoRoot, idx, ns, sib)
 	}
 }
 
--- a/internal/extractors/vbnet/extractor.go
+++ b/internal/extractors/vbnet/extractor.go
@@ -87,7 +87,7 @@ func (e *Extractor) Extract(_ context.Context, file extractor.FileInput) ([]type
 	if len(file.Content) == 0 {
 		return nil, nil
 	}
-	out := extractVBNet(string(file.Content), file.Path)
+	out := extractVBNet(string(file.Content), file.Path, file.RepoRoot)
 	// Both tags are load-bearing, not decoration: relLanguage
 	// (internal/resolve/refs.go:5893) reads Properties["language"] off each
 	// RELATIONSHIP, and that value is what selects the lang=="vbnet" arm in
@@ -99,7 +99,7 @@ func (e *Extractor) Extract(_ context.Context, file extractor.FileInput) ([]type
 }
 
 // extractVBNet is the testable core.
-func extractVBNet(src, filePath string) []types.EntityRecord {
+func extractVBNet(src, filePath, repoRoot string) []types.EntityRecord {
 	res := vbnet.Parse(src)
 
 	out := []types.EntityRecord{extractor.FileEntity(extractor.FileInput{
@@ -111,7 +111,7 @@ func extractVBNet(src, filePath string) []types.EntityRecord {
 	// The file carrier owns anything declared outside a type, so file-level and
 	// namespace-level call sites anchor there rather than being dropped.
 	appendCalls(&out[0], res.File)
-	emit(&out, res.File, filePath, 0, "")
+	emit(&out, res.File, filePath, repoRoot, 0, "")
 	return out
 }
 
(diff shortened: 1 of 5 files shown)


cajasmota:
## Independent review: REQUEST-CHANGES. Two claims in the body above are wrong, and they are mine.

Marking this draft while the fixes land. Correcting the record first, because both false claims were load-bearing for the design argument.

**1. "The anchor must be a pure function of one file's path" — false.** The guard `os.Stat`s the sibling, so the anchor depends on the *filesystem*, not on the path. That is a second input, and it reintroduces the exact full-vs-incremental identity disagreement (#6150) that this PR used to *reject* the cross-file grouping alternative. Two concrete divergences, both confirmed against `internal/extractors/incremental.go:700-720` (only changed+deleted files are re-extracted) and `:800-810` (eviction keys on `changedSet[e.SourceFile]`):

- Delete `Form1.vb` from a merged pair: incremental evicts the merged Component and never re-extracts the unchanged designer half, so **`Form1` disappears from the graph entirely**. A full index of the same tree yields `Form1` anchored at `Form1.Designer.vb`.
- Add `Form1.vb` beside an existing lone `Form1.Designer.vb`: incremental leaves the designer's old anchor, giving **two** Components where a full index gives one.

I rejected an alternative design for this defect class and then shipped it by another route. The fix will document this as an explicit known divergence rather than assert it away; making the incremental path invalidate the sibling is a larger change and is not in scope here.

**2. "structural refs moved, so that guard is load-bearing here" — false.** `TestNoNewFileAnchoredTypeRelationships` scans only `FromID` bound to a file-path expression (`internal/extractors/file_anchored_rels_guard_test.go:534-620`). This diff moved a **`ToID`**. The guard is blind to this change by construction, and there are no `vbnet` entries in `allowedFileAnchored` at all. It passing is evidence of nothing here. The exit code was real; my reading of what it proved was not.

**3. A third claim was partially wrong.** The second test direction pins *name-keyed* merging, not **namespace** collapse — and the code does collapse namespaces. `Widget.vb` in `Namespace Alpha` and `Widget.Designer.vb` in `Namespace Beta` fuse into one EntityID, because `vbnet_namespace` is a property and not part of the identity triple. The path-derived anchor was the only thing keeping them apart, and the rewrite removes it. `TestSameNamedClassesInDifferentNamespacesStayDistinct` looks like it covers this but uses two non-designer files with no `Partial`, so it passes with the entire feature deleted.

## What else the review found

**A strict regression versus no fix**, in the shape where the sibling exists but declares a *different* type: `Widget.Designer.vb` (`Partial Class Widget`) beside `Widget.vb` (`Public Class Gadget`) emits `name=Widget file=Widget.vb span=0-0` — wrong file, zero span, nothing merged. Same for any extra type a designer file declares. The existence guard checks that *a file* exists, never that it declares the same type.

**Platform-divergent output**, and macOS gets the worst of it: a sibling on disk named `Widget.VB` makes `os.Stat` succeed on case-insensitive filesystems and fail on Linux. On macOS that yields no merge, *plus* a Component pointing at a path no file entity claims, *plus* a zeroed span.

**Three of four mutants survived**, all compiling:

| Mutant | Exit | Meaning |
|---|---|---|
| span zeroing removed | 1 — killed | genuinely covered |
| `containsRef(kind, recFile, …)` → `filePath` | 0 — **survived** | the file→type CONTAINS re-key is untested |
| `st.IsDir()` guard removed | 0 — **survived** | directory guard untested |
| `repoRoot == ""` guard removed | 0 — **survived** | its test is **vacuous** |

The `containsRef` one matters most: it is a real behaviour change, not a no-op — the file entity would key its CONTAINS on `Foo.Designer.vb` while the Component lives at `Foo.vb`, dangling the edge. The test helper only inspects edges emitted *by the component*, never those emitted *by the file entity*, so half the diff had no coverage.

`TestPartialAnchorWithoutRepoRootDoesNotRewrite` passes with the guard it claims to pin deleted: `filepath.Join("", "Forms/Form1.vb")` is relative and fails to stat from the test CWD regardless. That is this repo's known vacuous-fixture class, caught in our own work.

## Unaffected

The merge direction itself works, and the attacks that failed are worth recording: `Partial Public Class`, `Public Partial Class`, `Partial Friend NotInheritable Class`, `Partial Module/Structure/Interface`, preceding comments and attribute lines all merge correctly; modifiers are order-independent and case-folded; `Dim partial As Integer` does not trigger a rewrite; a *directory* named `Widget.vb` is correctly refused. Applying the `Partial` gate to the designer half is the right choice and the corrected-measurement narrative in the body holds.

The corpus numbers were not re-measured here and remain as stated, with one caveat now attached: if any designer half's sibling declares a *different* type, those rows counted as "merged" in the 86→52 delta are actually mis-anchored, so **34 removed may slightly overstate the win**.

cajasmota:
## Independent re-review of `6efc1c796`: REQUEST-CHANGES. The prior blockers are closed; the fix introduced a new one.

**All three previous blockers are confirmed genuinely closed**, each by a mutant that now dies: the namespace over-merge (the previously-vacuous test is now real, driven by an actual designer pair carrying `Partial`), the wrong-file/zero-span regression when the sibling declares a different type, the file-carrier `containsRef` re-key — killed by a test that inspects **file-carrier** edges, which was precisely the gap the first review named — and the `repoRoot == ""` guard, whose test now uses an absolute path and **asserts its own precondition**, so the CWD accident cannot recur. The false "pure function of one file's path" claim is replaced by a `# KNOWN DIVERGENCE` block that matches `incremental.go` as cited.

### What the fix broke

**A FIFO sibling hangs the extractor forever.** `siblingPath` does not filter entry types, so a named pipe called `Widget.vb` is returned as the anchor and `os.ReadFile` blocks on `open(2)` until a writer appears. Confirmed: extraction blocked for over 3 seconds. That deadlocks an indexing worker.

This is a **new** regression — the pre-commit `os.Stat` returned instantly — and it slips past the file walker's own file-type vetting, because `partial.go` reads a path the walker never handed it. The round-1 argument for deleting the `IsDir` guard was that a non-regular file "fails the read and declares nothing." That holds for directories. **FIFOs do not fail; they block.** The argument was right about the case it was tested on and wrong about the case it wasn't.

### A false justification, and this one is partly on my brief

`partial.go:224-226` justifies the exact namespace compare with "the identity triple is case-sensitive, so a case-differing declaration would not fold into one id anyway." **The namespace is not in the identity triple at all** — `graph.EntityID` hashes only (repo, kind, name, sourceFile), `internal/graph/graph.go:259-269`. I said as much when I asked for this fix, and the justification that came back inverted it.

Consequence, confirmed: `Namespace ALPHA` and `Namespace Alpha` are the **same** namespace in VB.NET, which is case-insensitive — and they currently fail to merge. A silently dropped legitimate merge. The behaviour is conservative rather than corrupting, so it is not a blocker on its own, but the stated reason cannot stand: it is exactly what a future reader would cite to keep the compare exact.

### Three mutants survive against the new code

| Mutant | Result |
|---|---|
| drop nested-namespace qualification in `typeDecls` | **SURVIVED** |
| drop the nested-type recursion `walk(child, ns)` | **SURVIVED** |
| `n != 1` → `n == 0` (case-folding ambiguity refusal) | **SURVIVED**, and unreachable on macOS |

The first two are correct today but unpinned — dropping either silently loses every merge in a nested-namespace file, or every type declared inside a `Module`.

The third is the interesting one. `siblingPath` calls `os.ReadDir` itself with **no injection seam**, and APFS collapses `Widget.VB` and `Widget.Vb` into a single entry, so the refusal arm cannot be exercised on this machine at all — the test skips. "It will run on Linux CI" is not good enough when the local mutant survives silently, so the fix adds a seam that separates the pure decision from the I/O, making the arm testable everywhere.

### Smaller

A **symlinked** sibling rewrites and fails **open** — the same shape as the case-spelling defect this commit fixed. Being decided deliberately rather than left implicit. And the memoisation claim overstates itself: only `ReadFile`+parse is cached, while `siblingPath`'s directory read runs *before* the cache lookup, so it is paid per partial type.

Fixes are in flight. `vbnet-mini` still grades 20/20, 23/23, 0 forbidden, on a fresh binary whose grader printed `git HEAD 6efc1c796b97`.

Issue #6414, opened 2026-08-20
fastapi: fold include_router(prefix=) into the endpoint path — blocked on deciding the incremental contract, not on the alias resolution


cajasmota:
Split out of #6385 (reported by @arthurgeron). That issue keeps the small, incremental-safe half: emit a `url_mount_point` synthetic from the mount file so the linker's existing prefix-retry recovers cross-service link recall. **This issue is the other half — actually folding the prefix into the endpoint's `path`.**

It is filed separately because it cannot ship until the incremental contract is decided, and that decision is the substance of the work.

## Why the obvious implementation is wrong

Model it on `ApplyDjangoNestedURLConf` (`internal/engine/django_urlconf_nested.go:99`) — a Pass 2.6 that reads sibling files through a reader built over the classified set — and it breaks in three ways, because Django attributes the composed endpoint to the **mount** file (`django_urlconf_nested.go:227`, `SourceFile: relPath`):

- **Stale.** Edit only the route file. The composed endpoint lives on the unchanged mount file, so it is carried forward (`cmd/grafel/index.go:1623-1634`) and the graph keeps a path derived from the *old* route file. Rename a route → ghost endpoint that no longer exists.
- **Vanishing.** Edit only the mount file. Its previous entities drop out of carry-forward, the pass re-runs, but the reader returns nil for the unchanged route file (`index.go:4351-4353`; the pass then `continue`s at `django_urlconf_nested.go:180-183`) → composed endpoints disappear until the next full index.
- **Absent entirely on the daemon.** `internal/extractors/incremental.go` re-runs framework detection per changed file (`:1194-1197`) but never invokes the Django cross-file pass — grepping that file for it yields one hit, an unrelated migration-prune comment at `:1460`. So on the path most users actually hit, every reindex of a route file re-emits the short path and overwrites whatever a full index produced.

Django already carries this defect. **Adding a second instance is not neutral**: Django's composed entities are ANY-verb, while FastAPI's are per-verb and directly user-visible, so the blast radius is larger.

## What has to be decided before code

[...]

Issue #6385, opened 2026-08-19
fastapi include_router(prefix=) is not folded into the extracted path, so routes mounted with a prefix are emitted short


arthurgeron:
I index a set of Python services with grafel and join their HTTP endpoints to a frontend by path. Every endpoint those services define on a router mounted with `include_router(..., prefix=...)` is emitted short by that prefix segment.

`APIRouter(prefix="/network")` is folded into the extracted path. `app.include_router(router, prefix="/network")` is not. Only the constructor form is read, so a repository that declares the router bare and applies its prefix at the mount site gets a path with the segment missing.

This is the FastAPI case of the structural item deferred in #5688:

> Remaining structural item (non-Django `url_mount_point` / cross-module base-URL) is deferred to v0.1.9 per plan.

## Example

For example, given two files:

```python
# app/api/markets.py
from fastapi import APIRouter

router = APIRouter()

@router.get("/{chain_id}/terms")
async def terms(chain_id: int):
    return {}
```

```python
# app/main.py
from fastapi import FastAPI
from app.api import markets

app = FastAPI()
app.include_router(markets.router, prefix="/network")
```

The emitted `http_endpoint_definition` carries `path: "/{chain_id}/terms"`. The route the service serves is `/network/{chain_id}/terms`.

Moving the prefix to `APIRouter(prefix="/network")` in the first file produces the full path, which is what isolates the mount site as the cause.

## Where this is in the code

`synthesizeFastAPI` (`internal/engine/http_endpoint_synthesis.go:2809`) is per-file: it receives one file's content. Its only prefix source is `fastapiRouterPrefixes` (`:2793`), which scans for `<var> = APIRouter(...)` constructors, and the `compose` closure at `:2824-2829` prepends whatever that map holds. The stated policy is at `:2757-2761`.

No endpoint synthesizer reads the `prefix=` argument of `include_router`. Tree-wide, `include_router` appears in no non-test Go file; the matches are two YAML relationship rules, one Strawberry test, and one skills doc. Both YAML rules use the same pattern, which unescapes to:

```
(\w+)\.include_router\s*\(\s*(\w+)
```

Two capture groups, receiver and router variable, and `relationship: ROUTES_TO`. Neither captures `prefix=`.

grafel's own documented canonical FastAPI layout is the shape this cannot handle. `skills/grafel-tech-docs/conventions/fastapi.md:19-31`:

```
app/
  main.py             # FastAPI() instance + include_router calls
  api/
    v1/
[...]

Issue #2702, opened 2026-05-27
[P0] Consumer-side /api/v1/ prefix retry — ~80 frontend orphans


cajasmota:
## Source
Direct query against upvate group post-#2677, 2026-05-27.

\`archigraph_endpoints --action calls --repo_filter upvate-core-frontend --orphan_only=true\` returns 107 orphans, of which ~80 are the same pattern: frontend calls a path **without** the \`/api/v1/\` prefix that the backend defines.

## Examples (from the orphan list)
- \`GET /inspection-types\` ← backend has \`GET /api/v1/inspection-types\`
- \`GET /buildings/inspection-filters\` ← backend has \`GET /api/v1/buildings/inspection-filters\`
- \`POST /login\` ← backend has \`POST /api/v1/auth/login\` (note: also needs the \`/auth/\` part)
- \`GET /contracts/{id}\` ← backend has \`GET /api/v1/contracts/{pk}\`
- \`GET /version.txt\` ← static asset, won't-fix
- \`GET /dashboard/elv3\` ← backend has \`GET /api/v1/dashboard/elv3\`
- \`POST /buildings/bulk_delete\` ← backend has \`POST /api/v1/buildings/bulk_delete\`
- \`GET /domains\` ← backend has \`GET /api/v1/domains\`

These all use a network client that constructs the full URL by prepending a base path internally (axios baseURL, fetch wrapper, etc.). The static analyzer sees only the relative portion.

## Acceptance
1. When the cross-repo HTTP linker fails to find an exact match for a consumer call:
   - Try prepending each registered \`url_mount_point\` prefix (from #2677). The upvate group has \`/api/v1/\` as a mount point — retry the match.
   - Try common framework defaults if no mount points exist (\`/api/\`, \`/api/v1/\`, \`/api/v2/\`).
2. New telemetry counter: \`cross_repo_resolve_hits_by_strategy.mount_prefix_added\` so we can measure how many calls this strategy rescues.
3. Real-data verification: after fix, rebuild upvate. Expect frontend orphan_calls to drop from 194 to ≤120 (target ~80 fewer).
4. Integration test fixture under \`cmd/archigraph/testdata/audit_mount_prefix_retry/\`: simple client/server split with a mount point at \`/api/v1/\` and a client call without the prefix — assert resolution after retry.
5. Make sure the retry uses already-discovered \`url_mount_point\` entities (don't reinvent the discovery).

## Won't-fix categories (out of scope)
[...]

PR #6415, opened 2026-08-20
fix(#6385): FastAPI include_router(prefix=) now emits a url_mount_point, so the linker's existing prefix retry can see it

cajasmota:
First arm of #6385, reported by @arthurgeron. Refs #6414.

Routes mounted with `app.include_router(router, prefix="/network")` are emitted short by that segment. The constructor form `APIRouter(prefix=...)` already folds; the mount form was **not read by any Go code at all** — the reporter verified that, and it holds.

## This does not fold the prefix into the path

That is #6414, and it is deliberately out of scope. Folding needs the route file and the mount file at once, and a cross-file pass **does not survive incremental indexing**: edit only the route file and the composed path goes stale on the unchanged mount file; edit only the mount file and the endpoints vanish; and on the daemon fast path that class of pass never runs at all. Django already carries that defect, and a second instance would be worse here because FastAPI endpoints are per-verb and user-visible.

## What this does instead — strictly per-file

Emit one additive `http_endpoint` synthetic per mount site, carrying `pattern_type: "url_mount_point"` and `url_prefix`, mirroring `django_urlconf_nested.go:140-158` field for field and attributed to the mount file.

**It only ever reads the file it is given**, so it behaves identically on a full index, on `--incremental`, and on the daemon fast path. That property is the whole reason this arm ships first.

**The payoff needs zero linker changes**, and that was verified rather than assumed: `internal/links/http_pass.go:1047` gates solely on `e.Properties.Get("pattern_type") == patternTypeURLMountPoint`, reads `url_prefix` → `route` → `path` as fallbacks, feeds `normalizeMountPrefix` (`:3249`), and `continue`s so the mount is never indexed as a producer. **There is no `framework` test anywhere in that branch.** So the existing #2702 mount-prefix retry picks these up as-is.

Mounts are emitted *before* the decorator-marker guard, because a pure wiring `main.py` may carry no decorator at all. Dedup rides the existing `seen` map.

## A dead branch, found by mutation and deleted rather than shipped

The first draft had an explicit `if trimmed == "" { continue }` guard. Neutering it left the suite green — a degenerate prefix already canonicalises to `/` and is rejected by the next guard, so the branch was unreachable. Rather than ship a line no test could ever exercise, it was **deleted**, with a comment recording why, and the equivalent mutation re-run against the surviving guard: killed, with `empty`, `root`, and `trailing-slash-only` all failing on a `http:ANY:/:mount` leak.

That is the right outcome from a surviving mutant. An untestable branch is not "extra safety" — it is code that cannot be verified and will be trusted anyway.

| Mutant | `go vet` | Test | Result |
|---|---|---|---|
| `url_prefix` → `""` | 0 | 1 | killed |
| `SourceFile` → `""` | 0 | 1 | killed |
| `StartLine` → `1` | 0 | 1 | killed |
| `FindAll…(content, -1)` → `1` | 0 | 1 | killed |
| missing `prefix=` treated as a real prefix | 0 | 1 | killed |
| degenerate-prefix guard neutered | 0 | 1 | killed |

Every mutant compiled (`vet` = 0) and was grepped to confirm it landed before running — a mutant that fails to compile, or never lands, proves nothing.

## Two existing tests, handled explicitly

`TestSynth_Strawberry_FastAPIIntegration` is **vacuous with respect to prefixes** and already annotated as such (#6387): rewriting its fixture to `prefix="/zzz-not-used"` leaves it green, and so does deleting the `include_router` line. It was left alone here — its vacuity is not this PR's to fix, and it is not evidence of coverage either way.

`TestSynth_FastAPI_ImportedRouterByName` asserts the **short** path as expected behaviour, with a comment saying "no prefix since none is discoverable in this file". Since this change never touches route-path composition, that test is neither broken nor made vacuous — it now simply coexists with an unrelated mount synthetic. #6414 is where its intent has to be revisited.

## Verification

New test fails before the change (exit 1, `mounts: map[]`) and passes after (exit 0). `go test -count=1 ./internal/engine/` → 0 · `go vet` → 0 · `gofmt -l` → empty. `internal/links/` was deliberately not modified, so its suite was not run.

Not claimed: that this recovers link recall end to end. The harvest path is confirmed by reading; no corpus index was run to measure the recall delta.

### DIFF
--- a/internal/engine/http_endpoint_synthesis.go
+++ b/internal/engine/http_endpoint_synthesis.go
@@ -435,6 +435,16 @@ func applyHTTPEndpointSynthesis(args DetectorPassArgs) DetectorPassResult {
 	// regex (e.g. Spring's @GetMapping pattern). We only want one
 	// synthetic per endpoint per file.
 	seen := map[string]bool{}
+	// emitMountPoint appends an additive `url_mount_point` synthetic (#6385).
+	// These are not routable endpoints — the linker harvests them purely for
+	// their prefix — so they are deduped on the mount ID alone.
+	emitMountPoint := func(e types.EntityRecord) {
+		if seen[e.ID] {
+			return
+		}
+		seen[e.ID] = true
+		entities = append(entities, e)
+	}
 	// lastEndpointIdx records the index, in `entities`, of the http_endpoint
 	// (definition/call) entity appended by the most recent emit() call, or -1
 	// when that call emitted nothing (canonical-path empty, non-app file, or a
@@ -946,7 +956,7 @@ func applyHTTPEndpointSynthesis(args DetectorPassArgs) DetectorPassResult {
 		// (Spring-style placeholder), which the resolver dropped and
 		// the response-shape extractor could not parse. #753.
 		synthesizeFlask(string(content), emitDef)
-		synthesizeFastAPI(string(content), emitDef)
+		synthesizeFastAPI(string(content), path, emitDef, emitMountPoint)
 		// #2690 — Starlette / Tornado / Pyramid endpoint synthesis.
 		// Starlette and Pyramid use emitDef because the handler def lives in
 		// the same file as the routing site in the common case; when the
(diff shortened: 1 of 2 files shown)


cajasmota:
## Independent review: REQUEST-CHANGES. The design holds and the linker claim is confirmed — but it can publish a *wrong* prefix.

**The linker claim verified independently and HOLDS.** `internal/links/http_pass.go:1047` gates solely on `pattern_type`; grepping the whole harvest block (`:1043-1066`) for `framework` returns **exit 1** — no framework test anywhere. It falls back `url_prefix` → `route` → `path` and `continue`s, so a `verb=ANY` mount is never indexed as a producer and cannot match a consumer on its own. `normalizeMountPrefix` lowercases, so a `/Network` prefix normalises correctly. Zero linker changes needed, exactly as claimed.

**The per-file claim also holds**, confirmed by reading: the emitter takes only content and path, touches only `strings.*`, two package-level regexes, `Canonicalize` and `SyntheticID`. No corpus reader, no cross-file cache, no order dependence.

### What blocks

**A nested call publishes the wrong prefix.** The regex captures the whole argument tail and then takes the **first** `prefix=` anywhere inside it — including one belonging to an inner call:

```python
app.include_router(build_router(prefix="/internal"), prefix="/api/v1")
```

emits `url_prefix=/internal`. The real mount prefix is discarded.

That distinction matters more than it might look. A *missed* prefix leaves the graph where it is today. A *wrong* prefix is published: `normalizeMountPrefix` accepts it, it lands in `mountPrefixesByRepo`, and `http_pass.go:1423` takes the **union across every repo**, so a plausible-but-wrong `/internal/` becomes a global retry candidate. It is bounded — it can only mint a link if a real producer happens to sit at `/internal/<consumer-path>` — but this is precisely the "wrong prefix in the table" outcome I asked the reviewer to hunt, and it is the worst one available here.

**The PR's own headline design decision is unpinned.** I wrote in the body that mounts are emitted *before* the decorator-marker guard "because a pure wiring `main.py` may carry no decorator at all." Moving the emission **behind** that guard leaves the suite **green** — vet 0, test 0. The stated rationale is untested.

The cause is textbook: every new fixture contains `from fastapi import FastAPI` and `app = FastAPI()`, so no test ever exercises the case the placement exists for. The 6-mutant table in the body does not cover it. This is the repo's active vacuous-fixture class landing on the one decision the PR argues hardest for.

**And that placement has a cost I did not state.** Because emission precedes the marker guard, a file with zero FastAPI evidence emits `framework: "fastapi"` — a Flask file containing `routers.include_router(...)`, and `include_router` appearing inside a **docstring** or triple-quoted string. String content becomes graph data. The fix has to reconcile these: stay ahead of the decorator marker so wiring files work, without firing on files with no FastAPI signal at all.

Smaller: `prefix="/a\"b"` yields `url_prefix=/a\`, and `prefix=" /spaced "` yields a prefix with spaces (only `/` is trimmed, never whitespace) — inert for matching, but it pollutes entity IDs and the retry candidate list. Two `include_router` calls on one line match only the first, and two nesting levels abort the match entirely — so the nested-paren tolerance is asymmetric: enough to admit the wrong-prefix bug, not enough to be robust.

### Cleared

`requireContains` uses exact string equality, and `http:ANY:/graphql:mount` never equals `http:ANY:/graphql` — so no existing assertion is newly satisfiable. Both the Strawberry test and `TestSynth_FastAPI_ImportedRouterByName` still pass, and I checked specifically that neither was made vacuous rather than merely left green.

Comment-borne and `my_include_router(...)` attacks correctly fail — the line anchor and the `\.` do their job.

**Out of scope, filed separately:** the mount ID is not file-qualified, so two files in one repo mounting the same prefix collide. That is identical to Django's existing behaviour field-for-field, so it is pre-existing rather than introduced here.

cajasmota:
## Fixes pushed as `dadda7fde`. Coordinator mutation check below.

**The rule, now explicit in code:** emission stays **ahead** of the decorator-marker guard and carries its own evidence gate — emit only when the file has an independent FastAPI signal a pure wiring module would still have (a `fastapi` import, or a same-file `FastAPI(` / `APIRouter(` construction) **and** the mount call is real code rather than text. Comments and triple-quoted regions are masked before the scan.

That gate is deliberately **wider** than the marker guard — it accepts a lowercase `import fastapi`, which the guard rejects — so the pure-wiring case the placement exists for actually works. And it is strictly **narrower** than "the word `include_router` appears", so a Flask module is no longer published as FastAPI.

The wrong-prefix bug is fixed by replacing the tail-capturing regex with a bracket-balanced scanner that elides nested groups. That closed all three regex defects at once: the inner `prefix=` theft, two calls on one line, and two nesting levels aborting the match. Malformed literals are **rejected rather than normalised** — a missed prefix beats a wrong one, which is the whole lesson of the review.

The masking also turned out to be strictly stronger than the `(?m)^[ \t]*` anchor it replaced: an *indented* docstring slipped past the anchor, and the triple-single-quoted case had been passing only by accident of it.

### My own mutant, finer-grained than the fix report's

The report shows the masking as a single mutant. I split it — disabling **only** the triple-quoted branch while leaving `#`-comment masking intact:

| Mutant | `go vet` | Test | Result |
|---|---|---|---|
| triple-quote masking disabled, comment masking kept | 0 | **1** | **KILLED** — and only `docstring-only` and `triple-single-quoted-literal` fail |

That is the result I wanted: the coverage is *precise*, not blunt. Two subtests fail and nothing else, so the docstring and triple-quote cases are each pinned by an assertion that actually discriminates, rather than by one coarse test that would have died to any mutation anywhere in the masker.

File restored by `cp`, `shasum 8420b591…` verified identical. No `git checkout --` used.

**The fix report identified two of its own mutants as *equivalent* and discarded them** rather than reporting them as surviving defects — writing nested opening brackets verbatim, and dropping the elision space, both of which leave the inner text unreadable to `\bprefix` anyway. That is the correct call and the harder one to make; a surviving-mutant count is only meaningful if equivalent mutants are excluded from it.

**A residual limitation, stated rather than left implicit:** single-line string literals are **not** masked, so `X = "app.include_router(r, prefix='/x')"` on one line still emits. This is deliberate — the prefix value has to be readable out of the masked copy, so short literals are skipped rather than blanked, and a full Python tokenizer is the only clean fix. Docstrings and triple-quoted blocks, the cases from the review, are covered. **There is no test pinning the single-line case either way**, which is the part worth fixing; I am filing it as a follow-up rather than leaving the behaviour accidental.

PR #6419, opened 2026-08-20
fix(#6358,#6359): protobuf oneof members were silently dropped, and rpc type edges were bare-name IMPORTS anchored on the file

### LINKED ISSUE #6358: protobuf: every oneof field is silently dropped — the members nest one level deeper than buildMessage looks
**Component:** `internal/extractors/proto` · **Severity:** silent data loss

`buildMessage`'s `message_body` loop accepts only direct `field` and `map_field` children. The grammar nests `oneof` members one level deeper, under a `oneof` child, so they are never visited.

Measured on a file that parses **perfectly** (`error_ratio 0.0000`):

```proto
syntax = "proto3";
message M {
  oneof choice { string a = 6; int32 b = 7; }
  string c = 1;
}
```

Result: **1 entity, 1 CONTAINS edge** — both for the non-`oneof` field `c`. The `oneof` and both its members produce zero entities and zero edges. No warning, no error, no stats line. A consumer sees a one-field message.

`oneof` is ubiquitous in proto3 — it is the idiom for tagged unions and for optional presence, and it appears in most non-trivial service definitions. Since #6357 turned this extractor on for every `.proto` file, every such message now indexes as though those fields do not exist.

**Fix:** recurse into `oneof` children of `message_body` and emit each member through `buildField`, exactly as a direct field. The `oneof` group itself probably warrants its own entity so the union semantics survive — a `oneof` is not the same as N independent optional fields — but at minimum the members must stop vanishing.

Note this is a *different* failure mode from the proto2 gap: those files are rejected whole and produce nothing, which is at least honest. This one produces confident, incomplete output.

Refs #6356, #6357

### LINKED ISSUE #6359: protobuf: rpc request/response edges are bare-name, attributed to the file, duplicated, and can be literal "?"
**Component:** `internal/extractors/proto`, `buildRPC` · **Severity:** the most valuable fact in a proto file does not link

`buildRPC` emits:

```go
Relationships: []types.RelationshipRecord{
    {FromID: file.Path, ToID: reqType, Kind: "IMPORTS"},
    {FromID: file.Path, ToID: respType, Kind: "IMPORTS"},
}
```

Four problems in three lines:

1. **`FromID` is the file, not the rpc.** `rpc SayHello(HelloRequest) returns (HelloReply)` attributes both edges to `helloworld.proto`, so the graph never records that *SayHello* is the thing taking a `HelloRequest`. Observed on grpc-go's `helloworld.proto`: `IMPORTS from="…/helloworld.proto" to="HelloRequest"`, and the same for `HelloReply`.
2. **`ToID` is a bare identifier**, not a structural ref — while every other edge this extractor emits uses `BuildOperationStructuralRef` or the Format-B member ref. It cannot bind to the `message HelloRequest` entity sitting in the same file.
3. **Duplicated when request and response share a type.** `rpc Unary(Outer) returns (Outer)` emits the identical edge twice.
4. **Literal `"?"` targets.** `reqType, respType := "?", "?"`, and when the grammar yields no `message_or_enum_type` child the defaults are emitted verbatim as relationship targets.

Also: `Kind: "IMPORTS"` is the wrong verb for a request/response type. It shares a kind with the file-level `import "x.proto";` edges, which point at *file paths*, so consumers cannot tell the two families apart.

The consequence is that grafel indexes a gRPC service and cannot answer "what does this rpc take and return" — which is the single fact a protobuf file exists to state, and the cross-service contract topology grafel is for.

**Fix:** `FromID` the rpc entity; `ToID` a structural ref subject to the same resolvability rule as the field-type `REFERENCES` edges (see #6357's `dropUnresolvableTypeRefs`); dedupe identical request/response types; drop the `"?"` defaults rather than emitting them; and use a kind that distinguishes a signature type from a file import.

Cross-file request/response types will need the same proto-package index described in #6357's suppression rationale — until that exists, a same-file-only edge is the honest subset.

Refs #6356, #6357

cajasmota:
Two commits, one package. Fixes #6358. Fixes #6359.

They ship together because they collide in the same three test files and both change per-fixture edge counts — the assertion shape those files use — and because they must settle the same resolvability convention. #6358 lands first so that `oneof`-typed messages exist as entities before #6359's same-file-only rule is applied to rpc types; reversed, #6359 would ship an "honest subset" that #6358 then quietly widens with no test noticing.

## Commit 1 — `oneof` members were silently dropped

`buildMessage`'s `message_body` loop accepted only direct `field`/`map_field` children. The grammar nests `oneof` members one level deeper, so on a file that parses at `error_ratio 0.0000` a message with a `oneof` indexed as though those fields did not exist. No entity, no edge, no warning — `oneof` appeared **nowhere** in the package.

The loop is now a `switch` with an `addMember` closure; `field`/`map_field` behave exactly as before, and `case "oneof":` recurses one level through the same `fieldName` / `fieldTypeAndLabel` / `buildField` / `namedTypeRefs` path.

**Grammar shape confirmed against the real CST**, not assumed: `oneof` → `oneof_field` children whose layout is identical to `field`. Also confirmed that a `map` inside a `oneof` parses as an `ERROR` node — it is illegal proto — so oneof members always take the non-map path.

The member entity is appended **before** the CONTAINS edge that targets it, because `phantom_6357_test.go` walks every emitted edge asserting a backing entity exists.

**The `seen` dedupe stays per-message, deliberately.** proto3 requires field names unique across the whole message *including across sibling `oneof` blocks*, so a repeat is invalid proto rather than a case to support — and the member ref is itself per-message, so two same-named members would mint one address and the resolver would un-bind the pair as ambiguous. Per-message dedupe keeps the dedupe and the address space in agreement. That reasoning is in the code and pinned by a test, rather than inherited by accident.

**Out of scope, deliberately:** the `oneof` group entity. A `oneof` is a tagged union, and members-only loses the mutual-exclusivity semantics — that is real, but it needs a new 4-part ID form and reshapes every per-message CONTAINS count. Members-only converts silent data loss into correct-but-flat, which cannot regress. `TestProto_Oneof_GroupIsNotAnEntity` fails the day someone starts emitting it, which is the intent.

## Commit 2 — rpc request/response edges

`buildRPC` emitted `{FromID: file.Path, ToID: <bare type>, Kind: "IMPORTS"}` — file-anchored, bare-name, duplicated, and literally `"?"` when the types did not parse. The in-code comment asserted this shape was intentional.

Now: `REFERENCES` with an **empty FromID** (entity-anchored, matching `buildService`'s service→rpc convention), `ToID` built through `namedTypeRefs` so it flows through `dropUnresolvableTypeRefs` — the same convention #6358's refs use. No types parsed means no edges, so `"?"` can no longer reach an edge; it survives only in the human-readable signature. The same type on both arms emits **one** edge with `direction="request,response"` instead of a duplicate pair.

**The allow-list was deleted, not satisfied.** `allowedBareIMPORTS` exempted exactly these edges, and asserted the bare target *was* emitted — so a correct fix failed it twice. The whole map went, along with both consuming branches and the staleness bookkeeping. That file now has zero exemptions.

## A real bug the new tests caught

`types.Props` is a key-sorted, binary-searched slice. The first draft emitted `{via_rpc, direction}` unsorted, which made `Get()` silently return `""` for **both** keys. Fixed to sorted order. Worth recording because it fails silently rather than loudly — nothing panics, the properties simply are not there.

## Mutants

| # | Mutation | `./proto/` | `-run FileAnchored` |
|---|---|---|---|
| M1 | `oneof_field` → a name that never matches (pre-fix behaviour) | **1** — kills 3 of 4 new tests | — |
| M2-A | re-add `FromID: file.Path` to the new REFERENCES edges | **1** | **1** — `proto:buildRPC:REFERENCES (1 site)` |
| M2-B | restore the exact pre-fix literal (both IMPORTS, bare ToIDs, `file.Path`) | **1** — kills 5 tests incl. the phantom invariant | — |

All compiled (`go vet` = 0), all grep-verified as landed, all restored by `cp` with matching shasums.

**M2-A is the interesting one**: correcting the verb without correcting the anchor trades a silent mis-kind for a *correct* guard failure. Those edges passed the file-anchored guard before only because they were mis-kinded as `IMPORTS`, which the guard exempts. **M2-B proves deleting the allow-list was load-bearing** — with the exemption gone, the phantom invariant catches the very edges it was written to catch.

## Verification

`go test ./internal/extractors/proto/` → 0 · `go test ./internal/extractors/ -run FileAnchored` → 0 · `go vet ./internal/extractors/...` → 0 · `gofmt -l` → empty.

### DIFF
--- a/internal/extractors/proto/fields_test.go
+++ b/internal/extractors/proto/fields_test.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"testing"
 
-	"github.com/cajasmota/grafel/internal/extractor"
 	_ "github.com/cajasmota/grafel/internal/extractors/proto"
 	"github.com/cajasmota/grafel/internal/types"
 )
@@ -68,7 +67,7 @@ message User {
 	if userMsg == nil {
 		t.Fatal("User message entity not found")
 	}
-	wantRef := extractor.BuildOperationStructuralRef("proto", "u.proto", "Order")
+	wantRef := msgTypeRef("u.proto", "Order")
 	found := false
 	for _, r := range userMsg.Relationships {
 		if r.Kind == "REFERENCES" && r.ToID == wantRef {
@@ -112,7 +111,7 @@ message Cart {
 }
 `
 	entities := extract(t, "c.proto", src)
-	wantRef := extractor.BuildOperationStructuralRef("proto", "c.proto", "Order")
+	wantRef := msgTypeRef("c.proto", "Order")
 	found := false
 	for _, e := range entities {
 		if e.Subtype != "message" || e.Name != "Cart" {
--- a/internal/extractors/proto/oneof_6358_test.go
+++ b/internal/extractors/proto/oneof_6358_test.go
@@ -3,7 +3,6 @@ package proto_test
 import (
 	"testing"
 
-	"github.com/cajasmota/grafel/internal/extractor"
 	"github.com/cajasmota/grafel/internal/types"
 )
 
@@ -139,8 +138,8 @@ func TestProto_Oneof_MembersAreContainedAndReferenced(t *testing.T) {
 		}
 	}
 
-	orderRef := extractor.BuildOperationStructuralRef("proto", "ev.proto", "Order")
-	statusRef := extractor.BuildOperationStructuralRef("proto", "ev.proto", "Status")
+	orderRef := msgTypeRef("ev.proto", "Order")
+	statusRef := msgTypeRef("ev.proto", "Status")
 	if via, ok := refs[orderRef]; !ok {
 		t.Errorf("missing REFERENCES edge Event → Order (%q) from a oneof member", orderRef)
 	} else if via != "order" {
--- a/internal/extractors/proto/oneof_6358_test.go
+++ b/internal/extractors/proto/oneof_6358_test.go
@@ -1,0 +1,230 @@
+package proto_test
+
+import (
+	"testing"
+
+	"github.com/cajasmota/grafel/internal/extractor"
+	"github.com/cajasmota/grafel/internal/types"
+)
+
+// ---------------------------------------------------------------------------
+// #6358 — oneof members were silently dropped.
+//
+// buildMessage scanned message_body's DIRECT children for `field` / `map_field`
+// only. The grammar nests a oneof's members one level deeper:
+//
+//	message_body
+//	  field                ← seen
+//	  oneof
+//	    oneof_field        ← NEVER visited
+//	    oneof_field
+//
+// so a message whose fields live inside a oneof indexed as though those fields
+// did not exist: no entity, no CONTAINS edge, no REFERENCES edge, and no skip
+// counter or warning anywhere — on a file that parses at error_ratio 0.0000.
+// The word "oneof" appeared nowhere in the package.
+//
+// Scope: MEMBERS ONLY. No entity is emitted for the oneof group itself, so the
+// mutual-exclusivity semantics of the tagged union are still unmodelled. That
+// is tracked separately; it needs a new 4-part ID form and reshapes every
+// per-message CONTAINS count, which a data-loss fix must not smuggle in.
+// TestProto_Oneof_GroupIsNotAnEntity below pins that boundary so the follow-up
+// is a deliberate change rather than an accident.
+// ---------------------------------------------------------------------------
+
+const oneofSrc = `syntax = "proto3";
+
+message Order { string id = 1; }
+
+message Event {
+  string id = 1;
+  oneof payload {
+    string text = 2;
+    Order order = 3;
+    Status status = 4;
+  }
+  int32 seq = 5;
+}
+
+enum Status { STATUS_UNSPECIFIED = 0; }
+`
+
+// TestProto_Oneof_MembersBecomeFieldEntities is the direct data-loss assertion:
+// every member of the oneof must exist as a SCOPE.Schema/field entity carrying
+// its resolved type, exactly as a top-level field does. Before #6358 the three
+// oneof members were absent and only Event.id / Event.seq existed.
+func TestProto_Oneof_MembersBecomeFieldEntities(t *testing.T) {
+	entities := extract(t, "ev.proto", oneofSrc)
+
+	got := make(map[string]string)
+	for _, e := range entities {
+		if e.Kind == "SCOPE.Schema" && e.Subtype == "field" {
+			got[e.Name] = e.Properties["type"]
+		}
+	}
+	want := map[string]string{
+		"Order.id":     "string",
+		"Event.id":     "string",
+		"Event.text":   "string", // oneof member, scalar
+		"Event.order":  "Order",  // oneof member, named message type
+		"Event.status": "Status", // oneof member, named enum type
+		"Event.seq":    "int32",
+	}
+	for n, wt := range want {
+		gt, ok := got[n]
+		if !ok {
+			t.Errorf("no SCOPE.Schema/field entity for %q — oneof members are being dropped", n)
+			continue
+		}
+		if gt != wt {
+			t.Errorf("field %q Properties[\"type\"] = %q, want %q", n, gt, wt)
+		}
+	}
+	if len(got) != len(want) {
+		t.Errorf("emitted %d field entities (%v), want exactly %d", len(got), got, len(want))
+	}
+
+	// A oneof member carries no proto label: `repeated`/`optional` are illegal
+	// inside a oneof, so a non-empty label would mean the label reader latched
+	// onto the wrong token.
+	for _, e := range entities {
+		switch e.Name {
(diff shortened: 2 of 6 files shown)


cajasmota:
## Independent review: REQUEST-CHANGES. Both fixes are real, and one of them introduces a new silent edge loss.

Every grammar claim in the body re-verified true against the real CST, the allow-list deletion is genuinely clean (zero live references repo-wide), and the empty-`FromID` convention was confirmed to hold on **both** the full and incremental paths — which was the claim I most wanted checked, since holding on only one would have dangled the edges on the other.

### What blocks

**A self-loop when an rpc and its message share a name — and the edge is silently lost.** rpc entities and message entities share **one address space per file**, both addressed by `BuildOperationStructuralRef`. So `service S { rpc User(User) returns (User); }` emits a REFERENCES edge whose `ToID` is the rpc's *own* id. Assembly stamps `FromID` to the same value, and three separate call sites discard `FromID == ToID` as a self-loop. The intended rpc→message link disappears.

This is **new**: the pre-fix code emitted a bare `file → "User"` IMPORTS edge for that shape. And the phantom invariant cannot catch it — that test folds `"endpoint"` and `"message"` into the same namespace, so it is structurally blind here.

**A claim in this PR body is false, and I repeated it from the commit message.** I wrote that the member entity is appended before the CONTAINS edge "because `phantom_6357_test.go` walks every emitted edge asserting a backing entity exists." Reordering the two appends leaves the suite at **exit 0**. Both slices are returned together from `buildMessage`, and the phantom test checks set *membership*, not order — it structurally cannot enforce the ordering. The rationale is being deleted rather than propped up with a new test, because pinning a non-property would be worse than the wrong comment.

Also: one arm of `TestRPC_NeverEmitsQuestionMarkTarget` asserts over an **empty set** — `rpc Weird() returns ();` yields an entity with zero relationships, so "no `?` escaped" is trivially true. The `Stream` arm is fine, and the file as a whole is still pinned by the M2-B mutant.

### Two things the review confirmed that are worth keeping

The `direction="request,response"` collapse is **necessary, not merely tidy**: `cmd/grafel/index.go:6141` dedupes on `(from, to, kind)` ignoring Properties, so emitting two edges would have silently kept only `direction="request"`.

And the `Props` sorting regression *is* pinned — an independent mutant swapping the order to `{via_rpc, direction}` dies with 3 failures. That is the bug the implementer caught mid-flight, and it now cannot come back silently.

**Known limitation, being recorded rather than fixed:** `stream` cardinality is not modelled at all, so `rpc Stream(stream M) returns (stream M)` is indistinguishable from unary `Ping(M) returns (M)`. Filing separately.

cajasmota:
## Re-review of `d48ea3f6f`: APPROVE. The resolution claim is now measured, not inferred.

The fix's load-bearing claim — that `scope:schema:message:…` binds the message rather than the same-named rpc — was labelled INFERRED, read from `structuralKindFamilies`. That was the one thing worth spending a review on: if it were wrong, this would have traded a self-loop for a **dangling** edge, which is strictly worse, since a self-loop is at least discarded visibly at three known sites while a dangling ref points at an id no entity carries.

It was confirmed by running **real resolution** end to end — real tree-sitter parse, real `proto.Extractor.Extract`, real `ComputeID`, real `BuildIndex`, real `resolve.References` — not by reading the kind tables. Resolver stats: `rewritten=16 ambiguous=0 ToUnmatched=0`.

| Case | Result |
|---|---|
| `rpc Ping(Req) returns (Resp)` | **RESOLVES** — both arms bind the message entities |
| `rpc User(User) returns (User)` | **RESOLVES to the message** — distinct ids, no self-loop |
| field type ref via `buildMessage`'s changed path | **RESOLVES** |
| cross-file type ref | **DROPPED**, not dangling — no edge emitted at all |

**The negative control is what makes this meaningful.** Mutating `messageTypeRef` back to `BuildOperationStructuralRef` and re-running the identical probe flips case 2 to a self-loop while cases 1 and 3 keep resolving. So the new address space is precisely and solely what fixes the collision — not an incidental side effect of some other change in the diff.

I ran a separate mutant of my own beforehand: swapping the scope kind to a bogus-but-consistent string kills **10 tests**. Together those two results cover the failure mode I was actually worried about — that everything agrees on the same wrong address and the suite is green because it is self-consistent.

### Filed as follow-up, not fixed here

The reviewer demonstrated that `fileContainsRel` still addresses messages through the **operation** form, so in the collision case the file→message CONTAINS resolves onto the **rpc** and the message ends up with no inbound containment. Genuinely pre-existing — `fileContainsRel` is untouched by this diff — and this PR is a strict improvement, since before it the message had the same broken CONTAINS *and* no REFERENCES edge. Filed as #6422, with the note that the bug is invisible until a name collides, which is why it has gone unnoticed.

Also folded into that issue: `messageTypeRef` omits the `filepath.ToSlash` normalization every other `Build*StructuralRef` helper applies. It resolves anyway because `lookupStructural` normalizes defensively, but it is one line out of step with the convention.

Merging once CI reports.

PR #6426, opened 2026-08-20
feat(#6327): S7b — Handles and AddressOf produced no edges at all; 1,212 now, 78.5% resolved

cajasmota:
S7b of the VB.NET epic. Refs #6327.

`Handles` clauses and `AddressOf` expressions both name a member as a value, and both produced **no edge at all** before this. On the 302-file corpus: **0 edges before, 1,212 after, 952 of them resolved (78.5%).**

## Edge kind: `REFERENCES`, handler → handled

Both constructs reference a member as a value rather than invoking it, which is what `REFERENCES` already means in `golang/`, `java/` and `python/references.go`.

**CALLS was rejected**: neither site is call-shaped, so it would inflate the corpus's 8,702 measured `IsCall` sites with edges no execution path takes.

**A new kind (`HANDLES`/`SUBSCRIBES_TO`) was considered and rejected explicitly.** No other language models event wiring, so a VB-only kind would be invisible to every existing traversal and gate — a relationship nothing can query is not modelling, it is decoration. The construct is preserved as a `via` property (`handles` | `addressof`) instead. Direction is forced by the empty-`FromID` anchor.

## Target naming was decided by measurement, not taste

Three renderings were implemented and measured against the corpus:

| Target rendering | Edges | Resolved |
|---|---:|---:|
| as written, `Me`/`MyBase` folded | 1,070 | 17 |
| + bare names qualified by enclosing type | 1,213 | 434 |
| + `Handles` targets the **receiver**, event as a property | 1,212 | **952** |

The first row is the naive choice and it resolves 17 edges out of a thousand.

Two findings drove the rest. A bare `Worker` resolves for CALLS but **not** for REFERENCES — `internal/resolve/refs.go:3273` gates the leaf-name fallback on `relKind == "CALLS"` (#778/#6177) — while `declName` emits members as `<Type>.<member>`, so qualifying by the enclosing type is what makes them findable. And in `Handles Button1.Click` the *field* is in-tree (thanks to S7a's partial merge) while the *event* belongs to `System.Windows.Forms.Button` and can never resolve; targeting the event costs **518** resolved edges. The event is kept as a property rather than discarded, and `Me.`/`MyBase.` receivers still target the event.

## Corpus measurement

- Keyword tokens after stripping comments and literals: **Handles 596, AddressOf 585**
- Parsed: **654** `Handles` targets (multi-target clauses expand), **585/585** `AddressOf` operands — zero shortfall in any file
- Edges: **653** handles + **559** addressof (difference is per-owner dedup)
- Resolved: handles 521/653, addressof 431/559

Reaching 585/585 required one extra scan site: `walkProperty` scans no refs and pushes no frame, so auto-property initialisers were invisible — **142 of 585** operands live there. Only the operand scan was added; wiring `scanRefs` in would have moved S5's CALLS set, which is out of scope for this stage.

**The residual 260 are all dotted, and the split matters**: 85 name a node the graph holds but the lookup declined as *ambiguous* (one corpus project declares the same class in three files); 175 name nothing in-tree — qualified by a module, a local, or `My.Forms.*`, the same unnameable receiver `callTarget` already refuses. Neither group is a parse failure.

## Mutants

Fourteen ran. Two results worth stating rather than burying:

**M4 (`AddressOf` left boundary dropped) SURVIVED** on the first attempt. The boundary fixtures used `x = MyAddressOf`, which has no operand after the keyword — so they died on the *right* boundary instead and never exercised the left one. Fixtures now carry a trailing word; M4b kills.

That investigation also exposed a **dead `prev == '['` arm**: a `]` always follows a bracket escape, so the right-boundary check already refuses it and no test could ever have killed that branch. Removed, with the reasoning recorded — an unreachable branch is not extra safety, it is code that will be trusted anyway.

**M6 did not compile**, so it proved nothing and was re-run as M6b, which kills.

The other twelve — targeting the event instead of the receiver, dropping the type prefix, not folding `Me`/`MyBase`, case-sensitive keyword matching, removing the auto-property scan, `FromID` = file path, dropping the `event` property, `REFERENCES`→`CALLS`, keeping only the first of a multi-target `Handles`, and re-qualifying an already-qualified target — all die.

## Verification

`go test ./internal/vbnet/ ./internal/extractors/vbnet/ ./internal/quality/` → 0 · `./internal/extractors/ -run FileAnchored` → 0 · `go vet` → 0 · `gofmt -l` → empty.

**The `vbnet-mini` golden fixture did not move**: 20/20 entities, 23/23 relationships, 0 forbidden, on a freshly built binary under full isolation. Those three sources contain zero `Handles` and zero `AddressOf`, so it could not have. `expected.json`'s description and `NOTICE.md` were edited — both still claimed partial-merge, `Handles` and `AddressOf` were unimplemented S7 gaps, which S7a and this commit make false. **Doc text only; no expected counts touched.**

Out of scope and untouched: S7c (method-level `Implements`), S7d (With-blocks), S7e (`.aspx.vb`).


cajasmota:
## Independent review: REQUEST-CHANGES. A claim in the body above is false, and it is mine.

**"The event name is kept as the `event` property, so nothing the source said is discarded" — untrue for one shape.** `add()` dedups on **`ToID` only**, and because S7b targets the *receiver*, `Handles Btn.Click, Btn.MouseDown` collapses to a single edge carrying `event=Click`. **`MouseDown` is gone.** Confirmed by probe.

This is not the multi-target case the mutant table covers — `Handles A.Click, B.Click` (different receivers) works correctly. It is multiple *events on one receiver*, which M11 never exercised. The same collapse also merges a `Handles` edge with an `AddressOf` edge to the same target, keeping `via=handles` and hiding the addressof site. Corpus impact is small (654 targets → 653 edges), so this is about correctness and about the claim, not volume.

**A dead branch, and a test that appears to pin it but cannot fail.** Literals are already masked upstream before `scanRefs` hands text over, so deleting the whole `case '"'` literal-skip arm in `addressof.go` leaves both packages **green**. With it gone, `x = "AddressOf Handler"` and the property/field-initialiser variants still yield zero operands — meaning the `{"literal", …}` test case asserts nothing about that code.

That is the same standard this PR invoked *three lines above* to delete the `prev == '['` arm — "an unreachable branch is not extra safety, it is code that will be trusted anyway" — applied inconsistently. Being right about one and not the other in the same function is worse than being wrong about both, because it makes the surviving branch look deliberate.

Also found: `walkProperty` returns early without setting `.property` when a name is unparsable, so operands scan onto the *previously* declared property — `Property = AddressOf Bad` lands `Bad` on the preceding property's node. Malformed source only.

### What the review confirmed rather than took on trust

**The `refs.go:3273` claim holds exactly.** That line reads `strings.ToUpper(relKind) == "CALLS" && !strings.ContainsAny(ref, ":.#")`, gating the #778 leaf-name fallback to CALLS only, with a comment explicitly refusing to extend it. So the target-naming rationale — the reason the measured table moves from 17 to 434 resolved — is not a misreading. One nuance the body omits: the *ambiguous* path at `refs.go:3257` calls `lookupBareWithLocality` with no kind gate, so bare REFERENCES stubs are not universally excluded — only the unmatched path is.

**The `prev == '['` removal was correct.** `[AddressOf]` is refused by the right-boundary `]` check, and no reachable input distinguishes the arm. That deletion stands.

**The golden fixture is byte-identical to `origin/main`** on a semantic JSON compare: 20 entities, 23 relationships, 4 forbidden, all identical; only `description` changed. The reviewer was explicit that it did **not** run the grader end-to-end (that needs a full-repo compile, which the swap budget forbids), so the 20/20 · 23/23 figure follows from byte-identical expectations plus zero keyword occurrences in those three sources — not from an observed run. I would rather have that stated than assumed.

**Attacks that failed, worth recording** so they are not re-tested: case folding, `MyAddressOf`/`AddressOfThing`, `[AddressOf]`, `AddressOf [Stop]`, tab separators, `_` continuation across physical lines, `'`/`REM` comments, escaped `""`, attribute blanking, `:`-joined statements, lambdas, file-level operands, `Handles` as a parameter name, `Handles` in a string, and `Handles Btn.Click` following `As Dictionary(Of String, Integer)` — `cutKeyword` is depth-aware and multi-generic multi-target splitting is unbroken.

cajasmota:
## Fixes pushed as `01b32a4e7`. Coordinator mutation check below.

**The dedup key is now `(target, event, via)`, not `target` alone** — and `line` is deliberately kept **out**, which is the part worth reading twice. Including `line` would have broken a genuine dedup the PR already pins: `AddressOf X` beside `AddressOf Me.X` fold to one member and should stay one edge. That existing test stayed green unmodified, which is the evidence the finer key did not over-correct.

The false sentence in the package doc is rewritten to state the key explicitly rather than to claim nothing is discarded.

**The dead `case '"'` arm is deleted, and confirmed dead by reading the path rather than by mutant survival alone:** `walkLine` hands `BlankAttributes(MaskStringLiterals(…))` to every statement walk, and `StringFill` is `0x00` — literal content cannot spell `a`/`A`, and quotes are the only surviving bytes. That is the standard this PR applied to `prev == '['`, now applied consistently.

The `{"literal", …}` test case is **kept but re-commented**. It is a real end-to-end assertion — it fails if masking ever stops covering this path — and it now explicitly disclaims being a mutant-killer for the scanner. That distinction is the honest version: the test has value, just not the value its position implied.

The stale-property bug was fixed **at the source** rather than guarded: `walkProperty` now returns the node it opened (`nil` on the no-name path) and the S7b arm uses that return, so no stale-state path remains at all.

### Mutants

The fix round ran an **inverse** mutant worth highlighting: re-adding the deleted literal arm leaves both packages green, which *demonstrates* equivalence rather than asserting it. That is the right way to justify deleting code.

My own, finer-grained than the round's — dropping **only** `event` from the key while keeping `via`:

| Mutant | `go vet` | Landed | Test | Result |
|---|---|---|---|---|
| key → `target + via` (drop `event`) | 0 | yes | **1** | **KILLED** — `TestReferences_DedupKeepsDistinctEventsAndConstructs` |

The round's own mutants covered `target`-only and dropping `via`; this covers the third face. All three components of the key are now independently pinned, so none can be dropped silently later.

File restored by `cp`, `shasum 6200418d…` identical. No `git checkout --` used.

Golden fixture untouched — no fixture file appears in the diff at all.

Issue #6104, opened 2026-08-02
MergeWithCustom supersede: cross-kind destruction AND same-kind in-place corruption (280 entities, 5 kinds, 2 languages)


cajasmota:
`MergeWithCustom` supersedes a base-path entity with a custom-extractor entity **by Name**. When the two disagree on Kind, the base entity is destroyed and every edge produced by a *different* pass that still targets the old ID dangles.

This is not a hypothetical. It is the blocker holding #5989.

## Reproduction

A 723-file Django/DRF/Celery fixture indexed with in-process custom extraction enabled (the #5989 branch, default-off flag):

```
SCOPE.Pattern    0 → 801
SCOPE.Operation      +1000
SCOPE.Schema         +960
SCOPE.Service    0 → 240
SCOPE.Class      0 → 100
Task           240 → 0        ← destroyed
```

A custom Celery extractor emits an entity with the same Name as the base path's `Task` node but a different Kind, so the merge replaces it. `Task` goes to zero. Edges emitted elsewhere that pointed at the base node — `Function:send_receipt`, `scope:operation:method:python:...` — are left pointing at an ID that no longer exists.

#4402 re-keys the superseded node's own self-edges, but it cannot re-key edges produced by other passes, so that fix does not cover this.

## Why it generalises

Celery is only where this fixture caught it. The exposure is: **any repo containing a base-path entity whose Name collides with a custom-extractor entity of a different Kind**. Name is not a unique key across extractors and was never intended to be one — the graph's identity function is `graph.EntityID(repoTag, kind, name, sourceFile)`, which includes Kind precisely because Name alone does not identify an entity.

So the merge is keyed on strictly less information than the identity it is superseding.

## Consequence

Enabling custom extraction is **not purely additive** — it can delete entity kinds the base path produced, and leave dangling edges behind. That inverts the expected trade: a change meant to add ~21% more entities also silently removes a category.

## What a fix needs

Two things, and the second is the one that makes it safe:

1. Match on identity (Kind + Name + source file), not Name alone, so entities that merely share a name are not conflated.
2. When a supersede genuinely does occur, re-key **all** edges targeting the old ID, not just the superseded node's own. That requires the merge to see the relationship set, which today it does not.

## Status

Pinned by `TestInProcCustomExtractorsCelerySupersedeRegression` on the #5989 branch, which asserts the **broken** behaviour deliberately and documents what to change when this is fixed. The #5989 flag stays default-off until then.

Refs #5989, #4402

[...]

PR #6427, opened 2026-08-20
fix(#6369): one import could unresolve every edge to that name repo-wide — a real declaration now outranks a placeholder in byName

cajasmota:
The b2 arm of #6369. Refs #6369.

`BuildIndex` indexed **every** non-empty-Name record into `byName`, import placeholders included. Adding one file containing an import whose last segment collides with a real type name made the global entry ambiguous and dropped **every** edge to that name repo-wide — including in files that never imported it. The damage is global, not local to the importing file.

## What this is NOT: a blanket skip

I briefed this as "make `BuildIndex` skip import placeholders." **That would have been wrong, and the guard in the brief caught it.**

Two shapes resolve *through* a placeholder by bare name:

- **graphql** (`graphql.go:352-392`): `extend type User` emits the `Subtype:"import"` stub **plus** `IMPORTS(file → "User")` and `FEDERATES(file → "User")`, with the bare name as ToID.
- **css/scss/less** (`relationships.go:53`): `@import "theme.css"` emits `IMPORTS(file → "theme.css")`, ToID = module string = the placeholder's Name.

And the rescue path does not exist for them: `rewriteOneWithCaller` (`refs.go:3238-3280`) consults the locality tiers **only on `statusAmbiguous`** (plus an unmatched-CALLS special case). When the module is external, nothing else carries that name, so removing the placeholder yields `statusUnmatched` and **nothing rescues the edge** — permanently unresolved.

That is demonstrated, not argued: **mutant M4 is the naive skip, and it is killed by exactly those two guard tests.**

So what shipped is a **precedence**, not a skip:

- a real declaration always outranks a placeholder, **in either arrival order**, and the pairing never raises ambiguity;
- a placeholder may still occupy an **unclaimed** name — which is what keeps graphql and css working;
- two distinct placeholders still go ambiguous, but the ambiguity is remembered as *placeholder-only*, so a later real declaration reclaims the name (the issue's R3 variant).

`byKind`, `byLocation`, `byLocationKind`, `byQualifiedName` and `byPackageComponent` are untouched.

## Two corrections to my own brief

**"One file, small diff" was wrong.** There is a **second** `byName` writer inside `insertModuleEntry` (`symbol_index.go`), production-wired via `BuildIndexFromModulesOrdered`. Without changing it too, the fix would not apply in production and the parity harness would diverge. Both writers now delegate to a single shared `indexByName`. Mutant M5 — module path forgets the marker — is killed by `FULL-INDEX PARITY FAILURE … [byName ambigName]`.

**F# does not stamp the marker.** `fsharp/extractor.go:587-612` emits a bare `SCOPE.Component` with **no `Subtype`**. So the issue's own F# reproduction — the strongest evidence on the ticket, and the case I chose this arm for — is **not** fixed by a `Subtype=="import"`-keyed change. b2 is necessary but not sufficient for F#; that language needs the #6368-style stamp or the #742 route, both out of scope here. The tests therefore use the **vue-shaped** placeholder, which does carry the marker, on the identical F# topology.

I would rather state that plainly than let this PR imply it closes the reproduction it was dispatched for.

## Marker inventory, derived

15 non-test emission sites across 14 packages: razor, proto, css/css, css/scss, css/less, markdown, vue, just, cross/imports, kotlin, cpp (×2), graphql, javascript, fish. `cross/imports` and `kotlin` are immune by construction — they name the placeholder with the **full** module path and route through `Properties["ref"]`/qname, not `byName`.

## Tests and mutants

The new test drives `BuildIndex → ReferencesEmbedded` and **asserts the baseline first**, failing loudly on a vacuous fixture. Before: exit 1, two previously-resolved cross-file EXTENDS dropped by one import in an unrelated file, in all three orderings. After: exit 0.

| # | Mutation | Exit | Killed by |
|---|---|---|---|
| M1 | placeholder no longer yields to an already-indexed real declaration | 1 | cross-file EXTENDS + federation |
| M2 | real declaration no longer displaces a placeholder extracted first | 1 | placeholder-before-real |
| M3 | placeholder-only ambiguity no longer recoverable | 1 | two-placeholders (R3) |
| M4 | **blanket skip** | 1 | css `@import` + graphql federation |
| M5 | module-index path forgets the marker | 1 | full-index parity |

All compiled, all grep-verified as landed, all restored by `cp` with matching shasums.

## Not done, and the issue gates on it

`go test ./internal/resolve/` → 0 (2137 RUN, 376 top-level PASS, 0 FAIL) · `./internal/extractors/kotlin/` → 0 · `go vet` → 0 · `gofmt -l` → empty.

**The full-corpus edge-parity diff was NOT run.** The issue gates b2 on it. The scoped precedence removes the "binds to a placeholder today, unresolved after" failure mode by construction — the placeholder still occupies unclaimed names — and M4 shows that guard is live. But that is an argument plus unit evidence, **not** the corpus measurement the issue asks for before landing.

### DIFF
--- a/internal/resolve/import_placeholder_byname_6369_test.go
+++ b/internal/resolve/import_placeholder_byname_6369_test.go
@@ -318,4 +318,40 @@ func TestImportPlaceholderAmbiguityStillHonoured_6369(t *testing.T) {
 				"placeholder may claim it (#6369)", got)
 		}
 	})
+
+	// A placeholder and a REAL record of the same name in the same file share
+	// one EntityID by construction — EntityID is sha256(repo, kind, name,
+	// sourceFile) and does not include Subtype, so the extractors that mint a
+	// per-import SCOPE.Component emit two records on one ID (measured on the
+	// Go corpus in cmd/grafel/incremental_dup_rows_6094_test.go). The second
+	// record is a RE-INDEX of the same entity, not a collision, so it takes
+	// the same-ID path — and that path must retire the placeholder marker,
+	// because the entity holding the name is now known to be a real
+	// declaration. If the marker survives, the NEXT file's genuinely
+	// different declaration takes the "declaration displaces placeholder"
+	// branch instead of colliding: two real declarations of Animal then
+	// resolve to whichever one arrived last, with ambiguity never raised.
+	t.Run("placeholder re-indexed as a real record then a second declaration stay ambiguous",
+		func(t *testing.T) {
+			recs := []types.EntityRecord{
+				importPlaceholder6369(placeholderID6369, "Animal", "src/other/collide.vue"),
+				{
+					ID: placeholderID6369, Kind: "SCOPE.Component", Name: "Animal",
+					Subtype: "class", SourceFile: "src/other/collide.vue", Language: "vue",
+				},
+				{
+					ID: realAnimalID6369, Kind: "SCOPE.Component", Name: "Animal",
+					Subtype: "class", SourceFile: "src/domain/base.vue", Language: "vue",
+				},
+			}
+			idx := BuildIndex(recs)
+			if !idx.ambigName["Animal"] {
+				t.Errorf("two real declarations of Animal must be ambiguous, even when a " +
+					"placeholder record preceded one of them on the same ID (#6369)")
+			}
+			if got, ok := idx.byName["Animal"]; ok {
+				t.Errorf("byName[Animal] = %q, want absent — a name owned by two real "+
+					"declarations must not resolve to an arbitrary winner (#6369)", got)
+			}
+		})
 }
--- a/internal/resolve/import_placeholder_byname_6369_test.go
+++ b/internal/resolve/import_placeholder_byname_6369_test.go
@@ -1,0 +1,260 @@
+// Package resolve — #6369: an import placeholder must not poison the global
+// bare-name index.
+//
+// A dozen extractors mint one `SCOPE.Component` per import statement, marked
+// `Subtype:"import"` (css/scss/less, graphql, vue, kotlin, razor, proto,
+// markdown, cpp, just, fish, javascript, cross/imports). BuildIndex indexed
+// every one of them in `byName` exactly like a real declaration, so a
+// placeholder whose name collides with a real type flipped that name
+// AMBIGUOUS globally — and every bare-name edge to it, repo-wide, went
+// unresolved. Measured on #6369: adding ONE file containing `open Acme.Animal`
+// dropped both cross-file EXTENDS to `Animal` in a file that imported nothing.
+//
+// The fix is scoped, not a blanket skip: a placeholder may still OCCUPY an
+// otherwise-unclaimed name, because for css `@import` chains and graphql
+// federation stubs the placeholder is the only target the bare-name
+// IMPORTS/FEDERATES edge has (TestImportPlaceholderStillResolvesOwnEdge_6369
+// pins that). What it may never do is displace, or collide with, a real
+// declaration.
+package resolve
+
+import (
+	"testing"
+
+	"github.com/cajasmota/grafel/internal/types"
+)
+
+const (
+	realAnimalID6369   = "1860328bfaee8f13"
+	dogID6369          = "aaaa000000000001"
+	serviceID6369      = "aaaa000000000002"
+	placeholderID6369  = "eca4ea80cdd6b84d"
+	placeholder2ID6369 = "eca4ea80cdd6b84e"
+)
+
+// baseFixture6369 is the #6369 reproduction, transposed onto an extractor that
+// actually stamps the marker (vue: placeholder named after the module's last
+// segment, real component named after its own file).
+//
+//	src/domain/base.vue    declares component  Animal        (the real target)
+//	src/app/impl.vue       Dog, Service        EXTENDS "Animal" by bare name
+//
+// Both EXTENDS resolve cross-file on this input. Nothing in src/app imports
+// anything; the two edges are the "innocent bystanders" of the defect.
+func baseFixture6369() []types.EntityRecord {
+	return []types.EntityRecord{
+		{
+			ID: realAnimalID6369, Kind: "SCOPE.Component", Name: "Animal",
+			Subtype: "class", SourceFile: "src/domain/base.vue", Language: "vue",
+		},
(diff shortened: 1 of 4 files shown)


cajasmota:
## CI caught a REAL regression. Marking draft.

Exactly one failing test across the whole run — `TestIncremental_GuardSuppressesDuplicateImportRow_6094` in `./cmd/grafel/` — reproduced locally, and bisected to these two files: reverting only `refs.go` + `symbol_index.go` to `HEAD~2` makes it pass.

**It is not a stale test.** It is a *differential* test with no hardcoded target ID: it asserts only that the incremental graph's `IMPORTS` rows match a full rebuild of identical source. Nothing in it encodes a pre-fix target. The side that moved is the **full rebuild**.

- **Before:** `r07.go -IMPORTS-> ext:strings` — `SCOPE.External / package / strings`. Correct.
- **After:** `r07.go -IMPORTS-> 2ce12a5d6bf0ac16` — a `SCOPE.Component` named `strings` whose **`SourceFile` is `r11.go`**, an unrelated file.

All 12 fixture files collapse onto that single `r11.go` placeholder, **fabricating 12 bogus intra-repo dependency edges** where correct external-package edges existed. The incremental path — untouched by this change — still produces `ext:strings`, which is precisely why the parity test caught it.

The resolver counters pin the mechanism:

```
base: rewrote=108 ambiguous=36 unmatched=84   (to: rw=36 am=36 um=60)
head: rewrote=144 ambiguous=0  unmatched=84   (to: rw=72 am=0  um=60)
```

Base: those 36 bare-name resolutions were **ambiguous**, so they fell through to the external-library binder. Head: `am=0` — a placeholder now owns the name outright and **the external fallback never runs**.

## Why the review did not catch this

The reviewer's permutation harness verified `0 real + ≥2 placeholders → ambiguous`, exhaustively, and that result stands. This corpus has many same-named placeholders and still produces `am=0`, so those two facts have to be reconciled before repairing — the fix agent is establishing that mechanism first rather than guessing at a repair that merely turns the test green.

Worth being precise about what the earlier retargeting analysis got right and what it missed. It predicted the case where a **real declaration** outranks a placeholder. What actually bites is different: **a placeholder wins a name outright with no real declaration in play at all**, pre-empting an external binding that used to work by *staying* ambiguous. Ambiguity was load-bearing here, and the fix removed it.

## The repair has to satisfy both directions

A placeholder must not pre-empt the external-library binder — that is this regression. But a placeholder occupying an *unclaimed* name is load-bearing and must keep working: graphql's `extend type X` and css `@import` both emit bare-name edges targeting the placeholder's own name, and since `rewriteOneWithCaller` consults the locality tiers only on `statusAmbiguous`, turning those into `statusUnmatched` strands them permanently. The M4 blanket-skip mutant is killed by exactly those two guard tests.

So the line to draw is between "placeholder legitimately owns this name" and "placeholder is shadowing what should be an external binding" — and the fix has to justify where that line belongs rather than move the test.

**This also vindicates the process gate I flagged when opening this PR:** #6369 asks for a full-corpus edge-parity diff before b2 lands, and it was not run. A 12-file fixture in `cmd/grafel` found what unit tests in `internal/resolve` could not, because the defect only appears when a real corpus has many same-named placeholders and no real declaration. The corpus measurement is not ceremony here.

cajasmota:
## Regression fixed at `b9dba14fe`; coverage gap closed at `30bc0957c`. Marking ready.

### The root cause is a fact that has now bitten this project twice

`graph.EntityID` is `sha256(repo, kind, name, sourceFile)` — **`Subtype` is not an input**. So a per-import `SCOPE.Component` placeholder shares its ID *by construction* with any same-named `SCOPE.Component` in the same file. On the Go corpus that means **three records on one ID**: two unmarked from the language extractor, one `Subtype:"import"` from `cross/imports`.

That entity was therefore re-indexed under its **own** ID, skipping the collision branch entirely (`ok && existing != id` is false), and falling through to the insert tail — where the holder flag was rewritten from the *current record's* `isImport`. The trailing placeholder re-raised the flag on a name whose `byName` occupant was a real record. The flag stopped describing the entity in `byName` and started describing the last record that mentioned it, so the next file's real record matched "declaration displaces placeholder" instead of colliding. File after file: never ambiguous, last file wins, external binder never runs.

None of the four mechanisms I proposed was it — not folding, not `PruneImportPlaceholders` (which runs *after* the index), not `BuildIndexFromModulesOrdered` (default-off). It was a fifth thing.

**The same `EntityID` property is what made S7a's partial-class merge structurally hard.** It is a recurring sharp edge, and both times it produced a defect that looked like something else.

### The fix

Holder status is now **AND-ed, never re-raised**: an ID already claimed under a non-placeholder record stays a real declaration however many placeholder records for that ID follow.

Deliberately **not** in the resolver's fallback ordering — the resolver behaved correctly given the index it was handed. `rewriteOneWithCaller` consulting fallbacks only on `statusAmbiguous` is the documented contract; the index was lying about ambiguity. The collision branch, the #6104 facet rule, and placeholder-only-ambiguity recovery are all untouched. One flag, one bookkeeping bug.

Counters are restored byte-identical to the pre-regression baseline: `rewrote=108 ambiguous=36 unmatched=84`, and the rows are `r07.go→ext:strings` / `r07.go→ext:strconv` on **both** sides of the differential test.

### My mutant survived, and my explanation of it was wrong

I found that deleting the flag-clear on a real record left both suites green. I proposed the mechanism was `P → real(A) → real(B)` with distinct IDs.

**That was refuted, and I want it on the record.** With distinct IDs, `R1` takes the `!isImport && nameHolderImport[name]` branch, which carries **its own** `delete` — so the flag is cleared there, mutant or not. Measured under the mutant: `byName=""`, `ambigName=true`, identical to baseline.

The real sequence is the **same-ID** one — `P(id X) → real(id X) → real(id Y, other file)`:

| | `byName` | `ambigName` |
|---|---|---|
| baseline | absent | `true` |
| mutant | file B's declaration | `false` |

Two real declarations silently resolving to an arbitrary winner — the same damage class the reviewer flagged for `delete(nameAmbigImport, name)`.

I asked to be corrected rather than have a test written to match my story, and that is what happened. A test built on my mechanism would have passed for the wrong reason and pinned nothing.

`M4` (blanket skip) still dies on the css `@import` and graphql federation guards, which is the evidence the load-bearing direction was not weakened to make this green.

Issue #6323, opened 2026-08-18
the Operations RSS budget is resolved on every 5s poll (forking sysctl) and reports the next-start value as if it were live


cajasmota:
Two related follow-ups to #6319 (merged), found by review of that PR.

## 1. The budget is now resolved on every poll, and that forks a subprocess

`/api/system` is polled every 5 s by the dashboard (`webui-v2/src/hooks/use-operations.ts:20`). After #6319 that path calls `daemon.RSSBudgetMB()`, which:

- reads and unmarshals `settings.json`, and
- when nothing is persisted, calls `process.TotalMemoryMB()` — which **execs `sysctl -n hw.memsize`** on darwin (`internal/process/sysinfo_darwin.go:14`) and reads `/proc/meminfo` on linux.

So an open Operations tab now spawns roughly **12 subprocesses per minute, indefinitely**, on a path that previously did zero syscalls. Total physical memory cannot change while the process is running.

Fix: `sync.OnceValue` around the memory probe, or resolve the budget once at `dashboard.Server` construction.

## 2. The number shown is the next-start budget, not the live one

`/api/system` is documented as a live daemon status snapshot (`internal/dashboard/handlers_system.go:10`), but `daemon_rss_budget_mb` is explicitly `restart_required` (`handlers_settings.go:220-221`).

So after changing the setting, Operations immediately displays the new value while the running scheduler still enforces the old one. `grafel daemon --max-rss-budget N` (flag, not persisted) is likewise invisible to the dashboard entirely.

Pre-#6319 this field was a constant lie, which is at least easy to spot. Post-#6319 it is a plausible-looking lie, which is harder — it now moves when you change the setting, so it looks live.

The genuinely live value already exists: `proto.Status.RSSBudgetMB` (`internal/daemon/proto/proto.go:158`). It is simply not wired into the dashboard.

Fix: inject the effective budget into `dashboard.Server` at daemon startup, the same way `progressBroker` is injected. That closes (1) and (2) together and covers the flag case, which neither current path does.

## 3. Test gap

`internal/daemon/rss_budget_settings_test.go:86` only exercises the unexported pure `rssBudgetMB(int64)`. Deleting the persisted-setting branch from the exported `RSSBudgetMB()` leaves `./internal/daemon` **fully green** — it is caught only by a cross-package dashboard test. Add a `TestRSSBudgetMB` in `internal/daemon` with `GRAFEL_HOME` and a persisted value, so the exported function is covered where it lives.

[...]

Issue #6322, opened 2026-08-18
dashboard Settings still answers the RSS budget from its own 512 default, so the two screens disagree and saving the form quarters the budget


cajasmota:
Follow-up to #6319 (merged), found by review of that PR. Not introduced by it — but #6319 makes it user-visible in a worse way, so it should be fixed now.

## The third default

`internal/dashboard/handlers_settings.go:70` — `DefaultAppSettings()` hardcodes `DaemonRSSBudgetMB: 512`.

#6319 correctly routed the **Operations** screen through `daemon.RSSBudgetMB()`. The **Settings** screen still answers from its own constant. So on a 16 GB host with nothing persisted:

- Operations shows `2048` (auto: `total/8`, capped at 2048)
- Settings shows `512`

Before #6319 both screens were wrong in the same direction (500 vs 512), which read as a rounding quirk. Now they are wrong *and mutually contradictory*, which is a worse user story — the numbers actively disagree with each other.

## The part that costs the user something

`DaemonRSSBudgetMB: 512` is not merely displayed, it is the form's value. Saving the Settings form round-trips `512` into `settings.json`, after which `RSSBudgetMB()` legitimately returns 512 — so **opening Settings and pressing save quarters the admission budget** on that host, silently, with no indication that anything changed.

That is the same shape as several defects we shipped this cycle: a default that is only correct in one of the two places that own it.

## Fix

Either route `DefaultAppSettings()` through `daemon.RSSBudgetMB()`, or drop the hardcoded field default so the form renders the resolved value rather than a constant. The first is preferable — it leaves exactly one resolver.

Refs #6316, #6319

cajasmota:
**Triaged #6322/#6323/#6324 together, because they look like one bug and are not.** Recording the split so they get dispatched correctly — three agents on this would collide in the same file.

**There is a single source-of-truth function — `daemon.RSSBudgetMB()` (`internal/daemon/rss_budget_settings.go:50`) — but only two of the three symptoms trace to it.**

[...]

PR #6428, opened 2026-08-20
fix(#6324): the over-budget banner compared live whole-process RSS against a next-start delta budget, and rendered the two as a fraction

### LINKED ISSUE #6324: "Over memory budget" compares whole-process RSS against a delta-accounted admission budget, so it still fires on a correctly configured daemon
Found while reviewing #6319. That PR is a genuine improvement, but it does not close the symptom reported in #6316 for the default case, because the banner compares two different quantities.

## The mismatch

`webui-v2/src/routes/operations.tsx:346-348` raises "Over memory budget" by comparing `status.rss_mb` — the **whole process RSS** — against `rss_budget_mb`.

But `cmd/grafel/daemon.go:143-145` documents the budget as the **additional predicted RSS of concurrent index jobs**, delta-accounted, with the daemon's own idle RSS never subtracted.

Those are not the same number and the comparison is not meaningful. It is closer to "is the process bigger than the amount of headroom we allow new work to consume", which is routinely true and says nothing about admission being misconfigured.

## Why it still fires after #6319

On a 16 GB host with nothing persisted, the resolved budget is 2048 MB. This daemon's idle RSS routinely exceeds 2048 MB — see the measurements in the FD/RSS work earlier this cycle. So the banner still appears on a correctly configured daemon by default, which is exactly what #6316 was reporting.

## Fix

Either compare like with like — surface the admission-tracked delta rather than whole-process RSS — or change what the banner says, so it stops asserting a misconfiguration it cannot detect. The first is better; the second is honest.

Worth noting the reporter's original framing was right in substance: the dashboard *was* reporting misleading over-budget telemetry. #6319 fixed one of the two reasons.

Refs #6316, #6319

cajasmota:
Fixes #6324.

`operations.tsx` compared **whole-process RSS** against a budget that is **delta-accounted** — `cmd/grafel/daemon.go` says so explicitly: the budget covers "the ADDITIONAL predicted RSS of concurrently running index jobs only — the daemon's idle RSS is never subtracted." So the banner fired on a correctly configured daemon.

**Our own #6322/#6323 merge made it measurably worse.** That added `rss_budget_scope`, set unconditionally to `"next_start"`, and shipped it to `types.ts` — so the UI was provably comparing a **live** number against a value the backend itself labels **next-start** *and* delta-scoped. Two independent mismatches, both already documented in the reply the UI receives, and **no UI code read the field**.

## What changed

**Informational and scope-qualified, not like-for-like.** The banner can no longer emit `warning`/`danger` tone or the words "Over budget" from numbers that are not comparable.

It also stops rendering RSS as a fraction of the budget. `987 / 500 MB` **is itself the false claim** — it asserts a ratio between two quantities that do not share a denominator — so the value now renders as plain `987 MB` with an `info`-tone **"Next-start budget"** badge whose tooltip names what each number actually is.

**The real over-budget warning still exists**, gated behind an allow-list of scopes that denote a *live whole-process* ceiling. No daemon sends such a scope today, so it is dormant — but a future backend that becomes genuinely comparable gets the warning back **with no frontend change**. An **absent** `rss_budget_scope` is also treated as non-comparable, which is correct: pre-#6323 daemons budgeted the same delta, they simply did not say so.

## Why not the like-for-like comparison

The honest number is the admission counter — `s.usedMB` in `internal/daemon/sched/scheduler.go:478`, which is **unexported, has no accessor, and no JSON field**. Exposing it is scheduler → dashboard handler → `types.ts` → component plumbing, i.e. Go work in three subsystems. That is a separate issue, and half-building it would have been worse than either option.

## The testability trap, and how it was handled

There was **no test anywhere** for this banner — `overBudget` appeared only inside `operations.tsx`. And all existing webui tests live in `webui-v2/src/lib/*.test.ts`; there are **zero** tests under `src/routes/`. So the logic was untestable under this project's own convention, and a fix left in place would have shipped unverified.

The decision is extracted into a pure helper — `webui-v2/src/lib/rss-budget-banner.ts`, `rssBudgetPresentation(status) → {value, badgeLabel, tone, tooltip, comparable}` — with 7 tests beside it. **These are the first tests this banner has ever had.** `operations.tsx` now only consumes the result.

## Verification

RED before the helper existed (exit 1, module not found) → GREEN after (exit 0, 7/7).

**Killing mutant:** adding `"next_start"` to the comparable-scopes allow-list reintroduces the exact defect. Grep-confirmed it landed; **exit 1, three tests failed**. Restored by `cp`, shasum `b9c2cf68…` matching, re-run exit 0.

`npm run lint` (`tsc --noEmit`) → 0. `npm test` → 0, **29 files / 315 tests** (was 28/308).

**No Go was compiled** — this change is entirely within `webui-v2/`.

### DIFF
--- a/webui-v2/src/lib/rss-budget-banner.test.ts
+++ b/webui-v2/src/lib/rss-budget-banner.test.ts
@@ -1,0 +1,91 @@
+import { describe, it, expect } from "vitest";
+import { rssBudgetPresentation } from "./rss-budget-banner";
+
+describe("rssBudgetPresentation", () => {
+  it("never claims 'over budget' when live whole-process RSS exceeds a next-start, delta-scoped budget (#6324)", () => {
+    const p = rssBudgetPresentation({
+      rss_mb: 987,
+      rss_budget_mb: 500,
+      rss_budget_scope: "next_start",
+    });
+    expect(p.comparable).toBe(false);
+    expect(p.tone).not.toBe("warning");
+    expect(p.tone).not.toBe("danger");
+    expect(p.badgeLabel ?? "").not.toMatch(/over budget/i);
+    expect(p.tooltip ?? "").not.toMatch(/over (memory )?budget/i);
+  });
+
+  it("does not render RSS as a fraction of a non-comparable budget", () => {
+    const p = rssBudgetPresentation({
+      rss_mb: 987,
+      rss_budget_mb: 500,
+      rss_budget_scope: "next_start",
+    });
+    expect(p.value).toBe("987 MB");
+    expect(p.value).not.toContain("/");
+  });
+
+  it("qualifies the budget with its scope, and says what each number is", () => {
+    const p = rssBudgetPresentation({
+      rss_mb: 987,
+      rss_budget_mb: 500,
+      rss_budget_scope: "next_start",
+    });
+    expect(p.tone).toBe("info");
+    expect(p.badgeLabel).toBe("Next-start budget");
+    expect(p.tooltip).toContain("500 MB");
+    expect(p.tooltip).toContain("987 MB");
+    expect(p.tooltip).toMatch(/next start/i);
+    expect(p.tooltip).toMatch(/additional/i);
+  });
+
+  it("treats an absent scope as non-comparable too (pre-#6323 daemons budget the same delta)", () => {
+    const p = rssBudgetPresentation({ rss_mb: 987, rss_budget_mb: 500 });
+    expect(p.comparable).toBe(false);
+    expect(p.tone).toBe("info");
+    expect(p.value).toBe("987 MB");
+  });
+
+  it("stays silent when no budget is configured", () => {
+    expect(rssBudgetPresentation({ rss_mb: 412 })).toEqual({
+      value: "412 MB",
+      badgeLabel: null,
+      tone: null,
+      tooltip: undefined,
+      comparable: false,
+    });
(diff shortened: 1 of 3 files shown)


cajasmota:
## Coordinator mutation check

The implementation's own mutant added `"next_start"` to the comparable-scopes allow-list. I ran the **other** half of that guard — the `?? ""` fallback for an **absent** `rss_budget_scope`, changed to default to a comparable scope:

| Mutant | Test | Result |
|---|---|---|
| absent scope defaults to comparable instead of `""` | **1** | **KILLED** — 1 failed, 6 passed |

Both faces of the guard are now independently pinned: an explicitly-declared non-comparable scope, and a missing one. That second case is the one that would rot silently, since it is the shape every pre-#6323 daemon sends.

File restored by `cp`, `shasum b9c2cf68…` identical. No `git checkout --` used.

**One design point worth drawing out**, because it is better than what I briefed: the allow-list means the real over-budget warning is *dormant, not deleted*. A future backend that starts sending a genuinely live whole-process scope re-enables the warning with **zero frontend work**. I had asked only that the banner stop asserting something false; keeping the true version reachable is the stronger answer.

And dropping the fraction rendering was the right call for a reason worth stating: `987 / 500 MB` **is itself the false claim** — it asserts a ratio between two quantities that share no denominator. Rewording the label while leaving the fraction would have fixed the sentence and kept the lie.

Issue #6314, opened 2026-08-15
grafel_find (bm25) culls to a hard top-10 per repo before de-noise/fusion, so a file stem outranks an exact entity-name match and the relevant file can be absent rather than low-ranked


manuel1358000:
## Summary

Investigating a real duplicated-text bug (the same UI string appeared in two
sibling step components of a multi-branch checkout flow), grafel confirmed
the root cause well once I already knew the symbol name — but the initial
discovery step (finding the symbol in the first place) struggled, and one
category of relationship was invisible to the graph entirely.

## What worked

- `grafel_find` in **substring** mode, once I had the exact component name
  from manual grep, instantly returned both duplicate definitions (two
  sibling files under different step-flow branches of the same feature).
- `grafel_related` (`direction=callers`, depth 2) traced a 2-hop chain from
  the duplicated component to the checkout screen entry point: a direct
  `REFERENCES` edge from a sibling summary component, then a `RENDERS` edge
  from that summary compothese confirmed the

cajasmota:
Two reports in two days, both with a real reproduction behind them — thank you, this is the useful kind.

Before I answer the substance I have to flag a problem, because it's ours and it's eating your evidence: **this issue body is truncated.** It ends mid-sentence at "a `RENDERS` edge from that summary compothese confirmed the" — characters dropped mid-word, then the rest of the report gone. Your feedback report in #6313 has the identical signature (`Entities with valid startof 348687)`, `http_endpoint_definitio`, rows cut mid-cell). Two separate submissions with the same mid-word splicing, from output our own tooling produced, is enough for me to stop blaming the paste. Something in how we lay out and emit that text is corrupting it. I'm filing that as its own bug.

Practically, it means I can see your "What worked" section and your title, but the actual findings — the bm25 ranking problem and the `direction=uses` gap — are the part that got cut. Could you re-post the tail? If it's easier, attach it as a file rather than pasting; that'll dodge whatever we're doing to it.

From the title alone, both halves are things I want to chase:

**`grafel_find` bm25 ranking unrelated-domain matches above the relevant file** — worth knowing whether the relevant file was ranked low or absent entirely. Those are different bugs: a ranking problem is tuning, a missing candidate is a retrieval or indexing problem, and the fix is nowhere near the same. Your note that substring mode found both definitions *instantly* once you had the exact name suggests the entities were there and it's ranking, but I'd rather you confirm than have me assume.

[...]

PR #6430, opened 2026-08-20
fix(#6314): direction=uses traversed only NAVIGATES_TO, so every USES/REFERENCES/CALLS edge was unreachable from the parameter named uses

cajasmota:
The **edge half** of #6314, reported by @manuel1358000. Refs #6314 — this does not close it; the ranking half is still blocked on the reporter's evidence.

## The defect is wider than the report's title

`grafel_related direction=uses` did not merely "miss property/enum-member access." It traversed **only `NAVIGATES_TO`** — JS/TS router navigation — and nothing else. `USES`, `REFERENCES`, `CALLS` and `USES_HOOK` edges that genuinely exist in the model were **unreachable** from the parameter named `uses`.

An agent reading the parameter *name* got a silently-incomplete answer. That is the worst possible shape for a tool an AI is meant to trust, and it is why this ranks above cosmetic work.

## Kind set, chosen by measurement

`CALLS, USES, USES_HOOK, REFERENCES, NAVIGATES_TO` — every one confirmed emitted by real extractors (files containing each literal: CALLS 181, REFERENCES 53, USES 26, USES_HOOK 7, NAVIGATES_TO 13). **No named kind turned out to be un-emitted.**

Deliberately excluded, and documented in code rather than left implicit: `CONTAINS`/`IMPORTS`/`DEPENDS_ON` (containment and module-level dependency, not entity-uses-entity); `EXTENDS`/`IMPLEMENTS` (type hierarchy — including them would make "who uses this" inseparable from "who subclasses this"); `INJECTED_INTO` (direction inverted — it answers the opposite question).

**On the reporter's specific symptom:** there is no `ACCESSES` / `READS` / `ACCESSES_PROPERTY` / `USES_ENUM` kind anywhere in `internal/extractors|links|graph` — zero hits. Property and enum-member access is modelled as `REFERENCES`/`USES`, so widening to those kinds is the whole of the available repair from the MCP side. **No extractor work was started**, deliberately.

## A second defect: `used_by` could not have worked even after widening

The incoming `entity_id` filter compared only bare `rel.ToID`. Route destinations carry synthetic bare IDs (`route:/foo`), but a usage edge points at a real entity whose ID callers hold in prefixed `repo::id` form. So widening the kinds alone would have produced an empty result and looked like "no usages exist."

Fixed to accept either form. This is the kind of thing that only surfaces when you actually run the widened path.

## An existing test encoded the defect as the contract

`TestCoreRelatedDispatch` asserted that `uses` was **byte-identical** to `handleNavigates` — i.e. it pinned the bug as the intended behaviour. It now compares against the shared traversal with the usage kind set.

Worth noting: that test **genuinely failed on the real corpus fixture** with a newly-surfaced `CALLS` edge. That is independent evidence the widening bites in real data, not only in a hand-built fixture.

## The response is now self-describing

Edges carry `kind`, and the response echoes the `kinds` traversed. A caller can see what was searched instead of inferring it from a parameter name — which is the root failure this issue is about. `grafel_navigates` itself stays navigation-only.

## Mutants

| # | Mutation | Exit | Verdict |
|---|---|---|---|
| M1 | `usageEdgeKinds` narrowed back to `{NAVIGATES_TO}` | 1 | killed |
| M2 | drop `REFERENCES` — the reporter's symptom kind | 1 | killed |
| M3 | revert the prefixed-ID match in the incoming filter | 1 | killed |
| M4 | leak `usageEdgeKinds` into `grafel_navigates` | 1 | killed |

M1 is the widening-reverted mutant: a test that passed with the widening undone would have proved nothing. M4 guards the blast radius in the other direction.

## Verification

Red before: 3 new tests failed with `kinds returned: map[:1]` — one edge, NAVIGATES_TO only. `go test ./internal/mcp/` → 0 · `go vet` → 0 · `gofmt -l` → empty.

**One process note worth recording:** on the first full run the background notification reported **exit 0 while the output file said `FAIL`**. The run was repeated with the exit code written into the file itself. That is the truncation/false-green trap this repo already documents, firing in practice — trust the redirected exit code, not the wrapper.

### DIFF
--- a/docs/mcp-tools.md
+++ b/docs/mcp-tools.md
@@ -138,7 +138,7 @@ Entities related to X. One tool for every adjacency query — pick the `directio
 - `direction=callers` — inbound callers, **ranked by call frequency**. If `entity_id` starts with `/` and matches no entity, it is treated as an in-app route literal (searches NAVIGATES_TO edges; response carries `resolved_as: "navigation_route"`).
 - `direction=callees` — outbound callees.
 - `direction=neighbors` — graph neighbors (both directions).
-- `direction=uses` / `direction=used_by` — usage edges: `CALLS`, `USES`, `USES_HOOK`, `REFERENCES` and `NAVIGATES_TO` (outgoing for `uses`, incoming for `used_by`), with route/param filters and multi-hop flow (`mode=flow`). Each returned edge carries its `kind`. Containment (`CONTAINS`), module dependency (`IMPORTS`/`DEPENDS_ON`) and type hierarchy (`EXTENDS`/`IMPLEMENTS`) are deliberately excluded — see #6314.
+- `direction=uses` / `direction=used_by` — usage edges: `CALLS`, `USES`, `USES_HOOK`, `REFERENCES`, `RENDERS` and `NAVIGATES_TO` (outgoing for `uses`, incoming for `used_by`). Each returned edge carries its `kind`, and the response echoes the `kinds` traversed. Containment (`CONTAINS`), module dependency (`IMPORTS`/`DEPENDS_ON`) and type hierarchy (`EXTENDS`/`IMPLEMENTS`) are deliberately excluded — see #6314. Note `depth` does not apply here: this path reads `max_depth` (default 5) and only in `mode=flow`. `route`, `with_param` and `mode` reach the handler but are **not declared** in the `grafel_related` schema (#5784) — call `grafel_navigates` if you need them as a supported contract.
 
 Key params: `entity_id` (required), `direction`, `depth` (default 1), `token_budget` (default 800), `route`, `with_param`, `repo_filter[]`.
 
--- a/docs/mcp-tools.md
+++ b/docs/mcp-tools.md
@@ -138,7 +138,7 @@ Entities related to X. One tool for every adjacency query — pick the `directio
 - `direction=callers` — inbound callers, **ranked by call frequency**. If `entity_id` starts with `/` and matches no entity, it is treated as an in-app route literal (searches NAVIGATES_TO edges; response carries `resolved_as: "navigation_route"`).
 - `direction=callees` — outbound callees.
 - `direction=neighbors` — graph neighbors (both directions).
-- `direction=uses` / `direction=used_by` — NAVIGATES_TO and usage edges, with route/param filters and multi-hop flow (`mode=flow`).
+- `direction=uses` / `direction=used_by` — usage edges: `CALLS`, `USES`, `USES_HOOK`, `REFERENCES` and `NAVIGATES_TO` (outgoing for `uses`, incoming for `used_by`), with route/param filters and multi-hop flow (`mode=flow`). Each returned edge carries its `kind`. Containment (`CONTAINS`), module dependency (`IMPORTS`/`DEPENDS_ON`) and type hierarchy (`EXTENDS`/`IMPLEMENTS`) are deliberately excluded — see #6314.
 
 Key params: `entity_id` (required), `direction`, `depth` (default 1), `token_budget` (default 800), `route`, `with_param`, `repo_filter[]`.
 
--- a/internal/mcp/core_merges.go
+++ b/internal/mcp/core_merges.go
@@ -108,8 +108,8 @@ func (s *Server) handleCoreFind(ctx context.Context, req mcpapi.CallToolRequest)
 //	callers (default) → handleFindCallers   (inbound callers)
 //	callees           → handleFindCallees   (outbound callees)
 //	neighbors         → handleNeighbors(direction=both), messaging-aware
-//	uses              → handleNavigates(direction=outgoing)  (NAVIGATES_TO out)
-//	used_by           → handleNavigates(direction=incoming)  (NAVIGATES_TO in)
+//	uses              → usage-edge traversal, outgoing (#6314)
+//	used_by           → usage-edge traversal, incoming (#6314)
 //	messaging         → handleMessagingRelated (topic pub/sub/delivery, cross-repo)
 //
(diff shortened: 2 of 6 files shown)


cajasmota:
## Independent review: REQUEST-CHANGES. The reporter's own truncated sentence points at the gap we left.

**`RENDERS` is missing from the kind set, and the exclusion list never mentions it** — so it reads as an oversight rather than a decision. It is emitted by at least 8 extractors (javascript, angular, svelte, vue, astro, rescript, fsharp elmish); the file-count for the literal is **28**, the same order as `USES` and *above* `NAVIGATES_TO`. It is the component-uses-component edge in the JS/TS corpus this reporter works in.

And the detail that makes this sting: **the truncated issue body literally reads "a `RENDERS` edge from that summary compo…"**. @manuel1358000 was describing this exact edge when the paste was cut off. We widened the traversal and left out the one kind their evidence was pointing at.

Concretely: `direction=used_by` on a React/Vue/Svelte component would still return **nothing** for every parent that renders it — the same silently-incomplete answer this issue is about, one kind over.

**A line was changed that carries two behaviours, and only one got coverage.** The incoming filter is now `rel.ToID != entityID && prefixedID(r.Repo, rel.ToID) != entityID`. The new prefixed half is pinned. Deleting the **bare** half entirely leaves the full suite green — and that compare is what makes `grafel_navigates direction=incoming entity_id=route:/foo` and the documented `navigation_route` fallback work at all.

The reviewer also found a real hardening there: the bare compare is repo-agnostic, and the widening makes that consequential. `direction=used_by entity_id=Logger` in bare form now unions usages of **every** repo's `Logger` in the group. Visible rather than silent — edges carry `from_repo` — but the code comment asserts callers hold the prefixed form and nothing enforces it.

**The new `kind` field is unpinned in `mode=flow`.** Replacing it with a constant leaves the suite green, and `mode` is reachable on `grafel_related`, so a caller can receive flow edges whose `kind` is arbitrary with nothing noticing. The three new tests only exercise `mode=list`.

### On my own claim about the rewritten test

I wrote that `TestCoreRelatedDispatch` "now compares against the shared traversal with the usage kind set." **It is still the same tautology, rewritten** — it compares the implementation against itself, and it passed under the mutant that narrowed the kind set to one element. It pins dispatch *routing*, which is worth having, but it is not coverage of the kind set and I should not have implied it was. The three tests in `related_uses_test.go` are the real coverage, and those are not tautological — they assert concrete kinds against a hand-built fixture.

### Confirmed clean

Blast radius: `handleNavigatesKinds` has exactly three call sites, `collectNavigatesEdges`/`collectNavigatesFlow` have no others, and neither `dead_code` nor `pr_impact` touches this path. Performance: `matchesKind` is an O(5) scan inside the **existing single pass** — no per-kind rescan, complexity still O(edges). The outgoing filter already had the prefixed treatment, so the two directions are now symmetric.

**The prefixed branch cannot false-positive**: it builds `repo + "::" + ToID` and compares the whole string — constructed equality, not a suffix match — so a bare ID that is a suffix of another entity's prefixed ID does not collide.

**Recording separately, not fixing here:** `limit` defaults to 100 and the sort is lexicographic. With `CALLS` in the set, a hot entity's `used_by` now returns the alphabetically-first 100 callers, where `direction=callers` is advertised as "ranked by call frequency". `truncated:true` is set, so it is visible rather than silent.

cajasmota:
## Fixes pushed as `2187948fe`. Marking ready.

**`RENDERS` is in** — the kind the reporter was describing when their paste was cut off. **`INSTANTIATES` is deliberately out**, and the reason is now in the exclusion comment rather than left silent: its only emitter builds a synthetic `tfmodule-def:<dir>` directory marker that the resolver classifies as Dynamic, not an entity. Including it would report marker strings as usages, and **no `used_by` query could ever match one** — cost with no reachable answer. That is the right shape for an exclusion: measured, not asserted.

**Writing the flow test exposed a second defect the widening had introduced**, which was not in the review's list. The flow BFS dedup key was **endpoints-only**. With a single kind in the set, no two entities were ever joined by parallel edges — so the key was correct by accident. With six kinds they routinely are, and the key silently dropped all but the first, **reporting one kind where several applied**.

That is worth dwelling on: the original code was not wrong, it was *unfalsifiable*. Widening the kind set turned a latent assumption into a live bug, and it only surfaced because someone wrote a test for an adjacent thing.

### My mutant

The fix round's five mutants all die. Mine targeted the newly-found defect specifically, since new code is the surface nobody has attacked:

| Mutant | `go vet` | Landed | Test | Result |
|---|---|---|---|---|
| dedup key reverted to endpoints-only (`curr.entityID + "→" + ne.toID`) | 0 | yes, line 461 | **REALEXIT=1** | **KILLED** — `TestRelatedUses_FlowModeCarriesEdgeKind` |

Restored by `cp`, `shasum df624341…` identical. No `git checkout --` used.

### A correction to this PR's body that stands

The body claims `TestCoreRelatedDispatch` "now compares against the shared traversal with the usage kind set." **It does not.** It is still the same tautology — the implementation compared against itself — and it passed under the mutant that narrowed the kind set to one element. It is worth keeping as a pin on dispatch *routing*, but it is not kind-set coverage. The real coverage is `related_uses_test.go`, whose tests assert concrete kinds against a hand-built fixture and do die to the kind-set mutants.

I am leaving the body as written and correcting it here, so the claim and its refutation are both on the record rather than the mistake disappearing.

### Recorded for a separate issue

`limit` defaults to 100 with a lexicographic sort. With `CALLS` in the set, a hot entity's `used_by` returns the alphabetically-first 100 callers — whereas `direction=callers` is advertised in the docs as "ranked by call frequency". `truncated:true` is set, so it is visible rather than silent, but the ordering is arbitrary with respect to relevance.

Issue #6425, opened 2026-08-20
solidity: a Yul function inside assembly{} is emitted as a contract operation, and the CALLS scan both over-fires on builtins and drops capitalised targets


cajasmota:
Precision counterpart to #6423 (recall) and #6424 (measurement). Both found by reading `internal/extractors/solidity/extractor.go`; INFERRED, not yet measured — which is what #6424 is for.

## Phantom operations from Yul

`functionRE` (`:62`) is `(?m)^[ \t]*function\s+NAME\s*\(` and it scans the **entire contract body**, function bodies included. A Yul function inside an `assembly { ... }` block sits at line start and matches:

```solidity
function withdraw() external {
    assembly {
        function helper(x) -> y { y := add(x, 1) }
    }
}
```

`helper` is emitted as a `SCOPE.Operation` named `Contract.helper` — an entity that does not exist in the Solidity surface at all.

Brace *balance* is not the problem: `extractBracedBody` (`:597`) counts depth correctly and skips comments and strings. The problem is that **member scanning is depth-blind** — it looks anywhere inside the body rather than only at brace-depth 1. The depth machinery already exists in `findStateVariables` (`:745`), so the fix is to reuse it rather than build it.

**No test in the package mentions `assembly`** — grep returns nothing.

## CALLS noise from denylist gaps

`solidityKeywords` (`:88-107`) omits `returns`, `type`, `catch`, `try`, `unchecked`, `super`, `this` and `abi`. So:

- `type(uint256).max` emits `CALLS → type`
- `abi.encode(...)` emits `CALLS → abi.encode`

Eight strings in a map.

## And the opposite error, in the same pair of regexes

`callBareRE` (`:84`) requires a **lowercase** first character and at least two characters. So every capitalised bare call is dropped — `SafeMath(...)`, library-style `Address(...)` — and so is every single-character call `f(`.

That asymmetry is worth stating plainly: the same call scan is simultaneously **over-firing** on language builtins and **under-firing** on capitalised and one-character targets. A precision fix that only adds denylist entries will make the first number look better while leaving the second untouched.

## Sequencing

`docs/coverage/by-language/solidity.md:34` already marks call-line precision as `partial`, so this is a known-soft area rather than a regression. Do #6424 first: without a fixture, "precision improved" is an assertion, and the over-fire and under-fire cases move the same aggregate number in opposite directions — so a single count cannot even tell you whether a change helped.

cajasmota:
[...]

Issue #6423, opened 2026-08-20
solidity: a contract with base-constructor args (contract X is ERC20("n","s")) matches nothing, so the contract and every member inside it silently vanish


cajasmota:
**Severity: silent whole-contract loss.** Found by reading `internal/extractors/solidity/extractor.go:44-86`; not yet measured on a corpus (see the fixture issue this refs).

`contractRE` (`:56`) renders the inheritance list as `[A-Za-z_][A-Za-z0-9_,\s]*` and then requires `{`. That character class **cannot contain `(`**. So:

```solidity
contract MyToken is ERC20("My Token", "MTK") {
```

matches nothing — and because `findContracts` only walks matched contract bodies, **the contract and every function, event, modifier and state variable inside it vanish from the graph.** Not degraded: absent.

This is the standard OpenZeppelin shape — base-constructor arguments in the `is` clause are ordinary Solidity, not an edge case.

There is a tell that this was believed to work: `:183-186` contains code stripping `(...)` from parent names, **a defence that can never fire**, because the regex never matches such a contract in the first place.

## Three more recall gaps in the same scan, same cause

The member scan only recognises what its regexes name, and only inside a contract body:

- **Constructors, `receive()` and `fallback()` emit nothing.** `functionRE` (`:62`) requires the literal `function` keyword, and `constructor`/`receive`/`fallback` sit in the denylist at `:102`. In OpenZeppelin-style code the constructor is where the wiring happens, so all of it is invisible.
- **File-level (free) functions, `struct`, `enum`, `error`, and `type X is Y` are never emitted at all** — there is no code path for them anywhere in the file.
- **Modifier *usage* is never linked.** `declBody` (`:663`) jumps from the parameter list's `)` to the body's `{`, so `onlyOwner` in a signature produces no edge. Modifiers are declared and contained, never connected to the functions that use them.

## Fix shape

`contractRE`'s inheritance group needs to tolerate balanced parentheses (or match the head and hand the tail to the existing paren-aware scanner — `declBody` already does depth-aware skipping, so the machinery exists). The member gaps are additional regexes reusing `declBody`.

**Do not fix this without a fixture.** There is no `.sol` fixture in `testdata/fixtures/` and no measured Solidity precision/recall anywhere in the tree, so today a fix is unfalsifiable — and a test built only from the shapes named above would pass while leaving the rest of the class broken. Land the fixture first.

[...]

Issue #6231, opened 2026-08-07
AGENTS.md claims golden fixtures hold 100% must-have recall on every PR; 14 of 20 are red and 2 do not index at all


cajasmota:
`AGENTS.md:13` states:

> Quality fixtures in `internal/quality/golden/`; must hold 100% must-have recall on every PR

**They do not, and did not on `55c7fc7a5`.** Measured while gating the falcon/cherrypy bare-class patterns (#6152), across all 20 fixtures:

- **14 of 20 exit non-zero on baseline** — `csharp-hangfire` 0/5, `python-rq` 0/3, `python-dramatiq` 0/4, and others.
- `groovy-grails-mini` and `swift-swiftui-mini` **fail to index at all** (rc=1), before and after any change.
- Corpus totals sit at entities 229/258 (88.8%), relationships 90/121 (74.4%).

Entirely pre-existing. #6152 moved one fixture *up* (`python-django-mini` 82.1% → 92.9% entities, 25% → 58.3% relationships) and left nineteen byte-identical, so nothing here was introduced by it.

## Why this matters beyond the numbers

A gate documented as enforcing, which does not enforce, is worse than no gate. It reads as a guarantee in `AGENTS.md` — the file contributors are pointed at — so a change that drops recall looks safe, and a change that *raises* recall gets no credit. Neither signal is available today.

This is the same shape as several defects found this week: a check that looks like coverage and is not. #6223 (timeout tests that never exercised a timeout) and the `TestLabelStable` canary blind to the platform it guarded are the same class.

## What a fix needs

Decide which of these the sentence should become, and make the code match:

1. **Make it true** — fix the 14 red fixtures, wire the recall check into a job that runs on every PR, and let it fail the build. This is the version the sentence currently promises.
2. **Make it honest** — record the current per-fixture baseline as a ratchet: recall may not *drop* below the recorded figure, and raising it updates the record. This is cheaper, catches regressions, and does not require fixing 14 fixtures first.
3. **Delete the claim** and say what actually runs.

Option 2 is the recommendation: the two fixtures that fail to index are a real defect worth their own issue, and a ratchet gives a working gate today rather than after that defect is fixed.

Whichever is chosen, the two non-indexing fixtures (`groovy-grails-mini`, `swift-swiftui-mini`) should be split out — a fixture that cannot be indexed is not a recall problem.

Found while measuring #6152. Refs #6152, #6223

cajasmota:
[...]

PR #6431, opened 2026-08-20
test(#6424): the first graded Solidity fixture — 52.8% entity recall, and all seven suspected defects confirmed by measurement

### LINKED ISSUE #6424: solidity has no fixture and no measured precision/recall, so every claim about its accuracy is unfalsifiable
**There is no `.sol` fixture in `testdata/fixtures/` and no measured Solidity precision or recall anywhere in the tree.** `testdata/fixtures/` covers 27 languages; solidity is absent — `internal/extractors/file_anchored_rels_guard_test.go:42` already states this.

The only Solidity number on record (`docs/coverage/registry.json`, rendered at `docs/coverage/detail/lang.solidity.core.md`) is an *indexer* measurement, not an extractor one: on the OpenZeppelin contracts package it produced 61 interfaces, 71 contracts and 45 libraries, and pre-#6138 the file-fold silently deleted 42 of the 61 interfaces. That says nothing about whether the extractor found the right things in the first place.

Meanwhile `docs/coverage/by-language/solidity.md:34` shows core extraction as green and verified, with call-line precision and import resolution both marked partial.

## Why this is the prerequisite, not a nice-to-have

#6423 and #6424 name seven specific defects derived by **reading the regexes**. Each is plausible and each is INFERRED — none is measured. Without a fixture:

- a fix is unfalsifiable, and
- a test written from the seven named shapes would pass while leaving the rest of the class broken, which is this repo's active vacuous-fixture failure mode.

It also decides a strategic question. If a corpus run shows recall misses **concentrated** in the named constructs, the regex repair is right. If misses are **distributed**, the approach itself is the problem and a grammar becomes justifiable. Right now nobody can tell which, so any argument about parsing technique is unfalsifiable too.

## Scope

OpenZeppelin is already the reference corpus in the registry note, is MIT-licensed, and is therefore safe to commit excerpts from — unlike some corpora in use here. A graded fixture in the `vbnet-mini` shape (`internal/quality/golden/`) with `expected.json` and a forbidden list is the model to copy; that one was built to be **proven graded and proven non-vacuous**, which is the bar.

The fixture should deliberately include the shapes currently believed broken — a contract with base-constructor args, a constructor, a `receive`/`fallback`, a free function, a `struct`/`enum`/`error`, a modifier usage, and an `assembly` block — so the grade *starts* red on them and the improvement is visible as a number rather than asserted.

Refs #6423, #6424.

cajasmota:
Fixes #6424. Refs #6423, #6425.

Solidity had **no fixture and no measured accuracy anywhere in the tree**. Every claim about that extractor — including the seven defects filed as #6423 and #6425 — was inferred from reading its regexes. This makes them falsifiable.

## The numbers, and what each one is

```
entities:      19 / 40 expected  (47.5%)   [extracted total: 41]
relationships:  8 / 13 expected  (61.5%)   [extracted total: 108]
forbidden hits: 0
```

**47.5% is not the extractor's recall, and must not be quoted as if it were.** The must-have set was chosen to cover the shapes #6423/#6425 name, so it deliberately over-samples the broken ones. The honest figure over the same source is:

```
real declarations in src/:              52
currently extracted:                    31   (recall = 59.6%)
of those 31, asserted as must-haves:    19   (12 working ones are not asserted)
missing:                                21   (all 21 are asserted must-haves)
```

(31 = the 41 entities the run reports, minus 2 `Module` carriers, 4 `X.sol` file carriers, 3 `SCOPE.External` nodes, and the 1 phantom `Vault.helper` from #6425.)

So the fixture's rate tracks **the defect classes** and the baseline floor tracks **when they get fixed**; neither is a coverage statistic for Solidity. A targeted fixture is legitimate — presenting its number as the extractor's is not, which matters more than usual in a PR whose whole thesis is measured-not-read. `NOTICE.md` now says this on its face.

**All seven suspected shapes are CONFIRMED broken by measurement rather than by reading.** The worst is worse than predicted:

| Shape | Measured |
|---|---|
| base-constructor args (`contract X is ERC20("n","s")`) | **the whole file is lost** — yields a file carrier and 1 IMPORTS edge, nothing else. 5 entities + 2 edges gone |
| constructor / `receive` / `fallback` | 4 further entity misses (`MyToken.constructor` is inside the lost file above) |
| free function / struct / enum / error / `type X is Y` | 12 entity misses, asserted at **both** file and contract level |
| modifier usage | 2 edge misses — same-body and inherited-cross-file |
| Yul `function` in `assembly{}` | phantom `SCOPE.Operation` present in `graph.json` |
| CALLS over-fire | `SCOPE.External` nodes literally named `type` and `encode` |
| CALLS under-fire | `IERC20(token).transfer(...)` yields only `transfer` |

**Every `error` declaration in `src/` is asserted** — all six of them, not a sample. An earlier revision of this PR asserted only two and left four (`Ownable.OwnableUnauthorizedAccount`, `Ownable.OwnableInvalidOwner`, `ShortStrings.InvalidShortString`, `Vault.VaultEmpty`) out of the must-have set while asserting `StringTooLong` with the identical shape. That was inconsistent, and it let a partial `error` fix look complete. Adding them moved the must-have denominator 36 → 40; `entity_found` is unchanged at 19, so the ratchet floor did not move.

**A bonus finding the reading missed:** `Vault.deposit --[CALLS]--> computeFee` *is* emitted, but **unresolved**, because the free function is never extracted. So the free-function gap costs **resolution**, not merely an entity count — it produces a dangling edge, not a missing one.

## How a deliberately-red fixture was made to work

No gate was weakened and no marker invented. The recorded floor in `internal/quality/golden/baseline.json` already supports this: `scripts/quality/ratchet.py` gates each fixture against its found figures, and a floor below expected is an ordinary existing state — `csharp-aspnet-core-mini` already sits at 4/5.

Deleting a must-have to close the gap does not go unnoticed: `TestBaselineRecordedCountsAreSane` would still pass, but `ratchet.py:310-316` fails with `entity_expected changed 40 -> 39 — the fixture's expectations were edited`. That is a re-record prompt rather than a wall, but it forces the deletion into a baseline diff instead of letting it happen silently.

**A real gap was found and reported rather than worked around.** There is **no equivalent channel for precision**: `ratchet.py:293-295` treats any forbidden hit as "always fatal", and `Evaluate` counts them unconditionally — `nice_to_have` does not apply to the forbidden list. So the #6425 precision entries (the Yul phantom, the `type`/`encode` over-fires) **could not be committed without breaking the gate**. They are held verbatim in `NOTICE.md` to be added when #6425 lands.

That asymmetry is worth knowing: this repo can ratchet recall but cannot ratchet precision.

## Proven graded, proven non-vacuous

#6273 (fixtures never graded) and #6283 (grading a stale binary) are why this matters.

- **Graded:** the grader's own output lists the 21 missing entities and 5 missing relationships **by name** — including all four newly-added `error` entries — so it read `expected.json`; it did not merely find the files on disk.
- **Non-vacuous:** the `wantFixtures` 22→21 mutant **died**; the forbidden-list mutant **died** (`forbidden hits: 0` → `1`), which is also what double-confirms the Yul phantom is really in the graph.

## Correction: `to_bare_name` is **not** vacuous for Solidity

An earlier revision of this PR claimed, in bold, that the pipeline folds *every* bare `CALLS`/`EXTENDS` target into a `SCOPE.External` entity, so `vbnet-mini`'s `to_bare_name` idiom is vacuous here. **That was false**, and it was inverted guidance for the next fixture author. Counted in the emitted `graph.json`:

* **3** entities are `SCOPE.External` — `Context`, `encode`, `type`.
* **26** relationships still carry a **raw-string `ToID`** — `computeFee`, `_msgSender`, `abi.encode`, `ShortString.wrap`, `add`, `wrap`, `getStringSlot`, `unwrap`, `uint256`, `mstore`, the four `IMPORTS` paths, and more. The resolver prints it: `unmatched=29`.

Falsified by measurement, not by re-reading: rewriting the `Vault.deposit --[CALLS]--> computeFee` entry to `"to_bare_name": "computeFee"` moves relationships **8/13 → 9/13** — it matches.

So **some** targets fold and most do not. The two the fixture actually asserts with `to_name` + `"to_kind": "SCOPE.External"` — `Context` and `ERC20` — are ones that *do* fold, so those two entries are correct and are unchanged. For an **unresolved** target there is no entity to match `to_name` against, and `to_bare_name` (compared against the raw `ToID`) is the correct and only way to assert the current shape. The `computeFee` entry is deliberately **left** on the resolved form: that callee *should* be extracted (#6423), and asserting the unresolved bare name would record the defect as the intended shape. The generalisation came from a single folding target; one target is not a rule.

## Flagged, not fixed

`docs/coverage/by-language/solidity.md:34` still shows core extraction as green and verified. **59.6% contradicts that**, and it is user-facing. That file appears generated from `registry.json`, so it was left alone rather than hand-edited — filing separately.


cajasmota:
## Coordinator mutation check — and it corrected my own conclusion

I mutated a relationship assertion's `to_kind` from `SCOPE.External` to `SCOPE.Component` (line 335, the `Context` external-parent edge) and ran `./internal/quality/`.

**It survived — REALEXIT=0.** My first read was that this exposed a coverage gap in the fixture. It does not. I ran the wrong harness.

`go test ./internal/quality/` **does not grade fixtures**. It checks structure and recorded counts — `wantFixtures`, baseline sanity. Grading happens through the grader binary, which is why the reviewer's own mutants (delete a found entity, corrupt a found id) died with the **grader's** exit 2, while the `wantFixtures` mutant died in `go test`. Two different harnesses catching two different things, and I conflated them.

**What that does establish is worth recording.** `.github/workflows/quality.yml` is `workflow_dispatch` only — its own header says so — and grepping `.github/workflows/` for `run.sh`, `ratchet.py` or `grafel quality` returns **nothing**. So **no always-on CI job grades any fixture, including this one.**

That is pre-existing and already known — the header cites #6231, now closed — but it changes how this PR should be read. This fixture is *provably graded* when someone runs the grader, and the review verified that by running it. It is **not** continuously enforced. A corrupted `to_kind`, or extractor drift, would sit unnoticed until the next manual run.

That is not a reason to hold this PR — an unenforced measurement is still infinitely better than the zero measurement Solidity had an hour ago, and the recorded floor means a later run *will* catch regression. But "the fixture exists" and "the fixture is defended on every PR" are different claims, and only the first is true today.

### Verified state

`expected.json` restored by `cp`, `shasum addaf7d6…` identical. No `git checkout --` used.

The corrected figures read honestly now: **19/40 (47.5%)** on the must-have set, with the caveat stated in three places that this is a **red-enriched sample** and the extractor's recall over the same source is **31/52 = 59.6%**. Adding the four `error` declarations moved `entity_expected` 36→40 while `entity_found` stayed at 19, so the ratchet floor did not shift — the fixture got harder without lowering its own bar.

PR #6434, opened 2026-08-20
fix(#6354): gen never pruned by-language pages and the docs gate could not see untracked files, so orphans shipped

### LINKED ISSUE #6354: gen never prunes by-language pages and the docs gate cannot see untracked files, so orphaned pages ship and new ones can slip through
Found while reviewing #6351.

## Two one-directional gaps that compose

`go run ./tools/coverage gen` writes `docs/coverage/by-language/<slug>.md` for every derived language. **It never deletes a page whose slug no longer exists.** And the workflow's check is:

```yaml
git diff --exit-code docs/coverage/
```

which **does not see untracked files**.

Composed, that means:

- a page for a slug that has gone away survives a clean `gen` indefinitely, and
- a newly-emitted page is caught only *indirectly*, via the `summary.md` delta that references it — never on its own.

## Already shipped, on `main` today

Two committed orphans:

- `docs/coverage/by-language/avro.md`
- `docs/coverage/by-language/jsonschema.md`

Both slugs are in `extractorNonLanguageFormats`, so neither is derived; neither has a registry record. They correspond to nothing, they survive a clean `gen`, and nothing reports them.

## Why it matters beyond two stale files

The published matrix is the thing users read to decide whether grafel covers their stack. A page for a language we do not derive is a claim we are not making anywhere else — the inverse of #6332, where a language we *do* support was missing a reviewed label.

The untracked-file half is the sharper of the two. During the #6351 work, `gen` with a stray extractor directory present produced:

```
 M docs/coverage/summary.md
?? docs/coverage/by-language/testdata.md
```

`git diff --exit-code` fires on the first line, not the second. So the gate happened to catch that case, but only because the new page was also referenced from `summary.md`. A page emitted *without* a summary reference would be invisible to CI.

## Shape of a fix

1. **Prune in `gen`**: after writing, remove any `by-language/*.md` whose slug is not in the derived set. Then the two orphans disappear on the next run and the diff gate reports it.
2. **Make the gate see untracked files** — `git status --porcelain docs/coverage/` non-empty, or `git add -N docs/coverage/` before the diff. Cheap, and closes the half that reasoning alone would not have found.

Both are small. (1) without (2) leaves the gate half-blind; (2) without (1) turns the existing orphans into a permanent red. They should land together.

## Acceptance

- The two orphans are gone and `gen` is idempotent afterwards.
- Deleting a language directory makes CI red on *that* PR, naming the stale page.
- A hand-created `by-language/zzz.md` that nothing references makes CI red — this is the case that fails today.

Refs #6332, #6347, #6351

cajasmota:
Fixes #6354.

Two mechanisms, one PR, because they are genuinely coupled: prune alone leaves the gate half-blind, and the gate alone turns the two orphans already in the tree into permanent red on the very next PR.

## What was broken

**`gen` never pruned.** `tools/coverage/generate.go` held two write-only emit loops; the only `os.Remove` calls in the package were tmp-file cleanup. A by-language page whose language stopped being derived was never deleted.

**The gate could not see untracked files.** `.github/workflows/coverage-docs.yml:55` used `git diff --exit-code docs/coverage/`, and `git diff` does not report untracked files — so a newly-generated orphan slipped through silently.

**Two orphans were already shipped:** `avro.md` and `jsonschema.md`, both slugs listed in `extractorNonLanguageFormats`, so neither is a derived language. Both are deleted here.

## Prune scope, and why it is narrow

`generate` accumulates the pages it emits across **both** loops, then prunes `docs/coverage/by-language/` — a write-then-prune contract. A file is deleted only if all three hold: it ends in `.md`, its slug is absent from the keep set, **and its leading bytes equal `doNotEditMarker`**.

The marker is checked by reading exactly `len(doNotEditMarker)` bytes from the head of the file, **not** `strings.Contains` over the whole body — prose that quotes the banner deeper in a page is not a generator stamp, and a `Contains` check would delete a hand-written page that merely mentions it.

**No unmarked file lives there today** — confirmed by checking the head of every file in `by-language/` and `by-category/`; all carry the marker. So the marker scope costs nothing now; it exists so a future hand-written page survives.

**Deliberately not fixed:** `by-category/` and `detail/` have the identical write-only shape. Scoping matched the issue's acceptance criteria rather than widening; the class is worth its own issue.

## The trap this PR had to avoid

`gen_test.go` already asserted that two pages **do not exist** — and those assertions pass because the emit loop never writes them. **They would still pass with the prune deleted.** They were never coverage for pruning.

So the new tests create a stray page and assert `gen` removes it, and three mutants pin the three conditions independently:

| # | Mutation | Exit | Killed by |
|---|---|---|---|
| M1 | delete the `pruneGenerated` call | 1 | both new tests |
| M2 | ignore the marker scope — prune every `.md` | 1 | a must-survive unmarked page |
| M3 | drop the `keep[slug]` check — prune everything generated | 1 | two must-survive real pages |

M2 and M3 matter as much as M1: a prune that deletes too much is worse than one that deletes nothing, and each guard is now pinned by a test that fails for its own reason.

## Verification

New tests RED before (`REALEXIT=1`, both failing on the orphan surviving), green after. `go test ./tools/coverage/` → 0 · `go vet` → 0 · `gofmt -l` → empty.

`go run ./tools/coverage gen` deletes exactly `avro.md` and `jsonschema.md` and nothing else, and a second run produces no further change — **idempotent**.

The gate fix was exercised locally rather than assumed: with an untracked stray page present, `git add -N docs/coverage/` followed by `git diff --exit-code` exits **1** and **names the file**. That is the case that is green today.

## One behavioural subtlety, stated so it is not read as a gap

An untracked stray page that carries the marker is now pruned by `gen` *before* the diff step, so CI stays green — correctly, since nothing was ever committed. A **committed** orphan (the avro/jsonschema shape) is pruned, which surfaces as a deletion in `git diff` → red, naming the page. That is the intended behaviour, not a hole.

`test.yml` carries only the `validate` step, not `gen` + diff, so `coverage-docs.yml:55` was the sole gate needing the change — confirmed by grepping every workflow for `docs/coverage`.


cajasmota:
## Independent review: REQUEST-CHANGES. A confirmed false-deletion path, on the filesystem this project is developed on.

**`gen` can delete a live page it wrote seconds earlier.** On a case-insensitive filesystem — macOS/APFS, i.e. this repo's dev machine — if `by-language/python.md` exists on disk as `Python.md`, `renderToFile`'s temp+rename lands in the **existing** entry and the name stays `Python.md`. Prune then computes slug `Python`, misses `keep["python"]`, sees the marker, and deletes it.

Reproduced by execution: `mv python.md Python.md; go run ./tools/coverage gen` → **REALEXIT=0**, `git status` → ` D docs/coverage/by-language/python.md`. Silent, exit zero, page gone.

**Ubuntu CI is case-sensitive, so CI would never have caught this.** A macOS contributor loses a page. That is exactly the class of defect I asked the reviewer to hunt — a prune that deletes too much is the catastrophic direction — and it turned up in the one environment our gate cannot see.

### The PR's own headline safety claim is unpinned

I wrote that the marker is checked as **exact head bytes, not `strings.Contains`**. True of the implementation — and **no test distinguishes them**. A `strings.Contains`-over-the-whole-file version **survives** the suite, because no fixture has the marker anywhere except byte 0. The hand-written fixture is `"# Hand-written notes"`, which tests nothing about the marker.

The fixture that makes the claim real is a must-survive page that **quotes the marker on line 3**. Without it, the sentence in this PR body is an assertion about code nobody can break.

**The short-file branch is unpinned the same way.** A mutant accepting any *prefix* of the marker survives — and that mutant **deletes every zero-byte and truncated `.md`** in the directory. The `io.EOF` handling exists precisely for that case. Today's behaviour is correct (both file shapes were verified to survive); nothing keeps it correct.

Also: a `*.md` **symlink is deleted** (target unharmed) because `IsDir()` is false for symlinks and `os.Open` follows the link. And a dangling or directory symlink **aborts `gen` mid-run** — fail-safe, nothing deleted, but the abort lands after by-language is written and before by-category, leaving `docs/coverage/` half-regenerated.

### What the review confirmed as genuinely sound

The adversarial zoo mostly held: a page with **no marker**, a leading **BOM**, a **truncated** marker, a **zero-byte** file, a **directory** named `subdir.md`, and a `notes.txt` all survived — only `avro.md` and `jsonschema.md` were pruned. CRLF is handled because the marker excludes its own newline.

**Keep-set correctness is exact**: filenames are `s+".md"` and the keep keys are the same `s`, so there is no alias or slugify skew, and `multi`/empty slugs `continue` before both write and record.

**The gate change works and was checked, not assumed**: YAML parses, step order is gen→verify with verify last so the intent-to-add index never reaches a later step, the pathspec is confined to `docs/coverage/`, the no-change case exits 0, and an untracked stray exits 1 **naming the file**.

**The scope claim is true** — `by-category` and `detail` remain write-only and untouched, so that follow-up is real rather than an excuse.

Two of the three original mutants reproduce independently.

cajasmota:
## Fixes pushed as `287c82468`. Marking ready.

**The data-loss path is closed with a skip-only rule**, and the reasoning for choosing it over the alternative is worth recording. `keptFold` checks the exact key first, then folds — so it **cannot widen deletion by construction**, it only ever adds protection. Recording the exact written path would have changed what the keep set *means* and needed auditing against every emit site.

It also covers a case path-recording would not: a page a **previous** run wrote under a colliding name, already on disk before this run's emit loop.

**The cost is stated rather than hidden:** on a case-sensitive filesystem, a genuinely orphaned `Foo.md` sitting beside a derived `foo.md` now survives. That is the safe side of the trade — an orphan that lingers is a stale page; a false deletion is lost work.

The symlink findings collapse into one guard: `e.Type().IsRegular()` makes symlinks non-candidates, which makes the dangling/directory-symlink read-error path unreachable for them. For a **regular** file that still fails to open, the deliberate choice is to abort — that is a genuine environment fault worth surfacing — **except ENOENT**, which now skips, because the entry vanishing between `ReadDir` and `Open` means there is nothing left to prune.

### The two claims that were unbacked are now pinned

Both mutants that survived the review now die:

| Mutant | Before | After | Killed by |
|---|---|---|---|
| `strings.Contains` over the whole file | SURVIVED | **DIES** | a page quoting the marker in its body |
| accept any prefix of the marker | SURVIVED | **DIES** | zero-byte and truncated pages |

Those are the fixtures that make the prose true. Before, "exact head bytes, not `Contains`" was a description of code nobody could break.

### My own mutant — the opposite direction

The fix round proved the guard does not delete too much. I checked it had not stopped deleting at all: `keptFold` forced to `return true`, protecting everything.

| Mutant | `go vet` | Test | Result |
|---|---|---|---|
| `keptFold` always true — nothing is ever pruned | 0 | **REALEXIT=1** | **KILLED** — `TestGenPrunesOrphanedByLanguagePages`, `TestGenPrunesPlaceholderPageWhenExtractorDirDisappears` |

That matters because every other mutant here pushes the same way — *does it delete too much?* A narrowing fix can pass all of those by simply never deleting anything. It doesn't.

Restored by `cp`, `shasum 412289da…` identical.

### Re-verified on the real tree

Case-collision reproduction with three symlinks (good, dangling, directory) present: `gen` exits 0, **`Python.md` survives**, all three symlinks survive, the symlink target is unharmed, and `git status` shows no deletion. Restoring `avro.md` and `jsonschema.md` and re-running prunes **exactly** those two — 42 files to 40 — and stays idempotent.

Issue #6422, opened 2026-08-20
proto: fileContainsRel addresses messages as operations, so a message sharing a name with an rpc gets its file CONTAINS silently reparented onto the rpc


cajasmota:
Found while confirming #6419's new message address space actually resolves. **Pre-existing** — `fileContainsRel` is untouched by that PR — and demonstrated empirically rather than reasoned about.

`fileContainsRel` addresses **every** top-level entity through the operation form:

```
scope:operation:method:proto:<file>:<Name>
```

That is correct for a service or rpc. It is wrong for a message or enum, and it only *appears* to work because non-colliding names have nothing to collide with.

## The demonstration

In a fixture containing `service S { rpc User(User) returns (User); }` — an rpc and a message sharing a name, which proto permits — the resolver was run end to end (real extractor → real `ComputeID` → real `BuildIndex` → real `resolve.References`). Result:

- The **file → message `User` CONTAINS** edge resolves to `SCOPE.Operation/endpoint "User"` — **the rpc**, not the message. It becomes a duplicate of the existing service→rpc CONTAINS.
- The message `User` therefore has **no inbound CONTAINS at all**. After #6419 it is reachable only through the new REFERENCES edge; before #6419 it was reachable by nothing.

So the same entity is addressed through two address spaces that disagree with each other: type references now use the schema form (`scope:schema:message:…`) and bind correctly, while the containment edge still uses the operation form and binds the wrong entity.

Non-colliding messages (`Req`, `Resp`) are unaffected — the operation-form ref binds them because no operation shares the name. That is why this has gone unnoticed: **the bug is invisible until a name collides**, and then it silently reparents an entity rather than failing.

## Fix

`fileContainsRel` should choose a kind-appropriate ref per entity type — schema form for message and enum, operation form for service and rpc. That also restores the message's file CONTAINS in the collision case.

**Do not just suppress the duplicate edge.** The duplicate is a symptom; the defect is that a message is addressed as an operation. Suppressing it would leave the message with no containment at all, which is the worse half of the current behaviour.

## Also worth folding in

`messageTypeRef` (added in #6419) omits the `filepath.ToSlash` normalization that every `Build*StructuralRef` helper applies. `lookupStructural` calls `normalizePath` defensively so it resolves anyway, but it is an inconsistency with the surrounding convention and is one line to align. Cheap to do in the same change.

## Testing note

[...]

Issue #6367, opened 2026-08-19
file-anchored relationship sites: graphql FEDERATES is the last KNOWN OFFENDER — misanchored onto the synthetic file container, still inferred and never measured


cajasmota:
#6365 fixed verilog and astro and added a guard. Reviewing **that guard's allow-list** against the actual resolution rule found six entries blessed on a criterion that does not hold. They are labelled `KNOWN OFFENDER (#6298)` in `internal/extractors/file_anchored_rels_guard_test.go` so the guard stays green; none is fixed.

## The rule, stated operatively

`internal/resolve/refs.go` has **no path→entity index**. A path-valued `FromID` resolves **iff some emitted node carries that exact string** as `Name`/`QualifiedName` (or the package emits an `extractor.FileEntity`).

`looksLikeSourceFilePath` (`refs.go:2729`) resolves nothing — it routes to `DispositionDynamic`, which **suppresses the bug-extractor metric while the edge stays dangling**. That is why astro's dangle survived undetected until #6365. Anything whose extension is missing from `sourceFileExtensions` (`refs.go:2752-2767`) misses even that suppression and counts as `DispositionBugExtractor`.

The original allow-list criterion — "the owning record is genuinely file-scoped" — is a *conceptual* test. The operative one is stricter, and six entries fail it.

## The six

All **INFERRED from reading the site; none measured.** That distinction is the point of this issue.

1. **`graphql.go` FEDERATES** — misanchored. The site's own comment (`graphql.go:366-368`) says the edge is *"from the extending stub"*, but `FromID` is `filePath`, so it lands on the synthetic file container and every stub in the file merges onto one node.
2. **`proto.go` `fileContainsRel` CONTAINS** (3 call sites) — dangling. No proto node carries the *containing* file's path; only the *imported* path is path-named. `proto.go:260`'s sibling service→rpc edge correctly leaves `FromID` empty.
3. **`hcl/relationships.go:129,140` CONTAINS** — dangling. The HCL file component's `Name` is the **basename**, so a path-valued `FromID` resolves only by accident, for a root `main.tf`.
4. **`hcl/extractor.go:562` DEPENDS_ON** — dangling *and* the wrong owner (the resource/module record).
5. **`bicep/extractor.go:394` DEPENDS_ON** — dangling unconditionally, wrong owner, and `.bicep` is absent from `sourceFileExtensions` → **live `DispositionBugExtractor`**.
[...]

PR #6435, opened 2026-08-20
fix(#6367): hcl CONTAINS/DEPENDS_ON dangled at any nested path and only resolved at the repo root by accident

cajasmota:
The **hcl arm** of #6367. Refs #6367 — five arms remain (graphql, proto, bicep, dockerfile, swift).

## The issue's inferred claim was right in general and wrong in detail

Every entry in that allow-list is self-labelled *"INFERRED from the site, NOT measured."* Measuring hcl produced a split the inference could not have predicted.

Driving extractor → id-stamp → `ResolveImports` → `ReferencesEmbedded` on one fixture at **two paths**:

| path | CONTAINS | DEPENDS_ON |
|---|---|---|
| `infra/envs/prod/main.tf` | **5 of 5 DANGLING** | **2 of 2 DANGLING** |
| `main.tf` (repo root) | 5 of 5 **resolve** — onto the file component, i.e. the right node **by accident** | **2 of 2 MISANCHORED** |

The split is the file component's `Name`: `relationships.go` sets it to the **basename**, so a path-valued `FromID` only matches when the path *is* its own basename. A repo whose Terraform sits at the root looked fine; the same code one directory down dangled.

That is worth stating plainly — **a fixture written at the repo root would have shown this as working.** The new test measures nested *and* root precisely so the accidental resolution cannot hide a regression.

`DEPENDS_ON` is wrong in **both** columns, because `parseDependsOnTuple`'s records are appended to the resource/data/module entity — so even the resolving spelling moves the edge off its owner.

**After: 0 of 7 on both paths.**

## The `{2}` entry is deleted, not decremented

`emitFileLevelRelationships` had **exactly two** `CONTAINS` sites — the `locals` per-key branch and the default per-block branch. Both fixed, so the allow-list entry is removed entirely.

This matters because the guard's count check is **cardinality-only, not identity**: a `{2}` entry blesses *any* pair of same-kind sites in that function, so a half-fix leaving two sites of a different shape would have stayed green. Mutant 2 below is exactly that half-fix, and it is caught **because the entry is gone**, not because a count moved.

The other three `FromID: path` sites in that function are `IMPORTS`, exempt under the #120 convention, and are untouched.

## A pre-existing test encoded the bug

`relationships_test.go:137` asserted `CONTAINS FromID == "main.tf"` — it pinned the defect as the contract. Inverted to `== ""`.

Three now-stale claims in the guard's own header were also corrected: the `count>=2` live-key tally drops from four keys to one (swift), and the "blessed six" note now records that two are fixed.

## Mutants

| # | Mutation | hcl suite | guard |
|---|---|---|---|
| 1 | restore `FromID: fromPath` at `parseDependsOnTuple` | REALEXIT 1 | REALEXIT 1 |
| 2 | restore `FromID: path` at **one** of the two CONTAINS sites (half-fix) | REALEXIT 1 | REALEXIT 1, reporting `1 site(s)` |
| 3 | re-add the deleted allow-list entry with the fix in place | — | REALEXIT 1, `matches no site any more` |

Mutant 3 confirms the guard fails on a **stale** entry too, which is why deleting rather than decrementing was mandatory.

## Verification

`go test ./internal/extractors/hcl/` → REALEXIT 0 · `./internal/extractors/ -run FileAnchored` → REALEXIT 0 with both hcl entries deleted · `go vet` → 0 · `gofmt -l` → empty.

**No shared machinery was touched** — the graphql, proto, bicep, dockerfile and swift entries are byte-identical. proto in particular was left alone deliberately, since it overlaps open work in #6422.

### DIFF
--- a/internal/extractors/file_anchored_rels_guard_test.go
+++ b/internal/extractors/file_anchored_rels_guard_test.go
@@ -34,7 +34,8 @@
 // that exact path string exists" is. The first round of this allow-list used the
 // conceptual criterion and, on the strength of it, blessed six entries across
 // five languages that are real defects — they are re-labelled KNOWN OFFENDER
-// below rather than fixed, because each needs its own measurement.
+// below rather than fixed, because each needs its own measurement. #6367's hcl
+// arm has since measured and fixed two of the six.
 //
 // WHY A SOURCE SCAN AND NOT A RUNTIME ONE. The runtime form of this check —
 // drive every registered extractor and inspect the emitted records — needs a
@@ -229,13 +230,14 @@
 //     svelte:extractReactiveStatements:USES {2},
 //     hcl:emitFileLevelRelationships:CONTAINS {2}, and
 //     knownInvisibleOffenders["swift:extractTargets:DEPENDS_ON"] {2}. The two
-//     svelte entries went away with the #6366 fix; the remaining two are hcl and
-//     swift (RE-MEASURED 2026-08-21). What a "?"
+//     svelte entries went away with the #6366 fix and the hcl one with #6367;
+//     the only remaining count>=2 entry is swift (RE-MEASURED 2026-08-21).
+//     What a "?"
 //     key ADDS on top of that is collapsing ACROSS FORMS: a form-A site and a
 //     form-I site key identically, so one can be replaced by the other. The only
 //     "?" entry today (yaml:extractHelmHelpers:?) has count 1, so the
 //     across-forms half has no live consequence; the same-kind half is live at
-//     four keys right now, and it is where the next regression would hide.
+//     ONE key right now, and it is where the next regression would hide.
 //
 // This is a guard against the careless repetition of a known pattern, not a
 // proof. It is honest about being one.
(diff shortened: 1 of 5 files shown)


cajasmota:
## Independent review: APPROVE. The root-vs-nested claim is confirmed by execution, and it has consequences for the arms still open.

The reviewer reproduced the split rather than reading it. Restoring `FromID: fromPath` gives:

- `infra/envs/prod/main.tf` → `FROM = <UNRESOLVED:infra/envs/prod/main.tf>` — **2 dangling, 0 misanchored**
- `main.tf` → `FROM = file:main.tf` — **0 dangling, 2 misanchored**

So the claim holds both ways. **A root-only fixture would have shown this bug as working**, and that generalises: the four #6367 arms still open — graphql `FEDERATES`, proto `CONTAINS`, bicep `DEPENDS_ON`, dockerfile `USES` — carry reasons self-labelled *"INFERRED from the site, NOT measured."* They are genuinely at risk of the same mis-diagnosis. Whoever takes them should measure at a **nested** path, not the repo root.

**All sites are fixed, derived independently.** The package has exactly three `FromID` sites carrying a raw path; all three are fixed. Every other `FromID` in the package is a structural ref, not a path.

**The anchoring is correct rather than merely empty** — both assembly paths substitute the owner on an empty `FromID`, full and incremental, and the post-fix edges land on `resource:aws_s3_bucket.b` and `module:module.vpc` as intended. That was the question most worth asking, since dropping a field is only right if something else fills it on **every** path.

Two of the reviewer's own mutants are worth recording. The **mirror half-fix** (restoring `FromID` at the *locals* branch instead of the default branch) is killed — so both sites are independently pinned, not just the one the PR's own mutant used. And planting a **new unlisted site** on an unrelated CALLS edge trips the guard with `key "hcl:extractCalls:CALLS"` — which confirms deleting the allow-list entry **restored guard coverage for hcl**, rather than merely moving a number.

Also confirmed: the inverted `== ""` assertion in `relationships_test.go` is **not** vacuous — it fired 4 times under the half-fix mutant. That was my specific worry, since an assertion on an empty string is exactly the shape that passes when a field is simply never populated.

### Correcting this PR's body

I wrote that "the other **three** `FromID: path` sites in that function are `IMPORTS`". There are **two** — the module-source and provider sites. Body-only; no code or comment repeats the wrong number.

### Three low findings, being fixed before merge

The new test compares `Subtype+":"+Name` **labels** rather than IDs, so two records sharing both would compare equal and mask a misanchor — not reachable with this fixture, but the wrong thing to leave in a test whose whole purpose is anchoring. The CONTAINS pin rests **solely** on the nested path (for CONTAINS the owner *is* the file component, so the root subtest is structurally incapable of failing) — correct and intended, but if anyone later trims the fixture to one root path, that pin dies silently and nothing says so. And the guard header reflow left an orphan `// What a "?"` line mid-sentence.

Guard header numbers were verified against the file: every remaining `allowedFileAnchored` entry is `{1}`, and the only surviving `count>=2` key is swift.

cajasmota:
## Polish pushed as `56bc4279e`. Coordinator mutation check below.

The three low findings are closed, and one of them turned into a real robustness gain rather than a cosmetic fix: **dangling-vs-misanchored no longer splits on an `"<UNRESOLVED:"` string prefix.** It now keys on a genuine `byID` lookup, with labels retained only for the failure message. String-prefix matching on a sentinel is exactly the kind of assertion that keeps passing after the sentinel's format changes.

The comparison itself moved from `Subtype+":"+Name` to resolved **ids**, so two records sharing subtype and name can no longer compare equal and mask a misanchor.

And the fixture-path list now states what each path is load-bearing for: the **nested** path is the sole CONTAINS pin (the root subtest is structurally incapable of failing, since for CONTAINS the owner *is* the file component), the **root** path is the sole DEPENDS_ON-misanchor pin (nested only dangles), and removing either loses a distinct guarantee.

### My mutant — the site the earlier ones did not use

The fix round's mutant and the review's both restored `FromID` at the **default per-block** CONTAINS branch. I restored it at the **locals** branch instead, under the new ID-based assertion:

| Mutant | `go vet` | hcl suite | Guard |
|---|---|---|---|
| `FromID: path` restored at the **locals** branch only | 0 | **REALEXIT=1** — `2 dangling, 0 misanchored, of 5` at the nested path | **REALEXIT=1** — `key "hcl:emitFileLevelRelationships:CONTAINS" (1 site(s))` |

Two things worth drawing out. It reports **2** dangling where the default-branch mutant reported **3** — different sites, different counts, so both are genuinely pinned rather than one masking the other. And it fails **only** the nested subtest while the root subtest passes, which is direct empirical confirmation of the root-resolves-by-accident claim, now reproduced under the rewritten assertion rather than the old one.

Restored by `cp`, `shasum 47adb2bc…` identical. No `git checkout --` used.

This is the point of deleting the allow-list entry rather than decrementing it: with the entry gone, the guard catches a half-fix at **either** site. With a `{1}` left behind, one of these two mutants would have been blessed.

PR #6438, opened 2026-08-20
fix(#6423): solidity recall 19/40 to 40/40 — a contract with base-constructor args lost its entire file

### LINKED ISSUE #6423: solidity: a contract with base-constructor args (contract X is ERC20("n","s")) matches nothing, so the contract and every member inside it silently vanish

cajasmota:
Fixes #6423. Refs #6424, #6425.

The graded fixture that landed in `6b8d73859` made this measurable. The number moved from **19/40 to 40/40**.

| | before | after |
|---|---|---|
| entities | **19 / 40** (47.5%) | **40 / 40** (100%) |
| relationships | **8 / 13** (61.5%) | **13 / 13** (100%) |
| forbidden hits | 0 | 0 |
| strict grader exit | 2 | **0** |

`baseline.json` records `entity_found: 40`, `relationship_found: 13`. **`entity_expected` and `relationship_expected` are unchanged** — the floor rose to meet the bar, the bar did not move.

## The four defects

**1. `contract X is ERC20("n","s")` lost the entire file.** Two causes, and only the first was in the issue. The inheritance character class could not contain `(` — widened to `[^{};]*`. But the naive comma split was the other half: it yields the "parents" `ERC20(` and `)`. A new paren-aware `parseInheritanceList` handles that. The dead `(`-stripping code the issue flagged — a defence that could never fire — now actually runs.

**2. Constructors, `receive()` and `fallback()`** emitted nothing. New `constructorRE` and `specialFunctionRE`, folded into one member loop with functions.

**3. Free functions, `struct`, `enum`, `error`, `type X is Y`** were never emitted. Contract bodies are now masked — offsets and newlines preserved — and the remainder scanned at file level, with the same four shapes also emitted as contract members.

As the fixture predicted, this also **resolved** the dangling `Vault.deposit --[CALLS]--> computeFee`. That edge existed but pointed at nothing, because the callee was never extracted — so this defect was costing *resolution*, not just entity count.

**4. Modifier usage** was never linked. A new `declAttributes` reads the region `declBody` skips, and `modifierUsages` takes identifiers there while skipping attribute keywords and whole paren groups — so `returns (uint256)` does not mint `uint256`. Base-constructor calls in the same position are emitted too.

## A fixture assertion that was an unmeasurable guess

`MyToken --[EXTENDS]--> ERC20` was asserted as `to_name` + `to_kind: SCOPE.External`. That was written when `MyToken` did not exist — the whole file was being dropped — so nothing could have validated it.

**Falsified by experiment:** `ERC20` does **not** fold to a `SCOPE.External` node. Neither does a renamed `Fooble`, nor an arbitrary `Zonk` added to `Ownable`. `Context` folds only because that name is **allowlisted by ext-synthesis** — so the one case the original fixture generalised from was the exception, not the rule.

Changed to `to_bare_name: "ERC20"`, per the NOTICE's own rule. It is also the **stronger** assertion: `to_name` against a non-existent entity can never pass, whereas the raw-`ToID` compare still fails if the parent parses as `ERC20(` or `)` — which is exactly the defect being fixed. Counts unchanged; NOTICE updated with the reasoning.

## Mutants — 8 of 8 killed

One per defect, plus a second for each of the two with a distinct failure mode: the inheritance character class *and* the depth-blind comma split; `constructorRE` *and* `specialFunctionRE`; the file-level scan *and* the contract-level loop; `declAttributes` *and* the attribute-keyword filter.

**One mutant's first anchor was ambiguous** and landed in the function loop rather than the declaration loop. It was caught because the expected killing test was **absent from the kill list**, and re-run with a unique anchor. Worth recording: a mutant that lands in the wrong place reports a real exit code for the wrong reason, and only the *identity* of the failing test reveals it.

## Verification

`go test ./internal/extractors/solidity/ ./internal/quality/` → 0 · `go vet` → 0 · `gofmt -l` → empty. Graded with a freshly built binary under full isolation.

**Out of scope, untouched:** #6425 — the Yul `Vault.helper` phantom, the `type`/`abi.encode` over-fire, the `IERC20(...)` under-fire. The only remaining fixture gap is that issue's `nice_to_have` at `relationships 0/1`.

### DIFF
--- a/internal/extractors/solidity/extractor.go
+++ b/internal/extractors/solidity/extractor.go
@@ -2,14 +2,26 @@
 //
 // Extracted entities:
 //   - `contract Foo {…}`  / `library Foo {…}` / `interface Foo {…}` → SCOPE.Component (subtype="contract"/"library"/"interface")
-//   - `function name(…) …` → SCOPE.Operation (subtype="function")
+//   - `function name(…) …` → SCOPE.Operation (subtype="function"), inside a
+//     contract or at file level (a free function)
+//   - `constructor(…){…}`  → SCOPE.Operation (subtype="constructor")
+//   - `receive(){…}` / `fallback(){…}` → SCOPE.Operation (subtype="receive"/"fallback")
 //   - `event Name(…);`    → SCOPE.Operation (subtype="event")
+//   - `error Name(…);`    → SCOPE.Operation (subtype="error")
 //   - `modifier name(…){…}` → SCOPE.Operation (subtype="modifier")
+//   - `struct S {…}` / `enum E {…}` → SCOPE.Schema (subtype="struct"/"enum")
+//   - `type T is uint128;` → SCOPE.Schema (subtype="type")
 //   - `uint256 public cap;` → SCOPE.Schema (subtype="field") for contract-level state variables
 //   - `import "./Foo.sol"` / `import "…"` → IMPORTS relationship
-//   - `contract Foo is Bar, Baz` → EXTENDS edges (on the contract component)
+//   - `contract Foo is Bar, Baz` → EXTENDS edges (on the contract component),
+//     including the base-constructor form `is ERC20("n","s")`
 //   - Function-call expressions → CALLS edges
-//   - CONTAINS edges (contract → its functions/events/modifiers/state variables)
+//   - Modifier usage in a declaration's attribute section → CALLS edges
+//   - CONTAINS edges (contract → its members)
+//
+// Every declaration form except `function`/`event`/`modifier`/state variables
+// was added by #6423; before it, a contract with base-constructor arguments
+// matched nothing and took its whole file's contents with it.
 //
 // No tree-sitter grammar for Solidity is bundled in smacker/go-tree-sitter, so
 // this extractor uses regular expressions.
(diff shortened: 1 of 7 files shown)


cajasmota:
Fix round pushed as `ae2e40f86`, after an independent adversarial review. The review approved the 40/40 as earned but found two regressions this PR *introduced*, one surviving mutant, and a false causal claim. All five are addressed.

**Two regressions, both this PR's own:**

1. An unclosed contract brace leaked every member as a phantom **file-level** entity. When `extractBracedBody` returned `endLine == 0` the mask span degenerated to the header, leaving the whole body visible, so `findFileLevelDecls` emitted each member bare with no `CONTAINS` owner. On `origin/main` those entities did not exist at all. The mask now runs to `len(src)` — an unterminated contract runs to EOF by definition, and its members are lost with it, which is the pre-existing behaviour.

   The code comment there claimed masking less was "the conservative direction: it masks less, never more". **That has the sign backwards** and is now rewritten: whatever the mask leaves visible is what gets emitted, so masking less mints *more* phantoms, not fewer.

2. A `fallback();` or `receive(...)` **statement** at line start inside another function minted a phantom member, because the special-function and constructor patterns were matched against the whole contract body including nested function bodies. A new `braceDepths` helper labels each byte with its nesting depth and members are recognised only at depth 0.

**Bonus, measured rather than assumed:** that same depth guard removes the #6425 Yul phantom — `function helper(x) -> y` inside `assembly{}` sits two braces deep. Graph delta on the fixture is exactly −1 entity (`Vault.helper`) and −4 edges, with `Vault.roundUp CALLS helper` correctly reverting to a dangling bare name. The held-back forbidden row for it has been promoted into `expected.json`: the old binary scores it 1, the new one 0.

**The surviving mutant is dead.** `declEndOffset`'s bodied branch was entirely unpinned — collapsing every struct/enum `EndLine` onto its `StartLine` passed the full suite. Four span-producing declaration kinds had shipped with zero span assertions despite `issue6234_declaration_spans_test.go` existing for exactly this. Both call sites are now pinned.

**Correcting a claim I made.** The fixture NOTICE asserted that `Context` folds because that name is allowlisted by external synthesis. That was wrong, and it was wrong in a fixture, stated as measured fact. Every `"Context"` allowlist entry is language-gated and cannot fire for Solidity.

The real mechanism is `isKnownExternalPackage` (`internal/external/synth.go`), which **lower-cases** the candidate against `knownExternalPackages` — an allowlist with **no** language gate. `Context` → `context` hits the Go-stdlib section; `ERC20` → `erc20` is absent. Confirmed four ways by renaming the parent on the fixture: `Zebra` → no node; `Express` → `ext:Express`; `Fmt` → `ext:Fmt` (demonstrating the case-fold); and swapping the two files' import styles changed nothing, ruling out import-path form. The conclusion is unchanged — `ERC20` does not fold and the `to_bare_name` assertion stands — but the mechanism is now established rather than guessed.

**Re-graded:** entities **40/40**, relationships **13/13**, forbidden hits **0** — now against **5** forbidden rows rather than 3, including `Vault.ceiling CALLS uint256` so the graded number carries the `modifierUsages` precision guarantee it previously said nothing about.

**Independent mutant, run by me rather than taken from the agent's table.** I removed the base-argument truncation from `parseInheritanceList` — the headline fix, the thing that makes `contract X is ERC20("n","s")` parse at all — verified it compiled (`go vet` exit 0) and that it actually landed, then ran the package: **DIED**, killed by `TestSolidity_6423_InheritanceListIsParenAware`. Restored by `cp` with a matching shasum. The 40/40 is not resting on an unpinned headline.

One thing flagged and deliberately left out of scope: the two pre-existing `CALLS` forbidden rows (`string`, `emit`) use `to_name` + `to_kind: SCOPE.External` against targets that never resolve, so they are probably vacuous in the same way the new `uint256` row would have been had it not used `to_bare_name`. Worth a separate look.

Issue #2645, opened 2026-05-27
[bug] Daemon watcher not reindexing file changes — Cascade saw stale line numbers despite mid-session edits


cajasmota:
**Surfaced**: Cascade investigation 2026-05-27. Agent edited files in core-mobile (e.g. moved a cleanupInspection call from line 128 → 144) but inspect responses for callers still showed line=128, despite `whoami` reporting `indexed 32m ago` (well after the edit).

## Expected behavior
The daemon registers a file watcher per repo (fleet config). When a file changes on disk:
1. Watcher fires
2. Per-file partial reindex is queued
3. Index updates the affected entities' line ranges + outgoing edges within seconds

## Actual behavior
Edits were not picked up. `indexed_ref` SHA didn't change (still at the pre-edit commit). Line numbers stayed stale until a manual `archigraph rebuild upvate` was triggered.

## Investigation needed
1. Is the watcher actually subscribed to core-mobile's directory tree?
2. If yes — does it filter out .ts/.tsx files? Some watchers only watch .py / .go by default
3. Or is the watcher firing but the queued reindex sits in a backed-up queue?
4. Check daemon mode (background/workstation/readonly) — background mode may deprioritize reindexes
5. Check if reindex requires a git-state change (committed) vs working-dir change (uncommitted)

## Fix hypotheses
- **Watcher coverage**: extend file-extension allowlist to include all TS/JS variants (.ts, .tsx, .js, .jsx, .mjs, .cjs)
- **Queue priority**: reindex on file change should be high-priority, not background
- **Working-dir vs committed**: working-dir changes should trigger reindex too (per-file, not per-commit)

## Acceptance
- Edit a TS file in core-mobile → daemon log shows watcher fire → reindex completes within 30s
- Inspect on the modified entity returns updated line numbers without manual rebuild

## Evidence
Cascade report: "called_by on cleanupInspection reported line: 128 ... but after our edits the eviction call is at line ~144"

PR #6439, opened 2026-08-20
fix(#6267): WARM→COLD never released a descriptor, so a refused repo stayed unwatched for the life of the daemon

### LINKED ISSUE #6267: A default-branch-pinned repo refused for descriptor budget retries every ~65 min and can never succeed, because nothing returns descriptors
#6180's descriptor budget (`adff65308`) replaced EMFILE-at-repo-6 with a bounded refusal. But the refusal is **permanent and first-come-first-served**, so on a multi-repo store the first few repos claim the entire budget and every repo after them is silently unwatched for the life of the daemon.

## The mechanism

`fdUnwatched` (`internal/daemon/watch/watcher.go:142-144`) records a repo refused for lack of budget, set at `:441`. It is cleared in exactly two places:

- `watcher.go:459` — a **successful** `subscribeRepo`
- `watcher.go:491` — `RemoveRepo`

There is no path that retries a refused repo when budget frees. Nothing polls, nothing re-attempts on release at `:487`.

And budget effectively never frees, because the descriptors are never returned:

- `tier.go:665-674` — on WARM→COLD the graph is evicted but `wh.Pause` is **deliberately not called** (#2645: dropping the subscription made TS/TSX edits undetectable). The comment is explicit that removal is deferred to COLD→EXPIRED.
- `tier.go:651` — `if !s.isPinnedMain && idle >= expiredWindow`. Every repo's default branch is pinned, so COLD→EXPIRED never fires for it either.

So a subscription, once established, holds its ~10k descriptors permanently.

## What the user sees

On a 61,440 soft limit the budget is 46,080 (`fdbudget_darwin.go:43` — effective `RLIMIT_NOFILE` minus a 1/4 reserve, floor 1024). At the ~10.2k/repo calibrated in #6180 that is **about four repos**. Whichever four are queried or indexed first win, permanently.

Repo five onward: no file watching, so no incremental reindex on edit. Queries still answer — from an index that silently stops tracking the working tree. The warning at `watcher.go:446-452` is honest but fires once, at refusal time, and nothing surfaces the state afterwards. `grafel status` does not report it.

**This is the silent-wrongness shape**, and it is worse than the EMFILE it replaced: EMFILE was loud.

## What a fix has to settle

Retry-on-release is the small version and is probably not sufficient alone — with nothing ever releasing, there is nothing to retry into. The candidates interact:

- **Retry refused repos when budget frees** (`watcher.go:487`). Cheap, correct, and inert until something actually releases.
- **Make something release** — re-enabling `Pause` on COLD is the only mechanism that returns descriptors at scale (~10k per 60-min-idle repo). That reintroduces #2645 unless a resume-time catch-up rescan lands first; filed separately.
- **Refuse vs LRU-evict on exhaustion** (`watcher.go:430`). Currently refuse, i.e. first-come-wins. LRU would at least make the winners the *recently used* repos rather than the *first* ones. This is a policy call.

Whichever lands, the state needs to be visible: a repo that is indexed but unwatched should say so on a surface a user reads, not only in a one-shot log line.

Refs #6180, #2645, #6233

cajasmota:
Closes #6267

## What was wrong

`#2645` deliberately removed the `Pause` on WARM→COLD and deferred it to COLD→EXPIRED. The consequence: a WARM→COLD transition released the graph but **not the file descriptors**. On a machine where the fd budget is already saturated, a repo whose `AddRepo` was refused stayed unwatched for the life of the daemon — nothing ever freed the budget it was waiting on, because the slots holding it were COLD, not EXPIRED.

## The fix (arm (b) of the two-arm plan)

`Pause` is called from **both** WARM→COLD paths in `internal/daemon/tier/tier.go`:

1. the TTL scan's `toEvict` loop — the same loop as `m.onEvict(k)`, so graph release and subscription release cannot drift apart;
2. `scanPressureEvict()`'s `toEvict` loop — otherwise the identical bug arrives by the other road, with a pressure-COLD slot holding descriptors unreleasably until EXPIRED. This path needed its own `m.watcher` read under `m.mu`.

The COLD→EXPIRED `Pause` is **kept**. It is redundant for the normal path, but a `RegisterCold` slot can expire without this process ever observing its WARM→COLD, and `DefaultManager.Pause` is idempotent per `(repo, ref)`.

The refusal path in `watcher.go` is untouched — arm (b) needed nothing there.

## Evidence

The new test wires the **real** `tier.Manager` → **real** `watch.DefaultManager` → **real** fsnotify `Watcher`, with the budget set to exactly one repo. It is in-package so it can inject the kqueue cost model and therefore be the same test on Linux, macOS and Windows.

| step | observed |
|---|---|
| A queried first | watched, `used=4/4` |
| B queried | `AddRepo` refused, `IsPaused(B)`, tier slot B still **HOT** — the reload succeeded, only the subscription failed |
| A HOT→WARM | `used` stays 4 — WARM must not release |
| A WARM→COLD | **`used` 4 → 0**, `w.Repos()` empty |
| B's next query | re-subscribes, `used=4`, unwatched count 0 |
| edit under B | **sink fires with a non-bulk fsnotify event** — a live subscription, not bookkeeping |

That last row is the part that is not asserted from a counter. Which repos win the budget race is deliberately **not** asserted — that is map-iteration order.

### What this costs, stated as a trade and not as a strict improvement

**#2645's stated acceptance criterion is no longer met.** Its bar was: edit a TS file → the watcher fires → reindex within 30 s. After this change an edit during the COLD window produces **no watcher fire at all**. Recovery needs a *query*, and the `Resume` that query triggers dispatches `RescanRepo` on its own goroutine *after* `Touch` has already served from the on-disk graph — so the waking query is stale and only the one after it is fresh. That is the trade the maintainer decision accepts ("one query late" beats "never watched again"), but it is a real regression against #2645's own criterion, not a free win.

The staleness is separately not *observable in this package*: the sink double records `(repo, bulk)` without reindexing, so no graph freshness can be measured here at all (`resume_catchup_6269_test.go` says the same of itself). What the logs do show is the resume-side `watcher: catch-up rescan after re-subscribe`. Proving the one-query lag end to end needs the real scheduler and indexer.

## Mutants

| mutant | result |
|---|---|
| drop `Pause` from the TTL loop | 3 unit tests fail; e2e fails at `used=4 want 0` |
| drop `Pause` from the pressure loop | `want 1 Pause call on pressure-evict, got 0` |
| `Pause(k.RepoPath, "")` — wrong ref | `Pause fired for the wrong slot` |

The third is the one worth keeping: a wrong-ref `Pause` still drives the refcount to 0 and still releases the descriptors, so the end-to-end assertion alone would not have caught it.

## Tests that changed meaning

Three pre-existing tests encoded the `#2645` behaviour and now encode its reverse: `TestWatcherNotPausedOnCold` → `TestWatcherPausedOnCold`, `TestSlotNotPausedAfterWake` → `TestSlotPausedThenResumedAfterWake`, and the "still no Pause during COLD" precondition inside `TestWatcherPausedOnExpired`. Two stale comments were corrected: the tier package-doc transition table and `cmd/grafel/daemon_tier.go`, which still claimed the Pause was deferred.

## Review follow-up (second commit)

Adversarial review found one HIGH regression introduced by the first commit, fixed here in `DefaultManager.Pause`.

`Pause` decremented the repo-level `refCounts[repoPath]` even for a slot this manager had never seen. Nothing had ever incremented it for that slot, so the decrement took a descriptor back from a **sibling** ref holding a live subscription: `refCounts` hit 0 → `RemoveRepo` → the sibling's subscription died while its own `paused` flag stayed `false`, so every later `Resume` for it took the "already active" early return and never retried `AddRepo`. Permanently unwatched **and** permanently reported as watched — the shape of #6267 itself.

It is reachable because the tier keyspace and the watch keyspace are not required to agree: `tier.Touch` auto-registers an unknown key as HOT **without** calling `Resume`, so a ref queried but never indexed in this process has a tier slot and no watch slot. Before this branch it reached `Pause` only via COLD→EXPIRED (non-pinned refs, after 48 h); making WARM→COLD pause brought it to every ref on every ~1 h idle cycle. An unknown slot is now recorded as paused with the refcount left alone, pinned by `TestPauseOfUnregisteredSlotLeavesSiblingSubscriptionIntact`, which without the guard fails at `pausing an unregistered ref killed the sibling's subscription: w.Repos()=[]`.

Also from the review (LOW): `scan()` read `m.watcher` unsynchronised while the pressure loop read it under `m.mu`, and `SetWatcherHook` writes it under `m.mu`. Both sites now take the lock.

## Known-latent, deliberately not fixed here

- **`SubscribeGroup`'s sentinel slot.** `manager.go` creates a slot with ref `""` and bumps `refCounts[rp]`. It has no tier slot, so nothing ever Pauses it, the refcount never reaches 0, and the descriptor release in this PR does nothing on that path. Harmless today — `SubscribeGroupWatcher` has zero production callers — but **arm (a) may rewire it**, so it is recorded rather than assumed away.
- **What `TestWatcherPausedOnExpired` actually shows.** It asserts `pausedCount() == 2` against a fake hook with no idempotence, so it pins the *double* call (WARM→COLD then COLD→EXPIRED) without proving the second one harmless. The harmlessness rests on reading `DefaultManager.Pause`, which returns early when `st.paused` — not on this test.

## Follow-up

This is arm (b). Arm (a) — LRU eviction so the budget picks a victim deliberately rather than by whoever happens to age out — lands next, on top of this.

### DIFF
--- a/cmd/grafel/daemon_tier.go
+++ b/cmd/grafel/daemon_tier.go
@@ -394,11 +394,12 @@ func tierEvictCallback(key tier.SlotKey) {
 // graph was last indexed, enqueue a reactive reindex so the query is served
 // from the most up-to-date graph on the next request.
 //
-// #2645: also ensure the fsnotify subscription is live after a cold wake.
-// In the normal path the subscription is kept through WARM→COLD (the Pause
-// is now deferred to COLD→EXPIRED), but an EXPIRED slot that got re-indexed
-// will have had its subscription removed. Resume is idempotent, so this call
-// is safe even when the subscription is already active.
+// Also ensure the fsnotify subscription is live after a cold wake. Since
+// #6267 the subscription is released on WARM→COLD (that release is what
+// returns descriptors to the watch budget), and tier.Touch resumes it before
+// invoking this callback; an EXPIRED slot that got re-indexed has likewise had
+// its subscription removed. Resume is idempotent, so this call is safe — and
+// still required — whichever path got us here.
 func tierReloadCallback(key tier.SlotKey) error {
 	stateDir := daemon.StateDirForRepoRef(key.RepoPath, key.Ref)
 	// #5915 J1 FIX-3: prime the cache via the segment-aware ref entry point.
--- a/internal/daemon/tier/tier.go
+++ b/internal/daemon/tier/tier.go
@@ -12,7 +12,8 @@
 // # Transitions
 //
 //	HOT  → WARM  : 5 min idle  (status change only, no memory action)
-//	WARM → COLD  : 1 h  idle   (release in-memory reference; GC eligible)
+//	WARM → COLD  : 1 h  idle   (release in-memory reference AND the fsnotify
+//	                            subscription — #6267; GC eligible)
 //	COLD → HOT   : demand wake (reload from disk; ≤ 300–500 ms typical)
 //	COLD → EXPIRED : idle past ExpiredWindow (PH6: also triggers disk delete)
 //
@@ -669,16 +670,31 @@ func (m *Manager) scan() {
 	// This runs outside the lock so it can call m.onEvict safely.
 	pressureEvicted := m.scanPressureEvict()
 
-	// PH2a: release in-memory graphs for newly-COLD slots but keep the
-	// fsnotify subscription alive so file edits during the COLD window still
-	// trigger reindex. (#2645: removing the subscription here caused TS/TSX
-	// changes to go undetected when a repo had been idle for >60 min.)
+	// PH2a: release in-memory graphs for newly-COLD slots, AND release the
+	// fsnotify subscription with them (#6267).
+	//
+	// #2645 removed the Pause call from this loop so that edits during the COLD
+	// window still produced events. The measured cost of keeping it removed is
+	// worse than the problem it solved: descriptors are then never returned at
+	// scale, because a repo's only slot is usually its pinned default branch,
+	// which can go COLD but never EXPIRED — and COLD→EXPIRED was the only other
+	// caller of Pause. A repo refused for want of descriptor budget therefore
+	// retried once per idle cycle and failed EVERY time, for the life of the
+	// daemon: indexed, answering queries, receiving zero fs events. A COLD repo
+	// releases ~10k descriptors; nothing else releases anything.
+	//
+	// #2645's symptom does not return in full: the cold wake resumes the
+	// subscription and asks for a catch-up rescan of everything that changed
+	// while it was gone (#6269, watch.DefaultManager.Resume). That rescan is
+	// asynchronous by design, so the edit is detected ONE QUERY LATE rather
+	// than not at all — see the note on Touch. That staleness is the accepted
+	// trade (#6267).
 	wh := m.watcher // read under mu already released; field is write-once after init
 	for _, k := range toEvict {
 		m.onEvict(k)
-		// wh.Pause intentionally NOT called here — subscription removal is
-		// deferred to the COLD→EXPIRED path below so that watcher events keep
-		// firing while the graph is cold-but-still-on-disk.
+		if wh != nil {
+			wh.Pause(k.RepoPath, k.Ref)
+		}
 	}
 	if len(toEvict) > 0 || pressureEvicted > 0 {
 		runtime.GC() // nudge GC so released graph objects are reclaimed promptly
@@ -698,8 +714,10 @@ func (m *Manager) scan() {
 	}
 
 	// PH6: perform disk eviction for newly expired slots.
-	// Only EXPIRED slots lose their fsnotify subscription — at that point the
-	// graph.fb has been deleted and there is nothing to reindex into.
+	// The Pause here is belt-and-braces since #6267 made WARM→COLD release the
+	// subscription: a slot registered COLD at boot (RegisterCold) can expire
+	// without this process ever having observed its WARM→COLD, and Pause is
+	// idempotent per (repo, ref) in watch.DefaultManager.
 	for _, k := range toExpire {
 		if wh != nil {
 			wh.Pause(k.RepoPath, k.Ref)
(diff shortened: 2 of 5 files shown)


cajasmota:
Third round, test-only (`fc20d70e6`) — no behaviour change. Both idempotence guards already existed and were already correct; what was missing was any test that could tell.

## How this was found

After the review approved the unknown-slot guard, I ran one more mutant of my own before merging: delete `Pause`'s `if st.paused` early return, so a second pause on an already-paused slot falls through and decrements `refCounts` twice. It compiled, it landed, and the suite stayed **green**.

That mattered more than an ordinary uncovered line, because idempotence is the stated justification for a design decision in this PR: the COLD→EXPIRED `Pause` was **kept** on the reasoning that `Pause` is idempotent per `(repo, ref)`, so WARM→COLD followed by COLD→EXPIRED cannot double-decrement. The first review had already flagged that this rested entirely on reading `manager.go` — `TestWatcherPausedOnExpired` pins the double *call* against a fake hook with no idempotence, so it proves the call happens, not that it is safe. The mutant confirmed the gap was real.

## Not an equivalent mutant

The double `Pause` is reachable: a non-pinned ref takes WARM→COLD and then COLD→EXPIRED, so tier calls `Pause` twice for the same key. The new test drives both transitions through the **real tier scanner** (asserting `TierCold`, then `TierExpired`) rather than calling `Pause` twice by hand, so reachability is demonstrated rather than assumed. A test was the right answer here; had the branch been unreachable, deleting it would have been.

## Why the old coverage missed it

`manager_test.go:75-78` asserts on `PausedCount()` — which a double *decrement* does not disturb — and uses a single ref, so `refCounts` goes 1 → 0 → -1, clamps to 0, and `RemoveRepo` simply fires twice. Nothing observable changes. It takes a **sibling** ref for the corruption to surface.

`TestPauseAndResumeAreRefcountIdempotentAcrossSiblingRefs` subscribes `main` and `feat/x`, ages `feat` to COLD then EXPIRED while `main` keeps being queried, and asserts after the **second** pause that the repo is still subscribed, `SubscribedRepoCount() == 1`, and `main` is still active.

Mutant, re-verified independently after the fix (`go vet` exit 0, mutation confirmed landed, restored by `cp` with matching shasum):

```
--- FAIL: TestPauseAndResumeAreRefcountIdempotentAcrossSiblingRefs
    pause_on_cold_6267_test.go:406: double Pause of one ref unsubscribed the
    repo while main was active: w.Repos()=[]
```

## The mirror hazard on Resume — same shape, also now pinned

`Resume`'s `if !st.paused` early return exists and is correct, and was unpinned in precisely the same way: the existing double-resume assertion checks `ActiveCount()`, which a double *increment* does not disturb. A double increment means `refCounts` can never reach 0, `RemoveRepo` never fires, and descriptors never return — #6267 from the other side, arrived at by the opposite arithmetic.

Since the guard was already present this needed a test rather than a fix, so there is nothing to file separately:

```
--- FAIL: TestPauseAndResumeAreRefcountIdempotentAcrossSiblingRefs
    pause_on_cold_6267_test.go:422: double Resume double-incremented the
    refcount: pausing the last active ref left the repo subscribed
```

The pattern worth carrying forward: both gaps came from asserting on a **flag** (`PausedCount`, `ActiveCount`) where the property that actually matters is **arithmetic** (`refCounts`). A flag-based assertion is idempotent by construction, so it cannot observe a double-apply. That is the same reason the earlier clamped-decrement mutant survived a two-ref test — the clamp hid the arithmetic at its endpoint.

PR #6468, opened 2026-08-21
fix(#6416): a FIFO with an indexed extension blocked an indexing worker forever — the walker branched only on IsDir

### LINKED ISSUE #6416: walker: a FIFO with an indexed extension blocks an indexing worker forever — it branches only on IsDir(), and a FIFO does not fail the read, it waits
Found while fixing a FIFO hazard in the VB.NET partial-class extractor (#6327 S7a). That slice's own bug is fixed; **this one is upstream of it and is not**.

`internal/daemon/walk/walker.go:136-253` branches only on `d.IsDir()`. Anything that is not a directory is treated as a file and handed to an extractor, which reads it. A **FIFO** named `Foo.vb` (or `Foo.py`, or anything with an indexed extension) is not a directory, so it is passed through — and `os.ReadFile` on a FIFO **blocks on `open(2)` until a writer appears**. It does not fail; it waits.

That deadlocks an indexing worker indefinitely. No timeout, no error, no log line — the worker simply stops.

## Why this is easy to miss

The instinct is that a non-regular file will "fail the read and declare nothing." That is true for a **directory**, which is why a directory named `Foo.vb` behaves harmlessly. It is not true for a FIFO. We shipped exactly that reasoning inside the VB.NET extractor one commit before measuring it, and the measurement showed extraction blocked for over 5 seconds before the test's deadline fired.

The extractor-side fix (filter the listing so a path that cannot be opened without blocking is never chosen) is in flight for the VB.NET case. **It does not help here**, because the walker hands the path to the extractor directly — the extractor never gets a chance to vet a path it did not choose.

## Scope

**Derive the exposure — do not trust this issue's framing.** I have confirmed the `d.IsDir()` branch by reading `walker.go` and confirmed the blocking behaviour empirically in the extractor context. What I have **not** established:

- Whether any other guard upstream of the walker already excludes non-regular files (an ignore-file pass, a size check that stats first, a classifier gate). If one does, this may be unreachable in practice — that would make it a documentation fix, not a code fix.
- Whether the same hazard exists on the incremental path, which reaches extractors differently.
- Character/block devices and unix sockets, which have their own read semantics and may deserve the same treatment.

The likely fix is a `d.Type().IsRegular()` check (with a deliberate decision about symlinks, which the walker currently follows to files — `filepath.WalkDir` reports a symlink-to-file as a file, and the indexer **does** mint a file entity for it, so blanket rejection would lose legitimate coverage).

## Testing note

A test for this must run extraction on a goroutine against a deadline, so a regression **fails rather than hangs the suite**. A naive test that simply calls the walker on a FIFO will hang CI rather than report. The VB.NET equivalent uses `time.After` and a `//go:build unix` tag, since FIFOs are not creatable on Windows — worth copying that shape.

Realistically low-frequency in ordinary repos, but it is an unbounded hang with no diagnostic, triggered by a file an unprivileged user can create inside a watched directory.

cajasmota:
Closes #6416

`mkfifo` a file with an indexed extension and the indexing worker blocks forever. The walker branched only on `IsDir()`, and a FIFO does not fail the read — it waits.

**This PR grew from one package to seven across three rounds, because two independent reviews each found the hang still open on paths the previous round had not swept.** That history is the most useful thing in this description, so it is kept rather than tidied away.

## What was actually wrong, in the order it was found

**Round 1 — the walker.** `internal/daemon/walk` gained an entry-type gate (`irregularSkipRule`, `IrregularSkipReport`) and reported skips instead of dropping them.

**Round 2 — three more paths, found by executing rather than reading.** `grafel index` still wedged on a FIFO `package.json` **39 lines below the new gate**: `internal/install/detect` read three manifests by name. And a second, independent `filepath.WalkDir` in `internal/secrets` branched only on `IsDir()` before opening — reachable from the **live daemon's MCP tool** and a dashboard HTTP handler, so a remote caller could wedge a daemon goroutine. Its size guard was useless because a FIFO reports size 0.

**Round 3 — the class swept properly.** `internal/extractors/javascript/aliases.go` had **eight** name-chosen reads (`tsconfig.json`, `jsconfig.json`, vite, webpack, metro, babel, plus two behind an existing stat), reachable for *any* JS/TS file in the tree and not from the walker at all. `internal/extractors/golang/gomod.go` read `go.mod` twice with no gate — **unconditional on every Go repository, including this one**. `internal/extractors/yaml/helm.go` read sibling `Chart.yaml`/`Chart.yml` by name.

Ten sites were triaged; **seven of them are walker-gated** and are TOCTOU-hardening rather than open bugs — a `mkfifo` alone does not reproduce on them. That distinction is stated explicitly because this PR twice claimed a class was closed when it was not.

## The alias hang was worse than a hang

The RED stack showed a second goroutine parked in `resetAliasMapCache`. `loadRepoAliasState` holds the package-wide `aliasMapCache` **write lock across the parse**, so the abandoned goroutine never releases it. One FIFO did not lose one repo's alias map — **it wedged alias resolution for every repo in the process.**

Checked rather than assumed for the other two: `goModuleRoot`/`goModuleReplaces` both `RUnlock()` before parsing and take the write lock only after, and `helmSiblingSubcharts` holds no lock at all. Their blast radius is the abandoned worker only.

## `O_NONBLOCK` does not do what the fix originally assumed

The first design proposed reusing `gitignore_unix.go`'s `IsRegular` + `O_NONBLOCK` shape. **`O_NONBLOCK` does not make a FIFO fail to open** — the read-end open succeeds immediately whether or not a writer exists. The first draft asserted `ErrWouldBlock`, got `err = <nil>`, and because it then cleared the flag would have **moved** the hang from `open` to the first `Read`.

The layer that closes it is **`fstat` on the returned descriptor**, which is also what makes it unraceable: it asks the file we hold, not the path we asked for. `gitignore_unix.go` has the same shape and escapes the problem only because it `Lstat`-gates first.

## `internal/safeio`

Stat gate → `O_NONBLOCK` open → **fstat on the descriptor** → bounded-concurrency semaphore, platform-split like `gitignore_{unix,windows}.go`. `ReadFile` sites are byte-capped, because a character device opens fine and never reaches EOF — EOF is not a bound.

Round 2's review found two defects **introduced by round 2 itself**:

- **A silent-data-loss bug.** The semaphore acquired with a non-blocking `select`/`default` and returned `ErrWouldBlock` on miss, from a package-level global. Measured on 400 ordinary regular files: **325/293/300/234/329 spuriously refused across five runs.** Because `internal/secrets` mapped the error to `return nil`, the failure mode was **the secrets scanner reporting "clean" after reading a sixth of the tree**. Root cause was two-part — fail-fast acquire, *and* the release sitting after the send to the result channel, so slots accumulated unreleased. The acquire now waits; the caller releases on the fast path via `sync.Once`.
- **The non-Unix path had no TOCTOU layer and one leak bricked the package.** The abandoned worker never reached `<-openSlots`, so the doc's "degrades into bounded leakage" was false — it degraded into permanent, silent, total failure. It now has a descriptor-level gate (`os.Open` + `os.File.Stat` on the handle) and `TestDescriptorTypeGateIsUnraceable` runs on **every** GOOS.

Verified rather than asserted: with the build tag flipped onto darwin, `nonBlockingOpen` **still blocks on a real FIFO**, because Windows has no `O_NONBLOCK`. That fact is pinned in the doc instead of a parity claim. The caller is bounded; the worker cannot be rescued, because nothing interrupts a thread parked in `open(2)`.

## Every skip is reported, and the report names the file

The `#6338` lesson — a silent skip is worse than a loud one — had been satisfied only for a foreground `grafel index`. `walkSourceFiles` now returns its report and `tryIncremental` logs it, so a watcher-driven reindex no longer drops a FIFO with no record anywhere.

The extractor sites report through a shared convention: always-on stderr, deduped by path, capped, and gated to `ErrNotRegular`/`ErrWouldBlock` only — ENOENT is the normal case and would bury the signal.

**A coordinator mutant found the `ErrWouldBlock` arm unpinned in all three reporters.** That arm is the safety net for exactly the silent-mass-skip failure above, and removing it left every package green. Pinning it surfaced a further defect: `safeio` returns `ErrWouldBlock` **bare**, carrying no path, so the warning read `skipped alias config safeio: open would block` and **named no file at all**. Now decorated.

Stated plainly: those tests call the reporters directly. `ErrWouldBlock` has two producers, both in `openWithDeadline`, and **neither is reachable from `go test` on a healthy filesystem** — `O_NONBLOCK` means the one hangable object a test can create opens immediately and is refused by `fstat` as `ErrNotRegular`. The real producers are hung network/FUSE mounts and root-owned device nodes. So: *given* an `ErrWouldBlock`, the reporter announces it and names the file — pinned. *That a would-block read reaches the reporter* — not pinned; it is one shared `if err != nil` already driven end-to-end by the `ErrNotRegular` FIFO tests.

## Round 4 — the enumeration, which is the part that generalises

Four rounds had each fixed exactly the sites the previous review named and swept no further, and each claimed the class was closed. The thing nobody had done was the obvious one: **grep for `os.ReadFile`/`os.Open`/`os.OpenFile` with a literal filename argument.**

That enumeration now exists — **324 non-test call sites** across `internal/` and `cmd/`, classified name-chosen / walker-gated / not-applicable, in `docs/blocking-open-audit.md`. It immediately produced two sites no review had found:

- **`internal/daemon/walk/walker.go:343`** — `parseInheritedGrafelIgnore`'s raw `.grafelignore` read, **inside the very package whose entry-type gate this PR added**. `mkfifo .grafelignore` in any subdirectory hung `grafel index`.
- **`internal/gitmeta` ×5** — `.git`, `HEAD`, `commondir`, loose refs, `info/sparse-checkout`, which run on **every repo before any walk**. A hang there is an index that never starts, and no downstream gate can help. Living under `.git/` lowers likelihood and raises impact; nothing enforces file types there, so a hostile repo or a botched restore is enough.

Plus `internal/licenses` ×10 — five name-chosen at the repo root including `package.json` again, reachable only from the daemon's MCP license tool.

**26 sites remain**, all lower-severity and all listed, tracked in #6478 along with a spec for the AST lint that would keep the boundary closed. The doc is chosen over a test-embedded list deliberately: most not-applicable rows are provenance judgements no AST pass decides, and a list without a checker is a doc written in Go syntax.

## Round 5 — the reporting gap this PR had argued against

`internal/secrets` and `internal/install/detect` still mapped **both** reportable errors to a silent `return nil`. So `mkfifo creds.go` made the secrets scanner answer "clean" with no record anywhere — the #6338 shape this PR invokes as its own rationale, left in place for four rounds while the walker half was fixed for exactly that reason.

Both now report through the shared convention. Thirteen mutants, all dying — including one per `detect` parser, because un-wiring any single one left the other two green.

**Deliberately not built, with the cost stated:** `ScanPath` still returns `([]Finding, error)`, so the skip reaches the daemon log but not the MCP or dashboard *caller*. For a secrets scanner, "I did not check this file" and "I checked it and it was clean" are different answers and the API cannot express the difference. Closing it changes the signature at both entry points plus their JSON payloads — an API change, and this PR had already been extended twice mid-round. Filed as #6483.

## Two guards deleted for being unfalsifiable

An entry-type gate in `ScanPath`'s walk loop could only duplicate `scanFile`'s guard while leaving the race only the open site can close. Written, probed for a killing mutant, none found, removed. A guard nothing can test is a guard nobody can trust.

The semaphore was heading the same way — a mutant removing it entirely left three packages green — but under F2/F3 the right answer was to pin it, not delete it. `TestSemaphoreBoundsConcurrentOpens` now asserts the bound itself via an injected blocking opener. The injection is not testability theatre: a blocking open cannot be conjured on Windows, so without it the machinery that must work on `windows-latest` stays comment-only.

## Mutants

Three timeouts rather than assertions, which is the only honest way to pin a hang.

**Round 1-2:** `W1` walker gate un-wired → **times out** · `W2` gate below sparse filter · `W3` symlinked dir back in the `irregular:` namespace · `D1` detect back to `os.ReadFile` → **times out** · `S1` `scanFile` back to `os.Open` → **times out** · `IO1` fstat layer removed · `C1` index report block deleted · `E1` daemon report discarded.

**safeio round 3:** fail-fast acquire · semaphore removed · abandoned worker keeps its slot · caller deadline removed (**test timed out at 30s**) · non-Unix descriptor gate removed · Unix `fstat` gate removed.

**Sweep round 3:** each of the eight alias sites and the three `go.mod`/`Chart.yaml` sites reverted to `os.ReadFile` → **times out**; each skip-report call removed.

**Coordinator mutants:** the pre-open stat gate removed — died in `safeio` while `detect` and `secrets` stayed green, correctly showing the two layers are separately pinned rather than conflated. The `ErrWouldBlock` arm dropped — **survived in all three reporters**, now dies.

## Known gap, not fixed

The 8 MiB read cap has **no test**. Building a character-device fixture needs root, so the cap is argued, not measured. And `internal/extractors/cross/imports` has **no test files at all**, so reachability from that package is argued from the call site rather than exercised.

### DIFF
--- a/cmd/grafel/index.go
+++ b/cmd/grafel/index.go
@@ -1466,10 +1466,17 @@ func (i *Indexer) Run(ctx context.Context, absRepo string) (*graph.Document, err
 	// immediately. A second event follows once the walk completes with the
 	// real file count.
 	trk.PhaseStart(progress.PhaseScan, 0, 0)
-	files, _, err := walk.WalkRepo(absRepo, walkOpts)
+	files, walkSkipped, err := walk.WalkRepo(absRepo, walkOpts)
 	if err != nil {
 		trk.Fail(err.Error())
 		return nil, fmt.Errorf("walk repo: %w", err)
+	}
+	// #6416: non-regular files (FIFOs, devices, sockets) are dropped by the
+	// walker's entry-type gate because reading one can block an indexing
+	// worker forever. Say so unconditionally — a source-looking file that
+	// produced nothing, reported nowhere, is the failure mode #6338 is about.
+	if report := walk.IrregularSkipReport(walkSkipped); report != "" {
+		fmt.Fprintln(os.Stderr, report)
 	}
 	i.stats.files = len(files)
 	trk.SetFilesTotal(len(files))
--- a/cmd/grafel/index_fifo_report_6416_test.go
+++ b/cmd/grafel/index_fifo_report_6416_test.go
@@ -1,0 +1,76 @@
+//go:build unix
+
+package main
+
+import (
+	"context"
+	"io"
+	"os"
+	"path/filepath"
+	"strings"
+	"syscall"
+	"testing"
+	"time"
+)
+
+// TestIndexRunReportsIrregularSkipsOnStderr closes the review's MEDIUM-4 gap.
+//
+// "The skip is always reported" is a load-bearing claim of the #6416 fix, and
+// it was asserted nowhere outside internal/daemon/walk: IrregularSkipReport had
+// exactly one non-test caller, so deleting that block from Indexer.Run — or
+// quietly gating it behind --print-skipped — was invisible to every test.
+//
+// It also proves the FULL path end to end rather than the predicate: a FIFO
+// planted in a real indexed repo must not wedge Run (which is #6416 itself),
+// and must produce a line a user can actually see without opting in.
+func TestIndexRunReportsIrregularSkipsOnStderr(t *testing.T) {
+	repo := t.TempDir()
+	if err := os.WriteFile(filepath.Join(repo, "main.go"), []byte("package main\n\nfunc main() {}\n"), 0o644); err != nil {
+		t.Fatal(err)
+	}
+	if err := syscall.Mkfifo(filepath.Join(repo, "hang.go"), 0o644); err != nil {
+		t.Skipf("cannot create a fifo here: %v", err)
+	}
+
+	r, w, err := os.Pipe()
+	if err != nil {
+		t.Fatal(err)
+	}
+	orig := os.Stderr
+	os.Stderr = w
+	captured := make(chan string, 1)
+	go func() {
+		b, _ := io.ReadAll(r)
+		captured <- string(b)
+	}()
+
+	idx := newTestIndexer(t, "fiforepo", nil, t.TempDir())
+	done := make(chan error, 1)
+	go func() {
+		_, rerr := idx.Run(context.Background(), repo)
+		done <- rerr
+	}()
+
+	var runErr error
+	select {
+	case runErr = <-done:
+	case <-time.After(60 * time.Second):
+		// Restore before failing so the harness's own output is not lost.
+		os.Stderr = orig
+		_ = w.Close()
+		t.Fatal("Indexer.Run blocked on a fifo named hang.go — this is #6416 on the full path")
+	}
+	os.Stderr = orig
+	_ = w.Close()
+	out := <-captured
+
+	if runErr != nil {
+		t.Fatalf("Run: %v", runErr)
+	}
+	if !strings.Contains(out, "non-regular file") || !strings.Contains(out, "hang.go") {
+		t.Errorf("the skipped fifo was never reported on stderr; got:\n%s", out)
+	}
+	if !strings.Contains(out, "#6416") {
+		t.Errorf("the report does not point at the issue that explains it; got:\n%s", out)
+	}
+}
--- a/docs/blocking-open-audit.md
+++ b/docs/blocking-open-audit.md
@@ -1,0 +1,168 @@
+# Blocking-open audit (#6416)
+
+`os.ReadFile` and `os.Open` on a FIFO block in `open(2)` until a writer appears.
+Nothing bounds that wait: no timeout, no error, no log line. An unprivileged
+user plants one with a single `mkfifo`. `internal/safeio` exists to make that
+impossible; this document records **which call sites still do not use it**.
+
+## Why this document exists
+
+PR #6468 asserted "the class is closed" three times and was wrong three times.
+Each round fixed exactly the sites the previous reviewer had named and swept no
+further:
+
+| Round | Fixed | Missed, found by the next reviewer |
+|---|---|---|
+| 1 | `walk` entry-type gate | `install/detect`, `secrets` |
+| 2 | `install/detect`, `secrets` | six sites in `javascript/aliases.go` |
+| 3 | `aliases.go`, `gomod.go`, `helm.go` | ten sites in `internal/licenses` |
+
+The missed sites were never hard to find. A grep for a read with a literal
+filename under `internal/` takes seconds and would have caught every one of
+them in round 1. The enumeration below is that grep, done once, properly, so
+the next person does not have to rediscover the same class a fifth time.
+
+## Classification
+
+Every non-test `os.ReadFile` / `os.Open` / `os.OpenFile` call under `internal/`
+and `cmd/` falls into one of four buckets.
+
+- **name-chosen** — the path is built from a literal filename or a fixed list
+  of them, so `mkfifo <thatname>` in a scanned tree reproduces the hang. **No
+  walker sits in between, so the walker's entry-type gate cannot protect it.**
+  This is the bucket that matters.
+- **walker-gated** — the path comes from `walkSourceFiles`, a `filepath.WalkDir`
+  with an entry-type gate, or a `files []string` derived from one. Only a TOCTOU
+  residual remains, and `safeio`'s fstat-on-descriptor layer closes that at the
+  sites that already use it.
+- **already-safeio** — routed through `safeio` and reported. Done.
+- **not-applicable** — reads of grafel's own state (`~/.grafel`, sidecars, the
+  daemon log, pidfiles), writes, procfs, or a path the user typed as a CLI flag.
+  A hostile FIFO in any of those means the attacker already owns the account.
+
+A literal leaf joined onto a directory that itself came from a `ReadDir` still
+counts as **name-chosen** — the leaf is what the attacker names.
+
+## The open list: name-chosen sites not yet routed through safeio
+
+As of `c3e3720a8` plus this commit. `internal/licenses` is fixed here; the
+32 below are **not**, and are deliberately left for sized follow-up work rather
+than swept into an already-large PR.
+
+| Site | Literal name(s) | Note |
+|---|---|---|
+| `internal/daemon/walk/walker.go:343` | `.grafelignore` | **Highest severity.** An unguarded read *inside the walker package itself*, so the entry-type gate #6468 added never applies to it. `mkfifo .grafelignore` in any subdirectory of an indexed repo hangs the walk. |
+| `internal/gitmeta/cache.go:135` | `commondir` | Runs on every repo before any walk. |
+| `internal/gitmeta/cache.go:171` | `refs/…` from HEAD | |
+| `internal/gitmeta/cache.go:206` | `HEAD` | |
+| `internal/gitmeta/cache.go:279` | `.git` | |
+| `internal/gitmeta/sparse.go:105` | `info/sparse-checkout` | |
+| `internal/daemon/worktree/classify.go:125` | `.git` | `Lstat` + `!IsDir()` only; a FIFO passes. |
+| `internal/daemon/worktree/worktree.go:813` | `.git` | `os.Stat` + `!IsDir()` only. |
+| `internal/daemon/watch/gitignore.go:81` | `.grafel/watch.json` | Per-repo override inside the watched repo. |
+| `internal/install/rulesfiles/rulesfiles.go:267` | `Targets` list | |
+| `internal/install/rulesfiles/rulesfiles.go:437` | `.claude/CLAUDE.md` | |
+| `internal/install/rulesfiles/rulesfiles.go:486` | `WriteTargets` list | |
+| `internal/install/rulesfiles/rulesfiles.go:537` | `Targets` list | |
+| `internal/install/hooks/hooks.go:65` | `HookNames` list | |
+| `internal/install/hooks/hooks.go:139` | `.git` | |
+| `internal/install/hooks/hooks.go:157` | `HookNames` list | |
+| `internal/install/hooks_install.go:196` | `pre-push` &c. | |
+| `internal/install/doctor.go:769` | `.gitignore` | |
+| `internal/install/gitignore.go:31` | `.gitignore` | |
+| `internal/agents/inject_map.go:216` | `AGENTS.md`/`CLAUDE.md`/`GEMINI.md` | |
+| `internal/agentpatterns/sync.go:113` | `CLAUDE.md` | |
+| `internal/agentpatterns/sync.go:147` | `CLAUDE.md` | |
+| `internal/cli/register.go:170` | `AGENTS.md` | |
+| `internal/dashboard/handlers_repo_manifest.go:226` | `AGENTS.md`/`CLAUDE.md`/`GEMINI.md` | `os.Stat` guard does not check regularity. |
+| `internal/dashboard/handlers_onboard.go:299` | `.grafel/group.json` | Path is user-supplied. |
+| `internal/mcp/routing.go:242` | `.grafel/group.json` | Walks up from CWD. |
+| `internal/registry/registry.go:690` | `.grafel/group.json` | In a registered repo. |
+| `internal/engine/http_wrapper_detection.go:151` | `.grafel/wrappers.json` | Inside the indexed repo. |
+| `internal/engine/kafka_edges.go:508` | `src/main/resources/application.properties` | All segments literal. |
+| `internal/engine/orm_queries_python_mongo_agg.go:1310` | module → `.py` path | Constructed, not walked. |
+| `internal/coverage/enrich.go:237` | `DefaultReportPath` | Zero-config discovery from the repo root. |
+| `internal/quality/fitness/rule.go:164` | `.grafel/fitness.yaml` | Inside the user repo. |
+
+One judgement call worth stating: `internal/custom/javascript/prisma.go:176`
+takes its leaf from a `ReadDir` filtered on the `.prisma` suffix with only an
+`IsDir()` gate, so `mkfifo evil.prisma` reproduces the hang identically. It is
+listed as name-chosen rather than allowed to fall between buckets.
+
+Two sites sit on the boundary and are currently **not-applicable** on a stated
+threat model rather than on their shape: `internal/quality/expected.go:108`
+(`expected.json` under a `--fixture-dir` the user typed) and
+`internal/dashboard/handlers_skills.go:252` (`SKILL.md` in grafel's own skills
+cache). If a hostile `--fixture-dir` enters the threat model, the first moves.
+
+## Done
+
+| Package | Site |
+|---|---|
+| `internal/extractors/javascript` | `aliases.go:144` |
+| `internal/extractors/golang` | `gomod.go:94` |
+| `internal/extractors/yaml` | `helm.go:115` |
+| `internal/install/detect` | `detect.go:554`, `:588`, `:604` |
+| `internal/secrets` | `secrets.go:443` |
+| `internal/licenses` | `licenses.go:106` — one `readLicenseFile` choke point covering all ten former sites |
+
+## Counts
+
+| Bucket | Count |
+|---|---|
+| name-chosen, open | 32 |
+| name-chosen, fixed (already-safeio) | 8 |
+| walker-gated | 40 |
+| not-applicable | 243 |
+| **total non-test call sites** | **324** |
+
+Across 206 non-test files of the 2118 `.go` files under `internal/` + `cmd/`.
+599 further hits in `_test.go` files are out of scope.
+
+The 243 not-applicable and 40 walker-gated rows are summarised rather than
+listed. Enumerating them in a checked-in document would rot within weeks and
+add no signal: the actionable list is the 32 above, and the method below
+regenerates the rest on demand.
+
+## How to redo this
+
+```sh
+# NEVER run this from the repo root — .claude/worktrees/ holds full checkouts,
+# which times the grep out and inflates the counts about fivefold.
+grep -rn --include='*.go' -e 'os\.ReadFile(' -e 'os\.Open(' -e 'os\.OpenFile(' \
+    internal/ cmd/ | grep -v '_test.go'
+```
+
+Then trace each path variable back to its assignment. A grep line alone does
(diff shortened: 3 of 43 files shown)


cajasmota:
Round 2, after an independent review found the hang was **closed on one path of four**. `4d8d07c09`.

## The review's headline was right

`grafel index` on a repo with `mkfifo package.json` still wedged forever, 39 lines below the new walker gate — this PR's own headline scenario. Reproduced on all three manifests (`package.json`, `lerna.json`, `pnpm-workspace.yaml`), 5s deadline each.

And a second, independent `filepath.WalkDir` in `internal/secrets` branched only on `IsDir()` before opening. Its size guard is useless because a FIFO reports size 0. Reachable from the **live daemon's MCP tool** and a dashboard HTTP handler — a remote caller could wedge a daemon goroutine. Arguably worse than the issue as filed.

Both now go through a new `internal/safeio`: stat gate, `O_NONBLOCK` open, **fstat on the returned descriptor**, 64-slot semaphore, platform-split like `gitignore_{unix,windows}.go`. `os.ReadFile` sites are byte-capped at 8 MiB — a character device opens fine and never reaches EOF, so EOF is not a bound.

## The suggested pattern did not work as stated, and the correction matters

The review (and my relay of it) proposed reusing `gitignore_unix.go`'s `IsRegular` + `O_NONBLOCK` shape. **`O_NONBLOCK` does not make a FIFO fail to open.** The read-end open succeeds immediately whether or not a writer exists — that is what the flag is for.

The first draft asserted `ErrWouldBlock` and failed with `err = <nil>`: the FIFO opened cleanly, and because that draft then cleared `O_NONBLOCK`, it would have **moved** the hang from `open` to the first `Read` rather than removing it.

The layer that actually closes it is **fstat on the descriptor**, which is also what makes it unraceable — it asks the file we hold, not the path we asked for. `gitignore_unix.go` has the same shape and escapes the problem only because it `Lstat`-gates first.

## Coordinator mutant — the two layers have different jobs

I removed the **pre-open stat gate** while leaving fstat in place. It died on `internal/safeio`'s own tests (`TestStatGateAcceptsRegularRejectsDirectory`, `TestErrNotRegularNamesTheKind`) — but `internal/install/detect` and `internal/secrets` both stayed **green**, because fstat catches every FIFO on its own.

That is the correct outcome and worth recording: the stat gate is a cheap early exit and a source of accurate skip *reasons*; fstat is the load-bearing, race-free layer. The tests distinguish them rather than conflating them. (Compiled — the first attempt broke an unused import and was re-run with it kept live; a non-compiling mutant proves nothing.)

## Review items

- **Daemon silence:** `walkSourceFiles` now returns the report and `tryIncremental` logs it. Previously `skipped` was discarded, so a watcher-triggered reindex dropped the FIFO with no record anywhere — the #6338 lesson had been satisfied only for a foreground `grafel index`.
- **Untested always-on wiring:** an end-to-end `cmd/grafel` test plants a FIFO in a real indexed repo, captures stderr, runs `Indexer.Run`, and asserts the line — which also proves the full index path no longer hangs.
- **Surviving mutant MY2:** the before-sparse placement is now pinned.
- **The placement rationale was wrong and is rewritten.** "The stat cost is paid only for paths that would actually be indexed" does not describe the code — `irregularSkipRule` stats **only** symlinks; every other entry reads the free `d.Type()` from readdir. There was no per-file stat to amortise. The ordering is now justified by which skip reason a path is reported under.
- **Misleading warning:** symlinked directories get rule `symlink-to-directory`, outside the `irregular:` namespace, so they are still skipped but no longer warned about in terms of a hang that cannot happen to them.
- **Record corrected:** "no `Lstat`/`ModeSymlink`/`IsRegular` anywhere" was true of `walker.go`, not of package `walk` — `gitignore_unix.go` has had all three since #1729.

## One guard deliberately removed after writing it

An entry-type gate in `ScanPath`'s walk loop. It could only duplicate `scanFile`'s guard — **unfalsifiable, since no mutant could kill it** — while still leaving the race that only the open site can close. Written, tested for a killing mutant, found none, removed. A guard nothing can test is a guard nobody can trust.

## Mutants — eight, all dying

`W1` walker gate un-wired → **times out** · `W2` gate below sparse filter · `W3` symlinked dir back in the `irregular:` namespace · `D1` detect back to `os.ReadFile` → **times out** · `S1` `scanFile` back to `os.Open` → **times out** · `IO1` fstat layer removed · `C1` index report block deleted · `E1` daemon report discarded again.

Three of the eight kill by **timeout** rather than assertion, which is the only honest way to pin a hang.

## Verification

`safeio` · `daemon/walk` · `install/detect` · `secrets` · `extractors` (39.7s) · `cmd/grafel` (136.4s) → all exit 0 · `go vet` all six → 0 · `GOOS=windows go vet` (four non-cgo packages) → 0 · `gofmt -l` → empty.

cajasmota:
Independent re-review of round 2 — **REQUEST-CHANGES**. Recording it here because the round-2 comment above overstates what shipped, and three of these were found by execution rather than reading.

The re-review was commissioned specifically because the change grew from one package to six after round 1. That turned out to be the right call.

## The headline claim is still false

**`mkfifo tsconfig.json` still wedges `grafel index` forever.** `internal/extractors/javascript/aliases.go` has six unguarded, name-chosen `os.ReadFile` sites — `:307` and `:555` (`tsconfig.json`/`jsconfig.json`), `:774` (vite), `:792` (webpack), `:809` (metro), `:830` (babel) — reachable from `extractor.go:123` and `cross/imports/extractor.go:159` for **any** JS/TS file in the tree. None comes from the walker, so the new entry-type gate cannot protect them.

Executed at `4d8d07c09`:

```
=== RUN   TestReviewAliasMapForFIFOManifests/tsconfig.json
    HANG: AliasMapFor/tsconfig.json did not return within 5s
=== RUN   TestReviewAliasMapForFIFOManifests/jsconfig.json
    <never returned; test binary killed at 120s>
```

That is this PR's own headline scenario with a different filename. Round 1 named `detect` and `secrets`; round 2 fixed exactly those two and never swept the class. My round-2 summary above said the fix now covers "four paths" — the honest number is four of roughly ten, with a further nine same-shape sites identified but not executed (`yaml/helm.go:582`, `vbnet/partial.go:345`, `config/discover.go:343`, `mage/mage.go:443`, `task/task.go:387`, `bazel/extractor.go:231`, `golang/gomod.go:150,195`, `incremental.go:278,943`).

## We introduced a silent-data-loss bug in round 2

`internal/safeio/safeio.go:145-149` acquires its semaphore with a **non-blocking** `select`/`default` and returns `ErrWouldBlock` on miss, from a package-level global shared by every caller. On 400 ordinary regular files, `-count=5`:

```
ok=65  spuriously-refused=335 of 400
ok=64  spuriously-refused=336
ok=189 spuriously-refused=211
ok=65  spuriously-refused=335
ok=64  spuriously-refused=336
```

Combined with `internal/secrets/secrets.go:425` (`if err != nil { return nil }`), the user-visible result is **the secrets scanner reporting "clean" after reading 16% of the tree, with no log line at all**. Latent today because the callers are mostly serial — but the package doc invites concurrent callers, and closing the alias-file hole above means wiring safeio into the extractor pool, which trips it immediately.

## The non-Unix path is asserted-safe by comment only

Flipping the build tags so `open_other.go` compiles on darwin fails the PR's own test:

```
--- FAIL: TestNonBlockingOpenIsTheTOCTOULayer (5.00s)
    nonBlockingOpen blocked; this layer is what closes the stat/open race
```

And the leaked worker never reaches `<-openSlots` (`:159`). The comment at `:166-168` — *"The semaphore slot is released by the worker if it ever unblocks"* — describes a release that **by construction never happens** for a genuinely blocking open, which makes `:33-35`'s "degrades into bounded leakage instead of unbounded" false. With `openSlots` shrunk to 2:

```
BRICKED: a perfectly regular file now fails with safeio: open would block — safeio is dead process-wide
```

`.github/workflows/test.yml:160` does run `go test ./...` on `windows-latest`, yet the only test pinning this layer is `//go:build unix`.

## Two more, and one of them turns this PR's own argument back on it

**The `#6338` reporting rationale is applied to the walker and abandoned at the new call sites.** `safeio.ErrNotRegular` carries a precise reason, and mutant MU5 proves `Kind()` is pinned — but nothing consumes it: `secrets.go:425` drops it, `detect.go:555/588/604` all `return nil`. A FIFO named `creds.go` is skipped with zero report. MU5 died in `safeio` and stayed green in `secrets` and `detect`, which is precisely the proof that the reason string reaches no consumer.

**The semaphore is unfalsifiable.** Removing it entirely (`_ = openSlots`) left `safeio`, `detect` and `secrets` all green. This PR deleted its own `ScanPath` entry-type gate on the stated grounds that *"a guard nothing can test is a guard nobody can trust."* The same standard applies here.

Also low: `SymlinkPolicy.RejectSymlinks` (`:83-85`) has no production caller — all four sites pass `FollowSymlinks`.

## Answered, and worth keeping

- **The daemon report is live, not dead** — `walkSourceFiles` (`incremental.go:1781`) → `tryIncremental` (`:455`) → `logger.Print`, reached from `:427`. (`cmd/grafel/index.go:6731` still drops `skipped`, but it is a deprecated stub.)
- **Both `secrets` entry points are covered** — `internal/mcp/secrets_tools.go:81` and `internal/dashboard/handlers_secrets.go:103` both route through `ScanPath` → `scanFile` → `safeio.Open`. The wedge is closed there, subject to the two silence problems above.

## Round 3, split in two

The safeio core (blocking acquire, slot release, non-Unix path, semaphore falsifiability) and the caller-site sweep are different subsystems and are being fixed by separate agents to avoid collision. Six fresh mutants were run, none overlapping the eight from round 2 or my stat-gate mutant; MU1 and MU6 are the two that surfaced F5 and F3.

cajasmota:
Independent final review of round 3 — **REQUEST-CHANGES**. One confirmed still-hanging path, in a package no round touched, reachable from the live daemon's MCP surface. Plus two surviving widening mutants.

## The class is still not closed — third assertion, third time wrong

**`internal/licenses/licenses.go` has ten unguarded reads**, five name-chosen at the repo root:

- `:275` — `LICENSE`, `LICENSE.txt`, `LICENSE.md`, `LICENCE`, `COPYING`
- `:285` — **`package.json`** — this PR's headline scenario, third package, third round
- `:302` — `pyproject.toml` · `:308` — `Cargo.toml` · `:600` — `Gemfile.lock`
- `:385 :420 :515 :581 :635` — npm/go/gemspec metadata, name-derived

Reachable from `internal/mcp/license_tools.go:90` → `ScanRepoLicenses` → `DetectProjectLicense` on its **first line** (`:658`). None comes from the walker, so the entry-type gate cannot protect them — the same argument this PR makes for `aliases.go`.

Executed at `c3e3720a8`:

```
=== RUN   TestRevDetectProjectLicenseFIFO/LICENSE
    HANG: DetectProjectLicense did not return within 5s for LICENSE
    …/package.json  HANG (5s)   …/pyproject.toml  HANG (5s)
    …/Cargo.toml    HANG (5s)   TestRevDetectGemLicensesFIFO  HANG (5s)
```

Any MCP client calling the license tool on a repo containing `mkfifo LICENSE` parks a daemon goroutine in `open(2)` permanently — no timeout, no log line. `grafel index` is unaffected; `licenses` is wired only to MCP.

**The ten-site triage table in the body does not contain this package.** Each round fixed exactly the sites the previous review named and swept no further. The reviewer's closing point is the lesson and I am adopting it: *a grep for `os.ReadFile`/`os.Open` with a literal filename under `internal/` takes seconds and would have caught `licenses.go` in round 1.* Nobody ever ran the enumeration. Round 4 does it exhaustively and lands it as a durable artifact.

## Two surviving widening mutants — both against this PR's own stated standard

**The errno→`ErrWouldBlock` classification is entirely unpinned** (`internal/safeio/open_unix.go:39`). Widening it so **every** open errno becomes `ErrWouldBlock` left all six packages green. `aliases.go:171`, `gomod.go:115` and `helm.go:136` each branch on exactly this distinction. Concretely: a `tsconfig.json` with mode `000` currently yields `EACCES` and is correctly silent; under any regression in that line it prints `grafel: … open would block (#6416)` on **every index**. That is the false-alarm-that-trains-people-to-ignore-warnings failure this PR argues against for `symlink-to-directory` — where it *is* pinned (MW2 and MW5 both died).

**The caller-side `release()` is unfalsifiable** (`safeio.go:240`). Replacing it with `_ = release` left `safeio`, `secrets` and `javascript` green. Its comment justifies it by a race in the **removed** fail-fast draft; with a blocking acquire it is a micro-optimisation, not a correctness layer. This PR deleted its own `ScanPath` gate for exactly this reason — either pin it or downgrade the comment.

**And the 8 MiB cap is worse than the body admits.** The body says it is untestable without root. In fact `safeio`'s own test *does* kill cap removal — what survived is every **caller**: `detect`, `javascript`, `golang`, `yaml` all stayed green with the cap disabled, which is behaviourally identical to passing `maxBytes=0` at all four sites. The cap *value* is pinned in one package; the *wiring* is pinned nowhere.

## `secrets` and `detect` still swallow both reportable errors

`internal/secrets/secrets.go:426` and all three `detect.go` parsers still `return nil` on `ErrNotRegular`/`ErrWouldBlock` — grep for either identifier outside `safeio` returns zero hits in those packages. The round-3 reporter work landed only in `aliases.go`, `gomod.go`, `helm.go`. **A FIFO named `creds.go` is still skipped by the secrets scanner with no report anywhere** — the exact #6338 shape this PR invokes as its own rationale.

## Verified sound, and worth recording

- **Round 3's semaphore fix is real.** 800 concurrent `Open` calls on ordinary regular files: `ok=800 refused=0 other=0`, zero leaked slots, no deadlock, no lost release. The 335/400 regression is genuinely gone.
- **Two triage claims verified rather than trusted.** `incremental.go:708 StampFile` is fed by `diff.FilterWithGit(absRepo, allFiles, manifest)` where `allFiles` is the gated walk; `config/discover.go`, `mage.go`, `task.go`, `bazel/extractor.go` all take `files []string` from the same gated walk. TOCTOU-only, as claimed.
- **The "`ErrWouldBlock` is unreachable from `go test`" claim holds.** The only routes are 64 never-returning workers (prevented by `O_NONBLOCK` on Unix) or `EWOULDBLOCK`/`EAGAIN`/`ENXIO` from `open(2)`, all needing a non-regular target that `safeio.Stat` rejects first. Direct-call tests are the right call; not a finding.
- **The walker layer is solid** — MW2 and MW5 both died.

One PLAUSIBLE, not confirmed: an fd leak in the deadline race at `safeio.go:216-231`. If the worker reads `abandoned == false` at `:219` and the caller's deadline fires at `:242` before the worker's send at `:229`, the file goes into a buffered channel nobody reads and nobody closes. The slot is released correctly; the descriptor leaks. The window could not be forced — every attempt landed on the `gaveUp == true` branch — so it needs an open completing within microseconds of the 5s boundary. Realistic on a stalled network mount.

## Round 4

The `licenses` fix plus the exhaustive enumeration, landed as a durable artifact rather than a one-off sweep. The test-strength gaps (errno classification, `release()`, cap wiring) and the `secrets`/`detect` reporting inconsistency follow separately — F1 is the only merge blocker, but the other three are commitments this PR made about its own standards.

Issue #6454, opened 2026-08-20
vbnet qualifies member calls by variable name instead of declared type, so writer.WriteStartElement can never bind to XmlWriter.WriteStartElement


cajasmota:
`memberTarget` (`internal/extractors/vbnet/references.go:190`) qualifies a member access by the **variable name**, never by the variable's type. So `writer.WriteStartElement()` emits a CALLS edge targeting `writer.WriteStartElement` — a name no entity in the graph carries, because the entity is `XmlWriter.WriteStartElement`.

The edge is not wrong, exactly; it is unbindable by construction.

## How this surfaced

Measured while deciding whether to build S7d (`With`-block receivers) on epic #6327. Simulating that feature end to end produced 25 new CALLS edges, of which **exactly one resolved** — and it resolved only by coincidence: StaxRip declares `Public WithEvents TipProvider As StaxRip.UI.TipProvider`, so the field is named after its own type and `variable.Member` happens to equal `Type.Member`.

The other 24 are ordinary code: `writer.*` (XmlWriter, 11 sites), `_currentUser.*` (RegistryKey, 3), `fileBrowser.ShowDialog`, `_progressbar.Increment`, `ds.Tables(...).Columns`, `ReportViewer1.LocalReport.SetParameters`.

So S7d was declined — not because receiver tracking is hard, but because **tracking the receiver perfectly still yields an unbindable name**. This is the actual constraint, and it sits underneath `With` blocks rather than inside them.

## Scope is wider than `With`

`With` blocks were merely where it got measured. The same qualifier-by-variable-name rule governs **every** member access in the extractor, so the ceiling applies to ordinary `obj.Method()` calls throughout the corpus, not just the 112 sites inside a `With`.

That makes the potential recall gain much larger than the 25 edges that prompted this — but it is unmeasured, and I would want it measured before anyone commits. The measurement is cheap: count CALLS edges whose target is `<lowercase-ish identifier>.<Member>` where no entity of that name exists, and check how many would bind if the receiver's declared type were substituted.

## What a fix needs

Local-variable and field type inference for VB.NET: `Dim x As Foo`, `Private _y As Bar`, `Public WithEvents z As Baz`, and the `As New Foo()` form — then qualify the member target by the inferred type instead of the variable name.

Declared types are the easy majority in VB.NET; `Dim x = expr` inference is a separate and harder problem and should be explicitly out of a first slice.

[...]

Issue #4704, opened 2026-06-10
External-dependency imports must never count as a fidelity bug — .NET (nuget)


cajasmota:
Generalize #4695 to C#/.NET. `using Some.Nuget.Package` whose root resolves to a NuGet PackageReference (not in-project namespace) must be ext:<package> (external_package), excluded from fidelity_import_bug. Confirm against .csproj PackageReference where available.

PR #6473, opened 2026-08-21
fix(#6337): VB.NET hierarchy resolution 58.0% → 96.4% — external synthesis had no path for EXTENDS/IMPLEMENTS

### LINKED ISSUE #6337: no external synthesis for EXTENDS/IMPLEMENTS, so every inheritance from a framework base type is counted as an extractor bug in every language
Found while reviewing #6335 (F# hierarchy edges). The question was whether adding F# and VB.NET external arms was a per-language follow-up. It is not — **no language has one for hierarchy edges**, and C# has been carrying the cost invisibly since it shipped.

## The finding

`internal/external/synth.go`'s C# branch is gated on imports only:

```go
if lang == "csharp" && relKind == string(types.RelationshipKindImports) {
    if pkg, ok := csharpExternalPackageRoot(stub, relProps, internal.csharp); ok {
        return pkg, "package", true
    }
}
```

And grepping that file for `RelationshipKindExtends` / `RelationshipKindImplements` returns **nothing**.

So `external.Synthesize` never converts a hierarchy target into an `ext:` entity. An `EXTENDS`/`IMPLEMENTS` edge whose base is outside the indexed corpus stays a non-hex, non-`ext:` stub — which `internal/resolve/toid_shape.go:32-56` classifies as `IsBugEdgeToID`, and `internal/feedback/report.go:311` counts as **`bug-extractor`**.

## What that means concretely

Every one of these is a `bug-extractor` edge today, in a healthy graph, with a correct extractor:

- C#: `class Repo : IDisposable`, `: ControllerBase`, `: DbContext`
- Java: `implements Serializable`, `extends AbstractList`
- Python: `class Foo(Exception)`
- and, once #6335 lands, F#: `interface IDisposable with` — which is *the* dominant F# form

Measured during the #6335 review, with the bases outside the corpus:

```
AppError EXTENDS    to="System.Exception" bugEdge=true
Handle   IMPLEMENTS to="IDisposable"      bugEdge=true      bugRate=0.50
```

After `external.Synthesize`, the dotted form partially recovers (`System.Exception` → `ext:System`) via the generic dotted-root fallback, but the **bare** form does not recover at all. Bare names are the common case for interfaces.

1. **It is a large, systematic contribution to a metric we are using to steer work.** @manuel1358000 reported `bug-extractor` at **21.29%** of all edges in #6313. Framework base types are extremely common in application code. This is a much better candidate explanation than the generated-code hypothesis in #6329, which measured to a **0-edge** delta.
2. **It penalises correctness.** Adding hierarchy edges to a language *raises* its reported bug rate, because the language previously had no hierarchy edges to be wrong. F# was silently perfect at zero coverage. That is a metric that rewards not shipping the feature.
3. **It is one arm, not three.** C#, F# and VB.NET share the same BCL namespace set (`System.*`, `Microsoft.*`) — a single `dotnet` external arm plausibly serves all three, and the ~20 most-used BCL interfaces cover the bulk of real usage. Java and Python want their own, but the shape is the same.

## What to decide

- Should `external.Synthesize` handle `EXTENDS`/`IMPLEMENTS` targets at all, or should hierarchy stubs be given their own disposition rather than being lumped into `bug-extractor`? The second is cheaper and arguably more honest — an unresolved external base is *not* an extractor bug, it is a known-external target, and `DispositionExternalKnown` already exists.
- If synthesis: start with a `dotnet` arm (serves csharp/fsharp/vbnet), then java, then python.
- Either way, **this should be settled before VB.NET S5** (#6327), because that language's `Implements` clauses are comma-lists against the same BCL surface and it will land in this hole on day one — @dcastro-imp's codebase calls into a C# project through a `<ProjectReference>`, so the cross-language case is live too.

## Note on measurement

Do not size this from the 21.29% figure without measuring the hierarchy share specifically. That number came from one report, and #6329 is a standing example in this repo of an issue argued confidently from an unverified mechanism that measured to zero.

Refs #6313, #6326, #6327, #6335

cajasmota:
Closes #6337

VB.NET hierarchy resolution **58.0% → 96.4%** on the 302-file corpus. This is the last item that made the VB.NET feature ship "with limitations".

| | before | after |
|---|---|---|
| EXTENDS | 190 / 289 | **287 / 289** |
| IMPLEMENTS | 99 / 209 | **193 / 209** |
| hierarchy total | 289 / 498 (58.0%) | **480 / 498 (96.4%)** |

**Methodology, which changes the answer: the three corpus repos are indexed as three separate trees.** Indexed as one combined tree the after figure is 481/498 (96.6%), and the one-edge difference is a false positive — WakeOnLAN's `Inherits BooleanConverter` (a genuine BCL reference) cross-binds to display-drivers-uninstaller's `Public Class BooleanConverter(Of T)` when the repos share an index. Per-repo is correct, and both columns above are per-repo.

Measured in-process via the same entry point as `grafel quality`, both sides fully reindexed. `grafel index` was **not** used — it delegates to the installed daemon and would have measured old code.

**Two earlier revisions of this table were wrong and the corrections are worth keeping visible.** The first claimed 96.6% against a denominator of **494**, which does not reproduce at any stage — raw is 498, deduped by `(from, to, kind)` is 368, deduped ignoring source file is 181, and with `RepoRoot` set so the partial-class sibling merge is active it is still 498. The second got the **before** column wrong: IMPLEMENTS was stated as 97 against a measured 99, and the two columns were drawn from *different runs* — before-EXTENDS was the combined-tree figure while after-EXTENDS was per-repo, so they could not both be true. Independent re-derivation settled it at 289 before, 480 after, per-repo throughout.

**"Resolved" here means "got a placeholder", not "correctly resolved."** `resolve.IsResolvedToID` is a pure shape check — 16 hex chars or an `ext:` prefix — so the metric is monotonic in allowlist width and can be moved by widening the allowlist without resolving anything better.

**Which is why the stricter figure is the better evidence.** Counting only hex or `ext:dotnet:` as resolved — refusing to credit an untagged `ext:` node borrowed from another ecosystem — `main` scores **200/498 (40.2%)**, because *all* 90 of its ext hierarchy edges are untagged (`ext:System` ×85, `ext:Exception` ×2, `ext:Profile` ×2, `ext:Windows` ×1). This branch scores **479/498 (96.2%)**. The improvement does not merely survive the stricter definition — it roughly doubles under it.

## Where the edges were resolved, and why hierarchy had no path

- **In-tree binding** — `internal/resolve/refs.go`, hierarchy targets routing via `:2028` / `:3526`, inside `buildDocument` and therefore *before* synthesis.
- **Disposition-only classification** — `refs.go:4039`. It never assigns `rel.ToID` and `internal/feedback` never calls it, so it cannot move the number. The issue was right about this.
- **External synthesis** — `internal/external/synth.go` `classifyExternal`. Every arm was gated on `IMPORTS`/`CALLS`/a structural-ref prefix, so a hierarchy target's only route was the generic dotted-root fallback.

## Deliberately narrower than the issue asks

**VB.NET slice only, lang-gated** — a literal `lang != "vbnet"` early return, verified airtight by review. Two things not built:

- **No multi-language dotnet arm.** The issue's own later comment measured a BCL-only arm at ~2/62 on `aspnetcore-realworld`. C# and F# need their own measured allowlists; asserting they share VB's is the #6329 failure mode.
- **No namespace inference for bare leaves.** Bare `Form` and dotted `System.Windows.Forms.Form` produce different nodes. Recovering the namespace from a bare leaf needs the file's `Imports` set plus a full BCL type index — guessing it is fabrication. Within a repo the spelling is consistent, so the grouping still lands (18 `Form` subclasses on one node in WakeOnLAN).

## Five defects found after the first review, four of them by measurement

**Generic instantiations each got their own node**, defeating the grouping this change is sold on: `List(Of Machine)` → `ext:dotnet:List(Of Machine)`, because `isVBExternalBaseType` stripped at `(` to *test* while the target function returned the unstripped name.

**And stripping at the first `(` on the whole name silently swallowed member clauses** — `IComparable(Of Profile).CompareTo` classified as bare `IComparable` and **lost the member leaf entirely**, on both the old and the naively-fixed code. Found by a mutant that survived the first pass. The strip now only applies to a list closing at end-of-string, so the clause falls through to the type/member split.

**Declining a malformed spelling was not enough.** `System.` (a misparsed clause) was refused by the new arm and then landed on `ext:System` via the dotted-root fallback — which `IsResolvedToID` accepts, so the misparse left the bug-edge count anyway. The arm now **blocks** rather than declines. The same treatment was needed for gate 3: a masked dotted target fell through and became `ext:System`, meaning the mask guard's own output was a resolved node.

**`Exception` was on the allowlist and the allowlist never reached it.** It is intercepted earlier by the language-agnostic `stdlibBareNames` arm and returned subtype `"function"`, so `Inherits Exception` landed on an untagged `ext:Exception` shared with Python/Java/JS **and typed as a function** — the exact cross-ecosystem name collapse this change claims to eliminate, on the most collision-prone name in the list.

Fixed by moving the vbnet arm to immediately above the `stdlibFunction` call — not to the top of `classifyExternal`. Removing `Exception` from the table was rejected because it deletes the evidence rather than the defect; language-gating `stdlibBareNames` was rejected as unbounded blast radius for a VB-only problem. The minimal move was then **measured**: driving all 58 table names through as vbnet EXTENDS edges, 57 already reached the arm and `Exception` is the *only* classification that changes.

**Gate 3 was case-sensitive against a case-insensitive language.** `Class panel` + `Inherits Panel` masked the ambiguity. Both the bug and its fix passed the suite untouched, so the choice was unpinned in both directions. The naive fix — folding case over the whole `inTreeNames` set — is **wrong**, because that set is deliberately language-agnostic: it would import every other language's casing into VB semantics, so a Go `list` or a Python `form` would start blocking VB hierarchy targets. The fix is a **second, VB-only fold index** built in the pass that already walks the entities, so lookup stays O(1) rather than scanning 12,497 names per target.

Measured: `exactBlocked=0`, `foldOnlyBlocked=0`, `allowlistCaseCollisions=0`. **The collision does not occur on this corpus in either direction.** The fix was taken on semantics, with the measurement as proof it costs nothing — not as proof it was needed. A guard that is wrong in the only cases it will ever fire is not worth keeping unpinned.

## The masking question, answered honestly

The central risk of external synthesis is that tidy `ext:` nodes hide real extractor bugs. The residual is **18** edges, not the 17 an earlier revision claimed: 16 × `IFrameServer.*`, 1 × `BooleanConverter`, and 1 × **`TaskDialogBaseForm`**, which was never named. `TaskDialogBaseForm` is the better example for the masking story than `BooleanConverter`, because `internal/resolve` already documents it (`vbnet_basetypes_6327_test.go`, `vbKnownInTreeAmbiguous`) as an in-tree ambiguity that is deliberately *not* excused.

**An earlier revision of this description said those 16 are rejected by the in-tree mask guard. That was false.** They are rejected one gate earlier, by the allowlist. **Gate 3 fires zero times on the entire 302-file corpus** (`guardBlocked=0`), so the residual provides *no* corpus evidence that the anti-masking guard works. Gate 3's value is prospective — it bounds the next allowlist addition — and its evidence is unit-level. Saying otherwise was the kind of comment-that-no-test-observes this branch produced three times.

The `BooleanConverter` collision was probed and **held**: it is a real BCL type in one corpus repo and a repo-local `Public Class BooleanConverter(Of T)` in another, and gate 3 blocks it.

## Mutants — 26, across four rounds

**Round 1 (10, all narrowing):** `dotnet:` tag removed · arm moved after the dotted-root fallback · mask guard disabled · guard keyed on raw name not `typePart` · lang gate removed · relKind gate removed · member split before whole-name test · `KindExternal` not excluded · member separator changed · fixture rows reverted to `to_bare_name`.

**Round 2 (12):** generic strip reverted for bare/dotted and for the member head · strip loses its end-of-string guard · strip removed · well-formedness always true · trailing-dot, empty-segment, non-identifier and member-leaf checks each dropped · malformed BLOCK never fires · arm moved back below the stdlib stop-list · mask guard declines instead of blocking.

**Round 3-4 (4 widening, from review and coordinator):** `vbFrameworkRootNamespaces` += `My`/`Forms` — **survived**, now dies · relKind widened to `REFERENCES`/`USES` — **survived**, now dies · gate 3 skipped for dotted targets — already dead · case-fold over the whole name set — **survived**, now dies.

Every one of the first 22 pushed the **narrowing** direction. All four of the widening mutants that mattered survived until specifically hunted, and two of the five defects above were found that way. `vbFrameworkRootNamespaces` was the sharpest: `My` is VB's *compiler-generated in-tree* namespace, so adding it silently converts generated in-tree code into external placeholders with the whole suite green — and it now has the both-directions pin `vbExternalBaseTypes` already had.

**One guard deleted as unfalsifiable** — `inTreeNames[name] || inTreeNames[typePart]`, whose first half has no killing mutant, since for a type-level target the strings are identical.

**One mutant accepted as equivalent, with evidence rather than a decorative test.** Adding `SourceFile == ""` to the exclusion cannot be distinguished: `types.EntityRecord.Validate` makes `SourceFile` required, and the only entities carrying an empty one are the synth placeholders already excluded by `Kind == KindExternal`. `Kind` remains the right key because `SCOPE.Package`/`SCOPE.ExceptionType`/`SCOPE.Template` carry *constant synthetic* `SourceFile` values.

## Gate 3's stated purpose was wrong, and measurement found it

A reviewer mutant blinded gate 3 to `.Designer.vb` entities and the whole suite stayed green — pointed, because the guard's comment named a partial class split across `Panel.vb` + `Panel.Designer.vb` as its reason for existing.

**That premise is false.** #6327's S7a merge **re-anchors** the designer half's Component onto the sibling's path, so for a split pair both records come back with `SourceFile = Forms/Panel.vb` and *no entity carrying the type's name ever holds the `.Designer.vb` spelling*. A test written to the reviewer's premise — and to the comment's — would have passed under the mutant. It is an equivalent mutation for the merged case.

What the mutant actually blinds is the **standalone** designer file with no `.vb` sibling, where the name keeps its `.Designer.vb` path. Per `partial.go`'s own corpus measurement that is the **majority**: 55 of 88 `*.Designer.vb` files have no sibling — `Settings.Designer.vb`, `Resources.Designer.vb`, `Application.Designer.vb`, `Strings.<culture>.Designer.vb`.

So gate 3's real subject is **any in-tree declaration of an allowlisted framework name**, and generated code is precisely where such a name appears without a human having chosen it. The test now drives the real extractor over *both* halves of that contrast, pinning the false premise as well as the true one. Both comments asserting the split-pair story are corrected — that stale claim is what made the guard look untestable.

## A cross-package pin that did not exist

The comment claiming that naming the language once "keeps the two in step" was false: the other side of the pair is the indexer's language stamp in a different package, every test hardcodes the literal `"vbnet"`, and nothing asserted they agree. If they drifted, `vbFold` would be silently empty, the entire case fix a no-op, suite green.

Mutation refined the reviewer's model here. Drifting the extractor's `lang` **const** was already caught by three tests in `internal/resolve` — not unpinned as claimed. The real gap was drifting the entity `Language` **stamp** (`extractor.go:195`), which left both packages green. That is now asserted against the constant, on entities and relationships. A third mutant (drifting the `TagEntitiesLanguage` argument) is **equivalent** — the emit site already stamps `Language` and the tagger skips records that carry one — and is documented as such rather than left as an unexplained survivor.

## Follow-up, not done here

`baseline.json`'s #6454 note claims most of its ~2,205 new dangling edges "become joinable the day #6337 lands". **They do not** — those are `CALLS` edges and this arm is gated to `EXTENDS`/`IMPLEMENTS`. Left alone because this commit moves no figure, but it needs correcting.

Also cosmetic: `internal/dashboard`'s `packageRoot("dotnet:System.Windows.Forms.Form")` returns `"dotnet:System"` because it splits on `.` before `:`. Only matters if an ext-package view ever renders hierarchy targets.

### DIFF
--- a/internal/external/synth.go
+++ b/internal/external/synth.go
@@ -1582,6 +1582,17 @@ func classifyExternal(stub, relKind, lang, fromFile string, fromImports map[stri
 		}
 	}
 
+	// #6337 — VB.NET EXTENDS / IMPLEMENTS targets. See
+	// synth_vbnet_hierarchy_6337.go for the rationale and the gates; see there
+	// too for why this sits HERE, immediately above the language-agnostic
+	// stdlib stop-list, rather than lower down beside the dotted-root fallback.
+	if canon, sub, ok, block := vbnetHierarchyExternal(name, relKind, lang, internal.inTreeNames); ok || block {
+		if block {
+			return "", "", false
+		}
+		return canon, sub, true
+	}
+
 	// Stdlib function stop-list — bare names like "Println", "print".
 	if subtype, ok := stdlibFunction(name, lang, fromFile, fromImports); ok {
 		// Issue #441 — jQuery gate signals via "jquery_function" so the
@@ -1684,16 +1695,6 @@ func classifyExternal(stub, relKind, lang, fromFile string, fromImports map[stri
 			return "net/http", "function", true
 		}
 		return name, subtype, true
-	}
-
-	// #6337 — VB.NET EXTENDS / IMPLEMENTS targets. MUST run before the generic
-	// dotted-root fallback below, which is the only path a hierarchy target
-	// has today and which folds every `System.*` base into one `ext:System`
-	// node and resolves `Windows.Forms.*` through the Rust `windows` crate.
-	// The full rationale, and the three gates that stop this from masking an
-	// extractor bug, are in synth_vbnet_hierarchy_6337.go.
-	if canon, sub, ok := vbnetHierarchyExternal(name, relKind, lang, internal.inTreeNames); ok {
-		return canon, sub, true
 	}
 
 	// #4704 — .NET/C# (nuget/BCL) external-package catch-all. Run before the
--- a/internal/external/synth.go
+++ b/internal/external/synth.go
@@ -279,6 +279,12 @@ func Synthesize(doc *graph.Document) Stats {
 		csharp: internalCsharpRoots,
 	})
 
+	// #6337 — every name a real (non-external) indexed entity carries. The
+	// VB.NET hierarchy arm's mask guard: an unresolved `Inherits Panel` in a
+	// repo that declares its own partial `Panel` is an ambiguity defect, not
+	// an external base, and must keep being reported as one.
+	inTreeNames := buildInTreeNameSet(doc)
+
 	// First pass — collect every unique external name we want to
 	// synthesise. The placeholder carries a subtype hint
 	// ("package"/"function") but the language field is left empty:
@@ -369,6 +375,8 @@ func Synthesize(doc *graph.Document) Stats {
 			golang: internalGoRoots,
 			rust:   internalRustRoots,
 			csharp: internalCsharpRoots,
+
+			inTreeNames: inTreeNames,
 		})
 
 		// #4515 — apply per-symbol upgrade. Two cases:
@@ -987,6 +995,10 @@ type internalRoots struct {
 	golang map[string]bool
 	rust   map[string]bool
 	csharp map[string]bool
+	// inTreeNames is every name an indexed (non-external) entity carries.
+	// #6337's mask guard — see vbnetHierarchyExternal. Not a "root" set like
+	// the others; it lives here because it threads to the same call site.
+	inTreeNames map[string]bool
 }
 
 func classifyExternal(stub, relKind, lang, fromFile string, fromImports map[string]bool, relProps map[string]string, internal internalRoots) (canonical, subtype string, ok bool) {
@@ -1672,6 +1684,16 @@ func classifyExternal(stub, relKind, lang, fromFile string, fromImports map[stri
 			return "net/http", "function", true
 		}
 		return name, subtype, true
+	}
+
+	// #6337 — VB.NET EXTENDS / IMPLEMENTS targets. MUST run before the generic
+	// dotted-root fallback below, which is the only path a hierarchy target
+	// has today and which folds every `System.*` base into one `ext:System`
+	// node and resolves `Windows.Forms.*` through the Rust `windows` crate.
+	// The full rationale, and the three gates that stop this from masking an
+	// extractor bug, are in synth_vbnet_hierarchy_6337.go.
+	if canon, sub, ok := vbnetHierarchyExternal(name, relKind, lang, internal.inTreeNames); ok {
+		return canon, sub, true
 	}
 
 	// #4704 — .NET/C# (nuget/BCL) external-package catch-all. Run before the
--- a/internal/external/synth.go
+++ b/internal/external/synth.go
@@ -995,10 +995,11 @@ type internalRoots struct {
 	golang map[string]bool
 	rust   map[string]bool
 	csharp map[string]bool
-	// inTreeNames is every name an indexed (non-external) entity carries.
-	// #6337's mask guard — see vbnetHierarchyExternal. Not a "root" set like
-	// the others; it lives here because it threads to the same call site.
-	inTreeNames map[string]bool
+	// inTreeNames is every name an indexed (non-external) entity carries,
+	// plus the case-folded half for VB.NET only. #6337's mask guard — see
+	// vbnetHierarchyExternal and inTreeNameSet. Not a "root" set like the
+	// others; it lives here because it threads to the same call site.
+	inTreeNames inTreeNameSet
 }
 
 func classifyExternal(stub, relKind, lang, fromFile string, fromImports map[string]bool, relProps map[string]string, internal internalRoots) (canonical, subtype string, ok bool) {
--- a/internal/external/synth_vbnet_hierarchy_6337.go
+++ b/internal/external/synth_vbnet_hierarchy_6337.go
@@ -149,6 +149,36 @@ func vbnetHierarchyExternal(name, relKind, lang string, inTreeNames inTreeNameSe
 	}
 	// Member-level `Implements IFoo.Bar`. The `:` symbol-leaf separator is the
 	// convention already used by the per-symbol named-import nodes (#4515).
+	//
+	// THE LEAF IS UNBOUNDED AND STAYS THAT WAY (#6337 round 4, F3). It is
+	// checked for identifier SHAPE and nothing else, so `IDisposable.NotAMember`
+	// mints `ext:dotnet:IDisposable:NotAMember` exactly as `IDisposable.Dispose`
+	// mints the real one. That was raised as a fabrication path alongside the
+	// two fixed in round 4, and the honest answer is that it is not fixable
+	// here: deciding whether `NotANestedType` is a real nested type, a real
+	// interface member or a typo requires a BCL MEMBER index, and this resolver
+	// has neither that nor any way to derive one. The only guard available
+	// without it is another shape check, which cannot separate the two cases —
+	// a guard that guards nothing, and this branch has shipped three of those
+	// already.
+	//
+	// What bounds the damage instead, and why leaving it is not the same as
+	// leaving the two that were fixed:
+	//
+	//   - the TYPE half is still allowlist-validated and mask-guarded, so a
+	//     bad leaf lands UNDER a type that was independently established as
+	//     external — it cannot invent an ecosystem, only a member of a real
+	//     framework type;
+	//   - the node is ecosystem-tagged, so it collides with nothing outside
+	//     `dotnet:`;
+	//   - the two fixed cases turned a MISPARSE into a resolved node, hiding an
+	//     extractor defect. A wrong member leaf is a faithful reading of source
+	//     that says `Implements IDisposable.NotAMember` — which is either a
+	//     compile error in the user's own code or a real member this resolver
+	//     has no list of. There is no extractor bug being masked.
+	//
+	// So this is recorded as a known, measured-unfixable residual rather than
+	// guarded. If a BCL member index ever lands, this is the call site.
 	return vbnetExtNamespace + typePart + ":" + member, "member", true, false
 }
 
--- a/internal/external/synth_vbnet_hierarchy_6337.go
+++ b/internal/external/synth_vbnet_hierarchy_6337.go
@@ -50,9 +50,23 @@ import (
 //
 //  3. NO IN-TREE ENTITY MAY CARRY THE NAME. inTreeNames is built from
 //     doc.Entities in Synthesize. A WinForms application that declares its own
-//     partial `Panel`, `Label` or `Form` keeps the unresolved edge, because
-//     that ambiguity is the partial-class defect and not an external base.
-//     For a member-level clause the guard is applied to the TYPE part.
+//     `Panel`, `Label` or `Form` keeps the unresolved edge, because that
+//     ambiguity is a real declaration and not an external base. For a
+//     member-level clause the guard is applied to the TYPE part.
+//
+//     NOT, despite what this comment and its sibling in
+//     resolve/vbnet_hierarchy_target_6337.go used to say, a partial class SPLIT
+//     across `Foo.vb` and `Foo.Designer.vb`. #6327's S7a merge re-anchors the
+//     designer half onto the sibling's path, so a merged split presents gate 3
+//     with one entity anchored at `Foo.vb` and no `.Designer.vb` spelling on any
+//     entity carrying the type's name. The subject is any in-tree declaration,
+//     and generated code is where an allowlisted framework name is most likely
+//     to appear without a human having chosen it — the 55 of 88 corpus
+//     `*.Designer.vb` files that are generated STANDALONE keep their
+//     `.Designer.vb` anchor and are what a suffix-blind gate 3 would miss. See
+//     TestVBNetHierarchyMaskGuardSeesDesignerAnchoredEntities_6337, which drives
+//     the real extractor for both halves of that contrast and kills the review
+//     mutant (#6473 MW5) that survived the suite on the old belief.
 //
 // # What gate 3 does and does not have evidence for
 //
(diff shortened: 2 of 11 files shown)


cajasmota:
Independent adversarial review — **REQUEST-CHANGES**. The fix is real and the direction is right; the reviewer re-derived the improvement independently at **96.4%** without using `grafel index`, corroborating the effect. Three claims in the PR body are wrong about mechanism, and the widening direction is essentially unguarded.

## The denominator is not defensible as written

Independent re-derivation over the same 302 files, driving the extractor + `ResolveImports` + `ReferencesEmbedded` and applying the guard by hand:

```
EXTENDS     total=289  hexbind=154  arm=133  guard=0  residual=2   -> 287/289 (99.3%)
IMPLEMENTS  total=209  hexbind=47   arm=146  guard=0  residual=16  -> 193/209 (92.3%)
HIERARCHY   total=498  post=480 (96.4%)
```

The PR says IMPLEMENTS **205** and hierarchy **494**. The issue's own measurement says **209 / 498** (72+137+289). Four edges are unaccounted for, and 477/498 is **95.8%**, not 96.6%.

Quoting 96.6% means quoting the smaller of two denominators, in our own favour, with the drift undocumented — in a release note. Either the four edges get explained or we quote 498.

Also worth stating plainly in the body: **"resolved" here means "got a placeholder", not "correctly resolved."** `resolve.IsResolvedToID` (`internal/resolve/toid_shape.go:41`) is a pure shape check — 16-hex or an `ext:` prefix. The metric is monotonic in allowlist width, so it can be improved by widening the allowlist without resolving anything better.

## The anti-masking evidence does not exist

**Gate 3 fires zero times on the entire 302-file corpus** (`guardBlocked=0`).

The PR says the 16 `IFrameServer.*` residuals are "rejected by an in-tree mask guard." They are not. `VBNetExternalHierarchyTarget` returns `ok=false` at `vbnet_hierarchy_target_6337.go:50-58` because `IFrameServer` is not in `vbExternalBaseTypes` — gate 2. Gate 3 at `synth_vbnet_hierarchy_6337.go:71` is never reached. The unit-test comment in the branch already admits this while the file header comment at `:41-47` asserts the opposite.

**So the "residual is the right 17" story provides zero corpus evidence that the anti-masking guard works.** That story was the PR's entire answer to the central risk of external synthesis. Gate 3's only evidence is two synthetic unit cases.

Same shape at `refs.go:5843`, which claims "the nameExists guard at the call site covers" the `Windows` residual — untrue for any dotted spelling, because the vbnet extractor stamps namespace as a *property* (`vbnet_namespace`), not part of `Entity.Name` (`extractor.go:155-165`), so `inTreeNames` never contains `Windows.Foo`.

## Fabrication cases, all executed against `Synthesize` on this branch

| input (vbnet, EXTENDS) | result |
|---|---|
| `System.Foo.Bar.Baz.Nonexistent` | `ext:dotnet:System.Foo.Bar.Baz.Nonexistent` |
| `System.` (misparsed clause) | `ext:dotnet:System.` |
| `List(Of Machine)` | `ext:dotnet:List(Of Machine)` |
| `Windows.NativeMethods` + in-tree `NativeMethods` | `ext:dotnet:Windows.NativeMethods` |
| `Panel` + in-tree `panel` | `ext:dotnet:Panel` |

The dotted half is **not a curated allowlist** — `refs.go:5824` is `root ∈ {System, Microsoft, Windows}`. Any string with that prefix, including a truncated misparse, becomes a tidy resolved node and leaves the bug-edge count. `List(Of Machine)` gets its own node because `isVBExternalBaseType` strips at `(` to *test* while `VBNetExternalHierarchyTarget` returns the unstripped name — defeating the grouping this PR is sold on. And gate 3 is a plain map lookup, so it is case-sensitive while VB.NET is not: `Class panel` + `Inherits Panel` masks the ambiguity.

Credit where it is due: the `BooleanConverter` collision I asked to be probed **held** — it is a real BCL type in one corpus repo and a repo-local `Public Class BooleanConverter(Of T)` in another, and gate 3 would block it. But the PR tells the next contributor to "add it to the allowlist", and that is only safe because of a guard the PR misdescribes.

## `Exception` is on the allowlist and the allowlist never reaches it

Probe over 37 corpus targets, 36 changed, 1 did not:

```
"Exception"  OLD="ext:Exception"  NEW="ext:Exception"
```

`Exception` is in `vbExternalBaseTypes` (`refs.go:5900`) and is a live corpus base type, but the **language-agnostic** `stdlibBareNames` arm (`synth.go:3460`, entry at `:4136`, Python stdlib exceptions section) intercepts it first and returns subtype `"function"`. So VB's `Inherits Exception` lands on an untagged `ext:Exception` node shared with Python/Java/JS and typed as a function — exactly the cross-ecosystem canonical-name collapse this PR claims to have eliminated, on the most collision-prone name in the list, with no test covering it.

## Four of five widening mutants survived

M1–M8 are all *narrowing*. The widening direction is unguarded.

| # | Mutant | Result |
|---|---|---|
| W1 | dotted branch → always `true` | DIED |
| W4 | `vbFrameworkRootNamespaces` += `"My"`, `"Forms"` | **SURVIVED** |
| W2 | `buildInTreeNameSet` also excludes `SourceFile == ""` (weakens gate 3) | **SURVIVED** |
| W3 | relKind gate widened to `REFERENCES`/`USES` | **SURVIVED** |
| W6 | gate 3 skipped whenever `typePart` is dotted | **SURVIVED** |

**W4 is the finding.** `vbExternalBaseTypes` has a both-directions pin; `vbFrameworkRootNamespaces` has none — and it gates the higher-volume dotted half. `My` is VB's *compiler-generated in-tree* namespace (`My.MyProject.MySettings`), so adding it silently converts generated in-tree code into external placeholders with the whole suite green. The PR's own note that `Accessibility` and `Mono` were "unmeasured guesses, removed" shows the risk is live and that the removal was manual rather than test-enforced.

## Confirmed sound, not re-litigating

- **The lang gate is airtight** — a literal `lang != "vbnet"` early return at `synth_vbnet_hierarchy_6337.go:52`. C#/F# cannot reach it.
- **The position claim holds.** Nothing above `:1686` handles vbnet hierarchy except the global `stdlibBareNames` arm; below it the #4704 arm is `csharp && IMPORTS` only. What the dotted-root fallback used to produce was measured — `ext:System` for every dotted name, plus `ext:Windows`, `ext:Microsoft`, `ext:Form`, `ext:List` — and **36/37 targets strictly improve, with no regression surface found.**
- One cosmetic note: `internal/dashboard`'s `packageRoot("dotnet:System.Windows.Forms.Form")` returns `"dotnet:System"` because it splits on `.` before `:`. Only matters if an ext-package view ever renders hierarchy targets.

## Round 2 dispatched

Generics stripping, malformed-spelling rejection, the `Exception` decision, and the two false comments. The denominator question goes back for measurement before I touch the body — I have already been wrong about this corpus's baseline once this cycle and will not guess at it twice.

PR #6492, opened 2026-08-21
fix(#6459): a proto service sharing a name with a sibling claimed its ref and then dangled — fixed by an ordered tier, not a family widening

### LINKED ISSUE #6459: proto: SCOPE.Service is absent from operationKindFamily, so a service sharing a name with any sibling dangles — the mirror image of #6422
`internal/resolve/refs.go:1929-1932`:

```go
operationKindFamily = {"Operation", "Function", "Method", "SCOPE.Operation"}
```

**`SCOPE.Service` is not a member.** So a proto `service` addressed through the operation address space never matches on kind, and reaches its entity only by falling through to the kind-agnostic `byLocation` path — which fails the moment any other declaration shares its `(file, name)`.

Measured end to end on `message Foo { string id = 1; } service Foo { rpc Go(Foo) returns (Foo); }` in one file, after #6422's fix:

```
file -> message Foo  ->  SCOPE.Schema/message:Foo                              (resolves)
file -> service Foo  ->  UNRESOLVED:scope:operation:method:proto:b.proto:Foo   (dangles)
```

Before #6422 both dangled, so that change is a strict improvement — but it fixed the schema arm only. **This is the mirror image of #6422 on the service side**, and it was found while reviewing that fix rather than by a gate.

## Why it stayed hidden

`buildService`'s doc comment asserted that *"SCOPE.Service and SCOPE.Operation both resolve through the operation address space."* That is false, and it is exactly the kind of plausible-sounding claim that stops anyone checking. The comment has been corrected on the #6422 branch to state the truth, including that the service arm is unfixed — so the code no longer lies about it, but the defect is still there.

## Scope

Admit `SCOPE.Service` to `operationKindFamily`, or give services their own kind family. This is an `internal/resolve/` change, which is why it was deliberately kept out of #6422 — that PR was an extractor-side addressing fix, and widening a resolver kind family affects every language that emits a service-shaped entity, not just proto.

**Measure the blast radius before changing it.** `operationKindFamily` is consulted by the CALLS hint routing (`refs.go:2027-2032`) and by the package-scoped fallback (`refs.go:3480-3540`). Adding a kind to it changes what those match, so the question is not only "do proto services bind" but "does anything now bind that should not". Report before/after resolved counts, not just the proto case.

## Gradeability

Currently ungradeable: no golden fixture covers proto at all (#6453 — 22 fixture dirs, none proto, zero mentions in `baseline.json`). The `proto-mini` fixture proposed there should carry a `service`/`message` name-collision case for exactly this reason, alongside the rpc/message collision that #6422 needed. One fixture can pin both arms.

Note there is an inert `testdata/fixtures/protobuf/proto__sample.json` — stale absolute path from another machine, no Go test loads it, referenced only by a comment at `proto_test.go:168`. It is not a fixture in any useful sense; do not mistake it for coverage.

cajasmota:
Closes #6459

A proto service sharing a name with a sibling claimed its structural ref and then dangled. Fixed by an **ordered tier**, not by widening any kind family.

## Two earlier attempts widened a kind family, and both were corpus-wide regressions

Recorded because the lesson is the substance of this PR.

`lookupLocationKind` and `uniqueMatchInFamily` return `("", false)` when two distinct IDs in a family match. **Adding a member can therefore turn unique → ambiguous and destroy a binding that previously worked.**

- **Revision 1** added `SCOPE.Service` to the shared `operationKindFamily`. `SCOPE.Service` is emitted at ~60 sites, and `internal/custom/python/celery.go:94` names the task entity after the function in the same file. Result: `lookupStructural("scope:operation:method:python:app/tasks.py:send_email")` went from the function to **`handled=true, id=""`** — verbatim the failure mode #6459 exists to remove, imported into Python. Measured on `python-dramatiq-mini`: `rewrote=11 ambiguous=0` → `rewrote=9 ambiguous=2`, two actor CALLS edges lost.
- **Revision 2** gated that widening to proto only. It reimported the same failure one address space down: `buildService` addresses each rpc with `BuildOperationStructuralRef("proto", file, rpcName)`, so an rpc named `User` and a service named `User` in one file both entered the operation family and a **correct** `service Admin → rpc User` CONTAINS edge was destroyed — one edge lost, zero gained.

Both revisions' load-bearing safety arguments were falsified by execution. Revision 1's — *"`lookupByKindHint` requires a unique family match, so this cannot steal a target from a real function"* — argued only one direction of a two-directional property: uniqueness rules out *stealing* and **guarantees loss**.

## What landed: an ordered tier

`lookupProtoServiceTier` runs in `lookupStructural`'s Format-A path **after** `lookupLocationKind` misses and **before** the `ambigLocation`/`byLocation` fallbacks. It scans the (file, name) bucket for the presence of any `operationKindFamily` kind — blank sentinel counted as present — and **bails if it finds one**; otherwise it matches over `protoServiceKindFamily = {SCOPE.Service}`.

So the rpc/service collision never reaches it: an rpc named `User` is present, the tier bails, and the correct edge survives. Only the `message Foo` + `service Foo` shape — where the operation family finds *nothing* — gets the new resolution.

**No kind family is widened.** `operationKindFamily`, `componentKindFamily`, `schemaKindFamily`, `componentOrOperationKindFamily`, `hintKinds` and `memberFamilyMask` are all byte-identical to `main` (verified by comment-stripped sha256).

A separate `scope:service:…` address space was evaluated and rejected: it changes the ToID string of every proto `file → service` edge, which is also read by `internal/extractors/cross/endpoint`, so it would have to be taught to two emitters and the stub grammar at once — and it fixes nothing the tier doesn't, since the rpc/service collision is resolved by *ordering* either way.

## The precondition's family is load-bearing, and reachable

A coordinator mutant substituting `componentOrOperationKindFamily` for `operationKindFamily` in the precondition survived round 4. It looked latent. It is not.

`buildImportEntities` (`internal/extractors/proto/proto.go:1002-1006`) sets `Name` to the **verbatim quoted import string** — the grammar accepts any string and grafel never runs protoc — so `import "Foo";` mints `SCOPE.Component{Name:"Foo"}` in the same file as `service Foo`:

| build | `file → service Foo` CONTAINS |
|---|---|
| fixed | **1** |
| mutant | **0** |

Now pinned by `TestProtoServiceTierPreconditionScansTheWholeFamilyAndBase6492` (`scope-component`, `bare-component`) and the end-to-end `TestSelfNamedImportDoesNotBlockTheServiceTier6492`. The underlying property — an unvalidated string literal used as an entity `Name` — is filed separately as **#6497**, since it can collide a `SCOPE.Component` with any proto entity kind, not just services.

## What #6459 does NOT close, stated in the docs and pinned

`service Foo { rpc Foo(Bar) returns (Bar); }` leaves the service with **zero inbound CONTAINS** and the rpc with **two** — the `file → service` edge mis-binds onto the rpc. Both refs are the byte-identical string `scope:operation:method:proto:<file>:Foo`, so no tier can break the tie; closing it is a ref-*format* change belonging to its own issue.

`proto.go` and `NOTICE.md` previously claimed #6459 was closed outright. Both are now scoped to the `message Foo` + `service Foo` shape, and the residual is pinned by `TestSelfNamedRpcLeavesTheServiceOrphaned6459Residual`.

A separate unobserved claim — *"two rpcs in one service sharing a name behave the same way"* — was **deleted rather than asserted**: protoc rejects duplicate method names within a service, so a test would have pinned behaviour on input the language forbids. It was also simply wrong — that shape dangles, it does not mis-bind.

## Mutation coverage

Across five rounds, every serious defect was found by a mutant that made a predicate **more permissive**, and the suite was initially blind to that direction.

Killed: `SCOPE.Service` into `operationKindFamily` (revision 1's regression) · bare `"Service"` alongside the scoped form · `SCOPE.Controller` into the family · `SCOPE.Service` into `componentOrOperationKindFamily` · the round-2 proto-only widening · language gate removed / gated on the wrong language / widened to 8 more languages · precondition scanning `.real` only · precondition scanning one member only · `EqualFold` → `==` · the tier neutered · `SCOPE.Schema`, `Schema`, `Service`, `SCOPE.Service`, `Component` added to the precondition · the 27-row language table deleted (row-count floor plus a 15-entry required set).

The final round's independent review reports **no surviving mutants** — the first empty survivor table on this PR.

Two dead arms were removed rather than left in place: the precondition's `.real` scan (per entity `keys(.real) ⊆ keys(.base)`, so it could never fire alone — verified against both writers of `byLocationKind`), and `protoOperationKindFamily` itself.

`TestMemberFamilyMask_FamiliesAreDisjoint_6141` was also fixed: it iterated the **raw** `familyMaskByKind` while `memberFamilyMask` ORs the `SCOPE.`-trimmed form, so bare `"Service"` in one family and `SCOPE.Service` in another passed as two single-bit entries. It now iterates `memberFamilyMask` over the family union, with a non-vacuity guard.

## Verification

`gofmt -l` empty · `go vet ./internal/... ./cmd/...` → 0 · `go test -count=1 ./internal/resolve/... ./internal/extractors/proto/ ./internal/extractors/sresolver/ ./internal/quality/...` → 0, 7 packages ok. `-count=1` throughout: a cached result scored a mutant against stale output earlier in this work. `refs.go` restored by `cp` with shasum verified after every mutant; `git checkout --` never used.

### DIFF
--- a/internal/extractors/proto/proto.go
+++ b/internal/extractors/proto/proto.go
@@ -270,8 +270,9 @@ func fileContainsRel(filePath, toRef string) types.RelationshipRecord {
 // path — which drops any (file, name) that is not unique. The second revision
 // recorded that gap accurately and left it open as #6459.
 //
-// #6459 has since closed it — as an ORDERED TIER, not by widening any kind
-// family. SCOPE.Service is in NO family: not the shared operationKindFamily
+// #6459 has since closed it FOR THE `message Foo` + `service Foo` SHAPE — as an
+// ORDERED TIER, not by widening any kind family. That scope is the whole claim;
+// the residual it does not cover is stated at the bottom of this comment. SCOPE.Service is in NO family: not the shared operationKindFamily
 // (that slice also feeds hintKinds and the leaf-name family mask, and ~60
 // non-proto sites emit SCOPE.Service beside a same-named function or class, so
 // a global admission destroys those unique matches instead of adding any), and
@@ -302,12 +303,33 @@ func fileContainsRel(filePath, toRef string) types.RelationshipRecord {
 // by luck, so the pre-fix damage is only observable once the ambiguity exists —
 // which is why the guard fixture carries the colliding message.)
 //
+// WHAT IS STILL NOT RESOLVED, stated as plainly as the fix. When a service and
+// one of its OWN rpcs share a name —
+//
+//	service Foo { rpc Foo(Bar) returns (Bar); }
+//
+// — this function and buildService produce the BYTE-IDENTICAL ref
+// scope:operation:method:proto:<file>:Foo for two different entities. The tier
+// cannot help and must not try: its precondition sees the rpc in the operation
+// family and bails, because the alternative is a service outranking a real rpc.
+// Measured end-to-end through this extractor at the head that added the tier:
+// the service ends with ZERO inbound CONTAINS and the rpc carries TWO — its own
+// parent edge plus the file → service edge mis-bound onto it. That is #6459's
+// title symptom surviving in this one shape. It is a MIS-BINDING, not a dangle,
+// so nothing is left unresolved and no ref-integrity check sees it. Closing it
+// needs this file to stop addressing a service and its rpc identically — a ref
+// FORMAT change, not a resolver change — and belongs in its own issue with its
+// own measurement. The same applies to two rpcs in one service sharing a name.
+//
 // The guards are internal/resolve/proto_service_family_6459_test.go (the
 // service half) and internal/resolve/proto_rpc_service_collision_6492_test.go
 // (the rpc half — the two-service fixture above). The counter-guard, that a
 // non-proto operation-space ref (celery task, Spring stereotype) is unaffected
 // and that the language boundary admits proto and nothing else, is
-// internal/resolve/service_family_scope_6492_test.go.
+// internal/resolve/service_family_scope_6492_test.go. The residual above is
+// pinned as a characterisation test:
+// TestSelfNamedRpcLeavesTheServiceOrphaned6459Residual in
+// service_ref_e2e_6492_test.go.
 func fileContainsOperationRel(filePath, name string) types.RelationshipRecord {
 	return fileContainsRel(filePath, extractor.BuildOperationStructuralRef("proto", filePath, name))
 }
--- a/internal/extractors/proto/proto.go
+++ b/internal/extractors/proto/proto.go
@@ -270,9 +270,15 @@ func fileContainsRel(filePath, toRef string) types.RelationshipRecord {
 // path — which drops any (file, name) that is not unique. The second revision
 // recorded that gap accurately and left it open as #6459.
 //
-// #6459 has since closed it: internal/resolve/refs.go's operationKindFamily now
-// includes scopeKindPrefix+"Service", so BOTH arms bind through the kind-filtered
-// lookupLocationKind tier. Measured on
+// #6459 has since closed it, and #6492 re-scoped the closure: SCOPE.Service is
+// NOT in internal/resolve/refs.go's shared operationKindFamily (that slice also
+// feeds hintKinds and the leaf-name family mask, and ~60 non-proto sites emit
+// SCOPE.Service beside a same-named function or class, so a global admission
+// destroys those unique matches instead of adding any). It is admitted by
+// protoOperationKindFamily, reached only from
+// structuralKindFamilies("operation", "proto") — i.e. from the LANGUAGE segment
+// this very function stamps into the ref. Both arms therefore bind through the
+// kind-filtered lookupLocationKind tier for proto, and only for proto. Measured on
 //
 //	message Foo {…}; service Foo { rpc Go(Foo) returns (Foo); }
 //
@@ -281,7 +287,9 @@ func fileContainsRel(filePath, toRef string) types.RelationshipRecord {
 // before it dangled and the service ended with zero inbound CONTAINS. The guard
 // is internal/resolve/proto_service_family_6459_test.go; it also pins that
 // SCOPE.Service stays OUT of componentKindFamily, since nothing addresses a
-// service in the component address space.
+// service in the component address space. The counter-guard — that a non-proto
+// operation-space ref (celery task, Spring stereotype) is unaffected — is
+// internal/resolve/service_family_scope_6492_test.go.
 func fileContainsOperationRel(filePath, name string) types.RelationshipRecord {
 	return fileContainsRel(filePath, extractor.BuildOperationStructuralRef("proto", filePath, name))
 }
--- a/internal/extractors/proto/proto.go
+++ b/internal/extractors/proto/proto.go
@@ -260,29 +260,28 @@ func fileContainsRel(filePath, toRef string) types.RelationshipRecord {
 // fileContainsOperationRel is the file → service form, and the form the
 // service → rpc edges already use.
 //
-// A PRECISE STATEMENT OF WHAT THIS DOES AND DOES NOT RESOLVE, because an
-// earlier revision of this comment claimed "SCOPE.Service and SCOPE.Operation
-// both resolve through the operation address space" and that is FALSE.
-// internal/resolve/refs.go:1929-1932 defines
+// A PRECISE STATEMENT OF WHAT THIS RESOLVES, because two earlier revisions of
+// this comment were wrong in opposite directions.
 //
-//	operationKindFamily = {"Operation", "Function", "Method", "SCOPE.Operation"}
+// The first claimed "SCOPE.Service and SCOPE.Operation both resolve through
+// the operation address space", which was FALSE at the time: operationKindFamily
+// held only {"Operation", "Function", "Method", "SCOPE.Operation"}, so a service
+// reached its entity solely by falling through to the kind-agnostic byLocation
+// path — which drops any (file, name) that is not unique. The second revision
+// recorded that gap accurately and left it open as #6459.
 //
-// and SCOPE.Service is NOT in it. An rpc (SCOPE.Operation) binds through the
-// family; a service (SCOPE.Service) does not, and reaches its entity only by
-// falling through to the kind-agnostic byLocation path — which fails the
-// moment any other entity shares its (file, name).
-//
-// So the SERVICE arm still has the #6422 shape in mirror image. Measured on
+// #6459 has since closed it: internal/resolve/refs.go's operationKindFamily now
+// includes scopeKindPrefix+"Service", so BOTH arms bind through the kind-filtered
+// lookupLocationKind tier. Measured on
 //
 //	message Foo {…}; service Foo { rpc Go(Foo) returns (Foo); }
 //
-// in one file: the file → service Foo CONTAINS does not resolve, and the
-// service ends with zero inbound CONTAINS, while the message now correctly
-// has one. That is a strict improvement over the pre-#6422 state, where BOTH
-// dangled — but it is not a fix, and this comment does not claim to be one.
-// Closing it needs SCOPE.Service admitted to operationKindFamily (a
-// cross-language change to the resolver, not to this file) and belongs in its
-// own issue with its own measurement.
+// in one file — the collision that made the old byLocation fallback fail — the
+// file → service Foo CONTAINS now resolves to the SCOPE.Service entity, where
+// before it dangled and the service ended with zero inbound CONTAINS. The guard
+// is internal/resolve/proto_service_family_6459_test.go; it also pins that
+// SCOPE.Service stays OUT of componentKindFamily, since nothing addresses a
+// service in the component address space.
 func fileContainsOperationRel(filePath, name string) types.RelationshipRecord {
 	return fileContainsRel(filePath, extractor.BuildOperationStructuralRef("proto", filePath, name))
 }
--- a/internal/extractors/proto/proto.go
+++ b/internal/extractors/proto/proto.go
@@ -270,25 +270,43 @@ func fileContainsRel(filePath, toRef string) types.RelationshipRecord {
 // path — which drops any (file, name) that is not unique. The second revision
 // recorded that gap accurately and left it open as #6459.
 //
-// #6459 has since closed it, and #6492 re-scoped the closure: SCOPE.Service is
-// NOT in internal/resolve/refs.go's shared operationKindFamily (that slice also
-// feeds hintKinds and the leaf-name family mask, and ~60 non-proto sites emit
-// SCOPE.Service beside a same-named function or class, so a global admission
-// destroys those unique matches instead of adding any). It is admitted by
-// protoOperationKindFamily, reached only from
-// structuralKindFamilies("operation", "proto") — i.e. from the LANGUAGE segment
-// this very function stamps into the ref. Both arms therefore bind through the
-// kind-filtered lookupLocationKind tier for proto, and only for proto. Measured on
+// #6459 has since closed it — as an ORDERED TIER, not by widening any kind
+// family. SCOPE.Service is in NO family: not the shared operationKindFamily
+// (that slice also feeds hintKinds and the leaf-name family mask, and ~60
+// non-proto sites emit SCOPE.Service beside a same-named function or class, so
+// a global admission destroys those unique matches instead of adding any), and
+// not a proto-only variant of it either. A proto-only widening fails for a
+// reason specific to THIS file: buildService addresses every rpc child through
+// the same BuildOperationStructuralRef this function uses, so rpcs and services
+// occupy ONE address space. Put SCOPE.Service into the family that space is
+// filtered by, and
+//
+//	service User  { rpc Get(Foo)  returns (Foo); }
+//	service Admin { rpc User(Foo) returns (Foo); }
+//
+// — ordinary proto — makes the rpc User ref match two family members, and the
+// service Admin → rpc User CONTAINS edge that used to resolve now dangles.
+//
+// internal/resolve/refs.go's lookupProtoServiceTier instead consults
+// SCOPE.Service only when the unmodified operation family matched NOTHING at
+// all at that (file, name), and only for the proto LANGUAGE segment this very
+// function stamps into the ref. On
 //
 //	message Foo {…}; service Foo { rpc Go(Foo) returns (Foo); }
 //
-// in one file — the collision that made the old byLocation fallback fail — the
-// file → service Foo CONTAINS now resolves to the SCOPE.Service entity, where
-// before it dangled and the service ended with zero inbound CONTAINS. The guard
-// is internal/resolve/proto_service_family_6459_test.go; it also pins that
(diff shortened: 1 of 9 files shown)


cajasmota:
## Independent review: REQUEST-CHANGES

The one-line change is a **corpus-wide resolution regression**, not a proto-local fix — and the PR's load-bearing argument is falsified by execution. Round 2 is dispatched; this records why.

### The site enumeration is complete but audited the wrong axis

Five consult sites, all found. But the review only ever asks *who consumes* `operationKindFamily`. It never asks **who produces `SCOPE.Service`** — and that is where the damage is.

`SCOPE.Service` is emitted by roughly 60 sites across `internal/patterns/`, `internal/custom/{python,golang,rust,elixir}/` and `internal/extractors/{kotlin,yaml,proto}/`. It is not a proto concept. Several of those sites name the entity after a function or class **that already exists in the same file**.

### F1 — confirmed by execution: celery loses its binding

`internal/custom/python/celery.go:94` emits `entity(funcName, "SCOPE.Service", "task", …)` for every `@shared_task def <fn>` — same file, same name as the `SCOPE.Operation` function. `internal/extractors/python/celery.go:13-16` documents the duplication; `:231` mints `BuildOperationStructuralRef("python", file.Path, taskName)` for every `.delay()`/`.apply_async()` CALLS edge. `dramatiq.go:86` has the same shape.

Head vs. head-minus-the-new-line:

| probe | pre-fix | post-fix |
|---|---|---|
| `lookupStructural("scope:operation:method:python:app/tasks.py:send_email")` | `f1f1…` (the function) | **`handled=true, id=""`** |
| `lookupByKindHint("send_email","CALLS")` | `f1f1…`, ok=true | **`""`, ok=false** |

`handled=true, id=""` is *verbatim* the failure mode this PR exists to remove. Adding the member imports the defect into Python.

### F2 — the safety argument proves the opposite of what it claims

The PR says `lookupByKindHint` "requires a unique family match, so this cannot steal a target from a real function."

`uniqueMatchInFamily` (`refs.go:2440`) and `lookupLocationKind` (`:2997-3008`) return `("", false)` when two distinct IDs in the family match. Uniqueness rules out *stealing* — and **guarantees loss**. Only one direction of a two-directional property was argued. Widening a family can turn unique → ambiguous and destroy a binding that previously worked.

Damage is confined to `SCOPE.*`-only shapes, because the tier-1 `.real` bucket skips them. That is Python, Kotlin, proto, and every custom-extractor language. Most of the corpus.

### F3 — Kotlin, unmentioned

`internal/extractors/kotlin/kotlin.go:815` emits `Kind: "SCOPE.Service", Name: className` for Spring stereotypes. A `scope:operation:*:<file>:OrderService` ref moves from unmatched to binding the **stereotype marker entity**. Whether that is desirable is arguable; that it is undiscussed is not.

### F4 — the disjointness invariant does not cover what the PR credits it with

`TestMemberFamilyMask_FamiliesAreDisjoint_6141` iterates the **raw** `familyMaskByKind` map. But `memberFamilyMask` (`:2287-2294`) ORs a kind's mask with its `SCOPE.`-trimmed form's mask. Bare `"Service"` in one family and `SCOPE.Service` in another produce two single-bit map entries — the test passes — while `memberFamilyMask("SCOPE.Service")` is a two-bit mask. It is not a general net against this change.

### F5 / F6 — the stale-comment class, reintroduced

- `internal/extractors/sresolver/pkgmember.go:180-191` calls itself "the scoped port of `operationKindFamily`" and names a pin that "walks the same four kinds". Not adding `Service` there is behaviourally right; the comment is now false.
- `proto-mini/NOTICE.md` contradicts itself — the new paragraph says the fix landed, the paragraph below still asserts in present tense that `operationKindFamily` omits `SCOPE.Service`.

Both are the same defect class this PR fixes elsewhere.

### Surviving mutants — all three widening

- **N1** bare `"Service"` alongside the scoped form → survived (membership is exact-string).
- **N3** `SCOPE.Controller` added to the family → survived.
- **N8** bare `"Service"` in `componentOrOperationKindFamily` → survived.

Nothing in the suite guards against widening the family generally. This is the fifth PR this run where the *narrowing* mutants all died and the *widening* mutants all lived.

### Path to landable

1. Restrict the fix to the address space that needs it — the `structuralKindFamilies("operation")` path (`refs.go:2990`) — rather than the shared slice that also feeds `hintKinds` and `familyMaskByKind`; or gate on `Language == "protobuf"` / `Subtype == "service"`.
2. RED test for the celery collision, confirmed failing on current head.
3. Grade a fixture that can actually observe this. `proto-mini` cannot — its own NOTICE says so.
4. Fix F4 to iterate `memberFamilyMask` over the family union.
5. Fix F5 and F6.

cajasmota:
## Independent review of round 2 (`88d627c2e`): REQUEST-CHANGES

Round 2 correctly removes the corpus-wide regression, and most of its claims reproduce under execution. But **it reintroduces the same unique→ambiguous failure mode inside the proto address space**, reproduced end-to-end. Round 3 is dispatched to replace the mechanism.

### B1 (blocking) — the widening destroys a correct `service → rpc` CONTAINS

The safety comment argues that "proto, and only proto, points an operation-space ref at a `SCOPE.Service`". That is true and irrelevant. The collision that matters is *within* a `.proto`, between a service and an **rpc** — both now in the operation family.

`buildService` (`internal/extractors/proto/proto.go:372`) addresses each rpc child with `BuildOperationStructuralRef("proto", file, rpcName)`. When an rpc's name equals any service name in the same file, the ref matches two family members and `uniqueMatchInFamily` returns `("", false)`.

```proto
service User  { rpc Get(Foo)  returns (Foo); }
service Admin { rpc User(Foo) returns (Foo); }
```

| edge | pre-fix | at `88d627c2e` |
|---|---|---|
| `service Admin → rpc User` CONTAINS | `6544153a → c55020864eaf05ae` **resolved** | **DANGLING** |
| `file → service User` CONTAINS | `→ c55020864eaf05ae` (mis-bound to the rpc) | **DANGLING** |

**One correct edge destroyed, zero gained** — and `SCOPE.Service User` ends with no inbound CONTAINS, which is the literal symptom in this issue's title, now caused by the fix. `service User { rpc User(...) }` reproduces identically.

Measured by diffing the resolved graph through the in-process `Index(...)` at HEAD versus HEAD with only the language gate neutered; `go vet` clean both ways, `refs.go` restored by `cp` with shasum verified.

This is review F2 — *"uniqueness rules out stealing and guarantees loss"* — re-imported one address space down. Neither the PR nor any test covers it, and `proto-mini` cannot see it: no rpc there shares a service name.

**Widening a family cannot fix this**, because the ref is genuinely ambiguous between two legitimate proto entities. Round 3 replaces it with an ordered tier: try `operationKindFamily` first and use it if unique (so the rpc wins, unchanged); fall back to a service-kinded lookup only when the first tier yields *nothing at all* (which is where `message Foo` + `service Foo` gets its resolution).

### B2 (blocking) — the language gate is unpinned; a widening mutant survives

`case "proto", "protobuf":` → `case "proto", "protobuf", "go", "java", "typescript":` — target count 1, `go vet` 0:

```
go test ./internal/resolve/... ./internal/extractors/proto/  → exit 0   SURVIVED
```

The gate is the entire substance of round 2 over round 1, and it is pinned only at `python` and `kotlin`. `TestProtoOperationFamilyIsTheOnlyWidening6492` compares proto against one control language. Any *other* language admitted is invisible.

| # | mutation | dir | result |
|---|---|---|---|
| W1 | add `"go","java","typescript"` to the gate | widening | **SURVIVED** |
| W2 | gate removed, always proto family | widening | dead |
| M1 | gate removed, always `operationKindFamily` | narrowing | dead |
| W3 | extra member in `protoOperationKindFamily` | widening | dead |
| W4 | scoped → bare `"Service"` | — | dead |

### What holds

Byte-identity of all six untouched families and `hintKinds`/`memberFamilyMask` — comment-stripped sha256 identical to main. Language-segment consistency — every proto emitter stamps the literal `"proto"` (`proto.go:294,372`), single call site at `refs.go:2733`, bounds-guarded at `:2686`; the `"protobuf"` arm is dead but harmless. Reachability — on `message Foo` + `service Foo` the `file → service` CONTAINS dangles pre-fix and resolves post-fix through `Index()`, so this is not dead code. `componentOrOperationKindFamily` untouched. The dramatiq numbers reproduce exactly. F4 and F5 fixed.

### Non-blocking

- **N2** — `NOTICE.md` credits `python-dramatiq-mini` as the fixture that observes the difference. **It cannot fail**: `baseline.json` records `relationship_expected: 0` for it, and `TestJobFixturesAbsoluteRecall_6260/python-dramatiq-mini` passes under the round-1 regression. Only an unasserted stderr line moves. This is the same "cited a fixture that cannot observe it" class the PR fixes elsewhere — credit `TestCeleryTaskCallStillBindsToTheFunction6492` instead.
- **N3** — `proto.go:284-287` claims the service "ended with zero inbound CONTAINS" pre-fix. End-to-end it retained a resolved `file → service` CONTAINS. True of the synthetic unit index only.
- **N5** — `TestKotlinStereotypeMarkerIsNotAnOperationTarget6492` asserts only `id != marker.ID` and would pass on an empty index; the F4 non-vacuity guard (`len(seen) < 4`) is similarly loose.

N1 in the review (stale PR body) was already addressed — the body was rewritten for round 2 after the review started, and will be rewritten again for round 3.

The revert, the byte-identity, the celery and Kotlin guards, and the F4 fix are solid and should survive whatever replaces the family widening.

cajasmota:
## Coordinator mutant on round 4 — SURVIVED

Recording this before the review returns, because it bears on a consistency question rather than just on this PR.

**Mutation:** in `lookupProtoServiceTier`'s precondition, `for _, k := range operationKindFamily` → `for _, k := range componentOrOperationKindFamily`.

Target count asserted 1 · `go vet ./internal/resolve/` exit 0 **before** scoring · `GOMAXPROCS=4 go test ./internal/resolve/ ./internal/extractors/proto/` → **exit 0, SURVIVED** · restored by `cp`, shasum `f123d45c…` verified identical, tree clean.

### Assessment

**Non-equivalent in principle, unreachable in proto today.** `componentOrOperationKindFamily` adds `Component`/`Class`/`View`/`Model` and their `SCOPE.` forms. Proto's only `SCOPE.Component` entities come from `buildImportEntities` and are named after the import *target path*, so they cannot share a (file, name) bucket with a service name.

**Severity is lower than S2/S3.** The failure direction is the tier bailing *more* often — #6459 silently regressing — not a `SCOPE.Service` outranking a real rpc.

### Why it is worth raising anyway

The round-3 review explicitly refused *"unreachable today because `proto.go` emits exactly four kinds"* as a defence for S2 and S3, on the grounds that it is an accident of the current extractor and not an invariant — and round 4 pinned them. **This survivor rests on the identical argument.** Accepting it here while rejecting it there would be convenient rather than consistent.

The pin looks close to free: `TestProtoServiceTierPreconditionScansTheWholeFamilyAndBase6492` already builds synthetic buckets, so a `SCOPE.Component` row at the service's (file, name) should kill it.

Handed to the round-4 reviewer to confirm or refute rather than acted on unilaterally — in particular whether any non-proto indexer can write a `Component`/`Class`/`View`/`Model`-kinded entity under a `.proto` path, which would make it live rather than latent.

cajasmota:
## Round 4 review: REQUEST-CHANGES — and my own equivalence argument was wrong

I claimed the surviving precondition mutant was "non-equivalent in principle, unreachable in proto today," reasoning that proto's `SCOPE.Component` entities are import stubs *named after the import target path* and therefore cannot share a (file, name) bucket with a service name.

**That is false.** `buildImportEntities` (`internal/extractors/proto/proto.go:1002-1006`) sets `Name: path`, where `path` is the **verbatim quoted string** from `parseImport`. It is not a path in any enforced sense — the tree-sitter grammar accepts any string, and grafel never runs protoc. So `import "Foo";` mints `SCOPE.Component{Name:"Foo", SourceFile:<the .proto>}` in the same file as `service Foo`.

Proven end-to-end through the production pipeline (real extractor → real `graph.EntityID` → real `BuildIndex` → real `ReferencesEmbedded`):

```proto
syntax = "proto3";
import "Foo";
service Foo { rpc Go(Bar) returns (Bar); }
message Bar { string id = 1; }
```

Both entities are emitted at `(zz.proto, "Foo")`:

```
ENT kind="SCOPE.Service"   subtype="service" name="Foo" id="ecdf5122f9331d51"
ENT kind="SCOPE.Component" subtype="import"  name="Foo" id="43f4b781529e2cef"
```

| build | `file → service Foo` CONTAINS |
|---|---|
| head `9b7e7516d` | **1** (tier fires, #6459 closed) |
| head + mutant | **0** (tier bails, edge dangles) |

#6459's exact title symptom, reintroduced, on valid proto — **no hypothetical extractor change required.** The minimal form (`append(operationKindFamily, scopeKindPrefix+"Component")`) also survives, localising the gap to exactly `SCOPE.Component`/`Component`.

My severity read was correct and is confirmed: the direction is the tier bailing *more*, so the failure is a lost binding, never a `SCOPE.Service` outranking a real rpc — strictly less severe than S2/S3. But "less severe" is not "unpinnable", and the reachability finding removes the only defence I had.

**Consistency settles it.** Round 4 accepted the standard that "unreachable today is an accident of the current extractor, not an invariant" — that phrasing is in `TestProtoServiceTierPreconditionScansTheWholeFamilyAndBase6492`'s own comment, about `SCOPE.Method`. This gap is weaker still: it needs no accident at all. Round 5 pins it.

### Everything else on round 4 verified

**S2, S3, S5, S6 are genuinely dead** — each re-mutated from a shasum-verified backup, `go vet` 0, target count 1, restored and re-verified. Non-vacuity confirmed by reading the in-body controls, not asserted: the precondition test carries an empty-operation-entity control and a positive control, and a blank-sentinel mutant dies on both, so the tier is demonstrably firing when these tests pass.

**The S1 deletion is correct.** Exactly two writers of `byLocationKind` exist (`refs.go:1277/1280`, `symbol_index.go:704/707`), both with identical shape, so `keys(.real) ⊆ keys(.base)` per entity. Since the deleted arm tested presence only, it could never fire without the `.base` loop having already returned. Provably dead code; no third writer.

**The language table is adequate** — my concern about the row-count floor was already answered: there is a `required` list of 15 entries alongside `len(cases) >= 20`, so twenty all-proto rows would fail on the first required negative.

**The rescoped claims are actually observed**, which breaks the rounds 1-3 pattern. `TestSelfNamedRpcLeavesTheServiceOrphaned6459Residual` asserts the exact numbers both docs state — service inbound CONTAINS `== 0`, rpc `== 2` — end-to-end, plus a no-dangling-`scope:`-ToID sweep proving the mis-binding characterisation.

**One remaining doc nit:** both `proto.go` and `NOTICE.md` assert *"Two rpcs in one service sharing a name behave the same way"* — no test observes that shape. Round 5 either asserts it or scopes the sentence.

Also confirmed dead: `SCOPE.Component` into `protoServiceKindFamily`, bare `"Service"`, `SCOPE.Service` into `componentOrOperationKindFamily` (round 1's surviving N8), and `SCOPE.Service` into `operationKindFamily` (round 1's regression). `./internal/quality/...` green.

Issue #6449, opened 2026-08-20
the runtime_dynamic repair queue has been severed since #1217 by a stale kind literal, and its test fixture hardcodes the dead kind so nothing noticed


cajasmota:
`internal/enrichment/dynamic_baseurl.go:83` reads:

```go
if e.Kind != "http_endpoint" { continue }
```

Consumer entities have been minted as **`http_endpoint_call`** since #1217 split the kind (`internal/engine/http_endpoint_synthesis.go:534-537`). So on any graph indexed after that split, `CollectDynamicBaseURLCandidates` returns **zero candidates**, every time.

That severs the entire `runtime_dynamic` → repair-queue feed built for #732 / ADR-0015. The repair machinery downstream (`internal/enrichment/repair_apply.go`, five allowlisted resolutions, R1–R7 validated twice) is complete and tested — it simply never receives anything to work on.

## The test cannot catch it

`dynamic_baseurl_test.go:29` hardcodes the **legacy** kind in its fixture. So the test constructs an entity of a kind the pipeline no longer produces, matches it, and passes. The function is green and dead at the same time.

This is the vacuous-fixture class we have been tracking, in its purest form: an assertion that cannot fail because its input no longer resembles production. Fixing the comparison without fixing the fixture leaves the mutant alive.

## Fix

`types.IsHTTPEndpointKind` already exists (`internal/types/kinds.go:463`) and is documented as the thing to use for exactly this. Replace the raw string comparison with it, and rebuild the test fixture from a kind the current synthesis path actually emits.

One line of production code. The value is not the line — it is that a shipped feature has been disconnected for an unknown number of releases and nothing reported it.

## Verification bar

A mutant reverting the comparison to the legacy string must **die**. If it survives, the fixture is still wrong.

Worth also checking, in the same pass: are there other `!= "http_endpoint"` or `== "http_endpoint"` raw string comparisons that predate the #1217 split? **Derive that set** rather than fixing only this one — a claim of totality that holds only for the subset searched has bitten this repo repeatedly, and this defect is exactly what a stale literal looks like after a kind rename.

Issue #1639, opened 2026-05-22
flows: filter false-positive short flows (2-3 steps) + continue HTTP-call into the backend handler node (fix cross-repo tracing + mislabel)


cajasmota:
Owner, two flow-quality issues: 

(1) **False-positive short flows:** a "flow" with only 2-3 steps is most likely not a meaningful process. Filter/down-rank trivial flows (e.g. require >=N steps, default ~4, configurable) so the Flows list shows real multi-step processes, not noise. Keep them queryable but out of the default list. 

(2) **HTTP call stops at the call-site, not the backend — breaks cross-repo tracing + causes mislabel:** when a flow step is an HTTP call it shows the `http_endpoint_call` node and STOPS; it should CONTINUE into the resolved backend `http_endpoint_definition` / handler node so the process tracks deeper across the HTTP boundary into the backend repo. Use the cross-repo HTTP resolution (the #1615/#1625 structural matcher) to link the call → the backend route → its handler, extending the flow chain. This ALSO fixes the cross-repo MISLABEL the owner saw — flows are tagged "cross-repo" but on inspection are single-repo (the call node is in the caller repo but the flow never reaches the backend repo). Once flows extend to the backend node, the cross-repo tag is accurate (only tag a flow cross-repo when it genuinely spans repos). 

Files: the process-flow builder (internal/engine flows / process) + the flows endpoint/MCP + maybe webui flows list filter. Verify on real polyglot + upvate (isolated): no 2-3-step noise in the default flows list; an HTTP-crossing flow continues into the backend handler and is correctly labeled cross-repo; genuinely single-repo flows are not mislabeled.

Issue #4319, opened 2026-06-09
Endpoint→handler bridge missing (endpoints are graph islands): fix NestJS + GENERALIZE across all HTTP frameworks


cajasmota:
**Severity: high (rewrite-critical — the v3 NestJS backend is the rewrite agent's primary target; right now every endpoint's Paths panel shows empty Parameters/Response/Downstream/Tests).**

**Root cause (confirmed on live `upvate-v3`, endpoint `POST /inspections/me-manage/merge`, `building.controller.ts:108`):** the NestJS extractor emits TWO separate nodes for the same controller method and never connects them:
- `http:POST:/inspections/me-manage/merge` (`http_endpoint_definition`, line 108) → **neighbors(both) = EMPTY. It's a graph island.**
- `mergeMaintenanceEvaluations` (`Operation`, SAME file+line 108) → has the real edges: `VALIDATES → BuildingMeMergeBody` (the @Body payload DTO) and `CALLS → building.service.ts:mergeMaintenanceEvaluations` (downstream service → repo → mongo).

So all the rich data IS extracted (payload DTO, response DTO `BuildingMeMergeResponse`, contract test `ME_MANAGE_MERGE_PATH` in `building-me.contract.spec.ts` all exist as entities) — but the **dashboard Paths panel (and effective_contract, flows/traces, MCP neighbors) query from the disconnected endpoint node and find nothing.** Django doesn't have this problem because the ViewSet method is the unified endpoint+handler.

**Primary fix:** in the NestJS/TS controller extractor, emit an edge bridging the `http_endpoint_definition` to its handler `Operation` (same controller method, identifiable by file+line / decorator association). Direction/kind consistent with how endpoints relate to handlers elsewhere (e.g. an endpoint→handler `HANDLED_BY`/`DEFINES`, or unify them). Once bridged, traversing from the endpoint reaches the handler → its `VALIDATES`(payload) + `CALLS`(downstream) chain, so Paths "Parameters" and "Downstream" populate, and cross-repo flows can continue into the backend (relates to #4316 B2).

**Secondary edges the handler is ALSO missing (verify + add):**
1. **Response**: `handler → BuildingMeMergeResponse` (a `RETURNS`/response edge from the handler's return type) — currently absent; Paths "Response" shows (none).
[...]

Issue #1615, opened 2026-05-22
extraction: cross-repo HTTP link normalization — 26.5% coverage / 1115 orphan calls (strip /api/v1/ prefix, {id}≈{pk})


cajasmota:
Calibration: only 402 of 1517 cross-repo HTTP links resolve (26.5%); 1115 orphan calls. Frontend/mobile call e.g. /inspections/{id}/create-deficiencies/ while backend defines /api/v1/inspections/{pk}/create-deficiencies — they fail to match on the /api/v1/ prefix and {id} vs {pk}. FIX the cross-repo HTTP matching (internal/resolve / links P4 pass): normalize both sides before matching — strip the configurable API prefix (per upvate_core/urls.py:26 the backend mounts under /api/v1/), treat {id}≈{pk}≈{param} as equivalent path params, normalize trailing slashes. Verify on upvate via xrepo-verify + the served graph: cross-repo HTTP coverage rises well above 26.5% (toward the 1517 potential), orphan_calls drop; spot-check mobile create-deficiencies → backend route. This is the same gap behind benchmark q05/q11/q12 and the headline cross-repo value prop.

PR #6494, opened 2026-08-21
fix(#6458): repairing a dead kind gate revived a bug the dead gate had been hiding

### LINKED ISSUE #6458: three more consumer-side gates in process_flow.go still compare against the kind #1217 retired, so they are inferred dead on every current graph
Three consumer-side gates in `internal/engine/process_flow.go` still compare against the pre-#1217 `"http_endpoint"` kind literal, which no current code path emits:

- **`:778` `buildConsumerEndpointFileSet`** — `strings.ToLower(e.Kind) != "http_endpoint"`, then checks `pattern_type == "http_endpoint_client_synthesis"`. Only `http_endpoint_call` carries that property, so the kind gate rejects every entity that could have passed the second one.
- **`:957` `isConsumerHTTPEndpoint`** — identical shape.
- **`:1001`** — a `switch` whose endpoint-ish arm lists `"http_endpoint"` but neither of the split kinds.

INFERRED dead on every post-#1217 graph, by the same reasoning that was CONFIRMED for #6449: `internal/engine/http_endpoint_synthesis.go:534-537` stamps `http_endpoint_call` on consumer-side synthetics and `http_endpoint_definition` on producer-side ones, and the consts at `:63`/`:67` are literally those strings. Nothing emits the bare legacy kind any more.

## Why these were not fixed alongside #6449

Deliberate. `process_flow.go` sits in an 11MB, 277-test-file package, and **neither function has any direct test** — `grep -rln` for either name hits only `process_flow.go` itself. An untested drive-by change to process-flow output, made while clearing a release blocker, is the wrong trade. It needs its own TDD arm: a failing test first, then the fix, then a mutant proving the test can see it.

## Why it is worth doing

#6449 was exactly this shape and had silently disabled a whole shipped feature (the ADR-0015 repair queue) for an unknown number of releases. The root cause there generalises: `types.IsHTTPEndpointKind` was adopted across `internal/dashboard/` (11 call sites) and had **zero** callers in `internal/enrichment` or `internal/engine`. The post-#1217 sweep covered the dashboard and stopped, leaving raw string comparisons behind everywhere else.

So the question is not whether these three are dead — it is how much else the sweep missed.

## Scope

1. Write the missing tests for `buildConsumerEndpointFileSet` and `isConsumerHTTPEndpoint` **first**, against a kind the current synthesis path actually emits. Confirm they fail.
2. Replace the literals with `types.IsHTTPEndpointKind`.
3. Verify a mutant reverting each to the legacy literal dies.
4. Establish what the consumer-side behaviour change actually is once these gates start passing entities — this is not a no-op fix, it turns on code that has not run in a long time, so the output delta needs measuring rather than assuming.

## Also in this class, lower severity

`internal/enrichment/pricing.go:38` — `TokensPerEntity`'s `case "http_endpoint", …` omits both split kinds, so post-#1217 endpoints fall through to the default token estimate. Cost estimation only, no correctness impact, but the same one-word fix.

And a producer-side one worth a separate look: `internal/engine/webhooks_edges.go:111` still **emits** `Kind: "http_endpoint"`. If that is intentional it should be documented; if not, it is minting entities in a kind the rest of the pipeline has stopped recognising.

## Search boundary, stated

The #6449 sweep covered `internal/**/*.go` only — 335 hits partitioned by comparison shape, plus sweeps for `HTTPEndpointKindLegacy` and `EqualFold`. **`cmd/`, `site/` and non-Go files were not searched.** Anyone taking this arm should widen that before claiming the class is closed.

cajasmota:
Closes #6458

Three post-#1217 stale kind literals in `internal/engine/process_flow.go` — `:778` `buildConsumerEndpointFileSet`, `:957` `isConsumerHTTPEndpoint`, `:1001` `chainCrossesExternalLib`. Each compared against a kind that #1217 split, so two of the three ran and returned false unconditionally.

## Repairing a dead gate revived a bug the dead gate had been hiding

Widening `:778` broke a **pre-existing** test. Root cause: the #754 file-coarse fallback predates #1639. #1639 overturned its premise and encoded the correction *inside* `chainCrossesRepoBoundary` — but never gave it to the fallback, **because the `:778` kind gate had already rendered the fallback inert.** There was nothing to reconcile at the time. Un-breaking the gate restores pre-#1639 semantics along with it.

A uniform three-line repair would have shipped that silently — the same shape as #6449, where a stale literal disabled a shipped feature invisibly.

## Round 1's fix was wrong, and the review falsified it by execution

Round 1 added a `chainReachesConsumerEndpoint` guard asserting *"when the chain DID reach one, `chainCrossesRepoBoundary` has already given the authoritative answer."*

It has not. `hasContinuation` is computed **chain-globally** — one handler-continuation edge anywhere disables the unresolved-consumer check for *every* step. A chain that resolves one HTTP call in-repo and then makes a second, unresolved one is reported intra-repo. The #754 fallback had been accidentally compensating for exactly that; the guard removed the compensation and asserted it was never needed.

| probe | chain | round 1 | guard removed |
|---|---|---|---|
| A — call resolves same-repo, handler makes a **second unresolved** call (BFF/gateway) | `caller,callA,defA,handler,callB` | `cross_stack="false"`, reason empty | `"true"` |
| B — fan-out: one endpoint resolves in-repo, a second is unresolved | `caller,callA,defA,handler,repo` | `cross_stack="false"` | `"true"` |

In probe A the chain **terminates on an unresolved cross-repo consumer synthetic** and the verdict is `false` with an empty reason.

## What landed instead: per-endpoint resolution at the source

New `resolvedConsumerEndpoints(doc)` marks each consumer synthetic resolved iff `consumer --FETCHES--> http_endpoint_definition <--IMPLEMENTS-- handler`, all within the document. `chainCrossesRepoBoundary` consults it per endpoint; `buildConsumerEndpointFileSet` excludes resolved consumers, so a file qualifies for the #754 fallback only when it holds an *unresolved* call.

**Gating the fallback alone would not have worked:** probe A's entry file contains no consumer synthetic at all, so its `cross_stack` can only come from `chainCrossesRepoBoundary`. No change to the fallback's gate can reach it. The per-endpoint fix reaches both probes.

`chainReachesConsumerEndpoint` and its comment are **deleted** — reconciliation now happens in the file set, which closes the round-1 surviving mutant structurally rather than by adding an assertion.

## The corpus number, including the part that does not flatter this PR

Measured on a real indexed group graph (427k entities / 1.85M relationships / 282 emitted process flows / 114 consumer synthetics, 4 resolving in-repo):

- probe A shape (`consumer → continuation → consumer`): **0**
- probe B shape: **0**
- chains touching *any* consumer synthetic: **0**, and `cross_stack` is true for **0 of 282**

So the false-negative class this fix closes is empty on that corpus — **and so is the PR's headline delta there.** The consumer synthetics are not orphans (202 incoming CALLS/FETCHES, 71 outgoing); the BFS simply never lands on one within the `MaxProcesses=300` longest-chain-ranked sample. **The fix is justified by correctness and by the probes, not by a corpus delta**, and the commit message says so.

*Measurement caveat:* the corpus `graph.fb` is format v5 and this branch requires v6, so `minSupportedFBFormatVersion` was temporarily lowered (v5→v6 added only `end_line`, unused here). `internal/graph/load.go` was restored by `cp`, verified byte-identical, and is not in the commit.

## "Only `webhooks_edges.go:111` keeps the legacy arm live" was false

There are **nine** live non-test emitters of the legacy kind — eight via `httpEndpointKind` (`http_endpoint_synthesis.go:3158`, `django_urlconf_nested.go:182,:276`, `django_drf_actions.go:539,:2833,:3028`, `django_admin_routes.go:223`, `java_annotation_routes.go:864`) plus the literal at `webhooks_edges.go:111`. `types/kinds.go:186` ("No extractor emits this kind after #1217") was also wrong. Both corrected. The `:778`/`:957` inertness conclusion is unchanged — all nine are producer-side.

## Two documentation defects fixed

- The file header at `:41-44` claimed `crosses_external_lib` fires only when the chain *also terminates* in an external-lib node. The code has never done that — it is an OR over every step. Corrected, and annotated with the fact that the property has exactly three non-test references, all inside this file, so nothing reads it.
- `response_shape_corpus.go`'s `"Route": {"http_endpoint"}` widened to the three constants, with `TestLookupHandler_RouteEquivalence_6494` — **the first test that function has ever had** (mutant R1 survived before it, dies after).

## Mutants

| id | mutation | result |
|---|---|---|
| W1 | any endpoint kind counts as a consumer in `resolvedConsumerEndpoints` | KILLED |
| W2 | any endpoint kind is a cross-repo bridge (round 1's survivor) | KILLED |
| W3 | drop the resolved-consumer exclusion from the file set | KILLED |
| W4 | any edge kind resolves a consumer, not just FETCHES | KILLED |
| W5b | `served := defs` (IMPLEMENTS no longer required) | KILLED |
| N1 | invert the per-endpoint resolution test | KILLED |
| N2 | `resolvedConsumerEndpoints` always empty | KILLED |
| R1 | revert the `"Route"` handler-table widening | KILLED |

W5 (the first IMPLEMENTS-widening attempt) was an **equivalent mutant** — the `r.Kind != "IMPLEMENTS"` guard short-circuits it — and was replaced with W5b, the real widening, which dies on probe C.

**Coordinator mutant, independently applied and scored:** coarsen the exclusion from per-consumer to per-file, so any file containing one *resolved* consumer is excluded wholesale — realistic drift, given the surrounding code is explicitly file-coarse. `go vet` exit 0 before scoring; both mutation targets asserted count 1. **KILLED** by `TestProcessFlow_6494_ProbeB_FanOutUnresolvedSiblingInEntryFile`. Restored by `cp`, shasum `97ceeb9c…` verified identical.

## Verification

`gofmt -l` → 0 files · `go vet` engine/types/graph → 0 · `go test ./internal/engine/` → exit 0 (31.8s) · `./internal/quality/...` → 0 · `./internal/types/... ./internal/graph/...` → 0. No full repo suite.

### DIFF
--- a/internal/engine/process_flow.go
+++ b/internal/engine/process_flow.go
@@ -38,10 +38,21 @@
 //
 // The previous heuristic (`chainTouchesHTTP` — any step appears on an
 // IMPLEMENTS / ROUTES_TO / SERVES edge boundary) is preserved as a separate
-// `crosses_external_lib` property whenever the chain ALSO terminates in
-// an external-library SCOPE.External / SCOPE.ExternalAPI node. That
-// property captures the original intent ("this process bottoms out in a
-// third-party dep") without conflating it with cross-repo traversal.
+// `crosses_external_lib` property.
+//
+// #6494 — this header used to describe that property as firing "whenever
+// the chain ALSO terminates in an external-library SCOPE.External /
+// SCOPE.ExternalAPI node", i.e. an AND on the terminal step. The code has
+// never done that: chainCrossesExternalLib is an OR over EVERY step —
+// boundary-set membership, an HTTP-endpoint-ish kind, or an external-lib
+// kind, at any position. The function doc at chainCrossesExternalLib is
+// the accurate one ("this process touches an HTTP handler" OR bottoms out
+// in a third-party dep); this paragraph is corrected to match it rather
+// than the other way round, because no code path reads the property —
+// `crosses_external_lib` has exactly three non-test references, all inside
+// this file: this comment, the emit site, and the widening comment on the
+// function itself. Nothing consumes it, so widening it in #6458 cannot
+// have moved any downstream behaviour.
 //
 // Internal HTTP handlers (a controller method that implements a same-repo
 // route synthetic) are intentionally NOT cross_stack: the BFS never
@@ -182,7 +193,12 @@ func RunProcessFlowWithCompanions(doc *graph.Document, companions []*graph.Docum
 	// extractors catch up (eventual JS/TS class-field support) the BFS
 	// will reach the endpoint structurally and this fallback becomes a
 	// no-op overlay on top of the precise FETCHES-edge signal.
-	consumerEndpointFiles := buildConsumerEndpointFileSet(doc)
+	// #6494 — consumer synthetics whose call resolves into a same-repo
+	// handler are NOT cross-repo bridges (#1639). The set is per-endpoint
+	// and feeds both the file-coarse fallback below and
+	// chainCrossesRepoBoundary.
+	resolvedConsumers := resolvedConsumerEndpoints(doc)
+	consumerEndpointFiles := buildConsumerEndpointFileSet(doc, resolvedConsumers)
 
 	// Build CALLS adjacency across doc + companions. Edges with explicit
 	// `confidence < 0.5` are excluded so fuzzy global-fallback matches don't
@@ -337,7 +353,7 @@ func RunProcessFlowWithCompanions(doc *graph.Document, companions []*graph.Docum
 		if entry == nil || (terminal == nil && !terminalIsPhantom) {
 			continue
 		}
-		crossStack, crossReason := chainCrossesRepoBoundary(chain, byID, adj)
+		crossStack, crossReason := chainCrossesRepoBoundary(chain, byID, adj, resolvedConsumers)
 		if !crossStack {
 			// Fallback (#754): any chain whose entry file contains a
 			// consumer http_endpoint entity is treated as cross_stack
@@ -350,16 +366,15 @@ func RunProcessFlowWithCompanions(doc *graph.Document, companions []*graph.Docum
 			// synthetics (pattern_type=http_endpoint_synthesis) which
 			// buildConsumerEndpointFileSet excludes.
 			//
-			// #6458 — the fallback is for chains that could NOT reach a
-			// consumer bridge structurally. When the chain DID reach one,
-			// chainCrossesRepoBoundary has already given the authoritative
-			// answer (including #1639's same-repo handler-continuation
-			// rule), and a false verdict there means the call resolved
-			// in-repo. Overriding it from a file-coarse proxy would revive
-			// the pre-#1639 semantics that this gate being dead had been
-			// hiding.
-			if entry != nil && consumerEndpointFiles[entry.SourceFile] &&
-				!chainReachesConsumerEndpoint(chain, byID) {
+			// #6494 — reconciling the fallback with #1639 is done in
+			// buildConsumerEndpointFileSet, not here: a file counts only
+			// when it holds a consumer synthetic that does NOT resolve
+			// into a same-repo handler. Guarding on "the chain reached
+			// SOME consumer bridge" instead would be wrong — a chain can
+			// resolve one call in-repo while the entry file holds a
+			// second, unresolved one (fan-out), and that second call is
+			// exactly what the fallback exists to catch.
+			if entry != nil && consumerEndpointFiles[entry.SourceFile] {
 				crossStack = true
 				crossReason = "entry file contains consumer http_endpoint synthetics (BFS chain didn't reach the bridge structurally — see #754 fallback)"
 			}
@@ -782,10 +797,20 @@ func buildHTTPBoundarySetMulti(doc *graph.Document, companions []*graph.Document
 // a chain's entry lives in a file with consumer endpoints that the BFS
 // couldn't structurally reach (typically because the per-language
 // extractor doesn't surface class-field arrow methods as entities).
-func buildConsumerEndpointFileSet(doc *graph.Document) map[string]bool {
+//
+// #6494 — `resolved` (from resolvedConsumerEndpoints) carries the IDs of
+// consumer synthetics whose call resolves into a SAME-repo handler. Those
+// are intra-repo by #1639 and are excluded here, so a file qualifies only
+// when it holds at least one UNRESOLVED consumer call. That is what keeps
+// the fallback from reviving pre-#1639 semantics now that the kind gate
+// below no longer rejects every candidate.
+func buildConsumerEndpointFileSet(doc *graph.Document, resolved map[string]bool) map[string]bool {
 	out := map[string]bool{}
 	for i := range doc.Entities {
 		e := &doc.Entities[i]
+		if resolved[e.ID] {
+			continue
+		}
 		// #6458 — this gate compared against the bare pre-#1217
 		// "http_endpoint" literal, which the consumer synthesiser stopped
 		// emitting when #1217 split the kind into http_endpoint_call /
@@ -855,6 +880,7 @@ func chainCrossesRepoBoundary(
 	chain []string,
 	byID map[string]*graph.Entity,
 	adj *callsAdjacency,
+	resolvedConsumers map[string]bool,
 ) (bool, string) {
 	// #1639 — repo-aware cross-repo flag. A chain is cross-repo ONLY when it
 	// genuinely leaves the source repo. With the handler-continuation fix, an
@@ -866,19 +892,19 @@ func chainCrossesRepoBoundary(
 	// a cross-repo signal; we require an authoritative boundary marker:
 	//
 	//   1. A phantom cross-repo CALLS edge (#769): target_repo != this repo.
-	//   2. A consumer http_endpoint synthetic that the chain does NOT resolve
-	//      INTO a same-repo handler (no handler-continuation edge leaves it).
-	//      An unresolved consumer synthetic is, by construction, a call whose
-	//      backend lives in another repo (the cross-repo HTTP linker pairs it
-	//      with a producer elsewhere). When the chain DID continue into a
-	//      same-repo handler, the call resolved locally and is intra-repo.
-	hasContinuation := false
-	for i := 1; i < len(chain); i++ {
-		if adj != nil && adj.handlerCont[edgeKey{chain[i-1], chain[i]}] {
-			hasContinuation = true
-			break
-		}
-	}
+	//   2. A consumer http_endpoint synthetic that does NOT resolve INTO a
+	//      same-repo handler. An unresolved consumer synthetic is, by
+	//      construction, a call whose backend lives in another repo (the
+	//      cross-repo HTTP linker pairs it with a producer elsewhere). When
+	//      the call DID resolve into a same-repo handler, it is intra-repo.
+	//
+	// #6494 — that second test is PER-ENDPOINT (resolvedConsumers is keyed by
+	// consumer-synthetic ID). It used to be chain-GLOBAL: a single
+	// handler-continuation edge anywhere in the chain disabled the
+	// unresolved-consumer check for every step, so a chain that resolved one
+	// call in-repo and then made a second, unresolved one (the BFF/gateway
+	// shape: caller → callA → defA → handler → callB) was reported intra-repo.
+	// The #754 file-coarse fallback had been masking that; it no longer does.
 
 	// Walk pairwise — every transition has an originating edge.
 	for i := 1; i < len(chain); i++ {
@@ -892,13 +918,13 @@ func chainCrossesRepoBoundary(
 		}
 		// A consumer synthetic that did NOT resolve into a same-repo handler
 		// is a call whose backend lives in another repo.
-		if !hasContinuation && isConsumerHTTPEndpoint(byID[to]) {
+		if isConsumerHTTPEndpoint(byID[to]) && !resolvedConsumers[to] {
 			return true, fmt.Sprintf("unresolved consumer http_endpoint at step %d (%s) — backend in another repo", i, to)
 		}
 	}
 	// Also check the entry itself — a Process whose entry IS a consumer
 	// synthetic (unusual but possible) still crosses repos.
-	if !hasContinuation && len(chain) > 0 && isConsumerHTTPEndpoint(byID[chain[0]]) {
+	if len(chain) > 0 && isConsumerHTTPEndpoint(byID[chain[0]]) && !resolvedConsumers[chain[0]] {
 		return true, fmt.Sprintf("unresolved consumer http_endpoint at step 0 (%s)", chain[0])
 	}
 	return false, ""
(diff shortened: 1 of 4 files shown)


cajasmota:
## Round 2 review: REQUEST-CHANGES — and a claim I relayed in the PR body is false

I wrote in this PR's description that `crosses_external_lib` "has exactly three non-test references, all inside this file, so nothing reads it." **That is wrong, and it was the entire safety argument for the `:1001` widening.**

### F1 (HIGH) — the property drives a user-visible badge

The audit stopped at `internal/`. The property is serialised onto Process entities at `process_flow.go:434` and rendered in the dashboard:

```
webui-v2/src/data/types.ts:714     crosses_external_lib?: boolean;
webui-v2/src/routes/flows.tsx:254    {flow.crosses_external_lib && ( … "external lib" badge )}
webui-v2/src/routes/flows.tsx:1432   {flow.crosses_external_lib && ( … "external lib" badge )}
```

After this PR, a flow stepping through an **in-repo** `http_endpoint_call`/`http_endpoint_definition` — a purely first-party HTTP hop with no third-party dependency anywhere — renders a badge reading **"external lib"**.

The PR's own comment concedes the semantic gap ("the property's stated meaning: *this process touches an HTTP handler*") without noticing the UI string says something else. This is the same defect class the PR exists to fix: round 1 shipped unobserved comment claims; round 2 shipped one that is **observable and wrong**, and load-bearing.

The count is also wrong even scoped to `internal/` — four references (`:41`, `:52`, `:434`, `:1126`), not three.

Round 3 must either narrow the `:1001` repair so a first-party hop does not set the property, or correct the label, type and comment. Deleting the comment and keeping the behaviour is not an option.

### F2 (HIGH) — the `IMPLEMENTS` leg over-constrains "resolved", and the PR canonises the false positive

The `call --FETCHES--> definition` edge **already is** the in-repo resolution signal — `http_endpoint_resolve.go:424-437` stamps it `pattern_type="http_endpoint_split_resolved"`, `resolved="true"`, emitted only when a matching definition was found. The `IMPLEMENTS` leg asks a different question: whether *handler binding* also succeeded. Conflating them turns "the handler wasn't bound" into the verdict **"the backend lives in another repo."**

Handler-binding failure is a counted, first-class outcome, not a corner case: `bridgeEndpointToHandler` (`:945`) is the sole producer of that edge and is reached only on successful lookup, with `stats.NoHandlerProp`, `HandlerDropped`, `HandlerUnresolved` and `HandlerUnresolvedKept` (*"the endpoint survived unenriched"*) counting the failures — plus the #4319 co-location fallback that exists because binding fails on real NestJS code. And `http_endpoint_synthesis.go:518` gates `source_handler` on `if refName != ""`, so regex/YAML/mount-derived routes never carry a handler ref at all.

**`TestProcessFlow_6494_ProbeC_ResolutionBoundaries` case 1 locks this in** — "definition with no handler IMPLEMENTS edge" asserts `cross_stack == "true"` and emits *"backend in another repo"* while the definition sits in the same `doc`. This is the false-positive direction round 1's probes structurally cannot cover.

Note the "4 of 114 consumers resolve in-repo" figure was measured **with** this predicate, so it cannot bound the blast radius. The measurement that would: count `http_endpoint_definition` entities with no inbound `IMPLEMENTS` on the real graph. Round 3 takes it.

### F3 (MEDIUM) — companion-document asymmetry

Every sibling set in `RunProcessFlowWithCompanions` spans `doc` + `companions` (`byID` `:152-166`, `buildHTTPBoundarySetMulti` `:183`, `buildCallsAdjacencyMulti` `:209`). `resolvedConsumerEndpoints(doc)` at `:197` is the only one built doc-only, yet it is consulted against `byID[to]`, which does span companions — so a consumer synthetic reached in a companion can never be in the map. Reachable via `internal/cli/links.go:534`.

The incremental path is confirmed unaffected (`incremental.go:1386` runs over the finalized full-repo doc).

### Surviving mutants

| id | mutation | result |
|---|---|---|
| **A** | drop the `defs[r.ToID]` membership check → `served[r.ToID] = true` | **SURVIVED, proven non-equivalent** |
| **D** | drop `&& !resolvedConsumers[chain[0]]` from the entry-step check | **SURVIVED** |
| F | drop the `r.FromID == r.ToID` self-edge guard | survived; non-equivalence not claimed |

**Mutant A** was proven by discriminating fixture, not asserted: a consumer FETCHES a `SCOPE.Interface` that a `SCOPE.Class` IMPLEMENTS, with an unrelated definition present. Pristine → `false`; mutated → `true`. Under it, `served` degenerates to *every IMPLEMENTS target in the repo* — in Go/Java/Kotlin corpora, essentially every interface — silently marking consumers resolved and suppressing `cross_stack`. That is exactly the false-negative class #6458 exists to close, with the suite green. W5b (`served := defs`) was scored, but never the *type* check on the IMPLEMENTS target. Two independent reviewers converged on this gap.

**Mutant D:** N1 and N2 both perturb *both* consultation sites, so they die on the pairwise site alone and prove nothing about the entry-step site. The `chain[0]` term at `:1131` has zero coverage.

### Cleared, and worth recording

**F4 — the v5-read concern does not hold.** `fbversion/version.go:34` shows v6 added only the appended `end_line` slot; a v6 reader on a v5 buffer differs only in `EndLine == 0`, with `Kind`, `Properties` and the relationship table at unchanged vtable offsets. **None of the four zeros is explained by the lowered `minSupportedFBFormatVersion`** — the instrument is sound on that axis and the PR's disclosure of the caveat was accurate.

The nine-emitter enumeration and the `types/kinds.go:186` correction were spot-verified and hold, including that `webhooks_edges.go:111` stamps `pattern_type="webhook_synthesis"` so `isConsumerHTTPEndpoint` correctly rejects it.

The chain-global → per-endpoint correction remains the right move, and probes A and B are real. Round 3 fixes F1, F2, F3 and the two survivors.

cajasmota:
## Coordinator mutant on round 3 — SURVIVED, and it targets the exact property the new comment claims

`collectResolvedConsumerEndpoints`'s doc comment states:

> *"Classification and edge walk are both scoped to this document so a definition can only resolve a call from the same repo."*

**Nothing asserts that.** I replaced `resolvedConsumerEndpointsMulti`'s per-document loop with a merge-then-classify:

```go
merged := &graph.Document{}
// append doc.Entities/Relationships, then every companion's
collectResolvedConsumerEndpoints(merged, resolved)
```

`go vet ./internal/engine/` exit 0 **before** scoring · mutation target asserted count 1 · `GOMAXPROCS=4 go test -count=1 ./internal/engine/` → **exit 0, SURVIVED** · restored by `cp`, shasum `30a84106…` verified identical, tree clean.

### Why this is not equivalent, and why it matters

Merging before classification lets a `FETCHES` edge in document A resolve against a `http_endpoint_definition` that exists only in document B. Cross-repo `FETCHES` edges are precisely what the companions path exists to carry (`internal/cli/links.go:534`), so the shape is reachable by construction — not hypothetical.

And the consequence is the **inversion of this PR's whole purpose**: a consumer whose backend genuinely lives in another repo would be marked *resolved*, suppressing `cross_stack` on the one case where it is most certainly true. #6458 exists because those gates were inert and under-reported cross-repo hops; this mutant makes them under-report again, silently, with the suite green.

`M6` (drop companions entirely) and `M5` (drop `consumers[r.FromID]`) both die, so the companion *plumbing* is pinned — but neither perturbs the per-document **scoping**, which is the property the comment advertises.

### What would close it

A companion fixture where the consumer's `FETCHES` target definition lives in a *different* document than the consumer: the map must **not** mark it resolved, and `cross_stack` must stay true. The existing `CompanionConsumerResolves` test has the consumer and its definition in the same companion, so it cannot distinguish the two implementations.

Severity: below F1/F2 — it requires a cross-document FETCHES edge to fire — but it is the same class this PR has been iterating on for three rounds, and the comment asserting the property is exactly the kind of unobserved claim the last two rounds removed.

Handed to the round-3 reviewer to confirm or refute rather than acted on unilaterally.

cajasmota:
## Round 3 review: APPROVE, pending two one-line fixture additions

All three round-2 defects are properly fixed and the implementation is correct. Both surviving mutants are **test-coverage gaps, not behavioural defects**, and each closes with a single line inside an existing fixture — dispatched as a trivial follow-up commit rather than another round.

### My reachability claim was wrong

I argued my surviving merge-then-classify mutant was reachable, on the grounds that cross-repo `FETCHES` edges are what the companions path carries. **That is false**, and the refutation is three-layered:

- Cross-repo edges are materialised as **`Kind:"CALLS"` phantoms with `cross_repo=true`** (`internal/links/phantom_edges.go:148-160`), never FETCHES.
- `internal/cli/links.go:534` hands other repos' documents over as read-only companions and carries **no** FETCHES edge across a document boundary.
- Even accidental cross-document matching is impossible: `graph.EntityID` (`internal/graph/graph.go:259`) salts the sha256 with the repo tag, so no ID in doc A can collide with one in doc B.

So the mutant is a **defensive-invariant gap**, not the live inversion I described. Severity low. My points about the *tests* stood — `…Multi_6494_SpansCompanions`'s cross-document negative passes `companions=nil`, so under the mutant `merged == doc` and the assertion is identical; `CompanionConsumerResolves` has its consumer and definition in the same companion. Neither can discriminate.

That is twice now on this PR that I have over-claimed reachability for a surviving mutant and been corrected by execution. The mutants were worth raising both times; my reasoning about *why* was not load-bearing and should not have been stated as confidently as it was.

### MUT-1 — a second survivor, closer to reachable than mine

In `collectResolvedConsumerEndpoints`: `isHTTPEndpointDefinition(e)` → `types.IsHTTPEndpointKind(strings.ToLower(e.Kind))`. `go vet` 0, target count 1 → **exit 0, SURVIVED.**

`defs` then becomes *any* endpoint-kind entity that is not already a consumer — legacy `http_endpoint` (the webhook synthetics among the nine live producers this PR enumerates) and non-consumer `http_endpoint_call` stubs — so a consumer FETCHES edge landing on one marks it resolved and **suppresses `cross_stack`**.

**ProbeC case 3 pins the TO leg only against `SCOPE.Class` — a far miss.** The near miss is unpinned. This is round 2's mutant-A shape exactly, and it is the most likely drift direction *because* this PR widens the other two gates with that very call. The PR's own corpus note — 3 consumer FETCHES edges targeting a non-definition — means it cannot be ruled out on real data.

### Verified

- **Round-2 survivors are genuinely dead.** MUT3 (drop `!defs[r.ToID]`) → fails in `ProbeC_ResolutionBoundaries`; MUT2 (drop the entry-step guard) → fails in `ChainCrossesRepoBoundary_6494_EntryStepGuard`.
- **The F1 revert is real** — `chainCrossesExternalLib`'s switch is back to the legacy literal plus Endpoint/Route/ExternalAPI/`scope.external`; the split kinds are absent (`:1198-1205`).
- **The corrected claims are now observable and correct.** `webui-v2/src/data/types.ts:714`, `flows.tsx:254`, `flows.tsx:1432` all verified at the cited lines, and the badge string really is `external lib`. The four in-package refs are stated without line numbers so they cannot drift.
- **ProbeC case 1 is not an inversion-to-fit.** It asserts the resolver's own signal — `FETCHES` → in-doc definition means resolved — and the code evidence backs it independently of the corpus: `bridgeEndpointToHandler`'s failure counters and the `refName != ""` gate make handler binding a routinely-failing separate step.

### The F2 false-negative residual — bounded, and worth one sentence

Fuzzy path matching (#1615/#2547) can resolve a call against a same-repo route — a BFF proxy, a mock server, a path defined in both repos — while the runtime target is remote. But `chainCrossesRepoBoundary` checks the phantom `cross_repo` edge **first** (`:939`), so whenever the linker found the remote producer, `cross_stack` still fires. The gap is confined to calls the linker never paired, where there is no evidence of a remote producer anyway. Not blocking; worth a sentence in the function doc.

### The corpus numbers were NOT independently verified

Confirming them needs an export against the live daemon or a full in-process load of the 427k-entity graph, and with another Go agent active and RAM low the reviewer judged both out of budget rather than risk the user's MCP. **Stated as unverified rather than assumed true** — the right call, and the F2 conclusion does not rest on them: the code-level evidence (handler-binding failure counters, the `refName != ""` gate) is sufficient on its own.

Issue #6495, opened 2026-08-21
baseline.json's #6454 note says the dangling CALLS edges resolve with #6337 — that arm is gated to EXTENDS/IMPLEMENTS and never touched them


cajasmota:
Found while drafting the 0.3.0 CHANGELOG. Small, but it is a false statement sitting in a file people read to understand a gate.

`baseline.json`'s note for #6454 says the ~2,205 new dangling `CALLS` edges it introduced **"dangle until #6337"**.

That is wrong, and the #6337 commit says so explicitly: the arm in question is gated to `EXTENDS`/`IMPLEMENTS`, so #6337 never had any effect on those `CALLS` edges. They do not dangle "until" anything — the forward reference points at work that could not have resolved them.

The same wrong sentence was in `CHANGELOG.md`'s `[Unreleased]` text and is corrected in the 0.3.0 draft. This is the code-side copy, which the draft could not fix.

## Why it is worth fixing rather than leaving

A note in `baseline.json` explaining why a number is what it is, is load-bearing for whoever next re-measures that baseline — and #6490 has just established that the baseline is 82 commits stale and will need re-measuring. Someone doing that work will read this note, look up #6337, find it closed, and reasonably conclude the dangling edges should now be resolved. They will not be, and there will be no explanation for why.

The fix is to state what actually governs those edges instead of naming an issue that does not.

While in there, it is worth checking whether any other note in `baseline.json` carries a forward reference to an issue that has since closed without having the described effect — this one was only caught by chance.

Issue #6490, opened 2026-08-21
nine golden fixtures assert zero relationships — and erlang-otp-mini omits the key entirely, so the grader cannot fail it


cajasmota:
Found incidentally while correcting fixture paths for #6464 (PR #6489).

## The finding

**`erlang-otp-mini` and `haskell-warp-mini` both report `relationships: 0 / 0 expected`.** Every edge row in both fixtures is `nice_to_have`, so neither asserts a single must-have relationship.

They grade 20/20 and 24/24 on entities and 0/0 on relationships — which reads as a perfect score and is, for entities. For edges they assert nothing that can fail.

## Why this matters beyond two fixtures

The whole point of a golden fixture is that it goes red when the graph gets worse. An entity-only fixture pins that the extractor still *finds* things; it pins nothing about whether they are *connected correctly* — and connectivity is where most of this cycle's defects lived: dangling handler refs, self-referential IMPLEMENTS edges, hierarchy edges that were never emitted, import placeholders capturing names repo-wide.

For erlang specifically this compounds with **#6370**: erlang is one of the eight languages emitting no `EXTENDS`/`IMPLEMENTS` at all, and its `-behaviour(...)` declarations are collected into a *property* rather than an edge. Its golden fixture asserting no must-have relationships is why nothing objects. **#6370's grounding already noted the fixture bakes that absence in as correct** — this is the same gap seen from the other side.

## What to check before fixing

**Do not simply promote the existing `nice_to_have` rows to `must_exist`.** Several may be failing today for good reasons, and flipping them would produce a red fixture that says "someone flipped a flag" rather than a red fixture that says "the graph got worse". Grade them first and see which already pass.

Then, for each promoted row, **prove it load-bearing by mutation** — break the production code that emits the edge and confirm the fixture goes red. A row promoted without that check is the same class of decoration this repo keeps finding: `swift-package-mini` sat at `relationship_found: 0/7` for exactly this reason, and `solidity-mini` carries two forbidden rows that can never fire.

## Worth checking while here

Whether any **other** fixture has the same shape — all-`nice_to_have` edges, or a suspiciously low `relationship_expected` relative to its entity count. The audit for #6464 covered path resolvability across all 23 fixtures but did not look at must-have coverage. `baseline.json` has the per-fixture numbers to make that a quick scan.

## Related

- **#6488** — the fixture format cannot assert entity subtype, entity absence, or file-anchored edges. Different gap, same theme: what the harness can and cannot see.
[...]

Issue #6476, opened 2026-08-21
the grader reports every to_bare_name row as resolved, so a fixture row that cannot match is diagnosed as an extractor miss


cajasmota:
Found while auditing #6441. This is the reason that whole defect class survived as long as it did: **the grader's own diagnostic points the reader away from the fixture and at the extractor.**

## The defect

`internal/quality/diff.go:308`:

```go
toResolved := len(toCands) > 0 || er.ToBareName != ""
```

A row using `to_bare_name` is reported as **resolved unconditionally** — the bare name is never checked against anything. `toResolved` then feeds `report.go:167-172`, which renders the miss as:

```
(both endpoints exist; edge not emitted)
```

## Why that message is actively harmful

That message tells the reader: *the fixture is fine, the extractor failed to emit the edge.* For a `to_bare_name` row whose target actually resolves, the truth is the opposite — **the row cannot match anything**, because `toCands` is populated only when `er.ToName != ""` (`diff.go:299-337`), so the sole match path is literal `ToID` equality, and a resolved target carries an entity ID rather than a bare string.

Live instance: all seven `DEPENDS_ON` rows in `internal/quality/golden/swift-package-mini/expected.json` report exactly this message. `baseline.json` records `entity_found: 6/6, relationship_found: 0/7`. Every one of the six endpoints is declared `must_exist: true` and present. Anyone investigating that 0/7 is told, by the tool, to go and look at the swiftpm extractor.

## Scope

47 `to_bare_name` rows across the 23 fixtures. Every one of them currently reports `ToResolved == true` regardless of whether the name corresponds to anything, so **the diagnostic carries no information for that row shape at all** — it is not merely wrong for the seven swift rows, it is vacuous for all 47.

## Fix

Resolve `ToBareName` against the entity set the grader already holds (`doc.Entities`) before setting `toResolved`. That is the same lookup #6441's proposed mechanism-1 check needs, so the two should land together:

- if the bare name **does** match a declared entity, the row is the #6441 defect — report *that*, not "edge not emitted";
- if it matches nothing, `toResolved` is legitimately false and the existing message is correct.

## Acceptance

A test asserting that a `to_bare_name` row whose target resolves is **not** reported as "both endpoints exist" — driven through the real report path, not by calling the predicate directly. Without that, the fix is unfalsifiable: the current behaviour is a silent `||` short-circuit and no existing test observes `ToResolved` at all.

Worth checking whether `FromBareName` has the same shape on the from-side.

Issue #6471, opened 2026-08-21
django: the route path is passed as its own handler ref, so ~187 endpoints carry a self-referential IMPLEMENTS edge that #6374's metric scores as success


cajasmota:
Split out of #6374 after grounding it against `main` @ `640ea7784`. #6374 remains the parent for the Express and no-handler-ref components.

## The defect

`internal/engine/http_endpoint_synthesis.go:1757` hands the route **path** to `emit` as the handler reference:

```go
emit("ANY", canonical, "django", "Route", raw)   // raw == e.Name == the route PATH
```

`refKind="Route"` fails `synthesisHandlerStructuralRef` (`:240-256`), so the #4319 synthesis-time bridge never fires. The endpoint is stamped `source_handler = "Route:<own path>"` and falls through to phase-2 resolution.

There, the handler index (`internal/engine/http_endpoint_resolve.go:302-306`) excludes only the three `http_endpoint*` kinds — **`Route` entities are indexed** — so the name-based fallback at `:690` looks up `{"Route", "<path>", urls.py}`, finds the very Route entity the synthesizer was iterating, and emits:

```
Route:/foo --IMPLEMENTS--> http_endpoint_definition:http:ANY:/foo
```

The graph now asserts that the route `/foo` is implemented by the route `/foo`.

## This is #6429, unported

`e15170110` deleted exactly this shape for Java Spring; its commit body reads *"three bogus `IMPLEMENTS Route:/users -> http_endpoint_definition` self-edges are gone"* and states explicitly *"Out of scope and untouched: express/django's share of #6374."* Django was left on the old shape.

## Why it is worse than a missing edge

Roughly **187 Django endpoints** currently satisfy #6374's "has an IMPLEMENTS edge" metric via this self-edge. Consumers asking *which handler serves this route* get the route's own path back — a confidently wrong answer, which is worse for an agent reading the graph than a visibly absent edge.

It also means **#6374's headline number understates Django, and no count-based before/after can show this fix**: the bogus edge scores as a success today and the correct edge scores as a success tomorrow.

## Fix

Port the #6429 approach to `django_routes.go` and `http_endpoint_synthesis.go:1757` — stamp `handler_method` / `handler_class` on the Route entity at extraction, and emit `("Controller", <method>)` rather than `("Route", <path>)`.

[...]

PR #6496, opened 2026-08-21
docs: prepare the 0.3.0 CHANGELOG section, correcting four claims that no longer matched the code

cajasmota:
Refs #6456

Prepares the `0.3.0` CHANGELOG section. **Not a tag, and deliberately undated** — the heading carries a `<YYYY-MM-DD, fill on tag>` placeholder.

`## [Unreleased]` already held an earlier revision of this body, so most of this is supersession rather than fresh text. What actually changed is the part that was **wrong**.

## Four claims in the shipped text that no longer matched the code

1. **The VB.NET known-limitation bullet still said inheritance "does not resolve."** #6337 landed this cycle (`8adfe8028`, PR #6473): hierarchy resolution **58.3% → 96.6%** on the 302-file corpus. Rewritten from a limitation into a fix, with the residual 17 stated (16 `IFrameServer.*` resolver-routing misses, 1 unlisted BCL type) and the VB-only gating kept explicit.
2. **#6454's forward reference was false.** The text said its ~2,205 new dangling `CALLS` edges "dangle until #6337". The #6337 commit says the opposite — that arm is gated to `EXTENDS`/`IMPLEMENTS` and never touched them. Corrected here. **The same wrong sentence also sits in `baseline.json`**, which this PR cannot reach; filed as #6495, because whoever re-measures the stale baseline (#6490) will read it and reasonably conclude those edges should now resolve.
3. **#6314 read as a complete fix and is still OPEN.** Retitled `(#6314, partial)` — only the `direction=uses` edge-kind arm shipped. The bm25 ranking half is untouched. @manuel1358000's credit is kept, scoped to the part that is real.
4. **#6461 read as fixed; it is pinned, not fixed.** `810abc6db` added the parity test only. Stated as such.

## Newly changelogged — shipped but invisible until now

- **#6337** — above.
- **#6471** — Django DRF routes pointed at their own path, *and resolved*, so no count-based metric could see it.
- **#6416** — a FIFO with an indexed extension hung an indexing worker forever; an unprivileged local `mkfifo` was enough.
- **#6369 F# arm** — F# emitted its import placeholder with no subtype, so it never received the fix its own reproduction had motivated.

Plus a new `### Internal` section, modelled on 0.1.8.2's, covering the fixture/gate/harness work: #6199, #6332, #6347, #6352, #6361, #6373, #6387, #6424, #6453, #6461, #6464, #6476.

## The five @marcus555 bullets are deliberately left in `[Unreleased]`

`#5834, #5835, #5836, #5838, #5840` are **not** in `git log v0.2.2..main`. They were changelogged on 2026-07-21 (`7ef332237`), an ancestor of v0.2.2 — so they shipped in a 0.2.x tag that never received its own section. No section exists for `0.1.8.3`, `0.1.9`, `0.2.0`, `0.2.1` or `0.2.2`.

Carrying them into 0.3.0 would credit them to a release they did not ship in; deleting them would silently erase contributor credit. Neither is acceptable, so they stay where they are pending a backfill of the six missing sections. His in-range credits (#6320, #6318, #6319) are retained in 0.3.0.

## Verification

`git diff` touches only `CHANGELOG.md` (26 insertions, 4 deletions). Issue-reference sets for `[Unreleased]` and `[0.3.0]` were computed and intersect empty — no bullet appears twice. `grep -c '<!--'` over the whole file → 0, so no reviewer scaffolding survived. `[Unreleased]` remains well-formed with the five @marcus555 bullets in original order and wording. No date anywhere in the 0.3.0 section.

## Round 2 — a fifth false claim, found by the fact-check

An independent review caught one more, and it is the sharpest of the set: the `### Internal` bullet asserted that #6464 was **not** fixed and stayed open — written on top of `aee3a7453`, this branch's own parent, which fixed and closed it. The consequence was that `aee3a7453` became the only in-range change with **no** bullet, because the sentence that should have described it said the opposite.

Corrected, and three open issues that read as complete fixes are now labelled:

- **#6367** → `(#6367, partial)` — the HCL arm alone; graphql, proto, bicep, dockerfile and swift are unchanged.
- **#6370** → `(#6370, partial)` — the Groovy arm alone; eight further languages remain.
- **#6373** → partial; the flaky-PASS shape is fixed, the deadline that flaky-*fails* under load is not.
- **#6419** relabelled as a PR rather than an issue.

All 71 in-range commits now map to a bullet.

**Deliberately not changed:** the #6337 figures. `8adfe8028`'s *title* carries a stale `58.0% -> 96.4%`; its *body* measures `288/494 -> 477/494` = `58.3% -> 96.6%`, and the residual-17 breakdown matches verbatim. The changelog has the correct numbers — matching them to the commit title would introduce an error.


cajasmota:
## Independent fact-check: REQUEST-CHANGES

Round 2 dispatched. One blocking finding, three that should land in the same pass.

### Blocking — the `### Internal` bullet says #6464 is unfixed; its own parent commit fixed it

CHANGELOG.md:116 reads *"…the grader's underlying fallback, where a narrowed lookup silently falls back to an unnarrowed one, is **not** fixed and #6464 stays open."*

`52130f0c5`'s parent is `aee3a7453` — `fix(#6464): a row naming a file could be answered by a same-name entity in a different file (#6491)` — whose body states *"Three call sites, one rule: when the row specifies a file, the file-narrowed lookup is the ONLY answer accepted"* (`resolveEntity :225-230`, `from_file :284-289`, `to_file :303-307`). That is precisely the fallback the bullet denies was fixed. `gh issue view 6464` → `CLOSED / COMPLETED`, closed `2026-08-21T08:21:29Z`.

This is not inherited text — the whole `### Internal` section is among this PR's 26 added lines, written on top of the commit that falsifies it.

**Consequence:** `aee3a7453` is the one in-range change with **no bullet**, and the sentence that should describe it asserts the opposite. It is a real behavioural change to the fixture grader, including `TestKindBlankScanDoesNotRescueAWrongFile`.

That makes five false claims found in this text across two passes, all the same shape: a statement that was true when written and was falsified by work that landed afterwards. Worth noting that the mechanism catching them is an independent read against `git log` and `gh issue view` — not anything structural.

### Should fix — three open issues presented as complete fixes

The file already handles this correctly for #6314, #6360 and #6461.

- **#6367** — OPEN. `0e31e57ee` is *"the hcl arm only"*; graphql, proto, bicep, dockerfile and swift remain. The issue covers seven sites across six extractors.
- **#6370** — OPEN. `74cbc42d3` is *"groovy only"*; the other eight languages in the audit remain.
- **#6373** — OPEN. `745fd02ff` fixed the flaky-PASS shape; the "deadline flaky-fails under load" shape is still on the ticket.

### Minor

The Internal bullet at ~:115 cites "#6359, #6419 and #6422" as defects; **#6419 is a merged PR, not an issue.** It mirrors #6453's own title, so low severity.

### Explicitly NOT to be changed

The #6337 figures. The changelog's `58.3% → 96.6%` is **correct** — `8adfe8028`'s body measures `288/494` → `477/494` and the residual-17 breakdown (16 `IFrameServer.*`, 1 `BooleanConverter`) matches verbatim. The commit *title* carries the stale `58.0% → 96.4%`. Anyone "correcting" the changelog to match the title would be introducing an error.

### Verified clean

All 78 cited numbers checked against issue state; every issue described as fixed without qualification is CLOSED except the three above. All 71 commits in `v0.2.2..main` map to a bullet, and nothing credited to 0.3.0 is an ancestor of `v0.2.2`.

Numbers spot-checked against commit bodies and matching: #6454 (`5059→8044`, `3386→4166`, bug rate `0.0918→0.1666`, 2,994 added / 0 removed / 0 retargeted, sample 120, CI 1.9–10.6%, 66.9→51.8), #6412 (`1,046→1,012`, 88 Designer files), #6329/#6348 (`11,740 / 8,544 → 8,506`, ~0.1% vs 47.6%), #6423 (19/40→40/40), #6426 (1,212 / 78.5%), #6445 (2.1×), #6366 (15 of 15).

The #6454→#6337 forward-reference correction is right: `8adfe8028` gates `vbnetHierarchyExternal` to `EXTENDS`/`IMPLEMENTS` and never reaches a `CALLS` target.

**The @marcus555 attribution claim is TRUE.** `#5834/5835/5836/5838/5840` return zero commits in `v0.2.2..main`, and `git merge-base --is-ancestor 7ef332237 v0.2.2` confirms the 2026-07-21 changelogging predates the tag. The five bullets are untouched by the diff, original order and wording. Leaving them in `[Unreleased]` is correct — moving them into 0.3.0 would misfile contributor credit.

No date in the 0.3.0 section; no scaffolding; `[Unreleased]`/`[0.3.0]` issue sets intersect empty; heading depth, separator and heading style match the file's existing convention.

Issue #6500, opened 2026-08-21
JS/TS enclosing-function spans have no end bound, so a call site attributes to the nearest preceding declaration rather than the one containing it


cajasmota:
`jsFuncSpan` carries `offset` and `name` and nothing else, and `enclosingJSFuncAt`
is a nearest-preceding-offset walk with **no end bound**. A call site therefore
attributes to the last declaration that *starts* before it, whether or not that
declaration's body actually contains it.

This is the root cause of the mis-attribution half of #6447, which that arm
**masked rather than fixed**. In `internal/quality/golden/angular-http-mini/src/app/dynamic.service.ts`:

```ts
const base = (code: string) => BASE + '/things/' + code;        // :8
const legacyBase = (code: string) => BASE + '/legacy/' + code;  // :9

@Injectable({ providedIn: 'root' })
export class DynamicThingService {
  byCode(code: string) { return this.http.get<Thing>(base(code)); }        // was attributed to legacyBase
  legacyByCode(code: string) { return this.http.get<Thing>(legacyBase(code)); }
}
```

Before #6447 both call sites were stamped `legacyBase` — the arrow at :9 — because
the walk had no reason to leave it. #6447 indexed the class-method spans, so the
nearest preceding declaration is now `byCode` / `legacyByCode` and the answer is
right. But it is right *because the method header happens to sit nearer to the
call*, not because the helper's span ever ends. Every shape where the true
enclosing function has no declaration between it and the call still mis-attributes:

- a call in a file's trailing top-level code, after the last declaration;
- a call inside a class whose methods are all arrow properties assigned in a
  constructor (no shape 1–4 header between the helper and the call);
- a call in a nested block that follows a sibling function declaration.

## Why this was not fixed in #6447

Adding an `end` to `jsFuncSpan` is not a local change:

- `pyFuncSpan` is a **type alias** of `jsFuncSpan` (`http_endpoint_client_synthesis.go`),
  so any field added is added to the Python spans too, and
  `indexPyEnclosingFunctions` would have to compute a bound by indentation rather
  than by brace matching.
- `enclosingJSFuncAt` has **43 non-test invocations across 20 files** (measured on `main`
  2026-09-02; the earlier "~15 across 6" in this body was wrong by ~3x), several of which
  are not FETCHES at all — `sse_edges.go` (the caller becomes part of a Stream
  entity's **identity**: `channelPath := "/" + caller`), `websocket_edges.go`
[...]

PR #6501, opened 2026-08-21
fix(#6447): class-method spans were unindexed, so 5 Angular HTTP calls had no caller and 2 were stamped with the wrong one

### LINKED ISSUE #6447: FETCHES: enclosingJSFuncAt does not index class-method spans, so 4 of 5 Angular call sites get no caller and the fifth is attributed to the wrong function
`FETCHES` edges form only when `enclosingJSFuncAt` can name the calling function. It does not index **class-method** spans, so an HTTP call made from a class method has no caller to attribute to.

Measured on the new `angular-http-mini` fixture: **4 of 5** Angular call sites emit no `FETCHES` edge at all. The fifth does emit one, and attributes `this.http.get(base(code))` to the module-level arrow function `base` — the helper that builds the URL — rather than to `byCode`, the class method that actually makes the call.

So the failure has two faces: mostly a missing edge, and in at least one shape a **wrong** edge pointing at a plausible-looking but incorrect caller.

## Why this is filed rather than fixed in #6433

That arm's deliverable was emitting the call-site nodes, which is what made the reporter's `SCOPE.ExternalAPI` count zero on the frontend. The nodes now exist. This is the next layer: who *calls* them.

More importantly, the mis-attribution was deliberately **not** written into `angular-http-mini/expected.json`. Encoding an observed-but-wrong result as the expected one is exactly the #6441 defect — a fixture that is green because the bug exists, and that turns red when someone fixes it. So this behaviour is currently ungraded on purpose, and the fixture is honest about it rather than quietly pinning it.

That means a fix here needs to **add** the assertions, not amend them.

## Scope

Teach `enclosingJSFuncAt` to index class-method spans, so a call inside a method attributes to that method. Angular services are the motivating case, but this is not Angular-specific — any TypeScript or JavaScript class making HTTP calls from methods has the same gap, which is most of them.

Worth checking as part of the same arm: whether the arrow-function attribution is a *fallback* that fires when no method span is found, or an ordering problem where the innermost enclosing function wins over the method. Those need different fixes, and the observed `base`-instead-of-`byCode` case does not by itself distinguish them.

The fixture to grade it against already exists, so this arm is cheap to verify — it just needs the assertions it currently, and correctly, declines to make.

cajasmota:
Closes #6447

`jsFuncDeclRe` had three alternatives and an explicit comment declining the fourth — object/class method shorthand — "to avoid colliding with arbitrary function calls". `indexJSEnclosingFunctions` is its only consumer, so an HTTP call made from a class method had no caller to attribute to.

## The title undercounts it: 5 empty, 2 mis-attributed

Measured over `golden/angular-http-mini/src/app/*.ts`, not inferred:

| site | caller before |
|---|---|
| `list`, `create`, `byId`, `rename`, `strings` | `""` |
| `byCode`, `legacyByCode` | **`legacyBase`** |

`inject(HttpClient)` and `constructor(private http: HttpClient)` both fail alternative 3 — `= inject(` has an identifier between `=` and `(` — so four of the five files indexed **zero** spans.

**The mis-attribution:** `dynamic.service.ts` declares `const base = (code) =>` and `const legacyBase = (code) =>` at top level. Spans carry an offset and a name but **no end bound**, so the nearest-preceding walk never leaves `legacyBase` and stamps it on every call site in the class body below — including `byCode`, which actually calls `base(code)`.

**The empty case is silent by construction.** `refName == ""` ⇒ no `source_caller` ⇒ resolve step 3 never runs ⇒ even the broad same-file fallback in `resolveCallerToFetchesEdge` never fires. Zero FETCHES, not a coarse one.

## `legacyBase` is itself an entity — so there were two precise wrong edges

This decided which forbidden rows are falsifiable, and it was settled by dumping the graph (`grafel quality --keep-graph` at `de1d96481`) rather than assumed.

`legacyBase` **is** a `SCOPE.Operation` (`417d81566caa53c1`). So the graph carried two precise wrong edges, not a fallback fan-out:

- `legacyBase → http:GET:/{dynamic}/base_code`
- `legacyBase → http:GET:/{dynamic}/legacyBase_code`

Both are forbidden by name. The second is the subtler one: its *target* is correct and only the source is wrong, so a target-only assertion would never have caught it.

## RED first

7 `FETCHES` rows and 2 forbidden rows added with the fix withheld:

```
EXIT=2
  relationships: 10 / 17 expected  (recall=58.8%)
  forbidden hits: 2
  missing: list, create, byId, rename, strings, byCode, legacyByCode → their call sites
  FORBIDDEN present: legacyBase --[FETCHES]--> http:GET:/{dynamic}/base_code
                     legacyBase --[FETCHES]--> http:GET:/{dynamic}/legacyBase_code
```

After: `EXIT=0`, 17/17, 0 forbidden hits. `expected.json` is **99 insertions, 0 deletions** — the fixture previously contained zero occurrences of `FETCHES`, so this fix would otherwise have landed entirely unobservable.

## Blast radius: this is not FETCHES-only

`indexJSEnclosingFunctions` has 15 non-test call sites. Three are not FETCHES, and all three moved. Measured with a probe indexed by both binaries — no golden fixture exercises these in JS/TS, which is why the benchmark does not churn:

- **SSE — entity identity moves.** Two handlers in one class collapsed into a single file-named node `sse:/stream_ts`; they are now `sse:/streamEvents` and `sse:/streamAlerts`. That is an ID move **and** a per-file undercount fix (same class as #6446).
- **WebSocket:** `Function:sockets_ts → ws:/live | ws:/admin` becomes `SCOPE.Operation:connect | :connectAdmin` — resolving to a real entity instead of a dangling `Function:` placeholder.
- **GraphQL subscriptions:** `Function:subs_ts → graphql_sub:tock` becomes `SCOPE.Operation:watchTicksAgain`. `Subscription` entities key on the subscription name and did not move.

All in the improving direction, so absorbed rather than deferred. 52 sse/websocket/subscription unit tests pass unchanged.

## Mutants — and why the fixture alone is not enough

| | mutation | result | killed by |
|---|---|---|---|
| M1 | delete alternative 4 | KILLED | fixture 10/17 + 2 forbidden hits; 13 unit subtests |
| M2 | drop the required trailing `{` | KILLED | 2 unit subtests |
| M3 | widen `\([^()]*\)` → `\(.*\)` | KILLED | 1 unit subtest |
| M4 | empty the reserved-word reject list | KILLED | 2 unit subtests |
| M5 | `^[ \t]+` → `^[ \t]*` | KILLED | 1 unit subtest |

**The fixture kills only M1.** M2–M5 are the permissive direction and leave it 17/17 green, because the fixture's TypeScript has no `if (...) {` above a call site and no `it('...', function () {`.

**M3 and M5 initially survived everything.** Cases were added to kill them rather than recording them as equivalent. That asymmetry is the entire reason for the unit table: a permissive mutant does not delete an edge, it mints a **wrong caller** — which is worse than the empty one this issue started from.

## Deliberately not fixed: the root cause

Spans still have no end bound. `byCode` now resolves correctly only because the method header sits nearer to the call than the helper does — **the helper's span still never ends.** This fix masks the mis-attribution rather than curing it, and both the commit and the code comment say so.

`jsFuncSpan` is aliased by `pyFuncSpan`, and `enclosingJSFuncAt` is called by ~15 non-JS passes with their own span slices, so bounding spans is a separate change. Filed as **#6500**, which also records that the unbounded half currently has no fixture and is therefore unfalsifiable.

## One judgement call worth pushing back on

Alternative 4 requires a **non-zero** indentation, so a column-0 `foo(a) {` mints no span. That is the M5 killer. The rationale: method shorthand is by construction nested, and column-0 declarations are `function foo(`, which alternative 1 already owns. It is closer to pinning the implementation than the other four cases — the row to challenge if it looks wrong.

## Verification

`go vet ./internal/engine/ ./internal/quality/...` → 0 · `go test -count=1 ./internal/engine/ ./internal/quality/...` → 0 (engine 38.5s) · `go test -count=1 -run 'Quality|Fixture|Golden' ./cmd/grafel/` → 0 · `gofmt -l` clean.

### DIFF
--- a/internal/engine/http_endpoint_client_synthesis.go
+++ b/internal/engine/http_endpoint_client_synthesis.go
@@ -151,10 +151,25 @@ var axiosClientRe = regexp.MustCompile(
 //     `=`, `.`, `(`, `return`, …), a PARAMETER group containing no nested
 //     parens (which rules out `describe('x', () => {`), an optional TS
 //     return annotation, and a following `{`. The `()`-free rule applies to
-//     the parameter group only: the return annotation is `[^;{}]*`, because
-//     a function type in return position carries its own parens
+//     the parameter group only: the return annotation is `[^;{}\n]*`,
+//     because a function type in return position carries its own parens
 //     (`makeLoader(): (id: string) => Promise<void> {`) and excluding them
-//     silently mis-attributed an ordinary TS shape. The residue is the
+//     silently mis-attributed an ordinary TS shape.
+//
+//     The `\n` in that class is load-bearing and is NOT symmetry with the
+//     other exclusions: in Go's regexp `[^;{}]` matches a newline, so an
+//     annotation with no newline brake runs across lines until it finds a
+//     `{` — turning a JSDoc continuation ` * Word (prose):` into a span
+//     named `Word` that swallows the `export function` header below it, and
+//     letting a semicolon-free abstract signature swallow the next method's
+//     body. Measured on this repo's own webui-v2 frontend, the unbraked form
+//     changed the spans of 10 of 233 files and destroyed the caller of a real
+//     exported function. Every shape the annotation legitimately needs to
+//     admit — a function type, a parenthesised union — is single-line, so the
+//     brake costs nothing. Both crossings are pinned in
+//     TestJSMethodShorthandDoesNotSwallowNonDeclarations_6447.
+//
+//     The residue is the
 //     reserved words that are also written `(...) {` — `if`, `for`, `while`,
 //     `switch`, `catch`, … — and those are rejected by name in
 //     indexJSEnclosingFunctions rather than in the pattern, because RE2 has
@@ -164,15 +179,26 @@ var axiosClientRe = regexp.MustCompile(
 //     Unambiguous prefixes are allowed rather than declined: `*name(`,
 //     `async *name(` (generators) and `#name(` (private methods) cannot be
 //     call expressions, so there is no collision to trade against. The `#`
-//     stays IN the captured name — it is part of the identifier, and a class
-//     may declare both `#load` and `load`.
+//     stays IN the captured name, and that is a consistency requirement
+//     rather than a tie-break: handleMethodDefinition in
+//     internal/extractors/javascript/extractor.go:1073 names a method entity
+//     with `x.nodeText(nameNode)`, and for `#load` the name node is a
+//     `private_property_identifier` whose text INCLUDES the `#` — a probe of
+//     `class A { #load() {} load() {} }` emits entities named `#load` and
+//     `load`, not `load` twice. Keeping the
+//     `#` is what makes the `"Function:"+caller` reference resolvable at all;
+//     stripping it would dangle. It also keeps a class that declares both
+//     `#load` and `load` distinguishable.
 //
 // NOTE: a span still carries only (offset, name) — see jsFuncSpan. What makes
 // the attribution correct here is that a method declaration sits NEARER to its
 // call sites than a module-level helper does; the span is still unbounded, so
 // a call site trailing the last declaration in a file still attributes to it.
-// Bounding spans is a separate change — pyFuncSpan aliases jsFuncSpan and ~15
-// non-JS passes share enclosingJSFuncAt — and is filed as #6500.
+// Bounding spans is a separate change and is filed as #6500; its blast radius
+// is why it is separate. pyFuncSpan aliases jsFuncSpan, indexJSEnclosingFunctions
+// has 13 non-test call sites, and enclosingJSFuncAt has 43 — including the C#,
+// Kotlin, Swift, Rust, Ruby, Scala, Go, Java, Dart and PHP client passes, which
+// reuse the walk over their own span lists.
 //
 // Because spans are unbounded, DECLINING to match a shape is not a safe
 // no-op. A call site with no span above it in its own method attributes to the
@@ -181,7 +207,13 @@ var axiosClientRe = regexp.MustCompile(
 // these are places the fix does not reach rather than places it is careful.
 // The known set is pinned in TestJSKnownMisattributionShapes_6447: column-0
 // shorthand, computed keys (`['load']() {`), Allman braces, one-line class
-// bodies, and methods named after a rejected reserved word (`catch`, `return`).
+// bodies, methods named after a rejected reserved word (`catch`, `return`),
+// and the two shapes the type-parameter group `<[^<>()]*>` turns away —
+// a generic with a function-typed default (`load<T = () => void>(x): T {`,
+// rejected for the parens) and a nested generic (`load<Map<string, T>>(x) {`,
+// rejected for the inner angle brackets). Both are ordinary TypeScript, not
+// exotica; widening that group is a separate trade against the
+// `describe('x', () => {` collision and is not attempted here.
 // This is not confined to edge sourcing: sse_edges.go builds a Stream entity's
 // ID as `"/" + caller`, so a wrong caller is a wrongly NAMED entity.
 //
@@ -196,7 +228,7 @@ var jsFuncDeclRe = regexp.MustCompile(
 		`|(?m)(?:^|[^\w$])(?:const|let|var)\s+([A-Za-z_$][\w$]*)\s*=\s*(?:async\s*)?\(` +
 		`|(?m)(?:^|[\s{,;])([A-Za-z_$][\w$]*)\s*=\s*(?:async\s*)?\(` +
 		`|(?m)^[ \t]+(?:(?:public|private|protected|static|readonly|async|get|set|override|abstract)[ \t]+)*` +
-		`(?:\*[ \t]*)?(#?[A-Za-z_$][\w$]*)[ \t]*(?:<[^<>()]*>)?[ \t]*\([^()]*\)[ \t]*(?::[^;{}]*)?\{`,
+		`(?:\*[ \t]*)?(#?[A-Za-z_$][\w$]*)[ \t]*(?:<[^<>()]*>)?[ \t]*\([^()]*\)[ \t]*(?::[^;{}\n]*)?\{`,
 )
 
 // jsMethodShorthandReserved lists the identifiers that satisfy alternative 4
--- a/internal/engine/http_endpoint_client_synthesis.go
+++ b/internal/engine/http_endpoint_client_synthesis.go
@@ -131,10 +131,13 @@ var axiosClientRe = regexp.MustCompile(
 // We intentionally cast a wide net to cover four common shapes:
 //
 //  1. `function foo(` / `async function foo(`
+//
 //  2. `const/let/var foo = (` / `const/let/var foo = async (`
+//
 //  3. Class property arrow: `foo = (` / `foo = async (` (without var/const/let).
 //     This covers React component class methods and service-class patterns
 //     common in Angular/Vue/RN frontends, e.g. `login = (email) => $http.post(...)`.
+//
 //  4. Method shorthand: `foo(...) {` inside an object literal or a class
 //     body — `list(): Observable<T[]> { … }`, `async save(x) { … }`.
 //     This shape used to be declined outright, on the grounds that a bare
@@ -145,14 +148,24 @@ var axiosClientRe = regexp.MustCompile(
 //     at all (#6447). The alternative below keeps the collision closed by
 //     demanding four things at once that a call expression does not have:
 //     line-start plus indentation (a call in an expression is preceded by
-//     `=`, `.`, `(`, `return`, …), a parenthesis group containing no nested
+//     `=`, `.`, `(`, `return`, …), a PARAMETER group containing no nested
 //     parens (which rules out `describe('x', () => {`), an optional TS
-//     return annotation, and a following `{`. The residue is the reserved
-//     words that are also written `(...) {` — `if`, `for`, `while`,
+//     return annotation, and a following `{`. The `()`-free rule applies to
+//     the parameter group only: the return annotation is `[^;{}]*`, because
+//     a function type in return position carries its own parens
+//     (`makeLoader(): (id: string) => Promise<void> {`) and excluding them
+//     silently mis-attributed an ordinary TS shape. The residue is the
+//     reserved words that are also written `(...) {` — `if`, `for`, `while`,
 //     `switch`, `catch`, … — and those are rejected by name in
 //     indexJSEnclosingFunctions rather than in the pattern, because RE2 has
 //     no negative lookahead and spelling them into the alternation would
 //     bury the shape being matched.
+//
+//     Unambiguous prefixes are allowed rather than declined: `*name(`,
+//     `async *name(` (generators) and `#name(` (private methods) cannot be
+//     call expressions, so there is no collision to trade against. The `#`
+//     stays IN the captured name — it is part of the identifier, and a class
+//     may declare both `#load` and `load`.
 //
 // NOTE: a span still carries only (offset, name) — see jsFuncSpan. What makes
 // the attribution correct here is that a method declaration sits NEARER to its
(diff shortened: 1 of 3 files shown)


cajasmota:
## Independent review: REQUEST-CHANGES — the strict side of the predicate violates the PR's own standard

Every headline number reproduced exactly, and the `legacyBase` step was verified properly: it **is** `SCOPE.Operation 417d81566caa53c1`, both pre-fix wrong edges were precise rather than fan-out, and both forbidden rows fire in RED and clear in GREEN. RED `REL 10/17 ENT 21/21 FORBIDDEN 2`, GREEN `17/17, 0`. `#6500` exists and says what the comment claims.

### F1 (HIGH) — the column-0 rationale is factually wrong about its consequence

The PR presents *"a column-0 `foo(a) {` mints no span"* as the safe outcome. **Because spans are unbounded — the very thing #6500 records — a missing span does not yield an empty caller. It yields the previous declaration's name.**

Probed through `indexJSEnclosingFunctions` + `enclosingJSFuncAt` on post-fix code:

```
col0-class-method       caller="build"   want "load"
col0-object-literal     caller="build"
generator-method        caller="build"   # *stream() { fetch(…) }
async-generator         caller="build"
private-method          caller="build"   # #load() {
computed-method         caller="build"   # ['load']() {
method-named-catch      caller="then"    # reject-list eats a real method name
iterator-return-method  caller="next"    # return(v) { — iterator protocol
allman-brace-method     caller="build"
minified-class          caller="build"
fn-type-return-annot    caller="build"   # makeLoader(): (id: string) => Promise<void> {
paren-in-annot-union    caller="build"   # pick(): (A | B) {
```

Every one is a **wrong caller** — the `legacyBase` defect this PR exists to fix, reproduced by the fix's own strictness.

**Verdict on the column-0 rule you asked for pushback on: the rule is defensible, the rationale is not.** RE2 genuinely cannot distinguish `foo(a) {` at column 0 from other constructs. But the honest statement is *"column-0 shorthand falls back to the preceding span's name — wrong, but no worse than before"*, not *"mints no span"*. Same correction applies to the reject-list note: a rejected legitimate method name like `catch` or `return` also mis-attributes rather than going empty.

### F2 (MEDIUM) — a new poisoning source

```
comment-poison-after-method   post-fix caller="ghost"  spans=[load ghost]
                              pre-fix  caller=""       spans=[]
```

A block-commented-out method inside a live method body now mints a span and steals attribution. Same for method shorthand inside a template literal. Alternatives 1-3 have this hazard in kind, but **commented-out method bodies are far more common than commented-out `function foo(`**, so alternative 4 widens the surface materially. No test covers it.

### F3 (MEDIUM) — the identity move is broader than enumerated

`sse_edges.go:155` builds the Stream ID as `"/" + caller`, so F1 and F2 do not merely mis-source an edge — they **mint a wrongly-named entity** (`sse:/build`, `sse:/ghost`) where the old code minted an honest file-named one.

And "three consumers, all three moved" understates it. `indexJSEnclosingFunctions` has 13 non-test call sites; unenumerated JS/TS consumers inheriting the change include `websocket_edges.go:831/848` (socket.io client — a **third** WS site, not the two named), `http_endpoint_synthesis.go:1226`, `http_endpoint_client_synthesis.go:1313/1326/1337/1633/1643`, and `http_endpoint_client_ast.go:92`. All move in the same direction, but none of it is fixture-graded.

### Surviving mutants — non-equivalence disproven for each

| # | mutation | suite | fixture | why non-equivalent |
|---|---|---|---|---|
| N1 | drop `"with"` from the reject list | exit 0 | 17/17 | `  with (o) {` is valid sloppy-mode JS and matches alt 4 → mints span `with` |
| N2 | drop `"function"` from the reject list | exit 0 | 17/17 | `  function (x) {` (anonymous fn expression at line start) matches alt 4; alt 1 requires a name |
| N3 | `(?m)^[ \t]+` → `(?m)[ \t]+` | exit 0 | 17/17 | one-line class bodies `class A { load() { … } }` currently yield no span; matching changes with the anchor dropped |
| **N4** | `(?::[^;{}()]*)?` → `(?::[^;{}]*)?` | exit 0 | 17/17 | **a strict improvement** — fixes `makeLoader(): (id: string) => Promise<void> {` and `pick(): (A \| B) {`, both currently mis-attributed |

**N4 is the important one:** a surviving mutant that makes the code *better* means the annotation class is both under-tested and under-permissive on ordinary TypeScript.

### Required

1. Correct the column-0 comment and the M5 rationale — the outcome is the preceding span's name, not "no span". Same for the reject-list note.
2. Apply N4 and add the `makeLoader(): (id: string) => Promise<void> {` case; it is a live survivor against the full suite.
3. Allow `*`, `async *`, `#` and `[computed]` prefixes, or state in the comment that they are known mis-attribution shapes.
4. Reject-list coverage for `with` and `function`.
5. A comment/template-literal poisoning test, either accepted explicitly in the comment or gated.

The core is right and the evidence discipline is above this repo's average. But the PR's own standard — *a wrong caller is worse than an empty one* — is violated by the strict side of the same predicate, and it ships wrong-caller **entity IDs** for shapes that are ordinary TypeScript.

cajasmota:
## Round 2 review: REQUEST-CHANGES — and the review is retracting its own round-1 recommendation

**I called N4 "a strict improvement" in the round-1 review, and I relayed it to the implementer without verifying it. Both were wrong.** Applying it unmodified introduces a new defect class, now measured.

### First, the good news, verified rather than assumed

**The committed source is clean of the mutation-race residue.** `git diff HEAD~1 HEAD` on `http_endpoint_client_synthesis.go`, filtered to non-comment lines, is **exactly one changed line** — the alternative-4 regex. `jsMethodShorthandReserved` is byte-identical to the parent. No M5 residue, no stray edit.

All twelve mutants re-scored independently, sequentially, never backgrounded, shasum verified before and after each: **N1-N4 and M6-M8 all KILLED**, plus five of the reviewer's own (P4, P7, P9, P10, P11). The known-misattribution rows are exact-equality assertions that genuinely flip under N3 and P11 — real tripwires, not decoration.

### BLOCKING — N4 regresses 4% of this repo's own frontend

Root cause: **`[^;{}]` matches newlines in Go regexp.** Combined with round 2's new `(?:\*[ \t]*)?` generator prefix — which also matches a **JSDoc continuation `*`** — an indented `name(...)` followed by `:` swallows text across lines until the next `{`.

Measured with an exact replica of `indexJSEnclosingFunctions` over `webui-v2/src`, HEAD vs parent:

```
files=248 differing=10
```

`testdata/` (120 files) and `internal/quality/` (17) are unchanged — which is precisely why the suite stayed green. The damage is entirely outside the fixtures.

**The worst case destroys a real span.** `webui-v2/src/lib/coverage-provenance.ts` has a JSDoc line `* Precedence (best available wins, never a misleading authoritative %):` above `export function resolveCoverageProvenance(`:

```
old=[parseTime resolveCoverageProvenance resolveCoverageKindIndicator …]
new=[parseTime Precedence                resolveCoverageKindIndicator …]
```

The annotation runs from `:` through `*/`, the function header and its multi-line parameter list, to the body `{`; `FindAll` resumes past the match, so alternative 1 never sees the real name. **Every call in that exported function is attributed to a caller named `Precedence`** — and via `sse_edges.go:155` that is a Stream entity named `/Precedence`. Nine more files gain phantom callers from prose: `Limitation`, `Decoupling`, `Usage`, `Responsibilities`, `BUG`, `Payload`, `Layout`, `Sections`, `PH5`.

A second shape needs no JSDoc — in `semi: false` TypeScript an abstract member or overload signature swallows the next method:

```
abstract load(id: string): Promise<void>
save(x) { fetch('/api/save') }
parent: caller="save"   ← correct
HEAD:   caller="load"   ← wrong, and `save` has no span
```

**One-character fix, verified:** `(?::[^;{}]*)?` → `(?::[^;{}\n]*)?`. Suite stays green — `makeLoader(): (id: string) => Promise<void> {` and `pick(): (A | B) {` are both **single-line**, which is the only shape N4 ever needed. The `webui-v2` delta drops from **10 to 0**.

### Two surviving mutants, on that same untested axis

| # | mutation | proof of non-equivalence |
|---|---|---|
| P6 | `(?::[^;{}]*)?` → `(?::[^;]*)?` | on semicolon-free `class A { cfg(): void {…} load() { fetch() } }`, caller changes `load` → `cfg` and `load` mints no span |
| P12 | `(?::[^;{}]*)?` → `(?::[^;{}\n]*)?` | it **is** the fix — nothing tested the brake |

Nothing tests that `{`/`}` is the only thing stopping the annotation crossing lines. Two cross-line rows close both.

### Smaller items

- **F6's count is wrong and observable.** `indexJSEnclosingFunctions` has **13** non-test call sites, not 14; `enclosingJSFuncAt` has **43**, not "~30". Shape right, numbers wrong.
- **Two ordinary-TypeScript shapes are missing from the "known misattribution" set** — generic-with-default `load<T = () => void>(x): T {` and nested generic `load<Map<string, T>>(x) {`, both blocked by `<[^<>()]*>` and both silently inheriting the preceding method.
- **The `#`-in-capture choice is right for a better reason than the one given.** `javascript/extractor.go:1071` does `name := x.nodeText(nameNode)`, and for `#load` the name node is `private_property_identifier` **whose text includes the `#`** — so keeping it is what makes `"Function:"+caller` resolvable at all. That turns an arbitrary-looking tie-break into a consistency requirement.
- **The `describe('x', () => {` exclusion holds** — verified by execution, not reading: direct probe returns `caller=""`, and loosening the parameter group to `[^)]*` is killed by the collision test.

### The lesson worth keeping

Round 1 rejected this PR for a rationale that was wrong about its own consequence. Round 2's blocking finding is a **recommendation from that same review** that was wrong for the same reason — asserted from reading rather than execution, and only caught by measuring against a real corpus. The graded fixture could not see it, because the affected files are all outside `testdata/`.

cajasmota:
## Coordinator mutant on round 3 — SURVIVED. Reporting the direction as UNKNOWN, deliberately.

**Mutation:** widen the parameter group to admit one level of nesting —

```
\([^()]*\)  →  \((?:[^()]|\([^()]*\))*\)
```

`go vet ./internal/engine/` exit 0 **before** scoring · target count asserted 1 · `GOMAXPROCS=4 go test -count=1 ./internal/engine/` → **exit 0, SURVIVED** · restored by `cp`, shasum `0b87140f…` verified identical, tree clean.

This is a realistic drift: it is exactly what a "support default arguments like `foo(x = f())`" change would look like, and nothing in the suite objects.

### Why I am not calling it an improvement

Reasoning it through, the `describe('x', () => {` exclusion still appears to hold — the group would need a closing `)` before `{`, and `('x', () => {` never provides one. And admitting `foo(x = f())` would fix a shape that is currently mis-attributed. So the tempting conclusion is that this is another strict improvement.

**That is precisely the reasoning that made N4 a blocking regression.** N4 was called a strict improvement in round 1, relayed by me unverified, and turned out to change 10 of 233 real frontend files — including one where a real span was destroyed and an entity was renamed to `/Precedence`. The fixture suite was green throughout, because the damage lived entirely outside `testdata/`.

So the direction of this mutant is **unknown until measured**, and the measurement is the one round 3 already built: the `indexJSEnclosingFunctions` replica over `webui-v2/src`, parent vs mutant, reporting `files=N differing=M`.

### What this asks for

Not necessarily a code change — possibly just a test. Three outcomes:

- **differing = 0** → equivalent on the corpus; add a row pinning the parameter group's nesting behaviour so it stops being an untested axis, and record that nesting was considered and declined.
- **differing > 0 in the improving direction** → a separate, measured change with its own rows; not something to smuggle into this PR.
- **differing > 0 with any destroyed span** → the current narrow group is correct and should be pinned as such.

The parameter group is currently pinned only by P7 (`[^()]*` → `[^)]*`, killed by the collision test), which tests a *looser* failure. One-level nesting sits between the two and nothing observes it.

Handing this to the reviewer rather than acting on it — after N4 I would rather have the measurement than my own reasoning.

PR #6504, opened 2026-08-21
fix(#6483): a skipped file reached only stderr, so an MCP or dashboard caller was told a repo was clean

### LINKED ISSUE #6483: ScanPath reports a skipped file only to stderr, so an MCP or dashboard caller is told a repo is clean when a file was never read
Split out of #6416 (PR #6468), where it was identified and **deliberately not built** so the fix would not widen into an API change mid-round. The cost is written into `docs/blocking-open-audit.md` and into `reportSecretScanSkip`'s doc comment.

## What is already fixed

`internal/secrets.scanFile` no longer skips an irregular file silently. A FIFO named `creds.go` now produces:

```
grafel: skipped secret scan of safeio: not a regular file: <path> (named-pipe) — not read because reading one can block forever; this file was NOT checked for secrets (#6416)
```

always-on, deduped by path, capped, gated to `ErrNotRegular`/`ErrWouldBlock` only.

## What is still missing

That line goes to **stderr**. In the daemon that is the daemon log, so a record exists — this is no longer the #6338 "nowhere" case. But `ScanPath` returns `([]Finding, error)`, and the skip is not in the result.

So a caller reaching the scanner through the **MCP tool** (`internal/mcp/secrets_tools.go:81`) or the **dashboard HTTP handler** (`internal/dashboard/handlers_secrets.go:103`) asks *"is this repo clean?"* and gets an **unqualified yes**, while a file was skipped and never examined. The one channel that reports it is the one channel that caller is not reading.

For a secrets scanner specifically, "I did not check this file" and "I checked this file and it was clean" are not interchangeable answers, and the API currently cannot distinguish them.

## Cost, so the trade-off is visible

`ScanPath`'s signature changes — either an added `Report` return or a result struct — at both entry points, plus their JSON payloads and the tests covering them. That is why it was not folded into #6468: that PR had already been extended twice mid-round and a third widening would have made it unreviewable.

## Worth deciding rather than assuming

Whether the same applies to `internal/install/detect`. Its three manifest parsers now report their skips the same way, and a skipped `package.json` means **losing every workspace package** it would have declared — arguably a worse silent outcome than a missed secret, since it changes what gets indexed at all rather than what gets flagged. But `detect` is called during install/setup where stderr is genuinely in front of the user, so the case is weaker. Check the call sites before deciding.

## Acceptance

A caller of `ScanPath` can distinguish "clean" from "clean, and N files were not read", and the MCP and dashboard responses carry that distinction. A test asserting the MCP payload reflects a skipped file — because the whole defect is that the information exists and does not reach the caller, so a test at the package boundary would not observe the fix.

cajasmota:
Closes #6483

`ScanPath` returned `([]Finding, error)` and reported skipped files **only to stderr**. None of its three callers surfaces that, and for the two daemon callers stderr *is* the daemon log, which the client never reads. So an MCP or dashboard caller was told a repo was clean when a file was never opened.

## Three skips, not one — and the worst was reported nowhere at all

| skip | before | now |
|---|---|---|
| irregular file (FIFO, device, socket) | stderr only | `not_regular`, with `Kind` |
| open blocked | stderr only | `would_block` |
| **over `max_size`** | **nowhere** | `too_large` |
| **line over 64 KB** | **nowhere, findings discarded** | `line_too_long`, findings **kept** |

**The large-file gate matters because the dashboard exposes `max_size` as a query parameter** — a caller could pass `max_size=1024` and receive an unqualified "clean" for a repo where nearly every file was skipped.

**And `bufio.ErrTooLong` was reporting a real credential as clean.** `scanFile` returned `(findings, scanner.Err())` and the walk dropped `ff` wholesale:

```
findings=0 skipped=[]   // AKIA…REAL0000 on line 1 + one 64KB+ line
findings=1              // identical key, no long line
```

`bufio.MaxScanTokenSize` is 64 KB against a 512 KB cap, and the extension denylist excludes no `.json`, `.lock`, `.map` or minified `.js` — the files that carry long lines are exactly the ones on this path. More reachable than the FIFO case the issue was filed for. Tracked separately as #6505.

`line_too_long` is a distinct reason rather than a `Kind` on `too_large` because they are different facts with different remedies: `too_large` means never opened, under a cap **the caller chose** and can raise; `line_too_long` means opened and partly read, stopped at a limit **no caller can raise**.

Extension denylist and `isTestFile` stay unreported — intended filters, and reporting them would produce thousands of entries and destroy the signal.

## A struct, not a third return value

`ScanPath(root, max) (ScanResult, error)`. Not `([]Finding, []Skip, error)`: a third return is easy to `_`-discard, **which is the exact failure mode being fixed**. A struct forces every call site to be touched to compile, so the compiler performs the audit.

`Skip` reuses the existing `internal/daemon/walk.SkipEntry` vocabulary. `ErrWouldBlock` is returned **bare with no path**, so `Path` is set from the walk's own context rather than parsed out of the error.

## Partial findings are kept for `ErrTooLong` only — and the line numbers are provably right

Bytes before an overlong line were matched against real bytes; a mid-read I/O error puts the bytes *and the line numbers* in doubt. So `ff` is kept for `ErrTooLong` and still dropped elsewhere.

The obvious risk is a finding that looks trustworthy at the wrong line. Probed directly:

```
body: line 3 = AKIA…REAL000, line 5 = AKIA…REAL111, line 6 = 70 KB, line 7 = AKIA…REAL222
FINDING line=3   FINDING line=5
SKIP reason=line_too_long   nfindings=2 nskipped=1
```

`bufio` delivers every line before the overlong one exactly once and then stops; nothing about the partly-consumed line perturbs the counter. **`Finding.Line` is now asserted** (3 and 5) — previously it was checked nowhere in `internal/secrets` against a real scan, so `lineNum += 2` survived silently.

## Mutants — the ones that prove the layering

| mutation | result | what it proves |
|---|---|---|
| drop `Skipped` from the return, **keep stderr** | DEAD | every pre-#6483 test stayed green under it — the gap the old suite structurally could not see |
| drop `"skipped_files"` from the MCP payload | DEAD at MCP, **`./internal/secrets/` exit 0** | correct information reaching the *caller* is a separate property from the scanner knowing it |
| delete the dashboard `SkippedFiles` loop | DEAD | round 1 shipped without this; the endpoint with the `max_size` param had **no test at all** |
| `lineNum++` → `lineNum += 2` | DEAD | `first finding reported line 6, want 3` |
| delete the `ErrTooLong` arm / drop its partial findings | DEAD | |
| `errors.Is(ErrNotRegular)` → `err != nil` | DEAD | |
| delete the `os.Stat` re-stat in `classifyScanSkip` | DEAD | symlink→FIFO reports `named-pipe`, not `symlink` |
| disable the `ErrWouldBlock` arm | DEAD | |
| delete MCP truncation / total-after-truncation / nil-init | DEAD | |
| collapse `line_too_long` into `too_large` at the classify site | DEAD | the distinction is pinned, not just documented |

16 of 19 mutants killed. Two survivors are equivalent on all reachable paths; one (`SkipLineTooLong`'s literal value) is now pinned with a literal-string assertion, since the four reason strings are a client-facing JSON contract.

## Portability

A FIFO cannot be created on Windows. FIFO tests are `//go:build unix` with a `t.Skipf` guard and create the FIFO **only inside `t.TempDir()`**. The `too_large` and `line_too_long` tests need **no build tag and no syscall**, so Windows CI gets real coverage. The MCP test uses a `scanSecrets` seam; the dashboard test drives real HTTP against a temp `GRAFEL_HOME` at `?max_size=1`.

There was previously **no test at all** for `internal/mcp/secrets_tools.go` or `internal/dashboard/handlers_secrets.go`.

## Also

`omitempty` dropped and `SkippedFiles` initialised to `[]` — on a nil slice, "no `skipped_files` key" reads as "clean" to a JS client doing `if (r.skipped_files?.length)`, reproducing the exact ambiguity the field removes. MCP list capped at 32 with `skipped_files_total` / `skipped_files_truncated`; `total` is captured before the slice, so `total < len(list)` is unrepresentable.

A test doc comment claiming a kill it did not perform was corrected, and `TestScanPathDoesNotReportUnreadableFileAsSkip` now genuinely reaches the `errors.Is` gate. `WalkDir`'s `return nil` documents the EACCES-directory hole as a known deliberate exclusion rather than asserting coverage.

## Filed, not fixed here

- **#6505** — the `ErrTooLong` credential miss, so it stays tracked independently.
- **#6507** — the size gate lstats, so a symlink to an oversized file bypasses `max_size` and now mislabels as `line_too_long`. Pre-existing; only the mislabel is new.
- `internal/install/detect` — same reporting shape, but called where stderr **is** in front of a user, so it needs its own call-site audit.

## Verification

`go build ./...` → 0 · `gofmt -l` empty · `go vet` secrets/dashboard/mcp/cmd → 0 · `go test -count=1 ./internal/secrets/ ./internal/dashboard/ ./internal/mcp/` → 0.

### DIFF
--- a/cmd/grafel/rebuild_history.go
+++ b/cmd/grafel/rebuild_history.go
@@ -146,13 +146,17 @@ func appendRebuildHistory(root, group string, cfg *registry.GroupConfig, rebuilt
 	const maxSecretFileSizeBytes = 256 * 1024
 	totalSecrets := 0
 	for _, repoPath := range rebuiltPaths {
-		findings, err := secrets.ScanPath(repoPath, maxSecretFileSizeBytes)
+		// Only Findings is consumed here (#6483). The quality snapshot is a
+		// single per-run integer and deliberately does not grow a skip count:
+		// scan.Skipped exists for the two interactive callers that answer
+		// "is this repo clean?" to a human, not for the history series.
+		scan, err := secrets.ScanPath(repoPath, maxSecretFileSizeBytes)
 		if err != nil {
 			// Non-fatal: log and skip.
 			fmt.Fprintf(os.Stderr, "grafel: secret scan %s: %v (skipped)\n", repoPath, err)
 			continue
 		}
-		totalSecrets += len(findings)
+		totalSecrets += len(scan.Findings)
 	}
 	entry.Secrets = &totalSecrets
 
--- a/internal/dashboard/handlers_secrets.go
+++ b/internal/dashboard/handlers_secrets.go
@@ -63,7 +63,14 @@ type SecretScanReply struct {
 	// parameter: ?max_size=1024 would otherwise return total_findings:0 for a
 	// repo in which nearly every file was skipped unread, and the caller
 	// would have no way to tell that from a genuinely clean repo.
-	SkippedFiles []SecretSkippedFile `json:"skipped_files,omitempty"`
+	//
+	// Deliberately NOT omitempty, and initialised to a non-nil slice below.
+	// On a nil slice omitempty drops the key, and the natural client check
+	// `if (r.skipped_files?.length)` then reads "no key" as "clean" —
+	// recreating the exact ambiguity this field exists to remove. An
+	// always-present [] says "asked, and the answer is none". Files above
+	// already renders this way; this keeps the two uniform.
+	SkippedFiles []SecretSkippedFile `json:"skipped_files"`
 }
 
 // SecretSkippedFile is one file the secret scan did not read.
@@ -111,8 +118,9 @@ func (s *Server) handleQualitySecrets(w http.ResponseWriter, r *http.Request) {
 	}
 
 	reply := SecretScanReply{
-		Group:      groupName,
-		BySeverity: map[string]int{"critical": 0, "high": 0, "medium": 0, "low": 0},
+		Group:        groupName,
+		BySeverity:   map[string]int{"critical": 0, "high": 0, "medium": 0, "low": 0},
+		SkippedFiles: []SecretSkippedFile{},
 	}
 
 	for _, rp := range repoPaths {
--- a/internal/dashboard/handlers_secrets.go
+++ b/internal/dashboard/handlers_secrets.go
@@ -77,7 +77,8 @@ type SecretScanReply struct {
 type SecretSkippedFile struct {
 	Repo string `json:"repo"`
 	File string `json:"file"`
-	// Reason is secrets.SkipNotRegular / SkipWouldBlock / SkipTooLarge.
+	// Reason is secrets.SkipNotRegular / SkipWouldBlock / SkipTooLarge /
+	// SkipLineTooLong.
 	Reason string `json:"reason"`
 	Kind   string `json:"kind,omitempty"`
 }
--- a/internal/dashboard/handlers_secrets.go
+++ b/internal/dashboard/handlers_secrets.go
@@ -57,6 +57,22 @@ type SecretScanReply struct {
 	Files         []SecretFileRollup `json:"files"`
 	// ScannedRepos is the number of repos that were actually walked.
 	ScannedRepos int `json:"scanned_repos"`
+	// SkippedFiles are files the scan did NOT read (#6483).
+	//
+	// It matters most on THIS endpoint, because max_size is a query
+	// parameter: ?max_size=1024 would otherwise return total_findings:0 for a
+	// repo in which nearly every file was skipped unread, and the caller
+	// would have no way to tell that from a genuinely clean repo.
+	SkippedFiles []SecretSkippedFile `json:"skipped_files,omitempty"`
+}
+
+// SecretSkippedFile is one file the secret scan did not read.
+type SecretSkippedFile struct {
+	Repo string `json:"repo"`
+	File string `json:"file"`
+	// Reason is secrets.SkipNotRegular / SkipWouldBlock / SkipTooLarge.
+	Reason string `json:"reason"`
+	Kind   string `json:"kind,omitempty"`
 }
 
 // ─────────────────────────────────────────────────────────────────────────────
@@ -100,14 +116,23 @@ func (s *Server) handleQualitySecrets(w http.ResponseWriter, r *http.Request) {
 	}
 
 	for _, rp := range repoPaths {
-		findings, sErr := secrets.ScanPath(rp.Path, maxSize)
+		scan, sErr := secrets.ScanPath(rp.Path, maxSize)
 		if sErr != nil {
 			// Non-fatal: continue with remaining repos.
 			continue
 		}
 		reply.ScannedRepos++
 
-		report := secrets.BuildReport(rp.Path, findings)
+		for _, sk := range scan.Skipped {
+			reply.SkippedFiles = append(reply.SkippedFiles, SecretSkippedFile{
+				Repo:   rp.Slug,
+				File:   sk.Rel,
+				Reason: sk.Reason,
+				Kind:   sk.Kind,
+			})
+		}
+
+		report := secrets.BuildReport(rp.Path, scan.Findings)
 		for k, v := range report.BySeverity {
 			reply.BySeverity[k] += v
 		}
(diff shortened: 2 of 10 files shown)


cajasmota:
## Independent review: REQUEST-CHANGES — including a live instance of the bug this PR fixes

Both headline mutants reproduce exactly as claimed, and the layering argument is real: mutant 2 (drop `"skipped_files"` from the MCP payload) leaves `./internal/secrets/` at **exit 0** and fails only at MCP level, reporting `keys present: [by_severity files scanned_repos tip total_findings truncated]`. The boundary test earns its place.

### Blocking 1 — M-A is fatal to the PR's own thesis

Deleting the entire `for _, sk := range scan.Skipped` → `reply.SkippedFiles` loop (`handlers_secrets.go:126-133`) leaves `./internal/dashboard/` at **exit 0**. The JSON body loses `skipped_files` entirely and nothing notices.

**There is no dashboard test at all.** An MCP-level test was added; none for the endpoint this PR's own body calls *"the dangerous one"* — the one where `max_size` is a query parameter. The PR's argument that a package-boundary test would not observe the fix applies verbatim here and was not honoured.

### Blocking 2 — `bufio.ErrTooLong` reports a real secret as clean

`scanFile` returns `(findings, scanner.Err())`; the walk's `if err != nil` branch drops `ff` **wholesale**, and `classifyScanSkip` returns `ok=false`:

```
PROBE2         findings=0 skipped=[]   // AKIAIOSFODNN7REAL0000 on line 1 + one 64KB+ line
PROBE2-control findings=1              // same key, no long line
```

`bufio.MaxScanTokenSize` is 64 KB against a 512 KB cap, and `skipExtSet` denies no `.json`, `.lock`, `.map` or minified `.js` — **the files that carry >64 KB lines are exactly the ones on this path.** More reachable than the FIFO case the issue was filed for, and unmentioned.

Related and also unstated: `WalkDir`'s error callback `return nil // skip unreadable entries`. A `chmod 000` subdirectory containing a real key gives `err=<nil> findings=0 skipped=[]`. The PR's stated exclusion argues about *files*; one EACCES directory is N unread files.

### Blocking 3 — a claim in the tree that nothing observes

`TestScanPathSkipListStaysEmptyOnOrdinaryTrees`'s doc comment says it kills "dropping the `errors.Is` gate so every ENOENT lands in the list". **It does not** — an ordinary tree produces no `scanFile` error at all, so the gate is unobserved in both directions.

Proven by mutant **M-G**: `errors.Is(err, safeio.ErrNotRegular)` → `err != nil` survives, and every ENOENT/EACCES/`ErrTooLong` then becomes a reported `not_regular` skip.

### Other survivors

| # | mutation | non-equivalence |
|---|---|---|
| M-E | delete the `os.Stat` re-stat in `classifyScanSkip` | symlink `creds.go` → FIFO `pipe` gives `Kind:"named-pipe"` with the re-stat, `Kind:"symlink"` without |
| M-F | disable the `ErrWouldBlock` arm | **no test reaches it.** Genuinely reachable in production — `safeio.openWithDeadline` returns it from the semaphore timeout and the 5s open deadline, both on this path — and the `Path`-from-walk logic, the one behaviour the issue explicitly flagged, is asserted only by comment |

### The JSON asymmetry — ruled: make it uniform, in the MCP direction

The MCP side is right. For a scanner, an absent key is indistinguishable from an old server, and `[]` versus `[…]` is the entire distinction the field exists to carry. The dashboard's `omitempty` on a nil slice reproduces exactly the ambiguity being removed: **"no `skipped_files` key" reads as "clean"** to a JS client doing `if (r.skipped_files?.length)`. One line, and it takes a judgement call off a security surface.

Minor: the MCP `skipped_files` list is **uncapped** while `files` is truncated with a `truncated` flag — one repo with many oversized `package-lock.json`/`.map` files inflates the payload unboundedly with no signal.

### Verified correct

`SkipTooLarge`'s `>` is exclusive and pinned; `maxFileBytes <= 0` defaults to 512 KB before the walk and the dashboard clamps `v > 0`, so "every file too large" is unreachable. The `scanSecrets` seam has one assignment outside tests and no current test parallelises against it — safe today, though nothing prevents a future race. The rewritten `CHANNEL NOTE` is accurate assertion by assertion.

Round 2 is dispatched for the three blocking items plus the recommended coverage.

Issue #6450, opened 2026-08-20
links: delete the link-pass fold copy — it folds nothing at ANY group size, and nothing grades it today (decision made; measurement must come first)


cajasmota:
**Body rewritten 2026-09-02. Task 1 shipped; this issue is Task 2 only.** The original body argued Task 1 — that the engine's intra-repo matcher never sees a folded URL — and that is **no longer true**. The original text is preserved in the comment thread below.

## Task 1, for the record: fixed by `74ceab65a` (#6513)

It adds `internal/engine/http_endpoint_substrate_fold.go` (`FoldConsumerHTTPBaseURLs`), which builds a repo-scoped substrate resolver **at index time** and rewrites `path` / `url_kind` / `substrate_*` **before** `resolveHTTPEndpointHandlers`. It honours the identity constraint — only `path` is touched, `Name` and `EntityID` frozen (`:22-30`). The user-visible win, `FETCHES` instead of `UNRESOLVED_FETCH` in `graph.json` for the monorepo shape, has landed.

*(Read from the diff and the resulting file, not from the commit subject — a commit naming an issue has twice this week turned out not to close it.)*

## What is still live

The link-pass copy of the fold was never removed, so both halves the original review said to **decide between** are still present and still undecided:

- `internal/links/links.go:354` still calls `applyResolverToConsumerHTTP(graphs, resolver)`
- `internal/links/constant_propagation.go:450` still defines it; `:437-449` still documents in-memory-only mutation
- `internal/links/http_pass.go:973` still has `if len(graphs) < 2 {` — the bail is unchanged

So for a **multi-repo group** the fold now runs **twice** — once at index time, once in the link pass, the latter discarded. For a **single-repo group** the link-pass copy is dead work, and the `http_self` matcher in `runHTTPPass` still does not run at all.

## Blast radius: low, and that is the point

Wasted work and a redundant path. **Nothing silently wrong that a user can observe.** `substrate_resolved_value` still has zero readers (not re-verified in the latest pass). This should **not** outrank live silent-wrongness items on the board — it is filed to be correct about what remains, not to be picked up next.

## A decision is required BEFORE dispatch — do not let an agent choose

The two remedies are **mutually exclusive**:

- **Delete the call** — remove `applyResolverToConsumerHTTP` from the link pass, accepting that the link pass no longer folds at all and relying entirely on the index-time fold.
[...]

PR #6513, opened 2026-08-21
fix(#6450): the base-URL fold ran in the link pass, so the index-time matcher never saw a folded path

cajasmota:
Refs #6450 (Task 1 of 2 — `applyResolverToConsumerHTTP` and the `len(graphs) < 2` bail are untouched)

The substrate constant fold ran only in `grafel group-link`. The per-repo call→definition matcher runs at index time and joins on the `path` property, so the two passes structurally never met: `fetch(`${BASE}/things`)` was written to `graph.json` as `UNRESOLVED_FETCH` **even when the handler sat in the same tree**. No gate was involved.

New `internal/engine/http_endpoint_substrate_fold.go` runs the fold at index time, before `ResolveHTTPEndpointHandlers`. Folding is purely intra-repo — the links `Resolver`'s symbol table, imports index and file lookup are all repo-keyed — so nothing is lost by doing it per repo, and `internal/substrate` imports no grafel package, so `engine → substrate` adds no cycle.

## The issue's repro shape was incomplete, in a way that decides the fixture

The probe corrected it before any code changed. `` `${BASE}/things` `` does canonicalize to `/{BASE}/things` with the ident preserved, so the fold is live — but **if `BASE` is declared in the same file as the `fetch`, the JS canonicaliser already folds it at extraction time** (`path=/api/things`, `url_kind=literal`) via a different mechanism entirely, and the defect does not reproduce.

It reproduces only when `BASE` is **imported from another file** — which is exactly the case the substrate `Resolver`'s IMPORTS walk exists for. A same-file fixture here would have been **vacuously green**. `express-baseurl-mini` therefore uses cross-file constants.

## Identity contract held, and measured rather than asserted

Only `path` is rewritten; `Name` and the derived `EntityID` stay frozen at the unfolded canonical form. Diffing every `(kind, name, source_file, id)` tuple with and without the fold: **0 lost, 0 changed.** The only delta is 3 *new* `SCOPE.Process` cross-stack flows that could not previously exist. Folding at extraction time — i.e. changing `Name` — would churn the ID of every dynamic-base-URL call site on every incremental run, which is why the shipped links pass froze them and why that decision is preserved.

## RED

New fixture `internal/quality/golden/express-baseurl-mini`. Before the source change: `entities 8/8`, `relationships 0/5`, `forbidden hits: 3`, exit 1 — 5 missing `FETCHES` plus 3 `UNRESOLVED_FETCH` = **8 independent reds**. After: `8/8`, `5/5`, 0 forbidden, exit 0.

Both endpoints of the `FETCHES` edges resolve by real entity ID; `to_bare_name` was not needed, since the resolver rewrites the `ToID` stub before the graph is written.

## Mutants — and four the golden set provably cannot see

| # | mutant | golden | unit |
|---|---|---|---|
| M1 | fold disabled | 0/5 rel, 3 forbidden — **8 reds** | — |
| M2 | rewrite `Name` too (identity churn) | 5/8 ent, 2/5 rel | RED |
| M3 | drop scheme+host strip | 4/5 rel, 1 forbidden | RED |
| M4 | drop `url_kind` guard | **green** | RED |
| M5 | drop ident charset check | **green** | RED |
| M6 | fold definitions too | **green** | RED |
| M7 | bind ident from any file | **green** | RED |

M4–M7 are **permissive** mutants — the direction this milestone's defects have all come from. The golden cannot see them for a structural reason worth stating: `urlKindFromPath` always classifies a leading placeholder as `dynamic_baseurl`, and no sniffer emits a non-identifier `Binding.Ident`, so on any fixture expressible today the guards are unexercised. Rather than claim them killed, all four are pinned by unit tests in `http_endpoint_substrate_fold_6450_test.go`.

## Two limits documented rather than fixed

1. The sniffed file set is derived from record `SourceFile`s, so a constants module producing zero entities is invisible to the symbol table — inherited verbatim from `internal/links.buildResolver`.
2. The incremental one-file path (`ResolveHTTPEndpointHandlersFileScoped`) does **not** fold: a one-file slice cannot see the declaring module. Those calls keep the unfolded path until the next full index.

## Verification

`go vet` (3 pkgs) → 0 · `gofmt -l` clean · `./internal/engine/` → 0 · `./internal/quality/...` → 0 · `./cmd/grafel/ -run 'Quality|Fixture|Golden'` → 0 · `-run 'BuildDocument|Indexer|HTTPEndpoint|Incremental'` → 0. All `-count=1`, real exit codes. **All 25 goldens graded before and after — `express-baseurl-mini` is the only score that moves.** Sources restored by `cp` with shasums verified after every mutant.

### DIFF
--- a/cmd/grafel/incremental_baseurl_fold_6450_test.go
+++ b/cmd/grafel/incremental_baseurl_fold_6450_test.go
@@ -8,17 +8,35 @@
 // holds only the re-extracted files. The consequence was not a missed
 // improvement but a REGRESSION: touch the caller file with an unrelated edit
 // and the declaring constants module becomes invisible, every folded path
-// reverts to `/{BASE}/…`, and every FETCHES the previous full index resolved
-// turns back into UNRESOLVED_FETCH — and stays that way across further
-// incremental runs until the next full index.
+// reverts to `/{BASE}/…`, and the UNRESOLVED_FETCH edges the incremental run
+// emits go back to naming the unfolded path — and it stays that way across
+// further incremental runs until the next full index.
 //
-// Measured on this fixture before the fix, via the same Path-B pipeline this
-// test drives:
+// Measured on THIS fixture — two calls, two definitions — via the same Path-B
+// pipeline this test drives. Every figure below was run, not carried over from
+// the three-call golden fixture:
 //
-//	                              BEFORE the fix          AFTER the fix
-//	full index                    folded=3 path=/api      folded=3 path=/api
-//	append a comment to api.js    folded=0 path=/{BASE}   folded=3 path=/api
-//	no-change incremental         folded=0 path=/{BASE}   folded=3 path=/api
+//	                            BEFORE the fix        AFTER the fix
+//	full index                  path=/api/things      path=/api/things
+//	                            FETCHES=6 UNRES=0     FETCHES=6 UNRES=0
+//	append a comment to api.js  path=/{BASE}/things   path=/api/things
+//	                            FETCHES=2 UNRES=2     FETCHES=2 UNRES=2
+//	no-change incremental       path=/{BASE}/things   path=/api/things
+//	                            FETCHES=2 UNRES=2     FETCHES=2 UNRES=2
+//
+// The third row runs no fold at all — zero files changed, so `merged` is empty
+// and the pass emits no stat line. It is here to show that the second row's
+// damage PERSISTS: nothing re-folds it, and it does not recover until the next
+// full index.
+//
+// The fold's own stat line on the SECOND row is where the fix is visible:
+//
+//	before  candidates=2 folded=0 files_sniffed=1 files_indexed=6
+//	after   candidates=2 folded=2 files_sniffed=2 files_indexed=18
+//
+// 6 indexed paths versus 18 is the carried-forward prior graph coming into
+// view, and that is the whole fix. (The full index reports folded=2 in both
+// columns — it never had the problem.)
 //
 // A full-index-only test cannot see any of that, which is why this file
 // exists alongside the golden fixture. It asserts the property that matters —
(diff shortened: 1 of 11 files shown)


cajasmota:

The core mechanism is sound, the identity contract holds, and the tests are real. **Two measured defects block merge**, one of which is the bug this PR fixes coming back.

## 1. Is the fix dead in production? — No, it is live

`buildDocument` has exactly one non-test caller (`cmd/grafel/index.go:2132`, inside `Run`), and `Run` sets `i.absRepo` at `:1434` first. Proven on the daemon's real subprocess path (`grafel index-internal -repo …`, the child `sched` forks): `fold entered srcRoot=…/webui-v2`, then `candidates=3 folded=3 files_sniffed=202`, exit 0.

## 2. Corpus safety — measured, clean

Two full indexes of the in-repo `webui-v2` frontend (264 files, 8,904 records) from one binary with a locally-added no-fold gate. Diffing every `http_endpoint_*` entity and every `FETCHES`/`UNRESOLVED_FETCH`/`RESOLVES_TO` edge: **exactly 3 paths changed, all correct.** Entity IDs byte-identical. `FETCHES` 22 → 22, `UNRESOLVED_FETCH` 4 → 4 — **nothing stopped matching.** On the public `event-driven-ai` corpus the fold is a total no-op (`candidates=0`).

This is the check that mattered most, given a regex widening previously changed 10 of 233 real frontend files with every fixture green.

## 3. BLOCKING — the incremental path reverts the fix, and it is sticky

Via the production `index-internal -incremental <stateDir>` path, using this PR's own fixture as a repo:

```
full index          → folded=3,  FETCHES=9,  UNRESOLVED_FETCH=0
touch client/api.js → folded=0,  FETCHES=3,  UNRESOLVED_FETCH=3
```

One appended comment line — with the constants module `client/config.js` untouched — reverts all three paths to `/{BASE}/things`, `url_kind=dynamic_baseurl`. A subsequent no-change incremental **keeps it there**: it does not recover until the next full index.

Cause: `buildRepoSubstrateResolver` derives its sniff file set from `merged`, which on an incremental run holds only the changed files, so the declaring module is invisible.

The PR documents a *different* limitation — that `ResolveHTTPEndpointHandlersFileScoped` does not fold — and that documented case is correct as written. This is a third path, and it does not merely fail to fold: it **actively un-folds** what a previous full index got right. Entity IDs stayed stable across every run (0 churn, no rename/delete edges), so the identity contract is intact — it is the edges that flap.

Fix direction: seed the sniff file set from the carried-forward prior graph or from the walk, not from `merged`.

## 4. Mutants M4–M7 — genuinely killed, not vacuous

Each re-applied by hand, `go vet` clean (so each actually compiled), scored with `-count=1`, restored by `cp` with `shasum` verified. M4 → `TestFoldConsumerHTTPBaseURLs_Guards_6450`; M5 → `TestLeadingTemplateIdentEngine_6450`; M6 → `…Guards_6450`; M7 → `TestRepoSubstrateResolver_NoAmbientBinding_6450`. The four permissive mutants the golden set structurally cannot see are pinned for real.

## 5. BLOCKING — index-time file reads are unguarded, and the cost is not bounded

**(a) Plain `os.ReadFile`, not `internal/safeio.ReadFile`** — which exists for exactly this (#6416: non-blocking open, FIFO-behind-symlink safe, size cap). A harness with a FIFO named `config.js`, created inside `t.TempDir()`, made `FoldConsumerHTTPBaseURLs` **block >5s and never return.** Reachability is limited — `walk/irregular.go` filters non-regular files — but the fold re-reads **by path after the walk**, so this is an unguarded TOCTOU/symlink hole in a codebase that guards this everywhere else.

**(b) No size cap, no bound on the sniff.** A single candidate triggers a serial re-read and regex sniff of *every* substrate-language file in the repo. Measured on this repo (7,916 files, 121 MB): **+12.3s to an index, for `candidates=1 folded=0`.** The whole `webui-v2` index takes 4.7s by comparison. The pre-scan gates only on "≥1 candidate", which one stray `` fetch(`${BASE}/x`) `` anywhere satisfies.

## 6. baseline.json — clean

`+6 −0`, one surgical block in sorted position, no reflow. Fixture dirs on disk = 25, so the `wantFixtures` bump is correct.

## Not verified, stated rather than glossed

Whether the daemon's watcher picks `TryIncremental` or `index-internal -incremental` for a given edit — both exist, the first no-ops the fold and the second regresses it, so the conclusion is the same either way. Also unverified: behaviour on a >2GB file, and Task 2, which is explicitly out of scope here.

Issue #6382, opened 2026-08-19
convention test: a hand-maintained allow-list must fail on unlisted items too, not only on stale ones


cajasmota:
Promised in #6379, which added the `## Derive, don't list` section to `AGENTS.md`. That section's rule 2 says a hand-maintained list must be **load-bearing** — a test that fails in *both* directions, on a listed item that no longer matches and on an unlisted item that appears. This is the mechanical check for it.

## The shape, and the working example

`TestKnownInvisibleFileAnchoredOffenders` in `internal/extractors/file_anchored_rels_guard_test.go` is the in-repo instance. Verified during #6379's review: it errors when a pinned key matches no site (*"matches no site any more"*), when an unlisted site of the shape appears (*"at unaccounted key %q"*), on a count mismatch, and when the scan matches nothing at all. Its companion matcher walks independently of the main scanner specifically so narrowing one cannot silently narrow the other.

That is what a list that cannot rot looks like. Most lists in the repo are not that.

## What the check would do

Find package-level `map[string]…` / `[]string` vars in `_test.go` files whose names match `known*`, `allow*`, `expected*`, `skip*`, and assert each is referenced by a test that fails on **unlisted** items, not only on listed ones.

Structurally detectable: does the test iterate a **discovered** set and look each item up, or does it only iterate the list? The first is bidirectional; the second can only catch staleness in one direction, and never catches a new instance.

## Honest scoping — this is hygiene, not a gate

**It would have caught one of the eight instances behind #6361**, not two. An earlier draft claimed #6357 and #6345; on review that is wrong:

- **#6357** — yes. Its exclusion was a bare `continue` with a comment, which is exactly the "exclusion not pinned as a named allow-list" shape.
- **#6345** — no. A mis-identified control-flow claim ("one chokepoint" that was Pass 1 of 3), not a package-level var.
- **#6349** — no. An AST-walk scope bug.
- The four #6365/#6375 instances — no. All prose.

So: worth building because it keeps *future* lists honest and the population is small enough that false positives are bounded. **Not** worth presenting as something that would have prevented this week. The other three rules in that section rest on judgement and cannot be gated — the section says so, and this issue does not change that.

## Two cheaper adjacent wins

- Fail any test-file exclusion expressed as a bare `continue` inside a "checks every X" loop unless it references a named allow-list. Grep-level, mostly mechanical, and it is the #6357 shape directly.
[...]

PR #6516, opened 2026-08-21
fix(#6353): the coverage matrix's second input was unguarded, so one registry tag shipped a spurious language row

### LINKED ISSUE #6353: the coverage matrix's language rows come from two sources and #6351 guards only one — a registry tag still ships an unreviewed row
Found while reviewing #6351.

## The gap

The coverage matrix's language rows are the **union** of two sources:

1. slugs derived from `internal/extractors/<lang>/` directories, and
2. `language` tags on records in `docs/coverage/registry.json`.

#6351 binds the **first** source to the real tree — a new extractor directory can no longer reach the matrix unreviewed. **The second source remains completely unguarded.** Adding a registry record with a novel `language` value still ships a row into the published matrix with nothing checking it.

## Live evidence, already shipped

`registry.json` carries exactly **1** record tagged `language: "javascript"` against **84** tagged `jsts`. The result is in `docs/coverage/summary.md:37` today: a spurious `[javascript](by-language/javascript.md)` row sitting next to `[JS/TS]`.

That is precisely the collapse `extractorDirAliases` exists to enforce — `javascript` and `typescript` fold into `jsts` — bypassed by coming in through the registry rather than through a directory. One record defeats the alias table.

## Why it is the same defect as #6332, not a new one

#6332's shape was: the derivation is not purely structural, it filters through hand-maintained tables whose permissive default is silence. This is the same sentence with a different input. The alias table encodes a decision ("javascript is not its own language row"), and the registry can contradict that decision without consulting it.

So the roster added in #6351 guards half the axis and reads as though it guards all of it. Worth fixing before someone relies on it — and specifically before VB.NET S2, which will add both an extractor directory *and* registry records.

## Shape of a fix

The cheap version: assert that every distinct `language` tag in `registry.json` is either a derived slug or an explicitly-recorded exception, and run it in the same real-tree test file that #6351 added. That would fail today on `javascript`, which is the point — the failure is a real defect, not a false positive, and fixing it means retagging that one record to `jsts`.

The question to settle first is whether a registry `language` tag is *allowed* to introduce a row with no extractor behind it. There may be legitimate cases (a language covered only by generic/framework rules), in which case the exception list needs to be explicit rather than implied by silence — and it should be guarded for staleness the way #6351 now guards the alias tables, since this cycle has deleted 34 dead config entries across #6330 and #6351.

## Acceptance

- Retagging the stray `javascript` record makes the spurious row disappear from `summary.md`.
- Adding a record with a fabricated `language` value fails the new test, naming what to do about it — including the alias option, which #6351's first draft omitted from its remedy list and had to be corrected for.

Refs #6330, #6332, #6347, #6351

cajasmota:
Closes #6353

The coverage matrix's language rows are the union of two sources: slugs derived from `internal/extractors/<lang>/` directories, and `language` tags on records in `docs/coverage/registry.json`. #6351 bound the first to the real tree. **The second was completely unguarded**, and the defect was live in shipped docs: one record tagged `javascript` against 84 tagged `jsts` rendered a spurious `[javascript]` row in `summary.md` directly below the real `[JS/TS]` one.

One record defeated the alias table, by coming in through the registry rather than through a directory.

## RED, by execution

```
go test -count=1 -run TestRegistryLanguageTagsAreDerivedSlugsOrJustifiedExceptions ./tools/coverage/   → exit 1

  registry.json tags 1 record(s) with language "javascript", which is neither a slug
  derived from internal/extractors/ nor an entry in registryLanguageExceptions
```

After the retag and regeneration: `go test -count=1 ./tools/coverage/` → **exit 0**, `gofmt -l` clean, `go vet` → 0, `go run ./tools/coverage fmt --check` → canonical.

## The exception list was derived mechanically, and that mattered

The brief handed the implementer a five-value candidate list — `multi`, `c-cpp`, `lisp`, `idris`, `jcl` — with an instruction to derive the set in the test rather than paste the list.

**Four of the five were wrong.** `c-cpp` comes from `internal/extractors/cpp/` via `extractorDirAliases`; `lisp`, `idris` and `jcl` each have their own extractor directory. All four are already derived slugs. Pasting the list would have shipped four dead exception entries on day one — the exact rot this issue exists to prevent, introduced by the fix for it.

Deriving `SupportedLanguages(repoRoot)` against the distinct registry tags leaves **exactly one** genuine exception: `multi` (126 records).

`multi` is an exception rather than a retag because it is real: `generate.go:681,797` skip `n == "multi" || n == ""` from the pivot, so it renders no row. Retagging cross-cutting Docker/Terraform/Actions records to any single real language would be a lie. `c-cpp` does render a row (`summary.md:11`, `[C/C++]`) — correctly, as a derived slug.

That justification is **load-bearing rather than a comment**: the test asserts no excepted tag renders a `](by-language/<k>.md)` row in the shipped `summary.md`.

## Bidirectional, because a one-way list rots

Staleness is guarded in both remaining directions — an entry with zero records (dead config), and an entry that is actually a derived slug (dead *and* misleading). Anti-vacuity fatals cover an empty registry, an empty derivation, and a `summary.md` whose pivot shape changed.

This had to be inline: #6382 proposed a generic meta-check for exactly this property and was **closed today on measurement showing it would fire on nothing**. There is no separate guard coming.

## Mutants

| # | mutation | result |
|---|---|---|
| M1 | retag the record back to `javascript` | KILLED, exit 1 |
| M2 | add a stale `registryLanguageExceptions["fortran"]` | KILLED, exit 1 (staleness half) |
| M3 | **permissive** — drop `n == "multi"` from the `generate.go:681` pivot filter | KILLED — a `[multi]` row appeared at `summary.md:41` |
| M4 | **permissive** — delete `extractorDirAliases["javascript"]`, record retagged | KILLED — **but by the sibling `TestSupportedLanguagesMatchesRealTree`, not by this gate** |

**M4 is the honest caveat and is recorded rather than glossed.** Widening the alias table makes `javascript` a derived slug, so the new gate accepts it *by construction*. The two guards compose — this one binds registry → derivation, the roster binds derivation → tree — and neither covers the other's axis alone. Each was `go vet`-clean before scoring, and every mutated file was restored by `cp` with its shasum re-verified.

## Docs delta

`infra.state-machine.xstate` retagged `javascript` → `jsts`. Spurious row gone, `by-language/javascript.md` deleted, JS/TS Other 9 → 10, headline `40 (39 active · 1 placeholder)` → `39 (38 active · 1 placeholder)`. The placeholder is `bicep` — a derived slug with zero registry records, unaffected.

Note for whoever touches VB.NET next: `vbnet` already has a registry record, so that work already flows through this path.


cajasmota:

Verified by execution, not by reading. Highlights:

- **The gate is not vacuous.** Reverting only the `registry.json` retag (`jsts` → `javascript`) and leaving everything else on the branch → **exit 1** with the claimed message. The issue's acceptance criterion is met.
- **The exception list is correct and minimal, independently re-derived.** `internal/extractors/` contains `cpp/`, `lisp/`, `idris/`, `jcl/`, and `languages.go:113` maps `"cpp": "c-cpp"`. All four of the candidates I supplied are derived slugs. An independent enumeration (822 records, 39 distinct tags, `jsts`=85, `multi`=126, `vbnet`=1) leaves exactly `multi`.
- **Staleness fails in both directions.** Injecting `registryLanguageExceptions["fortran"]` (zero records) and `["lisp"]` (a derived slug) → **exit 1** with three distinct errors: dead-config, derived-slug-is-dead, and the shipped-summary row check.
- **Docs are generator output, not hand-edited.** `go run ./tools/coverage gen --out <scratch>` then `diff -r` → no diff in the pivot rows, `by-language/jsts.md`, or `by-category/platform.md`. The only `summary.md` delta is a pure `--out` artifact, because `generate.go:729` calls `SupportedLanguages(outRoot)` and a temp dir has no `internal/extractors/`. No dangling `by-language/javascript` reference survives outside one explanatory comment.
- **The anti-vacuity fatal fires.** Renaming the jsts pivot link → **exit 1**, *"the pivot's shape changed and the row check below would assert nothing; refusing to pass vacuously."*

### A second construction-dependency, found by the reviewer and worth recording

The PR already names one: mutant M4 (deleting `extractorDirAliases["javascript"]`) is caught by the sibling `TestSupportedLanguagesMatchesRealTree`, not by this gate, because widening the derivation makes the tag acceptable *by construction*.

There is another. The **"no matrix row" assertion is checked against the committed `summary.md`, not against `generate.go`'s filter.** Retagging a record to `fortran`, adding a `fortran` exception, and *not* regenerating docs → gate **exit 0**, whole package **exit 0**. An exception whose justification is false passes.

It is caught downstream by `coverage-docs.yml`'s regenerate-and-diff, with the authoritative unconditional copy in `test.yml` per #6352 — so the guards compose and this is not a blocker. But it is a second place where this gate depends on something outside itself, and that belongs in the record rather than in someone's head.

## Coordinator killing mutants — the acceptance criterion nobody had executed

Both the author and the reviewer tested the gate by **reverting an existing tag**. The issue states a second criterion: *adding a record with a fabricated `language` value must fail the test, naming what to do about it.* That path was untested, so I ran it.

**Mutant 1 — a brand-new record with `language: "cobol99"`:** `go test -count=1 -run TestRegistryLanguageTags ./tools/coverage/` → **exit 1**, and the message names both remedies, including the alias option that #6351's first draft had to be corrected for omitting.

**Mutant 2 — a new record with `language: ""`**, the case the reviewer explicitly listed as not verified: → **exit 1**. That gap is closed.

`registry.json` restored by `cp` after each, shasum `f0b02c0d…` re-verified identical both times, worktree swept.

Merging.

Issue #4406, opened 2026-06-10
Production dedup path drops base-node QualifiedName + edges (the live-graph half of #4402)


cajasmota:
Split out from #4402 (ref epic #4334). #4402 (PR #4405) fixed the documented `MergeWithCustom` boundary — but that boundary is invoked only by repro/unit harnesses. The PRODUCTION daemon (`internal/daemon/extract/subproc.go`) emits base+custom entities independently and dedups downstream via a different path, so the live-graph re-orphaning is NOT fixed by #4405.

Evidence it's a real production bug: the per-language workarounds (#4366 CONTAINS re-emit; #4379 settings-wiring tier-2 leaf fallback; the #4328 analog) were each live-validated against the real `upvate_core` graph and DID fix it — proving production drops the base node's QualifiedName/CONTAINS edges somewhere in the subproc dedup, and the workarounds compensate by re-emitting before that dedup.

**Task:** trace the production dedup path (`internal/daemon/extract/subproc.go` + whatever cross-subprocess entity dedup/merge it feeds). Where a custom entity supersedes a base entity of the same identity, apply the SAME semantics #4405 added to `MergeWithCustom`: inherit QualifiedName when the survivor's is empty, union + re-key the base node's edges (esp. #526 class→field CONTAINS), conservative gap-fill. Ideally route both paths through the shared `supersedeBase` helper from #4405 so there's ONE merge semantics, not two.

**Live-validate (MANDATORY, not fixtures):** this must be validated on the REAL graph — coordinator live-reindex of `upvate-v3` + `upvate` at a deploy window, then verify (a) model fields stay members and (b) the listed workarounds can be DISABLED without re-orphaning. Only after that passes should the redundant workarounds be removed (separate cleanup PR). Until then, KEEP all per-language workarounds.

This is the true long-term root fix; #4405 was the boundary-hygiene half.

PR #6517, opened 2026-08-21
fix(#6440): VB.NET overloads shared an entity id — 279 collision groups, not the 4 this was deprioritised on

### LINKED ISSUE #6440: vbnet: two overloads of the same method in one type share an entity id, so they merge into one node and each other's edges
`declName` in `internal/extractors/vbnet/extractor.go:274-284` renders a member as `<EnclosingType>.<member>` with **no parameter list**. Entity identity is

```
graph.EntityID = sha256(repo, kind, name, sourceFile)
```

and `Subtype` is not part of that hash. So two overloads of the same method, declared in the same type in the same file, produce the **same** name, kind and source file — and therefore the same entity id. They merge into one node.

Nothing reports this. The merge is silent: the graph shows one operation where the source has two, and every edge from either overload lands on the survivor.

## Measured, not hypothesised

Found while grounding the method-level `Implements` slice of #6327, on the 302-file VB.NET corpus. Two files contain overload pairs where **each overload implements a different interface member**:

- `GlobalizedDescriptor.vb` — `GetEvents`, `GetProperties`
- `Dynamic.vb` — `GetEvents`, `GetProperties`

These are the `ICustomTypeDescriptor` two-arity pairs (`GetEvents()` and `GetEvents(attributes)`). Both arities collapse onto a single entity, and both `Implements` targets would attach to it.

Same-name members elsewhere in the corpus (`Dispose`, `Info`, `CompareTo` in `FrameServer.vb`, `Main.vb`, `SimpleUI.vb`) are **not** affected — those are distinct *types* sharing one file, so the `<EnclosingType>.` prefix already separates them. The collision is specifically overload-vs-overload within one type.

## Why it matters more now

This was survivable while VB.NET emitted no member-level edges. Once member-targeted edges exist, a collapsed overload pair stops being a missing node and becomes a **wrong** one: two different interface contracts attributed to a single operation, with no way for a consumer to tell which arity satisfies which.

## Scope

Whether to include an arity or parameter-type discriminator in `declName` is the decision, and it is not free — changing `declName` changes entity ids for every VB.NET member, which is a reindex-visible change and interacts with incremental indexing. That trade-off is the substance of this issue, not the detection.

Filed as measurement, not as a blocker on the current slice. The count here is small (2 files of 302), but the failure is silent, which ranks it above louder breakage.

cajasmota:
Closes #6440

Two overloads of the same method in one type hashed to the same `EntityID`, so the #4406 dedup kept the first, dropped the rest, and **unioned the dropped records' edges onto the survivor** — with no counter and no log. Option B: only members that actually collide get a discriminator; every non-colliding member keeps its exact current name and ID.

## The number this issue was deprioritised on was ~70× low

Measured with the **real extractor** on the same 302-file, 3-repo public MIT corpus #6327 used:

```
before:  279 colliding EntityID groups,  364 dropped records   (of 13,479)
after:     5 groups,  5 records
```

Not 4. The prior figure was scoped to the `Implements`-carrying subset it happened to be measured for; the regex estimate of ~257 was roughly right. **277 of the 279 are member-vs-member overloads.**

The 5 residual groups are **not overloads** — they are two files each declaring one *type name* twice, plus that type's members. `declName`'s deliberate bare-name rule for types accepts that, and this change does not touch it.

## The two gates that could have killed option B

**Partial classes — absent, and structurally so.** Zero collision groups span two physical files. `partialAnchor` is guarded by `isPartialType(n)`: it rewrites `recFile` for **types only**, so a member in a designer half always keeps its own path and member overloads can never be re-anchored into a shared file. A per-file pre-pass is sufficient **by construction**, not by corpus luck — which is the difference between "works here" and "works".

**Incremental re-index — not a half-rekeyed graph.** This was the one genuinely silent failure mode in the whole issue, so it was run on the production `index-internal --incremental` path, seeding state with a pre-fix binary and re-indexing with the fixed one. (The first attempt silently fell back to a full reindex and was corrected — worth noting, because that fallback would have made the gate look green while testing nothing.)

- Edit a **colliding** file → incremental graph identical to a clean full index: 86 VB entities, 0 key diffs, **0 ID mismatches**.
- Edit a **non-colliding** file → the untouched colliding file stays un-split (85 vs 86): stale, but **0 ID mismatches and no *additional* dangling endpoints**.

Two precisions on those two lines, both from review, because the first drafting of each was stronger than the measurement supports:

- **"0 dangling" is wrong as an absolute; "no additional dangling" is the claim.** 4 unresolved targets are present in the pre-fix graph, the clean full index and the incremental graph alike. They pre-date this change and this change neither adds to them nor clears any.
- **"no existing ID ever moves" holds for DECLARATION entities, not for every node in the graph.** Three `SCOPE.Process` flow nodes whose derived `proc:` names embed a split member's name do move (`X.Dispose → Y` becomes `X.Dispose() → Y`), and 2 new flows appear. There is no `fbversion` consequence — flow nodes are re-derived on every index and are stripped from carried-forward state on the incremental path — but the flat form of the claim was wrong and is corrected here rather than dropped.

With those two corrections: collision-only means no *declaration* ID moves, so incremental staleness here is "a node not yet split" — identical in kind to any extractor recall improvement, and it self-heals on the next touch or full index.

## The bare-name question, which was the strongest argument for B over A

`byLocation` is keyed `(file, name)` and retained only when unique, and call targets are emitted as bare names with no arity. That reasoning was established by **reading** the resolver, so it was measured:

```
273 at-risk overloaded members · 566 resolved CALLS aim at them
before → after:  CALLS 8,044 total / 4,166 resolved
                 at-risk 566 resolved / 707 dangling   —  IDENTICAL
```

All 566 still land, on the bare-named overload. And bare-name refs do **not** survive without a bare-named entity: mutant M2 discriminates the first overload too, and `StructPtr.Finalize -CALLS-> StructPtr.Dispose` immediately dangles. The rule is load-bearing, proven by executing the resolver rather than by inspection.

**Which overload keeps the bare name: first in source order.** Deterministic from file content, no map iteration, and it matches the #4406 dedup's existing survivor, so no ID moves. It *would* move if someone inserted a new overload above the current first — stated as a known property rather than discovered later.

This splits **declarations**, not call-site arity precision: call targets are still bare names.

## Discriminator choice

Declared **parameter-type list** (`StructPtr.Dispose()`). Not an ordinal — that renumbers every later overload on insert. Not arity — same-arity overloads are the common case here. Parameter *names* are excluded so a rename cannot move an ID.

## The golden fixture pinned that a discriminator EXISTS, and nothing pinned what it CONTAINS

Found by a mutant run in review, not by me: `paramTypeList` mutated to append parameter **names** — the one thing its doc comment says it must never do — **survived the entire committed suite**, `./internal/extractors/vbnet/`, `./cmd/grafel/ -run 'Quality|Fixture|Golden|VBNet'` and the #6440 fixture test alike. The reason is in the fixture note itself: the only discriminated member in `vbnet-mini` is `StructPtr.Dispose()`, whose parameter list is **empty**, so the discriminator renders as a bare `()` and every character `paramTypeList` emits between the parentheses was unobserved.

That is a comment asserting a property no test observes, inside the PR whose purpose is to make overload identity trustworthy. Fixed with unit tests on the two functions directly (`internal/extractors/vbnet/overload_suffix_6440_test.go`, `package vbnet` so the unexported helpers are visible), **not** by growing the fixture:

| test | property pinned |
|---|---|
| `TestOverloadSuffixRendersParameterTypes_6440` | a NON-empty list: `(String)`, `(String, Integer)`, and type params leading a generic overload |
| `TestOverloadSuffixExcludesParameterNames_6440` | two types whose overloads differ ONLY in parameter names get the SAME discriminator |
| `TestOverloadSuffixKeepsArrayRank_6440` | `Byte()` does not fold into `Byte` |
| `TestOverloadSuffixOrdinalFallback_6440` | three same-rendering siblings → `#2`, and the per-file name set stays injective |
| `TestOverloadSuffixEmptyWhenNotColliding_6440` | option B's mechanism: a non-colliding member gets no suffix at all |
| `TestOverloadSuffixGroupsByKind_6440` | a Field and a Sub of one name are different kinds, already distinct, and neither is discriminated |
| `TestParamTypeListDirect_6440` | 11-case table on the renderer alone: names excluded, ByRef/Optional/default folded, untyped → `Object`, qualified type kept as written, return type excluded |

Writing them surfaced a **real defect the hand probes had not**: the parser lifts the rank marker off the type (`payload As Byte()` → `TypeName "Byte"` + `IsArray`), so reading `TypeName` alone folded an array overload into its scalar sibling and pushed it onto the ordinal fallback. `paramTypeList` now emits the rank marker. VB overloads on exactly that difference, so the discriminator does too.

## Mutants

| # | mutation | result |
|---|---|---|
| M0 | revert `base + overloadSuffix(n)` → `base` | **KILLED** — 6 assertions |
| M1 | `len(group) < 2` → `< 1` (permissive) | **EQUIVALENT** — a dead guard, now documented in code so it is not misread as a suite hole |
| M2 | discriminate the first overload too (permissive) | **KILLED** — bare name gone *and* the `Finalize → Dispose` CALLS dangles |
| M3 | every sibling joins the group (permissive, option-A shaped) | **KILLED** — `StructPtr.ObjSize` renamed, pinned ID moves |
| M4 | `paramTypeList` appends parameter NAMES (the review's surviving mutant) | **KILLED** — 5 top-level tests, 10 subtests. Survived everything before this commit |
| M5 | drop the array rank marker | **KILLED** — `TestOverloadSuffixKeepsArrayRank_6440` + the `array rank kept` subtest |

M3 is the one that makes this option B rather than option A. M4 is the one that shows why the fixture alone was not enough.

## Fixture — the defect was previously unobservable

`vbnet-mini/src/Win32Native.vb` already contained the `StructPtr.Dispose` pair and **nothing asserted it**; `expected.json` documented the merge in prose while asserting nothing. Now: both overloads as expected entities, a `forbidden_relationships` entry for the phantom self-`CALLS` (`Dispose()` calls `Dispose(True)`, so the merge minted a self-call on a method that does not call itself), the `IMPLEMENTS` `from_name` amendment, and five prose notes corrected.

**Option-B proof pinned as a literal:** `StructPtr.ObjSize` ID `cfb4b4428aa5b24d`, recorded before the change and asserted unchanged after.

RED → GREEN: `-run TestVBNetOverloadIdentity_6440 ./cmd/grafel/` → exit 1 (4 assertions, including the self-CALLS) → exit 0. `baseline.json` entities 20/20 → 22/22; relationships unmoved at 33/33.

## Verification

`./internal/extractors/vbnet/` → 0 · `./internal/quality/...` → 0 · `./cmd/grafel/ -run 'Quality|Fixture|Golden|VBNet'` → 0 · `gofmt -l` clean · `go vet` → 0. All `-count=1`.


cajasmota:
**Recording a measurement that was made during review but never landed in this PR's body**, because the 0.3.0 CHANGELOG cites it and a reader following the citation would not find it here.

The PR body measures two things: the incremental probe (86 VB entities, 0 key diffs, **0 id mismatches**) and collision groups on the corpus (**279 → 5**). Neither is a corpus-wide shared-key id diff — and I compressed them into one when drafting the CHANGELOG, which a reviewer caught.

## What the independent review of this PR actually measured

Two binaries built from this tree differing **only** in `internal/extractors/vbnet/extractor.go` (fixed vs `1456968af`'s version), each indexing the 302-file / 3-repo public VB corpus, then diffing every `(kind, name, source_file) → id` tuple:

```
entities   13,234 → 13,595
keys added        364
keys removed        3
shared keys whose ID changed   0
```

The **3 removals are the `SCOPE.Process` flow nodes** whose derived `proc:` names embed a split member (`X.Dispose → Y` becoming `X.Dispose() → Y`), plus 2 new flows. Those are re-derived every index and stripped from carried-forward state on incremental, which is the reason they carry no `fbversion` consequence — an argument that applies **to the flow nodes specifically**, not to the id-diff result as a whole. I had attached it to the wrong measurement.

The same review also confirmed, on the fixture, 93 → 94 entities with exactly **1 key added** (`SCOPE.Operation StructPtr.Dispose()`), 0 removed, 0 shared-key id changes.

## Why this is worth a comment rather than leaving it in a transcript

"0 shared keys whose id changed, corpus-wide" is the single load-bearing claim behind shipping this as option B — collision-only — instead of option A with an `fbversion` bump and a global cross-language reindex. It decides whether a user's existing graph survives the upgrade.

A number that important should be reachable from the PR that relies on it, not only from a review thread. Filed here so the CHANGELOG's statement has a citable source and the next person auditing the release can check it.

Issue #6461, opened 2026-08-21
the daemon runs no cross-file composition pass and prunes only by SourceFile, so a composed endpoint on an unchanged mount file goes stale and coexists with a fresh short one


cajasmota:
The daemon fast path (`internal/extractors/incremental.go`) prunes only entities whose `SourceFile` is in the changed set (`:779-792`, and the `removedEntityIDs` collection immediately after). It never invokes any Pass-2.6 cross-file composition pass — `grep -n "NestedURLConf\|DjangoDRFRoutes\|urlconf" internal/extractors/incremental.go` returns **zero matches**.

Combine those two facts and any entity **attributed to a file other than the one that produced its content** goes stale and is never corrected.

Django's nested-URLconf pass is the known instance: `ApplyDjangoNestedURLConf` attributes the composed endpoint to the **mount** file (`internal/engine/django_urlconf_nested.go:227`, `SourceFile: relPath`). So on the daemon:

1. A route file changes. Its own entities are pruned and re-extracted, producing a fresh **short-path** endpoint.
2. The composed endpoint lives on the *mount* file, which did not change — so it is not pruned, and the composition pass never re-runs.
3. The graph now carries a **stale composed endpoint alongside a correct short one**. Rename a route and the ghost persists indefinitely.

This is worse than the failure mode described in #6414, which assumed the short path would *overwrite* the composed one. It does not overwrite it — both coexist.

## Why this was not noticed

Two accidents, both confirmed:

**On the CLI path, Django is masked by a filename coincidence.** `internal/indexer/diff/diff.go` cross-invalidates by **basename** in both filter entry points (`Filter` at `:239-254`, and `FilterWithGit` — the one `cmd/grafel/index.go:1559` actually calls — at `:636-650`; `moduleBase` at `:780-787` is the stem without extension). Django's convention is that the mount file and the route file are *both* named `urls.py`, so editing either drags the other into the changed set and the pass recomposes correctly. The comment at `:226-228` is explicit that this is a stand-in: *"a proper import-graph traversal is left for a future pass"*.

[...]

Issue #2678, opened 2026-05-27
[Epic] Audit endpoint attribution across all web frameworks — extension of #2677


cajasmota:
## Source
#2677 confirmed DRF/Django endpoints are misattributed to the routing/registration file instead of the actual handler method file. This bug is **likely systemic across all framework extractors**, because the same architectural pattern (decorator-based / registration-based routing with handler methods in separate files) is the norm in modern web frameworks.

### Python (sub-audit)
- Flask (@app.route, Blueprint.route)
- FastAPI (@router.get, @app.get, APIRouter)
- Starlette (Route, Mount)
- Tornado (RequestHandler subclasses + Application route table)
- Pyramid (config.add_route + view_config)

### JavaScript/TypeScript (sub-audit)
- Express (app.get/post/put/delete, Router)
- Koa (router.get/post, Koa-Router)
- NestJS (@Controller, @Get/@Post)
- Fastify (fastify.get, fastify.register)
- Hono (app.get, app.post)
- Next.js API routes (pages/api/*, app/api/*/route.ts)
- tRPC (router.procedure)

### Go (sub-audit)
- net/http (http.HandleFunc, ServeMux)
- gin (router.GET, .POST, .Group)
- echo (e.GET, .POST, .Group)
- chi (r.Get, .Post, r.Mount, r.Route)
- gorilla/mux (r.HandleFunc, .Methods)
- fiber (app.Get, .Post)
- huma (OpenAPI-driven)

### JVM (sub-audit)
- Spring Boot (@RestController, @RequestMapping, @GetMapping)
- JAX-RS (@Path, @GET)
- Play (routes file + Action class)

### Others (sub-audit)
- Ruby on Rails (config/routes.rb + controllers)
- Sinatra (get '/path' do ... end)
- PHP Laravel (Route::get + Controllers, route files)
- PHP Symfony (#[Route] attributes, YAML routes)
- C# ASP.NET Core ([HttpGet], [Route], Controllers)
- Rust Axum (Router::route + handler fns)
- Rust Actix (App::route + handlers)
- Rust Rocket (#[get(\"/path\")])
- Elixir Phoenix (scope/get/post in router.ex + Controller modules)

## Audit procedure (for each framework)

1. Find the extractor in \`internal/engine/\` or \`internal/extractors/\`.
2. Identify which file path the endpoint entity's \`source_file\` resolves to.
3. **Bug condition**: the source_file points to the routing/registration site instead of the file where the handler function/method body lives.
[...]

PR #6528, opened 2026-08-21
fix(#6461): the daemon ran no cross-file composition pass, so a urlconf edit left one endpoint stale and invented another

cajasmota:
Refs #6461 — **the ghost shape only. This does not close the issue.**

The daemon's incremental path ran **no cross-file composition pass** and pruned only by `SourceFile`, so a Django urlconf edit left the graph carrying a stale mount endpoint and inventing another. New `internal/extractors/incremental_django_compose.go` plus Step 7b in `incremental.go` recompute-and-reconcile: run `engine.ApplyDjangoNestedURLConf` over the current tree, prune what the graph carries under this pass's ownership (`pattern_type` ∈ `urlconf_nested_include` / `url_mount_point`) that the recomputation no longer produces, add what it lacks. Gated on "a `.py` changed AND the repo has a `*urls.py`".

## The brief's premise was wrong, and that changed the shape of the fix

`#6461` ships with `TestMountParity_6461_Django_RoutePathRename_PathA` self-skipping as "#6461 UNFIXED", and the obvious plan was to flip that skip to a pass.

**That test is red for three independent reasons, only one of which is the ghost.** Flipping it would have required fixing all three — or, worse, invited narrowing the assertion until it passed.

Instead: an **ungated** ratchet scoped to the ghost, on the *same fixture, same edit, same helpers* — no new fixture. `TestMountParity_6461_Django_RoutePathRename_PathA_EndpointCensus` asserts the `http_endpoint*` / `url_mount` census matches a clean full rebuild.

| | exit | measured |
|---|---|---|
| census, before | **1** | `LOST http:ANY:/network/conditions` + `INVENTED http:ANY:/network/terms` |
| census, after | **0** | 0 lost, 0 invented |
| gated `…PathA` | still fails | **22 → 6** divergences |
| gated Django ROUTE / MOUNT | still fail | 18 → 6, 5 → 3 |

The gated tests stay gated and honest about what is still broken.

## Confirming the incremental path actually ran

The trap on this path has caught several attempts: a scratch repo inside a worktree makes `FilterWithGit` resolve the *enclosing* archigraph repo, mark everything dirty, and silently turn an "incremental" run into a full reindex — making a broken path look green.

Test repos are `t.TempDir()` (`/var/folders/…`), outside any git tree. Every run's daemon log reads `incremental: done changed=1 …` plus the new `incremental: django-urlconf recompose composed=2 added=1 stale_pruned=1 rels=1 (#6461)`, and **never** `falling back to full reindex`. `changed=1` also proves basename cross-invalidation did not fire.

## Mutants — and one that survived the end-to-end gate entirely

| # | mutant | direction | result |
|---|---|---|---|
| M1 | disable the pass (required revert) | — | **KILLED** — census RED, unit RED |
| M2 | drop the guard on adds | permissive | **KILLED** by the unit ratchet — **census SURVIVED it** |
| M3 | `isComposedDjangoEndpoint` owns every endpoint kind | permissive | **KILLED** — prunes a foreign `http_endpoint_synthesis` |
| M4 | prune edges by `FromID` only | permissive | **SURVIVED** the census gate *and* the whole `internal/extractors` package |

**M4 is the finding.** It left the ghost correctly removed while the graph gained a dangling inbound `IMPLEMENTS` row — so every entity-set assertion stayed green. Same for M2, which produces duplicates. An end-to-end census is structurally blind to both: it compares *entity sets*, and these defects live in edges.

Fixed by extracting `prunesRel` and pinning **both directions** at the seam in `incremental_django_compose_6461_test.go`; M4 is now KILLED. This is why the permissive direction has to be asserted at the seam rather than end-to-end.

## The entity-loss half is NOT fixed — confirmed, not assumed

`[ENTITY-LOST] SCOPE.Operation | GET /conditions | mpsite/urls.py` remains. Root cause located: that entity comes from a **cross** extractor (`internal/extractors/cross/endpoint`), and **`TryIncremental` runs no cross extractors at all** — stated in the tree's own comment at `entityRecordToGraphEntity`. Different root cause, different fix, filed separately.

It plus its two edges account for 3 of the 6 residual divergences. The other 3 are scoped-resolver binding differences (`IMPORTS` binding to `Module/mpsite.views` vs `SCOPE.Component/mpsite/views.py`, and a lost self-`ROUTES_TO`) — also unrelated to composition.

## Verification

`gofmt -l cmd/ internal/` clean · `go vet ./cmd/grafel/ ./internal/extractors/... ./internal/engine/` → 0 · `go test -count=1 ./cmd/grafel/` → **ok** (127s, whole package) · `./internal/extractors/...` → **ok**, 75 packages, 0 FAIL · `./internal/engine/` → **ok**. All `-count=1`, exit codes captured separately.

---

## Review round: three fixes, one of them a regression this PR introduced

Independent review confirmed M2/M4 die non-vacuously, the incremental path genuinely runs, cost is **11.3ms at 10,000 files** (I/O is bounded — `ApplyDjangoNestedURLConf` reads only `*urls.py`), and the gated test was **not** narrowed. It found three problems.

### 1. The ownership predicate deleted another pass's live entities — BLOCKING

`pattern_type: "url_mount_point"` is **not exclusive to Django.** `fastapiMountPointSynthetics` (`http_endpoint_synthesis.go:3169`, from #6385) stamps the identical value on a kind `isComposedDjangoEndpoint` accepted. Proven by probe:

```
--- FAIL: TestREV6528_PrunesForeignFastAPIMountPoint
  pruned a FOREIGN FastAPI url_mount_point synthetic http:ANY:/api/v2:mount@svc/main.py
```

Reachable in any Django+FastAPI monorepo, silent, and permanent until a full reindex. The predicate now requires `framework == "django"`, and an entity carrying **no** `framework` is also not claimed — this pass always writes one, so its absence means another producer.

The original M3 only varied `pattern_type`; it never covered **a foreign producer of the same `pattern_type`**, which is the realistic case. That case is now in `TestRecomposeDjangoURLConf_PrunesOnlyEndpointsItOwns`. The underlying non-uniqueness of `pattern_type` is filed as **#6530** — this predicate fix closes the instance, not the class.

### 2. A read failure pruned everything

`reader` swallowed `os.ReadFile` errors to `nil`, so a transient failure on a root `urls.py` left the recomputation empty while the graph still held composed endpoints — and the pass pruned **all** of them. A `readFailed` flag now **skips the prune entirely**.

A path *outside* the walked set is deliberately **not** treated as a read failure: that is the security guard's definite "no such file", and treating it as failure would let a bogus `include("does.not.exist")` disable the prune permanently.

### 3. The invented endpoint was introduced by this PR, not exposed by it

The reviewer flagged a DRF mechanism without a repro. Measuring it changed the disposition:

```
pass disabled:  lost=6  invented=0
pass enabled:   lost=6  invented=1   (http:ANY:/api/widgets)
```

The 6 losses are pre-existing (#6529) and identical either way. **The invented endpoint is this change's.** A regression a PR creates cannot be filed for later, so it is fixed here.

**Mechanism: abstention.** `drfUnknownCoveragePaths` collects the prefix of every `router.register()` in a file changed this tick; routes under those prefixes are neither added nor pruned. Same decision as the read-failure path — act on nothing when the input is untrustworthy. Re-derivation was rejected as #6529's territory, and because it reintroduces the per-tick whole-repo I/O that the graph-backed `dropDRFCovered` substitution exists to avoid.

Two properties keep it from over-reaching, both pinned: it keys on the **registered prefix** rather than "a DRF file changed" (so a `urls.py` mixing `router.register("widgets")` with a plain `path("terms/", …)` still recomposes `/…/terms`), and it matches by **suffix, not containment** (the composition puts the prefix in the last segment, so `/api/widgets-archive` and `/widgets/detail` are not swept in).

**The prune half is load-bearing, which was not obvious.** Suppressing only the add leaves the record in `freshEnts`, so the prune looks unaffected — but the full path handler-resolves composed endpoints onto the *handler's* file (#2678), so a graph copy can carry a different id than this pass computes once the DRF entities are gone. It then reads as stale, and an add-only abstention would **delete the very route it just declined to re-derive.**

| layout | before | after |
|---|---|---|
| ordinary (ViewSet in `views.py`) | lost=0 invented=0 | **lost=0 invented=0** |
| same-file (ViewSet in the edited `urls.py`) | lost=6 invented=**1** | **lost=6 invented=0** |

Both layouts are subtests of one test, so a fix that suppressed composition everywhere cannot pass.

### Round-2 mutants

| # | mutant | direction | result |
|---|---|---|---|
| M5 | ownership by `pattern_type` alone | permissive | **KILLED** — only the FastAPI assertion fails |
| M6 | drop the `readFailed` guard | permissive | **KILLED** — only the read-failure test fails |
| M7 | never abstain | permissive | **KILLED** — same-file subtest fails; **ordinary subtest still passes** |
| M8 | containment instead of suffix | over-reach | **KILLED** — only the suffix test fails |

Each fails on its own assertion rather than through collateral breakage. Gated parity counts unchanged at 6 / 6 / 3, all cross-extractor or scoped-resolver, none composition.

---

## Round 2 review → round 3: the abstention re-opened the ghost

Round-2 review verified `readFailed` is correct, all four of M5–M8 die on their **own** assertion with no collateral, M7 independently reproduced the `invented 0→1` measurement, and the ghost fix had no regression (census 0-lost/0-invented, gated parity 6/6/3 with no composition among the residuals, skip reason unchanged).

It found one blocking defect, and it is the sharpest one available: **the DRF abstention re-created the very ghost this PR fixes.**

`pathMatchesAnySuffix` matched `strings.HasSuffix(path, "/"+prefix)` against **every composed path in the repo**, with nothing tying the suffix to the mount the registering file sits under:

```
changed=[mpsite/urls.py]                  -> ghost /network/conditions PRUNED     (correct)
changed=[mpsite/urls.py, apiapp/urls.py]  -> ghost NOT pruned   removedIDs=map[]
```

A `router.register(r"conditions")` in `apiapp/` suppressed pruning of an unrelated route in `mpsite/` mounted at `/network/`. In a monorepo, `users`, `items` and `orders` are the realistic collisions.

**Why the existing tests missed it:** the narrowness claim was true on the *containment* axis — M8 proves that — and said nothing about **scope**. `TestPathMatchesAnySuffix_IsSuffixNotContainment` pinned the axis that was already right. Containment and scope are two axes; treating them as one property is what let the gap hide, and the comments now say so.

### The fix: differential, not by name

Re-run the composition over the same file list with every `router.register(...)` **call site in the changed files textually removed**. Whatever composed path disappears between the two runs came, by construction, from one of those registrations — **mount prefix and all**. Matching is then equality on the full composed path.

This avoids duplicating module-path resolution (engine's `modulePathToFilePath` is unexported) and re-deriving scope by name. The mount scoping is **inherited from the composition itself**. It also handles the mixed-file case more precisely than the original: stripping only the register call sites leaves `path("terms/", …)` composing, where the previous version took the whole file's word for it.

Cost: one extra composition over **already-cached** content — the reader memoises, so no second read — and only when a changed `.py` actually declares a registration.

### Proof

The fixture mounts two apps sharing last segments and registers **two** prefixes, so a scope bug is caught on both sides — `/network/terms` on the add side, a stale `/network/legacy` on the prune side.

**M10** (the defect re-applied: bare last-segment suffix, repo-wide) → `go vet` 0, `go test -count=1 ./internal/extractors/` **exit 1**, dying on both halves:

```
--- FAIL: TestRecomposeDjangoURLConf_UnrelatedRouteStillPrunesAcrossMounts
    the ghost http:ANY:/network/legacy@mpsite/views.py was NOT pruned … removedIDs=map[]
    /network/terms was not composed … added=[/network /api]
--- FAIL: TestPathIsUnknownCoverage_ExactPathNotSuffix
```

`removedIDs=map[]` reproduces the reviewer's probe output verbatim. With the fix: exit 0, ghost pruned, `/network/terms` composed, `/api/terms` and `/api/legacy` still abstained. **M11** (add-loop-only abstention, re-scored because the code changed) → KILLED.

New tests: `TestDRFUnknownComposedPaths_ScopedToTheRegisteringMount`, `TestRecomposeDjangoURLConf_UnrelatedRouteStillPrunesAcrossMounts`, `TestPathIsUnknownCoverage_ExactPathNotSuffix` (replacing the containment-only test), and the retained regex-parity test.

### Two non-blocking notes, both addressed

- **Prune-side evidence.** The pinning test now uses `mpsite/urls.py`, a file the fixture really has, and its comment states plainly that the id divergence is **constructed by attribution, not observed** from a real resolution flip. Kept — worst case is one stale tick.
- **A claim that was literally false.** Round 1's "reads only `*urls.py`" was quoted in this body and is corrected: the scan reads every changed `.py`, because a registration can live in `routers.py` or any module an `include()` reaches. The bound that does hold is the changed set, capped by the too-many-changed gate.

### DIFF
--- a/cmd/grafel/incremental_mount_parity_6461_test.go
+++ b/cmd/grafel/incremental_mount_parity_6461_test.go
@@ -821,3 +821,135 @@ func TestMountParity_6461_Django_RoutePathRename_PathA_EndpointCensus(t *testing
 			len(lost), lost, len(invented), invented)
 	}
 }
+
+// ─────────── #6528 — DRF coverage, BOTH layouts (the regression this
+// pass created, and the layout that never had it) ───────────
+//
+// `dropDRFCovered` in internal/extractors/incremental_django_compose.go decides
+// whether a composed ANY endpoint is superseded by per-verb
+// `drf_router_expanded` entries by reading those entries out of the GRAPH,
+// because `ApplyDjangoDRFRoutes` has exactly one caller (cmd/grafel/index.go)
+// and re-running it per tick would mean reading every .py file in the repo.
+//
+// That is sound only while the entries are still IN the graph, and there is one
+// layout where they are not. Both are exercised here, in ONE test, because a
+// fix that suppressed composition everywhere would pass a test that only
+// covered the reproducing layout:
+//
+//	ORDINARY  — ViewSet in views.py. `bridgeEndpointToHandler` re-attributes
+//	  every drf_router_expanded entity onto views.py, which the urls.py edit
+//	  does not touch, so the coverage survives Step 5 and dropDRFCovered
+//	  reaches the right verdict unaided. MEASURED lost=0 invented=0 both
+//	  before and after the #6528 fix — it must STAY that way.
+//
+//	SAME-FILE — ViewSet declared in the edited urls.py. Those entities are
+//	  attributed to that file, Step 5 prunes them, nothing re-derives them, and
+//	  coverage reads ABSENT when it is merely INVISIBLE. MEASURED lost=6
+//	  invented=1 (`http:ANY:/api/widgets`) — the 1 INVENTED introduced purely by
+//	  the composition pass, confirmed against the same fixture with the pass
+//	  disabled (lost=6 invented=0). After the fix: lost=6 invented=0.
+//
+// `lost=6` in the SAME-FILE case is asserted as-is rather than fixed: it is
+// #6529 (drf_router_expanded entities are never re-derived on the incremental
+// path), it is identical with this pass disabled, and papering over it here
+// would hide a defect that belongs to a different change.
+func mpDRFOrdinary(t *testing.T, repo string, pass int) {
+	t.Helper()
+	dvWriteFile(t, repo, "mpdrfapp/views.py", `from rest_framework import viewsets
+
+
+class WidgetViewSet(viewsets.ModelViewSet):
+    queryset = []
+`)
+	dvWriteFile(t, repo, "mpdrfapp/urls.py", fmt.Sprintf(`from rest_framework import routers
+
+from . import views
+
+router = routers.DefaultRouter()
+router.register(r"widgets", views.WidgetViewSet)
+
+urlpatterns = router.urls
+
+MP_DRF_PASS = %d
+`, pass))
+	dvWriteFile(t, repo, "mpdrfproj/urls.py", `from django.urls import include, path
+
+urlpatterns = [
+    path("api/", include("mpdrfapp.urls")),
+]
+`)
+}
+
+func mpDRFSameFile(t *testing.T, repo string, pass int) {
+	t.Helper()
+	dvWriteFile(t, repo, "mpsameapp/urls.py", fmt.Sprintf(`from rest_framework import routers, viewsets
+
+
+class WidgetViewSet(viewsets.ModelViewSet):
+    queryset = []
+
+
+router = routers.DefaultRouter()
+router.register(r"widgets", WidgetViewSet)
+
+urlpatterns = router.urls
+
+MP_SAME_PASS = %d
+`, pass))
+	dvWriteFile(t, repo, "mpsameproj/urls.py", `from django.urls import include, path
+
+urlpatterns = [
+    path("api/", include("mpsameapp.urls")),
+]
+`)
+}
+
+func TestMountParity_6461_DRFCoverage_PathA_BothLayouts(t *testing.T) {
+	cases := []struct {
+		label    string
+		write    func(*testing.T, string, int)
+		wantLost int // #6529, pre-existing, asserted not fixed
+	}{
+		{"ordinary (ViewSet in views.py)", mpDRFOrdinary, 0},
+		{"same-file (ViewSet in the edited urls.py)", mpDRFSameFile, 6},
+	}
+	for _, tc := range cases {
+		t.Run(tc.label, func(t *testing.T) {
+			repo := t.TempDir()
+			stateDir := t.TempDir()
+			tc.write(t, repo, 0)
+			dvFullRebuild(t, repo, stateDir)
+
+			endRepo := t.TempDir()
+			tc.write(t, endRepo, 1)
+			full := dvFullRebuild(t, endRepo, t.TempDir())
+
+			dvSeedManifest(t, repo, stateDir)
+			tc.write(t, repo, 1) // only the DRF urls.py content changes
+			inc := dvIncremental(t, repo, stateDir)
+
+			mpLogEndpointDelta(t, "DRF/"+tc.label, full, inc)
+			lost, invented := mpEndpointDelta(full, inc)
+
+			// THE ASSERTION THIS TEST EXISTS FOR. An INVENTED endpoint is one
+			// the daemon put in the graph that a full rebuild does not have —
+			// here, a composed ANY that DeduplicateNestedURLConfDRF removes on
+			// the full path. It must be zero in BOTH layouts.
+			if len(invented) != 0 {
+				t.Errorf("#6528 %s: the incremental graph INVENTED %d endpoint(s) a clean full "+
+					"rebuild does not have: %v. A composed ANY endpoint survives here only because "+
+					"the pass could not see the drf_router_expanded coverage that supersedes it — "+
+					"see drfUnknownCoveragePaths, which must abstain from composing under a "+
+					"router.register() prefix declared in a file that changed this tick.",
+					tc.label, len(invented), invented)
+			}
+			if len(lost) != tc.wantLost {
+				t.Errorf("#6529 %s: expected exactly %d LOST endpoint(s) (the pre-existing "+
+					"never-re-derived drf_router_expanded entities, identical with the composition "+
+					"pass disabled), got %d: %v. A DIFFERENT number means either that defect moved "+
+					"or the #6528 abstention over-reached and started suppressing live routes.",
+					tc.label, tc.wantLost, len(lost), lost)
+			}
+		})
+	}
+}
--- a/cmd/grafel/incremental_mount_parity_6461_test.go
+++ b/cmd/grafel/incremental_mount_parity_6461_test.go
@@ -263,11 +263,19 @@ func mpEndpointSet(d *graph.Document) []string {
 	return out
 }
 
-// mpLogEndpointDelta logs the endpoint census on both sides plus the symmetric
-// difference, so the measured entity delta is in the test output whether the
-// gate passes or fails.
-func mpLogEndpointDelta(t *testing.T, label string, full, inc *graph.Document) {
-	t.Helper()
+// mpEndpointDelta returns the symmetric difference of the two endpoint/mount
+// censuses: `lost` is present in the full rebuild and absent from the
+// incremental result, `invented` is the reverse.
+//
+// It is a FUNCTION rather than inlined into the logger because the census
+// delta is the #6461 GHOST property on its own — a stale composed endpoint
+// carried forward on an unchanged file shows up here as one LOST (the
+// recomposed path) plus one INVENTED (the pre-edit path), with no reference to
+// any other divergence class. Asserting it directly (see
+// TestMountParity_6461_Django_RoutePathRename_PathA_EndpointCensus) gives a
+// ratchet on the ghost that does not have to wait for the unrelated defects
+// the full parity comparator also sees on this fixture.
+func mpEndpointDelta(full, inc *graph.Document) (lost, invented []string) {
 	fa, ib := mpEndpointSet(full), mpEndpointSet(inc)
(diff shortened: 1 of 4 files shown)


cajasmota:
## Coordinator mutant round — one SURVIVOR, blocking

I scored two permissive mutants nobody else ran against the round-3 differential mechanism. Both `go vet`-clean before scoring, `-count=1`, restored by `cp` with shasum `6ba98546b…` verified after each.

| # | mutation | direction | result |
|---|---|---|---|
| P1 | `strippedReader` strips `router.register(...)` in **every** file, not only the changed ones | **permissive** | **SURVIVED** — `./internal/extractors/` ok, and the `Mount\|6461\|Django\|Incremental` arc of `./cmd/grafel/` ok |
| P2 | `composedRecordPath` admits **mount points** as well as nested paths | **permissive** | **KILLED** — `TestDRFUnknownComposedPaths_ScopedToTheRegisteringMount:468` |

### P1 is a genuine hole, not a contrived one

```go
strippedReader := func(relPath string) []byte {
	content := reader(relPath)
	if !registering[relPath] || len(content) == 0 {   // <- drop `!registering[relPath] ||`
		return content
	}
	return drfRouterRegisterRe.ReplaceAll(content, nil)
}
```

The `len(registering) == 0` early return still guards entry, so P1 only fires once **some** changed file registers. But from that point it strips **every** registration in the repo, so `survives` shrinks, `unknown` grows, and the pass abstains from pruning composed paths whose DRF coverage was never in question.

The direction is the one that matters: abstaining too much leaves the **#6461 ghost in place** — the exact defect this pass exists to remove. Reachable in any monorepo where two apps use DRF routers and you edit one of them: the other app's registrations get stripped too, its composed paths vanish from the counterfactual, and they are all marked unknown.

The commit message states the scoping — *"with every `router.register(...)` call site in the CHANGED files textually removed"* — and the doc comment repeats it. Nothing observes it. That is the same shape as the round-3 finding: a property asserted in prose on an axis no test covers.

### What round 4 needs

A test where a changed `urls.py` registers a router **and** an unchanged file elsewhere also registers one, asserting that the unchanged app's composed path is **not** in `unknown` and therefore prunes normally. Then re-score P1 — it must die.

P2 dying is good news and worth keeping as-is: the mount-point exclusion is pinned by the fixture's own composition assertion, so the doc comment's "excluded on purpose" is observed rather than merely claimed.

CI at `b148598e0` is green except the Windows leg, still in flight. Not merging until P1 is pinned.

cajasmota:
## Round-4 verified, and a second axis confirmed SURVIVING

**P1 is dead.** I re-scored it against `90c987b8c`: `go vet` clean first, `go test -count=1 ./internal/extractors/` in full, restored by `cp` with shasum `6ba98546bda6e96a69a14de0e28e9927d00d602a` verified. The new `dcTwoRouterRepo` fixture kills it, and the RED output is the point — `removedIDs=map[]` is the **#6461 ghost surviving verbatim on an app the edit never touched**. That is the defect this whole PR exists to remove, so the pin is load-bearing rather than decorative.

The round-4 diagnosis of why nothing saw it is also worth keeping: `dcScopeRepo` has exactly **one** registering file and it is **always** in the changed set, so "strip the changed files" and "strip every file" were byte-identical there. The only other exercised shape returns early at `len(registering) == 0` and never builds `strippedReader` at all. **Unobservable, not merely untested.**

## P3 — the breadth of the `registering` scan — SURVIVED

Round 4 flagged this and was right. I scored it:

```go
for _, rel := range changed {
	if !strings.HasSuffix(strings.ToLower(rel), "urls.py") {   // was: !isPythonPath(rel)
		continue
	}
```

`go vet` clean, `go test -count=1 ./internal/extractors/` → **exit 0, `ok … 55.581s`**. Restored by `cp`, shasum re-verified.

### It is the ghost-preserving direction after all, not the safe one

Round 4 rated this "lower severity — the restrictive direction (fewer abstentions, more pruning)". I do not think that reading survives contact with the early return.

Narrowing the scan does not merely abstain less. When the only registration in a changed set lives in `routers.py`, `registering` comes back **empty**, `drfUnknownComposedPaths` returns `nil` at the `len(registering) == 0` guard, and the pass proceeds to prune composed paths **whose DRF coverage this tick destroyed** — which is precisely the case the abstention was added to prevent. More pruning here is data loss, not conservatism.

And the doc comment claims the breadth explicitly:

> *Note that the scan reads every changed `.py`, not only `*urls.py` — a registration can live in `routers.py` or any module an `include()` reaches.*

Every fixture in the file puts the registration in a `urls.py`. So this is the same shape as P1 and as round 3: **prose asserting a breadth no fixture exercises.** Third occurrence on this PR, which is itself the finding.

## What round 5 needs

One fixture where the `router.register(...)` lives in a **non-`urls.py`** module (`routers.py`) reachable from a changed set, asserting the composed path is in `unknown` and therefore abstains. Then re-score P3 — it must die.

Test-only again, as round 4 was. No production behaviour should change; if pinning it shows the breadth is actually wrong, stop and say so rather than narrowing the scan.

PR #6533, opened 2026-08-21
fix(#6507): the max_size gate lstatted, so a symlink bypassed the cap and then mislabelled itself line_too_long

### LINKED ISSUE #6505: secrets: a file with one >64KB line reports a real credential as clean — bufio.ErrTooLong discards the findings AND the skip
Found while reviewing #6504. **A file containing a real credential is reported as clean, with no finding and no skip**, if it also contains one line longer than 64 KB.

## Reproduction

Same file scanned twice, differing only by the presence of one long line:

```
PROBE2         findings=0 skipped=[]   // AKIAIOSFODNN7REAL0000 on line 1, plus one 64KB+ line, 65587 bytes total
PROBE2-control findings=1              // identical key, no long line
```

## Mechanism

`scanFile` returns `(findings, scanner.Err())`. When `bufio.Scanner` hits `bufio.ErrTooLong`, the walk's `if err != nil` branch drops `ff` **wholesale** — including findings already collected from earlier lines — and `classifyScanSkip` does not recognise the error, so it is not reported as a skip either.

The result is the worst combination available: a real secret is silently discarded, and the caller receives an unqualified clean.

## Why this is reachable rather than theoretical

- `bufio.MaxScanTokenSize` is **64 KB**, while the scanner's own file cap is **512 KB** — so there is a wide band of files that pass the size gate and then fail mid-scan.
- `skipExtSet` denies **no** `.json`, `.lock`, `.map`, or minified `.js`. Those are precisely the files that carry >64 KB lines in real repositories.

So the affected population is not exotic — it is lockfiles, source maps, bundled JS and large JSON fixtures, which are exactly the places a leaked key tends to end up.

This is more reachable than the FIFO case in #6483, which was filed for the same class of defect.

## A second, related silent path

`filepath.WalkDir`'s error callback is `return nil // skip unreadable entries`. A directory the scanner cannot read is dropped whole:

```
PROBE3 err=<nil> findings=0 skipped=[]   // sub/ chmod 000, containing a real AWS key
```

#6483's stated exclusion for ordinary `ENOENT`/`EACCES` is argued for **files**. One unreadable *directory* is N unread files, and the caller still gets a clean.

## What a fix needs to decide

Two independent questions, and the second is easy to miss:

1. **Should `ErrTooLong` be a reported skip?** Almost certainly yes — #6504 adds the `Skipped` channel that makes this expressible.
2. **Should findings collected before the error be kept?** Dropping `ff` on a scanner error is a separate decision from not reporting the skip. Keeping them would have surfaced the key in the probe above even with the skip unreported. Whichever way this goes, it should be stated rather than incidental.

Also worth a stated reason: the `WalkDir` error callback currently explains nothing.

## Observability

Go-level; the golden harness cannot express this (`ScanPath` is not an extractor and produces no graph entities).

**Killing mutant:** revert whichever handling lands — the test must go red on a fixture with a real-looking key plus one >64 KB line, asserting either a finding or a skip naming that file. The control case (same key, no long line) must stay green so the test cannot pass vacuously.

Note this needs **no build tag and no syscall** — just `os.WriteFile` with a long line — so unlike the FIFO fixtures it gives Windows CI real coverage.

## Status

Being addressed in #6504 round 2. Filed separately so it stays tracked if that round chooses to document the behaviour rather than change it — the defect exists on `main` today either way.

### LINKED ISSUE #6507: secrets: the max_size gate lstats, so a symlink to an oversized file bypasses the cap — and now reports line_too_long, a limit the caller cannot raise
Found while reviewing #6504. **Pre-existing — the bypass predates that PR**; only the mislabel is new, because the skip vocabulary is new.

## The bypass

`ScanPath`'s size gate reads `d.Info()`, which is an **lstat**. For a symlink that returns the *link's* size, not the target's. The file is then opened through `safeio.Open(FollowSymlinks)` and read in full.

So a symlink pointing at an oversized file **passes a cap it should fail**.

Probed: symlink → 70 KB target, `max_size=1024` → the target is fully read, one finding is reported, and the skip says `reason=line_too_long`.

## Why the reported reason is now actively misleading

#6504 defines a deliberate distinction:

- `too_large` — never opened, because of a cap the **caller** chose. Raise `max_size` and it goes away.
- `line_too_long` — opened, partly read, stopped at `bufio.MaxScanTokenSize`, **a hard limit no caller can raise.**

The symlink case reports `line_too_long` for a file the caller's `max_size` *would* have excluded. A caller reading that reason concludes nothing can be done, when raising the cap is exactly the remedy — or, read the other way, that the file was within budget when it was not.

The direct (non-symlink) path is correct: the cap is evaluated in the walk before `scanFile` is called, so precedence is deterministic and size-gate-first. Probed on a file that is both oversized and long-lined:

| cap | reason | findings |
|---|---|---|
| 1024 | `too_large` | 0 |
| default 512 KB | `line_too_long` | 1 |

Only the symlink path diverges.

## What a fix needs to decide

1. **Should the gate stat the target?** Switching `d.Info()` to an `os.Stat` on the resolved path would make the cap mean what a caller expects. Cost: one extra stat per symlinked entry.
2. **Or should symlinks be skipped outright**, with their own reason? That is a policy question — the scanner currently follows them deliberately.
3. Either way, **the reported reason must not claim an unraisable limit for a file the caller's cap would have excluded.**

Worth checking whether the same lstat-vs-stat asymmetry affects any other size or type gate on this path.

## Observability

Go-level; the golden harness cannot express this (`ScanPath` is not an extractor).

**Killing mutant:** revert whichever gate change lands — a test with a symlink to an oversized target under a small `max_size` must go red. Note the assertion should check the **reason**, not just that a skip exists, since the current behaviour does produce a skip; it is the *wrong* one.

⚠️ Symlink creation needs Developer Mode or elevation on Windows and this repo runs `windows-latest`, so this test needs a `//go:build unix` tag or a `t.Skipf` guard — the same pattern the FIFO tests already use.

cajasmota:
Closes #6507
Closes #6505

Two issues, one file region. **Only one of them needed a behavioural change** — that is the main thing to read here.

## #6507 — real, reproduced, fixed

`ScanPath`'s size gate read `d.Info()`, an **lstat**. For a symlink that returns the *link's* size, not the target's, and the file is then opened through `safeio.Open(FollowSymlinks)` and read in full. So a symlink pointing at an oversized file passed a cap it should have failed — and then reported `line_too_long`, a limit **no caller can raise**, for a file the caller's own `max_size` would have excluded.

Probed on `main` before writing anything:

```
PROBE6507  max_size=1024, symlink -> ~70KB target
           findings=1  skipped=[{Rel:link.go Reason:line_too_long}]
```

RED:

```
--- FAIL: TestScanPathSizeGateStatsSymlinkTarget
    got 1 findings, want 0
    reason = "line_too_long", want "too_large"
    wire reason = "line_too_long", want "too_large"
```

GREEN: `ok internal/secrets 0.584s`.

### The two decisions the issue asked to be stated rather than left incidental

**Stat the target; do not skip symlinks.** The scanner follows them deliberately (`safeio.FollowSymlinks`), and monorepo/vendor link farms are the ordinary case — refusing symlinks would trade a size-gate bug for a coverage hole, which is the same silent-clean class #6483 was filed for. The extra stat is paid **only** on symlinked entries (`info.Mode()&os.ModeSymlink != 0`), so the common path costs nothing. `info` is deliberately not reassigned: its `Mode()` feeds `classifyScanSkip`, which documents that it receives the **walk's lstat** mode.

**A dangling target stays silent, with no skip.** Same class as the vanished-entry case immediately above it, and `safeio.Open` would fail on it moments later anyway.

## #6505 — the code was already correct on `main`. The guard was not.

The headline defect does not reproduce:

```
PROBE6505  err=<nil>  findings=1  skipped=[{Rel:creds.go Reason:line_too_long}]
```

#6504 already fixed it, says so in its body, and #6505 itself anticipated this outcome — it was filed precisely so the item stayed tracked if that round chose to document rather than change. Both decisions #6505 demanded be stated *are* stated in code.

Its third item — the `WalkDir` error callback that "explains nothing" — now carries a rationale on `main` treating the EACCES-directory hole as a **deliberate, documented exclusion with a named follow-up**. Deliberately not reversed here: that is a policy call a merged PR made explicitly, and the probe confirms the behaviour matches the comment (`err=<nil> findings=0 skipped=[]`) rather than having regressed. It does not belong inside a size-gate fix.

So what this PR adds for #6505 is **not** a behavioural change — it is the missing guard, which the mutant round found.

## Mutants

`go vet` first on every one (a mutant that does not compile proves nothing), `-count=1` (a cached green is not evidence), restored by `cp` with a verified shasum — never `git checkout -- <file>`.

|---|---|---|
| #6507's killing mutant — revert the gate to the lstat size | — | **DEAD** |
| skip every symlink as `too_large` | **permissive** | **DEAD** (4 tests, incl. the pre-existing `TestScanPathNamesFifoKindThroughSymlink`) |
| report a dangling target as `too_large` | **permissive** | **DEAD** |
| #6505's killing mutant — drop `ErrTooLong` partial findings | — | **DEAD** |
| #6505's killing mutant — disable the `line_too_long` classify arm | — | **DEAD** |
| keep findings on **every** `scanFile` error (the PREDICATE, `keepPartialFindings` -> `return true`) | **permissive** | **DEAD** |
| keep findings on **every** `scanFile` error (the CALL SITE, `if keepPartialFindings(err)` -> `if true`) | **permissive** | **SURVIVED — equivalent under the suite, not dead** |
| symlink -> **directory** answered `too_large` (round 2, the guard this PR now adds) | — | **DEAD** |
| never gate a symlinked entry (round 2) | **permissive** | **DEAD** |
| `IsRegular()` -> `!IsDir()` (round 2) | **permissive** | **SURVIVED — equivalent, see below** |

### The survivor was a genuine gap — and it is still open at the call site

Widening the keep to every error survived the **entire package**. The two `safeio` sentinels come from a failed *open*, so `ff` is nil and appending it is a no-op — the widening is structurally invisible end to end. The one arm where it changes an answer is a **mid-read I/O error**, which can leave `ff` non-empty while the bytes *and every line number in them* are in doubt. No portable `os`/`filepath` fixture provokes that, so the distinction was pinned nowhere.

Extracted to `keepPartialFindings(err error) bool` and asserted directly — the remedy `TestClassifyScanSkipReportsWouldBlock` already applies to its own unreachable arm, so this follows the package's precedent rather than inventing a seam. Both directions of the **predicate** now die (`return true` and `return false`).

**Correction (round 2).** An earlier revision of this table scored the widening `SURVIVED -> pinned -> DEAD`. That was wrong, and the review was right to call it: the extraction pinned the **function**, not the **behaviour**. Re-verified at the call site,

```
internal/secrets/secrets.go:584
-  if keepPartialFindings(err) {
+  if true {
```

`go vet` 0, `go test -count=1 ./internal/secrets/` **exit 0, ok — SURVIVES**. The mutant is **equivalent under this suite, not dead**, and it is recorded that way rather than rounded down.

It is equivalent for precisely the reason the predicate had to be extracted: the only error that separates the two branches is a **mid-read I/O failure**, and nothing reachable through `os`/`filepath` produces one. Two ways to force the difference were considered and rejected — a fabricated I/O-error fixture, and a package-level indirection over `scanFile` injected purely so a test could return an error the real path cannot. Both **manufacture** the difference instead of observing it, and the second adds exactly the production machinery whose own removal is unfalsifiable that the `NOTE (#6416)` in this same walk rules out.

`keepPartialFindings`' doc comment and `TestKeepPartialFindingsIsExclusiveToErrTooLong` now both state, in as many words, that the call site is unpinned — so no later reader over-reads the coverage.

## Round 2 — a defect this PR introduced, on the platform CI runs

`os.Stat` on a symlink to a **directory** answers with the directory's own size. That is a filesystem artefact, not a payload: **64 bytes on APFS**, so nothing changes locally, but **≥ 4096 on ext4** and larger still for a directory with many entries.

So any caller cap below that made the new gate fire and answer **`too_large` with `Kind` empty** — telling the caller to raise `max_size` for an entry no cap will ever make scannable. That is the **same wrong-bucket class #6507 was filed about**, reintroduced on the path #6507's own fix added.

**Fix:** the gate is guarded by `sizeInfo.Mode().IsRegular()`, and a non-regular target **falls through** rather than being skipped at the gate. Falling through keeps the naming with the single authority that already owns it — `safeio.Open` refuses the entry and `classifyScanSkip` reports `not_regular` with `Kind=directory`. An early skip at the gate would have to re-derive that vocabulary, and would reintroduce the duplicate entry-type gate `NOTE (#6416)` explicitly rules out.

RED (on APFS, where the directory is only 64 bytes — the cap is **1**, the one value guaranteed to sit below every filesystem's directory size):

```
--- FAIL: TestScanPathSymlinkToDirectoryIsNotTooLarge/tiny_cap
    reason = "too_large", want "not_regular"
    kind   = "",          want "directory"
```

GREEN: `ok internal/secrets 0.620s`.

The test asserts the **reason and the kind, never a byte count** — an assertion written against a directory's size would pass on APFS by accident and prove nothing about the platform where the bug bites. Its second subtest runs the identical fixture under the **default** cap, where old and new code both answer `not_regular`, so what is pinned is *the reason must not depend on the cap*, not one number.

**The `!IsDir()` survivor** is equivalent for a stateable reason rather than a missing test: every non-regular, non-directory entry type reports `st_size` 0 — FIFOs, sockets, character and block devices alike — so the gate can never fire on one whatever the cap. `IsRegular()` is still the right predicate to write: it says what the gate **means** (bytes that will actually be read) instead of enumerating the one type observed to break.

## Portability

`size_gate_symlink_6507_unix_test.go` is `//go:build unix` **plus** a `t.Skipf` guard — symlink creation needs elevation on `windows-latest`, the same pattern the FIFO tests use. `partial_findings_6505_test.go` has **no build tag and no syscall**, so Windows CI gets real coverage from it. No FIFOs created outside `t.TempDir()`.

## Verification

`gofmt -l` clean · `go build ./...` 0 · `go vet` (secrets, mcp, dashboard, cmd) 0 · `go test -count=1` on **whole packages, no `-run` filter**: `./internal/secrets/` 0, `./internal/mcp/` 0 (128s), `./internal/dashboard/` 0 (37s). Exit codes captured separately from the output.

### DIFF
--- a/internal/secrets/partial_findings_6505_test.go
+++ b/internal/secrets/partial_findings_6505_test.go
@@ -1,0 +1,81 @@
+// Tests for #6505's second decision: which scanFile errors may keep the
+// findings collected before the failure.
+//
+// No build tag and no syscall — this is pure predicate coverage, so Windows CI
+// exercises it too, alongside the end-to-end TestScanPathReportsOverlongLineAsSkip.
+package secrets
+
+import (
+	"bufio"
+	"errors"
+	"fmt"
+	"io"
+	"io/fs"
+	"os"
+	"testing"
+
+	"github.com/cajasmota/grafel/internal/safeio"
+)
+
+// TestKeepPartialFindingsIsExclusiveToErrTooLong pins the PERMISSIVE direction
+// of the predicate, which no end-to-end test can see.
+//
+// Widening it to `return true` is invisible through ScanPath: the two safeio
+// sentinels come from a failed OPEN, so scanFile returned a nil slice and
+// appending it is a no-op. The full package suite stays green under that mutant
+// — verified — which is exactly why the decision is asserted here rather than
+// left to the walk's comment.
+//
+// The mid-read I/O arm is the one where the widening would actually change an
+// answer: a truncated or erroring read can leave ff non-empty while the byte
+// stream, and therefore every line number in it, is in doubt. No portable
+// fixture provokes that through os/filepath, so it is asserted at the level the
+// decision is made.
+func TestKeepPartialFindingsIsExclusiveToErrTooLong(t *testing.T) {
+	cases := []struct {
+		name string
+		err  error
+		want bool
+	}{
+		{"bufio.ErrTooLong", bufio.ErrTooLong, true},
+		{"wrapped ErrTooLong", fmt.Errorf("scan %s: %w", "creds.go", bufio.ErrTooLong), true},
+
+		// Failed opens: ff is empty, so keeping it is a no-op end-to-end.
+		// Asserted anyway — a `return true` mutant must die somewhere, and the
+		// invariant being protected is "only a partial READ keeps findings",
+		// not "only cases where it happens to matter today".
+		{"safeio.ErrNotRegular", safeio.ErrNotRegular, false},
+		{"safeio.ErrWouldBlock", safeio.ErrWouldBlock, false},
+		{"wrapped ErrNotRegular", fmt.Errorf("open: %w", safeio.ErrNotRegular), false},
+
+		// Mid-read failures: ff CAN be non-empty and the line numbers are
+		// untrustworthy, so these must drop it.
+		{"io.ErrUnexpectedEOF", io.ErrUnexpectedEOF, false},
+		{"fs.ErrPermission", fs.ErrPermission, false},
+		{"os.ErrClosed", os.ErrClosed, false},
+		{"generic I/O error", errors.New("input/output error"), false},
+
+		// bufio's other sentinel: a negative advance is a scanner bug, not a
+		// clean partial read.
+		{"bufio.ErrAdvanceTooFar", bufio.ErrAdvanceTooFar, false},
+	}
+
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
+			if got := keepPartialFindings(tc.err); got != tc.want {
+				t.Errorf("keepPartialFindings(%v) = %v, want %v: only a PARTIAL "+
+					"READ leaves the collected findings and their line numbers "+
+					"trustworthy", tc.err, got, tc.want)
+			}
+		})
+	}
+}
+
+// TestKeepPartialFindingsRejectsNil guards the degenerate input. The walk only
+// calls it inside `if err != nil`, but a predicate that answered true for nil
+// would silently invert its own contract if that guard ever moved.
+func TestKeepPartialFindingsRejectsNil(t *testing.T) {
+	if keepPartialFindings(nil) {
+		t.Error("keepPartialFindings(nil) = true; nil is not a partial read")
+	}
+}
--- a/internal/secrets/partial_findings_6505_test.go
+++ b/internal/secrets/partial_findings_6505_test.go
@@ -31,6 +31,20 @@ import (
 // stream, and therefore every line number in it, is in doubt. No portable
 // fixture provokes that through os/filepath, so it is asserted at the level the
 // decision is made.
+//
+// WHAT THIS DOES NOT PIN, stated plainly so no later reader over-reads it.
+// This test pins the FUNCTION. The CALL SITE remains unpinned: mutating
+// `if keepPartialFindings(err)` in ScanPath to `if true` vets clean and leaves
+// the whole package green — re-verified. That mutant is EQUIVALENT UNDER THIS
+// SUITE, not dead, and it is equivalent for the same reason the predicate had
+// to be extracted: the only error that would distinguish the two is a mid-read
+// I/O failure, and nothing reachable from os/filepath produces one. Rejected
+// alternatives, both of which would have manufactured the difference rather
+// than observed it: a fabricated I/O-error fixture, and a package-level
+// indirection over scanFile injected purely so a test could return an error the
+// real code path cannot. The second would add exactly the kind of production
+// machinery whose own removal is unfalsifiable that the NOTE (#6416) in
+// ScanPath rules out. It is reported as a survivor rather than rounded down.
 func TestKeepPartialFindingsIsExclusiveToErrTooLong(t *testing.T) {
 	cases := []struct {
 		name string
--- a/internal/secrets/secrets.go
+++ b/internal/secrets/secrets.go
@@ -552,7 +552,23 @@ func ScanPath(root string, maxFileBytes int64) (ScanResult, error) {
 			sizeInfo = ti
 		}
 
-		if sizeInfo.Size() > maxFileBytes {
+		// IsRegular, because the gate weighs BYTES THAT WILL BE READ and a
+		// non-regular target has none. os.Stat on a symlink to a DIRECTORY
+		// answers with the directory's own size — a filesystem artefact, 64
+		// bytes on APFS but at least 4096 on ext4 and larger for a directory
+		// with many entries — which meets any modest cap. Without this guard
+		// the entry came back too_large with Kind empty, telling the caller to
+		// raise max_size for something no cap will ever make scannable: the
+		// same wrong-bucket defect #6507 is about, on the path #6507's own fix
+		// introduced.
+		//
+		// It FALLS THROUGH rather than skipping here, so the naming stays with
+		// the single authority that already owns it: safeio.Open refuses the
+		// entry and classifyScanSkip reports not_regular with Kind=directory.
+		// An early skip at this site would have to re-derive that vocabulary,
+		// and would reintroduce exactly the duplicate entry-type gate the
+		// NOTE (#6416) above rules out.
+		if sizeInfo.Mode().IsRegular() && sizeInfo.Size() > maxFileBytes {
 			skipped = append(skipped, Skip{Path: path, Rel: rel, Reason: SkipTooLarge})
 			return nil
 		}
@@ -620,6 +636,13 @@ func ScanPath(root string, maxFileBytes int64) (ScanResult, error) {
 // fixture can provoke a mid-read I/O error through os/filepath, so the
 // distinction is unfalsifiable end-to-end and would otherwise be pinned
 // nowhere.
+//
+// Extraction moved the assertion; it did not close the gap. The CALL SITE above
+// is still unpinned — replacing `if keepPartialFindings(err)` with `if true`
+// leaves the package green — because the one error that would tell the two
+// apart cannot be reached from a test. That mutant is equivalent under the
+// suite, not dead, and TestKeepPartialFindingsIsExclusiveToErrTooLong says so
+// in as many words.
 func keepPartialFindings(err error) bool {
 	return errors.Is(err, bufio.ErrTooLong)
 }
--- a/internal/secrets/secrets.go
+++ b/internal/secrets/secrets.go
@@ -513,7 +513,46 @@ func ScanPath(root string, maxFileBytes int64) (ScanResult, error) {
 			// tree; not a skip worth reporting.
 			return nil
 		}
-		if info.Size() > maxFileBytes {
+
+		// The gate must weigh the bytes that will actually be READ (#6507).
+		//
+		// d.Info() is an LSTAT: for a symlink it reports the link's own size,
+		// a few dozen bytes, never the target's. scanFile then opens with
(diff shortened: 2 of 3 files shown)


cajasmota:
Independent adversarial review came back **REQUEST-CHANGES**, and I re-scored its central claim myself before acting on it. **No behavioural defect in the shipped fix** — but one claim in the PR body is false as written, and the review found a real defect this PR introduces on the platform CI actually runs.

## The headline mutant claim does not hold — confirmed independently

The table says *"keep findings on EVERY `scanFile` error (permissive) — SURVIVED → pinned → DEAD"*. The extraction pinned the **function body**, not the **behaviour**. Mutating the call site instead:

```
internal/secrets/secrets.go:584
-  if keepPartialFindings(err) {
+  if true {
```

My own run: `go vet ./internal/secrets/` → **0**, `go test -count=1 ./internal/secrets/` → **EXIT=0, `ok … 1.157s`**. Restored by `cp`, shasum `989c4ebf375ea0af286fb773860890e039c2000e` verified.

**SURVIVES.** Same semantics as the mutant the PR claims to have killed, one call frame out.

So what the new test pins is that a helper named `keepPartialFindings` returns `false` for non-`ErrTooLong` inputs — which the call site is free to ignore. The widening itself remains unpinned end to end.

No code change is required for this: the shipped behaviour is correct, and the PR's own reasoning explains why it stays invisible — the `safeio` sentinels make it a no-op today, and a mid-read I/O error is not portably provokable. **What is required is that the record stop calling it DEAD.** It is *equivalent under the current suite at the call site*, and this milestone's rule is that permissive survivors get reported, not rounded down. The PR body is the permanent record.

## A real defect, on Linux, of exactly the class #6507 was filed about

`os.Stat` on a **symlink→directory** returns the directory's own size, which then meets the cap.

On macOS/APFS a directory is 64 bytes, so nothing changes locally and every probe here passes. **On ext4 a directory is ≥4096** and a grown one can be much larger. So on Linux, any caller cap between the link's path length and the directory size flips the reason to `too_large` with `Kind` empty — telling the caller *"raise `max_size`"* for something that raising the cap will never scan.

That is the same wrong-bucket failure #6507 exists to fix, re-created on a path this PR introduced. Going back for a guard, with the test asserting the **reason and Kind rather than byte counts**, so it cannot pass vacuously on APFS.

## What the review re-probed and confirmed — not being churned

- **The #6505 "no behavioural change needed" claim holds**, verified on the merge-base `f95959323` rather than the PR head: `PROBE6505 err=<nil> findings=1 skipped=[{creds.go line_too_long}]`, control `findings=1`. The third item reproduces as still-open too — `PROBE6505-dir err=<nil> findings=0 skipped=[]` on a `chmod 000` directory holding a real key, which is now tracked as **#6534**.
- **Size-gate precedence is deterministic on both paths.** Same fixture, oversized *and* >64KB line: direct cap=1024 → `too_large`/0 findings, cap=512K → `line_too_long`/1; symlink cap=1024 → `too_large`/0, cap=512K → `line_too_long`/1.
- **`keepPartialFindings` is textually equivalent to what it replaced and non-vacuous** — both `return true` and `return false` die.
- **The `info`/`sizeInfo` split produces no wrong `Kind`** across symlink→dir, symlink→fifo, symlink→symlink→70KB, relative `../` links and symlink loops, at caps 1/16/1024/512K.

## On my own calibration mutant

I had scored one permissive mutant myself — letting a dangling symlink target fall through and be scanned instead of returning early — which survived, and I judged it **equivalent** rather than a gap. The review traced the path and agrees: `safeio.Stat`'s `os.Stat` returns ENOENT **raw** (`safeio.go:142-144`), not wrapped in `ErrNotRegular`, so `reportSecretScanSkip` returns early at its sentinel guard (`secrets.go:726`), `classifyScanSkip` matches no arm, `keepPartialFindings(ENOENT)` is false, and `ff` is nil. Findings, skips and stderr are byte-identical; the early return's only value is one saved open.

It found a second near-equivalent of its own — widening the re-stat trigger from `ModeSymlink` to `ModeType` — which also survives and is also genuinely equivalent, since a direct FIFO stats to size 0 and lands in the same bucket.

Issue #6535, opened 2026-08-21
Feedback report: MMF1App (vbnet+csharp) — Schema 90% / Operation 60% orphan, 0% field extraction


dcastro-imp:
### Anonymization verification

- [x] I have verified this report contains no real source code, file paths, or identifier names (all entity names look like `ent_a3f7` or `op_92c1`; all paths look like `<go>/<seg-1>/<seg-2>.go`).
- [ ] I generated this report with `grafel feedback` v0.9+ (generated with v0.3.0)

### Feedback report

_Pasted below rather than attached — filing via `gh` CLI, not the web UI drag-and-drop, so a file attachment wasn't available to me. Full contents, not truncated._

```markdown
# grafel feedback report

Generated: 2026-08-21T21:25:18Z
grafel version: v0.3.0 (commit b037c3f, built 2026-08-21T18:24:57Z)
Group profile: 2 language(s) (vbnet, csharp), 45000-46000 entities, 121000-122000 relationships
Confidence: 90% (9/10 sanity checks passed)

## 1. Extractor Coverage

### Entities by language

| Language | Entity count (range) |
|---|---|
| csharp | 100+ |
| vbnet | 100+ |

### Entity kind distribution

| Kind | Language | Count (range) |
|---|---|---|
| SCOPE.Component | csharp | 21-100 |
| SCOPE.Operation | csharp | 100+ |
| SCOPE.Schema | csharp | 100+ |
| SCOPE.Component | vbnet | 100+ |
| SCOPE.Operation | vbnet | 100+ |
| SCOPE.Process | vbnet | 21-100 |
| SCOPE.Schema | vbnet | 100+ |

### Source-window completeness

Entities with valid start/end line: **99.0%** (45192 of 45663)

### Annotation coverage

Framework-annotated entities: **0.0%** (0 of 45663)

### Field extraction rate

Class/Model entities with zero fields: **100.0%** (253 total class-like entities)

## 2. Orphan Rate

An entity is orphan when it has no semantic edge in EITHER direction (CONTAINS/DECLARES excluded, in both directions). The table below counts only orphans of kinds that carry a semantic edge SOMEWHERE in this group; kinds where no entity does are listed separately under **Expected/terminal orphans**, so a kind reading 0 here may still be entirely unwired there.

| Kind | Total | Orphan | Orphan % |
|---|---|---|---|
| Module | 85 | 1 | 1.2% |
| SCOPE.Component | 2618 | 1003 | 38.3% |
[...]

PR #6541, opened 2026-08-21
fix(#6536): the field-extraction metric never sampled a class, so a working subsystem reported 100% failure

### LINKED ISSUE #6536: the field-extraction metric never samples a class: classLikeKindTails omits 'component', so the C#/VB.NET 100% zero-field rate is guaranteed by construction
Found while grounding the feedback report in #6535. **The reported metric is vacuous by construction — it has never sampled a real class in any C#-family codebase**, and the 100% figure it produced was guaranteed before any code was extracted.

## The metric

`internal/feedback/report.go:592-603` defines `classLikeKindTails` as exactly `class`, `struct`, `model`, `schema`, matched against the lower-cased, last-dot-stripped kind tail (`kindTail`, `report.go:580-586`). The candidate filter is `report.go:277`: `isClassLikeKind(kind) && e.Subtype != "field"`.

## The extractor side does not use those tails

VB.NET's kinds (`internal/extractors/vbnet/extractor.go:240-260`), and per `extractor.go:234-238` C# is the same:

| declaration | Kind | tail | class-like? |
|---|---|---|---|
| Class / Module / Structure / Interface | `SCOPE.Component` | `component` | **no** |
| Delegate | `SCOPE.Component` | `component` | **no** |
| Enum | `SCOPE.Schema` | `schema` | yes |
| Const | `SCOPE.Schema` | `schema` | yes |
| Field | `SCOPE.Schema` (subtype `field`) | `schema` | excluded at `:277` |
| Method / Property / Event | `SCOPE.Operation` | `operation` | no |

`component` is **not** in `classLikeKindTails`. So every class, structure, module and interface is invisible to this metric.

## What the 100% actually measured

In #6535's report the denominator was **253** class-like entities against **2,618** `SCOPE.Component` and 45,663 entities total. Those 253 are the `SCOPE.Schema` residue after the `field` subtype is removed — i.e. **enums and consts**, which structurally cannot own field children.

**A 100.0% zero-field rate on that population is guaranteed, not measured.** The real classes were never sampled, and their fields — which *are* extracted and correctly attached via `BuildSchemaFieldStructuralRef` (`internal/extractors/vbnet/extractor.go:219-228`, `:275`) — are simply uncounted.

This is worse than a wrong number. It is a metric that reports a **catastrophic-looking 100% failure** for a subsystem that is working, which is exactly the signal that sends a maintainer to fix code that has no defect. A contributor acted on it in good faith.

## What a fix needs to decide

1. **Add `component` to `classLikeKindTails`** — one line, and it makes the metric sample real classes for the whole C# family.
2. **Decide whether enums and consts should be exempt** rather than counted as zero-field failures. They cannot own fields, so leaving them in keeps a permanent floor of false failures under the metric.
3. **Audit the other kind-tail lists in `internal/feedback/` for the same asymmetry.** This one was found by accident because a contributor's report made it visible; the population of similar lists is unmeasured, and each is a metric that can be silently vacuous.

## Observability

Go level, `internal/feedback` — its tests already build a `Report` literal directly, so this needs no fixture and no corpus.

**Killing mutant:** remove `component` again. A test asserting a `SCOPE.Component` with field children reports a **non-**100% zero-field rate must go red. Today no test observes the tail list against any extractor's actual kind strings, which is the gap that let this ship.

Refs #6535

cajasmota:
Closes #6536

Found while grounding a contributor's feedback report (#6535), where the metric announced **"100% of 253 class-like entities have zero fields"** on a codebase whose fields were being extracted and attached correctly the entire time.

## The mechanism, verified rather than assumed

`entityKind` (`internal/extractors/vbnet/extractor.go:240-262`) returns `SCOPE.Component` for Class / Module / Structure / Interface / Delegate, and `SCOPE.Schema` for enum / const / field. `classLikeKindTails` admitted only `class`, `struct`, `model`, `schema`.

So **no real class was ever sampled**. The 253 in that report were the `SCOPE.Schema` residue after field leaves were filtered — enums and consts, which structurally cannot own field children. A 100% zero-field rate on that population was guaranteed before any source was parsed.

The new test does not hard-code the kind string; it asks the extractor. The RED output is the extractor telling the test what it emits:

```
--- FAIL: TestClassLikeKindTailsAdmitsExtractorTypeKinds_6536
    type Widget is emitted as kind "SCOPE.Component", which isClassLikeKind rejects:
    the field-extraction metric cannot sample it
--- FAIL: TestFieldExtractionSamplesRealClasses_6536
    zero-field rate is 100% on a graph whose Widget owns 2 fields and Box owns 1: ClassTotal=2
--- FAIL: TestEnumAndConstExemptFromFieldDenominator_6536
    ClassTotal = 2 on an enum/const-only graph, want 0 (zero-field rate 100)
```

That third line reproduces #6535's symptom in miniature: **100% on a population that cannot pass.** GREEN: `ok internal/feedback 0.259s`, whole package, no `-run` filter.

## A second instance of the same defect, from the other side

**Adding `component` alone re-creates the bug in the opposite direction.** Every extractor's per-file carrier is a `SCOPE.Component` (`internal/extractor.FileEntity`, `extractor.go:326-340`), so the metric would have enrolled **one guaranteed-zero-field entity per indexed file** — a denominator that grows with repo size and can only drag the rate toward 100%.

This was not in the issue's analysis. It was caught by the test written for the *first* half (`ClassTotal=3` where 2 was right), which is the argument for building the observation before the fix rather than after.

## The enum/const decision, stated in code

**Exempt.** Recorded as `nonClassSubtypes` (`report.go:610-635`), replacing the bare `Subtype != "field"` at the candidate filter.

Reason, checked per emitter rather than asserted: an enum's members are enum members and a const is a single value, and **no extractor emitting those subtypes — vbnet, csharp, cpp, php, proto, avro, solidity — ever emits a `field` child under one**. Counting them is a permanent floor of false failures.

## Round 2 — the carrier exemption leaked, and the PR's own mutant table was wrong

Two defects were found by adversarial review and reproduced against real extractor output.

**(a) The exemption was one population short — three, in fact.** Round 1 applied the rule *"a population that cannot pass must not be in the denominator"* to `Subtype: "file"` alone. Still counted:

- **`Subtype: "import"` placeholders.** `cross/imports` emits one `SCOPE.Component`/`import` per imported module for C#, Java, Go, Python, Ruby, Rust, JS/TS and Elixir; eighteen more extractor packages emit the same pair. That is one to several guaranteed-zero-field entities **per file** — a population that outnumbers classes in most repos. Measured, not argued: a 3-`using` C# file with one class gave `ClassTotal = 2 … zero-field rate 100 is guaranteed, not measured`.
- **The ML-family per-file `Subtype: "module"` carriers** — a file carrier under another name (haskell, elm, ocaml, fsharp, idris, reasonml, rescript, crystal, erlang).
- **Haskell's import placeholders**, which set no `Subtype` at all, so no subtype exclusion could ever see them.

**(b) The tail set was unconstrained from above.** No fixture in the package carried a kind outside `SCOPE.{Class,Component,Schema,Operation,Function}`, so adding `"external": true` to `classLikeKindTails` vetted clean and left the package green. `SCOPE.External` is the sresolver/swift external-dependency placeholder — admitting it is exactly the mistake #6536 is about. It is now pinned.

### Two proposed exemptions were narrowed after checking the extractors

Both would have deleted a population that genuinely **passes**, which biases the rate the same way counting one that cannot does.

- **`"module"` is NOT globally exempt.** `Public Module Util` emits `SCOPE.Component`/`module` and `CONTAINS` its `Shared`/`Private` fields — a real field-bearing container in exactly the C#-family population #6535 is about. The ML-family carriers are exempted **by language** instead: those nine extractor packages emit no `Subtype "field"` anywhere, so every entity they produce is guaranteed zero-field.
- **`"interface"` is exempt in every language EXCEPT Kotlin.** VB.NET/C# forbid interface fields (probe: `IThing` contains only its method); a Java interface's constants come out as a `SCOPE.Enum` group, not field children; Go interfaces hold only method specs; JS/TS emits no `field` subtype at all. But `interface Shape { val sides: Int }` really does emit `Shape` CONTAINS `Shape.sides` subtype `field`.

### Final exemption set (`isFieldExtractionCandidate`)

| exemption | why it cannot own a field child |
|---|---|
| subtype `field` | the field LEAF is the child, not a container |
| subtype `enum` | enum members are enum members; no emitter puts a `field` under one |
| subtype `const` | a const is a single value |
| subtype `file` | the per-file carrier is a container but not a class |
| subtype `import` | a reference to a module, never a declaration |
| subtype `delegate` | a signature; VB.NET/C# give it no members (zero CONTAINS) |
| subtype `interface`, language ≠ kotlin | C#/VB.NET/Java/Go/TS interfaces emit no field children |
| language ∈ {haskell, elm, ocaml, fsharp, idris, reasonml, rescript, crystal, erlang} | the package emits no `Subtype "field"` anywhere, so every entity is guaranteed zero-field |

### RED, combined (all round-2 exemptions removed, predicate intact)

```
--- FAIL: TestInterfaceAndDelegateExemptModuleCounted_6536
    ClassTotal = 3, want 1 … ZeroFieldsPct = 66.66666666666667, want 0
--- FAIL: TestKotlinInterfaceStaysInDenominator_6536
    a VB.NET interface is still in the denominator; it cannot own fields
--- FAIL: TestImportPlaceholdersExemptFromFieldDenominator_6536
    ClassTotal = 2 on a graph of 2 import placeholders and nothing else, want 0;
    zero-field rate 100 is guaranteed, not measured
--- FAIL: TestNonFieldBearingLanguageCarriersExempt_6536
    ClassTotal = 3 on a Haskell graph, want 0
```

GREEN: `ok github.com/cajasmota/grafel/internal/feedback 0.647s` — 73 tests, 0 FAIL, whole package, no `-run` filter, exit code captured separately.

## Mutants — 7 scored, 7 killed

**Correction to the earlier table in this PR.** It claimed *"5 scored, 5 killed, no survivors, so nothing left to pin"*. That was false: `"external": true` in `classLikeKindTails` **SURVIVED** the whole package. The re-scored table:

`go vet` first on each (a mutant that does not compile proves nothing), `-count=1` (a cached green is not evidence), restored by `cp` with the shasum asserted after every one — never `git checkout -- <file>`.

| mutant | direction | round-1 verdict | verdict now |
|---|---|---|---|
| admit `external` tail | permissive | **SURVIVED** (unscored) | **KILLED** — `TestClassLikeKindTailsConstrainedFromAbove_6536` |
| drop the `import` exemption | permissive | n/a | **KILLED** |
| drop the `delegate` exemption | permissive | n/a | **KILLED** |
| drop the `interface` exemption entirely | permissive | n/a | **KILLED** — 2 tests |
| drop the non-field-bearing-language exemption | permissive | n/a | **KILLED** |
| exempt `interface` in every language incl. Kotlin | restrictive | n/a | **KILLED** — drops a passer |
| exempt `module` globally | restrictive | n/a | **KILLED** — drops VB.NET Modules |

The five round-1 mutants (remove `component`; admit `operation`; drop the subtype filter; re-admit `enum`/`const`; re-admit `file`) remain killed by the tests already in the package.

## The golden D2 cross-check is tautological — recorded, not fixed

`report_golden_test.go`'s D2 recomputes `isFieldExtractionCandidate(...)` from the production symbols, so **no wrong tail and no wrong exemption can make it fail** — all it asserts is that `Generate` calls the predicate and counts the result. It was already half-tautological before this PR and lost its last literal here. A `TAUTOLOGY WARNING` comment now says so at the check itself, and it is **not** counted as coverage. Rebuilding it into an independent expectation is out of scope and tracked separately.

SQL / `SCOPE.Datastore` (needs the numerator widened for subtype `column` too) is out of scope and filed separately.

## Two existing expectations moved — flagged for a reviewer to check, not to wave through

Both moved because the measurement genuinely changed, not because they were wrong:

- `report_golden_test.go` — its D2 cross-check recomputes the predicate inline; it now uses the same exclusion set as production.
- `TestGenerate_FieldChildrenAndContainerTerminalsClassifiedCorrectly` — 11 candidates / 0.0% → 13 / ~15.4%. Its two `SCOPE.Component` subtype-`class` fixtures own no field children and are now counted. **That is real signal where there was previously silence.**

## Out of scope

#6536's item 3 — auditing the other kind-tail lists in `internal/feedback` for the same asymmetry — is deliberately untouched and dispatched separately. The population of similar lists is unmeasured, and each one is a metric that can be silently vacuous the same way this one was.

## Verification

`gofmt -l internal/feedback/` clean · `go vet` 0 · `go test -count=1 ./internal/feedback/` 0, whole package, no `-run` filter, exit code captured separately.

### DIFF
--- a/internal/feedback/report.go
+++ b/internal/feedback/report.go
@@ -269,12 +269,10 @@ func Generate(_ context.Context, docs []*graph.Document, opts Opts) (*Report, er
 			// the namespace-stripped tail (see kindTail, mirroring
 			// internal/graph/coverage.go) and include schema.
 			//
-			// A Subtype == "field" entity is itself a field LEAF, not a
-			// class/model container — classLikeKindTails includes "schema"
-			// which also matches these leaves (SCOPE.Schema/field), so they
-			// must be excluded here or every field would double as a "class"
-			// and guarantee a 100% zero-fields rate.
-			if isClassLikeKind(kind) && e.Subtype != "field" {
+			// Entities that are not field-bearing class containers are
+			// excluded from the denominator — see nonClassSubtypes for the
+			// list and the reasoning (#6536).
+			if isClassLikeKind(kind) && !nonClassSubtypes[e.Subtype] {
 				classCandidates = append(classCandidates, classCandidate{
 					id:            e.ID,
 					fieldCountRaw: e.PropGet("field_count"),
@@ -589,11 +587,50 @@ func kindTail(kind string) string {
 // class/model/field-bearing semantics for the field-extraction metric. Real
 // FB-loaded graphs use canonical kinds (SCOPE.Class, SCOPE.Schema, SCOPE.Model,
 // bare Model) — never the lowercase literals the in-memory unit fixtures used.
+// "component" is here because the entire C# family emits types under it: every
+// C#/VB.NET class, structure, module, interface and delegate is SCOPE.Component
+// (internal/extractors/vbnet/extractor.go entityKind, which follows
+// internal/extractors/csharp/csharp.go). Omitting it meant the metric had never
+// sampled a real class in any C#-family codebase — the only survivors of the
+// candidate filter were the SCOPE.Schema enums and consts, which cannot own
+// fields, so the reported rate was a guaranteed 100% rather than a measurement
+// (#6536, surfaced by #6535). Any change to this set must be checked against
+// the kinds the extractors actually emit, not against these literals.
 var classLikeKindTails = map[string]bool{
-	"class":  true,
-	"struct": true,
-	"model":  true,
-	"schema": true,
+	"class":     true,
+	"struct":    true,
+	"model":     true,
+	"schema":    true,
+	"component": true,
+}
+
+// nonClassSubtypes are the subtypes that carry a class-like KIND but are not
+// field-bearing class/model CONTAINERS, and so are excluded from the
+// field-extraction denominator (#6536).
+//
+// The decision, stated rather than left incidental: they are exempt.
+//
+//   - "field" — a field LEAF is itself the child, not a container. The tail
+//     "schema" matches these leaves too, so without this exclusion every field
+//     would double as a "class" with zero fields.
+//   - "enum" / "const" — a const is a single value and an enum's members are
+//     enum members, never fields; no extractor that emits these subtypes
+//     (vbnet, csharp, cpp, php, proto, avro, solidity) emits a Subtype "field"
+//     child under one. Counting them means every such entity is permanently a
+//     zero-field failure, which puts a floor of false failures under the
+//     metric and is exactly what produced the misleading 100% in #6535.
+//   - "file" — every extractor's file carrier is a SCOPE.Component
+//     (internal/extractor.FileEntity). It is a container, but it is not a
+//     class, and admitting "component" above would otherwise enrol one
+//     guaranteed-zero-field entity per indexed file.
+//
+// A metric whose denominator contains populations that cannot pass reports
+// noise, not coverage.
+var nonClassSubtypes = map[string]bool{
+	"field": true,
+	"enum":  true,
+	"const": true,
+	"file":  true,
 }
 
 // isClassLikeKind reports whether kind is a class/model/schema-shaped entity
--- a/internal/feedback/report.go
+++ b/internal/feedback/report.go
@@ -270,9 +270,9 @@ func Generate(_ context.Context, docs []*graph.Document, opts Opts) (*Report, er
 			// internal/graph/coverage.go) and include schema.
 			//
 			// Entities that are not field-bearing class containers are
-			// excluded from the denominator — see nonClassSubtypes for the
-			// list and the reasoning (#6536).
-			if isClassLikeKind(kind) && !nonClassSubtypes[e.Subtype] {
+			// excluded from the denominator — see isFieldExtractionCandidate
+			// for the exemption set and the reasoning (#6536).
+			if isFieldExtractionCandidate(kind, e.Subtype, e.Language) {
 				classCandidates = append(classCandidates, classCandidate{
 					id:            e.ID,
 					fieldCountRaw: e.PropGet("field_count"),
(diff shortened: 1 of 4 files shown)


cajasmota:
## Coordinator mutant on round 2 — KILLED

I scored the mutant that attacks the new **language-based** exemption, which is the widest of the new surfaces and the one none of the seven listed mutants targets from this angle:

```go
var nonFieldBearingLanguages = map[string]bool{
	"vbnet":    true,   // <- added
	"haskell":  true,
	...
```

Admitting a language that **does** emit fields exempts the entire C#/VB.NET population — the exact codebase #6535 reported — re-creating the original defect at full scale rather than at the margins.

`go vet ./internal/feedback/` → **0**. `go test -count=1 ./internal/feedback/` → **EXIT=1**, three tests down:

```
--- FAIL: TestFieldExtractionSamplesRealClasses_6536
--- FAIL: TestInterfaceAndDelegateExemptModuleCounted_6536
--- FAIL: TestClassLikeKindTailsConstrainedFromAbove_6536
```

Restored by `cp`, shasum `136bca680bceabfc966ff5ea3f85b695c6947caf` verified, working tree clean. **DEAD.**

## The two narrowings are the strongest part of this round

Round 2 was told to exempt `module` globally and `interface` everywhere. It did **neither**, and refused on measured evidence rather than argument:

- **`module` stays counted.** Probed against the real VB.NET extractor: `Public Module Util` → `SCOPE.Component`/`module` with `CONTAINS → Util.Total`, `Util.cache` (subtype `field`). Exempting it would delete **real passers** from exactly the population #6535 is about.
- **`interface` is exempt everywhere except Kotlin.** Probed: `interface Shape { val sides: Int }` → `Shape` CONTAINS `Shape.sides` subtype `field`.

That distinction matters more than it looks. **Dropping a passing population biases the rate in the same direction as counting a non-passer** — both push the reported figure away from the truth, and a fix that over-exempts would have looked like an improvement while quietly re-introducing the defect. The review's asks were directionally right and wrong in the specifics; taking them literally would have shipped a subtler version of the same bug.

The `""`-subtype exemption the review asked for also turned out to be unnecessary — Haskell's empty-subtype import placeholders are already covered by the language exemption, so it was dropped rather than added.

## Recorded honestly, which is the point

The false claim is corrected: **"5 scored, 5 killed, no survivors" was wrong** — the `external` tail mutant survived the whole package. It is now killed by `TestClassLikeKindTailsConstrainedFromAbove_6536`, and the table says so.

`report_golden_test.go`'s D2 check carries a `TAUTOLOGY WARNING` stating it recomputes the production predicate, cannot fail from a wrong tail or exemption, and is **not coverage**. Rebuilding it is out of scope and flagged rather than quietly counted.

SQL / `SCOPE.Datastore` deliberately untouched — filed as #6543, because admitting that tail without widening the numerator would guarantee 100% for SQL, which is this same defect in a new place.

Issue #5822, opened 2026-07-17
grafel status residuals: relationship over-count, stale commit label, --ref on a transient branch, count swings between runs


pmedranocurantis:
# grafel v0.1.8.3 — `--full` rebuild of a large monorepo silently killed by the 30-min watchdog (+ 3 secondary issues)

**Version:** `v0.1.8.3` (commit `baca1ca`, built `2026-07-17T18:21:20Z`), arm64 macOS, background daemon mode
**Repo under test:** single-repo group `<monorepo>` — ~33,074 candidate files, ~448k entities, ~1.96M relationships, ~25k files extracted
**Context:** doing a from-scratch reindex right after installing v0.1.8.3.

---

## 1. (Blocker) A valid `--full` rebuild is SIGKILLed at the 30-min per-repo watchdog, and the failure is not surfaced to the CLI

`grafel rebuild <group> <slug> --full` on this monorepo runs the whole pipeline (extraction → resolvers → prune/fold → `data-lineage`) and then gets killed by the daemon's per-repo rebuild watchdog at exactly 30m, right as it reaches the final `data-lineage` pass. Nothing commits, so `grafel status` keeps showing the **previous** index with no indication anything went wrong.

**Evidence (`~/.grafel/logs/daemon.err`):**
```
2026/07/17 13:39:39  subprocess-indexer: foreground rebuild gomaxprocs=10 ...
2026/07/17 13:39:43  [child] discovered 33074 candidate files ...
2026/07/17 14:08:13  [child] data-lineage: data_access_seen=578 edges=49 ...   ← reached the LAST pass
grafel: rebuild <monorepo> STALLED — no result after 30m0s; surfacing as timeout and continuing with remaining repos
grafel: rebuild <monorepo> took 30m0.052s [FAILED: repo index timed out after 30m0s (still running in background)]
grafel: rebuild exit group=<monorepo> took=30m0.053s
2026/07/17 14:09:36 ERROR subprocess-indexer: exited with error pid=79553 err="signal: killed"
```

**Two distinct problems here:**

**1a. The watchdog default (30m) is too tight for a large monorepo, and killing at the last pass wastes the entire run.** The knob exists — env var `GRAFEL_REBUILD_REPO_TIMEOUT` (Go duration string, e.g. `90m`) — but:
  - It's not documented in `grafel help` / `grafel help advanced`.
[...]

PR #6542, opened 2026-08-21
fix(#5822): status TOTAL accumulated per registry entry while rows were keyed by slug, so a double-registered repo doubled every counter

cajasmota:
Refs #5822

The last open residual of a contributor's report: **`grafel status` over-reported the relationship count by roughly 2x.** They measured `status` showing **3,487,888** against the indexer's own committed **1,963,736** on the same graph, and said plainly that the number was alarming because it is exactly what graph corruption would look like. The on-disk graph was fine. The line describing it was not.

## Mechanism, verified against the code rather than taken on trust

`ComputeStatusSummaryForRef` (`internal/cli/status_stats.go`) loops over **registry entries** and accumulates every group counter once per iteration, while rows are stored as `s.RepoStats[r.Slug]` and printed **by slug**.

So a second entry with an already-seen slug **overwrites the row** and **adds again to TOTAL**. One repo, registered twice, is summed twice and printed once — which is precisely the "TOTAL exceeds the sum of the visible rows" shape reported, with a 2x factor when it is double-registered. There was no dedupe guard on either side.

## Three counters the issue did not name

Audited rather than assumed. Affected: `TotalEntities` (`:296`, `:319`), `TotalRelationships` (`:297`, `:320`), `UnsupportedExt[ext]` (`:307`), `HTTPEndpoints` (`:467`), `ProcessFlows` (`:468`), `EnrichmentCandidates` (`:475`), `RepairCandidates` (`:476`) — **plus `ReindexRequired`, `MigrationFailed` and `MigrationNotAccepted` (`:245-252`)**, which the issue's analysis missed.

`MigrationLive` is a bool set to `true` and was already idempotent.

That matters beyond arithmetic: the three unnamed ones are **health signals**. A double-registered repo inflated "how much of your graph needs attention", which is the kind of number an operator acts on.

## RED

```
--- FAIL: TestStatusTotalsCountADoubleRegisteredRepoOnce
    TOTAL relationships = 3927472, but the printed rows sum to 1963736
    TOTAL entities      = 896000,  but the printed rows sum to 448000
    HTTPEndpoints = 4, want 2 · ProcessFlows = 2, want 1
    EnrichmentCandidates = 6, want 3 · RepairCandidates = 4, want 2
    UnsupportedExt[.bin] = 8, want 4
--- FAIL: TestStatusTotalsDedupeSameSlugAcrossDifferentPaths   (same shape)
```

**3,927,472 against 1,963,736 — the contributor's own committed figure, doubled exactly.** The synthetic registry reproduces their measurement without needing their monorepo, which is why they were told not to re-collect anything.

GREEN: `ok internal/cli 49.729s`, whole package, no `-run` filter, `-count=1`.

## The observation had to be built first

Nothing in the tree could observe this: **no test anywhere built two registry entries sharing a slug.** That is why a 2x over-count survived to a user. The test is the deliverable as much as the guard is.

## Fix, and why the key is the slug specifically

One dedupe on `r.Slug` at the top of the loop — keyed on the slug **because that is what the row map is keyed by**. The identity being restored is "one printed row, one contribution to TOTAL", so the dedupe key must be the row key or the two halves disagree again.

Path-keying would miss same-slug/different-path. Anything coarser converts an over-count into a **silent under-report**, which is strictly worse — the three permissive mutants below are what establish that, rather than an argument.

Skipped entries also no longer perform a redundant graph open.

## Mutants — 4 scored, 4 killed

All `go vet`-clean before scoring (a mutant that does not compile proves nothing), `-count=1` (a cached green is not evidence), restored by `cp` with shasum `e6283104…` re-verified — never `git checkout -- <file>`.

|---|---|---|
| revert the dedupe guard | — | **KILLED** |
| key on `strings.ToLower(TrimSpace(slug))` | **permissive** | **KILLED** — `api` vs `API` |
| collapse slugs sharing a prefix | **permissive** | **KILLED** — `api` vs `api-gateway` |
| key on `filepath.Dir(r.Path)` | **permissive** | **KILLED** — two distinct repos collapsed into one row |

No survivors, so nothing left to pin.

## One fixture note worth keeping

macOS's case-insensitive filesystem made directories `api` and `API` **the same directory**, which would have made the case-variant mutant unfalsifiable on this platform. The fixture therefore uses a distinct directory name and varies the slug independently — documented in the test, because the next person on a case-sensitive filesystem will not hit it and could undo it by accident.

## Verification

`gofmt -l` clean · `go vet` 0 · `go test -count=1 ./internal/cli/` 0, whole package, no `-run` filter, exit code captured separately.

### DIFF
--- a/internal/cli/status_dupslug_5822_test.go
+++ b/internal/cli/status_dupslug_5822_test.go
@@ -1,0 +1,271 @@
+package cli
+
+// status_dupslug_5822_test.go — #5822, the relationship over-count.
+//
+// The contributor measured `grafel status` reporting 3,487,888 relationships
+// against the indexer's own committed 1,963,736 on the same graph — very close
+// to exactly 2x. It is NOT a write-path defect: the graph stores no duplicates.
+//
+// ComputeStatusSummaryForRef walks the group's REGISTRY ENTRIES and accumulates
+// every group-level counter once per entry, while the per-repo rows it renders
+// are keyed by SLUG (s.RepoStats[r.Slug] = rs) and printed by slug. Two
+// registry entries that resolve to the same slug are therefore SUMMED TWICE and
+// PRINTED ONCE — which is exactly the reported shape: "TOTAL exceeds the sum of
+// the visible rows", with a ~2x factor when one repo is double-registered.
+//
+// Nothing in the tree could observe this before these tests: no test anywhere
+// built two registry entries sharing a slug.
+//
+// The invariant pinned here is deliberately expressed as an IDENTITY rather
+// than as a magic number — every group-level counter must equal the sum over
+// the rows that are actually rendered. That is the property the user can check
+// on screen, and it is the property the defect broke.
+
+import (
+	"encoding/json"
+	"os"
+	"path/filepath"
+	"testing"
+	"time"
+
+	"github.com/cajasmota/grafel/internal/daemon"
+	"github.com/cajasmota/grafel/internal/graph"
+	"github.com/cajasmota/grafel/internal/graph/fbwriter"
+	"github.com/cajasmota/grafel/internal/registry"
+)
+
+// dup5822Repo materialises one repo with a fully-populated state directory:
+// a graph-stats.json sidecar (entity/relationship/unsupported-extension
+// counts), a graph.fb carrying http-endpoint and process entities, and an
+// enrichment-candidates.json carrying both enrichment and repair candidates.
+//
+// Every group-level accumulator in ComputeStatusSummaryForRef is fed by exactly
+// one of these files, so a single fixture exercises all of them at once.
+func dup5822Repo(t *testing.T, root, dirName string, ents, rels int) string {
+	t.Helper()
+	repoPath := filepath.Join(root, dirName)
+	if err := os.MkdirAll(repoPath, 0o755); err != nil {
+		t.Fatalf("mkdir repo: %v", err)
+	}
+	stateDir := daemon.StateDirForRepo(repoPath)
+	if err := os.MkdirAll(stateDir, 0o755); err != nil {
+		t.Fatalf("mkdir state: %v", err)
+	}
+
+	side := graph.GraphStatsSidecar{
+		Version:               1,
+		ComputedAt:            time.Now().Add(-10 * time.Minute),
+		TotalFiles:            5,
+		TotalEntities:         ents,
+		TotalRelationships:    rels,
+		UnsupportedExtensions: map[string]int{".bin": 4},
+	}
+	data, err := json.Marshal(side)
+	if err != nil {
+		t.Fatalf("marshal sidecar: %v", err)
+	}
+	if err := os.WriteFile(filepath.Join(stateDir, "graph-stats.json"), data, 0o644); err != nil {
+		t.Fatalf("write sidecar: %v", err)
+	}
+
+	doc := &graph.Document{
+		Repo:        dirName,
+		GeneratedAt: time.Now().Add(-10 * time.Minute),
+		IndexedRef:  "main",
+		IndexedSHA:  "abcdef012345",
+		Entities: []graph.Entity{
+			{ID: "h1", Name: "GET /a", Kind: "http_endpoint_definition", SourceFile: "a.go", Language: "go"},
+			{ID: "h2", Name: "GET /b", Kind: "http_endpoint_definition", SourceFile: "b.go", Language: "go"},
+			{ID: "p1", Name: "flow", Kind: "process", SourceFile: "c.go", Language: "go"},
+		},
+		Relationships: []graph.Relationship{},
+	}
+	if err := fbwriter.WriteAtomic(filepath.Join(stateDir, "graph.fb"), doc); err != nil {
+		t.Fatalf("write graph.fb: %v", err)
+	}
+
+	cands := []map[string]string{
+		{"kind": "repair_edge", "subject_id": "r1"},
+		{"kind": "repair_edge", "subject_id": "r2"},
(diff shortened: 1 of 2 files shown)


cajasmota:
## Coordinator mutant — one SURVIVOR, and it is the fourth instance of the same class this milestone

I scored a mutant nobody else ran, on an axis the four existing mutants do not touch. All of those vary the **dedupe key**; this one keeps the key exactly as written and varies **which duplicate survives**.

```go
// real code: first entry wins
seenSlugs := make(map[string]struct{}, len(repos))
for _, r := range repos {
	if _, dup := seenSlugs[r.Slug]; dup { continue }
	seenSlugs[r.Slug] = struct{}{}

// mutant: last entry wins
lastIdx := make(map[string]int, len(repos))
for i, r := range repos { lastIdx[r.Slug] = i }
for i, r := range repos {
	if lastIdx[r.Slug] != i { continue }
```

`go vet ./internal/cli/` → **0**. `go test -count=1 ./internal/cli/` → **EXIT=0, `ok … 51.769s`**. Restored by `cp`, shasum `e6283104ea4861ab2fe922d777dfbce80adb3280` verified. **SURVIVES.**

## Why it survives, and why that is a gap rather than a curiosity

Both variants contribute **exactly once per slug**, so every count assertion in the new tests is satisfied by construction. What differs is **which duplicate's row is rendered** — and the row carries `Path`.

The code states the choice and gives it a justification:

> *First entry wins, which matches the pre-existing precedence for the group's OTHER per-slug state and is deterministic in registry order.*

Nothing observes it. `TestStatusTotalsCountADoubleRegisteredRepoOnce` registers **both entries at the same `repoPath`**, so first and last are indistinguishable there. `TestStatusTotalsDedupeSameSlugAcrossDifferentPaths` is the one test where they differ — and it asserts counts, not which path survived.

So a claim about **matching pre-existing precedence** — the sort of thing a future refactor would silently invert — rests on nothing.

This is the same shape as the three findings already recorded on #6528 this milestone: prose asserting a property, no fixture exercising the axis, mutant survives. The consequence here is smaller than a pruned endpoint — a user sees one registration's path instead of another's — but it is user-visible, and the comment is the only thing currently defending it.

## What round 2 needs

One assertion in `TestStatusTotalsDedupeSameSlugAcrossDifferentPaths` that the surviving row carries the **first** entry's `Path`. Then re-score the last-wins mutant; it must die.

Worth doing rather than deleting the comment: if first-wins really does match the group's other per-slug precedence, that is a genuine invariant and should be pinned. If it turns out **not** to match, say so and fix the comment instead — but establish which, rather than leaving the claim standing unobserved.

Everything else on this PR holds. The four key-varying mutants are all genuinely dead, the RED reproduces the contributor's exact 3,927,472 / 1,963,736, and the three counters the issue never named (`ReindexRequired`, `MigrationFailed`, `MigrationNotAccepted`) are a real find — those are health signals, so a double-registered repo was inflating "how much of your graph needs attention".

cajasmota:
## The survivor turned out to be worse than unpinned — the guard silently inverted a user-visible precedence

Round 2 was asked to pin first-wins and instead **stopped and refuted the claim**, which was the right call. The comment says first-entry-wins *"matches the pre-existing precedence for the group's OTHER per-slug state."* It does not.

Every group-scoped per-slug map in the tree is **last-wins**:

| site | precedence |
|---|---|
| `status_stats.go:312`, `:541` — `s.RepoStats[r.Slug] = rs` (the very row this dedupe governs) | last wins |
| `links.go:105` — `srcPaths[r.Slug] = r.Path` | last wins |
| `links.go:208` — mirror pass | last wins |
| `links.go:475-476` — `docs[r.Slug]`, `graphPaths[r.Slug]` | last wins |

Nothing in `internal/registry/` keys anything by slug at all, so there is no registry-side precedence to inherit. The only first-wins analogue, `quarantine.go:248-266`, is explicitly a **cross-group lookup** — documented as such — not group-scoped aggregation.

**And the decisive evidence is this function's own history.** `git diff main...HEAD` on `status_stats.go` is `+34 / -0` — pure addition, no deletion. Before the guard, both entries ran and the second `s.RepoStats[r.Slug] = rs` overwrote the first, so the `Path` a user saw for a double-registered slug came from the **last** entry. The comment at `:225` says so itself.

So the guard does not preserve the precedence — it **inverts** it, user-visibly, while claiming the opposite. A counting fix was carrying an undeclared rendering change.

## Decision: switch to last-wins

Three reasons, in order of weight:

1. **A counting fix should not silently change what is rendered.** Preserving the pre-fix `Path` keeps this PR to the defect it is about.
2. **It matches every group-scoped per-slug map in the tree**, rather than introducing a second convention.
3. **It costs nothing.** The one merit first-wins had — skipping the duplicate before any graph open — is fully retained by pre-computing the last index per slug:

```go
lastIdx := make(map[string]int, len(repos))
for i, r := range repos { lastIdx[r.Slug] = i }
for i, r := range repos {
	if lastIdx[r.Slug] != i { continue }
```

That is the exact mutant I scored earlier, so it is already known to vet clean and keep the package green.

Round 3 is switching the guard, rewriting the comment to state what is actually true, and pinning the direction so the first-wins mutant dies.

## Worth naming

This started as *"a documented property nothing observes"* — the fourth such finding this milestone. It ended as a behaviour change nobody had noticed, hiding inside prose that asserted the opposite. **The unobserved claim was not merely unproven; it was false**, and only mutating that axis surfaced it.

Everything else on this PR stands: the four key-varying mutants are genuinely dead, the RED reproduces the reporter's exact 3,927,472 against 1,963,736, and the three counters the issue never named — `ReindexRequired`, `MigrationFailed`, `MigrationNotAccepted` — remain a real find.

Issue #6366, opened 2026-08-19
svelte dangles RENDERS/USES/NAVIGATES_TO/CONTAINS the same way astro did — no FileEntity, so the raw path reaches the graph


cajasmota:
Found by the general scan added in #6365 (for #6298), not by inspection. Svelte has **astro's exact defect** — not the misanchoring shape @arthurgeron fixed for Solidity in #6297, but the worse one.

## The shape

`internal/extractors/svelte/` emits no `extractor.FileEntity` anywhere, and `entities[0]` is the component named from the basename. `RENDERS`, `USES`, `NAVIGATES_TO` and `CONTAINS` all pass `file.Path` / `filePath` as `FromID`.

Because assembly substitutes the owning record's id **only** when `FromID == ""` (`cmd/grafel/index.go:6063-6069`, `internal/extractors/incremental.go:2215-2219`), and because there is no file entity for the path to resolve onto, the raw path reaches the graph verbatim.

Measured: `Card.svelte` rendering `<Button />` yields

```
FROM=<UNRESOLVED:src/lib/Card.svelte>   TO=Button@src/lib/Button.svelte
```

`UNRESOLVED` is literal. The ToID binds; the FromID does not.

**Only `RENDERS` was measured.** The other three are the same construction on the same record and are recorded as *inference, not finding* — they are allow-listed in #6365 as `KNOWN OFFENDER (#6298)` so the scan stays green while this issue is open, and each carries that qualifier.

## Why this is worse than the Solidity/verilog case

Solidity and verilog **misanchor**: the edge lands on the file component, so several types in one file merge their edges onto one node. Wrong, but present and traceable.

Svelte and astro **dangle**: the edge points at an id no entity carries. A misanchored edge can be found and corrected; a dangling one may vanish or point at nothing, and nothing in the graph records that it was ever meant to exist.

## Scope

Eleven sites across four relationship kinds. The fix is the same one-line shape each time — drop `FromID` and let assembly stamp the owner — but it deserves its own change and its own measurement rather than riding along on an issue that reviewed verilog and astro:

- `RENDERS` — measured above.
- `USES`, `NAVIGATES_TO`, `CONTAINS` — same construction, unmeasured. **Measure each before changing it**; #6298 is the standing example of why (astro was assumed to match verilog's pattern and turned out to be a different, worse failure).

[...]

PR #6546, opened 2026-08-22
fix(#6413): three dead filePath params in the svelte extractor, not the one the issue named

### LINKED ISSUE #6413: svelte: #6366 left dead path parameters behind, and a dead filePath next to a RelationshipRecord literal invites the exact bug it removed
Follow-up to #6366 (merged in 7cb724922).

That fix dropped `FromID` at all eleven svelte emission sites so assembly stamps `entities[0]`, the component. The path values those sites used to consume came in as function parameters, and **at least some of those parameters are now dead** — carried through the signature and never read.

Confirmed instance:

```go
// internal/extractors/svelte/extractor.go:1026
func extractChildComponents(template string, templateStartLine int, filePath, componentName string) []types.RelationshipRecord {
```

`filePath` appears nowhere in that function's body any more.

## Why this is worth a change rather than a shrug

A dead `filePath` parameter sitting next to a `types.RelationshipRecord` literal is an *invitation*. The next person adding a relationship kind to that function sees a path in scope, and `FromID: filePath` is the obvious-looking thing to write. That is precisely the dangling-FROM defect #6366 just spent a measurement removing. The allowlist scan in `internal/extractors/file_anchored_rels_guard_test.go` would catch it — but the cheaper fix is to not leave the loaded gun on the table.

Go does not flag unused function parameters, and `go vet` does not either, so nothing in CI notices.

## Scope

**Derive the set — do not trust this issue's list.** The one site above is the only one confirmed by reading; the other ten sites have not been checked. Whoever picks this up should enumerate every function in `internal/extractors/svelte/` that takes a path-ish parameter and determine, per function, whether the body still reads it. Some of those parameters are probably still live (span/line computation, or genuinely used for a `ToID`), so this is not a blanket removal.

Removing a parameter changes call sites, so the fix is mechanical but not zero-risk. `TestSvelte_ComponentRelsAnchoredOnComponent` and `TestNoNewFileAnchoredTypeRelationships` both stay green throughout — they are the regression net, and neither one is what proves this change correct. This is a readability/trap-removal change; the honest test story is "no behaviour change", so the verification is that both existing guards still pass and the package still compiles, not a new assertion.

Small, self-contained, good first issue.

cajasmota:
Closes #6413

Dead `filePath` parameters left behind by #6366, removed from three Svelte extractor functions and their call sites.

## Three, not one

The issue's body names one. Grounding against `main` found **three**, each taking `filePath, componentName` and reading only `componentName` (as `caller` / `from_component`):

- `extractScriptNavigation` — `internal/extractors/svelte/extractor.go:698`
- `extractRouteDirectives` — `extractor.go:733`
- `extractChildComponents` — `extractor.go:1026`

All three were re-read in full before removal; none had to be dropped from the change.

## Why this is not a blanket sweep

The issue warns that it is not, and the warning is load-bearing. **Seven** other path-taking functions in the same file genuinely read `filePath` for `SourceFile:` — `:362`, `:511`, `:572`, `:646`, `:772`, `:850`, `:944` — and are untouched. The live/dead split is **7 live, 3 dead**.

Diff is 6 insertions / 6 deletions in one file: three signatures plus their call sites at `:264`, `:287`, `:299`.

## The mutant had to be aimed at the guards, not at the change

There is no new assertion to write here — removing an unused parameter cannot fail a test that did not already exist. So the question worth answering is whether the guards this change relies on actually observe the defect it makes unrepresentable.

Mutant: re-anchor `RENDERS` onto `filePath` in `extractChildComponents` (`FromID: filePath`, parameter reintroduced) — i.e. re-create the #6366 defect. `go vet` clean first, so the mutant is real rather than a compile error.

**Both guards fired.**

```
--- TestSvelte_ComponentRelsAnchoredOnComponent
    RENDERS -> Button@src/lib/Button.svelte
    FROM = <UNRESOLVED:src/lib/Card.svelte>, want Card@src/lib/Card.svelte
    RENDERS -> <UNRESOLVED:Link>
    FROM = <UNRESOLVED:src/lib/Card.svelte>, want Card@src/lib/Card.svelte

--- TestNoNewFileAnchoredTypeRelationships
    key "svelte:extractChildComponents:RENDERS" (1 site(s))
      svelte/extractor.go:1041 FromID: filePath
```

So the safety net is real: `issue6366_anchoring_test.go` catches the semantic regression, and the allow-list scan in `file_anchored_rels_guard_test.go` catches the *new site* independently. **No gap to report and no new test needed** — which is the outcome the issue predicted, now verified rather than assumed.

Restored by `cp`; shasum back to `32a0cb15d186eebd2ac8d0ce31d3ad1da548266d`, identical to pre-mutant. Never `git checkout -- <file>`.

## Verification

`gofmt -l` clean · `go vet` clean · `go test -count=1` on whole packages, no `-run` filter:

```
ok  internal/extractors/svelte  0.905s
ok  internal/extractors        29.262s
```

### DIFF
--- a/internal/extractors/svelte/extractor.go
+++ b/internal/extractors/svelte/extractor.go
@@ -261,7 +261,7 @@ func (e *Extractor) Extract(ctx context.Context, file extractor.FileInput) ([]ty
 		entities = append(entities, cfEnts...)
 
 		// Navigation (#2856): imperative svelte-routing navigate()/push().
-		navRels := extractScriptNavigation(scriptContent, scriptStartLine, file.Path, componentName)
+		navRels := extractScriptNavigation(scriptContent, scriptStartLine, componentName)
 		entities[0].Relationships = append(entities[0].Relationships, navRels...)
 
 		// Lifecycle (#2856): state_setter_emission — store .set/.update and
@@ -284,7 +284,7 @@ func (e *Extractor) Extract(ctx context.Context, file extractor.FileInput) ([]ty
 	// ── 3. Extract RENDERS edges + branch conditions from the template ───────
 	templateContent, templateStartLine := extractTemplateSection(src)
 	if templateContent != "" {
-		renderRels := extractChildComponents(templateContent, templateStartLine, file.Path, componentName)
+		renderRels := extractChildComponents(templateContent, templateStartLine, componentName)
 		if len(renderRels) > 0 {
 			// Attach RENDERS relationships to the component entity.
 			entities[0].Relationships = append(entities[0].Relationships, renderRels...)
@@ -296,7 +296,7 @@ func (e *Extractor) Extract(ctx context.Context, file extractor.FileInput) ([]ty
 
 		// Navigation (#2856): <Route path="…"> / <Link to="…"> svelte-routing
 		// directives emit NAVIGATES_TO edges.
-		linkRels := extractRouteDirectives(templateContent, templateStartLine, file.Path, componentName)
+		linkRels := extractRouteDirectives(templateContent, templateStartLine, componentName)
 		entities[0].Relationships = append(entities[0].Relationships, linkRels...)
 
 		// Svelte Internals (#2877): `use:` action directives.
@@ -695,7 +695,7 @@ func extractBranchConditions(template string, templateStartLine int, filePath, c
 // has no built-in router, so this targets the ecosystem libraries:
 // svelte-routing's navigate('/x') and svelte-spa-router's push('/x')/
 // replace('/x'). Returns NAVIGATES_TO edges from the component file.
-func extractScriptNavigation(script string, scriptStartLine int, filePath, componentName string) []types.RelationshipRecord {
+func extractScriptNavigation(script string, scriptStartLine int, componentName string) []types.RelationshipRecord {
 	var rels []types.RelationshipRecord
 	seen := map[string]bool{}
 	emit := func(route, via string, off int) {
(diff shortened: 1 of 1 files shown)


cajasmota:
## Coordinator mutant — KILLED, and deliberately aimed at a different function than the PR's own

The PR's mutant re-anchored `RENDERS` in `extractChildComponents`. That proves the guards observe **that** function. It does not prove they observe the other two this PR touches — and a guard that happens to cover one of three call sites is exactly the shape that reads as coverage and is not.

So I re-anchored `NAVIGATES_TO` in **`extractRouteDirectives`** instead (`internal/extractors/svelte/extractor.go:733`), reintroducing the `filePath` parameter and setting `FromID: filePath`.

`go vet` clean first — my first attempt did **not** compile (`undefined: filePath` at the call site, since the enclosing scope uses `file.Path`), and a non-compiling mutant proves nothing, so I fixed it and re-scored.

**Both guards fired, on the other function too:**

```
./internal/extractors/svelte/  EXIT=1
  --- FAIL: TestSvelte_ComponentRelsAnchoredOnComponent

./internal/extractors/         EXIT=1
  --- FAIL: TestNoNewFileAnchoredTypeRelationships
```

Restored by `cp`, shasum `32a0cb15d186eebd2ac8d0ce31d3ad1da548266d` verified, working tree clean.

So the safety net this PR relies on is **general rather than incidental**: `issue6366_anchoring_test.go` catches the semantic regression and the allow-list scan catches the new site independently, in whichever of the three functions the anchor comes back. That is the claim the PR makes, now tested at a second site rather than extrapolated from one.

Merging.

Issue #6474, opened 2026-08-21
diff.SaveManifest stamps live HEAD at save time, so a commit moving mid-pass labels the graph with a commit it was not built from


cajasmota:
Split out of #5822 (reported by @auxmedrano) after grounding it against `main` @ `640ea7784`. The load-bearing half of that residual was fixed by `1277c728f` (#6226); this is what it left behind.

## The defect

`diff.SaveManifest` stamps `git_commit` from **live HEAD at save time** (`internal/indexer/diff/diff.go:157-158`), not from the commit whose bytes the pass actually read.

Three callers still use it:

- `cmd/grafel/index.go:1286` — full index
- `internal/extractors/incremental.go:678` — zero-change reconcile
- `internal/extractors/incremental.go:1560` — successful incremental

All three read their file bytes at pass start and stamp HEAD at save. If HEAD moves **during** the pass, the manifest labels a graph built at commit A as being at commit B.

That label is not cosmetic: `daemon.indexedCommitShortNoGit` and `daemon.IndexedCommitForRepo` both **prefer `git_commit` over the `graph.fb` header**, so the wrong value is what users and the daemon see. It also disarms the #5710 head-advance detector, which is what let the original instance of this bug conceal itself.

## Why it is worth fixing rather than accepting

The window is small in an idle tree and **not small in an active one**. The reporter's environment flips HEAD every minute or two, which is exactly the condition that produced #5822's stale-label observation in the first place.

## The fix already exists in the codebase

`1277c728f` introduced `diff.SaveManifestAtCommit` (`internal/indexer/diff/diff.go:179`) precisely to thread a captured commit through, and wired it at `incremental.go:612` and `:758`. This issue is extending that treatment to the three remaining callers: capture the commit at pass start and pass it down, rather than re-asking git at save time.

## Acceptance

A test that moves HEAD **between** the byte-read and the save, and asserts the manifest records the commit the graph was built from, not the one HEAD points at afterwards. Without that, the fix is unfalsifiable — the window is timing-dependent and will not reproduce incidentally.

Worth checking while in here whether `SaveManifest` should keep its live-HEAD behaviour at all once the three callers are converted, or whether it should be removed in favour of the explicit-commit form so no future caller can reintroduce this.

cajasmota:
Grounded at `de1d96481`, read-only. **Premise confirmed on all three sites — but they are not symmetric, and one belongs in its own issue.**

## The three residual callers

[...]

Issue #5710, opened 2026-07-10
Incremental indexer no-ops on HEAD advance (changed_files=0), serves stale graph silently


pmedranocurantis:
## Summary

The background daemon's **incremental indexer silently no-ops** when the repo's checked-out `HEAD` advances: it emits `incremental_ok changed_files=0` and leaves the served graph pinned to the previously-indexed commit. In my case the graph stayed pinned at a commit that was **~22 hours / 56 commits / 207 changed files** behind the actual working-tree `HEAD`, and kept answering MCP queries against stale source until I ran a manual `grafel rebuild`.

This is a correctness problem for anyone relying on the daemon to auto-track a moving branch: queries return confidently-wrong results (old source, missing/renamed symbols) with **no visible staleness signal** to the agent, because `grafel status` reports the pin as the current `@ main`.

## Environment

| | |
|---|---|
| grafel version | `v0.1.7.4` (commit `fa521e5`, built `2026-07-09T18:30:29Z`) |
| Platform | macOS (darwin 25.5.0), Apple Silicon |
| Daemon mode | `background` |
| Indexing | `incremental=on subprocess=on` |
| Worktree indexing | disabled — `GRAFEL_MAX_WORKTREES_PER_REPO=0` (in-place: log line `worktree: indexing disabled`) |
| Repo scale | 1 repo, ~19,576 indexed files, ~291k entities, ~1.38M relationships, `predicted=22359MB` |
| mem_limit | `2560MB (fraction-of-RAM (capped))` |
| Watcher at time of bug | `repos=1 dirs=2816 events=8 dropped=0 slots_active=1 slots_paused=0` |

## Timeline / evidence

- Graph was last built **`2026-07-09T19:36:53Z`**, pinned `@ main (2b824d777179)` (a 2026-07-08 commit).
- Working-tree `HEAD` at time of investigation = **`cf020bff34db`** (current `origin/main` tip). `git rev-parse HEAD` and `git symbolic-ref HEAD` → `refs/heads/main` both confirmed the checkout was on `main` at the newer commit.
- Range `2b824d777179..cf020bff34db` = **56 commits, 207 files**, including **24 `.go`, 63 `.ts`, 11 `.java`, 19 `.sql`** — all indexed languages. So there were substantial real source changes on disk.
- The daemon nonetheless logged:

```
2026-07-10T17:37:57Z  index_start     …  predicted=22359MB ref=main
[...]

PR #6547, opened 2026-08-22
fix(#6502): a full index stamped the manifest and fb header from a post-extraction HEAD read

### LINKED ISSUE #6502: full index stamps both the manifest AND doc.IndexedSHA from a post-extraction HEAD read, so a mid-pass commit mislabels the graph — and fixing only one makes them disagree
Split out of #6474, which covers the two incremental `SaveManifest` call sites. This is the full-index site, and it is **not symmetric with those** — it carries a second defect on the same path.

## Site A

`Indexer.commitManifest` (`cmd/grafel/index.go:1251`, called at `:948`) reaches `diff.SaveManifest(...)` at `:1286`, which stamps live HEAD at save time. Unlike the incremental path, **there is no pass-start commit in scope** at either the call site or its caller.

## The second defect: `doc.IndexedSHA` has it too

The only commit capture on this path is `gi := gitmeta.Capture(absRepo)` at `index.go:842` — and that runs **after** `idx.Run(...)` at `:830`, i.e. after the walk and the whole extraction.

So the `graph.fb` header's `IndexedSHA` is stamped from a post-extraction HEAD read. **Fixing only the manifest would leave the manifest and the fb header disagreeing about the same graph** — arguably worse than both being wrong the same way, because the disagreement is what a debugging user would trust least.

That is why this cannot be a copy of #6474's fix. It requires a decision: does `gitmeta.Capture` move ahead of `idx.Run` so both stamps come from one instant, or does the manifest get its own capture and the fb header keep its current behaviour with the difference documented?

## Two format complications

- **`gitmeta.Capture` uses `rev-parse --short=12`** (`internal/gitmeta/gitmeta.go:307`) while `diff.headCommit` uses bare `--short` (~7 chars). Reusing `gi.SHA` for the manifest would silently change the manifest's short-form length — and `indexedCommitShortNoGit` surfaces that value verbatim to the statusline, so it is user-visible.
- **`gitmeta.Info` has no full-SHA field at all** (`gitmeta.go:126-137`). `SaveManifestAtCommit` requires short *and* full and documents "pass both or neither", so site A cannot source `GitCommitFull` from the existing capture regardless.

Recommendation: capture via the `diff` helpers rather than reusing `gi`, so the manifest's format is unchanged.

## Why this window is worse than the incremental one

The incremental sites' exposure is walk → extract → save. Site A's window is **the entire full index** — minutes on a large repo — *and* its capture is already post-extraction. A commit landing mid-pass mislabels essentially every time it happens, rather than occasionally.

## What consumes it

Established while grounding #6474: the one *functional* consumer is the #5710 head-advance detector at `internal/extractors/incremental.go:505` — an over-advanced label makes the next pass see `HEAD == manifest`, find zero changes, and return `Done: true` over a graph built from an earlier commit. Self-concealing; nothing re-requests the work.

The rest are reporting surfaces: `statuswriter.go:67` → statusline and dashboard, and `daemon/index_commit.go:55` → RPC status and `grafel_index_status`, which will report `AtHead: true` over a stale graph. No automatic reindex trigger is keyed on `AtHead`.

## Observability

Go-level test, not a fixture row — the golden harness covers extractor output and cannot express a manifest sidecar's stamp.

The window is timing-dependent so it needs a deterministic seam. `cmd/grafel` already has the precedent: the test-only `enrichmentOrderHook` at `index.go:728-734`. **Note its existing `graph_written` stage at `:966` fires *after* `commitManifest` at `:948`, so it is useless for this** — a new `git_captured` stage is needed.

**Killing mutant:** revert the threaded call to `diff.SaveManifest(...)`; the test must go red on the pass-start commit.

⚠️ **Vacuity trap, same as #6474:** against a plain `t.TempDir()` the stamp is unconditionally `""` and **the test passes against the defective code**. It needs a real git repo, an explicit `c1 != c2` assertion, and an assertion that the intended path actually ran.

cajasmota:
Closes #6502

A full index read HEAD **after** extraction and stamped both the manifest and the fb header from it. A commit landing mid-index was therefore recorded as the indexed commit, and the next incremental diffed against work the graph had never seen.

## RED

Written first, against a real git repo, using a new test-only `gitCaptureHook` seam that fires once — right after this pass reads HEAD, before extraction — so the test can move HEAD mid-pass deterministically rather than racing it:

```
manifest GitCommit     = "f8bb81d",        want pass-start "867a6fd"        (mid-pass HEAD "f8bb81d")
manifest GitCommitFull = "f8bb81d...024",  want pass-start "867a6fd...dc2"
header   IndexedSHA    = "f8bb81ddec94",   want pass-start "867a6fd052b5"
--- FAIL: TestFullIndex_StampsPassStartCommit_NotMidPassHEAD (4.30s)
```

GREEN: `--- PASS (1.15s)`, and the whole package — `ok cmd/grafel 129.313s`, no `-run` filter, exit 0.

## `doc.IndexedSHA` moves too — the decision the issue left open

Yes. `gitmeta.Capture` (and the `ProbeRepo` describing the same checkout) now runs before `idx.Run`; only the assignment onto `doc` stays after, since `doc` does not exist earlier.

Reason: fixing only the manifest would leave **the manifest and the fb header naming different commits for the same graph**. That disagreement is worse than both being wrong the same way — it is precisely the state a user debugging a stale index trusts least, because neither number can be believed and nothing says which is which.

## The `--short=12` vs bare `--short` mismatch

Resolved by moving **both producers** rather than sharing one value.

`gitmeta.Info` has no full-SHA field, and `SaveManifestAtCommit` requires short *and* full. Reusing `gi.SHA` for the manifest would silently change the manifest's short length — and `indexedCommitShortNoGit` surfaces that string **verbatim to the statusline**, so the change would have been user-visible for no reason.

So: the manifest keeps its 7+40 pair from a single `diff.HeadCommitPair` call, where short and full can never describe different commits; the header keeps its 12-character form; and both reads now sit adjacent at pass start.

## Why a new hook rather than an `enrichmentOrderHook` stage

The issue suggested a `git_captured` stage on the existing hook. That hook's **first event is asserted to be `graph_written`**, so prepending a stage would have rewritten an unrelated contract to buy a test seam. A separate `gitCaptureHook` costs less and breaks nothing.

## Mutants — 3 scored, 3 killed

All `go vet`-clean before scoring (a mutant that does not compile proves nothing), `-count=1` (a cached green is not evidence), restored by `cp` with shasum `2bea81fc…` re-verified after each — never `git checkout -- <file>`.

|---|---|---|
| revert both halves to the post-extraction read | killing | **KILLED** — all three assertions red |
| `SaveManifestAtCommit(dest, m, passShort, "")` — empty full SHA where "pass both or neither" forbids it | **permissive** | **KILLED** on `GitCommitFull` |
| short from live HEAD, full from pass start — a mismatched pair | **permissive** | **KILLED** on `GitCommit` |

## Context

The sibling incremental path was already fixed under #6474 (`internal/extractors/incremental.go:643` calls `SaveManifestAtCommit` with the pass-start commit). This was the **last site** where the two halves of that invariant disagreed, which is the strongest argument for the change — not the theoretical race, but that half the codebase had already decided the question.

## Verification

`gofmt -l` clean · `go vet` clean · `go test -count=1 ./cmd/grafel/` → 0, whole package, no `-run` filter, exit code captured separately.

### DIFF
--- a/cmd/grafel/full_index_commit_stamp_6502_test.go
+++ b/cmd/grafel/full_index_commit_stamp_6502_test.go
@@ -1,0 +1,139 @@
+package main
+
+// #6502 — a FULL index must label the graph it just built with the commit that
+// was checked out when the pass STARTED, in both places that label is written:
+// the diff manifest's GitCommit/GitCommitFull, and the graph header's
+// IndexedSHA.
+//
+// Before this change neither did. `gitmeta.Capture` ran AFTER `idx.Run`, so
+// IndexedSHA was a post-extraction read, and `commitManifest` reached
+// `diff.SaveManifest`, which reads live HEAD at save time — later still. A
+// commit landing during the pass (minutes, on a real repo) was therefore
+// recorded as the commit the graph was built from, for work it never saw. The
+// #5710 head-advance detector then sees HEAD == manifest on the next pass,
+// finds zero changes and returns Done over a graph built from an earlier
+// commit — self-concealing, because nothing re-requests the work.
+//
+// The sibling incremental sites were fixed in #6474
+// (internal/extractors/incremental.go passes its pass-start commit to
+// diff.SaveManifestAtCommit); this is the last unfixed one, so until now the
+// two halves of one invariant disagreed.
+//
+// VACUITY TRAP. Against a plain t.TempDir() every commit stamp is
+// unconditionally "" and this test would pass against the defective code. So
+// it uses a REAL git repo, asserts the mid-pass commit actually differs from
+// the pass-start one, and asserts the seam fired exactly once — no assertion
+// below is reachable without the intended path having run.
+
+import (
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
+	"testing"
+
+	"github.com/cajasmota/grafel/internal/graph"
+	"github.com/cajasmota/grafel/internal/indexer/diff"
+)
+
+// gitOut runs git in dir and returns trimmed stdout, failing the test on error.
+func gitOut(t *testing.T, dir string, args ...string) string {
+	t.Helper()
+	cmd := exec.Command("git", args...)
+	cmd.Dir = dir
+	cmd.Env = append(os.Environ(),
+		"GIT_CONFIG_GLOBAL=/dev/null", "GIT_CONFIG_SYSTEM=/dev/null",
+	)
+	out, err := cmd.Output()
+	if err != nil {
+		t.Fatalf("git %v in %s: %v", args, dir, err)
+	}
+	return strings.TrimSpace(string(out))
+}
+
+func TestFullIndex_StampsPassStartCommit_NotMidPassHEAD(t *testing.T) {
+	if _, err := exec.LookPath("git"); err != nil {
+		t.Skip("git unavailable")
+	}
+	root := t.TempDir()
+	// Never touch the real ~/.grafel or the running daemon's state.
+	t.Setenv("HOME", filepath.Join(root, "home"))
+	t.Setenv("USERPROFILE", filepath.Join(root, "home"))
+	t.Setenv("GRAFEL_HOME", filepath.Join(root, "grafelhome"))
+	t.Setenv("GRAFEL_DAEMON_ROOT", filepath.Join(root, "daemonroot"))
+
+	repo := filepath.Join(root, "repo")
+	if err := os.MkdirAll(filepath.Join(repo, "svc"), 0o755); err != nil {
+		t.Fatal(err)
+	}
+	writeFixtureFile(t, repo, "go.mod", "module stampfixture\n\ngo 1.21\n")
+	writeFixtureFile(t, repo, "svc/widget.go", "package svc\n\ntype Widget struct{ Name string }\n\nfunc (w *Widget) Render() string { return w.Name }\n")
+	seedGitRun(t, repo, "init", "-q", "-b", "main")
+	seedGitRun(t, repo, "add", "-A")
+	seedGitRun(t, repo, "commit", "-q", "-m", "fixture")
+
+	// c1 — the commit the graph is actually built from, in all three
+	// abbreviations the two stamps use.
+	c1Short, c1Full := diff.HeadCommitPair(repo)
+	c1Short12 := gitOut(t, repo, "rev-parse", "--short=12", "HEAD")
+	if c1Full == "" || c1Short == "" || c1Short12 == "" {
+		t.Fatalf("fixture is not a git repo (short=%q short12=%q full=%q) — every stamp "+
+			"below would be \"\" and this test would pass against the defect", c1Short, c1Short12, c1Full)
+	}
+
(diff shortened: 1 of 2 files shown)


cajasmota:
## Coordinator mutants — one equivalent, one SURVIVOR that is not

Two mutants nobody else ran, both aimed at the **non-git** arm rather than at the commit-timing arm the three existing mutants cover. `go vet` clean before each, `-count=1`, restored by `cp` with shasum `2bea81fca0251c6fd71d60f42fe6190eb56f6602` verified after both.

### M1 — live-HEAD fallback when the pass-start capture is empty: SURVIVED, and equivalent

```go
shortSHA, fullSHA := i.passHeadShort, i.passHeadFull
if shortSHA == "" || fullSHA == "" {
	shortSHA, fullSHA = idiff.HeadCommitPair(absRepo)   // "helpful" fallback
}
```

`ok cmd/grafel 137.865s`. **Equivalent**, for a stateable reason: the branch only fires when the pass-start pair came back empty, which happens on a **non-git directory** — where `HeadCommitPair` also returns empty. The two paths agree everywhere a portable fixture can reach. The only divergence is a transient git failure on a real repo, which nothing here can provoke, so it is recorded rather than pinned.

### M2 — placeholder instead of empty: SURVIVED, and this one is a real gap

```go
if shortSHA == "" { shortSHA = "0000000" }
if fullSHA  == "" { fullSHA  = "0000000000000000000000000000000000000000" }
```

`ok cmd/grafel 131.525s`. **Not equivalent** — on a non-git directory the manifest now carries a fabricated commit where the code says it must carry none.

The claim is stated twice and observed nowhere:

> *Empty on a non-git directory, which `SaveManifestAtCommit` writes through as empty — **the correct default**.*

Nothing asserts it. And the consequence is not cosmetic: the same comment explains that the **#5710 head-advance detector compares HEAD against the manifest**. A manifest carrying `0000000` for a directory that has no HEAD is a different input to that comparison than an absent value, on a path where there is no git to correct it. That is the silent-wrongness direction — the detector reports on a graph whose provenance is fabricated.

This is the **sixth** instance this milestone of the same shape: a doc comment asserting a property, no fixture exercising the axis, mutant survives.

## What round 2 needs

One test running a full index against a **non-git** directory, asserting the manifest's `GitCommit` and `GitCommitFull` are **empty** — not merely "not equal to some SHA", since a placeholder would satisfy that. Then re-score M2; it must die.

**Do not add a placeholder to make the test convenient.** Empty is the documented and correct behaviour; the test exists to pin it, not to renegotiate it.

## Everything else on this PR holds

The three existing mutants are genuinely dead, the RED reproduces the real defect at all three stamping points, and the `doc.IndexedSHA` decision is right for the reason given — a manifest and fb header naming different commits for one graph is worse than both being wrong the same way. The choice to add a separate `gitCaptureHook` rather than prepend a stage to `enrichmentOrderHook`, whose first event is asserted to be `graph_written`, also avoids rewriting an unrelated contract to buy a test seam.

Issue #6433, opened 2026-08-20
Angular: HttpClient call sites not extracted, so no frontend→endpoint cross-stack links


auxmedrano:
**Impact:** Cross-stack linkage (frontend/backend edges not connected)
**grafel version:** v0.2.2-40-g6f14cb873 (dev build from source, `make build-go-only`)

### Summary

On a real Angular frontend that talks to a Spring backend in the same monorepo, grafel recognizes the Angular framework and indexes the components/services, but it extracts **zero outbound HTTP call sites** from the Angular code. `HttpClient` requests (`this.http.get/post/patch/...`) never become `SCOPE.ExternalAPI` (or any HTTP-call) nodes, so there is nothing to link to the backend `http_endpoint_definition`s — the cross-stack "frontend calls this endpoint" graph is empty.

### Expected vs. actual

- **Expected:** each `HttpClient` call site becomes an outbound-HTTP node (e.g. `SCOPE.ExternalAPI`) carrying its verb + path, so it can be matched/linked to the backend endpoint that serves it.
- **Actual:** `SCOPE.ExternalAPI` count in the whole group = **98, all in the backend, 0 in the frontend**, despite the frontend making HTTP calls throughout. No HTTP-client edge kind is emitted for the Angular side, and no frontend→endpoint links exist.

### Evidence (genericized)

- **42 frontend files** use `HttpClient` with concrete calls, e.g.:
  ```ts
  this.http.get<readonly Thing[]>(base(code))
  this.http.post<Thing>(`${BASE}/${id}`, body)
  this.http.patch<Thing>(`${BASE}/${id}`, patch)
  ```
- Framework recognition **does** tag the app as `angular` (confirmed in `grafel feedback` → Framework Recognition).
- Graph facts:
  - `SCOPE.ExternalAPI` total **98** → frontend **0**, backend **98**.
  - No `-[:CALLS_ENDPOINT]->` / HTTP-client edge kind anywhere.
  - Consequently `frontend/ → http_endpoint_definition` edges: **0** (only 2 incidental non-HTTP edges exist between the two trees).

### Why it matters

This is the frontend half of end-to-end request tracing. Even with the backend endpoints fully mapped (94 routes extracted), you cannot answer "which screen calls this endpoint" or "what breaks in the UI if I change this route's payload" — grafel's headline cross-stack question — because the client call sites are never modeled. It is the companion to the backend-side Spring route→handler resolution gap reported separately.

### Notes that may bear on the fix

[...]

Issue #6451, opened 2026-08-20
decision needed: SCOPE.ExternalAPI joins nothing — normalise it, drop it, or document it as terminal (angular-http-mini now baselines the third)


cajasmota:
`SCOPE.ExternalAPI` is minted by `internal/extractors/cross/httpclient/extractor.go:423` and **participates in no matcher**. Zero occurrences of `ExternalAPI` anywhere in `internal/links/`. It is read in one place — `process_flow.go:1004` — as a **terminal** node.

So it is a parallel, non-joining representation of the same call sites that `http_endpoint_call` represents, and only the latter can ever reach a `http_endpoint_definition`.

## Why this is worth naming as a defect rather than a quirk

It is the metric users measure. In #6433 the reporter's headline evidence was `SCOPE.ExternalAPI = 98, all backend, zero frontend`. Those 98 nodes join nothing. A user counting them reasonably concludes the graph knows about 98 outbound calls; the graph's linker has never heard of them.

The two producers also disagree on **name format** for the same source line. `` this.http.get(`/api/things/${id}`) `` yields:

- `SCOPE.ExternalAPI` named `/api/things/{*}` — the `{*}` wildcard dialect is Path-1-only and no join understands it (`extractor.go:40-47`)
- `http_endpoint_call` named `http:GET:/api/things/{id}` — the canonical form

Both are asserted in the same fixture. And the caller granularity differs too: Path 1's `CALLS` edge originates from a file stub `scope:component:http_caller:<file>` (`extractor.go:395-401`), while synthesis emits `FETCHES` from a function. Two edges for one call, never reconciled, and the only dedup in the system is `patternType + "\x00" + id` **within a single synthesis invocation** (`http_endpoint_synthesis.go:500`) — so anything counting "HTTP calls" across kinds double-counts.

`internal/extractors/cross/endpoint/extractor.go:161-163` has the same problem from the other side: `SCOPE.Operation` named `GET /users/{id}`, space-separated with a file-scoped ID. Also joins nothing.

## The decision this needs

Not "add a third index". The question is whether `SCOPE.ExternalAPI` should:

1. **feed the matcher** — normalise its name to the canonical `http:VERB:/path` form so it can join; or
2. **be derived from** `http_endpoint_call` rather than independently extracted, so one producer feeds both views; or
3. **be documented as deliberately terminal** — a "this code calls the outside world" annotation that is *not* a cross-stack participant — and have consumers told so.

[...]

PR #6552, opened 2026-08-22
fix(#6551): fold same-file string-const URLs in Angular http client calls

### LINKED ISSUE #6551: Angular: module-level URL constants not folded, so every frontend→endpoint cross-stack link orphans (#6433 extracts the call, but never binds)
## Summary

On a Spring (Java) + Angular (TypeScript) monorepo, **frontend→endpoint cross-stack links never resolve because the Angular URL extractor does not fold module-level string constants** used as the request path. `#6433` fixed *call-site extraction* (the calls are now emitted), but every call addressed through a `const` lands as an orphan `/{dynamic}/<CONST_NAME>`, so it can't be matched to the backend route it literally names.

This is a very common Angular idiom (`const BASE = '/api/...'; this.http.get(BASE)`), so in practice **0 of the frontend calls resolve** even though both tiers are fully indexed.

Tested on **grafel v0.3.0** (commit b037c3f) after a full `grafel index`.

## Environment

- grafel `v0.3.0`
- Repo: single-repo group, Spring Boot backend (`@RestController`, 94 endpoints extracted, 89 via AST) + Angular admin portal.
- Full reindex (`grafel index <repo>`), graph current.

## What works (v0.3.0)

- `#6433`: Angular `HttpClient` call sites are extracted — `angular_http_client: {calls: 29}` (was 0 on v0.2.2).
- `#6429`: Spring routes resolve to their handlers — e.g. `http:GET:/api/admin/v1/people` → `PersonController.java:76`, no dead-ends.

## The bug: module-level URL constants are not folded

`endpoints stats` after a clean index:

```
angular_http_client: calls 29 | nestjs_http_service: calls 37
spring: def 5 | spring_mvc: def 89
totals: definitions 94, calls 66, orphan_calls 34, cross_repo_resolved 0, cross_repo_links 0
```

Every orphan is a `dynamic_kind: env-var-baseurl` (or `leading-path-placeholder`) whose `static_path_suffix` is a **constant identifier**, not the literal it holds.

### Minimal repro

`frontend/admin-portal/src/app/features/people/people-api.ts`:

```ts
const PEOPLE = '/api/admin/v1/people';                 // module-level string literal
// ...
return this.http.get<Person>(PEOPLE, { params });      // grafel records:  GET /{dynamic}/PEOPLE
return this.http.get<Person>(`${PEOPLE}/${personId}`); //                  (suffix = the const NAME)
```

Backend endpoint that exists in the same graph:

```
http:GET:/api/admin/v1/people   backend/.../PersonController.java:76   (id ea820a0a7dbda170)
```

Expected: the call binds to `/api/admin/v1/people` (fold `PEOPLE` → its literal, then match the route).
Actual: the call stays `GET /{dynamic}/PEOPLE`, `orphan_calls`, no link.

### Scope on this repo

38 orphaned HTTP calls. The dominant, cleanly-foldable pattern is a same-file `const`:

| const (in `*-api.ts`) | literal value | files using it |
|---|---|---|
| `BASE` | `/api/admin/v1/<feature>` | users, units, billing-cycles, charge-concepts, extraordinary-charges, overdue-payments-report, … |
| `PEOPLE` | `/api/admin/v1/people` | people |
| `ESPACIOS` | `/api/admin/v1/rental-spaces` | rentals |
| `TIPOS_UNIDAD` | `/api/admin/v1/unit-types` | rentals, extraordinary-charges |
| `LOTS` / `DEVICES` | `/api/admin/v1/access-device-lots` /`…/access-devices` | access-devices |

All are `const X = '/literal'` at module scope in the same file as the call — no environment indirection, no interpolation of the base. A constant-folding pass over same-file `const … = '<string>'` would resolve the large majority.

(Two residuals are genuinely interprocedural — `this.occupanciesDe(unitId)` returns a URL from a private method — and legitimately need runtime/`resolve` treatment. Those are not part of this report.)

## Additional findings while trying to resolve these via the repair workflow

Attempted to land the links manually via `grafel_docgen_apply(kind=repairs, bind_to_entity)` (the `grafel-resolve` flow). Two rough edges:

1. **`grafel_find` entity_id ≠ the id `bind_to_entity` accepts.** `grafel_find(substring, "http:GET:/api/admin/v1/people")` returns `entity_id: "vf::605348763bd011e5"`; submitting that (and its bare form) as `target_entity_id` is rejected `target_entity_not_found`. The id that works is the one from `grafel_inspect` (`ea820a0a7dbda170`) — a different value entirely, not just a prefix difference. Scripting the resolve flow off `grafel_find` therefore always fails until you also `inspect` each target.

2. **`bind_to_entity` repairs for `dynamic_baseurl_endpoint` residuals go stale and never apply.** Submitted 12 binds (all `ok:true`, `repair_count` 1→12, persisted to `refs/_unknown/repair.json`). Re-indexed with `--enable-repair-candidates --enable-repair-apply`. Result: `repair_stats.json` → `applied: 0, stale: 12` — every repair `R-stale` ("edge_id no longer matches any current candidate"), so `orphan_calls` stayed 34 and `cross_repo_resolved` stayed 0. The documented "resolve → `grafel index` to apply" loop does not materialize these links. (Repo has no `.git`, so the ref is `_unknown` — possibly relevant to the candidate-id regeneration.)

Net: on this stack there is currently **no path** — automatic or via the repair workflow — to get frontend→endpoint links, even though every call names its route through a plain constant.

## Suggested fix

Fold same-file module-level `const <NAME> = '<string-literal>'` (and simple `` `${NAME}/...` `` interpolations) when extracting Angular `HttpClient` URLs, so the path is the literal rather than the identifier. Secondarily, stabilize `dynamic_baseurl_endpoint` candidate ids across reindex (or fix the `_unknown`-ref case) so submitted binds don't go stale, and make `bind_to_entity` accept the `grafel_find` id (or document that only the `inspect` id is valid).

auxmedrano:
Fixes #6551.

## Problem

#6433 got Angular `this.http.<verb>(...)` **call sites** extracted, but a call whose URL argument is a **bare identifier bound to a same-file string-literal constant** — the idiomatic Angular api-service shape — fell straight through to the `/{dynamic}` marker:

```ts
const PEOPLE = '/api/admin/v1/people';
// ...
this.http.get<Person>(PEOPLE, { params });   // emitted as GET /{dynamic}/PEOPLE
this.http.post<Person>(PEOPLE, body);        //             POST /{dynamic}/PEOPLE
```

On a real Spring + Angular monorepo this left **0 cross-stack links** even though both tiers were fully indexed and every backend route existed — the frontend calls named their routes through path constants (`BASE`, `PEOPLE`, `ESPACIOS`, `LOTS`, …) and none folded.

## Why this is a small change

The machinery already exists. `buildJSConstantSymbolTable` captures exactly these `const NAME = '<string>'` declarations, and its result is **already threaded into** `synthesizeReceiverClientResidualCalls`. The string-literal and template-literal arms of that function's `switch` already consult it — only the **bare-identifier arm was missing**, so the lookup never happened for `get(PEOPLE)`.

This adds a `default:` arm that folds the identifier through the same symbol table before emitting the dynamic marker.

## On #6446's deliberate choice

The #6446 comment states base-URL constant folding was *deliberately not attempted*, on the reasoning that a dynamic-marked node is "reviewable via the repair flow." In practice that path doesn't currently deliver the link: submitting `bind_to_entity` repairs for these `dynamic_baseurl_endpoint` residuals and re-indexing marks them **`applied: 0, stale: N`** (details in #6551). So for the common same-file string-const case, folding at extraction time is the reliable way to get the link. This change is deliberately narrow:

- Only **same-file string-literal** consts fold (what `buildJSConstantSymbolTable` already captures).
- A **non-literal** const (`const BASE = environment.apiBase`) is *not* captured and stays dynamic — so #6433's `DynamicArg` guarantee is unchanged (that test still passes).
- Method-return URLs (`this.occupanciesDe(unitId)`) and local-variable bases stay dynamic — correctly left for the repair flow.

## Measured on the reporter's repo (Spring + Angular)

| | before | after |
|---|---|---|
| `/{dynamic}` endpoint orphans | 19 | **9** |
| frontend calls resolved | few | **57 / 66** |
| frontend→backend `FETCHES` edges | **0** | **46** |

## Tests

- New: `TestSynth_AngularHttpClient_StringConstURL_6551` — asserts `get(PEOPLE)`/`post(PEOPLE)` fold to `http:GET|POST:/api/admin/v1/people` and are not left `{dynamic}`.
- Existing `internal/engine` suite passes unchanged (incl. `TestSynth_AngularHttpClient_DynamicArg_6433`, which pins that non-literal consts stay dynamic).

Verified with `CGO_ENABLED=1 go test ./internal/engine/`.

### DIFF
--- a/internal/engine/http_endpoint_angular_constfold_6552_test.go
+++ b/internal/engine/http_endpoint_angular_constfold_6552_test.go
@@ -1,0 +1,200 @@
+// Scope-safety tests for the bare-identifier constant fold added by #6551.
+//
+// The fold itself is right: `const PEOPLE = '/api/admin/v1/people';
+// http.get(PEOPLE)` should resolve, and before #6551 it dangled at
+// /{dynamic}/PEOPLE. What was wrong was the BINDING-SELECTION rule. The fold
+// read buildJSConstantSymbolTable, a regex table over raw file text that keeps
+// the FIRST `(?:const|let|var) NAME = '...'` match per name, never strips
+// comments, and has no notion of scope.
+//
+// The older consumers of that table tolerated all of this because they only
+// ever REFINED a path that was already partly literal. The bare-identifier arm
+// makes the table the sole source of truth for a call that was otherwise an
+// honest /{dynamic} marker, so a wrong binding ships as a confident
+// runtime_dynamic=false endpoint and the cross-stack linker FETCHES-links it to
+// a route the call never makes.
+//
+// Each test below is a shape where the regex table binds the wrong value, and
+// where declining is strictly better than answering: `main` emitted
+// /{dynamic}, which is honest.
+package engine
+
+import (
+	"strings"
+	"testing"
+)
+
+// requireNoStaticFold fails when any emitted id resolves to a concrete path
+// under one of the given prefixes — i.e. the fold answered when it should have
+// declined. A /{dynamic} id is the correct outcome for every caller here.
+func requireNoStaticFold(t *testing.T, got []string, badPaths []string, label string) {
+	t.Helper()
+	for _, id := range got {
+		for _, bad := range badPaths {
+			if strings.Contains(id, bad) && !strings.Contains(id, "{dynamic}") {
+				t.Errorf("%s: folded a binding it could not prove: %q (got=%v)", label, id, got)
+			}
+		}
+	}
+}
+
+// requireAnyDynamic asserts the call site was still emitted, just honestly.
+func requireAnyDynamic(t *testing.T, got []string, label string) {
+	t.Helper()
+	for _, id := range got {
+		if strings.Contains(id, "{dynamic}") {
+			return
+		}
+	}
+	t.Errorf("%s: expected the unresolvable call to survive as a {dynamic} marker, got=%v", label, got)
+}
+
+// A — a commented-out declaration wins. tree-sitter never yields a `comment`
+// node as a declaration; a first-match-wins regex over raw text does.
+func TestSynth_AngularConstFold_CommentedOutDeclaration_6552(t *testing.T) {
+	src := `
+import { Injectable, inject } from '@angular/core';
+import { HttpClient } from '@angular/common/http';
+
+// const PEOPLE = '/api/OLD/people';        // dead, commented out
+const PEOPLE = '/api/admin/v1/people';      // live
+
+@Injectable({ providedIn: 'root' })
+export class PeopleService {
+  private http = inject(HttpClient);
+  a() { return this.http.get<any>(PEOPLE); }
+}
+`
+	got, _ := runDetect(t, "typescript", "people.service.ts", src)
+	requireNoStaticFold(t, got, []string{"/api/OLD/people"}, "commented-out-decl")
+	requireContains(t, got, []string{"http:GET:/api/admin/v1/people"}, "commented-out-decl-live")
+}
+
+// D — the identifier is not in scope at the call at all: declared inside an
+// unrelated class's method body, referenced from a different class where the
+// name is undefined. A per-file flat map cannot see that; a scope tree can.
+func TestSynth_AngularConstFold_OutOfScopeDeclaration_6552(t *testing.T) {
+	src := `
+import { Injectable, inject } from '@angular/core';
+import { HttpClient } from '@angular/common/http';
+
+export class UnrelatedService {
+  build() {
+    const endpoint = '/api/unrelated/thing';
+    return endpoint;
+  }
+}
+
+@Injectable({ providedIn: 'root' })
+export class CallerService {
+  private http = inject(HttpClient);
+  a() { return this.http.get<any>(endpoint); }
+}
+`
+	got, _ := runDetect(t, "typescript", "caller.service.ts", src)
+	requireNoStaticFold(t, got, []string{"/api/unrelated/thing"}, "out-of-scope-decl")
+	requireAnyDynamic(t, got, "out-of-scope-decl")
+}
+
+// C — a reassignable binding. `let url = '/api/first'; url = '/api/second';`
+// folds the stale first value. A `let` can never be folded safely, and
+// jsConstStringRe literally alternates (?:const|let|var).
+func TestSynth_AngularConstFold_ReassignedLet_6552(t *testing.T) {
+	src := `
+import { Injectable, inject } from '@angular/core';
+import { HttpClient } from '@angular/common/http';
+
+@Injectable({ providedIn: 'root' })
+export class ReassignService {
+  private http = inject(HttpClient);
+  a() {
+    let url = '/api/first';
+    url = '/api/second';
+    return this.http.get<any>(url);
+  }
+}
+`
+	got, _ := runDetect(t, "typescript", "reassign.service.ts", src)
+	requireNoStaticFold(t, got, []string{"/api/first", "/api/second"}, "reassigned-let")
+	requireAnyDynamic(t, got, "reassigned-let")
+}
+
+// H — a module-level const shadowed by a method PARAMETER of the same name.
+// At the call site `path` is the parameter, whose value is a caller's runtime
+// argument; the module const is not what is passed.
+func TestSynth_AngularConstFold_ShadowedByParameter_6552(t *testing.T) {
+	src := `
+import { Injectable, inject } from '@angular/core';
+import { HttpClient } from '@angular/common/http';
+
+const path = '/api/module/level';
+
+@Injectable({ providedIn: 'root' })
+export class ShadowService {
+  private http = inject(HttpClient);
+  a(path: string) { return this.http.get<any>(path); }
+}
+`
+	got, _ := runDetect(t, "typescript", "shadow.service.ts", src)
+	requireNoStaticFold(t, got, []string{"/api/module/level"}, "shadowed-by-parameter")
+	requireAnyDynamic(t, got, "shadowed-by-parameter")
+}
+
+// A class FIELD is not the module const of the same name. `this.url` and `url`
+// are different bindings, and a class field's initialiser is not something this
+// fold tracks. The review of #6552 named this exact move — trimming a `this.`
+// prefix and retrying the lookup — as a permissive mutant that left the whole
+// suite green while emitting a wrong value; this test is what kills it.
+func TestSynth_AngularConstFold_ThisFieldIsNotTheModuleConst_6552(t *testing.T) {
+	src := `
+import { Injectable, inject } from '@angular/core';
+import { HttpClient } from '@angular/common/http';
+
+const url = '/api/module/const';
+
+@Injectable({ providedIn: 'root' })
+export class FieldService {
+  private http = inject(HttpClient);
+  private url = '/api/field/value';
+  a() { return this.http.get<any>(this.url); }
+}
+`
+	got, _ := runDetect(t, "typescript", "field.service.ts", src)
+	requireNoStaticFold(t, got, []string{"/api/module/const", "/api/field/value"}, "this-field-not-module-const")
+	requireAnyDynamic(t, got, "this-field-not-module-const")
+}
+
+// Regression guard for what already correctly DECLINES and must keep
+// declining: a concatenation, and a same-named const in ANOTHER file (the
+// table is per-file, built at http_endpoint_client_synthesis.go:1707).
+func TestSynth_AngularConstFold_KeepsDeclining_6552(t *testing.T) {
+	concat := `
+import { Injectable, inject } from '@angular/core';
+import { HttpClient } from '@angular/common/http';
+
+const base = '/api/base';
+
+@Injectable({ providedIn: 'root' })
+export class ConcatService {
+  private http = inject(HttpClient);
+  a() { return this.http.get<any>(base + '/things'); }
+}
+`
+	got, _ := runDetect(t, "typescript", "concat.service.ts", concat)
+	requireNoStaticFold(t, got, []string{"/api/base"}, "concat-declines")
+	requireAnyDynamic(t, got, "concat-declines")
+
+	otherFile := `
+import { Injectable, inject } from '@angular/core';
+import { HttpClient } from '@angular/common/http';
+
+@Injectable({ providedIn: 'root' })
+export class OtherFileService {
+  private http = inject(HttpClient);
+  a() { return this.http.get<any>(PEOPLE); }
+}
+`
+	got2, _ := runDetect(t, "typescript", "otherfile.service.ts", otherFile)
+	requireNoStaticFold(t, got2, []string{"/api/admin/v1/people"}, "cross-file-declines")
+	requireAnyDynamic(t, got2, "cross-file-declines")
+}
--- a/internal/engine/http_endpoint_angular_httpclient_6433_test.go
+++ b/internal/engine/http_endpoint_angular_httpclient_6433_test.go
@@ -60,6 +60,54 @@ func TestSynth_AngularHttpClient_StaticURL_6433(t *testing.T) {
 	}, "angular-httpclient-static")
 }
 
+// TestSynth_AngularHttpClient_StringConstURL_6551 — the URL argument is a bare
+// identifier bound to a same-file string-literal constant, the idiomatic Angular
+// api-service shape:
+//
+//	const PEOPLE = '/api/admin/v1/people';
+//	this.http.get<Person>(PEOPLE, opts);
+//	this.http.post<Person>(PEOPLE, body);
+//
+// Reported by @auxmedrano (#6551): #6433 got the call site extracted, but a bare
+// identifier fell straight through to the /{dynamic} marker even when its value
+// was a statically-known path, so every call routed through a path constant
+// dangled instead of linking to its producer endpoint. The symbol table
+// (buildJSConstantSymbolTable) already captures such consts and is already
+// threaded into the receiver residual pass; this folds the identifier through it.
+// A non-literal const (const BASE = environment.apiBase) is NOT captured and
+// stays dynamic — see TestSynth_AngularHttpClient_DynamicArg_6433.
+func TestSynth_AngularHttpClient_StringConstURL_6551(t *testing.T) {
+	src := `
+import { Injectable, inject } from '@angular/core';
+import { HttpClient, HttpParams } from '@angular/common/http';
+
+const PEOPLE = '/api/admin/v1/people';
+
+@Injectable({ providedIn: 'root' })
+export class PeopleService {
+  private http = inject(HttpClient);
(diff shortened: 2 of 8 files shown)


cajasmota:
Thank you — and this one lands on a question that is currently **open and escalated**, which makes it more useful than a standalone fix.

Folding a same-file string constant before the URL is read is exactly the missing step behind a class of unresolvable client calls: `private url = '/api/things'` followed by `http.get(this.url)` previously produced a call whose target could not be matched to any endpoint. Constant-folding it at the client-synthesis site is the right place — it keeps the fold local and same-file, so it cannot reach across files and invent a match.

Where this gets interesting: **#6451** is an open decision about `SCOPE.ExternalAPI` and the `{*}` wildcard sentinel — entities minted for client calls whose URL could not be resolved, which then join nothing. Every URL your change folds is one fewer `{*}` entity. So this may shrink the population that decision is about, and possibly change which option is right. I will re-measure `SCOPE.ExternalAPI` counts against this branch before that decision is made, and say so on #6451 either way.

CI is green on all legs. An independent review is running — that is our standard path for every change including our own, not a comment on yours. I will post its result here.

One thing worth saying plainly: filing the issue with the mechanism already isolated, then the fix with a test beside it, is the shape that makes a change reviewable in one pass. It is appreciated.

cajasmota:
Review is back, and I owe you a correction before the findings.

**I described this PR wrongly in my comment above.** I said it fixes `private url = '/api/things'` + `http.get(this.url)`. It does not — that shape still emits `/{dynamic}/this.url` on this branch, unchanged. What the PR actually folds is a **bare identifier** (`const PEOPLE = '/api/admin/v1/people'; http.get(PEOPLE)`). My example was wrong and I should have checked it against the diff rather than the title.

**The idea is right and the change does work for the clean single-declaration case.** The objection below is the *binding-selection rule*, not the approach — and the measured gains in your PR body are credible.

## Blocking: the symbol table is scope-blind, and the new arm makes that consequential

`buildJSConstantSymbolTable` (`http_endpoint_client_synthesis.go:741`) runs `jsConstStringRe` over **raw file text**, keeps the **first** match per name, does not strip comments, and accepts `let`/`var` as well as `const`.

The string and template arms already tolerated that, because they only ever *refined* a path that was already partly literal. **The new bare-identifier arm makes the table the sole source of truth** for a call that was previously an honest `/{dynamic}` marker. A wrong binding now ships as a confident `runtime_dynamic=false` endpoint, and the cross-stack linker will FETCHES-link it to a route the call never hits.

**A — a commented-out declaration wins:**

```ts
// const PEOPLE = '/api/OLD/people';        // dead, commented out
const PEOPLE = '/api/admin/v1/people';      // live
@Injectable() export class S {
  private http = inject(HttpClient);
  a() { return this.http.get<any>(PEOPLE); }
}
```

Base: `/{dynamic}/PEOPLE`. This branch: **`http:GET:/api/OLD/people`**, `runtime_dynamic=false`.

**D — an identifier that is not in scope at all:** `const endpoint = '/api/unrelated/thing'` inside an unrelated class's method, with `this.http.get(endpoint)` in a *different* class where `endpoint` is undefined. Base declines; this branch emits `http:GET:/api/unrelated/thing`.

Also wrong, less severe: `let url = '/api/first'; url = '/api/second';` folds to the stale first value; a module-level `const` shadowed by a method **parameter** of the same name folds the parameter.

What correctly **declines**, and is worth keeping: `base + '/things'`, `` `${BASE}/spaces` ``, and a same-named const in another file — the table is genuinely per-file (`http_endpoint_client_synthesis.go:1707`).

## What I would ask for

1. Build the symbol table over **comment-blanked** source.
2. Restrict to `const` — drop `let`/`var`, since a reassignable binding cannot be folded safely.
3. **Decline the fold when the name is declared more than once in the file.** That one alone kills A, C and H, and it fails safe: the call goes back to being an honest `/{dynamic}` marker rather than a confident wrong answer.

A surviving permissive mutant confirms the gap is real rather than theoretical: adding a `this.` trim plus a leading-slash retry in the new `default:` arm leaves the entire suite green while binding a class field to a module const of the same name and emitting the wrong value.

## One process note

`angular-http-mini/expected.json` still states in prose that *"Base-URL constant folding is deliberately NOT attempted."* This PR flips that documented invariant without touching the fixture, so nothing grades the new behaviour. A fixture row would make the change observable — and would have caught A.

## Your fix does not affect #6451, and that is useful to know

I said I would measure `SCOPE.ExternalAPI` before deciding that issue. Results:

| corpus | base | this branch |
|---|---|---|
| 26 golden fixtures — ExternalAPI total | 6 | 6 |
| 26 golden fixtures — of which `{*}` wildcard | **3** | **3** |
| reporter-shape probe fixture | 1 (`{*}/{*}`) | 1 (`{*}/{*}`) |

**No movement.** `SCOPE.ExternalAPI` is minted by `internal/extractors/cross/httpclient/extractor.go`, which by design never mints a node for a non-literal URL argument — so those calls contributed zero ExternalAPI entities before your change. The population #6451 is about is untouched either way. That resolves a question I had flagged as possibly blocking it, so thank you: the answer just came from a different direction than I expected.

Happy to push the comment-blanking and duplicate-declaration guard to your branch if you would rather not context-switch — say the word.

cajasmota:
Reworked the binding lookup onto the AST and pushed it as `723bb69f1`. Your commit is untouched — one commit on top, no rebase, squash, amend or force-push, and the message credits your diagnosis.

**Your idea was right; the symbol table underneath it was the problem.** Rather than patch the regex, the fold now reads a scoped table built from tree-sitter `lexical_declaration` nodes. That was cheaper than it sounds: `internal/engine/http_endpoint_client_ast.go:85` already parses the file, in this package, on the same `content`, and `http_endpoint_client_synthesis.go:1695-1696` runs that pass immediately before the regex one. The tree was one function call away.

## What that fixes structurally rather than by patching

New `internal/engine/http_endpoint_client_constscope_6552.go`: `buildJSScopedConstTable` walks `lexical_declaration` nodes and records **every** binding — parameters, imports, destructuring, function and class names — with the byte span of the scope it governs. Non-foldable bindings are recorded anyway, precisely so they can still **shadow**. `resolve` takes the innermost binding visible at the call's byte offset, and declines on a tie, a non-`const`, a non-string initialiser, any assignment to the name, or an unparseable file.

Tree-sitter never yields a comment as a declaration, the node exposes `const` vs `let`/`var`, and byte offsets make shadowing decidable — so all four defects go away as a consequence of the representation rather than as four special cases.

Built lazily: a file with no bare-identifier argument pays no second parse.

RED before, all four of your reported shapes:

```
CommentedOutDeclaration  folded "http:GET:/api/OLD/people"
OutOfScopeDeclaration    folded "http:GET:/api/unrelated/thing"
ReassignedLet            folded "http:GET:/api/first"
ShadowedByParameter      folded "http:GET:/api/module/level"
```

GREEN after, with the clean case still folding and `TestSynth_AngularHttpClient_DynamicArg_6433` still passing.

## Scope discipline — your five fellow consumers are untouched

`buildJSConstantSymbolTable` is also read by `sse_edges.go:186`, `websocket_edges.go:660` and `http_endpoint_client_synthesis.go:1387`, `:1445`, `:1707`. The AST table sits **alongside** and is wired only to the new bare-identifier arm.

That is deliberate: those five only ever *refine* an already-literal path, so the loose table degrades an answer they were giving anyway — and their `let`/`var` arm is load-bearing for the template-literal folds. Byte-identical by construction beat proving it by measurement.

## Mutants

| mutant | verdict |
|---|---|
| point the arm back at the regex table | **DEAD** — 4 unit tests, 2 forbidden fixture hits, both recalls regressed |
| prefer the **outermost** visible binding (permissive) | **DEAD** — `ShadowedByParameter` |
| trim a `this.` prefix before lookup (permissive) | **DEAD** — via a new test |

That third one is the mutant the review reported as **surviving** on the original PR. It now dies.

## The fixture grades it now

`angular-http-mini`: entities **21/21 → 23/23**, relationships **17/17 → 18/18**, 0 forbidden hits. Baseline re-recorded through the writer, ratchet green across all 26 gated fixtures.

The `expected.json` note saying *"Base-URL constant folding is deliberately NOT attempted"* now describes what actually happens, and `app/const-fold.service.ts` grades it with two must-have entities, a FETCHES row, and **two forbidden rows** for your shapes A and H — so the defects you found are now permanently pinned rather than just fixed.

One implementation detail worth knowing: the new fixture method is `people()` rather than `list()`, because a second `list` SCOPE.Operation made an existing #6447 caller-attribution row ambiguous and turned it red. Noted in the fixture source.

## On the earlier round-trip

I asked you for changes when I should have just made them — that was the wrong call and the norm here now is that the maintainer does this work. Thank you for the report and the reproduction; the four shapes you found are what made the rework testable.

auxmedrano:
Independently verified `723bb69f1` before reading past the diff — not just the write-up:

- `go build ./...` clean, `go vet ./internal/engine/...` clean
- Full `internal/engine` and `internal/quality` suites green
- Ran `grafel quality` myself against `angular-http-mini`: **23/23 entities, 18/18 relationships, 0 forbidden hits** — matches exactly

The AST rework is the right call — scoping this off real `lexical_declaration`/parameter/scope nodes instead of patching the regex is what actually closes shape D (out-of-scope identifier), which I didn't have a clean answer for. Good outcome all around. Thanks for taking it the rest of the way.

cajasmota:
Ran one more mutant before merging — one nobody else had tried — and it **survived**. Reporting it rather than rounding it down, because a surviving permissive mutant is how every serious defect this milestone got in.

**The mutant:** drop the assignment tracker from `resolve`, so a name reassigned anywhere in the file still folds its binding.

```go
if t == nil || name == "" {   // was: || t.assigned[name]
```

`go vet` clean, then the full unfiltered `go test -count=1 ./internal/engine/` → **EXIT=0**. Green. The check is unobserved.

## Chasing it down turned up a false claim in my own comment

I wrote, on the decline-reason list, that the tracker is what declines shape C (`let url = '/api/first'; url = '/api/second'`). It is not. Measured directly against the table:

```
shapeC: assigned=true  anyFoldableBinding=false  resolve="",false
```

`foldable` is only ever set when the declaration is a `const` (`recordDeclarators`). Shape C's binding is a `let`, so it is already non-foldable and dies on `!best.foldable` — **before the tracker is consulted at all.** My attribution was backwards.

And because `foldable` implies `const`, and valid JS cannot assign to a `const` binding, the tracker's only *reachable* effect is the opposite of what the comment claimed — it over-declines:

```ts
const base = '/api/v1';
class S { go() { return this.http.get(base); } }
function other() { let base; base = '/api/v2'; }   // unrelated binding
```

```
assigned[base]=true  resolve="",false
```

A correct module-level fold, suppressed because an unrelated function happens to assign its own same-named `let`. The tracker carries no scope span, so it is file-global by construction.

## Verdict: equivalent mutant, and the check stays

It is equivalent under the current suite, and not because the suite is weak — it is redundant *by construction* with the const-only rule. So there is no fixture to write that would kill it honestly, and manufacturing one would be pinning a coincidence.

I kept the check anyway. An over-decline costs a `/{dynamic}` path, which is honest and is what `main` emitted; a mis-fold invents an endpoint that does not exist. That asymmetry is the whole doctrine of this PR and it is worth some redundancy.

What I did fix is the comment — pushed as `70d13b621`, **comments only**, verified mechanically (every changed line matches `^[+-]\s*//`), `go vet` clean, `go test -count=1 ./internal/engine/` EXIT=0. It now credits shape C to the branch that actually declines it, states the tracker's real (over-declining) effect, and records that dropping it leaves the suite green — so a future reader who runs the same mutant finds the reasoning already there instead of deleting a fail-safe on a green run.

Your commits are untouched: commits added on top, no rebase, no squash, no amend, no force-push, and the message marks the correction as mine.

Merging once the checks land. If a Windows leg goes red it is very likely not yours — this repo carries several intermittent Windows-only failures on bystander tests, and I re-run the identical commit to prove flakiness rather than asking you to chase it.

auxmedrano:
Verified `70d13b621` — mechanically confirmed it's comments-only (`git diff` shows every changed line matches `^[+-]\s*//`, nothing else), and `go vet` + `go test -count=1 ./internal/engine/... ./internal/quality/...` still exit clean after pulling it in.

Agreed on the logic itself, independently traced: `foldable` is only ever set for a `const` binding, and valid JS can't reassign a `const`, so `t.assigned[name]` can only fire on a binding that was already going to be declined by `!best.foldable` — it's reachable only in its over-decline form (a same-named `let` in an unrelated function suppressing a real module-level `const`). Keeping it as a fail-safe is the right call given the asymmetry you're optimizing for (an honest `/{dynamic}` vs. a fabricated endpoint), and flagging the equivalent-mutant result instead of quietly dropping it is exactly the standard this thread has been holding.

Nothing further from me — good to merge whenever checks land.

cajasmota:
Merged as `65180662d`. Thank you — this one improved twice over, and both times because of something you did.

**Your diagnosis was the load-bearing part.** You didn't just report that Angular URL constants weren't folding; you found four distinct shapes that broke it — a commented-out declaration, an out-of-scope declaration, a reassigned `let`, and a parameter shadowing a module constant. That set is what made the rework testable, and it's why the fix ended up reading a scoped symbol table off the AST instead of patching a regex. A bug report that hands you the failure taxonomy is worth several that hand you a patch.

**And you were clearer than I was on the mechanism.** When I described #6554 as folding the prefix into the endpoint path, you corrected me — it ships only the join key. You were right and I'd read the title instead of the diff. I also told you this PR fixed the `private url = '/api/things'` + `http.get(this.url)` shape; it does not, that shape is unchanged. Both corrections are on the record above.

Before merging I ran one more mutant of my own — dropping the scope-visibility window so a sibling-scope binding becomes visible file-wide. `go vet` 0, full unfiltered `go test -count=1 ./internal/engine/` EXIT=1, killed by `TestSynth_AngularConstFold_OutOfScopeDeclaration_6552` by name. Production file restored by `cp` with its shasum re-verified. So your shape-D case is genuinely pinned, not just passing.

The one loose end is documented rather than papered over: the file-global assignment tracker is redundant with the const-only rule, and dropping it leaves the package green. It stays as a fail-safe, with a comment now recording that it's an equivalent mutant and why it shouldn't be deleted on a green run alone — so whoever runs that mutant next finds the reasoning instead of a puzzle.

`angular-http-mini` now grades 23/23 entities and 18/18 relationships with two forbidden rows pinning your shapes A and H, so the four defects can't come back quietly.

One process note that cost you time and was entirely ours: your fork's CI had been parked in `action_required` awaiting maintainer approval since your first commit — on every new head sha, including the commits I pushed to your branch. That's why nothing appeared to run for so long. I've written the check into our merge routine (count the test legs rather than look for failures), so the next contributor PR won't sit like that.

Please do keep filing these.

Issue #5688, opened 2026-07-09
HTTP client→server joins unresolved for versioned/base-prefixed URLs (88%/79% endpoint orphan)


cajasmota:
Reported in #5680/#5681. http_endpoint_definition 88% / http_endpoint_call 79% orphan. Stack: Spring MVC (@RestController x160) + JAX-RS (@Path x133) defined; RestTemplate/WebClient x188 + Angular HttpClient x481 consumed. Paths versioned + base-URL-prefixed (`/api/v1/<r>/{id}`) built from constants.

**Assessment (per grounding):** the HTTP matcher (`internal/links/http_pass.go`) is already mature; residual orphans are mostly (a) client repos not in the same group, (b) dynamic/prefixed URL construction. Not a pass rewrite.

**Fix directions:** index the client repos into the same group; add a base-URL/prefix template via `grafel-resolve` (dynamic_baseurl residual surface exists); optionally feed OpenAPI specs (x60) through the openapi pass. Lower priority than B6/C1.

**Effort:** S–M (config/tuning).

cajasmota:
**Corpus validation (post-merge #5735 + #5738).** Rebuilt the aliased corpus group (a PHP backend + PHP framework + JS frontend monorepo) on the merged batch and compared against the pre-fix graph:

| metric | before | after |
|---|---|---|
| cross-repo edges | 0 | 27 |
| endpoints | 5 | 11 |

The version/prefix-aware HTTP joins from #5735 took the group from **0 → 27 cross-repo edges** — client calls in the frontend/framework now resolve to backend endpoints instead of orphaning. Combined with #5738 (PHP Slim endpoint synthesis) the endpoint count rose 5 → 11 as the previously-invisible PHP backend routes became real `http_endpoint_definition`s for the joins to land on. Remaining structural item (non-Django `url_mount_point` / cross-module base-URL) is deferred to v0.1.9 per plan. Refs #5729 #5733.

cajasmota:
Fixed by #5735: version/prefix-aware HTTP joins with prefix-stripped aliases, longest-first mount-prefix matching, and a base-URL dynamic-suffix matcher (internal/links/http_pass.go). Reopen with numbers if orphan rates have not moved on a re-measure.

Issue #6530, opened 2026-08-21
pattern_type is not unique across producers: Django and FastAPI both stamp url_mount_point, so it reads like an identifier and is not


cajasmota:
Found by the independent review of PR #6528, and proven with an executed probe rather than by reading. **The instance is fixed in that PR; the class is not, and it will recur.**

## The finding

`pattern_type: "url_mount_point"` is stamped by **two unrelated producers**:

- `internal/engine/django_urlconf_nested.go` — Django's nested urlconf composition
- `internal/engine/http_endpoint_synthesis.go:3169` — `fastapiMountPointSynthetics` (from #6385)

Both stamp it on kinds that overlap. So `pattern_type` alone does **not** identify which pass produced an entity.

That went unnoticed while nothing consumed it as an ownership signal. #6528 introduced the first pass that does — a recompute-and-reconcile that **deletes** graph entities it believes it owns — and immediately pruned live FastAPI mount synthetics:

```
--- FAIL: TestREV6528_PrunesForeignFastAPIMountPoint
  pruned a FOREIGN FastAPI url_mount_point synthetic http:ANY:/api/v2:mount@svc/main.py
```

Reachable in any repo with a FastAPI `include_router(prefix=)` mount plus any `*urls.py`. Silent, and permanent until the next full reindex.

## Why the #6528 fix does not close this

#6528 adds `framework == "django"` to its predicate. That is correct and sufficient **for that pass**. It does not change the fact that `pattern_type` is a non-unique key that reads like a unique one.

The next pass that keys on `pattern_type` — for ownership, for pruning, for attribution — hits the same trap, and there is nothing that would stop it or even flag it. The property is invisible: nothing asserts that a `pattern_type` maps to one producer, and nothing warns when two producers pick the same string.

This is the same shape as **#6514** (two byte-identical `include_router` rules in different YAML files, so neither can be tested in isolation): a value that looks like an identifier and is not, discovered only when something finally relies on it.

## What a fix could look like

Options, in rough order of cost:

1. **A test asserting `pattern_type` values are unique across producers** — enumerate the stamping sites and fail when two produce the same string without a distinguishing `framework`. Cheap, and it fires on the next collision rather than after one deletes data.
2. **Require `framework` wherever `pattern_type` is stamped**, making the pair the identifier. Narrow change, but it needs every producer audited.
[...]

PR #6554, opened 2026-08-23
fix(#6414): the mount synthetic and the route endpoint had no join key, so nothing could tell which routes a prefix belongs to

arthurgeron:
This is the first slice you recommended in the grounding comment on #6414, and deliberately not the fold. Your comment names that slice as stamping `mounted_module` and `mounted_router` on the mount synthetic in `fastapiMountPointSynthetics`, resolved from the mount file's own imports, and holds the fold back in your words "until Q2 (does a Pass-2.6 caller have access to unchanged-file bytes?) is answered and Q3 (one router mounted at two prefixes) is decided". Neither question is touched here. No `path` moves, so `_NoRouteRegression_6385` and `TestSynth_FastAPI_ImportedRouterByName` both stand unmodified.

## The defect, from the code rather than from a count

Line numbers below are on `main` at `8595d7a6b`. `fastapiMountPointSynthetics` (`http_endpoint_synthesis.go:3118`) emits the mount record with `url_prefix` and four other properties. `synthesizeFastAPI` emits the route endpoints through the record builder at `:512`. Neither carries anything naming the other: the mount states a prefix, the endpoint states a path, and no property relates them. This is unconditional, not distribution dependent.

**MEASURED.** Searched for `mounted_module`, `mounted_router`, `router_recv` and `mount_prefix_unapplied` with `rg` across the whole tree at `8595d7a6b`: zero matches, 6,800 files searched. Slice 1 was unbuilt.

Why the missing key matters more than a short path, as a constructed example rather than a report from any particular repository:

```python
# internal_api/main.py
from fastapi import FastAPI
from internal_api.routes import items
app = FastAPI()
app.include_router(items.router, prefix="/internal")

# external_api/main.py
from fastapi import FastAPI
from external_api.routes import items
app = FastAPI()
app.include_router(items.router, prefix="/public")
```

If both `items.py` declare `@router.get("/things")`, the two emitted `http_endpoint_definition` entities have the same verb and the same `path`, and no property separates them. Two mount synthetics exist with the right prefixes, and before this change nothing said which mount went with which endpoint. That is undecidable for a consumer, not merely truncated. The key does not fix it, but the fold cannot without it.

## What the change does

`mounted_module` holds the dotted Python module that defines the mounted router. `mounted_router` holds its name in that module. Both are resolved from the mount file's own imports, so the pass stays per file and a full index, `--incremental` and the daemon produce the same record. Both spellings that occur in real code resolve: `from app.api import markets` with `markets.router`, and `from app.api.markets import router` with a bare `router`.

Three decisions came out of measurement rather than reading.

**1. A dotted module, not a file path.** Your comment asked for `app/api/markets.py`. I built that first and measured it: of 54 resolvable mounts across the corpus below, **49 named a file that does not exist**, all 49 in one `src/`-layout repo where `dispatch.incident.views` sits at `src/dispatch/incident/views.py`. `ApplyDjangoNestedURLConf` converts the same way, but it can test the result and fall back to `modulePathToFilePath_relToParent` (`django_urlconf_nested.go:449`) against a file reader. A per-file pass has no reader, so publishing a converted path is a layout guess that is usually wrong. This publishes what the source states and leaves the file mapping to the consumer that holds the file set. With the dotted form, 54 of 54 name a module that maps to a real `.py` file in its repo under a dotted-suffix match.

**2. `parseImports` cannot resolve the dotted form.** It binds the alias and discards the original name (`django_drf_actions.go:1843`), so `from app.api import markets as mkts` with `mkts.router` resolves to `app.api.mkts`. `pythonFromImportBindings` reuses `flattenParenthesised` and `drfImportFromRe`, so parenthesised multi-line imports behave identically, and keeps both names. Two subtests pin the aliased module and the aliased router.

**3. The first argument is read with a new `fastapiFirstCallArg`, not `fastapiTopLevelArgs`.** The existing helper elides nested groups, which is right for finding the top-level `prefix=` and wrong here: under elision `markets.build_router(prefix="/internal")` arrives as `markets.build_router` and would be published as a router that does not exist. Mutation-verified below.

## Absent, never guessed

Each of these leaves both properties off the record rather than approximating:

- A **non-literal prefix**, including `prefix=f"{settings.API_PREFIX}"`. Unchanged and deliberate: `fastapiRouterPrefixKwargRe` only matches a literal, and following a module-level constant across files is a capability nothing here has. I filed #6385 partly on that shape and it stays excluded. Worth saying that the mount record's own `url_prefix` is equally absent for it, so a route under a non-literal prefix is still silently short with nothing marking it. Recording that observation is the `mount_prefix_unapplied` item from #6385's thread, which needs cross-file knowledge and is not in this slice.
- A **relative import** (`from .api import markets`). It is anchored on the mount file's package, so publishing `.api.markets` as absolute would name a module that is not there. The dot-depth walk-up is separate work, as your comment says.
- A plain `import a.b.c`, which `drfImportFromRe` does not match.
- A router **defined in the mount file itself**. There is no import to read, and the case is already whole: a same-file `APIRouter(prefix=...)` is folded by #5688, so no key is needed.
- An argument that is a **call**, or a local assigned from one. There is no router object to name. This is the largest unresolvable class on the corpus.
- **One prefix mounted by two different routers in one file.** The mount ID keys on the prefix alone (`:3150`), so these collapse into one record; naming whichever mount was scanned first asserts a join that is only half true. Two mounts of the same router are not ambiguous and keep the key. This is the ambiguity your comment flagged. It is now detectable rather than fixed: the ID still collapses, and changing it is entity-identity churn that belongs to your call, not mine.

## Evidence

**Real-tree measurement, per AGENTS.md.** Built the binary from this branch and indexed six FastAPI repositories with `HOME`, `GRAFEL_HOME` and `GRAFEL_DAEMON_ROOT` all isolated: `langflow-ai/langflow`, `fastapi-users/fastapi-users`, `tiangolo/full-stack-fastapi-template`, `mlflow/mlflow`, `Netflix/dispatch`, `nsidnev/fastapi-realworld-example-app`. 7,442 Python files. Counted `url_mount_point` entities with `framework=fastapi` in the exported `graph.json`.

| | count |
|---|---|
| mount synthetics with a literal prefix | 70 |
| carrying the join key | 54 |
| of those, naming a module that maps to a real file | 54 |
| no key, argument is a call or a local assigned from one | 15 |
| no key, two different routers at one prefix | 1 |

The last row is the collapse rule firing on real code (`fastapi-realworld-example-app/app/api/routes/articles/api.py:7-8` mounts two different routers at `/articles`), not only in a fixture. The 15 are `fastapi_users.get_auth_router(auth_backend)` and siblings, plus one langflow local assigned from a factory call.

**Mutation-verified.** Each of these turns the new suite red: swapping `fastapiFirstCallArg` for `fastapiTopLevelArgs`; dropping the trailing `\s*$` from `fastapiMountedRouterArgRe`, which then publishes `mounted_router=build_router` for the call-expression case; and taking first-wins instead of dropping the ambiguous key, which then publishes one of the two `/network` routers.

**Gates.** `go test ./...` green on 233 packages. `go vet ./...`, `go vet -tags perf ./...`, `gofmt -l $(git ls-files '*.go')` and `go run ./cmd/lint-localematch` clean. `tools/coverage validate`, `backfill --check` and `fmt --check` exit 0, and `gen` leaves `docs/coverage` byte-identical. `scripts/quality/run.sh --ratchet` holds, 26 gated fixtures at baseline.

Two failures are not from this change, stated rather than left to be found:

- `TestBaselineMeasuredOnIsReachable` fails identically on unmodified `origin/main`: `baseline.json` records `measured_on: "0ca8d585e"`, and `git cat-file -t 0ca8d585e` reports it is not a valid object in the public repository. Verified on a clean `origin/main` worktree.
- `TestSelfDefenseCheck_RunRefusesWhenCalledFromTmpBinary` failed only while the measurement binary above was sitting in `/tmp`, and passes once it is removed. My own doing, not a defect.

**No coverage registry change, and the reasoning so it is cheap to overrule.** No `path` and no capability status moves; the change adds two properties to an existing synthetic. The direct precedent in this subsystem, `ee89d4a83` for #6385, also shipped without a registry change. There is no mount-composition capability key in the taxonomy, and AGENTS.md says a missing slot is proposed in a separate schema PR rather than invented in the implementation PR. Say the word and I will add a `Routing` cell in a follow-up.

**On the golden set.** `python-fastapi-mini` now exists, so #6414's "no FastAPI fixture at all" is out of date. It cannot grade this change either way: its `app.include_router(router)` has no prefix, and **MEASURED** by grepping every `expected.json` under `internal/quality/golden/`, the schema has no `properties` key anywhere, so it cannot express a property assertion at all. The Go-level pins are the only instrument that fits. The ratchet confirms all 26 fixtures are unmoved.

## Tests

`internal/engine/http_endpoint_synthesis_fastapi_mounted_router_6414_test.go`, driven through the existing `runDetect` and reusing `fastapiMountSynths` from the #6385 file. Seven positive cases, seven absence cases, and one that re-pins the paths as unfolded at the point where that now matters most. Each positive case fails against `8595d7a6b`:

```
--- FAIL: TestSynth_FastAPI_MountedRouter_DottedModuleAttr_6414 (0.03s)
    http_endpoint_synthesis_fastapi_mounted_router_6414_test.go:62: #6414: mounted_module = "" (present=false), want "app.api.markets"
    http_endpoint_synthesis_fastapi_mounted_router_6414_test.go:62: #6414: mounted_router = "" (present=false), want "router"
```

Reproduce it by checking out this branch's test file over `8595d7a6b`'s `http_endpoint_synthesis.go`: all seven positive assertions report `present=false`, and the seven absence cases pass trivially, which is why the absence rules are mutation-verified above rather than left to the table alone.

Refs #6414

### DIFF
--- a/internal/engine/http_endpoint_synthesis.go
+++ b/internal/engine/http_endpoint_synthesis.go
@@ -2956,6 +2956,166 @@ var fastapiFileEvidenceRe = regexp.MustCompile(
 // is acceptable; a garbage one is not.
 var fastapiBadPrefixCharRe = regexp.MustCompile(`[\s"'\\]`)
 
+// fastapiMountedRouterArgRe matches the first positional argument of an
+// `include_router` call when, and only when, it is a bare name or a
+// single-level attribute access on one: `router` or `markets.router`. It is
+// fully anchored, so a call expression, a subscript, an attribute chain deeper
+// than one level, or anything else yields no join key at all rather than a
+// half-parsed one (#6414).
+var fastapiMountedRouterArgRe = regexp.MustCompile(`^\s*(?:([A-Za-z_]\w*)\.)?([A-Za-z_]\w*)\s*$`)
+
+// pyImportBinding is one name bound by an absolute `from <module> import
+// <name> [as <bound>]`: the module it came from and its ORIGINAL name there.
+type pyImportBinding struct {
+	module string
+	name   string
+}
+
+// pythonFromImportBindings maps bound-name -> pyImportBinding for every
+// `from <module> import <names>` in src, reusing flattenParenthesised and
+// drfImportFromRe so multi-line parenthesised imports behave exactly as they
+// do for parseImports.
+//
+// It differs from parseImports (#6414) in KEEPING the original name, which
+// parseImports discards for an aliased import. The original is what a dotted
+// mount argument needs: `from app.api import markets as mkts` + `mkts.router`
+// names the module `app.api.markets`, while resolving the ALIAS would name
+// `app.api.mkts`, which does not exist. A join key naming a module that is not
+// there is worse than no key, so the alias is resolved through to the name the
+// defining module actually uses.
+//
+// Later bindings win, as they do in Python and in parseImports.
+func pythonFromImportBindings(src string) map[string]pyImportBinding {
+	out := map[string]pyImportBinding{}
+	for _, m := range drfImportFromRe.FindAllStringSubmatch(flattenParenthesised(src), -1) {
+		module := strings.TrimSpace(m[1])
+		names := strings.TrimSpace(m[2])
+		if i := strings.Index(names, "#"); i >= 0 {
+			names = names[:i]
+		}
+		names = strings.Trim(names, "() \t")
+		for _, raw := range strings.Split(names, ",") {
+			raw = strings.TrimSpace(raw)
+			if raw == "" {
+				continue
+			}
+			name, bound := raw, raw
+			if i := strings.Index(raw, " as "); i >= 0 {
+				name = strings.TrimSpace(raw[:i])
+				bound = strings.TrimSpace(raw[i+4:])
+			}
+			if name == "" || bound == "" {
+				continue
+			}
+			out[bound] = pyImportBinding{module: module, name: name}
+		}
+	}
+	return out
+}
+
+// fastapiFirstCallArg returns the source text of the first argument of the
+// bracket-balanced call whose opening `(` sits at index open, stopping at the
+// first TOP-LEVEL comma or at the call's own closing bracket. Nested bracket
+// groups are copied through verbatim rather than elided the way
+// fastapiTopLevelArgs elides them, so a call-expression argument comes back
+// whole and is rejected by fastapiMountedRouterArgRe instead of looking like
+// the bare name that precedes its paren. Returns "" for an unterminated call.
+func fastapiFirstCallArg(src string, open int) string {
+	depth := 0
+	start := open + 1
+	for i := open; i < len(src); i++ {
+		switch c := src[i]; c {
+		case '(', '[', '{':
+			depth++
+		case ')', ']', '}':
+			depth--
+			if depth == 0 {
+				return src[start:i]
+			}
+		case ',':
+			if depth == 1 {
+				return src[start:i]
+			}
+		case '"', '\'':
+			j := i + 1
+			for j < len(src) {
+				if src[j] == '\\' {
+					j += 2
+					continue
+				}
+				if src[j] == c || src[j] == '\n' {
+					break
+				}
+				j++
+			}
+			if j < len(src) && src[j] == c {
+				j++
+			}
+			if j > len(src) {
+				j = len(src)
+			}
+			i = j - 1
+		}
+	}
+	return ""
+}
+
+// fastapiResolveMountedRouter resolves an `include_router` first argument to
+// the DOTTED module path that defines the router and the router's name in that
+// module, using ONLY the mount file's own imports (#6414). Both results are ""
+// together whenever the mount cannot be resolved that way, which keeps the
+// properties absent rather than guessed.
+//
+// WHY A DOTTED MODULE AND NOT A FILE PATH. Converting it here with
+// modulePathToFilePath would be a guess about the repo's layout, and measured
+// on six real FastAPI repositories the guess is wrong far more often than it is
+// right: of 54 resolvable mounts, 49 named a file that does not exist, all 49
+// in one `src/`-layout repo where `dispatch.incident.views` lives at
+// `src/dispatch/incident/views.py`. Django converts the same way but only
+// because it can TEST the result — ApplyDjangoNestedURLConf tries
+// modulePathToFilePath and falls back to modulePathToFilePath_relToParent
+// against a file reader (django_urlconf_nested.go:400). A per-file pass has no
+// reader, so it publishes the value the source actually states and leaves the
+// file mapping to the consumer that holds the file set.
+//
+// Deliberately not resolved here, each because it needs input this pass does
+// not have:
+//   - a RELATIVE import (`from .api import markets`), which is anchored on the
+//     mount file's own package and needs a dot-depth walk-up to become
+//     absolute.
+//   - a plain `import app.api.markets`, which drfImportFromRe does not match.
+//   - a router defined in the mount file itself, which has no import to read.
+//     That case is already whole: a same-file `APIRouter(prefix=...)` is
+//     folded into the route path by #5688, so no join key is needed for it.
+//   - an argument that is a CALL rather than a name
+//     (`fastapi_users.get_auth_router(backend)`) or a local variable assigned
+//     from one. There is no router object to name. This is the single largest
+//     unresolvable class on the measured corpus, 15 of 16 absences.
+func fastapiResolveMountedRouter(arg string, imports map[string]pyImportBinding) (module, router string) {
+	m := fastapiMountedRouterArgRe.FindStringSubmatch(arg)
+	if m == nil {
+		return "", ""
+	}
+	qualifier, attr := m[1], m[2]
+	// `markets.router`: the qualifier is the imported module and the attribute
+	// is the router. `router`: the imported name IS the router.
+	lookup := qualifier
+	if lookup == "" {
+		lookup = attr
+	}
+	b, ok := imports[lookup]
+	if !ok || strings.HasPrefix(b.module, ".") {
+		return "", ""
+	}
+	dotted, router := b.module, attr
+	if qualifier == "" {
+		router = b.name
+	} else {
+		dotted = b.module + "." + b.name
+	}
+	return dotted, router
+}
+
 // pythonMaskInertRegions returns a copy of src of the SAME byte length, with
 // the same newline positions, in which the contents of `#` comments and of
 // triple-quoted string literals (docstrings, embedded samples) are replaced by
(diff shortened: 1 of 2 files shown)


cajasmota:
Thank you for these — four PRs, each with the issue filed alongside it and a test beside the fix. All four are green on every leg.

What stands out is that the **titles name the mechanism rather than the symptom**. *"the `*.csproj` manifest entry was glob text passed to `os.Stat`"*, *"the decorator path capture rejected an empty literal, so a prefixed router's own path was dropped"*, *"a pydantic field's default was read for optionality and then discarded"* — each of those is the sentence a maintainer needs and usually has to reconstruct. It makes them reviewable in one pass.

Two are worth calling out specifically:

**#6558** — glob text handed to `os.Stat` is the kind of defect that cannot fail loudly. `os.Stat("*.csproj")` simply returns not-exist, so a .NET project directory was silently never treated as a backend boundary, and nothing anywhere reported a problem. Silent-wrongness defects are ranked above loud breakage here precisely because a user has no way to notice them.

**#6554** — #6414 has been open a while and blocked on exactly the thing your title names: there was no join key between the mount synthetic and the route endpoint, so no amount of care at either site could tell which routes a prefix belonged to. Establishing the key is the part that was missing.

Each will get an independent adversarial review before merge. That is our standard path for every change including our own — the maintainer's own PRs in flight right now are on their third and fifth review rounds — so please read it as process rather than as doubt about the work. I will post each result on its PR, and I will run a killing mutant myself before merging.

If a review turns up something, it will come with the reproduction attached so you can judge it rather than take it on trust.

cajasmota:
Independent review is back: **approve on the behaviour, with two test cases worth adding before merge.** No defect found in the code — both findings are places where the suite would not notice if the code were wrong.

First, a correction on our side: my reviewer was briefed that this PR "folds the prefix into the endpoint path." **It does not, and you were clearer about that than I was.** It ships the join key — `mounted_module` + `mounted_router` on the `url_mount_point` synthetic — and changes no `path`, `ID`, `Name`, `Kind` or `SourceFile`. `TestSynth_FastAPI_MountedRouter_PathIsStillUnfolded_6414` pins exactly that with a `requireNotContains` on `http:GET:/network/items`. Shipping the key without the fold is the right decomposition; the mis-briefing was mine.

## What held up under attack

- **Alias resolution is correct**, which I did not expect to survive: `from app.api import markets as mkts` resolves to the *defining* module `app.api.markets`, not `app.api.mkts`.
- **Two apps with same-named routers stay distinguishable** — `internal_api.routes.items` vs `external_api.routes.items`.
- **Cross-mount contamination is not reachable.** `mountID` is built over the *full* canonical prefix with no suffix or segment matching anywhere in the new code. A probe with `/api/users` and `/admin/users` sharing a last segment resolved each to its own module — and kept passing even under both mutants below. That matters because that exact shape survived three review rounds on #6461/#6528 in this same subsystem.
- **No golden can move.** `expected.json` gates on `name`/`kind`/`source_file`/relationships and this PR adds only properties.
- **Two different routers at one prefix** correctly drops the key from the collapsed record.

## Two test gaps — the code is right, the suite would not catch it going wrong

Both were found by mutating the shipped code; in both cases the **real code rejects the shape correctly** and the mutant still passed the whole package.

**1. Mixed resolvability at one prefix.** Guarding the ambiguity-drop on resolvability (`mountedModule != "" && …` at `http_endpoint_synthesis.go:3327`) leaves `internal/engine` green. With one resolvable mount and one call-expression mount at the same prefix, the collapsed record would keep `mounted_module=app.api.markets` while also covering `fastapi_users.get_auth_router(backend)` — the "join that is only half true" your own comment says the drop prevents. The existing `two-routers-one-prefix` case uses two *resolvable* routers, so it cannot see this. Your PR's own corpus row reporting **15 call-expression mounts in fastapi-users** is what makes this the higher-value of the two.

**2. The `^` anchor in `fastapiMountedRouterArgRe` is load-bearing.** Dropping it leaves the package green, and it is not equivalent: `container.markets.router` with `from app.api import markets` in scope then suffix-matches and publishes `app.api.markets`/`router` for an unrelated object. The `plain-import` fixture misses it because `markets` is not from-import-bound there, so it returns absent either way.

**The ask:** add those two shapes to the absence table — mixed-resolvability-same-prefix, and a deep attribute chain. Happy to add them myself if you would rather not context-switch; say the word and I will push them to your branch.

## One thing for the issue, not for you

Both producers build the mount ID with the byte-identical formula — `django_urlconf_nested.go:174` and `http_endpoint_synthesis.go:3325` both emit `httproutes.SyntheticID("ANY", canonical) + ":mount"`. So a repo with a Django `path("api/", include(...))` and a FastAPI `include_router(prefix="/api")` mints the same entity ID from both.

That is **pre-existing (#6530), not introduced here** — but this PR makes it consequential for the first time, because whichever record survives dedup now decides whether the join key is present, absent, or attached to a Django mount. I am recording it on #6530 rather than expanding this PR.

## Still open, and correctly so

The key is **many-to-many**: the same router mounted at two prefixes produces an identical key on both records. That is #6414's deferred Q3, the fold cannot proceed on this key alone, and nothing here pretends otherwise — flagging it only so we agree it stays open rather than reading as solved.

arthurgeron:
Both shapes are pushed as `48ce7b0c3`, so no need for you to context-switch.

`mixed-resolvability-same-prefix` and `deep-attribute-chain` are rows in `TestSynth_FastAPI_MountedRouter_AbsentWhenUnresolvable_6414`. I ran both of your mutants to confirm the rows actually kill them:

- Guarding the drop as `mountedModule != "" && (...)` at `http_endpoint_synthesis.go:3327` fails `mixed-resolvability-same-prefix` with `mounted_module present as "app.api.markets", want absent`.
- Removing the `^` from `fastapiMountedRouterArgRe` fails `deep-attribute-chain` with the same message.

Each mutant is killed only by the row aimed at it, and `go test ./internal/engine/` is green once both are reverted.

One detail worth recording rather than hiding: I also added the reversed order, `mixed-resolvability-same-prefix-reversed`, and it does **not** fail under the resolvability mutant. When the call-expression mount is scanned first the surviving record has no key to keep, so only the resolvable-first order can carry a stale key into it. That row is there to pin the drop as order-independent, not because it adds mutation coverage, and its comment says so. Drop it if you would rather the absence table only carried rows that kill something.

Agreed on the last two points. The mount-ID collision belongs on #6530, not here, and the same-router-at-two-prefixes case keeps the key many-to-many, so #6414's Q3 stays open and the fold cannot proceed on this key alone.

cajasmota:
Both rows verified on my side, independently of your run. I re-scored each mutant against `48ce7b0c3` — `go vet` clean first, `go test -count=1 ./internal/engine/` as a full package run, file restored by `cp` with shasum `05b191e6…` re-verified after each:

```
M1  guard the drop on resolvability
    --- FAIL: …_AbsentWhenUnresolvable_6414/mixed-resolvability-same-prefix

M2  drop the ^ from fastapiMountedRouterArgRe
    --- FAIL: …_AbsentWhenUnresolvable_6414/deep-attribute-chain
```

Each dies to exactly the row aimed at it and nothing else. That is the property I care about more than the kill itself — a row that takes down half the table tells you a mutant is bad, not that the coverage is precise.

## Keep the reversed row

**Keep it, and thank you for flagging it rather than letting it read as coverage.** That disclosure is worth more than the row.

Your reasoning is right: when the call-expression mount is scanned first the surviving record has no key to keep, so only the resolvable-first order can carry a stale key into the collapse. So it adds no mutation coverage — but it does pin **order-independence**, which is a genuine property of the drop and one a future refactor could break without touching either regex.

The thing this project actually gets bitten by is a row that looks like coverage and is not. A row whose comment states plainly what it does and does not pin is the opposite of that problem. We have had two fixtures this month that could not observe the thing they were written for — one where every registering file was always in the changed set, another where the marker sat below the level the mutant operated on — and both times the fixture read as proof while proving nothing. Yours says what it is.

## Merging once the Windows leg lands

Everything else is green; the Windows leg is still running. For context so a red one does not alarm you: this repo currently has **four intermittent Windows-only failures** (#6512, #6493, #6421, plus a `cmd/grafel` timing test), all bystanders unrelated to any change under review. If one trips on your PR I will re-run the identical commit to prove it flaky rather than ask you to chase it.

Both remaining points stand as you put them: the mount-ID collision is recorded on #6530 with the byte-identical-formula detail added, and #6414's Q3 stays open — the key is many-to-many and the fold cannot proceed on it alone.

Issue #6555, opened 2026-08-23
deriveOwningBackend never tests the repository root and stops at intra-service entry-point files, so every endpoint in a single-service repo is tagged app or src


arthurgeron:
I consume grafel's `http_endpoint_definition` records and their `owning_backend` property. In a repository holding several services the property names the service. In a repository holding one it names a source-root directory: `app`, `src`, or whatever the first path segment happens to be.

`deriveOwningBackend` (`internal/engine/http_endpoint_synthesis.go:6381`) walks the handler file's directory upward and returns `filepath.Base(dir)` for the first directory holding a manifest or a framework marker, falling back to the top-level path segment.

MEASURED by calling it against temp trees (Go test in `internal/engine`, one `os.Chdir` per case, all trees constructed):

| tree | handler file | result |
|---|---|---|
| `pyproject.toml` + `app/main.py` | `app/api/endpoints/notifications.py` | `app` |
| `pyproject.toml` only | `src/service/api/routes.py` | `src` |
| `svc-a/pyproject.toml`, `svc-b/pyproject.toml` | `svc-a/app/routers/x.py` | `svc-a` |

The third row is the behaviour the property is for. The first two are the degradation, and two independent mechanisms cause them.

**1. The repository root is never tested.** The loop breaks when `dir` reaches `"."`, before `directoryHasManifest` is called on it, so a manifest at the repository root is never a candidate. Row 2 is that path: `pyproject.toml` sits at the root, the walk exhausts `src/service/api`, `src/service` and `src`, breaks at `.`, and the fallback returns the top-level segment. Testing `.` would not fix it alone, because `filepath.Base(".")` is `.`.

**2. `frameworkMarkerFiles` holds intra-service entry points.** `main.py`, `app.py`, `asgi.py`, `wsgi.py`, `index.js`, `server.js` and `app.js` are modules inside a service, not boundaries between services. Row 1 is that path: in the layout FastAPI's own docs use (root `pyproject.toml`, package `app/` with `app/main.py`), the walk reaches `app/main.py` first because it is deeper, returns `app`, and the root manifest is never consulted. A framework marker below therefore beats a real manifest above, whenever both exist.

The property's own doc comment (`internal/types/kinds.go:166`) describes the derivation as walking up "until a framework manifest (pyproject.toml, package.json, go.mod) or framework marker is found". Row 1 does not satisfy that reading: `pyproject.toml` is present and is not found.

## Consequence inside grafel

[...]

PR #6558, opened 2026-08-23
fix(#6556): the *.csproj manifest entry was glob text passed to os.Stat, so a .NET project directory was never a backend boundary

### LINKED ISSUE #6556: the *.csproj manifest entry is glob text passed to os.Stat, so a .NET service directory is never recognised as a backend boundary
Context: I consume `http_endpoint_definition` records and their `owning_backend` property.

`manifestFileNames` (`internal/engine/http_endpoint_synthesis.go:6348`) is fourteen entries, thirteen literal filenames and one glob:

```go
"*.csproj",         // C#
```

`directoryHasManifest` (`:6409`) tests each entry with `os.Stat(filepath.Join(dir, name))`. `os.Stat` does no glob expansion, so that entry can only match a file literally named `*.csproj`, and the C# arm of the list never fires.

MEASURED by calling `deriveOwningBackend` against temp trees (Go test in `internal/engine`, one `os.Chdir` per case, all trees constructed):

| tree | handler file | `owning_backend` |
|---|---|---|
| `services/orders/go.mod` | `services/orders/handlers/order.go` | `orders` |
| `services/orders-api/package.json` | `services/orders-api/src/routes/order.ts` | `orders-api` |
| `services/Orders.Api/Orders.Api.csproj` | `services/Orders.Api/Controllers/OrderController.cs` | `services` |

Rows 1 and 2 find their manifest one directory above the handler's directory. Row 3 has the same shape and walks straight past `services/Orders.Api`, ending on the top-level-segment fallback.

Scope: `manifestFileNames` and `frameworkMarkerFiles` hold no other glob, so `*.csproj` is the only affected entry. `.fsproj`, `.vbproj` and `.sln` are absent from the list entirely, which is a separate gap and not this one.

PR follows.

Refs #6555

arthurgeron:
Closes #6556

`manifestFileNames` holds thirteen literal filenames and one glob, `*.csproj`. `directoryHasManifest` tests every entry with `os.Stat`, which does no glob expansion, so that entry could only match a file literally named `*.csproj` and the C# arm never fired. A .NET project directory was therefore never a backend boundary, and `deriveOwningBackend` fell through to the top-level-segment fallback.

The fix routes an entry containing `*`, `?` or `[` through `filepath.Glob` and leaves every other entry on `os.Stat`.

## The measurement, and its unit

Unit of analysis: one source file, calling `deriveOwningBackend` directly after `chdir` into each checkout. Not one emitted endpoint entity, since these trees are not indexed here.

| tree | files measured | values changed | distinct values before | after |
|---|---|---|---|---|
| `dotnet/eShop` @ `ae71a06` | 544 `.cs` | 542 | 2 (`src` 495, `tests` 49) | 29 |
| `jasontaylordev/CleanArchitecture` @ `10f1a45` | 110 `.cs` | 108 | 3 (`src` 76, `tests` 32, `templates` 2) | 14 |
| `fastapi/full-stack-fastapi-template` @ `68adb40` | 86 `.py` and `.ts` | 0 | 6 | 6 |

The 29 buckets on eShop name real projects: `Ordering.API` 68, `Identity.API` 63, `Catalog.API` 39, `Webhooks.API` 27, `Basket.API` 11, and so on. Two `.cs` files stay on `src` because they sit outside any project directory, which is correct. The third row is the no-regression control: a tree with no `.csproj` anywhere is untouched.

## Mutant

`TestDeriveOwningBackend_ProjectFileManifest_6556` is a three-row table. Rows 1 and 2 are controls that fail if a fix widens the directory walk instead of fixing the match. Against `main` before the change:

```
--- FAIL: TestDeriveOwningBackend_ProjectFileManifest_6556 (0.00s)
    owning_backend_csproj_6556_test.go:67: deriveOwningBackend("services/Orders.Api/Controllers/OrderController.cs") = "services", want "Orders.Api"
    --- PASS: TestDeriveOwningBackend_ProjectFileManifest_6556/go.mod_one_level_above_the_handler (0.00s)
    --- PASS: TestDeriveOwningBackend_ProjectFileManifest_6556/package.json_one_level_above_the_handler (0.00s)
    --- FAIL: TestDeriveOwningBackend_ProjectFileManifest_6556/csproj_one_level_above_the_handler (0.00s)
```

All three rows pass after it.

## Limitations

`filepath.Glob` reads its whole argument as a pattern, so a directory name containing `*`, `?` or `[` is now interpreted where `os.Stat` treated it literally. This only reaches the glob branch for the `*.csproj` entry, whose joined path includes the directory being tested.

`.fsproj`, `.vbproj` and `.sln` are still absent from `manifestFileNames`. That is a separate gap, not this one, so this PR does not add them.

`owning_backend` degrading in a single-service repository (#6555) is a different mechanism and is not fixed here.

No `docs/coverage/registry.json` change: no capability cell cites `owning_backend`, checked by search.

## Verification

`gofmt -l internal/engine` and `go vet ./...` are both silent.

`go test -race -count=1 -timeout 40m ./...` on this branch: 233 packages ok, including `internal/engine`, which is the only package the diff touches. Two packages fail, neither of them for a reason the diff can reach:

`internal/quality`

```
--- FAIL: TestBaselineMeasuredOnIsReachable (0.07s)
    baseline_test.go:204: baseline.json: measured_on "0ca8d585e" is not a commit in this repository
```

The same command on `main` at `8595d7a6b` gives 234 ok and this as its only failure, so it is a property of the checkout.

`internal/daemon`

```
--- FAIL: TestEngineSupervisor_UnkeepableEngineIsFatal (15.01s)
    supervise_test.go:271: supervisor never surfaced a fatal for an unkeepable engine
```

Load dependent. It did not appear in the `main` baseline run, and on this branch it passes at `-count=3` in isolation and again across the whole `./internal/daemon/` package.

### DIFF
--- a/internal/engine/http_endpoint_synthesis.go
+++ b/internal/engine/http_endpoint_synthesis.go
@@ -6406,9 +6406,19 @@ func deriveOwningBackend(filePath string) string {
 // directoryHasManifest reports whether dir contains a manifest file or
 // framework marker. Uses os.Stat so it works with both real file trees
 // (during actual indexing) and in-memory test scenarios.
+//
+// #6556: an entry carrying a glob metacharacter is matched with filepath.Glob
+// instead. os.Stat does no glob expansion, so "*.csproj" could only ever match
+// a file of that literal name, and the C# entry never fired.
 func directoryHasManifest(dir string) bool {
 	allMarkers := append(manifestFileNames, frameworkMarkerFiles...)
 	for _, name := range allMarkers {
+		if strings.ContainsAny(name, "*?[") {
+			if m, err := filepath.Glob(filepath.Join(dir, name)); err == nil && len(m) > 0 {
+				return true
+			}
+			continue
+		}
 		if _, err := os.Stat(filepath.Join(dir, name)); err == nil {
 			return true
 		}
--- a/internal/engine/owning_backend_csproj_6556_test.go
+++ b/internal/engine/owning_backend_csproj_6556_test.go
@@ -13,9 +13,11 @@ import (
 // recognised as a backend boundary: the walk continued past it and
 // owning_backend came out as an ancestor directory.
 //
-// The three rows below are the same layout in three ecosystems. Go and Node
-// are the control: they find their manifest one directory above the handler,
-// and must keep doing so.
+// The first three rows below are the same layout in three ecosystems. Go and
+// Node are the control: they find their manifest one directory above the
+// handler, and must keep doing so. The fourth row pins the other half of the
+// property: the glob is anchored to the directory being tested, so a .csproj
+// in a cousin directory must not make an ancestor a boundary.
 
 // writeTree creates each named file (with its parent directories) under root.
 func writeTree(t *testing.T, root string, files ...string) {
@@ -55,6 +57,12 @@ func TestDeriveOwningBackend_ProjectFileManifest_6556(t *testing.T) {
 			tree:    []string{"services/Orders.Api/Orders.Api.csproj", "services/Orders.Api/Controllers/OrderController.cs"},
 			handler: "services/Orders.Api/Controllers/OrderController.cs",
 			want:    "Orders.Api",
+		},
+		{
+			name:    "csproj in a cousin directory is not matched",
+			tree:    []string{"repo/a/b/Orders.Api/Orders.Api.csproj", "repo/a/b/web/Controllers/OrderController.cs"},
+			handler: "repo/a/b/web/Controllers/OrderController.cs",
+			want:    "repo",
 		},
 	}
 
--- a/internal/engine/owning_backend_csproj_6556_test.go
+++ b/internal/engine/owning_backend_csproj_6556_test.go
@@ -1,0 +1,71 @@
+package engine
+
+import (
+	"os"
+	"path/filepath"
+	"testing"
+)
+
+// #6556: manifestFileNames carries one glob, "*.csproj", among thirteen
+// literal filenames, and directoryHasManifest tests every entry with os.Stat,
+// which does no glob expansion. So the C# entry could only ever match a file
+// literally named "*.csproj", and a .NET project directory was never
+// recognised as a backend boundary: the walk continued past it and
+// owning_backend came out as an ancestor directory.
+//
+// The three rows below are the same layout in three ecosystems. Go and Node
+// are the control: they find their manifest one directory above the handler,
+// and must keep doing so.
+
+// writeTree creates each named file (with its parent directories) under root.
+func writeTree(t *testing.T, root string, files ...string) {
+	t.Helper()
+	for _, f := range files {
+		p := filepath.Join(root, f)
+		if err := os.MkdirAll(filepath.Dir(p), 0o755); err != nil {
+			t.Fatalf("MkdirAll(%q): %v", filepath.Dir(p), err)
+		}
+		if err := os.WriteFile(p, nil, 0o644); err != nil {
+			t.Fatalf("WriteFile(%q): %v", p, err)
+		}
+	}
+}
+
+func TestDeriveOwningBackend_ProjectFileManifest_6556(t *testing.T) {
+	cases := []struct {
+		name    string
+		tree    []string
+		handler string
+		want    string
+	}{
+		{
+			name:    "go.mod one level above the handler",
+			tree:    []string{"services/orders/go.mod", "services/orders/handlers/order.go"},
+			handler: "services/orders/handlers/order.go",
+			want:    "orders",
+		},
+		{
+			name:    "package.json one level above the handler",
+			tree:    []string{"services/orders-api/package.json", "services/orders-api/src/routes/order.ts"},
+			handler: "services/orders-api/src/routes/order.ts",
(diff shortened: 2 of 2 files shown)


arthurgeron:
The `test` job is green on both `ubuntu-latest` and `windows-latest`, so both failures listed under Verification are artefacts of my local checkout and load, not of the diff.

cajasmota:
Independent review: **approve on the behaviour.** One coverage gap worth a row before merge, and one finding that came back negative in a way worth telling you about.

## The fix is correctly anchored

`filepath.Glob(dir + "/*.csproj")` does not cross a separator, and that was probed rather than assumed:

- `sub/dir/x.csproj` with the handler at `sub/other/y.cs` → falls back to `"sub"`, not matched
- grandchild-only `repo/a/b/Proj/Proj.csproj` with the handler at `repo/a/b/c/x.cs` → falls back to `"repo"`, no descent
- two `.csproj` in one directory → fine
- a directory literally *named* `*.csproj` → matches, but `os.Stat` matched it too, so not a regression

`directoryHasManifest` has exactly one caller (`deriveOwningBackend`, `http_endpoint_synthesis.go:6388`), so the blast radius is contained.

## I asked whether other manifest entries had the same bug. They do not.

This was the finding I expected to be bigger than the instance, and it is not — which is worth stating plainly rather than leaving implied.

`*.csproj` is the **only** entry carrying a metacharacter across all 13 `manifestFileNames` and all 9 `frameworkMarkerFiles`; every other entry is a literal filename. Sweeping the whole `internal/engine` package for the bug class turns up only three non-test `os.Stat` sites — `http_endpoint_synthesis.go:6422`, `orm_queries_python_mongo_agg.go:1371`, `kafka_edges.go:482` — and the latter two build fully literal paths. So this really was a single-entry defect.

## One row I would like, and I will add it myself if you prefer

A permissive mutant survives the whole package: adding a second, one-level-deeper glob alongside the anchored one (`filepath.Glob(filepath.Join(dir, "*", name))`). `go vet` clean, `./internal/engine/` **ok 35.6s**.

It is a real behaviour change — the grandchild-only tree above flips from `"repo"` to `"b"`. Your three-row table pins the *manifest-in-this-directory* case, so nothing currently watches the **"must not match below the directory"** half, which is the anchoring property the PR body advertises.

A fourth row — handler in a sibling subtree, `.csproj` in a cousin directory, expect the fallback — kills it. Say the word and I will push it rather than have you context-switch.

Good news on the other direction: `len(m) > 0` → `len(m) >= 0` (every directory becomes a boundary) **dies** on all three rows including both literal-manifest controls, so the controls do bite.

## Two nits, neither blocking

- `.CSPROJ` is not matched — MSBuild is case-insensitive, `filepath.Match` is not.
- A `.csproj` under `bin/` matches. Same class as a `package.json` under `node_modules`, and pre-existing walk policy rather than anything you introduced.

Your disclosed limitation (a directory name containing `[1]`) is real and also missed *before* the fix, so it is not a regression — and disclosing it up front is the right call.

## Golden movement: none

`csharp-aspnet-core-mini` scores 4/5 entities, 0/0 relationships, 0 forbidden on both your branch and the base tree — byte-identical. `python-fastapi-mini` unchanged. Neither fixture contains a `.csproj`, and no `expected.json` asserts on `owning_backend`.

The `./internal/quality/` failure you may see is **pre-existing and not yours**: `TestBaselineMeasuredOnIsReachable` fails because `baseline.json`'s `measured_on 0ca8d585e` is not an ancestor of `origin/main`. It is a checkout property and fails on main too — exactly as you reported. I am filing it separately so it stops looking like ambient red on everyone's PR.

cajasmota:
Pushed as `10ec05ed1`.

Added a fourth row to `TestDeriveOwningBackend_ProjectFileManifest_6556` — *"csproj in a cousin directory is not matched"*: handler at `repo/a/b/web/Controllers/OrderController.cs`, `.csproj` at `repo/a/b/Orders.Api/Orders.Api.csproj`, expecting the `"repo"` fallback. Header comment extended to describe the anchoring half your three original rows did not cover.

**Permissive mutant** (an extra `filepath.Glob(filepath.Join(dir, "*", name))` alongside the anchored one) — before: package green. After: **dies** with `deriveOwningBackend(...) = "b", want "repo"`.

---

**I pushed the row rather than asking you to — and I was wrong to ask in the first place.** That is the norm here going forward: a round-trip for something a maintainer can do in minutes costs you a context-switch and turns a merge into a wait.

Your commits are untouched: one commit added on top, no rebase, no squash, no amend, no force-push. The commit message says the addition was mine so you are not credited with work you did not do.

Scoring method, so you can check the claim rather than take it: `go vet` clean before each mutant (so each is a real behaviour change and not a compile error), full unfiltered `go test -count=1 ./internal/engine/`, production file restored by `cp` with its shasum re-verified afterwards.

Your disclosed `[1]`-in-a-directory-name limitation stays as you described it — real, and also missed before the fix, so not a regression.

Everything else about the PR stands as reviewed. Merging once the checks land.

If a Windows leg goes red it is very likely not yours — this repo currently carries four intermittent Windows-only failures on bystander tests. I re-run the identical commit to prove flakiness rather than asking you to chase it.

PR #6562, opened 2026-08-23
fix(#6559): the fastapi decorator path capture rejected an empty literal, so a prefixed router's own path was dropped

### LINKED ISSUE #6559: the two fastapi decorator regexes require at least one path character, so a route registered at a prefixed router's own path is dropped before any prefix fold runs
Found while auditing FastAPI endpoint recall against `main` at 8595d7a6b.

Both FastAPI decorator regexes require at least one character between the quotes of the path argument. The capture, identical in `fastapiVerbDecoratorRe` (`internal/engine/http_endpoint_synthesis.go:2855`) and `fastapiApiRouteDecoratorRe` (`:2863`):

```go
\s*\(\s*["']([^"'\n\r]+)["']
```

A decorator whose path literal is `""` therefore never matches, and nothing downstream can recover it: the two `FindAllStringSubmatchIndex` loops at `:3246` and `:3263` are the only decorator sources `synthesizeFastAPI` (`:3177`) has, `synthesizeFastAPIAddRoute` handles only `.add_api_route(`, and the match fails before the receiver gate, the prefix fold or the verb are consulted. There is no branch to reach.

## The correct path is not ambiguous

An empty route path is a normal FastAPI spelling, not a degenerate value. `APIRouter.add_api_route` composes `self.prefix + path`, so on a router built as `APIRouter(prefix="/markets")` a handler decorated `@markets.get("")` serves `/markets`. It is how the collection-root endpoint is declared when a trailing slash is not wanted, and it usually sits directly above the `"/{id}"` route in the same file.

#5688 already makes the prefix available at the point where the drop happens: `fastapiRouterPrefixes` (`:2913`) builds the receiver-to-prefix map and `compose` (`:3207`) applies it to every non-empty path. `TestSynth_FastAPI_RouterPrefixFold_NoDoubleSlashOnRoot` already pins `@router.get("/")` on `APIRouter(prefix="/users")` to `http:GET:/users`, the same canonical path the empty-path form resolves to.

## Constructed example

This file was written for this issue, not quoted from a repository:

```python
from fastapi import FastAPI, APIRouter

app = FastAPI()
markets = APIRouter(prefix="/markets")

@markets.get("")
async def list_markets():
    return []

@markets.get("/{id}")
async def get_market(id: int):
    return {}
```

Run through `runDetect(t, "python", "routers/markets.py", src)` on `main`, the entities are:

```
kind=Route                     name=markets
kind=Service                   name=app
kind=Route                     name=/{id}
kind=http_endpoint_definition  name=http:GET:/markets/{id}
```

No entity of any kind names the empty-path route. The sibling route is emitted normally, which is what makes the result read as a partially indexed router rather than a parsing failure.

## Consequence

The collection-root endpoint of a prefixed router produces no `http_endpoint_definition`, so it gets no IMPLEMENTS edge to its handler and no path for a consumer-side call to match against.

## Bounded scope

Only the two decorator regexes are involved. `fastapiAddAPIRouteRe` (`:2658`) carries the same `+`, but `synthesizeFastAPIAddRoute` is called at `:3199`, before `prefixes` exists, and is passed no compose function, so it folds no router prefix for any path, empty or not. That is a separate gap, not part of this one.

arthurgeron:
Closes #6559

`fastapiVerbDecoratorRe` and `fastapiApiRouteDecoratorRe` required one or more characters between the quotes of the path argument, so `@router.get("")` never matched. The capture is now `*` instead of `+`, and both decorator loops skip a match whose composed path is empty.

## Why the composed path is unambiguous

`compose` already returns the prefix unchanged for an empty route path, because `joinPathFragments(prefix, "")` returns `prefix`. So `APIRouter(prefix="/markets")` plus `@markets.get("")` yields `/markets`, which is what FastAPI serves and is the same canonical path `TestSynth_FastAPI_RouterPrefixFold_NoDoubleSlashOnRoot` already pins for `@router.get("/")` on `APIRouter(prefix="/users")`. No new path arithmetic was needed.

## What still emits nothing, deliberately

An empty path with no same-file prefix names no path at all, so both loops `continue` rather than let `Canonicalize` turn `""` into a bare `/`. Three receivers fall in that bucket:

- the FastAPI app itself, `@app.get("")`
- a router constructed with no `prefix=` kwarg
- a router recognised only by name, `@router.get("")` where `router` was constructed in another file, so `fastapiRouterPrefixes` has no entry for it

That keeps the change additive: no receiver that emitted nothing before starts emitting a root endpoint. `TestSynth_FastAPI_EmptyRoutePathWithoutPrefixNotSynthesized` pins the first two.

## Failing first

The three new tests, run against `8595d7a6b` with only the test file applied:

```
=== RUN   TestSynth_FastAPI_RouterPrefixFold_EmptyRoutePath
    http_endpoint_synthesis_test.go:1103: FastAPI APIRouter prefix fold, empty route path: missing synthetic "http:GET:/markets" (got: [http:GET:/markets/{id}])
    http_endpoint_synthesis_test.go:1103: FastAPI APIRouter prefix fold, empty route path: missing synthetic "http:POST:/markets" (got: [http:GET:/markets/{id}])
--- FAIL: TestSynth_FastAPI_RouterPrefixFold_EmptyRoutePath (0.04s)
=== RUN   TestSynth_FastAPI_RouterPrefixFold_EmptyRoutePathApiRoute
    http_endpoint_synthesis_test.go:1129: FastAPI APIRouter prefix fold, empty route path via api_route: missing synthetic "http:GET:/markets" (got: [])
    http_endpoint_synthesis_test.go:1129: FastAPI APIRouter prefix fold, empty route path via api_route: missing synthetic "http:PUT:/markets" (got: [])
--- FAIL: TestSynth_FastAPI_RouterPrefixFold_EmptyRoutePathApiRoute (0.05s)
=== RUN   TestSynth_FastAPI_EmptyRoutePathWithoutPrefixNotSynthesized
--- PASS: TestSynth_FastAPI_EmptyRoutePathWithoutPrefixNotSynthesized (0.03s)
FAIL
FAIL	github.com/cajasmota/grafel/internal/engine	1.773s
```

The third one passes on `main` by construction: it is the regression pin for the half of the change that must keep emitting nothing.

## The negative half is observed, not assumed

The `continue` guard is the only thing standing between an unprefixed empty path and a bare `/` endpoint, so it is worth checking that a test actually watches it. Mutant: drop the guard from the verb loop, keeping `path := compose(recv, raw)`. `go vet` clean first, so the mutant is a behaviour change and not a compile error.

```
=== RUN   TestSynth_FastAPI_EmptyRoutePathWithoutPrefixNotSynthesized
    http_endpoint_synthesis_test.go:1157: FastAPI empty path without prefix: phantom synthetic "http:GET:/" should NOT be present (got: [http:GET:/ http:GET:/health])
--- FAIL: TestSynth_FastAPI_EmptyRoutePathWithoutPrefixNotSynthesized (0.02s)
```

Restored by `cp`; `shasum` back to `5d656c389162f183489db734fa1bf26e53fb79ac`, byte-identical to pre-mutant.

## Verification

`gofmt -l $(git ls-files '*.go')` clean and `go vet ./...` clean. `go test -race -count=1 -timeout 40m ./...` on the whole tree: 234 packages ok, 13 with no test files, 1 FAIL. The package this PR changes:

```
ok  	github.com/cajasmota/grafel/internal/engine	561.585s
```

### The one failure is pre-existing and unrelated

```
--- FAIL: TestBaselineMeasuredOnIsReachable (0.08s)
    baseline_test.go:204: baseline.json: measured_on "0ca8d585e" is not a commit in this repository
FAIL  github.com/cajasmota/grafel/internal/quality  11.179s
```

`go test -race -count=1 ./internal/quality/ -run TestBaselineMeasuredOnIsReachable` reproduces it identically on unmodified `8595d7a6b`, and in a full clone of `refs/heads/*` `git cat-file -e 0ca8d585e` answers `fatal: Not a valid object name 0ca8d585e`. The cause is the one the test's own doc comment names: `internal/quality/golden/baseline.json` records a `measured_on` sha that resolves in no pushed history. CI does not see it because `actions/checkout` clones shallow, and the test skips on a shallow clone. Nothing in this PR touches `internal/quality`, and it is left alone.

## Not covered

`fastapiAddAPIRouteRe` carries the same restriction on `.add_api_route("")`, but `synthesizeFastAPIAddRoute` runs before `prefixes` is built and is passed no compose function, so it folds no router prefix for any path. Widening it alone would produce a path of `""`, not the router's path, so it is left alone and stated in #6559 as a separate gap.

### DIFF
--- a/internal/engine/http_endpoint_synthesis.go
+++ b/internal/engine/http_endpoint_synthesis.go
@@ -2851,16 +2851,25 @@ func parseFlaskMethods(args string) []string {
 // terminate the match prematurely (the inner `)` previously aborted the scan,
 // dropping the whole endpoint — #3628).
 //
+// The path capture accepts an EMPTY literal (`*`, not `+`): `@router.get("")`
+// on a prefixed router is the idiomatic spelling of the collection-root
+// endpoint without a trailing slash, and FastAPI composes it to the prefix
+// itself. synthesizeFastAPI drops the match when no prefix is available to
+// compose, so an empty path never becomes a bare `/` endpoint.
+//
 // Capture groups: 1 = receiver, 2 = verb, 3 = path, 4 = handler name.
-var fastapiVerbDecoratorRe = regexp.MustCompile(`@(\w+)\.(get|post|put|patch|delete|head|options|trace)\s*\(\s*["']([^"'\n\r]+)["'](?:[^()]*(?:\([^()]*\)[^()]*)*)\)\s*[\r\n]+(?:\s*@[^\n\r]*[\r\n]+)*\s*(?:async\s+)?def\s+(\w+)`)
+var fastapiVerbDecoratorRe = regexp.MustCompile(`@(\w+)\.(get|post|put|patch|delete|head|options|trace)\s*\(\s*["']([^"'\n\r]*)["'](?:[^()]*(?:\([^()]*\)[^()]*)*)\)\s*[\r\n]+(?:\s*@[^\n\r]*[\r\n]+)*\s*(?:async\s+)?def\s+(\w+)`)
 
 // fastapiApiRouteDecoratorRe captures the generic `@<recv>.api_route("/path",
 // methods=[...])` decorator form. FastAPI's `api_route` differs from the
 // single-verb shorthands (`.get`/`.post`/...) in that the HTTP verb(s) are
 // declared via a `methods=[...]` kwarg rather than the method name itself.
 //
+// The path capture accepts an EMPTY literal for the same reason as
+// fastapiVerbDecoratorRe.
+//
 // Capture groups: 1 = receiver, 2 = path, 3 = kwargs tail, 4 = handler name.
-var fastapiApiRouteDecoratorRe = regexp.MustCompile(`@(\w+)\.api_route\s*\(\s*["']([^"'\n\r]+)["']((?:[^()]*(?:\([^()]*\)[^()]*)*))\)\s*[\r\n]+(?:\s*@[^\n\r]*[\r\n]+)*\s*(?:async\s+)?def\s+(\w+)`)
+var fastapiApiRouteDecoratorRe = regexp.MustCompile(`@(\w+)\.api_route\s*\(\s*["']([^"'\n\r]*)["']((?:[^()]*(?:\([^()]*\)[^()]*)*))\)\s*[\r\n]+(?:\s*@[^\n\r]*[\r\n]+)*\s*(?:async\s+)?def\s+(\w+)`)
 
 // fastapiRouterConstructorRe captures a same-file `<var> = APIRouter(...)`
 // constructor call (#5688). Group 1 is the receiver variable, group 2 is the
@@ -3203,6 +3212,11 @@ func synthesizeFastAPI(content, relPath string, emit emitDefFn, emitMount func(t
 	// receiver → prefix map once and compose it into every decorator route
 	// registered on that receiver. Receivers absent from the map (e.g. `app`,
 	// the FastAPI application itself) compose as a no-op.
+	// An empty route path (`@router.get("")`) names the prefix itself, so
+	// compose returns "" whenever there is no same-file prefix to supply the
+	// path. Callers skip such a match rather than canonicalising "" into a
+	// bare `/`, which would attribute a root endpoint to a router whose real
+	// mount point this file does not state.
 	prefixes := fastapiRouterPrefixes(content)
 	compose := func(recv, raw string) string {
 		if pfx, ok := prefixes[recv]; ok && pfx != "" {
@@ -3254,7 +3268,11 @@ func synthesizeFastAPI(content, relPath string, emit emitDefFn, emitMount func(t
 		verb := strings.ToUpper(content[idx[4]:idx[5]])
 		raw := content[idx[6]:idx[7]]
 		handler := content[idx[8]:idx[9]]
-		canonical := httproutes.Canonicalize(httproutes.FrameworkFastAPI, compose(recv, raw))
+		path := compose(recv, raw)
+		if path == "" {
+			continue
+		}
+		canonical := httproutes.Canonicalize(httproutes.FrameworkFastAPI, path)
 		defLine := lineOfOffset(content, idx[8])
 		emit(verb, canonical, "fastapi", "Controller", handler, defLine)
 	}
@@ -3271,7 +3289,11 @@ func synthesizeFastAPI(content, relPath string, emit emitDefFn, emitMount func(t
 		raw := content[idx[4]:idx[5]]
 		tail := content[idx[6]:idx[7]]
 		handler := content[idx[8]:idx[9]]
-		canonical := httproutes.Canonicalize(httproutes.FrameworkFastAPI, compose(recv, raw))
+		path := compose(recv, raw)
+		if path == "" {
+			continue
+		}
+		canonical := httproutes.Canonicalize(httproutes.FrameworkFastAPI, path)
 		defLine := lineOfOffset(content, idx[8])
 		methods := parseFlaskMethods(tail) // same methods=[...] shape
 		if len(methods) == 0 {
--- a/internal/engine/http_endpoint_synthesis_test.go
+++ b/internal/engine/http_endpoint_synthesis_test.go
@@ -1071,6 +1071,92 @@ async def item(id: int):
 	requireContains(t, got, want, "FastAPI APIRouter prefix fold — api_route decorator")
 }
 
+// TestSynth_FastAPI_RouterPrefixFold_EmptyRoutePath covers a route path of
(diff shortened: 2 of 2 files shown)


cajasmota:
Independent review: **the shipped behaviour is correct on every shape probed — but half the guard this PR adds is untested, and the mutant proving it is the same one you already demonstrated on the other half.**

No production change needed.

## Blocking: the `api_route` arm of the empty-path guard is unpinned

Deleting `if path == "" { continue }` from the **`api_route`** loop only — leaving the verb loop's guard intact — leaves the whole package green: `go vet` clean, `go test -count=1 ./internal/engine/` → **ok 30.6s**.

What that mutant lets through:

```python
from fastapi import FastAPI, APIRouter
app = FastAPI()
bare = APIRouter()

@app.api_route("", methods=["GET"])
async def a(): ...

@bare.api_route("", methods=["POST"])
async def b(): ...
```

Shipped code: `[]`. Mutant: `[http:GET:/ http:POST:/]` — exactly the phantom root endpoint your PR body says the guard exists to prevent.

You demonstrated this mutant on the verb loop and it died. `TestSynth_FastAPI_EmptyRoutePathWithoutPrefixNotSynthesized` exercises only `@app.get("")` and `@bare_router.get("")`, so the identical mutation on the `api_route` loop survives. **Extending that test with an `api_route` receiver having no prefix closes it** — nothing else changes.

## Everything else held

| shape | result |
|---|---|
| `@app.get("")`, no prefix | `[]` — as claimed, no bare `/` |
| `@bare_router.get("")` | `[]` |
| `@app.api_route("", methods=[...])`, no prefix | `[]` in shipped code |
| no path argument at all (`@markets.get()`, `@markets.post`) | correctly declined |
| `APIRouter(prefix="")` | guard holds |

Both `fastapiVerbDecoratorRe` and `fastapiApiRouteDecoratorRe` are referenced only at their two guarded call sites (3260, 3281), so no unguarded consumer can turn `""` into `/`.

## Two notes, neither blocking

**The change is not quite purely additive.** `*` lets the capture succeed on any decorator whose first argument merely *begins* with an empty string literal. On a prefixed router, `@markets.get(''.join(segments))` now emits `http:GET:/markets` where it previously emitted nothing, and `@markets.get("""/tri""")` emits `http:GET:/markets` — the wrong path, since the real route is `/markets/tri`. Both are rare and neither can produce a bare `/`, because your guard still covers the unprefixed case. RE2 has no lookahead, so anchoring this properly is not available; a sentence in the regex comment is the honest fix rather than a code change.

**`""` and `"/"` collapse** on a prefixed router — both canonicalize to `/markets`, one entity survives, attributed to whichever decorator appears first. FastAPI serves those as two distinct routes, but grafel's trailing-slash canonicalization already made that choice pre-PR; only the winner flipped (from the `/` handler to the `""` handler). Policy, not a regression.

**Keyword form** `@markets.get(path="")` is still unsupported — pre-existing, and `+`→`*` does not change it.

## Golden movement: none

`python-fastapi-mini` scores 5/5 entities, 4/4 relationships, 0 forbidden on both your branch and the base tree; the fixture contains no empty route path. `csharp-aspnet-core-mini` unchanged at 4/5.

Any `./internal/quality/` red you see is **pre-existing and not yours** — `TestBaselineMeasuredOnIsReachable`, where `baseline.json`'s `measured_on 0ca8d585e` is not an ancestor of `origin/main`. It fails on main too. Filing it separately.

Happy to push the `api_route` row myself if you would rather not context-switch — otherwise I will merge as soon as it lands.

cajasmota:
Pushed as `db2f6fddc`.

Added to `TestSynth_FastAPI_EmptyRoutePathWithoutPrefixNotSynthesized`: `@app.api_route("", methods=["GET"])` and `@bare_router.api_route("", methods=["POST"])`, plus `http:POST:/` and `http:POST:` in the `requireNotContains` set.

**api_route mutant** (drop `if path == "" { continue }` from the api_route loop only) — before: package green, `ok 31.7s`. After: **dies**, `got: [http:GET:/ http:GET:/health http:POST:/]`, both phantom ids reported. The **verb-loop mutant still dies** too, so neither half is unobserved now.

---

**I pushed the row rather than asking you to — and I was wrong to ask in the first place.** That is the norm here going forward: a round-trip for something a maintainer can do in minutes costs you a context-switch and turns a merge into a wait.

Your commits are untouched: one commit added on top, no rebase, no squash, no amend, no force-push. The commit message says the addition was mine so you are not credited with work you did not do.

Scoring method, so you can check the claim rather than take it: `go vet` clean before each mutant (so each is a real behaviour change and not a compile error), full unfiltered `go test -count=1 ./internal/engine/`, production file restored by `cp` with its shasum re-verified afterwards.

Everything else about the PR stands as reviewed. Merging once the checks land.

If a Windows leg goes red it is very likely not yours — this repo currently carries four intermittent Windows-only failures on bystander tests. I re-run the identical commit to prove flakiness rather than asking you to chase it.

Issue #6444, opened 2026-08-20
fastapi rules anchor edges on Route:<function> while Route entities are named by path, so every decorated handler emits a permanently-dangling edge parked in the dynamic bucket


cajasmota:
`internal/engine/rules/python/frameworks/fastapi.yaml:110-116` emits a relationship whose source is `source_type: Route` with `source_group: 1` — the captured **function** name. But FastAPI `Route` entities are named by **path**. The two never agree, so the edge names a Route that cannot exist.

Confirmed by running the detector on a minimal FastAPI file:

```
PROBEENT kind=Route name="/things"
PROBEREL kind=DECORATES from="Route:list_things" to="Service:list_things"
         props=[{framework python} {pattern_type yaml_driven}]
```

The entity is `/things`. The edge points at `Route:list_things`. It is permanently unresolvable — not "unresolved until the resolver improves", but naming something that is never minted under that name.

## Why nobody has seen it

The edge carries `framework` and `pattern_type` but **no `language`/`lang` property**, so `relLanguage()` (`internal/resolve/refs.go:6231`) returns `""`. That routes it to the `lang == ""` arm of the Dynamic escape hatch, where it is classified `DispositionDynamic` — accounted as "framework-mediated, not an extractor bug". So every decorated FastAPI handler contributes one permanently-dangling edge and the resolver-bug rate stays at 0.0%.

This is the same accounting shape that hid #6429 for as long as it did: a real defect parked in the dynamic bucket, where nothing counts it.

## Found by

An adversarial review of #6443. That PR proposed dropping the `Route:` arm of the Dynamic hatch, on the stated grounds that `Route:<path>` reaches the classifier only as a `ROUTES_TO` edge's FROM side. The reviewer disproved that by execution — the probe above — and the arm was narrowed to `lang == "java"` so the reclassification would not silently land on Python.

Worth being explicit: reclassifying these to `DispositionBugExtractor` would arguably be *honest*, since they really are rule bugs. But it would raise the reported resolver-bug rate across the whole Python corpus as a side effect of a Java change, unmeasured. That is why it was kept out. Fixing the rule is the right order; re-classifying after is then a measurement, not a surprise.

## Same shape suspected in seven more rule files

`target_type: Route` or `source_type: Route` appears with the same group-capture pattern in:

- `internal/engine/rules/python/frameworks/flask.yaml:115`
- `internal/engine/rules/python/frameworks/django.yaml:141`
[...]

PR #6565, opened 2026-08-24
fix(#6511): the Python DI passes addressed both INJECTED_INTO endpoints by bare name, so they could bind anywhere

### LINKED ISSUE #6511: the AST fastapi_depends pass emits INJECTED_INTO with both endpoints as unresolved bare-name stubs — proven by a mutant in #6444, not by inspection
Found while fixing #6444, and proven by a mutant there rather than by inspection.

## The defect

The AST `fastapi_depends` pass emits its own `INJECTED_INTO` edge, and **both endpoints are unresolved bare-name stubs**:

```
STUB<get_db> -> STUB<list_things>
```

No kind prefix, no resolution. The corresponding entities exist — the generic Python extractor mints `SCOPE.Operation:get_db` and `SCOPE.Operation:list_things` — so these are edges that *could* bind and do not.

## How it was found, which is the interesting part

#6444 repaired the YAML `INJECTED_INTO` rule, whose target had been anchored on the unmintable `Route:<function>`. Mutant **M4** reverted that anchor, expecting the row to survive on the strength of the AST pass emitting the same relationship.

**It did not survive.** M4 died — which proves the AST-emitted edge does **not** satisfy a `must_exist` row naming real entities, because it is itself a dangling stub. Had the AST edge been resolving correctly, M4 would have been an equivalent mutant and the YAML repair would have been redundant.

So the AST pass and the YAML rule were both emitting `INJECTED_INTO`, both broken, in different ways, and neither observable.

## Why nothing reports it

Same hatch as #6444: these stubs carry no `language`/`lang` property, so `relLanguage` (`internal/resolve/refs.go:6713`) returns `""` and `classifyDispositionLang` parks them in `DispositionDynamic` — "framework-mediated runtime dispatch", **explicitly excluded from the resolver-bug rate**.

A bare-name stub with no kind prefix may take a different arm of that classifier than the `Route:`-prefixed case; that is worth establishing as part of the fix, because it determines whether the dangling-edge count would move at all once this is repaired.

1. **What should these endpoints be?** `SCOPE.Operation:<name>` for both, presumably — matching what #6444 settled for the YAML rule, where `target_type: Operation` was confirmed to bare-name-bind via `BuildIndex`'s dual-indexing of `SCOPE.*` kinds under their trimmed key (`refs.go:1190-1219`).
2. **Should the AST pass emit this at all**, now that the YAML rule resolves correctly? Two passes emitting the same relationship is how #6442's phantom-duplicate suspicion started. If both are kept, dedup needs deciding; if one is dropped, which.
3. Whether the same shape affects the other AST-driven FastAPI passes.

## Observability

`python-fastapi-mini` now exists (added by #6444) and is the natural home. It already asserts the YAML `INJECTED_INTO` edge between real entities.

**Killing mutant:** revert whichever anchor lands. Note the fixture must distinguish the two producers — if the YAML rule alone satisfies the row, an AST-side regression is invisible. Establishing that separation is part of the task, not an afterthought.

cajasmota:
All three INJECTED_INTO emit sites in `internal/custom/python/di_graph.go`
(`fastapi_depends`, `dependency_injector_inject`, `litestar_provide`) wrote raw
bare identifiers into FromID/ToID: `get_db -> list_things`. No kind, no file
anchor — the resolver reached those through its global byName tier, where any
same-named entity in the repo is a candidate, and where an ambiguous name
resolves to nothing at all and parks in `DispositionDynamic`, outside the
resolver-bug rate.

## Endpoint addressing

**consumer (ToID, all three sites)** — `extractor.BuildOperationStructuralRef`,
i.e. `scope:operation:method:python:<file>:<fn>`. The consumer IS the file being
extracted, so the anchor is knowable and honest. `lookupStructural` returns
`handled=true` even on a miss, so this shape can NEVER fall through to bare-name
matching. Same builder the http-endpoint handler bridge and every class→method
CONTAINS edge already use.

**provider (FromID, fastapi + litestar)** — the kind prefix
`SCOPE.Operation:<name>` **only when this extractor has observed the provider to
be a module-level callable**, i.e. a `def` / `async def` at **column 0** of the
file being extracted (`pyLocalDefNames` / `rePyModuleDef`). Otherwise **bare**.

**provider (FromID, `dependency_injector_inject`)** — left as the container
attribute. It is a DI token, the same address its BINDS edge uses, not an
operation; prefixing it would assert a kind the symbol lacks.

## Why the provider end is gated (review fix, `9fe52ae0f`)

The first pass of this branch (`fda47b4a7`) prefixed the provider
**unconditionally**. That is a fabricated kind: a FastAPI provider is very often
a **class**, and both idiomatic forms land in the same code path —

```python
def me(svc = Depends(SvcClass)):          # class passed positionally
def me(svc: AuthService = Depends()):     # bare Depends() → the annotation
```

This is not cosmetic. `internal/resolve`'s `LookupStatusHint` probes
`byKind[kind][name]` **before** the kind-agnostic `byName` tier, so a wrong kind
prefix does not merely fail to help — it **promotes** a wrong-kind, same-named
entity ahead of the entity actually being injected.

Measured with two real indexer binaries. Fixture: `app/routers.py` defines
`class AuthService` and `def me(svc: AuthService = Depends())`; `app/legacy.py`
defines an unrelated `def AuthService`.

```
fda47b4a7 : INJECTED_INTO  SCOPE.Operation|AuthService (app/legacy.py)   ->  SCOPE.Operation|me
parent    : INJECTED_INTO  SCOPE.Component|AuthService (app/routers.py)  ->  UNRESOLVED<me>
```

The prefix bound the provider to a function in **another file** with nothing to
do with the injection — the exact defect class #6511 exists to fix, newly
introduced on the provider side. Delete `legacy.py` and both bind; the
same-name collision is the trigger.

Bare is the honest default here because it is kind-**agnostic**, not kind-wrong:
it resolves on `byName`, and when that name is ambiguous it falls to
`hintKinds("INJECTED_INTO") = componentKindFamily`, which is precisely what
binds the class correctly. This is the reasoning this PR already applied to the
`dependency_injector_inject` token — "prefixing it would assert a kind the
symbol lacks" — applied verbatim to a class provider.

The consumer half of the original fix is unchanged.

## Why the gate is MODULE-LEVEL, not any `def` (review follow-up, `e0bc32152`)

The first version of the gate was built on `pyFunctions`, whose `rePyDef` begins
`^[ \t]*def`. That leading `[ \t]*` means an **indented** def — a **method** —
satisfied it:

```python
class Registry:
    def AuthService(self): ...

class AuthService: ...

def me(svc: AuthService = Depends()): ...
#  => localDefs["AuthService"] == true
```

So a class provider colliding with any method anywhere in the same file was
prefixed again and the byKind-before-byName promotion returned. Provider factory
methods on a container/registry class are a real DI idiom, so this was not
theoretical.

`pyLocalDefNames` now uses a **new** `rePyModuleDef` anchored at column 0.
`rePyDef` is deliberately untouched — `pyFunctions` is shared with the consumer
side and with endpoint-handler extraction and must keep seeing methods. The
tightening is strictly more conservative (fewer prefixes, more bare refs), and
it is a **column** rule, not a ban on async: `async def` at column 0 still
qualifies.

## Node identity decoupled from endpoint addressing

`addEdge` built the owner `SCOPE.Pattern` entity's `Name` out of
`rel.FromID`/`rel.ToID`, so re-addressing the endpoints renamed **every** DI
pattern entity and changed its ComputeID:

```
di:INJECTED_INTO:get_db->list_things@9
  -> di:INJECTED_INTO:SCOPE.Operation:get_db->scope:operation:method:python:app/routers/things.py:list_things@9
```

— churn for entities outside the three emit sites, and a nested structural ref
inside an entity `Name`. `addEdgeNamed` now takes the semantic provider/consumer
names, so the owner names are byte-identical to the pre-#6511 ones. Pinned by
`TestPyDIProvider6511_OwnerEntityNameIsStableUnderEndpointReaddressing`.

## Measured, `internal/quality/golden/python-fastapi-mini`

| binary | relationships | resolver stats |
|---|---|---|
| parent `8595d7a6b` | 57 | rewrote=27 ambiguous=2 (from am=1, to am=1) |
| 1st pass `fda47b4a7` | 56 | rewrote=29 ambiguous=0 |
| `9fe52ae0f` | 57 | rewrote=28 ambiguous=1 (from am=1, to am=0) |
| this head `e0bc32152` | 57 | rewrote=28 ambiguous=1 (from am=1, to am=0) |

Recall unchanged throughout: **5/5 entities, 4/4 relationships**, exactly the
`baseline.json` figures, so no re-record. The module-level tightening moved
nothing — the fixture's `get_db` is a module-level `async def`, so it still
qualifies and still folds identically.

**The consumer fix is worth +1 rewrite and to-side ambiguous 1 → 0** — the
dangling `INJECTED_INTO UNRESOLVED|get_db -> UNRESOLVED|list_things` now has a
resolved consumer end. That is end-to-end proof the consumer half works, and it
was missing from the earlier body.

The remaining from-side ambiguity is `get_db`, which shares its bare name with a
`fastapi.yaml` `Dependency` entity **in the same file**, so neither a file
anchor nor the component-family hint can separate them. The first pass bought
that last rewrite — and the 57→56 dedup into the YAML edge — with a guessed kind
that is wrong for class providers. One honestly-unresolved endpoint is the
correct trade for one confidently-wrong binding. Recovering it *without*
fabricating a kind means widening `hintKinds("INJECTED_INTO")` to
`componentOrOperationKindFamily` in `internal/resolve`: a resolver-side change,
out of scope here, and one that would make the `AuthService` case honestly
ambiguous rather than correct.

## Tests

Red first on `fda47b4a7`:

- `TestPyDIProviderDoesNotBindToUnrelatedSameNamedFunction_6511` (package
  `python_test`; the real extractor plus `resolve.BuildIndex` /
  `resolve.ReferencesEmbedded`) failed with the provider bound to
  `def AuthService` in `app/legacy.py` — the reviewer's collision, reproduced
  through the real resolver. Independently confirmed to have teeth: dropping
  `INJECTED_INTO` from the `componentKindFamily` case in `refs.go` kills it.
- `TestPyDIProvider6511_ClassProviderIsNotAddressedAsAnOperation` failed on all
  three sub-cases (positional class, annotation class, imported callable), and
  `..._LitestarClassProviderIsNotAddressedAsAnOperation` with it.

Red first on `9fe52ae0f`:

- `TestPyDIProvider6511_MethodDoesNotMakeAClassProviderCallable` and its
  litestar twin failed with `SCOPE.Operation:AuthService`, want bare
  `AuthService`.
- `TestPyDIProvider6511_ModuleLevelAsyncDefStillCounts` is the non-vacuity
  guard on that tightening — already green, and it pins that the rule is about
  the column, not about `async`.

All green after. The expectations that blessed the wrong behaviour
(`SCOPE.Operation:SvcClass`, `SCOPE.Operation:AuthService` in `di_graph_test.go`,
and the two `SCOPE.Operation:get_db` providers in `di_graph_refs_6511_test.go`)
were **changed**, not worked around. Every consumer-end assertion stayed as it
was and stayed green.

Mutants scored — `go vet` first, `go test -count=1`, whole
`./internal/custom/python/` package, no `-run` filter, production file restored
by `cp` from a byte-level backup and re-verified with `shasum`:

|---|---|---|
| M3 — every provider treated as callable (the reverted rule) | permissive | **DEAD** — both class cases, the imported case, the resolver-level collision gate. Re-scored against the tightened gate, still dead |
| M4 — never prefix; the rule collapses to "always bare" | restrictive | **DEAD** — `..._LocallyDefinedCallableProviderKeepsItsKind` plus the two pre-existing local-`def` fixtures |
| M5 — "known-callable" widened to include `class` declarations | permissive | **DEAD** — the positional-class case and the collision gate |
| M6 — owner name rebuilt from the endpoint refs | — | **DEAD** — `..._OwnerEntityNameIsStableUnderEndpointReaddressing` |
| P1 — `rePyModuleDef` regains `[ \t]*`, any indentation | permissive | **DEAD** — both method-collision tests |
| P2 — `pyLocalDefNames` body reverted to the `pyFunctions` loop | permissive | **DEAD** — both method-collision tests |
| R1 — `rePyModuleDef` drops the `async` alternation | restrictive | **DEAD** — `..._ModuleLevelAsyncDefStillCounts` and the pre-existing `TestPyDI_LitestarProvideBindsAndInject` |

Observed Go-level on `PyDIExtractor.Extract` and at the resolver seam, not in
the golden fixture — `python-fastapi-mini` pins `INJECTED_INTO get_db ->
list_things` but cannot tell the #6444 YAML rule from this pass, which is why
the extractor-side edge was unobserved for as long as it was.

## Not observable by a test here

- Whether a **non-golden, real-world** Python corpus contains provider names
  that the module-level gate now leaves bare and that were previously resolving
  through the prefix. The golden fixture is the only Python DI corpus wired into
  this repo's gates; the fixture's own from-side ambiguity is reported above,
  but the corpus-wide delta was not measured.
- Whether widening `hintKinds("INJECTED_INTO")` to the component∪operation union
  is safe for the NestJS / Angular / Spring DI emitters that share that hint.
  Asserted from reading `refs.go`, not measured — which is why it is left as
  follow-up rather than done here.
- A module-level `def` inside an `if`/`try` block is indented and therefore does
  NOT satisfy the gate, so such a provider stays bare. That is the conservative
  direction (kind-agnostic, never kind-wrong) and is deliberate, but no test
  distinguishes "correctly conservative" from "missed a real callable" for that
  shape.

Closes #6511

### DIFF
--- a/internal/custom/python/di_graph.go
+++ b/internal/custom/python/di_graph.go
@@ -103,13 +103,35 @@ func pyDIProviderRef(localDefs map[string]bool, name string) string {
 	return name
 }
 
-// pyLocalDefNames is the set of `def` names declared in the file being
-// extracted — the only providers whose Operation kind pyDIProviderRef has
-// actually observed.
+// rePyModuleDef is rePyDef anchored at COLUMN 0: a module-level `def` /
+// `async def`, never an indented one. Deliberately a separate matcher rather
+// than a tightening of rePyDef, which pyFunctions shares with the consumer
+// side and with endpoint-handler extraction and which MUST keep seeing methods.
+var rePyModuleDef = regexp.MustCompile(`(?m)^(?:async[ \t]+)?def[ \t]+(\w+)[ \t]*\(`)
+
+// pyLocalDefNames is the set of MODULE-LEVEL `def` names declared in the file
+// being extracted — the only providers whose Operation kind pyDIProviderRef
+// has actually observed.
+//
+// Module-level, not any `def`. rePyDef begins `^[ \t]*def`, so an INDENTED def
+// — a METHOD — would satisfy the gate, and a provider factory method on a
+// container/registry class is a real DI idiom. That would let
+//
+//	class Registry:
+//	    def AuthService(self): ...
+//	class AuthService: ...
+//	def me(svc: AuthService = Depends()): ...
+//
+// stamp "SCOPE.Operation:" on a CLASS provider again and reopen the
+// byKind-before-byName promotion this whole gate exists to prevent. A method
+// is not evidence about a module-level name.
+//
+// Strictly more conservative on purpose: fewer prefixes, more bare refs, and
+// bare is kind-AGNOSTIC rather than kind-wrong.
 func pyLocalDefNames(src string) map[string]bool {
 	defs := make(map[string]bool)
-	for _, fn := range pyFunctions(src) {
-		defs[fn.name] = true
+	for _, m := range rePyModuleDef.FindAllStringSubmatch(src, -1) {
+		defs[m[1]] = true
 	}
 	return defs
 }
--- a/internal/custom/python/di_graph.go
+++ b/internal/custom/python/di_graph.go
@@ -49,6 +49,40 @@ type PyDIExtractor struct{}
 
 func (e *PyDIExtractor) Language() string { return "python_di_graph" }
 
+// pyDIConsumerRef addresses the CONSUMER end of an INJECTED_INTO edge — the
+// enclosing `def` — as the canonical file-anchored operation structural-ref
+// (issue #6511). The consumer's file is known exactly: it is the file being
+// extracted. Emitting the bare name instead let the resolver's global byName
+// tier bind the edge to ANY same-named function anywhere in the repo, which is
+// precisely the mis-binding #6511 reports.
+//
+// internal/resolve's lookupStructural handles this shape through
+// lookupLocationKind(file, name, operationKindFamily) and returns handled=true
+// even on a miss — so an unresolvable consumer stays honestly unresolved
+// rather than degrading to a bare-name guess. Same builder the http-endpoint
+// handler bridge and every language's class→method CONTAINS edge use.
+func pyDIConsumerRef(filePath, fn string) string {
+	return extractor.BuildOperationStructuralRef("python", filePath, fn)
+}
+
+// pyDIProviderRef addresses the PROVIDER end of an INJECTED_INTO edge when the
+// provider is a source-level callable or class (fastapi `Depends(x)`, litestar
+// `Provide(x)`). Unlike the consumer, the provider is routinely IMPORTED from
+// another module — python-fastapi-mini's `get_db` lives in app/deps.py while
+// its consumer lives in app/routers/things.py — so its file is NOT knowable
+// here and a file anchor would be a fabrication. It carries the kind prefix
+// only, the same address #6444 settled on for fastapi.yaml's INJECTED_INTO
+// rule; BuildIndex dual-indexes SCOPE.* kinds under their trimmed key, so
+// "SCOPE.Operation:<name>" binds to the generic extractor's Operation entity.
+//
+// Deliberately NOT applied to the dependency-injector @inject provider: that
+// endpoint is a container attribute (a DI token, the same address its BINDS
+// edge uses), not an operation, and prefixing it would assert a kind the
+// symbol does not have.
+func pyDIProviderRef(name string) string {
+	return "SCOPE.Operation:" + name
+}
+
 var (
 	// rePyDef captures a `def name(` head with the byte offset of the opening
 	// paren so the signature can be balanced-scanned. Group 1 = function name.
@@ -148,8 +182,8 @@ func (e *PyDIExtractor) Extract(ctx context.Context, file extractor.FileInput) (
 					"via":       "fastapi_depends",
 				},
 				types.RelationshipRecord{
-					FromID: provider,
-					ToID:   fn.name,
+					FromID: pyDIProviderRef(provider),
+					ToID:   pyDIConsumerRef(file.Path, fn.name),
 					Kind:   string(types.RelationshipKindInjectedInto),
 					Properties: types.Props{
 						{K: "consumer", V: fn.name},
@@ -225,7 +259,7 @@ func (e *PyDIExtractor) Extract(ctx context.Context, file extractor.FileInput) (
 				},
 				types.RelationshipRecord{
 					FromID: token,
-					ToID:   fn.name,
+					ToID:   pyDIConsumerRef(file.Path, fn.name),
 					Kind:   string(types.RelationshipKindInjectedInto),
 					Properties: types.Props{
 						{K: "consumer", V: fn.name},
@@ -306,8 +340,8 @@ func (e *PyDIExtractor) Extract(ctx context.Context, file extractor.FileInput) (
 						"via":       "litestar_provide",
 					},
 					types.RelationshipRecord{
-						FromID: provider,
-						ToID:   fn.name,
+						FromID: pyDIProviderRef(provider),
+						ToID:   pyDIConsumerRef(file.Path, fn.name),
 						Kind:   string(types.RelationshipKindInjectedInto),
 						Properties: types.Props{
 							{K: "consumer", V: fn.name},
--- a/internal/custom/python/di_graph.go
+++ b/internal/custom/python/di_graph.go
@@ -65,22 +65,53 @@ func pyDIConsumerRef(filePath, fn string) string {
 	return extractor.BuildOperationStructuralRef("python", filePath, fn)
 }
 
-// pyDIProviderRef addresses the PROVIDER end of an INJECTED_INTO edge when the
-// provider is a source-level callable or class (fastapi `Depends(x)`, litestar
-// `Provide(x)`). Unlike the consumer, the provider is routinely IMPORTED from
-// another module — python-fastapi-mini's `get_db` lives in app/deps.py while
-// its consumer lives in app/routers/things.py — so its file is NOT knowable
-// here and a file anchor would be a fabrication. It carries the kind prefix
-// only, the same address #6444 settled on for fastapi.yaml's INJECTED_INTO
-// rule; BuildIndex dual-indexes SCOPE.* kinds under their trimmed key, so
-// "SCOPE.Operation:<name>" binds to the generic extractor's Operation entity.
+// pyDIProviderRef addresses the PROVIDER end of an INJECTED_INTO edge for the
+// fastapi `Depends(x)` and litestar `Provide(x)` passes.
 //
-// Deliberately NOT applied to the dependency-injector @inject provider: that
-// endpoint is a container attribute (a DI token, the same address its BINDS
-// edge uses), not an operation, and prefixing it would assert a kind the
-// symbol does not have.
-func pyDIProviderRef(name string) string {
-	return "SCOPE.Operation:" + name
+// Unlike the consumer, the provider's file is NOT knowable here — it is
+// routinely IMPORTED from another module (python-fastapi-mini's `get_db` lives
+// in app/deps.py while its consumer lives in app/routers/things.py) — so a
+// file anchor would be a fabrication.
+//
+// A KIND is a fabrication too, unless it was observed. A FastAPI provider is
+// very often a CLASS rather than a function: `Depends(SvcClass)` and
+// `svc: AuthService = Depends()` are both idiomatic and both arrive here. So
+// the kind prefix is emitted ONLY when this extractor has actually seen a
+// `def` of that name in the file it is extracting — `localDefs` — which is the
+// one case where "Operation" is a fact rather than an assertion.
+//
+// Why this matters, measured (#6511 review): internal/resolve's
+// LookupStatusHint probes byKind[kind][name] BEFORE the kind-agnostic byName
+// tier. An unconditional "SCOPE.Operation:" prefix therefore does not merely
+// fail to help a class provider — it actively PROMOTES a wrong-kind, same-named
+// entity ahead of the entity being injected. With `class AuthService` in
+// app/routers.py and an unrelated `def AuthService` in app/legacy.py, the
+// prefixed form bound the edge to app/legacy.py; the bare form bound it to the
+// class, because an ambiguous bare name falls to
+// hintKinds("INJECTED_INTO") = componentKindFamily.
+//
+// Bare is therefore the honest default here: it is kind-AGNOSTIC, not
+// kind-wrong, and it keeps both the class and the imported-callable case
+// reachable. This is the same reasoning the dependency-injector @inject
+// provider already followed — that endpoint is a container attribute (a DI
+// token, the same address its BINDS edge uses), and prefixing it "would assert
+// a kind the symbol lacks". A class provider is the identical situation.
+func pyDIProviderRef(localDefs map[string]bool, name string) string {
+	if localDefs[name] {
+		return "SCOPE.Operation:" + name
+	}
+	return name
+}
+
+// pyLocalDefNames is the set of `def` names declared in the file being
+// extracted — the only providers whose Operation kind pyDIProviderRef has
+// actually observed.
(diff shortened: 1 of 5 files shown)


cajasmota:
Independent adversarial review — **REQUEST-CHANGES on my own PR.** Recording it in full because the finding is a regression this PR introduces, and because mutation scoring could not have caught it.

## The defect: the provider end binds worse than `main` does

`pyDIProviderRef` stamps a `SCOPE.Operation:` prefix onto the INJECTED_INTO `FromID` for `fastapi_depends` and `litestar_provide`. But a FastAPI provider is very often a **class** — `Depends(SvcClass)` and `svc: AuthService = Depends()` are both idiomatic — so the prefix asserts a kind the symbol does not have.

Measured end-to-end, PR binary against parent binary. Fixture: `app/routers.py` defines `class AuthService` and `def me(svc: AuthService = Depends())`; `app/legacy.py` defines an unrelated `def AuthService()`.

```
PR  (fda47b4a7): INJECTED_INTO  SCOPE.Operation|AuthService (app/legacy.py)   ->  SCOPE.Operation|me
parent         : INJECTED_INTO  SCOPE.Component|AuthService (app/routers.py)  ->  UNRESOLVED<me>
```

The PR binds the provider to **an unrelated function in a different file**. `resolveOne` probes `byKind["SCOPE.Operation"][name]` before falling back to `byName`, so the fabricated prefix actively *promotes* a wrong-kind same-named entity ahead of the class actually being injected. Delete `legacy.py` and both binaries bind correctly — the same-name collision is the trigger.

So on the provider end this PR takes a correct binding and makes it wrong. That is the exact defect class #6511 exists to fix, reintroduced from the other side.

Worse, the answer was already written in this PR. For the `dependency_injector_inject` site the body says prefixing the container token "would assert a kind the symbol lacks", and leaves it bare. That reasoning applies verbatim to a class provider and was not applied there.

**The PR's own tests bless the wrong behaviour** — `SCOPE.Operation:SvcClass` and `SCOPE.Operation:AuthService` are expected values. Those expectations have to change.

## Why the mutants missed it

Every mutant of `di_graph.go` died, including two new permissive ones:

| # | mutation | vet | test | result |
|---|---|---|---|---|
| M1 | site 1 `FromID`/`ToID` back to bare names | 0 | 1 | **DEAD** — failure names `fastapi_depends` |
| M2 | `pyDIConsumerRef` → `"SCOPE.Operation:"+fn` | 0 | 1 | **DEAD** — 8 fails, all 3 sites + guard |
| P1 *(new, permissive)* | consumer ref drops the directory: `filepath.Base(filePath)` | 0 | 1 | **DEAD** — the guard's two same-basename files do real work |
| P4 *(new, permissive)* | BINDS `FromID: token` → `pyDIProviderRef(token)` | 0 | 1 | **DEAD** |
| P4b *(new, permissive)* | INJECTED_INTO `FromID: token` → `pyDIProviderRef(token)` | 0 | 1 | **DEAD** |

No mutant survived. The suite is genuinely tight against that file — and the defect is still there, because **it is a defect in the fix as written, not a gap in what the fix pins.** Mutation testing asks "would a change to this code be noticed"; it cannot ask "is the thing this code asserts true." Worth remembering the next time an all-DEAD table reads as sufficient evidence.

## What the review confirmed as good

The **consumer end is correct and stays**. `pyDIConsumerRef` genuinely uses `extractor.BuildOperationStructuralRef("python", filePath, fn)` — the same helper with the same three arguments as the http-endpoint handler bridge and the class→method CONTAINS edges. `lookupStructural` does return `handled=true` on a miss (`refs.go:2938`, sole caller `refs.go:2062`), so that shape cannot fall through to bare-name matching; verified in code, and it is rationale rather than a tested property, which is fine here.

The `dependency_injector` decision is not just documented, it is **observed**: mutating that field to `pyDIProviderRef(token)` is killed by two named tests, so nobody can "fix" it green.

Golden re-verified: `python-fastapi-mini` grades entities 5/5, relationships 4/4, forbidden 0.

## Two things the body omitted

**A real result, unstated.** Total relationships went 57 → 56, resolver stats `rewrote 27→29, ambiguous 2→0` — the dangling `INJECTED_INTO UNRESOLVED|get_db -> UNRESOLVED|list_things` now resolves and dedups into the YAML edge. That is end-to-end proof the consumer fix works and it belongs in the body.

**Node-identity churn, undocumented.** `addEdge`'s owner name embeds `FromID->ToID`, so every DI `SCOPE.Pattern` entity was renamed and its ComputeID changed:

```
di:INJECTED_INTO:get_db->list_things@9
di:INJECTED_INTO:SCOPE.Operation:get_db->scope:operation:method:python:app/routers/things.py:list_things@9
```

Not a correctness break — the CONTAINS fixup keys on IDs — but it churns identity for entities outside the three emit sites and nests a `scope:` ref inside an entity Name.

One minor note: the three site tests compare the emitter's output to `BuildOperationStructuralRef(...)`, so they are tautological with respect to that helper. Only the guard's `strings.Contains(ToID, path)` has independent teeth, and it does not pin the name segment. Low severity — `internal/extractor/structural_ref_test.go:21` pins the literal shape — but worth knowing.

Fix in progress on this branch. Not merging until the provider end is addressed and the collision case is pinned by a test that is red on `fda47b4a7`.

cajasmota:
Coordinator verification on `e0bc32152`. Two killing mutants of my own, neither previously run, plus the probe that found the method gap.

## Mutant 1 — the resolver hint the fix depends on

Dropped `"INJECTED_INTO"` from the component-family case at `internal/resolve/refs.go:2268`, so a bare provider ref loses the `componentKindFamily` narrowing:

```go
case "BINDS", "GRAPH_RELATES", "REGISTERS",
	"HANDLES_SIGNAL", "FEDERATES":
	return componentKindFamily
```

`go vet ./internal/resolve/ ./internal/custom/python/` → 0. `go test -count=1 ./internal/custom/python/` → **EXIT=1**, `--- FAIL: TestPyDIProviderDoesNotBindToUnrelatedSameNamedFunction_6511`. **DEAD.**

This is the one worth having. "Bare is kind-agnostic and still binds the class correctly" is the load-bearing claim of the whole approach, and it lives in a *different package*. The new test reaches it through the real resolver rather than a proxy, so the claim is observed rather than asserted. Restored by `cp`, shasum `90fcd59e3…` re-verified.

## Mutant 2 — is the column rule actually per-line?

Dropped the `(?m)` flag from `rePyModuleDef`, leaving `^` anchored to file start only:

```go
var rePyModuleDef = regexp.MustCompile(`^(?:async[ \t]+)?def[ \t]+(\w+)[ \t]*\(`)
```

`go vet` → 0. `go test -count=1` (full package) → **EXIT=1**. **DEAD**, by four tests:

```
--- FAIL: TestPyDI_FastAPIDependsCallable
--- FAIL: TestPyDI_LitestarProvideBindsAndInject
--- FAIL: TestPyDIProvider6511_LocallyDefinedCallableProviderKeepsItsKind
--- FAIL: TestPyDIProvider6511_ModuleLevelAsyncDefStillCounts
```

Two of those are pre-existing tests, which is the useful part: the column rule is not propped up only by the tests written for it. Restored by `cp`, shasum `f4a3912e6…` re-verified, tree clean.

## How the method gap was found, for the record

Not by a mutant. `rePyDef` matched `^[ \t]*def`, so I probed the gate directly with a throwaway test (since deleted):

```
class Registry:
    def AuthService(self): ...
class AuthService: ...
→ localDefs["AuthService"] = true
```

A class provider colliding with any method name in the file was prefixed again, reopening the collision the previous commit had just closed.

**That is the third defect on this branch found by asking what the code claims about the world rather than whether tests cover it** — and each time the mutant table was fully DEAD beforehand. Mutation testing answers "would a change here be noticed"; it cannot answer "is the thing this code asserts true." When a change invents a fact about the domain — a kind, a prefix, an address, a definition of *callable* — that second question needs its own check, usually a fixture built to violate the assertion, run against the new binary and the parent.

Worth noting the fix went further than asked on two counts: the node-identity churn was eliminated rather than documented (owner names are byte-identical to pre-#6511 and pinned by a test), and `TestPyDIProvider6511_ModuleLevelAsyncDefStillCounts` was added specifically as a non-vacuity guard so the new rule is pinned as a *column* rule and not an accidental ban on `async def`. The `R1` mutant confirms that guard has teeth.

Merging once the Windows leg lands. The three residual unobserved items are named in the body rather than buried, which is the right call.

Issue #6293, opened 2026-08-11
Windows: deleting a watched subdirectory under-counts the FD ledger (2 vs 3) — the failure the 40-minute hang was masking


cajasmota:
Surfaced by the new per-PR test gate (#6291) on its own PR, #6292 — the first Windows run of this package at a 15-minute timeout instead of 40, and without `-race`.

```
--- FAIL: TestRemoveOfAWatchedDirectoryReleasesAndUnmaps (10.23s)
    fdbudget_6268_test.go:370: a watched subdirectory deleted:
        ledger read 2 (held 0 of 30 polls), want a settled 3
```

Job: https://github.com/cajasmota/grafel/actions/runs/31522772370/job/93883705320

**This was invisible until now.** In the earlier full-matrix run (31510723706), `TestSkippedDirectoryPushesASubscriptionOverBudget` hung for 40 minutes and panicked, which blanked every test after it in the package — including this one. It is the "second failure masked by the hang" that #6289 explicitly could not rule out.

Not flaky: `held 0 of 30 polls` means the ledger settled at 2 and stayed there.

## Direction, not a diagnosis

The ledger reads **2 where 3 is expected** — an **under**-count on directory removal, the opposite direction from #6287's Linux over-count. Deleting a watched subdirectory releases more than it should, or the removal path releases something it never charged. An under-count is the original #6268 defect: the ledger understates what the process holds, so grafel keeps subscribing past the real ceiling and walks toward EMFILE instead of refusing.

Windows-only as far as we know — macOS is green on this test and Linux did not report it.

Also in the same run, unexplained and worth ruling in or out while here:

```
ERROR watcher: error err="CancelIo: The handle is invalid."
ERROR watcher: error err="CloseHandle: The handle is invalid."
```

Emitted during `TestEventChargesDuringTheWalkSurviveTheWalksBookkeeping` (401 dirs). A handle being closed twice, or closed after being invalidated, is plausibly the same accounting defect seen from the other side.

## Not the hang, and not the interleaved test

- The 40-minute hang did **not** recur in this run (no `panic: test timed out`), consistent with it being a `-race`-timing-dependent manifestation of the `Stop`/`Close` deadlock fixed in #6289.
- `TestInterleavedDirectoryFillsAcrossReposAreNotDoubleCharged` **also fails on Windows** here (`662 vs 660`, two quarantines, same signature as Linux) and is already fixed by #6289's detach — this issue is not about that one.

## Verify before assuming it is separate

[...]

PR #6566, opened 2026-08-24
fix(#6493): handleEvent called fsnotify.Add on the only goroutine draining Events, deadlocking the Windows backend

### LINKED ISSUE #6493: Watcher.RemoveRepo can deadlock forever against fsnotify's Windows backend, hanging the daemon
`Watcher.RemoveRepo` can deadlock forever against fsnotify's Windows backend, wedging the daemon — caught as a 15-minute CI timeout on an unrelated PR.

## The deadlock

Two goroutines, from the CI goroutine dump (run 32459537306, `test (windows-latest)`, PR #6491 — a PR touching only `internal/quality/diff.go`, so this is unrelated to its change):

```
goroutine 180 [chan send, 14 minutes]:
  fsnotify.(*readDirChangesW).Remove       backend_windows.go:162
  fsnotify.(*Watcher).Remove
  watch.(*Watcher).unwatchAll              internal/daemon/watch/watcher.go:882
  watch.(*Watcher).RemoveRepo              internal/daemon/watch/watcher.go:855
  watch.TestInterleavedDirectoryFillsAcrossReposAreNotDoubleCharged
                                           internal/daemon/watch/fdbudget_6268_test.go:1360

goroutine 181 [select, 14 minutes, locked to thread]:
  fsnotify.(*readDirChangesW).sendEvent    backend_windows.go:76
  fsnotify.(*readDirChangesW).readEvents   backend_windows.go:642
```

fsnotify's Windows backend (`readDirChangesW`) serialises `Remove` by **sending a command on a channel to the same goroutine that delivers events**. Here `readEvents` is blocked in `sendEvent` trying to hand an event to a consumer that is not draining, so it never returns to service the pending `Remove` — and `RemoveRepo` blocks forever.

**This is Windows-specific by construction.** The inotify and kqueue backends perform removal as a direct syscall on the calling goroutine, so neither can exhibit it. That is why it has never been seen on the Linux or macOS legs.

## Why it matters beyond CI

`RemoveRepo` is not test-only. Anything that unregisters a repo on Windows — daemon shutdown, repo removal, the reconcile loop dropping a repo that disappeared — can enter this state and hang the daemon goroutine that called it. The daemon is what serves MCP; a wedged `RemoveRepo` is a silent hang, not a crash, so there is no error to surface and nothing in the logs after the last `watcher: registered` line.

## Why it is intermittent, and why that is worse

It requires `readEvents` to be mid-`sendEvent` at the moment `Remove` is issued, with the event channel unattended. `test (windows-latest)` passed on the four most recently merged PRs (#6484, #6486, #6487, #6489) and failed here, on a commit that cannot have caused it. So it fires on a timing window, and will be misread as CI flake every time it does. It already was — my first reading of this failure was "unrelated package, probably flaky."

The immediately preceding log line is from a *different* test (`TestRefusalReturnsEventChargesToo`), which suggests the wedged consumer may be inherited across tests in the package rather than being specific to `TestInterleavedDirectoryFillsAcrossReposAreNotDoubleCharged`.

## What needs establishing before a fix

1. **Who is supposed to drain the event channel during `RemoveRepo`, and why isn't it?** The fix is probably at that end, not at the `Remove` call.
2. **Is there a bounded-wait option?** A `Remove` that can block forever is not safe to call from a shutdown path regardless of the drain question.
3. **Does the reconcile loop (`reconcile.go:176`) or `quarantineSweep` call `RemoveRepo` on a live daemon?** If so this is reachable in production, not only in tests. Goroutine 185 was parked in `reconcileLoop` here, so it was not the caller in this instance.
4. **Whether the same shape exists at any other `fsnotify.Remove`/`Add` call site** in the package.

Not a release blocker for the Linux/macOS path, but it is a real hang on a supported platform, and the failure mode — silent, intermittent, indistinguishable from flake — is the class this milestone has been prioritising.

cajasmota:
Closes #6493

`Watcher.loop` is the only consumer of the fsnotify Events channel, and
`handleEvent` called `subscribeDirRecursive` — and therefore `fsnotify.Add` —
synchronously on that goroutine. fsnotify's Windows backend is a
single-goroutine actor: `Add`/`Remove` are requests posted to the very goroutine
that delivers events, and its `sendEvent` is a blocking send. So the moment the
backend had an event to deliver while `loop` was inside `Add`, the two wedged
permanently, and every later caller — `RemoveRepo` → `unwatchAll` — blocked
forever posting into a dead actor. That is why the symptom always surfaced at
`RemoveRepo`; `RemoveRepo` was the victim.

The invariant this establishes: **`loop` never makes a blocking call into the
backend, so the Events channel is always being drained.**

## The change

`handleEvent` hands the directory to `subscribeOwner`, a single goroutine that
owns subscriptions discovered during event handling, over a bounded channel
(`subCh`, cap 256) written with a **non-blocking** send. A full queue costs the
subscription (counted in `droppedSubscribes`, logged at WARN, reported by
`ExtendedStats`), never the drain — a blocking send would reinstate the deadlock
exactly.

`subInFlight` counts, per root, the subscriptions handed over and not yet
finished. A count and not a set: one directory can be handed over twice (a
delete-then-recreate reports Create twice), and a set cleared by the first walk's
exit unmarks the directory while the second is still queued.

`subscribeDirRecursive` binds `w.fsAdd` once, under `w.mu`, at entry: the owner
outlives a backend restart, and the heartbeat re-points that field under the same
lock, so one walk can no longer straddle two backend generations. No backend call
is made while holding `w.mu` (#6309).

`Stop` gains one bounded wait for the owner, placed **before** the loop wait and
after `closeBackend`, so the loop is still draining while an owner parked in
`Add` unwinds. Queued-but-unstarted work is abandoned rather than drained.

## Accounting: two regressions introduced here, one pre-existing defect

Being precise about which is which, because "found latent bugs" and "broke two
things and fixed them" must not read the same.

**Regression introduced by this refactor, fixed before merge — Create-then-listing
double charge.** `fdPreCharged` suppressed *listing-then-Create* duplicates. On
the parent commit that was airtight for a structural reason: `subscribeDirRecursive`
and `chargeEventOpen` each had exactly one caller, `handleEvent`, on the single
`loop` goroutine, so the opposite order could not occur. Making the walk
concurrent created that order. `fdEventCharged` is the mirror set:
`chargeEventOpen` records a non-directory path it charged while a walk covering
it was in flight, and `settleListing` deducts those from what the listing would
charge. Reclaiming and publishing happen in one critical section, because
splitting them leaves one of the two windows open. Directories are excluded — a
subdirectory created during a walk is charged `perDir` by that walk and
`perEntry` by its event, and its own `subscribeDirRecursive` deducts that
`perEntry` unconditionally, which is the arithmetic the synchronous code already
produced.

**Regression introduced by this refactor, fixed before merge — short `dirEntries`
tally.** An entry charged by the loop after the listing read the directory is in
neither the listing count nor the tally (the tally bump is skipped while the
directory is not yet in `dirToRepo`). Same structural argument: on the parent
there was no goroutine that could charge during a walk. A short tally makes
`reconcileDir` repair a directory that needs no repair. Those entries are now
counted in the same critical section that publishes `dirToRepo`, which closes the
window.

**Pre-existing defect, genuinely found — duplicate entry Remove.** kqueue can
report one entry removal twice; `releaseEventClose` released twice for one close.
`#6293` had already built the marker mechanism for exactly this and applied it
only to directories. It now covers entries too, and the marker dies as soon as a
descriptor could exist again: `chargeEventOpen` clears it on a Create, and
`forgetReleasedEntriesLocked` clears every child of a re-listed directory. The
refactor is what made this reproduce reliably; **it is not what caused it**, and
the earlier claim that concurrent draining is why kqueue started duplicating was
an unverified story and is withdrawn — the release path was wrong either way.

## Call sites: routed vs. not

| Site | Goroutine | Routed? |
|---|---|---|
| `subscribeDirRecursive` → `fsAdd` (`watcher.go`) | **was the loop** | **Yes** — this is the bug |
| `subscribeRepo` → `fsAdd` | `AddRepo` caller / heartbeat | No |
| `unwatchAll` → `fsRemove` (`RemoveRepo`; budget-refusal unwind) | caller / `AddRepo` caller | No |
| `repairDir` → `fsAdd` / `fsRemove` (`reconcile.go`) | reconcile sweep | No |

The three unrouted sites already satisfy the invariant: none runs on the loop
goroutine and none holds `w.mu` across the call. Routing them would make
`AddRepo`, `RemoveRepo` and the sweep depend on the owner's queue and need a
reply channel — a new blocking path, and a new shutdown hazard, for no deadlock
benefit. `AddRepo` also returns the directory count it subscribed, which a
fire-and-forget hand-off cannot produce.

That leaves one ordering hazard the hand-off creates: a subscription discovered
before `RemoveRepo` now runs after it, and could re-`Add` a directory
`RemoveRepo` just unwatched. `RemoveRepo` clears `dirToRepo` under `w.mu`
*before* it touches the backend, and `subscribeDirRecursive` resolves its owning
repo on entry, so the deferred work becomes a no-op. Tested.

## Tests

The bug cannot reproduce on macOS or Linux with a real backend, so the test is
structural. `windowsActorBackend` models the Windows actor semantics — one
goroutine both delivers events with a blocking send and services `Add`/`Remove`
over a channel round-trip — reusing the existing `fsAdd`/`fsRemove` seams and
`testEvents`; **no new production seam was needed**. Event delivery has strict
priority over request servicing, which makes the pre-fix deadlock deterministic
rather than a scheduling coin-flip. All waits are bounded, so a regression is a
FAIL and not a hang.

- `TestLoopKeepsDrainingWhileSubscribingANewDirectory` — the regression. Two
  events queued together; the second arrives only if the loop kept draining
  across the subscription. Verified RED on the pre-fix code
  (`only 1 of 2 events reached the loop goroutine within 10s`), green after. It
  also asserts the directory was actually `Add`ed, so a "fix" that dropped the
  work would not pass.
- `TestDuplicateEnqueueKeepsTheDirectoryMarkedUntilEveryWalkIsDone` — the
  refcount. Red on `a214a784f`
  (`subInFlight lost .../d while a subscription walk over it was still running`).
- `TestFullSubscribeQueueDropsTheSubtreeAndKeepsTheLoopDraining` — with a 1-slot
  queue and the owner parked in `Add`, exactly one of three directories is
  dropped, all four events still reach the loop, and the drop is visible through
  `ExtendedStats`.
- `TestOnlyEntriesInsideTheWalkedSubtreeCountAsAlreadyCharged` — the ancestry
  walk, with a positive control so it cannot pass vacuously.
- `TestASubdirectoryCreatedDuringAWalkIsNotCountedAsAnEntryOfItsParent` — the
  directory exclusion, observed through `dirEntries`.
- `TestSubscribeQueuedBeforeRemoveRepoDoesNotResurrectTheWatch` — the ordering
  guard above.
- `TestStopRetiresTheSubscribeOwnerWithQueuedWorkOutstanding` — `Stop` returns
  with five subscriptions still queued and completes none of them.

Two existing tests needed a wait rather than a rewrite, because subscription is
now asynchronous: `drainLedger` treats "a subtree is still queued" as not
settled, and the #6293 re-subscribe test waits for the owner to publish. Both are
the same assertion as before, taken at the right moment.

### One test was racy, and its positive control is what caught it

`test (ubuntu-latest)` failed on `b49835552` — on the **positive control** of
`TestOnlyEntriesInsideTheWalkedSubtreeCountAsAlreadyCharged`, not on its real
assertion. The cause was a test-ordering defect, not a platform difference in
the fix: `releaseForever` was called *before* `enqueueSubscribe`, so the gate
never held, the walk ran to completion racing the test goroutine, and the
subtree was no longer in flight by the time the Creates were driven. It held on
kqueue and lost on inotify. Without the control it would have passed **vacuously
on Linux**.

Nothing in the in-flight suppression is platform-dependent — it is map and mutex
logic — and these tests synthesise their events by calling `handleEvent`
directly rather than taking them from a backend. The fix for this was confined
to the test file; no production line changed.

Every stepped test now parks its walk by channel blocking rather than by
scheduling luck, and calls `requireInFlight` as an explicit premise check so a
finished walk fails loudly instead of quietly making the real assertion vacuous.
The release gate also became a token buffer: a pump goroutine could not work,
because its `t.Cleanup` is registered after `windowsActorWatcher`'s and cleanups
run LIFO, so the pump was torn down *before* the `Stop` that needed it — which
is where the `subscribe owner did not retire` line in the CI log came from.

### Mutants scored (`go vet` clean first, `go test -count=1`, whole package, no `-run`)

Re-scored on the deterministic harness. Every kill below now rests on an
assertion whose premise is enforced by channel blocking, so it does not depend
on the scheduler — but all runs are **macOS/kqueue**; Linux and Windows are
CI's word, not mine.

| Mutant | Result | Killed by |
|---|---|---|
| `handleEvent` calls `subscribeDirRecursive` inline again (permissive: blocking call back on `loop`) | **DEAD** | 3 tests |
| `enqueueSubscribe` uses a blocking send (permissive: loop blocks on a full queue) | **DEAD** | queue-full test |
| drop the `fdEventCharged` reclaim in `settleListing` | **DEAD** | 2 ledger tests |
| drop the entry release marker in `releaseEventClose` | **DEAD** | interleaved-fill test |
| `releaseInFlight` deletes the key instead of decrementing | **DEAD** | duplicate-enqueue test |
| record directories in `fdEventCharged` too | **DEAD** (0.02s) | `dirEntries` test |
| `underSubscribeInFlightLocked` returns `len(subInFlight) > 0` | **DEAD** (0.02s, was 10.01s) | ancestry test |
| `enqueueSubscribe` drops `releaseInFlight` on the refusal path (permissive: refcount leak) | **DEAD** (0.00s) | queue-full test's new mark assertion |

The last three previously depended, wholly or partly, on the racy test. The
refusal-path mutant is now killed by a fresh assertion in the queue-full test —
that a refused hand-off gives back the in-flight mark it took, with the queued
directory checked alongside so it cannot pass vacuously — rather than by a
premise that happened to hold.

## Not observed by any test

- **The Windows behaviour itself.** Every assertion here is against the fake
  actor. CI's `test (windows-latest)` leg exercises the real backend; nothing in
  this PR pins it.
- **Every mutant score is a macOS result.** The kills are now deterministic by
  construction rather than by timing, but I cannot run Linux or Windows locally,
  so "DEAD on Linux" is an inference from the harness, not a measurement. The
  ubuntu leg on this branch is the only evidence that settles it.
- **Cross-test interference from an abandoned owner: checked, not real.** `subCh`
  is a per-`Watcher` field, so an owner abandoned by one test's bounded `Stop`
  can only drain its own watcher's queue and call its own seams. The only shared
  mutable state is two package vars (`subscribeQueueCap`, `watcherStopTimeout`),
  both restored by `t.Cleanup`, and the process-wide fd ledger, which every test
  here bypasses with an explicit `FDBudget`. The `timeout=2s` in the CI log can
  only come from `TestStopRetiresTheSubscribeOwnerWithQueuedWorkOutstanding`,
  where parking the owner across `Stop` is the point of the test; slog writes
  unbuffered to stderr while the framework buffers per-test output, so its
  apparent adjacency to another test's failure is an artefact.
- **Recovery from a dropped subtree.** The counter is now exported and the drop
  is logged at WARN, but nothing rediscovers the lost tree: the reconcile sweep
  only visits directories already in `dirToRepo`, and repairs per entry, so a
  dropped ROOT stays unwatched until its repo is re-registered. Recovery is
  deliberately **out of scope** for this PR.
- **The `fdPreChargedCap` wholesale reset** on `fdEventCharged`, and markers
  whose walk finished before the Create arrived — the same bounded-leak
  treatment `fdPreCharged` already has, and equally unobserved.
- **A `RemoveRepo` that lands mid-walk.** The guard is evaluated once, on entry;
  a repo removed while the walk is already descending can still see a few
  `Add`s. That race predates this PR (the loop had it too) and no test pins it.
- **The `subInFlight` prefix walk cost.** Guarded by a `len(...) == 0` fast
  path; no benchmark measures it.

## Known, pre-existing, deliberately NOT fixed here

`subscribeRepo` runs off the loop goroutine and publishes `dirEntries[p]` after
its `fsAdd` with no `settleListing` equivalent, so the shape of the second
regression above exists there independently of this change. It is pre-existing
and out of scope; fixing it belongs in its own issue.

### DIFF
--- a/internal/daemon/watch/fdbudget_6268_test.go
+++ b/internal/daemon/watch/fdbudget_6268_test.go
@@ -245,6 +245,17 @@ func drainLedger(t *testing.T, w *Watcher, accept func(int) bool) (int, bool) {
 	last, longestGap := -1, time.Duration(0)
 	runStart := time.Now()
 	for time.Now().Before(deadline) {
+		// #6493: subscribing a new subtree no longer happens on the event
+		// goroutine, so the ledger can sit at a plateau simply because the
+		// subscribe owner has not reached the queued directory yet. That is a
+		// quiet QUEUE, not a settled ledger, and reading it as one hands back a
+		// figure two directories short. Restart the hold — and with last = -1,
+		// so this wait cannot inflate longestGap either.
+		if subscribesPending(w) {
+			last, runStart = -1, time.Now()
+			time.Sleep(ledgerPollEvery)
+			continue
+		}
 		used, _ := w.fdb.snapshot()
 		switch {
 		case accept != nil && !accept(used):
--- a/internal/daemon/watch/fdbudget_6293_test.go
+++ b/internal/daemon/watch/fdbudget_6293_test.go
@@ -144,6 +144,9 @@ func TestARecreatedDirectoryIsReleasableAgain(t *testing.T) {
 	// Nothing on disk changes — src is still there, only the ledger was told it
 	// went away — so no genuine event can arrive alongside and be counted too.
 	w.handleEvent(fsnotify.Event{Name: dir, Op: fsnotify.Create})
+	// The re-subscribe is performed by the subscribe owner, not by handleEvent
+	// (#6493), so it completes after handleEvent returns rather than inside it.
+	awaitSubscribed(t, w, dir)
 	w.mu.Lock()
 	_, watched := w.dirToRepo[dir]
 	w.mu.Unlock()
--- a/internal/daemon/watch/loop_backend_6493_test.go
+++ b/internal/daemon/watch/loop_backend_6493_test.go
@@ -1,0 +1,529 @@
+package watch
+
+import (
+	"errors"
+	"os"
+	"path/filepath"
+	"sync"
+	"sync/atomic"
+	"testing"
+	"time"
+
+	"github.com/fsnotify/fsnotify"
+)
+
+// ---------------------------------------------------------------------------
+// #6493 — the loop goroutine must never make a blocking call into the backend.
+//
+// fsnotify's Windows backend is a single-goroutine actor: Add and Remove are
+// requests posted to the very goroutine that delivers events
+// (`w.input <- in; return <-in.reply`, backend_windows.go:141-146, :162-167),
+// and that same goroutine is the only sender on Events, via a sendEvent that
+// blocks until grafel receives (:69-91).
+//
+// Watcher.loop is the ONLY consumer of Events. handleEvent used to call
+// subscribeDirRecursive -> fsAdd synchronously, ON the loop goroutine. So the
+// instant the backend had a second event to deliver while loop was inside Add,
+// the two wedged against each other permanently: the backend cannot service
+// the Add until someone receives the event, and the only receiver is inside
+// the Add. Every later caller — RemoveRepo -> unwatchAll — then blocks forever
+// posting into a dead actor, which is where the symptom always surfaced.
+//
+// kqueue and inotify do Add/Remove as a direct syscall on the calling
+// goroutine, so the cycle is unreachable off Windows with a real backend. The
+// fake below models the Windows actor semantics exactly, which makes the
+// property — "loop keeps draining while a subscription is in flight" — a
+// deterministic assertion on any host.
+// ---------------------------------------------------------------------------
+
+// errActorClosed stands in for fsnotify.ErrClosed: what a real backend returns
+// from Add/Remove once it has been closed. The fake returns it rather than
+// parking forever so a REGRESSION of this bug is a clean FAIL instead of a
+// wedged test binary.
+var errActorClosed = errors.New("windowsActor: closed")
+
+type actorReq struct {
+	op    string
+	path  string
+	reply chan error
+}
+
+// windowsActorBackend is a stand-in for fsnotify's Windows backend, faithful in
+// the one dimension this bug lives in: ONE goroutine both delivers events
+// (blocking send) and services Add/Remove (channel round-trip). Anything that
+// takes the receiver away from the Events channel therefore stops Add and
+// Remove from ever completing.
+type windowsActorBackend struct {
+	ev     chan fsnotify.Event
+	errs   chan error
+	reqs   chan actorReq
+	outbox chan fsnotify.Event // events the test hands the actor to deliver
+
+	delivered chan fsnotify.Event // one send per event whose blocking send RETURNED
+	stop      chan struct{}
+	stopOnce  sync.Once
+	done      chan struct{}
+
+	mu    sync.Mutex
+	calls []string
+}
+
+func (b *windowsActorBackend) record(op, path string) {
+	b.mu.Lock()
+	b.calls = append(b.calls, op+" "+path)
+	b.mu.Unlock()
+}
+
+func (b *windowsActorBackend) callsSnapshot() []string {
+	b.mu.Lock()
+	defer b.mu.Unlock()
+	return append([]string(nil), b.calls...)
+}
+
+// request is the Add/Remove seam: post to the actor, wait for its reply. This
+// is the shape that makes the backend un-callable from a goroutine the actor
+// needs.
+func (b *windowsActorBackend) request(op, path string) error {
+	r := actorReq{op: op, path: path, reply: make(chan error, 1)}
+	select {
+	case b.reqs <- r:
+	case <-b.stop:
+		return errActorClosed
+	}
+	select {
+	case err := <-r.reply:
+		return err
+	case <-b.stop:
+		return errActorClosed
+	}
+}
+
+// run is the actor. Event delivery has strict priority over servicing
+// requests, which is what makes the pre-fix deadlock deterministic rather than
+// a scheduling coin-flip: with an event already queued, the actor is inside
+// its blocking send by the time the loop goroutine posts an Add.
+func (b *windowsActorBackend) run() {
+	defer close(b.done)
+	for {
+		select {
+		case <-b.stop:
+			return
+		default:
+		}
+		select {
+		case e := <-b.outbox:
+			select {
+			case b.ev <- e:
+			case <-b.stop:
+				return
+			}
+			select {
+			case b.delivered <- e:
+			case <-b.stop:
+				return
+			}
+			continue
+		default:
+		}
+		select {
+		case <-b.stop:
+			return
+		case e := <-b.outbox:
+			select {
+			case b.ev <- e:
+			case <-b.stop:
+				return
+			}
+			select {
+			case b.delivered <- e:
+			case <-b.stop:
+				return
+			}
+		case r := <-b.reqs:
+			b.record(r.op, r.path)
+			r.reply <- nil
+		}
+	}
+}
+
+func (b *windowsActorBackend) shutdown() {
+	b.stopOnce.Do(func() { close(b.stop) })
+	<-b.done
+}
+
+// windowsActorWatcher builds a Watcher whose loop drains the actor's Events and
+// whose Add/Remove seams post to that same actor.
+//
+// The REAL backend is still constructed by NewWatcherConfig and still gets a
+// discard sink for its own channels — withholding events from the loop is not
+// the same as leaving the backend unread, and #6380 is the bill for conflating
+// them (see withheldEventsWatcher).
+func windowsActorWatcher(t *testing.T, cfg Config) (*Watcher, *windowsActorBackend) {
+	t.Helper()
+	b := &windowsActorBackend{
+		ev:        make(chan fsnotify.Event),
+		errs:      make(chan error),
+		reqs:      make(chan actorReq),
+		outbox:    make(chan fsnotify.Event, 8),
+		delivered: make(chan fsnotify.Event, 8),
+		stop:      make(chan struct{}),
+		done:      make(chan struct{}),
+	}
+	go b.run()
+
+	cfg.testEvents, cfg.testErrors = b.ev, b.errs
+	// The automatic sweep would call the same seams on its own schedule and
+	// race the sequence this test drives by hand.
+	cfg.reconcileInterval = -1
+	cfg.disableQuarantine = true
+	w := newBudgetedWatcherCfgNoCleanup(t, cfg)
+
+	w.mu.Lock()
+	fw := w.fs
+	w.fsAdd = func(p string) error { return b.request("Add", p) }
+	w.fsRemove = func(p string) error { return b.request("Remove", p) }
+	w.mu.Unlock()
+	drainedReal := discardBackendChannels(fw.Events, fw.Errors)
+
+	realClose := w.closeBackend
+	w.closeBackend = func() error {
+		err := realClose()
+		select {
+		case <-drainedReal:
+		case <-time.After(5 * time.Second):
+			t.Errorf("the backend discard sink outlived Close by 5s")
+		}
+		// Retire the actor BEFORE closing the channels it sends on: a real
+		// backend closes its Events only from inside its own I/O goroutine, so
+		// closing them out from under a live actor is a harness bug, not a
+		// property of the code under test.
+		b.shutdown()
+		close(b.ev)
+		close(b.errs)
+		return err
+	}
+	t.Cleanup(func() {
+		w.Stop()
+		b.shutdown()
+	})
+	return w, b
+}
+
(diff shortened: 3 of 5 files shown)


cajasmota:
Independent adversarial review — **REQUEST-CHANGES.** The deadlock fix proper is correct and would be approved on its own; four things sit on top of it, one of them a defect reproduced on this head.

## The blocker: `subInFlight` has no refcount

It is a set keyed by root, and `subscribeDirRecursive`'s `defer` deletes the key unconditionally. Two Create reports for the **same** directory — delete-then-recreate during a checkout, the pattern `fdbudget_6293_test.go` already models — enqueue it twice. Walk #1's defer clears the mark while walk #2 is still to run.

From that point `chargeEventOpen` records no `fdEventCharged` marker for that subtree, **reopening the exact double-charge window `settleListing` was added to close**, and `subscribesPending` — which `drainLedger` now depends on — reports "settled" with a subscribe outstanding.

Reproduced deterministically on unmodified `a214a784f`, `go vet` 0, full package `go test -count=1` exit 1:

```
--- FAIL: TestProbeDuplicateEnqueueClearsInFlightWhileAWalkIsStillRunning
    subInFlight lost .../d while a subscription walk over it was still running
```

## The three "accounting defects" — two are this PR's own regressions

The body says the hand-off "exposed three **real** accounting defects that existing tests caught". Checked against the parent `cd06ab4aa`:

| defect | verdict |
|---|---|
| Create-then-listing double charge | **INTRODUCED-THEN-PATCHED** |
| short `dirEntries` tally | **INTRODUCED-THEN-PATCHED** |
| duplicate entry `Remove` | **PRE-EXISTING**, proven |

On the parent, `subscribeDirRecursive` has exactly one caller (`handleEvent`, `watcher.go:1387`) and `chargeEventOpen` exactly one (`handleEvent`, `:1430`), both on the single `loop` goroutine — so the concurrency window the first two defects require is **structurally absent**. The PR's own in-code comments say so ("Before #6493 only one direction could occur"), which contradicts its summary line.

The third is genuine, and proven rather than argued: a probe on the parent fails (ledger 4 after two `Remove` reports of one closed entry handle, want 5) and passes on this head. `releasedDirLocked` is only ever armed for watched directories, so any backend reporting an entry removal twice double-releases. The PR's causal story — that kqueue only started duplicating because the loop now drains — is unverified and beside the point; the code was wrong either way.

This is still a good outcome. Two regressions were caught by existing tests and fixed before merge, which is the system working. But **"found latent bugs" and "broke two things and fixed them" must not read the same in a PR body**, because the second one is evidence about the refactor's blast radius and the first is not.

## Two new permissive mutants survived

Both `go vet` 0, `go test -count=1` exit 0, zero failures:

- **N1** — record directories in `fdEventCharged` too. The body asserts directories must be excluded or "recording it here would deduct it twice". Nothing observes that claim.
- **N2** — replace the whole ancestry walk in `underSubscribeInFlightLocked` with `return len(w.subInFlight) > 0`. Entirely unpinned.

Pin both, or record each as equivalent-under-the-suite with the reason. A survivor gets reported, not rounded down — that rule is why the last three defects on other branches were caught.

## Silent subtree loss

A dropped subscribe loses its subtree **permanently**: the reconcile sweep only visits directories already in `dirToRepo` and repairs per entry, so a dropped **root** is never rediscovered. `droppedSubscribes` is not in `Stats`/`ExtendedStats`, which the body admits — so in production this is invisible. Export it at minimum, and say plainly if recovery is out of scope.

## Verified, for the record

Everything the fix claims about the deadlock itself holds. `handleEvent` reaches only `chargeEventOpen`, `releaseEventClose`, `repoFor`, `quarantine.Observe`, `enqueueSubscribe` and `recordAndArm` — none touch `fsAdd`/`fsRemove`; `backendClosed` only does a non-blocking send on `restartCh`, and it is the *heartbeat* goroutine, not `loop`, that runs `restartBackend` → `subscribeRepo`. The three unrouted call sites are each unreachable from `loop` as claimed. #6309 is not reintroduced: `fsAdd` is bound under `w.mu` and called with the lock released, and `enqueueSubscribe` releases before its send. The fake models the actor faithfully — one goroutine, blocking `b.ev <- e`, unbuffered `reqs` round-trip, strict event priority — and the pre-fix failure is for the right reason. Shutdown cannot hang: the owner exits on `stopCh` and abandons the queue, and `subCh` is never closed.

One adjacent pre-existing issue, **out of scope here** but worth a line in the body: `subscribeRepo` runs off-loop and publishes `dirEntries[p]` after its `fsAdd` with no `settleListing` equivalent — the defect-2 shape already exists there.

Fixes in progress on this branch.

PR #6567, opened 2026-08-24
fix(#6543): the field-extraction metric could not see SQL, and counted its columns as failures

### LINKED ISSUE #6543: the field-extraction metric still cannot see SQL: SCOPE.Datastore is unadmitted, and admitting it alone would guarantee 100% because columns are not counted as fields
Found by the independent review of PR #6541, which fixed the same class of defect for `SCOPE.Component`. **SQL is the remaining field-bearing kind the field-extraction metric still cannot see** — and unlike #6536, fixing it needs the numerator widened too, which is why it is filed separately rather than folded in.

## The gap

`classLikeKindTails` (`internal/feedback/report.go`) admits `class`, `struct`, `model`, `schema` — and, after #6536, `component`. Of the 20 `SCOPE.*` kind strings in the vocabulary, the field-bearing one still unadmitted is **`SCOPE.Datastore`** (tail `datastore`, **13 emit sites**).

`internal/extractors/sql/sql.go:249` emits `SCOPE.Datastore` / `table` with `CONTAINS(contained_kind=column)` children. cobol, jcl and erlang emit the same kind.

A SQL table is exactly what this metric is for — a class-like container with declared members — and **it has never been sampled**, for the same reason C# classes never were: the tail list and the extractor disagreed, and nothing compared them.

## Why this is not a one-line addition

Admitting the tail is not sufficient. `report.go:340` counts field children as `entitySubtype[rel.ToID] == "field"`, and **SQL children are subtype `"column"`** (`internal/extractors/sql/sql.go:358`, `:1021`).

So adding `datastore` alone would enrol every table into the denominator while the numerator stayed blind to its columns — producing a **guaranteed 100% zero-field rate for SQL**. That is precisely the defect #6536 was filed for, re-created in a new place by an incomplete fix.

Both halves have to move together, or neither should.

## What a fix needs to decide

1. **Admit `datastore` to the tail list**, and audit the other three emitters (cobol, jcl, erlang) rather than assuming they match SQL's shape.
2. **Widen the field-child predicate** to recognise `column` alongside `field` — or generalise it so a container declares which child subtype counts as its members, which is the version that stops this recurring for the next kind.
3. **Check the remaining kinds** for any other field-bearing tail still unadmitted. The population was established once by this review; it should be established by a test, not by a review.

## Observability

Go level, `internal/feedback` — its tests build a `Report` literal directly, so this needs no fixture and no corpus.

**Killing mutant:** remove `datastore` again, or narrow the child predicate back to `field` only. A test asserting a SQL table with column children reports a **non-**100% zero-field rate must go red in both cases. Today neither is observed.

Refs #6535, #6536, #6541

cajasmota:
Closes #6543

> Three rounds. Round 2 fixed the gate granularity; round 3 closed the empty-emit-site hole. See "Review rounds" below.

## What was wrong

`classLikeKindTails` did not admit the `datastore` tail, so SQL tables — emitted as `SCOPE.Datastore` with `CONTAINS(contained_kind=column)` children (`internal/extractors/sql/sql.go:249`) — were never in the field-extraction denominator.

The RED run surfaced something the issue did not predict. On a three-column `CREATE TABLE`, the metric reported `ClassTotal = 3` at `ZeroFieldsPct = 100`. Those three were the **columns**, not the table: a column is `SCOPE.Schema`/`column`, whose tail `schema` was already admitted, and `column` was absent from `nonClassSubtypes`. The only SQL entities the metric ever sampled were member leaves that structurally cannot own members — a guaranteed 100%, with the container invisible.

## The three moving parts

1. **`classLikeKindTails` admits `datastore`.**
2. **`memberChildSubtypes` replaces the `entitySubtype[rel.ToID] == "field"` literal** at the child-counting site. A container's declared members are `field` **or** `column`. `column` is added to `nonClassSubtypes` in the same change, restoring the leaf/container symmetry `field` already had.
3. **`datastoreMemberBearingKinds`** — an allowlist of **emit sites**, keyed `"<language>/<subtype>"`, currently `{"sql/table"}`.

Result on the fixture: `ClassTotal = 1`, `ZeroFieldsPct = 0`.

## The gate rule, and why the unit is the emit site

`SCOPE.Datastore` is not one population, and neither the kind, the language, nor the subtype is a unit you can gate on without re-creating #6536 somewhere new.

- **Not the kind** — 13 non-test emit sites, one of which owns members.
- **Not the language.** `sql` is *seven* emit sites: `table` (`sql.go:249`), `view` (`:419`), `index` (`:444`), `function`/`trigger_function` (`:527`), `procedure` (`:585`), `trigger` (`:634`), `dbt_source` (`:695`). Only `table` emits column children, and none of the other six is in `nonClassSubtypes`. Killed by **M7**.
- **Not the subtype.** Erlang's ETS/Mnesia datastores are `Subtype: d.engine + "_table"` (`erlang/otp_deepen.go:407`) — `ets_table`, `mnesia_table`. A bare `{"table"}` allowlist misses them today by luck and admits any future `duckdb/table`. Killed by **M8**.

Both dimensions are load-bearing and each is pinned by a mutant only the other kills.

**Allowlist, not denylist**, so an un-enumerated emit site is excluded until someone checks it. Guessing wrong that way costs an unmeasured population; guessing wrong the other way puts a guaranteed-failure population in the denominator, which is what #6535/#6536/#6543 are about. **An unset dimension is an unknown emit site, not a wildcard**, and is excluded for exactly the same reason. Both halves of that contract are now observed (M6, M9).

### Round 2 — gate granularity

Reproduced on the first commit with the extended fixture: `ClassTotal = 6`, `ZeroFieldsPct = 83.3` on a schema whose single table is fully columned. Indexes alone outnumber tables in most real schema files, so the reported SQL rate would have been dominated by the population that cannot pass. `datastoreMemberBearingLanguages` → `datastoreMemberBearingKinds` + `datastoreEmitSiteKey`. `sqlSource6543` now exercises **every** `SCOPE.Datastore` subtype the extractor emits — a table-only fixture is what hid this, and `TestSQLTableColumnsMeasuredNonVacuously_6543` now `t.Fatal`s if the fixture ever loses its non-passing siblings.

**M6** (allowlist → denylist of the three known languages) had SURVIVED — the stated design argument had no observation. `TestUnknownDatastoreEmitSitesAreExcluded_6543` now asserts `duckdb/table`, `prisma/model`, `sql/materialized_view` and erlang's real `ets_table`/`mnesia_table` are all excluded.

**Cobol citation corrected.** `cobol/extractor.go:748` data items are `SCOPE.Schema`/`field` whose group-item parents carry `Subtype "field"` too, so `nonClassSubtypes` excludes them as leaves — they pass nothing and were the wrong counter-example. The genuinely member-bearing cobol population is **IMS segments**: `kindIMSSegment = "SCOPE.Schema"` / `ims-segment` (`ims.go:295`) with `Subtype "field"` children (`ims.go:490`, edge at `:498-505`), admitted by the `schema` tail and untouched by this gate.

**Fixtures.** The erlang arm of `TestNonColumnBearingDatastoresExcluded_6543` passed with the gate deleted (already covered by `nonFieldBearingLanguages`) — moved to the un-enumerated-emit-site test where it discriminates; the cobol arms now use real emitted subtypes. `TestColumnlessTableStillCountsAsZeroField_6543` no longer hand-writes a `SCOPE.Operation`/`trigger` that no extractor emits.

### Round 3 — the empty emit site

**M9** (`subtype != "" &&` short-circuits the gate) had SURVIVED. `datastoreEmitSiteKey("", "")` is `"/"`, absent from the map, so an unset-subtype datastore was *already* excluded — the safe direction, unobserved. The un-enumerated-site test listed unknown **named** sites and never the empty one, so a gate that skips itself on an empty subtype passed the whole package. An extractor that forgets to set `Subtype`, or any path that drops it, lands a columnless `SCOPE.Datastore` in the denominator.

Three cases added, expectation kept as proposed — all excluded:

```
isFieldExtractionCandidate("SCOPE.Datastore", "",      "sql") == false
isFieldExtractionCandidate("SCOPE.Datastore", "table", "")    == false
isFieldExtractionCandidate("SCOPE.Datastore", "",      "")    == false
```

New `TestDatastoreMemberBearingKindsKeysAreWellFormed_6543` pins the key *shape*: every key must be `"<language>/<subtype>"` with both halves non-empty and lower-case, and must actually admit the emit site it names. A key `datastoreEmitSiteKey` can never produce reads as an admitted emit site while admitting nothing — a blind spot that looks like coverage. It checks only addressability by the lookup that reads it; it does **not** compare keys against the extractor tree.

## Replacement of the #6536 guard

The `SCOPE.Datastore` arm of `TestClassLikeKindTailsConstrainedFromAbove_6536` asserted the kind was *not* class-like, because "admitting it puts a guaranteed-zero-field population back in the denominator". Right about the danger, wrong about the kind: excluding it wholesale traded a false-failure population for an unmeasured one.

The arm is replaced in place by a comment stating that justification and naming the five tests that assert strictly more. Its precondition is discharged, not deleted — and the narrowing goes one level *further* than the arm did: not merely "not this kind", but "of the six SQL datastore subtypes, only `table`".

## Tests

New file `internal/feedback/report_datastore_column_6543_test.go`. Kind and subtype strings come from the **real SQL extractor**, never from literals. Every test carries a premise check that `t.Fatal`s rather than passing vacuously.

- `TestSQLTableColumnsMeasuredNonVacuously_6543` — the killing test.
- `TestOnlyTableSubtypeIsMemberBearing_6543` — gate granularity, every emitted subtype checked individually.
- `TestUnknownDatastoreEmitSitesAreExcluded_6543` — the allowlist choice, unknown named sites, the empty/unset sites, and case folding in both dimensions.
- `TestDatastoreMemberBearingKindsKeysAreWellFormed_6543` — every map key is reachable by the lookup that reads it.
- `TestColumnlessTableStillCountsAsZeroField_6543` — a non-member child does not make a container pass.
- `TestNonColumnBearingDatastoresExcluded_6543` — jcl/cobol datastores do not move `ClassTotal`.

`gofmt -l` clean, `go vet ./internal/feedback/` = 0, `go test -count=1 ./internal/feedback/` = 0.

## Mutants scored

`go vet` first on each; each restored by `cp` from a byte-level backup with `shasum` re-verified.

| # | Mutation | Direction | Result | Killed by |
|---|---|---|---|---|
| M1 | drop the emit-site gate entirely | permissive | **DEAD** | 4 tests |
| M2 | numerator back to the `"field"` literal | narrowing | **DEAD** | 1 |
| M3 | remove `datastore` from `classLikeKindTails` | narrowing | **DEAD** | 6 |
| M4 | remove `column` from `nonClassSubtypes` | permissive | **DEAD** | 1 |
| M5 | count **any** structural child as a member | permissive | **DEAD** | 1 |
| M6 | allowlist → denylist of the known languages | permissive | **DEAD** | 3 |
| M7 | gate keyed on **language** only | permissive | **DEAD** | 3 |
| M8 | gate keyed on **subtype** only | permissive | **DEAD** | 1 |
| M9 | empty subtype bypasses the gate | permissive | **DEAD** | 1 |
| M10 | unreachable key `"duckdb"` in the map | — | **DEAD** | 1 |
| M11 | upper-case key `"SQL/VIEW"` in the map | — | **DEAD** | 1 |

M5 survived the first pass of round 1, M6 survived round 1, M9 survived round 2. All three are fixed by strengthening tests, never by weakening the mutant.

## Not observed by any test — stated explicitly

- **That `{"sql/table"}` is the complete set of member-bearing emit sites.** The tests pin every emit site that exists today, enumerated by hand from the 13 non-test sites; the un-enumerated-site test makes the conservative default explicit and the well-formedness test pins the key shape — but nothing compares the map against the extractor tree automatically. A *new* extractor emitting a member-bearing `SCOPE.Datastore` is silently unmeasured until someone adds it. That generalisable guard is the issue's third bullet and is deliberately out of scope.
- **That `{field, column}` is the complete member-subtype set.** Same shape: hand-maintained, checked against today's extractors, not derived from them. The `contained_kind` property the CONTAINS edge already carries is the obvious raw material and is not used here.
- **That cobol IMS segments actually pass the metric.** The comment cites `ims.go:295`/`:490`/`:498-505` as the genuinely member-bearing cobol population; that claim is verified by reading, not by a test in this package.

### DIFF
--- a/internal/feedback/report.go
+++ b/internal/feedback/report.go
@@ -336,7 +336,12 @@ func Generate(_ context.Context, docs []*graph.Document, opts Opts) (*Report, er
 				// children so the field-extraction metric reflects the graph
 				// instead of reading a property the dominant extractors never
 				// write.
-				if entitySubtype[rel.ToID] == "field" {
+				//
+				// "field" is not the only name a container gives its
+				// declared members: a SQL table's members are Subtype
+				// "column" (#6543). Match the MEMBER-subtype set rather
+				// than the single literal — see memberChildSubtypes.
+				if memberChildSubtypes[entitySubtype[rel.ToID]] {
 					fieldChildCount[rel.FromID]++
 				}
 			} else {
@@ -596,12 +601,73 @@ func kindTail(kind string) string {
 // fields, so the reported rate was a guaranteed 100% rather than a measurement
 // (#6536, surfaced by #6535). Any change to this set must be checked against
 // the kinds the extractors actually emit, not against these literals.
+//
+// "datastore" is here because every SQL table is SCOPE.Datastore
+// (internal/extractors/sql/sql.go:249) with CONTAINS(contained_kind=column)
+// children — a class-like container with declared members, and exactly what
+// this metric is for, which had never been sampled (#6543). Admitting the tail
+// is only half the fix: the members are Subtype "column", so the numerator has
+// to see them (memberChildSubtypes) and the non-SQL emitters of the same kind
+// have to be exempted (datastoreMemberBearingLanguages), or the tail merely
+// re-creates #6536 for a new population.
 var classLikeKindTails = map[string]bool{
 	"class":     true,
 	"struct":    true,
 	"model":     true,
 	"schema":    true,
 	"component": true,
+	"datastore": true,
+}
+
+// memberChildSubtypes are the child subtypes that count as a container's
+// DECLARED MEMBERS — the numerator of the field-extraction metric (#6543).
+//
+// The metric asks "does this container's members show up in the graph", and
+// different extractors give the same relationship different names. Matching
+// the single literal "field" meant SQL's members were invisible: a table's
+// children are Subtype "column" (internal/extractors/sql/sql.go:358, :1021),
+// declared on the CONTAINS edge itself as contained_kind=column (sql.go:240-242).
+//
+// Every subtype here must also appear in nonClassSubtypes below: a member is a
+// LEAF, and a leaf that counts as its parent's member must never be counted as
+// a container in its own right, or it enters the denominator as a permanent
+// zero-field failure. That symmetry is what "field" already had and what
+// "column" was missing — before #6543 the three columns of a three-column
+// table were themselves 3 of the 3 entities in the denominator.
+var memberChildSubtypes = map[string]bool{
+	"field":  true,
+	"column": true,
+}
+
+// datastoreMemberBearingLanguages are the languages whose SCOPE.Datastore
+// entities are real member-bearing containers, and so belong in the
+// field-extraction denominator (#6543).
+//
+// SCOPE.Datastore has 13 non-test emit sites and only the SQL ones are
+// container-shaped. The other three emitters produce datastores with no member
+// children anywhere in their output:
+//
+//   - jcl — a DD dataset (extractor.go:664, :779) is a file reference; the
+//     jcl extractor emits no Subtype "field" or "column" at all.
+//   - cobol — an IMS database / message queue (ims.go:170) and a CICS queue
+//     or file resource (depth.go:701, :864) are external resources, not
+//     declarations. cobol DOES emit Subtype "field" (extractor.go:748,
+//     ims.go:491) for its data items, which is why the exemption is keyed on
+//     the KIND-plus-language rather than on the language wholesale: a blanket
+//     cobol exemption would drop its genuine field-bearing records.
+//   - erlang — an ETS table (otp_deepen.go:407) is a runtime table with no
+//     declared columns. (erlang is already exempt wholesale via
+//     nonFieldBearingLanguages; it is listed here so the population is
+//     recorded in one place and stays excluded if that ever changes.)
+//
+// Stated as an allowlist rather than a denylist deliberately: a new extractor
+// emitting SCOPE.Datastore is excluded until someone checks that its entities
+// actually own members. The failure mode of guessing wrong in that direction
+// is an unmeasured population; guessing wrong the other way puts a
+// guaranteed-zero-field population in the denominator, which is the defect
+// this whole line of issues (#6535, #6536, #6543) is about.
+var datastoreMemberBearingLanguages = map[string]bool{
+	"sql": true,
 }
 
 // nonClassSubtypes are the subtypes that carry a class-like KIND but are not
@@ -637,8 +703,14 @@ var classLikeKindTails = map[string]bool{
 //
 // A metric whose denominator contains populations that cannot pass reports
 // noise, not coverage.
+//   - "column" — the SQL analogue of "field": a member LEAF, not a container.
+//     Its kind tail is "schema" (internal/extractors/sql/sql.go:358), which
+//     this set already admits, so before #6543 every column was in the
+//     denominator as a permanent zero-field failure — a three-column table
+//     contributed three guaranteed failures and zero passes.
 var nonClassSubtypes = map[string]bool{
 	"field":    true,
+	"column":   true,
 	"enum":     true,
 	"const":    true,
 	"file":     true,
@@ -712,6 +784,11 @@ func isFieldExtractionCandidate(kind, subtype, language string) bool {
 		return false
 	}
 	if subtype == "interface" && !interfaceSubtypeFieldBearingLanguages[strings.ToLower(language)] {
+		return false
+	}
+	// SCOPE.Datastore is admitted for SQL tables, which own Subtype "column"
+	// members; the jcl/cobol/erlang emitters of the same kind own none (#6543).
+	if kindTail(kind) == "datastore" && !datastoreMemberBearingLanguages[strings.ToLower(language)] {
 		return false
 	}
 	return true
--- a/internal/feedback/report.go
+++ b/internal/feedback/report.go
@@ -608,8 +608,9 @@ func kindTail(kind string) string {
 // this metric is for, which had never been sampled (#6543). Admitting the tail
 // is only half the fix: the members are Subtype "column", so the numerator has
 // to see them (memberChildSubtypes) and the non-SQL emitters of the same kind
-// have to be exempted (datastoreMemberBearingLanguages), or the tail merely
-// re-creates #6536 for a new population.
+// have to be exempted (datastoreMemberBearingKinds) — including the six SQL
+// subtypes that are not `table` — or the tail merely re-creates #6536 for a
+// new population.
 var classLikeKindTails = map[string]bool{
 	"class":     true,
 	"struct":    true,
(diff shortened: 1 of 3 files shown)


cajasmota:
Independent adversarial review — **REQUEST-CHANGES.** The RED-phase discovery and the leaf/container symmetry work are correct and stay. One structural problem, and it is the same defect this issue exists to fix.

## The fix re-creates the defect a fourth time, one level out

The exemption is keyed on **language**. But `SCOPE.Datastore`/`sql` is not one population — it is **seven** emit sites:

| subtype | site |
|---|---|
| `table` | `sql.go:249` |
| `view` | `:419` |
| `index` | `:444` |
| `function` / `trigger_function` | `:527` |
| `procedure` | `:585` |
| `trigger` | `:634` |
| `dbt_source` | `:695` |

**Only `table` emits `CONTAINS(contained_kind=column)` children.** None of the other six is in `nonClassSubtypes`, so all of them now enter the field-extraction denominator as guaranteed zero-field failures — the #6536 shape verbatim.

Measured on this head, unmodified production code, through the PR's own extractor-driven helpers:

```
CREATE TABLE users (id INTEGER PRIMARY KEY, email VARCHAR(255) NOT NULL);
CREATE VIEW active_users AS SELECT id FROM users;
CREATE INDEX idx_users_email ON users (email);
→ ClassTotal=3   ZeroFieldsPct=66.67

+ CREATE FUNCTION, CREATE PROCEDURE
→ ClassTotal=5   ZeroFieldsPct=80
```

One fully-columned table, and the metric reports **80% failure**. Index count exceeds table count in most real schema files, so SQL's reported rate would be dominated by the population that cannot pass — a metric that is confidently wrong about a population it enrols but cannot measure. That is precisely what #6543 was filed to stop.

**The PR's own replacement justification is the argument against its gate:** *"`SCOPE.Datastore` is not one population."* That reasoning is right, and it applies one level further down than it was applied.

The reason the suite never saw it: `sqlSource6543` is a bare `CREATE TABLE` only.

Fix is to gate on **subtype** — admit `table`, and `view` only if it ever gains column children — and to extend the fixture so the population the metric enrols is the population the extractor actually emits.

## A surviving mutant kills the central design claim

**M6 — allowlist → denylist.** `go vet` 0, `go test -count=1 ./internal/feedback/` **0 — SURVIVED**, re-confirmed on a fresh worktree.

```diff
-	if kindTail(kind) == "datastore" && !datastoreMemberBearingLanguages[strings.ToLower(language)] {
+	if kindTail(kind) == "datastore" && map[string]bool{"jcl": true, "cobol": true, "erlang": true}[strings.ToLower(language)] {
```

`TestDatastoreMemberBearingLanguagesIsTheOnlyGate_6543` probes only the three known languages, so *"a new `SCOPE.Datastore` emitter is excluded until someone checks it"* — the stated reason an allowlist was chosen over a denylist — has **no observation at all**. One line kills it: `isFieldExtractionCandidate("SCOPE.Datastore", "table", "duckdb")` must be false.

## The cobol justification is right, citing the wrong code

`cobol/extractor.go:748` data items are `SCOPE.Schema`/`field` whose **parents are also subtype `field`** (group items get the same subtype in the same append), so `nonClassSubtypes["field"]` already excludes them — they pass nothing.

The genuinely-passing cobol population is **IMS segments**: `kindIMSSegment = "SCOPE.Schema"` (`ims.go:295`), subtype `ims-segment`, admitted tail, absent from `nonClassSubtypes`, with `CONTAINS` → subtype `field` children (`ims.go:490`, edge at `:498-505`). Conclusion sound, citation wrong.

## Verified, for the record

- The gate really is tail+language (`report.go:786-790`), not language alone.
- The exemption is denominator-only and coherent with the numerator; an empty/unset `Language` misses the allowlist and is excluded — the safe direction, though also unobserved.
- `column` genuinely had to join `nonClassSubtypes`: columns are `SCOPE.Schema` (`sql.go:1021`), an already-admitted tail, which is exactly the RED-phase `ClassTotal=3, 100%`. Nothing else with an admitted tail is enrolled or dropped by either half.
- `{field, column}` has **no counter-example**. A sweep for member-shaped subtypes across the extractor tree (`property`, `attribute`, `member`, `variable`, `record`, `slot`, `param`, `element`, `entry`) found only Vue `slot` and `variable`/`record`/`param` leaves, none a container's declared member.
- M1–M5 all reproduce, failure counts matching the table exactly.
- **M5's fixture strengthening is real and load-bearing.** Re-weakening `TestColumnlessTableStillCountsAsZeroField_6543` by deleting its trigger `CONTAINS` edge makes M5 survive again. The self-report that M5 initially survived was honest, and the fix for it was the right one.

## Two fixture notes

- `TestNonColumnBearingDatastoresExcluded_6543`'s erlang arm is doubly covered — erlang is already in `nonFieldBearingLanguages`, so that arm passes with the new gate deleted.
- `TestColumnlessTableStillCountsAsZeroField_6543` hand-writes its trigger as `SCOPE.Operation`/`trigger`, but the extractor emits `SCOPE.Datastore`/`trigger` (`sql.go:634`). That departure from the file's own stated principle — *"kind and subtype strings come FROM THE EXTRACTOR"* — is exactly where the defect above hid.

Fixes in progress on this branch.

Issue #6564, opened 2026-08-24
internal/quality is red on a clean main: baseline.json's measured_on points at a commit no longer reachable from origin/main, and nothing prevented writing it


cajasmota:
`go test ./internal/quality/` fails on a **clean checkout of `origin/main`**, and has now been independently reproduced by three separate reviews plus an outside contributor. It is ambient red on every PR, which is the real cost.

## The failure

```
--- FAIL: TestBaselineMeasuredOnIsReachable
    baseline.json measured_on "0ca8d585e" is not an ancestor of origin/main
```

`git merge-base --is-ancestor 0ca8d585e origin/main` exits **1**. The commit exists in some local clones — it is an orphaned pre-amend/pre-rebase sibling — but it is not reachable from `origin/main`, so the assertion is correct and the data is wrong.

## Why this matters more than one red test

**It fails on work that does not touch quality at all.** It has now surfaced on a contributor's `.csproj` fix, on a FastAPI decorator fix, on a Python DI change, and on an Angular URL-folding change — none of which go near `internal/quality/`. Every one of those authors had to work out for themselves that the red was not theirs.

That is the same corrosion #6512 describes for the Windows legs: **a persistent failure unrelated to the change under review teaches everyone to skip the signal.** The next real regression in this package is the one that gets waved through.

It is also unfriendly to outside contributors specifically, who have no way to know which reds are baseline noise and which are theirs.

## The mechanism worth fixing, not just the value

`measured_on` records the commit the baseline was measured against. It was written by a run whose commit **later stopped being reachable** — amended, rebased, or force-pushed away — and nothing at write time checks that the commit will survive, nor does anything re-anchor it afterwards.

So a correct fix is not only "re-record the baseline". It is deciding what should happen when the recorded commit becomes unreachable:

1. **Re-record** — run `scripts/quality/run.sh --runs 1 --update-baseline` through the writer (never by hand; #6466 landed a canonicality gate on that file). Fixes today's instance, leaves the mechanism.
2. **Validate at write time** — refuse to stamp a `measured_on` that is not reachable from the remote's default branch, so the unreachable value can never be written.
3. **Degrade instead of failing** — treat an unreachable `measured_on` as "unknown provenance" with a warning rather than a hard test failure, on the grounds that it says nothing about whether the *measurements* are still valid.

(2) and (3) are not exclusive, and (1) is needed either way to clear the current red.

[...]

Issue #6466, opened 2026-08-21
baseline.json is a generated ratchet floor with no canonicality gate, so unrelated encoding churn lands silently in a test gate


cajasmota:
`internal/quality/golden/baseline.json` is a generated artifact — `scripts/quality/ratchet.py --update-baseline` writes it — and **nothing gates its canonical form**.

Its neighbour `docs/coverage/registry.json` has a strict one: `go run ./tools/coverage fmt --check`, wired into `.github/workflows/coverage-docs.yml` (regression #2907). It fails CI on any byte that diverges from what the Go writer produces. `baseline.json` has no equivalent.

## How it surfaced

While editing `baseline.json` for the VB.NET declared-type work, an agent round-tripped it through Python with `ensure_ascii=False`. That silently converted **four pre-existing `—` escapes into literal em-dashes**, in lines nobody was editing.

The intended change was three lines. The actual diff was seven. CI would not have caught it — the canonicality check is `tools/coverage`-scoped — and the file would have drifted from what `--update-baseline` produces, so the *next* legitimate ratchet update would have shown a spurious four-line diff that had nothing to do with that update.

It was caught by the agent noticing and re-serialising with `ensure_ascii=True`, not by any gate.

## Why it matters more than a cosmetic diff

`baseline.json` is the ratchet's floor. Its whole job is that a change to it is deliberate and visible. Unrelated churn in the same file defeats that: a reviewer scanning a baseline diff to check that a recall number moved for a good reason now has to separate signal from encoding noise, and the noise looks exactly like content.

The failure mode is the same one the coverage registry gate exists to prevent — and it is arguably worse here, because the coverage pages are documentation while this file is a **test gate**.

## Scope

Add a canonicality check for `baseline.json`. The cheapest honest form is to have the check re-run whatever writer owns the file and assert no diff, the way `fmt --check` does — rather than hand-rolling a formatter, which would then be a second source of truth that can itself drift.

Establish first which writer is authoritative: `scripts/quality/ratchet.py --update-baseline` is the documented path, but confirm nothing else writes the file (Go-side updates from the quality tooling would mean two writers with two encodings, which is its own bug and would need settling before a gate makes sense).

Worth checking in the same pass whether other generated JSON under `internal/quality/` has the same gap.

## Related traps worth knowing while working here

[...]

PR #6568, opened 2026-08-24
fix(#6564): the writer stamps a merge-base, so the manual re-pin that failed three times is gone

cajasmota:
Refs #6564. This is the **mechanism** half only — the baseline is deliberately **not** re-recorded here, so `TestBaselineMeasuredOnIsReachable` stays red on `a71cdf3d0` until the owner's re-record, which will now stamp a merge-base by itself. Re-recording before this landed would have stamped a fourth feature-branch HEAD.

## What was actually broken

Not the design. `baseline.json`'s own `measured_on_note` already said the field is the merge-base, and `ratchet.py`'s comment above the write already said it is "a BASE commit, not the commit this file lands in". What did not exist was enforcement: `git_sha()` was `git rev-parse --short HEAD`, and the correct value depended on a human re-pinning the field after every re-record. That step has now been skipped three times, most recently in PR #6552; this repo squash-merges, so each stamped feature-branch commit was orphaned by the very PR carrying it.

## 1. The writer stamps the merge-base

`git_sha()` now returns `git merge-base HEAD <default branch>`, shortened. Reachable by construction at write time, and permanently — squash, rebase, amend and force-push do not touch it. Verified against both recorded failures: `merge-base(a71cdf3d0) = 8595d7a6b` and `merge-base(0ca8d585e) = 3516777a4`, both ancestors of `origin/main`.

The default branch is resolved in this order, first ref that exists winning: `QUALITY_BASE_REF` (override for a checkout that knows its base but cannot advertise it), `refs/remotes/origin/HEAD`, `origin/main`, `origin/master`, local `main`, local `master`. Remote-tracking refs come first because they are what the merge actually lands on; a local `main` can be stale or diverged.

**Degradation, and why.** When none of those refs exists — a shallow clone (`actions/checkout` defaults to `fetch-depth: 1`, and `test.yml` sets no depth), a repo with no remote, a detached CI checkout — or when `merge-base` finds no common ancestor, the value degrades to the string `"unknown"` with a warning on stderr naming `QUALITY_BASE_REF` and the full-clone fix. It never degrades to `HEAD`. `HEAD` is the value that *looks* right and is not durable, which is exactly how this bug survived three occurrences; `"unknown"` is visibly not a sha, so it is caught in review, and `internal/quality`'s existing hex-shape check refuses it too. Note that `--update-baseline` only ever runs on a developer machine — no workflow invokes it — so the degraded path is a safety net, not the normal one.

## 2. The property is observable at write time

`build()` now calls `ensure_durable_measured_on()` before it grades anything: a `measured_on` that is not an ancestor of the resolved default branch is **refused**, and no baseline is written. Previously nothing at write time objected, so the only guard was a sentence in a note, and the failure was discovered days later by someone reviewing an unrelated change. The guard is a no-op only when there is no default branch to check against, which is the same state that produces `"unknown"`.

## Where the test lives, and why there is a Go wrapper

The tests are Python, in **`scripts/quality/test_ratchet.py`** (stdlib `unittest`, no new dependency, runnable as `python3 scripts/quality/test_ratchet.py`). That is the right home: the property is a property of the **writer**, and the writer is Python. A Go assertion on the committed `baseline.json` structurally cannot catch this — by the time the file is on `main` the orphaning squash has already happened, which is precisely how all three occurrences were found. The tests build a synthetic repo shaped like a re-record (a default branch plus a feature branch that has moved past it) and drive `git_sha()` and `build()` against it, which no test of the checked-in JSON can do.

There is also one thin Go test, `internal/quality/ratchet_measured_on_6564_test.go`, whose entire body runs the Python file. It is not a second assertion: it exists so the Python tests actually run in CI alongside the gate they protect, rather than only when someone remembers to invoke `python3`. A remembered manual step is the failure mode this PR removes, so leaving the new tests behind one would have been ironic. It skips if `python3` or `git` is missing, matching the existing `requirePython3` convention in that package.

## RED → GREEN, and mutation scores

Against unmodified `ratchet.py` (RED, 6 of 7 failing):

```
FAILED (failures=6)
AssertionError: 'c39c0d9' == 'c39c0d9' : git_sha() stamped the feature-branch HEAD; squash-merge orphans it
AssertionError: 1 != 0 : '70328e7' is not an ancestor of the default branch
AssertionError: '4ce8cb7' == '4ce8cb7' : degraded to HEAD, which is the bug
AssertionError: SystemExit not raised
```

The seventh — the note-carry test — passed from the start, deliberately: it is the guard on what must **not** change.

After the fix: `Ran 7 tests ... OK`, and `go test -count=1 ./internal/quality/` fails on nothing but the expected pre-existing `TestBaselineMeasuredOnIsReachable`.

Both mutants named in #6564 die:

| Mutant | Result |
|---|---|
| revert the sha source to `rev-parse --short HEAD` | `FAILED (failures=4, errors=2)` |
| delete the `ensure_durable_measured_on()` call from `build()` | `FAILED (failures=1)` |

The second is scored at the **call site**, not on the helper: an extracted-and-tested predicate that `build()` is free to ignore would prove nothing.

## The 4KB note survives

`measured_on_note` is ~9KB of provenance (solidity-mini and vbnet-mini measurement exceptions, the #6454 disposition cost, the #6440 corpus effect) that a fixed-key-set rebuild deleted once already. This PR touches `build()`, the function that does that rebuild, so the survival is tested twice: the existing `TestRatchetUpdateCarriesMeasuredOnNote_6260` (still green) and a new Python `test_build_carries_the_measured_on_note_across_a_rebuild` that carries a 4KB note through a real `build()`. The note grew 9016 → 9407 chars in this PR and lost nothing — the `#6440`, `#6454`, solidity-mini and vbnet-mini passages are all still present.

## 3. The note correction

The instruction "Re-record with `--update-baseline` if you like, but re-pin this field afterwards" is now wrong, so it says instead that the writer records the merge-base itself and refuses a non-ancestor, that there is no manual step, and that hand-editing the field is what #6466's canonicality gate exists to stop. The history is kept and corrected from "twice" to three times, naming #6552. Everything else in the note is byte-identical.

The note was edited **through `ratchet.serialize()`**, not by hand-formatting, and `ratchet.py canon` passes. `measured_on` itself was not touched.

## Not observed by any test — stated plainly

- **That a real squash-merge preserves the stamped value.** The tests assert ancestry of the merge-base against the default branch in a synthetic repo; nothing here performs an actual squash-merge and re-checks. The property is structural (a merge-base is already on the default branch), but it is reasoned, not executed.
- **The shallow-clone degradation as CI actually produces it.** The `"unknown"` path is exercised by a repo with no default-branch ref, not by a genuine `--depth 1` clone. Same code path, a different way of getting there.
- **`QUALITY_BASE_REF`.** The override is implemented and documented but has no test of its own; the ref-resolution order is only exercised through the remote and local-branch cases.
- **That the resolved default branch is the branch the PR will actually merge into.** If someone works from a base other than `main`/`master` without setting `QUALITY_BASE_REF`, the stamp anchors to `main` — still durable, but one merge-base further back than they may expect.


cajasmota:
Independent adversarial review — **APPROVE**, with two non-blocking follow-ups.

The two properties that would have made this PR self-defeating were both verified **by execution**, not by reasoning.

## The squash-merge scenario, actually run

The body admitted this was "reasoned structurally, never executed". It has now been executed, in a throwaway repo (bare upstream + clone): advance `main`, branch `feature`, run `git_sha()` + `ensure_durable_measured_on()` on the branch, then `git merge --squash`, push, delete the branch, `reflog expire --all`, `gc --prune=now`.

```
STAMPED=3e3ae71   FEATURE_HEAD=54f572f
ancestor_of_origin_main_EXIT=0
ancestor_of_HEAD_EXIT=0
```

The stamp is the merge-base and it survives the squash **and** a full GC that destroyed the feature head. That is the property #6564 exists for, confirmed rather than argued.

## The Go wrapper genuinely fails the build

This was the one that mattered most: a wrapper that swallows a non-zero exit would have re-created the exact defect — a guard nobody notices isn't running. Probed by injecting `self.fail("DELIBERATE-WRAPPER-PROBE")` into the Python suite:

```
go test -count=1 -run TestRatchetMeasuredOnSelfTests_6564 ./internal/quality/   → EXIT=1
--- FAIL: TestRatchetMeasuredOnSelfTests_6564 (2.77s)
    AssertionError: DELIBERATE-WRAPPER-PROBE
```

`CombinedOutput()`'s error is checked and escalated with `t.Fatalf`, and `unittest.main()` exits non-zero. Restored, shasum re-verified.

## `"unknown"` is genuinely caught, and not only locally

`internal/quality/baseline_test.go:190-194` rejects any non-hex rune, so `unknown` fails on `u`. Crucially the shape check sits **above** the shallow-clone `t.Skip`, so it still fires on the shallow CI checkouts. And no real CI configuration reaches the degraded path at all: `quality.yml` is `workflow_dispatch`-only with `mode: [ratchet, strict]`, `--update-baseline` is not among its options, and `scripts/verify2/run-quality.sh` never passes it. Shallowness affects readers, not the writer.

`ensure_durable_measured_on` is called on the **only** path that writes the field (`ratchet.py:298-299`, consumed at `:328`); `serialize`/`canon` do not re-derive it. No unguarded second writer.

## The note survives, and the test observes survival rather than presence

9016 → 9407 chars; `#6440`, `#6454`, `#6466`, `#6552`, solidity-mini and vbnet-mini passages all intact. A sentence-level diff shows only the four intended sentences changed. `test_build_carries_the_measured_on_note_across_a_rebuild` seeds a 4KB note into the *prior* baseline, runs a real `build()`, and asserts equality on the output — so deleting the carry-forward makes the key absent and the test fail. That is a survival test, not a presence check.

## Mutants

| # | mutant | exit | result |
|---|---|---|---|
| 1 | `git_sha()` → `rev-parse --short HEAD` (re-run) | 1 | **DEAD**, matches the body |
| 2 | delete `ensure_durable_measured_on(...)` from `build()` (re-run) | 1 | **DEAD**, matches the body |
| 3 | **new, permissive** — local heads before remote-tracking in `DEFAULT_BRANCH_REFS` | 0 | **SURVIVED** |
| 4 | **new, permissive** — `ensure_durable` checks ancestry of `HEAD`, not `ref` | 1 | **DEAD** |

Plus my own, reported earlier: degrade to `HEAD` instead of `"unknown"` → **DEAD**.

## Follow-up 1 — mutant 3, and where it sits

```diff
 DEFAULT_BRANCH_REFS = (
-    "refs/remotes/origin/main",
-    "refs/remotes/origin/master",
     "refs/heads/main",
     "refs/heads/master",
+    "refs/remotes/origin/main",
+    "refs/remotes/origin/master",
 )
```

`Ran 7 tests … OK`. The comment directly above asserts *"remote-tracking refs come first because they are what the merge actually lands on; a local `main` can be stale or diverged"* — and nothing observes it, because `make_repo()` points local `main` and `refs/remotes/origin/main` at the same commit, making the ordering structurally unobservable.

That is this milestone's dominant defect class — prose asserting a property no test observes — **one layer down inside the fix for that same class**. Worth stating plainly rather than filing quietly.

Impact is bounded: a stale local `main` still yields a *durable* merge-base, just an older one. Follow-up, not a blocker.

## Follow-up 2 — `QUALITY_BASE_REF` is self-referential

`ensure_durable_measured_on` validates against **the same ref it computed from**, so the guard can never object to a bad override. Demonstrated on unmutated `c06305f28`:

```
QUALITY_BASE_REF=refs/heads/feature2
stamped: b66c860   feature HEAD: b66c860     ← the #6564 bug, reproduced
ensure_durable_measured_on: ACCEPTED
that sha ancestor of origin/main EXIT=1
```

Requires deliberate misuse of an env var the body already flags as untested, so not blocking. Cheap hardening: cross-check against `refs/remotes/origin/HEAD`/`origin/main` when one exists, rather than only against the override.

The body's other named unobserved item — "the resolved default branch is the branch the PR will really merge into" — is **acceptable**: a non-`main` base without the override still produces a durable, merely older, ancestor.

Merging once the Windows leg lands. Both follow-ups get their own issue; neither justifies holding a fix for an ambient red that is currently costing every contributor a false signal.

PR #6587, opened 2026-08-25
fix(#6514): strawberry_graphql.yaml's include_router twin double-emitted the edge and made both copies untestable

### LINKED ISSUE #6514: strawberry_graphql.yaml duplicates fastapi.yaml's include_router rule byte-for-byte, so neither can be tested in isolation
Found by an independent review of PR #6510, and proven by a mutant rather than by reading.

## The finding

`internal/engine/rules/python/frameworks/strawberry_graphql.yaml:118` contains an `include_router` rule that is **byte-identical** to the one in `internal/engine/rules/python/frameworks/fastapi.yaml:120` — same pattern, same `source_type`, same `target_type`, same capture groups.

Either one alone emits `ROUTES_TO Service:app -> Route:router`. So each is a complete substitute for the other, and **neither can be tested in isolation.**

## How it surfaced, which is the part worth reading

PR #6510 added `python-fastapi-mini` and pinned the `ROUTES_TO` edge with a note claiming the row would catch a `fastapi.yaml`-side regression.

The reviewer disabled `fastapi.yaml`'s `include_router` rule outright — renamed the pattern token, asserted the occurrence count was 1, rebuilt — and the fixture stayed **exit 0, 4/4 relationships, 57 edges**. The edge was still emitted, still stamped `pattern_type: yaml_driven`, sourced from the strawberry file.

So a rule can be deleted from the file it belongs to and every gate stays green.

## Why this is worth fixing rather than noting

The immediate consequence is narrow — one fixture note was overclaiming, and that is corrected in #6510. The durable consequence is not: **any future FastAPI `include_router` coverage is unfalsifiable by construction.** Someone will write a test for that rule, watch it pass, and have measured nothing. That is the same shape as the vacuity problems this milestone has been closing one at a time, except here the duplication *guarantees* it rather than merely permitting it.

It is also a latent correctness question in its own right. `include_router` is a FastAPI/APIRouter construct. Whether Strawberry's GraphQL router genuinely needs its own identical copy — or whether this is a copy-paste that has been silently double-emitting and getting deduped downstream — has **not** been established, and should be step 1.

1. **Does the strawberry copy have an independent reason to exist?** If Strawberry apps really do use `include_router`, the two rules need to differ in something observable — a distinct `pattern_type` tag, a distinct source anchor — so each is separately testable. If it is a copy-paste, delete it.
2. **Are both currently firing on the same input?** If so, check whether anything downstream is deduping them and what the edge count would be without that dedup. A double emission that gets deduped is invisible until the dedup key changes — see #6482, where an ignore-map entry silently turned a red parity test green.
3. Whether any other rule file in `internal/engine/rules/` carries a duplicated pattern of this kind. This was found by accident, not by a sweep, so the population is unknown and should not be assumed to be one.

## Acceptance

Disabling either rule must make some test fail. Today, disabling `fastapi.yaml`'s does not.

**Killing mutant:** rename the pattern token in `fastapi.yaml`'s `include_router` rule and rebuild. That must go red once this is fixed. It is green today, and the reviewer's run is the evidence.

Refs #6444, #6510

cajasmota:
Closes #6514.

## What the two rules were

`strawberry_graphql.yaml:118` and `fastapi.yaml:119` carried the same relationship rule with every field equal — pattern, `source_type: Service`, `target_type: Route`, `ROUTES_TO`, `source_group: 1`, `target_group: 2`. I read both files in full before deleting: nothing around them differentiated the copies either. The strawberry file's own three `strawberry.Schema(...)` rules are unrelated and untouched.

## Which copy survives, and why the other one is not needed

`fastapi.yaml`'s. `include_router` is FastAPI/APIRouter API.

The strawberry copy existed on the theory that a Strawberry-on-FastAPI mount needs its own rule. It does not: `Detect` resolves compiled rule sets by `file.Language` alone, so **every** python rule set runs on **every** python file. `relationship_rules` carry no `requires_framework` gate (only `source_patterns` do). fastapi.yaml's copy therefore already fires on `app.include_router(graphql_app, prefix="/graphql")` in a Strawberry app.

## Both copies were firing, and the emission was doubled

`Detect`'s dedup (`seenEntities`) covers entities only; relationship records are appended unconditionally. Running the real embedded rules over `internal/quality/golden/python-fastapi-mini/src`, `main.py` emitted `Service:app -ROUTES_TO-> Route:router` **twice**, with byte-identical properties — `framework` is the language key `python`, not the rule file, so the two records were indistinguishable in the output. Downstream dedup collapsed them, which is why the fixture graded 4/4 with fastapi.yaml's copy disabled.

**Before/after is not byte-identical, and the sole difference is the duplicate:**

```
-REL main.py | Service:app -ROUTES_TO-> Route:router | props=[{framework python} {pattern_type yaml_driven}]
```

The other 14 entity/relationship records over the fixture are unchanged.

## The pin

Two tests in `internal/engine`, both over the real embedded rules:

- `TestIncludeRouter6514_EmitsExactlyOneRoutesToEdge` — counts **raw, pre-dedup** records for a minimal FastAPI mount. Pre-dedup is the only place the double emission was ever visible. `== 1` fails from both directions: the surviving rule going away (0) and a twin coming back (2). `>= 1` would have been satisfied by the duplication itself and measured nothing.
- `TestIncludeRouter6514_RuleIsLoadedOnlyFromFastAPI` — pins the carrier, which an edge count cannot: identical properties mean "fix the duplication by keeping strawberry's copy instead" would otherwise pass.

## Mutants

`go vet ./internal/engine/` first on every mutant (exit 0 each time), then `go test -count=1 ./internal/engine/`, full package, no `-run` filter.

| Mutant | Result |
|---|---|
| **PERMISSIVE** — delete the *surviving* `fastapi.yaml` rule | **DEAD**, exit 1 |
| Re-add the byte-identical strawberry twin (the pre-fix tree) | **DEAD**, exit 1, both tests |
| `fastapi.yaml` `source_group: 1` → `2` | **DEAD**, exit 1, behavioural test |

The permissive mutant is the load-bearing one, and it says more than "the test works": **only the two new tests failed.** Deleting the FastAPI `include_router` rule outright was invisible to every other test in the package. That is the acceptance criterion from the issue.

Clean tree: `go vet` 0, `gofmt -l` clean, `go test -count=1 ./internal/engine/` exit 0.

## Also updated

`python-fastapi-mini/expected.json`'s ROUTES_TO note documented this defect as deliberately-unfixed. It now describes the fix and keeps the remaining honest limitation — `must_exist` grading runs post-dedup, so that row still does not pin the rule on its own; the Go test does.

## What I did NOT do

Step 3 of the issue — the wider sweep for duplicated patterns across `internal/engine/rules/` (`\.Use\s*\(` ×4, `@Inject` ×4, several inside `python/`) — is out of scope here. Not all of those are substitutes (the two `Depends(...)` copies are both inside `fastapi.yaml` and serve different relationships), so a blanket "no two rules share pattern+types+groups" guard would need a reviewed allowlist. That is its own change; this PR fixes the one instance and pins it.

Per repo policy I ran `./internal/engine/` only, not the full suite. `internal/quality` was not run — the only change there is prose inside a `note` field, and nothing in Go asserts that text.


cajasmota:
Both surviving mutants addressed in 8d5058143. The fix itself is unchanged.

**Mutant D — a duplicate does not have to be spelled identically.** You are right that the prefixed form is the one the deleted rule existed for, and my probe never drove it. The count test is now table-driven over both mounts: the un-prefixed FastAPI one and `app.include_router(graphql_app, prefix="/graphql")`, each asserting exactly one raw pre-dedup record.

**Mutant E — the name is not the carrier.** Replaced with a test that walks the rule tree on disk and asserts the rule *file*, `python/frameworks/fastapi.yaml`, since the loaded rule sets carry no file identity. Membership is now decided by **behaviour, not pattern text**: a rule carries the mount if its regex matches a mount line *and* its capture groups name the same two endpoints. That makes a differently-spelled re-add caught at file level too, not just by the count. The failure message names what is actually asserted.

## Verdicts

`go vet ./internal/engine/` exit 0 before each; then `go test -count=1 ./internal/engine/`, full package, no `-run` filter.

| Mutant | Verdict |
|---|---|
| **D** — prefixed-only twin appended to `strawberry_graphql.yaml` | **DEAD**, exit 1 — both tests; the `/prefixed` subtest catches the emission |
| **E** — rule moved to a new `fastapi_extras.yaml` named `FastAPI Extras` | **DEAD**, exit 1 — file test only, correctly: the edge count *is* 1 when the rule moves, which is exactly why the carrier assertion has to exist |

Originals re-run against the tightened guard — all still DEAD, and all now fail both tests rather than one:

| Mutant | Verdict |
|---|---|
| **A** — delete the surviving `fastapi.yaml` rule (PERMISSIVE) | **DEAD**, exit 1 |
| **B** — re-add the byte-identical strawberry twin (pre-fix tree) | **DEAD**, exit 1 |
| **C** — `source_group: 1` → `2` | **DEAD**, exit 1 |

Clean tree: `gofmt -l` clean, `go vet` 0, `go test -count=1 ./internal/engine/` exit 0. Rule YAML restored by `cp` from byte-backups between every mutant, shasums verified equal.

Scope held: no baseline touched, no widening to the other duplicate-pattern families.

PR #6600, opened 2026-08-25
feat(#6537): Arm B — a real vbnet rule bucket, after measuring that an alias onto csharp fires nothing

### LINKED ISSUE #6537: v0.3.0 shipped a VB.NET extractor with no rules/vbnet bucket at all, so framework detection is structurally 0% for the language
Found while grounding the feedback report in #6535, which measured **0.0% framework-annotated entities across 45,663** on a real legacy VB.NET WinForms application. The report itself could not distinguish "detector did not fire" from "vanilla codebase". From our side it is not ambiguous.

## There is no `vbnet` rule bucket at all

- `internal/engine/rules/` has **37 language buckets. None is `vbnet`.**
- The entire `internal/engine/` Go tree contains **zero occurrences of the string `vbnet`**.
- The only mention anywhere in the ruleset is an unrelated `'*.vbproj'` glob at `internal/engine/rules/csharp/build_tools.yaml:8`.

The loader buckets rules by top-level dirname — `rules/<lang>/frameworks/*.yaml` (`internal/engine/loader.go:79-81`) — and `Detect` resolves the compiled set by `file.Language` (`internal/engine/detector.go:235-238`). So `d.compiled["vbnet"]` is empty, no aliasing exists, and **no rule can fire on a VB.NET file regardless of what it contains**. 0.0% of 45,663 is fully explained.

## A WinForms rule exists — under the wrong key, and aimed at the wrong era

`internal/engine/rules/csharp/frameworks/winforms.yaml` is registered in `internal/engine/rules/csharp/_manifest.yaml:1`. Two separate reasons it cannot help here:

1. It is keyed to `csharp`, and the files are `vbnet`.
2. Its detection payload is `.csproj` / `*.Designer.cs`-oriented, and its own `notes` field explicitly defers **legacy pre-SDK projects** — which is precisely what a legacy VB.NET WinForms app is.

## This is a known latent class, now with a confirmed instance

`internal/engine/detector.go:234-252` already documents this exact hazard for the `cicd`, `ansible` and `kubernetes` buckets, and `:224-231` for `javascript_typescript`. Those comments describe the trap; this issue is the first case where a **whole language shipped with no bucket** and a user measured the consequence.

Worth stating plainly: v0.3.0 shipped a VB.NET extractor that produces 45k entities, with **zero framework recognition capability** for that language. The extractor works; the detector was never extended to meet it.

## What a fix needs to decide

1. **Does `vbnet` get its own bucket, or alias onto `csharp`?** Aliasing is cheaper and the frameworks genuinely overlap (WinForms, ASP.NET, ADO.NET), but the *patterns* differ — VB.NET syntax is not C# syntax, so a shared rule's regex will not match. Measure before assuming aliasing buys anything.
2. **Whether a language shipping with no rule bucket should fail a check.** This is the generalisable half: the detector has no assertion that every extractor-supported language has a reachable rule set, so the next language lands the same way.
3. **Which frameworks to target first** — a legacy WinForms/.NET Framework stack is the concrete, reported case.

## Observability

Go level, `internal/engine` — a test asserting that every language key produced by the extractor registry resolves to a non-empty compiled rule set needs no fixture.

**Killing mutant:** delete a language's rule bucket. A test must go red naming that language. Today deleting any bucket is silent, which is how this shipped.

Refs #6535

cajasmota:
`rules/vbnet/` did not exist and nothing aliased onto it, so `d.compiled["vbnet"]`
was empty and no rule could fire regardless of file content.

ALIAS-VS-OWN-BUCKET, MEASURED FIRST
Realistic legacy VB WinForms source was run through Detect with
Language="csharp" — exactly what an alias vbnet -> csharp would produce. The
csharp bucket compiles 46 rules (27 source_patterns / 6 relationship_rules /
13 file_conventions) and emitted 0 entities and 0 relationships. C# rule regexes
key off braces, `void`, `[Attribute]` and `//`; VB.NET uses Sub/End Sub,
Inherits, Handles, WithEvents and `'`. An alias would have turned the Arm A
guard green while firing nothing. Hence a real bucket.

WHAT LANDS
  rules/vbnet/_manifest.yaml
  rules/vbnet/frameworks/winforms.yaml — 7 source_patterns, 2 relationship_rules,
  2 file_conventions, all VB syntax:
    Inherits ...Form           -> View
    Inherits ...UserControl    -> Component
    Friend WithEvents X As ... -> Component
    Sub X(...) Handles C.Event -> Operation, plus Component -HANDLES-> Operation
    Module X                   -> Module (gated on import_markers)
    Application.Run(New F(     -> Config, plus Module -INSTANTIATES-> View
    Me.Controls.Add(Me.X)      -> Component

The csharp winforms.yaml `notes` field defers legacy pre-SDK projects; that
deferral does NOT apply here. It defers them because its detection is
Sdk="Microsoft.NET.Sdk.WindowsDesktop" / <UseWindowsForms>true</UseWindowsForms>,
which a pre-SDK project has neither of. Every pattern here keys off VB *source*
syntax, which a pre-SDK project emits identically — no project-file era gate is
involved.

HANDSHAKE WITH ARM A (same commit, as designed)
  - deleted the `vbnet` row from knownLanguageRuleGaps
  - inverted the assertion in language_rule_coverage_6537_test.go: vbnet must
    now NOT reappear in the gap list

MUTANTS (go vet first on every one; full ./internal/engine/ package, no -run)
  M1 delete winforms.yaml entirely        vet=0 test=1  DEAD (4 tests red, incl. the Arm A guard)
  M2 PERMISSIVE form pattern -> (?i)class vet=0 test=1  DEAD (fired on C# source)
  M3 NULLIFY WithEvents pattern           vet=0 test=0  SURVIVED -> test strengthened -> DEAD
  M4 NULLIFY Handles-Sub pattern          vet=0 test=1  DEAD
  M5 PERMISSIVE HANDLES rel -> word pairs vet=0 test=1  DEAD (2 tests)
  M6 NULLIFY Me.Controls.Add pattern      vet=0 test=1  DEAD
  M7 NULLIFY Module pattern               vet=0 test=1  DEAD
  M8 PERMISSIVE WithEvents -> any field   vet=0 test=1  DEAD (3 assertions)

M3 is the finding. The first version of the test asserted Component:SaveButton,
which Me.Controls.Add ALSO emits — so anchoring the WithEvents pattern
impossibly left the suite green and the pattern's firing unobserved. The fixture
now declares CustomerTextBox/OrdersGrid WithEvents-only and StatusLabel
Controls.Add-only, so each source_pattern is pinned by an assertion no other
pattern can satisfy. M6 and M7 were added afterwards to confirm the separation
holds in both directions.

All fixture VB/C# source is hand-written for this test; nothing copied from any
corpus.

Closes #6537
Refs #6535

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_0111KEDUVWEWm9G5RRnJqXft

### DIFF
--- a/internal/engine/language_rule_coverage_6537_test.go
+++ b/internal/engine/language_rule_coverage_6537_test.go
@@ -254,13 +254,49 @@ func TestLanguageRuleCoverage6537_ListIsCurrent(t *testing.T) {
 	}
 
 	// vbnet was the reported instance (#6535) and the reason this guard exists.
-	// Arm B of #6537 landed rules/vbnet/frameworks/winforms.yaml, so it is no
-	// longer a gap — the assertion that it still was one was deleted in the same
-	// commit, which is the intended, visible handshake between the two arms.
-	// The direction is now inverted: vbnet must NOT reappear in the gap list.
-	if contains(stillOpen, "vbnet") {
-		t.Errorf("vbnet is an open rule gap again; rules/vbnet no longer compiles anything "+
-			"(#6537 Arm B) — open gaps are now: %v", stillOpen)
+	// Arm B of #6537 landed rules/vbnet/frameworks/winforms.yaml, so its entry
+	// was deleted from knownLanguageRuleGaps in the same commit — which is what
+	// makes the two arms hand off visibly.
+	//
+	// There is deliberately no "vbnet must not be in stillOpen" assertion here.
+	// stillOpen is built by iterating knownLanguageRuleGaps, so with vbnet gone
+	// from that list such a check is unsatisfiable under every input — dead
+	// code wearing a comment. The reachable version is
+	// TestLanguageRuleGaps6537_ClosedGapsStayClosed, which checks
+	// CompiledRuleCount directly; the unlisted-gap guard in
+	// TestLanguageRuleCoverage6537_EveryProducibleLanguage catches the same
+	// regression from the other side.
+}
+
+// TestLanguageRuleGaps6537_ClosedGapsStayClosed is the ratchet's pawl. A gap
+// that has been closed leaves knownLanguageRuleGaps and thereby leaves that
+// guard's population entirely; this test keeps watching it, by asking the
+// detector directly rather than by consulting any list.
+func TestLanguageRuleGaps6537_ClosedGapsStayClosed(t *testing.T) {
+	rules, err := engine.LoadAllRules()
+	if err != nil {
+		t.Fatalf("LoadAllRules: %v", err)
+	}
+	d := engine.New(rules)
+
+	if len(closedLanguageRuleGaps) == 0 {
+		t.Fatal("closedLanguageRuleGaps is empty; this test would assert nothing")
+	}
+	for _, closed := range closedLanguageRuleGaps {
+		if closed.Reason == "" {
+			t.Errorf("closed gap %q has no Reason", closed.Language)
+		}
+		if got := d.CompiledRuleCount(closed.Language); got == 0 {
+			t.Errorf("%q was recorded as a CLOSED rule gap (%s) but compiles 0 rules again; "+
+				"its bucket has been removed or has stopped loading",
+				closed.Language, closed.Reason)
+		}
+		for _, open := range knownLanguageRuleGaps {
+			if open.Language == closed.Language {
+				t.Errorf("%q appears in BOTH knownLanguageRuleGaps and closedLanguageRuleGaps; "+
+					"a gap is one or the other", closed.Language)
+			}
+		}
 	}
 }
 
--- a/internal/engine/language_rule_coverage_6537_test.go
+++ b/internal/engine/language_rule_coverage_6537_test.go
@@ -253,12 +253,14 @@ func TestLanguageRuleCoverage6537_ListIsCurrent(t *testing.T) {
 			len(knownLanguageRuleGaps), len(stillOpen), stillOpen)
 	}
 
-	// vbnet is the reported instance (#6535) and the reason this guard exists.
-	// When Arm B lands a `vbnet` bucket this assertion fails and the entry gets
-	// deleted — which is the intended, visible handshake between the two arms.
-	if !contains(stillOpen, "vbnet") {
-		t.Errorf("vbnet is no longer an open rule gap; remove it from knownLanguageRuleGaps "+
-			"(and update this assertion) — open gaps are now: %v", stillOpen)
+	// vbnet was the reported instance (#6535) and the reason this guard exists.
+	// Arm B of #6537 landed rules/vbnet/frameworks/winforms.yaml, so it is no
+	// longer a gap — the assertion that it still was one was deleted in the same
+	// commit, which is the intended, visible handshake between the two arms.
+	// The direction is now inverted: vbnet must NOT reappear in the gap list.
+	if contains(stillOpen, "vbnet") {
+		t.Errorf("vbnet is an open rule gap again; rules/vbnet no longer compiles anything "+
+			"(#6537 Arm B) — open gaps are now: %v", stillOpen)
 	}
 }
 
--- a/internal/engine/language_rule_gaps_allowlist_test.go
+++ b/internal/engine/language_rule_gaps_allowlist_test.go
@@ -91,11 +91,10 @@ var knownLanguageRuleGaps = []languageRuleGap{
 	{Language: "nim", Reason: "C: no rules/nim bucket; ORM/migration recognition is 10 custom_nim_* Go extractors"},
 
 	// --- (D) no bucket, no framework recognition ----------------------------
-	// The reported instance. Arm B of #6537 decides alias-vs-own-bucket and
-	// lands the content; when it does, this entry must be deleted or the guard
-	// fails.
-	{Language: "vbnet", Reason: "D: no rules/vbnet bucket and no VB.NET cross-extractor; measured as 0.0% framework annotation across 45,663 entities on a real WinForms codebase", Issue: "#6537"},
-
+	// vbnet USED to be listed here — the reported instance from #6535. Arm B of
+	// #6537 measured alias-vs-own-bucket (an alias onto csharp emits 0 entities
+	// on VB source) and landed rules/vbnet/frameworks/winforms.yaml, so the gap
+	// is closed and the entry is gone. See vbnet_winforms_rules_6537_test.go.
 	{Language: "assembly", Reason: "D: no rule bucket; no framework layer is expected for assembly"},
 	{Language: "astro", Reason: "D: no rules/astro bucket; Astro is itself the framework, and its integrations are unrecognised"},
 	{Language: "avro", Reason: "D: no rule bucket; schema IDL with no framework layer"},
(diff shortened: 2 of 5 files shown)


cajasmota:
Pushed `3fbfb3e` addressing blocking 1-5. Verdicts below were all scored `go vet` first, then full `go test -count=1 ./internal/engine/` with no `-run` filter; `TestRuleInventory_EveryEmbeddedRuleFileParses` was confirmed absent from every failure set, so none of these is a parse-failure false DEAD.

|---|---|---|---|
| N1 | truncate Operation pattern at `\(` (drop `Handles`) | vet=0 test=0 **SURVIVED** | vet=0 test=1 **DEAD**, 3 assertions |
| N2 | delete `requires_framework: true` on Module | vet=0 test=0 **SURVIVED** | vet=0 test=1 **DEAD** |
| N3a | empty `closedLanguageRuleGaps` | n/a | vet=0 test=1 **DEAD** |
| N3b | delete `rules/vbnet` with the ratchet in place | n/a | vet=0 test=1 **DEAD**, names `vbnet` |
| N4b | narrow HANDLES rel pattern to `\w*Button` (drops `MyBase`) | n/a | vet=0 test=1 **DEAD** |
| N5 | re-add the `Application.Run` Config source_pattern | n/a | vet=0 test=1 **DEAD** |

I reproduced your N1 and N2 survivals on the parent commit before changing anything — both confirmed vet=0 test=0.

**N1.** Fixture now carries three subs with handler-shaped signatures and no `Handles` clause (`InitializeComponent`, `NotAHandler`, `AlsoNotAHandler`), asserted not to be Operations.

**N2.** Added `TestVBNet6537_ModuleGateRequiresFrameworkMarker` with a marker-free negative, a positive control, and — deliberately — an assertion that the comment-only case *does* fire. Your `' see System.Windows.Forms docs` example is exactly right, so the rule file now states the gate's real strength ("excludes only files with no textual mention anywhere") instead of the stronger claim. If the gate is ever tightened, that assertion goes red and points at the stale comment rather than letting it quietly become wrong in the other direction.

**N3.** You were right that it was unreachable, and the fix is the second option you offered: `closedLanguageRuleGaps` in the allowlist file plus `TestLanguageRuleGaps6537_ClosedGapsStayClosed`, which asks `CompiledRuleCount` directly and never consults `stillOpen`. It also rejects a language listed as both open and closed. N3b is the reachability proof — deleting the bucket produces `"vbnet" was recorded as a CLOSED rule gap ... but compiles 0 rules again`.

**N4 — Config/View mismatch.** Reconciled by **removing the entity**, not by renaming it. Emitting the startup form as `View` from `Program.vb` would have minted a second View for a form defined in another file — trading an orphan for a duplicate. The `Application.Run` text is still read; it now lives only inside the INSTANTIATES rule, whose `View:<Form>` target resolves to the entity the form's own file already emits. The test asserts the edge *and* asserts no `Config:<Form>` is emitted, so N5 (re-adding the pattern) is now caught.

**5.** Header wording corrected. The conclusion survives for a different, measured reason — `rules/csharp` is keyed to `csharp` and its regexes are C# syntax, so nothing in it can fire on a `.vb` file — and the header now says that instead of the "explicit deferral" over-read. Rules unchanged.

**Also now asserted:** `Handles MyBase.<Event>` → `Component:MyBase`, which was a rule-file claim no test covered (N4b).

### Non-blocking items

Filed **#6605** for the two I think are real engine limitations rather than local debt:

1. **Rules see raw text** — no comment/string stripping, so `Private Const Doc As String = "Me.Controls.Add(Me.GhostLabel)"` mints a `Component`. Not vbnet-specific; any unanchored call-shaped pattern in any bucket has it. The `import_markers` weakness from N2 is the same root cause, which is why they belong in one issue.
2. **Multi-clause `Handles A.Click, B.Click`** — needs a schema change (repeated-group fan-out on `RelationshipRule`), not a regex change; several other languages have the same shape (stacked Python route decorators, Java `@RequestMapping` path arrays).

The unexercised `file_conventions` globs I left as a comment only — that is test-coverage debt local to this one file, and an issue for it would be noise.

Full package green: `gofmt -l` clean, `go vet` exit 0, `go test -count=1 ./internal/engine/` exit 0.

cajasmota:
Pushed `221f34b`. Reproduced your survivor on the parent first — vet=0, `go test -count=1 ./internal/engine/` **0**, confirmed.

**Verdict: `N6` drop `Application\.Run` from INSTANTIATES — vet=0 test=1 DEAD.** Both the existing positive and the new negative go red; the negative names it directly:

```
INSTANTIATES edge targets View:SqlConnection, which is not a form — the rule is
pairing the Module with an arbitrary constructor rather than with the type
passed to Application.Run
```

`TestRuleInventory_EveryEmbeddedRuleFileParses` absent from the failure set, YAML list marker verified intact, restored by `cp` against a verified shasum.

**Negative case:** `TestVBNet6537_FiresOnEntryPointModule`. `vbModuleSource` now constructs two non-form types, one on each side of the `Application.Run` call, and the test asserts no INSTANTIATES edge targets `View:SqlConnection` or `View:Timer`. The ordering is load-bearing and commented as such — the pattern is non-greedy, so a rule that stops requiring the prefix pairs with the *first* reachable constructor, which is why `SqlConnection` sits ahead of `Application.Run` rather than behind it. Put it behind and your mutant survives again.

You framed this as the same defect class and that is exactly right, with the sides swapped: the Config fix removed an entity no edge could point at; this one removes edges no entity will ever satisfy. Asserting an edge exists catches neither. Only asserting which edges must *not* exist does.

### The `{0,600}` window — my read: real hazard, not a tolerable imprecision

You asked for a judgement; I measured instead, because the answer turned out to be checkable. The window is positional, not structural — it has no notion of `End Module`. A file declaring two Modules, which VB permits and which is not unusual:

```vb
Module DiagnosticsHelpers
    Public Sub Log(ByVal msg As String)
        System.Diagnostics.Debug.WriteLine(msg)
    End Sub
End Module

Module Program
    Public Sub Main()
        Application.Run(New OrderEntryForm())
    End Sub
End Module
```

emits `Module:DiagnosticsHelpers -INSTANTIATES-> View:OrderEntryForm`, and **nothing at all** for `Module:Program`. `FindAllStringSubmatch` is non-overlapping, so the first match consumes through the `Application.Run` and the correct pairing is never reachable. Both halves are wrong — a false edge on one module, a missing edge on another — from entirely well-formed source.

So: **fold it into #6605**, which I have done, with the reproduction. Two things push it past "stated limitation" for me:

1. It is misattribution, not fuzziness. A missing edge is a recall number; an edge attributed to the wrong entity is a graph that answers "who starts this app?" with the wrong name, which is precisely the question this rule exists to answer.
2. It is **not vbnet-specific**. Every `relationship_rules` entry that joins across a window has the exposure. `rules/csharp/frameworks/asp_net_mvc.yaml` joins `class (\w+Controller)` to `public \w+Controller(` across `{0,400}`, and a file with two controllers is entirely ordinary. That is why it belongs in #6605 next to the raw-text item rather than as a vbnet bug — same root cause, a regex layer with no structural awareness, and the plausible fix (a `within` scope on `RelationshipRule`, sourced from the `TSTree` already on `FileInput`) is the same kind of engine change.

I kept the rule rather than dropping it: on the single-module entry-point file, which is the overwhelmingly common shape, it is correct, and it is the only edge tying an application entry point to its startup form. But it is now **pinned rather than silently carried** — `TestVBNet6537_InstantiatesWindowMispairsAcrossModules` asserts the wrong behaviour in *both* directions, so whoever fixes it gets a red test and is forced to update the `KNOWN LIMITATIONS` note in `winforms.yaml` instead of leaving stale prose behind.

Full package green: `gofmt -l` clean, `go vet` 0, `go test -count=1 ./internal/engine/` 0.

Issue #6612, opened 2026-08-25
pythonMaskInertRegions panics on a file ending with a backslash inside an unterminated triple-quoted string — the clamp preventing it is untested


cajasmota:
`pythonMaskInertRegions` **panics** on a Python file that ends with a backslash inside an unterminated triple-quoted string. Reproduced:

```
input: x = """abc\
panic: runtime error: index out of range [11] with length 11
```

This is a crash in the indexer, not a wrong edge. Any Python file of that shape takes the process down.

## The mechanism

In the triple-quoted branch of `internal/engine/http_endpoint_synthesis.go` (~L3166), the scan advances `j += 2` when it sees a backslash. With a backslash as the final byte and no closing delimiter, `j` runs to `len(b)+1`. A clamp catches it:

```go
				if j > len(b) {
					j = len(b)
				}
				for k := i; k < j; k++ {
					blank(k)
				}
```

Remove the clamp and `blank(k)` indexes past the end. **Nothing observes it** — `go vet` 0, full package `go test -count=1 ./internal/engine/` **exit 0**. So the clamp is load-bearing and completely untested.

That is the finding: not that the code is wrong today, but that the single line preventing a panic has no test, and the input that reaches it is exactly the input class we already know is unrepresented in the fixtures.

## Why this is reachable

The same reachability argument #6598 makes applies here, and it applies harder because the failure is a crash:

- partially-written files and generated fragments;
- anything mid-edit when an incremental index fires;
- templates and non-Python content that happens to carry `.py`.

**A file does not have to be valid Python to be indexed.** It does have to not crash the indexer.

## Relationship to the neighbouring issues, since there are now several

These are four distinct holes in and around the same function. Keeping them separate has been deliberate:

- **#6418** — a *well-formed* single-line literal whose contents spell `include_router` mints a mount. Pinned as a known limitation.
- **#6598** — an *unterminated single-line* quote unmasked the rest of the file. Pinned; PR in review.
- **#6596** — the YAML rule bypasses the masker entirely, so it mints from comments and docstrings regardless. Exposure being measured.
- **this** — the *triple-quoted* sibling branch, where the same unterminated-literal shape reaches an unclamped index and panics.

#6598 pinned the single-line branch. This is its sibling, with the same root cause and a worse failure mode.

## What is needed

[...]

Issue #6598, opened 2026-08-25
an unterminated single-line quote makes pythonMaskInertRegions skip to EOF, silently unmasking every comment and docstring after it — and no fixture has one


cajasmota:
`pythonMaskInertRegions` treats an **unterminated** single-line quote as running to end-of-file, so every `#` comment and every triple-quoted docstring *after* it is silently left unmasked. The mount scan then reads all of it as live code.

## The mechanism

`internal/engine/http_endpoint_synthesis.go:3184`, in the single-line-literal branch:

```go
			for j < len(b) {
				if b[j] == '\\' {
					j += 2
					continue
				}
				if b[j] == c || b[j] == '\n' {
					break
				}
				j++
			}
```

The `|| b[j] == '\n'` is the entire protection. It is what stops an unclosed quote from swallowing the file. Remove it and the scan index jumps to EOF, so every subsequent inert region — comments, docstrings, triple-single-quoted blocks — is skipped by the masker rather than blanked.

## Nothing observes it

Deleting `|| b[j] == '\n'` is a clean permissive mutant:

```diff
-				if b[j] == c || b[j] == '\n' {
+				if b[j] == c {
 					break
 				}
```

`go vet` **0**, full package `go test -count=1 ./internal/engine/` **exit 0**. Nothing in the repo notices.

The reason is coverage, not luck: **no fixture in the `_6385` or `_6414` tables contains an unterminated quote.** Every case is well-formed Python, so the guard never has to do its job in a test.

## Why it matters beyond a mutant score

The masker's whole purpose is to stop comments and docstrings from minting endpoints — that is what #6415 fixed. This is a single-character condition standing between that fix and its complete reversal for the remainder of any file containing one unclosed quote.

Unterminated quotes are not exotic in the inputs a graph indexer sees: partially-written files, templates, generated fragments, files with a `\` line-continuation edge case, and anything mid-edit when an incremental index fires. A file does not have to be valid Python to be indexed.

Note also that this is the **narrow** hole. #6596 records that `fastapi.yaml:119` bypasses the masker entirely, so the YAML rule mints from comments and docstrings regardless of any of this. The two together are why the synthesis path's carefulness has not been translating into a clean graph.

## What is needed

1. A test with an **unterminated single-line quote** followed by a commented or docstring'd `include_router`, asserting that the later inert region is still masked. That kills the mutant above and is the whole ask.
[...]

Issue #6596, opened 2026-08-25
YAML relationship rules see raw source, so fastapi.yaml:119 mints ROUTES_TO from comments and docstrings — measured at 4 of 128 matches; pin it as a known limitation


cajasmota:
`pythonMaskInertRegions` has exactly **one** non-test caller — `internal/engine/http_endpoint_synthesis.go:3310`. Every other consumer of Python source reads it raw. In particular the YAML relationship rule at `internal/engine/rules/python/frameworks/fastapi.yaml:119`:

```yaml
  - pattern: "(\\w+)\\.include_router\\s*\\(\\s*(\\w+)"
```

runs against unmasked source, so it mints a `Service -ROUTES_TO-> Route` edge from inside **comments and docstrings** — exactly the cases #6415 fixed on the synthesis path.

## Why this is worse than the limitation it sits next to

#6418 pins a known false positive on the synthesis path: a *single-line string literal* containing `include_router` still emits a mount point, because the masker deliberately leaves single-line literals intact so a `prefix="/x"` value stays readable. That is a bounded, deliberate trade-off, and it is now pinned by a test.

The YAML path has no such trade-off, because it has no masking at all:

| construct | synthesis path (`:3310`, masked) | YAML rule (`fastapi.yaml:119`, raw) |
|---|---|---|
| `# app.include_router(r)` in a comment | excluded | **emits** |
| `"""app.include_router(r)"""` docstring | excluded | **emits** |
| `'''…'''` triple-single-quoted | excluded | **emits** |
| commented-out import block | excluded | **emits** |
| single-line `"app.include_router(r, prefix='/x')"` | emits (#6418, known) | emits |

So the narrow, deliberate hole is documented and tested, while the wide, undeliberate one is neither.

## Why it has stayed invisible

The edge minted by the YAML rule and the edge minted by the synthesis path are not the same record, so the synthesis path's exclusions do not protect the graph — they only protect *its own* output. Anything reading `ROUTES_TO` sees the YAML rule's edges regardless of how careful the synthesis path is.

This also explains why #6415's fix looked complete: it was measured on the synthesis path, which is genuinely clean afterwards. Nothing measured the rule.

## What needs deciding

1. **Should the YAML rule be masked at all?** Rule evaluation is generic across every language; `pythonMaskInertRegions` is Python-specific. There may be no clean seam, in which case the honest answer is that YAML relationship rules cannot exclude inert regions and the rule should be narrowed or removed rather than pretended to be safe.
[...]

Issue #6418, opened 2026-08-20
fastapi mount scan: a single-line string containing include_router still emits — the limitation is deliberate but nothing pins it


cajasmota:
Follow-up from the #6415 review. The masker added there blanks `#` comments and triple-quoted regions before scanning for `include_router`, so a mount inside a docstring no longer becomes graph data. **Single-line string literals are deliberately not masked**, so this still emits:

```python
X = "app.include_router(r, prefix='/x')"
```

The omission is intentional and the reason is real: the prefix *value* has to be readable out of the masked copy, so short literals are skipped over rather than blanked. Blanking them would break the feature. A full Python tokenizer is the only clean fix and is not cheap.

**The part actually worth fixing is that no test pins this either way.** The behaviour is currently accidental — nothing would notice if it changed in either direction, and nothing tells a future reader it was a decision rather than an oversight. Given that the whole point of the #6415 review was that an untested branch is trusted anyway, leaving this one unpinned reproduces the same mistake in miniature.

## What to do

Add a test asserting the **current** behaviour, with a comment stating it is a known limitation rather than a desired property. That is the honest shape: it documents the trade-off, it fails loudly if someone changes the masker in a way that accidentally covers or worsens this case, and it does not pretend the limitation is a feature.

Do **not** attempt the tokenizer. If someone later wants the real fix, this test becomes the thing they update, which is exactly the signal that should be there.

## Assessing the exposure first

Worth a quick measurement before writing anything: how often does a full `include_router(...)` call, with a literal `prefix=`, appear inside a single-line string in real Python? My guess is close to never outside of test fixtures and code-generation templates — but that is a guess, and if it is right, the test-only outcome above is clearly correct. If it turns out to be common in generated code, this changes shape and should be re-scoped.

The false positive it produces is bounded: it mints a `url_mount_point` synthetic with a prefix that does not correspond to a real mount. Per #6417 that pollutes the per-repo prefix set and can only mint a link if a real producer happens to sit under that prefix. Bad, but narrow.

Refs #6385, #6415, #6417.

cajasmota:
**Re-grounded against `main`. Premise HOLDS — no comments, nothing re-scoped, and nothing in the body is wrong.**

[...]

PR #6622, opened 2026-08-25
fix(#6614): an unterminated string literal's own tail reached the mount scan as live code

### LINKED ISSUE #6614: the contents of an unterminated single-line literal stay live and reach the mount scan — the choice is unrecorded and blanking them passes the whole suite
`pythonMaskInertRegions` leaves the **contents of an unterminated single-line literal** live — the text between the opening quote and the end of that line is passed to the mount scan as code. Nothing observes that choice, and it is not obvious it is the right one.

## The surviving mutant

Blanking the literal when it turns out to be unterminated (the terminated case untouched):

```diff
 			if j < len(b) && b[j] == c {
 				j++
+			} else {
+				for k := i; k < j; k++ {
+					blank(k)
+				}
 			}
```

`go vet` **0**, full package `go test -count=1 ./internal/engine/` **exit 0**. Nothing notices either behaviour.

## Why this is a real question and not a nit

Consider:

```python
x = "oops forgot to close it, app.include_router(other.router, prefix="/x")
```

- **Today**: the literal is unterminated, so the branch ends it at the newline and does *not* blank it. The `include_router(...)` text inside the broken string is live, and the mount scan reads it. A mount point is minted from the contents of a string.
- **Under the mutant**: the runaway span is blanked, and nothing is minted.

Both are defensible. Leaving it live matches the deliberate #6418 trade-off — single-line literals are left intact *on purpose*, because the mount prefix itself lives inside one (`prefix="/x"`), and blanking short literals would blank the value the scan needs to read. Blanking it matches the intuition that an unterminated literal is broken source whose contents should not be trusted as code.

What is not defensible is that the choice is unrecorded. The function's own doc comment explains why *terminated* single-line literals are left intact; it says nothing about the unterminated case, which reaches a different code path.

## Relationship to the neighbours

This is the fourth distinct hole in or around this function, and they are deliberately separate:

- **#6418** — a *terminated* single-line literal whose contents spell `include_router` mints a mount. Pinned as a known limitation.
- **#6598** — an unterminated quote unmasked everything *after* it. Fixed and pinned (PR #6611).
- **#6612** — the *triple-quoted* sibling branch panics on a trailing backslash at EOF; the clamp preventing it is untested.
- **this** — the unterminated literal's *own line contents*, which #6598's test does not cover because its assertions are about the regions that follow.

#6611's test is sound and this does not undermine it: that test pins what happens *after* the quote, and this mutant changes only what happens *within* the runaway span.

## What is needed

1. **Decide** which behaviour is intended. My inclination is to keep the current one — it is consistent with #6418, and blanking would make an unterminated quote silently swallow a legitimate `prefix=` on the same line, which is the failure #6418 exists to avoid — but that is a call to make deliberately, not by default.
2. **Pin it**, so the mutant above dies and the decision is recorded where the next reader will find it.
3. **Say it in the doc comment.** The comment currently explains the terminated case only.

Cheap: the #6598 fixture already carries an unterminated quote; this needs one more assertion about the text on that same line.

## Trap

Do not "fix" this by blanking without checking the `prefix=` interaction. A `prefix="/x"` sitting after an unterminated quote **on the same line** would be blanked too, and #6418 records that the masker deliberately preserves single-line literals precisely so that value stays readable.

## Provenance

Found by a coordinator mutant against PR #6611 while verifying #6598. Pre-existing on `main`; the PR touches no non-test code.

Refs #6598, #6418, #6612.

cajasmota:
`pythonMaskInertRegions` ended an unterminated single-line literal at the newline but did not blank the span it had just walked, so everything from a runaway opening quote to the end of that line was handed to the FastAPI mount scan as code. A broken line whose string contents happen to spell `include_router(..., prefix=...)` minted a `url_mount_point` for a mount that does not exist.

Once a quote opens and never closes, everything to the end of the line IS string content by definition - there is no legitimate `prefix=` "after" it on the same line for blanking to swallow. So the change is strictly one-directional: leaving the tail live can only mint false mounts, blanking it can only remove them. This is the same direction #6598 took for the regions AFTER a runaway quote, applied now to the runaway span itself.

The TERMINATED path is untouched, which is the load-bearing constraint: #6418 records that a closed single-line literal must stay INTACT because the mount prefix the scan reads lives inside one. It appears as unmodified context in the diff. #6418's own test file is untouched and green.

## What the new test actually detects

Stated plainly, because three dead mutants should not read as three pieces of evidence:

- **M1 is the whole of the new test's unique kill power.** Reverting to "leave the tail live" survives the pre-existing suite (exit 0) and dies only against this test.
- **M2 and M3 were already dead before this test existed** - M2 by 10+ pre-existing tests, M3 by #6598's. They are honest probes of the dangerous direction, but they are not this test's detection.
- **The two terminated controls duplicate existing coverage.** They are documentation - they make #6418's pin observably unmoved by this change rather than merely believed to be. They are not new detection.

Per-clause, on the premise guards: c2 (exactly one `"` on the line) and c3 (`include_router` on that line) are **load-bearing**; c0 and c1 are structural anchors that admit no new inputs when dropped - decorative for kill power.

## Bounds introduced by the fix, now pinned

The second commit is test-only; the implementation is byte-identical to the first. Three bounds inside the new blanking branch were live and unobserved - the fix had inherited the defect family it was written next to:

| Mutant | BEFORE (reviewed test) | AFTER |
|---|---|---|
| M5 drop the `j > len(b)` clamp | vet 0, test 0 **SURVIVED** | vet 0, test 1 **DEAD** (panic) |
| N1 blank from `i+1` | vet 0, test 0 **SURVIVED** | vet 0, test 1 **DEAD** (span) |
| N2 blank to `j-1` | vet 0, test 0 **SURVIVED** | vet 0, test 1 **DEAD** (span) |

M5 is a real panic path: a literal ending in a backslash at EOF pushes `j` one past the buffer via the escape skip, the literal is then unterminated, and the blanking loop indexes out of range. N1 leaves an unbalanced opening quote in the masked output - the exact shape that caused #6598, reintroduced inside the fix for its sibling.

N1 and N2 are pinned by one byte-exact assertion rather than by fixture gymnastics: the masked runaway span must be ALL spaces, from the opening quote through the last byte before the newline. A `Contains` check for `include_router` is blind to both off-by-ones; a whole-span equality is not.

M1 remains DEAD against the hardened test (vet 0, test exit 1).

## Guard hole closed

Same class as #6620's. The premise guards inspected the fixture only from the anchor RIGHTWARD, so a `#` to the LEFT of the opening quote (`BROKEN = x#= "oops ...`) would route the whole physical line through the COMMENT branch - blanked under every implementation - leaving the fixture vacuous while every clause reported healthy. An enumeration found 170 such inputs. The shipped fixture has no `#` and was never affected; this is hardening. Verified by construction: injecting one left of the anchor makes the new clause `Fatal`.

## Out of scope

- **M4** (a quote escaping a consumed `#` comment) survives, is pre-existing on the parent, and is tracked as **#6623**.
- **#6418** (terminated literal mints), **#6596** (YAML rules bypass the masker), **#6612/#6617** (index bounds) - unchanged.
- No Python tokenizer: #6418 records that as the real fix for this family and as deliberately out of scope.

## Verification

`go test -count=1 ./internal/engine/` exit 0, full package, no `-run` filter. Goldens re-run and unchanged: `./internal/quality/...` and `./cmd/grafel/` exit 0, including the only `url_mount_point` fixture. No baseline was re-recorded or hand-edited.

Closes #6614

### DIFF
--- a/internal/engine/http_endpoint_synthesis.go
+++ b/internal/engine/http_endpoint_synthesis.go
@@ -3134,9 +3134,25 @@ func fastapiResolveMountedRouter(arg string, imports map[string]pyImportBinding)
 // pythonMaskInertRegions returns a copy of src of the SAME byte length, with
 // the same newline positions, in which the contents of `#` comments and of
 // triple-quoted string literals (docstrings, embedded samples) are replaced by
-// spaces. Single-line string literals are left INTACT so a `prefix="/x"` value
-// can still be read out of the masked copy, and byte offsets in the result
-// address exactly the same source positions as in the original.
+// spaces. Byte offsets in the result address exactly the same source positions
+// as in the original.
+//
+// Single-line string literals are handled in two ways, and the split is
+// deliberate:
+//
+//   - TERMINATED — left INTACT, so a `prefix="/x"` value can still be read out
+//     of the masked copy. That is load-bearing: the mount prefix the scan
+//     reads lives inside a closed literal. The cost is #6418 — a closed
+//     literal whose contents merely spell a call is not inert to the scan and
+//     mints a mount that does not exist. That is a recorded known limitation,
+//     fixable only by a real Python tokenizer.
+//   - UNTERMINATED — BLANKED from the opening quote to the end of the line
+//     (#6614). Once a quote opens and never closes, everything to the end of
+//     the line IS string content, so there is no legitimate `prefix=` after it
+//     on that line for blanking to swallow; leaving the tail live could only
+//     mint false mounts from the inside of a broken string. This is the same
+//     direction as #6598, which stopped a runaway quote from unmasking the
+//     regions that FOLLOW it.
 func pythonMaskInertRegions(src string) string {
 	b := []byte(src)
 	blank := func(i int) {
@@ -3174,7 +3190,9 @@ func pythonMaskInertRegions(src string) string {
 				i = j
 				continue
 			}
-			// Single-line literal: skipped over, not blanked.
+			// Single-line literal: terminated ones are skipped over and
+			// left intact (#6418); an unterminated one is string content to
+			// the end of the line, so its span is blanked (#6614).
 			j := i + 1
 			for j < len(b) {
 				if b[j] == '\\' {
@@ -3188,6 +3206,13 @@ func pythonMaskInertRegions(src string) string {
 			}
 			if j < len(b) && b[j] == c {
 				j++
+			} else {
+				if j > len(b) {
+					j = len(b)
+				}
+				for k := i; k < j; k++ {
+					blank(k)
+				}
 			}
 			if j > len(b) {
 				j = len(b)
--- a/internal/engine/http_endpoint_synthesis_python_mask_6614_test.go
+++ b/internal/engine/http_endpoint_synthesis_python_mask_6614_test.go
@@ -57,15 +57,59 @@ BROKEN = "oops app.include_router(other.router, prefix='/tailofbroken')
 		t.Fatalf("#6614: the contested include_router must sit INSIDE the runaway literal, on the same line, "+
 			"or this case is vacuous: %q", line)
 	}
+	// A `#` anywhere LEFT of the opening quote sends the whole physical line
+	// through the COMMENT branch, which blanks it under both implementations —
+	// so the fixture would satisfy every assertion below while observing
+	// nothing about the literal branch at all. The guards above only inspect
+	// from the anchor rightward and cannot see that; this one looks left.
+	if lineStart := strings.LastIndexByte(broken[:q], '\n') + 1; strings.ContainsRune(broken[lineStart:q], '#') {
+		t.Fatalf("#6614: a `#` sits LEFT of the opening quote, so this line is masked by the COMMENT branch "+
+			"under every implementation and observes nothing about unterminated literals: %q", broken[lineStart:q+nl])
+	}
 
-	if masked := pythonMaskInertRegions(broken); strings.Contains(masked, "include_router") {
+	masked := pythonMaskInertRegions(broken)
+	if len(masked) != len(broken) {
+		t.Fatalf("#6614: masking changed the byte length: got %d, want %d", len(masked), len(broken))
+	}
+	if strings.Contains(masked, "include_router") {
 		t.Errorf("#6614: the tail of an unterminated single-line literal reached the masked copy as live code "+
 			"— everything from the opening quote to the end of the line is string content and must be blanked. "+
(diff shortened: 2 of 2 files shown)


cajasmota:
Merging. Two coordinator mutants run against the hardened test; one DEAD, one recorded equivalent with its reason.

## DEAD — the newline subclass

Nobody had run a mutant that keeps the blanking but narrows *when* it applies. M1 disabled it outright, so it could not tell a subclass hole from a total one.

```diff
-			} else {
+			} else if j >= len(b) {
```

Blanking then happens only for a literal unterminated at **EOF**, and a literal unterminated at a **newline** — the ordinary case, a broken line in the middle of a file — keeps its tail live. `go vet` 0, full package `go test -count=1 ./internal/engine/` **exit 1**, failing exactly `TestSynth_FastAPI_MountPrefix_UnterminatedLiteralTailIsBlanked_6614` and nothing else.

## Equivalent, with the reason — the mixed-quote terminator

I also tried making the terminator check accept either quote character regardless of which opened the literal:

```diff
-			if j < len(b) && b[j] == c {
+			if j < len(b) && (b[j] == '"' || b[j] == '\'') {
```

`go vet` 0, full package **exit 0** — survived. It is **equivalent**, not a hole, and the reason is worth writing down because it is not obvious from the diff: the scan loop above breaks only on `b[j] == c` or `b[j] == '\n'`, so at the check `b[j]` can never be the *other* quote character. The added disjunct is unreachable.

I probed it rather than argued it. First probe fixture was wrong in an instructive way — `x = "abc' app.include_router(r, prefix="/p")` masks identically under both, but because the later `"` in `prefix="` closes the literal, not because of anything the mutant does. Only the second fixture, `x = "abc' app.include_router(r, prefix=p)` with no further `"` on the line, isolates the branch; both versions blank from the opening quote to end of line, byte for byte. Recording it as equivalent, never as DEAD.

## On the rest

The three bounds the fix shipped in its own new code (M5 clamp, N1 `i+1`, N2 `j-1`) are now each reproduced as a survivor first and each dies with its own message. Pinning N1/N2 with one byte-exact span equality rather than by reshaping the fixture is the better call — it states the behaviour instead of encoding it in punctuation, and it kills both off-by-ones whatever the line ends in.

The corrected body is what I wanted: it says plainly that the unique kill power rests solely on M1, that M2 and M3 were already dead, and that the terminated controls are documentation rather than new detection. c0/c1 recorded as decorative for kill power is the same honesty as #6625's C5.

The `__pycache__` leakage is real and separate — filing it.

Issue #6624, opened 2026-08-25
the backslash-escape skip in the single-line literal branch is unobserved — j += 2 vs j += 1 passes the whole suite (prove non-equivalence first)


cajasmota:
The backslash-escape skip in `pythonMaskInertRegions`'s **single-line literal** branch is unobserved. Changing how far it skips passes the whole suite:

```diff
 			for j < len(b) {
 				if b[j] == '\\' {
-					j += 2
+					j += 1
 					continue
 				}
```

`go vet` **0**, full package `go test -count=1 ./internal/engine/` **exit 0**. Nothing notices.

**Pre-existing on `main`.** Found by a coordinator mutant against PR #6620 (#6617), which touches no non-test code.

## FIRST: prove non-equivalence, or close this

I have **not** probed this. The reasoning is that `j += 2` exists to step *over* an escaped character, so `j += 1` should let an escaped quote (`\"`) be read as a terminator on the next iteration, ending the literal early and leaving the rest of the line live — e.g.

```python
x = "a\"b" and then app.include_router(r, prefix="/x")
```

But that is an argument, not a measurement, and this milestone has repeatedly shown arguments of exactly this shape to be wrong (#6613's annotated-tag fixture, #6592's "always a synthetic ID", #6614's non-existent trap — all reasoned, all false).

**So step 1 is a probe**, not a test: construct an input where the two skips produce different masked output, and show it. If no such input exists, this mutant is **equivalent under the current suite** and possibly outright equivalent — record that with the reason and close the issue. A recorded negative result is a complete outcome here, exactly as `AnnotatedTagCandidateIsNotADistinguisher` was for #6613.

Only if the probe succeeds does step 2 apply: pin it.

## If it is non-equivalent

Pin it with the discipline this function's other holes now use:

1. **Score the mutant before and after** — exit 0 on pristine `main`, exit 1 after.
2. **Premise-guard the fixture so the branch under test actually runs**, not merely that the input contains a backslash. Note the existing rows in `TestPythonMaskInertRegions_MustNotPanic` deliberately **forbid** backslashes in their fixtures (rows A and D) or require an **odd** trailing run (row C) — so this fixture is in tension with all three and needs its own guard reasoning, not a copy.
[...]

Issue #6617, opened 2026-08-25
two more unobserved index bounds in pythonMaskInertRegions panic the indexer — x = "" and x = """abc"" both crash when weakened


cajasmota:
Two more index bounds in `pythonMaskInertRegions` are load-bearing anti-panic guards that **no test observes**. Both were proved non-equivalent with probes — weakening either crashes the indexer on ordinary input.

This is the same family as **#6612** (the triple-quote clamp), found while reviewing its fix. That one is now pinned; these two are not.

## The two bounds

**`internal/engine/http_endpoint_synthesis.go:3155`**

```diff
-			if i+2 < len(b) && …
+			if i+2 <= len(b) && …
```

`go vet` **0**, full package `go test -count=1 ./internal/engine/` **exit 0**. Probe: input `x = ""` →

```
panic: runtime error: index out of range [6] with length 6
```

**`internal/engine/http_endpoint_synthesis.go:3162`**

```diff
-				if b[j] == c && j+2 < len(b) && …
+				if b[j] == c && j+2 <= len(b) && …
```

`go vet` **0**, full package **exit 0**. Probe: input `x = """abc""` →

```
panic: runtime error: index out of range [12] with length 12
```

Neither input is exotic. `x = ""` is an ordinary empty-string assignment. `x = """abc""` is a triple-quoted string closed with two quotes instead of three — a typo, or a partially-written file.

## Why they matter together

`pythonMaskInertRegions` now has **three** known unobserved bounds, of which one is pinned:

| bound | shape | status |
|---|---|---|
| triple-quote clamp `j > len(b)` | trailing backslash in an unterminated `"""` | **pinned** (#6612, PR #6615) |
| `i+2 < len(b)` at `:3155` | `x = ""` | **this issue** |
| `j+2 < len(b)` at `:3162` | `x = """abc""` | **this issue** |

The pattern is consistent and worth stating: this function walks a byte slice with several look-ahead offsets, and **every fixture in the suite is well-formed Python**, so no bound is ever approached. The failure mode is a crash in the indexer, not a wrong edge — and a file does not have to be valid Python to be indexed.

## What is needed

1. A test per bound, with the probe input above, asserting `pythonMaskInertRegions` returns normally. Weakening the bound must make it panic.
2. **Score each mutant before and after** — exit 0 on pristine `main` (survivor reproduced) and exit 1 after. A kill not first reproduced as a survivor is not evidence.
[...]

PR #6625, opened 2026-08-25
test(#6623): a quote inside a `#` comment escaped the masker with the suite green

### LINKED ISSUE #6623: a quote inside an already-consumed # comment can escape into the string branch of pythonMaskInertRegions, and nothing observes it
`pythonMaskInertRegions`'s **`#`-comment branch** has an unobserved hole: a mutant that lets a quote inside an already-consumed comment escape into the string branch passes the whole suite.

`go vet` **0**, full package `go test -count=1 ./internal/engine/` **exit 0**. Under it, a *terminated* quoted span inside a `#` comment is left live rather than blanked, so its contents reach the mount scan.

**Pre-existing** — confirmed to survive identically on `bce761f1e`, before the #6614 change that surfaced it.

## Why it is filed separately

#6614 changed the **literal** branch (blanking the tail of an unterminated single-line literal). This is the **comment** branch, which that change does not touch, so folding it in would have widened a behaviour-change PR into an unrelated one. The agent implementing #6614 reported it as a survivor rather than fixing it, which was the right call.

## Where this sits in the family

This is now the fifth item in and around one 60-line function, and worth listing so the next reader sees the shape rather than five separate reports:

| item | branch | status |
|---|---|---|
| #6418 | terminated single-line literal mints a mount | known limitation, pinned |
| #6598 | unterminated quote unmasked everything after it | fixed, pinned |
| #6612 / #6617 | four unobserved look-ahead index bounds (`:3155`, `:3162`, `:3168`, `:3189`) | pinned / in review |
| #6614 | unterminated literal's own tail stayed live | fixed |
| **this** | **`#`-comment branch: a quote inside a consumed comment escapes** | **unobserved** |

The uniform cause across all of them: the function walks a byte slice with several look-ahead offsets and branch dispatches, and **every fixture in the suite is well-formed Python**, so no branch boundary is ever approached.

## What is needed

1. A test pinning that a quoted span inside a `#` comment is blanked, so the mutant above dies.
2. **Score it before and after** — exit 0 on pristine `main` (survivor reproduced), exit 1 after. A kill not first reproduced as a survivor is not evidence.
3. A premise guard that asserts **the comment branch actually runs**, not merely that the fixture contains a `#`. This exact class has now failed three times here (#6611, #6615, #6620): a guard checking the fixture's *shape* while the branch under test is never entered.
4. **Enumerate rather than hand-pick.** #6620's review brute-forced every string ≤7 bytes over `{x, space, =, ", ', #, \, a, \n}`, scoring guard-pass × mutant-survival, and found two holes that three rounds of hand-crafted attacks had missed. The space here is the same size. Do that, and report the per-clause drop analysis so decorative guard clauses are not claimed as load-bearing.

## Traps

2. If the fix belongs in the shared table-driven panic test from #6617, put it there rather than starting a sixth idiom — but note that test asserts *"must not panic"*, and this is a *content* claim, so it may genuinely need its own home. Decide deliberately and say why.
3. The `:3192` single-line clamp is **equivalent under the current suite**, verified three times with the reason. Record it as equivalent if you mention it; never as DEAD.

## Provenance

Found by the implementing agent on PR #6622 (#6614) while probing the dangerous direction of its own change, and confirmed pre-existing on the parent commit.

Refs #6614, #6617, #6418.

cajasmota:
`pythonMaskInertRegions`'s `#`-comment branch blanks the rest of the line. Nothing in the package observed that: a mutant stopping the comment scan at a quote —

```go
for ; i < len(b) && b[i] != '\n' && b[i] != '"' && b[i] != '\''; i++ {
```

— lets the quoted span escape into the single-line-string branch, which **skips** rather than blanks. A commented-out `app.include_router(legacy, prefix="/ghost")` then keeps `"/ghost"` live for the mount scan. `go vet` 0, full package `go test -count=1` **exit 0**. Pre-existing.

**No behaviour change.** `internal/engine/http_endpoint_synthesis.go` is byte-identical to `main`: `2081d244d0f9ae4040b20f80c9c9bdb56faf2511`.

## Review round 2 — the escape hatch is per quote character

The first revision pinned only half its own premise. The **narrower** mutant, adding only the single-quote disjunct —

```go
for ; i < len(b) && b[i] != '\n' && b[i] != '\''; i++ {
```

— was `go vet` 0 and full-package `go test -count=1 ./internal/engine/` **exit 0: a survivor**, because the fixture carried a double-quoted literal only. After a PR whose whole premise is *"a quote inside a comment escapes"*, the `'` half of that premise was unobserved.

The fixture now carries a literal of **each** kind:

```python
# app.include_router(legacy, prefix="/ghost", tags=['ghost'])
```

and the premise guard requires both, per quote character, failing loudly if a later edit drops either half.

## Scoring — three mutants, each reproduced as a survivor first

| run | `go vet` | `go test -count=1 ./internal/engine/` | verdict |
|---|---|---|---|
| pristine + new test | 0 | **0** | green |
| `"` and `'` (the pair), test parked | 0 | **0** | SURVIVED |
| `"` and `'` (the pair) + new test | 0 | **1** | **DEAD** — `byte 36 … survived masking as '"'` |
| `"` only, test parked | 0 | **0** | SURVIVED |
| `"` only + new test | 0 | **1** | **DEAD** — `byte 36 … survived masking as '"'` |
| `'` only, test parked | 0 | **0** | SURVIVED |
| `'` only + new test | 0 | **1** | **DEAD** — `byte 52 … survived masking as '\''` |

The fix does not trade one half for the other: the `"`-only mutant is still dead.

## The kill does not rest on a length bound

The review's analytic argument is correct and the header now carries it, replacing the sweep as the load-bearing evidence. Once every clause passes, each mutant dies for input of **arbitrary** length:

- **C1+C2** put the scanner at the `#`. `h` is the *first* `#`, and every byte before it is neither `#` nor a quote, so each takes the switch's `default` arm and the cursor arrives at `h` with the comment branch selected.
- **C3** gives the mutant a byte to halt on: let `q` be the first quote inside the comment that the mutant's disjunct names. The loop blanks `[h,q)` and stops with the cursor **on** `q`, which the switch dispatches into the quote arm.
- **C4** says `q` does not open a triple — and the mutant has rewritten only bytes *before* `q`, so the look-ahead reads the source unchanged — so the arm taken is the single-line one, which **skips**. The cursor only advances, so nothing writes to `q` afterwards. `got[q]` is still a quote, `h <= q < eol`, and the assertion fails. Dead.
- Symmetrically on pristine, the comment loop blanks every byte of `[h,eol)`; `blank()` rewrites everything but `\n` and `\r`; the region holds no `\n` by construction and no `\r` by **C5**. All spaces, assertion passes.

Neither half mentions `len(src)`. I found no gap in it.

## Enumeration — re-run against the changed guard

C3 and C4 now score **per quote character** rather than over the first quote of either kind, so the guard did move and the previous sweep does not carry over. Re-run against the new guard, and against all three mutants: every string of length ≤ 9 over `{x, space, =, ", ', #, \, a, \n, \r}` — **1,111,111,111 inputs, 29,304,608 passing the guard** — scored for guard-pass × mutant-survival and guard-pass × pristine-failure.

**Zero survivors for every mutant, zero pristine failures.**

Per-clause drop analysis over the ≤ 7 prefix (11,111,111 inputs, 215,784 passing):

| clause dropped | passing | `"`+`'` alive | `"` only alive | `'` only alive | first witness | + pristine fails | verdict |
|---|---|---|---|---|---|---|---|
| C1 a `#` exists | 617,556 | 2,832 | 2,832 | 2,832 | `"""xxx'` | 398,940 (`xxxxx"'`) | LOAD-BEARING |
| C2 no quote before the `#` | 281,560 | 156 | 156 | 156 | `x"""#"'` | 56,932 (`xxx"#"'`) | LOAD-BEARING |
| C3a a `"` inside the comment | 650,502 | 0 | **434,718** | 0 | `xxxxx#'` | 0 | LOAD-BEARING |
| C3b a `'` inside the comment | 650,502 | 0 | 0 | **434,718** | `xxxxx#"` | 0 | LOAD-BEARING |
| C4 that quote is not a triple opener | 218,018 | 1,118 | 1,118 | 1,118 | `xx#"""'` / `xx#"'''` | 0 | LOAD-BEARING |
| C5 no CR in the region | 314,448 | **0** | **0** | **0** | — | 98,664 (`xxx#"'\r`) | load-bearing for the assertion, **admits no mutant survivor** |

C3a and C3b are the new rows and they are exactly symmetric — which is the point: dropping either one leaves the corresponding half's mutant alive by 434,718 inputs. C5 is reported honestly as before: it keeps the all-spaces assertion from failing on **correct** code (`blank()` preserves `\r`); it does not contribute to killing anything.

## Controls

All four on pristine source, all exit 1 with the premise message rather than passing silently:

- `'ghost'` removed → `… the comment "…" contains no '\'', so the mutant that stops the comment scan at '\'' has nothing to escape through …`
- `"/ghost"` removed → the same message for `'"'`
- no quote at all in the comment → the same message for `'"'`
- `x = "q"  ` prepended before the `#` → `… the source before the `#` at 9 must contain no quote, or the string branch consumes the `#` and the COMMENT branch never runs …`

## Where it lives, and the cross-reference

New file `internal/engine/http_endpoint_synthesis_python_mask_comment_test.go`, **not** a fifth row in `..._python_mask_panic_test.go` (#6617). That table pins four look-ahead **bounds** against a panic, and its only post-condition is `len(got) == len(src)`. This is a **content** claim about produced bytes. A row asserting content would smuggle a differently-shaped assertion through a table whose entire contract is the crash.

The previous body claimed "the header of each file cross-references the other." **That was false** — only the new file pointed at the panic test; the panic test was untouched and named neither #6623 nor this test. Resolved by making it true rather than by softening the sentence: `..._python_mask_panic_test.go` now carries the reverse pointer, so a reader landing in the panic table is told where content claims about this function belong. The reference is now mutual **in the tree**, which is the only place it counts.

`gofmt -l internal/engine/` clean.

## Not touched

#6418 (terminated literal outside a comment mints a mount — known limitation), #6596, #6614 (PR #6622), #6624 (the `j += 2` backslash skip), and the **backslash disjunct in this same loop** — a second pre-existing survivor found by the review (witness `# a \ prefix="/ghost"`), filed separately and deliberately left alone here. The `:3192` single-line clamp remains **equivalent** under this suite; it is not scored here.

Closes #6623

### DIFF
--- a/internal/engine/http_endpoint_synthesis_python_mask_comment_test.go
+++ b/internal/engine/http_endpoint_synthesis_python_mask_comment_test.go
@@ -14,6 +14,14 @@ import (
 // `app.include_router(..., prefix="/ghost")` keeps its literal live and the
 // value reaches the mount scan.
 //
+// The fixture carries BOTH quote kinds on purpose. The escape hatch is per
+// quote character, so there are three mutants, not one: the disjunct naming
+// the double quote, the disjunct naming the single quote, and the pair. A
+// fixture holding only a double quote leaves the single-quote half of the very
+// premise this test states unobserved. The guard below therefore requires a
+// quote of each kind inside the comment and fails loudly if a later edit drops
+// either.
+//
 // This lives beside TestPythonMaskInertRegions_MustNotPanic (#6617) rather than
 // as a fifth row IN it deliberately: that table pins "must not panic" for four
 // look-ahead BOUNDS and asserts only that the output length is unchanged. A row
@@ -21,29 +29,51 @@ import (
 // assertion through a table whose whole contract is the crash. Same function,
 // different claim, own home.
 //
-// Guard clauses are annotated from a brute-force per-clause drop analysis over
-// every string of length <= 7 on the alphabet {x, space, =, ", ', #, \, a, \n,
-// \r} (11,111,111 inputs), scoring guard-pass x mutant-survival and guard-pass x
-// pristine-failure. With every clause on: ZERO inputs pass the guard and leave
-// the mutant alive, and ZERO pass the guard and fail on pristine. Each clause
-// below carries the witness its removal admits.
+// The kill does not rest on a bounded search. Once every clause below passes,
+// each mutant dies for input of ARBITRARY length, by construction:
+//
+//	C1+C2 put the scanner at the `#`. `h` is the FIRST `#`, and every byte
+//	before it is neither `#` nor a quote, so each takes the switch's default
+//	arm and the cursor arrives at h with the COMMENT branch selected.
+//	C3 gives the mutant a byte to stop on: let q be the first quote inside the
+//	comment that the mutant's disjunct names. The loop blanks [h,q) and halts
+//	with the cursor ON q, which the switch then dispatches into the quote arm.
+//	C4 says q does not open a triple — and the mutant has only rewritten bytes
+//	BEFORE q, so the look-ahead sees the source unchanged — so the arm taken is
+//	the single-line one, which SKIPS. Nothing writes to q afterwards, since the
+//	cursor only advances. got[q] is therefore still a quote, h <= q < eol, and
+//	the assertion below fails. Dead.
+//	Symmetrically on pristine: the comment loop blanks every byte of [h,eol),
+//	blank() rewrites everything except `\n` and `\r`, the region holds no `\n`
+//	by construction and no `\r` by C5, so the region comes out all spaces and
+//	the assertion passes. Neither argument mentions the length of src.
+//
+// The brute-force sweep below corroborates that argument rather than carrying
+// it: every string of length <= 9 over the alphabet {x, space, =, ", ', #, \,
+// a, \n, \r} — 1,111,111,111 inputs, 29,304,608 of which pass the guard —
+// scored for guard-pass x mutant-survival and guard-pass x pristine-failure,
+// against all three mutants. Zero survivors and zero pristine failures for
+// every one of them. The per-clause witnesses annotated below are counted over
+// the <= 7 prefix of that sweep (11,111,111 inputs, 215,784 passing).
 //
 // Refs #6623. Out of scope and separately tracked: #6418 (a terminated literal
 // OUTSIDE a comment still mints a mount — a known limitation, not touched
 // here), #6614 (the unterminated literal's own tail), #6624 (the `j += 2`
-// backslash skip).
+// backslash skip), and the backslash disjunct in this same loop, which is a
+// distinct pre-existing survivor filed on its own.
 func TestPythonMaskInertRegions_QuotedSpanInsideCommentIsBlanked(t *testing.T) {
 	// A commented-out mount, the shape that makes the mutant's effect concrete:
-	// under it `"/ghost"` survives masking and is read as a live prefix.
-	const src = `# app.include_router(legacy, prefix="/ghost")`
+	// under it `"/ghost"` (or `'ghost'`) survives masking and is read as a live
+	// prefix. Both quote kinds appear so that all three mutants are scored.
+	const src = `# app.include_router(legacy, prefix="/ghost", tags=['ghost'])`
 
 	// --- premise -----------------------------------------------------------
 	// The claim is about the COMMENT branch, so the premise must establish that
 	// the branch RUNS, not merely that the fixture contains a `#`.
 
-	// LOAD-BEARING. Without it any fixture with no `#` at all — `x` — leaves an
-	// empty region and the assertion below observes nothing (5,380,840 such
-	// survivors in the enumeration).
+	// LOAD-BEARING. Without it any fixture with no `#` at all — `"""xxx'` —
+	// leaves the region empty or misplaced and the assertion observes nothing
+	// (2,832 survivors, and 398,940 pristine failures, in the enumeration).
 	h := strings.IndexByte(src, '#')
 	if h < 0 {
 		t.Fatalf("#6623 premise broken: fixture %q contains no `#`, so the comment branch "+
@@ -52,11 +82,11 @@ func TestPythonMaskInertRegions_QuotedSpanInsideCommentIsBlanked(t *testing.T) {
 
 	// LOAD-BEARING, and the trap that has caught this function three times
 	// (#6611, #6615, #6620): a quote EARLIER in the source consumes the `#` in
-	// the string branch, so the comment branch is never entered. `xx"""#"`
-	// passes every other clause here and leaves the mutant alive; `xxxx"#"`
-	// passes them and fails on PRISTINE. The scanner dispatches on the first
-	// quote-or-comment byte, so requiring the prefix to hold no quote is what
-	// proves `#` is that byte and the comment branch is what executes.
+	// the string branch, so the comment branch is never entered. `x"""#"'`
+	// passes every other clause here and leaves all three mutants alive;
+	// `xxx"#"'` passes them and fails on PRISTINE. The scanner dispatches on the
+	// first quote-or-comment byte, so requiring the prefix to hold no quote is
+	// what proves `#` is that byte and the comment branch is what executes.
 	if strings.ContainsAny(src[:h], "\"'") {
 		t.Fatalf("#6623 premise broken: the source before the `#` at %d must contain no quote, "+
 			"or the string branch consumes the `#` and the COMMENT branch never runs; got %q. "+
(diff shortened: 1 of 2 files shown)


cajasmota:
**REQUEST-CHANGES**, on one blocking finding and one false claim. Everything else the review touched came back verified, including the parts that were easiest to take on trust.

## Blocking — the test pins only half the quote alphabet

The narrow mutant, adding **only** the single-quote disjunct:

```diff
-			for ; i < len(b) && b[i] != '\n'; i++ {
+			for ; i < len(b) && b[i] != '\n' && b[i] != '\''; i++ {
```

`go vet` 0, full package `go test -count=1 ./internal/engine/` **exit 0 — SURVIVOR**. Witness: `# prefix='/ghost'` masks to `"                 "` pristine and `"         '/ghost'"` under the mutant.

So after a PR whose entire premise is *a quote inside a comment escapes*, the `'` half of that premise is unobserved. Worth naming the root cause precisely, because the enumeration is not what failed: **the sweep varied the input against a fixed mutant.** That is a rigorous answer to "does the guard admit a defeating input" — and the zero is real — but it is silent on "does the fixture pin the mutant *family*". Eleven million inputs against one mutant is one mutant's worth of evidence.

That is a genuinely new lesson. The enumeration method has earned its place; this says it answers the guard question, not the coverage question, and the two need separate work.

## The independent verification, since it is unusual to see this much survive

- The sweep reproduced **digit-for-digit** on an independent harness — every per-clause witness count matched exactly — then extended to length ≤9, **1,111,111,111 inputs, still 0 survivors and 0 pristine failures**.
- Better, the length bound turns out not to be load-bearing at all: there is an analytic argument that once all five clauses pass, the mutant dies for **arbitrary** length. C2 guarantees the scanner reaches the anchor; the mutant's loop halts at the first in-comment quote `p`; `p..p+2` are untouched by blanking, so C4's test on `src` is exactly the branch predicate on `b`; C4 forces the single-line branch, which skips, leaving `got[p]` a quote rather than a space. A proof beats a bound, and it should replace the bound in the body.
- Alphabet confirmed sufficient: the function dispatches on exactly `#`, `"`, `'`, `\n`, `\\`, plus `\r` in `blank()` — all six present, with four default-class representatives.
- **C5's de-claiming is accurate** — 0 mutant survivors, kept only to keep the all-spaces assertion honest on correct code. Recording a clause as admitting no survivor, rather than claiming it carries weight, is the same honesty #6622 landed with, and it is now the expected standard here.
- Both controls fail loudly with their own premise messages; neither can pass silently. My off-by-one (`i < len(b)-1`) is dead, and dead by *this* test alone.
- Placement judged correct: `TestPythonMaskInertRegions_MustNotPanic` asserts `len(got) == len(src)` against a crash, and a content row would need a second, differently-shaped post-condition inside a table whose contract is the panic. Own file is right.

## The false claim

The body says "the header of each file cross-references the other." Only one direction exists — the panic test file is untouched by this PR and mentions neither #6623 nor the new test. Either add the reverse reference or fix the sentence. A body claiming something the tree does not contain is the same defect class this milestone keeps finding, one level up.

## Out of scope, filed separately

The review found a second pre-existing survivor in the same loop — the backslash disjunct, witness `# a \ prefix="/ghost"`. Filing it rather than folding it in. The over-masking direction (dropping `b[i] != '\n'` so the comment eats to EOF) is already **DEAD**, killed by `TestSynth_FastAPI_MountedRouter_AbsentWhenUnresolvable_6414`.

Sent back.

cajasmota:
Rework verified. Merging.

## My own mutant, scored both ways

A content mutation nobody had run — leave the `#` characters themselves live while blanking everything else in the comment:

```diff
 			for ; i < len(b) && b[i] != '\n'; i++ {
-				blank(i)
+				if b[i] != '#' {
+					blank(i)
+				}
 			}
```

- With the new test **parked**: `go vet` 0, full package `go test -count=1 ./internal/engine/` **exit 0 — SURVIVOR**.
- With it **present**: `go vet` 0, **exit 1 — DEAD**, failing `TestPythonMaskInertRegions_QuotedSpanInsideCommentIsBlanked` and nothing else.

So the added kill power is attributable, not incidental — the test is the sole reason a comment's own delimiters are now observed. `http_endpoint_synthesis.go` byte-identical to `main` at `2081d244…` before and after, restored by `cp`; the PR is two test files, +174/-1.

## On the three things sent back

**The guard moved, and re-running the sweep rather than assuming was the right call.** C3/C4 now score per quote character instead of over "the first quote of either kind", which is a different predicate — the old sweep's zero would not have transferred. 1,111,111,111 inputs, 29,304,608 guard-passing, 0 survivors across all three mutants and 0 pristine failures. The two new drop-analysis rows being exactly symmetric (dropping C3a leaves `"`-only alive on 434,718 inputs; C3b does the same for `'`-only) is the useful evidence: it shows each half is carrying its own weight rather than one clause covering for both.

**Strengthening the guard rather than only the fixture was correct.** A fixture-only fix would have left a future edit free to drop either half silently — which is precisely the defect class that produced this PR, one level up.

**The analytic proof now carries the claim and the sweep is corroboration.** That is the right ordering: the length bound was never load-bearing, and a proof that neither half mentions `len(src)` retires the question rather than bounding it.

**The false cross-reference was made true rather than softened.** The panic test now carries the reverse pointer and says why content claims live elsewhere. Deleting the sentence would have been easier and would have left the two files as unaware of each other as they were.

## What is deliberately not in here

The backslash survivor (#6629) and the `j += 2` escape skip (#6624) stay out, and are named as such in the body. Keeping a test-only PR to one escape character is what made this review tractable twice.

PR #6628, opened 2026-08-25
chore(#6626): ignore Python bytecode so the quality helpers stop dirtying git status

### LINKED ISSUE #6626: __pycache__/ is not ignored — a .pyc was staged by accident on #6622 and only caught by hand
`scripts/quality/__pycache__/` is not ignored, so every agent that runs the Python quality helpers gets untracked build output in `git status` — and one `.pyc` has already been staged by accident and had to be amended out of a commit (PR #6622).

## Reproduced on `main` (`1c5c98936`), clean checkout

```
$ grep -c pycache .gitignore
0
$ git status --porcelain
?? scripts/quality/__pycache__/
$ git ls-files | grep -c '\.pyc$'
0
```

So: nothing is tracked yet, and nothing is ignored either. The only thing standing between a `.pyc` and `main` is whoever is staging remembering to look.

## Why it is worth a fix rather than a habit

Agents in this repo stage with `git add -A` or by directory more often than by explicit path, and the byte-level discipline we use for mutant restores does not extend to staging. The near-miss on #6622 was caught because that agent staged by explicit path on the second commit *after* noticing the first one; that is a person catching it, not the repo preventing it.

A committed `.pyc` is also a genuinely annoying artefact — architecture-and-interpreter-specific, unreadable in review, and it goes stale silently against the `.py` beside it.

## What to do

Add to `.gitignore`:

```
__pycache__/
*.py[cod]
```

Repo-wide rather than scoped to `scripts/quality/`, because the pattern is not specific to that directory and a scoped rule would need editing the next time a Python helper lands elsewhere.

## Verification

Not "it looks ignored" — assert it:

1. `git status --porcelain` is EMPTY on a clean checkout after running the quality helpers (today it is not).
2. `git check-ignore -v scripts/quality/__pycache__/ratchet.cpython-*.pyc` names the new rule.
3. `git ls-files | grep '\.pyc$'` stays empty — a guard against the "fix" being to commit them instead.

Small, but it is the kind of thing that gets fixed once or gets rediscovered by every agent forever.

Found while landing #6622.

cajasmota:
Closes #6626

`scripts/quality/__pycache__/` was neither tracked nor ignored, so every agent that runs the Python quality helpers picks up untracked build output — and a `.pyc` was staged by accident on #6622 and only caught by hand.

## Diff

```diff
@@ -21,6 +21,10 @@
 coverage.out
 coverage.html

+# Python bytecode — generated by the scripts/quality helpers on every run (#6626).
+__pycache__/
+*.py[cod]
+
 # Environment
 .env
 .env.local
```

Repo-wide rather than scoped to `scripts/quality/`, per the issue. Placed in its own labelled section alongside the other generated-output sections; nothing else in the file was touched.

### 1. `git status --porcelain` is empty after running the helpers

Before the fix, on a clean checkout of `1c5c98936`:

```
$ git status --porcelain
$ python3 -m pytest scripts/quality/test_ratchet.py -q
19 passed
$ git status --porcelain
?? scripts/quality/__pycache__/
```

After the fix (pycache removed, helpers re-run):

```
$ git status --porcelain
$ python3 -m pytest scripts/quality/test_ratchet.py -q
19 passed
$ ls scripts/quality/__pycache__/
ratchet.cpython-314.pyc  test_ratchet.cpython-314-pytest-9.0.2.pyc
$ git status --porcelain
```

Empty — the bytecode is generated and no longer reported.

### 2. `git check-ignore -v` names the new rule

```
$ git check-ignore -v scripts/quality/__pycache__/ratchet.cpython-314.pyc
.gitignore:25:__pycache__/	scripts/quality/__pycache__/ratchet.cpython-314.pyc
$ git check-ignore -v scripts/quality/__pycache__/
.gitignore:25:__pycache__/	scripts/quality/__pycache__/
```

Both exit 0 and cite line 25 of `.gitignore`, i.e. the rule added here.

### 3. Nothing is already tracked

```
$ git ls-files | grep -c '\.pyc$'
0
```

So this is not hiding tracked files, and the fix is not "commit them instead".

### 4. No currently-tracked file becomes ignored

`git ls-files -i -c --exclude-standard`, run on the parent commit and again with the change applied, is byte-identical:

```
$ git ls-files -i -c --exclude-standard   # before AND after — diff exits 0
cmd/grafel/testdata/audit_mount_prefix_retry/consumer/graph.json
cmd/grafel/testdata/audit_mount_prefix_retry/producer/graph.json
internal/quality/golden/go-chi-mini/.grafel/group.json
internal/quality/golden/php-slim-mini/.grafel/group.json
```

Those four are pre-existing (the `graph.json` and `.grafel/` rules), unrelated to this change and unchanged by it. The set is identical before and after, so the broad `*.py[cod]` rule captures zero tracked files. Independently: no tracked path matches `\.py[cod]$` at all.

No Go or full-repo suite was run — a `.gitignore` rule cannot affect them.


cajasmota:
Merging. Verified independently rather than on the report, because the one real risk in a broad ignore rule is that it swallows something already tracked.

`git ls-files -i -c --exclude-standard` before and after the change: **4 entries, byte-identical** (`diff` exit 0) — two `graph.json` testdata files and two golden `.grafel/group.json`, all from pre-existing rules and untouched by this. So `*.py[cod]` captures zero tracked files.

I also exercised the rule the reported verification did not: `check-ignore` matched `__pycache__/` for a bytecode file *inside* the directory, but the second rule only earns its place if it catches one outside. It does — `scripts/quality/foo.pyo` → `.gitignore:26:*.py[cod]`. Both lines are load-bearing, which is worth knowing rather than assuming, since the directory rule alone would have satisfied the issue's own reproduction.

Placement in its own labelled section with the `#6626` reference matches how the rest of the file is organised.

PR #6630, opened 2026-08-25
test(#6627): the durability guard's ref choice and its second early return were both unobserved

### LINKED ISSUE #6627: two survivors in ensure_durable_measured_on: the ancestry check may ignore the ref it found, and the ref-is-None guard is unobserved
Two mutants in `ensure_durable_measured_on` survive the whole `test_ratchet.py` suite. Both were found during #6621's review; both re-verified on `main` (`1c5c98936`, `ratchet.py` sha `ce30c02a…`) before filing, since issue bodies here go stale.

Baseline: `python3 -m pytest scripts/quality/test_ratchet.py -q` → **19 passed**.

## Survivor 1 — the ancestry check does not have to use the ref it found

```diff
-        ["git", "merge-base", "--is-ancestor", sha, ref],
+        ["git", "merge-base", "--is-ancestor", sha, "origin/main"],
```

**19 passed.** The function calls `default_branch_ref()` on the line above and then need not use the result: every fixture's default branch ref *is* `origin/main`, so hardcoding it is indistinguishable.

This matters because `QUALITY_BASE_REF` exists precisely for checkouts that cannot be discovered from refs — #6613's own fixture is one. Under this mutant an operator sets `QUALITY_BASE_REF`, watches `default_branch_ref()` honour it in the sha, and then has durability checked against a ref that may not exist in that checkout at all. `merge-base --is-ancestor` against a missing ref exits non-zero, so the failure mode is a confident refusal to write a sha that is in fact perfectly durable — with an error message naming a ref the user never chose.

The fix is a fixture whose default branch ref is **not** `origin/main`: set `QUALITY_BASE_REF` to a differently-named ref (or a repo whose default is `origin/trunk`) and assert the ancestry decision follows it. Assert the *decision*, not that the ref string appears somewhere — the error message already interpolates `ref`, so a message-only assertion would go green on a call that used the wrong ref and merely reported the right one.

## Survivor 2 — the second early return is unobserved

```diff
     ref = default_branch_ref()
-    if ref is None:
+    if False:
         return
```

**19 passed.** No fixture reaches `ensure_durable_measured_on` with a resolvable-looking sha *and* no default branch ref, so the guard never fires. Note this is not a silent behaviour change if it ever did fire — `merge-base --is-ancestor <sha> None` raises `TypeError` inside `subprocess`, i.e. a traceback instead of the intended quiet return.

This is the sibling of the trap #6613 already recorded from the other side: there, `ensure_durable_measured_on`'s **first** guard (`sha == "unknown"`) was dead in the obvious "no default branch" fixture because the second one carried the case. The obvious fixture cannot pin both, and today neither shape pins this one.

The shape that does: `sha` is a real, resolvable commit (so the first guard passes) while `default_branch_ref()` returns `None`. `QUALITY_BASE_REF` unset, no `origin/HEAD`, no `DEFAULT_BRANCH_REFS` candidate present — a local repo with a commit and no remote. Guard the fixture on all three independently: the sha resolves, `default_branch_ref()` really is `None`, and the first guard really is not the one returning.

## Rules

- Score every mutant **before and after** — each must be `19 passed` on pristine `main` and a failure after. A kill not first reproduced as a survivor is not evidence.
- `ratchet.py` must stay byte-identical to `main`; this is **test-only**. Restore by `cp` from a byte backup and re-verify with `shasum` — never `git checkout --`.
- If a shape turns out not to distinguish the mutants, record the negative result in-tree with its reason, as `AnnotatedTagCandidateIsNotADistinguisher` did for #6613. That is a complete outcome.
- Do not widen into the rest of `ratchet.py`.

Refs #6613, #6621, #6564.

cajasmota:
Closes #6627.

Two mutants in `ensure_durable_measured_on` survived the whole `test_ratchet.py` suite on `main` (`1c5c98936`). Both are now dead. **Test-only**: `scripts/quality/ratchet.py` is byte-identical to `main` — `shasum` is `ce30c02a1bdc794549bb1ef7df862bf75ccaa57a` before and after, verified after every mutant restore (restored by `cp` from a byte backup, never `git checkout --`).

## Mutant table

| # | Mutant | Before (on `main`) | After |
|---|---|---|---|
| 1 | `["git","merge-base","--is-ancestor",sha,ref]` → `…, sha, "origin/main"]` | **19 passed** (survives) | **3 failed, 21 passed** |
| 2 | `if ref is None:` → `if False:` (2nd early return) | **19 passed** (survives) | **1 failed, 23 passed** |

Pristine baseline: 19 passed before, **24 passed** after. `python3 scripts/quality/test_ratchet.py` also exits 0.

### `AncestryFollowsTheDiscoveredRef` (kills mutant 1)

New fixture `make_repo_divergent_default_ref` builds the shape no existing fixture had — a checkout whose default branch ref is **not** `origin/main`:

```
r1 ── r2      <- refs/remotes/origin/release   (discovered via QUALITY_BASE_REF)
 └── f1       <- feature, HEAD
o1            <- refs/remotes/origin/main      (resolvable, orphan history)
```

The two refs disagree in **both** directions, so the ancestry **decision** is observable:

- `test_accepts_a_sha_only_the_discovered_ref_can_reach` — `r1` is reachable from the discovered ref and not from `origin/main`; the writer must not refuse it. Under the mutant it raises `SystemExit`.
- `test_refuses_a_sha_only_the_hardcoded_ref_can_reach` — `o1` is reachable from `origin/main` and not from the discovered ref; the writer must refuse it. Under the mutant it is accepted.
- `test_the_writer_completes_on_a_checkout_whose_base_is_not_origin_main` — the operator-visible half, through `build()`: `--update-baseline` must write a baseline here, not refuse outright.

**No assertion touches the error message text.** That was deliberate and the mutant proves why: under the mutant the failure output reads `not an ancestor of refs/remotes/origin/release` for a call that asked git about `origin/main`. A message-only assertion would have gone green.

Premises asserted independently of `ratchet.py` (via a new `is_ancestor()` helper that asks git directly): the discovered ref is the override and not `origin/main`; `origin/main` really resolves; all four reachability facts; neither sha is `"unknown"`.

### `SecondEarlyReturnIsReached` (kills mutant 2)

New fixture `make_repo_no_default_branch_ref` — a local repo on branch `wip`, one real commit, no remote, no `origin/HEAD`, no local `main`/`master`. This is the combination nothing had: **a sha that resolves** *and* **no default branch ref**. Existing no-ref shapes all degrade `sha` to `"unknown"` and are carried by the FIRST guard.

- `test_returns_quietly_for_a_real_sha_with_no_ref_to_check_against` — must return silently. Under the mutant, `merge-base --is-ancestor <sha> None` raises `TypeError` inside `subprocess` (confirmed as the actual failure mode, not an incidental assertion miss).
- `test_the_writer_cannot_reach_this_guard_on_its_own` — control: `git_sha()` is `"unknown"` here, so `build()` can only ever hand the first guard's value in. There is no writer-level route to the second guard, which is why nothing observed it and why the test above is a direct unit call.

Three premises asserted separately: the sha resolves (asked of git); every candidate slot — `refs/remotes/origin/HEAD` and each entry of `DEFAULT_BRANCH_REFS` — is enumerated and absent, and `default_branch_ref()` is `None`; and `stderr` is empty, which is how the first guard would have announced itself had it been the one returning.

## Per-clause drop analysis

Each clause neutralised (premise assertion → `pass`), the fixture degraded to the shape that clause forbids, the mutant re-applied. Load-bearing ⇔ the mutant then survives. All measured.

| Clause | Degradation | Mutant | Result | Verdict |
|---|---|---|---|---|
| Discovered ref and `origin/main` disagree on reachability | point `origin/main` at `origin/release` | 1 | 3 passed | **load-bearing** |
| `origin/main` resolves | delete the ref | 1 | 2 failed, 1 passed | **partial — see below** |
| Sha is not `"unknown"` | pass `"unknown"` | 1 | see below | **partial — see below** |
| `default_branch_ref()` is `None` | add a local `main` at HEAD | 2 | 2 passed | **load-bearing** |
| Sha is not `"unknown"` | pass `"unknown"` | 2 | 2 passed | **load-bearing** |

Two clauses were only partially load-bearing, and are recorded in-tree that way rather than credited with weight they do not carry:

**`origin/main` must resolve** — with the ref deleted, per-test: `test_accepts…` still fails, `test_the_writer_completes…` still fails, `test_refuses_a_sha_only_the_hardcoded_ref_can_reach` **passes** (survives). So the clause is genuinely load-bearing for exactly one of the three tests. For the other two the mutant still dies, but only because `merge-base --is-ancestor` cannot resolve a missing ref — an argument about a broken checkout rather than about which ref the decision follows. The clause is kept for all three because it is what makes the kill an argument about ref *choice*. Recorded in `make_repo_divergent_default_ref`'s docstring.

**Sha is not `"unknown"` (mutant 1)** — at class granularity the mutant still dies, because the other two tests carry it. Restricted to the test the degradation actually targets, `test_accepts_a_sha_only_the_discovered_ref_can_reach` **passes** under the mutant: the first early return carries the call and the ancestry check is never reached. So it is load-bearing per-test. Separately verified that the `err.getvalue() == ""` assertion is the detector: with the degraded fixture and that guard intact, the test fails on **pristine** `ratchet.py`. It is not decorative. Recorded in the class docstring.

## On the issue body

Everything in #6627 checked out — both survivors reproduced exactly as described (19 passed each), both fixture shapes it proposed are the ones that work, and mutant 2's `TypeError`-inside-`subprocess` failure mode is exactly what it predicted. No corrections.

No negative results to record: both proposed shapes distinguished their mutants on the first attempt.

### DIFF
--- a/scripts/quality/test_ratchet.py
+++ b/scripts/quality/test_ratchet.py
@@ -345,6 +345,114 @@ def make_repo_criss_cross(root):
 STAMP = "test-stamp-6564"
 
 
+def is_ancestor(repo, sha, ref):
+    """True if `sha` is reachable from `ref`, asked of git directly.
+
+    The tests below use this to state a fixture's ancestry facts independently
+    of `ratchet.py` — the whole point of #6627's first survivor is that the
+    code can consult the WRONG ref and still produce a plausible answer, so the
+    premises cannot be read back out of the function under test.
+    """
+    return subprocess.call(
+        ["git", "merge-base", "--is-ancestor", sha, ref],
+        cwd=repo, stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL,
+    ) == 0
+
+
+def make_repo_divergent_default_ref(root):
+    """A checkout whose default-branch ref is NOT `origin/main`, and where the
+    two disagree about what is reachable (Refs #6627).
+
+    Returns (branch_point, orphan_sha), short.
+
+    Every other fixture in this file makes `refs/remotes/origin/main` the
+    answer `default_branch_ref()` gives, so `ensure_durable_measured_on` could
+    ignore that answer and hardcode `origin/main` with nothing to observe it.
+    Here the discovered ref is `refs/remotes/origin/release`, fed in through
+    `QUALITY_BASE_REF` — which exists precisely for a checkout that knows its
+    own base and cannot be discovered from refs — and `origin/main` is present
+    but sits on an ORPHAN history:
+
+        r1 ── r2            <- refs/remotes/origin/release   (the real base)
+         └── f1             <- feature, HEAD
+        o1                  <- refs/remotes/origin/main      (disjoint)
+
+    So the two refs disagree in BOTH directions, which is what makes the
+    ancestry decision observable rather than merely reported:
+
+      * `r1` (the branch point, and what `git_sha()` stamps) is reachable from
+        the discovered ref and NOT from `origin/main`;
+      * `o1` is reachable from `origin/main` and NOT from the discovered ref.
+
+    `origin/main` deliberately RESOLVES, and that clause was MEASURED rather
+    than assumed (Refs #6627). Deleting the ref from this fixture and
+    re-applying the mutant: the two permissive tests still fail, but only
+    because `merge-base --is-ancestor` cannot resolve a missing ref — an
+    argument about a broken checkout, not about which ref the decision follows.
+    The strict test, `test_refuses_a_sha_only_the_hardcoded_ref_can_reach`,
+    goes GREEN under the mutant without it, because with no `origin/main` the
+    mutant refuses `o1` for the same reason the real code does. So the clause
+    is load-bearing for exactly one of the three tests, and is kept for all
+    three because it is what makes the kill an argument about ref choice.
+    """
+    git(root, "init", "--quiet", "--initial-branch=trunk")
+    git(root, "config", "user.email", "t@example.com")
+    git(root, "config", "user.name", "t")
+    git(root, "config", "commit.gpgsign", "false")
+    with open(os.path.join(root, "a"), "w") as fh:
+        fh.write("1")
+    git(root, "add", "a")
+    git(root, "commit", "--quiet", "-m", "base")
+    branch_point = git(root, "rev-parse", "--short", "HEAD")
+    git(root, "branch", "feature")
+    with open(os.path.join(root, "b"), "w") as fh:
+        fh.write("someone else landed this")
+    git(root, "add", "b")
+    git(root, "commit", "--quiet", "-m", "release branch moves on")
+    git(root, "update-ref", "refs/remotes/origin/release", "HEAD")
+
+    # An unrelated root commit, so `origin/main` can resolve while sharing no
+    # history at all with the branch the base ref names.
+    git(root, "checkout", "--quiet", "--orphan", "unrelated")
+    with open(os.path.join(root, "c"), "w") as fh:
+        fh.write("a different project's main")
+    git(root, "add", "c")
+    git(root, "commit", "--quiet", "-m", "unrelated root")
+    orphan = git(root, "rev-parse", "--short", "HEAD")
+    git(root, "update-ref", "refs/remotes/origin/main", "HEAD")
+
+    git(root, "checkout", "--quiet", "--force", "feature")
+    with open(os.path.join(root, "a"), "w") as fh:
+        fh.write("2")
+    git(root, "commit", "--quiet", "-am", "feature work")
+    assert branch_point != orphan
+    return branch_point, orphan
+
+
+def make_repo_no_default_branch_ref(root):
+    """A local repo with ONE REAL COMMIT and no candidate default-branch ref
+    whatsoever (Refs #6627).
+
+    Returns the short HEAD sha.
+
+    The branch is `wip`, there is no remote, no `origin/HEAD`, and no local
+    `main` or `master`, so every entry in `DEFAULT_BRANCH_REFS` misses and
+    `default_branch_ref()` returns None. Unlike
+    `test_degrades_to_unknown_when_no_default_branch_is_available`, the caller
+    here holds a sha that RESOLVES — which is the combination that reaches
+    `ensure_durable_measured_on`'s second early return.
+    """
+    git(root, "init", "--quiet", "--initial-branch=wip")
+    git(root, "config", "user.email", "t@example.com")
+    git(root, "config", "user.name", "t")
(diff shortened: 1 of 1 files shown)


cajasmota:
Merging. Approved on an independent review that reproduced every number, plus two coordinator mutants of my own.

## My mutants — both directions of the guard

| mutant | result |
|---|---|
| `raise SystemExit(…)` → `print(…)` — refusal downgraded to a warning | **2 failed**, incl. the new `…refuses_a_sha_only_the_hardcoded_ref_can_reach` |
| `if ok != 0:` → `if True:` — refuse unconditionally | **4 failed**, incl. both new accept-direction tests |

The second matters more than it looks: it proves the new tests observe **acceptance**, not just refusal. A durability guard that refuses everything is the easy thing to ship by accident, and a refusal-only pin would never have seen it.

## The sharpest thing here

The claim that a message-text assertion would have gone green on mutant 1 is **true, and the reviewer reproduced the exact string**: the failure reads `not an ancestor of refs/remotes/origin/release` for a call that asked git about `origin/main`. The `SystemExit` f-string interpolates `ref`, which the mutant leaves untouched — so the error faithfully reports the ref the code *chose* while git was asked about a different one.

That is the cleanest example this milestone of why a diagnostic string is not a regression pin. Asserting the decision in both directions is what makes these tests evidence.

## The two "partial" gradings hold

Both were checked rather than credited, and both are honest:

- *"`origin/main` resolves"* — under M1, degrading it leaves exactly `test_refuses_a_sha_only_the_hardcoded_ref_can_reach` passing, test-for-test as claimed. Keeping the clause is right: without it the kills argue "a missing ref fails", not "the decision follows the discovered ref".
- *"sha is not `unknown`"* — a genuine per-test survivor at single-test granularity, carried by siblings at class granularity. And the `err == ""` assertion is a real detector: the same degradation with it kept fails on **pristine**.

Under-claiming hides coverage the way over-claiming invents it, so I wanted these measured. They were.

## One survivor found, filed rather than folded in

`os.environ.get("QUALITY_BASE_REF", "").strip()` → without `.strip()` survives all 24 tests. With a trailing newline — exactly what `$(git symbolic-ref …)` and YAML block scalars produce — the override is silently discarded, `git_sha()` degrades to `"unknown"`, and a durable sha is refused against a ref the operator never chose. That is #6627's own failure mode re-entering through a different door, and `.strip()` is asserted only in a docstring. Filing it.

Also recorded: `sha` → `sha + "^{commit}"` in the ancestry call is **equivalent** — `measured_on` is a commit sha by construction and the `"unknown"` case returns earlier — and the `QUALITY_BASE_REF` pop without an `addCleanup` restore is a real but currently harmless nit, worth tidying when that file is next touched.

Issue #6560, opened 2026-08-23
axumRouteRe is the only route source in the axum pass, so a utoipa_axum routes!() registration yields no endpoint definition while a sibling .route() in the same function does


arthurgeron:
Found while auditing Rust endpoint recall against `main` at 8595d7a6b.

`synthesizeAxumRoutes` (`internal/engine/http_endpoint_axum.go:71`) has exactly one route source, the loop at `:144` over `axumRouteRe` (`:39`):

```go
`\.route\s*\(\s*"([^"\n\r]+)"\s*,\s*(get|post|put|patch|delete|head|options)\s*\(\s*([A-Za-z_]\w*)`
```

That requires a literal path string in the registration call. `utoipa_axum` registers differently: the verb and path live on the handler's `#[utoipa::path]` attribute, and the registration call names only the handler, `OpenApiRouter::routes(routes!(list_items))`. No `.route(` call exists, so the loop finds nothing and no `http_endpoint_definition` is emitted. `internal/custom/rust/axum.go:31` carries the same `.route(`-only shape.

`internal/custom/rust/utoipa.go:165` does read the attribute, but it emits `SCOPE.Operation` / Subtype `endpoint` carrying a `route_path` property, a different entity family. `internal/engine/http_endpoint_vapor.go:7` records what that costs for the identical Swift case: those markers are not `http_endpoint_definition` entities, so neither `ResolveHTTPEndpointHandlers` nor the e2e route-test linker sees them. Swift got a bridging synthesizer for that reason (#4749); Rust has no equivalent for `utoipa_axum`.

`OpenApiRouter` appears in no non-test file in the tree. `utoipa` appears in four non-test `.go` files: `internal/custom/rust/utoipa.go`, `internal/custom/rust/observability.go`, `internal/custom/golang/observability.go` and `internal/extractors/rust/struct_fields.go`. None is under `internal/engine`, and no file under `internal/engine/rules` mentions it, so `utoipa.go` is the only reader of the attribute's `path`.

`docs/coverage/detail/lang.rust.framework.utoipa.md` grades both `Route extraction` and `Endpoint synthesis` as `full`, citing `internal/custom/rust/utoipa.go`, with the note "enriches bare axum/actix routes with documented contract". Read against the canonical endpoint family, that reads as more coverage than the `routes!` form gets: the attribute is parsed, but a route registered only through `routes!` produces no `http_endpoint_definition` for anything to enrich.

## Constructed example

[...]

PR #6631, opened 2026-08-25
docs(#6560): utoipa coverage graded itself full for a form that minted nothing (Arm C)

cajasmota:
Arm C of #6560, from `arthurgeron`'s report. Docs only — no Go file is touched.

## What was wrong

`docs/coverage/detail/lang.rust.framework.utoipa.md` graded both `Endpoint synthesis` and `Route extraction` as `✅ full`, cited only `internal/custom/rust/utoipa.go`, and carried the note "enriches bare axum/actix routes with documented contract". The report is the evidence that `full` was never true: before Arm A (`defb267e6`) a `routes!()` registration minted no `http_endpoint_definition` at all.

## The regrade

Both cells → `🟢 partial`, the vocabulary the sibling Rust records already use (`tower.md:21`, `hyper.md:21`, `warp.md:21`). No new symbol invented.

Justification, taken from the pass's own scope note at `internal/engine/http_endpoint_utoipa_axum.go:66-79`:

- multi-handler `routes!(a, b)` and path-qualified `routes!(crate::items::list)` do not match `utoipaRoutesMacroRe` (`:102`) and mint nothing — pinned by `TestUtoipaAxum_MultiHandlerRoutesMacroNotMinted` (`http_endpoint_utoipa_axum_test.go:336`);
- a handler whose `#[utoipa::path]` lives in another file is skipped, because the attribute map is built from one file only (`:73-76`, `TestUtoipaAxum_UnknownHandlerNotMinted` at `:176`);
- `OpenApiRouter::nest` prefix composition is not threaded, so a nested router yields the unprefixed attribute path (`:77-79`);
- cross-file double mint is not deduped (`:54-58`).

## The "enriches" claim — corrected in review (`dbc6d69a3`)

The first push **deleted** the sentence and replaced it with an uncited negative: "no code merges an attribute into a `.route(...)`-derived endpoint". That negative is **false at the `SCOPE.Operation` layer**, and it is exactly the defect class this doc exists to prevent.

The merge is real, and it is the house mechanism:

- `internal/custom/rust/utoipa.go:164-165` names its op `method + " " + routePath`, Kind `SCOPE.Operation` / subtype `endpoint`;
- `internal/custom/rust/axum.go:148-149` names its route op `method + " " + fullPath`, identical Kind and subtype;
- `MergeWithCustom` keys on `(SourceFile, Kind, Name)` since #6104 — stated outright by the sibling stampers at `internal/custom/rust/endpoint_response_codes.go:29` and `:587`.

So the note now states **both layers, separately**, and does not collapse them:

- **`SCOPE.Operation` layer** — the attribute op merges onto the plain axum route op by Name. Real precondition, now stated: the Names must be identical, i.e. same source file **and** attribute path equal to mount path. axum composes its nest prefix into `fullPath` (`axum.go:143-144`) while the attribute path stays unprefixed, so a nested or prefixed mount does **not** merge.
- **Canonical `http_endpoint_definition` layer** — no enrichment at all. `synthesizeUtoipaAxumRoutes` (`internal/engine/http_endpoint_utoipa_axum.go:261`) mints a **separate** definition and explicitly **skips** any handler `synthesizeAxumRoutes` or `synthesizeRocket` already registers from the same file (`utoipaRegisteredElsewhere`, `:232`).

## Citation defects fixed in the same commit

- **DRIFTED:** `internal/custom/rust/utoipa.go:165-172` was cited for "normalises `{id}`/`:id`/`<id>` and captures the decorated handler fn". Neither is in that range. Split into `:151` (`rustNormalizePath`), `:158-162` (handler fn capture), `:174` (`handler_name` stamp), `:164-165` (the op's Name and Kind).
- **UNCITED:** "not a member of the canonical endpoint family" now cites the gates that skip it — `applyEndpointPagination` (`internal/engine/http_endpoint_pagination.go:80`) and `applyEndpointResponseCodes` (`internal/engine/http_endpoint_response_codes.go:100`), both `Kind != httpEndpointDefinitionKind`. The note also says explicitly that the op is **not** thereby inert, so the gating fact cannot be read as "the op does nothing".

## Cites

Now include `internal/engine/http_endpoint_utoipa_axum.go` and its test (and `http_endpoint_synthesis.go` on the synthesis cell, for the dispatch at `:1350`), so the file carrying the behaviour is reachable from the grade.

## Other docs checked

Aggregate tables are unaffected by construction: `groupDigest` counts `full+partial` as covered (`tools/coverage/generate.go:509-516`) and `supportGlyph` still returns `🟡` while three Routing cells stay `missing`, so Routing remains `🟡 3/6` in `by-language/rust.md:44` and `by-category/http_framework.md:158`. All 17 docs mentioning utoipa were checked; no other one grades its routing cells — the rest reference utoipa only for `ToSchema` DTO extraction.

## Note for review

`registry.json` and the generated `.md` were edited together by hand rather than through `go run ./tools/coverage gen`, because this lane was told not to compile. Both notes were verified byte-identical between the JSON and the rendered table cell, in `detail.md.tmpl` field order, so a regen should be a no-op — **still worth confirming with a `gen` run on a lane that can build.**

Arms B1/B2 remain open on #6560, so this does not close it.

Refs #6560


cajasmota:
Merging. Generator gate re-run on this head: `go run ./tools/coverage gen` → 822 records, `git status` clean. The hand-edited `registry.json` and the rendered `.md` agree with the generator byte for byte.

## The correction is right, and it went one step further than the review

I spot-checked the crux myself rather than relaying it: `internal/extractors/custom_dispatch.go:346` is `type identityKey struct{ file, kind, name string }`, and `internal/custom/rust/axum.go:143-149` composes the nest prefix into `fullPath` *before* building `name`. So the merge key is **(SourceFile, Kind, Name)**, not Name alone — the note states the stricter, correct precondition, which the review's own framing had understated.

That matters for the claim being made: same-file *and* equal path. A nested mount composes its prefix into the axum op's Name while the attribute path stays unprefixed, so the Names differ and no merge happens. Stating both halves is what makes the sentence checkable instead of merely plausible.

## What this PR actually fixed, in order

1. The original doc graded a form `✅ full` that minted nothing.
2. The first fix removed a true-but-imprecise sentence and replaced it with a **false uncited negative** — the same defect class, one level up.
3. This fix separates the two layers and says which is true where: enrichment is **real** at the `SCOPE.Operation` layer by merge-by-Name, and **absent** at the canonical `http_endpoint_definition` layer, where a separate definition is minted and already-registered handlers are skipped. Neither fact generalises to the other, which is exactly how step 2 went wrong.

Both citation defects resolved properly: the drifted `:165-172` range split four ways to the lines that actually carry each claim, and "not a member of the canonical endpoint family" now cites the two `Kind != httpEndpointDefinitionKind` gates — with a guard sentence so the gating fact cannot be misread as "the op is inert". That guard is the difference between a citation and an explanation.

## Two judgement calls I agree with

**`handler_attribution` left `✅ full`.** The `partial` elsewhere is driven by a *cross-file* failure; handler attribution has no cross-file dependency, since an attribute and the item it decorates are adjacent by language rule. Downgrading it for symmetry would have been the tidy answer and the wrong one.

**The `registry.json` re-serialisation reverted.** A 221k-line diff replaced by 4 changed lines across 2 files, reproducing Go's `<`/`>` escaping by hand. A generated file reformatted wholesale is unreviewable, and catching that before pushing is the right reflex.

Noted for a future arm, not this one: `reUtoipaFnName` searches all of `src[endOff:]`, so an attribute not followed by a `fn` in that file would attribute the next `fn` anywhere below. Explicitly speculative — no failing case was constructed — and I would rather see it recorded as speculative than asserted.

PR #6634, opened 2026-08-25
fix(#6314): grafel_find culled each repo to a hard top-10 before fusion, so rank 11+ was absent, not low-ranked

### LINKED ISSUE #6314: grafel_find (bm25) culls to a hard top-10 per repo before de-noise/fusion, so a file stem outranks an exact entity-name match and the relevant file can be absent rather than low-ranked

cajasmota:
`internal/mcp/tools.go:670` called `r.getBM25().Search(question, 10)` — a bare inline literal, with the semantic index capped identically at `:681`. That cull is the **first** stage of the bm25 pipeline:

```
per-repo top-10 cull -> RRF fusion (:688) -> de-noise -> kind filter
  -> rerankScored (:737) -> min_score -> max_results
```

So an entity at rank 11+ **in its own repo** was *absent* from the answer, not low-ranked, and nothing downstream could recover it. `max_results` (default 50, ceiling 200) gates only the *returned* set at the end, and the schema's `limit` applies solely to the `search=substring` branch — so `max_results=200` still saw at most 10 candidates per repo. That is why @manuel1358000's substring workaround behaved so differently from bm25 mode.

The reason ten slots fill so easily: the heaviest indexed field is the **file stem** (weight `1.5`, `scoring.go:23-26`) while the entity label is only `1.0`, and there is no exact-name bonus anywhere in the path. Eleven files whose *stems* share a query token are enough to hide the entity whose *name* **is** that query.

## Red step first — it is the point of this PR

**Nothing in this repo measured retrieval quality.** The nearest test, `find_ranking_1747_test.go`, uses a 7-entity fixture — smaller than the cap — so the cull can never fire in it. Changing `10` to `maxResults` would have passed the entire suite unchanged and proved nothing.

`internal/mcp/find_recall_cull_6314_test.go` builds a 15-entity single-repo corpus: 14 decoys carry the token `checkout` through their **file stems** (`src/stepNN/checkout.ts`), one entity carries it as its **name** (`Checkout`, in `src/promo/banner.ts`). The fixture asserts its own validity first — the target must be present in the *unlimited* BM25 result set and rank past 10 — so it cannot silently stop exercising the cull.

On unmodified `main`, with `min_score=0` so the cull is the only stage able to drop it:

```
--- FAIL: TestFindRecall_ExactNameSurvivesCrowdedRepo_6314
    target ranks 15 of 15 in the raw BM25 index
    entity named exactly the query "Checkout" is ABSENT from results in a crowded repo
    got: {"matches":[ ...exactly the ten decoys, identical scores... ]}
--- FAIL: TestFindRecall_CullTracksMaxResults_6314
```

## The fix

Derive the per-repo cull from the already-clamped `maxResults`, at both call sites. **No BM25 weight is touched** — weight tuning is deliberately out of scope.

## The semantic half: PINNED, not declared

Review caught that reverting **only** the semantic call site left the full package green — half the fix pinned, half silently revertible, even though the vector index has the identical defect.

Option taken: **pin it.** Standing up a vector index turned out not to be disproportionate — `embed.NewStore` + `Put` accepts explicit vectors, so semantic rank is set *by construction* (2-D unit vectors, cosine decreasing monotonically) rather than inferred from an embedding model, and the query side is a two-line fake backend.

`TestSemanticRecall_VectorHitSurvivesCrowdedRepo_6314` builds a repo where the target is **invisible to BM25** (name, path and docstring share no token with the query) and reachable **only** through the vector index at rank 15 of 15. Both premises are asserted as fixture-validity checks, so the test cannot quietly stop exercising the semantic cull.

`installQueryEmbedder` swaps the package-level `globalQueryEmbedder` and burns its `sync.Once`. It deliberately does **not** call `t.Parallel`: Go defers every top-level parallel test until the serial ones finish, so a serial test owns the global outright and no concurrent reader can observe the swap.

## The pool is pinned to `max_results`, not to a constant

A first attempt bounded only the *widening* direction (`perRepoCull := maxResults * 4`). Review then found the survivor that mattered:

```diff
-	perRepoCull := maxResults
+	perRepoCull := 200
```

`go vet` 0, full package green — **survivor**, reproduced here at 114.1s. The cull could be fully decoupled from `max_results` with nothing noticing, while this body claimed it was "pinned to the caller-visible argument rather than to any particular constant". `TestFindRecall_CullTracksMaxResults_6314` claimed to close exactly that and did not: it only counted returned rows, which a constant cull satisfies. Its doc comment asserted that lowering `max_results` below the target's rank drops it — **nothing observed that.**

What separates a derived pool from a constant one is a **tight** `max_results`. `TestFindRecall_PoolTracksMaxResultsNotAConstant_6314` queries the existing 260-entity corpus with `max_results=3`:

- derived → the pool admits 3, nothing is discarded, **no note**;
- constant → the pool admits 200, the surplus is dropped after the fact:

```
the per-repo pool did not follow max_results down to 3 in a 260-entity repo:
note: truncation_note":"capped at max_results=3; 197 additional hits omitted ..."
```

That single assertion kills the whole `perRepoCull := <constant>` family, not just `200`, and retires the unbacked prose. The `*4` widening direction is closed by the ceiling test below, using the clamp the contract already defines — no invented number in either.

## What the `truncation_note` actually carries — re-credited

The note was previously credited with observing the 200 clamp. **It does not.** It observes only *pool ≤ max_results*; it says nothing about absolute pool size. Deleting the clamp at `tools.go:606` is caught by the **row-count** assertion (260 rows returned), not by the note — with the clamp gone `max_results` is 1000 and a 260-entity pool never exceeds it, so no note is emitted.

The test is renamed `TestFindRecall_ClampBoundsPoolAndRows_6314` and each assertion now names the property it carries. Together they bound the pool at `min(max_results, 200)` per repo.

## Known limits, stated rather than implied

Both are now written into the test file as well:

1. **Every fixture is single-repo.** The real pre-filter pool is `len(repos) × perRepoCull`, so a 20-repo group still admits up to 4,000 candidates per request. That multiplier is **unobserved**; only the per-repo bound is pinned.
2. **`truncation_note` is one-directional.** It witnesses a pool wide *relative to `max_results`*, never one wide *in absolute terms* while `max_results` is wider still — which is precisely the mutant that survived until now.

## Mutants scored against the final committed state

`go vet` clean on each; `-count=1`, full package, **no `-run` filter**; every restore by `cp` from a byte backup with `shasum` re-verified.

| mutant | before | after |
|---|---|---|
| `perRepoCull := 10` (the original defect) | RED | RED |
| `perRepoCull := maxResults / 4` | RED | RED |
| `semIDs := r.Semantic.Search(qVec, 10)` | **survivor** | **RED** |
| `perRepoCull := maxResults * 4` | **survivor** | **RED** |
| `perRepoCull := 200` | **survivor** (114.1s, rc=0) | **RED** |

Baseline with all five tests: `ok`, 107.8s.

## Memory shape

- The pre-filter `all` slice grows from `10` to at most `maxResults` candidates per repo. `maxResults` is clamped to `[1, 200]` at `:602-607`, **before** the loop — so the bound is the ceiling the caller-visible contract already imposes, and **no new constant is introduced**. That bound is now asserted, not merely commented.
- Each element is a small `scored{repo, hit}` struct, transient and freed with the request. A 20-repo group moves from ~11KB to ~224KB worst case — **and that `len(repos)` multiplier is the part no test observes** (see Known limits).
- `BM25Index.Search` allocates its hits slice at full candidate width (`make([]Hit, 0, len(scored))`) *regardless* of the limit, and scores every matching doc either way — so the index-side footprint and the scoring cost are **unchanged**. The real cost is up to 200 rather than 10 `resolve()` materializations per repo. That is the latency fee, and it is the acceptable kind.

## Blast radius

`getBM25().Search` has **exactly one** non-test caller — this one. (The original grounding said two; the second cap at `:681` is `r.Semantic.Search`, a different index with its own type. Both are changed here, and both are now pinned.)

The returned-set contract is unchanged: `max_results` still gates output at `:826`, and `TestFindRecall_CullTracksMaxResults_6314` pins that a tight `max_results=3` still returns at most 3 rows.

Closes #6314

### DIFF
--- a/internal/mcp/find_recall_cull_6314_test.go
+++ b/internal/mcp/find_recall_cull_6314_test.go
@@ -12,10 +12,13 @@
 package mcp
 
 import (
+	"context"
 	"fmt"
+	"math"
 	"strings"
 	"testing"
 
+	"github.com/cajasmota/grafel/internal/embed"
 	"github.com/cajasmota/grafel/internal/graph"
 )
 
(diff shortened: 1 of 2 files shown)


cajasmota:
Red step is exactly what I asked for, and the honesty about the surviving widening mutant is the right instinct. One finding sends this back.

## My mutant — the semantic half of the fix is unobserved

The change replaces two `10` literals with `perRepoCull`. Reverting **only the semantic one**:

```diff
-				semIDs := r.Semantic.Search(qVec, perRepoCull)
+				semIDs := r.Semantic.Search(qVec, 10)
```

`go vet` 0, full package `go test -count=1 ./internal/mcp/` **exit 0 — SURVIVOR** (98.8s).

So half the fix is pinned and half is free to be reverted by any future edit. The semantic index has the identical defect the issue describes — a relevant entity at vector rank 11+ is absent rather than low-ranked, and no later stage can recover it — and nothing now observes that it was fixed.

Two acceptable outcomes, and I do not mind which:

1. **Pin it** — a second recall case that crowds the semantic path the way the BM25 one crowds the index.
2. **Declare it honestly** — if the test harness has no vector index and standing one up is disproportionate, say so **in the test file**, naming what is unobserved and why, so the next reader does not mistake the passing suite for coverage.

What is not acceptable is leaving it silent, because a two-line fix with a one-line pin reads as fully tested.

## The widening mutant

`perRepoCull := maxResults * 4` surviving is fine as a recall claim — the test asserts presence, not pool width, and you said so rather than hiding it. But note which direction that leaves open: **more permissive**, i.e. a larger pre-filter pool. Every serious defect this milestone came from a mutant that made a predicate more permissive, and here the cost is memory in a package whose RSS is already contended, with concurrent MCP agents as the primary consumer.

Not blocking, and I do not want a magic number invented to close it. But if the `[1,200]` clamp is the real bound, that is a property worth asserting once — that the cull never exceeds the caller-visible ceiling — rather than left as a comment. Your call; say which you chose.

## On the grounding error

You are right and I was wrong: `getBM25().Search` has **one** non-test caller, not two. The second site is `r.Semantic.Search` — a different index with its own type. Thank you for checking rather than repeating it; the blast-radius claim in my brief was off by one index, and it is exactly the kind of thing that gets copied forward into three more briefs if nobody says so.

## What stands

The fixture asserting **its own validity first** — target present in the unlimited BM25 set and ranked >10 — is the part I would keep in every future recall test here. Without it, this test silently stops exercising the cull the moment the weights shift, which is precisely how the existing 7-entity ranking test came to observe nothing. And the memory analysis is the right shape: identifying that `BM25Index.Search` scores every matching doc regardless of the limit, so the cost is `resolve()` materialisations rather than slice growth, is a real answer instead of a reassurance.

cajasmota:
Merging. The blocking survivor is closed by an assertion that kills its whole family, and the note is now credited to the assertion that actually carries it.

## My own mutant — the tightest boundary

I ran a one-element off-by-one nobody had scored:

```diff
-	perRepoCull := maxResults
+	perRepoCull := maxResults + 1
```

`go vet` 0, full package `go test -count=1 ./internal/mcp/` **exit 1**, failing **both** new tests. A pool one candidate wider than the caller asked for is now observable, which is about as tight as this bound can be pinned without asserting on internals.

## Why the new assertion is the right shape

Presence-of-target and row-count are both satisfied by a constant cull, so neither can distinguish a *derived* pool from a *fixed* one. A **tight** `max_results` can: derived admits 3 and discards nothing, constant admits 200 and drops the surplus after the fact, which the response body reports as `197 additional hits omitted`. That kills every `perRepoCull := <constant>` mutant at once rather than the specific `200` I sent back — the difference between fixing a finding and fixing the class it came from.

## The re-crediting is the part I want on the record

The original test's name and comment gave `truncation_note` a property it did not carry. The corrected split states what each assertion actually observes:

- **absence of `truncation_note`** → only *pool ≤ max_results*. It cannot see the 200 clamp: delete the clamp and `max_results` becomes 1000, which a 260-entity pool never exceeds, so no note fires.
- **row count ≤ 200** → this is what catches clamp removal, and therefore carries the *≤ 200* half.

Two assertions, two properties, neither borrowing the other's credit. The doc comment on the older test also lost a claim about drop-below-rank behaviour that nothing observed — deleted rather than reworded, since the honest version is "this test does not check that" plus a pointer to the one that does.

## Limits stated rather than implied

Both are now in the test file *and* the body:

1. Every fixture is **single-repo**, so the real `len(repos) × perRepoCull` pool is unobserved — a 20-repo group still admits up to 4,000 candidates. Flagged inline in the memory-shape section so the ~224KB figure is not read as a pinned bound.
2. `truncation_note` is **one-directional** — it witnesses a pool wide *relative to* `max_results`, never one wide in absolute terms while `max_results` is wider. That is precisely the gap the `200` mutant lived in, and naming it is what stops the next reader assuming the note covers both.

## For the reporter

The original report — *"culled before fusion, so absent rather than low-ranked"* — was precise about the mechanism, and that is why this was checkable in one grounding pass instead of becoming a ranking-opinion thread. The fix is two lines; everything else here is the apparatus needed to prove those two lines stay correct, which did not exist before because nothing in the repo measured retrieval quality at all.

`tools.go` verified at `8644cac7e…` before and after every mutant, restored by `cp`. Pre-existing `-shuffle` order coupling left alone and filed as #6639.

Issue #6613, opened 2026-08-25
ratchet.py: three degradation branches are unobserved because every fixture is a well-formed linear repo — the "unknown" writer guard has a user-visible failure mode


cajasmota:
Three permissive mutants in `scripts/quality/ratchet.py` pass the whole suite. All three are **pre-existing on `main`** and were found by an adversarial review of PR #6609 while verifying a different fix. They are filed together because they share one cause: the fixtures exercise the happy path and nothing else, so every degradation branch is unobserved.

Verified against `main` with `python3 -m unittest discover -s scripts/quality -p 'test_*.py'`.

## 1. The `"unknown"` early return in `ensure_durable_measured_on` — real behaviour change

```diff
-    if sha == "unknown":
-        return
+    if False:
+        return
```

**exit 0.** And this one is not cosmetic. A shallow or detached checkout that legitimately degrades to `"unknown"` currently prints a WARNING and continues. Without the guard it reaches `merge-base --is-ancestor unknown ref`, which fails, and the writer raises **`SystemExit`** — a hard failure where the design is an advisory warning.

The `"unknown"` path is written about at length in `git_sha()`'s docstring — *"Degrades to `unknown` when no default branch is available. Never to HEAD"* — and #6569 pinned that `git_sha()` **produces** it. Nothing observes what the **writer** does when it receives it. This is the most valuable of the three: it has a user-visible failure mode.

## 2. The tag-peel in `default_branch_ref()` — unobserved

```diff
-        if _git("rev-parse", "--verify", "--quiet", ref + "^{commit}"):
+        if _git("rev-parse", "--verify", "--quiet", ref):
```

**exit 0.** Every ref in every fixture is already a commit, so the peel never does work. It matters when a ref names an annotated tag: without `^{commit}` the candidate resolves to the tag object, and the later `merge-base` against a tag behaves differently from a commit.

Cheap to pin: a fixture whose candidate ref is an annotated tag.

## 3. `merge-base` vs `merge-base --all` — unobserved

```diff
-    base = _git("merge-base", "HEAD", ref)
+    base = _git("merge-base", "--all", "HEAD", ref)
```

**exit 0.** All fixtures are linear, so exactly one merge-base exists and `--all` prints one line. With criss-cross history `--all` returns several lines, `rev-parse --short` fails on the multi-line input, and `git_sha()` silently degrades to `"unknown"` — a provenance field quietly emptied on exactly the repositories with the most complex history.

[...]

PR #6638, opened 2026-08-25
test(#6633): QUALITY_BASE_REF's .strip() was asserted only in a docstring

### LINKED ISSUE #6633: QUALITY_BASE_REF is stripped and nothing observes it — a trailing newline silently discards the override
`default_branch_ref()` strips whitespace off the `QUALITY_BASE_REF` override, and nothing observes it. Dropping `.strip()` survives all 24 tests in `scripts/quality/test_ratchet.py`:

```diff
 def default_branch_ref():
-    override = os.environ.get("QUALITY_BASE_REF", "").strip()
+    override = os.environ.get("QUALITY_BASE_REF", "")
```

**Non-equivalence witness**, with `QUALITY_BASE_REF="refs/remotes/origin/release\n"` — a trailing newline, which is exactly what `$(git symbolic-ref …)` and a YAML block scalar produce:

| | `default_branch_ref()` | `git_sha()` | `ensure_durable_measured_on(branch_point)` |
|---|---|---|---|
| pristine | `refs/remotes/origin/release` | `f3ae2d7` | accepted |
| mutant | `refs/remotes/origin/main` | `"unknown"` | **REFUSED** |

## Why this is worth its own issue

This is **#6627's failure mode re-entering through a different door.** #6627 was about the ancestry check being free to ignore the ref `default_branch_ref()` chose. Here the override is discarded *before* the choice is even made: the operator sets `QUALITY_BASE_REF`, the trailing newline makes `rev-parse --verify` fail, the search falls through to `origin/main`, and durability is judged against a ref they never picked — refusing a sha that is in fact perfectly durable, with an error naming the wrong ref.

`QUALITY_BASE_REF` exists precisely for checkouts that cannot be discovered from refs. An override that silently does nothing is worse than no override, because the operator has no signal that it was ignored.

The `.strip()` is asserted only in prose — the docstring says *"QUALITY_BASE_REF overrides the search for a checkout that knows its own base"* — and prose asserting a property nothing observes is the dominant defect class in this repo.

## What is needed

1. A test with a whitespace-padded `QUALITY_BASE_REF` (trailing `\n` at minimum; consider a leading space too) asserting the override is honoured — the mutant above must die.
2. **Score it before and after**: 24 passed on pristine `main` (survivor reproduced), failing after. A kill not first reproduced as a survivor is not evidence.
3. Assert the **outcome**, not the string: check the ancestry *decision* or the resolved sha, not that the ref name appears in a message. #6630 established why — the `SystemExit` f-string interpolates `ref`, so a message assertion reports the ref the code chose even when git was asked about a different one.
4. Guard the premise: the padded value must genuinely resolve after stripping and genuinely fail to resolve unstripped, or the test proves nothing about `.strip()`.

## Also worth folding in while that file is open

`SecondEarlyReturnIsReached` calls `os.environ.pop("QUALITY_BASE_REF", None)` with no `addCleanup` restore, so an ambient value is dropped for the rest of the run. Harmless today — nothing else in the suite reads it — but it is a cross-test coupling waiting to produce an order-dependent failure.

## Rules

- **Test-only.** `scripts/quality/ratchet.py` must stay byte-identical to `main`. Restore mutants by `cp` from a byte backup, verify with `shasum`, never `git checkout --`.
- Do not widen into the rest of `ratchet.py`.

Found by the adversarial review of PR #6630. Refs #6627, #6613, #6564.

cajasmota:
Closes #6633.

`default_branch_ref()` strips whitespace off the `QUALITY_BASE_REF` override, and nothing observed it. **Test-only** — `scripts/quality/ratchet.py` is byte-identical to `main` (`ce30c02a1bdc794549bb1ef7df862bf75ccaa57a`), verified with `shasum` after every mutant restore and before each push.

## Mutant scoring — five families, each scored separately

| mutant | on `main` | with this PR |
|---|---|---|
| `.strip()` removed entirely | **24 passed** (survivor) | **3 failed**, 24 passed |
| `.strip()` → `.rstrip()` | **24 passed** (survivor) | **3 failed**, 24 passed |
| `.strip()` → `.lstrip()` | **24 passed** (survivor) | **3 failed**, 24 passed |
| `.strip()` → `.strip(" \n")` | **24 passed** (survivor) | **3 failed**, 24 passed |
| strip moved past the emptiness test | — | **EQUIVALENT**, see below |

Pristine: 24 passed before, 27 after. Each live mutant kills all three new tests.

The fourth family is why `PADDED` carries **four** distinct whitespace characters, `"\t refs/remotes/origin/release\r\n"`: `.strip()` with no argument removes a *set*, so a padding drawn from part of that set cannot pin the set. Tab and CR both have real sources — a hand-indented `env:` line, and a CRLF-checked-out workflow file on any runner.

## What the new tests assert

`PaddedBaseRefOverrideIsHonoured`, on `make_repo_divergent_default_ref` (where the override `refs/remotes/origin/release` and the fallback `refs/remotes/origin/main` disagree in both directions):

1. `default_branch_ref()` **is** the override — the resolved ref itself.
2. `git_sha()` **is** the branch point, not the `"unknown"` degradation.
3. `ensure_durable_measured_on(branch_point)` does not raise, and writes nothing to stderr.

None reads the message text. Per #6630 the `SystemExit` f-string interpolates `ref`, so a message assertion reads back the ref the code *chose* and can go green on a call that consulted a different one.

Premises are asked of **git**, not of `ratchet.py`: the stripped spelling resolves; each of the four unstripped spellings (padded, `rstrip()`-only, `lstrip()`-only, `strip(" \n")`-only) does **not** — one per mutant family, so no mutant is equivalent on this fixture. The fallback ref is asserted to resolve, and the two refs to disagree about the branch point.

## Per-clause drop analysis (measured over all five families, 3 selected tests)

| clause dropped | no strip | `rstrip()` | `lstrip()` | `strip(" \n")` | verdict |
|---|---|---|---|---|---|
| (none) | 3 fail | 3 fail | 3 fail | 3 fail | — |
| pad right only | 3 fail | **SURVIVES** | 3 fail | 3 fail | leading whitespace load-bearing |
| pad left only | 3 fail | 3 fail | **SURVIVES** | 3 fail | trailing whitespace load-bearing |
| pad space + newline only | 3 fail | 3 fail | 3 fail | **SURVIVES** | tab + CR load-bearing |
| `origin/main` deleted | 2 fail | 2 fail | 2 fail | 2 fail | load-bearing for the durability test only — with no ref at all, `ensure_durable_measured_on` takes its second early return and accepts quietly |
| durability test deleted | 2 fail | 2 fail | 2 fail | 2 fail | **DECORATIVE** for this family |

On the decorative call, both halves are now stated in the docstring, because "decorative here, load-bearing elsewhere" is the distinction that keeps getting lost: it admits **no survivor** for any strip family, *and* it is the **suite-wide unique** killer of a mutant that consumes the override unstripped downstream of `default_branch_ref()`.

## Recorded equivalence (not a kill)

Moving the strip past the emptiness test — `if override:` on the raw value, `candidates.append(override.strip())` — is **EQUIVALENT**, recorded in the class docstring as such and explicitly not as DEAD. The only shape that could diverge is a whitespace-only value, where the mutant appends `""`, `rev-parse --verify "^{commit}"` fails, and the search falls through to exactly the ref the pristine code reaches by not appending at all. Re-measured here over 12 inputs, five of them whitespace-only: **0 divergences**.

## Ambient-environment coupling

`set_base_ref(case, value)` saves and restores `QUALITY_BASE_REF`, including the "was unset" case. All six sites now use it — the two bare `os.environ.pop(...)` calls in `SecondEarlyReturnIsReached`, plus the four `addCleanup(os.environ.pop, ...)` sites that deleted unconditionally instead of restoring.

Verified the way the finding was made — all 27 tests driven **one at a time**, each in its own process, under an ambient `QUALITY_BASE_REF=AMBIENT-SENTINEL`:

| | tests destroying the sentinel |
|---|---|
| before this fix (positive control) | **6** |
| after | **0** |

One correction to the review note: it reported exactly one destroyer, `AncestryFollowsTheDiscoveredRef.test_accepts_a_sha_only_the_discovered_ref_can_reach`. There were six — that test plus both of its siblings (all three route through the shared `_fixture`), and one each in `MeasuredOnIsDurable`, `CandidateRefsMustPeelToACommit`, and `AnnotatedTagCandidateIsNotADistinguisher`. The prescribed fix — all four sites — was right; the count was low.

## Note on the issue body

Everything in #6633 checked out; the witness table reproduces exactly, `"unknown"` included (under the mutant `merge-base` is taken against a disjoint `origin/main` and returns empty).

### DIFF
--- a/scripts/quality/test_ratchet.py
+++ b/scripts/quality/test_ratchet.py
@@ -359,6 +359,32 @@ def is_ancestor(repo, sha, ref):
     ) == 0
 
 
+def set_base_ref(case, value):
+    """Set (or, with `value=None`, unset) `QUALITY_BASE_REF` for one test and
+    RESTORE whatever was there before, ambient value included.
+
+    `os.environ.pop("QUALITY_BASE_REF", None)` with no restore drops an ambient
+    value for the remainder of the process, which is a cross-test coupling: the
+    next test to read it sees a different environment depending on whether this
+    one ran first. Nothing in this file depends on that today — which is
+    exactly why it would surface later, as an order-dependent failure in a test
+    that has nothing to do with the one that broke it.
+    """
+    prior = os.environ.get("QUALITY_BASE_REF")
+
+    def restore():
+        if prior is None:
+            os.environ.pop("QUALITY_BASE_REF", None)
+        else:
+            os.environ["QUALITY_BASE_REF"] = prior
+
+    case.addCleanup(restore)
+    if value is None:
+        os.environ.pop("QUALITY_BASE_REF", None)
+    else:
+        os.environ["QUALITY_BASE_REF"] = value
+
+
 def make_repo_divergent_default_ref(root):
     """A checkout whose default-branch ref is NOT `origin/main`, and where the
     two disagree about what is reachable (Refs #6627).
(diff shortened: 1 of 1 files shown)


cajasmota:
Merging. Five mutant families scored, the reviewer's survivor closed, and the environment coupling removed with a positive control behind it.

## My own mutant — the permissive direction

I ran one nobody had: let the override skip verification entirely.

```diff
     if override:
-        candidates.append(override)
+        return override
```

`python3 -m pytest scripts/quality/test_ratchet.py -q` → **1 failed, 26 passed**, killed by the pre-existing `CandidateRefsMustPeelToACommit::test_a_ref_that_resolves_but_is_not_a_commit_is_skipped`.

That is the answer I wanted. `QUALITY_BASE_REF` is the one input an operator can point anywhere, so "trust the override without checking it resolves to a commit" is the most attractive wrong simplification in this function — and it was already covered, by a test written for a different reason (#6613's tag-on-a-tree fixture). Worth recording that the earlier control is doing work nobody planned for it.

## The correction that matters

The review reported **one** test destroying an ambient `QUALITY_BASE_REF`. There were **six** — all three `AncestryFollowsTheDiscoveredRef` cases route through the shared `_fixture`, plus one each in `MeasuredOnIsDurable`, `CandidateRefsMustPeelToACommit`, and `AnnotatedTagCandidateIsNotADistinguisher`.

The prescribed fix was right and the count was wrong, and the difference is only visible because the fix was verified with a **positive control**: 6 destroyers before, 0 after. Without that, "one destroyer, now fixed" would have read as complete while five remained. Checking the before-count is the whole reason the discrepancy surfaced.

## The fifth family

`.strip()` → `.strip(" \n")` was a genuine survivor: the padding pinned space and newline only, while tab and CR are whitespace `.strip()` removes. `PADDED` is now `"\t refs/remotes/origin/release\r\n"` and the premise loop gained a fourth spelling, so the guards extend automatically rather than needing their own case. The drop matrix re-measured across all five families, with the new row confirming tab and CR are load-bearing rather than decorative padding.

The equivalence is recorded properly too — moving the strip past the emptiness test diverges only on a whitespace-only value, where the mutant appends `""` and `rev-parse --verify` fails identically. 12 inputs, 0 divergences, written down as EQUIVALENT with the reason and explicitly not as DEAD.

## And the scratchpad hazard was real

This agent's earlier harness had in fact been clobbered by another agent writing to the scratchpad root — the private-subdirectory instruction was not a precaution, it was a fix for something that had already happened twice today. Every script this round lived in `m6633/` with issue-tagged filenames.

`scripts/quality/ratchet.py` unchanged throughout at `ce30c02a…`, verified in the worktree before push and in the main checkout after. One file, no `.pyc`.

PR #6641, opened 2026-08-25
feat(#6560): B1 — routes!(a, b) minted nothing, so a multi-handler utoipa router contributed no endpoints

cajasmota:
Arm **B1** of #6560, reported by @arthurgeron. B2 stays open, so this deliberately does not close the issue.

## The gap

`utoipaRoutesMacroRe` matched a single identifier followed by the closing paren:

```go
`\broutes!\s*\(\s*([A-Za-z_]\w*)\s*\)`
```

so `routes!(list_items, create_item)` did not match at all and **neither** handler produced an `http_endpoint_definition` — even though both verbs and paths were declared statically in the same file, on the handlers' own `#[utoipa::path]` attributes. A router that registers its handlers in one grouped macro contributed no endpoints, so nothing got an IMPLEMENTS edge and no path could be matched by a caller or by `linkE2ERouteTestsToEndpoints`.

## The change

```go
`\broutes!\s*\(\s*([A-Za-z_]\w*(?:\s*,\s*[A-Za-z_]\w*)*)\s*,?\s*\)`
```

Capture 1 is the whole argument list; each argument is then resolved **independently** against the same-file attribute map, exactly as the single-handler case already was. A sibling that another producer owns, or that has no same-file contract, is skipped without suppressing the rest of the macro.

Everything stays **file-scoped**, so the per-file dedupe against `axumRouteRe` and `rocketRouteAttrRe` is untouched and the #6530 duplicate-producer hazard is not reopened.

## Two boundaries, and what actually distinguishes them

An earlier draft of this body justified the whole-macro skip on a path-qualified argument by calling the alternative "a silent half-registration" — which was rhetorically inconsistent, since a bare-name argument imported from another module produces exactly that half-registration and is accepted. The behaviour is right; that was the wrong reason for it. The real distinction is **what the text tells you about its own completeness**:

- `crate::items::create_item` is **self-evidently** a handler this pass cannot resolve — the `::` says so in the token itself. Accepting the macro while dropping that argument would be recording a router as fully read when the source visibly says otherwise. Rejecting the macro whole keeps recognition and resolution in agreement: this pass either understood every argument or claims none of them.
- A bare `create_order` imported from a sibling module is **indistinguishable** at this scope from a typo, a handler with no attribute, or one whose attribute is genuinely elsewhere. There is no signal to act on, so the only available behaviour is per-argument: mint what has a same-file contract, skip what does not. That is also what Arm A already did for the single-handler form, so it is not a new stance.

Both cases converge on the same guarantee, which is the one worth stating: **this pass never mints an endpoint whose verb and path it did not read in this file.** Nothing is fabricated in either direction.

## The B2 boundary is carried by the regex, not by a post-filter

No `(`, `)` or `:` appears in any character class, which gives three properties for free:

- a **path-qualified** argument makes the whole macro fail to match, per the reasoning above;
- a match can never **span** from one `routes!(...)` into the next, since neither paren is admissible mid-capture;
- the trailing `\)` stays a required **anchor**, so an unterminated macro registers nothing.

`\s` spans newlines, so the rustfmt multi-line form with a trailing comma reads normally. Three narrower shapes also fail to match, all in the safe direction and none a regression from Arm A — a comment between arguments, a raw identifier (`r#type`), and `routes ! ( a , b )` with Rust-legal whitespace before the `!`. These are now recorded in the header scope note so the next reader does not rediscover them.

## Replacing the pin rather than deleting it

`TestUtoipaAxum_MultiHandlerRoutesMacroNotMinted` asserted **0** definitions for both handlers of `routes!(a, b)`. That pin was honest when written — Arm A excluded the multi form deliberately, because minting only the first of N drops routes the source declared statically and in full view.

It is **inverted, not removed**: `TestUtoipaAxum_MultiHandlerRoutesMacroMinted` asserts each handler mints its own definition with the right verb and path, and its doc comment names the Arm A limitation and the B1 ruling that lifts it. A silently deleted pin is indistinguishable from a regression.

## Premise guards

Every positive fixture asserts the `framework` property **per handler**, not just the endpoint ID or the count. `utoipaRegisteredElsewhere` skips any handler another producer already registered from that file, so a fixture that also carried a `.route()` mount would have observed a passing test with the branch under test never executing. `framework=utoipa_axum` proves this pass is what minted; the dedupe test asserts `framework=axum` for the handler that `.route()` owns and `framework=utoipa_axum` for its sibling in the same macro.

*(Round 2: this claim was false when first written — `TestUtoipaAxum_HandlerRepeatedAcrossMacrosMintedOnce` asserted only `countDefsForHandler == 1`, which is satisfiable by this pass minting none and another producer minting one. That is precisely the test the M6 argument below leans on, so the missing guards are added rather than the claim softened.)*

## Mutants scored

Full package, `-count=1`, no `-run` filter. Every one reproduced RED before the fix / after the mutant.

|---|---|---|---|
| M1 | revert to single-identifier regex | **RED** | 5 tests incl. `MultiHandlerRoutesMacroMinted` |
| M2 | drop the trailing `\)` anchor | **RED** | `MultiHandlerPathQualifiedNotMinted`, `UnterminatedRoutesMacroNotMinted` |
| M3 | admit `::` into arguments (widen past B1 into B2) | **RED** | `MultiHandlerPathQualifiedNotMinted` |
| M4 | greedy `([A-Za-z_][^;]*)` capture slurping past the macro | **RED** | 10 tests, incl. the pre-existing Arm A ones |
| M5 | mint only the first argument of N | **RED** | 5 tests |
| M6 | drop the in-pass `minted[handler]` guard | **EQUIVALENT** | see below |
| M7 | drop the `registeredElsewhere` dedupe | **RED** | `DedupeAgainstRouteCall`, `RocketAttributeSameFileNotDoubleMinted`, `MultiHandlerDedupeAgainstRouteCall` |
| M8 | resolve a qualified argument by its last `::` segment | **RED** | both path-qualified tests |
| **N1** | **`continue` → `break` on an unresolvable argument** | **RED** *(survivor in round 1)* | `MixedResolvableAndImportedArguments` |

**N1 was a genuine survivor**, caught in review: `go vet` 0 and the full package green. It abandons the whole macro at the first argument with no same-file contract, which means the one behaviour B1 actually introduces — independent per-argument resolution — was unobserved. Every fixture in round 1 had a uniformly resolvable or uniformly unresolvable argument list, so none could see it.

`TestUtoipaAxum_MixedResolvableAndImportedArguments` closes it with the shape that motivated the whole review: a router module aggregating handlers imported by bare name from sibling modules, which is the idiomatic utoipa_axum layout rather than a corner case. Unresolvable arguments sit **first and in the middle** of four, so an early exit at any position is visible; both same-file siblings must still mint with `framework=utoipa_axum`, and the imported pair must mint nothing. Survivor reproduced (exit 0), then RED under the same mutant.

**M6 is equivalent, not merely unkilled**, and the mechanism is worth recording: `dedupKey` collapses on verb+path, and `utoipaHandlerRoutes` is first-wins per handler, so one handler can only ever produce one `(verb, path)` pair no matter how many times it is named. The in-pass `minted` guard is therefore unobservable defence-in-depth that predates this change. The behaviour — one handler never yields two definitions — *is* pinned, by `HandlerRepeatedAcrossMacrosMintedOnce`, at the artefact rather than at the pass's internal bookkeeping.

Two further mutants are likewise equivalent by construction: dropping the `h != ""` guard in `utoipaMacroHandlerArgs` (capture 1 always begins and ends with `\w` and cannot contain adjacent commas, since the trailing comma is matched outside the group), and `canonical == "" continue → break` (unreachable — `utoipaAttrPathRe` requires a non-empty path and `Canonicalize` always returns a `/`-prefixed string).

## Nothing was relaxed to make a test pass

The one place widening was tempting is `routes!(crate::items::list)`: trivially matchable, and the shape most likely to appear next in a real repo. It is left minting nothing, because resolving it means the attribute map stops being per-file, and the file scope is precisely what stops the cross-file double mint the header at `http_endpoint_utoipa_axum.go:54-58` still admits is unfixed. That is the dedupe-scope decision B2 is waiting on. A gap is visible; a phantom is not.

## Verification

`gofmt -l internal/engine/` clean, `go vet ./internal/engine/` clean, `go test -count=1 ./internal/engine/` green (full package). The round-2 non-test diff is comment-only.

Refs #6560.

### DIFF
--- a/internal/engine/http_endpoint_utoipa_axum.go
+++ b/internal/engine/http_endpoint_utoipa_axum.go
@@ -63,13 +63,16 @@
 // merely be duplicated — it would be relabelled framework=utoipa_axum. The
 // Rocket guard tests assert the framework stamp for exactly that reason.
 //
-// Scope — Arm A only (#6560)
-// --------------------------
-// Deliberately NOT handled here; these are Arm B and are left to emit nothing
+// Scope — Arms A and B1 only (#6560)
+// ----------------------------------
+// Arm B1 added the multi-handler form `routes!(a, b, c)`: every argument is
+// resolved against the same-file attribute map exactly as the single-handler
+// case already was, and the macro is skipped WHOLE if any argument is not a bare
+// identifier. Everything stays file-scoped, so the dedupe below is untouched.
+//
+// Deliberately NOT handled here; these are Arm B2 and are left to emit nothing
 // rather than to emit a guess:
 //
-//   - multi-handler `routes!(a, b)` — the regex requires a single identifier
-//     followed by the closing paren, so the multi form simply does not match.
 //   - cross-module handlers (`routes!(crate::items::list)`, or a bare name whose
 //     `#[utoipa::path]` attribute lives in another file) — the attribute map is
 //     built from THIS file only, and an unknown handler is skipped. A fabricated
@@ -93,13 +96,44 @@ import (
 	"github.com/cajasmota/grafel/internal/engine/httproutes"
 )
 
-// utoipaRoutesMacroRe matches the SINGLE-handler `routes!(handler)` form.
+// utoipaRoutesMacroRe matches a `routes!(a)` / `routes!(a, b, c)` registration
+// whose arguments are ALL plain identifiers (#6560 Arm B1 widened this from the
+// single-identifier form Arm A shipped).
 //
-// The trailing `\s*\)` is what confines this pass to Arm A: `routes!(a, b)`
-// does not match, and neither does a path-qualified `routes!(crate::x::h)`.
+// Capture group 1 = the whole argument list, split by utoipaMacroHandlerArgs.
 //
-// Capture group 1 = handler identifier.
-var utoipaRoutesMacroRe = regexp.MustCompile(`\broutes!\s*\(\s*([A-Za-z_]\w*)\s*\)`)
+// Two properties of this pattern carry the Arm B2 boundary and are the reason it
+// is written as one alternation rather than as a loose `([^)]*)` capture:
+//
+//   - The argument list admits ONLY `[A-Za-z_]\w*` separated by commas, so a
+//     path-qualified argument (`crate::items::list`) makes the WHOLE macro fail
+//     to match. Such a handler's `#[utoipa::path]` attribute lives in another
+//     module, and this pass's attribute map is file-scoped, so it could not be
+//     resolved anyway — and minting only the macro's plain arguments would be a
+//     silent half-registration, the exact failure the Arm A pin forbade.
+//   - The trailing `\)` is a required anchor. Without it the capture runs past
+//     the macro it belongs to and registers identifiers that are not arguments
+//     at all.
+//
+// Neither `(` nor `)` is in any character class here, so a match can never span
+// from one `routes!(...)` into the next; `\s` spans newlines so the rustfmt
+// multi-line form is read normally. A trailing comma is Rust-legal and accepted.
+var utoipaRoutesMacroRe = regexp.MustCompile(
+	`\broutes!\s*\(\s*([A-Za-z_]\w*(?:\s*,\s*[A-Za-z_]\w*)*)\s*,?\s*\)`)
+
+// utoipaMacroHandlerArgs splits capture group 1 of utoipaRoutesMacroRe into its
+// handler identifiers, in source order. The regex has already guaranteed every
+// element is a bare identifier, so this only has to trim.
+func utoipaMacroHandlerArgs(argList string) []string {
+	parts := strings.Split(argList, ",")
+	out := make([]string, 0, len(parts))
+	for _, p := range parts {
+		if h := strings.TrimSpace(p); h != "" {
+			out = append(out, h)
+		}
+	}
+	return out
+}
 
 // utoipaPathAttrStartRe matches the start of a `#[utoipa::path(` attribute. The
 // argument list is then read with a depth-counted paren scan because it nests
@@ -246,9 +280,11 @@ func utoipaRegisteredElsewhere(content string) map[string]bool {
 	return out
 }
 
-// synthesizeUtoipaAxumRoutes scans a Rust source file for single-handler
-// `routes!(handler)` registrations and emits one http_endpoint_definition per
-// handler whose `#[utoipa::path(...)]` attribute is in the SAME file.
+// synthesizeUtoipaAxumRoutes scans a Rust source file for `routes!(...)`
+// registrations — one handler or several — and emits one
+// http_endpoint_definition per handler whose `#[utoipa::path(...)]` attribute is
+// in the SAME file. A handler named more than once, in one macro or across
+// several, still mints exactly once.
 //
 // Handlers that another producer already registers out of this file — a
 // `.route(...)` call or a Rocket verb attribute — are skipped; see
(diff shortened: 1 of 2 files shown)


cajasmota:
Merging. Both blockers closed, and the rationale is now consistent instead of merely defended.

## My own mutant — granularity, not presence

The review scored dropping `registeredElsewhere` entirely (M7, RED). I ran the subtler variant nobody had: consult it **once per macro** on the first argument, rather than per argument.

```diff
-		for _, handler := range utoipaMacroHandlerArgs(m[1]) {
-			if registeredElsewhere[handler] || minted[handler] {
+		margs := utoipaMacroHandlerArgs(m[1])
+		if len(margs) > 0 && registeredElsewhere[margs[0]] {
+			continue
+		}
+		for _, handler := range margs {
+			if minted[handler] {
```

`go vet` 0, full package `go test -count=1 ./internal/engine/` **exit 1 — DEAD**, sole failure `TestUtoipaAxum_MultiHandlerDedupeAgainstRouteCall`.

That is the assertion I wanted: the dedupe is pinned at **per-argument granularity**, not merely as present-or-absent. A macro whose first handler is owned by another producer must not suppress its siblings, and a sibling that is owned must not be minted just because the first one wasn't. Both directions are observed.

## The fixture that closed N1 is better than the one I asked for

I asked for a macro mixing a resolvable and an unresolvable argument. What landed puts the unresolvable ones **first and in the middle** of a four-argument list, so an early exit is visible at any position rather than only at the head. It is also the only fixture in the file with a non-uniform argument list — which is exactly why nothing observed N1 before, and worth saying out loud: every previous fixture was uniform, so "resolve each argument independently" had no fixture that could distinguish it from "resolve them all or none".

## The rationale, dropped rather than defended

The half-registration argument did not survive the imported-bare-name case and has been replaced by what actually distinguishes the two:

> `crate::items::create_item` is **self-evidently** unresolvable — the `::` says so in the token — so accepting the macro while dropping that argument would record a router as fully read when the source visibly says otherwise. A bare `create_order` is **indistinguishable** at this scope from a typo or an un-annotated handler, so there is no signal to act on.

Both converge on one statable guarantee: *this pass never mints an endpoint whose verb and path it did not read in this file.* That is a better sentence than the one it replaces because it is checkable.

## The body now says it was wrong

The "every positive fixture asserts `framework` per handler" claim was false when written — `HandlerRepeatedAcrossMacrosMintedOnce` asserted only `count == 1`, which would pass if this pass minted none and another producer minted one. It is now true, and the body carries a parenthetical saying it was false rather than quietly becoming correct. That is the right way to fix a claim: a body that silently repairs itself teaches the next reader nothing.

## Recorded as equivalent, with mechanisms

M6 (`minted` guard), the `h != ""` guard in `utoipaMacroHandlerArgs`, and `canonical == "" continue → break` — all equivalent, each with the reason stated rather than asserted. M6's is the useful one: `dedupKey` collapses on verb+path and `utoipaHandlerRoutes` is first-wins per handler, so one handler can only ever produce one `(verb, path)`. Killing it would have meant asserting a raw emit count — testing the counter instead of the artefact.

Three fail-to-mint shapes are now in the header scope note (comment between arguments, raw identifier, `routes ! ( a , b )`), all in the safe direction and none a regression from Arm A.

**B2 remains open on #6560 and is ours** — cross-file resolution and `OpenApiRouter::nest`, after the dedupe-scope decision that widening the attribute map forces.

PR #6642, opened 2026-08-25
docs(#6560): B1 stale-dated Arm C — the utoipa coverage cells cited an inverted test and a lifted limitation

cajasmota:
Refs #6560 — Arm C refresh after B1. **Not** closing: B2a/B2b stay open.

Merging B1 as `954e57786` stale-dated the coverage record Arm C (`ef36c0c30`) had just corrected. Both cells of `docs/coverage/detail/lang.rust.framework.utoipa.md` and their `registry.json` source still cited `TestUtoipaAxum_MultiHandlerRoutesMacroNotMinted` at `http_endpoint_utoipa_axum_test.go:336` — a test B1 **inverted** — and still described `routes!(a, b)` as minting nothing, a limitation B1 **lifted**.

## What changed

Two `notes` fields and two `verified_at` fields, in `registry.json` and the generated `.md`, kept byte-identical. 6 changed lines total; no re-serialisation of `registry.json`.

Every citation in both cells was re-read against the tree rather than carried over, which turned up more drift than the two known items:

| cited as | actually |
|---|---|
| `TestUtoipaAxum_MultiHandlerRoutesMacroNotMinted` `_test.go:336` | `TestUtoipaAxum_MultiHandlerRoutesMacroMinted` `:367` |
| `synthesizeUtoipaAxumRoutes` `:261` | `:304` |
| `utoipaRegisteredElsewhere` `:232` | `:273` |
| `utoipaRoutesMacroRe` `:102` | `:128-129` |
| scope note `:66-79` | `:66-96` |
| pagination gate `http_endpoint_pagination.go:80` | `:79` (`:80` is the `continue`) |

Verified as still accurate and left alone: `internal/custom/rust/utoipa.go:151`, `:158-162`, `:164-165`, `:174`; `internal/custom/rust/axum.go:143-144` and `:148-149`; `endpoint_response_codes.go:29`/`:587`; `http_endpoint_response_codes.go:100`; the dispatch at `http_endpoint_synthesis.go:1350`.

## Remaining limitations, re-stated from the pass's own header

Arm B2, unchanged by B1: path-qualified arguments `routes!(crate::items::list)` (`:115-120`), cross-file handlers (`:76-79`), `OpenApiRouter::nest` prefix composition (`:80-82`), cross-file double mint (`:54-58`).

Plus the three fail-to-mint shapes B1 recorded at `:84-89` — a comment between arguments, a raw identifier `routes!(r#type)`, and `routes ! ( a , b )` with whitespace before `!`. Both cells now say these are observed by **no test**; that is what the header says and what a `grep` of the test file confirms, and neither cell previously said it.

Both cells also now record what B1 actually bought, cited: per-argument resolution that skips an unresolvable sibling without suppressing the rest (`:323-331`, `_test.go:724`) and mint-once across repeated names (`:315`/`:336`, `_test.go:550`, `:586`).

## Grade: held at `🟢 partial`, argued rather than left

Not raised. The sibling bar reserves `partial` for a statically-resolvable subset with an unresolved **composition** form — `lang.rust.framework.warp.md:21` is `partial` precisely because `.or()` filter-tree composition is not statically attributable, and `tower.md:21` / `hyper.md:21` sit at `partial` for less. `OpenApiRouter::nest` composition (`http_endpoint_utoipa_axum.go:80-82`) and cross-file handler resolution (`:76-79`) are exactly that class, so widening the argument list does not clear the bar. The reasoning is now written into both cells, so the next reader does not have to re-derive it.

Not lowered either: B1 is a real widening and both cells say so, with the positive pins cited.

## Scope

Docs only — nothing outside `docs/` is touched, and no code defect was found in the pass while reading it.

`utoipa_axum` and the inverted test name appear in exactly two files under `docs/`, both in this diff; no other coverage doc carries the stale claim.

**A `go run ./tools/coverage gen` no-op check is still required and was not run on this lane.**


cajasmota:
Merging. Generator gate run on this head: `go run ./tools/coverage gen` → 822 records, `git status` clean, so the hand-edited `registry.json` and the rendered `.md` agree with the generator byte for byte.

I spot-checked the corrections rather than accepting them, including the one that contradicts my own brief.

## Six stale citations, not the two I named

I sent this out to fix a test name and a lifted limitation. Reading every citation instead of carrying them over found four more — `synthesizeUtoipaAxumRoutes` `:261`→`:304`, `utoipaRegisteredElsewhere` `:232`→`:273`, `utoipaRoutesMacroRe` `:102`→`:128-129`, the scope note `:66-79`→`:66-96`, and `http_endpoint_pagination.go:80`→`:79` (`:80` is the `continue`, not the `Kind !=` gate).

That last one is the useful detail: it was already stale **before** B1, and Arm C introduced it while fixing a different citation defect. So the drift is not one bad merge — it is that nothing checks whether a cited line still says what the citation claims, and the doc most vulnerable to that is the one whose entire value is its citations.

## The correction to my brief

My brief said the macro is skipped whole when an argument is unresolvable. The code says two different things and I had blurred them — verified myself:

- `:70-71` — the macro is skipped **whole** if any argument is not a bare identifier. That is regex-level, and it is the path-qualified case.
- `:323-331` — an **unresolvable but bare** identifier is skipped **per-argument**, `continue`, siblings unaffected. Pinned by `_test.go:724`.

Both cells now state them separately. This is the third time this milestone that reading the code beat repeating my summary of it, and the right response is the one taken here: say the brief was wrong rather than quietly do the correct thing.

## Grade held at partial, argued rather than assumed

`warp.md:21` is `partial` for an unresolved *composition* form; `tower.md:21` and `hyper.md:21` sit at `partial` for less. `OpenApiRouter::nest` composition and cross-file handler resolution are that same class, so widening the argument list does not clear the bar. Not lowered either — both cells now cite B1's positive pins and what it actually bought: per-argument skip that does not suppress siblings, and mint-once across repeats.

## The finding I did not ask for

**The three fail-to-mint shapes are observed by no test** — no `r#`, no `/*`, no `routes !` anywhere in the test file; they exist only in the header comment. Both cells now say so explicitly, where neither said it before.

That is this milestone's dominant defect class, one level up: a scope note asserting behaviour that nothing checks, inside the file whose header is the source of truth for the coverage doc. Not filing it as a bug — they are documented limitations in the safe direction — but a doc that says "this is claimed and unobserved" is worth more than one that repeats the claim.

PR #6644, opened 2026-08-25
fix(#6639): a test left group "g" in the docgen run registry, so the next test to use that name resumed its run

### LINKED ISSUE #6639: an internal/mcp test writes docgen staging into the live worktree, so the package fails under some shuffle orderings
A test in `internal/mcp` writes docgen staging state into the **live working tree** at `<repo>/.grafel/staging` (`docgen.go:102`) despite `sandboxGrafelHome(t)`, and a later docgen test in the same run resumes it. The package therefore fails under some test orderings.

## Reproduction

```
go test -count=1 -shuffle=1724578000 ./internal/mcp/   → exit 1
```

Sole failure: `TestWorkflowDocgenDispatch`, `meta_merges_test.go:53` — the response carries `"resumed":true` and a `staging_path` pointing at the worktree root's `.grafel/staging` rather than a `t.TempDir()`.

Under `-shuffle=on` with a random seed the package passes (exit 0, ~129s), so this is **order-dependent, not deterministic**. Reproduced with `.grafel` wiped beforehand, which rules out cross-run residue: it is intra-run coupling between two tests in the same package.

## Why it is worth fixing rather than reseeding around

1. **It escapes the sandbox.** `sandboxGrafelHome(t)` exists precisely so a test cannot touch the developer's real state, and this path goes around it into the checkout. On a developer machine that means a test run mutates the repo it is being run from; in a worktree-per-agent setup — which this project uses heavily — it means concurrent agents share one staging directory.
2. **The failure is a false accusation.** It surfaces as a docgen *resumption* bug in whichever test happens to run second, so the test that reports the problem is not the test that caused it. That is expensive to diagnose from a CI log, and it is the kind of order-dependent flake that gets reseeded away rather than fixed.
3. **This project has a standing cost here.** Windows-only timing failures already number four, and each one makes the next genuine regression less likely to be believed. An order-dependent failure in the package that serves the MCP tools has the same corrosive effect.

## What is needed

1. Find the test that leaves `<repo>/.grafel/staging` behind — start from `docgen.go:102` and its callers under `internal/mcp`.
2. Route it through the sandbox so the staging root is inside `t.TempDir()`, and assert that it is: the fix must be observable, not just applied.
3. **Add a guard that fails loudly if any test writes under the repo root's `.grafel/`.** That is the property worth pinning — one test cleaning up after itself does not stop the next one from doing the same thing.
4. Confirm with `-shuffle=1724578000` (the seed above) **and** a handful of other seeds. A single passing seed proves nothing; that is how this survived until now.

## Rules

- **Score it before and after**: the named seed must fail on `main` (reproduce the survivor first) and pass after. A fix not first reproduced as a failure is not evidence.
- Do **not** simply pin the shuffle seed or disable shuffling — that hides the coupling rather than removing it.
- Do not widen into unrelated docgen behaviour; the target is where the staging root comes from.
- The `internal/mcp` suite runs ~120-190s under load. Slow is contention, not a hang.

## Provenance

Pre-existing; found by the adversarial review of PR #6634 while running `-shuffle=on` against that PR. Explicitly **not** caused by #6634 — its four tests create no staging state (`-run '6314'` against a clean `.grafel` leaves `.grafel/staging` non-existent).

Refs #6314.

cajasmota:
Closes #6639

Test-only. `git show --stat` touches five `*_test.go` files and nothing else.

## Reproduced before fixed

`go test -count=1 -shuffle=1724578000 ./internal/mcp/` on `29d9b8e16`, in a fresh worktree with no `.grafel/` residue:

```
--- FAIL: TestWorkflowDocgenDispatch (0.41s)
    meta_merges_test.go:53: action=start did not route to handleDocgenStartRun:
      {"resumed":true,"run_id":"2026-08-25-73f7b4ee",
       "staging_path":"<worktree>/.grafel/staging/2026-08-25-73f7b4ee"}
```

Sole failure. The aged staging directories reported in the first issue comment are disk residue of past runs, not the trigger — this reproduction had none.

## The production code did not move

`stagingRootPath` (`docgen.go:99-101`) anchors on the **project** root, derived from git — a different axis from `sandboxGrafelHome`, which redirects `HOME`/`USERPROFILE`/`GRAFEL_HOME`. Project-anchored staging is what makes cross-process resume possible, so it stays exactly where it is.

## Item 1 alone does not fix it — measured, not assumed

With only `"cwd": dir` added at `server_test.go:3237`, the same seed **still fails**; only the path changes:

```
{"resumed":true,"run_id":"2026-08-25-1413473d",
 "staging_path":"/var/folders/.../TestElapsedMSCoverageAllTools1732032595/001/.grafel/staging/..."}
```

The resume branch never consults the disk. It consults `docgenRunsByGroup`.

## What actually fixes it

`TestElapsedMSCoverageAllTools` left an entry under group `"g"` in the package-level `docgenRunsByGroup` (`docgen.go:52-53`); `TestWorkflowDocgenDispatch` uses the same literal group name and took the resume branch (`docgen.go:223-234`).

**Drain per test, and assert globally — they do different jobs.**

- `newDocgenServer` already cleared the map, but on the way **in**. Clearing at setup protects the test doing the clearing from its predecessors and does nothing for its successors — which is the direction this failure travels. `releaseDocgenRuns(t)` drains on the way **out**, via `t.Cleanup`, so a test's effect on package state ends when the test does.
- Draining alone would fix today's failure and leave the property unobserved. So `TestMain` also asserts the map is empty after `m.Run()`. That assertion cannot fix anything — it runs last by construction; its job is to turn "no test leaks docgen run state" from a convention into a checked property. Asserting the **map** rather than statting the staging directory is deliberate: the map is the coupling, the directory is only its artefact, and a test can break this package through the map without touching disk at all.

A second `TestMain` check pins item 1: staging directories created inside the checkout **during** the run. It snapshots the entries before `m.Run()` and reports only new ones, so pre-existing residue (this checkout had two such directories, dated 28 and 31 July, invisible because `.grafel/` is gitignored) cannot masquerade as a leak. It derives the root through the production helpers the polluter itself went through, and disables itself rather than guessing if that derivation fails.

## The worse bug (item 3)

`TestElapsedMSCoverageAllTools` isolated **no** home while driving `grafel_save_finding`, `grafel_persona_event`, `grafel_feedback_event`, `grafel_mcp_metrics` and `grafel_apply_doc_semantics`, whose writers resolve under `registry.HomeDir()`. Every run wrote findings, telemetry and metrics into the developer's real `~/.grafel`. It sits exactly in the blind spot the existing guard documents about itself (`home_isolation_guard_6288_test.go:43-45`): the guard is keyed on the presence of a `Setenv("HOME", …)` call and says nothing about a test that isolates nothing at all. Now calls `sandboxGrafelHome(t)`.

**Positive control.** Every one of the package's **1295** top-level tests (`-test.list '.*'`, same on both revisions) was run alone against a fresh surrogate `HOME`, before and after:

| | tests writing under a real `HomeDir()` |
|---|---|
| before | **33** / 1295 |
| after | **32** / 1295 |

The single removed row is `TestElapsedMSCoverageAllTools`; no other row changed in either direction. **The gap is endemic, not local** — the other 32 are pre-existing and are deliberately left for a separate issue rather than widened into this PR.

## Both guards positively controlled

| mutant | result |
|---|---|
| remove `"cwd": dir` | `mcp: 1 staging dir(s) written into the CHECKOUT during this run (#6639)` — exit 1 with **both tests passing**; the guard alone killed it |
| remove `releaseDocgenRuns(t)` from the polluter | the original `"resumed":true` failure returns |
| same, run with only the polluter selected | `mcp: 1 docgen run(s) left in docgenRunsByGroup after the suite (#6639)`, naming group `"g"` |

## Seeds

Full package, no `-run` filter, `-count=1`:

| seed | result |
|---|---|
| 1724578000 (the named one) | FAIL on `main` → **ok** after |
| 20260826 | ok |
| 7 | ok |
| 424242 | ok |
| 99999 | ok |

No seed was pinned and shuffling was not disabled. No `.grafel/` directory exists in the worktree after all five runs. `gofmt -l internal/mcp/` clean; `go vet ./internal/mcp/` clean.

---

## Review follow-up (second commit)

The guards above were themselves checked by nothing — each reporter had exactly one caller (`TestMain`) and nothing captured its stderr. Two non-equivalent mutants survived the committed suite; both are now dead, scored before and after on the full package with no `-run` filter:

| mutant | before | after |
|---|---|---|
| M1 `reportNewRepoStaging`: `!before[name]` → `before[name]` | **exit 0 — survives** | exit 1, `TestReportNewRepoStagingSeparatesResidueFromLeaks` |
| M2 `reportLeakedDocgenRuns`: `len(docgenRunsByGroup) == 0` → `true` | **exit 0 — survives** | exit 1, both map-guard tests |
| M3 drop `sort.Strings(groups)` | — | exit 1, `TestReportLeakedDocgenRunsSortsGroups` |
| M4 drop the group name from the report line | — | exit 1, both map-guard tests |

M1 is the worse of the two: inverting the residue filter turns the guard into a residue **accuser**, failing the package with every test passing on any checkout carrying an aged staging dir — which the checkout this was filed from does.

M3 is scored on **six groups over eight reports**, not one report of three. Go randomises map iteration, so deleting the sort leaves a 1-in-6 chance of coming out ordered anyway with three keys; a kill on that premise would be scheduling luck rather than evidence.

Both helpers already took their input as parameters (an `io.Writer`, and `root`+`before`), so this is two table tests rather than a redesign. Seeding the package global is deterministic: the only reader of `docgenRunsByGroup` is `handleDocgenStartRun`, and no test that reaches it calls `t.Parallel()` (`docgen_test.go` and `meta_merges_test.go` have none; `server_test.go`'s single `t.Parallel` is in `TestMCPInstructionsOrientationMap`).

### Documented, no behaviour change

- The staging guard watches `.grafel/staging` only — narrower than #6639 item 3's "any test writes under the repo root's `.grafel/`". A leak into `.grafel/store` or `.grafel/events` escapes it, as does a staging dir created and removed within the run.
- Neither new guard survives a **panic**: both report after `m.Run()`, and a panic kills the binary before `os.Exit`. The pre-existing home guard runs *before* `m.Run()` precisely to survive one; these two cannot, because what they assert is a property of the finished run. A panicking suite is already failing loudly, so the asymmetry costs a diagnosis rather than a detection.
- `releaseDocgenRuns` **keeps** draining the whole map rather than the caller's groups, matching what `newDocgenServer` already does on entry: a caller cannot reliably know which group names its own table produced (the polluter's `"g"` came from a table shared with 30 other tools). The cost — no test holding a docgen run may call `t.Parallel()` — is now stated in the doc comment rather than implied.

Full package after the follow-up: exit 0; seed 1724578000 still passes; no `.grafel/` in the worktree afterwards; `gofmt` and `go vet` clean.

### DIFF
--- a/internal/mcp/docgen_run_registry_6639_test.go
+++ b/internal/mcp/docgen_run_registry_6639_test.go
@@ -43,13 +43,30 @@ package mcp
 // convention. Asserting the map rather than statting the staging directory is
 // deliberate: the map is the coupling, the directory is only its artefact, and
 // a test can fail this package through the map without touching disk at all.
+//
+// # What these guards do NOT cover
+//
+// The staging guard watches <repo>/.grafel/staging and nothing else, which is
+// narrower than #6639 item 3 asked for ("any test writes under the repo root's
+// .grafel/"): a leak into .grafel/store or .grafel/events escapes it, and so
+// does a staging directory that a test creates and removes within the run.
+//
+// Neither guard survives a PANIC. Both report after m.Run(), and a panic kills
+// the test binary before os.Exit is reached — the pre-existing home-isolation
+// guard runs BEFORE m.Run() precisely so its diagnosis outlives one (see the
+// header of home_isolation_guard_6288_test.go). These two cannot: what they
+// assert is a property of the FINISHED run, so there is nothing to report
+// until it has finished. A panicking suite is already failing loudly, so the
+// asymmetry costs a diagnosis rather than a detection.
 
 import (
+	"bytes"
 	"fmt"
 	"io"
 	"os"
 	"path/filepath"
 	"sort"
+	"strings"
 	"testing"
 )
 
@@ -60,6 +77,13 @@ import (
 // run stays in the map until an abort or promote removes it, and the tests
 // that deliberately leave a run in flight are the common case rather than the
 // exception.
+//
+// It drains the WHOLE map, not just the caller's groups, matching what
+// newDocgenServer already does on entry. Kept that way deliberately: a caller
+// does not reliably know which group names its own table produced (the #6639
+// polluter's "g" came from a table shared with 30 other tools). The cost is
+// that no test holding a docgen run may call t.Parallel() — none does today,
+// and the alternative is a per-group API that silently under-drains.
 func releaseDocgenRuns(t *testing.T) {
 	t.Helper()
 	t.Cleanup(func() {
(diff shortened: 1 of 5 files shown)


cajasmota:
Merging. The guards are now guarded, and the diagnostic text is pinned as an artefact rather than trusted as prose.

## My own mutant — the staging guard's names

The rework scored M4 (dropping the group name from the *map* report). I ran its counterpart on the *staging* report, which nobody had:

```diff
-	for _, name := range added {
-		fmt.Fprintf(w, "mcp:   %s\n", filepath.Join(root, name))
-	}
 	return len(added)
```

`go vet` 0, full package **exit 1 — DEAD**, failing `…SeparatesResidueFromLeaks/a_dir_created_during_the_run_is_reported_by_name` and `…/every_new_dir_is_counted`.

So both halves of both reports are observed: the count *and* the names. That matters more here than usual — the file's own prose says the names are "the diagnosis", and a guard that fires with an accurate count and no names tells you the suite is dirty without telling you which test dirtied it. That is a guard that costs a debugging session rather than saving one.

## The M3 handling is the best thing in this PR

Map iteration is randomised, so deleting `sort.Strings` has a real chance of emerging ordered by luck — with three keys, roughly one run in six. A kill scored that way is scheduling luck wearing a green tick.

The rework used **six groups over eight reports**, survival probability under 1e-23, and wrote down why in the doc comment. Nobody asked for that; the brief said "score M3". Getting a deterministic premise for a kill without being told is the standard this milestone has been trying to reach, and it is the difference between a test that passes and evidence that a property holds.

## What the review found and what closed it

The PR's contribution was turning "no test leaks docgen run state" from a convention into a checked property — but the checkers themselves were unobserved: inverting one character in the staging comparison (`!before[name]` → `before[name]`), or disabling the map guard outright, both survived the committed suite. The inverted form was a false-accusation shape with a live trigger on my own checkout, which carries two aged staging dirs.

Four table tests on the pure helpers close M1–M4 at no suite-time cost, using seams that were already there. `swapDocgenRuns` substitutes and restores the global so the new tests cannot become the leak they detect — the right instinct for a test that manipulates package state.

## Corrections carried, not quietly absorbed

- The top-level test count is **1295**, not 1280; the numerator (33 → 32) was exact. Fixed in all three places rather than in one.
- Both caveats are now file-header prose: the guard covers `.grafel/staging` only, so a leak into `.grafel/store` or `.grafel/events` — or a staging dir created and removed within the run — escapes it; and neither new guard survives a **panic**, because that kills the binary before `os.Exit`, whereas the pre-existing home guard runs before `m.Run()` specifically to survive one.
- `releaseDocgenRuns` keeps the whole-map drain, with the reason stated: a caller cannot reliably know which group names its own table produced — the polluter's `"g"` came from a table shared with 30 tools — so a per-group API would silently under-drain. The non-parallelisable cost is now written down instead of implied.

## On the stall

This agent stalled mid-scoring with M1 still applied, and I restored it from the committed blob before resuming. The cause turned out to be its own scoring script `rm -rf`-ing a surrogate `HOME` that had become `GOPATH`, then spending 600s failing on read-only module-cache files — diagnosed and fixed by pinning the real `GOPATH`/`GOMODCACHE`/`GOCACHE`. Worth recording because the symptom (a silent 10-minute hang) looked nothing like the cause.

Five test files, +446/-2, no production file touched. Seed 1724578000 green, full package green, no `.grafel/` left in the worktree afterwards.

I will delete the two aged staging directories from my checkout now that this is merged.

Issue #100, opened 2026-05-09
markdown CONTAINS edges become bug-extractor for every README heading (Refs #91)


cajasmota:
Across every language repo in the post-#96 corpus, the markdown extractor emits Document --CONTAINS--> <path>::<heading-slug> edges. The slug ToIDs are not registered as resolvable entity targets, so the resolver classifies them as bug-extractor (categories: structural-ref / bare-kind-prefixed).

Histogram evidence (issue #91 report):
- go-gin: 3,395 of 7,555 bug-extractor edges (45%) are markdown structural-refs
- go-chi: 987 of 2,396 (41%)
- php-symfony-demo: 202 of 265 (76%)
- csharp-aspnetcore-realworld: 89 of 394 (22.6%)

This is the single highest-leverage fix in the bug-extractor breakdown. One-place fix in either internal/extractors/markdown/markdown.go (heading entity QualifiedName must match CONTAINS ToID exactly) or internal/resolve (doc::slug shape should look up the heading by file+slug).

Acceptance: aggregate bug-rate drops by >=15pp on go-gin and >=10pp on csharp-aspnetcore-realworld vs. the issue-91 baseline.

Refs: #91 (reports/issue-91-per-language-bug-extractors.md)

cajasmota:
WIP — implementer subagent dispatched. This is the highest-leverage bug-extractor fix per #91 baselines (45% of go-gin bugs).

cajasmota:
## Subagent investigation — issue #100

**Root cause:** `resolve.Index` had no QualifiedName index; markdown `Document --CONTAINS--> "<file>::<heading-slug>"` ToIDs went through `splitStub` which split on the first `:`, producing a non-existent kind segment, so every README heading landed in `bug-extractor`.

**Files touched:**
- `internal/resolve/refs.go` — add `byQualifiedName` field to `Index`, populate in `BuildIndex`, probe first in `Lookup` and `LookupStatusHint` (collisions blank to ambiguous).
- `internal/resolve/refs_test.go` — 2 new tests.

**Test count delta:** +2 (`TestReferences_QualifiedNameMatch`, `TestReferences_QualifiedNameAmbiguous`). Full suite green, `go vet` clean, `gofmt` clean for touched files.

**Commit / PR:** b0967e7 — https://github.com/cajasmota/archigraph/pull/110

**Measured (single-repo verify2, baseline = origin/main tip 8ece547):**
[...]

PR #6646, opened 2026-08-25
fix(#6636): the two "single tracking issue" crons searched only open issues, so a close forked the thread

### LINKED ISSUE #6636: grammar freshness compares against a fixed vendored snapshot date, so 23 of 27 grammars read as permanently stale
The grammar-freshness check compares every grammar against a **fixed snapshot date**, so it reports "stale" permanently and unconditionally. The signal is saturated and carries no information about any particular release.

`tools/grammar-freshness/check.go:70-75` compares the live upstream commit date against `lock.Binding.PinnedDate` — the vendored snapshot date, now two years old. Essentially every grammar is upstream of it, so essentially every grammar is "stale", every month, forever.

The workflow itself is fine and genuinely fires: `.github/workflows/grammar-freshness.yml`, `cron: '0 6 1 * *'`, real scheduled runs on the 1st of July and the 1st of August, both success, no dispatch-only trigger and no fork guard. **The cron is not the problem.** The comparison it performs is.

## Why a permanently-true alarm is worse than no alarm

It exits 0 regardless — the entire verdict is issue prose, so nothing can fail on it. And because it fires identically whether one grammar drifted yesterday or all 28 drifted two years ago, a reader cannot use it to answer the only question it is asked at release time: *did the grammar for the language that just shipped fall behind?*

`docs/language-release-calendar.md` names this check as one of two mechanisms that would catch a language-release regression (see #5673). It cannot, in its current form.

## A second, separable defect in the same workflow

The idempotency logic searches only `--state open` when deciding whether to reuse its tracking issue. So once someone closes the issue, the next run opens a **brand-new** one. The evidence is already in the repo's history: #5674 and #6066 are the same recurring report filed twice, each closed in turn. The "single idempotent tracking issue" property holds only while nobody ever closes it — which is the opposite of how a tracking issue is meant to be used.

## What is needed

1. **Make staleness relative to something that moves.** Compare against the currently-pinned grammar revision per language, not one global snapshot date, so "stale" means "this grammar drifted since we pinned it" rather than "time has passed".
2. **Report per-language drift, with a magnitude** — a commit count or a date delta — so a release-window reader can see which grammar moved and by how much.
3. **Fix the idempotency search** to consider closed issues too, or key on a stable marker rather than issue state.

## Rules

- **Do not make this workflow fail the build.** An advisory that opens an issue is the right shape for drift; the defect is that its verdict is uninformative, not that it is advisory. Making a permanently-true condition fail CI would be strictly worse.
- **Score the fix**: construct a lockfile state where exactly one language has drifted and confirm the report names that one and not the other 27. Today every input produces the same output — reproduce that as a survivor first, because a fix that still reports everything as stale will pass any test that only checks "the report is non-empty".
- The three items above are separable; take them one at a time rather than as one change.

Found by auditing #5673 rather than acting on its checklist. Refs #5673, #5359, #6066, #5674.

cajasmota:
Refs #6636 — **item 3 only**. Items 1 and 2 of that issue body are withdrawn by the author's own correction comment (item 1 is unimplementable: there is no per-language grammar provenance; item 2 already exists). The staleness comparison in `tools/grammar-freshness/check.go` is untouched.

## The defect

Two scheduled workflows document that they maintain "a **SINGLE** tracking issue", and both looked it up like this:

```bash
existing=$(gh issue list --state open --label "$ISSUE_LABEL" ...)
if [ -z "$existing" ]; then
  existing=$(gh issue list --state open --search "in:title \"$ISSUE_TITLE\"" ...)
```

`--state open` means a **closed** issue is invisible. The moment a human closes the tracking issue — the normal way a tracking issue gets used — the next monthly cron sees nothing and opens a brand-new one, with no reference to the old. The evidence is already in this repo's history: **#5674 and #6066 are the same recurring report, filed twice**, each closed in turn, with nothing linking them.

Affected, and (verified) only these two: `.github/workflows/grammar-freshness.yml` and `.github/workflows/language-release-calendar.yml`.

## Decisions

**Do not reopen; file a successor and link it.** The brief offered reopen / link / stable-marker. Reopening was rejected deliberately: a human closing the issue is a decision — "this instance is handled" — and silently reopening overrides that decision every month. But the failure mode that actually matters is the *silent* fork, not the fork. So the run files a successor and posts the link as a comment on it; GitHub renders the back-reference on the closed issue automatically, so the chain is traceable from both ends.

Note what this deliberately does **not** solve: if the maintainer's intent in closing was "stop telling me", this still tells them next month. That is the pre-existing contract of an advisory monthly cron, and is out of scope here — the fix makes the result a linked chain instead of an anonymous pile.

**Identity: label primary, marker verified locally, title not a key at all.** The repo-controlled **label** is the primary key — an exact server-side filter, and one `gh issue list --label --json number,state,createdAt,body` call returns the candidates' bodies in the same request. The **marker** (`<!-- grafel-tracker:… -->`) is then verified *locally*, and only when it occupies a column-0 line of its own. The **title** is no longer consulted, so a retitle cannot fork the thread either.

Precedence among candidates, documented in the script because it is load-bearing: `open+marker` → `open` → `closed+marker` → `closed`. State dominates the marker, so "the newest open candidate wins" holds unconditionally. Ties on `createdAt` break on issue number, so the pick is deterministic.

**Divergence.** The two workflows carried byte-identical copies of this logic, and copies rot independently — that is precisely how one of two paired checks silently goes wrong. Both now call one shared script, `.github/scripts/upsert-tracking-issue.sh`, parameterised by `ISSUE_TITLE` / `ISSUE_LABEL` / `ISSUE_MARKER` / `ISSUE_KIND` / `BODY_FILE`. Identity of behaviour is structural, not something a reviewer has to diff.

## Corrections from review (`52ed1554`)

The first cut of this PR claimed *"the workflow can never leave two open issues"*. **That claim was false**, and it is withdrawn rather than quietly patched. Review found three HIGH defects, each reaching the same destination as the original bug — a duplicate issue, or the wrong issue overwritten — through a different door:

- **A — a failed lookup created a duplicate.** `lookup()` ended `2>/dev/null || true`, making a non-zero exit indistinguishable from "no match". A rate limit or a 502 forked the thread. Every `gh` call's exit status is now checked, and any failure aborts *without writing anything*. A missed month is free; a duplicate is not.
- **B — garbage output produced a nonsense public issue.** Positional parsing of `not json{{{` gave `number=not`, and the closed branch then filed an issue reading *"Continues from #not, which was closed."* The number must now be all digits and the state exactly `open`/`closed`; anything else routes to A's abort path.
- **C — marker-first identity overwrote a human's issue.** An open contributor question that merely *quoted* the marker was selected and its body replaced with the grammar table. This was not hypothetical: the marker string appears in these YAML files and in this PR. Fixed by the label-primary / verify-locally restructure above (the reviewer's option (b)), which also removes the unverifiable dependence on GitHub's tokenised search. The search fallback — for the "label removed, marker survived" case — additionally requires the candidate to be **bot-authored**, which is what closes the hole for an issue carrying no label at all.
- **D (MEDIUM) — the forward pointer was erased a month later.** "Continues from #N" lived in the body, which the open-update path rewrites. It is now posted as a **comment**, which nothing rewrites, so a chain of three or more stays reconstructible.
- **E (LOW)** — deterministic tie-break on issue number, as above.

## Does GitHub index text inside HTML comments?

**I could not determine this offline, so I am not asserting it.** No issue body in this repo currently contains an HTML comment to probe with, and creating one to find out would mean writing to GitHub, which this work is not permitted to do.

The design was changed so the answer no longer matters on the path that runs every month: the primary key is the label (an exact filter, not full-text), and the marker is checked against the body text we fetch ourselves. The `in:body` search survives only as a fallback for the label-removed case; if the index does not reach inside HTML comments, that branch is simply inert and identity degrades to label-only — still correct. Its results are re-verified locally either way, so a loose match can never select a foreign issue.

## Verification

The scheduled workflows were **not** triggered, and **no issue was created, edited, reopened or closed on GitHub**. The script was exercised against a stubbed `gh` that filters a fixture issue list by `--state` / `--label` / `in:body`, can inject failures and malformed output, and logs every mutating call.

Baseline (pre-fix logic, one closed issue, nothing open) reproduces the reported defect: `CREATE, link=none, marker=none`.

| Case | Decision | Writes |
|---|---|---|
| no existing issue | create the first one | `CREATE` (marker embedded) |
| one OPEN (marker) | update in place | `EDIT #100`, `+label` |
| one CLOSED (marker) | successor + linked comment | `CREATE #9999`, `COMMENT … #5674` |
| CLOSED older + OPEN newer | update the open one | `EDIT #6066`, `+label` |
| OPEN older + CLOSED newer | update the **open** one | `EDIT #5674`, `+label` |
| retitled open issue | update in place — retitle does not fork | `EDIT #6066`, `+label` |
| marker stripped, label intact | update via label | `EDIT #6066`, `+label` |
| label removed, marker intact | update via verified marker | `EDIT #6066`, `+label` |
| unrelated issue (no marker, no label) | create the first one | `CREATE` |
| **A** `gh issue list` exits 1 while an open tracker exists | **abort**, exit 1 | **none** |
| **B** `gh` prints `not json{{{` | **abort**, exit 1 | **none** |
| **B** valid JSON, `state:"MYSTERY"` | **abort**, exit 1 | **none** |
| **C** contributor issue quotes marker + real tracker exists | picks the **real tracker** #100 | `EDIT #100`, `+label` |
| **C** only the contributor issue exists (quoted, indented) | no match — create our own | `CREATE` |
| **C** contributor pastes marker at column 0 in a code fence | no match — create our own | `CREATE` |
| **D** run 1 (only #5674, closed) → run 2 (#8000 now open) | successor, then update in place | `CREATE #8000` + `COMMENT`; then `EDIT #8000` |
| **E** two candidates, identical `createdAt` | picks #201 on all three runs | `EDIT #201` |

For **D**, the run-1 body written by the script contains `Continues from` **0 times** and the run-2 body **0 times** — the link is in the comment, which the monthly body rewrite cannot touch.

The same script was also driven under the `language-release-calendar` env (different title/label/marker/kind) and produced the corresponding successor-with-link, confirming the parameterisation.

Static checks: `actionlint` clean on both workflows (exit 0), `shellcheck -x` clean on the script (exit 0), both YAML files parse under Python `yaml.safe_load`. `yamllint` is not installed on this machine. No Go was built or run.

Not `Closes #6636` on purpose — the rest of that issue is rescoped (B2 decouple), not done.

### DIFF
--- a/.github/scripts/upsert-tracking-issue.sh
+++ b/.github/scripts/upsert-tracking-issue.sh
@@ -6,29 +6,45 @@
 # copy of this logic, and both copies searched only `--state open` — so once a
 # human closed the issue, the next cron run opened a brand-new one with no link
 # back. #5674 and #6066 are the same recurring report, filed twice, silently
-# forked. That silent fork is the failure mode this script exists to prevent.
+# forked. That silent fork is the failure mode this script exists to prevent,
+# and every rule below is written so that no other door leads to the same place.
 #
-# Policy (see the PR for #6636 for the reasoning):
-#   * Identity is keyed on a stable MARKER embedded in the issue body, with the
-#     label as an equally stable secondary key. Neither a retitle nor a label
-#     edit alone can fork the thread. The title is no longer consulted.
-#   * The search covers ALL states, not just open ones.
-#   * If the newest match is OPEN  -> update it in place (unchanged behaviour).
-#   * If the newest match is CLOSED -> do NOT reopen it. A human closing the
-#     issue is a decision ("this instance is handled"), and silently reopening
-#     it overrides that decision. Instead open a fresh issue whose body links
-#     the previous one, so the chain stays traceable. GitHub renders the
-#     back-reference on the closed issue automatically.
-#   * If there is no match at all -> create the first issue.
+# Identity
+#   The LABEL is the primary key: it is an exact, server-side filter, and one
+#   `gh issue list --label` call returns the candidates' bodies too. The MARKER
+#   (`<!-- grafel-tracker:... -->`) is then verified LOCALLY, and only when it
+#   occupies a line of its own — merely quoting the marker inside prose, a code
+#   fence or a review comment does not make an issue ours. The title is not a
+#   key at all, so a retitle cannot fork the thread.
+#
+# Precedence among candidates (documented because it is load-bearing):
+#   1. open + marker   2. open, no marker   3. closed + marker   4. closed
+#   State dominates the marker, so "the newest OPEN candidate wins" holds
+#   unconditionally and two open issues for one report are unreachable.
+#   Ties on createdAt are broken by issue number, so the choice is deterministic.
+#
+# Action
+#   * open match   -> update it in place.
+#   * closed match -> do NOT reopen. A human closing the issue is a decision
+#     ("this instance is handled"), and silently reopening overrides it. File a
+#     successor instead and post the "Continues from #N" link as a COMMENT, not
+#     in the body: the body is rewritten every month, comments are not, so a
+#     chain of three or more stays reconstructible.
+#   * no match     -> create the first issue.
+#
+# Failure handling
+#   A lookup that ERRORS must never be read as "no match" — that would create a
+#   duplicate on a rate limit or a 502, i.e. fork the thread through a different
+#   door. Every gh call's exit status is checked and any failure, or any
+#   malformed field, aborts without writing anything. A missed month is free;
+#   a duplicate is not.
 #
 # Required environment:
 #   ISSUE_TITLE   title used when creating a new issue
 #   ISSUE_LABEL   stable label, applied on create and re-applied on update
-#   ISSUE_MARKER  stable identity token appended to every body
+#   ISSUE_MARKER  stable identity token, written as its own HTML-comment line
 #   ISSUE_KIND    human word for log lines, e.g. "tracking" or "reminder"
 #   BODY_FILE     path to the rendered markdown body
-#
-# `gh` must be authenticated. Nothing here writes to a closed issue.
 set -euo pipefail
 
 : "${ISSUE_TITLE:?ISSUE_TITLE is required}"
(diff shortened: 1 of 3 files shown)


cajasmota:
Merging. All five findings closed, and the one that mattered most is verified by my own stub rather than the reported table.

## My own mutant — the column-0 rule is load-bearing

`marked` requires the marker to occupy a line of its own, with **trailing** whitespace trimmed and leading whitespace deliberately preserved. I mutated `trim` to strip leading whitespace too:

```diff
-def trim: sub("[[:space:]]+$";"");   # trailing only: our marker is written at column 0
+def trim: sub("[[:space:]]+$";"") | sub("^[[:space:]]+";"");
```

Fixture: the real tracker `#100` (open, older, marker at column 0) and a contributor's question `#7777` (open, newer, marker **indented** inside prose).

| | decision |
|---|---|
| pristine | `EDIT 100` — the real tracker |
| mutant | **`EDIT 7777`** — the contributor's issue, body overwritten and relabelled |

One character of whitespace handling separates "update our tracker" from "destroy a contributor's issue on a public repo". That is exactly the hazard the review found, and it is now pinned by construction rather than by intention.

## The fix went further than the finding

I asked for local marker verification. What landed **restructured the identity model**: the label is now the primary key, fetched in one call *with bodies*, so the marker is verified locally and the monthly path never depends on GitHub's tokenised search at all. The `in:body` fallback survives only for the label-removed case, and there a candidate must **also be bot-authored**.

That last guard closes a case neither of us had listed: a contributor pasting the marker **at column 0** inside a code fence still satisfied the line rule. The line check alone was not enough, and finding that required building the attack rather than reasoning about it.

## On the question that could not be answered

Whether GitHub indexes text inside HTML comments was left **unresolved and unasserted** — no issue body in the repo contains one to probe, and manufacturing one means writing to GitHub. Rather than assume, the dependency was removed: the monthly path never uses full-text search, and the fallback that does is inert-but-harmless if the index does not reach into comments, degrading to label-only identity.

That is the right handling of an unknown. The alternative — assuming it works and shipping a feature that silently does nothing — is this milestone's most-repeated defect.

## The claim was withdrawn, not quietly repaired

"The workflow can never leave two open issues for the same report" was false when written: a rate-limited `gh issue list` was indistinguishable from "no match" and would have created a duplicate. The body now marks that sentence as having been false and lists A–E as the corrections, instead of editing it into truth. A body that silently becomes correct teaches the next reader nothing.

## The rest

- **A** — every lookup now distinguishes exit status from empty output; a transient failure **aborts and writes nothing**. A missed month is free; a duplicate is not.
- **B** — number must be all digits, state exactly `open`/`closed`, validated before branching. No more "Continues from #not".
- **D** — the back-reference moved out of the body and into a **comment**, because bodies are rewritten monthly and comments are not. Verified over two runs: `grep -c 'Continues from'` on the written body is 0 both times, and the link survives.
- **E** — `sort_by(.createdAt, .number)`, deterministic on ties.

18 stubbed cases, the original nine unchanged. `actionlint` 0, `shellcheck -x` 0. Nothing was created, edited, reopened or closed on GitHub during any of it — the whole review and rework happened against stubs, which is the only responsible way to test something whose failure mode is public issue spam.

Issue #6637, opened 2026-08-25
a backslash-escaped newline continues a single-line literal, and nothing observes it — two fixtures pin two shapes, not a rule


cajasmota:
A backslash-escaped **newline** inside a single-line Python literal is a line continuation, and the masker's handling of it is unobserved. A mutant that honours escapes only when the escaped byte is the quote or a backslash passes the whole engine package:

```diff
-				if b[j] == '\\' {
+				if b[j] == '\\' && j+1 < len(b) && (b[j+1] == c || b[j+1] == '\\') {
 					j += 2
 					continue
 				}
```

`go vet` **0**, full package `go test -count=1 ./internal/engine/` **exit 0**.

## Non-equivalence witness — measured, not argued

Input (a real Python line continuation inside a string):

```python
x = "a\
b" and app.include_router(r, prefix=p)
```

| | masked output |
|---|---|
| pristine | unchanged — the whole continuation stays intact |
| mutant | `"x =    \nb                                     \n"` |

Under the mutant the literal is treated as unterminated at the newline, its tail is blanked, and then `b"` opens a *new* literal that blanks the rest of the second line — **losing the `include_router` mount entirely.** That is the over-masking direction: a false negative, which is the harder failure to notice, because a missing endpoint looks like a repo that simply does not have one.

## Why it is filed rather than folded into #6624

PR #6632 pins two escape shapes — escaped **quote** (`x = "\""`) and escaped **backslash** (`x = "\\"`). Its own enumeration measured this third shape and recorded it: 11,110 inputs in the ≤7 space kill all three of its distinct mutants, and every one of them escapes a newline. The agent named it as out of scope rather than quietly leaving it, which is why this issue exists at all.

My mutant is precisely the fixture-shaped one: it handles exactly the two shapes #6632 pins and nothing else. It survives because only the newline shape distinguishes it. That makes this a clean test of whether a pin generalises — and the honest answer is that two fixtures pin two shapes, not a rule.

## What is needed

1. A test pinning that a backslash-escaped newline continues the literal — the mutant above must die.
2. **Score it before and after** — exit 0 on pristine `main` (survivor reproduced), exit 1 after.
3. **Premise-guard so the branch actually executes**, not merely that the fixture contains `\` and a newline. This class has failed five times in this function.
[...]

PR #6647, opened 2026-08-25
test(#6629): a backslash inside a `#` comment ended the comment, and 0 tests saw it

### LINKED ISSUE #6629: a backslash inside a '#' comment can end the comment and nothing observes it
A third disjunct added to `pythonMaskInertRegions`'s `#`-comment loop survives the whole engine package:

```diff
-			for ; i < len(b) && b[i] != '\n'; i++ {
+			for ; i < len(b) && b[i] != '\n' && b[i] != '\\'; i++ {
```

`go vet` **0**, full package `go test -count=1 ./internal/engine/` **exit 0**.

**Non-equivalence witness** (so this is a real hole, not an unreachable branch): `# a \ prefix="/ghost"` masks to all spaces pristine, and to `"    \ prefix=\"/ghost\""` under the mutant — the tail of the comment stays live and its `prefix=` reaches the mount scan.

**Pre-existing.** Found by the adversarial review of PR #6625, which was fixing the quote-escape hole in the same loop. Filed rather than folded in, because #6625 is a test-only PR for a different escape character and widening it would have made the two indistinguishable in review.

## What is needed

1. A test pinning that a backslash inside a `#` comment does not end the comment — the mutant above must die.
2. **Score it before and after** — exit 0 on pristine `main` (survivor reproduced), exit 1 after. A kill not first reproduced as a survivor is not evidence.
3. A premise guard asserting **the comment branch actually runs**, not merely that the fixture contains a `#` and a `\`. This class has failed four times in this function (#6611, #6615, #6620, and the `'`-half gap the #6625 review found).
4. **Enumerate rather than hand-pick**, and report the per-clause drop analysis. But note the specific lesson from #6625: its sweep of 11,111,111 inputs was reproduced digit-for-digit and *still* missed this family, because **it varied the input against a fixed mutant**. Enumeration answers "does the guard admit a defeating input"; it says nothing about "does the fixture pin the mutant family." So vary **both** here — at minimum score the `'`-only, `"`-only, `\`-only and combined disjuncts separately.

## Traps

2. The over-masking direction is already covered: dropping `b[i] != '\n'` so the comment eats to EOF is **DEAD**, killed by `TestSynth_FastAPI_MountedRouter_AbsentWhenUnresolvable_6414`. Do not re-pin it.
3. `:3192`'s single-line clamp is **equivalent under the current suite**, verified repeatedly with the reason. Never re-score it as DEAD.
4. The mixed-quote terminator (`b[j] == c` → either quote) is likewise **equivalent**, and the reason is worth keeping: the scan loop breaks only on the opening quote character or `\n`, so the other quote is unreachable at that check.
5. This is the **string** branch's neighbour but not the same as #6624, which is the `j += 2` escape skip inside single-line literals and still needs a non-equivalence probe of its own. Keep them apart.

## Where this sits

Seventh item in and around one 60-line function: #6418 (known limitation), #6598 (fixed), #6612/#6617 (four bounds pinned), #6614 (fixed, #6622), #6623 (in review, #6625), #6624 (probe first), and this. The uniform cause has not changed: the function walks a byte slice with several look-ahead offsets and branch dispatches, and every fixture in the suite is well-formed Python, so no branch boundary is ever approached.

Refs #6623, #6625, #6617, #6418.

cajasmota:
`pythonMaskInertRegions`'s `#`-comment branch blanks the rest of the line, and nothing may end the comment except the newline. #6623 pinned that for a **quote** inside the comment. The **backslash** half was a live pre-existing survivor on `32d9e1162`:

```diff
 		case c == '#':
-			for ; i < len(b) && b[i] != '\n'; i++ {
+			for ; i < len(b) && b[i] != '\n' && b[i] != '\\'; i++ {
```

`go vet` 0, full `go test -count=1 ./internal/engine/` **exit 0 — reproduced as a survivor before anything was written.**

Under that mutant the comment loop halts **on** the backslash, which the switch dispatches to its *default* arm (a `\` is neither `#` nor a quote), so the cursor steps past it and the **rest of the comment is scanned as live Python**. A commented-out `prefix="/ghost"` sitting after the backslash is then skipped intact by the single-line-literal branch and reaches the mount scan. Witness from the issue: `# a \ prefix="/ghost"` masks to all spaces pristine, and to `    \ prefix=\"/ghost\"` under the mutant.

**Test-only.** `http_endpoint_synthesis.go` is byte-identical to `main` (`shasum 70a007154803093f6b8e7ef719994f72801e52c0`, `git diff` empty).

## Four mutant families, scored separately

Varying the *input* against a *fixed* mutant answers "does the guard admit a defeating input", not "does the fixture pin the family" — the lesson #6625 paid for. So the mutant was varied too. Full package, no `-run` filter, `go vet` first on each.

| mutant | before | after |
|---|---|---|
| `&& b[i] != '\\'` | **SURVIVOR** | **DEAD** |
| `&& b[i] != '\''` | DEAD | DEAD |
| `&& b[i] != '"'` | DEAD | DEAD |
| combined three-disjunct | DEAD | DEAD |

Only the `\\` family was open. The quote families die on #6623's fixture, and the **combined form dies on its quote clauses alone** — it was never a distinguishing family, which is worth recording since the issue asked for it to be scored as one.

## What the fixture asserts

Fixture: `` `# app.include_router(legacy, \ prefix="/ghost", tags=['ghost'])` `` — the backslash sits *before* the prefix, so the mutant's damage is concrete.

Claim: every byte of `[h, eol)` in the masked copy is a space, plus length invariance.

Premise guards, so the **comment branch actually executes** rather than the fixture merely containing a `#` and a `\`:

- **C1** — a `#` exists.
- **C2** — no quote before it, or the string branch consumes the `#` and the comment branch never runs (the trap that caught #6611, #6615, #6620, and the `'`-half of #6623).
- **C3** — a backslash inside the comment, so the mutant has a byte to stop on.
- **C4** — no `\r` in the comment, since `blank()` preserves CR.

The kill is argued by construction for input of arbitrary length (written out in the file), not by a bounded sweep: the loop halts with the cursor **on** the backslash, nothing afterwards writes to that offset, so `got[bs]` is still `\` and the assertion fails.

## Per-clause drop analysis (measured, isolated runs)

Each guard neutralised in turn, the fixture degraded to the shape that guard forbids, every mutant re-applied.

| dropped | degraded fixture | pristine | verdict |
|---|---|---|---|
| C1 | `'''x \ prefix="/ghost"'''` | PASS | all four **SURVIVE** — load-bearing against all |
| C2 | `"""# \ prefix='/ghost'"""` | PASS | all four **SURVIVE** — load-bearing against all |
| C3 | same comment, no `\` | PASS | `\\` **SURVIVES**, quotes DEAD — load-bearing **against the `\\` mutant only** |
| C4 | a `\r` in the comment | **FAIL** | no survivor — **decorative** as a mutant guard |
| — (control) | this fixture | PASS | all four DEAD |

C3 is the #6632 point in miniature: with it dropped the test still kills three of the four mutants, so a verdict that does not name the family it is scored against would report the clause as unnecessary.

C4 is stated plainly as decorative: it admits no mutant survivor and exists only so a later fixture edit cannot break the assertion on correct code.

## Deliberately not touched

- No widening into a Python tokenizer — #6418 records that as the real fix for this family and as out of scope.
- The over-masking direction (dropping `b[i] != '\n'`) is already DEAD under `TestSynth_FastAPI_MountedRouter_AbsentWhenUnresolvable_6414`; not re-pinned.
- `:3192`'s single-line clamp and the mixed-quote terminator remain **equivalent**; not re-scored.
- The `j += 2` escape skip (#6624) and the escaped-newline shape (#6637) are separate and untouched.

Closes #6629

### DIFF
--- a/internal/engine/http_endpoint_synthesis_python_mask_6629_test.go
+++ b/internal/engine/http_endpoint_synthesis_python_mask_6629_test.go
@@ -1,0 +1,164 @@
+package engine
+
+import (
+	"strings"
+	"testing"
+)
+
+// pythonMaskInertRegions' `#`-comment branch blanks the rest of the line
+// unconditionally, and NOTHING may end the comment early except the newline.
+// #6623 pinned that for a QUOTE inside the comment. The backslash half was
+// left open and is a real, pre-existing hole: adding a third disjunct
+//
+//	for ; i < len(b) && b[i] != '\n' && b[i] != '\\'; i++ {
+//
+// left `go vet` at 0 and the full engine package at exit 0 on 32d9e1162.
+//
+// The effect is the same class of damage as #6623, by a different route.
+// Under that mutant the comment loop halts ON the backslash, which the switch
+// dispatches to its DEFAULT arm (a backslash is neither `#` nor a quote), so
+// the cursor simply steps past it and the REST OF THE COMMENT is scanned as
+// live Python. A commented-out `prefix="/ghost"` that sits after the backslash
+// is then read by the single-line-literal branch, which SKIPS rather than
+// blanks, and the ghost prefix reaches the mount scan. The witness from the
+// issue: `# a \ prefix="/ghost"` masks to all spaces pristine, and to
+// `    \ prefix=\"/ghost\"` under the mutant.
+//
+// Why the kill holds for input of ARBITRARY length, by construction rather
+// than by a bounded search:
+//
+//	C1+C2 put the scanner at the `#`. `h` is the FIRST `#` and every byte
+//	before it is neither `#` nor a quote, so each takes the switch's default
+//	arm and the cursor arrives at h with the COMMENT branch selected.
+//	C3 gives the mutant a byte to stop on: the first `\` inside the comment.
+//	The loop blanks [h,bs) and halts with the cursor ON bs, and nothing
+//	afterwards writes to bs — the cursor only ever advances, and the byte is
+//	dispatched to the default arm, which writes nothing. got[bs] is therefore
+//	still `\`, h <= bs < eol, and the all-spaces assertion below fails. Dead.
+//	Symmetrically on pristine: the comment loop blanks every byte of [h,eol),
+//	blank() rewrites everything except `\n` and `\r`, the region holds no `\n`
+//	by construction and no `\r` by C4, so it comes out all spaces and the
+//	assertion passes. Neither argument mentions the length of src.
+//
+// Four mutant families were scored separately, because a sweep that varies the
+// INPUT against a FIXED mutant answers "does the guard admit a defeating
+// input" and says nothing about "does the fixture pin the family" — the lesson
+// #6625 paid for. Before this file: `\\` SURVIVOR; `'`, `"` and the combined
+// three-disjunct form all DEAD, killed by #6623's fixture. After: all four
+// DEAD. The combined form was therefore never a distinguishing family — it
+// dies on its quote clauses alone — and `\\` alone is what this file adds.
+//
+// Per-clause drop analysis, measured (each guard neutralised in turn, the
+// fixture degraded to the shape that guard forbids, then every mutant
+// re-applied and this test run in isolation):
+//
+//	C1 dropped, fixture `'''x \ prefix="/ghost"'''`      pristine PASS, all four SURVIVE
+//	C2 dropped, fixture `"""# \ prefix='/ghost'"""`      pristine PASS, all four SURVIVE
+//	C3 dropped, fixture without the `\`                  pristine PASS, `\\` SURVIVES, quotes DEAD
+//	C4 dropped, fixture with a `\r` in the comment       pristine FAIL, no survivor
+//	control, all guards, this fixture                    pristine PASS, all four DEAD
+//
+// So C1 and C2 are load-bearing against every mutant; C3 is load-bearing
+// against the `\\` mutant ONLY — the quote mutants die on this fixture with C3
+// gone, which is precisely why a verdict has to name the family it is scored
+// against (#6632); and C4 is DECORATIVE as a mutant guard, load-bearing only
+// for the assertion's correctness on pristine code.
+//
+// Out of scope and separately tracked: #6418 (a terminated literal OUTSIDE a
+// comment still mints a mount — the known limitation whose real fix is a
+// Python tokenizer), #6624 (the `j += 2` escape skip inside a single-line
+// literal), #6637 (the escaped-newline shape, where the backslash is the last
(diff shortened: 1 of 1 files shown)


cajasmota:
Merging. Every claim in the body reproduced under independent measurement, and the construction argument holds up to a line-by-line audit.

## My own scoring

The `\\` mutant on this head: `go vet` 0, full package `go test -count=1 ./internal/engine/` **exit 1 — DEAD**, sole failure `TestPythonMaskInertRegions_BackslashInsideCommentIsBlanked_6629`. Source restored and re-verified at `70a00715…`.

I also ran a mutant that **survived** — stopping the comment loop at `\r` as well as `\n`. That is not a gap in this PR: it is a genuine question about lone-`\r` line termination, where the *current code* is arguably the wrong one, and it is filed as **#6648**.

## The construction argument earns its place

The kill is argued for arbitrary-length input rather than by a bounded sweep, and the audit traced all three steps against the source rather than the prose: `h` is the first `#` and C2 forbids a quote before it, so the cursor lands on `h` exactly and the switch has no fourth arm to escape through; `[h,eol)` holds no `\n`, so the mutant halts on the first `\`; and `i` is strictly non-decreasing on every path, with every blanking loop writing only to `[i,j)`, so nothing rewrites that offset afterwards — including the `j += 2` overrun, which is clamped before `i = j`.

That is strictly stronger than #6625's eleven-million-input sweep, and worth saying plainly: **the sweep is what missed this family in the first place.** A proof about the cursor beats an enumeration over inputs.

## The per-clause verdicts are exact

All five rows reproduced. C3 load-bearing against the `\\` mutant **only** — drop it and the three quote mutants still die, so an unqualified verdict would have called it unnecessary. C4 admits no mutant and its removal breaks the pristine assertion, which is exactly the "decorative as a mutant guard, load-bearing for correctness" shape it was recorded as.

This is #6632's corollary landing for the third time: *load-bearing against what* is part of the claim, not a footnote.

## Two survivors found, neither this PR's

Both in `blank()`, the helper C4's justification cites:

- `if b[i] != '\n' && b[i] != '\r'` → dropping the `\r` clause destroys a CR.
- Dropping the `\n` clause turns **a newline inside a triple-quoted docstring into a space**, silently merging two source lines in the masked copy.

The second is the more serious — any downstream per-line scan sees a collapsed line — and it is distinct from #6648, which is about the comment loop's *terminator* rather than the blanker's *write predicate*. Filing it separately.

Also recorded so nobody re-runs them: skipping the `#` itself, and the `i < len(b)-1` bound, are both DEAD — killed by this test **and** by #6623's.

Issue #6648, opened 2026-08-25
a lone carriage return ends a Python line but not our '#' comment, so we blank live code — a false negative


cajasmota:
`pythonMaskInertRegions`'s `#`-comment branch runs to `\n`, treating a lone `\r` as ordinary comment content. **Python does not** — a bare carriage return is a line terminator under universal newlines, so the bytes after it are live code on the next line. We blank them.

This is a **false negative**: a real mount is masked away and never reaches the scan.

## Measured, both directions

Mutant: `for ; i < len(b) && b[i] != '\n' && b[i] != '\r'; i++` — `go vet` **0**, full package `go test -count=1 ./internal/engine/` **exit 0 — SURVIVOR** (against `main` plus #6629's new fixture).

| input | pristine | mutant |
|---|---|---|
| `# a\rapp.include_router(r, prefix=p)\n` | `"   \r                               \n"` | `"   \rapp.include_router(r, prefix=p)\n"` |
| `# a\r\nx = 1\n` (well-formed CRLF) | identical | identical |

So CRLF files — the common Windows case — are unaffected either way, because the `\r` is immediately followed by `\n` and everything before it was already blanked. The divergence needs a **lone** `\r` mid-line.

## Why this is a behaviour question, not just a coverage gap

The mutant is **closer to Python's semantics than the current code is.** Under universal newlines a lone `\r` ends the line, so `app.include_router(...)` above is genuinely live code that we are silently discarding.

That makes this different from the rest of the cluster, where the pass fails safe by minting nothing. Here it fails *unsafe* in the reporting sense — a mount that exists is not reported, and a missing endpoint is invisible in a way a phantom endpoint is not.

I am not asserting the fix is simply "stop at `\r` too". That changes masking behaviour for every branch that shares the assumption, and the decision belongs with someone who has looked at all of them:

- does the triple-quoted branch treat a lone `\r` as content? (It should — inside a string literal it *is* content.)
- does the single-line literal branch? Its unterminated-tail blanking (#6614) has the same question.
- does `blank()` preserving `\r` interact with a per-line reading of the buffer?

## Where it was found, and the honest limitation of the fixture that missed it

[...]

PR #6650, opened 2026-08-25
test(#6649): blank() preserved \n and \r for line structure, and 0 tests saw either clause

### LINKED ISSUE #6649: blank() preserves \n and \r to keep line structure, and neither clause is observed — dropping \n merges docstring lines
`blank()` in `pythonMaskInertRegions` preserves `\n` and `\r` so the masked copy keeps the source's line structure. **Neither clause of that predicate is observed.** Both mutants pass the whole engine package:

```diff
-		if b[i] != '\n' && b[i] != '\r' {
+		if b[i] != '\n' {            // "blankcr"  — SURVIVOR
+		if b[i] != '\r' {            // "blanknl"  — SURVIVOR
```

`go vet` **0**, full package `go test -count=1 ./internal/engine/` **exit 0** for each.

## Non-equivalence witnesses

| mutant | input | pristine | mutant |
|---|---|---|---|
| blankcr | `"# a\rb"` | `"   \r "` | `"     "` — the CR is destroyed |
| **blanknl** | `"\"\"\"doc\nline2\"\"\"\napp..."` | `"      \n        \napp..."` | `"               \napp..."` |

## blanknl is the serious one

It turns **a newline inside a triple-quoted docstring into a space**, silently merging two source lines in the masked copy. The function's contract — stated in its own doc comment — is that the result has *the same byte length and the same newline positions* as the source, so byte offsets in the masked copy address exactly the same source positions. Under this mutant that contract is broken for any multi-line docstring, which is to say: for most real Python files.

Anything downstream that reads the masked copy per line, or maps an offset back to a line number, silently sees a collapsed line. That is the kind of wrongness this project ranks above loud breakage, and nothing observes it.

## Why it was not caught by the recent work

The `pythonMaskInertRegions` cluster now runs to nine items and every one of them has been about the **scanner** — which branch is entered, how far it runs, where it stops. `blank()` is the **writer**, and no fixture has ever asserted anything about it beyond "the span became spaces". #6629's C4 guard is the only place it is even mentioned, and only to explain why a fixture may not contain a `\r`.

So this is a genuinely different axis, not a tenth variation on the same one.

## What is needed

1. Pin both clauses — each mutant above must die.
2. **Score before and after**: exit 0 on pristine `main` (survivor reproduced), exit 1 after. A kill not first reproduced as a survivor is not evidence.
3. Assert the **contract**, not a fixture's spelling: same byte length, and the set of newline offsets in the output equals that of the input. That statement kills both mutants at once and states the property the doc comment already claims.
4. Include a **multi-line triple-quoted docstring**, since that is the shape that makes blanknl produce a real defect rather than a curiosity.
5. Premise-guard so the blanking path actually runs on the region under test.
6. **Vary the mutant family**, not just the input — score each clause separately as well as the combined drop, and report which fixture pins which.

## Traps

1. `blank()` is called from several branches. A fixture that only exercises the comment branch pins one caller, not the helper — say which callers your fixtures cover.
2. Do **not** widen into a Python tokenizer (#6418 records that as out of scope for this whole family).
3. Distinct from **#6648**, which is about the comment loop's *terminator* treating a lone `\r` as content. This issue is the blanker's *write predicate*. They touch the same byte and are not the same bug; do not merge them.
4. Test-only. `internal/engine/http_endpoint_synthesis.go` must end byte-identical to `main`.

Found by the adversarial review of PR #6647. Refs #6629, #6648, #6418.

cajasmota:
Closes #6649

`pythonMaskInertRegions`' doc comment promises a masked copy of the **same byte length** with the **same newline positions**, so a byte offset in the masked copy addresses exactly the same source position as in the original. The only thing delivering the second half of that promise is one predicate in the writer, `blank()`:

```go
if b[i] != '\n' && b[i] != '\r' {
	b[i] = ' '
}
```

Neither clause was observed on `d7eec1d33`.

## Three mutant families, scored separately

`go vet` **0** throughout; full package `go test -count=1 ./internal/engine/`, no `-run` filter.

| mutant | predicate | before | after |
|---|---|---|---|
| `blankcr` | `if b[i] != '\n' {` | exit 0 — **SURVIVOR** | exit 1 — **DEAD** |
| `blanknl` | `if b[i] != '\r' {` | exit 0 — **SURVIVOR** | exit 1 — **DEAD** |
| `both` | `if true {` | exit 0 — **SURVIVOR** | exit 1 — **DEAD** |

`blanknl` is the damaging one: inside a multi-line triple-quoted docstring it rewrites a newline as a space, silently **merging two source lines** in the masked copy while every later offset still lines up byte-for-byte. Anything downstream reading the masked copy per line, or mapping an offset back to a line number, then reads a collapsed line.

## What the test asserts — the contract, not a spelling

```
len(got) == len(src)                                   AND
ordered offsets of '\n' in got == those in src         AND
ordered offsets of '\r' in got == those in src
```

One property statement, exactly what the doc comment already claims, and it kills both clauses at once. It also **sidesteps** the trap that forced #6629's C4: an all-spaces assertion cannot be used on a fixture containing `\r`, precisely because `blank()` preserves it. Asserting preserved-byte *positions* needs no such guard — a `\r` in the region is the point, not a hazard.

Each subtest is premise-guarded: a marker substring inside the region must come out all spaces, proving the blanking path actually ran rather than the fixture being copied through untouched.

## Caller coverage

`blank()` has three call sites and they are **not** equally powerful — two of them can never be handed a `'\n'`:

| caller | site | can see `\n`? | fixtures |
|---|---|---|---|
| **A** | `#`-comment loop (`b[i] != '\n'` halts on it) | no | `commentCRLF`, `commentLoneCR` |
| **B** | triple-quoted span — the only span crossing line boundaries | **yes** | `docstringMultiline` (#6649's own witness), `docstringCRLF` |
| **C** | unterminated single-line literal (#6614) | no | `unterminatedCR` |

So `blanknl` is killable **only** through caller B.

## Per-fixture drop analysis (measured failure matrix; X = subtest failed)

| fixture | caller | blankcr | blanknl | both |
|---|---|---|---|---|
| commentCRLF | A | X | | X |
| commentLoneCR | A | X | | X |
| docstringMultiline | B | | X | X |
| docstringCRLF | B | X | X | X |
| unterminatedCR | C | X | | X |

Subtests are independent, so each single-fixture drop is exact. Group drops were measured directly:

- drop **both caller-B** fixtures → `blanknl` **SURVIVES** (measured exit 0); `blankcr` and `both` still die on A/C.
- drop **callers A and C** → no survivor (measured exit 1 for all three); caller B alone kills the whole table.

## Two negative results, recorded in-tree

1. **No single fixture is load-bearing against any mutant.** Only the caller-B *pair* is, and only against `blanknl`; the two docstring fixtures are mutually redundant as kills. Both are kept: `docstringMultiline` is #6649's witness, and `docstringCRLF` is the one `blanknl`/`blankcr` kill that #6648 cannot disturb.
2. **The caller-A and caller-C fixtures kill nothing caller B does not already kill.** They are load-bearing for *caller coverage* only — they are what makes this a statement about `blank()` rather than about the triple-quote branch. Optimising for the mutant table alone would delete them and leave two of three call sites unobserved.

## Scope

Test-only. `internal/engine/http_endpoint_synthesis.go` is **byte-identical to `main`** (`shasum 70a0071548…` on both).

Deliberately **not** touched: #6648 (the comment loop's *terminator* treating a lone `\r` as content — the scanner's stop condition, not the writer's predicate; caller B's fixtures are immune to it either way), #6418 (a real Python tokenizer, the recorded eventual fix for this family), and the recorded equivalences at the single-line clamp and the mixed-quote terminator.

### DIFF
--- a/internal/engine/http_endpoint_synthesis_python_mask_6649_test.go
+++ b/internal/engine/http_endpoint_synthesis_python_mask_6649_test.go
@@ -1,0 +1,289 @@
+package engine
+
+import (
+	"strings"
+	"testing"
+)
+
+// pythonMaskInertRegions' doc comment promises a copy of src of the SAME BYTE
+// LENGTH, with THE SAME NEWLINE POSITIONS, so that a byte offset in the masked
+// copy addresses exactly the same source position as in the original. The only
+// thing that delivers the second half of that promise is one predicate inside
+// the writer, blank():
+//
+//	if b[i] != '\n' && b[i] != '\r' {
+//		b[i] = ' '
+//	}
+//
+// Neither clause was observed on d7eec1d33. Dropping either one left `go vet`
+// at 0 and the full engine package at exit 0:
+//
+//	if b[i] != '\n' {   // "blankcr" — SURVIVOR, the CR is destroyed
+//	if b[i] != '\r' {   // "blanknl" — SURVIVOR, the LF is destroyed
+//	if true {           // "both"    — SURVIVOR, line structure gone entirely
+//
+// blanknl is the damaging one. Inside a multi-line triple-quoted docstring it
+// rewrites a newline as a space, so two source lines MERGE in the masked copy
+// while every later offset still lines up byte-for-byte. Anything downstream
+// that reads the masked copy per line, or maps an offset back to a line
+// number, then silently reads a collapsed line — wrong quietly, which this
+// project ranks above breaking loudly. The witness from #6649:
+//
+//	src       "\"\"\"doc\nline2\"\"\"\napp..."
+//	pristine  "      \n        \napp..."
+//	mutant    "               \napp..."
+//
+// Why the whole nine-item pythonMaskInertRegions cluster missed it: every
+// earlier item was about the SCANNER — which branch is entered, how far it
+// runs, where it stops. blank() is the WRITER, and no fixture had ever
+// asserted anything about it beyond "the span became spaces", an assertion
+// that cannot see a preserved byte at all. #6629's C4 is the only place it is
+// even mentioned, and only to explain why a fixture may NOT contain a `\r` —
+// which is exactly the shape needed to observe the CR clause. So the guard
+// that kept the old assertions honest is what kept this hole open, and the
+// fix is to stop asserting the SPELLING of the masked copy and assert the
+// CONTRACT instead.
+//
+// The contract, stated once, kills both clauses at once:
+//
+//	len(got) == len(src), AND
+//	the ORDERED offsets of '\n' in got == those in src, AND
+//	the ORDERED offsets of '\r' in got == those in src.
+//
+// This also sidesteps the trap that forced #6629's C4: an all-spaces
+// assertion cannot be used naively on a fixture containing `\r`, precisely
+// because blank() preserves it. Asserting the preserved-byte POSITIONS needs
+// no such guard — a `\r` in the region is the point, not a hazard.
+//
+// blank() has THREE call sites, and a fixture that exercises one pins one
+// caller, not the helper. All three are covered here, and they do not have
+// equal power, because two of them can never see a '\n' at all:
+//
+//	CALLER A — the `#`-comment loop, `for ; i < len(b) && b[i] != '\n'; i++`.
+//	  Halts ON the newline, so blank() is never handed a '\n' from here. It
+//	  CAN be handed a '\r' (the CR of a CRLF, or a lone CR mid-comment).
+//	  Pins blankcr and both; CANNOT pin blanknl.
+//	CALLER B — the triple-quote span, `for k := i; k < j; k++`. The only site
+//	  whose span crosses line boundaries, so the only one that can hand
+//	  blank() a '\n'. Pins blankcr, blanknl and both.
+//	CALLER C — the UNTERMINATED single-line literal span (#6614), same loop
+//	  shape but j stops at the quote or the '\n'. Like caller A it never sees
+//	  a '\n', and like caller A it can see a '\r'.
+//	  Pins blankcr and both; CANNOT pin blanknl.
+//
+// So blanknl is killable ONLY through caller B, and the two docstring
+// fixtures below are what do it. That asymmetry is the reason the fixtures are
+// labelled by caller rather than by shape.
+//
+// Three mutant FAMILIES were scored separately, not one family against varied
+// input — the distinction #6625 paid for. Measured on d7eec1d33, full package
+// `go test -count=1 ./internal/engine/`, `go vet` 0 throughout:
+//
+//	                  before this file    after
+//	blankcr           exit 0 (SURVIVOR)   exit 1 DEAD
+//	blanknl           exit 0 (SURVIVOR)   exit 1 DEAD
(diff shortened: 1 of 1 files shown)


cajasmota:
Merging. Both negative results reproduced cell-for-cell by an independent reviewer, and the honest one turned out to be the important one.

## My own scoring

`blanknl` (`if b[i] != '\r'`, dropping the `\n` clause) on this head: `go vet` 0, full package **exit 1 — DEAD**, failing `…PreservesLineStructure_6649/docstringMultiline` and `/docstringCRLF`. Source restored and verified at `70a00715…`.

I ran a second mutant earlier — `blank()` also preserving `"` and `'` — which is **DEAD, but killed by three pre-existing tests** (#6614's, #6629's, #6623's), not by this PR's. The reviewer found the same pattern independently: the cluster's *older* tests cover the writer's **filler and range**, incidentally and unplanned, while this PR covers its **predicate**. Nobody designed that division; it is worth knowing it exists.

## The negative results were right, and the caller-A/C judgement was the right call

Both reproduced by measurement: no single fixture is load-bearing against any mutant, and only the caller-B *pair* is, against `blanknl`. Dropping callers A and C leaves every mutant dead — caller B alone kills the whole table.

Keeping them anyway is correct, and the reviewer put the reason better than the PR did: the mutant table is a table over **one predicate**, and callers A and C are the two sites that can never hand `blank()` a `\n`, so by construction they can only duplicate caller B's `\r` kill. Deleting them for redundancy would optimise the suite against the three mutants that happen to have been written, and silently convert a claim about `blank()` into a claim about the triple-quote branch. That is precisely how this hole stayed open through nine cluster items.

One over-claim to note rather than fix: the body calls `docstringCRLF` "the one kill that #6648 cannot disturb". **Both** caller-B fixtures are immune to #6648 — neither touches the comment loop. `docstringCRLF` is merely the one that kills both clauses at once.

## Two gaps found, both filed rather than folded in

- **The contract assertion cannot see preserve-set *additions*.** Making `blank()` also preserve `\t` survives — on this head *and* on `main`. Length and both offset vectors are unchanged, so the assertion is structurally blind to it. The PR observes bytes being *removed* from the preserve set; nothing observes bytes being *added*. Same predicate, opposite direction.
- **The premise guard's region starts at byte 0.** It checks `tc.src[:at+len(tc.mustVanish)]`, a prefix from the start of the fixture, so a `\n` belonging to a *live code line* satisfies "the fixture carries a preserved byte inside a blanked region". The reviewer proved it is not theoretical by constructing a fixture that passes every guard and every assertion while observing nothing. All five shipped fixtures are genuinely non-vacuous — the mutant matrix proves it — so nothing is broken today, but the guard whose whole job is to stop a future vacuous fixture does not do it.

Also flagged and filed separately: an off-by-one in the triple-quote *blanked range* (`for k := i` → `i + 1`) survives on both this head and `main`, leaving the opening quote live in the masked copy. That is the caller's range rather than the writer's predicate, so it is not this PR's to close.

## Hygiene note worth repeating

The reviewer's own run was interrupted mid-mutation and left a mutant applied — which they caught themselves via the `shasum` check and restored by `cp`. Second instance today. The rule earns its place precisely because an interruption is when the restore does not happen.

Issue #1420, opened 2026-05-21
extractor: axum (Rust) HTTP endpoint extraction


cajasmota:
Rust axum Router routes/handlers not extracted. ShipFast pricing. #1409 coverage.

Issue #6643, opened 2026-08-25
the utoipa pass's three fail-to-mint shapes are recorded in a header comment and observed by no test


cajasmota:
`synthesizeUtoipaAxumRoutes`'s header records three shapes that deliberately mint nothing, and **no test observes any of them**:

- a comment between macro arguments — `routes!(\n a, // note\n b,\n)`
- a raw identifier — `routes!(r#type, a)`
- Rust-legal whitespace before the bang — `routes ! ( a , b )`

Grepped `internal/engine/http_endpoint_utoipa_axum_test.go` for `r#`, `/*` and `routes !`: **zero hits**. The three shapes exist only in the header comment at `http_endpoint_utoipa_axum.go:84-89`.

## Why pin a limitation rather than fix it

All three fail in the **safe** direction — they mint nothing rather than minting something wrong — and that is a defensible design choice for a regex-level pass. This issue is not asking to change the behaviour.

It is asking to **observe** it, for the reason #6418 is pinned rather than merely written down: an unobserved limitation cannot be distinguished from an unobserved regression. If someone later widens `utoipaRoutesMacroRe` and accidentally makes `routes ! ( a , b )` match, nothing objects; if someone tightens it and breaks a shape that *did* work, nothing objects either. The header becomes prose about behaviour instead of a description of it — which is this milestone's dominant defect class, and here it sits in the file whose header is the **source of truth for the coverage doc** (`docs/coverage/detail/lang.rust.framework.utoipa.md` now cites these three shapes explicitly).

## What is needed

1. Three table-driven cases asserting each shape mints **zero** definitions, with the doc comment saying they are deliberate Arm-A/B1 boundaries and not defects.
2. **Score a mutant per shape, before and after** — reproduce each as a survivor first. Suggested: widen the regex to tolerate whitespace before `!`; widen it to admit `r#` identifiers; make `utoipaMacroHandlerArgs` strip `//`-comments. Each must be a survivor on `main` and dead after.
3. **Vary the mutant family, not just the input** — the lesson from #6625 and #6632 this milestone. A fixture that pins one shape while another regex change slips through is the failure mode.
4. Premise-guard each fixture so the *macro-matching* branch is the thing observed, not an unrelated early return. `utoipaHasRoutesMacro` returns early when no `routes!` is present at all, so a fixture must still contain a recognisable macro for the case to mean anything.

## Traps

[...]

PR #6651, opened 2026-08-25
fix(#6560): B2a — same-file OpenApiRouter::nest prefix composition for utoipa_axum

cajasmota:
Implements **B2a** of #6560, as defined in the dedupe-scope decision comment: same-file `OpenApiRouter::nest("/prefix", …)` prefix composition for the utoipa_axum pass. **B2b — cross-file handler resolution and cross-file nest — is untouched and stays open**, so this deliberately does not say `Closes`.

## Withdrawn claim (review round 1)

The first version of this body said the scope-count clause "already counts `OpenApiRouter` correctly". **That claim is withdrawn, not edited into truth.** It was true for `OpenApiRouter::new()` and false for `OpenApiRouter::with_openapi(…)`, which is utoipa-axum's canonical constructor — and the fixture I wrote pinned the guard only in the spelling that happened to satisfy it. Review caught a real regression on that gap; `330343048` fixes it. Everything under "Review round 1" below is the correction.

## The defect

`synthesizeUtoipaAxumRoutes` minted the bare `#[utoipa::path]` path regardless of any `.nest(...)` that mounted the router. A service mounted at `/api` reported `GET /items`, not `GET /api/items`. The pass's own scope note already admitted it ("a nested utoipa router yields the unprefixed attribute path") — nothing observed it, which is why it stood.

## Extract or duplicate — the call, and why

**Extracted.** The grounding note said B2a was "pure reuse of `nestPrefixFor`", and that was not available as stated: `nestPrefixFor` was a closure inside `synthesizeAxumRoutes`, capturing a local `nests` slice.

Measured blast radius before choosing: `nestPrefixFor` had no other reference in the tree, and `axumNestRe` no reader outside `http_endpoint_axum.go`. Extraction therefore touches exactly the two passes — the floor — so duplicating buys nothing the mutant score buys more cheaply, and would leave two copies of a windowed heuristic to drift. (Review round 1 vindicated this concretely: the constructor-inventory fix below is one edit that repairs both passes. Under duplication it would have been two, and the axum copy's turbofish miss would likely still be there.)

Extracted as `rustNestEntries` / `rustNestPrefixFor` / `rustJoinNestPrefix`, window promoted to a named `rustNestWindow` constant. Both strategies preserved verbatim — inline chain and outer-function nest — with nearest-by-byte-distance winning across both.

**Axum behaviour proven unchanged, not observed green:** M1, M6, M7, M10b, M11 all kill the pre-existing `TestAxum_NestPrefix` / `TestAxum_NestMultipleRoutes`.

## Review round 1 — the phantom, and the constructor inventory

The scope guard counted the literal substring `Router::new()`. `OpenApiRouter::with_openapi(ApiDoc::openapi())` does not contain it, so two independently built routers read as one scope and the second's `.nest("/admin", …)` reached back into the first.

**Reproduced before the fix, in three spellings — each losing the real endpoint and inventing one:**

| Spelling | Pre-fix `ids` | Post-fix `ids` |
|---|---|---|
| `OpenApiRouter::with_openapi(ApiDoc::openapi())` | `[http:GET:/admin/health]` | `[http:GET:/health]` |
| `OpenApiRouter::default()` | `[http:GET:/admin/health]` | `[http:GET:/health]` |
| `OpenApiRouter::<AppState>::new()` | `[http:GET:/admin/health]` | `[http:GET:/health]` |

The substring is replaced by `rustRouterCtorRe` with a **stated inventory**, because one near-miss found by accident argues against patching spelling by spelling.

**Matched:**
- `Router::new()` / `OpenApiRouter::new()` and any `*Router` type, including a local `type` alias
- `::default()` — both `Router` and `OpenApiRouter` implement `Default`
- `OpenApiRouter::with_openapi(…)` — utoipa-axum only
- the turbofish form of all three, `Router::<S>::new()`

**Deliberately not matched, each stated with its reason:**
- a rename-import erasing the suffix (`use …OpenApiRouter as Api;` → `Api::new()`) — nothing in the text identifies `Api` as a router without resolving the import
- `OpenApiRouter::from(axum_router)` — real, but it *converts* an existing router rather than starting a chain, so counting it would split a scope that is genuinely one
- a constructor behind a user helper (`fn base() -> OpenApiRouter`)

Every miss above makes the guard *more permissive* — two scopes may read as one — which is the failure the fix addresses; but widening the matched set makes it *stricter*, so no miss can invent a path. The turbofish case was additionally a **pre-existing miss in the axum pass since #1420**, fixed here as a consequence of having extracted rather than duplicated.

## Scope discipline

Everything stays **file-scoped**. Neither the attribute map nor `utoipaRegisteredElsewhere` moves, so the #6530 duplicate-producer hazard is not approached. Resolution is per *macro*, at the `routes!` offset, so every handler in `routes!(a, b)` shares one mount point.

## Mutant score

Round 1/2 — survivor reproduced first (4 fixtures RED pre-fix); all DEAD after.

| # | Mutant | Killed by |
|---|---|---|
| M1 | `rustJoinNestPrefix` never applies the prefix | axum ×2 + all 4 utoipa positives |
| M2 | farthest nest wins instead of nearest | `NestPrefixNearestOfTwo` |
| M3 | window widened 2000 → 20000 | `NestPrefixBeyondWindow` |
| M4 | window narrowed 2000 → 500 | `NestPrefixWithinWindow` |
| M5 | scope clause dropped | `NestPrefixNotAcrossRouterScope` |
| M6 | strategy 2 dropped (nest AFTER ignored) | axum ×2, `…OuterFunction`, `…WithinWindow` |
| M7 | strategy 1 dropped (nest BEFORE ignored) | `…InlineChain`, `…NearestOfTwo` |
| M8b | utoipa **call site**: prefix computed then discarded | all 4 utoipa positives |
| M10b | axum **call site**: prefix discarded | `Axum_NestPrefix`, `Axum_NestMultipleRoutes` |
| M11 | join separator dropped (`prefix+rawPath`) | fixtures in **both** passes |
| M12 | nest resolved at offset 0, not the macro's | `…OuterFunction`, `…NearestOfTwo`, `…WithinWindow` |

Round 3 — the constructor inventory and the two review survivors. Phantom reproduced first (three fixtures RED).

|---|---|---|---|
| N1 | revert to the substring `Router::new()` count | **DEAD** | all three scope twins |
| N2 | inventory: drop `with_openapi` | **DEAD** | `…Scope_WithOpenapi` |
| N3 | inventory: drop `default` | **DEAD** | `…Scope_Default` |
| N4 | inventory: drop the turbofish group | **DEAD** | `…TurbofishScope` |
| N5 | inventory: exact `\bRouter\b` (excludes `OpenApiRouter`) | **DEAD** | all four scope fixtures |
| N6 | scope clause dropped entirely | **DEAD** | all four scope fixtures |
| N7 | ctor count always 0 (guard never fires) | **DEAD** | all four scope fixtures |
| N8 | scope clause over-strict (`> 0`) | **DEAD** | axum ×2 + 6 utoipa positives |
| A′ | `dist > rustNestWindow` → `>=` | **DEAD** *(was a survivor)* | `NestPrefixAtExactWindow` |
| C′ | `dist < bestDist` → `<=` (tie → last wins) | **SURVIVED** *(documented, see below)* | — |

N8 is the counterweight to N1–N7: it proves the wider inventory did not become a blanket rejection, since an over-strict guard costs eight fixtures their prefix.

### Survivor A — now pinned

`TestUtoipaAxum_NestPrefixAtExactWindow` builds a source whose `routes!`/`.nest` gap is **exactly** `rustNestWindow`, by measuring the unpadded gap and inserting the precise shortfall as comment filler, then asserts `gap == rustNestWindow` so drift fails as a broken premise. The bracketing fixtures pin the constant's magnitude; this pins the boundary byte.

### Survivor C — documented, not pinned

The tie-break between two nests equidistant on opposite sides is arbitrary by design: there is no reason a nest 300 bytes before beats one 300 bytes after. A fixture would freeze an arbitrary choice as though it were a decision, so it is written down above `rustNestPrefixFor` instead. What *is* pinned is that distance, not source order, decides when the distances differ.

### Negative result, recorded in-tree

**M9 — `prefix + "/" + rawPath` (drop the Trims): SURVIVED and is equivalent.** Both callers feed `httproutes.Canonicalize`, which ends in `normaliseSlashes`. Review confirmed it independently with a differential probe over 7.8M `(prefix, path)` pairs — 0 diffs, covering empty prefix, prefix exactly `/`, path starting `//`, and trailing slash. The reason is written above `rustJoinNestPrefix`. M11 is its observable neighbour and dies.

### Deliberately unpinned, upheld in review

`rustNestWindow` 2000 → 2100 survives, so the constant is pinned only to roughly (1.8k, 2.3k). That is correct strength for a heuristic with no semantic meaning: a byte-exact pin would fail on any legitimate re-tuning, which is a false alarm rather than a test.

## What the fixtures assert, and per-clause verdicts

Every fixture goes through `requireUtoipaDef`, which asserts the endpoint is minted **by this pass** — `framework=utoipa_axum` plus `source_handler=Controller:<handler>`. This is the premise guard: `utoipaRegisteredElsewhere` hands a handler to `synthesizeAxumRoutes` or `synthesizeRocket` whenever they also register it from the same file, so a fixture asserting only the ID would happily observe *their* nest handling and pass while this pass did nothing. The window fixtures additionally assert their measured byte gap.

Per clause, each verdict naming the mutant it is load-bearing against:

- **`dist > rustNestWindow` (the window)** — magnitude against M3/M4 via the bracketing pair; **boundary byte** against A′ via `NestPrefixAtExactWindow`. The band between the bracketing gaps stays unpinned, deliberately.
- **`n.offset < routeOffset` two-strategy split** — strategy 1 against M7, strategy 2 against M6. Neither fixture covers the other's branch, which is the point.
- **`dist < bestDist` (nearest wins)** — against M2 via `NestPrefixNearestOfTwo`, where both nests are in range and both pass the scope test, so distance is the only discriminator. The `<` vs `<=` tie-break is C′: unpinned, documented.
- **`rustRouterCtorCount(between) > 1`** — the clause against M5/N6/N7; the `> 1` threshold specifically against N8 (over-strict) in one direction and N6 (absent) in the other; the *inventory* against N1–N5 via the four scope fixtures.
- **`if prefix == ""` early return in the join** — load-bearing against nothing directly; kept for legibility, not claimed as pinned.
- **`prefix` resolved at `m[0]`** — against M12.

## Verification

- `go vet ./internal/engine/` clean on every mutant before running tests.
- `go test -count=1 ./internal/engine/` — full package, no `-run` filter — green; each mutant scored the same way, files restored by `cp` from byte snapshots and re-verified by `shasum`.
- `gofmt -l internal/engine/` clean.

Refs #6560.

### DIFF
--- a/internal/engine/http_endpoint_axum.go
+++ b/internal/engine/http_endpoint_axum.go
@@ -56,6 +56,53 @@ func axumHasAxum(content string) bool {
 		(strings.Contains(content, ".route(") || strings.Contains(content, "Router::"))
 }
 
+// rustRouterCtorRe counts router CONSTRUCTIONS, which is how rustNestPrefixFor
+// tells one router-building expression from the next.
+//
+// This was a literal `strings.Count(content, "Router::new()")` until #6560 B2a
+// review. That spelling is not the inventory — it is one member of it — and the
+// gap was not theoretical: `OpenApiRouter::with_openapi(ApiDoc::openapi())` is
+// utoipa-axum's CANONICAL constructor (the `::new()` form is the one that omits
+// the OpenAPI document), and it contains no `Router::new()`. Two independent
+// routers built that way read as one scope, so a `.nest("/admin", …)` in the
+// second reached back into the first and replaced a real `GET /health` with a
+// phantom `GET /admin/health`. Same for `Router::<AppState>::new()`, where the
+// turbofish splits the substring — that one had been missing from the axum pass
+// since #1420.
+//
+// STATED INVENTORY, so the next reader does not have to rediscover it by
+// accident. Matched:
+//
+//   - `Router::new()` / `OpenApiRouter::new()` and any `*Router` type,
+//     including a local alias (`type ApiRouter = OpenApiRouter;`)
+//   - `Router::default()` / `OpenApiRouter::default()` — both types implement
+//     Default, and `::default()` is a legal empty-router constructor
+//   - `OpenApiRouter::with_openapi(...)` — utoipa-axum only
+//   - the turbofish form of all three, `Router::<S>::new()`
+//
+// Deliberately NOT matched, and each is a MISS in the permissive direction (two
+// scopes may read as one, as the bug above did):
+//
+//   - a rename-import that erases the `Router` suffix
+//     (`use ...::OpenApiRouter as Api;` then `Api::new()`) — nothing in the text
+//     identifies `Api` as a router without resolving the import
+//   - `OpenApiRouter::from(some_axum_router)` — real, but it CONVERTS an
+//     existing router rather than starting a chain, so counting it would split
+//     a scope that is genuinely one
+//   - a constructor returned by a user helper (`fn base() -> OpenApiRouter`)
+//
+// Widening this set makes the guard STRICTER — more nests are rejected as
+// out-of-scope — so every miss above costs recall of a prefix, never a phantom.
+// That is the direction to fail in.
+var rustRouterCtorRe = regexp.MustCompile(
+	`\w*Router\b(?:::<[^>\n]*>)?::(?:new|default|with_openapi)\s*\(`)
+
+// rustRouterCtorCount reports how many router constructions appear in src. It
+// replaces a substring count; see rustRouterCtorRe for the inventory.
+func rustRouterCtorCount(src string) int {
+	return len(rustRouterCtorRe.FindAllStringIndex(src, -1))
+}
+
 // rustNestEntry records one `.nest("prefix", ...)` occurrence: the byte offset
 // of the `.nest` token and the static prefix it mounts at.
 type rustNestEntry struct {
@@ -110,10 +157,21 @@ const rustNestWindow = 2000
 //     registration but within rustNestWindow bytes ahead, as long as no second
 //     Router::new() separates them.
 //
-// The "at most one Router::new()" clause is what keeps a nest from reaching
-// across into a different router scope; note that `OpenApiRouter::new()`
-// contains `Router::new()` as a substring, so the utoipa spelling is counted
-// by the same test.
+// TIE-BREAK, documented rather than pinned (#6560 B2a review): when two nests
+// are EQUIDISTANT on opposite sides of the registration, the first one scanned
+// wins, because the comparison is strict `<`. That is arbitrary — there is no
+// reason a nest 300 bytes before is a better mount than one 300 bytes after —
+// and the mutant `<=` (last wins) survives the suite with a real witness. It is
+// left unpinned deliberately: a fixture would freeze an arbitrary choice as if
+// it were a decision, and either answer is equally defensible. What IS pinned is
+// that distance, not source order, decides when the distances DIFFER — see
+// TestUtoipaAxum_NestPrefixNearestOfTwo.
+//
+// The "at most one router construction" clause is what keeps a nest from
+// reaching across into a different router scope. Which spellings count is
+// rustRouterCtorRe's business, and its inventory is stated there — a substring
+// test for `Router::new()` looked like it covered the utoipa spellings and did
+// not.
 func rustNestPrefixFor(content string, nests []rustNestEntry, routeOffset int) string {
 	bestDist := -1
 	bestPrefix := ""
@@ -136,10 +194,10 @@ func rustNestPrefixFor(content string, nests []rustNestEntry, routeOffset int) s
 			start, end = routeOffset, n.offset
 		}
 		between := content[start:end]
-		// Allow at most one Router::new() (the one that .nest() or
+		// Allow at most one router construction (the one that .nest() or
 		// .route() is being called on). More than one means a different
-		// router scope.
-		if strings.Count(between, "Router::new()") > 1 {
+		// router scope. See rustRouterCtorRe for which spellings count.
+		if rustRouterCtorCount(between) > 1 {
 			continue
 		}
 		if bestDist < 0 || dist < bestDist {
--- a/internal/engine/http_endpoint_axum.go
+++ b/internal/engine/http_endpoint_axum.go
@@ -56,15 +56,127 @@ func axumHasAxum(content string) bool {
 		(strings.Contains(content, ".route(") || strings.Contains(content, "Router::"))
 }
 
+// rustNestEntry records one `.nest("prefix", ...)` occurrence: the byte offset
+// of the `.nest` token and the static prefix it mounts at.
+type rustNestEntry struct {
+	offset int
+	prefix string
+}
+
+// rustNestEntries collects every `.nest("prefix", ...)` call in a Rust source
+// file, in source order, with its byte offset.
+//
+// Extracted from synthesizeAxumRoutes (#6560 B2a) so the utoipa_axum pass can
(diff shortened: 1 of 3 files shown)


cajasmota:
Merging. The regression is closed, the inventory replaced the patch, and the extract choice paid for itself in a way nobody predicted.

## The phantom, reproduced in three spellings

I reported one (`with_openapi`). The rework found the wrong answer on `main`-correct input in **three**, and fixed all of them:

| constructor | before | after |
|---|---|---|
| `OpenApiRouter::with_openapi(ApiDoc::openapi())` | `/admin/health` — real endpoint lost | `/health` |
| `OpenApiRouter::default()` | `/admin/health` | `/health` |
| `OpenApiRouter::<AppState>::new()` | `/admin/health` | `/health` |

## The inventory found a bug that predates this PR

`Router::<AppState>::new()` — the turbofish form — has been missed by the axum pass **since #1420**, because the turbofish splits the counted substring. It is not a B2a regression; it is a pre-existing hole that surfaced only because the guard was being enumerated rather than patched.

And it is fixed here **as a consequence of having extracted rather than duplicated**: one edit repaired both passes. That is the concrete vindication of the extract-over-duplicate call, and it is worth recording, because the argument for extraction at the time was maintainability — a hypothetical. It turned out to be a bug fix in a pass nobody was looking at.

The stated misses are all in the **permissive** direction — a rename-import (`OpenApiRouter as Api`), `::from(…)` which converts rather than starts a chain, and a constructor behind a user helper. Widening the matched set makes the guard *stricter*, so a miss costs a prefix and can never invent one. That asymmetry is the right thing to have written down.

## N8 is the mutant I most wanted to see

Making the clause over-strict (`> 0`) is DEAD, killed by both axum fixtures and six utoipa positives. That is the counterweight proving the wider inventory did not quietly become a blanket rejection — the failure mode a bigger constructor set invites. Scoring the over-correction as well as the under-correction is what makes the widening safe rather than merely well-intentioned.

## A pinned, C documented — both correct calls

**A** (`dist > rustNestWindow` → `>=`) is now dead, with the fixture measuring the unpadded gap and inserting the exact shortfall so the gap is precisely `rustNestWindow`, asserting `gap == rustNestWindow` so drift fails as a broken premise rather than silently testing nothing. Twenty lines; deferring it would not have been defensible.

**C** (the equidistant tie-break) is documented rather than pinned, with the reason: the tie is arbitrary by design, and a fixture would freeze an arbitrary choice as though it were a decision. What stays pinned is that distance beats source order when distances differ. That is the right line.

## My own mutant — a survivor, filed not folded

Adding `from` to the constructor alternation **survives** (`go vet` 0, full package exit 0). So the in-tree decision that `::from(…)` converts an existing router rather than starting a new scope — and therefore must not be counted — is asserted in prose and observed by nothing.

Direction is safe (counting it would only suppress a prefix, never invent one) and the decision is written down, so this is not worth a third round on this PR. Filing it separately, the same way #6643 handled the fail-to-mint shapes.

## The claim was withdrawn, not repaired

The body now has its own section naming what was true (`::new()`), what was false (`with_openapi`), and that the fixture pinned the guard only in the spelling that satisfied it. That is the third time this shift a body has been corrected rather than edited into truth, and it is the habit worth keeping.

Issue #6655, opened 2026-08-25
the triple-quote blanked range is off-by-one-able: shifting its lower bound leaves the opening quote live


cajasmota:
The triple-quoted branch's blanking range is unobserved at its lower bound. Shifting it by one byte leaves the **opening quote live** in the masked copy and passes the whole engine package:

```diff
-				for k := i; k < j; k++ {
+				for k := i + 1; k < j; k++ {
```

(`internal/engine/http_endpoint_synthesis.go`, the triple-quote span loop.)

`go vet` **0**, full package `go test -count=1 ./internal/engine/` **exit 0** — on current `main` *and* with #6650's tests applied.

Non-equivalence witness: `"""doc"""` → pristine masks to 9 spaces; mutant yields `"\"        "` — the leading `"` survives into the masked copy.

## Why #6650 does not cover it

#6650 pinned `blank()`'s **write predicate** (which bytes are preserved). This is the **caller's range** (which bytes are handed to it). Same helper, different half — and the review that found this was explicit that it should not be held against that PR.

It is worth stating why the distinction keeps mattering here: the `pythonMaskInertRegions` cluster now runs to eleven items, and they divide into scanner / writer-predicate / writer-range. Each division was found only after the previous one was pinned. A mutant in one half is routinely invisible to every test written for the other.

## Is it harmful?

Probably mild, and the issue should establish this rather than assume it. A stray `"` at the start of a masked docstring cannot spell an identifier, but it *can* re-open a quote for any consumer that scans the masked copy for string boundaries — and `pythonMaskInertRegions`'s whole purpose is to hand downstream scans a copy with no live string content. Establish whether any consumer is affected before deciding how hard to pin it.

## What is needed

1. A fixture asserting the **full** triple-quoted span is blanked, opening delimiter included — the mutant above must die.
2. **Score before and after**: exit 0 on pristine `main` (survivor reproduced), exit 1 after.
3. Check the **upper** bound too (`k < j` → `k < j-1`, and the closing delimiter) — an off-by-one at the other end is the obvious sibling and nobody has scored it.
4. **Premise-guard so the triple-quote branch actually executes**, not merely that the fixture contains `"""`. A single-line-literal or comment fixture would observe nothing.
5. **Vary the mutant family** — lower bound, upper bound, and both — reporting which fixture pins which. And name each per-clause verdict *against which mutant*: an unqualified "load-bearing" has been an incomplete claim three times in this cluster.

## Traps

[...]

Issue #6654, opened 2026-08-25
blank()'s preserve set is pinned only against removals — adding \t to it survives, and the premise guard's region starts at byte 0


cajasmota:
#6649 pinned `blank()`'s preserve set in one direction only. Bytes **leaving** the set are now observed; bytes **joining** it are not, and the contract assertion is structurally incapable of seeing them.

## The survivor

```diff
-		if b[i] != '\n' && b[i] != '\r' {
+		if b[i] != '\n' && b[i] != '\r' && b[i] != '\t' {
```

`go vet` **0**, full package `go test -count=1 ./internal/engine/` **exit 0** — on current `main` *and* with #6650's new tests applied.

Non-equivalence witness: `"# a\tb\n"` → pristine `"     \n"`, mutant `"   \t \n"`. A tab inside a comment survives into the masked copy.

**Why the assertion cannot see it.** #6650 asserts `len(got) == len(src)` plus the ordered `\n` and `\r` offsets. Adding a byte class to the preserve set changes none of those: the length is identical and no terminator moved. The premise guard would catch it only if some fixture's `mustVanish` marker contained a tab — none does.

Severity is genuinely low: a stray tab cannot spell an identifier the mount scan reads, so no downstream defect is known. But it is **non-equivalent and unobserved**, on the very predicate #6649 set out to pin, which makes it worth closing while the context is fresh.

## The second, related gap — the premise guard's region starts at byte 0

`http_endpoint_synthesis_python_mask_6649_test.go` documents that each fixture must carry a preserved byte **"INSIDE a region that gets blanked"**. The check is:

```go
region := tc.src[:at+len(tc.mustVanish)]
```

That is a prefix from **byte 0**, so it includes live code *before* the inert region. A `\n` belonging to an ordinary code line satisfies it.

This was proven non-theoretical by the adversarial review, with a fixture that passes every premise guard and the contract assertion while observing nothing:

```go
{name: "guardHole6650", caller: "A: `#`-comment loop",
 src: "import os\n# app.include_router(r)\napp = FastAPI()\n",
 mustVanish: "include_router"},
```

Under `blankcr` it passes — its only terminator is the code line's `\n`, which is never handed to `blank()`.

All five shipped fixtures are genuinely non-vacuous (the mutant matrix proves each fails under at least one mutant), so **nothing is broken today**. The defect is that the guard whose entire job is to stop a *future* vacuous fixture does not do it.

## What is needed

[...]

Issue #6653, opened 2026-08-25
the single-line literal scan breaks on \n only, so a lone carriage return moves the #6418/#6614 boundary


cajasmota:
`pythonMaskInertRegions`'s **single-line literal scan** breaks on `\n` only:

`internal/engine/http_endpoint_synthesis.go:3202`

```go
			if b[j] == c || b[j] == '\n' {
				break
			}
```

Python's universal newlines treat a bare `\r` as a line terminator, so an unterminated `'…` literal is a syntax error at the `\r`, not at the `\n`. We keep scanning past it.

## Why this is split out of #6648

#6648 covers the `#`-comment loop, where the answer is simple and one-directional: a lone `\r` ends the comment, so a `\r`-aware change means **less masking** and fixes a false negative.

This site is **not** one-directional, and that is the whole reason it needs its own issue and its own mutants:

- **Terminated literals** — a closing quote sitting after a `\r` would stop counting as terminated, so the literal falls into the unterminated path and its span gets blanked. **More** masking.
- **Unterminated literals** — the blanked span would end at the `\r` instead of at the `\n`. **Less** masking of the tail.

Those two effects move the **#6418/#6614 boundary**, which is a deliberate and tested trade-off: #6418 records that a *terminated* single-line literal is left intact on purpose, because the mount prefix the scan reads lives inside one; #6614 established that an *unterminated* literal's own tail must be blanked. Changing which literals count as terminated touches both decisions at once.

So this is a behaviour change with a real risk budget, not a regex tweak that can ride along on a comment-loop fix.

## Reachable with real files

Established while grounding #6648: there is **no CR normalisation anywhere on the Python path** — `os.ReadFile`-raw at `cmd/grafel/index.go:3954`, into `FileInput.Content` at `:3988`, forwarded as `DetectorPassArgs.Content` at `internal/engine/detector.go:552`, stringified into `synthesizeFastAPI` at `:965`, and nothing in `internal/treesitter` strips CR. Old-Mac line endings, or a CR introduced mid-line by a generator or a merge, reach this in production.

## What is needed

1. **Decide, with the #6418/#6614 trade-off stated explicitly.** Both directions have a cost: leaving it as-is keeps a false negative (we scan past a line boundary Python respects); changing it may blank a literal that today legitimately yields a `prefix=` value. Say which cost is being accepted and why.
[...]

PR #6658, opened 2026-08-25
fix(#6652): mount StartLine ignored lone-CR line terminators, under-reporting one line per CR

### LINKED ISSUE #6652: StartLine counts only \n, so a lone carriage return makes every later entity report the wrong line
`StartLine` for synthesised FastAPI endpoints counts `\n` only, so in any Python file containing a **lone `\r`** our reported line number is lower than Python's — and every tool that navigates to that line lands on the wrong one.

`internal/engine/http_endpoint_synthesis.go:3388`:

```go
StartLine: 1 + strings.Count(content[:loc[0]], "\n"),
```

Python's universal newlines treat a bare `\r` as a line terminator. We do not count it, so every entity after the first lone `\r` in a file is off by one line — cumulatively, one per CR.

## This is independent of the masking question

Filed separately from **#6648** on purpose. #6648 asks whether the `#`-comment scanner should *stop* at a lone `\r`; this is the offset→line map, and it is wrong **regardless of how #6648 is decided**. Nothing about the masking behaviour changes `strings.Count(…, "\n")`.

It is also more directly user-visible than the masking question: a wrong `StartLine` is what an MCP consumer, the dashboard, or a jump-to-definition uses. A mis-masked region produces a missing or spurious endpoint; a mis-counted line sends a reader to the wrong place in a file that is otherwise correct.

## It is reachable with real files

Grounding for #6648 established there is **no CR normalisation anywhere on the Python path**: bytes are `os.ReadFile`-raw at `cmd/grafel/index.go:3954`, put into `extractor.FileInput.Content` at `:3988`, forwarded verbatim as `DetectorPassArgs.Content` at `internal/engine/detector.go:552`, and stringified into `synthesizeFastAPI` at `:965`. Nothing in `internal/treesitter` strips CR either. The only `\r\n`→`\n` normalisations in the repo are in the Astro and VB.NET paths.

So this needs a real file with old-Mac line endings, or a CR introduced mid-line by a generator or a bad merge — not a synthetic unit-test-only construction.

## The same assumption is broader than this line

Grounding also found several `\n`-only assumptions in the same neighbourhood, listed here as context rather than scope: `fastapiFileEvidenceRe` uses `(?m)^`, which in Go anchors after `\n` only; `fastapiTopLevelArgs` and `fastapiFirstCallArg` break their scans on `\n`; and `drfImportFromRe` (`django_drf_actions.go:86`) runs its name list through a `\r`. Interestingly the **regex layer already treats `\r` as a terminator** in places — `fastapiRouterPrefixKwargRe` uses `[^"'\n\r]*` — so the two layers disagree with each other today.

**Do not fix all of those here.** This issue is the line map. The rest is worth a survey once the line map has a decided answer.

## What is needed

1. Decide what `StartLine` should mean for a lone `\r` and make it agree with Python. Write the decision in the doc comment.
2. **Score a mutant before and after** — reproduce the miscount as a survivor first, then dead. A fix not first reproduced as a failure is not evidence.
3. Assert the **observable**: the reported `StartLine` of an entity that follows a lone `\r`, not an internal counter.
4. Check the neighbours that share the assumption and say which ones your change does and does not touch, so the next reader knows what remains.

## Traps

1. **Do not** normalise CR globally as a shortcut. That changes byte offsets, and `pythonMaskInertRegions` has a documented contract that its output has the same byte length and the same newline positions as the source — several tests now pin exactly that (#6649).
2. Well-formed CRLF is already correct under the current code (`\r\n` contains a `\n`), so a CRLF fixture observes nothing. The fixture needs a **lone** `\r`.
3. Sibling scanners share the `\n`-only assumption (`solidity/extractor.go:1322`, `erlang/extractor.go:1201`) but are out of scope here.

Found while grounding #6648. Refs #6648, #6649.

cajasmota:
Closes #6652

## The defect

`internal/engine/http_endpoint_synthesis.go:3388` computed a synthesised FastAPI mount's line as:

```go
StartLine: 1 + strings.Count(content[:loc[0]], "\n"),
```

CPython's universal newlines terminate a line on `\n`, on `\r\n`, **and on a lone `\r`**. Counting only `\n` reports a line number that is low by one for every lone CR that precedes the entity — cumulatively, one per CR. This is the number an MCP consumer, the dashboard, or a jump-to-definition navigates to, so the reader lands in the wrong place in a file that is otherwise analysed correctly.

It is reachable with real bytes: there is no CR normalisation anywhere on the Python path (`os.ReadFile`-raw at `cmd/grafel/index.go:3954` → `FileInput.Content` at `:3988` → `DetectorPassArgs.Content` at `internal/engine/detector.go:552` → `synthesizeFastAPI` at `:965`), and nothing in `internal/treesitter` strips CR.

## The decision

`pythonLineOfOffset` reports the line **the reader of the file sees** — the line CPython's tracebacks, `linecache`, and any editor honouring old-Mac endings report. All three terminators count as exactly one, so `\r\n` advances the line by one, not two. The decision and its scope are written into the helper's doc comment.

It is a **read-only view**: nothing is normalised. Normalising CR would change byte offsets and break `pythonMaskInertRegions`' documented contract that its output has the same byte length and the same newline positions as its input, pinned by #6649/#6650.

## Reproduced before fixed

The failing assertions are on the **observable** — the emitted entity's `StartLine` — never an internal counter (#6613). On `main`:

```
LoneCR              : StartLine = 1, want 4
MixedCRandLF        : StartLine = 4, want 6
CRAtOffsetBoundary  : StartLine = 3, want 4
```

All pass after the fix. The expected values come from an oracle in the test file that re-derives the line by scanning terminators one at a time, written deliberately unlike the production helper so it cannot inherit the same mistake.

While building the first fixture, a `#` comment before a lone `\r` suppressed the mount **entirely** (0 synthetics) — that is #6648, a different mechanism. The fixture was rewritten without a comment so it observes the line map and nothing else; the reason is recorded in a comment on the fixture.

## Mutants scored

| Mutant | Before | After |
|---|---|---|
| count `\n` only (the original bug) | SURVIVED | KILLED — 6 tests |
| count `\r` only | — | KILLED — MixedCRandLF, **LFOnly control**, LFThenCR, Adjacency |
| count `\r\n` as two | — | KILLED — **CRLF control**, CRAtOffsetBoundary, SplitCRLFPair, Adjacency |
| delete the split-pair guard | — | KILLED — SplitCRLFPair |
| also subtract `\n\r` | **SURVIVED** (review) | KILLED — LFThenCR, Adjacency |
| drop the guard's `\r` half | **SURVIVED** (review) | KILLED — OffsetOnLoneLF |
| relax guard bounds to `off <= len` | **SURVIVED** (review) | KILLED — OffsetAtEOFAfterLoneCR (**panic**) |
| revert site `:3547` to `lineOfOffset` | — | KILLED — DecoratorRoute |
| revert site `:3568` to `lineOfOffset` | **SURVIVED** (round 1) | KILLED — ApiRouteDecorator |
| revert site `:3618` to `lineOfOffset` | **SURVIVED** (round 1) | KILLED — AddApiRouteFallback |

### The `\n\r` survivor (found in review)

The first round of this PR shipped a fix that was correct but a suite that could not tell:

```go
line := … - strings.Count(prefix, "\r\n") - strings.Count(prefix, "\n\r")  // ← passed everything
```

`go vet` clean and the **full** `go test -count=1 ./internal/engine/` package suite exit 0. Only `\r\n` collapses to one terminator; `\n\r` is **two** — the `\n` ends a line and the bare `\r` ends the empty line after it. Witness: for `"a\n\rb\n\rc"` the helper returns `5` (Python's answer), the mutant `3`.

The original five fixtures covered CRLF, lone CR, mixed CR/LF, LF-only and the split pair, but none placed a `\n` **immediately** before a `\r`. That shape is real — a CRLF file edited by a tool that writes bare LF, or endings reversed by a bad transform. Now pinned at both levels: the entity fixture reports `3` instead of `5` under the mutant, and a helper adjacency table covers it directly.

The adjacency table also pins two shapes **no existing case reached**: `\r\n\r\n` (two pairs → two lines) and `\r\r\n` (lone CR then a pair → two lines); the pre-existing `SplitCRLFPair` source `"a\r\nb\r\nc"` never places two terminators next to each other. `\r\r` and `\n\n` are covered too.

### The guard's `\r` half — a second survivor, three lines away

Review then caught a survivor *inside the guard we had just decided to keep*:

```go
if off > 0 && off < len(content) && content[off] == '\n' {   // ← `content[off-1] == '\r' &&` removed; passed everything
```

`go vet` clean, full package suite exit 0. Without the `\r` check the guard fires on **every plain Unix newline**, decrementing a line number that was never incremented. For `"a\nb\nc"`:

| offset | pristine | mutant |
|---|---|---|
| 1 (lands on a lone `\n`) | 1 | **0** |
| 3 (lands on a lone `\n`) | 2 | **1** |

Line 0 is not a wrong line, it is an impossible one — and this is precisely the failure the keep-the-guard argument was about. `pythonLineOfOffset` was kept general because unreachable-today is a property of one caller's regex, not of the helper; a general helper that returns 0 for an ordinary offset is broken for exactly the callers that generality was preserved for.

`SplitCRLFPair` already covered the `content[off] == '\n'` half — its offsets land on the `\n` **of a pair**. Nothing passed an offset landing on a **lone** `\n`, where the guard must not fire. Now pinned, together with an explicit assertion that no offset may ever yield a line below 1.

The symmetric neighbour — an offset landing on a **lone `\r`** — turned out to be **already covered**: offset 2 of `"a\n\rb\n\rc"` in the adjacency table, where the `\r` is followed by `b` and so is not part of a pair. Relabelled to say so rather than duplicated.

### The split-pair guard — kept, by review ruling

The `\r\n`-split-pair guard was an unpinned survivor in the first round. It is unreachable from the entity path today: `fastapiIncludeRouterCallRe` starts its match on `[A-Za-z_]`, so a mount offset is never the `\n` of a `\r\n` pair.

Review ruled to **keep** it. Unreachable-today is a property of one caller's regex, not of the helper — `pythonLineOfOffset` is a general offset→line function and the next caller may pass an arbitrary offset, so deleting a correct guard because the only current caller cannot reach it would make the helper quietly caller-specific. It is pinned by a direct table test on the helper, which observes it at the level where it is actually a contract rather than manufacturing an artificial entity-path fixture.

### Withdrawn: this fix was PARTIAL, and I described it as complete

Round 1 changed the line map at **one** site. Review found two more **inside `synthesizeFastAPI` itself**, so a single lone-CR file produced a correct mount line and a wrong route line from the same pass, same language, same observable:

```
http_endpoint            http:ANY:/api:mount  StartLine=4   ← correct
http_endpoint_definition http:GET:/ping       StartLine=1   ← Python line 6
```

**What I claimed, and what is actually true.** In my round-3 report I wrote that the shared `lineOfOffset` "serves the NestJS/JS path, not Python". That is **false**, and worse than the reviewer knew:

- `lineOfOffset` (`:6562`) has **29 call sites, and not one of them is JS**. 28 are Python synthesizers (Flask, Starlette, Tornado, Pyramid, Sanic, Litestar, Robyn, Aiohttp, Bottle, CherryPy, Falcon, Hug, Quart, Strawberry, Graphene, Ariadne) plus `findPyDefLine`; the 29th, `synthesizeGqlgen`, is **Go**.
- The JS `lineOfOffset` I was thinking of is a **different function in a different package** — `internal/substrate/jsts.go:198`. I conflated two same-named helpers and then reasoned from the conflation.
- So `lineOfOffset` was, and still is, the live line map for ~28 **Python** call sites. The Python line map is **not** correct after this PR.

The claim was in my report rather than in this body, but the body's silence about the other sites implied the same thing, so it is withdrawn here where the reviewer of record can see it.

**What this PR now fixes:** 4 sites — the mount synthetic, `:3547`, `:3568`, `:3618`.
**What remains wrong:** the other 29, tracked in #6660 with the full per-function line list.

One live gap worth naming precisely: `synthesizeFastAPIAddRoute` prefers `findPyDefLine` and only falls back to the fixed helper. So `add_api_route` is now correct for a handler defined **elsewhere** and still wrong for one defined **in the same file**. The fixture here deliberately uses an external handler so it tests the branch this PR actually changed.

**JS, as asked:** yes, `internal/substrate/jsts.go:198` is `\n`-only and ECMAScript's `LineTerminator` includes CR, so it is wrong for its own JS/TS callers too. Not touched; recorded in #6660.

### The guard conjunction, enumerated

Three reviews each found one survivor in the same eight lines, one per term. Rather than wait for a fourth, all four terms are now mutated and scored:

| Term | Mutation | Result |
|---|---|---|
| `off > 0` | drop / relax to `>= 0` | KILLED — SplitCRLFPair (panic, index `-1`) |
| `off < len(content)` | relax to `<=` / drop | KILLED — OffsetAtEOFAfterLoneCR (panic, index `6`) |
| `content[off-1] == '\r'` | drop / flip to `'\n'` | KILLED — OffsetOnLoneLF, GuardConjunction |
| `content[off] == '\n'` | drop / flip to `'\r'` | KILLED — 5 tests / SplitCRLFPair |

8 mutations, 8 dead. The bounds one is not a wrong line but a **panic**: `pythonLineOfOffset("x = 1\r", 6)` indexes out of range, and every prior case reaching `off == len(content)` had a non-`\r` byte before it, so the `&&` short-circuited first.

## Controls

- **CRLF** — well-formed `\r\n` was already correct under the `\n`-only count, so it observes nothing about the bug. Its job is to fail if the fix double-counts a pair, and it does exactly that against the `\r\n`-as-two mutant.
- **LF-only** — the ordinary Unix file keeps the line it has always reported; it kills the `\r`-instead-of-`\n` mutant.

## Neighbours still carrying the `\n`-only assumption — NOT changed here

Separately tracked, deliberately untouched:

- **#6648** — the `#`-comment loop's terminator (`http_endpoint_synthesis.go:3167`).
- **#6653** — the single-line literal scan's terminator (`:3202`, and the sibling scans at `:3061` / `:3258`).
- **#6654 / #6655** — `blank()`'s preserve set and the triple-quote blanked range.

Surveyed and left alone, so the next reader knows what remains:

- `fastapiFileEvidenceRe` (`:2963`) uses `(?m)^`, which in Go anchors after `\n` only.
- `fastapiTopLevelArgs` and `fastapiFirstCallArg` break their scans on `\n`.
- `drfImportFromRe` (`django_drf_actions.go:85`) runs its name list through a `\r`.
- The two layers **disagree today**: `fastapiRouterPrefixKwargRe` (`:2897`) already excludes `\r` via `[^"'\n\r]*`, while the hand-written scanners do not.

Sibling scanners in other languages share the assumption and are out of scope: `solidity/extractor.go:1322`, `erlang/extractor.go:1201`.

**The other 29 `lineOfOffset` call sites in this file are tracked in #6660**, with the full per-function line list, the Go-not-Python exception (`synthesizeGqlgen`), the `findPyDefLine` gap, and the JS/TS helper.

## Verification

`gofmt -l internal/engine/` clean, `go vet ./internal/engine/` clean, and the full `go test -count=1 ./internal/engine/` package suite passes (no `-run` filter).

All mutant scores and suite runs in this PR were re-derived through `rtk proxy` and scored on exit codes: a bare `go test` was hook-rewritten to the `rtk` wrapper mid-session and reported a fabricated `Go test: 8 passed` for a run that produced no such result. That wrapper has previously fabricated `grep` counts and stale `git` state in this repo, so none of the numbers above come from it.

### DIFF
--- a/internal/engine/http_endpoint_synthesis.go
+++ b/internal/engine/http_endpoint_synthesis.go
@@ -3544,7 +3544,7 @@ func synthesizeFastAPI(content, relPath string, emit emitDefFn, emitMount func(t
 			continue
 		}
 		canonical := httproutes.Canonicalize(httproutes.FrameworkFastAPI, path)
-		defLine := lineOfOffset(content, idx[8])
+		defLine := pythonLineOfOffset(content, idx[8])
 		emit(verb, canonical, "fastapi", "Controller", handler, defLine)
 	}
 
@@ -3565,7 +3565,7 @@ func synthesizeFastAPI(content, relPath string, emit emitDefFn, emitMount func(t
 			continue
 		}
 		canonical := httproutes.Canonicalize(httproutes.FrameworkFastAPI, path)
-		defLine := lineOfOffset(content, idx[8])
+		defLine := pythonLineOfOffset(content, idx[8])
 		methods := parseFlaskMethods(tail) // same methods=[...] shape
 		if len(methods) == 0 {
 			methods = []string{"GET"}
@@ -3615,7 +3615,7 @@ func synthesizeFastAPIAddRoute(content string, emit emitDefFn) {
 			defLine = findPyDefLine(content, refName)
 		}
 		if defLine == 0 {
-			defLine = lineOfOffset(content, idx[0])
+			defLine = pythonLineOfOffset(content, idx[0])
 		}
 		for _, verb := range methods {
 			emit(verb, canonical, "fastapi", refKind, refName, defLine)
--- a/internal/engine/http_endpoint_synthesis.go
+++ b/internal/engine/http_endpoint_synthesis.go
@@ -3279,6 +3279,53 @@ func fastapiTopLevelArgs(src string, open int) (string, bool) {
 	return "", false
 }
 
+// pythonLineOfOffset returns the 1-based line number of byte offset `off` in a
+// Python source file, using CPython's UNIVERSAL-NEWLINE rule (#6652): a line is
+// terminated by "\n", by "\r\n", or by a LONE "\r". All three count as exactly
+// one terminator, so "\r\n" advances the line number by one, not two.
+//
+// THE DECISION (#6652). We report the line the READER of the file sees — the
+// same line CPython's tracebacks, `linecache`, and every editor that honours
+// old-Mac line endings report. The alternative, counting "\n" only, silently
+// under-reports by one line per lone CR, cumulatively, so an MCP consumer or a
+// jump-to-definition lands somewhere earlier in a file that is otherwise
+// correctly analysed.
+//
+// This is a READ-ONLY view over the source: it does NOT normalise anything.
+// Normalising CR would change byte offsets and break the pythonMaskInertRegions
+// contract that its output has the same byte length and the same newline
+// positions as its input (#6649/#6650).
+//
+// SCOPE. This is the offset->line map for synthesised Python endpoints only.
+// Several neighbours still assume "\n" is the sole terminator and are tracked
+// separately: the `#`-comment scanner's terminator (#6648), the single-line
+// literal scan's terminator (#6653), and blank()'s preserve set and the
+// triple-quote blanked range (#6654/#6655). Also unchanged here, and noted for
+// the next reader: fastapiFileEvidenceRe's `(?m)^` (Go anchors after "\n"
+// only), the "\n" break in fastapiTopLevelArgs and fastapiFirstCallArg, and
+// drfImportFromRe in django_drf_actions.go, which runs its name list through a
+// "\r". The regex layer already disagrees with the hand-written scanners:
+// fastapiRouterPrefixKwargRe excludes "\r" today.
+func pythonLineOfOffset(content string, off int) int {
+	if off < 0 {
+		off = 0
+	}
+	if off > len(content) {
+		off = len(content)
+	}
+	prefix := content[:off]
+	// Count "\r" and "\n" and remove the double-count on each "\r\n" pair that
+	// lies wholly inside the prefix.
+	line := 1 + strings.Count(prefix, "\n") + strings.Count(prefix, "\r") - strings.Count(prefix, "\r\n")
+	// A pair SPLIT by the offset — prefix ends on the "\r", the "\n" is at
+	// content[off] — has not been crossed yet: the offset still sits on the
+	// terminator of the current line, so the trailing "\r" must not advance it.
+	if off > 0 && off < len(content) && content[off-1] == '\r' && content[off] == '\n' {
+		line--
+	}
+	return line
+}
+
 // fastapiMountPointSynthetics emits one ADDITIVE `url_mount_point` synthetic
 // per `include_router(..., prefix="<literal>")` mount site found in THIS file
 // (#6385, reported by @arthurgeron).
@@ -3385,7 +3432,7 @@ func fastapiMountPointSynthetics(content, relPath string) []types.EntityRecord {
 			Name:               mountID,
 			Kind:               httpEndpointKind,
 			SourceFile:         relPath,
-			StartLine:          1 + strings.Count(content[:loc[0]], "\n"),
+			StartLine:          pythonLineOfOffset(content, loc[0]),
 			Language:           "python",
 			EnrichmentRequired: false,
 			EnrichmentStatus:   types.StatusPending,
--- a/internal/engine/mount_point_startline_cr_6652_test.go
+++ b/internal/engine/mount_point_startline_cr_6652_test.go
@@ -231,7 +231,10 @@ func TestIssue6652_PythonLineOfOffset_TerminatorAdjacency(t *testing.T) {
 		// "\n\r" — TWO terminators. This is the survivor the rest of the suite
 		// missed: a `- strings.Count(prefix, "\n\r")` term returns 3 here.
 		{`"\n\r" adjacency, end of "a\n\rb\n\rc"`, "a\n\rb\n\rc", 7, 5},
-		{`"\n\r" adjacency, on the first \r`, "a\n\rb\n\rc", 2, 2},
+		// Offset lands ON a LONE "\r" (content[3] is 'b', so this \r is not
+		// part of a pair). This is the symmetric neighbour of the lone-"\n"
+		// offsets pinned by TestIssue6652_PythonLineOfOffset_OffsetOnLoneLF.
+		{`"\n\r" adjacency, ON a lone \r`, "a\n\rb\n\rc", 2, 2},
 		{`"\n\r" adjacency, first byte of line 3`, "a\n\rb\n\rc", 3, 3},
 		// "\r\n\r\n" — two well-formed pairs, so two lines, not four.
 		{`two CRLF pairs "a\r\n\r\nb"`, "a\r\n\r\nb", 5, 3},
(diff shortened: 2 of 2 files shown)


cajasmota:
Merging. Four rounds, four survivors found in roughly eight lines of arithmetic, and the last one was found by the author against their own work.

## My own mutant

Reverting **only** the `add_api_route` fallback site (`defLine = pythonLineOfOffset(…)` → `lineOfOffset(…)`): `go vet` 0, full package **exit 1 — DEAD**, sole failure `TestIssue6652_LoneCR_AddApiRouteFallbackStartLine`. Source restored, worktree swept.

I picked that site deliberately, because it is exactly the one the author reported as having been **fixed but not pinned** in their first pass.

## The self-caught finding is the most valuable thing here

> "my first pass fixed all three but only pinned one — reverting `:3568` or `:3618` alone SURVIVED. I'd have shipped two sites 'fixed but merely assumed'."

That is the same defect class as the blocking finding, one level down, and it was found by re-scoring rather than by review. Three sites changed, three fixtures, and the mount test additionally asserts the two sites agree **with each other** — because intra-file disagreement is what the original witness actually showed.

## The framing was worse than either of us thought

I asked for a claim to be withdrawn on the grounds that `lineOfOffset` serves ~30 Python sites rather than JS. The truth: it has **29 call sites and not one is JS** — 28 Python synthesizers plus `findPyDefLine`, and one Go. The JS helper is a different function in a different package; two same-named helpers were conflated and then reasoned from.

Also correctly noted: the claim was never in the PR body, only in a report to me — and it was withdrawn in the body anyway, because the body's *silence* about the other sites implied the same completeness. That is the right instinct about what a body claims by omission.

## The exception that justifies a deliberate pass

`synthesizeGqlgen:5746` is **Go**, not Python, and a lone `\r` is not a Go line terminator — so it must keep `lineOfOffset`. A bulk rename across the 29 sites would have broken it silently. That single exception is why #6660 argues for a per-function pass rather than sed, and it is worth more than the site list itself.

Two more from #6660: `findPyDefLine:4244` is a live gap **inside code this PR touched** — `add_api_route` prefers it and only falls back to the fixed helper, so it is now correct for an external handler and still wrong for a same-file one, with the fixture deliberately exercising the branch that changed. And JS **is** wrong for its original callers, since ECMAScript's `LineTerminator` includes CR while `internal/substrate/jsts.go:198` counts `\n` only.

## The guard, finally exhausted

Eight mutations across all four conjunction terms, all dead — including two that **panic** rather than mis-count (`off > 0` dropped → index `-1`; `off < len` relaxed → index `6`). Three of those four terms had to be demanded by a separate round of review. Walking a conjunction term-by-term is now the standing expectation here, not an optimisation.

The arithmetic itself is unchanged since it was proven correct at all lengths by construction and by 398,579 enumerated (string, offset) pairs.

Production diff: 6 lines across 3 call sites. Everything else is fixtures and the follow-up map in #6660.

Issue #6635, opened 2026-08-25
the parse-error canary's spike verdict has zero consumers, and the one real assertion cannot detect a new-syntax gap


cajasmota:
The parse-error canary's verdict has **zero consumers**. On a spike, production writes a stderr WARN and a JSON field, and nothing anywhere reads either.

`cmd/grafel/index.go:3466-3475`:

```go
fmt.Fprintf(os.Stderr, "grafel: WARN parse-error canary SPIKE ...")
```

The boolean propagates to `sidecar_build.go:47` as `parse_error_spike`. A tree-wide search for `ParseErrorSpike` outside tests returns exactly three hits: the write, the struct field, and a comment. **No exit code, no CI step, no assertion.**

## Why this matters more than it looks

`docs/language-release-calendar.md` names the canary as one of the two mechanisms that would catch a language-release regression, and the generated watch issue #5673 repeats that. tree-sitter is error-tolerant, so a grammar that stops recognising new syntax **never breaks indexing** — it silently emits `ERROR` nodes. That is exactly the silent-wrongness shape this project ranks above loud breakage, and the canary exists to be the instrument that sees it.

An instrument whose reading nothing consumes is not an instrument. Worse than absent: it is *cited* as coverage, so the gap looks closed.

## The second half — the one real assertion cannot detect the thing it is cited for

`internal/treesitter/parse_canary_test.go:833` does assert, against `maxDominatedRatio = 0.50` (`:76`). But 50% means *more than half the tree is ERROR nodes*. A grammar that fails to parse one new statement produces an error ratio of a few percent. That threshold detects a total recovery storm — the ABI-mismatch failure the file header (`:5-32`) says it was written for — and **cannot** detect a new-syntax gap.

The corpus says so itself, at `parse_canary_test.go:82`:

> "Keep these idiomatic and **within the currently-pinned grammars' syntax (no bleeding-edge constructs)** so the canary stays a regression net, **not a feature gate**."

That comment is honest and the design is defensible on its own terms. The defect is that the calendar cites this instrument for a job its own source code declines to do. The Go sample (`:264`) is pre-1.18 — no generics, no range-over-func.

[...]

PR #6661, opened 2026-08-25
docs(#6636): say plainly that A2 and A4 cannot catch a language-release regression

cajasmota:
`docs/language-release-calendar.md` named the **A2 grammar-freshness cron** and the **A4 parse-error canary** as the two mechanisms that would catch a language-release regression. Grounding against `main` confirms **neither can do that job**. Both work as designed — the design does not cover that question — and the doc did not say so, which made the gap look closed.

Docs-only. No code changed; both defects stay tracked where they belong.

## What the doc now says

A new section, **"What A2 and A4 cannot detect today"**, states plainly what each mechanism *can* and *cannot* detect, with the reason and a pointer to the blocking work. The intro, the irregular-languages note, the alarm table, and checklist step 1 all link to it.

**The per-release checklist is kept.** The procedure was never the problem — what was wrong is the implied automation behind it. Step 1 is now marked manual, with instructions to read `parse_error_canary.languages[]` yourself rather than wait for a verdict nothing consumes.

## A2 — accurate, but the baseline never moves

Every grammar is compared against **one global date**: `bundled := parseDate(lock.Binding.PinnedDate)` (`tools/grammar-freshness/check.go:50`, stamped at `:56`, compared at `:70-75`). There is no per-grammar baseline available to use instead:

- The vendored bundle carries **no per-grammar provenance** — `grammars.lock` `binding.notes` (line 15): *"only ABI LANGUAGE_VERSION numbers in parser.h, so bundled_version below is recorded as the binding snapshot date"*. All 27 entries share `"bundled_via": "smacker@2024-08-27"` (`:26-52`).
- The pin **cannot advance**: `binding_is_at_upstream_head: true`, head date 2024-08-27, "appears unmaintained" (`:12-15`).

Blocked on the **B2 decouple** (`decouple_target_b2`, `grammars.lock:17-23`).

The check *does* report magnitude (`check.go:73` `Behind`, sorted `:98-99`, rendered `:137`/`:169`) and *is* advisory by design — recorded as such rather than as a defect.

## A4 — measured, but nothing reads the verdict

- **No consumer** (#6635 Arm A): stderr WARN (`cmd/grafel/index.go:3466-3475`) plus a JSON field (`sidecar_build.go:47` → `internal/graph/graph.go:586-588`) whose only reader in the tree is `sidecar_build_test.go:125`, which builds the sidecar directly rather than running an index.
- **No calibrated zero**: `docs/grammar-canary-baseline.json` is empty, so the relative test is unreachable (`canary.go:341`, `minBaselineNodes = 200` at `:172`) and only the 2 % absolute rule applies (`canary.go:161`).
- **The test arm cannot substitute**: assertions trip at `maxDominatedRatio = 0.50` (`parse_canary_test.go:76`, `:833`, `:894`) against a corpus whose own comment excludes new syntax *"so the canary stays a regression net, not a feature gate"* (`:82-84`).

## Corrections found while grounding

- The provenance note lives in `binding.notes`, **not** `_comment`, and the field it describes is `bundled_via` (the text's own "bundled_version" is stale). Cited by its real location.
- **"All 27 grammars are permanently stale" is not accurate.** By the dates in `grammars.lock`, **23 of 27** are behind the pin; `lua`, `proto`, `toml` and `yaml` are not, because their own upstreams stopped committing in 2021-2022. The doc says 23 of 27.
- The A4 ERROR-ratio threshold is asserted in **two** places (`:833` and `:894`), not one.
- #6636 item 3 (idempotency) is **already fixed on main** — `.github/workflows/grammar-freshness.yml:31-33` keys on a stable `ISSUE_MARKER`. Not mentioned in the doc.

## Also changed

`docs/new-language-feature-triage.md` repeated the same claim in its signal table; it now carries a short caveat linking to the new section. Both files' "maintained as of" dates refreshed to 2026-08-26.

Refs #6636, #6635.

### DIFF
--- a/docs/language-release-calendar.md
+++ b/docs/language-release-calendar.md
@@ -1,6 +1,6 @@
 # Language-release calendar (A3, epic #5359 — milestone 0.1.4)
 
-_Calendar maintained as of 2026-06-23. Companion to
+_Calendar maintained as of 2026-08-26. Companion to
 [`grammars.lock`](../grammars.lock) and
 [`docs/grammar-freshness-audit.md`](./grammar-freshness-audit.md)._
 
@@ -23,6 +23,12 @@ grammar has moved or that parse errors have spiked. This **calendar is the
 proactive trigger**: it fires *ahead* of a known release window so we verify
 grammar + extractors against version N around the day it lands, rather than
 waiting for an alarm to catch up.
+
+> **Read [What A2 and A4 cannot detect today](#what-a2-and-a4-cannot-detect-today)
+> before relying on either alarm as release coverage.** Neither one can currently
+> answer *"did the grammar for the language that just shipped fall behind?"* — for
+> structural reasons, not because of a bug. The per-release checklist below is
+> still the right procedure; what it cannot do is delegate step 1 to automation.
 
 ## How to use this on a release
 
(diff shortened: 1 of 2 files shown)


cajasmota:
Merging. This corrected my framing in four places, and I verified the one that mattered most rather than taking it on trust.

## "Every grammar is permanently stale" was wrong

I wrote that in my #6636 correction. Recomputed from `grammars.lock` directly: **23 of 27** are behind the pinned date, and four are not — `lua` (2022-08-14), `proto` (2021-06-12), `toml` and `yaml` (2021-05-11) — because their own upstreams stopped moving. The doc says 23 of 27, names the exceptions, and attributes the number to the lock rather than to a live query, which is the right hedge since the checker queries upstream at runtime.

That is a correction to a correction: I originally called the signal "saturated", then corrected the issue's fix items, and the saturation claim itself was still overstated. Worth noting because I had already been careful once on this issue and was still wrong.

The other three: the provenance sentence lives in `binding.notes`, not `_comment` (I misattributed it, and the note's own phrase "bundled_version below" is stale — the field is `bundled_via`); there are **two** `maxDominatedRatio` assertions at 0.50, not one; and #6636 item 3 is already fixed on `main` via the `ISSUE_MARKER` keying from #6646, so the doc correctly says nothing about it.

## What the doc now does

It states plainly, at each point where the calendar cites them, what A2 and A4 **can** and **cannot** detect, with the structural reason and the blocking work — and it keeps the per-release checklist intact while marking step 1 **manual**: read `parse_error_canary.languages[]` yourself, and use the A2 issue as a magnitude lookup rather than an alarm.

That is the right shape. The procedure was never the problem; the implied automation behind it was. A reader following the old text would have assumed two crons were watching for them.

The same claim was repeated in `docs/new-language-feature-triage.md`'s signal table and is fixed there too — worth having checked, since a corrected claim that survives in a sibling doc is how the original one propagated.

## One thing left for a separate look

`docs/grammar-freshness-audit.md`'s "Refreshing the baseline" section prescribes a manual refresh procedure for `grammar-canary-baseline.json` — a file **nothing has ever written**. Flagged rather than folded in. That is the same shape as #6635 (a mechanism cited as coverage that cannot provide it), one document over.

Issue #6288, opened 2026-08-11
Windows: homeguard lets every home path through, atomicfile asserts POSIX modes, and docgen writes into the real user profile


cajasmota:
Three packages fail on Windows at `fb4c6fd4b`. All three trace to the #6246 symlink/mode/home-resolution cluster, which landed **after** `v0.2.1` — the last commit ever verified on three platforms.

Run: https://github.com/cajasmota/grafel/actions/runs/31510723706 (`internal/daemon/watch` fails separately, tracked in #6287)

## 1. `internal/homeguard` — the guard does not recognise a Windows path

```
--- FAIL: TestEscapes/the_home_itself
    homeguard_test.go:36: Escapes("\home\dev", "\home\dev") = false, want true
--- FAIL: TestEscapes/a_dotfile_in_the_home
    homeguard_test.go:36: Escapes("\home\dev\.claude.json", "\home\dev") = false, want true
--- FAIL: TestEscapes/nested_under_the_home
    homeguard_test.go:36: Escapes("\home\dev\.cursor\mcp.json", "\home\dev") = false, want true
```

Added by `0642468e0` (#6246). This is the containment guard that decides whether a write escapes the user's home — the thing standing between us and the #5443 incident. It returns `false` (does not escape → allowed) for every one of these on Windows.

**Decide whether this is only the test's POSIX-shaped fixtures, or the guard itself.** If `Escapes` is separator-naive, the guard is not protecting Windows users at all, and a "the test used the wrong separator" conclusion would paper over a live hole in a security-relevant check. Establish which by evidence, not by whichever is cheaper to fix.

## 2. `internal/atomicfile` — mode assertion is POSIX-only

```
--- FAIL: TestExistingPerm_ReportsTheTargetNotTheLink
    writethrough_test.go:219: got 0666, want the target's 0600
```

Added by `c028a360d` (#6246). Windows Go reports `0666`/`0444` and nothing finer, so a `0600` expectation cannot hold there. Note the adjacent warning the same run emits, which is real behaviour worth a look while you are here:

```
WARN atomicfile: cleared the read-only attribute on a destination in order to
replace it; the file's read-only protection is now gone
```

The likely correct move is to assert the *property* (the target's mode is reported, not the link's) in a way Windows can express, or to skip with a stated reason — not to delete the assertion.

## 3. `internal/mcp` docgen — resolves the REAL home instead of the test's

```
--- FAIL: TestDocgenPromote_Atomic
[...]

Issue #2083, opened 2026-05-24
Tests leak fixture stores into real `~/.archigraph/store/` (~100+ directories)


cajasmota:
## Symptom
`ls ~/.archigraph/store/` shows ~100+ directories from unit tests that weren't sandboxed:
- `001-*` (a3 hashes, ~70 directories, each 12KB)
- `crossfile_go-*`, `crossfile_python-*` (~30 directories)
- `cdk_app-*`, `spring_app-*`, `django_app-*`
- `src-*`, `admin-*`, `cdc-*`, `cloudformation-*` (tiny fixture remnants)

These pollute the real user's archigraph state. They're harmless in size (~1.5MB combined) but:
1. `archigraph doctor` traverses them
2. Future cleanup operations have to filter them
3. Indicates tests are writing into `os.UserHomeDir()` instead of a sandbox

## Likely cause
Tests calling indexing code that respects `~/.archigraph/store/<repo-hash>/` but not setting an override (e.g. `XDG_STATE_HOME` or `ARCHIGRAPH_STATE_DIR`) for the test process.

## Expected fix
1. Add `ARCHIGRAPH_STATE_DIR` env override (if not already) honored everywhere paths are resolved.
2. Test helpers (e.g. `testutil.WithTempStateDir(t)`) that set it to `t.TempDir()`.
3. Sweep + delete the leaked directories.
4. Lint/test that fails if a test writes outside `t.TempDir()`.

## Related
- The daemon `archigraph cleanup` command exists but didn't catch these because they're not orphaned-registry entries — they were never registered, they're test artifacts that wrote raw graph.fb.

PR #6662, opened 2026-08-25
fix(#6645): 32 internal/mcp tests wrote into the real ~/.grafel — seven seeders, one seam, measured 32 → 0

### LINKED ISSUE #6645: 32 tests in internal/mcp write into a real grafel home — the isolation guard checks that isolation was called for, not that it happened
**32 tests in `internal/mcp` write into a real `HomeDir()` when run.** Measured, not estimated: every one of the package's **1280** top-level tests was run alone against a fresh surrogate `HOME`, and 33 of them left state behind. #6644 removed one (`TestElapsedMSCoverageAllTools`); the other 32 are pre-existing and untouched.

None are incidental — all 33 wrote under `.grafel`, specifically `store/*/refs/_unknown/graph.fb` and `events/*.jsonl`.

## The leaking rows

They cluster in the group-lifecycle tests rather than being scattered:

- `TestSweepToMemoryBudget_*` — 7
- `TestSweepIdleGroups_*` — 5
- `TestEvictGroup_*` — 5
- `TestSegmentSet_*` — 3
- `TestGroup_*` — 3
- one each: `TestFlowOverlay_*`, `TestContentHashSkip_*`, `TestDescriptionSidecar_*`, `TestFlowSidecar_*`, `TestSingleFile_*`, `TestMetaEventDispatch`

The clustering is the useful signal: this is one shared setup path missing isolation, not 32 independent mistakes. Fixing the shared helper is likely to move most of the count at once — which also means the fix must be **measured**, not assumed, because a helper change that looks right and moves 6 rows instead of 26 would pass any test that only checks a couple of them.

## Why it matters beyond tidiness

On a developer machine, running `go test ./internal/mcp/` writes into the developer's own grafel state — the same state the daemon and the MCP tools read. That is a test suite mutating the environment it is supposed to be observing.

`internal/mcp/home_isolation_guard_6288_test.go` exists to prevent exactly this, and its own comment says why it cannot (`:43-45`):

> "It is keyed on the PRESENCE of a `Setenv("HOME", …)` call and says nothing about a function that isolates nothing at all."

So the guard checks that isolation was *called for*, not that isolation *happened*. 32 rows sit in that gap. This is the milestone's recurring shape — a guard whose own comment documents the hole, and nothing observing the hole.

## What is needed

1. **Reproduce the count first.** The harness is described in PR #6644's body: run each top-level test alone against a fresh surrogate `HOME` and record which ones write under it. Confirm 32 on current `main` before changing anything. A fix not preceded by a measured baseline cannot be scored.
2. **Fix the shared setup path**, not 32 call sites, if the clustering holds. Say which helper and why.
3. **Re-measure and report the new count.** The positive control is the whole point: #6638 shipped with a reviewer believing one test polluted when six did, and only the before-count exposed it.
4. **Strengthen the guard so the gap closes structurally** — a check that observes *isolation happened* rather than *isolation was called for*. #6644 added a `TestMain` post-`m.Run()` pattern for a neighbouring property; the same shape may apply.

## Traps

1. Do **not** simply add `sandboxGrafelHome(t)` to 32 tests and call it done without re-measuring. Several of these construct their own servers or registries, and the leak may originate below the helper.
2. Do **not** widen into `internal/mcp`'s production code. Every finding here so far has been test-side; #6639 established that the production path (project-anchored staging) is correct and must not move.
3. Do not pin or disable shuffle ordering anywhere.
4. Test-only. No production file may change.

Found while fixing #6639. The count and the row list come from that PR's measurement harness. Refs #6639, #6644, #6288.

cajasmota:
Closes #6645

Test-only. `git show --stat` touches eight `*_test.go` files and nothing else. `internal/mcp`'s production code is unchanged, per #6639 — the project-anchored paths are by design.

## The measurement, before anything was changed

The #6644 harness, re-run on `c67b1539e`: every top-level test in the package compiled once and run **alone** in its own process against a **fresh surrogate `HOME`** (`HOME`/`USERPROFILE` redirected, `GRAFEL_HOME` unset, real `GOPATH`/`GOMODCACHE`/`GOCACHE` pinned so a surrogate home can never become one of them), then the surrogate walked for anything written under it.

| | tests writing under a real `HomeDir()` |
|---|---|
| **before** (`c67b1539e`) | **32** / 1284 |
| **after** | **0** / 1285 |

The 32 rows are exactly the ones the issue lists — same names, same clusters. The **total** differs from the issue's 1295: `main` moved between the two measurements. **32 is the number that reproduced**, so the issue body's count is good.

No test failed in either scan.

## The shared origin

The clustering held, and it is not one helper — it is one *derivation*, reached by seven setup paths.

`daemon.StateDirForRepo` resolves to `$GRAFEL_HOME`-or-`~/.grafel``/store/<slug>-<hash>/refs/<ref>/`. Six on-disk **seeders** called it with no sandbox in force and then wrote a fixture `graph.fb` there — which is why every leaking row landed under `store/*/refs/_unknown/`:

| seeder | leaking tests |
|---|---|
| `seedRepoOnDisk` (`reload_cost_3377_test.go`) | 12 — `TestEvictGroup_*`, `TestGroup_*` (2), `TestContentHashSkip_*`, `TestDescriptionSidecar_*`, `TestFlowSidecar_*` |
| `seedGroupsOnDisk` (`idle_group_evict_5872_pr2_test.go`) | 12 — all `TestSweepToMemoryBudget_*` and `TestSweepIdleGroups_*` |
| `writeSegmentSet` (`segment_serve_5912h_test.go`) | 3 `TestSegmentSet_*` |
| the inlined mmap seeder, duplicated in two `flows_overlay_test.go` tests | 2 `TestFlowOverlay_*` |
| `seedTwoGroups` (`group_lock_scope_6060_test.go`) | `TestGroup_ColdWakeDoesNotBlockOtherGroups` |
| the inlined single-file seeder in `TestSingleFile_ServedViaReaderForDirParity_5912h` | 1 |

The seventh is a different axis. `coreTestServer` builds its graph through `writeGraph`, which pins `GRAFEL_DAEMON_ROOT` (#2083) — so the **graph** was isolated and the **home** was not, and `TestMetaEventDispatch` drove `grafel_event` into `<real home>/.grafel/events/*.jsonl` on every run.

The last two rows in that table were **not** visible from the issue's clustering: they only surfaced when the first re-measurement came back **2**, not 0. Both are inline copies of a seeder rather than a call to one. They are now extracted (`newMmapFlowServer`, `writeSingleFileGraph`) so that they are seeders the guard can enumerate, which is the whole reason they were invisible.

## The fix: one seam, and it checks its own answer

`sandboxStateDirs(t)` (`testhelpers_test.go`) sandboxes the grafel home **once** for the calling test and returns the resolver every seeder now uses instead of a bare `daemon.StateDirForRepo`.

Three things about its shape are load-bearing:

- **It returns a closure, not a path.** The multi-repo seeders write several repos per call; a per-repo `sandboxGrafelHome` would hand each group a different `t.TempDir()` home and leave every earlier group unreadable once the last `t.Setenv` won.
- **It neutralises `GRAFEL_DAEMON_ROOT`.** That variable is consulted *before* `GRAFEL_HOME`, so a developer who exports it (a normal debugging move in this repo) gets a perfectly sandboxed home and a state dir outside it anyway.
- **It checks the resolved path.** Setting three variables is isolation being asked for; the path landing inside the sandbox is isolation happening. That distinction is the one #6288's guard documents about itself.

## The guard

`home_isolation_guard_6288_test.go` is unchanged. The new observation lives beside it, in `home_isolation_seeders_6645_test.go`, and it is the #6645 harness reduced to one in-process table: point `HOME`/`USERPROFILE`/`GRAFEL_HOME` at a surrogate "real" home, run a seeder, assert the surrogate is still **empty**. One row per seeder, named after the seeder — a failure names the setup path, not one of its 32 downstream tests.

The `coreTestServer` row **drives** `grafel_event` rather than just constructing the server. A row that only built the server would pass against a completely unisolated home, because the server is not what writes.

Two things it deliberately does not do:

- It does not diff the developer's actual `~/.grafel`. Over a ~2-minute suite that reports every write the running grafel daemon made during it — a guard that fails for reasons unrelated to this package. The surrogate makes it deterministic.
- It does not see a **new** test that reaches `daemon.StateDirForRepo` inline without going through the seam — the same blind spot #6288 has. **Nothing else sees it either** — see the withdrawal below.

This is a **regression pin on seven named seeders**, not structural closure. Adding a row is a manual step, and a seeder written without one is not caught.

This is a new file plus one call per seeder. No redesign of the existing guard.

## Mutants, scored

| mutant | before the new tests | after |
|---|---|---|
| **M1** revert `seedRepoOnDisk` to a bare `daemon.StateDirForRepo` | — | harness **0 → 12**, the exact dependents of that one seeder; `TestSeedersDoNotWriteIntoTheRealGrafelHome/seedRepoOnDisk` fails naming `.grafel/store/…/refs/_unknown/graph.fb` |
| **M2** delete the seam's `t.Setenv("GRAFEL_DAEMON_ROOT", "")` | **exit 0 — survives the whole table** | exit 1, `TestSandboxStateDirsNeutralisesDaemonRoot` |
| **M3** downgrade the seam's escape-check `t.Fatalf` to `t.Logf` | **exit 0 — SURVIVOR**, `go vet` 0, full package | exit 1, `TestSandboxStateDirsEscapeCheckIsFatal` |
| **M4** delete the seam's whole `if !strings.HasPrefix(dir, prefix) { … }` block | **exit 0 — SURVIVOR**, `go vet` 0, full package | exit 1, same test |
| **M5** weaken the seam's `strings.HasPrefix(dir, prefix)` to `strings.Contains(dir, ".grafel")` | **exit 0 — SURVIVOR** even against the first version of the probe | exit 1, same test, once the hostile root ends in `.grafel` |

M2 is the interesting one, and it is why there is a second test. Every row of the table enters through `surrogateHome`, which already clears `GRAFEL_DAEMON_ROOT` — so the table cannot see a seam that forgot to. Measured, not reasoned: the mutant was run and came back green before the second test existed. `TestSandboxStateDirsNeutralisesDaemonRoot` exports the variable and then asks the seam directly, and the kill comes **through the resolver's own escape check**, naming the offending path — which is what makes that check an observation rather than a comment.

### M3/M4 — the escape check was itself unobserved

The paragraph above claimed the escape check was "an observation rather than a comment". **It was not.** `TestSandboxStateDirsNeutralisesDaemonRoot` asserts the resolved **path**, so it passes whether the check *fails* the test or merely *reports*. Two independent reviews found the same survivor from opposite directions — downgrading the `t.Fatalf` to `t.Logf`, and deleting the guard block outright — and **both left `go vet` at 0 and the full package at exit 0**. A guard that logs instead of failing is this milestone's signature defect, and it was sitting inside the guard this PR adds.

`TestSandboxStateDirsEscapeCheckIsFatal` now observes the check rather than its answer. A check that fails the test cannot be watched from inside the same test — `t.Fatalf` marks the running test failed and a subtest's failure propagates — so the observation runs in a **child process**: the child asks the seam for its resolver and *only then* exports a hostile `GRAFEL_DAEMON_ROOT`, a state the seam cannot have neutralised in advance. The parent scores the child on its **exit code**, which is what kills both shapes at once: a `t.Logf` downgrade and a deleted block both exit 0 while handing back a path a seeder would write a real `graph.fb` into.

Mutants were restored by `cp` from a byte-level backup and verified with `shasum`, not by `git checkout`.

## Withdrawn: the seam is not a chokepoint

An earlier revision of this body claimed the seam is the only path from a test in this package to the daemon state dir, and that the seam's own escape check therefore covers a newly-written inline seeder.

**That claim was false and is withdrawn, not softened: 18 other `_test.go` files in `internal/mcp` call `daemon.StateDirForRepo{,Ref}` directly, the seam is not a chokepoint, its escape check only fires for code that went through it, and a newly-written seeder in the inline style is not caught — this PR is a regression pin on seven measured seeders, not the structural closure requirement 4 of #6645 asks for.**

Demonstrated rather than argued: a reviewer added one new seeder written exactly the way the seven leakers were, plus a single test calling it. `go vet` clean, full package **exit 0**, and **6 entries leaked into the surrogate home**. The leak count goes 0 → 1 and nothing notices. The inline pattern is used by 18 files — it is the normal way to write a seeder here, not an exotic corner.

`writeGraph` (`internal/mcp/server_test.go:25`) is the same story on the other axis. It *is* the split this PR fixes — it pins `GRAFEL_DAEMON_ROOT` only if unset and never sandboxes `HOME` — but it has ~100 callers, so the fix was applied one level above it, in `coreTestServer` only. No other caller leaks today (measured 0). The consequence is stated rather than left to be found: **the `events` axis is patched per-caller, not at its origin.** `writeGraph` is deliberately untouched here.

Structural closure — a chokepoint, or a vet-style check over those 18 call sites, plus `writeGraph` itself — is tracked in **#6663**.

## Verification

- `go test -count=1 ./internal/mcp/` — full package, no `-run` filter: **exit 0** (three runs, 126–157 s).
- `go vet ./internal/mcp/` clean; `gofmt -l internal/mcp/` clean.
- The after-count was re-taken on the final tree for all 32 previously-leaking rows plus the new tests: **0**.
- **M5** is why the hostile `GRAFEL_DAEMON_ROOT` in both probes is `filepath.Join(t.TempDir(), ".grafel")` and not a bare `t.TempDir()`. A bare temp dir has no `.grafel` segment, so a merely-*permissive* predicate keeps rejecting it and survives. The escape that actually costs a developer their store is one **into** a real `.grafel` — the guard means "inside **my** sandbox", not "inside **some** `.grafel`". Re-scored with the realistic root: `go vet` 0, full package **exit 1** (91.8 s), sole failure `TestSandboxStateDirsEscapeCheckIsFatal`. M3 (89.6 s) and M4 (97.8 s) still exit 1; pristine **exit 0** (90.9 s).
- M3 and M4 scored on the full package with no `-run` filter: **exit 0 before** the new test (99.5 s / 99.0 s), **exit 1 after** (102.2 s / 104.0 s), failing only `TestSandboxStateDirsEscapeCheckIsFatal`. Pristine tree back to **exit 0** (98.6 s).
- The package suite was run with `HOME`, `GRAFEL_HOME` and `GRAFEL_DAEMON_ROOT` all redirected to a scratch sandbox (real `GOCACHE`/`GOMODCACHE`/`GOPATH` pinned), so scoring these mutants wrote nothing into the developer's `~/.grafel`.

### DIFF
--- a/internal/mcp/core_merges_test.go
+++ b/internal/mcp/core_merges_test.go
@@ -21,6 +21,13 @@ import (
 // coreTestServer builds a one-group/one-repo server over the standard fixture.
 func coreTestServer(t *testing.T) *Server {
 	t.Helper()
+	// #6645: writeGraph pins GRAFEL_DAEMON_ROOT, so the GRAPH is isolated — but
+	// the grafel HOME is not, and the meta/event handlers this server exists to
+	// drive (grafel_event → feedback/persona) resolve their writers under
+	// registry.HomeDir(). TestMetaEventDispatch was therefore appending to the
+	// developer's real ~/.grafel/events/*.jsonl on every run. Sandbox the home
+	// FIRST so writeGraph's own t.TempDir() and the event writers agree.
+	sandboxGrafelHome(t)
 	dir := t.TempDir()
 	repo := filepath.Join(dir, "r1")
 	if err := os.MkdirAll(repo, 0o755); err != nil {
--- a/internal/mcp/flows_overlay_test.go
+++ b/internal/mcp/flows_overlay_test.go
@@ -99,6 +99,21 @@ func newFlowServer(t *testing.T) (*Server, string) {
 	if err != nil {
 		t.Fatal(err)
 	}
+	// The loader keeps a resident mmap Reader over graph.fb for the process
+	// lifetime (S8, #2159). On POSIX that file can still be unlinked while
+	// mapped, so leaking the handle costs nothing at teardown; on Windows the
+	// mapping LOCKS graph.fb and t.TempDir's RemoveAll fails with "Access is
+	// denied" (#4285).
+	//
+	// That never surfaced while this seeder wrote into the real ~/.grafel,
+	// which nothing cleans up. #6645 moved the write into a t.TempDir, so the
+	// leak now has somewhere to show — the isolation seam did not create this,
+	// it revealed it.
+	//
+	// Registered AFTER the t.TempDir() calls above on purpose: cleanups run
+	// LIFO, so registering last is what makes Close run FIRST, before either
+	// temp dir is removed.
+	t.Cleanup(srv.Close)
 	return srv, stateDir
 }
 
--- a/internal/mcp/flows_overlay_test.go
+++ b/internal/mcp/flows_overlay_test.go
@@ -226,13 +226,17 @@ func TestFlowOverlay_StaleFallback(t *testing.T) {
 	}
 }
 
-// TestFlowOverlay_MmapPath_Replace: the REPLACE merge also holds on the flag-ON
-// mmap read path (baked flow in graph.fb, cross-repo flow in the sidecar).
-func TestFlowOverlay_MmapPath_Replace(t *testing.T) {
-	forceServeFromMMap(t, true)
+// newMmapFlowServer (#6645) is the flag-ON mmap seeder shared by the two mmap
+// overlay tests. It was duplicated inline in both, and BOTH copies resolved the
+// state dir with a bare daemon.StateDirForRepo — writing a fixture graph.fb into
+// the developer's real ~/.grafel/store on every run. Extracting it makes the
+// sandbox seam a property of the seeder rather than of each copy of it.
+func newMmapFlowServer(t *testing.T) (*Server, string) {
+	t.Helper()
+	stateDirFor := sandboxStateDirs(t)
 	dir := t.TempDir()
 	repo := filepath.Join(dir, "r1")
-	stateDir := daemon.StateDirForRepo(repo)
+	stateDir := stateDirFor(repo)
 	if err := os.MkdirAll(stateDir, 0o755); err != nil {
 		t.Fatal(err)
 	}
@@ -248,6 +252,14 @@ func TestFlowOverlay_MmapPath_Replace(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
+	return srv, stateDir
+}
+
+// TestFlowOverlay_MmapPath_Replace: the REPLACE merge also holds on the flag-ON
+// mmap read path (baked flow in graph.fb, cross-repo flow in the sidecar).
+func TestFlowOverlay_MmapPath_Replace(t *testing.T) {
+	forceServeFromMMap(t, true)
+	srv, stateDir := newMmapFlowServer(t)
 	ents, rels := crossRepoFlowDelta()
 	if err := flows.Upsert(stateDir, ents, rels); err != nil {
 		t.Fatalf("upsert: %v", err)
@@ -268,23 +280,7 @@ func TestFlowOverlay_MmapPath_Replace(t *testing.T) {
 // key in the adjacency and this test fails.
 func TestFlowOverlay_StepAdjReplace_Mmap(t *testing.T) {
 	forceServeFromMMap(t, true)
-	dir := t.TempDir()
-	repo := filepath.Join(dir, "r1")
-	stateDir := daemon.StateDirForRepo(repo)
-	if err := os.MkdirAll(stateDir, 0o755); err != nil {
-		t.Fatal(err)
-	}
-	if err := fbwriter.WriteAtomic(filepath.Join(stateDir, "graph.fb"), bakedFlowDoc("r1")); err != nil {
-		t.Fatalf("write graph.fb: %v", err)
-	}
-	reg := Registry{Groups: map[string]RegistryGroup{"g": {Repos: map[string]RegistryRepo{"r1": {Path: repo}}}}}
-	regPath := filepath.Join(dir, "registry.json")
-	d, _ := json.MarshalIndent(reg, "", "  ")
-	_ = os.WriteFile(regPath, d, 0o644)
-	srv, err := NewServer(Config{RegistryPath: regPath})
-	if err != nil {
-		t.Fatal(err)
-	}
+	srv, stateDir := newMmapFlowServer(t)
 	ents, rels := crossRepoFlowDelta()
 	if err := flows.Upsert(stateDir, ents, rels); err != nil {
 		t.Fatalf("upsert: %v", err)
--- a/internal/mcp/flows_overlay_test.go
+++ b/internal/mcp/flows_overlay_test.go
@@ -99,20 +99,9 @@ func newFlowServer(t *testing.T) (*Server, string) {
 	if err != nil {
 		t.Fatal(err)
 	}
-	// The loader keeps a resident mmap Reader over graph.fb for the process
-	// lifetime (S8, #2159). On POSIX that file can still be unlinked while
-	// mapped, so leaking the handle costs nothing at teardown; on Windows the
-	// mapping LOCKS graph.fb and t.TempDir's RemoveAll fails with "Access is
-	// denied" (#4285).
-	//
-	// That never surfaced while this seeder wrote into the real ~/.grafel,
-	// which nothing cleans up. #6645 moved the write into a t.TempDir, so the
-	// leak now has somewhere to show — the isolation seam did not create this,
-	// it revealed it.
-	//
-	// Registered AFTER the t.TempDir() calls above on purpose: cleanups run
-	// LIFO, so registering last is what makes Close run FIRST, before either
-	// temp dir is removed.
+	// Release the server's graph state at teardown. This helper is not the one
+	// #6645's Windows failure came from — see newMmapFlowServer — but a Server
+	// that is never closed is a leaked mmap handle either way.
 	t.Cleanup(srv.Close)
 	return srv, stateDir
 }
@@ -267,6 +256,25 @@ func newMmapFlowServer(t *testing.T) (*Server, string) {
 	if err != nil {
 		t.Fatal(err)
 	}
+	// THIS is the helper the three Windows failures come from
+	// (TestFlowOverlay_MmapPath_Replace, TestFlowOverlay_StepAdjReplace_Mmap and
+	// the newMmapFlowServer row of TestSeedersDoNotWriteIntoTheRealGrafelHome).
+	//
+	// The loader keeps a resident mmap Reader over graph.fb for the process
+	// lifetime (S8, #2159). On POSIX that file can still be unlinked while
+	// mapped, so leaking the handle costs nothing at teardown; on Windows the
+	// mapping LOCKS graph.fb and t.TempDir's RemoveAll fails with
+	// "Access is denied" (#4285).
+	//
+	// It never surfaced while this seeder wrote into the real ~/.grafel, which
+	// nothing cleans up. #6645 moved the write into a t.TempDir, so the leak
+	// finally has a cleanup to obstruct — the isolation seam did not create it,
+	// it revealed it.
+	//
+	// Registered AFTER both t.TempDir() calls above (sandboxStateDirs' home and
+	// dir) on purpose: cleanups run LIFO, so registering last is what makes
+	// Close run FIRST, before either temp dir is removed.
+	t.Cleanup(srv.Close)
 	return srv, stateDir
 }
 
--- a/internal/mcp/group_lock_scope_6060_test.go
+++ b/internal/mcp/group_lock_scope_6060_test.go
@@ -27,7 +27,6 @@ import (
 	"testing"
 	"time"
 
-	"github.com/cajasmota/grafel/internal/daemon"
 	"github.com/cajasmota/grafel/internal/graph/fbwriter"
 )
 
@@ -244,6 +243,9 @@ func TestGroup_NoCallerObservesHalfRevivedGroup(t *testing.T) {
 // "b"), each with a real graph.fb on disk, both resident.
 func seedTwoGroups(t *testing.T) *State {
 	t.Helper()
+	// #6645: both groups must share ONE sandboxed home, so the seam is entered
+	// once outside the loop and its resolver used per repo.
+	stateDirFor := sandboxStateDirs(t)
 	reg := &Registry{Groups: map[string]RegistryGroup{}}
 	type seeded struct {
 		lr *LoadedRepo
@@ -251,7 +253,7 @@ func seedTwoGroups(t *testing.T) *State {
 	seeds := map[string]seeded{}
 	for _, name := range []string{"a", "b"} {
 		repoDir := t.TempDir()
-		stateDir := daemon.StateDirForRepo(repoDir)
+		stateDir := stateDirFor(repoDir)
 		if err := os.MkdirAll(stateDir, 0o755); err != nil {
 			t.Fatalf("mkdir state dir: %v", err)
 		}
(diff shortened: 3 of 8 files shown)


cajasmota:
One more round on `deb267a31`, and it is the last. The escape-check probe is right in shape and kills both shapes you and the reviewer scored — but it cannot see the **permissive** direction, and that is the direction this guard actually fails in.

## My mutant — SURVIVOR

```diff
-		if !strings.HasPrefix(dir, prefix) {
+		if !strings.Contains(dir, ".grafel") {
```

`go vet` **0**, full package `go test -count=1 ./internal/mcp/` with `HOME`+`GRAFEL_HOME`+`GRAFEL_DAEMON_ROOT` all sandboxed: **exit 0 — SURVIVES** (94.1s), `TestSandboxStateDirsEscapeCheckIsFatal` included.

## Why it survives, and why that matters more than the mutant

Both probes build their hostile root as a bare `t.TempDir()` (`:183` and `:233`). That path contains no `.grafel` segment, so the weakened predicate still rejects it and the check still fires. The probe passes for a reason unrelated to what it means to assert.

The realistic value of `GRAFEL_DAEMON_ROOT` is **the developer's actual `~/.grafel`** — that is what the daemon sets in production. Under this mutant, `dir` then contains `.grafel`, the check does not fire, and the seeder writes into the real store. That is precisely the leak #6645 exists to stop, and the PR's own evidence shows it is not hypothetical: the reviewer put 31 directories into the live `~/.grafel/store` scoring these mutants.

So the guard's meaning is "inside **my** sandbox", and the fixture only ever supplies a hostile root that looks nothing like a grafel home. It cannot distinguish that from "inside **some** `.grafel`".

## The fix, measured — one line, both call sites

```diff
-	hostile := t.TempDir()
+	hostile := filepath.Join(t.TempDir(), ".grafel")
```

| tree | vet | full package (sandboxed) |
|---|---|---|
| pristine seam + this change | 0 | **exit 0** (102.9s) — no regression |
| permissive mutant + this change | 0 | **exit 1** (97.0s), sole failure `TestSandboxStateDirsEscapeCheckIsFatal` |

The mutant dies through the probe you added, not through some incidental assertion. Apply it at both `:183` and `:233` — `TestSandboxStateDirsNeutralisesDaemonRoot` has the same blind spot as the child probe.

Then re-score **all three** shapes (`t.Logf`, block deleted, `HasPrefix`→`Contains`) and report the table. I restored my worktree by `cp` from a byte backup; `shasum` matches `deb267a31` and `git status --porcelain` is empty.

## While you are in the file

Two comment blocks now overclaim in the same way the body did before you fixed it. `:178-181` says deleting the `t.Setenv` "fails only here … which is what makes that check an observation rather than a comment", and `:221-224` says the exit-code score means "both shapes of the defect die". Both were true of the shapes then in evidence and are too strong now — there is a third shape, and until this change lands it survives. Say *which* mutants are scored rather than claiming the check is observed in general. That is the same correction you already made well in the body; it just has not reached these two comments.

Nothing else is outstanding. Finding 2's withdrawal sentence is the right shape — it states the bound plainly instead of softening it — and #6663 correctly carries the structural work.

cajasmota:
**Blocking, and it is not flaky.** `test (windows-latest)` fails on `dae6716c6`, and it failed identically on `deb267a31` and `6c311f93a` — all three heads, same three tests, same error. Four Windows-only *timing* failures are on this repo's record, so I checked before saying it: this is not one of them. It is reproducible and it is caused by this PR.

```
--- FAIL: TestFlowOverlay_MmapPath_Replace (1.73s)
    testing.go:1464: TempDir RemoveAll cleanup: unlinkat
    …\TestFlowOverlay_MmapPath_Replace792386671\001\.grafel\store\r1-0066115268565f59\refs\_unknown\graph.fb:
    Access is denied.
--- FAIL: TestFlowOverlay_StepAdjReplace_Mmap (1.90s)          … same shape
--- FAIL: TestSeedersDoNotWriteIntoTheRealGrafelHome/newMmapFlowServer (1.40s)  … same shape
```

## What this actually means — read this before fixing it

All three are **mmap** paths, and the failure is Windows refusing to unlink a file that is still mapped.

The PR did not create that defect. It **revealed** it. Before this change those `graph.fb` files were written into the real `~/.grafel`, which `t.TempDir()` never tries to clean up — so a leaked mmap handle had nowhere to make itself known. Now the writes land inside the temp dir, cleanup runs, and Windows says no.

So the honest reading is: **the PR is working exactly as intended, and the first thing it caught is a real handle leak that has been invisible on every platform for as long as the leak existed.** POSIX unlinks a mapped file happily, which is why ubuntu and macOS are green and always were. That is a good outcome, not a setback — but it cannot merge red.

Note the second-order point, because it is the more valuable one: this is direct evidence for #6663's argument. The isolation seam is not merely hygiene; routing test writes into a sandbox is what makes resource leaks *observable*. Add that to #6663 — it is a stronger argument for structural closure than anything currently in it.

## What to do

Fix the leak, not the symptom.

1. Find who maps `graph.fb` on the `newMmapFlowServer` path and ensure it is **unmapped/closed before the test ends** — a `t.Cleanup` that closes the server, registered so it runs *before* `t.TempDir()`'s own cleanup (Go runs cleanups LIFO, so the close must be registered **after** the TempDir call).
2. Do **not** work around it by `t.Skip`ping on Windows, by moving these three tests back out of the sandbox, or by ignoring `RemoveAll` errors. Any of those re-hides exactly what the PR exists to expose, and would make the next genuine Windows regression less believable — which is the running cost of the four we already have.
3. If the close cannot be made to happen before cleanup without a production change, **stop and tell me**. A production change is out of scope for a test-only PR and I would rather split it than let it grow.

## Verification I will hold you to

- You cannot reproduce this locally on macOS — POSIX will unlink the mapped file without complaint. **Do not claim a local green as evidence.** The proof is the Windows leg, so push and read CI.
- Re-run the identical commit only if you believe it flaky; I have already established it is not, across three heads.
- The three mutants scored in round 2 stay scored — do not re-derive them. My separate `prefix` survivor is recorded on #6663 and is deliberately not blocking.

Everything else about the PR is ready to merge: withdrawal of the false claim reads correctly, the comment blocks now name which mutants they score rather than claiming general observation, and the hostile-root change killed the permissive mutant I raised.

cajasmota:
**My grounding on the previous round was wrong, and it sent this investigation down two blind alleys.** Correcting it explicitly rather than letting the fix quietly land, because the mistake is more instructive than the bug.

I wrote:

> `newFlowServer` (`flows_overlay_test.go:83`) **is** the right helper — all three failing tests reach it. Your cleanup is on the correct function.

That is false. The three failing tests reach **`newMmapFlowServer`** (`:238`), which has no `Close`. Verified:

```
PR branch 5252d97dd:  83: func newFlowServer      238: func newMmapFlowServer
main      65c5bef10:  83: func newFlowServer      (no newMmapFlowServer)
```

**I grepped `main` for a helper that exists only on this branch.** `flows_overlay_test.go` is a file this PR modifies, and I checked the pre-change version of it — then stated the result as established fact, in bold, as the thing the agent should not re-derive.

The cost was real: with `newFlowServer` wrongly excluded as the cause, the only explanations left were exotic ones, and I supplied two — an undrained `retireHandle` borrow, and a reader opened outside `State`'s ownership. Both plausible, both wrong, both expensive to chase. The actual cause was a scripted replace matching the first of two byte-identical helper tails.

This is the same defect class this milestone keeps finding, committed by me while enforcing it: **a claim asserted with confidence that nothing had checked.** It is also exactly the failure I have been correcting in issue bodies for two days — grounding against a stale tree. Ten grounding passes this week caught someone else's stale line reference; this one was mine, and the tell was available (`git show <branch>:<file>` instead of a working-tree grep) and I did not use it.

Two things I got right and will keep: excluding the "nobody called Close" explanation for `newFlowServer` was itself sound and correctly reasoned from `State.Close`'s handling of `s.groups` and `s.evicted` — it just did not generalise to a helper I had not looked at. And the agent tested my hypotheses against call sites rather than deferring to them, which is what found the truth.

The fix on `5252d97dd` is the right one: `t.Cleanup(srv.Close)` on `newMmapFlowServer`, registered after both `t.TempDir()` calls so LIFO runs it first, no production change, and the `newFlowServer` cleanup kept as hygiene with a comment that no longer claims to be the Windows fix. The Windows leg is the only verdict that counts here — macOS cannot reproduce this, and the local green is a no-regression check, not evidence.

Issue #6659, opened 2026-08-25
TRACKING: pythonMaskInertRegions — scanner / writer-predicate / writer-range / consumer, and which issue is which


cajasmota:
`pythonMaskInertRegions` and its immediate neighbourhood have accumulated **thirteen** issues. Six are open. They are routinely mistaken for one another — three briefs this milestone had to be corrected because two of them were conflated — so this is the map, not new work.

**Do not implement anything here.** Each row is its own issue with its own decision.

## The three axes

The function divides into parts that fail independently, and every axis was found only *after* the previous one was pinned:

| axis | what it decides | issues |
|---|---|---|
| **Scanner** | which branch runs, how far it goes, where it stops | #6418, #6598, #6612/#6617, #6614, #6623, #6624, #6629, **#6637**, **#6648**, **#6653** |
| **Writer — predicate** | which bytes `blank()` preserves | #6649, **#6654** |
| **Writer — range** | which bytes the caller hands to `blank()` | **#6655** |
| **Consumer** | what the masked copy is used for | **#6652** |

Nine scanner issues landed before anyone scored the writer at all. That is the single most useful fact here: **a mutant in one axis is routinely invisible to every test written for another**, and the accumulated scanner tests turned out to cover the writer's filler and range only incidentally.

## Open, with the distinction that keeps getting lost

- **#6648** — the `#`-comment loop treats a lone `\r` as content. Python ends the line there, so we blank live code. **Less** masking if fixed. Comment branch only; the triple-quoted branch and `blank()` are already correct and must not be touched.
- **#6653** — the single-line literal scan, same byte, **different direction**: fixing it makes some *terminated* literals count as unterminated (**more** masking) while shortening unterminated spans (**less**). It moves the deliberate #6418/#6614 boundary, so it needs its own decision.
- **#6652** — `StartLine` counts `\n` only. Wrong **regardless** of how #6648 and #6653 resolve, because nothing about masking changes a `\n` count. In flight.
- **#6654** — `blank()`'s preserve set is pinned against *removals* only; adding `\t` survives. Also: the premise guard's region starts at byte 0, so a `\n` on a live code line satisfies it.
- **#6655** — the triple-quote blanked *range*: shifting its lower bound leaves the opening quote live.
- **#6637** — a backslash-escaped newline continues a literal; unobserved.

## Recorded equivalences — never re-score these as DEAD

Each was verified more than once and cost real time to re-derive:

- `:3192`'s single-line clamp.
[...]

PR #6667, opened 2026-08-25
fix(#6648): a lone \r ends a Python comment, and we blanked the live code after it

### LINKED ISSUE #6648: a lone carriage return ends a Python line but not our '#' comment, so we blank live code — a false negative

cajasmota:
`pythonMaskInertRegions`' `#`-comment loop ran to `\n` only. Under Python's universal newlines a bare carriage return **is** a line terminator, so the bytes after it are live code on the next line — and we blanked them. A real mount sitting after a lone CR never reached the scan.

That is a **false negative**: a mount that exists is not reported, which is invisible in a way a phantom endpoint is not. The fix moves in the direction of **less masking**, which can only stop discarding real code and can never invent an endpoint. That asymmetry is why this branch was decidable on its own while its neighbours are not.

## The change

One clause, in the comment branch only:

```go
for ; i < len(b) && b[i] != '\n' && b[i] != '\r'; i++ {
```

Plus a `LINE TERMINATORS (#6648)` block in the function's doc comment recording the decision for **all three** sites, since that comment is the source of truth the coverage docs cite.

**Deliberately untouched**, each separately tracked:

| site | why not here |
|---|---|
| triple-quoted branch | a lone `\r` inside a triple-quoted literal genuinely **is** content — already correct |
| `blank()` | preserving `\n` and `\r` is the contract pinned by #6649/#6650 — already correct (#6654 owns its preserve set) |
| triple-quote blanked range | #6655 |
| single-line literal scan | **#6653** — same byte, *mixed* direction: some **terminated** literals would stop counting as terminated (more masking) while unterminated spans shorten (less), moving the deliberate #6418/#6614 boundary |

## Mutant table

Full package `go test -count=1 ./internal/engine/`, no `-run` filter, `go vet` 0 throughout, scored on exit codes.

| family | mutant | before | after |
|---|---|---|---|
| **M1** | stop at `\r` only | DEAD | DEAD |
| **M2** | stop at neither | DEAD | DEAD |
| **M3** | CRLF-only — `\r\n` ends the comment, a lone `\r` stays content | **SURVIVOR** | **DEAD** |
| **M4** | `b[i] >= 0x20` — stop at any control byte, incl. a **tab** | **SURVIVOR** | **DEAD** |

M1 and M2 were predicted survivors and **were not**, recorded rather than quietly corrected: both are the **over**-masking direction, already dead under `TestSynth_FastAPI_MountedRouter_AbsentWhenUnresolvable_6414`. **M3 and M4 are the families this file kills at package level.**

### M4 — the permissive direction, found by review

M3 and M4 are the two **halves** of the claim, and they were not found together. M3 is *"a `\r` does end the comment"* — the false negative this issue reports. M4 is *"and nothing else does"*, the half the first round of this PR left entirely unobserved: the doc comment asserted the terminator set was exactly `{\n, \r}` while every fixture only exercised the first half.

`b[i] >= 0x20` reads naturally as "stop at a line terminator" and also stops at a **tab**:

```
'# TODO\tapp.include_router(markets.router, prefix="/ghost")\n'
 correct: fully blanked
 M4:      '      \tapp.include_router(...prefix="/ghost")\n'   →  /ghost is minted
```

That is **strictly worse** than the false negative being fixed here: a missing endpoint is a false negative, a minted one is a phantom the graph then asserts. It is the recurring shape #6659 records — *prose asserts what no test observes*, and the permissive direction is the one to mutate first.

The fix is fixture-only: one line added to `crFixture`, the CRLF control's `want` literal extended to match, and `C6` as a premise guard so the tab cannot silently vanish. **The production change is unchanged — `http_endpoint_synthesis.go` is byte-identical to the previous commit.** The kill runs through the test's **existing** direction check (no unexpected mount), making it the only clause here whose kill is an *unexpected* mount rather than a *missing* one.

Separately, the fix itself was scored as a mutant: reverting `&& b[i] != '\r'` fails the new test. That is the bug, reproduced as a failing assertion first and confirmed failing again after reverting.

## What the fixture asserts

A mount file with two mounts, because different families need different rows:

- `/network` sits after a **lone `\r`** inside a comment — masked away pristine and under M3, live under the fix.
- `/people` sits after a comment terminated by a plain `\n` — masked away under M1 and M2, whose loops run past the `\n` and eat the rest of the file.

Three layers of assertion: the comment **text** is blank (a *branch-ran* assertion — proof the comment branch executed, not merely that the input contains a `#`), the code after the `\r` survives masking byte for byte, and both mounts reach the synthesis pass end to end. Plus a direction check that no *third* mount was minted.

## CRLF control

A well-formed `\r\n` ends the comment under the old rule and the new one alike, so **a `\r\n` fixture observes nothing about this change**. It is carried as a control asserting the common Windows case **did not move**, pinning the masked buffer byte for byte. Measured: with the fixture degraded to CRLF, the pristine terminator and M3 both survive — which is exactly the control's point.

## Per-clause verdicts, with the family named

Guard-dropping cannot be run directly (dropping C1 leaves `h == -1` and the next line panics), so drops were measured with a claim-only probe carrying the degraded fixture and no premise guards, against each family in turn.

| fixture degradation | pristine | M1 | M2 | M3 | fixed |
|---|---|---|---|---|---|
| *(control)* the real fixture | kills | kills | kills | kills | PASS |
| C1 dropped: no `#` at all | SURV | SURV | SURV | SURV | PASS |
| C2 dropped: `#` inside a triple quote | kills | kills | kills | kills | **FAIL** |
| C3 dropped: CRLF instead of a lone `\r` | SURV | kills | kills | SURV | PASS |
| C4 dropped: nothing live after the `\r` | SURV | kills | kills | SURV | PASS |
| C5 dropped: no second `\n`-ended comment | kills | SURV | kills | kills | PASS |

- **C1** — load-bearing against **all four**. No `#`, no comment branch, terminator unobservable.
- **C2** — load-bearing against **none**. Its degradation admits no survivor; it makes the test fail on *correct* code. So it guards the assertion's own correctness — the same honest de-claim #6629's C4 made.
- **C3** — load-bearing against the **pristine terminator and M3 only**. M1/M2 die on a CRLF fixture too.
- **C4** — same verdict, same reason.
- **C5** — load-bearing against **M1 only**, and only for this test **in isolation**; at package level M1 was already dead via #6414.
- **C6** *(added after review)* — load-bearing against **M4 only**; no other family stops at a sub-0x20 byte. Dropping the tab from the `/ghost` line leaves M4 a survivor.

## One neighbour moved, and it was verified rather than assumed

The loop now halts **on** a `\r`, so `blank()` is never handed a carriage return from that call site again. #6649's two caller-A rows (`commentCRLF`, `commentLoneCR`) could previously observe `blank()`'s `\r` clause and no longer can; one fixture's marker had to move inside the comment, and both `kills` fields are **restated** rather than left overstating a kill they no longer make.

That is a real cost, so the claim was measured: with `blank()`'s `\r` clause removed, the package still **fails**, via `docstringCRLF` (caller B) and `unterminatedCR` (caller C). All three `blank()` families — `blankcr`, `blanknl`, `blankboth` — were re-scored after the fix and remain DEAD.

Review also caught that the replacement label was **literally false**: `"kills nothing since #6648"` overstated the de-claim, because both rows *do* fail under the inverted `blank()` predicate. Measured — under that mutant **all five** rows fail, so nothing there is load-bearing. The label and the prose above it now say that instead. De-claiming was right; the wording was not.

## Interference from the cluster, as predicted

#6659 warned that a fixture written for one axis gets intercepted by another. It happened in reverse here: the fix broke `#6649`'s caller-A premise guard, which had encoded the *old* comment-loop behaviour as its fixture shape. The guard fired loudly rather than silently mis-measuring — which is the guard working.

Closes #6648

---

### DIFF
--- a/internal/engine/http_endpoint_synthesis.go
+++ b/internal/engine/http_endpoint_synthesis.go
@@ -3153,6 +3153,27 @@ func fastapiResolveMountedRouter(arg string, imports map[string]pyImportBinding)
 //     mint false mounts from the inside of a broken string. This is the same
 //     direction as #6598, which stopped a runaway quote from unmasking the
 //     regions that FOLLOW it.
+//
+// LINE TERMINATORS (#6648). Python's universal newlines make a lone `\r` a
+// line terminator as much as `\n` is, and the three sites that care about it
+// are decided SEPARATELY because they are not the same question:
+//
+//   - the `#`-COMMENT loop stops at `\n` AND at `\r`. A bare carriage return
+//     ends the comment, so what follows it is live code on the next line; the
+//     old `\n`-only test blanked that code away and hid real mounts. Well-
+//     formed CRLF reads identically under either rule.
+//   - the TRIPLE-QUOTED branch has no terminator test at all, deliberately: a
+//     lone `\r` inside a triple-quoted literal genuinely IS content.
+//   - the SINGLE-LINE literal scan still breaks on `\n` only. Making it
+//     `\r`-aware pulls in two opposite directions at once — a TERMINATED
+//     literal whose closing quote sits after a `\r` would stop counting as
+//     terminated (MORE masking) while an UNTERMINATED literal's blanked tail
+//     would end sooner (LESS) — which moves the #6418/#6614 boundary above.
+//     That is a separate decision, tracked as #6653, not a shared tweak.
+//
+// blank() preserving both `\n` and `\r` (#6649/#6650) is what keeps byte
+// offsets in the result addressing the same source positions; CR must never be
+// "normalised" away here as a shortcut.
 func pythonMaskInertRegions(src string) string {
 	b := []byte(src)
 	blank := func(i int) {
@@ -3164,7 +3185,14 @@ func pythonMaskInertRegions(src string) string {
 		c := b[i]
 		switch {
 		case c == '#':
-			for ; i < len(b) && b[i] != '\n'; i++ {
+			// A lone `\r` ENDS the comment, exactly as `\n` does (#6648):
+			// under Python's universal newlines a bare carriage return is a
+			// line terminator, so the bytes after it are live code on the next
+			// line. Stopping only at `\n` blanked them — a false negative that
+			// hid a real mount. Well-formed CRLF is unaffected either way,
+			// since the `\r` is immediately followed by the `\n` the loop
+			// already stopped on.
+			for ; i < len(b) && b[i] != '\n' && b[i] != '\r'; i++ {
 				blank(i)
 			}
 		case c == '"' || c == '\'':
--- a/internal/engine/http_endpoint_synthesis_python_mask_6648_test.go
+++ b/internal/engine/http_endpoint_synthesis_python_mask_6648_test.go
@@ -1,0 +1,372 @@
+package engine
+
+import (
+	"sort"
+	"strings"
+	"testing"
+
+	"github.com/cajasmota/grafel/internal/types"
+)
+
+// #6648 — pythonMaskInertRegions' `#`-comment loop ran to `\n` ONLY, so a bare
+// carriage return was treated as ordinary comment content. Python disagrees:
+// under universal newlines a lone `\r` IS a line terminator, so the bytes after
+// it are live code on the NEXT line. We blanked them, and a real mount was
+// silently masked away before the scan ever saw it.
+//
+// That makes this a FALSE NEGATIVE, and the fix moves in the direction of LESS
+// masking — which can only stop discarding real code, never mint an endpoint
+// that does not exist. That asymmetry is why the comment branch is decidable on
+// its own while its neighbours are not.
+//
+// Scope, deliberately: the comment branch's terminator test and nothing else.
+// Explicitly untouched, each separately tracked —
+//
+//   - the TRIPLE-QUOTED branch: a lone `\r` inside a triple-quoted literal
+//     genuinely IS content, so having no terminator test there is already
+//     correct (#6655 owns its blanked range);
+//   - `blank()`: preserving `\n` AND `\r` is the contract that keeps byte
+//     offsets meaningful, pinned by #6649/#6650 (#6654 owns its preserve set);
+//   - #6653, the single-line literal scan. Same byte, MIXED direction: making
+//     it `\r`-aware would make some TERMINATED literals count as unterminated
+//     (more masking) while shortening unterminated spans (less), which moves
+//     the deliberate #6418/#6614 boundary. It needs its own decision.
+//
+// Reachability is not hypothetical: nothing on the Python path normalises CR.
+// Bytes are read raw and forwarded verbatim into the synthesis pass, so an
+// old-Mac-ending file, or a CR introduced mid-line by a generator or a bad
+// merge, hits this in production.
+//
+// ---------------------------------------------------------------------------
+// MUTANT FAMILIES, scored separately (`go vet` 0, then the FULL package with
+// `go test -count=1 ./internal/engine/`, scored on the exit code). A sweep
+// varies the INPUT against a FIXED mutant and says nothing about whether the
+// fixture pins the FAMILY — the lesson #6625 paid for — so the family is what
+// is varied here.
+//
+//	M1  stop at `\r` ONLY   `for ; i < len(b) && b[i] != '\r'; i++`
+//	M2  stop at NEITHER     `for ; i < len(b); i++`
+//	M3  CRLF-only           stop at `\r` only when it is followed by `\n`,
+//	                        i.e. `\r\n` ends the comment, a LONE `\r` stays
+//	                        content. On a lone CR this is behaviourally the
+//	                        PRISTINE code, which is why it is the family that
+//	                        matters here.
+//
+// Measured on 65c5bef10, `go vet` 0 throughout, scored on the exit code of the
+// FULL package (no `-run` filter):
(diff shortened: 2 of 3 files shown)


cajasmota:
I priced the M4 fix at ~2 lines and questioned the ~85 spent, most of it the **C6** premise guard. I tested that rather than accepting or trimming it, and **C6 earns its place — my estimate was wrong.**

## The test

I degraded the fixture the way it would realistically degrade: replaced the tab in `crFixture` with a space (both the fixture line and the `strings.Repeat` `want` literal, so the test stays self-consistent — exactly what a well-meaning later edit or a whitespace-normalising tool would produce).

`go vet` 0, full package `go test -count=1 ./internal/engine/` sandboxed: **exit 1**, sole failure `TestPythonMaskInertRegions_LoneCREndsComment_6648`, with:

> `#6648 premise broken: the fixture holds no sub-0x20 byte other than `\n` and `\r`, so a comment loop that stops at ANY control byte behaves exactly like the correct one here and SURVIVES. This test is VACUOUS for M4.`

That is the right failure. Without C6 the degraded fixture stays **green** while silently ceasing to pin M4 — the test would keep passing and keep claiming, in its own doc comment, that the terminator set is exactly `{\n, \r}`.

## Why my estimate was wrong

I priced the **kill** and forgot the **durability of the kill**. Those are different costs, and this cluster is where the difference shows: the dominant defect here is prose asserting what no test observes, and a fixture that quietly degrades converts a real test into that exact shape without anyone editing a claim. A one-character change to a string literal would have done it.

C6 is also the correct *kind* of guard — it asserts a property of the **input** (a sub-0x20 byte other than `\n`/`\r` is present) rather than restating the expected output, so it cannot pass for the wrong reason. Compare the several fixtures in this same function that failed by never executing the branch they claimed to test.

Recording the general form, since it generalises past this PR: **when a fixture's kill depends on one incidental byte, guard the byte.** The guard is not overhead on the kill; it is what makes the kill survive the next editor.

Withdrawing the ~2-line pricing. Nothing to trim.

Issue #6669, opened 2026-08-25
utoipa_axum: consume the #6668 cross-file registration marker and mint the canonical endpoint


cajasmota:
#6668 ships the PRODUCER half of the cross-file utoipa_axum case: a `routes!(h)` whose `#[utoipa::path]` lives in another module now emits an additive `SCOPE.Route` / `Subtype=utoipa_handler_registration` marker carrying a join key — `handler_module` + `handler_name`, resolved from the registration file's own `use` declarations — plus `mount_prefix` when a same-file `.nest(...)` supplies one.

Nothing consumes it yet. Filing that explicitly rather than leaving it implied, because this repo's precedent is that markers sit unredeemed: **#6414**, the consumer of #6385's FastAPI mount join key, has been open since it was filed, and #6668's own body says so.

## What the consumer has to do

At resolve/link time — where cross-file binding already exists — pair each marker with the handler's declaring module and mint (or enrich) the canonical `http_endpoint_definition`:

1. Find the handler entity for `(handler_module, handler_name)`. The Rust custom extractor already emits it as a `SCOPE.Operation` / `Subtype=endpoint` carrying `route_path` (`internal/custom/rust/utoipa.go`), which is the verb/path contract the registration file could not see.
2. Compose `mount_prefix` (when present) with that `route_path`, using the same join the mint uses (`rustJoinNestPrefix`).
3. Mint ONE definition per (marker, handler) pair, `framework=utoipa_axum`, and dedupe against any definition another producer already minted for that handler — including the **cross-file** `.route(...)` case, which is #6530 and is NOT fixed by #6668.

## The two things that will bite

- **#6530 is still open.** Two files minting one identical synthetic ID is the failure mode this whole arm has been steering around. A consumer that mints from the marker without a group-scoped identity decision will reproduce it, not avoid it. `TestUtoipaCrossFile_PathsAgreeStillDoubleMints_6668` pins the current (unfixed) behaviour and will need to move when it is fixed.
- **`handler_module` is a Rust module path, not a file path.** `crate::items` → `src/items.rs` is the common mapping but not the only one (`mod.rs`, `#[path = "..."]`, re-exports). Resolve by module identity where possible rather than by string-munging a path.

## Deliberately still out of scope

- path-qualified arguments, `routes!(crate::items::list)` — `utoipaRoutesMacroRe` fails the whole macro on those, which is a change to the MINT's recognition rule;
[...]

Issue #6656, opened 2026-08-25
the decision that OpenApiRouter::from does not start a router scope is prose — adding it to the ctor regex survives


cajasmota:
`rustRouterCtorRe` (added by #6560 B2a, `internal/engine/http_endpoint_axum.go`) deliberately does **not** count `OpenApiRouter::from(…)` as a router construction. The reasoning is written in-tree — `from` *converts* an existing router rather than starting a new one, so counting it would split a scope that is genuinely one — and **no test observes it**.

## The survivor

```diff
-	`\w*Router\b(?:::<[^>\n]*>)?::(?:new|default|with_openapi)\s*\(`
+	`\w*Router\b(?:::<[^>\n]*>)?::(?:new|default|with_openapi|from)\s*\(`
```

`go vet` **0**, full package `go test -count=1 ./internal/engine/` **exit 0 — SURVIVOR**, on `main` at `c67b1539e`.

Non-equivalent: a file with `OpenApiRouter::from(axum_router)` between a `.nest(…)` and a `routes!(…)` counts one extra construction under the mutant, trips the scope clause, and drops the prefix that pristine applies.

## Severity: low, and the direction is why

Counting `from` can only make the guard **stricter** — a miss costs a prefix and can never invent one. #6560's inventory work established that asymmetry deliberately, and a phantom endpoint is the failure this pass is most careful about. So nothing is silently wrong today.

What is wrong is that the decision is **prose**. If someone later adds `from` to the alternation — reasonably, since it *is* a constructor in the ordinary sense — nothing objects, and a legitimate prefix silently disappears from every file that uses the conversion form.

That is this milestone's most-repeated defect shape, and it is worth closing while B2a's context is still fresh.

## What is needed

1. A fixture where `OpenApiRouter::from(…)` sits between the nest and the `routes!` macro, asserting the prefix **is** applied — the mutant above must die.
2. **Score before and after**: exit 0 on pristine `main` (reproduce the survivor), exit 1 after.
3. **Premise-guard it** so the fixture genuinely exercises the scope clause — assert the endpoint is minted by this pass (`framework=utoipa_axum`, `source_handler=Controller:<h>`), which `requireUtoipaDef` already does, and that the `from` call really falls between the two offsets rather than outside the window.
[...]

Issue #6417, opened 2026-08-20
url_mount_point IDs are not file-qualified, so two files mounting the same prefix collapse into one entity whose SourceFile flaps under incremental indexing


cajasmota:
Found while reviewing #6415 (FastAPI `url_mount_point`). **Pre-existing in the Django path — not introduced by that PR**, which is why it is filed separately rather than blocking it.

The mount-point synthetic's ID is `http:ANY:<canonical>:mount`. It is **not file-qualified**.

So two different files in the same repo that mount at the same prefix produce the **same entity ID** with different `SourceFile` values:

```python
# app/main.py
app.include_router(users.router, prefix="/api")

# app/admin.py
admin_app.include_router(admin.router, prefix="/api")
```

Both yield `http:ANY:/api:mount`. The store collapses them into one entity, and **which file wins the surviving attribution is order-dependent**. Under incremental indexing the survivor can flap between files as one or the other is re-extracted.

Verified field-for-field that `django_urlconf_nested.go:135-158` has the identical property, so this is a shared characteristic of the mount-point shape rather than anything the FastAPI emitter added.

## Why this is probably minor, and the one case where it might not be

For the **linker**, this is close to harmless: `internal/links/http_pass.go:1047` harvests mounts into a per-repo prefix *set*, and `continue`s so a mount is never indexed as a producer. A set does not care that two sources collapsed into one member — `/api` is `/api` either way.

The exposure is in the **graph as a browsable artifact**: an entity whose `SourceFile` points at `main.py` when `admin.py` mounts the same prefix is simply wrong to a human reading it, and it flaps under incremental reindexing, which looks like churn with no cause.

## What to decide

Three options, none obviously right:

1. **File-qualify the ID** (`http:ANY:<canonical>:mount:<sourceFile>`). Correct attribution; produces N entities for N mount sites at the same prefix, which is arguably right — they *are* distinct declarations. Changes Django's IDs too, so it is a graph-shape change for existing users on reindex.
2. **Leave it and document** that a mount synthetic is a repo-level fact about a prefix, not a file-level declaration, and that its `SourceFile` is arbitrary among the mounting files. Cheapest, and honest — but "arbitrary" in a field users read as authoritative is a trap.
3. **Deduplicate deliberately**, picking a stable winner (lexicographically smallest path) so at least the flapping stops without changing cardinality.

[...]

PR #6670, opened 2026-08-25
feat(#6668): the cross-file utoipa registration emitted nothing; an additive marker with a join key, and the bound it does not cross

### LINKED ISSUE #6668: utoipa_axum B2b: cross-file handler resolution via an additive marker resolved at link time, not a group-wide mint
Final arm of #6560 (`@arthurgeron`). A, C, B1 and B2a are merged; this is the remainder, and it is **ours** — filed as its own issue so it can be dispatched as one bounded task rather than as "the rest of the epic".

The design decision was made and grounded on #6560 and is **not** reopened here: **marker-plus-linker, file-scoped mint.** Read that comment before starting.

## Defect

`utoipaRegisteredElsewhere` can only see the file it is scanning. When the `#[utoipa::path]` handler, the `routes!()` registration and a sibling `.route()` live in different files, the pass either mints nothing for the cross-file case or mints a second definition on a different path for the same handler:

```rust
// src/items.rs   #[utoipa::path(get, path = "/items")]  pub async fn list_items()
// src/router.rs  OpenApiRouter::new().routes(routes!(list_items))          → http:GET:/items
// src/main.rs    Router::new().route("/api/items", get(items::list_items))  → http:GET:/api/items
```

One handler, two definitions, two paths. When the paths **agree** it is worse: two files mint one identical synthetic ID — one ID, two claimants, which is **#6530, still open**.

## What to build

Follow the FastAPI mount-point precedent at `internal/engine/http_endpoint_synthesis.go:3286-3290`, which faced this exact choice and refused to go cross-file, for a reason written in-tree: a cross-file read "would not survive incremental indexing or the daemon fast path."

So: emit an **additive marker** carrying a join key resolved from **this file's imports only**, and let `ResolveHTTPEndpointHandlers` fold it at resolve time, where cross-file binding already exists. Apply the #6150 rule — when the corpus is not visible, leave the endpoint **unenriched**; never delete and never invent.

## What NOT to build

- **Not a group-wide attribute map.** `seen` for `dedupKey` is declared at `http_endpoint_synthesis.go:443`, *inside* a per-file detector pass, and is reconstructed per file. A group-wide dedupe is not a bigger version of that map; it is unimplementable in that layer.
- **Not an out-of-band pass** like the Django nested URLconf minter. Its cost is the argument against it: a pass called from `cmd/grafel/index.go:4434`, a pass-wide `seen` *plus* a per-file `scanned` guard (added by #6417 to fix an identity bug the pass itself caused), a per-file mount dedupe that deliberately does not consult the pass-wide map, and two post-hoc subtraction functions. Landing that on top of an open #6530 would be trading a missing endpoint for a corrupted identity.
- **Do not touch** `rustRouterCtorRe`, the pinned window-boundary fixture `TestUtoipaAxum_NestPrefixAtExactWindow`, or the documented tie-break — all settled in #6651. #6656 covers the `::from` question separately.

## Stated bound, to be repeated in the PR body

This does **not** produce a canonical `http_endpoint_definition` for the cross-file handler on day one. It is a correct partial, chosen over an incorrect whole.

That bound needs saying plainly because a marker is a promise, and this repo's precedent is that promises sit unredeemed: **#6414, the FastAPI prefix-folding follow-up, is still open.** If the marker ships without a linker consumer, say so in the body rather than implying the marker is the finished thing — and file the consumer as its own issue rather than leaving it implied.

## Verification

- The fixture must be genuinely multi-file; a single-file fixture cannot distinguish this fix from B2a's same-file composition, which is already merged.
- **Premise-guard it**: assert the endpoint is minted by this pass (`framework=utoipa_axum`, `source_handler=Controller:<h>`), as `requireUtoipaDef` already does.
- Score the **permissive** direction first. Every serious defect this milestone — four in a row in one seam today — came from a mutant making a predicate accept more, not less. A mutant that mints the cross-file definition *twice*, or folds a marker whose join key does not match, must die.
- Add the paths-agree case explicitly, and state whether it is fixed here or still lands on #6530. Do not leave that implied.

Refs #6560, #6530, #6150, #6417, #6414, #6651.

cajasmota:
Closes #6668. Final arm of #6560 (@arthurgeron).

> **Revision 4** — three rounds of independent review, eight real defects, all fixed and scored below. Round 2 **refuted a justification I had written into the tree**; round 3 found the same blind spot twice more. Every claim this body previously made that turned out to be wrong is corrected in place, not quietly dropped.

## The defect

`synthesizeUtoipaAxumRoutes` resolves a `routes!(h)` argument against an attribute map built from **this file only**. In the canonical utoipa_axum layout the handler and its `#[utoipa::path]` live in one module and the router in another, so `attrs[h]` misses and the registration is invisible to every consumer of the endpoint family.

## What this does

Emits an **additive** `SCOPE.Route` / `Subtype=utoipa_handler_registration` marker carrying a join key — `handler_module` + `handler_name`, resolved from **the registration file's own `use` declarations** — plus `mount_prefix` when a same-file `.nest(...)` supplies one.

This follows the FastAPI mount-point precedent (`fastapiMountPointSynthetics`), which faced this exact choice and refused to go cross-file, for a reason written in-tree: *"the router is constructed in a different file in the canonical FastAPI layout, and a cross-file read would not survive incremental indexing or the daemon fast path."* Nothing is folded, no path is invented, no existing entity changes shape.

The design ruling on #6560 is applied, not reopened: the group-wide attribute map is unimplementable in a layer whose `seen` map is reconstructed per file, and the Django-style out-of-band minter costs a pass-wide `seen` **plus** a per-file `scanned` guard added by #6417 to repair an identity bug that pass caused.

**Why the marker is not in the http_endpoint family.** `ResolveHTTPEndpointHandlers` rewrites every `Kind == http_endpoint` record to `http_endpoint_definition` unconditionally (only `http_endpoint_client_synthesis` escapes), so a pathless marker would become a phantom in the canonical family — the invention #6150 forbids. `SCOPE.Route` also appears in no `resolverKindEquivalents` key or value, so the marker can never be mistaken for a *handler*.

## The bound, stated plainly

**This does not produce a canonical `http_endpoint_definition` for the cross-file handler.** It is a correct partial chosen over an incorrect whole, and `TestUtoipaCrossFile_MarkerCarriesJoinKey_6668` asserts that (`countDefsForHandlerIn(create_item) == 0`) rather than implying it.

A marker is a promise, and this repo's promises have a track record of sitting unredeemed — **#6414**, the consumer of #6385's FastAPI mount join key, is still open. The consumer is **filed as #6669**, not implied.

**The marker is additive but NOT free.** It is one graph node per cross-file registration, it is counted in `internal/coverage/reachability.go`'s denominators, and `internal/links/reachability.go` + `internal/mcp/dead_code.go` seed `SCOPE.Route` as a framework entry point. The enrichment cost — which was the largest of these — is now removed rather than absorbed (see R3).

## Paths-agree verdict — NOT fixed here, still #6530

In the layout where `items.rs` mints `http:GET:/items` from its own attribute and `main.rs` mints the same ID from a literal `.route("/items", get(list_items))`, two files produce one identical synthetic ID with two claimants. That is #6530 and it is untouched: **neither file emits a marker**, so this PR neither fixes nor worsens it. `TestUtoipaCrossFile_PathsAgreeStillDoubleMints_6668` pins the count at 2 with a comment telling the next reader to move the pin when #6530 lands.

## Emission is refused when

- the handler has a **same-file** `#[utoipa::path]` contract — Arm A/B1 already mints a real definition, and a marker beside it is a second claimant;
- another producer reading this file registers it — `axumRouteRe` or `rocketRouteAttrRe`;
- no `use` declaration supplies a join key: absent, glob, or a **nested** group;
- two declarations bind one local name to a different **module** *or* a different **item** (see the collision policy below);
- the module path is malformed after the leaf split (`crate::`, `crate::1bad`);
- the module is `self::`- or `super::`-rooted — resolvable only against the declaring file's own position in the module tree, which a per-file pass cannot know.

Still unhandled: path-qualified `routes!(crate::items::list)` (a change to the mint's recognition rule), and cross-file `.nest(...)`.

Untouched, per #6651/#6656: `rustRouterCtorRe`, `TestUtoipaAxum_NestPrefixAtExactWindow`, the documented tie-break, and the `::from` question.

---

## Review round 1 — three permissive defects

All three were genuine, all permissive, all in code the previous body called guarded. `go vet` before every score; full-package `go test -count=1`, no `-run` filter.

### R1 — the #6150 "never guess a module" clause was unpinned

Deleting `if !bound { continue }`:

| state | vet | test | verdict |
|---|---|---|---|
| **before** (PR head `968baff61`, its own fixtures) | 0 | **0** | **SURVIVED** |
| **after** | 0 | **1** | **DEAD** — `UnimportedHandlerGetsNoMarker` |

Under the mutant every subcase emitted `utoipa:registration:src/router.rs:::` with an **empty** `handler_module`/`handler_name`, and the test passed anyway: `requireNoUtoipaMarkerFor` filters on `handler_name`, so an empty one is invisible to it, and unlike its siblings that test had no total-count backstop. This is the same assertion-helper defect I self-reported and fixed in `utoipaMarkers` last round — present in a second helper. Fixed by adding `if n := len(utoipaMarkers(ents)); n != 0`.

### R2 — the deleted nested-group guard fabricated a join key

Last round's deletion rested on the claim that leaf validation subsumes the guard. **That claim is false.** On PR-head code:

```rust
use crate::{items::{create_item, purge}, admin};   routes!(purge)
  → handler_module=crate  handler_name=purge      (real path: crate::items::purge)
```

The scan cut the group at the **first** `}`, splitting out a bare `purge` that passes leaf validation and is attributed to the **outer** base. That is an invented key, and because #6669 resolves on `(handler_module, handler_name)` it **mis-joins** rather than missing — worse than a phantom. Compounding it, the subcase labelled `nested-group` contained no nested braces, so the fixture never exercised the case its label named.

| state | vet | test | verdict |
|---|---|---|---|
| **before** (PR head + new fixture) | 0 | **1** | fabricates `crate::purge` |
| **after**, guard deleted again | 0 | **1** | **DEAD** — `NestedUseGroupFabricatesNothing` |
| **after**, unmutated | 0 | **0** | green |

Both candidate guards (reject on body braces / scan to the last `}`) each fix the case alone, which left each individually unkillable. **Exactly one is kept and pinned.**

I was wrong to delete this on a subsumption argument that was asserted rather than demonstrated. Deleting an unkillable guard is right only when the subsumption is shown.

### R3 — the marker was a critical-band enrichment candidate

`qualifyHTTPKinds` contains `"SCOPE.Route"` and is consulted **before any structural signal**; `EnrichmentRequired: false` is never reached on that path. A record shaped exactly as emitted scored `qualifies=true, signals=[http_endpoint], score=80, band=critical` — real model budget spent describing a join-key stub named `utoipa:registration:src/router.rs:crate::items::create_item`.

Excluded by **Subtype**, not Kind, in `qualifiesForEnrichment`. `TestQualifies_UtoipaRegistrationMarker_NotACandidate_6668` asserts the exclusion, asserts a plain `SCOPE.Route` **still qualifies** (so the exclusion cannot widen unnoticed), and re-checks through the full `CollectCandidates` path.

| mutant | vet | test | verdict |
|---|---|---|---|
| drop the exclusion (= previous behaviour) | 0 | **1** | **DEAD** |

"Additive and inert" no longer stands unqualified; the residual costs are listed under the bound above.

### Open items — ruled explicitly

- **Path-shaped group leaves now resolve.** `use crate::{items::create_item, admin::purge};` is the most common rustfmt grouping and previously resolved **neither** leaf. Restrictive rather than wrong, but it would have shipped the feature near-inert on real routers. Mutant `strings.LastIndex → strings.Index` (take the first segment) is **DEAD** via a multi-segment leaf `api::v1::admin::purge`.
- **`self::` / `super::` are refused** (#6150) — the key would be a guess in the shape of an answer. Mutant removing the check: **DEAD**.
- ~~**The `dup` first-wins guard in `rustAddUseLeaf` is deleted.**~~ **Withdrawn in round 2 — the E0252 premise was false.** See the collision policy above.
- **The `//`-commented `routes!` asymmetry is KEPT and pinned.** A line-commented `use` binds nothing, but a line-commented `routes!` is still read as a registration. `utoipaAttrIsLineCommented` could suppress it — but `synthesizeUtoipaAxumRoutes` and `synthesizeAxumRoutes` both read a commented registration as real. A marker stricter than the mint that shares its regex would make one `routes!` occurrence a registration for one pass and not the other, which is worse than the false positive it removes. Fix it for the whole Rust route family or not at all. `TestUtoipaCrossFile_CommentedRoutesMacroStillMarks_6668` pins it as-is, labelled as documentation rather than endorsement.
- **M11's equivalence is caveated in-tree.** Widening `rustUseIdentRe` to admit `*` is equivalent **only under the current mint recognition rule** — a `*` binding is unreachable because `utoipaRoutesMacroRe` admits only `[A-Za-z_]\w*`. If that rule widens to path-qualified macros the equivalence lapses, so the two must widen together or not at all.

---

## Review round 2 — a false language claim, an unpinned guard, a silent gap

`go vet` before every score; full-package `go test -count=1`, no `-run` filter.

### The collision policy, and why the previous revision had none

The deleted first-wins guard was justified in-tree with: two `use` declarations binding one local name is E0252, does not compile, therefore unkillable. **E0252 is per-scope, and that claim is false.** Both of these compile, and both reached the binding map:

```rust
use crate::items::create_item;
mod tests { use crate::mocks::create_item; }        // different scope

#[cfg(feature = "x")]      use crate::real::create_item;
#[cfg(not(feature = "x"))] use crate::stub::create_item;
```

The map was **last-wins**, so a `mod tests` block — which by convention sits at the bottom of a file — always beat the top-level declaration. PR head published `handler_module=crate::mocks` and `crate::stub`, which **mis-joins** at #6669 rather than missing: the failure this body itself calls worse than a phantom.

**Policy chosen: DROP BOTH and poison the name for the rest of the file. Not first-wins.** First-wins would return the common cases to `crate::items`, but it still encodes "the first declaration in byte order is the one the macro meant" — true by convention, not by the language. This pass models neither scopes nor `cfg` evaluation, so it cannot know which binding is live at the registration site, and #6150's rule for exactly that situation is to leave the endpoint **unenriched** rather than guess. It is also what every other unresolvable shape here already does (glob, nested group, `self::`/`super::`), so the file has one answer to ambiguity instead of two. An *identical* duplicate binding names one item, is not ambiguous, and is kept.

| state / mutant | vet | test | verdict |
|---|---|---|---|
| **before** (PR head `bedd339b1` + new fixtures) | 0 | **1** | publishes `crate::mocks`, `crate::stub` |
| **M-C** — delete the policy (= last-wins, PR-head behaviour) | 0 | **1** | **DEAD** |
| **M-D** — first-wins instead of drop | 0 | **1** | **DEAD** |
| **M-F** — poison set never blocks a later resurrection | 0 | **1** | **DEAD** |

**This is the second guard I deleted on a subsumption that was asserted rather than demonstrated** (R2 was the first). Standing rule adopted for this PR and after: *a guard may only be deleted after a genuine attempt to construct a compiling counter-example, and that attempt is reported.* "No valid input can reach this" is a claim about a language, and language claims get checked against the language.

### M-B — `rustUsePathRe` was the only defence for the new concatenation, and nothing named it

`if module == "" || !rustUsePathRe.MatchString(module)` → `if module == ""`:

| state | vet | test | verdict |
|---|---|---|---|
| **before** (`bedd339b1` + its own fixtures) | 0 | **0** | **SURVIVED** |
| **after** | 0 | **1** | **DEAD** — `MalformedUsePathRefused` |

Not equivalent — under the mutant each input publishes a garbage join key: `use crate::{::create_item};` → `handler_module="crate::"`, `{a::::create_item}` → `"crate::a::"`, `{1bad::create_item}` → `"crate::1bad"`. All three are named in the fixture.

### Path leaf + alias resolved to nothing

`use crate::{items::create_item as mk};` produced **no marker**: `rustUseLeafAsRe` accepted only a bare identifier on its left, so it ran *before* the `::` split and could not match, and the split then left `create_item as mk` as the item name. Restrictive, therefore invisible — the direction this PR's own rationale argues against.

The alias is now split off **first** and its left side may be a path, and the condition is spelled `if !aliased` rather than `if local == leaf`.

*Corrected in round 3:* the previous body called the old condition "a case that could not reach it". That was true only of the **old identifier-only regex**. Under the new regex the two predicates agree pointwise — `local` is reassigned only inside the alias branch, to an identifier, and an aliased leaf always contains `" as "` — so `if local == leaf` would still be correct today. The rewrite is a readability change, not a defect fix.

| state / mutant | vet | test | verdict |
|---|---|---|---|
| **before** (PR head + new fixture) | 0 | **1** | 0 markers for an aliased path leaf |
| **M-E** — alias regex back to identifier-only | 0 | **1** | **DEAD** |

### Coverage — taken, not just disclosed

`internal/coverage/reachability.go` counted every marker as production surface, and `isTestEntity` filters only `subtype == test_suite`, so each marker was production-and-forever-untested: reported coverage fell by one per cross-file registration on every utoipa repo. Round 1 fixed the enrichment half of this and only *disclosed* the coverage half; it is now fixed with the same two-line Subtype exclusion, scored **both** directions.

| mutant | vet | test | verdict |
|---|---|---|---|
| **M-G** — drop the exclusion (= previous behaviour) | 0 | **1** | **DEAD** |
| **M-H** — widen it to the whole `SCOPE.Route` kind | 0 | **1** | **DEAD** |

The two remaining consumers are confirmed benign: the marker seeds BFS as its own entry, has no outbound edges, and propagates reachability to nothing.

---

## Review round 3 — one guard pinned on one axis, blind on the other. Twice.

Both survivors were the same shape: a predicate covered thoroughly along one axis and not at all along the other. Naming it, because it is the shape that let nine scanner issues land in `pythonMaskInertRegions` before anyone scored the writer.

### X1 — the ambiguity fixtures pinned the module axis and left the item axis open

Dropping the `prev.name == name` conjunct from the identical-duplicate test:

| state | vet | test | verdict |
|---|---|---|---|
| **before** (`6e9530a1e` + its own fixtures) | 0 | **0** | **SURVIVED** |
| **after** | 0 | **1** | **DEAD** |

Every ambiguity fixture from rounds 2 and 3 varies the **module** (`crate::items`/`crate::mocks`, `crate::real`/`crate::stub`). **None varied the item under a shared module** — a shape reachable only through the alias path, which is the newest code in the delta:

```rust
use crate::m::f as h;  use crate::m::g as h;
//  unmutated → no marker          (ambiguous, poisoned — correct)
//  X1        → h = crate::m::f    (first-wins, fabricated pair → mis-joins at #6669)
```

Two subcases added, including the realistic spelling — a `cfg` pair selecting between two handler versions in one module. X1 now dies on exactly those two, by name. The **complementary** mutant (dropping the *module* conjunct instead) is also **DEAD**, so both conjuncts are individually pinned rather than one covering for the other.

### C1 — the coverage transplant was pinned on the kind axis, not the subtype axis

`coverageMarkerSubtypes[e.Subtype]` → `e.Subtype != ""`:

| state | vet | test | verdict |
|---|---|---|---|
| **before** (`6e9530a1e`) | 0 | **0** | **SURVIVED** |
| **after** | 0 | **1** | **DEAD** |

Round 2's test covered widening along the **Kind** axis (M-H, dead) and the empty-subtype control, but not the **Subtype** axis — the one the guard actually keys on, and the one whose widening silently removes real entities from the coverage denominator where nothing would notice. Three production entities carrying other subtypes (`async_handler`, `method`, `http_route`) are now asserted to stay production.

### The overstated comment, corrected — and #6675 filed

The header claimed the file now has **"one answer to ambiguity"**. That is false, and it is corrected in place rather than dropped. The poison check is consulted **after** the module-validity, `self::`/`super::` and glob/nested-group rejections, so a competitor of one of those shapes returns early, never poisons, and the survivor is published as if unambiguous:

```rust
#[cfg(a)] use crate::real::create_item;
#[cfg(b)] use super::stub::create_item;   // rejected upstream → does not poison
//   → handler_module = crate::real, a guess at which cfg arm is live
```

**Not a regression** — every revision of this file has behaved so — and closing it means poisoning at every rejection point, which is a restructure of `rustAddUseLeaf`/`rustUseBindings`. Filed as **#6675** with both witness shapes rather than expanded into this PR at round 4 of a contributor's final arm. The header now states the limit and cites it.

### Disclosure

There is no `rustc` in this environment. Every claim in rounds 2 and 3 that a witness shape *compiles* is stated **from construction, not from a compiler run**. After the E0252 episode an unqualified "this compiles" is not something this PR asserts.

---

## Verification

Every fixture is **genuinely multi-file** (2–4 files) and **premise-guarded** by `requireUtoipaDefIn`, which asserts `framework=utoipa_axum` + `source_handler=Controller:list_items` on a definition minted by the pass under test.

Round-1 mutants, re-derived after the fixes:

| # | Mutant (permissive unless noted) | vet | test | Verdict |
|---|---|---|---|---|
| M2 | `attrs` always empty — same-file contract never suppresses | 0 | 1 | DEAD |
| M3 | `registeredElsewhere` always empty | 0 | 1 | DEAD |
| M4 | under `as`, join key binds the LOCAL alias | 0 | 1 | DEAD |
| M5 | drop the `(?m)^[ \t]*` anchor — commented `use` binds | 0 | 1 | DEAD |
| M6 | stamp `mount_prefix` unconditionally, including empty | 0 | 1 | DEAD |
| M8 | drop the ID dedupe — marker emitted **twice** | 0 | 1 | DEAD |
| M10 | drop `relPath` from the marker ID — two routers collapse | 0 | 1 | DEAD |
| M9 | *control (restrictive)*: emit nothing | 0 | 1 | DEAD (4 tests) |
| M11 | `rustUseIdentRe` admits `*` | 0 | 0 | EQUIVALENT — caveated above |
| M1 | drop the function-local `emitted` map | 0 | 0 | SURVIVED → deleted as dead |
| M7 | drop the nested-group guard | 0 | 0 | **SURVIVED for the wrong reason — see R2; guard RESTORED** |
| *(round 1)* | delete the `dup` first-wins guard | 0 | 0 | **SURVIVED for the wrong reason — see round 2; guard RESTORED as a drop policy** |

Round-2 mutants: **R1, R2, R4 (`self`/`super`), R5 (leaf segment), R6 (enrichment) — all DEAD**, each with `go vet` 0 and full-package exit 1.

Round-2 mutants: **M-B, M-C, M-D, M-E, M-F, M-G, M-H — all DEAD**, each `go vet` 0 and full-package exit 1.

Round-3 mutants: **X1** (item axis), **X1b** (module axis, complementary), **C1** (subtype axis) — **all DEAD**, each `go vet` 0 and full-package exit 1. X2, X3, X5 and C2 were confirmed dead by the reviewer.

`gofmt -l internal/engine/ internal/enrichment/ internal/coverage/` clean; `go vet` 0 on all three; `go test -count=1` **0** for `./internal/engine/`, `./internal/enrichment/` and `./internal/coverage/`.

## One correction to #6668's body

The FastAPI precedent is cited as `internal/engine/http_endpoint_synthesis.go:3286-3290`. On `main` (`65c5bef10`) that range is inside `pythonLineOfOffset`'s doc comment; the `fastapiMountPointSynthetics` header and the "would not survive incremental indexing" sentence are at **:3329-3341**. Everything else verified exactly, including `seen` at `:443`.

Refs #6560, #6530, #6150, #6414, #6417, #6651, #6656, #6669.

### DIFF
--- a/internal/coverage/reachability.go
+++ b/internal/coverage/reachability.go
@@ -113,6 +113,17 @@ func isProductionEntity(e types.EntityRecord) bool {
 	if isTestEntity(e) {
 		return false
 	}
+	// #6668 — additive JOIN-KEY MARKERS are not executable surface. The utoipa
+	// cross-file registration marker is emitted as a SCOPE.Route (it must stay
+	// out of the http_endpoint family, or the #1217 legacy-kind migration would
+	// turn a pathless record into a phantom definition), but it has no verb, no
+	// path and no body: no test path can reach it, so counting it as production
+	// makes it permanently UNTESTED and drops reported coverage by one for every
+	// cross-file registration in the repo. Excluded by SUBTYPE, so every genuine
+	// SCOPE.Route is unaffected.
+	if coverageMarkerSubtypes[e.Subtype] {
+		return false
+	}
 	switch types.EntityKind(e.Kind) {
 	case // executable / addressable surface — production
 		types.EntityKindFunction,
@@ -156,6 +167,15 @@ func isTestEntity(e types.EntityRecord) bool {
 }
 
 const subtypeTestSuite = "test_suite"
+
+// coverageMarkerSubtypes lists Subtypes of additive join-key markers that share
+// a kind with executable surface but are not reachable by any test path. The
+// value is duplicated from engine.utoipaCrossFileSubtype rather than imported —
+// internal/coverage must not depend on internal/engine — and is pinned on both
+// sides by TestIsProductionEntity_UtoipaRegistrationMarker_6668.
+var coverageMarkerSubtypes = map[string]bool{
+	"utoipa_handler_registration": true, // #6668
+}
 
 // entitiesGraph indexes a batch of entities for traversal: id→entity and the
 // adjacency list restricted to reachability edge kinds.
--- a/internal/coverage/reachability_test.go
+++ b/internal/coverage/reachability_test.go
@@ -211,3 +211,33 @@ func TestCrossSignal(t *testing.T) {
 		}
 	}
 }
+
+// TestIsProductionEntity_UtoipaRegistrationMarker_6668 pins the coverage cost of
+// #6668's cross-file join-key marker.
+//
+// The marker is emitted as a SCOPE.Route by deliberate design — it must stay out
+// of the http_endpoint family so the #1217 legacy-kind migration cannot turn a
+// pathless record into a phantom definition. But isProductionEntity counts every
+// SCOPE.Route, and isTestEntity filters only subtype == test_suite, so each
+// marker was counted as production-and-forever-untested: reported coverage fell
+// by one per cross-file registration on every utoipa repo.
+//
+// The exclusion keys on Subtype, so a genuine SCOPE.Route stays production. That
+// half is asserted too, or the exclusion could widen to every route unnoticed.
+func TestIsProductionEntity_UtoipaRegistrationMarker_6668(t *testing.T) {
+	marker := types.EntityRecord{
+		ID:      "utoipa:registration:src/router.rs:crate::items::create_item",
+		Name:    "utoipa:registration:src/router.rs:crate::items::create_item",
+		Kind:    string(types.EntityKindRoute),
+		Subtype: "utoipa_handler_registration",
+	}
+	if isProductionEntity(marker) {
+		t.Error("#6668: registration marker counts as production surface; no test path can reach it, " +
+			"so it is permanently untested and depresses reported coverage")
+	}
+
+	route := types.EntityRecord{ID: "r1", Name: "/api/users", Kind: string(types.EntityKindRoute)}
+	if !isProductionEntity(route) {
+		t.Error("#6668: a plain SCOPE.Route is no longer production; the exclusion must key on Subtype, not Kind")
+	}
+}
--- a/internal/coverage/reachability_test.go
+++ b/internal/coverage/reachability_test.go
@@ -240,4 +240,20 @@ func TestIsProductionEntity_UtoipaRegistrationMarker_6668(t *testing.T) {
 	if !isProductionEntity(route) {
 		t.Error("#6668: a plain SCOPE.Route is no longer production; the exclusion must key on Subtype, not Kind")
 	}
+
+	// THE SUBTYPE AXIS — the axis the guard actually keys on, and the one that
+	// matters most here: widening it silently DELETES real entities from the
+	// coverage denominator, which nothing else would notice. Without this
+	// assertion, `coverageMarkerSubtypes[e.Subtype]` → `e.Subtype != ""` scores
+	// vet 0 / exit 0 — SURVIVED.
+	for _, e := range []types.EntityRecord{
+		{ID: "f1", Name: "handleOrder", Kind: string(types.EntityKindFunction), Subtype: "async_handler"},
+		{ID: "o1", Name: "Svc.Do", Kind: string(types.EntityKindOperation), Subtype: "method"},
+		{ID: "r2", Name: "/api/orders", Kind: string(types.EntityKindRoute), Subtype: "http_route"},
+	} {
+		if !isProductionEntity(e) {
+			t.Errorf("#6668: %s (kind=%s subtype=%q) is no longer production; the exclusion must name "+
+				"marker subtypes exactly, not every entity that carries a subtype", e.Name, e.Kind, e.Subtype)
+		}
+	}
 }
--- a/internal/engine/http_endpoint_synthesis.go
+++ b/internal/engine/http_endpoint_synthesis.go
@@ -441,16 +441,22 @@ func applyHTTPEndpointSynthesis(args DetectorPassArgs) DetectorPassResult {
 	// regex (e.g. Spring's @GetMapping pattern). We only want one
 	// synthetic per endpoint per file.
 	seen := map[string]bool{}
-	// emitMountPoint appends an additive `url_mount_point` synthetic (#6385).
-	// These are not routable endpoints — the linker harvests them purely for
-	// their prefix — so they are deduped on the mount ID alone.
-	emitMountPoint := func(e types.EntityRecord) {
+	// emitAdditiveSynthetic appends a fully-formed ADDITIVE record, deduped on
+	// its own ID alone. Additive records are not routable endpoints and carry no
+	// (verb, path) identity to key on — the mount ID / registration ID IS the
+	// identity — so they share this one-line contract rather than makeEmit's
+	// side-scoped dedupKey.
+	emitAdditiveSynthetic := func(e types.EntityRecord) {
 		if seen[e.ID] {
 			return
 		}
 		seen[e.ID] = true
 		entities = append(entities, e)
 	}
+	// emitMountPoint appends an additive `url_mount_point` synthetic (#6385).
+	// These are not routable endpoints — the linker harvests them purely for
+	// their prefix — so they are deduped on the mount ID alone.
+	emitMountPoint := emitAdditiveSynthetic
 	// lastEndpointIdx records the index, in `entities`, of the http_endpoint
 	// (definition/call) entity appended by the most recent emit() call, or -1
 	// when that call emitted nothing (canonical-path empty, non-app file, or a
@@ -1348,6 +1354,16 @@ func applyHTTPEndpointSynthesis(args DetectorPassArgs) DetectorPassResult {
 		// Runs AFTER synthesizeAxumRoutes and skips any handler that pass has
 		// already registered, so one handler never gets two producers (#6530).
 		synthesizeUtoipaAxumRoutes(string(content), emit)
+		// #6668 — ADDITIVE cross-file registration markers. A `routes!(h)` whose
+		// `#[utoipa::path]` attribute lives in ANOTHER module resolves to nothing
+		// above (the attribute map is file-scoped, by the #6560 ruling). This
+		// emits a SCOPE.Route marker carrying the join key resolved from THIS
+		// file's own `use` declarations, mirroring fastapiMountPointSynthetics.
+		// It mints no endpoint and folds no path: see the header of
+		// http_endpoint_utoipa_crossfile.go for the stated bound.
+		for _, e := range utoipaCrossFileRegistrations(string(content), path) {
+			emitAdditiveSynthetic(e)
+		}
 		// Producer side (#2692): Rocket attribute macros
 		// (#[get("/path")], #[post("/path")], ...).
 		synthesizeRocket(string(content), emit)
--- a/internal/engine/http_endpoint_utoipa_crossfile.go
+++ b/internal/engine/http_endpoint_utoipa_crossfile.go
@@ -144,6 +144,13 @@ var rustUseDeclRe = regexp.MustCompile(`(?m)^[ \t]*(?:pub(?:\s*\([^)]*\))?[ \t]+
 var rustUsePathRe = regexp.MustCompile(`^[A-Za-z_]\w*(?:::[A-Za-z_]\w*)*$`)
 
 // rustUseIdentRe validates a leaf name.
+//
+// Widening it to admit `*` is EQUIVALENT under the CURRENT mint recognition
+// rule and only that rule: a `*` binding is unreachable because
+// utoipaRoutesMacroRe admits only `[A-Za-z_]\w*` as a macro argument. If that
+// rule is ever widened to path-qualified arguments (`routes!(crate::items::x)`,
+// listed as future work on #6669), the equivalence LAPSES and this validation
+// becomes load-bearing — so widen the two together or not at all.
 var rustUseIdentRe = regexp.MustCompile(`^[A-Za-z_]\w*$`)
 
 // rustUseLeafAsRe splits a `name as alias` leaf.
@@ -157,12 +164,22 @@ var rustUseLeafAsRe = regexp.MustCompile(`^([A-Za-z_]\w*)[ \t\r\n]+as[ \t\r\n]+(
 // `self`, `*`, and any leaf that is not a bare identifier (or `ident as ident`)
 // are dropped: none of them names a resolvable single item.
 func rustAddUseLeaf(out map[string]rustUseBinding, module, leaf string) {
-	if module == "" || !rustUsePathRe.MatchString(module) {
-		return
-	}
 	name, local := leaf, leaf
 	if m := rustUseLeafAsRe.FindStringSubmatch(leaf); m != nil {
 		name, local = m[1], m[2]
+	}
+	// A group leaf may itself be a PATH — `use crate::{items::create_item,
+	// admin::purge};` is the most common rustfmt grouping, and rejecting it
+	// would leave the dominant real-world spelling silently unresolved. The
+	// leading segments belong to the module, the last to the item.
+	if cut := strings.LastIndex(name, "::"); cut >= 0 {
+		module, name = module+"::"+name[:cut], name[cut+2:]
+		if local == leaf {
+			local = name
+		}
+	}
+	if module == "" || !rustUsePathRe.MatchString(module) {
+		return
 	}
 	if !rustUseIdentRe.MatchString(name) || !rustUseIdentRe.MatchString(local) {
 		return
@@ -170,11 +187,21 @@ func rustAddUseLeaf(out map[string]rustUseBinding, module, leaf string) {
 	if name == "self" || local == "self" {
 		return
 	}
-	if _, dup := out[local]; dup {
-		// Two declarations bind one local name. Rust would not compile this;
-		// keep the first rather than guess which one the macro meant.
+	// A `self::`- or `super::`-rooted module is resolvable only against the
+	// declaring file's own position in the module tree, which this per-file pass
+	// does not know. Stamping it verbatim would publish a join key that cannot
+	// match anything by identity — a guess wearing the shape of an answer, which
+	// is what #6150 forbids. `crate::…` and an absolute crate path are both
+	// resolvable and are kept.
+	if root := module; root == "self" || root == "super" ||
+		strings.HasPrefix(root, "self::") || strings.HasPrefix(root, "super::") {
 		return
 	}
+	// No first-wins guard here on purpose. Two `use` declarations binding one
+	// local name is E0252 — it does not compile — so a guard for it defends
+	// against input Rust cannot produce, and a mutant deleting it cannot be
+	// killed by any valid fixture. This file deletes unkillable guards rather
+	// than keeping them (the same disposition applied to the `emitted` map).
 	out[local] = rustUseBinding{module: module, name: name}
 }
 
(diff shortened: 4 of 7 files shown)


cajasmota:
Independent adversarial review: **REQUEST-CHANGES**. Three defects, all in the permissive direction, all in code this body describes as guarded. Sent to the author; recording it here so the reasoning is public.

## 1. R1 — the `use`-binding guard is unpinned, and the test naming the case cannot observe it

Deleting `if !bound { continue }` (the #6150 "never guess a module" clause): `go vet` 0, full package **exit 0 — SURVIVED**. Under the mutant all three subcases emit a blank-join-key marker and the test still passes:

```
no-binding / glob / nested-group → markers=1  ID=utoipa:registration:src/router.rs:::  handler_module=  handler_name=
```

`requireNoUtoipaMarkerFor` filters on `e.Properties["handler_name"] == name`, so a marker with an **empty** handler_name is invisible to it — and unlike `SameFileHandlerGetsNoMarker`, `BraceGroupAndAlias` and `PathsAgree`, it carries no total-count backstop.

This is the same shape the author already found and fixed in `utoipaMarkers` (a helper that silently collapsed what its test existed to count), surviving in a second helper. That it recurred within one PR is the useful part: **the assertion layer needs the same scrutiny as the fixture layer**, every time.

## 2. The deleted brace-count guard is NOT subsumed — the code fabricates a join key

On unmutated head:

```
use crate::{items::{create_item, purge}, admin};  routes!(purge)
  → handler_module=crate  handler_name=purge     (real path: crate::items::purge)
```

`rustUseBindings` cuts the group at the **first** `}` — the inner one — so a leaf after a comma inside a nested group is attributed to the outer base. The subsumption argument ("the comma split tears the inner braces apart and every fragment fails leaf validation") is false for this shape: the fragment is a bare identifier and passes validation.

Since #6669 resolves on `(handler_module, handler_name)`, this **mis-joins** rather than misses — worse than a phantom, and exactly what #6150 exists to prevent. The `nested-group` subcase also contains no nested braces, so it never exercised the case it names.

**This one is my error, not the author's.** I approved deleting that guard on the strength of a summary saying it was subsumed. My standing rule — an unkillable guard should be deleted rather than propped up with a manufactured fixture — is right, but it requires the subsumption to be **demonstrated**, and here it was asserted. Deleting code on an unverified subsumption claim is just a silent regression with better manners.

## 3. The marker is not inert

`internal/enrichment/candidates.go:430` sets `qualifyHTTPKinds["SCOPE.Route"] = true`, consulted unconditionally at `:599` before any structural signal, and `kindBaseScore` returns `80 / base_http_endpoint:80`. A record shaped as emitted scores `qualifies=true, score=80, band="critical"`; `EnrichmentRequired: false` is never consulted on that path.

Every marker therefore becomes a **critical-band `describe_entity` candidate** — LLM budget spent describing a join-key stub. Lower impact but same class: `internal/coverage/reachability.go:124` counts `EntityKindRoute` in coverage denominators, and `internal/links/reachability.go:95` + `internal/mcp/dead_code.go:192` seed `SCOPE.Route` as a framework entry point.

The **kind choice remains correct** — verified both directions. `http_endpoint_resolve.go:289-305` rewrites every `Kind == httpEndpointKind` to `httpEndpointDefinitionKind` unconditionally (only `http_endpoint_client_synthesis` escapes), so a pathless marker in that family really would become a phantom definition; and `SCOPE.Route` appears in no `resolverKindEquivalents` key or value. What does not follow is "therefore inert".

## Confirmed, and worth stating because they were argued rather than assumed

- **FastAPI did not move.** `emitAdditiveSynthetic`'s body is byte-identical to the removed `emitMountPoint`, so a behaviour change is impossible by construction — and a no-op positive control fails 10+ `TestSynth_FastAPI_*` tests, proving that call site is exercised rather than merely present.
- **The other deletion is safe**: two local names aliasing one declaring pair still yields exactly one marker via the ID dedupe.
- **M11's equivalence holds under the current recognition rule only.** Widening to path-qualified `routes!(crate::items::list)` — already listed here as future work — makes it non-equivalent. That belongs in the comment, since an equivalence recorded without its precondition is how a stale claim starts.

cajasmota:
Independent **re-review** at `bedd339b1` (a different reviewer from round 1): **REQUEST-CHANGES**. All three round-1 defects are genuinely fixed; these are new, and the first one changes a rule rather than just a line.

## 1. The `dup` guard deletion rests on a false claim about Rust

`http_endpoint_utoipa_crossfile.go:200-204` argues the guard is unkillable because two `use` declarations binding one local name is E0252 and "does not compile". **E0252 is per-scope.** This compiles:

```rust
use crate::items::create_item;
mod tests { use crate::mocks::create_item; }   // different scope — no E0252
OpenApiRouter::new().routes(routes!(create_item))
```

Head emits `handler_module=crate::mocks`. The `#[cfg(feature)]` / `#[cfg(not(feature))]` pair compiles as well and head picks `crate::stub`. The map is **last-wins**, and a `mod tests` block conventionally sits at the bottom of the file — so it always beats the top-level `use`. Since #6669 joins on `(handler_module, handler_name)`, this **mis-joins**, which this PR's own body calls "worse than a phantom" and attributes to #6150.

Restoring the guard returns all three shapes to `crate::items`, so a valid fixture kills it and the justification is simply false.

## The rule this changes — mine, and I am tightening it

I have been telling agents: *an unkillable guard should be deleted rather than propped up with a manufactured fixture.* That is still right, and it is what the `emitted`-map deletion correctly did. But on this PR it was satisfied **twice** by an argument rather than a demonstration, and both arguments were wrong:

- round 1, the nested-brace guard — "the comma split tears the inner braces apart and every fragment fails leaf validation", false for `use crate::{items::{create_item, purge}, admin};`
- round 2, this one — "E0252, does not compile", false for any second scope

Both produced a fabricated join key: the same defect, reached twice, by the same shortcut. **New standing rule: a guard may be deleted only after a reported attempt to construct a *compiling* counter-example.** "No valid input can reach this" is a claim about a language, not about the code, and language claims are precisely where confident reasoning fails. I accepted the first one on a summary, which is my error, not the author's.

Related and worth stating: a deletion argument written *before* a rework may not survive it. The `dup` guard predated the new `LastIndex` path splitting, which creates collision shapes the original argument never contemplated.

## 2. The new path-resolution code's only defence is unpinned

`M-B`: `if module == "" || !rustUsePathRe.MatchString(module)` → `if module == ""`. `go vet` 0, full package **exit 0 — SURVIVED**, and non-equivalent:

| input | head | under M-B |
|---|---|---|
| `use crate::{::create_item};` | no marker | `handler_module=crate::` |
| `use crate::{a::::create_item};` | no marker | `handler_module=crate::a::` |
| `use crate::{1bad::create_item};` | no marker | `handler_module=crate::1bad` |

`rustUsePathRe` is the only thing between the new `module+"::"+name[:cut]` concatenation and a published garbage join key, and no fixture names it.

## 3. `path-leaf + alias` resolves to nothing, and its guard is dead code

`use crate::{items::create_item as mk};` → no marker: `rustUseLeafAsRe` runs *before* the `LastIndex` split and cannot match a leaf containing `::`. Consequently `if local == leaf { local = name }` at `:177` is **invariantly true** — an alias match yields a `name` that cannot contain `::`, so `cut >= 0` implies no alias matched. The condition documents a case that cannot reach it while the case it appears to handle is broken. Restrictive, therefore invisible.

## Verified rather than accepted — three claims that held

- **The commented-`routes!` ruling is correct.** Checked directly: `synthesizeAxumRoutes` mints from a line-commented `.route(`, and `synthesizeUtoipaAxumRoutes` mints from a line-commented `.routes(routes!(...))` — remove the registration and it mints nothing, so the commented line really is what is read. A marker stricter than the mint sharing its regex would split one occurrence across two passes.
- **The enrichment exclusion is scored in both directions.** The reviewer killed the permissive mutant; I killed the over-strict one myself (`enrichmentMarkerSubtypes[e.Subtype]` → `true`: 6 failures including `TestQualifiesForEnrichment_Signals`). Settled.
- **Two of the three `SCOPE.Route` consumers are benign** — the marker seeds BFS as its own entry point, carries no outbound edges, and propagates reachability to nothing.

## 4. A fourth consumer, disclosed but unfixed

`internal/coverage/reachability.go:112` counts every marker as a production entity, and `isTestEntity` filters only `subtype == test_suite` — so each marker is counted **untested**, and reported coverage drops by one per cross-file registration on every utoipa repo. The body discloses this and fixes only the enrichment half; the same two-line Subtype exclusion transplants verbatim.

Issue #6499, opened 2026-08-21
spring_routes_kotlin still targets Controller:<bareMethod> while #6429 moved Java to SCOPE.Operation:<Class>.<method> — one Spring construct, two shapes, split by file language


cajasmota:
Found while grounding #6442. Not that issue's defect, but a live inconsistency introduced by #6429.

## The asymmetry

**#6429 moved the Java Spring AST route edge** to target `SCOPE.Operation:<Class>.<method>`, and stamped `handler_class` / `handler_method` properties alongside it (`e15170110`).

**Kotlin was left behind.** `internal/engine/spring_routes_kotlin.go:214,223` still emits:

- `ToID` of `Controller:<bareMethod>`, and
- `source_handler=Controller:<method>`

So the same Spring construct produces a class-qualified operation target in a `.java` file and a bare-method controller target in a `.kt` file. Anything consuming these edges — resolution, handler attribution, disposition accounting, any consumer keying on `handler_class` — now sees two different shapes for one concept, split by source language.

## Why it matters beyond tidiness

`Controller:<bareMethod>` is precisely the shape that makes handler identity ambiguous across classes: two Kotlin controllers in one file declaring the same method name are indistinguishable at the ToID level. That is the same root property behind the over-dedup filed alongside this, and #6429's whole point on the Java side was to remove it.

It also means a Kotlin Spring repo silently gets the pre-#6429 behaviour while a Java one gets the fixed behaviour, with nothing reporting the difference — the language of the file decides which semantics apply.

## Worth establishing before fixing

1. **Does the Kotlin path have the class name available** at the point the edge is minted, the way `#6429` found it for Java? If not, this is a larger change than a retarget.
2. **What reads `source_handler`?** If any consumer branches on the `Controller:` prefix, changing the Kotlin shape without updating it moves a second thing.
3. **Is there a Kotlin Spring fixture at all?** If not, the fix lands unobservable and the fixture is part of the task. (`java-spring-mini` exists; a Kotlin equivalent was not found while grounding #6442, but that was not an exhaustive search.)

**Killing mutant:** revert the Kotlin target to `Controller:<bareMethod>`; a fixture asserting the class-qualified target must go red.

## Note

Both directions of this are defensible — align Kotlin to Java, or establish that the two genuinely differ and document why. What is not defensible is the current state, where the difference is undocumented and unasserted, so nobody can tell whether it is intentional.

cajasmota:
[...]

Issue #6488, opened 2026-08-21
golden fixtures: arms A+B+C shipped (subtype, absence, from_bare_name) — only arm D remains, now unblocked


cajasmota:
Three distinct assertions the golden fixture format **cannot express**, all found by mutants that survived a fully green fixture. These are not wrong assertions — they are assertions that cannot be written, which is why no audit has caught them.

## 1. Entity `Subtype` is unassertable

`internal/quality/expected.go` has **no subtype field**. `ExpectedEntity` carries name, kind and file; nothing more.

Measured on `proto-mini` (PR #6486): changing `buildEnumValue`'s `Subtype` from `"enum_value"` to `"field"` leaves the fixture at **18/18 entities, 16/16 relationships, 0 forbidden, exit 0**.

**This compounds with a known identity property.** `graph.EntityID = sha256(repo, kind, name, sourceFile)` — **`Subtype` is not hashed.** So a subtype change is invisible to identity *and* invisible to every fixture in the repo. It is visible to consumers: `internal/mcp/denoise.go` filters on it, dashboard views group by it, and several resolver paths branch on it.

Subtype has been the proximate cause of real defects this cycle — #6440 (VB.NET overloads), #6422 (proto message/enum), and the F# import marker whose entire fix was *"stamp the right Subtype"* (#6369/PR #6480). **A field that decides behaviour, is not hashed into identity, and cannot be asserted by any fixture is the weakest link in the measurement chain.**

## 2. Entity over-emission is unassertable

`Fixture` has `ForbiddenRelationships` but **no `forbidden_entities`**, and `baseline.json` records no extracted totals — only `entity_found` / `entity_expected`.

So a fixture cannot say *"this entity must NOT exist"*, and nothing notices when extracted totals grow. `proto-mini` extracts **29** entities and asserts 18; the unasserted 11 are near-duplicate families — `proto_message:User` alongside `User`, `proto_enum:Role`, `grpc_service:UserService`, and **three** `SCOPE.Operation` entities per rpc (`User`, `RPC /UserService/User`, `UNARY /UserService/User`).

Those could double and every fixture would stay green.

## 3. File-anchored edges are unassertable

`ExpectedRelationship` has `to_bare_name` but **no `from_bare_name`**, and `resolveExpectedEdge` requires at least one resolvable `from` candidate.

[...]

Issue #6429, opened 2026-08-20
Spring MVC: Route ROUTES_TO targets stay unresolved placeholders, so endpoint→handler traces dead-end


auxmedrano:
**Impact:** Resolver bug (edges missing or wrong disposition)
**grafel version:** v0.2.2-40-g6f14cb873 (dev build from source, `make build-go-only`)
**Report anonymization:** reviewed — hashed identifiers, scrubbed paths, no source.

### Summary

On a real Spring MVC (Java) service, grafel extracts routes, `http_endpoint_definition`s **and** the concrete handler methods, and it wires the handler's downstream calls (`method → service → repository`). But the **route→handler hop is never resolved**: the `Route` node emits `ROUTES_TO` edges whose targets are unresolved placeholder symbols instead of the concrete `SCOPE.Operation` handler that exists in the same graph. Because that one hop dangles, an endpoint-rooted downstream DAG dead-ends at the `Route`, even though everything below the handler is fully connected.

### Expected vs. actual

- **Expected:** `Route ─[ROUTES_TO]→ SCOPE.Operation(<Controller>.<method>)` — the concrete handler node.
- **Actual:** `Route ─[ROUTES_TO]→ Controller:<method>` (a synthetic placeholder id) and, for some verbs, `─[ROUTES_TO]→ ext:<method>` (the method mis-bucketed as an external symbol). Neither placeholder is the real handler node, which is present in the graph with its own outgoing `CALLS` edges to the service layer.

### Evidence (one endpoint, genericized)

```
Route  /api/.../{id}/<resource>
   ─[ROUTES_TO]→ Controller:create      # placeholder — does NOT point at the real method node
   ─[ROUTES_TO]→ ext:list               # 'list' handler bucketed as external

http_endpoint_definition (POST /.../{id}/<resource>)
   ─[IMPLEMENTS]→ … , ─[ENTRY_POINT_OF]→ proc:…      # route-level process wired
   (no edge into the handler SCOPE.Operation)

SCOPE.Operation  <Controller>.create        # the REAL handler, present in the graph
   ─[CALLS]→ <Service>.<method>              # downstream method→service→repo fully wired
```

So the graph has both endpoints of the missing edge; only the resolution that would join `Controller:create` / `ext:list` to the concrete `SCOPE.Operation` is absent. A forced clean `rebuild` did not change it (not a cache/tier artifact).

### Why it matters

[...]

Issue #67, opened 2026-05-08
Spring class-level @RequestMapping prefix not composed with method-level @GetMapping


cajasmota:
Spring controllers commonly nest:

```java
@RequestMapping("/api")
@RestController
public class OrderController {
  @GetMapping("/orders")
  public List<Order> list() { ... }
}
```

Real route is `/api/orders`, but the YAML rules in `internal/engine/rules/java/frameworks/spring_mvc.yaml` emit `Route:/api` and `Route:/orders` as separate orphan entities (no composition).

**Why not blocking #51:** strict improvement over zero ROUTES_TO output. But for graph quality this needs class+method composition — likely requires moving Spring routes from regex/YAML to AST-aware extractor (`spring_ecosystem_extractor` exists in Python codebase already).

Surfaced by review of 96bf33e (Refs #51).

cajasmota:
Implementer subagent dispatched in isolated worktree.

cajasmota:
## Fix: AST-driven Spring route composition

**Approach: new AST pass in the engine package, NOT a YAML extension.**

Composing class-level `@RequestMapping` with method-level verb annotations is fundamentally a lexical-scope problem. The YAML rules engine compiles a regex per pattern and emits matches with no class context, so it cannot tie a class-level prefix to the methods inside that class. Adding it as a YAML extension would require either a new "scope-aware" rule kind in the engine schema (large surface change) or per-framework regex hacks that break on nested classes / multi-class files. The AST pass is small, isolated, and uses the existing `treesitter.ParserFactory` already used everywhere else.

### Architecture
1. New file `internal/engine/spring_routes.go` runs as a post-process inside `engine.Detect` after the YAML rules have produced their raw entities/relationships.
2. Cheap pre-filter: only Java files containing `@RestController` or `@Controller` are parsed by tree-sitter (avoids paying for files that obviously aren't controllers).
3. Walks `class_declaration` nodes, identifies HTTP-controller classes carrying a class-level `@RequestMapping`, captures the prefix, and for every method with a verb annotation emits a composed `Route:<prefix><method_path>` entity + matching `ROUTES_TO` edge.
4. The pass also reports the bare method paths and class-level prefix it consumed so the surrounding engine can filter out the now-redundant orphan Routes from the YAML output.

### Double-emit handling
Both halves run on every Java file, so without explicit suppression the YAML pass and the AST pass would emit overlapping entities. The AST pass returns three sets used to filter the YAML output:
[...]

PR #6702, opened 2026-08-26
fix(#6498): the Spring handler claim was keyed on a bare method name, so a sibling controller lost its only route edge

### LINKED ISSUE #6498: spring_routes: claimedHandlerMethods is keyed on the bare method name and scoped to the file, so an unmapped controller class loses its only route edge to a same-named method in a sibling class
Found while disproving #6442's stated mechanism. That issue's `ext:` causal claim does not hold, but the dedup it points at has a **different, real defect running in the opposite direction** — it drops edges it should keep.

## The defect

`applySpringRouteComposition` builds `claimedHandlerMethods` and then filters at `internal/engine/spring_routes.go:115`:

```go
if r.Kind == "ROUTES_TO" && strings.HasPrefix(r.ToID, "Controller:") {
    method := strings.TrimPrefix(r.ToID, "Controller:")
    if composed.claimedHandlerMethods[method] { continue }
}
```

**The set is keyed on the bare method name and scoped to the file, not to the class.**

So for one `.java` file containing two controller classes:

- `OrdersController` has a class-level `@RequestMapping`, so the AST pass composes its routes and claims method name `list`.
- `ReportsController` has **no** class-level `@RequestMapping`, so the AST pass skips it entirely (the `hasClassMapping` gate from #67) and its only edge is the YAML one.
- If `ReportsController` also declares `list`, its **legitimate** YAML `ROUTES_TO` is dropped — claimed by a method on a different class.

That is silent recall loss: an endpoint that exists in the source has no route edge in the graph, and nothing reports it.

## Why the unmapped-class case is not hypothetical

`HealthController` in `java-spring-mini` is exactly that shape — no class-level `@RequestMapping`, AST pass skips it, and the YAML `Controller:health` edge is the **only** edge it has. It is asserted at `java-spring-mini/expected.json:117`. So the "YAML edge is sole survivor" path is live and already load-bearing in a fixture; the collision case just needs a same-named method in a mapped sibling class.

This is also the reason a blanket "drop the YAML twin" fix to the sibling problem would be wrong — it would delete `HealthController`'s only edge.

## The fix shape

Key the claim on class-qualified identity rather than the bare method name. #6429 already moved the Java AST edge to `SCOPE.Operation:<Class>.<method>` and stamps `handler_class`/`handler_method` properties, so the qualifying information is present at the point the set is built — this looks like a keying change, not new plumbing.

Worth checking whether `django_routes.go:274,327,350` shares the shape; its in-pass invariant is sound (that was checked while grounding #6442) but the scoping question is separate and was not examined.

## Observability

**Not expressible in the golden harness.** `internal/quality/expected.go:25-104` has `ForbiddenRelationships` only — no cardinality field, and `must_exist` is presence-only.

It *is* expressible as a Go-level detector test, which is where the existing coverage lives: `internal/engine/spring_routes_issue67_test.go:80-97` already counts `ROUTES_TO` across a whole `Detect()` result.

**Killing mutant:** a fixture with two controller classes in one file — one mapped, one unmapped — sharing a method name. Revert the key to the bare method name and the unmapped class's edge disappears → assertion red. Note the fixture must use a **quoted** path argument (`@GetMapping("/x")`), because every `spring_mvc.yaml` relationship regex requires one; a bare `@GetMapping` produces no YAML edge at all and the test would pass vacuously.

## Unverified

Whether this actually fires on the corpus — establishing that needs an index run and a query for controller classes sharing a method name within one file. It may be rare. It is silent when it happens, which is the argument for fixing it regardless of frequency.

cajasmota:
Fixes #6498. Follow-ups filed: #6703, #6704.

## The defect

`internal/engine/spring_routes.go` built `claimedHandlerMethods` as a **file-scoped set keyed on the bare handler method name**. Two controller classes in one `.java` file sharing a handler method name collided: the mapped class (class-level `@RequestMapping`, so the AST pass composes it) claimed the name, and the unmapped sibling — which the `hasClassMapping` gate from #67 skips, so the YAML edge is its **only** edge — had that edge dropped. A real endpoint ends up with no route edge. Nothing dangles, nothing goes red: silent recall loss.

## The fix

Re-key the claim on the **qualifying pair** (`springHandlerClaimKey`), NUL-separated.

### Why the pair is (path, method) and not (class, method)

The issue proposed `className` + method. `className` is in scope at the *claim* site but not at the *filter* site, which is where the lookup happens: `applySpringRouteComposition` iterates `rawRels` and sees only `FromID`/`ToID`/`Kind`. All six `ROUTES_TO` rules in `spring_mvc.yaml` are `source_type: Route` / `target_type: Controller`, so the edge is `Route:<barePath> -> Controller:<bareMethod>` with no class component, and it is the only Java rule file emitting `ROUTES_TO`. A class-keyed claim would never match — it degenerates into "remove the filter" (mutant 2). The class-distinguishing information carried on **both** sides is the annotation path. *(This deviation was independently verified from source during review.)*

Field renamed `claimedHandlerMethods` -> `claimedHandlerEdges`, since it is no longer a set of method names.

### F1 — the two halves are computed by different engines (fixed in 485325e)

The first revision of this PR introduced a precision regression, caught in review. The claim side (`annotationNameAndPath`) prefers the `value=`/`path=` named argument; the YAML side (`[^)]*["']([^"'\n\r]+)["'][^)]*\)`) is **greedy and captures the LAST literal** in the argument list. They disagree on an ordinary Spring shape:

```java
@PostMapping(value = "/things", consumes = "application/json")
public String create() { return "x"; }
```

The pair never matched, so `Route:application/json -> Controller:create` stopped being suppressed — an edge dangling on **both** ends (no `Route:application/json` entity exists, since the entity rule anchors on the FIRST literal; and `Controller:` stubs resolve to nothing, #6429). That is exactly the hop #6429 existed to remove, and the old bare-method key was immune to it because it never looked at the path.

Fixed by also claiming **the literal the YAML regex actually lands on** — `annotationLastStringLiteral` returns the LAST string literal in the verb annotation's argument list, since each of the six rules is greedy and captures exactly one literal per annotation.

The widening is bounded on three sides, and each bound is pinned by a test rather than asserted in prose:

- **the verb annotation only** — an `@Operation(summary = "/probe")` literal is not claimed (`NonVerbAnnotationLiteralsAreNotClaimed`);
- **the enclosing method only** — a literal is claimed against its own handler, never against every method in the class (`LiteralClaimIsBoundToItsOwnMethod`, review finding A);
- **the reachable literal only** — earlier literals in `value = {"/a", "/b"}` are not claimed, because the regex can never emit an edge for them (`ArrayValueClaimsOnlyTheReachableLiteral`, review finding B).

Without those bounds the F1 fix re-creates #6498 through its own widening; findings A and B were both live defects on `485325e`.

F1 red control, on `7ecf6d70` before the fix:

```
--- FAIL: TestDetect_SpringRoute_NamedArgsWithMediaTypeLiteral (0.03s)
    spring_routes_issue6498_test.go:251: UNEXPECTED ROUTES_TO edge: Route:application/json -> Controller:create
          got edges: [Route:/api/things -> SCOPE.Operation:ProbeController.create Route:/api/widgets -> SCOPE.Operation:ProbeController.read Route:application/json -> Controller:create Route:text/plain -> Controller:read]
    spring_routes_issue6498_test.go:251: UNEXPECTED ROUTES_TO edge: Route:text/plain -> Controller:read
```

### Findings A and B — the widening was too wide (fixed in 9306c74)

Two over-suppression defects on `485325e`, both caught in review, both of which let the F1 widening delete a real edge.

**B — array recursion over-claimed literals the regex can never reach.** All six rules are `…[^)]*["']([^"'\n\r]+)["'][^)]*\)`: greedy, one capture per annotation, always the LAST literal. So for `@GetMapping(value = {"/a", "/b"})` only `/b` is reachable, and claiming `/a` claimed something never emitted — swallowing a sibling controller's genuine `/a` edge. Red control on `485325e`:

```
--- FAIL: TestDetect_SpringRoute_ArrayValueClaimsOnlyTheReachableLiteral (0.02s)
    spring_routes_issue6498_test.go:394: MISSING ROUTES_TO edge: Route:/a -> Controller:multi
          got edges: [Route:/api -> SCOPE.Operation:ArrController.multi]
```

That fixture pins **both** halves at once: `/a` must survive and `/b`'s YAML twin must still be suppressed. Mutant S below (recursion into the array initializer disabled) fails it from the opposite side.

**A — the enclosing-method bound was unobserved.** The doc comment said each literal is claimed against its enclosing method; claiming it against every method in the class instead passed the entire `internal/engine` + `internal/quality` suite. Not equivalent — mutant R now dies on a ~35-line fixture where `/alpha` belongs to `MappedController.one` and a sibling controller's only edge hangs off a handler that merely shares the name `two`:

```
--- FAIL: TestDetect_SpringRoute_LiteralClaimIsBoundToItsOwnMethod (0.02s)
    spring_routes_issue6498_test.go:348: MISSING ROUTES_TO edge: Route:/alpha -> Controller:two
```

Same standard as mutant 5: killed with a fixture rather than recorded as equivalent.

The doc comment also named `array_initializer`; tree-sitter-java's node here is **`element_value_array_initializer`**. Behaviour was fine (the walk is generic), but a mutant written against the wrong name is a silent no-op, so the comment now names the node the code actually sees.

### The `apath` claim removed — it was over-suppressing (fixed in 588b4b7)

Review round 4 ran a mutant nobody had: deleting `claimedHandlerEdges[springHandlerClaimKey(apath, methodName)]`, leaving only the last-literal claim. It survived the whole suite, and the reasoning for why was correct — every relationship rule is `@XMapping\s*\([^)]*["']([^"'\n\r]+)["'][^)]*\)`, so the emitted edge is always `Route:<last literal>` and the filter can only ever look the last literal up. `apath` is consulted only when it *is* the last literal, where it is exactly redundant.

Probing before deleting showed it is worse than redundant. When the two differ, the `apath` claim claims an edge the regex never emitted and destroys someone else's instead. `name` is a real `@RequestMapping` attribute, so:

```java
@GetMapping(path = "/x", name = "foo")   // AST path = /x, regex capture = foo
public String m() { ... }
```

claimed `/x`, which deleted an unmapped sibling controller's only edge. Measured on the same source with only that line differing:

```
with the apath claim:     [Route:/api/x -> SCOPE.Operation:NamedController.m]
without it:               [Route:/api/x -> SCOPE.Operation:NamedController.m,
                           Route:/x -> Controller:m]      <-- sibling's only edge
```

So the line went, and the deletion is **pinned** rather than merely made: `TestDetect_SpringRoute_ApathIsNotClaimedWhenItIsNotTheLastLiteral` fails with `MISSING ROUTES_TO edge: Route:/x -> Controller:m` if the line comes back (mutant T).

The comment's old justification — *"`apath` is claimed explicitly so the `apath == ""` (bare `@GetMapping`) case still registers"* — described a case that cannot arise: every rule requires `@XMapping\s*\(` **and** a quoted literal, so a bare `@GetMapping` emits no edge and there is nothing to suppress. It has been replaced with the regex argument and a pointer to the test.

### F2 — prose weakened

`springHandlerClaimKey`'s comment claimed the NUL separator "cannot be forged". No test observes NUL-ness (`|` passes the suite), and the absolute is false for `$`/`_`, which are legal in Java identifiers. Rewritten to state what `TestDetect_SpringRoute_ConcatAmbiguousClaimKey` actually observes — that the halves are separated rather than concatenated — with NUL named as one adequate choice, not as an unforgeable one.

**Not touched:** `spring_routes_kotlin.go` (#6499), and no fixture-format field (#6488).

## Tests — axes varied vs held constant

Eight Go-level pins in `internal/engine/spring_routes_issue6498_test.go`. All assert the **exhaustive `ROUTES_TO` edge set** (from -> to), never a count — an extra edge is reported as `UNEXPECTED` rather than ignored, so a wrong-but-equal number cannot pass.

| Test | VARIED | HELD CONSTANT |
|---|---|---|
| `SameMethodNameInTwoControllers` | the controller class (mapped / unmapped) | the handler method name (`list` in both) |
| `SamePathTwoControllers` | the handler method name (`ping` / `legacyPing`) | the annotation path (`/ping` in both) |
| `ConcatAmbiguousClaimKey` | both halves at once | their concatenation (`/ab`+`c` == `/a`+`bc`) |
| `NamedArgsWithMediaTypeLiteral` | which argument holds the literal (`value=` vs `consumes=`/`produces=`) | class, method name, real path |
| `NonVerbAnnotationLiteralsAreNotClaimed` | which annotation holds `/probe` (`@Operation` vs `@GetMapping`) | the handler method name (`check` in both) |
| `LiteralClaimIsBoundToItsOwnMethod` | which method inside the mapped class owns the literal (`one` owns `/alpha`) | the literal `/alpha` and the method name `two`, split across different methods |
| `ArrayValueClaimsOnlyTheReachableLiteral` | position in the array initializer (first vs last literal) | the handler method name (`multi` in both) |
| `ApathIsNotClaimedWhenItIsNotTheLastLiteral` | whether the literal is the AST's path (`path=`) or the regex's capture (`name=`) | the handler method name (`m` in both) and the path string `/x` |

Test 1 is the issue. Test 2 pins the *other* half of the pair. Tests 4-8 are F1, its three bounds, and the removal of the redundant `apath` claim.

### Red control for #6498 itself (before the fix)

```
--- FAIL: TestDetect_SpringRoute_SameMethodNameInTwoControllers (0.04s)
    spring_routes_issue6498_test.go:116: MISSING ROUTES_TO edge: Route:/reports -> Controller:list
          got edges: [Route:/orders/detail -> SCOPE.Operation:OrdersController.detail Route:/orders/list -> SCOPE.Operation:OrdersController.list]
```

## Mutation score

`go vet` clean before every run; `-count=1` throughout; file restored byte-for-byte by `cp` and verified by `shasum` between mutants, never `git checkout`; one mutant per invocation.

Re-scored on `588b4b7` after the `apath` deletion: **6** (its target block is now the *sole* claim, so its premise genuinely changed) and the new mutant **T**. Nothing else is affected: the deleted line was provably redundant on every single-literal annotation, and every other fixture except the array and named-argument ones uses single-literal annotations, whose verdicts are governed by the last-literal claim that this change does not touch.

Previously re-scored on `9306c74` after the A/B narrowing: **1, 6, 7, R, S**. Mutants **2, 3, 4 and 5** have not been re-run since their original scoring, for a structural reason rather than a budgetary one: each is killed by tests 1, 2 or 3, whose annotations each carry exactly **one** string literal — where "the last literal", "every literal" and "the AST's preferred literal" are the same string. No change to which literals get claimed can alter the claim set those three fixtures produce. (Mutant 2 additionally deletes the *reader* of the claim set, so no change to its contents can reach it; it was independently re-run during review and came back DEAD with 7 failures.)

| # | Mutant | Result | Evidence |
|---|---|---|---|
| 1 | Revert key to the bare method name (the original defect) | **DEAD** (re-scored on `9306c74`) | `MISSING ROUTES_TO edge: Route:/reports -> Controller:list` |
| 2 | Remove the claim filter entirely (permissive) | **DEAD** | `UNEXPECTED ... Route:/list -> Controller:list`, `Route:/detail -> Controller:detail`, `Route:/ping -> Controller:ping`; also failed `Issue67` and `TestDetect_SpringRoutes` |
| 3 | Key on the route path alone (drop the method half) | **DEAD** | `MISSING ROUTES_TO edge: Route:/ping -> Controller:legacyPing` |
| 4 | Permissive: match if **either** half matches (`OR` instead of the pair) | **DEAD** | `MISSING ... Route:/reports -> Controller:list` **and** `MISSING ... Route:/ping -> Controller:legacyPing` |
| 5 | Drop the `\x00` separator (bare concatenation) | **DEAD** | `MISSING ROUTES_TO edge: Route:/a -> Controller:bc` |
| 6 | Delete the literal claim (now the sole claim in the pass) | **DEAD** (re-scored on `588b4b7`) | 10 failing tests, incl. `UNEXPECTED ... Route:/list -> Controller:list`, `Route:/ping -> Controller:ping`, `Route:application/json -> Controller:create` |
| 7 | Permissive: claim literals from **non-verb** annotations too (move the claim above the `verbAnnotations` check) | **DEAD** (re-scored) | `MISSING ROUTES_TO edge: Route:/probe -> Controller:check` |
| R | Permissive: claim each literal against **every method in the class**, dropping the enclosing-method bound (review finding A) | **DEAD** | `MISSING ROUTES_TO edge: Route:/alpha -> Controller:two` — was ALIVE on `485325e` with the full suite green |
| S | Disable recursion into `element_value_array_initializer`, so `/b` stops being claimed | **DEAD** | `UNEXPECTED ROUTES_TO edge: Route:/b -> Controller:multi` |
| T | Re-add the `apath` claim alongside the last-literal claim | **DEAD** | `MISSING ROUTES_TO edge: Route:/x -> Controller:m` — this line was ALIVE as a *deletion* mutant on `9306c74` |

Mutants 5 and R were both ALIVE at some point — 5 on the first revision, R on `485325e` with the entire suite green — and both were killed with a fixture (~35 lines each) rather than recorded as equivalent. Mutants R and S are each other's mirror: R over-claims and deletes a real edge, S under-claims and lets a bogus one through, so the pair brackets the widening from both sides. Mutant T is the fourth ALIVE-then-killed case: a reviewer's deletion mutant survived, and probing it turned up not dead code but an over-suppression, so the line was removed and its removal pinned. A `HasPrefix(r.FromID, "Route:")` guard I had briefly added also scored ALIVE and is provably unreachable (every ROUTES_TO rule declares `source_type: Route`), so it was removed rather than shipped as unkillable code.

Mutant 2 being DEAD answers the issue's open question: the suppression **is** observed — only the class-collision case was not.

## Corpus measurement for F1 — the corpus cannot decide this

Scanned all **645 Java files** in `archigraph-corpora`:

| measurement | count |
|---|---|
| Spring `@(Get\|Post\|Put\|Delete\|Patch\|Request)Mapping` annotations | **20** (spring-petclinic 18, awesome-compose 2) |
| ...using `value=` / `path=` named arguments | **0** |
| ...mentioning `produces` / `consumes` at all | **0** |
| ...with >= 2 string literals in the argument list | **0** |
| F1 shape (named path plus a differing LAST literal) | **0** |

All 20 are the positional single-literal form, where the AST's literal *is* the last literal, so the two engines cannot disagree. The honest read is **not** "F1 is rare" — it is that this corpus contains no Spring code using named annotation arguments at all, so it has no power to measure the shape either way. What it does establish is that this PR is a **no-op on every Spring annotation in the corpus**: 20/20 single-literal, and the extra claims are identical to the existing one. `@GetMapping(value="/x", produces=MediaType.APPLICATION_JSON_VALUE)` (constant, not literal) was already safe and remains so.

## `java-spring-mini` did not move

`internal/quality/golden/java-spring-mini/expected.json:117` (`to_bare_name: "Controller:health"`, `must_exist: true`) is intact and untouched. `HealthController.java` is its own file with no class-level `@RequestMapping`, so `composed.entities` is empty and `applySpringRouteComposition` returns before the filter runs. *(Independently confirmed in review by building both binaries and diffing `grafel quality` output — byte-identical.)*

`go test -count=1 ./internal/engine/ ./internal/quality/...` green across engine, quality, quality/analytics, quality/audit, quality/fitness. `gofmt -l ./internal/` clean.

## Not fixed

- **#6703** — the entity filter and the relationship filter disagree, so in the test-2 shape the surviving edge's source `Route` entity has been removed. Confirmed **not** a regression (the parent behaves identically). Split out because `claimedMethodPaths` cannot be pair-keyed: a `Route` entity record carries no method.
- **#6704** — `django_routes.go:274,327,350` has the same claim-and-suppress architecture, keyed on the path half rather than the method half, with one filter suppressing by *name shape* (`isDRFGlobalRegisterName`) rather than by a recorded claim. Unexamined; needs the same fixture-and-mutant treatment, and may well come back "sound".
- **#6706** — array-valued `@GetMapping(value = {"/a", "/b"})` loses both paths and composes to the bare class prefix, because `annotationNameAndPath` only reads `element_value_pair` values that are directly a `string_literal`. Found while measuring F1; verified pre-existing (parent and HEAD emit the identical single edge). That is the *path-extraction* side; review finding B above was the *suppression* side of the same annotation shape, and is fixed here. #6706 stays out of this PR.
- Identical path **and** identical method name in two controllers in one file still collides. Unfixable at the filter site: the YAML `Route` entity is file-scope-deduped, so there is no second entity to attach a second edge to. Documented limitation.

### DIFF
--- a/internal/engine/spring_routes.go
+++ b/internal/engine/spring_routes.go
@@ -276,16 +276,27 @@ func processSpringClass(class ts.Node, src []byte, path string, out *composedSpr
 			composedPath := joinRoutePaths(prefix, apath)
 
 			out.claimedMethodPaths[apath] = true
-			// #6702/F1 — also claim the literal the YAML relationship regex
-			// actually captures (the LAST one in the argument list), which is
-			// not always the one the AST picked as the path. Both claims are
-			// bound to THIS annotation's own handler method: claiming a
-			// literal against any other method in the class would let one
-			// handler's path suppress a sibling controller's edge (finding A,
-			// pinned by TestDetect_SpringRoute_LiteralClaimIsBoundToItsOwnMethod).
-			// `apath` is claimed explicitly so the `apath == ""` (bare
-			// `@GetMapping`) case still registers.
-			out.claimedHandlerEdges[springHandlerClaimKey(apath, methodName)] = true
+			// #6702/F1 — claim the literal the YAML relationship regex actually
+			// captures: the LAST one in the argument list, which is not always
+			// the one the AST picked as the path (`apath`).
+			//
+			// `apath` itself is deliberately NOT claimed. Every relationship
+			// rule in spring_mvc.yaml has the shape
+			// `@XMapping\s*\([^)]*["']([^"'\n\r]+)["'][^)]*\)`, so the edge's
+			// FromID is ALWAYS `Route:<last literal>` and the filter can only
+			// ever look up the last literal. Claiming `apath` as well is
+			// therefore either redundant (single-literal annotations, where
+			// `apath` IS the last literal) or actively wrong: on
+			// `@GetMapping(path = "/x", name = "foo")` it claims `/x`, which
+			// the regex never emitted, and suppresses a sibling controller's
+			// real `/x` edge instead — #6498 all over again. Pinned by
+			// TestDetect_SpringRoute_ApathIsNotClaimedWhenItIsNotTheLastLiteral.
+			//
+			// The claim is bound to THIS annotation's own handler method:
+			// claiming a literal against any other method in the class would
+			// let one handler's path suppress a sibling controller's edge
+			// (finding A, pinned by
+			// TestDetect_SpringRoute_LiteralClaimIsBoundToItsOwnMethod).
 			if lit, ok := annotationLastStringLiteral(a, src); ok {
 				out.claimedHandlerEdges[springHandlerClaimKey(lit, methodName)] = true
 			}
--- a/internal/engine/spring_routes.go
+++ b/internal/engine/spring_routes.go
@@ -276,13 +276,17 @@ func processSpringClass(class ts.Node, src []byte, path string, out *composedSpr
 			composedPath := joinRoutePaths(prefix, apath)
 
 			out.claimedMethodPaths[apath] = true
-			// #6702/F1 — claim every literal the YAML relationship regex could
-			// have captured, not just the one the AST picked as the path.
-			// `apath` itself is among them for every annotation that has a
-			// path argument; it is claimed explicitly so the `apath == ""`
-			// (bare `@GetMapping`) case still registers.
+			// #6702/F1 — also claim the literal the YAML relationship regex
+			// actually captures (the LAST one in the argument list), which is
+			// not always the one the AST picked as the path. Both claims are
+			// bound to THIS annotation's own handler method: claiming a
+			// literal against any other method in the class would let one
+			// handler's path suppress a sibling controller's edge (finding A,
+			// pinned by TestDetect_SpringRoute_LiteralClaimIsBoundToItsOwnMethod).
+			// `apath` is claimed explicitly so the `apath == ""` (bare
+			// `@GetMapping`) case still registers.
 			out.claimedHandlerEdges[springHandlerClaimKey(apath, methodName)] = true
-			for _, lit := range annotationStringLiterals(a, src) {
+			if lit, ok := annotationLastStringLiteral(a, src); ok {
 				out.claimedHandlerEdges[springHandlerClaimKey(lit, methodName)] = true
 			}
 
@@ -412,37 +416,48 @@ func annotationNameAndPath(anno ts.Node, src []byte) (string, string) {
 	return name, positional
 }
 
-// annotationStringLiterals returns EVERY string literal in the annotation's
-// argument list, in source order, including ones nested inside an
-// array_initializer (`value = {"/a", "/b"}`).
+// annotationLastStringLiteral returns the LAST string literal in the
+// annotation's argument list, which is the only literal the spring_mvc.yaml
+// relationship rules can ever capture.
 //
 // #6702/F1 — the AST side and the YAML side disagree about which literal is
-// the path: annotationNameAndPath prefers the `value=`/`path=` pair, while the
-// spring_mvc.yaml relationship regex (`[^)]*["']([^"'\n\r]+)["'][^)]*\)`) is
-// greedy and captures the LAST literal in the argument list. On
+// the path: annotationNameAndPath prefers the `value=`/`path=` pair, while each
+// of the six relationship rules is `[^)]*["']([^"'\n\r]+)["'][^)]*\)` —
+// greedy, one capture per annotation, always the LAST literal. On
 // `@PostMapping(value = "/things", consumes = "application/json")` that is the
 // media type, so the YAML edge is `Route:application/json -> Controller:create`.
-// Claiming only the AST's preferred literal would leave that edge unsuppressed
-// — dangling on both ends, since no `Route:application/json` entity exists (the
+// Claiming only the AST's preferred literal left that edge unsuppressed —
+// dangling on both ends, since no `Route:application/json` entity exists (the
 // entity rule anchors on the FIRST literal) and `Controller:` stubs resolve to
-// nothing (#6429). Claiming every literal the regex could land on restores
-// suppression without giving up the pair key that #6498 needs.
-func annotationStringLiterals(anno ts.Node, src []byte) []string {
+// nothing (#6429).
+//
+// #6702/finding B — this returns the last literal, NOT every literal. The
+// regex emits exactly one edge per annotation, so any earlier literal is
+// claimed for something that was never emitted. For
+// `@GetMapping(value = {"/a", "/b"})` only `/b` is reachable; claiming `/a` as
+// well would swallow a sibling controller's genuine `/a` edge, which is #6498
+// re-entering through this widening. Pinned by
+// TestDetect_SpringRoute_ArrayValueClaimsOnlyTheReachableLiteral.
+//
+// The walk is generic rather than node-type-driven so it reaches literals
+// nested inside an `element_value_array_initializer` (tree-sitter-java's node
+// type for the `{...}` in `value = {"/a", "/b"}`) without naming it.
+func annotationLastStringLiteral(anno ts.Node, src []byte) (string, bool) {
 	if anno == nil {
-		return nil
+		return "", false
 	}
 	args := anno.ChildByFieldName("arguments")
 	if args == nil {
-		return nil
+		return "", false
 	}
-	var out []string
+	last, found := "", false
 	var walk func(n ts.Node)
 	walk = func(n ts.Node) {
 		if n == nil {
 			return
 		}
 		if n.Type() == "string_literal" {
-			out = append(out, stripStringLiteral(nodeText(n, src)))
+			last, found = stripStringLiteral(nodeText(n, src)), true
 			return
 		}
 		for i := 0; i < int(n.ChildCount()); i++ {
@@ -450,7 +465,7 @@ func annotationStringLiterals(anno ts.Node, src []byte) []string {
 		}
 	}
 	walk(args)
-	return out
+	return last, found
 }
 
 // stripStringLiteral removes the surrounding quotes from a Java string
--- a/internal/engine/spring_routes.go
+++ b/internal/engine/spring_routes.go
@@ -144,9 +144,14 @@ func applySpringRouteComposition(args DetectorPassArgs) DetectorPassResult {
 // springHandlerClaimKey builds the key for claimedHandlerEdges: the bare
 // annotation path the YAML rules matched, paired with the handler method name.
 // Both halves are required — see the claimedHandlerEdges doc comment (#6498).
-// The separator is a NUL byte, which cannot occur in a Java identifier or in a
-// route path captured by the spring_mvc.yaml regexes, so the pair cannot be
-// forged by a path/method combination that concatenates ambiguously.
+//
+// The two halves are separated rather than concatenated, because bare
+// concatenation is ambiguous: ("/ab","c") and ("/a","bc") both flatten to
+// "/abc", which would let one class's claim swallow a sibling's only edge.
+// TestDetect_SpringRoute_ConcatAmbiguousClaimKey pins exactly that. The
+// separator must be a byte that appears in neither half; NUL is one such byte
+// for the halves this pass produces, but the test observes the separation, not
+// the choice of NUL.
 func springHandlerClaimKey(barePath, methodName string) string {
 	return barePath + "\x00" + methodName
 }
@@ -271,7 +276,15 @@ func processSpringClass(class ts.Node, src []byte, path string, out *composedSpr
 			composedPath := joinRoutePaths(prefix, apath)
 
 			out.claimedMethodPaths[apath] = true
+			// #6702/F1 — claim every literal the YAML relationship regex could
+			// have captured, not just the one the AST picked as the path.
+			// `apath` itself is among them for every annotation that has a
+			// path argument; it is claimed explicitly so the `apath == ""`
+			// (bare `@GetMapping`) case still registers.
 			out.claimedHandlerEdges[springHandlerClaimKey(apath, methodName)] = true
+			for _, lit := range annotationStringLiterals(a, src) {
+				out.claimedHandlerEdges[springHandlerClaimKey(lit, methodName)] = true
+			}
 
 			routeProps := map[string]string{
 				"framework":    "java",
@@ -397,6 +410,47 @@ func annotationNameAndPath(anno ts.Node, src []byte) (string, string) {
 		return name, byKey
 	}
 	return name, positional
+}
+
+// annotationStringLiterals returns EVERY string literal in the annotation's
+// argument list, in source order, including ones nested inside an
+// array_initializer (`value = {"/a", "/b"}`).
+//
+// #6702/F1 — the AST side and the YAML side disagree about which literal is
+// the path: annotationNameAndPath prefers the `value=`/`path=` pair, while the
+// spring_mvc.yaml relationship regex (`[^)]*["']([^"'\n\r]+)["'][^)]*\)`) is
+// greedy and captures the LAST literal in the argument list. On
+// `@PostMapping(value = "/things", consumes = "application/json")` that is the
+// media type, so the YAML edge is `Route:application/json -> Controller:create`.
+// Claiming only the AST's preferred literal would leave that edge unsuppressed
+// — dangling on both ends, since no `Route:application/json` entity exists (the
+// entity rule anchors on the FIRST literal) and `Controller:` stubs resolve to
+// nothing (#6429). Claiming every literal the regex could land on restores
+// suppression without giving up the pair key that #6498 needs.
+func annotationStringLiterals(anno ts.Node, src []byte) []string {
+	if anno == nil {
+		return nil
+	}
+	args := anno.ChildByFieldName("arguments")
+	if args == nil {
+		return nil
+	}
+	var out []string
+	var walk func(n ts.Node)
+	walk = func(n ts.Node) {
+		if n == nil {
+			return
+		}
+		if n.Type() == "string_literal" {
+			out = append(out, stripStringLiteral(nodeText(n, src)))
+			return
+		}
+		for i := 0; i < int(n.ChildCount()); i++ {
+			walk(n.Child(i))
+		}
+	}
+	walk(args)
+	return out
 }
 
 // stripStringLiteral removes the surrounding quotes from a Java string
--- a/internal/engine/spring_routes.go
+++ b/internal/engine/spring_routes.go
@@ -59,10 +59,19 @@ type composedSpringRoutes struct {
 	// drop the duplicate orphan Route entities the YAML rules emitted for
 	// the same paths inside the same controller class.
 	claimedMethodPaths map[string]bool
-	// claimedHandlerMethods is the set of handler method names whose
-	// ROUTES_TO edge this pass replaced. Used to drop the orphan YAML
+	// claimedHandlerEdges is the set of YAML ROUTES_TO edges this pass
+	// replaced, keyed by the QUALIFYING PAIR (bare annotation path, handler
+	// method name) — see springHandlerClaimKey. Used to drop the orphan YAML
 	// ROUTES_TO edges that point at uncomposed source Routes.
-	claimedHandlerMethods map[string]bool
+	//
+	// #6498 — this was keyed on the bare method name alone and scoped to the
+	// file, so two controller classes in one file sharing a handler method
+	// name collided: the first class's claim suppressed the second class's
+	// legitimate edge, silently deleting a real endpoint from the graph. The
+	// YAML edge carries no class (its ToID is `Controller:<bareMethod>`), so
+	// the class-distinguishing information available on BOTH sides is the
+	// annotation path the edge's FromID (`Route:<barePath>`) is built from.
+	claimedHandlerEdges map[string]bool
 	// claimedClassPrefixes is the set of class-level @RequestMapping
 	// prefixes consumed (e.g. "/api"). Used to drop the orphan class-level
 	// Route entity the YAML rules emit from the bare class annotation.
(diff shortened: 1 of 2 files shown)


cajasmota:
Addressed at `485325e096f7d4e4e86a4e8e9b9876aacbfeeb36`. PR body rewritten with the full detail; summary:

**F1 (blocking) — confirmed and fixed.** Reproduced your exact shape as a red control before touching the code:

```
--- FAIL: TestDetect_SpringRoute_NamedArgsWithMediaTypeLiteral (0.03s)
    spring_routes_issue6498_test.go:251: UNEXPECTED ROUTES_TO edge: Route:application/json -> Controller:create
          got edges: [Route:/api/things -> SCOPE.Operation:ProbeController.create Route:/api/widgets -> SCOPE.Operation:ProbeController.read Route:application/json -> Controller:create Route:text/plain -> Controller:read]
    spring_routes_issue6498_test.go:251: UNEXPECTED ROUTES_TO edge: Route:text/plain -> Controller:read
```

Fixed as you suggested: `annotationStringLiterals` returns every string literal in the annotation's argument list (recursively, so `value = {"/a","/b"}` is covered too) and each is claimed against the enclosing method.

I bounded the widening rather than leaving it open: the loop sits **after** the `verbAnnotations` check, so only the verb annotation's literals are claimed. `TestDetect_SpringRoute_NonVerbAnnotationLiteralsAreNotClaimed` pins that — an `@Operation(summary = "/probe")` on a mapped handler must not swallow a sibling controller's real `/probe` edge. Moving the loop above the check is mutant 7 and it dies on that test. Without the bound, the F1 fix would have been a fresh route back into #6498.

**F2 — weakened, not re-absolutised.** The comment now says the halves are *separated rather than concatenated*, points at `ConcatAmbiguousClaimKey` as the thing that observes it, and names NUL as one adequate choice instead of an unforgeable one. You are right that `$` and `_` break the old claim.

**Corpus measurement — it has no power here, and I would rather say so than quote a reassuring zero.** All 645 Java files: 20 Spring mapping annotations total (spring-petclinic 18, awesome-compose 2), **all positional single-literal**, zero using `value=`/`path=`, zero mentioning `produces`/`consumes`. So the F1 shape occurs 0 times — but only because the corpus contains no Spring code using named arguments at all. It cannot tell us the shape is rare in the wild. What it does establish: this PR is a no-op on 20/20 corpus annotations, since with one literal the AST's pick *is* the last literal.

**Mutants.** Re-scored 1, 3 and 5 on `485325e` (all touch the key F1 feeds) — all still DEAD. 6 and 7 are new and cover the F1 code. Mutant 2 left at its original score: F1 only adds entries to a set whose *reader* mutant 2 deletes, so the change cannot affect it, and the two new tests can only strengthen it.

**Follow-ups filed** with the reasoning preserved, both referencing #6498: #6703 (entity/relationship filter divergence — includes why `claimedMethodPaths` cannot be pair-keyed: a `Route` record carries no method) and #6704 (`django_routes.go` audit — notes it keys on the path half, the mirror of #6498, and that `isDRFGlobalRegisterName` suppresses by name shape rather than by a recorded claim, which is the permissive direction).

**One new thing found while measuring F1, not fixed:** array-valued `@GetMapping(value = {"/a", "/b"})` loses both paths and composes to the bare class prefix — `annotationNameAndPath` only reads `element_value_pair` values that are directly a `string_literal`. I checked it on both sides before assuming: parent and HEAD emit the identical single edge `Route:/api -> SCOPE.Operation:ArrController.multi`, so it is pre-existing and unchanged by this PR. Same silent-recall family as #6498 though, and it should get its own issue.

Not merged, CI not waited on.

cajasmota:
Both findings confirmed and fixed at `9306c74675b47f843fe3c9a9f7ff8485760df974`. You were right on both counts, and B in particular was me re-creating #6498 inside the fix for #6498.

**Finding B — red control on `485325e`, before touching the code:**

```
--- FAIL: TestDetect_SpringRoute_ArrayValueClaimsOnlyTheReachableLiteral (0.02s)
    spring_routes_issue6498_test.go:394: MISSING ROUTES_TO edge: Route:/a -> Controller:multi
          got edges: [Route:/api -> SCOPE.Operation:ArrController.multi]
```

**Finding A — red control is mutant R on the fixed tree:**

```
--- FAIL: TestDetect_SpringRoute_LiteralClaimIsBoundToItsOwnMethod (0.02s)
    spring_routes_issue6498_test.go:348: MISSING ROUTES_TO edge: Route:/alpha -> Controller:two
```

**The fix is the one you specified**, in one place: `annotationStringLiterals` (every literal) became `annotationLastStringLiteral` (the one the greedy regex can actually land on). `consumes="application/json"` *is* the last literal, so F1 stays dead; `/a` in `value = {"/a","/b"}` is no longer claimed; the `@Operation` bound is untouched.

The array fixture pins both halves as you asked — `/a` must survive **and** `/b`'s YAML twin must still be suppressed. Mutant S (recursion into the initializer disabled) fails it from the other side with `UNEXPECTED ROUTES_TO edge: Route:/b -> Controller:multi`, so neither over- nor under-claiming can slip through it now. R and S are mirrors: R deletes a real edge, S admits a bogus one.

**Mutants re-scored on `9306c74`: 1, 6, 7, R, S — all DEAD.**

| # | Result | Evidence |
|---|---|---|
| 1 | DEAD | `MISSING ... Route:/reports -> Controller:list` |
| 6 | DEAD | `UNEXPECTED ... Route:application/json -> Controller:create`, `Route:text/plain -> Controller:read` |
| 7 | DEAD | `MISSING ... Route:/probe -> Controller:check` |
| R | DEAD | `MISSING ... Route:/alpha -> Controller:two` (ALIVE on `485325e`, full suite green) |
| S | DEAD | `UNEXPECTED ... Route:/b -> Controller:multi` |

**2, 3, 4 and 5 not re-run**, and the reason is structural rather than a budget call: each is killed by test 1, 2 or 3, whose annotations carry exactly **one** string literal — the case where "last literal", "every literal" and "the AST's preferred literal" are the same string. The narrowing cannot change the claim set those fixtures produce, so it cannot move those verdicts. Mutant 2 is doubly unaffected: it deletes the *reader* of the claim set, so no change to the set's contents can reach it (and you re-ran it anyway — DEAD, 7 failures).

**Prose nit fixed.** The comment now names `element_value_array_initializer` and explains that the walk is generic *so that* it reaches that node without depending on the name — which is also why your first mutant attempt was a no-op. Worth flagging as a trap: a node-type-named mutant against a generic walk is unfalsifiable by construction.

`internal/engine` + `internal/quality/...` green, `gofmt -l ./internal/` clean. Not merged, CI not waited on. #6706 kept out of this PR as the path-extraction side; finding B was the suppression side.

cajasmota:
Fixed at `588b4b761af9e01a83af6c9a54eb034e36207341`. **I took (a) — deleted the line — but the reason turned out to be stronger than "redundant".**

Your regex argument is right and I verified it against `spring_mvc.yaml` rather than taking it on trust: greedy `[^)]*` before the capture, cannot cross `)`, so the edge's `FromID` is always `Route:<last literal>` and the filter can only ever look the last literal up. `apath` is consulted only when it *is* the last literal — exactly redundant. That is why your mutant survived.

Before deleting I tried to build the (b) fixture — an input where `apath` is consulted and the literal claim is not. I could not, which by your own framing settles it for (a). But the attempt turned up something else: **when the two differ, the `apath` claim is not inert, it deletes a real edge.** `name` is a genuine `@RequestMapping` attribute, so:

```java
@GetMapping(path = "/x", name = "foo")   // AST path = /x, regex capture = foo
```

claims `/x` — which the regex never emitted for this handler — and swallows an unmapped sibling controller's only edge. Measured on identical source with only that line differing:

```
with the apath claim:   [Route:/api/x -> SCOPE.Operation:NamedController.m]
without it:             [Route:/api/x -> SCOPE.Operation:NamedController.m,
                         Route:/x -> Controller:m]        <-- sibling's only edge
```

That is #6498 for the third time in this PR, and the third time it came in through something I added to fix it. The pattern is consistent enough to be worth naming: every regression here has been the claim set growing past what the YAML rules actually emit. The invariant that holds is "claim exactly what the regex can capture, once, bound to the handler that carries it" — findings A, B and now this are all the same rule violated on a different axis.

So it is (a) **plus a fixture**, because the deletion turned out to be observable after all — just in the opposite direction from the one (b) was looking for. Mutant T (re-add the line) is now DEAD:

```
--- FAIL: TestDetect_SpringRoute_ApathIsNotClaimedWhenItIsNotTheLastLiteral (0.02s)
    spring_routes_issue6498_test.go:452: MISSING ROUTES_TO edge: Route:/x -> Controller:m
```

**Comment fixed, not reworded.** You were right that the old justification was refuted by the rules it appealed to — a bare `@GetMapping` matches no pattern and emits no edge, so there is nothing to suppress and the stated case cannot arise. It now carries the regex argument and names the test.

**Re-scored: 6 and T only.** Mutant 6's premise genuinely changed — it now deletes the *sole* claim in the pass, and went from 2 failing tests to **10**, which is the positive evidence that the one remaining line carries everything. Nothing else is affected: the deleted line was redundant on every single-literal annotation, and every fixture except the array and named-argument ones is single-literal, so no other claim set moved.

`internal/engine` + `internal/quality/...` green, `gofmt -l ./internal/` clean. Not merged, CI not waited on.

Issue #6481, opened 2026-08-21
import placeholders emitted with no recognisable Subtype are indexed as real declarations, so #6369's fix never reaches those languages


cajasmota:
Split out of #6369 after fixing the F# instance (PR #6480). **#6369's fix does not reach ~26 other languages, for the same reason it did not reach F#.**

## The shape

#6427 made an import placeholder rank below a real declaration in the repo-wide by-name index, keyed on `isImportPlaceholderKind` reading the record's `Subtype`. An extractor that emits its per-import placeholder **without a recognisable `Subtype`** never joins that family, so the placeholder is still indexed as though it were a real declaration and the #6369 defect stays live: one import whose last segment collides with a type name flips that name ambiguous and drops every bare-name edge to it **repo-wide**, including in files that import nothing.

F# was found this way. It is not special.

## Shape (B) — synthetic per-import `SCOPE.Component`, no `Subtype`

golang `extractor.go:2451` · rust `rust.go:~929` · csharp `:485` · scala `:897` · groovy `:1176` · dart `:344` · elixir `:444` · clojure `:538` · php `:485` **and** `:650` · lua `:535` · shell `:255` · erlang `:412`+`:448` · assembly `:482` · verilog `:652`+`:690` · vhdl `:287`+`:319` · pony `:331` · lisp `:641`

Note golang confirms a note already sitting in `internal/resolve/imports.go`.

**Plus the ten sharing the very same `buildImportEntities` template F# used** — haskell `:460`, ocaml `:346`, reasonml `:311`, rescript `:282`, sml `:401`, idris `:363`, elm `:317`, nim `:335`, zig `:258`.

**That template group is the highest risk after go and rust**, because it names the placeholder after the **bare last segment** — the exact collision shape #6369 describes. The others vary.

## Sub-class (B′) — stamps `Subtype: "module"`

ruby `:385`, crystal `:155`, swift `:641`. Whether `"module"` should join the marker family is a **design decision, not a bug fix** — decide it before touching them. swift is mitigated regardless: its `Name` is namespaced, so it cannot collide.

Ambiguous: `html/extractor.go:639` (`Subtype: "style_include"`) is structurally (B), but its names are asset hrefs rather than identifiers, so the collision may be unreachable. Check before changing.

## Clean (C) — no placeholder; IMPORTS hangs on the file carrier

java, python, javascript, solidity, yaml, hcl, astro, dockerfile, task, cobol, vbnet, bicep.

[...]

PR #6711, opened 2026-08-26
fix(#6481): A1 — elm/haskell/ocaml import placeholders were invisible to the #6427 marker

cajasmota:
Arm **A1** of #6481 only — elm, haskell, ocaml. Three emission sites, three lines of production change. No other language is touched.

## The defect

`internal/resolve/refs.go:1602-1604` defines the import-placeholder marker:

```go
kind == "SCOPE.Component" && subtype == "import"
```

consumed at `refs.go:1589`, `symbol_index.go:209`, `symbol_index.go:448`. #6427 taught `BuildIndex` that a record carrying that marker is **not** a declaration of the name it holds, so a real declaration outranks it instead of flipping the name AMBIGUOUS and dropping every bare-name edge to it.

`buildImportEntities` in these three extractors minted its per-import record with `Kind: "SCOPE.Component"`, no `StartLine`/`EndLine`, a single `IMPORTS` relationship — and **no `Subtype` at all**. The predicate therefore never recognised it, so #6369's fix never reached elm/haskell/ocaml. All three name the placeholder by the module's **last segment** (`importDisplayName`), the collision-prone shape: one `import Acme.Animal` anywhere in the repo collided with a real type named `Animal` and silently deleted every bare-name edge to it — repo-wide, no flag, nothing reporting it.

| lang | site |
|---|---|
| elm | `internal/extractors/elm/extractor.go:315` |
| haskell | `internal/extractors/haskell/extractor.go:458` |
| ocaml | `internal/extractors/ocaml/extractor.go:344` |

## The test is the work

A golden fixture **cannot** express this: `Subtype` on a bodiless stub is not surfaced in golden output (#6488), so a fixture passes byte-identically before and after the stamp. It has to be asserted on the extractor's emitted records.

One test file per language, in that language's own package — **not** one table-driven test, so a language that emits nothing cannot hide behind its neighbours (proved by mutant M1 below).

Each file asserts on the **emitted artefact**, never a count:

- the placeholder's `Kind` AND `Subtype` AND `Name` AND the module its `IMPORTS` edge targets;
- membership is decided by *carrying an `IMPORTS` edge*, i.e. independently of the field under test, so the sweep cannot be satisfied by the very thing it asserts.

### Non-vacuity (mandatory, and the whole point)

Each test compares an expected `name -> module` table for **exact equality**:

| lang | expected placeholders |
|---|---|
| elm | `Animal`→`Acme.Animal`, `Attributes`→`Html.Attributes`, `Browser`→`Browser` |
| haskell | `Animal`→`Acme.Animal`, `Strict`→`Data.Map.Strict`, `Prelude`→`Prelude` |
| ocaml | `Animal`→`Acme.Animal`, `Tree`→`Base.Map.Tree`, `Stdio`→`Stdio` |

Each table carries a 1-segment row (covers `importDisplayName`'s no-dot branch) and a 3-segment row (the shape that actually occurs). An extractor emitting **zero** placeholders fails here rather than passing trivially — that trivially-passing shape is exactly what let this drift unnoticed across 25 languages. The neighbouring-axis test carries its own independent non-vacuity fatal (`sawMarked == 0`).

### Neighbouring axis (varied vs held constant)

Each fixture declares real entities through *separate* code paths that already carry their own `Subtype`, and the probe deliberately makes the collision real — the placeholder `Animal` and a real declaration named `Animal` (or `animal`) coexist in the same file:

| lang | real declarations pinned |
|---|---|
| elm | `Domain.Core`/`module`, `Animal`/`type`, `view`/`function` |
| haskell | `Domain.Core`/`module`, `Animal`/`data`, `speak`/`function` |
| ocaml | `Animal`/`module`, `animal`/`type`, `speak`/`function` |

Two controls:

1. an **exclusivity sweep** — nothing may carry `Subtype: "import"` without an `IMPORTS` edge (the marker *demotes* whatever carries it out of the by-name index, so stamping it on a real declaration is the same defect with the sign flipped; it bites hardest in OCaml, where a `module` declaration is precisely what an `open` targets);
2. **positive** per-declaration subtype assertions, so the control cannot be satisfied by the declaration simply disappearing.

**Axes varied:** segment count of the imported module (1 / 2 / 3), the language, and whether a name collides with a real declaration. **Axes held constant:** `Kind` is `SCOPE.Component` throughout, one probe file per language, and the marker value itself.

## Red control (verbatim, before the production change)

```
--- FAIL: TestElmImportEmitsMarkedImportPlaceholder_6481 (0.00s)
    import_placeholder_6481_test.go:95: import placeholder "Animal": Kind="SCOPE.Component" Subtype="", want SCOPE.Component/import (resolve.isImportPlaceholderKind will not recognise this record, so it stays in the by-name index as a declaration and flips "Animal" ambiguous)
    import_placeholder_6481_test.go:95: import placeholder "Attributes": Kind="SCOPE.Component" Subtype="", want SCOPE.Component/import (...)
    import_placeholder_6481_test.go:95: import placeholder "Browser": Kind="SCOPE.Component" Subtype="", want SCOPE.Component/import (...)
--- FAIL: TestElmRealDeclarationsDoNotAcquireImportMarker_6481 (0.00s)
    import_placeholder_6481_test.go:173: vacuous: no entity carries Subtype="import" at all
FAIL	github.com/cajasmota/grafel/internal/extractors/elm	1.588s
--- FAIL: TestHaskellImportEmitsMarkedImportPlaceholder_6481 (0.00s)
    import_placeholder_6481_test.go:92: import placeholder "Animal": Kind="SCOPE.Component" Subtype="", want SCOPE.Component/import (...)
--- FAIL: TestOCamlOpenEmitsMarkedImportPlaceholder_6481 (0.00s)
    import_placeholder_6481_test.go:94: import placeholder "Animal": Kind="SCOPE.Component" Subtype="", want SCOPE.Component/import (...)
```

It fails because `Subtype=""`, not incidentally: the expected-table equality check and every real-declaration lookup both passed at red, so the fixtures were already producing exactly 3 placeholders and all 3 real declarations before the fix.

## Mutation score

All mutants applied byte-level, restored by `cp` and verified by `shasum`; never `git checkout --`. `go vet` (or `go build`) first, `-count=1` always, and the applied mutant confirmed by an occurrence count before every run.

| # | mutant | verdict | output |
|---|---|---|---|
| M1 | revert the stamp at **one site only** (elm) | **DEAD**, and isolated | `FAIL .../elm 0.658s` with `import placeholder "Animal": … Subtype=""`; `ok .../haskell 0.250s`, `ok .../ocaml 0.454s` — proves the three tests are independent, not one test carrying all three |
| M2 | **permissive:** stamp `Subtype: "import"` on the real-declaration path (the module declaration) in all three | **DEAD** | `entity "Domain.Core" (Kind="SCOPE.Component") is marked Subtype="import" but carries no IMPORTS edge — the placeholder marker landed on a real declaration, demoting it out of the by-name index` (elm + haskell), and `entity "Animal" …` (ocaml). All three `RealDeclarationsDoNotAcquireImportMarker` tests fail |
| M3a | `buildImportEntities` returns nil (extractor emits nothing), tests intact | **DEAD** | `vacuous or wrong: got 0 import placeholders map[], want 3 map[Animal:Acme.Animal Attributes:Html.Attributes Browser:Browser]` + `vacuous: no entity carries Subtype="import" at all`, in all three packages |
| M3b | M3a **plus** delete the non-vacuity assertions from the elm test | **passes (`ok .../elm 0.448s`)** — the control proving the non-vacuity assertion is load-bearing, not decoration. Restored immediately |

Nothing recorded as *equivalent under the current suite*; all three requested mutants were killed by portable fixtures and no fixture was manufactured to force a kill.

## Verification

- `go test -count=1 ./internal/extractors/elm/ ./internal/extractors/haskell/ ./internal/extractors/ocaml/` — all three packages green (full package, not just the new tests).
- `go vet` on all three, `gofmt -l` clean.
- `go test -count=1 ./internal/quality/` — **`ok … 30.047s`**. **No golden figure moved**; `baseline.json` and every `expected.json` are untouched. (`haskell-warp-mini` is the only golden fixture in these three languages; its three `IMPORTS` rows are `nice_to_have` and still score the same.)

## Found but NOT fixed (out of A1 scope)

Stamping the marker makes `resolve.PruneImportPlaceholders` start *recognising* these records. `placeholderModuleSpecifier`'s precedence is `import_module` > `module` > `QualifiedName` > `Name`, and these three placeholders carry **none** of the first three — so the #6156 restore would rewrite an incoming edge to the bare **last segment** (`Generic`) rather than the full dotted specifier (`System.Collections.Generic`). This is the same gap #6369's F# arm closed by adding `Properties["import_module"]`. It is not reachable from any current fixture (nothing binds an incoming edge to these placeholders today) and `internal/quality` is unaffected, but the same fix — carrying the specifier on `Properties["import_module"]`, and **not** on `Properties["module"]` (module-rollup label) or `QualifiedName` (probed ahead of `byName`) — will be wanted for these three languages as a follow-up. Flagged rather than done, because A1 is scoped to the marker alone.

Closes part of #6481 (arm A1).

Refs #6481 — arm A1 only (elm, haskell, ocaml). This does NOT close the issue: nine other groups remain, php/erlang/verilog/vhdl have two emission sites each, and `erlang/extractor.go:446` must never be stamped.

### DIFF
--- a/internal/extractors/elm/extractor.go
+++ b/internal/extractors/elm/extractor.go
@@ -318,6 +318,14 @@ func buildImportEntities(filePath string, imports []string) []types.EntityRecord
 			Subtype:    "import",
 			SourceFile: filePath,
 			Language:   "elm",
+			// The FULL dotted specifier, for resolve.placeholderModuleSpecifier
+			// (#6156). Name is only the last segment, so without this the prune
+			// restore would rename the external node to that bare segment. It
+			// must NOT travel on Properties["module"] (the module-rollup label,
+			// which EnsureModule and stampModuleOnEntities both treat as
+			// authoritative) nor on QualifiedName (probed ahead of byName and
+			// never given #6427's placeholder precedence).
+			Properties: map[string]string{"import_module": mod},
 			Relationships: []types.RelationshipRecord{
 				{
 					FromID: filePath,
--- a/internal/extractors/elm/extractor.go
+++ b/internal/extractors/elm/extractor.go
@@ -315,6 +315,7 @@ func buildImportEntities(filePath string, imports []string) []types.EntityRecord
 		out = append(out, types.EntityRecord{
 			Name:       displayName,
 			Kind:       "SCOPE.Component",
+			Subtype:    "import",
 			SourceFile: filePath,
 			Language:   "elm",
 			Relationships: []types.RelationshipRecord{
--- a/internal/extractors/elm/import_placeholder_6481_test.go
+++ b/internal/extractors/elm/import_placeholder_6481_test.go
@@ -46,11 +46,22 @@ const elmProbeFile6481 = "src/Domain/Core.elm"
 // `Html.Attributes` is a 3-segment import, the shape that actually occurs, and
 // `Browser` is single-segment so importDisplayName's no-dot branch is covered
 // too.
+//
+// EVERY real-declaration path the extractor has is represented — module, `type
+// alias` (subtype "typealias"), custom `type`, and a function. That is not
+// padding: the marker DEMOTES whatever carries it, so a construct absent from
+// this fixture is a construct on which a Subtype:"import" mutant survives
+// unobserved. Review-measured on the sibling Haskell arm: the type-synonym
+// site was stamped `"import"` and lived through the entire package suite,
+// purely because no fixture declared one.
 const elmProbeSrc6481 = `module Domain.Core exposing (..)
 
 import Acme.Animal
 import Html.Attributes exposing (class, id)
 import Browser
+
+type alias Flags =
+    { name : String }
 
 type Animal
     = Dog
@@ -64,7 +75,9 @@ view model =
 // points at, and whether the record carries such an edge at all. Carrying an
 // IMPORTS edge is what makes a record an import placeholder independently of
 // the marker under test — so the sweep below cannot be satisfied by the very
-// field it is asserting.
+// field it is asserting. Enumerated against the extractor's full output: no
+// real declaration emits an IMPORTS edge (module -> CONTAINS, function ->
+// CALLS, type/typealias -> none), so the discriminator is sound.
 func elmImportsTarget6481(e types.EntityRecord) (string, bool) {
 	for _, r := range e.Relationships {
 		if r.Kind == "IMPORTS" {
@@ -78,7 +91,7 @@ func TestElmImportEmitsMarkedImportPlaceholder_6481(t *testing.T) {
 	ents := runElm(t, elmProbeSrc6481, elmProbeFile6481)
 
 	// display name (the LAST SEGMENT — the collision-prone shape that is the
-	// premise of the whole issue) -> the full module the IMPORTS edge targets.
+	// premise of the whole issue) -> the FULL dotted module.
 	want := map[string]string{
 		"Animal":     "Acme.Animal",
 		"Attributes": "Html.Attributes",
@@ -97,10 +110,41 @@ func TestElmImportEmitsMarkedImportPlaceholder_6481(t *testing.T) {
 				"stays in the by-name index as a declaration and flips %q ambiguous)",
 				e.Name, e.Kind, e.Subtype, e.Name)
 		}
+
+		// THE SPECIFIER CHANNEL (#6156). Now that the marker is stamped,
+		// resolve.PruneImportPlaceholders recognises this record and
+		// buildPlaceholderModuleRestores (imports.go:3502) rewrites incoming
+		// edges to whatever placeholderModuleSpecifier reads. Its precedence is
+		// import_module > module > QualifiedName > Name, so with none of the
+		// first three set it would fall back to the bare last segment and
+		// rename the external node from "Acme.Animal" to "Animal".
+		spec := e.Properties["import_module"]
+		if spec != target {
+			t.Errorf("import placeholder %q: Properties[\"import_module\"]=%q, want the "+
+				"full dotted module %q — the #6156 restore will otherwise record the "+
+				"bare display Name", e.Name, spec, target)
+		}
+		// The specifier must NOT travel on Properties["module"]: that key is
+		// the module-rollup label, and BOTH module.EnsureModule and
+		// stampModuleOnEntities treat an extractor-supplied value as
+		// authoritative, so a dotted specifier there fabricates a Module node
+		// on the never-pruning incremental path.
+		if got, ok := e.Properties["module"]; ok {
+			t.Errorf("import placeholder %q pre-stamps Properties[\"module\"]=%q; that key "+
+				"is the path-derived module-rollup label, not an import specifier", e.Name, got)
+		}
+		// Nor on QualifiedName: byQualifiedName is probed BEFORE byName and
+		// never received #6427's placeholder precedence, so the placeholder
+		// would shadow the very module it points at.
+		if e.QualifiedName != "" {
+			t.Errorf("import placeholder %q sets QualifiedName=%q; byQualifiedName is probed "+
+				"ahead of byName and has no placeholder precedence", e.Name, e.QualifiedName)
+		}
+
 		if _, dup := got[e.Name]; dup {
 			t.Errorf("duplicate import placeholder for %q", e.Name)
 		}
-		got[e.Name] = target
+		got[e.Name] = spec
 	}
 
 	// NON-VACUITY. Exact equality, not a subset and not a count floor: an
@@ -111,7 +155,7 @@ func TestElmImportEmitsMarkedImportPlaceholder_6481(t *testing.T) {
 	}
 	for name, mod := range want {
 		if got[name] != mod {
-			t.Errorf("import placeholder %q: IMPORTS target = %q, want %q", name, got[name], mod)
+			t.Errorf("import placeholder %q: specifier = %q, want %q", name, got[name], mod)
 		}
 	}
 }
@@ -133,23 +177,29 @@ func elmFindBySubtype6481(ents []types.EntityRecord, name, subtype string) *type
 // index, so stamping it on a real declaration deletes that declaration's own
 // name instead of protecting it — the same defect with the sign flipped.
 //
-// Without this control, a mutant that stamps Subtype:"import" on every
-// emission path would satisfy the sweep above and survive.
+// Without this control, a mutant that stamps Subtype:"import" on an emission
+// path would satisfy the sweep above and survive.
 func TestElmRealDeclarationsDoNotAcquireImportMarker_6481(t *testing.T) {
 	ents := runElm(t, elmProbeSrc6481, elmProbeFile6481)
 
-	// The exclusivity sweep: nothing may carry the marker except a record that
-	// actually stands in for an import.
+	// The exclusivity sweep: nothing may carry the marker, or the specifier
+	// key, except a record that actually stands in for an import.
 	sawMarked := 0
 	for _, e := range ents {
-		if e.Subtype != "import" {
-			continue
+		_, isImport := elmImportsTarget6481(e)
+		if e.Subtype == "import" {
+			sawMarked++
+			if !isImport {
+				t.Errorf("entity %q (Kind=%q) is marked Subtype=\"import\" but carries no "+
+					"IMPORTS edge — the placeholder marker landed on a real declaration, "+
+					"demoting it out of the by-name index", e.Name, e.Kind)
+			}
 		}
-		sawMarked++
-		if _, isImport := elmImportsTarget6481(e); !isImport {
-			t.Errorf("entity %q (Kind=%q) is marked Subtype=\"import\" but carries no "+
-				"IMPORTS edge — the placeholder marker landed on a real declaration, "+
-				"demoting it out of the by-name index", e.Name, e.Kind)
+		if spec, ok := e.Properties["import_module"]; ok && !isImport {
+			t.Errorf("entity %q (Kind=%q subtype=%q) carries Properties[\"import_module\"]=%q "+
+				"but is not an import placeholder; no layer filters unknown property keys, so "+
+				"that false provenance claim is persisted into the graph, rendered into docgen "+
+				"output and served in MCP payloads", e.Name, e.Kind, e.Subtype, spec)
 		}
 	}
 	if sawMarked == 0 {
@@ -157,9 +207,11 @@ func TestElmRealDeclarationsDoNotAcquireImportMarker_6481(t *testing.T) {
 	}
 
 	// Each real declaration keeps its OWN subtype, asserted positively so the
-	// check cannot be satisfied by the declaration disappearing.
+	// check cannot be satisfied by the declaration disappearing. One row per
+	// declaration path the extractor has.
 	for _, want := range []struct{ name, subtype string }{
 		{"Domain.Core", "module"},
+		{"Flags", "typealias"},
 		{"Animal", "type"},
 		{"view", "function"},
 	} {
--- a/internal/extractors/elm/import_placeholder_6481_test.go
+++ b/internal/extractors/elm/import_placeholder_6481_test.go
@@ -1,0 +1,175 @@
+package elm_test
+
+// #6481 arm A1 (elm) — the import placeholder must carry the marker the
+// resolver keys on.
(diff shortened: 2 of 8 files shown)


cajasmota:
## Review round 2 — addressed at `1b89dd6a`

Both asks are in. The three `Subtype: "import"` stamps are unchanged, so **M1 / M2 / M3a / M3b stand as scored**.

### 1. The surviving type-synonym mutant — now DEAD

`internal/extractors/haskell/extractor.go:321`, `Subtype: "type"` → `"import"`:

```
--- FAIL: TestHaskellRealDeclarationsDoNotAcquireImportMarker_6481 (0.00s)
    import_placeholder_6481_test.go:201: entity "Alias" (Kind="SCOPE.Component") is marked Subtype="import" but carries no IMPORTS edge — the placeholder marker landed on a real declaration, demoting it out of the by-name index
    import_placeholder_6481_test.go:231: no entity "Alias" with Subtype="type"; the real-declaration paths are not emitting, so this control is vacuous
FAIL	github.com/cajasmota/grafel/internal/extractors/haskell	0.592s
```

Killed by the exclusivity sweep exactly as predicted, plus a second independent failure from the positive row.

### …and the rest of the fixture space, enumerated rather than hand-picked

One hand-picked construct would have left the same hole open elsewhere, so **every** declaration path in all three extractors is now represented in its probe fixture with a positive row in the real-declaration table:

| lang | added to fixture | new row |
|---|---|---|
| elm | `type alias Flags = { name : String }` | `{"Flags", "typealias"}` |
| haskell | `type Alias = Int` | `{"Alias", "type"}` |
| haskell | `newtype Wrapper = Wrapper Int` | `{"Wrapper", "newtype"}` |
| haskell | `class Speaker a where …` | `{"Speaker", "typeclass"}` |
| haskell | `instance Speaker Animal where …` | `{"Speaker Animal", "instance"}` |
| ocaml | — (module / type / let were already all three of its paths) | — |

Then **all 14 declaration sites across the three extractors were swept with the permissive stamp, one at a time** — apply, assert the file changed, `go vet`, `go test -count=1`, restore by `cp`, verify with `cmp`:

```
elm:147     (module)     -> DEAD   entity "Domain.Core"  is marked Subtype="import" but carries no IMPORTS edge
elm:175     (typealias)  -> DEAD   entity "Flags"
elm:211     (type)       -> DEAD   entity "Animal"
elm:245     (function)   -> DEAD   entity "view"
haskell:167 (module)     -> DEAD   entity "Domain.Core"
haskell:215 (function)   -> DEAD   entity "speak"
haskell:260 (data)       -> DEAD   entity "Animal"
haskell:289 (newtype)    -> DEAD   entity "Wrapper"
haskell:321 (type)       -> DEAD   entity "Alias"          <- the reviewer's survivor
haskell:351 (typeclass)  -> DEAD   entity "Speaker"
haskell:389 (instance)   -> DEAD   entity "Speaker Animal"
ocaml:158   (module)     -> DEAD   entity "Animal"
ocaml:210   (function)   -> DEAD   entity "speak"
ocaml:245   (type)       -> DEAD   entity "animal"
```

**14/14 DEAD.** Worth recording: `elm:245` first reported **ALIVE** — the patch had silently failed to apply because that literal is aligned `Subtype:       "function",` (extra spaces) and the applier's exact-line assertion tripped. The assert caught it; re-applied correctly it is DEAD, message above. A sweep without an applied-ness check would have banked a false ALIVE, or worse a false DEAD.

The discriminator the sweep rests on was also enumerated, not assumed — dumped from real extractor output, **no** real declaration emits an `IMPORTS` edge (module/data → `CONTAINS`, function → `CALLS`, instance → `IMPLEMENTS`, type/typealias/newtype/typeclass → none). It is stated in the helper doc comment in each file.

### 2. `import_module` — folded in, not deferred

Agreed with the trace and took the recommendation. The placeholder literal now carries:

```go
Properties: map[string]string{"import_module": mod},
```

Reasoning recorded at the site: stamping the marker is what makes `PruneImportPlaceholders` recognise these records at all, so `buildPlaceholderModuleRestores` (`imports.go:3502`) newly reads them, and `placeholderModuleSpecifier`'s `import_module > module > QualifiedName > Name` precedence would otherwise fall through to `importDisplayName`'s bare last segment — renaming the external node `Acme.Animal` → `Animal`. One line, same struct literal, identical to #6369's F# arm.

The assertion was indeed free — the `want` tables already held the full dotted module, so the placeholder test now keys on `Properties["import_module"]` and cross-checks it against the `IMPORTS` edge target.

Both wrong channels are pinned negatively as well, carrying forward #6369's two shipped regressions: the specifier must **not** be on `Properties["module"]` (module-rollup label; `EnsureModule` and `stampModuleOnEntities` both treat an extractor value as authoritative → fabricates a Module node on the never-pruning incremental path) nor on `QualifiedName` (probed ahead of `byName`, no #6427 placeholder precedence → the placeholder shadows the module it points at).

And the exclusivity sweep now covers the new key too, because it is not inert — nothing filters unknown property keys, so it reaches the persisted graph, docgen output and MCP payloads.

**Two new mutants scored:**

| # | mutant | verdict | output |
|---|---|---|---|
| M5 | drop `import_module` from the haskell placeholder literal | **DEAD** | `import placeholder "Animal": Properties["import_module"]="", want the full dotted module "Acme.Animal" — the #6156 restore will otherwise record the bare display Name` (×3, plus 3 specifier-table failures) |
| M6 | **permissive:** stamp `import_module` on the haskell module declaration | **DEAD** | `entity "Domain.Core" (Kind="SCOPE.Component" subtype="module") carries Properties["import_module"]="Domain.Core" but is not an import placeholder; … persisted into the graph, rendered into docgen output and served in MCP payloads` |

### Verification at `1b89dd6a`

```
ok  github.com/cajasmota/grafel/internal/extractors/elm      0.600s
ok  github.com/cajasmota/grafel/internal/extractors/haskell  0.785s
ok  github.com/cajasmota/grafel/internal/extractors/ocaml    1.007s
ok  github.com/cajasmota/grafel/internal/quality            25.367s
```

`go vet` clean on all four, `gofmt -l` clean. **No golden figure moved** — `internal/quality` is green with the `import_module` fold-in in place, and `baseline.json` and every `expected.json` remain untouched. Scope is unchanged: three languages, `erlang:446` and the other nine groups still untouched.

cajasmota:
## CI regression in `internal/feedback` — fixed at `dc271bee`

Reproduced locally, verbatim, before touching anything:

```
--- FAIL: TestNonFieldBearingLanguageCarriersExempt_6536 (0.00s)
    report_classlike_component_6536_test.go:379: premise gone: haskell emitted 1 module carriers and 0 bare-subtype placeholders; this test would exempt nothing
FAIL	github.com/cajasmota/grafel/internal/feedback	1.675s
```

After:

```
ok  	github.com/cajasmota/grafel/internal/feedback	8.805s
```

### Direction taken: assert the marked form, and split the exemption's two halves

**The premise guard was right and is not weakened.** It is a non-vacuity assertion of exactly the shape this PR's own tests rely on, and it caught its fixture population emptying out instead of passing on an exemption that no longer exempts anything.

Reading #6536's intent rather than the failing line: `isFieldExtractionCandidate` (`report.go:966`) checks `nonClassSubtypes[subtype]` **first**, and only then `nonFieldBearingLanguages[language]`. `TestNonFieldBearingLanguageCarriersExempt_6536` covered *two* populations at once, and the `nonFieldBearingLanguages` doc comment says why neither was reachable by the subtype rule:

> …and Haskell's import placeholders, **which set no Subtype at all so the subtype exclusions above cannot see them.**

**That sentence is precisely what this PR makes false**, and the two halves have now diverged:

| half | status after #6481 | exempted by |
|---|---|---|
| import placeholders | **obsolete for haskell** — the exemption existed *only* because they were unmarked | `nonClassSubtypes["import"]`, present since #6536 itself — one step earlier, and independent of the language |
| per-file `module` carriers | **unchanged, still uniquely load-bearing** | `nonFieldBearingLanguages["haskell"]` only — `"module"` is deliberately absent from `nonClassSubtypes` because a VB.NET `Public Module Util` genuinely owns Shared/Private field children |

So the answer to your question 1 is the first branch, for one half only: the placeholders are exempted **more strongly** now, by the same rule that already covered C#'s `cross/imports` placeholders in `TestImportPlaceholdersExemptFromFieldDenominator_6536`. No fixture was retargeted to another language and none was manufactured.

The test now:
- counts `marked` (Subtype `"import"`) in the premise instead of `bare`;
- **asserts `bare == 0`** — the stale counter becomes a live regression pin, so `internal/feedback` now fails if the #6481 stamp is ever reverted;
- pins the two halves **rule-by-rule**, because an obsolete half is a standing invitation to delete the whole entry.

`report.go`'s doc comment asserted the old shape as fact; corrected, and what it claims is now *observed* by the test rather than only stated.

### Answering question 3 plainly

**The exemption is now language-specific in its import half, and it will need revisiting as the other arms land.** `nonFieldBearingLanguages` is deliberately **not** pruned: only haskell, elm, ocaml (A1) and fsharp (#6369) stamp the marker; **idris, reasonml, rescript, crystal, erlang** still emit bare-subtype placeholders and still need the language exemption for *both* halves. That is stated in the test and in `report.go` rather than left for the next arm to rediscover. And no entry may ever be dropped on the strength of its import half, because the `module` carriers depend on it in all nine.

### Mutants on the new assertions

| # | mutant | verdict | output |
|---|---|---|---|
| M7 | revert the haskell `Subtype:"import"` stamp | **DEAD** | `premise gone: haskell emitted 1 module carriers and 0 import placeholders; this test would exempt nothing` — the consumer-side pin fires |
| M8 | drop `"haskell"` from `nonFieldBearingLanguages` (**the deletion the obsolete half invites**) | **DEAD** | `the haskell per-file `module` carrier is in the field-extraction denominator; it is a guaranteed zero-field failure` + `ClassTotal = 1 on a Haskell graph, want 0 … (zero-field rate 100)` |
| M9 | add `"module"` to `nonClassSubtypes` (**the naive alternative fix** — it would have made CI green) | **DEAD** | `"module" entered nonClassSubtypes; that deletes VB.NET's genuinely field-bearing \`Public Module Util\` from the denominator` + `a VB.NET \`Module\` is excluded from the denominator, but it really does own Shared and Private field children` + a third failure in `TestInterfaceAndDelegateExemptModuleCounted_6536` |
| M10 | drop `"import"` from `nonClassSubtypes` | **DEAD** | `"import" left nonClassSubtypes; every import placeholder in every language re-enters the field-extraction denominator` + `a vbnet import placeholder is in the field-extraction denominator` + `a kotlin …` |

M9 is the one worth noting: excluding `"module"` globally is the shortest path to a green CI and it silently deletes a population that genuinely passes. It now fails three ways.

M10 also confirms the belt-and-braces: haskell is absent from its failure list because the language rule still catches it as a fallback — only `vbnet`/`kotlin` are exposed. Both rules are real, and neither is redundant.

### Verification at `dc271bee`

```
ok  github.com/cajasmota/grafel/internal/extractors/elm      4.012s
ok  github.com/cajasmota/grafel/internal/extractors/haskell  5.032s
ok  github.com/cajasmota/grafel/internal/extractors/ocaml    6.601s
ok  github.com/cajasmota/grafel/internal/quality           34.305s
ok  github.com/cajasmota/grafel/internal/feedback           8.805s
```

Plus a risk check on the two packages most likely to observe this change — `internal/resolve` (defines the marker predicate) and `internal/extractors` (registry-level tests over every extractor):

```
ok  github.com/cajasmota/grafel/internal/resolve      1.156s
ok  github.com/cajasmota/grafel/internal/extractors  31.769s
```

`go vet` and `gofmt -l` clean throughout. No golden figure moved; `baseline.json` and every `expected.json` untouched. No extractor was touched by this commit, so the three stamps and the 14-site sweep are unaffected.

### Other packages that observe extractor subtypes

`internal/feedback` is the **only** package outside `internal/extractors/` that directly imports the elm/haskell/ocaml extractor packages (the sole other reference is `registry_gen.go`, registration only). Packages that observe subtypes *generically*, via the registry rather than by import, and so could be exposed by a later arm: `internal/quality`, `internal/extractors`, `internal/resolve`, `internal/coverage`, `internal/links`, `internal/docgen`, `internal/dashboard`, `internal/engine`, `internal/graph`. The first three are the plausible ones and all three are green here.

Issue #6709, opened 2026-08-26
no workflow runs any Python test: ratchet.py's 33-test suite — the quality gate's own guard — has no CI trigger at all


cajasmota:
`scripts/quality/test_ratchet.py` — the repo's **only** Python test file — is never executed by CI. **No workflow runs any Python test at all.**

## The measurement

```
grep -rn "unittest\|pytest\|python3 -m" .github/workflows/    -> 0 hits
find scripts -name "test_*.py"                                -> scripts/quality/test_ratchet.py
```

The only workflow that mentions the ratchet is `quality.yml`, and it is **`workflow_dispatch:` only** — it invokes `scripts/verify2/run-quality.sh` (which `exec`s `scripts/quality/run.sh`), never the unit tests. `test.yml` runs the Go suite and nothing else.

So the file currently holds **33 tests** guarding `ratchet.py` — the component that decides whether a quality regression blocks a merge — and every one of them runs only when a human types the command locally.

## Why this is worth a P2 rather than a shrug

`ratchet.py` is a **gate**. It decides regression-vs-improvement, expectation drift, forbidden hits, missing reports and missing baseline entries. Its unit tests are the only thing standing between a subtle edit and a gate that silently stops objecting — and this repo has already shipped, and had to fix, exactly that shape more than once:

- **#6627** — the durability guard's ref choice and its second early return were both unobserved.
- **#6633** — `QUALITY_BASE_REF`'s `.strip()` was asserted only in a docstring.
- **#6570** — the guard validated a sha against the very ref it was computed from; the fix landed with a suite of new tests that, per this issue, **will never run again automatically**.

That last one is the sharp version of the problem: we keep investing in tests for this file, and the investment does not compound, because nothing re-runs them.

It is a member of the **"gates that only fire when someone asks them to"** family (see the 0.3.1 tracking issue), and arguably the purest instance: not a gate that is hard to trigger, but a test suite with **no trigger at all**.

## What a fix has to establish

Add a `pull_request`-triggered job that runs `python3 -m unittest scripts.quality.test_ratchet` (or the project's preferred runner). Constraints worth stating up front:

[...]

Issue #6586, opened 2026-08-25
the recall ratchet is dispatch-only so nothing gates it per-PR, and quality.yml's header defers that question to #6231, which is closed


cajasmota:
`scripts/quality/run.sh` — the golden-fixture recall ratchet — is run by exactly one workflow, and that workflow is **dispatch-only**. So nothing gates recall per-PR, and the open question about whether it should is **pointing at a closed issue**.

## Grounded on `main` (`55d420669`)

`.github/workflows/quality.yml:1-9`:

```yaml
on:
  # Dispatch-only — so this is NOT a per-PR gate; run
  # `scripts/quality/run.sh --runs 1 --ratchet` locally instead. That is a
  # property of THIS workflow, not a repo convention: cross-platform-compile.yml
  # and windows-installers.yml are deliberately always-on and their headers
  # forbid making them dispatch-only. Whether quality should join them is open
  # (#6231) — the ratchet makes it viable, since it can now actually pass.
  workflow_dispatch:
```

`grep -rln "quality/run.sh" .github/workflows/` returns **only** `quality.yml`. And **#6231 is CLOSED**.

So the header's forward reference — "whether quality should join them is open (#6231)" — names an issue that no longer tracks anything. The decision it defers is now deferred to nobody.

## Why this matters more than one un-run workflow

The ratchet exists so a recall regression surfaces as a moving number without anyone remembering to look. Two things this milestone show what that is worth:

- **#6572** — a full re-record moved **zero** figures across all 26 fixtures the day after five extractors changed. That is either a coverage gap in the fixtures or evidence the ratchet cannot see what those changes altered; it is being investigated. Either way it was only discovered because a re-record happened to be run by hand.
- **#6564** — `internal/quality` was ambiently red on `main` for days, failing on PRs that never touched quality. It was noticed by contributors tripping over it, not by a gate.

A ratchet nobody runs is a ratchet that has stopped ratcheting. The instrument is in good shape now — #6564 fixed the durability defect, and `main` is green — which makes this the moment to decide, not a reason to keep deferring.

## What needs deciding

1. **Should `quality` become a per-PR gate?** The header says the ratchet "makes it viable, since it can now actually pass" — that claim is testable and, as of `55d420669`, true.
[...]

PR #6712, opened 2026-08-26
ci: run the ratchet Python suite on a Go-free path as well as through the Go wrapper (refs #6709)

cajasmota:
Refs #6709.

**The issue's premise was wrong, and this PR no longer rests on it.** #6709 claimed that no workflow ran any Python test, on the strength of a grep over `.github/workflows/` that found no `unittest` / `pytest` / `python3 -m`. That grep was a **false negative** — the trigger is in Go, not in YAML:

```go
// internal/quality/ratchet_measured_on_6564_test.go  (#6568, on main,
// no build tag, no testing.Short() guard)
func TestRatchetMeasuredOnSelfTests_6564(t *testing.T) {
	requirePython3(t)
	if _, err := exec.LookPath("git"); err != nil { t.Skip(...) }
	script, _ := filepath.Abs(filepath.Join("..", "..", "scripts", "quality", "test_ratchet.py"))
	out, err := exec.Command("python3", script).CombinedOutput()
	if err != nil { t.Fatalf("scripts/quality/test_ratchet.py failed: %v\n%s", err, out) }
}
```

`test_ratchet.py` ends in `if __name__ == "__main__": unittest.main(verbosity=2)`, so that wrapper is not a vacuous call. `test.yml`'s `test` job runs `go test ./...` on `pull_request`. **All 33 Python tests have been running on every PR, on all three matrix OSes, all along.** #6709 is closed as `not_planned`.

Every claim in the original version of this PR that depended on the false premise — "no workflow ran any Python test", "ran only when a human typed the command locally", "#6570 / #6627 / #6633 landed suites that were never re-run automatically" — has been **deleted** from both the YAML comments and this body. They were false.

## What this job actually is

Cheap, deliberate **redundancy** with a narrow but real resilience value. The Go wrapper is the primary execution path; this is a second, Go-free one. It survives three things the wrapper does not:

1. **A Go-side compile failure.** `go test ./...` cannot report on the Python suite if any package in the repo fails to build — the wrapper never runs. This job needs no Go toolchain at all.
2. **`requirePython3(t)`**, which `t.Skip()`s when `python3` is off PATH. A skip is green. Here a missing `python3` fails the step.
3. **The wrapper's `exec.LookPath("git")` guard**, which likewise `t.Skip()`s — also green. Measured: `PATH=<python-only> python3 -m unittest scripts.quality.test_ratchet` → `FAILED (errors=33)`, exit 1.

It is also independent of the `test` job's `go test -timeout` (15m/40m) and its 90-minute leg cap. ~20s on `ubuntu-latest`, no toolchain, no cache restore.

## The change

One new job, `python-quality-tests`, appended to `.github/workflows/test.yml`. No source file is touched — `ratchet.py` and `test_ratchet.py` are byte-identical to `main` (`shasum` `011c006ceec3b7d5f894577687c3527e6809de00`, verified after the temporary red-direction edit was reverted by `cp` from a byte backup).

```yaml
  python-quality-tests:
    name: python ratchet tests
    if: github.event_name != 'pull_request_target'
    runs-on: ubuntu-latest
    timeout-minutes: 10
    steps:
      - uses: actions/checkout@v4
      - name: Python version
        run: python3 --version
      - name: Run ratchet unit tests
        run: python3 -m unittest -v scripts.quality.test_ratchet
```

## Why `test.yml` — the trigger, verified

```
$ python3 -c "import yaml;d=yaml.safe_load(open('.github/workflows/test.yml'));print(list(d[True].keys()))"
['workflow_call', 'workflow_dispatch', 'pull_request_target', 'pull_request']
```

The bare `pull_request:` (added by #6291) fires on open, synchronize and reopen. Confirmed empirically too — `test.yml` has recent real `pull_request` runs on other branches.

**Not `pre-merge.yml` and not `quality.yml` — both are `workflow_dispatch:`-only**, so a job wired there would never fire on a PR. The workflows with a real `pull_request` trigger are exactly: `board-hygiene.yml`, `coverage-docs.yml`, `cross-platform-compile.yml`, `test.yml`, `windows-installers.yml`.

**Separate job, not a step in `test`:** the `test` job carries a four-branch `if:`, a 3-OS matrix and a 90-minute leg cap. This job is a sibling with its own `if:`, excluding exactly one event — `pull_request_target`, where GitHub checks out the **base** ref, so running the tests there would prove nothing about the PR.

## Acceptance: both directions

**Red** — a temporary `assertEqual(1, 2)` case appended to the suite, reverted before commit:

```
$ python3 -m unittest scripts.quality.test_ratchet ; echo EXIT=$?
..............................F...
FAIL: test_deliberately_broken (...TemporaryRedDirectionProof.test_deliberately_broken)
AssertionError: 1 != 2 : deliberate failure to prove the CI job can fail
Ran 34 tests in 19.882s
FAILED (failures=1)
EXIT=1
```

**Green** — the unmodified suite:

```
$ python3 -m unittest -v scripts.quality.test_ratchet ; echo EXIT=$?
Ran 33 tests in 23.585s
OK
EXIT=0
```

**Zero-collection is a hard failure, not merely "visible".** Corrected and measured — on Python ≥3.12 an empty suite exits **5**:

```
$ python3 -m unittest -v empty_mod ; echo EXIT=$?
Ran 0 tests in 0.000s

NO TESTS RAN
EXIT=5
```

`ubuntu-24.04` ships Python 3.12.3, so that holds today. It **degrades to exit 0 if the runner image ever regresses below 3.12** — hence `-v`, which at least keeps an empty collection legible in the log on such an image. `python3 -m unittest` also exits non-zero on an import error.

Invocation checked from the repo root: `scripts.quality.test_ratchet` resolves via implicit namespace packages — no `__init__.py` anywhere, none needed. Stdlib-only, so no `setup-python` and no `pip install`.

`actionlint` reports 4 shellcheck findings on `test.yml` — all at lines 193/232/403, inside the pre-existing `test` job, all present on `main`. The new job adds **zero**.

## Git identity: the suite sets its own — no `git config` step added

Checked, not assumed. 12 `git init` calls, each immediately followed by repo-local config:

```python
git(root, "init", "--quiet", "--initial-branch=main")
git(root, "config", "user.email", "t@example.com")
git(root, "config", "user.name", "t")
git(root, "config", "commit.gpgsign", "false")
```

Verified empirically with no ambient identity at all:

```
$ env HOME=$EMPTY GIT_CONFIG_GLOBAL=$EMPTY/.gitconfig GIT_CONFIG_SYSTEM=/dev/null \
      python3 -m unittest scripts.quality.test_ratchet ; echo EXIT=$?
Ran 33 tests in 19.052s / OK / EXIT=0
```

Adding a `git config` step would be dead configuration masking the day a new `git init` forgets its own.

## No `paths:` filter

None, on the job or the workflow. A path filter is how this class of gate dies quietly (#6332 is the same shape one file over), and at ~20s there is no wall-clock argument for one.

## Is the Go wrapper now the redundant one? — judgement, no change made

**No. Keep it.** Recording the reasoning rather than acting on it:

- The wrapper runs the suite on **macOS and Windows** (via the `test` matrix). This job runs Ubuntu only. `test_ratchet.py` drives `git` through `subprocess` and builds repositories on the filesystem — precisely the surface where Windows path and line-ending behaviour diverges. Deleting the wrapper would silently drop two platforms.
- It runs under the tag/`ci:full`/`workflow_dispatch` legs too, where this job's coverage is incidental.
- It is, per below, the only thing that would still exercise the suite if this job were mutated to a no-op.

The overlap is one Ubuntu execution per PR, roughly 20s. That is not worth removing either side for. **Recommendation: no follow-up removal.** If one is ever wanted, the wrapper is the *more* valuable of the two, not the less.

## Known limit: two surviving mutants, recorded not fixed

Reviewer-supplied, and I agree they survive:

1. Appending `|| true` to the run step.
2. Flipping the `if:` to `== 'pull_request_target'`.

Either leaves the job **green forever while testing nothing**, and nothing in the repo would catch it — `actionlint` reports only the same 4 pre-existing findings under both. They are equivalent mutants under the current suite. The only thing that would close them is a guard test asserting that the `pull_request` workflow contains an unguarded invocation of the ratchet suite; **deliberately not built here** — it is a distinct piece of work with its own design questions.

Ironically the pre-existing Go wrapper partially covers that loss: if either mutant landed, `go test ./...` would still run all 33 Python tests, so the suite would keep executing even with this job neutered — which is another argument for keeping the wrapper.

## Also found, not fixed

- **`test_ratchet.py`'s own `__main__` git guard is dead code.** It checks the *return code* of `subprocess.call(["git", "--version"])`, but an absent `git` raises `FileNotFoundError` before any return code exists, so the script form tracebacks with exit 1 instead of taking its intended `sys.exit(0)` skip. Harmless in practice (both forms end up red), and out of scope — but the guard does not do what it reads as doing.
- **`acceptance.yml` has a `pr-linux` job whose `if: github.event_name == 'pull_request'` can never match**, because that workflow's `pull_request` trigger was removed. The header calls the dormancy intentional; it is still a live instance of the "job that cannot fire" shape and deserves its own issue.
- `quality.yml` remains `workflow_dispatch:`-only, so `run.sh` — the ratchet in anger, as opposed to its unit tests — still never runs per-PR. That is #6231 / #6586, out of scope.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -412,23 +412,40 @@ jobs:
           fi
 
   # ---------------------------------------------------------------------------
-  # Python unit tests for the quality ratchet (#6709).
+  # Python unit tests for the quality ratchet (refs #6709).
   #
-  # `scripts/quality/test_ratchet.py` is this repo's ONLY Python test file, and
-  # until this job existed NO workflow ran ANY Python test — its 33 tests, which
-  # guard `scripts/quality/ratchet.py` (the component that decides whether a
-  # quality regression blocks a merge), ran only when a human typed the command
-  # locally. #6570 / #6627 / #6633 all landed new tests for this file that were
-  # then never re-run automatically.
+  # WHAT THIS IS NOT: it is NOT the first time the Python suite runs in CI.
+  # #6709 claimed no workflow ran any Python test, on the strength of a grep
+  # over .github/workflows/ that found no `unittest`/`pytest`/`python3 -m`.
+  # That grep was a FALSE NEGATIVE — the trigger is in Go, not in YAML.
+  # `internal/quality/ratchet_measured_on_6564_test.go` (#6568) shells out to
+  # `python3 scripts/quality/test_ratchet.py`, and the script ends in
+  # `if __name__ == "__main__": unittest.main(verbosity=2)`, so all 33 tests
+  # already run under `go test ./...` in the `test` job above — on every PR, on
+  # every matrix OS. #6709 was closed as invalid.
+  #
+  # WHAT THIS JOB IS: deliberate, cheap redundancy with a narrow resilience
+  # value. The Go wrapper is the primary execution path; this is a second,
+  # Go-free one, and it survives three things the wrapper does not:
+  #   1. A Go-side compile failure. `go test ./...` cannot report on the Python
+  #      suite if any package in the repo fails to build — the wrapper never
+  #      runs. This job needs no Go toolchain at all.
+  #   2. `requirePython3(t)`, which t.Skip()s when python3 is off PATH. A skip
+  #      is green. Here, a missing python3 fails the step.
+  #   3. The wrapper's own `exec.LookPath("git")` guard, which t.Skip()s when
+  #      git is off PATH — again green. Here a git-less runner lands red:
+  #      measured, `PATH=<python-only> python3 -m unittest
+  #      scripts.quality.test_ratchet` -> "FAILED (errors=33)", exit 1.
+  # It is also independent of the `test` job's 15m/40m `go test -timeout` and
+  # its 90-minute leg cap.
   #
   # WHY IT LIVES IN test.yml: this workflow has a plain `pull_request:` trigger
   # (see the `on:` block above), so this job runs on every PR open/sync/reopen.
   # Do NOT move it to `pre-merge.yml` or `quality.yml` — BOTH are
-  # `workflow_dispatch:`-only, so a job wired there would never fire on a PR,
-  # reproducing the exact defect #6709 is about. As of this commit the
-  # workflows with a real `pull_request` trigger are: board-hygiene.yml,
-  # coverage-docs.yml, cross-platform-compile.yml, test.yml (this file) and
-  # windows-installers.yml.
+  # `workflow_dispatch:`-only, so a job wired there would never fire on a PR.
+  # As of this commit the workflows with a real `pull_request` trigger are:
+  # board-hygiene.yml, coverage-docs.yml, cross-platform-compile.yml, test.yml
+  # (this file) and windows-installers.yml.
   #
   # SEPARATE JOB, ON PURPOSE: it deliberately does not sit under the `test`
   # job's matrix or its multi-branch `if:`. It is Go-free (no toolchain, no
@@ -449,7 +466,7 @@ jobs:
   # HOME pointed at an empty directory and both GIT_CONFIG_GLOBAL and
   # GIT_CONFIG_SYSTEM neutralised: 33 tests, OK. So this job deliberately does
   # NOT `git config` a committer identity; adding one here would be dead
-  # configuration that hides the day the suite stops setting its own.
+  # configuration that hides the day a new `git init` forgets its own.
   python-quality-tests:
     name: python ratchet tests
     if: github.event_name != 'pull_request_target'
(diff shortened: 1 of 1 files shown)


cajasmota:
Review addressed at `ef538f30c`.

The blocking finding is correct and I have removed every claim that rested on the false premise. `internal/quality/ratchet_measured_on_6564_test.go` (#6568) already runs all 33 Python tests under `go test ./...` on every PR and every matrix OS, so this job is **redundancy**, not a missing gate. The YAML comment block and the PR body now say so, name that file, and state the three things this job survives that the wrapper does not (a Go-side compile failure; `requirePython3`'''s skip; the wrapper'''s `exec.LookPath("git")` skip). Closure keyword changed to `Refs #6709`.

Two corrections while re-verifying, both now in the diff:

1. **Zero-collection is stronger than I claimed, as the reviewer said.** Measured: `Ran 0 tests / NO TESTS RAN / EXIT=5` on Python >= 3.12. ubuntu-24.04 ships 3.12.3, so it is a hard red today; it degrades to exit 0 only if the image regresses below 3.12. Both stated.

2. **I briefly wrote a claim that was itself wrong, and caught it by measuring.** I had asserted that `python3 scripts/quality/test_ratchet.py` silently `sys.exit(0)`s when git is off PATH, and that `-m unittest` avoids that. It does not: the script'''s `__main__` guard tests the *return code* of `subprocess.call(["git", "--version"])`, but an absent git raises `FileNotFoundError` first, so the script form tracebacks with exit 1. The real green-on-missing-git path is the **Go wrapper'''s** `exec.LookPath("git")` `t.Skip`. Corrected in the YAML, and the dead guard is recorded as found-not-fixed.

**Is the Go wrapper now redundant? No — keep it.** It runs the suite on macOS and Windows via the `test` matrix; this job is Ubuntu-only, and `test_ratchet.py` drives `git` and the filesystem through `subprocess`, which is exactly where Windows diverges. It also covers the tag / `ci:full` / dispatch legs. And as you note, it is what would still exercise the suite if either surviving mutant landed. No follow-up removal recommended; if one is ever wanted, the wrapper is the more valuable of the two.

The two mutants (`|| true` on the run step; `if:` flipped to `== '''pull_request_target'''`) are recorded in the body as equivalent under the current suite, with the guard-test remedy named and explicitly not built here.

Issue #6360, opened 2026-08-19
the 10% syntax-error gate is a whole-file average, so a one-typo file emits silently-incomplete entities — direction 1 shipped in 1592e9a94, the degradation marker is what remains


cajasmota:
**Component:** `internal/treesitter` (the gate) + `internal/extractors/proto` (the consumer) · **Severity:** wrong data, worse than no data

`maxErrorRatio = 0.10` is a whole-file average. For protobuf that produces a bimodal outcome with a dangerous middle band:

- **proto2 / edition-2023 files** pin at 0.20–0.28 — the whole body collapses into one ERROR node — and are dropped cleanly. No output. Correct behaviour.
- **A proto3 file with one malformed message among many** parses far under the gate and is accepted **in full, including the broken part.**

Measured: a 12-message proto3 file with one syntactically broken message parses at **error_ratio 0.0033** and emits 35 entities. The broken message `M0` still gets a `message` entity, and its malformed field is **silently dropped** — `M0` indexes with 1 field where the source has 2, with nothing distinguishing it from a genuinely 1-field message. Three fixture variants (missing `=`, missing field name, missing `;`) all reproduce the same class.

*Stated as unverified:* an adversarial review reported a variant emitting a field whose **name is wrong** (`field:M0.string` instead of `M0.b`). Three attempts to reproduce that specific output produced the silent-drop above instead. Both are the same defect class; only the silent drop is measured here.

No file in the current corpus lands in the `0 < r < 0.10` band, so this is **11 files of evidence rather than a guarantee** — and the population that uses protobuf heavily is not represented locally.

## Why the shape is wrong

A whole-file average conflates "this file is not the language I think it is" with "this file has one typo". The first deserves rejection; the second deserves partial, *marked* extraction. The gate currently gets the extreme right and the middle exactly backwards: it is confident precisely where it should be cautious.

Two directions worth weighing:

1. **Skip the subtrees under ERROR nodes** and emit only the clean part. Simple, and makes the output a true subset of reality rather than a wrong description of it.
2. **Mark affected entities as degraded**, so a consumer can distinguish "1 field" from "1 field that we could read". More work, but it preserves the signal that something was unreadable — which (1) discards.

This is not protobuf-specific. The gate is shared, so any language whose grammar produces a localised ERROR subtree on a small syntax slip has the same exposure; protobuf is simply where it was measured first, because #6357 was the first change to make this extractor run at all.

Refs #6347, #6356, #6357

[...]

PR #6737, opened 2026-08-31
fix(#6736): a Kotlin controller that isn't last vanishes — the grammar, not the ERROR filter

### LINKED ISSUE #6736: a Kotlin @RestController followed by any later top-level declaration emits ZERO routes, with no ERROR node — every fixture hides it by putting the controller last
An annotated Kotlin `@RestController` class followed by **any** later top-level declaration produces **zero** route entities — and there is **no ERROR node in the tree**, so nothing looks wrong.

Measured on `96947e91a` (the #6499 arm-3 branch) while reviewing it; **pre-existing**, not introduced there.

| fixture | routes emitted |
|---|---|
| annotated class only | **1** (`SCOPE.Operation:RealController.kept`) |
| class + `fun loose()` (expression body) | **0** |
| class + `fun loose() { }` (block body) | **0** |
| class + `val loose` | **0** |
| class + a second annotated class | **1** — and it is the *second* class; the first is gone |
| `fun` **before** the class (class last) | **1** |

The pattern: only the **last** top-level declaration is seen. Anything after the controller class silently erases it.

## Why nothing is red

**Every Kotlin Spring fixture in the suite happens to put its single controller last in the file.** That is the only reason this passes. It is a whole-suite instance of the one-axis fixture problem — every fixture varies the controller's *contents* and holds its *position* constant.

In real Kotlin this is not an edge case. Top-level extension functions, top-level `val` constants, and a second class in the same file are all idiomatic. **For a realistic Kotlin Spring file this pass mostly emits nothing.**

## It also invalidates a stated mechanism in #6499 arm 3

That PR's `TestKotlinSpring6499_TopLevelFunctionIsNotAHandlerShape` asserts that top-level handlers are unreachable *because `walkKotlinClasses` is class-scoped*. That is **not** what is operating. The `loose` fixture does not parse at all — the tree carries `ERROR`, and the annotated top-level `fun` degrades to `prefix_expression → anonymous_function (ERROR (simple_identifier))`. The test would pass identically if the walk visited every node in the file. Being corrected on that PR; recording it here because the same misreading produced the "top-level handlers are unreachable" conclusion.

## What to establish

1. **Where the last-declaration-wins behaviour comes from.** Is `walkKotlinClasses` overwriting rather than appending, is it descending into the wrong node, or is the tree-sitter-kotlin parse itself collapsing siblings? The absence of an `ERROR` node in the multi-declaration cases (unlike the `loose` case) suggests the parse is fine and the walk is at fault — but that is inference, not measurement. **Measure it before fixing.**
2. **Whether the same shape affects the other Kotlin passes**, not just Spring routes. `walkKotlinClasses` has other callers; the extractor's own `class_declaration` walk is a separate path and appears healthy (arms 1+2 pin nested/top-level naming across many shapes), but that should be checked rather than assumed.

## Acceptance

A fixture with an annotated controller **followed by** a top-level function, a top-level `val`, and a second class — asserting the controller's routes are all present. **Non-vacuity:** it must fail on today's code. The reviewer's table above is the red baseline; reproduce it first.

And the fixture-position axis should be varied deliberately somewhere in the Kotlin Spring suite, since holding it constant is what hid this.

## Not in scope

`#6499` arm 3 itself (the route-target rename) — that is correct and independently verified as `resolved`, not excused. This defect bounds how much of a real Kotlin repo that fix can reach, which is why it should be settled **before arm 4** baselines the golden fixture on top of it.

Refs #6499.

cajasmota:
Fixes #6736.

A Kotlin `@RestController` that is not the last construct in its file emitted **zero** routes. Since real Kotlin routinely follows a controller with a top-level extension function, a constant `val`, or a second class, the Kotlin Spring pass emitted almost nothing on realistic input.

## The mechanism — measured, and not what the issue said

Both the issue body and my grounding comment on it were wrong, in opposite directions. Recorded here because the wrong theory is instructive.

The body reported "no tree-sitter ERROR node" as a puzzle. I commented that the observation was untrustworthy, because `pr.TSTree` is wrapped by the #6360 error-skipping filter, which makes ERROR nodes unreachable — and I named that wrapper the leading candidate.

**That candidate is dead.** The raw `official.Tree` was compared against `pr.TSTree` directly:

```
RAW official.Tree, controller-then-declarations: total=110 errNodes=0 ratio=0.0000
RAW  top=[package_header import_list prefix_expression function_declaration property_declaration class_declaration]
FILT top=[package_header import_list prefix_expression function_declaration property_declaration class_declaration]
```

Zero ERROR nodes, so `newErrorSkippingTree` is never constructed and raw and filtered are byte-identical. The body's "no ERROR node" was simply **true**, on the raw tree. My correction pointed away from the answer.

The real cause is a fwcd/tree-sitter-kotlin 0.3.8 grammar ambiguity, and its trigger is **narrower and different** from what the issue claims:

```
1 annotation, class then fun    err=0  top=[class_declaration function_declaration]   FINE
2 annotations, class LAST       err=0  top=[class_declaration]                        FINE
2 annotations, class then fun   err=0  top=[prefix_expression function_declaration]   BROKEN
```

It is not "only the last declaration is seen". It is **two or more consecutive annotations on a top-level declaration that is not last**. The annotations are read as a chain of `prefix_expression` operators instead of the declaration's `modifiers`; the class header degrades to `infix_expression (simple_identifier) (simple_identifier)` and the body to a trailing `lambda_literal`.

So the fixture corpus is blind to **two** axes, not one: position *and* annotation count. That is why 63 Spring fixtures all miss it.

## Blast radius is wider than the issue states

The issue frames this as a Spring-routes defect. It is not. The Kotlin **extractor** loses the class too: on `main` this file emits `SCOPE.Operation kept` (unqualified) and **no** `SCOPE.Component RealController`.

A Spring-walk-only fix would therefore have emitted a route pointing at an entity that does not exist — which is exactly the class of defect #6499 just spent four arms repairing. The fix is placed at the parse layer for that reason.

## The repair

Inserting an explicit `;` at the end of the misparsed construct disambiguates in favour of the declaration reading — verified across `class` / `object` / `interface` / `fun` carriers, with 2 and 3 annotations, followed by a `fun` / `val` / `class` / comment. So: parse, detect the signature, insert one `;` per misparse, re-parse, and present the result with byte offsets mapped **back** to the original source, so callers keep slicing their own unmodified buffer.

Scope and cost:
- **Kotlin only**, and only when a top-level `prefix_expression` whose first child is an `annotation` is actually present — a scan over the root's direct children. A clean file pays that one scan and is returned completely untouched: same tree, no wrapper, no re-parse.
- Deliberately unlike the shared #6360 wrapper, no other grammar is reachable from this code.
- The re-parse loop is bounded by `maxKotlinRepairRounds`, and each round must strictly reduce the misparse count or the loop stops. Every measured shape converges in one round; the ceiling is belt-and-braces against a non-converging grammar shape, not an expected cost.

**Known limit, disclosed:** each inserted `;` shifts the *columns* of anything after it on the same line by one. The insertion point is the end of a top-level declaration, which is end-of-line in any real Kotlin file, so in practice nothing follows. Byte offsets are mapped exactly and rows are never affected, because no newline is ever inserted.

## Java is unaffected — measured, not assumed

Java's root node is `program`, not `source_file`, and it parses the same shape correctly, so the detector cannot fire. My grounding comment guessed Java was probably safe on grammar grounds (no ASI scanner); that guess was right, and this is now the measurement behind it.

## Tests

The reproducing test lives in `internal/engine/spring_routes_kotlin_test.go` (general-reach suite, not the arm-scoped `_6499_` file) and carries a **non-vacuity control** on the raw unwrapped parse proving the `class_declaration` exists — without it a red is indistinguishable from a fixture that failed to parse, which is the confusion that produced the wrong theory above.

Plus 7 treesitter tests: the carrier/annotation-count matrix, offset mapping, clean-file no-op, ERROR-skip still intact, and the detector discriminator.

## Mutation score

| # | Direction | Mutant | Verdict |
|---|---|---|---|
| M1 | permissive | drop the annotation-first-child guard | ALIVE → **DEAD** (pin added) |
| M2 | permissive | remove the `language == "kotlin"` gate | **ALIVE** |
| M3 | permissive | `swallowsDeclaration` always false | DEAD (8 failures) |
| M4 | revert | disable the repair call | DEAD (11 failures) |
| M5 | permissive | `unshift()` returns the offset unchanged | DEAD |

**M2 is reported ALIVE deliberately.** A sweep of all 28 vendored grammars, plus Groovy / Swift / Rust / Scala on language-appropriate annotated sources, found none that trips the detector. The language gate is defence-in-depth, not load-bearing, and no natural fixture kills it. Recorded rather than papered over with a manufactured fixture.

## Suites

Package set derived from the changed directories plus their parents: `internal/engine`, `internal/engine/httproutes`, `internal/treesitter` and all 28 `ts/grammars/*` plus `ts/official`, and `internal/extractors/...` (treesitter is their parse chokepoint). `gofmt -l` clean, `go vet` clean, `-count=1` throughout.

### DIFF
--- a/internal/engine/spring_routes_kotlin_test.go
+++ b/internal/engine/spring_routes_kotlin_test.go
@@ -5,9 +5,13 @@ package engine
 
 import (
 	"context"
+	"strings"
 	"testing"
 
 	"github.com/cajasmota/grafel/internal/extractor"
+	"github.com/cajasmota/grafel/internal/treesitter/ts"
+	tskotlin "github.com/cajasmota/grafel/internal/treesitter/ts/grammars/kotlin"
+	tsofficial "github.com/cajasmota/grafel/internal/treesitter/ts/official"
 )
 
 // sampleKotlinSpringController mirrors the Java fixture in spring_routes_test.go.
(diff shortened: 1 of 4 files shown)


cajasmota:
## REQUEST-CHANGES — independent adversarial review

Five of the six attack surfaces came back clean, with evidence. Two blocking defects, one of which makes this change a **regression** on a shape more idiomatic than the one it fixes.

### Clean, and better than the PR body claims

- **Offset remapping (attack 1) — exhaustively verified.** A probe compared `original[Start:End)` against the repaired slice with insertions removed, for **every node** in the tree, recursing to full depth — not just identifiers. Zero mismatches across four fixtures (nested annotation args, raw strings, comments, two controllers). `Parent()`, `PrevSibling()`, `ChildByFieldName()` all return remapped nodes; every accessor rewraps, and `ts.Node` exposes no `NextSibling` and no cursor API, so the surface is closed. `len(buf) == len(src)+len(insertions)`, and stripping the insertions reproduces `src` byte-for-byte.
- **Column shift (attack 2) — harmless, for a stronger reason than the PR gives.** `Point.Column` has **zero consumers in the repository**; the only occurrence constructs it in the official adapter. Kotlin consumers read `StartPoint().Row`. The "in practice nothing follows on that line" argument is weaker than the truth and should be replaced by it.
- **Ownership/lifetime (attack 3) — clean.** Round 0 has `curTree == tree`, so the guard prevents a double-close; each later round closes the previous intermediate. The failure path returns `(nil,false)` with `tree` untouched. No caller of `parseOfficial` retains a pre-repair node.
- **Caching/incremental (attack 4) — clean.** `official.Parse` never reuses an old tree; `internal/extractors/incremental.go` closes `input.TSTree` per file and caches none.
- **`;` into a string or comment (attack 5) — clean.** Raw-string and `//`-comment `@`-lookalike fixtures produced correct insertion points.
- **M2 (attack 6) — honest, and strengthenable.** A symbol sweep of all 27 vendored grammars found only `kotlin` carries all three of `prefix_expression`, `"annotation"`, `"source_file"`; Groovy has two of three. That belongs in the doc comment as the reason the gate is unkillable.

---

### D1 (blocking) — the loop has no convergence check, and a non-converging repair is *kept*

`internal/treesitter/kotlin_annot_repair.go:96-122`.

The doc at `:73` and the PR body both state: *"Each round must strictly reduce the number of misparsed constructs or the loop stops."* **No such check exists.** I confirmed this by reading the loop: the only exits are the round ceiling and `len(pts) == 0`, with no comparison against the previous round's count. `:122` then accepts any result where `len(insertions) != 0`, contradicting the neighbouring claim that it returns `(nil,false)` when the repair did not help.

This is the house defect class — prose asserting a property no code implements, let alone any test.

It is not theoretical. Reproducer, an idiomatic single-line class body with two annotations, not last:

```kotlin
package io.demo

@A
@B
class One { fun a() = 1 }

@A
@B
class Two { fun b() = 2 }

fun tail() = 3
```

The repaired buffer receives `};;;` at **both** sites (3 rounds, `insertions=[48 49 50 84 84 84]`), the tree is **still** `prefix_expression` — nothing fixed — and it now carries ERROR nodes that were not there before. Measured end-to-end through the factory on a 20-class file:

```
main:        errNodes=0   ratio=0.0000
this PR:                  ratio=0.0347
```

So on this shape the change strictly degrades: no recovery, an invented syntax error, and the #6360 error-skipping wrapper activating where it previously did not. With `maxErrorRatio` at 0.10, this is a path to a whole file being dropped with `ErrHighSyntaxErrorRate` that `main` accepts today.

**Required:** implement the convergence check the doc already claims — if a round does not strictly reduce `len(pts)`, stop and return `(nil, false)`, discarding the partial repair rather than keeping a mangled buffer. Never return a tree with a higher error count than the one you were given; that invariant deserves its own test. And pin this fixture: the single-line class body is more idiomatic Kotlin than the multi-line shape the PR fixes.

### D2 (blocking) — `unshift`'s boundary is unpinned, and a real off-by-one survives

Mutant at `:296`: `sort.SearchInts(ins, int(off))` → `int(off)+1`, turning "strictly before" into "at or before". `go vet` clean. **All 8 new treesitter tests and the engine `6736` tests still pass.**

It is genuinely behavioural: a `class_declaration` at `[17,136)` becomes `[17,135)`, losing its closing `}`, and the `}` token collapses to `[136,136)=""`. The doc's claim that mapping is "exact for both start (inclusive) and end (exclusive) bounds" is asserted by no test. Add one that pins the end-bound boundary specifically — an insertion at exactly a node's `EndByte`.

### D3 — one new test is vacuous

With the repair call disabled, `TestKotlinAnnotRepair_OffsetsAddressTheOriginalSource` **passes**. Only `RecoversDeclarationShapes` failed of the eight. It is killed by M5 but is not a regression pin, and combined with D2 the offset-mapping axis is under-pinned at exactly its boundary.

### Not checked

Live corpus indexing (no daemon run). The grammar sweep was symbol-presence plus three live parses, not 27 live parses.

cajasmota:
## Coordinator verification of `33432a3c2` — accepting the D1 deviation

I asked for "stop and return `(nil, false)`, discarding the partial repair." You implemented **discard the failing round, keep rounds that made progress**, and argued that strict discard-everything loses a repairable controller because of an unrelated sibling. I checked that rather than taking it on trust, and the argument holds — accepted as implemented.

**My own mutant, C1, which no one else ran.** I wanted to attack the *bookkeeping* behind the new guard rather than the guard itself, since that is where a permissive slip would hide. Deleting the single line

```go
remaining = len(newPts)
```

leaves `remaining` pinned at the ORIGINAL misparse count forever, so the guard still fires — but only against the starting count, never against the previous round. Rounds that make no progress relative to each other sail through as long as they stay under the original. `go vet` clean, so it is a real mutant and not a compile artefact.

**DEAD**, and killed by exactly the test that pins your deviation:

```
--- FAIL: TestKotlinAnnotRepair_PartialRepairSurvivesANonConvergingSibling
    the repairable `class Multi` was NOT recovered; an unrepairable sibling in the
    same file discarded the whole repair.
    top children: [package_header prefix_expression prefix_expression function_declaration]
```

That is the result I wanted to see. The mixed-file fixture is not decoration around a judgement call — it is load-bearing, and it is the thing that would catch a future refactor quietly reverting the partial-keep semantics. `internal/engine` 6736 tests stayed green under the mutant, confirming the kill is specific to the treesitter layer rather than incidental.

Restored byte-for-byte and confirmed `git diff` against **HEAD** is empty, not merely that the file matches my backup — a stale backup has silently reverted an intervening commit in this repo before.

### On D1b staying ALIVE

Correct call, and the comment is the right shape. You searched for an input where a strictly-progressing round also raises the error count, found none across ten malformed tails, and wrote that down instead of manufacturing a fixture to force a kill. Keeping the never-worse invariant as defence in depth while **saying in the code that it is currently unreachable** is exactly the discipline whose absence caused D1 in the first place — a comment claiming a property the code did not implement. Do not let anyone talk you into "killing" it later with a synthetic case.

### What I especially want on the record

You established something the review did not: the single-line class body is **not repairable at all** — `pts=[31]` is invariant across all four rounds, because the body is already read as a trailing lambda, so `;` can never disambiguate it. That converts D1 from "the loop lacks a check" into "there is a shape this strategy provably cannot fix, and the correct behaviour is a clean refusal." The refusal path is now the specified behaviour rather than an accident, which is a materially better outcome than the guard I asked for.

Merging once CI is green.

Issue #6735, opened 2026-08-28
test(#6735): the effective-effects tests isolate HOME but not GRAFEL_HOME, so any doc-compliant sandbox run is red


cajasmota:
**`main` is currently RED, and CI cannot tell us — `test.yml` has no `push` trigger, so the suite has not run on `main` since the 0.3.0 tag.**

## The failure

On a clean checkout of `main` = `28c9e47e0`, `-count=1`:

```
--- FAIL: TestV2PathDetail_EffectiveEffects_AggregatesDownstreamWrite
    v2_paths_effective_effects_test.go:144: effective_effects: want db_write, got []
--- FAIL: TestV2PathDetail_EffectiveEffects_DirectSinkTaggedDirect
    v2_paths_effective_effects_test.go:185: effective_effects: want db_write, got []
```

Independently reproduced twice: once by an agent that stashed its branch and re-ran against base, once by me on the working tree at `28c9e47e0`. No skip gates, no env dependence, no build tags in that file.

## Why nothing caught it — this is the real defect

`.github/workflows/test.yml` triggers on `workflow_call` (release tag), `workflow_dispatch`, `pull_request_target: [labeled]`, and `pull_request`. **There is no `push` trigger.**

So every PR is tested against its own head, and **`main` is never tested after a merge**. Two PRs that are each green independently can land a semantic conflict, and nothing runs. The most recent `test.yml` run on `main` is `b037c3ffb` — the 0.3.0 tag commit, 2026-08-22 — and it was green.

**91 PRs have merged since.** The regression is somewhere in that range, and it has been invisible the entire time.

Note the irony recorded in that workflow's own comment on the `pull_request` trigger: *"#6291 — 24 merges landed with the suite never running once."* That fixed the PR side. The `main` side has the same hole and is still open.

## Two separate pieces of work

**1. Find and fix the regression.** `internal/dashboard/v2_paths_effective_effects_test.go` has not changed since `6bb88465e` (August, a grafel-home path refactor), so a **production** change broke it — most likely in effects aggregation or in the resolution the aggregation reads. `want db_write, got []` means the downstream write is no longer being found at all, not mis-labelled.

Bisect between `b037c3ffb` (green, tagged) and `28c9e47e0` (red). That range is well-defined and the test is fast, so this should be mechanical.

[...]

Issue #6548, opened 2026-08-22
PARTIAL: all four home-boundary arms shipped, incl. the other-user-home ruling (a4dbdbcc7) — the reported iCloud consent dialog is still unexplained


cajasmota:
**Reported from real use on macOS.** The user was shown a system consent dialog:

> **"grafel" wants to access files managed by "iCloud Drive".**

They never pointed grafel at iCloud. Something in grafel reached a path under `~/Library/Mobile Documents/` — or another TCC-guarded location — on its own.

## Why this is a real defect and not cosmetic

A tool that indexes source code has **no business touching iCloud Drive, Photos, Contacts, or another user's home directory**, and being asked to approve it is the point at which a user stops trusting the tool with filesystem access. The prompt is the visible symptom; the traversal behind it is the defect.

It is also not macOS-specific. The same unbounded traversal on other platforms means:

- **Linux** — climbing out of `/home/<user>` into `/home/<other>`, or into `/etc`, `/proc`, `/var`.
- **Windows** — climbing out of `C:\Users\<user>` into `C:\Users\<other>`, or into `C:\Windows`.

macOS is simply the platform that *asks*. The other two fail silently, which is worse: a permission error is logged and swallowed, or the read succeeds and grafel has quietly walked a directory that is none of its business.

## The suspected mechanism — being verified, not assumed

grafel has several **walk-up / ancestor traversals**: climbing from a repo path toward the filesystem root looking for a marker (a `.grafel/group.json`, a git boundary, a state directory). Two known ones, with the caveat that the enumeration is in progress and will be posted here:

- `canonicalizePath` walks **every ancestor** of every repo path (`internal/daemon/`).
- the `.grafel/group.json` walk-up around `internal/mcp/routing.go:242`.

The dangerous shape is a loop whose **only** stop condition is finding what it is looking for. If the marker is absent, it climbs until `filepath.Dir` stops changing — i.e. all the way to `/` — passing through `$HOME` and everything under it on the way.

A recursive **scan** that follows symlinks would explain the prompt equally well, so that is being checked too rather than assumed away.

## What a fix needs

1. **A shared boundary predicate, used by every traversal.** Not a fix at one call site — the whole point is that there are several, and the next one added will repeat the mistake.
2. **Stop at the user's home directory** when the start path is inside it. A repo under `$HOME` has no marker above `$HOME` worth finding.
[...]

PR #6787, opened 2026-09-03
fix(#6548): refuse to climb into another user's home — the package named it as the harm and then allowed it

cajasmota:
Refs #6548 — implements **requirement 3**. Deliberately NOT `Closes`: the reported macOS iCloud dialog is not proven fixed by this arm.

## The contradiction the owner ruled on

`pathboundary.go:13` named *climbing into other users' home directories* as the harm the package exists to prevent. `:86` then accepted the opposite: *"a user explicitly pointed grafel at it, so the climb must continue."*

Both could not be true. On 2026-09-02 the owner ruled **`:86` wrong** — an explicit path is a claim about **intent**, not about **permission**. My recommendation had been to amend the issue instead; this implements the owner's ruling.

## Shape

`internal/pathboundary/otherhome.go` — a fourth `dirClass`, `otherHome`, refused **at or under**, and checked **first** in `classifyDir`, before `mayBeProtected`. It is purely lexical, so it costs no filesystem work of its own: `resolved` is already in hand.

The class split against `classTCC` is load-bearing. `classTCC` is refused at the **outermost** directory only, which is what preserves the explicit-path escape hatch inside *the current user's own* home. `otherHome` gets no such treatment — see the coordinator mutant below.

Home references are derived from the **real uid** via `os/user`. Held in **both** literal and resolved spellings.

## Review round 1 → 2: five defects, all fixed

**D1 — the boundary was inert behind a symlink.** Only literal spellings were held, so a climb starting under a symlinked root resolved past the check. Measured: `/var/folders/…` (symlinked) read `resolved=false` and **visited 11 directories**; `/private/tmp/…` (direct) refused correctly. Both now `resolved=true, VISITED=0`.

**D2 — no exemptions.** `/Users/Shared` is not a home, and refusing it breaks a legitimate shared checkout. Now a per-platform list: darwin `Shared, Guest, .localized`; windows `Public, Default, "Default User", "All Users"`; else `linuxbrew, lost+found, Public`. Matched case-folded on **every** platform via a dedicated `eqFold`, deliberately **not** through the identity comparisons, which keep filesystem semantics. Over-firing control: `Sharedish` is still refused.

**D3 — the root guard mis-classified containers.** "Parent of a home unless it is the filesystem root" made `/var` a home container, so `/var/root` (the darwin root account) read as another user's home. Replaced by **container plausibility**: the container's own name is `Users` or `home` (folded), or a platform convention. `/var/root`→`/var`, `/var/lib/postgresql`→`/var/lib`, and `/root`→`/` are all now rejected.

**D4 — `$HOME` could launder a foreign home into "ours".** The reference set unioned the real-uid home with `$HOME`, so exporting `HOME=/Users/bob` made bob's home read as the current user's. The `$HOME` leg is **removed**, not documented as a bypass. Nothing breaks: the real uid is authoritative, `climbHome` is not consulted for this class, and test isolation goes through `OverrideHomeReferences` (now single-argument).

**D5** — a prose sentence asserting behaviour the code did not have, deleted.

Author + reviewer mutants: refuse-unconditionally, allow-unconditionally, trust-`$HOME`-not-uid, prefix-compare, root-as-container, drop-resolved-leg, literal-spellings-only, re-add-`$HOME`-union, any-parent-is-container, drop-not-a-home-list — **all DEAD**.

## Coordinator-verified: is "at or under" actually pinned, or only "at"?

That is the one question this arm rests on, and it is exactly the reading the owner overruled. So I gave `otherHome` the same outermost-only treatment `classTCC` gets:

```go
if (curClass.otherHome && !parentClass.otherHome) || curClass.media || (curClass.tcc && !parentClass.tcc) {
```

Under this, a climb starting at `/Users/bob/proj/pkg` sees its parent `/Users/bob` also classed `otherHome`, so the refusal never fires and the climb proceeds — precisely the "they pointed us at it" laundering.

**DEAD on six tests**, with diagnostics that name the offending reads rather than a bare boolean:

```
--- FAIL: TestClimb_RefusesAnotherUsersHome
    Climb READ another user's home: started at ".../Users/bob/proj/pkg" under ".../Users/bob"
    while the current user's home is ".../Users/alice"; visited [.../Users/bob/proj/pkg .../Users/bob/proj]
--- FAIL: TestClimb_SymlinkedPathIntoAnotherUsersHome
--- FAIL: TestClimb_HOMECannotLaunderAnotherUsersHome
--- FAIL: TestClimb_SiblingHomeIsNotAPrefixMatch
--- FAIL: TestOtherHome_NotAHomeNamesAreExempt
--- FAIL: TestClimb_RefusesUnderAnotherUsersProtectedTree
```

Note the last two: the symlink leg (D1) and the exemption over-firing control (D2) both die to a mutant aimed at neither, which is what a non-vacuous suite looks like. `go vet` clean before scoring — a non-compiling mutant proves nothing. Restored by `cp` to shasum `e123218d`, worktree clean at `bda06308`.

## Interaction with #6735

`internal/registry`'s sweep guard caught two tests in this change that redirected `HOME`. They now pin `GRAFEL_HOME` beside it rather than going on the ledger — the guard did its job on the first change written after it landed.

## What this does NOT close

The **reported symptom** on #6548 is a macOS iCloud permission dialog. This arm refuses other users' homes; it is not demonstrated to be what raised that dialog. #6548 stays open on that.

`gofmt` clean, `go.mod`/`go.sum` untouched. `-count=1` green on `./internal/pathboundary/` and `./internal/registry/`. No extractor touched, so no golden fixtures and no `cmd/grafel` digest exposure.

### DIFF
--- a/internal/gitmeta/gitdir_home_boundary_6548_test.go
+++ b/internal/gitmeta/gitdir_home_boundary_6548_test.go
@@ -60,18 +60,30 @@ func TestHasGitDirInTree_FindsGitAtHome(t *testing.T) {
 	}
 }
 
-// TestHasGitDirInTree_OutsideHomeStillClimbs — a start path outside $HOME keeps
-// climbing; a boundary that fires on any home-LOOKING ancestor would break
+// TestHasGitDirInTree_OutsideHomeStillClimbs — a start path outside every home
+// keeps climbing; a boundary that fired on any ancestor at all would break
 // repos under /opt, /var/folders or a CI checkout root.
+//
+// #6548 requirement 3 (owner decision 2026-09-02) INVERTED half of this test.
+// It used to start under <root>/Users/someone — a sibling of the current
+// user's home — and assert the .git above the Users level was still found.
+// That case is now a refusal, and is asserted as one below; the
+// still-climbs half moves to a non-home root, which is what it was really for.
 func TestHasGitDirInTree_OutsideHomeStillClimbs(t *testing.T) {
 	root, _ := fakeHomeUnder6548(t, "Users", "me")
 
-	// The .git sits ABOVE the "Users" level — the exact spot a home-LOOKING
-	// boundary would cut the climb off.
 	mkdir6548(t, filepath.Join(root, ".git"))
-	start := mkdir6548(t, filepath.Join(root, "Users", "someone", "src", "deep"))
+	start := mkdir6548(t, filepath.Join(root, "srv", "ci", "src", "deep"))
 	if !HasGitDirInTree(start) {
-		t.Fatalf("HasGitDirInTree stopped before a legitimate .git outside $HOME (a home-LOOKING ancestor is not a boundary; start %s)", start)
+		t.Fatalf("HasGitDirInTree stopped before a legitimate .git outside every home (start %s)", start)
+	}
+
+	// The inverted case: the same .git, reached from a SIBLING HOME, must not
+	// be found — the climb never enters /Users/someone at all.
+	sibling := mkdir6548(t, filepath.Join(root, "Users", "someone", "src", "deep"))
+	if HasGitDirInTree(sibling) {
+		t.Fatalf("HasGitDirInTree climbed out of another user's home %s (site: internal/gitmeta/gitmeta.go HasGitDirInTree)",
+			filepath.Join(root, "Users", "someone"))
 	}
 }
 
--- a/internal/gitmeta/gitdir_home_boundary_6548_test.go
+++ b/internal/gitmeta/gitdir_home_boundary_6548_test.go
@@ -5,6 +5,7 @@ import (
 	"path/filepath"
 	"testing"
 
+	"github.com/cajasmota/grafel/internal/pathboundary"
 	"github.com/cajasmota/grafel/internal/testsupport"
 )
 
@@ -70,7 +71,10 @@ func TestHasGitDirInTree_FindsGitAtHome(t *testing.T) {
 // That case is now a refusal, and is asserted as one below; the
 // still-climbs half moves to a non-home root, which is what it was really for.
 func TestHasGitDirInTree_OutsideHomeStillClimbs(t *testing.T) {
-	root, _ := fakeHomeUnder6548(t, "Users", "me")
+	root, home := fakeHomeUnder6548(t, "Users", "me")
+	// $HOME is no longer consulted by the other-user-home class (it can be
+	// laundered), so the fixture's home is injected through the seam.
+	t.Cleanup(pathboundary.OverrideHomeReferences(home))
 
 	mkdir6548(t, filepath.Join(root, ".git"))
 	start := mkdir6548(t, filepath.Join(root, "srv", "ci", "src", "deep"))
--- a/internal/gitmeta/gitdir_home_boundary_6548_test.go
+++ b/internal/gitmeta/gitdir_home_boundary_6548_test.go
@@ -74,7 +74,7 @@ func TestHasGitDirInTree_OutsideHomeStillClimbs(t *testing.T) {
 	root, home := fakeHomeUnder6548(t, "Users", "me")
 	// $HOME is no longer consulted by the other-user-home class (it can be
 	// laundered), so the fixture's home is injected through the seam.
-	t.Cleanup(pathboundary.OverrideHomeReferences(home))
+	t.Cleanup(pathboundary.OverrideHomeReferences(home, pathboundary.RealUIDHomeForTest()))
 
 	mkdir6548(t, filepath.Join(root, ".git"))
 	start := mkdir6548(t, filepath.Join(root, "srv", "ci", "src", "deep"))
--- a/internal/gitmeta/other_user_home_6548_test.go
+++ b/internal/gitmeta/other_user_home_6548_test.go
@@ -33,7 +33,7 @@ func TestHasGitDirInTree_RefusesAnotherUsersHome(t *testing.T) {
 	alice := mkdir(t, filepath.Join(users, "alice"))
 	bob := mkdir(t, filepath.Join(users, "bob"))
 
-	restore := pathboundary.OverrideHomeReferences(alice)
+	restore := pathboundary.OverrideHomeReferences(alice, pathboundary.RealUIDHomeForTest())
 	defer restore()
 
 	repo := mkdir(t, filepath.Join(bob, "proj"))
@@ -64,7 +64,7 @@ func TestHasGitDirInTree_SiblingHomeIsNotAPrefixMatch(t *testing.T) {
 	alice := mkdir(t, filepath.Join(users, "alice"))
 	alice2 := mkdir(t, filepath.Join(users, "alice2"))
 
-	restore := pathboundary.OverrideHomeReferences(alice)
+	restore := pathboundary.OverrideHomeReferences(alice, pathboundary.RealUIDHomeForTest())
 	defer restore()
 
 	repo := mkdir(t, filepath.Join(alice2, "proj"))
--- a/internal/gitmeta/other_user_home_6548_test.go
+++ b/internal/gitmeta/other_user_home_6548_test.go
@@ -33,7 +33,7 @@ func TestHasGitDirInTree_RefusesAnotherUsersHome(t *testing.T) {
 	alice := mkdir(t, filepath.Join(users, "alice"))
 	bob := mkdir(t, filepath.Join(users, "bob"))
 
-	restore := pathboundary.OverrideHomeReferences(alice, alice)
+	restore := pathboundary.OverrideHomeReferences(alice)
 	defer restore()
 
 	repo := mkdir(t, filepath.Join(bob, "proj"))
@@ -64,7 +64,7 @@ func TestHasGitDirInTree_SiblingHomeIsNotAPrefixMatch(t *testing.T) {
 	alice := mkdir(t, filepath.Join(users, "alice"))
 	alice2 := mkdir(t, filepath.Join(users, "alice2"))
 
-	restore := pathboundary.OverrideHomeReferences(alice, alice)
+	restore := pathboundary.OverrideHomeReferences(alice)
 	defer restore()
 
 	repo := mkdir(t, filepath.Join(alice2, "proj"))
--- a/internal/gitmeta/other_user_home_6548_test.go
+++ b/internal/gitmeta/other_user_home_6548_test.go
@@ -1,0 +1,76 @@
+package gitmeta
+
+import (
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/cajasmota/grafel/internal/pathboundary"
+)
+
+// #6548 requirement 3, NON-VACUITY. The boundary is only real if it refuses on
+// a REAL traversal path — extracting a predicate and unit-testing it does not
+// pin the call site (#6533). HasGitDirInTree is a production climber that does
+// an os.Stat of <ancestor>/.git per level, so its answer is a direct
+// observation of WHAT WAS READ: the .git exists, and the only way to report
+// false is to never have looked.
+//
+// Site named by these tests: internal/gitmeta/gitmeta.go HasGitDirInTree, via
+// pathboundary.Climb.
+
+func mkdir(t *testing.T, p string) string {
+	t.Helper()
+	if err := os.MkdirAll(p, 0o755); err != nil {
+		t.Fatalf("mkdir %s: %v", p, err)
+	}
+	return p
+}
+
+// TestHasGitDirInTree_RefusesAnotherUsersHome — a git checkout sitting in
+// another user's home must not be discovered, even though it is right there.
+func TestHasGitDirInTree_RefusesAnotherUsersHome(t *testing.T) {
+	users := mkdir(t, filepath.Join(t.TempDir(), "Users"))
+	alice := mkdir(t, filepath.Join(users, "alice"))
+	bob := mkdir(t, filepath.Join(users, "bob"))
+
+	restore := pathboundary.OverrideHomeReferences(alice, alice)
+	defer restore()
+
+	repo := mkdir(t, filepath.Join(bob, "proj"))
+	mkdir(t, filepath.Join(repo, ".git"))
+	start := mkdir(t, filepath.Join(repo, "internal", "pkg"))
+
+	if HasGitDirInTree(start) {
+		t.Fatalf("HasGitDirInTree read another user's home (site: internal/gitmeta/gitmeta.go HasGitDirInTree): found %q while the current user's home is %q",
+			filepath.Join(repo, ".git"), alice)
+	}
+
+	// UNDER-FIRING CONTROL: the identical layout inside the CURRENT user's own
+	// home must still resolve. A boundary that refuses unconditionally passes
+	// the assertion above and fails this one.
+	ownRepo := mkdir(t, filepath.Join(alice, "proj"))
+	mkdir(t, filepath.Join(ownRepo, ".git"))
+	ownStart := mkdir(t, filepath.Join(ownRepo, "internal", "pkg"))
+	if !HasGitDirInTree(ownStart) {
+		t.Fatalf("HasGitDirInTree failed to find %q inside the current user's own home — the boundary refuses everything",
+			filepath.Join(ownRepo, ".git"))
+	}
+}
+
+// TestHasGitDirInTree_SiblingHomeIsNotAPrefixMatch — /Users/alice2 is not
+// inside /Users/alice, on a real traversal path.
+func TestHasGitDirInTree_SiblingHomeIsNotAPrefixMatch(t *testing.T) {
+	users := mkdir(t, filepath.Join(t.TempDir(), "Users"))
+	alice := mkdir(t, filepath.Join(users, "alice"))
+	alice2 := mkdir(t, filepath.Join(users, "alice2"))
+
+	restore := pathboundary.OverrideHomeReferences(alice, alice)
+	defer restore()
+
+	repo := mkdir(t, filepath.Join(alice2, "proj"))
+	mkdir(t, filepath.Join(repo, ".git"))
+
+	if HasGitDirInTree(mkdir(t, filepath.Join(repo, "sub"))) {
+		t.Fatalf("prefix comparison on a real traversal path: %q was treated as inside %q", alice2, alice)
+	}
+}
--- a/internal/mcp/cwd_home_boundary_6548_test.go
+++ b/internal/mcp/cwd_home_boundary_6548_test.go
@@ -90,7 +90,7 @@ func TestGroupFromCWD_OutsideHomeStillClimbs(t *testing.T) {
 	root, home := fakeHomeUnder(t, "Users", "me")
 	// $HOME is no longer consulted by the other-user-home class (it can be
 	// laundered), so the fixture's home is injected through the seam.
-	t.Cleanup(pathboundary.OverrideHomeReferences(home))
+	t.Cleanup(pathboundary.OverrideHomeReferences(home, pathboundary.RealUIDHomeForTest()))
 
 	writeGroupMarker(t, root, "workspace-root")
 
--- a/internal/mcp/cwd_home_boundary_6548_test.go
+++ b/internal/mcp/cwd_home_boundary_6548_test.go
@@ -5,6 +5,7 @@ import (
 	"path/filepath"
 	"testing"
 
+	"github.com/cajasmota/grafel/internal/pathboundary"
 	"github.com/cajasmota/grafel/internal/testsupport"
 )
 
@@ -86,7 +87,10 @@ func TestGroupFromCWD_FindsMarkerAtHome(t *testing.T) {
 // user's home — and require the marker above the Users level to be found.
 // That case is now a refusal and is asserted as one below.
 func TestGroupFromCWD_OutsideHomeStillClimbs(t *testing.T) {
-	root, _ := fakeHomeUnder(t, "Users", "me")
+	root, home := fakeHomeUnder(t, "Users", "me")
+	// $HOME is no longer consulted by the other-user-home class (it can be
+	// laundered), so the fixture's home is injected through the seam.
+	t.Cleanup(pathboundary.OverrideHomeReferences(home))
 
 	writeGroupMarker(t, root, "workspace-root")
 
--- a/internal/mcp/cwd_home_boundary_6548_test.go
+++ b/internal/mcp/cwd_home_boundary_6548_test.go
@@ -77,20 +77,30 @@ func TestGroupFromCWD_FindsMarkerAtHome(t *testing.T) {
 }
 
 // TestGroupFromCWD_OutsideHomeStillClimbs guards the other permissive failure:
-// a start path that is NOT inside $HOME (a /var/folders temp dir, a /opt
-// checkout, another user's tree) must keep climbing. A boundary that fires on
-// any directory that merely LOOKS like a home — a child of /Users, /home — would
-// silently stop a repo from resolving its group.
+// a start path that is NOT inside any home (a /var/folders temp dir, an /opt or
+// /srv checkout) must keep climbing. A boundary that fired on every ancestor
+// would silently stop such a repo from resolving its group.
+//
+// #6548 requirement 3 (owner decision 2026-09-02) INVERTED half of this test:
+// it used to start under <root>/Users/someone — a SIBLING of the current
+// user's home — and require the marker above the Users level to be found.
+// That case is now a refusal and is asserted as one below.
 func TestGroupFromCWD_OutsideHomeStillClimbs(t *testing.T) {
 	root, _ := fakeHomeUnder(t, "Users", "me")
 
-	// Start under a DIFFERENT users-style tree, with the marker ABOVE the
-	// "Users" level — the exact spot a home-LOOKING boundary would cut off.
 	writeGroupMarker(t, root, "workspace-root")
 
-	start := mkdir(t, filepath.Join(root, "Users", "someone", "src", "deep", "pkg"))
+	start := mkdir(t, filepath.Join(root, "srv", "ci", "src", "deep", "pkg"))
 	if got := groupFromCWD(start); got != "workspace-root" {
-		t.Fatalf("groupFromCWD stopped before a legitimate marker outside $HOME (a home-LOOKING ancestor is not a boundary): got %q, want %q", got, "workspace-root")
+		t.Fatalf("groupFromCWD stopped before a legitimate marker outside every home: got %q, want %q", got, "workspace-root")
+	}
+
+	// The inverted case: the same marker, reached from another user's home,
+	// must be unreachable — the climb never enters /Users/someone at all.
+	sibling := mkdir(t, filepath.Join(root, "Users", "someone", "src", "deep", "pkg"))
+	if got := groupFromCWD(sibling); got != "" {
+		t.Fatalf("groupFromCWD climbed out of another user's home %s (site: internal/mcp/routing.go groupFromCWD): got %q, want \"\"",
+			filepath.Join(root, "Users", "someone"), got)
 	}
 }
 
--- a/internal/pathboundary/other_user_home_6548_test.go
+++ b/internal/pathboundary/other_user_home_6548_test.go
@@ -1,0 +1,176 @@
+package pathboundary
+
+import (
+	"path/filepath"
+	"runtime"
+	"testing"
+)
+
+// #6548 requirement 3 — grafel must not traverse into a home directory that is
+// not the current user's. Owner decision 2026-09-02 overrode the package doc's
+// former position that "another user's tree a user explicitly pointed grafel
+// at" must keep climbing.
+//
+// Grading note: this is a REFUSAL, not an emission, so recall cannot see it and
+// a boundary that refuses EVERYTHING looks identical from the graph to one that
+// refuses correctly. Every positive case below is therefore paired with an
+// under-firing control asserting a legitimate climb inside the user's OWN home
+// still works, and the tests observe WHAT WAS VISITED (i.e. read), not what was
+// produced.
+//
+// Every fixture is a TempDir with a synthetic /Users-like layout. No test here
+// reads, enumerates or stats anything under a real home directory.
+
+// usersLayout builds <tmp>/Users/{alice,bob,alice2} and returns the container.
+func usersLayout(t *testing.T, names ...string) (container string, homes map[string]string) {
+	t.Helper()
+	container = mk(t, filepath.Join(t.TempDir(), "Users"))
+	homes = make(map[string]string, len(names))
+	for _, n := range names {
+		homes[n] = mk(t, filepath.Join(container, n))
+	}
+	return container, homes
+}
+
+// TestClimb_RefusesAnotherUsersHome is the killing test for the Climb site: a
+// climb that starts inside a directory belonging to another user must visit
+// NOTHING — not the start path, not the other home, not the container above it.
+func TestClimb_RefusesAnotherUsersHome(t *testing.T) {
+	container, homes := usersLayout(t, "alice", "bob")
+	start := mk(t, filepath.Join(homes["bob"], "proj", "pkg"))
+
+	seen := visited(t, start, homes["alice"])
+	if len(seen) != 0 {
+		t.Fatalf("Climb READ another user's home (site: internal/pathboundary/pathboundary.go Climb): started at %q under %q while the current user's home is %q; visited %v",
+			start, homes["bob"], homes["alice"], seen)
+	}
+
+	// The refusal is observable through the return value too, not only the log:
+	// a marker inside the other user's tree must be unreachable.
+	marker := mk(t, filepath.Join(homes["bob"], "proj"))
+	if ClimbWithHome(start, homes["alice"], func(dir string) bool { return samePath(dir, marker) }) {
+		t.Fatalf("Climb resolved a marker inside another user's home %q", marker)
+	}
+	_ = container
+}
+
+// TestClimb_OwnHomeStillClimbs is the UNDER-FIRING CONTROL. A boundary that
+// refuses unconditionally passes the test above and fails this one.
+func TestClimb_OwnHomeStillClimbs(t *testing.T) {
+	_, homes := usersLayout(t, "alice", "bob")
+	proj := mk(t, filepath.Join(homes["alice"], "proj"))
+	start := mk(t, filepath.Join(proj, "pkg", "deep"))
+
+	seen := visited(t, start, homes["alice"])
+	want := []string{start, filepath.Join(proj, "pkg"), proj, homes["alice"]}
+	if len(seen) != len(want) {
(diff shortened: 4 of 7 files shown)


cajasmota:
## CI: Windows RED, and it is a real production defect — not a test-environment artefact

macOS and Ubuntu pass. Windows fails **26 tests across five packages** (`internal/cli`, `internal/gitmeta`, `internal/mcp`, `internal/pathboundary`, `internal/daemon`).

**Windows was green before this change** — `workflow_dispatch` on `main` at `a6f86acc9` was a full 3-platform success, and every PR merged this milestone was green on Windows. This is ours.

(Aside on the matrix: `test.yml:160` runs PRs on ubuntu+windows only, so the macOS leg above exists solely because I added the `ci:full` label. Worth doing on any change with `runtime.GOOS`-conditional behaviour — this one has three such switches.)

### Root cause: `C:\Users` is an unconditional container, and 8.3 short names break the identity comparison

`platformHomeContainers()` (`otherhome.go`) returns `C:\Users` on Windows **unconditionally**, independent of where the current user's home actually is. On the runner:

- `TempDir` is `C:\Users\RUNNER~1\AppData\Local\Temp\...`
- so `C:\Users\RUNNER~1` has a container parent (`C:\Users`) and a name that is not on the not-a-home list → **classified as another user's home**
- the current user's home from `user.Current()` is the **long** form, `C:\Users\runneradmin`
- `samePath("C:\Users\RUNNER~1", "C:\Users\runneradmin")` is false, so the current user's own home is not recognised as their own

`underOtherUserHome` is at-or-under, so **everything beneath the temp root is refused**. That is exactly what the diagnostics say:

```
other_user_home_6548_test.go:55: HasGitDirInTree failed to find
  "C:\Users\RUNNER~1\...\Users\alice\proj\.git" inside the current user's own home
  — the boundary refuses everything
```

and why unrelated tests fall over with it — `TestHasGitDirInTree_OutsideHomeStillClimbs` uses `\srv\ci\src\deep`, which is *also* under the temp root, and `TestCheckRepoGitDetection` reports `missing .git warning = true, want false` because the climb that would have found `.git` was refused.

`bothSpellings` holds the literal and the `EvalSymlinks`-resolved spelling. Whatever Go's Windows `EvalSymlinks` does here, it demonstrably does not unify `RUNNER~1` with `runneradmin` — so the short-name spelling is a third spelling the reference set does not hold.

### Why this is production, not CI

8.3 short names are not a CI curiosity. `%TEMP%` commonly resolves to a short form, `GetShortPathName` output flows through legacy tooling, and any path with a space or over eight characters in a component has one. A Windows user whose path reaches grafel in short form has **their own home classified as another user's**, and the boundary then refuses every climb under it — grafel declines to index the user's own repositories. That is a full denial of service on the user's own machine, caused by a privacy boundary.

### The design question underneath, which I want answered rather than patched around

Adding a short-name spelling to `bothSpellings` fixes this instance. It does not answer the general question: **when the boundary cannot reliably decide whether a directory is the current user's own home, which way should it fail?**

Today it fails closed, and closed here means refusing the user's own files. For the *foreign* home case, fail-closed is obviously right. For the *is-this-mine* case, the harm being prevented is zero — the files are the user's own — and the cost is total. Those are different questions and this change answers them with one rule.

So the fix should state, and pin, what happens when the current user's home cannot be resolved at all (`user.Current()` failing, an empty `HomeDir`, a static binary with no passwd row). The package doc already claims that case is "a narrowing of this boundary, never a widening of it" — that claim needs a test, and on the evidence above the narrowing can be total.

### Other failures to triage separately, not assumed to be the same cause

- `internal/daemon`: `TestCanonicalizePathTimesOutAndDegrades` — `canonicalizePath took 2.54s; expected prompt return under the 10s block`
- `internal/daemon/watch`: FAIL with no test named in the summary lines

Both are under the temp root too, so the same root cause is plausible — but plausible is not measured. Confirm each individually rather than folding them in.

Not merging until Windows is green.

cajasmota:
## Windows fix at `91933c2` — root cause held in outcome, was wrong about the mechanism

Your diagnosis of the **symptom** was exactly right: the running user's own home is classified as another user's, and `underOtherUserHome` being at-or-under then refuses everything beneath the temp root. Confirmed against the CI log (`gh run view 33704615270 --log-failed`), which shows the temp root as `C:\Users\RUNNER~1\...` on every failing package.

**Refuted:** *"whatever Go's Windows `EvalSymlinks` does here, it demonstrably does not unify `RUNNER~1` with `runneradmin`."* It does. `filepath.EvalSymlinks` on Windows is `walkSymlinks` followed by `toNorm`, and `toNorm` normalises **every component** through `normBase`, which is `FindFirstFile` returning `Win32finddata.FileName` — the long name (`path/filepath/symlink_windows.go`; `TestToNorm` in `path_windows_test.go` pins the per-component normalisation). `os/user` on Windows returns `GetUserProfileDirectory`, also long. So both sides canonicalise to `C:\Users\runneradmin`.

**The actual mechanism** is one line in `classifyDir`:

```go
if underOtherUserHome(dir, refs) || underOtherUserHome(resolved, refs) {
```

The refusal is a **disjunction over spellings** while the identity exemption is a lookup in a set that holds only two. `resolved` said "ours" and was overruled by `dir`, which said "foreign". D1 (the symlink-inert bug) was fixed by holding both spellings in `refs` *and* testing both spellings of `dir`; only the first half was needed, and the second half is what shipped the DoS.

The read lands at the canonical location, so the canonical location decides. `classifyDir` now tests `resolved` and only `resolved` — `resolveOrSelf` already falls back to the literal path when resolution fails, and `refs` still holds both spellings, so nothing D1 caught is let through.

### Second, independent defect: the fixtures, not just production

Even with (1), every fixture using `OverrideHomeReferences` still failed on Windows, for a reason that is invisible on darwin and linux: **`t.TempDir()` on Windows lives inside the running user's home.** A fixture that *replaced* the reference set with `<tmp>/Users/alice` turned `C:\Users\runneradmin` — an **ancestor** of the fixture — into another user's home. On darwin (`/var/folders`) and linux (`/tmp`) the temp root is outside every home container, so the whole failure mode is unobservable.

`OverrideHomeReferences` is now variadic and every fixture passes `RealUIDHomeForTest()` alongside its synthetic home.

## Your design question: which way should "is this mine?" fail

**They are two questions and they must fail in opposite directions.** Implemented and pinned:

- *"Is this a **foreign** home?"* — **fail closed.** Refusing wrongly costs an un-indexed directory; allowing wrongly reads another user's files. Unchanged.
- *"Is this **my own** home?"* — **fail open.** Refusing wrongly denies the user their own files and the harm prevented is zero, because the files are theirs.

Concretely: **with no reference for the current user's home at all, the class stands down** — `resolveHomeReferences` returns an empty set, so no directory on the machine is classified. Grafel then behaves exactly as it did before requirement 3: no new exposure is invented, an existing guarantee is simply not offered on a host whose premise does not hold.

The doc claim you flagged — *"a narrowing of this boundary, never a widening of it"* — was true and useless. The narrowing is **total**: with no left-hand side for "not MY home", every home under `/Users`, `/home` or `C:\Users` is somebody else's, the user's own included. It had no test; it now has `TestOtherHome_NoHomeReferenceStandsDown`, asserted in **both** directions (the with-a-reference control is what keeps the stand-down assertion from being vacuous), and the doc says what the code does.

**Flagged for you rather than decided quietly:** the residual exposure is real and bounded — a host with no password-database row for the running uid (cgo-less static binary in a scratch container) gets no other-user-home boundary at all. I judged that better than refusing the user their own machine, and wrote the reversal down: it is a one-line change in `resolveHomeReferences` plus the test. Overrule me if you read the balance differently.

## Mutants — permissive direction first, `go vet` before every score

|---|---|---|---|
| M1 | delete the `otherHome` check at the call site (never fires) | **DEAD** | 10 tests across `pathboundary` + `gitmeta` |
| M2 | reference set always empty (class always stands down) | **DEAD** | same 10 |
| M6 | at-only instead of at-or-under (the laundering reading) | **DEAD** | 10 incl. `TestHasGitDirInTree_RefusesAnotherUsersHome` |
| M3 | restore the shipped `dir \|\| resolved` OR — i.e. re-introduce this bug | **DEAD** | `TestClimb_OwnHomeUnderASecondSpellingIsStillOurs` |
| M4 | decide on the literal spelling only | **DEAD** | `TestClimb_SymlinkedPathIntoAnotherUsersHome` + the above |
| M5 | delete the stand-down (fail **closed** on an unknown home) | **DEAD** | `TestOtherHome_NoHomeReferenceStandsDown` |
| M8 | `OverrideHomeReferences` keeps only `homes[0]` | **DEAD** | `TestOtherHome_EveryReferenceInTheSetIsHonoured` |
| M7 | `RealUIDHomeForTest` returns `""` | **ALIVE — not pinnable off Windows** | — |

M7 is reported as alive, not scored. It only bites where `t.TempDir()` sits inside a home container, which is Windows and nowhere else; the function is a one-line passthrough to `realUIDHome`, so the only portable assertion available is tautological. Windows CI is its test.

Note that M1/M2/M6 also kill `TestOtherHome_NoHomeReferenceStandsDown` — via its *with-a-reference* control, not its stand-down assertion. That is the control doing its job.

### New tests, and what each varies

- **`TestClimb_OwnHomeUnderASecondSpellingIsStillOurs`** — varies the **spelling** used to reach the current user's own home (canonical vs. a second name for the same directory sitting in the same container); holds the layout, reference set, foreign home and climb bound constant. **It does not reproduce the 8.3 short name** — a symlink is not a short name, and this is stated in the test. It reproduces the *class*: one directory, two names, one of them absent from the reference set. Over-firing controls in the same fixture: a foreign home by its canonical name, **and** a foreign home reached through a symlink of its own, both still refused.
- **`TestOtherHome_NoHomeReferenceStandsDown`** — varies whether a reference exists at all; holds the path, the container and the bound constant. Deliberately uses the **platform** container, because with no reference a synthetic container is not derived either and a TempDir layout would pass vacuously. The name belongs to no user, is never created, and the climb stops at the first level.
- **`TestOtherHome_EveryReferenceInTheSetIsHonoured`** — varies how many homes the set holds. Pins the property the Windows fix rests on (every member contributes its identity *and* its container), and says explicitly that it does **not** reproduce the Windows topology, which needs the enclosing container to be the platform's own absolute path.

The stale claim at the top of `other_user_home_6548_test.go` ("no test here reads, enumerates or stats anything under a real home directory") is corrected rather than left standing — on Windows the temp root *is* under the real home.

## Triage of the two `internal/daemon` failures — measured, and **neither is ours**

**`internal/daemon/watch`** — not a test failure at all. It is a Go **runtime fatal**:

```
watcher: error err="CancelIo: The handle is invalid."
fatal error: runtime.semawakeup
... watch.(*Watcher).Stop-fm.(*Watcher).Stop.func1() watcher.go:1147
```

No `pathboundary` frame anywhere in the dump, and `internal/daemon/watch` has no call into the package (its git work goes through `gitmeta.Capture`/`ResolveCommonDir`, not `HasGitDirInTree`). This is a Windows watcher-teardown crash. Green locally (`./internal/daemon/watch/` ok, 59s). Re-run the leg; if it reproduces it is an independent bug and wants its own issue.

**`TestCanonicalizePathTimesOutAndDegrades`** — also not ours, and the log says why. `canonicalizePath` walks the path **segment by segment**, and each segment's stubbed `ReadDir` blocks until `readDirBounded` cancels it:

```
WARN canonicalizePath: os.ReadDir timed out ... dir=C:\
WARN ... dir=C:\Users
WARN ... dir=C:\Users\RUNNER~1
WARN ... dir=C:\Users\RUNNER~1\AppData
... (10 segments, 01:56:01 → 01:56:03)
```

Ten segments at ~250ms of actual cancel-to-reschedule latency each = 2.54s against a hardcoded 2s budget. The Windows temp path is ten segments deep *because* it is `C:\Users\<user>\AppData\Local\Temp\...`; on linux `/tmp/...` it is four. `internal/daemon` does not import `pathboundary`, and `canonicalizePath` does its own `ReadDir` walk — the elapsed time is entirely inside the injected stub. This is a marginal, load-sensitive budget (segment count × cancel latency vs. a fixed 2s), not a regression from this change. It wants its own fix — a budget proportional to segment count — and I have not touched it here rather than guess at a threshold.

## Verification

`gofmt` clean, `go vet` clean, `go.mod`/`go.sum` untouched, no extractor touched. `-count=1` green on `./internal/pathboundary/ ./internal/gitmeta/ ./internal/mcp/ ./internal/cli/ ./internal/registry/` and `./internal/daemon/watch/`. Every mutant `go vet`-ed before scoring and restored by `cp` with a verified `shasum`.

Windows is the only place the fix can actually be observed end-to-end, so this needs the `ci:full` leg before it means anything.

---

## Coordinator-verified, a direction none of the eight mutants took

My root cause for this failure was **right on the symptom and wrong on the mechanism**, and the fix's author demonstrated that from the Go source rather than asserting it: Windows `EvalSymlinks` *does* unify `RUNNER~1` with `runneradmin` (`toNorm` → `normBase` → `FindFirstFile`, returning the long `Win32finddata.FileName` per component), and `os/user` returns the long form too. Both sides canonicalise. The defect was that `classifyDir` refused when **either** spelling looked foreign — the literal short-name spelling overruled a correct resolved verdict. Holding both spellings in `refs` was right; testing both spellings of `dir` is what shipped the DoS.

That makes the fix's remaining load-bearing assumption a **symmetry**: it now decides on `resolved` only, so the reference set MUST hold the resolved spelling of our own home. If it held only the literal, our own home would fail to match a resolved `dir` — the identical DoS wearing macOS clothes instead of Windows.

So I dropped the resolved spelling from `refs.current` alone, leaving it in place for containers:

```go
- bothSpellings(&refs.current, home)
+ appendUnique(&refs.current, home)
```

**DEAD on five tests across two packages** — `TestClimb_OwnHomeStillClimbs`, `TestClimb_RealUIDHomeIsOursEvenWhenHOMEDisagrees`, `TestClimb_OwnHomeUnderASecondSpellingIsStillOurs`, `TestOtherHome_EveryReferenceInTheSetIsHonoured`, and `gitmeta`'s `TestHasGitDirInTree_RefusesAnotherUsersHome`. The third names the property directly:

```
the current user's OWN home ".../Users/alice", reached as ".../Users/alice-second-spelling",
was classified as another user's ... A path can name one directory in more than one way
(a symlink here; a Windows 8.3 short name in #6787) and the identity comparison must
resolve before it compares.
```

That the symmetry is pinned on darwin — where the CI leg that would catch it exists only because this PR carries `ci:full` — is what makes the Windows fix portable rather than a Windows patch.

`go vet` clean before scoring. Restored by `cp` to shasum `3e8cf066`, tree clean at `91933c2`.

## On M7, reported ALIVE rather than scored

The author declined to score `RealUIDHomeForTest → ""` as DEAD because it is observable only where `TempDir` sits inside a home container — i.e. Windows. That is the correct call: a mutant that cannot be distinguished by the suite is *equivalent under the current suite*, not dead, and recording it honestly is worth more than a table of twelve kills. Same for the 8.3 short name itself, where the new test reproduces the *class* — one directory, two names — via a symlink and says so in its own diagnostic.

cajasmota:
## Windows fix at `91933c2` — root cause held in outcome, was wrong about the mechanism

Your diagnosis of the **symptom** was exactly right: the running user's own home is classified as another user's, and `underOtherUserHome` being at-or-under then refuses everything beneath the temp root. Confirmed against the CI log (`gh run view 33704615270 --log-failed`), which shows the temp root as `C:\Users\RUNNER~1\...` on every failing package.

**Refuted:** *"whatever Go's Windows `EvalSymlinks` does here, it demonstrably does not unify `RUNNER~1` with `runneradmin`."* It does. `filepath.EvalSymlinks` on Windows is `walkSymlinks` followed by `toNorm`, and `toNorm` normalises **every component** through `normBase`, which is `FindFirstFile` returning `Win32finddata.FileName` — the long name (`path/filepath/symlink_windows.go`; `TestToNorm` in `path_windows_test.go` pins the per-component normalisation). `os/user` on Windows returns `GetUserProfileDirectory`, also long. So both sides canonicalise to `C:\Users\runneradmin`.

**The actual mechanism** is one line in `classifyDir`:

```go
if underOtherUserHome(dir, refs) || underOtherUserHome(resolved, refs) {
```

The refusal is a **disjunction over spellings** while the identity exemption is a lookup in a set that holds only two. `resolved` said "ours" and was overruled by `dir`, which said "foreign". D1 (the symlink-inert bug) was fixed by holding both spellings in `refs` *and* testing both spellings of `dir`; only the first half was needed, and the second half is what shipped the DoS.

The read lands at the canonical location, so the canonical location decides. `classifyDir` now tests `resolved` and only `resolved` — `resolveOrSelf` already falls back to the literal path when resolution fails, and `refs` still holds both spellings, so nothing D1 caught is let through.

### Second, independent defect: the fixtures, not just production

Even with (1), every fixture using `OverrideHomeReferences` still failed on Windows, for a reason that is invisible on darwin and linux: **`t.TempDir()` on Windows lives inside the running user's home.** A fixture that *replaced* the reference set with `<tmp>/Users/alice` turned `C:\Users\runneradmin` — an **ancestor** of the fixture — into another user's home. On darwin (`/var/folders`) and linux (`/tmp`) the temp root is outside every home container, so the whole failure mode is unobservable.

`OverrideHomeReferences` is now variadic and every fixture passes `RealUIDHomeForTest()` alongside its synthetic home.

## Your design question: which way should "is this mine?" fail

**They are two questions and they must fail in opposite directions.** Implemented and pinned:

- *"Is this a **foreign** home?"* — **fail closed.** Refusing wrongly costs an un-indexed directory; allowing wrongly reads another user's files. Unchanged.
- *"Is this **my own** home?"* — **fail open.** Refusing wrongly denies the user their own files and the harm prevented is zero, because the files are theirs.

Concretely: **with no reference for the current user's home at all, the class stands down** — `resolveHomeReferences` returns an empty set, so no directory on the machine is classified. Grafel then behaves exactly as it did before requirement 3: no new exposure is invented, an existing guarantee is simply not offered on a host whose premise does not hold.

The doc claim you flagged — *"a narrowing of this boundary, never a widening of it"* — was true and useless. The narrowing is **total**: with no left-hand side for "not MY home", every home under `/Users`, `/home` or `C:\Users` is somebody else's, the user's own included. It had no test; it now has `TestOtherHome_NoHomeReferenceStandsDown`, asserted in **both** directions (the with-a-reference control is what keeps the stand-down assertion from being vacuous), and the doc says what the code does.

**Flagged for you rather than decided quietly:** the residual exposure is real and bounded — a host with no password-database row for the running uid (cgo-less static binary in a scratch container) gets no other-user-home boundary at all. I judged that better than refusing the user their own machine, and wrote the reversal down: it is a one-line change in `resolveHomeReferences` plus the test. Overrule me if you read the balance differently.

## Mutants — permissive direction first, `go vet` before every score

|---|---|---|---|
| M1 | delete the `otherHome` check at the call site (never fires) | **DEAD** | 10 tests across `pathboundary` + `gitmeta` |
| M2 | reference set always empty (class always stands down) | **DEAD** | same 10 |
| M6 | at-only instead of at-or-under (the laundering reading) | **DEAD** | 10 incl. `TestHasGitDirInTree_RefusesAnotherUsersHome` |
| M3 | restore the shipped `dir \|\| resolved` OR — i.e. re-introduce this bug | **DEAD** | `TestClimb_OwnHomeUnderASecondSpellingIsStillOurs` |
| M4 | decide on the literal spelling only | **DEAD** | `TestClimb_SymlinkedPathIntoAnotherUsersHome` + the above |
| M5 | delete the stand-down (fail **closed** on an unknown home) | **DEAD** | `TestOtherHome_NoHomeReferenceStandsDown` |
| M8 | `OverrideHomeReferences` keeps only `homes[0]` | **DEAD** | `TestOtherHome_EveryReferenceInTheSetIsHonoured` |
| M7 | `RealUIDHomeForTest` returns `""` | **ALIVE — not pinnable off Windows** | — |

M7 is reported as alive, not scored. It only bites where `t.TempDir()` sits inside a home container, which is Windows and nowhere else; the function is a one-line passthrough to `realUIDHome`, so the only portable assertion available is tautological. Windows CI is its test.

Note that M1/M2/M6 also kill `TestOtherHome_NoHomeReferenceStandsDown` — via its *with-a-reference* control, not its stand-down assertion. That is the control doing its job.

### New tests, and what each varies

- **`TestClimb_OwnHomeUnderASecondSpellingIsStillOurs`** — varies the **spelling** used to reach the current user's own home (canonical vs. a second name for the same directory sitting in the same container); holds the layout, reference set, foreign home and climb bound constant. **It does not reproduce the 8.3 short name** — a symlink is not a short name, and this is stated in the test. It reproduces the *class*: one directory, two names, one of them absent from the reference set. Over-firing controls in the same fixture: a foreign home by its canonical name, **and** a foreign home reached through a symlink of its own, both still refused.
- **`TestOtherHome_NoHomeReferenceStandsDown`** — varies whether a reference exists at all; holds the path, the container and the bound constant. Deliberately uses the **platform** container, because with no reference a synthetic container is not derived either and a TempDir layout would pass vacuously. The name belongs to no user, is never created, and the climb stops at the first level.
- **`TestOtherHome_EveryReferenceInTheSetIsHonoured`** — varies how many homes the set holds. Pins the property the Windows fix rests on (every member contributes its identity *and* its container), and says explicitly that it does **not** reproduce the Windows topology, which needs the enclosing container to be the platform's own absolute path.

The stale claim at the top of `other_user_home_6548_test.go` ("no test here reads, enumerates or stats anything under a real home directory") is corrected rather than left standing — on Windows the temp root *is* under the real home.

## Triage of the two `internal/daemon` failures — measured, and **neither is ours**

**`internal/daemon/watch`** — not a test failure at all. It is a Go **runtime fatal**:

```
watcher: error err="CancelIo: The handle is invalid."
fatal error: runtime.semawakeup
... watch.(*Watcher).Stop-fm.(*Watcher).Stop.func1() watcher.go:1147
```

No `pathboundary` frame anywhere in the dump, and `internal/daemon/watch` has no call into the package (its git work goes through `gitmeta.Capture`/`ResolveCommonDir`, not `HasGitDirInTree`). This is a Windows watcher-teardown crash. Green locally (`./internal/daemon/watch/` ok, 59s). Re-run the leg; if it reproduces it is an independent bug and wants its own issue.

**`TestCanonicalizePathTimesOutAndDegrades`** — also not ours, and the log says why. `canonicalizePath` walks the path **segment by segment**, and each segment's stubbed `ReadDir` blocks until `readDirBounded` cancels it:

```
WARN canonicalizePath: os.ReadDir timed out ... dir=C:\
WARN ... dir=C:\Users
WARN ... dir=C:\Users\RUNNER~1
WARN ... dir=C:\Users\RUNNER~1\AppData
... (10 segments, 01:56:01 → 01:56:03)
```

Ten segments at ~250ms of actual cancel-to-reschedule latency each = 2.54s against a hardcoded 2s budget. The Windows temp path is ten segments deep *because* it is `C:\Users\<user>\AppData\Local\Temp\...`; on linux `/tmp/...` it is four. `internal/daemon` does not import `pathboundary`, and `canonicalizePath` does its own `ReadDir` walk — the elapsed time is entirely inside the injected stub. This is a marginal, load-sensitive budget (segment count × cancel latency vs. a fixed 2s), not a regression from this change. It wants its own fix — a budget proportional to segment count — and I have not touched it here rather than guess at a threshold.

## Verification

`gofmt` clean, `go vet` clean, `go.mod`/`go.sum` untouched, no extractor touched. `-count=1` green on `./internal/pathboundary/ ./internal/gitmeta/ ./internal/mcp/ ./internal/cli/ ./internal/registry/` and `./internal/daemon/watch/`. Every mutant `go vet`-ed before scoring and restored by `cp` with a verified `shasum`.

Windows is the only place the fix can actually be observed end-to-end, so this needs the `ci:full` leg before it means anything.

---

## Coordinator-verified, a direction none of the eight mutants took

My root cause for this failure was **right on the symptom and wrong on the mechanism**, and the fix's author demonstrated that from the Go source rather than asserting it: Windows `EvalSymlinks` *does* unify `RUNNER~1` with `runneradmin` (`toNorm` → `normBase` → `FindFirstFile`, returning the long `Win32finddata.FileName` per component), and `os/user` returns the long form too. Both sides canonicalise. The defect was that `classifyDir` refused when **either** spelling looked foreign — the literal short-name spelling overruled a correct resolved verdict. Holding both spellings in `refs` was right; testing both spellings of `dir` is what shipped the DoS.

That makes the fix's remaining load-bearing assumption a **symmetry**: it now decides on `resolved` only, so the reference set MUST hold the resolved spelling of our own home. If it held only the literal, our own home would fail to match a resolved `dir` — the identical DoS wearing macOS clothes instead of Windows.

So I dropped the resolved spelling from `refs.current` alone, leaving it in place for containers:

```go
- bothSpellings(&refs.current, home)
+ appendUnique(&refs.current, home)
```

**DEAD on five tests across two packages** — `TestClimb_OwnHomeStillClimbs`, `TestClimb_RealUIDHomeIsOursEvenWhenHOMEDisagrees`, `TestClimb_OwnHomeUnderASecondSpellingIsStillOurs`, `TestOtherHome_EveryReferenceInTheSetIsHonoured`, and `gitmeta`'s `TestHasGitDirInTree_RefusesAnotherUsersHome`. The third names the property directly:

```
the current user's OWN home ".../Users/alice", reached as ".../Users/alice-second-spelling",
was classified as another user's ... A path can name one directory in more than one way
(a symlink here; a Windows 8.3 short name in #6787) and the identity comparison must
resolve before it compares.
```

That the symmetry is pinned on darwin — where the CI leg that would catch it exists only because this PR carries `ci:full` — is what makes the Windows fix portable rather than a Windows patch.

`go vet` clean before scoring. Restored by `cp` to shasum `3e8cf066`, tree clean at `91933c2`.

## On M7, reported ALIVE rather than scored

The author declined to score `RealUIDHomeForTest → ""` as DEAD because it is observable only where `TempDir` sits inside a home container — i.e. Windows. That is the correct call: a mutant that cannot be distinguished by the suite is *equivalent under the current suite*, not dead, and recording it honestly is worth more than a table of twelve kills. Same for the 8.3 short name itself, where the new test reproduces the *class* — one directory, two names — via a symlink and says so in its own diagnostic.

---

# Round 3 — `87fbea0`: the unresolvable branch

Windows went 26 → 8 failures. **Your D1 mechanism held exactly as written**, and I confirmed it against the code before touching anything: `resolveOrSelf` (`pathboundary.go`) returns the literal `path` when `EvalSymlinks` errors, and `EvalSymlinks` errors outright if any component is missing. Round two removed the literal leg because it could *overrule* a correct resolved verdict; on this branch there is no resolved verdict and the literal is the only input. Same defect, narrower branch.

The evidence you flagged is decisive on its own — `resolvedeepest_boundary_6548_test.go:78` reports `got` in the **short** spelling and `want` in the **long** one, which only happens if the resolved and unresolved branches disagree about what the path is called.

## The fix: "cannot resolve" is its own case

`canonicalForIdentity(dir, resolved)` in `pathboundary.go`. When `resolveOrSelf` actually resolved, it is returned untouched — zero extra filesystem calls on the common path. When it did not, the deepest **existing** ancestor is resolved and the missing tail re-appended. That is the shape `registry.resolveDeepestExisting` already uses for exactly this reason; it cannot be reused here because it climbs through `Climb`, which is this function's caller.

Two things I checked rather than assumed:

- **Re-appending the tail is load-bearing, not decoration.** A home that does not exist *yet* is still a home. Classifying `<container>/ghost/x` from the resolved container alone would find nobody's home there and let the climb through. Pinned, and mutant R3 dies on it.
- **The scope is surgical.** `resolved` is still what `mayBeProtected` receives; only the other-user-home comparison uses the canonical spelling. Widening the TCC/media gate's input is a separate question with a separate fail direction, and I did not fold it in.

Per the fail-direction rule from round 2: this can only make the class fire **less**. The missing components are below the home, and it is the ancestors that decide.

## `OverrideHomeReferences` audit — 12 call sites, one bug

| Call site | Passes `RealUIDHomeForTest()` | Verdict |
|---|---|---|
| `gitmeta/gitdir_home_boundary_6548_test.go:77` | yes | ok |
| `gitmeta/other_user_home_6548_test.go:36, :67` | yes | ok |
| `mcp/cwd_home_boundary_6548_test.go:93` | yes | ok |
| `pathboundary/other_user_home_6548_test.go:47, :215, :232, :264` | yes | ok |
| `pathboundary/other_user_home_6548_test.go:390` | no | **deliberate** — classifies a path under the platform container, no temp path involved |
| `pathboundary/other_user_home_6548_test.go:399` | no | **deliberate** — this *is* the no-reference case |
| `pathboundary/other_user_home_6548_test.go:435, :445` | no | **D2, fixed** |

The two legs of `TestOtherHome_EveryReferenceInTheSetIsHonoured` now both carry it, so the axis under test stays "how many synthetic homes" rather than silently becoming "is the process's real home referenced".

## D3 triage — confirmed individually, and one of the two is not a failure at all

**`TestGroupFromRegistryWithCandidates_CaseInsensitive` — D1, measured.** The fixture creates `<tmp>/Acme/acme_core` and then uses `<tmp>/acme/acme_core/src/views` as cwd. `src/views` is **never created**, so the child does not exist. `mcp.pathContains` (`routing.go:148`) resolves the ancestor, fails on the child, and takes the `ancestorResolved && !childResolved` branch, which climbs from the child's parent through `pathboundary.Climb` to find a resolvable ancestor. That climb was refused at level 0, so containment could not be established, `hits` stayed empty, and the candidate list came back empty. Not a shared cause assumed from proximity — the code path is the D1 path.

**The `wrap_panic_test.go` panic — refuted, it is not a failure.** `TestWrapRecoversPanic` does not appear in any `--- FAIL` line. The stack is the deliberate diagnostic `Server.wrap` prints when it recovers, from a tool the test registers as `grafel_boom` specifically to panic. Expected output of a passing test.

That accounts for all 8: seven D1 (`registry` ×4, `mcp` ×1, plus the two `pathcontains_*` cases), one D2.

## Mutants — one per branch, `go vet` before each, `-count=1`

|---|---|---|---|
| R1 | revert to round 2 — classify from `resolveOrSelf`, literal on failure (**this bug**) | **DEAD** | `TestClimb_UnresolvablePathIsClassifiedFromItsDeepestExistingAncestor` |
| R2 | canonical spelling always empty — class never fires (permissive) | **DEAD** | 7 tests |
| R3 | drop the missing tail — classify from the resolved ancestor alone (permissive) | **DEAD** | the ghost-home assertion + `TestOtherHome_NoHomeReferenceStandsDown` |
| R5 | fallback walk runs zero levels | **DEAD** | `TestClimb_UnresolvablePathIsClassifiedFromItsDeepestExistingAncestor` |
| R4 | drop the `resolved != Clean(dir)` early-out | **EQUIVALENT — not scored** | see below |
| R6 | `RealUIDHomeForTest` returns `""` | **ALIVE — not pinnable off Windows** | — |

**R4 is provably equivalent, not merely indistinguishable.** The early-out is a pure optimisation: when it is removed, control reaches `EvalSymlinks(dir)`, which by construction succeeded (that is why `resolved != Clean(dir)`), so the function returns the same `resolved`. One extra syscall, identical value. No test can distinguish it and none should be written to.

**R6 is reported alive, not scored DEAD.** It only bites where `t.TempDir()` sits inside a home container — Windows and nowhere else. Same call as round 2.

## New test — both directions, because this class has now failed both ways

`TestClimb_UnresolvablePathIsClassifiedFromItsDeepestExistingAncestor`. **Axis varied:** whose home the not-yet-created path sits under. **Held constant:** the layout, the reference set, the missing tail, the climb bound, and — asserted by an `os.Lstat` guard on every case — the fact that no component of the tail exists, so the test cannot quietly stop exercising the unresolvable branch.

Four cases, in both directions:

- **ours** — a missing path under the current user's own home reached under a second spelling **climbs** (the DoS);
- **theirs** — the identical shape under another user's home is **refused** (without this, the first case is satisfied by a boundary that stopped firing on missing paths altogether);
- **theirs, aliased** — a missing path under a symlink whose name looks innocent is still **refused**, so resolving the ancestor folds an alias onto its target rather than becoming a bypass;
- **ghost home** — a home-shaped child of the container that merely does not exist yet is still **refused**.

Plus an under-firing control for the plain canonical case, so the fix cannot be read as "only aliased paths are handled". The 8.3 short name itself is still not reproduced off Windows and the test says so; what it reproduces is the class — one directory, more than one name, and a spelling the reference set does not hold.

## Verification

`gofmt` clean, `go vet` clean, `go.mod`/`go.sum` untouched, no extractor touched. `-count=1` green on `./internal/pathboundary/ ./internal/gitmeta/ ./internal/mcp/ ./internal/cli/ ./internal/registry/`. Every mutant `go vet`-ed before scoring, backed up byte-for-byte and restored by `cp` with a verified `shasum`; no `git checkout --` anywhere.

`internal/daemon/watch` left alone per #6790.

---

## Coordinator round 3: the deliberate asymmetry is unpinned — reported, not blocking

Round 3 routes the other-user-home check through `canonicalForIdentity(dir, resolved)` but deliberately leaves `mayBeProtected(dir, resolved, ...)` on the **old** `resolved` value, with the reasoning stated: widening the TCC gate's input is a separate question with a separate fail direction. That is a defensible call, and it is exactly the kind of deliberate asymmetry worth testing, because nothing distinguishes "chosen" from "overlooked" once it is in the file.

```go
- if !mayBeProtected(dir, resolved, home, homeReal) {
+ if !mayBeProtected(dir, canonicalForIdentity(dir, resolved), home, homeReal) {
```

`go vet` clean, `-count=1`: `./internal/pathboundary/` and `./internal/registry/` both **exit 0**. **ALIVE** — the asymmetry is held by no test in either direction.

**Why this does not block.** `mayBeProtected(dir, resolved, ...)` is the pre-existing call; this PR does not change it or make it worse. Merging leaves the TCC gate exactly as it has always been.

**The hypothesis it raises, labelled as unverified — I did not build a repro.** The gate passes only when `dir` or `resolved` is at or under `$HOME`, or names a media bundle. For a path whose components do not all exist, `resolveOrSelf` falls back to the literal. So a **symlink outside `$HOME` whose target does not fully exist, but whose deepest existing ancestor is inside a protected tree**, would have `resolved` land outside `$HOME`, fail the cheap gate, and skip the TCC check — where the canonical spelling would have caught it. That is precisely the case `mayBeProtected`'s own comment says the gate resolves symlinks to catch. The ordinary case is unaffected: a non-existent path under `~/Library/...` still has a literal under `$HOME`, so the gate passes.

Filed separately rather than expanded here — this PR is already three review rounds deep on a different question, and widening a TCC input is a fail-direction decision that deserves its own grounding.

## On the two mutants the author did NOT score DEAD

Both calls are right and I want them on the record rather than buried in a table:

- **R4 recorded as EQUIVALENT with a proof**, not as a kill: removing the `resolved != Clean(dir)` early-out reaches `EvalSymlinks(dir)`, which by construction already succeeded, returning the same value — one extra syscall, identical result.
- **R6 reported ALIVE and explicitly not pinnable off Windows**, the same verdict as round 2 rather than a quietly upgraded one.

A mutant table with two honest non-kills in it is worth more than one with twelve kills, because it tells you where the suite actually ends.

## Also worth recording: one reported "failure" was refuted

The `wrap_panic_test.go` stack in the Windows log is **not a failure**. `TestWrapRecoversPanic` appears in no `--- FAIL` line; the stack is the deliberate diagnostic `Server.wrap` prints when recovering the `grafel_boom` tool it registers *in order to* panic. Triaging it as a defect would have cost a round. All 8 real failures are accounted for: **7 × D1, 1 × D2**.

PR #6798, opened 2026-09-03
fix(#6796): a parenthesised enum member value resolved to nothing, and member_count still counted the member

### LINKED ISSUE #6796: python: an enum member value wrapped in parentheses is dropped, and member_count still counts the member
I index a set of Python services with grafel and read `SCOPE.Enum` value sets to decide whether a route parameter accepts a given path literal.

`pythonEnumLiteralValue` (internal/extractors/python/types.go:543) switches on the member's RHS node type and handles `integer`, `float`, `string`, `true`/`false`/`none`, `tuple` and `expression_list`. tree-sitter-python parses a lone value wrapped in parentheses as `parenthesized_expression`, which no case matches, so the function returns `""` and the member is recorded with no value. `member_count` still counts the member, so the entity reports more members than it has values.

For example, given:

```python
import enum

class TransactionType(enum.StrEnum):
    SHORT = "short"
    A_VERY_LONG_MEMBER_NAME = (
        "some-value"
    )
```

the emitted entity carries `member_count` `2` and `values` `SHORT=short`.

Parenthesising is what a formatter does to a member whose name and value together exceed the line limit, so the members this hits are the longest-named ones rather than an arbitrary sample. It is distinct from the computed forms `pythonEnumLiteralValue` documents as deliberately value-less: this is a static string literal that the existing `string` case resolves as soon as the parentheses are gone.

The consumer-visible effect is that absence from `values` reads as "the enum declares no such value" when the real cause is a capture gap. `member_count` is the only signal that the set is short, and only if you compare the two.

arthurgeron:
Closes #6796.

`pythonEnumLiteralValue` (internal/extractors/python/types.go:543) had no `parenthesized_expression` case, so a member written as `NAME = (\n    "value"\n)` reached the value set with no value while `member_count` still counted it.

## Shape

One case, recursing on the parenthesised node's first named child, so the value is resolved by whichever case already handles it. The string, numeric, keyword and Django-tuple forms are unchanged whether or not parentheses wrap them.

## Evidence

Before, with the new test and the old `types.go`:

```
--- FAIL: TestEnumValueSet_PythonParenthesizedValue
    types_test.go:127: values = "SHORT=short, DJANGO_LIKE=db", want "SHORT=short, A_VERY_LONG_MEMBER_NAME=some-value, DJANGO_LIKE=db"
```

After:

```
ok  	github.com/cajasmota/grafel/internal/extractors/python	1.462s
ok  	github.com/cajasmota/grafel/internal/extractor	0.737s
```

The fixture also holds `enum.auto()` and a Django-style `("db", "Label")` tuple, so the test pins that the recursion does not give a value to a computed member and does not change which tuple element is stored.

## What this does not fix

`member_count` still counts members rather than resolved values, so it stays a way for the two numbers to disagree for the other value-less forms the function documents as deliberate. Nothing here changes that.

### DIFF
--- a/internal/extractors/python/types.go
+++ b/internal/extractors/python/types.go
@@ -549,6 +549,12 @@ func pythonEnumLiteralValue(rhs ts.Node, src []byte) string {
 		return extractor.StripLiteralQuotes(nodeText(rhs, src))
 	case "true", "false", "none":
 		return nodeText(rhs, src)
+	case "parenthesized_expression":
+		// `NAME = (\n    "value"\n)`: the grammar wraps a lone parenthesised
+		// value distinctly from the comma-bearing forms below.
+		if el := rhs.NamedChild(0); el != nil {
+			return pythonEnumLiteralValue(el, src)
+		}
 	case "tuple", "expression_list":
 		// Django (Integer|Text)Choices: `ACTIVE = "active", "Active"` → first
 		// element is the stored value. The grammar exposes the bare comma form
--- a/internal/extractors/python/types_test.go
+++ b/internal/extractors/python/types_test.go
@@ -101,6 +101,36 @@ class Status(StrEnum):
 	}
 }
 
+// TestEnumValueSet_PythonParenthesizedValue asserts a member value wrapped in
+// parentheses across lines (`NAME = (\n    "value"\n)`) still resolves to its
+// literal, while auto() stays empty-valued and a Django-style tuple still
+// resolves to its first element.
+func TestEnumValueSet_PythonParenthesizedValue(t *testing.T) {
+	src := `
+import enum
+
+class TransactionType(enum.StrEnum):
+    SHORT = "short"
+    A_VERY_LONG_MEMBER_NAME = (
+        "some-value"
+    )
+    AUTOD = enum.auto()
+    DJANGO_LIKE = ("db", "Label")
+`
+	ents := extractPy(t, src, "app/transaction.py")
+	en := findEnum(ents, "TransactionType")
+	if en == nil {
+		t.Fatal("SCOPE.Enum:TransactionType value-set node not found")
+	}
+	want := "SHORT=short, A_VERY_LONG_MEMBER_NAME=some-value, DJANGO_LIKE=db"
+	if got := en.Properties["values"]; got != want {
+		t.Fatalf("values = %q, want %q", got, want)
+	}
+	if got := en.Properties["member_count"]; got != "4" {
+		t.Fatalf("member_count = %q, want 4", got)
+	}
+}
+
 // TestEnumValueSet_NonEnumNoNode asserts an ordinary (non-enum) class emits NO
 // SCOPE.Enum node — the negative case.
 func TestEnumValueSet_NonEnumNoNode(t *testing.T) {


cajasmota:
Merging — thank you. The fix is the right shape: recursing rather than re-implementing means the value is resolved by whichever case already handles it, so the string, numeric, keyword and Django-tuple forms behave identically with or without parentheses. That is a smaller change than it looks, and it is why it is safe.

I checked the things a recursion invites, and it holds up:

- **Termination is well-founded** — `NamedChild(0)` is a strict descendant. `((("x")))` resolves to `"x"`, and nesting at depth 100 / 1 000 / 5 000 / 20 000 all resolve cleanly with no stack growth problem.
- **`X = ()` never reaches the new case at all** — the grammar parses it as `tuple` with zero named children, not `parenthesized_expression`.
- **No over-firing in the shipped code.** Every computed form your issue names stays value-less through the new case: `(enum.auto())`, `(foo())`, `(a + b)`, `(CONST)`, `(lambda: 1)`, `(-1)`, `(yield 5)`, `("a" if b else "c")`. The recursion only ever hands control to a case that was already trusted, which is exactly the property that makes this safe.
- **`cmd/grafel`'s whole-graph digest did not move**, and it could not: `selftestdata/fixture/util.py` and every `.py` under `internal/quality/golden/` contain no Python enum and no parenthesised assignment. No baseline note needed.

## One thing I am adding in a follow-up rather than asking you to

Your PR body says the fixture "pins that the recursion does not give a value to a computed member and does not change which tuple element is stored". Both of those members turn out to route through **pre-existing** paths — `AUTOD = enum.auto()` is a bare `call`, and `DJANGO_LIKE = ("db", "Label")` parses as `tuple`, not `parenthesized_expression`. So neither one actually enters the new case, and deleting the new case entirely leaves both assertions passing.

That leaves the new code with one graded axis and no negative control, which I confirmed by measurement: replacing the recursion with the parenthesised node's raw text — `stripQuotes(nodeText(el, src))` — keeps the whole package **green** while producing `(enum.auto())` → `"enum.auto()"`, `(foo())` → `"foo()"`, `(a + b)` → `"a + b"`. Fabricated values for exactly the computed forms your issue was careful to exclude.

Adding one fixture line closes it:

```python
AUTOD_PAREN = (enum.auto())
```

With that line present, the shipped code stays **green** (it correctly gives the member no value) and the raw-text mutant goes **red**: `values = "…, AUTOD_PAREN=enum.auto(), …"`. `member_count` moves 4→5 with it.

This is a gap in the *test*, not in your fix — the production change is correct as written, which is why it is going in as-is. I will push the control plus a parenthesised-tuple case in a follow-up.

## Two things you found the edge of, filed separately

Both are **pre-existing** and neither is caused by this change, but your fix extends their reach, so they are worth their own issues rather than scope creep here:

- **f-strings fabricate a value.** `X = f"hi{n}"` already resolves to the literal text `f"hi{n}"` through the existing `string` case, because `stripQuotes` sees a leading `f` rather than a quote. By the function's own rule an f-string is computed and should be value-less.
- **A leading comment defeats the fix in the exact case it targets.** `X = (  # why\n    "v"\n)` — which is what a formatter emits when a comment forces the wrap — has `comment` as `NamedChild(0)`, so it resolves to `""`. Same shape as #6796, one step further out.

Your `member_count` caveat is accurate and I have left it exactly as you scoped it: the test asserts `member_count` against a `values` set of different size, so it pins the disagreement rather than papering over it.

Issue #6831, opened 2026-09-05
Examine the three data-dependent t.Skip sites in internal/graph — the original scope (internal/types, internal/entkinds) has zero members and the two known instances are fixed


cajasmota:
`TestSCOPEPrefixedKindsOutsideTheEnumAreStillCounted` has been skipping ENTIRELY since arm B4 — the skip drops the whole test instead of the one row it excuses

## The defect

The test carries a guard that skips when its `SCOPE.`-prefixed sample kind has become enum-valid. **#6776 arm B4 made `SCOPE.Process` valid**, so that condition has been true ever since — and because the skip is at **function** scope rather than around the single row it excuses, **the entire test has been a no-op for several arms.**

Its un-prefixed control — the half that does not depend on the prefixed sample at all — has not been reached since B4.

## Why this is worse than a deleted test

A deleted test is visible in a diff and in coverage. A test that runs, prints `--- SKIP`, and reports success looks like coverage. Nothing in the suite output distinguishes "this property is guarded" from "this property has not been checked since August".

It also **decays silently by design**: the skip condition is *"has this kind become valid yet"*, and #6776 exists precisely to make more kinds valid. Every arm of that issue makes the skip more likely to fire, so the test was guaranteed to switch itself off as the work it guards progresses. That is the opposite of what a ratchet-adjacent test should do.

## Found while shipping arm B7

Arm B7 (`Operation, Route, Service, Config` → ledger 8→4) updated the control string but **deliberately did not touch the skip**, because changing skip semantics inside a migration batch would mix two unrelated risks. That was the right call for that PR and is why this is filed separately.

Note the same batch found a *second*, different instance of the same family: `TestEntityKindsAreCountedNotDropped`'s doc claimed two non-enum kinds while its body observed nothing of the sort — substituting a **valid** kind left it green. That one was fixed in B7 by constructing the document from the same values the guard asserts about. **Two tests in the same area, both reporting success while checking nothing.**

## What a fix should do

1. **Narrow the skip to the row it excuses**, not the function. The un-prefixed control has no dependency on the prefixed sample and should always run.
2. **Prefer a sample that cannot go valid** over a skip at all — e.g. a kind held on a ledger the test itself reads, so the sample tracks reality instead of going stale.
3. If a skip must remain, **make it fail loudly when it becomes permanent**. A skip that has been unconditionally true for months is not an excuse, it is a dead test; something should say so.

## Not established

[...]

Issue #6830, opened 2026-09-05
AllEntityKinds() is ungraded for a constant declared OUTSIDE kinds.go — the "passes every existing guard" premise was false (internal/cli catches the duplicate)


cajasmota:
`AllEntityKinds()` can contain duplicate entries undetected — the pin is set-based and counts source constants, not slice elements

## The gap

Listing the same constant twice in `AllEntityKinds()` passes every existing guard. Proved as an ALIVE mutant during the review of #6776 arm B6 (`EntityKindModelBare` listed twice → `internal/types`, `internal/entkinds`, `internal/graph`, `internal/graph/fbwriter` all green).

Why nothing catches it:

- `TestEntityKindDeclarations6776_MatchAllEntityKindsExactly` compares **sets**, so a duplicate collapses before comparison.
- Its `len(declared)` counts **constants parsed from the source file**, not elements of the returned slice — so the count check is measuring a different thing from the slice being validated.
- The vocabulary pin (#6779) is likewise set-based and sorts its own copy.

Every guard on this function therefore treats it as a set, while it is declared and returned as a **slice**.

## Why it is not purely cosmetic

`AllEntityKinds()` is a **public, slice-returning API**. Any caller that does `len(AllEntityKinds())`, iterates to emit one row per element, builds a histogram, or drives a table-test from it would be wrong in a way no test would reveal.

This is the same shape as the roster defect #6776 arm B5 fixed: "sorted" and "no duplicates" were prose nothing observed, because the pin compared sets and sorted its own copy. That one was closed with a strictly-increasing check whose two comparison branches were then shown to grade sortedness and uniqueness **independently**. `AllEntityKinds()` never received the equivalent treatment.

## The fix is small

A one-line duplicate check in the existing declarations test — build a `map[EntityKind]bool` while walking the slice and fail on a repeat. It should be scored with its own mutant (duplicate an entry, confirm it dies **alone**), because a check that only fires alongside the set comparison would be the masking pattern this batch has now hit eight times.

## Not established

- **Whether any current caller is actually sensitive to duplicates.** No sweep of `AllEntityKinds()` callers has been done. If some caller already de-duplicates, the practical blast radius is smaller than the API contract suggests — but the contract is still wrong.
- **Whether a duplicate has ever existed** in the list historically. Unknown; nothing would have recorded it.
- Whether the same set-vs-slice asymmetry affects other slice-returning vocabulary accessors in `internal/types`.

[...]

Issue #6829, opened 2026-09-05
fbwriter's copy of the rule-declared-kind ledger can shrink consistently and undetected — nothing couples it to the real ledger in internal/entkinds


cajasmota:
`fbwriter`'s copy of the rule-declared-kind ledger can shrink consistently and undetected — nothing couples it to the real ledger in `internal/entkinds`

## The gap

`internal/graph/fbwriter/non_enum_entity_kinds_6776_test.go:435` holds `ruleDeclaredKinds6776`, a **hand-transcribed copy** of the deferred rule-declared kind ledger that actually lives in `internal/entkinds`.

Its guards cover one direction only:
- a `len(...) != 8` premise check, and
- a per-subtest `types.IsValidEntityKind(kind)` fatal.

Together those catch **adding a migrated kind** — i.e. leaving a row behind after it has entered the enum.

They do **not** catch a **consistent shrink**. Removing a row *and* lowering the literal from 8 to 7 in the same edit leaves `internal/types`, `internal/entkinds` and `internal/graph/fbwriter` all green. Proved as an ALIVE mutant during the review of #6776 arm B6.

Both variables are **unexported, in different packages**, so nothing couples them. The transcription can drift from the ledger it claims to mirror, in the direction that quietly reduces coverage.

## Why it matters now specifically

The doc comment asserts the test covers **"ALL 8 ledger entries"**. That is prose asserting what no test observes — the repo's most common defect class, and one that has appeared in nine consecutive PRs.

**#6776 arm B6 re-asserted it.** That batch moved the ledger 12 → 8 and edited both numbers, so the "ALL 8" claim is freshly restated and still unobserved. Arms **B7** (`Operation, Route, Service, Config` → 4) and **B8** (`Constraint, Endpoint, Plugin, Template` → 0) will each edit both numbers again. Every one of those edits is an opportunity for the copy to silently stop mirroring, and the failure mode is a shrinking test that still reports success.

It is **not** a regression from B6 — the coupling has never existed.

## Two candidate fixes

1. **Export the entkinds ledger keys behind a test-only accessor** and have fbwriter derive its list rather than transcribe it. Removes the class entirely.
2. **Have fbwriter assert its transcription against `entkinds`** — keeping the local copy but failing when the two disagree in either direction.

(1) is stronger; (2) is smaller and preserves the local list's readability. Either is fine, but a fix that only re-checks the count is not — the count is exactly what a consistent shrink also updates.

## Not established

[...]

Issue #6821, opened 2026-09-05
Entity counts are non-deterministic across runs of an identical tree — SCOPE.Process and SCOPE.External both observed varying, so it is not flows-specific


cajasmota:
`SCOPE.Process` entity counts are non-deterministic across runs of an identical tree — observed twice, independently, and it now affects a test gate

## The observations

**Two independent measurements, months apart in the work, agreeing on the kind and not on much else.**

**#6776 arm A** — a runtime entity-kind counter over two corpora. Every one of the 25 ledger kinds reproduced exactly across two environments. `SCOPE.Process` did **not**: 58+82 in one environment, 93+60 in another. It was deterministic *per machine* across three runs each side, and the two corpora moved in **opposite directions** — which rules out a tree-state or version explanation, since either would push both the same way.

**#6776 arm B5** — while replacing a measurement control, `total_entities` came out **5461 / 5462 / 5463 across three in-process indexes of an identical tree**. The full entity-kind histogram was then compared between two commits: 51 distinct kinds, identical set, and every per-kind count identical **except `SCOPE.Process` (55 vs 56)**. An independent reviewer separately saw `total_entities` at 5453–5456 on *both* sides of the same comparison.

So the variance is real, it is isolated to this one kind, and it is present within a single machine and a single tree — not only across environments.

## Where it comes from

`SCOPE.Process` is `flows.kindProcess` — a **Go producer** (`internal/graph/flows/`), not a rule-pack declaration. It is a linearised call/data-flow chain. `internal/engine/process_flow.go:497` is the emit site the entity-kind sweep reports.

Not diagnosed further. A flow-linearisation pass producing a count that wobbles by one suggests iteration over a map, a concurrency-ordering dependency, or a cycle-breaking tie that resolves differently per run — but **that is a hypothesis, not a finding.** Nobody has read the pass with this question in mind.

## Why it matters now, beyond tidiness

**It is about to sit under a test gate.** #6488 arm D persists `entity_extracted_total` / `relationship_extracted_total` into the ratchet so unnamed over-emission becomes visible. A total that varies run-to-run is exactly what that gate compares.

[...]

PR #6833, opened 2026-09-05
feat(#6776): arm B7 — Operation, Route, Service and Config into the enum; Route pays BOTH ratchets

cajasmota:
feat(#6776): arm B7 — Operation, Route, Service and Config into the enum; Route pays BOTH ratchets

`Operation`, `Route`, `Service` and `Config` join `types.AllEntityKinds()` in
their CURRENT un-prefixed spellings. No rename, no `SCOPE.` prefix, no
stored-graph migration, `KindVocabularyVersion` still 1 (its guard bumps only
when `len(removed) > 0`, and this change removes nothing).

`ruleDeclaredKindsDeferredMax` drops 8 -> 4, and `Route` is on the SECOND
ledger as well, so `goUnprefixedKindsDeferredMax` drops 6 -> 5 in the same
commit. A mutant that lowers only the rule ratchet — the exact slip B5 made
with `Migration` — is DEAD on
`TestProducerEntityKinds6776_UnprefixedLedgerIsExact`.

Each pair was read at its cited site before being relied on, not relayed:

  Operation  internal/resolve/refs.go:1940 is literally
             `k == "SCOPE.Operation" || k == "Operation"`.
  Route      internal/enrichment/candidates.go:98 (and pricing.go:38) list
             "Route" beside "SCOPE.Route" in ONE case arm.
  Service    classfold.go:44/62 pairs "Service":100 with "SCOPE.Service":100
             and :91 gives both canon-rank 4. Separately, a docker compose
             service is extracted twice over: bare `Service` from
             rules/docker/frameworks/docker_compose.yaml:39 and
             "SCOPE.Service" from patterns/docker_compose_extractor.go:68.
  Config     both spellings name a configuration file — rule YAML emits bare
             `Config` (ansible_core.yaml:31,34; docker_compose.yaml:30,33),
             patterns/config_detector.go:49 emits "SCOPE.Config" subtype
             "config_file".

ONE CORRECTION to the grounding that selected this batch. It carried a caveat
that the `Config` twin uses a SYNTHETIC SourceFile for id collapse.
config_detector.go passes the REAL filePath to makeEntity, so the caveat does
not describe that producer; and `IsValidEntityKind` is membership in a string
set that never reads SourceFile, so an id-shaping difference could not bear on
enum membership either way. Recorded in the const block rather than dropped.

#6818 clause (a) covers all four and NO allow-list entry was added:
`enumOnlyEntityKinds` still holds exactly `SCOPE.ScopeUnknown` and
`enumOnlyEntityKindsMax` is still 1. Fabricating any one of the four constants'
values (`ConfigZZ`, `OperationZZ`, `RouteZZ`, `ServiceZZ`) is DEAD on
`TestEveryEnumEntityKind6818_CorrespondsToSomethingOutsideTheEnum`, per member,
individually.

FIXTURE CHURN. `Route` and `Config` were `internal/graph/fbwriter`'s negative
controls and its `:67` guard hard-failed "fixture is inert", as designed. Both
were re-picked from the four kinds still on the ledger — `Route` -> `Endpoint`,
`Config` -> `Plugin`, and the one remaining `Operation` slot -> `Template`.
`internal/types`' `TestEntityKindEnum6776_B4_KindsOutsideTheEnumStayOutside`
mirrors that fbwriter fixture in two rows and was moved in the SAME commit,
because the mirroring is what makes those rows a documented dependency rather
than two arbitrary strings.

A defect found by mutating rather than by reading:
`TestEntityKindsAreCountedNotDropped`'s doc comment claimed "one enum kind plus
two kinds the enum does NOT hold", and the body observed no such thing —
swapping "Template" for the valid "Model" left the test GREEN. It now derives
the document from the same `enumKind`/`nonEnumKinds` values it asserts about; a
guard that reads its own literals leaves the fixture free to ignore it. That
mutant is now DEAD.

Measured on a full in-process index of `testdata/fixtures` with
`qualityIndexOptions()` (custom extractors on), isolated
GRAFEL_HOME/GRAFEL_DAEMON_ROOT — the same method arm B6 used:
`entity_entities_kind_not_in_enum` 409 -> 97 (-312), distinct 17 -> 13. The
drop is exactly the four populations: Config 121, Operation 122, Route 52,
Service 17. Control is the full per-kind histogram, not `total_entities`
(#6821): 65 distinct kinds before and after, identical set, every count
identical except `SCOPE.External` 539 -> 537, one of the two kinds already
known to vary across runs on identical input.

No golden fixture or `baseline.json` figure moved; `minFixtures` is still 26
and the three exact-count constants are still 31. `go run ./tools/coverage
validate` reports the same 822 records / 17532 warnings before and after.

TWO KNOWN GAPS CONFIRMED STILL OPEN, neither made worse and neither fixed here:
#6829 — shrinking fbwriter's transcribed ledger copy from 4 to 3 with its own
length check is ALIVE, both variables being unexported and in different
packages. #6830 is untouched. Separately surfaced:
`TestSCOPEPrefixedKindsOutsideTheEnumAreStillCounted` has been SKIPPING
entirely since arm B4 made `SCOPE.Process` valid — its skip drops the whole
test rather than the one row it excuses, so its un-prefixed control was never
reached. The control string was updated with the rest; the skip is not touched
here.

Refs #6776

## Mutation table

|---|---|---|---|
| M1 | `ruleDeclaredKindsDeferredMax` 4 → 5 | DEAD | `TestRuleDeclaredKindsRatchetOnlyShrinks` |
| M2 | `ruleDeclaredKindsDeferredMax` 4 → 3 | DEAD | `TestRuleDeclaredKindsRatchetOnlyShrinks` |
| M3 | `goUnprefixedKindsDeferredMax` 5 → 6 | DEAD | `TestProducerEntityKinds6776_UnprefixedLedgerIsExact` |
| M4 | `goUnprefixedKindsDeferredMax` 5 → 4 | DEAD | same |
| M5 | **the `Route` trap** — rule ratchet lowered, `goUnprefixedKindsDeferred` left holding `Route` at max 6 | DEAD | same |
| M6 | `pinnedEntityKindVocabulary`: `Config` misordered, set intact | DEAD | `TestEntityKindVocabularyIsStrictlyIncreasing` |
| M7 | `pinnedEntityKindVocabulary`: `Config` deleted, order intact | DEAD | `TestEntityKindVocabularyIsPinnedToItsVersion` |
| M8a–d | each new constant's VALUE fabricated (`ConfigZZ`, `OperationZZ`, `RouteZZ`, `ServiceZZ`) | DEAD ×4 | `TestEveryEnumEntityKind6818_CorrespondsToSomethingOutsideTheEnum` (+ `TestNoUndeclaredRuleEntityKinds`) |
| M9 | fbwriter inert-guard controls reverted to `Route`/`Config` | DEAD | `TestStreamingWriterTalliesOnlyEntityKindsAbsentFromTheEnum` |
| M12 | fbwriter flat-path control → the valid kind `Component` | DEAD | `TestWriteGraphGenReportWiresTheFlatEntityProducerPath` |
| M13 | counted-not-dropped slot `Template` → the valid kind `Model` | ALIVE before the fix, **DEAD after** | `TestEntityKindsAreCountedNotDropped` |
| M14 | delete the guard added for M13, fixture unmutated | **equivalent** | a guard that only fires on a bad fixture is unobservable on a good one; M13 is its positive control |
| M10 | fbwriter's transcribed ledger copy shrinks 4 → 3 with its own length check | **ALIVE** | pre-existing **#6829**, deliberately not fixed here |

M6 and M7 are killed by *different* tests, so the sortedness pin and the set pin
are graded separately — that is the exact pair B5 found mutually masked.

Every mutant edit asserted an exact match count and aborted otherwise; one did
abort at MATCH COUNT 2 (a literal shared by the new guard and the document),
which is how the M13 coupling defect was found. Every scoring run executed the
whole package with `-count=1` after `go vet`, and the `=== RUN` count is
recorded per run rather than trusting a `-run` filter. Restores were by `cp`
from a `shasum`-verified backup, never `git checkout`.

---

## Review round (`c63ae1735`) — D1, D2, D3

Test-integrity only; no production code, no ratchet and no measurement moves.

|---|---|---|---|
| N1 | `TestSCOPEPrefixedKindsOutsideTheEnumAreStillCounted`: un-prefixed control → the valid kind `Model` | ALIVE | **DEAD** |
| N1b | every prefixed row dropped, i.e. `len(prefixed) == 0` | — | **DEAD** |
| N2 | `TestEntityKindsAreCountedNotDropped`: both non-enum slots → `Endpoint` | ALIVE | **DEAD** |
| N7 | `internal/types` row `Endpoint` → `Constraint` (mirror diverges) | ALIVE | ALIVE — by design; the comment now says convention, not constraint |

**D1** — the skip stopped the whole test rather than the one row it excused, so
arm B4's declaration of `SCOPE.Process` left it running zero assertions while
reporting SKIP. A declared prefixed row is now dropped from the population and
every assertion is sized off the surviving slice; `len(prefixed) == 0` is a hard
failure. The test PASSES rather than SKIPS. Filed as **#6831**, fixed here
because the PR was already editing the test.

**D2** — one added assertion rejecting `nonEnumKinds[0] == nonEnumKinds[1]`.

**D3** — softened, not guarded. Enforcing the mirror means one package reading
another's unexported fixture literals, which is exactly **#6829**'s shape and
out of scope for this arm; the comment now states the convention, why it is not
enforced, and what *is* enforced.

Seven packages green, `gofmt -l` clean, `go run ./tools/coverage validate`
unchanged at 822 records / 17532 warnings.


cajasmota:
## Coordinator verification — and the D3 deferral is now graded, not just argued

The review left **N7 ALIVE** (swapping the types-side non-enum row `Endpoint` -> `Constraint` is green in both packages) and the author declined to enforce the cross-package mirror, on the grounds that enforcing it means one package reading another's unexported fixture literals — #6829's shape, which the brief excluded.

That is an argument. Per our own rule that a deferred mutant kill is a claim until it is priced, I ran the mutant that grades the **premise** of the deferral rather than the mirror itself: put `Endpoint` — a B8 kind, deliberately not one of B7's four — into `AllEntityKinds()`.

**DEAD, across three packages and eleven tests**, and more importantly it fails *the right way*:

```
non_enum_entity_kinds_6776_test.go:69   fixture is inert: "Endpoint" is expected to be ABSENT FROM THE ENUM
non_enum_entity_kinds_6776_test.go:233  ... the flat path has nothing non-enum to report; pick one still on internal/entkinds' ledger
non_enum_entity_kinds_6776_test.go:337  ... this document no longer covers the droppable side; re-pick from internal/entkinds' ledger
non_enum_entity_kinds_6776_test.go:551  ... the un-prefixed control "Endpoint" ...; re-pick from internal/entkinds' ledger
```

Five distinct inert-guards fire, each naming what went stale **and where to re-pick from**. That is the enforcement. A textual mirror guard would have caught the same drift later and told the reader less. **The author's call is right, and it is right for a better reason than the one they gave:** the mirror does not need coupling because each side independently hard-fails the moment its own control goes valid. I have asked them to record that reason in place of the "convention" wording.

### This measurement also hands B8 its brief

B8 migrates `Constraint, Endpoint, Plugin, Template` — and `Endpoint` and `Plugin` are exactly the controls this fbwriter test uses. So **B8 will open on these eleven red tests.** That is not a trap it has to discover; it is the above output, in advance, with the fix instruction already written into each message.

One thing B8 must not do: re-pick from the rule ledger. B8 takes that ledger 4 -> 0, so after it there is no non-enum kind left there to borrow. The controls have to come from the kinds still outside the enum in the corpus — the residue behind `entity_entities_kind_not_in_enum`, which stays at ~97 after B7 precisely because it is not rule-declared. Picking one deliberately, with its provenance recorded, is part of B8's work rather than an afterthought.

### Ledger coupling, checked rather than assumed

I also tried to break the "Route pays BOTH ratchets" claim by putting an already-valid kind (`Model`, valid since B6) back on the Go ledger with the pin raised to match. It cannot survive: `assertLedgerExact`'s second loop derives `found` from `invalidGoKindSites`, so a kind that becomes valid drops out of `found` and trips *"ledgered kind is no longer declared by any Go producer (or has become valid)"*. The two ledgers are coupled to `IsValidEntityKind` by construction, not by a count that happens to agree. Recording it because "the ratchet only counts" was a plausible reading of the diff and it is wrong.

Mutant hygiene: byte-level backup, restored by `cp`, shasum equal before and after (`898c9688...`), worktree clean at `c63ae1735`.

Issue #6823, opened 2026-09-05
internal/links has ELEVEN unhardened source-path reads, not one — #6416's FIFO hang was fixed at one of twelve sites


cajasmota:
`internal/links` still reads source files with plain `os.ReadFile` — the #6416 FIFO hang was closed on the engine twin only

## The asymmetry

#6416 ("a FIFO with an indexed extension blocks an indexing worker forever — it branches only on `IsDir()`, and a FIFO does not fail the read, it waits") was closed 2026-08-21. The fix reached the **engine** copy of the substrate resolver:

`internal/engine/http_endpoint_substrate_fold.go:290-294`
```go
// safeio, not os.ReadFile: this pass re-opens by path after the walk's
// …
content, err := safeio.ReadFile(filepath.Join(r.srcRoot, file), safeio.FollowSymlinks, substrateMaxFileBytes)
```
That file's header (`:51`, `:83`) spells out why — non-blocking open, FIFO-behind-symlink safe, and a size bound because *"a character device never reaches EOF."*

It did **not** reach the near-identical resolver in the link pass:

`internal/links/constant_propagation.go:355`
```go
content, err := os.ReadFile(abs)
```

`safeio` appears nowhere in that file. The two resolvers were independently confirmed to be algorithmically identical (same depth, same lookup keys, both repo-scoped) — they are twins, and only one was hardened.

## Why the closed issue does not cover it

#6416's fix is correct and complete *for the path it touched*. The link pass runs in `group-link`, over graphs loaded from disk, and re-opens source files by path exactly as the engine pass does — so it has the same exposure to the same construct, reached through a different entry point.

A FIFO with an indexed extension in a linked repo therefore still blocks a `group-link` worker indefinitely. The engine's own comment explains precisely why `os.ReadFile` is unsafe here; that reasoning applies unchanged to the link copy.

## Found while grounding #6450, and deliberately not bundled there

#6450 Task 2 proposes **deleting** the link-pass fold as dead work — the engine sets `url_kind="literal"` at index time and `applyResolverToConsumerHTTP` skips anything not `dynamic_baseurl`, so the link copy folds nothing in the normal case.

**Deleting the fold would not close this**, and the two should not be conflated:
- `buildResolver` is called for its own reasons and would survive a fold deletion; and
- even if a deletion did remove the only live caller today, leaving an unhardened `os.ReadFile` in the tree re-opens the hole the moment anything calls it again.

[...]

PR #6838, opened 2026-09-05
fix(#6823): the link pass never got #6416's FIFO fix — eleven source reads in internal/links could still park a group-link worker in open(2)

cajasmota:
# fix(#6823): the link pass never got #6416's FIFO fix — and it was not a pair, it was eleven copies

`Refs #6823`

## The asymmetry was still present, exactly as described

Verified on `main` @ `b671aa7a2` before touching anything:

- **Hardened:** `internal/engine/http_endpoint_substrate_fold.go:302` reads through `safeio.ReadFile(..., safeio.FollowSymlinks, substrateMaxFileBytes)`, with the rationale at `:51`/`:83`.
- **Not hardened:** `internal/links/constant_propagation.go:355` was plain `os.ReadFile(abs)`. `safeio` appeared **nowhere** in `internal/links`.

Two line numbers in the issue had drifted (engine `:294` → `:302`); the substance was correct.

## What the sweep found — the issue undercounts by ten

The issue asked whether any *other* consumer in `internal/links` reads a source path with `os.ReadFile`. Eleven do, across eleven passes, and eight of them are the byte-identical block:

```go
abs := filepath.Join(srcRoot, file)
content, err := os.ReadFile(abs)
```

| Site | Shape | Action |
|---|---|---|
| `constant_propagation.go:355` | substrate sniff | **hardened** |
| `dataflow_pass.go:117`, `def_use_pass.go:108`, `effect_propagation.go:115`, `payload_drift.go:348`, `reachability.go:233`, `taint_flow.go:122`, `template_pattern_pass.go:78` | same block, verbatim | **hardened** |
| `complexity_pass.go:68`, `dataflow_pass.go:328` | `os.ReadFile(filepath.Join(srcRoot, rel))` body cache | **hardened** |
| `string_pass.go:295` | `scanFile` — stat, `IsDir()` branch, then read. The literal #6416 shape. | **hardened + skip arm** |
| `string_pass.go:285` (scan cache), `links.go:644` (`ReadLinkPassStats`), `phantom_edges.go:187` (`LoadLinksDocument`), `payload_drift.go:435/810/851` (links + drift sidecars), `candidates.go:18` (`readDoc`) | grafel's **own artefacts**, under a grafel-owned dir | **deliberately not hardened** — not plantable by whoever owns the indexed repo, and their callers depend on distinguishing `fs.ErrNotExist` from a real failure. Hardening them would trade a non-exposure for an error-semantics change. |

All eleven now go through one helper (`internal/links/safe_source_read.go`), not eleven hardened call sites — eleven copies is how the divergence happened in the first place.

## The repro, measured on unmodified behaviour

`internal/links/safe_source_read_6823_test.go` (`//go:build !windows`; every FIFO inside `t.TempDir()`).

**A FIFO test does NOT fail the suite on the unhardened code — it wedges it.** So each blocking-shaped body runs on its own goroutine under a 10s deadline (`runWithin`), which converts "never returns" into a named failure. A blocked goroutine is abandoned, not cancelled — there is no portable way to interrupt a thread parked in `open(2)`, which is precisely why `safeio`'s guard is a stat gate and not a timeout.

Measured against main's read (helper body = `os.ReadFile`, call sites untouched):

```
--- FAIL: TestReadSourceFile_FIFORefusedWithoutBlocking            (10.01s)
--- FAIL: TestReadSourceFile_SymlinkToFIFORefusedWithoutBlocking   (10.02s)
--- PASS: TestReadSourceFile_RegularFileAndSymlinkToRegularFile...  (0.01s)   <- positive control
--- FAIL: TestReadSourceFile_StopsAtTheByteBound                    (0.00s)
--- FAIL: TestBuildResolver_FIFOInSourceTreeDoesNotBlock           (10.00s)   <- constant_propagation call site
--- FAIL: TestScanRepo_FIFOIsSkippedAndDoesNotAbortTheRepo         (10.00s)   <- string_pass call site
--- FAIL: TestNoUnhardenedSourcePathReadsInPackage                  (0.01s)   <- listed all 11 sites
```

`buildResolver` and `scanRepo` **do not return at all**. The exposure is no longer inferred.

## `string_pass` needed more than a swap

A read error out of `scanFile` propagates through `scanRepo` and zeroes the repo's **entire** extraction set (the #5523 shape). Swapping the read alone would have converted the #6416 hang into a whole-repo blackout on the same input. `ErrNotRegular` is therefore a per-file skip; `ErrWouldBlock` still propagates, per `safeio`'s own guidance that mapping it to "nothing found" turns a bounded degradation into a silent one.

## The bound, and what it costs

`maxSourceFileBytes = 1 << 20` — the engine's value, chosen deliberately.

#6823 exists because the twins diverged once. A second axis of divergence — one bounded at 1 MiB, the other at something else — would be a new instance of the bug, not a fix for it. So the cost is inherited whole and stated in the code: `safeio.ReadFile` **truncates** at the bound (`io.LimitReader`), so a constant declared past the first megabyte is silently not found, with no telemetry. #6450 recorded exactly that as a permanent capability loss on the engine path; this path now shares it. `TestBuildResolver_OneMiBBoundIsObservable` makes it an observed result rather than a claim in a comment.

`stringScanMaxFileBytes = 4 MiB + 1` is the one deliberate divergence: that pass already discards files over 4 MiB after reading them whole, so this bound is one byte past its own threshold — over-length files still arrive over-length and are still discarded. Pure memory saving, no behavioural change.

## Mutation testing

`go vet` clean before every score; whole package each time (`-count=1`, exit code, no `-run` filter); every edit asserted to land at exactly one occurrence; restore by `cp` with `shasum` verified both sides and a clean `git status` assertion.

|---|---|---|---|
| M1 | helper `FollowSymlinks` → `RejectSymlinks` | **DEAD** | `..._RegularFileAndSymlinkToRegularFileStillRead` |
| M2 | helper falls back to `os.ReadFile` | **DEAD** | 7 tests, incl. 3 deadline failures |
| M3 | `constant_propagation` call site bound → `0` (unbounded) | **DEAD** | `TestBuildResolver_OneMiBBoundIsObservable` |
| M4 | `ErrNotRegular` skip arm → `return nil, err` | **DEAD** | `TestScanRepo_FIFOIsSkippedAndDoesNotAbortTheRepo` |
| M5 | `stringScanMaxFileBytes` → `100` | **DEAD** (was ALIVE) | `TestScanRepo_ReadsPastTheFirstBytesOfAFile` |
| M6 | one call site (`taint_flow.go`) reverted to `os.ReadFile` | **DEAD** | `TestNoUnhardenedSourcePathReadsInPackage` |

### Round 2 — permissive direction (review found six survivors)

Round 1's mutants all asked "did the fix stop working". None asked "does it fire too broadly, or at the wrong boundary". Six permissive mutants survived that first suite; all six were reproduced here (vet clean, whole package, 387 `=== RUN`) before being fixed, and all six are now dead (390 `=== RUN`).

| # | Permissive mutant | Before | After | Killed by |
|---|---|---|---|---|
| D1 | twelfth copy as `os.ReadFile(filepath.Join(g.FileRoot, file))` — no `srcRoot` local | ALIVE | **DEAD** | guard now matches case-insensitively and knows `FileRoot` |
| D1b | twelfth copy via `os.OpenFile` | ALIVE | **DEAD** | guard now matches `os.OpenFile(` (which `"os.Open("` does not substring-match) |
| D2 | `string_pass` swallows **every** read error, not just `ErrNotRegular` | ALIVE | **DEAD** | `TestScanRepo_NonSkippableReadErrorStillPropagates` |
| D3 | `stringScanMaxFileBytes` `4 MiB + 1` → `4 MiB` | ALIVE | **DEAD** | `TestScanFile_DiscardsAFileOverFourMiB` |
| D3b | delete the `len(body) > 4*1024*1024` discard | ALIVE | **DEAD** | same test — the bound and the discard are a pair, and it observes both |
| D4 | `maxSourceFileBytes` `1<<20` → `1<<24` | ALIVE | **DEAD** | `TestSourceReadBoundMatchesTheEngineTwin` |

D1 is the one that mattered most: the guard exists so this class cannot recur, and skipping the `srcRoot :=` local — the *more* natural way to write the next copy — walked straight past it.

D3's paragraph in round 1 said the `+ 1` was behaviour-preserving. True at the chosen value, false one byte away: at `4 MiB` exactly, a 4 MiB + 1 file is truncated to 4 MiB, fails the discard, and is **scanned** instead of discarded. That is now observed rather than asserted.

D4 killed the round-1 caveat too: `TestSourceReadBoundMatchesTheEngineTwin` reads `../engine/http_endpoint_substrate_fold.go` as text (the constant is unexported) and fails if either twin's byte bound moves. It is the one check here that sees across the twin boundary — it covers the **bound only**, not the engine's lazy sniff or its per-run read cap.

M5 was **ALIVE on the first pass** — every string-pass fixture's interesting literal sat inside the first 100 bytes, so a bound that truncated ordinary files changed nothing observable. A test was added rather than the mutant excused.

**Mutually-masking guards:** the type gate and the byte bound are orthogonal and each is exercised where the other cannot fire — a FIFO has zero size, so the bound never rejects it; the 1 MiB fixture is a regular file, so the type gate never rejects it. Neither masks the other.

## What the anti-divergence guard does and does not catch

`TestNoUnhardenedSourcePathReadsInPackage` scans this package's non-test `.go` files for `os.ReadFile`/`os.Open` on an argument mentioning `abs`/`srcRoot`/`fileRoot` (comment lines excluded).

- **Catches:** a twelfth copy pasted from any of the eleven — the exact regression shape.
- **Does NOT catch:** divergence from the **engine** twin. It cannot see `internal/engine`, so if the two drift again in symlink policy, byte bound, or read cap, it stays green. Collapsing the duplication is the only thing that closes that, and it is out of scope here (#6450 owns the resolver's future).
- **Does NOT run on Windows** (the file is `//go:build !windows`, since `syscall.Mkfifo` is Unix-only).

### Round 3 — the guard was itself vacuous

`os.ReadDir(".")` → `os.ReadDir(t.TempDir())` **passed**. The guard reported a clean package having scanned zero files: no entries, no loop bodies, no offenders, PASS. That is the failure `52aaa84f1` fixed in the secrets scanner ("reported a repo it never read as clean") and the one `internal/types` guards with `scanGoEntityKinds`' `FilesParsed` floor. A guard is the thing that is supposed to notice, so a vacuous one is worse than none — it occupies the slot and reports success.

The guard now counts what it scanned and fails below a floor of 15 (26 non-test files today, measured), with a message that says **the walk is broken, not that the package is dirty** — opposite diagnoses.

A floor alone was **not** enough, which the second mutant proved: this package has 35 test files against 26 non-test ones, so inverting the `.go`/`_test.go` filter clears a count-based floor comfortably while never opening a single production source. The guard therefore also anchors on three files that must be in the scanned set — `constant_propagation.go`, `string_pass.go`, `safe_source_read.go`.

| # | Mutant | Before | After | Killed by |
|---|---|---|---|---|
| V1 | `os.ReadDir(".")` → `os.ReadDir(t.TempDir())` | ALIVE | **DEAD** | scanned-file floor (`scanned 0 … want at least 15`) |
| V2 | `.go`/`_test.go` filter inverted — reads only test files | ALIVE | **DEAD** | must-scan anchor (`constant_propagation.go was never scanned`, 35 files walked) |

`break` after the first matching call token: **reachable, so fixed rather than excused.** `os.Open(abs); os.ReadFile(k)` on one line was checked against `os.ReadFile` first, so the examined remainder began *after* it and `abs` was never in view. Two opens on one line is unidiomatic but legal Go, and this guard's value is not needing the next copy to be idiomatic. Every call token is now examined; a `flagged` bool keeps one offender per line. D1 and D1b were re-scored after the restructure and both still die.

### Round 4 — the guard's matcher was ungraded

`roots := []string{"abs","srcroot","fileroot"}` → `[]string{"zzz_no_such_token"}` **passed**. The guard scanned all 26 files, cleared the floor, hit every anchor — and could not detect anything. Three rounds had proved:

1. did it read **anything**? → the file-count floor (round 2)
2. did it read the **right** things? → the must-scan anchors (round 3)
3. could it **detect** anything? → **nothing graded this**

Layers 1 and 2 both grade the walk. Neuter `roots` or `calls` and the guard becomes a very thorough no-op that reports success — the exact failure mode it exists to prevent, reachable through the one part of it never checked.

**How the positive control exercises the real loop.** The matcher was **not** extracted and unit-tested on its own — that leaves the loop free to ignore it, which this repo has shipped before. Instead the scan was parameterised over its *input*: `scanUnhardenedReads([]guardFile) []string` holds the entire loop **and** the entire matcher, and the walk above it now contributes nothing but a file list — it has no matching logic left to drift out of step. The real guard calls it with the package directory; the control calls the same function with one synthetic file. There is one loop, and both callers go through it.

The control asserts an **exact** offender set, line numbers included, over a fixture that mixes four offending lines with three safe ones that are safe for three different reasons (a comment quoting the pattern verbatim, a real `os.ReadFile` naming no source root, and the hardened helper itself). That grades over-firing as well as recall: a matcher that flagged every line fails it just as an empty one does.

|---|---|---|---|
| R1 | `roots` → a token matching nothing | **DEAD** | `TestScanUnhardenedReads_DetectsAndDiscriminates` |
| R2 | `calls` → a token matching nothing | **DEAD** | same |
| R3 | comment-skip widened to skip every line | **DEAD** | same |
| R4 | `strings.ToLower` dropped from the argument (D1's fix) | **DEAD** | same |
| V2 | filter inverted (re-score after restructure) | **DEAD** | must-scan anchor |
| D1 | twelfth copy via `g.FileRoot` (re-score) | **DEAD** | the guard |
| D1b | twelfth copy via `os.OpenFile` (re-score) | **DEAD** | the guard |

R1–R4 leave the real guard test **green** — it has nothing to find either way — and die only on the control. That is the shape of the hole: the guard cannot grade its own matcher, so something else has to.

### Round 5 — the guard's reaction was ungraded

`if offenders := scanUnhardenedReads(files); false && len(offenders) > 0` **passed**. The production guard could discard the matcher's entire output and the package stayed green. Four layers had been proved and the fourth was opened by the refactor that fixed the third — extracting a predicate leaves the call site free to ignore it.

The guard's four moving parts, each ungraded in turn and each now covered:

| Layer | Question | Graded by |
|---|---|---|
| 1 | did the walk read **anything**? | file-count floor (≥15; 26 today) |
| 2 | did it read the **right** things? | must-scan anchors |
| 3 | can the matcher **detect and discriminate**? | `TestScanUnhardenedReads_DetectsAndDiscriminates` |
| 4 | does the guard **act** on what the matcher returns? | `TestCheckNoUnhardenedReads_ActsOnWhatTheMatcherReturns` |

`checkNoUnhardenedReads(tb testing.TB, files []guardFile)` holds the decision and the `Fatalf`. The real test passes `t`; the control passes a `recordingTB` that records a failure instead of aborting, and asserts **both** directions — a dirty file set must fail (naming the offending line), a clean one must not. Nothing follows the `Fatalf`, so the function is correct whether or not `Fatalf` is terminal for the TB it was handed; `runtime.Goexit` is never called and no goroutine gymnastics are needed.

|---|---|---|---|
| C1 | call site discards matcher output (`false &&`) | **DEAD** | layer-4 control |
| C2 | reaction inverted (`>= 0`) — fails on a clean set | **DEAD** | layer-4 control **and** the real guard |
| R1 | `roots` neutered (re-score) | **DEAD** | layers 3 **and** 4 |
| R2 | `calls` neutered (re-score) | **DEAD** | layers 3 and 4 |
| R3 | comment-skip skips every line (re-score) | **DEAD** | layers 3 and 4 |
| R4 | `ToLower` dropped (re-score) | **DEAD** | layer 3 |
| V2 | filter inverted (re-score) | **DEAD** | layer 2 |
| D1 | twelfth copy via `g.FileRoot` (re-score) | **DEAD** | the guard |
| D1b | twelfth copy via `os.OpenFile` (re-score) | **DEAD** | the guard |

The restructure strengthened rather than unhooked: R1–R3 now die on **two** layers, because a neutered matcher also stops the reaction control's dirty set from failing.

### Round 6 — the anchors verified identity, not content

`guardFile{name: n, body: string(src)}` → `body: ""` **passed**. The walk handed the matcher the right file *names* with empty bodies: the floor saw 26, the anchors saw all three required files, both controls passed on their synthetic inputs, and the production guard scanned nothing.

The guard is a three-stage pipeline, and every stage is now observed:

| stage | layer | graded by |
|---|---|---|
| input | 1 read anything | file-count floor (≥15; 26 today) |
| input | 2 read the right files | must-scan anchors |
| input | 3 read their actual content | required-token check on the collected bodies |
| detect | 4 matcher detects **and** discriminates | `TestScanUnhardenedReads_DetectsAndDiscriminates` |
| react | 5 call site acts on the result | `TestCheckNoUnhardenedReads_ActsOnWhatTheMatcherReturns` |

The three input modes are **one** check with one `t.Fatalf` — they answer one question ("is the input real?") and three sites read as three unrelated rules. Every mode says *the walk is broken*, not *the package is dirty*.

Anchor tokens are chosen against volatility: `package links` is the package clause, present in every file here and moved by no refactoring short of renaming the package; `readSourceFile` is the helper the guard exists to push callers towards, so a rename breaking it would already require rewriting the guard's message.

**A defect the scoring caught in the fix itself.** The first cut checked a parallel `bodies` map built alongside `files` — and both mutants stayed **ALIVE**, because the thing verified was not the thing used. The check now interrogates the `files` slice handed to the matcher. The same-shaped trap as extracting a predicate and leaving the call site free to ignore it, one level down, and only visible because the mutant was scored rather than assumed.

|---|---|---|---|
| B1 | `body: ""` | **DEAD** (ALIVE against the parallel-map cut) | layer 3 |
| B2 | `body: n` — the name as the body | **DEAD** (ALIVE against the parallel-map cut) | layer 3 |
| V1 | `os.ReadDir(t.TempDir())` (re-score) | **DEAD** | layer 1 |
| V2 | filter inverted (re-score) | **DEAD** | layer 2 |
| C1 | call site discards matcher output (re-score) | **DEAD** | layer 5 |
| C2 | reaction inverted (re-score) | **DEAD** | layer 5 **and** the real guard |
| R1 | `roots` neutered (re-score) | **DEAD** | layers 4 and 5 |
| R2 | `calls` neutered (re-score) | **DEAD** | layers 4 and 5 |
| R3 | comment-skip skips every line (re-score) | **DEAD** | layers 4 and 5 |
| R4 | `ToLower` dropped (re-score) | **DEAD** | layer 4 |
| D1 | twelfth copy via `g.FileRoot` (re-score) | **DEAD** | the guard |
| D1b | twelfth copy via `os.OpenFile` (re-score) | **DEAD** | the guard |

### Round 7 — presence is weaker than completeness

A walk truncating every body to 4000 bytes **passed** the token check: `package links` is line one, `readSourceFile` appears early in `safe_source_read.go` — while the real read in `constant_propagation.go` sits at **byte 12961**. A truncating walk cleared every layer and could not see a single one of the eleven sites this guard protects.

Layer 3 now asserts the collected body's **length equals the file's size on disk** (`os.Stat`). That is exact, unsatisfiable by any prefix, and depends on no string surviving a refactor. The tokens are kept as a cheaper second opinion, not as the load-bearing check.

Walk integrity is still **one** `t.Fatalf` with four failure modes, each with its own diagnosis, all opening *the walk is broken, not the package is dirty*:

- `it scanned only 0 non-test .go files, want at least 15`
- `constant_propagation.go was never scanned`
- `the walk collected 4000 bytes for constant_propagation.go but the file is 16015 — it read a PREFIX`
- `... the body collected for it does not contain "package links"`

|---|---|---|---|
| T1 | bodies truncated to 4000 bytes | **DEAD** | layer 3 length |
| T2 | `body[:len(body)-1]` — one byte short | **DEAD** | layer 3 length (a token check never would) |
| B1 | `body: ""` (re-score) | **DEAD** | layer 3 |
| B2 | `body: n` (re-score) | **DEAD** | layer 3 |
| V1 | empty dir (re-score) | **DEAD** | layer 1 |
| V2 | filter inverted (re-score) | **DEAD** | layer 2 |
| C1 | `false &&` (re-score) | **DEAD** | layer 5 |
| C2 | reaction inverted (re-score) | **DEAD** | layer 5 and the real guard |
| R1 | `roots` neutered (re-score) | **DEAD** | layers 4 and 5 |
| R2 | `calls` neutered (re-score) | **DEAD** | layers 4 and 5 |
| R3 | comment-skip skips everything (re-score) | **DEAD** | layers 4 and 5 |
| R4 | `ToLower` dropped (re-score) | **DEAD** | layer 4 |
| D1 | twelfth copy via `g.FileRoot` (re-score) | **DEAD** | the guard |
| D1b | twelfth copy via `os.OpenFile` (re-score) | **DEAD** | the guard |

## Corrections made under review

**Blast radius was understated, and is now stated correctly in the code.** An error out of `scanFile` does not merely zero the repo's extraction set: it reaches `runStringPass`, and `links.go:388` returns it out of `RunAllPasses`, aborting **the entire group link run** — HTTP, drift, taint and reachability never run. Loud-fail is kept, with that price now named: `ErrWouldBlock` is reachable only from safeio's terminal state (64 abandoned opens), it was **not** reachable on this path before this change, and mapping it to "no findings" is the #6338 shape safeio's own doc forbids. A wrong graph is worse than an absent one.

**Cross-site policy inconsistency, recorded rather than papered over.** `string_pass` propagates; the other ten sites `continue` or cache `""` on **any** error, so they *do* map `ErrWouldBlock` to "nothing found". Those arms pre-date this change — what is new is that the error class is now reachable there. Aligning them means giving ten best-effort passes an abort path they have never had, which is a bigger behavioural decision than #6823 should make alone. The note lives at the one site that does propagate.

**The guard no longer needs a build tag.** It is pure static text scanning, so it moved to an untagged `source_read_guard_6823_test.go` and now runs on the Windows CI leg too, along with both bound tests. Only the `mkfifo` tests stay behind `//go:build !windows`.

**Equivalent under the current suite:** `string_pass.go`'s `fi.IsDir()` early return is now dead — a directory fails `safeio`'s type gate and returns `ErrNotRegular`, which the new skip arm handles identically. Deleting it changes no observable behaviour. Left in place as cheap defence-in-depth ahead of the read.

## Verification

- `go test -count=1 ./internal/links/` — ok
- `go test -count=1 ./internal/engine/` — ok
- `go test -count=1 ./cmd/grafel/` — ok (141s)
- `gofmt -l internal cmd` — 0 lines
- `go run ./tools/coverage validate` — ok, 822 records / 223 mapping records

## Out of scope, untouched

`internal/engine`'s copy; collapsing the duplication; `internal/types`, `internal/entkinds`, `internal/graph/fbwriter`. #6450's deletion of the link-pass fold is unaffected — `buildResolver` survives at `constant_propagation.go:236` for the RESOLVES_TO seeds, which is the path that still runs this read.

### DIFF
--- a/internal/links/complexity_pass.go
+++ b/internal/links/complexity_pass.go
@@ -24,7 +24,6 @@ package links
 
 import (
 	"fmt"
-	"os"
 	"path/filepath"
 	"sort"
 
@@ -65,7 +64,7 @@ func runComplexityPass(graphs []repoGraph, _ Paths) (PassResult, error) {
 				cache[rel] = ""
 				return ""
 			}
-			buf, err := os.ReadFile(filepath.Join(srcRoot, rel))
+			buf, err := readSourceFile(filepath.Join(srcRoot, rel), maxSourceFileBytes)
 			if err != nil {
 				cache[rel] = ""
 				return ""
--- a/internal/links/constant_propagation.go
+++ b/internal/links/constant_propagation.go
@@ -352,7 +352,7 @@ func buildResolver(graphs []repoGraph) *Resolver {
 				srcRoot = g.FileRoot
 			}
 			abs := filepath.Join(srcRoot, file)
-			content, err := os.ReadFile(abs)
+			content, err := readSourceFile(abs, maxSourceFileBytes)
 			if err != nil {
 				// File missing on disk (graph indexed from a prior
 				// snapshot, etc.) — skip; the pass is best-effort and
--- a/internal/links/dataflow_pass.go
+++ b/internal/links/dataflow_pass.go
@@ -114,7 +114,7 @@ func runDataFlowPass(graphs []repoGraph, paths Paths, rejects map[string]bool) (
 				srcRoot = g.FileRoot
 			}
 			abs := filepath.Join(srcRoot, file)
-			content, err := os.ReadFile(abs)
+			content, err := readSourceFile(abs, maxSourceFileBytes)
 			if err != nil {
 				continue
 			}
@@ -325,7 +325,7 @@ func resolveCrossFileFlows(g *repoGraph, handlerFile, srcRoot, lang string, xfil
 		if c, ok := cache[rel]; ok {
 			return c, c != ""
 		}
-		buf, err := os.ReadFile(filepath.Join(srcRoot, rel))
+		buf, err := readSourceFile(filepath.Join(srcRoot, rel), maxSourceFileBytes)
 		if err != nil {
 			cache[rel] = ""
 			return "", false
--- a/internal/links/def_use_pass.go
+++ b/internal/links/def_use_pass.go
@@ -105,7 +105,7 @@ func runDefUsePass(graphs []repoGraph, paths Paths) (PassResult, error) {
 				srcRoot = g.FileRoot
 			}
 			abs := filepath.Join(srcRoot, file)
-			content, err := os.ReadFile(abs)
+			content, err := readSourceFile(abs, maxSourceFileBytes)
 			if err != nil {
 				continue
 			}
--- a/internal/links/effect_propagation.go
+++ b/internal/links/effect_propagation.go
@@ -112,7 +112,7 @@ func runEffectPropagationPass(graphs []repoGraph, paths Paths, _ map[string]bool
 				srcRoot = g.FileRoot
 			}
 			abs := filepath.Join(srcRoot, file)
-			content, err := os.ReadFile(abs)
+			content, err := readSourceFile(abs, maxSourceFileBytes)
 			if err != nil {
 				continue
 			}
--- a/internal/links/payload_drift.go
+++ b/internal/links/payload_drift.go
@@ -345,7 +345,7 @@ func sniffPayloadShapes(graphs []repoGraph) (map[shapeKey]*shapeBucket, int) {
 				srcRoot = g.FileRoot
 			}
 			abs := filepath.Join(srcRoot, file)
-			content, err := os.ReadFile(abs)
+			content, err := readSourceFile(abs, maxSourceFileBytes)
 			if err != nil {
 				continue
 			}
--- a/internal/links/reachability.go
+++ b/internal/links/reachability.go
@@ -230,7 +230,7 @@ func runReachabilityPass(group string, graphs []repoGraph, paths Paths) (PassRes
 				srcRoot = g.FileRoot
 			}
 			abs := filepath.Join(srcRoot, file)
-			content, err := os.ReadFile(abs)
+			content, err := readSourceFile(abs, maxSourceFileBytes)
 			if err != nil {
 				continue
 			}
--- a/internal/links/safe_source_read.go
+++ b/internal/links/safe_source_read.go
@@ -1,0 +1,71 @@
+package links
+
+import (
+	"github.com/cajasmota/grafel/internal/safeio"
+)
+
+// safe_source_read.go — the link pass's single entry point for reading a
+// file out of a SCANNED SOURCE TREE (#6823).
+//
+// WHY THIS EXISTS. #6416: a FIFO with an indexed extension — `mkfifo
+// config.js`, one unprivileged command — parks whichever goroutine opens it
+// in open(2) forever. No timeout, no error, no log line. The fix for that
+// reached internal/engine/http_endpoint_substrate_fold.go, whose header
+// spells out the reasoning; it did not reach this package, which held ELEVEN
+// copies of the same `os.ReadFile(filepath.Join(srcRoot, rel))` read across
+// eleven passes. The link pass re-opens source files by path in group-link,
+// long after the indexing walk's irregular-file filter has run, so every one
+// of those copies had the same exposure through a different entry point.
+//
+// It is one helper rather than eleven hardened call sites because eleven
+// copies is how the divergence happened in the first place.
+//
+// WHAT IT DOES NOT COVER. Reads of grafel's OWN artefacts — links.json, the
+// resolves-to / drift sidecars, the string-scan cache, candidate documents —
+// deliberately stay on os.ReadFile. Those paths are constructed by grafel
+// inside a grafel-owned directory rather than taken from a scanned tree, so
+// they are not plantable by whoever owns the repo being indexed, and their
+// callers depend on distinguishing fs.ErrNotExist from a real failure.
+
+// maxSourceFileBytes caps a single source file read by the link pass.
+//
+// THE CHOICE, and its cost, stated rather than inherited silently. This is
+// the same 1 MiB the engine twin uses (substrateMaxFileBytes), and it is
+// deliberately the same value: #6823 exists because the two resolvers
+// diverged once already, and a second axis of divergence — one twin bounded
+// at 1 MiB, the other at something else — would be a new instance of the bug
+// being fixed, not a fix for it.
+//
+// The cost is real and is inherited whole. safeio.ReadFile TRUNCATES at the
+// bound (io.LimitReader) rather than failing, so a declaring file larger than
+// 1 MiB is parsed from its first megabyte only: a constant declared past that
+// point is silently not found, with no telemetry. #6450 recorded exactly that
+// as a permanent capability loss on the engine path. This path now shares it.
+// The trade is accepted because a bound is not optional — safeio needs one, a
+// character device never reaches EOF — and because base-URL constants,
+// import lines and handler bodies live in small modules; a source file over
+// 1 MiB is generated or minified, not hand-written.
+const maxSourceFileBytes int64 = 1 << 20
+
+// stringScanMaxFileBytes bounds the string-literal scan instead. That pass
+// already discards any file larger than 4 MiB after reading it whole, so the
+// bound is set one byte past that threshold: a file over the limit still
+// arrives over-length and is still discarded by the existing check, making
+// this a pure memory saving with no behavioural change. Every file the pass
+// accepts today is read whole today.
+const stringScanMaxFileBytes int64 = 4*1024*1024 + 1
+
+// readSourceFile reads abs, which must be a path inside a scanned source
+// tree, refusing anything that is not a regular file and reading at most
+// maxBytes.
+//
+// FollowSymlinks, matching the engine twin: the indexing walk mints file
+// entities for a symlink to a regular file, so rejecting symlinks outright
+// would delete legitimate coverage. The policy judges the TARGET, so a
+// symlink pointing at a FIFO is still refused.
+//
+// The bound is a parameter rather than a package default because the two
+// callers that need a different one should have to say so at the call site.
+func readSourceFile(abs string, maxBytes int64) ([]byte, error) {
+	return safeio.ReadFile(abs, safeio.FollowSymlinks, maxBytes)
+}
--- a/internal/links/safe_source_read_6823_test.go
+++ b/internal/links/safe_source_read_6823_test.go
@@ -1,0 +1,337 @@
+//go:build !windows
+
+package links
+
+import (
+	"errors"
+	"os"
+	"path/filepath"
+	"strings"
+	"syscall"
+	"testing"
+	"time"
+
+	"github.com/cajasmota/grafel/internal/safeio"
+)
+
+// safe_source_read_6823_test.go — #6823.
+//
+// #6416 hardened the ENGINE copy of the substrate resolver against a FIFO
+// planted in a scanned tree (internal/engine/http_endpoint_substrate_fold.go
+// reads through internal/safeio). The near-identical read in the link pass
+// was left on plain os.ReadFile, so a `mkfifo config.js` in a linked repo
+// parked a group-link worker in open(2) forever. This file grades the link
+// side of that.
+//
+// WHY EVERY TEST HERE IS WRAPPED IN A DEADLINE. The defect's symptom is a
+// hang, not a failure: run against the unhardened code these bodies never
+// return, and an un-bounded test would wedge the whole package suite instead
+// of reporting a red. runWithin turns "blocked" into a named test failure.
+// Every FIFO lives inside t.TempDir() and is removed with it.
+
+// fifoDeadline bounds each blocking-shaped call. The hardened path decides on
+// a stat(2) and returns in microseconds, so this is slack, not a measurement:
+// its only job is to be shorter than "forever" and longer than any plausible
+// scheduling delay on a loaded CI box.
+const fifoDeadline = 10 * time.Second
+
+// runWithin runs fn on its own goroutine and fails the test if it has not
+// returned within d. A blocked goroutine is deliberately abandoned rather
+// than cancelled: there is no portable way to interrupt a thread parked in
+// open(2), which is the whole reason safeio's guard is a stat gate and not a
+// timeout. The test binary exits at the end of the run regardless.
+func runWithin(t *testing.T, d time.Duration, what string, fn func()) {
+	t.Helper()
+	done := make(chan struct{})
+	go func() {
+		defer close(done)
+		fn()
+	}()
+	select {
+	case <-done:
+	case <-time.After(d):
+		t.Fatalf("%s did not return within %s: the read blocked on an irregular file (#6416, reached through the link pass)", what, d)
+	}
+}
+
+// mkfifo creates a FIFO at dir/name. dir MUST be a t.TempDir().
+func mkfifo(t *testing.T, dir, name string) string {
+	t.Helper()
+	p := filepath.Join(dir, name)
+	if err := syscall.Mkfifo(p, 0o600); err != nil {
+		t.Skipf("mkfifo unsupported on this filesystem: %v", err)
+	}
+	return p
+}
+
+// TestReadSourceFile_FIFORefusedWithoutBlocking pins the helper itself: a
+// FIFO with an indexed extension is refused as not-a-regular-file, and the
+// call returns. Against plain os.ReadFile this call never returns at all.
+func TestReadSourceFile_FIFORefusedWithoutBlocking(t *testing.T) {
+	dir := t.TempDir()
+	p := mkfifo(t, dir, "config.js")
+
+	var (
+		body []byte
+		err  error
+	)
+	runWithin(t, fifoDeadline, "readSourceFile on a FIFO", func() {
+		body, err = readSourceFile(p, maxSourceFileBytes)
+	})
+	if !errors.Is(err, safeio.ErrNotRegular) {
+		t.Fatalf("readSourceFile(FIFO) err = %v, want safeio.ErrNotRegular", err)
+	}
+	if len(body) != 0 {
+		t.Fatalf("readSourceFile(FIFO) returned %d bytes, want none", len(body))
+	}
+}
+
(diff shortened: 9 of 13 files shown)


cajasmota:
## Coordinator verification — seven rounds, and the guard is why

The hardening itself was solid from round 2: eleven `os.ReadFile`/`os.Open` calls on attacker-plantable source paths routed through `safeio`, with the one asymmetry that matters — `ErrNotRegular` skips the file, `ErrWouldBlock` still propagates — because a bare swap in `string_pass` would have converted a hang into a **silent repo-wide blackout** (#5523's shape). The before-state was reproduced independently by a reviewer: 4 tests wedged at 10.00s, 6 of 7 red, positive control green.

**Rounds 3–7 were all about the anti-divergence guard**, and they are worth reading as one finding rather than seven fixes.

I mutated the guard five times; **every one was ALIVE on arrival**:

| # | mutant | what it proved |
|---|---|---|
| 1 | walk root → `t.TempDir()` | reported the package clean having scanned **0** of 26 files |
| 2 | `roots` → a token matching nothing | scanned all 26, hit every anchor, **detected nothing** |
| 3 | `if false && len(offenders) > 0` | call site **discards** the matcher's entire output |
| 4 | `body: ""` | right file **names**, empty bodies |
| 5 | `body[:4000]` | right content, **truncated** — the real read sits at byte **12961** of a 16015-byte file |

Final state: **14 mutants, all DEAD**, each scored individually with `go vet` first, `-count=1`, and exact-count edits that abort on ≠1 match.

### The generalisable part

Two things came out of this that apply well beyond this file, and both are now on **#6834**:

**A guard is a three-stage pipeline — acquire → detect → react — and each stage can independently be a silent no-op.** That gives five layers (read anything / read the right files / read their full content / detect / act), and it *terminates*: once every stage is observed the guard cannot be vacuously green. Fixing one layer repeatedly exposed the next — layer 5 exists precisely **because** closing layer 4 means parameterising the matcher, which frees the call site to ignore it.

**Every weak fix pinned a *witness* of the property rather than the property.** Three of the five layers were closed twice for this reason:
- a **count** witnesses "read the right files" — defeated here because the package has 35 test files against 26 non-test, so inverting the filter scans *more* while opening zero production sources;
- a **parallel map** witnesses what the walk collected — the author caught this on themselves, having verified a `bodies` map while the guard consumed the `files` slice;
- a **token** witnesses content — defeated by truncation, since `package links` is on line 1.

The last one was **my** spec's fault: I asked for "assert a known string is present", and presence is strictly weaker than completeness. The fix is a length assertion against `os.Stat`, which is exact, subsumes the token check, and depends on no string surviving a refactor.

### Precedent this is not

`52aaa84f1` on main — *"the secrets scanner reported a repo it never read as clean"* — is the same defect that shipped. `scanGoEntityKinds` in `internal/types` carries layers 1 and 2 only, and it carries them because it was bitten first. This is the third instance, and the first where all five layers are closed.

### Verified before merge

Test-only since `f5d6dca63`, confirmed by diff — so the author's skip of `./internal/engine/` and `./cmd/grafel/` on the last two rounds is sound, and both were green at that SHA with identical non-test content. `./internal/links/` green at 392 `=== RUN`, `gofmt -l` clean, `go run ./tools/coverage validate` green.

`ci:full` added: this change carries `//go:build !windows` FIFO tests, and the guard was deliberately **untagged** so it runs on the Windows leg too.

Issue #6820, opened 2026-09-05
Electron IPC channels leak into the HTTP path and OpenAPI dashboard panes — the dashboard keys on bare "Endpoint", a different concept from SCOPE.Endpoint


cajasmota:
Electron IPC channels leak into the HTTP path and OpenAPI dashboard panes — the dashboard keys on bare `"Endpoint"`, which is a different concept from `SCOPE.Endpoint`

## The bug

There are two distinct entity kinds whose names differ only by prefix, and they mean **different things**:

- **`SCOPE.Endpoint`** — the HTTP entrypoint kind.
- **bare `"Endpoint"`** — declared only by the Electron rule pack (`internal/engine/rules/.../electron.yaml:41,46,52`), where it means an **Electron IPC channel**.

The dashboard keys on the **bare** spelling: `handlers_paths.go`, `handlers_export_openapi.go:186`, and 8 further sites.

So an indexed Electron app's IPC channels are presented as HTTP paths, and are exported into the **OpenAPI** output as if they were HTTP endpoints. They are not; they never had a URL.

## Status: pre-existing, user-visible, and not caused by any in-flight work

Found during a read-only grounding pass for #6776 arm B6, at `dab49ec0e`. It is **not** introduced by #6776 — the two spellings and the dashboard's choice of the bare one both predate it. Filing separately so it is not mistaken for migration fallout, and so it is not fixed *inside* an enum-membership batch where it would be reviewed as a naming change rather than as the output defect it is.

This is also the one genuinely **different-concept** pair among the twelve `SCOPE.`-prefixed/un-prefixed twins. The other eleven are same-concept (9) or need a concept note (2), and the codebase already folds those deliberately — `internal/.../classfold.go:36-38` states that both spellings must appear so lookups match. `Endpoint` is the exception, and the dashboard picked the wrong side of it.

## Why it matters more than a cosmetic mislabel

The OpenAPI export is a **published artefact**. An IPC channel appearing there is not a display glitch; it is a wrong claim about the surface a service exposes. Anyone reading the export to enumerate HTTP attack surface, generate a client, or diff API changes gets entries that do not exist over HTTP.

## What a fix needs to decide

1. **Which kind the dashboard means.** Almost certainly `SCOPE.Endpoint` for the HTTP panes — but that must be checked per site, because 10 sites are named and they may not all want the same answer.
[...]

Issue #6818, opened 2026-09-05
Nothing asserts an EntityKind enum member corresponds to a real rule declaration or producer — an invented kind passes every guard


cajasmota:
Nothing asserts that an `EntityKind` enum member corresponds to a real rule declaration or producer — an invented kind passes every guard

## The gap

Adding a fabricated kind to the enum and updating every pin that mentions it leaves the whole guard suite green:

```go
EntityKindBogus6776 EntityKind = "Bogus6776"
```

added to the constant block, to `AllEntityKinds()`, to `pinnedEntityKindVocabulary`, and to both declaration counts → `./internal/types/`, `./internal/entkinds/` and `./internal/graph/fbwriter/` **all pass**.

Found as an over-firing mutant (R11) during the independent review of #6776 arm B5, at `b8a8f4858`.

## Why the existing guards do not catch it

The suite is well built in the direction it was designed for — **a kind that exists in the rules must be in the enum** — and structurally blind in the other: **a kind in the enum need not exist anywhere else.**

- `TestNoUndeclaredRuleEntityKinds` walks rule declarations and checks they are accounted for. An enum member with no rule declaration is not a rule declaration, so it is never visited.
- `TestEntityKindDeclarationsMatchAllEntityKindsExactly` checks the constant block against `AllEntityKinds()` — both of which the fabrication updates.
- The vocabulary pin compares the roster against the enum — likewise updated.
- The ledger ratchets count *deferred* kinds, and a fabricated kind is not deferred.

Every guard is satisfied by construction, because they all check the enum against **itself** or against a set the fabrication has already joined.

Note a **misspelling** of a real kind IS caught (mutating `EntityKindMiddleware` to `"MiddleWare"` dies on both the vocabulary pin and `TestNoUndeclaredRuleEntityKinds`), because the real rule declaration then goes unaccounted-for. The hole is specifically an *addition* with no counterpart, not a corruption of an existing one.

## Why it matters now

It is **pre-existing and structural**, and #6776 arm B5 is safe — its 13 kinds were each verified against `entkinds.ScanRuleYAML` by hand, and a reviewer independently confirmed all 13 appear only in `kinds.go` with zero other references.

[...]

Issue #6776, opened 2026-09-02
DONE: rule-declared kind ledger retired 532→0 sites — IsValidEntityKind was the target, and no rename or stored-graph migration was needed


cajasmota:
Carved out of #6744 (merged `1fa7b35d5`), which froze the population behind a ratcheted ledger but deliberately did not migrate it. Figures below were produced by the scanner and **independently reproduced by a reviewer using a different method** (regex walk over the YAML keys, not the new package).

## The measurement

**532 entity-kind declaration sites in rule YAML, 28 distinct values, 26 of them not valid entity kinds.**

| value | sites | files |
|---|---|---|
| `Config` | 91 | 31 |
| `Route` | 73 | 31 |
| `Operation` | 68 | 28 |
| `Service` | 52 | 28 |
| `Model` | 42 | 17 |
| `Controller` | 37 | 21 |
| `Middleware` | 29 | 16 |
| `Schema` | 27 | 5 |
| … 18 more | | |

Kinds are declared in 21 of the top-level directories under `internal/engine/rules/`. Full ledger: `internal/entkinds/rule_declared_kinds_sweep_guard_6744_test.go`.

Only **two** values validate today: `SCOPE.IngressHost`, and `Module` — and `Module` validates by **coincidence**, because `types.EntityKindModule` happens to be un-prefixed itself.

## The question this issue exists to settle

Every one of the 26 is spelled without the `SCOPE.` prefix that Go-declared entity kinds carry. Two readings:

1. **A deliberate second namespace** for rule-declared kinds. If so it needs documenting, and `IsValidEntityKind` needs to know about it.
2. **An accident.** The evidence points here: nothing in the code treats `Route` differently from `SCOPE.Route`; `internal/engine/detector.go` writes the declared string **verbatim** into `EntityRecord.Kind`, so `IsValidEntityKind` is never consulted on this path at all; no doc describes a second namespace; and no consumer keys on the un-prefixed spelling.

**If it is an accident, the fix is large and cannot be a rename.** ~530 sites across 21 rule directories, plus a graph re-baseline, plus a **stored-graph migration** — every already-indexed group holds entities kinded with the old spelling, and a bare rename orphans them.

## Why it was not folded into #6744

[...]

Issue #6744, opened 2026-08-31
rules: three YAML rule files mint an un-prefixed 'ExternalAPI' kind that the Go-literal producer scan cannot see — the #6451 collision, unsplit, outside the guard


cajasmota:
Follow-up from #6451, which split `SCOPE.ExternalAPI` into `SCOPE.ExternalEndpoint` (HTTP-client calls, URL-path dialect) and `SCOPE.IngressHost` (k8s Ingress, hostname dialect) because one kind carried two unrelated meanings.

**That split covers the Go producers. A third producer family lives in the rule YAML and was deliberately left untouched — it still has the collision.**

## The sites

```
internal/engine/rules/javascript_typescript/frameworks/electron.yaml:70,74,78   entity_type: ExternalAPI
internal/engine/rules/kubernetes/extras.yaml:275,300                            entity_mapping: ExternalAPI
internal/engine/rules/kubernetes/extras.yaml:801                                entity_type: ExternalAPI
```

`internal/engine/detector.go:411` writes that value straight through as `Kind`.

## Two distinct problems

**1. It is the same collision, in the rule layer.** `kubernetes/extras.yaml:810` maps "Service type=LoadBalancer → ExternalAPI (cloud load balancer endpoint)" and `:801` extracts "hostname/path combinations as ExternalAPI endpoints" — while `electron.yaml` uses the same name for Electron IPC/HTTP surfaces. Exactly the two-meanings problem #6451 just separated on the Go side, unseparated here.

**2. The name is un-prefixed, so it is a different string entirely.** These emit `ExternalAPI`, not `SCOPE.ExternalAPI`. That means:
   - they were never affected by the #6451 rename, and
   - **they escape `internal/types/producer_kinds_test.go`, which scans Go literals only.**

That second point is the more important one. A guard that enumerates producers by scanning Go source cannot see a producer declared in YAML, so the rule layer can mint any kind string it likes — valid, retired, or misspelled — with nothing to catch it. #6451 retired `SCOPE.ExternalAPI` from the enum precisely so a stale producer would fail `IsValidEntityKind`; the YAML producers sit outside that net.

## Scope to establish first

- **Enumerate the real population.** How many distinct `entity_type` / `entity_mapping` values do the rule files declare, and how many are not valid entity kinds? `ExternalAPI` may not be the only one. Do this before deciding the shape of the fix — the answer determines whether this is a three-site rename or a systemic gap.
[...]

PR #6840, opened 2026-09-05
feat(#6776): arm B8 — Constraint, Plugin and Template into the enum; Endpoint held back for its concept ruling

cajasmota:
# #6776 arm B8 (partial) — Constraint, Plugin and Template into the enum; the ledger lands at 1, not 0

`Constraint`, `Plugin` and `Template` join `types.AllEntityKinds()` in their CURRENT
un-prefixed spellings. No rename, no `SCOPE.` prefix, no stored-graph migration.
`KindVocabularyVersion` stays **1** — verified against the guard's own body, not relayed:
`kind_vocabulary_bump_guard_6779_test.go:271` demands a bump only under
`len(removed) > 0 && KindVocabularyVersion <= pinned`, and this change removes nothing.

## `Endpoint` is deliberately excluded, so the ledger goes 4 → 1

`Endpoint` is the one pair on #6744's ledger that is **not a synonym**. The bare spelling is
declared only by `javascript_typescript/frameworks/electron.yaml` (Electron IPC channels);
`SCOPE.Endpoint` names the HTTP concept. Every pair migrated in B5–B7 was a true synonym.
Migrating this one as though it were would put two different concepts under a single enum
member — that is **#6820**, and it needs an owner ruling that has not been given. So
`ruleDeclaredKindsDeferredMax` lands at **1**, not 0, and `Endpoint` stays on the ledger.
This is a decision, not an oversight, and it is written into the kinds.go const block, into
the fbwriter file header, and into internal/types' negative-test row so no reader has to
infer it.

## Ratchets, both read from disk

| ratchet | before | after | why |
|---|---|---|---|
| `ruleDeclaredKindsDeferredMax` (internal/entkinds) | 4 | **1** | three migrated |
| `goUnprefixedKindsDeferredMax` (internal/types) | 5 | **5, unchanged** | see below |
| `enumOnlyEntityKindsMax` (#6818 allow-list) | 1 | **1, unchanged** | no row added |
| `KindVocabularyVersion` | 1 | **1** | nothing removed |
| EntityKind constant counts | 90/89 | **93/92** | read from the file |

`Route` paid both ratchets in B7 because it was on `goUnprefixedKindsDeferred` too. **None of
B8's three is**, derived from the file rather than assumed: that ledger holds exactly
`ChannelEvent`, `File`, `Stream`, `Subscription`, `relationship`
(`producer_entity_kinds_6776_test.go:148-154`). Graded rather than asserted — mutant **M5**
below puts `Plugin` on it with the pin raised to 6 and dies naming the coupling.

## Evidence per kind, read at the site

`#6818` clause **(a)** — rule-pack declaration — covers all three, and the site inventory was
re-derived from the tree, matching the ledger comments this change deletes:
`Constraint` 1 site (`python/frameworks/sqlalchemy.yaml:79`), `Plugin` 5 sites in 2 files
(`kotlin/frameworks/ktor.yaml:40,62`; `javascript_typescript/frameworks/fastify.yaml:35,38,48`),
`Template` 2 sites in 2 files (`python/frameworks/flask.yaml:52`;
`ansible/frameworks/ansible_core.yaml:37`). **No allow-list entry was added**:
`enumOnlyEntityKinds` still holds exactly `SCOPE.ScopeUnknown`.

Concept notes, each read at both ends:

- **Constraint** — bare: `sqlalchemy.yaml:79`, pattern `ForeignKey("table.id")`, a constraint
  declared on a SQLAlchemy table/model. `SCOPE.Constraint` (`kinds.go:150-153`, #749): a Django
  `Model.Meta.constraints` entry bound to its parent Model by CONTAINS. One concept — a
  declarative constraint on a persisted model — through two Python ORMs.
- **Plugin** — bare: both source-pattern sites are the *act of registration*, `ktor.yaml:62`
  matching `install(<Name>)` and `fastify.yaml:48` matching `<app>.register(<Name>)`, with
  `*/plugins/*` file conventions beside them. `SCOPE.Plugin` (`kinds.go:1295-1306`, #3628 area
  #25): "a single plugin / extension that a build tool, application, or framework registers",
  linked by REGISTERS_PLUGIN. Same concept, same relationship, different extractor.
- **Template** — bare: `flask.yaml:52` (`templates/**/*.html`) and `ansible_core.yaml:37`
  (`roles/*/templates/*.j2`), a server-side template FILE. `SCOPE.Template`
  (`kinds.go:310-323`, #3628): the server-side view template by normalized logical path, and
  its doc names Flask `render_template` as a producer. Same concept, from the two ends of one
  render.

**A citation I was briefed toward and did NOT use.** `internal/engine/classfold.go:52` has
`"Plugin": 80` in `FrameworkClassKindPriority`. It is **not** pairing evidence: there is no
`"SCOPE.Plugin"` row in that map, even though the map's own doc at `classfold.go:35-38` says
"Both bare kind names AND their SCOPE.-prefixed forms must appear so that lookups match".
That is the same half-pair shape B6 found for `Component`, one map further along. `Constraint`
has no row either, and *that* one is deliberate — `classfold.go:26-27` names "a Django Meta
`Constraint`" as its example of a nested artifact that must never win a fold. Both facts are
recorded in the const block instead of a citation that would not survive being opened.

## Fixture churn — budgeted, and the replacement control is durable

`Plugin` and `Template` were `internal/graph/fbwriter`'s non-enum controls, so its
`fixture is inert` guards fired as designed. `Endpoint` survives as one control; the other slots
were re-picked as **`File`**.

The guards' own repair advice — "re-pick from internal/entkinds' ledger" — is **wrong for this
arm**, and the file header now says so: that ledger holds one entry, it is `Endpoint`, it is
already in use here, and it is itself a migration candidate once #6820 is ruled on. `File` is
chosen from the other direction — the kinds outside the enum because *nothing proposes to move
them*:

- `internal/engine/commit_coupling_edges.go:117`, `const KindFile = "File"` — a live Go
  producer, never rule-declared, so not on #6776's rule-migration path at all;
- `goUnprefixedKindsDeferred`'s own comment: "an internal commit-coupling artefact rather than
  a first-class kind; **it is ledgered here, not promoted**";
- `internal/types`' arm-B4 negative test has carried it as a permanent non-enum row for that
  same reason since B4, so this is not a new claim about it.

Because internal/types already had a `File` row, the mirror shrinks from two rule-declared rows
to one plus `File` rather than gaining a row, and the prose says which two kinds fbwriter now
names. That the new control is load-bearing is **graded**, not argued (M6).

## Mutants — permissive first, scored individually, `go vet` first, `-count=1`, each edit proved to land

Every mutant was applied by an exact-count replace that aborts on 0 or >1 matches, and every
restore was `cp` from a byte-for-byte backup with a verified `shasum`. No `git checkout --`.

| # | mutant | verdict |
|---|---|---|
| M1 | `EntityKindConstraintBare = "ConstraintZZ"` (fabricate the value) | **DEAD** — `TestEveryEnumEntityKind6818_CorrespondsToSomethingOutsideTheEnum`, `TestNoUndeclaredRuleEntityKinds`, `TestEntityKindVocabularyIsPinnedToItsVersion` |
| M2 | `EntityKindPluginBare = "PluginZZ"` | **DEAD** — same three |
| M3 | `EntityKindTemplateBare = "TemplateZZ"` | **DEAD** — same three |
| M4 | ledger shrunk to 1 but `ruleDeclaredKindsDeferredMax` left at 4 | **DEAD** — `TestRuleDeclaredKindsRatchetOnlyShrinks` (the shrink half of the ratchet, alone) |
| M5 | `"Plugin"` added to `goUnprefixedKindsDeferred`, pin raised 5→6 | **DEAD** — `TestProducerEntityKinds6776_UnprefixedLedgerIsExact`: *"ledgered kind \"Plugin\" is no longer declared by any Go producer (or has become valid)"*. This is the positive control on "B8 owes nothing to the second ratchet" |
| M6 | declare the NEW control `File` in the enum | **DEAD** — 5 tests across `fbwriter` and `types`, 3 of them through explicit `fixture is inert` guards. `entkinds` stays green, correctly: `File` is not rule-declared |
| M7 | empty `ruleDeclaredKinds6776` and re-pin its length to 0 | **ALIVE at first, now DEAD** — see below |
| M8 | `ruleDeclaredKinds6776` grown to 2, pin left at 1 | **DEAD** — on the exact pin ALONE (`:530`), the new floor passing. The two clauses isolate; no mutual masking |

### The defect M7 found, in prose I had just written myself

I wrote a comment saying an empty transcription "would make the loop unreachable and this test
vacuously green, which is why zero is a failure". Nothing observed that. The length check was
`!= 1`, so emptying `ruleDeclaredKinds6776` and re-pinning to `!= 0` left
`TestEveryRuleDeclaredKindOnTheLedgerIsCountedByTheWritePath` **green over an empty loop** —
measured ALIVE. That is exactly the "prose asserts what no test observes" class, and it matters
here rather than hypothetically: the ledger is at 1, and the moment #6820 rules on `Endpoint`
this test can be quietly emptied instead of deleted.

Fixed by a **separate absolute floor** (`len(...) == 0` is a hard failure) that the pin cannot
absorb, because re-pinning is what defeats the pin. M7 re-scored **DEAD**; M8 shows the pin
still fires alone, so the two are not a mutually-masking pair.

## Prose corrected while touching it

- `internal/entkinds`' file header said "the ledger now holds the remaining 12" — stale since
  B6. Replaced with the arm-by-arm breakdown and an explicit instruction not to restate the
  remainder as a number, since that sentence went stale three arms running; the ledger variable
  is pinned exactly and is always current.
- `internal/graph/fbwriter`'s header carried the same stale 12.
- `internal/types`' arm-B4 header still named `Route`/`Config` as fbwriter's controls, two arms
  after they stopped being. Rewritten to describe the mechanism and defer to the row list.
- `kinds.go` still called `Template` "a deferred twin" in the arm-B5 zero-producing paragraph.

## Verification

`go test -count=1` green on `./internal/types/`, `./internal/entkinds/`, `./internal/graph/`,
`./internal/graph/fbwriter/`, `./internal/quality/`, plus `./internal/engine/` and
`./internal/resolve/` (both read these kinds). `gofmt -l` clean over `internal`, `cmd`, `tools`;
`go vet ./internal/...` clean. `go run ./tools/coverage validate` reports **822 records / 17532
warnings**, identical to arm B7's figure. No extractor, resolver or graph-assembly code is
touched, so `./cmd/grafel/`'s whole-graph digest pin is not in scope.

## Not done, and flagged

- **No corpus re-measurement.** Arms B6/B7 quoted `entity_entities_kind_not_in_enum` from a full
  in-process index of `testdata/fixtures`; I did not run one. The three kinds' populations are
  therefore unquantified — `Constraint`/`Template` are on arm A's zero-producing list, so the
  expected drop is small, but that is an inference and should be measured before being quoted.
- **A forward hazard for the #6820 arm.** `TestEnumEntityKinds6818_NoDeferredLedgerOverlapsTheEnum`
  (`enum_member_corresponds_6818_test.go:446`) hard-fails when `ruleDeclaredKindsDeferred` is
  EMPTY, as an anti-vacuity guard. With the ledger at 1, migrating `Endpoint` empties it, so that
  arm must retire the ledger and its guards deliberately rather than ratchet to zero.
- **#6829 and #6830 are untouched** and neither is made worse.

---

Refs #6776

Deliberately **not** `Closes` — arm B8 migrates three of the ledger's four kinds and leaves
`Endpoint` on it pending the concept ruling on #6820, so `ruleDeclaredKindsDeferred` lands at 1
rather than 0 and #6776 stays open.


cajasmota:
## Coordinator verification — APPROVE, and the review corrected me three times

Independent review: **APPROVE**, 12 mutants scored individually, with the branch and base both verified green on five packages plus `./cmd/grafel`, `./internal/engine`, `./internal/resolve` and `./internal/mcp` (run because `kinds.go` is extractor-adjacent and carries a digest-pin hazard).

**My own killing mutant, which nobody else ran:** replace the ledger's remaining entry with a kind no rule pack declares (`"Endpoint"` → `"ZZNotDeclaredAnywhere"`, comment and family left intact). **DEAD on three guards** — `TestNoUndeclaredRuleEntityKinds`, `TestYAMLHalfObservesDeclaredKinds`, `TestLedgerIsDescribedByItsOwnScan`. So the ledger is pinned to *reality*, not merely to its own size; a row cannot claim rule-declaredness it does not have. Worth stating because "the ratchet only counts" was a plausible reading of the diff and is wrong here in the same way it was wrong for B7.

### Three corrections the review made to my framing — all confirmed

1. **The `Endpoint` forward hazard is three guards, not one.** I briefed `enum_member_corresponds_6818_test.go:446` as the single empty-ledger trap. It is `:447`, and there are two more: `rule_declared_kinds_sweep_guard_6744_test.go:516` (`TestYAMLHalfObservesDeclaredKinds`) and fbwriter's own floor.
2. **`Endpoint` is the sole non-enum control in two further fbwriter fixtures** the header does not enumerate (`:267`, `:607`), both still carrying the stale "re-pick from the ledger" advice. So #6820's arm is **four fixtures and three guards**, not the two the header implies.
3. **My phrasing "two different concepts under one enum member" was imprecise.** `EntityKindEndpointBare` and `EntityKindEndpoint` would be *two* members, as with every B5–B8 pair. The hazard is a **misleading vocabulary name**, not a member collision. The conservative call to exclude stands; the reason I gave for it did not.

Also confirmed against me: my `classfold.go` citation for `Plugin` was **wrong** — `:52` has `"Plugin": 80` with no `"SCOPE.Plugin"` row, while the map's own doc at `:35-38` requires both spellings. The author refused the citation and used the rule sites instead. That is the second time I have cited that file for a kind it does not cover.

### Evidence re-derived with the right tool

Site counts were re-run through **`ScanRuleYAML`, not grep** (714 YAML files, 532 sites, 0 unresolved): **Constraint 1, Plugin 5, Template 2** — exact, and line-exact at every cited site. `Endpoint` is 3 sites, **all** in `electron.yaml`, which is the whole basis for excluding it.

### Report-only, carried forward rather than fixed here

- **Duplicates in `AllEntityKinds()` are still undetected** (mutant ALIVE) — confirms **#6830**. B8 does not worsen the mechanism but adds three more entries to a list declared as a slice and consumed as a set.
- **The `Template` concept note over-claims for one of its two sites.** `SCOPE.Template`'s doc is explicitly *server-side view templates*; `flask.yaml:52` fits, but `ansible_core.yaml:37` (`roles/*/templates/*.j2`) is a config-file template rendered against a host — same engine, different concept. This matters only because the same "is it a synonym?" test is what justifies excluding `Endpoint`; applying it loosely here slightly weakens that argument.
- **`classfold.go`'s doc over-claims well beyond `Plugin`** — 11 bare rows have no `SCOPE.`-prefixed twin (`Controller, Middleware, Repository, Worker, Job, Topic, TestClass, Plugin, Implementation, Interface, Task`) against a doc saying both must appear. Pre-existing, untouched here, and no issue covers the general case. **Filing separately.**
- The commit message's "DEAD across 5 tests in 2 packages" for the `File` control is reproducible only for a partially-completed mutant variant (the reviewer measured 7/2 minimal, 3/1 fully-completed). Not wrong, but not reproducible from the message alone.

**No corpus figure is quoted anywhere in this change**, deliberately — no full index was run. B6 and B7 both quoted one, so the absence is a stated choice rather than an omission.

Issue #6834, opened 2026-09-05
Arm 2: sweep the remaining 16 of 19 guard sites for the five vacuity layers — arm 1 shipped, layer 3 measured open


cajasmota:
Sweep every static-scan guard for the vacuous-walk hole: three instances found, only one carries the check

## The defect shape

A guard that walks a tree, collects offenders, and fails if the offender list is non-empty **passes when the walk reads nothing**. Empty input → no offenders → green. The report is indistinguishable from a genuinely clean tree, and it is *more* dangerous than having no guard at all, because it occupies the slot and reports success.

This is not a hypothetical class in this repo. It is the third confirmed instance:

| where | status |
|---|---|
| the secrets scanner | **shipped defect**, fixed in `52aaa84f1` — *"reported a repo it never read as clean, including a caged root it scanned zero bytes of"* |
| `scanGoEntityKinds`, `internal/types/producer_entity_kinds_6776_test.go` | **has the check**: hard-fails on `FilesParsed < 500` and `len(Sites) < 200`, with the comment *"a walk that reaches no files reports no offenders and looks like a clean tree"* |
| `TestNoUnhardenedSourcePathReadsInPackage`, `internal/links` (#6823) | **no check** — proved ALIVE below |

## Proof, on the third one

Mutating the walk root to an empty directory:

```
os.ReadDir(".")  ->  os.ReadDir(t.TempDir())
--- PASS: TestNoUnhardenedSourcePathReadsInPackage (0.00s)
```

Green, having scanned **zero** of the package's 26 non-test `.go` files. The same hole opens if the `.go`/`_test.go` filter is inverted, or if a rename stops the suffix matching. That one is being fixed in #6823's own PR, so this issue is about the **other** guards nobody has checked.

## Why a sweep rather than three fixes

The two instances that had the hole were written by different people at different times, and the one that carries the check carries it because it was *bitten first*. That is the signature of a class nobody has swept for, not of two unlucky files. Every guard of this shape needs the same one-line floor, and the ones that already have it should be confirmed rather than assumed.

**Scope:** every test that walks a directory (or a rule pack, or a fixture corpus) and asserts on what it *did not* find. `internal/entkinds`, `internal/types`, `internal/links`, `internal/quality` and `scripts/quality/` are the obvious places to start; the search is for the pattern, not for a fixed file list.

[...]

Issue #6842, opened 2026-09-05
home_sweep_guard_6178_test.go walks into .claude/worktrees/ — the repo-root trap, inside a test


cajasmota:
`home_sweep_guard_6178_test.go` walks into `.claude/worktrees/` — the repo-root trap, inside a test

## The defect

`scanHandRolledHomePaths` (`internal/registry/home_sweep_guard_6178_test.go:475`) is called at `:216` with `root := repoRoot(t)`, and its `SkipDir` switch at `:483-:485` lists:

```
".git", "node_modules", "vendor", "testdata", "dist", "build"
```

**`.claude` is not there.** Every other walker in the repo excludes it — `internal/entkinds/scan.go:220` (`skippedDir`) does, and the local walk in `internal/types/producer_entity_kinds_6776_test.go` does.

In any development checkout that has agent worktrees under `.claude/worktrees/`, this walk therefore parses **every worktree's full copy of the production tree**. Three consequences:

1. It `t.Fatalf`s on a parse error inside any worktree — including one belonging to a *different* branch mid-edit, which has nothing to do with the invariant being guarded.
2. It reports offences under keys like `.claude/worktrees/agent-X/internal/...` that **can never match the allow-lists**, which are written against real repo paths.
3. It does the work N+1 times, where N is the number of live worktrees.

## Why it is worth fixing rather than tolerating

This is the same trap we hit repeatedly at the command line — greps against the repo root time out and inflate match counts roughly 5× because `.claude/worktrees/` holds full checkouts. Here it is baked into a **guard**, where the failure mode is worse than slow: a test that fails because of an unrelated in-flight branch is a test people learn to re-run rather than read.

CI never sees it (no worktrees there), so this is a **local-development-only** failure — the kind that costs an afternoon exactly once per person and is invisible to the people who could fix it.

## The fix

Add `.claude` to the `SkipDir` switch. Better: have the three walkers share one exclusion list rather than each maintaining its own — `entkinds`' `skippedDir` already has the right set, and the divergence between the three is what produced this.

Scoring it needs care, because the obvious test ("does it skip `.claude`?") passes trivially in CI where no such directory exists. A real check creates a `.claude/worktrees/...` directory under a `t.TempDir()` root with a deliberately unparseable Go file in it, and asserts the walk neither fails nor reports it.

## Not established

- **Whether this has actually bitten anyone yet.** It would surface as a confusing local failure, and nothing records that.
[...]

PR #6844, opened 2026-09-05
test(#6834): arm 1 — grade scanGoEntityKinds against all five vacuity layers; three were ungraded, one stays measured-open

cajasmota:
# #6834 arm 1 — `scanGoEntityKinds` graded against all five vacuity layers

Refs #6834

Branch `lane/6834-guard-vacuity`, base `a1900a432`. Test files only —
`internal/types/producer_entity_kinds_6776_test.go`.

Review round 2 folded in: the walk check is now itself graded, the ledger control's masked direction
is unmasked, layer 3 is recorded as **measured open**, and the "wrong diagnosis" claim is re-grounded.

## What was actually open on `main`

| layer | mutant | verdict on `main` |
|---|---|---|
| 1 | `root` → `t.TempDir()` | DEAD, correct diagnosis |
| 2 | invert `parseGoTree`'s `.go`/`_test.go` filter | DEAD — but **both floors cleared**: 2662 files > 500, 2888 sites > 200. No count floor can see this; only anchors can. |
| 3a | arbitrary truncation `src[:4000]` | DEAD by construction — the parser rejects it and `parseGoTree` hard-errors |
| 3b | `f.Decls = f.Decls[:len/2]` | DEAD, but 214 sites — 14 above the floor |
| 3c | **truncate at the last top-level `\n}\n` inside 85% of each file** | **ALIVE**, and still ALIVE on this branch — see "Layer 3" below |
| 4a | neuter the bare-identifier resolver | DEAD (`ResolverSeesWhatLiteralsCannot`) |
| 4b | `key.Name != "Kind"` → `"ZzzNoSuchField"` | DEAD (site floor, correct diagnosis) |
| 5a | `if false && !ledger[kind]` | **ALIVE** |
| 5b | return from `assertLedgerExact` before comparing | **ALIVE** |

**Layer 4 needed nothing.** The fixture-tree tests drive the *real* `entkinds.ScanGo` in both
directions with exact expected sets. Both matcher mutants died on `main`. Nothing was added there.

## The fixes

**1. `checkWalk(tb testing.TB, root string, res, anchors)`** — the floors plus a must-scan anchor set
(`goScanAnchors`), extracted so it can be observed, with the anchor set as a parameter so the control
drives *this* function rather than a copy.

**2. `TestCheckWalk_ActsOnWhatTheScanReturns`** — because the first cut of the anchor check was itself
an ungraded no-op in two ways: emptying `goScanAnchors` was green, and `if false && broken != ""` was
green. That is this issue's own defect class reappearing one function above where it had just been
closed, exactly as #6823 predicted. Five directions, each on an otherwise **healthy** synthetic
`Result`: healthy passes, a missing anchor fails, an **empty anchor set** fails, a below-floor file
count fails, a below-floor site count fails.

**3. `TestAssertLedgerExact_ActsOnWhatTheScanReturns`** — the reaction control that closed 5a/5b.
`assertLedgerExact` takes `testing.TB`; a `recordingTB` observes it.

**4. Direction 3 of that control was masked** (review B2). Its fixture was `found={}` against a
one-row ledger, which *also* violates direction 2 — so with the pin check deleted, execution fell
through to the stale-row check, that fired on the same fixture, and the mutant read DEAD on a failure
it did not cause. Two guards that only fire together grade neither. The sets now agree and only the
pin is wrong.

**5. `minFilesParsed` / `minSites` constants** — the message used to hardcode `want at least 500`
beside a comparison against the constant, so a re-pin would print a self-contradicting message.

## Layer 3: measured OPEN, and my structural argument was wrong

I claimed a prefix of a Go file is not a Go file, so `parseGoTree` hard-errors. That holds only for
*arbitrary* truncation. Truncating each file at the last top-level `\n}\n` inside 85% of its length —
verified independently here:

```
PROBE files= 2015 sites= 453 anchorsHit= 3 of 3
go test -count=1 ./internal/types/   ->  ok   (4.294s)
```

Every file parses, `FilesParsed` is unchanged at 2015, all three anchors are hit, **566 sites become
453** — a fifth of the producers this file exists to grade go invisible — and the package is green.

**Chosen outcome: retract the claim, record the layer as open.** The doc comment now names the
mutant, the 566→453 figure, and the fact that closing it needs the #6823 form (collected length vs
`DirEntry.Info().Size()`), which `parseGoTree` does not expose. Closing it properly means changing
production code in `internal/entkinds` to surface per-file sizes, which is a larger change than arm 1
should carry and belongs in its own arm with `./cmd/grafel/` in scope. The sentence claiming the
anchors covered content ("they sit deep in long files on purpose") was prose asserting what no test
observes — in the file whose subject is that defect — and is deleted.

## Re-grounding the "wrong diagnosis" claim (review B4)

My earlier framing said the layer-2 mutant's only diagnosis misdirects. That is **overstated and I
withdraw it.** Measured: it kills six tests with ~1085 lines; my `Stream` line is 3 of those; two of
the six are fixture-tree tests printing `collected = map[]`, which no ledger edit can silence; and
deleting the three stale rows leaves 57 `NEW Go-declared entity kind` failures plus four other
failing tests. Following the misleading rows does **not** yield a green broken scan.

What survives, and is the real result: **the two count floors provably cannot see this mutant**
(2662 files, 2888 sites) — only the anchors distinguish "did work" from "did *this* work" — and the
diagnosis consolidates from ~1085 lines that never name the walk to one line that does.

**6. The wiring itself was ungraded** (review round 3, and the worst of the three). Deleting the
`checkWalk` call from `scanGoEntityKinds` entirely was **ALIVE** — the package stayed green with no
walk guard at all in production, because `TestCheckWalk_ActsOnWhatTheScanReturns` drives `checkWalk`
directly. B3 loses a fifth of the sites; this loses the whole guard, floors and anchors and the
empty-anchor case together, while the file still reads as though every layer were covered.

The mechanism deserves stating once, because it produced B1, B2 and this in succession:
**parameterising a decision so a control can drive it is exactly what frees the caller to ignore
it.** Every extraction has to be followed by a control one level up, or it trades a graded no-op for
an ungraded one. So `scanGoEntityKindsIn(tb testing.TB, root string)` now holds the wiring,
`scanGoEntityKinds` supplies the real root, and `TestScanGoEntityKindsIn_RunsTheWalkCheck` drives the
wiring with a `recordingTB` against a `t.TempDir()`. Its expected message is pinned precisely enough
to say *which* check ran (`parsed only 0 files`), so passing `nil` anchors at the call site — which
would report the empty anchor set instead — is caught there too.

## Every mutant re-scored after the restructure (verdicts do not carry)

Whole-package runs, `go vet` before each, `-count=1`, exact-count replace aborting on ≠1 matches with
a post-check that the edit landed, restore by `cp` with shasum equality verified after every mutant.

| id | mutant | verdict | killed by |
|---|---|---|---|
| W1 | **`checkWalk` call deleted from the scan** | DEAD | `TestScanGoEntityKindsIn_RunsTheWalkCheck` **alone** |
| W2 | `nil` anchors passed at the call site | DEAD | `TestScanGoEntityKindsIn_…` first (+4) |
| N1 | `goScanAnchors` emptied | DEAD | `TestCheckWalk_…` (+4 live-tree tests) |
| N2 | `if false && broken != ""` | DEAD | `TestCheckWalk_…` **alone** |
| N3 | walk check fails unconditionally (over-fire) | DEAD | `TestCheckWalk_…` (+4) |
| N4 | file floor discarded | DEAD | `TestCheckWalk_…` alone |
| N5 | site floor discarded | DEAD | `TestCheckWalk_…` alone |
| N6 | anchor comparison discarded | DEAD | `TestCheckWalk_…` alone |
| M7 | pin check discarded | DEAD | `TestAssertLedgerExact_…` alone |
| M5 | found→ledger direction discarded | DEAD | `TestAssertLedgerExact_…` alone |
| M6 | ledger→found direction discarded | DEAD | `TestAssertLedgerExact_…` alone |
| M13 | whole ledger reaction neutered | DEAD | `TestAssertLedgerExact_…` alone |
| M8 | ledger reaction fails unconditionally (over-fire) | DEAD | `TestAssertLedgerExact_…` + `UnprefixedLedgerIsExact` |
| M1 | `t.TempDir()` walk root | DEAD | 4 live-tree tests |
| M2 | inverted `.go`/`_test.go` filter | DEAD | 6 live-tree tests |
| M4b | `key.Name != "ZzzNoSuchField"` | DEAD | 6 live-tree tests |
| **3c** | **85% top-level truncation** | **ALIVE** | — layer 3, recorded open above |

Eight of the sixteen die by the three new controls **alone** — no live-tree test sees them.

## Still not graded (stated, not hidden)

- **Layer 3**, per above.
- The anchors are three pairs. A walk reading exactly those three files and nothing else passes.
  Closing that needs `FilesParsed` compared to an independently computed count of non-test `.go`
  files under `internal/`, i.e. a second copy of `skippedDir` in the test — the trap this class is
  about. Deliberately not done.
- The other consumers of `scanGoEntityKinds` hold their reaction inline in the test body, where there
  is nothing to discard it into; not re-shaped.

## Verification

- `go test -count=1 ./internal/types/ ./internal/entkinds/` — ok (4.6s / 11.5s)
- `gofmt -l` on the touched file — clean
- `go run ./tools/coverage validate` — exit 0, `822 record(s) … 1163 test(s) checked`
- `./cmd/grafel/` not run: test files only, no non-test file touched.

### DIFF
--- a/internal/types/producer_entity_kinds_6776_test.go
+++ b/internal/types/producer_entity_kinds_6776_test.go
@@ -88,6 +88,7 @@ package types_test
 // internal/entkinds/rule_declared_kinds_sweep_guard_6744_test.go.
 
 import (
+	"fmt"
 	"go/ast"
 	"go/parser"
 	"go/token"
@@ -156,9 +157,46 @@ var goUnprefixedKindsDeferred = map[string]bool{
 // goUnprefixedKindsDeferredMax pins the ledger's exact size.
 const goUnprefixedKindsDeferredMax = 5
 
+// goScanAnchors are production sites the scan MUST reach, named as
+// (file, kind) pairs. They are the must-scan half of the walk-integrity check
+// and exist because a COUNT is only a witness of "read the right files", not
+// the property.
+//
+// Concretely: internal/ holds 2663 _test.go files against 2101 non-test ones,
+// so inverting parseGoTree's `.go` / `_test.go` filter parses MORE files than
+// the correct walk and clears any sane floor while opening not one production
+// source. Measured, not assumed — that mutant leaves both floors below intact
+// (#6834 arm 1).
+//
+// Each pair is a kind the ledgers below already name with its file, so a
+// producer that genuinely moves updates both in one change. They sit deep in
+// long files on purpose: a walk that reads a PREFIX of the tree's content
+// tends to keep the first declarations and lose the last.
+var goScanAnchors = []struct{ file, kind string }{
+	{"engine/workflow_edges.go", "SCOPE.StateMachine"},
+	{"engine/commit_coupling_edges.go", "File"},
+	{"engine/sse_edges.go", "Stream"},
+}
+
 // scanGoEntityKinds runs the shared resolver over internal/ — the same subtree
-// the literal guard walked — and fails loudly if the walk read nothing, since a
-// walk that reaches no files reports no offenders and looks like a clean tree.
+// the literal guard walked — and fails loudly if the walk did not do this
+// work, since a walk that reaches no files reports no offenders and looks like
+// a clean tree.
+//
+// Three questions, one failure site, because they are one question — "is the
+// input the ledgers are about to be compared against real?":
+//
+//  1. did it read ANYTHING? The FilesParsed floor. 2015 files parse today, so
+//     500 is a floor and not a count.
+//  2. did it read the RIGHT files? goScanAnchors. The floor alone cannot say
+//     so; see that variable's doc for the measurement.
+//  3. did it get anything OUT of them? The site floor. 566 sites today.
+//
+// The message must diagnose the WALK, not the tree. Without these checks the
+// mutants above still fail the package — but through the ledger tests, saying
+// `ledgered kind "Stream" is no longer declared by any Go producer. Delete its
+// row and lower the pin by one.` That is the opposite diagnosis, and following
+// it deletes the ledger rows and leaves the broken walk permanently green.
 func scanGoEntityKinds(t *testing.T) entkinds.Result {
 	t.Helper()
 	root := filepath.Join(repoRoot(t), "internal")
@@ -166,11 +204,31 @@ func scanGoEntityKinds(t *testing.T) entkinds.Result {
 	if err != nil {
 		t.Fatalf("ScanGo(%s): %v", root, err)
 	}
-	if res.FilesParsed < 500 {
-		t.Fatalf("ScanGo parsed %d files under internal/; the walk is broken", res.FilesParsed)
+	broken := ""
+	switch {
+	case res.FilesParsed < 500:
+		broken = fmt.Sprintf("it parsed only %d files, want at least 500", res.FilesParsed)
+	case len(res.Sites) < 200:
+		broken = fmt.Sprintf("it resolved only %d entity-kind sites, expected hundreds", len(res.Sites))
+	default:
+		for _, a := range goScanAnchors {
+			seen := false
+			for _, s := range res.Sites {
+				if s.File == a.file && s.Kind == a.kind {
+					seen = true
+					break
+				}
+			}
+			if !seen {
+				broken = fmt.Sprintf("it never resolved %q in %s — it read something, but not the production sources these ledgers grade", a.kind, a.file)
+				break
+			}
+		}
 	}
-	if len(res.Sites) < 200 {
-		t.Fatalf("ScanGo resolved %d entity-kind sites; expected hundreds", len(res.Sites))
+	if broken != "" {
+		t.Fatalf("THE SCAN'S WALK IS BROKEN, not the tree: %s (walked %s, %d files, %d sites). "+
+			"An empty or shrunken result from a walk that did not read internal/'s production sources "+
+			"is not evidence about the ledgers (#6834; cf. 52aaa84f1).", broken, root, res.FilesParsed, len(res.Sites))
 	}
 	return res
 }
@@ -205,10 +263,15 @@ func sortedKeys(m map[string][]entkinds.Site) []string {
 // assertLedgerExact compares found against ledger in BOTH directions and pins
 // the ledger's size, so neither an appended row nor a silently-migrated kind
 // can leave the ledger stale.
-func assertLedgerExact(t *testing.T, label string, found map[string][]entkinds.Site, ledger map[string]bool, pin int) {
+// It takes testing.TB rather than *testing.T so
+// TestAssertLedgerExact_ActsOnWhatTheScanReturns can hand it a recorder and
+// observe both outcomes. Nothing follows the Fatalf, so the function behaves
+// the same whether or not Fatalf is terminal for the TB it was handed.
+func assertLedgerExact(t testing.TB, label string, found map[string][]entkinds.Site, ledger map[string]bool, pin int) {
 	t.Helper()
 	if len(ledger) != pin {
 		t.Fatalf("%s: ledger has %d entries but the pin says %d; move both or neither", label, len(ledger), pin)
+		return
 	}
 	for _, kind := range sortedKeys(found) {
 		if !ledger[kind] {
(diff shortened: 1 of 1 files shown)


cajasmota:
## Coordinator verification — and the mechanism this arm actually found

**My killing mutant, which nobody else ran:** delete the production call entirely.

```go
checkWalk(t, root, res, goScanAnchors)  →  _ = root
go test ./internal/types/  →  ok        ← ALIVE
```

`scanGoEntityKinds` could stop invoking the walk guard **altogether** and the package stayed green, because `TestCheckWalk_ActsOnWhatTheScanReturns` drives `checkWalk` directly. That is worse than the layer-3 gap this arm records open: layer 3 loses 20% of sites; this loses **100% of the walk guard**, and every layer closed in the previous round goes unobserved at once. Now **DEAD**, killed by `TestScanGoEntityKindsIn_RunsTheWalkCheck` **alone** — no live-tree test sees it.

### The through-line, which is the arm's real output

Three rounds, three blocking findings, **one mechanism**:

| round | extraction | what it freed |
|---|---|---|
| B1 | anchors pulled into a var | the anchor set can be emptied silently |
| B2 | the reaction split into a helper | one of its three directions is masked by a neighbour |
| W1 | the decision split into `checkWalk` | **the caller can skip it entirely** |

**Parameterising a decision so a control can drive it is exactly what frees the caller to ignore it.** Each extraction needs a control one level up, or it trades a graded no-op for an ungraded one. That is now written into the file's comments rather than only the commit message, because it is the reusable part.

### What is deliberately NOT closed

- **Layer 3 is open and MEASURED open.** Truncating each file at the last top-level `}` at 85% parses cleanly, keeps `FilesParsed` at 2015, hits all three anchors, and leaves the package green while the scan has silently lost **113 of 566 sites (20%)**. Closing it needs `parseGoTree` to surface per-file sizes — a production change in `internal/entkinds` that pulls `./cmd/grafel/` into scope, correctly out of scope for arm 1. The prose claiming the anchors covered content has been **deleted**, and the mutant, the figure and the required `DirEntry.Info().Size()` fix are recorded in its place.
- **M1 (tempdir root) is still killed only by the live-tree tests**, not by the new control — the control drives `scanGoEntityKindsIn`, so mutating the *wrapper's* root is one seam further out. The author raised this unprompted and chose to state it rather than add another wrapper. **I agree**: it is benign today (four tests fail loudly), and the alternative is chasing the regress indefinitely. Stated beats implied.

### Two claims withdrawn, both mine

1. **"DEAD but with the wrong diagnosis" was materially overstated** — I relayed it onward before it was checked. Measured: the inverted-filter mutant kills **six** tests over ~1085 lines on main, two of them fixture-tree tests printing `collected = map[]` (an unambiguous *"the walk read nothing"* that no ledger edit can silence), and following the misleading advice leaves 57 further failures rather than a green tree. What survives is the real result and is worth having: the mutant **clears both floors** (2662 files > 500, 2888 sites > 200), so a count floor provably cannot see it and anchors can — plus one clear diagnosis instead of 1085 noisy lines. The narrative is re-grounded on that.
2. **Layer 4 needed nothing**, and my brief was wrong to imply doubt. Both directions — recall and over-fire — are measured DEAD on `main`, killed by fixture-tree tests driving the real `ScanGo` with exact expected sets. The author's refusal to add anything was correct.

### Enumeration

19 guard sites across 8 packages; **layer 5 missing at 16 of 19**, so one shared `recordingTB` is the cheapest sweep. Spot-check hit rate **3/4**, and the miss clusters in the *extension-inversion* mutant (an incidental non-Go file in the walk root turns it into an accurate parse error). Arm 2 must **score-then-fix per row**, not batch-apply the template, and should budget ~1 in 4 rows evaporating on contact. That warning is in the enumeration file.

Filed separately from this arm: **#6842** (`home_sweep_guard_6178_test.go` omits `.claude` from its `SkipDir`, so it walks every agent worktree's full checkout — invisible to CI, which has no worktrees).

Issue #6178, opened 2026-08-05
grafel links ignores GRAFEL_HOME for <group>-links.json: internal/cli/links.go passes an empty grafelHome to links.PathsFor


arthurgeron:
I index several repositories with grafel and compare two builds by pointing `GRAFEL_HOME` at a scratch root, which is how I hit this.

`links.PathsFor` takes the grafel home as a parameter and treats the empty string as "use the real one" (`internal/links/links.go:257`):

```go
// PathsFor returns the canonical paths under grafelHome ("" → ~/.grafel)
// for the given group.
func PathsFor(grafelHome, group string) (Paths, error) {
	if group == "" {
		return Paths{}, errors.New("group name required")
	}
	if grafelHome == "" {
		home, err := os.UserHomeDir()
		if err != nil {
			return Paths{}, err
		}
		grafelHome = filepath.Join(home, ".grafel")
	}
```

Nothing in that fallback consults `GRAFEL_HOME`, and both non-test callers that run a real link pass pass the empty string:

- `internal/cli/links.go:201`, `runLinksForGroup`, the `grafel links` command
- `internal/cli/links.go:112`, `RunLinksForGroup`, the hook the rebuild path runs

```go
res, err := links.RunAllPasses(group, graphsDir, "")
```

So a link pass on a group under a scratch `GRAFEL_HOME` puts the store where it was told, since `StoreDir` does go through a home helper that honours the variable, and puts `<group>-links.json` in the real `~/.grafel/groups/` next to the state of whatever groups the user actually has. Nothing warns, and a second isolated run reads the first one's links. The `grafel links` site is in-process, so this reproduces without involving a daemon or its environment.

The parameter is not vestigial and the fix shape is already in the tree: `cmd/grafel/xrepo_verify.go:86` calls the same function with a real home and gets hermetic behaviour from it.

```go
res, err := links.RunAllPasses(group, tmpGraphs, tmpHome)
```

`registry.HomeDir()` (`internal/registry/registry.go:198`) is the resolver those two call sites want, and `internal/daemon/state_path.go:249` is an existing example of a helper that defers to it.

I am happy to send that as a PR if you would rather it came from me.

[...]

PR #6845, opened 2026-09-05
test(#6842): home_sweep_guard walked into .claude/worktrees/ — the guard fails outright in any dev checkout, not just slowly

### LINKED ISSUE #6842: home_sweep_guard_6178_test.go walks into .claude/worktrees/ — the repo-root trap, inside a test

cajasmota:
# fix(#6842): the #6178 home sweep walked into `.claude/worktrees/`, parsing every other branch's checkout

Closes #6842

## The defect

`scanHandRolledHomePaths` (`internal/registry/home_sweep_guard_6178_test.go`) is called with
`root := repoRoot(t)` and its `SkipDir` switch listed
`".git", "node_modules", "vendor", "testdata", "dist", "build"` — **`.claude` was missing**.

In any development checkout with agent worktrees under `.claude/worktrees/`, the guard therefore
parsed every worktree's full copy of the production tree, and would:

1. `t.Fatalf` on a parse error inside an unrelated in-flight branch mid-edit;
2. report offences under `.claude/worktrees/agent-X/...` keys that can never match the
   allow-lists, which are written against real repo paths;
3. do the work N+1 times.

CI never sees it — there are no worktrees there — so this was a local-development-only failure.

Measured side effect of the fix on this machine: `go test ./internal/registry/` went from
**18.3s to 2.6s**.

## The fix

Added `.claude` to the `SkipDir` switch, with prose explaining why it is not an optimisation.

## On sharing one exclusion list — declined, deliberately

The brief asked whether `registry` should share `entkinds`' `skippedDir`. I did not, for three
reasons:

1. **`skippedDir` is unexported production code in `internal/entkinds`.** Sharing means either
   exporting it from a production package purely so a test in another package can call it, or
   creating a new package for a six-element string switch. Both add an import edge that exists
   only to serve a test.
2. **The two lists mean different things.** `entkinds.skippedDir` is "directories the entity-kind
   rule scan refuses"; the registry switch is "directories the home-path guard refuses". They
   happen to agree today. Coupling them means a future `entkinds`-specific addition silently
   narrows a *guard's* scope — which is precisely the #6834 defect class (a guard whose scope
   nothing verifies), reintroduced through the back door.
3. A three-way share is out of scope anyway: `internal/types/` is under an open PR and untouchable.

Instead, the exclusion is now **pinned by a test** rather than by convention, which is the thing
that was actually missing. Divergence is no longer silent for this walker.

## The test

The obvious spelling — "run the real sweep, assert nothing under `.claude` was reported" — is
**vacuous in CI**, where no `.claude` directory exists, so it passes whether or not the exclusion
is present. `TestHandRolledHomeSweepSkipsWorktreeCheckouts` therefore *builds* the condition
under a `t.TempDir()` root. `scanHandRolledHomePaths` already accepted an arbitrary `root`, so no
parameterisation — and hence no "extracted decision the production caller can now skip" — was
needed.

Fixture axes, varied vs held constant:

| File | Axis it varies | Must be |
|---|---|---|
| `internal/cli/hand_rolled.go` | **nested** location, valid Go, genuine offender | reported (positive control) |
| `.claude/worktrees/agent-x/broken.go` | unparseable Go under `.claude` | never parsed |
| `.claude/worktrees/agent-x/internal/registry/shadow.go` | **parseable** offender under `.claude` | never reported |

The positive control is deliberately nested rather than at the temp root — see mutant D below.
The `shadow.go` case exists so the two failure modes are graded independently: the parse error
alone would mask an over-reporting walk that happened to find only valid Go.

## Mutation scoring

All with `go vet` first (a non-compiling mutant proves nothing), `-count=1`, every mutant edit
applied by an exact-count replace that aborts on 0 or >1 matches, and `-v | grep -c '^=== RUN'`
confirming the `-run` filter selected exactly 1 test each time. Baseline restored by `cp` with
verified shasums between every run — never `git checkout`.

|---|---|---|---|
| **A** *(permissive, first)* | remove `.claude` from the switch **and** soften the parse-error `t.Fatalf` to `return nil` — i.e. the walk *reports* the `.claude` file instead of skipping it | **DEAD** | the `.claude/` prefix assertion **and**, independently, the exact-count assertion |
| **B** | remove `.claude` from the switch (the shipped defect) | **DEAD** | `parse .claude/worktrees/agent-x/broken.go: expected ')', found '{'` |
| **C** | mutant B **plus** neutering the fixture — `broken.go` rewritten as valid Go | **DEAD** | the `shadow.go` over-report assertion. The test stays meaningful with a valid-Go fixture, which was the explicit ask. |
| **Cctl** *(control)* | neuter the fixture **alone**, on fixed code | **ALIVE / green**, as required — proves C's kill comes from the missing exclusion, not from the fixture edit |
| **D** | over-skip: `SkipDir` on every directory below the root | **DEAD** | the pre-existing `found no hand-rolled home-path shape at all` binding fatal |

Mutant A was run first and deliberately compounds two edits so the parse-fatal cannot mask the
over-report assertion. Both assertions kill A on their own — they are not mutually masking.

## Something the brief got wrong, and a gap found by scoring

- **The brief said there are three walkers with their own exclusion lists.** There are more, and
  one of them is in the *same package*: `home_isolation_sweep_guard_6735_test.go`'s
  `walkTestFiles` (shared by `scanUnpinnedGrafelHome` and `countTestFiles`) has an identical
  switch about thirty lines away — and it **already lists `.claude`**, with a comment saying why.
  So the divergence was not `registry` vs `entkinds`; it was `registry` vs `registry`. Only the
  #6178 walker was missing it. That also weakens the case for a cross-package share further:
  the nearest correct copy was already in the same file's neighbour.
- **Mutant D initially SURVIVED.** The first version of the test put the positive-control
  offender at the temp root, so a walk that refused to descend anywhere still found it — the
  test pinned "skips `.claude`" while leaving "descends at all" ungraded. The control was moved
  to `internal/cli/hand_rolled.go` and D now dies. This is the "fixtures pin one axis and leave
  the neighbour open" failure, caught only because the mutant was actually run.

Also updated the file's header doc comment, which claimed the scan covers "every non-test .go
file in the repository" — a coverage claim the exclusion changes.

## Verification

- `go test -count=1 ./internal/registry/` — ok (18.3s → 2.6s)
- `go test -count=1 ./internal/entkinds/` — ok
- `gofmt -l` — clean
- `go run ./tools/coverage validate` — exit 0 (822 records; warnings pre-existing and unchanged)

`./cmd/grafel/` was not run: the change touches a single `_test.go` file in `internal/registry`,
nothing under `internal/extractors/`, `internal/resolve/`, or graph assembly.

### DIFF
--- a/internal/registry/home_sweep_guard_6178_test.go
+++ b/internal/registry/home_sweep_guard_6178_test.go
@@ -13,8 +13,10 @@ package registry_test
 // sidecar reader re-derived its path by hand too. Grep-and-patch does not
 // end this — a twelfth sidecar next year re-opens it exactly the same way.
 //
-// This test scans every non-test .go file in the repository for a function
-// whose body contains BOTH (a) a call to os.UserHomeDir() or
+// This test scans every non-test .go file in the repository — excluding the
+// vendored, generated and worktree directories scanHandRolledHomePaths
+// refuses to descend into, notably .claude (#6842) — for a function whose
+// body contains BOTH (a) a call to os.UserHomeDir() or
 // os.Getenv("HOME"/"USERPROFILE"), AND (b) a string literal containing
 // ".grafel" — the two ingredients of a hand-rolled "resolve the grafel
 // home" join. Every hit must be declared in EXACTLY one of the two
@@ -130,6 +132,7 @@ import (
 	"go/parser"
 	"go/token"
 	"io/fs"
+	"os"
 	"path/filepath"
 	"runtime"
 	"sort"
@@ -481,7 +484,11 @@ func scanHandRolledHomePaths(t *testing.T, root string) []homePathOffence {
 		}
 		if d.IsDir() {
 			switch d.Name() {
-			case ".git", "node_modules", "vendor", "testdata", "dist", "build":
+			// .claude holds full worktree checkouts of this same repo; walking it
+			// would parse (and report) other branches' source under paths the
+			// allow-lists can never name, and t.Fatalf on any mid-edit parse
+			// error in an unrelated in-flight branch (#6842).
+			case ".git", ".claude", "node_modules", "vendor", "testdata", "dist", "build":
 				return filepath.SkipDir
 			}
 			return nil
@@ -520,4 +527,69 @@ func repoRoot(t *testing.T) string {
 	}
 	// here = <root>/internal/registry/home_sweep_guard_6178_test.go
 	return filepath.Clean(filepath.Join(filepath.Dir(here), "..", ".."))
+}
+
+// TestHandRolledHomeSweepSkipsWorktreeCheckouts pins that the sweep refuses to
+// descend into .claude, which in a development checkout holds full worktree
+// copies of this same repository (#6842).
+//
+// The obvious spelling of this check — "run the real sweep and assert nothing
+// under .claude was reported" — is vacuous in CI, where no .claude directory
+// exists at all, so it would pass whether or not the exclusion is present.
+// This test therefore BUILDS the condition under a t.TempDir() root:
+//
+//   - internal/cli/hand_rolled.go is a genuine offender, deliberately NESTED
+//     rather than sitting at the root. It is the positive control: the walk
+//     must still descend to it and report it, so neither a "skip everything"
+//     mutant nor a "never descend past the root" one can pass by reporting
+//     nothing. A root-level control would survive both.
+//   - .claude/worktrees/agent-x/broken.go is not parseable Go. The sweep
+//     t.Fatalf's on any parse error, so descending into .claude fails the
+//     test outright — the exact way an unrelated in-flight branch mid-edit
+//     breaks this guard locally.
+//   - .claude/worktrees/agent-x/internal/registry/shadow.go is a perfectly
+//     parseable offender. It catches the weaker failure the parse error
+//     would mask: a walk that descends but happens to find only valid Go
+//     still reports offences under paths the allow-lists can never name.
+//     This assertion is what keeps the test meaningful even if broken.go
+//     were ever repaired into valid Go.
+func TestHandRolledHomeSweepSkipsWorktreeCheckouts(t *testing.T) {
+	root := t.TempDir()
+	write := func(rel, src string) {
+		t.Helper()
+		p := filepath.Join(root, filepath.FromSlash(rel))
+		if err := os.MkdirAll(filepath.Dir(p), 0o750); err != nil {
+			t.Fatalf("mkdir for %s: %v", rel, err)
+		}
+		if err := os.WriteFile(p, []byte(src), 0o600); err != nil {
+			t.Fatalf("write %s: %v", rel, err)
+		}
+	}
+
+	offender := `package p
+
+import "os"
+
+func handRolled() string {
+	h, _ := os.UserHomeDir()
+	return h + "/.grafel/groups/x-links.json"
+}
+`
+	write("internal/cli/hand_rolled.go", offender)
+	write(".claude/worktrees/agent-x/internal/registry/shadow.go", offender)
+	write(".claude/worktrees/agent-x/broken.go", "package p\n\nfunc broken( { this is not Go\n")
+
+	got := scanHandRolledHomePaths(t, root)
+
+	for _, o := range got {
+		if strings.HasPrefix(o.file, ".claude/") {
+			t.Errorf("sweep reported %s — it descended into .claude, which holds full worktree "+
+				"checkouts of this repo; offences there can never match the allow-lists (#6842)", o.String())
+		}
+	}
+	if len(got) != 1 || got[0].file != "internal/cli/hand_rolled.go" || got[0].fn != "handRolled" {
+		t.Fatalf("sweep over the synthetic tree reported %v, want exactly "+
+			"[internal/cli/hand_rolled.go:handRolled] — the nested offender is the positive control "+
+			"that the walk still descends into real source directories", got)
+	}
 }


cajasmota:
## Coordinator verification — APPROVE, with one correction the PR body should absorb

**My killing mutant, which nobody else ran:** `case ".git", ".claude", …` → `".Claude"` — a wrong-case spelling. **DEAD**, on `TestHandRolledHomeSweepSkipsWorktreeCheckouts` alone. That was the reviewer's explicitly open question (*"the one thing that would have caught a wrong `.claude` spelling"*), and it is now answered: the fix is pinned to the exact spelling, not merely to something claude-shaped.

## The review found the defect is worse than this PR claims

It built the condition independently (no worktree has a `.claude` of its own — `repoRoot(t)` resolves via `runtime.Caller` to the *worktree* root) by copying `internal/` + `cmd/` into a fake `.claude/worktrees/`:

- **main**: `TestNoHandRolledGrafelHomePaths` **FAILS**, reporting 20 bogus offences under paths the allow-lists can never name.
- **branch**: same tree, **PASS**, 9.67s → 1.17s on that test.

So in the real checkout (five live worktrees) this guard is **currently failing**, not just slow. That is a much stronger justification than the PR body gives, and it should lead.

**Correspondingly, the `18.3s → 2.6s` figure is not reproducible as stated** and should be restated or dropped. In a clean worktree the review measured main 3.35s vs branch 3.03s — no real difference, because there is nothing to skip. The 18.3s looks like a **cold** (compile-inclusive) run compared against a warm one. The magnitude is credible with the mechanism actually present (9.67s → 1.17s for one test, ~5× plausible across five worktrees), but the comparison as written is cold-vs-warm.

## The unfixed clone — I confirmed it myself

**`internal/atomicfile/guard_6018_test.go:435`** is a byte-for-byte clone of this walk — same `WalkDir`, same `repoRoot(t)`, same parse-`t.Fatalf` — and its switch reads:

```go
case ".git", "node_modules", "vendor", "testdata", "dist", "build":
```

No `.claude`. The reviewer proved it live by planting three lines of invalid Go under a fake worktree and getting `--- FAIL: TestNoDeterministicTempNames`. **Filing separately** — it is the same bug in a different package and does not belong in this PR.

That makes **seven** `repoRoot(t)`-rooted walks with their own exclusion list, across `atomicfile`, `types` (×2), `verify`, `entkinds`, `relkinds`, `registry` (×2) — not the three I briefed.

## On the declined share, the reviewer disagrees with the author and I side with the reviewer

The author declined to share the exclusion list, arguing that coupling a guard's scope to `entkinds`' unexported production list would let a future change silently narrow the guard. The reviewer's counter is stronger: **that risk is hypothetical, while divergence is demonstrated and currently live** — five copies, one still wrong. And mutants **I** (`Contains(name,"claude")`) and **J3** (widening with `engine, graph, extractors, security`) are both **ALIVE on the whole package**, so *nothing detects narrowing in either design* — which means the coupling argument does not discriminate between them.

**Still not blocking this PR.** The author's scope discipline is right; the fix plus its pin is a complete unit. The share belongs with the `atomicfile` fix, and I am filing them together.

## One clause grades nothing, and that is fine

The `.claude/` prefix `Errorf` loop **never fires alone** — every mutant that trips it leaves two entries in `got`, so the exact-count `Fatalf` trips too, and deleting the loop kills nothing extra. It is **subsumed** by the count assertion rather than masking it: the surviving assertion is the stronger one and the loop buys a better failure message. Not a defect, but the doc block should not read as though it were an independent check.

## Corrections to my own framing

- I said "three walkers, one per package". It is **seven call sites**, and the nearest divergence was **two switches ~30 lines apart in the same package**, one already correct.
- My proposed attack — *"the `.claude/` prefix assertion satisfied by a path merely containing `.claude`"* — is **not reachable**. The switch matches `d.Name()` exactly, non-`.go` files are never parsed, and the co-located exact-count assertion pins the expected file.
- The author's "mutant D left 'descends at all' ungraded" is true of the **new test in isolation** only; D was always killed by the pre-existing `len(out)==0` fatal. The fix (nesting the control) is still correct and makes the new test self-sufficient — but no mutant was ever globally ALIVE.
- Worth knowing: the `.claude/` prefix check is not Windows-vacuous only because of a `filepath.ToSlash` twenty lines away.

Issue #6518, opened 2026-08-21
proto CONTAINS is anchored on a file path, and #6367's hand-off to #6459 is stale — that issue shipped without touching it


cajasmota:
Split out of #6367, because that thread hands this arm to **#6459** and **#6459 shipped without touching it**.

#6459 is in `main` at `de1d96481`. It fixed the proto service *tier* — an ordered-tier lookup in the resolver — which is a different concern entirely from where a CONTAINS edge is anchored. Anyone reading #6367 would reasonably conclude this was handled. It was not, and it had no owner.

## The defect

`internal/extractors/proto/proto.go:252-257` anchors the relationship's `FromID` on a **file path** rather than on the owning entity.

The consequence is not a dangling edge that something reports. `refs.go`'s `sourceFileExtensions` (around `:3196`) does **not** include `.proto`, so the resolver does not recognise the `FromID` as a file path at all — it is classified as `DispositionBugExtractor` and lands in a bucket that reads as a generic extractor bug rather than as a misanchored edge. The signal is there, but it points at the wrong thing.

## Why it is worth fixing rather than folding back into #6367

The three remaining #6367 arms (swift, bicep+dockerfile, graphql) each need their own observation built, and swift additionally needs seven `to_bare_name` rows repaired before its gate can move. Bundling proto behind that sequencing is what let it go unowned in the first place.

## What a fix needs

1. Anchor `FromID` on the owning entity, matching the pattern the shipped hcl fix established (`0e31e57ee`).
2. **A regression test on the `TestHCL_ContainsAndDependsOnAnchoredOnOwner` model.** There is no proto fixture that would catch this — note #6453 separately records that no proto golden fixture exists at all, which is why three previously-fixed proto defects can regress silently.
3. Decide whether `.proto` belongs in `sourceFileExtensions`. Adding it would move this class from `DispositionBugExtractor` into the file-anchored bucket where it can be counted — **but that is a reporting change with its own blast radius** and should be a deliberate decision, not a side effect of this fix.

**Killing mutant:** restore the file-path anchor. The new test must go red. If only the disposition bucket changes and no test fails, the observation has not been built.

Refs #6367, #6453

Issue #6118, opened 2026-08-04
Python custom extractor omits EndLine; C# file components unpositioned when the class shadow-fold starves foldFileComponentDuplicates


cajasmota:
With `GRAFEL_INPROC_CUSTOM_EXTRACTORS` enabled, **not one of the 61,131 entities the custom extractors add carries a complete source span.**

Measured on a ~427k-entity corpus, both graphs read back via `graph.LoadGraphFromDir`:

| check | flag OFF | flag ON | delta |
|---|---|---|---|
| `StartLine <= 0` | 87,532 | 87,745 | +213 |
| `EndLine == 0` (with `StartLine > 0`) | 339,753 | 398,595 | **+58,842** |

`213 + 58,842 = 59,055` — **exactly the net entity delta**. Every added entity is missing either its start or its end position. Zero have both.

## Why this blocks default-on

The base path already has a 79.5% `EndLine == 0` rate, so incomplete spans are a pre-existing problem rather than a new class. But at **100%** it is categorical rather than partial: any consumer that slices source by extent — `grafel_get_source`, docgen, the dashboard's code panels, anything computing entity size — gets a zero-length or unanchored region for every single custom-extractor entity.

That turns a +13.8% entity gain into 59k nodes that cannot be located in their own source file.

It is also systemic rather than per-language: C#, Java, TypeScript, Go, Python, JavaScript and Rust all contribute added entities, and the shortfall accounts for all of them. So this is likely a property of how `internal/custom/**` extractors populate their records — or of how `MergeWithCustom` / the record→entity conversion carries positions through — rather than 340 independent bugs.

## A second, smaller regression in the same area

**187 C# file-level component entities regress from a real start line to line 0** — verified directionally (11, 12, 13, 14, 15 → 0). This is the entire `StartLine <= 0` delta.

This one is worth separating because it is the **only** hygiene metric where the flag makes an existing value *worse* rather than merely diluting the average. An entity that had a correct position loses it.

## Where to look first

The likely candidates, in order:
1. Whether `internal/custom/**` extractors populate `EndLine` at all on the records they emit — if most do not, this is a systematic gap in the extractor base rather than in the merge.
2. Whether the record→entity conversion carries `EndLine` through. Note `convertExtractedRecords` was recently extracted in #6094 and the analogous `FromID` omission there caused a different silent defect, so this conversion has form.
[...]

PR #6848, opened 2026-09-05
fix(#6815): erlang, nim and groovy emitted IMPORTS edges whose FROM end was a raw path no node carried

### LINKED ISSUE #6815: erlang, nim and groovy emit path-anchored IMPORTS with no FileEntity carrier — the edges resolve from nothing and no fixture asserts one
erlang, nim and groovy emit path-anchored `IMPORTS` with no `extractor.FileEntity` carrier — so those edges resolve from nothing, and no fixture asserts one

## The gap

Three extractors emit `IMPORTS` relationships whose `from` end is a **path**, without emitting an `extractor.FileEntity` for that path:

| language | site |
|---|---|
| erlang | `internal/extractors/erlang/extractor.go:424`, `:460` |
| nim | `internal/extractors/nim/nim.go:358` |
| groovy | `internal/extractors/groovy/groovy.go:1180` |

With no carrier entity, the edge's source end has nothing to resolve to.

`#6518` already did the corresponding fix for proto by emitting a `FileEntity`, so there is a worked precedent in-tree for what this should look like.

## Why it went unnoticed

The guard that would normally report this — `knownInvisibleOffenders` — **is now empty (`{}`)**, and every surviving `allowedFileAnchored` entry has a resolving carrier. So the mechanism that catches file-anchored edges with no carrier currently reports a clean board.

It does not cover `IMPORTS`, by the #120 convention. That exclusion is why these three have stayed invisible: the population moved into the one relationship kind the guard does not look at.

**No fixture asserts a single `IMPORTS` row for any of the three.** `erlang-otp-mini` has one at `internal/quality/golden/erlang-otp-mini/expected.json:112-115`, but it is `nice_to_have`, so it fails silently — and separately it could not match anyway, because the expected-edge resolver cannot match an edge whose `from` end resolves to no entity (that half is #6488 arm C).

## A false note to correct while here

That same `erlang-otp-mini` note claims erlang is **unique** in lacking a file carrier. It is not — nim and groovy are in exactly the same position. The note should name all three or stop making the claim.

## Two fixes, deliberately kept apart

- **This issue (the graph):** emit `FileEntity` for the three languages so the edges have a real source end. The existing `nice_to_have` row would then resolve on its own.
- **#6488 arm C (the assertion):** make `from_bare_name` actually work — today it exists only in prose, because every match path in `resolveExpectedEdge` (`internal/quality/diff.go:537-638`) is nested inside `for _, fc := range fromCands`, so an empty `fromCands` can never match by any route.

They are complementary, not alternatives. They are filed separately because they touch different subsystems with different blast radii, and bundling them would put an extractor change and a fixture-format change in one review.

## Not established

- **Predicted, not measured.** The claim that deleting erlang's `-include` IMPORTS record produces zero fixture delta was reasoned from the source, not run. Whoever picks this up should measure it first rather than inheriting the prediction.
- **No corpus count.** How many real `IMPORTS` edges across the three languages currently dangle is unknown. A zero measurement here would be corpus-relative and would not mean the shape is rare — but nobody has looked, and the number should inform how urgent this is.
- Whether emitting the carrier moves any golden fixture's entity counts. It plausibly does — new entities are new rows — so the three exact-count constants and affected baselines need deriving from disk, not incrementing.

Grounded read-only at `2ed41a989`; the file:line references above were read directly.

cajasmota:
fix(#6815): erlang, nim and groovy emit path-anchored IMPORTS with no FileEntity carrier — the edge's FROM end resolved to nothing

Closes #6815

## What was wrong

`erlang`, `nim` and `groovy` each emit `IMPORTS` relationships whose `FromID` is the source
file's path, and none of the three emitted an `extractor.FileEntity` (the #577 per-file
`SCOPE.Component(subtype="file")`) carrying that path. `internal/resolve/refs.go` has no
path→entity index: a path-valued `FromID` resolves **if and only if** some emitted node
carries that exact string as its `Name`. Nothing did, so `ReferencesEmbeddedWithAllowlist`
had nothing to rewrite the `FromID` onto and the raw path reached the graph as the FROM end.

This is an **entity-side** defect, not a relationship-side one — the edges were correct, the
node they name was simply never emitted. That distinction sets the fix: emit the missing
carrier, do not re-anchor the edges.

## Grounding — three things in the issue are now WRONG

Ground-truth read at `a1900a432`, from the code, not from the issue body.

1. **All three languages still exhibit it.** Verified by enumerating `extractor.FileEntity`
   call sites under `internal/extractors/` — 25 packages call it; erlang, nim and groovy are
   absent from that list. Confirmed at runtime by the RED test runs recorded below.

2. **STALE — "the erlang golden row is `nice_to_have`, so it fails silently."** It is not.
   `internal/quality/golden/erlang-otp-mini/expected.json` already carried
   `"must_exist": true` on that row. #6488 arm C landed between the issue being filed and
   this work.

3. **STALE — "`from_bare_name` exists only in prose … an empty `fromCands` can never match
   by any route."** `from_bare_name` is fully implemented (`internal/quality/diff.go:591`,
   plus nine tests in `internal/quality/from_bare_name_6488_test.go`), and the erlang row was
   matching through it. This mattered directly: it is *why* the fix moves recall backwards
   unless the fixture row is moved with it (see the measurement below).

4. **ALREADY FIXED — "the note claims erlang is unique in lacking a file carrier."** The note
   already named nim and groovy and already pointed at #6815. Nothing to correct there.

5. **MEASURED, replacing a prediction.** The issue flagged "deleting erlang's `-include`
   IMPORTS record produces zero fixture delta" as reasoned-not-run. It is now moot in the
   direction that matters: with the carrier in place the edge is no longer inert — deleting
   the carrier drops `erlang-otp-mini` from 26/26 to 25/26 must-have relationships.

## The fix

One shared mechanism, three call sites — `internal/extractor/file_carrier.go`:

```go
func FileCarrierFor(path, lang string, records []types.EntityRecord) (types.EntityRecord, bool)
func PrependFileCarrier(path, lang string, records []types.EntityRecord) []types.EntityRecord
```

The carrier is emitted **conditionally**, following the proto precedent (#6518), not by
calling `FileEntity` unconditionally as ~25 other extractors do. These three languages have
no other file-anchored edge, so an unconditional carrier would mint one bare orphan node per
source file across a whole repo — a change no recall-shaped assertion can see, because every
such assertion only asks whether the carrier *exists*.

Three guard clauses, each of which rejects on its own and each graded by a mutant only
that clause's fixture kills (see the table):

1. `path` is non-empty — a nameless carrier resolves nothing;
2. some relationship in `records` has `FromID == path` — the exact resolution requirement;
3. no record is *already* named `path` — two nodes under one id would make the rewrite
   target ambiguous. Checked for **every** record, before the loop may short-circuit on
   clause 2; that ordering is load-bearing and is pinned by M12.

The carrier owns no relationships: the per-import stub records still carry the `IMPORTS`
edges, so re-homing them would double every edge.

## Mutants — scored individually, `go vet` first, `-count=1`, every edit proven to land

Every mutant was applied by an exact-count replace that aborts on 0 or >1 matches and
re-reads the file to prove the anchor is gone (`scratchpad/mutate.py`), then restored by
`cp` from a byte backup with verified shasums. Never `git checkout --`.

| # | Direction | Mutation | Verdict | Killed by |
|---|---|---|---|---|
| M1 | PERMISSIVE | drop clause 1 — carrier emitted unconditionally | **DEAD** | `TestFileCarrierFor_NoCarrierWhenNoEdgeIsPathAnchored_6815`, `TestFileCarrierFor_AnotherFilesAnchorDoesNotCount_6815`, `TestPrependFileCarrier_NoOpWhenNothingIsDue_6815`, `TestErlang_/TestNim_/TestGroovy_NoCarrierWithoutAnythingToCarry_6815` — **and at fixture level**: `forbidden_entity_hits` 0 → 3 (erlang), 0 → 2 (nim) |
| M2 | PERMISSIVE | drop the clause entirely — allow a second carrier when one is already named after the path | **DEAD** | RE-SCORED after M12's fixture landed: **both** `…NoSecondCarrierWhenOneIsAlreadyNamedAfterThePath_6815` and `…NoSecondCarrierWhenThePathNamedRecordComesLast_6815` fail. Deleting the clause kills both; merely reordering it (M12) kills only the second — so the two cases are not one duplicated, and the clause is graded ALONE in both |
| M3 | PERMISSIVE | `FromID != ""` instead of `FromID == path` | **DEAD** | `TestFileCarrierFor_AnotherFilesAnchorDoesNotCount_6815` |
| M4 | RESTRICTIVE | delete the erlang call site (the pre-fix state) | **DEAD** | 5 erlang tests; nim and groovy stay GREEN |
| M5 | RESTRICTIVE | delete the nim call site | **DEAD** | 5 nim tests; erlang and groovy stay GREEN |
| M6 | RESTRICTIVE | delete the groovy call site | **DEAD** | 5 groovy tests; erlang and nim stay GREEN |
| M7 | RESTRICTIVE | anchor only on `Kind == "IMPORTS"` | **DEAD** | `TestFileCarrierFor_AnchorIsAboutFromIDNotKind_6815` |
| M8 | ORDER | append the carrier last instead of first | **DEAD** | `TestPrependFileCarrier_CarrierIsFirstAndOrderIsKept_6815` |
| M9 | PERMISSIVE | remove the empty-path guard | **DEAD** | `TestFileCarrierFor_EmptyPathNeverCarries_6815` |
| M10 | — | pass `""` as the language token at erlang's call site | **ALIVE — equivalent under the suite, with the reason** | Not a defect and not manufacturable into a kill: all three callers run `extractor.TagEntitiesLanguage` after `PrependFileCarrier`, and that helper fills an *empty* `Language` with the extractor's own token. The parameter is therefore observably redundant *for these callers* and load-bearing for one that does not tag. Recorded in the doc block rather than asserted. |
| M11 | — | pass a WRONG language token (`"beam"`) at erlang's call site | **DEAD** | `TestErlang_CarrierIsLanguageTagged_6815` — the pair M10/M11 pins exactly which half of the parameter the suite grades |
| M12 | ORDERING (**found in review**) | hoist `if anchored { continue }` **above** the clause-3 `Name == path` test | **DEAD** *(was ALIVE at review time)* | `TestFileCarrierFor_NoSecondCarrierWhenThePathNamedRecordComesLast_6815`, added for it. The pre-existing clause-3 case stays GREEN under this mutant — that asymmetry is the evidence the two cases grade different things rather than one twice |

M4/M5/M6 killing only their own language is also the evidence that the nim and groovy
fixtures were RED before the fix: those two packages' tests were written first but the erlang
call site landed in the same commit, so a same-commit red run is not otherwise on record.

## Fixtures — the axis each varies and what it holds constant

Every fixture body was read against its own name.

**`internal/extractor/file_carrier_6815_test.go`** (8 cases, unit-level guard grading)

| Test | VARIES | HOLDS CONSTANT |
|---|---|---|
| `EmitsWhenAnEdgeIsPathAnchored` | — (baseline positive) | one record, one path-anchored edge, no record named after the path |
| `NoCarrierWhenNoEdgeIsPathAnchored` | the edge's `FromID` (empty) | the record, the kind, and **clause 2 satisfied** — so clause 1 alone rejects |
| `AnotherFilesAnchorDoesNotCount` | *which* path the edge anchors on | the edge stays path-anchored, so "non-empty FromID" would still pass |
| `NoSecondCarrierWhenOneIsAlreadyNamedAfterThePath` | an existing record named after the path | the path-anchored edge is present — so clause 2 alone rejects |
| `AnchorIsAboutFromIDNotKind` | the relationship `Kind` (CONTAINS) | the anchoring `FromID` |
| `EmptyPathNeverCarries` | the path (empty) | the records |
| `PrependFileCarrier_CarrierIsFirstAndOrderIsKept` | position | contents |
| `PrependFileCarrier_NoOpWhenNothingIsDue` | nothing due | contents |

**`internal/extractors/erlang/file_carrier_6815_test.go`** (6 cases)

| Test | VARIES | HOLDS CONSTANT |
|---|---|---|
| `IncludeGetsAFileCarrier` | the `-include` attribute | no `-import`, one module, one function |
| `FunctionImportGetsAFileCarrier` | the import KIND (`-import(Mod,[f/1])`) | no `-include` — a fix keyed on `includeRE` alone passes the row above and fails this one |
| `NoCarrierWithoutAnythingToCarry` | imports absent | module, export and function body, byte for byte identical to the first case |
| `OneCarrierPerFileNotPerImport` | the NUMBER of anchored edges (3) | one file — catches a carrier appended inside the import loop |
| `CarrierCarriesNoRelationshipsOfItsOwn` | — | asserts the edge is still emitted exactly once, so a re-homed (doubled) edge fails |
| `CarrierIsLanguageTagged` | — | `Language` and `SourceFile` |

**`internal/extractors/nim/file_carrier_6815_test.go`** (6 cases) — same shape; the second
case varies the import FORM (`include` vs `import`, two of the three syntaxes
`collectImports` feeds into one producer) and the count case uses 4 edges from
`import a, b, c` + `from os import getEnv`. `CarrierIsSubtypeFileNotImport` pins
`subtype == "file"`, not `"import"`: #6481 keys the import-placeholder marker on
`SCOPE.Component` + `subtype=="import"`, and a carrier stamped that way would be excluded
from the very by-name index it exists to populate.

**`internal/extractors/groovy/file_carrier_6815_test.go`** (6 cases) — the second case varies
the import FORM to `import static java.lang.Math.*`, which `buildImportRecord` routes down a
different branch with a different Properties set and a different `Name`.

## Golden fixtures — measured, not predicted

`grafel quality` was run with a binary built from this branch. `entity_extracted_total` and
`relationship_extracted_total` are gated one-directionally by `scripts/quality/ratchet.py`
(growth is fatal), so every figure below was re-derived from disk.

| fixture | entity_found/expected | rel_found/expected | ent_total | rel_total |
|---|---|---|---|---|
| erlang-otp-mini | 20/20 → **21/21** | 26/26 (unchanged) | 22 → **23** | 53 → **54** |
| nim-objects-mini | 21/21 → **22/22** | 11/11 (unchanged) | 22 → **23** | 40 → **41** |
| groovy-grails-mini | 13/16 → **14/17** | 15/18 (unchanged) | 23 → **24** | 53 → **54** |

**Enumerated delta, per fixture.** +1 entity is the carrier itself. The +1 relationship is a
NET figure, not a single addition: the module pass synthesises one new
`Module -[CONTAINS]-> carrier` edge (the CONTAINS source is the **Module** node — `kind:
"Module"`, name `erlang-otp-mini` — not the repo), and the pre-existing IMPORTS edge is
REWRITTEN rather than added, its FromID moving from the raw path onto the carrier's
stamped id. One added plus one rewritten. Read out of the kept `graph.json`, not inferred:

```
ENTITY a21736596f013e33 SCOPE.Component file cache_server.erl
  EDGE erlang-otp-mini -[ CONTAINS ]-> cache_server.erl
  EDGE cache_server.erl -[ IMPORTS ]-> cache.hrl     <- pre-existing, now anchored on an entity
```

Only one file per fixture acquires a carrier: `cache_server.erl` is the only `.erl` with an
`-include`, `store.nim` the only `.nim` with an `import`, and `PostController.groovy` is the
fixture's only file.

**`erlang-otp-mini/expected.json` — the row had to move, and this was measured.** With the
carrier in place and the row still spelled `from_bare_name`, the fixture reported
**25/26** must-have relationships: the bare string no longer matches because there is no
longer a raw path at the FROM end. Both the must_exist row and its forbidden over-firing
counterpart are now `from_name` + `from_kind` + `from_file`, exactly as that row's own note
said they must be when #6815 landed. The forbidden row stays a live control rather than a
tautology: if the extractor mis-anchored `cache_server.erl`'s include onto `cache_sup.erl`,
that path *would* acquire a carrier and the row would fire.

**New over-firing rows at fixture level** (`forbidden_entities`), proven non-vacuous by M1:
`cache_sup.erl` and `cache_app.erl` (no include, no import), plus `cache.hrl` — the include
*target*, held because a fix that minted a carrier per import ENDPOINT rather than per
anchoring FILE would pass both other rows and fail this one. nim adds `errors.nim` and
`model.nim`. groovy-grails-mini is a single file, so it can only grade the positive
direction; its negative direction is graded by the unit tests and by the two multi-file
fixtures.

`internal/quality/subtype_assertion_6488_test.go`'s corpus control ledger is updated for the
three new subtype-asserting rows. The subtype is load-bearing there, not decorative: the
carrier's `Kind` (`SCOPE.Component`) is shared with every import stub and every erlang module
entity, so a row without `subtype: "file"` could be satisfied by an unrelated same-named
record.

## `cmd/grafel` digest

**Did not move.** `go test -count=1 ./cmd/grafel/` is GREEN. The #6118 fixture
(`writeSpanFixture6118`) contains Python, Java, JS/TS and C# sources and **no** `.erl`,
`.nim` or `.groovy` file, so this change cannot reach it. That is a statement about the
fixture's membership, verified by enumerating it — not an assumption that a property-only
change is invisible.

## Verification

- `go test -count=1` GREEN on `./internal/extractor/`, `./internal/extractors/...`,
  `./internal/quality/...`, `./cmd/grafel/`
- `gofmt -l` clean over `internal/extractor` and `internal/extractors`
- `go run ./tools/coverage validate` — `ok: 822 record(s) … 1163 test(s) checked`

## Review round — three findings, all addressed

1. **Clause 3 was order-dependent and nothing caught it** (coordinator's mutant, ALIVE at
   review). Added `TestFileCarrierFor_NoSecondCarrierWhenThePathNamedRecordComesLast_6815`:
   same two records as the existing clause-3 case, with the path-named one placed *after* the
   anchoring one. M12 now DIES; M2 re-scored and still kills both cases. No current caller
   emits in that order — stated in the test, and the reason to pin the property rather than
   rely on it.
2. **A fixture label claimed more than its body.** nim's over-firing control said "the same
   proc, byte for byte, as the first case" while the positive used
   `result = name.toUpperAscii()` and the control `result = name`. All three nim sources now
   share one proc body, and the claim is proven mechanically, not by eye: the control is the
   positive with `import strutils\n\n` deleted and nothing else. erlang's and groovy's pairs
   were already byte-for-byte. M1 and M5 re-scored after the change — both still DEAD, so the
   ablation did not go vacuous.
3. **The doc block said two clauses; there are three.** `path == ""` is clause 1 and is
   separately graded by M9. The block now enumerates all three, names the pinning test for
   each, and states the ordering property M12 grades.

## Scope — adjacent findings NOT fixed here

- `knownInvisibleOffenders` in `internal/extractors/file_anchored_rels_guard_test.go` is `{}`
  and the guard skips `IMPORTS` by the #120 convention, so it structurally cannot report this
  class. Whether it *should* grow an IMPORTS-carrier arm now that a carrier is the expected
  shape for all three is a separate question with its own blast radius.
- No corpus-wide count of dangling `IMPORTS` edges across the three languages was taken; the
  issue asked for one and it remains open. A zero there would be corpus-relative anyway.

### DIFF
--- a/internal/extractor/file_carrier.go
+++ b/internal/extractor/file_carrier.go
@@ -1,0 +1,91 @@
+// file_carrier.go — the conditional form of FileEntity (#6815).
+//
+// FileEntity (extractor.go) is the per-file SCOPE.Component(subtype="file")
+// that #577 introduced so a path-anchored FromID has something to resolve to.
+// Most extractors call it unconditionally at the top of Extract. Three did not
+// call it at all — erlang, nim and groovy each emit IMPORTS edges whose FromID
+// is the source path, with no record carrying that path, so
+// ReferencesEmbeddedWithAllowlist had nothing to rewrite the FromID onto and
+// the edge reached the graph with a raw path at its FROM end.
+//
+// The fix is deliberately NOT "call FileEntity unconditionally". These three
+// languages have no other file-anchored edge, so an unconditional carrier would
+// mint one bare orphan node per source file across an entire repo — invisible
+// to every recall-shaped assertion, which only ever asks whether the carrier
+// EXISTS. proto took the same decision in #6518: the carrier is emitted when
+// the file has something for it to carry, and not otherwise.
+
+package extractor
+
+import "github.com/cajasmota/grafel/internal/types"
+
+// FileCarrierFor returns the FileEntity that gives records' path-anchored
+// relationships a real FROM end, and ok=false when no such entity should be
+// emitted.
+//
+// The carrier is emitted when BOTH clauses hold. Each rejects on its own:
+//
+//  1. Some relationship in records has FromID == path. This is the exact
+//     resolution requirement — refs.go has no path→entity index, so a
+//     path-valued FromID resolves if and only if some emitted node carries that
+//     exact string as its Name. A file with entities but no path-anchored edge
+//     needs no carrier and gets none; that clause alone rejects, e.g., an
+//     erlang module that declares functions and imports nothing.
+//  2. No record in records is ALREADY named path. That clause alone rejects a
+//     file which does have a path-anchored edge but whose extractor already
+//     minted a path-named container for it — emitting a second one would put
+//     two nodes under one id and make the rewrite target ambiguous.
+//
+// lang is stamped explicitly rather than taken from FileInput.Language so the
+// carrier cannot become the one record in an extraction that disagrees with the
+// classifier token every other record carries (proto's #6356 trap).
+//
+// MEASURED grading status of that parameter, stated rather than implied: a
+// WRONG token is caught (mutating erlang's "erlang" to "beam" fails
+// TestErlang_CarrierIsLanguageTagged_6815), but an EMPTY one is not — all three
+// current callers run extractor.TagEntitiesLanguage afterwards, and that helper
+// fills an empty Language with the extractor's own token, so passing "" is
+// equivalent under the suite. It is not equivalent for a caller that does not
+// tag, which is why the parameter exists and is passed.
+//
+// The returned record owns no relationships. Callers that DO have file-scoped
+// edges to re-home (proto's file-level CONTAINS) assign them afterwards; for
+// erlang, nim and groovy the per-import stub records still carry the IMPORTS
+// edges themselves, so hanging them off the carrier as well would double them.
+func FileCarrierFor(path, lang string, records []types.EntityRecord) (types.EntityRecord, bool) {
+	if path == "" {
+		return types.EntityRecord{}, false
+	}
+	anchored := false
+	for i := range records {
+		if records[i].Name == path {
+			return types.EntityRecord{}, false
+		}
+		if anchored {
+			continue
+		}
+		for j := range records[i].Relationships {
+			if records[i].Relationships[j].FromID == path {
+				anchored = true
+				break
+			}
+		}
+	}
+	if !anchored {
+		return types.EntityRecord{}, false
+	}
+	return FileEntity(FileInput{Path: path, Language: lang}), true
+}
+
+// PrependFileCarrier returns records with FileCarrierFor's carrier at index 0,
+// or records unchanged when no carrier is due. Index 0 matches the #577
+// convention that the file entity is the first record an extractor appends,
+// which python/re_exports.go and python/prune_import_placeholders.go both rely
+// on.
+func PrependFileCarrier(path, lang string, records []types.EntityRecord) []types.EntityRecord {
+	carrier, ok := FileCarrierFor(path, lang, records)
+	if !ok {
+		return records
+	}
+	return append([]types.EntityRecord{carrier}, records...)
+}
--- a/internal/extractor/file_carrier.go
+++ b/internal/extractor/file_carrier.go
@@ -23,18 +23,38 @@ import "github.com/cajasmota/grafel/internal/types"
 // relationships a real FROM end, and ok=false when no such entity should be
 // emitted.
 //
-// The carrier is emitted when BOTH clauses hold. Each rejects on its own:
+// The carrier is emitted when ALL THREE clauses hold. Each rejects on its own,
+// and each is graded by a mutant that only that clause's fixture kills:
 //
-//  1. Some relationship in records has FromID == path. This is the exact
+//  1. path is non-empty. A nameless carrier is not a carrier — it would resolve
+//     nothing and add a blank node. That clause alone rejects records that DO
+//     carry an anchoring relationship (an empty FromID trivially equals an empty
+//     path), which is why it is tested separately and not folded into clause 2.
+//     Pinned by TestFileCarrierFor_EmptyPathNeverCarries_6815.
+//  2. Some relationship in records has FromID == path. This is the exact
 //     resolution requirement — refs.go has no path→entity index, so a
 //     path-valued FromID resolves if and only if some emitted node carries that
 //     exact string as its Name. A file with entities but no path-anchored edge
 //     needs no carrier and gets none; that clause alone rejects, e.g., an
-//     erlang module that declares functions and imports nothing.
-//  2. No record in records is ALREADY named path. That clause alone rejects a
+//     erlang module that declares functions and imports nothing. Pinned by
+//     TestFileCarrierFor_NoCarrierWhenNoEdgeIsPathAnchored_6815 and, for the
+//     narrower "some OTHER file's path is not this file's anchor" reading, by
+//     TestFileCarrierFor_AnotherFilesAnchorDoesNotCount_6815.
+//  3. No record in records is ALREADY named path. That clause alone rejects a
 //     file which does have a path-anchored edge but whose extractor already
 //     minted a path-named container for it — emitting a second one would put
 //     two nodes under one id and make the rewrite target ambiguous.
+//
+// Clause 3 is checked for EVERY record, before the loop may short-circuit on
+// clause 2 being satisfied — deliberately, and the order is load-bearing.
+// Hoisting the `if anchored { continue }` above the Name test would stop the
+// scan as soon as anchoring is established, so a path-named record emitted
+// AFTER the anchoring one would go unseen and a second carrier would be minted.
+// That reordering is a mutant the fixture set kills:
+// TestFileCarrierFor_NoSecondCarrierWhenThePathNamedRecordComesLast_6815 exists
+// for it specifically, because the ordinary clause-3 case places the path-named
+// record first and cannot see it. No current caller emits in that order — which
+// is the reason to pin the property rather than to rely on it.
 //
 // lang is stamped explicitly rather than taken from FileInput.Language so the
 // carrier cannot become the one record in an extraction that disagrees with the
--- a/internal/extractor/file_carrier_6815_test.go
+++ b/internal/extractor/file_carrier_6815_test.go
@@ -84,6 +84,37 @@ func TestFileCarrierFor_NoSecondCarrierWhenOneIsAlreadyNamedAfterThePath_6815(t 
 	}
 }
 
+// CLAUSE 2 ALONE rejects, REORDERED. Axis VARIED: the POSITION of the
+// path-named record — after the anchoring record instead of before it. HELD
+// CONSTANT: both records, their contents, and the verdict, which must not
+// depend on emission order.
+//
+// This is a separate case rather than a table row on the one above because the
+// two grade different mutants. The case above is satisfied by an
+// implementation that stops scanning for a path-named record as soon as
+// anchoring is established (hoisting `if anchored { continue }` above the Name
+// check): the path-named record is seen first there, so the early exit is never
+// reached. Under that implementation THIS case mints a second carrier — two
+// nodes under one id, which is exactly the ambiguous rewrite target clause 2
+// exists to prevent.
+//
+// NOT justified by a known in-tree emission order: none of the three callers
+// mints a path-named container at all today, so neither ordering is currently
+// produced by a real extractor. That is the reason to pin it rather than a
+// reason not to — clause 2's correctness must not rest on an ordering
+// assumption nothing in the tree states or enforces.
+func TestFileCarrierFor_NoSecondCarrierWhenThePathNamedRecordComesLast_6815(t *testing.T) {
+	recs := []types.EntityRecord{
+		recWithRel("cache.hrl", types.RelationshipRecord{
+			FromID: carrierPath, ToID: "cache.hrl", Kind: "IMPORTS"}),
+		{Name: carrierPath, Kind: "SCOPE.Component", Subtype: "file", SourceFile: carrierPath},
+	}
+	if _, ok := FileCarrierFor(carrierPath, "erlang", recs); ok {
+		t.Fatal("a record named after this path rejects wherever it sits in the slice, " +
+			"not only when it precedes the anchoring record")
+	}
+}
+
 // Any edge kind anchors, not only IMPORTS: the resolution requirement is about
 // the FromID, not about the kind. Axis VARIED: the relationship Kind. HELD
 // CONSTANT: the anchoring FromID.
--- a/internal/extractor/file_carrier_6815_test.go
+++ b/internal/extractor/file_carrier_6815_test.go
@@ -1,0 +1,140 @@
+package extractor
+
+// #6815 — unit grading for FileCarrierFor's two guard clauses. Each clause has
+// at least one input where THAT clause alone rejects, so neither is graded only
+// through the other.
+
+import (
+	"testing"
+
+	"github.com/cajasmota/grafel/internal/types"
+)
+
+const carrierPath = "src/cache_server.erl"
+
+func recWithRel(name string, rel types.RelationshipRecord) types.EntityRecord {
+	return types.EntityRecord{
+		Name:          name,
+		Kind:          "SCOPE.Component",
+		SourceFile:    carrierPath,
+		Relationships: []types.RelationshipRecord{rel},
+	}
+}
+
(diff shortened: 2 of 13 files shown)


cajasmota:
## Coordinator verification — APPROVE, and my mutant found the one thing the review missed

**My killing mutant, which nobody else ran:** hoist `if anchored { continue }` above the `records[i].Name == path` check.

```
./internal/extractor/  ./internal/extractors/{erlang,nim,groovy}/  →  all ok   ← ALIVE
```

Under that mutant, once anchoring is satisfied later records are never checked for the path name — so a file whose extractor mints its path-named container **after** the anchoring record gets a **second** carrier, which is exactly the "two nodes under one id, rewrite target ambiguous" outcome that clause exists to prevent. The existing fixture put the path-named record first, so the ordering was unobserved.

Now **DEAD**, and the author supplied the asymmetry check that proves the two cases are not one merged case grading neither:

| mutant | effect |
|---|---|
| reorder (`M12`) | kills **only** the new last-position case; the original stays green |
| delete the clause (`R3`) | kills **both** cases |

Neither case is graded solely through the other. `M1` and `M5` were re-scored after the nim body change and stayed DEAD, so the ablation did not go vacuous.

Worth recording: **none of the three callers mints a path-named container today**, so neither ordering is currently produced. The test was kept anyway, and says so — a clause's correctness must not rest on an ordering assumption nothing in the tree enforces.

**I verified the load-bearing claim myself** rather than relaying it: `file_carrier.go`'s diff in this round is **comment-only** (filtered the diff to non-comment, non-blank lines — empty), so no extractor behaviour changed and the golden numbers cannot have moved.

## What the review established, independently of the author

- **The conditional-carrier decision is right, with in-tree evidence.** The orphan definition is stated at `internal/quality/audit/orphan_structural_test.go:31` (*"only CONTAINS → orphan"*), and an unconditional carrier was measured to mint **3 erlang + 2 nim structural orphans** and to **cost erlang a graded relationship** (26→25), because `cache.hrl` acquires a competing path-named node.
- **Golden deltas re-derived from disk on both sides**, matching `baseline.json` exactly: erlang 20/20→21/21 (totals 22/53→23/54), nim 21/21→22/22, groovy 13/16→14/17. The diff of the two kept graphs shows the defect and fix in four lines — on `main` the edge's `from_id` is a literal path string that is not any node's id; on the branch it is the carrier's id.
- **The digest genuinely cannot move**: `writeSpanFixture6118`'s 21 members were enumerated independently — all `.cs`/`.java`/`.ts`/`.js`/`.py`/`.go`/`.rs`, no `.erl`/`.nim`/`.groovy`. My brief's warning that it "plausibly moves" was wrong, and the *membership* check settled it rather than any reasoning about entities-vs-properties.
- **`M10` stays ALIVE and that is the correct disposition**: `TagEntitiesLanguage` fills an empty `Language` and never overwrites a wrong one, so an empty token is genuinely equivalent under the suite while a wrong token (`M11`) dies. Recorded with its reason instead of being forced into a kill.

## Measured, not assumed: neither test layer is redundant

An over-broad anchor (`FromID != ""` instead of `== path`) was **invisible to all three golden fixtures** and survived only into the unit suite. Conversely the wrong-`subtype` mutant reverted the golden totals to exactly main's numbers and was caught there. The two layers each catch what the other misses — that is now measured rather than hoped.

## Three corrections to MY issue body, all verified

1. The erlang golden row was **already `must_exist: true`**, not `nice_to_have`.
2. **`from_bare_name` was fully landed, not prose** — #6488 arm C shipped it (`diff.go:591`, 16 tests), and the row *was* matching through it. This one mattered: leaving it in place with the carrier added drops erlang-otp-mini to **25/26**, reproduced to the digit.
3. The "erlang is unique" note **already** named all three languages and pointed at #6815.

## Filed separately, not fixed here

**#6847** — nothing would catch a *fourth* extractor doing this. `file_anchored_rels_guard_test.go` skips `IMPORTS` by the #120 convention with an empty `knownInvisibleOffenders`. Framed as the narrower invariant (every path-anchored `FromID` names an emitted record) rather than un-skipping `IMPORTS`, with the measurement required **before** the fix — a guard added green tells you nothing.

Issue #6837, opened 2026-09-05
The feedback report's resolution section blames a missing resolution property that the code stopped reading


cajasmota:
The feedback report's resolution section blames a missing `resolution` property that the code stopped reading

## The defect

`internal/feedback/render.go:193` renders, when the resolution table is empty:

> `_No relationship resolution data available (no `resolution` property found on edges)._`

That message names a mechanism **that no longer exists**. `report.go:449-456` retired the property switch entirely — dispositions are now derived structurally from the `ToID` shape via `resolve.IsResolvedToID`. The comment sitting right there says so:

> *"the pipeline never writes a `Properties["resolution"]` tag, so the previous property switch always reported 'no resolution property found on edges'"*

So the code knows the old mechanism was removed **because it never worked**, and the empty-branch text still reports its absence as the diagnosis. A reader who hits this line is sent to look for a missing edge property, which is not a thing the pipeline has written for some time and is not why the table is empty.

## Why it matters more than a typo

This is the failure mode of a diagnostic, not of a label: it does not merely describe the wrong thing, it actively **directs the next investigation somewhere false**. Someone debugging an empty resolution table will go looking for why edges lack a `resolution` property, find that nothing writes one anywhere, and have to work backwards to discover the message is stale. That is a grounding pass spent to learn nothing.

The correct message should name whatever actually produces an empty table under the current structural derivation.

## Not established

- **What now makes the table legitimately empty.** Under the `ToID`-shape derivation, an empty table presumably means zero relationships were examined — but that has not been traced, and writing a replacement message requires knowing. Do not guess it into the string.
- Whether any other empty-branch or error text in this file names a retired mechanism. **Only the resolution section was examined at this depth.** The broader label-vs-computation sweep that found this covered *labels*, not *diagnostic strings*, so the diagnostic strings are an unswept surface.
- Whether this message has ever actually rendered in practice.

## Provenance

Found during a label-vs-computation sweep of `internal/feedback/render.go` while fixing #6827, and independently confirmed by an adversarial review. **Pre-existing on `main`**, not introduced by that branch. Filed separately because writing a correct replacement needs the "what makes it empty now" answer above, which is real investigation rather than a string edit.

Sibling from the same sweep, different mechanism and more serious: three resolution-table rows driven by counters that are never incremented, rendering a permanent `0.00%` as though measured.

cajasmota:
## Evidence for this issue: the "never writes" comment is false, and here are the seven writers

[...]

Issue #6826, opened 2026-09-05
"Annotation coverage" divides framework-tagged entities by EVERY entity, so its value is set by graph composition rather than framework detection


cajasmota:
"Annotation coverage" divides framework-tagged entities by EVERY entity, so its value is set by graph composition rather than by framework detection

## The defect

`internal/feedback/report.go:258` computes annotation coverage over **every entity in the graph** as the denominator.

But the `framework` property is only ever stamped by `internal/patterns/*` and `internal/custom/csharp/*`. Most entity kinds can never carry it — schema fields, import placeholders, file carriers, markdown code blocks — and `internal/extractors/csharp/` stamps it nowhere at all.

So the annotatable surface is a few hundred entities **by construction**, while the denominator is the whole graph. The resulting percentage is dominated by how many non-annotatable entities a corpus happens to contain.

## Why this is worse than an imprecise metric

A metric whose value moves mainly with unrelated graph composition is not just noisy — it is **actively misleading**, because it looks like a quality signal and is reported as one.

Concretely: on the corpus in **#6726** (external report, ~73k entities, C#/F#/Java/protobuf), annotation coverage read **1.0% — 696 of 73,182**. That corpus contains roughly 35,000 schema fields. The 1% is very close to what a healthy system should produce there, but it reads as near-total failure.

**It misled me.** In my first reply on #6726 I treated the 1.0% as supporting a detector/edge-builder mismatch. It cannot carry that reading, and I have posted a correction. The reporter reasonably drew the same conclusion from the same number — which is the strongest evidence that the metric's presentation is the problem, not their reading of it.

## What a fix has to decide

1. **What the denominator should be.** The obvious candidate is "entities of a kind that a framework detector could plausibly tag". That set needs defining explicitly rather than inferred, or the metric acquires a *second* implicit assumption to go stale.
2. **Whether the label survives.** If the denominator narrows, "annotation coverage" starts meaning something different from what past reports meant by it. Any historical comparison across that change is invalid and should be called out where the number is rendered.
3. **Whether it should be a ratio at all.** "696 entities carry a framework tag, of ~N annotatable" may be more honest than any single percentage — a count plus its population states the two facts separately instead of collapsing them into a figure that hides one.

## Related, and the reason to be careful here specifically

[...]

Issue #1085, opened 2026-05-20
[indexer][core] Stop emitting placeholder External entities for stdlib targets — they pollute graph + MCP


cajasmota:
## Problem

When a resolver classifies an edge as `DispositionDynamic` via a stdlib pattern (`int`, `console.log`, `fmt.Sprintf`, `List.map`, etc.), the indexer STILL emits a placeholder `External` entity for the target name.

These placeholder entities:
- Pollute the graph (every codebase has 50+ stdlib targets)
- Inflate bug-rate (counted as unresolved)
- Skew MCP queries (`find_callers_of_int` returns spurious 158 entries)
- Mislead the dashboard graph (hub nodes that aren't user code)

Earlier #1081 patched the dashboard cosmetically with `?include_external=false`, but MCP still sees them and bug-rate still counts them.

## Real example
upvate Python: `int(value)` calls scattered across the codebase. Resolver correctly marks the edge `DispositionDynamic` via our Python pattern catalog. But an `External(name='int', kind='External', from='upvate_core', language='python')` entity is created with 158 inbound edges.

## Fix

When a resolver classifies an edge `DispositionDynamic` AND the target matches a known dynamic-pattern catalog entry:
1. Keep the edge
2. Set a property like `dynamic_target="int"` on the edge
3. **Do not create a placeholder External entity** for the bare-name target
4. The edge's `ToID` becomes empty or null, with the target reconstructible from the property

For edges to KNOWN external libraries (not stdlib stubs — actual npm/pypi packages), keep emitting External entities — those carry useful metadata.

## Affected languages (per-language stdlib catalogs already exist)

Python (#961, #1067 had partial fix), TS/JS (#962), Go (#945 + #1029), Ruby (#977), Kotlin (#990), Scala (#1005), C# (#1006), Rust (#1009), PHP (#1011), Swift (#1008), Lua (#1012), Dart (#1014), Bash+Just+Shell (#1015), Elixir (#1019), Zig (#1021), Crystal (#1043), Nim (#1039), Lisp family (#1063), Verilog (#1073), VHDL (#1075), Pony+Idris (#1080), Reason+ReScript (#1051/1052), Elm (#1049), F# (#1045), Haskell (#1046), OCaml (#1050), SML (#1061), Erlang (#1047), Clojure (#1038), Groovy (#1030), Razor (#1031)

## Acceptance

- Re-index upvate
[...]

Issue #95, opened 2026-05-08
External synth swallows getattr/setattr/eval as external-unknown before dynamic check fires


cajasmota:
Surfaced by #90 review: external synthesiser tags Python builtins like getattr/setattr/eval/exec with ext: prefix BEFORE the dynamic-pattern classifier sees them. ~40 endpoints in chi corpus that should be dynamic are mis-labeled external-unknown.

**Fix:** in classifyDispositionLang or upstream, run the dynamic-pattern check BEFORE the external classifier. Reflection builtins are inherently dynamic, not external imports.

Refs #90.

cajasmota:
## Implemented

**Root cause:** `classifyDispositionLang` checked the `ext:<pkg>` prefix before the per-language dynamic-pattern catalog, so reflection builtins (Python `getattr`/`setattr`/`hasattr`/`delattr`/`eval`/`exec`/`compile`/`__import__` and JS `Function`/`eval`) — which the external synthesiser stamped as `ext:builtins` / `ext:globalThis` via its stdlib stop-list — were classified as `ExternalUnknown` instead of `Dynamic`.

**Files touched:**
- `internal/resolve/refs.go` — reordered checks in `classifyDispositionLang`; extended `pythonDynamicPatterns` with `hasattr`/`delattr`/`compile` and `jsDynamicPatterns` with bare `eval`/`Function`.
- `internal/external/synth.go` — removed `getattr`/`setattr`/`__import__` from `stdlibBareNames` (broader exclusion documented in comment).
- `internal/resolve/refs_test.go` — added 2 new test functions (23 sub-cases total).
- `internal/external/synth_test.go` — added 1 new test function (8 sub-cases).

**Tests added:** 3 new test functions, 31 sub-cases total.

**Verification:** `go build ./...` clean, `go test ./...` all packages pass, `gofmt -l` clean for touched files, `go vet ./...` clean.

**Commit:** ee7e948
**PR:** https://github.com/cajasmota/archigraph/pull/97

**Follow-up gotchas:**
- VERIFY-2 corpus run skipped (long-running). Worth a follow-up to confirm the aggregate bug-rate moves the right direction on the standard corpus.
- The per-language dynamic catalog is now the sole source of truth for reflection-builtin classification — any language that adds reflection primitives must add patterns there, NOT to the synthesiser stop-list.

cajasmota:
Spec review (PR #97): **LGTM**

[...]

PR #6850, opened 2026-09-05
fix(#6836): three resolution rows rendered a permanent 0.00% — they were uncounted, not unreachable, and carry 20% of the edges

### LINKED ISSUE #6836: Three rows of the feedback report's resolution table are driven by counters that are never incremented — a permanent 0.00% presented as a measurement
Three rows of the feedback report's resolution table are driven by counters that are never incremented — a permanent `0.00%` presented as a measurement

## The defect

`internal/feedback/report.go` declares six resolution-disposition counters. **Three of them are never incremented anywhere.**

| counter | declared | incremented |
|---|---|---|
| `resResolved` | `:241` | `:464` |
| `resExternalKnown` | `:239` | `:462` |
| `resBugExtractor` | `:244` | `:467` |
| **`resExternalUnknown`** | `:240` | **nowhere** |
| **`resBugResolver`** | `:242` | **nowhere** |
| **`resDynamic`** | `:243` | **nowhere** |

All six are summed into `total` at `:623` and divided into percentages at `:630/:632/:633`. So `render.go:200/202/203` renders `external-unknown`, `bug-resolver` and `dynamic` as `0.00%` — **always**, on every corpus, forever — underneath a `_Total edges examined: N_` footer that tells the reader all six dispositions were evaluated against N edges.

Verified by grepping every write to each name, not by reading the declarations.

## Why this is worse than a missing row

`0.00%` and "not measured" are different claims, and the report makes the false one. A reader — including the external reporters we point at this report — sees three dispositions confidently measured at zero and reasonably concludes the resolver never produces them. The correct statement is that nobody knows, because nothing counts them.

This is the same misreading #6827 exists to prevent, on the report's **headline** resolution table rather than in a section label. It is also a sharper instance of a lesson already recorded in this repo: **a zero measurement is corpus-relative** — except here the zero is not even corpus-relative, it is structural, and no corpus can ever move it.

## Not established

- **Whether the three dispositions are genuinely unreachable, or merely uncounted.** These are different bugs with different fixes. `:449-456` derives dispositions structurally from `ToID` shape via `resolve.IsResolvedToID`; whether that classification can ever yield external-unknown / bug-resolver / dynamic has **not** been determined. If it can, the counters are simply missing their increments. If it cannot, the rows should be deleted rather than wired up.
- Whether the counters were ever incremented historically, and lost their call sites in a refactor. Not investigated.
- Whether `total` being computed from three live and three dead counters skews the *other three* percentages. It does not change their ratios to each other, but it does mean `total` is not "edges examined" in the sense the footer claims — worth checking against `:642` before trusting any of the six.

`resExternalKnown` was checked for the same defect and is **not** dead: `resolve.IsResolvedToID` accepts an `ext:` prefix (`toid_shape.go:41`), so `:462` is reachable.

## Provenance

Found during a full label-vs-computation sweep of `internal/feedback/render.go` while fixing #6827, then independently confirmed by an adversarial review. **Pre-existing on `main`**; not introduced by #6827's branch, which touches neither file's logic. Filed separately rather than folded in, because wiring or deleting these rows is a behavioural change and #6827 is a label fix.

Sibling from the same sweep, different mechanism: the stale empty-branch text naming a retired `resolution` property switch. Related by class: #6826 (annotation coverage's denominator), #6827 (source-window label).

cajasmota:
# fix(#6836): wire the three dead resolution counters to the indexer's real classifier

Closes #6836

## Path (a) — wired, not deleted. My first verdict was wrong.

My first pass deleted `external-unknown`, `bug-resolver` and `dynamic` as structurally
unreachable. **That was wrong, and the review's counter-example is correct.** I reproduced it
independently before acting on it: on this branch's own `internal/feedback/testdata/golden/graph.fb`
(796 edges), classifying with the real machinery yields

```
resolved 488  external-known 118  external-unknown 59  external-sql 0
dynamic 84    bug-extractor 30    bug-resolver 17     unclassified 0
```

The three "dead" rows carry **160 of 796 edges — 20% of the population.** My error was
conflating *"`Generate` does not take an allowlist"* with *"no allowlist is reachable"*. It can
simply call the predicate. Verified myself: `resolve.BuildIndex` (`refs.go:1014`),
`resolve.Index.ClassifyEndpoints` (`refs.go:929`), `external.IsKnownExternalPackage`
(`synth.go:17486`), and `go list -deps ./internal/external` / `./internal/resolve` contain no
`internal/feedback` — no cycle in either direction.

## The structural fix, not just the wiring

Rather than re-adding six hand-maintained counters, the table is now **driven by
`resolve.AllDispositions` on both sides**: `report.go` populates one entry per disposition from
`ClassifyEndpoints`, and `render.go` iterates the same list to emit rows. A disposition can no
longer exist in the taxonomy and be silently absent from — or unfed by — the table, which is
the defect class #6836 filed. `ResolutionVector`'s six named fields become
`ByDisposition map[string]float64` with a `Pct(name)` accessor. `sanity.go`'s sums-to-100%
check sums the taxonomy too, so a missing bucket fails the check instead of quietly shrinking
its scope.

This also picks up `external-sql` and `unclassified`, which the old six-row table omitted
entirely despite both having live `return` sites in `classifyDispositionLang`
(`refs.go:4190`, `refs.go:4249`). They are 0 on the golden corpus — a corpus-relative zero,
not a structural one.

## Correction to the review: `ToOriginal = ToID` is NOT exact

The review asked me to verify this rather than take it, which was the right call — **it is
inexact, and I measured by how much.**

The claimed justification was that "per #95 reflection builtins are no longer synthesised to
`ext:`". That is not so. `resolve.IsDynamicPatternLang` — the exported guard whose doc comment
says the synthesiser uses it to avoid stamping dynamic stubs — **has no caller anywhere outside
`refs.go`** (`grep -rn IsDynamicPatternLang internal/ cmd/` returns only its definition). The
guard does not exist, and `refs.go:4118-4126` says the opposite in prose: the synthesiser *does*
stamp `getattr`/`eval` with `ext:` because they sit in the stdlib stop-list, which is exactly why
`classifyDispositionLang` runs its dynamic check **before** the `ext:` check.

Measured consequence, directly:

```
IsDynamicPatternLang("getattr", "python")     = true
IsDynamicPatternLang("ext:getattr", "python") = false
```

so feeding the raw ToID silently defeats #95. On `testdata/golden` it moves **6 of 796 edges**
out of `dynamic` and into `external-unknown`.

`ToOriginal: strings.TrimPrefix(rel.ToID, "ext:")` is the fix. It is **approximate, not
exact** — see round 2 below, where a second review corrected my own overreach on this point.
Pinned by `TestResolution_ExtPrefixedDynamicBuiltinIsDynamicNotExternal`; M1 is the mutant that
reverts it.

## The other review items

- **Non-allowlisted `ext:` fixture** — added.
  `TestResolution_ExternalKnownAndUnknownAreSeparated` uses `ext:react` (allowlisted) against
  `ext:not_a_real_pkg_xyzzy` (not), and pins **both** percentages, so collapsing the split in
  either direction fails. M2 (allowlist always true) and M3 (always false) are both DEAD on it.
- **The vacuous `1-5` clause** — dropped. The comment was also false: `countRangeLabel` returns
  `1-5` for any `n <= 5`, so 2 and 4 edges are indistinguishable through it and the clause could
  never reject alone. The test now pins the two surviving percentages instead, which *do* move
  if an empty-target edge is given a disposition.
- **`skills/grafel-feedback/SKILL.md:215`** — updated to name the full taxonomy and its source.
  (The rename that made it stale is gone; the row names are the original ones again.)

## Golden test D3: the old invariant now asserts the bug

`TestGenerate_GoldenFB` compared `resolved + external` against `IsResolvedToID` (hex **or**
`ext:`). Post-fix that is off by 4 edges — precisely the ext:-stamped builtins the classifier
now correctly calls `dynamic`. Keeping it would pin the #95 defect, so it is replaced by two
tighter clauses:

- **D3a** — `resolved` must equal the **hex-only** fraction. `classifyDispositionLang`
  short-circuits on `isHexID` before anything else, and `isHexID` is the same
  16-lowercase-hex test, so this is an identity rather than an approximation.
- **D3b** — `external-unknown`, `bug-resolver` and `dynamic` must all be **> 0 on the real
  fixture**. This is the regression guard a hand-built fixture cannot provide: the original
  defect was that *no corpus could ever move these*.

## Mutants — re-scored in full after the round-2 restructure

Permissive first. `go vet` before every score, `-count=1`, exact-1-match edits aborting on ≠1,
edit-landed asserted, byte restore verified by sha.

| # | mutant | direction | verdict | killed by |
|---|---|---|---|---|
| M1 | `ToOriginal` keeps the `ext:` prefix (defeats #95) | permissive | DEAD | `...ExtPrefixedDynamicBuiltinIsDynamicNotExternal` |
| M2 | allowlist always true (every external "known") | permissive | DEAD | `TestGenerate_GoldenFB`, `...ExternalKnownAndUnknownAreSeparated` |
| M3 | allowlist always false (every external "unknown") | permissive | DEAD | `TestGenerate_ResolutionDisposition`, `...ExternalKnownAndUnknownAreSeparated` |
| M4 | empty-ToID guard dropped, alone | permissive | **EQUIVALENT** | masking pair — see below |
| M5 | `Language` dropped (per-language dynamic catalog never runs) | permissive | DEAD | `...DynamicIsSeparatedFromBugExtractor`, `...ExtPrefixedDynamicBuiltinIsDynamicNotExternal` |
| M11 | `total := len(endpoints)`, alone | permissive | **EQUIVALENT** | masking pair — see below |
| M12 | **compound**: guard dropped AND `total := len(endpoints)` | permissive | DEAD | `...FooterDoesNotClaimEveryEdgeWasExamined` |
| M6 | name index built from no entities | restrictive | DEAD | `TestGenerate_GoldenFB`, `...BugResolverIsSeparatedFromBugExtractor` |
| M7 | render only the first disposition row | restrictive | DEAD | 8 tests |
| M8 | sanity sums a hand-list instead of the taxonomy | restrictive | DEAD | `TestRunSanityChecks_AllPass` |
| M9 | footer reverts to "Total edges examined" | restrictive | DEAD | `...FooterDoesNotClaimEveryEdgeWasExamined` |
| M10 | vector omits the three formerly-dead dispositions | restrictive | DEAD | `TestGenerate_GoldenFB` + 4 |

**The masking pair (M4 / M11 / M12), named explicitly in the code.** `ClassifyEndpoints` records
a disposition only `if ep.ToID != ""`, and `ResolutionTotal` is summed from `DispositionCounts`.
So dropping the guard alone is equivalent, and switching the total to `len(endpoints)` alone is
equivalent — but **doing both counts empty-target edges in the denominator**, and the footer
test kills that. Two equivalent mutants whose compound is not: recorded as EQUIVALENT with the
reason, never as DEAD, and no fixture manufactured to force a kill. M2/M3/M8 were each first
scored NON-COMPILING (unused import) and re-run in a compiling form that still references the
real symbol — a non-compiling mutant proves nothing.

## Round 2 — four blocking items, all confirmed by measurement

- **B1 — the round-1 false premise was still shipped, verbatim.** `report_test.go` carried
  *"the finer splits need the resolver's allowlist, name index and pre-resolution stubs, which
  the persisted graph does not carry"* directly above a test whose code path **builds** that
  allowlist and index and measures all three — and *"those three are the whole vector"* when
  the vector has eight entries. Rewritten. I then swept every prose site in the diff for the
  same framing; the three remaining "persisted graph does not store" phrases are accurate
  statements about the *stub*, which genuinely is not stored.
- **B2 — the golden comment pinned my own pre-fix number.** It said 59 external-unknown edges;
  that is the count with my `TrimPrefix` fix **reverted**. Measured on the branch: **53**. 17
  and 90 were right, and 53+17+90 still sums to 160, which is exactly why the headline survived
  and the per-row figure did not.
- **B3 — "two cases, and both are exact" was false, and I have corrected my own claim.**
  `synth.go` stores a **canonical** name, optionally `<pkg>:<leaf>`. **14 of the 82 distinct
  `ext:` ids in `testdata/golden` are not the original stub** (`ext:fastapi:Depends` — the stub
  was `Depends`), and the fixture holds **both** `ext:ArrayList` and `ext:java:ArrayList`: one
  stub canonicalised two ways. The block now says **approximate**, bounds the approximation to
  its single consumer (the dynamic-pattern check), and says why that bound is safe — the
  catalogs anchor `^…$` on leaf names, so the error direction is always "not dynamic", never a
  false dynamic. **Verified rather than relayed:** across all 175 `ext:` edges in the fixture,
  trimmed and last-colon-leaf give identical verdicts, 0 disagreements. The residual gap is
  named rather than papered over: an Elixir `Repo.insert` canonicalised to `ecto:insert` would
  land in `external-unknown` where the indexer says `dynamic` — a silent disagreement with
  `orient view=stats` that does not occur in this corpus and needs the synthesiser to persist
  the pre-canonical stub to close.
- **B4 — my fix REMOVED coverage that `main` had.** `sanity.go` promises *"a disposition
  missing from the vector must fail this check"*, but this branch had rewritten both fixtures
  down to exactly the three dispositions a hand-list would name, making the sum's scope
  unobservable — the hand-list mutant survived here and **fails on main**. Fixtures now
  populate `dynamic` and `external-sql`, outside any plausible hand-list. M8 is DEAD.

### Round-2 report-only items, fixed because they are prose

- **R3** — the six affected edges are **stdlib leaf-callee dispatch**, not reflection builtins:
  `print`(zig), `filter`(clojure), `first`(swift), `insert`(lua, elixir), `all`(elixir). No
  `getattr`/`eval`/`Function` edge exists in the corpus; the unit test manufactures one.
  Corrected in both doc blocks.
- **R2** — six edges are routed `ext:`→`dynamic`; **four** is the net on that identity after
  two non-`ext:` callees move the other way. Right number, wrong mechanism; now says both.
- **R4** — added `TestResolution_ExternalSQLCanFire` and `TestResolution_UnclassifiedCanFire`.
  Both are 0.00% on every fixture, and their honesty rested only on a doc comment — a prose
  claim nothing checks, inside the fix for that defect class. Both fire.
- **R6** — dropped a doc block justifying two `0.00%` clauses as a double-count guard. Pinning
  `dynamic | 100.00%` already forces every other bucket to zero, so neither clause can reject
  alone; they are now described as what they are.

### Round-2 item I did NOT fix

**R1** — `AllDispositions` is itself a hand-maintained slice, and dropping `DispositionExternalSQL`
from it removes the row *and* its bucket from the denominator with everything green. The
coordinator is filing this. I have removed "by construction" phrasing where the construction
rests on that slice.

## Verification

- `go test -count=1 ./internal/feedback/ ./internal/resolve/ ./internal/external/ ./internal/cli/` — all ok.
- `go build ./...` — clean. `go vet ./internal/feedback/` — clean.
- `gofmt -l internal/ cmd/ tools/` — empty.
- `go run ./tools/coverage validate` — exit 0.
- `-run 'TestResolution_'` proven to select 6 (`-v` + `grep -c '^=== RUN'`).
- Section scoping re-confirmed: every assertion goes through `resolutionSection()`, and M7
  (which leaves the section present but truncated) is killed by 6 tests.

## Report-only, not fixed here

`resolve.IsDynamicPatternLang`'s doc comment claims "used by the external synthesiser … see
issue #1085", but nothing outside `refs.go` calls it. Behaviour is unaffected (the ordering
inside `classifyDispositionLang` is what implements #95), but it is a stale claim in
`internal/resolve`, which is another lane's file. Flagging, not touching.

The coordinator is separately recording `report.go`'s "the pipeline never writes a
`Properties["resolution"]` tag" as false under #6837; that comment is gone from this file now
anyway, replaced by the classifier call.

## Not touched

`#6826` and `#6837`. #6837's `ResolutionTotal == 0` empty branch in `render.go` is one line
above my edit and is byte-identical to `main`.

### DIFF
--- a/internal/feedback/render.go
+++ b/internal/feedback/render.go
@@ -196,11 +196,13 @@ func Render(w io.Writer, r *Report) error {
 	} else {
 		rv := r.Resolution
 		fmt.Fprintf(w, "| Disposition | Percentage |\n|---|---|\n")
-		// Rows are driven by resolve.AllDispositions, not by a hand-written
-		// list: before #6836 three of six rows were wired to counters nothing
+		// Rows are driven by resolve.AllDispositions, not by a list maintained
+		// here: before #6836 three of six rows were wired to counters nothing
 		// incremented and rendered a permanent 0.00%. Iterating the taxonomy
-		// makes that shape impossible — every disposition the classifier can
-		// return has a row, and every row is fed by the classifier.
+		// removes that failure mode for every disposition the slice names —
+		// each gets a row, and each row is fed by the classifier. It does NOT
+		// cover a Disposition that reaches the enum without reaching
+		// AllDispositions, which nothing currently checks.
 		for _, d := range resolve.AllDispositions {
 			fmt.Fprintf(w, "| %s | %.2f%% |\n", d.String(), rv.Pct(d.String()))
 		}
--- a/internal/feedback/render.go
+++ b/internal/feedback/render.go
@@ -6,6 +6,8 @@ import (
 	"sort"
 	"strings"
 	"time"
+
+	"github.com/cajasmota/grafel/internal/resolve"
 )
 
 // Render writes the markdown feedback report to w. Section order is stable.
@@ -194,18 +196,15 @@ func Render(w io.Writer, r *Report) error {
 	} else {
 		rv := r.Resolution
 		fmt.Fprintf(w, "| Disposition | Percentage |\n|---|---|\n")
-		fmt.Fprintf(w, "| resolved | %.2f%% |\n", rv.ResolvedPct)
-		fmt.Fprintf(w, "| external | %.2f%% |\n", rv.ExternalPct)
-		fmt.Fprintf(w, "| unresolved | %.2f%% |\n\n", rv.UnresolvedPct)
-		// Say what these three buckets do NOT separate. The report used to
-		// render external-unknown / bug-resolver / dynamic rows at a permanent
-		// 0.00%, which read as "measured, and zero" when the truth is that the
-		// persisted graph cannot tell those cases apart (#6836).
-		fmt.Fprintf(w, "Classified by ToID shape alone, which is all a persisted graph retains. "+
-			"`external` does not separate allowlisted packages from unknown ones, and `unresolved` "+
-			"does not separate extractor bugs from resolver misses or from dynamic dispatch: those "+
-			"splits need the resolver's allowlist, name index and pre-resolution stubs, none of "+
-			"which survive into the graph this report reads.\n\n")
+		// Rows are driven by resolve.AllDispositions, not by a hand-written
+		// list: before #6836 three of six rows were wired to counters nothing
+		// incremented and rendered a permanent 0.00%. Iterating the taxonomy
+		// makes that shape impossible — every disposition the classifier can
+		// return has a row, and every row is fed by the classifier.
+		for _, d := range resolve.AllDispositions {
+			fmt.Fprintf(w, "| %s | %.2f%% |\n", d.String(), rv.Pct(d.String()))
+		}
+		fmt.Fprintf(w, "\n")
 		fmt.Fprintf(w, "_Edges with a target to resolve: %s (edges with an empty target are excluded)_\n\n",
 			countRangeLabel(r.ResolutionTotal))
 	}
--- a/internal/feedback/render.go
+++ b/internal/feedback/render.go
@@ -195,12 +195,19 @@ func Render(w io.Writer, r *Report) error {
 		rv := r.Resolution
 		fmt.Fprintf(w, "| Disposition | Percentage |\n|---|---|\n")
 		fmt.Fprintf(w, "| resolved | %.2f%% |\n", rv.ResolvedPct)
-		fmt.Fprintf(w, "| external-known | %.2f%% |\n", rv.ExternalKnownPct)
-		fmt.Fprintf(w, "| external-unknown | %.2f%% |\n", rv.ExternalUnknownPct)
-		fmt.Fprintf(w, "| bug-extractor | %.2f%% |\n", rv.BugExtractorPct)
-		fmt.Fprintf(w, "| bug-resolver | %.2f%% |\n", rv.BugResolverPct)
-		fmt.Fprintf(w, "| dynamic | %.2f%% |\n\n", rv.DynamicPct)
-		fmt.Fprintf(w, "_Total edges examined: %s_\n\n", countRangeLabel(r.ResolutionTotal))
+		fmt.Fprintf(w, "| external | %.2f%% |\n", rv.ExternalPct)
+		fmt.Fprintf(w, "| unresolved | %.2f%% |\n\n", rv.UnresolvedPct)
+		// Say what these three buckets do NOT separate. The report used to
+		// render external-unknown / bug-resolver / dynamic rows at a permanent
+		// 0.00%, which read as "measured, and zero" when the truth is that the
+		// persisted graph cannot tell those cases apart (#6836).
+		fmt.Fprintf(w, "Classified by ToID shape alone, which is all a persisted graph retains. "+
+			"`external` does not separate allowlisted packages from unknown ones, and `unresolved` "+
+			"does not separate extractor bugs from resolver misses or from dynamic dispatch: those "+
+			"splits need the resolver's allowlist, name index and pre-resolution stubs, none of "+
+			"which survive into the graph this report reads.\n\n")
+		fmt.Fprintf(w, "_Edges with a target to resolve: %s (edges with an empty target are excluded)_\n\n",
+			countRangeLabel(r.ResolutionTotal))
 	}
 
 	// Section 4 — Framework Recognition
--- a/internal/feedback/report.go
+++ b/internal/feedback/report.go
@@ -37,14 +37,31 @@ type KindStats struct {
 	OrphanPct   float64
 }
 
-// ResolutionVector holds the disposition breakdown for graph relationships.
+// ResolutionVector holds the disposition breakdown for graph relationships,
+// to the resolution a ToID SHAPE can actually support.
+//
+// This report reads a PERSISTED graph: per edge it has the final ToID and
+// nothing else. Three outcomes are derivable from that — a 16-hex bound entity
+// ID, an "ext:"-prefixed external, and any other non-empty stub the resolver
+// never bound — and this struct carries exactly those three.
+//
+// internal/resolve's Disposition taxonomy is finer (it splits external into
+// known/unknown and unresolved into bug-extractor / bug-resolver / dynamic),
+// but every one of those splits needs an input the persisted graph does not
+// retain: the ExternalAllowlist (known vs unknown), the resolver's name Index
+// (bug-resolver vs bug-extractor) and the PRE-resolution stub plus its language
+// (dynamic). Fields for those three dispositions existed here and were never
+// written, rendering a permanent and false "0.00%" (#6836); they are gone
+// rather than wired up, because there is nothing here to wire them to.
 type ResolutionVector struct {
-	ResolvedPct        float64
-	ExternalKnownPct   float64
-	ExternalUnknownPct float64
-	BugExtractorPct    float64
-	BugResolverPct     float64
-	DynamicPct         float64
+	// ResolvedPct — ToID is a 16-char lowercase-hex entity ID.
+	ResolvedPct float64
+	// ExternalPct — ToID is "ext:"-prefixed. Mixes allowlisted (known) and
+	// unknown externals; the allowlist is not available here.
+	ExternalPct float64
+	// UnresolvedPct — ToID is a non-empty stub of any other shape. Mixes
+	// extractor bugs, resolver misses and intrinsically dynamic dispatch.
+	UnresolvedPct float64
 }
 
 // EntityKindLang is one row in the entity kind × language table.
@@ -144,8 +161,13 @@ type Report struct {
 	OrphanLeafByKind map[string]KindStats
 
 	// Section 3 — Resolution Disposition
-	Resolution      ResolutionVector
-	ResolutionTotal int // total edges examined
+	Resolution ResolutionVector
+	// ResolutionTotal is the number of edges carrying a NON-EMPTY ToID —
+	// the denominator of ResolutionVector. It is NOT the number of edges
+	// examined: an edge with an empty ToID is iterated and then skipped
+	// (there is nothing to resolve), so it appears in TotalRelationships
+	// and not here (#6836).
+	ResolutionTotal int
 
 	// Section 4 — Framework Recognition
 	FrameworkHits          map[string]int // framework → entity count
@@ -238,12 +260,9 @@ func Generate(_ context.Context, docs []*graph.Document, opts Opts) (*Report, er
 
 	// Resolution disposition counters.
 	var (
-		resResolved        int
-		resExternalKnown   int
-		resExternalUnknown int
-		resBugExtractor    int
-		resBugResolver     int
-		resDynamic         int
+		resResolved   int
+		resExternal   int
+		resUnresolved int
 	)
 
 	annotated := 0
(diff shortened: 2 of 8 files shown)


cajasmota:
## Coordinator verification — three rounds, two independent reviews, and the filed defect was the smallest of four

**My killing mutant, which nobody else ran:** rotate the disposition→count mapping by one, so every row keeps its label and receives its neighbour's count. Total unchanged, every row present, still sums to 100% — **a table that is entirely self-consistent and entirely wrong.**

**DEAD — 9 tests**, including both newly-added `TestResolution_ExternalSQLCanFire` and `..._UnclassifiedCanFire`. That matters twice over: the table is pinned per-label rather than only in aggregate, and the two "can this row ever fire" tests added in the last round are load-bearing rather than decorative.

## What this issue actually turned out to be

I filed it as *"three counters are never incremented, so three rows show a permanent 0.00%."* That was the smallest of four defects in the same table:

1. **The three rows were uncounted, not unreachable** — round 1 concluded unreachable and **deleted** them; round 1's reviewer disproved it with a compiling counter-example and measured that they carry **160 of 796 edges, 20%** of the population on the project's own fixture.
2. **`external-known` was reachable but mislabelled.** Unknown externals also persist as `ext:`, so every one of them rendered as *known*; dynamic and resolver-miss edges keep raw stubs and rendered as `bug-extractor`. Deleting the three dead rows alone would have left the survivors making a **stronger** false claim.
3. **The taxonomy and the table were coupled by hand.** Driving both rows and counts from `resolve.AllDispositions` surfaced **two more dispositions the table had silently omitted** — `external-sql` and `unclassified`, both with live return sites, both since proven to fire.
4. **The golden assertion was encoding a bug.** `D3`'s `resolved+external == IsResolvedToID` identity was off by exactly the `ext:`-stamped builtins — it had been *passing because* the classification was wrong. Split into a hex-only identity plus a non-zero check on real data.

## Corrections, in both directions

**The round-1 reviewer was wrong** that `ToOriginal = ToID` is exact — `IsDynamicPatternLang("getattr","python")` is true but `IsDynamicPatternLang("ext:getattr","python")` is false, so raw `ToID` silently defeats #95 on **6 of 796** edges. The author caught it and fixed it with `TrimPrefix`. Round 2 confirmed the author against the reviewer, and confirmed `IsDynamicPatternLang` has no caller outside `refs.go` at all.

**The author was wrong twice, and both were caught by round 2:**
- `TrimPrefix` was documented as **exact**. It is not — **14 of 82** distinct `ext:` ids in the fixture are canonicalised (`ext:fastapi:Depends` came from stub `Depends`; the fixture holds both `ext:ArrayList` and `ext:java:ArrayList`). Consequence on golden is **nil** — verified across all **175** `ext:` edges, 0 disagreements between trimmed and last-colon-leaf — but the claim is now stated as *approximate*, bounded to its one consumer, with the residual `ecto:insert` gap named as a real potential disagreement with `orient view=stats`.
- **The fix removed coverage `main` had.** `sanity.go` promises that a disposition missing from the vector must fail — but the branch had narrowed both fixtures to exactly the three dispositions a hand-list would name, making the sum's scope unobservable. On `main` the equivalent mutant fails. Fixtures now populate `dynamic` and `external-sql`; **M8 is DEAD**.

Also fixed: the round-1 **false premise was still shipped verbatim** in `report_test.go` above a test that measures the very splits it said were unmeasurable, and a golden comment pinned **59** where the post-fix value is **53** — the author's own pre-fix number.

## A masking pair, named

`M4` (drop the empty-`ToID` guard) and `M11` (`total := len(endpoints)`) are **each equivalent alone**, and their **compound `M12` is DEAD**. Both are equivalent only because `ResolutionTotal` sums `DispositionCounts` rather than `len(endpoints)` — that is now the code comment rather than a comment claiming the guard does the excluding, which it does not.

## Left open deliberately

**#6849** — `resolve.AllDispositions` is a hand-maintained slice nothing grades against the `Disposition` enum. Removing a member drops its row **and** its bucket from the denominator, silently redistributing its share across the survivors: a table that sums to 100% and is wrong, with no trace. That is #6836 one level up, and it is worse than the original defect, since a permanent `0.00%` row is at least visible. The four comments that claimed "by construction" now name this residual dependency instead.

**#6837** — `render.go:194`'s zero-total branch still blames a missing `resolution` property. Recorded there with the seven sites that do write one.

PR #6851, opened 2026-09-05
fix(#6846): eleven hand-maintained copies of one exclusion list, two of them live bugs — share the predicate and add the missing lower bound

### LINKED ISSUE #6846: internal/atomicfile/guard_6018_test.go walks into .claude/worktrees/ too — and seven copies of one exclusion list is the actual defect
`internal/atomicfile/guard_6018_test.go` walks into `.claude/worktrees/` too — and seven copies of one exclusion list is the actual defect

## The immediate bug

`scanDeterministicTempNames` (`internal/atomicfile/guard_6018_test.go:435`) is a byte-for-byte clone of the walk fixed in #6842 — same `WalkDir`, same `repoRoot(t)`, same parse-`t.Fatalf` — and its `SkipDir` switch reads:

```go
case ".git", "node_modules", "vendor", "testdata", "dist", "build":
```

**No `.claude`.** So in any development checkout with agent worktrees it parses every worktree's full copy of the production tree, `t.Fatalf`s on a parse error inside an unrelated in-flight branch, and reports offences under paths the allow-lists can never name.

Proved live during the review of #6842: planting three lines of invalid Go at `.claude/worktrees/fake-agent/broken.go` produces

```
--- FAIL: TestNoDeterministicTempNames
    guard_6018_test.go:453: parse .claude/worktrees/fake-agent/broken.go: ... expected ')', found '{'
```

while the same condition against the fixed `internal/registry` walk passes. **CI never sees it** — no worktrees there — so this is a local-development-only failure, which is why it has survived.

## The real defect is the duplication

A survey during that review found **seven** `repoRoot(t)`-rooted walks that each carry their own exclusion list:

| location | `.claude`? |
|---|---|
| `internal/entkinds/scan.go:220` (`skippedDir`) | yes |
| `internal/registry/home_isolation_sweep_guard_6735_test.go:513` | yes |
| `internal/relkinds/undeclared_kinds_sweep_guard_6757_test.go:503` | yes |
| `internal/types/producer_entity_kinds_6776_test.go:266` | yes |
| `internal/registry/home_sweep_guard_6178_test.go` | yes, **as of #6842** |
| **`internal/atomicfile/guard_6018_test.go:435`** | **no — this issue** |
| `internal/types/producer_kinds_test.go:70` | no (`testdata vendor fixtures` only — a narrower walk, may be deliberate) |

#6842 fixed one copy by hand. That is grep-and-patch, and it produces exactly the outcome this repo has hit repeatedly: a fix applied at one of N sites while the others stay broken and nobody knows how many N is. Note the two `internal/registry` copies sit **~30 lines apart in the same package**, one of which was already correct — so this is not even cross-package drift.

## Nothing detects narrowing, in any of them

Two mutants scored during #6842's review, both **ALIVE on the whole package**:

- `case ".claude"` → `strings.Contains(d.Name(), "claude")` — over-broad, undetected.
- widening the switch with `engine, graph, extractors, security` — silently narrows the guard's scope, undetected.

So an exclusion list can be widened to cover real production directories and **no test fails**. That is the direction that actually matters for a *guard*: excluding too much makes it report a clean tree it never looked at, which is #6834's whole subject.

This also means "share the list" and "keep them separate" are **not** discriminated by safety today — the argument for separation (a shared change could silently narrow a guard) describes a risk that already exists unguarded in both designs.

## What a fix should do

1. Add `.claude` to `atomicfile`'s switch — the immediate bug.
2. **Share one exclusion list** across the walks that genuinely want the same thing, rather than adding a seventh hand-maintained copy. `entkinds.skippedDir` already holds the right set but is unexported; exporting it, or lifting it to a small shared helper, is the obvious shape.
3. **Add the lower bound that is missing everywhere**: a test that fails if the exclusion list grows to cover a directory the guard is supposed to inspect. Without it, item 2 is genuinely riskier than item 1, which is the author's stated reason for declining the share in #6842.

Do item 3 before or with item 2, not after.

## Not established

- **Whether `internal/types/producer_kinds_test.go:70`'s narrower list is deliberate.** It omits `.git` as well as `.claude`, which suggests a different intent rather than the same drift. Not investigated; `internal/types` was off-limits during the review.
- **Whether all seven walks want identical semantics.** They are all `repoRoot(t)`-rooted Go-file walks, but "wants the same exclusions" has been assumed from shape, not verified per site. Sharing a list across walks that legitimately differ would be a regression.
- Whether any of the other five has a *different* omission not covered by the `.claude` question.

Found during the independent review of #6842; pre-existing and untouched by it. Related: **#6834** (guards that report success while scanning the wrong thing) — a guard that skips too much is that defect's most direct cause.

cajasmota:
# fix(#6846): two guards walked into `.claude/worktrees/`, eleven copies of one exclusion list, and nothing anywhere graded it for excluding too much

Closes #6846

### 1. The lower bound, first — `internal/repowalk/repowalk_test.go`

`TestSkippedDir_NeverHidesRepositorySource` fails if the exclusion list grows to
cover a directory a guard must descend into to reach real source. It walks the
**repository root**, collects every directory NAME on the path from the root to
a non-test `.go` file (219 of them), and requires `SkippedDir` to return false
for each.

Two design points, both load-bearing:

- **It does not steer by the predicate under test.** It prunes with its own
  literal replica of the list. A lower bound that pruned with `SkippedDir` would
  prune exactly the directories it is meant to prove are read, find nothing
  missing, and pass — #6834's defect inside the test for it.
- **Path COMPONENTS, not the immediate parent.** `internal`, `cmd` and `tools`
  hold no `.go` file directly, so a parent-name-only domain never mentions them
  — and "widen the list with `internal`", which hides the entire source tree, is
  the single most destructive widening available.

`TestSkippedDir_MatchesTheWholeNameOnly` covers over-firing at the string level.
It generates the name space around each skipped entry (16 mutations each —
suffix, prefix, dot added, dot stripped, case-folded, `-backup`, `_old`, …),
filters the exact names, and requires none of the ~100 survivors to be skipped,
with a floor of 80 generated variants so a collapsed generator fails loudly.

`TestSkippedDir_SkipsEveryForeignTree` keeps the recall direction.

### 2. Two live `.claude` omissions, not one

| walk | root | before | symptom |
|---|---|---|---|
| `internal/atomicfile/guard_6018_test.go` | repo root | `.git, node_modules, vendor, testdata, dist, build` | parses every worktree; `t.Fatalf`s on a mid-edit parse error in an unrelated branch |
| `internal/extractors/incremental_phasetrace_6199_test.go` | **module root** | `.git, node_modules, testdata, vendor` | **fails outright, no parse error required** |

The second was not in the issue, not in the review, and is **worse**. Its
`findCallSites` enumerates every call to `tryIncremental` across the module and
`TestPhaseTrace_EveryReturnIsTraced` asserts there is exactly ONE. Every
worktree contains `internal/extractors/incremental.go`, hence its own call site.
Measured with one planted worktree, before the fix:

```
incremental_phasetrace_6199_test.go:103: tryIncremental must have exactly ONE call site
(inside TryIncremental) … found 2: [{shadowCaller …/.claude/worktrees/fake-agent/
internal/extractors/shadow.go:3:27} {TryIncremental …/internal/extractors/incremental.go:427:8}]
```

So this test fails on every local run in a checkout with agent worktrees, and
has since #6199. Both are now pinned by `t.TempDir()` regression tests, because
the ambient checkout is exactly what makes them pass in CI and fail locally.

### 3. The share — seven of eleven

`internal/repowalk.SkippedDir` is a new leaf package (no grafel dependencies, so
no cycle and no odd edge).

| site | shares? | why |
|---|---|---|
| `internal/entkinds/scan.go` | yes | production predicate |
| `internal/relkinds/scan.go` | yes | production predicate |
| `internal/atomicfile/guard_6018_test.go` | yes | sole walk; **had the bug** |
| `internal/extractors/incremental_phasetrace_6199_test.go` | yes | sole walk; **had the bug** |
| `internal/registry/home_sweep_guard_6178_test.go` | yes | sole walk |
| `internal/registry/home_isolation_sweep_guard_6735_test.go` | yes | sole walk |
| `internal/safeio/name_chosen_open_sweep_guard_6478_test.go` | yes | sole walk (its `.claude` case was already dead — it roots at `root/internal` and `root/cmd`, so this is tidying, not a fix) |
| `internal/entkinds/…6744_test.go` `countSourceFiles` | **no** | independent floor vs `entkinds.Scan` |
| `internal/relkinds/…6757_test.go` `countSourceFiles` | **no** | independent floor vs `relkinds.Scan` |
| `internal/types/…6776_test.go` `literalOnlyEntityKinds` | **no** | independent cross-measure vs `entkinds.Scan` |
| `internal/types/producer_kinds_test.go` `walkGoFiles` | **no** | different root (`internal/`), different set (`fixtures`) |

Each of the four exclusions is documented at the walk itself, so the next
grep-and-patch pass does not "finish the job" and couple a floor to what it
measures.

## Why sharing an independent floor would be a regression — demonstrated, not argued

Widening `SkippedDir` with `"security","atomicfile","registry"` and running
`relkinds.TestRepoSweepIsNotVacuous` both ways:

| `countSourceFiles` | result |
|---|---|
| independent (as shipped) | **FAIL** — `scan parsed 2079 non-test .go files; an independent walk … finds 2086` |
| switched to `repowalk.SkippedDir` | **PASS**, while **7 real production Go files went unread by both sides** |

Qualifier: the equality is backstopped by that test's absolute
`wantGo < 1000 || wantYAML < 500` floor, which independently catches *large*
widenings (a first attempt was caught at `18 .yaml files`). The backstop is
coarse; the independence is what makes a small widening visible at all.

## Mutant scores

`go vet` clean before every score, `-count=1`, exact-count replaces aborting on
≠1 occurrence, byte backups restored by `cp` with verified shasums.

**Predicate mutants** (scored across `repowalk` + `atomicfile` + `registry`):

| # | mutant | on `main` | on branch | killed by |
|---|---|---|---|---|
| M1 | `case ".claude"` → `strings.Contains(name,"claude")` | ALIVE | **DEAD** | `MatchesTheWholeNameOnly` (12 over-matches) + atomicfile's `.claude-backup` control |
| M2 | widen with `engine, graph, extractors, security` | ALIVE | **DEAD** | `NeverHidesRepositorySource` + `TestNoDeterministicTempNames` |
| M4 | widen with `cmd, tools` | ALIVE *(also alive on my first draft)* | **DEAD** | `NeverHidesRepositorySource` — names both with example files |
| M5 | widen with `internal` (hides the whole tree) | ALIVE *(also alive on my first draft)* | **DEAD** | `NeverHidesRepositorySource`, plus atomicfile and registry |
| M7 | drop `.claude` (the original #6846 defect, promoted to the shared list) | n/a | **DEAD** | `SkipsEveryForeignTree` + both worktree tests + `TestHandRolledHomeSweepSkipsWorktreeCheckouts` |

**Call-site mutants** (`if repowalk.SkippedDir(…)` → `if false && …`):

| site | verdict |
|---|---|
| `internal/atomicfile` | **DEAD** |
| `internal/extractors` (new) | **DEAD** |
| `internal/registry/…6735` | ALIVE before this PR → **DEAD** (new twin test, below) |
| `internal/entkinds` ×2, `internal/relkinds` ×2, `internal/registry/…6178` | **DEAD** on pre-existing tests |

**Vacuity control:** raising the lower bound's floor to 100000 reports
`found only 219 directory names`, so the floor assertion and the walk both fire.

### The one remaining call-site gap, closed

`internal/registry/home_isolation_sweep_guard_6735_test.go`'s `walkTestFiles`
was the only ungraded exclusion left: neutralising it left the whole
`internal/registry` package GREEN, while the same mutant on the 6178 walk ~30
lines away was caught by the test #6842 added. `TestUnpinnedHomeSweepSkipsWorktreeCheckouts`
is that walk's twin, and the mutant is now DEAD.

## Corrections to the issue, the brief, and my own earlier report

1. **Eleven copies, not seven** (verified against `0ce393bc3`): nine carrying the
   full 7-name literal, plus `producer_kinds_test.go`'s 3-name set and
   `incremental_phasetrace_6199_test.go`'s 4-name set — **three different sets**.
   The issue's table listed seven. My own first report said nine; that was also
   wrong, because I enumerated over the issue's file list instead of over the
   literal. Enumerating the literal is what surfaced the second live bug.
2. **Two live omissions, not one** — see §2. The extractors one is unconditional.
3. **`.claude` in `producer_entity_kinds_6776_test.go` and in
   `name_chosen_open_sweep_guard_6478_test.go` is dead code**: both root under
   `internal/` (and `cmd/`), where `.claude` is unreachable. They counted as
   "yes" in the survey from shape alone.
4. **`producer_kinds_test.go`'s narrower list is not a deliberate semantic
   choice** as the issue framed it — the walk is rooted differently.
   *But my first draft's justification for that was itself false*: I wrote that
   `dist` is "unreachable" under `internal/`, and `internal/dashboard/dist` is
   tracked and reachable. It holds no `.go` today, so behaviour is unaffected;
   the comment now records what was checked, per name, including that `dist` is
   a latent gap rather than an unreachable one.
5. **My "known limits" item was wrong in the favourable direction.** `if false &&`
   is DEAD at both `entkinds` sites, both `relkinds` sites and `registry/…6178`.
   The real gap was exactly one site (`registry/…6735`), and it is now closed.
6. **Layer 3 of the vacuity checklist was over-claimed.** Layers 1, 2, 4 and 5
   are killed by real mutants; layer 3 (full-content reading) is satisfied by
   construction only — the offence sits on the last line of a 3-line file, but
   nothing mutates the reader. The comment now says so.
7. **`SkippedDir` takes a base name**: `SkippedDir("a/b/.claude")` is false. All
   seven call sites pass `fs.DirEntry.Name()`, so this is latent, and now
   documented on the function.

## Verification

- `go test -count=1` green: `./internal/{repowalk,atomicfile,registry,entkinds,relkinds,types,safeio,extractors,verify}/`
- `gofmt -l internal cmd tools` — empty. `go vet ./...` — clean. `go build ./...` — clean.
- `go run ./tools/coverage validate` — `ok: 822 record(s) … 1163 test(s) checked`.
- `go mod tidy -diff` — clean.
- `./cmd/grafel` not re-run: the only non-test change is the new leaf package;
  `go list -deps ./cmd/grafel` does not reach `entkinds`, `relkinds` or
  `repowalk`, and the `internal/extractors` change is test-only.

### DIFF
--- a/internal/atomicfile/guard_6018_test.go
+++ b/internal/atomicfile/guard_6018_test.go
@@ -46,12 +46,15 @@ import (
 	"go/parser"
 	"go/token"
 	"io/fs"
+	"os"
 	"path/filepath"
 	"runtime"
 	"sort"
 	"strconv"
 	"strings"
 	"testing"
+
+	"github.com/cajasmota/grafel/internal/repowalk"
 )
 
 // notYetConverted lists the files that STILL build a deterministic temp name,
@@ -431,8 +434,10 @@ func scanDeterministicTempNames(t *testing.T, root string) []tempNameOffence {
 			return err
 		}
 		if d.IsDir() {
-			switch d.Name() {
-			case ".git", "node_modules", "vendor", "testdata", "dist", "build":
+			// The exclusion list is shared (#6846): seven hand-maintained
+			// copies had already drifted apart, and #6842 fixed exactly one
+			// of them. internal/repowalk states why each name is on it.
+			if repowalk.SkippedDir(d.Name()) {
 				return filepath.SkipDir
 			}
 			return nil
@@ -474,4 +479,82 @@ func repoRoot(t *testing.T) string {
 	}
 	// here = <root>/internal/atomicfile/guard_6018_test.go
 	return filepath.Clean(filepath.Join(filepath.Dir(here), "..", ".."))
+}
+
+// ---------------------------------------------------------------------------
+// #6846 — the walk must not descend into agent worktrees
+// ---------------------------------------------------------------------------
+
+// writeGoFile writes body to root/rel, creating parents.
+func writeGoFile(t *testing.T, root, rel, body string) string {
+	t.Helper()
+	path := filepath.Join(root, filepath.FromSlash(rel))
+	if err := os.MkdirAll(filepath.Dir(path), 0o755); err != nil {
+		t.Fatalf("mkdir for %s: %v", rel, err)
+	}
+	if err := os.WriteFile(path, []byte(body), 0o600); err != nil {
+		t.Fatalf("write %s: %v", rel, err)
+	}
+	return path
+}
+
+// offenderSource is a minimal non-test Go file carrying one of the three
+// spellings the guard catches, so a walk that reaches it MUST report it.
+const offenderSource = `package p
+
+func tmpFor(dest string) string { return dest + ".tmp" }
+`
+
+// TestScanDeterministicTempNames_DoesNotDescendIntoAgentWorktrees pins the
+// #6846 fix AT THE CALL SITE.
+//
+// .claude/worktrees/ holds full checkouts of this same repository. Before this
+// fix the guard's walk parsed every one of them, so in any development
+// checkout it reported offences under paths the notYetConverted ledger can
+// never name, and t.Fatalf'd on any mid-edit parse error in an unrelated
+// in-flight branch. CI has no worktrees there, which is why it survived from
+// #6018 through #6842 — so this test builds the condition itself under
+// t.TempDir() rather than relying on the ambient checkout, which would make it
+// pass vacuously exactly where it is needed least.
+//
+// The tree exercises all five vacuity layers at once: the scan must read
+// something (the ordinary package), read the RIGHT thing (that offence and no
+// other), read its full content (the offence is on the file's last line), the
+// matcher must fire, and the walk must ACT on the exclusion — a walk that
+// descends both trips the broken file's parse Fatalf and reports the shadow
+// offence.
+func TestScanDeterministicTempNames_DoesNotDescendIntoAgentWorktrees(t *testing.T) {
+	root := t.TempDir()
+
+	// Ordinary source: MUST be reported. Without this the test would pass on a
+	// walk that reads nothing at all.
+	writeGoFile(t, root, "internal/pkg/writer.go", offenderSource)
+
+	// An agent worktree: a full second copy of the tree. Its offence must NOT
+	// be reported, and its mid-edit parse error must not fail this package.
+	writeGoFile(t, root, ".claude/worktrees/agent-x/internal/pkg/writer.go", offenderSource)
+	writeGoFile(t, root, ".claude/worktrees/agent-x/internal/pkg/broken.go", "package p\n\nfunc (\n")
+
+	// A directory that merely CONTAINS "claude" is ordinary source and must
+	// still be scanned — the exclusion is an exact base-name match, not a
+	// substring one.
+	writeGoFile(t, root, ".claude-backup/internal/pkg/writer.go", offenderSource)
+
+	var got []string
+	for _, o := range scanDeterministicTempNames(t, root) {
+		got = append(got, o.file)
+	}
+	sort.Strings(got)
+
+	want := []string{".claude-backup/internal/pkg/writer.go", "internal/pkg/writer.go"}
+	if len(got) != len(want) {
+		t.Fatalf("scan reported %v; want exactly %v.\n"+
+			"An extra entry under .claude/worktrees/ means the walk descended into an agent "+
+			"worktree (#6846); a missing entry means it stopped reading real source.", got, want)
+	}
+	for i := range want {
+		if got[i] != want[i] {
+			t.Fatalf("scan reported %v; want exactly %v", got, want)
+		}
+	}
 }
--- a/internal/atomicfile/guard_6018_test.go
+++ b/internal/atomicfile/guard_6018_test.go
@@ -434,9 +434,12 @@ func scanDeterministicTempNames(t *testing.T, root string) []tempNameOffence {
 			return err
 		}
 		if d.IsDir() {
-			// The exclusion list is shared (#6846): seven hand-maintained
-			// copies had already drifted apart, and #6842 fixed exactly one
-			// of them. internal/repowalk states why each name is on it.
+			// The exclusion list is shared (#6846): ELEVEN hand-maintained
+			// copies had already drifted into three different sets, TWO of
+			// them missing `.claude` outright, and #6842 fixed exactly one by
+			// hand. Seven share this list now; internal/repowalk states why
+			// each name is on it, and which four walks deliberately keep
+			// their own copy.
 			if repowalk.SkippedDir(d.Name()) {
 				return filepath.SkipDir
 			}
@@ -517,12 +520,17 @@ func tmpFor(dest string) string { return dest + ".tmp" }
 // t.TempDir() rather than relying on the ambient checkout, which would make it
 // pass vacuously exactly where it is needed least.
 //
-// The tree exercises all five vacuity layers at once: the scan must read
-// something (the ordinary package), read the RIGHT thing (that offence and no
-// other), read its full content (the offence is on the file's last line), the
-// matcher must fire, and the walk must ACT on the exclusion — a walk that
-// descends both trips the broken file's parse Fatalf and reports the shadow
-// offence.
+// The tree covers four of the five vacuity layers WITH OBSERVED MUTANTS —
+// layer 1, the scan must read something; layer 2, it must read the RIGHT thing
+// (that offence and no other, which is why the assertion is an exact set and
+// not a floor); layer 4, the matcher must fire; layer 5, the walk must ACT on
+// the exclusion, since a walk that descends both trips the broken file's parse
+// Fatalf and reports the shadow offence.
+//
+// Layer 3 — reading a file's FULL content — is satisfied only by construction
+// here: the offence sits on the last line of a three-line file, so a truncating
+// reader would miss it, but nothing in this package mutates the reader, so that
+// is an argument rather than a measurement. Do not read this test as grading it.
 func TestScanDeterministicTempNames_DoesNotDescendIntoAgentWorktrees(t *testing.T) {
 	root := t.TempDir()
 
--- a/internal/entkinds/rule_declared_kinds_sweep_guard_6744_test.go
+++ b/internal/entkinds/rule_declared_kinds_sweep_guard_6744_test.go
@@ -418,6 +418,22 @@ func TestRepoSweepIsNotVacuous(t *testing.T) {
 // countSourceFiles is this test's OWN walk of the tree, deliberately written
 // out rather than delegating to the package under test: a floor that asks the
 // scanner how many files the scanner read is not a floor.
+//
+// For the same reason this list is NOT internal/repowalk.SkippedDir (#6846),
+// even though it is character-for-character the same set and entkinds.Scan now
+// uses that shared copy. Sharing it here would put one predicate on both sides
+// of the equality above. Measured on the twin of this test in internal/relkinds,
+// widening the shared list with "security","atomicfile","registry":
+//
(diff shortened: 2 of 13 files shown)


cajasmota:
## Coordinator verification — and the count was wrong three times

I filed this as **seven** copies of the exclusion list. The author's first pass said **nine**. Enumerating the *literal* rather than the issue's file list gives **eleven, in three different sets** — and the author named their own error correctly: the nine came from grepping only the six files the issue happened to mention, which is the hand-picking failure the brief warns about.

That corrected enumeration is what found the second bug.

## A second live `.claude` omission, worse than the one this issue was filed for

`internal/extractors/incremental_phasetrace_6199_test.go` roots at the **module root** with only `.git, node_modules, testdata, vendor`. `TestPhaseTrace_EveryReturnIsTraced` asserts `tryIncremental` has exactly **one** call site — and every worktree contains its own `incremental.go`:

```
found 2: [{shadowCaller …/.claude/worktrees/fake-agent/…/shadow.go:3:27}
          {TryIncremental …/internal/extractors/incremental.go:427:8}]
```

**No parse error required.** Unlike the `atomicfile` and `registry` cases, this one fails on **every local run** in any checkout with worktrees, and has since #6199. CI never sees it — there are no worktrees there. Fixed with a `t.TempDir()` regression test; the call-site mutant is DEAD.

## My killing mutant, which nobody else ran: the *narrowing* direction

Every mutant scored so far attacked **widening** — the list growing to hide directories a guard should inspect. Nobody had asked whether *removing* a needed exclusion is caught. I dropped `testdata` from the shared predicate:

**DEAD in four packages** — `TestSkippedDir_SkipsEveryForeignTree`, `TestRepoSweepIsNotVacuous`, `TestScanSkipsWorktreeCheckouts`, `TestScanGo_SkipsExcludedDirs`. So the change grades **both** directions: widening by the new lower bound, narrowing by the per-package tests that were already there.

## The blocking finds, both reproduced by the author before fixing

- **The lower bound's headline claim was false.** Its domain was base names of directories *under `internal/`* holding a non-test `.go` file — so widening with `cmd, tools` was ALIVE, and widening with **`internal`** (the single most destructive widening available) was ALIVE **in the very test built to catch widening**. Now rooted at `repoRoot` and collecting every **path component** on the way to a source file (219 names), which is what makes `internal`/`cmd`/`tools` nameable at all — they hold no `.go` directly, so a parent-only domain could never have covered them. The prune stays a hard-coded literal, so the self-vacuity property survives.
- **A new comment asserted `dist` was unreachable.** `internal/dashboard/dist` is tracked and the walk descends into it. Now recorded as a **latent gap**, per name, rather than an unreachable one.

## The strongest argument in the change, demonstrated rather than asserted

Three of the eleven walks were **deliberately not shared** because they are *independent floors* against production scanners. Widening `SkippedDir` with three carefully-sized names and running `relkinds.TestRepoSweepIsNotVacuous` both ways:

- independent walk → **FAIL**: `scan parsed 2079 non-test .go files; an independent walk … finds 2086`
- shared predicate → **PASS**, with **7 real production Go files unread by both sides**

One predicate on both sides of an equality, with receipts. Qualifier worth keeping: an absolute `wantGo < 1000 || wantYAML < 500` floor backstops it, so only widenings small enough to clear that floor exhibit the failure — it had to be constructed deliberately.

## Also closed

The one genuinely ungraded exclusion call site (`registry/…6735`'s `walkTestFiles`) now has a test; **no ungraded exclusion call sites remain**. My earlier restatement that only `atomicfile` had call-site coverage relayed the author's error — three of those four sites were already graded by pre-existing tests.

## Corrections to my own framing

- **"Seven copies"** → eleven. All three call-site comments carried my number; now fixed.
- **"Two mutants ALIVE"** was stale by the time I wrote it — both were DEAD, but the *class* was not closed, which is exactly what the `cmd, tools` and `internal` widenings showed.
- My question about `producer_kinds_test.go`'s "deliberately narrower" list was **malformed** — different root, so the omitted names are unreachable rather than missing.
- One correction *to the review*: `internal/safeio/…6478` shares the list, but its `.claude` case was **already dead** (it roots at `root/internal` and `root/cmd`). Sharing there is tidying, not a third bug, and the PR body says so.

`cmd/grafel`'s digest is not in play: no non-test file under `internal/extractors/` changed, and `go list -deps ./cmd/grafel` does not reach `repowalk`, `entkinds` or `relkinds` — verified rather than relayed.

PR #6853, opened 2026-09-05
test(#6849): grade AllDispositions against the Disposition enum — the slice is complete today, and the sanity check that looked like it caught this cannot

### LINKED ISSUE #6849: resolve.AllDispositions is a hand-maintained slice nothing grades against the Disposition enum — a disposition can exist and be silently absent from every report
`resolve.AllDispositions` is a hand-maintained slice that nothing grades against the `Disposition` enum — a disposition can exist and be silently absent from every report

## The gap

#6836's fix drives both the report's **rows** and its **counts** from `resolve.AllDispositions`, so a disposition can no longer exist in the taxonomy while being absent from or unfed by the table. That is the right shape — but the guarantee holds only relative to `AllDispositions` itself, and **`AllDispositions` is a hand-maintained slice that nothing checks against the enum it is supposed to enumerate.**

Proved during the round-2 review of #6836: deleting `DispositionExternalSQL` from the slice removes its row **and silently drops that bucket from the denominator**, with `./internal/resolve` and `./internal/feedback` both green.

`Disposition.String()` even carries an `"unknown"` fallback for values outside the known set, which is exactly the escape hatch that lets an unenumerated value pass unnoticed.

The percentages are normalised over the sum of the enumerated buckets. So dropping a row does not merely hide it — it **redistributes its share across the survivors**, making every other number larger and still internally consistent. A reader gets a table that sums to 100% and is wrong, with nothing anywhere reporting a discrepancy.

That is strictly worse than the defect #6836 was filed for. A permanent `0.00%` row is at least visible; a silently-absent one changes the numbers a reader is tracking and leaves no trace.

## This is #6836 one level up

#6836: three counters existed in the code and were never fed, so three rows were permanently zero.
This: a disposition can exist in the enum and never reach the table at all.

Both are "the taxonomy and the thing reporting it are coupled by hand". #6836's fix removed the hand-coupling between *rows* and *counts*; it did not remove the hand-coupling between *the enum* and `AllDispositions`.

## The fix

A test asserting `AllDispositions` covers the `Disposition` enum exactly — no member missing, no extra. The usual way here is a source-level scan of the const block compared against the slice, the same shape `internal/types`' entity-kind guards use.

**Score it with a mutant that removes one member and confirm it dies alone** — a check that only fires alongside something else would be the masking pattern this batch has hit repeatedly. And per #6834, apply the vacuity layers to the new guard itself: a source scan that reads nothing reports a clean tree.

## Not established

- **Whether `AllDispositions` is currently complete.** The round-2 review confirmed `external-sql` and `unclassified` are in it and reachable, but nobody has diffed the full slice against the full enum. **That measurement should come first** — a guard added green tells you nothing, and if it is already incomplete that is a live reporting bug rather than a latent one.
- Whether any other consumer of `AllDispositions` exists outside `internal/feedback`. Not surveyed.
- Whether the `"unknown"` fallback in `Disposition.String()` is load-bearing anywhere, or vestigial.

Found during the independent round-2 review of #6836; pre-existing in the sense that the slice predates that change, though the fix is what made the report depend on it. Related: **#6834** (guards that report success while checking nothing) and **#6830** (`AllEntityKinds()` duplicates undetected) — the same set-vs-slice, enum-vs-list asymmetry in a third place.

cajasmota:
# Round 3 addendum — grading the widening itself

*(Appends to the round-2 PR body. Everything below is re-derived after the restructure; round-2 verdicts do not carry.)*

## The defect: the widening was ungraded

Round 2 replaced a single-file scan with a package-directory scan, which killed P4. But **nothing observed that the scan was still package-wide.** Narrowing the filter back to `refs.go` — a plausible edit — silently restored the file-scoped shape #6849 is about.

Reproduced exactly as reported:

```
filter → !strings.HasSuffix(name, "refs.go")
zz_coordmut_extra.go → const DispositionCoordProbe Disposition = 8
go test ./internal/resolve/   →  ok      ← ALIVE
```

**Positive control** (same planted constant, filter unmutated): `--- FAIL: TestAllDispositions_CoversTheEnumExactly` + `--- FAIL: TestDispositionString_HasALabelForEveryDeclaredConstant`. The guard works; only its input scope was unobserved.

Layer 2 for a *directory* scan is not "did it read `refs.go`" — that is what M6b covered. It is "did it read **the rest of the package**", which is the entire content of the widening.

## The fix: must-scan anchors

`readPackageGoSources` now calls `assertScannedThePackage`, following `internal/links/source_read_guard_6823_test.go`. Failures say **"the scan is broken, not the enum"** — deliberately the opposite diagnosis from "a constant is missing".

**The floor is measured, not eyeballed.** Pointed at `100000` once, the scan reported:

```
the scan is broken, not the enum: it read only 47 non-test .go file(s) from ., want at least 100000
```

**Measured file count: 47.** Floor set to **30**, documented as deliberately slack — it catches a filter that *collapses* the scan, not the file count, which churns.

**A count floor alone is provably not enough here, and this package is a sharp instance of the 35-vs-26 trap:** `internal/resolve` has **61 `_test.go` files against 47 non-test ones**. A filter inverted to test files reads *more* files than the real scan and clears **any** floor the real scan could satisfy. Only the named anchors catch it — confirmed by the recorded failure message, not assumed (see N2).

Anchors (file + tokens that must appear in the bytes actually collected): `refs.go` / `imports.go` / `symbol_index.go` / `stdlib_builtins.go`.

The standalone `refs.go` check in the non-vacuity test is removed as redundant — the shared reader now enforces four files for every caller of the scan.

## Scoring (round 3)

`go vet` before every score; `-count=1`; packages scored separately; exact-count replace aborting on ≠1; backups restored by `cp` with shasums verified at the end (`refs.go eabe9ab7…`, guard `d1b9c8f4…`, both matched; planted file confirmed absent).

### The two new mutants

| # | Mutant | `./internal/resolve` | `./internal/feedback` | Verdict |
|---|---|---|---|---|
| **N1** | filter narrowed to `refs.go`, **with** the planted constant | NonVacuous + `CoversTheEnumExactly` + `String_HasALabel…` | green | **ALIVE before → DEAD** — *"read only 1 non-test .go file(s) from ., want at least 30"* |
| **N1b** | filter narrowed to `refs.go`, **clean tree, no planted constant** | same three | green | **DEAD** |
| **N2** | filter inverted to `_test.go` only | same three | green | **DEAD via the ANCHORS, not the floor** — *"refs.go was never read from .. The walk read **61** file(s)"* |

**N1b is the stronger result and worth calling out.** The narrowing dies on a *clean tree*, with no defect planted. That means the widening is graded **on its own**, not merely when a defect happens to coincide with it — so this cannot regress into a check that only fires alongside something else.

**N2 confirms the trap empirically rather than by argument:** 61 > 30, the floor is cleared, and the *anchor* is what fires. Had I set the floor to the measured 47, it would still have been cleared.

### Re-scored (verdicts do not carry across a restructure)

| # | Mutant | resolve | feedback | Verdict |
|---|---|---|---|---|
| P4 | `const DispositionArchived Disposition = 8` in a new file | `CoversTheEnumExactly` + `String_HasALabel…` | green | DEAD |
| P4b | same, value `3` (aliases `ExternalSQL`) | same two | green | DEAD |
| M9 | literal evaluator neutered + P4 | 4 tests, via the **refusal** path | green | DEAD |
| M1 ×8 | remove each member individually | `CoversTheEnumExactly` **alone, all 8** | incumbents | DEAD 8/8, solo in-package |
| M2 | add `Disposition(99)` | `CoversTheEnumExactly` alone | **green** | DEAD — only test in the tree |
| M3 | duplicate a member | `CoversTheEnumExactly` alone | 4 incumbents | DEAD |
| M4 | **reorder** (control) | green | green | **ALIVE, intended** |
| M5 | delete a `String()` case | `String_HasALabel…` alone | 2 incumbents | DEAD |
| M6b | scan `../feedback` (real parseable Go) | 3 tests | green | DEAD — **note: now fires on the FLOOR** (*"read only 5 non-test .go file(s) from ../feedback"*), where in round 2 it fired on the missing anchor. Recorded because the assertion that fires changed, even though the verdict did not. |
| M7 | truncate every read to 300 bytes | 3 tests | green | DEAD |
| M8 | call site `false &&` + real defect | **green** | incumbents | layer-5 positive control **PASSED** |

## Known limitation, recorded in the code

The anchors pin **four** files, so a filter that excluded some *other* single non-anchor file would still clear both the floor and the list. That is inherent to the pattern — the same is true of `internal/links/source_read_guard_6823_test.go`, which this follows. It catches a scan that **collapses**, not one that loses one file. This is stated in the `dispositionMustScan` doc block rather than left for the next reviewer to rediscover, the same way the sentinel limitation was.

## Verification

- `go test -count=1 ./internal/resolve/ ./internal/feedback/` — ok
- `go build ./...` — ok
- `gofmt -l internal/resolve/ internal/feedback/` — empty
- `go run ./tools/coverage validate` — exit 0 (`822 record(s)`; warnings pre-existing)
- Planted files (`zz_coordmut_extra.go`, `zz_mutant_extra_6849.go`) deleted; `git status` shows only the intended edits.

---

Closes #6849

The slice and the enum agree today, so this ships as a **ratchet**, not a fix for a live
divergence — the measurement is on the issue. What it does close is the gap that made the
divergence undetectable: nothing graded `AllDispositions` against the `Disposition` enum,
and `resolution-vector-sums-to-100pct` is algebraically incapable of noticing a dropped
member in production.

### DIFF
--- a/internal/feedback/render.go
+++ b/internal/feedback/render.go
@@ -200,9 +200,10 @@ func Render(w io.Writer, r *Report) error {
 		// here: before #6836 three of six rows were wired to counters nothing
 		// incremented and rendered a permanent 0.00%. Iterating the taxonomy
 		// removes that failure mode for every disposition the slice names —
-		// each gets a row, and each row is fed by the classifier. It does NOT
-		// cover a Disposition that reaches the enum without reaching
-		// AllDispositions, which nothing currently checks.
+		// each gets a row, and each row is fed by the classifier. A Disposition
+		// that reaches the enum without reaching AllDispositions is covered
+		// too, since #6849 grades the slice against the enum at its root.
+		// Row ORDER follows the slice; the numbers do not depend on it.
 		for _, d := range resolve.AllDispositions {
 			fmt.Fprintf(w, "| %s | %.2f%% |\n", d.String(), rv.Pct(d.String()))
 		}
--- a/internal/feedback/report.go
+++ b/internal/feedback/report.go
@@ -51,9 +51,10 @@ type KindStats struct {
 // The breakdown is produced by the SAME classifier the indexer uses —
 // resolve.Index.ClassifyEndpoints — not by a private re-derivation, so the
 // report does not drift from the taxonomy it claims to report. That guarantee
-// is only as strong as resolve.AllDispositions being complete: it is a
-// hand-maintained slice, graded against nothing, so a Disposition added to the
-// enum and not to the slice would still go missing here. #6836: three of
+// is only as strong as resolve.AllDispositions being complete, and since #6849
+// that completeness is itself graded: TestAllDispositions_CoversTheEnumExactly
+// in internal/resolve scans the package for Disposition constants and fails if
+// the slice omits, duplicates, aliases or invents one. #6836: three of
 // six hand-maintained counters here had no increment and rendered a permanent
 // 0.00%; keying off resolve.AllDispositions removes the whole defect class,
 // because a disposition can no longer exist in the taxonomy and be silently
@@ -756,9 +757,9 @@ func Generate(_ context.Context, docs []*graph.Document, opts Opts) (*Report, er
 	// Resolution disposition vector, computed by the indexer's own classifier
 	// rather than re-derived here (#6836). One entry per resolve.AllDispositions
 	// member — a 0.00% is corpus-relative, never a disposition nothing counts.
-	// Caveat: AllDispositions is a hand-maintained slice and nothing grades it
-	// against the Disposition enum, so completeness of the table still rests on
-	// that slice staying in step.
+	// AllDispositions is hand-maintained, but since #6849 it is graded against
+	// the Disposition enum by TestAllDispositions_CoversTheEnumExactly in
+	// internal/resolve, so a disposition cannot exist and be missing here.
 	//
 	// Only ToID is populated on each EndpointPair, so ClassifyEndpoints
 	// records exactly one disposition per edge with a target and the counts
--- a/internal/feedback/sanity.go
+++ b/internal/feedback/sanity.go
@@ -306,10 +306,21 @@ func runSanityChecks(r *Report) ([]SanityResult, int) {
 	// 3. Resolution vector sums to 100% ± 0.1%.
 	if r.ResolutionTotal > 0 {
 		// Summed over resolve.AllDispositions rather than a hand-listed set
-		// (#6836): a disposition missing from the vector must fail this check
-		// instead of silently shrinking the sum's scope. The dispositions
-		// partition the edges that have a target, so anything but 100% means a
-		// bucket double-counted or dropped one.
+		// (#6836), so this check follows the taxonomy instead of drifting from
+		// it. It catches a vector whose percentages do not add up — e.g. one
+		// assembled by a caller from a different source than report.go's.
+		//
+		// What it does NOT catch, despite an earlier version of this comment
+		// claiming otherwise (#6849): a disposition MISSING FROM
+		// resolve.AllDispositions. report.go:768-779 derives the denominator by
+		// summing DispositionCounts over that same slice and then divides each
+		// bucket by it, so this loop computes Σ(100·c_d / Σc_d) over an
+		// identical index set — exactly 100 for ANY subset. A duplicated member
+		// is invisible too: the double-count in the denominator and the
+		// double-iteration here cancel. Dropping a disposition therefore
+		// redistributes its share across the survivors and still sums to 100%,
+		// silently. Completeness of the slice is enforced at its root instead,
+		// by TestAllDispositions_CoversTheEnumExactly in internal/resolve.
 		sum := 0.0
 		for _, d := range resolve.AllDispositions {
 			sum += r.Resolution.Pct(d.String())
--- a/internal/resolve/disposition_enum_completeness_6849_test.go
+++ b/internal/resolve/disposition_enum_completeness_6849_test.go
@@ -6,62 +6,126 @@ import (
 	"go/parser"
 	"go/token"
 	"os"
+	"path/filepath"
 	"sort"
+	"strconv"
+	"strings"
 	"testing"
 )
 
-// dispositionSourceFile is the single file that declares BOTH the Disposition
-// const block and the AllDispositions slice. The guard below reads the const
-// block from source rather than from a hand-written roster of names — a
-// hand-maintained roster would be the very defect it is guarding against, one
-// level up (#6849).
-const dispositionSourceFile = "refs.go"
+// dispositionSourceDir is the package directory the guard scans. It is the
+// DIRECTORY, not a single file, on purpose.
+//
+// The first cut of this guard parsed only refs.go, which is where the
+// Disposition const block lives today. That made it a FILE-scoped roster
+// guarding a slice-scoped roster — a constant declared in any other file of
+// this package was invisible to it, so the exact defect #6849 exists to
+// prevent (a Disposition that exists and never reaches AllDispositions) could
+// still be introduced with the whole package green. Scanning the directory is
+// what makes the guard enum-scoped rather than file-scoped.
+const dispositionSourceDir = "."
 
-// dispositionConst is one constant of the `Disposition = iota` const block as
-// it is actually written in refs.go. Value is the constant's iota value, which
-// is the index of its ValueSpec within the const block — `_` placeholders and
-// specs of other types still consume an iota slot, so the index is taken over
-// every spec in the block, not only the ones this walk keeps.
+// dispositionAnchorFile is where the const block lives today. It is asserted to
+// be among the files actually read (a layer-2 pin), but it is NOT the limit of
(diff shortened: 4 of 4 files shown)


cajasmota:
## Coordinator verification — four rounds, and the agent disproved my last justification

**The measurement, done before the guard as asked: the slice and the enum AGREE today.** 8 constants, 8 entries, same set, no duplicates. **A ratchet, not a live bug** — worth knowing before anyone reads #6849 as an outage.

### My filed premise was wrong, and the real cause is a repeatable trap

I wrote that deleting `DispositionExternalSQL` leaves `./internal/resolve` and `./internal/feedback` **both green**. Removing **each of the 8 in turn** already fails an incumbent `internal/feedback` test.

The confound is worth recording precisely, because it repeats easily: **`go test ./internal/resolve/` alone IS green for all 8**, because every incumbent killer lives in `internal/feedback`. "Green in resolve" became "both green". Not a `-run` slip — a **package-selection** slip. The rule we apply to mutants (*prove the filter selected something*) applies just as much to the runs that establish a defect exists in the first place.

### What survives is sharper than what I filed

`report.go` derives `total` by summing over `AllDispositions`, `byDisposition` divides by that same total over the same slice, and `sanity.go` sums `Pct()` over the slice again — so the result is Σ(100·c_d / Σc_d) over an identical index set: **exactly 100 for any subset**, and it stays 100 under a duplicated member (the double-count and double-iteration cancel).

So `resolution-vector-sums-to-100pct` is **structurally incapable of firing on this defect in production**. It only reddens under mutants because the sanity fixture is hand-built. Confirmed empirically: deleting `DispositionExternalUnknown` or `DispositionBugResolver` leaves `TestRunSanityChecks_AllPass` passing. Its comment claimed the opposite and is corrected here — leaving a disproved capability claim in the file we just disproved it about would be the defect class in miniature.

### My two killing mutants, and how each was answered

**1. The guard was file-scoped, not enum-scoped.** A `Disposition` constant in a *second file* of the package, absent from the slice — the exact defect #6849 exists to prevent — left the package green. Widened to scan the package directory (47 files, ~20 ms, so the cost argument did not materialise). That widening **forced something neither I nor the reviewer anticipated**: positional value inference becomes unsafe, since `Disposition = 8` read as "index in block" is value 0 and collides with `DispositionResolved` — the guard would have reported **success** on the very mutant it was built for. Values are now *evaluated*, with a **loud refusal** rather than a guess for anything non-literal.

**2. The widening was itself ungraded.** Narrowing the filter back to `refs.go` with the same planted constant: **ALIVE**, positive control confirming the unmutated guard fails. Closed with must-scan anchors. The 35-vs-26 trap is sharper here and was measured, not assumed: `internal/resolve` has **61 `_test.go` against 47 non-test**, so an inverted filter reads *more* than the real scan and clears **any** floor the real scan could satisfy.

**3. Then the anchor list could be emptied silently** — the same hole found on `internal/links`' guard. Closed with a floor on the anchor set.

### The agent disproved my justification for that last fix, then justified it properly

I claimed emptying the anchors re-opens the inverted-filter case. **It does not** — a test-file-only scan finds zero `Disposition` constants, so the zero-constants floor fires regardless. My prescribed fix would have shipped **ungraded**: no mutant in my set showed the anchors doing work nothing else did.

So the agent built the one that does: **`refs.go` + the 36 `dynamic_patterns_*.go` files** — 37 files clears the file floor, all 8 constants clear the zero-constants floor *and* the sentinels, and only the **anchors** catch it. Anchors emptied: **ALIVE pre-fix, DEAD post-fix.** A realistic shape — a filter that keeps most of the package — and the justifying mutant is written into the comment above the check.

That is "a guard added green tells you nothing" applied to my own instruction, which is exactly right.

### Limitations kept rather than papered over

- The anchors pin four files, so a filter excluding some *other* single non-anchor file clears both floor and list. Inherent to the pattern (`internal/links` shares it); stated in the doc block.
- `M4` (reordering) is **ALIVE and intended** — the guard is deliberately not order-tight. `render.go` and `cmd/grafel/index.go` do emit rows in slice order, so the earlier "every consumer sums or maps over it" justification was inaccurate and has been corrected.
- The value half of the sentinels can never fire alone; kept for error quality, and now honest about it.
- `"unknown"` in `String()` is **vestigial** — no int→`Disposition` conversion and no deserialisation anywhere in `internal/` or `cmd/`. Kept for totality; the "read back from disk" justification dropped, since it contradicted the vestigial finding.

### Blast radius, verified

The guard grades the **shared root**, so it protects every consumer: `internal/feedback/{report,render,sanity}.go`, `internal/resolve/refs.go:876,879`, `cmd/grafel/index.go:1438,1440,1452,1453,3057` (`disposition_counts`, `disposition_samples`, and the `grafel doctor` printer — which derives its own total by summing the slice, so it has the same structural 100% property), and `internal/extractors/vbnet/corpus_6327_test.go:202`.

`cmd/grafel` is not in play: **I verified myself** that the branch's own diff is test-only for `internal/resolve` and comment-only for `internal/feedback` (diff filtered to non-comment, non-blank lines: empty).

Issue #6852, opened 2026-09-05
extractors: path-anchored FromID with no file carrier — COMPLETE, 12 of 12 arms shipped (+dockerfile unledgered); the twelve was a lower bound


cajasmota:
Twelve extractors emit a path-anchored `IMPORTS` whose FROM end resolves to nothing — the same defect #6815 fixed in three

## The measurement

#6847 asked for the measurement that had never been run: does any currently-registered extractor emit a path-anchored `FromID` naming no emitted record? It does. Twelve of them.

The new runtime guard (`internal/extractors/file_anchored_carrier_runtime_6847_test.go`) drives every corpus file under a `testdata/` directory through the production classifier, the production `treesitter.ParserFactory` and the production `extractors.Extract`, then asks the exact resolution question: is there a record whose `Name` or `QualifiedName` is this file's path? Measured 2026-09-05 over 909 candidate files, 41 registered languages exercised:

| language | evidence |
|---|---|
| astro | `astro/extractor.go:328` `FromID: filePath` |
| bicep | `bicep/extractor.go:296` `FromID: path` |
| clojure | path-anchored `IMPORTS` |
| cobol | path-anchored `IMPORTS` (COPY members) |
| crystal | `crystal/extractor.go:161` `FromID: filePath` |
| dart | path-anchored `IMPORTS` |
| fsharp | `fsharp/extractor.go:673` `FromID: filePath` |
| html | path-anchored `IMPORTS` (script/link refs) |
| lua | `lua/lua.go:540` `FromID: file.Path` |
| shell | `shell/shell.go:260` `FromID: file.Path` |
| terraform | `hcl/relationships.go:160`, `:180` `FromID: path` |
| zig | `zig/zig.go:272` `FromID: filePath` |

Every one is `IMPORTS`. Independently confirmed against the source: **none of the twelve packages calls `extractor.FileEntity` or `extractor.PrependFileCarrier` anywhere in its non-test files.** The only three extractor packages in that set that do are erlang, nim and groovy — the three #6815 fixed. So the class was three of fifteen, not three of three.

[...]

PR #6855, opened 2026-09-05
test(#6847): a runtime guard for path-anchored FromIDs — the measurement came back RED, at least fifteen extractors violate the invariant

### LINKED ISSUE #6847: Nothing would catch a fourth extractor emitting a path-anchored IMPORTS with no file carrier — the guard that should skips IMPORTS by convention
Nothing would catch a fourth extractor emitting a path-anchored `IMPORTS` with no file carrier — the guard that should skips `IMPORTS` by convention

## The gap

#6815 fixed erlang, nim and groovy: each emitted `IMPORTS` relationships whose `FromID` was the source path, with **no record carrying that path**, so the edge reached the graph with a raw path string at its FROM end — a `from_id` that is not any node's id.

The guard that ought to catch this class is `internal/extractors/file_anchored_rels_guard_test.go`. It cannot:

- it **skips `IMPORTS`** by the #120 convention (`:635`, `:1060`);
- `knownInvisibleOffenders` is `{}` (`:506`).

So the three known instances are fixed and **no mechanism was added that would catch the fourth**. The next extractor to emit a path-anchored `IMPORTS` without a carrier reproduces the defect silently, and — as #6815 established — recall-shaped assertions cannot see it, because they only ever ask whether the carrier *exists*, never whether an edge's FROM end resolves.

## Why the existing skip is not obviously wrong

The #120 convention presumably had a reason, and this issue is **not** a claim that `IMPORTS` should simply be un-skipped. Un-skipping it wholesale would likely light up every extractor that legitimately anchors `IMPORTS` on something other than a path. **Establishing whether that is true is the first task here**, and it has not been done.

The narrower, safer shape: a table-driven check over the registered extractors asserting that **every path-anchored `FromID` names an emitted record** — which is the actual invariant, independent of relationship kind. That does not require deciding anything about `IMPORTS` specifically.

## Why it is worth doing

#6815's own evidence is that the two existing layers each catch what the other misses, measured rather than assumed:

- an over-broad anchor predicate (`FromID != ""` instead of `== path`) was **invisible to all three golden fixtures** and survived only into the unit suite;
- a wrong `subtype` stamp was caught by the golden totals reverting exactly to main's, and by the #6481 placeholder pruner.

Neither layer is a structural guard over *all* extractors, though — both are per-language. That is the hole: the invariant is repo-wide, the checks are local, and the class was found by inspection rather than by anything failing.

## Not established

- **Whether any currently-registered extractor already violates the invariant** in a kind other than `IMPORTS`. Nobody has run the check, because the check does not exist. That measurement is most of the value here and should come *before* any fix — a guard added green tells you nothing.
- **Why #120 skipped `IMPORTS`.** Read the issue before changing the convention; there may be a good reason that the narrower invariant above sidesteps entirely.
- Whether `knownInvisibleOffenders` being empty is deliberate (nothing to defer) or vestigial.

Found during the independent review of #6815; pre-existing and explicitly out of that issue's scope. Related: **#6834** (guards that report success while checking nothing) — a guard that skips the kind carrying the defect is that class by a different route.

cajasmota:
# test(#6847): the invariant does NOT hold — at least 15 registered extractors emit a path-anchored IMPORTS with no carrier

Closes #6847. Refs #6852 (the live defects this found), #6815, #120, #6834, #6849.

## THE MEASUREMENT CAME FIRST, AND IT IS RED

The issue asked whether any currently-registered extractor already emits a path-anchored `FromID` naming no emitted record. **Twelve do, as measured by this corpus — and twelve is a lower bound, not a population.**

907 candidate files (every file under a `testdata/` directory) driven through the production classifier → production `treesitter.ParserFactory` → production `extractors.Extract`. 41 registered languages exercised, 34 producing at least one path-anchored edge. Offenders, all on `IMPORTS`:

```
astro bicep clojure cobol crystal dart fsharp html lua shell terraform zig
```

Independent review then drove the registered languages this corpus **cannot reach** from hand-written minimal sources and found three more — `dockerfile` (`dockerfile/dockerfile.go:347`), `reasonml` (`reasonml/extractor.go:325`), `rescript` (`rescript/extractor.go:296`). So the known total is **≥15 offenders / 18 instances** counting the three #6815 fixed, and it is unbounded above until every registered language is driven.

Confirmed against the source, not inferred from the run: **none of the fifteen packages calls `extractor.FileEntity` or `extractor.PrependFileCarrier` anywhere in its non-test files.** The only extractor packages that do are erlang, nim and groovy — exactly the three #6815 fixed. `refs.go` has no path→entity index, so every one of these edges reaches the graph with a raw path at its FROM end.

Nothing is fixed here. Each needs the per-language conditional-carrier judgement #6815/#6518 made (an unconditional carrier mints one bare orphan node per source file), and each moves the golden fixtures and `cmd/grafel`'s whole-graph digest. Tracked in **#6852**.

## What #120 actually says, and why it does not change the design

#120 is *"Java cross-file receiver binding (follow-up to #105)"*. The convention lives in its closing implementation comment, not its body: **"Source-file-path FromIDs (every IMPORTS edge across every language) tagged Dynamic instead of bug-extractor."**

That is a decision about a **metric** — such an edge stops counting as an extractor bug — not a claim that the edge resolves. It does not. #6815 treated exactly this shape as a defect and fixed it, so the repo's own current position is that a dangling path-anchored `IMPORTS` is a bug whose metric is merely suppressed.

Design consequence: the source scan's `IMPORTS` skip is **not** touched. This guard never asks what kind an edge is, only whether its FROM end resolves — the actual invariant — so the `IMPORTS` question is sidestepped rather than reopened.

## The corpus gap is named, not implied

The neighbouring source scan's rationale — a corpus "grows only when someone remembers to grow it", while a source scan covers a new extractor the day it lands — is **vindicated** by `dockerfile`/`reasonml`/`rescript`, not superseded. The two checks are complements; neither replaces the other. The file says so, and further records:

- `exercisedLanguages6847` is **hand-maintained and NOT derived from `extractor.List()`**, stated plainly because a slice that relates to nothing outside itself makes every omission read as deliberate (#6849's pattern). A `List()`-relative pin is not available: `List()` returns **445** names, ~380 of them framework extractors (`custom_*`, `python_*`, `lua_*`) that no classifier language dispatches.
- The **24 unexercised base languages** are listed by name, three of them confirmed offenders.
- `noExtractorLanguages6847` (`objective_c`, `prisma`) pins the languages the corpus **does** classify but no extractor handles — an external relation, so registering an extractor for either turns the omission red instead of silent. Graded by mutant MN.
- **`dockerfile` is not a missing fixture.** Three corpus files exist — `testdata/fixtures/dockerfile/sample.Dockerfile`, `testdata/fixtures/sources/dockerfile/sample.Dockerfile`, `testdata/fixtures/real-world/docker/Dockerfile.multi_stage` — and all three classify `lang="" skip=true reason="unsupported_extension"`: the classifier's basename table maps the exact name `Dockerfile`, not the `*.Dockerfile` / `Dockerfile.<variant>` spellings the fixtures use. The extractor is registered and the fixtures exist; the classifier is what the corpus cannot cross.

## The guard

`internal/extractors/file_anchored_carrier_runtime_6847_test.go`:

1. `TestFileAnchoredCarrier_NoNewOffender_6847` — the ratchet. Offending-language set must equal `knownMissingCarrier6847` **exactly, in both directions**: a thirteenth corpus-visible offender fails, and so does fixing one of the twelve without deleting its line. A floor would have been blind to the permissive direction.
2. `TestFileAnchoredCarrier_CorpusCoverage_6847` — vacuity layers 1-3 plus the no-extractor pin.
3. `TestPathAnchoredKindsMissingCarrier_Detects_6847` — the matcher's positive control, 9 cases.

### #6834's five vacuity layers, graded not claimed

| layer | mechanism | graded by |
|---|---|---|
| 1 read anything | `minCorpusFiles6847` floor | **nothing — M10 SURVIVES**, recorded in the source rather than papered over |
| 2 the RIGHT things | `exercisedLanguages6847` (41), `noExtractorLanguages6847` (2) | M7 DEAD, MN DEAD |
| 3 their FULL content | `anchoringLanguages6847` (34) | M8 DEAD |
| 4 can the matcher detect | 9-case unit table | M1, M2, M3, M9, MJ DEAD |
| 5 does the CALL SITE act | the ledger is produced by the walk | M4, M5, M6 DEAD — each with the unit test still PASSING |

### Mutants — 12, scored individually against the final file, `go vet` first, `-count=1`

| # | mutant | direction | verdict | killed by |
|---|---|---|---|---|
| M1 | carrier test → "any record is a carrier" | permissive | **DEAD** | walk + unit table |
| M2 | anchor `FromID == path` → `!= ""` (#6815's own escaped predicate) | permissive | **DEAD** | walk, **forward** branch (`language "sql"`) |
| M3 | matcher returns nil unconditionally | permissive | **DEAD** | walk + unit table |
| M4 | call site never records a finding | permissive | **DEAD** | walk only — unit table PASSED |
| M5 | call site passes the unclassified path | permissive | **DEAD** | walk only — unit table PASSED |
| M6 | detection lost for exactly one language (zig) | permissive | **DEAD** | walk, **reverse** branch |
| M7 | corpus shrinks (repo-root `testdata/` dropped) | permissive | **DEAD** | exercised + anchoring sets |
| M8 | every file truncated to 120 bytes | permissive | **DEAD** | anchoring set (13 languages) |
| M9 | kinds reverse-sorted (deterministic, not racy map order) | strict | **DEAD** | ordered comparison |
| M10 | `minCorpusFiles6847` 700 → 0 | permissive | **ALIVE — equivalent** | nothing; a floor grades layer 1 only |
| MJ | delete the `path == ""` clause | permissive | **DEAD** *(was ALIVE before the fixture fix)* | the `empty path` subtest |
| MN | walk stops recording no-extractor languages | permissive | **DEAD** | `noExtractorLanguages6847` |

Every edit applied by an exact-count replace aborting unless it matches exactly one site (M9's first anchor matched 2 and aborted — re-anchored, re-run). Restored by `cp` from a private-scratchpad backup with shasums verified equal after every mutant; never `git checkout --`.

## Review defects fixed, each proved rather than argued

- **B3 — the `an empty path is not an anchor` fixture was vacuous.** Its record left `QualifiedName` `""`, so `"" == path` matched the **carrier** branch and returned nil first: the subtest passed for a reason unrelated to its label, and MJ was **ALIVE**. The record now carries `QualifiedName: "std"`; MJ is **DEAD**, killed by that exact subtest (`want [] got [IMPORTS]`). This is precisely the clause `file_carrier.go` calls out as *"tested separately and not folded into clause 2"* — it was folded.
- **B2 — a doc claim contradicted by the code.** `carrierScanResult` claimed the tests "share a single traversal and cannot disagree" while two call sites ran two independent walks with no caching. Now memoised in a `sync.Once`; the coverage test drops from ~2.5s to **0.00s**, which is the observable proof the claim is now true.
- **B1 — the population sentence.** Reworded to a corpus-bound lower bound, with the three out-of-corpus offenders named and the coverage gap documented as above.

## Corrections to the record

1. **My own item-6 was half wrong, and the reviewer is right to push back.** I reported the source scan's rationale as "stale". Only its parenthetical is: "27 languages" is 41 now, and astro *is* present (it is one of the twelve). `verilog` and `solidity` still have zero corpus files. **The reasoning is not stale — it is vindicated**, by my own `dockerfile`/`reasonml`/`rescript` gap. Calling it stale inverted the lesson. (Nothing false landed: the committed file made no claim about it; the error was in my report.)
2. **My earlier numbers were wrong and are fixed everywhere.** 907 candidate files, not 909 — the 909 came from a scaffold whose filter matched `testdata` as a substring rather than as a path segment. And the un-rewritten-path trap is **total**: **0** files classify, not 2, because every candidate contains a `/testdata/` segment by construction and `classifier.depDirs` skips it outright.
3. **One correction back to the review.** The `dockerfile` evidence pair showed `.../docker/Dockerfile lang="dockerfile" skip=false` alongside the dropped `Dockerfile.multi_stage`, implying the corpus holds a classifiable bare `Dockerfile`. It does not — `find` over the corpus returns exactly three docker fixtures and **all three are dropped**. The `skip=false` line reads as a synthetic probe of the classifier, not a corpus file. The substance is unaffected and if anything stronger: the corpus has no reachable dockerfile at all.
4. **Report-only, unchanged:** the Windows leg is worth watching rather than merging on local green, though `filepath.Join` normalises the `ToSlash`'d relative paths on its own. `M-K`/`M-L` (deleting a set-pin assertion from its call site) are trivially ALIVE and grade nothing — the call site is graded by M4/M5/M6 and M7/M8/MN — so assertion-deletion is not counted as call-site coverage anywhere above.

## Verification

- `go test -count=1 ./internal/extractors/` — **ok, 56.0s**
- `gofmt -l internal/extractors internal/extractor` — clean
- `go vet ./internal/extractors/` — clean
- `go run ./tools/coverage validate` — `ok: 822 record(s) ... 1163 test(s) checked`
- `-run '6847' -v` selected **12 `=== RUN` lines**; every absence claim rests on the whole-package run, not on a filter.
- `./cmd/grafel/` ran green (144.2s, digest unmoved) on the previous revision and is skipped here per review: this revision is test-only and touches no non-test source.

Test-only change: one new `_test.go` file.

### DIFF
--- a/internal/extractors/file_anchored_carrier_runtime_6847_test.go
+++ b/internal/extractors/file_anchored_carrier_runtime_6847_test.go
@@ -1,0 +1,591 @@
+// Package extractors — file_anchored_carrier_runtime_6847_test.go
+//
+// Issue #6847. A RUNTIME guard for the invariant #6815 fixed by hand in three
+// languages: EVERY relationship whose FromID is the source file's own path must
+// have some emitted record carrying that exact path as its Name or
+// QualifiedName. internal/resolve/refs.go has no path→entity index, so a
+// path-valued FromID resolves if and only if such a node exists; when none
+// does, the raw path reaches the graph as the edge's FROM end and the edge
+// points at nothing.
+//
+// WHY A SECOND GUARD, NEXT TO file_anchored_rels_guard_test.go. That file is a
+// SOURCE SCAN, and it deliberately skips IMPORTS (the #120 convention). Every
+// instance of this defect found so far — solidity, verilog, astro, svelte,
+// erlang, nim, groovy — was an IMPORTS edge, so the one guard aimed at the
+// class cannot see the kind that carries it. This file sidesteps the question
+// entirely: it never asks what KIND an edge is. It asks only whether the FROM
+// end resolves, which is the actual invariant, and it answers by RUNNING the
+// registered extractors rather than by reading their source.
+//
+// WHAT #120 ACTUALLY SAYS, since the skip is justified by it. #120 is "Java
+// cross-file receiver binding"; the convention lives in its closing comment:
+// "Source-file-path FromIDs (every IMPORTS edge across every language) tagged
+// Dynamic instead of bug-extractor". That is a decision about a METRIC — such
+// an edge stops counting as an extractor bug — not a statement that the edge
+// resolves. It does not. #6815 treated exactly this shape as a defect and fixed
+// it in three languages, so the repo's own position is that a dangling
+// path-anchored IMPORTS is a bug whose metric is merely suppressed. Nothing in
+// #120 licenses skipping the invariant checked here.
+//
+// WHAT THIS COSTS AND WHAT IT CANNOT SEE, stated rather than implied:
+//
+//   - COVERAGE IS CORPUS-BOUND, AND THE GAP IS REAL. The corpus is every file
+//     under a testdata/ directory in this repo, classified by the production
+//     classifier and parsed by the production parser factory. A registered
+//     language with no corpus file is NOT measured, and a new extractor is
+//     therefore NOT automatically covered. That is the opposite of the
+//     neighbouring source scan's property, and the source scan's stated reason
+//     for preferring a derive-shaped check — that a corpus "grows only when
+//     someone remembers to grow it" — is VINDICATED here, not superseded:
+//     dockerfile, reasonml and rescript are offenders this corpus cannot see.
+//     The two checks are complements. Neither replaces the other.
+//   - exercisedLanguages6847 IS HAND-MAINTAINED AND IS NOT DERIVED FROM
+//     extractor.List(). Stating that plainly, because a pinned slice that
+//     relates to nothing outside itself makes every omission read as
+//     deliberate (#6849's pattern). A List()-relative pin is not available:
+//     List() returns 445 names, of which ~380 are framework extractors
+//     (custom_*, python_*, lua_*) that no classifier language ever dispatches,
+//     so "registered minus exercised" is not a meaningful coverage number.
+//     What IS pinned against something external is noExtractorLanguages6847
+//     below. The unexercised BASE languages, measured, are the 24:
+//     assembly avro c commonlisp dockerfile elm erlang haskell hcl idris
+//     jsonschema nim ocaml pony racket reasonml rescript scheme sml solidity
+//     swift_package systemverilog verilog vhdl — three of which (dockerfile,
+//     reasonml, rescript) are confirmed offenders.
+//   - ONE CORPUS DROP IS NOT A MISSING FIXTURE. dockerfile has three corpus
+//     files and reaches the extractor with NONE of them: sample.Dockerfile
+//     (x2, under testdata/fixtures/dockerfile and .../sources/dockerfile) and
+//     real-world/docker/Dockerfile.multi_stage all classify as
+//     lang="" skip=true reason="unsupported_extension" — the classifier's
+//     basename table maps the exact name "Dockerfile", not the
+//     `*.Dockerfile` / `Dockerfile.<variant>` spellings the fixtures use. The
+//     extractor is registered and the fixtures exist; the classifier is what
+//     the corpus cannot cross. Filed separately as a classifier question.
+//   - IT IS PER-FILE. A carrier emitted by some OTHER file's extraction would
+//     satisfy the resolver's repo-wide byName index but not this check. No
+//     in-tree extractor mints a node named after a file it is not extracting,
+//     so the two coincide today; if one ever does, this guard reports it and
+//     the entry belongs in the ledger with that reason.
+//   - IT RUNS ONE PASS. Pass 2.5 engine rules and the custom-extractor
+//     dispatch are not driven here, so a carrier minted by a later pass is not
+//     seen. Same disposition as above: report, then justify in the ledger.
+//
+// MEASURED STATE WHEN THIS GUARD LANDED (2026-09-05, before any fix). The
+// invariant does NOT hold today. Twelve registered languages violate it AS
+// MEASURED BY THIS CORPUS — all on IMPORTS, and none of them emitting a file
+// carrier at all:
+//
+//	astro bicep clojure cobol crystal dart fsharp html lua shell terraform zig
+//
+// TWELVE IS A LOWER BOUND, NOT A POPULATION. The count is corpus-bound, and the
+// bound is not theoretical: driving the registered languages this corpus never
+// reaches from hand-written minimal sources found THREE MORE, confirmed against
+// the source the same way —
+//
+//	dockerfile (dockerfile/dockerfile.go:347 `FromID: file.Path`)
+//	reasonml   (reasonml/extractor.go:325)
+//	rescript   (rescript/extractor.go:296)
+//
+// none of which calls FileEntity or PrependFileCarrier either. So the known
+// total is at least FIFTEEN offenders / EIGHTEEN instances counting the three
+// #6815 fixed, and the population is unbounded above until every registered
+// language is driven. All of it is tracked in #6852. None is fixed here: each
+// needs the same per-language judgement #6815 made (conditional carrier vs. one
+// orphan node per source file) and each moves the golden fixtures and
+// cmd/grafel's whole-graph digest.
+//
+// THE LEDGER IS AN EXACT SET, NOT A FLOOR, IN BOTH DIRECTIONS. A thirteenth
+// language fails this test; so does fixing one of the twelve without removing
+// its entry. A guard that only checked "no NEW offender" would be blind to the
+// permissive direction — a walk that silently stopped detecting anything would
+// stay green under a floor and goes red here.
+package extractors_test
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"path/filepath"
+	"sort"
+	"strings"
+	"sync"
+	"testing"
+
+	"github.com/cajasmota/grafel/internal/classifier"
+	"github.com/cajasmota/grafel/internal/extractors"
+	"github.com/cajasmota/grafel/internal/treesitter"
+	"github.com/cajasmota/grafel/internal/types"
+)
+
+// knownMissingCarrier6847 is the exact set of registered languages measured to
+// emit a path-anchored FromID with no record carrying that path. Every entry
+// was produced by the walk below, not by reading source, and each was confirmed
+// against the extractor package: none of the twelve calls extractor.FileEntity
+// or extractor.PrependFileCarrier anywhere in its non-test sources.
+//
+// To REMOVE an entry: fix the extractor (see internal/extractor/file_carrier.go
+// for the conditional-carrier shape #6815 and #6518 settled on) and delete the
+// line. To ADD one: do not. A new offender is the defect this file exists to
+// catch; adding a line to keep it green is the failure mode #6834 names.
+var knownMissingCarrier6847 = map[string]string{
+	"astro":     "extractor.go:328 `FromID: filePath` on IMPORTS; no carrier emitted",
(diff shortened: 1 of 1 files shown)


cajasmota:
## Coordinator verification — the measurement came back RED, and that is the whole point

I filed #6847 as *"nothing would catch the fourth extractor."* Putting the measurement **before** the guard answered it: **the fourth through fifteenth already exist.** A guard added green would have said "ratchet in place, nothing to see."

**My killing mutant, which nobody else ran:** claim `"go"` — a language that plainly has a registered extractor — belongs in `noExtractorLanguages6847`, the external pin added in round 2. **DEAD** on `TestFileAnchoredCarrier_CorpusCoverage_6847`. So the newest mechanism is graded in the **permissive** direction too, not just the recall one.

## The population is a lower bound, and the review proved it

The twelve were re-derived **independently** — a second walk, a different path-rewrite segment (`corpuslib` rather than `src`), and a matcher written from the invariant statement rather than reusing the author's — producing the identical list and identical exercised (41) / anchoring (34) sets.

Then the reviewer drove the languages **the corpus never reaches** with hand-written sources and found three more: `dockerfile`, `reasonml`, `rescript`, cross-checked against source (`dockerfile.go:347`, `reasonml/extractor.go:325`, `rescript/extractor.go:296`, none calling `FileEntity`/`PrependFileCarrier`). The other 21 unexercised languages are clean.

So it is **≥15 offenders / 18 instances**, and the file now says *"twelve as measured by this corpus"* rather than asserting a total. **#6852 has been corrected and retitled** so nobody inherits the wrong number.

## #120 does not say what the skip implied

#120 is *Java cross-file receiver binding*. Its `IMPORTS` sentence — *"tagged Dynamic instead of bug-extractor"* — is a **metric** decision, not a resolution claim. So the source-scan guard's skip was left alone and this guard never asks an edge's kind, only whether its FROM end resolves. That is the narrow invariant, and it sidesteps the convention rather than fighting it.

## The corpus-derivation trap, and why the numbers moved twice

`classifier.depDirs` skips `/testdata/`, so classifying corpus paths as-is exercises **nothing**. The author's first run reported 4 violations instead of 22.

Then two of their own reported numbers failed to reproduce and were corrected: **907 files, not 909** (their scaffold matched `testdata` as a substring rather than a path segment) and **0 files classify un-rewritten, not 2** — same cause. The rewrite itself was proved non-distorting: `rewriteDisagree=0`, `srcSkip == altSkip`, byte-identical classified-language sets across two different replacement segments.

The lesson worth keeping: **a low number from a derived corpus is a claim about the corpus first.**

## Two defects the review proved, both fixed

- **`M-J` ALIVE** — the subtest labelled *"an empty path is not an anchor"* passed for an unrelated reason: its fixture left `QualifiedName` empty, so `"" == path` hit the **carrier** branch and returned before the clause under test mattered. Setting `QualifiedName: "std"` kills it. Same clause `file_carrier.go` says is *"tested separately and not folded"* — it was folded.
- **A doc claiming three tests "share a single traversal and cannot disagree"** when there were **two** call sites and no caching. Now memoised, with observable proof: the coverage test drops ~2.5s → **0.00s**.

## A correction that runs back up the chain

The author reported the sibling source-scan guard's stated reason as **stale**. The review found the committed file makes no such claim (so nothing false landed), and that the claim as relayed is **half wrong**: the parenthetical is stale (27→41 languages, and `astro` is present), but **the rationale is vindicated** — a corpus check grows only when someone remembers to grow it, while a source scan covers a new extractor the day it lands. The `dockerfile`/`reasonml`/`rescript` gap is an in-tree demonstration. Calling it stale inverts the lesson; the file now says the two checks are complements.

Separately, the reviewer's evidence pair implied the corpus holds a classifiable bare `Dockerfile`. It does not — **all three** docker fixtures are dropped as `unsupported_extension`, and the classifier matches only the exact basename, not `*.Dockerfile` (which is what both of this repo's own fixtures use). I had propagated that into **#6854**, which is now corrected and retitled.

## Honest limits, kept

- **`M10` (corpus floor 700→0) is ALIVE and equivalent**, recorded in the constant's own doc. A floor grades layer 1 only.
- **`M-K`/`M-L`** (deleting a set-pin assertion from its call site) are trivially ALIVE and grade nothing. The call site is properly graded by mutating **real extractors** (`M-C` renames golang's `FileEntity` off the path; `M-D` gives zig a carrier — each fires **alone**) and by mutating the walk (`M-F` truncation, `M-H` dropping the root `testdata/`). Assertion-deletion is not counted as call-site coverage.
- **`exercisedLanguages6847` is hand-maintained and documented as such**, with the 24 unexercised base languages listed and `noExtractorLanguages6847` as the one thing pinned against something external.

**Watch the Windows leg rather than merging on local green** — `filepath.Join` mixes separators and the walk assumes a full checkout. New always-on test in a package CI runs on Windows.

Issue #1613, opened 2026-05-22
extraction: class entities are line-less INFERRED Component shadows duplicating the real node — fold + give real line spans + qualified_name


cajasmota:
Validated on upvate via inspect/search/find. Every class is represented by an inferred `SCOPE.Component` shadow with provenance=INFERRED_FROM_CLASS_HIERARCHY, start_line=0, end_line=0, qualified_name="", language="" — DUPLICATING the real typed node. E.g. LoginViewSet appears as BOTH Component(id 55d3..,line 0) AND View(id 071a..,line 22); same for RegistrationViewSet/RefreshViewSet/LoginSerializer/TokenAuthenticationMiddleware. Consequences: (a) get_source unusable on classes (no bound → empty/whole-file dump), (b) find/inspect return :0 pointers (useless), (c) duplicate-node noise (~2-3x/class per the calibration report). FIX: fold the inferred Component shadow INTO its real typed node (View/Model/class/Operation at the same source_file+symbol), OR backfill the shadow with the real class declaration start_line/end_line + qualified_name from the AST. After: one node per class with a real line span + qualified_name; no line-0 shadow. Verify on upvate: search "LoginViewSet"/"TokenAuthenticationMiddleware" returns a single class node with a real start_line; quantify how many start_line=0 / INFERRED_FROM_CLASS_HIERARCHY nodes exist before/after.

PR #6856, opened 2026-09-05
fix(#6841): grade the classfold twin-declaration claim with internal/entkinds — a quoted-literal scan was blind to the rule YAML that mints these kinds

### LINKED ISSUE #6841: classfold's "both spellings must appear" is false for thirteen rows — eight of them vacuous (the prefixed spelling is not a kind at all)
`classfold.go`'s doc claims every bare kind appears alongside its `SCOPE.`-prefixed twin — eleven rows violate it, and nothing checks

## The over-claim

`internal/engine/classfold.go:35-38` states that both the bare kind names **and** their `SCOPE.`-prefixed forms must appear in the map, so lookups match whichever spelling a caller holds.

**Eleven bare rows have no prefixed twin:**

`Controller`, `Middleware`, `Repository`, `Worker`, `Job`, `Topic`, `TestClass`, `Plugin`, `Implementation`, `Interface`, `Task`

Nothing observes the invariant, so the doc has been asserting a property the data does not have.

## Why it kept costing time

This map is a natural place to look for "are these two spellings the same concept?", which is exactly the question #6776's migration arms ask for every kind. **I cited it twice as evidence and was wrong both times:**

- **arm B6, `Component`** — I briefed "one classfold citation covers all four". `classfold.go:29-33` in fact says `SCOPE.Component` is *intentionally absent* (a generic AST node, folded away, never a fold candidate), and the bare form is absent too. The agent found the real evidence in `internal/resolve/refs.go` instead.
- **arm B8, `Plugin`** — I cited it again. `:52` has `"Plugin": 80` and there is **no** `"SCOPE.Plugin"` row. The agent refused the citation and used the rule-pack sites plus `SCOPE.Plugin`'s own doc.

Both times the doc's blanket claim is what made the citation look safe without opening the file. That is the cost: a doc that promises completeness invites exactly the shortcut it cannot support.

Note `Constraint`'s absence is **deliberate and correctly documented** — `:26-27` names "a Django Meta `Constraint`" as the thing that must never win a fold. So the map does distinguish intentional absences; it just does not distinguish them *in the doc's blanket claim*, and there is no test either way.

## The fix

Either make the invariant true, or make the doc describe what is actually there — a set of rows where pairing is per-entry and sometimes deliberately one-sided — and **add the check** for whichever version survives. A doc that says "must" with nothing enforcing it will drift again; this is its third recorded instance.

If pairing genuinely is per-entry, the useful guard is not "every bare row has a twin" but "every row's pairing state is *declared*", so an accidental one-sided addition is distinguishable from a deliberate one. The two intentional absences already on record (`Component`, `Constraint`) show the vocabulary for that exists — it is just in prose rather than in data.

## Not established

- **Whether any of the eleven is a deliberate absence** like `Component` and `Constraint`, or all eleven are drift. Nobody has checked per row; that check is most of the work here.
- **Whether the missing twins cause real lookup misses at runtime.** The doc says they would, but no failing case has been exhibited — it is possible callers only ever hold one spelling for these particular kinds, which would make this documentation-only.
- Whether `FrameworkClassKindCanonRank` (the sibling map) has the same shape. **Not examined.**

Found during the independent review of #6776 arm B8; **pre-existing** and untouched by that change. Related by class: prose asserting what no test observes — the dominant defect of this batch.

cajasmota:
fix(#6841): classfold's "both spellings must appear" was false for thirteen rows — declare pairing per row, in data, and grade it with internal/entkinds

Closes #6841

## The measurement came first, and it changed the fix

The issue asked, per row, whether each of the eleven unpaired bare kinds is a
deliberate absence or drift. **Eight are neither — they are vacuous**, one is a
live miss, and two are produced-but-never-class. Making the old doc true would
have added eight rows for kinds that cannot reach the fold.

| row | verdict | evidence |
|---|---|---|
| `Controller` | **vacuous** | `"SCOPE.Controller"` is not in `types.AllEntityKinds`; `entkinds.Scan` reports 0 declaration sites; its 3 Go mentions (`enrichment/candidates.go:101,457`, `pricing.go:47`) all read it |
| `Repository` | **vacuous** | 0 declarations, 0 mentions |
| `Worker` | **vacuous** | 0 declarations, 0 mentions |
| `Job` | **vacuous** | 0 declarations, 0 mentions |
| `Topic` | **vacuous** | 0/0; `SCOPE.MessageTopic` is a different kind, not this one prefixed |
| `TestClass` | **vacuous** | 0 declarations, 0 mentions |
| `Implementation` | **vacuous** | 0 declarations, 0 mentions |
| `Task` | **vacuous** | 0 declarations; 2 consumer case arms in `links/effect_propagation.go:587,605` |
| `Middleware` | **produced, never class-shaped** | `custom/scala/frameworks.go` emits `"SCOPE.Middleware"` at 9 sites — every Name is built from a literal prefix (`"middleware:akka:"+directive`), so no AST class node can carry it. Note the subtype does NOT settle this: `"middleware"` **is** in `ClassLikeComponentSubtypes` |
| `Plugin` | **produced, never class-shaped** | `SCOPE.Plugin` is a real enum kind reaching the graph via `plugin_system_edges.go:59`'s `string(types.EntityKindPlugin)`, for a plugin *registration* in a build/config file. **This is the row #6776 arm B8 cited as a classfold pair; it is not one** |
| `Interface` | **DRIFT — a live, exhibited miss** | `custom/scala/type_system.go:391,411` emits `"SCOPE.Interface"` for a `trait` and an `abstract class`, on the same `(source_file, name)` `extractors/scala/scala.go` emits `SCOPE.Component` subtype `trait` for — which IS a fold source |

Nothing was left undetermined. **Two more unpaired rows the issue did not
count**, from the other side: `SCOPE.UIComponent` and `SCOPE.GrpcService` have
no bare twin. Thirteen, not eleven.

## Does a missing twin cause a real lookup miss? YES — once

```
SCOPE.Interface: folded=0 rows=[SCOPE.Component|Repo|trait|a.scala SCOPE.Interface|Repo|trait|a.scala]
bare Interface:  folded=1 rows=[Interface|Repo|trait|a.scala]
```

A Scala trait reaches the fold twice and does not fold: two nodes for one class,
against #1613. Pinned by `…_ProducedClassLikeTwinMissesTheFold`, with the bare
spelling as the control and an `IsClassFoldSource` premise check so the exhibit
cannot pass on a record shape that was never a fold source.

**Deliberately NOT fixed here.** Adding `"SCOPE.Interface": 80` makes the
survivor carry a kind `IsValidEntityKind` rejects while deleting the valid
`SCOPE.Component`. The fix belongs at the Scala producer or in the enum; the
test fails the day it lands.

## `FrameworkClassKindCanonRank`

Same unpaired shape, no independent defect: every row it leaves unpaired is a
row `FrameworkClassKindTwins` declares unpaired. That containment is now
**asserted** (`…_CanonRankUnpairedRowsAreDeclaredUnpaired`), one-way on purpose
— a kind it does not rank at all makes no pairing statement.

## The fix

Prose "must" replaced by data: `ClassKindTwinState`
(`TwinInMap`/`TwinUnproduced`/`TwinProducedNonClass`/`TwinProducedClassLike`)
and `FrameworkClassKindTwins`, a declaration for **every** row carrying
`Producers`, `KnownSites`, `NotClassShaped` and `FoldSourceSubtype`.

**The scan is `internal/entkinds`, not a local matcher.** The first version of
this guard rolled its own walk for double-quoted spellings and was blind to the
dominant producer of exactly these kinds: rule YAML writes them **unquoted**
(`entity_type: Controller`, ×37). Measured: `entity_type: Config` →
`entity_type: SCOPE.Repository` in `gin.yaml` was **ALIVE** against that
version. `entkinds.ScanRuleYAML` reads those scalars and
`entkinds.ScanGoReferences` resolves `string(types.EntityKindX)` conversions and
kinds passed as function arguments — the two holes that package exists for
(#6744, #6818). Both blind spots were load-bearing here: all three produced
twins live in one or the other.

Ten tests grade it. The two that carry the weight:
`…_UnproducedTwinIsDeclaredNowhere` (0 declaration sites, both halves) and
`…_EveryGoMentionOfAnUnpairedTwinIsDeclared` (every Go mention accounted for as
Producer or KnownSite, both directions).

## Review items

- **B1 closed.** MYAML now DEAD. Doc claims corrected: the struct doc no longer
  says "string literal", and the unproduced block now states the rule-YAML
  counts explicitly.
- **B2 closed.** MN and MO now DEAD.
- **R1 answered.** The walk is no longer mine to narrow — the root is derived
  from `runtime.Caller` and `entkinds` walks it with `repowalk` skipping. What
  anchors coverage is `internal/entkinds.TestRepoSweepIsNotVacuous`, which
  requires the scan to have parsed *exactly* the files an independent walk
  finds; the comment now says that instead of implying the local floor does it.
- **R2 fixed.** `Plugin`'s Producer is `internal/engine/plugin_system_edges.go`,
  and `internal/types/kinds.go` moved to `KnownSites` (it declares the enum
  member, it emits nothing). MR2 — repointing the real producer at another kind
  — is now DEAD.
- **R3 fixed.** `Controller`'s `Why` now says three mentions, and names
  `candidates.go:457` as a map-literal key.
- **R4 closed, both directions.** Each state now requires its own evidence:
  `ClassLike` needs a `FoldSourceSubtype` that is in `ClassLikeComponentSubtypes`
  *and* appears in the producer file; `NonClass` needs `NotClassShaped`;
  `Unproduced` may have neither, nor any Producers. MB (promote `Plugin`) and MC
  (demote `Interface`) are both DEAD, and MC no longer depends on `Interface`
  being the only ClassLike row.
- **R5** — the four rows dead in both spellings — left for the reviewer's issue.

## Mutants — 20 scored individually, 19 DEAD, 1 ALIVE and recorded

All verdicts re-scored after the scanner switch. `go vet` before every scoring
run (MGO was rejected three times as non-compiling before it vetted); exact-count
replace aborting on ≠1; `cp` restore with verified shasums; `-count=1`; the
`-run` filter proven to select (10 tests ran); ALIVE claims scored
whole-package.

|---|---|---|---|
| MYAML | `entity_type: SCOPE.Repository` in `gin.yaml` | **DEAD** (was ALIVE) | `UnproducedTwinIsDeclaredNowhere` |
| MGO | function-local `"SCOPE.Repository"` literal in `grpc_edges.go` | DEAD | `EveryGoMentionOfAnUnpairedTwinIsDeclared` |
| MR2 | repoint `plugin_system_edges.go` at another kind | **DEAD** (R2) | `EveryGoMention…` (stale Producer) |
| ME | revert `Plugin`'s Producer to `kinds.go` | DEAD | `EveryGoMention…` |
| MN | drop `"SCOPE.Model"` from CanonRank | **DEAD** (was ALIVE) | `CanonRankUnpairedRowsAreDeclaredUnpaired` |
| MO | drop `"SCOPE.Schema"` from CanonRank | **DEAD** (was ALIVE) | same |
| MB | promote `Plugin` to `TwinProducedClassLike` | **DEAD** (was ALIVE) | `EachStateCarriesItsRequiredEvidence`, `ProducedClassLike…` |
| MC | demote `Interface` to `TwinProducedNonClass` | DEAD | same two |
| MD | `FoldSourceSubtype` → `"controller"` (class-like but wrong) | DEAD | `EachState…` (subtype absent from producer) |
| M3 | downgrade `Middleware` to `TwinUnproduced` | DEAD | `EachState…` (Producers + NotClassShaped forbidden) |
| M5 | empty `Interface`'s `Producers` | DEAD | `EveryGoMention…`, `EachState…` |
| MA | empty `Controller`'s `KnownSites` | DEAD | `EveryGoMention…` |
| M1 | delete the `Interface` declaration | DEAD | 3 tests |
| M2 | declare `Interface` `TwinInMap` | DEAD | 3 tests |
| M4 | upgrade `Task` to `TwinProducedNonClass` | DEAD | `EachState…` |
| M9 | priority row `"Handler": 70` with no declaration | DEAD | `DeclaresEveryPriorityRow…` |
| M10 | one-sided `"SCOPE.Repository": 100` (the original defect shape) | DEAD | 3 tests |
| M14 | add `"SCOPE.Interface": 80` — close the declared gap | DEAD | 4 tests, incl. the exhibit |
| M15 | rank a kind absent from the priority map | DEAD | `CanonRankRanksOnlyPriorityRows`, `EveryGoMention…` |
| **MF** | neuter the Go file-count floor (`< 1000` → `< 0`) | **ALIVE** | — |

**MF is genuinely equivalent under this suite, and left that way.** The floor is
a local canary; nothing else here observes it, because the real coverage anchor
is `internal/entkinds.TestRepoSweepIsNotVacuous`, which compares the scan's
parsed-file counts against an independent walk of the repository. Duplicating a
derived floor in this package would be a worse assertion in a worse place. The
comment now says this instead of implying the floor is load-bearing. What this
file *does* add — and what a file count cannot give — are the two find-controls
(`SCOPE.Component` for the Go half, `Controller` for the YAML half), which go to
zero if a half stops resolving while still parsing every file.

## Verification

- `go test -count=1 ./internal/engine/` — ok (41s)
- `go test -count=1 ./cmd/grafel/` — ok (153s; graph digest unchanged)
- `go test -count=1 ./internal/entkinds/` — ok (new dependency)
- `gofmt -l internal/engine/` — clean
- `go run ./tools/coverage validate` — exit 0

## Follow-up worth an issue (not done here)

`SCOPE.Interface` (Scala traits and abstract classes, 2 sites) and
`SCOPE.Middleware` (9 sites) are emitted today, are **not** in
`types.AllEntityKinds`, and are invisible to `entkinds.Scan` because both pass
the kind as a function argument — the call-site hole #6776 arm B4 documented and
left open. For `Interface` that also costs a fold: every Scala trait in an
indexed repo is two nodes, one carrying a kind `IsValidEntityKind` rejects.
`internal/types/kinds.go:505` already records the synonym state; the fold
consequence is new.

### DIFF
--- a/internal/engine/classfold.go
+++ b/internal/engine/classfold.go
@@ -32,10 +32,18 @@ import "github.com/cajasmota/grafel/internal/types"
 // resolves to ONE node), and it is itself the survivor only when no
 // framework-typed node exists (handled separately, never as a candidate here).
 //
-// Both bare kind names (emitted by Java/Django custom extractors) AND their
-// "SCOPE."-prefixed forms (emitted by Kotlin, TypeScript, proto, and pattern
-// extractors) must appear so that FrameworkClassKindPriority[r.Kind] matches
-// regardless of which extractor emitted the survivor. Issue #1700.
+// Rows come in BOTH spellings — bare kind names (emitted by Java/Django custom
+// extractors) and "SCOPE."-prefixed forms (emitted by Kotlin, TypeScript, proto
+// and pattern extractors) — because FrameworkClassKindPriority[r.Kind] is a
+// plain map lookup on whatever spelling the emitting extractor chose (#1700).
+//
+// PAIRING IS PER ROW, NOT UNIVERSAL. This doc used to claim that every kind
+// appears in both spellings. It never did: eleven bare rows and two prefixed
+// rows have no twin here, and the claim was cited twice as proof that a
+// particular pair exists — wrongly both times (#6841). The reason a row has no
+// twin is now recorded per row in FrameworkClassKindTwins below, and graded by
+// classfold_twin_declaration_6841_test.go, so a one-sided addition can no
+// longer be mistaken for a deliberate omission.
 var FrameworkClassKindPriority = map[string]int{
 	// Bare names (Java/Django/Spring-boot custom extractors)
 	"Model":          100,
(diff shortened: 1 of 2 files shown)


cajasmota:
## Coordinator verification

**My own killing mutant (nobody else ran this one) — DEAD.**

The interesting property of this guard is not that it scans, but that it scans through **two** entry points split by claim. So I attacked the split rather than the walk: leave the declaration half intact, and neuter only the Go-mentions half while it still visibly "scans".

```
cfRefRes, cfRefErr = entkinds.ScanGoReferences(cfRepoRoot(t), cfUnpairedOpposites())
  →                  entkinds.ScanGoReferences(cfRepoRoot(t), []string{"zzNoSuchSpelling6841"})
```

`go vet` clean first (a non-compiling mutant proves nothing), then `-count=1` on the whole package:
`--- FAIL: TestClassKindTwins6841_EveryGoMentionOfAnUnpairedTwinIsDeclared`, exit 1. **DEAD** — the split is graded, not decorative.

File restored from a byte-level backup by `cp`, shasums matched both ways (`9785aee2…`), worktree clean.

**Why the split is the load-bearing part**, and it is worth recording because it inverts the obvious design: the author measured that `entkinds.Scan` alone reports **0 declaration sites for all three produced twins** (`SCOPE.Middleware`, `SCOPE.Interface`, `SCOPE.Plugin`) — `ScanGo` reads `Kind:` fields, and all three reach the graph another way. So the single-entry-point version of this guard would have reported "unproduced" for **precisely the rows this issue turns on**, green, in the shape of a measurement. It takes both entry points, split by claim: `Scan` for declarations, `ScanGoReferences` for mentions.

**One mutant recorded ALIVE-equivalent, not DEAD:** neutering the file-count floor. `internal/entkinds.TestRepoSweepIsNotVacuous` already derives a floor over the same sweep, so a second floor here is redundant rather than missing. Recording it rather than adding a wrapper — the outermost seam fails loudly by other means, which is where this regress stops.

**A correction to the review, which inherited it from the original wording rather than introducing it:** `Middleware`'s *subtype* does not settle whether it is class-like — `"middleware"` **is** in `ClassLikeComponentSubtypes`, so the subtype is evidence in the wrong direction if anything. What actually settles it is the **Name shape**: the literal prefix `"middleware:akka:" + directive`, which is not a declaration site. The conclusion stands; the stated reason did not, and the review passed it through unchallenged.

Issue #6839, opened 2026-09-05
Ten of eleven hardened reads in internal/links mapped a read failure to "nothing found" — one turned it into a positive "this code is dead" claim; all three arms shipped


cajasmota:
Ten of the eleven hardened source reads in `internal/links` map `ErrWouldBlock` to "nothing found" — the thing safeio's own package doc forbids

> **Corrected 2026-09-05 by the arm-1 measurement** ([comment](https://github.com/cajasmota/grafel/issues/6839#issuecomment-5552693238)). Three premises below were wrong and are struck through in place; the "Not established" section is now established and replaced. The headline inconsistency is real and unchanged.

## The inconsistency

#6823 routed eleven source-path reads in `internal/links` through `safeio`. They now disagree about what a blocking read means:

- **`string_pass`** propagates a non-`ErrNotRegular` error, so it reaches `runStringPass` → `links.go:388` → out of `RunAllPasses`. Loud, and deliberately so.
- **The other ten** `continue` or cache `""` on any error. A read that would have blocked is therefore indistinguishable from a file that legitimately contained nothing.

`safeio`'s package doc explicitly says a would-block condition must not be mapped to "nothing found". Ten sites do exactly that.

Measured split: **8 `continue` + 2 cache-`""` + 1 propagate**, across **ten passes** (`dataflow_pass` owns two of the eleven reads). Site list in the arm-1 comment.

## Why this is being filed now rather than fixed in #6823

Those `continue`/cache-`""` arms are **pre-existing** — they predate the hardening and were written when the only errors reachable there were ordinary I/O failures. What #6823 changed is that **`ErrWouldBlock` is now reachable on those paths at all**; ~~before the hardening, a FIFO simply hung instead of returning an error. So the arms did not become wrong, they became *reachable* while wrong.~~

**Correction:** a planted FIFO does **not** reach these sites as `ErrWouldBlock`. `safeio.Open` calls `Stat` first, which refuses any non-regular file with `ErrNotRegular` — and `string_pass` maps `ErrNotRegular` to a skip too. On the #6416 input all eleven sites behave identically. On Unix `ErrWouldBlock` is reachable only from safeio's global 64-slot saturation / terminal state, which a FIFO cannot cause (`O_RDONLY|O_NONBLOCK` returns immediately).

[...]

Issue #6861, opened 2026-09-05
file_carrier.go states its grading status as a MEASURED fact about its caller set — #6852 re-falsifies it twelve times, once per language


cajasmota:
`internal/extractor/file_carrier.go` states its grading status as a **measured fact about its caller set** — and #6852 will invalidate that fact twelve times over, once per language.

## The claim

`file_carrier.go:63-69`:

> **MEASURED** grading status of that parameter, stated rather than implied: a WRONG token is caught (mutating erlang's `"erlang"` to `"beam"` fails `TestErlang_CarrierIsLanguageTagged_6815`), but an EMPTY one is not — **all three current callers** run `extractor.TagEntitiesLanguage` afterwards … so passing `""` is equivalent under the suite. **It is not equivalent for a caller that does not tag**, which is why the parameter exists and is passed.

And at `:72-76`: "for **erlang, nim and groovy** the per-import stub records still carry the IMPORTS edges themselves".

Verified on `3bcb17807`: exactly three non-test callers — `internal/extractors/{nim/nim.go, erlang/extractor.go, groovy/groovy.go}`. The doc is **accurate today**.

## Why it is worth an issue anyway

The doc anticipated its own falsification and said what would falsify it — *"not equivalent for a caller that does not tag"* — and #6852's first language is exactly that caller. bicep does **not** call `TagEntitiesLanguage`, so an empty `lang` **is** graded there (that mutant is DEAD on bicep's branch), and the sentence "passing `""` is equivalent under the suite" stops being true the moment it lands.

#6852 has **twelve** languages to work through, one PR each. Every one of them re-opens this same
sentence, and the count "all three" drifts by one each time. A doc that is re-falsified twelve times
by a known, already-filed work queue is not going to survive by diligence.

**This is the repo's dominant defect class in its most durable form.** A `MEASURED`-labelled claim
reads as verified — it *was*, once — and nothing observes it now. The three miscitations #6841 and
#6858 corrected all had this shape: true when written, load-bearing, and unobserved thereafter.

## What the fix should do — grade the claim, do not restate it

Restating "all four current callers" just resets the clock. The claim to grade is the **caller
set** itself:

- enumerate the non-test callers of `PrependFileCarrier` / `FileCarrierFor` from source;
- assert, per caller, whether it runs `TagEntitiesLanguage`;
- fail when a caller appears that the doc's stated equivalence does not cover.

That is the same shape as the anchor-list self-checks now used in `internal/types`,
[...]

PR #6872, opened 2026-09-05
fix(#6857): string_pass answered from cache for files it never opened — probe through the same hardened path, at a measured +47% on warm scans

### LINKED ISSUE #6857: string_pass's loud-failure guarantee is bypassed on a cache hit — stat-only validation returns extractions for a file the hardened read never opened
`string_pass` is the *one* hardened read in `internal/links` that fails loudly instead of silently skipping — that is the whole reason #6839 treats it as the reference behaviour. But the loud path sits **behind a cache that can satisfy the call without ever reaching it**.

## The shape

`scanFile` (`internal/links/string_pass.go:261`):

```go
fi, err := os.Stat(absPath)          // :270
...
h := sha256.Sum256([]byte(absPath))  // :284   key is the PATH, not the content
cacheFile := ...
if b, err := os.ReadFile(cacheFile); err == nil {
    if e.Mtime == fi.ModTime().UnixNano() && e.Size == fi.Size() {
        return e.Values, nil          // :289  ← returns here
    }
}
body, err := readSourceFile(absPath, stringScanMaxFileBytes)  // :296  the hardened read
```

The cache entry is keyed on the **absolute path** and validated on **mtime + size only**. On a
hit, `scanFile` returns before `readSourceFile` is called at all — so `safeio`'s refusal, and the
error propagation that makes this site the loud one, do not happen.

## What this does and does not mean

**It is narrower than "the pass is deaf on a re-index."** A cache hit requires mtime and size to
still match, and an ordinary unchanged regular file is still readable — so in the common warm-cache
case the skipped read costs nothing. The hole is the case where **mtime and size match while the
file is no longer a regular readable file**: the path is replaced by a non-regular file that
preserves both (size 0 is the easy half; mtime is settable), or the entry is stale for any other
reason the two fields cannot distinguish. That is a narrow window and I am not claiming a corpus
has hit it.

**What is not narrow is the property it breaks.** #6839's whole framing rests on `string_pass`
being the site that cannot silently under-read. On a cache hit that guarantee is not weakened, it
is **absent** — the pass returns cached extractions for a path it did not open, and the caller
cannot tell that apart from a successful read. Under `stat`-only validation, "this file's contents
were classified" is an inference, not an observation.

`os.Stat` also follows symlinks and applies no symlink policy, whereas the hardened read does — so
the two paths do not even agree on what they would accept.

## Why this is filed separately from #6839

#6839 is about ten sites mapping a read failure to "nothing found". This is the eleventh site
mapping a read it never performed to "here is what the file contained". Different defect, and
fixing the ten does not touch it.

## Not established

- Whether any real corpus produces a mtime+size-preserving replacement. Unmeasured, and I would
  not gate this on that — the cheaper argument is that the validation is weaker than the property
  the pass is documented to have.
- Whether content-hash validation is affordable here. It would mean reading the file, which is
  most of what the cache exists to avoid — so the fix is more likely to be *"say that the cache
  hit is an inference"* (or validate through the same hardened path) than to re-hash everything.

## Acceptance

A test in which a cached entry satisfies `scanFile` for a path the hardened read would refuse, and
which **fails before the fix**. Assert the emitted artefact — the extractions the pass returns —
not a cache-hit counter. Found while measuring #6839 arm 1; verified against the source at
`1ade9c017` rather than taken from the arm-1 report, whose phrasing was broader than the code
supports.

cajasmota:
`string_pass` is the one hardened source read in `internal/links` that fails loudly instead of mapping a read failure to "nothing found" — #6839 treats that as the reference behaviour, and it is the whole reason the other ten sites are considered the deviation. `scanFile` could return before the hardened read ever ran: the string-scan cache is keyed on the **absolute path** and validated on **mtime + size alone**, so a hit returned extractions for a path the pass never opened, and no caller could tell that apart from a successful read.

## Measured first

The issue asks for the reachable window to be established rather than assumed, and this repo has twice built a guard for a population that was already handled. `scanFile` stats **before** it consults the cache, which rules out most of the obvious shapes — verified empirically against the pre-fix tree, not argued:

| shape | reaches the cache? |
|---|---|
| directory at the path | no — `fi.IsDir()` is already a silent skip |
| path removed (`ENOENT`) | no — mapped to a skip at the stat |
| self-referential symlink (`ELOOP`) | no — the stat itself errors and propagates |
| **regular file that is no longer readable** | **yes** |
| **FIFO / socket / device at the path** | **yes** |

So the window is real, and it is these two:

- **Unreadable regular file.** `chmod` changes neither mtime nor size, so nothing has to be replaced at all: the entry validates and the pass hands back confident extractions for a file whose open fails with `EACCES` — not `ErrNotRegular`, and therefore exactly the error this site's contract says must propagate.
- **Non-regular, non-directory file.** `os.Stat` succeeds, `IsDir` is false, size is 0 and mtime is settable, so a stamped entry validates against a named pipe (or a unix socket — same class).

Both are asserted on the **emitted artefact** — the extractions `scanFile` returns — not on a cache-hit counter, because this package has already been burned by a test that graded the helper while the real call site went unobserved (#6450 Task 2).

## The fix

Validate the hit through the **same hardened path the read uses, minus the read**. `probeSourceFile` opens through `safeio` with the same symlink policy and closes immediately, and the cache branch repeats the read arm's error mapping (`ErrNotRegular` → skip, everything else propagates) — deliberately **duplicated rather than shared**, because the two arms must agree and a caller must not be able to tell which one answered.

**Rejected — content-hash validation:** it means reading the file, which is most of what the cache exists to avoid.
**Rejected — surfacing the hit as an inference to the caller:** it labels the gap rather than closing it, changes `scanFile`'s signature and every caller, and still leaves the pass returning extractions for a path it could not open.

## What it costs — measured, because "one `open(2)`" understates it

The open drags `safeio.openWithDeadline`'s harness with it: a goroutine, two `time.Timer`s, a buffered channel, a `sync.Once`, a mutex and a 64-slot semaphore acquire. Over 2000 fully-warm files, 3×20 iterations:

```
pre-fix:   544 / 589 / 608 ms    52,042 allocs   (~290 µs/file)
post-fix:  788 / 838 / 927 ms    86,043 allocs   (~425 µs/file)
```

**+47% wall clock on a fully-warm scan, +17 allocations per hit exactly**, all of it on `scanRepo`'s serial critical path. Accepted rather than hidden: correctness beats warm-scan throughput at the one site whose contract is that it does not infer — but a future reader optimising this loop should know what they are paying for, which is why the numbers are in the code comment too.

## A new failure mode, recorded in the code

The probe routes warm cache hits through safeio's **process-wide 64-slot semaphore**, so a warm hit can now fail with `ErrWouldBlock` — which this arm propagates, and which per the read arm's own comment aborts the **entire group link run** (HTTP, drift, taint and reachability never run). Pre-fix a warm hit could not do that at all: it touched no descriptor. The slots are shared with `internal/secrets`, `internal/install/detect` and `internal/mcp`, so validating hits roughly doubles the opens contending for them.

Propagating is nonetheless the deliberate choice, for the reason the read arm already gives: swallowing `ErrWouldBlock` on a warm hit would reintroduce, one branch lower, exactly the inference this change removes. The inheritance is now written down at the cache branch rather than left implicit.

## Both directions

A legitimate warm hit must still be **served**. The controls seed a cache entry whose value never appears on disk, so a "fix" that disables the cache, or that re-reads and re-classifies on a hit, returns the on-disk value and fails. A stale entry must still fall through to the read, and a warm hit through a symlink to a regular file must still be served — the read this validation stands in for accepts those, and a stricter probe would silently drop them from the pass.

Each edit applied with an asserted exact match count, `go vet` clean, whole-package `-count=1`:

| # | edit | verdict |
|---|---|---|
| M1 | delete the probe on a cache hit (pre-fix behaviour) | **DEAD** — both negative tests |
| M2 | map every probe error to a skip | **DEAD** — the unreadable file |
| M3 | drop the `ErrNotRegular` skip arm | **DEAD** — the FIFO |
| M4 | drop the cache-hit return (cache effectively disabled) | **DEAD** — both warm-cache controls |
| M5 | probe with `RejectSymlinks` | **DEAD** — the symlinked warm hit |

## Which CI leg actually grades this

Both negative tests are POSIX-shaped: the FIFO file is `//go:build !windows`, and the chmod test skips wherever the file mode is not enforced, which includes Windows. **On Windows only the positive controls run** — deleting the probe there would be ALIVE. Per `test.yml` a plain PR runs **ubuntu + windows**, so **Ubuntu is the sole leg grading this fix**; the `ci:full` label adds macOS, which enforces file mode and has `mkfifo` and is therefore a second real grading leg rather than a compile check.

## One thing worth flagging — the guard caught it

An earlier draft extracted the cache-filename derivation into a helper. That removed the `".json"` literal from `scanFile` and dropped the site from `internal/safeio`'s name-chosen-open sweep — and `TestNoUnguardedNameChosenOpens` **failed loudly, naming `internal/links/string_pass.go:scanFile`**. Its stale-entry arm did exactly its job on a refactor nobody would otherwise have thought to re-check. Reverted; the test re-derives the filename itself, with a comment saying why, and a drift there cannot pass silently either (`[:32]` → `[:31]` dies on both warm-hit controls).

## Verification

`go test -count=1` green on `./internal/links/`, `./internal/safeio/` and `./cmd/grafel/`; `gofmt` clean. Not run: everything else, including `./internal/mcp/`.

Arm 3 of #6839 — counting skips at the other eight sites — is deliberately untouched.

Closes #6857

### DIFF
--- a/internal/links/safe_source_read.go
+++ b/internal/links/safe_source_read.go
@@ -69,3 +69,27 @@ const stringScanMaxFileBytes int64 = 4*1024*1024 + 1
 func readSourceFile(abs string, maxBytes int64) ([]byte, error) {
 	return safeio.ReadFile(abs, safeio.FollowSymlinks, maxBytes)
 }
+
+// probeSourceFile answers "would readSourceFile accept abs?" without reading
+// it: same package, same symlink policy, same two-layer type gate — it just
+// closes the descriptor instead of draining it.
+//
+// It exists for the string-scan cache (#6857). A cache hit there is validated
+// on mtime+size, which says the file has not changed; it does not say the path
+// still holds a regular file that opens. Returning cached extractions on that
+// evidence alone makes "the contents were classified" an inference, at the one
+// site in this package whose contract is that a read failure is loud. This is
+// the cheapest way to keep that contract on the cache path: one open(2) and a
+// close, not a re-read and a re-hash of the file the cache exists to avoid
+// reading.
+//
+// The bound readSourceFile takes is absent on purpose — nothing is read, so
+// there is nothing to bound, and safeio's own stat gate plus O_NONBLOCK are
+// what keep the open itself from parking on a FIFO.
+func probeSourceFile(abs string) error {
+	f, err := safeio.Open(abs, safeio.FollowSymlinks)
+	if err != nil {
+		return err
+	}
+	return f.Close()
+}
--- a/internal/links/string_pass.go
+++ b/internal/links/string_pass.go
@@ -287,6 +287,67 @@ func scanFile(absPath, relPath, cacheDir string) ([]Extraction, error) {
 			var e scanCacheEntry
 			if json.Unmarshal(b, &e) == nil {
 				if e.Mtime == fi.ModTime().UnixNano() && e.Size == fi.Size() {
+					// #6857: mtime+size say the file has not CHANGED; they do
+					// not say it is still a file this pass would agree to
+					// read. os.Stat above follows symlinks and applies no
+					// type or openability policy, so a named pipe, socket or
+					// device at the path (size 0, mtime settable), or a
+					// regular file that is merely no longer readable (chmod
+					// changes neither field), all validate. Returning here
+					// without asking would make "this file's contents were
+					// classified" an inference — and this site is precisely
+					// the one whose contract is that it does not infer: it is
+					// the single read in this package that fails loudly
+					// rather than mapping a failure to "nothing found"
+					// (#6839).
+					//
+					// So the hit is validated through the SAME hardened path
+					// the read uses, minus the read: probeSourceFile opens
+					// and closes, which costs one open(2) and answers exactly
+					// the questions readSourceFile would have. Re-hashing the
+					// contents would answer more, and would cost the whole
+					// read — which is most of what this cache exists to
+					// avoid.
+					//
+					// The error mapping is the read arm's, deliberately
+					// duplicated rather than shared: the two arms must agree,
+					// and a caller cannot be able to tell which one answered.
+					//
+					// WHAT THIS ARM NOW INHERITS, stated because it is NEW
+					// here rather than merely copied. The probe goes through
+					// safeio's process-wide 64-slot semaphore, so a warm
+					// cache hit can now fail with safeio.ErrWouldBlock — and
+					// this arm propagates it, which per the read arm's own
+					// comment below aborts the ENTIRE group link run, not
+					// just this repo. Before this change a warm hit could not
+					// do that at all: it touched no descriptor. The slots are
+					// shared with internal/secrets, internal/install/detect
+					// and internal/mcp, and validating hits roughly doubles
+					// the opens contending for them.
+					//
+					// Propagating anyway is the deliberate choice, for the
+					// same reason the read arm gives: this is the one site in
+					// this package that does not map a failure it could not
+					// rule out to "nothing found". A warm hit that silently
+					// swallowed ErrWouldBlock would be exactly the inference
+					// #6857 removed, reintroduced one branch lower.
+					//
+					// The cost is not free either, and "one open(2)" would
+					// understate it: the open drags openWithDeadline's
+					// harness with it — a goroutine, two timers, a buffered
+					// channel, a sync.Once, a mutex and the semaphore
+					// acquire. Measured over 2000 fully-warm files, 3x20
+					// iterations: 544ms -> 788ms wall clock (+47%) and +17
+					// allocations per hit, all of it on scanRepo's serial
+					// critical path. Correctness beats warm-scan throughput
+					// here, but a future reader optimising this loop should
+					// know what they are paying for.
+					if err := probeSourceFile(absPath); err != nil {
+						if errors.Is(err, safeio.ErrNotRegular) {
+							return nil, nil
+						}
+						return nil, err
+					}
 					return e.Values, nil
 				}
 			}
--- a/internal/links/string_pass_cache_6857_test.go
+++ b/internal/links/string_pass_cache_6857_test.go
@@ -1,0 +1,175 @@
+package links
+
+import (
+	"crypto/sha256"
+	"encoding/hex"
+	"encoding/json"
+	"os"
+	"path/filepath"
+	"testing"
+)
+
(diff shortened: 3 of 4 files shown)


cajasmota:
## Coordinator verification — the cost this PR now documents is graded by nothing

The review scored ten mutants DEAD and I did not repeat any of them. Mine went at the thing the review's F1 exposed: this PR measured a real cost, wrote the number into the source, and **nothing observes where that cost is incurred.**

The probe lives inside the cache-hit branch. I hoisted it to run on **every** `scanFile` call — after the `IsDir` gate, before the cache lookup, with the identical error mapping:

```go
	if fi.IsDir() {
		return nil, nil
	}
+	if err := probeSourceFile(absPath); err != nil {
+		if errors.Is(err, safeio.ErrNotRegular) {
+			return nil, nil
+		}
+		return nil, err
+	}
```

`go vet` clean, exact match count asserted. `go test -count=1 ./internal/links/` → **ok. ALIVE.**

### What that does and does not mean

**Behaviourally it is equivalent**, deliberately so — I placed it after the stat so `ENOENT` still maps to a skip and the error classes still land in the same arms. I am recording it as *equivalent under the current suite*, not as a defect, and I am **not asking for a change**.

What it costs is the point. Under this mutant every cache **miss** pays the probe too — a double open on the cold path — and the +47% the comment now attributes to warm hits silently becomes a cold-path cost as well. The suite cannot tell the two apart.

### Why it is worth writing down rather than fixing

The comment added in this round says: *"a future reader optimising this loop should know what they are paying for."* That reader will move this call. When they do, every test still passes — the numbers in the comment become wrong, and nothing says so.

Pinning it properly would mean asserting an open-count, which needs a seam that does not exist and which I am not going to manufacture for a cost property. **The honest state is: correctness is graded in both directions and at both error classes; placement is not.** Better stated than quietly assumed — this is the same shape as the `MEASURED` paragraph in `file_carrier.go` (#6861), one abstraction level down: a measurement written into a comment, load-bearing for a future decision, observed by nothing.

### On the two platform legs

Both negative tests are POSIX-shaped, so Ubuntu was the sole leg grading this on a plain PR. I have added **`ci:full`**, which makes macOS a second real grading leg rather than a compile check — it enforces file mode and has `mkfifo`, so both negatives actually execute there.

Issue #2370, opened 2026-05-26
[tech-debt][PR-2365-followup] Add language field to graph.fbs Entity table — retire property-tunnel workaround


cajasmota:
## Context
PR #2365 fixed #2341 by tunneling Language through `Properties["language"]` in the FlatBuffers serializer (since `graph.fbs`'s `Entity` table has no `language` slot). Works, but it's a workaround — adding the proper field is cleaner.

## Fix
1. Add `language: string` to `Entity` table in `internal/graph/schema/graph.fbs`
2. Run `flatc` codegen (creates new generated Go bindings)
3. Bump `FormatVersion` so old `graph.fb` files trigger reindex
4. Update `buildEntity` to write directly to the slot
5. Update `fbEntityToGraphEntity` to read directly from the slot
6. Drop the property-tunnel workaround
7. Test backward compat: old graph.fb files should still load (with empty Language) via the load fallback OR trigger reindex via FormatVersion bump

## Effort / Priority
M-L (cross-cutting: schema + codegen + readers). Priority: low — current workaround is correct and tested.

Refs: PR #2365.

PR #6876, opened 2026-09-05
fix(#6862): the whole-graph digest did not hash entity Language — a field with its own FlatBuffers slot, blind in two of four measured arms

### LINKED ISSUE #6862: cmd/grafel's whole-graph digest does not hash entity Language — the broad net every extractor change must run is blind to a field with its own FlatBuffers slot
`cmd/grafel`'s whole-graph digest — the broadest safety net this repo has, and the one every extractor change is required to run — does not hash entity `Language`.

## Verified, not inferred

`cmd/grafel/custom_extractor_spans_6118_test.go:393`:

```go
func entityTupleKey(e *graph.Entity) string {
	return fmt.Sprintf("%s|%s|%s|%s|%s|%d|%d",
		e.Kind, e.Name, e.QualifiedName, e.Subtype, e.SourceFile, e.StartLine, e.EndLine)
}
```

Seven fields. `Language` is not among them — and neither are entity properties, which is already
recorded in the coordinator notes.

`Language` is not incidental state: it is a **dedicated top-level FlatBuffers slot** persisted by
`fbwriter/writer.go:167-182` (#2370), i.e. it round-trips into the shipped graph exactly like the
fields the digest does hash.

So a change that altered the language token on **every entity a given extractor emits** would move
the digest by zero, over a fixture that deliberately spans C#, Java, TypeScript, JavaScript, Python,
Go and Rust.

`cmd/grafel`'s digest occupies a specific role: nothing in `internal/extractors/...` points at it, so
a diff-derived package set never reaches it, and the standing rule is that **any** extractor change
must run it. It is treated — including by me, in briefs — as the broad net that catches what
narrower package tests miss. It is narrower than that in a way that is not written down anywhere it
would be read.

And this is not a hypothetical field. `internal/extractor/file_carrier.go` explicitly stamps `lang`
rather than taking it from `FileInput.Language` to avoid "**proto's #6356 trap**" — a shipped defect
that was precisely a language-token disagreement. The field has produced real bugs, and the broad
net does not see it.

Concretely, this surfaced while reviewing #6852's first language: a mutant setting the carrier's
`lang` to `""` is DEAD in `internal/extractors/bicep` — but it would have been ALIVE in the digest
even if the fixture had contained a `.bicep` file, because the digest cannot observe the field.

## What is NOT established

**Whether anything else grades a repo-wide language regression.** I have not measured that, and the
measurement is the first half of this issue, not an afterthought — this repo has twice filed
"nothing catches X" claims that were false because a package was never run, and once filed a guard
request that the measurement showed was already satisfied twelve times over.

Per-extractor language assertions clearly exist (`TagEntitiesLanguage` has its own test; several
extractors assert their own token). The open question is whether any **whole-graph** assertion would
notice an extractor's entities all changing language at once.

## Suggested shape

1. **Measure first.** Determine what, if anything, fails when a registered extractor's emitted
   `Language` is mutated wholesale. Report the state either way — if something already covers it,
   that is the answer and this issue closes.
2. Only if nothing does: add `Language` to `entityTupleKey`. That will move the pinned digest
   constant, so **enumerate the fixture members producing the delta** rather than bumping the
   number — the standing rule for that constant.
3. Decide separately, and state, whether entity **properties** belong in the tuple. That is a much
   larger blast radius and should not be bundled in.

Found while reviewing #6852 (bicep). Verified against `3bcb17807` by reading `entityTupleKey` and
`fbwriter/writer.go`, not by trusting the review that raised it.

cajasmota:
cmd/grafel's semantic digest is the broadest net this repo has and the one every
extractor change is required to run. entityTupleKey hashed seven fields and
Language was not among them, even though Language is a dedicated top-level
FlatBuffers slot (fbwriter, #2370) that round-trips into the shipped graph
exactly like Kind and SourceFile.

MEASURED FIRST, per the issue. Four wholesale language mutants at b0951cb0a,
each run against ./cmd/grafel/ + the producing package + ./internal/extractors/
+ ./internal/quality/ + ./internal/resolve/ + ./internal/graph/... +
./internal/mcp/ + ./internal/dashboard/ + ./internal/extractor/:

  python  "python"->"zzlang" (2 TagEntitiesLanguage calls + 8 `Language:`
          literals): digest UNCHANGED. Caught only by
          TestLanguageTagPersistsAfterFBRoundtrip, which hardcodes
          `.py` -> "python" and so covers exactly one language, and by
          TestContentParity_PathA_6129.
  go      "go"->"zzlang" (1 tag call + 6 literals): digest moved, but at
          141/286 instead of 142/287 — a STRUCTURAL side effect of the token,
          not the field being observed.
  bicep   `const lang` -> "zzlang": nothing outside internal/extractors/bicep
          failed.
  engine  detector.go's three `Language: file.Language` rule-emit sites ->
          "zzlang": nothing outside internal/engine failed.

So the gap is real: the digest cannot see the field at all, and two of the four
arms had no grader outside the package that produced them.

THE FIX. entityTupleKey hashes e.Language. The graph does not move: counts stay
at 142/287, no entity is added, deleted, re-kinded, renamed or re-positioned,
no edge is rewired. This re-pin therefore has no "fixture members producing the
delta" to enumerate — the delta is one field appended to all 142 entity lines —
so the comment enumerates the census the tuple now observes instead
(python 39, csharp 32, empty 31, typescript 11, java 10, rust 9, javascript 6,
go 4). The 31 empty ones are all SCOPE.External / Module synthetics with no
source file; nothing that came out of a file is language-less.

Entity Properties remain unhashed. That is a separate decision with a much
larger blast radius and is deliberately not bundled here.

GRADED, not just re-pinned. TestSemanticDigestGradesEntityLanguage6862 asserts
the consequence in three parts — the tuple distinguishes two entities differing
only in Language; the fixture still spans >= 5 languages and >= 100 tagged
entities; and re-stamping every Python entity's language in the loaded gate-OFF
graph changes the digest. It keeps holding across future re-pins of the
constant. With the field back out of the tuple, parts 1 and 3 each fail
independently, and the digest test hits a dedicated gateOffDigest6742 branch
that names the cause. Re-run of the python mutant against the new tuple fails
the digest at entities=142 relationships=287 — the field itself, with the
structure held constant.

ONE TUPLE, NOT TWO. entityTuples (inproc_custom_extractors_5989_test.go)
inlined its own copy of the same seven fields while
TestCustomExtractorAddedEntitiesCarryACompleteSpan resolved its output through
a map keyed by entityTupleKey. Adding a field to one desynced them immediately.
entityTuples now calls entityTupleKey, and the three tuple parsers that read
StartLine/EndLine off fixed indices 5/6 read the trailing pair instead.

Two of those parsers fed "no span may narrow" checks that skip silently on a
parse failure — verified vacuous by a probe that makes the parser match
nothing, which both tests passed. They now carry a vacuity guard; the same
probe kills both.

Closes #6862

---

### DIFF
--- a/cmd/grafel/custom_extractor_spans_6118_test.go
+++ b/cmd/grafel/custom_extractor_spans_6118_test.go
@@ -390,9 +390,25 @@ func TestCustomExtractorGateDoesNotRegressFileComponentStartLines(t *testing.T) 
 
 // entityTupleKey is the same content tuple entityTuples aggregates on, for a
 // single entity.
+//
+// LANGUAGE IS PART OF THE TUPLE (#6862). Language is a dedicated top-level
+// FlatBuffers slot (fbwriter/writer.go, #2370) — it round-trips into the
+// shipped graph exactly like Kind and SourceFile do, and grafel_find
+// --language reads it. Before #6862 the tuple hashed seven fields and Language
+// was not among them, so an extractor that re-stamped the language token on
+// EVERY entity it emits moved this digest by zero. That was measured, not
+// assumed: mutating the Python extractor's token wholesale
+// ("python" -> "zzlang", both TagEntitiesLanguage calls plus all eight direct
+// `Language:` literals) left this digest byte-identical. The Go extractor's
+// equivalent mutant did move it, but only via a structural side effect
+// (141/286 instead of 142/287) — the field itself was still unobserved.
+//
+// Entity Properties are still NOT hashed. That is a deliberately separate
+// decision with a much larger blast radius (#6862 declines to bundle it).
 func entityTupleKey(e *graph.Entity) string {
-	return fmt.Sprintf("%s|%s|%s|%s|%s|%d|%d",
-		e.Kind, e.Name, e.QualifiedName, e.Subtype, e.SourceFile, e.StartLine, e.EndLine)
+	return fmt.Sprintf("%s|%s|%s|%s|%s|%s|%d|%d",
+		e.Kind, e.Name, e.QualifiedName, e.Subtype, e.SourceFile, e.Language,
+		e.StartLine, e.EndLine)
 }
 
 // semanticDigest6118 is a stable content digest of a graph: entities as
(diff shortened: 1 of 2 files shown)


cajasmota:
## Coordinator verification — the de-duplication is structurally held, not merely tidy

The review scored M1, M1b, M2 and the M2-control; I did not repeat any of them. Mine went at the extension rather than the fix: this PR removed a **second copy** of the tuple, and "two copies that agree today" is the setup for divergence later — a risk I have raised on another PR this week and never actually scored.

So I put the second copy back. In `entityTuples`, replacing the shared call with an inlined tuple of the original seven fields:

```go
out[entityTupleKey(e)]++
  →  out[fmt.Sprintf("%s|%s|%s|%s|%s|%d|%d",
         e.Kind, e.Name, e.QualifiedName, e.Subtype, e.SourceFile, e.StartLine, e.EndLine)]++
```

Exact match count asserted, `go vet` clean. `go test -count=1 ./cmd/grafel/` → **exit 1. DEAD**, on four tests:

```
--- FAIL: TestCustomExtractorAddedEntitiesCarryACompleteSpan
--- FAIL: TestInProcCustomExtractorsJavaEnrichment
--- FAIL: TestInProcCustomExtractorsSupersedeIsNonDestructive
--- FAIL: TestInProcCustomExtractorsThreeCollisionShapesAreClosed
```

**That is the good answer, and it is structural rather than lucky.** `TestCustomExtractorAddedEntitiesCarryACompleteSpan` looks its output up in a map keyed by `entityTupleKey`, so the two producers are joined by a lookup that fails the moment their formats differ. The copies cannot drift apart silently — a divergence is a miss, not a mismatch nobody checks.

Worth stating because it is the opposite of the usual finding here. The comment at `:270-274` says two copies of one format is a hazard; the tests make that hazard *loud*, which is what a de-duplication should be able to claim and usually cannot.

## On the two residual items from review

Neither blocks, and I am not asking for changes:

- The new vacuity guards are conditioned on `len(destroyed) > 0`, so if the gate went inert on those fixtures the guard would fall silent again. Pre-existing shape, not introduced here.
- Part 2's `tagged < 100` floor sits at 111 — an 11-entity margin. Thin, but a floor derived from a measurement is still the right form.

Both are recorded here rather than filed, since neither is a defect in this change.

PR #6877, opened 2026-09-05
fix(#6839): the last nine hardened reads in internal/links kept their skip but never reported it — bounded and REPORTED, as safeio requires

### LINKED ISSUE #6839: Ten of eleven hardened reads in internal/links mapped a read failure to "nothing found" — one turned it into a positive "this code is dead" claim; all three arms shipped

cajasmota:
## Nine reads across eight passes — the two numbers count different things

Re-derived from source rather than from arm 1's table (`grep -rn readSourceFile internal/links/`,
non-test): eleven hardened reads, of which

- 1 is **group A** — `string_pass.go:357`, propagates by design, untouched here;
- 1 is **group B** — `reachability.go:308`, fixed in #6863, untouched here;
- **9 are group C** — this PR.

The issue says "the other eight". Arm 1 was not off by one — its group-C row enumerates eight
**passes**, written as `dataflow ×2`, so the 8 was never a site count. Counted as reads, group C is
**7 `continue` + 2 cache-`""` = 9 reads across 8 passes**. Both numbers are right about different
things; this PR touches nine reads.

| # | site | pass |
|---|---|---|
| 1 | `constant_propagation.go` (in `buildResolver`) | constant_propagation |
| 2 | `effect_propagation.go` | effect_propagation |
| 3 | `taint_flow.go` | taint_flow |
| 4 | `payload_drift.go` (in `sniffPayloadShapes`) | payload_drift |
| 5 | `def_use_pass.go` | def_use |
| 6 | `template_pattern_pass.go` | template_patterns |
| 7 | `complexity_pass.go` (cache-`""`) | complexity |
| 8 | `dataflow_pass.go` in-file read | data_flow |
| 9 | `dataflow_pass.go` in `resolveCrossFileFlows` (cache-`""`) | data_flow |

## What changed

Every one of the nine **keeps its skip** — no abort path was added anywhere, and no pass returns an
error it did not return before. Not *no* behaviour change, though: `noteUnreadableSource` also emits
an **uncapped stderr warning per unreadable read**, matching `reachability.go`'s convention, so a
wholly unreadable tree now prints one line per file where it printed none. What each site now does is **record** the skip, through the channel
arm 2 already built: `PassResult.UnreadableSourceFiles`, serialised as `unreadable_source_files` in
`<group>-link-pass-stats.json`. No second channel was invented.

One shared helper, `noteUnreadableSource` in `safe_source_read.go`, carries the reasoning once
(rather than nine times, which is how eleven divergent copies of this read happened in the first
place) and returns `1`/`0` so a site reads `unreadable += noteUnreadableSource(...)` directly above
the `continue` it already had. It also warns on stderr, matching `reachability.go`.

**It counts READS, not distinct files,** and the docs now say so — they previously disagreed with
each other. Measured: one unreadable `src/svc.ts` reached both from `data_flow`'s top-level
`fileSet` and as a cross-file callee counts **2** in a single run, because `resolveCrossFileFlows`'s
cache is per call. An unreadable callee shared by N handler files counts N. Behaviour is fine; read
the field as an upper bound on lost input, not a count of paths.

`fs.ErrNotExist` does **not** count, matching the field's existing doc: under a root that exists, a
file the graph names but disk does not hold is a known fact, not a hidden one. Without that
exemption the counter would fire on every ordinary re-index over a changed tree and mean nothing.
The consequence is now written into the helper doc rather than left to be discovered: a source root
that has **moved** makes every file `ErrNotExist`, so group C reports **zero** for that repo while
producing nothing. That is the one scenario where this field is silent by design; the loud signal
for it is reachability's `degraded_repos`, which stats the root.

Two signatures moved, both because the count could not otherwise survive an early return:

- `sniffPayloadShapes` gains a third return — `runPayloadDriftPass` returns early on `scanned == 0`,
  which is exactly the wholly-unreadable case.
- `buildResolver` gains a second return rather than a field on `*Resolver`, because it returns
  **nil** when it read no file at all — again, precisely the case that must still be reported.
  Six test call sites updated mechanically.
- `resolveCrossFileFlows` gains a second return for the cross-file callee read.

## Assert the emitted artefact — and a retraction

**Retracted:** an earlier draft of this PR claimed M12 (dropping the serialisation) was arm 2's
still-ALIVE mutant, now killed here. It was **already DEAD at `HEAD~1`**.
`reachability_degraded_test.go:159-181`, shipped *in* arm 2's own commit `1d1cb47e9`, already parses
the written stats file and asserts `unreadable_source_files == 1`; that file and `links.go` are
byte-identical between `HEAD~1` and `HEAD`. Arm 2's review closed that gap. The assertion added here
is a legitimate independent second grader over eight more pass rows — it did not close anything.

What the episode still dictates is the method: arm 2's field spent part of its life written and
never observed, so the assertions here land on the **emitted artefact** — `unreadable_source_files`
parsed out of the written `link-pass-stats.json` by its documented JSON name — not only on the
struct counter. And the same lesson one level up: **score the call site, not the helper.**

Stated plainly rather than implied: **no MCP tool projects `unreadable_source_files` today.** It is
an operator-facing record. That is now written into the helper's doc as well as the field's, so the
next reader does not infer a consumer that does not exist.

## Tests — three directions, not one

`internal/links/unreadable_source_count_6839_test.go`, a table over all eight group-C passes plus a
dedicated case for the ninth read:

1. **unreadable → counted.** A **directory** planted at the read path inside `t.TempDir()`
   (`safeio.Open` stats first → `ErrNotRegular`). No FIFO, no socket, nothing outside the temp dir,
   no dependence on permissions.
2. **readable → counts nothing.** A fix that reports unconditionally passes (1) and is useless;
   only this catches it.
3. **absent → counts nothing.** The third case the field's own doc distinguishes.
4. **the serialised artefact** carries `unreadable_source_files: 1` for every one of the eight pass
   rows.
5. `TestCrossFileFlowUnreadableCalleeIsCounted` drives `resolveCrossFileFlows` directly with the
   handler file **readable** and only the callee file planted — a fix that double-counted the
   handler read would not pass it.
6. `TestDataFlowPassReportsTheCrossFileCalleeRead` grades the **call site** rather than the helper:
   it runs `runDataFlowPass` end to end and asserts the cross-file count reaches `PassResult` and the
   serialised stats file, in both directions.

Nothing here is platform-conditional: a directory-at-the-path fails identically on Windows and
Unix, and no test provokes `ErrWouldBlock` (unreachable here without a FIFO, per arm 1).

## Mutation score

Every mutant: exact edit, asserted match count proving it landed, `go vet` clean, whole-package
`go test -count=1 ./internal/links/`. Permissive direction first.

**Per-site — stop reporting the skip, keep the skip** (`unreadable += ...` → `_ = ...`, so the
warning still prints and only the count is lost). This is also the red-before-green evidence: each
mutant is the pre-change state of exactly one site.

|---|---|---|
| S1 `complexity_pass.go` | **DEAD** | `…CountAnUnreadableSourceFile/complexity` + sidecar |
| S2 `dataflow_pass.go` in-file | **DEAD** | `…/data_flow` + sidecar |
| S3 `dataflow_pass.go` cross-file | **DEAD** | `TestCrossFileFlowUnreadableCalleeIsCounted` |
| S4 `taint_flow.go` | **DEAD** | `…/taint_flow` + sidecar |
| S5 `def_use_pass.go` | **DEAD** | `…/def_use` + sidecar |
| S6 `payload_drift.go` | **DEAD** | `…/payload_drift` + sidecar |
| S7 `constant_propagation.go` | **DEAD** | `…/constant_propagation` + sidecar |
| S8 `template_pattern_pass.go` | **DEAD** | `…/template_patterns` + sidecar |
| S9 `effect_propagation.go` | **DEAD** | `…/effect_propagation` + sidecar |

Each mutant is killed by its OWN sub-test, not by a neighbour — nine sites, nine independently
graded rows.

**Behavioural mutants:**

|---|---|---|
| M10′ narrow the exemption `fs.ErrNotExist` → `fs.ErrPermission` (i.e. start counting absent files) | **DEAD** | `TestGroupCPassesDoNotCountAnAbsentFile` (all 8) + the byte-identical golden |
| M11 `res.UnreadableSourceFiles = 1` unconditionally in `taint_flow` | **DEAD** | `TestGroupCPassesCountNothingOnAReadableTree/taint_flow` |
| M12 drop `UnreadableSourceFiles:` from `writeLinkPassStats` | **DEAD** | `TestGroupCUnreadableCountReachesTheStatsSidecar` **and** arm 2's own `TestReachabilityUnreadableEntryPointFileDoesNotClaimDead` |
| M13 move the report BELOW `effect_propagation`'s `scannedFiles == 0` early return | **DEAD** | `…CountAnUnreadableSourceFile/effect_propagation` |
| M14 move the report BELOW `constant_propagation`'s nil-resolver early return | **DEAD** | `…CountAnUnreadableSourceFile/constant_propagation` |

**Found ALIVE by review, now fixed and DEAD:**

|---|---|---|
| M13 discard `xunreadable` at the CALL SITE in `runDataFlowPass` | **DEAD** | `TestDataFlowPassReportsTheCrossFileCalleeRead` (new) |
| M16 drop `cache[rel] = ""` on complexity's failure arm | **DEAD** | `…CountAnUnreadableSourceFile/complexity` (fixture now has two entities in one file) |

Both were ALIVE in the first push and the review was right about each. **M13 is the important one:**
`TestCrossFileFlowUnreadableCalleeIsCounted` calls `resolveCrossFileFlows` **directly**, so site 9
was graded inside the helper and nothing observed the count reaching `PassResult` — the very
"written but never observed" shape this arm exists to end. Testing the helper does not pin the
caller. The new test runs `runDataFlowPass` over the `TestDataFlowPass_JSTS_CrossFile_DBWrite`
fixture with only `src/svc.ts` planted, and asserts **2**, not 1 — see the counts-reads note above.
M16 pinned prose that was true but unobserved: with one entity the fixture could not tell one read
per file from one per entity.

**16 applied, 16 DEAD, 0 ALIVE.** One earlier attempt at M10 (deleting the guard outright) was
discarded as a VET-FAIL rather than scored — it left `errors` imported and unused, so it never
compiled and graded nothing. M10′ is the compiling replacement.

## Verification

`go test -count=1 ./internal/links/`, `./internal/mcp/`, `./cmd/grafel/`.
**Not run:** every other package. No production symbol outside `internal/links` changed, and the
three changed signatures are all package-private.

Closes #6839.

### DIFF
--- a/internal/links/complexity_pass.go
+++ b/internal/links/complexity_pass.go
@@ -43,6 +43,8 @@ const MethodComplexity = "complexity"
 func runComplexityPass(graphs []repoGraph, _ Paths) (PassResult, error) {
 	res := PassResult{Pass: MethodComplexity}
 	stamped := 0
+	// #6839: scanned-source reads this pass skipped because they failed.
+	unreadable := 0
 
 	for ri := range graphs {
 		g := &graphs[ri]
@@ -66,6 +68,13 @@ func runComplexityPass(graphs []repoGraph, _ Paths) (PassResult, error) {
 			}
 			buf, err := readSourceFile(filepath.Join(srcRoot, rel), maxSourceFileBytes)
 			if err != nil {
+				// #6839 group C: skipping is the deliberate choice for a
+				// supplementary pass — the file's functions go unstamped,
+				// which is an absence rather than a wrong number. Record
+				// the skip so it is bounded and reported; see
+				// noteUnreadableSource. The cache entry keeps the count at
+				// one per file, not one per entity in it.
+				unreadable += noteUnreadableSource(res.Pass, g.Repo, rel, err)
 				cache[rel] = ""
 				return ""
 			}
@@ -114,5 +123,8 @@ func runComplexityPass(graphs []repoGraph, _ Paths) (PassResult, error) {
 	}
 
 	res.LinksAdded = stamped
+	// #6839: report the skipped reads, so a pass that stamped nothing
+	// still says WHY it stamped nothing.
+	res.UnreadableSourceFiles = unreadable
 	return res, nil
 }
--- a/internal/links/constant_propagation.go
+++ b/internal/links/constant_propagation.go
@@ -233,7 +233,11 @@ func (r *Resolver) resolveDepth(repo, file, ident string, depth int, seen map[st
 // count.
 func runConstantPropagationPass(graphs []repoGraph, paths Paths, rejects map[string]bool) (PassResult, error) {
 	res := PassResult{Pass: "constant_propagation"}
-	resolver := buildResolver(graphs)
+	resolver, unreadable := buildResolver(graphs)
+	// #6839: report the skipped reads BEFORE the nil-resolver early return —
+	// a tree whose every file is unreadable takes that return, and is exactly
+	// the case where "resolved nothing" needs to say why.
+	res.UnreadableSourceFiles = unreadable
 	if resolver == nil {
 		return res, nil
 	}
@@ -317,13 +321,19 @@ func runConstantPropagationPass(graphs []repoGraph, paths Paths, rejects map[str
 // buildResolver constructs the in-memory symbol table + import index from
 // the per-repo graphs. Returns nil when no T1-language entities are
 // present (the pass is a no-op for unsupported groups).
-func buildResolver(graphs []repoGraph) *Resolver {
+// The second return is the number of scanned-source reads it skipped because
+// they FAILED (#6839). It is a return value rather than a field on the
+// Resolver because buildResolver yields NIL when it read no file at all —
+// which is precisely what a wholly unreadable tree produces, and precisely the
+// case that still has to be reported.
+func buildResolver(graphs []repoGraph) (*Resolver, int) {
 	r := &Resolver{
 		bindings:      map[string]map[string]map[string]substrate.Binding{},
 		importsByFile: map[string]map[string][]importTarget{},
 		fileLookup:    map[string]map[string]string{},
 	}
 	totalFiles := 0
+	unreadable := 0
 	for _, g := range graphs {
 		// Collect unique source files referenced by any entity. Use a
 		// set so we sniff each file once even when many entities share it.
@@ -357,6 +367,12 @@ func buildResolver(graphs []repoGraph) *Resolver {
 				// File missing on disk (graph indexed from a prior
 				// snapshot, etc.) — skip; the pass is best-effort and
 				// must never fail the whole link pipeline.
+				//
+				// #6839 group C: skipping stays the deliberate choice —
+				// the file's bindings go unresolved, which loses edges
+				// rather than inventing them. Record the skip so it is
+				// bounded and reported; see noteUnreadableSource.
+				unreadable += noteUnreadableSource("constant_propagation", g.Repo, file, err)
 				continue
 			}
 			bindings := sniff(string(content))
@@ -396,9 +412,9 @@ func buildResolver(graphs []repoGraph) *Resolver {
 		}
 	}
 	if totalFiles == 0 {
-		return nil
+		return nil, unreadable
 	}
-	return r
+	return r, unreadable
 }
 
 // indexFileForLookup registers file under several canonical key forms so
--- a/internal/links/constant_propagation_test.go
+++ b/internal/links/constant_propagation_test.go
@@ -33,7 +33,7 @@ func TestResolverInFile(t *testing.T) {
 			SourceFile: "src/config.ts",
 		}},
 	}}
-	r := buildResolver(graphs)
+	r, _ := buildResolver(graphs)
 	if r == nil {
 		t.Fatal("expected non-nil resolver")
 	}
@@ -63,7 +63,7 @@ fetch(`+"`"+`${API_URL}/things`+"`"+`);
 			{ID: "e2", Name: "fetch", Kind: "SCOPE.Function", SourceFile: "src/app.ts"},
 		},
 	}}
-	r := buildResolver(graphs)
+	r, _ := buildResolver(graphs)
 	if r == nil {
 		t.Fatal("expected non-nil resolver")
 	}
@@ -116,7 +116,7 @@ func TestBackendSubstrateImportResolutionFixture(t *testing.T) {
 		},
 	}}
 
-	r := buildResolver(graphs)
+	r, _ := buildResolver(graphs)
 	if r == nil {
 		t.Fatal("expected non-nil resolver; substrate found no JS/TS bindings")
 	}
@@ -171,7 +171,7 @@ def go(): return API_URL
 			{ID: "e2", Name: "go", Kind: "SCOPE.Function", SourceFile: "pkg/client.py"},
 		},
 	}}
-	r := buildResolver(graphs)
+	r, _ := buildResolver(graphs)
 	if r == nil {
 		t.Fatal("expected non-nil resolver")
 	}
--- a/internal/links/dataflow_pass.go
+++ b/internal/links/dataflow_pass.go
@@ -76,6 +76,8 @@ func runDataFlowPass(graphs []repoGraph, paths Paths, rejects map[string]bool) (
 
 	var links []Link
 	scanned := 0
+	// #6839: scanned-source reads this pass skipped because they failed.
+	unreadable := 0
 
 	for ri := range graphs {
 		g := &graphs[ri]
@@ -116,6 +118,11 @@ func runDataFlowPass(graphs []repoGraph, paths Paths, rejects map[string]bool) (
 			abs := filepath.Join(srcRoot, file)
 			content, err := readSourceFile(abs, maxSourceFileBytes)
 			if err != nil {
+				// #6839 group C: skipping is the deliberate choice for a
+				// supplementary pass — this file's output is now MISSING
+				// rather than wrong. Record the skip so it is bounded and
+				// reported instead of silent; see noteUnreadableSource.
+				unreadable += noteUnreadableSource(res.Pass, g.Repo, file, err)
 				continue
 			}
 			scanned++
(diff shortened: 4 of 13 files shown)


cajasmota:
## Coordinator verification — the count is graded at all nine sites; the warning beside it is graded nowhere

Sixteen mutants DEAD across author and review, with each of the nine per-site mutants failing exactly **one** sub-test — no co-firing, which is the specific way a table-driven arm looks stronger than it is. I did not repeat any of them.

Mine went at the other half of what `noteUnreadableSource` emits. Every site passes its own pass name into the operator-facing warning; I made the helper name a different one:

```go
	fmt.Fprintf(os.Stderr,
		"grafel: warning: %s: %s: source file %s unreadable (%v); …",
-		pass, repo, file, err)
+		"zzWrongPass", repo, file, err)
```

Exact match count asserted, `go vet` clean. `go test -count=1 ./internal/links/` → **ok. ALIVE.**

### What that does and does not mean

The **count** is thoroughly graded — nine sites, the serialisation, the `ErrNotExist` exemption, the unconditional-report direction, and two early-return placements all die. What is not graded is the **message**, and this arm's entire subject is that a skip must be *reported* rather than silent. A report that names the wrong pass sends an operator chasing an unreadable file to the wrong place.

So the honest statement is narrower than "the skip is now reported": the *fact* of the skip is reported and graded; the *content* of the report is neither.

### Recorded, not required

Pinning stderr means substituting `os.Stderr` in a test, which is racy with parallel tests in this package and buys a diagnostic string. I am recording it as **equivalent under any affordable suite**, with the reason — the same call I made on #6857's probe placement, and for the same reason: manufacturing machinery to pin a non-behavioural property is worse than stating the gap.

Two things make me comfortable leaving it: nine call sites each pass a literal that matches their own pass, so the failure mode requires someone to edit the helper rather than drift into it; and the field itself — `unreadable_source_files` — carries the load-bearing information and *is* graded.

Worth one sentence in the helper's doc so the next reader knows the boundary: the returned count is graded, the warning text is not.

### One thing this arm got right that is easy to miss

The `fs.ErrNotExist` exemption means a **moved source root** reports zero for group C — every file is `ErrNotExist`, so the field is silent exactly when the tree is most broken. That consequence is now documented, with `degraded_repos` named as the loud signal instead. Documenting the scenario where your own new field says nothing is the kind of thing that normally only surfaces in an incident.

PR #6878, opened 2026-09-05
fix(#6868): backfill --check was NOT fully masked — validate skipped ungrouped records entirely; collapse the two predicates and pin their agreement

### LINKED ISSUE #6868: backfill --check was NOT masked: validate skipped ungrouped records entirely, so a taxonomy-declaring record with no cells was invisible to step 1/5
`backfill --check` is masked by `validate` — no whole-gate input can attribute a failure to it

## What was measured

While building `coverage check` (#6866) I needed one input per gate step that fails at *that* step
and no other, so each step could be graded independently. Four of the five have one. The backfill
step does not, and cannot.

Take a grouped record and delete a cell its subcategory taxonomy declares — the exact defect
`backfill --check` exists to catch (#2971). Run the gate. It fails at **step 1/5, `validate`**:

```
error: records[1] (lang.jsts.framework.nestjs): lane key "route_extraction" (group "Routing")
       declared by subcategory "http_backend" taxonomy but absent from record
```

`validateGroupedCompleteness` (`tools/coverage/validate.go`) is the validate-time mirror of the
same check — its own doc comment says so: *"It is the validate-time mirror of the `backfill`
subcommand: every key the taxonomy promises should have an explicit cell."* With
`completenessGateIsError` true those messages land in `Errors`, not `Warnings`, so `validate` fails
first, on the identical input, one step earlier.

The two guards share a presence set (`rec.AllCapabilities()`), the same gate condition
(`rec.Subcategory != "" && validSubcategory(...)` plus a non-empty taxonomy) and the same
render-order walk. `planBackfill` and `validateGroupedCompleteness` are the same predicate written
twice.

## Why it matters

This is the mutually-masking-guards shape: **neither guard is graded end-to-end by any test that
runs both.** Any input built to exercise the backfill step actually exercises `validate`; a test
asserting "the gate failed" on such an input would pass with the backfill step deleted entirely.
That is precisely the class of ungraded guard that a green suite hides.

In #6866 I did not paper over it. `TestCheckFailsOnBackfillGap` invokes the backfill step in
isolation via `stepNamed(...).Run(env)`, with a positive control that the same step returns nil once
the gap is refilled, plus `TestGateRosterMatchesTheWorkflow` so the step cannot vanish from
`gateSteps()` unnoticed. That grades the step, but it does not make it reachable through the gate,
and the masking is stated in the test comment rather than hidden.

## The decision I think this needs

Does `backfill --check` still earn a slot in the gate? Three options, in the order I would consider
them:

1. **Drop the step.** If `completenessGateIsError` is true and stays true, `validate` strictly
   subsumes it and the step is ~2s of CI buying nothing. The risk is entirely in that constant: if
   it ever flips to false the completeness messages become advisory warnings and the backfill step
   becomes the *only* thing failing on an incomplete record. So dropping it requires either pinning
   the constant or accepting that coupling knowingly.
2. **Keep it and pin the relationship.** A test asserting that the two guards agree on the same
   input — validate errors ⟺ backfill plans a seed — turns the duplication into a checked
   invariant instead of an accident, and makes the redundancy deliberate and visible.
3. **Collapse the implementations.** Have `validateGroupedCompleteness` call `planBackfill` (or a
   shared predicate) so there is one walk rather than two that happen to agree today. This is the
   only option that removes the possibility of them drifting apart.

My preference is (3) then (2). (1) is tempting on cost but the `completenessGateIsError` coupling is
exactly the kind of thing that gets flipped later by someone who has no reason to know the backfill
step was deleted because of it.

## Pointers

- `tools/coverage/validate.go` — `validateGroupedCompleteness`, and the `completenessGateIsError`
  constant that sets its severity.
- `tools/coverage/backfill.go` — `planBackfill` / `cmdBackfill`'s `--check` path.
- `tools/coverage/check.go` — `gateSteps()`, step 2 of 5.
- `tools/coverage/check_test.go` — `TestCheckFailsOnBackfillGap`, which documents the masking and
  the isolation workaround.

Surfaced by the agent working #6866, which measured it rather than assuming — the masking was found
because a whole-gate test for the backfill step kept reporting the validate hint instead.

cajasmota:
## The issue's premise did not survive verification

#6868 states that `planBackfill` and `validateGroupedCompleteness` use "the identical gate, the
identical presence set and the identical key walk", so **no** whole-gate input can attribute a
failure to gate step 2/5. That was confirmed twice — by shared-predicate analysis and by dropping a
grouped cell and watching step 1/5 fire.

Both confirmations used the same input shape, and it is the one shape where the two guards agree.

They did **not** use the same gate:

| | `planBackfill` | `validateGroupedCompleteness` (before) |
|---|---|---|
| reached when | `rec.Subcategory` declares a group taxonomy | only from `validateGroupedRecord`, i.e. only when `rec.IsGrouped()` |

`IsGrouped()` is `len(rec.Groups) > 0`. So a record whose subcategory declares a taxonomy but which
carries **no grouped cells** — the shape `coverage add` produces, and the shape `planBackfill`'s own
doc comment calls out ("a freshly `add`-ed record has an empty flat capabilities map and
`IsGrouped()==false`") — was invisible to `validate` entirely. `validateFlatRecord`'s
"flat capability shape forbidden" error needs `len(rec.Capabilities) > 0`, so it did not catch it
either.

Measured, by adding one such record to the gate fixture and running the whole gate:

```
==> [1/5] Validate schema + cites          <- passed
==> [2/5] Guard against incomplete grouped records (#2971)
error: backfill --check: 50 cell(s) would be seeded
::error::docs/coverage/registry.json has grouped records missing cells the taxonomy declares. ...
```

Step 2/5, attributably, with only its own annotation. The masking was real for grouped records and
total for nothing else. The two guards were a mirror in the doc comment and not in the code — and
the difference was a silent hole in `validate`.

## What I chose: option 3, with option 2's pin

**`missingTaxonomyCells` (`backfill.go`) is now the single implementation of the decision.**

- `planBackfill` applies the `--language` / `--subcategory` filters and calls it per record.
- `validateGroupedCompleteness` calls it and only renders the result as messages. It derives
  nothing.
- Validate's call site moves from `validateGroupedRecord` to `validateRegistry`'s record loop. The
  predicate self-gates on the subcategory, so gating the *call* on `IsGrouped()` is precisely what
  reintroduces the divergence — and that is now a graded mutant, not a comment.

This closes the validate hole as a side effect: the shape that used to reach only step 2/5 now fails
at step 1/5, which is why the masking the issue describes is, after this change, actually total. It
is total **by construction and deliberately**, not by coincidence.

`recordCompletenessMessages` takes the severity as a parameter rather than reading
`completenessGateIsError` directly, so the knob-false world can be run rather than argued about.

**The backfill gate step is kept.**

### What I rejected

**Option 1, delete the step** — rejected, and the `completenessGateIsError` objection is answered by
running it, not by reasoning. I flipped the constant to `false` and ran the whole gate on a grouped
record missing one declared cell:

```
completenessGateIsError=false
==> [1/5] Validate schema + cites
warning: records[1] (lang.jsts.framework.nestjs): lane key "middleware_coverage" (group "Middleware")
         declared by subcategory "http_backend" taxonomy but absent from record
==> [2/5] Guard against incomplete grouped records (#2971)
error: backfill --check: 1 cell(s) would be seeded
::error::docs/coverage/registry.json has grouped records missing cells the taxonomy declares. ...
gate failed at step 2/5
```

Validate emits the message as a **warning** and passes. Step 2/5 is the only thing that fails. So
the answer to "what still catches an incomplete record if the step is deleted and the knob is
flipped" is: nothing. The coupling the issue worried about is real and measured, and the ~2s of CI
is the price of the knob staying safe to flip. `TestCompletenessSeverityKnobDecidesWhoCatchesIt`
drives both values of the severity and asserts the gate step's outcome, so this is pinned rather
than recounted in a comment.

**Option 2 alone, pin the relationship** — rejected as insufficient on its own. A test that the two
agree leaves two walks that can drift between test runs of the invariant, and it would have
codified the `IsGrouped()` divergence as "the invariant", since the two genuinely disagreed. Fixing
the divergence first and *then* pinning is the order that makes the pin true. The pin is included.

## Tests

- `TestBackfillAndValidateAgreeOnCompleteness` — the invariant, over five inputs including the shape
  that used to diverge. It compares the two guards as **sets** (record, key, owning group), not as
  "both non-empty": two guards that both fire on *different* cells is the drift it exists to catch,
  and a non-emptiness assertion cannot see that. `wantMissing` per case stops it passing vacuously
  on a predicate that reports nothing for every input.
- `TestValidateFlagsTaxonomyRecordThatCarriesNoGroups` — the direction the collapse changed, on its
  own, naming a specific declared lane.
- `TestCompletenessSeverityKnobDecidesWhoCatchesIt` — both severities, plus the gate step's outcome
  with a positive control (green after refill).
- `TestCompletenessGateSeverityIsHonoured` — the control that `validateGroupedCompleteness` still
  passes the constant rather than a hardcoded `true`.
- `TestCheckFailsOnBackfillGap`'s comment is corrected: it asserted the masking was total when it
  was not.

Each: exact edit, asserted match count of 1 proving it landed, `go vet` clean, whole-package
`-count=1`. Both mutated files restored by `cp` from shasum-verified backups; no mutant is applied.

| # | Edit | Direction | Verdict |
|---|---|---|---|
| M1 | re-gate validate's call site on `if rec.IsGrouped()` — the exact pre-fix divergence | permissive | **DEAD** — 4 tests, incl. `TestBackfillAndValidateAgreeOnCompleteness` and `TestValidateFlagsTaxonomyRecordThatCarriesNoGroups` |
| M2 | `if _, ok := existing[key]; ok {` → `ok \|\| true` in `missingTaxonomyCells` (nothing ever reported) | permissive | **DEAD** — 9 tests |
| M3 | same line → `ok && false` (presence check removed; every key reported even when present) | restrictive | **DEAD** — 13 tests |
| M4 | validate drops the last cell of the shared result (`missing[:len(missing)-1]`) — both guards still fire, on different sets | permissive | **DEAD** — `TestBackfillAndValidateAgreeOnCompleteness` **only** |

M2 is the option-3 evidence the brief asks for: it edits **`backfill.go`** and kills the
**validate-only** tests. A `validateGroupedCompleteness` that reimplemented the walk instead of
calling the shared one would have survived it.

M4 is what shows the pin distinguishes rather than merely detects. It keeps both guards non-empty
and is caught by nothing else in the suite — a "both fired" assertion would have scored it ALIVE.

## Verification

- `go test -count=1 ./tools/coverage/` — **ok**, exit 0.
- `go vet ./tools/coverage/` clean; `gofmt -l` empty.
- `go run ./tools/coverage check` on a clean tree — **exit 0, gate passed (5/5 steps)**.
  This is safe by construction, not by luck: `backfill --check` was already green on the real
  registry, and `validate` now runs the identical predicate, so widening its scope can add no new
  error there.
- `docs/coverage/registry.json` and every generated page are untouched (`git status`).
- The step list is unchanged — five steps, same names, same order — so
  `TestGateRosterMatchesTheWorkflow` and `TestWorkflowInvokesCheck` hold, both verbatim `::error::`
  hints are byte-identical, and nothing was re-inlined into `coverage-docs.yml`. No "step N/5"
  wording moved.

## One thing that contradicts the brief

The brief's "established fact — do not re-derive it, verify it and move on" is false as stated, and
verifying it is what found that. Two independent confirmations agreed because both used a grouped
record; the divergence lives in the un-grouped one. Worth recording as the "N repros in one env is
ONE repro" shape: the second confirmation varied the method (static vs empirical) but held the input
shape constant, which is the axis that mattered.

Closes #6868

### DIFF
--- a/tools/coverage/backfill.go
+++ b/tools/coverage/backfill.go
@@ -25,78 +25,107 @@ type seedPlan struct {
 }
 
 // planBackfill computes the set of lane cells that are declared by each
-// grouped record's subcategory group taxonomy but absent from the
-// record's cells. It mutates nothing — callers decide whether to seed.
+// record's subcategory group taxonomy but absent from the record's
+// cells. It mutates nothing — callers decide whether to seed.
 //
-// Only GROUPED records whose subcategory carries a group taxonomy are
-// considered. A lane key is "declared" if it appears in the subcategory's
-// group taxonomy; its canonical owning group is groupForCapability (the
-// first group in render order that lists the key), mirroring the
-// auto-placement in cmdUpdate. A key already present anywhere on the
-// record (any group) is never re-seeded — this is the no-clobber
-// guarantee. framework_specific cells are deliberately ignored: the
-// completeness denominator here is the canonical lane set only.
+// Every record whose SUBCATEGORY carries a group taxonomy is considered,
+// whether or not the record is currently in the grouped shape — the
+// comment this replaced said "only GROUPED records", which was never
+// what the code did and is the misreading #6868 was filed on.
+// framework_specific cells are deliberately ignored: the completeness
+// denominator here is the canonical lane set only.
 //
 // langFilter / subFilter, when non-empty, scope the plan to a single
-// language slug / subcategory slug.
+// language slug / subcategory slug. Note that they scope the PLAN, not
+// the predicate: validate's completeness check is unscoped, so a filtered
+// `backfill --check` is deliberately narrower than `validate` and the
+// agreement invariant is stated over the unfiltered plan.
+//
+// The per-record decision lives in missingTaxonomyCells, which is also
+// what validate's completeness check runs — see that function.
 func planBackfill(reg *Registry, langFilter, subFilter string) []seedPlan {
 	var plans []seedPlan
 	for i := range reg.Records {
 		rec := &reg.Records[i]
-		// Gate on whether the record's SUBCATEGORY declares a group
-		// taxonomy, not on the record's current IsGrouped() state. A
-		// freshly `add`-ed record has an empty flat capabilities map
-		// and IsGrouped()==false, but if its subcategory has a group
-		// taxonomy it still needs backfilling into that taxonomy.
-		if rec.Subcategory == "" || !validSubcategory(rec.Category, rec.Subcategory) {
-			continue
-		}
-		groups := groupsForSubcategory(rec.Subcategory)
-		if len(groups) == 0 {
-			continue
-		}
 		if langFilter != "" && rec.Language != langFilter {
 			continue
 		}
 		if subFilter != "" && rec.Subcategory != subFilter {
 			continue
 		}
-		existing := rec.AllCapabilities()
-		// Walk the taxonomy in canonical render order and seed each
-		// declared key exactly once, into its canonical owning group.
-		// A key declared under multiple groups (e.g. db_effect) is owned
-		// by the first group in render order — groupForCapability returns
-		// precisely that group, so seeding stays deterministic.
-		seeded := map[string]struct{}{}
-		for _, g := range groups {
-			for _, key := range g.Keys {
-				if _, ok := existing[key]; ok {
-					continue
-				}
-				if _, done := seeded[key]; done {
-					continue
-				}
-				owner := groupForCapability(rec.Subcategory, key)
-				if owner == "" {
-					owner = uncategorizedGroup
-				}
-				if owner != g.Name {
-					// Defer to the canonical owning group; we'll reach it
-					// later in the render-order walk.
-					continue
-				}
-				seeded[key] = struct{}{}
-				plans = append(plans, seedPlan{
-					RecordID: rec.ID,
-					Language: rec.Language,
-					Group:    owner,
-					Key:      key,
-				})
-			}
-		}
+		plans = append(plans, missingTaxonomyCells(rec)...)
 	}
 	sortSeedPlans(plans)
 	return plans
+}
+
+// missingTaxonomyCells is THE grouped-completeness predicate: the lane
+// cells rec's subcategory group taxonomy declares but rec does not
+// carry. It is the ONLY implementation of that decision in this tool —
+// `backfill` (plan/--dry-run/--check) and `validate`'s
+// validateGroupedCompleteness both call it, and neither re-derives it.
+//
+// That single-implementation property is the point (#6868). The two used
+// to be separate walks described as mirrors of each other, and they were
+// not mirrors: this one gates on whether the record's SUBCATEGORY
+// declares a group taxonomy, while validate's copy only ran for records
+// with rec.IsGrouped() true. A freshly `add`-ed record has an empty flat
+// capabilities map and IsGrouped()==false, so validate's copy skipped
+// the exact record shape backfill exists to fill, and the two guards
+// disagreed on it. TestBackfillAndValidateAgreeOnCompleteness grades
+// that they no longer can.
+//
+// A lane key is "declared" if it appears in the subcategory's group
+// taxonomy; its canonical owning group is groupForCapability (the first
+// group in render order that lists the key), mirroring the auto-placement
+// in cmdUpdate. A key already present anywhere on the record (any group)
+// is never reported — this is the no-clobber guarantee. Order is the
+// taxonomy's render order; callers that need a total order sort.
+func missingTaxonomyCells(rec *Record) []seedPlan {
+	// Gate on whether the record's SUBCATEGORY declares a group
+	// taxonomy, not on the record's current IsGrouped() state.
+	if rec.Subcategory == "" || !validSubcategory(rec.Category, rec.Subcategory) {
+		return nil
+	}
+	groups := groupsForSubcategory(rec.Subcategory)
+	if len(groups) == 0 {
+		return nil
+	}
+	existing := rec.AllCapabilities()
+	// Walk the taxonomy in canonical render order and report each
+	// declared key exactly once, under its canonical owning group.
+	// A key declared under multiple groups (e.g. db_effect) is owned
+	// by the first group in render order — groupForCapability returns
+	// precisely that group, so the result stays deterministic.
+	var out []seedPlan
+	seen := map[string]struct{}{}
+	for _, g := range groups {
+		for _, key := range g.Keys {
+			if _, ok := existing[key]; ok {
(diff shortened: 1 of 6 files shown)


cajasmota:
## Correction: my M7 verdict was wrong — the edit never landed

I reported that adding an early return for empty-capabilities records to `validateGroupedCompleteness` was **ALIVE**, and concluded this PR's motivating shape had no regression test. **The mutant is DEAD.** The author could not reproduce my result, and they are right.

### What actually happened

My substitution silently did nothing:

```sh
perl -0pi -e 's/\Q\tmissing := missingTaxonomyCells(rec)\E/…/'
```

Inside single quotes `\t` is a backslash followed by `t`, and `\Q` quotes it as exactly that — a literal two-character sequence which never matches a real tab. Reproduced in isolation: zero substitutions. So I ran the suite on an **unmutated** tree, got `ok`, and read it as ALIVE.

I asserted the match count *before* the edit and my post-edit `grep` for the inserted line printed nothing — which I did not notice. The standing rule in this repo is **prove the mutant edit landed**, precisely because a silently-unapplied edit reads as ALIVE and argues the opposite conclusion. I wrote that rule into the briefs for this series and then broke it.

The author's diagnosis of my census was also right: I searched for the literal `Capabilities: map[string]Capability{}`, but this PR's `ungroupedRecord()` **omits the field entirely** — a nil map, `len() == 0`, invisible to that grep. Four tests already carried the shape.

### The instinct was right one shape over, and that part stands

The sibling I asked them to check is a real hole, and it was **ALIVE**:

```go
if !rec.IsGrouped() && len(rec.Capabilities) > 0 { return }
```

A flat-shaped record under a taxonomy subcategory had a test for `flat shape forbidden` but **nothing asserted its completeness**. Validate could have stopped reporting it silently — the same hole this PR closes, one shape over.

That is the useful generalisation: *"not grouped"* is **three** values of `rec.Capabilities` — nil, explicitly empty, and non-empty-with-no-groups — and a re-gating predicate readmits them one at a time. The test is now a table over all three, with the flat subcase also asserting that the lane the record *does* carry is not reported, so it cannot pass on a predicate that ignores the presence set.

The explicitly-empty case is now spelled out rather than expressed as a nil map, so the next person's grep census finds what mine missed.

### Net

Two of my three claims here were wrong — the ALIVE verdict and the census — and the third (ask about the sibling shape) found the defect. Recording it this way round because a scorecard that quietly drops its wrong entries is worth less than one that keeps them.

PR #6879, opened 2026-09-05
fix(#6852): html's asset refs anchored on a path nothing declared — arm 3 of twelve, and the first where BOTH depths dangled

cajasmota:
# #6852 arm 3 of twelve — html. The multiplicity case, and the first arm to hit #6861's caller-set guard.

`internal/extractors/html` builds **every** asset `IMPORTS` edge through one constructor,
`buildAssetImportRel`, which stamps `FromID: file.Path`. Nothing this extractor emits is *named* after
the file — every record it emits is named after the **reference** (the `src`/`href` value) — and
`internal/resolve/refs.go` has no path→entity index, so a path-valued `FromID` resolves only onto a
record carrying that exact string as its `Name`. Nothing did, so the raw path reached the graph as the
edge's FROM end. Same defect #6815 fixed in erlang/nim/groovy and #6852 fixed in bicep (#6864) and
terraform (#6871); same fix, `extractor.PrependFileCarrier`.

`html`'s line is deleted from `knownMissingCarrier6847`.

## One anchor, six source sites, five reachable — from source, then measured

`buildAssetImportRel` has six call sites: `<link>` and `` in `visitElement`, `<script>` in
`visitScriptElement`, and `` / `<script>` / `<link>` in `visitSelfClosingTag`. They are **six
producers of one anchor string**, not six anchors — the resolution requirement is one string, so one
carrier serves them all. The ledger row names "script/link refs"; `` anchors identically and is
covered here too.

The self-closing `<script src="..."/>` site is **unreachable through the grammar**, and that is a
measurement, not an omission: tree-sitter treats `<script>` as a raw-text element, so that spelling
opens a `script_element` whose unterminated raw text swallows the rest of the document and the file
extracts to **zero** records. A fixture for it would assert nothing about the carrier. Stated in the
test file rather than left as a silent gap in the fixture set.

## Both path depths dangled — the axis differs from terraform's

hcl already emitted a file-level `SCOPE.Component` named `basename(path)`, so its **root** case
resolved by accident (#6367) and only the nested case could fail. html emits **no file-scoped record
of any kind**, so `index.html` and `src/pages/index.html` both dangled. Both are pinned, per depth per
site, and both were read as RED before the change.

`FileCarrierFor` clause 3 therefore fires for html only when a page references **its own path** —
contrived but reachable, and otherwise clause 3 would be entirely ungraded for this caller, so
`TestHTML_SelfReferencingRefGetsNoSecondCarrier_6852` drives it.

## Multiplicity is what this arm adds

bicep had one anchored edge shape, terraform two. Real html pages reference many assets, so
`TestHTML_OneCarrierPerFileNotPerImport_6852` drives **seven** anchored edges spanning all five
reachable sites, at both depths, and requires exactly **one** carrier.

## Graded in both directions

An unconditional carrier would mint one bare orphan node per `.html` file across a whole repo — a
change no recall-shaped assertion can see. Four forbidden-row controls forbid it, each on a distinct
return path or shape:

| control | what it varies |
|---|---|
| `TestHTML_NoCarrierWithoutAPathAnchoredImport_6852` | a full record set (form, fields, custom element, mustache, Jinja2) with **no** local asset ref |
| `TestHTML_ExternalRefsGetNoCarrier_6852` | the tags are present but every ref is a CDN URL (#506 drops them before any record is built) |
| `TestHTML_EmptyFileGetsNoCarrier_6852` | `Extract`'s empty-content early return |
| `TestHTML_EmailTemplateGetsNoCarrier_6852` | `Extract`'s #506 skip, on a fixture that **would** anchor if extracted |

## Mutation score — permissive direction first

Every mutant: exact edit, asserted match count proving it landed, `go vet` clean, whole-package
`-count=1`, restored by `cp` from a shasum-verified backup.

|---|---|---|---|
| 1 | **unconditional** carrier (`extractor.FileEntity(...)` prepended always) | **DEAD** | the no-anchor, external-refs and self-referencing controls — three independent rows |
| 2 | `lang` argument `"html"` → `""` | **DEAD** | `TestHTML_CarrierShape_6852` at `carrier Language = "", want "html"` (and, incidentally, the pre-existing `TestHTMLExtractor_AllEntitiesHaveQualityScore`) |
| 3 | **placement**: call moved above the CST walk | **DEAD** | resolution + multiplicity + shape — clause 2 can never hold before the anchored records exist |
| 4 | anchor spelling `file.Path` → `assetBasename(file.Path)` | **DEAD** | **nested subtests only** — every root subtest passes, which is the measured proof that the nested depth is load-bearing and a root-only fixture set would have scored this ALIVE |
| 5 | roster entry for `internal/extractors/html` removed | **DEAD** | `TestCarrierCallerSetIsGradedFromSource_6861` |
| 6 | `html` re-added to `knownMissingCarrier6847` | **DEAD** | `TestFileAnchoredCarrier_NoNewOffender_6847` — "no longer reproduces" |
| 7 | a **different** language's ledger line (`astro`) deleted | **DEAD** | same guard, other direction — new offender |

One mutant is deliberately **not** scored because it is equivalent: moving the call between the CST
walk and the Jinja2 append. `PrependFileCarrier` returns the carrier at index 0 and the Jinja2 records
anchor nothing, so the result is byte-identical. Saying so beats reporting an ALIVE verdict that means
nothing.

## The #6861 obligation, and what the roster does not prove

`internal/extractors/html` runs **no** `extractor.TagEntitiesLanguage`, so it is the third
non-tagging caller and the guard fails until the roster names a grading test. Added
`{"internal/extractors/html", []string{"TestHTML_CarrierShape_6852"}}` and the matching `mustScan`
anchor (8 anchors now, floor 6; 6 external callers, floor 5).

That guard's own doc is explicit that it grades the named test's **existence and location, never its
content** — a gutted test with the right name satisfies it. So the burden was taken as mine: the token
was scored **directly** by mutant 2 above, not inferred from a green guard.

Worth noting for the next arm: the pre-existing `TestHTMLExtractor_AllEntitiesHaveQualityScore` also
fails under the empty-token mutant, because it asserts `e.Language == "html"` over *all* entities and
the carrier is now one of them. It is not named in the roster — it grades the token incidentally
rather than by design, and naming an incidental test in the roster would overstate what the entry
records.

## Costs

- **`./cmd/grafel/` is green and that is not coverage of this change**, stated rather than implied:
  `writeSpanFixture6118`'s members are cs/java/ts/js/py/go/rs with **no `.html` file**, so
  `entityTupleKey`'s new `Language` hashing (#6862) cannot reach this. No golden moved, no digest
  moved.
- **The coverage gate cost ZERO this arm**, contrary to bicep (16 lines / 2 citations) and terraform
  (22 lines / 3 citations across 2 records). No record in `docs/coverage/registry.json` cites
  `internal/extractors/html` at all — checked by re-deriving every citation path, not by trusting the
  gate's silence — so the 23-line shift invalidated nothing. `go run ./tools/coverage check` passes
  5/5.

Package set run green at `-count=1`: `./cmd/grafel/`, `./internal/extractor/`, `./internal/extractors/`,
`./internal/extractors/html/`, `./internal/resolve/`, `./internal/quality/`.

Refs #6852

### DIFF
--- a/internal/extractor/carrier_caller_set_6861_test.go
+++ b/internal/extractor/carrier_caller_set_6861_test.go
@@ -143,6 +143,7 @@ var nonTaggingCallers = []rosterEntry{
 		"TestTerraform_CarrierShape_6852",
 		"TestHCLToken_ImportsFromEndResolves_6852",
 	}},
+	{"internal/extractors/html", []string{"TestHTML_CarrierShape_6852"}},
 }
 
 // scannedFile is one file the walk read, kept as a name plus the FULL body the
@@ -196,6 +197,7 @@ func TestCarrierCallerSetIsGradedFromSource_6861(t *testing.T) {
 		{"internal/extractors/groovy/groovy.go", []string{prependFileCarrier}},
 		{"internal/extractors/bicep/extractor.go", []string{prependFileCarrier}},
 		{"internal/extractors/hcl/extractor.go", []string{prependFileCarrier}},
+		{"internal/extractors/html/extractor.go", []string{prependFileCarrier}},
 	}
 
 	nonTest := 0
--- a/internal/extractors/file_anchored_carrier_runtime_6847_test.go
+++ b/internal/extractors/file_anchored_carrier_runtime_6847_test.go
@@ -140,7 +140,6 @@ var knownMissingCarrier6847 = map[string]string{
 	"crystal": "extractor.go:161 `FromID: filePath` on IMPORTS; no carrier emitted",
 	"dart":    "path-anchored IMPORTS; no carrier emitted",
 	"fsharp":  "extractor.go:673 `FromID: filePath` on IMPORTS; no carrier emitted",
-	"html":    "path-anchored IMPORTS (script/link refs); no carrier emitted",
 	"lua":     "lua.go:540 `FromID: file.Path` on IMPORTS; no carrier emitted",
 	"shell":   "shell.go:260 `FromID: file.Path` on IMPORTS (source/.); no carrier emitted",
 	"zig":     "zig.go:272 `FromID: filePath` on IMPORTS; no carrier emitted",
--- a/internal/extractors/html/extractor.go
+++ b/internal/extractors/html/extractor.go
@@ -212,12 +212,26 @@ func (e *Extractor) Extract(ctx context.Context, file extractor.FileInput) ([]ty
 	// drops those before any record is built) or which reference nothing local
 	// at all.
 	//
-	// Placed AFTER both passes, deliberately: the anchored edges are carried by
-	// the script/link/img records the CST walk produces, so before that append
-	// FileCarrierFor's clause 2 can never hold and no carrier would ever be
-	// minted. The two earlier returns — empty content and the #506
-	// email-template skip — return nil without reaching here, and neither
-	// anchors anything, so the helper would be a guaranteed no-op on both.
+	// Placed AFTER BOTH passes, deliberately, and each pass makes it load-bearing
+	// for a DIFFERENT clause:
+	//
+	//   - against the CST walk, for clause 2. The anchored edges are carried by
+	//     the script/link/img records that walk produces, so before its append
+	//     FileCarrierFor's clause 2 can never hold and no carrier would ever be
+	//     minted at all.
+	//   - against the Jinja2 pass, for clause 3. Those records own no
+	//     relationships, so they cannot affect clause 2 — but they ARE NAMED
+	//     (after the directive argument), so they are exactly what clause 3
+	//     scans. Running the helper before them would let a record named after
+	//     the path slip in AFTERWARDS and put two nodes under one id. Today that
+	//     needs a path whose whole value is a Jinja2 keyword ("include"), so it
+	//     is not production-reachable — which is a fact about today's key space,
+	//     not a property of the placement, and is why the call sits after the
+	//     pass rather than between them.
+	//
+	// The two earlier returns — empty content and the #506 email-template skip —
+	// return nil without reaching here, and neither anchors anything, so the
+	// helper would be a guaranteed no-op on both.
 	entities = extractor.PrependFileCarrier(file.Path, "html", entities)
 
 	span.SetAttributes(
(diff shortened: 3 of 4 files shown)


cajasmota:
## Coordinator verification — the cross-PR integration holds

Nine mutants were scored by author and review; I repeated none of them. Mine went at the question neither could ask, because it spans two PRs: **#6861's caller-set guard shipped four days before this arm existed. Does it actually protect a roster entry added later?**

html is the first non-tagging caller added since that guard landed, so this is the first time the mechanism is exercised by a genuinely new entry rather than by the entries it was written against.

I pointed html's roster entry at a real test that lives in **another** package:

```go
-	{"internal/extractors/html", []string{"TestHTML_CarrierShape_6852"}},
+	{"internal/extractors/html", []string{"TestBicep_CarrierShape_6852"}},
```

`go vet` clean, post-edit match count asserted = 1. `go test -count=1 ./internal/extractor/` → **exit 1. DEAD**:

```
carrier_caller_set_6861_test.go:328: nonTaggingCallers says TestBicep_CarrierShape_6852 grades
internal/extractors/html's carrier token, but no _test.go in that package declares
func TestBicep_CarrierShape_6852(
```

Exactly the intended diagnostic, naming both the package and the missing symbol. The location half of the roster is live for new entries, not just for the ones present when the guard was written — which is the part that would have quietly rotted.

**What this does not show, and the file already says so:** the roster grades a named test's **existence and location, never its content**. A gutted `TestHTML_CarrierShape_6852` would satisfy it. That limitation is documented at the roster and was measured when #6861 shipped; this arm carries the burden the right way round by scoring the token directly (M2 fails the named test at `carrier Language = "", want "html"`), rather than treating a green guard as evidence.

## One note on my own method

My first attempt at this mutant **failed to apply** — a `perl -0pi` substitution with double quotes in the pattern died on a syntax error — and the post-edit count gate caught it and stopped. I had added that gate an hour earlier, after making precisely the opposite mistake on #6878: asserting a match count *before* an edit, not after, and reading an unmutated tree's `ok` as an ALIVE verdict.

Recording it because it is the same failure mode twice in one session, caught the second time only because the check was mechanical rather than a line of output to eyeball.

Issue #6326, opened 2026-08-18
the F# extractor emits no EXTENDS or IMPLEMENTS edges, so every F# graph has no inheritance topology at all


cajasmota:
Found while researching the resolution contract for a new line-scanner language. This is shipped behaviour, not a regression.

## What is wrong

**The F# extractor emits no `EXTENDS` and no `IMPLEMENTS` edges. At all.**

Its entire edge vocabulary is `CALLS`, `IMPORTS`, `CONTAINS`, `USES`, `RENDERS`. So every F# codebase we have ever indexed has a graph with **zero inheritance and zero interface-implementation topology** — object hierarchies simply do not exist in it.

Worse than absent: `inherit` and `interface` appear in `fsharpKeywords` (`internal/extractors/fsharp/extractor.go:650`), i.e. they are deliberately *suppressed* as call targets. So the tokens are seen, recognised as significant, and then dropped — rather than being turned into the edges they represent.

## Why it happens

A registration gap, not a technical limit. `internal/extractors/cross/hierarchy/extractor.go:1018-1023` has a `supportedLanguages` set, with a matching dispatch switch at `:1046`. **F# is absent from both.**

## Why this is not a line-scanner limitation

Solidity, Verilog, Crystal and Lua are all regex/line-scanner extractors and all emit `EXTENDS` from their own extractor code — `internal/extractors/solidity/extractor.go`, `verilog/extractor.go`, `crystal/extractor.go`, `lua/oop.go`. There is nothing about not having a CST that prevents this.

F# inheritance syntax is line-anchored and keyword-led:

```fsharp
type Derived() =
    inherit Base()
    interface IDisposable with
        member _.Dispose() = ()
```

`inherit <Type>` and `interface <Type> with` are exactly the shape the existing `letRE`-style scanning already handles well.

## Why it matters now

Two reasons beyond the obvious.

1. **It is silent.** Nothing reports that a language produces no inheritance edges. An F# user asking "what extends this type" gets an empty answer that is indistinguishable from "nothing extends it". Same class as #6321, where 670 `.vb` files produced zero entities while `doctor` reported the group healthy.
2. **It is a live warning for the VB.NET work.** VB.NET is being added as a line-scanner extractor, and `Inherits X` / `Implements I1, I2` are its most important edges. If we register VB.NET the way F# was registered, we ship the same hole. The fix here should establish the pattern the VB.NET extractor follows.

## Fix

[...]

PR #6880, opened 2026-09-05
fix(#6852): fsharp's open-directive IMPORTS anchored on a path nothing declared — arm 4 of twelve, and the second caller where clause 3 fires in production

cajasmota:
# #6852 arm 4 of twelve — fsharp. One anchoring site, N edges, both depths dangling, and the arm where clause 3 meets the #6369 `Subtype` hazard

`internal/extractors/fsharp/extractor.go`'s `buildImportEntities` stamps `FromID: filePath` on every
`open` placeholder's `IMPORTS` edge, and **nothing this extractor emits is named after the file** —
every record is named after a module, namespace, type, operation, DU case or active pattern.
`internal/resolve/refs.go` has no path→entity index, so a path-valued `FromID` resolves if and only
if some emitted node carries that exact string as its `Name`. Nothing did, so the raw path reached
the graph as the edge's FROM end. Same defect #6815 fixed in erlang/nim/groovy and #6852 fixed in
bicep (#6864), terraform (#6871) and html; same fix, `extractor.PrependFileCarrier`.

`fsharp`'s line is deleted from `knownMissingCarrier6847`.

## The four questions the brief asked, answered from source

| question | fsharp's answer |
|---|---|
| how many anchored sites? | **one** — `buildImportEntities` (`extractor.go:673`). It is the only bare `FromID: path` in the package: `collectHierarchyEdges` leaves `FromID` **empty** deliberately (#6295) and `elmish_feliz.go`'s `RENDERS`/`USES` edges do too. |
| do they share one anchor string? | one site, one string — but **N edges**, one per distinct `open`, each on its own record. So the multiplicity question is html's, not bicep's. |
| pre-existing file component? | **none**, of any kind. Unlike hcl's `basename(path)` `SCOPE.Component`. |
| both depths or only nested? | **both**. `Core.fs` and `src/Domain/Core.fs` were each read as RED before the change. |

## Clause 3 fires here, by a different route than hcl's — and it is what handles the `Subtype` hazard

The brief flagged that #6369 / PR #6480's entire fix was *"stamp the right `Subtype`"*, and that
`Subtype` is **not** hashed into `graph.EntityID`. That matters in exactly one place here, and it is
not the import marker:

- **The module/namespace record CAN be named after the file.** `moduleRE` captures a dotted
  `[\w.]+`, so a **root** file `Core.fs` declaring `module Core.fs` emits a `SCOPE.Component` named
  exactly the path — while still anchoring an `IMPORTS` on it. `graph.EntityID` hashes
  `(repo, kind, name, sourceFile)` and not `Subtype`, so a carrier there would land a **second**
  `SCOPE.Component` under the module record's id and make the resolver's rewrite target ambiguous.
  `FileCarrierFor` **clause 3**'s `records[i].Name == path` is what rejects it — clause 1 is the
  empty-path guard — and **not** anything `Subtype`-aware. Checked, not assumed:
  `TestFSharp_ModuleNamedLikeThePathGetsNoSecondCarrier_6852` asserts the one path-named record is
  the **module** record, and its nested subtest is the contrast that stops the whole test passing on
  a carrier that is never emitted anywhere.
- **The import marker CANNOT be.** Its `Name` is `importDisplayName(mod)` — the last dot-segment of
  the `open` target — while every fsharp path carries a `.fs`/`.fsi`/`.fsx`/`.fsproj` dot, so the
  marker's name is at most the extension. Driven directly rather than argued:
  `TestFSharp_ImportMarkerIsNeverThePathNamedRecord_6852` puts `open Core.fs` in a file at `Core.fs`.
  **The load-bearing fact is the classifier's, not this extractor's**, and the test now says where it
  lives: `importDisplayName` returns `mod` *unchanged* when it has no dot, so the argument reduces to
  "every path routed to fsharp contains a dot" — true because
  `internal/classifier/classifier.go:428-431` is what maps `.fs`/`.fsi`/`.fsx`/`.fsproj` to this
  extractor. A dotless extension mapped to fsharp later breaks the premise **there**, not here.

`file_carrier.go`'s clause-3 note records the MECHANISM and deliberately carries **no ordinal**.
It first said fsharp was "a second such caller"; the html arm landed on main mid-flight with its own
clause-3 case (`TestHTML_SelfReferencingRefGetsNoSecondCarrier_6852`), which made that ordinal false
on the rebased tree. Given this is the file whose self-labelled counts have already gone stale twice
(#6861), the fix is to stop writing a count there at all rather than to bump it to three.

## Graded in both directions

An unconditional carrier mints one bare orphan node per `.fs` file across a whole repo — a change no
recall-shaped assertion can see. Two forbidden-row controls forbid it, on distinct return paths:

| control | what it varies |
|---|---|
| `TestFSharp_NoCarrierWithoutAnOpen_6852` | a full record set (module, record type, DU, operations) with **no** `open`, at **both** depths |
| `TestFSharp_EmptyFileGetsNoCarrier_6852` | `Extract`'s `len(file.Content) == 0` early return |

`TestFSharp_OneCarrierPerFileNotPerOpen_6852` drives **four** `open`s (including F# 5's
`open type`) at both depths and requires exactly **one** carrier.
`TestFSharp_SignatureFileImportsFromEndResolves_6852` varies the file form: a `.fsi` routes to this
same extractor and emits no hierarchy edges (#6326) but still anchors, so a carrier wired to some
`.fs`-only branch would leave every signature file dangling.

## Mutation score — permissive direction first

Every mutant: exact edit, **match count asserted after the edit as a gate** (`assert count==1` on the
replacement and `==0` on the original, inside the same script that wrote the file — not read off
stdout), `go vet` clean, whole-package `-count=1`, restored by `cp` from a shasum-verified backup.
The final worktree diff is 5 files and no mutant.

|---|---|---|---|
| 1 | **unconditional** carrier (`extractor.FileEntity(...)` prepended always) | **DEAD** | the no-`open` control at both depths **and** the clause-3 collision (`got 2` records named the path) — two independent shapes |
| 2 | **placement**: call hoisted above the CST/regex scan | **DEAD** | resolution, multiplicity and shape — clause 2 can never hold before the import records exist |
| 3 | `lang` `"fsharp"` → `""` | **DEAD** *(see below)* | `TestFSharp_CarrierShape_6852`, on `Properties["language"]` |
| 4 | `lang` `"fsharp"` → `"fs"` | **DEAD** | `TestFSharp_CarrierShape_6852` at `carrier Language = "fs", want "fsharp"` |
| 5 | anchor spelling `filePath` → `basename(filePath)` | **DEAD** | **nested subtests only** — every root subtest passes, the measured proof that the nested depth is load-bearing and a root-only fixture set would have scored this ALIVE |
| 6 | `fsharp` re-added to `knownMissingCarrier6847` | **DEAD** | `TestFileAnchoredCarrier_NoNewOffender_6847` — "no longer reproduces" |
| 7 | a **different** language's ledger line (`astro`) deleted | **DEAD** | same guard, other direction — new offender, naming two astro corpus files |
| 8 | `internal/extractors/fsharp` added to `nonTaggingCallers` | **DEAD** | `TestCarrierCallerSetIsGradedFromSource_6861` — "is listed as a NON-TAGGING caller but its package does call TagEntitiesLanguage" |

**Mutant 3 was ALIVE when first scored, and the honest classification is neither of the brief's two
options — it was *distinguishable and ungraded*, so it was graded rather than recorded.** fsharp
runs `extractor.TagEntitiesLanguage`, which fills an empty `Language` with `"fsharp"`, so `Language`
alone cannot tell `""` from `"fsharp"` and the whole package was green. What *does* tell them apart
is provenance: `TagEntitiesLanguage` only touches a record whose `Language` is empty, and when it
does it **also** stamps `Properties["language"]`. Every other fsharp record sets `Language`
explicitly and therefore carries no such key, so under the mutant the carrier is the one record in
the extraction whose provenance differs from its siblings' — proto's #6356 trap, which is the reason
`file_carrier.go` takes the token as a parameter at all. `TestFSharp_CarrierShape_6852` now asserts
that absence and the mutant dies at
`carrier carries Properties["language"]="fsharp"`.

Not scored, and stated rather than reported as a verdict: **removing fsharp's `mustScan` anchor** from
`carrier_caller_set_6861_test.go`. That list is documented in its own file as redundancy — every walk
mutant fires on the first anchor, `minAnchors` is 6 and the list holds **9** after this arm — so the
mutant would be ALIVE **by design**, not by a gap. The anchor is added anyway, as bicep's, hcl's and
html's arms did, because it names a file this guard makes a claim about.

## #6861: fsharp is a TAGGING caller and owes no roster entry

Determined from source (`Extract` calls `extractor.TagEntitiesLanguage(out, "fsharp")`) and then
confirmed **from the guard**, which reads the caller set out of the tree on every run: adding a
`nonTaggingCallers` entry for fsharp is **DEAD** with the drift message above. So no entry, and the
roster keeps exactly the three entries main already has (bicep, hcl, html — all non-tagging).
External carrier call sites: **7** on this tree, re-derived by raising `minExternalCallers` to 999
and reading the number the guard itself printed, not by grepping and not by adding 1 to a
pre-rebase figure. `minExternalCallers` is 5, so the floor still holds.

The brief's warning was taken seriously anyway — that guard grades a named test's *existence and
location, never its content* — which is why the token is scored directly (mutants 3 and 4) rather
than treated as covered because a guard is green.

## Costs, measured

- **`./cmd/grafel/` is green and that is NOT coverage of this change.** `writeSpanFixture6118`'s
  members are cs/java/ts/js/py/go/rs — no `.fs` file — and `entityTupleKey` does not hash `Language`
  (#6862). No golden moved and no digest moved. Run because it is mandatory, not credited.
- **The coverage gate cost ZERO — the html outcome, not bicep's or terraform's.**
  `docs/coverage/registry.json` cites `internal/extractors/fsharp/*.go` as **files only**: no fsharp
  record carries a symbol-anchored `file.go:NNN` citation (bicep had two, terraform three), so the
  17-line shift invalidated nothing. `go run ./tools/coverage check` passes **5/5** and
  `docs/coverage/` is untouched.

## Package set

`./cmd/grafel/`, `./internal/extractors/fsharp/`, `./internal/extractors/`, `./internal/extractor/`,
`./internal/resolve/`, `./internal/quality/` — all `ok` at `-count=1`. `gofmt -l` clean over
`internal/ cmd/ tools/`.

## Nothing contradicting the brief

The brief's fsharp ledger row, the three prior arms' mechanisms, and `FileCarrierFor`'s clause order
all held as written. The only place the brief's framing did not fit was mutant 3's verdict, addressed
above.

Refs #6852

---

## Rebased onto the html arm (`361fd1db8`), and every shared counter re-derived from the rebased tree

The html arm (#6879) merged to main while this was in flight. Merge-base `87e4cedad`; both arms touch
the same two shared files, and both conflicted textually — which was the good outcome, because the
dangerous version of this is the change that rebases *clean* and silently drops half of itself.

| file | resolution |
|---|---|
| `internal/extractor/carrier_caller_set_6861_test.go` | **both** `mustScan` anchors kept — html's and fsharp's. html's `nonTaggingCallers` entry was never in conflict and is untouched. |
| `internal/extractors/file_anchored_carrier_runtime_6847_test.go` | **both** ledger deletions kept — neither `"html"` nor `"fsharp"` remains. |
| `internal/extractor/file_carrier.go` | no conflict; html's arm does not touch this file. |

**Every shared counter was re-read from disk after the rebase rather than carried forward**, and
three of the four had moved:

| counter | pre-rebase (what this body first said) | re-derived from the rebased tree |
|---|---|---|
| external carrier call sites | 5 → 6 | **7** (read off the guard by probing `minExternalCallers = 999`) |
| `mustScan` anchors | "list now 8" | **9** (`minAnchors` = 6, unchanged, still holds) |
| `len(knownMissingCarrier6847)` | 10 → 9 | **8** — two deletions since the merge-base (`astro clojure cobol crystal dart lua shell zig`) |
| `nonTaggingCallers` | bicep, hcl | bicep, hcl, **html** — fsharp is still absent, correctly, because it tags |

**Mutants re-scored on the rebased tree**, since a rebase invalidates a verdict as readily as a code
edit does — each re-applied with the match count asserted after the edit and restored from a
shasum-verified backup of the *rebased* file:

| mutant | verdict on the rebased tree |
|---|---|
| unconditional carrier | **DEAD** — same two independent shapes (no-`open` at both depths; the clause-3 collision) |
| `fsharp` re-added to `knownMissingCarrier6847` | **DEAD** — "no longer reproduces" |
| **cross-check:** `html` re-added to `knownMissingCarrier6847` | **DEAD** — "no longer reproduces" |

That last row is not part of this arm's grade; it exists because "html's line is absent from the
ledger" is a statement about *text*, and what actually has to survive the rebase is html's *carrier*.
Re-adding html's line and watching the runtime guard reject it is the positive proof that html's fix
is live in this tree, not merely that its deletion was preserved.

Re-run on the rebased tree at `-count=1`: `./cmd/grafel/`, `./internal/extractors/fsharp/`,
`./internal/extractors/`, `./internal/extractor/`, `./internal/resolve/`, `./internal/quality/` — all
`ok`. `gofmt -l` clean over `internal/ cmd/ tools/`; `go run ./tools/coverage check` still 5/5 with
`docs/coverage/` untouched. Final worktree diff is 5 files and no mutant.

---

## Review: APPROVE, three prose/assertion fixes folded into the same push

The independent review verified both claims the brief singled out — that clause 3 really is what
fires (the `Name == path` test returns *inside* the loop while `!anchored` is checked *after* it, so
clause 2 cannot reject first, and deleting clause 3 kills only the root subtest, which is possible
only if anchoring was already established), and that the `lang → ""` resolution is real rather than
vacuous (a 19-record / 13-combo probe found `Properties["language"]` absent on every fsharp record).
13 mutants scored, 12 DEAD; the one ALIVE is the `mustScan` anchor deletion, correctly by design.

Three fixes landed in the same commit rather than a follow-up, so this costs one CI cycle:

**1. Clause misnumbering — mine and the brief's.** `file_carrier.go`'s numbered list makes
`records[i].Name == path` **clause 3**; clause 1 is the `path == ""` guard. The test file's header
said "clause 1", and `TestFSharp_ModuleNamedLikeThePathGetsNoSecondCarrier_6852`'s docstring opened
with "drives CLAUSE 3" and then credited "clause 1's `Name == path`" — self-contradictory inside one
comment. Fixed in the test file, in the commit body, and above in this PR body. The comment added
*inside* `file_carrier.go` was correctly placed and is unchanged.

**2. The mutant-killing assertion rested on an unasserted premise.** It checks
`Properties["language"]` is absent **on the carrier**; that no *sibling* carries the key was asserted
nowhere — and that is precisely what makes the check distinguish `""` from `"fsharp"`. If a future
fsharp record shipped without an explicit `Language`, `TagEntitiesLanguage` would fill it, stamp the
key, and the check would go on passing while grading nothing, with the empty-token mutant quietly
back to ALIVE. `TestFSharp_CarrierShape_6852` now loops over every record and forbids the key.

**Pinned, not just added** — the new loop was scored in its own direction rather than trusted to be
non-vacuous. Deleting `Language: "fsharp"` from `buildImportEntities`' record (the exact drift it
guards against) makes it fire: `record kind="SCOPE.Component" subtype="import" name="Text" carries
Properties["language"]="fsharp"`. And the empty-token mutant was re-scored after the rewrite —
still **DEAD**, now on both assertions.

**3. The marker-collision argument belongs to the classifier.** Stated in the test file and above:
`importDisplayName` returns `mod` unchanged when it has no dot, so "the marker can never be named the
path" is not a fact about this extractor at all — it reduces to "every path routed to fsharp contains
a dot", which `internal/classifier/classifier.go:428-431` is what makes true.

The review's item 2 (dropping the "second such caller" ordinal from `file_carrier.go`) had already
been fixed during the rebase — html's arm landed its own clause-3 case, so the ordinal was false by
the time it reached main. The wording is now the review's: *"another such caller, reached by a
different route"*, with an explicit note that no roster belongs in that file (#6861).

`carrier_caller_set_6861_test.go:174-175`'s stale comment numbers are **deliberately untouched** —
the coordinator is filing them separately. Both are floors, so neither can misfire.

Re-run after the fixes at `-count=1`: `./internal/extractors/fsharp/` and `./internal/extractor/`
both `ok`; `gofmt -l` clean over `internal/ cmd/ tools/`; `go vet` clean. Working tree clean, no
mutant or probe left applied.

### DIFF
--- a/internal/extractor/carrier_caller_set_6861_test.go
+++ b/internal/extractor/carrier_caller_set_6861_test.go
@@ -198,6 +198,7 @@ func TestCarrierCallerSetIsGradedFromSource_6861(t *testing.T) {
 		{"internal/extractors/bicep/extractor.go", []string{prependFileCarrier}},
 		{"internal/extractors/hcl/extractor.go", []string{prependFileCarrier}},
 		{"internal/extractors/html/extractor.go", []string{prependFileCarrier}},
+		{"internal/extractors/fsharp/extractor.go", []string{prependFileCarrier}},
 	}
 
 	nonTest := 0
--- a/internal/extractor/file_carrier.go
+++ b/internal/extractor/file_carrier.go
@@ -63,6 +63,19 @@ import "github.com/cajasmota/grafel/internal/types"
 //     property of every root .tf. Pinned end-to-end by
 //     TestTerraform_RootPathGetsNoSecondCarrier_6852, with the block-less case
 //     pinned separately by TestTerraform_EmptyFileGetsNoCarrier_6852.
+//     fsharp (#6852) is another such caller, reached by a different route.
+//     (No ordinal, and no roster of which callers those are: this paragraph is
+//     inside the file whose doctrine below says the caller set is deliberately
+//     not written down here, having been stated as measured fact three times
+//     and falsified twice — #6861.) The route: nothing fsharp emits is named
+//     after the file as a rule, but moduleRE captures
+//     a dotted `[\w.]+`, so a ROOT file Core.fs declaring `module Core.fs`
+//     emits a SCOPE.Component named exactly the path — and since graph.EntityID
+//     does not hash Subtype, a carrier there would land a second node under the
+//     module record's id (the #6369/#6480 hazard). Pinned by
+//     TestFSharp_ModuleNamedLikeThePathGetsNoSecondCarrier_6852, whose nested
+//     subtest is the contrast that stops it passing on a carrier that is never
+//     emitted at all.
 //
 // Clause 3 is checked for EVERY record, before the loop may short-circuit on
 // clause 2 being satisfied — deliberately, and the order is load-bearing.
--- a/internal/extractors/file_anchored_carrier_runtime_6847_test.go
+++ b/internal/extractors/file_anchored_carrier_runtime_6847_test.go
@@ -139,7 +139,6 @@ var knownMissingCarrier6847 = map[string]string{
 	"cobol":   "path-anchored IMPORTS (COPY members); no carrier emitted",
 	"crystal": "extractor.go:161 `FromID: filePath` on IMPORTS; no carrier emitted",
 	"dart":    "path-anchored IMPORTS; no carrier emitted",
-	"fsharp":  "extractor.go:673 `FromID: filePath` on IMPORTS; no carrier emitted",
 	"lua":     "lua.go:540 `FromID: file.Path` on IMPORTS; no carrier emitted",
 	"shell":   "shell.go:260 `FromID: file.Path` on IMPORTS (source/.); no carrier emitted",
 	"zig":     "zig.go:272 `FromID: filePath` on IMPORTS; no carrier emitted",
--- a/internal/extractors/fsharp/extractor.go
+++ b/internal/extractors/fsharp/extractor.go
@@ -519,7 +519,22 @@ func extractFSharp(src, filePath string) []types.EntityRecord {
 	// + emits RENDERS, stamps Cmd dispatch USES edges).
 	applyElmishFeliz(src, filePath, imports, entities)
 
-	return entities
+	// #6852: buildImportEntities anchors each `open` placeholder's IMPORTS on
+	// filePath, and nothing above is NAMED after the file — every record is
+	// named after a module, namespace, type, operation, DU case or active
+	// pattern. resolve/refs.go has no path→entity index, so that FROM end
+	// resolved to nothing and the raw path reached the graph. The carrier is
+	// CONDITIONAL (#6815, #6518): an .fs file with no `open` anchors nothing,
+	// and an unconditional carrier would mint one bare orphan node per source
+	// file across a whole repo — a change no recall-shaped assertion can see.
+	//
+	// Placement is load-bearing: FileCarrierFor clause 2 asks whether some
+	// record already anchors on the path, so the call must come AFTER the
+	// import placeholders exist. Clause 3 matters here too — a root file
+	// `Core.fs` declaring `module Core.fs` already emits a record named
+	// exactly the path, and graph.EntityID does not hash Subtype, so a second
+	// SCOPE.Component would land under the module record's id (#6369/#6480).
+	return extractor.PrependFileCarrier(filePath, "fsharp", entities)
 }
 
 // classifyTypeSubtype determines the F# type subtype from the declaration context.
--- a/internal/extractors/fsharp/file_carrier_6852_test.go
+++ b/internal/extractors/fsharp/file_carrier_6852_test.go
@@ -1,0 +1,444 @@
+package fsharp_test
+
+// file_carrier_6852_test.go — #6852, fsharp arm.
+//
+// buildImportEntities (extractor.go) stamps `FromID: filePath` on the IMPORTS
+// edge of every `open` placeholder, and NOTHING this extractor emits is named
+// after the file: every record is named after a module, namespace, type,
+// operation, DU case or active pattern. internal/resolve/refs.go has no
+// path→entity index, so a path-valued FromID resolves if and only if some
+// emitted node carries that exact string as its Name — nothing did, so the raw
+// path reached the graph as the edge's FROM end. Same defect #6815 fixed in
+// erlang/nim/groovy and #6852 fixed in bicep (#6864), terraform (#6871) and
+// html; same fix, extractor.PrependFileCarrier.
+//
+// ONE ANCHORING SITE, MANY EDGES. buildImportEntities is the only producer of a
+// bare `FromID: path` in the package (collectHierarchyEdges deliberately leaves
+// FromID EMPTY — #6295 — and elmish_feliz.go's RENDERS/USES edges leave it
+// empty too). But it emits ONE record per distinct `open`, each carrying its
+// own path-anchored IMPORTS, so a real F# file anchors N edges and must still
+// gain exactly ONE carrier.
+//
+// BOTH DEPTHS DANGLED, unlike terraform. hcl already emitted a file component
+// named BASENAME(path), so its ROOT case resolved by the accident #6367
+// documents and only nested files dangled. fsharp emits no file-scoped record
+// of any kind, so `Core.fs` and `src/Domain/Core.fs` both dangled — the same
+// shape as html.
+//
+// THE Subtype HAZARD (#6369 / PR #6480), stated because this is the language
+// where it lives. graph.EntityID hashes (repo, kind, name, sourceFile) and NOT
+// Subtype, so two SCOPE.Component records with the same Name and SourceFile
+// collide on one id however their Subtype differs. Two record shapes could in
+// principle be named after the file:
+//
+//   - the IMPORT MARKER (Subtype "import"). It CANNOT be: its Name is
+//     importDisplayName(mod), the last dot-segment of an `open` target, while
+//     every fsharp path ends in `.fs`/`.fsi`/`.fsx`/`.fsproj` and therefore
+//     contains a dot — so the marker's name is at most the extension, never the
+//     path. WHERE THAT LOAD-BEARING FACT LIVES is worth stating, because it is
+//     not here: importDisplayName returns mod UNCHANGED when it has no dot, so
+//     this is not an argument about the extractor at all. It reduces to "every
+//     path routed to fsharp contains a dot", which holds because
+//     internal/classifier/classifier.go:428-431 is what maps
+//     .fs/.fsi/.fsx/.fsproj to this extractor. A dotless extension mapped to
+//     fsharp later would break the premise there, not in this package.
+//     Pinned by TestFSharp_ImportMarkerIsNeverThePathNamedRecord_6852.
+//   - the MODULE / NAMESPACE record (Subtype "module"/"namespace"). It CAN be:
+//     moduleRE captures `[\w.]+`, so a root file `Core.fs` declaring
+//     `module Core.fs` emits a record named exactly the path. That is
+//     FileCarrierFor CLAUSE 3 (`records[i].Name == path`; clause 1 is the
+//     empty-path guard) — NOT anything Subtype-aware — and it is what stops a
+//     second node landing under the one id. Pinned by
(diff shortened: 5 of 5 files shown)


cajasmota:
## Coordinator verification — carrier position, and a note on what this arm did unprompted

Thirteen mutants were scored by the review and eight by the author; I repeated none. Mine went at an axis that mattered in an earlier arm of this series and was not covered here: **where in the slice the carrier lands.**

bicep's arm scored a `PrependFileCarrier` → append-at-end mutant (its M7) and it was DEAD on the `#577` index-0 row. That row is a property of the *shared helper's contract*, not of any one language, so it needed re-asking for a caller whose records are built differently.

```go
-	return extractor.PrependFileCarrier(filePath, "fsharp", entities)
+	if c, ok := extractor.FileCarrierFor(filePath, "fsharp", entities); ok {
+		return append(entities, c)
+	}
+	return entities
```

Post-edit match count asserted as a gate, `go vet` clean. `go test -count=1 ./internal/extractors/fsharp/` → **exit 1. DEAD**, on `TestFSharp_CarrierShape_6852`.

So the prepend is graded here too, and it is graded by the shape test rather than incidentally by a resolution test — which is the right place for it, since resolution does not care about ordering and would have passed either way.

## Two things this arm did that are worth recording

**It scored its own new assertion.** Asked to pin the sibling premise behind the `Properties["language"]` check, it did not simply add the loop — it noted that *a loop asserting an absence over a set that never contains the thing is a no-op that reads like a guard*, and drove it: deleting `Language: "fsharp"` from `buildImportEntities` makes it fire with the offending record named. That is the vacuity lesson applied to its own fix, unprompted, and it is the difference between adding a guard and adding a graded guard.

**It widened a premise rather than restating it.** The marker-collision argument reduces to "every path routed to fsharp contains a dot", which is the **classifier's** fact, not the extractor's. Asked to say where that fact lives, it also found `classifier.go:431` routes `.fsproj` — so the three-extension phrasing understated the population the premise depends on. Four extensions, all dotted, premise intact.

## One correction to me

I told this agent the remote was still at `44c22d11d`. It was already at `24da98bbc` — the rebase push had landed between my two checks and I quoted the earlier one. The worktree check that found the outstanding clause misnumbering at `:42`/`:302` was accurate; the branch-state claim beside it was stale.

Issue #577, opened 2026-05-18
[P0] Python/Go/Java/etc extractors emit file-path FromIDs for IMPORTS edges (same pattern as #570)


cajasmota:
**Surfaced from PR #575 investigation.** The file-path FromID anti-pattern in JS extractor (fixed in #570) exists in other extractors too. Python/Go/Java/Ruby/PHP IMPORTS edges currently land in `DispositionDynamic` via `looksLikeSourceFilePath` (no bug-rate impact) but their cross-repo import links remain invisible.

## Impact
client-fixture-a (Django) and client-fixture-c (RN+Expo) currently produce 0 cross-repo import-channel links between Python ↔ JS even though there are no shared Python/JS imports — the bigger gap is cross-repo Python/Python (when there ARE shared dependencies between two Django apps).

For corpus repos like django-realworld + flask-realworld in the same group: both import `pytest`, `requests`, `json`, `datetime` — these should produce cross-repo import-channel links between corpus repos when grouped.

## Fix
Apply the same pattern as #570 to each per-language extractor:
- Python: `internal/extractors/python/extractor.go` — emit file-level SCOPE.Component (subtype=\"file\") per .py source
- Go: `internal/extractors/go/extractor.go`
- Java: `internal/extractors/java/extractor.go`
- Ruby, PHP, Scala, Kotlin, Swift, etc.

Each language's IMPORTS edges then use file-entity hex ID as FromID (after resolver rewrite).

## Acceptance
- Cross-repo import links available when grouping multi-repo monorepos in any language
- New test corpus pair (2 Python repos sharing django) → produces cross-repo import links
- No bug-rate regression on per-language corpora >0.5pp (or if it does occur, follow same #575 precedent: net signal quality improvement justifies)

PR #6882, opened 2026-09-05
fix(#6852): shell's source/. IMPORTS anchored on a path only sometimes declared — arm 5 of twelve, and the first three-way depth split

cajasmota:
internal/extractors/shell/shell.go's makeImportStub stamps `FromID: file.Path`
on the IMPORTS edge of every `source <path>` / `. <path>` stub.
internal/resolve/refs.go has no path→entity index, so a path-valued FromID
resolves iff some emitted node carries that exact string as its Name. Same
defect #6815 fixed in erlang/nim/groovy and #6852 fixed in bicep (#6864),
terraform (#6871), html (#6879) and fsharp (#6880); same fix,
extractor.PrependFileCarrier, conditional.

shell's line is deleted from knownMissingCarrier6847, both directions scored.

ONE ANCHORING SITE, N EDGES. makeImportStub is the only bare `FromID: path` in
the whole package — buildScriptComponent's CONTAINS and extractCallRelationships'
CALLS both leave FromID EMPTY, and extractRegex (the TSTree == nil fallback)
emits no relationships at all. One stub is emitted per directive, so a script
with four of them anchors four edges and must still gain exactly ONE carrier
(TestShell_OneCarrierPerFileNotPerSource_6852, both depths).

THE DEPTH SPLIT IS SHELL'S OWN SHAPE — neither terraform's nor html's, and this
is the arm's main result. buildScriptComponent DOES emit a file-level
SCOPE.Component named BASENAME(file.Path), so at a ROOT path that name already
IS the path and the edge resolved by the #6367 accident, exactly as hcl's did.
But it is emitted ONLY when the file declares a function (`len(fnEntities) > 0`).
So shell dangled at:

  - every NESTED path, with or without functions; AND
  - the ROOT path of a script with NO function definitions, where no file-level
    component is emitted at all.

A script that only sources libraries and runs top-level commands is the ordinary
shell script, not a corner, so the root hole is real rather than notional. Both
holes were read as RED before the change; the surviving accident (root +
functions) was read as GREEN before the change and is now pinned as a clause-3
REJECTION rather than left to chance:

  TestShell_SourceImportsFromEndResolves_6852           nested FAIL, root FAIL
  TestShell_ScriptWithFunctionsImportsFromEndResolves   nested FAIL, root PASS

CONDITIONAL, NOT UNCONDITIONAL. Most shell scripts source nothing, and an
unconditional carrier would mint one bare orphan node per .sh/.bash/.zsh/.ksh
file across a whole repo — a change no recall-shaped assertion can see. Three
forbidden-row controls forbid it: a full record set (script component, two
functions, a CALLS edge) with no `source` at both depths; Extract's
empty-content early return; and the TSTree == nil regex fallback. The
unconditional mutant is DEAD on the first of them at both depths.

CLAUSE 3 IS REACHED BY TWO ROUTES HERE, AND ONE IS NEW TO THIS ISSUE.
FileCarrierFor clause 3 (`records[i].Name == path`) fires for shell when the
SCRIPT COMPONENT is named the path — hcl's route, root depth only — and, newly,
when an IMPORT STUB is: makeImportStub names the stub after the SOURCED path
VERBATIM, so a script that sources itself by the spelling of its own path is a
path-named record at ANY depth. fsharp's import marker provably cannot be the
path; shell's can. graph.EntityID does not hash Subtype (#6369/#6480), so a
carrier at either would land a second SCOPE.Component under one id. Pinned by
TestShell_ScriptComponentNamedLikeThePathGetsNoSecondCarrier_6852 (whose nested
subtest is the contrast that stops it passing on a carrier never emitted) and
TestShell_SelfSourcingImportStubGetsNoSecondCarrier_6852 (driven at both
depths, over a function-less fixture so the script component cannot be what
satisfies it).

PLACEMENT IS LOAD-BEARING AT BOTH ENDS. Above pass 1 the import stubs do not
exist yet and clause 2 rejects — carrier never minted (DEAD, 5 tests). Above
pass 4 the script component does not exist yet and clause 3 cannot see it — a
second SCOPE.Component lands under its id for any ROOT script that declares a
function AND has a `source`/`.` directive (DEAD, killed by exactly the clause-3
test written for it and by nothing else).

WHAT THAT SECOND ONE IS AND IS NOT, stated precisely because "reachable" is
easy to over-read. It has ZERO instances in THIS repo today: the only
root-level `.sh` here is `install.sh`, which declares functions but sources
nothing. So it is **producible by any real repo**, not observed in this one.
That is still meaningfully different from html's Jinja2 case, which needed a
file PATH whose whole value equals a template keyword: this one needs only an
ordinary script shape, not a coincidence of spelling.

shell CALLS extractor.TagEntitiesLanguage, so it owes NO nonTaggingCallers
entry, and the guard's negative was scored rather than assumed: adding one is
DEAD with the exact intended diagnostic ("is listed as a NON-TAGGING caller but
its package does call TagEntitiesLanguage"). THE #6861 INTERACTION IN THIS ARM
IS THAT NEGATIVE — shell correctly staying OUT of the roster — and not a roster
entry, because there is none to add.

The `mustScan` anchor line added alongside fsharp's is NON-GRADING, said plainly
so this diff is not read as having added a graded roster entry. Scored: removing
it leaves `internal/extractor` green — correctly ALIVE-and-equivalent, since
that file's own prose states the anchors are deliberate redundancy (every walk
mutant fires on the first anchor, `file_carrier.go`, so any six of the seven
could go). It is added because each anchor names a file the guard makes a claim
about, not because it grades anything.

The lang argument is graded anyway. Because shell tags, `""` cannot be caught on
Language alone; it IS caught on provenance — TagEntitiesLanguage only touches a
record whose Language is empty and stamps Properties["language"] when it does,
while every other shell record sets Language explicitly and carries no such key.
TestShell_CarrierShape_6852 asserts that absence, so both the empty token (DEAD,
"carrier carries Properties[\"language\"]=\"shell\"") and a wrong one (DEAD,
`Language = "bash", want "shell"`) die. THE ABSENCE ASSERTION IS SCORED IN ITS
OWN DIRECTION rather than left as a no-op that reads like a guard: deleting
`Language: "shell"` from makeImportStub makes the sibling-premise loop fire,
naming the offending record.

MUTATION SCORE: 11 applied, 10 DEAD, 1 ALIVE — the added `mustScan` anchor's
removal, classified *equivalent under the current suite* (that guard's anchors
are stated redundancy; only the floor on the list is graded), and scored rather
than relayed. Every mutant asserted its match count AFTER the edit as a
hard gate (an insertion mutant gates on a marker absent before the edit and
present exactly once after), `go vet` clean, whole-package `-count=1`, restored
by `cp` from a shasum-verified backup with the sha re-checked.

./cmd/grafel/ is green and that is NOT coverage of this change, stated rather
than implied: writeSpanFixture6118's members are cs/java/ts/js/py/go/rs with no
.sh file, and entityTupleKey hashes Language only as of #6862, which still does
not reach a language absent from the fixture. No golden moved and no digest
moved.

Coverage gate cost: ZERO. docs/coverage/registry.json carries no citation of
internal/extractors/shell at all — not a file path, not a symbol-anchored
`file.go:NNN` — so the 20-line shift invalidated nothing.
`go run ./tools/coverage check` passes 5/5 and regenerates docs byte-identical.

COST AND COUNT ARE DIFFERENT CLAIMS, and the earlier arms are easy to conflate.
fsharp's gate cost was zero too, but not for shell's reason: fsharp HAS nine
citations of its extractor.go, which simply carry no line numbers, so nothing
could go stale. shell has no citation at all. Only bicep (two) and terraform
(three) had line-anchored citations to re-derive. Quote the cost, not the count.

Refs #6852

---

## Axes varied vs held constant (per fixture)

| test | VARIED | HELD CONSTANT |
|---|---|---|
| `TestShell_SourceImportsFromEndResolves_6852` | path DEPTH (nested, root) | source: sources two libs, declares NO function |
| `TestShell_ScriptWithFunctionsImportsFromEndResolves_6852` | path DEPTH | source: sources one lib AND declares two functions |
| `TestShell_NoCarrierWithoutASource_6852` | presence of `source`/`.` (absent) | full record set — script component, two functions, a CALLS edge — at both depths |
| `TestShell_OneCarrierPerFileNotPerSource_6852` | NUMBER of directives (4: two `source`, two `.`) | one file, one path, both depths |
| `TestShell_ScriptComponentNamedLikeThePathGetsNoSecondCarrier_6852` | path DEPTH (root fires clause 3, nested does not) | source (functions + one `source`) |
| `TestShell_SelfSourcingImportStubGetsNoSecondCarrier_6852` | path DEPTH (clause 3 fires at BOTH) | the script sources itself by its own path spelling; no functions |
| `TestShell_EmptyFileGetsNoCarrier_6852` | Extract return path (empty content) | — |
| `TestShell_RegexFallbackGetsNoCarrier_6852` | Extract return path (`TSTree == nil`) | source text still contains a `source` directive |
| `TestShell_CarrierShape_6852` | — | carrier fields, `Properties` provenance, #577 index 0 |

## Mutation ledger — 11 applied, 10 DEAD, 1 ALIVE-equivalent

| # | mutant | file | verdict | killed by |
|---|---|---|---|---|
| 1 | unconditional carrier (`FileEntity` prepended always) | shell.go | DEAD | `NoCarrierWithoutASource` (both depths), `ScriptComponentNamedLikeThePath` root, `SelfSourcingImportStub` (both depths) |
| 2 | placement moved ABOVE pass 4 (clause 3 blind to the script component) | shell.go | DEAD | `ScriptComponentNamedLikeThePath` root — and nothing else |
| 3 | placement moved ABOVE the walk (clause 2 blind to the stubs) | shell.go | DEAD | 5 tests incl. both resolution tests |
| 4 | `lang` → `""` | shell.go | DEAD | `CarrierShape` — `Properties["language"]` present |
| 4b | delete `Language: "shell"` from `makeImportStub` (scores the ABSENCE assertion in its own direction) | shell.go | DEAD | `CarrierShape` sibling-premise loop |
| 5 | `file.Path` → `basename(file.Path)` | shell.go | DEAD | 5 tests, every nested subtest |
| 6 | `PrependFileCarrier` → `FileCarrierFor` + append at end | shell.go | DEAD | `CarrierShape` #577 index-0 row |
| 7 | re-add shell to `knownMissingCarrier6847` | ledger | DEAD | `TestFileAnchoredCarrier_NoNewOffender_6847` ("no longer reproduces") |
| 9 | remove the carrier call with the ledger line still deleted | shell.go | DEAD | same guard, other direction ("NEW path-anchored FromID") |
| 10 | `lang` → `"bash"` | shell.go | DEAD | `CarrierShape` — `Language = "bash", want "shell"` |
| 8 | add shell to `nonTaggingCallers` | caller-set guard | DEAD | `TestCarrierCallerSetIsGradedFromSource_6861` — shell DOES tag |
| 11 | remove the shell `mustScan` anchor | caller-set guard | **ALIVE — equivalent under the current suite** | nothing; see below |

**The one ALIVE verdict, classified three ways rather than two.** Mutant 11 is
*equivalent under the current suite*, not *distinguishable and ungraded*: no
fixture can separate it, because `carrier_caller_set_6861_test.go`'s own prose
states the anchors are deliberate REDUNDANCY — every walk mutant scored against
that guard fires on the first anchor (`file_carrier.go`), so any six of the
seven could be dropped without a verdict changing. What the guard does grade is
the FLOOR on the anchor list (`minAnchors`), i.e. emptying it. Scored here
rather than relayed: `./internal/extractor` is green with the line removed.
So the `mustScan` line this PR adds is documentation of a file the guard makes
a claim about, and this diff adds NO graded roster entry — the #6861 interaction
in this arm is the negative, shell staying out of `nonTaggingCallers`.

Every mutant gated on its match count AFTER the edit (insertion mutants gate on a marker absent before and present exactly once after), `go vet` clean, whole-package `-count=1`, restored by `cp` from a shasum-verified backup with the sha re-checked. Final worktree diff carries no mutant.

## Packages run (`-count=1`, all green)

`./cmd/grafel/` `./internal/extractors/shell/` `./internal/extractors/` `./internal/extractor/` `./internal/resolve/` `./internal/quality/` — no golden moved, no digest moved. `go run ./tools/coverage check` 5/5.

Refs #6852

---

## Review round (commit `f335a7131`)

One real defect found, and it was in the file whose whole purpose is to stop it. `file_carrier.go`'s invariant paragraph said an empty `lang` is caught "only for a caller that does NOT run `TagEntitiesLanguage`", so for a tagging caller `""` is "equivalent under the suite". **shell tags, and shell's empty-token mutant is DEAD** — caught on provenance rather than on the token. The sentence was over-broad in the **permissive** direction: it under-claims grading for tagging callers.

Softened rather than deleted, per review: "equivalent ON THE `Language` FIELD ALONE — unless some test in that package distinguishes the two some other way, as shell's does", with the mechanism stated so a reader can check their own caller. Deliberately **not** turned into a roster of which tagging callers manage it — that is the shape #6861 rejected. The paragraph now also states the asymmetry that makes leaving it loose safe: under-claiming grading costs nothing, over-claiming is the failure.

Also corrected in this body, from the reviewer: the `mustScan` anchor is non-grading (scored ALIVE-equivalent, see the ledger); the above-pass-4 placement mutant has **zero corpus instances here** and is producible-by-any-real-repo rather than occurring in this tree; and gate **cost** vs citation **count** are different claims — fsharp had nine citations that merely carried no line numbers, while shell has none at all.

Comment-only change. `./internal/extractor/` and `./internal/extractors/shell/` green at `-count=1`, gofmt and `go vet` clean.

### DIFF
--- a/internal/extractor/carrier_caller_set_6861_test.go
+++ b/internal/extractor/carrier_caller_set_6861_test.go
@@ -199,6 +199,7 @@ func TestCarrierCallerSetIsGradedFromSource_6861(t *testing.T) {
 		{"internal/extractors/hcl/extractor.go", []string{prependFileCarrier}},
 		{"internal/extractors/html/extractor.go", []string{prependFileCarrier}},
 		{"internal/extractors/fsharp/extractor.go", []string{prependFileCarrier}},
+		{"internal/extractors/shell/shell.go", []string{prependFileCarrier}},
 	}
 
 	nonTest := 0
--- a/internal/extractor/file_carrier.go
+++ b/internal/extractor/file_carrier.go
@@ -108,9 +108,17 @@ import "github.com/cajasmota/grafel/internal/types"
 // TestErlang_CarrierIsLanguageTagged_6815). An EMPTY one is caught only for a
 // caller that does NOT run extractor.TagEntitiesLanguage: that helper fills an
 // empty Language with the extractor's own token, so for a caller that tags,
-// passing "" is equivalent under the suite, while a caller that does not tag
-// keeps whatever token this parameter is given. The second shape is the one the
-// parameter exists for.
+// passing "" is equivalent ON THE Language FIELD ALONE — unless some test in
+// that package distinguishes the two some other way, as shell's does (#6852):
+// TagEntitiesLanguage only touches a record whose Language is empty, and stamps
+// Properties["language"] when it does, so a package whose every other record
+// sets Language explicitly can observe the fill as provenance rather than as a
+// token. A caller that does not tag keeps whatever token this parameter is
+// given, and that second shape is the one the parameter exists for. The
+// asymmetry to keep in view is the direction of the error: this paragraph is
+// safe when it UNDER-claims grading for a tagging caller (a mutant dies that it
+// did not promise would), and unsafe the other way, so the clause above is a
+// softening and not a second measured fact about which callers do it.
 //
 // WHICH callers are which is deliberately NOT written down here. This paragraph
 // stated it as a MEASURED fact three times — "all three current callers", then
--- a/internal/extractor/file_carrier.go
+++ b/internal/extractor/file_carrier.go
@@ -76,6 +76,15 @@ import "github.com/cajasmota/grafel/internal/types"
 //     TestFSharp_ModuleNamedLikeThePathGetsNoSecondCarrier_6852, whose nested
 //     subtest is the contrast that stops it passing on a carrier that is never
 //     emitted at all.
+//     shell (#6852) reaches it by BOTH of those routes and by a third that is
+//     new: it emits hcl's BASENAME(path) file component (root depth only, and
+//     only for a script that declares a function), AND its `source`/`.` import
+//     stub is named after the SOURCED path VERBATIM — so a script that sources
+//     itself by the spelling of its own path is a path-named record at ANY
+//     depth. That is the case fsharp's import marker provably cannot produce,
+//     and it means clause 3 is not a root-depth phenomenon. Pinned by
+//     TestShell_ScriptComponentNamedLikeThePathGetsNoSecondCarrier_6852 and
+//     TestShell_SelfSourcingImportStubGetsNoSecondCarrier_6852.
 //
 // Clause 3 is checked for EVERY record, before the loop may short-circuit on
 // clause 2 being satisfied — deliberately, and the order is load-bearing.
--- a/internal/extractors/file_anchored_carrier_runtime_6847_test.go
+++ b/internal/extractors/file_anchored_carrier_runtime_6847_test.go
@@ -140,7 +140,6 @@ var knownMissingCarrier6847 = map[string]string{
 	"crystal": "extractor.go:161 `FromID: filePath` on IMPORTS; no carrier emitted",
 	"dart":    "path-anchored IMPORTS; no carrier emitted",
 	"lua":     "lua.go:540 `FromID: file.Path` on IMPORTS; no carrier emitted",
-	"shell":   "shell.go:260 `FromID: file.Path` on IMPORTS (source/.); no carrier emitted",
 	"zig":     "zig.go:272 `FromID: filePath` on IMPORTS; no carrier emitted",
 }
 
--- a/internal/extractors/shell/file_carrier_6852_test.go
+++ b/internal/extractors/shell/file_carrier_6852_test.go
@@ -1,0 +1,531 @@
+package shell_test
+
+// file_carrier_6852_test.go — #6852, shell arm.
+//
+// makeImportStub (shell.go) stamps `FromID: file.Path` on the IMPORTS edge of
+// every `source <path>` / `. <path>` stub. internal/resolve/refs.go has no
+// path→entity index, so a path-valued FromID resolves if and only if some
+// emitted node carries that exact string as its Name. Same defect #6815 fixed
+// in erlang/nim/groovy and #6852 fixed in bicep (#6864), terraform (#6871),
+// html (#6879) and fsharp (#6880); same fix, extractor.PrependFileCarrier.
+//
+// ONE ANCHORING SITE, N EDGES. makeImportStub is the only bare `FromID: path`
+// in the whole package — buildScriptComponent's CONTAINS edges and
+// extractCallRelationships' CALLS edges leave FromID EMPTY, and extractRegex
+// (the TSTree == nil fallback) emits no relationships at all. But one stub is
+// emitted per `source`/`.` command, so a script with four of them anchors four
+// edges and must still gain exactly ONE carrier.
+//
+// THE DEPTH SPLIT IS SHELL'S OWN SHAPE, and it is neither terraform's nor
+// html's. buildScriptComponent emits a file-level SCOPE.Component named
+// BASENAME(file.Path) — so at a ROOT path that name already IS the path and the
+// edge resolved by the accident #6367 documents, exactly as hcl's did. BUT it
+// is emitted ONLY when the file declares at least one function
+// (`if len(fnEntities) > 0`). So shell dangles at:
+//
+//   - every NESTED path, function or no function (basename != path); and
+//   - the ROOT path of a script with NO function definitions, where no
+//     file-level component is emitted at all.
+//
+// A shell script that only sources libraries and runs top-level commands is the
+// ordinary case, not a corner, so the root hole is real rather than notional.
+// Both holes are driven below, and the surviving accident (root + functions) is
+// pinned as a clause-3 rejection rather than left to chance.
+//
+// CLAUSE 3 IS REACHED BY TWO ROUTES HERE, one of which no earlier arm had.
+// FileCarrierFor clause 3 is `records[i].Name == path` (clause 1 is the
+// empty-path guard, clause 2 the anchoring test). It fires for shell when:
+//
+//   - the SCRIPT COMPONENT is named the path — root depth only, as hcl's is; or
+//   - an IMPORT STUB is named the path, because makeImportStub names the stub
+//     after the SOURCED path verbatim, so a script that sources ITSELF by the
+//     spelling of its own path mints a record named exactly the path. Unlike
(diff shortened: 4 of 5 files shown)


cajasmota:
## Coordinator verification — the second clause-3 route is graded, and graded where it matters

Eleven mutants from the author and eight from the review; I repeated none. Mine went at the claim that is genuinely new to this arm: that clause 3 is reachable by **two** routes here, and that the second — a self-sourcing script — works at *any* depth, where the script-component route can only fire at root.

`makeImportStub` names the stub after the sourced path verbatim, so that claim rests entirely on one field. I changed it to the basename the function already computes:

```go
 	return types.EntityRecord{
-		Name:       path,
+		Name:       leaf,
```

Post-edit match count asserted as a gate, `go vet` clean. `go test -count=1 ./internal/extractors/shell/` → **exit 1. DEAD.**

The kill is the informative part:

```
--- FAIL: TestShell_SelfSourcingImportStubGetsNoSecondCarrier_6852
    --- FAIL: .../scripts/ci/deploy.sh
```

It fails at the **nested** subtest. That is the cell the script-component route cannot reach — at nested depth the component is named `deploy.sh` while the path is `scripts/ci/deploy.sh`, so only the stub can be path-named there. So the test is not passing incidentally alongside the root case; it is pinning the specific route the arm claims is new.

## On the invariant softening

The added clause is right, and the sentence the author added beyond what was asked is the better half: naming the **direction** of the error. That paragraph is safe when it under-claims grading for a tagging caller — a mutant dies that it did not promise would — and unsafe in the other direction. A reader now knows the clause is a softening rather than a second measured fact, which is what stops it becoming the next thing #6861 has to police.

Declining to turn it into a roster of which tagging callers manage it was also correct. That is precisely the shape #6861 removed from this file.

## Recorded, not requested

The `mustScan` anchor this arm adds is **non-grading** — removing it leaves `./internal/extractor` green — and the author scored that themselves rather than relaying the review's verdict, classifying it as equivalent with the file's own redundancy prose as the reason. So this diff adds **no graded roster entry**; the #6861 interaction here is the negative, shell correctly staying out of `nonTaggingCallers`.

PR #6883, opened 2026-09-05
fix(#6854): the Dockerfile classifier matched only the exact basename — and routing the files revealed a new #6852 offender, so the carrier lands with it

### LINKED ISSUE #6854: The classifier recognises only the exact basename Dockerfile — every Docker fixture in this repo is silently unreachable
The classifier drops `Dockerfile.<variant>` filenames as `unsupported_extension`

## The defect

```
src/fixtures/real-world/docker/Dockerfile.multi_stage   lang=""            skip=true   reason=unsupported_extension
src/fixtures/real-world/docker/Dockerfile               lang="dockerfile"  skip=false
```

A bare `Dockerfile` classifies. `Dockerfile.multi_stage` does not — the trailing `.multi_stage` is read as an extension, found unsupported, and the file is skipped entirely.

`Dockerfile.<something>` is a conventional naming form (`Dockerfile.dev`, `Dockerfile.prod`, `Dockerfile.alpine`, and multi-stage variants), so this is not an exotic case. Every such file in an indexed repository is currently invisible.

## How it surfaced

During the independent review of #6847. A runtime guard drives every corpus file through the production classifier, and `dockerfile` did **not** appear among the exercised languages — despite the corpus containing two Docker fixtures. The natural reading was "no fixture exists for that language". The actual cause was that one of the two fixtures is dropped before extraction.

That matters beyond Docker: it is why a language with genuine corpus coverage can look unexercised, which in turn made an offender list look complete when it was not.

## Why it is worth fixing rather than tolerating

The failure is **silent**. `skip=true reason=unsupported_extension` is a normal, expected outcome for the vast majority of files in a repository, so a legitimately-supported file landing in that bucket produces no signal anywhere. There is no error, no warning, and no count that would look wrong.

## The fix

Recognise `Dockerfile` as a **prefix** form, not only as an exact basename — the same way other extensionless conventional filenames are presumably handled. Read how the classifier treats those before choosing the shape; do not special-case Docker if a general mechanism already exists.

**Score it against the real classifier**, not a unit table: assert `Dockerfile.multi_stage` classifies as `dockerfile`, and — the permissive direction, which matters more — that something like `Dockerfile.md` or `Dockerfilex` does **not**. A prefix match is exactly the shape that over-fires, and this repo has repeatedly found guards that grade recall and leave over-firing unobserved.

## Not established

- **Whether other conventional extensionless filenames have the same problem.** `Makefile.<variant>`, `Jenkinsfile.<variant>`, `Vagrantfile.<variant>` are the obvious candidates. **Not checked** — and checking is probably most of the value here, since the fix shape depends on whether this is one bug or a class.
- **How many such files appear in real corpora.** Unmeasured; the in-tree corpus has exactly one.
- Whether anything downstream assumes a Docker entity's source file is named exactly `Dockerfile`.

Found during the review of #6847; pre-existing and unrelated to that change. Related by consequence: **#6852** (path-anchored `IMPORTS` with no carrier), whose measured language list was incomplete partly because of this.

cajasmota:
Closes #6854
Refs #6852

Two changes shipped together. The ordering constraint between them runs **one
way**, and this section previously claimed a circularity that does not exist —
corrected here and in both code comments that asserted it, because the review
BUILT the split and ran it.

- The **classifier** widening **cannot land first**. It makes all three docker
  fixtures reach the extractor, where each becomes a NEW path-anchored-FromID
  offender, reddening `internal/extractors/file_anchored_carrier_runtime_6847_test.go`.
- The **carrier** arm **can** precede it. With the carrier line, its unit tests
  and the `carrier_caller_set_6861` row applied but `classifier.go` and the 6847
  guard at the parent, `./internal/extractors/`,
  `./internal/extractors/dockerfile/` and `./internal/extractor/` are all GREEN
  and every carrier mutant still dies — `file_carrier_6852_test.go` drives
  `Extract` with synthetic paths and never needs a real docker file to be
  routed.

They ship together because splitting them buys nothing, not because splitting
them is impossible. A stated impossibility that one command disproves is this
repo's dominant defect class, which is why the wording is corrected rather than
quietly dropped.

One PR. `./internal/extractors` is GREEN on this branch.

---

# Half 1 — #6854: the classifier matched one exact basename

## Step 1: the measurement, before any change

`find` over the whole tree for `Dockerfile`, `Dockerfile.*`, `*.Dockerfile`,
`Containerfile*`. Three fixtures, no bare `Dockerfile`, no `Containerfile`
anywhere:

| on-disk path | as the corpus walk spells it | before | after |
|---|---|---|---|
| `testdata/fixtures/dockerfile/sample.Dockerfile` | `src/fixtures/dockerfile/sample.Dockerfile` | `skip=true unsupported_extension` | `lang=dockerfile skip=false` |
| `testdata/fixtures/sources/dockerfile/sample.Dockerfile` | `src/fixtures/sources/dockerfile/…` | `skip=true unsupported_extension` | `lang=dockerfile skip=false` |
| `testdata/fixtures/real-world/docker/Dockerfile.multi_stage` | `src/fixtures/real-world/docker/…` | `skip=true unsupported_extension` | `lang=dockerfile skip=false` |

The other four `dockerfile`-named paths in the tree are Go sources, a YAML rules
file and directories — none is a container build file.

The decision lives in `internal/classifier/classifier.go`: `basenameLanguageMap`
maps the exact strings `Dockerfile` and `Containerfile`, consulted only after
`extensionLanguageMap` misses. So `*.Dockerfile` was read as extension
`.dockerfile` (unknown) and `Dockerfile.<variant>` as `.multi_stage` (unknown).
The issue's corrected comment is confirmed in full.

### A third correction the issue does not have

At their REAL repo-root paths those three files never reach `detectLanguage` —
`testdata` is a `depDir`, so `universalPathSkip` returns `vendor_testdata` two
steps earlier (and `build/` returns `vendor_build`, which cost the first draft of
the test a false red on `build/Dockerfile.ci`). The `src/fixtures/...` rows come
from the #6847 corpus walk, which rewrites `/testdata/` to `/src/` before
classifying. Every path in the new test is spelled the way the classifier
actually receives it, so the test grades the docker rule and not the vendor rule.

## Step 2: the change

Two routes, because they are two different facts:

- **`<name>.Dockerfile` is an extension.** `.dockerfile` / `.containerfile`
  added to `extensionLanguageMap`. Routing it here rather than by a basename
  rule is what makes `LanguageForExtension(".dockerfile")` answer, which stops
  the unsupported-extension report listing an extension the router does claim.
  Case-insensitive by construction.
- **`Dockerfile.<variant>` is a basename.** New `containerVariantLanguage`,
  called last in `detectLanguage`, case-sensitive like the bare-name table it
  sits beside. The variant must be exactly one non-empty segment.

### Over-firing, and why ordering alone was not enough

Running last means anything whose trailing segment names a known language
returned from the extension lookup already, so `Dockerfile.md` stays markdown and
`Dockerfile.py` stays python with no denylist naming them.

That argument does **not** reach a segment the router claims nothing for. The
first cut was caught by a pre-existing test —
`TestMX1100_DockerfileWithExtension_NotDockerfile`, which has forbidden
`Dockerfile.bak` since long before this issue.

**Review round: the blocklist still accepted everything it did not name**, and
the harm is concrete rather than theoretical — a backup of a real Dockerfile
still contains its FROM lines, so it clears the extractor's `stageCount == 0`
early return and mints a duplicate `SCOPE.Component` plus a file carrier for a
file nobody builds. Three changes:

1. **Two SHAPE rules**, because no list can enumerate these.
   - A trailing `~` is **rejected outright, not trimmed before the lookup**.
     Trimming rejects `Dockerfile.bak~` and *accepts* `Dockerfile.dev~`, since
     "dev" is not a blocked segment — so the two implementations differ on
     exactly the second escape the review named. Mutant **M14** applies the
     trim-then-lookup version and dies **only** on the "tilde on an accepted
     segment" assertions; a test set listing only `.bak~` would have scored it
     ALIVE.
   - A **purely numeric** segment is a revision or date stamp (`Dockerfile.1`,
     `Dockerfile.20260101`). `v1`, `1x`, `go1`, `node20` are not purely numeric
     and stay accepted — the contrast that kills **M16**, which widens the rule
     to "starts with a digit" while still rejecting `Dockerfile.1`.
2. **19 more named segments**: rpm and dpkg conffile artefacts (both renamers
   leave the file they replace beside a real Dockerfile in any distro-based
   image), encryption and key material, and the remaining backup/state words.
3. **The set is graded by DERIVATION, not by a roster.** The old floor test is
   gone — a floor catches shrinkage and never a key added with no row beside it,
   which the review correctly called out as zero slack.
   `TestDockerfileDenylistIsExhaustivelyGraded6854` is *internal* so it can read
   the map, and per entry asserts both that the classifier rejects
   `Dockerfile.<entry>` **and** that `extensionLanguageMap` claims nothing for
   `.<entry>` (a shadowed entry grades nothing — the #6861 shape).

**That derived test earned its keep on its first run.** `sig` failed it: `.sig`
is Standard ML's signature extension and is *already routed*, so the entry was
shadowed dead weight and `Dockerfile.sig` classifies as **sml**, not dockerfile.
The entry is removed and the forbidden row now asserts the `sml` answer — the
property that matters is "not dockerfile", and here the *ordering* is what
delivers it. A roster-shaped check would have carried that entry indefinitely.

### The bare dotfile is a choice now, not an accident

`.dockerfile` (and `foo/.dockerfile`) classifies, because Go's
`filepath.Ext(".dockerfile")` returns the whole name. Kept and pinned as an
**accept**, because it is not a docker carve-out: `.md` is markdown and `.go` is
go by the same route, so excluding it would make container files the one
extension in the table that answers differently for a dotfile.
`TestDockerfileBareDotfileIsDeliberate6854` asserts the siblings too, so if that
uniformity ever changes this row goes red with it.

### Two forbidden rows deleted for being vacuous

`Dockerfile/README` and `src/Dockerfile/notes` passed but could not **fail**
under any mutation of this rule — `containerVariantLanguage` only ever receives
`path.Base(norm)`, so they graded `path.Base`. A row that cannot fail is a claim
wearing a forbidden row's clothes, and the file's own standard is that forbidden
rows constrain the width of *this* rule.

### Grading, both directions

15 accepted shapes and 61 forbidden ones in the emitted-decision table, all
asserted on `ClassifyResult` (`Language` **and** `Skip`), never on a map or a
counter — plus the two derived internal tests above, which are exhaustive over
the map rather than hand-listed.

---

# Half 2 — #6852 arm: dockerfile's FROM IMPORTS

## What was measured (as every arm reports)

- **Anchoring sites: ONE.** `collectFrom` stamps `FromID: file.Path` on the
  IMPORTS edge of every FROM. It is the only producer of a path-anchored FromID
  left in the package — #6367 already moved USES and CONTAINS off the path (they
  leave FromID empty so graph assembly stamps the owning record's id) and left
  the IMPORTS anchor under the #120 convention.
- **They share ONE anchor**, but the site runs once per FROM, so a file anchors
  N edges on that one path and must still gain exactly ONE carrier. On the real
  fixtures: `sample.Dockerfile` ×2 anchor **2** each, `Dockerfile.multi_stage`
  anchors **3**.
- **Does the extractor already emit something named after the file?** Yes, but
  not the path: `buildDockerfileEntity` names its single `SCOPE.Component`
  **BASENAME(path)**. So this is the **terraform/hcl shape** — the ROOT path
  resolved BY ACCIDENT (basename == path) and only nested paths dangled — not
  the html/fsharp shape where both depths dangle. #6367 measured the same thing:
  "3 of 3 DANGLING at services/api/Dockerfile, and correct only BY ACCIDENT at a
  root Dockerfile".
- **All three fixtures were offenders**, not two. The two-path list in my earlier
  report was a truncated grep of the failure output, not a smaller population.

## The change

`extractor.PrependFileCarrier(file.Path, "dockerfile", …)` at Extract's final
return, CONDITIONAL. Both of FileCarrierFor's rejections are exercised:

- **clause 2** is unreachable from that line — a FROM-less Dockerfile already
  returned with no records at all — so the over-emission control is that early
  return, asserted rather than assumed;
- **clause 3 fires in production at every ROOT path**, where the component's
  BASENAME name IS the path. A carrier there would land a second
  `SCOPE.Component` under the component's own `graph.EntityID`, which does not
  hash Subtype (#6369/#6480).

Acceptance, both depths and both directions:

| test | what it grades |
|---|---|
| `TestDockerfile_FromImportsFromEndResolves_6852` | FROM end resolves at `services/api/Dockerfile` **and** `Dockerfile`; the root case passed before the change, so a root-only test would have graded nothing |
| `TestDockerfile_FixtureSpellingsResolve_6852` | the five real spellings — both `sample.Dockerfile` fixture paths, `Dockerfile.multi_stage`, `Containerfile`, `api.Containerfile` — so a basename-shaped fix cannot pass |
| `TestDockerfile_RootPathGetsNoSecondCarrier_6852` | clause 3 at root, with a nested contrast subtest, plus root `Containerfile` |
| `TestDockerfile_OneCarrierPerFileNotPerFrom_6852` | 3 FROMs → 3 anchored edges → exactly 1 carrier; also re-asserts #6367 (only IMPORTS may anchor) |
| `TestDockerfile_NoFromGetsNoCarrier_6852` | over-emission: a FROM-less file gains nothing, at both depths |
| `TestDockerfile_EmptyFileGetsNoCarrier_6852` | Extract's other return path |
| `TestDockerfile_CarrierShape_6852` | token, SourceFile, zero relationships, record 0 (#577), exactly 2 records |

### `TagEntitiesLanguage`: called — and the token is still directly graded

`internal/extractors/dockerfile` **does** call `extractor.TagEntitiesLanguage`
(dockerfile.go:145), so it owes **no `nonTaggingCallers` entry** — and this is
the one caller where that classification understates the grading rather than
overstating it. The tagging call sits ABOVE the carrier line and only fills an
EMPTY `Language`, so a carrier prepended afterwards keeps whatever the argument
gave it: `lang=""` stays `""` and is caught on `Language` alone. The premise is
pinned too — `TestDockerfile_CarrierShape_6852` asserts the carrier carries no
`Properties["language"]`, which is what a post-hoc tag would add; if the carrier
were ever moved above the tagging call, that assertion goes red rather than
quietly grading nothing.

Its `mustScan` anchor is added to `carrier_caller_set_6861_test.go` alongside
fsharp's. `knownMissingCarrier6847` gains **nothing** — dockerfile was never in
it, and that file forbids adding an entry to go green.

---

# Mutants — 23 run, 23 DEAD

Each: exact edit, **match count asserted after the edit as a hard gate** (a
deletion gate additionally checks the byte delta), `go vet` clean, whole-package
`go test -count=1`, restored by `cp` from a shasum-verified backup, with the
restored sha re-verified after every mutant.

**Classifier half (8):**

|---|---|---|---|
| M1 | prefix rule claims every basename that reaches it | DEAD | forbidden test + 12 pre-existing tests |
| M2 | both routes removed — back to the exact basename | DEAD | accept test, router-visibility test |
| M3 | suffix route only removed (the prefix-only reading of the issue) | DEAD | accept test, router-visibility test |
| M4 | `nonContainerVariantSuffixes` check deleted | DEAD | forbidden test + `TestMX1100_…` |
| M5a | empty-variant guard deleted (`Dockerfile.`) | DEAD | forbidden test |
| M5b | single-segment guard deleted (`Dockerfile.a.b`) | DEAD | forbidden test |
| M6 | docker rule moved ABOVE the extension lookup | DEAD | forbidden test |
| M7 | prefix rule made case-insensitive | DEAD | forbidden test |
| M13 | trailing-`~` rejection deleted | DEAD | forbidden rows + both tilde subtests |
| M14 | `~` **trimmed** before the lookup instead of rejected | DEAD | **only** the "tilde on an accepted segment" subtest + the `Dockerfile.dev~` / `Containerfile.prod~` rows |
| M15 | purely-numeric rejection deleted | DEAD | forbidden rows + numeric subtest |
| M16 | numeric rule widened to "starts with a digit" | DEAD | **only** the numeric subtest's accepted contrast rows (`v1`, `1x`, `go1`, `node20`) |
| M17 | blocklist lookup deleted (re-score of M4 over the enlarged set) | DEAD | derived test on all 35 entries + `TestMX1100_…` |
| M18 | `isContainerBuildTarget` returns true for everything (re-score of M1) | DEAD | derived test + forbidden rows |
| M19 | **the separator widened from `.` to also accept `-`** | **was ALIVE, now DEAD** | the three new separator rows, and nothing else |

**M19 is the one that was ALIVE on review**, and it is worth reading before the
rest. `containerVariantLanguage` cuts `stem + "."` and nothing else; a mutant
adding a `stem + "-"` fallback survived `./internal/classifier`,
`./internal/extractors`, `./internal/extractors/dockerfile` and
`./internal/extractor` — every package that could see it — so `Dockerfile-dev`,
`Containerfile-prod` and `app/Dockerfile-slim` all classified as dockerfile
under it. Verdict three of the three, *distinguishable and ungraded*: not
equivalent, not production-unreachable, and demonstrated with a throwaway probe
plus the positive control that the same probe passes on the restored tree.

`grep 'Dockerfile-\|Containerfile-'` over `internal/classifier/` returned
**nothing**, tests included. Every other part of the accept predicate was
graded — stem, dot count, case, denylist, both shape rules — while the separator
that gates all of them was not: one axis pinned and its neighbour left open, in
the file whose thesis is that over-matching is the invisible direction. Fixed
with **three forbidden rows and no production change** (the behaviour was
correct, just unobserved), the nested one holding the depth axis explicitly since
`path.Base` runs first. M19 now dies on exactly those three subtests and nothing
else, which is the evidence they are what grades it.

M5a/M5b are the two halves of one compound condition, scored separately so
neither is graded only in company with the other. M14 and M16 are the two that
matter most: each is a *plausible* implementation that passes every assertion
except one, and in both cases that one assertion was added because the mutant
was written first.

**Carrier half (6) and the guard sets (2):**

|---|---|---|---|
| D1 | UNCONDITIONAL carrier | DEAD | `RootPathGetsNoSecondCarrier` (+9 pre-existing) |
| D2 | carrier deleted — pre-#6852 return | DEAD | 5 of the arm's tests |
| D3 | `lang=""` | DEAD | `CarrierShape` (on `Language`, directly) |
| D4 | `lang="docker"` | DEAD | `CarrierShape` |
| D5 | carrier named BASENAME instead of the path | DEAD | 5 of the arm's tests |
| D6 | `FileCarrierFor` clause 3 deleted | DEAD | `RootPathGetsNoSecondCarrier`, scored against `./internal/extractors/dockerfile/` alone so another arm's suite cannot take the credit |
| G1b | `dockerfile` removed from `exercisedLanguages6847` | DEAD | `CorpusCoverage_6847` |
| G2b | `dockerfile` removed from `anchoringLanguages6847` | DEAD | `CorpusCoverage_6847` |

G1b/G2b are **re-scores after the shell rebase**, not the pre-rebase results.
Both set literals gained `shell`, so the pre-rebase edits no longer match the
file — the old mutants would have failed their own match-count gate rather than
reporting a verdict, which is exactly why a mutant whose premise is a set that
moved gets re-run rather than carried.

D5 is the one worth naming: a carrier named BASENAME is emitted at every depth,
looks exactly like the fix to any "a file carrier exists" assertion, and resolves
nothing at a nested path — it reinstates the #6367 accident wearing the fix's
clothes.

No mutant survived, so the three-way ALIVE classification does not arise.

---

# Package set

Run, `-count=1`, whole package, all **green**:

- `./internal/classifier` — and `gofmt -l` / `go vet` clean
- `./internal/extractors/dockerfile`
- `./internal/extractors` — **the guard that this branch was blocked on**
- `./internal/extractor`
- `./internal/resolve`
- `./internal/quality`, `./internal/quality/{analytics,audit,fitness}` — no
  golden moved; there is no docker fixture in `internal/quality/golden/`, so the
  extraction-quality ratchet is untouched
- `./internal/cli` — where the unsupported-extension report is asserted
  end-to-end (run before the carrier half; unaffected by it)
- `./cmd/grafel` — 139s. **The whole-graph digest did NOT move**, and this is
  NOT coverage of the carrier change: `writeSpanFixture6118` has no docker
  member.
- `go run ./tools/coverage check` — 5/5. Coverage cost zero:
  `docs/coverage/registry.json` cites `internal/extractors/dockerfile/dockerfile.go`
  as a FILE only, with no symbol-anchored `file.go:NNN` citation, so the line
  shift invalidated nothing.

**Not run**, and the absence claim does not extend to them: every other package
in the module, including `./internal/engine`, `./internal/substrate` and the
other per-language `./internal/extractors/...` subpackages.

# Rebase: PR #6882 (shell) landed first

Rebased onto `20fbb6e50`. One conflict (`mustScan`, resolved by keeping both
rows). `file_anchored_carrier_runtime_6847_test.go` **auto-merged with no
conflict**, which is precisely the case where a silent drop hides, so every
shared counter below was **re-derived from disk after the rebase**, not carried
forward. Three of the numbers in the previous version of this section were
wrong:

| counter | pre-rebase claim | re-derived from disk | note |
|---|---|---|---|
| `len(knownMissingCarrier6847)` | 8 | **7** | shell's line removed by its arm; astro, clojure, cobol, crystal, dart, lua, zig |
| `mustScan` anchors | 10 | **11** | shell's row + dockerfile's, both present after the conflict resolution |
| external carrier call sites | 8 | **9** | + shell |
| `nonTaggingCallers` | 3 | **3** | bicep, hcl, html — shell added none, so shell tags too |
| `exercisedLanguages6847` | 42 | **42** | shell was ALREADY in it (it was a `knownMissingCarrier` entry, so its corpus files were always exercised); 41 + dockerfile |
| `anchoringLanguages6847` | 35 | **35** | same reason |
| `minExternalCallers` / `minAnchors` / `minCorpusFiles6847` | — | **5 / 6 / 700** | floors, untouched |

Both set literals were also checked for sortedness and duplicates after the
merge (both sorted, no dupes), since an auto-merge is exactly what would produce
a duplicate or a dropped entry without a conflict marker.

## One thing the rebase surfaced, in `internal/extractor/file_carrier.go`

Shell's arm narrowed the `MEASURED` invariant to say an empty `lang` is, for a
tagging caller, equivalent "ON THE `Language` FIELD ALONE — unless some test in
that package distinguishes the two some other way, as shell's does". That clause
covers shell and fsharp, which observe the fill as **provenance**
(`Properties["language"]`).

dockerfile escapes it by a mechanism no test-side clause describes: **call
order**. `TagEntitiesLanguage` fills only the records handed to it, and
`dockerfile.go` tags at `:145` and prepends the carrier at `:183`, so the carrier
is never offered to it and an empty token is caught on `Language` directly. For
this caller the paragraph's claim is simply false. The paragraph's own asymmetry
note makes under-claiming the safe direction, so nothing shipped broken — but
leaving it would tell the next arm its tagging caller *cannot* grade the empty
token on `Language`, and it can. Named as a **mechanism** with dockerfile as its
example, not as a roster (#6861's subject is that roster sentences here get
falsified); a caller leaves the category the moment its carrier line moves above
its tagging call, which is what
`TestDockerfile_CarrierShape_6852`'s `Properties["language"]` assertion catches.

# Also not established

- Whether `Makefile.<variant>`, `Jenkinsfile.<variant>`, `Vagrantfile.<variant>`
  and their `*.Makefile`-style suffix forms have the same classifier defect. The
  issue calls this "probably most of the value" and it is still **unchecked**.
  `basenameLanguageMap` holds `Justfile`, `justfile`, `.justfile`, `nginx.conf`,
  `Caddyfile`, `luarocks.lock` and the two container names; none of the other
  three is in the table at all, so the question is whether they belong there —
  a different issue from this one.
- Whether anything downstream assumes a Docker entity's source file is named
  exactly `Dockerfile`. `./cmd/grafel` and `./internal/quality` green is evidence
  against, not proof.

### DIFF
--- a/internal/classifier/classifier.go
+++ b/internal/classifier/classifier.go
@@ -605,6 +605,15 @@ var extensionLanguageMap = map[string]string{
 	// does claim. Being an extension route also makes this form
 	// case-insensitive, unlike the case-sensitive `Dockerfile.<variant>` rule
 	// in containerVariantLanguage.
+	//
+	// A CONSEQUENCE THAT IS CHOSEN, NOT OVERLOOKED: a bare dotfile named
+	// `.dockerfile` classifies, because filepath.Ext(".dockerfile") returns the
+	// whole name. That is how EVERY entry in this map behaves — a file named
+	// `.md` is markdown, `.go` is go — so carving out an exception here would
+	// make container files the one extension in the table that answers
+	// differently for a dotfile. The uniform behaviour is kept and pinned
+	// (TestDockerfileBareDotfileIsDeliberate6854) rather than left as an
+	// unexamined accept.
 	".dockerfile":    "dockerfile",
 	".containerfile": "dockerfile",
 	// Markdown / Documentation
@@ -789,14 +798,20 @@ func detectLanguage(norm string) string {
 // markdown and `Dockerfile.py` stays python without this function having to
 // name them.
 //
-// THE OTHER HALF IS nonContainerVariantSuffixes, AND ORDERING ALONE IS NOT
-// ENOUGH — that is not a hypothetical. `Dockerfile.bak` is a backup file, not a
-// build target, and `.bak` is routed by nothing, so the ordering argument does
-// not reach it; TestMX1100_DockerfileWithExtension_NotDockerfile has forbidden
+// THE OTHER HALF IS isContainerBuildTarget, AND ORDERING ALONE IS NOT ENOUGH —
+// that is not a hypothetical. `Dockerfile.bak` is a backup file, not a build
+// target, and `.bak` is routed by nothing, so the ordering argument does not
+// reach it; TestMX1100_DockerfileWithExtension_NotDockerfile has forbidden
 // exactly that name since long before #6854 and caught the first cut of this
-// function. Every entry in that set is one the extension router does NOT claim,
-// which is what keeps the set load-bearing rather than shadowed (asserted in
-// dockerfile_variants_6854_test.go).
+// function.
+//
+// A BLOCKLIST OVER AN UNBOUNDED SEGMENT SPACE ACCEPTS EVERYTHING IT DOES NOT
+// NAME, so the two rules that are NOT list membership carry most of the weight.
+// The harm is concrete rather than theoretical: a backup of a real Dockerfile
+// still contains its FROM lines, so it clears the extractor's own
+// `stageCount == 0` early return and mints a duplicate SCOPE.Component plus a
+// file carrier for a file nobody builds. `~`-suffixed and purely numeric
+// segments are rejected by SHAPE precisely because no list can enumerate them.
 //
 // The variant must be exactly one non-empty segment: `Dockerfile.` and
 // `Dockerfile.a.b` are rejected, and a bare `Dockerfilex` has no separator at
@@ -810,7 +825,7 @@ func containerVariantLanguage(base string) string {
 		if !ok || variant == "" || strings.Contains(variant, ".") {
 			continue
 		}
-		if _, blocked := nonContainerVariantSuffixes[strings.ToLower(variant)]; blocked {
+		if !isContainerBuildTarget(variant) {
 			continue
 		}
 		return "dockerfile"
@@ -818,31 +833,111 @@ func containerVariantLanguage(base string) string {
 	return ""
 }
 
-// nonContainerVariantSuffixes are trailing segments that mark a file's STATE or
-// its format rather than a build target, so `Dockerfile.<one of these>` is not
-// a Dockerfile. None of them is claimed by extensionLanguageMap — if one ever
-// is, its entry here becomes dead weight, which is why the test asserts the
-// absence rather than trusting this comment.
+// isContainerBuildTarget reports whether variant names a BUILD TARGET rather
+// than a file's state, revision or encoding. Three rules, in order, and the
+// first two exist because the third cannot be complete:
+//
+//  1. A TRAILING `~` is the editor backup marker, and it is the marker whatever
+//     precedes it — `Dockerfile.dev~` is a backup OF `Dockerfile.dev`, not a
+//     target named "dev~". Rejecting on the tilde rather than trimming it
+//     before the lookup is the difference between forbidding both
+//     `Dockerfile.bak~` AND `Dockerfile.dev~`, and forbidding only the first:
+//     trimming would hand "dev" to the lookup, which accepts it.
+//  2. A PURELY NUMERIC segment is a revision or a date stamp — `Dockerfile.1`,
+//     `Dockerfile.20260101` — never a target. `v1` is not numeric and is
+//     accepted, which is the point of testing digits rather than "starts with
+//     a digit".
+//  3. Membership of nonContainerVariantSuffixes, for the named conventions
+//     that have neither shape.
+func isContainerBuildTarget(variant string) bool {
+	if strings.HasSuffix(variant, "~") {
+		return false
+	}
+	if isAllDigits(variant) {
+		return false
+	}
+	_, blocked := nonContainerVariantSuffixes[strings.ToLower(variant)]
+	return !blocked
+}
+
+// isAllDigits reports whether s is non-empty and every byte is an ASCII digit.
+// Deliberately byte-wise rather than unicode.IsDigit: a variant carrying, say,
+// an Arabic-Indic digit is not a revision suffix any tool produces, and
+// widening this would reject names nothing generates.
+func isAllDigits(s string) bool {
+	if s == "" {
+		return false
+	}
+	for i := 0; i < len(s); i++ {
+		if s[i] < '0' || s[i] > '9' {
+			return false
+		}
+	}
+	return true
+}
+
+// nonContainerVariantSuffixes are trailing segments naming a file's STATE,
+// FORMAT or ENCODING rather than a build target, so `Dockerfile.<one of these>`
+// is not a Dockerfile.
+//
+// Grouped by the convention that produces them, because that is what makes the
+// list auditable — a reader can ask "is my package manager's suffix here?"
+// rather than scanning an alphabetical blur.
+//
+// Every entry is graded, and derived rather than rostered:
+// TestDockerfileDenylistIsExhaustivelyGraded6854 (internal, so it can read this
+// map) walks it and asserts, per entry, that the classifier rejects
+// `Dockerfile.<entry>` AND that extensionLanguageMap claims nothing for
+// `.<entry>` — an entry the router already claims would be shadowed dead
+// weight. Adding a key here therefore cannot go ungraded, which a floor over a
+// hand-written table could not promise.
 //
 // Deliberately NOT listed: `example`, `sample`, `template`, `in`, `dist`. Those
 // name a Dockerfile that has not been rendered yet, not a non-Dockerfile.
 var nonContainerVariantSuffixes = map[string]struct{}{
-	"bak":    {},
-	"backup": {},
-	"old":    {},
-	"orig":   {},
-	"rej":    {},
-	"save":   {},
-	"swp":    {},
-	"swo":    {},
-	"tmp":    {},
-	"temp":   {},
-	"log":    {},
-	"txt":    {},
-	"json":   {},
-	"patch":  {},
-	"diff":   {},
-	"lock":   {},
+	// Editor and generic backups.
+	"bak":      {},
+	"backup":   {},
+	"bkp":      {},
+	"old":      {},
+	"new":      {},
+	"copy":     {},
+	"sav":      {},
+	"save":     {},
+	"swp":      {},
+	"swo":      {},
+	"tmp":      {},
+	"temp":     {},
+	"disabled": {},
+	// Merge / patch artefacts.
+	"orig":  {},
+	"rej":   {},
+	"patch": {},
+	"diff":  {},
+	// Package-manager conffile artefacts. dpkg and rpm both rename the file
+	// they are replacing rather than deleting it, so these land next to a real
+	// Dockerfile in any image built from a distro base.
+	"rpmsave":   {},
+	"rpmnew":    {},
+	"rpmorig":   {},
+	"dpkg-old":  {},
+	"dpkg-new":  {},
+	"dpkg-dist": {},
+	"dpkg-bak":  {},
+	// Encrypted or signed payloads. The bytes are not text, and handing them to
+	// a text extractor is worse than dropping them.
+	"gpg": {},
+	"enc": {},
+	"age": {},
+	"asc": {},
+	"pem": {},
+	"crt": {},
+	"key": {},
+	// Data / document formats the extension router does not claim.
+	"log":  {},
+	"txt":  {},
+	"json": {},
+	"lock": {},
 }
 
 // ---------------------------------------------------------------------------
--- a/internal/classifier/classifier.go
+++ b/internal/classifier/classifier.go
@@ -597,6 +597,16 @@ var extensionLanguageMap = map[string]string{
 	// COBOL programs a job invokes. Routed to the jcl extractor, which emits
 	// the JCL→COBOL cross-language bridge (#2843).
 	".jcl": "jcl",
+	// Container build files (#6854). `<name>.Dockerfile` / `<name>.Containerfile`
+	// is the spelling editors and language servers key on, and it is the one
+	// this repo's own fixtures use. It is routed HERE rather than by a basename
+	// rule so that LanguageForExtension(".dockerfile") answers too: otherwise
+	// the unsupported-extension report keeps listing an extension the router
+	// does claim. Being an extension route also makes this form
+	// case-insensitive, unlike the case-sensitive `Dockerfile.<variant>` rule
+	// in containerVariantLanguage.
+	".dockerfile":    "dockerfile",
+	".containerfile": "dockerfile",
 	// Markdown / Documentation
 	".md":       "markdown",
 	".mdx":      "markdown",
@@ -762,7 +772,77 @@ func detectLanguage(norm string) string {
 	}
 	// Fall back to basename matching for files like Dockerfile / Containerfile
 	// that carry no extension.
-	return basenameLanguageMap[base]
+	if lang, ok := basenameLanguageMap[base]; ok {
+		return lang
+	}
+	return containerVariantLanguage(base)
+}
+
+// containerVariantLanguage recognises the `Dockerfile.<variant>` /
+// `Containerfile.<variant>` prefix convention (#6854) — Dockerfile.dev,
+// Dockerfile.prod, Dockerfile.multi_stage — which the exact-basename table
+// above cannot express.
+//
+// IT RUNS LAST, AND THAT ORDERING IS HALF THE OVER-FIRING DEFENCE. Anything
+// whose trailing segment names a language the router already knows returned
+// from the extension lookup before reaching here, so `Dockerfile.md` stays
+// markdown and `Dockerfile.py` stays python without this function having to
+// name them.
+//
+// THE OTHER HALF IS nonContainerVariantSuffixes, AND ORDERING ALONE IS NOT
+// ENOUGH — that is not a hypothetical. `Dockerfile.bak` is a backup file, not a
+// build target, and `.bak` is routed by nothing, so the ordering argument does
+// not reach it; TestMX1100_DockerfileWithExtension_NotDockerfile has forbidden
+// exactly that name since long before #6854 and caught the first cut of this
+// function. Every entry in that set is one the extension router does NOT claim,
+// which is what keeps the set load-bearing rather than shadowed (asserted in
+// dockerfile_variants_6854_test.go).
+//
+// The variant must be exactly one non-empty segment: `Dockerfile.` and
+// `Dockerfile.a.b` are rejected, and a bare `Dockerfilex` has no separator at
+// all. The match is case-SENSITIVE, matching basenameLanguageMap — bare
+// lowercase `dockerfile` does not classify today and `dockerfile.dev` does not
+// either. The `<name>.Dockerfile` SUFFIX form is not handled here: it is a real
+// extension and is routed by extensionLanguageMap.
+func containerVariantLanguage(base string) string {
+	for _, stem := range [...]string{"Dockerfile", "Containerfile"} {
+		variant, ok := strings.CutPrefix(base, stem+".")
+		if !ok || variant == "" || strings.Contains(variant, ".") {
+			continue
+		}
+		if _, blocked := nonContainerVariantSuffixes[strings.ToLower(variant)]; blocked {
+			continue
+		}
+		return "dockerfile"
+	}
+	return ""
+}
+
+// nonContainerVariantSuffixes are trailing segments that mark a file's STATE or
+// its format rather than a build target, so `Dockerfile.<one of these>` is not
+// a Dockerfile. None of them is claimed by extensionLanguageMap — if one ever
+// is, its entry here becomes dead weight, which is why the test asserts the
+// absence rather than trusting this comment.
+//
+// Deliberately NOT listed: `example`, `sample`, `template`, `in`, `dist`. Those
+// name a Dockerfile that has not been rendered yet, not a non-Dockerfile.
+var nonContainerVariantSuffixes = map[string]struct{}{
+	"bak":    {},
+	"backup": {},
+	"old":    {},
+	"orig":   {},
+	"rej":    {},
+	"save":   {},
+	"swp":    {},
+	"swo":    {},
+	"tmp":    {},
+	"temp":   {},
+	"log":    {},
+	"txt":    {},
+	"json":   {},
+	"patch":  {},
+	"diff":   {},
+	"lock":   {},
 }
 
 // ---------------------------------------------------------------------------
--- a/internal/classifier/dockerfile_denylist_6854_internal_test.go
+++ b/internal/classifier/dockerfile_denylist_6854_internal_test.go
@@ -1,0 +1,123 @@
+// Package classifier — dockerfile_denylist_6854_internal_test.go
+//
+// #6854 review item 2. The external table in dockerfile_variants_6854_test.go
+// grades the emitted ClassifyResult for a hand-written list of shapes. A
+// hand-written list over a MAP has one failure mode it cannot cover: a key
+// added to nonContainerVariantSuffixes with no row added beside it. A floor on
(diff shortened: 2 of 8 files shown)


cajasmota:

Reviewed at `77d34fa83` in a separate detached worktree; merge-base `21c7b7dbd` is still `main`, so nothing moved underneath. All six claims in the PR body verified; **13/13 mutants DEAD**, each gated on an asserted post-edit substitution count and restored by `cp` from a shasum-verified backup.

Green, `-count=1`: `./internal/classifier/`, `./internal/extractors/dockerfile/`, `./internal/extractor/`, `./internal/resolve/`, `./internal/extractors/`, `./internal/quality/...`, `./cmd/grafel/`. `go vet` clean, `gofmt -l` empty. Not run: the extraction-quality ratchet (CI-only) — confirmed unaffected, `internal/quality/golden` contains no docker member, the only three docker files in the tree are under `testdata/fixtures/`.

### What held up

The measurement is right, and **the `src/` spelling is load-bearing**: at their on-disk paths all three fixtures return `skip=true reason=vendor_testdata` two steps before `detectLanguage`, and only the #6847 walk's `/testdata/` → `/src/` rewrite gets them to the docker rule. A test written against the on-disk path would grade the vendor rule instead — the shipped tests use the walk's spelling.

The arm reproduces: reverting only the carrier line reddens the guard with **all three** fixtures listed, `kinds=[IMPORTS]`. Exactly one non-empty `FromID` in the package's non-test sources (`dockerfile.go:371`, `collectFrom`) — the three other hits are #6367 comments explaining the deliberately-empty ones — and 3 FROMs yielding exactly 1 carrier is asserted with a kind-check, so the premise cannot pass via a USES/CONTAINS regression.

Claim 5 is the interesting one and it is confirmed. `TagEntitiesLanguage` at `:145` sits **above** the carrier at `:183`, so an empty token survives to the carrier and dies on `Language` directly — unusual for a tagging caller. `CarrierShape_6852` pins that premise by asserting `Properties["language"]` is absent, so moving the carrier above the tagging call goes red rather than silently ungrading.

### What blocks

**1 — the stated circularity is false in one direction.** The reviewer built the carrier-first split and ran it: carrier line, its unit tests, and the `carrier_caller_set_6861` row, with `classifier.go` and the 6847 guard at the parent. `./internal/extractors/`, `./internal/extractors/dockerfile/`, `./internal/extractor/` all **green**, and M9–M12 all still die — `file_carrier_6852_test.go` drives `Extract` with synthetic paths and never needs a routed docker file. The real constraint is one-directional: the *classifier* half cannot precede the carrier.

The PR is not being re-split; the prose is being corrected, in all three places that assert it (this body, the comment in `dockerfile.go`, the guard header). Shipping a stated circularity that one command disproves is this repo's dominant defect class — prose asserting what no test observes — and it is worth the two-line fix.

**2 — the denylist has an open near-miss tail, which is the invisible direction.** `nonContainerVariantSuffixes` is a 16-entry blocklist over an unbounded segment space, so everything unlisted accepts, and no recall-shaped assertion can see it. It is not free: a backup of a real Dockerfile *has FROM lines*, so it clears the `stageCount == 0` early return and mints a duplicate `SCOPE.Component` plus a carrier — exactly the harm the denylist exists to prevent. Blast radius is bounded (different `Name` ⇒ different `EntityID`, so noise rather than collision).

Measured escapes classifying as `dockerfile` today: `Dockerfile.bak~` and `Dockerfile.dev~` (a trailing `~` defeats the exact-lowercase lookup on entries deliberately chosen — the same shape `TestMX1100_…` has forbidden since long before this issue); `Dockerfile.rpmsave`, `.rpmnew`, `.dpkg-old`, `.dpkg-dist`; `Dockerfile.1`, `.2`, `.20260101`, `.copy`, `.new`, `.sav`, `.bkp`, `.disabled`; and `Dockerfile.gpg`, `.enc`, `.age`, `.asc`, `.sig`, `.pem`, `.crt` — binary content fed to a text extractor.

Three minor items also going back: `.dockerfile` as a bare dotfile classifies unforbidden (Go's `filepath.Ext(".dockerfile")` returns the whole name); the anti-vacuity test's doc comment says "twelve" where the floor is 16; and two forbidden rows (`Dockerfile/README`, `src/Dockerfile/notes`) are vacuous — `containerVariantLanguage` is only ever handed `path.Base(norm)`, so they grade `path.Base`, not this rule.

The engineering underneath is strong: both directions of the classifier rule graded against the emitted `ClassifyResult`, clause 3 driven in production with a nested contrast, and multiplicity and over-emission both controlled. What blocks is one factual claim about the PR's own shape and one widening whose tail is open where nothing can see it.

cajasmota:
## Coordinator verification — one ALIVE mutant, third category (distinguishable and ungraded)

The independent re-review APPROVEs `8b287ac90`: 14/14 delta mutants dead, the `sig` shadowing confirmed real **and confirmed to be the only one** (all 35 keys audited against the full extension map independently of the test), and all five guard constants re-derived from disk after the silent auto-merge. The restructure is a genuine improvement on what was reviewed the first time.

My own mutant is the one thing still outstanding, and it is **ALIVE**.

### The mutant

`containerVariantLanguage` cuts `stem+"."` and nothing else. I made it fall back to `stem+"-"`:

```go
variant, ok := strings.CutPrefix(base, stem+".")
if !ok {
    variant, ok = strings.CutPrefix(base, stem+"-")
}
```

Post-edit gate asserted (exactly one occurrence of `stem+"-"`), `go vet` clean. `./internal/classifier`, `./internal/extractors`, `./internal/extractors/dockerfile` and `./internal/extractor` are **all green** under it.

### Demonstrated, not argued

A throwaway internal probe on `detectLanguage`, under the mutant:

```
detectLanguage("Dockerfile-dev")      = "dockerfile"
detectLanguage("Containerfile-prod")  = "dockerfile"
detectLanguage("app/Dockerfile-slim") = "dockerfile"
```

with the positive control that matters: the same probe **passes** on the restored tree, so it is not an always-red assertion. Backup shasum verified before restore, probe deleted, worktree confirmed byte-identical to `8b287ac90`.

### Why it is worth three lines rather than a note

This is the third verdict category — *distinguishable and ungraded* — so the disposition is to grade it, not to record it. `grep 'Dockerfile-\|Containerfile-'` across `internal/classifier/` returns **nothing**, tests included.

Every other component of the accept predicate is now graded in both directions: the stem, the case, the dot-count, the two shape rules, and all 35 denylist entries. The **separator that gates all of them** is the one part with no forbidden row. That is the "one axis pinned, its neighbour left open" shape that has cost this series repeatedly — and it is a little sharper here, because the PR's own thesis is that over-matching is the invisible direction and the forbidden rows are the only thing that can see it.

Asked for: three rows in `TestDockerfileNearMissesStayUnclassified6854` — `Dockerfile-dev`, `Containerfile-prod`, and a nested `app/Dockerfile-slim` so the depth axis is held explicitly rather than by luck. **No production change** — the current behaviour is correct, it was simply unobserved.

Merging on the new sha once that lands. Findings A–C from the review are filed as **#6886** (the `Dockerfile.<lang>` mis-routing, the numeric rule's "never", and the incomplete pacman/ucf/vim families) and are explicitly not blockers here.

PR #6893, opened 2026-09-06
fix(#6820): nine dashboard panes matched the Electron IPC kind and rejected SCOPE.Endpoint outright — repoint all ten consumers and give the channels a home

### LINKED ISSUE #6820: Electron IPC channels leak into the HTTP path and OpenAPI dashboard panes — the dashboard keys on bare "Endpoint", a different concept from SCOPE.Endpoint

cajasmota:
fix(#6820): Electron IPC channels were the only thing the HTTP panes accepted under the name `Endpoint` — 10 consumer sites repointed at `SCOPE.Endpoint`

Closes #6820

Implements the owner's ruling of 2026-09-06: **fix the 10 consumer sites; no rename, no
stored-graph migration.** Bare `"Endpoint"` keeps its spelling as the Electron IPC kind, so no
already-indexed group is orphaned.

---

### 1a. The count is right, but the issue understates what the sites were doing

`10` is correct for sites that **select entities** on the bare spelling. Grounded from the tree at
`b6d677b0e`, there are **13 non-test lines** in `internal/dashboard` mentioning bare `"Endpoint"`;
10 of them gate entity selection and are changed here, 3 do not and are deliberately left alone
(table below).

The issue's framing — "the dashboard keys on the bare spelling" — is true but **understates the
defect by one step**. Nine of the ten sites compared the **raw** kind:

```go
e.Kind == "Endpoint" || e.Kind == "Route"
```

`types.IsHTTPEndpointKind` accepts only `http_endpoint`, `http_endpoint_definition` and
`http_endpoint_call`; `httpEndpointKind` is `"http_endpoint"`. Neither matches `Endpoint`. So the
bare clause was not one accepted kind among several — **it was the only clause that could ever
match either spelling of `Endpoint`, and it matched exactly the wrong one.** `SCOPE.Endpoint` was
rejected by every one of those nine panes. The panes did not "also" admit IPC channels; for this
pair they admitted IPC channels *instead of* HTTP endpoints. The RED test run shows this directly:
before the fix, a `SCOPE.Endpoint` fixture is absent from every pane while the IPC channel is
present in all of them.

The tenth site (`graphstate.go`) compared the SCOPE-**stripped** kind, so it genuinely accepted
both spellings; it is the one site where the fix could have silently dropped the `SCOPE.Endpoint`
side, and it has its own positive control for that.

### 1b. The 10 changed sites, read individually

| # | Site | What it is for | Verdict |
|---|---|---|---|
| 1 | `handlers_paths.go:134` | v1 Paths pane — the endpoint list | repoint |
| 2 | `handlers_paths.go:436` | v1 path **detail** (selects by `hashStr(path)`) | repoint |
| 3 | `v2_paths.go:372` | v2 Paths pane — per-backend route tree | repoint |
| 4 | `v2_paths.go:867` | v2 path **detail** | repoint |
| 5 | `v2_paths_downstream_dag.go:288` | resolves the DAG **root** for a path hash | repoint |
| 6 | `v2_paths_posture.go:151` | auth/rate-limit **posture** report for a path | repoint |
| 7 | `search_index.go:127` | prebuilt **HTTP side-index** for path search | repoint |
| 8 | `handlers_export_openapi.go:186` | the **OpenAPI export** (priority half) | repoint |
| 9 | `handlers_search.go:192` | `isHTTPEndpointEntity` — **linear** search fallback | repoint |
| 10 | `graphstate.go:1558` | `groupTopFrameworks` — group "top HTTP frameworks" badge | repoint |

**No site legitimately wanted both kinds.** Every one of the ten answers a question of the form
"is this entity an HTTP entrypoint", for which an IPC channel is a false yes. Sites 7 and 9 are the
same user-facing feature reached by two code paths (`handleSearch` picks between them on
`grp.Search`), and both are changed — they are graded by separate mutants because a fix to one
leaves the other unguarded.

### 1c. Three sites deliberately NOT changed

| Site | What it is for | Why unchanged |
|---|---|---|
| `topology_compound.go:157` (`renderableKind`) | which kinds are architecturally significant enough to draw | reads the **stripped** kind, so it accepts both spellings — and it **should**. An Electron IPC boundary is a real architectural noun. This is where the channels go. |
| `topology_compound.go:199` (`nodeTier`) | which lane a node sits in | same; IPC channels belong on the **edge** tier |
| `handlers_export_dsl.go:249` (`kindStyles`) | a hex-colour lookup for diagram export | selects **nothing**; it colours a node the renderer already decided to draw. Note it keys on the raw kind, so `SCOPE.Endpoint` falls through to the default style — a pre-existing cosmetic gap, out of scope here and deliberately not swept in. |

### 1d. Does the OpenAPI export already filter downstream? **Yes, partly — answering the question the issue left open**

Yes. `handlers_export_openapi.go:199` runs `isHTTPEndpointPath(path)`, which requires the path to
start with `/` or `http(s)://`. Only **three** of the ten sites apply it (`handlers_export_openapi`,
`handlers_paths` list, `v2_paths` list).

So for the OpenAPI export specifically, **the blast radius is narrower than an unqualified
reading would suggest**: an IPC channel leaks into the published document only when its channel
name is itself shaped like an HTTP path — `ipcMain.handle('/api/config', …)`, which is legal
Electron but not the common convention (the repo's own fixtures use `dialog:openFile` and
`log:write`). A channel named `app:quit` was already rejected downstream.

**The issue asked this question; this is the answer, not a correction to it.** Its "Not established"
section pre-registered exactly this — *"whether the OpenAPI export applies any further filtering
downstream that would already exclude these. If it does, the blast radius is smaller than stated"* —
so the finding is the answer it asked for. Nothing in the issue is overstated: it never claimed
`SCOPE.Endpoint` was accepted, and if anything it **understates** the defect, since the panes turn
out to have rejected the HTTP kind outright rather than merely also admitting IPC channels.

The narrowing is real but **local to the export**, for two reasons:

1. The narrowing does not apply to the other seven sites. The **search side-index** and the
   **linear search fallback** apply no path filter at all, so *every* IPC channel — `app:quit`
   included — was offered to users as an HTTP path. From there the four **detail** handlers select
   by `hashStr(path)` with no path filter either, so clicking that search result produced a full
   HTTP path detail page, a posture report and a downstream DAG for a channel that has no URL.
   The leak chain is end-to-end and is reproduced in the tests.
2. `groupTopFrameworks` has no path filter, and the engine stamps `framework: <language>` on every
   rule-pack source-pattern entity (`internal/engine/detector.go:450`), so an Electron app reported
   `javascript_typescript` among its top **HTTP** frameworks purely on the strength of its IPC
   channels.

I have not edited the issue body. Nothing in it needs correcting — the one place it could be read
as overstating the OpenAPI blast radius is a question it explicitly left open, and the answer is
recorded here.

### 1e. Corpus figure: **0 — and corpus-relative, exactly as predicted**

The only registered group is `monorepo-corpus` (one repo, `curantis-monorepos-main`,
`~/.config/grafel/monorepo-corpus.fleet.json`). A grep for `ipcMain|ipcRenderer|contextBridge`
across its `.js/.ts/.jsx/.tsx` (excluding `node_modules`) returns **zero files**, so the Electron
rule pack cannot have fired and there are **0 bare-`Endpoint` entities** in any indexed graph on
this machine.

This is a **corpus-relative zero, not a capability zero.** The producer is live: `electron.yaml`
lines 41/46/52 are the sole declarations of the kind (confirmed against `entkinds`' ledger, which
records `"Endpoint": "rule_namespace" // 3 sites`), and the repo ships three Electron fixtures
(`testdata/fixtures/typescript/electron_ipc.ts`, `electron_preload.ts`,
`substrate_electron/substrate_electron.ts`). Index any Electron app and the leak is immediate.

### 1f. One thing worth knowing before reviewing

**`SCOPE.Endpoint` is an admitted kind with no producer.** Being precise about this, because the
obvious shorthand is wrong: it **is** a member of the enum — `EntityKindEndpoint` is declared at
`types/kinds.go:60` and listed in `AllEntityKinds()` at `kinds.go:728`, which I verified directly.
It is therefore a kind producers are *permitted* to emit; it is simply that none does. No rule pack
declares it, and the cross-endpoint extractor emits `SCOPE.Operation` for endpoints instead
(`internal/extractors/cross/endpoint/extractor.go:21,204`).

That extractor's comment about a "14-type SCOPE allowlist" not containing a standalone `Endpoint`
is **its own prose about its own emission choice**, not a statement about `AllEntityKinds()`, and it
should not be quoted as evidence of enum membership — it disagrees with the enum. *Admitted but
unproduced* is the accurate description.

So on today's extractors the `SCOPE.Endpoint` arm is **inert**, and the load-bearing effect of this
change is the removal of the bare clause. I have implemented the ruling literally rather than
deleting the clause, because (a) that is what was decided, (b) it is what the panes *mean*, and
(c) it is forward-compatible if an emitter ever appears. But the change should not be reviewed as
"the panes now pick up SCOPE.Endpoint instead" — on the current corpus they pick up nothing extra.

**The inert half is `Endpoint` only.** The adjacent, untouched clause on nine of the same lines —
`e.Kind == "Route"`, raw, sitting next to a stripped `kind :=` — has the identical defect, and
`SCOPE.Route` **is** produced: I confirmed emitters in `internal/custom/java/vaadin_gwt.go:146,294`
(Vaadin `@Route` pages, history tokens) and `internal/custom/lua/routing.go:199,228,243,265`
(nginx/Kong/OpenResty routes). Those routes are therefore silently missing from both Paths panes,
all four detail handlers, both search paths and the OpenAPI export **today** — a live recall loss,
not an inert clause. It is outside the owner's ruling for #6820 and is being filed separately; I
have deliberately not swept it in here.

---

## Step 2 — where the IPC channels go instead

Not deleted from the UI. Two destinations, both asserted:

1. **The compound architecture topology** (`/api/v2/topology/{group}/compound`).
   `renderableKind` and `nodeTier` read the SCOPE-stripped kind and therefore keep accepting the
   bare spelling. This is deliberate and is now pinned:
   `TestElectronIPCChannels_RemainOnTheArchitectureTopology` asserts both channels appear as nodes
   **and** that they land on the `edge` tier. If someone later "finishes the job" by pointing those
   two switches at `SCOPE.Endpoint` too, that test fails — which is the point. Mutants `M12`/`M13`
   score the two halves separately, because `renderableKind` and `nodeTier` otherwise only ever
   fire together.

2. **Generic entity search.** `buildSearchIndex` appends every entity to the general index before it
   builds the HTTP side-index, so a channel stays findable by name.
   `TestElectronIPCChannels_RemainInEntitySearch` asserts this through the rendered search response
   on **both** the indexed and the linear path.

This is the recall half, and it is invisible to every other test in the file: all of those pass just
as happily if the entities stop existing entirely.

---

## Tests — all on the emitted artefact

`internal/dashboard/electron_ipc_not_http_6820_test.go`. Every assertion reads a rendered payload
(the exported OpenAPI document, a pane's JSON, the search response, the framework list) — no
internal counter is asserted anywhere.

Fixtures are two Electron IPC channels shaped exactly as the rule pack emits them (bare kind, **no**
`path` property — the channel name is the entity name; `framework=javascript_typescript`,
`pattern_type=yaml_driven` per `detector.go:450`), plus a real `http_endpoint_definition` and a
`SCOPE.Endpoint`:

- `/api/config` — a channel **named like a path**, the only shape that survives
  `isHTTPEndpointPath` and therefore the only shape that proves the *kind* check is what excludes it
  from the OpenAPI export and the two list panes.
- `app:quit` — a conventionally-named channel, which reaches the sites that apply **no** path
  filter, where the kind check is the sole guard.

**Every absence assertion is paired with a positive control in the same payload**, checked first and
with `t.Fatalf`, so an absence loop over an empty set cannot pass as a guard. For the four detail
handlers the control is stronger still: the *same handler* driven by the *same* `pathHash`
mechanism must resolve the real endpoint before the IPC assertion runs.

| Test | Pins |
|---|---|
| `TestOpenAPIExport_ExcludesElectronIPCChannels` | site 8 — the published artefact |
| `TestPathsPanes_ExcludeElectronIPCChannels` | sites 1, 3 (v1 + v2 list, separate subtests) |
| `TestSearchPaths_ExcludeElectronIPCChannels` | sites 7, 9 (indexed + linear, separate subtests) |
| `TestPathDetail_ElectronIPCChannelIsNotResolvable` | sites 2, 4, 5, 6 (four separate subtests) |
| `TestGroupTopFrameworks_ExcludesElectronIPCChannels` | site 10 |
| `TestElectronIPCChannels_RemainOnTheArchitectureTopology` | the destination + tier |
| `TestElectronIPCChannels_RemainInEntitySearch` | the destination, both search paths |

RED before the change: every one of the first five failed, and for the right reason (the IPC channel
present, the `SCOPE.Endpoint` entity absent). The two destination tests passed before **and** after
— they are the recall guard, and their job is to stay green.

### One pre-existing fixture corrected

`handlers_phase1_test.go` built two **gin HTTP routes** (`src/routes.go`, `verb`, `path`,
`framework: gin`) using `Kind: "Endpoint"` — the Electron IPC spelling — as a stand-in for the HTTP
kind. The label claimed more than the content: they are HTTP routes and are now
`string(types.EntityKindEndpoint)`. This is why those seven tests went red on the first run of the
fix; the fixture, not the fix, was wrong.

---

## Mutation score

Driver: a lane-private script (uniquely named subdirectory — the shared scratchpad has had generic
drivers overwritten mid-run). Every mutant: exact string replacement, **post-edit match count
asserted as a hard gate** (a silently unapplied edit reads as ALIVE and argues the opposite),
`go vet` (which compiles tests; `go build` does not), whole-package `go test -count=1`, restored by
`cp` from a shasum-verified backup — never `git checkout`. A green baseline is verified before the
first mutant, and the tree is confirmed pristine after the last.

**Round 1 — baseline verified GREEN before the first mutant. 16 applied, 16 DEAD, 0 ALIVE,
0 ERROR.** The tree was confirmed pristine after the run.

All-DEAD on the set you chose proves the set, not the code. **It did not hold here:** review ran a
superset and found the `SCOPE.Endpoint` arm ungraded at three sites round 1 never mutated. Round 2
below closes that and re-scores the arm at every site, and one genuinely-ALIVE mutant is triaged as
*equivalent under the current suite* with its reason. Read this round-1 table as "these 16 died",
not as "the suite is complete".

| # | Mutant | Direction it probes | Verdict |
|---|---|---|---|
| M1 | OpenAPI export -> bare `"Endpoint"` | revert one site (the priority half) | DEAD |
| M2 | `search_index.go` -> bare | revert; the **indexed** search path | DEAD |
| M3 | `handlers_search.go` -> bare | revert; the **linear** fallback, graded apart from M2 | DEAD |
| M4 | `v2_paths_posture.go` -> bare | revert **one** detail site alone | DEAD |
| M5 | `v2_paths_downstream_dag.go` -> bare | revert a **second** detail site alone | DEAD |
| M6 | `graphstate.go` -> bare | revert the framework summary | DEAD |
| M7 | OpenAPI export accepts **both** kinds | permissive — invisible to a recall-only test | DEAD |
| M8 | v1 Paths pane accepts both (2 sites) | permissive | DEAD |
| M9 | v2 Paths pane accepts both (2 sites) | permissive | DEAD |
| M10 | OpenAPI: **delete** the `SCOPE.Endpoint` arm | positive direction, at **one** site (see round 2) | DEAD |
| M11 | `graphstate`: delete the `SCOPE.Endpoint` arm | positive direction, at **one** site (see round 2) | DEAD |
| M12 | `renderableKind` drops `"Endpoint"` | the **destination** — the recall loss no pane can see | DEAD |
| M13 | `nodeTier` drops `"Endpoint"` | the destination's **tier** half, graded apart from M12 | DEAD |
| M14 | invert the OpenAPI absence check | proves the absence assertion is not a no-op | DEAD |
| M15 | invert the search-pane absence check (`app:quit`) | same, on the unfiltered channel | DEAD |
| M16 | invert the topology **presence** check | proves the destination assertion is not a no-op | DEAD |

Three pairs were split specifically because each pair would otherwise only ever fire together and
so would grade neither half: **M2/M3** (indexed vs linear search), **M4/M5** (two detail sites that
share one test), and **M12/M13** (`renderableKind` vs `nodeTier`). Likewise **M1/M7/M10** split the
OpenAPI predicate three ways — revert, widen, and delete — so the restrictive direction, the
permissive direction and the positive arm are each scored on their own.

Driver and full transcript: `mutate6820.py` / `mutants.txt` (lane-private scratchpad; not committed).
Note the driver's exact-match strings target the code as committed **before** the two explanatory
comments were added to `topology_compound.go`; the comments sit above the `case` lines, so the
matched strings themselves are unchanged and the run reproduces.

### Round 2 — a review found the `SCOPE.Endpoint` claim was prose, and it was right

Review ran a superset of the above (36 mutants) and caught a real hole. Round 1 deleted the
`SCOPE.Endpoint` arm at **two** sites and I wrote that this "grades that arm so it cannot rot into a
no-op unnoticed." That sentence asserted something no test observed at three of the ten sites:
deleting the arm was **ALIVE** at `v2_paths.go` (detail), `v2_paths_downstream_dag.go` and
`v2_paths_posture.go`.

The cause is exactly the compound-masking failure this PR claims to guard against, in my own test.
`TestPathDetail_ElectronIPCChannelIsNotResolvable`'s positive control was `realHTTPPath`, an
`http_endpoint_definition` — a kind `types.IsHTTPEndpointKind` **already** matches. So at those
three sites the two conditions only ever fired together and the control graded neither. It survived
at the other seven only incidentally: sites 1–2 via the phase1 fixture corrected above, and sites 3
and 7–10 via the new tests' `scopeEndpointPath` / `scopeframework` controls.

**Fix (10 lines):** the fixture already carried a `SCOPE.Endpoint` entity at `/api/scoped`, so each
of the four detail subtests now also drives `hashStr(scopeEndpointPath)` through the same handler
and asserts it resolves. No other clause in those predicates matches that kind, so the arm is now
observed at every site that carries it.

**Round 2 re-scored the arm at all eight sites that carry it, not just the three:**

| # | Mutant | Round 1 | Round 2 |
|---|---|---|---|
| R1 | `v2_paths.go` (list + detail): delete the arm | **ALIVE** (detail) | DEAD |
| R2 | `v2_paths_downstream_dag.go`: delete the arm | **ALIVE** | DEAD |
| R3 | `v2_paths_posture.go`: delete the arm | **ALIVE** | DEAD |
| R4 | `handlers_paths.go` (list + detail): delete the arm | not run | DEAD |
| R5 | `search_index.go`: delete the arm | not run | DEAD |
| R6 | `handlers_search.go`: delete the arm | not run | DEAD |
| R7 | `handlers_export_openapi.go`: delete the arm | DEAD (M10) | DEAD |
| R8 | `graphstate.go`: delete the arm | DEAD (M11) | DEAD |
| R9 | invert the **new** `SCOPE.Endpoint` control | n/a | DEAD |

9 applied, 9 DEAD, 0 ALIVE, baseline verified green first, tree pristine after. R9 exists because a
new positive control is itself a candidate no-op: inverting it must fail on correct code, and it does.

**The fourth ALIVE the review found needs no change and I agree with its verdict:** inverting
`if h.found(body)` in the IPC branch is *equivalent under the current suite*. Post-fix the IPC hash
returns non-200, so the test returns early at the status check above it and the branch is
unreachable while the code is correct — and it **is** reached, and fires, under every revert mutant
(M1–M6), which is where it earns its place. Classifying it as equivalent rather than deleting it is
the right call.

Round-2 driver and transcript: `mutate6820_r2.py` / `mutants_r2.txt`, same lane-private directory
and the same discipline (post-edit match count as a hard gate, `go vet`, whole-package `-count=1`,
`cp` restore from a shasum-verified backup).

---

## Packages run

- `./internal/dashboard/` — full suite, `-count=1`, green (and green under every mutant restore).
- `./internal/types/`, `./internal/entkinds/`, `./internal/graph/...` — green.
- `./cmd/grafel/` — green. Run because it digest-pins the whole graph; no entity or relationship
  kind changed here, but a diff-derived package set would have missed it.

**Not run:** the rest of the tree. The absence claim in this PR is only as wide as those packages.
In particular I did not run `./internal/engine/...`, `./internal/mcp/...` or `./internal/links/...`:
no file in them is touched, and the two remaining non-dashboard bare-`Endpoint` consumers
(`internal/engine/precision_dedup.go:75`, a dedup priority table, and
`internal/engine/ws_channel_grouping.go:445`) are engine-side and are not HTTP-surface claims.

---

## Consequence for #6776

The ambiguity that made bare `Endpoint` a *different-concept* blocker is gone: no consumer now keys
on the bare spelling for the HTTP concept. Whether #6776's ledger then goes to 0 is a separate
decision and is **not** implied by this one.

### DIFF
--- a/internal/dashboard/electron_ipc_not_http_6820_test.go
+++ b/internal/dashboard/electron_ipc_not_http_6820_test.go
@@ -1,0 +1,601 @@
+package dashboard
+
+import (
+	"encoding/json"
+	"net/http"
+	"net/http/httptest"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/cajasmota/grafel/internal/graph"
+	"github.com/cajasmota/grafel/internal/types"
+)
+
+// ---------------------------------------------------------------------------
+// #6820 — Electron IPC channels must not be presented as HTTP endpoints.
+//
+// Two entity kinds differ only by a "SCOPE." prefix and mean different things:
+//
+//   - SCOPE.Endpoint  — the HTTP entrypoint kind (types.EntityKindEndpoint).
+//   - bare "Endpoint" — declared ONLY by the Electron rule pack
+//     (internal/engine/rules/javascript_typescript/frameworks/electron.yaml
+//     lines 41/46/52 — ipcMain / ipcRenderer / contextBridge), where it names
+//     an Electron IPC channel. It is never an HTTP route and never had a URL.
+//
+// Ten dashboard sites keyed on the BARE spelling, so IPC channels rendered in
+// the HTTP panes and were written into the exported OpenAPI document — a
+// published artefact people read to enumerate HTTP attack surface.
+//
+// Note the pre-fix predicate `e.Kind == "Endpoint"` compared the RAW kind, so
+// it admitted *only* the Electron kind and never matched SCOPE.Endpoint at
+// all: the panes accepted exactly the wrong one of the pair.
+//
+// The tests below assert the EMITTED ARTEFACT (rendered pane payload, exported
+// OpenAPI document, search response, framework summary) rather than any
+// internal counter, and every absence assertion is paired with a positive
+// control in the same payload so the loop can never pass vacuously.
+// ---------------------------------------------------------------------------
+
+// ipcChannelPathLike is the Electron IPC channel whose *name* is shaped like an
+// HTTP path. This is the load-bearing fixture for the OpenAPI export and the
+// paths list: both apply isHTTPEndpointPath() downstream, which rejects a
+// channel name such as "app:quit" on its own. Only a channel named like a path
+// survives that filter, so only this shape proves the kind check is what
+// excludes it. `ipcMain.handle('/api/config', ...)` is legal Electron.
+const ipcChannelPathLike = "/api/config"
+
+// ipcChannelPlain is a conventionally-named channel. It reaches the panes that
+// do NOT filter by path shape (the search side-index and the linear search
+// fallback), where the kind check is the only thing standing in the way.
+const ipcChannelPlain = "app:quit"
+
+// realHTTPPath is the positive control: a genuine backend handler that must
+// remain in every pane after the fix.
+const realHTTPPath = "/api/users"
+
+// scopeEndpointPath is carried by a SCOPE.Endpoint entity — the kind the panes
+// are now supposed to key on. It grades the POSITIVE arm of the replacement:
+// without it, swapping `e.Kind == "Endpoint"` for a constant that nothing in
+// the fixture matches would be indistinguishable from deleting the clause.
+const scopeEndpointPath = "/api/scoped"
+
+// ipc6820Entities returns one repo's worth of entities: two Electron IPC
+// channels exactly as the electron.yaml source_patterns emit them (bare kind,
+// no "path" property — the channel name IS the entity name, and the engine
+// stamps framework=<language> and pattern_type=yaml_driven on every
+// source-pattern entity, see internal/engine/detector.go:450), one real
+// http_endpoint_definition, and one SCOPE.Endpoint.
+func ipc6820Entities() []graph.Entity {
+	return []graph.Entity{
+		graph.Entity{
+			ID:   "ipc-pathlike",
+			Name: ipcChannelPathLike,
+			Kind: "Endpoint", // bare — Electron IPC, NOT HTTP
+		}.WithProperties(map[string]string{
+			"framework":    "javascript_typescript",
+			"pattern_type": "yaml_driven",
+		}),
+		graph.Entity{
+			ID:   "ipc-plain",
+			Name: ipcChannelPlain,
+			Kind: "Endpoint", // bare — Electron IPC, NOT HTTP
+		}.WithProperties(map[string]string{
+			"framework":    "javascript_typescript",
+			"pattern_type": "yaml_driven",
+		}),
+		graph.Entity{
+			ID:   "real-http",
+			Name: "GET " + realHTTPPath,
+			Kind: "http_endpoint_definition",
+		}.WithProperties(map[string]string{
+			"method":    "GET",
+			"verb":      "GET",
+			"path":      realHTTPPath,
+			"framework": "gin",
+		}),
+		graph.Entity{
+			ID:   "scope-ep",
+			Name: "GET " + scopeEndpointPath,
+			Kind: string(types.EntityKindEndpoint), // "SCOPE.Endpoint" — HTTP
+		}.WithProperties(map[string]string{
+			"method":    "GET",
+			"verb":      "GET",
+			"path":      scopeEndpointPath,
+			"framework": "scopeframework",
+		}),
+	}
+}
+
+// ipc6820Server wires a Server holding ipc6820Entities under group "g".
+// withIndex controls whether the pre-built SearchIndex is attached, which
+// selects between the two independent search code paths (search_index.go's
+// side-index and handlers_search.go's linear fallback).
+func ipc6820Server(t *testing.T, withIndex bool) *Server {
+	t.Helper()
+	srv, err := NewServer(DefaultConfig(), newFakeStore())
+	if err != nil {
+		t.Fatalf("NewServer: %v", err)
+	}
+	grp := openAPITestGroup("g", ipc6820Entities())
+	if withIndex {
+		grp.Search = buildSearchIndex(grp)
+		if grp.Search == nil {
+			t.Fatal("buildSearchIndex returned nil; the indexed search path would not be exercised")
+		}
+	}
+	srv.graphs.mu.Lock()
+	srv.graphs.entries["g"] = &cacheEntry{group: grp, loadedAt: time.Now()}
+	srv.graphs.mu.Unlock()
+	return srv
+}
+
+// serveJSON runs h against GET url with the given path values and returns the
+// decoded body, failing the test on a non-200.
+func serveJSON(t *testing.T, h http.HandlerFunc, url string, pathValues map[string]string) map[string]any {
+	t.Helper()
+	req := httptest.NewRequest(http.MethodGet, url, nil)
+	for k, v := range pathValues {
+		req.SetPathValue(k, v)
+	}
+	w := httptest.NewRecorder()
+	h(w, req)
+	if w.Code != http.StatusOK {
+		t.Fatalf("%s: status = %d, want 200; body=%s", url, w.Code, w.Body.String())
+	}
+	var body map[string]any
+	if err := json.Unmarshal(w.Body.Bytes(), &body); err != nil {
+		t.Fatalf("%s: bad JSON: %v; body=%s", url, err, w.Body.String())
+	}
+	return body
+}
+
+// ---------------------------------------------------------------------------
+// Site 8 (priority) — handlers_export_openapi.go: the exported document
+// ---------------------------------------------------------------------------
+
+// TestOpenAPIExport_ExcludesElectronIPCChannels is the priority half of #6820.
+// The OpenAPI document is a published artefact; an IPC channel in it is a
+// false claim about the HTTP surface a service exposes.
+func TestOpenAPIExport_ExcludesElectronIPCChannels(t *testing.T) {
+	srv := ipc6820Server(t, false)
+	req := httptest.NewRequest(http.MethodGet, "/api/export/g/openapi?format=json", nil)
+	req.SetPathValue("group", "g")
+	w := httptest.NewRecorder()
+	srv.handleExportOpenAPI(w, req)
+	if w.Code != http.StatusOK {
+		t.Fatalf("status = %d, want 200; body=%s", w.Code, w.Body.String())
+	}
+	var doc struct {
+		Paths map[string]json.RawMessage `json:"paths"`
+	}
+	if err := json.Unmarshal(w.Body.Bytes(), &doc); err != nil {
+		t.Fatalf("bad JSON: %v", err)
+	}
+
+	// Positive control FIRST. Without it the absence check below could pass on
+	// an empty document — a vacuous guard that reads like a real one.
+	if _, ok := doc.Paths[realHTTPPath]; !ok {
+		t.Fatalf("exported paths %v are missing the real endpoint %q; the absence assertion "+
+			"below would be vacuous", keysOf(doc.Paths), realHTTPPath)
+	}
+	if _, ok := doc.Paths[scopeEndpointPath]; !ok {
+		t.Errorf("exported paths %v are missing the SCOPE.Endpoint entry %q; the export must "+
+			"key on SCOPE.Endpoint, not merely drop the bare kind", keysOf(doc.Paths), scopeEndpointPath)
+	}
+
+	// The forbidden direction.
+	if _, ok := doc.Paths[ipcChannelPathLike]; ok {
+		t.Errorf("exported OpenAPI paths contain %q, an Electron IPC channel (bare %q kind) — "+
+			"it is not an HTTP endpoint and never had a URL; paths=%v",
+			ipcChannelPathLike, "Endpoint", keysOf(doc.Paths))
+	}
+	if _, ok := doc.Paths[ipcChannelPlain]; ok {
+		t.Errorf("exported OpenAPI paths contain the IPC channel %q; paths=%v",
+			ipcChannelPlain, keysOf(doc.Paths))
+	}
+}
+
+func keysOf(m map[string]json.RawMessage) []string {
+	out := make([]string, 0, len(m))
+	for k := range m {
+		out = append(out, k)
+	}
+	return out
+}
+
+// ---------------------------------------------------------------------------
+// Sites 1 and 3 — handlers_paths.go / v2_paths.go: the HTTP path panes
+// ---------------------------------------------------------------------------
+
+// TestPathsPanes_ExcludeElectronIPCChannels covers both the v1 Paths pane and
+// the v2 Paths pane. Each is a separate predicate in a separate file, so each
+// gets its own subtest rather than one loop that a single fix would satisfy.
+func TestPathsPanes_ExcludeElectronIPCChannels(t *testing.T) {
+	cases := []struct {
+		name  string
+		paths func(t *testing.T, s *Server) []string
+	}{
+		{
+			name: "v1 /api/paths/{group}",
+			paths: func(t *testing.T, s *Server) []string {
+				body := serveJSON(t, s.handlePathsList, "/api/paths/g",
+					map[string]string{"group": "g"})
+				return collectStrings(asSlice(body["paths"]), "path")
+			},
+		},
+		{
+			name: "v2 /api/v2/paths/{id}",
+			paths: func(t *testing.T, s *Server) []string {
+				body := serveJSON(t, s.handleV2PathsList, "/api/v2/paths/g",
+					map[string]string{"id": "g"})
+				// The v2 payload nests routes inside a per-backend tree rather
+				// than a flat array, so gather every "path" field at any depth.
+				// Keyed on the field name, NOT a whole-body substring match:
+				// a substring assertion over the whole response would survive
+				// deleting the thing it checks.
+				return collectPathsDeep(body["data"])
+			},
+		},
+	}
+
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
+			srv := ipc6820Server(t, false)
+			paths := tc.paths(t, srv)
+
+			if !ipc6820Contains(paths, realHTTPPath) {
+				t.Fatalf("pane paths %v are missing the real endpoint %q; the absence "+
+					"assertion below would be vacuous", paths, realHTTPPath)
+			}
+			if !ipc6820Contains(paths, scopeEndpointPath) {
+				t.Errorf("pane paths %v are missing the SCOPE.Endpoint entry %q", paths, scopeEndpointPath)
+			}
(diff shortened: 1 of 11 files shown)


cajasmota:
## Coordinator verification — DEAD, via the one door all 36 scored mutants left alone

Between the author's 16 and the review's 36, every one of the ten call sites was mutated three ways, plus the topology destination and eight test-assertion inversions. All of them mutate **call sites**. So I went at the shared predicate instead: if the leak can be reopened through `types.IsHTTPEndpointKind`, then repointing ten callers at it buys less than it appears to.

### The mutant

```go
case string(HTTPEndpointKindLegacy),
     string(EntityKindHTTPEndpointDefinition),
     string(EntityKindHTTPEndpointCall),
     "Endpoint":                       // ← readmit the Electron IPC kind, centrally
     return true
```

Gate asserted (new case present exactly once), `go vet` clean.

### Result: DEAD, on four dashboard tests

```
--- FAIL: TestPathsPanes_ExcludeElectronIPCChannels/v1_/api/paths/{group}
    electron_ipc_not_http_6820_test.go:255: pane paths [/api/config /api/scoped /api/users]
      contain "/api/config", an Electron IPC channel, not an HTTP route
--- FAIL: TestSearchPaths_ExcludeElectronIPCChannels
--- FAIL: TestPathDetail_ElectronIPCChannelIsNotResolvable
--- FAIL: TestGroupTopFrameworks_ExcludesElectronIPCChannels
```

Note the pane now carries **both** `/api/config` and `/api/scoped` — the IPC channel readmitted alongside the HTTP endpoint the fix restored. So the absence assertions are load-bearing against a central widening and not only against a per-site revert. Restored by `cp` from a shasum-verified backup, worktree removed.

### The part worth recording: `./internal/types/` stayed GREEN

The predicate's **own package has no forbidden row for it.** Admitting the Electron IPC kind into `IsHTTPEndpointKind` is invisible in `internal/types` and is caught entirely by the dashboard's absence assertions, one package away.

That is not a defect in this PR — it is a fact about where the protection now lives, and it is worth knowing before someone "tidies" that switch. The kind list a predicate accepts is the kind of thing edited by people who never run the dashboard suite. A forbidden row in `internal/types` asserting that bare `"Endpoint"` is **not** an HTTP endpoint kind would cost one line and put the guard next to the thing it guards; I am not asking for it in this PR, and it is a reasonable candidate for the `Route` follow-up (**#6894**) which touches the same predicate family.

### Standing

The review's three ALIVE mutants at the detail sites are the blocking item and are with the author. This mutant is orthogonal to them — it fails at every site including those three, because it widens the predicate they all share rather than the arm each site carries.

Issue #6866, opened 2026-09-05
No local command reproduced the coverage-docs gate — and git add -N had blinded the gate's own deletion direction since #6354


cajasmota:
The `coverage-docs` gate is five steps, and **no single local command reproduces it**. Two of the five are only reachable by replaying the workflow by hand, so the normal way to discover a failure is a red CI run.

## Measured, not inferred

The bicep arm of #6852 (PR #6864) burned **two CI cycles** on this. Reproducing the exact failing tree locally and running every `./tools/coverage` subcommand against it:

| local command | on the tree CI was red on |
|---|---|
| `go run ./tools/coverage validate` | exit 0, **0 errors** |
| `go run ./tools/coverage backfill --check` | exit 0 |
| `go run ./tools/coverage fmt --check` | exit 0 |

All three green; CI red.

## Why — it is structural, not a bug

`.github/workflows/coverage-docs.yml` runs five steps. Three are subcommands (`validate`,
`backfill --check`, `fmt --check`). The other two are not:

```yaml
- run: go run ./tools/coverage gen
- run: |
    git add -N docs/coverage/
    if ! git diff --exit-code docs/coverage/; then
      echo "::error::docs/coverage/ is out of sync with registry.json …"
```

The docs-sync check is `gen` **plus a comparison against git state**. No `./tools/coverage`
subcommand performs that, and none could — it is a property of the working tree, not of the
registry. (The `git add -N` is itself a #6354 fix so an untracked new page is not invisible.)

So `validate` catches the stale-citation half and is **blind** to the docs-sync half, and nothing
tells you that.

## Why it is worth fixing now rather than later

`coverage-docs.yml` subscribes to `internal/extractors/**` (deliberate, #6332 — the language matrix
is derived from those directories). So **any extractor edit that shifts a line number can red this
workflow**, and #6852 has **eleven more language arms to go**, each of which adds a carrier call and
pushes every declaration below it down.

The bicep arm shifted declarations by exactly 16 lines and hit both halves: two stale citations,
then the regeneration. Expect each remaining arm to need the same two steps, and to cost the same
two round trips until this is fixed.

## Suggested shape

Add a `coverage check` subcommand that runs the whole gate — `validate`, `backfill --check`,
`fmt --check`, `gen`, and the `git add -N` + `git diff --exit-code docs/coverage/` comparison — and
make `coverage-docs.yml` call **that one command** rather than listing five steps. Two properties
matter:

[...]

PR #6897, opened 2026-09-06
fix(#6852): clojure's require IMPORTS anchored on a path only the ns form can ever name — arm 8 of twelve, and the arm where the placement conjunct turned out entailed

cajasmota:
# fix(#6852): arm 8 — clojure anchored IMPORTS on a path nothing was ever named

`buildImportEntities` (`internal/extractors/clojure/clojure.go:630` on this branch; `:590` on `main`,
before the placement comment shifted it) stamps `FromID: filePath` on the
`IMPORTS` edge of every `(:require ...)` / `(:use ...)` / `(:import ...)` entry.
`internal/resolve/refs.go` has no path→entity index, so a path-valued `FromID` resolves **iff** some
emitted node carries that exact string as its `Name`. Nothing clojure emits does, so the raw path
reached the graph as the edge's FROM end.

One conditional `extractor.PrependFileCarrier` at the end of `extractClojure`; clojure's line is
deleted from `knownMissingCarrier6847`.

Refs #6852

---

### 1. How many anchored sites, and do they share one anchor string?

**ONE site, N edges, one anchor string** — fsharp's / lua's / astro's multiplicity shape.

`buildImportEntities` is the only bare `FromID: <path>` in the whole package. Everything else leaves
`FromID` **empty** on purpose: the namespace record's `CONTAINS`, `collectCalls`' `CALLS`, and every
`IMPLEMENTS`/`EXTENDS` edge `hierarchy.go` scans — `hierarchy.go:359-361` states why (a non-empty
non-hex `FromID` would be rewritten away from the owning component).

One import stub is emitted per deduped `(module, localName)` pair, so an `ns` form with six entries
across `:require`/`:use`/`:import` anchors six edges on one string and must still gain **exactly one**
carrier. Driven by `TestClojure_OneCarrierPerFileNotPerImport_6852` at both depths.

The brief's grounding (`FromID: filePath`, the only non-test site) is **confirmed** — at `clojure.go:590`
on `main`, `:630` on this branch after the placement comment shifted it. The ledger row carried no
`file:line` and now has one.

### 2. Does clojure emit anything named after the file or its basename?

**No — nothing, under any condition. BOTH depths dangled.** html's / fsharp's / lua's / astro's shape;
not terraform's root-by-accident, not shell's three-way, not dockerfile's `BASENAME(path)`.

Clojure names its records after the namespace symbol, the `defn`/`defmacro` symbol, the
`defrecord`/`deftype` symbol, or — for an import stub — `topSegment(module)`, the module cut at its
**first** dot. There is no file component of any kind and no basename anywhere, so there is no
condition axis of the shell/terraform sort.

But the brief asked the sharper question — **can any emitted record equal the path string?** — and the
answer is yes, by exactly one route, at exactly one depth. The table has four cells and three are
driven in one test:

| | nested `src/app/core.clj` | root `core.clj` |
|---|---|---|
| ns name ≠ path | **dangled** (pre-fix FAIL) | **dangled** (pre-fix FAIL) |
| ns name *written* as the path | **dangled** — nsRE admits no `/`, so `(ns src/app/core.clj)` captures `src` and nothing is named the path (pre-fix FAIL) | **already resolved** — `(ns core.clj)` in `core.clj` IS the path; clause 3 correctly mints nothing |

The bottom-left cell is **driven, not asserted** — that is what makes "clause 3 is a root-depth
phenomenon for clojure" a measurement rather than an argument (`nested_ns_spelled_like_path` in
`TestClojure_ImportsFromEndResolves_6852`).

**Clause 3 has TWO routes for clojure, and the first version of this PR claimed one.** Review
falsified the character-set closure with a real-world path; the correction is recorded here rather
than quietly edited, and it *added* grading.

- **The namespace route, root depth only.** fsharp's (#6880): `nsRE` captures a dotted `[\w\-\.]+`,
  so a root `core.clj` whose header reads `(ns core.clj)` emits a `SCOPE.Component` named exactly the
  path — and `graph.EntityID` does not hash `Subtype` (#6369 / PR #6480), so a carrier there would
  land a second node under the namespace record's id. `nsRE` admits no `/`, closing this route at
  nested depth — the driven cell in the table above.
- **The import-stub route, ANY depth, for a path whose first byte is a dot.** lua's and shell's route
  (#6885, #6882) after all. `topSegment` (`clojure.go:648`) guards with **`dot > 0`**, so a module
  beginning with a dot comes back **verbatim** — slashes and extension included — and
  `requireVecRE`'s class is `[\w\-\./]+`, which admits both. Measured with the production
  extractor:

| path | stub name | carriers |
|---|---|---|
| `.core.clj` + self-require | `.core.clj` | **0** — clause 3 |
| `.clj-kondo/hooks/foo.clj` + self-require | `.clj-kondo/hooks/foo.clj` | **0** — clause 3 |
| `src/app/core.clj` + self-require (control) | `src/app/core` | **1** |

  `classifier.Classify` routes both dot-leading paths `Skip:false, Tier:1`, and
  `.clj-kondo/hooks/*.clj` is a ubiquitous real Clojure repo layout — not a contrivance.

**Three shipped statements were false**, in the test header and repeated here: that the stub name
"never contains a dot and can never equal a routed path"; that clause 3 is reachable "at root depth
only, and by exactly one route"; and that this is "where clojure differs from lua and shell" — under a
dot-leading path it does exactly what they do.

**There is no behavioural defect.** Clause 3 declines, the `IMPORTS` edge resolves onto the path-named
stub, nothing is fabricated. What was wrong was a **stated proof of unreachability** — the same shape
`hierarchy.go`'s own `:gen-class` argument got wrong (it now carries a 40k-input panic proof) and the
astro arm got wrong about `componentNameFromPath` two arms ago. The repair is a **driven cell, not a
better argument**: `TestClojure_SelfRequiringStubUnderADotLeadingPathGetsNoSecondCarrier_6852` drives
all three rows above, with a classifier premise row so "production-reachable" is a measurement about
the classifier and not an inference from the extractor.

**The correction bought real grading, which is the part worth keeping.** Mutating `topSegment`'s guard
to `dot >= 0` is **DEAD** — and killed by the new test *only*; nothing in the package observed that
guard before. (`dot >= 0` is also not a fix to make: it mints an empty-named stub.) Deleting clause 3
now fires on **both** routes at both depths rather than one. And a full revert of the carrier reddens
exactly the control row of the new test, so the control is a premise rather than decoration.

**What survives, and is what the tests drive:**

- `defnRE` / `defmacroRE` / `deftypeRE` capture `[\w\-\?!\*'+]+` — no `.`, no `/`. Every path the
  classifier routes to clojure ends `.clj`/`.cljs`/`.cljc`/`.edn` (`classifier.go:406-409`) and so
  contains a dot. No operation or class record can equal any routed path, at any depth. **This holds
  unconditionally.**
- the stub route is closed for every module **not beginning with `.`** — `topSegment` then cuts at a
  dot at index > 0 and the result contains none.

### 3. Coverage-gate cost — **zero**, and for a fourth distinct reason

`docs/coverage/registry.json` cites `internal/extractors/clojure` in three records
(lines 9896, 9903, 9910) and **every citation is a bare path** — no `file.go:NNN` anywhere in the
clojure records — so the 24-line shift the carrier comment introduces invalidates nothing.

`go run ./tools/coverage check` (#6866): **exit 0**, all five steps, no doc regeneration diff
(`git status` clean after the run). bicep's 2 and terraform's 3 stale citations still do not predict
an arm.

### 4. `mustScan`

Row added at `internal/extractor/carrier_caller_set_6861_test.go:207` (the list, now 13 entries):

```go
{"internal/extractors/clojure/clojure.go", []string{prependFileCarrier}},
```

Scored by **corrupting** it (`clojure.go` → `hierarchy.go`): **DEAD**, with the intended diagnostic —
*"internal/extractors/clojure/hierarchy.go was scanned but the body collected for it does not contain
"PrependFileCarrier" — the walk passed the right file NAME with the wrong content"*.
**Deleting** the row is **ALIVE**, as #6887 records: the roster is redundancy above `minAnchors = 6`
and 13 > 6. Out of scope, stated rather than worked around.

**`nonTaggingCallers` gets NO entry, and the negative is scored.** clojure calls
`extractor.TagEntitiesLanguage(out, "clojure")` at `clojure.go:136`, so *adding* an entry must be
REJECTED — and it is: **DEAD**, *"internal/extractors/clojure is listed as a NON-TAGGING caller but
its package does call TagEntitiesLanguage … an empty lang is no longer observable there"*.

Except that it **is** observable, by the provenance route — see the `lang → ""` mutant below.

---

### A placement conjunct that reads load-bearing and is ENTAILED

The brief pointed at lua's #6885 hazard and asked whether clojure keeps state keyed by index into the
entity slice. **It does.** `extractClojure` builds `comps []compOffset{idx, name}` where `idx` is an
index into `entities`, and step 3b dereferences it — `rec := &entities[cp.idx]` (`clojure.go:283`) —
to attach the `extend-type` / `extend-protocol` edges `hierarchy.go` scanned by **name**. An insertion
at position 0 before that step re-homes every one of those edges onto the record *before* its owner.
Silent, not fatal: the graph is confidently wrong.

The first draft of the placement comment said so, in exactly lua's words. **It was measured and it is
not true of this call.** Moving the carrier call above step 3b, or above step 4, is DEAD — but on the
*emptiness of the result*, never on ordering. `PrependFileCarrier` judges anchoring over the slice it
is handed, and `importEntities` — the only records carrying a path-anchored `FromID` — do not enter
`entities` until the head-prepend at `clojure.go:316`, which is **after** both steps. So clause 2
rejects at every earlier point, and the index hazard is **entailed by the clause-2 requirement, not an
independent constraint on this call**.

The hazard is nonetheless real, and it is graded — by the only shape that can reach it, an
**unconditional** carrier placed before step 3b:

| | shipped | under the mutant |
|---|---|---|
| `Triangle` | owns `IMPLEMENTS -> Shape` | keeps nothing |
| `Hexagon` | owns `IMPLEMENTS -> Sized` | keeps nothing |
| the carrier (index 0) | owns nothing | **owns `IMPLEMENTS -> Shape`** |
| `Triangle` | — | **owns `IMPLEMENTS -> Sized`** |

Four wrong-owner rows, at both depths. The **same unconditional carrier placed after step 3b produces
ZERO** wrong-owner rows — which is what separates the placement half from the conditional half, and is
why the pair is reported rather than the single verdict.

Pinned by `TestClojure_CarrierPlacementDoesNotShiftTheExtendPass_6852`, whose premise row asserts a
carrier really is emitted (a full revert makes that row go fatal, so the test is not vacuous).
`opOffset.idx` has **zero** consumers — step 4 reads only `op.name` — so it is not a third reason, and
no test is written for something unobservable. Both index-valued fields were enumerated; there are
exactly two, with one and zero consumers respectively.

**The generalisable half:** lua's lesson was *a conjunct that is load-bearing only in combination
reads identically to one that stands alone*. This is its neighbour — **a conjunct that is entailed by
another reads identically to one that is independent**, and the way to tell is the same: move each
part separately and read *which assertion* fires, not merely whether the suite goes red. Both
placement mutants here are DEAD, and both are DEAD for a reason that has nothing to do with what the
comment claimed.

### An ALIVE mutant graded rather than recorded

astro (#6891) killed the permissive widening of clause 3 —
`if records[i].Name == path && len(records[i].Relationships) > 0` — but with an astro fixture. Against
**clojure's package alone it was ALIVE**, because the root clause-3 fixture declares a `defn` and its
namespace record therefore owns a `CONTAINS`. Verdict: *distinguishable and ungraded*, so it was
graded, not recorded. The separating input is the ordinary re-export/facade namespace — a root file
whose `ns` form has a `:require` and **no** top-level declaration, whose namespace record owns zero
relationships. New subtest `root_namespace_owning_no_edges`; mutant now **DEAD**.

### `file_carrier.go`'s MEASURED paragraph is TRUE of clojure — and by the provenance route

`lang → ""` is **DEAD**, and not on the `Language` field: clojure tags, and the carrier is prepended
before the tagging pair, so `TagEntitiesLanguage` fills the empty token. It is killed by
`Properties["language"]` — `extractor.FileEntity` never sets that key and `TagEntitiesLanguage` stamps
it **only on the fill path**, while every other clojure record sets `Language` explicitly. That is
fsharp's / shell's provenance route, and the paragraph already names it. `lang → "clj"` is killed by
the `Language` row instead. **Two rows, two mutants, scored apart** — the second row is the premise
for reading the first as evidence, and `TestClojure_CarrierShape_6852` asserts both.

No falsification this arm; a third data point for a paragraph that four consecutive arms falsified.

**`file_carrier.go` is not edited, deliberately.** Its clause-3 paragraph names hcl, fsharp and shell,
and clojure is a fourth production caller of that clause by a route none of them uses (the namespace
name, not a basename component and not a self-referencing stub). Adding it would grow exactly the
roster #6861 removed from that file, and lua and astro — both clause-3 production callers — did not
add themselves either. The route is recorded in `clojure/file_carrier_6852_test.go`'s header instead,
where it is next to the fixture that drives it.

---

## Both directions

**Resolution** (`TestClojure_ImportsFromEndResolves_6852`): three cells, all FAIL before the carrier,
each with a premise row asserting exactly one record is named the path and that it is the carrier — so
the measurement cannot pass because something unrelated happened to be named the path.

**Over-emission** — each control matched to a **distinct return path**:

| control | return path |
|---|---|
| `TestClojure_EmptyFileGetsNoCarrier_6852` | `Extract`'s `len(file.Content) == 0` |
| `TestClojure_EmptyPathGetsNoCarrier_6852` | `FileCarrierFor` **clause 1** (`path == ""`) |
| `TestClojure_NoCarrierWithoutAnImport_6852` | `FileCarrierFor` **clause 2** (`!anchored`) |
| `TestClojure_NoNsFormGetsNoCarrier_6852` | clause 2 again, by a route the row above cannot reach |
| `TestClojure_NamespaceNamedLikeThePathGetsNoSecondCarrier_6852` | `FileCarrierFor` **clause 3** |

Two of those deserve a note.

**Clause 1 is doing real work for clojure, not shadowing clause 2.** `buildImportEntities` stamps
`filePath` verbatim, so with an **empty** path the `IMPORTS` edge's `FromID` is *also* empty and
clause 2's `FromID == path` test is trivially **satisfied** — only the empty-path guard rejects. That
is precisely the case `file_carrier.go`'s clause-1 paragraph describes, driven here through a
production extractor rather than a unit fixture. The premise row asserts the anchoring *would* pass,
so the test cannot go green because nothing anchored.

**The no-`ns`-form control is a different mutant-killer from the no-import one.** `collectImports` is
fed `findNsForm`'s body, so a file with **no** `(ns ...)` form collects no imports however many
`require` vectors its text contains. The fixture therefore carries a bare
`(:require [clojure.string :as str])` at top level — text that looks exactly like an import and
produces none. A carrier keyed on the source **text** rather than on the emitted relationships dies
here and nowhere else.

Both over-emission tests also carry forbidden rows (`Subtype == "file"` anywhere; any record named the
path), so a carrier smuggled in under a different Kind or Subtype is caught too.

---

## Mutation score — 20 mutants, 19 DEAD, 1 ALIVE (known, filed)

Every mutant: exact edit with the match count asserted as a **hard gate before and after** the write
(a silently-unapplied edit reads as ALIVE and argues the opposite), `go vet` clean, whole-package
`-count=1`, restored by `cp` from a shasum-verified backup, `git status` confirmed clean between runs.

| # | mutant | scope | verdict |
|---|---|---|---|
| 1 | carrier made **unconditional** | clojure | DEAD — 5 tests |
| 2 | `lang → ""` | clojure | DEAD — `Properties["language"]` provenance row |
| 3 | `lang → "clj"` | clojure | DEAD — `Language` row |
| 4 | `filePath → basename(filePath)` | clojure | DEAD — 5 tests, nested only |
| 5 | conditional carrier moved **before step 3b** | clojure | DEAD — on emptiness, **not** ordering |
| 6 | conditional carrier moved **before step 4** | clojure | DEAD — same reason |
| 7 | **unconditional** carrier **before step 3b** | clojure | DEAD — 4 wrong-owner rows |
| 8 | **unconditional** carrier **before step 4** (control) | clojure | DEAD, **0** wrong-owner rows |
| 9 | `PrependFileCarrier` → `append(records, carrier)` | clojure | DEAD — #577 index-0 row |
| 10 | `FileCarrierFor` clause 3 deleted | clojure | DEAD — root subtest |
| 11 | clause 3 widened `&& len(Relationships) > 0` | clojure | **ALIVE → graded → DEAD** |
| 12 | clojure re-added to `knownMissingCarrier6847` | extractors | DEAD |
| 13 | clojure dropped from `exercisedLanguages6847` | extractors | DEAD |
| 14 | clojure dropped from `anchoringLanguages6847` | extractors | DEAD |
| 15 | `mustScan` row **corrupted** | extractor | DEAD |
| 16 | `mustScan` row **deleted** | extractor | **ALIVE** — #6887, by design |
| 17 | `nonTaggingCallers` gains a clojure entry | extractor | DEAD — correctly rejected |
| 18 | `topSegment`'s guard `dot > 0` → `dot >= 0` | clojure | DEAD — **new test only** |
| 19 | clause 3 deleted, re-scored after the correction | clojure | DEAD — now **both** routes, both depths |
| 20 | carrier call fully reverted (non-vacuity control) | clojure | DEAD — reddens exactly the new test's control row |

Mutants 5–8 are the compound scored part by part. Mutants 2 and 3 are the `lang` argument scored part
by part. Mutants 13 and 14 are why **no** set-line edit was needed: clojure was already in both
`exercisedLanguages6847` and `anchoringLanguages6847` (the corpus has three `.clj` files and they do
anchor), and dropping it from either is red — so both memberships are graded and the measurement says
they belong where they already were.

Mutant 16's verdict is *alive by design, filed*: the roster is redundancy above `minAnchors`, and
#6887 records the bar for a real fix (deleting an **arbitrary** row must go red).

---

## Package set

`./cmd/grafel/` ✅ · `./internal/extractors/clojure/` ✅ · `./internal/extractors/` ✅ ·
`./internal/extractor/` ✅ · `./internal/resolve/` ✅ · `./internal/quality/...` ✅ ·
`go vet` clean · `gofmt -l` empty · `go run ./tools/coverage check` exit 0.

**`cmd/grafel` does not grade this change, plainly.** `writeSpanFixture6118`
(`cmd/grafel/custom_extractor_spans_6118_test.go:32`) has **no `.clj`/`.cljs` member**. Its 18 members
are cs (7), py (6), java, ts, js, **go** (`svc/handler.go`, line 240) and **rs** (`rs/routes.rs`, line
251) — the `files` map runs to line 260, and both trailing entries are inside it; review flagged go/rs
as absent, and re-reading the map by line number is what settles it. It is run because an extractor
change makes it mandatory (its digest pins the whole graph), not because it covers anything here. And
per #6862 `entityTupleKey` does not hash `Language`, so even with a member it could not catch mutants
2 or 3.

**`clojure-protocols-mini` (#6370) DOES move, and the first version of this section said it did not.**
`expected.json` is unchanged — its rows are `must_exist` plus forbidden, with no count and no constant
to bump — but `internal/quality/golden/baseline.json` records `entity_extracted_total` and
`relationship_extracted_total` as **ceilings gated in the growth direction** (#6488 arm D), and both
moved: 20 → 22 and 47 → 49.

**The gap that let it through, stated so the next arm does not repeat it:** that ratchet is
**CI-only**. It runs through `scripts/quality/run.sh --ratchet` (or the `grafel quality` CLI), not
through `go test ./internal/quality/...`, so a green local `internal/quality` package says **nothing**
about it. "The quality suite is green" and "the extraction-quality gate is green" are two different
facts. Reproduced locally before touching anything — `scripts/quality/run.sh --runs 1 --ratchet` fails
on exactly the two rows CI reported.

### Both deltas enumerated by member, not derived from counts

Dumped the fixture's full entity and relationship lists through the production indexer
(`newTestIndexer` → `idx.Run`, the same path `daemonRecallFunc` uses) with the carrier and with the
call reverted, and diffed. The delta is four lines and nothing else:

**+2 entities — exactly the two carriers, nothing drifted in alongside:**

```
+ ENT SCOPE.Component/file/extend.clj@extend.clj
+ ENT SCOPE.Component/file/shapes.clj@shapes.clj
```

`protocols.clj` has no `:require` and correctly gains none — the conditional's over-emission property
demonstrated on a real corpus fixture rather than a unit one.

**+2 relationships — NOT extractor edges.** `FileEntity` mints an entity with **zero** relationships,
and the extractor's own output is byte-for-byte unchanged: the indexer reports `pass1_rels=28` both
before and after, and a direct dump of `Extract`'s records over the three files gives 28 relationships
either way. The two new edges are minted by **graph assembly**, one per new entity:

```
+ REL CONTAINS | Module//clojure-protocols-mini -> SCOPE.Component/file/extend.clj
+ REL CONTAINS | Module//clojure-protocols-mini -> SCOPE.Component/file/shapes.clj
```

`internal/module/aggregate.go` emits one synthetic `Module → entity` CONTAINS edge per entity
(`KindModule` / `KindContains`, and `module_gds.go:337` calls them "Module→entity scaffolding"), so on
this fixture the relationship total moves **1:1 with the entity total** by construction. Both are
legitimate; neither is a new extractor edge, a survived dedup, or a re-homed owner.

**The rest of the diff is the fix working, and it is worth reading:** the six `IMPORTS` edges from
`shapes.clj` and `extend.clj` are the same six edges, same count, same targets — their FROM end simply
stops being a raw path:

```
- REL IMPORTS | UNRESOLVED<extend.clj> -> SCOPE.Component/namespace/app.protocols
+ REL IMPORTS | SCOPE.Component/file/extend.clj -> SCOPE.Component/namespace/app.protocols
```

…and five more of the same shape. That is #6847's defect and its repair, visible end-to-end on a real
corpus fixture rather than a unit one.

### The baseline edit is SURGICAL, and `--update-baseline` would have swept in three unrelated rows

`scripts/quality/run.sh --runs 1 --update-baseline` re-records the **whole file**, and on this tree it
moved four fixtures, not one:

| fixture | baseline | this tree | direction |
|---|---|---|---|
| `clojure-protocols-mini` | 20 / 47 | **22 / 49** | grew — **this arm** |
| `python-django-mini` | – / 255 | – / **251** | shrank |
| `python-fastapi-mini` | 28 / 57 | **27 / 54** | shrank |
| `python-rq-mini` | 22 / 43 | **20 / 39** | shrank |

The three python rows are **not this arm's**: the only production file this PR touches is
`internal/extractors/clojure/clojure.go`, which cannot reach python extraction. They are pre-existing
drift on `main` — `baseline.json` records `measured_on: dab49ec0e` and this branch sits on
`9706a5cde` — and they went unnoticed because **the ratchet gates growth only**, so a shrink passes in
silence. Reproduced identically across two independent runs, so not run-to-run noise.

Recording them here would tighten three ceilings on a measurement this arm did not investigate and
would silently absorb whatever caused the shrink. So the committed edit is **two numbers**, matching
the #6815 precedent (`0ce393bc3` edited only its affected fixtures' rows and no metadata):

```
-      "entity_extracted_total": 20,
-      "relationship_extracted_total": 47
+      "entity_extracted_total": 22,
+      "relationship_extracted_total": 49
```

`measured_at` / `measured_on` deliberately untouched, since the file as committed is not a full
re-record. **Verified rather than assumed:** `scripts/quality/run.sh --runs 1 --ratchet` with only that
edit reports *"quality ratchet OK — 31 gated fixtures held their baseline (22 at 100% must-have
recall)"*, exit 0 — the growth-only gate accepts the three shrinks as it always has.
`./internal/quality/...` green, `expected.json` untouched.

**Worth filing separately** (not done here, to keep this PR one subject): a shrink in
`*_extracted_total` is invisible to the ratchet by construction, so extraction can lose four
relationships on a gated fixture between two baseline re-records and nothing says so. That is the
mirror of what #6488 arm D added the growth check for.

### DIFF
--- a/internal/extractor/carrier_caller_set_6861_test.go
+++ b/internal/extractor/carrier_caller_set_6861_test.go
@@ -204,6 +204,7 @@ func TestCarrierCallerSetIsGradedFromSource_6861(t *testing.T) {
 		{"internal/extractors/dockerfile/dockerfile.go", []string{prependFileCarrier}},
 		{"internal/extractors/lua/lua.go", []string{prependFileCarrier}},
 		{"internal/extractors/astro/extractor.go", []string{prependFileCarrier}},
+		{"internal/extractors/clojure/clojure.go", []string{prependFileCarrier}},
 	}
 
 	nonTest := 0
--- a/internal/extractors/clojure/clojure.go
+++ b/internal/extractors/clojure/clojure.go
@@ -318,7 +318,28 @@ func extractClojure(src, filePath string) []types.EntityRecord {
 		entities = append(importEntities, entities...)
 	}
 
-	return entities
+	// 5. File carrier (#6852). buildImportEntities stamps `FromID: filePath` on
+	// every IMPORTS edge, and internal/resolve/refs.go has no path→entity index:
+	// a path-valued FromID resolves iff some emitted node carries that exact
+	// string as its Name. Nothing clojure emits does — the namespace record is
+	// named the ns symbol, operations and classes are named their declaration
+	// symbol, and an import stub is topSegment(module) — so without this the
+	// raw path reached the graph as the edge's FROM end.
+	//
+	// CONDITIONAL, deliberately: extractor.PrependFileCarrier emits nothing when
+	// no relationship is anchored on the path (a namespace with no :require) and
+	// nothing when some record is ALREADY named the path (a root core.clj whose
+	// header reads `(ns core.clj)` — nsRE captures a dotted `[\w\-\.]+`). An
+	// unconditional carrier would mint one bare orphan node per .clj file across
+	// a whole repo, which no recall-shaped assertion can see (#6815, #6518).
+	//
+	// PLACEMENT IS LOAD-BEARING, and for ONE reason that is graded rather than
+	// argued: compOffset.idx (step 3b, `&entities[cp.idx]`) is an INDEX into
+	// this slice, so an insertion at position 0 before that step re-homes every
+	// extend-type / extend-protocol edge onto the record before its owner.
+	// Pinned by TestClojure_CarrierPlacementDoesNotShiftTheExtendPass_6852.
+	// opOffset.idx has no consumer at all, so it is not a second reason.
+	return extractor.PrependFileCarrier(filePath, "clojure", entities)
 }
 
 // importSpec describes one resolved import entry.
--- a/internal/extractors/clojure/clojure.go
+++ b/internal/extractors/clojure/clojure.go
@@ -333,12 +333,31 @@ func extractClojure(src, filePath string) []types.EntityRecord {
 	// unconditional carrier would mint one bare orphan node per .clj file across
 	// a whole repo, which no recall-shaped assertion can see (#6815, #6518).
 	//
-	// PLACEMENT IS LOAD-BEARING, and for ONE reason that is graded rather than
-	// argued: compOffset.idx (step 3b, `&entities[cp.idx]`) is an INDEX into
-	// this slice, so an insertion at position 0 before that step re-homes every
-	// extend-type / extend-protocol edge onto the record before its owner.
-	// Pinned by TestClojure_CarrierPlacementDoesNotShiftTheExtendPass_6852.
-	// opOffset.idx has no consumer at all, so it is not a second reason.
+	// PLACEMENT. The call goes LAST, and the reason is CLAUSE 2 rather than the
+	// index hazard below — the two read alike in prose and were measured apart.
+	// PrependFileCarrier judges anchoring over the slice it is HANDED, and
+	// importEntities (the only records carrying a path-anchored FromID) do not
+	// enter `entities` until the head-prepend just above. Move this call any
+	// earlier and the carrier is not misplaced, it is never minted at all: the
+	// "carrier before step 3b" and "carrier before step 4" mutants both die on
+	// the EMPTINESS of the result, never reaching the ordering question.
+	//
+	// THE INDEX HAZARD IS REAL BUT ENTAILED, and which of those it is is the
+	// point. compOffset.idx (step 3b, `&entities[cp.idx]`) is an INDEX into this
+	// slice, so an insertion at position 0 before that step re-homes every
+	// extend-type / extend-protocol edge onto the record BEFORE its owner —
+	// lua's #6885 hazard, present here too, silent rather than fatal. It cannot
+	// be reached by moving THIS call anywhere, because clause 2 rejects at every
+	// earlier point; it is reachable only by an UNCONDITIONAL carrier placed
+	// there, which is how it is scored. Pinned by
+	// TestClojure_CarrierPlacementDoesNotShiftTheExtendPass_6852: under that
+	// mutant the carrier takes Triangle's IMPLEMENTS -> Shape and Hexagon's
+	// IMPLEMENTS -> Sized lands on Triangle, at both depths. The SAME
+	// unconditional carrier placed after step 3b produces ZERO wrong-owner rows,
+	// which is what separates the placement half from the conditional half.
+	//
+	// opOffset.idx has ZERO consumers — step 4 reads only op.name — so it is not
+	// a third reason and no test is written for something unobservable.
 	return extractor.PrependFileCarrier(filePath, "clojure", entities)
 }
 
--- a/internal/extractors/clojure/file_carrier_6852_test.go
+++ b/internal/extractors/clojure/file_carrier_6852_test.go
@@ -487,6 +487,43 @@ func TestClojure_NamespaceNamedLikeThePathGetsNoSecondCarrier_6852(t *testing.T)
 				"the path as its Name, got %d", n)
 		}
 	})
+	// The path-named namespace record with NO relationships of its own. This is
+	// the PERMISSIVE WIDENING control, and it is a separate cell rather than a
+	// variation of the row above: the plausible maintainer edit to clause 3 is
+	//
+	//	if records[i].Name == path && len(records[i].Relationships) > 0 {
+	//
+	// on the reasoning that "a path-named record carrying no edges is not really
+	// the file's node, so minting a proper carrier beside it is harmless". astro
+	// (#6891) scored that mutant and killed it — but only with an astro fixture.
+	// Against clojure's package alone it was ALIVE, because the root fixture
+	// above declares a defn and its namespace record therefore owns a CONTAINS.
+	// A root file whose ns form has a :require and NO top-level declaration is
+	// the ordinary shape that separates them, and it is production-reachable —
+	// a re-export/facade namespace is exactly that. Distinguishable and
+	// ungraded, so it is graded here rather than recorded.
+	t.Run("root_namespace_owning_no_edges", func(t *testing.T) {
+		const path = "core.clj"
+		src := "(ns " + path + "\n  (:require [clojure.string :as str]))\n"
+		recs := runClj6852(t, src, path)
+		if n := len(cljPathAnchored6852(recs, path)); n == 0 {
+			t.Fatal("premise: the fixture must still anchor an IMPORTS on the path")
+		}
+		named := cljNamedExactly6852(recs, path)
+		if len(named) != 1 {
+			t.Fatalf("exactly 1 record may be named %q, got %d", path, len(named))
+		}
+		if n := len(named[0].Relationships); n != 0 {
+			t.Fatalf("premise: the path-named namespace record must own ZERO relationships "+
+				"here, got %d — otherwise this cell is the same as the one above and the "+
+				"permissive widening of clause 3 is not under test", n)
+		}
+		if n := len(cljCarriers6852(recs, path)); n != 0 {
+			t.Errorf("no file carrier may be minted beside a path-named record that owns no "+
+				"edges, got %d — graph.EntityID does not hash Subtype (#6369/#6480), so the "+
+				"two land under one id", n)
+		}
+	})
 	t.Run("nested_contrast", func(t *testing.T) {
 		const path = "src/app/core.clj"
 		src := "(ns " + path + "\n  (:require [clojure.string :as str]))\n\n(defn render [p] (str/join \",\" p))\n"
@@ -556,9 +593,11 @@ func TestClojure_CarrierShape_6852(t *testing.T) {
 	}
 }
 
-// TestClojure_CarrierPlacementDoesNotShiftTheExtendPass_6852 pins the ONE
-// placement conjunct in this package that is load-bearing, and it is graded on
-// its own rather than left inside a bulleted comment.
+// TestClojure_CarrierPlacementDoesNotShiftTheExtendPass_6852 pins the index
+// hazard the carrier's position could reach, graded on its own rather than left
+// inside a bulleted comment — and the grading is what showed the hazard to be
+// ENTAILED by clause 2 rather than an independent placement constraint. See the
+// paragraph below the mechanism.
 //
 // extractClojure keeps `comps`, a slice of compOffset{idx, name} where idx is an
 // INDEX INTO THE ENTITY SLICE. Step 3b dereferences it — `rec := &entities[cp.idx]`
@@ -569,12 +608,23 @@ func TestClojure_CarrierShape_6852(t *testing.T) {
 // the graph is quietly wrong. That is lua's #6885 hazard, in a package that has
 // it too.
 //
+// THE MUTANT THIS KILLS IS NOT "the carrier moved", AND THAT IS THE FINDING.
+// Moving the SHIPPED (conditional) call above step 3b — or above step 4 — is
+// DEAD, but on the EMPTINESS of the result, never on ordering: importEntities
+// do not enter `entities` until the head-prepend at clojure.go:316, so clause 2
+// rejects at every earlier point and no carrier is minted at all. The index
+// hazard is therefore ENTAILED by the clause-2 requirement rather than an
+// independent constraint on that call, and the only shape that reaches it is an
+// UNCONDITIONAL carrier placed before step 3b. Scored: that mutant produces
+// FOUR wrong-owner rows here (the carrier takes Triangle's IMPLEMENTS -> Shape;
+// Hexagon's IMPLEMENTS -> Sized lands on Triangle), at both depths — and the
+// SAME unconditional carrier placed AFTER step 3b produces ZERO, which is what
+// separates this test's subject from the over-emission tests' subject.
+//
 // The enumeration behind "the ONE conjunct": compOffset.idx has exactly one
 // consumer (clojure.go:283, step 3b). opOffset.idx has ZERO consumers — step 4
 // reads only op.name — so a shift cannot be observed through it, and no test is
-// written for something that cannot be observed. The final
-// `entities = append(head, entities...)` invalidates both, which is why the
-// carrier is prepended AFTER it and not before.
+// written for something that cannot be observed.
 //
 // The fixture is the ordinary Clojure shape that makes this production-reachable
 // — a namespace that requires its protocols and then extends its own records —
--- a/internal/extractors/clojure/file_carrier_6852_test.go
+++ b/internal/extractors/clojure/file_carrier_6852_test.go
@@ -1,0 +1,632 @@
+package clojure_test
+
+// file_carrier_6852_test.go — #6852, clojure arm.
+//
+// buildImportEntities (clojure.go) stamps `FromID: filePath` on the IMPORTS
+// edge of every (:require ...) / (:use ...) / (:import ...) entry.
+// internal/resolve/refs.go has no path→entity index, so a path-valued FromID
+// resolves if and only if some emitted node carries that exact string as its
+// Name. Same defect #6815 fixed in erlang/nim/groovy and #6852 fixed in bicep
+// (#6864), terraform (#6871), html (#6879), fsharp (#6880), shell (#6882),
+// dockerfile, lua (#6885) and astro (#6891); same fix,
+// extractor.PrependFileCarrier.
+//
+// ONE ANCHORING SITE, N EDGES — fsharp's / lua's / astro's multiplicity shape,
+// not bicep's single edge and not terraform's two sites. buildImportEntities is
+// the only bare `FromID: <path>` in the whole package: the CONTAINS edges the
+// namespace record owns, the CALLS edges collectCalls builds and every
+// IMPLEMENTS/EXTENDS edge hierarchy.go scans all leave FromID EMPTY, and
+// hierarchy.go's header says why (a non-empty non-hex FromID would be rewritten
+// away from the owning component). One import record is emitted per deduped
+// (module, local name) pair, so an ns form with four requires anchors four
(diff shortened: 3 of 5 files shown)


cajasmota:
## Coordinator verification — DEAD, and the new dot-leading test is the only thing covering one of the cells

The arm's own mutants and the review between them cover placement, the two `lang` signals, the sets, the roster and clause 3's deletion. The axis left alone was the shape a maintainer would plausibly *write*: keep the anchoring condition, drop the clause-3 consultation.

### The mutant

Replace the `PrependFileCarrier` call with a hand-rolled equivalent that mints the carrier whenever **anything anchors on the path**, skipping `FileCarrierFor` entirely:

```go
for i := range entities {
    for j := range entities[i].Relationships {
        if entities[i].Relationships[j].FromID == filePath {
            return append([]types.EntityRecord{
                extractor.FileEntity(extractor.FileInput{Path: filePath, Language: "clojure"}),
            }, entities...)
        }
    }
}
return entities
```

This is the tempting simplification — it looks like it preserves the conditional (no anchor, no carrier), and it keeps every over-emission control green. What it silently discards is clause 1 and clause 3. Gate asserted, `go vet` clean.

### Result: DEAD, on three tests and four cells

```
--- FAIL: TestClojure_EmptyPathGetsNoCarrier_6852                                    (clause 1)
--- FAIL: …NamespaceNamedLikeThePathGetsNoSecondCarrier_6852/root                    (clause 3, ns route)
--- FAIL: …NamespaceNamedLikeThePathGetsNoSecondCarrier_6852/root_namespace_owning_no_edges
--- FAIL: …SelfRequiringStubUnderADotLeadingPathGetsNoSecondCarrier_6852/dot_leading_root
--- FAIL: …SelfRequiringStubUnderADotLeadingPathGetsNoSecondCarrier_6852/dot_leading_nested_clj_kondo
    file_carrier_6852_test.go:627: exactly 1 record may be named ".clj-kondo/hooks/foo.clj", got 2
      — two records under one graph.EntityID, which does not hash Subtype (#6369/#6480)
```

**The last line is the point.** `dot_leading_nested_clj_kondo` is a cell that did not exist before this round, and it is the **only** cell covering clause 3 at nested depth — the arm originally claimed clause 3 was reachable "at root depth only, by exactly one route". Under the old test set my mutant would still have died at root, and the nested duplicate would have shipped unobserved.

So the review's finding did not merely correct three sentences: it closed a cell no test occupied. Together with the author's own re-score — `dot > 0` → `dot >= 0` is DEAD **and killed by the new test only**, meaning nothing in the package had ever observed that guard — the correction converted two ungraded branches into graded ones.

Restored by `cp` from a shasum-verified backup, positive control green, worktree removed.

### One correction to my own relay

I passed on the review's "minor" that `writeSpanFixture6118` has no `go` or `rs` members. **The author is right and the review was wrong** — I checked the file: `"svc/handler.go"` at `custom_extractor_spans_6118_test.go:240` and `"rs/routes.rs"` at `:251`, inside a map running from `:36` to `:260`, 18 members total.

The author's explanation of the miscount is worth keeping: a `grep -n '^func '` lands on line 247, `func (h *Handler) ServeHTTP` — which is **inside the Go source string literal**, making the map appear to end early. A line-oriented scan over a file that embeds source in string literals will be fooled exactly there, and this repo has been bitten by the same class before (`const K = string(types.X)` defeating an AST scanner, #6118's own family).

The load-bearing half is unaffected and confirmed: **no `.clj` member, so `cmd/grafel` does not grade this arm.**

Issue #6894, opened 2026-09-06
Nine dashboard sites match e.Kind == "Route" raw while graphstate.go matches the stripped kind — the fix is a WIDENING to accept both spellings, not a repoint


cajasmota:
Nine dashboard sites match `e.Kind == "Route"` raw, so every `SCOPE.Route` is missing from both Paths panes, all four detail handlers, both search paths and the OpenAPI export

Found by the independent review of PR #6893 (#6820). **Not caused by that PR** — it is the identical defect on the identical lines, for a different kind, and unlike `Endpoint` this one is **live**.

## The defect

Nine of the ten sites #6893 repoints also carry a sibling clause comparing the **raw** kind:

```go
e.Kind == "Route"
```

sitting next to a `kind :=` that has already been stripped of its `SCOPE.` prefix. So the clause matches bare `"Route"` and **not** `SCOPE.Route`.

## Why this is worse than the `Endpoint` case #6893 fixes

`SCOPE.Endpoint` turned out to be an **admitted-but-unproduced** kind — it is in `types.AllEntityKinds()` (`kinds.go:728`) but no producer emits it, so the repoint arm in #6893 is inert today.

**`SCOPE.Route` is produced, by at least seven sites:**

- `internal/custom/lua/kong.go`
- `internal/custom/lua/routing.go` (×5)
- `internal/custom/java/vaadin_gwt.go`
- `internal/engine/http_endpoint_utoipa_crossfile.go:531`
- `api_gateway_routing_edges.go`
- `frontend_route_edges.go`

So Kong/nginx/Lua routes and Vaadin `@Route` pages are **silently absent** from the HTTP surface today. Probed at PR #6893's HEAD with a two-entity fixture:

```
v1 paths pane      = [/bare/route]      # /scope/route absent
openapi paths      = [/bare/route]      # /scope/route absent
groupTopFrameworks = [gin kong]         # the stripped-kind site: accepted
```

That last line is the tell. `groupTopFrameworks` reads the **stripped** kind and accepts `SCOPE.Route` fine; the nine raw-comparison sites do not. The inconsistency is within one feature.

## Evidence it is an accident, not a design

`graphstate.go` writes `kind != "Route"` against the **stripped** kind — the one site of the ten that does accept `SCOPE.Route`. A deliberate decision to treat the two spellings differently would not have one site disagreeing with nine on the same predicate in the same feature.

## Scope

[...]

Issue #5782, opened 2026-07-16
Enhancement: connect the config/dependency/messaging layers to support migration investigations


cajasmota:
## Type

Enhancement / product feedback — grounded in a real investigation, with tool-level evidence.

## Context

I used grafel to plan a transport migration on an event-driven microservices codebase
(9 Quarkus services, Kafka via SmallRye Reactive Messaging): **replace Kafka with SNS+SQS
and introduce Terraform**. grafel was excellent at reconstructing the *event topology*
(`find` + `inspect` + the now-fixed `orient view=topology`), but weak at the
*migration/config/infra* layer. I then re-tested the tools I hadn't tried
(`impact_radius`, `related`, `trace effects`, `patterns`, `debt`, `subgraph`) to check
whether the gaps were real or just me missing a tool. Findings below are evidence-based.

**grafel version:** `v0.1.8-4-g61c99101b`

I have reviewed the anonymized report at the bottom and confirm it contains no source code,
no real file paths, and no un-hashed identifier names.

---

## What already works (self-correction)

Credit where due — one thing I initially assumed needed `grep` is already a one-call query:

- **`grafel_find(query="quarkus-messaging-kafka", search="substring", cross_repo=true)`**
  returns all 16 nodes (8 `SCOPE.Component` @ `build.gradle` + 8 `SCOPE.Package`) across
  every repo. So "which services depend on artifact X" — the core inventory question of a
  dependency migration — is fully answerable today. 👍

## The core gap: indexed-but-not-connected layers

grafel indexes dependencies, message topics, and config files, but does **not link them to
each other or to the code that uses them**. For structure questions this is fine; for
*change/migration* questions it's the whole ballgame. Concrete, reproducible evidence:

### 1. Dependency nodes are islands
```
grafel_related(entity_id=<quarkus-messaging-kafka Component>, direction="neighbors", depth=2)
=> {"callees":[], "callers":[]}
```
You can find *who declares* a dependency, but there is no edge from the dependency to the
`MessageTopic`/`@Channel`/emitter entities that actually exercise it. So "if I remove
`quarkus-messaging-kafka`, what breaks?" is unanswerable from the graph.

**Ask:** a `DEPENDS_ON` / `PROVIDED_BY` edge linking a dependency to the framework
entities it enables (topics, channels, endpoints). That turns dependency removal into a
real blast-radius query.

### 2. `impact_radius` does not follow pub/sub across repos
```
[...]

PR #6907, opened 2026-09-06
fix(#6902): four reachability predicates dropped bare "Route" — and the corpus figure #6894 could not get is **156 real HTTP routes reported as dead code today**

### LINKED ISSUE #6902: isEndpointReachKind and FOUR siblings select on the RAW kind and drop bare "Route" — the mirror of #6894, and here the excluded spelling is the CI-gated one
`isEndpointReachKind` and three siblings select on the RAW kind and drop bare `"Route"` — the mirror of #6894, and this time the excluded spelling is the CI-gated one

Found by the independent review of PR #6900 (#6894). **Not caused by that PR** — it is the same defect one site over, in the opposite direction, and it was missed by #6894's own site enumeration (which counted ten, not eleven).

## The defect

`internal/dashboard/handlers_coverage_reach.go:157` — `isEndpointReachKind(e.Kind)` selects on the **raw** kind and accepts `SCOPE.Endpoint`, `SCOPE.Route` and `http_endpoint_definition`. It therefore **drops bare `"Route"`**.

That is the exact mirror of #6894, where nine sites accepted bare `"Route"` and dropped `SCOPE.Route`. Here the accepted set is the prefixed one and the dropped spelling is the bare one.

## Why this half is the worse half

#6894's headline — that the dropped spelling is a live recall loss — could **not** be confirmed for `SCOPE.Route`: no live count was obtainable and three purpose-built end-to-end indexes produced zero entities of either spelling.

**For bare `"Route"` it is confirmed, from inside the repo.** Four golden fixtures assert bare-`"Route"` entities with `must_exist: true`:

```
internal/quality/golden/java-spring-mini/expected.json
internal/quality/golden/go-chi-mini/expected.json
internal/quality/golden/python-fastapi-mini/expected.json
internal/quality/golden/python-django-mini/expected.json
```

and that gate has been always-on since `fad4c5f33`. So bare `"Route"` provably survives indexing end-to-end **today**, and the producers are numerous and Go-side: `internal/custom/java/{play,spring_webflux,akka_http,javalin,vertx,struts}_routes.go` plus `internal/engine/{spring,django}_routes.go`. Zero rule packs declare it.

So entities that CI guarantees exist are excluded from the coverage-reach view.

## The same shape recurs outside the dashboard

Three further sites select on the raw kind in the same family, and the dashboard's own comment says the first of them is mirrored deliberately:

- `internal/mcp/reachability_tools.go:305` — `isEndpointKind`
- `internal/links/reachability.go:103` — `frameworkEntryKinds`
- `internal/enrichment/candidates.go:435`

**Whether each has the same exclusion is not established** — only `handlers_coverage_reach.go` was measured. They are listed because a fix that repairs one and leaves three mirrored copies diverging is how this defect family propagates; #6846 found eleven hand-maintained copies of one exclusion list, two of them live bugs.

1. **Per site, which spellings belong.** #6894's lesson is that the analogous case does not carry over: for `Endpoint` the two spellings are *different concepts* (Electron IPC vs HTTP), for `Route` they are the *same concept*. Read each of the four sites individually rather than applying one rule.
2. **The forbidden direction.** A test showing bare `"Route"` now reaches the coverage-reach view cannot see that something else leaked in. Assert the absence and score that assertion in its own direction.
3. **Whether the four sites should share a predicate.** They are already near-duplicates and one is a documented mirror of another. Sharing would make the next divergence impossible; #6846 is the precedent for doing it, and for putting a lower bound under the shared list.
4. **The trap from #6893, restated because it caught that PR:** a positive control matched by *another* clause in the same predicate grades nothing — control and arm fire together and the arm can die unnoticed. Each control must be matched by nothing else in the predicate under test.

## Not established

- Whether the three non-dashboard sites share the exclusion.
- Whether anything downstream depends on bare `"Route"` being absent from coverage-reach.
- A corpus count of bare-`Route` entities excluded in practice — the golden fixtures prove existence, not volume.

Refs #6894, #6900, #6893, #6846.

cajasmota:
fix(#6902): four reachability predicates dropped bare "Route" — and the corpus figure #6894 could not get is **156 real HTTP routes reported as dead code today**

Closes #6902

The issue's premise is **confirmed**, with two corrections to its site list and one
measurement it explicitly marked "not established" that I *was* able to get.

---

### 1a. Every site in the four packages that selects on a `Route` spelling

| # | Site | Shape | What the view is for | `SCOPE.Route` | bare `Route` | Verdict |
|---|---|---|---|---|---|---|
| A | `internal/dashboard/handlers_coverage_reach.go:157` `isEndpointReachKind` | `switch types.EntityKind(kind)` on the RAW kind | endpoint-level test-reachability roll-up + orphan list on `GET /api/quality/coverage/{group}` | accepted | **dropped** | **fixed** |
| B | `internal/mcp/reachability_tools.go:305` `isEndpointKind` | the same switch, **byte-identical body** | `grafel_test_reachability` — the `Endpoints`/`Endpoints reachable` roll-up, the `endpoints_only` listing, and the `endpoint ` row tag | accepted | **dropped** | **fixed** |
| C | `internal/links/reachability.go:104` `frameworkEntryKinds` | `map[string]bool` keyed on the RAW kind | the link pass's BFS **seed set** — kinds that are entry-points with no inbound edge; decides `reachable` in the persisted `<group>-reachability.json` | accepted | **dropped** | **fixed** |
| D | `internal/mcp/dead_code.go:228` `frameworkEntryKindsMCP` | the same map, documented as a mirror of C | `grafel_dead_code` — the same seed set, applied to the sidecar and to the live-recompute fallback | accepted | **dropped** | **fixed** — **and the issue does not name this site** |
| E | `internal/enrichment/candidates.go:433` `qualifyHTTPKinds` | `map[string]bool` | which kinds always qualify for LLM enrichment | accepted | **accepted** | **no defect** |
| F | `internal/enrichment/candidates.go:98` | `case` arm | enrichment tier assignment | accepted | accepted | no defect |
| G | `internal/enrichment/pricing.go:38` | `case` arm | enrichment cost band | accepted | accepted | no defect |

**Correction 1 — the issue's fourth candidate is clean.** `internal/enrichment` was listed
as a suspect at all three of its sites. It already lists `"Route"` beside `"SCOPE.Route"`
in every one of them, and `internal/types/kinds.go:614` already records that fact in a
comment. Nothing to fix there; it is the *counter-example* that shows the repo already
knows both spellings are the same concept.

**Correction 2 — there is a fifth site, and it is the worst one.**
`internal/mcp/dead_code.go`'s `frameworkEntryKindsMCP` is a hand-copied mirror of C
(its own comment says so) and carries the same gap. It is the site with the most
user-visible harm: an unseeded HTTP route is reported to an agent as **dead code**.

**Confirmed as stated: A and B are literal duplicates, not "a documented mirror".**
`shasum` of the two switch bodies before the change: `16446cab4e5b…` for both.
After the change, both are `341c5608…`. The comment on each now says so and says
to keep them in step.

### 1b. Which spellings belong, per site — established here, not inherited

#6894's own lesson is that the analogous case did **not** carry over: for `Endpoint`,
bare and `SCOPE.` are different concepts (Electron IPC vs HTTP), so #6893 *repointed*;
for `Route` they are the same concept, so #6900 *widened*. I read each of the five
sites rather than applying that answer:

- **A and B** ask "is this entity an HTTP endpoint surface, for a test-reachability
  roll-up". A bare-`Route` entity from `internal/engine/spring_routes.go` is exactly
  that. There is no reading on which a Spring route is an endpoint surface and a
  Vaadin `@Route` is not. **Both belong.**
- **C and D** ask "is this entity a framework-managed entry-point that needs no inbound
  edge". A bare-`Route` entity is invoked by the framework by construction — that is
  what makes it a route. **Both belong**, and D must equal C because D is declared to
  mirror it.
- **E/F/G** already accept both.

**No site should keep excluding bare `Route`.** All four exclusions are accidents of
the raw-kind comparison, exactly as in #6894.

The underlying fact both spellings share (`types.EntityKindRoute` = `"SCOPE.Route"`,
`types.EntityKindRouteBare` = `"Route"`, added together by #6776 arm B7 with the note
"Both members of each pair are live") is unchanged from #6900 §1b; I re-read the kind
declarations rather than citing that PR.

### 1c. Existence: CI guarantees it

Nine `must_exist: true` rows across four always-on golden fixtures
(`java-spring-mini` 2, `go-chi-mini` 2, `python-fastapi-mini` 2, `python-django-mini` 3)
assert bare-`Route` entities, on a gate always-on since `fad4c5f33`. Producers are
Go-side and numerous:
`internal/custom/java/{play,spring_webflux,akka_http,javalin,vertx,struts}_routes.go`
and `internal/engine/{spring,django}_routes.go`. Zero rule packs declare it.

### 1d. **The corpus figure — #6894 could not get one. This one can.**

This is the part the issue lists under "Not established", and it is now established.

#6900 could not read the stored graph (`graph.fb` is format version 5; this build
requires 6, `internal/graph/fbversion/version.go:41`), and that is still true — so the
*coverage-reach* half (sites A and B) still has **no** volume figure, only the golden
fixtures' proof of existence. Stated plainly, as the brief asks.

**But sites C and D do not need the graph.** The link pass writes its verdict to a
plain-JSON sidecar, `~/.grafel/groups/<group>-links-reachability.json`, which is
version-independent and readable today. The one on this machine is a real index of the
registered corpus group (86 MB, 427,300 entities, written 2026-08-05):

| kind | entries in the sidecar | persisted **unreachable** |
|---|---|---|
| `http_endpoint_definition` | 2,061 | **0** |
| `SCOPE.Route` | 30 | **0** |
| **bare `Route`** | **575** | **156** |

Zero seeded HTTP kinds are unreachable, as you would expect of a seed set. Bare `Route`
is the only HTTP kind in the corpus with a non-zero unreachable count, because it is the
only HTTP kind that was not seeded. **156 real HTTP routes were persisted as unreachable
and are reported by `grafel_dead_code` as dead code today** (the handler prefers the
sidecar over live recompute). The other 419 were rescued incidentally by an inbound
`HANDLES`/`ROUTES_TO`/`REGISTERS` edge or by the source sniffer — which is why the
number is 156 and not 575, and why a fixture with no edges is the only honest way to
grade the seed arm.

So the claim the issue calls "the worse half" is not just existence-confirmed, it is
**volume-confirmed** for two of the four sites. (Counts only; no repo identifiers,
paths or entity names are reproduced here.)

---

## Step 2 — the change

Four one-line widenings, each with a comment saying which two spellings match and why:

```go
 	case types.EntityKindEndpoint,
 		types.EntityKindRoute,
+		types.EntityKindRouteBare,
 		types.EntityKindHTTPEndpointDefinition:
```

at A and B (keeping the two bodies byte-identical), and

```go
 	"SCOPE.Route":              true,
+	"Route":                    true, // #6902
```

at C and D.

---

## The forbidden direction, scored in its own right

Each of the four site tests is a **four-way verdict**, not a presence check:

1. **vacuity guard**, checked first with `t.Fatalf` so nothing below can pass on an
   empty payload — an `http_endpoint_definition` in the emitted list at A/B, a seeded
   `http_endpoint_definition` marked reachable at C, and (inverted, because dead_code
   asserts absence) the decoy actually appearing in the dead list at D;
2. **bare `"Route"` present** — the #6902 recall arm;
3. **`SCOPE.Route` still present** — the direction a #6893-style repoint would break;
4. **a `SCOPE.Class` decoy ABSENT** — the forbidden direction. It carries the same
   `test_reachable` prop / the same file / the same shape as everything else, so nothing
   upstream filters it out for us; a decoy something else rejects would grade nothing.

Direction 4 is graded on its own by `I-A-decoy`, `I-B-decoy`, `I-C-decoy` and
`I-D-vacuity`, and by the three `*3-widen-any` mutants, which are killed **by the decoy,
not by the Route control**.

## The trap that caught #6893, checked explicitly per predicate

#6893's positive control was an `http_endpoint_definition` — a kind the predicate under
test already matched — so control and arm only ever fired together and the arm sat ALIVE
at three of ten sites with the suite green.

Every control here is a bare-`Route` entity, checked against **every other clause of the
predicate it grades**:

| site | the predicate's other clauses | does any match bare `"Route"`? |
|---|---|---|
| A | `== SCOPE.Endpoint`, `== SCOPE.Route`, `== http_endpoint_definition` (exact string compares) | no |
| B | identical to A | no |
| C | the other 7 map keys; framework-invocation edge targets; the source sniffer | no — **the fixture has no edges and an empty `FileRoot`**, so the seed map is the only thing that can make anything reachable |
| D | the other 8 map keys; `isFrameworkOrHandler(e)`; `HANDLES/HANDLES_SIGNAL/NAVIGATES_TO/ROUTES_TO/REGISTERS` targets | no — the fixture has no edges, and `isFrameworkOrHandler` cannot match the control: its name `/orders-6902` fails `routeNameRe` (`^(get\|post\|…)[\s:]`), is not a lifecycle name, is not a constructor (`X.X`), does not start with `on`+upper or `handle`, and mcp's `isHTTPEndpointKind` classifies `"Route"` as `endpointKindNone` |

Site D is where this check earned itself: a route entity named `GET /orders` would have
been seeded by `isFrameworkOrHandler` on its **name** and the seed-map arm would have
graded nothing. The `http_endpoint_definition` entities in A and B are vacuity guards
only, and each test says so at the point of use.

## Mutation score — 29 mutants, 29 DEAD, 0 ALIVE, permissive direction first

**Round 1 — 12 site mutants, three per site**, at **every** site carrying the arm
(including D, which the issue does not name):

| id | site | delete-the-new-arm | repoint (drop `SCOPE.Route`) | widen-to-anything |
|---|---|---|---|---|
| A | `handlers_coverage_reach.go` `isEndpointReachKind` | DEAD | DEAD | DEAD |
| B | `reachability_tools.go` `isEndpointKind` | DEAD | DEAD | DEAD |
| C | `links/reachability.go` `frameworkEntryKinds` | DEAD | DEAD | DEAD |
| D | `mcp/dead_code.go` `frameworkEntryKindsMCP` | DEAD | DEAD | DEAD |

`*1-delete-arm` reverts the fix. `*2-repoint-drop-scope` is the #6893 shape — it keeps
bare `Route` and drops `SCOPE.Route`, and is killed by assertion 3, so the widening is
graded in **both** directions. `*3-widen-any` is the permissive direction
(`if kind != "" { return true }` at A/B, `if e.Kind != "" {` / `if e.Kind != "" || …` at
the C/D use sites) and is killed by the decoy.

**Round 2 — 17 assertion-inversion mutants.** A fresh positive control is itself a no-op
candidate, so every new assertion is inverted in place, vacuity guards included:
`I-A-{bare,scope,decoy,total,vacuity}`, `I-B-{bare,scope,decoy,rollup,vacuity}`,
`I-C-{bare,scope,decoy,vacuity}`, `I-D-{bare,scope,vacuity}`. **All 17 DEAD** — no new
assertion is vacuous.

Discipline per mutant: exact string edit whose **pre-image count is asserted as 1
before** the edit and whose **post-image is asserted present and pre-image absent
after** it (a silently-unapplied edit reads as ALIVE and argues the opposite);
`gofmt -l` clean; `go vet` clean; whole-package `go test -count=1`; restore by `cp` from
a shasum-verified backup with the digest re-checked after restore. Every round ended
`TREE PRISTINE` (all seven touched files byte-identical to baseline), and
`git diff <commit>..` is empty.

---

## The design question — measured, recommended, NOT refactored here

**Sharing is warranted, and this PR deliberately does not do it.**

What the measurement shows: **two byte-identical switch bodies in two packages, and two
hand-copied maps in two packages, and each pair's duplication is asserted only by a code
comment.** That is the #6846 shape (eleven hand-maintained copies of one exclusion list,
two of them live bugs) at a smaller scale, and it is how this defect propagated: #6894
fixed ten dashboard sites and left these four untouched one file away.

Recommended shape, if you file it:

- **`internal/types` is the obvious home.** It already hosts `IsHTTPEndpointKind` and
  gained a forbidden-row test in #6900 — the exact place a lower bound under a shared
  list belongs. Add `types.IsHTTPRouteKind(kind)` (both `Route` spellings) and
  `types.IsEndpointSurfaceKind(kind)` (the A/B switch), each with an accepting-row AND a
  forbidden-row table test.
- **Who imports what**: `internal/dashboard` and `internal/mcp` already import
  `internal/types`, so A and B are a straight substitution with no new edge.
- **The seed maps are harder and should be a second step.** C and D are *data*, not a
  predicate, and D exists specifically to avoid an `mcp → links` dependency (its comment
  says so, as does `reachabilityEdgeKindsMCP` beside it). Sharing them means moving the
  seed set into a leaf package both can import — worth doing, but it moves a
  link-pass-owned concept out of `internal/links`, which is a design call, not a bug fix.
- **Cost**: roughly 30 lines in `internal/types` plus a forbidden-row test, four call-site
  substitutions, and one leaf-package decision for the maps.
- **Also worth a row in that issue**: `reachabilityEdgeKindsMCP` (`mcp/dead_code.go:216`)
  is a *third* hand-copy of `reachabilityEdgeKinds` (`links/reachability.go:75`), of the
  same pair of packages, with no test asserting the two are equal. Nothing is wrong with
  it today; nothing would catch it if something were.

Doing any of that inside a bug fix would mean refactoring four packages behind a
one-line defect, so it is left as a follow-up.

---

## Packages run

Green, `-count=1`, exit codes checked (no pipes):

- `./internal/dashboard/` (36.7s)
- `./internal/mcp/` (109–143s)
- `./internal/links/` (10.7s)
- `./internal/types/` (6.0s)
- `./internal/enrichment/` (1.1s) — run even though it is unchanged, because the
  measurement's "no defect" claim about E/F/G is about that package
- `./cmd/grafel/` (141.3s) — run rather than argued about. `internal/links` is an
  index-time pass, so its output is not purely read-side: the widening makes more
  entities carry `reachable=true` and changes the sidecar. The digest test hashes
  `Kind|Name|QualifiedName|Subtype|SourceFile|StartLine|EndLine` plus relation rows, none
  of which this diff moves, and it passes.

`gofmt -l` and `go vet` clean on all four changed packages.

**NOT run**: the rest of the tree, including `./internal/quality/` (the golden gate — no
extractor changes here, so no `must_exist` row can move), every extractor package,
`./internal/graph/`, `./internal/docgen/`. An absence claim is only as wide as the
packages run.

---

## Things that contradict the brief or the issue

1. **The issue's fourth candidate is not defective.** `internal/enrichment` already
   accepts both spellings at all three of its `Route` sites.
2. **There is a fifth site the issue does not list**, `internal/mcp/dead_code.go`, and
   it is the one with the demonstrable production harm.
3. **A corpus figure *was* obtainable**, contrary to the brief's expectation that the
   fb-v5/v6 mismatch would block it as it blocked #6894. The reachability sidecar is
   plain JSON and does not go through the flatbuffer loader. The figure only covers
   sites C and D; sites A and B still have existence (golden fixtures) and not volume.
4. **Three adjacent `isEndpointKind` functions carry no `Route` arm at all** and are
   therefore out of scope, recorded so the next enumeration does not re-derive them:
   `internal/dashboard/handlers_security.go:152` and `internal/graph/coverage.go:608`
   (both endpoint-only, no route clause), and `internal/docgen/llm_bundle.go:1724`
   (already accepts both — it lowercases, strips `SCOPE.` and does
   `strings.Contains(k, "route")`).

---

## Corrections before merge — the headline is over-claimed in two independent ways

Both found after the body was written (one by the coordinator's mutant, one by the independent review). The **fix** is unchanged and correct at all five sites; what changes is what this PR may claim.

### 1. The 156 do not clear on upgrade — they clear on the next link-pass run

`frameworkEntryKindsMCP` (site D) is consulted **only on the recompute path**. `handleDeadCode` prefers the on-disk sidecar, and when one exists it projects the persisted verdict without ever consulting the seed map. The 156 are already persisted as `reachable: false`.

So: *"156 real HTTP routes are reported as dead code today"* — **accurate**. *"This PR stops them being reported"* — **only after the link pass re-runs** and rewrites the sidecar using site C's corrected seed set.

Established by a coordinator mutant nobody else ran: forcing `loadReachabilitySidecar` to never load is **DEAD**, but killed by `TestDeadCodeSurfacesDegradedRepos` and `TestDeadCodeUndegradedResponseHasNoPartialMarker` — both **#6839** degraded-repo tests, neither a Route test, neither from this PR. The sidecar path is exercised; **no test covers a sidecar carrying a bare-`Route` entry marked unreachable**, which is precisely the state the 156 live in.

### 2. ~124 production routes, not 156

The independent review read the sidecar directly and reproduced every figure (575 bare `Route`, 156 unreachable; `SCOPE.Route` 30 → 0; `http_endpoint_definition` 2,061 → 0), and confirmed the 156 are route-shaped: 143 URL-path names, 156 distinct on `(repo, name, file)` — no duplicates, no vendored code.

But **32 of the 156 are in Java test-source trees**, and all 156 come from a single repo. The honest figure is **~124 production routes in one repo**, not "156 real HTTP routes" across a corpus.

### 3. One cosmetic error in the report

The *before* shasum `16446cab` does not reproduce (the reviewer gets `c48fe09f`); the *after* `341c5608` matches exactly. The identical-before / identical-after claim itself holds — only the transcribed digest is wrong.

## Two findings filed rather than fixed here

- **The two seed maps still differ after this PR.** `frameworkEntryKindsMCP` carries `"SCOPE.ChannelBinding"` (#5782 / ADR-0025) and `links`' `frameworkEntryKinds` does not. Because the handler prefers the sidecar, ADR-0025's *"never report a ChannelBinding as dead code"* is defeated in the normal case. No volume evidence (zero such entities in the measured index), but structurally real.
- **`isFrameworkOrHandler` is ungraded as a seeding disjunct** — deleting it leaves the whole `./internal/mcp/` suite green. (Its being ALIVE is also the *positive* result for this PR: it proves empirically that the site-D control cannot be seeded by anything but the map, so the #6893 trap is cleared by measurement rather than by argument.)

## Verification summary

**35 mutants, 34 DEAD, 1 deliberately ALIVE** (the `isFrameworkOrHandler` disjunct above). The review reproduced all 29 of the author's using true *inversions* rather than deletions — a stronger vacuity check — and added six, including a precise `widen-class` over-widen at each site killed by the decoy rather than by the Route control.

`internal/quality` and the extraction-quality ratchet **cannot move**, with a reason rather than an assumption: neither imports any of the three packages, and no golden fixture carries a `reachable` field — reachability is a post-index link pass, not extraction. CI's Extraction-quality gate is green independently.

### DIFF
--- a/internal/dashboard/bare_route_reach_6902_test.go
+++ b/internal/dashboard/bare_route_reach_6902_test.go
@@ -1,0 +1,152 @@
+package dashboard
+
+import (
+	"encoding/json"
+	"testing"
+
+	"github.com/cajasmota/grafel/internal/coverage"
+	"github.com/cajasmota/grafel/internal/graph"
+	"github.com/cajasmota/grafel/internal/types"
+)
+
+// ---------------------------------------------------------------------------
+// #6902 — the mirror of #6894, one site over and in the opposite direction.
+//
+// isEndpointReachKind switches on the RAW kind and accepts SCOPE.Endpoint,
+// SCOPE.Route and http_endpoint_definition. It therefore drops bare "Route"
+// (types.EntityKindRouteBare) — the spelling every Java route extractor and
+// internal/engine/{spring,django}_routes.go emits, and the spelling four
+// golden fixtures (java-spring-mini, go-chi-mini, python-fastapi-mini,
+// python-django-mini) assert with must_exist:true on an always-on gate.
+//
+// So entities CI guarantees exist were excluded from the coverage-reach
+// roll-up: not counted in TotalEndpoints, and never listed as orphans even
+// when no test reaches them.
+//
+// Both Route spellings mean the SAME concept — an HTTP route (#6894 §1b,
+// #6776 arm B7: "Both members of each pair are live"). Unlike the Endpoint
+// pair, where bare "Endpoint" is Electron IPC and "SCOPE.Endpoint" is HTTP,
+// there is no reading on which this view wants one spelling and not the
+// other. The fix is a WIDENING.
+//
+// THE POSITIVE CONTROL, and the trap that caught #6893. Its control was an
+// http_endpoint_definition — a kind the predicate under test already matched
+// — so control and arm only ever fired together and the arm sat ALIVE.
+// The control here is the bare-"Route" entity, and the predicate's other
+// three arms are exact string comparisons:
+//
+//	kind      == SCOPE.Endpoint  == SCOPE.Route  == http_endpoint_definition
+//	"Route"   no                 no              no
+//
+// so nothing else in isEndpointReachKind can match it. The
+// http_endpoint_definition below is present ONLY as a vacuity guard.
+//
+// Every assertion reads the EMITTED ARTEFACT: the JSON of the
+// ReachabilitySummary that handleGroupCoverage embeds verbatim under
+// "reachability" in GET /api/quality/coverage/{group}.
+// ---------------------------------------------------------------------------
+
+const (
+	// bareRouteReachName grades the RECALL direction — the #6902 arm.
+	bareRouteReachName = "/bare/route-6902"
+	// scopeRouteReachName grades the direction a repoint-style "fix" would
+	// break: SCOPE.Route was already accepted and must stay accepted.
+	scopeRouteReachName = "/scope/route-6902"
+	// decoyReachName grades the FORBIDDEN direction. SCOPE.Class is matched by
+	// no arm of the predicate, before or after the fix; it carries the same
+	// test_reachable prop as everything else so nothing upstream filters it
+	// out for us — a decoy something else rejects would grade nothing.
+	decoyReachName = "/decoy/class-6902"
+	// vacuityReachName is the vacuity guard only, never a control.
+	vacuityReachName = "/real/http-6902"
+)
+
+// reach6902Doc returns one repo document whose four entities all carry
+// test_reachable=false, so every entity the predicate admits must appear in
+// the emitted orphan list.
+func reach6902Doc() *graph.Document {
+	mk := func(id, name, kind string) graph.Entity {
+		return graph.Entity{ID: id, Name: name, Kind: kind, SourceFile: "src/routes.go"}.
+			WithProperties(map[string]string{coverage.PropTestReachable: "false"})
+	}
+	return &graph.Document{Entities: []graph.Entity{
+		mk("bare-route-6902", bareRouteReachName, string(types.EntityKindRouteBare)),
+		mk("scope-route-6902", scopeRouteReachName, string(types.EntityKindRoute)),
+		mk("decoy-class-6902", decoyReachName, string(types.EntityKindClass)),
+		mk("real-http-6902", vacuityReachName, "http_endpoint_definition"),
+	}}
+}
+
+// orphanNamesFromPayload marshals the summary exactly as the coverage handler
+// does and reads the names back out of the wire bytes.
+func orphanNamesFromPayload(t *testing.T, s *ReachabilitySummary) []string {
+	t.Helper()
+	buf, err := json.Marshal(s)
+	if err != nil {
+		t.Fatalf("marshal ReachabilitySummary: %v", err)
+	}
+	var wire struct {
+		TotalEndpoints int `json:"total_endpoints"`
+		Orphans        []struct {
+			Name string `json:"name"`
+		} `json:"orphans"`
+	}
+	if err := json.Unmarshal(buf, &wire); err != nil {
+		t.Fatalf("unmarshal ReachabilitySummary: %v", err)
+	}
+	out := make([]string, 0, len(wire.Orphans))
+	for _, o := range wire.Orphans {
+		out = append(out, o.Name)
+	}
+	return out
+}
+
+func contains6902(hay []string, needle string) bool {
+	for _, h := range hay {
+		if h == needle {
+			return true
+		}
+	}
+	return false
+}
+
+func TestCoverageReachSummary_AcceptsBareRouteKind_6902(t *testing.T) {
+	var a reachAccumulator
+	a.accumulate(reach6902Doc(), "repoA")
+	s := a.summarize()
+
+	names := orphanNamesFromPayload(t, s)
+
+	// 1. vacuity guard — if the pane emits nothing, nothing below grades.
+	if !contains6902(names, vacuityReachName) {
+		t.Fatalf("orphans = %v, missing the plain http_endpoint_definition %q; "+
+			"every assertion below would be vacuous", names, vacuityReachName)
+	}
+	// 2. the #6902 recall direction.
+	if !contains6902(names, bareRouteReachName) {
+		t.Errorf("orphans = %v, missing the bare-\"Route\" entity %q. "+
+			"isEndpointReachKind switches on the raw kind and has no "+
+			"types.EntityKindRouteBare arm, so every Java/Spring/Django route "+
+			"entity — the spelling four golden fixtures assert with "+
+			"must_exist:true — is excluded from the coverage-reach view.",
+			names, bareRouteReachName)
+	}
+	// 3. the direction a repoint would break: SCOPE.Route must stay.
+	if !contains6902(names, scopeRouteReachName) {
+		t.Errorf("orphans = %v, missing the %q entity %q. Both Route spellings "+
+			"name the same concept; accepting the bare one must not cost the "+
+			"prefixed one.", names, types.EntityKindRoute, scopeRouteReachName)
+	}
+	// 4. the FORBIDDEN direction.
+	if contains6902(names, decoyReachName) {
+		t.Errorf("orphans = %v contains %q, a %q entity. Widening this predicate "+
+			"to bare \"Route\" must not widen it to non-route kinds.",
+			names, decoyReachName, types.EntityKindClass)
+	}
+	// The roll-up counters must move with the list, not just the list.
+	if s.TotalEndpoints != 3 {
+		t.Errorf("total_endpoints = %d, want 3 (bare Route, SCOPE.Route, "+
+			"http_endpoint_definition — and NOT the SCOPE.Class decoy)",
+			s.TotalEndpoints)
+	}
+}
--- a/internal/dashboard/handlers_coverage_reach.go
+++ b/internal/dashboard/handlers_coverage_reach.go
@@ -153,11 +153,25 @@ func (a *reachAccumulator) summarize() *ReachabilitySummary {
 
 // isEndpointReachKind reports whether a kind is an HTTP endpoint surface for
 // the reachability roll-up. Mirrors the MCP tool's isEndpointKind (#5060) so
-// the dashboard and MCP agree on what counts as an endpoint.
+// the dashboard and MCP agree on what counts as an endpoint — the two switch
+// bodies are byte-identical and must stay that way.
+//
+// #6902: BOTH Route spellings match. "SCOPE.Route" (types.EntityKindRoute) is
+// emitted by the Lua routing extractors, Vaadin @Route pages and the engine's
+// gateway/frontend-route synthesisers; bare "Route"
+// (types.EntityKindRouteBare) is emitted by internal/custom/java/{play,
+// spring_webflux,akka_http,javalin,vertx,struts}_routes.go and
+// internal/engine/{spring,django}_routes.go. Unlike the Endpoint pair — where
+// bare "Endpoint" is Electron IPC and "SCOPE.Endpoint" is HTTP (#6820/#6893)
+// — the two Route spellings name the SAME concept, an HTTP route, and #6776
+// arm B7 added both kinds together because both are live. Four golden
+// fixtures assert bare-"Route" entities with must_exist:true, so accepting
+// only the prefixed spelling excluded entities CI guarantees exist.
 func isEndpointReachKind(kind string) bool {
 	switch types.EntityKind(kind) {
 	case types.EntityKindEndpoint,
 		types.EntityKindRoute,
+		types.EntityKindRouteBare,
 		types.EntityKindHTTPEndpointDefinition:
 		return true
 	}
--- a/internal/links/bare_route_entry_6902_test.go
+++ b/internal/links/bare_route_entry_6902_test.go
@@ -1,0 +1,108 @@
+package links
+
+import (
+	"encoding/json"
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+
+	"github.com/cajasmota/grafel/internal/types"
+)
+
+// ---------------------------------------------------------------------------
+// #6902 — frameworkEntryKinds drops bare "Route" (types.EntityKindRouteBare).
+//
+// frameworkEntryKinds is the link pass's SEED set: kinds whose mere presence
+// makes an entity a framework-managed entry-point, no inbound edge required.
+// It holds "SCOPE.Route" and not "Route", so every bare-"Route" entity — the
+// spelling internal/custom/java/{play,spring_webflux,akka_http,javalin,vertx,
+// struts}_routes.go and internal/engine/{spring,django}_routes.go emit, and
+// the spelling four golden fixtures assert with must_exist:true — is left
+// unseeded and written to the reachability sidecar as UNREACHABLE, together
+// with everything only it reaches.
+//
+// Both Route spellings name the same concept (#6894 §1b; #6776 arm B7 added
+// EntityKindRoute and EntityKindRouteBare together, noting both are live), so
+// both belong in the seed set. internal/mcp/dead_code.go's
+// frameworkEntryKindsMCP is a documented mirror of this map and carries the
+// same gap; it is fixed and graded in that package.
+//
+// THE POSITIVE CONTROL, against the #6893 trap. This fixture has NO EDGES at
+// all and an EMPTY FileRoot with an unrecognised source extension, so neither
+// the framework-invocation edge kinds (HANDLES / ROUTES_TO / REGISTERS / …)
+// nor the source sniffer can seed anything. The ONLY thing that can make an
+// entity reachable here is frameworkEntryKinds[e.Kind], and "Route" is a key
+// no other entry in that map equals. The http_endpoint_definition entity is
+// the vacuity guard only.
+//
+// The assertion reads the EMITTED ARTEFACT — the persisted
+// <group>-reachability.json sidecar — not an in-memory counter.
+// ---------------------------------------------------------------------------
+
+func TestReachability_BareRouteIsAFrameworkEntrySeed_6902(t *testing.T) {
+	graphs := []repoGraph{{
+		Repo:     "repo-6902",
+		FileRoot: t.TempDir(), // empty: the source sniffer can seed nothing
+		Entities: []entityNode{
+			{ID: "bare-route-6902", Name: "/orders-6902",
+				Kind: string(types.EntityKindRouteBare), SourceFile: "src/x.unknown"},
+			{ID: "scope-route-6902", Name: "/invoices-6902",
+				Kind: string(types.EntityKindRoute), SourceFile: "src/x.unknown"},
+			{ID: "decoy-6902", Name: "LedgerReconciler6902",
+				Kind: string(types.EntityKindClass), SourceFile: "src/x.unknown"},
+			{ID: "real-http-6902", Name: "/health-6902",
+				Kind: "http_endpoint_definition", SourceFile: "src/x.unknown"},
+		},
+		// deliberately no edges
+	}}
+
+	paths := Paths{Links: filepath.Join(t.TempDir(), "g-links.json")}
+	if _, err := runReachabilityPass("g", graphs, paths); err != nil {
+		t.Fatalf("runReachabilityPass: %v", err)
+	}
+
+	sidecar := strings.TrimSuffix(paths.Links, ".json") + "-reachability.json"
+	buf, err := os.ReadFile(sidecar)
+	if err != nil {
+		t.Fatalf("read sidecar: %v", err)
+	}
+	var doc reachabilityDocument
+	if err := json.Unmarshal(buf, &doc); err != nil {
+		t.Fatalf("unmarshal sidecar: %v", err)
+	}
+	got := map[string]bool{}
+	seen := map[string]bool{}
+	for _, e := range doc.Entries {
+		got[e.EntityID] = e.Reachable
+		seen[e.EntityID] = true
+	}
+
+	// 1. vacuity guard — if the sidecar lists nothing, or lists nothing as
+	//    reachable, the assertions below grade nothing.
+	if !seen["real-http-6902"] || !got["real-http-6902"] {
+		t.Fatalf("the http_endpoint_definition seed is not in the sidecar as "+
+			"reachable (seen=%v reachable=%v); every assertion below would be "+
+			"vacuous. entries=%d", seen["real-http-6902"], got["real-http-6902"], len(doc.Entries))
+	}
+	// 2. the #6902 recall direction.
+	if !got["bare-route-6902"] {
+		t.Errorf("the bare-%q entity is reachable=%v in the sidecar. "+
+			"frameworkEntryKinds holds \"SCOPE.Route\" but not \"Route\", so every "+
+			"Java/Spring/Django route is unseeded and persisted as unreachable.",
+			string(types.EntityKindRouteBare), got["bare-route-6902"])
+	}
+	// 3. SCOPE.Route must stay seeded.
+	if !got["scope-route-6902"] {
+		t.Errorf("the %q entity is reachable=%v in the sidecar; adding the bare "+
+			"spelling must not cost the prefixed one.",
+			string(types.EntityKindRoute), got["scope-route-6902"])
+	}
+	// 4. the FORBIDDEN direction — a SCOPE.Class with no edges is NOT a
+	//    framework entry-point and must stay unreachable.
+	if got["decoy-6902"] {
+		t.Errorf("the %q decoy is reachable=true with no inbound edge and no "+
+			"sniffable source. Adding bare \"Route\" to frameworkEntryKinds must "+
+			"not seed anything else.", string(types.EntityKindClass))
+	}
+}
--- a/internal/links/reachability.go
+++ b/internal/links/reachability.go
@@ -101,11 +101,22 @@ var reachabilityEdgeKinds = map[string]bool{
 // frameworkEntryKinds are entity kinds whose presence implies a
 // framework-managed entry-point. Used to seed the BFS without needing
 // inbound edges.
+//
+// #6902: BOTH Route spellings are seeds. "SCOPE.Route" comes from the Lua
+// routing extractors, Vaadin @Route pages and the engine's gateway /
+// frontend-route synthesisers; bare "Route" comes from
+// internal/custom/java/{play,spring_webflux,akka_http,javalin,vertx,struts}_routes.go
+// and internal/engine/{spring,django}_routes.go. They name the same concept —
+// an HTTP route — and #6776 arm B7 added both kinds together because both are
+// live. Seeding only the prefixed one persisted every Java/Spring/Django
+// route into the sidecar as unreachable, along with everything only it reaches.
+// internal/mcp/dead_code.go's frameworkEntryKindsMCP mirrors this map.
 var frameworkEntryKinds = map[string]bool{
 	"http_endpoint_definition": true,
 	"http_endpoint":            true,
 	"SCOPE.Endpoint":           true,
 	"SCOPE.Route":              true,
+	"Route":                    true, // #6902
 	"SCOPE.MessageTopic":       true,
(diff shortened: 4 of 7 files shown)


cajasmota:
## Coordinator mutant — DEAD, but *what killed it* qualifies the headline

I scored a mutant structurally different from the author's 29 (all of which target the four predicates: delete-arm, repoint-drop-`SCOPE.Route`, widen-to-anything, plus assertion inversions). Mine targets the **mechanism the headline depends on** rather than the predicates: the sidecar preference in `internal/mcp/dead_code.go`.

**Mutant:** make `loadReachabilitySidecar` never load — so every call falls through to live recompute.

First attempt inserted an early `return nil, false` and **`go vet` rejected it as unreachable code**. A mutant that does not vet proves nothing, so I restructured it to `if path == path {` — same effect, vet-clean (`rc=0`). Occurrence gate asserted 1 match after the edit. Restored by `cp` from a shasum-verified backup (`8e16d658…`), tree confirmed pristine.

**Verdict: DEAD.**

```
--- FAIL: TestDeadCodeSurfacesDegradedRepos
--- FAIL: TestDeadCodeUndegradedResponseHasNoPartialMarker
FAIL  github.com/cajasmota/grafel/internal/mcp
```

So the sidecar-projection path **is** covered — the preference is not free-floating prose.

### The qualification, which is the reason I'm posting this

Look at *which* tests killed it: both are **#6839's degraded-repo tests**. Neither is a Route test, and neither is from this PR. The sidecar path is exercised, but **no test covers a sidecar that contains a bare-`Route` entry marked unreachable** — which is precisely the state the 156 live in.

That matters for the causal claim in the PR title, and I do not think the body currently says it:

- Site D's `frameworkEntryKindsMCP` is consulted **only on the recompute path** (`computeDeadCodeLive`). The `if !ok` branch.
- When a sidecar exists — the production case that produced the 156 — the handler **projects the persisted verdict** and never consults the seed map at all.
- So the fix does **not** clear those 156 on upgrade. It clears them when the **link pass re-runs** and rewrites the sidecar with site C's corrected seed set.

**"156 real HTTP routes are reported as dead code today" is accurate; "this PR stops them being reported" is true only after a link-pass re-run.** Those are different claims and the title currently reads as the second.

I am not asking for a code change on this basis — the fix is right at all four sites and the ordering (C rewrites, D catches the no-sidecar case) is coherent. I am asking for two things:

1. **Say it in the body**, so nobody upgrades, sees 156 unchanged, and files a regression against this PR.
2. **Reviewer:** treat this as a live question rather than settled. Is the stale-sidecar interaction worth a cell — a sidecar fixture carrying a bare-`Route` entry with `reachable: false`, asserting what the handler reports before and after a re-run? My mutant says that cell does not exist today. It is the difference between the fix being *correct* and the fix being *observably effective on the population it was filed for*, which is the distinction #6533 turned on.

This does not change my read that the PR is sound. It changes what the PR should claim.

Refs #6902, #6839, #6533.

cajasmota:

Reviewed at `73a53bce6` in my own detached worktree. Everything below was re-derived from
source and re-run, not taken from the PR body. Counts only — no corpus identifiers.

---

### Verdict on the three contradictions with the issue

**1. `internal/enrichment` is clean — the issue body is wrong, the PR is right.** Read all
three sites myself:

- `candidates.go:98` `kindBaseScore` — `case "http_endpoint", "http_endpoint_definition", "HTTPEndpoint", "SCOPE.HTTPEndpoint", "Route", "SCOPE.Route":` — both spellings, one arm.
- `candidates.go:433` `qualifyHTTPKinds` — map holds `"Route"` **and** `"SCOPE.Route"`.
- `pricing.go:38` `TokensPerEntity` — `case "http_endpoint", "HTTPEndpoint", "SCOPE.HTTPEndpoint", "Route", "SCOPE.Route":`.

`internal/types/kinds.go` (the `Route` twin note) cites `candidates.go:98` and `pricing.go:38`
as the *evidence* that the pair is one concept. Nothing to fix. **The issue's site list needs
correcting.**

**2. The fifth site is real and in scope.** `mcp/dead_code.go:228` `frameworkEntryKindsMCP`
exists, its own comment declares it a mirror of `links`' `frameworkEntryKinds`, and the parent
commit's map has `"SCOPE.Route"` with no `"Route"`. It is the same defect in the same seed
concept, one hop from site C, and it is the site the harm figure lands on — including it is
correct, not scope creep. Excluding it would have shipped a fix that is inert for the tool
users actually call.

**3. The corpus figure is real — I reproduced 156 exactly, independently.** Read-only, straight
from the on-disk sidecar with my own script:

| kind | entries | persisted unreachable |
|---|---|---|
| `http_endpoint_definition` | 2,061 | 0 |
| `SCOPE.Route` | 30 | 0 |
| bare `Route` | **575** | **156** |

Byte-for-byte the PR's table. Two follow-on checks the brief asked for:

- **The sidecar preference is real and unconditional.** `handleDeadCode` takes the recompute
  path only when `from != ""`, or when `loadReachabilitySidecar` fails (missing/unparseable
  file). With a sidecar present and no `from`, every unreachable entry is projected verbatim
  into `dead_code`, `matchesBareKind` passes on an empty `kind_filter`, and `total`/`unreachable`
  report the full count regardless of the default `limit: 200`. So yes — those 156 are served
  to an agent as dead code today.
- **The 156 are route-shaped.** 143/156 names are URL paths; 156/156 distinct on
  `(repo, name, source_file)` — **no duplicates**; zero under `vendor/`, `node_modules/` or
  `third_party/`; source extensions 142 `.java`, 7 `.py`, 5 `.yaml`, 2 `.go`; all carry an
  empty `entry_source`, i.e. nothing seeded them. A count of the right kind of thing.

  Two honest qualifications on the headline wording: **32 of the 156 sit under a `test`/`tests`
  path segment** (Java test-source trees), so "156 real HTTP routes" is nearer ~124 production
  routes plus 32 test-tree routes — all still route entities the seed set should cover, but the
  adjective is slightly over-claimed. And all 156 come from **one repo**, so this is one
  framework's route surface, not a corpus-wide spread.

---

### Mutant scorecard — 29/29 reproduced DEAD, plus 6 of my own

Baseline first: `go vet` clean, `go test -count=1` green on `./internal/dashboard/` (45.7s),
`./internal/links/` (13.7s), `./internal/mcp/` (121.1s). Every mutant: exact-string edit with
**pre-image count asserted `== 1`** and post-image asserted present before scoring, `go vet`
before each run, `-count=1`, restore by `cp` from a shasum-verified backup. Scored at the call
sites (`reachAccumulator.accumulate`, `handleTestReachability`, `runReachabilityPass`,
`handleDeadCode` → `computeDeadCodeLive`), never through an extracted helper.

**Author's 29 — all DEAD, independently reproduced:** 12 site mutants
(`{A,B,C,D}{1-delete-arm, 2-repoint-drop-scope, 3-widen-any}`) and 17 assertion mutants. I used
true **inversions** (negate the condition) rather than deletions, which is the stronger check
for vacuity; all 17 still DEAD, vacuity guards included.

**Six I added (the brief's part-by-part and precise-over-widen requirements):**

| id | mutant | verdict |
|---|---|---|
| `X-{A,B}-widen-class` | add `types.EntityKindClass` to the switch | DEAD (killed by the decoy, not the Route control) |
| `X-{C,D}-widen-class` | add `"SCOPE.Class"` to the seed map | DEAD |
| `X-D-drop-map-disjunct` | `frameworkEntryKindsMCP[e.Kind] \|\| isFrameworkOrHandler(e)` → `isFrameworkOrHandler(e)` | DEAD |
| `X-D-drop-handler-disjunct` | same → `frameworkEntryKindsMCP[e.Kind]` | **ALIVE** — see below |

`X-D-drop-handler-disjunct` surviving is the *good* outcome and it settles the #6893 trap
**empirically** rather than by argument: if `isFrameworkOrHandler` could seed the control, the
new test would have failed when it was deleted. It did not. The two disjuncts are graded
separately, and the author's exoneration chain — which I re-derived line by line at
`flow_tools.go:2098-2160` and `endpoint_tools.go:86-105` — holds: `/orders-6902` fails
`routeNameRe` (`^(get|post|put|delete|patch|head|options|http)[\s:]`, the name starts `/`), is
not in `frameworkLifecycleNames`, is not a constructor (no `.`), does not start `on`+upper or
`handle`, and `classifyEndpointKind("Route")` → `endpointKindNone`. Also confirmed
`isLiveCodeKind("Route") == true`, so assertion 2 at site D is genuinely observable rather than
passing because the kind is filtered out upstream.

**All-DEAD is not truth — checked.** The fix fabricates nothing: `types.EntityKindRouteBare`
is literally `"Route"` (`kinds.go:645`, in the canonical kind list at `:857`), and 575 entities
in a real corpus sidecar carry that exact kind string. The spelling is real, and the parent
build differs on it (that is what `{A,B,C,D}1-delete-arm` demonstrate).

**Fixture isolation verified as claimed.** The links fixture has zero edges and
`FileRoot: t.TempDir()` (empty), and the mcp dead-code fixture has zero edges — so the
`HANDLES/HANDLES_SIGNAL/NAVIGATES_TO/ROUTES_TO/REGISTERS` seeding loops and the source sniffer
can seed nothing. The seed map is the only live mechanism. Confirmed by reading
`links/reachability.go:241-250` and `mcp/dead_code.go:284-296`.

Tree restored: `git status --porcelain` empty, `git diff 73a53bce6` empty, `gofmt -l` clean.

---

### Byte-identical switches — confirmed both halves, with one nit in the report

Extracting the function bodies (everything below the signature line) at the parent and at HEAD:

- **before:** both `c48fe09f9929…` — identical
- **after:** both `341c560872f9…` — identical, and a `diff` of the two whole functions at HEAD
  differs on the signature line only

So the load-bearing claim is true in both directions. **Nit:** the report's *before* digest
`16446cab…` does not reproduce — I get `c48fe09f…`. The *after* digest `341c5608…` matches
exactly, so it looks like the before-hash was taken over a different span (probably including
the doc comment, which the change rewrites). Cosmetic; the claim itself stands.

---

### `reachabilityEdgeKindsMCP` — confirmed, and the neighbouring map is **actually divergent**

Confirmed as stated: `reachabilityEdgeKindsMCP` (24 keys) and `links`' `reachabilityEdgeKinds`
(24 keys) are set-identical today, and a repo-wide grep over `*_test.go` finds **no test naming
either symbol** — the only mentions are the comments added by this PR. Agreed: own issue, not
this PR.

**But while checking that I found a live divergence in the pair this PR *does* touch.** After
the fix, the two seed maps still differ by one key:

```
links  frameworkEntryKinds     : 10 keys
mcp    frameworkEntryKindsMCP  : 11 keys  — extra "SCOPE.ChannelBinding"
```

`SCOPE.ChannelBinding` was added to the MCP copy by #5782 / ADR-0025 ("a ChannelBinding is a
config-side messaging declaration with no callers by design — never report it as dead code")
and **never landed in the link-pass original**. Because `handleDeadCode` prefers the sidecar,
ADR-0025's guarantee is defeated on every group that has a sidecar — i.e. the normal case — and
holds only on the `live_recompute` fallback. That is precisely the #6846 shape and precisely the
failure mode this PR's comments warn about, in the very map it edits.

I have **no volume evidence**: the corpus sidecar I read contains zero `ChannelBinding` entities,
so I cannot say it bites today. Structurally it is a real, separate bug. Not a blocker here —
this PR did not introduce it and fixing it would widen the diff into ADR-0025 territory — but it
belongs in the shared-list follow-up issue as a row with teeth, not a hypothetical.

---

### `internal/quality` / the extraction ratchet — cannot move, with a reason

Not "probably fine": `internal/quality` imports **none** of `internal/links`, `internal/mcp` or
`internal/dashboard` (checked by grep for the import paths — zero hits). The golden gate compares
extractor output (entities + relationships); reachability is a post-index link pass and no golden
`expected.json` or `baseline.json` carries a `reachable` field — the only `reachab*` matches in
golden data are English prose in fixture notes. Nothing in this diff can move a `must_exist` row
or the ratchet. Skipping it was correct, and the author's `cmd/grafel` digest run already covers
the one index-time file touched.

---

### Corrections to the brief itself

1. The brief inherits the issue's claim that `internal/enrichment` is a fourth defective site.
   It is not — see above. The issue body should be corrected.
2. The brief's framing that the fb-v5/v6 mismatch would block a corpus figure is wrong for
   sites C and D: the reachability sidecar is plain JSON on a `version: 1` schema of its own and
   never goes near the flatbuffer loader. The author's route around it is legitimate. (It stays
   correct for sites A and B, which still have existence-only evidence.)
3. One thing neither the brief nor the report says out loud and a reader should know: **site C's
   fix does not retroactively repair an existing sidecar.** Those 156 stay unreachable on disk
   until the group is re-indexed. The MCP `live_recompute` and `from_entry` paths are fixed
   immediately; the sidecar path — the default — is fixed at the next index. Worth a line in the
   changelog so nobody reports the fix as ineffective.

---

### Follow-ups (neither blocks this PR)

- The shared-list issue the PR recommends, with `SCOPE.ChannelBinding` added as a row that is
  *already* divergent rather than merely at risk, plus `reachabilityEdgeKindsMCP` as the third
  hand-copy with no equality assertion.
- `isFrameworkOrHandler` as a disjunct of `computeDeadCodeLive`'s seeding is **ungraded**:
  deleting it leaves the entire `./internal/mcp/` suite green (`-count=1`, 0 `--- FAIL`, rc 0).
  Distinguishable, not equivalent — a `handleX`-named operation with no inbound edge is seeded
  only by that disjunct. Pre-existing, not introduced here.

**APPROVE.** The premise is confirmed, the headline number reproduces exactly, the site list is
better than the issue's, the tests grade the recall *and* the forbidden direction at every site
with controls nothing else in the predicate can match, and the mutant table survives being
re-run by someone trying to break it.

PR #6908, opened 2026-09-06
fix(#6776): arm B9 — bare `Endpoint` is a produced kind that is nobody's synonym, so it enters the enum as its own member and #6744's ledger is retired, not ratcheted to zero

### LINKED ISSUE #6776: DONE: rule-declared kind ledger retired 532→0 sites — IsValidEntityKind was the target, and no rename or stored-graph migration was needed

cajasmota:
## What this does

`Endpoint` — the last entry on #6744's rule-YAML ledger — enters `types.AllEntityKinds()` as `EntityKindEndpointBare`, and that ledger is **retired**, not ratcheted to zero. This closes #6776's B series: `ruleDeclaredKindsDeferredMax` goes **1 → 0**, and every kind `internal/engine/rules/**/*.yaml` declares is now a kind `IsValidEntityKind` accepts.

Refs #6776. Refs #6820.

## The disposition, and why it is not a synonym migration

Every pair migrated in arms B5–B8 was a *same-concept, different-spelling* pair, and each `Bare` constant's doc says so. **This one is not**, and the change does not pretend otherwise.

Bare `Endpoint` is an **Electron IPC channel** — three sites, all in `internal/engine/rules/javascript_typescript/frameworks/electron.yaml:41,46,52` (`ipcMain.handle` / `ipcRenderer.invoke` / `contextBridge.exposeInMainWorld`). `SCOPE.Endpoint` is the HTTP entrypoint. They share seven letters and nothing else; an IPC channel never had a URL.

So the argument for membership is **not** "it is the same thing as SCOPE.Endpoint". It is #6776's own ruling, which is about producers rather than about vocabulary design:

- `internal/engine/detector.go` writes `SourcePattern.EntityType` **verbatim** into `EntityRecord.Kind`. Those three sites put the string `Endpoint` into every Electron graph whether or not the enum carries it.
- The only question membership answers is whether `IsValidEntityKind` then rejects an entity grafel itself produced. #6776 exists to make that answer "no".
- A distinct concept is a reason for a **distinct member** — which is what this is, alongside `SCOPE.Endpoint`, both spellings kept — not a reason to keep a produced kind outside the vocabulary. No rename, no stored-graph migration, consistent with the standing ruling on both issues.

**The naming debt this incurs, stated rather than denied.** With both spellings in the enum meaning *different* concepts, the `SCOPE.` prefix becomes semantically **load-bearing** for this one pair — which contradicts the framing arms B4–B8 ran on, and which the retired ledger family note still states: the un-prefixed spelling is *"an accident, not a namespace"*. Every earlier arm could say that honestly because both spellings meant the same thing; this one cannot. It is a documented naming debt rather than a namespace admitted by the back door, because the honest alternative for a distinct concept is a distinct **name** (`IPCChannel`) — and the owner priced that and **declined** it in #6820 on 2026-09-06: *"No rename, no stored-graph migration… enum membership is the target, not re-spelling kinds."* With the rename off the table the only two options are *member* or *permanently-invalid produced kind*, and the second is what #6776 exists to end. Recorded here so the older "prefix is an accident" comments do not mislead the next reader; a follow-up issue tracks it.

**Why it is safe now and was not at arm B8.** #6820 was decided on 2026-09-06 and shipped (`9706a5cde`, plus `fa05d3e2c` for the `Route` twin): the ten consumer sites were repointed at `SCOPE.Endpoint`, so no consumer reads the bare spelling as HTTP. The hazard B8 named — a misleading vocabulary name with IPC channels reachable through HTTP-shaped consumers — is closed at the consumers. B8's own words were that reaching zero "was never a goal" and one honestly-justified entry is a legitimate end state; the justification is what expired, not the arm's judgement.

**`internal/types/http_endpoint_kind_forbidden_6894_test.go` is untouched in its assertions.** Only its comment beside the `Endpoint` row changed, and it changed to say the row matters *more* now: "it is in the vocabulary" is precisely the argument that would fold the kind into `IsHTTPEndpointKind` and re-export IPC channels into the OpenAPI document. That row is now written as `string(EntityKindEndpointBare)` so the constant and the prohibition are visibly the same string.

## Retiring the ledger, rather than ratcheting it to zero

Three guards treat an empty ledger as evidence of a broken scan, and they were right to while the ledger was the only evidence. The retirement replaces that premise with a **stronger, positive one** instead of deleting it.

New: `internal/entkinds/rule_ledger_retired_6776_test.go`.

- `TestRuleDeclaredLedger6776_IsRetiredBecauseThePopulationIsEmpty` — scans the rule tree with `ScanRuleYAML` and asserts **zero** of its sites names a kind outside the enum, over a **floor** (`minRuleYAMLSites = 300`) that catches a collapsed scan, plus `Unresolved() == 0`. It carries a **positive control on the predicate itself** (`IsValidEntityKind` must still reject an invented kind), and it pins `EntityKindEndpointBare`'s three `electron.yaml` sites **line-exactly**. Then it asserts the ledger, the ratchet and the family map are all empty. Emptiness stops being an assumption and becomes a measurement with its own guard.

**What that measurement is stronger and weaker than — corrected.** An earlier revision of this body said "strictly stronger than any ledger row". **"Strictly" was wrong**, and review mutant R11 is what showed it:

- **Stronger** in the direction the ledger was built for: any kind outside the enum, anywhere in the rule tree, fails here with a line-exact diagnosis and no row available to absorb it.
- **Weaker**, on its own, in the direction the ledger held *incidentally*: a row named a kind, so the sweep noticed when that kind's sites disappeared. "Is this set empty" cannot notice that a **member's producers vanished**. Re-spelling electron.yaml's three `entity_type: Endpoint` rows to `SCOPE.Endpoint` — the silent rename #6820 rejected — left all six packages **green** on the first revision of this branch, and fails **three** tests on the parent `37f39ca57`.

The site pin is the repair, and it restores the coupling deliberately rather than incidentally: it is asserted as an exact `file:line` set, not a count, because a count alone would accept the three sites moving to a file that has nothing to do with Electron IPC — the concept the member is named for.
- `TestRuleDeclaredLedger6776_EveryFamilyIsUsedByARow` — the reverse coupling that `TestYAMLHalfObservesDeclaredKinds`'s `perFamily` loop can no longer provide (its body is now unreachable). An entry needs a family (checked there); a family needs an entry (checked here).
- `enum_member_corresponds_6818_test.go:447`'s `t.Fatal` on an empty ledger becomes a `t.Logf`, with the replacement named in the comment. Its row loop still grades a ledger that comes back — proven by mutant M4.
- The `perFamily` loop is **kept, not deleted**, with its unreachability stated: it fires again the moment a family is reintroduced with no live YAML site behind it (proven by mutant M8).
- **A FOURTH guard went vacuous, and it is named here rather than left to be tripped over:** `TestLedgerIsDescribedByItsOwnScan` (`rule_declared_kinds_sweep_guard_6744_test.go`) loops over the now-empty ledger, so it is a green over zero entries. It is **kept without a floor**, because the only floor available — "the ledger is non-empty" — is exactly the assertion this arm falsifies. What it actually observed, *"a kind this file names resolves to a located site"*, is not lost: it moved into the new test's line-exact `electron.yaml` pin, made about the one kind that still needs it (the member admitted on the strength of those sites) instead of about a ledger row that no longer exists. The test now carries that paragraph in its own doc comment. It remains as the drift guard that fires again the moment a row comes back.

That is **four** guards the retirement relaxed, not three. The whole risk of this arm is anti-vacuity guards being relaxed by the change they exist to catch, so the list is exhaustive on purpose: `:447`'s empty-ledger fatal, the `perFamily` traffic loop, fbwriter's ledger transcription, and `TestLedgerIsDescribedByItsOwnScan`.

## Evidence — `ScanRuleYAML`, never grep

Re-derived from disk on this branch, not from the brief:

```
yaml files=732  sites=532  unresolved=0
INVALID KINDS: 1
  Endpoint: 3
     internal/engine/rules/javascript_typescript/frameworks/electron.yaml:41
     internal/engine/rules/javascript_typescript/frameworks/electron.yaml:46
     internal/engine/rules/javascript_typescript/frameworks/electron.yaml:52
```

(The brief said 714 YAML files; the live scan says **732**. Sites and the invalid set match.) `ScanGo` reports **0** Go sites for `Endpoint` — see the Go ratchet below.

## Counters, each re-derived from disk

| counter | before | after | why |
|---|---|---|---|
| `ruleDeclaredKindsDeferredMax` | 1 | **0** | retired; the row is now an enum member |
| `ruleDeclaredKindsDeferred` | 1 row | **0 rows** | ditto |
| `ruleDeclaredFamily` | 1 family | **0** | a family with no row is a slot inviting a decision-free row |
| EntityKind constants pin | 93 | **94** | one constant added |
| constants named `EntityKind*` | 92 | **93** | ditto |
| `pinnedEntityKindVocabulary` | — | `+"Endpoint"` | added, not renamed or retired |
| `KindVocabularyVersion` | 1 | **1** | its guard bumps only when `len(removed) > 0`; this removes nothing |
| `goUnprefixedKindsDeferredMax` | 5 | **5** | `Endpoint` is not on the Go ledger — **graded by mutant M7**, not assumed |
| `enumOnlyEntityKindsMax` | 1 | **1** | `Endpoint` has three rule-YAML sites, so it is corroborated; no allow-list row |

`assertLedgerExact` derives its population from the site scanners and is coupled to `IsValidEntityKind` by construction — M7 is exactly that coupling firing.

## The four (actually five) re-picked fbwriter controls

B8's guard messages said to re-pick from `internal/entkinds`' ledger. **That advice expires with this change** — the ledger is now empty and cannot be drawn from. Every replacement comes from `internal/types`' `goUnprefixedKindsDeferred`, the Go half of the same accident, which survives precisely because it is *not* rule-declared and nothing proposes to move it. The guard messages have been rewritten to say so.

| fixture | slot | new control | provenance |
|---|---|---|---|
| `TestStreamingWriterTalliesOnlyEntityKindsAbsentFromTheEnum` (:101) | 2 controls | `ChannelEvent` + `File` | `engine/websocket_edges.go`, `engine/commit_coupling_edges.go` |
| `TestWriteGraphGenReportWiresTheFlatEntityProducerPath` (:265) | **sole** | `Stream` | `engine/sse_edges.go` |
| `TestEntityKindsAreCountedNotDropped` (:364) | 2 controls | `Subscription` + `File` | `engine/graphql_subscriptions.go` |
| `TestEntitySummaryIsSeparableFromTheRelationshipSummary` (:463) | **sole** | `ChannelEvent` | as above |
| `TestSCOPEPrefixedKindsOutsideTheEnumAreStillCounted` (:590) | **sole** un-prefixed | `File` | as above |

**That is FIVE fixtures, not four.** B8's estimate named four (two the header enumerated plus `:267` and `:607`). `TestEntitySummaryIsSeparableFromTheRelationshipSummary` is a fifth: it carries no explicit validity guard, but its `fixture is inert` check on a non-empty `EntitySummary()` fires the moment its control goes valid — confirmed by mutant M6b, which is exactly the failure mode. Also updated for coherence, though neither is validity-gated: the nil-tally `observeEntity` call (:212) and the `ApplyToSidecar` report literals (:416/:435).

`ruleDeclaredKinds6776` — the ledger transcription — and `TestEveryRuleDeclaredKindOnTheLedgerIsCountedByTheWritePath` are replaced rather than deleted. Its own floor said "if the ledger really has reached zero, delete this test and say so". Deleting outright would have thrown away the **property** along with the population, and the property is what arm A's measurement rests on: a kind invisible to the write path reports a zero meaning "not measurable" rather than "not produced". So the population is re-aimed at `nonEnumProducerKinds6776` (the four producer rows of `goUnprefixedKindsDeferred`; `relationship` is excluded because its own ledger comment calls it unreachable) and the mechanism — floor, exact pin, per-kind assertion — is kept intact as `TestEveryNonEnumProducerKindIsCountedByTheWritePath`.

- **`endpoint_bare_kind_6776_test.go`** — *varies*: the spelling (bare vs `SCOPE.`-prefixed) and the vocabulary it is tested against (entity kinds vs HTTP endpoint kinds). *Holds constant*: both validators, and that both spellings name a really-produced kind. Positive control on `IsHTTPEndpointKind` so its rejections are not satisfied by a function that returns false for everything.
- **`rule_ledger_retired_6776_test.go`** — *varies*: nothing; it is a population floor plus an emptiness pin. *Holds constant*: the live scan of the shipped rule tree.
- **Re-picked fbwriter fixtures** — each keeps its original varied axis (kind membership / producer path / prefix) and holds its neighbour constant; the swap changes only the identity of the non-enum control, never what the fixture varies. The two-slot fixtures keep their existing "the two slots must differ" guard.

## Mutant table

Every edit's match count was asserted **after** application (an edit that silently fails to apply reads as ALIVE and argues the opposite). All scored at the call site with `-count=1`, restored by `cp` from shasum-verified backups.

|---|---|---|---|
| M1 | drop `EntityKindEndpointBare` from the `AllEntityKinds()` roster (keep the const) | **DEAD** | 6 tests / 2 pkgs: roster-exactly-once, declarations-match, both new `EndpointBare` tests, `TestNoUndeclaredRuleEntityKinds`, the new retired-ledger test |
| M2 | `EntityKindEndpointBare = "SCOPE.Endpoint"` (migrate it *as a synonym* — the thing this arm must not do) | **DEAD** | vocabulary pin, roster duplicate-value guard, both `EndpointBare` tests, both entkinds population tests |
| M3 | add `case string(EntityKindEndpointBare)` to `IsHTTPEndpointKind` | **DEAD** | `TestIsHTTPEndpointKind_ForbiddenRows`, the new `MembershipIsNotHTTPMembership`, and **four #6820 dashboard tests** (`TestPathsPanes_ExcludeElectronIPCChannels`, `TestSearchPaths_…`, `TestPathDetail_…`, `TestGroupTopFrameworks_…`) |
| M4 | reinstate the ledger row + ratchet + family, keeping the enum change (undo only the retirement) | **DEAD** | `TestEnumEntityKinds6818_NoDeferredLedgerOverlapsTheEnum` (its row loop still grades), `TestNoUndeclaredRuleEntityKinds` (stale-entry half), the new retired-ledger test |
| M5 | *positive control*: point the new test's scan at `t.TempDir()` | **FLOOR FIRES** with the right diagnosis — "produced only 0 site(s) from 0 file(s) (floor: 300)". It does **not** pass on "zero invalid kinds", which is the failure this floor exists to prevent |
| M6 | swap all four validity-gated re-picked controls to the valid `Model` | **DEAD** | all 5 fixtures' `fixture is inert` guards, each naming the stale control |
| M6b | swap `:463`'s control to `Model` (the fixture with no explicit validity guard) | **DEAD** | its `fixture is inert: Summary()=… EntitySummary()=""` check — this is what makes it the fifth fixture |
| M7 | put `Endpoint` on `goUnprefixedKindsDeferred` with the pin raised to 6 | **DEAD** | `TestProducerEntityKinds6776_UnprefixedLedgerIsExact`: *"ledgered kind \"Endpoint\" is no longer declared by any Go producer"* — this is the grading of the "the Go ratchet does not move" claim |
| M8 | add a family to `ruleDeclaredFamily` with no ledger row | **DEAD** | `TestYAMLHalfObservesDeclaredKinds` (the loop kept for exactly this), the retired-ledger test, and the new `EveryFamilyIsUsedByARow` |

### Review round — R10 / R11, and one correction to the review's own reasoning

| # | mutant | verdict |
|---|---|---|
| R11 | re-spell electron.yaml's three `entity_type: Endpoint` rows to `SCOPE.Endpoint` (match count asserted = 3) | was **SURVIVOR**, now **DEAD** — the new line-exact site pin: `got [] want [electron.yaml:41 :46 :52]` |
| R10 | `IsValidEntityKind` returns true for any non-empty string | was **partial** (`internal/entkinds` green), now **DEAD** in `internal/entkinds` too: *"fixture is inert: IsValidEntityKind(\"SCOPE.ZZNotAKindAnyProducerWrites\") is true, so it accepts everything"* |
| R10b | *control on the repair*: apply the R10 mutant with the site pin present but the predicate control removed | **`internal/entkinds` GREEN, exit 0** |

**R10b is a correction to the review.** The review states the site block *"also closes a second gap… one fix for two findings"*. **It does not.** A site-level positive fact is satisfied whether the validator is honest or constant-true — the sites exist either way — so with the pin in place and the one-line control removed, the always-true validator sails through this package. The two findings needed two edits, and the reviewer's own cheaper alternative (the explicit `fixture is inert` line) is the load-bearing half of the R10 fix, not the fallback. Both are in.

No survivors, so there is no equivalent/unreachable verdict to defend.

Two observations worth recording rather than leaving to be rediscovered:

- Under **M1**, `./internal/graph/fbwriter` stays **green**. That is the re-pick working as intended: after this change no fbwriter fixture depends on `Endpoint`'s enum status in either direction.
- **M6b** is the one B8's cost estimate missed, and it is the cheap direction to get wrong: a fixture whose inertness is detected through an empty summary rather than a named `IsValidEntityKind` call.

## Verification

Run, all `-count=1`, all green:

- `go test -count=1` on `./internal/entkinds/`, `./internal/types/`, `./internal/graph/fbwriter/`, `./internal/engine/`, `./internal/dashboard/`, **`./cmd/grafel/`**.
- `./cmd/grafel/` is the mandatory leg for any kind change (`custom_extractor_spans_6118_test.go` digest-pins the whole graph and nothing in the package tree points at it). **The digest did not move**, and that is the expected result rather than a lucky one: the digest hashes `Kind|Name|QualifiedName|Subtype|SourceFile|StartLine|EndLine`, and enum membership changes no entity's `Kind` string — `detector.go` was already writing `Endpoint` verbatim. No constant was bumped.
- **`scripts/quality/run.sh --ratchet`** — `quality ratchet OK — 31 gated fixtures held their baseline (22 at 100% must-have recall)`, exit 0.
- `gofmt -l internal/` clean; `go vet` clean on all touched packages.
- Review round: `go test -count=1` re-run on `./internal/entkinds/`, `./internal/types/`, `./internal/graph/fbwriter/` and `./cmd/grafel/` after the fix.

## What I did NOT run

- **No full corpus index**, so **no `entity_entities_kind_not_in_enum` figure is quoted.** B6 and B7 quoted one; B8 did not, and neither does this. The expected drop is the Electron IPC population, and #6893 measured that at **0 across the corpus** (no indexed group contains `ipcMain`/`ipcRenderer`/`contextBridge`) — but that is a corpus-relative zero and an *inference* about this arm, not a measurement of it. It should be measured before anyone repeats it as a number.
- No full-repo `go test ./...`. Package selection is not suite selection, and the set above is chosen, not derived from the diff: `./cmd/grafel/` and `./internal/dashboard/` are in it precisely because no diff-derived set would reach them.

### DIFF
--- a/internal/entkinds/enum_member_corresponds_6818_test.go
+++ b/internal/entkinds/enum_member_corresponds_6818_test.go
@@ -435,6 +435,19 @@ func TestEnumOnlyEntityKinds6818_AllowListDoesNotRot(t *testing.T) {
 // an entry the enum also carries, "non-deferred enum member" stops being a
 // synonym for "enum member" and the sweep needs the exclusion it currently does
 // without.
+//
+// # THE ANTI-VACUITY FATAL ON AN EMPTY LEDGER IS GONE, AND WAS NOT SIMPLY DROPPED
+//
+// This test used to t.Fatal when ruleDeclaredKindsDeferred was empty, on the
+// premise that an empty ledger cannot be told apart from a scan that stopped
+// looking. #6776 arm B9 retired the ledger — the last row, `Endpoint`, is now
+// an enum member — so that premise had to be either honoured (leaving the arm
+// unable to finish) or REPLACED. It is replaced, by a stronger statement of the
+// same thing: rule_ledger_retired_6776_test.go asserts, against a live scan
+// with a floor under it, that ZERO of the 532 rule-YAML sites names a kind
+// outside the enum. Emptiness is now a measured fact with its own guard, so
+// this test asserts it rather than refusing it — a ledger that comes BACK is
+// then the anomaly, and the row loop below still grades it.
 func TestEnumEntityKinds6818_NoDeferredLedgerOverlapsTheEnum(t *testing.T) {
 	inEnum := map[string]bool{}
 	for _, k := range enumEntityKindStrings() {
@@ -443,8 +456,11 @@ func TestEnumEntityKinds6818_NoDeferredLedgerOverlapsTheEnum(t *testing.T) {
 	if len(inEnum) == 0 {
 		t.Fatal("types.AllEntityKinds() is empty; the disjointness below would hold vacuously")
 	}
-	if len(ruleDeclaredKindsDeferred) == 0 {
-		t.Fatal("ruleDeclaredKindsDeferred is empty; the disjointness below would hold vacuously")
+	if len(ruleDeclaredKindsDeferred) != 0 {
+		t.Logf("ruleDeclaredKindsDeferred has %d row(s); #6776 arm B9 retired it, so the loop "+
+			"below is grading a ledger that came back — check "+
+			"TestRuleDeclaredLedger6776_IsRetiredBecauseThePopulationIsEmpty first",
+			len(ruleDeclaredKindsDeferred))
 	}
 	for k := range ruleDeclaredKindsDeferred {
 		if inEnum[k] {
--- a/internal/entkinds/rule_declared_kinds_sweep_guard_6744_test.go
+++ b/internal/entkinds/rule_declared_kinds_sweep_guard_6744_test.go
@@ -13,13 +13,23 @@ package entkinds_test
 // validation, so a rule file can mint any string it likes.
 //
 // The live scan of the rule tree finds 532 declaration sites and 27 distinct
-// values. Exactly TWO of them are valid entity kinds — `Module` (valid only
+// values. Exactly TWO of them were valid entity kinds — `Module` (valid only
 // because types.EntityKindModule is itself un-prefixed) and `SCOPE.IngressHost`
-// (added by this change). The other 25 were ledgered below; #6776 arms B5-B8
-// declared twenty-four of them in types.AllEntityKinds() with their un-prefixed
-// spellings unchanged (thirteen at B5, four at B6, four at B7, three at B8).
+// (added by this change). The other 25 were ledgered below; #6776 arms B5-B9
+// declared ALL of them in types.AllEntityKinds() with their un-prefixed
+// spellings unchanged (thirteen at B5, four at B6, four at B7, three at B8, one
+// at B9).
 //
-// DO NOT RESTATE THE REMAINDER AS A NUMBER HERE — it went stale three arms
+// # THE LEDGER IS RETIRED, and this is the one number safe to state here
+//
+// Not "ratcheted to zero": the POPULATION is empty, and that is measured rather
+// than inferred from an empty map. rule_ledger_retired_6776_test.go scans the
+// rule tree and asserts, over a floor that catches a collapsed scan, that not
+// one of the 532 sites names a kind AllEntityKinds() rejects. Every guard in
+// this file that treated an empty ledger as evidence of a broken scan was
+// correct to, while the ledger was the only evidence; it is not any more.
+//
+// DO NOT RESTATE A REMAINDER AS A NUMBER HERE — it went stale three arms
 // running before this sentence replaced it. ruleDeclaredKindsDeferred below is
 // the population, and it is pinned by exact set equality in both directions
 // plus ruleDeclaredKindsDeferredMax, so reading it is always current and
@@ -78,6 +88,13 @@ package entkinds_test
 // migration is triaged — anything new that is not on the ledger fails,
 // immediately, naming the file and line.
 //
+// That stance was about the ORDER of the work, not about never finishing it,
+// and #6776 kept it: the population was published, ranked and migrated arm by
+// arm over B5-B9, each arm naming its sites, and only then did the ledger go
+// empty. With the ratchet at 0 the sweep is now strictly stricter than it was
+// — any rule-declared kind outside the enum fails, with no ledger row
+// available to absorb it.
+//
 // # The ratchet, and why it is spelled this way
 //
 // ruleDeclaredKindsDeferredMax pins the ledger's EXACT size, so an author who
@@ -117,9 +134,18 @@ import (
 // Every entry was produced by the live scan, not transcribed from the issue —
 // the issue named three sites and the scan found 532, of which 25 distinct
 // values are invalid. This list must only ever SHRINK, and that is ENFORCED.
-var ruleDeclaredKindsDeferred = map[string]string{
-	"Endpoint": "rule_namespace", // 3 sites; javascript_typescript/frameworks/electron.yaml:41
-}
+// RETIRED, and EMPTY is the finished state rather than a hole. #6776 arm B9
+// declared the last entry (`Endpoint`, the Electron IPC kind) in
+// internal/types/kinds.go, and the population this ledger stood in for is now
+// measured directly by rule_ledger_retired_6776_test.go: 532 rule-YAML sites,
+// zero naming a kind AllEntityKinds() rejects. That measurement is what makes
+// the emptiness readable as "nothing left to defer" rather than "the scan
+// stopped looking", which is the only reading an empty map could otherwise
+// carry.
+//
+// The map is kept, not deleted, so the next drift lands on a mechanism instead
+// of inventing one — and so the ratchet below can hold it at zero.
+var ruleDeclaredKindsDeferred = map[string]string{}
 
 // ruleDeclaredKindsDeferredMax is the RATCHET on ruleDeclaredKindsDeferred: the
 // EXACT number of entries the ledger is allowed to hold.
@@ -135,23 +161,23 @@ var ruleDeclaredKindsDeferred = map[string]string{
 //   - SHRINKS (a kind was declared or removed, which is the point) → this fires
 //     and requires the constant to come down with it, so the bar is never left
 //     slack for a later append to slip under.
-const ruleDeclaredKindsDeferredMax = 1
+const ruleDeclaredKindsDeferredMax = 0
 
 // ruleDeclaredFamily explains each family tag. A ledger entry without a stated
 // reason is not a decision, it is a silence.
+// EMPTY because the ledger is. The one family it held, `rule_namespace`,
+// explained the un-prefixed rule-YAML spelling as an accident rather than a
+// second namespace; #6776 acted on that reading across arms B4-B9 and there is
+// now no deferred kind for it to explain. It is deliberately not left standing
+// with no rows: TestYAMLHalfObservesDeclaredKinds requires live YAML traffic
+// for every family here, so an unused family is either a red suite or, once
+// that check goes unreachable, a documented slot inviting a row with no
+// decision behind it. TestRuleDeclaredLedger6776_EveryFamilyIsUsedByARow is the
+// standing check on that direction.
 var ruleDeclaredFamily = map[string]struct {
 	Origin string
 	Why    string
-}{
-	"rule_namespace": {
-		Origin: entkinds.OriginRuleYAML,
-		Why: "an un-prefixed name declared by internal/engine/rules/**/*.yaml. " +
-			"internal/engine/detector.go:411 writes SourcePattern.EntityType straight into " +
-			"types.EntityRecord.Kind with no validation, so the string reaches the graph exactly " +
-			"as spelled. The un-prefixed spelling is an accident, not a namespace — see the file " +
-			"header. Fixing it is a ~530-site migration filed separately.",
-	},
-}
+}{}
 
 func repoRoot(t *testing.T) string {
 	t.Helper()
@@ -523,6 +549,15 @@ func TestYAMLHalfObservesDeclaredKinds(t *testing.T) {
 	// Positive, per-mechanism traffic for the ledger itself: the YAML half must
 	// carry ledgered sites, or the exact-set comparison above is measuring an
 	// empty set against an empty set.
+	//
+	// UNREACHABLE TODAY, and deliberately kept: ruleDeclaredFamily is empty
+	// since #6776 arm B9 retired the ledger, so this loop drives nothing. It is
+	// the check that fires again the moment a family is reintroduced with no
+	// live YAML site behind it. What it can no longer do — object to a family
+	// with no ledger ROW — is carried by
+	// TestRuleDeclaredLedger6776_EveryFamilyIsUsedByARow, and the sentinels
+	// above still hold the YAML half of the scan to real, located traffic
+	// without depending on the ledger at all.
 	perFamily := map[string]int{}
 	for _, s := range yamlSites(res) {
 		if fam, ok := ruleDeclaredKindsDeferred[s.Kind]; ok {
--- a/internal/entkinds/rule_declared_kinds_sweep_guard_6744_test.go
+++ b/internal/entkinds/rule_declared_kinds_sweep_guard_6744_test.go
@@ -815,6 +815,24 @@ func TestBoundPathsMirrorEngineSchema(t *testing.T) {
 // TestLedgerIsDescribedByItsOwnScan keeps this file honest about how it was
 // built: every ledger entry names a real, currently-scanned site with a
 // location, so the failure text a future author sees can point at one.
+//
+// FULLY VACUOUS TODAY, disclosed rather than left to be tripped over. #6776 arm
+// B9 emptied ruleDeclaredKindsDeferred, so this loop drives nothing and this
+// test is a green over zero entries. It is the FOURTH guard the retirement
+// relaxed, alongside enum_member_corresponds_6818_test.go's empty-ledger fatal,
+// the perFamily traffic loop above, and internal/graph/fbwriter's ledger
+// transcription.
+//
+// It is kept, and NOT given a floor, because the only floor available —
+// "the ledger is non-empty" — is the assertion arm B9 deliberately falsified.
+// What it actually observed, "a kind this file names resolves to a located
+// site", is not lost: it moved to
+// TestRuleDeclaredLedger6776_IsRetiredBecauseThePopulationIsEmpty, which pins
+// EntityKindEndpointBare's three electron.yaml sites LINE-EXACTLY. That is the
+// same property, made about the one kind that still needs it — the member
+// admitted on the strength of those sites — instead of about a ledger row that
+// no longer exists. This test remains as the drift guard that fires again the
+// moment a row comes back.
 func TestLedgerIsDescribedByItsOwnScan(t *testing.T) {
 	res := scanRepo(t)
 	for kind := range ruleDeclaredKindsDeferred {
--- a/internal/entkinds/rule_ledger_retired_6776_test.go
+++ b/internal/entkinds/rule_ledger_retired_6776_test.go
@@ -1,0 +1,122 @@
+package entkinds_test
+
+// rule_ledger_retired_6776_test.go — #6776 arm B9 retires #6744's ledger.
+//
+// # Why a RETIREMENT needs its own guard, and is not "the ratchet reached zero"
+//
+// Every anti-vacuity guard in this area is built on the same premise: an empty
+// ledger is indistinguishable from a scan that stopped looking. That premise is
+// correct while the ledger is the only evidence, and it is the reason
+// TestEnumEntityKinds6818_NoDeferredLedgerOverlapsTheEnum used to t.Fatal on an
+// empty map.
+//
+// The population, not the ledger, is what actually reached zero: the rule-YAML
+// half of the scan declares 532 sites and NOT ONE of them names a kind
+// types.AllEntityKinds() rejects. That is a positive, measured statement, and
+// it is strictly stronger than any ledger row — it is what the ledger was a
+// stand-in for. So this file asserts the population directly, and the empty
+// ledger becomes a consequence rather than an assumption.
+//
+// Varies: nothing — this is a population floor plus an emptiness pin.
+// Holds constant: the live scan of the shipped rule tree.
+
+import (
+	"fmt"
+	"sort"
+	"strings"
+	"testing"
+
+	"github.com/cajasmota/grafel/internal/entkinds"
+	"github.com/cajasmota/grafel/internal/types"
+)
+
+// minRuleYAMLSites is a slack FLOOR under the rule-YAML half, not a count.
+// Measured on this branch with entkinds.ScanRuleYAML: 732 YAML files, 532
+// sites, 0 unresolved. The floor catches a scan that COLLAPSES — which is the
+// only way "zero invalid kinds" could become a lie — and deliberately does not
+// track the churn of the rule tree.
+const minRuleYAMLSites = 300
+
+// TestRuleDeclaredLedger6776_IsRetiredBecauseThePopulationIsEmpty
+//
+// This is the assertion that replaces the ledger. Read it as: the thing the
+// ledger recorded no longer exists, and here is the measurement, not the
+// absence of a measurement.
+func TestRuleDeclaredLedger6776_IsRetiredBecauseThePopulationIsEmpty(t *testing.T) {
+	res, err := entkinds.ScanRuleYAML(repoRoot(t))
+	if err != nil {
+		t.Fatalf("ScanRuleYAML: %v", err)
+	}
+
+	// FLOOR FIRST. "No invalid kind was found" is the same sentence for a live
+	// scan and for a scan that read nothing, and #6534 is this repo's standing
+	// example of the second one shipping as the first.
+	if len(res.Sites) < minRuleYAMLSites {
+		t.Fatalf("the rule-YAML scan produced only %d site(s) from %d file(s) (floor: %d). "+
+			"An emptied scan reports every rule file clean; the zero below would then mean "+
+			"'nothing was read', not 'nothing is invalid'.",
+			len(res.Sites), res.YAMLFilesParsed, minRuleYAMLSites)
+	}
+	if res.Unresolved() != 0 {
+		t.Errorf("%d rule-YAML site(s) are unresolved; a site whose kind could not be read is "+
+			"outside the sweep below and must not be counted as clean", res.Unresolved())
+	}
+
+	invalid := map[string][]string{}
+	for _, s := range res.Sites {
+		if types.IsValidEntityKind(s.Kind) {
+			continue
+		}
+		invalid[s.Kind] = append(invalid[s.Kind], fmt.Sprintf("%s:%d", s.File, s.Line))
+	}
+	if len(invalid) > 0 {
+		var lines []string
+		for k, where := range invalid {
+			lines = append(lines, fmt.Sprintf("%s (%d site(s)): %s", k, len(where), strings.Join(where, ", ")))
+		}
+		sort.Strings(lines)
+		t.Errorf("the rule-YAML half declares %d entity kind(s) types.AllEntityKinds() rejects:\n  %s\n\n"+
+			"#6744's ledger was RETIRED by #6776 arm B9 on the strength of this set being empty. "+
+			"Do not reinstate it to hold a new entry: declare the kind in internal/types/kinds.go, "+
+			"which is what every arm B5-B9 did.",
+			len(invalid), strings.Join(lines, "\n  "))
+	}
+
+	// And the ledger itself must be gone, in BOTH of its halves — a retired
+	// ledger that keeps a family tag is a ledger with room for a row.
+	if len(ruleDeclaredKindsDeferred) != 0 || ruleDeclaredKindsDeferredMax != 0 {
+		t.Errorf("ruleDeclaredKindsDeferred has %d entry/entries and the ratchet reads %d, but the "+
+			"population above is empty. The ledger is retired: a row here now names a kind the "+
+			"live scan does not produce, which TestNoUndeclaredRuleEntityKinds reports as stale.",
+			len(ruleDeclaredKindsDeferred), ruleDeclaredKindsDeferredMax)
+	}
+	if len(ruleDeclaredFamily) != 0 {
+		t.Errorf("ruleDeclaredFamily still explains %d family/families for an empty ledger. "+
+			"TestYAMLHalfObservesDeclaredKinds requires live YAML traffic for every family it "+
+			"holds, so a family with no rows is either a broken guard or an invitation to add one.",
+			len(ruleDeclaredFamily))
+	}
+}
+
+// TestRuleDeclaredLedger6776_EveryFamilyIsUsedByARow is the reverse coupling
+// the perFamily traffic check in TestYAMLHalfObservesDeclaredKinds used to
+// provide and cannot any more.
+//
+// With the ledger empty, that check's loop body is unreachable, so a future
+// author could reintroduce ruleDeclaredFamily["something"] with no ledger row
+// and nothing would object. This says the two maps move together in the
+// direction the ledger's own family check does not cover: an entry needs a
+// family (checked there), and a family needs an entry (checked here).
+func TestRuleDeclaredLedger6776_EveryFamilyIsUsedByARow(t *testing.T) {
+	used := map[string]bool{}
+	for _, fam := range ruleDeclaredKindsDeferred {
+		used[fam] = true
+	}
+	for fam := range ruleDeclaredFamily {
+		if !used[fam] {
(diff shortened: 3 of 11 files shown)


cajasmota:
## Independent adversarial review — REQUEST-CHANGES (one cheap fix, ~5 lines)

Reviewed at `d2a8a438f` in an own detached worktree, with a second detached worktree at the parent
`37f39ca57` for before/after differentials. Every number below was re-derived from disk; no figure in
the report was taken on trust.

**The work is good.** The disposition argument is sound, the evidence re-derives exactly, and 10 of my
11 mutants died — several with better diagnostics than the guards they replaced. The one survivor is a
real, reproduced coverage loss caused by the retirement, and it lands squarely on the claim the arm's
whole justification rests on. It is cheap to close.

---

### The finding — R11: the three producing sites are no longer anchored to the member

Mutant: re-spell all three `entity_type: Endpoint` in
`internal/engine/rules/javascript_typescript/frameworks/electron.yaml` to `SCOPE.Endpoint`
(match count asserted = 3). This is the silent rename #6820 explicitly rejected, and it makes
`EntityKindEndpointBare` a vocabulary member **no producer writes**.

| tree | result |
|---|---|
| **this PR** (`d2a8a438f`) | **GREEN** — `internal/entkinds`, `internal/types`, `internal/graph/fbwriter`, `internal/engine`, `internal/dashboard`, `cmd/grafel`. All six. |
| **parent** (`37f39ca57`) | **RED** — three tests in `internal/entkinds`: `TestNoUndeclaredRuleEntityKinds` (*"ledger entries the live scan no longer produces: [Endpoint]"*), `TestYAMLHalfObservesDeclaredKinds` (:534), `TestLedgerIsDescribedByItsOwnScan` (*"ledger entry \"Endpoint\" has no site in the live scan"*, :788) |

So the retirement did not only trade one premise for a stronger one. It also **dropped** the coupling
between this kind and the sites that justify it. That coupling was incidental before — it existed
because the ledger row happened to name the kind — but it was the only thing observing the sentence
this PR repeats in five places (PR body, `kinds.go`'s const doc, `endpoint_bare_kind_6776_test.go`'s
header, `all_entity_kinds_roster` comment, `prefixed_producer_enum_membership`): *"three sites, all in
electron.yaml:41,46,52"*.

The PR body says the new test is *"strictly stronger than any ledger row"*. It is strictly stronger in
one direction (any kind outside the enum, anywhere in the rule tree, fails — I confirmed this, see
R9). It is **weaker** in the other: it cannot notice that a member's producers vanished. "Strictly" is
the word that does not survive R11.

Why the existing orphan sweep does not catch it either: `TestEveryEnumEntityKind6818_CorrespondsToSomethingOutsideTheEnum`
accepts a Go *reference* as corroboration, and `internal/dashboard/handlers_export_dsl.go:249` holds a
bare `"Endpoint"` map key. So the member stays corroborated by a colour table after its last producer
is gone. That guard is asking a different question by design; it is not the one to change.

**Requested fix**, in `rule_ledger_retired_6776_test.go`, which already holds the scan result:

```go
// The sites that justify EntityKindEndpointBare's membership. Without this,
// re-spelling electron.yaml's three entity_type rows leaves the whole suite
// green and the member behind with no producer (measured: it does).
var bare []string
for _, s := range res.Sites {
    if s.Kind == string(types.EntityKindEndpointBare) {
        bare = append(bare, fmt.Sprintf("%s:%d", s.File, s.Line))
    }
}
if len(bare) != 3 {
    t.Errorf("the rule-YAML scan finds %d site(s) for %q, want 3 (electron.yaml:41,46,52): %v. "+
        "#6776 arm B9 admitted this kind on the strength of those three producers; if they are "+
        "gone the member is enum-only and needs a KindVocabularyVersion decision, not silence.",
        len(bare), types.EntityKindEndpointBare, bare)
}
```

That single block also closes a second gap I measured, so it is one fix for two findings — see R10
below.

---

### Secondary findings

**R10 — the new test's detection step is ungraded inside its own package.** Mutating
`types.IsValidEntityKind` to return `true` for any non-empty string leaves `internal/entkinds`
**GREEN** (`internal/types` and `internal/graph/fbwriter` go red, so the suite as a whole catches it —
this is not a hole, it is a missing in-fixture positive control). Of the five ways a
scan-and-assert-absence guard can be a no-op, the floor covers "fail to read", R9 covers "wrong
files / wrong content / detect / act", and this is the one left: the predicate itself going
permissive. The `bare` block above fixes it — a site-level positive fact cannot be satisfied by an
over-permissive validator. Cheaper alternative if you prefer: one line,
`if types.IsValidEntityKind("SCOPE.ZZNotAKind") { t.Fatal("fixture is inert") }`. Note the author
already applied exactly this pattern to `IsHTTPEndpointKind` in
`TestEndpointBare6776_MembershipIsNotHTTPMembership`; it is the same fixture discipline, just not
applied here.

**Third guard went vacuous, undisclosed.** The PR body names three guards affected by the empty
ledger (`:447`'s fatal, the `perFamily` loop, the fbwriter ledger transcription). There is a fourth:
`TestLedgerIsDescribedByItsOwnScan` (`rule_declared_kinds_sweep_guard_6744_test.go:818`) loops over
`ruleDeclaredKindsDeferred` with **no floor and no note**, so it is now a fully vacuous green. The
`perFamily` loop got an explicit "UNREACHABLE TODAY, and deliberately kept" paragraph; this one got
nothing. It deserves the same one-paragraph treatment (it is also the test that fired at :788 on the
parent under R11, so it is the one whose loss is being reported above).

**`EveryFamilyIsUsedByARow` is vacuous today but genuinely gradeable** — M8 kills it with the right
message. Fine as a drift guard; just do not read it as evidence of anything present.

---

### Mutant scorecard — 11 mutants, 10 DEAD, 1 SURVIVOR

All at the call site, `-count=1`, `go vet` before each score, match count asserted **after** every
edit, restored by `cp` from shasum-verified backups (`git status --porcelain` empty after each).

|---|---|---|---|
| M1 | drop `EntityKindEndpointBare` from the `AllEntityKinds()` roster | **DEAD** | 4 in `internal/types` + 2 in `internal/entkinds`. **`internal/graph/fbwriter` stays green** — the author's decoupling claim, confirmed |
| M2 | `EntityKindEndpointBare = "SCOPE.Endpoint"` (migrate as a synonym) | **DEAD** | vocabulary pin, roster duplicate-value guard, both `EndpointBare` tests, both entkinds population tests |
| M3 | add `string(EntityKindEndpointBare)` to `IsHTTPEndpointKind` | **DEAD** | `TestIsHTTPEndpointKind_ForbiddenRows` + `MembershipIsNotHTTPMembership`, **and all four #6820 dashboard tests** — I ran the dashboard half too, not just `internal/types` |
| M4 | reinstate ledger row + ratchet + family | **DEAD** | and it settles the `t.Fatal`→`t.Logf` question: `:467` fired — *"ruleDeclaredKindsDeferred holds \"Endpoint\", which types.AllEntityKinds() now also carries"*. **The row loop still grades.** The demoted line logged, as designed |
| M5 | *positive control*: scan `t.TempDir()` | **FLOOR FIRES** | *"produced only 0 site(s) from 0 file(s) (floor: 300)"* — not a vacuous pass on "zero invalid" |
| M6 | all five re-picked controls → the valid `Model` | **DEAD** | all five `fixture is inert` guards, each naming the stale control. **Five, not four** — confirmed |
| M6b | `:463`'s control only → `Model` | **DEAD** | *"fixture is inert: Summary()=… EntitySummary()=\"\""*. B8's estimate genuinely missed this one |
| M7 | `Endpoint` onto `goUnprefixedKindsDeferred`, pin 5→6 | **DEAD** | *"ledgered kind \"Endpoint\" is no longer declared by any Go producer"* — the Go-ratchet claim is graded, not assumed |
| M8 | family in `ruleDeclaredFamily` with no row | **DEAD** | `TestYAMLHalfObservesDeclaredKinds` :571, the retired-ledger test :94, and `EveryFamilyIsUsedByARow` :117 — the kept-loop claim holds |
| **R9** *(mine)* | inject `entity_type: ZZReviewerBogusKind` at `electron.yaml:41` | **DEAD** | *"ZZReviewerBogusKind (1 site(s)): …electron.yaml:41"*. This is the strong one: it closes no-op modes 2–5 at once (right files, right content, detect, act) with a line-exact diagnosis. The new test is a real guard |
| **R10** *(mine)* | `IsValidEntityKind` returns true for any non-empty string | **partial** | `internal/entkinds` GREEN; `internal/types` + `fbwriter` red. In-package detection is ungraded — see above |
| **R11** *(mine)* | re-spell electron.yaml's 3 `entity_type` rows to `SCOPE.Endpoint` | **SURVIVOR** | green across all six packages; **red in three tests on the parent**. Three-way verdict: *distinguishable and ungraded* — so I am grading it above rather than excusing it |

---

### Claim 1 — DESIGN VERDICT: the disposition is correct. Concur.

**The load-bearing half checks out.** `internal/engine/detector.go:466` is `Kind: sp.entityType`,
fed from `sp.EntityType` at :168 — verbatim, no validation on that path. Stronger than the report
claims: **`IsValidEntityKind` has zero non-test production callers anywhere in `internal/` or
`cmd/`.** Its only non-test appearance is in `internal/graph/fbwriter/non_enum_entity_kinds_6776.go`,
which *counts* rather than validates. So membership decides exactly two things — whether grafel's own
metric flags an entity grafel itself emitted, and what the test suite accepts. The argument does not
invert; it is on firmer ground than stated.

**On "distinct concept → distinct member", which is the real question.** The worry is legitimate:
the enum now carries `Endpoint` (Electron IPC) and `SCOPE.Endpoint` (HTTP) as *different concepts*
told apart only by the prefix — which makes the `SCOPE.` prefix semantically load-bearing, in a
series whose whole framing (arms B4–B8, and the ledger's own retired family note) was that the
un-prefixed spelling is *"an accident, not a namespace"*. Every earlier arm could say that honestly
because both spellings meant the same thing. This one cannot. That is a genuine cost and the PR
should not be read as denying it.

But it is the right call anyway, for a reason that is not the author's framing:

1. The honest fix for a distinct concept is a distinct **name** (`IPCChannel`), not a distinct member
   with a colliding one — and the owner **ruled that out** on 2026-09-06 in #6820: *"No rename, no
   stored-graph migration… enum membership is the target, not re-spelling kinds."* I read the ruling
   rather than taking the PR's summary of it; the summary is accurate.
2. With the rename off the table, the only two options are *member* or *permanently-invalid produced
   kind*. The second is what #6776 exists to end, and it costs a false `entity_entities_kind_not_in_enum`
   signal on every Electron graph forever. Membership is the residual.
3. The hazard is closed where it actually lives — at the consumers (#6893, `9706a5cde`) — and this PR
   pins it twice (`http_endpoint_kind_forbidden_6894_test.go` and the new
   `MembershipIsNotHTTPMembership`), with a positive control so the rejection is not satisfied by a
   function that rejects everything. M3 kills both halves plus four dashboard tests. That is the part
   that had to be right, and it is.

So: **not a second namespace by the back door — a naming debt the owner priced and declined to pay.**
It is documented loudly rather than hidden, which is the correct handling. Worth a follow-up issue
recording that `Endpoint`/`SCOPE.Endpoint` is now the one pair in the vocabulary where the prefix
carries meaning, so the next reader of the "prefix is an accident" comments is not misled by them.

---

### Everything re-derived from disk

Via `entkinds.ScanRuleYAML` / `ScanGo` in a throwaway probe (deleted), never grep:

```
yaml_files=732  sites=532  unresolved=0  distinct=27  invalid=0
Endpoint sites: electron.yaml:41, :46, :52   (exactly three, line-exact — confirmed)
go_sites=567    go_endpoint=0
```

- **732 YAML files** — the author's correction of the brief's 714 is right; the brief was wrong.
- Counters: `93→94` / `92→93` both required by `TestEntityKindDeclarations6776_MatchAllEntityKindsExactly`.
  `KindVocabularyVersion` staying at 1 is correct — `kind_vocabulary_bump_guard_6779_test.go:275`
  gates the bump on `len(removed) > 0`, and :291 requires the pin list to match exactly, so adding
  `"Endpoint"` to `pinnedEntityKindVocabulary` was mandatory and is graded (M2 kills it).
  `goUnprefixedKindsDeferredMax` staying at 5 is graded by M7. `enumOnlyEntityKindsMax` staying at 1
  is right: three YAML sites corroborate the kind, so no allow-list row.
- The four re-picked controls (`ChannelEvent`, `File`, `Stream`, `Subscription`) are **exactly** the
  four producer rows of `goUnprefixedKindsDeferred` (five rows minus `relationship`), and they are
  held outside the enum by `assertLedgerExact`, which derives its population from `ScanGo` and is
  coupled to `IsValidEntityKind` by construction. No #6893 trap: no control is matched by another
  clause of its own predicate — `:364` has an explicit "both slots differ" guard that fires,
  `:101`'s arm assertion (`EntityDistinctKinds != 2`) would catch a collapsed pair, and `:590`'s
  control is checked for the absence of the `SCOPE.` prefix separately from its validity.
- **Claim 5 verified**: `http_endpoint_kind_forbidden_6894_test.go`'s assertions are substantively
  unchanged — `"Endpoint"` → `string(EntityKindEndpointBare)` is the same string, everything else in
  the diff is comment. M3 confirms it still fires.

### Ran

`go vet` (clean) and `go test -count=1` on `./internal/entkinds/`, `./internal/types/`,
`./internal/graph/fbwriter/`, `./internal/engine/`, `./internal/dashboard/`, `./cmd/grafel/` — all
**ok**, exit 0. `gofmt -l internal/ cmd/` clean. `scripts/quality/run.sh --ratchet` reproduced:
*"quality ratchet OK — 31 gated fixtures held their baseline (22 at 100% must-have recall)"*, exit 0.

### Did NOT run

Full-repo `go test ./...`. No corpus index, so I did not verify the
`entity_entities_kind_not_in_enum` inference — and I agree with the author's decision to state it as
an inference rather than quote a number. The digest-unmoved claim I verified only as "`cmd/grafel`
is green", not by reading the digest itself.

---

**Verdict: REQUEST-CHANGES** on R11 alone (~5 lines, in a file that already has the scan result in
hand, and which closes R10 in the same block). Everything else in this PR is in good shape, and two
of the author's mutants (M6b, M7) are the kind that reviews usually have to add rather than find
already present.

cajasmota:
## Coordinator mutant — DEAD. My hypothesis was wrong, and the reason is worth recording.

I attacked the one no-op mode the existing mutants could not reach. R9 injected a bogus kind at `electron.yaml:41` — a file the scan **does** read — which closes *right-content / detect / act*. M5 (scan pointed at `t.TempDir()`) closes *reads nothing*. Neither can see the mode in between: **the scan reads, but reads too few files.**

That looked live, because the new test's floor is `minRuleYAMLSites = 300` against **532** actual sites — **43% slack**. My hypothesis: a scan that silently loses a large subtree would stay above the floor, and "not one of 532 sites names an invalid kind" would quietly become "not one of the 350 I still look at".

**Mutant:** make `ScanRuleYAML`'s `WalkDir` skip a subtree.

### First attempt — a kill I threw out

I skipped paths containing `"python"` or `"java"`. Seven tests failed, including the new site pin. **But `"java"` also matches `javascript_typescript/`, which is where `electron.yaml` lives** — so the site pin fired because I had deleted the pinned sites themselves, not because anything detected the coverage loss. That kill rested on an accident of my chosen string and proves nothing about the hypothesis.

Recording it rather than quietly re-running: a kill whose premise is accidental is the failure this repo keeps finding, and I nearly reported it.

### Second attempt — clean premise

Skip only `"/python"`, a subtree containing **none** of the three pinned `Endpoint` sites. Occurrence gate 1, `go vet` clean.

**DEAD** — 4 failures:

```
--- FAIL: TestZeroProducingKinds6776_SiteCountsHold
    zero_producing_kinds_6776_test.go:73: Fixture: scan finds 0 rule-YAML declaration sites, arm A measured 1.
--- FAIL: TestRepoSweepIsNotVacuous
--- FAIL: TestEveryEnumEntityKind6818_CorrespondsToSomethingOutsideTheEnum
--- FAIL: TestScanRuleYAMLReadsBoundSourcePatterns
```

### The finding: coverage is graded, but not by anything in this PR

**My hypothesis was wrong** — scan coverage is genuinely pinned. But the thing pinning it is **not** the 300 floor, which the mutant sails past. It is the **pre-existing per-kind site pins** (`TestZeroProducingKinds6776_SiteCountsHold`, which asserts exact `file:line` sets per kind and so notices any site vanishing) plus `TestRepoSweepIsNotVacuous`.

Two consequences worth stating in the body:

1. **The floor is not load-bearing for coverage** and should not be described as if it were. It catches a scan that *collapses*; the per-kind pins catch a scan that *shrinks*. The file's comment already says the floor catches a collapse — that is accurate, and I am confirming it rather than correcting it.
2. **The protection is inherited, not added.** This arm empties its own population, so its safety against a shrinking scan comes entirely from tests that other arms left behind. That is fine — but it means the property is only as durable as those neighbours. If a later change retires `zeroKindSites()` the way this one retired the ledger, the coverage property leaves with it and nothing in `rule_ledger_retired_6776_test.go` will notice.

Not a blocker, and no code change requested. Restored by `cp` from a shasum-verified backup (`cbd49c15…`); tree confirmed pristine.

Refs #6776, #6908.

PR #6913, opened 2026-09-06
fix(#6852): dart's import IMPORTS anchored on a path only `dartTopName` can spell — arm 11, and a mutant narrowed the placement window by one loop

cajasmota:
Arm 11 of 12. `buildImportRecord` (`dart.go:429`; **349 was the pre-patch line** — the same commit that added the reference moved the site, review nit 1) stamps `FromID: filePath` on every `import '...';` IMPORTS edge; a conditional `extractor.PrependFileCarrier` now mints the carrier, and dart's ledger line is deleted. **One remains: zig.**

## The site table, grounded

`FromID` sites in `internal/extractors/dart/` (non-test): **exactly one** — `dart.go:349`. Every other edge dart emits sets no `FromID` at all: `extractCallRelationships` builds CALLS with only ToID/Kind/Properties, and the method pass's CONTAINS sets only ToID/Kind.

Six `Name` sites, and the closure is split because **only one of them is reachable**:

| # | site | Name | verdict |
|---|---|---|---|
| 1 | `buildImportRecord` | `dartTopName(uri)` | **REACHABLE at root depth — driven, not asserted** |
| 2 | class pass | `classRE` `(\w+)` | non-offender, observed |
| 3 | method pass | `methodRE` `(\w+)` | non-offender, observed |
| 4 | `dartEnums` | `dartEnumRE` `(\w+)` via `EnumEntity` | non-offender, observed (Name **and** QualifiedName) |
| 5 | `dartTypeAlias` | both `typedef` spellings, `(\w+)` | non-offender, observed |
| 6 | `dartModifiedClasses` | `dartModifiedClassRE` `(\w+)` | non-offender, observed |

**Site 1 is reachable and is DRIVEN.** `dartTopName` strips a scheme at the first `:`, truncates at the first `/`, then `TrimSuffix`es ONE `".dart"`. So `import 'main.dart.dart';` in a ROOT `main.dart` yields an import record named exactly `main.dart` — clause 3 fires and no second carrier is minted (two nodes under one `graph.EntityID`, the #6369/#6480 hazard). `TestDart_ImportTopNameSpelledLikeThePathGetsNoSecondCarrier_6852/root` drives it. At NESTED depth it is unreachable, and that is a *slicing* property rather than a character class: the return is always a substring of `uri` ending before the first `/` (`s = s[:i]`), so it cannot hold one — driven by the `nested_contrast` subtest, which is also what stops the root cell from passing on a carrier that is never emitted at all.

**The closure the non-offenders rest on, and why concatenation cannot defeat it.** Sites 2-6 are each a VERBATIM slice of a `(\w+)` capture, `src[m[a]:m[b]]`. There is **no builder** — that is the whole argument. cobol's closure fell because a builder concatenated a prefix onto a literal-derived operand; here the only concatenations in the package go into `Signature` (`"class "+name`, `name+"("+params+")"`, `"typedef "+name`, `modifier+" class "+name`) and into `Properties`, neither of which clause 3 or `refs.go` reads. `\w` is `[0-9A-Za-z_]`, so such a Name holds no `'.'`; `classifier.go:372` maps `.dart` and that extension entry is the only route into this extractor, so every production path holds one. Site 4 additionally sets `QualifiedName = scope:enum:<path>:<Name>` — the crystal **length** invariant, `11 + len(path) + 1 + len(Name) > len(path)`.

And it is **observed, not argued**: `TestDart_WordCharacterSitesCannotSpellThePath_6852` drives a source that names every one of sites 2-6 the path's stem *and* attempts the literal path spelling on each (`class api.dart {`, `enum api.dart {`, `typedef api.dart = …`, `sealed class api.dart`), then asserts over EVERY record's Name and QualifiedName, at both depths. A site that started concatenating shows up there.

## Placement — and the mutant that disproved this PR's own first draft

**One independent, production-reachable reason; one entailed-in-effect. Named as such.**

**INDEPENDENT (graded):** the index hazard. `classSpan.idx` is an INDEX into `entities`, dereferenced in the method pass to hang each method's CONTAINS on its enclosing class. **But the window is one loop narrower than the first draft of this comment claimed, and scoring is what found it.** The draft said a carrier between the IMPORT pass and the CLASS pass shifts the stored indices. That mutant (M5) came back **ALIVE** — `idx := len(entities)` is evaluated *inside* the class loop, so a head insertion before that loop is absorbed. The window that actually bites is strictly **between the class loop and the method loop**, and that move (M5b) is **DEAD** with wrong-owner CONTAINS rows. Both source comments were corrected to the narrower claim rather than left as prose the tests do not support. It is not entailed by clause 2 — dart's import pass runs FIRST, so a conditional carrier in that window really is emitted, unlike clojure (#6897).

**ENTAILED IN EFFECT (named, not dressed up):** clause-3 visibility of the type pass. In general the call must run after the `extractDartTypes` append; for dart it buys nothing, because every name that pass emits is a `(\w+)` slice and the one reachable clause-3 route is emitted in the FIRST pass. M5c confirms it: moving the call before the type pass is **ALIVE**.

The `"dart"` token is graded through the **provenance** route, not the `Language` field alone: dart runs `TagEntitiesLanguage` in `Extract`, AFTER `extractDart` returns, so an empty token would be filled; `TagEntitiesLanguage` stamps `Properties["language"]` only on that fill path and `FileEntity` never sets that key, so its absence is what makes the Language row evidence. `TestDart_CarrierShape_6852` asserts both halves.

## Mutant table

Scored at the CALL SITE, `-count=1`, `go vet` clean before every score, each edit gated on an exact occurrence count asserted **after** the edit landed, restored by `cp` from a `shasum`-verified backup (never `git checkout --`).

|---|---|---|---|
| M1 | call deleted (`return entities`) | **DEAD** | ImportsFromEndResolves (both depths), OneCarrierPerFile, nested_contrast, CarrierShape, Placement |
| M2 | unconditional carrier (`FileEntity` prepended always) | **DEAD** | NoCarrierWithoutAnImport (both depths), EmptyPathGetsNoCarrier, ImportTopName/root |
| M3 | empty token `""` | **DEAD** | CarrierShape — the `Properties["language"]` provenance row |
| M4 | wrong token `"flutter"` | **DEAD** | CarrierShape — Language row |
| M5 | placement moved before the class pass | **ALIVE — equivalent on the index axis, not-production-reachable on the clause-3 axis** | `idx := len(entities)` is evaluated inside the class loop, so the **index-axis** equivalence is unconditional and exact, and it **falsified the placement claim this PR started with** (both comments were rewritten to the narrower window). But the move also puts the call where clause 3 cannot see the class/method/type records, so *full* equivalence additionally rests on Claim 1 — **the same dependency M5c rests on**, which an earlier draft of this table attributed to M5c alone. The reviewer's R1 mutant demonstrates it: add a builder to the class Name site and M5 and HEAD become distinguishable. Labelled in full so the verdict is not read as broader than it is. |
| M5b | placement moved between the class loop and the method loop | **DEAD** | Placement test, with the exact wrong-owner rows (`CONTAINS -> …:handle is owned by [flutter], want ["Handler"]`) |
| M5c | placement moved before the type pass | **ALIVE — not production-reachable** | Input that would distinguish it: a type-pass record (enum / typedef / modified class) named the path. Impossible — those Names are `(\w+)` slices and every production dart path holds a `'.'`. This is exactly the "entailed in effect" conjunct above, and M5c is its evidence. |
| M6 | carrier named `BASENAME(path)` | **DEAD** | ImportsFromEndResolves (nested), OneCarrierPerFile, nested_contrast, CarrierShape Name row, Placement |
| **M7a** | **travelling:** clause 3 `&& len(rels) > 0` | **DEAD** | Not by dart's cells (dart's only path-named record, the import stub, always owns exactly one IMPORTS edge) — killed in `internal/extractor` by `TestFileCarrierFor_NoSecondCarrierWhenOneIsAlreadyNamedAfterThePath_6815` and `…WhenThePathNamedRecordComesLast_6815` |
| **M7b** | **travelling:** clause 3 `&& len(rels) == 0` | **DEAD** | `TestDart_ImportTopNameSpelledLikeThePathGetsNoSecondCarrier_6852/root` — dart's own cell |

**Travelling-mutant tally now: DEAD on astro · ALIVE on clojure · ALIVE on cobol until a new cell was added · fully graded on crystal by disjoint cells · DEAD on dart, but only half of it by dart's own cell.** Stated precisely because the halves are not symmetric here: dart's reachable clause-3 record is an import stub that always owns exactly one relationship, so the `> 0` variant is indistinguishable from the original *within this package* and its kill comes from the generic #6815 fixtures. That is a real gap in dart's local grading of clause 3's relationship-independence, recorded rather than papered over.

## Fixture axes

Every multi-row test states them, and each holds its neighbour constant:

- `ImportsFromEndResolves`, `NoCarrierWithoutAnImport`, `WordCharacterSites`, `CarrierShape`, `Placement` — **varied:** path depth (`lib/src/api.dart` vs `api.dart`); **held constant:** the source, byte for byte.
- `ImportTopNameSpelledLikeThePath` — **varied:** path depth; **held constant:** the source, byte for byte (the same `main.dart.dart` import in both rows), so the only difference between "clause 3 declines" and "the carrier is minted" is depth.
- `NoCarrierWithoutAnImport` — **varied:** the `import` directives (absent); **held constant:** a full record set (class + superclass + method, top-level function, enum, typedef, sealed class, and the CONTAINS/CALLS edges they own), asserted as a premise so "no carrier" is a decision and not an empty extraction.
- `OneCarrierPerFileNotPerImport` — **varied:** the number of anchored edges (3); **held constant:** depth and the single class.
- `WordCharacterSites` — **varied:** within the source, the spelling of each declared name (stem vs literal path); **held constant:** the set of declaration kinds present, so every site is exercised in both rows.

## Shared counters, re-derived from disk

- `mustScan` in `internal/extractor/carrier_caller_set_6861_test.go` — row added for `internal/extractors/dart/dart.go`. **It is NOT the load-bearing counter, and the brief that framed it as one was wrong.** The anchor list is documented as *redundancy* with a floor of 6 against 16 entries; the reviewer's R2 (delete dart's row) is **ALIVE by design** (#6887) and leaves both packages green. Adding the row is right for consistency and costs nothing, but a future arm that forgets it will not go red there.
- `knownMissingCarrier6847` — dart's line deleted; `len()` is now 1 (`zig`). **This is the load-bearing counter**: the reviewer's R3 (re-add `dart`) goes **RED** with the exact-set message, which confirms the fix at *corpus runtime* rather than only under unit fixtures.
- `exercisedLanguages6847` and `anchoringLanguages6847` keep `dart`: it is still exercised and still anchors.

## Verification

- `go test -count=1` GREEN on `./internal/extractors/dart/`, `./internal/extractors/`, `./internal/extractor/`, `./internal/resolve/`, **`./cmd/grafel/`** (145s; the #6118 whole-graph digest did **not** move — no dart member in that mixed-language fixture, so no constant was touched and none needed enumerating), and `./internal/quality/...`.
- `scripts/quality/run.sh --ratchet` — **exit 0**, "31 gated fixtures held their baseline (22 at 100% must-have recall)". `baseline.json` is untouched: there is no dart fixture under `internal/quality/golden/`, so neither `entity_extracted_total` nor `relationship_extracted_total` moved.
- `gofmt -l internal/ cmd/` clean. Tree verified free of mutants (`git diff` on `internal/extractor/file_carrier.go` is empty; dart.go's only non-comment change is the one-line call).

**Not run:** the full repo suite, `-race`, the Windows/macOS legs, and `grafel index` over a real corpus. Left to the coordinator.

---

## Review round 2 — comment/body only (`1d5fc9479`)

**No code changed.** `git diff HEAD~1 HEAD` filtered to lines that are neither `//` nor blank returns **0 lines** — verifiable with a diff. Three prose corrections, all from the review:

1. **`dart.go`'s clause-3 paragraph misattributed the `QualifiedName` invariant.** `FileCarrierFor` clause 3 compares `records[i].Name` and nothing else; the enum site's `scope:enum:<path>:<Name>` length invariant belongs to `refs.go`'s `byQualifiedName` exact-match tier (`refs.go:2103-2109`, ahead of `lookupStructural` — a grounding pass has driven that tier and confirmed it binds). Both mechanisms are now named as real, separate and adjacent, with the reason the distinction is written down: **zig is the last arm and will read this file**, and inherited prose is how a false closure propagates.
2. **Stale line cite fixed** in the test header *and* above in this body: `dart.go:349` → **`429`**. The header also records that 349 was the parent's number and says to read the *function name* as the durable half; `dart.go`'s own header now names `dartTopName` with **no line at all**, since this very round shifted the file again.
3. **The WordCharacter header over-claimed.** It said a literal-path attempt is made on "each" of sites 2-6; the method site gets only the stem. **Narrowed the sentence** (the alternative, extending the case, would have been a code change) to *four of the five* declaration kinds, and added why the omission is not a gap: `methodRE`'s `(\w+)` must be followed by `\s*\(`, so a dotted method name is not expressible. The axis sentence was narrowed to match. The R1 mutant is now named in that header as what it grades.

Re-verified after the edits: `go vet` clean, `go test -count=1 ./internal/extractors/dart/` **GREEN**, `gofmt -l internal/ cmd/` clean.

Refs #6852, #6847.

### DIFF
--- a/internal/extractor/carrier_caller_set_6861_test.go
+++ b/internal/extractor/carrier_caller_set_6861_test.go
@@ -207,6 +207,7 @@ func TestCarrierCallerSetIsGradedFromSource_6861(t *testing.T) {
 		{"internal/extractors/clojure/clojure.go", []string{prependFileCarrier}},
 		{"internal/extractors/cobol/extractor.go", []string{prependFileCarrier}},
 		{"internal/extractors/crystal/extractor.go", []string{prependFileCarrier}},
+		{"internal/extractors/dart/dart.go", []string{prependFileCarrier}},
 	}
 
 	nonTest := 0
--- a/internal/extractors/dart/dart.go
+++ b/internal/extractors/dart/dart.go
@@ -239,7 +239,75 @@ func extractDart(src, filePath string) []types.EntityRecord {
 	// class/method records above.
 	entities = append(entities, extractDartTypes(src, filePath)...)
 
-	return entities
+	// ── #6852 — file carrier for the path-anchored IMPORTS above ─────────
+	//
+	// buildImportRecord stamps `FromID: filePath` on every import's IMPORTS
+	// edge, and internal/resolve/refs.go has no path→entity index: a
+	// path-valued FromID resolves iff some emitted node carries that exact
+	// string as its Name (or QualifiedName). Nothing here does as a rule, so
+	// without this the raw path reached the graph as the edge's FROM end —
+	// #6847's measurement, #6815's fix shape.
+	//
+	// CONDITIONAL, via extractor.PrependFileCarrier. An unconditional carrier
+	// mints one bare orphan node per .dart file across a whole repo, which no
+	// recall-shaped assertion can see (#6518, #6815).
+	//
+	// THE ONE Name SITE THAT CAN SPELL THE PATH is dartTopName, and it is a
+	// DRIVEN cell rather than a closure: for a ROOT path main.dart the URI
+	// `main.dart.dart` yields exactly `main.dart`, so clause 3 declines in
+	// favour of the import record that already carries the anchor. The other
+	// five Name sites (class / method / enum / typedef / modified class) are
+	// verbatim `(\w+)` captures with no builder concatenating onto them, and
+	// `\w` is [0-9A-Za-z_], so none can hold the '.' every production dart
+	// path carries; the enum site's QualifiedName is
+	// `scope:enum:<path>:<Name>`, strictly longer than the path. Both halves
+	// are OBSERVED by file_carrier_6852_test.go, at both depths, rather than
+	// asserted — four arms of #6852 shipped a closure a driven cell later
+	// disproved.
+	//
+	// PLACEMENT: last, and the compound is scored PART BY PART with the
+	// entailed conjunct NAMED (the cobol lesson, #6899).
+	//
+	//  1. INDEPENDENT AND PRODUCTION-REACHABLE — THE INDEX HAZARD.
+	//     classSpan.idx is an INDEX INTO `entities`, dereferenced in the
+	//     method pass as `entities[cls.idx].Relationships = append(...)` to
+	//     hang each method's CONTAINS edge on its enclosing class.
+	//     THE HAZARD WINDOW IS NARROWER THAN "ANYWHERE EARLIER", and saying so
+	//     precisely is the point: an earlier draft of this comment claimed a
+	//     carrier placed between the IMPORT pass and the CLASS pass shifts the
+	//     stored indices, and MUTANT SCORING DISPROVED IT — `idx :=
+	//     len(entities)` is evaluated inside the class loop, so a head
+	//     insertion BEFORE that loop is absorbed and the move survives every
+	//     test (recorded as M5 in the PR, ALIVE and equivalent). The window
+	//     that really bites is strictly BETWEEN the class loop and the method
+	//     loop: the spans are recorded and not yet dereferenced, so a head
+	//     insertion there shifts every stored idx by one and each class's
+	//     CONTAINS edges land on the record before it. Nothing crashes; the
+	//     graph is quietly wrong.
+	//     It is not entailed by clause 2 — dart's import pass runs FIRST, so
+	//     the path-anchored edge is already in `entities` and a CONDITIONAL
+	//     carrier placed in that window really is emitted, unlike clojure
+	//     (#6897) where the same move minted nothing and the hazard graded
+	//     itself away. Graded by
+	//     TestDart_CarrierPlacementDoesNotShiftTheContainsPass_6852, which
+	//     kills that move with wrong-owner CONTAINS rows (M5b).
+	//  2. ENTAILED IN EFFECT — CLAUSE-3 VISIBILITY OF THE TYPE PASS.
+	//     FileCarrierFor can only decline for a record it is handed, so in
+	//     general the call must run after the extractDartTypes append above.
+	//     For dart that reason buys nothing and is named here rather than
+	//     dressed up as a second independent one: every name extractDartTypes
+	//     emits is a `(\w+)` slice and cannot spell a '.'-bearing path, and
+	//     the one reachable clause-3 route is emitted in the FIRST pass, so
+	//     clause 3 sees it wherever after the import loop the call sits. Its
+	//     unreachability is what
+	//     TestDart_WordCharacterSitesCannotSpellThePath_6852 observes.
+	//
+	// The "dart" token is stamped explicitly rather than left to Extract's
+	// TagEntitiesLanguage, which runs AFTER this and would fill an empty one —
+	// so the token is graded through the Properties["language"] provenance
+	// route, not through the Language field alone. See
+	// TestDart_CarrierShape_6852.
+	return extractor.PrependFileCarrier(filePath, "dart", entities)
 }
 
 // methodStartIsClass reports whether the byte offset is inside a class's
--- a/internal/extractors/dart/dart.go
+++ b/internal/extractors/dart/dart.go
@@ -252,18 +252,30 @@ func extractDart(src, filePath string) []types.EntityRecord {
 	// mints one bare orphan node per .dart file across a whole repo, which no
 	// recall-shaped assertion can see (#6518, #6815).
 	//
-	// THE ONE Name SITE THAT CAN SPELL THE PATH is dartTopName, and it is a
-	// DRIVEN cell rather than a closure: for a ROOT path main.dart the URI
+	// THE ONE Name SITE THAT CAN SPELL THE PATH is dartTopName (named, not
+	// line-cited: a line number in this file goes stale the next time someone
+	// edits a comment above it, which is nit 1 of this arm's review), and it is
+	// a DRIVEN cell rather than a closure: for a ROOT path main.dart the URI
 	// `main.dart.dart` yields exactly `main.dart`, so clause 3 declines in
 	// favour of the import record that already carries the anchor. The other
 	// five Name sites (class / method / enum / typedef / modified class) are
 	// verbatim `(\w+)` captures with no builder concatenating onto them, and
 	// `\w` is [0-9A-Za-z_], so none can hold the '.' every production dart
-	// path carries; the enum site's QualifiedName is
-	// `scope:enum:<path>:<Name>`, strictly longer than the path. Both halves
-	// are OBSERVED by file_carrier_6852_test.go, at both depths, rather than
-	// asserted — four arms of #6852 shipped a closure a driven cell later
-	// disproved.
+	// path carries. Both halves are OBSERVED by file_carrier_6852_test.go, at
+	// both depths, rather than asserted — four arms of #6852 shipped a closure
+	// a driven cell later disproved.
+	//
+	// TWO MECHANISMS, ADJACENT AND SEPARATE — do not merge them. FileCarrierFor
+	// CLAUSE 3 compares `records[i].Name` and NOTHING ELSE, so everything above
+	// is what clause 3 sees. The enum site's QualifiedName
+	// (`scope:enum:<path>:<Name>`, strictly longer than the path — crystal's
+	// LENGTH invariant, #6905) belongs to a DIFFERENT consumer: refs.go's
+	// `byQualifiedName` exact-match tier (refs.go:2103-2109, ahead of
+	// lookupStructural), which a grounding pass has driven and confirmed binds.
+	// The invariant matters because it is the other way a record could satisfy
+	// the resolution question #6847 measures — not because clause 3 reads it.
+	// Said explicitly here because this is the file the LAST arm of #6852 (zig)
+	// will read, and inherited prose is how a wrong attribution propagates.
 	//
 	// PLACEMENT: last, and the compound is scored PART BY PART with the
 	// entailed conjunct NAMED (the cobol lesson, #6899).
--- a/internal/extractors/dart/file_carrier_6852_test.go
+++ b/internal/extractors/dart/file_carrier_6852_test.go
@@ -17,7 +17,13 @@
 //
 // FromID sites in internal/extractors/dart/ (non-test): exactly ONE.
 //
-//	dart.go:349  buildImportRecord  FromID: filePath   — path-anchored, IMPORTS
+//	dart.go:429  buildImportRecord  FromID: filePath   — path-anchored, IMPORTS
+//
+// (349 in the parent commit. The commit that added this reference MOVED the
+// site, so the number it shipped with pointed at the wrong function — review
+// nit 1. Read the FUNCTION NAME as the durable half; the line drifts with any
+// comment edit above it, which is why dart.go's own header now names
+// dartTopName without a line at all.)
 //
 // Every other relationship dart emits carries no FromID at all: the CALLS edges
 // extractCallRelationships builds set only ToID/Kind/Properties, and the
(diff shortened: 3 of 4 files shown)


cajasmota:
## Coordinator mutant — DEAD, and it grades the arm's *premise* rather than its behaviour

Every mutant in the author's table (M1–M7b) attacks the **carrier call** — delete it, make it unconditional, empty/wrong token, move it, widen clause 3. None attacks the thing the whole arm rests on: **that dart's offender is reachable at all.**

That premise is one `TrimSuffix`. `dartTopName` strips a scheme at the first `:`, truncates at the first `/`, then trims **exactly one** `".dart"` — which is the only reason `import 'main.dart.dart';` in a root `main.dart` can name a record exactly the path.

**Mutant:** make the trim exhaustive.

```go
for strings.HasSuffix(s, ".dart") { s = strings.TrimSuffix(s, ".dart") }
```

`main.dart.dart` now yields `main`, so **clause 3 can never fire for dart** and the offender ceases to exist. Occurrence gate 1; `go vet` clean before scoring.

**DEAD**, on exactly the cell that matters:

```
--- FAIL: TestDart_ImportTopNameSpelledLikeThePathGetsNoSecondCarrier_6852/root
    file_carrier_6852_test.go:548: the one record named "main.dart" is the CARRIER — clause 3
      should have declined in favour of the import record dartTopName already named "main.dart"
    file_carrier_6852_test.go:553: want 0 carriers when a record is already named "main.dart", got 1
```

### Why this is the verdict the arm needed

The cell **detects the premise breaking**, not merely today's behaviour. Had it been written to assert only "a carrier exists for an anchored dart file", this mutant would have sailed through: the carrier still gets minted, the import edge still resolves, and the arm would look fine while the case it was written for had silently become unreachable.

That is the cobol shape, and it is the reason cobol's exoneration was worth trusting: I stripped the prefix from `BuildOperationStructuralRef` — a helper in a *different package* — and its test failed with a diagnosis naming the broken premise. Same property here, one package in.

It also settles something the brief asked the reviewer to attack. The author's site-1 argument is a **slicing** property, which I flagged as weaker than crystal's length invariant and stronger than cobol's character-class closure. It is now graded: if the slicing behaviour changes in either direction, a named cell fails with a diagnosis that says which. The strength of the *argument* matters less once the *cell* is the thing holding it.

### What this does not cover

- **Only the root depth.** The mutant kills at `/root`; I did not construct a nested case, and cobol's "root depth only" turned out to be an artefact of a false closure. The reviewer is checking whether dart's root-only reachability is real or the same artefact — this mutant does not answer that.
- **It says nothing about M7a.** The travelling mutant being DEAD-but-not-by-dart's-cells is a separate question and remains with the reviewer.

Restored by `cp` from a shasum-verified backup (`efa692b0c…`); tree confirmed pristine. Worktree at a unique path — the shared `<scratch>/wt` collision destroyed this author's first tree mid-run, and that was my doing.

Refs #6852.

cajasmota:

Reviewed at `d0c21d6ad` in a private worktree. Nothing below is taken from the PR body: every mutant was re-applied at the call site with an exact occurrence count asserted **after** the edit landed, `go vet` clean before each score, `-count=1`, restored by `cp` from `shasum`-verified backups. **All ten reported verdicts reproduce exactly.**

### Mutant scorecard (re-scored independently)

| # | mutant | my verdict | agrees with PR |
|---|---|---|---|
| M1 | call deleted | **DEAD** (ImportsFromEnd ×2, OneCarrierPerFile, nested_contrast, CarrierShape ×2, Placement ×2) | ✅ |
| M2 | unconditional `FileEntity` | **DEAD** (NoCarrierWithoutAnImport ×2, EmptyPathGetsNoCarrier, ImportTopName/**root**) | ✅ |
| M3 | empty token `""` | **DEAD** (CarrierShape ×2 — the `Properties["language"]` provenance row) | ✅ |
| M4 | wrong token `"flutter"` | **DEAD** (CarrierShape ×2) | ✅ |
| M5 | moved before the class loop | **ALIVE** | ✅ (with a caveat below) |
| M5b | moved between the class loop and the method loop | **DEAD** (Placement ×2) | ✅ |
| M5c | moved before the type pass | **ALIVE** | ✅ |
| M6 | carrier named BASENAME(path) | **DEAD** (nested rows of ImportsFromEnd / CarrierShape / Placement, OneCarrierPerFile, nested_contrast) | ✅ |
| M7a | clause 3 `&& len(rels) > 0` | **DEAD — and `./internal/extractors/dart/` is GREEN under it** | ✅ |
| M7b | clause 3 `&& len(rels) == 0` | **DEAD** by `…ImportTopName…/root` — dart's own cell | ✅ |

**Four mutants I added.**

| # | mutant | verdict |
|---|---|---|
| **M5z** | call placed **before the import loop** — the fifth placement position the PR did not score | **DEAD** (clause 2 rejects, no carrier ever: the clojure failure mode). The placement window is now exhaustively enumerated: 5 positions, 3 DEAD, 2 ALIVE. |
| **R1** | a builder added to the **class** Name site (`Name: name + ".dart"`) — the exact shape that falsified cobol | **DEAD** by `TestDart_WordCharacterSitesCannotSpellThePath_6852/api.dart`. The closure is genuinely *observed*, not argued. |
| **R2** | dart's `mustScan` row deleted from `carrier_caller_set_6861_test.go` | **ALIVE — by design**, see the note on the brief below. |
| **R3** | `dart` re-added to `knownMissingCarrier6847` | **RED**, with the exact-set message. The ledger deletion is load-bearing and confirms the fix at **corpus runtime**, not only under unit fixtures. |

### Claim 1 — the "there is no builder" closure: **survives**

I enumerated the Name sites myself rather than reading the table. `internal/extractors/dart/` (non-test) contains exactly five `types.EntityRecord{…}` literals (dart.go:152, 205, 410; types.go:167, 211) plus the `extractor.EnumEntity` call (types.go:96) — six Name sites, matching the PR. Sites 2–6 assign a `(\w+)` capture slice directly, with no derivation between capture and assignment. Every `+` in the package feeds `Signature`, `Properties`, a line count, or `dartModulePrefix` (which is not a Name). `FileCarrierFor` clause 3 reads `records[i].Name` only; `refs.go` additionally indexes `byQualifiedName`. One `FromID` site, as stated.

Site 1's nested half — the slicing argument — I did not take on the reading. An exhaustive probe over `{a . / : d r t _}` up to length 6 (**299,592 URIs**) confirmed `dartTopName`'s return is always a substring of its input and **never contains `/`**, with `dartTopName("main.dart.dart") == "main.dart"` as the positive control. Root depth is the only reachable depth, and that is a property of the function rather than an artefact of the closure — unlike cobol.

Site 4's `QualifiedName`: `EnumQualifiedName` = `"scope:enum:"` (11 bytes) `+ path + ":" + name`, `name` non-empty, so strictly longer than `path`. Both fields are asserted at both depths by the WordCharacter test.

I also read the fixture's record set rather than its label: the source produces 6 records named `api` (class, method, enum, typedef, and **two** modified classes), and every literal-path attempt yields either no record or a truncated `api` capture. The fixture varies what it claims and holds its neighbour constant.

### The two ALIVE mutants — adjudicated

**M5c — ALIVE, not production-reachable. Confirmed.** The distinguishing input would be a type-pass record (enum / typedef / modified class) named the path; impossible, and that impossibility is Claim 1, which I verified independently above rather than inheriting from the PR. Correctly classified and correctly named as the *entailed* conjunct.

**M5 — ALIVE. The verdict is right; the label under-states its dependence.** `idx := len(entities)` really is evaluated inside the class loop (dart.go:147), so the index axis is a genuine, unconditional equivalence — that half of the PR's account is exact, and finding it by scoring rather than by reading is the good part of this arm. But M5 also moves the call to a point where clause 3 cannot see the class / method / type records, so its full equivalence needs Claim 1 as well — the *same* second reason M5c rests on, which the PR attributes only to M5c. **R1 demonstrates the dependence:** with a builder on the class Name site, M5 and HEAD are distinguishable. So the precise three-way verdict is *"alive; equivalent on the index axis, not-production-reachable on the clause-3 axis"* rather than plain "equivalent". No grading is missing; the sentence is.

### The M7a distinction — verified, and it holds

`go test ./internal/extractors/dart/` is **green** under M7a: dart's cells genuinely cannot distinguish the variant. It dies only in `internal/extractor`, on exactly the two fixtures named — `TestFileCarrierFor_NoSecondCarrierWhenOneIsAlreadyNamedAfterThePath_6815` and `…WhenThePathNamedRecordComesLast_6815` — and nothing else in the four packages fires. The "always exactly one edge" premise is a struct literal with a single `RelationshipRecord`, and nothing in the package appends to an import record's `Relationships` (the method pass appends only through `classSpan.idx`).

**My verdict on the gap: acceptable, not a defect.** A local cell would have to hand clause 3 a path-named record owning zero relationships, which dart cannot emit — the only route is the import stub, which owns one by construction. Manufacturing that cell would mean asserting on a record shape production cannot produce, which is worse than the disclosure. The variant is graded, once, where the shape is free. The PR states this precisely and does not dress it up; that is the right call.

### Standing checks — re-derived from disk

- `knownMissingCarrier6847` = `{zig}`, `len` 1. `exercisedLanguages6847` and `anchoringLanguages6847` both still carry `dart` (verified in the file, not in the report). The `astro bicep … zig` list at line 103 is an explicitly historical snapshot and correctly still names dart. `knownMissingCarrier` has no other reference in the tree.
- `mustScan` row present. See the brief note below.
- **Ratchet run, not reported:** `scripts/quality/run.sh --ratchet --runs 1` → **exit 0**, *"31 gated fixtures held their baseline (22 at 100% must-have recall)"* — the PR's figure, reproduced. Premise verified: `internal/quality/golden/` holds 31 fixture directories plus `baseline.json` and **no dart fixture**; `run.sh` iterates only that directory, so the `testdata/fixtures/*/dart/` sources cannot reach it. `baseline.json` untouched (`git status` clean).
- **`cmd/grafel` digest:** no `.dart` string occurs anywhere under `cmd/grafel/`; `./cmd/grafel/` green in 166s, so the #6118 digest did not move.
- **`file_carrier.go` diff is empty.** Confirmed (`git diff HEAD~1 --` returns nothing).
- Suites, all `-count=1` and all GREEN: `./internal/extractors/dart/`, `./internal/extractors/`, `./internal/extractor/`, `./internal/resolve/`, `./cmd/grafel/`, `./internal/quality/...`. `gofmt -l internal/ cmd/` clean. Worktree `git status` clean at the end.
- **Not run by me:** the full repo suite, `-race`, the Windows/macOS legs, `grafel index` over a real corpus, and the ratchet at the default `--runs 5`.

### Nits (none blocking)

1. **Stale line reference.** `dart.go:349` — quoted in the PR body *and* in the shipped `file_carrier_6852_test.go` header — is the **pre-patch** line. At `d0c21d6ad` the `FromID: filePath` site is `dart.go:417`. The same commit that added the reference moved it.
2. **`dart.go`'s clause-3 paragraph mixes two consumers.** The `QualifiedName` length invariant sits inside the sentence about clause 3, but `FileCarrierFor` clause 3 compares `Name` **only**; the QN invariant is what `refs.go`'s `byQualifiedName` index needs. The test-file header draws the line correctly; the source comment could borrow its wording.
3. **One over-claim in the WordCharacter header.** It says the fixture "attempts the literal path spelling on **each** of them" for sites 2–6, but the method site gets only the stem (`void api()`) — there is no `void api.dart()` row. The parenthetical enumeration right after it is honest (four kinds). Harmless, since a dotted method name is not expressible, but the sentence claims a fifth attempt the body does not hold.

### One correction to the review brief, not to the PR

The brief asks for `mustScan` to be treated as a load-bearing shared counter. It is not one, and the file says so: the anchor list is documented as **redundancy** with a floor of 6 (16 entries today), and "any six of the seven could be dropped without a verdict changing". **R2 confirms it empirically** — deleting dart's row leaves `./internal/extractor` and `./internal/extractors` fully green. Adding the row is correct for consistency and costs nothing, but a future arm that forgets it will not go red *there*. The counter that is actually load-bearing is the #6847 ledger, and R3 proves it: re-adding `dart` fails with the exact-set message.

**APPROVE.** The closure is real and, for the first time in this series, exhaustively probed on its weakest half. The placement window is fully enumerated and the one narrowing found by scoring was written into the source rather than smoothed over. The M7a disclosure is accurate and the gap it names is the right thing to leave open.

Issue #5338, opened 2026-06-22
wizard completion: auto-index + group-name default to container folder + TUI hints/space + watcher install robustness


cajasmota:
Follow-up to #5336 (action-first wizard). Live testing surfaced:

## 1. Auto-index on group create / repo-or-module add
Today `grafel wizard` only registers + installs watchers (lazy index on first query). Make it ALWAYS trigger an explicit index after a group is configured (and after add-repo/add-module to a group), reusing `indexGroup → c.Rebuild` (`internal/cli/group.go:296`) and rendering the live CLI phase progress. Gate behind `--no-index` for scripting. Matches the dashboard (which rebuilds explicitly with progress).

## 2. Group-name default = container/parent folder, not a repo
From `another group/`, the wizard pre-fills the group name with `backend` (a child repo). Default to the **container/cwd folder name** (`another group`). Slug from the parent dir, not a selected repo.

## 3. TUI navigation hints
huh steps show no instructions; new users don't know **space** toggles selection. Add per-step hints (↑/↓ navigate · space select · enter confirm · / filter). 

## 4. TUI uses a tiny slice of the terminal
Each step renders in a few rows. Use more height (taller select/multiselect, scrollable), nicer per-step layout with the title + hint + body.

## 5. Watcher install robustness + delete cleanup
- `launchctl bootstrap com.grafel.watcher.<group>.<repo>` intermittently returns **exit 5 (EIO)**; despite bootout-before-bootstrap (`internal/daemon/service/manager.go`), it isn't retried → **retry bootout+bootstrap on err 5**.
- A watcher activation failure currently **aborts the wizard** (`internal/install/install.go:199`) even though the group config saved → make it a **non-fatal warning** (group still registered + indexable).
- **Delete (web + CLI) must bootout + remove the watcher launchd jobs/plists** — today deleting a group leaves `com.grafel.watcher.<group>.*` loaded + plists on disk, so recreate fights stale state.

labels: cli, dashboard, ux, bug, daemon

cajasmota:
## 6. `[ ]`/`[✓]` glyphs not actually rendering (secondary)
Live screenshots show the multiselect still uses huh's default `✓`/`•`, not the intended `[ ]`/`[✓]` brackets — the glyph customization in #5337 didn't take effect (wrong huh theme field, or default theme overrode it). Lower priority than the navigation instructions (#3), but fix the theme so brackets render — and verify it actually shows, don't just set the field.

Issue #5675, opened 2026-07-09
Monorepo + git worktrees: `missing .git` false positive on setup, then the daemon melts down re-indexing every worktree (and `MAX_WORKTREES_PER_REPO=0` won't disable it)


pmedranocurantis:
## Environment
- grafel `v0.1.7.3` (commit `fb8056e`, built 2026-06-27)
- macOS (Darwin 25.5.0), Apple Silicon, 10 cores
- Single-`.git` monorepo, registered in monorepo mode with **16 module slugs** (`domains`, `infrastructure`, `packages/*`)
- Repo contains **16 live `git worktree`s** under `.worktrees/` (created by an unrelated dev workflow)
- Installed via `grafel install`; daemon managed by launchd (`com.grafel.daemon`, `KeepAlive=true`)

## Summary
On a monorepo that also uses git worktrees, the daemon enters a permanent re-index storm and never becomes usable. Three distinct issues, in order of severity:

### 1. (Critical) Worktree indexing storm — each worktree indexed as a full-monorepo checkout, per module slug
In monorepo mode every registered module slug independently discovers **all** git worktrees and indexes each one. Because a worktree is a full checkout of the monorepo, each worktree index is a *full-monorepo* index (~19.5k files, ~288k entities, ~380 MB peak heap, ~9 min each in our repo).

With ~15 kept worktrees this enqueues ~15 full-monorepo indexes on **every daemon boot**. Observed in `daemon.log`:
```
indexer: completed repo=.../.worktrees/NOVA-1842 took=9m25.972s peak_heap_mb=383
indexer: completed repo=.../.worktrees/SUP-19810 took=1m7.312s peak_heap_mb=208
worktree: per-parent cap enforced group=. slug=curantis-monorepos-<each-of-16> total=15 kept=10 skipped=5 cap=10
```
Real (non-worktree) module slugs (`aws-common`, etc.) hang indefinitely behind this queue — after 14 min only `domains` and `infrastructure` had indexed; the other 14 never got a turn.

### 2. (Critical) fd exhaustion crashes the daemon → KeepAlive crash-loop
Mid-storm the daemon hits the file-descriptor ceiling and dies while persisting its worktree store:
```
ERROR worktree: failed to persist store err="open …/.grafel/worktrees.json.tmp: too many open files"
```
[...]

PR #6919, opened 2026-09-06
feat(#6918): `group add --track-worktrees` — worktree tracking without the watcher-unit install, for containers with no service manager

### LINKED ISSUE #6918: group add: no way to enable features.track_worktrees without also requesting OS watcher units (breaks containers with no service manager)
## The problem

Per-worktree tracking (`features.track_worktrees`) and OS watcher units (`features.watchers`) are two
independent capabilities, but `grafel group add` only exposes a switch for the second one — and turning
that switch on is currently the only way through the CLI to get the first.

`cmd/grafel/daemon.go:1152` opts a group into worktree discovery when `TrackWorktrees || Watchers`, so
`--watchers` happens to enable worktree tracking as a side effect. That coupling breaks for a real
consumer running grafel inside minimal Docker containers:

- PID 1 is `docker-init`; there is no systemd and no `systemctl` on `PATH`.
- The in-process fsnotify path that backs worktree tracking works fine there.
- The **watcher-unit install** does not, and is inapplicable in a container.

So the only way to get the capability they want is to also request one they cannot use.

## What `--watchers` does on a systemctl-less host today

It does **not** error — `internal/install/install.go:370` deliberately downgrades a `loader.Load`
failure to a `WatcherWarnings` entry (per #5338, so a flaky service manager cannot fail a whole wizard
run). That graceful degradation is correct and is not what this issue asks to change.

But the unit *file* is still written before that point, and `internal/install/watchers/loader_linux.go`
has **no `LookPath` guard and no unsupported-platform branch**, so a container install ends up with a
systemd unit on disk that nothing will ever load, plus a warning. That is a separate defect and is
worth filing on its own — it should NOT be fixed here.

## Proposed change

Add a `--track-worktrees` bool flag (default false) to `grafel group add` that sets
`features.track_worktrees=true` **without** the watcher-unit install transaction:

- It must not touch `SkipWatchers` in the install options — `--track-worktrees` alone must write no
  unit file and shell out to no service manager.
- It is independent of `--watchers`; the two compose.
- The daemon's read of `TrackWorktrees` is retroactive (re-evaluated per call), so an existing group
  can be re-registered with the flag and pick the capability up.

`internal/registry/registry.go:184` and `internal/cli/status.go:197` already model and report the
field; only the `group add` wiring is missing.

## Acceptance

A four-cell table over the two flags, asserting both the persisted `features` and — the point of the
change — the `SkipWatchers` the install transaction is handed:

| flags | `TrackWorktrees` | `Watchers` | `SkipWatchers` |
|---|---|---|---|
| `--track-worktrees` | true | false | **true** |
| `--watchers` | false | true | false |
| both | true | true | false |
| neither | false | false | true |

The first row is the regression this exists to prevent: wiring `--track-worktrees` into `SkipWatchers`
would silently re-introduce the container problem, and a test that only checked the field assignment
would not notice.

cajasmota:
Closes #6918.

## What

`grafel group add --track-worktrees` (bool, default false) sets
`features.track_worktrees` **without** the OS watcher-unit install transaction.

The daemon opts a group into worktree discovery when `TrackWorktrees || Watchers`
(`cmd/grafel/daemon.go:1152`, a retroactive per-call read), so before this the only CLI route to
per-worktree tracking was `--watchers` — which also writes a unit file and activates it through
launchd/systemd/schtasks. Inside a minimal Docker container (PID 1 = `docker-init`, no systemd, no
`systemctl` on PATH) the in-process fsnotify path works fine and the unit install does not, so the
consumer had to request a capability they cannot use in order to get the one they can.

The flag composes with `--watchers` rather than replacing it. `internal/registry/registry.go:184`
and `internal/cli/status.go:197` already model and report the field; only the wiring was missing.

## The load-bearing property

**`--track-worktrees` must never move `SkipWatchers`.** That is the whole point, and it is also the
one thing an obvious test would miss.

`install.Apply` gates its watcher branch on `!opts.SkipWatchers && opts.Config.Features.Watchers` —
**both** (`internal/install/install.go:303`). So the regression this PR exists to prevent, folding the
new flag into the option (`SkipWatchers: !(f.watchers || f.trackWorktrees)`), writes **no unit file
either**, because `Features.Watchers` is still false. A test asserting only the emitted artefact
would score that mutant ALIVE while it sat one unrelated edit away from re-introducing the container
problem.

So the option is asserted directly, through a one-line seam (`applyGroupConfigFn`) at the `group add`
call site, **and** the artefact is asserted alongside it.

## The four-cell table

`TestGroupAdd_TrackWorktreesIsIndependentOfWatchers` — each row reads the persisted config back off
disk (not the in-memory struct the command happened to build), the captured install option, and the
unit file:

| flags | `track_worktrees` | `watchers` | `SkipWatchers` | unit file |
|---|---|---|---|---|
| `--track-worktrees` | true | false | **true** | none |
| `--watchers` | false | true | false | written |
| both | true | true | false | written |
| neither | false | false | true | none |

Rows 2 and 4 were green before the implementation landed — they pin today's behaviour as unchanged.

Plus `TestGroupAddCmd_TrackWorktreesFlagDefaultsOff` (registration, bool, default false — the default
is load-bearing, `track_worktrees` is documented "opt-in to preserve existing behaviour") and
`TestGroupAddCmd_TrackWorktreesFlagIsBoundToTheField`, which drives the real cobra command so the
`RunE` closure binding is covered too, not only a hand-built `groupAddFlags`.

## Service-manager isolation, as evidence rather than as a stub

The `--watchers` rows reach `loader.Load`, a mutating verb, and are stubbed via
`watchers.StubServiceCallsForTest()` (`internal/install/watchers/test_isolation_guard.go` exists
because a test once drove a real `launchctl bootout` against the developer's machine).

The rows **without** `--watchers` are left deliberately unstubbed. The claim being made is
"`--track-worktrees` shells out to no service manager", and the guard is what enforces it: any
mutating verb reaching a real exec under `go test` panics. Stubbing those rows would replace that
proof with a no-op.

Verified the guard is a live control, not a dormant one — forcing the watcher branch on for the
track-worktrees-only row (SkipWatchers mutant + `Watchers = watchers || trackWorktrees`) panics:

```
panic: watchers: test attempted a REAL service-manager call that is not a known read-only probe:
launchctl [enable gui/501/com.grafel.watcher.demo.alpha-450a9830]
```

## Mutants

Scored at the call site (`internal/cli/group.go`), `-count=1`, `go vet` before each score, each edit's
post-application match count asserted as a hard gate (a silently unapplied edit reads ALIVE and
argues the opposite), restored by `cp` from a shasum-verified backup.

| # | mutant | applied | verdict |
|---|---|---|---|
| M1 | `SkipWatchers: !(f.watchers \|\| f.trackWorktrees)` — the regression this PR prevents | 1 | **DEAD** |
| M2 | flag default `false` → `true` | 1 | **DEAD** |
| M3 | drop `cfg.Features.TrackWorktrees = f.trackWorktrees` | 1 | **DEAD** |
| M4 | `cfg.Features.TrackWorktrees = f.watchers` | 1 | **DEAD** |
| M5 | `cfg.Features.Watchers = f.watchers \|\| f.trackWorktrees` | 1 | **DEAD** |
| M6 | drop the `RunE` binding (`trackWorktrees: false`) | 1 | **DEAD** |

No survivors, so no three-way verdicts to record. All-DEAD proves pinning rather than truth on its
own, which is why the positive control above is included: it demonstrates the isolation assertion is
capable of failing.

## Follow-up worth filing separately — NOT fixed here

`group add --watchers` on a systemctl-less host **degrades gracefully rather than erroring**:
`internal/install/install.go:370` turns a `loader.Load` failure into a `WatcherWarnings` entry, which
is deliberate (#5338 — a flaky service manager must not fail a whole wizard run). That is correct and
untouched.

But `internal/install/watchers/loader_linux.go` has **no `LookPath` guard and no unsupported-platform
branch**, so a container install still writes a systemd unit file to disk that nothing will ever load,
alongside the warning. Worth its own issue; out of scope here.

## Verified

- `go test -count=1 ./internal/cli/` — green (53s), full package, not a `-run` subset.
- `go vet ./internal/cli/` — clean.
- `gofmt -l internal/cli internal/registry` — clean.

**Not run, and why:** `./internal/registry/` — the package is read-only in this change, no file in it
was touched. `./cmd/grafel/`'s graph-digest test — this is a CLI-flag change touching no extractor and
no entity kind, so the digest is not implicated. The rest of the repo suite was skipped by instruction
(concurrent Go-compiling agents on a memory-constrained machine); CI covers it.

### DIFF
--- a/internal/cli/group.go
+++ b/internal/cli/group.go
@@ -37,6 +37,7 @@ func newGroupAddCmd() *cobra.Command {
 		reposCSV  string
 		groupDocs string
 		watchers  bool
+		trackWT   bool
 		gitHooks  bool
 		rules     bool
 		mcp       bool
@@ -73,22 +74,25 @@ new group immediately via the 'group' MCP parameter.
 
 Examples:
   grafel group add new-backend --repo core=/abs/path/to/repo --index
-  grafel group add legacy --repo /abs/api --repo /abs/web --no-watchers --json`,
+  grafel group add legacy --repo /abs/api --repo /abs/web --no-watchers --json
+  # container / no service manager: worktree tracking, no OS watcher units
+  grafel group add app --repo /abs/app --track-worktrees --json`,
 		Args: cobra.ExactArgs(1),
 		RunE: func(cmd *cobra.Command, args []string) error {
 			gaFlags := groupAddFlags{
-				repoArgs:  repoArgs,
-				reposCSV:  reposCSV,
-				groupDocs: groupDocs,
-				watchers:  watchers,
-				gitHooks:  gitHooks,
-				rules:     rules,
-				mcp:       mcp,
-				runInst:   runInst,
-				doIndex:   doIndex,
-				jsonOut:   jsonOut,
-				tools:     toolsCSV,
-				projGuide: projGuide,
+				repoArgs:       repoArgs,
+				reposCSV:       reposCSV,
+				groupDocs:      groupDocs,
+				watchers:       watchers,
+				trackWorktrees: trackWT,
+				gitHooks:       gitHooks,
+				rules:          rules,
+				mcp:            mcp,
+				runInst:        runInst,
+				doIndex:        doIndex,
+				jsonOut:        jsonOut,
+				tools:          toolsCSV,
+				projGuide:      projGuide,
 			}
 			return runGroupAddImpl(cmd, args[0], gaFlags, "")
 		},
@@ -102,6 +106,8 @@ Examples:
 		"optional path to shared group docs")
 	cmd.Flags().BoolVar(&watchers, "watchers", false,
 		"install OS watcher units for each repo")
+	cmd.Flags().BoolVar(&trackWT, "track-worktrees", false,
+		"enable per-worktree tracking (features.track_worktrees): the daemon discovers this group's linked git worktrees and registers them as ephemeral children. Independent of --watchers — it installs NO OS watcher units and shells out to no service manager, so it is usable in containers and on hosts with no systemd/launchd")
 	cmd.Flags().BoolVar(&gitHooks, "git-hooks", true,
 		"install git hooks (post-merge/checkout reindex)")
 	cmd.Flags().BoolVar(&rules, "rules", true,
@@ -122,19 +128,30 @@ Examples:
 }
 
 type groupAddFlags struct {
-	repoArgs  []string
-	reposCSV  string
-	groupDocs string
-	watchers  bool
-	gitHooks  bool
-	rules     bool
-	mcp       bool
-	runInst   bool
-	doIndex   bool
-	jsonOut   bool
-	tools     string // --tools CSV → GroupConfig.Tools (#5701)
-	projGuide bool
+	repoArgs       []string
+	reposCSV       string
+	groupDocs      string
+	watchers       bool
+	gitHooks       bool
+	rules          bool
+	mcp            bool
+	runInst        bool
+	doIndex        bool
+	jsonOut        bool
+	trackWorktrees bool   // --track-worktrees → Features.TrackWorktrees (#6918)
+	tools          string // --tools CSV → GroupConfig.Tools (#5701)
+	projGuide      bool
 }
+
+// applyGroupConfigFn indirects applyGroupConfig at the `group add` call site.
+//
+// It is a test seam and nothing else: `group add`'s job is to turn flags into a
+// GroupConfig plus a groupApplyOptions, and the second half of that — which
+// side effects the install transaction is authorised to perform — is otherwise
+// unobservable without performing them. install.Apply gates its watcher branch
+// on `!SkipWatchers && Features.Watchers`, so an option wired wrongly can look
+// identical on disk to one wired correctly (#6918).
+var applyGroupConfigFn = applyGroupConfig
 
 type groupAddRepo struct {
 	Slug string `json:"slug"`
@@ -180,6 +197,14 @@ func runGroupAddImpl(cmd *cobra.Command, group string, f groupAddFlags, socketPa
 
 	cfg := &registry.GroupConfig{Name: group, GroupDocs: f.groupDocs}
 	cfg.Features.Watchers = f.watchers
+	// #6918: an INDEPENDENT capability from Watchers, deliberately not folded
+	// into SkipWatchers below. The daemon opts a group into worktree discovery
+	// when TrackWorktrees || Watchers, so before this flag existed the only CLI
+	// route to worktree tracking was --watchers — which also runs the OS
+	// watcher-unit install, inapplicable inside a container whose PID 1 is
+	// docker-init and which has no systemctl. Setting this flag must leave the
+	// install transaction's watcher branch exactly as --watchers alone decides.
+	cfg.Features.TrackWorktrees = f.trackWorktrees
 	cfg.Features.GitHooks = f.gitHooks
 	// Per-tool selection (#5701): an explicit --tools value wins and is validated
 	// before any registration. When omitted, cfg.Tools is left empty so
@@ -205,7 +230,7 @@ func runGroupAddImpl(cmd *cobra.Command, group string, f groupAddFlags, socketPa
 	if f.jsonOut {
 		applyOut = io.Discard
 	}
-	res, err := applyGroupConfig(applyOut, cfg, groupApplyOptions{
+	res, err := applyGroupConfigFn(applyOut, cfg, groupApplyOptions{
 		RunInstall:      f.runInst,
 		SkipHooks:       !f.gitHooks,
 		SkipWatchers:    !f.watchers,
--- a/internal/cli/group_track_worktrees_6918_test.go
+++ b/internal/cli/group_track_worktrees_6918_test.go
@@ -1,0 +1,247 @@
+package cli
+
+// group_track_worktrees_6918_test.go — `--track-worktrees` enables per-worktree
+// tracking WITHOUT the OS watcher-unit install transaction.
+//
+// The consumer this exists for runs grafel in minimal Docker containers: PID 1
+// is `docker-init`, there is no systemd and no `systemctl` on PATH. The
+// in-process fsnotify path behind `features.track_worktrees` works fine there;
+// the watcher-unit install does not. Before #6918 the only CLI route to
+// `track_worktrees` was `--watchers` (the daemon opts a group in when
+// `TrackWorktrees || Watchers`, cmd/grafel/daemon.go), which drags the whole
+// unit-install transaction along with it.
+//
+// ── Why this asserts SkipWatchers and not just the unit file ────────────────
+//
+// install.Apply gates its watcher branch on `!opts.SkipWatchers &&
+// opts.Config.Features.Watchers` (internal/install/install.go:303) — BOTH.
+// So the regression this file exists to prevent, wiring --track-worktrees into
+// SkipWatchers (`SkipWatchers: !(f.watchers || f.trackWorktrees)`), writes no
+// unit file either, because Features.Watchers is still false. A test that only
+// looked at the emitted artefact would score that mutant ALIVE while it sat one
+// unrelated edit away from re-introducing the container problem. The option
+// handed to the transaction is therefore asserted directly, through the
+// applyGroupConfigFn seam, AND the artefact is asserted alongside it.
+
+import (
+	"bytes"
+	"io"
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/spf13/cobra"
+
+	"github.com/cajasmota/grafel/internal/install"
+	"github.com/cajasmota/grafel/internal/install/watchers"
+	"github.com/cajasmota/grafel/internal/registry"
+)
+
+// captureGroupApplyOptions swaps the applyGroupConfigFn seam for one that
+// records the options `group add` derives from its flags and then delegates to
+// the real implementation, so the install transaction still runs for real.
+func captureGroupApplyOptions(t *testing.T, got *groupApplyOptions) {
+	t.Helper()
+	prev := applyGroupConfigFn
+	applyGroupConfigFn = func(out io.Writer, cfg *registry.GroupConfig, ga groupApplyOptions) (*install.Result, error) {
+		*got = ga
+		return prev(out, cfg, ga)
+	}
+	t.Cleanup(func() { applyGroupConfigFn = prev })
+}
+
+// TestGroupAdd_TrackWorktreesIsIndependentOfWatchers is the four-cell table
+// over the two flags. The `--track-worktrees` row is the load-bearing one.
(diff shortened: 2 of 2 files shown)


cajasmota:
## Coordinator mutant — DEAD, but it is graded in a package this PR did not run

All six author mutants attack the **flag wiring** — the default, the assignment, the `RunE` binding, and `SkipWatchers` contamination. None attacks the thing that makes the flag *worth having*: that setting `features.track_worktrees` actually opts the group into worktree discovery.

That is a different file, a different package, and the entire point of the feature. For a container with no service manager, `--track-worktrees` alone must be sufficient — if it is not, the flag writes a field nothing acts on and the user is no better off than before.

**Mutant:** flip the daemon's opt-in from OR to AND at `cmd/grafel/daemon.go:1152`.

```go
- if !cfg.Features.TrackWorktrees && !cfg.Features.Watchers {   // opt in on EITHER
+ if !cfg.Features.TrackWorktrees || !cfg.Features.Watchers {   // require BOTH
```

Under the mutant, `--track-worktrees` alone no longer enables tracking — **precisely the container case this PR exists for**, and it would ship green against the PR's own test package. Occurrence gate 1; `go vet` clean before scoring.

**Verdict: DEAD.**

```
--- FAIL: TestDaemonWorktreeParents_SingleRepo
--- FAIL: TestDaemonWorktreeParents_MonorepoCollapses
--- FAIL: TestDaemonWorktreeParents_SeparateReposStaySeparate
--- FAIL: TestDaemonWorktreeParents_NonGitGraceful
FAIL  github.com/cajasmota/grafel/cmd/grafel
ok    github.com/cajasmota/grafel/internal/cli
```

### The finding: the grading is inherited, not added — and it lives where you did not look

Note the second line. **`./internal/cli/` passes under the mutant.** The kill comes entirely from `cmd/grafel/daemon_worktree_parents_5675_test.go`, which is **pre-existing on `main`** (it already sets `cfg.Features.TrackWorktrees = true` at `:46`) and which this PR does not touch — the diff is two files, both in `internal/cli`.

So the property holds, and no change is requested. But two things follow:

1. **Your `./cmd/grafel/` skip was right for the wrong reason.** You skipped it as "a CLI-flag change touching no extractor, so the digest is not implicated" — correct about the digest. But `cmd/grafel` also holds the only tests that grade this flag's *purpose*. Had they been red, the PR would have looked green everywhere you ran. The reasoning "no extractor, therefore not implicated" does not follow for a package that holds more than the digest.

2. **This arm's end-to-end value rests on a neighbour.** Same shape as #6908, where an arm's safety against a shrinking scan turned out to be inherited from `zeroKindSites()` rather than added by the arm. If `daemon_worktree_parents_5675_test.go` were ever narrowed, nothing in this PR would notice that `--track-worktrees` had stopped working.

Worth one line in the PR body naming `cmd/grafel/daemon_worktree_parents_5675_test.go` as what grades the flag end-to-end, so the next person editing that file knows a CLI feature depends on it.

Restored by `cp` from a shasum-verified backup (`e8544ed76…`); tree confirmed pristine; worktree removed.

Refs #6918.

cajasmota:

Reviewed at `7eb8ac54a` in a private detached worktree. Every claim below was re-derived, not taken
from the PR body. `go vet` clean, `gofmt -l internal/cli internal/registry` clean,
`go test -count=1 ./internal/cli/` green (57s, full package, restored tree verified `git status`
clean and the file's shasum equal to its pre-mutation backup).

### Claim 1 — the four-cell table: CONFIRMED, and the disk read-back is real

`registry.ConfigPathFor("demo")` → `registry.LoadGroupConfig(path)`. That is a genuine read off disk,
not an echo of the in-memory `cfg` the command built — the command's struct goes out of scope at
`runGroupAddImpl`, and the assertion runs against a freshly unmarshalled config. All four rows assert
three independent things (persisted `track_worktrees`, persisted `watchers`, the captured
`SkipWatchers`) plus the artefact.

### Claim 2 — the gate is a conjunction, so the seam is load-bearing: CONFIRMED, with direct evidence

`internal/install/install.go:303` reads verbatim:

```go
if !opts.SkipWatchers && opts.Config.Features.Watchers {
```

Both conditions, as claimed. I did not settle for reading it — **M1's failure output is the proof**.
Under `SkipWatchers: !(f.watchers || f.trackWorktrees)` the *only* assertions that fired were the two
`SkipWatchers` ones:

```
group_track_worktrees_6918_test.go:167: install options SkipWatchers = false, want true
group_track_worktrees_6918_test.go:245: --track-worktrees cleared SkipWatchers on the install transaction
```

The unit-file assertion stayed silent — because `Features.Watchers` is still false and no unit is
written. An artefact-only test scores that mutant ALIVE. My own **N2** (`SkipWatchers: false`
unconditionally) reproduces the same asymmetry on rows 1 and 4. The seam is not a test-only
convenience; it is the only thing that can observe this class of regression. It is one line of
production indirection, `var applyGroupConfigFn = applyGroupConfig`, and it earns it.

### Claim 3 — the positive control: REPRODUCED

Forcing the watcher branch on for the deliberately-unstubbed row (M5 + M1 applied together, match
count asserted at 2) panics under `go test`:

```
panic: watchers: test attempted a REAL service-manager call that is not a known read-only probe:
launchctl [enable gui/501/com.grafel.watcher.demo.alpha-4c821009]
```

The guard is a **live control, not a dormant one**, so the unstubbed rows passing is positive evidence
that `--track-worktrees` reaches no service manager. Confirmed `watchers.StubServiceCallsForTest()` is
invoked only under `if tc.watchers` and nowhere else in the file — the isolation claim is not
laundered through a stub.

### Claim 4 — mutants: all six reproduced DEAD, plus two of my own

Scored at the call site in `internal/cli/group.go`, each edit's post-application match count asserted
as a hard gate before the score (a silently unapplied edit reads ALIVE), `go vet` before every run,
`-count=1`, restored by `cp` from a shasum-verified backup — never `git checkout`.

| # | mutant | matches | verdict | caught by |
|---|---|---|---|---|
| M1 | `SkipWatchers: !(f.watchers \|\| f.trackWorktrees)` | 1 | **DEAD** | seam only (artefact silent) |
| M2 | flag default `false` → `true` | 1 | **DEAD** | registration test |
| M3 | drop `cfg.Features.TrackWorktrees = f.trackWorktrees` | 1 | **DEAD** | disk read-back, rows 1+3 |
| M4 | `TrackWorktrees = f.watchers` | 1 | **DEAD** | rows 1 **and 2** (both directions) |
| M5 | `Watchers = f.watchers \|\| f.trackWorktrees` | 1 | **DEAD** | disk read-back, row 1 |
| M6 | drop the `RunE` binding | 1 | **DEAD** | the real-cobra test |
| **N1** *(mine)* | `TrackWorktrees = f.trackWorktrees \|\| f.watchers` — **over-firing** | 1 | **DEAD** | row 2 (`--watchers` alone) |
| **N2** *(mine)* | `SkipWatchers: false` unconditionally | 1 | **DEAD** | seam only, rows 1+4 |

N1 matters because all-DEAD on the author's set could have meant the table only grades the
*under*-firing direction. It does not: row 2 (`--watchers` alone, `wantTrackWorktrees: false`) is a
forbidden row, and it is what kills both M4 and N1. Both directions are graded.

On the brief's suggested gap — *does anything assert the flag's effect on the daemon's opt-in read?*
Not in this PR, and it does not need to: `cmd/grafel/daemon_worktree_parents_5675_test.go` already
drives `daemonWorktreeParents()` against a config with `cfg.Features.TrackWorktrees = true`
(four cases). Dropping `TrackWorktrees` from `cmd/grafel/daemon.go:1152`'s condition would take all
four to zero parents. The producer (this PR) and the consumer (#5675) are each pinned; the seam
between them is the on-disk `track_worktrees` key, which both sides go through. That is adequate.

### Also assessed

**Help text and the `Examples:` block.** The new line is accurate and I verified it *parses*
(`newGroupAddCmd().Flags().Parse`): `grafel group add app --repo /abs/app --track-worktrees --json`
resolves cleanly. The flag's usage string is unusually long for a one-liner but every clause in it is
true, and it is the clause "installs NO OS watcher units and shells out to no service manager" that
the container user needs to read.

**One drive-by defect, pre-existing, in the line directly above the new example:**

```
grafel group add legacy --repo /abs/api --repo /abs/web --no-watchers --json
```

There is no `--no-watchers` flag. Only `cmd.Flags().BoolVar(&watchers, "watchers", false, ...)` is
registered, there is no `SetNormalizeFunc` and no `NoOptDefVal` anywhere in `internal/cli`, and pflag
does not synthesise a `--no-` form. I confirmed it empirically rather than by reasoning:

```
args [--repo /abs/api --no-watchers --json]: unknown flag: --no-watchers
```

So a documented example fails outright, and it is also redundant (`--watchers` already defaults
false). Not introduced here — but this PR reflows that exact line, and the container user who reads
the new example is reading it one line below a broken one. **Worth fixing in this PR** (delete
`--no-watchers`, or register the negative flag) or filing immediately; either is fine, it does not
block.

**Scope: `group add` only.** Right scope for the reported consumer, who scripts `group add`. Two
notes on the neighbours, neither a blocker:

- `internal/dashboard/v2_group_settings.go:392` loads the config **off disk** and overwrites only
  `Features.Watchers` / `Features.GitHooks`, so `track_worktrees` **survives** a dashboard PATCH. No
  data loss — just no way to turn it on from the UI.
- `internal/cli/wizard.go:137` builds `cfg := &registry.GroupConfig{}` **fresh**, so running the
  wizard over an existing group **silently drops `track_worktrees`**. Same for `onboard.go:55`, which
  hardcodes `Features.Watchers = true`. That is the inconsistency a user will actually hit — the
  container consumer who sets the flag and later re-runs the wizard loses it. It is pre-existing
  (nothing set the field before this PR, so nothing could be dropped), and correctly out of scope, but
  it is now reachable and deserves its own issue.

**The declared follow-up** (`internal/install/watchers/loader_linux.go`: no `LookPath` guard, no
unsupported-platform branch, so a container `--watchers` install writes a systemd unit nothing loads,
alongside the `WatcherWarnings` entry from `install.go:370`). Correctly out of scope — it is a
different code path in a different package, and this PR's whole point is that the container user need
not go down it at all. It is real and worth its own issue.

**Skipping `./cmd/grafel/`'s digest test: I agree.** The change adds a CLI flag and one assignment to
a config struct field that already existed. It touches no extractor, no entity kind, no ID
derivation. The `cmd/grafel` digest pin is for extractor-set changes; nothing here can move it.
(`internal/registry` untouched — agreed, no need to run it either, though it is cheap.)

### Anything wrong in the PR body

Nothing material. Every claim I checked held, including the ones that would have been easiest to
overstate — the disk read-back, the conjunction at `install.go:303`, and the panic transcript, which
reproduces down to the verb and the label shape (my digest differs, as it must — different temp path).

**APPROVE.** Recommend fixing or filing the `--no-watchers` example, and filing the two follow-ups
(the `loader_linux.go` guard, and the wizard/onboard path dropping `track_worktrees`).

PR #6920, opened 2026-09-06
fix(#6852): zig's import IMPORTS anchored on a path only `importTopSegment` can spell — arm 12, the LAST, and the ledger reaches EMPTY

### LINKED ISSUE #6852: extractors: path-anchored FromID with no file carrier — COMPLETE, 12 of 12 arms shipped (+dockerfile unledgered); the twelve was a lower bound

cajasmota:
Arm 12 of 12. Zig has **exactly one** `FromID` site — `buildImportEntities` → `FromID: filePath` on IMPORTS — now fixed with a conditional `extractor.PrependFileCarrier`. `knownMissingCarrier6847` goes from `{zig}` to **`{}`**.

**The site table, all four Name producers, proven not asserted.** `zig.go` has 4 `Name:` sites and 1 `FromID:` site, and those counts are now READ OUT OF THE SOURCE on every run by `TestZig_SourceSiteCountsAreWhatThisFileAssumes_6852` (which also pins `QualifiedName:` at 0 — zig sets none, so the resolution question here is a Name question only). Three sites are verbatim `(\w+)` captures, OBSERVED by driving an adversarial fixture. The fourth is `importTopSegment`.

**The closure survives concatenation because the enumeration IS the anti-concatenation property.** `import_top_segment_6852_test.go` brute-forces all **5461** targets over `{a _ . /}` up to length 6 and establishes that `importTopSegment`'s result is always a **SUBSTRING of its input**. Cobol's closure died to a builder concatenating a prefix onto a literal-derived operand; that exact edit fails this test on its first input rather than silently invalidating prose. The second enumerated property — a '/'-bearing result always **ends** in '/' — is what makes the depth split a fact about the function.

**Both clause-3 cells are DRIVEN, neither is argued away.** Root: `importTopSegment` trims exactly ONE extension, so `main.zig` importing `"main.zig.zig"` names the import stub exactly the path; clause 3 declines and the second node under one `graph.EntityID` (which does not hash Subtype — #6369/#6480) is never minted. Nested: the only shape the enumeration leaves open is a path ending in '/', so that cell is driven too (path `src/`) and reported as **alive but not production-reachable — the input is a path ending in '/'**, which no file on disk has. A nested contrast subtest stops both passing on a carrier that is never emitted at all.

**Placement enumerated, not argued — 6 positions.** P1 top of `extractZig` (M3 **DEAD**), P2 between the fn passes (M4 **DEAD**), P3 before the import pass (M5 **DEAD**) — all three mint nothing, because zig's ONLY path-anchored edge is built in the import pass, so clause 2 rejects and the defect survives silently. P4 end of `extractZig` (shipped). P5 in `Extract` before tagging (M6 **ALIVE**) and P6 after tagging (M7 **ALIVE**) are **equivalent, not merely ungraded**: the carrier owns zero relationships and carries a non-empty explicit Language, so neither tagging call can touch it and the output is byte-identical. Tail-append instead of index 0 is M10, **DEAD**. The compound is scored part by part with the **entailed** conjunct named: clause-2 reachability is the independent reason; clause-3 visibility is **entailed**, because the only record that can be named after the path is an import stub built in the same pass as the anchor. Zig stores no index into `entities`, so dart's index-shift hazard has no analogue here — stated rather than inherited.

**Mutants — 13 author mutants, every edit gated on an exact match count before scoring, `go vet` before each, restored by `cp` from shasum-verified backups.**

| # | mutant | verdict |
|---|---|---|
| M1 | drop the `PrependFileCarrier` call | DEAD |
| M2 | unconditional `FileEntity` instead | DEAD (over-firing control + empty-path) |
| M3 | placement P1 | DEAD |
| M4 | placement P2 | DEAD |
| M5 | placement P3 | DEAD |
| M6 | placement P5 (Extract, pre-tagging) | ALIVE — **equivalent** (identical output) |
| M7 | placement P6 (Extract, post-tagging) | ALIVE — **equivalent** (identical output) |
| M8 | empty lang token | DEAD via the `Properties["language"]` provenance stamp |
| M9 | wrong token `"ziglang"` | DEAD on `Language` |
| M10 | append at tail, not index 0 | DEAD (#577's index-0 convention) |
| M11 | **premise mutant** — `importTopSegment` strips ALL extensions | DEAD: the clause-3 offender ceases to exist and the root cell says so |
| T1 | travelling: clause 3 `&& len(rels) > 0` | ALIVE in zig, **DEAD in `./internal/extractor/`** |
| T2 | travelling: clause 3 `&& len(rels) == 0` | **DEAD on zig's own cells**, ALIVE in `./internal/extractor/` |

**The travelling mutant, with the honest split.** T1 is *indistinguishable inside this package* and the reason is checkable: zig's only path-named record is an import stub, and `buildImportEntities` gives every stub exactly one IMPORTS edge, so a path-named record with zero relationships is a shape zig cannot emit. Graded elsewhere, gap disclosed — dart's verdict, same mechanism. T2 is the interesting one: it is **ALIVE on the generic #6815 fixtures** (whose path-named records own no edges) and **DEAD on zig's cells**. Zig is where the `== 0` variant is graded at all. Series tally: DEAD on astro · ALIVE on clojure · ALIVE on cobol until a new cell · graded on crystal by disjoint cells · graded-elsewhere on dart · **T1 graded-elsewhere / T2 graded HERE on zig**.

**The emptied guard was restated, because emptying it opened a hole.** With entries, `TestFileAnchoredCarrier_NoNewOffender_6847`'s "every known offender still reproduces" loop was this file's strongest anti-vacuity check; empty, it iterates nothing and can no longer fail. Measured, not assumed:

- **V1** — `pathAnchoredKindsMissingCarrier` returns nil unconditionally: with the ledger empty this used to be a clean green run. Now **DEAD**, via a new per-file positive control run *through the walk's own call*: for every anchoring corpus file, blanking the carrier's name fields must make the matcher REPORT.
- **V2** — the walk stops recording offenders: **ALIVE**, and correctly so — the offender list is now the diagnostic, not the assertion. Its safety is proved rather than asserted by **V4**.
- **V3** — zig loses its carrier again, guard unmutated: **DEAD**, three assertions firing.
- **V4** — zig loses its carrier **AND** V2 is applied: still **DEAD** ("3 corpus files anchor on their own path but only 0 carry"). That is the proof the emptied guard remains a meaningful assertion.

`TestFileAnchoredCarrier_EveryAnchoringFileCarries_6847` states the invariant in the positive direction over a **non-empty** population — per language and per file count, with a floor of 100 accounted files — and its `carried` count is derived **directly from the records, not from the offender list**. Deriving it from the matcher was tried first and was wrong: a walk that stopped calling the matcher would then count every anchoring file as carried, and a real regression would read green twice over. The positive control also **un-names** the carrier rather than deleting it — deleting was measured and was wrong, because for java, python, typescript, javascript and graphql the path-named record is the same record that OWNS the anchored edges.

**Prose, since nothing downstream corrects the last arm.** `file_carrier.go`'s clause-3 paragraph gains zig's route and the enumeration lesson, plus an explicit note that "the ledger is empty" is a statement about **#6847's CORPUS**, not about the registered language set — reasonml and rescript remain confirmed offenders that corpus never reaches, and the population is still unbounded above. Same-file references name functions, not line numbers.

**Fixture axes.** `TestZig_ImportStubNamedLikeThePath_6852` varies exactly one axis per case — the path's **depth** (root / nested / trailing-slash) — holding the source bytes and the import target **byte-identical** across the first two. `TestZig_NoCarrierWithoutAnImport_6852` varies only the **presence of an `@import`**, holding every other pass (pub fn, private fn with a call, struct with a method) constant. `TestZig_OneCarrierPerFileNotPerImport_6852` varies only the **number of distinct import targets** (5 vs 2), holding path, depth and passes constant.

Verification: `./internal/extractors/zig/`, `./internal/extractors/`, `./internal/extractor/`, `./internal/resolve/`, `./cmd/grafel/` all green at `-count=1`; the whole-graph digest is unmoved (`custom_extractor_spans_6118_test.go` has no zig member — grepped, 0 occurrences); ratchet `scripts/quality/run.sh --ratchet` exit 0, 31 gated fixtures held baseline, 22 at 100% must-have recall, premise confirmed (no zig fixture exists under `internal/quality/golden/`). NOT run: the full repo suite, `-race`, and any Windows/Linux leg — CI has those.

Closes #6852. Refs #6847, #6815, #6861.

### DIFF
--- a/internal/extractor/carrier_caller_set_6861_test.go
+++ b/internal/extractor/carrier_caller_set_6861_test.go
@@ -208,6 +208,7 @@ func TestCarrierCallerSetIsGradedFromSource_6861(t *testing.T) {
 		{"internal/extractors/cobol/extractor.go", []string{prependFileCarrier}},
 		{"internal/extractors/crystal/extractor.go", []string{prependFileCarrier}},
 		{"internal/extractors/dart/dart.go", []string{prependFileCarrier}},
+		{"internal/extractors/zig/zig.go", []string{prependFileCarrier}},
 	}
 
 	nonTest := 0
--- a/internal/extractor/file_carrier.go
+++ b/internal/extractor/file_carrier.go
@@ -111,6 +111,29 @@ import "github.com/cajasmota/grafel/internal/types"
 //     over "every Name this package emits" is the claim most likely to be
 //     wrong, because one builder concatenating a prefix onto a literal-derived
 //     operand defeats it — clojure's `.clj-kondo` correction, one package over.
+//     zig (#6852) is the LAST arm, and it answers that lesson with the strongest
+//     form the series found. Its one candidate name site, importTopSegment,
+//     trims exactly ONE trailing extension, so a ROOT main.zig importing
+//     "main.zig.zig" names an import stub exactly the path and clause 3
+//     declines — driven, not argued. What settles the NESTED question is not a
+//     character-class closure at all: the arm brute-forces the whole input
+//     space and establishes that importTopSegment's result is always a
+//     SUBSTRING of its input. That is the anti-concatenation property itself,
+//     asserted over the function rather than over a paragraph, so the edit that
+//     defeated cobol's closure — concatenating a prefix onto a literal-derived
+//     operand — fails a test here instead of quietly invalidating prose. The
+//     nested cell that remains (a path ending in '/') is DRIVEN too, and
+//     reported as alive-but-not-production-reachable by naming the input.
+//     Pinned by TestZig_ImportStubNamedLikeThePath_6852 and, for the
+//     enumeration, TestZig_ImportTopSegmentIsAlwaysASubstring_6852.
+//
+// WITH THE ZIG ARM, #6852'S LEDGER REACHED EMPTY over #6847's corpus:
+// knownMissingCarrier6847 has no entries left. That is a statement about a
+// CORPUS, not about the registered language set — reasonml and rescript are
+// confirmed offenders that corpus never reaches, and the population stays
+// unbounded above until every registered language is driven. Read
+// len(knownMissingCarrier6847) and the coverage sets beside it, never this
+// sentence, for the current state.
 //
 // Clause 3 is checked for EVERY record, before the loop may short-circuit on
 // clause 2 being satisfied — deliberately, and the order is load-bearing.
--- a/internal/extractors/file_anchored_carrier_runtime_6847_test.go
+++ b/internal/extractors/file_anchored_carrier_runtime_6847_test.go
@@ -126,12 +126,17 @@
 //
 // THE LEDGER IS AN EXACT SET, NOT A FLOOR, IN BOTH DIRECTIONS. A language not
 // in knownMissingCarrier6847 fails this test; so does fixing one that IS in it
-// without removing its entry. NOTE THAT THE SET SHRINKS AS LANGUAGES ARE
-// FIXED — it opened at twelve and #6852 removes one per landed arm (bicep is
-// the first), so any count written here goes stale by design. Read
-// len(knownMissingCarrier6847), not a number in prose. A guard that only checked "no NEW offender" would be blind to the
-// permissive direction — a walk that silently stopped detecting anything would
-// stay green under a floor and goes red here.
+// without removing its entry. THE SET SHRANK AS LANGUAGES WERE FIXED — it
+// opened at twelve, #6852 removed one per landed arm (bicep first, zig last),
+// and it is now EMPTY over this corpus. Read len(knownMissingCarrier6847), not
+// a number in prose.
+//
+// AN EMPTY LEDGER CHANGES WHAT THIS FILE PROVES, and the change is written up
+// at the declaration below rather than left implicit: the "still reproduces"
+// direction now iterates nothing, so the permissive failure it used to catch —
+// a walk that silently stopped detecting anything — is caught by
+// TestFileAnchoredCarrier_EveryAnchoringFileCarries_6847 instead, which
+// accounts for a non-empty population in the positive direction.
 package extractors_test
 
 import (
@@ -163,9 +168,22 @@ import (
 // for the conditional-carrier shape #6815 and #6518 settled on) and delete the
 // line. To ADD one: do not. A new offender is the defect this file exists to
 // catch; adding a line to keep it green is the failure mode #6834 names.
-var knownMissingCarrier6847 = map[string]string{
-	"zig": "zig.go:272 `FromID: filePath` on IMPORTS; no carrier emitted",
-}
+// THE SET IS NOW EMPTY, AND THAT CHANGES WHICH ASSERTION IS LOAD-BEARING.
+// While it had entries, the second loop of TestFileAnchoredCarrier_NoNewOffender_6847
+// ("every known offender still reproduces") doubled as this file's strongest
+// anti-vacuity check: a walk that silently stopped detecting anything went RED
+// there. Empty, that loop iterates nothing and can no longer fail, so "zero
+// offenders" became a conclusion a broken walk reaches just as easily as a
+// fixed tree does — the #6908 shape, an anti-vacuity guard relaxed by the very
+// change it exists to catch.
+//
+// The replacement is TestFileAnchoredCarrier_EveryAnchoringFileCarries_6847
+// below, which states the SAME invariant in the positive direction over a
+// population that is not empty and cannot become empty without failing:
+// every one of the anchoringLanguages6847 files must be MATCHED to a carrier,
+// per language and per file count. Deleting the detection step from the walk
+// now fails there instead of passing here.
+var knownMissingCarrier6847 = map[string]string{}
 
 // exercisedLanguages6847 is the exact set of registered languages the corpus
 // actually drives. It is pinned so that a corpus file being deleted, renamed
@@ -268,6 +286,8 @@ type carrierScanResult struct {
 	candidates       int
 	exercised        map[string]int
 	anchoring        map[string]int
+	carried          map[string]int
+	detected         map[string]int
 	noExtractorLangs map[string]int
 	offenders        []carrierFinding
 }
@@ -335,6 +355,8 @@ func scanCorpusForMissingCarriers() (carrierScanResult, error) {
 		candidates:       len(files),
 		exercised:        map[string]int{},
 		anchoring:        map[string]int{},
+		carried:          map[string]int{},
+		detected:         map[string]int{},
 		noExtractorLangs: map[string]int{},
 	}
 	cls := classifier.New(nil)
@@ -383,6 +405,46 @@ func scanCorpusForMissingCarriers() (carrierScanResult, error) {
 		}
 		if anchored {
 			res.anchoring[cr.Language]++
+			// PER-FILE POSITIVE CONTROL FOR THE MATCHER, run through the walk's
+			// own call rather than beside it. Take the carrier away and the file
+			// becomes a genuine offender; if the matcher does NOT report it, the
+			// matcher is not detecting anything and the empty offender list
+			// below is a no-op that reads like a guard. This is the check that
+			// replaces what an entry in knownMissingCarrier6847 used to provide:
+			// with the ledger empty, nothing else here fails when the detection
+			// step is neutered.
+			//
+			// THE CARRIER IS UN-NAMED, NOT DELETED, and the distinction is not
+			// cosmetic — deleting it was measured and was WRONG. For java,
+			// python, typescript, javascript and graphql the record that carries
+			// the path is the same record that OWNS the path-anchored edges, so
+			// removing it removes the anchor too and the matcher correctly
+			// reports nothing. Blanking the two name fields leaves the anchor in
+			// place and isolates exactly the property under test.
+			stripped := make([]types.EntityRecord, len(recs))
+			copy(stripped, recs)
+			for i := range stripped {
+				if stripped[i].Name == vrel || stripped[i].QualifiedName == vrel {
+					stripped[i].Name = ""
+					stripped[i].QualifiedName = ""
+				}
+			}
+			if len(pathAnchoredKindsMissingCarrier(vrel, stripped)) > 0 {
+				res.detected[cr.Language]++
+			}
+			// The POSITIVE half of the invariant, derived DIRECTLY from the
+			// records and NOT from the offender list below. Deriving it from
+			// `kinds` was measured and was wrong: a walk that stopped calling
+			// the matcher would then count every anchoring file as carried and
+			// report no offenders, so a real regression would produce a green
+			// run twice over. Read straight from the records, that same walk
+			// mutant leaves this count intact and the regression still fails.
+			for i := range recs {
+				if recs[i].Name == vrel || recs[i].QualifiedName == vrel {
+					res.carried[cr.Language]++
+					break
+				}
+			}
 		}
 		if kinds := pathAnchoredKindsMissingCarrier(vrel, recs); len(kinds) > 0 {
 			res.offenders = append(res.offenders, carrierFinding{lang: cr.Language, path: vrel, kinds: kinds})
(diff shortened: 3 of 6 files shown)


cajasmota:

Reviewed at `c8fec2307` in a detached worktree of my own. Everything below is measured in that tree, `-count=1`, `go vet` clean before each score, every edit gated on an exact match count, restores by `cp` from shasum-verified backups (`git status --porcelain` empty at the end).

Baseline green: `./internal/extractors/zig/` `./internal/extractors/` `./internal/extractor/` `./internal/resolve/` `./cmd/grafel/` (the last at 167s — that is the whole-graph digest, verified by running rather than by the grep premise).

**Not run:** the full repo suite, `-race`, any Windows/Linux leg, and `scripts/quality/run.sh --ratchet` (premise verified by enumeration instead — see below).

### Scorecard (my own reproductions)

| mutant | reported | measured |
|---|---|---|
| M1 drop `PrependFileCarrier` | DEAD | **DEAD** — 6 zig tests + both 6847 tests, with the 3 offending corpus files named |
| M2 unconditional `FileEntity` | DEAD | **DEAD** (over-firing + empty-path + clause-3) |
| M5 placement P3 (before the import pass) | DEAD | **DEAD** — 5 zig tests |
| M7 placement P6 (`Extract`, post-tagging) | ALIVE/equivalent | **ALIVE — equivalent, and by a stronger argument than claimed** (below) |
| M8 empty lang token | DEAD | **DEAD** on `TestZig_CarrierShape_6852` |
| M9 wrong token `ziglang` | DEAD | **DEAD** |
| M10 tail-append | DEAD | **DEAD** — `TestZig_ImportsFromEndResolves_6852` + `TestPrependFileCarrier_CarrierIsFirstAndOrderIsKept_6815` |
| M11 premise: strip ALL extensions | DEAD | **DEAD** — the root clause-3 cell is a real shape, not a fabricated one |
| T1 `&& len(rels) > 0` | ALIVE zig / DEAD extractor | **exactly that** (extractor: `..._NoSecondCarrierWhenOneIsAlreadyNamedAfterThePath_6815`, `..._ComesLast_6815`) |
| T2 `&& len(rels) == 0` | DEAD zig / ALIVE extractor | **exactly that** |
| V1 neuter the matcher | DEAD | **DEAD** — and see the hole check |
| V2 stop recording offenders | ALIVE | **ALIVE** |
| V3 zig loses its carrier | DEAD | **DEAD** |
| V4 = V3 + V2 | DEAD | **DEAD** (`zig: 3 anchor, 0 carry`) |
| *(mine)* delete-instead-of-un-name in the positive control | — | **DEAD** — see correction #1 |

### Claim 1 — the enumeration's alphabet: ADEQUATE

`5461 = sum(4^k, k=0..6)` confirmed. I did not take the alphabet on trust: I re-ran both properties over a **7-symbol alphabet** — `a`, a **digit** `9`, `-`, a **backslash**, `.`, `/`, and a **multi-byte rune** `é` — to length 5, **19,608 inputs**. Both properties hold with zero counterexamples.

That is empirical; the structural reason is what actually closes it, and it is checkable by reading `importTopSegment`: the function branches on exactly three things (the `./` / `../` prefixes, `'/'`, `'.'`) and is otherwise byte-opaque, so `a` is a faithful representative of every other byte. It only ever *slices* — prefix strip, suffix-after-last-`/`, prefix-before-last-`.`, two `return mod` fall-backs — hence substring-only unconditionally, not just on 5461 inputs. And after the slash step `trimmed` provably contains no `/` at all, so a `/`-bearing result can only come from a `return mod`, both of which are reached only when the basename slice came out empty, i.e. when the input ends in `/`. UTF-8 aliasing checked too: continuation bytes are ≥0x80, so no multi-byte rune can contain 0x2E or 0x2F and `LastIndexByte` cannot be fooled. **Digits, `-`, `\`, unicode, empty target, leading `/` — none can change the outcome.** The closure is sound, and unlike astro/clojure/cobol it is sound for a reason that does not depend on the enumeration being wide enough.

### Claim 2 — the emptied #6847 guard: the hole was real, and the restatement closes it

I reproduced the hole rather than accepting the report. With V1 applied (matcher returns nil unconditionally):

- `-run 'TestFileAnchoredCarrier_NoNewOffender_6847|TestFileAnchoredCarrier_CorpusCoverage_6847'` → **`ok`**. The pre-existing assertions are completely blind to it with an empty ledger.
- `-run TestFileAnchoredCarrier` → **FAIL, and only `..._EveryAnchoringFileCarries_6847` fails.**

So the new test is the sole killer, the hole was exactly as described, and it is closed.

**The five no-op ways, audited individually** (not just the one the positive control covers):

1. *fail to read* — `res.candidates >= minCorpusFiles6847` (700) plus the new `total >= 100` floor. Covered.
2. *read the wrong files* — three exact set diffs (`exercised`, `noExtractor`, `anchoring`). Covered.
3. *right files, wrong content* — **partially covered.** Nothing pins `res.anchoring[lang]` to a number; `carried` and `detected` are both compared *to `anchoring`*, so they are self-relative, and the only floors are global. A per-language erosion (zig's corpus going 3 files → 1) fails nothing. Not a blocker — the global floors bound total erosion and the language *set* is exact — but it is the one residual vacuity shape here, and I would take a per-language floor in a follow-up.
4. *fail to detect* — the new `detected` control. Covered, measured (V1).
5. *detect and not act* — V2 is ALIVE, correctly: `carried` is the assertion now, and V4 proves the consequence is still caught.

**`carried` derived from the records, not the matcher — verified independently.** It re-implements the `Name == vrel || QualifiedName == vrel` test directly over `recs`; it does not call `pathAnchoredKindsMissingCarrier`. Under V1 the count is intact and V3+V1 still fails. The self-correction is real.

**Is the guard still worth what it costs, with the population at zero?** Yes — but its centre of gravity has moved and that should be said plainly. V3 shows `NoNewOffender` still fires on a real regression with the three offending paths named, so it is a live regression net, not a shape. What is now dead is the *second loop* over `knownMissingCarrier6847`: it can never fail while the map is empty. The prose says so, which is the right mitigation, so the rot risk is documentation drift rather than a false green. I would not delete the file; I would revisit if the map is still empty in three arms' time.

### Claim 3 — the travelling mutant's split

Reproduced exactly, and the split is genuinely disjoint: T1 is graded only outside the package, T2 **only inside it**. Zig is the first arm in the series where the `== 0` variant is graded at all — that claim holds.

The **"exactly one edge" premise is real**, verified by reading rather than inherited from dart: `buildImportEntities` builds every stub with a single-element `[]types.RelationshipRecord` literal, unconditionally, on the only path that emits a stub. There is no code path producing a path-named record with zero relationships in this package. T1's in-package ALIVE is therefore genuinely indistinguishable, and the out-of-package kill is on the two dedicated clause-3 fixtures — not incidental.

### Claim 4 — the two ALIVE placement mutants

**Verdict: equivalent — and by a stronger argument than the PR body gives.** The body says "neither tagging call can touch it". I checked both taggers: `TagEntitiesLanguage` skips any record with a non-empty explicit `Language` (and the carrier's is `"zig"`); `TagRelationshipsLanguage` only walks `records[i].Relationships`, and the carrier owns none (pinned). So P5/P6 are not "equivalent under the current suite" — they are **equivalent under any suite**, because the output is identical, and no test could distinguish them. That is the correct three-way verdict and it does not need a follow-up.

**The index-shift claim verified by reading, not inherited:** `grep -n 'entities\['` in `zig.go` returns nothing; the struct pass hangs CONTAINS on `BuildOperationStructuralRef`, and the import pass does `append(importEntities, entities...)` rather than positional surgery. Dart's hazard genuinely has no analogue here.

### Claim 5 — the source-site count test: NOT fooled

It uses `strings.Count` over the whole file, not a line-oriented grep, so the string-literal trap does not apply — and `zig.go` embeds no Go source in literals anyway. Counts confirmed independently: 1 `FromID:`, 4 `Name:`, 0 `QualifiedName:` (the single `QualifiedName` occurrence is prose, no colon), and no `.Name =` assignment sites outside composite literals. Two nits, neither blocking:

- `"Name:"` is a substring of `"QualifiedName:"`, so the two rows are coupled. Harmless today (QN = 0) and it fails loudly rather than silently, but the 4 is not independent of the 0.
- The test reads **`zig.go` only, not the package**. A second non-test `.go` file in package `zig` with a `Name:` site would evade it. Today `zig.go` is the only non-test file in the package (verified), so the count is complete.

### Corrections

1. **PR body, the un-name-vs-delete measurement, is understated.** I applied the delete variant myself. Reported affected: java, python, typescript, javascript, graphql. **Measured: those five (all → 0 detected) *plus* `markdown`, which drops 17 → 13.** The self-correction is right and load-bearing; the roster attached to it is one language short.
2. **The brief is wrong that `file_carrier.go` should be untouched.** It gains 23 lines. All 23 are comment — the code is byte-identical, confirmed by filtering the diff. The author disclosed this; the brief did not.
3. Ratchet premise confirmed by enumeration rather than by a 40-minute run: 31 golden dirs, **zero** `.zig` files under `internal/quality/`, zero `zig` hits anywhere in `internal/quality/golden/`. The only `.zig` files in the repo are the three under `testdata/fixtures/` that the #6847 corpus walk reads. A zig extractor change provably cannot move the golden recall numbers. I did not run the ratchet; I am relying on that enumeration and on the author's exit-0 report for the run itself.

### Shared counters, re-derived from disk

`knownMissingCarrier6847` = `{}` ✓ · `exercisedLanguages6847` and `anchoringLanguages6847` unchanged and both still contain `zig` (it always anchored — only the carrier was missing) ✓ · `mustScan` gains the `zig.go` row, 17 entries against a floor of 6 ✓ · `carried`'s set is compared against `anchoringLanguages6847`, so *every* anchoring language must now fully carry — a real tightening ✓.

Nothing here rests on prose the tests do not observe. The series closes cleanly.

cajasmota:
## Coordinator mutant — ALIVE, and it locates the guard's sensitivity boundary exactly

The review named a residual gap but did not score it: *"nothing pins `res.anchoring[lang]` to a number, and `carried`/`detected` are compared **to** `anchoring`, so a per-language corpus erosion (zig 3→1) fails nothing."* That is the one no-op mode the restated guard does not close, and it is worth a measurement rather than a note — especially on the arm that empties the ledger, because from here the guard is the only thing standing.

**Mutant: simulate corpus erosion.** Make one language contribute only its first anchoring file. `exercised`, `anchoring`, `carried` and `detected` all shrink together, so every **relative** comparison still balances; only an absolute per-language floor could notice. Occurrence-gated at 1, `go vet` clean.

Two runs, and the boundary is sharp:

| erosion | verdict |
|---|---|
| zig 3 → **1** anchoring file | **ALIVE — suite green** |
| zig 3 → **0** anchoring files | **DEAD** — `EveryAnchoringFileCarries_6847` and `CorpusCoverage_6847` both fail with `missing (expected, not observed): [zig]` |

The second run is the non-vacuity proof: the mechanism *does* fire, so the first result is a real gap rather than a mutant that failed to apply.

**So the guard detects "a language disappears entirely" and not "a language erodes partially."** `diffStringSets(anchoringLanguages6847, sortedStringKeys(res.carried))` is a **set** comparison — presence, not count. As long as one file of a language still anchors, the language stays in the set and every count comparison balances at the reduced number.

### Why this is worth recording rather than fixing here

Not blocking, and I am not asking for a change in this PR — the arm is correct and the restatement genuinely closed the hole it was written for (V1 is DEAD, and `EveryAnchoringFileCarries_6847` is its sole killer).

But the failure mode is the quiet kind this repo keeps finding. If a future change causes zig to stop anchoring in 2 of its 3 corpus files — a regex narrowed, a fixture moved, an extractor arm rewritten — **the guard reports green while measuring a third of what it used to.** That is the "vacuous walk reports clean" shape one level up: not a walk that does nothing, but a walk whose population silently shrinks under it.

The fix is cheap and mechanical: pin `res.anchoring[lang]` to an expected count per language with a floor, so erosion trips a number rather than passing a set membership test. Filing it separately rather than expanding this PR — the twelve-arm series should close on the arm that finished it, not on a guard refactor.

Also worth noting for whoever takes that issue: the review found the author's un-name-vs-delete roster **one language short** — `markdown` is affected too (17→13 detected), on top of the five named. A per-language count pin would have caught that as a number rather than as a reviewer's second look.

Restored by `cp` from a shasum-verified backup (`073e0d1ef…`); tree confirmed pristine; worktree removed.

Refs #6852, #6847.

Issue #6180, opened 2026-08-05
fsnotify on macOS opens an FD per file: 39,894 descriptors for one repo against a 61,440 ceiling


cajasmota:
**Measured on a live engine process: 39,894 file descriptors held for a single repo, against a hard per-process ceiling of 61,440.** One repo is at 65% of the limit. A 140-repo store cannot work on macOS.

## Why the cost is per-file, not per-directory

`Watcher.AddRepo` → `subscribeRepo` walks the tree and calls `w.fs.Add(p)` on every surviving **directory** (`internal/daemon/watch/watcher.go:284-340`).

But `fsnotify v1.10.1` has **no FSEvents backend at all** — `backend_kqueue.go` carries `//go:build … || darwin`. This is a library fact, not a configuration choice. In kqueue:

- `addWatch` opens an FD per watched path (`unix.Open(name, openMode, 0)`).
- When the path is a directory it calls `watchDirectoryFiles(d)`, which `ReadDir`s it and opens **another FD per file in that directory**.

fsnotify says so itself, in `Close()`: *"On macOS a single directory watch opens an fd for every file in the dir … ran out of fds with EMFILE (#732)."*

So one `AddRepo` costs **(non-skipped dirs) + (files in those dirs)** descriptors.

## The measurement

Live engine (`grafel engine --foreground`, up 1d11h), **one** repo subscribed:

```
lsof -p <pid> | wc -l  →  40079
  32073 REG    (the watchDirectoryFiles cost)
   7995 DIR
      3 KQUEUE
```

39,894 of those belong to a single repo. Ceiling: `kern.maxfilesperproc: 61440`. The daemon plist requests `NumberOfFiles 65536`, which **exceeds `maxfilesperproc` and is silently clamped** — `internal/daemon/fdlimit/fdlimit.go` documents exactly this Darwin quirk.

Calibration on a mid-size repo (excluding `.git`/`node_modules`): 1,158 dirs + 9,034 files ≈ **10,200 FDs**. Extrapolated to 140 such repos: **~1.4 million FDs against 61,440 — a 23× overrun. EMFILE lands somewhere around repo 6.**

## The existing cap does not see this

`walk.DefaultWatchDirCap = 100000` (`internal/daemon/walk/protected.go:69`, enforced at `watcher.go:305`) counts **directories only** — it is blind to the per-file FDs, which are 4× larger here. At 100k directories it is effectively infinite.

[...]

Issue #6014, opened 2026-07-27
serve reports in_flight=0 always in split mode — the increment is on a branch that is never reached


cajasmota:
## Summary

The `in_flight` counter reported by `serve` is structurally always `0` in the default configuration, because the code that increments it sits on a branch split mode never reaches.

## Mechanism

`internal/daemon/service.go:685-712` acquires `groupRebuildMu` only on the fall-through path. Split mode — which is **default-on** since ADR-0024 — never reaches it; `internal/daemon/requests/requests_drain.go:70-77` acknowledges this explicitly and adds an engine-side guard instead.

So `atomic.AddInt64(&s.inFlight, 1)` at `service.go:684` is never exercised in the shipped configuration, and every consumer of the serve-side counter reads a constant zero.

## Why this matters

This is very likely what #5991's reporter actually saw. That issue is filed as "`grafel reset` can silently no-op, exits 0" — and one of the symptoms is the daemon appearing to be doing nothing. If `in_flight` is structurally zero, then "no rebuild is happening" is what status reports *regardless of the truth*, which would make a working rebuild indistinguishable from a dropped one.

It is worth separating the two: #5991 is about the CLI not requesting completion and the engine silently declining; this is about the status plane being unable to report the truth even when the rebuild is running fine.

## Suggested fix

Either move the counter to the path split mode actually takes, or have the serve-side status read the engine-side value (the engine already tracks this — `requests_drain.go` guards on it). Whichever way, add a test that asserts `in_flight` is non-zero while a rebuild is genuinely in flight **in split mode**, since the current tests presumably exercise only the monolith path.

## Provenance

Found during a full audit of the open issue backlog. Not covered by any open issue.

Refs #5991

PR #6925, opened 2026-09-06
fix(#6921): an fsnotify queue overflow dropped events forever and only logged — detect it, rescan every subscribed repo, and surface the staleness

### LINKED ISSUE #6921: fsnotify's `ErrEventOverflow` is handled nowhere — an inotify queue overflow logs one line and leaves the graph silently stale
fsnotify's `ErrEventOverflow` is handled nowhere — an inotify queue overflow logs one line and leaves the graph silently stale

Found while evaluating git polling as a watcher alternative (#6726 consumer context). **This is a silent-staleness path independent of watch exhaustion**, and it needs no unusual configuration to hit.

## The defect

`internal/daemon/watch/watcher.go:1367` is the only consumer of the fsnotify error channel:

```go
case err, ok := <-errs:
    if !ok {
        w.backendClosed()
        return
    }
    if err != nil && !w.stopping() {
        w.logger.Error("watcher: error", "err", err)
    }
```

It logs and continues. `ErrEventOverflow` appears **nowhere** in `internal/` or `cmd/` — verified by grep across both trees.

## Why that matters

inotify raises `IN_Q_OVERFLOW` when more events arrive than `fs.inotify.max_queued_events` (kernel default **16384**) can hold. fsnotify surfaces that as `ErrEventOverflow` on the error channel. **When it fires, an unknown set of events has been dropped and is never redelivered.**

fsnotify is edge-triggered: a dropped event is lost permanently. So after an overflow the watcher keeps running, keeps reporting healthy, and the graph is stale for every file whose event was dropped — until something unrelated (a branch switch, a manual reindex, a daemon restart) happens to re-cover it.

The triggering conditions are ordinary in the target environment: `npm install`, a codegen run, a large branch checkout, or a bulk file operation by an AI coding agent. This is not a pathological-tree scenario.

## Why it is worse in the container/agent setting

The consumer in #6726 runs agents that edit files inside per-task worktrees. Each tracked worktree subscribes its **entire** working tree, so a single lane can hold ~10k watch descriptors (measured: **976 directories per worktree** on this repo through the real `AddRepo`, and `GRAFEL_MAX_WORKTREES_PER_REPO` defaults to 10). More subscriptions on one inotify instance means a shared queue and a proportionally higher chance of overflow under a burst.

And the symptom is indistinguishable from the watch-exhaustion failure mode already noted in `fdbudget.go` — *"a repo that is registered but silently receives no events"*. Two different causes, one silent outcome, neither reported.

## What a fix has to do

1. **Detect it.** `errors.Is(err, fsnotify.ErrEventOverflow)` at `watcher.go:1367`.
2. **Recover.** An overflow means "your event history is incomplete", so the only correct response is a **full rescan of the affected repo** — not a retry, and not a log line. The existing bulk/coalesce path (`skip.go`'s debouncer) is the natural place to hand off to.
3. **Report it.** It should be visible in `grafel status` / the group doctor, not only in the daemon log. A staleness event the user cannot see is the actual defect here.
4. **Decide the scope.** fsnotify's overflow is per *watcher instance*, not per repo. Establish whether one instance backs all repos (in which case an overflow invalidates everything) or one per repo. **This is not established** and it determines how big the rescan must be.

## Not established

- **Whether an overflow has ever actually fired here.** No reproduction was attempted; this is a code-path finding, not an observed incident. Reproducing it (thousands of rapid creates under a watched tree) is the first thing a fix should do.
- Whether the Windows `ReadDirectoryChangesW` backend and macOS kqueue have equivalent overflow semantics — fsnotify's error surface differs per backend, and only the Linux/inotify case is described above.
- Whether `walk.WatchDirCap` or the quarantine tracker incidentally bound queue pressure in practice.
- Whether any existing test drives the error channel at all.

Refs #6726, #6852.

cajasmota:
`internal/daemon/watch/watcher.go`'s error arm was the only consumer of fsnotify's error channel, and it treated an `ErrEventOverflow` exactly like any other backend error: one `logger.Error`, then carry on. An overflow means an **unknown set of events was dropped**, and fsnotify is edge-triggered — nothing redelivers them. The watcher kept reporting healthy and the graph stayed silently stale for every file whose event was lost.

## STEP 1 — the scope question the issue left open

**One fsnotify instance backs every repo.** Not inferred from a type name:

- `internal/daemon/engineplane.go:203` — `watch.NewWatcherConfig(...)` is the **only** non-test construction of a `Watcher` in `internal/` or `cmd/`, and it runs once per daemon process.
- `internal/daemon/watch/watcher.go:180` — that one `Watcher` holds a single `fs *fsnotify.Watcher`, shared by every repo `AddRepo` subscribes.
- `internal/daemon/watch/manager.go:6-12` says the same thing from the other end: pause/resume "delegate to those APIs rather than introducing a second Watcher instance."

So an overflow invalidates **everything currently watched**, and the rescan has to be all of it. `IN_Q_OVERFLOW` also carries no path, so no narrower scope exists even in principle.

## STEP 1 — reproduction: NOT reproduced, and why

Attempted, honestly, and it failed for a structural reason rather than a lack of pressure: **20 000 file creates in 2.7 s across 50 watched directories → 4481 events delivered, 0 overflows.** The positive control is in that same line — events flowed, so the harness was live.

This machine is macOS and no Linux container was available. fsnotify documents the reason at `fsnotify.go:262-270`: **"kqueue, fen: Not used."** Only `backend_inotify.go:397-400` and `backend_windows.go:580-583` ever raise it. The branch is therefore unreachable on darwin *by construction*, and the Linux/Windows behaviour here is reasoned from fsnotify's source, not observed. That is a real limit on the confidence of everything downstream and it is stated in the commit message too.

## STEP 2 — the fix

1. **Detect** — `errors.Is(err, fsnotify.ErrEventOverflow)` at the error-channel site. Every other error keeps the old behaviour.
2. **Recover** — `ForceRescan()`, i.e. the existing `EventSink(repoPath, bulk=true)` contract, which is precisely the hand-off `restartBackend` already uses for the identical "we lost events" condition. No second reindex path was invented. The fan-out is dispatched to its own goroutine: the loop goroutine is the sole receiver on the backend's `Events`, and blocking it deadlocks the Windows backend's next `Add` (the invariant at `watcher.go:172-185`).
3. **Report** — `OverflowStats()` → `StatusReply.Watcher{Overflows,OverflowRescans,LastOverflow}` → an **ungated** line in `grafel status`:

   ```
   watcher: 3 fsnotify queue overflow(s) — file events were DROPPED and never
   redelivered; 1 full rescan(s) triggered to recover (last: 2026-09-06T12:00:00Z)
   ```

   Ungated for the same reason the `#6180` descriptor line is: the `watcher:` activity block is gated on counters describing events that *did* arrive, and an overflow is about events that did not. The same three fields land in `/diagnostics`.

**The storm trap.** A burst raises several overflows and the rescans they ask for are identical, so all but the first inside `overflowRescanCooldown` (30 s, measured on the package's clock seam) are **coalesced — still counted, still reported, acted on once**. Same shape of bound as #5675's activation semaphore and `quarantine.go`'s tracker.

## Tests

Every timing outcome is asserted against the injected `manualClock`; there is not a `time.Sleep` in the suite. The harness points the loop at channels the test owns and gives the real backend a discard sink, so it is safe on Windows (#6380's lesson).

| Test | Pins |
|---|---|
| `TestQueueOverflowRescansEverySubscribedRepo` | **both** repos rescanned, `bulk=true` — the scope claim above |
| `TestPlainBackendErrorNeverRescans` | the over-firing direction: a plain error must not reindex the daemon |
| `TestOverflowStormCoalescesIntoOneRescan` | 5 overflows → 1 rescan; one tick short of the cooldown → still 1; past it → 2 |
| `TestOverflowDuringStopDoesNotRescan` | no reindex armed against a watcher being torn down |
| `TestDaemonDiagnosticsReportsQueueOverflows6921` + omission case | the `/diagnostics` fill, at `buildDaemonDiagnostics` |
| `TestPrintDaemonDetail_OverflowLine…` + silence case | the `grafel status` print gate (#6014's defect class) |
| `TestFillWatcherStatus…` + `TestStatusCallsTheWatcherFill6921` | the reply mapping **and** the call site — a real `*watch.Watcher` proves `Status` still calls it |

## Mutant table — 12 applied, 12 DEAD

Each applied by a script that refuses to write unless the anchor matches exactly once and the replacement lands exactly once; `go vet` before every score; `-count=1` throughout; restored by `cp` from shasum-verified backups (never `git checkout`), with a `diff` proving the restore.

| # | Mutant (call site) | Verdict | Killed by |
|---|---|---|---|
| M1 | delete the overflow detection in `loop` | **DEAD** | scope + storm tests |
| M2 | keep detection, drop `go w.ForceRescan()` (**the fifth no-op mode**) | **DEAD** | scope + storm tests |
| M3 | delete the cooldown → a rescan per overflow | **DEAD** | storm test |
| M4 | rescan only one repo instead of all | **DEAD** | scope test |
| M5 | drop the `stopping()` guard | **DEAD** | shutdown test |
| M6 | widen the predicate to `if true` | **DEAD** | plain-error test |
| M8 | drop `d.WatcherOverflows` in diagnostics | **DEAD** | dashboard test |
| M9 | delete the `grafel status` line | **DEAD** | CLI print-gate test |
| M10 | drop `reply.WatcherOverflows` | **DEAD** | fill test |
| M11 | stop calling `fillWatcherStatus` from `Status` | **DEAD** | real-watcher control |
| M12d / M12s | remove the `!last.IsZero()` guard (year-1 timestamps) | **DEAD** | the two omission tests |

No survivors, so there is no three-way verdict to render.

## Verification

`go vet ./internal/...` clean; `gofmt -l` clean. `go test -count=1` green on `./internal/daemon/watch/`, `./internal/daemon/`, `./internal/cli/`, `./internal/dashboard/`.

One caveat worth recording: a first full run showed 4 `TestStaleReindexGuard_*` failures that were **my own scratch `GRAFEL_DAEMON_ROOT` being reused across runs** — leftover request files, not this change. Re-run against a fresh root: green.

Nothing in the change is `runtime.GOOS`-conditional, so no `ci:full` label is needed on that account.

Closes #6921

### DIFF
--- a/internal/cli/status_daemon_detail.go
+++ b/internal/cli/status_daemon_detail.go
@@ -39,6 +39,21 @@ func printDaemonDetail(w io.Writer, st proto.StatusReply) {
 		fmt.Fprintf(w, "  watcher: %d repo(s) NOT WATCHED — file-descriptor budget full (%d/%d used); edits there will not re-index\n",
 			st.WatcherUnwatched, st.WatcherFDUsed, st.WatcherFDLimit)
 	}
+	// #6921: an fsnotify queue overflow dropped file events that are never
+	// redelivered. Printed on its own line and NOT gated on the watcher line
+	// above, for the same reason the descriptor line is not: the condition is
+	// about events that did NOT arrive, so the counters that gate that line say
+	// nothing about whether this one applies. A rescan has already re-covered
+	// the dropped window; the line exists so a user who saw a stale answer can
+	// tell an overflow from a healthy watcher, which is otherwise impossible.
+	if st.WatcherOverflows > 0 {
+		fmt.Fprintf(w, "  watcher: %d fsnotify queue overflow(s) — file events were DROPPED and never redelivered; %d full rescan(s) triggered to recover",
+			st.WatcherOverflows, st.WatcherOverflowRescans)
+		if st.WatcherLastOverflow != "" {
+			fmt.Fprintf(w, " (last: %s)", st.WatcherLastOverflow)
+		}
+		fmt.Fprintln(w)
+	}
 	if st.QueueLen > 0 || len(st.IndexInFlight) > 0 ||
 		len(st.PendingAlgo) > 0 || len(st.PendingLinks) > 0 {
 		fmt.Fprintf(w, "  scheduler: queue=%d in_flight=%d pending_algo=%d pending_links=%d\n",
--- a/internal/cli/status_overflow_6921_test.go
+++ b/internal/cli/status_overflow_6921_test.go
@@ -1,0 +1,61 @@
+package cli
+
+// status_overflow_6921_test.go — #6921, the print gate for fsnotify queue
+// overflows.
+//
+// The failure being reported is silent staleness: an overflow drops an unknown
+// set of file events, fsnotify never redelivers them, and the watcher keeps
+// reporting healthy. `grafel status` is where a user looks when the graph does
+// not match the tree, so a number that reaches StatusReply and is never printed
+// is the same defect #6014 was — right on the wire, absent from the screen.
+
+import (
+	"bytes"
+	"strings"
+	"testing"
+
+	"github.com/cajasmota/grafel/internal/daemon/proto"
+)
+
+// TestPrintDaemonDetail_OverflowLinePrintsWithNoWatcherActivity is the
+// regression AND the gate assertion. WatcherRepos/WatcherEvents are zero here
+// on purpose: the overflow line must not be nested inside the `watcher:`
+// activity block, because that block is gated on counters describing events
+// that DID arrive, and an overflow is about events that did not.
+func TestPrintDaemonDetail_OverflowLinePrintsWithNoWatcherActivity(t *testing.T) {
+	var buf bytes.Buffer
+	printDaemonDetail(&buf, proto.StatusReply{
+		WatcherRepos:           0,
+		WatcherEvents:          0,
+		WatcherOverflows:       3,
+		WatcherOverflowRescans: 1,
+		WatcherLastOverflow:    "2026-09-06T12:00:00Z",
+	})
+	out := buf.String()
+
+	for _, want := range []string{
+		"3 fsnotify queue overflow(s)",
+		"DROPPED",
+		"1 full rescan(s)",
+		"2026-09-06T12:00:00Z",
+	} {
+		if !strings.Contains(out, want) {
+			t.Errorf("`grafel status` output is missing %q — an overflow the user cannot see is "+
+				"the actual defect in #6921.\ngot:\n%s", want, out)
+		}
+	}
+}
+
+// TestPrintDaemonDetail_NoOverflowLineWhenNoneHappened: the line is a warning,
+// so it must be silent in the overwhelmingly common case. A line that always
+// prints trains the reader to ignore it.
+func TestPrintDaemonDetail_NoOverflowLineWhenNoneHappened(t *testing.T) {
+	var buf bytes.Buffer
+	printDaemonDetail(&buf, proto.StatusReply{
+		WatcherRepos:  2,
+		WatcherEvents: 400,
+	})
+	if out := buf.String(); strings.Contains(out, "overflow") {
+		t.Errorf("overflow line printed for a watcher that never overflowed:\n%s", out)
+	}
+}
--- a/internal/daemon/proto/proto.go
+++ b/internal/daemon/proto/proto.go
@@ -88,15 +88,28 @@ type StatusReply struct {
 	// file-descriptor budget was full, and WatcherFDUsed/WatcherFDLimit are
 	// the ledger behind that decision (#6180). A non-zero WatcherUnwatched
 	// means edits in those repos are not being picked up at all.
-	WatcherUnwatched int                `json:"watcher_unwatched,omitempty"`
-	WatcherFDUsed    int                `json:"watcher_fd_used,omitempty"`
-	WatcherFDLimit   int                `json:"watcher_fd_limit,omitempty"`
-	QueueLen         int                `json:"queue_len,omitempty"`
-	IndexInFlight    []string           `json:"index_in_flight,omitempty"`
-	PendingAlgo      []string           `json:"pending_algo,omitempty"`
-	PendingLinks     []string           `json:"pending_links,omitempty"`
-	IndexedRepos     []IndexedRepoState `json:"indexed_repos,omitempty"`
-	RecentLog        []SchedLogEntry    `json:"recent_log,omitempty"`
+	WatcherUnwatched int `json:"watcher_unwatched,omitempty"`
+	WatcherFDUsed    int `json:"watcher_fd_used,omitempty"`
+	WatcherFDLimit   int `json:"watcher_fd_limit,omitempty"`
+
+	// WatcherOverflows counts fsnotify queue overflows (#6921): the backend
+	// dropped an unknown set of file events, and because fsnotify is
+	// edge-triggered nothing redelivers them. WatcherOverflowRescans is how
+	// many full reindexes of every subscribed repo those overflows triggered —
+	// the recovery — and WatcherLastOverflow (RFC3339) is when the most recent
+	// one arrived, so the user can correlate it with what they were doing.
+	//
+	// A non-zero count is the only way this failure is visible at all: the
+	// watcher keeps running and keeps reporting healthy through an overflow.
+	WatcherOverflows       uint64             `json:"watcher_overflows,omitempty"`
+	WatcherOverflowRescans uint64             `json:"watcher_overflow_rescans,omitempty"`
+	WatcherLastOverflow    string             `json:"watcher_last_overflow,omitempty"`
+	QueueLen               int                `json:"queue_len,omitempty"`
+	IndexInFlight          []string           `json:"index_in_flight,omitempty"`
+	PendingAlgo            []string           `json:"pending_algo,omitempty"`
+	PendingLinks           []string           `json:"pending_links,omitempty"`
+	IndexedRepos           []IndexedRepoState `json:"indexed_repos,omitempty"`
+	RecentLog              []SchedLogEntry    `json:"recent_log,omitempty"`
 
 	// GroupAlgoRunning names the groups whose ANNOTATION (group-algo) pass is
 	// executing right now, and GroupAlgoInFlight counts them. The pass produces
--- a/internal/daemon/service.go
+++ b/internal/daemon/service.go
@@ -396,17 +396,7 @@ func (s *Service) Status(_ *proto.StatusArgs, reply *proto.StatusReply) error {
 	reply.DaemonMode = s.daemonMode
 
 	if s.watcher != nil {
-		repos, dirs, events, dropped, unwatched := s.watcher.Stats()
-		reply.WatcherRepos = repos
-		reply.WatcherDirs = dirs
-		reply.WatcherEvents = events
-		reply.WatcherDropped = dropped
-		// #6180: surface repos that are NOT watched for want of file
-		// descriptors, plus the ledger, so `grafel status` says so out loud.
-		reply.WatcherUnwatched = unwatched
-		fdUsed, fdLimit, _, _ := s.watcher.FDBudgetStats()
-		reply.WatcherFDUsed = fdUsed
-		reply.WatcherFDLimit = fdLimit
+		fillWatcherStatus(reply, s.watcher)
 	}
 	// PH2a (#2096): report pause/resume slot counts.
 	if s.watcherMgrStats != nil {
--- a/internal/daemon/service_watcher_status.go
+++ b/internal/daemon/service_watcher_status.go
@@ -1,0 +1,59 @@
+package daemon
+
+// service_watcher_status.go — the watcher half of Service.Status' reply.
+//
+// Extracted from Status (#6921) so the mapping can be driven by a stand-in
+// watcher. Status itself needs a live *watch.Watcher to have counters worth
+// reading, and the one condition that matters most here — a queue overflow —
+// cannot be provoked from outside package watch on any platform, and cannot be
+// provoked at all on macOS, where the kqueue backend never raises it.
+//
+// The extraction moves the mapping out of Status but NOT the decision to call
+// it: Status still owns the `s.watcher != nil` gate, and a test that builds a
+// Service around a real Watcher pins that gate independently — otherwise this
+// would be the classic extraction that pins a helper and leaves the call site
+// free to skip it.
+
+import (
+	"time"
+
+	"github.com/cajasmota/grafel/internal/daemon/proto"
+)
+
+// watcherStatusSource is the read-only subset of *watch.Watcher that
+// Service.Status reports. Narrow by construction: everything here is a counter
+// snapshot, so a stand-in is a struct of fields rather than a simulation.
+type watcherStatusSource interface {
+	// Stats returns (repos, dirs, totalEvents, dropped, unwatched).
+	Stats() (int, int, uint64, uint64, int)
+	// FDBudgetStats returns (used, limit, unwatched, unwatchedRepos) (#6180).
+	FDBudgetStats() (int, int, int, []string)
+	// OverflowStats returns (overflows, rescans, coalesced, last) (#6921).
+	OverflowStats() (uint64, uint64, uint64, time.Time)
+}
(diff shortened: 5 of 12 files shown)


cajasmota:
## Coordinator mutant — DEAD, and the run answers the PR's own open question

The 12 author mutants attack **detection** (drop it, widen it, drop the `stopping()` guard), **recovery** (no-op it, drop the cooldown, rescan one repo instead of all) and **plumbing** (drop the status line, the reply fill, the diagnostics fill). None attacks the **fidelity of the ledger itself** — whether the numbers reported are the right numbers, as opposed to being reported at all.

That matters because #6921's stated defect is not "the code does nothing", it is **"a staleness event the user cannot see"**. A ledger that under-reports is the same defect wearing the fix's clothes.

**Mutant:** count only the overflows that trigger a rescan — move `atomic.AddUint64(&w.overflows, 1)` below the cooldown early-return. Recovery behaviour is **completely unchanged**: same detection, same single rescan, same cooldown, same coalescing. The only difference is that a storm of N overflows reports as **1**.

Gated both directions (top-add removed → 0 matches; single add present → 1 match), `go vet` clean.

**Verdict: DEAD** — `TestOverflowStormCoalescesIntoOneRescan`.

So the ledger is graded on *count*, not merely on presence. Good — that is the half that makes the fix worth having.

### The more useful finding: the overflow path IS driven, on any platform

The failing run's log shows the test feeding the error channel directly:

```
level=ERROR msg="watcher: error" err="sentinel: not an overflow"
level=WARN  msg="watcher: fsnotify queue overflow — ... forcing a full reindex of every subscribed repo" repos=1 overflows=1
level=INFO  msg="watcher: force-rescan" repo=...
level=WARN  msg="... a full rescan is already in flight for this burst, coalescing" overflows=1
```

Both a **positive** (a real `ErrEventOverflow` reaching `handleOverflow`) and a **negative** control (`sentinel: not an overflow` taking the plain-log path) go through the real error-channel site at `watcher.go:1397`, not by calling `handleOverflow` directly.

That substantially narrows the "could not reproduce" caveat, and the PR body undersells itself. There are two separable claims:

1. **Does grafel handle the error correctly once it arrives?** — **Driven and graded on every platform, including this one.** Not a gap.
2. **Does the inotify backend actually raise it, and does it survive un-wrapped to `errors.Is`?** — Still unobserved, and only reasoned from fsnotify's source.

Only (2) is unverified. Worth saying that way in the body, because "not reproduced" currently reads as though the whole fix is untested, and it is not.

Restored by `cp` from a shasum-verified backup (`5ae90626a…`); tree confirmed pristine; worktree removed.

Refs #6921.

cajasmota:

Reviewed at `70e274b3d` in a private detached worktree, fresh `GRAFEL_HOME` + `GRAFEL_DAEMON_ROOT`. Every mutant applied by a script that aborts unless the anchor matches **exactly once** and the replacement lands **exactly once**, `go vet` before each score, `-count=1` throughout, restored by `cp` from a sha256-verified backup.

### 1. Would this actually fire on Linux? — YES. Verified from fsnotify's source, not the quoted line numbers.

This was the highest-risk claim in the PR (the author correctly flags that they could not observe it on darwin), so I traced the sentinel end to end in `$(go env GOMODCACHE)/github.com/fsnotify/fsnotify@v1.10.1`:

- `backend_inotify.go:398` — `if inEvent.Mask&unix.IN_Q_OVERFLOW != 0 { w.sendError(ErrEventOverflow) }`.
- `backend_windows.go:582` — `w.sendError(ErrEventOverflow)` on a zero-length read.
- `shared.go:34-45` (`*inotify` embeds `*shared`) — `case w.Errors <- err`. **The sentinel is put on the channel raw.** No `fmt.Errorf`, no `%w`, no wrapper type on either backend.
- `fsnotify.go:285` — the public facade is `&Watcher{b: b, Events: ev, Errors: errs}`; the same `errs` channel, not a re-forwarded one.
- grafel binds it directly: `watcher.go:527` and `:1276`, `w.events, w.errs = fw.Events, fw.Errors`.

So the value reaching `errors.Is(err, fsnotify.ErrEventOverflow)` is the sentinel **by identity**. `errors.Is` matches on the first comparison. There is no wrapping anywhere on the path, and therefore no way for the fix to be inert on the platform it targets.

Grepping `Overflow` across `backend_kqueue.go`, `backend_fen.go`, `backend_other.go` returns **zero hits** — the "unreachable on darwin by construction" claim is confirmed, and the non-reproduction is a structural fact rather than insufficient pressure. Two nits: the cited `397-400` is actually `398-401` and `580-583` is `582`; immaterial, but the comments in `overflow.go` will drift.

### 2. Can the overflow path be driven without a Linux box? — YES, and the tests already do it through the real code path.

`overflow_6921_test.go` feeds `fsnotify.ErrEventOverflow` onto the loop's error channel via the **pre-existing** `cfg.testErrors` seam (`watcher.go:529`), so the sentinel goes through the real `loop()` select and the real `errors.Is` branch — it does **not** call `handleOverflow` directly. That is proven, not asserted: **M1 mutates the `errors.Is` line in `watcher.go` itself** (not `overflow.go`) and the overflow tests fail. The graded site is the production branch.

The harness also keeps the real backend live with a discard sink so `AddRepo` subscribes a real watch set — "every subscribed repo" means something, and Windows `Add` cannot deadlock (#6380's lesson, applied).

Combined with §1, the only thing no test on any platform observes is the *kernel* raising `IN_Q_OVERFLOW`. Nothing in CI could observe that anyway without `sysctl fs.inotify.max_queued_events` tuning inside the runner.

### 3. `ci:full` — agreed, not needed, but state the limit precisely.

Nothing is `runtime.GOOS`-conditional and the tests are not platform-gated, so the **Linux leg does exercise the new branch** — it drives grafel's handling of the sentinel identically to macOS. What no leg exercises is a real kernel overflow. The honest framing is not "the Linux leg gives real coverage that macOS doesn't"; it is "all three legs give the same synthetic coverage, and that is the most that is obtainable here." On that basis `ci:full` buys nothing and I agree with omitting it.

### 4. Scope claim — verified with **one correction**.

- `internal/daemon/engineplane.go:203` is indeed the **only** non-test `watch.NewWatcherConfig` in `internal/` + `cmd/` — confirmed.
- `watcher.go:180` — one `fs *fsnotify.Watcher` per `Watcher`, shared by every `AddRepo` — confirmed. `restartBackend` (`watcher.go:1252`) replaces that instance in place and re-binds `w.events/w.errs`; the overflow counters and cooldown live on `w`, so they survive a backend swap correctly.
- `ForceRescan` (`watcher.go:1068`) really is `w.sink(p, true)` over every entry of `w.repos` — the same `EventSink(repo, bulk=true)` contract, no second reindex path. Confirmed.

**The correction:** there *is* a second process-wide `fsnotify.NewWatcher()` — `internal/daemon/worktree/worktree.go:674`, watching each parent's `.git/worktrees/`. Its error channel is drained and **discarded** (`case _, ok := <-fw.Errors:`), so an overflow on *that* inotify fd is still unhandled. So `overflow.go`'s "grafel builds exactly one Watcher per daemon process" is true of `watch.Watcher` but **not** of `*fsnotify.Watcher` instances in the process, and inotify queues are per-fd.

This does **not** change the rescan scope and is not a blocker: that watcher holds ~1 descriptor per parent, and `Watcher.Start`'s 60s reconciliation poll re-runs `Sync()` unconditionally, so a dropped event there is self-healing within a tick — structurally the opposite of the defect being fixed. But the SCOPE paragraph as written would mislead the next reader. Worth one sentence: *"a second fsnotify instance exists in internal/daemon/worktree for `.git/worktrees`; its overflows are covered by that watcher's reconciliation poll, not by this path."*

### 5. Deadlock hazard and the goroutine — the goroutine is the right fix.

`watcher.go:172-185` documents a real invariant, not a style rule: fsnotify's Windows backend serialises `Add`/`Remove` through the same I/O goroutine that produces `Events`, so blocking the sole receiver wedges the next `Add`. `handleOverflow` runs on the loop goroutine and `w.sink` is the daemon's reindex scheduler — an unbounded-latency callback. Dispatching it is correct and matches `RescanRepo`. It is not papering over lock ordering: `handleOverflow` releases `w.mu` before `go w.ForceRescan()`, and `ForceRescan` takes `w.mu` itself, so there is no nesting to invert.

Goroutine-per-overflow cannot storm: the cooldown gates the dispatch, not the goroutine. One residual, acceptable and worth a comment: the cooldown is stamped at **dispatch** time, not completion, so a rescan that outlives 30s allows a second concurrent `ForceRescan`. `handleOverflow` also increments `overflowRescans` when `w.repos` is empty, reporting a rescan that rescanned nothing.

### 6. Mutant scorecard — 10 of 12 reproduced independently, **10/10 DEAD**

| # | Mutation (call site) | Verdict | Killed by |
|---|---|---|---|
| M1 | `errors.Is(...)` → `false && errors.Is(...)` in **`watcher.go`'s loop** | **DEAD** | `TestQueueOverflowRescansEverySubscribedRepo`, `TestOverflowStormCoalescesIntoOneRescan` |
| **M2** | keep detection, `go w.ForceRescan()` → `_ = 0` (**the fifth no-op mode**) | **DEAD** | scope + storm tests |
| M3 | cooldown predicate → `if false` | **DEAD** | `TestOverflowStormCoalescesIntoOneRescan` |
| M4 | rescan only the first repo (real replacement code, not a stub) | **DEAD** | `TestQueueOverflowRescansEverySubscribedRepo` |
| M5 | drop the `w.stopping()` guard | **DEAD** | `TestOverflowDuringStopDoesNotRescan` |
| M6 | widen to always-true (`errors.Is(err, err) \|\| ...`) | **DEAD** | `TestPlainBackendErrorNeverRescans` (+2) |
| M8 | drop `d.WatcherOverflows` in `buildDaemonDiagnostics` | **DEAD** | `TestDaemonDiagnosticsReportsQueueOverflows6921` |
| M9 | disable the `grafel status` line | **DEAD** | `TestPrintDaemonDetail_OverflowLinePrintsWithNoWatcherActivity` |
| M10 | drop `reply.WatcherOverflows` | **DEAD** | `TestFillWatcherStatusReportsQueueOverflows6921` |
| **M11** | `fillWatcherStatus(reply, s.watcher)` → `_ = reply` | **DEAD** | `TestStatusCallsTheWatcherFill6921` |
| M12 | remove the `!lastOverflow.IsZero()` guard | **DEAD** | `TestFillWatcherStatusOmitsOverflowTimeWhenNoneHappened6921` |

M2 and M11 were the two I most expected to survive, and both die on their named test. M11's control does build a **real `*watch.Watcher`** and a real `Service`, so the extraction into `service_watcher_status.go` is not a helper nobody calls — the call site is graded one level up, which is the right shape.

Not re-scored (accepted on the author's report): M12s, and the second half of the M12 pair.

### 7. Tree hygiene after the M6 incident — **pristine**, verified.

`git status --porcelain` empty and `git diff 70e274b3d --stat` empty at `70e274b3d`. Nothing from the reported non-unique-replacement incident survives, and every mutant I applied restored with a byte-for-byte equality assertion.

### 8. Two pre-existing things found, neither this PR's fault

- **`TestStaleReindexGuard_*` is not isolated from a reused `GRAFEL_DAEMON_ROOT` — confirmed independently and deterministically.** Fresh root: green. Same root reused after one prior run: `durable pending reindex requests after one heartbeat = 428, want 2`. The author's caveat is correct as diagnosis, but the underlying defect is real: those tests count durable request files in the daemon root without scoping to the run. Deserves its own issue.
- `TestInterleavedDirectoryFillsAcrossReposAreNotDoubleCharged` (~96s) failed in 2 of 8 mutant runs where the applied mutation could not plausibly reach it (M1, M5), and passed in baseline. Load-sensitive flake, pre-existing.

### Verification run here

`go vet ./internal/daemon/... ./internal/cli/... ./internal/dashboard/...` clean. `go test -count=1` green on `./internal/daemon/watch/` (74.6s), `./internal/daemon/` (33.6s), `./internal/cli/` (59.8s), `./internal/dashboard/` (36.5s). Not run: everything else, including `./cmd/grafel` — no extractor or graph surface is touched by this change.

### Verdict: **APPROVE**

The fix is correct, the branch is genuinely reachable on the platforms it targets and provably dead on darwin, the recovery reuses the existing bulk-reindex contract, the storm bound is real and asserted on the clock seam, and the mutant grading holds up under independent reproduction at the call sites. The two follow-ups — the `internal/daemon/worktree` second-instance caveat in `overflow.go`'s SCOPE paragraph, and the dispatch-time-vs-completion-time cooldown note — are comment-only and do not block.

Issue #6898, opened 2026-09-06
The quality ratchet gates extracted totals in the growth direction only — a fixture that LOSES members passes in silence; MEASURED: 8 fixtures across 4 languages sit below their ceilings (15 fields, all drops), and only the 3 python ones have an explanation


cajasmota:
The quality ratchet gates `*_extracted_total` in the growth direction only, so a fixture that loses entities or relationships passes in silence — and three python fixtures are sitting 4/3/4 below their ceilings right now

Two halves: a **mechanism** that cannot see a loss, and a **live instance** of that loss which is expected but unrecorded. Found while diagnosing PR #6897's ratchet failure (#6852 clojure arm).

## The mechanism

`entity_extracted_total` and `relationship_extracted_total` in `internal/quality/golden/baseline.json` are **ceilings**. The ratchet fails when extraction emits *more* than the recorded number — that is #6488 arm D, added because unnamed over-emission was invisible on all 31 fixtures.

It does not fail when extraction emits **fewer**. `scripts/quality/run.sh --ratchet` reports OK, and the code's own comment acknowledges a fall is "deliberately allowed and silent" — while also saying a fall "weakens this gate until the next re-record", which is the part nobody acts on.

So a gated fixture can lose four relationships between two re-records and **nothing says so**. That is the exact defect shape #6488 arm D exists to prevent, in the opposite direction.

## The live instance — expected, but never recorded

The clojure lane ran `--update-baseline` and found it would move **four** fixtures on current `main`, not one:

| fixture | baseline | measured on `9706a5cde` | |
|---|---|---|---|
| clojure-protocols-mini | 20 / 47 | 22 / 49 | grew (that arm; being handled in #6897) |
| python-django-mini | – / 255 | – / 251 | **shrank** |
| python-fastapi-mini | 28 / 57 | 27 / 54 | **shrank** |
| python-rq-mini | 22 / 43 | 20 / 39 | **shrank** |

Reproduced identically across two independent runs, so not noise. `baseline.json` records `measured_on: dab49ec0e`.

[...]

Issue #6923, opened 2026-09-06
The #6847 runtime guard detects "a language disappears" but not "a language erodes" — nothing pins `res.anchoring[lang]` to a number


cajasmota:
The #6847 runtime guard detects "a language disappears entirely" but not "a language erodes partially" — measured, with the boundary located exactly

Found by a coordinator mutant on PR #6920, the **final** arm of #6852. Non-blocking there — the arm is correct and its restatement closed the hole it was written for — but worth its own issue now that the ledger is empty and this guard is the only thing standing.

## The defect

`internal/extractors/file_anchored_carrier_runtime_6847_test.go` builds four per-language counters — `exercised`, `anchoring`, `carried`, `detected` — and then asserts:

- `diffStringSets(anchoringLanguages6847, sortedStringKeys(res.carried))` — a **set** comparison: presence, not count.
- `res.carried[lang] != res.anchoring[lang]` — relative.
- `res.detected[lang] != res.anchoring[lang]` — relative.

**Nothing pins `res.anchoring[lang]` to a number.** So if a language's corpus contribution shrinks, all four counters shrink together, every relative comparison still balances, and the language remains present in the set.

## Measured, with the boundary located

A mutant made one language contribute only its first anchoring file:

| erosion | verdict |
|---|---|
| zig 3 → **1** anchoring file | **ALIVE — suite green** |
| zig 3 → **0** anchoring files | **DEAD** — `EveryAnchoringFileCarries_6847` and `CorpusCoverage_6847` both fail with `missing (expected, not observed): [zig]` |

The second row is the non-vacuity proof: the mechanism *does* fire, so the first row is a real gap and not an unapplied edit.

## Why it matters now specifically

Arm 12 took `knownMissingCarrier6847` to **empty**. From here, this guard is what stands between the twelve fixed languages and a silent regression. If a future change causes a language to stop anchoring in some of its corpus files — a regex narrowed, a fixture moved, an extractor arm rewritten — **the guard reports green while measuring a fraction of what it used to.**

That is the "vacuous walk reports clean" shape one level up: not a walk that does nothing, but a walk whose *population* shrinks underneath it while every assertion stays satisfied.

There is already evidence this class of drift is easy to miss: the same PR's review found the author's un-name-vs-delete roster **one language short** — `markdown` was affected (17→13 detected) on top of the five named. A per-language count pin would have surfaced that as a failing number instead of requiring a reviewer's second look.

## The fix

[...]

Issue #6260, opened 2026-08-08
Every internal/custom extractor is switched off during the golden benchmark: 22 of 26 missing must-have entities are one default-off gate, not a missing capability


cajasmota:
Measured while fixing #6231, confirmed independently in its review. **This one gap accounts for 22 of the 26 missing must-have entities across the whole golden corpus** — five red fixtures, one capability.

## The fixtures

| Fixture | Recall | Missing |
|---|---|---|
| `csharp-hangfire-mini` | 0/5 | `EmailJob`, `CleanupJob` as `SCOPE.Service`; `EmailService.SendConfirmation` as `SCOPE.Operation` |
| `csharp-quartz-net-mini` | 0/5 | same shape |
| `java-quartz-mini` | 2/7 | same shape |
| `python-dramatiq-mini` | 0/4 | same shape |
| `python-rq-mini` | 0/3 | `send_email.enqueue` / `enqueue_call` as `SCOPE.Operation`; `Worker(...)` as `SCOPE.Service` |

Reproduce: `scripts/quality/run.sh --runs 1` (strict mode exits 2; the per-fixture JSON reports name every miss).

## The shape

Two things are consistently unextracted:

1. **Job/worker classes are not emitted as `SCOPE.Service`.** A Hangfire `EmailJob`, a Quartz `IJob` implementation, a dramatiq actor, an RQ `Worker` — each is a service in the architectural sense, and none is recognised as one.
2. **Enqueue and schedule sites are not emitted as `SCOPE.Operation`.** `BackgroundJob.Enqueue(...)`, `scheduler.ScheduleJob(...)`, `send_email.send(...)`, `queue.enqueue(...)`, `enqueue_call(...)` — the call that hands work to the queue is the edge between the producer and the consumer, and it is invisible.

Together those are what make a background-job architecture legible: without them a graph of one of these codebases shows the HTTP surface and then simply stops at the queue boundary.

## Why this is the highest-leverage recall work

The other four causes behind #6231's red fixtures are one or two fixtures each. This is five fixtures and **22 of 26 misses**. It is also a single coherent capability rather than five independent framework fixes — the frameworks differ in spelling, not in shape, and the same producer/consumer/queue triple appears in all of them.

It is worth checking before implementing whether an existing extractor already models this for a framework that *is* green — if some queue-like capability exists and simply has no rule for these five, that is a much smaller change than building it.

## Guardrails

[...]

PR #6926, opened 2026-09-06
fix(#6917): sinatra's `^`-anchored patterns compiled without `(?m)` matched start-of-TEXT — a modular app yielded ZERO Route entities; the sweep found 27 such patterns in 8 files and one that would over-fire

### LINKED ISSUE #6917: sinatra's `^`-anchored rule patterns compile without `(?m)`, so `^` means start-of-TEXT — a realistic sinatra app yields ZERO Route entities
sinatra's `^`-anchored rule patterns are compiled without `(?m)`, so `^` means start-of-TEXT — a realistic sinatra app yields ZERO Route entities

Found while grounding #6914's sinatra sibling. **This is not a middleware defect and it is much larger than the one it was found under** — it silently zeroes classic-style sinatra route extraction.

## The defect

`internal/engine/detector.go:156` and `:177` compile rule patterns with plain `regexp.Compile`. Without the `(?m)` flag, Go's `^` matches **start of text**, not start of line.

Every `^\s*`-anchored pattern in `sinatra.yaml` therefore fires only if its construct is the **first thing in the file**.

Measured by driving the real `LoadAllRules()` + `Detector.Detect()` path:

- `use Rack::Session::Cookie` at byte 0 of `config.ru` → **fires**.
- The same line placed after a `require './app'` → **does not fire**.

This affects `before`, `after`, `error`, `helpers`, `configure` — **and the classic-style Route pattern** `^\s*(get|post|put|delete|patch)\s+["']path["']\s+do`.

## The consequence

On a realistic sinatra app — routes indented inside `class MyApp < Sinatra::Base`, which is the idiomatic modular style — the probe produced **zero `Route` entities and zero `Middleware` entities**.

So sinatra route extraction is not degraded, it is **absent** for any file that does not begin with a route. A single-file toy app whose first line is `get '/' do` would extract exactly one route and nothing after it.

## Why nothing caught it

`detector.go:1141-1143` carries a comment naming this exact hazard — but scoped to **terminator-bearing** rules. Sinatra's patterns are not terminator-bearing, so they were never audited against it.

There is **no sinatra fixture** under `internal/quality/golden/`, so the always-on extraction-quality gate has nothing to say about it. Same blindness class as #6906, #6912 and #6914: a defect in a language the gate does not cover cannot be caught or graded.

## What a fix has to establish FIRST — this is the sizing question

**Whether the omission extends beyond `sinatra.yaml`.** Only sinatra was measured. The rules tree holds ~200 files and **was not swept** for `^`-anchored patterns compiled without `(?m)`.

That number decides everything about this issue:

- If sinatra is the only offender, the fix is `(?m)` on ~8 patterns plus a fixture — small.
- If the omission is systemic, then adding `(?m)` at the **compile site** (`detector.go:156,177`) fixes every language at once — but that is a behaviour change across every rule in the repo, and patterns written *around* the current semantics (deliberately relying on start-of-text to match only a file's opening construct) would start over-firing.

**Do the sweep before choosing.** A per-file `(?m)` is safe and narrow; a compile-site `(?m)` is a one-line fix with corpus-wide blast radius. The sweep is what tells you which you are looking at, and it is cheap: parse the rule YAML structurally (`ScanRuleYAML` in `internal/entkinds` already walks all 732 rule files) rather than grepping.

## Recall, not over-firing, is the risk direction here

Note which way this fails: patterns currently match **too little**. So a fix moves recall **up**, and the extraction-quality ratchet gates entity/relationship totals **growth-only** — meaning a correct fix will trip the ceiling and require a deliberate `--update-baseline`, not a silent pass. That is the desired behaviour, but budget for it.

And per #6902's lesson, a recall assertion alone cannot detect over-firing. If the compile-site fix is chosen, it needs **forbidden rows** proving the newly-multiline patterns do not start matching things they should not — the recall direction will look green either way.

## Not established

- **Whether any other rule file has the same omission.** The single most important unknown; sinatra only.
- Whether any existing pattern *deliberately* relies on start-of-text semantics. If one does, a compile-site fix breaks it silently.
- The real-world split between classic-style and modular-style sinatra apps — the defect is proven for both (neither begins with a route), but no corpus frequency was measured.
- Whether the same construct in other Ruby frameworks (rails, hanami, roda) is affected — only sinatra's rules were read.

Refs #6914, #6726, #6912, #6902.

cajasmota:
Closes #6917

`internal/engine/detector.go:156,177` compile rule patterns with plain `regexp.Compile`. In Go, `^` without `(?m)` matches **start of text**, not start of line — so every `^\s*`-anchored pattern in `sinatra.yaml` could only fire when its construct was the first thing in the file.

Driven, not argued: `internal/engine/sinatra_multiline_anchor_6917_test.go` runs the real `LoadAllRules()` + `Detector.Detect()` over a modular-style app. Before the fix it reported `Routes found: []` — zero Route, Middleware, Config and Service entities. The new golden fixture graded **38.5% entity recall / 33.3% relationship recall** on a pre-fix binary and **100% / 100%** after.

## STEP 1 — the sweep, and the scope it drove

Swept structurally, not by grep: every rule YAML under `internal/engine/rules/<lang>/{frameworks,orms,queues}/` decoded through `engine.FrameworkRule` — the same schema `detector.go` compiles — so the sweep population **is** the compiled population. 520 rule files, 412 patterns.

**27 `^`-anchored patterns without `(?m)`, in 8 files across 7 languages:**

| file | count |
|---|---|
| `ruby/frameworks/sinatra.yaml` | 9 |
| `cicd/frameworks/github_actions.yaml` | 6 |
| `docker/frameworks/docker_compose.yaml` | 3 |
| `ansible/frameworks/ansible_core.yaml` | 2 |
| `python/frameworks/pytest.yaml` | 2 |
| `python/frameworks/sqlalchemy.yaml` | 2 |
| `scala/frameworks/play_framework.yaml` | 2 |
| `graphql/frameworks/graphql_schema.yaml` | 1 |

So the omission **is systemic**. It does not follow that the compile-site fix is the right one, and the sweep is also what says why:

1. **`(?m)` changes `$` too**, and the sweep found 4 `$`-bearing patterns without `(?m)` (in `docker_compose`, `ansible_core` ×2, `csharp/net_maui`). Two of them are effectively *dead today* and would become *live and very broad*: `docker_compose`'s `^\s{2}(\w[\w-]*):\s*$` currently needs start-of-text AND end-of-text; under `(?m)` it matches **every two-space-indented key in any YAML file the docker bucket sees**. `ansible_core`'s `\s+(\w+\.\w+\.\w+):\s*$` is the same shape. That is precisely the "pattern written around the current semantics starts over-firing" hazard the issue names — found, not hypothesised.
2. **6 of the 8 files have no golden fixture** (only `graphql-schema-mini` and `scala-play-mini` exist), so a compile-site change would widen 7 languages with nothing grading the widening. A green suite would not be a demonstration.

The review re-derived both figures independently (`candidates == loaded == 520`, all 412 compile, same 27/4 split) and found the `$` hazard **understated**: `dormantBucketAliases` maps `docker → {dockerfile, yaml}` and that pattern carries no `requires_framework`, so under a compile-site `(?m)` it is ungated on *every* `.yaml` in a repo — minting `labels`/`selector` from a k8s Deployment and `push`/`build`/`lint` from a GitHub workflow, measured.

**Decision: fix sinatra per-pattern — and be precise about what that rests on.** The `$` hazard rejects the **compile-site** fix specifically, and nothing more: 16 of the remaining 18 caret patterns carry no `$` at all, and the ansible FQCN pattern has no `^`, so a per-pattern fix would never touch the dangerous ones. What actually holds the line for deferring the other 7 files is the second reason — **6 of the 8 have no golden fixture** (only `graphql-schema-mini` and `scala-play-mini` exist), and the review confirmed the adjacent claim that could have undercut it: the existing `python-*` fixtures contain no *indented* `class X(Y)` or `def test_` that would incidentally grade `pytest.yaml` / `sqlalchemy.yaml`. Those five languages are genuinely ungraded, so widening them here would be a behaviour change nothing measures.

So the honest framing is **"the remaining 16 are a per-pattern fix waiting on fixtures"**, not "the sweep says stop". **A P1 is fixed in 1 of 7 affected languages by this PR, and #6927 must land** — it carries the measured, per-file plan (recall-first: `pytest`/`sqlalchemy`, then the two already-fixtured files, then `github_actions`, and the two `$` patterns last, since those need rewriting rather than flagging). Closing #6917 on this PR should not read as the matter being settled.

## STEP 2 — what changed

- `(?m)` added to sinatra's 9 `^`-anchored patterns (7 source patterns + 2 relationship rules). The two un-anchored patterns (`class X < Sinatra::Base`, `Sinatra::Application.run!`) are untouched — they already worked, and they serve as the control in the tests.
- The `helpers <Module>` include capture tightened from `(\w+)` to `([A-Z]\w*(?:::\w+)*)`. `\w+` matched the `do` of `helpers do` and minted a junk `Service` entity literally **named "do"** beside the real block entity. It was unreachable while the anchor defect kept both patterns dead and reachable the instant `(?m)` landed, so it is this change's business. Ruby module names are constants, so requiring one is the language's own rule.

## Fixture: `ruby-sinatra-modular-mini`

The first Ruby golden fixture — which is exactly why the always-on gate had nothing to say about a defect that zeroed a language.

**Varied axis (one):** the **vertical position** of each Sinatra construct. Nothing in any file is on line 1; every DSL call is indented inside `class MyApp < Sinatra::Base`, and `error 404 do` sits near the end of the file — the furthest possible point from start-of-text.

**Held constant:** the constructs' spellings (idiomatic sinatra throughout); the un-anchored `class MyApp < Sinatra::Base` declaration, which was green before and after and is the control saying only anchor-gated behaviour moved; and the four `http_endpoint_definition` entities, which come from the separate **Ruby extractor** rather than from `sinatra.yaml` and so distinguish "the YAML rules came back" from "something moved in Ruby extraction".

**Forbidden rows (3), each killed by a distinct mutant** — per #6902, the recall arm alone cannot see over-firing, and `(?m)` is precisely the change that lets `^\s*` reach an indented line for the first time:
- `Service "do"` — killed by widening the helpers capture back to `\w+`.
- `Service "Whatever"` — `server.helpers Whatever` in `src/receiver_calls.rb`. The capture would happily accept it, so **only** the line anchor keeps it out; killed by dropping the `^`.
- `Route "/never-registered"` — appears only inside a `#` comment; killed by dropping the `^`.

A vacuous fourth row (`Service "server"`, an entity no pattern could ever mint) was written and then removed rather than left in as decoration.

## Ratchet

`scripts/quality/run.sh --ratchet --runs 1` → **exit 0, 32 gated fixtures held their baseline (23 at 100% must-have recall)**.

`baseline.json` gains exactly one entry — the new fixture, at `entity_found 13/13`, `relationship_found 3/3`, `entity_extracted_total 61`, `relationship_extracted_total 118`. **No existing fixture's figures changed.**

One honest note on that: the `--update-baseline` run also wanted to re-record *lower* `extracted_total` figures across three Python fixtures — **five figures, not three**: `python-django-mini` rel 255→251, `python-fastapi-mini` ent 28→27 **and** rel 57→54, `python-rq-mini` ent 22→20 **and** rel 43→39. That drift is **not from this change** — a binary with `sinatra.yaml` reverted produces byte-identical lower numbers (reproduced independently in review, deterministic across repeat runs), and this change touches no Python rule. **It is already filed as #6898**, which records exactly this ("three python fixtures sit 4/3/4 below their ceilings now"); nothing new needs filing here. Since the totals are gated **growth-only** (`scripts/quality/ratchet.py`, `EXTRACTED_TOTAL_KEYS`), re-recording them would have silently *tightened* the ceiling on three fixtures over pre-existing drift that is not mine. Those three entries were restored to their prior values instead, so the baseline diff is one addition plus the `measured_at`/`measured_on` stamps.

Delta on the new fixture, enumerated rather than summarised (44→52 entities, 88→98 relationships before the `helpers` tightening was added; 61/118 in final form with the extra fixture files): the 8 entities the fix restores are `Route /invoices`, `Route /invoices/:id`, `Config production`, `Middleware 404`, `Middleware "after do"`, `Middleware "before '/admin/*' do"`, `Service "helpers do"` and (post-tightening) `Service AuthHelpers` — plus the 2 `Route --[CALLS]--> json_helper` edges and their `CONTAINS` companions. **Nothing was lost.**

## Mutant table

Scored at the call site — the shipped `sinatra.yaml` — with `go vet` clean and `-count=1` before every score. Each edit's match count was asserted **after** applying it (a silently unapplied edit reads ALIVE and argues the opposite), and every restore was `cp` from a shasum-verified backup (`fe2e0830…`), never `git checkout`.

| # | mutant | edits applied | verdict | killed by |
|---|---|---|---|---|
| M1 | revert `(?m)` on all 9 patterns | 9/9 | **DEAD** | `TestIssue6917_ModularSinatraAppYieldsRoutes` — `Routes found: []` |
| M2 | revert `(?m)` on the classic Route pattern **only** | 1/1 | **DEAD** | same, Route arm alone |
| M3 | widen helpers-include capture back to `(\w+)` | 1/1 | **DEAD** | forbidden row `Service "do"` |
| M4 | drop the `^` anchor from all 9, keep `(?m)` | 9/9 | **DEAD** | forbidden rows: comment route, `server.configure`, `server.before`/`after`, `Whatever` |
| M5 | drop `^` on the Route pattern only | 1/1 | **DEAD** | forbidden row `Route "/commented-out"` |
| M6 | drop `^` from the `helpers do` **block** pattern | 1/1 | **DEAD** (was ALIVE) | blanket sweep in `TestIssue6917_ReceiverPrefixedCallsAreNotSinatraDSL` |
| M7 | allow a receiver prefix on `error`: `^\s*` → `^\s*[\w.]*` | 1/1 | **DEAD** (was ALIVE) | `Middleware "500"` — forbidden row + decoy line |

Scored a second time **at the fixture level**, with mutant binaries built and run through `grafel quality`, so the `expected.json` rows are graded too and not merely present:

| mutant | fixture result |
|---|---|
| full `(?m)` revert | entity recall **38.5%** (8 must-haves missing), relationship recall **33.3%** — ratchet FAIL |
| M3 | `forbidden entity hits: 1` — `do [Service] in app.rb` |
| M4 | `forbidden entity hits: 2` — `Whatever [Service]`, `/never-registered [Route]`, both in `receiver_calls.rb` |
| M7 | `forbidden entity hits: 1` — `500 [Middleware] in receiver_calls.rb` |

### M6 and M7 — two mutants that survived review, and why they mattered

Both were found by the adversarial review, both were gaps in the **test surface** rather than in the fix, and both are #6902's lesson landing on my own work: a recall assertion cannot see over-firing, and **an aggregate ceiling is not a substitute for a named forbidden row**.

**M7** (`^\s*` → `^\s*[\w.]*` on the error rule) minted a false `Middleware "500"` from `server.error 500 do` while must-have recall stayed **13/13** and forbidden hits stayed **0** — so `grafel quality` returned **exit 0** and the engine test stayed green. The only thing that noticed was the ratchet's `entity_extracted_total` 61→62: an aggregate that cannot say *what* went wrong, and one somebody eventually re-baselines away. The evidence was already sitting in the fixture — `src/receiver_calls.rb` contained `server.error 500 do` and **nothing named it** — while the engine decoy enumerated the receiver-prefixed forms in its header comment and omitted the `error` one from its actual source. That gap between prose and content is now called out in the decoy's comment, so the next person adding a form is told the enumeration is load-bearing.

**M6** (drop `^` from the `helpers do` block pattern) survived by a different mechanism worth naming: **it hid by deduping**. `server.helpers do` matches as `helpers do` under `name_group: 0` — the identical name the decoy's own heredoc produces — and `detector.go:432` dedupes on `entityType:name:file`, so the false entity collapsed into a legitimate one and changed nothing observable. Closed with a second decoy file (`sinatraReceiverOnlyDecoys`) that has no heredoc to collide with, swept across **every** kind `sinatra.yaml` can mint. The named rows grade the forms somebody thought to name; the blanket sweep grades the ones nobody did.

No survivors remain.

## One limitation, pinned in the positive direction rather than hidden

`(?m)^\s*` reaches every line, and a regex cannot know a line lives inside a `<<~HEREDOC` body. Route-shaped text in a usage string **is** extracted, and this change does widen that case (before, such text had to open the file). Writing it as a forbidden row would make the test fail today and say nothing, so `TestIssue6917_MultilinePatternsDoNotOverFire` asserts it as **current behaviour**: whoever teaches the detector about heredocs gets a red test telling them to move those cases into the forbidden list. The pin covers **all three** YAML-minted heredoc entities — `Route "/from-a-heredoc"`, `Config "from_a_heredoc"` and `Service "helpers do"`. The third was missing at first review because the sweeps covered `Route` and `Middleware` but not `Service`; a `Service` sweep was added, which is also what kills M6. Comment decoys are genuinely excluded — `#` opens the line, so `^\s*get` cannot reach past it.

Two FINDINGs are pinned in `expected.json` rather than fixed here, both pre-existing and independent of the anchor defect:
- The fixture registers **four** routes and the graph carries **two** Route entities: `name_group: 2` names a Route by its path alone and `detector.go:432` dedupes on `entityType:name:file`, so the verb is lost and `get`/`post` on the same path collapse.
- `use Rack::Session::Cookie` — half the issue's driven evidence — lives in `src/config.ru`, and the indexer **skips that file entirely** (`config.ru` and `Gemfile` are not recognised as Ruby by extension). The `use` rule is fixed but ungradeable by the fixture, so `TestIssue6917_UseDirectiveAfterRequire` drives it directly at the engine level instead. **Filed as #6930** — `config.ru` is Rack's entry point for Rails, Sinatra, Roda and Hanami alike, so Rack middleware is structurally invisible for every Ruby repo, not just this fixture.

## Verification

| ran | result |
|---|---|
| `go test -count=1 ./internal/engine/` | ok (42s) |
| `go test -count=1 ./internal/quality/...` | ok — needed three corpus-size constants re-derived 31→32 (`absent_relationships_6490`, `zero_relationships_6490`, `subtype_assertion_6488`), which is what those guards exist to force |
| `go test -count=1 ./cmd/grafel/` | ok (156s) — the `custom_extractor_spans_6118` digest pin did **not** move; verified by running it, not assumed from "the fixture has no Ruby" |
| `go vet ./internal/engine/ ./internal/quality/ ./cmd/grafel/` | clean |
| `gofmt -l` on all touched trees | clean |
| `scripts/quality/run.sh --ratchet --runs 1` | exit 0, 32/32 held |

**Not run:** the full repository suite (only the touched packages plus `cmd/grafel`); `--runs 5` on the ratchet (used `--runs 1`, as the regenerate line prescribes); and no measurement of how the other 7 offending rule files behave under `(?m)` beyond reading their patterns — that is deliberately #6927's job, not a claim made here.


cajasmota:
## Independent adversarial review — **APPROVE** (two non-blocking asks)

Reviewed at `a4508468f` in a detached worktree of my own. Everything below was re-derived, not taken from the report. Mutants restored by `cp` from a shasum-verified backup and the restore verified with `shasum -c` after every score; `go vet` clean before each; `-count=1` throughout.

Every headline claim in the PR body reproduced. The two things I found are gaps in the *test* surface, not in the fix, and one reporting inaccuracy.

---

### Claim 1 — the sweep and the scope call: **verified, and the scope call is defensible but not fully argued**

I re-derived the sweep independently (walk `rules/<lang>/{frameworks,orms,queues}/*.yaml`, decode through `engine.FrameworkRule`, count `^`/`$` in **anchor position** — not inside a character class, not escaped):

```
DISK:  rule files=520  patterns=412
EMBED: candidates=520  loaded=520  failures=0  patterns=412
CARET-no-(?m): disk=18  embed=18   (at HEAD, i.e. post-fix)
DOLLAR-no-(?m): 4
caret files=7  langs=6
```

18 + sinatra's 9 = **27, in 8 files across 7 languages**. Exact match. The disk figure and the `LoadAllRulesReport()` figure agree, `candidates == loaded`, zero failures, and all 412 compile — so the structural decode genuinely **is** the compiled population, neither superset nor subset. `FileConvention.Pattern` is correctly excluded: `detector.go:156,177` never compiles it. The 4 `$`-bearing patterns are exactly the four named.

(A naive `strings.Contains(p, "^")` gives 155 across 22 files — that is character classes like `[^"']`, not anchors. Worth saying only because it is the count anyone re-checking by grep will get.)

### The `$` hazard is **real, and if anything understated**

Measured rather than read. `dormantBucketAliases` maps `docker → {dockerfile, yaml}`, and `docker_compose.yaml`'s service-key pattern carries no `requires_framework` and its rule file declares no `import_markers` — so it is **ungated on every `.yaml`/`.yml` in a repo**, not just compose files:

| pattern | corpus | plain | `(?m)` | captures under `(?m)` |
|---|---|---|---|---|
| `^\s{2}(\w[\w-]*):\s*$` | compose file | **0** | 4 | `web db pgdata default` |
| `^\s{2}(\w[\w-]*):\s*$` | k8s manifest | **0** | 2 | `labels selector` |
| `^\s{2}(\w[\w-]*):\s*$` | GH workflow | **0** | 3 | `push build lint` |
| `^\s{2}(\w[\w-]*):\s*$` | ansible playbook | **0** | 1 | `tasks` |
| `\s+(\w+\.\w+\.\w+):\s*$` | ansible playbook | **0** | 2 | `ansible.builtin.apt`, `ansible.posix.synchronize` |

Dead on every corpus today; under a compile-site `(?m)` it mints `Service` entities named `labels` and `selector` from a Deployment and `push`/`build`/`lint` from a workflow — and even on a genuine compose file it over-fires on volume and network keys. **Rejecting the compile-site fix is correct.**

**But that is the only thing the hazard rejects.** Of the 18 remaining caret patterns, **16 carry no `$` at all**, and the ansible FQCN pattern has no `^`, so a per-pattern caret fix would never touch it. The sweep therefore argues against flipping the compile site — it does *not* argue for stopping at sinatra. What is left holding the line is the second reason:

**"6 of 8 files have no golden fixture" — verified.** Only `graphql-schema-mini` and `scala-play-mini` exist. I also checked the adjacent claim that could have undercut it: the existing `python-*` fixtures contain no *indented* `class X(Y)` or `def test_` that would incidentally grade `pytest.yaml` / `sqlalchemy.yaml`'s multiline direction. They are genuinely ungraded.

**Verdict on scope:** narrow was the right call *for this PR*, but the honest framing is "the remaining 16 are a per-pattern fix waiting on fixtures", not "the sweep says don't". #6927 is filed and its numbers match mine exactly (18 in 7 files). Rot risk is real — the #5989 → #6260 shape is closing an issue while the thing it tracked stays unflipped — and the difference here is that #6927 carries a measured, per-file plan rather than a gate to be flipped later. I would hold it to that: a P1 is currently fixed in 1 of 7 affected languages.

### Claim 2 — the `helpers` capture: **verified in both directions**

Junk entity genuinely reachable post-`(?m)` and now excluded (M3 kills it at both levels). The direction that matters — does the tightening *miss* anything:

| form | Services extracted |
|---|---|
| `helpers AuthHelpers` | `[AuthHelpers]` |
| `helpers Foo::Bar` | `[Foo::Bar]` |
| `helpers Auth_Helpers` | `[Auth_Helpers]` |
| `helpers Foo, Bar` | `[Foo]` — second module lost, **but lost identically under the old `(\w+)`**, so no regression |
| `helpers my_helpers` | `[]` — the only loss, and Ruby modules are constants, so this is not a real form |
| `helpers do` | `[helpers do]` (block rule, correct) |

No regression.

### Claim 3 — the pre-existing python drift: **reproduced, and already filed**

Built a binary with `sinatra.yaml` reverted to `HEAD~1` (asserted `(?m)` count = 0 before building), and ran both binaries per fixture in a fresh isolated `HOME`/`GRAFEL_HOME`/`GRAFEL_DAEMON_ROOT`:

| fixture | baseline | HEAD | HEAD (2nd run) | sinatra-**reverted** |
|---|---|---|---|---|
| `python-django-mini` rel_total | 255 | 251 | 251 | **251** |
| `python-fastapi-mini` ent/rel_total | 28 / 57 | 27 / 54 | 27 / 54 | **27 / 54** |
| `python-rq-mini` ent/rel_total | 22 / 43 | 20 / 39 | 20 / 39 | **20 / 39** |

Byte-identical under the reverted binary, and deterministic across repeat HEAD runs. **The drift predates this change; restoring the three entries rather than re-recording was the right call** (re-recording would tighten someone else's growth-only ceiling over drift that is not this PR's).

Two corrections to the report:

1. The PR names **three** shrunk figures; there are **five**. `fastapi`'s relationship total also fell 57→54 and `rq`'s entity total also fell 22→20. Same conclusion, understated magnitude.
2. This does **not** need a new finding filed — **#6898 is already open and already records exactly this** ("three python fixtures sit 4/3/4 below their ceilings now"). The PR should cite it; a reader of this body would reasonably conclude it is unfiled.

### Claim 4 — fixture quality: **body matches label**

Read `src/*` against the claims. Nothing on line 1 anywhere; every DSL call indented inside `class MyApp < Sinatra::Base`; `error 404 do` at line 50 of 53. The control (`class MyApp < Sinatra::Base`, never `^`-anchored) is present and is exactly the construct that is green on both binaries. The 4 `http_endpoint_definition` rows do come from the Ruby extractor. No label-over-claiming here.

**`config.ru` is skipped — confirmed.** `nice_relationship_found: 0 / 1` in the fixture report, and the extension maps recognise `.rb`/`.rake` only. The `use` rule is genuinely ungradeable by any fixture and is driven at engine level instead. **I agree this is its own defect worth filing** — `config.ru` is the canonical Rack entrypoint and it is precisely where `use` middleware lives, so Rack middleware is structurally invisible to the indexer for every Ruby repo, not just this fixture.

**The heredoc pin is right in principle but incomplete.** Dumping every entity the shipped decoy produces:

```
Route                    "/from-a-heredoc"      <- pinned
Config                   "from_a_heredoc"       <- pinned
Service                  "helpers do"           <- NOT pinned, newly reachable via (?m)
http_endpoint_definition "http:GET:/from-a-heredoc"   } Ruby extractor,
SCOPE.Operation          "<inline GET /from-a-heredoc>" } pre-existing
```

The heredoc's `helpers do` line mints a `Service` too, and it is unpinned because the test blanket-sweeps `Route` and `Middleware` but has **no blanket `Service` sweep**. Pinning known-wrong behaviour positively is the right choice here (a forbidden row would be red today and say nothing); the pin just needs to cover all three YAML-minted heredoc entities.

### Claim 5 — mutants: 5 reproduced DEAD, **2 of mine survive**

| # | mutant | edits asserted | engine test | fixture (`grafel quality`) | ratchet |
|---|---|---|---|---|---|
| M1 | revert `(?m)` on all 9 | 9/9 | **DEAD** — `Routes found: []` | **DEAD** — 5/13 ent (**38.5%**), 1/3 rel (**33.3%**) | FAIL |
| M2 | revert `(?m)` on Route only | 1/1 | **DEAD** — `Routes found: []` | — | — |
| M3 | widen helpers capture to `(\w+)` | 1/1 | **DEAD** — `Service "do"` | **DEAD** — `forbidden_entity_hits=1`, rc=2 | FAIL |
| M4 | drop `^` on all 9, keep `(?m)` | 9/9 | **DEAD** — 8 over-fire errors | — | — |
| M5 | drop `^` on Route only | 1/1 | **DEAD** — `Route "/commented-out"` ×2 | — | — |
| **M6** | *(mine)* drop `^` from the `helpers do` **block** pattern only | 1/1 | **ALIVE** | **ALIVE** — 61/118, identical to baseline | pass |
| **M7** | *(mine)* allow a receiver prefix on `error`: `^\s*` → `^\s*[\w.]*` | 1/1 | **ALIVE** | **ALIVE** — 13/13 recall, `forbidden_entity_hits=0`, **rc=0** | **FAIL** (61→62, 118→119) |

Every mutant's edit count was asserted after applying it, and each produced a distinct file digest, so none read ALIVE because it silently failed to land. The M1 fixture figures reproduce the author's 38.5%/33.3% exactly.

**M7 is the finding.** It is the over-firing direction the forbidden rows do not cover. `src/receiver_calls.rb` already contains `server.error 500 do`, and there is **no forbidden row for `Middleware "500"`** — so a rule change that lets a receiver-prefixed `error` through mints a false entity, keeps 13/13 recall, keeps 0 forbidden hits, and the fixture's own gate returns **exit 0**. Only the ratchet's growth ceiling (`entity_extracted_total` 61→62) catches it, i.e. it is caught by an aggregate that cannot say *what* was wrong, not by an assertion that names it. The engine decoy misses it for a different reason: it has `server.helpers` / `server.configure` / `server.before` / `server.after` but **no `server.error` line at all**, despite its comment enumerating the receiver-prefixed forms.

Three-way verdict on M7: not equivalent (it demonstrably emits a new entity), not a bad mutant (a receiver-prefix leak is the same class the file's other decoys exist to catch) — it is a **genuine assertion gap**, cheap to close.

M6 three-way verdict: **behaviourally live but inert on both corpora** — the fixture has no `receiver.helpers do`, and in the engine decoy the false entity's name (`name_group: 0`) collides with the heredoc-born `Service "helpers do"` and dedupes away. Real gap, low value; I would not block on it.

---

### Asks (non-blocking, ~3 lines total)

1. Add `{ "name": "500", "kind": "Middleware", "must_exist": false }` to the fixture's `forbidden_entities`, and a `server.error 500 do` line to `sinatraDecoys` — that kills M7 at both levels and uses evidence already sitting in the fixture.
2. Pin `Service "helpers do"` alongside the two heredoc cases (or add a blanket `Service` sweep with the known cases excepted), so the "whoever teaches the detector about heredocs gets a red test" promise covers all three.
3. Cite **#6898** where the python drift is discussed, and correct the enumeration to five figures (fastapi rel 57→54 and rq ent 22→20 are also down).
4. Consider filing the `config.ru` / `Gemfile` extension gap — Rack `use` middleware is invisible to the indexer for every Ruby repo.

### Verification I ran

| ran | result |
|---|---|
| `go vet ./internal/engine/ ./internal/quality/... ./cmd/grafel/` | clean |
| `gofmt -l` on touched trees | clean |
| `go test -count=1 ./internal/engine/` | ok (42s) |
| `go test -count=1 ./internal/quality/...` | ok (4 packages) |
| `go test -count=1 ./cmd/grafel/` | ok (142s) — **the #6118 digest pin did not move**, confirmed by running the whole package, not by name-matching a test |
| `scripts/quality/run.sh --ratchet --runs 1` (isolated `HOME`+`GRAFEL_HOME`+`GRAFEL_DAEMON_ROOT`, `GRAFEL_BIN` = a binary I built) | **exit 0 — 32 gated fixtures held their baseline (23 at 100% must-have recall)** |
| baseline diff | one addition + the two stamps; no existing fixture's figures changed |

**Not run:** the full repository suite; `--runs 5`; any measurement of the other 7 rule files under `(?m)` beyond the two `$` patterns I probed directly; Windows/Linux legs.

cajasmota:
Thanks — this was a genuinely useful review, and both mutants were real. Addressed in `4c1f5e9`, pushed to the same branch.

## M7 — closed at both levels

Exactly as you specified: a `Middleware "500"` forbidden row in the fixture, and the missing `server.error 500 do` line in `sinatraDecoys`.

| level | before | after |
|---|---|---|
| engine test | ALIVE | **DEAD** — `over-fired: Middleware "500"` |
| `grafel quality` | ALIVE (13/13, 0 forbidden, rc=0) | **DEAD** — `forbidden_entity_hits=1`, `500 [Middleware] in receiver_calls.rb` |

Edit count asserted at 1/1 after applying, `go vet` clean, `-count=1`, restored by `cp` from a shasum-verified backup (`fe2e0830…`).

The part worth keeping is your diagnosis of *how* it hid: the fixture already contained the evidence and nothing named it, while the decoy's header comment enumerated the receiver-prefixed forms and its source omitted the `error` one. That gap between prose and content is now stated in the decoy's own comment, so the next person adding a form is told the enumeration is load-bearing rather than decorative.

## Heredoc pin — extended, and it also killed M6

Added the blanket `Service` sweep. It pins `Service "helpers do"` as the third YAML-minted heredoc entity, so the "red test tells you to flip these" promise now covers all three.

I did not stop at your non-blocking verdict on **M6**, because your dedupe analysis pointed straight at the fix. The false entity hid by colliding with a legitimate name in the same file — so I isolated the receiver-prefixed forms into a second decoy (`sinatraReceiverOnlyDecoys`) with no heredoc to collide with, swept across every kind `sinatra.yaml` can mint:

```
M6: DEAD — over-fired: Service "helpers do" — every sinatra-shaped call in this
    file is prefixed with a `server.` receiver and none of them is a DSL invocation
```

Named rows grade the forms somebody thought to name; the blanket sweep grades the ones nobody did. **No survivors remain.**

## Body corrections

- **Five shrunk figures, not three** — corrected, with `fastapi` rel 57→54 and `rq` ent 22→20 spelled out.
- **#6898 cited.** You were right that a reader would have concluded it was unfiled.
- **Scope framing reworked**, taking your point that the `$` hazard rejects only the *compile-site* fix — 16 of 18 carry no `$`, and the ansible FQCN pattern has no `^` for a per-pattern fix to touch. The body now says what actually holds the line (6 of 8 files ungraded, including your check that the `python-*` fixtures contain no indented `class X(Y)` / `def test_`), and states plainly that **a P1 is fixed in 1 of 7 affected languages and #6927 must land**. Your rot-risk point is the reason that is in the body rather than in a comment.
- **#6930 filed** for the `config.ru` / `Gemfile` gap, with your framing: Rack's entry point is where `use` middleware lives, so it is invisible for every Ruby repo — Rails, Roda and Hanami included, not just this fixture. #6926's `use` rule is now correct and still emits nothing on a real repo until that lands.

## Re-verified after the change

| ran | result |
|---|---|
| `go test -count=1 ./internal/engine/` | ok (42s) |
| `go test -count=1 ./internal/quality/...` | ok (4 packages) |
| `go vet` + `gofmt -l`, touched trees | clean |
| `scripts/quality/run.sh --ratchet --runs 1` | **exit 0 — 32 gated fixtures held** |
| `baseline.json` | unchanged by this commit — a forbidden row and test-only decoys move no recorded figure |

**Not re-run:** `./cmd/grafel/` (this commit touches only a `_test.go` file and an `expected.json`; neither is an input to the #6118 digest, which was already confirmed unmoved at `a4508468f` by both of us), the full repo suite, and `--runs 5`.

cajasmota:
## Coordinator mutant — the relationship half, split verdict

Every mutant scored so far targets the **source** patterns or the `^` anchor: M1 (revert all 9), M2 (Route only), M3 (widen the helpers capture), M4/M5 (drop `^`), M6, M7. Nobody isolated the **two `relationship_rules` patterns** — `:126` (the `get … *_helper` CALLS rule) and `:142` (the `use` rule).

That matters because the two are graded by *different artefacts*, and one of them by only one.

| mutant | `./internal/engine/` | ratchet |
|---|---|---|
| revert `(?m)` on **both** relationship patterns | **DEAD** — `TestIssue6917_UseDirectiveAfterRequire` | — |
| revert `(?m)` on **`:142` (`use`) only** | **DEAD** — same test | — |
| revert `(?m)` on **`:126` (CALLS) only** | **ALIVE — whole engine suite green** | **DEAD** — `ruby-sinatra-modular-mini: relationship_found REGRESSED 3 -> 1 (of 3 must-have)` |

Each edit gated on its exact occurrence count (10 → 8, then 10 → 9), restored by `cp` from a shasum-verified backup (`fe2e08300…`), tree confirmed pristine.

### The finding

**The CALLS relationship rule has no engine-level coverage at all.** `./internal/engine/` is entirely green with `(?m)` reverted on it. It is graded only by the golden fixture's two must-have rows (`/invoices -> json_helper`, `/invoices/:id -> json_helper`).

That is **not** an M7-shaped hole — those are *named* must-have assertions, not an aggregate ceiling, and they fail with a precise diagnosis. So the rule is genuinely graded and I am not asking for a change.

But it is worth stating in the body, because of who pays: an author iterating on `sinatra.yaml` runs `./internal/engine/` — it is seconds, the ratchet is minutes. Under that loop, breaking the CALLS rule looks green. The `use` rule has an engine test precisely because `config.ru` is unreadable (#6930) and no fixture could grade it; the CALLS rule has the mirror-image gap and no note saying so.

**One line in the decoy or the rule comment naming `ruby-sinatra-modular-mini`'s relationship rows as the CALLS rule's only grader** would close the loop for the next person — the same reason #6923 and the `mustScan` correction were worth writing down.

### Incidental

`TestApplyCommitCoupling_Idempotent` also failed in the first run and is unrelated to this rule (commit coupling, no sinatra involvement); it passed in the isolated re-run. Consistent with the load-sensitive flakes the review already noted under concurrent agent load — flagging it, not attributing it.

Refs #6917, #6930.

Issue #6932, opened 2026-09-06
DECIDED (0.3.3): auto-switch fsnotify -> git+manifest polling when the inotify budget will not fit, with a flag override to force it on


cajasmota:
**Owner-decided, 2026-09-06.** Target: **0.3.3**, alongside whatever else lands meantime.

## The decision

Replace fsnotify with a polling change-detector **when the inotify budget will not fit**, switching automatically on the right trigger, with a **flag override** so a known environment (containers) can force it on permanently.

## Why — the constraint, measured

On Linux, inotify costs **one watch descriptor per DIRECTORY**, recursively. Measured on this repo through the real `AddRepo`: **976 directories per worktree.** Each tracked linked worktree subscribes its entire working tree, and `GRAFEL_MAX_WORKTREES_PER_REPO` defaults to 10, so **one lane can reach ~10,700 descriptors**.

`fs.inotify.max_user_watches` is **per-UID, host-level, and not namespaced** — every container running as the same UID draws from one pool, and a container cannot raise it. Nothing in grafel bounds this today: `walk.WatchDirCap` is 100,000 per repo (larger than the whole kernel pool, so it cannot bind first) and `defaultFDBudgetLimit()` returns **0 on Linux**, disabling accounting entirely.

**The failure mode is silent** — a watcher that is registered and receives no events. That is the same class as #6921 (overflow logged and ignored), #6923 (a guard that cannot see partial erosion) and #6928 (a second watcher discarding its errors).

## The mechanism — measured, three variants tried

A spike drove a 14-case matrix against real fsnotify, real `git status`, and grafel's real `diff.Filter`/`walk.WalkRepo`:

| approach | inotify wd | complete? | idle cost |
|---|---|---|---|
| fsnotify | **976/worktree** | yes | ~0 |
| `git status` polling alone | 0 | **misses 6 of 12** | ~30 ms |
| hybrid A — `-uall` for candidates | 0 | **misses 2**, incl. permanent corruption | 173 ms |
| **hybrid B — `-unormal` discovery + manifest-key sweep** | **0** | **yes** (see prerequisites) | **60 ms** |
| full manifest sweep | 0 | yes | 271 ms |

[...]

PR #6933, opened 2026-09-06
fix(#6931,#6922): the walker consulted .gitignore for directories only, and never read the other two ignore sources git has

### LINKED ISSUE #6922: `.git/info/exclude` and `core.excludesFile` are read by nothing — grafel indexes files git will never report at any `-u` level
`.git/info/exclude` and `core.excludesFile` are read by nothing — grafel indexes files git will never report at any `-u` level

Found by the git-polling spike (#6726 consumer context). Measured, not inferred.

## The defect

Git has **three** ignore sources, not one:

1. `.gitignore` files in the tree
2. `.git/info/exclude` — per-clone, uncommitted
3. `core.excludesFile` — per-user global (commonly `~/.config/git/ignore`)

grafel reads only the first. `.git/info/exclude` and `core.excludesFile` appear **nowhere** in `internal/daemon/walk/` or `internal/daemon/watch/` — verified by grep across both packages.

## Measured consequence

On a driven fixture, a file excluded via `.git/info/exclude`:

| observer | result |
|---|---|
| grafel `walk.WalkRepo` | **indexed** (file set 6 → 7) |
| fsnotify watcher | **HIT** (52 ms) |
| `git status --porcelain` | MISS |
| `git status --porcelain -uall` | MISS |
| `git status -uall` + `git diff HEAD` | MISS |

So grafel indexes a file that git will not report **at any `-u` level**. Editing that file afterwards is likewise invisible to git and visible to grafel.

## Why this is worth fixing on its own merits

Two separate reasons, and the second is the one that generalises:

**It breaks any git-derived change signal.** This is what the spike hit: a git-status poller cannot be made complete while grafel's index set is a superset of git's reported set. That is a blocker for the "replace watchers with git polling" direction, and it will be a blocker for any future git-based incremental input.

**It is a correctness gap regardless of polling.** `.git/info/exclude` is the standard place to ignore machine-local scratch files without touching a shared `.gitignore` — build scratch, editor droppings, local fixtures. grafel currently indexes all of it and presents it as repo architecture. That is noise in the graph today, with no watcher involved.

## The pleasant surprise, recorded so nobody re-investigates it

**The feared `.gitignore` divergence measures ZERO.** grafel applies `.gitignore` in three places — directory subscription (`ShouldSkipDirGitignore`), the event boundary (`ShouldSkipPathForRepo`, `skip.go:229`), and `walk.WalkRepo` — and all agree with git. On a real checkout of this repo: `walk.WalkRepo` yields **7069** files, every one inside `git ls-files` (**8209**), with the only exceptions being spike scratch files. So the `.gitignore` layer is sound; it is specifically the other two ignore sources that are missing.

1. **Where to read them.** `walk.ParseIgnoreFile`'s caller is the natural seam — the parser already exists, it is only ever pointed at `.gitignore`.
2. **Precedence.** Git's order is: command-line, `.gitignore` (deepest first), `.git/info/exclude`, `core.excludesFile`. Getting precedence wrong produces a *different* wrong answer, not a safer one.
3. **`core.excludesFile` is a config read, not a fixed path.** It must come from `git config --get core.excludesFile` (or the `$XDG_CONFIG_HOME` fallback), not a hardcoded `~/.config/git/ignore`.
4. **Whether this changes the indexed set on real repos, and by how much.** It shrinks the graph — so it will move `entity_extracted_total` **down**, which the growth-only ratchet permits silently. **Measure the delta deliberately** rather than letting it pass unnoticed; a silent shrink is the shape #6898 exists for.
5. **The forbidden direction.** A test proving excluded files are skipped cannot see over-exclusion. Assert both.

## Not established

- Any corpus count of how many files this currently over-indexes. Measured on a purpose-built fixture only.
- Whether `.grafelignore` interacts with these (it is a separate layer and was not exercised).
- Sparse-checkout and submodule behaviour — untested.
- Whether the watcher and the walker would need the fix in both places or can share one resolver. Two call sites were identified; a shared implementation was not designed.

Refs #6726, #6912.

### LINKED ISSUE #6931: walk.WalkRepo applies .gitignore to DIRECTORIES only, never to files — a pattern like *.pb.go is ignored by git and indexed by grafel
`walk.WalkRepo` applies `.gitignore` to DIRECTORIES only, never to files — a file-level pattern like `*.pb.go` is ignored by git and indexed by grafel

Found by the git+manifest hybrid spike. **This corrects a claim I made on #6922** and is a second, larger divergence than the one that issue names.

## The defect

In `internal/daemon/walk/walker.go`, the `igStack.Match(rel)` call sits **inside the `d.IsDir()` branch** (~line 208). The file branch below it (~line 245 onward) consults the extension filter, the FIFO gate and the sparse filter — and **never consults the ignore stack at all.**

So `.gitignore` prunes ignored *directories* but never excludes an ignored *file*.

Driven on a fixture — three patterns, each matched by `git check-ignore -v`, none affecting grafel:

| `.gitignore` entry | `git check-ignore` | grafel walk |
|---|---|---|
| `untracked_indexed.go` | matched | `files=7` — unchanged |
| `/untracked_indexed.go` | matched | `files=7` — unchanged |
| `*.go` | matched | `files=7` — unchanged |

## Correcting the record on #6922

I reported there that *"the feared `.gitignore` divergence measures ZERO"*, based on `walk.WalkRepo`'s 7069 files all being inside `git ls-files`' 8209.

**That measurement only tested one direction.** `walk ⊂ git ls-files` shows grafel does not index files git has never heard of; it cannot show grafel skipping files git *ignores*, because an ignored file is not in `ls-files` either — both sets simply exclude it. The fixture used to spot-check was also a **directory** pattern, which is precisely the case that does work.

So the conclusion was right for directory patterns and untested for file patterns, and I stated it as general. The `.gitignore` layer is **not** sound; it is half-implemented.

## Practical size today, and why that is not reassuring

On this repo: **2 of 7069 walked files** are `git check-ignore`-positive (two `graph.json` testdata files). Small.

But it is unbounded in principle, and the shape that would hurt is ordinary: a repo with `*.pb.go`, `*_generated.go`, `*.min.js` or `dist/**/*.js` in `.gitignore` gets every one of those files indexed as architecture. Generated code is exactly what a `.gitignore` file-pattern is usually *for*.

## Why it matters beyond graph noise

It blocks the change-detection work. A git-derived signal can only be complete if grafel's indexed set is a **subset** of what git reports. Two things break that today:

1. `.git/info/exclude` and `core.excludesFile` are read by nothing (**#6922**).
2. This — file-level `.gitignore` patterns are not applied.

The hybrid spike measured that fixing **both** would make its `hybB` variant complete on all 14 detection cases; fixing only #6922 would not. So this is on the critical path for replacing fsnotify, not merely a tidiness issue.

1. **Apply the ignore stack in the file branch**, matching git's precedence (deepest `.gitignore` first, then `.git/info/exclude`, then `core.excludesFile` — the last two being #6922).
2. **Measure the shrink deliberately.** This removes files from the index, so `entity_extracted_total` moves **down**, which the growth-only ratchet permits silently. #6898 exists because that class of change slips through unnoticed — enumerate the delta rather than letting a `--update-baseline` absorb it.
3. **Assert both directions.** A test proving ignored files are skipped cannot see over-exclusion. #6902's lesson.
4. **Check whether a downstream consumer already re-filters.** `ShouldSkipPathForRepo` in `skip.go` was **not** read by the spike; if the watcher path re-filters through it, the practical scope differs between the walker and the watcher, and the two should agree.

## Not established

- Whether any downstream consumer of `WalkRepo` output re-applies the ignore stack (see 4 above — `skip.go` unread).
- `.grafelignore` interaction, sparse checkout, submodules, symlinks — untouched.
- Any corpus count beyond this repo's 2/7069.
- Whether the watcher's directory-subscription path has the same file/directory asymmetry.

Refs #6922, #6898, #6902.

cajasmota:
Arm C of #6932. Both defects share one seam, so they land together — fixing either alone leaves the gap open.

> **Revised after REQUEST-CHANGES** (`af23b5560`). All seven items addressed; the shrink section below is re-measured on the main clone and the earlier numbers in it were wrong. Two findings from the redo are filed as #6936 and #6937 rather than folded in — **#6937 needs a coordinator decision**, see the last section.

## The defect

`walk.WalkRepo` called `igStack.Match(rel)` **inside** the `d.IsDir()` branch. The file branch below it consulted the extension filter, the entry-type gate and the sparse filter, and **no ignore layer at all**. `.gitignore` pruned ignored *directories* and never excluded an ignored *file* (#6931). Separately, git reads three ignore sources and grafel read one — `.git/info/exclude` and `core.excludesFile` appeared nowhere in `internal/daemon/walk/` or `internal/daemon/watch/` (#6922).

## The fix

- **`IgnoreFile.MatchPath(rel, isDir)` / `IgnoreStack.MatchFile`.** `MatchDir` is now `MatchPath(rel, true)`, so directory behaviour is unchanged by construction. `isDir` is load-bearing for exactly one pattern form: `dirOnly` (`build/`) was parsed and never consulted, which is harmless while only directories ask and becomes an over-exclusion the moment files do.
- **The walker's file branch consults the stack** — after the entry-type gate and before the sparse filter, and the skip is **reported** through the same `SkipEntry` + `PrintSkipped` channel as every other layer (#6338).
- **The nested-ignore pop now runs for both entry kinds** (blocker 1 below).
- **`gitexclude.go`** resolves `.git/info/exclude` via `rev-parse --git-common-dir` (a linked worktree shares the main clone's) and `core.excludesFile` via `git config --get` with git's own `$XDG_CONFIG_HOME/git/ignore` fallback and `~/` expansion. Both are pushed **below** everything in the tree, so any `.gitignore` or `.grafelignore` rule — including a negation — outranks them. Both are read through `readIgnoreFile`, not `os.ReadFile`.

---

### 1. BLOCKER — new over-exclusion: a nested `.gitignore` deleted files above itself. FIXED.

`WalkRepo` popped nested ignore files only inside the `d.IsDir()` branch. `filepath.WalkDir` is lexical DFS, so after a subtree finished, every **file** visited before the next **directory** entry still saw that subtree's `.gitignore` on the stack. Harmless while the stack only decided directories — a directory entry pops first — and live the moment the file branch asks.

Reproduced from the reviewer's probe before touching the code:

```
skipped pkg/sub/victim.go rule=.gitignore line 1
skipped pkg/victim.go     rule=.gitignore line 1   <- git keeps it
skipped victim.go         rule=.gitignore line 1   <- repo root, killed from two levels down
walked = [pkg/sub/.gitignore pkg/sub/inside.go]
```

The pop loop is hoisted out of the `IsDir` branch and now runs for every entry, before either branch. `TestWalkRepo_NestedIgnoreDoesNotLeakToSiblingFiles_6931` pins it with an exact-set assertion; **M12** (revert the hoist) is DEAD.

The ordering *is* the bug, so every row is chosen for its DFS position, not its name — and all four rows share the basename `victim.go`, so a row that passed by name would prove nothing:

| row | position relative to `pkg/sub/.gitignore` | verdict |
|---|---|---|
| `pkg/sub/inside.go` | inside the subtree | walked (rule does not match it) |
| `pkg/sub/victim.go` | inside the subtree | **skipped** — genuinely governed |
| `pkg/victim.go` | sibling one level up, sorts after `sub` | **walked** |
| `victim.go` | repo root, sorts after `pkg` | **walked** |

### 2. BLOCKER — claims 5 and 6 were false. RE-MEASURED on the main clone.

The reviewer is right about the cause and right that the direction was wrong; the corrected figure is **larger than either of us had**. My first measurement ran inside a linked worktree. The reviewer's correction was taken on the main clone but reused my `2` for the `.gitignore` half, so the published total of 7 is also short.

Re-measured on `/Users/jorgecajas/Projects/archigraph`, three builds, same tree, read-only:

| build | walked |
|---|---|
| C — `main` | 7090 |
| B — this branch, `gitExcludeIgnoreFiles` returning nil | 7084 |
| A — this branch as-is | **7079** |

**Total shrink 11 files. Nothing added** (`A − C` is empty). Complete, both directions, each attributed by `git check-ignore -v`:

| file | rule | source |
|---|---|---|
| `.DS_Store` | `.gitignore:35` | #6931 |
| `grafel` | `.gitignore:15` `/grafel` | #6931 |
| `mcp.test` | `.gitignore:19` `*.test` | #6931 |
| `webui-v2/tsconfig.tsbuildinfo` | `webui-v2/.gitignore:7` `*.tsbuildinfo` (nested) | #6931 |
| `cmd/grafel/testdata/audit_mount_prefix_retry/consumer/graph.json` | `.gitignore` `graph.json` | #6931 — **see #6937** |
| `cmd/grafel/testdata/audit_mount_prefix_retry/producer/graph.json` | `.gitignore` `graph.json` | #6931 — **see #6937** |
| `presentation-kit/01-what-is-grafel.md` … `05-talking-points-and-faq.md` (5) | `.git/info/exclude:19` `presentation-kit/` | #6922 |

**#6931 alone: 6 files. #6922 alone: 5 files.** Claim 6 ("#6922 contributes zero here") was false; the linked worktree simply does not contain `presentation-kit/`.

**Entity delta re-derived: 7, not 0.** Measured, not argued — six of the eleven are classifier-skipped (`unsupported_extension` for `.DS_Store` / `grafel` / `mcp.test` / `.tsbuildinfo`, `vendor_testdata` for both `graph.json`), so they contribute nothing. The five `presentation-kit/*.md` classify as `markdown` and do reach an extractor; running it over them yields **1+1+1+3+1 = 7** Pass-1 entities. The `cmd/grafel` whole-graph digest is unchanged and the ratchet holds, consistent with the two `graph.json` files being `vendor_testdata`-skipped either way.

Both figures are **machine-relative**, and that is inherent to #6922: `presentation-kit/`, `.DS_Store`, `grafel` and `mcp.test` are untracked local state, and `core.excludesFile` is per-developer. On CI the shrink is smaller.

### 3. Test environment isolation. FIXED via `testsupport.IsolateHome`, with both controls re-established.

`hermeticGitEnv(t)` calls `testsupport.IsolateHome(t)` — which sets `HOME`, `USERPROFILE`, `GRAFEL_HOME`, `GRAFEL_DAEMON_ROOT` and both XDG vars together and asserts the redirect actually left the real user home — and keeps its own `GIT_CONFIG_GLOBAL` / `GIT_CONFIG_SYSTEM` empty-file pins, which `IsolateHome` does not touch and which are what neutralise a hostile global git config.

The first attempt redirected `HOME` by hand and was **red in CI** on `internal/registry`'s #6735 sweep guard. The guard is right about the mechanism, not merely about style: `registry.HomeDir()` prefers `GRAFEL_HOME` and only falls back to the OS home, so a fixture that moves `HOME` while an ambient `GRAFEL_HOME` survives has the code under test resolving against a directory the fixture never wrote to — and the sandbox recipe used here sets `GRAFEL_HOME`. Not added to `grafelHomePinDeferred` and not `deliberatelyUnpinnedHome`: this fixture asserts nothing about home resolution, it only needs isolation.

Worth recording: the guard is a **static sweep over test sources**, not a runtime check, which is why it caught this in CI even though CI never sets `GRAFEL_HOME` and so could never have reproduced the failure it prevents.

Both controls verified **after** the swap rather than assumed:

1. **Hostile environment** — `XDG_CONFIG_HOME` pointing at a dir whose `git/ignore` holds `*.go` and `*.json`, plus a `GIT_CONFIG_GLOBAL` setting `core.excludesFile`: `./internal/daemon/walk/` green, whole package. On the pre-fix tests the same environment produced `--- FAIL: TestInfoExcludeIsSharedWithLinkedWorktrees_6922` (its own `git add` skipped the seed files, so the seed commit was empty) and `--- FAIL: TestGlobalExcludesPath_ConfigThenXDGFallback_6922`.
2. **MC-4** — the XDG pin removed. It now lives inside `IsolateHome`, so the equivalent mutant overrides `XDG_CONFIG_HOME` back to empty immediately after the call. Still **RED on a clean environment**, still on `TestGlobalExcludesPath_ConfigThenXDGFallback_6922`. The grading did not move: the kill rests on the XDG-before-home *ordering*, and `HOME` stays pinned either way, so the pin remains self-graded rather than a manual procedure.

M12, R2, R3 and MC-2b were re-scored after the swap — all still DEAD.

### 4. MC-2b — the tilde was graded only as a helper. FIXED at the call site.

The fixture now writes the global ignore under the fake home and configures `core.excludesFile = ~/global_ignore` — the spelling git's own docs use — instead of an absolute path, driving `expandTilde` end to end. **MC-2b is now DEAD** (kills `TestWalkRepo_AppliesIgnoreStackToFiles_6931` and `…IgnoredFilesAreReported…`), and MC-2 with it.

### 5. R2 and R3 — both ALIVE mutants graded. Two FIFO tests, both under `t.TempDir()`.

- **R2** — `TestIgnoredFIFOIsReportedAsAHazardNotAsIgnored_6931`: `mkfifo ignored.pb.go` with `*.pb.go` ignored. Both orderings `return nil`, so only the reported reason differs: head reports `irregular:…`, the mutant reports `.gitignore line 1`. The fixture holds the *name* constant across a FIFO row and a regular-file row and varies only the entry type, so the control cannot pass by name. **DEAD.**
- **R3** — `TestInfoExcludeFIFODoesNotHang_6416`: `mkfifo .git/info/exclude`. Pinned as a deadline plus the always-on `reportIgnoreSkip` line, because the mutant produces no value to assert on. Under `os.ReadFile` it parks in `open(2)`: `HANG: WalkRepo with a FIFO .git/info/exclude did not return within 10s`. **DEAD.**

### 6. The stale doc block. DELETED, count corrected.

The orphaned first block on `gitExcludeIgnoreFiles` — which asserted the rewrite is reused, the opposite of the code — is gone. The surviving block now states the cost as **three** bounded git calls (`rev-parse --show-toplevel`, `config --get core.excludesFile`, `rev-parse --git-common-dir`), all outside the `WalkDir` callback, per walk and not per entry.

### 7. `IgnoreFile.Dir` is dead. FILED as #6936; the claim is softened, in the test file and here.

**Chosen: file, not fix.** The reason is not size. `filepath.Rel(d, filepath.Join(d, rel))` is `rel` for every input, so `ig.Dir` never applies — but the fix has to change `MatchPath`, and `MatchDir` is `MatchPath(rel, true)`. This PR's central safety property, the one the reviewer independently verified, is that **directory behaviour is unchanged by construction**; a `Dir` fix breaks that guarantee and needs its own directory-side measurement and its own shrink enumeration, in the same direction (it removes more files, so the growth-only ratchet stays silent, #6898).

The claim is now bounded rather than restated: the fixture's nested rows are **bare-name patterns only**, and *nested × anchored* and *nested × slash-bearing glob* are knowingly ungraded, with the reason written at the top of the test file.

---

## Fixture: what is varied, what is held constant

One fixture, asserted as **exact set equality**, so over-exclusion fails as loudly as under-exclusion (#6902) — every kept file is a forbidden row for the ignore predicate.

| form | pattern | positive row | forbidden row (must still be walked) |
|---|---|---|---|
| bare name | `bare_ignored.go` | `bare_ignored.go`, `pkg/bare_ignored.go` | — |
| rooted | `/rooted_ignored.go` | `rooted_ignored.go` | `pkg/rooted_ignored.go` |
| glob | `*.gen.go` | `a.gen.go`, `pkg/b.gen.go` | — |
| negation | `!kept.gen.go` | — | `kept.gen.go` |
| dir-only | `dironly.go/` | `pkg/dironly.go/inside.go` (a DIR) | `dironly.go` (a FILE, same name) |
| nested overriding a parent | `pkg/sub/.gitignore`: `!bare_ignored.go` | — | `pkg/sub/bare_ignored.go` |
| nested scope does not leak up | `pkg/sub/.gitignore`: `nested_only.go` | `pkg/sub/nested_only.go` | `nested_only.go` at root |
| `.git/info/exclude` | `excluded_by_info.go` | `excluded_by_info.go` | — |
| `core.excludesFile`, via `~/` | `excluded_by_global.go` | `excluded_by_global.go` | — |
| precedence `.gitignore` > `info/exclude` | `!prec_a.go` vs `prec_a.go` | — | `prec_a.go` |
| precedence `info/exclude` > `core.excludesFile` | `!prec_b.go` vs `prec_b.go` | — | `prec_b.go` |
| plain control | — | — | `keep.go`, `pkg/keep.go` |

Plus, in their own tests: nested-ignore DFS position (4 rows, one basename); FIFO vs regular file under one ignore rule; `.git/info/exclude` through a real `git worktree add`; the top-level-only scope limit; `core.excludesFile` config-vs-XDG order; `~/` expansion.

**Held constant:** extension `.go` on every row of the main fixture, so the extension filter can never be the thing doing the work; one repo root that IS the git top-level; no sparse checkout; no `.grafelignore`; a fully pinned git environment.

**Knowingly ungraded:** nested × anchored and nested × slash-bearing glob (#6936).

## Mutants — 15, all DEAD

Every mutant is at the call site, applied by exact-occurrence-count assertion, and restored by `cp` from a shasum-verified backup.

| # | mutant | killed by |
|---|---|---|
| M1 | delete the file-branch ignore check (the #6931 fix) | exact-set + report + watch agreement |
| M2 | `MatchFile` → `Match` (isDir=true) at the call site | exact-set (`dironly.go`) |
| M3 | delete the `gitExcludeIgnoreFiles` push (the #6922 fix) | exact-set + report |
| M4 | push the exclude sources ABOVE `.gitignore` | exact-set (`prec_a.go`) |
| M5 | swap `info/exclude` vs `core.excludesFile` order | exact-set (`prec_b.go`) |
| M6 | invert the dirOnly guard | exact-set + dirOnly unit |
| M7 | drop ignored files silently (no `SkipEntry`) | report test |
| M8 | `--git-dir` instead of `--git-common-dir` | linked-worktree test |
| M9 | drop the `core.excludesFile` config lookup | exact-set + resolution order |
| M10 | `expandTilde` helper is a no-op | expandTilde unit |
| M11 | drop the top-level gate | sub-root scope test |
| **M12** | **revert the pop hoist (back inside `d.IsDir`)** | **nested-pop test** |
| **R2** | **ignore block above the entry-type gate** | **ignored-FIFO test** |
| **R3** | **`readIgnoreFile` → `os.ReadFile`** | **FIFO-hang deadline** |
| **MC-2b** | **`expandTilde` CALL SITE deleted, helper intact** | **exact-set (tilde-configured fixture)** |

R1, R4, R5 and MC-1 were already DEAD against the reviewer's and coordinator's own scoring and are unchanged by this commit.

## Verification

`-count=1`, full suites: `./internal/daemon/walk/`, `./internal/daemon/watch/`, `./internal/registry/` (the #6735 guard's package), `./internal/gitmeta/`, `./internal/classifier/`, `./internal/extractors/` — all ok. `./cmd/grafel/` ok (148 s), digest unchanged. `scripts/quality/run.sh --ratchet`: 32 fixtures held. `gofmt -l` clean, `go vet ./internal/daemon/...` clean. Plus the hostile-environment run in item 3.

## Findings filed, not folded in

- **#6936** — `IgnoreFile.Dir` is dead; nested anchored rules never fire. Pre-existing, an under-exclusion, and it cannot be fixed without changing `MatchDir`.
- **#6937** — **needs a decision.** grafel's ignore layers are index-unaware, and git never ignores a **tracked** file. Both `graph.json` files above are **tracked** (`git ls-files --error-unmatch` matches them; `git check-ignore` reports them NOT ignored, because it excludes tracked paths by default). So this PR drops two committed files that git keeps. This also **corrects #6931's own ground truth** — that issue cites the pair as `check-ignore`-positive, and they are not. The entity cost here is 0 (both are `vendor_testdata`-skipped, which is why no test caught it) and it does not threaten arm A, whose requirement is that grafel's set be a *subset* of git's. But it is an over-exclusion of committed content, and the fix — an index read vetoing the ignore decision — is a design change I did not make unilaterally. Say the word and I will do it in this PR or a follow-up.

## Answers the brief asked for either way

- **`ShouldSkipPathForRepo` does apply the root `.gitignore` to files** (it walks every prefix *including the full path*), so the walker and watcher disagreed before this PR, with the watcher stricter — which is why it never surfaced as churn. `TestWalkerAndEventBoundaryAgreeOnIgnoredFiles_6931` pins the agreement; M1 kills it. Residual watcher divergences filed as **#6934**.
- **The watcher's directory-subscription path has no file/directory asymmetry** — `subscribeRepo`'s `WalkDir` callback returns early on `!d.IsDir()` (`watcher.go:707`). Nothing to fix, nothing to file.
- **Incidental for arm A**: `WalkRepo` rooted below the repo root does **not** inherit the top-level `.gitignore`, and the one inheriting rewrite (`rewriteInheritedIgnoreLine`) **drops anchoring** — measured, `/sub/x.go` under relRoot `sub` becomes the un-anchored `x.go`. The exclude sources are therefore applied only at the git top-level, documented and pinned by `TestExcludeSourcesAreNotAppliedBelowTopLevel_6922` (which also asserts the same tree walked from the top-level *does* apply them), killed by M11.
- **Not established, unchanged:** `.grafelignore` interaction with the two new sources beyond stack ordering, sparse checkout, submodules, symlinks, any corpus count beyond this repo.

Closes #6931
Closes #6922
Refs #6932, #6898, #6902, #6338, #6416, #6934, #6936, #6937

### DIFF
--- a/internal/daemon/walk/gitexclude.go
+++ b/internal/daemon/walk/gitexclude.go
@@ -1,0 +1,173 @@
+// Package walk — gitexclude.go
+//
+// #6922: git reads THREE ignore sources, not one.
+//
+//  1. .gitignore files in the tree      (deepest first — highest precedence)
+//  2. .git/info/exclude                 (per-clone, uncommitted)
+//  3. core.excludesFile                 (per-user global — lowest precedence)
+//
+// The walker read only the first, so a file excluded via .git/info/exclude was
+// indexed by grafel and reported by `git status` at no -u level. This file
+// resolves sources 2 and 3 and hands them to the walker as ordinary IgnoreFile
+// values, so they go through the same parser, the same precedence stack and the
+// same SkipEntry reporting as .gitignore.
+package walk
+
+import (
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/cajasmota/grafel/internal/gitmeta"
+)
+
+// gitExcludeSourceInfo is the label used for .git/info/exclude in SkipEntry
+// rules; gitExcludeSourceGlobal is the label for core.excludesFile.
+const (
+	gitExcludeSourceInfo   = "info/exclude"
+	gitExcludeSourceGlobal = "core.excludesFile"
+)
+
+// gitExcludeIgnoreFiles returns the non-.gitignore ignore sources for the repo
+// containing root, ordered LOWEST precedence first — core.excludesFile, then
+// .git/info/exclude — so a caller can Push them onto an IgnoreStack before the
+// tree's own .gitignore files, which must outrank both.
+//
+// Entries that do not exist, are empty, or contain no patterns are omitted, so
+// a non-git directory yields nil and costs two bounded git calls.
+//
+// Patterns in both files are anchored at the git TOP-LEVEL. When root is a
+// subdirectory of the top-level (a monorepo registered by package root) the
+// patterns are rewritten relative to root by the same helper that handles
+// inherited .grafelignore files, so an anchored `/build` in info/exclude does
+// not silently become "anything named build" under the child root.
+// gitExcludeIgnoreFiles returns the non-.gitignore ignore sources for the repo
+// at root, ordered LOWEST precedence first — core.excludesFile, then
+// .git/info/exclude — so a caller can Push them onto an IgnoreStack before the
+// tree's own .gitignore files, which must outrank both.
+//
+// Entries that do not exist, are empty, or contain no patterns are omitted, so
+// a non-git directory yields nil and costs one bounded git call.
+//
+// SCOPE, deliberately: patterns in both files are anchored at the git
+// TOP-LEVEL, so they are applied ONLY when root IS the top-level. A walk rooted
+// at a subdirectory (a monorepo registered by package root) gets neither — the
+// same treatment root .gitignore already gets, which is read from root and not
+// inherited from the top-level either. The one inheriting layer, .grafelignore,
+// rewrites its patterns through rewriteInheritedIgnoreLine, and that rewrite
+// DROPS anchoring: `/sub/x.go` under relRoot `sub` becomes the un-anchored
+// `x.go`, which then matches at any depth. Reusing it here would trade an
+// under-exclusion for an over-exclusion, so it is not reused. Applying the
+// patterns unrewritten would mis-anchor them just as badly.
+func gitExcludeIgnoreFiles(root string) []*IgnoreFile {
+	// RunGit, not Capture: only the top-level is needed here and Capture costs
+	// five git invocations to answer four questions this function never asks.
+	topLevel := gitmeta.RunGit(root, "rev-parse", "--show-toplevel")
+	if topLevel == "" {
+		return nil // not a git repo (or git unavailable) — nothing to read
+	}
+	if !isSamePathResolved(topLevel, root) {
+		return nil // see SCOPE above
+	}
+
+	var out []*IgnoreFile
+	if ig := parseExcludeFile(globalExcludesPath(root), gitExcludeSourceGlobal); ig != nil {
+		out = append(out, ig)
+	}
+	if ig := parseExcludeFile(infoExcludePath(root), gitExcludeSourceInfo); ig != nil {
+		out = append(out, ig)
+	}
+	return out
+}
+
+// isSamePathResolved reports whether a and b name the same directory once both
+// are made absolute and symlink-resolved (macOS /var → /private/var).
+func isSamePathResolved(a, b string) bool {
+	resolve := func(p string) string {
+		abs, err := filepath.Abs(p)
+		if err != nil {
+			return ""
+		}
+		if r, err := filepath.EvalSymlinks(abs); err == nil {
+			abs = r
+		}
+		return filepath.Clean(abs)
+	}
+	ra, rb := resolve(a), resolve(b)
+	if ra == "" || rb == "" {
+		return false
+	}
+	return samePath(ra, rb)
+}
+
+// parseExcludeFile reads path as gitignore syntax and returns an IgnoreFile
+// anchored at the repo root, or nil when the file is absent, unreadable or pattern-free.
+//
+// readIgnoreFile (not os.ReadFile) because this is a NAME-CHOSEN read that runs
+// before the walk, so the entry-type gate has nothing to say about it — the
+// same hazard #6416 fixed for .grafelignore applies verbatim to a
+// `mkfifo .git/info/exclude`.
+func parseExcludeFile(path, source string) *IgnoreFile {
+	if path == "" {
+		return nil
+	}
+	b, err := readIgnoreFile(path)
+	if err != nil || len(b) == 0 {
+		return nil
+	}
+	ig, err := parseIgnoreReader("", source, strings.NewReader(string(b)))
+	if err != nil || ig == nil || len(ig.patterns) == 0 {
+		return nil
+	}
+	return ig
+}
+
+// infoExcludePath returns the absolute path of <git-common-dir>/info/exclude,
+// or "" when the git dir cannot be resolved.
+//
+// git-common-dir, not git-dir: a linked worktree has its own git dir but shares
+// the main clone's info/exclude, and git honours the shared one.
+func infoExcludePath(root string) string {
+	common := gitmeta.RunGit(root, "rev-parse", "--git-common-dir")
+	if common == "" {
+		return ""
+	}
+	if !filepath.IsAbs(common) {
+		common = filepath.Join(root, common)
+	}
+	return filepath.Join(common, "info", "exclude")
+}
+
+// globalExcludesPath resolves core.excludesFile for the repo at root, falling
+// back to git's own default of $XDG_CONFIG_HOME/git/ignore (then
+// ~/.config/git/ignore) when the config key is unset.
+//
+// It is a CONFIG read, not a fixed path: a developer who points
+// core.excludesFile somewhere else gets that file, which is the whole reason
+// #6922 asks for the config lookup rather than a hardcoded ~/.config path.
+func globalExcludesPath(root string) string {
+	if p := gitmeta.RunGit(root, "config", "--get", "core.excludesFile"); p != "" {
+		return expandTilde(p)
+	}
+	if xdg := strings.TrimSpace(os.Getenv("XDG_CONFIG_HOME")); xdg != "" {
+		return filepath.Join(xdg, "git", "ignore")
+	}
+	home, err := os.UserHomeDir()
+	if err != nil || home == "" {
+		return ""
+	}
+	return filepath.Join(home, ".config", "git", "ignore")
+}
+
+// expandTilde expands a leading "~/" using the current user's home directory,
+// matching git's own expansion of core.excludesFile.
+func expandTilde(p string) string {
+	if p == "~" || strings.HasPrefix(p, "~/") {
+		home, err := os.UserHomeDir()
+		if err != nil || home == "" {
+			return p
+		}
+		return filepath.Join(home, strings.TrimPrefix(strings.TrimPrefix(p, "~"), "/"))
+	}
+	return p
+}
--- a/internal/daemon/walk/gitexclude.go
+++ b/internal/daemon/walk/gitexclude.go
@@ -29,25 +29,15 @@ const (
 )
 
 // gitExcludeIgnoreFiles returns the non-.gitignore ignore sources for the repo
-// containing root, ordered LOWEST precedence first — core.excludesFile, then
-// .git/info/exclude — so a caller can Push them onto an IgnoreStack before the
-// tree's own .gitignore files, which must outrank both.
-//
-// Entries that do not exist, are empty, or contain no patterns are omitted, so
-// a non-git directory yields nil and costs two bounded git calls.
-//
-// Patterns in both files are anchored at the git TOP-LEVEL. When root is a
-// subdirectory of the top-level (a monorepo registered by package root) the
-// patterns are rewritten relative to root by the same helper that handles
-// inherited .grafelignore files, so an anchored `/build` in info/exclude does
-// not silently become "anything named build" under the child root.
-// gitExcludeIgnoreFiles returns the non-.gitignore ignore sources for the repo
 // at root, ordered LOWEST precedence first — core.excludesFile, then
 // .git/info/exclude — so a caller can Push them onto an IgnoreStack before the
 // tree's own .gitignore files, which must outrank both.
 //
-// Entries that do not exist, are empty, or contain no patterns are omitted, so
-// a non-git directory yields nil and costs one bounded git call.
+// Entries that do not exist, are empty, or contain no patterns are omitted. The
+// cost is at most THREE bounded git calls per WalkRepo — `rev-parse
+// --show-toplevel`, `config --get core.excludesFile`, `rev-parse
+// --git-common-dir` — all outside the WalkDir callback, so it is per walk and
+// not per entry. A non-git directory yields nil after the first.
 //
 // SCOPE, deliberately: patterns in both files are anchored at the git
 // TOP-LEVEL, so they are applied ONLY when root IS the top-level. A walk rooted
--- a/internal/daemon/walk/gitexclude_fifo_6416_test.go
+++ b/internal/daemon/walk/gitexclude_fifo_6416_test.go
@@ -1,0 +1,121 @@
+//go:build unix
+
+package walk
+
+import (
+	"bytes"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
+	"syscall"
+	"testing"
+)
+
+// TestIgnoredFIFOIsReportedAsAHazardNotAsIgnored_6931 grades the ORDER of the
+// new file-branch ignore check against the entry-type gate.
+//
+// The walker's comment argues the check "runs AFTER the entry-type gate so a
+// FIFO is still reported as the hazard it is rather than disappearing under an
+// ignore rule". Nothing observed that: both orderings `return nil`, so the walked
+// SET is identical and only the REPORTED REASON differs. Swapping the two blocks
+// left the whole package green — prose asserting what no test observes.
+//
+// A FIFO planted in a source tree is exactly the thing someone has to explain,
+// and `*.pb.go` is exactly the pattern a repo ignores, so the two overlap in
+// practice. Under the swapped order the operator is told the file was ignored by
+// their own .gitignore and never learns a named pipe is sitting in their tree.
+//
+// VARIED: the entry type only. HELD CONSTANT: the name `ignored.pb.go` matches
+// the ignore rule in BOTH rows, so the control cannot pass by name.
+func TestIgnoredFIFOIsReportedAsAHazardNotAsIgnored_6931(t *testing.T) {
+	root := t.TempDir()
+	if err := os.WriteFile(filepath.Join(root, ".gitignore"), []byte("*.pb.go\n"), 0o644); err != nil {
+		t.Fatal(err)
+	}
+	// Row 1: a FIFO whose name matches the ignore rule.
+	mkfifoInWalkTemp(t, root, "ignored.pb.go")
+	// Row 2: a REGULAR file whose name matches the same rule.
+	if err := os.WriteFile(filepath.Join(root, "regular.pb.go"), []byte("package p\n"), 0o644); err != nil {
+		t.Fatal(err)
+	}
+
+	var files []string
+	var skipped []SkipEntry
+	mustReturnWalk(t, "WalkRepo over an ignored FIFO", func() {
+		files, skipped, _ = WalkRepo(root, nil)
+	})
+
+	rules := map[string]string{}
+	for _, s := range skipped {
+		rel, _ := filepath.Rel(root, s.AbsPath)
+		rules[filepath.ToSlash(rel)] = s.Rule
+	}
+	if got := rules["ignored.pb.go"]; !strings.HasPrefix(got, "irregular:") {
+		t.Errorf("FIFO ignored.pb.go reported with rule %q, want an irregular: hazard — "+
+			"the ignore layer must not swallow the entry-type gate's reason", got)
+	}
+	if got := rules["regular.pb.go"]; !strings.Contains(got, ".gitignore") {
+		t.Errorf("regular.pb.go reported with rule %q, want the .gitignore rule", got)
+	}
+	// Both are still omitted from the walked set — only the reason differs.
+	for _, f := range files {
+		if f == "ignored.pb.go" || f == "regular.pb.go" {
+			t.Errorf("%s should not be walked", f)
+		}
+	}
+}
+
+// TestInfoExcludeFIFODoesNotHang_6416 grades the reason parseExcludeFile reads
+// through readIgnoreFile rather than os.ReadFile.
+//
+// .git/info/exclude is a NAME-CHOSEN read that runs BEFORE the walk, so the
+// entry-type gate #6468 added has nothing to say about it — the #6416 hazard
+// verbatim. os.ReadFile on a FIFO does not fail, it parks in open(2) waiting
+// for a writer: no timeout, no error, no log line, and `grafel index` never
+// returns. Both the PR body and the doc comment named that hazard and no test
+// observed it, so the mutant survived.
+//
+// It is a hang, so the pin is a deadline plus the always-on report — the
+// mutant produces no value to assert on.
+func TestInfoExcludeFIFODoesNotHang_6416(t *testing.T) {
+	if _, err := exec.LookPath("git"); err != nil {
+		t.Skip("git not on PATH")
+	}
+	hermeticGitEnv(t)
+	root := t.TempDir()
+	mustGit(t, root, "init", "-q")
+
+	// git init writes a regular .git/info/exclude; replace it with a FIFO.
+	excl := filepath.Join(root, ".git", "info", "exclude")
+	if err := os.Remove(excl); err != nil && !os.IsNotExist(err) {
+		t.Fatalf("remove %s: %v", excl, err)
+	}
+	if err := os.MkdirAll(filepath.Dir(excl), 0o755); err != nil {
+		t.Fatal(err)
+	}
+	if err := syscall.Mkfifo(excl, 0o600); err != nil {
+		t.Fatalf("mkfifo %s: %v", excl, err)
+	}
+	t.Cleanup(func() { _ = os.Remove(excl) })
+
+	if err := os.WriteFile(filepath.Join(root, "keep.go"), []byte("package p\n"), 0o644); err != nil {
+		t.Fatal(err)
+	}
+
+	var buf bytes.Buffer
+	restore := setIgnoreSkipOutput(&buf)
+	defer restore()
+
+	var files []string
+	mustReturnWalk(t, "WalkRepo with a FIFO .git/info/exclude", func() {
+		files, _, _ = WalkRepo(root, nil)
+	})
+
+	if len(files) == 0 {
+		t.Fatal("walk produced nothing; the refused exclude file must not stop the walk")
+	}
+	if !strings.Contains(buf.String(), excl) {
+		t.Errorf("refused .git/info/exclude was not reported; got %q", buf.String())
+	}
+}
--- a/internal/daemon/walk/gitignore.go
+++ b/internal/daemon/walk/gitignore.go
@@ -192,6 +192,18 @@ func ParseIgnoreFile(dir, path, source string) (*IgnoreFile, error) {
 // Returns (skip=true, lineNum) when the last matching pattern is a
 // positive (non-negated) rule.
 func (ig *IgnoreFile) MatchDir(relPath string) (bool, int) {
+	return ig.MatchPath(relPath, true)
+}
+
+// MatchPath is MatchDir generalised to an entry that may be a FILE.
+//
+// isDir is load-bearing for exactly one pattern form: a trailing-slash rule
+// (`build/`) matches a DIRECTORY named build and never a file named build.
+// Every other form applies to both kinds. Before #6931 the walker only ever
+// asked about directories, so dirOnly was parsed and never consulted; once the
+// file branch consults the stack it becomes the difference between git's answer
+// and an over-broad one.
+func (ig *IgnoreFile) MatchPath(relPath string, isDir bool) (bool, int) {
 	if len(ig.patterns) == 0 {
 		return false, 0
 	}
@@ -216,6 +228,9 @@ func (ig *IgnoreFile) MatchDir(relPath string) (bool, int) {
 	matchedLine := 0
 
 	for _, pat := range ig.patterns {
+		if pat.dirOnly && !isDir {
+			continue
+		}
 		var ok bool
 		if pat.anchored {
 			// Anchored: match against path-from-dir using full path match.
@@ -345,11 +360,25 @@ func (s *IgnoreStack) Pop() {
 // pattern in a child IgnoreFile overrides a positive match in a parent.
 // Returns (skip=true, "source:lineN") when the final decision is to skip.
 func (s *IgnoreStack) Match(relPath string) (bool, string) {
+	return s.MatchPath(relPath, true)
+}
+
+// MatchFile is Match for a regular file. It exists so the walker's file branch
+// reads as deliberately as its directory branch — before #6931 the file branch
+// consulted no ignore layer at all, so `*.pb.go` in .gitignore was matched by
+// `git check-ignore` and indexed by grafel.
+func (s *IgnoreStack) MatchFile(relPath string) (bool, string) {
+	return s.MatchPath(relPath, false)
+}
+
+// MatchPath is Match with an explicit entry kind. See IgnoreFile.MatchPath for
+// why the kind matters (trailing-slash, directory-only patterns).
+func (s *IgnoreStack) MatchPath(relPath string, isDir bool) (bool, string) {
 	// Walk the stack from bottom (root) to top (most-nested).
 	matched := false
 	rule := ""
 	for _, ig := range s.files {
-		ok, line := ig.MatchDir(relPath)
+		ok, line := ig.MatchPath(relPath, isDir)
 		if ok {
 			matched = true
 			rule = ruleLabel(ig.Source, line)
--- a/internal/daemon/walk/gitignore_files_6931_test.go
+++ b/internal/daemon/walk/gitignore_files_6931_test.go
@@ -1,0 +1,350 @@
+package walk
+
+import (
+	"os"
+	"os/exec"
+	"path/filepath"
+	"sort"
+	"strings"
+	"testing"
+)
+
+// --- #6931 / #6922 -----------------------------------------------------------
+//
+// WalkRepo applied the ignore stack in the d.IsDir() branch only, so a
+// file-level .gitignore pattern (`*.pb.go`, `graph.json`, ...) was matched by
+// `git check-ignore` and INDEXED by grafel (#6931). The other two ignore
+// sources git reads — `.git/info/exclude` and `core.excludesFile` — were read
+// by nothing at all (#6922).
+//
+// The fixture below enumerates the pattern space rather than hand-picking a
+// few cases, and asserts the walked set EXACTLY (set equality, not "contains"),
+// so over-exclusion is graded as loudly as under-exclusion (#6902).
+//
+// VARIED, one axis per row:
+//   - pattern form: bare name, rooted `/name`, glob, negation `!`,
+//     trailing-slash dir-only
+//   - pattern SOURCE: root .gitignore, nested .gitignore, .git/info/exclude,
+//     core.excludesFile
+//   - entry type for one identical name (`dironly.go` as a file vs as a dir)
+//   - depth of the matched path (repo root vs nested) for each of the bare,
+//     rooted and glob forms
+//   - precedence between adjacent sources, in BOTH directions
+//
+// HELD CONSTANT: file extension (.go) for every row so the extension filter
+// can never be the thing doing the work; no sparse checkout; no .grafelignore;
+// a single repo root that IS the git top-level.
+
+// writeFile writes p (relative to root) with trivial content.
+func writeIgnoreFixtureFile(t *testing.T, root, p, content string) {
+	t.Helper()
+	abs := filepath.Join(root, filepath.FromSlash(p))
+	if err := os.MkdirAll(filepath.Dir(abs), 0o755); err != nil {
+		t.Fatalf("mkdir %s: %v", abs, err)
+	}
+	if err := os.WriteFile(abs, []byte(content), 0o644); err != nil {
+		t.Fatalf("write %s: %v", abs, err)
+	}
+}
+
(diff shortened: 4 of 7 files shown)


cajasmota:
Coordinator pass, part 1 — from reading the source, before any compile. Two items, one certain and one a question.

## 1. `gitexclude.go` carries two stacked doc comments, and the first contradicts the code

```
31: // gitExcludeIgnoreFiles returns the non-.gitignore ignore sources for the repo
44: // gitExcludeIgnoreFiles returns the non-.gitignore ignore sources for the repo
62: func gitExcludeIgnoreFiles(root string) []*IgnoreFile {
```

Lines 31–43 are an orphaned earlier draft. Only 44–61 is the real doc comment. They do not merely
overlap — they assert **opposite** behaviour:

- **31–43:** *"the patterns are rewritten relative to root by the same helper that handles inherited
  `.grafelignore` files, so an anchored `/build` in info/exclude does not silently become 'anything
  named build' under the child root"* — and *"costs two bounded git calls"*.
- **44–61 (correct):** the rewrite is **deliberately not reused**, because
  `rewriteInheritedIgnoreLine` **drops anchoring**; a sub-root walk gets **neither** source — and
  *"costs one bounded git call"*.

The stale block describes the design you rejected, in the affirmative, immediately above the block
explaining why you rejected it. Whoever reads this next is told the subdirectory case is handled.
Delete 31–43.

This is worth more than a nit here because the second block is the one piece of prose arm A is
being sent to read (it is the only written record that `WalkRepo` rooted below the top-level does
not inherit the top-level `.gitignore`, and that the one inheriting rewrite drops anchoring). A
reader who stops at the first block gets the opposite answer to the exact question they came with.

## 2. Unverified question — is the walk now machine-dependent, and is the suite still hermetic?

`globalExcludesPath` falls back to `$XDG_CONFIG_HOME/git/ignore`, then `~/.config/git/ignore`. That
is right — it is what git does, and it is the point of #6922. But it makes `WalkRepo`'s output a
function of the *developer's own global git config*, which nothing in the walk controls.

Two consequences I have **not** yet measured and am not asserting:

- **Test hermeticity.** If `internal/daemon/walk`'s new tests do not pin `XDG_CONFIG_HOME` (or
  `HOME`) to a scratch dir, a contributor with `*.go` or `*.json` in their global ignore file gets a
  different walked set than CI. Green here would not mean green there.
- **Ratchet stability.** The extraction-quality gate counts entities from a walked set. If the
  global file is consulted during a ratchet run, `baseline.json` becomes machine-relative — and a
  *shrink* passes that gate silently (#6898), so it would not announce itself.

Both may already be handled. If they are, say where; if the fixture pins the env, that pin is itself
worth a line of comment, because the failure it prevents is invisible on the machine that has no
global ignore file.

I will score my own mutants against the two paths nobody has: the XDG/home fallback and
`expandTilde`. Refs #6922, #6931, #6898.

cajasmota:
## Independent adversarial review — REQUEST CHANGES

Own checkout of `51c3f902d` (`git worktree add --detach`), never the author's. Baseline
`go test -count=1 ./internal/daemon/walk/ ./internal/daemon/watch/` green before any mutation;
all three touched files restored to their original shasums afterwards and `git status --porcelain`
clean.

The fix is the right fix and the seam is the right seam. Three things block it: one **new
over-exclusion regression** that the fixture misses by lexical luck, two **false claims in the PR
body** that I measured directly, and one **environment-dependent test**. Plus two live mutants.

---

### BLOCKER 1 — a nested `.gitignore` now deletes files it has no authority over

`WalkRepo` pops nested ignore files **only inside the `d.IsDir()` branch**
(`walker.go`, the `depthEntries` loop). `filepath.WalkDir` is DFS in lexical order, so after a
subdirectory's subtree completes, every **file** visited before the next *directory* entry is still
matched against that subdirectory's `.gitignore`. Until this PR that stale entry could only decide
directories — which pop first — so it was invisible. The file branch makes it decide files.

Demonstrated in my checkout (`TestProbe_StaleNestedIgnoreLeaksToSiblingFiles`):

```
tree:  pkg/sub/.gitignore  -> "victim.go"
       pkg/sub/inside.go
       pkg/victim.go
       victim.go            <- REPO ROOT

WalkRepo walked: [pkg/sub/.gitignore pkg/sub/inside.go]
skipped: pkg/victim.go  rule=.gitignore line 1
skipped: victim.go      rule=.gitignore line 1     <- root file, killed by a rule two levels down
git check-ignore pkg/victim.go ignored=false
git check-ignore victim.go     ignored=false
```

`"sub" < "victim.go"`, so no directory entry intervenes and no `Pop` runs. Both files are dropped
from the index and attributed to a rule that does not govern them.

The fixture misses this by ordering alone: in `ignoreSourceFixture`, `nested_only.go` sorts before
`pkg`, `pkg/sub` is the last entry in `pkg`, and the root files that *do* follow `pkg`
(`prec_a.go`, `prec_b.go`, `rooted_ignored.go`) happen to match neither rule in
`pkg/sub/.gitignore`. One file named `zz_nested_only.go`… actually one file named `nested_only.go`
placed after `pkg` would have caught it. The exact-set assertion is the right shape; its population
is one row short.

This is production-reachable in any repo with a nested `.gitignore` whose parent has later-sorting
files — `frontend/.gitignore: dist`, `testdata/.gitignore: *.out`, and so on. It is the
over-permissive direction, which is the direction this milestone keeps losing in.

Fix: pop by comparing the entry's *parent* directory before matching, or (cheaper) do the pop
bookkeeping for both entry kinds rather than only in the `IsDir` branch.

### BLOCKER 2 — claims 5 and 6 are false on the real repo. Measured, not argued.

I re-ran the enumerated shrink against `/Users/jorgecajas/Projects/archigraph` (the main clone, not
a linked worktree), using this branch's `WalkRepo`, with and without
`gitExcludeIgnoreFiles` neutered:

| build | walked |
|---|---|
| this branch, `gitExcludeIgnoreFiles` returning nil | 7084 |
| this branch as-is | **7079** |

The complete delta attributable to **#6922 alone** is five files, not zero:

```
presentation-kit/01-what-is-grafel.md
presentation-kit/02-how-it-works.md
presentation-kit/03-grep-vs-grafel.md
presentation-kit/04-demo-walkthrough.md
presentation-kit/05-talking-points-and-faq.md
```

`presentation-kit/` is in this clone's `.git/info/exclude` and **not** in `.gitignore` — it is not
"already covered". So:

- **Claim 6 ("#6922's own contribution on this repo is ZERO files") is FALSE** — it is 5 files.
- **Claim 5 ("2 of 7084 walked files… nothing else removed") is FALSE** — the total shrink vs `main`
  is 2 (`graph.json`) + 5 (`presentation-kit/`) = **7 files**.

The likely cause is that the measurement ran inside a linked worktree, where the untracked
`presentation-kit/`, `.claude/hooks/`, `.claude/skills/`, `.claude/dev/` simply do not exist. All
four are present in the main clone. This is a measurement-environment error, not a behaviour bug —
the new exclusions are *correct* — but "entity delta 0" now needs re-deriving: five markdown files
leaving the index is not obviously a zero-entity change, and the `cmd/grafel` digest pin was taken
in the same environment that produced the wrong file count.

### BLOCKER 3 — `TestGlobalExcludesPath_ConfigThenXDGFallback_6922` is red on a common dev machine

The test asserts that with `core.excludesFile` unset the path falls back to
`$XDG_CONFIG_HOME/git/ignore`. It sets `XDG_CONFIG_HOME` but does nothing about `~/.gitconfig`, so
`git config --get core.excludesFile` still reads the developer's real global config. Git's own
documentation recommends exactly that setting.

```
$ HOME=<fake home with [core] excludesFile=…> go test -count=1 -run TestGlobalExcludesPath ./internal/daemon/walk/
--- FAIL: TestGlobalExcludesPath_ConfigThenXDGFallback_6922
```

(Green on this machine only because `core.excludesFile` happens to be unset here — verified
`git config --global --get core.excludesFile` → rc=1.)

Also note the PR body's "the test never reads or writes the developer's real global git config" is
true of `ignoreSourceFixture` only. `TestExcludeSourcesAreNotAppliedBelowTopLevel_6922` and
`TestInfoExcludeIsSharedWithLinkedWorktrees_6922` set no local `core.excludesFile` either, so they
also read the real one — they just tolerate it today.

Fix: `t.Setenv("GIT_CONFIG_GLOBAL", filepath.Join(t.TempDir(), "gitconfig"))` and
`t.Setenv("GIT_CONFIG_SYSTEM", os.DevNull)` in `mustGit`'s repos.

---

### My mutants — 5, at the call site, none of them the author's

Applied by exact-occurrence assertion in Python (`assert s.count(old)==1`), `go vet` clean before
each score, `-count=1`, restored by `cp` from a shasum-verified backup.

|---|---|---|---|
| R1 | `isSamePathResolved`: delete the `filepath.EvalSymlinks` call | **DEAD** | 4 tests fail (`AppliesIgnoreStackToFiles`, `IgnoredFilesAreReported`, `InfoExcludeIsSharedWithLinkedWorktrees`, `ExcludeSourcesAreNotAppliedBelowTopLevel`) — `t.TempDir()` under `/var` → `/private/var` |
| R2 | walker: swap the new ignore block **above** the entry-type gate | **ALIVE — distinguishable and ungraded** | walk `ok`, watch `ok` |
| R3 | `parseExcludeFile`: `readIgnoreFile` → `os.ReadFile` | **ALIVE — distinguishable and ungraded** | walk `ok` |
| R4 | `IgnoreStack.MatchPath`: delete the cross-file negation reset (`else if line != 0`) | **DEAD** | `AppliesIgnoreStackToFiles` fails |
| R5 | `infoExcludePath`: `filepath.Join(root, common)` unconditionally, dropping the `IsAbs` guard | **DEAD** | `InfoExcludeIsSharedWithLinkedWorktrees` fails |

**R2 (ordering).** The code comment states the check "runs AFTER the entry-type gate so a FIFO is
still reported as the hazard it is rather than disappearing under an ignore rule". Nothing observes
that. Distinguishing input: `mkfifo ignored.pb.go` with `*.pb.go` in `.gitignore` — head build
reports `Rule: irregular…`, mutant reports `Rule: .gitignore line 1`. Both `return nil`, so only
the *reported reason* differs, which is precisely the #6338 property the comment invokes. Grade it
with one `SkipEntry.Rule` assertion on a FIFO whose name matches an ignore pattern.

**R3 (`#6416` hazard).** Both the PR body and the doc comment on `parseExcludeFile` justify
`readIgnoreFile` by name — "`mkfifo .git/info/exclude` is the #6416 hazard verbatim". No test
observes it, and the `maxIgnoreFileBytes` cap goes with it. Distinguishing input:
`mkfifo .git/info/exclude` before `WalkRepo` — the mutant blocks in `open(2)`; head refuses it and
reports through `reportIgnoreSkip`. This is the "prose asserts what no test observes" shape: one
test, and the justification stops being an argument.

---

### Smaller, non-blocking

**`gitExcludeIgnoreFiles` carries two stacked doc comments, and the first one is stale and wrong.**
Lines 26–39 of `gitexclude.go` say the patterns "are rewritten relative to root by the same helper
that handles inherited `.grafelignore` files" — the second block (and the code) say the exact
opposite, that reusing that helper would trade an under-exclusion for an over-exclusion and is
therefore *not* done. The first block also says "two bounded git calls", the second says "one", and
the actual cost is **three** (`rev-parse --show-toplevel`, `config --get core.excludesFile`,
`rev-parse --git-common-dir`). Delete the first block; fix the count in the second.

**The pattern space is not fully enumerated.** `IgnoreFile.Dir` is dead —
`filepath.Rel(ig.Dir, filepath.Join(ig.Dir, relPath))` is `relPath` for every input — so a nested
`.gitignore` is matched against the **root**-relative path. Anchored and slash-bearing nested rules
therefore never fire (`TestProbe_NestedAnchoredPatternNeverFires`: `pkg/.gitignore` containing
`/anchored.go` and `subdir/deep.go` excludes neither). Pre-existing, and an *under*-exclusion, so
not a blocker for this PR — but the fixture's nested rows are bare names only, so "the pattern
space is enumerated, not hand-picked" overstates what is graded. Two missing axes: nested × anchored,
nested × glob. Worth a follow-up issue alongside the residual watcher divergences you already listed.

I did check the forms you did enumerate through the file branch — character class `[abc]bad.go`,
`**/deep/*.go`, `tmp?.go` — and all three behave correctly, including the `dbad.go` control.

### Claims I checked and found TRUE

- **2 / dirOnly.** `MatchDir` → `MatchPath(rel, true)` and the only new branch is
  `pat.dirOnly && !isDir`, which is unreachable when `isDir` is true. Directory behaviour is
  genuinely unchanged, in both directions.
- **3.** `dirOnly` was parsed in `parseIgnoreReader` and read nowhere before this diff.
- **4.** Independently checked `internal/classifier/classifier.go`: the generic `.json` branch
  requires `openapi.json`/`swagger.json`/`*.openapi.json`/`*.swagger.json`, `ocelot*`, the
  `cdc/`/`debezium/`/`kafka-connect/` anchors, or one of the `*.schema.json`/`*.tofu.json`/
  `elm.json`/`rescript.json`/`bsconfig.json` special cases. A bare `graph.json` matches none.
- **7.** `ShouldSkipPathForRepo` does walk every prefix *including the full path*, so the walker and
  watcher did disagree with the watcher stricter.
- **8.** `watcher.go:707` is `if !d.IsDir() { return nil }` in `subscribeRepo`'s `WalkDir` callback.
  No file branch exists there.
- **Cost.** Three `gitmeta.RunGit` calls per `WalkRepo`, all outside the `WalkDir` callback. Per
  walk, not per entry — fine for a performance arm.

---

**Verdict: REQUEST CHANGES.** Blocker 1 is a live over-exclusion that silently drops indexed files;
blockers 2 and 3 are a false measurement and a test that is red on a normal developer machine.
R2 and R3 should be graded before merge, since both are behaviours the PR argues for in prose.

cajasmota:
Coordinator mutants, part 2 — four scored against `gitexclude.go`, the two paths nobody had touched
(the XDG/home fallback and the tilde expansion) plus an environment probe. **One is ALIVE and one is
a real test defect.** Tree verified pristine after every restore (`cp` from a shasum-verified backup,
`6bebec1f4…`, matched byte-for-byte three times; `git status --porcelain` empty).

| # | mutant | verdict |
|---|---|---|
| MC-1 | `globalExcludesPath` — delete the XDG **and** home fallback, so an unset config key yields `""` | **DEAD** — `TestGlobalExcludesPath_ConfigThenXDGFallback_6922` |
| MC-2 | `expandTilde` body → `return p` (the **helper**) | **DEAD** — `TestExpandTilde_6922` |
| **MC-2b** | **the CALL SITE: `return expandTilde(p)` → `return p`, helper left intact** | **ALIVE — whole `./internal/daemon/walk/` package green** |
| MC-3 | no code edit: run the package with `XDG_CONFIG_HOME` pointing at a dir whose `git/ignore` holds `*.go` / `*.json` | **suite goes RED** |

## MC-2b — distinguishable and ungraded

Pre-edit anchor count 1, post-edit call-site count 0, `func expandTilde` still present (1), `go vet`
clean, `-count=1`. So the edit landed and the package is genuinely green without it.

`expandTilde` is graded **only as a helper**. The integration fixture sets `core.excludesFile` to an
absolute path (`gitignore_files_6931_test.go:99`), so no test ever drives a `~/`-prefixed value
through `globalExcludesPath`. Deleting the only call site is invisible.

This is the #6533 shape verbatim: extracting a predicate and unit-testing it does not pin the
behaviour — the call site stays free to ignore it. And it is production-reachable by the most
ordinary setting there is: `core.excludesFile = ~/.gitignore_global` is the spelling git's own
documentation uses. Under MC-2b that user's entire global ignore layer silently no-ops.

The sharpest part is that **your own test comment already states this exact failure** (`:334-335`:
*"Without it a configured `~/.gitignore_global` is opened as a literal directory named `~` and the
whole layer silently no-ops"*). The consequence is asserted in prose and graded only on the helper —
the dominant defect class on this board.

**Grade it at the call site.** Cheapest form: in the integration fixture write the global ignore
file under a temp home, `t.Setenv("HOME", tmp)` (and `USERPROFILE` for Windows), configure
`core.excludesFile` as `~/<name>` rather than the absolute path, and assert the excluded file is
still excluded. That kills MC-2b and MC-2 together and costs a few lines.

## MC-3 — the suite is not hermetic, and it fails, not merely diverges

With `XDG_CONFIG_HOME` set to a controlled dir containing `git/ignore` = `*.go`, `*.json`:

```
--- FAIL: TestInfoExcludeIsSharedWithLinkedWorktrees_6922 (0.12s)
    gitignore_files_6931_test.go:244: git commit -qm seed: exit status 1
        nothing to commit (create/copy files and use "git add" to track)
```

The fixture's own `git add` is subject to the developer's global ignore, so the seed commit is
empty and the test dies before it tests anything. This machine has `core.excludesFile` unset and no
`XDG_CONFIG_HOME`, which is the only reason it is green here — a contributor with the very
`~/.gitignore_global` this feature exists to support gets a red suite and no clue why.

Note the two findings pull in opposite directions and both need fixing: MC-2b says the tilde path is
**under**-exercised, MC-3 says the fixture is **over**-exposed to the real environment. The answer to
both is the same — own the environment inside the test rather than inheriting it.

I am not asserting anything about the ratchet: it runs in CI where the global file is absent. Worth
a thought only if a ratchet run ever executes on a developer machine.

MC-1 is DEAD and worth saying so plainly: the fallback chain is properly graded, including the
XDG-before-home ordering. Refs #6922, #6931, #6533.

cajasmota:
Coordinator verification of `af23b5560`. All seven items land. Two checks of my own on the new head,
both on the isolation fix, since that was the item most likely to be cosmetic.

**Verified, not taken:** under a hostile environment (`XDG_CONFIG_HOME` holding `git/ignore` with
`*.go` + `*.json`, and a `GIT_CONFIG_GLOBAL` setting `core.excludesFile`) the package is now **green,
rc=0**. That same environment was red on `51c3f902d`. The claim is true.

**MC-4 — delete the `XDG_CONFIG_HOME` pin from `hermeticGitEnv`, leave everything else:** **DEAD in
both environments** — hostile (2 failures, incl. `TestInfoExcludeIsSharedWithLinkedWorktrees_6922`)
**and clean** (1 failure). Anchor count 1 pre-edit, 0 post-edit, `go vet` clean, `-count=1`, file
restored by `cp` and shasum-matched (`327c0d4cd…`), `git status --porcelain` empty.

The clean-env result is the one that matters and it corrects a concern I was about to raise: I
expected the pin to be graded only by an environment CI does not have, which would have made it a
manual procedure rather than a test. It is not — `TestGlobalExcludesPath_ConfigThenXDGFallback_6922`
derives its expectation from the pinned value, so deleting the pin fails in CI. The isolation is
genuinely self-graded.

## #6937 — my call: correct as filed, fix as a follow-up, not here

You were right not to do it unilaterally, and right that this PR makes the *file* half live (the
directory half predates it — an ignored directory has always been pruned with its tracked contents
inside). I am taking **option 2, both halves, in its own change**, for the reason you gave on #6936:
it must alter `MatchPath`, which is this PR's one independently-verified safety property, and it
needs its own shrink enumeration in *both* directions.

I am not blocking on it, because the measured cost here is 2 files and **0 entities**, and the
direction is the safe one for arm A — a smaller indexed set is still a subset of what git reports,
so the change signal stays complete.

**One thing to add to #6937 before it is picked up:** it prices the directory half as the hard one
("a per-file index veto cannot recover a subtree that was never entered"). That is true of a per-file
veto, but the same single `git ls-files -z` answers the directory question too — insert every
ancestor of every tracked path into a set, and the directory branch simply declines to prune a
directory in it. One call, one pass, both halves, no second data source. The real cost is memory
(~8k paths plus ancestors here), not reachability. Worth writing down so whoever takes it does not
price it as a two-phase job.

**Rejecting a partial fix explicitly:** a file-only veto would leave a tracked `graph.json` indexed
while a tracked file inside an ignored `dist/` stayed dropped. That inconsistent rule is harder to
reason about than today's uniform "the ignore layers are pattern-based and index-unaware", and it
would have to be undone when the directory half lands. Both halves or neither.

Merging once CI is green. Refs #6931, #6922, #6936, #6937, #6932.

Issue #6812, opened 2026-09-05
DECIDED + MEASURED: wire OCaml's vendored tree-sitter grammar — the regex depth walker gets 230 of 408 inherit attributions WRONG (precision 48.3%); fixture and evidence shipped, the wiring is what remains


cajasmota:
OCaml has a vendored tree-sitter grammar that is registered and never used — the extractor is 100% regex, and `inherit` is the case where that costs correctness.

## The finding

`internal/extractors/ocaml/extractor.go` is **610 lines and contains zero occurrences of `TSTree`** (verified by count at `8dc150619`). Every construct is matched with `regexp.MustCompile`.

Meanwhile OCaml **is** wired into the tree-sitter layer:

- `internal/treesitter/adapters.go:51` declares `migratedLanguages`, documented as "maps each supported language to its official ts.Language".
- `:69` — `"ocaml": tsocaml.Language()` — registers the vendored grammar.
- `:22` imports it; `:109` even carries an OCaml smoke sample (`let f x = x + 1`).

So the grammar is vendored, compiled in, and reachable, and the extractor never asks for it.

## Why this matters more for OCaml than for the other regex extractors

This surfaced while grounding #6370, which tracks languages that emit no `EXTENDS`/`IMPLEMENTS`. OCaml is one of the remaining ones, and it is the only entry on that list where a grammar is already available.

The reason it matters is that in OCaml, **`inherit` only means inheritance inside `object … end`**. Outside that context the same keyword is not a class-inheritance edge. That is precisely the kind of context a CST gives you for free, and precisely what a regex has to approximate — the extractor already hand-rolls a depth walker at `:422` to fake structural context.

Unlike groovy, there is no ERROR-node collapse pushing us back toward text scanning.

## Two options, and the tradeoff

**(a) Add `inherit` handling as more regex**, matching the shape of the other seven #6370 languages (erlang, graphql, nim, pony, lisp shipped so far). Small, consistent, and lands with the rest of the batch — but it grows the depth-walker approximation rather than replacing it, in the one place where the approximation is most likely to be wrong.

**(b) Wire the existing grammar** and derive `inherit` from the CST. Bigger diff, and it touches parsing for every OCaml construct rather than just hierarchy, so it needs its own measurement and its own review. But it removes the approximation instead of extending it.

[...]

Issue #6370, opened 2026-08-19
DONE (8/8): erlang, graphql, nim, pony, lisp, clojure, ocaml, razor all ship hierarchy edges — the nine-language silence audit is complete


cajasmota:
#6326 asked, at the end, whether any *other* language is silently emitting no hierarchy edges. It is. This is that audit, filed separately so #6326 stays closed on what it fixed.

A language gets EXTENDS/IMPLEMENTS by one of two paths: its own extractor emits them, or it is registered in `supportedLanguages` (`internal/extractors/cross/hierarchy/extractor.go:1018`) with a matching arm in the dispatch switch (`:1046`). **Nine languages with a real inheritance construct are in neither.** Their graphs have no inheritance topology at all, and nothing reports that — a user asking "what extends this type" gets an empty answer indistinguishable from "nothing does". Same silence class as #6321 and pre-fix F#.

## What was searched, and where

Stated separately from what was concluded, because this repo has shipped repeated enumeration-totality errors and an audit is exactly where that hurts.

Over all 60 directories under `internal/extractors/` (excluding `cross/`, `assembly/`, `testdata/`, `sresolver/`):

1. `grep -rE '"(EXTENDS|IMPLEMENTS)"'` across `*.go`, split non-test vs test-only, classifying each package EMITS / NOEMIT.
2. `grep -roE 'Register\("[a-z0-9_]+"'` per package to recover the **language tokens actually registered** rather than directory names — this matters: `golang`→`go`, `javascript`→`javascript typescript`, `lisp`→`commonlisp racket scheme`, `verilog`→`verilog systemverilog`, `hcl`→`hcl terraform`, `cpp`→`c cpp`, `proto`→`protobuf`. A directory-name comparison would have produced a wrong answer here.
3. Cross-referenced against the 14-entry `supportedLanguages` map.
4. For each gap, grepped the extractor for that language's actual inheritance construct, to separate *never parsed* from *parsed then dropped*.

**Steps 1-3 are exhaustive over the registered extractor set.** Step 4's judgement — "does this language have inheritance worth modelling" — is judgement, not a search, so the table below is ranked rather than claimed total.

## The nine

| Language | Construct | State | Note |
|---|---|---|---|
[...]

PR #6938, opened 2026-09-06
test(#6812): OCaml had no golden fixture at all, and nobody had measured the depth walker — it gets more `inherit` attributions wrong than right

cajasmota:
## What this is

Arm 1 of #6812: the two things that issue lists under **"Not yet established"**, both of which gate the grammar wiring and are needed whichever way it goes. This is **not** the migration — `internal/treesitter/` is untouched, no regex was added to the OCaml extractor, and `internal/extractors/ocaml/extractor.go` is byte-identical to `main` (`git diff` on it is empty).

Two deliverables:

1. **`internal/quality/golden/ocaml-objects-mini/`** — the first golden fixture for OCaml. Until this, nothing about `internal/extractors/ocaml` was graded, so every statement about what the regex path produces was an impression.
2. **`internal/extractors/ocaml/depth_walker_inherit_accuracy_6812_test.go`** — the measurement of how often the hand-rolled depth walker at `extractor.go:422` gets `inherit` context wrong.

---

### The number

Over a clone of **github.com/ocaml/ocaml at `70165ff23e58ebde4e50bae2c1f9948dbe43ee41`**:

| | |
|---|---|
| `*.ml` files in corpus | 2,420 |
| files containing a real `inherit` token | **68** (233 tokens) |
| `object … end` blocks in those files, matched by the reference | 633 |
| blocks the reference could not match (excluded) | 218 |
| **blocks where the walker's block end is wrong** | **479 / 633 = 75.7%** |
| `(block, inherit)` pairs truly enclosed | 216 |
| **`inherit` RECALL** — truly-enclosed pairs inside the walker's span | 197 / 216 = 91.2% (19 missed) |
| **`inherit` PRECISION** — pairs the walker attributes that are right | **197 / 408 = 48.3%** (211 spurious) |

The walker attributes 408 pairs and **230 of them are wrong** — more wrong than right.

Recall and precision are reported separately and never summed. The first version of this printed `230/216 = 106.5%`, which is not a rate: `spurious` is drawn from every `(block, inherit)` pair in the file while `enclosed` counts only the truly-enclosed ones, so the two have different denominators. Fixed in the test as well as here.

### Corpus provenance, stated rather than implied

**The repo vendors no OCaml.** `find internal testdata scripts -name '*.ml'` returns zero, `~/.grafel/corpus` holds none, and there is no OCaml repo in any local group. So the headline figure is taken against an **external** clone, named above, and the test that produces it is **env-gated** (`GRAFEL_OCAML_CORPUS`) and skips without it. It is not silently substituted for an in-repo measurement. The constructed enumeration below runs unconditionally in CI and is what makes the finding reproducible with no network.

### Method

`extractModuleBody` is the depth walker. "Is this `inherit` inside `object … end`" is the same question it answers for `module … struct … end`, so it is driven exactly the way production drives it (`extractor.go:153` hands it a position with the block's opening keyword still ahead, so the opener increments depth) and pointed at `object` instead:

```
walker span = [openerStart, openerStart+len(extractModuleBody(src, openerStart)))
truth span  = [openerStart, offset of the block's real matching `end`)
MIS-ATTRIBUTED = exactly one of the two spans contains the inherit
```

Span ends are compared **whitespace-tolerantly** (`endsAgree`): the walker returns body text and the reference returns the `end` token's offset, so counting that gap would score formatting as a defect. Every difference reported is a real difference in what the block contains.

### The reference is validated — and independently, against a real parser

Ground truth is a small block matcher written for this test. `TestReferenceBlockMatcher_PositiveControls` now pins it against **22** hand-computed answers (14 originally): nested `(* (* … *) … *)` comments, `{|…|}` quoted strings, `"…"` with escapes, char literals vs `'a` type variables, nested and sibling blocks, an unterminated block (which must return −1, not a guess), identifiers *ending in* `end`/`sig`/`object`, and — added after review — identifiers *starting with* `end`/`sig`/`object`/`begin`/`struct`, plus three blocks opened by a 3- or 5-byte keyword.

**The controls alone were not enough, and the review is what established that.** Two defects lived inside the reference:

- A mutant deleting `tokenAt`'s **trailing** identifier-boundary check survived the whole package. Every control tested only the *leading* boundary (`append`, `myobject`, `design`); none had an identifier *starting* with a keyword, so `endian`, `signature`, `objects`, `beginning` and `structure` were all mis-tokenised. Five controls added; the mutant now fails six subtests.
- `refMatchEnd` advanced by a fixed `i += 6` after any opener, with a comment claiming the exact length was irrelevant. It is not: `sig` is 3 bytes and `begin` is 5, so the scan stepped over real source and `module type T = sig end` returned −1 while `sig (* end *) … end` returned the `end` **inside the comment** — defect (3) from the list below, living in the oracle. Invisible to every control because they all opened with the 6-byte `object`. Now `i += len(kw)`, pinned by three controls.

**Neither moved the corpus figures**, which is the strongest available evidence that both were latent rather than load-bearing: 479/633 and 197/408 are byte-identical before and after.

Independently of any of that, the reviewer cross-checked `refMatchEnd` against **the repo's own vendored tree-sitter OCaml grammar** over the same 68 files — a real GLR parser rather than a second approximation: **595 blocks compared, 595 agree, 0 disagreements**, none outside an ERROR subtree.

**The 218 excluded blocks are a positive signal, not a hole.** Every one is in `testsuite/tests/generated-parse-errors/errors.ml` — a generated file of *deliberately malformed* OCaml carrying 256 `object` against 38 `end`. The reference fails exactly where the input is not valid OCaml, and nowhere else. Those blocks are excluded from every figure, so they neither help nor hurt the walker.

### Why it is wrong — four independent defects

1. **The scan never terminates on a balanced block.** The opener takes depth 0→1 and the matching `end` takes it 1→0 *without stopping*, because the loop only breaks on an `end` seen **at** depth 0. A well-formed block therefore always falls through to the indentation fallback, and the depth scan's `found` branch fires mainly when something else has already corrupted the count.
2. **`\b` is evaluated against a slice, not the source.** `openKW`/`closeKW` are re-run on `rest[i:]` and tested for `om[0] == 0`; at offset 0 of a slice, start-of-string satisfies `\b`. So every identifier *ending in* `end`, `sig`, `object`, `struct` or `begin` is read as a keyword — `append`, `send`, `extend`, `backend`, `legend`, `myobject`, `reconstruct`.
3. **OCaml block comments nest; the skipper does not.** It stops at the first `*)`, so `(* (* x *) end *)` leaks an `end`.
4. **`{|…|}` quoted strings are unknown to it** and are scanned as code.

Plus, not a correctness defect but worth knowing before anyone moves this code: the scan runs two regexes **at every byte** of the remaining file, so it is quadratic per block. Scoping the corpus run to the 68 inherit-bearing files is what makes it finish in ~5 minutes.

### Enumerated, not hand-picked — and two of the five axes were still inert

Three rounds of hand-written cases were tried first and **every one agreed with the reference** — a canonically formatted single block is rescued by the indentation fallback, so toy inputs measure the fallback and not the depth scan. Hand-picking would have concluded "no divergence" and been wrong.

`TestDepthWalker_ObjectBlockEnd_EnumeratedSpace` crosses five binary axes exhaustively (32 inputs):

| axis | varied | status |
|---|---|---|
| A | the block's `end` at column 0 vs indented | live |
| B | body contains an identifier ending in `end` | **was inert, now live** |
| C | body contains a nested `(* (* … *) … *)` comment | live |
| D | body contains a `{|…|}` quoted string holding `end` | live |
| E | a further top-level declaration follows the block | **inert, now asserted to be** |

**Axis B graded nothing, and it was the axis that mattered.** It exists to grade defect (2), the slice-relative `\b`, and the generator wrote `method append_all x = x`. **Go's `\b` treats `_` as a word character**, so `\bend\b` never matches inside `end_all` and the hazard never fired. Respelled `append`: the diverging set goes **12/32 → 18/32**, and B is live in both directions. The most interesting row it exposes is `-B---`, which does *not* diverge — the phantom `end` inside `append` takes depth 1→0 early, so the real `end` is then seen at depth 0 and terminates the scan in the right place, defect (2) accidentally cancelling defect (1). Add a comment or a quoted string on top and it stops cancelling and starts truncating.

**Axis E is genuinely inert**, and rather than leave it padding the row count it is now *asserted* inert: the fallback's stop condition is the next column-0 line, so a trailing declaration cannot change where the span ends, and an E that ever starts mattering fails the test instead of quietly widening the table.

So the honest description is **five axes, four of which change outcomes, 18 of 32 combinations diverging** — not the "32 combinations, 12 diverge" the first version implied over two axes that crossed nothing. C and D still diverge only *together*; neither alone is in the set.

### What the enumeration does NOT show

Stated because the first version of this PR claimed the opposite. It said *"a divergence **is** a mis-attribution rather than a proxy for one."* **That is false.** A mutant reducing the criterion to the attribution term alone found *nothing*: **0 of 32 inputs mis-attributes its `inherit`**, because the generator always writes `inherit` as the first body line, so every disagreeing span still contains it. The criterion's attribution term was dead code.

The enumeration measures **block-end divergence** — the walker's structural judgement, and the mechanism an attribution would rest on, but a **proxy**. It is now written that way: the dead term is gone, the 0 is asserted positively by a guard inside the loop, and the log line reads "put the WRONG END on the block … 0/32 mis-attribute the `inherit`".

Actual `inherit` mis-attribution is carried by the other two things: `TestDepthWalker_InheritIsMisAttributed_Constructed` (a sibling pair inside a module, where one block's span swallows the other's `inherit`) and the corpus measurement's 230 pairs. Both are unaffected.

### Is a zero possible here?

No zero was measured, so the "a zero is corpus-relative" caveat does not have to carry this result. Recording what it *would* have meant anyway: a zero over this corpus would have said only that these 68 files happen not to contain the shapes above. The input that distinguishes "cannot be wrong" from "was not wrong here" is any of the 12 enumerated combinations — which is why they are committed as an always-on test rather than left in the corpus run.

---

## Deliverable 1 — the fixture

`ocaml-objects-mini`, three files, on the `pony-traits-mini` template (read alongside `lisp-clos-mini`; pony is the closest precedent because it is the other #6370 arm whose fixture had to grade over-firing as well as recall).

**19 / 19 entities, 6 / 6 relationships, 0 forbidden hits, 0 forbidden entity hits.**

### Axes VARIED

- **`inherit` context** — inside `object … end` at a class declaration (`shapes.ml`); inside a **class type** (`containers.ml`); inside an `object … end` nested in a **`let` body** (`render.ml`). The third is the only context where the keyword is reachable by the scanner today, and therefore the only one gradable in both directions.
- **module-composition keyword** — `open` (an import, and an edge), `include` (not an import and not inheritance), functor `module Make (C : Comparable) = struct`, `module type … = sig`.
- **inheritance arity** — single `inherit`, `inherit … as super`, and multiple inheritance (`tagged_circle` inherits twice).
- **declaration keyword** — `module` vs `module type`; `class` vs `class virtual` vs `class type`.
- **type RHS** — record, record-of-records, variant.
- **identifier hazard** — `backend` and `legend` end at a word boundary, so Go's `\bend\b` re-run against the remaining slice *does* match them; `append_all` does not, because `_` is a word character. Two triggering spellings and one non-triggering one. The first version spelled all three with a trailing `_word`, so the axis it claimed to vary ("three different positions") contained only the non-triggering case — a label claiming more than its content, corrected here and in the fixture's notes. Neither spelling is *graded* by an assertion, because nothing in this fixture asserts a module span and the hazard only reaches `extractModuleBody`; they put both outcomes in the corpus, and that is now what the notes say.
- **presence of the keyword** — `make_logger` (object *with* `inherit`) against `make_counter` (same `let x = object … end` shape, *no* `inherit`).
- **forbidden target dialect** — one forbidden row with a bare target, one with a resolved target (see the trap below).

### Axes HELD CONSTANT

- Language (OCaml only), one fixture, no framework.
- Every graded declaration at column 0 (the extractor's anchor) — indented declarations appear in the source but are deliberately not asserted.
- No `Caqti`/`alcotest` construct, so `depth.go`'s framework pass stays out of the measurement.
- No commented-out or in-string declaration: unlike pony, this fixture claims no string/comment fence, because the extractor has no such guard to grade and a fence nothing can reach would be decoration.

### What is deliberately NOT asserted, so nobody reads absence as approval

**Hierarchy is ungraded, on purpose.** OCaml emits no `EXTENDS`/`IMPLEMENTS` (#6370 group (a)) and #6370's OCaml arm is blocked. The fixture asserts **no hierarchy row in either direction** for the object and class-type forms, so it does not repeat the `erlang-otp-mini` trap of pinning a missing edge as correct. It *does* forbid the shapes that are not inheritance in any OCaml — a module `include`, in both `EXTENDS` and `IMPLEMENTS` spellings — because those stay forbidden after #6370's arm lands.

Known divergences present in the source and left unasserted:

- The **functor** `module Make (C : Comparable) = struct` is not extracted at all — `moduleRE` requires `=`/`:` immediately after the name and a parameter list intervenes.
- No `class` / `class type` / `method` / `val` construct produces any entity, so `shape`, `circle`, `square`, `tagged_circle`, `printer`, `verbose_printer` and `stream_writer` are invisible.
- Indented `let`s inside a module body are not extracted, which also means module `CONTAINS` is empty here.
- Every module in `containers.ml` reports the **same** end line — the depth-walker defect above, visible in the fixture itself.

These are filed rather than fixed, per the brief.

### The row the fixture exists for

```
make_logger --[CALLS]--> stream_writer
```

`inherit stream_writer prefix` inside `make_logger`'s object body. **The parent name survives extraction today** — but only as a `CALLS` edge from the enclosing function, because the extractor has no class vocabulary and `inherit` itself is keyword-suppressed. Recording it as `CALLS` is not an endorsement; it is the enumeration of *what the regex path currently produces* that the owner's #6812 comment requires **before** a grammar swap replaces it, and it is the evidence of what that swap must not silently drop.

---

## Mutant verdicts

Every mutant is applied with an asserted **exact match count of 1**, checked to have landed with `diff`, rebuilt, re-graded, and restored **from a backup file** — never `git checkout`, which silently discarded uncommitted work once during this task. `extractor.go` is byte-identical to `main`.

### Round 1 — the fixture's own assertions

|---|---|---|---|
| M1 | delete `"inherit": true` from `ocamlKeywords` | **DEAD** | forbidden hit `make_logger --[CALLS]--> inherit`, exit 2 |
| M2 | delete the `break` in `extractLetBody` so a let body runs to EOF | **DEAD** | 2 forbidden hits: `→ bump`, `→ backend`, exit 2 |
| M3 | `callBareRE` `\s+` → `\s\s+` | **DEAD** | relationship recall 6/6 → 3/6 |
| M4 | `openRE` `^open\s+` → `^open\s\s+` | **DEAD** | relationship recall 6/6 → 4/6 |
| M5 | `moduleRE` `^module` → `^ module` | **DEAD** | entity recall 19/19 → 16/19 |
| M6 | `typeRE` `^type` → `^ type` | **DEAD** | entity recall 19/19 → 16/19 |

All six re-scored after the fixture changed; all six still DEAD.

### Round 2 — fence FIREABILITY, one mutant per row

The review's blocking finding (N5) was that two fences were **silently unfireable**: `Extended EXTENDS Base` and its `IMPLEMENTS` twin used `to_bare_name` for a name that always resolves, so a mutant genuinely emitting the edge left them green. The stated reason — *"OCaml's edge vocabulary contains no hierarchy kind at all, so no mutant can emit an EXTENDS"* — was an argument, and it was wrong: the cause was the **target dialect**, and on the day #6370's OCaml arm emits that edge both fences would have passed.

This is the same trap found earlier on `backend_label`, about which the first version of this PR wrote "assume there are others". There were two more, in the same file, and reading did not find them. **So every remaining forbidden row was re-audited by firing it**, not by reasoning about it:

| # | mutant (injected into `extractOCaml`) | row | verdict |
|---|---|---|---|
| N5a | emit `Extended --[EXTENDS]--> Base` from the `include` | `Extended EXTENDS Base` | **FIRES**, 1 forbidden hit, exit 2 |
| N5b | emit `Extended --[IMPLEMENTS]--> Base` | `Extended IMPLEMENTS Base` | **FIRES** — scored with its own mutant, because a fence verified through its sibling is verified once |
| N5c | emit `Comparable --[EXTENDS]--> printer` | `Comparable EXTENDS printer` | **FIRES** — the one hierarchy fence that *keeps* `to_bare_name`, deliberately: `printer` is a `class type` and produces no entity, so a resolved to-side could never match it |
| N6a | append a `SCOPE.Operation` named `inherit` | forbidden entity `inherit` | **FIRES**, 1 forbidden entity hit |
| N6b | append a `SCOPE.Component` named `object` | forbidden entity `object` | **FIRES**, scored separately |
| N4b | set `Subtype = "method"` on `make_logger`/`make_counter` | the two restored subtype rows | **DEAD**, entity recall 19/19 → 17/19 |

**All six forbidden relationships and both forbidden entities are now demonstrated capable of firing.** No row in this fixture rests on an argument about why it cannot fire.

### Round 3 — the harness's own assertions

| # | mutant | verdict |
|---|---|---|
| N1 | delete `tokenAt`'s trailing identifier-boundary check | **DEAD** — 6 failing subtests (was ALIVE; five controls added) |
| N2 | reduce the enumeration's criterion to the attribution term | **DEAD** — "no combination diverged" (was ALIVE; the term was dead code, now removed and the 0 asserted) |
| NIT | restore `i += 6` in `refMatchEnd` | **DEAD** — 4 failing subtests (was ungraded; three controls added) |

### One principle, applied in both ledgers

The review noted an asymmetry: three hierarchy rows were **kept** as future-facing while ten subtype rows were **dropped** as redundant — both arguments about a producer that does not exist yet, resolved in opposite directions, and N5 showed the kept ones did not work. The fixture now states and applies one rule: **a row is kept only if it has been demonstrated capable of firing, and dropped otherwise.** Under it the hierarchy fences stay (now demonstrated), eight subtype rows stay dropped (`function` is the only Operation subtype the extractor can emit, so they distinguish nothing), and **`make_logger:function` and `make_counter:function` are restored** — the `let x = object … end` pair is the one place a CST-backed producer has a real choice about what kind of thing a `let` bound to an object body is, and N4b proves the rows are graded.

---

## Verification

- **Ratchet: `scripts/quality/run.sh --runs 1 --ratchet` → exit 0**, "33 gated fixtures held their baseline (24 at 100% must-have recall)". Re-run after the review fixes, not inherited from the first round.
- `go test -count=1 ./internal/extractors/ocaml/ ./internal/quality/... ./cmd/grafel/` → all `ok`. `gofmt -l internal/ cmd/` clean.
- **`./cmd/grafel` was run explicitly** (the `custom_extractor_spans_6118_test.go` digest pin). It passed unchanged; no constant was touched.

### Enumerated ratchet delta — re-derived after the fixture changed

`--update-baseline` was re-run from scratch after the identifier renames and the fence rewrites, rather than assuming the earlier delta still held. It does: `baseline.json` changes are **exactly two things**, and the `ocaml-objects-mini` figures are unchanged by this round.

1. `+ ocaml-objects-mini` — `entity_found 19 / entity_expected 19`, `relationship_found 6 / relationship_expected 6`, `entity_extracted_total 28`, `relationship_extracted_total 68`.
2. `measured_on` `de290e8e7` → `fc9124eef`.

**No figure belonging to a fixture I did not touch was recorded.** The re-run again wanted to change three of them, and they were **again restored, not recorded**:

| fixture | field | on `main` | run wanted |
|---|---|---|---|
| `python-django-mini` | `relationship_extracted_total` | 255 | 251 |
| `python-fastapi-mini` | `entity_extracted_total` / `relationship_extracted_total` | 28 / 57 | 27 / 54 |
| `python-rq-mini` | `entity_extracted_total` / `relationship_extracted_total` | 22 / 43 | 20 / 39 |

That is **pre-existing drift** (#6898) in the *diagnostic* totals only — `*_found` and `*_expected` are unchanged for all three, and the ratchet is green with the restored values, which confirms these fields are diagnostics rather than gated figures. `measured_at` was likewise left at `2026-09-06` rather than stamped with the re-run's date, so the file's provenance describes the run whose numbers it actually records.

**Ratchet re-run after every change above: exit 0**, "33 gated fixtures held their baseline (24 at 100% must-have recall)".

### Four counters bumped 32 → 33

Adding a fixture moves four corpus-size controls in `internal/quality`. Each is a deliberate re-derivation, not a silent bump:

- `absent_relationships_6490_test.go`, `zero_relationships_6490_test.go`, `subtype_assertion_6488_test.go` — coverage-claim counts.
- `TestBaselineCoversEveryFixture` / `TestGoldenSetIsFullyGraded_6273` — satisfied by the baseline entry, no code change.

`TestOnlyTheIntendedGoldenRowsAssertSubtype_6488` is the one that needed judgement rather than a number. It is a control against bulk-adding decorative subtypes, so the fixture's rows were re-measured against its own standard and **eight of nineteen were dropped** (ten at first; two restored on review): the remaining `SCOPE.Operation` rows no longer assert `subtype: "function"`, because `function` is the only Operation subtype the base extractor emits and asserting it there would be decoration. `make_logger` and `make_counter` keep theirs under the cross-ledger principle above, graded by N4b. The nine `SCOPE.Component` rows keep theirs and the ledger entry says why — OCaml is the **first fixture in the corpus where one Kind carries three subtypes at once** (`file`, `module`, `type`, all inside this one fixture), so a row naming only the Kind states something strictly weaker than the author means, and a producer that reclassified a type as a module would satisfy it.

### What the review changed, in one place

Nothing here moves the 75.7% or the 230/408. Every fix is in the **always-on half** — the part that has to carry the finding for anyone without the corpus:

| finding | fix |
|---|---|
| N5: two fences silently unfireable | resolved to-sides; **all 8 forbidden rows now demonstrated to fire** |
| axis B inert (`_` is a word char to Go's `\b`) | respelled `append`; 12/32 → 18/32; same fix in the fixture's three identifiers |
| axis E inert | asserted inert rather than padding the count |
| N2: "a divergence *is* a mis-attribution" false | dead term removed, the 0 asserted, corrected in the test, this body and the #6812 comment |
| N1: trailing identifier boundary ungraded | 5 controls; mutant now DEAD |
| `i += 6` for 3- and 5-byte openers | `i += len(kw)`; 3 controls; corpus figures unchanged, confirming it was latent |
| `106.5%` printed as a rate | separate recall (91.2%) and precision (48.3%) |
| subtype rows dropped too far | `make_logger`/`make_counter` restored, graded by N4b |

### Tree measured in

Everything above was measured in the worktree at `.../scratchpad/impl-6812a-w7n3/wt` on `lane/6812-ocaml-fixture-measure`, branched from `fc9124eef`. The corpus clone lives outside both trees and is not committed.

---

Refs #6812
Refs #6370


cajasmota:

Reviewed in my own detached worktree at `f61625fd4`, with my own shallow clone of `github.com/ocaml/ocaml` at `70165ff23e58ebde4e50bae2c1f9948dbe43ee41` (2,420 `.ml` files — matches the PR's figure exactly). Five mutants scored, three of the six alive-or-vacuous.

**Scope held.** `git diff origin/main..HEAD -- internal/extractors/ocaml/extractor.go internal/treesitter/` is empty against *current* `main`, not just the merge base. No grammar wiring, no new regex, nine files, all test/fixture. Env gating is clean: without `GRAFEL_OCAML_CORPUS` the corpus test SKIPs, the other three pass, no hang and no false green.

---

## Verdict on the 75.7%: it is a measurement, not two approximations disagreeing

I reproduced it byte-for-byte on my own clone: `files=2420 files_with_inherit=68`, `blocks matched=633 (unmatched=218)`, `BLOCK-END WRONG: 479/633 (75.7%)`, `missed 19/216`, `spurious 211`.

Then I graded the grader, which is the thing that mattered here. **I did not take the 14 positive controls as sufficient** — I cross-checked `refMatchEnd` against the repo's own vendored tree-sitter OCaml grammar (`tsocaml.Language()` + `tsofficial.New()`), a genuinely independent oracle, over the same 68 files:

```
tree-sitter cross-check: compared=595 agree=595 DISAGREE=0 uncovered=256 (parse-failed files=0)
disagreements OUTSIDE any tree-sitter ERROR subtree: 0
```

595 of the 633 measured blocks are confirmed exactly right against a real GLR parser, with **zero** disagreements; the 256 uncovered are 224 in the malformed generated file plus 32 where my crude node-kind filter found no wrapper node, not disagreements. And the 218-exclusion claim is exactly true — I histogrammed it: **218 / 218 in one file**, `/testsuite/tests/generated-parse-errors/errors.ml`, and nowhere else.

So the headline stands, and it is the strongest thing in this PR. Everything below is about the parts that do *not* stand.

### The reference is nonetheless not defect-free, and its controls are structurally blind

`refMatchEnd` advances `i += 6` after **any** opener, with the comment *"longest opener; exact length is irrelevant, all are >0"*. That is false: `sig` is 3 bytes and `begin` is 5, so the scan skips 3 (resp. 1) bytes of real source, and if a comment or string opener sits in that window it is scanned **as code**. Demonstrated on valid OCaml:

| input | want | `refMatchEnd` |
|---|---|---|
| `module type T = sig end` | 20 | **−1** |
| `module type T = sig (* end *) val f : int end` | 42 | **23** (the `end` *inside the comment*) |
| `module type T = sig\n(* end *)\n val f : int end` | 43 | **23** |
| `let x = object method a = begin(* end *) 1 end end` | 47 | **43** |

That is defect (3) from your own list — comment leakage — living inside the reference. The 14 controls cannot see it because **every** comment/string/quoted-string control opens with `object`, which is exactly 6 bytes, so the skip is a no-op in all of them. Corpus-latent (the tree-sitter check found 0 real instances), so the number survives, but "the reference is validated, not asserted" is stronger than the controls actually support. Please either fix `i += len(kw)` or state the limitation.

---

## Mutant table

| # | mutant | verdict |
|---|---|---|
| N1 | `tokenAt`: delete the **trailing** identifier-boundary check (`if j < len(src) && isIdentByte(src[j]) { return "" }`) — makes the reference more permissive | **ALIVE, distinguishable and ungraded.** Whole package green. Distinguishers: `object method endian = 1 end` → 22 (want 33); `signature` / `structure` / `beginning` / `objects` → **−1**. The 14 controls test only the *leading* boundary (`append`, `myobject`, `design`); no control has an identifier *starting* with a keyword. Add `endian` / `signature` / `objects` controls. |
| N2 | `EnumeratedSpace`: divergence criterion `!endsAgree(...) \|\| inTruth != inWalker` → `inTruth != inWalker` only | **ALIVE — the attribution term is dead code.** Test fails "no combination diverged": **0 of 32 inputs mis-attributes the `inherit`.** All 12 divergences are block-*end* differences (`in_walker=true` on every logged line, and the test `Fatalf`s unless `inTruth` too). See below — this falsifies a claim in the PR body and in the #6812 comment. |
| N3 | `expected.json`: `total_area CALLS area_of` resolved `to_name/to_kind/to_file` → `to_bare_name` (permissive) | **DEAD** — `relationship_found 6 → 5`, exit 2. The row's "only row whose TO side resolves" claim is real and graded. |
| N4 | `expected.json`: `point` `subtype: "type"` → `"module"` (permissive/false) | **DEAD** — `entity_found 19 → 18`, exit 2. Subtype assertions are live, so the nine kept ones carry weight. |
| N5 | `extractor.go`: emit `Extended --[EXTENDS]--> Base` from `include` (i.e. fire the absence fence for real) | **ALIVE — the fence is silently unfireable.** Edge emitted (`relationship_extracted_total 68 → 70`), `forbidden_hits 0`, exit 0. Same edge as `CALLS`: still 0. Change the row's target to `to_name: "Base", to_kind: "SCOPE.Component", to_file: "containers.ml"` and the *same* mutant gives `forbidden_hits 1`, exit 2. |

### N5 is the finding I would block on

`Extended EXTENDS Base` and `Extended IMPLEMENTS Base` are written with `to_bare_name: "Base"`. `Base` is a top-level module that binds to an entity, so any edge to it carries a **resolved** target and `to_bare_name` can never match. This is *precisely* the trap you found on `make_logger → backend_label`, documented as "silently unfireable", and then wrote "assume there are others" about — there are two more, in the same file, and I found them by firing the edge rather than by reading.

It also falsifies the stated reason. The PR says these rows are unfireable because *"OCaml's edge vocabulary contains no hierarchy kind at all, so no mutant of this extractor can emit an EXTENDS"*. I emitted one from the extractor in ~8 lines and the fence stayed green. The reason is the target dialect, not the vocabulary. The day #6370's OCaml arm lands and emits `Extended EXTENDS Base`, **both fences pass**.

`Comparable EXTENDS printer` is fine as `to_bare_name` — `printer` is a `class type` and produces no entity.

---

## Two of the five enumeration axes are inert, and one of them is the one that matters

Read the recorded `divergingCombinations6812` pairwise:

- **Axis B (`endIdent`) is inert in all 16 pairs.** `-----`/`-B---` both diverge, `A----`/`AB---` both don't, `--CD-`/`-BCD-` both do, and so on for every pair.
- **Axis E (`trailer`) is inert too** — the E-half of the set is the non-E half with an `E` appended, identically.

So "32 combinations, 12 diverge" is really **8 combinations, 3 diverge** over (A, C, D). The two extra axes multiply the row count by 4 and change no outcome.

Axis B is the damaging one, because it exists to grade defect (2) — the `\b`-on-a-slice bug — and it cannot. The generator writes `method append_all x = x`. Go's `\b` treats `_` as a word character, so `\bend\b` against `end_all…` does **not** match and the hazard never fires. I re-ran the (A,C,D) cube varying only the identifier:

```
endIdent=""            diverging = [---  -CD  ACD]     (3/8)
endIdent="append_all"  diverging = [---  -CD  ACD]     (3/8)   <- inert
endIdent="append"      diverging = [-C-  AC-  --D  A-D  -CD  ACD]  (6/8)
endIdent="backend"     diverging = [-C-  AC-  --D  A-D  -CD  ACD]  (6/8)
endIdent="legend"      diverging = [-C-  AC-  --D  A-D  -CD  ACD]  (6/8)
```

Dropping the `_all` **doubles** the divergence rate. The always-on test — the part that carries the finding without a network — currently *under*-states the walker's defect and grades defect (2) not at all.

**The same `_` problem is in the fixture.** `append_all`, `backend_label`, `legend_of` are all `end` immediately followed by `_<word>`, so none of the three exercises the identifier hazard they are there for, and the claimed axis *"three identifiers containing `end`, in three different positions"* is not varied — all three are structurally `…end_<word>` (`legend_of`'s note, "`end` is neither a prefix nor a suffix", is not true of its content). One of them should lose the underscore, e.g. `let backend lvl = …` or a `method append x = x` in an object body. Note `containers.ml:27` already has a real `List.append xs` that does trigger it, inside the functor nothing extracts.

---

## The "mis-attribution" claim in the PR body and in the issue comment is false

The committed test logs `12/32 enumerated combinations mis-attribute the `inherit``, and the PR says *"each input holds exactly one `object … end` block with exactly one `inherit` in it, so a divergence **is** a mis-attribution rather than a proxy for one."* N2 shows the opposite: `inTruth != inWalker` is never true across all 32, so a divergence is *exactly* a proxy. The enumeration measures block-end divergence and says nothing about `inherit` attribution. Fix the log line and both write-ups, or add a case where the walker's span actually crosses the `inherit` (the sibling-objects-in-a-module shape in `TestDepthWalker_InheritIsMisAttributed_Constructed` does it; the enumerated generator never can, because the `inherit` is always the first body line).

Related, minor but committed: the corpus test prints `INHERIT MIS-ATTRIBUTED, either direction: 230/216 (106.5%)`. `spurious` is drawn from all (block, inherit) pairs in the file while the denominator is only the truly-enclosed ones. The "more wrong than right" claim is still defensible as a *precision* statement (the walker attributes 197 + 211 = 408 pairs, 211 of them wrong), but it should be normalised that way rather than printed as a >100% rate.

---

## Ratchet and baseline

Confirmed correct and complete: `baseline.json` changes are exactly the `ocaml-objects-mini` block and `measured_on`. **No python fixture figure was recorded** — the three drifting `*_extracted_total` values are untouched. Restoring rather than recording was the right call; recording them would have moved someone else's diagnostic figures under cover of this PR. `entity_found 19 / 19`, `relationship_found 6 / 6`, `forbidden_hits 0`, `forbidden_entity_hits 0` reproduced from a clean build.

Fixture-content claims I checked against the bodies rather than the labels: `as super` (shapes.ml:24) ✓, multiple inheritance (`tagged_circle`, two `inherit` lines) ✓, functor (containers.ml:24) ✓, `module type` ✓, class-type `inherit` ✓, `let`-body `object` with and without `inherit` ✓. Only the three-`end`-identifiers axis fails to match its label (above).

---

## The dropped subtype assertions — second opinion

The justification checks out on the facts: `Subtype:` appears four times in `extractor.go`, and `"function"` (`:210`) is the only one on a `SCOPE.Operation`, so the ten dropped rows would each have asserted the single value the producer can emit. And N4 proves subtype assertions are graded, so this is a redundancy judgement, not a gradedness one. Eight of the ten I would leave dropped.

**I would restore two: `make_logger:function` and `make_counter:function`.**

Not for today's producer — for the one this PR exists to justify. Those two are the `let x = object … end` pair, and they are the one place in the corpus where a CST-backed producer has a genuine choice about what kind of thing a `let` bound to an object body *is*; a grammar swap that reclassified them (method, class, object) is exactly the silent recall change the owner's #6812 comment says must not happen invisibly, and `make_logger` is already the row the fixture exists for. Restoring them costs two strings and makes the ledger's own argument — "a row naming only the Kind states something strictly weaker than the author means" — apply where it actually bites.

I also note the asymmetry: three hierarchy rows were **kept** on the grounds that they are unreachable today but aimed at a future arm, while ten subtype rows were **dropped** on the grounds that they are redundant today. Both are arguments about a producer that does not exist yet, resolved in opposite directions. That is fine as a judgement, but it is worth stating rather than letting the two ledgers argue past each other — especially since N5 shows the kept-for-the-future rows do not actually work.

---

## What I would need to approve

1. **N5** — rewrite the two `Extended` hierarchy fences with resolved targets, and correct the stated reason for their non-firing. This one is load-bearing for #6370's OCaml arm.
2. **Axis B** — drop the `_` so the axis grades defect (2), and re-record `divergingCombinations6812`. Same in the fixture for at least one of the three identifiers.
3. **N2 / the mis-attribution claim** — correct the log line, the PR body and the #6812 comment, or make the enumeration actually cross an `inherit`.
4. **N1** — three controls for the trailing identifier boundary.
5. Nits: `i += len(kw)` in `refMatchEnd`; normalise the 106.5% as precision.

None of these move the 75.7%, which I independently confirmed against tree-sitter and consider sound. They are all about the always-on half — the half that has to carry the finding for everyone who does not have the corpus.

🤖 Independent review by Claude Opus 5

cajasmota:
Coordinator mutant on `10c13b08f`. One mutant, **DEAD**, and it is killed by the guard added in this round — which is the most useful thing I can report about it.

## Choosing the mutant

The enumeration asserts an **exact diverging set** (`strings.Join(got,",") != strings.Join(divergingCombinations6812,",")`), not a count or a threshold, so any mutant that makes the walker *worse* dies trivially and proves little. The sharper question for a measurement PR is the opposite one: **do these tests actually observe the defect, or do they pass for some other reason?** So I mutated toward "less broken".

## MC-6938 — neuter the indentation fallback in `extractModuleBody`

`return extractLetBody(src, afterPos)` → `return ""`, in `extractModuleBody` only, leaving `extractLetBody` and every other caller intact. Anchor count 1, `go vet` clean, `-count=1`, baseline green first (`ok 0.593s`), restored by `cp` from a shasum-verified backup (`94fad9af5…`), `git status --porcelain` empty.

**Verdict: DEAD**, by:

```
--- FAIL: TestDepthWalker_ObjectBlockEnd_EnumeratedSpace
    depth_walker_inherit_accuracy_6812_test.go:435: -----: this generator is not
    supposed to be able to produce an inherit MIS-ATTRIBUTION — it varies where
    the block ENDS, and the inherit is always the first body line. If it now can,
    the enumerated finding changed and both write-ups need re-deriving.
```

## Why this is worth more than a tick

That is the **0/32 guard** — the thing that did not exist two rounds ago, and that I asked for only because a reviewer showed the "a divergence *is* a mis-attribution" claim was false. It is now load-bearing: it caught a change to a *different* function that silently turned the generator into something capable of producing the mis-attribution the write-up says it cannot produce, and it said so in a message that explains what to do about it rather than just failing.

A guard added to correct a false claim usually reads as bookkeeping. This one earned its place within one round.

**One thing my hypothesis got wrong, stated because it bears on the finding:** I expected the fallback to be the over-reach mechanism behind the constructed mis-attribution, so I predicted `TestDepthWalker_InheritIsMisAttributed_Constructed` would fail under this mutant. It did not — it still sees mis-attribution with the fallback neutered. So the sibling-swallow the constructed test pins does **not** come from the indentation fallback alone, and my reading of defect (1) was incomplete. That does not change any number in the PR, and I am not asking for anything; it is recorded so nobody inherits my wrong mental model from this comment.

Accepted from your round: all six forbidden relationships and both forbidden entities now demonstrated fireable by injection rather than argued; `N5c` correctly **keeping** `to_bare_name` because a `class type` binds to no entity is exactly the right kind of exception — a rule applied with its reason rather than uniformly; axis E asserted inert instead of padding a count; and the corpus figures being **byte-identical** after the reference-matcher fixes is the evidence that those two defects were latent rather than load-bearing, which is the strongest thing that could have been said about them.

Refs #6812, #6370.

PR #6941, opened 2026-09-06
fix(#6929): TestStaleReindexGuard_* counted the durable reindex ledger absolutely — 428-vs-2 on a reused daemon root, green on a fresh one

### LINKED ISSUE #6929: TestStaleReindexGuard_* counts durable reindex requests unscoped to the run — green on a fresh root, 428-vs-2 on a reused one

cajasmota:
## Both sides reproduced first

Same worktree, same binary, only the root differs. `GRAFEL_HOME` and
`GRAFEL_DAEMON_ROOT` both pointed at scratch dirs; the "reused" root is literally
the fresh one after one prior `go test` run.

| root | exit | result |
|---|---|---|
| fresh `GRAFEL_DAEMON_ROOT` | 0 | `ok  github.com/cajasmota/grafel/internal/daemon  5.074s` |
| same root, 2nd run | 1 | 4 failures (below) |

```
--- FAIL: TestStaleReindexGuard_ExactlyOnceAcrossHeartbeats
    pending reindex requests = 2, want exactly 1 (no storm)
--- FAIL: TestStaleReindexGuard_SelfClearsAfterReindex
    after in-flight heartbeats, pending = 3, want 1
--- FAIL: TestStaleReindexGuard_FirstHeartbeatDoesNotStampede
    durable pending reindex requests after one heartbeat = 144, want 2
--- FAIL: TestStaleReindexGuard_StuckSlotExpires
    wedged repo has 2 pending requests, want exactly 1
```

## The mechanism — verified, and one step deeper than the issue said

The issue's framing ("counts durable requests unscoped to the run") is correct as
far as it goes, but it does not say why the tests are exposed to another run's
state at all, given that they *do* isolate themselves. They isolate the wrong
variable:

- `repoBaseDir` (state_path.go:601) resolves `GRAFEL_DAEMON_ROOT` **first** and only
  falls back to `GRAFEL_HOME`/`StoreDir()`. The older members of the family set
  `t.Setenv("GRAFEL_HOME", t.TempDir())` and nothing else, so whenever
  `GRAFEL_DAEMON_ROOT` is set in the environment — which is what every agent and
  every isolated daemon does — that isolation is a **no-op** for the requests dir.
- The fake repo paths (`/repo/000`…`/repo/139`, `/repo/stale`, `/repo/wedged`) are
  constants, so every run hashes to the same `state/<hash>/refs/_unknown/requests`
  directory. Nothing ever drains those requests in a test, so the directory grows
  by one file per write, forever.
- `countPendingReindex` then counted the directory absolutely.

That also explains the exact numbers, which are not arbitrary. Within one run,
`FirstHeartbeatDoesNotStampede` writes 2 and `140RepoUpgrade_Simulation` writes 140
**for the same repo paths** — so each run leaves 142 behind: 2 → 144 → 286 → **428**,
the number in the issue, i.e. the 4th run in that root. (The review's own reused
root gave 570 — the 5th.) The count also confirms the issue's open question:
leftovers accumulate from **normal** runs, not only interrupted ones.

Not the mechanism, checked and ruled out: `reconcileLocked` (stale_reindex.go:377,
reached on first `maybeEnqueue`) globs the same shared root for durable
`migration.json` markers, which would be a *behavioural* leak rather than a
miscount. After 8 runs the root held **1665** leftover `*.request.json` files and
**0** markers — the markers are cleared on slot release, so that path never fired.
Worth knowing it exists (an interrupted run could strand one), but it is not this
defect.

## The fix: scope by identity, not by clearing

`reindexLedger` snapshots the request IDs already on disk when the test begins;
`countNew(repo)` reports only IDs absent from that snapshot. IDs are a fresh uuid
per `requests.Write`, so "not in the snapshot" is exactly "written by this test".

- **Nothing is weakened.** The absolute count only ever asserted "the dir started
  empty", which is not a property of a deliberately durable store. A stampede that
  writes 140 requests still reports 140 — see the mutants.
- **Nothing is cleared or reset.** No `t.TempDir()` was added, no root is wiped. The
  whole-root snapshot means no caller has to enumerate the repos it is about to
  touch, so the next test to use the helper is scoped by construction rather than
  by remembering to be.
- Building the ledger late fails **loudly** (0 vs the wanted 1), not vacuously.

### Review finding: the snapshot itself was not a pure read (fixed in c0ce5f2)

`requests.ListPending` **deletes** any request whose ack is already on disk
(requests.go:236) as exactly-once catch-up cleanup. The first version of
`newReindexLedger` called it over the **entire** daemon root, so taking a baseline
wrote into every repo directory in a shared root, 12× per package run — a fix for
cross-run interference that itself interfered across runs, and a direct
contradiction of this body's "nothing is cleared" claim. The snapshot now reads
directory entries with `os.ReadDir` and strips the suffix; the request ID is the
filename stem, which is all a baseline needs.

That property is now **pinned, not asserted in prose** —
`TestReindexLedgerSnapshotIsPure` plants an *acked* request (the one `ListPending`
removes) in an unrelated repo dir and requires it to survive the sweep, plus a
positive control that the baseline actually observed the planted ID. Without the
control, the survival assertion is satisfied by a snapshot that reads nothing.

## Siblings — checked, and the split is real

The `TestStaleReindexGuard_*` family is **33** tests; **15** call sites across 2 files
now go through the ledger.

- The *later* members, added in review rounds, already set
  `t.Setenv(EnvRoot, t.TempDir())` **and** use `realRepoDirs` (per-run `t.TempDir()`
  repo paths) — doubly isolated, never exposed.
- The 4 that failed, plus `CurrentRepo_NoRequest` and `stale_reindex_6175_test.go`'s
  `DeclinedRepoIsNotRetriedForever`, were exposed. The latter two are only
  *accidentally* safe: they assert 0 for a repo path nothing else writes to.
- **`IdleWindowBetweenBatches` and `140RepoUpgrade_Simulation` do NOT share the
  defect** — neither makes any durable-count assertion at all (`grep countNew|ListPending`
  over both function bodies: 0 hits). They are the **polluters**, not victims: the
  simulation is what writes the 140 requests, against the same constant repo paths
  the other tests use, that everything else then trips over. A future reader
  chasing accumulated cruft should look at *them*, and at nothing draining those
  requests — not at their assertions.

Swept the rest of `internal/daemon` for the same pattern —
`requests_drain_test.go`, `requests_drain_unblock_test.go`, `index_split_mode_test.go`
all read `requests.ListPending` absolutely, and all set `EnvRoot` to a `t.TempDir()`
**and** use a `t.TempDir()` repo path. The review extended the sweep to
`rebuild_wait_test.go` and `status_rebuild_inflight_6014_test.go` (the former isolated
through the `withSplitWaitTest` helper) and found them safe for the same reason.

## Mutants (all `-count=1`, each scored on BOTH a fresh and a reused root)

Every mutant edit was applied by a script that asserts an exact anchor occurrence
count of 1 before writing, and the applied text was then grepped back.

| # | mutant | verdict |
|---|---|---|
| A | drop the batch-admission gate (`if false && !g.admitLocked(...)`) | **DEAD** both roots — but killed by the `admitted 140 of 140` check at line 263, *before* the durable count. So it does not grade the changed line; A2 exists because of that. |
| A2 | duplicate the durable `requests.Write` (admissions unchanged at 2, ledger gets 4) | **DEAD** both roots, killed by the changed assertions themselves and with **identical messages on both**: `durable pending reindex requests after one heartbeat = 4, want 2`, plus 6 sibling assertions. This is the killing mutant: the guard's durable output genuinely breaks, admissions look fine, and the scoped count still catches it. |
| B | send the durable write to a different directory (guard admits but queues nothing) | **DEAD** both roots — `= 0, want 1` / `= 0, want 2`. Grades the under-count direction: the assertion observes the durable artefact, not the in-memory admission decision it could otherwise be re-asserting. |
| C | make `newReindexLedger` snapshot nothing (`pre` empty ⇒ the old absolute count) | **DEAD** on a reused root (`= 994, want 2`; `= 7, want 1`). **Equivalent on a pristine root** — necessarily so: that is exactly the defect's premise, and it is why the pre-fix suite was green for everyone who only ever ran it once. Reported as distinguishable-and-graded, not as a clean kill. |
| D | restore the `requests.ListPending` whole-root sweep (the reviewed defect) | **DEAD** — `newReindexLedger destroyed another run's durable request …: no such file or directory`. |
| E | the C mutant scored against the purity test's positive control | **DEAD** — `baseline did not observe the planted request … the sweep read nothing, so purity is untested`. Both halves of the new test are graded. |

`stale_reindex.go` is byte-identical to `HEAD` after the mutation round (`git status`
shows only the test files; `shasum` matched the pre-mutation backup).

## Verification

Full `./internal/daemon/` suite, `-count=1`, on the final head:

- reused root (holding 4 runs of leftovers): **PASS**, exit 0, 38.568s
- same root again: **PASS**, exit 0, 33.191s
- fresh root: **PASS**, exit 0, 30.113s

`gofmt -l internal/daemon/` clean; `go vet ./internal/daemon/` exit 0. Test-only
change; no package outside `internal/daemon` touched, and nothing under
`internal/daemon/watch` (#6935).

Closes #6929

### DIFF
--- a/internal/daemon/stale_reindex_6175_test.go
+++ b/internal/daemon/stale_reindex_6175_test.go
@@ -31,6 +31,7 @@ import (
 func TestStaleReindexGuard_DeclinedRepoIsNotRetriedForever(t *testing.T) {
 	t.Setenv("GRAFEL_HOME", t.TempDir())
 	t.Setenv(EnvRoot, t.TempDir())
+	led := newReindexLedger(t)
 	clk := &fakeClock{t: time.Unix(1_700_000_000, 0)}
 
 	const declined = "/repo/worktree"
@@ -62,7 +63,7 @@ func TestStaleReindexGuard_DeclinedRepoIsNotRetriedForever(t *testing.T) {
 	if admissions != 0 {
 		t.Errorf("a repo the scheduler declines to accept was admitted %d times over 12 simulated hours, want 0", admissions)
 	}
-	if n := countPendingReindex(t, declined); n != 0 {
+	if n := led.countNew(declined); n != 0 {
 		t.Errorf("wrote %d reindex request(s) for a repo whose enqueue the scheduler drops, want 0", n)
 	}
 	if !othersMigrated {
--- a/internal/daemon/stale_reindex_test.go
+++ b/internal/daemon/stale_reindex_test.go
@@ -18,17 +18,79 @@ import (
 	"github.com/cajasmota/grafel/internal/indexstate"
 )
 
-// countPendingReindex returns how many KindReindex requests are queued for
-// repoPath's control-plane requests dir.
-func countPendingReindex(t *testing.T, repoPath string) int {
+// pendingReindexIDs returns the IDs of the KindReindex requests currently
+// queued for repoPath's control-plane requests dir.
+func pendingReindexIDs(t *testing.T, repoPath string) map[string]bool {
 	t.Helper()
 	recs, err := requests.ListPending(requestsDirForRepo(repoPath))
 	if err != nil {
 		t.Fatalf("ListPending: %v", err)
 	}
-	n := 0
+	ids := make(map[string]bool, len(recs))
 	for _, r := range recs {
 		if r.Kind == requests.KindReindex && r.RepoPath == repoPath {
+			ids[r.ID] = true
+		}
+	}
+	return ids
+}
+
+// reindexLedger scopes durable request counts to the CURRENT RUN (#6929).
+//
+// The requests ledger is durable ON PURPOSE and its root is shared: repoBaseDir
+// keys off GRAFEL_DAEMON_ROOT when that is set and only falls back to
+// GRAFEL_HOME otherwise, so a test that isolates itself with
+// `t.Setenv("GRAFEL_HOME", t.TempDir())` alone still writes its requests into
+// whatever daemon root the developer's environment names. The fake repo paths
+// below ("/repo/000"…) are stable across tests AND across runs, so the requests
+// dir for a given fake repo accumulates one file per write, forever. An
+// absolute count therefore measured "every request ever written into this root
+// for this repo", not "the requests this test caused": green against a fresh
+// root, 144 / 286 / 428-vs-2 against a root reused for a 2nd / 3rd / 4th run.
+//
+// Scoping is by IDENTITY, not by clearing the root: the ledger snapshots the
+// request IDs already on disk when the test begins, and countNew reports only
+// IDs that were NOT in that snapshot. Requests are keyed by a fresh uuid per
+// write, so "not in the snapshot" is exactly "written during this test". The
+// assertions keep their full strength — a stampede that writes 140 requests
+// still reports 140 — and no other run's durable state is destroyed.
+//
+// Build the ledger BEFORE the action under test. If it is built afterwards the
+// test's own requests land in the baseline and countNew under-reports, which
+// fails the assertion loudly (0 vs the wanted 1) rather than passing vacuously.
+type reindexLedger struct {
+	t   *testing.T
+	pre map[string]bool
+}
+
+// newReindexLedger snapshots every reindex request ID already durably present
+// anywhere under the active requests root, so no caller has to enumerate the
+// repos it is about to touch.
+func newReindexLedger(t *testing.T) *reindexLedger {
+	t.Helper()
+	l := &reindexLedger{t: t, pre: map[string]bool{}}
+	dirs, err := discoverRequestsDirs(requestsRoot())
+	if err != nil {
+		t.Fatalf("discoverRequestsDirs: %v", err)
+	}
+	for _, d := range dirs {
+		recs, err := requests.ListPending(d)
+		if err != nil {
+			t.Fatalf("ListPending(%s): %v", d, err)
+		}
+		for _, r := range recs {
+			l.pre[r.ID] = true
+		}
+	}
+	return l
+}
+
+// countNew returns how many KindReindex requests THIS RUN added for repoPath.
+func (l *reindexLedger) countNew(repoPath string) int {
+	l.t.Helper()
+	n := 0
+	for id := range pendingReindexIDs(l.t, repoPath) {
+		if !l.pre[id] {
 			n++
 		}
 	}
@@ -40,6 +102,7 @@ func countPendingReindex(t *testing.T, repoPath string) int {
 // fingerprint) writes exactly ONE reindex request, not N.
 func TestStaleReindexGuard_ExactlyOnceAcrossHeartbeats(t *testing.T) {
 	t.Setenv("GRAFEL_HOME", t.TempDir())
+	led := newReindexLedger(t)
 	const repo = "/repo/stale"
 	g := newStaleReindexGuard()
 
@@ -55,7 +118,7 @@ func TestStaleReindexGuard_ExactlyOnceAcrossHeartbeats(t *testing.T) {
 	if firing != 1 {
 		t.Errorf("maybeEnqueue returned true %d times across 12 heartbeats, want exactly 1", firing)
 	}
-	if got := countPendingReindex(t, repo); got != 1 {
+	if got := led.countNew(repo); got != 1 {
 		t.Fatalf("pending reindex requests = %d, want exactly 1 (no storm)", got)
 	}
 }
@@ -64,6 +127,7 @@ func TestStaleReindexGuard_ExactlyOnceAcrossHeartbeats(t *testing.T) {
 // current-format repo (required=false) never enqueues.
 func TestStaleReindexGuard_CurrentRepo_NoRequest(t *testing.T) {
 	t.Setenv("GRAFEL_HOME", t.TempDir())
+	led := newReindexLedger(t)
 	const repo = "/repo/current"
 	g := newStaleReindexGuard()
 
@@ -72,7 +136,7 @@ func TestStaleReindexGuard_CurrentRepo_NoRequest(t *testing.T) {
 			t.Fatalf("current-format repo must never enqueue (heartbeat %d)", i)
 		}
 	}
-	if got := countPendingReindex(t, repo); got != 0 {
+	if got := led.countNew(repo); got != 0 {
 		t.Fatalf("pending reindex requests = %d, want 0 for a current repo", got)
 	}
 }
@@ -84,6 +148,7 @@ func TestStaleReindexGuard_CurrentRepo_NoRequest(t *testing.T) {
 // one fresh request. This is the anti-#5891 loop-guard end to end.
 func TestStaleReindexGuard_SelfClearsAfterReindex(t *testing.T) {
 	t.Setenv("GRAFEL_HOME", t.TempDir())
+	led := newReindexLedger(t)
 	const repo = "/repo/lifecycle"
 	g := newStaleReindexGuard()
 
@@ -100,7 +165,7 @@ func TestStaleReindexGuard_SelfClearsAfterReindex(t *testing.T) {
 			t.Fatalf("in-flight heartbeat %d must not re-enqueue (same stale generation)", i)
 		}
 	}
-	if got := countPendingReindex(t, repo); got != 1 {
+	if got := led.countNew(repo); got != 1 {
 		t.Fatalf("after in-flight heartbeats, pending = %d, want 1", got)
 	}
 
@@ -119,7 +184,7 @@ func TestStaleReindexGuard_SelfClearsAfterReindex(t *testing.T) {
 			t.Fatalf("second-generation heartbeat %d must not re-enqueue", i)
 		}
 	}
-	if got := countPendingReindex(t, repo); got != 2 {
+	if got := led.countNew(repo); got != 2 {
 		t.Fatalf("total pending across two stale generations = %d, want 2 (one per generation)", got)
 	}
 }
@@ -182,6 +247,7 @@ func heartbeat(g *staleReindexGuard, repos []string, stale map[string]bool) []st
 // requests. Pre-fix this admits all 140.
 func TestStaleReindexGuard_FirstHeartbeatDoesNotStampede(t *testing.T) {
 	t.Setenv("GRAFEL_HOME", t.TempDir())
+	led := newReindexLedger(t)
 	clk := &fakeClock{t: time.Unix(1_700_000_000, 0)}
 	g := newTestGuard(clk, 2, 30*time.Second, 15*time.Minute)
 
@@ -199,7 +265,7 @@ func TestStaleReindexGuard_FirstHeartbeatDoesNotStampede(t *testing.T) {
 
 	total := 0
 	for _, r := range repos {
-		total += countPendingReindex(t, r)
+		total += led.countNew(r)
 	}
 	if total != 2 {
 		t.Fatalf("durable pending reindex requests after one heartbeat = %d, want 2", total)
@@ -282,6 +348,7 @@ func TestStaleReindexGuard_IdleWindowBetweenBatches(t *testing.T) {
 // progress for the others, it does not retry the failure.
 func TestStaleReindexGuard_StuckSlotExpires(t *testing.T) {
 	t.Setenv("GRAFEL_HOME", t.TempDir())
+	led := newReindexLedger(t)
 	clk := &fakeClock{t: time.Unix(1_700_000_000, 0)}
 	g := newTestGuard(clk, 1, 30*time.Second, 15*time.Minute)
 
@@ -309,7 +376,7 @@ func TestStaleReindexGuard_StuckSlotExpires(t *testing.T) {
 	if len(got) != 1 || got[0] != "/repo/next" {
 		t.Fatalf("after the slot TTL admitted %v, want [/repo/next] — a wedged repo must not block the migration", got)
 	}
-	if n := countPendingReindex(t, "/repo/wedged"); n != 1 {
+	if n := led.countNew("/repo/wedged"); n != 1 {
 		t.Fatalf("wedged repo has %d pending requests, want exactly 1 — TTL expiry must not re-enqueue a known failure", n)
 	}
 }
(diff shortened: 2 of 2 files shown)


cajasmota:
## Independent adversarial review — APPROVE (two body corrections + one small follow-up)

Own worktree at `32ccc28c4`, own `GRAFEL_HOME` + `GRAFEL_DAEMON_ROOT` scratch roots, `-count=1`
everywhere, every mutant scored on **both** a fresh root and a root already holding 4 runs of
leftovers. Every mutant edit was applied by a script asserting an exact anchor occurrence count of 1
and grepped back; every restore was `cp` from a shasum-verified backup, never `git checkout --`.

### Both sides reproduced independently, and the arithmetic lands exactly

| | fresh root | reused root (4 prior runs) |
|---|---|---|
| pre-fix (`fc9124eef` test files) | exit 0 | **exit 1**, `= 570, want 2` |
| PR head | exit 0 | exit 0 |

`570 = 2 + 4×142`. I ran the family 4 times before the pre-fix run, so 570 is the 5th run in the
sequence `2 → 144 → 286 → 428 → 570` — the issue's 428 is the 4th, exactly as claimed. The
whole-family run leaves **153** `*.request.json` files per pass in my root, from *normal* completed
`go test` runs, so the issue's "Not established" question is answered: no interruption is needed.

`repoBaseDir` (`state_path.go:601`) read directly: `if root := os.Getenv(EnvRoot); root != ""` returns
`<root>/state/<hash>` and only *otherwise* falls back to `StoreDir()` (`$GRAFEL_HOME/store`).
`GRAFEL_HOME`-alone isolation is indeed a no-op whenever a daemon root is set. Verified, not taken on
trust.

**Rival mechanism ruling-out reproduced.** After 8 whole-family runs my root held ~1200
`*.request.json` and `find … -name migration.json` returned **0**. `reconcileLocked`'s marker glob
never fires here. The discarded hypothesis holds up.

### Mutants — mine (M1–M5), plus A2 and C reproduced

| # | mutant | fresh root | reused root |
|---|---|---|---|
| **A2** (yours) | duplicate the durable `requests.Write` in `stale_reindex.go` | **DEAD** — 7 assertions, `= 4, want 2` etc. | **DEAD** — *byte-identical failure text and numbers* |
| **C** (yours) | `newReindexLedger` snapshots nothing | equivalent (exit 0) | **DEAD** — `= 12/19/1562, want 1/1/2` |
| **M1** | `countNew` *consumes* IDs (`l.pre[id] = true` after counting) — permissive on repeat calls | **DEAD** — `two generations = 1, want 2`; `re-admitted: 3 -> 0` | **DEAD**, identical |
| **M2** | drop `r.RepoPath == repoPath` from `pendingReindexIDs` | **ALIVE** | **ALIVE** |
| **M3** | build the ledger **after** the action in `ExactlyOnceAcrossHeartbeats` | **DEAD** — `= 0, want exactly 1` | **DEAD**, identical |
| **M4** | production `requests.Write` gives a *deterministic* ID per (kind, repo) — kills the uuid-uniqueness premise | **DEAD** — 3 assertions, `= 1, want 2` / `= 0, want 1` | **DEAD**, identical |
| **M5** | drop `r.Kind == requests.KindReindex` | **ALIVE** | **ALIVE** |

**The guard can still fail — yes, and root-independently.** A2, M1, M3 and M4 all die on *both*
roots with byte-identical messages and identical numbers. That identity is the property this PR is
actually claiming, and it is the thing I most wanted to see fail and could not make fail.

- **A2 confirmed as the killing mutant.** 7 assertions fire; the durable count is genuinely
  load-bearing, not shadowed by the `admitted 140 of 140` pre-check.
- **M4 is the one I'd add to your set.** The whole design rests on "a fresh uuid per write ⇒ not in
  the snapshot == written by this test". I broke exactly that premise in production
  (`requests.Write` line 149–151) and three assertions caught it. The premise is graded, not assumed.
- **M1** kills the one site A2 misses: `FailedRepoIsRetriedThenReported`'s `before`/`after` delta
  needs `countNew` to be *cumulative since the snapshot*, not *since the last call*.
- **M3 confirms your "fails loudly" claim** verbatim (`0, want exactly 1`). One caveat worth writing
  down: that loudness is **polarity-dependent**. The three `!= 0` sites (`CurrentRepo_NoRequest`,
  6175's `declined`, the `ReconcileReEnqueues` precondition) would pass *vacuously* under a
  late-built ledger. They happen to write nothing, so it is not a live hole today — but the doc
  comment's "fails loudly" should say "for any site asserting a non-zero count".
- **M2/M5 alive**: both filters are redundant given `requestsDirForRepo(repoPath)` is already
  per-repo-hash keyed and nothing writes a non-reindex request there. Defensive, not distinguishable
  by any input this suite can produce. Reported as *equivalent under the current suite*, no action.

`stale_reindex.go` and `requests/requests.go` restored and shasum-matched to `32ccc28c4` and to
`fc9124eef` (`53c3a327…` / `8377d57d…`); `git status --porcelain` empty afterwards.

### Finding — the ledger now *writes* to every repo dir in the shared root

`requests.ListPending` is not a pure read: `requests.go:236-239` deletes any `*.request.json` that
has a matching ack ("best-effort catch-up cleanup"). `newReindexLedger` calls it over **every** dir
`discoverRequestsDirs(requestsRoot())` finds — the whole daemon root, not just the fake repos. So the
PR body's "no other run's durable state is destroyed" is not quite right. Demonstrated:

```
$ mkdir -p $ROOT/state/deadbeefcafe/refs/_unknown/requests   # an unrelated repo
$ …/probe1.request.json  …/probe1.ack.json
$ go test ./internal/daemon/ -run TestStaleReindexGuard_CurrentRepo_NoRequest -count=1   # exit 0
$ ls …/requests
probe1.ack.json          # probe1.request.json is gone
```

Pre-fix, `countPendingReindex` only ever touched `requestsDirForRepo(<fake repo>)`. This PR widens a
test helper's write radius to the entire root, 12× per `go test ./internal/daemon` — including the
real `~/.grafel` for anyone who forgets to set a scratch root. Only *already-acked* files are removed,
which is idempotent and mirrors what the engine's own drain does, so I am **not blocking on it** — but
the fix is small: the ID is the filename stem, so the snapshot can use `os.ReadDir` + strip
`requestSuffix` instead of `ListPending` and be side-effect free (~5 lines). Your call whether that
lands here or as a follow-up. Related and worth a line in the comment: the root still grows 153
files/run forever (nothing in this PR bounds it), and the ledger is now O(all requests in the root)
per test — mine passed 2000 files during this review.

### Claims

1. **"All 16 call sites"** — miscount. `grep -c countPendingReindex` on the parent gives 16
   occurrences in `stale_reindex_test.go` but one of those is the `func` definition: **15 call sites**
   (14 + 1 in the 6175 file), across 12 `newReindexLedger` constructions. All 15 are converted — no
   `countPendingReindex` survives (it could not, the symbol is gone). ✅ on substance, ✗ on the number.
2. **Siblings** — partly wrong, in the safe direction. The family is **33** `TestStaleReindexGuard_*`
   tests, not 16. More importantly `IdleWindowBetweenBatches` and `140RepoUpgrade_Simulation` do **not**
   share the defect: I read both, and neither makes *any* durable-count assertion — they assert
   in-memory admissions and idle-window runs only. They are the *polluters* (140RepoSim writes the
   140 files per run), not victims. The accidental-safety claim is right for the two that do assert:
   `CurrentRepo_NoRequest` and 6175's `DeclinedRepoIsNotRetriedForever` both assert `0` on a path
   nothing else writes — one shared fixture path away from breaking, and both are now converted, so
   the point is moot. No conversion is owed to IdleWindow/140RepoSim.
3. **Sweep** — verified and extended. `requests_drain_test.go`, `requests_drain_unblock_test.go`,
   `index_split_mode_test.go` all do `root := t.TempDir()` + `t.Setenv(EnvRoot, root)` + a
   `t.TempDir()` repo path: doubly isolated, confirmed line by line. I also swept the four
   `ListPending` users you did **not** name — `rebuild_split_mode_test.go`,
   `requests_drain_resume_test.go`, `rebuild_wait_test.go`, `status_rebuild_inflight_6014_test.go`.
   The last two count absolutely against a **constant** group name (`requestsDirForGroup("g")` /
   `("acme")`), which is the same shape as the defect — but every one of them reaches `EnvRoot =
   t.TempDir()`, `rebuild_wait_test.go` via the `withSplitWaitTest` helper. Clean, no gap.
4. **Mutant C** — confirmed exactly as reported: equivalent on a pristine root, DEAD on a reused one.
   That asymmetry is the defect's own premise and is also a useful positive control that my "reused"
   root really was reused.
5. **`stale_reindex.go` byte-identical** — verified by shasum against both `fc9124eef` and `32ccc28c4`. ✅

### Rebase

**Not needed.** Parent is `fc9124eef`; `main` is `2d6d3c97b` (3 commits ahead). Zero file overlap —
main touched `internal/daemon/{engineplane,server}.go`, `walk/`, `watch/`, `registry/`, `quality/`;
this PR touches only the two `stale_reindex*_test.go` files. I rebased onto `origin/main` anyway to
be sure: clean, and the **full** `./internal/daemon/` suite is green on both roots after it —
reused root exit 0 (34.5s), fresh root exit 0 (31.7s), `gofmt -l` clean, `go vet` exit 0.

### Verdict

**APPROVE.** The diagnosis is right and one step deeper than the issue; the arithmetic reproduces to
the file; the rival mechanism is genuinely ruled out; the assertion is scoped, not weakened; and the
guard demonstrably still fails, identically on both roots, under four independent mutants including
one that attacks the uuid premise the design rests on. Before merge I'd like the two body corrections
(15 not 16 call sites; IdleWindow/140RepoSim don't share the defect, they cause it) and a decision on
the `ListPending` write side-effect.

cajasmota:
Coordinator mutant on the final head `c0ce5f210`. One mutant, **ALIVE**, and it refines the review's conclusion rather than contradicting it.

## MD-1 — drop BOTH filters at once

`if r.Kind == requests.KindReindex && r.RepoPath == repoPath` → `if true`, in `pendingReindexIDs`. Anchor count 1 pre-edit, 0 post-edit, `if true {` 1 post-edit, `go vet` clean, `-count=1`, restored by `cp` from a shasum-verified backup (`47c8ce0d9…`), `git status --porcelain` empty.

I picked it because the review scored **M2** (drop `RepoPath`) and **M5** (drop `Kind`) separately, both ALIVE, and concluded "redundant defensive filters". Two guards that are only ever exercised together grade neither, so a compound mutant is the one that settles it.

**Verdict: ALIVE on a fresh root AND on a reused root** (`rc=0` both). So the compound behaves as the parts did — no masking. The review's empirical result stands.

## But "redundant" is the wrong word for one half of it

The `RepoPath` half is genuinely redundant: the requests directory is already keyed per repo hash, so a request for another repo cannot appear in the set being iterated. Nothing can distinguish it.

The `Kind` half is a different matter, and I do not think it should be described as defensive. `internal/daemon` writes **`requests.KindRebuild`** in production at `service.go:717`, and **`requests.KindCancelGroup`** at `service.go:1206`, plus `KindRebuild` from at least four tests (`rebuild_split_mode_test.go:190`, `status_rebuild_inflight_6014_test.go:171`, `requests_drain_resume_test.go:43,97,141`). A rebuild request written **for the same repo path** lands in the **same directory** — so the `Kind` filter is load-bearing for that input; it is simply that no current test produces it against the stale-reindex fixtures' constant fake repos (`/repo/stale` and friends).

So the precise verdict is **equivalent under the current suite, with a named distinguishing input**: any test that writes a `KindRebuild` for one of these repo paths. That is not exotic — it is the ordinary next test somebody writes in this package, and this PR exists precisely because these fixtures use constant paths that outlive a run.

**No change requested.** Keep the filter — it costs nothing and it is the reason such a test would not silently corrupt these counts. I am recording the distinction because "redundant defensive filter" invites a future cleanup that deletes it, and the deletion would be invisible until the day it was not. A guard that is unreachable *today* because of a fixture's choice of constants is not the same as one that is unreachable by construction.

## On the purity fix

`newReindexLedger` now snapshots with `os.ReadDir` and strips `.request.json` — correct, and it makes the point that `ListPending` was never needed for a baseline, since the ID is the filename stem.

The part worth calling out is **mutant E**, the positive control: a helper whose sweep reads nothing passes the survival assertion trivially, because a file nobody looked for is a file nobody deleted. Asserting that the baseline actually observed the planted ID is what makes the purity claim mean something. That is the same vacuity shape as a scan-and-assert-absence guard, and catching it unprompted is the right instinct.

Merging on green. Refs #6929.

Issue #6942, opened 2026-09-06
`except <mod>.<Class>` binds to the real class on a full rebuild and to a synthesised `SCOPE.ExceptionType` incrementally — masked until #6927 removed the junk entity causing the collision


cajasmota:
`except <mod>.<Class>` binds to the real class on a full rebuild and to a synthesised `SCOPE.ExceptionType` on the incremental path — the divergence was masked by a junk entity for four months

Found while landing the python arm of #6927. Not caused by it: #6927 only removed the thing that was hiding it.

## The divergence

`cmd/grafel/incremental_content_parity_6129_test.go`'s fixture has

```python
# cperrs_static.py   (never touched between passes)
class CpNotFound(Exception):
    pass

# cphandler_delta.py (in the delta)
except cperrs_static.CpNotFound:
```

With `internal/engine/rules/python/frameworks/sqlalchemy.yaml`'s Model pattern gated (see below), the two paths disagree on four rows:

```
[EDGE-LOST]       SCOPE.Operation/cp_handle@cphandler_delta.py→SCOPE.Component/CpNotFound@cperrs_static.py:CATCHES
[EDGE-INVENTED]   SCOPE.Operation/cp_handle@cphandler_delta.py→SCOPE.ExceptionType/exception:CpNotFound@<exception>:CATCHES
[EDGE-INVENTED]   Module/test-repo@→SCOPE.ExceptionType/exception:CpNotFound@<exception>:CONTAINS
[ENTITY-INVENTED] SCOPE.ExceptionType|exception:CpNotFound|scope:exceptiontype:CpNotFound|<exception>
```

LOST = present on the full rebuild and absent incrementally. So the **full** rebuild resolves the caught exception to the real class in the untouched file; the **incremental** run does not find it and synthesises a placeholder `SCOPE.ExceptionType` anchored at the `<exception>` pseudo-file instead.

Which side is *right* is a separate question — a dedicated exception kind may well be the better model — but they are not the same graph, and a query joining on the CATCHES endpoint sees different rows depending on how the repo was indexed.

## Why nobody saw it

[...]

Issue #6927, opened 2026-09-06
DONE: all 18 patterns shipped and both dead-on-arrival arms are now demonstrably live on real repos


cajasmota:
18 more `^`-anchored rule patterns in 7 files are compiled without `(?m)` — and two of them would OVER-fire if `(?m)` were simply switched on

Split out of #6917, which fixed `sinatra.yaml` (9 patterns) and left the rest deliberately alone. This issue carries the measured remainder.

## The sweep

Structural, not grep: every rule YAML under `internal/engine/rules/<lang>/{frameworks,orms,queues}/` decoded through `engine.FrameworkRule` — the same schema `internal/engine/detector.go` compiles at `:156,177` — so the sweep population **is** the compiled population. **520 rule files, 412 patterns.**

**27 `^`-anchored patterns with no `(?m)`, in 8 files across 7 languages.** #6917 fixed the 9 in sinatra. The remaining **18 in 7 files**:

| file | count |
|---|---|
| `cicd/frameworks/github_actions.yaml` | 6 |
| `docker/frameworks/docker_compose.yaml` | 3 |
| `ansible/frameworks/ansible_core.yaml` | 2 |
| `python/frameworks/pytest.yaml` | 2 |
| `python/frameworks/sqlalchemy.yaml` | 2 |
| `scala/frameworks/play_framework.yaml` | 2 |
| `graphql/frameworks/graphql_schema.yaml` | 1 |

Every one of them reads as a *line* anchor — `^def\s+(test_\w+)\s*\(` obviously wants every test function in a file, not the first line of one; `^-\s+name:` wants every ansible task; `^(?:GET|POST|…)\s+(\S+)\s+controllers\.` wants every row of a Play routes file. So the recall loss is the same defect as #6917's, in six more languages.

## Why this is NOT a one-line compile-site fix

`(?m)` changes `$` as well as `^`. The sweep also found **4 `$`-bearing patterns without `(?m)`**:

- `docker/frameworks/docker_compose.yaml` — `^\s{2}(\w[\w-]*):\s*$`
- `ansible/frameworks/ansible_core.yaml` — `\s+(\w+\.\w+\.\w+):\s*$`
- `ansible/frameworks/ansible_core.yaml` — `^\s+(apt|yum|dnf|pip|copy|template|…):\s*$`
- `csharp/frameworks/net_maui.yaml` — `(?:public|internal)\s+(?:partial\s+)?class\s+(\w+ViewModel)\s*(?::|$)`

[...]

PR #6943, opened 2026-09-06
fix(#6927): pytest and sqlalchemy `^` anchors were start-of-TEXT — a realistic python module extracted ZERO tests and ZERO models

cajasmota:
The **python arm** of #6927: `pytest.yaml` and `sqlalchemy.yaml`, 4 of the 18 remaining `^`-anchored patterns compiled without `(?m)`.

`internal/engine/detector.go` compiles rule patterns with plain `regexp.Compile` at `:156,177`, so `^` means start-of-**text**. `^def\s+(test_\w+)\s*\(` could only fire when a test function was the FIRST thing in a file. Every realistic pytest module opens with a docstring or an import block, so it extracted **nothing** — not degraded, absent. Same for `^class\s+(Test\w+)` and sqlalchemy's two class patterns.

## Red, then green

Driven end-to-end on the new golden fixture `python-pytest-sqlalchemy-mini`, same binary path both times (`grafel quality`, isolated `GRAFEL_HOME`/`GRAFEL_DAEMON_ROOT`/`HOME`):

| | pre-fix binary | post-fix |
|---|---|---|
| entity must-have recall | **3 / 10 (30.0%)** | **10 / 10 (100%)** |
| relationship must-have recall | **0 / 2 (0.0%)** | **2 / 2 (100%)** |
| forbidden entity hits | 0 | 0 |
| exit | 2 | 0 |

The 3 rows the pre-fix binary did find are the fixture's deliberate **controls** — `Fixture customer`, `Relationship Invoice`, `Constraint customers.id` — all from un-anchored patterns, green before and after. Without them "the rules came back" and "rule loading broke entirely" would look the same.

At the engine level, `TestIssue6927_RealisticPytestModuleYieldsTests` and `TestIssue6927_SQLAlchemyModelsYieldSchemaAndModels` fail on the parent commit with zero `Test`, zero `TestClass`, zero `Schema`, zero `Model`.

## Why `(?m)` per pattern, and why not only `(?m)`

**Per-pattern, not the compile site.** #6917 rejected the compile-site variant because `(?m)` changes `$` too. Checked pattern by pattern here: **none of the four carries a `$`**, so that second effect cannot reach them. The compile-site cleanup stays a later, no-op step.

**`requires_framework: true` on the Model pattern.** Read literally, `^class\s+([A-Z]\w*)\s*\(\w+\)\s*:` describes "a module-level class with exactly one simple base" — plain Python syntax, not an ORM construct. `class ConfigError(Exception):` and `class Serializer(object):` both satisfy it, and `Detect` resolves rule sets by `file.Language` alone, so `(?m)` **without a gate would type every such class in every indexed Python file as a `Model`**. The start-of-text anchor was holding that back by accident. This is the shape #6152 already had to gate for `falcon.yaml` and `cherrypy.yaml` (`class Foo:` → Controller) and it carries the same remedy. The trade is stated in the YAML: a models module that imports its `Base` from a sibling and never names sqlalchemy loses the entity (the `*/models.py` file_convention still types the file).

The other three patterns are left ungated on purpose: the `DeclarativeBase` literal is its own evidence, and gating pytest would lose the very common plain-`assert` module that imports nothing.

## Forbidden rows, each proven capable of firing

A recall assertion is structurally blind to over-firing (#6902), and `(?m)` is precisely the change that lets `^` reach a non-first line for the first time — so over-firing only becomes *possible* with this fix. Seven forbidden rows, and **none is decoration**: each was fired by an actual mutant (#6938's trap — a fence written against a spelling the producer never emits reads green forever).

| forbidden row | held by | mutant that fires it |
|---|---|---|
| `Test test_commented_out` | line anchor (`# def …`) | M5 |
| `Test test_nested_helper` | line anchor (indented in a `def`) | M5 |
| `TestClass TestCommentedOut` | line anchor (`# class …`) | M6 |
| `Model LegacyInvoice` | line anchor, **inside a file that DOES import sqlalchemy** | M7 |
| `Model ConfigError` | framework gate | M8 |
| `Model Serializer` | framework gate | M8 |
| `Schema ConfigError` | the `DeclarativeBase` literal | M9 |

`LegacyInvoice` is deliberately in `app/models.py` rather than the un-imported file: the `serializers.py` rows are kept out by the **gate** and say nothing about the **anchor**, so the two fences are graded separately and their mutants are distinguishable.

## Vertical position is the axis varied

Nothing in the fixture sits on line 1. Tests appear after a docstring + imports + fixture, one indented inside a class, one **last in the file**; `Invoice` is the last model; `Registry` lives in a third file. The constructs are held at their idiomatic spellings.

## Mutants — scored at BOTH levels (#6926)

`internal/engine` package suite (`-count=1`) and the golden-fixture gate. **All nine DEAD at both.** Every mutant's pre-image count was asserted `== 1` before writing, so a silently-unapplied edit cannot read as ALIVE.

| # | mutation | engine suite | fixture gate |
|---|---|---|---|
| M1 | pytest `Test`: drop `(?m)` | FAIL | 8/10, exit 2 |
| M2 | pytest `TestClass`: drop `(?m)` | FAIL | 9/10, exit 2 |
| M3 | `DeclarativeBase`: drop `(?m)` | FAIL | 9/10 (`Registry` missing), exit 2 |
| M4 | `Model`: drop `(?m)` | FAIL | 7/10 + 0/2 edges, exit 2 |
| M5 | pytest `Test`: drop `^` | FAIL | 2 forbidden hits |
| M6 | pytest `TestClass`: drop `^` | FAIL | 1 forbidden hit |
| M7 | `Model`: drop `^` | FAIL | 1 forbidden hit |
| M8 | `Model`: remove `requires_framework` | FAIL | 2 forbidden hits |
| M9 | widen the `DeclarativeBase` literal to `[^)]*` | FAIL | 1 forbidden hit |

M3 is worth calling out. It was **ALIVE at the fixture level** on the first draft: in `models.py` both class patterns match `class Base(DeclarativeBase):` and the downstream class-shadow fold collapses the result, so removing `(?m)` there changed *nothing* observable — identical 57 entities, identical recall. `src/app/db.py` was added for that reason: `class Registry(MappedAsDataclass, DeclarativeBase):` has a two-name base list the Model pattern's `\(\w+\)` cannot match, so that `Schema` entity has exactly one producer and the pattern is graded.

## The latent bug the widening made reachable

Two, and they went opposite ways.

1. **`Base` is minted twice**, as `Schema` *and* as `Model` — the Model pattern's `(\w+)` accepts `DeclarativeBase` as a single simple base. Unreachable while the anchor kept both patterns dead. Pinned as a must-have row + a positive assertion so a later narrowing is deliberate, not accidental.
2. **`Test\w+` is not an over-fire.** It accepts `class Testing:` and `class TestHarness(Base):`, which looks wrong until you check pytest: `python_classes` defaults to the glob `Test*`. Left as written rather than narrowed to a spelling pytest itself does not use. Recorded in the YAML so the next reader does not "fix" it.

Known limitations are pinned in the **positive** direction, the #6917 heredoc idiom: an indented `def test_…` method and test-shaped text inside a `"""…"""` literal. Written as assertions so whoever teaches the detector about string literals (or widens the anchor to `^\s*def`) is told by a red test to come move them into the forbidden list.

## Ratchet — enumerated, not absorbed

`scripts/quality/run.sh --ratchet` → **OK, 33 gated fixtures held their baseline (24 at 100% must-have recall)**. `baseline.json` gains **exactly one entry**, hand-written:

```
+    "python-pytest-sqlalchemy-mini": {
+      "entity_found": 10, "entity_expected": 10,
+      "relationship_found": 2, "relationship_expected": 2,
+      "entity_extracted_total": 62, "relationship_extracted_total": 157
+    },
```

The three known python drifts (#6898) showed up again and were **restored, not recorded**:

```
python-django-mini    relationship_extracted_total 255 -> 251
python-fastapi-mini   entity_extracted_total 28 -> 27, relationship_extracted_total 57 -> 54
python-rq-mini        entity_extracted_total 22 -> 20, relationship_extracted_total 43 -> 39
```

They are **not** this change's doing, and that is measured rather than assumed: the same `--ratchet --runs 1` sweep with a **pre-fix** binary produces the identical three deltas. A post-fix-vs-pre-fix comparison across all 33 reports shows movement in **exactly one fixture** — the new one:

```
python-pytest-sqlalchemy-mini  entity_found 3->10  relationship_found 0->2
                               entity_extracted_total 54->62  relationship_extracted_total 142->157
```

Zero movement on the other 32. Recording the drifts would have tightened someone else's growth-only ceiling.

## Filed, not absorbed: #6942

`TestContentParity_PathA_6129` in `cmd/grafel` went red, and the cause is worth reading rather than skipping.

`cperrs_static.py` in that fixture opens with `class CpNotFound(Exception):` **on line 1** — the one position start-of-text semantics can reach — so the ungated Model pattern minted a junk `Model|CpNotFound` beside the real `SCOPE.Component|CpNotFound` in the same file. Two same-named entities in one file made the reference ambiguous, so the **full** rebuild also declined to bind `except cperrs_static.CpNotFound` and also fell back to a synthesised `SCOPE.ExceptionType` — and the two paths **agreed, for the wrong reason**. Gate the pattern, the junk entity goes, full starts resolving to the real class, and the incremental path's failure to do so becomes visible.

Isolated by measurement: with `(?m)` applied and `requires_framework` **removed**, that test is green; with the gate on it reports exactly four rows. Filed as **#6942**, with four shape-keyed `cpKnown` entries citing it. The allow-list is a shrink-only ratchet, so whoever fixes it is told by a stale-entry failure to delete them.

Also bumped: three `internal/quality` corpus-size constants `32 -> 33` (`want 32` guards that exist precisely so a corpus change is deliberate).

## Verification

- `go test ./internal/engine/... ./internal/quality/... ./cmd/grafel/... -count=1` → **all ok** (`internal/engine` 41.9s, `internal/quality` 34.6s, `cmd/grafel` 138.9s).
- `scripts/quality/run.sh --ratchet` → **exit 0**, 33 gated fixtures.
- `gofmt -l internal cmd` clean; `go vet` on the three trees clean.
- Measured in the linked worktree at `lane/6927-python-multiline`, parent `fc9124eef`.

## What this arm does NOT do

`Refs #6927` — it does not close it. **Five files remain**, 14 patterns:

| file | patterns |
|---|---|
| `cicd/frameworks/github_actions.yaml` | 6 (no fixture yet) |
| `docker/frameworks/docker_compose.yaml` | 3 |
| `ansible/frameworks/ansible_core.yaml` | 2 |
| `scala/frameworks/play_framework.yaml` | 2 (fixture-covered) |
| `graphql/frameworks/graphql_schema.yaml` | 1 (fixture-covered) |

`docker_compose.yaml` and `ansible_core.yaml` are **untouched on purpose**. They carry the `$`-bearing over-fire hazard #6927 measured: under `(?m)`, `^\s{2}(\w[\w-]*):\s*$` becomes "every two-space-indented key in any YAML the docker bucket sees". Those patterns need rewriting, not a flag, and they have no golden fixture to grade the rewrite. Separate arm.

Refs #6927, #6917, #6902, #6152, #6942.

### DIFF
--- a/cmd/grafel/incremental_content_parity_6129_test.go
+++ b/cmd/grafel/incremental_content_parity_6129_test.go
@@ -783,6 +783,54 @@ var cpKnownPathA = []cpKnown{
 		Contains: []string{"SCOPE.Operation/cp_use_class@cphandler_delta.py→«unbound»:CALLS"},
 	},
 
+	// ── #6942 family: `except <mod>.<Class>` binds differently on the two paths ──
+	//
+	// The FULL rebuild resolves `except cperrs_static.CpNotFound` to the real
+	// class entity in the untouched file; the INCREMENTAL run does not find it
+	// and synthesises a placeholder `SCOPE.ExceptionType` at the `<exception>`
+	// pseudo-file instead. Which side is right is genuinely open — a dedicated
+	// exception kind may be the better model — but they are not the same graph.
+	//
+	// These four entries are NEW here and the reason they are new is worth
+	// keeping, because it is not "a fix regressed something". Until #6927 the
+	// two paths AGREED, for the wrong reason: `cperrs_static.py` opens with
+	// `class CpNotFound(Exception):` on LINE 1, and sqlalchemy.yaml's
+	// `^class\s+([A-Z]\w*)\s*\(\w+\)\s*:` -> Model was compiled without
+	// `(?m)`, so start-of-text was the one position it could reach and it minted
+	// a junk `Model|CpNotFound` beside the real `SCOPE.Component|CpNotFound` in
+	// the same file. Two same-named entities in one file made the reference
+	// ambiguous, so the full rebuild ALSO declined to bind and also fell back to
+	// the synthetic. #6927 gates that pattern behind `requires_framework`, the
+	// junk entity goes, full starts resolving, and the incremental gap shows.
+	//
+	// Isolated by measurement rather than argued: with `(?m)` applied and
+	// `requires_framework` removed from that one pattern, this test is green;
+	// with the gate on, it reports exactly these four rows.
+	{
+		Issue:    "#6942",
+		Why:      "Incremental cannot resolve the caught exception to the class in an untouched file and synthesises a SCOPE.ExceptionType placeholder instead. Unfixed; which path is correct is undecided.",
+		Bucket:   cpEntityInvented,
+		Contains: []string{"SCOPE.ExceptionType|exception:CpNotFound", "<exception>"},
+	},
+	{
+		Issue:    "#6942",
+		Why:      "The CATCHES edge onto that synthesised placeholder.",
+		Bucket:   cpEdgeInvented,
+		Contains: []string{"SCOPE.Operation/cp_handle@cphandler_delta.py→SCOPE.ExceptionType/exception:CpNotFound@<exception>:CATCHES"},
+	},
+	{
+		Issue:    "#6942",
+		Why:      "Module CONTAINS for the synthesised placeholder. Listed separately from the CATCHES edge so one of them going quiet cannot be absorbed by the other.",
+		Bucket:   cpEdgeInvented,
+		Contains: []string{"→SCOPE.ExceptionType/exception:CpNotFound@<exception>:CONTAINS"},
+	},
+	{
+		Issue:    "#6942",
+		Why:      "The LOST half: the full rebuild's CATCHES edge, bound to the real class in the untouched file. This row is the whole of the disagreement — it is what incremental does not produce.",
+		Bucket:   cpEdgeLost,
+		Contains: []string{"SCOPE.Operation/cp_handle@cphandler_delta.py→SCOPE.Component/CpNotFound@cperrs_static.py:CATCHES"},
+	},
+
 	// ── #6094 family: a duplicated row persisted into graph.fb ── ALL FIXED ──
 	//
 	// This block held three entries and now holds none. They came off the
--- a/internal/engine/python_multiline_anchor_6927_test.go
+++ b/internal/engine/python_multiline_anchor_6927_test.go
@@ -1,0 +1,364 @@
+package engine
+
+import (
+	"context"
+	"testing"
+
+	"github.com/cajasmota/grafel/internal/extractor"
+)
+
+// #6927 (python arm) — the same defect #6917 fixed for sinatra, in two Python
+// rule files. `internal/engine/detector.go` compiles rule patterns with plain
+// `regexp.Compile` at :156 and :177, so `^` means START OF TEXT, not start of
+// line. Four `^`-anchored patterns are affected:
+//
+//	rules/python/frameworks/pytest.yaml
+//	  ^def\s+(test_\w+)\s*\(          -> Test
+//	  ^class\s+(Test\w+)\s*           -> TestClass
+//	rules/python/frameworks/sqlalchemy.yaml
+//	  ^class\s+(\w+)\s*\([^)]*DeclarativeBase[^)]*\)\s*:  -> Schema
+//	  ^class\s+([A-Z]\w*)\s*\(\w+\)\s*:                   -> Model
+//
+// None of the four carries a `$`, so `(?m)`'s second effect (end-of-line
+// semantics for `$`) cannot reach them — checked pattern by pattern, which is
+// why the fix is per-pattern here and NOT at the compile site (#6917 rejected
+// the compile-site variant because this repo has ungated `$`-bearing patterns
+// elsewhere, notably docker_compose and ansible_core).
+//
+// These tests drive the REAL embedded YAML through LoadAllRules() and
+// Detector.Detect(): the defect IS the shipped pattern text, so grading a
+// hand-written copy would leave the shipped one unobserved.
+//
+// Graded in BOTH directions (#6902 — a recall assertion is structurally blind
+// to over-firing):
+//
+//   - recall:    TestIssue6927_RealisticPytestModuleYieldsTests,
+//     TestIssue6927_SQLAlchemyModelsYieldSchemaAndModels
+//   - forbidden: TestIssue6927_PythonPatternsDoNotOverFire,
+//     TestIssue6927_ModelPatternDoesNotClaimEveryPythonClass
+
+// pytest6927Module is deliberately NOT a toy whose first line is a test. It
+// opens with a docstring and an import block — the shape every real pytest
+// module has — then places tests at three DIFFERENT vertical positions: after
+// the fixture, inside a class, and last-in-file. Vertical position IS the
+// defect, so it is the one axis this source varies; the constructs themselves
+// are held at their idiomatic spellings.
+const pytest6927Module = `"""Invoice service tests."""
+
+import pytest
+
+from app.invoices import InvoiceService
+
+
+@pytest.fixture
+def service():
+    return InvoiceService()
+
+
+def test_creates_invoice(service):
+    assert service.create()
+
+
+class TestInvoiceTotals:
+    def test_sums_lines(self, service):
+        assert service.total() == 0
+
+
+def test_rejects_negative_amount(service):
+    with pytest.raises(ValueError):
+        service.create(amount=-1)
+`
+
+// sqlalchemy6927Models likewise opens with a docstring + imports, and declares
+// its last model near EOF.
+const sqlalchemy6927Models = `"""ORM models."""
+
+from sqlalchemy import Column, ForeignKey, Integer, String
+from sqlalchemy.orm import DeclarativeBase, relationship
+
+
+class Base(DeclarativeBase):
+    pass
+
+
+class Customer(Base):
+    __tablename__ = "customers"
+
+    id = Column(Integer, primary_key=True)
+    name = Column(String)
+    invoices = relationship("Invoice")
+
+
+# class LegacyInvoice(Base):
+#     __tablename__ = "legacy_invoices"
+
+
+class Invoice(Base):
+    __tablename__ = "invoices"
+
+    id = Column(Integer, primary_key=True)
+    customer_id = Column(Integer, ForeignKey("customers.id"))
+`
+
+func detect6927Python(t *testing.T, path, src string) *DetectResult {
+	t.Helper()
+	rules, err := LoadAllRules()
+	if err != nil {
+		t.Fatalf("LoadAllRules: %v", err)
+	}
+	det := New(rules)
+	res, err := det.Detect(context.Background(), extractor.FileInput{
+		Path:     path,
+		Content:  []byte(src),
+		Language: "python",
+	})
+	if err != nil {
+		t.Fatalf("Detect: %v", err)
+	}
+	return res
+}
+
+func has6927Entity(res *DetectResult, kind, name string) bool {
+	for _, e := range res.Entities {
+		if e.Kind == kind && e.Name == name {
+			return true
+		}
+	}
+	return false
+}
+
+func entity6927Names(res *DetectResult, kind string) []string {
+	var out []string
+	for _, e := range res.Entities {
+		if e.Kind == kind {
+			out = append(out, e.Name)
+		}
+	}
+	return out
+}
+
+// TestIssue6927_RealisticPytestModuleYieldsTests is the pytest recall arm.
+// Before the fix it found ZERO Test and ZERO TestClass entities: the module
+// opens with a docstring, so nothing could satisfy start-of-TEXT.
+func TestIssue6927_RealisticPytestModuleYieldsTests(t *testing.T) {
+	res := detect6927Python(t, "tests/test_invoices.py", pytest6927Module)
+
+	for _, want := range []string{"test_creates_invoice", "test_rejects_negative_amount"} {
+		if !has6927Entity(res, "Test", want) {
+			t.Errorf("Test %q not extracted; Tests found: %v", want, entity6927Names(res, "Test"))
+		}
+	}
+	if !has6927Entity(res, "TestClass", "TestInvoiceTotals") {
+		t.Errorf("TestClass TestInvoiceTotals not extracted; TestClasses found: %v",
+			entity6927Names(res, "TestClass"))
+	}
+
+	// CONTROL, held constant across the change: the fixture pattern
+	// `@pytest\.fixture(?:\s*\([^)]*\))?\s*\ndef\s+(\w+)\s*\(` is NOT
+	// `^`-anchored, so it already worked and must keep working. Without it a
+	// total-failure mode (rules stop loading at all) would be indistinguishable
+	// from the recall win above.
+	if !has6927Entity(res, "Fixture", "service") {
+		t.Errorf("CONTROL FAILED: the un-anchored @pytest.fixture pattern produced no "+
+			"Fixture entity; Fixtures found: %v — this is not the #6927 defect, "+
+			"rule loading itself is broken", entity6927Names(res, "Fixture"))
+	}
+
+	// KNOWN LIMITATION, pinned in the POSITIVE direction (the #6917 heredoc
+	// idiom): `^def` requires the `def` to open the line, so a test METHOD
+	// indented inside a class is still not extracted. `(?m)` does not change
+	// that and this change deliberately does not widen the anchor to `^\s*def`
+	// — that is a second behaviour change with its own over-fire direction
+	// (every nested helper `def test_...` in non-test code) and belongs in its
+	// own issue with its own forbidden rows. Written as an assertion rather
+	// than a comment so whoever does widen it is told by a red test to come
+	// here and move this case up into the recall list.
+	if has6927Entity(res, "Test", "test_sums_lines") {
+		t.Errorf("indented test methods are now extracted (Test %q) — good news; "+
+			"move this case into the recall assertions above", "test_sums_lines")
+	}
+}
+
+// TestIssue6927_SQLAlchemyModelsYieldSchemaAndModels is the sqlalchemy recall
+// arm. Before the fix: ZERO Schema and ZERO Model entities from the class
+// patterns.
+func TestIssue6927_SQLAlchemyModelsYieldSchemaAndModels(t *testing.T) {
+	res := detect6927Python(t, "app/models.py", sqlalchemy6927Models)
+
+	if !has6927Entity(res, "Schema", "Base") {
+		t.Errorf("Schema Base (class Base(DeclarativeBase)) not extracted; Schemas: %v",
+			entity6927Names(res, "Schema"))
+	}
+	for _, want := range []string{"Customer", "Invoice"} {
+		if !has6927Entity(res, "Model", want) {
+			t.Errorf("Model %q not extracted; Models found: %v", want, entity6927Names(res, "Model"))
+		}
+	}
+
+	// CONTROLS, held constant: neither pattern is `^`-anchored, so both were
+	// green before the fix and must be green after it.
+	if !has6927Entity(res, "Relationship", "Invoice") {
+		t.Errorf("CONTROL FAILED: un-anchored `= relationship(\"Invoice\")` produced no "+
+			"Relationship entity; found: %v", entity6927Names(res, "Relationship"))
+	}
+	if !has6927Entity(res, "Constraint", "customers.id") {
+		t.Errorf("CONTROL FAILED: un-anchored `ForeignKey(\"customers.id\")` produced no "+
+			"Constraint entity; found: %v", entity6927Names(res, "Constraint"))
+	}
+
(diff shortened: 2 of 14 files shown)


cajasmota:

Reviewed in my own detached worktree at `45fbfd221` (`review-6943-zx71/wt`), isolated `HOME`/`GRAFEL_HOME`/`GRAFEL_DAEMON_ROOT`, `-count=1` throughout, every mutant restored by `cp` from a shasum-verified backup (final tree verified clean against HEAD).

The core of this PR holds up under independent measurement, and the `requires_framework` argument is **understated**, not overstated. One blocking defect and two ungraded axes.

---

### BLOCKING — the 32→33 bump collides with #6938 and git merges it *silently*

`main` is now `2d6d3c97b`; #6938 (`2de485618`) added `ocaml-objects-mini` and bumped the **same three** corpus-size guards 32→33. This PR also bumps them 32→33. Both sides made the *identical* edit on the *same lines*, so there is **no conflict** — GitHub reports `MERGEABLE` and the merge is clean. The corpus becomes 34; the guards still say 33.

Proven, not argued — merged `origin/main` into a checkout of this head and ran the three guards:

```
merge exit=0   (clean, no conflict)
fixtures on disk: 34

--- FAIL: TestEveryGoldenFixtureDeclaresExpectedRelationships_6490
    inspected 34 golden fixtures, want 33
--- FAIL: TestOnlyTheIntendedGoldenRowsAssertSubtype_6488
    parsed 34 fixtures, want 33
--- FAIL: TestNoGoldenFixtureClaimsTheOptIn_6490
    inspected 34 golden fixtures, want 33
FAIL github.com/cajasmota/grafel/internal/quality
```

**Rebase onto `2d6d3c97b` and make the three constants `33 -> 34`.** `baseline.json` merges correctly on its own (34 entries, both new fixtures present) — only the three counters are wrong. This is the exact "identical-change rebase drops silently" trap: a clean merge is *not* evidence here.

---

### The `requires_framework` gate IS load-bearing — and by more than the PR claims

Measured independently against `~/Projects/archigraph-corpora` (3365 real `.py` files), applying the widened pattern with and without the marker gate:

| | matches | files |
|---|---|---|
| `(?m)^class\s+([A-Z]\w*)\s*\(\w+\)\s*:` **ungated** | **4291** | 1445 |
| same, gated by `sqlalchemy.yaml`'s `import_markers` | **1** | 1 |

A ~4300x reduction. Shipping `(?m)` without the gate would have typed 43% of the corpus's Python files as containing `Model`s. I also read `Detect` rather than inferring: `frameworkPresent` (detector.go:96) is a literal substring scan of the whole file over `frameworks.detection.import_markers`, evaluated once per (ruleset, file) at :402 and consumed at :409 — it gates exactly as described.

Independently reproduced the red→green measurement with a pre-fix binary built from `fc9124eef`'s two YAMLs against the new fixture: **3/10 entities, 0/2 relationships, extracted 54/142**, and the 3 found are exactly the three un-anchored controls (`Fixture customer`, `Relationship Invoice`, `Constraint customers.id`). Post-fix: **10/10, 2/2, 62/157** — matches the PR body digit for digit.

Verified the `$` claim by reading all four patterns: none carries a `$`. Correct.

---

### My mutants (5, none of the author's 9), scored at both levels

| # | mutation | engine suite | fixture gate | verdict |
|---|---|---|---|---|
| R1 | **call site**: `if sp.requiresFramework && !frameworkSeen` → `if false && …` (detector.go:409, pre-count asserted `==1`) | FAIL ×3 incl. `TestIssue6927_ModelPatternDoesNotClaimEveryPythonClass`, `TestBarePythonClass_IsNotController_6152`, `TestVBNet6537_ModuleGateRequiresFrameworkMarker` | 8/10, **2 forbidden entity hits**, exit 2 | **DEAD** — the mechanism is graded, not just the YAML key |
| R2 | Model pattern `\(\w+\)` → `\([^)]*\)` (any base list) | **ok — GREEN** | 9/10, exit 2, **forbidden hits 0**; missing `Registry [Schema]` | **distinguishable and ungraded** (see below) |
| R3 | pytest `(?m)^class\s+(Test\w+)` → `(?m)^\s*class\s+(Test\w+)` | **ok — GREEN** | **exit 0, 10/10, 2/2, 0 forbidden** | **ALIVE at both levels** |
| R4 | delete one of the four `#6942` `cpKnown` entries (the CONTAINS row) | — | `TestContentParity_PathA_6129` FAIL, reports exactly the one dropped `[EDGE-INVENTED] …:CONTAINS` | **DEAD** — entries are minimal, not padded |
| R5 | add a 5th `cpKnown` entry matching nothing | — | FAIL: `allow-list entry for #6942 … matched NOTHING` | **DEAD** — the shrink-only ratchet really does force deletion |

#### R3 is the finding worth acting on

`TestIssue6927_PythonPatternsDoNotOverFire` pins the *def* anchor on **two** axes — `#`-commented (`test_commented_out`) and **indented** (`test_nested_helper`, with the note "widening it to `^\s*def` mints it"). The *class* anchor is pinned on only **one**: `TestCommentedOut`, held by the leading `#`. Nothing anywhere pins the class pattern's column-0 half. So `^class` → `^\s*class` is invisible to the 42s engine suite **and** to the fixture gate — and it is production-reachable (a nested `class TestX:` inside a class or a helper). This is the same axis the def pattern got right; the class pattern just lost its second row.

Fix: add an indented `class TestNestedHelper:` to `src/tests/helpers.py` + `decoy` const, and a `{"name": "TestNestedHelper", "kind": "TestClass", "must_exist": false}` forbidden row.

#### R2 — over-fire caught only by accident

Widening the Model base list mints `Model|Registry` on `class Registry(MappedAsDataclass, DeclarativeBase):`. The gate catches it, but with **`forbidden_entity_hits: 0`** — it fires only because the new `Model` shadows the `Schema` in the class-shadow fold, i.e. as a *recall* loss. `expected.json`'s own description says "a recall assertion cannot see a pattern that started matching too much (#6902)", and here it is the only thing seeing it. `db.py`'s docstring asserts "the Model pattern … cannot match this line" — an "only X" claim with no forbidden row behind it. Add `{"name": "Registry", "kind": "Model", "must_exist": false}` so the claim is graded in the direction it is made.

#### And a factual over-claim in `pytest.yaml`

> "That is not an over-fire — it is pytest's OWN collection rule, whose `python_classes` default is the glob `Test*`."

pytest's collection is **conjunctive**: `python_classes = Test*` applies only inside files matching `python_files = test_*.py *_test.py`. The rule as written drops the file half and applies `Test*` to every indexed `.py`. That is broader than pytest, so the justification does not hold as stated. It is not hypothetical — the corpus scan found a real hit:

```
TestConfig   flask-realworld/conduit/settings.py     -> typed TestClass
```

`class TestingConfig(Config):` / `class TestConfig(Config):` in a settings module is a Flask idiom, and this PR is what makes it reachable (pre-fix it needed to be on line 1). One hit in 3365 files, so low severity and I would not gate on it — but please reword the comment to say what is actually true ("broader than pytest's own collection, accepted because the file half is not available here"), or narrow it. Under [prose-asserts-what-no-test-observes] a comment that states a false "it is exactly pytest's rule" is worse than no comment. Same corpus scan found **zero** non-test-file hits for the `def test_*` pattern, so that one is genuinely safe.

---

### `cpKnown` allow-list: **honest, not masking**

- The causal story is factually grounded: `cpStatic` writes `cperrs_static.py` opening with `class CpNotFound(Exception):` **on line 1** (incremental_content_parity_6129_test.go:426) — the one position start-of-text could reach. The junk `Model|CpNotFound` colliding with the real `SCOPE.Component|CpNotFound` making the reference ambiguous is a coherent and checkable mechanism, and the pre-existing defect (incremental cannot bind an exception to a class in an untouched file) is unchanged by this PR — the PR only makes the *full* path stop failing in the same way. The two paths were agreeing for the wrong reason. **Pre-existing, unmasked — not introduced.**
- **Minimal and shape-keyed**: every entry carries a fully-resolved id with file attribution and edge kind. R4 proves none is padding — deleting one surfaces exactly one unexpected divergence.
- **The ratchet works**: R5 proves an entry that stops matching fails the test by name, so whoever fixes #6942 is forced to delete all four.

`#6942` is filed, OPEN, and its title states the defect accurately.

---

### Summary

- **Rebase required** — blocking, and the danger is that it merges *clean*. `33 -> 34` on all three guards after rebasing onto `2d6d3c97b`.
- **Add two forbidden rows** (R3's indented `TestClass`, R2's `Model|Registry`).
- **Reword the `python_classes` comment** in `pytest.yaml` — the claim as written is false.
- Everything else I attacked held: the gate is load-bearing (measured at 4291:1), the red→green numbers reproduce exactly, the `$` claim is correct, the allow-list is honest and ratcheted, and the mechanism-level mutant on the gate call site is dead at both levels.

Nothing merged, nothing pushed to the branch.

cajasmota:
**Coordinator note — do NOT merge this on its green checks. It is green now and would be red immediately after.**

Recording this on the PR rather than only in my own notes, because the signal a merger normally trusts is actively misleading here:

```
#6943  MERGEABLE/CLEAN   ALL GREEN
```

Both are true, and both stay true right up to the merge. #6938 (`2de485618`) landed `ocaml-objects-mini` and bumped the same three corpus-size guards **32 → 33**. This branch makes the **identical 32 → 33 edit on the same lines**. Git therefore merges it with **no conflict** — the two edits agree textually — while the post-merge corpus is **34** and the guards still say **33**. The reviewer proved it by merging `origin/main` into a checkout of this head: all three guards fail with `inspected 34 … want 33`.

So this is the identical-change trap in its most dangerous form. The usual version is a rebase that silently *drops* your change because someone else made the same one. This is the inverse: the change is silently *kept*, and it is now wrong, because what the constant means depends on a count neither branch owns.

**The general rule this is the second instance of: a clean merge is not evidence.** A shared counter must be re-derived from disk after any rebase or merge, never carried over from the diff — the diff is exactly where the information is missing.

Held with the author for a rebase onto `2d6d3c97b` with the constants moved to **34**. Two further items are outstanding (an ALIVE mutant on the ungraded column-0 half of the class anchor, and a factual over-claim in a rule comment about pytest's collection semantics); details in the review above.

Refs #6927, #6938.

cajasmota:
All four addressed, rebased onto `2d6d3c97b`. Head is now `50a72888c`.

## 1 — rebase + the identical-change trap (BLOCKER)

Rebased. Re-derived the corpus size **from disk, not from the diff**: 34 fixture directories under `internal/quality/golden/`, 34 entries in `baseline.json`. The three guards read **34**.

The trap is exactly as described and is worth leaving on the record: #6938's `32 -> 33` and mine were textually identical on the same lines, so git resolved them to one copy with no conflict while the merged corpus is 34. A clean merge was not evidence of anything.

## 2 — R3 was ALIVE at both levels (BLOCKER)

Confirmed and fixed. The `def` anchor was pinned on two axes (`#`-commented, indented) and the `class` anchor on only the commented one, so its column-0 half was graded by nothing.

`src/tests/helpers.py` now declares `class TestNestedHelper:` inside `build_client()`, forbidden row added, and the engine decoy carries the same. Re-scored:

```
M10  (?m)^class\s+(Test\w+) -> (?m)^\s*class\s+(Test\w+)
     engine suite: FAIL (TestIssue6927_PythonPatternsDoNotOverFire)
     fixture gate: exit 2, 1 forbidden hit — TestNestedHelper [TestClass]
```

M6 (dropping `^` entirely) now reports **two** forbidden hits where it previously reported one, which is the same gap showing from the other side.

## 3 — R2 distinguishable and ungraded

Confirmed, including your reading of why it hid: on `Registry` the widened capture mints a second entity for a class that already has a Schema entity, the class-shadow fold absorbs one of the pair, and the over-fire surfaces as an accidental **recall** loss with zero forbidden hits.

`src/app/db.py` now also declares `class SessionFactory(LoggingMixin, BaseFactory):` — two-name base list, no `DeclarativeBase`, no second producer — with `Model SessionFactory` forbidden:

```
M11  \(\w+\)\s*: -> \([^)]*\)\s*:   (Model pattern)
     engine suite: FAIL (TestIssue6927_ModelAnchorIsGradedInsideAFrameworkFile)
     fixture gate: exit 2, 1 forbidden hit — SessionFactory [Model] in app/db.py
                   (and the Registry recall loss alongside it, which is the fold
                    doing exactly what you described)
```

The db.py docstring's "the Model pattern cannot match this line" is gone. It now states which class pins which direction and why `Registry` alone could not.

## 4 — the factual over-claim

Reworded. `pytest.yaml` no longer says the breadth "is pytest's OWN collection rule". It now says the **name** half follows `python_classes`' `Test*` default, that pytest's collection is **conjunctive** (`python_files` AND `python_classes`), that this rule carries no file half at all because `source_patterns` are resolved by `file.Language` and have no way to express one, and records the measured consequence — `class TestConfig(Config):` in a settings module, 1 hit in 3365 files, against zero non-test hits for `^def test_`. Pattern unchanged, as you asked.

## Re-verification

- **Mutants re-scored after the fixture changed: eleven, all DEAD at BOTH levels** (engine package suite exit 1, fixture gate exit 2). Every pre-image count asserted `== 1` before writing; both YAML files restored and shasum-verified against their backups afterwards.
- `scripts/quality/run.sh --ratchet` → **exit 0, 34 gated fixtures held their baseline (25 at 100% must-have recall)**.
- **Ratchet delta re-enumerated post-rebase, not assumed.** Against `origin/main`'s baseline the only movements are the three known python drifts (#6898) plus the new fixture. A **pre-fix binary built at the same base** produces the identical three drifts, so they are still not this branch's doing; post-fix-vs-pre-fix across all 34 reports moves **exactly one fixture**:

```
python-pytest-sqlalchemy-mini  entity_found 3->10  relationship_found 0->2
                               entity_extracted_total 59->67  relationship_extracted_total 156->171
```

  `baseline.json`'s single entry was updated to the fixture's new totals (67 / 171) — the fixture grew by two decoy classes and a nested one.
- `go test ./internal/engine/... ./internal/quality/... ./cmd/grafel/... -count=1` → **all ok**.
- `gofmt -l internal cmd` clean, `go vet` clean.
- Measured in the linked worktree at `lane/6927-python-multiline`, parent `2d6d3c97b`.

Thanks for the corpus measurement on the gate — 4291 matches in 1445 files ungated vs 1 in 1 gated is a much sharper statement of the hazard than the fixture can make, and I have not tried to restate it as mine.

cajasmota:
Coordinator mutant on the final head `50a72888c`. One mutant, **DEAD** — but killed by something other than this arm's own grading, which is the part worth reporting.

## ME-1 — widen the NAME half, not the anchor

`(?m)^class\s+(Test\w+)\s*` → `(?m)^class\s+(\w+)\s*` in `pytest.yaml`. Anchor count 1, `go vet` clean, `-count=1`, restored by `cp` from a shasum-verified backup (`0543bcd10…`), `git status --porcelain` empty.

Every mutant so far — yours M1–M11, the review's R1–R5 — attacked the **anchor** axis (`^`, `(?m)`, indentation) or the framework gate. Nobody had touched the **name** axis, which is the other half of what makes this rule specific. It is also the permissive direction: `(\w+)` types *every* top-level class in a pytest-gated file as a `TestClass`.

**Verdict: DEAD.**

```
--- FAIL: TestZeroProducingKinds6776_NoSiteOverFires/TestClass_—_python/frameworks/pytest.yaml:60
    rule over-fires: TestClass emitted on tests/test_orders.py
    (a class not named Test* is not a pytest test class);
    got: TestClass/Orders, Test/test_orders
```

## What killed it, and what did not

It died on **`TestZeroProducingKinds6776_NoSiteOverFires`** — a *general* over-fire guard that walks rule sites and checks each kind against its own semantics. It did **not** die on this arm's fixture: `python-pytest-sqlalchemy-mini`'s forbidden rows pin `TestCommentedOut` and the indented `TestNestedHelper`, both of which are *anchor*-axis rows. A plain `class Orders:` in a pytest file is not forbidden anywhere in this fixture, so the fixture gate would have let ME-1 through.

That is a good outcome and a lucky one. The kind's specificity is graded — by a guard from #6776 that nobody working on this arm had in mind — and it happens to run in the **fast** engine suite, so an author iterating on `pytest.yaml` would hit it. It is worth knowing that is where the protection lives, because the fixture's own description says it "grades BOTH directions", and for the name axis it does not; the general guard does.

**No change requested.** The rule is graded, by a real named assertion with a precise message, at the level an author actually runs. I am recording where the grading lives so that a future arm of #6927 — five files and fourteen patterns remain — does not assume its fixture's forbidden rows cover the name axis just because they cover the anchor axis. They are separate axes and this fixture pins one of them.

It also usefully bounds the earlier `TestConfig` finding: the review measured `class TestConfig(Config):` typing as `TestClass` (1 hit in 3365 files). ME-1's kill confirms the guard catches *unnamed* over-fire but, by construction, cannot catch a class that genuinely **is** named `Test*` — so that residual is exactly as narrow as reported, and the reworded comment naming it is the right treatment.

Merging on green. Refs #6927, #6776.

Issue #6939, opened 2026-09-06
The first OCaml golden fixture exposed four ungraded extraction gaps: functors invisible, no class vocabulary, dead module CONTAINS, wrong module spans


cajasmota:
The first OCaml golden fixture landed in #6938 (`internal/quality/golden/ocaml-objects-mini`). Building it surfaced four extraction gaps that were previously ungraded and unrecorded. They are filed rather than fixed because #6812's owner decision forbids adding regex to `internal/extractors/ocaml`, and each of these is either fixed for free by the grammar wiring or is a deliberate scope call for it.

All four are present in the fixture source and **deliberately not asserted** there, with the reason written into `expected.json`'s description, so nobody reads their absence from the golden set as approval.

## 1. A functor declaration is not extracted at all

`internal/extractors/ocaml/extractor.go:48`:

```go
moduleRE = regexp.MustCompile(
    `(?m)^module(?:\s+type)?\s+([A-Z][a-zA-Z0-9_']*)\s*(?:=|:)`,
)
```

The name must be followed by `=` or `:` after optional whitespace. A functor's parameter list intervenes:

```ocaml
module Make (C : Comparable) = struct
  type item = C.t
  let sort_items xs = List.sort C.compare xs
end
```

`Make` produces **no entity**. Verified against the fixture: `containers.ml` yields `Comparable`, `Base` and `Extended` and not `Make`. Functors are not a corner of OCaml — they are how the language does what other languages do with generics — so this is a population question, not a curiosity, and it should be sized against a real corpus before anyone decides it is small.

## 2. No class construct produces any entity

There is no `classRE`. `class`, `class virtual`, `class type`, `method`, `val` and `inherit` are all invisible to entity extraction. In the fixture that means `shape`, `circle`, `square`, `tagged_circle`, `printer`, `verbose_printer` and `stream_writer` do not exist in the graph.

The consequence worth recording, because it is the thing #6812 turns on: the **parent name in an `inherit` clause does survive**, but only when the clause sits inside a `let` body, and only as a `CALLS` edge from the enclosing function — `make_logger --[CALLS]--> stream_writer` in the fixture. An `inherit` at a top-level `class` declaration is dropped entirely, because nothing scans there. So the regex path's `inherit` recall is not zero and is not one; it is "whatever happens to sit inside a scanned body". #6938 pins the surviving case so a grammar swap cannot silently drop it.

## 3. Module `CONTAINS` is empty for idiomatic code

[...]

PR #6944, opened 2026-09-06
fix(#6812): OCaml's vendored tree-sitter grammar was compiled in and never asked for — wiring it takes `inherit` attribution from 48.3% precision to 100%

### LINKED ISSUE #6812: DECIDED + MEASURED: wire OCaml's vendored tree-sitter grammar — the regex depth walker gets 230 of 408 inherit attributions WRONG (precision 48.3%); fixture and evidence shipped, the wiring is what remains

cajasmota:
Arm 2 of #6812 — the migration itself, now that arm 1 (`2de485618`, #6938) has measured what it replaces. Per the owner's decision on the issue: wire the grammar, add **no more regex** to `internal/extractors/ocaml`.

`extractModuleBody`'s hand-rolled depth walker is deleted from production and block spans come from `tsocaml.Language()` — the grammar registered at `internal/treesitter/adapters.go:69`, imported at `:22`, compiled in, and until now never asked for.

## Before / after, measured in one run

`github.com/ocaml/ocaml@70165ff23e` — 2420 files, the 68 containing a real `inherit` token, 633 reference-matched `object … end` blocks. Both producers scored over the same files, the same blocks and the same reference:

| | retired regex depth walker | CST-backed `extractModuleBody` |
|---|---|---|
| block end wrong | **479/633 (75.7%)** | **37/633 (5.8%)** |
| …of those, in files the grammar parses cleanly | 336 | **0** |
| `inherit` recall | 197/216 (91.2%), 19 missed | **216/216 (100%), 0 missed** |
| `inherit` precision | 197/408 (48.3%), 211 spurious | **216/216 (100%), 0 spurious** |
| attributions wrong | **230 of 408** | **0 of 216** |

> **Read the 100%/100% with this caveat, not without it** (it is the first thing anyone quoting this table needs, so it sits on the table rather than three paragraphs below): the reference matcher is **no longer an independent oracle for the grammar**, and the arithmetic bounds exactly how far that goes. Arm 1 established `refMatchEnd` ≡ the vendored grammar over **595** blocks (595/595). This measurement's population is **633**. So **38 blocks were never inside the established equivalence — and the residual disagreements number 37.** Essentially every disagreement sits outside the range where the two are known to coincide; over the ~595 where they do coincide, agreement is *definitional modulo the wiring*. What these figures genuinely grade is therefore the **wiring layer** — opener node types, same-parent `end` pairing, MISSING rejection, the binary search, the memo — which is where this arm could have been wrong, and none of which is in the grammar. They are a strong result about that layer; they are **not** 633 independent confirmations of the grammar. (Also in the corpus test's own doc comment, so it does not live only here.)

The legacy column reproduces arm 1 byte-for-byte on the same clone, which is the evidence that the harness was not quietly re-specified to make the new number look good.

**Every one of the 37 residual disagreements is in `testsuite/tests/generated-parse-errors/errors.ml`** — the corpus's one deliberately-malformed generated file. That is asserted rather than observed: the corpus test classifies each file by whether the *grammar* reports an ERROR node (never by filename) and requires **exactly zero** disagreements outside them, plus a vacuity guard that fails if no file in the corpus needed recovery. A threshold was rejected — it would let a real regression on valid source hide under a budget the malformed file already fills.

## Why the grammar was registered and never used

The issue asked for this before any of it was written, and *"nobody got round to it"* is a hypothesis. Checked: nothing was in the way. No adapter mismatch, no version pin blocking it, no build-tag path — `internal/treesitter/ts/grammars/ocaml`'s ABI smoke test has been parsing OCaml on every CI run the whole time, and `migratedLanguages` has routed `.ml` through it for the dispatcher's own parse. The extractor simply predates the grammar's arrival and nothing forced the two together. Its package comment still asserted *"No tree-sitter grammar for OCaml is bundled in smacker/go-tree-sitter"* — a premise that had been false since the grammar landed, and which is corrected in this diff.

## What the CST buys, defect by defect

Arm 1 named four independent causes. Three are free here because the grammar tokenises:

- **slice-relative `\b`** — `openKW`/`closeKW` were re-run on `rest[i:]` and tested for `om[0] == 0`, and at offset 0 of a slice start-of-string satisfies `\b`; `append`, `send`, `backend`, `legend`, `myobject` all read as closing keywords. Now `append` is a `method_name`.
- **non-nesting comment skipping** — `(* (* q *) end *)` is one `comment` node.
- **`{|…|}` quoted strings** — a `quoted_string` node with a `quoted_string_content` child.

The fourth, *the scan never terminating on a balanced block*, does not arise at all: a block's extent is a node span, not a counter's output.

## Scope — what was deliberately not done

- **No `EXTENDS`/`IMPLEMENTS`.** That is #6370's OCaml arm, which this unblocks. Hierarchy stays ungraded in the fixture, exactly as arm 1 left it.
- **#6939's four gaps are not fixed.** One falls out for free and is named rather than hunted: *every module in a file reporting the same end line* is now correct, because module `EndLine` derives from this span. The other three are untouched — functors are still invisible (`moduleRE` needs `=`/`:` immediately after the name), there is still no class vocabulary, and module `CONTAINS` is still dead (`letRE` is column-0-anchored, so an indented `let` inside a correct body still matches nothing).
- **Nothing else moved.** The wiring forced no rewrite: `extractLetBody`, `collectCalls`, `stripStringsAndComments`, the import/module/type/let discovery regexes and `depth.go` are untouched. `extractModuleBody` keeps its exact signature so the harness drives the new code through the same entry point with the same arguments.

## The parity choice, stated because it is a choice

`blockIndex` answers the same question the walker answered, including its unbounded one: *the first block opener at or after `from`*. It does not bound how far ahead that opener may be, so `module Foo = Bar` — an alias with no block of its own — still adopts the next block in the file. That mis-span is pre-existing, it is one of #6939's four, and bounding it here would have made the before/after figures incomparable. Named in `blocks.go` rather than left to be rediscovered.

## ERROR nodes — the decision, and it is graded

tree-sitter does not refuse malformed OCaml; it recovers, with ERROR subtrees and zero-width MISSING tokens. So the grammar *will* hand back an answer for an unclosed block and the question is what to do with it.

**Decision:** an opener whose `end` is absent or zero-width is reported as unclosed, and `extractModuleBody` hands back to the indentation heuristic — the same fallback the old code already used, and in fact the one it used for the *majority* of real blocks, since the non-terminating scan meant it almost never found a closer. It does **not** adopt a later block's `end` (a fabricated span) and it does **not** silently return nothing, which the brief explicitly rejected.

Graded by `TestBlockIndex_MalformedInput_FallsBack` (unclosed opener → declines, falls back, body stops at the next column-0 `let` and does not swallow the next block's `inherit`) and `TestBlockIndex_UnparseableInput_FallsBack` (an index with `parsed=false` declines; positive control that the *same* index answers once `parsed` is set, so the refusal is the guard firing and not the lookup missing).

Worth recording, because it shaped the design: **`hasError` is not a usable proxy for "this block is unclosed"**. Four candidate malformed inputs were tried before one produced an ERROR node — `let a = object end while ;;`, an unclosed `object`, an unclosed `struct`, and one of `errors.ml`'s own lines were all parsed *cleanly* into plausible trees with MISSING tokens. That is why the index declines on a zero-width `end` rather than trusting a file-level flag. `hasError` exists only so the corpus measurement can separate "this file is malformed" from "this span is wrong"; it is graded in both directions by `TestBlockIndex_HasErrorTracksRecovery`, since a `hasError` stuck true would make the corpus's zero-disagreement assertion vacuous.

## What happened to the harness, and to its reference matcher

This is the part the brief asked to be decided deliberately rather than allowed to become a tautology.

**The retired walker is preserved verbatim** as test-only `legacyDepthWalkerBody`. Without it every figure in #6812 becomes a claim in a PR body that nothing in the repo can reproduce. With it, the enumeration and the corpus measurement grade **two producers against one reference in a single run**, so "the CST implementation is better" is a measurement taken here, not a before/after taken across two checkouts.

**The reference matcher stays, and its role changed — say so, do not let an agreement read as a confirmation.** Arm 1's reviewer cross-checked `refMatchEnd` against this repo's own vendored grammar over these same files: 595 compared, 595 agree. So "the CST-backed producer agrees with `refMatchEnd`" is *not* the independent confirmation it was when the producer was a regex walker, and the corpus test now says that in full, in the test's own doc comment.

**The arithmetic that bounds it**, added at the reviewer's suggestion because it is more defensible than either "the reference is independent" or "the reference is no longer independent": arm 1's equivalence covered 595 blocks, arm 2's population is 633, so 38 blocks were never inside it — and there are 37 residual disagreements. That says exactly *where* the independence ran out.

It is not vacuous either, and the distinction is the code in between. `refMatchEnd` is hand-written and shares no line with `blocks.go`. What it still grades is everything the wiring added *on top of* the grammar: which node types count as openers, pairing an opener with the `end` of the **same parent node**, rejecting zero-width MISSING closers, the binary search for "first opener at or after `from`", and the per-source memo handing back the right file's index. Every one of those is a place this arm could have been wrong and none of them is in the grammar. What the reference can no longer do is *arbitrate* a disagreement — if the two differ, this test says so but cannot say which is right.

## The always-on enumeration: re-derived, not deleted

`TestDepthWalker_ObjectBlockEnd_EnumeratedSpace` now runs both producers over the same 32 inputs:

- **legacy: unchanged at the recorded 18/32.** Arm 1's finding is still reproducible in-repo.
- **CST: 0/32**, recorded as its own named empty literal `cstDivergingCombinations6812`, with the per-axis account of why each of the four axes (indented `end`, `append`, nested comment, quoted string) is one of the four measured defects and why each disappears.

An empty set looks weak, so: `endsAgree` fails on *any* disagreement in either direction, so a producer that truncates, over-reaches, returns `""` or silently falls back to the indentation heuristic all land in that set and fail. The direction that would be invisible — agreeing for the wrong reason — is what running the legacy walker alongside answers, and **a guard fails if the two sets ever coincide** (the wiring reverted, or the table pointed at one producer twice). The axis-E inertness and E-symmetry checks are asserted against the **legacy** set, where they are non-vacuous; over an empty set they would be two guards passing by having nothing to look at.

The 0/32 inherit-mis-attribution guard is kept and now applies to **both** producers — a CST-backed span starting *before* the `inherit` would break the generator's premise just as surely as one stopping short.

`TestDepthWalker_InheritIsMisAttributed_Constructed` keeps its input and now asserts a **pair**: the retired walker still mis-attributes on those bytes (1/4), the CST-backed producer does not (0/4). Asserting only the second would be an absence over a population a broken generator could empty.

## Verification

- `go test -count=1 ./internal/extractors/ocaml/ ./internal/quality/... ./cmd/grafel/` — all green. **`./cmd/grafel/` included and passing**: `custom_extractor_spans_6118_test.go` digest-pins the whole graph, nothing in the diff points at it, and the digest did **not** move, so no constant was bumped and there is no delta to enumerate.
- `scripts/quality/run.sh --ratchet` — `quality ratchet OK — 34 gated fixtures held their baseline (25 at 100% must-have recall)`.
- **Shrink enumerated by hand, since #6898 lets one pass silently.** Compared per-fixture `entity_found` / `relationship_found` / `entity_extracted_total` / `relationship_extracted_total` against `baseline.json` for all 34: five deltas, all three known python drifts (`python-django-mini` rel-total 255→251; `python-fastapi-mini` 28→27 / 57→54; `python-rq-mini` 22→20 / 43→39). **Proved pre-existing** by building `origin/main@2e68fb052` into a separate binary and re-running the ratchet against it with `GRAFEL_BIN`: identical five deltas, and **branch vs pre-change binary is 0 deltas across all 34 fixtures on all six counters**. Restore, do not record.
- **Corpus-size guards re-derived from disk, not from the diff**: `ls internal/quality/golden/` is 34 directories; the three guards (`absent_relationships_6490_test.go:232`, `subtype_assertion_6488_test.go:291`, `zero_relationships_6490_test.go:181`) all say 34. This diff adds no fixture, so none moves.
- `ocaml-objects-mini`: 19/19 entities, 6/6 relationships, 0 forbidden hits, **zero expectation changes** — no row was rewritten, so nothing here asserts the one-week-old fixture was wrong. Its `entity_extracted_total` (28) and `relationship_extracted_total` (68) are byte-identical to origin/main.

### Review follow-ups (Q1, Q2), applied

- **Q1 — `blockOpeners` was pinned only in the deletion direction.** M2/M3 delete `begin`/`sig` and die; adding an entry was ALIVE, and it is the more dangerous mistake: a phantom opener sitting *before* the real one wins the "first opener at or after `from`" search, has no `end`, makes `bodyEnd` decline, and the block silently falls back — visible only as recall loss, the direction recall-shaped assertions are blind to (#6902). Same shape as the axis lesson from #6943. Fixed with two paired assertions rather than one: `TestBlockIndex_NonBlockKeywordsAreNotOpeners` is **enumerated, not hand-picked at `then`** — the fixture carries `if/then/else`, `match/with`, `try/with`, `fun`, `for/do/done`, `while/do/done` and `let/in` around an `object … end` holding an `inherit`, and asserts the opener count is exactly 1, that `bodyEnd` still answers (the half that catches a phantom preceding the real opener), and that the span still contains its own `inherit`. `TestBlockOpeners_ExactSet` is the residue for a keyword the fixture does not happen to contain.
- **Q2 — the vacuity guard on the corpus assertion was one-sided.** `blockEndWrongOK != 0` was guarded against *nothing* being classified as recovered, but not against *everything* being so classified — in which case the zero is unreachable and the assertion passes by never being able to fail. The five-ways-to-be-a-no-op shape. Now also requires `legacy.blockEndWrongOK > 0`, pinned through the **legacy producer** rather than a file count: the retired walker is known to get block ends wrong on valid OCaml, so if not one of its 479 errors lands in a cleanly-parsed file, the classifier has swallowed the corpus. **Proved reachable, not asserted:** a mutant forcing `recovered := true` for every file fails the corpus test on this new guard specifically.

### Mutants — 9 scored, 9 DEAD

Each applied to a hash-verified pristine tree, occurrence count asserted **exactly 1** before and 0 after, an **independent** site-verification script (grepping all eight mutation sites from disk) confirming the mutant landed, and the unmutated suite proven **green immediately before every verdict**.

| | mutant | verdict | killed by |
|---|---|---|---|
| M1 | accept a zero-width MISSING `end` as a real closer | DEAD | `TestBlockIndex_MalformedInput_FallsBack` |
| M2 | drop `begin` from `blockOpeners` | DEAD | `TestBlockIndex_AllFourOpeners` |
| M3 | drop `sig` from `blockOpeners` | DEAD | `TestBlockIndex_AllFourOpeners` |
| M4 | do not recurse into children | DEAD | 5 tests, incl. `…NestedBlocksPairWithTheirOwnEnd` |
| M5 | answer from an index whose parse failed | DEAD | `TestBlockIndex_UnparseableInput_FallsBack` |
| M6 | always take the file's FIRST opener | DEAD | `…NestedBlocksPairWithTheirOwnEnd`, `…InheritIsMisAttributed_Constructed` |
| M7 | accept an `end` that precedes `from` | DEAD | `TestBlockIndex_EndBeforeFromIsRefused` |
| M8 | memo ignores its key | DEAD | `TestBlockIndex_AllFourOpeners` |
| M9 | no fallback when the CST declines | DEAD | `TestBlockIndex_MalformedInput_FallsBack` |

Three of these are worth calling out:

- **M2 and M3 are why `TestBlockIndex_AllFourOpeners` exists.** The enumeration generates `object` and nothing else, so deleting `begin`, `sig` or `struct` left the whole package green — three quarters of the table ungraded, and `sig` is the entry production reaches most often via `module type T = sig … end`. The new test pins the exact body text per opener, not a length, so a span off by the width of the opener keyword (the class of defect that lived undetected in the harness's own reference as `i += 6`) fails rather than passing a byte count.
- **M7 was ALIVE and was killed rather than argued equivalent.** `e < from` and `e < 0` are equivalent through `buildBlockIndex` — the binary search returns an opener at or after `from`, and a real `end` is strictly after its opener — but "unreachable through today's constructor" is an argument, not a fence: `bodyEnd` is a method and the guard is what stops `src[afterPos:end]` being a negative-length slice. Cost of the kill, since *"too expensive"* deserves a number: **nine lines and no production change**.
- **The first two mutant runs are void and are reported as such.** The scoring harness silently left a mutant applied, so a subsequent "DEAD" was scored against an already-red tree and meant nothing. Caught by adding a per-mutant positive control (unmutated suite must be green *immediately before* each verdict) and an independent on-disk site check that must **notice** every mutant and must confirm every restore. The table above is from the rebuilt harness; the tree was re-verified pristine on all eight sites and green before committing.

Closes #6812
Refs #6370
Refs #6939

### DIFF
--- a/internal/extractors/ocaml/blocks.go
+++ b/internal/extractors/ocaml/blocks.go
@@ -1,0 +1,265 @@
+// blocks.go — CST-backed block spans for the OCaml extractor (#6812).
+//
+// # Why this file exists
+//
+// `struct … end`, `sig … end`, `object … end` and `begin … end` are the only
+// structural judgement this extractor owns, and until this file landed it was
+// made by a hand-rolled depth walker inside extractModuleBody: two regexes
+// re-run at every byte, counting `end` against `struct|sig|object|begin`.
+//
+// #6812 measured that walker against a clone of github.com/ocaml/ocaml at
+// 70165ff23e, over the 68 files that contain a real `inherit` token:
+//
+//   - 479 of 633 `object … end` blocks (75.7%) got the WRONG end;
+//   - of the 408 (block, inherit) pairs it attributed, 230 were wrong —
+//     recall 91.2% (19 missed), precision 48.3% (211 spurious).
+//
+// Four independent causes, three of which a concrete syntax tree removes by
+// construction:
+//
+//  1. the scan never terminated on a balanced block — the opener took depth
+//     0→1 and the matching `end` took it 1→0 without stopping, because the
+//     loop only broke on an `end` seen AT depth 0, so well-formed blocks
+//     always fell through to the indentation fallback;
+//  2. `\b` was evaluated against a SLICE, not the source: `openKW`/`closeKW`
+//     were re-run on `rest[i:]` and tested for `om[0] == 0`, and at offset 0
+//     of a slice start-of-string satisfies `\b`, so `append`, `send`,
+//     `backend`, `legend`, `myobject` all read as keywords;
+//  3. OCaml block comments NEST and the skipper stopped at the first `*)`;
+//  4. `{|…|}` quoted strings were unknown to it and scanned as code.
+//
+// (2), (3) and (4) are free here: the grammar tokenises. (1) does not arise:
+// a block's extent is a node span, not the output of a counter.
+//
+// The grammar itself was already vendored, registered at
+// internal/treesitter/adapters.go:69 and compiled in — it had simply never
+// been asked for. There was no adapter mismatch, no version pin and no build
+// tag standing in the way; the smoke test in
+// internal/treesitter/ts/grammars/ocaml has been parsing OCaml on every CI run
+// the whole time. The answer to "why was it registered and never used" is that
+// this extractor predates the grammar's arrival and nothing forced the two
+// together.
+//
+// # Contract
+//
+// blockIndex answers exactly one question, and deliberately the SAME question
+// the depth walker answered, so the #6812 measurement harness compares like
+// with like: given a position `from` from which the block's OPENING keyword is
+// still ahead (extractor.go hands it the end of the `module Foo =` match, with
+// `struct` still to come), where does that block's `end` token start?
+//
+// It does NOT bound how far ahead the opener may be. That is a deliberate
+// parity choice, not an oversight: `refMatchEnd`, the independent reference in
+// the measurement harness, has the same contract, and a `module Foo = Bar`
+// alias with no block of its own therefore still adopts the next block in the
+// file. That mis-span is pre-existing (the depth walker did the same thing),
+// it is one of #6939's four gaps, and closing it here would have made the
+// before/after figures incomparable.
+//
+// # When the CST cannot answer
+//
+// Three cases, all of which hand back to the indentation heuristic
+// (extractLetBody) — which is where the depth walker already sent the majority
+// of real blocks anyway, since defect (1) meant it almost never terminated:
+//
+//   - the grammar produced no tree at all;
+//   - no block opener exists at or after `from`;
+//   - the nearest opener has no `end` — either genuinely absent or supplied by
+//     tree-sitter as a zero-width MISSING node inside an ERROR subtree.
+//
+// The third is the one the corpus forced: `testsuite/tests/generated-parse-
+// errors/errors.ml` in the OCaml tree is generated, deliberately malformed
+// source with 256 `object` against 38 `end`. Answering it with the nearest
+// LATER block's `end` would be a fabricated span, so an unclosed opener is
+// reported as unclosed and the caller falls back. This is graded by
+// TestBlockIndex_MalformedInput_FallsBack.
+package ocaml
+
+import (
+	"sort"
+	"sync"
+
+	"github.com/cajasmota/grafel/internal/indexstate"
+	"github.com/cajasmota/grafel/internal/treesitter/ts"
+	tsgrammar "github.com/cajasmota/grafel/internal/treesitter/ts/grammars/ocaml"
+	"github.com/cajasmota/grafel/internal/treesitter/ts/official"
+)
+
+// blockOpeners are the four anonymous tokens that open an `… end` block in the
+// tree-sitter-ocaml grammar. They appear as unnamed children of, respectively,
+// `structure`, `signature`, `object_expression`/`class_body_type` and
+// `parenthesized_expression`. Keying on the TOKEN rather than on the parent
+// node type is what keeps `( … )` — which shares parenthesized_expression with
+// `begin … end` — out of the index.
+var blockOpeners = map[string]bool{
+	"struct": true, "sig": true, "object": true, "begin": true,
+}
+
+// blockIndex records, for every block opener in a source file, the byte offset
+// at which its closing `end` token starts (or -1 when the block has none).
+type blockIndex struct {
+	// openers is ascending, so a lookup is a binary search.
+	openers []int
+	end     map[int]int
+	// parsed is false when the grammar produced no usable tree; every lookup
+	// then declines and the caller falls back.
+	parsed bool
+	// hasError records whether the tree contains an ERROR node anywhere, i.e.
+	// whether tree-sitter had to recover. It changes no answer this file gives
+	// — a recovered tree is still consulted, block by block — and exists so a
+	// caller can separate "this file is malformed" from "this block span is
+	// wrong". The #6812 corpus measurement uses it to assert that every
+	// residual disagreement with the reference is in a file the grammar itself
+	// reports as broken, which is a much stronger claim than a bounded count.
+	hasError bool
+}
+
+var (
+	ocamlAdapter = official.New()
+	ocamlGrammar = tsgrammar.Language()
+)
+
+// buildBlockIndex parses src with the vendored OCaml grammar and records every
+// block span in it.
+func buildBlockIndex(src string) *blockIndex {
+	bi := &blockIndex{end: make(map[int]int)}
+	parser, err := ocamlAdapter.NewParser(ocamlGrammar)
+	if err != nil {
+		return bi
+	}
+	defer parser.Close()
+
+	// #5630 — this is a second parse of a file the dispatcher has usually
+	// already parsed, so it must take the same daemon-wide parse slot that
+	// treesitter.ParserFactory.Parse takes. GOMAXPROCS does not bound cgo, so
+	// a parse outside the gate makes the in-process ceiling illusory. The slot
+	// is held around the parse ALONE (not the walk below), and released via
+	// defer inside the closure so a panic in Parse cannot leak it permanently.
+	tree, perr := func() (ts.Tree, error) {
+		indexstate.AcquireParseSlot()
+		defer indexstate.ReleaseParseSlot()
+		return parser.Parse([]byte(src))
+	}()
+	if perr != nil || tree == nil {
+		return bi
+	}
+	defer tree.Close()
+	root := tree.RootNode()
+	if root == nil {
+		return bi
+	}
+	bi.parsed = true
+	bi.collect(root)
+	sort.Ints(bi.openers)
+	return bi
+}
+
+// collect walks every node, named and anonymous, recording each block-opening
+// token against the `end` token of the same parent node.
+func (bi *blockIndex) collect(n ts.Node) {
+	if n.IsError() {
+		bi.hasError = true
+	}
+	opener, end := -1, -1
+	for i := 0; i < int(n.ChildCount()); i++ {
+		c := n.Child(i)
+		if c == nil {
+			continue
+		}
+		if !c.IsNamed() {
+			switch {
+			case blockOpeners[c.Type()] && opener < 0:
+				opener = int(c.StartByte())
+			case c.Type() == "end":
+				// A MISSING `end` inserted by tree-sitter's error recovery is
+				// zero-width. Treating it as a real closer would invent a span
+				// ending where the parser gave up, which is precisely the kind
+				// of fabricated answer this file exists to stop producing.
+				if c.EndByte() > c.StartByte() {
+					end = int(c.StartByte())
+				}
+			}
+		}
+		bi.collect(c)
+	}
+	if opener < 0 {
+		return
+	}
+	if _, seen := bi.end[opener]; !seen {
+		bi.openers = append(bi.openers, opener)
+	}
+	bi.end[opener] = end
+}
+
+// bodyEnd returns the byte offset at which the closing `end` of the first
+// block opening at or after `from` starts. ok is false when the index cannot
+// answer — see the "When the CST cannot answer" section of the package note.
+func (bi *blockIndex) bodyEnd(from int) (int, bool) {
+	if !bi.parsed {
+		return 0, false
+	}
+	i := sort.SearchInts(bi.openers, from)
+	if i >= len(bi.openers) {
+		return 0, false
+	}
+	e := bi.end[bi.openers[i]]
+	if e < from {
+		// Either the block is unclosed (-1) or the `end` precedes `from`,
+		// which would produce a negative-length body.
+		return 0, false
+	}
+	return e, true
+}
+
+// ---------------------------------------------------------------------------
+// Per-source memo
+// ---------------------------------------------------------------------------
+
+// extractModuleBody is called once per module entity and again per module from
+// addModuleContains, and the #6812 harness calls it once per block. Parsing on
+// every call would re-parse the same file up to a few dozen times, so the last
+// few indexes are memoised by their exact source text.
+//
+// Keyed on string EQUALITY, not on a hash: a hash collision would silently
+// hand one file's block spans to another, and this package has no way to
+// notice. Equality is cheap in the case that matters — every caller inside one
+// extraction passes the same string value, so the comparison short-circuits on
+// the data pointer.
+//
+// The ring is small on purpose. It is a re-parse avoider for one file being
+// walked, not a cache of the corpus; extraction workers run concurrently and a
+// large ring would pin whole files' worth of trees' worth of maps for as long
+// as the process lives.
+var blockCache struct {
+	mu   sync.Mutex
+	ring [8]blockCacheEntry
+	next int
+}
+
+type blockCacheEntry struct {
+	src string
+	idx *blockIndex
+}
+
+func blockIndexFor(src string) *blockIndex {
+	blockCache.mu.Lock()
+	for _, e := range blockCache.ring {
+		if e.idx != nil && e.src == src {
+			idx := e.idx
+			blockCache.mu.Unlock()
+			return idx
+		}
+	}
+	blockCache.mu.Unlock()
+
+	// Parse OUTSIDE the lock: it takes the parse gate, and holding a package
+	// mutex across a gated cgo call would serialise every OCaml extraction in
+	// the process behind one file.
+	idx := buildBlockIndex(src)
+
+	blockCache.mu.Lock()
+	blockCache.ring[blockCache.next] = blockCacheEntry{src: src, idx: idx}
+	blockCache.next = (blockCache.next + 1) % len(blockCache.ring)
+	blockCache.mu.Unlock()
+	return idx
+}
--- a/internal/extractors/ocaml/blocks_6812_test.go
+++ b/internal/extractors/ocaml/blocks_6812_test.go
@@ -235,3 +235,87 @@ func TestBlockIndex_EndBeforeFromIsRefused(t *testing.T) {
 		t.Fatalf("got (%d,%v), want (9,true)", e, ok)
 	}
 }
+
+// TestBlockIndex_NonBlockKeywordsAreNotOpeners pins blockOpeners in the
+// ADDITION direction. Review Q1.
+//
+// M2 and M3 delete `begin` and `sig` from the set and both die. Nothing
+// covered the other direction, and an over-broad opener is the more dangerous
+// mistake of the two: a phantom opener sitting BEFORE the real one wins the
+// "first opener at or after `from`" search, has no `end` of its own, and makes
+// bodyEnd decline — so the block silently falls back to the indentation
+// heuristic and its `inherit` is dropped. That shows up only as recall loss,
+// which is the direction recall-shaped assertions are structurally blind to
+// (#6902), and it is the same shape as #6943: one direction of a set graded,
+// its neighbour left open.
+//
+// The fixture is ENUMERATED rather than hand-picked at `then`, because a fence
+// aimed at one keyword is a fence one keyword wide. The source carries every
+// OCaml keyword that could plausibly be mistaken for a block opener —
+// including the four that legitimately introduce nesting in the surface syntax
+// (`then`, `else`, `do`, `with`) — so adding ANY of them to the set moves the
+// opener count off 1 and fails here.
+func TestBlockIndex_NonBlockKeywordsAreNotOpeners(t *testing.T) {
+	src := "let c = if cond then object\n" +
+		"    inherit base\n" +
+		"    method a = match x with Some y -> y | None -> 0\n" +
+		"    method b = try (fun z -> z) 1 with _ -> 0\n" +
+		"    method d = for i = 0 to 2 do ignore i done\n" +
+		"    method e = while false do () done\n" +
+		"    method f = let g = 1 in g\n" +
+		"  end else other\n"
+
+	bi := blockIndexFor(src)
+	if !bi.parsed {
+		t.Fatal("the grammar produced no tree")
+	}
+	if len(bi.openers) != 1 {
+		var got []string
+		for _, o := range bi.openers {
+			got = append(got, src[o:min(o+8, len(src))])
+		}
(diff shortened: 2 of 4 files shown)


cajasmota:

Measured in my own detached worktree at `60678f4f2` (PR head, one commit on top of `origin/main@2e68fb052`), `git status --porcelain` empty, `blocks.go` sha `00fd7a1d…` before and after every mutant. Not the author's worktree.

### Re-scored mutants from a verified-clean tree

The two void runs are the reason to re-score rather than to read the table. I re-scored three of the nine, including M7, with a green baseline proved **immediately before each verdict**, an exact occurrence count (`1` before, `0` after) and a sha check that the file actually changed on disk.

| | mutant | my verdict | killed by |
|---|---|---|---|
| M7 | `e < from` → `e < 0` | **DEAD** | `TestBlockIndex_EndBeforeFromIsRefused` |
| M1 | accept a zero-width MISSING `end` | **DEAD** | `TestBlockIndex_MalformedInput_FallsBack` |
| M6 | always take the file's FIRST opener | **DEAD** | `…NestedBlocksPairWithTheirOwnEnd`, `…InheritIsMisAttributed_Constructed` |

Same verdicts, same killers as the table. **The re-scored verdicts are clean of the contamination** — each was taken against a baseline observed green on the pristine file seconds earlier, so none of them can be inheriting a red tree. M7 in particular is genuinely dead, not dead-by-leftover.

### Four mutants of my own

| | mutant | verdict |
|---|---|---|
| Q1 | `blockOpeners` **gains** `"then"` (over-broad opener set) | **ALIVE — distinguishable and ungraded** |
| Q2 | memo key `e.src == src` → `len(e.src) == len(src)` | DEAD (`…ParenthesesAreNotBlocks`) — but see below |
| Q3 | an unclosed opener adopts the **next** block's `end` (the fabricated span) | DEAD (`…MalformedInput_FallsBack`) |
| Q4 | `bodyEnd` always declines (total CST bypass, agreeing "for the wrong reason") | DEAD — 6 tests incl. **`…EnumeratedSpace`** |

**Q1 is the one finding.** `blockOpeners` is graded only in the *deletion* direction (M2/M3 → `TestBlockIndex_AllFourOpeners`). Adding a key survives the whole package. Demonstrated distinguishable, on `let g = if a then b else c\nlet c = object\n  inherit base\n …`:

```
unmutated: openers=[35]     bodyEnd(0)=(74,true)   body="…let c = object\n  inherit base\n  method a = 1\n  "
mutated:   openers=[13 35]  bodyEnd(0)=(0,false)   body="let g = if a then b else c"
```

i.e. an over-broad opener silently **loses the block's `inherit`** — a recall loss, the direction #6902 says assertions are structurally blind to. The corpus run would catch it; the always-on suite does not. Fix is one case in `TestBlockIndex_ParenthesesAreNotBlocks` (`if … then … else`, `for … do … done`, `match … with` → `len(bi.openers) == 0`). Not a blocker: no such token is in the shipped set, and the wiring's other five directions are fenced.

Q4 is the answer to "an empty expected set is the classic vacuous assertion": a producer that silently falls back **does** land in `cstDivergingCombinations6812` and fails, and the coincide-guard is reachable. The 0/32 is observed, not asserted.

Q2 is DEAD but *incidentally* — the kill depends on two unrelated test sources happening to share a byte length. The memo's key-equality contract (the thing the comment argues for over a hash) is not pinned on purpose.

### The 100%/100% — how much is independently established

Split it, because the arithmetic makes the split sharp:

- Arm 1's cross-check was **595 blocks compared, 595 agree** (218 unmatched, all in `errors.ml`). Arm 2's population is **633** reference-matched blocks. So **38 of the 633 were never inside the established `refMatchEnd` ≡ grammar equivalence — and the residual disagreements number 37.** The residual sits almost exactly in the region where equivalence was never claimed, which is coherent, and it means: **over the ~595 where equivalence holds, agreement is definitional modulo the wiring.** The recall/precision figures are derived from the same spans, so 100%/100% is definitional in exactly the same measure.
- What is **not** definitional, and is the real content of the number: the wiring layer between the grammar and the answer — opener token set, same-parent pairing, zero-width MISSING rejection, the `SearchInts` lower bound, the memo. Every one of those is independently graded; my M1/M6/M7/Q3/Q4 kills confirm five of six directions are fenced, with Q1 the sixth.
- The genuinely *independent* evidence for OCaml truth is `refMatchEnd`'s 22 hand-computed positive controls and the enumeration — not the corpus agreement.

**The honesty is materially complete**, and further than most PRs go: the caveat is in the PR body *and* in the corpus test's own doc comment, and it says the reference can no longer arbitrate. Two sharpenings I would take:

1. The headline table quotes 100%/100% with no pointer to the caveat five sections down. One parenthetical would fix it.
2. Neither the PR nor the test states the **595-of-633** arithmetic — that the 37 residuals are precisely the blocks arm 1's equivalence never covered. That is the strongest honest thing available to say about them, stronger than "they are in `errors.ml`".

**One real gap in the corpus assertion (test-only, env-gated).** `cst.blockEndWrongOK != 0` is guarded against one vacuity (`filesWithErrors == 0`) but not against its mirror: if every measured file were classified as recovered, the zero would be unreachable and nothing would say so. `TestBlockIndex_HasErrorTracksRecovery` covers the unit direction, but the corpus assertion's own denominator is unguarded. Free fix, non-vacuous by the author's own run: also require `legacy.blockEndWrongOK > 0` (it is 336).

### The other claims — checked

- **`hasError` is not a usable proxy** — **confirmed independently.** All four parse with `hasError=false`: `let a = object end while ;;` (openers=[8] end=15), unclosed `object` (end=-1), unclosed `struct` (end=-1), `if UIdent then with` (no openers). Only `let ) = ( ;; %%%` sets it. The decline-on-zero-width-`end` design is load-bearing, not belt-and-braces. This deserves to live somewhere more durable than a PR body — a line in `blocks.go`'s package note would do it.
- **Zero expectation changes** — structural, not a claim: the diff is 4 files, all in `internal/extractors/ocaml`, no `expected.json`, no `baseline.json`.
- **Shrink, re-derived independently.** `scripts/quality/run.sh --ratchet` in an isolated `GRAFEL_HOME`/`GRAFEL_DAEMON_ROOT`: exit 0, *"34 gated fixtures held their baseline (25 at 100% must-have recall)"*. Enumerated all six counters × 34 fixtures against `baseline.json` myself: **exactly 5 deltas, exactly the three python drifts** (`django` 255→251, `fastapi` 28→27 / 57→54, `rq` 22→20 / 43→39) — untouchable by a diff confined to the OCaml extractor. `ocaml-objects-mini` measured 19/19, 6/6, totals **28/68 — byte-identical**.
- **Corpus-size guards** — `ls -d internal/quality/golden/*/` is **34** on disk; the three guards all say 34. Re-derived, not read off the diff.
- **Legacy walker** — `grep` across `internal/` finds `legacyDepthWalkerBody` in exactly one file, the test. `go test -count=1 ./internal/quality/... ./cmd/grafel/` green (158s on `cmd/grafel` — the digest pin did not move).
- **Scope** — no `EXTENDS`/`IMPLEMENTS` anywhere in the diff; `moduleRE`/`letRE`/`extractLetBody`/`collectCalls`/`depth.go` untouched, so #6939's other three gaps are demonstrably not fixed. The one that fell out for free is named in the body.
- **Corpus test skips cleanly** without `GRAFEL_OCAML_CORPUS`; `go vet` clean.

### Verdict

**APPROVE.** The re-scored verdicts are clean; the 100%/100% is honestly framed and its definitional half is bounded by an arithmetic the PR should state; the wiring — the part the number actually grades — survives everything I threw at it except one over-broad-opener direction. Three follow-ups, none blocking: the Q1 forbidden-opener case, the `legacy.blockEndWrongOK > 0` mirror guard, and the 595-of-633 sentence.

cajasmota:
Coordinator mutant on the final head `472fc1d38`. One mutant, **DEAD** — killed by a test written for an entirely different purpose, which is the interesting part.

## MF-1 — widen the memo key

`blockIndexFor`'s ring compares source text by **exact string equality**, and the comment says why:

> Keyed on string EQUALITY, not on a hash: a hash collision would silently hand one file's block spans to another, **and this package has no way to notice.**

That last clause is a claim about a hazard, and nothing obviously observes it — the "prose asserts what no test observes" shape. The review's Q2 (key by length) died only *incidentally*, via a length collision between unrelated test sources, so the memo's correctness was graded by luck rather than by design. I wanted a mutant that models the real failure.

**A pure widening** — keep exact equality, additionally accept a 32-byte prefix match:

```go
if e.idx != nil && (e.src == src ||
    (len(e.src) >= 32 && len(src) >= 32 && e.src[:32] == src[:32])) {
```

This is not a contrived collision. A shared licence header, a `(* ... *)` module docstring, or an `open Core` prelude gives two different OCaml files an identical 32-byte prefix routinely. Anchor count 1 pre-edit, 0 post-edit, new form 1, `go vet` clean, baseline green first (`ok 0.709s`), `-count=1`, restored by `cp` from a shasum-verified backup (`b454682ed…`), `git status --porcelain` empty.

**Verdict: DEAD** at package level (the quality gate stays green, so this is graded only in the fast suite — which is the right place, since it is the suite an author actually runs).

```
--- FAIL: TestDepthWalker_ObjectBlockEnd_EnumeratedSpace
    ----- [legacy regex depth walker] BLOCK-END DIVERGES: span end=51, true end=47
    A---- [CST-backed extractModuleBody] BLOCK-END DIVERGES: span end=47, true end=45
    -B--- [CST-backed extractModuleBody] BLOCK-END DIVERGES: span end=47, true end=69
```

## Why it died, and why that is worth writing down

The enumeration exists to compare two block-end producers across 32 combinations of five axes. To do that it generates 32 sources that **differ from each other only in small, localised ways** — which makes it, incidentally, a near-ideal collision detector for a prefix-keyed cache. Every variant shares a long prefix with its siblings, so a widened key hands one variant's spans to another and the CST column immediately reports spans that belong to a different input (`span end=47` recurring against three different true ends).

So the hazard the comment describes **is** observed, and the claim is not unpinned after all. But it is observed by a test that has no idea it is doing so, and the protection is a side effect of how the generator varies its inputs. **One line in `blockIndexFor`'s comment naming `TestDepthWalker_ObjectBlockEnd_EnumeratedSpace` as what would catch a loosened key** would make that durable — otherwise the next person to "optimise" this to a hash sees only a comment asserting a danger with no visible guard, which is exactly the situation that invites the change.

This is the same lesson as #6943's name-axis mutant dying on a *general* over-fire guard rather than the arm's own fixture: the protection is real, it just does not live where you would look for it. Worth a pointer in both directions.

**No change requested beyond that optional comment line.** Everything from the review round landed: `blockOpeners` is now fenced in the addition direction and — better than I asked — **enumerated over seven OCaml constructs rather than hand-picked at `then`**, since a fence aimed at one keyword is one keyword wide; the corpus mirror guard is closed and **demonstrated** by a mutant that forces every file to classify as recovered; and the 595-of-633 arithmetic now sits both on the headline table and in the test's own doc comment, so the figure travels with its caveat.

Merging on green. Refs #6812, #6370.

Issue #6152, opened 2026-08-05
Every bare Python class is typed Controller: falcon/cherrypy YAML class patterns are ungated by framework detection


cajasmota:
## The behaviour

Every bare Python class is typed `Controller` in the graph, on both the full and (since #6148) the incremental path.

Measured directly against the classifier, with no framework anywhere in the file:

```go
engine.New(engine.LoadAllRules()).Detect(ctx, extractor.FileInput{
    Path: "cphandler_delta.py", Language: "python",
    Content: []byte("class CpPlainProbe:\n    def cp_probe_noop(self):\n        return 1\n"),
})
// → entity kind="Controller" name="CpPlainProbe" props=[framework:python pattern_type:yaml_driven]
```

Two rule sets match a bare class declaration, neither gated on the framework actually being present:

- `internal/engine/rules/python/frameworks/falcon.yaml` — `pattern: "class\\s+(\\w+)\\s*(?:\\([^)]*\\))?\\s*:"` → `entity_type: Controller`
- `internal/engine/rules/python/frameworks/cherrypy.yaml` — `pattern: "class\\s+(\\w+)\\s*(?:\\(object\\))?\\s*:"` → `entity_type: Controller`

Detection metadata exists in both files (`import_markers: [import falcon, …]`) and is consumed by the modular loader, but `Detector.Detect` resolves compiled rule sets by `file.Language` alone — so the extraction patterns fire on every Python file in every repo, framework or not.

## Evidence it is already treated as noise

`internal/extractors/migration_prune.go:63-77` exists to delete some of the output:

```go
// "Controller" is included to prune the scaffolding noise introduced by
// … extractors that run on ALL Python files including migration files and
// emit a Controller entity for every `class Migration(...)` declaration.
// #3173: Falcon/CherryPy YAML `class X:` -> Controller patterns fire on …
"Controller": true,
```

That prune is scoped to migration files. Everything else keeps the label.

## Why this is filed rather than fixed

It surfaced while diagnosing #6148 ("incremental classifies a delta-file class as SCOPE.Component where full rebuild says Controller"). That issue is about the two paths disagreeing, and the correct scope discipline there was to make the incremental path match the full one — not to change what the full one decides. Changing this reclassifies every Python class in every indexed repo and needs its own corpus measurement and re-index.

## What a fix would have to establish first

1. How many `Controller` entities across the corpus come from these two ungated patterns versus from a real framework — measured, not estimated.
[...]

PR #6945, opened 2026-09-06
fix(#6927): GitHub Actions `^` anchors were start-of-TEXT — three of six rules could never fire, and three more were plain YAML the anchor was gating by accident

cajasmota:
The **GitHub Actions arm** of #6927 — 6 of the 14 remaining patterns, the largest block.
`docker_compose.yaml`, `ansible_core.yaml`, `play_framework.yaml` and `graphql_schema.yaml` are
untouched and stay separate arms.

`internal/engine/detector.go` compiles rule patterns with plain `regexp.Compile` at `:156,177`, so
`^` means start of **text**. In `rules/cicd/frameworks/github_actions.yaml` that left **three
patterns unable to fire at all** — `^\s{2}…` (job), `^\s+-\s+name:` (step) and `^env:\s*\n…` all
require something before the match, and start-of-text forbids it — and the other three firing only
on a file whose very first byte is `name:` or `on:`.

## The first question was not "add `(?m)`"

It was, per pattern: **is this actually specific to GitHub Actions, or has start-of-text been doing
the gating?** #6943 found exactly that in sqlalchemy. Answered here by measurement, not by reading —
every `.yml`/`.yaml` in a 59-repo corpus (`starter-workflows`, `argocd-example-apps`,
`prometheus-helm`, `awesome-compose`, `kafka`, `django`, `rails`, …): **671 files, 269 of them
GitHub Actions workflows/composite actions by location, 402 not.** Matches, files in brackets:

| pattern | kind | current (start-of-text) wf / other | `(?m)` ungated wf / other | `(?m)` + gate wf / other | verdict |
|---|---|---|---|---|---|
| `^name:\s+["']?([^"'\n]+)` | Service | 99 (99) / 24 (24) | 264 (264) / **34 (34)** | 264 (264) / 0 (0) | **gated** |
| `^\s{2}(\w[\w-]*):\s*\n\s+(?:runs-on\|uses):` | Operation | 0 / 0 | 152 (131) / **0 (0)** | — | ungated |
| `^\s+-\s+name:\s+["']?([^"'\n]+)` | Task | 0 / 0 | 1138 (232) / **194 (49)** | 1138 (232) / 0 (0) | **gated** |
| `^on:\s*\n\s+(\w+):` | Config | 1 (1) / 0 | 234 (234) / **0 (0)** | — | ungated |
| `^on:\s+(\w+)` | Config | 1 (1) / 0 | 235 (235) / **0 (0)** | — | ungated |
| `^env:\s*\n\s+(\w+):` | Config | 0 / 0 | 35 (35) / **3 (3)** | 35 (35) / 0 (0) | **gated** |

Three of the six are plain YAML — a column-0 `name:`, a `- name:` list item, a column-0 `env:`
block — and the cicd bucket is aliased onto the whole `yaml` language
(`detector.go` `dormantBucketAliases`), so `(?m)` alone would type every helm `Chart.yaml`,
`pubspec.yaml`, `compose.yaml`, ansible task and travis config in a repo as a workflow part. They
get `requires_framework: true` plus a `frameworks.detection.import_markers` block
(`runs-on:`, `uses: actions/`, `${{`, `.github/workflows/`) — the #6152 falcon/cherrypy remedy.
The marker set selects **265 of the 269** workflow files and **zero of the 402** others; the 4 it
declines are not workflows (`dependabot.yml`, `labeler.yml`, `auto_assign.yml`,
`.pre-commit-config.yaml`, which merely live in `.github/`). The gate costs no recall: 264→264 for
the name rule, 35→35 for the env rule.

The other three are self-gating and stay ungated **because that was measured, not assumed**: the
job rule names `runs-on:`/`uses:` itself, and `on:` at column 0 is GitHub Actions' trigger key —
zero matches across all 402 non-workflow files. Their anchors are graded by forbidden rows instead.

**`$`**: none of the six carries one, so `(?m)`'s second effect cannot reach them. Asserted, not
claimed — `TestIssue6927_GitHubActions_NoDollarInWidenedPatterns` fails if a `$` is ever added to a
`(?m)` pattern in this file, and pins the count at 6.

## Graded in both directions, at two levels

New golden fixture **`github-actions-workflows-mini`** (5 files: two workflows, a helm chart, an
ansible playbook, a travis config): **25.0% entity recall on the pre-fix binary (3/12), 100% after**
(12/12 entities, 3/3 relationships, 0 forbidden hits). The 3 found pre-fix are exactly the three
un-anchored controls.

New engine-level test file **`github_actions_multiline_anchor_6927_test.go`**, driving the real
embedded YAML through `LoadAllRules()` + `Detect()` against a detector holding **only** this rule
set. It exists because the fixture structurally cannot see two things:

* **The step rule's blast radius.** `ansible_core.yaml` carries an **un-anchored** `- name:` → Task
  rule in the same `yaml` bucket, which subsumes this one outright, so those 194 over-fired Tasks
  are already in the graph from the other rule file and the gate is invisible at graph level. The
  gate is still right: it stops this file depending on another rule file's accident for its own
  blast radius, and #6927's ansible arm is going to narrow that pattern.
* **Which rule produced an entity** — the fixture asserts what the graph contains.

### Axes varied vs held constant

| axis | varied | held constant |
|---|---|---|
| vertical position | nothing on line 1 (7-line comment header); second job far down; `env:` block is the LAST construct in `nightly.yml` | — |
| anchor (column 0 vs indented vs commented) | a JOB's `name:` and a STEP's `- name:` (dash-less vs dashed — these are different mutants), job-level `env:` indented, `strategy.matrix` deeper, a commented-out step | — |
| framework gate | 3 non-workflow files naming GitHub Actions nowhere | the 2 workflow files satisfy the gate, so their forbidden rows grade the anchor alone |
| alternation branch | `runs-on:` job **and** `uses:` reusable-workflow job | — |
| trigger form | block `on:` and single-line `on:` | — |
| entity name / spelling | idiomatic throughout | held constant deliberately — the name axis is graded by `TestZeroProducingKinds6776_NoSiteOverFires` and by the two different-shaped rows per fence |

Anchor rows and gate rows are kept in **separate files** on purpose (the #6943 R3 lesson): a row
held by the gate says nothing about the anchor, so each fence has rows only it can hold, and each
has two rows of different shapes so no single accident of spelling carries a fence.

### Forbidden rows, each proven capable of firing

| row | fence | mutant that mints it |
|---|---|---|
| `Service:storefront-chart` (Chart.yaml) | name-rule gate | delete `requires_framework` |
| `Config:MATRIX_KEY` (legacy-ci.yml) | env-rule gate | delete `requires_framework` |
| `Service:Build and test` (a JOB's indented `name:`, inside build.yml) | name-rule anchor | `^name:` → `^\s*name:` |
| `Service:Compile assets` (a STEP's `- name:`, inside build.yml) | name-rule anchor | `^name:` → `name:` |
| `Config:BUNDLE_PATH` (job-level `env:`) | env-rule anchor | `^env:` → `^\s*env:` |
| `Operation:matrix` (`strategy.matrix.runs-on`) | job-rule indent | `\s{2}` → `\s+` |
| `Config:needs` (job named `notification:`, which ENDS in `on:`) | both `on:` anchors | drop `^` |
| `Config:20` (`node-version: 20` contains `on: 20`) | single-event `on:` anchor | drop `^` |
| `Task:Install nginx` / `Restart nginx`, `Service:language`, `Task:Deploy to staging` (commented step) | step-rule gate + anchor | engine-level only — see above |

## Findings the widening surfaced

1. **`.github` is on the walker's hard-coded skip list** (`internal/daemon/walk/walker.go`
   `hardcodedSkipDirs`), so a workflow stored where GitHub actually looks for it is never walked —
   these rules and this file's `.github/workflows/*.yml` `file_conventions` globs cannot fire on a
   real repo at all. The fixture stores its workflows at `ci/workflows/`, the layout GitHub's own
   `starter-workflows` repo uses. Filed separately.
2. **The job rule only looks at the immediately-next line**, so a job whose first key is `needs:`
   (or `if:`, `name:`, `permissions:`) is missed. Pinned as an *edge* rather than as prose: the
   fixture's `notification` job is asserted through the structured extractor's
   `CONTAINS` edge, which sees it while the regex rule does not.
3. **The single-event `on:` rule also reads the block form** — `\s` matches a newline — which is why
   its corpus count (235) sits one *above* the block-form rule's (234) rather than complementing it.
   The old `# on: single event` comment denied this. Now pinned by
   `TestIssue6927_GitHubActions_SingleEventRuleAlsoReadsBlockForm` against the shipped pattern text,
   because at entity level the two rules mint the same Config and the duplicate folds away.

### Mutants — scored at BOTH levels

Eighteen mutants, each applied to a pristine copy with the anchor-text occurrence count asserted
**exactly 1** (a silently-unapplied edit reads as ALIVE and argues the opposite, #6926), with a
green positive control before the first and after the last.

| mutant | engine suite | fixture gate |
|---|---|---|
| `^name:` → `^\s*name:` | DEAD | DEAD |
| `^name:` → `name:` (drop anchor) | DEAD | DEAD |
| name rule: drop `(?m)` | DEAD | DEAD |
| name rule: drop `requires_framework` | DEAD | DEAD |
| job rule: `\s{2}` → `\s+` | DEAD | DEAD |
| job rule: drop `(?m)` | DEAD | DEAD |
| step rule: drop `(?m)` | DEAD | **ALIVE** |
| step rule: drop `^` | DEAD | **ALIVE** |
| step rule: drop `requires_framework` | DEAD | **ALIVE** |
| block `on:`: drop `^` | DEAD | DEAD |
| block `on:`: drop `(?m)` | DEAD | **ALIVE** |
| single `on:`: drop `^` | DEAD | DEAD |
| single `on:`: drop `(?m)` | DEAD | DEAD |
| `^env:` → `^\s*env:` | DEAD | DEAD |
| env rule: drop `(?m)` | DEAD | DEAD |
| env rule: drop `requires_framework` | DEAD | DEAD |
| replace the marker set with one that matches nothing | DEAD | DEAD |

**All 18 DEAD at engine level. Four are ALIVE at the fixture gate**, and all four for reasons this
PR documents rather than discovers: the three step-rule mutants are invisible in the graph because
`ansible_core.yaml` mints those Tasks anyway, and `block on:` losing `(?m)` is masked because the
*single-event* rule reads the block form too and still mints `pull_request`. That masking is
precisely why the single-event rule's real behaviour is now pinned.

**One mutant was ALIVE at BOTH levels on first scoring and is now DEAD.** `^name:` → `^\s*name:`
survived every row: a step's `- name:` opens with a **dash**, so `\s*` cannot reach it, and the
`Service:Compile assets` row was passing for the wrong reason. What the widened anchor actually
claims is a **job's** `name:` — the display-name key, indented and dash-less, which every real
workflow may carry. `build.yml` now declares one and the fixture forbids `Service:Build and test`.
Two anchor rows for that fence, because the first one was not grading it.

### Ratchet delta

`scripts/quality/run.sh --ratchet` → **OK, 35 gated fixtures held their baseline** (26 at 100%
must-have recall), exit 0, no `GREW` line. The three known python drifts of #6898
(`python-django-mini`, `python-fastapi-mini`, `python-rq-mini`) did not appear — and that is **not
python recovering**, which is worth stating because "no drift" invites exactly that inference. The
new baseline row was written **by hand** rather than by `--update-baseline`, so no re-record
happened; #6898's drift is a *downward* movement of the extracted totals, `ratchet.py` fails on
`got > want` only, and a fall is silent by design. With no re-record there was no write for a drift
to appear in — the mechanism was **not exercised**, not satisfied. The baseline diff is purely
additive (one object, six keys), so nothing else could be re-recorded as a side effect.

### Review round (#6945 review, `24b2310d8`)

Independent review scored five mutants of its own and found **three fences that read as covered and
were not**. All three are closed; no shipped rule text changed.

| gap | before | now |
|---|---|---|
| no SINGLE `import_marker` was graded — deleting only `runs-on:` cost a `run:`-only workflow its Service and every Task | ALIVE at engine, fixture **and** ratchet | **DEAD** — one minimal workflow per marker, each asserting it carries exactly one marker first |
| the `(?:runs-on\|uses)` alternation — the entire justification for leaving the job rule ungated | ALIVE at engine and fixture; died only *incidentally* on the ratchet's generic growth ceiling | **DEAD at both** — forbidden `Operation:notification` plus an exact-set assertion |
| `SingleEventRuleAlsoReadsBlockForm` selected its rule by pattern **text** | red on a behaviour-**preserving** rewrite; its "kill" of a real anchor weakening was selector loss | rules selected by **behaviour**; the weakening `^on:` → `^\s*on:` now DEAD on its own test, and the behaviour-preserving control is **GREEN** |

The control is the one that matters: `(\w+)` → `([\w]+)` changes nothing and used to turn the suite
red with the same message a real break produced. A test that fails when you rename something and
passes for the wrong reason when you break something reads as coverage and is not.

Two caveats are now recorded beside the marker list rather than left implicit: **`${{` is Azure
Pipelines' template syntax too** (and an `azure-pipelines.yml` carries a column-0 `name:` *and* a
`- name:` variables list, so both gated rules would over-fire on one), and a **composite `action.yml`
whose steps are all `run:`** carries none of the four markers. Zero of 402 is a measurement, not a
proof.

## Verification

* `go test ./internal/engine/ ./internal/quality/... ./cmd/grafel/ -count=1` — green.
* `scripts/quality/run.sh --ratchet` — green at 35 gated fixtures, delta enumerated above.
* Corpus-size guards re-derived **from disk**, not from the diff: 35 fixture directories, 35
  baseline entries, guards moved 34 → 35 in `zero_relationships_6490_test.go`,
  `subtype_assertion_6488_test.go` and `absent_relationships_6490_test.go`.

Refs #6927, #6917, #6902, #6152, #6943.

### DIFF
--- a/internal/engine/github_actions_multiline_anchor_6927_test.go
+++ b/internal/engine/github_actions_multiline_anchor_6927_test.go
@@ -1,0 +1,429 @@
+package engine
+
+import (
+	"context"
+	"regexp"
+	"sort"
+	"strings"
+	"testing"
+
+	"github.com/cajasmota/grafel/internal/extractor"
+)
+
+// #6927 — the cicd arm. The same defect #6917 fixed for sinatra and #6943 for
+// python, in rules/cicd/frameworks/github_actions.yaml. detector.go compiles
+// rule patterns with plain regexp.Compile at :156 and :177, so `^` means START
+// OF TEXT. Six patterns were affected:
+//
+//	^name:\s+["']?([^"'\n]+)                       -> Service
+//	^\s{2}(\w[\w-]*):\s*\n\s+(?:runs-on|uses):     -> Operation
+//	^\s+-\s+name:\s+["']?([^"'\n]+)                -> Task
+//	^on:\s*\n\s+(\w+):                             -> Config
+//	^on:\s+(\w+)                                   -> Config
+//	^env:\s*\n\s+(\w+):                            -> Config
+//
+// Three of them could never fire AT ALL — `\s{2}`, `\s+-` and a preceding
+// `env:` block all require something before the match, and start-of-text
+// forbids it. The other three fired only on a file whose very first byte was
+// `name:` or `on:`.
+//
+// Two things are graded here that the golden fixture
+// (internal/quality/golden/github-actions-workflows-mini) structurally cannot
+// see, and that is why this file exists alongside it:
+//
+//  1. The Task rule's blast radius. ansible_core.yaml carries an UN-anchored
+//     `-\s+name:` -> Task rule in the same `yaml` bucket, so every Task the
+//     GitHub Actions rule could over-emit is already in the graph from the
+//     other rule file. At graph level the gate is invisible; against a
+//     detector holding ONLY this rule set it is not.
+//  2. Which RULE produced an entity. The fixture asserts what the graph
+//     contains; these tests assert what this one rule file contributes.
+//
+// Graded in BOTH directions (#6902), because a recall assertion is
+// structurally blind to over-firing.
+
+// ghaOnlyRules returns the loaded rule map reduced to the GitHub Actions rule
+// set alone. The rules are the REAL embedded YAML — the defect IS the shipped
+// pattern text, so a hand-written copy would leave the shipped one unobserved
+// — but every other yaml rule set is dropped, so nothing another rule file
+// mints can be mistaken for evidence about this one.
+func ghaOnlyRules(t *testing.T) map[string][]FrameworkRule {
+	t.Helper()
+	all, err := LoadAllRules()
+	if err != nil {
+		t.Fatalf("LoadAllRules: %v", err)
+	}
+	out := map[string][]FrameworkRule{}
+	for lang, frs := range all {
+		for _, fr := range frs {
+			if fr.Frameworks.Name == "GitHub Actions" {
+				out[lang] = append(out[lang], fr)
+			}
+		}
+	}
+	// The loader keys rule sets by their BUCKET directory (`cicd`); it is
+	// compile() that aliases the bucket onto the `yaml` language via
+	// dormantBucketAliases, which is why these tests hand Detect a
+	// Language of "yaml" while the map key stays "cicd". If that stops being
+	// true the tests are measuring nothing, so it is asserted rather than
+	// assumed.
+	if len(out["cicd"]) != 1 {
+		t.Fatalf("expected exactly one GitHub Actions rule set in the cicd bucket, got %d",
+			len(out["cicd"]))
+	}
+	return out
+}
+
+func detect6927GHA(t *testing.T, path, src string) *DetectResult {
+	t.Helper()
+	det := New(ghaOnlyRules(t))
+	res, err := det.Detect(context.Background(), extractor.FileInput{
+		Path:     path,
+		Content:  []byte(src),
+		Language: "yaml",
+	})
+	if err != nil {
+		t.Fatalf("Detect: %v", err)
+	}
+	return res
+}
+
+func has6927GHA(res *DetectResult, kind, name string) bool {
+	for _, e := range res.Entities {
+		if e.Kind == kind && e.Name == name {
+			return true
+		}
+	}
+	return false
+}
+
+func names6927GHA(res *DetectResult, kind string) []string {
+	var out []string
+	for _, e := range res.Entities {
+		if e.Kind == kind {
+			out = append(out, e.Name)
+		}
+	}
+	sort.Strings(out)
+	return out
+}
+
+// gha6927Workflow is the fixture's build.yml, held here as a const so the
+// engine-level assertions and the golden fixture describe the SAME source. It
+// opens with a comment header — the shape every real workflow has, and the
+// reason start-of-text found a `name:` in only 99 of the 269 workflow files in
+// the measured corpus.
+const gha6927Workflow = `# Storefront build pipeline.
+#
+# Deliberately NOT under .github/.
+name: Storefront CI
+on:
+  pull_request:
+    branches: [main]
+  push:
+    branches: [main]
+env:
+  REGISTRY: ghcr.io
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    name: Build and test
+    env:
+      BUNDLE_PATH: vendor/bundle
+    strategy:
+      matrix:
+        runs-on: [ubuntu-latest, macos-latest]
+    steps:
+      - uses: actions/checkout@v4
+      - name: Compile assets
+        run: make assets
+      # - name: Deploy to staging
+      #   run: make deploy
+      - uses: actions/setup-node@v4
+        with:
+          node-version: 20
+      - name: Upload bundle
+        run: make upload
+  publish:
+    runs-on: ubuntu-latest
+    needs: build
+    steps:
+      - uses: actions/upload-artifact@v4
+      - name: Push image
+        run: docker push "$REGISTRY/storefront"
+  notification:
+    needs: [build, publish]
+    runs-on: ubuntu-latest
+    steps:
+      - name: Post build result
+        run: ./ci/notify.sh
+`
+
+// gha6927ReusableCaller exercises the `uses:` half of the job rule's
+// alternation: a reusable-workflow call has no `runs-on:` anywhere in the job.
+const gha6927ReusableCaller = `# Nightly smoke run.
+name: Nightly Smoke
+on: workflow_dispatch
+jobs:
+  smoke:
+    uses: ./.github/workflows/build.yml
+env:
+  SMOKE_REGION: eu-west-1
+`
+
+// TestIssue6927_GitHubActions_WorkflowYieldsJobsTriggersAndEnv is the recall
+// direction. Vertical position is the axis it varies: nothing sits on line 1,
+// the second job is far down the file, and the reusable caller's `env:` block
+// is the last construct in its file.
+func TestIssue6927_GitHubActions_WorkflowYieldsJobsTriggersAndEnv(t *testing.T) {
+	res := detect6927GHA(t, "ci/workflows/build.yml", gha6927Workflow)
+
+	for _, want := range []struct{ kind, name string }{
+		{"Service", "Storefront CI"},
+		{"Operation", "build"},
+		{"Operation", "publish"},
+		{"Config", "pull_request"},
+		{"Config", "REGISTRY"},
+		{"Task", "Compile assets"},
+		{"Task", "Upload bundle"},
+		{"Task", "Push image"},
+		{"Task", "Post build result"},
+	} {
+		if !has6927GHA(res, want.kind, want.name) {
+			t.Errorf("%s %q not extracted from a realistic workflow; Services=%v Operations=%v "+
+				"Configs=%v Tasks=%v", want.kind, want.name,
+				names6927GHA(res, "Service"), names6927GHA(res, "Operation"),
+				names6927GHA(res, "Config"), names6927GHA(res, "Task"))
+		}
+	}
+
+	res2 := detect6927GHA(t, "ci/workflows/nightly.yml", gha6927ReusableCaller)
+	for _, want := range []struct{ kind, name string }{
+		{"Service", "Nightly Smoke"},
+		// The `uses:` half of `(?:runs-on|uses)`. Without this the alternation
+		// is graded on one branch only.
+		{"Operation", "smoke"},
+		// `on: <event>` on one line, which the block-form rule cannot reach.
+		{"Config", "workflow_dispatch"},
+		// Last construct in the file — the far end of the vertical axis.
+		{"Config", "SMOKE_REGION"},
+	} {
+		if !has6927GHA(res2, want.kind, want.name) {
+			t.Errorf("%s %q not extracted from a reusable-workflow caller; Services=%v "+
+				"Operations=%v Configs=%v", want.kind, want.name,
+				names6927GHA(res2, "Service"), names6927GHA(res2, "Operation"),
+				names6927GHA(res2, "Config"))
+		}
+	}
+}
+
+// TestIssue6927_GitHubActions_AnchorsHoldInsideAWorkflow is the forbidden
+// direction on the ANCHOR axis, inside a file that SATISFIES the framework
+// gate. Every row here is held by `^` alone; none of them is held by
+// requires_framework, which is what keeps the two fences distinguishable.
+func TestIssue6927_GitHubActions_AnchorsHoldInsideAWorkflow(t *testing.T) {
+	res := detect6927GHA(t, "ci/workflows/build.yml", gha6927Workflow)
+
+	for _, bad := range []struct{ kind, name, why string }{
+		{"Service", "Compile assets",
+			"`^name:` is column-0 only; a step's indented `name:` is a Task, not the workflow"},
+		{"Service", "Push image", "same, for a step in the second job"},
+		{"Service", "Build and test",
+			"a JOB's indented `name:` — the display-name key every workflow may carry. This is " +
+				"the row that kills `^name:` -> `^\\s*name:`: a step's `- name:` opens with a " +
+				"dash, so `\\s*` alone cannot reach it and the mutant survived every other row"},
+		{"Config", "BUNDLE_PATH",
+			"`^env:` distinguishes the WORKFLOW-level env block from the job-level one this key sits in"},
+		{"Operation", "matrix",
+			"`\\s{2}` is exactly two; `strategy.matrix` is deeper and is not a job"},
+		{"Operation", "strategy", "same, one level up"},
+		{"Config", "needs",
+			"`notification:` is a legal job name ENDING in `on:`; without `^` both on: rules read the next key"},
+		{"Config", "20",
+			"`node-version: 20` contains the substring `on: 20`; without `^` the single-event rule mints it"},
+		{"Task", "Deploy to staging",
+			"a commented-out step. `^\\s+-` requires the dash to open the line's content"},
+	} {
+		if has6927GHA(res, bad.kind, bad.name) {
+			t.Errorf("over-fire: %s %q was extracted from a workflow — %s", bad.kind, bad.name, bad.why)
+		}
+	}
+}
+
+// gha6927HelmChart, gha6927AnsiblePlay and gha6927TravisConfig are plain YAML.
+// None names GitHub Actions anywhere, and each carries one of the three shapes
+// the widened patterns would otherwise claim.
+const gha6927HelmChart = `apiVersion: v2
+name: storefront-chart
+description: Storefront deployment chart
+type: application
+version: 0.4.1
+`
+
+const gha6927AnsiblePlay = `- hosts: web
+  become: true
+  tasks:
+    - name: Install nginx
+      apt:
+        name: nginx
+        state: present
+    - name: Restart nginx
+      service:
+        name: nginx
+        state: restarted
+`
+
+const gha6927TravisConfig = `language: ruby
+rvm:
+  - 3.1
+env:
+  MATRIX_KEY: fast
+script:
+  - bundle exec rake
+`
+
+// TestIssue6927_GitHubActions_GateHoldsPlainYAML is the forbidden direction on
+// the FRAMEWORK-GATE axis, and the reason this change is not `(?m)` alone.
+//
+// `^name:`, `^\s+- name:` and `^env:` are plain YAML with nothing about GitHub
+// Actions in them, and Detect resolves rule sets by file.Language while the
+// cicd bucket is aliased onto `yaml`, so under `(?m)` alone every chart,
+// playbook and CI config in a repo would be typed as workflow parts. Measured
+// on 671 real YAML files from 59 upstream repos: 34 `^name:` matches in 34
+// non-workflow files, 194 step matches in 49, 3 env matches in 3 — against
+// ZERO for all three once `requires_framework: true` is on.
+//
+// The Task half of this cannot be graded by the golden fixture at all:
+// ansible_core.yaml's un-anchored `- name:` rule mints those Tasks anyway, so
+// the graph looks identical with the gate on or off. Here it does not.
+func TestIssue6927_GitHubActions_GateHoldsPlainYAML(t *testing.T) {
+	for _, tc := range []struct {
+		path, src string
+		bad       []struct{ kind, name string }
+	}{
+		{"deploy/helm/Chart.yaml", gha6927HelmChart, []struct{ kind, name string }{
+			{"Service", "storefront-chart"},
+		}},
+		{"ops/provision.yml", gha6927AnsiblePlay, []struct{ kind, name string }{
+			{"Task", "Install nginx"},
+			{"Task", "Restart nginx"},
+		}},
+		{"ops/legacy-ci.yml", gha6927TravisConfig, []struct{ kind, name string }{
+			{"Service", "language"},
+			{"Config", "MATRIX_KEY"},
+		}},
+	} {
+		res := detect6927GHA(t, tc.path, tc.src)
+		for _, bad := range tc.bad {
+			if has6927GHA(res, bad.kind, bad.name) {
+				t.Errorf("over-fire: %s %q extracted from %s, which names GitHub Actions "+
+					"nowhere — requires_framework is not holding", bad.kind, bad.name, tc.path)
+			}
+		}
+		if len(res.Entities) != 0 {
+			var got []string
+			for _, e := range res.Entities {
+				got = append(got, e.Kind+":"+e.Name)
+			}
+			sort.Strings(got)
+			t.Errorf("%s: the GitHub Actions rule set extracted %d entities from plain YAML: %v",
+				tc.path, len(res.Entities), got)
+		}
+	}
+}
+
+// ghaSourcePatterns returns the shipped GitHub Actions source_patterns.
+func ghaSourcePatterns(t *testing.T) []SourcePattern {
+	t.Helper()
+	return ghaOnlyRules(t)["cicd"][0].SourcePatterns
+}
+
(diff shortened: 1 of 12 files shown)


cajasmota:
Filed the `.github` finding as #6946 — measured, not read: two byte-identical workflow files, one at `src/.github/workflows/probe.yml` and one at `src/ci/workflows/probe.yml`, and the indexer discovered **1** candidate file, not 2. `.github` is in `hardcodedSkipDirs`, so this rule file's three `.github/...` `file_conventions` globs and every pattern in it are unreachable on a repo that stores workflows where GitHub requires them. That is why this fixture's workflows live at `ci/workflows/` (the `starter-workflows` layout) rather than under `.github/`.

cajasmota:
## Independent adversarial review — REQUEST-CHANGES (three cheap asks; the change itself is right)

Own detached worktree at `3f662dcc1`, own `HOME`/`GRAFEL_HOME`/`GRAFEL_DAEMON_ROOT`. `go vet ./internal/engine/` clean. Green positive control before the first mutant (`go test ./internal/engine/ -count=1` exit 0; `grafel quality` on the new fixture exit 0, 12/12, 3/3, 0 forbidden, totals 55/72 — the recorded baseline row is exact) and after the last. Every mutant applied by exact-count-1 replacement, restored by `cp` from a shasum-verified backup (`792ac5698…` before and after each), tree `git status --porcelain` empty at the end.

### Mutants of my own — 5, scored at three levels

The third level matters: `grafel quality` on a single fixture does **not** apply the `entity_extracted_total` ceiling; `scripts/quality/ratchet.py:617` (`if got > want: … GREW`) does. So a mutant that mints one extra entity dies at the ratchet even when the arm's own forbidden rows never see it.

| # | mutant (call site, in the shipped YAML) | `./internal/engine/` | fixture gate | ratchet ceiling (55/72) |
|---|---|---|---|---|
| M1 | **add** `- "name:"` to `import_markers` (marker that matches ~everything) | **DEAD** — `TestIssue6927_GitHubActions_GateHoldsPlainYAML` | **DEAD** — exit 2, `storefront-chart [Service] in deploy/helm/Chart.yaml` | DEAD (56 > 55) |
| M2 | **delete only** `- "runs-on:"` from `import_markers` | **ALIVE** (exit 0) | **ALIVE** (exit 0, 12/12, 0 forbidden) | ALIVE (55/72 unchanged) |
| M3 | job rule `(?:runs-on\|uses):` → `(?:runs-on\|uses\|needs):` | **ALIVE** (exit 0) | **ALIVE** (exit 0, 0 forbidden — but totals moved 55→**56**, 72→**73**) | DEAD, *incidentally* (56 > 55) |
| M4 | single-event rule `(?m)^on:` → `(?m)^\s*on:` | red, but **not a kill** — see M5 | **ALIVE** (exit 0, 55/72) | ALIVE |
| M5 | control for M4: `(?m)^on:\s+(\w+)` → `(?m)^on:\s+([\w]+)` — **behaviour-preserving** | **RED**, same line, same message | (not run) | (not run) |

**M2 — each marker is individually ungraded, in the recall direction, and it is production-reachable.** The only marker mutant scored in the PR is "replace the set with one that matches nothing". The set is over-determined on the fixture: `build.yml` satisfies three markers, `nightly.yml` satisfies `.github/workflows/` via its `uses:` target — so dropping any single marker changes nothing anywhere. The input that exercises it is the most ordinary workflow there is:

```yaml
name: CI
on: push
jobs:
  build:
    runs-on: ubuntu-latest
    steps:
      - run: make
```

`runs-on:` is its only marker. Under M2 the gate declines it and the `name:` Service and every step Task are lost — and the PR's own corpus number (265/269) is the claim that this would break. Ask: four one-line engine cases, one minimal file per marker.

**M3 — the alternation is the entire justification for leaving the job rule ungated, and nothing pins it.** The PR argues the job rule needs no `requires_framework` because "the regex names `runs-on:`/`uses:` itself". Adding a third key to that alternation walks straight past the arm's own fences: it mints `Operation:notification` (totals 55→56) with zero forbidden hits. It dies only on the generic growth ceiling — which fires for *any* extra entity and would not fire at all for a same-count substitution. Finding #2 in the body ("a job starting with `needs:` is missed, pinned as an edge") pins the *structured extractor's* `CONTAINS` edge; it says nothing about the regex rule's blast radius. Ask: forbid `Operation:notification` (or an engine-level assertion that the job rule mints exactly `build`/`publish`/`smoke`).

**M4/M5 — `TestIssue6927_GitHubActions_SingleEventRuleAlsoReadsBlockForm` selects by text substring, so its verdict is not about behaviour.** It finds the rule with `strings.Contains(sp.Pattern, "^on:\\s+(\\w+)")` and `t.Fatal`s when that misses. M4 weakens the column-0 anchor to `^\s*on:` — a real, permissive behaviour change on an **ungated** rule that then fires on an indented `on:` in any YAML file in any repo — and the test goes red with *"the single-event on: pattern is no longer in github_actions.yaml"*, i.e. it lost its selector, not its behaviour. M5 proves that directly: `(\w+)` → `([\w]+)` changes nothing at all and produces the identical failure. So the anchor's partial weakening is behaviourally ungraded, and the test is a false-alarm on any innocuous rewrite. Ask: select the pattern by index/entity behaviour and assert the block-form read, so the `t.Fatal` means "no rule reads the block form" rather than "my substring moved".

Everything else I attacked held. `TestIssue6927_GitHubActions_NoDollarInWidenedPatterns` really fails on a 7th `(?m)` pattern — I added `(?m)^permissions:\s+(\w+)$` and **both** halves fired (the `$` error and `found 7 … want 6`). The forbidden-row dialect is fine: M1's hit came back fully resolved (`storefront-chart [Service] in deploy/helm/Chart.yaml`, id present), so these rows are not silently unfireable the way #6938's fences were. Anchor rows do live inside gate-satisfying files and gate rows inside non-workflow files — I read the five fixture bodies against the axes table, and it holds.

### #6946 — the claim is right, and stronger than the issue states

Reproduced independently (`walk.WalkRepo` on a temp tree, two identical files): `.github/workflows/ci.yml` and `.github/actions/x/action.yml` are **not returned**; `ci/workflows/ci.yml` is. Skip entry: `.github (hardcoded)`.

"The walker skips it" and "grafel never sees it" are different claims, and the second one holds too:

* **No re-admission by config.** `hardcodedSkip` runs at `walker.go:214`, **before** the gitignore/`.grafelignore` stack at `:226`. A `.grafelignore` carrying `!.github` / `!.github/**` re-admits nothing — measured, in the same run. `AdditionalSkipDirs` only ever adds.
* **All four enumeration paths are the same walk.** `WalkRepo` has exactly four non-test callers — `cmd/grafel/index.go`, `internal/extractors/incremental.go`, `internal/daemon/watch/change_poller.go`, and the walker itself.
* **The git-derived paths are intersected with the walked set.** `incremental.go`'s head-advance union takes `rangeChanged` from `git diff --name-only old..new` and admits `f` only `if allFilesSet[f]`, so a changed `.github/workflows/ci.yml` cannot enter the changed-file set.
* **The watcher drops it at the event boundary.** `watch/skip.go` `ShouldSkipPath` splits the path and calls `ShouldSkipDir` on every segment, which delegates to `walk.IsHardcodedSkip`.
* **Gate is content-only.** `compiledRuleSet.frameworkPresent` does `strings.Contains(content, m)`; the path never participates, so `file_conventions` cannot re-admit anything either.

One genuine escape, worth a line in the issue: if `.github` is registered as the repo/package **root** itself, its contents are walked (measured: root=`.github` yields `workflows/ci.yml`) — but the rel path is then `workflows/ci.yml`, which the `.github/workflows/*.yml` `file_conventions` glob does not match. So the finding stands: no workflow file stored where GitHub requires it is reachable today.

### Are the three ungated patterns safely ungated?

As shipped, yes — but say what the zero is over. The 402 is a CI/k8s/helm/app-repo corpus; the *population* the rules face is not corpus-shaped and is not constrained at all: `dormantBucketAliases` aliases `cicd` onto `yaml`, so these rules are offered **every** `.yml`/`.yaml` in every indexed repo, and the two `on:` rules and the job rule carry no `requires_framework`. The anchors are the only fence, and per M3/M4 that fence is graded only against *total* removal of `^` (or of the alternation), not against weakening it. That is the gap I am asking you to close; I am not asking you to gate the three.

Two caveats on the marker set, non-blocking, for the ansible arm to keep in view:

* **`${{` is not GitHub-only.** Azure Pipelines template expressions are literally `${{ … }}`, and an `azure-pipelines.yml` carries a column-0 `name:` (build-number format) and a `- name:` variables list — i.e. both gated rules would over-fire on it. Zero of 402 is real; it is not "cannot".
* **A composite `action.yml` with only `run:` steps** satisfies none of the four markers (no `runs-on:` in a composite action). The 4 declined files you name are genuinely not workflows — `dependabot.yml`, `labeler.yml`, `auto_assign.yml`, `.pre-commit-config.yaml` — but the declined set is not obviously closed at 4 outside this corpus.

### #6944 guard collision — none, and the premise is wrong

Re-derived from disk, not from the diff: 35 fixture directories, 35 `baseline.json` entries, and `want 35` in all three guards (`zero_relationships_6490_test.go:181`, `absent_relationships_6490_test.go:233`, `subtype_assertion_6488_test.go:292`). `origin/main` is at 34 dirs / `want 34`, so 34→35 is correct against the branch this will land on.

**#6944 does not touch these files at all** — it is the OCaml tree-sitter arm (`internal/extractors/ocaml/*`, 4 files). I swept every open PR for `internal/quality/golden|_6490_test|subtype_assertion` and #6945 is the only one. So there is no identical-textual-edit hazard in flight; the guards will still be semantically right whatever order things merge in. If a second fixture-adding PR opens before this lands, re-derive again — a clean merge of two identical 34→35 edits is exactly the failure mode.

### The ratchet drift, explained rather than confirmed

"The three known python drifts did not appear" is not a recovery, and the PR body already contains the reason without drawing it: the new baseline row was written **by hand**, not by `--update-baseline`. #6898's drift is a *downward* movement of `entity_extracted_total`/`relationship_extracted_total` that becomes visible only when a re-record writes the lower figure; `ratchet.py` fails on `got > want` only, and a fall is silent by design. With no re-record there is no write for a drift to appear in — the mechanism was not exercised, not satisfied. Worth phrasing that way in the body, because "no drift at all" reads as "python recovered", and nothing here measured that. The baseline diff is purely additive (one object, six keys), so I agree nothing else was re-recorded.

### Verdict

The fix is correct, the measurement is the right measurement, and the two-level scoring found a real ALIVE-at-both mutant and fixed it — that is the standard this board wants. **REQUEST-CHANGES only on the three grading gaps above** (M2: per-marker; M3: the alternation; M4/M5: the substring selector), each of which is a handful of lines and none of which changes the shipped rule text. #6946 is confirmed independently and I would strengthen its title-level claim rather than weaken it.

cajasmota:
## All three addressed in `24b2310d8`. No shipped rule text changed.

Every one of the three was a fence that read as covered and was not, and I accept all three readings — particularly M4/M5, which is the sharper finding: a test that goes red when you rename something and green-for-the-wrong-reason when you break something is worse than no test. That one was mine to catch and I did not.

### 1. Each `import_marker` is now graded on its own (M2)

`TestIssue6927_GitHubActions_EachImportMarkerAdmitsAWorkflowAlone` — one minimal workflow per marker, and each case **asserts it carries exactly one marker before asserting the gated entity survives**, so a case that quietly grows a second marker fails as "it grades the SET, not this marker" rather than passing. Your `run:`-only workflow is case 1 verbatim (plus a named step, so the Task half is graded too). The test also pins the marker count at 4, so a fifth marker is ungraded loudly instead of silently.

### 2. The alternation is graded as an exact set (M3)

Two rows now, at both levels: a forbidden `Operation:notification` in the fixture, and an engine-level **exact-set** assertion that the job rule mints exactly `build`/`publish` in `build.yml` and exactly `smoke` in `nightly.yml`. Exact rather than membership, precisely because a membership check is blind to the direction this mutant moves in — and because the ratchet ceiling that killed it incidentally would not fire on a same-count substitution at all.

### 3. The `on:` rules are selected by behaviour, and the anchor is graded against *weakening* (M4/M5)

`gha6927OnRules` classifies the compiled patterns by what they **do** — which one reads a single-line `on: push`, which reads only the block form — with no substring anywhere. The two canonical documents carry nothing but the trigger, because a `name:` line makes the workflow-name rule answer to the same probe (found while writing it). Then:

* `SingleEventRuleAlsoReadsBlockForm` asserts the behaviour, so `t.Fatal` now means "no rule reads the block form".
* New `TestIssue6927_GitHubActions_OnRulesAreColumnZeroOnly` grades the half the forbidden rows never reached: `Config:needs` / `Config:20` grade *removing* `^`; this grades *weakening* it, on both rules and both trigger shapes.

### Re-scored — your five, plus your control

| mutant | engine | fixture | before |
|---|---|---|---|
| delete only `- "runs-on:"` | **DEAD** | ALIVE | ALIVE at all three |
| `(?:runs-on\|uses)` → `(?:runs-on\|uses\|needs)` | **DEAD** | **DEAD** | ALIVE at both, ratchet-incidental |
| single `on:` → `^\s*on:` | **DEAD** | ALIVE | selector loss, not a kill |
| block `on:` → `^\s*on:` | **DEAD** | ALIVE | ungraded |
| **control**: `(\w+)` → `([\w]+)` (behaviour-preserving) | **GREEN** | GREEN | RED — false alarm |

The control is the one that matters: the old test failed on it identically to a real weakening. It now passes, so the suite is grading behaviour rather than spelling.

### Caveats recorded, not waved through

Both are in the rule comment beside the marker list, with the shape named: `${{` is Azure Pipelines' template syntax too and an `azure-pipelines.yml` carries both a column-0 `name:` and a `- name:` variables list, so **both gated rules would over-fire on one**; and a composite `action.yml` whose steps are all `run:` carries none of the four (a composite action has no `runs-on:`), so it is declined and its `name:` is then extracted by nothing. The comment now says "zero of 402 is a measurement, not a proof" rather than leaving the set looking closed.

### The ratchet drift phrasing — you are right, and I have corrected the body

"No drift at all" invited exactly the inference you name. The new baseline row was written **by hand**, so no re-record happened; #6898's drift is a *downward* movement of the extracted totals and `ratchet.py` fails on `got > want` only, so a fall is silent by design and there was no write for it to appear in. The mechanism was **not exercised**, not satisfied. Body updated to say that.

### #6946 strengthened

Your five-path argument and the root-registration escape are now [a comment on the issue](https://github.com/cajasmota/grafel/issues/6946#issuecomment-5562319345) — including the part that makes the escape not a rescue: root = `.github` does walk `workflows/ci.yml`, but `matchesFile` then tests that rel path against the `.github/workflows/*.yml` glob, which does not match (and the `path.Base` fallback is slash-free globs only).

And thank you for re-deriving the #6944 guard premise from disk rather than taking the brief's word for it — 34 → 35 stands, and the "if a second fixture-adding PR opens before this lands, re-derive again" note is the right standing instruction.

### Verification after the changes

`gofmt`/`go vet` clean. `go test ./internal/engine/ ./internal/quality/... ./cmd/grafel/ -count=1` → exit 0. `scripts/quality/run.sh --ratchet` → **OK, 35 gated fixtures held**, exit 0, no `GREW` line. Fixture unchanged at 12/12 entities, 3/3 relationships, 0 forbidden hits, totals 55/72 — so the recorded baseline row is still exact.

cajasmota:
Coordinator mutant on the final head `24b2310d8`. One mutant, **DEAD twice over** — and the notable thing is that the rule comment predicted it by name.

## MG-1 — widen the job rule's indent precision

`(?m)^\s{2}(\w[\w-]*):\s*\n\s+(?:runs-on|uses):` → `(?m)^\s+(...)`. Exactly-two-spaces becomes any depth, so keys nested arbitrarily deep under `jobs:`, `strategy:` or `with:` become candidate jobs.

I chose it because every mutant so far — your 18, the review's 5 — attacked the **alternation** (`runs-on|uses`), the **markers**, or the **`^` anchor**. Nobody had touched the **indent width**, which is the third thing making this rule specific and the one carrying "a job is a top-level key under `jobs:`". Anchor count 1 pre-edit, `go vet` clean, baseline green first (`ok 50.529s`), `-count=1`, restored by `cp` from a shasum-verified backup (`d70d5e9e4…`), `git status --porcelain` empty.

**Verdict: DEAD, by two independent named assertions:**

```
over-fire: Operation "matrix" was extracted from a workflow —
`\s{2}` is exactly two; `strategy.matrix` is deeper and is not a job

job rule minted Operations [build matrix publish], want exactly [build publish].
`notification:` is a job whose first key is `needs:`, so the rule must NOT see it
```

## Why this one is worth reporting rather than just ticking

The rule comment says, before I ran anything:

> `\s{2}` — exactly two — is what keeps it off deeper keys such as a `strategy.matrix` whose axis is named `runs-on`; **widening it to `\s+` mints an Operation called `matrix`, which is a forbidden row in `github-actions-workflows-mini`.**

That is the mutant, its consequence, and its grader, named in advance. It is the exact inverse of the defect class this board keeps hitting — a comment asserting a hazard that nothing observes. Here the comment asserts a hazard, names what catches it, and the claim survives being driven.

**That is the standard the remaining four files of #6927 should copy**, and it is worth saying explicitly because two of the three gaps the review found on this very PR were the opposite shape: a fence that read as coverage and was not.

The second kill is the better of the two, incidentally: the exact-set assertion (`want exactly [build publish]`) catches the widening *and* would catch a same-count substitution, which the growth ceiling never would. Membership would have missed it. Your choice of exact-set over membership there is doing real work.

## On the round as a whole

The control passing is the proof that mattered. `(\w+)` → `([\w]+)` is behaviour-preserving, the old substring-selecting test failed on it **identically to a genuine break**, and the new structural selector passes it while still killing all four real weakenings. A test that cannot tell a rename from a break is worse than no test, because it reads as coverage — and the only way to find that is to run a control you expect to pass.

Also accepted: each `import_marker` now graded alone, with each case asserting it carries **exactly one** marker first, so a case that silently grows a second fails as "grades the SET, not this marker" rather than passing. That guard-on-the-guard is the same shape as the positive control in #6941's purity test, and it is the right instinct.

**#6946** is now airtight, including the root-registration escape that looks like a rescue and is not: registering `.github` as a root does walk `workflows/ci.yml`, but `matchesFile` then tests that rel path against `.github/workflows/*.yml`, which does not match.

Merging on green. Refs #6927, #6946.

Issue #6947, opened 2026-09-06
MEASURE FIRST: OCaml inheritance across files/modules emits NO edge — the gate is deliberate, the population is unmeasured


cajasmota:
#6370's OCaml arm emits a hierarchy edge only when the parent class is declared in the SAME FILE. Two shapes produce nothing, and neither is exotic:

```ocaml
(* 1. module-qualified parent *)
class relay = object
  inherit Remote.printer
end

(* 2. unqualified parent declared in another file / brought in by `open` *)
open Shapes
class circle = object
  inherit shape
end
```

Both are deliberate under the arm's constraint — `internal/quality/analytics/scan.go` (`accumulateRel`) keys `touched[from]` unconditionally, so a **dangling** edge de-orphans its source exactly as well as a real one, and the orphan metric cannot tell them apart. The arm therefore refused to emit anything it could not itself show would bind, which for a per-file extractor means "this file declares the class". `internal/extractors/ocaml/hierarchy.go` states the rule and the cost in its package comment; the fixture `ocaml-objects-mini` fences both shapes (`relay_printer --[EXTENDS]--> printer` for (1), `remote_relay --[EXTENDS]--> external_base` for (2)), each demonstrated firing under its own mutant.

## What is NOT known, and should be measured before it is fixed

**The population.** Nobody has measured what fraction of real OCaml `inherit` clauses name a parent in another file or module. That number decides whether this is a rounding error or the majority of the construct, and it is cheap to get: #6812 left an env-gated corpus harness in `internal/extractors/ocaml/depth_walker_inherit_accuracy_6812_test.go` that already walks a clone of github.com/ocaml/ocaml at 70165ff23e and finds every `inherit`. Extend it to classify each clause as same-file / unqualified-cross-file / module-qualified, and report the three counts. **Do that before designing anything** — a fix aimed at (1) when the population is (2) buys nothing.

**Whether (1) can be resolved at all.** `inherit Remote.printer` needs `Remote` mapped to a compilation unit, which is what the module system does and this extractor does not model (`#6939` (1) records that a functor is not even extracted). Taking the last segment is NOT an option and is the specific thing the arm refused: this fixture declares a class type named `printer`, so `Remote.printer` reduced to `printer` would produce an edge that exists, resolves, and points at the wrong node — #6369's standing example.

[...]

PR #6948, opened 2026-09-06
feat(#6370): OCaml emitted no EXTENDS/IMPLEMENTS — arm 7 of 8, CST-backed off the grammar #6812 wired

cajasmota:
OCaml is arm 7 of the eight #6370 found. It had no inheritance topology by **any** of the three producer paths, and the third is the one a Go-only grep does not see: its own extractor suppressed `inherit` as a keyword, it is absent from `supportedLanguages` in `cross/hierarchy`, and **no YAML rule pack under `internal/engine/rules/` mentions it** — checked, because the graphql arm found that path the hard way.

The blocker was #6812, which closed in `18652763f`: `inherit` attribution went from **48.3% precision (230 of 408 wrong) to 100%** when the vendored tree-sitter grammar was wired. An edge built on the old regex walker would have been wrong about half the time.

## What resolves, and what does not

`internal/extractors/ocaml/hierarchy.go` emits:

| construct | edge |
|---|---|
| `class circle = object inherit shape … end` | `circle --[EXTENDS]--> shape` |
| `class type verbose_printer = object inherit printer … end` | `verbose_printer --[EXTENDS]--> printer` |
| `class basic_printer : printer = object … end` | `basic_printer --[IMPLEMENTS]--> printer` |

and deliberately emits **nothing** for:

- `inherit Remote.printer` — a module-qualified parent. Its last segment is `printer`, which this fixture declares, so reducing a path to its final identifier would produce an edge that exists, resolves, and points at the **wrong node** (#6369).
- `inherit external_base` where nothing in the file declares `external_base` — no entity, no edge.
- `let make_logger prefix = object inherit stream_writer … end` — an anonymous object has no class binding, so there is nothing to anchor a hierarchy edge on. The parent name still reaches the graph as the CALLS edge #6938 pinned.
- `include`, `open`, functor application — separate grammar productions that cannot reach the producer. Fenced anyway: "cannot happen" is an argument, a forbidden row is a measurement.

Cross-file and cross-module inheritance is therefore **not** covered. That is the arm's chosen cost, not an oversight, and it is filed as **#6947** with the measurement that has to come before anyone relaxes it — including the note that a single-corpus reading is corpus-relative.

Both guards are graded by unit tests that reduce a target past the last `:`, so they hold whatever spelling `ToID` carries, and the qualified-path guard is graded **arm by arm**: a class qualifies with `module_path`, a class type with `extended_module_path`, and deleting only the second left the whole suite green until a `class type d = object inherit Make(Int).printer end` case was added.

**Why the gate lives in the producer.** `internal/quality/analytics/scan.go` (`accumulateRel`) keys `touched[from]` unconditionally, so a **dangling** edge de-orphans its source exactly as well as a real one. The orphan metric improves either way and can never be evidence the edge works — so "emit only when the target resolves" is enforced where it is checkable, and no orphan figure is quoted anywhere in this PR.

## This arm had to introduce the class vocabulary

#6939 (2) records that no `class` construct produced any entity at all — the extractor's vocabulary was module / type / let. A hierarchy edge needs a FROM node and #6370's bar is that it anchors on the **type**, so the twelve new class entities are a precondition of the six edges rather than a separate feature. `class` → subtype `class`; `class type` → subtype `class_type`, **not** `interface`: a class type is OCaml's interface-shaped construct and is what IMPLEMENTS points at, but it is also legal as a plain type abbreviation, and the relation is what makes it findable. The other three #6939 gaps (functors invisible, dead module CONTAINS, wrong module spans) are untouched.

No regex was added: #6812's owner decision forbids it, and the CST makes every distinction here a node type rather than a judgement.

## The `ToID` design, and the two wrong turns on the way to it

`ToID` is the entity's **Format A structural ref**, not the bare written name the Solidity / Crystal / F# / Groovy arms use. Those emit a bare name because their parent usually lives in another file, so a file-pinned ref would never bind. Rule (2) inverts that premise: this producer only ever names a class it has just declared in the **same** file, so the ref always binds and is strictly more specific.

It took two corrections to get there, and both are recorded in the package comment because each is easy to repeat.

**First wrong turn — "the ref is ungradeable".** With a ref, deleting `if !declared[p.target] { continue }` emits `scope:component:class:ocaml:<file>:external_base`, which no *portable* forbidden row can name; the fence `remote_relay --[EXTENDS]--> external_base` reported 0 hits with the whole suite green. That premise is true. The conclusion was not: a fixture row is not the only instrument. `TestOCamlHierarchy_UndeclaredParentYieldsNoEdge_6370` kills the same mutant under **either** spelling, because `hier()` reduces a target past the last `:`. "The fixture cannot express it" is a fact about the fixture.

**Second wrong turn — "the bare name is sufficient".** Measured, on two `.ml` files each declaring `class shape` with a local subclass, in **both file orders**:

| `ToID` | bound to own file | cross-file mis-binds |
|---|---|---|
| bare name | **0 / 2** | 0 |
| structural ref | **2 / 2** | 0 |

The bare name never binds to the *wrong* file — ambiguity leaves the edge unresolved, so #6369's wrong-node failure does not occur either way — but it **silently loses the edge**, and same-named classes across files are ordinary in OCaml, where a module's classes are named for their role (`t`, `printer`, `writer`).

The fixture fence survives the choice rather than being dropped: spelled with the ref, it fires under the gate-deletion mutant, because the ref is the only string such an edge can carry. So the gate carries two independent instruments — a fixture row that names the dialect and breaks loudly if it changes, and a unit test that names no dialect at all.

## Forbidden rows: 15, every one demonstrated capable of firing

Nine are new; **two pre-existing rows had to change DIALECT**, which is the trap this arm was warned about and is exactly what happened:

- `make_logger --[CALLS]--> stream_writer` (expected) used `to_bare_name`. `stream_writer` is now a class entity, so the CALLS target resolves and its `ToID` is a content hash — the row **failed** until it was rewritten to `to_name` + `to_kind` + `to_file`. It is stronger now: it asserts the edge binds to the right node rather than that a string survived.
- `Comparable --[EXTENDS]--> printer` (forbidden) used `to_bare_name` on the stated ground that `printer` "is a class type and produces no entity at all". This arm makes that false, so the fence would have gone **silently unfireable** in precisely the way its own note warned about. Re-dialected and re-verified.

Four class declarations were added to `containers.ml` purely to give four fences a FROM endpoint (`basic_printer`, `relay_printer`, `remote_relay`, `spawning_printer`) — a fence with no anchor cannot fire.

| fence | mutant | verdict |
|---|---|---|
| `relay_printer --[EXTENDS]--> printer` | drop the `module_path` arm of `pathLeafName` | **DEAD**, 1 hit |
| (unit) qualified parent, class-type and functor-application forms | drop the `extended_module_path` arm | **DEAD**, disjoint sub-tests |
| `remote_relay --[EXTENDS]--> scope:component:class:ocaml:containers.ml:external_base` | drop the `declared[target]` gate | **DEAD**, 1 hit — and independently by unit test |
| `spawning_printer --[EXTENDS]--> basic_printer` | make `inheritEdges` search descendants | **DEAD**, 1 hit |
| `tagged_circle --[IMPLEMENTS]--> circle` | swap the kind, **class arm only** | **DEAD**, 1 hit, rel 8/12 |
| `verbose_printer --[IMPLEMENTS]--> printer` | swap the kind, **class-type arm only** | **DEAD**, 1 hit, rel 11/12 |
| `basic_printer --[EXTENDS]--> printer` | annotation emits EXTENDS | **DEAD**, 1 hit, rel 11/12 |
| `containers.ml --[EXTENDS]--> List` (`open`) | direct emission | **FIRES**, 1 hit |
| `containers.ml --[EXTENDS]--> Comparable` (functor) | direct emission | **FIRES**, 1 hit |
| `make_logger --[EXTENDS]--> stream_writer` (anon object) | direct emission | **FIRES**, 1 hit |
| `Comparable --[EXTENDS]--> printer` (re-dialected) | direct emission | **FIRES**, 1 hit |
| `Extended --[EXTENDS]--> Base` (`include`) | direct emission | **FIRES**, 1 hit |
| `Extended --[IMPLEMENTS]--> Base` (`include`) | direct emission | **FIRES**, 1 hit |

The kind-swap mutants are split per arm on purpose: the two arms build their edges from different node types, and one shared kind literal is an assumption. The six direct emissions each produced a **distinct** edge, scored separately, against a harness-off control that produced zero hits — no fence claims credit for another's mutant. The `make_logger --[EXTENDS]--> stream_writer` fence exists because the FROM side would be a `SCOPE.Operation` extending a class, a shape the graph has no reading for; paired with the expected CALLS row, the two together say which edge the construct is allowed to produce.

## Behavioural mutants and honest non-results

| mutant | verdict |
|---|---|
| delete the whole emit site | **DEAD** — 19/31 entities (61.3%), 5/12 relationships (41.7%) |
| keep entities, drop the edges | **DEAD** — entities unmoved at 31/31, relationships 6/12; exactly the six hierarchy rows missing, so the fixture grades this change and not the extractor in general |
| drop the self-edge guard | **DEAD** at unit level |
| drop the parent dedup | **DEAD** at unit level |
| read only the first `class_binding` (`and` chains) | **DEAD** at unit level |
| EXTENDS line ← body start instead of clause | **DEAD** at unit level |
| IMPLEMENTS line ← binding start instead of annotation | **ALIVE when first probed.** The two coincide on every single-line class header, which is every class in the fixture, so the field was ungraded. A multi-line-header test now separates them; **DEAD** after. |
| search `class_type_path` recursively under a class_binding | **ALIVE when first probed — and it disproved a claim I had just "demonstrated".** I enumerated five annotation sites (self-type `object (self : printer)`, `val`, method return, ascription `(p : printer)`, coercion `(p :> printer)`), found `class_type_path` in none, and wrote that the search depth was therefore equivalent. A reviewer found the sixth: `#printer`, the hash-type form, parses as `hash_type > class_type_path`, and the mutant emits `g IMPLEMENTS printer` from `method m (x : #printer)`. **Five sites chosen by one person is one confirmation, not five** — the new claim rested on the very enumeration that had just been shown incomplete. The depth is now graded and **DEAD**. |
| `firstChildText`'s slice bound | **Unexercised, and labelled as such in the code** rather than described as if it filtered something: `src` is the exact string the tree was parsed from (`blockIndexFor` keys its memo on string EQUALITY, so no collision can cross files) and a named node's span is non-empty. |

Two `#6938` forbidden **entity** rows (`inherit`, `object`) were not re-verified: this producer names entities from `class_name` / `class_type_name` text only and cannot reach either spelling.

## Ratchet, digest and corpus guards

`scripts/quality/run.sh --ratchet --runs 1` over the whole golden set reported **exactly six changed figures, all six naming `ocaml-objects-mini`** — so "no other fixture moved" is a measurement here, not an inference. Re-recorded by hand for that fixture only; **`--update-baseline` was never run**, and the delta is enumerated in `measured_on_note` member by member:

- `entity_extracted_total` **28 → 40**: twelve class entities and nothing else — five in `shapes.ml`, six in `containers.ml`, one in `render.ml`.
- `relationship_extracted_total` **68 → 86**: six hierarchy edges plus twelve module-rollup CONTAINS, one per new entity. Taken from two independent counters the binary prints and it closes exactly: `pass1_rels` 41 → 47 = the six edges; `module-agg contains` 26 → 38 = the twelve.
- It arrives in two separable steps, both measured: the producer alone against untouched fixture source gives 28→36 and 68→81 (eight classes, five EXTENDS, eight CONTAINS); the four added class declarations take it to 40 and 86 (four classes, one IMPLEMENTS, four CONTAINS).

`baseline.json` was written through `ratchet.py`'s own `serialize()` encoding and `canon()` passes.

**Semantic digests do not move**, and that is derived rather than assumed: `writeSpanFixture6118`'s file set was re-read from the test source and is C#, Java, Python, Go, TypeScript and JavaScript with **no `.ml` file in it**. `go test -count=1 ./cmd/grafel/` was run regardless and is green.

**Corpus-size guards re-derived from disk, not from the diff:** `ls -d internal/quality/golden/*/ | wc -l` = **35**, `ls internal/quality/golden/*/expected.json | wc -l` = **35**, matching the `fixtures != 35` constant in `subtype_assertion_6488_test.go`. This arm adds no fixture directory, so no count constant moves.

The subtype ledger in that test gains twelve rows. Every class row asserts its subtype and the subtype is the load-bearing half: this fixture now carries **five** meanings under one `SCOPE.Component` Kind (file, module, type, class, class_type), so a Kind-only row for a class distinguishes almost nothing — and specifically would be satisfied by a producer that emitted `printer` as a class rather than a class type, the exact collapse the forbidden rows fence in the edge direction.

## Verification

- `go test -count=1 ./internal/extractors/ocaml/ ./internal/extractors/ ./internal/quality/... ./cmd/grafel/` — **exit 0** (`./internal/extractors/` included for `file_anchored_rels_guard_test.go`; every hierarchy edge leaves `FromID` empty so assembly anchors it on the class, asserted directly by `TestOCamlHierarchy_FromIDIsEmpty_6370`).
- `scripts/quality/run.sh --ratchet --runs 1` — **exit 0**, 35 gated fixtures held their baseline.
- `gofmt -l internal/ cmd/ scripts/` — clean. Every mutant edit was applied through a helper that **asserts an occurrence count of exactly 1** before writing, and a green baseline was confirmed immediately before each score.

Refs #6370
Refs #6812
Refs #6939
Refs #6947


cajasmota:

Own checkout at `8b35652ec`, own `GRAFEL_HOME`/`GRAFEL_DAEMON_ROOT`, own build. Every mutant applied
through a helper that aborts unless the occurrence count is exactly 1, restored by `cp` from a
shasum-verified backup, green baseline re-confirmed before each score, `-count=1` throughout.

The work is unusually well evidenced and most of the claims survive checking. Four items do not, and
all four are cheap. None is a live production defect: two are guard-coverage gaps and two are prose
that asserts more than the code does.

---

### 1. The bare-name-vs-structural-ref trade — the premise is TRUE, the conclusion is not. A third option exists.

I reproduced all four cells. The premise holds exactly as stated:

| design | gate | `remote_relay --[EXTENDS]--> external_base` | verdict |
|---|---|---|---|
| bare name (as shipped) | intact | 0 hits, 31/31, 12/12 | green |
| bare name | deleted | **1 hit**, `pass1_rels` 47→48 | mutant **DEAD** |
| `scope:component:class:ocaml:<file>:<name>` | intact | 0 hits, 31/31, 12/12 | green — the ref does work end to end |
| structural ref | deleted | **0 hits**, `pass1_rels` 47→48, `go test ./internal/extractors/ocaml/ ./internal/quality/` **exit 0** | mutant **ALIVE** |

So "the ref makes the gate ungradeable" is true *at the fixture level*, and I confirmed the dangling
edge really is emitted (48th `pass1_rels`) while every fence and the whole suite stay green. That is
a genuine hazard and you were right to refuse to ship it unscored.

**But the generalisation — "the only thing that grades the rule" — is false.** The gate is gradeable
under the ref, at the unit level, in three lines, using *your own helper from your own file*:

```go
func TestOCamlHierarchy_UndeclaredParentYieldsNoEdge_6370(t *testing.T) {
	wantEdges(t, "class remote_relay = object\n  inherit external_base\nend\n")
}
```

Scored under all four cells:

- bare name + gate intact → **PASS**
- ref + gate intact → **PASS**
- ref + gate **deleted** → **FAIL** (`hierarchy edges: got ["remote_relay EXTENDS external_base"]`) — the ALIVE
  mutant above is **DEAD**
- (bare + gate deleted is already dead at the fixture)

`hier()` reduces the target past the last `:`, so the assertion is dialect-independent by
construction. This is exactly the pattern you already used for the self-edge guard
(`TestOCamlHierarchy_SelfEdgeSuppressed_6370`) and for module composition
(`TestOCamlHierarchy_ModuleCompositionMintsNoEdge_6370`) — it was applied to two guards and not to
the two guards that are the arm's entire precision story. The qualified-path guard is in the same
position: it has **no** unit test either, only the `relay_printer` fence.

I am **not** asking you to switch to the ref. The bare name is independently defensible and I
checked the downstream hazard the brief flagged: a two-file probe with `class parent` declared in
both files binds the edge to the *owning* file's node, in both alphabetical orders — the locality
tiers hold, and #6369 is not reintroduced. What I am asking is that the design note stop resting on
a false premise, and that the two gates get unit coverage that does not depend on the ToID dialect.
Then the choice becomes a real trade (simplicity + convention vs. precision) rather than a forced
move.

### 2. Half of `pathLeafName`'s guard is ungraded, and the ungraded half is reachable

```
mutant: `c.Type() == "module_path" || c.Type() == "extended_module_path"` → `c.Type() == "module_path"`
```

Fixture: 31/31, 12/12, **0 forbidden hits**. Unit suite: green. **ALIVE.**

And it is production-reachable, not equivalent. Probed against the real grammar:

```ocaml
class type printer = object method p : int end
class type d = object inherit Make(Int).printer end
```

emits `d EXTENDS printer` under the mutant (0 edges unmutated). `Make(Int).printer` parses as
`extended_module_path`, so the class-**type** arm reaches the arm no fixture input exercises — and
it is the identical #6369 wrong-node shape that `relay_printer` fences on the sibling arm. The
compound guard was scored as a unit; each half needs its own input. One line in `containers.ml` plus
one forbidden row, or one unit test, closes it.

### 3. The `class_type_path` equivalence claim is refuted — there is a sixth site

The doc comment says the recursive search is "equivalent here", on five probed annotation sites all
yielding `type_constructor_path`. I re-probed those five (all confirmed) plus seven more, dumping
every `class_type_path` / `type_constructor_path` inside a `class_binding`. Site 07:

```ocaml
class g = object method m (x : #printer) = x end
```

produces **`class_type_path`** at depth 8 inside the binding. The `#` open-object-type form is a
class-type position that appears in an ordinary *method parameter annotation*. With the recursive
mutant applied, this emits `g IMPLEMENTS printer` — a fabricated IMPLEMENTS from a method signature,
which is precisely what the doc comment you deleted warned about and what the replacement text now
denies. Fixture and unit suite are **green** under the mutant, so it is **distinguishable and
ungraded**.

Production is correct — the direct-child read rejects it. What needs changing is the claim (it is
stated as demonstrated, and the demonstration had a hole: none of the five sites used the `#` form)
and the guard's coverage: one unit test on the `#printer` shape kills the mutant and makes the
comment true-as-tested.

### 4. Three stale passages in `expected.json`'s `description`, now false

The fixture description says "HIERARCHY IS NOW GRADED" and then, five sentences later, still says:

- *"OCaml emits no EXTENDS/IMPLEMENTS (#6370 group (a)); this fixture asserts no hierarchy row in EITHER direction for the object/class-type forms"* — it now asserts six.
- *"all six forbidden relationships and both forbidden entities carry a FIRES verdict"* — there are now **15** forbidden relationships.
- *"no `class`/`class type`/`method`/`val` construct produces any entity, so shape/circle/square/printer/stream_writer are invisible"* — false for the class half as of this PR.

This is the artefact a reviewer reads to learn what the fixture asserts. Same defect class the PR
itself is careful about everywhere else.

---

### My mutant table (mine, not yours)

| # | mutant | verdict |
|---|---|---|
| M0 | delete the `classEntities` emit site | **DEAD** — 19/31 (61.3%), 5/12 (41.7%); totals 28/68 |
| M1 | `pathLeafName`: drop the `module_path` arm only | **DEAD** — 1 hit (`relay_printer → printer`) |
| M2 | `pathLeafName`: drop the `extended_module_path` arm only | **ALIVE — distinguishable and ungraded.** `class type d = object inherit Make(Int).printer end` → `d EXTENDS printer`. No fixture input, no unit test |
| M3 | annotation search `childrenOfType` → recursive descendants | **ALIVE — distinguishable and ungraded.** `method m (x : #printer)` → `g IMPLEMENTS printer`. Refutes the "equivalent under the grammar" claim |
| M4 | delete `if !declared[p.target]` (bare-name ToID, as shipped) | **DEAD** — 1 hit |
| M5 | ToID → structural ref **and** delete the gate | **ALIVE** — 0 forbidden hits, `./internal/extractors/ocaml/ ./internal/quality/` exit 0. **DEAD** once the 3-line unit test above exists |
| M6 | `classBindingDecl`: `object_expression` direct-child → descendants | **DEAD** — 1 hit (`spawning_printer → basic_printer`). A *different* site from the `inheritEdges` mutant you scored; the fence covers both |

### The hand-edited `baseline.json` is CORRECT — re-derived from runs, not checked as arithmetic

Every figure measured on my own build, not read off the diff:

- as shipped: `entities=40 relationships=86 pass1_rels=47 module-agg contains=38`, **31/31 · 12/12 · 0 forbidden** → matches all four recorded figures.
- emit site deleted: `entities=28 relationships=68 pass1_rels=41`, **19/31 · 5/12** → reproduces the *old* recorded 28 / 68 / `entity_found` 19 exactly. (`relationship_found` reads 5 rather than the old 6 only because the re-dialected `make_logger → stream_writer` CALLS row needs the class entity; not a discrepancy.)
- producer only, `containers.ml` reverted to `HEAD~1`: `entities=36 relationships=81 pass1_rels=46` → the claimed 28→36 / 68→81 intermediate step, confirmed.
- `scripts/quality/run.sh --ratchet --runs 1`, all 35 fixtures, my own isolated env: **exit 0 — 35 gated fixtures held their baseline.** Nothing outside `ocaml-objects-mini` moved.

### Also verified independently

- **Forbidden-row sample, direct-emission arm (the weaker evidence class):** re-emitted `Comparable --[EXTENDS]--> printer` (the re-dialected one, the row most at risk of silent unfireability) → **1 hit**; re-emitted `make_logger --[EXTENDS]--> stream_writer` → **1 hit**. Both fences fire.
- 35 golden dirs / 35 `expected.json`; `ocaml-objects-mini` is the **only** fixture containing a `.ml` file. `writeSpanFixture6118`'s file set read from the test source: C#/Java/TS/JS/Python/Go, **no `.ml`** — the digest reason is correct, not lucky. `go test -count=1 ./cmd/grafel/ ./internal/extractors/` **exit 0**.
- No YAML rule pack mentions ocaml; ocaml absent from `cross/hierarchy`; no regex added to the package; `gofmt -l` clean; no orphan figure quoted as evidence anywhere.
- **New vocabulary is minimal and downstream-safe.** 12 entities, all 12 subtype-asserted. `class` matches 18 existing sites; `class_type` is the repo's first, so I checked the two subtype consumers that could care — `internal/mcp/mro.go` (`isClassEntity`/definition check) and `internal/graph/coverage.go` (`nonTestableKinds`) — both key on `Kind`/`SCOPE.Component`, so nothing regresses. The `class_type`-not-`interface` reasoning holds.
- **#6947's premise holds.** `depth_walker_inherit_accuracy_6812_test.go` exists and is env-gated on `GRAFEL_OCAML_CORPUS` as described, so the proposed measurement is one classification pass over an instrument that already runs. One nit for the issue body: `ocaml/ocaml` is a *compiler*, and a single repo — whatever fraction comes back is corpus-relative and should be reported as "in this corpus", not as "the population".

### What I'd need to approve

1. Add the resolution-gate and qualified-path unit tests (3–6 lines), and correct the `hierarchy.go` note so it says the gate is ungradeable *at the fixture level under a ref*, not that nothing else can grade it. Keep the bare name if you still prefer it — just stop justifying it with a premise that a unit test disproves.
2. Grade the `extended_module_path` arm with its own input (M2).
3. Grade the `#printer` shape and fix the equivalence claim (M3).
4. Fix the three stale `description` passages.

cajasmota:
## Review round 1 addressed — `d8c39ecee`. One item resolved AGAINST the reviewer's stated preference, with the measurement.

### 1. Unit coverage added; justification rewritten — and the reviewer's replacement premise measured FALSE

Item 1 is two halves and they came apart.

**(a) Unit coverage — done, and the reviewer was right that my own helper was the instrument sitting unused.** Three new tests, each with a positive control holding everything but the guard constant:

- `TestOCamlHierarchy_UndeclaredParentYieldsNoEdge_6370` — the resolution gate.
- `TestOCamlHierarchy_QualifiedParentYieldsNoEdge_6370` — the qualified-path guard, three sub-tests (see item 2).
- `TestOCamlHierarchy_AnnotationIsTheBindingsOwnDirectChild_6370` — the annotation search depth (see item 3).

All three grade the guard, not the spelling: `hier()` reduces a target past the last `:`, so they hold under either ToID dialect. That property is what made the next half decidable.

**(b) The justification is rewritten, but not to the sentence you asked for.** You asked for "the bare name is sufficient and the fixture cannot name a structural ref". I could not write it, because the first clause is false.

Your reason for keeping the bare name was that its downstream hazard is not realized — "a two-file same-name-class probe binds to the owning file in both file orders". I reproduced that probe before relaying it: two `.ml` files each declaring `class shape`, each with a local subclass, graded in both file orders.

| ToID | bound to own file | cross-file mis-binds |
|---|---|---|
| bare name | **0 / 2** | 0 |
| structural ref | **2 / 2** | 0 |

The bare name never binds to the *wrong* file — the ambiguity leaves the edge unresolved instead, so #6369's wrong-node failure does not occur either way, and your "hazard not realized" reading is correct as far as it goes. But it **silently loses the edge**, which is a recall defect the ref does not have, and same-named classes across files are ordinary in OCaml where a module's classes are named for their role (`t`, `printer`, `writer`).

With gradeability now supplied at unit level, the only argument that had ever favoured the bare name was gone, and the measurement favours the ref. **So this PR switches to the structural ref.** I am flagging the disagreement rather than burying it: it is a one-line revert plus the fence re-spelling below, and you can overrule with the numbers in front of you.

The fixture fence survives the switch rather than being dropped. `remote_relay --[EXTENDS]--> scope:component:class:ocaml:containers.ml:external_base` **fires** under the gate-deletion mutant — the ref spelled out is the only string such an edge can carry, since a target with no entity never resolves. So the gate now has two independent instruments: a fixture row that names the dialect and breaks loudly if it ever changes, and a unit test that names no dialect at all.

The package comment now carries both corrections — "ungradeable" was a conclusion that did not follow from a true premise, and "the bare name is sufficient" was measured false — because each mistake is easy to repeat.

### 2. `extended_module_path` arm — confirmed ALIVE, now graded separately

Reproduced exactly: `class type d = object inherit Make(Int).printer end` emits `d EXTENDS printer` with only that arm deleted. The two arms are reached by **different constructs** — a class's `class_path` qualifies with `module_path`, a class type's `class_type_path` with `extended_module_path` — so one test naming only the first can never reach the second. The new sub-tests are class-form/module-qualifier, class-type-form/module-qualifier, and class-type-form/**functor-application** qualifier, every one declaring a local `printer` so a dropped guard produces a wrong-node edge rather than merely an extra one. Deleting either arm now fails on a disjoint set of sub-tests.

### 3. Recursive annotation search — confirmed ALIVE; the equivalence claim was mine and it was wrong

Reproduced: `class g = object method m (x : #printer) = x end` parses as `hash_type > class_type_path` and the recursive mutant emits `g IMPLEMENTS printer`.

The failure mode is the one worth keeping. I disproved my *original* comment by enumerating five annotation sites, then wrote a *new* claim resting on that same enumeration — **five sites chosen by one person is one confirmation, not five**, and the second claim inherited every blind spot of the first. The comment now says the direct-child read is load-bearing, records both wrong versions and names the sixth site, and the depth is graded.

### 4. Fixture description — three false passages corrected

- "asserts no hierarchy row in EITHER direction" → rewritten in the past tense with what the arm added.
- "all six forbidden relationships" → **fifteen**, six from #6812 and nine from this arm.
- "no `class`/`class type`/`method`/`val` construct produces any entity" → `method` and `val` still produce none; `class` and `class type` produce twelve, asserted above it.

### 5. #6947 nit — filed as a comment on the issue

Its single-corpus measurement is corpus-relative (ocaml/ocaml at `70165ff23e` is one unusually module-dense tree), and its "one deleted `if` away" premise no longer holds now that ToID is a ref — relaxing the gate is two changes, and the second is the harder one.

### Re-verification (everything re-run after the dialect switch, nothing carried over)

All **15** forbidden rows re-scored under the ref — 9 by code mutants, 6 by distinct direct emissions against a harness-off control at 0 hits:

| mutant | verdict |
|---|---|
| drop `module_path` arm | DEAD — `relay_printer EXTENDS printer` |
| drop `extended_module_path` arm | DEAD — unit, disjoint sub-tests **(new)** |
| drop the resolution gate | DEAD — fixture hit **and** unit **(new)** |
| recursive body walk | DEAD — `spawning_printer EXTENDS basic_printer` |
| recursive annotation search | DEAD — unit **(new)** |
| swap kind, class arm only | DEAD — 1 hit, rel 8/12 |
| swap kind, class-type arm only | DEAD — 1 hit, rel 11/12 |
| annotation emits EXTENDS | DEAD — 1 hit, rel 11/12 |
| delete whole emit site | DEAD — 19/31 ent, 5/12 rel |
| entities kept, edges dropped | DEAD — 31/31 ent, 6/12 rel |
| self-edge / dedup / `and`-chain / both clause-line fields | DEAD — unit |
| six direct emissions (open, functor, anon object, cross-construct, include×2) | all FIRE, 1 distinct hit each; control 0 |

- `go test -count=1 ./internal/extractors/ocaml/ ./internal/extractors/ ./internal/quality/... ./cmd/grafel/` — **exit 0**
- `scripts/quality/run.sh --ratchet --runs 1` — **exit 0**, 35 gated fixtures held their baseline
- `gofmt -l internal/ cmd/ scripts/` — clean
- Extracted totals unchanged at **40 / 86**, so the hand-recorded baseline still holds without a re-record; the ToID spelling changes what an edge points at, not how many exist.

`ci:full` survived the force-push and is still on the PR — checked, not assumed. Not merging.

cajasmota:
**Coordinator: I was wrong, and the author was right to overrule me. Recording it here because I said the opposite in writing.**

I relayed the review's finding as *"the bare-name downstream hazard is not realized — a two-file same-name-class probe binds to the owning file in both file orders — so the bare name stands on its own merits"*, and told the author not to switch to the structural ref.

The author measured it before acting on it:

| ToID | bound to own file | cross-file mis-binds |
|---|---|---|
| bare name | **0 of 2** | 0 |
| structural ref | **2 of 2** | 0 |

**"No mis-binding" is not "sufficient".** Ambiguity does not point the edge at the wrong file — it drops the edge entirely. Both the review and I checked only the direction that would have been *loudly* wrong, and neither of us checked whether the edge survived at all. A silent loss of the thing the arm exists to produce is worse than the failure we did test for, and it was one probe away.

So the design decision inverts, and for the right reason. The original justification — "the more precise design is ungradeable" — was false: a three-line unit test using the arm's own `wantEdges` helper grades it. Once gradeability was supplied at unit level, the **only** argument that had ever favoured the bare name was gone, and the measurement above shows the bare name actively loses edges. The PR now ships the structural ref.

The fence survives the switch rather than being dropped, which is the part I would have got wrong if I had insisted: `remote_relay --[EXTENDS]--> scope:component:class:ocaml:containers.ml:external_base` fires under the gate mutant, so the gate has **two independent instruments** — one that names the dialect and will break loudly if it ever changes, one that names none. That is strictly better than either design alone, and it exists because the author treated a coordinator instruction as a claim to be checked rather than an order to be followed.

**The general rule, since this is the second time this cycle it has bitten:** a relayed finding is a claim, not a fact, and it inherits every gap in the check that produced it. I asked an agent to verify a reviewer's advice before relaying it; I then relayed a reviewer's advice without verifying it myself. The author's habit — measure, then act, then say which — is the one that held.

Also settled in this round, both by reproduction:

- **The `extended_module_path` arm was ungraded** because the two arms are reached by *different constructs* (`class_path` → `module_path`, `class_type_path` → `extended_module_path`), so no class-form test can reach the second. `inherit Make(Int).printer` emitting `d EXTENDS printer` is the functor-application version of #6369's wrong-node hazard. Three sub-tests now; deleting either arm fails a disjoint set.
- **The annotation equivalence claim was wrong**, and its shape is worth keeping: the author disproved their original doc comment by enumerating five annotation sites, then built a *new* claim on that same enumeration. `method m (x : #printer)` parses `hash_type > class_type_path` and is the sixth. Five sites chosen by one person is one confirmation, not five. Both wrong versions and the sixth site are now recorded in the comment.

Extracted totals are unchanged at 40/86 across the dialect switch — the spelling changes what an edge points at, not how many exist — so the hand-recorded baseline still holds without a re-record. I will run my own mutant on this head before merging.

cajasmota:
Coordinator mutant on the final head `d8c39ecee`. One mutant, **DEAD** — and it establishes a structural property of this design that is worth stating, because it is why the gate is hard to widen by accident.

## MH-1 — widen the gate's DICTIONARY, not the gate

Every mutant so far — yours, and the review's M1–M6 — attacked the gate itself (`declared[target]`), the path arms, the annotation depth, or the emit site. Nobody attacked the gate's **input**: the set of names that count as a declared class.

So I taught `collectClasses` that a `type_definition` is a class declaration. The gate then resolves `inherit some_record_type` and emits an `EXTENDS` to a plain record type — an over-fire the gate itself cannot see, because from the gate's point of view the target *is* declared.

Anchor count 1 pre-edit, 1 post-edit for the new arm, `go vet` clean, baseline green first (`ok 0.636s`), `-count=1`, restored by `cp` from a shasum-verified backup (`f9be67f43…`), `git status --porcelain` empty.

**Verdict: DEAD** at package level (quality gate stays green):

```
--- FAIL: TestOCaml_TypeDeclarations
    type "color": expected subtype=type, got "class"
    type "option": expected subtype=type, got "class"
    type "point":  expected subtype=type, got "class"
```

## Why it died, and why that is a property rather than luck

`declared` is built at `hierarchy.go:390` from the **same `decls` slice that becomes the class entities**. So the gate's dictionary and the entity population are one list: you cannot widen what the gate will resolve without also minting entities, and the entity subtypes are independently pinned by a pre-existing test that has nothing to do with hierarchy.

That is a genuinely good shape. The equivalent design — a separate lookup table assembled just for the gate — would have let this mutant through silently, and no hierarchy forbidden row would have caught it, because every fence here names a *relationship* and the defect is in *what counts as a class*.

**Worth one line in the comment above `declared`** saying so: the dictionary is the entity set on purpose, and that is what makes widening it visible. Right now the property is real but accidental-looking, and the next refactor that "optimises" the gate into its own map would remove the protection without touching a single test.

This is the third instance this cycle of protection living somewhere other than where you would look for it — a name-axis mutant dying on a general over-fire guard (#6943), a memo-key widening dying on an enumeration written for another purpose (#6944), and now a dictionary widening dying on an entity-subtype test. In all three the right fix was the same: name the guard in the comment next to the thing it guards.

## On the round

Your `FIVE SITES CHOSEN BY ONE PERSON IS ONE CONFIRMATION` comment at `:226` is the right way to leave that lesson — attached to the code that taught it, not in a PR body nobody reads twice. Both wrong versions plus the sixth site being recorded is better than a corrected comment that hides its own history.

Merging on green, with `ci:full` re-applied. Refs #6370, #6812, #6947.

PR #6949, opened 2026-09-06
fix(#6927): docker's `^\s{2}(\w[\w-]*):\s*$` and ansible's two `$`-anchored module rules needed BOTH ends of the file at once — and `(?m)` alone would have over-fired 1867 times

cajasmota:
The docker + ansible arm of #6927 — the two files the previous three arms deliberately
deferred, because they hold the four `$`-bearing patterns and two of them were the measured
over-fire hazard. 5 of the 8 remaining patterns; `play_framework.yaml` and
`graphql_schema.yaml` are a separate arm.

## What `$` actually was, and what it becomes

`detector.go` compiles rule patterns with plain `regexp.Compile` at :156/:177, so `^` is start
of TEXT — and `$` is end of TEXT. Go's `$` is stricter than Perl's: it does not even allow a
trailing newline before it. Three of these five patterns carry a `$`, so they needed both ends
at once. **All five measured ZERO matches across 992 real YAML files** (671 from the 59
upstream repos in `archigraph-corpora`, plus 321 from four upstream ansible repos —
`ansible-examples`, `ansible-for-devops`, `geerlingguy/ansible-role-{docker,nginx}` — because
the existing corpus contains **no ansible at all**, which is itself a measurement).

The first question this arm was asked was not "add `(?m)`" but "is each pattern
framework-specific, or has start-of-text been gating it by accident?". Answered per pattern, by
measurement, before anything was widened:

| pattern | today | `(?m)` ungated, OUT of framework | decision |
|---|---|---|---|
| compose service `^\s{2}(\w[\w-]*):\s*$` | 0 | **1867 in 611 / 941 files** + **40 junk Services inside real compose files** | **REWRITTEN**, `$` removed, + `requires_framework` |
| compose `^volumes:\s*\n\s{2}(…):` | 0 | **0 / 941** | `(?m)`, ungated |
| compose `^networks:\s*\n\s{2}(…):` | 0 | **0 / 941** | `(?m)`, ungated |
| ansible FQCN `\s+(\w+\.\w+\.\w+):\s*$` | 0 | **0 / 683** | `(?m)` + `requires_framework` anyway (free; see below) |
| ansible shorthand `^\s+(apt\|…):\s*$` | 0 | **59 in 43 / 683 files** | `(?m)` + `requires_framework` |

`$` per pattern, reasoned explicitly:

* **compose service** — `$` is **deleted**. Under `(?m)` it would become end-of-line, and
  `^\s{2}(\w[\w-]*):\s*$` then reads *"every two-space key alone on its line"*, which is not
  what a service is. `[ \t]*\n` carries the same "nothing else on this line" half without the
  end-of-text/end-of-line ambiguity, and lets the rule additionally require the next line.
* **volumes / networks** — no `$`. Unaffected.
* **ansible FQCN and shorthand** — `$` becomes end-of-line, and that is exactly right: a module
  key with nothing after the colon *is* the block form, arguments in the mapping below. Both
  rules still mean what their names say; they simply stop requiring the module to be the last
  construct in the file. The inline `key=value` form stays invisible — a pre-existing recall
  gap, now pinned from the other side by forbidden rows.

## Why the compose service rule needed a rewrite and not a gate

`(?m)` alone is the change #6927 proposed, and the measurement is why it was not made.
`requires_framework` cannot help with 40 of its false positives, because **the file IS compose**:
a named volume, network or secret is also a two-space key alone on its line (`db-data`,
`db-password`, `pgdata`, `default`). So the rule now says what a service is structurally — the
same shape as #6945's job rule: a two-space key whose block opens with a compose *service-level*
key at a deeper **literal-space** indent.

Measured after: **109 matches in 47 files = 109 of the 111 real services** in the corpus's 50
compose files (98.2%, up from 0), **zero in-compose false positives**, 5 matches in 3 non-compose
files ungated, **0 under the gate**. The 2 missed are the price of the vocabulary narrowing below
and are pinned by a test, not asserted here.

Every character class in it is load-bearing, and the review round below found three of them wrong
in the first version. Two `\s`-related traps, both named in the rule comment and pinned:

* `\s+` for the next-line indent **matches a newline**, so `volumes:\n  db-data:\n\nsecrets:`
  reads the column-0 `secrets:` as the service body. Hence `[ ]{3,}`.
* 7 of the corpus's compose files put a comment between the service key and its first key
  (`db:` then a note about the image), so the rule carries an explicit comment arm.

And the same trap one position to the LEFT, which the review caught: `^\s{2}` under `(?m)` matches
at a blank line and then eats newlines, so a column-0 key preceded by two blank lines satisfied it.
It is now `^[ ]{2}`.

**The in-compose false-positive class is now empty by the Compose Spec's schema, not by a corpus
count.** Per the spec, the only service-level keys an entry under top-level `volumes:`,
`networks:`, `secrets:` or `configs:` may also carry are `labels` and `environment`, so the
next-line vocabulary excludes both. That is a deliberate precision-for-recall trade with a stated
price: a service whose FIRST non-comment key is one of those two is missed — 2 of the corpus's 111
— because the rule inspects only that one key.

## The marker sets, and their blind spots

**docker — one marker, `services:`.** It admits 50 of 51 compose-shaped corpus files and 10 of
941 others (django's and phoenix's CI workflows, which declare service *containers*; three
prometheus helm values files). Those ten are admitted by the gate and then **declined by the
pattern** — which is the point: the two fences are independent, and `ci/test.yml` in the fixture
is that exact shape. Product: **0 of 941**.

One marker rather than a set, deliberately: on this corpus `depends_on:` and `container_name:`
are strictly redundant with `services:`, so shipping them would add markers no input can grade —
the shape #6945 called out. Blind spots, stated rather than implied: `services:` is not
compose-only (a GHA job's service containers, a k8s `spec.template`, a helm values file all
write it — every such corpus file is held by the *pattern*); and a compose **v1** file has no
`services:` key at all, though it also cannot match the rule, whose `^\s{2}` requires an indented
key.

**ansible — seven markers**, each measured at **zero** of the 683 non-ansible files on its own:
`ansible.builtin.`, `become:`, `gather_facts`, `ansible_`, `with_items:`, `.j2`,
`state: present`. Product with the shorthand rule: 59/43 → **0/0**, keeping 128 of its 156
in-framework matches.

The 28 declined in-framework matches are named rather than rounded away, because **15 of them
are the gate working**: they are one file, `ansible-for-devops`'s `kube-flannel-vagrant.yml`, a
Kubernetes manifest that happens to live in an ansible repo, whose `template:` keys are pod
templates. The other 13 in 11 files are real losses with one shape — a tiny
`roles/*/handlers/main.yml` holding `service:` with `state: restarted`. `notify:` and `handlers:`
would recover them but each admits two non-ansible files, so the trade was declined and recorded.

Blind spots: `state: present` is a module-*argument* idiom, not an ansible-only token; and
`ansible_` / `ansible.builtin.` are measured on repos that predate collections, which is why
`ansible.builtin.` — the most obviously correct marker here — is load-bearing for only 6 corpus
files and costs nothing to drop *on this corpus*. It is kept because the corpus is old, not
because the corpus grades it — and deleting it alone is DEAD at both levels anyway, because the
fixture's role `handlers/main.yml` carries it as its only marker. Zero-of-683 is a measurement,
not a proof.

**The FQCN rule is gated even though it measured zero over-fire**, because the dotted-three-segment
shape only looks like a framework signal: a traefik middleware name, a java-style config key and a
k8s annotation map all have it. The gate was checked file by file rather than in aggregate and
costs **literally nothing** — all 9 in-framework files and all 22 matches survive it. A fence that
is free is not worth declining on the strength of one corpus.

## Graded in both directions, on two new golden fixtures

`docker-compose-services-mini` and `ansible-playbook-modules-mini`.

**Recall on the pre-fix binary (built from the base commit, same fixtures): 2/7 entities
(28.6%) each.** The 2 found are exactly the un-anchored controls. After: **7/7 entities, 2/2
relationships, 0 forbidden hits** on both. Docker relationships were **0/2** pre-fix, because
both endpoints of the `depends_on` edge are Services the dead rule never minted.

**Both axes pinned, listed explicitly** (#6943's lesson — a fixture that pins one does not pin
its neighbour):

*docker-compose-services-mini*

| row | axis VARIED | held constant |
|---|---|---|
| `db-data` / Service | the `$`-form revert, AND the spec-legal `labels:` on a named volume | inside compose, gate satisfied |
| `cache-data` / Service | indent **class** (`[ ]{3,}` vs `\s+`) | same file, same vocabulary |
| `db-password` / Service | next-line **vocabulary** | same file, indent correct |
| `deploy` / Service | the **gate** (file declares no service block) | shape matches: its first key is `secrets:` |
| `unit` / Service | **vocabulary**, in a file the gate ADMITS | gate satisfied |
| `postgres` / Service | leading-indent **count**, in a gate-admitted file | vocabulary in-set |
| `extraArgs`, `hostAliases` / Config | column-0 `^` of the two UNGATED rules | no gate involved |

*ansible-playbook-modules-mini*

| row | axis VARIED | held constant |
|---|---|---|
| `copy` / Module | the `$` (inline vs block) | gate satisfied, name in-set |
| `systemd` / Module | the `^` (commented-out module) | gate satisfied, name in-set |
| `lineinfile` / Module | the **name** alternation | gate satisfied, anchors intact |
| `ansible.builtin.debug` / Module | the `$` of the FQCN rule (its only anchor) | gate satisfied |
| `nginx.conf` / Module | the FQCN **name** shape (3 segments, not 2) | gate satisfied |
| `command` / Module | the **gate** (k8s Deployment) | shape would match |
| `storefront.rate.limiter` / Module | the **gate**, FQCN rule | shape would match |

The child-indent count (`[ ]{3,}` vs `[ ]{2,}`), the leading-indent class (`[ ]{2}` vs `\s{2}`),
the head's whitespace class (`[ \t]` vs `\s`), CRLF, and the spec-legal section-entry shapes are
graded at engine level, because a realistic compose file cannot be shaped into any of them.

Every forbidden row was checked to be **fireable** — each one is emitted by at least one scored
mutant below, so none is the #6938 shape (a row the producer can never spell). `command` rather
than `template` for the k8s gate row precisely because `template` is a must-have from the tasks
file and a forbidden row that collides with an expected one is unfireable by construction.

## Companion engine-level tests

`internal/engine/docker_ansible_multiline_anchor_6927_test.go`, for what a golden fixture
structurally cannot see: which RULE produced an entity (both buckets are aliased onto the whole
`yaml` language, so several rule files see every file and one file's output can stand in for
another's), and each import_marker on its own (a realistic fixture file carries three or four at
once, which is how #6945's single-marker hole survived).

**Rules are selected by BEHAVIOUR, never by pattern text**, and
`TestIssue6927_DockerAnsible_SelectorSurvivesABehaviourPreservingRewrite` is the **control that
must PASS**: it expands every `\w` to its exact RE2 set (`[0-9A-Za-z_]`, class-depth aware so a
`\w` inside `[\w-]` does not become a nested bracket) and asserts the same construct is still
read out of the same probe. That control caught a real defect while it was being written — the
first version rewrote `\w` to `[\w]`, which nests inside `[\w-]` and silently changes the
language; it failed, and the failure was the selector doing its job.

## Mutants

**28 mutants, all DEAD at the fast `./internal/engine/` suite.** Scored at both levels; the
fixture column is the golden gate (`grafel quality` on the two new fixtures plus
`github-actions-workflows-mini`, the one the ansible change touches). A green baseline was
verified immediately before scoring, every edit was proven to land by an exact occurrence count,
and the rule files were restored by `cp` from a shasum-checked backup after each one.

| mutant | engine | fixture | what fired |
|---|---|---|---|
| D1 compose service: revert to the `(?m)`-only form **this issue proposed** | DEAD | DEAD | `Service:db-data`, `cache-data`, `db-password`, `unit` |
| D2 `[ ]{3,}` -> `\s+` (indent CLASS; `\s` matches a newline) | DEAD | DEAD | `Service:cache-data` |
| D3 `^\s{2}` -> `^\s+` (indent COUNT) | DEAD | DEAD | `Service:options`, `Service:postgres` |
| D4 next-line vocabulary -> `\w+` | DEAD | DEAD | `Service:db-password` |
| D12 next-line vocabulary -> `[\w-]+` | DEAD | DEAD | `Service:db-password`, `Service:unit` |
| D5 compose service: drop `requires_framework` | DEAD | DEAD | `Service:deploy` |
| D6 compose service: drop the comment arm | DEAD | DEAD | recall 6/7 (`db` lost) |
| D7 volumes rule: drop `^` | DEAD | DEAD | `Config:extraArgs` |
| D8 networks rule: drop `^` | DEAD | DEAD | `Config:hostAliases` |
| D9 volumes rule: drop `(?m)` | DEAD | DEAD | recall 6/7 |
| D10 compose service: drop `(?m)` | DEAD | DEAD | recall 4/7 |
| D11 replace the docker marker with one nothing carries | DEAD | DEAD | recall 4/7 |
| A1 shorthand: drop `\s*$` (the `$` fence) | DEAD | DEAD | `Module:copy` |
| A2 shorthand: drop `^` | DEAD | DEAD | `Module:systemd` |
| A3 shorthand: alternation -> `(\w+)` | DEAD | DEAD | `Module:lineinfile` |
| A4 FQCN: drop `\s*$` (its only anchor) | DEAD | DEAD | `Module:ansible.builtin.debug` |
| A5 FQCN: third dotted segment optional | DEAD | DEAD | `Module:nginx.conf` |
| A6 shorthand: drop `requires_framework` | DEAD | DEAD | `Module:command` |
| A7 FQCN: drop `requires_framework` | DEAD | DEAD | `Module:storefront.rate.limiter` |
| A8 shorthand: drop `(?m)` | DEAD | DEAD | recall 5/7 |
| A9 FQCN: drop `(?m)` | DEAD | DEAD | recall 4/7 |
| delete only `ansible.builtin.` | DEAD | **DEAD** | recall 6/7 |
| delete only `become:` / `gather_facts` / `ansible_` / `with_items:` / `.j2` / `state: present` (6 mutants) | DEAD | **ALIVE** | — |

**Every one of the 16 forbidden rows is demonstrated capable of firing** by at least one mutant
above — the identities were read back out of the mutant graph, not inferred from a hit count, so
none of them is #6938's unfireable shape.

Two results worth stating rather than averaging away:

* **Six of the seven single-marker deletions are ALIVE at the fixture gate and DEAD only at the
  engine suite.** That is #6945's finding reproduced exactly: a realistic task file carries three
  or four markers at once, so no single deletion is observable on one. The engine test carries
  one minimal task file per marker and asserts *exactly one marker is present* before asserting
  the gated Module survives, so a case that grows a second marker fails loudly instead of
  quietly grading the set. The seventh (`ansible.builtin.`) is DEAD at both, because the
  fixture's role `handlers/main.yml` happens to carry it as its only marker.
* **D1 is the mutant that matters most**, and it is the change this issue originally proposed:
  `(?m)` and nothing else. It fires four forbidden rows across two files — the 40-per-corpus
  in-compose over-fire and a workflow job — and it is why this arm rewrote the pattern instead.

## Finding recorded, not absorbed

**A fixture's own comment can satisfy a substring marker.** Three of the four gate-direction
files were first written with headers saying "declares no `services:` key" and "no `become:`, no
`gather_facts`, no `.j2`" — and `frameworkPresent` is `strings.Contains` over whole file content,
comments included, so every one of those files was admitted and **all four gate rows were
silently unfireable**. Caught only by dumping what the detector actually emitted for each file
before writing `expected.json`, not by any assertion. The fixture headers now say why they do not
name the markers, and the engine tests assert the premise ("this case must NOT carry the marker" /
"this case must satisfy the gate") before asserting the consequence, so a case that drifts into
the other class fails loudly instead of passing for the wrong reason.

## Verification

* `./internal/engine/`, `./internal/quality/...`, `./cmd/grafel/` — green, `-count=1`. `go vet`
  clean, `gofmt -l` clean.
* Ratchet `scripts/quality/run.sh --runs 1 --ratchet`: **OK, 37 gated fixtures**. Delta
  enumerated: two new fixture entries, plus **exactly one** existing figure moved —
  `github-actions-workflows-mini`'s `entity_extracted_total` 55 → 57. Identified by diffing the
  graph of the pre-fix and post-fix binaries on that fixture: `Module:apt` and `Module:service`,
  added, nothing removed. They come from `ops/provision.yml`, which is a real ansible playbook
  the GA arm added as a non-workflow control and which carries `become:` and `state: present`;
  the shorthand module rule can finally read it. Recorded by hand rather than with
  `--update-baseline`, so no other fixture could re-record silently (#6898). **No python drift
  appeared, and as in #6945 that is not python recovering** — a hand-edit means the
  downward-drift mechanism was never exercised.
* Corpus-size guards re-derived **from disk**, not from the diff: 37 fixture directories, 37
  `expected.json`, 37 baseline entries; the three guards move 35 → 37.

Refs #6927, #6917, #6902, #6152, #6943, #6945.

---

## Review round (PR #6949 REQUEST-CHANGES)

All four rule-file changes are in the compose service pattern:

```
-(?m)^\s{2}(\w[\w-]*):[ \t]*\n(?:[ \t]*(?:#[^\n]*)?\n)*[ ]{3,}(?:…|environment|…|labels|…):
+(?m)^[ ]{2}(\w[\w-]*):[ \t]*\r?\n(?:[ \t]*(?:#[^\r\n]*)?\r?\n)*[ ]{3,}(?:…):
```

**F1 CRLF — fixed.** `[ \t]*\n` cannot match `\r\n`; the `\s*$` it replaced could and the
sibling `^volumes:\s*\n` rules still can, so the rewrite was the only compose rule with the
blindness. Now `\r?\n` in both the head and the comment/blank arm. Graded by
`TestIssue6927_Compose_ServiceRuleReadsCRLF`, which builds the CRLF input in Go source rather than
checking in a CRLF file — a checked-in one is at the mercy of autocrlf and would stop grading
silently. Both halves are separately scored (N1, N2).

**F2 "zero in-compose false positives" — the claim is withdrawn and the defect fixed.** The
reviewer is right that it was a corpus measurement being read as a property of the format. Rather
than restate it as "none on this corpus", the vocabulary now excludes `labels` and `environment`,
which per the Compose Spec are the **only** service-level keys a top-level
`volumes`/`networks`/`secrets`/`configs` entry may also carry — so the class is empty by the
schema. Graded four ways at engine level
(`TestIssue6927_Compose_TopLevelSectionEntriesAreNotServices`, one case per section) and at
fixture level: `db-data` now carries a spec-legal `labels:` block, and putting `labels` back in
the vocabulary fires that forbidden row (N3, DEAD at both levels). The price is stated, measured
(111 → 109 of 111 corpus services) and pinned by
`TestIssue6927_Compose_EnvironmentFirstServiceIsMissed`, which asserts the current imperfect
behaviour so a later improvement goes red rather than drifting.

**F3 `^\s{2}` — fixed** to `^[ ]{2}`, graded by
`TestIssue6927_Compose_ServiceRuleIndentCountsAreExact`.

**M1 / M2 / M3 — all three closed, and M3 was not equivalent.**

| review mutant | was | now |
|---|---|---|
| M1 `[ ]{3,}` → `[ ]{2,}` (child-indent COUNT) | ALIVE / ALIVE | **DEAD** at engine |
| M2 `^[ ]{2}` → `^\s{2}` (leading-indent CLASS) | ALIVE / ALIVE | **DEAD** at engine — and the shipped form is now the tight one |
| M3 `[ \t]*` → `\s*` (head whitespace CLASS) | ALIVE / ALIVE | **DEAD** at engine |

M3 was called "equivalent under the suite"; it is not equivalent at all. `\s` is `[\t\n\f\r ]`
and therefore crosses line boundaries, so `\s*` silently turns "nothing else on **this** line"
into "nothing but whitespace for some number of lines" — duplicating the blank/comment arm and
then over-reaching it, because that arm deliberately skips only blank and comment lines. The
distinguishing input was found by **enumerating** 39690 generated key/separator/body combinations
rather than by picking one: 7812 separate the two spellings, and the minimal case is a line
containing a lone form feed. `TestIssue6927_Compose_HeadWhitespaceCannotCrossALine` grades it with
two positive controls (a blank line and a comment line, both of which the arm *does* skip) so a
failure cannot be "the arm broke".

**A forbidden row was un-fired by the F2 fix, and re-scoring caught it.** Removing `environment`
from the vocabulary made `deploy` — the gate row — unreachable, because
`ci/release.yml`'s job opened with `environment:`. That is exactly the #6938 shape, produced by a
fix. Its first key is now `secrets:`, which is both real reusable-workflow syntax and a compose
service-level key, so the rule's shape matches the job and only the gate declines it; D5 fires
`Service:deploy` again at both levels. **All 12 docker mutants were re-scored against the new
pattern text rather than carried over** — the ansible half is untouched and its scores stand.

**The comment-satisfies-marker scan is now a test**, in this PR:
`internal/quality/comment_only_marker_6949_test.go`. It loads every rule set's `import_markers`,
splits each golden fixture source into code and comment lines by extension, and fails on any
marker whose occurrences are *all* in comments. Reproduces the reviewer's result exactly — one
hit, the cross-language lua `stub` in a C# file, recorded in an allow-list **with its reason**
rather than filtered out by a rule, so a same-language instance still fails. Two guards on the
guard: it fails if fewer than 200 markers or 100 fixture files are seen, and it fails on any
source file whose extension the comment map does not know (that guard fired immediately on
`Gemfile` and `config.ru`). Verified fireable by appending `# it has no services: key` to
`ci/release.yml` — it reported exactly the incident — then restored and re-verified by shasum.
Its limits are in the doc comment: comments only (a string literal naming a marker does the same
thing and is invisible to it), no block comments, and a hand-written extension map.

Re-verified after all of the above: `./internal/engine/`, `./internal/quality/...`,
`./cmd/grafel/` green with `-count=1`, `go vet` and `gofmt -l` clean, ratchet **OK at 37 gated
fixtures** with **no baseline change** (both new fixtures still 7/7, 2/2, 0 forbidden, totals
48/68 and 34/32; the GitHub-Actions 55 → 57 already recorded).

---

## MI-1 round: the ansible shorthand alternation's MEMBERSHIP

Fixed in `b29f9eac5`. The coordinator's mutant was right and the gap was real: three axes of that
rule were graded — the `^`, the `$`, the framework gate — plus its recall direction (a module
*outside* the list, `lineinfile`). Nobody attacked what the list **contains**, which is the only
axis where the rule gets wider without any regex construct changing.

`when:` with a list of conditions under it is ordinary ansible and bare-colon-legal, so it has
exactly the shape the rule matches — and it already has its own `when:\s+(.+)` → Config rule
further down the same file. Same for `tags:`, `loop:`, `with_items:`, `vars:`, `args:`,
`environment:`, `block:`/`rescue:`/`always:`. **No corpus frequency is claimed for any of them**;
reachability is the property, and a count here would be the corpus-measurement-as-property trap
running in the permissive direction.

* **Three fixture forbidden rows** — `when`, `tags`, `loop` — one per syntactic family
  (conditional / metadata / loop) so they cannot all rest on one accident, each carried by a bare
  colon block in a task file that already satisfies the gate, so they are held by the alternation
  and not by the gate or the anchors.
* **An exact-set assertion** over the minted Module names at engine level, which membership cannot
  do: it also catches a **same-count substitution** (`debug` → `when`) that no growth ceiling would
  see.
* **A ten-case behavioural table**, one keyword per case with a real module as positive control in
  the same file. Scalar-only keywords (`register`, `become`, `until`, `delegate_to`) are
  deliberately excluded — a bare one is legal YAML but not legal ansible, so a case built on one
  would argue rather than demonstrate.

| mutant | engine | fixture | fired |
|---|---|---|---|
| insert `when` / `tags` / `loop` (3) | DEAD | **DEAD** | `Module:when` / `tags` / `loop` |
| same-count substitution `debug` → `when` | DEAD | **DEAD** | `Module:when` |
| insert `vars` / `args` / `environment` / `block` (4) | DEAD | ALIVE | — |
| alternation → generic `(\w+)` (re-scored) | DEAD | DEAD | 4 rows, was 1 |

The four ALIVE-at-fixture rows are the expected shape: the fixture carries a bare colon for three
keywords, not for ten.

Baseline: `ansible-playbook-modules-mini` `entity_extracted_total` 34 → 35 and
`relationship_extracted_total` 32 → 33, recorded by hand. The one new entity is **identified, not
assumed**: `Config:- ansible_os_family == "Debian"`, minted by this file's own pre-existing
`when:` rule from the bare `when:` block the fixture gained. Recall, forbidden hits and every
other fixture unchanged; ratchet **OK at 37 gated fixtures**.

### DIFF
--- a/internal/engine/docker_ansible_multiline_anchor_6927_test.go
+++ b/internal/engine/docker_ansible_multiline_anchor_6927_test.go
@@ -1,0 +1,795 @@
+package engine
+
+import (
+	"context"
+	"regexp"
+	"sort"
+	"strings"
+	"testing"
+
+	"github.com/cajasmota/grafel/internal/extractor"
+)
+
+// #6927 — the docker_compose + ansible_core arm, i.e. the two files the sweep
+// flagged as the `$`-side over-fire HAZARD and the previous three arms
+// deliberately deferred.
+//
+// detector.go compiles rule patterns with plain regexp.Compile at :156 and
+// :177, so `^` means start of TEXT — and, less obviously, `$` means end of
+// TEXT. Go's `$` is stricter than Perl's: it does not even allow a trailing
+// newline before it. Five patterns were affected:
+//
+//	docker_compose.yaml  ^\s{2}(\w[\w-]*):\s*$                  -> Service
+//	docker_compose.yaml  ^volumes:\s*\n\s{2}(\w[\w-]*):         -> Config
+//	docker_compose.yaml  ^networks:\s*\n\s{2}(\w[\w-]*):        -> Config
+//	ansible_core.yaml    \s+(\w+\.\w+\.\w+):\s*$                -> Module
+//	ansible_core.yaml    ^\s+(apt|yum|…|import_role):\s*$       -> Module
+//
+// The three `$`-bearing ones needed start-of-text and end-of-text at the same
+// time (or, for the FQCN rule which has no `^`, needed the module key to be the
+// last construct in the file), so they were near-dead: ZERO matches across 992
+// real YAML files.
+//
+// `(?m)` ALONE is what this issue predicted would over-fire, and the
+// measurement is why the service rule was REWRITTEN rather than merely
+// widened. Under `(?m)`, `^\s{2}(\w[\w-]*):\s*$` reads "every two-space key
+// alone on its line": 1867 matches in 611 of the 941 non-compose corpus files,
+// AND 40 junk Services inside real compose files, where no framework gate can
+// help.
+//
+// What is graded HERE rather than in the two golden fixtures
+// (internal/quality/golden/{docker-compose-services-mini,
+// ansible-playbook-modules-mini}), and why this file exists alongside them:
+//
+//  1. WHICH RULE produced an entity. The docker and ansible buckets are both
+//     aliased onto the whole `yaml` language (dormantBucketAliases), so at
+//     graph level several rule files see every YAML file and one file's output
+//     can stand in for another's. Every case below runs against a detector
+//     holding ONE rule set.
+//  2. Each import_marker on its own. A realistic fixture file carries several
+//     markers at once, so deleting one is invisible there — the failure #6945
+//     had to be told about by a reviewer.
+//  3. Rule SELECTION by behaviour rather than by pattern text. #6945 shipped a
+//     test that selected its rule with strings.Contains on the pattern and so
+//     went red on a behaviour-preserving rewrite; its apparent kill of a real
+//     anchor weakening was selector loss, not grading. Everything here selects
+//     by what a pattern DOES, and TestIssue6927_DockerAnsible_SelectorSurvivesA
+//     BehaviourPreservingRewrite is the control that must PASS.
+//
+// Graded in BOTH directions (#6902): a recall assertion is structurally blind
+// to over-firing.
+
+// rules6927 returns the loaded rule map reduced to a single framework by name.
+// The rules are the REAL embedded YAML — the defect IS the shipped pattern
+// text, so a hand-written copy would leave the shipped one unobserved — but
+// every sibling rule set is dropped, so nothing another rule file mints can be
+// mistaken for evidence about this one.
+func rules6927(t *testing.T, frameworkName, wantBucket string) map[string][]FrameworkRule {
+	t.Helper()
+	all, err := LoadAllRules()
+	if err != nil {
+		t.Fatalf("LoadAllRules: %v", err)
+	}
+	out := map[string][]FrameworkRule{}
+	for lang, frs := range all {
+		for _, fr := range frs {
+			if fr.Frameworks.Name == frameworkName {
+				out[lang] = append(out[lang], fr)
+			}
+		}
+	}
+	// The loader keys rule sets by their BUCKET directory (`docker`,
+	// `ansible`); compile() aliases the bucket onto the `yaml` language via
+	// dormantBucketAliases, which is why these tests hand Detect a Language of
+	// "yaml" while the map key stays the bucket name. If that stops being true
+	// the tests measure nothing, so it is asserted rather than assumed.
+	if len(out[wantBucket]) != 1 {
+		t.Fatalf("expected exactly one %q rule set in the %q bucket, got %d",
+			frameworkName, wantBucket, len(out[wantBucket]))
+	}
+	if targets := dormantBucketAliases[wantBucket]; !containsStr6927(targets, "yaml") {
+		t.Fatalf("bucket %q is no longer aliased onto yaml (%v); these tests hand Detect "+
+			"Language=yaml and would grade nothing", wantBucket, targets)
+	}
+	return out
+}
+
+func containsStr6927(hay []string, needle string) bool {
+	for _, h := range hay {
+		if h == needle {
+			return true
+		}
+	}
+	return false
+}
+
+func detect6927(t *testing.T, rules map[string][]FrameworkRule, path, src string) *DetectResult {
+	t.Helper()
+	res, err := New(rules).Detect(context.Background(), extractor.FileInput{
+		Path:     path,
+		Content:  []byte(src),
+		Language: "yaml",
+	})
+	if err != nil {
+		t.Fatalf("Detect: %v", err)
+	}
+	return res
+}
+
+func has6927(res *DetectResult, kind, name string) bool {
+	for _, e := range res.Entities {
+		if string(e.Kind) == kind && e.Name == name {
+			return true
+		}
+	}
+	return false
+}
+
+func names6927(res *DetectResult, kind string) []string {
+	var out []string
+	for _, e := range res.Entities {
+		if string(e.Kind) == kind {
+			out = append(out, e.Name)
+		}
+	}
+	sort.Strings(out)
+	return out
+}
+
+func composeRules6927(t *testing.T) map[string][]FrameworkRule {
+	t.Helper()
+	return rules6927(t, "Docker Compose", "docker")
+}
+
+func ansibleRules6927(t *testing.T) map[string][]FrameworkRule {
+	t.Helper()
+	return rules6927(t, "Ansible Core", "ansible")
+}
+
+// composeFixtureSrc is a minimal but complete compose file. It carries the
+// marker, three services and a top-level volumes/networks pair, and is reused
+// wherever a case needs "a real compose file" as its positive control.
+const composeFixtureSrc6927 = `services:
+  api:
+    image: ghcr.io/example/api:1
+    depends_on:
+      - db
+  db:
+    # a comment between the key and its first argument
+    image: postgres:16
+
+volumes:
+  db-data:
+  cache-data:
+
+secrets:
+  db-password:
+    file: ./secrets/db.txt
+
+networks:
+  backend:
+    driver: bridge
+`
+
+// ansibleTasksSrc6927 is a minimal role tasks file in block form, carrying
+// several markers on purpose — it is the "realistic input" control, and the
+// per-marker cases below are the ones that carry exactly one.
+const ansibleTasksSrc6927 = `- name: Install nginx
+  ansible.builtin.apt:
+    name: nginx
+    state: present
+
+- name: Render the config
+  template:
+    src: nginx.conf.j2
+    dest: /etc/nginx/nginx.conf
+`
+
+// TestIssue6927_DockerAnsible_WidenedPatternsCarryMultiline pins the fix itself
+// at the level the defect lives at: the compiled pattern text.
+//
+// Rules are selected by BEHAVIOUR — which compiled pattern reads which
+// construct out of a real file — and the `(?m)` flag is then read off the
+// pattern that was selected. Selecting by `strings.Contains(sp.Pattern, "^\\s{2}")`
+// would make this test a spelling assertion, which is exactly the #6945 defect.
+func TestIssue6927_DockerAnsible_WidenedPatternsCarryMultiline(t *testing.T) {
+	type probe struct {
+		framework, bucket string
+		// what the rule must read out of src, which is how it is identified
+		src, wantCapture   string
+		entityType         string
+		describe           string
+		mustNotCarryDollar bool
+	}
+	probes := []probe{
+		{
+			framework: "Docker Compose", bucket: "docker",
+			src:         "services:\n  api:\n    image: nginx\n",
+			wantCapture: "api", entityType: "Service",
+			describe: "the compose service rule",
+			// The `$` is GONE from this rule on purpose: `(?m)` would have
+			// turned it into end-of-line, i.e. "every two-space key alone on
+			// its line", which is not what a service is. `[ \t]*\n` carries
+			// the same "nothing else on this line" half without the
+			// end-of-text / end-of-line ambiguity.
+			mustNotCarryDollar: true,
+		},
+		{
+			framework: "Docker Compose", bucket: "docker",
+			src:         "volumes:\n  db-data:\n",
+			wantCapture: "db-data", entityType: "Config",
+			describe: "the compose named-volume rule",
+		},
+		{
+			framework: "Docker Compose", bucket: "docker",
+			src:         "networks:\n  backend:\n    driver: bridge\n",
+			wantCapture: "backend", entityType: "Config",
+			describe: "the compose named-network rule",
+		},
+		{
+			framework: "Ansible Core", bucket: "ansible",
+			src:         "- name: x\n  ansible.builtin.apt:\n    name: nginx\n",
+			wantCapture: "ansible.builtin.apt", entityType: "Module",
+			describe: "the ansible FQCN module rule",
+		},
+		{
+			framework: "Ansible Core", bucket: "ansible",
+			src:         "- name: x\n  template:\n    src: a.j2\n",
+			wantCapture: "template", entityType: "Module",
+			describe: "the ansible shorthand module rule",
+		},
+	}
+
+	for _, p := range probes {
+		fr := rules6927(t, p.framework, p.bucket)[p.bucket][0]
+		var selected []SourcePattern
+		for _, sp := range fr.SourcePatterns {
+			if sp.EntityType != p.entityType {
+				continue
+			}
+			re, err := regexp.Compile(sp.Pattern)
+			if err != nil {
+				t.Fatalf("%s: rule file carries an uncompilable pattern %q: %v", p.describe, sp.Pattern, err)
+			}
+			m := re.FindStringSubmatch(p.src)
+			if len(m) > sp.NameGroup && m[sp.NameGroup] == p.wantCapture {
+				selected = append(selected, sp)
+			}
+		}
+		if len(selected) != 1 {
+			t.Errorf("%s: expected exactly ONE pattern to read %q out of the probe, got %d. "+
+				"Either the rule is gone or two rules now overlap, and in both cases the "+
+				"assertions below would be about the wrong pattern", p.describe, p.wantCapture, len(selected))
+			continue
+		}
+		sp := selected[0]
+		if !strings.Contains(sp.Pattern, "(?m)") {
+			t.Errorf("%s lost its `(?m)`: %q. Without it `^` is start of TEXT and `$` is end "+
+				"of TEXT (Go allows no trailing newline before it), which is the whole of "+
+				"#6927 — this rule measured ZERO matches in 992 real YAML files that way",
+				p.describe, sp.Pattern)
+		}
+		if p.mustNotCarryDollar && strings.Contains(sp.Pattern, "$") {
+			t.Errorf("%s grew a `$` back: %q. Under `(?m)` that is end of LINE, i.e. `every "+
+				"two-space key alone on its line`, which mints a Service for every named "+
+				"volume, network and secret in a real compose file — 40 of them across the "+
+				"measured corpus", p.describe, sp.Pattern)
+		}
+	}
+}
+
+// TestIssue6927_DockerAnsible_GatesAreDeclaredAndLoaded pins the OTHER half of
+// the remedy. requires_framework is inert without import_markers — detector.go
+// logs that combination and frameworkPresent then returns false for every file,
+// so a gate that looks present in the YAML can be an off switch — and a gate
+// that quietly disappears is a silent 1867-match over-fire, not a failure.
+func TestIssue6927_DockerAnsible_GatesAreDeclaredAndLoaded(t *testing.T) {
+	for _, tc := range []struct {
+		framework, bucket string
+		wantGated         int
+		wantMarkers       int
+	}{
+		{"Docker Compose", "docker", 1, 1},
+		{"Ansible Core", "ansible", 2, 7},
+	} {
+		fr := rules6927(t, tc.framework, tc.bucket)[tc.bucket][0]
+		markers := fr.Frameworks.Detection.ImportMarkers
+		if len(markers) != tc.wantMarkers {
+			t.Errorf("%s declares %d import_markers, want %d: %v — every marker is graded "+
+				"individually below, so a new one would be ungraded and a lost one is a gate "+
+				"getting weaker in silence", tc.framework, len(markers), tc.wantMarkers, markers)
+		}
+		if len(markers) == 0 {
+			t.Errorf("%s declares no import_markers at all, so every requires_framework "+
+				"pattern in it can NEVER fire", tc.framework)
+		}
+		gated := 0
+		for _, sp := range fr.SourcePatterns {
+			if sp.RequiresFramework {
+				gated++
+			}
+		}
+		if gated != tc.wantGated {
+			t.Errorf("%s has %d requires_framework patterns, want %d", tc.framework, gated, tc.wantGated)
+		}
+	}
+}
+
+// TestIssue6927_Compose_ServiceRuleNeedsBothFences is the core of the docker
+// arm. The rewritten service rule has TWO independent fences — its shape (a
+// two-space key whose block opens with a compose service-level key, at a
+// deeper LITERAL-SPACE indent) and its framework gate — and this table holds
+// each of them on inputs where the other one says nothing.
+//
+// Every "must not mint" row here is a shape measured in the real corpus, not
+// an invention: the workflow `deploy:` / `environment:` pair is what GitHub's
+// own starter-workflows carries in eleven files.
+func TestIssue6927_Compose_ServiceRuleNeedsBothFences(t *testing.T) {
+	rules := composeRules6927(t)
+	marker := rules["docker"][0].Frameworks.Detection.ImportMarkers[0]
+
+	cases := []struct {
+		name       string
+		src        string
+		wantHeld   []string // Service names that must NOT be minted
+		wantMinted []string // Service names that must be
+		fence      string
+	}{
+		{
+			name:       "a real compose file yields exactly its services",
+			src:        composeFixtureSrc6927,
+			wantMinted: []string{"api", "db"},
+			// The volume, network and secret names are the 40-per-corpus
+			// over-fire the un-rewritten `(?m)` form produced.
+			wantHeld: []string{"db-data", "cache-data", "db-password", "backend"},
+			fence:    "shape",
+		},
+		{
+			name: "a workflow job whose first key is compose vocabulary, WITHOUT the marker",
+			src: `name: Release
+jobs:
+  deploy:
+    environment:
+      name: production
+    runs-on: ubuntu-latest
+`,
+			wantHeld: []string{"deploy"},
+			fence:    "gate",
+		},
+		{
+			name: "a workflow that DOES carry the marker is still declined by shape alone",
+			src: `name: Test
+jobs:
+  unit:
+    runs-on: ubuntu-latest
+    services:
+      postgres:
+        image: postgres:16
+`,
+			// `unit:` is declined because `runs-on` is not compose vocabulary;
+			// `postgres:` because it is six-space indented, not two.
+			wantHeld: []string{"unit", "postgres"},
+			fence:    "shape",
+		},
+	}
+
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
+			// Premise, asserted rather than eyeballed: a "gate" case that
+			// quietly grew the marker would grade nothing, and a "shape" case
+			// that lost it would pass for the wrong reason.
+			gated := strings.Contains(tc.src, marker)
+			if tc.fence == "gate" && gated {
+				t.Fatalf("this case is meant to be held by the GATE but carries the marker %q, "+
+					"so it is held by the shape instead", marker)
+			}
+			if tc.fence == "shape" && !gated {
+				t.Fatalf("this case is meant to be held by the SHAPE but does not carry the "+
+					"marker %q, so the gate declines it before the shape is ever consulted", marker)
+			}
+			res := detect6927(t, rules, "deploy/case.yaml", tc.src)
+			for _, want := range tc.wantMinted {
+				if !has6927(res, "Service", want) {
+					t.Errorf("Service %q missing; got %v", want, names6927(res, "Service"))
+				}
+			}
+			for _, bad := range tc.wantHeld {
+				if has6927(res, "Service", bad) {
+					t.Errorf("Service %q was minted and must not be — the %s fence is not holding. "+
+						"Got %v", bad, tc.fence, names6927(res, "Service"))
+				}
+			}
+		})
+	}
+}
+
+// TestIssue6927_Compose_ServiceRuleMintsAnExactSet grades the rule as a SET
+// rather than by membership, because membership is blind to the direction the
+// dangerous mutants move in: widening the indent, the indent CLASS or the
+// vocabulary all ADD names, and every `has(...)` assertion above stays green
+// while they do.
+func TestIssue6927_Compose_ServiceRuleMintsAnExactSet(t *testing.T) {
+	res := detect6927(t, composeRules6927(t), "deploy/compose.yaml", composeFixtureSrc6927)
+	got := names6927(res, "Service")
+	want := []string{"api", "db"}
+	if strings.Join(got, ",") != strings.Join(want, ",") {
+		t.Errorf("compose Services = %v, want exactly %v.\n"+
+			"  extra `db-data`/`cache-data`  -> the `[ ]{3,}` indent CLASS or the `$`-form revert\n"+
+			"  extra `db-password`           -> the next-line VOCABULARY was widened\n"+
+			"  missing `db`                  -> the comment arm `(?:#[^\\n]*)?\\n` was dropped",
+			got, want)
+	}
+}
+
+// TestIssue6927_Compose_TheImportMarkerAdmitsAComposeFileAlone grades the
+// docker marker set marker by marker in the RECALL direction. The set has one
+// entry today, deliberately: on the measured corpus `depends_on:` and
+// `container_name:` are strictly redundant with it, and #6945 records that
+// shipping a marker no input can grade is the shape to avoid. If a later arm
+// adds one, this test fails until a case for it is added.
+func TestIssue6927_Compose_TheImportMarkerAdmitsAComposeFileAlone(t *testing.T) {
+	rules := composeRules6927(t)
+	markers := rules["docker"][0].Frameworks.Detection.ImportMarkers
+
+	cases := []struct{ marker, src, wantName string }{
+		{
+			marker: "services:",
+			src: `services:
+  api:
+    image: nginx
+`,
+			wantName: "api",
+		},
+	}
+	if len(cases) != len(markers) {
+		t.Fatalf("%d cases for %d markers (%v) — a new marker is ungraded", len(cases), len(markers), markers)
+	}
+	for _, tc := range cases {
+		var present []string
+		for _, m := range markers {
+			if strings.Contains(tc.src, m) {
+				present = append(present, m)
+			}
+		}
+		if len(present) != 1 || present[0] != tc.marker {
+			t.Errorf("case %q carries markers %v, want exactly [%q]", tc.marker, present, tc.marker)
+			continue
+		}
+		res := detect6927(t, rules, "deploy/compose.yaml", tc.src)
+		if !has6927(res, "Service", tc.wantName) {
+			t.Errorf("marker %q alone did not admit a compose file: Service %q missing. "+
+				"Deleting this marker turns the service rule off entirely", tc.marker, tc.wantName)
+		}
+	}
+}
+
+// TestIssue6927_Compose_VolumeAndNetworkRulesAreColumnZeroOnly grades the two
+// UNGATED compose rules. They take no requires_framework — measured ZERO
+// matches across 941 non-compose corpus files — so the column-0 `^` is their
+// entire fence, and it has to be graded against WEAKENING and not only against
+// removal (#6945's third review finding).
+func TestIssue6927_Compose_VolumeAndNetworkRulesAreColumnZeroOnly(t *testing.T) {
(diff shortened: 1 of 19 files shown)


cajasmota:

Own checkout (detached at `bd8ee1b94`), own `HOME`/`GRAFEL_HOME`/`GRAFEL_DAEMON_ROOT`, `go vet` clean, every mutant edit proven to land by an exact occurrence count, rules restored by `cp` from a shasum-verified backup, tree confirmed clean and `./internal/engine/` green at the end.

This is the most carefully measured of the four arms and the reasoning is right in the large. I am asking for changes on two **production** defects in the rewritten compose service rule, both demonstrated end-to-end against the **unmutated shipped rule** with the real binary, and both a two-character fix.

---

### Verified independently first

Built from this branch, ran `grafel quality` on the three touched fixtures:

| fixture | ent | rel | forbidden | `entity_extracted_total` |
|---|---|---|---|---|
| `docker-compose-services-mini` | 7/7 | 2/2 | 0 | **48** |
| `ansible-playbook-modules-mini` | 7/7 | 2/2 | 0 | **34** |
| `github-actions-workflows-mini` | 12/12 | 3/3 | 0 | **57** |

Matches `baseline.json` exactly, **including the 55 → 57 move**. The cross-fixture effect is correct behaviour, not a loose gate: `ops/provision.yml` is a real playbook carrying `become:` and `state: present`, and `apt` / `service` are real module keys. Claim 6 confirmed.

Claim 3 (`$` → end-of-line on both ansible rules) holds: both still mean their names, and the inline `key=value` gap is genuinely pinned from the other side by the `copy` and `ansible.builtin.debug` rows. Claim 5 confirmed by reading `roles/storefront/handlers/main.yml` — `ansible.builtin.service:` + `state: restarted` really is a one-marker file (`state: restarted` is not `state: present`), so that DEAD-at-both is genuine.

---

### F1 — BLOCKER: the rewrite makes the rule blind to CRLF, and it is the only compose rule that is

`[ \t]*\n` cannot match `\r\n`. The `\s*$` it replaced could (`\s*` eats the `\r`), and the sibling `^volumes:\s*\n` / `^networks:\s*\n` rules still do — so this is a regression introduced *by the rewrite*, and it is asymmetric across the file.

Converting **this PR's own** `deploy/compose.yaml` to CRLF and changing nothing else:

```
docker-compose-services-mini  ent=4/7 rel=0/2
   MISSING: api/Service, db/Service, worker/Service
```

Every service and both `depends_on` edges gone. A Windows-authored compose file extracts nothing. This repo runs a Windows CI leg. Fix: `[ \t]*\r?\n` in the head and `(?:[ \t]*(?:#[^\n]*)?\r?\n)*` in the comment/blank arm — plus a fixture line or engine case, because M3 below shows that position is currently unobserved in both directions.

### F2 — BLOCKER: "ZERO in-compose false positives" is corpus-relative, and the counter-example is spec-legal

Named volumes and networks may carry `labels:`; secrets and configs may carry `environment:`. All four are in the rule's own next-line vocabulary. Probing the shipped pattern directly:

| input | captured |
|---|---|
| `volumes:` → `db-data:` → `labels:` | `api`, **`db-data`** |
| `networks:` → `backend:` → `labels:` | `api`, **`backend`** |
| `secrets:` → `db-password:` → `environment:` | `api`, **`db-password`** |
| `configs:` → `cfg:` → `environment:` | `api`, **`cfg`** |

End-to-end proof through the binary: adding a spec-legal `labels:` to `db-data` in this PR's own fixture, with the rule untouched, takes `forbidden_entity_hits` **0 → 1** — it fires the fixture's own `db-data` forbidden row.

The corpus's 50 compose files happen to give their volumes `driver:` and their secrets `file:`, neither of which is in the vocabulary. So the measurement is honest and the claim built on it is not: the rewrite **reduces** the 40-per-corpus in-compose over-fire, it does not eliminate it. Either drop the four keys that `services`/`volumes`/`networks`/`secrets`/`configs` share, or say "none on this corpus" in the rule comment and the PR body and add a graded row. Right now the rule comment's "ZERO in-compose false positives — full recall of the raw `(?m)` form with none of its junk Services" is a stronger claim than any input observes.

### F3 — `^\s{2}`: the exact `\s`-matches-a-newline trap this comment names, still live one position to the left

The comment correctly explains why `\s+` is wrong for the next-line indent. The same `\s` is still the *leading* indent. Under `(?m)`, `^` matches at a blank line and `\s{2}` then consumes newlines:

* a **1-space**-indented key preceded by one blank line matches;
* a **column-0** key preceded by two blank lines matches.

`services:\n  a:\n    image: x\n\n\nlabels:\n   image: y\n` captures `a` **and `labels`**. Reach is narrow, but the fix is `^[ ]{2}` and M2 shows nothing would notice either way.

Minor, no action needed: a tab-indented compose file yields zero (YAML-illegal anyway), and an `x-` extension field under `services:` is minted as a Service.

---

### My five mutants — scored at BOTH levels

Engine = full `./internal/engine/ -count=1`. Fixture = `grafel quality` on the three touched fixtures. Green baseline verified immediately before each.

| # | mutant | engine | fixture | verdict |
|---|---|---|---|---|
| M1 | compose service next-line indent `[ ]{3,}` → `[ ]{2,}` | ALIVE | ALIVE | **distinguishable and ungraded.** D2 grades the indent *class*, D3 the service key's *count* — the child key's *count* has no assertion at either level. |
| M2 | compose service `^\s{2}` → `^[ ]{2}` | ALIVE | ALIVE | **equivalent under the suite** — and the shipped, looser spelling is F3. |
| M3 | compose service `[ \t]*\n` → `\s*\n` | ALIVE | ALIVE | **distinguishable and ungraded**, and this is the position that causes F1. `\s*\n` is CRLF-correct; nothing prefers it or the shipped form. |
| M4 | ansible marker `state: present` → `state:` | **DEAD** | ALIVE | Killed cleanly by the per-marker premise: *"case `become:` carries markers [become: state:], want exactly [become:] — it grades the marker SET, not this marker"*. This guard is the best thing in the PR. |
| M5 | docker marker `services:` → `service` | **DEAD** | **DEAD** (feh 0→1, etot 48→49) | Marker specificity graded at both levels. |

Which level caught what: only the engine suite caught anything (M4, M5); the fixture gate caught only M5. Three of five permissive/spelling mutants on the rewritten pattern survive both — the rewrite's *shape* is well graded (D1–D12 are real kills), but its *character classes* are not.

---

### The comment-satisfies-marker finding — confirmed, and generalised

Mechanism confirmed at `internal/engine/detector.go:96-103`: `frameworkPresent` is `strings.Contains` over whole content, and it is evaluated at :402 *before* the per-pattern `requiresFramework` skip at :409. Any occurrence anywhere — comment, docstring, string literal — admits the file. The four fixture headers now avoid the token and the engine tests assert the premise before the consequence, which is the right remedy.

**Repo-wide scan: no other fixture has this problem.** I parsed `import_markers` out of every rule file under `internal/engine/rules` (12 rule sets carry them, 1002 distinct marker strings) and checked every file under `internal/quality/golden/` for a marker whose occurrences are *all* on comment lines. Exactly one hit, and it is benign: `csharp-aspnet-core-mini/src/Services/UserService.cs`'s `// stubbed: real impl would query a DbContext` matches `lua/test_patterns.yaml`'s `stub` marker — cross-language, so `Detect` never offers that file to the lua rule set. No same-language instance exists anywhere in the corpus.

Worth a follow-up issue rather than this PR: **that scan should be a test.** It is ~30 lines, it makes the class un-reintroducible, and it is the only thing that would have caught these four without dumping detector output by hand. The hazard is wider than comments — a string literal or docstring naming a marker does the same thing.

---

### #6948 collision — **NO, they do not collide.** Checked, not assumed.

`git merge-tree --write-tree bd8ee1b94 d8c39ecee` → exit 0, clean tree `7d54c7d4c`, no conflict.

* `baseline.json` — both branch from the same blob `da35e0931`. This PR's hunks are `@@ -13` / `@@ -45` / `@@ -74` (two new fixture blocks and the GA 55→57). #6948's are `@@ -150` (the `ocaml-objects-mini` block) and `@@ -287` (`measured_on_note`). Disjoint with context.
* `subtype_assertion_6488_test.go` — this PR at `@@ -288` (the 35→37 constant); #6948 at `@@ -311` and `@@ -340`. Disjoint.
* Semantically independent as well: #6948 adds **no** fixture directory, so the re-derived 37 stays correct under either merge order. Whoever lands second merges, no rebase and no re-derivation needed.

---

### Other claims

* **Claim 1 (gating).** One docker marker is the right call and M5 shows it is graded at both levels. The seven ansible markers are graded individually and the "exactly one marker" premise (M4) is what makes that real rather than nominal. The 13 `handlers/main.yml` losses **are** worth naming and you named them — but note the fixture's own handlers file carries `ansible.builtin.`, so the lossy real-world shape (bare `service:` + `state: restarted`) exists only in prose. Fine to defer; say so in the issue.
* **Claim 2 (the two ungated zeros).** Corpus-relative and honestly labelled as such. The bucket population really is every YAML in every repo, so `^volumes:` at column 0 with a two-space child is a thin fence — a kustomize patch or a split multi-doc that opens with `volumes:` would hit it. Low enough risk, disclosed, accept.
* **Claim 4 (forbidden rows fireable).** Sampled independently: `db-data` fired under F2's spec-legal edit, and M5 fired a docker row. Reading identities back out of the mutant graph is the right method and it holds up.

Fix F1 and F2 (F3 while you are in the pattern), add a case for whichever of M1–M3 you choose to close, and this is a clear approve.

cajasmota:
All four points addressed in `0a7701fb1`. Both blockers were real; F2 in particular was a claim I built on my own measurement without checking it against the format.

**F1 CRLF — fixed.** `\r?\n` in the head *and* in the comment/blank arm, scored separately (both DEAD at engine). Graded by `TestIssue6927_Compose_ServiceRuleReadsCRLF`, which builds the CRLF input in Go source rather than checking in a CRLF file — a checked-in one is at the mercy of autocrlf and would stop grading silently.

**F2 — claim withdrawn, and the defect fixed rather than documented.** You are right that it was a corpus count read as a property of the format. `labels` and `environment` are, per the Compose Spec, the *only* service-level keys a top-level `volumes`/`networks`/`secrets`/`configs` entry may also carry, so the vocabulary now excludes both and the class is empty by the schema. Your counter-example is now a graded row: `db-data` carries a spec-legal `labels:` block, and putting `labels` back fires it (DEAD at both levels). Price stated and pinned rather than rounded away — 111 → **109 of 111** corpus services, and `TestIssue6927_Compose_EnvironmentFirstServiceIsMissed` asserts the imperfect behaviour so a later improvement goes red instead of drifting.

**F3 — fixed**, `^[ ]{2}`.

**M1 / M2 / M3 all DEAD at engine. M3 is not equivalent.** `\s` is `[\t\n\f\r ]` and crosses line boundaries, so `\s*` turns "nothing else on **this** line" into "nothing but whitespace for some number of lines" — duplicating the blank/comment arm and then over-reaching it, since that arm skips only blank and comment lines. Rather than hand-pick an input I enumerated 39690 key/separator/body combinations: **7812 separate the two spellings**, minimal case a line holding a lone form feed. Graded with two positive controls (a blank line and a comment line, both of which the arm *does* skip) so a failure cannot be "the arm broke".

**A forbidden row was un-fired by the F2 fix, and re-scoring caught it.** Dropping `environment` left `deploy` unreachable — `ci/release.yml`'s job opened with `environment:`. That is the #6938 shape produced *by a fix*. Its first key is now `secrets:` (real reusable-workflow syntax and a compose service-level key), so the shape matches and only the gate declines it; D5 fires `Service:deploy` again at both levels. **All 12 docker mutants were re-scored against the new pattern text** rather than carried over; the ansible half is untouched and its scores stand.

**The scan is a test, in this PR** — `internal/quality/comment_only_marker_6949_test.go`. It reproduces your result exactly: one hit, the cross-language lua `stub` in `UserService.cs`, allow-listed **with its reason** rather than filtered by a rule, so a same-language instance still fails. Two guards on the guard: it fails below 200 markers or 100 fixture files, and it fails on any source extension the comment map does not know — which fired immediately on `Gemfile` and `config.ru`. Verified fireable by re-creating the incident (`# it has no services: key` appended to `ci/release.yml` → reported exactly that), then restored and re-checked by shasum. Its limits are in the doc comment: comments only (a string literal naming a marker is invisible to it), no block comments, hand-written extension map.

Two of your notes deliberately not actioned, recorded instead: an `x-` extension field under `services:` is still minted as a Service, and the `handlers/main.yml` recall shape (`service:` + `state: restarted`) still exists only in prose — both belong in the #6927 issue rather than here.

Re-verified: `./internal/engine/`, `./internal/quality/...`, `./cmd/grafel/` green with `-count=1`, `go vet` + `gofmt -l` clean, ratchet **OK at 37 gated fixtures with no baseline change** (48/68 and 34/32 unchanged; the GHA 55 → 57 already recorded). Not merged.

cajasmota:
Coordinator mutant on the final head `0a7701fb1`. One mutant, **ALIVE at both levels** — the ansible shorthand alternation's *membership* is ungraded.

## MI-1 — widen the alternation by one plausible entry

`^\s+(apt|yum|dnf|…|debug|set_fact|…):\s*$` → the same list with **`when`** inserted.

Every mutant on this arm — your 28, the review's 5, and the re-scored 12 — attacked the **anchor**, the **`$`**, the **indent**, the **gate**, or an individual **marker**. Nobody attacked what the alternation *contains*. That is the third axis, and it is the permissive direction: a keyword added to that list mints a `Module` for every task that uses it.

Anchor count 1 pre-edit, 1 post-edit, `go vet` clean, baseline green first (`ok 47.637s`), `-count=1`, restored by `cp` from a shasum-verified backup (`5c91f26f5…`), `git status --porcelain` empty.

**Verdict: ALIVE — `./internal/engine/` exit 0 AND `./internal/quality/...` exit 0.**

The fixture contains **no bare `when:` at all**, so nothing can distinguish the mutant. This is the same shape the review found in your GitHub Actions sibling — where the `runs-on|uses` alternation died only *incidentally* on the growth ceiling — except here it does not die at all.

## Why `when` specifically, and the limit of what I established

`when:` is a task-level keyword, not a module, and the pattern requires nothing after the colon. A bare `when:` followed by an indented list is ordinary, documented ansible:

```yaml
- name: restart nginx
  service:
    name: nginx
  when:
    - ansible_os_family == "Debian"
    - nginx_enabled
```

So the mutant is production-reachable in the sense that matters: real playbooks contain the line it needs. **I did not measure how frequently**, and I am not claiming a corpus figure — that is exactly the "a zero is corpus-relative" trap running the other way, and one of your own blockers this round was a corpus count read as a property of the format.

**Grade the membership.** The cheapest form is a forbidden row for a task-level keyword that must never become a `Module` — `when`, `register`, `notify`, `become` and `tags` are all bare-colon-legal and all already have their own rules elsewhere in this file, so they are the natural negatives. An exact-set assertion over the minted Module names would be stronger still, and would also catch a same-count substitution, which no ceiling would; that is what the review asked for on the GHA alternation and it is worth the symmetry.

## On the round

The F2 fix is the right kind. Narrowing the vocabulary so the false-positive class is **empty by the Compose Spec** rather than zero-by-corpus-count is a genuinely different claim, and pricing it honestly — 111 → 109 of 111, with `TestIssue6927_Compose_EnvironmentFirstServiceIsMissed` asserting the imperfect behaviour so a later improvement goes red instead of drifting — is better than rounding the loss away.

Two things you caught that the suite did not, both worth carrying:

- **Dropping `environment` un-fired the `deploy` gate row**, because a fixture job opened with `environment:` — the #6938 silently-unfireable shape produced *by a fix*. You only saw it by re-scoring all 12 docker mutants against the new pattern text instead of carrying the old verdicts over.
- **The `cp` restore silently reverted an edit made after the backup**, caught by `git diff HEAD~1` before pushing and by no test. That is the shasum-proves-the-backup-not-HEAD trap; a backup is a claim about a moment, not about the tree.

And the M3 enumeration is the right instinct: 39690 combinations with 7812 separating the two spellings, minimal case a lone form feed, beats any hand-picked input — including the one I would have picked.

Refs #6927, #6946.

cajasmota:
MI-1 fixed in `b29f9eac5`. You were right, and the framing was right too — this is the one axis where the rule gets wider with no regex construct changing, and the four earlier rounds all attacked constructs.

**Three fixture forbidden rows** — `when`, `tags`, `loop` — one per syntactic family (conditional / metadata / loop) so they cannot all rest on one accident, each carried by a bare-colon block in a task file that already satisfies the gate, so the alternation holds them rather than the gate or the anchors. I stopped at the keywords I could give a bare colon to honestly: `register`, `become`, `until` and `delegate_to` are scalar-only, so a bare one is legal YAML but not legal ansible, and a row built on one would argue rather than demonstrate.

**The exact-set assertion is in**, and it earned its keep immediately: the same-count substitution `debug` → `when` is DEAD only because of it. Also a ten-case behavioural table (one keyword per case, a real module as positive control in the same file) so the axis is graded beyond the three the fixture can shape.

| mutant | engine | fixture |
|---|---|---|
| insert `when` / `tags` / `loop` | DEAD | **DEAD**, each firing its own row |
| same-count `debug` → `when` | DEAD | **DEAD** |
| insert `vars` / `args` / `environment` / `block` | DEAD | ALIVE — the fixture carries a bare colon for three keywords, not ten |
| alternation → generic `(\w+)` (re-scored) | DEAD | DEAD, now 4 rows instead of 1 |

I took your "did not measure frequency and am not claiming a corpus figure" literally and carried it into the rule comment — reachability is the property, and a count there would be my own blocker running the other way.

Baseline moved by one entity and one edge on that fixture, **identified rather than assumed**: `Config:- ansible_os_family == "Debian"`, minted by this file's own pre-existing `when:\s+(.+)` rule from the bare `when:` block the fixture gained. Recorded by hand; recall, forbidden hits and every other fixture unchanged.

`./internal/engine/`, `./internal/quality/...`, `./cmd/grafel/` green `-count=1`; `go vet` + `gofmt -l` clean; ratchet **OK at 37 gated fixtures**. Not merged.

cajasmota:
Coordinator re-score on the final head `b29f9eac5`. **MI-1 is DEAD**, by two named assertions with precise messages:

```
--- FAIL: TestIssue6927_Ansible_TaskKeywordsAreNotModules/when
    Module "when" minted from a task KEYWORD — it is in the shorthand rule's
    alternation, which lists MODULES. Got [apt when]
--- FAIL: TestIssue6927_Ansible_ModuleRulesReadBlockFormOnly
```

Anchor applied and verified (1 occurrence), `go vet` clean, `-count=1`, restored by `cp` from a shasum-verified backup (`9d261f507…`), `git status --porcelain` empty.

## An instrument distinction I nearly reported wrong

My run also shows `go test ./internal/quality/... → exit 0` under the mutant, and I was one sentence away from writing that the fixture missed it. **It did not — that command is the wrong instrument.** `./internal/quality/...` runs that package's own unit tests; the golden-fixture assertions run through the ratchet (`scripts/quality/run.sh --ratchet`), which I did not re-run. So I have verified the engine level myself and am **taking your fixture-level result on report**, which is the honest description.

Worth stating because it is this board's "package selection is not suite selection" trap in a new place: a green `go test` over a directory whose *name* matches the thing you care about is not a check of that thing.

## On your fix

**The exact-set assertion is the piece that mattered**, and it earned its keep on the mutant nobody would have thought to write: `debug` → `when` is a **same-count substitution**, so it moves no total and no ceiling — membership assertions and the growth gate are both blind to it, and only an exact set catches it. That is the second time this cycle an exact-set beat membership (the GitHub Actions job rule was the first), and it should be the default for any rule whose specificity lives in a list.

**The restraint is the other half.** Stopping at `when`, `tags` and `loop` — because `register`, `become`, `until` and `delegate_to` are scalar-only, so a bare colon on them is legal YAML but not legal ansible — is the right call. A forbidden row built on an input the language cannot produce argues rather than demonstrates, and it would have read as three more rows of coverage. Choosing one keyword per syntactic family so the three cannot all rest on one accident is the same instinct applied a level up.

Carrying the "no corpus figure" instruction into the rule comment was correct and I want to be explicit about why, since it cuts against the reflex to quantify: reachability is the property under test. A frequency claim there would be your own F2 blocker — a corpus count read as a property of the format — running in the permissive direction instead of the restrictive one.

The baseline delta being **identified rather than assumed** (`Config:- ansible_os_family == "Debian"`, minted by that file's own pre-existing `when:\s+(.+)` rule from the bare `when:` block the fixture gained) is the standard: one entity, named, with the rule that produced it.

Merging on green. Refs #6927, #6946.

Issue #6940, opened 2026-09-06
poll mode never converges on a path git reports and the walker refuses via the extension, ignore or sparse gate — a reindex every interval, forever


cajasmota:
Found reviewing #6935 (#6932 arm A). **Not a regression** — it is the third instance of the defect class that arm A's blocker 1 and MA-1 were the first two, reached through the gates the poller cannot consult. Filed so it is not discovered a fourth time.

## The shape

`watch.ChangePoller.pollRepo` builds its candidate set as *manifest keys* ∪ *git-reported paths that pass the entry-type gate*, then hands it to `diff.Filter`. `diff.isChanged` returns true for anything absent from the manifest.

`walk.WalkRepo`'s file branch applies **four** gates, in this order:

1. `shouldSkipFileByExt` — the indexed-extension filter (#1629)
2. `irregularSkipRule` — the entry-type gate (#6416)
3. `igStack.MatchFile` — the file-level `.gitignore` / `.grafelignore` layer (#6931/#6933, landed in #6933)
4. the sparse-checkout filter (#2181)

`existsOnDisk` → `walk.IndexableEntryType` reproduces **gate 2 only**, and cannot reproduce 3 or 4 at all from a single path (3 needs the walk's inherited ignore stack, built by descending from the root; 4 needs `opts.Sparse`).

So a path that git reports, that clears the entry-type gate, and that one of gates 1/3/4 then refuses, enters the candidate set, is stamped by nothing, and reads as new on **every cycle, forever**.

## Driven

Both on `lane/6932-arm-a-pollmode` rebased onto `b92b633c3`, macOS/APFS, with the real poller and the real `diff` manifest round-trip, running an index pass between every cycle:

**Gate 1 — untracked file with a non-indexed extension**

```
cycle 1: [photo.png]
cycle 2: [photo.png]
cycle 3: [photo.png]
cycle 4: [photo.png]
cycle 5: [photo.png]
manifest keys: [alpha.go beta.go]      <- photo.png is never stamped, and cannot be
```

**Gate 3 — tracked but gitignored** (the case #6933 opened; per #6937 this repo has 2 such files today)

```
.gitignore contains "gen.go"; gen.go is tracked (git add -f) and then edited
manifest keys after index: [.gitignore alpha.go beta.go]   <- gen.go excluded by the ignore layer
cycle 1: [gen.go]
cycle 2: [gen.go]
cycle 3: [gen.go]
cycle 4: [gen.go]
[...]

Issue #6928, opened 2026-09-06
A second process-wide fsnotify.NewWatcher() at worktree.go:674 drains and discards its Errors channel — #6921 overflow recovery does not cover it


cajasmota:
Found by the independent review of PR #6925 (#6921). **Not caused by that PR** — it is the same silent-staleness shape, one watcher instance over, and it makes that PR's scope comment overstated.

## The defect

#6921 established that `internal/daemon/engineplane.go:203` is the only non-test `watch.NewWatcherConfig` call, and that the single `watch.Watcher` it builds holds one `fs *fsnotify.Watcher` shared by every `AddRepo`. That is true, and it is what made "an overflow invalidates everything watched" the correct recovery scope.

But there is a **second, independent `fsnotify.NewWatcher()`** at `internal/daemon/worktree/worktree.go:674`, outside that `Watcher` entirely. **Its Errors channel is drained and discarded.**

inotify queues are **per file descriptor**, so this instance has its own queue and its own overflow condition. #6921's handler is wired to the other instance and cannot see it. An `IN_Q_OVERFLOW` on this fd is dropped exactly the way every overflow was before #6921 — logged nowhere, counted nowhere, surfaced nowhere.

## Why it is P3 and not higher

Two mitigations, both real:

1. **Small subscription.** It holds roughly one descriptor per parent repo, not per directory, so the queue pressure that produces an overflow is far lower than on the main watcher — which subscribes ~976 directories per worktree on this repo.
2. **It self-heals.** `Watcher.Start`'s 60-second reconciliation poll re-derives state independently of events, so a dropped event costs up to a minute of staleness rather than permanent staleness. That is categorically better than the main watcher's pre-#6921 behaviour, where a dropped event was lost forever.

So this is a real gap with a bounded cost, not a repeat of the original defect.

## What a fix has to establish

1. **Whether the two instances should be one.** `manager.go:6-12` already argues against a second `watch.Watcher` ("rather than introducing a second Watcher instance"). This fd predates or sidesteps that reasoning — establish which, and whether the worktree poller genuinely needs its own fd or could share.
2. **If they stay separate, wire the same handler.** #6921's `handleOverflow` and `OverflowStats` are reusable; the recovery for this instance is a worktree re-discovery rather than a repo rescan, so the hand-off differs even though the detection does not.
[...]

PR #6950, opened 2026-09-07
feat(#6932) arm B part 1: project the inotify watch-descriptor cost and say it out loud — the probe, not the auto-switch

cajasmota:
#6932 **arm B, part 1**: the inotify budget becomes observable. This is the probe and the announcement — **not** the auto-switch. `auto` still resolves to fsnotify, `PollingEnabled()` is untouched, and no configuration subscribes anything it did not subscribe before.

## Why the arm is split

An auto-switch needs a threshold, and every number #6932 records is macOS/APFS with **no concurrency measurement**. Guessing a threshold and calling it automatic would be worse than the flag arm A already ships. This half produces the data the threshold needs, on the owner's own containers. Part 2 — the switch rule — changes exactly one function, and this PR does not touch it.

## The projection is compared against reality — one predicate, twice, plus one independent pin

`ProjectInotifyBudget` walks a repo set and reports what subscribing it would cost, without subscribing it. Graded against **fsnotify's own `WatchList()`** — the library's record of what it handed the kernel, the instrument the #6935 review used — after a real subscription of the same tree:

| | dirs |
|---|---|
| dirs on disk in the fixture | 8 |
| probe's projection | **4** |
| `AddRepo`'s own return | **4** |
| fsnotify `WatchList()` after subscribing | **4** |

**These are not three independent confirmations, and the body previously implied they were.** The projection and `AddRepo` are *one predicate (`watchDirSkip`) evaluated twice*, and `WatchList()` is a backend echo of the second evaluation — a real check that `AddRepo`'s counter is not lying about what it handed fsnotify, but not an independent opinion about *which* directories should be watched. Widen the shared predicate and all four numbers move together.

The one leg outside the predicate is the **named pin** at `inotifybudget_6932_test.go:100`: `dirsOnDisk` is a bare `WalkDir` counter and the `8`/`4` are literals. It carries the test, and it is correct — root + `src` + `src/inner` + `pkg` are watched, `node_modules{,/deep}` is pruned by `SkipDirs`, and `generated{,/nested}` by the `generated/` gitignore rule (`generated` is *not* in `SkipDirs`, so the gitignore layer really is the thing firing). Mutant 4 below is the demonstration: it is the only assertion that catches a widened shared predicate.

`Watcher.InotifyBudget()` answers with a **measurement** in fsnotify mode (the subscribed set *is* the demand, no walk needed) and a **projection** in poll mode (nothing is subscribed, so the number worth having is what fsnotify would have asked this host for). The report says which of the two it is. It is cached for 60 s because the projection walk is not free and `/diagnostics` is polled; the staleness is bounded and that bound is tested.

## One definition of "which directories are watched"

`subscribeRepo` and the probe now call the same `watchDirSkip` for layers 1–3 (hard-coded skips, the per-Watcher excludes, then `.gitignore`/`watch.json`). Two walks asking that question two ways is how a projection drifts from the thing it projects. The TCC protected-path guard and the watch-dir cap stay at their call sites — the first also applies to the repo root, the second is stateful — applied in the same order by both walks, and their agreement is measured rather than assumed.

## Honesty, where it was easy to be dishonest

- **The pool is shared and the probe is half-blind to it.** `max_user_watches` is per-UID and host-level, not namespaced: every container running as the same UID draws from one pool and none of them can raise it. The report says, in its own text, that it sees grafel's own demand and **cannot** see what other processes have taken, and that any headroom shown is an **upper bound, not a reservation**. That caveat travels over the wire *with* the summary, because it is a property of how the number was obtained and no client can reconstruct it.
- **Off Linux the answer is "not applicable on `<goos>`", never a limit of 0.** `defaultFDBudgetLimit()` returns 0 to mean *accounting disabled*; reusing that 0 as a ceiling would read as *no headroom at all* on a platform where the question does not arise. The directory count is still reported there, because it is a property of the tree, and the summary never renders an `N of 0` comparison.
- **"WOULD EXCEED" is a claim**, so it is made only when grafel's own demand is *strictly* larger than a ceiling that was actually read. Equal fits. An unreadable ceiling is not evidence of crossing one.
- **`GRAFEL_INOTIFY_MAX_USER_WATCHES`** lets an operator ask the what-if from a machine that is not the container. A report built from it names the env var as its source, so it can never pass itself off as a kernel read.
- **#6940 is not implied away.** The poll-mode note says only what it can support — that zero descriptors are held right now. It makes no claim about the poller's candidate set agreeing with the indexer's file set.

## The announcement

`grafel status` and `/diagnostics`, with the numbers and the caveats rather than a boolean, on its own line and gated on nothing above it — the case worth seeing is precisely the one where the watcher looks healthy. IntelliJ's precedent is exactly this: report hitting the limit rather than quietly degrading. Silent degradation is the class #6921/#6923/#6928 all belong to.

```
  inotify budget: 9760 of 8192 watch descriptors (projected, 9760 dirs across 10 repos) — WOULD EXCEED
    note: the inotify watch pool is per-UID and host-level, NOT per-container: ... This probe sees only
          grafel's own demand — it cannot see what other processes have already taken, so any headroom
          shown is an upper bound, not a reservation.
    note: demand is PROJECTED from a directory walk; nothing was subscribed to produce it.
```

## Review round: the blocker, and the mutants it left alive

The reviewer's blocker was real and is fixed in `46c1e85d8`. In fsnotify mode the demand came from `w.Stats()` — the watches the kernel **granted** — so every failure made the report *cheaper*, and `GRAFEL_WATCH_FD_BUDGET=2` rendered `inotify budget: 0 of 8192 watch descriptors (measured, 0 dirs across 0 repos) — fits` for a repo that was not watched at all. The demand now counts what was **attempted**: a refused repo's tree is projected back in, a directory whose `fsnotify` `Add` failed (Linux `ENOSPC`) increments a counter and is added back, and `Summary()` carries a `NOT WATCHED` clause plus a `DEGRADED` note. The healthy case stays silent. The verdict word is now `fits (grafel's demand alone)`.

Also addressed: the rendered line is asserted as a *string* (the measured/projected distinction was only ever checked on the struct field, so poll mode could print `(measured, …)` for a walk); darwin + `GRAFEL_INOTIFY_MAX_USER_WATCHES` is graded and renders as an explicit what-if instead of a Linux-shaped verdict; the Linux sysctl read is graded on the ubuntu leg (real read end to end, plus the unparseable and absent branches) rather than only asserted; `Headroom()` has a production caller; `proto.go`'s doc-comment hijack of `WatcherOverflows` is fixed.

## Mutants — 12 applied, 12 DEAD

Each applied with an asserted exact occurrence count, scored with `-count=1` against a green baseline taken immediately before, and reverted with a second asserted count.

| # | mutant (all in the permissive direction) | verdict | killed by |
|---|---|---|---|
| 1 | `WouldExceed`: `>` → `>=` | **DEAD** | boundary case — 100 of 100 must *fit* |
| 2 | non-Linux sets `PoolApplies = true` | **DEAD** | `PlatformAnswerIsHonest` (darwin leg) |
| 3 | projection drops the skip layers entirely | **DEAD** | projected 8 vs 4 actually subscribed |
| 4 | shared verdict drops the `.gitignore` layer — **both walks widen together, so every equality check stays satisfied** | **DEAD** | the named fixture pin, the only assertion outside the predicate |
| 5 | the cached report never expires | **DEAD** | `CacheIsBounded` |
| 6 | the caveats are dropped on the wire | **DEAD** | stand-in *and* the real-watcher call-site control |
| 7 | `grafel status` prints the number without the caveats | **DEAD** | `PrintsInotifyBudgetWithItsCaveats` |
| 8 | refused repos are not counted back into the demand | **DEAD** | `RefusedRepoIsInTheDemandAndAnnounced` |
| 9 | backend-rejected dirs (`ENOSPC`) are not counted | **DEAD** | `BackendRejectedDirsAreInTheDemand` |
| 10 | `Summary()` always prints `measured` *(reviewer's M6, was ALIVE)* | **DEAD** | `SummaryNamesHowTheDemandWasObtained` |
| 11 | the what-if ceiling renders as a measured one *(reviewer's M5, was ALIVE)* | **DEAD** | `WhatIfCeilingIsNeverPrintedAsAMeasuredOne` |
| 12 | projection drops the `dirCap` gate *(reviewer's M2, was ALIVE-unreachable)* | **DEAD** | `HonoursTheWatchDirCap`, via `GRAFEL_WATCH_DIR_CAP` |

Mutants 8 and 9 are the blocker's pins: without them the report gets *cheaper* the worse the failure is.

## Verification

`go test -count=1` green on `./internal/daemon/watch/`, `./internal/daemon/`, `./internal/registry/`, `./internal/cli/`, `./internal/dashboard/`, and `./cmd/grafel/` (the graph digest pin). `go vet ./...` clean, `gofmt` clean.

**`ci:full` applied** — this is `GOOS`-conditional behaviour and the macOS leg does not otherwise run tests. Re-apply it after any push.

## Not established

**Every measurement here is macOS/APFS.** The Linux leg now *grades* the sysctl read end to end (`inotifylimit_linux_test.go`) rather than only asserting arithmetic, but it runs against CI's kernel with a trivial demand — nothing has been measured on overlayfs, in a container, or under concurrency, and the 976-dirs-per-worktree figure remains one repo on one machine. This PR exists so those numbers can start coming from the target.

The unrelated flake the reviewer saw once (`TestChurnReturnsTheLedgerToItsSubscriptionBaseline`, `fdbudget_6268_test.go`, untouched here) is pre-existing and worth its own issue.

Refs #6932, #6940, #6921, #6923, #6928

### DIFF
--- a/internal/cli/status_daemon_detail.go
+++ b/internal/cli/status_daemon_detail.go
@@ -54,6 +54,18 @@ func printDaemonDetail(w io.Writer, st proto.StatusReply) {
 		}
 		fmt.Fprintln(w)
 	}
+	// #6932 arm B: the inotify budget probe. Announced with the reason and the
+	// numbers rather than as a boolean, and NOT gated on anything above — the
+	// case worth seeing is precisely the one where the watcher looks healthy.
+	// The caveats print with the number, never instead of it: the pool is
+	// per-UID and host-level, so this is grafel's own demand against a ceiling
+	// other processes are also drawing on, and grafel cannot see their share.
+	if st.InotifyBudgetSummary != "" {
+		fmt.Fprintf(w, "  %s\n", st.InotifyBudgetSummary)
+		for _, n := range st.InotifyBudgetNotes {
+			fmt.Fprintf(w, "    note: %s\n", n)
+		}
+	}
 	if st.QueueLen > 0 || len(st.IndexInFlight) > 0 ||
 		len(st.PendingAlgo) > 0 || len(st.PendingLinks) > 0 {
 		fmt.Fprintf(w, "  scheduler: queue=%d in_flight=%d pending_algo=%d pending_links=%d\n",
--- a/internal/cli/status_inotify_6932_test.go
+++ b/internal/cli/status_inotify_6932_test.go
@@ -1,0 +1,52 @@
+package cli
+
+// status_inotify_6932_test.go — #6932 arm B, the print gate for the inotify
+// budget probe.
+//
+// A number on the wire that is never printed is the #6014 defect. The caveats
+// are printed WITH the number rather than summarised away, because the number
+// alone reads as headroom the daemon owns — and the pool is per-UID and shared
+// with every other process running as the same user.
+
+import (
+	"bytes"
+	"strings"
+	"testing"
+
+	"github.com/cajasmota/grafel/internal/daemon/proto"
+)
+
+func TestPrintDaemonDetail_PrintsInotifyBudgetWithItsCaveats(t *testing.T) {
+	var buf bytes.Buffer
+	printDaemonDetail(&buf, proto.StatusReply{
+		// Zero watcher activity on purpose: the budget line must not be
+		// nested inside the `watcher:` block, whose gate is about events that
+		// arrived and says nothing about what the watch set costs.
+		InotifyBudgetSummary: "inotify budget: 9760 of 8192 watch descriptors (projected, 9760 dirs across 10 repos) — WOULD EXCEED",
+		InotifyBudgetNotes:   []string{"the inotify watch pool is per-UID and host-level", "demand is PROJECTED from a directory walk"},
+		InotifyBudgetExceeds: true,
+	})
+	out := buf.String()
+
+	for _, want := range []string{
+		"9760 of 8192 watch descriptors",
+		"WOULD EXCEED",
+		"per-UID and host-level",
+		"PROJECTED from a directory walk",
+	} {
+		if !strings.Contains(out, want) {
+			t.Errorf("`grafel status` output is missing %q — a budget the user cannot see is no "+
+				"improvement on a watcher nobody can see failing.\ngot:\n%s", want, out)
+		}
+	}
+}
+
+// A daemon whose reply carries no probe (an older daemon, or one with no
+// watcher) prints nothing rather than an empty line or a fabricated zero.
+func TestPrintDaemonDetail_NoInotifyLineWithoutAProbe(t *testing.T) {
+	var buf bytes.Buffer
+	printDaemonDetail(&buf, proto.StatusReply{WatcherRepos: 2, WatcherEvents: 400})
+	if out := buf.String(); strings.Contains(out, "inotify") {
+		t.Errorf("inotify budget line printed with no probe in the reply:\n%s", out)
+	}
+}
--- a/internal/daemon/proto/proto.go
+++ b/internal/daemon/proto/proto.go
@@ -101,6 +101,22 @@ type StatusReply struct {
 	//
 	// A non-zero count is the only way this failure is visible at all: the
 	// watcher keeps running and keeps reporting healthy through an overflow.
+	// InotifyBudget* is the #6932 arm B probe: what this configuration costs
+	// the host's per-UID inotify watch pool, and what that pool allows.
+	// Reported rather than acted on — nothing switches mode on these numbers.
+	//
+	// InotifyBudgetSummary is the one-line announcement and
+	// InotifyBudgetNotes are the caveats that go with it. The caveats are
+	// carried across the wire rather than reconstructed by the client,
+	// because the most important one — the pool is per-UID and host-level, so
+	// grafel can see its own demand and NOT anyone else's — is a property of
+	// how the number was obtained, and the client has no way to know it.
+	InotifyBudgetSummary string   `json:"inotify_budget_summary,omitempty"`
+	InotifyBudgetNotes   []string `json:"inotify_budget_notes,omitempty"`
+	// InotifyBudgetExceeds is true only when grafel's OWN projected demand is
+	// larger than the host ceiling. False when no ceiling could be read: not
+	// knowing the limit is not evidence of exceeding it.
+	InotifyBudgetExceeds   bool               `json:"inotify_budget_exceeds,omitempty"`
 	WatcherOverflows       uint64             `json:"watcher_overflows,omitempty"`
 	WatcherOverflowRescans uint64             `json:"watcher_overflow_rescans,omitempty"`
 	WatcherLastOverflow    string             `json:"watcher_last_overflow,omitempty"`
--- a/internal/daemon/proto/proto.go
+++ b/internal/daemon/proto/proto.go
@@ -101,6 +101,10 @@ type StatusReply struct {
 	//
 	// A non-zero count is the only way this failure is visible at all: the
 	// watcher keeps running and keeps reporting healthy through an overflow.
+	WatcherOverflows       uint64 `json:"watcher_overflows,omitempty"`
+	WatcherOverflowRescans uint64 `json:"watcher_overflow_rescans,omitempty"`
+	WatcherLastOverflow    string `json:"watcher_last_overflow,omitempty"`
+
 	// InotifyBudget* is the #6932 arm B probe: what this configuration costs
 	// the host's per-UID inotify watch pool, and what that pool allows.
 	// Reported rather than acted on — nothing switches mode on these numbers.
@@ -116,16 +120,13 @@ type StatusReply struct {
 	// InotifyBudgetExceeds is true only when grafel's OWN projected demand is
 	// larger than the host ceiling. False when no ceiling could be read: not
 	// knowing the limit is not evidence of exceeding it.
-	InotifyBudgetExceeds   bool               `json:"inotify_budget_exceeds,omitempty"`
-	WatcherOverflows       uint64             `json:"watcher_overflows,omitempty"`
-	WatcherOverflowRescans uint64             `json:"watcher_overflow_rescans,omitempty"`
-	WatcherLastOverflow    string             `json:"watcher_last_overflow,omitempty"`
-	QueueLen               int                `json:"queue_len,omitempty"`
-	IndexInFlight          []string           `json:"index_in_flight,omitempty"`
-	PendingAlgo            []string           `json:"pending_algo,omitempty"`
-	PendingLinks           []string           `json:"pending_links,omitempty"`
-	IndexedRepos           []IndexedRepoState `json:"indexed_repos,omitempty"`
-	RecentLog              []SchedLogEntry    `json:"recent_log,omitempty"`
+	InotifyBudgetExceeds bool               `json:"inotify_budget_exceeds,omitempty"`
+	QueueLen             int                `json:"queue_len,omitempty"`
+	IndexInFlight        []string           `json:"index_in_flight,omitempty"`
+	PendingAlgo          []string           `json:"pending_algo,omitempty"`
+	PendingLinks         []string           `json:"pending_links,omitempty"`
+	IndexedRepos         []IndexedRepoState `json:"indexed_repos,omitempty"`
+	RecentLog            []SchedLogEntry    `json:"recent_log,omitempty"`
 
 	// GroupAlgoRunning names the groups whose ANNOTATION (group-algo) pass is
 	// executing right now, and GroupAlgoInFlight counts them. The pass produces
--- a/internal/daemon/service_watcher_status.go
+++ b/internal/daemon/service_watcher_status.go
@@ -30,6 +30,10 @@ type watcherStatusSource interface {
 	FDBudgetStats() (int, int, int, []string)
 	// OverflowStats returns (overflows, rescans, coalesced, last) (#6921).
 	OverflowStats() (uint64, uint64, uint64, time.Time)
+	// InotifyBudgetReport returns (summary, notes, exceeds) for the probe
+	// that projects this configuration's cost against the host's per-UID
+	// inotify watch pool (#6932 arm B).
+	InotifyBudgetReport() (string, []string, bool)
 }
 
 // fillWatcherStatus copies the watcher's counters into the status reply.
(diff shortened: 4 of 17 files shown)


cajasmota:

Own detached worktree at `bb4493fb`, isolated `GRAFEL_HOME`/`GRAFEL_DAEMON_ROOT`, `go vet` clean, every score `-count=1` against a scoped green baseline (66 tests selected by `-run 'Inotify|Skip|Subscribe|AddRepo|Gitignore|Exclude|Watch'`, exit 0) taken immediately before, restored by `cp` from shasum-verified backups and confirmed with `git status --porcelain` = 0 + matching shasums. **All measurements below are darwin/APFS.**

The line the arm must not cross held, and the extraction is clean. Three things are not yet graded, and one of them means the announcement says the wrong thing in exactly the situation the arm exists for.

---

### 1. The `subscribeRepo` extraction IS behaviour-preserving

Line-by-line against `origin/main`: same three layers, same order, same call site position (after the TCC guard and after the dir cap, both left inline), `p != abs` preserved inside `watchDirSkip`, `w.shouldSkipDir(base)` faithfully re-expressed as `ShouldSkipDir(base) || extraSkip[base]` (the method is exactly that), and only the `watchSkipIgnored` branch logs — matching main, where layer 1+2 pruned silently and only the gitignore layer logged. No layer dropped, no order change. Agreed.

### 2. The three-number agreement IS circular — but the pin is real, and it carries the whole test

`projection`, `AddRepo`'s return and `WatchList()` are **one predicate evaluated twice plus a backend echo of the second evaluation**, not three witnesses. `WatchList()` is a genuine check that `AddRepo`'s counter is not lying about what it handed fsnotify; it is not an independent evaluation of *which* directories should be watched. The only leg that is independent of `watchDirSkip` is the literal `8`/`4` on line 99 — and `dirsOnDisk` is a bare `WalkDir` counter, so the pin is genuinely outside the predicate. Not circular.

**I reproduced mutant 4.** Dropping the `ShouldSkipDirGitignore` layer from `watchDirSkip` (asserted count=1):

```
--- FAIL: TestProjectInotifyBudget_MatchesWhatSubscriptionActuallyTakes
    inotifybudget_6932_test.go:100: fixture no longer exercises both skip layers:
        8 dirs on disk, 6 subscribed, want 8 and 4
```

That was the **only** failing assertion in the package — 104 (`proj.Dirs != actual`), 108 (`!= added`) and the `onDisk <= actual` premise all still passed at 6/6/6. The author's characterisation is exact and the pin is load-bearing.

**Is 8/4 right?** Yes. Root + `src` + `src/inner` + `pkg` = 4 watched; `node_modules{,/deep}` pruned by `SkipDirs`, `generated{,/nested}` pruned by the `generated/` gitignore rule (`generated` is not in `SkipDirs` — `_generated` is — so the gitignore layer really is the thing firing). The pin encodes the correct expectation.

**Please state the "one predicate, twice" fact in the PR body.** The table as written reads as three independent confirmations; it is one, plus a literal. That distinction is the whole reason mutant 4 was interesting.

### 3. BLOCKING — the announcement says **"fits"** precisely when the subscription was refused

`Watcher.InotifyBudget()` in the default (fsnotify) mode reports `MeasuredInotifyBudget(w.Stats())`, i.e. **the watches that succeeded**. Watches that were refused are not in that number. So the report cannot exceed a ceiling the kernel already enforced, and a degraded watcher renders as healthy. Demonstrated in-tree (darwin, `GRAFEL_WATCH_FD_BUDGET=2`, `GRAFEL_INOTIFY_MAX_USER_WATCHES=8192`, fixture with 4 watchable dirs):

```
AddRepo added=0 err=watch: file-descriptor budget exhausted ... backendWatchList=0
projection Dirs=4
SUMMARY: inotify budget: 0 of 8192 watch descriptors (measured, 0 dirs across 0 repos) — fits
```

The repo is **not watched at all** and the announcement says `fits`. On Linux the same shape arrives via `w.fsAdd` returning `ENOSPC` — logged `WARN`, `return nil`, not counted — so measured demand shrinks by exactly the shortfall and the line still reads `fits`. This is #6921/#6923/#6928's class, reproduced by the surface built to expose it.

The shared-pool note does not cover this: it warns that *headroom* is an upper bound, not that the *demand* is a lower bound. Minimum fix — a measured-mode note saying the figure counts only watches that succeeded, and that refused ones are not in it (`FDBudgetStats`/`OverflowStats` are already on the same interface, so the refused count is available). Pin it.

### 4. Mutants (mine — 4 scored, none of the author's 7 repeated except the reproduction above)

| # | mutant (call site preferred, permissive direction) | verdict |
|---|---|---|
| M1 | **call site**: `subscribeRepo` drops the `watchSkipIgnored` case — ignored dirs get subscribed, projection untouched | **DEAD** — pin (8 on disk, 6 subscribed) + `PollModeProjects…` |
| M2 | `projectWatchDirs` drops the `dirCap` gate (projection unbounded, subscription still capped) | **ALIVE, ungraded.** Not reachable in practice: `DefaultWatchDirCap = 100000`, per repo. Reachable with `GRAFEL_WATCH_DIR_CAP` set low — the projection then over-reports by the truncated tail |
| M5 | `newInotifyBudget` drops the `"%s has no inotify watch pool of its own; this is a what-if against a Linux ceiling"` note on the env path | **ALIVE — distinguishable and ungraded.** `EnvSuppliedLimitIsLabelledAsSuch` only asserts the env-var name appears; `PlatformAnswerIsHonest` `Unsetenv`s the override, so **darwin + `GRAFEL_INOTIFY_MAX_USER_WATCHES` is never graded** — and on that path `Summary()` drops `"not applicable on darwin"` and renders a Linux-shaped `N of 8192 — fits` verdict on a host with no inotify at all. Add a case for it |
| M6 | `Summary()`: `demand := "measured"` unconditionally (drop the projected/measured distinction from the **printed line**) | **ALIVE — distinguishable and ungraded.** Green on `watch`, `cli`, `dashboard`, `daemon`. The PR claims "The report says which of the two it is"; the tests assert the `Measured` **field**, never the string. Poll mode can print `(measured, 9760 dirs)` for a walk that subscribed nothing and the suite stays green. One `strings.Contains(b.Summary(), "projected")` fixes it |

M4 (the author's) reproduced above: **DEAD**, sole killer the named pin.

### 5. Can the announcement be misread as headroom?

**For the shared-pool dimension: no** — `inotifySharedPoolNote` says per-UID, host-level, cannot see other processes, upper bound not a reservation; it is on the wire, in `grafel status` and in `/diagnostics`; the caveat travels with the number on both the env and the sysctl paths; `WouldExceed` is strict, `LimitKnown` gates it, and off-Linux there is no `N of 0`. That half is done well.

**For the truncation dimension: yes** — §3. And the `fits` verdict itself is a stronger word than the data supports even in the healthy case: it is `grafel's demand ≤ ceiling`, which is not "it will fit". Suggest `fits (grafel's demand alone)`.

`Headroom()` is exported and has no non-test caller — either wire it or drop it.

### 6. Smaller

- **`proto.go` doc-comment hijack.** The new `InotifyBudget*` paragraph is appended to the *existing* `WatcherOverflows` doc block, so `WatcherOverflows` now has no doc of its own and its sentence ("A non-zero count is the only way this failure is visible at all…") documents `InotifyBudgetSummary`. Blank line + move.
- **#6940 is respected.** The poll-mode note claims only "zero descriptors right now" and says nothing about the poller's candidate set. Correct.
- **The Linux leg is asserted, not graded here.** `PlatformAnswerIsHonest`'s Linux half, `inotifyMaxUserWatchesPath` and `readInotifyLimit` run only on the ubuntu leg — the "Not established" section is honest about it, but note that no Linux leg exercises the `LimitErr` path or a *real* `max_user_watches` read against a real demand, so on Linux only the arithmetic is graded, never the probe end to end.
- **Unrelated flake seen once**: `TestChurnReturnsTheLedgerToItsSubscriptionBaseline` (`fdbudget_6268_test.go:599`, file untouched by this PR) failed in a loaded full-package run and passed `-count=3` in isolation. Pre-existing, not a blocker for this PR — but worth an issue.

---

**REQUEST-CHANGES** on M6 (one assertion), M5 (one test case), §3 (a measured-mode caveat + its pin), and the `proto.go` comment. The design, the extraction and the shared-pool honesty are all sound; what is missing is that three of the report's user-visible claims are made in prose and observed by nothing.

cajasmota:
Coordinator mutant on the final head `46c1e85d8`. One mutant, **DEAD by two separate tests — one per half of the guard**, which is the result I was hoping for and did not assume.

## MJ-1 — score the compound guard PART BY PART

`degradedClause`'s suppression is a compound condition:

```go
if b.UnwatchedRepos == 0 && b.FailedDirs == 0 { return "" }
```

Two guards that are only ever exercised together grade neither, so I flipped `&&` to `||`. Under that mutant the clause is suppressed unless **both** counters are non-zero — meaning a repo refused with no rejected directories, or every directory rejected with no refused repo, prints a clean healthy line. It is a pure widening of the silence.

Anchor count 1 pre-edit, 1 post-edit, `go vet` clean, baseline green first (`ok 130.321s`), `-count=1`, restored by `cp` from a shasum-verified backup (`1d20f4cc4…`), `git status --porcelain` empty.

**Verdict: DEAD, and the two halves are graded separately:**

```
TestWatcherInotifyBudget_RefusedRepoIsInTheDemandAndAnnounced:
  a repo that receives no events at all is announced as healthy: "… 4 of 8192
  watch descriptors (measured, plus a projection of 1 refused repo(s), …) — fits …"

TestWatcherInotifyBudget_BackendRejectedDirsAreInTheDemand:
  every directory was rejected and the line still reads healthy: "… 4 of 8192
  watch descriptors (measured, 4 dirs across 1 repos) — fits …"
```

Each half fails on its own, on an input the other cannot produce, and both failure messages quote the **rendered line** rather than a field. That is the difference between a compound guard that is graded and one that merely happens to fire.

## On the blocker fix

The correction is the right one and it is the arm's whole point. The demand now counts what was **attempted** — a refused repo's tree projected back in, and a directory whose `fsAdd` failed (the Linux `ENOSPC` shape, previously WARN-and-carry-on) counted and added back — so a failure can no longer make the report *cheaper*. The reviewer's repro now reads `4 dirs / 1 refused repo` instead of `0 of 8192 — fits`.

The verdict word `fits (grafel's demand alone)` is doing real work: it is a claim about grafel's own demand and it says so, in a place where the surrounding numbers invite the reader to hear "you have headroom" from a pool that is per-UID, host-level and shared with every other process running as the same user.

Keeping the healthy case **silent** is right for the reason stated in the comment — a line that always prints trains the reader to ignore it — and it is exactly why the compound guard needed scoring part by part. Silence is the default, so any bug in the condition fails toward silence.

Two of the reviewer's items came back better than asked: **M2 was graded rather than recorded** (`WatchDirCap()` reads `GRAFEL_WATCH_DIR_CAP` per call, so a cap of 2 pins projection == subscription), and the Linux leg is now **graded end to end** by a real sysctl read plus the unparseable and absent branches, instead of asserting arithmetic. Turning "not reachable in practice" into a graded case beats recording it, whenever the environment can be bent to reach it.

Merging on green, `ci:full` confirmed still applied. Refs #6932, #6940.

Issue #6934, opened 2026-09-06
The watcher's event boundary reads only the ROOT .gitignore — nested .gitignore, .grafelignore and both #6922 exclude sources are invisible to it, so it disagrees with the walker in both directions


cajasmota:
Found while fixing #6931 / #6922 (PR #6933). Reported, not fixed there — it is a different seam: the walker builds an `IgnoreStack` as it descends, the watcher caches one parsed file per repo in `loadRepoIgnoreState`.

## What the watcher's event boundary consults

`ShouldSkipPathForRepo` (`internal/daemon/watch/skip.go:229`) applies **only** `<repo>/.gitignore`, loaded once by `loadRepoIgnoreState` (`internal/daemon/watch/gitignore.go:64`). After PR #6933 the walker applies four things the watcher still does not:

1. **Nested `.gitignore` files.** A rule in `pkg/sub/.gitignore` — including a negation that *un*-ignores a path the root ignores — is invisible to the event boundary.
2. **`.grafelignore`**, at any level, including the inherited-ancestor form `walk.inheritedGrafelIgnores` handles.
3. **`.git/info/exclude`** (#6922).
4. **`core.excludesFile`** (#6922).

For 1 and 2 the watcher is the *stricter* observer (it can drop an event for a path a child `.gitignore` re-includes); for 3 and 4 it is the *looser* one (it arms a reindex for a path the walker will not index). Both directions are a walker/watcher disagreement.

## A second, smaller one: `dirOnly` over-match

`ShouldSkipPathForRepo` calls `MatchDir` on every prefix of the path **including the full path**, i.e. it asks the file question with `isDir=true`. PR #6933 added `IgnoreFile.MatchPath(rel, isDir)` precisely because a trailing-slash rule (`build/`) must match a directory named `build` and never a *file* named `build`. At the event boundary a file named `build` in a repo whose `.gitignore` says `build/` is therefore dropped, where git and (now) the walker keep it.

Not trivially fixable by passing `isDir=false`: the boundary sees paths from fsnotify events, including removals, so it cannot always stat the path to learn its kind. Whether to stat, to assume "file" for the last segment and "dir" for every prefix, or to leave it is a design question, not a one-liner.

## Practical size — not measured

Unknown. The consequence of 1/2 is a **missed reindex** (a dropped event for a file the walker does index), which is the silent class #6932 catalogues. The consequence of 3/4 is wasted reindex churn, which is loud and self-correcting. No corpus count was taken.

[...]

Issue #1629, opened 2026-05-22
indexer: walker indexes junk (.windsurf, assets/images/*.png, docs, metro.config) — skip non-source/asset/generated dirs


cajasmota:
Owner: indexing is slow/bloated because the file walker indexes non-source noise — the per-module progress showed it processing `.windsurf/skills/*.md`, `assets/images/icon-dark.png`, `docs/*.md` (generated), `metro.config.js`, etc. (528 "files" for core-mobile, mostly junk). FIX the candidate-file discovery to SKIP: tool/agent dirs (.windsurf, .github, .vscode, .idea), asset/binary dirs + files (assets/, images/, *.png/*.jpg/*.svg/*.ico/fonts), generated docs (the docs/ dir — esp. now that docs move to the daemon store), build output (dist/build/out/.next/coverage), deps (node_modules/.venv/vendor) — index only real source. Reduces index time + graph noise. Coordinate with #1626 (both touch cmd/archigraph/index.go discovery) — sequence AFTER #1626. ACCEPTANCE: re-indexing a repo discovers only source files (no pngs/.windsurf/generated-docs), index is faster + smaller.

PR #6951, opened 2026-09-07
fix(#6946): `.github` was on the walker's pre-ignore skip list, so every GitHub Actions rule was dead on a real repo — measured the blast radius, then un-skipped

### LINKED ISSUE #6946: Every GitHub Actions rule in the engine is unreachable on a real repo: `.github` is on the walker's hard-coded skip list
Every GitHub Actions rule in the engine is unreachable on a real repo: `.github` is on the walker's hard-coded skip list

Found while measuring the GitHub Actions arm of #6927.

`internal/daemon/walk/walker.go` lists `".github"` in `hardcodedSkipDirs` (the "tool / agent dirs"
group, issue #1629), and `hardcodedSkip` skips the directory outright. So no file under
`.github/` is ever walked, and therefore never classified, never handed to `Detect`, and never
reaches the yaml extractor.

That makes the whole GitHub Actions rule set dead on any repo that stores its workflows where
GitHub requires them to be:

* `internal/engine/rules/cicd/frameworks/github_actions.yaml` `file_conventions` declares
  `.github/workflows/*.yml`, `.github/workflows/*.yaml` and `.github/actions/*/action.yml` —
  three globs that can never match a walked path.
* All eleven `source_patterns` in that file, and the structured workflow extractor in
  `internal/extractors/yaml`, only ever see workflow YAML that happens to live somewhere else
  (GitHub's own `starter-workflows` repo is such a layout; a normal repo is not).

## Measured, not read

Two byte-identical workflow files, one at `src/.github/workflows/probe.yml` and one at
`src/ci/workflows/probe.yml`, indexed through `grafel quality`:

```
grafel: discovered 1 candidate files in .../probe-dotgithub/src
grafel: processed=1 extracted=1 skipped=0 failed=0
```

One, not two. The `.github/` copy was not discovered.

## Why this matters beyond GitHub Actions

`.gitlab` and `.circleci` are on the same list, so the same argument applies to any rules targeting
those. The list's stated rationale is "checked-in tool-config noise that is not source and should
never enter the graph" — which is a defensible call for `.claude` or `.vscode`, but CI pipeline
definitions are exactly the kind of architecture the graph claims to model (this repo's own
`vera_scope` for GitHub Actions is "Service (pipeline as Service, job as Operation)").

## Not established

* Whether the exclusion was deliberate for CI dirs specifically, or inherited wholesale from the
  #1629 tool-dir sweep.
* The right fix. Options: drop `.github`/`.gitlab`/`.circleci` from `hardcodedSkipDirs`; or keep the
  directory skipped but carve out `workflows/` and `actions/`; or leave the skip and delete the
  unreachable rules and globs so the coverage figures stop counting them.
* The blast radius of un-skipping: `.github/` also holds issue templates, `dependabot.yml`,
  `CODEOWNERS` and markdown, which is presumably what the skip was protecting the graph from.

Refs #6927, #1629.

cajasmota:
`hardcodedSkip` runs at walker.go:214, BEFORE the ignore stack, so `.github` on
`hardcodedSkipDirs` is not a default a repo can undo — it is an unconditional
erasure. Every `source_pattern` in
internal/engine/rules/cicd/frameworks/github_actions.yaml, and its three
`.github/...` file_conventions globs, were unreachable on any repo that stores
its workflows where GitHub requires them. Same for `.gitlab` and `.circleci`.

Measured first, because un-skipping changes what EVERY repo indexes, and the
primary consumer is concurrent MCP agents on a 16GB machine.

## Blast radius

58 corpus repos in ~/Projects/archigraph-corpora + this repo. Walk-level from a
`walk.WalkRepo` harness; entity-level end-to-end with `grafel quality` on
rsync'd copies, base binary vs un-skipped binary, isolated HOME/GRAFEL_HOME.

  walked files      31279 -> 31468   (+189, +0.60%)
  repos affected    30 of 59; 29 have no CI dir on disk at all (several are
                    sparse checkouts, where Layer 5 would drop the tree anyway)
  content           189 files / 277KB: 121 .yml, 21 .yaml, 38 .md,
                    5 extensionless, 2 .tpl, 1 .sh, 1 .json

`.gitlab` was measured on NOTHING. The corpus contains zero `.gitlab/`
directories, so these numbers cover `.github` plus three `.circleci` files. The
`.gitlab` un-skip is graded only by the synthetic walker test. It is the same
argument — a pipeline definition is architecture — but it is an argument, not a
measurement, and should not be read as covered by the figures above.

  entities, base -> un-skipped:
    django             87897 -> 88313  (+0.47%),  rels +690 (+0.08%)
    apollo-server      11551 -> 11617  (+0.57%)
    gin                 3967 ->  4091  (+3.1%)
    starter-workflows   3745 ->  3861  (+3.1%)
    flask               3261 ->  3354  (+2.9%)
    click               2421 ->  2513  (+3.8%)
    terraform-aws-vpc   2091 ->  2185  (+4.5%)
    requests            1723 ->  1894  (+9.9%)
    express             1663 ->  1798  (+8.1%)
    spring-petclinic     934 ->   991  (+6.1%)
    dart-samples         736 ->   897  (+21.9%, one 56KB workflow)
    tide                 307 ->   371  (+20.8%)

### Cost

The load-bearing figure is structural, not a timing run: django's `.github` is
43,365 bytes against 27.7MB of .py/.html/.txt/.js source — 0.16% of the source
bytes fed to the extractors. Cost here is linear in bytes walked, so the
expected peak-RSS delta is below the floor of any single measurement by
construction.

The observed peak RSS on django (`/usr/bin/time -l`, n=1) was 2058MB base ->
2047MB un-skipped. That is consistent with no cost and rules out a large
regression; it cannot establish "free". macOS max-RSS counts MADV_FREE pages,
so it is an upper bound rather than a footprint, and n=1 carries no variance
estimate. Read it as a ceiling, and the 0.16% as the reason.

### Outliers

"No outlier" holds for directory SIZE: largest `.github` is django's at 24
files / 43KB, largest single file is `dart-samples/.github/workflows/dart.yml`
at 56KB. It does NOT hold for entity YIELD, and the outlier is at the small end,
not the large one: dart-samples +21.9% and tide +20.8%, the former from that one
56KB workflow. A single oversized workflow moves a small repo's entity mix by a
fifth — and #6932's RSS-constrained container case is a small-repo case.

The untested mega-repo end (a kubernetes-scale `.github` of ~100 workflows) is
acceptable rather than merely unmeasured: cost is linear in bytes and a big
repo's source tree grows faster than its CI config, so the ratio only improves.

### Signal ratio, both denominators

By ENTITY yield, 83% of requests' +170 and 69% of tide's +64 come from
`.github/workflows/*`, as Service / Task / SCOPE.Operation / SCOPE.ScheduledJob.
By FILE count the split is much worse: only 90 of the 164 corpus `.github`
files (55%) live under `workflows/` or `actions/`; 74 are issue/PR templates,
FUNDING.yml and markdown. Notably the un-skip re-admits this repo's own
`.github/copilot-instructions.md` — a file `internal/install/rulesfiles`
GENERATES as tool config, i.e. exactly the #1629 class the skip list exists for.
Trimming that tail needs no new mechanism (a basename entry for
`ISSUE_TEMPLATE` plus a loose-`.md` rule removes 29% of the addition in two
lines); filed as #6954 rather than done here.

## Why option 1 and not the alternatives

  - Option 2 (walk only `.github/workflows` + `.github/actions`) is feasible —
    every call site already holds the full path — but costs one new
    `IsHardcodedSkipPath(rel)`, a change at each of four sites in walk, watch
    (x2) and sched, plus a genuine file-level rule, since `hardcodedSkip` only
    returns SkipDir for directories. #6954's two basename rules capture most of
    the benefit for none of that, and should be tried first.
  - Option 3 (leave the skip, retarget the globs) makes the rules fire on files
    stored in the WRONG place while still never seeing the real ones. It is now
    graded (M3 below) rather than merely rejected in prose.

## Grading, both directions

  - internal/daemon/walk/ci_dirs_6946_test.go — the three CI dirs are walked;
    the tool-agent dirs and node_modules must still be skipped; IsHardcodedSkip
    agrees with the walker in both directions. A second test pins that an
    un-skipped dir is still governed by .grafelignore, with .circleci as the
    negative control.
  - internal/daemon/watch/ci_dirs_parity_6946_test.go — ShouldSkipDir /
    ShouldSkipPath must not drop these three. `watch.SkipDirs` is a genuine
    SECOND hand-maintained copy of the skip names, consulted BEFORE the
    delegation to walk.IsHardcodedSkip; until this test the walker/watcher
    parity this PR asserts held only by luck.
  - cmd/grafel/ci_dirs_stale_parity_6946_test.go — the THIRD copy,
    shouldSkipDirForStale ("keep in sync with watch.ShouldSkipDir"), same rows.
  - internal/engine/github_actions_reachable_6946_test.go — grades both axes:
    every declared file_conventions glob must match a path walk.WalkRepo
    actually emits for a standard GitHub layout (reachability), and must NOT
    match ordinary YAML at paths GitHub never reads (specificity). Neither
    axis alone catches anything: matchesFile was always true for the real
    globs, and a glob of `*.yml` reaches everything.

## Mutants

Each edit's occurrence count asserted; baseline verified green immediately
before each score; restored by cp + shasum.

  M1  re-add the three dirs to hardcodedSkipDirs                      -> DEAD
  M2  glob -> "workflows/*.yml"                                       -> DEAD
  M3  globs -> "*.yml" / "*.yaml"; matchesFile falls back to a BASENAME
      match for slash-free globs, so every .yml in the repo becomes a
      GHA Config — option 3 in a spelling neither the issue nor I named
      -> ALIVE against the first version of the engine test (green on
      walk, engine, quality, extractors and cmd/grafel). Added the decoy
      rows; re-scored DEAD.
  M6  add the three names to watch.SkipDirs — reproduces #6934 exactly:
      walker indexes the dir, watcher never re-indexes it on change
      -> ALIVE against everything in the first push (green on walk,
      watch, sched, engine). Added the parity tests; re-scored DEAD.

M1/M2 were scored by me, M3/M6 by the adversarial review of #6951.

## Verification

go test -count=1 green on internal/daemon/walk, internal/daemon/watch,
internal/daemon/sched, internal/engine, internal/quality, internal/extractors
and ./cmd/grafel (the graph digest pin did NOT move). gofmt and go vet clean.

The extraction ratchet did not move either — and that is a finding, not a
reassurance: no fixture under internal/quality/golden contains a .github,
.gitlab or .circleci dir, and cmd/grafel's fixtures are all t.TempDir()
synthesis, so both the ratchet and the digest pin are blind to this change by
construction. Filed as #6955; the end-to-end cicd golden fixture belongs there.

Closes #6946
Refs #6954, #6955

### DIFF
--- a/cmd/grafel/ci_dirs_stale_parity_6946_test.go
+++ b/cmd/grafel/ci_dirs_stale_parity_6946_test.go
@@ -1,0 +1,24 @@
+package main
+
+import "testing"
+
+// TestShouldSkipDirForStale_CIDirs_6946 pins the THIRD hand-maintained copy of
+// the watcher skip list (shouldSkipDirForStale, whose doc comment says "keep in
+// sync with internal/daemon/watch.ShouldSkipDir"). A copy that drifts here
+// makes the daemon's stale-detection walk blind to a change under .github, so
+// a workflow edit never marks the graph dirty even though the walker indexes
+// it. See TestShouldSkipDir_CIDirsMatchWalker_6946 for the second copy.
+func TestShouldSkipDirForStale_CIDirs_6946(t *testing.T) {
+	for _, base := range []string{".github", ".gitlab", ".circleci"} {
+		if shouldSkipDirForStale(base) {
+			t.Errorf("shouldSkipDirForStale(%q) = true, want false (#6946) — the walker indexes "+
+				"this directory, so stale detection must see writes into it", base)
+		}
+	}
+	for _, base := range []string{".git", "node_modules", ".idea"} {
+		if !shouldSkipDirForStale(base) {
+			t.Errorf("shouldSkipDirForStale(%q) = false, want true — the #6946 un-skip reached "+
+				"further than the three CI dirs", base)
+		}
+	}
+}
--- a/internal/daemon/walk/ci_dirs_6946_test.go
+++ b/internal/daemon/walk/ci_dirs_6946_test.go
@@ -1,0 +1,121 @@
+package walk
+
+import (
+	"strings"
+	"testing"
+)
+
+// TestWalkRepo_CIDirsAreWalked_6946 grades BOTH directions of the #6946
+// change: the CI-definition dirs that came off hardcodedSkipDirs must now be
+// walked, and the tool-agent dirs that stayed on it must still be skipped.
+//
+// The skip mattered because hardcodedSkip runs BEFORE the ignore stack
+// (walker.go), so a `.github` on the list is not a default a repo can undo —
+// it is an unconditional erasure. Every GitHub Actions rule in
+// internal/engine/rules/cicd/frameworks/github_actions.yaml was therefore
+// unreachable on any repo that stores its workflows where GitHub requires
+// them. See TestGitHubActionsFileConventionsAreReachable_6946 in
+// internal/engine for the other half of that claim.
+func TestWalkRepo_CIDirsAreWalked_6946(t *testing.T) {
+	root := t.TempDir()
+
+	// Must be walked (#6946).
+	wantWalked := []string{
+		".github/workflows/ci.yml",
+		".github/workflows/release.yaml",
+		".github/actions/setup/action.yml",
+		".github/dependabot.yml",
+		".gitlab/ci-templates/build.yml",
+		".circleci/config.yml",
+		"src/main.go",
+	}
+	// Must still be skipped: tool-agent config that is not source, and the
+	// build/VCS dirs. If one of these ever leaks the un-skip was too wide.
+	wantSkipped := []string{
+		".claude/settings.json",
+		".cursor/rules/foo.json",
+		".windsurf/skills/coding.md",
+		".husky/pre-commit",
+		".devcontainer/devcontainer.json",
+		"node_modules/pkg/index.js",
+	}
+	for _, p := range append(append([]string{}, wantWalked...), wantSkipped...) {
+		mkfile(t, root, p, "x")
+	}
+
+	files, _, err := WalkRepo(root, nil)
+	if err != nil {
+		t.Fatalf("WalkRepo: %v", err)
+	}
+	got := make(map[string]bool, len(files))
+	for _, f := range files {
+		got[f] = true
+	}
+
+	for _, want := range wantWalked {
+		if !got[want] {
+			t.Errorf("#6946: %q was not walked; files=%v", want, files)
+		}
+	}
+	for _, notWant := range wantSkipped {
+		if got[notWant] {
+			t.Errorf("%q leaked into the walk results — the #6946 un-skip is too wide", notWant)
+		}
+	}
+
+	// The exported predicate the watcher and the RSS scanner consult must
+	// agree with the walker, or a repo would be walked and then never
+	// re-indexed on change (the walker/watcher divergence class of #6934).
+	for _, base := range []string{".github", ".gitlab", ".circleci"} {
+		if IsHardcodedSkip(base) {
+			t.Errorf("IsHardcodedSkip(%q) = true, want false (#6946) — the watcher would still "+
+				"drop the event for a directory the walker now indexes", base)
+		}
+	}
+	for _, base := range []string{".claude", ".cursor", ".windsurf", ".husky", ".devcontainer", "node_modules"} {
+		if !IsHardcodedSkip(base) {
+			t.Errorf("IsHardcodedSkip(%q) = false, want true — the #6946 un-skip removed more "+
+				"than the three CI dirs", base)
+		}
+	}
+}
+
+// TestWalkRepo_CIDirsStillHonourIgnoreStack_6946 pins the reason the skip was
+// wrong rather than merely unhelpful: the hard-coded layer runs before the
+// ignore stack and so cannot be overridden, while an un-skipped directory is
+// governed by the repo's own .gitignore/.grafelignore like anything else.
+//
+// Without this a "fix" that hard-coded `.github` as always-walked would pass
+// the test above and still take the choice away from the user.
+func TestWalkRepo_CIDirsStillHonourIgnoreStack_6946(t *testing.T) {
+	root := t.TempDir()
+	mkfile(t, root, ".github/workflows/ci.yml", "name: ci")
+	mkfile(t, root, ".circleci/config.yml", "version: 2.1")
+	mkfile(t, root, "src/main.go", "package main")
+	writeIgnoreFile(t, root, ".grafelignore", ".github/\n")
+
+	files, _, err := WalkRepo(root, nil)
+	if err != nil {
+		t.Fatalf("WalkRepo: %v", err)
+	}
+	for _, f := range files {
+		if strings.HasPrefix(f, ".github/") {
+			t.Errorf("%q was walked despite a .grafelignore entry — an un-skipped dir must "+
+				"remain governed by the ignore stack (#6946)", f)
+		}
+	}
+	got := make(map[string]bool, len(files))
+	for _, f := range files {
+		got[f] = true
+	}
+	// Negative control: the ignore file named only .github, so .circleci must
+	// still be walked. Without this row a walker that skipped every CI dir
+	// again would pass.
+	if !got[".circleci/config.yml"] {
+		t.Errorf(".circleci/config.yml was not walked; the .grafelignore named .github only; files=%v", files)
+	}
+	if !got["src/main.go"] {
+		t.Errorf("src/main.go was not walked; files=%v", files)
+	}
+
+}
--- a/internal/daemon/walk/gitignore_test.go
+++ b/internal/daemon/walk/gitignore_test.go
@@ -548,15 +548,18 @@ func TestWalkRepo_SkipsLinguistGeneratedDir(t *testing.T) {
 }
 
 // TestWalkRepo_SkipsToolAgentDirs verifies issue #1629: AI / pair-programmer
-// and CI metadata dirs are filtered at the walker so they never appear in
-// the index. These dirs hold .md / .json config — not source — and were
-// previously inflating per-module file counts (e.g. .windsurf/skills/*.md).
+// dirs are filtered at the walker so they never appear in the index. These
+// dirs hold .md / .json config — not source — and were previously inflating
+// per-module file counts (e.g. .windsurf/skills/*.md).
+//
+// #6946 removed the CI dirs (.github / .gitlab / .circleci) from this claim:
+// a pipeline definition IS source-graph content. See
+// TestWalkRepo_CIDirsAreWalked_6946.
 func TestWalkRepo_SkipsToolAgentDirs(t *testing.T) {
 	root := t.TempDir()
 	mkfile(t, root, ".windsurf/skills/coding.md", "skill")
 	mkfile(t, root, ".cursor/rules/foo.json", "{}")
 	mkfile(t, root, ".claude/settings.json", "{}")
-	mkfile(t, root, ".github/workflows/ci.yml", "name: ci")
 	mkfile(t, root, "src/main.go", "package main")
 
 	files, skipped, err := WalkRepo(root, nil)
@@ -568,14 +571,14 @@ func TestWalkRepo_SkipsToolAgentDirs(t *testing.T) {
 	for _, s := range skipped {
 		skippedNames[filepath.Base(s.AbsPath)] = true
 	}
-	for _, want := range []string{".windsurf", ".cursor", ".claude", ".github"} {
+	for _, want := range []string{".windsurf", ".cursor", ".claude"} {
 		if !skippedNames[want] {
 			t.Errorf("expected %q to be skipped (issue #1629); skipped=%v", want, skipped)
 		}
 	}
 
 	for _, f := range files {
-		for _, prefix := range []string{".windsurf/", ".cursor/", ".claude/", ".github/"} {
+		for _, prefix := range []string{".windsurf/", ".cursor/", ".claude/"} {
 			if strings.HasPrefix(f, prefix) {
 				t.Errorf("tool-agent file leaked into walk results: %q", f)
 			}
@@ -758,7 +761,7 @@ func TestDefaultWalkerHelpers(t *testing.T) {
 // watcher and scheduler (which use IsHardcodedSkip to short-circuit).
 func TestIsHardcodedSkip_NewEntries(t *testing.T) {
 	for _, name := range []string{
-		".windsurf", ".cursor", ".claude", ".github",
+		".windsurf", ".cursor", ".claude",
 		"assets", "images", "fonts", "media", "docs",
 		".cache", "bin", "obj", ".terraform", ".ruff_cache",
 	} {
--- a/internal/daemon/walk/walker.go
+++ b/internal/daemon/walk/walker.go
@@ -543,10 +543,22 @@ var hardcodedSkipDirs = map[string]struct{}{
 
 	// Tool / agent dirs (issue #1629) — checked-in tool-config noise that
 	// is not source and should never enter the graph. Cover the popular
-	// AI / pair-programming and CI metadata dirs.
-	".github":          {},
-	".gitlab":          {},
-	".circleci":        {},
+	// AI / pair-programming dirs.
+	//
+	// #6946: ".github", ".gitlab" and ".circleci" were on this list and are
+	// NOT any more. They are not tool-config noise: they are where CI
+	// pipeline definitions live, and a pipeline is architecture the graph
+	// claims to model (the GitHub Actions rule set scopes a workflow as a
+	// Service and a job as an Operation). Because this check runs BEFORE the
+	// ignore stack, the skip made every GitHub Actions rule — and the three
+	// `.github/...` file_conventions globs in
+	// internal/engine/rules/cicd/frameworks/github_actions.yaml —
+	// unreachable on any repo that stores its workflows where GitHub
+	// requires them. Measured blast radius of the un-skip over 58 corpus
+	// repos + this one: +189 walked files (+0.60%), +0.5% entities on the
+	// largest repo measured (django), whose .github is 0.16% of its source
+	// bytes. The corpus holds no .gitlab dir, so that one is graded by the
+	// synthetic walker test only.
 	".husky":           {},
 	".devcontainer":    {},
 	".claude":          {},
--- a/internal/daemon/watch/ci_dirs_parity_6946_test.go
+++ b/internal/daemon/watch/ci_dirs_parity_6946_test.go
@@ -1,0 +1,37 @@
+package watch
+
+import "testing"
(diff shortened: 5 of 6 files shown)


cajasmota:
## Independent adversarial review — REQUEST-CHANGES (two cheap items; the change itself is right)

The bug is real, the fix is minimal, and the two new tests are good ones — they killed 2 of my 4
mutants, including the one they were written for. My objections are to two claims in the PR body
that are load-bearing and currently unpinned, not to the un-skip.

Measured in my own worktree at `df0c6635a` (`scratchpad/review-6951-adv9q/wt`, detached), baseline
`go test -count=1 ./internal/daemon/walk/... ./internal/engine/...` green (exit 0) immediately
before every score, every edit's occurrence count asserted, restore by `cp` from a shasum-verified
backup, tree verified clean vs HEAD afterwards.

### Mutants

| # | Mutant (production, permissive direction preferred) | Result |
|---|---|---|
| M3 | `github_actions.yaml`: `.github/workflows/*.yml` → `*.yml`, `…*.yaml` → `*.yaml` (2 edits; `.github/actions/*/action.yml` left intact so the `dotGitHub` row still sees one). `matchesFile` falls back to basename matching when a glob has no `/`, so this makes **every `.yml` in the repo** a GitHub Actions `Config`. | **ALIVE — distinguishable and ungraded.** Green on `walk`, `engine`, and also on `./internal/quality/... ./internal/extractors/... ./cmd/grafel/...` |
| M4 | `walker.go:208`: `igStack.Match(rel); skip` → `skip && base != ".github"` (1 edit) — the "hard-code `.github` as always-walked" non-fix. | **DEAD** — `TestWalkRepo_CIDirsStillHonourIgnoreStack_6946`. The negative control earns its place. |
| M5 | `IsHardcodedSkip`: early `if base == ".github" { return true }`, map unchanged — walker walks it, watcher/RSS scanner drop it (#6934 shape). | **DEAD** — `TestWalkRepo_CIDirsAreWalked_6946`. Note `./internal/daemon/watch/...` stayed **green**: the divergence is graded only from the `walk` side. |
| M6 | `internal/daemon/watch/skip.go`: add `".github"`, `".gitlab"`, `".circleci"` to **`watch.SkipDirs`** — the watcher's own hand-maintained list (1 edit, 3 entries, gofmt+vet clean). Reproduces the exact #6934 divergence: walker indexes the dir, watcher never re-indexes it on change. | **ALIVE — distinguishable and ungraded.** Green on `walk`, `watch`, `sched`, `engine`. |

**M6 is the ask.** The PR body states "watch.ShouldSkipDir keeps no list of its own for these names …
no walker-only divergence of the #6934 class." That is true *today* and nothing pins it. `SkipDirs`
is a genuine second copy (it duplicates `.git`, `.claude`, `node_modules`, `.idea`, …) with a comment
that invites additions, and `cmd/grafel/daemon_tier.go:518` carries a third, hand-maintained
"keep in sync" copy. Neither contains these three names, so this PR introduces no divergence — but
one assertion in `internal/daemon/watch` would make the claim a graded one:

```go
for _, b := range []string{".github", ".gitlab", ".circleci"} {
    if ShouldSkipDir(b) { t.Errorf("ShouldSkipDir(%q) = true; the walker indexes it (#6946)", b) }
}
```

M3 is a pre-existing gap, not something this PR introduced — but it is worth recording that the
*specificity* of those globs is graded by nothing in the tree, in either the engine or the golden
tiers. Reachability and specificity are different axes and this PR only closes the first.

### RSS: the number is fine, the sentence is not

"2058MB → 2047MB … no measurable cost; the delta is run-to-run noise" is n=1 with no variance
estimate, on macOS `/usr/bin/time -l`, whose *maximum resident set size* counts `MADV_FREE` pages
and is therefore an upper bound rather than a footprint. A single pair of numbers 0.53% apart cannot
distinguish "free" from "+0.5%"; it can only rule out a large regression.

What the data *does* support is stronger and doesn't need a repeat: django's `.github` is
**43,365 bytes against 27.8MB of `.py`/`.html`/`.txt`/`.js` — 0.16% of the source bytes fed to the
extractors.** The expected peak-RSS delta is structurally below the noise floor of any single
measurement. Please reword to that, e.g. *"the addition is 0.16% of django's source bytes; the
observed peak-RSS delta (−11MB on 2058MB) is below the measurement floor and consistent with no
cost"* — and drop "no measurable cost", which will be quoted as "measured free". A repeat that would
actually license the original phrasing is 5 runs per arm with the spread reported; I do not think it
is worth the machine time, which is why the reword is the cheaper fix.

### Mega-repo gap: acceptable — and the corpus outlier is at the *other* end

Cost here is linear in bytes walked, and a kubernetes-scale `.github` is ~100 workflows of a few KB
each against a source tree two orders of magnitude larger; the ratio only gets *better* as the repo
grows. I would not block on it. The untested end is the opposite one, and the PR body already has
the evidence without drawing the conclusion: **tide +20.8%, dart-samples +21.9% — the latter from a
single 56KB `dart.yml`.** "No outlier directory anywhere in the corpus" is a claim about directory
*size* and reads, in context, as covering the entity yield it does not cover. Please qualify it: one
oversized workflow file is exactly the shape that moves a small repo's entity mix by a fifth, and
that is an outlier by any reading. #6932's RSS-constrained container case is a small-repo case too.

### Verified independently

- **Ratchet blindness: confirmed.** `find internal/quality -type d \( -name .github -o -name .gitlab
  -o -name .circleci \)` → **0**. Repo-wide the only such dir is grafel's own `./.github`. The
  `cmd/grafel` tests synthesize fixtures into `t.TempDir()` and none writes a `.github`, so the
  digest pin is blind for the same reason. The finding is stated accurately and it is the right
  framing — this is a change to what every repo indexes, protected end-to-end by nothing but the two
  unit tests in this PR.
- **The +189 corroborates.** Independently over `~/Projects/archigraph-corpora`: 164 `.github` files
  + 3 `.circleci` + 19 in grafel's own `.github` = **186**, against the claimed 189. Good.
- **`.gitlab` was measured on nothing.** There is not a single `.gitlab/` directory in the corpus
  (`find */.gitlab` → 0 matches). The blast-radius numbers cover `.github` and 3 `.circleci` files;
  the `.gitlab` un-skip is graded only by the synthetic walker test. Fine on the merits, but "58
  corpus repos" should not be read as covering it.

### Option 2: overstated, but the conclusion survives — and there is a cheaper narrowing

I read all four call sites. Every one of them already has the full path in hand at the moment it
consults the basename predicate: `walker.go` has `rel` two lines above the `hardcodedSkip` call;
`watcher.go:733` computes `relPath` immediately *below* `w.shouldSkipDir(base)`; `sched/rss.go:37`
has `p` and the repo root; `watch.ShouldSkipPath` already reconstructs the path segment by segment.
So "a new path-aware mechanism in three packages" overstates it — realistically one new exported
`IsHardcodedSkipPath(rel string)` plus a one-to-three-line change at each of four sites. The
file-level rule for loose files directly under `.github` is a genuine extra, since `hardcodedSkip`
only returns `SkipDir` for directories.

The conclusion still holds, but not for the reason given, and the "15–30%" is the wrong denominator.
By **entity** yield the non-workflow share is 17–31% (your 83%/69% signal ratios). By **file** count
it is much larger: of the 164 corpus `.github` files, only **90 (55%) live under `workflows/` or
`actions/`** — 74 are issue/PR templates, `FUNDING.yml`, `CODE_OF_CONDUCT.md` and friends, 35 of them
`.md`. Please label the 15–30% as the entity figure.

That 45% matters because the walker hard-skips `docs/`, `doc/`, `_posts/` on the stated grounds that
"hand-authored markdown is not source for the graph" — and this PR now admits
`.github/ISSUE_TEMPLATE/*.md` and `.github/PULL_REQUEST_TEMPLATE.md` under the same roof. Concretely,
grafel will now index its own **`.github/copilot-instructions.md`** — a file `internal/install/rulesfiles`
*generates* as tool config, i.e. precisely the #1629 class the skip list exists for.

None of that requires path-awareness to fix. A basename entry for **`ISSUE_TEMPLATE`** alone removes
26 of the 164 files, and a loose-`.md`-under-`.github` file rule another 22 — **29% of the addition
for two lines and no new mechanism**. Worth a follow-up; I am not asking for it here.

### Asks before merge

1. The M6 assertion in `internal/daemon/watch` (3 lines). It pins the PR's own walker/watcher claim,
   which is the one claim here with a history of going wrong silently.
2. Reword the RSS sentence to what n=1 supports (0.16%-of-source-bytes framing), and qualify "no
   outlier" as being about directory size, with dart-samples/tide named as the entity-yield outliers.

Follow-ups, not blockers: the end-to-end cicd golden fixture (agreed it belongs with #6927 — but
please file it now rather than leaving it implied, since it is the only thing that would make the
ratchet non-blind); a glob-specificity assertion (M3); and the `ISSUE_TEMPLATE` / loose-`.md` trim.

cajasmota:
All four asks addressed; both ALIVE mutants are now DEAD. Pushed as `78810ceb` (force-with-lease over `df0c663`).

**1. M6 — pinned, in two places.** `internal/daemon/watch/ci_dirs_parity_6946_test.go` asserts `ShouldSkipDir` and `ShouldSkipPath` do not drop the three, with `.claude` / `node_modules` / `.idea` as negative controls so it cannot pass against a watcher that skips nothing. You named a third copy, so I pinned that too: `cmd/grafel/ci_dirs_stale_parity_6946_test.go` on `shouldSkipDirForStale`. Re-scored your M6 (3 entries added to `watch.SkipDirs`, occurrence count 3 asserted, gofmt clean, baseline green immediately before) — **DEAD**, 6 failing rows. Restored by `cp` from a shasum-verified backup.

**2. M3 — graded.** Reachability and specificity are, as you say, different axes, and the first test only closed the first. The engine test now also requires every declared glob to REJECT decoy YAML at paths GitHub never reads (`config/app.yml`, `deploy/values.yaml`, `src/actions/setup/action.yml` — the last one aimed at the `actions/*/action.yml` glob). Re-scored your M3 (`*.yml` + `*.yaml`, 2 edits) — **DEAD**, 3 rows naming the basename fallback.

**3. RSS reworded.** "No measurable cost" is gone. The body now leads with the structural figure — 43,365 B against 27.7 MB of source, **0.16%**, which I re-derived independently (I get 26.4 MiB where you wrote 27.8 MB; same bytes, different unit convention) — and demotes the RSS pair to a ceiling, explicitly naming n=1 and MADV_FREE.

**4. "No outlier" qualified.** Restated as a claim about directory *size*, with dart-samples +21.9% (one 56 KB `dart.yml`) and tide +20.8% named as the entity-yield outliers at the small end, and #6932's container case noted as a small-repo case. Your mega-repo reasoning is in the body as the reason that gap is acceptable rather than merely unmeasured.

**5. Option-2 argument corrected.** "A new path-aware mechanism in three packages" was overstated and is now "one `IsHardcodedSkipPath(rel)` plus a change at each of four sites, plus a genuine file-level rule" — with your point that all four call sites already hold the path. The 15–30% is labelled as the *entity* figure, and the file-count denominator is stated: **90 of 164 (55%)** under `workflows/` + `actions/`.

**6. `.gitlab`.** Now stated in its own paragraph: zero `.gitlab/` directories in the corpus, so the blast radius covers `.github` plus three `.circleci` files, and the `.gitlab` un-skip rests on the argument plus the synthetic walker test — not on the measurement.

**7. `copilot-instructions.md`** is named in the body, as a file `internal/install/rulesfiles` generates as tool config.

**Follow-ups filed, not implied:** #6955 (cicd golden fixture — the ratchet and the `cmd/grafel` digest pin are both blind for the reason you confirmed) and #6954 (the `ISSUE_TEMPLATE` + loose-`.md` trim, with your 26 + 22 = 29%-in-two-lines framing and the `copilot-instructions.md` case). M3's specificity gap is closed here, so it needed no follow-up.

Verification: `go test -count=1` green on `internal/daemon/walk`, `internal/daemon/watch`, `internal/daemon/sched`, `internal/engine`, `internal/quality`, `internal/extractors` and `./cmd/grafel` (digest pin unmoved); `gofmt -l` and `go vet` clean. Not merged.

cajasmota:
Coordinator mutant on the final head `78810cebf`. One mutant, **DEAD by four tests** — and one near-miss of my own worth recording.

## MK-1 — the skip list's REMAINING membership

This PR removes three entries from `hardcodedSkipDirs`. Every mutant so far — your M1/M2/M3/M6 and the review's M3–M6 — attacked the three names, the globs, the ignore-stack interaction, or the parity copies. Nobody asked the obvious neighbouring question: **is the rest of that map pinned, or is it only the three you touched?**

So I made the same shape of edit one entry further down: delete `"node_modules"`.

**Verdict: DEAD**, by four tests — `TestWalkRepo_HardcodedSkip`, `TestIsHardcodedSkip`, `TestWalkRepo_SkipsMediaBundleAndNodeModules`, and your own new `TestWalkRepo_CIDirsAreWalked_6946`, whose negative controls catch it. The list's membership is genuinely graded, not just its recent edit. Baseline green first (walk 7.682s, watch 76.911s), `go vet` clean, `-count=1`, restored by `cp` from a shasum-verified backup (`8047347493…`), `git status --porcelain` empty.

## The near-miss — my first edit silently did not land

My first anchor was `^\t"node_modules": {},` — one space before the brace. The map is column-aligned, so the real line is `"node_modules":  {},` with **two**. `perl` matched nothing, the file was byte-identical to the backup, and the suite came back **green**.

Had I not asserted the pre-edit occurrence count, I would have reported **MK-1 ALIVE — the skip list's membership is ungraded**, and sent you a fabricated finding with a confident diagnosis attached to it. The whole edifice of the report would have been wrong in the *permissive* direction: an unapplied edit always reads as a survivor.

That is why the rule is "prove the edit landed with an exact occurrence count", and it is worth restating with a real instance rather than as advice. The check that saved it was `pre=1` failing as `pre=0` — a cheap assertion, and the only thing standing between a green run and a false report.

## On the round

All five asks landed, and two beyond what I asked:

- **The reviewer named a third parity copy and you pinned that too** (`cmd/grafel`'s `shouldSkipDirForStale`), rather than fixing only the two under discussion. Three copies of a skip list is the underlying defect; pinning all three is what makes the "no divergence" claim mean something. The `.claude`/`node_modules`/`.idea` negative controls are what stop the parity test passing against a watcher that skips nothing — the same guard-on-the-guard shape as #6950's positive control.
- **M3's fix grades specificity, not just reachability.** Requiring every declared glob to *reject* decoy YAML at paths GitHub never reads (`config/app.yml`, `deploy/values.yaml`, and `src/actions/setup/action.yml` aimed squarely at the `actions/*/action.yml` glob) is the right generalisation of a mutant that exploited the basename fallback.

The rewordings are all correct, and leading with **0.16% of source bytes** instead of the n=1 RSS pair is the right call — it is a structural argument that does not depend on a measurement whose floor is above the effect. Naming `copilot-instructions.md`, giving `.gitlab` its own "measured on nothing" paragraph, and filing **#6954** and **#6955** rather than leaving them implied are exactly what I wanted.

Merging on green. Refs #6946, #6927, #6934.

Issue #6952, opened 2026-09-07
conf/routes is never classified, so every Play (and Revel) routes producer in the tree is unreachable — three of them, in two packages


cajasmota:
Same shape as #6946, in a second language. `conf/routes` — the file Play Framework's own detection block calls "highly reliable — that path is unique to Play" — is never classified, so nothing that reads it ever runs.

## The gate

`internal/classifier/classifier.go`'s `detectLanguage` resolves, in order: three compound-suffix special cases, a `.json` routing block, `exactBasenameLanguageMap`, `extensionLanguageMap` keyed on `filepath.Ext`, `basenameLanguageMap`, then `containerVariantLanguage`. A bare `routes` has no extension and appears in neither basename map, so `detectLanguage` returns `""` and both `classifyInner` (:154) and `classifyWithSizeInner` (:190) return `Skip{unsupported_extension}`.

Measured, not read:

```
conf/routes                            lang=""       skip=true  reason=unsupported_extension
modules/admin/conf/routes              lang=""       skip=true  reason=unsupported_extension
conf/routes.conf                       lang=""       skip=true  reason=unsupported_extension
app/controllers/HomeController.scala   lang="scala"  skip=false
```

Pinned by `TestIssue6927_Play_ConfRoutesIsNotClassified` (`internal/engine/play_graphql_multiline_anchor_6927_test.go`), which carries the `.scala` row as a positive control so the skip is a fact about the filename and not about a malformed path.

## What is unreachable behind it

Three producers, in two packages, all keyed on a path that never arrives:

- `internal/engine/rules/scala/frameworks/play_framework.yaml` — the `Route` source_pattern and the `ROUTES_TO` relationship_rule. #6927 just fixed both (they were compiled without `(?m)`, so they found 4 of 135 real route lines); the fix is correct and changes nothing in production.
- `internal/custom/scala/frameworks.go:644` — `rePlayRoute`, which emits `SCOPE.Operation "GET:/path"` with `http_method`, `http_path` and `controller` properties and `provenance: PLAY_ROUTES_FILE`.
- `internal/custom/java/play_routes.go` — `isPlayRoutesFile` matches `routes`, `routes.*`, `*.routes` and any path containing `conf/routes`, and `extractPlayRoutes` parses the DSL into Route + Handler entities. `internal/custom/golang/revel.go:59` is the same shape for Revel, whose routes file is also `conf/routes`.

[...]

PR #6953, opened 2026-09-07
fix(#6927): play's routes rules needed `(?m)` and found 4 of 135 route lines without it — and graphql's field rule was measured away instead, because widening it silently repointed two SERVES edges

cajasmota:
**Refs #6927 — deliberately not `Closes`.** The 18 patterns it enumerated are done, but only **one of the four arms is live in production**: this PR's graphql half. The play half changes nothing (`conf/routes` → `Skip{unsupported_extension}`, #6952) and the GitHub Actions arm changes nothing (`.github` is on `walk.hardcodedSkipDirs`, #6946; a fix is in flight as #6951 but not merged). An issue closed while two of its arms are latent tells the next reader the work is live. #6946 and #6952 are what it should close against.

The last 3 of the 18 patterns: `scala/frameworks/play_framework.yaml` (2) and `graphql/frameworks/graphql_schema.yaml` (1). **The two files got opposite treatments**, and the reason is the measurement rather than the pattern text.

## Per pattern: matches today, ungated `(?m)`, and shipped

Corpus: `~/Projects/archigraph-corpora` has 20 `.scala` files, 1 `conf/routes` and **zero** `.graphql`, so — as #6949 did for ansible — real upstream repos were cloned: `playframework/play-samples`, `http4s/http4s`, `scalatra/scalatra`, `facebook/relay`, `99designs/gqlgen`, `graphql/graphiql`, `graphql/graphql-js`, `apollographql/apollo-client`.

| pattern | today | `(?m)` ungated | shipped |
|---|---|---|---|
| play `^(?:GET\|…)\s+(\S+)\s+controllers\.` (source_pattern) | **4 of 135** route lines, 4 of 33 files | 135 of 135, 32 of 33 files. **0 of 1052** `.scala`/`.sc` files | `(?m)`, **ungated** |
| play, same regex as relationship_rule (a separate `regexp.Compile` at `detector.go:177`) | same | same | `(?m)`, ungated |
| graphql `^\s+(\w+)\s*\([^)]*\)\s*:` → `Operation` | **0 of 1907** `.graphql`/`.gql` files | 1429 in 160 files | **REMOVED** |

**Play is genuinely framework-specific and takes no gate**, and that is measured, not assumed: a column-0 HTTP verb followed by a literal `controllers.` is Play's routes grammar and is not Scala syntax, and it matches zero times in 1052 real Scala files in *both* modes. `requires_framework` would in fact be worse than useless here — `play_framework.yaml` declares no `import_markers`, so `detector.go:161` would log it and `frameworkPresent` would return false for every file, turning the gate into an off switch. `TestIssue6927_Play_RouteRuleIsNotFrameworkGated` records that.

**GraphQL did not over-fire across file kinds** — the thing three arms in a row found, and the thing this one was sent to look for. All 160 matching files are SDL documents (classified by whether they carry a `type`/`input`/`interface`/`enum`/`union`/`scalar`/`extend`/`schema`/`directive` definition at all); **zero** matches in a pure operation document, across relay's 1737 `.graphql` fixtures. The reason it is still wrong is narrower and worse — see below.

## `$`, explicitly

Neither pattern carries a `$`, so `(?m)`'s other half cannot reach either of them. That is asserted rather than noted: `TestIssue6927_Play_BothRouteRulesAreMultiline` fails if a `$` appears in either rule, because a new one would need the measurement #6949's docker rewrite needed.

## The graphql finding: recall could not see it

`internal/extractors/graphql` emits the fields of a type **definition** as `SCOPE.Component|Owner.field` with a `CONTAINS` edge from the owner (`graphql.go:236-270`). This rule captured a subset of those — only fields carrying an argument list — under a different kind and a **bare** name, dropping the owner.

**Correction from review, and it is load-bearing.** That subsumption does **not** hold everywhere: the extractor's `extend type Foo { … }` branch (`graphql.go:350`) emits an IMPORTS/FEDERATES stub and **no field entities at all** — which is how every Apollo Federation subgraph declares its fields, and both `gqlgen` and `apollo-client` are in the measured corpus. So "the extractor already covers it" is **not** why the rule is gone, and an earlier version of this PR (and of the test's failure message) said it was. It is gone because at HEAD it produces nothing, and widened it does active harm:

Measured by running the pipeline on `graphql-schema-mini` with `(?m)` added:

```
entities:      8 / 8 (100.0%)     forbidden hits: 0
relationships: 6 / 6 (100.0%)     forbidden entity hits: 0
```

Nothing moved — and two `SERVES` edges silently repointed. **The mechanism**, which review supplied and which belongs here: `resolve/refs.go`'s `isOperationKind` is literally `k == "SCOPE.Operation" || k == "Operation"`, so `Operation` is not an inert label — it makes the entity a first-class candidate when the resolver binds an operation reference by **bare** name. The http_endpoint synthesis emits `SCOPE.Operation "QUERY /node"` carrying a bare `handler_ref: node`, so the new bare `Operation|node` outranked `SCOPE.Component|Query.node` and the edge landed on it. `QUERY /user` did the same. `QUERY /posts` and `QUERY /settings` were untouched, because their fields take no arguments — which is what makes the two moved edges attributable to this rule and not to the schema.

So the rule was removed, not widened. Removal is behaviour-preserving on every input measured (0 matches in 1907 files).

**What is NOT closed off**, said explicitly in the rule comment, the fixture prose and the test's failure message, because the earlier framing foreclosed it on a false premise: widening the rule and **owner-qualifying** the capture — emitting `Owner.field`, or any kind `isOperationKind` does not accept — would reach 1429 sites across 160 SDL files, *including* the `extend type` fields the extractor does not model. Only a **bare-named `Operation`** is banned.

## Forbidden rows, and proof each can fire

**graphql — 2 forbidden entities + 3 `SERVES` rows** in `graphql-schema-mini/expected.json`. Fired by re-adding the pattern with `(?m)`:

```
forbidden entity hits: 2
  - node [Operation] in schema.graphql
  - user [Operation] in schema.graphql
missing relationships:
  - QUERY /node --[SERVES]--> Query.node  (both endpoints exist; edge not emitted)
  - QUERY /user --[SERVES]--> Query.user  (both endpoints exist; edge not emitted)
```

`QUERY /posts --SERVES--> Query.posts` is the **control expected to pass in both worlds**; a failure there would mean something other than this rule moved the resolution.

**play — 4 forbidden rows**, one per axis, each a real construct of the routes format and each fired by exactly the mutant its `why` names:

| forbidden row | shape | mutant that mints it | rows that actually fired |
|---|---|---|---|
| `/legacy` | `# GET /legacy …` — a commented-out route | delete `^` | `/indented`, `/legacy` |
| `/indented` | indented verb line (Play's parser rejects it) | `(?m)^` → `(?m)^[ \t]*` | `/indented` |
| `/trace` | `TRACE` — a verb outside the alternation | add `TRACE` to the alternation | `/trace` |
| `/health` | `services.HealthCheck.ping` — a non-controller target | `controllers\.` → `\w+\.` | `/health` |

The right-hand column is a real run, not a claim: each mutant was applied to the shipped pattern and the forbidden rows it minted were collected.

Varied on the **anchor** axis: column-0 / indented / commented-out / mid-line, name held constant (`controllers.<Class>.<method>` throughout). Varied on the **name-and-content** axis: verb inside vs outside the alternation, target package `controllers` vs `services`, anchor held constant at column 0. #6943's lesson is that a fixture pinning one pins nothing about the other, so the two are separate cases with separate mutants — and M2 (delete `^`) firing *both* anchor rows while touching neither name row is the evidence the split is real.

Where the rule's specificity lives in a **list**, an **exact set** is asserted, not membership: `TestIssue6927_Play_VerbAlternationIsExactlyTheSevenHTTPVerbs` parses the alternation out of both compile sites and compares it to `{DELETE,GET,HEAD,OPTIONS,PATCH,POST,PUT}`. Adding `TRACE` is caught there *and* by the `/trace` row; a same-count substitution — the mutant that survived a growth ceiling and a membership assertion in both #6945 and #6949 — is caught only there.

Rules are selected **structurally** in every test: the Route rule by `EntityType == "Route"`, the edge rule by `Relationship == "ROUTES_TO"`, the graphql rule by the kind it emits — never by text substring, which is what went red on a rename in #6945. `TestIssue6927_Play_ControllerRuleStillMatches` is the control expected to pass: it drives an untouched Play rule on a `.scala` file, so a forbidden-row result elsewhere cannot be an artefact of the harness producing nothing.

## Mutant verdicts, both levels

Play has **no fixture level available** — see the reachability finding — so its seven mutants are scored at the engine suite only, and each is named in the rule's own comment the way #6945's job rule does.

|---|---|---|---|
| M1 | `(?m)^` → `(?m)^[ \t]*` | DEAD | `RoutesFileYieldsEveryRoute`, `…RoutesToEdge`, `RouteRuleForbiddenShapes` |
| M2 | `(?m)^` → `(?m)` (anchor deleted) | DEAD | the same three + `BothRouteRulesAreMultiline` |
| M3 | alternation + `TRACE` | DEAD | the same three + `VerbAlternationIsExactlyTheSevenHTTPVerbs` |
| M4 | `controllers\.` → `\w+\.` | DEAD | the same three + the alternation test |
| M5 | `(?m)` dropped from BOTH sites | DEAD | both recall tests + `BothRouteRulesAreMultiline` |
| M6 | `(?m)` dropped from the **source_pattern only** | DEAD | `RoutesFileYieldsEveryRoute` + `BothRouteRulesAreMultiline` — **not** the edge test |
| M7 | `(?m)` dropped from the **relationship_rule only** | DEAD | `…RoutesToEdge` + `BothRouteRulesAreMultiline` — **not** the entity test |

M6/M7 are the point of splitting them: the two compile sites are separate `regexp.Compile` calls and each is graded by a test the other does not fail. Review's MU6 confirms the `$` claim from the other side: appending `.*$` to one of the two sites dies on `BothRouteRulesAreMultiline`.

**MU2 is ALIVE and is recorded rather than fixed.** Widening the capture `controllers\.(\w+\.\w+)` → `(\w+(?:\.\w+)?)` at both sites survives every level, because over the 32 real `conf/routes` files that carry a controller route the two produce **identical match sets** — every real line names `Controller.method`, so the optional second segment is never exercised. It is equivalent on measured input, so there is no defect to demonstrate; what it establishes is that the **capture axis is ungraded**, distinct from the anchor and name axes that are. Named in the rule comment so a later change to the capture is known to be uncovered rather than assumed covered.

| # | mutant | engine suite | ratchet |
|---|---|---|---|
| G1 | re-add the graphql pattern **with** `(?m)` | DEAD — `NoOperationSourcePattern`, `SchemaMintsNoBareFieldEntity` | DEAD — 2 forbidden entity rows, `relationship_found` 9→7, and both extracted-total ceilings |
| G2 | re-add it **without** `(?m)` (the pre-#6927 text) | DEAD — `NoOperationSourcePattern`, `LedgerOfAnchoredPatternsWithoutMultiline` | not run (G1 covers the graph-level harm; G2 is a no-op at the graph level *by construction* — 0 matches — which is the ledger test's whole job) |
| MU1 (review) | re-add the same regex with `entity_type: Component` | DEAD — but on `SchemaMintsNoBareFieldEntity` **only**; `NoOperationSourcePattern` is kind-scoped and passes | — |

MU1 is worth stating plainly: the **structural selector is evadable**, and the entity-level test is what actually holds the line. `NoOperationSourcePattern` names the kind and the reason (`isOperationKind`); it is not the fence. Its comment now says so.

Baseline verified green immediately before each score, and restored and re-verified green after.

## Ratchet delta, enumerated

`scripts/quality/run.sh --ratchet` before recording anything reported **exactly two lines, both mine, both `graphql-schema-mini`**:

```
- graphql-schema-mini: relationship_found IMPROVED 6 -> 9 (of 9 must-have)
- graphql-schema-mini: relationship_expected changed 6 -> 9
```

Both are the three `SERVES` rows added above. **No pre-existing drift appeared, so #6898's downward-drift mechanism was not exercised here** — that is a statement about this run, not about drift being absent. The baseline row was **hand-edited** (two integers), not re-recorded with `--update-baseline`, so nothing else could be silently re-recorded. Re-run: `quality ratchet OK — 37 gated fixtures held their baseline (28 at 100% must-have recall)`.

The corpus-size guards were re-derived from disk rather than trusted: `ls -d internal/quality/golden/*/` is **37**, `baseline.json` carries 37 fixtures, and the three guards (`absent_relationships_6490_test.go:232`, `zero_relationships_6490_test.go:180`, `subtype_assertion_6488_test.go:291`) all read 37. This arm adds no fixture, so none moved.

## The finding that dwarfs the play fix — #6952

**`conf/routes` is never classified, so both Play rules are unreachable in production**, and the fix does not change that. It has no extension and no entry in `classifier.go`'s `exactBasenameLanguageMap` or `basenameLanguageMap`, so `detectLanguage` returns `""` and both `classifyInner` and `classifyWithSizeInner` return `Skip{unsupported_extension}`. Same shape as #6946's `.github` finding, in a second language.

Three producers wait behind it, in two packages: these rules, `internal/custom/scala/frameworks.go:644`'s `rePlayRoute` (which already carries `(?m)` and emits a **richer** `SCOPE.Operation "GET:/path"` with method, path and controller), and `internal/custom/java/play_routes.go` (plus `internal/custom/golang/revel.go` for Revel, whose routes file is also `conf/routes`). All have passing unit tests, because those construct the file record directly and never go through the classifier.

`TestIssue6927_Play_ConfRoutesIsNotClassified` pins it — with a `.scala` **positive control**, so the skip is a fact about the filename and not about a broken classifier or a malformed path. It goes red when #6952 is fixed, which is the point: whoever fixes it has to decide which producer owns the Route entity, because `rePlayRoute` is strictly richer *except* that it emits no edge, and the `ROUTES_TO` edge is the only reason these rules were fixed rather than deleted like the graphql one.

## The ledger does not reach zero, and now says so

`TestIssue6927_LedgerOfAnchoredPatternsWithoutMultiline` decodes every compiled rule set — 290 sets, 385 patterns, the same population `detector.go` compiles — and lists every pattern carrying a **real** `^`/`$` anchor without `(?m)`, stepping over escapes and character classes. That distinction matters: a naive scan counts `[^)]` and `\$` and reports **134** offenders where there are **2**.

The 18 this issue enumerated are done. Two survivors remain, both predating the enumeration and both recorded on it as unmeasured, and they are an **allow-list, not a ceiling** — a third one fails, and so does fixing one of these without deleting its entry:

- `csharp/.NET MAUI` — `(?::|$)`. `$`-only; #6927 itself flagged it as "probably a recall gain, unmeasured".
- `ansible/Ansible Core` — `^-\s+name:`. A third anchored pattern in a file the issue tabulated as carrying two. `ansible_core.yaml` also carries an un-anchored `-\s+name:` → Task rule, so the plays this one misses are already partly in the graph under a different kind.

The test carries floors (250 sets / 350 patterns) so a loader returning nothing reads as a failure rather than as a clean ledger.

**Two holes review found in it, both now closed and graded.** The sweep cannot grade its own detection — both survivors happen to carry `^` at index 0 and neither carries a `(?m)` — so `TestIssue6927_LedgerHelperDetectsAnchorsAndFlagPlacement` grades the helper directly, over an enumerated table of shapes rather than the two live offenders:

- **MU3 was ALIVE**: reducing anchor detection to `strings.HasPrefix(p, "^")` left the sweep green. The table carries `foo$`, `(?:a|b)^x` and `x(?::|$)` — anchors that are not at index 0 — alongside the non-anchors `[^)]`, `\$`, `\^` and `["$]`.
- **MU5 mattered more**: the flag check was `strings.Contains(p, "(?m)")`, which is **placement-blind**. Go applies a flag group from its own position onward, so `^foo(?m)` is still start-of-TEXT and carried the exact defect #6927 is about — and it satisfied the ledger. The check now compares the flag's index against the **first** anchor's, which is the conservative reading (`^a(?m)b$` is an offender: the later anchor being multiline does not rescue the first).

Both are DEAD against the new test. The helper now returns an index rather than a bool, which is what the placement comparison needs.

## Verification

- `go test ./internal/engine/ -count=1` — ok (49s)
- `go test ./internal/quality/... -count=1` — ok, all four packages
- `go test ./cmd/grafel/ -count=1` — ok, which is the digest-pin over the whole graph and mandatory for any rule change. **Flagged rather than smoothed over:** the first run of the post-review round exited non-zero, and I lost the cause by tailing its output to three lines. Three consecutive full re-runs are green (exit 0, no `--- FAIL`), and the pre-review round was green, so I cannot attribute it. It is recorded here rather than dropped; CI runs this leg too.
- `go test ./internal/extractors/graphql/ ./internal/classifier/ ./internal/custom/scala/ -count=1` — ok
- `go vet ./internal/engine/ ./internal/quality/... ./cmd/grafel/` — clean; `gofmt -l` — clean
- `scripts/quality/run.sh --ratchet` — exit 0, 37 fixtures
- `git diff HEAD~1` read before pushing, because a `cp` restore proves the backup and not HEAD (#6949)


cajasmota:

Measured in my own worktree at `scratchpad/review-6953-adv9k/wt`, detached at `0a3099b19`, isolated
`HOME`/`GRAFEL_HOME`/`GRAFEL_DAEMON_ROOT`. Tree verified clean (`git status --porcelain` empty,
`git diff --stat HEAD` empty) before and after every score; `go vet ./internal/engine/` clean;
`go test ./internal/engine/ -count=1` green as the baseline before each mutant and again at the end.

The change itself is correct and safe. What I am asking to change is a **load-bearing claim that is
false**, a permanent guard that rests on it, and a hole in the standing ledger that is being offered
as the closing statement for #6927.

---

### 1. The repointing reproduces exactly. Independently confirmed.

Re-added the pattern with `(?m)`, rebuilt `cmd/grafel`, ran `grafel quality` on the fixture:

```
                      HEAD                        rule re-added with (?m)
entities         8 / 8  (100.0%)  total 34        8 / 8  (100.0%)  total 36
relationships    9 / 9  (100.0%)  total 82        7 / 9  ( 77.8%)  total 84
forbidden hits   0                                0
forbidden ents   0                                2   node [Operation], user [Operation]
missing:                                          QUERY /node --[SERVES]--> Query.node
                                                  QUERY /user --[SERVES]--> Query.user
```

`QUERY /posts` held in both worlds. This is the best part of the PR: entity recall, forbidden-edge
hits and forbidden-entity hits at the *old* expected.json would all have stayed put while two edges
moved, and the new rows are what make it visible.

### 2. Every corpus figure re-derived from the clones. All correct.

| claim | PR | mine |
|---|---|---|
| `.graphql`/`.gql` files | 1907 | **1907** |
| matches today (start-of-text) | 0 | **0** |
| matches under `(?m)` | 1429 in 160 files | **1429 in 160 files** |
| of those, SDL / operation documents | 160 / 0 | **160 / 0** |
| `.scala`/`.sc` files, matches both modes | 1052, 0 | **1032 in the clones + 20 in archigraph-corpora = 1052, 0 / 0** |
| `conf/routes` files | 33 | **32 in the clones + 1 in archigraph-corpora = 33** |
| route lines, start-of-text vs `(?m)` | 4 of 135 | **4 of 131 in the clones + the corpora file = 4 of 135, in 4 files** |

Say plainly on the issue *what the zero is over*: `~/Projects/archigraph-corpora` carries **20**
`.scala` files and **one** `conf/routes`. The 1052/1907 figures live entirely in transient clones
under a scratchpad. They happen to still be on this machine, which is the only reason I could check
them; the next reader cannot.

### 3. The subset claim is FALSE — `extend type` fields

This is the one that matters, because "strict subset of what the extractor already emits" is the
*entire* stated reason the rule was deleted rather than widened, and it is asserted in four places:
the YAML comment, `expected.json`'s `second_subject_6927`, the PR body, and the failure message of
`TestIssue6927_GraphQL_NoOperationSourcePattern`.

`internal/extractors/graphql` emits per-field `SCOPE.Component` entities only from the `typeDefRE`
branch, gated on `subtype == "type" || "interface" || "input"` (graphql.go:218). The **`extend type`
branch (graphql.go:350)** emits an IMPORTS stub and optional FEDERATES edges and **no field
entities and no CONTAINS edges at all**. Probed directly through `Extractor.Extract`:

```
type Query { posts: [Post!]! }
extend type Query @key(fields: "id") {
  search(term: String!): [Post!]!
  byAuthor(id: ID!): [Post!]!
}

extractor:  SCOPE.Component/field "Query.posts"   SCOPE.Schema "Query"   SCOPE.Component/import "Query"
            (no Query.search, no Query.byAuthor)
deleted rule under (?m) would mint:  Operation|search, Operation|byAuthor
```

`extend type Query { … }` is how an Apollo Federation subgraph declares its query fields — and
`gqlgen` and `apollo-client` are two of the eight repos in the very corpus this arm measured. So the
widened rule was **not** a strict subset; on federation subgraphs it was the only producer.

This does **not** make the deletion a regression. At HEAD the rule matched 0 of 1907 files, so
removing it is behaviour-preserving on every measured input, and the *harm* the PR demonstrates
(bare name outranking the owner-qualified field) is real and reproduced above. But two things follow
that the PR does not say:

- The correct alternative was never priced. Widen **and** owner-qualify (or emit `SCOPE.Component`
  `Owner.field` from the extend branch) is a recall gain over 160 SDL files / 1429 sites that the
  "subset" framing declared impossible without measuring it.
- `TestIssue6927_GraphQL_NoOperationSourcePattern` forbids **any** `Operation` source_pattern in that
  file **forever**, and its failure text tells the next reader the rule is a duplicate. That is an
  over-broad guard resting on a false premise. Keep the test — it is the right shape — but fix its
  message and the YAML/expected.json prose to say *subset except inside `extend type` blocks, which
  the extractor does not cover*, and file the federation gap.

### 4. `Operation` vs `SCOPE.Component` IS a kind difference, and it is the mechanism

`internal/types/kinds.go:610` records `EntityKindOperationBare`, and `internal/resolve/refs.go`'s
`isOperationKind` is literally `k == "SCOPE.Operation" || k == "Operation"`. So the deleted rule's
entity was, to the resolver, an **operation** — the same class as the `QUERY /node` endpoint's
`handler_ref` target — while `Query.node` is a component. That is *why* the bare candidate won. This
supports the deletion and independently refutes the subset claim on the kind axis too; it is worth a
sentence in the comment, because "bare name" alone under-describes the mechanism.

### 5. The ungated Play verdict HOLDS

Verified in code, not assumed. `compiledRuleSet.frameworkPresent` (detector.go:96) returns false when
`importMarkers` is empty, and the match loop skips `sp.requiresFramework && !frameworkSeen`
(detector.go:409). `play_framework.yaml`'s `frameworks.detection` declares `config_files`,
`build_signals`, `import_patterns`, `code_patterns` — and **no `import_markers`**, which is the only
key the engine reads. So `requires_framework: true` would be an exact off switch. Confirmed.

The 0/1052 measurement reproduces, and the specificity argument (column-0 verb + literal
`controllers.`) is sound. `TestIssue6927_Play_RouteRuleIsNotFrameworkGated` is a good record of it.

### 6. Claims 1–4 spot-checked

- **`$` assertion fires.** Appended `.*$` to the source_pattern site only (1 of 2 occurrences):
  `TestIssue6927_Play_BothRouteRulesAreMultiline` FAILs. Asserted, not noted — confirmed.
- **Two compile sites are independently graded.** `frameworkPresent`/compile inspection confirms
  detector.go:156 and :177 are separate `regexp.Compile` calls, and MU6 above fired on a
  single-site edit. M6/M7's shape is real.
- **Forbidden row sampled.** Relaxed `\s+controllers\.` → `\s+\w+\.` at both sites: exactly
  `Route "/health"` was minted and no other forbidden row. Fires by its named mutant, only it.
- **Ratchet delta.** Not re-run in full (37 fixtures × 5 runs), but bounded from disk instead:
  `ls -d internal/quality/golden/*/` is 37, and the only fixtures containing a `.graphql`/`.gql`/
  `.scala`/`routes` file are `graphql-schema-mini` and `scala-play-mini`. `scala-play-mini` graded
  23/23, 5/10, totals 81/201 — **byte-identical to its baseline row**, and it carries no routes file,
  so the Play change cannot move any fixture. `graphql-schema-mini` graded 9/9 as recorded. The
  two-line delta is exactly right.
- **No pre-existing drift.** Consistent with the above; the two fixtures this arm can touch are the
  only ones that moved, and the python drifts seen in prior arms are fixture-level facts about other
  fixtures I did not re-run. The wording ("a statement about this run") is accurate as written.

### 7. The ledger does not close #6927 as tightly as claimed

The sweep and the count are right — I re-ran it and the two survivors are `.NET MAUI`'s `(?::|$)`
and ansible's `^-\s+name:`, both carrying their reasons, and a third violation does fail. But the
guard has a hole in the direction that matters:

**`strings.Contains(p, "(?m)")` is placement-blind.** In Go, an inline flag group applies from that
point onward, so `^foo(?m)` still means start-of-**text** — and reads as clean to this ledger.

> **MU5** — moved `(?m)` to the end of pytest's rule: `"^def\\s+(test_\\w+)\\s*\\((?m)"`.
> `TestIssue6927_LedgerOfAnchoredPatternsWithoutMultiline` **PASSED**. The mutant died only because
> the pytest arm shipped its own tests (`TestIssue6927_RealisticPytestModuleYieldsTests`,
> `TestIssue6927_PythonPatternsDoNotOverFire`). For any of the 385 patterns without an arm test, this
> is alive and silent.

Fix is small: require the flag group to appear *before* the first real anchor (the helper already
walks the string; return the anchor's index and compare).

---

### Mutant table (mine, not the author's 7+2)

| # | mutant | verdict |
|---|---|---|
| **MU1** | re-add the graphql field rule with `(?m)` but `entity_type: **Component**` (evades the kind-scoped selector) | **DEAD** — `SchemaMintsNoBareFieldEntity` (name-based, kind-independent). Note `NoOperationSourcePattern` did **not** fire, so the structurally-selected test alone is evadable; the pair is what covers it. |
| **MU2** | play `controllers\.(\w+\.\w+)` → `controllers\.(\w+(?:\.\w+)?)` at both sites (permissive target shape) | **ALIVE** — whole engine suite green. Distinguishable only on a synthetic `GET /x controllers.Assets`; over the 32 real `conf/routes` files the match sets are **identical**, so equivalent on measured input. Fixture gap on the capture axis, not a defect: the comment's "every character before the capture is graded" is true, and the capture itself is not. |
| **MU3** | ledger helper: count `^` as an anchor **only at index 0** | **ALIVE — distinguishable and ungraded.** Both allow-listed survivors carry `^` at index 0, so nothing can tell. A future `(?:^\|\n)foo` or `a\|^b` slips the ledger. |
| **MU4** | ledger helper: delete the `case c == '\\': i++` escape stepping | **DEAD** — 2 new offenders reported, ledger FAILs. The over-count direction *is* graded. |
| **MU5** | move `(?m)` to the END of an anchored pattern (pytest) | **DEAD only by another arm's tests; the ledger PASSED it.** See §7. |
| **MU6** | append `.*$` to the play source_pattern (1 of 2 sites) | **DEAD** — `BothRouteRulesAreMultiline`. Confirms both the `$` assertion and per-site grading. |

---

### Verdict on the deletion

**Right change, wrong reason.** Removing the rule loses nothing on any measured input and removes a
real landmine — the repointing is reproduced and the `Operation` kind is genuinely the wrong kind.
But it is **not** a strict subset: `extend type` fields have no producer at all, and the PR forecloses
the widen-and-qualify option with a permanent test whose message states the false premise.

### Should #6927 close?

**Not as written.** The 18 enumerated patterns are done and that is real. But the closing statement
should say, on the issue:

1. The play arm and #6946's GitHub Actions arm **changed nothing in production** — `conf/routes` is
   `Skip{unsupported_extension}` (#6952, whose premise I accept: `detectLanguage` has no basename
   entry, and the PR's own pinning test passes) and `.github` is on `hardcodedSkipDirs`. Three
   producers in two packages sit behind #6952. Closing #6927 without that sentence reads as if the
   graph gained routes; only the graphql arm is live today.
2. The ledger **does not reach zero** — 2 survivors — and is currently bypassable by flag placement
   (§7). Fix that before it is cited as the thing that keeps the count at 2.
3. #6952 and #6946 stay open and are the reason two of the six arms are latent.

Not merging, not pushing.

cajasmota:
Coordinator mutants on the final head `ffc676963`. Two, **both DEAD**, and they establish something about the ledger that is worth writing down: it is defended by three different mechanisms, and each covers a case the others miss.

I went at the ledger's **population** rather than its helpers. The helpers are now graded directly, and the review's MU3/MU5 closed the two holes in them — but a scan-and-assert-absence guard has more ways to be a no-op than its predicate, and #6949's vacuous-walk shape is the one this board keeps meeting.

## ML-1 — narrow the sweep by ONE language that has a survivor

`if lang == "ansible" { continue }` inside the sweep loop. Well above the floor, so the floor cannot see it.

**DEAD** — and not by the floor:

```
expected residual offender ansible/Ansible Core/source_pattern to still read
"^-\\s+name:\\s+[\"']?([^\"'\\n]+)", got "" — if it was fixed, delete its entry here
```

**The allow-list is doubling as a positive control on the sweep's population.** Each `want` entry asserts not just "this offender is still here" but "the language it lives in is still being swept". That is a second job the map is doing for free, and it is why an entry must never be deleted casually — deleting one removes a population pin as well as a record.

## ML-2 — narrow by three languages that have NO survivor

`ruby`, `python` and `cicd` — chosen precisely because no `want` entry lives in them, so the allow-list cannot notice.

**DEAD**, by the floor: `swept 250 rule sets / 277 patterns — too few to be the real population`.

## What the pair establishes

The two mechanisms cover **disjoint** failure modes:

- the **allow-list** catches a narrowing that drops a language *with* a survivor, at any size;
- the **floor** catches a narrowing large enough to move the totals, in any language.

The gap between them is a narrowing that is both small **and** confined to languages with no survivor — one or two rule sets in `ruby` alone would have slipped both. That gap is real but shallow, and I am not asking for anything: the floors at 250/350 sit close enough to the true population (250 sets / 277 patterns after dropping three languages, against the live figures) that the room to hide is small. Worth a sentence in the test comment naming the two mechanisms as covering different cases, so the next person to relax a floor knows what they are relaxing.

## On the round

All four items landed, and the first one properly: the false "strict subset" premise is corrected in **all four places it had leaked** — the test's failure message, the rule comment, the fixture prose and the PR body — with the true mechanism (`isOperationKind` at `refs.go:1939` matching both `SCOPE.Operation` and `Operation`, which is *why* the bare candidate outranks the field) put in its place. Stating explicitly that widen-and-owner-qualify remains open, and that only a bare-named `Operation` is closed, is what stops a correct deletion from foreclosing a correct future change on a wrong basis.

Recording MU1's lesson in the test's own comment — the kind-scoped selector is evadable, the entity-level test is the fence — is the right place for it. And `Refs #6927` rather than `Closes` is correct: counting patterns is not counting fixes that reach a user, and with #6952 and #6946 open, graphql is the only live arm.

**On the lost `cmd/grafel` failure:** flagging it rather than smoothing it over was right, and the lesson is the one this repo keeps paying for — a failure tailed to three lines is a failure discarded. Redirect to a file and keep it; `rtk`'s truncation is the same trap with a different cause. Three green re-runs plus a green pre-review round is not attribution, and you did not claim it was.

Merging on green. Refs #6927, #6952, #6946.

Issue #6955, opened 2026-09-07
No golden fixture contains a .github, so the ratchet is blind to the #6946 walker change — the whole GitHub Actions rule set is graded by two unit tests and nothing else


cajasmota:
Split from the #6946 review.

#6946 un-skipped `.github` / `.gitlab` / `.circleci` at the walker, which is a change to what
**every** repo grafel indexes. It is protected end-to-end by nothing: the extraction ratchet cannot
see it.

Confirmed independently, twice:

* `find internal/quality -type d \( -name .github -o -name .gitlab -o -name .circleci \)` → **0**.
  No golden fixture contains a CI directory, so the ratchet's entity/relationship counts are
  invariant under the change.
* `cmd/grafel`'s graph digest pin is blind for the same reason — its fixtures are synthesised into
  `t.TempDir()` and none writes a `.github`.

So the whole GitHub Actions rule set — eleven `source_patterns`, three `file_conventions` globs, the
structured workflow extractor in `internal/extractors/yaml` — is graded today by two unit tests
(walker reachability + glob reachability/specificity, both added in #6951) and by nothing that
measures what it actually extracts.

## Ask

A `cicd-github-actions-mini` golden fixture under `internal/quality/golden/`, laid out the way
GitHub requires (`src/.github/workflows/*.yml`, `src/.github/actions/*/action.yml`), with
`expected_entities` / `expected_relationships` covering the `vera_scope` the rule file claims —
workflow as `Service`, job as `Operation`, `needs:` as the job→job edge, `uses:` as the dependency
on an external action — plus forbidden rows for the over-fire direction.

It would be the first fixture in the tree that grades a CI extractor at all, and the only thing
that would make the ratchet non-blind to a walker change of this class.

Refs #6946, #6927, #6951.

PR #6956, opened 2026-09-07
fix(#6952): `conf/routes` has no extension, so four Play/Revel routes producers were unreachable — two table entries, not a new mechanism

### LINKED ISSUE #6952: conf/routes is never classified, so every Play (and Revel) routes producer in the tree is unreachable — three of them, in two packages

cajasmota:
Play's router file carries no extension, so `detectLanguage` returned `""` and both
`classifyInner` and `classifyWithSizeInner` answered `Skip{unsupported_extension}`. Four producers
were keyed on a file that never arrived — the two rules in
`engine/rules/scala/frameworks/play_framework.yaml`, `internal/custom/scala`'s `rePlayRoute`,
`internal/custom/java/play_routes.go` and `internal/custom/golang`'s Revel arm. All four have
passing unit tests, because those tests construct the file record directly.

> **Round 2.** Rebased onto `origin/main` (12 commits, including `59de3205d`, #6953's `(?m)` fix).
> Every figure below is re-derived from a run on the rebased tree, not carried across. The rebase
> changed an argument, not just numbers — see "What the rebase changed".

## Part 1 — the mechanism, read rather than assumed

**There was no missing hook.** `basenameLanguageMap` has routed extensionless well-known files
since `Dockerfile` / `Containerfile` / `Justfile` / `.justfile` / `Caddyfile` / `luarocks.lock`, and
`extensionLanguageMap` routes an unknown-to-the-OS extension like `.routes` exactly as it routes
`.tf`. The bug was **two absent table entries**, not an absent mechanism — so `Dockerfile`,
`Makefile`, `Gemfile`, `Procfile`, `CODEOWNERS` need no new machinery either; they need a measured
entry each. There is nothing here to generalise.

**Basename or path?** Both are expressible — `detectLanguage` receives the full normalised path and
already keys on it in four places (three compound suffixes and the Debezium `dirAnchor`). So the
narrow `conf/routes` anchor was available and was rejected on measurement.

```go
basenameLanguageMap["routes"]   = "scala"   // conf/routes, src/main/resources/routes
extensionLanguageMap[".routes"] = "scala"   // conf/admin.routes sub-router includes
```

## Part 2 — measured

Corpus `~/Projects/archigraph-corpora` (60 repos) plus cloned `playframework/playframework`,
`playframework/play-samples`, `lichess-org/lila`, `revel/examples`. Measured in this branch's own
worktree; the repo root was never walked.

| rule | files | what it misses |
|---|---|---|
| path suffix `conf/routes` | **70** | 9 at `src/main/resources/routes` + 60 `*.routes` includes |
| basename `routes` | **79** | the 60 `*.routes` includes |
| suffix `*.routes` | **60** | — |
| **both entries (this PR)** | **139** | — |

The `conf/` anchor reaches **half** the population, and what it misses is structural, not an
artefact of which repos were cloned: it cannot express `src/main/resources/routes` (Play's
maven-layout / sbt-plugin location) or any `*.routes` include at all.

**Over-fire: ZERO**, and the review widened the population it holds over — the ansible/infra corpus,
every other corpus tree on the machine, `*.route` anywhere, and extensionless files under any
`routes/` directory all measure zero. All 139 are genuine routes DSLs (the four `conf/routes` with
no column-0 verb are Play `->` aggregators, still routers). This repo's own blast radius is **zero
files**: `git ls-files` matches two routes files, both under `testdata/`, dropped by
`universalPathSkip`.

It is a **corpus-relative zero** and is named as one in the code comment. The shape that would break
it — a `files/etc/network/routes` static-route table in an ansible role, a `bin/routes` wrapper —
has no instance here and would be called `"scala"`.

**End to end, lila's five real routes files (923 column-0 verb lines), base vs branch:**

```
  base    0 entities / 0 relationships
  branch  1741 / 2647

  SCOPE.Operation  900   custom/scala, provenance PLAY_ROUTES_FILE
  Route            833   play_framework.yaml
  SCOPE.Component    5   one per-file carrier
  Module/External    3
```

**1738 of 1741 entities come from a routes file**, and every one is a `Route`, a `SCOPE.Operation`
or a carrier. The three that are not are Module/External aggregates. This is not a file-count
artefact: the fixture contains *nothing but* routes files, and the base binary extracts zero.

## What the rebase changed — and it is not just numbers

The first push priced #6959's duplicate-producer collision at "~0.4%, 4 `Route` vs 900, and 0
`ROUTES_TO` edges", and deferred it on that basis. `59de3205d` gives `play_framework.yaml` its
`(?m)`, and it is on `origin/main`. On the merge target:

- **833 `Route` against 900 `SCOPE.Operation` — ~93% duplication, not ~0.4%.**
- `ROUTES_TO` now fires: **907 edges, of which 905 have an unresolved to-side** (dangling
  `Operation:controllers.X.y` bare names that never bind, even in `scala-play-mini` where the
  target `SCOPE.Operation` exists in the same fixture).

So the rule pack's sole advantage over the custom extractor — the edge — is today almost entirely
unlanded, while its entity is a near-complete duplicate. **#6959 has been re-priced against these
numbers rather than left standing on the void ones**, and the same two figures are now recorded on
the rule itself in `play_framework.yaml` and on the inverted tripwire.

## Part 3 — why `"scala"`, and what that does and does not scope

All three custom producers hard-gate on their own base language
(`file.Language != "scala"`, `ctx.Language != "java"`, `language != "go"`). One file has one
language token, so **no single token reaches more than one of them**, and a routes-specific token
reaches **none**. `"scala"` is chosen because it is the only one whose producer already
content-sniffs what it is handed (`detectScalaFramework` answers `"play"` on a column-0 verb line).

**Those gates scope cross-language leakage only.** Same-language they do nothing:
`file.Language == "scala"` is now true for every routes file, so **all of `internal/custom/scala`'s
gated passes receive routes DSL as input** — anorm looking for SQL, caliban for GraphQL, akka for
actors, di for injection. The lila run is the evidence they produce nothing on it (1738/1741 from
routes files, zero spurious kinds), and that is now stated in the code so the next person adding a
Scala pass knows routes files are in its input.

**The cost of a wrong guess is small but not zero**, also now stated: `detectScalaFramework`'s
default arm returns `"scala"` rather than nothing, so a misclassified file still mints a
`SCOPE.Component` carrier and its `CONTAINS`, and `reGenericAuth` can mint a `SCOPE.Security`
`auth_check` from the word `Authorization`. "Yields no route entities rather than wrong ones" is
true about *routes*, not about entities.

Java and Revel producers stay unreachable → **#6957**.

## Grading — both directions

`internal/classifier/conf_routes_6952_test.go`

- **Reachability**: 7 paths must classify `scala` through **both** `Classify` and
  `ClassifyWithSize`. Carries the issue's `.scala` positive control.
- **Over-fire, basename half**: 8 rows that must keep their extension's language (every one an
  actual corpus path), 9 that must stay unclassified, 3 proving `universalPathSkip` still wins.
- **Over-fire, extension half** *(new in round 2 — this direction had zero rows and N1 was alive)*:
  `.route` singular, `.routez`, `.routes.bak`, `.routes~`.
- **`routes/` as a directory** *(new in round 2 — N2 proved the two rows advertised for this could
  not fire, since they carry extensions and extension lookup answers first)*: three EXTENSIONLESS
  files under a `routes/` directory. Synthetic by necessity — the corpus has six `routes`
  directories and zero extensionless files in any of them.
- The case asymmetry is pinned in both directions; the two halves genuinely differ.
- A row that was **tried and rejected** is recorded rather than omitted: bare `routes/` with a
  trailing slash classifies (`path.Base` strips it), but the classifier is only ever handed file
  paths, so a row for it would assert a contract that does not exist.

`internal/quality/golden/scala-play-mini/` — four files under `src/`: `conf/routes`,
`conf/admin.routes` (the *other* rule), **`main/resources/routes`** *(new in round 2 — the
maven-layout shape, 9 of the 139)*, and `conf/routes.bak` as the forbidden row. Base binary scores
**23/30**; branch **30/30**.

`internal/extractors/routes_incremental_reextract_6952_test.go` *(new in round 2)* — see below.

`internal/daemon/watch/routes_event_parity_6952_test.go` — the change event is not dropped; `.bak` /
`.tmp` / `.swp` are the negative control and `.bak` genuinely is on `SkipExts`.

## Corrected claim: FOUR consumers, not one

The first push said the classifier is consulted at one boundary. **That was false.** The consumers
are `cmd/grafel/index.go:3973`, `daemon/extract/coordinator.go:662`, `daemon/extract/subproc.go:105`
and **`internal/extractors/incremental.go:885/1002`** — the fourth being the UPDATE path, which is
exactly where the #6934 class lives. All four call the same classifier, so none can disagree, but
the enumeration was wrong and the watcher test only graded the *event*.

The re-extract is now graded too, on both conditions that can drop a changed routes file: the
classify step (`cr.Skip || cr.Language == ""`) and the registry lookup (`Get(cr.Language)`). The
second is why `"scala"` rather than a routes-specific token is a requirement and not a preference —
an unregistered token passes gate 1 and dies at gate 2, silently, after Step 5 has already evicted
the file's entities.

Stated and not claimed: `incremental.go` never calls `RunCustomExtractors`, so **no** incremental
re-extraction dispatches **any** of the ~340 custom extractors. That is pre-existing, universal, and
the same shape as the default-off `GRAFEL_INPROC_CUSTOM_EXTRACTORS` gate — not something #6952
introduced or can fix from the classifier.

## Rebase fallout, fixed rather than worked around

- **`TestIssue6927_Play_ConfRoutesIsNotClassified`** is a tripwire that pins the defect and goes red
  on the fix *by design*, instructing whoever fixes it to read the two consequences on the Route
  rule. Inverted to `..._IsClassified_6952`, carrying both consequences with the post-rebase
  numbers, plus a `routes.bak` negative control so the new rows are not satisfiable by a classifier
  that skips nothing. The rule's own comment is updated the same way.
- **`TestGoldenFixtureMarkersAreNotCommentOnly_6949`** fails when a fixture file has an extension its
  comment-prefix map does not know — a false-negative guard working exactly as designed. Three
  entries added (`.routes`, `routes`, `routes.bak`, all `#`). `routes.bak` is keyed by basename
  rather than adding a global `.bak` extension entry, which would claim every backup in the corpus.

## Mutants

Every edit's occurrence count asserted before writing and the replacement asserted after; green
baseline verified immediately before each score; `cp`-restore from a shasum-verified backup with
`shasum -c` after each.

| # | Mutant | Result |
|---|---|---|
| M1 | delete `basenameLanguageMap["routes"]` | DEAD — unit + fixture |
| M2 | delete `extensionLanguageMap[".routes"]` | DEAD — unit + fixture |
| M3 | widen to a `routes.` **prefix** (`isPlayRoutesFile`'s spelling) | DEAD — unit **and forbidden entity hits: 1**, the demonstration that the fixture's forbidden row is fireable |
| M4 | case-INsensitive basename lookup | DEAD |
| M5 | route both entries to `"yaml"` — classified as something nothing consumes | DEAD — unit + fixture drops to the base binary's score |
| M6 | replace both with the `conf/` path anchor | DEAD |
| **N1** | add `".route": "scala"` (singular) — **ALIVE at review**, the `.routes` half had zero forbidden rows | **re-scored DEAD** |
| **N2** | `strings.Contains(norm, "/routes/")` at the natural edit site — **ALIVE at review**; the two rows advertised for it could not fire | **re-scored DEAD** (3 rows) |
| **N3** | swap the extension/basename lookup order | **Equivalent under the current tables — recorded, not fixed.** The consequence is that the "extension consulted first" rationale is *inert*; the comment now says what actually holds: `basenameLanguageMap` is keyed on the FULL basename, so `routes.rb` is simply not the string `routes`. |
| **M6′** | the `conf/` **directory** anchor — scored **fixture-blind at 29/29** in review | **now DEAD by the fixture too (29/30, extracted 94)**, because `main/resources/routes` was added |

## Ratchet

`scala-play-mini` re-baselined **from a run on the rebased tree**, not carried across the rebase:

```
  entity_found                 29 -> 30   the maven-layout route
  entity_expected              29 -> 30
  entity_extracted_total       90 -> 97   +5 Route (the rule pack, now (?m)-enabled)
                                          +1 SCOPE.Operation + 1 carrier (main/resources/routes)
  relationship_extracted_total 210 -> 224 +7 ROUTES_TO (all unresolved), +7 CONTAINS
  relationship_found / expected  5 / 10, unchanged
```

Base binary on this fixture: **23/30 (76.7%), 81, 201**. Diffed entity-by-entity: 16 added, **0
removed, 0 re-kinded, 0 renamed**.

`bash scripts/quality/run.sh --ratchet` → exit 0, *"37 gated fixtures held their baseline (28 at
100% must-have recall)"*.

**The ratchet is not blind to this change** — the fixture is in the PR. **The `cmd/grafel` digest pin
still is**: `selftestdata/fixture/` is `main.go` + `util.py`, so `./cmd/grafel` passing unchanged is
a fact about the pin, not evidence about this change.

## Verification

`go test -count=1` exit 0 on `internal/classifier`, `internal/custom/scala`, `internal/quality`,
`internal/extractors`, `internal/engine`, `internal/daemon/watch` and `./cmd/grafel` (147s).
`gofmt -l` clean, `go vet ./internal/... ./cmd/...` clean.

Two `internal/daemon/watch` inotify-accounting tests
(`TestEventReleaseIsDeductedFromTheOwningRepo`, `TestInterleavedDirectoryFillsAcrossReposAreNot-
DoubleCharged`) failed once under load and passed on a clean re-run of the full package; they
concern watch-descriptor budgeting and touch no classification path. Reported rather than hidden.

Closes #6952
Refs #6927, #6934, #6946, #6949, #6951, #6953, #6957, #6959

### DIFF
--- a/internal/classifier/classifier.go
+++ b/internal/classifier/classifier.go
@@ -715,7 +715,29 @@ var basenameLanguageMap = map[string]string{
 	// handed (detectScalaFramework returns "play" on a column-0 HTTP verb line),
 	// so a non-Play file that happens to be named `routes` yields no route
 	// entities rather than wrong ones. Reaching the Java and Revel producers is
-	// a second change to those gates, filed separately.
+	// a second change to those gates, filed separately (#6957).
+	//
+	// WHAT THE LANGUAGE GATES DO NOT COVER, AND WHY IT IS STILL FINE. Those
+	// gates scope the routes token away from OTHER languages. They do nothing
+	// same-language: `file.Language == "scala"` is now true for every routes
+	// file, so ALL of internal/custom/scala's gated passes receive a routes DSL
+	// as if it were Scala source — anorm looking for SQL, caliban for GraphQL
+	// schemas, akka for actors, di for injection. Measured rather than argued:
+	// on lichess-org/lila's five real routes files (923 column-0 verb lines),
+	// this classification yields 1741 entities of which 1738 come from a routes
+	// file and every one is a Route, a SCOPE.Operation or a per-file carrier.
+	// Those passes produce ZERO spurious entities on real routes content. A
+	// future Scala custom pass should know routes files are in its input.
+	//
+	// THE COST OF A WRONG GUESS IS SMALL BUT NOT ZERO, so do not read the
+	// measured-zero over-fire as "harmless if it ever fires".
+	// detectScalaFramework's default arm returns "scala", not nothing, so a
+	// misclassified file still mints one SCOPE.Component carrier and its
+	// CONTAINS edge, and reGenericAuth (`\b(?:JWT|BearerToken|ApiKey|
+	// Authorization)\b`) fires for that arm — a stray `routes` file containing
+	// the word Authorization would mint a SCOPE.Security auth_check. "Yields no
+	// route entities rather than wrong ones" is true about ROUTES; it is not
+	// true about entities in general.
 	"routes": "scala",
 }
 
--- a/internal/classifier/classifier.go
+++ b/internal/classifier/classifier.go
@@ -616,6 +616,16 @@ var extensionLanguageMap = map[string]string{
 	// unexamined accept.
 	".dockerfile":    "dockerfile",
 	".containerfile": "dockerfile",
+	// Play / Revel routes DSL (#6952). Play splits a router across
+	// `conf/routes` plus per-module `conf/<name>.routes` includes
+	// (`-> /admin admin.Routes`); 60 of the 139 routes files measured across
+	// the Play upstreams use this suffix form, so routing the bare basename
+	// alone would reach fewer than half of them. Routed to "scala" — the
+	// language of the only registered producer that gates on a routes DSL
+	// (internal/custom/scala's `play` framework arm, selected by a content
+	// sniff for a column-0 HTTP verb) — see the basenameLanguageMap entry
+	// below for why "scala" and not a routes-specific token.
+	".routes": "scala",
 	// Markdown / Documentation
 	".md":       "markdown",
 	".mdx":      "markdown",
@@ -674,6 +684,39 @@ var basenameLanguageMap = map[string]string{
 	// reaches the _cross_manifest extractor (which exact-name-matches it and
 	// parses the resolved dependency tree as kind=locked deps).
 	"luarocks.lock": "lua",
+	// Play / Revel routes file (#6952). `conf/routes` carries no extension, so
+	// before this entry detectLanguage returned "" and the classifier answered
+	// Skip{unsupported_extension} — which made FOUR producers unreachable in
+	// production: the two rules in engine/rules/scala/frameworks/play_framework.yaml,
+	// internal/custom/scala's `rePlayRoute`, internal/custom/java/play_routes.go
+	// and internal/custom/golang's Revel arm. Every one of them had passing unit
+	// tests, because those tests construct the file record directly.
+	//
+	// WHY THE BARE BASENAME AND NOT A `conf/` PATH ANCHOR. `conf/routes` is the
+	// narrower rule and it is the one the Play docs describe, but it is not the
+	// one the corpus supports: of 139 routes files measured across
+	// playframework/playframework, playframework/play-samples, lichess-org/lila,
+	// revel/examples and the 60-repo corpus, only 70 sit at `conf/routes`. Nine
+	// more are Play's own `src/main/resources/routes` (the maven-layout and
+	// sbt-plugin form), and 60 are `*.routes` module includes that a `conf/`
+	// anchor also misses. The over-fire the anchor would buy protection against
+	// was measured at ZERO: every one of the 139 is a genuine routes DSL, and
+	// the corpus contains exactly one file named `routes` outside the Play
+	// upstreams (play-scala-starter's, which is one).
+	//
+	// WHY "scala" AND NOT A ROUTES-SPECIFIC LANGUAGE TOKEN. All three custom
+	// producers hard-gate on their own base language — `file.Language != "scala"`
+	// (custom/scala/frameworks.go), `ctx.Language != "java"`
+	// (custom/java/play_routes.go), `language != "go"` (custom/golang/revel.go)
+	// — so ONE token cannot reach all three; a routes-specific token reaches
+	// none of them without also widening all three gates and teaching the scala
+	// rule-pack loader a second language. "scala" is chosen because it is the
+	// only one of the three whose producer already content-sniffs the file it is
+	// handed (detectScalaFramework returns "play" on a column-0 HTTP verb line),
+	// so a non-Play file that happens to be named `routes` yields no route
+	// entities rather than wrong ones. Reaching the Java and Revel producers is
+	// a second change to those gates, filed separately.
+	"routes": "scala",
 }
 
 // detectLanguage returns the language token for the given normalised path, or
--- a/internal/classifier/conf_routes_6952_test.go
+++ b/internal/classifier/conf_routes_6952_test.go
@@ -114,7 +114,18 @@ func TestRoutesNameDoesNotOverFire6952(t *testing.T) {
 	c := New(nil)
 
 	// Rows that must keep the language the EXTENSION router already gave them.
-	// These are the ones a `Contains`-style rule would steal.
+	//
+	// WHAT ACTUALLY PROTECTS THESE, corrected after review. It is tempting to
+	// say "extensionLanguageMap is consulted before basenameLanguageMap, so
+	// routes.rb is ruby before the routes entry is ever reached". That ordering
+	// is INERT: swapping the two lookups is equivalent under the current tables
+	// (probed on nginx.conf, luarocks.lock, .justfile, conf/routes and
+	// config/routes.rb — identical output both ways, because `.conf` and
+	// `.lock` are not in extensionLanguageMap and `.justfile` is in both with
+	// the same value). The thing that actually holds is that
+	// basenameLanguageMap is keyed on the FULL basename, exactly: `routes.rb`
+	// is simply not the string `routes`. A reader tightening the rule should
+	// preserve exact-full-basename keying, not the lookup order.
 	keepsLanguage := []struct{ path, lang, corpus string }{
 		{"config/routes.rb", "ruby", "rails-realworld/config/routes.rb"},
 		{"config/routes.yaml", "yaml", "symfony-demo/config/routes.yaml"},
@@ -157,6 +168,57 @@ func TestRoutesNameDoesNotOverFire6952(t *testing.T) {
 			if !got.Skip || got.Language != "" {
 				t.Errorf("Classify(%q) = {lang=%q skip=%v}, want an unclassified skip — %s",
 					tc.path, got.Language, got.Skip, tc.why)
+			}
+		})
+	}
+
+	// THE `.routes` EXTENSION HALF, graded on its own. Every row above keys on
+	// the BASENAME entry; before review this direction had zero forbidden rows,
+	// and adding `".route": "scala"` (singular) to extensionLanguageMap was
+	// ALIVE against the whole suite and the golden fixture. Two populations were
+	// measured (79 basename / 60 suffix) and only one was graded.
+	extensionHalfStaysSkipped := []struct{ path, why string }{
+		{"conf/admin.route", "SINGULAR. Zero instances in 139 measured files; Play's include suffix is plural, and a `.route` entry is invisible to every basename row"},
+		{"conf/admin.routez", "a one-character-off suffix — the lookup is exact, not a prefix or fuzzy match"},
+		{"conf/admin.routes.bak", "filepath.Ext takes the LAST segment, so this is `.bak` and must stay skipped even though `.routes` appears in the name"},
+		{"conf/admin.routes~", "the editor backup marker; `.routes~` is not `.routes`"},
+	}
+	for _, tc := range extensionHalfStaysSkipped {
+		t.Run("ext/"+tc.path, func(t *testing.T) {
+			if got := c.Classify(context.Background(), tc.path); !got.Skip || got.Language != "" {
+				t.Errorf("Classify(%q) = {lang=%q skip=%v}, want an unclassified skip — %s",
+					tc.path, got.Language, got.Skip, tc.why)
+			}
+		})
+	}
+
+	// A `routes/` DIRECTORY must not pull its contents in. The two rows in
+	// keepsLanguage that advertise this (express-realworld, nextjs) CANNOT
+	// grade it — review's N2 proved it: they carry extensions, so the extension
+	// lookup answers first and a `strings.Contains(norm, "/routes/")` widening
+	// at the natural edit site never reaches them. Killing that shape needs an
+	// EXTENSIONLESS file under a routes/ directory, which is what these are.
+	// The corpus has six `routes` directories (symfony-demo, nextjs x2,
+	// rails-actionpack, express-realworld, awesome-compose) and zero
+	// extensionless files in any of them, so these rows are synthetic by
+	// necessity — the widening is not production-reachable on the measured
+	// corpus, but it is the cheapest wrong turn a future editor can take.
+	routesDirStaysSkipped := []string{
+		"routes/health",
+		"app/routes/index",
+		"src/routes/admin/handler",
+		// NOT a row: a bare `routes/` with a trailing slash. path.Base strips it
+		// and the file classifies — but the classifier is only ever handed FILE
+		// paths (the walker decides directories, and does so before this), so a
+		// row for it would assert a contract that does not exist. Recorded
+		// rather than silently omitted: it was tried and rejected, not missed.
+	}
+	for _, p := range routesDirStaysSkipped {
+		t.Run("dir/"+p, func(t *testing.T) {
+			if got := c.Classify(context.Background(), p); !got.Skip || got.Language != "" {
+				t.Errorf("Classify(%q) = {lang=%q skip=%v}, want an unclassified skip — a "+
+					"segment match on `routes/` would fire here; the rule is a basename "+
+					"lookup and must not read the parent directory", p, got.Language, got.Skip)
 			}
 		})
 	}
--- a/internal/classifier/conf_routes_6952_test.go
+++ b/internal/classifier/conf_routes_6952_test.go
@@ -1,0 +1,198 @@
+// Package classifier — conf_routes_6952_test.go
+//
+// Issue #6952. Play Framework's router file carries NO EXTENSION, so
+// detectLanguage returned "" for it and both classifyInner and
+// classifyWithSizeInner answered Skip{unsupported_extension}. Four producers
+// were keyed on a file that never arrived — the two rules in
+// internal/engine/rules/scala/frameworks/play_framework.yaml, internal/custom/
+// scala's rePlayRoute, internal/custom/java/play_routes.go, and internal/custom/
+// golang's Revel arm — and every one of them had passing unit tests, because
+// those tests construct the file record directly and never go through here.
+//
+// THERE WAS NO MISSING MECHANISM. basenameLanguageMap has routed extensionless
+// well-known files since Dockerfile / Containerfile / Justfile / Caddyfile, and
+// extensionLanguageMap routes an unknown-to-the-OS extension like `.routes`
+// exactly as it routes `.tf` or `.hcl`. The bug was two absent table entries,
+// not an absent hook, which is why Dockerfile / Makefile / Gemfile / Procfile
+// need no new machinery either.
+//
+// WHY THE BARE BASENAME AND NOT A `conf/` PATH ANCHOR. detectLanguage CAN
+// express a path rule — the Debezium `.json` block's dirAnchor does — so the
+// narrower `conf/routes` rule was available and was rejected on measurement,
+// not on taste. Across playframework/playframework, playframework/play-samples,
+// lichess-org/lila, revel/examples and the 60-repo corpus at
+// ~/Projects/archigraph-corpora, 139 routes files:
+//
+//	basename `routes`      79 files   (70 at conf/routes, 9 at src/main/resources/routes)
+//	suffix   `*.routes`    60 files   (conf/<module>.routes sub-router includes)
+//	path     `conf/routes` 70 files   — HALF the population
+//
+// and the over-fire the path anchor buys protection against measured ZERO:
+// every one of the 139 is a genuine routes DSL, and the whole 60-repo corpus
+// contains exactly ONE file named `routes` (play-scala-starter's, which is a
+// Play routes file). The generic-name risk is real in the abstract and was not
+// observed once; the forbidden rows below are what hold it, together with the
+// fact that the consumer content-sniffs (detectScalaFramework only answers
+// "play" for a column-0 HTTP verb line).
+//
+// WHY "scala". All three custom producers hard-gate on their own base language
+// (`file.Language != "scala"`, `ctx.Language != "java"`, `language != "go"`),
+// so no single token reaches more than one of them and a routes-specific token
+// reaches NONE. "scala" is the only one of the three whose producer already
+// content-sniffs what it is handed. Reaching the Java and Revel producers needs
+// those gates widened and is filed separately.
+package classifier
+
+import (
+	"context"
+	"testing"
+)
+
+// TestConfRoutesIsClassified6952 grades the REACHABILITY direction: the paths
+// that carry a Play/Revel routes DSL must arrive at a language with a producer.
+//
+// The `.scala` row is a POSITIVE CONTROL, carried from the issue: without it a
+// blanket regression in the classifier (an early skip, a broken normalisePath)
+// would look identical to the bug this file pins, and the fix would read as
+// necessary when it was not.
+func TestConfRoutesIsClassified6952(t *testing.T) {
+	c := New(nil)
+	cases := []struct {
+		path string
+		why  string
+	}{
+		{"conf/routes", "the canonical Play/Revel location; 70 of 139 measured files"},
+		{"modules/admin/conf/routes", "a Play sub-project's own conf/ — path depth must not matter"},
+		{"src/main/resources/routes", "Play's maven-layout / sbt-plugin location; 9 measured files, and every one of them is invisible to a `conf/` path anchor"},
+		{"routes", "repo root, e.g. a sub-project indexed with its own directory as root"},
+		{"conf/admin.routes", "`-> /admin admin.Routes` sub-router include; 60 measured files"},
+		{"a.routes", "the include form at the tree root (playframework's routes-compiler-incremental-compilation fixtures)"},
+		{"documentation/manual/working/scalaGuide/main/http/code/scalaguide.http.routing.routes", "a multi-dot include name — filepath.Ext takes only the LAST segment, so this must route on `.routes` and not on `.http`"},
+	}
+	for _, tc := range cases {
+		t.Run(tc.path, func(t *testing.T) {
+			got := c.Classify(context.Background(), tc.path)
+			if got.Skip {
+				t.Fatalf("Classify(%q) = Skip{%s} — %s", tc.path, got.SkipReason, tc.why)
+			}
+			if got.Language != "scala" {
+				t.Fatalf("Classify(%q).Language = %q, want %q — internal/custom/scala's "+
+					"framework extractor gates on `file.Language != \"scala\"`, so any "+
+					"other token leaves the Play producer as unreachable as an outright "+
+					"skip does", tc.path, got.Language, "scala")
+			}
+			// ClassifyWithSize is a SECOND copy of the decision (classifier.go
+			// :190) and is the one the extraction pipeline actually calls
+			// (cmd/grafel/index.go:3973). A fix applied to only one of them
+			// would be green on Classify and dead in production.
+			ws := c.ClassifyWithSize(context.Background(), tc.path, 4096)
+			if ws.Skip || ws.Language != "scala" {
+				t.Fatalf("ClassifyWithSize(%q, 4096) = {lang=%q skip=%v reason=%s}, want "+
+					"{scala,false} — the two entry points disagree", tc.path, ws.Language, ws.Skip, ws.SkipReason)
+			}
+		})
+	}
+
+	// Positive control (#6952 as filed): a co-located ordinary Scala file was
+	// ALWAYS classified. The skip was a fact about the filename, not about the
+	// path or the repo layout.
+	if got := c.Classify(context.Background(), "app/controllers/HomeController.scala"); got.Skip || got.Language != "scala" {
+		t.Fatalf("premise: Classify(HomeController.scala) = {lang=%q skip=%v} — the control "+
+			"is broken, so nothing above grades the routes rules", got.Language, got.Skip)
+	}
+}
+
+// TestRoutesNameDoesNotOverFire6952 grades the OTHER direction, which is where
+// the whole risk of this change lives: `routes` is a plausible basename for
+// something that is not a Play router.
+//
+// Every row is a name that EXISTS in the measured corpus (the file it names is
+// noted) except where marked synthetic, and every row would flip if the rule
+// were spelled one notch wider — a `strings.Contains(base, "routes")`, a
+// `routes.` prefix arm, or a case-insensitive lookup.
+func TestRoutesNameDoesNotOverFire6952(t *testing.T) {
+	c := New(nil)
+
+	// Rows that must keep the language the EXTENSION router already gave them.
+	// These are the ones a `Contains`-style rule would steal.
+	keepsLanguage := []struct{ path, lang, corpus string }{
+		{"config/routes.rb", "ruby", "rails-realworld/config/routes.rb"},
+		{"config/routes.yaml", "yaml", "symfony-demo/config/routes.yaml"},
+		{"src/routes.rs", "rust", "actix-examples/databases/mysql/src/routes.rs"},
+		{"app/Http/routes.php", "php", "laravel-quickstart/app/Http/routes.php"},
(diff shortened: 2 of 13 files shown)


cajasmota:
Follow-ups filed rather than folded in: #6957 (the Java + Revel producers stay unreachable — each hard-gates on its own base language, so no single token reaches all three) and #6959 (the Route / ROUTES_TO ownership merge, measured at 900 vs 4 on lila and 0 ROUTES_TO edges — it becomes 1:1 the moment #6953 lands).

cajasmota:
Coordinator mutant on `76a48158d`. One mutant, **ALIVE** — and chasing why it lives exposed a question your safety argument does not cover. I am not asking for a code change.

## MM-1 — drop ONE producer's language gate

`internal/custom/scala/anorm.go:85`: `if len(file.Content) == 0 || file.Language != "scala"` → `if len(file.Content) == 0`. Pre-edit count 1, post-edit 1, `go vet` clean, baseline green first (classifier 1.540s, custom/scala 0.583s), `-count=1`, restored by `cp` from a shasum-verified backup (`f8f56bc50…`), `git status --porcelain` empty.

**Verdict: ALIVE** at `./internal/classifier/` + `./internal/custom/scala/`. The hard gates are ungraded — but that is **pre-existing and not yours**, so it is not the finding.

## The finding: the gates do not gate routes files, because routes files ARE Scala now

Your "forced, not preferred" argument is that all three routes producers hard-gate on their own base language, so a routes token reaches at most one of them. That is right, and I verified it — but it answers **cross-language leakage only**.

The unanswered half is **same-language**: `routes` is classified as `"scala"`, so `file.Language == "scala"` is *true* for every routes file, and **all 11 gated Scala custom passes now receive Play routes DSL as if it were Scala source** — anorm looking for SQL, caliban for GraphQL schemas, akka for actors, di for injection. The gate that stops a Go producer seeing it does nothing to stop these.

**Your own end-to-end number is the evidence, and it is good evidence** — lila **909/909 signal** across ~900 routes files means those eleven passes produced exactly zero spurious entities on real routes content. That is a stronger answer than any fixture would give.

But it is evidence you gathered for a different question, and nothing in the PR *states* it answers this one. **Add a sentence**: the routes token is scoped away from other languages by their gates, and same-language passes were measured to produce nothing on it, at 909/909 on lila. Otherwise the next person adding a Scala custom pass has no idea routes files are in its input.

That also sharpens #6959: two producers own the Play `Route` because the routes file legitimately reaches both the custom lane and the rule pack — the ownership question is a consequence of this classification, not an accident beside it.

## On the round

The mechanism answer is the best kind — **there was no missing hook.** `basenameLanguageMap` has routed extensionless well-known files since `Dockerfile`, so Dockerfile/Makefile/Gemfile need a measured entry each and not a new mechanism. A two-entry fix that reports "the general thing you asked me to build already exists" is worth more than a general thing built.

Two calls I want to credit specifically, because both are places the previous arm went the other way:

- **You put the fixture in rather than deferring it.** `scala-play-mini` scores 23/29 on the base binary and 29/29 on the branch, so the ratchet genuinely sees this change — where #6951 had to file #6955 because no fixture could see it. Adding `conf/routes.bak` as a negative alongside is what makes the positive mean something.
- **You reported your own fixture's blind spot.** M6 — the rejected `conf/` path anchor — dies on the unit test and **not** on the golden fixture, because every routes file in the fixture lives under `conf/`. Saying so in the PR rather than letting 29/29 speak for it is the difference between a fixture and a claim about a fixture.

And `cmd/grafel`'s digest pin passing being "a fact about the pin, not evidence" is exactly the right way to report a green that means nothing.

Refs #6952, #6957, #6959, #6927.

cajasmota:

Own checkout: detached worktree at `scratchpad/review-6956-zx9/wt` @ `76a48158d`, isolated `HOME`/`GRAFEL_HOME`/`GRAFEL_DAEMON_ROOT` under `scratchpad/review-6956-zx9/iso`. The author's `impl-6952-k7q2` and `.claude/worktrees/` were not touched. `go vet` clean, `-count=1` throughout, every mutant restored by `cp` from a shasum-verified backup (`8e1a37f8…`), tree verified clean and byte-identical at the end.

**The design is right and the evidence mostly reproduces. One finding blocks: the branch is 12 commits behind `origin/main` and the deferral of #6959 rests on a number that is void on the merge target.**

---

### 1. BLOCKER — the "~0.4% collision, filed rather than pre-empted" argument does not survive the merge

The PR body's own caveat is that `play_framework.yaml` produces "4 `Route` and 0 `ROUTES_TO`" against 900 custom routes, and that "#6953's `(?m)` fix will make it 1:1". `59de3205d` is **already on `origin/main`**; this branch is based on `fc9124eef` and does **not** contain it. So the headline measurements describe a tree that no longer exists.

Measured, not argued. I built a fixture from lila's five real routes files (`conf/routes`, `appeal/clas/report/team.routes`) and ran `grafel quality` three ways:

| binary | entities | relationships |
|---|---|---|
| base (both entries deleted from this branch) | **0** | **0** |
| this branch as-is | **910** | **913** |
| this branch **merged with `origin/main`** | **1741** | **2647** |
| merged, with only the two `(?m)` in `play_framework.yaml` reverted | 910 | 913 |

The last row is the causal isolation: reverting nothing but `(?m)` restores 910/913 exactly, so **+831 entities and +1734 relationships are attributable to this PR's classifier change meeting `59de3205d` on main**. The collision is not ~0.4%; on the merge target it is ~91% of the entities this arm adds, and it nearly triples the edge count from the same five files.

`scala-play-mini` moves too: merged, it scores 29/29 but `entity_extracted_total` **90 → 94** and `relationship_extracted_total` **210 → 220**. The `baseline.json` figures this PR just re-recorded do not describe the tree that will exist after merge.

Asked for: rebase onto current `origin/main`, re-derive the lila / fixture figures and the baseline, and re-state whether #6959 is still deferrable at 1:1 rather than at 4-vs-900. If it is, say so against the real number.

### 2. The zero over-fire — verified, and widened

Independently re-derived on `~/Projects/archigraph-corpora`: exactly **one** file named `routes` (`play-scala-starter/conf/routes`, a Play router), and **six** `routes` *directories* (symfony-demo, nextjs ×2, rails-actionpack, express-realworld, awesome-compose) that a basename rule structurally cannot reach. Matches the PR exactly.

I then extended the measurement to populations the PR did not measure, because "framework sample apps" is the wrong corpus to prove a generic name safe:

- the scratchpad **ansible/infra corpus** (`ansible-for-devops` et al.) — **0** bare `routes`, **0** `*.routes`;
- every other corpus tree on this machine outside the Play upstreams — **0**;
- **0** `*.route` files anywhere;
- **0** extensionless files under any `routes/` directory (27 `.ts`, 4 `.yaml`, 2 `.erb`, 1 `.js`).

So the zero holds over a wider population than claimed. The honest residual: the shape that would break it — a `files/etc/network/routes` static-route table in an ansible role, a `bin/routes` wrapper script — has no instance here, and grafel will call it `"scala"`. That is a corpus-relative zero and it should be named as one.

**What actually happens when the guess is wrong.** I verified the three gates (`internal/custom/scala/frameworks.go:465 file.Language != "scala"`, `internal/custom/java/play_routes.go`, `internal/custom/golang/revel.go`) — the mutual-exclusion argument is correct, one token reaches one producer. But the cost is not zero:

- `detectScalaFramework` returns `"scala"` by default, not "nothing" — the file still goes down the general scala arm and still mints **one `SCOPE.Component` file carrier plus a `CONTAINS`**. The PR's own play-scala-rest-api row (`+1`, "file carrier, no verbs") is that cost, observed.
- `reGenericAuth` is `\b(?:JWT|JwtToken|BearerToken|ApiKey|Authorization)\b` and fires for `framework == "scala"`. A stray `routes` file containing the word `Authorization` mints a `SCOPE.Security` `auth_check` with `framework: "scala"`.

"Yields no route entities rather than wrong ones" is true about *routes*. It is not true about *entities*. Worth one sentence in the comment block, since that block is the thing a future reader will trust.

### 3. basename-vs-path — the rejection is correct

Structurally verified, not just counted: a `conf/` anchor cannot express `src/main/resources/routes` or any `*.routes` include, which is the majority of the population by construction and not an artefact of which repos were cloned. I re-spelled the rejected design as a `conf/` **directory anchor** (the strongest form of it, gating both entries on the path) and it fails 5 of the 7 reachability rows plus `TestRoutesExtensionIsRoutedByTheRouter6952`. The basename+extension spelling is the right call.

### 4. Reproduced claims

- **Re-baseline, re-derived from my own runs**, not read from the diff: base binary `entity_found` 23/29 (79.3%), `entity_extracted_total` 81, `relationship_extracted_total` 201 → branch 29/29, 90, 210, `relationship_found` 5/10 unchanged, `forbidden_entity_hits` 0. Exact match with the PR.
- **End-to-end signal holds.** Base extracts **zero** entities and **zero** relationships from lila's five real routes files; the branch extracts 910/913. 923 raw column-0 verb lines in those files, so ~900 unique after `canonicalScalaColonPath` collapsing is the right order. This is not a file-count artefact — the fixture contains *nothing but* routes files.
- **`cmd/grafel`'s digest pin is blind**, confirmed: `selftestdata/fixture/` is `main.go` + `util.py`, full stop.
- **"No missing hook"** — confirmed by reading. `basenameLanguageMap` already carries `Dockerfile`/`Containerfile`/`Justfile`/`justfile`/`.justfile`/`nginx.conf`/`Caddyfile`/`luarocks.lock`. Two entries, not a mechanism.

### 5. Not confirmed — the consumer enumeration is incomplete

> The classifier is consulted at ONE boundary (`cmd/grafel/index.go:3973`, `internal/daemon/extract/{coordinator,subproc}.go`)

There is a fourth site: **`internal/extractors/incremental.go:885`** (`classifier.New(nil)`) and **`:1002`** (`ClassifyWithSize`). It is named in neither the PR body nor the header of `routes_event_parity_6952_test.go`.

Behaviourally this is fine — it is the same classifier, so a changed `conf/routes` does get re-extracted. But "there is no second consumer" is false as written, and it matters for the #6934 class specifically: the watcher test proves the *event* is not dropped, and nothing here proves the *re-extract* lands. The update path is `incremental.go`, and it is ungraded for a routes file. Please fix the enumeration, and consider one row in the incremental path.

### 6. Mutants (mine, not the six in the PR)

Every edit asserted at **exactly one** occurrence before writing, replacement asserted present after writing, green baseline verified immediately before each score, `cp`-restore + shasum after each.

| # | Mutant (production, all more-permissive except N3) | Result |
|---|---|---|
| **N1** | add `".route": "scala"` to `extensionLanguageMap` | **ALIVE — distinguishable and ungraded.** `internal/classifier` green (exit 0), fixture blind. All 17 over-fire rows grade the *basename* half; **the `.routes` half has zero forbidden rows**. No corpus instance of `*.route`, so no observed harm — but the direction is entirely ungraded. |
| **N2** | `if strings.Contains(norm, "/routes/") \|\| strings.HasPrefix(norm, "routes/") { return "scala" }`, inserted at the natural edit site (after the extension lookup) | **ALIVE — ungraded.** The two rows advertised as grading exactly this (`express-realworld … "a segment match would fire on the path, a basename match cannot"`, and the nextjs "file INSIDE a routes/ directory" row) **cannot** kill it: they carry extensions, and the extension lookup runs first, so the widening never reaches them. A row would need an *extensionless* file under a `routes/` dir. Not production-reachable on the measured corpus (0 such files). |
| **N3** | swap the `extensionLanguageMap` and `basenameLanguageMap` lookups | **Equivalent under the current tables.** Probed `nginx.conf`, `luarocks.lock`, `.justfile`, `conf/routes`, `config/routes.rb` — identical output both ways (`.conf` and `.lock` are not in `extensionLanguageMap`; `.justfile` is in both maps with the same value). Consequence worth recording: the "`extensionLanguageMap` is consulted first" rationale that both the PR body and the `keepsLanguage` comment lean on is **inert today**. What actually protects `config/routes.rb` is that `basenameLanguageMap` is keyed on the *full* basename. |
| **M6′** | the rejected design re-spelled as a `conf/` **directory anchor** (gate both entries on `conf/`) | DEAD by unit (5 rows), **fixture blind: 29/29, 90, 210 — byte-identical to unmutated.** Confirms the PR's own M6 admission. |
| **M6″** | the *narrow* spelling — `norm == "conf/routes" \|\| HasSuffix(norm, "/conf/routes")` only | DEAD by unit **and by the fixture at 28/29** — the `conf/admin.routes` row catches it. So the fixture is blind to the dir-anchored shape, not to every path-anchored shape. Worth stating precisely; the PR's M6 line is slightly harder on itself than the evidence requires. |
| **BASE** | delete both entries (= the pre-PR classifier), built as its own binary | DEAD — fixture 23/29. Positive control for every figure above. |

### 7. Should the fixture's `conf/` blindness block?

**No, not on its own** — `TestConfRoutesIsClassified6952` kills M6′ on 5 rows and runs in CI on every PR, so the shape is defended, just not in the end-to-end harness. But since a rebase and a re-baseline are required anyway (§1), add the file while you are there: **one `src/main/resources/routes` in `scala-play-mini` plus one `expected_entities` row**. That is a real Play shape (9 of the 139 measured files), it closes the blindness at the exact axis the design decision turns on, and it costs one file.

### Summary of what would flip this to APPROVE

1. **Rebase onto `origin/main`, re-measure, re-baseline** — and re-argue #6959's deferral against 1:1, not against 4-vs-900. (blocking)
2. One `src/main/resources/routes` fixture file, closing the M6′ blindness. (cheap, recommended)
3. A forbidden row or two for the `.routes` **extension** half (N1). (recommended)
4. Fix the "one boundary" enumeration to include `internal/extractors/incremental.go`, and either grade the incremental re-extract of a routes file or say out loud that it is ungraded. (recommended)
5. Soften "yields no route entities rather than wrong ones" to name the file carrier and the `reGenericAuth` path. (recommended)

The corpus work, the basename-vs-path decision, the two-directions unit test and the end-to-end reproduction are all sound and independently reproduced. It is the staleness against `main` that stops this being an approve today.

cajasmota:
## Round 2 — all five addressed. The blocker was right, and it was right about more than the numbers.

Rebased onto `origin/main` at `59de3205d`; force-pushed as `d170d6863`. Every figure re-derived from
a run on the rebased tree — nothing carried across the rebase, per the "identical-change rebase
drops silently" rule.

### 1. BLOCKER — rebased, and the re-derivation reproduces your measurement exactly

`1741 / 2647` on the merged tree, `0 / 0` on base. Your causal isolation holds. What I then measured
that the isolation implies but does not show:

```
  SCOPE.Operation  900   custom/scala
  Route            833   play_framework.yaml     -> 93% duplication, not 0.4%
  ROUTES_TO        907   of which 905 UNRESOLVED on the to-side
```

**905 of 907 `ROUTES_TO` edges dangle** — bare `Operation:controllers.X.y` that never binds, and it
is not a lila artefact: `scala-play-mini` emits 7 and every one is unresolved *even though the target
`SCOPE.Operation` exists in the same fixture*. So the rule pack's only claimed advantage — the edge —
is almost entirely unlanded, while its entity is a near-complete duplicate.

**Answering the question you actually asked** (does 91% change whether it can be deferred): it changes
the *shape* of #6959 more than the *urgency*. It is still a merge, not a deletion, but the ordering
inverts — resolving the to-side is now the load-bearing half, not choosing an owner. #6959's body is
**re-priced against the post-merge numbers**, with the void ones retained below the line rather than
quietly replaced. The same two figures are recorded on the rule in `play_framework.yaml` and on the
inverted tripwire, so the next reader meets them at the code and not only in an issue.

`scala-play-mini` re-baselined from a run: **30/30, 97, 224** (base binary 23/30, 81, 201). The
+3/+4 over your predicted 94/220 is item 5's new fixture file.

### 2. N1 and N2 — both re-scored DEAD

You were right that I measured two populations and graded one.

- **N1**: four forbidden rows for the `.routes` half — `.route` singular, `.routez`, `.routes.bak`,
  `.routes~`.
- **N2**: three **extensionless** files under a `routes/` directory. Your diagnosis was exact — the
  two rows I advertised for this carry extensions, so extension lookup answers first and the
  widening never reaches them. An assertion whose population is narrower than its name.

One row I **tried and rejected**, recorded in the test rather than silently omitted: bare `routes/`
with a trailing slash. `path.Base` strips it and the file classifies — but the classifier is only
ever handed file paths, so a row for it asserts a contract that does not exist.

### 3. N3 — recorded, not fixed

Agreed and corrected in place. The `keepsLanguage` comment now says the ordering is **inert** (with
your probe list), and that what actually protects `config/routes.rb` is exact-full-basename keying —
`routes.rb` is simply not the string `routes`. A reader tightening the rule should preserve the
keying, not the order.

### 4. The "ONE boundary" claim — false, corrected, and the gap graded

Four consumers now enumerated, including `internal/extractors/incremental.go:885/1002`. You were
right that the watcher test proves the *event* and nothing proved the *re-extract*, so
`internal/extractors/routes_incremental_reextract_6952_test.go` grades the two conditions that can
drop a changed routes file there — and the second (`Get(cr.Language)`) is why routing to a
registered language is a **requirement**, not a preference: an unregistered token passes gate 1 and
dies at gate 2 after Step 5 has already evicted the file's entities.

Found while writing it, stated rather than claimed: **`incremental.go` never calls
`RunCustomExtractors` at all**, so no incremental re-extraction has ever dispatched any of the ~340
custom extractors. Pre-existing, universal, same shape as the default-off gate — not something #6952
introduced or can fix from the classifier.

### 5. `src/main/resources/routes` added — and it closes M6′

Added to `scala-play-mini`. **M6′ now dies on the fixture too: 29/30, extracted 94** (it was a
byte-identical 29/29 before). Your M6″ point is taken and reflected: the fixture was blind to the
*dir-anchored* shape specifically, not to every path-anchored shape.

### Your mutant, folded in

Both halves are now in the code, not just the PR body. The classifier comment states that the
language gates scope **cross-language leakage only**, that same-language every `custom/scala` pass
now receives routes DSL as input, and that lila's 1738-of-1741 is the measured answer that they
produce nothing on it. And your second point — that the cost of a wrong guess is small but **not
zero** — is stated with its mechanism: `detectScalaFramework`'s default arm returns `"scala"`, so a
misclassified file still mints a carrier and its `CONTAINS`, and `reGenericAuth` can mint a
`SCOPE.Security` `auth_check` from the word `Authorization`. "No route entities rather than wrong
ones" is true about routes, not about entities.

### Rebase fallout — fixed, not worked around

Two of your merged-tree neighbours have guards that fire on this change by design, and both were
worth reading rather than silencing:

- **`TestIssue6927_Play_ConfRoutesIsNotClassified`** is a tripwire pinning the defect, whose failure
  message instructs whoever fixes it to read the two consequences on the Route rule. Inverted to
  `..._IsClassified_6952` carrying both, with post-rebase numbers, plus a `routes.bak` negative
  control so the new rows are not satisfiable by a classifier that skips nothing.
- **`TestGoldenFixtureMarkersAreNotCommentOnly_6949`** rejects a fixture file whose extension its
  comment-prefix map does not know — the false-negative guard doing its job. Three entries added
  (`#`), with `routes.bak` keyed by **basename** rather than a global `.bak` extension entry, which
  would have claimed every backup in the corpus.

### Verification

`-count=1` exit 0 on `internal/classifier`, `internal/custom/scala`, `internal/quality`,
`internal/extractors`, `internal/engine`, `internal/daemon/watch`, `./cmd/grafel` (147s), and
`run.sh --ratchet` (37 fixtures, exit 0). `gofmt -l` and `go vet ./internal/... ./cmd/...` clean.

Reported rather than hidden: `TestEventReleaseIsDeductedFromTheOwningRepo` and
`TestInterleavedDirectoryFillsAcrossReposAreNotDoubleCharged` failed once under load in a combined
six-package run and passed on a clean re-run of the full `watch` package. Both are watch-descriptor
budget accounting and touch no classification path.

Not merged.

cajasmota:
## Coordinator mutant on the final head (`d170d6863`)

My earlier mutant was scored on the pre-rebase `76a48158d`, so it did not count for this head. Scored in my own detached worktree at `d170d6863`, `git status --porcelain` empty, restored by `cp` from a shasum-verified backup (`179e5d18…` both ways).

**MC-1 — make the *Java* Play routes producer also accept the language this PR now stamps.** Nobody scored the producer side; all nine prior mutants (your six, the reviewer's three) stayed inside the classifier.

`internal/custom/java/play_routes.go:145` — `ctx.Language != "java"` → `(ctx.Language != "java" && ctx.Language != "scala")`. Edit proven landed: `pre=1`, `post_old=0`, `post_new=1`. `go vet` clean.

**Result: ALIVE** on `internal/custom/java`, `internal/classifier`, `internal/custom/scala`, `internal/engine` (`-count=1`, exit 0, baseline verified green immediately before).

**Verdict: alive but not production-reachable** — and the reason is the interesting part.

`ctx.Language` is a **constant by construction**. `patterns_dispatch.go:255` gates the whole dispatcher on `strings.ToLower(file.Language) != "java"`, and `:277` then **hardcodes** `Language: "java"` into every `PatternContext` it builds. The `ctx.Language != "java"` check inside `ExtractPlay` can never be false in production. No fixture can kill this mutant, and none should be manufactured to try.

## What that means for a claim in this PR

> all three routes producers hard-gate on their own base language (`!= "scala"` / `!= "java"` / `!= "go"`) so no token reaches more than one

**The conclusion is right and the mechanism is wrong for two of the three.** The separation is real — I am satisfied there is no over-fire and this is safe to merge — but it is carried by **dispatch in all three cases**, not by the in-function language gates you cited:

| producer | cited gate | what actually separates it |
|---|---|---|
| scala | `!= "scala"` | registration under `custom_scala_frameworks` |
| java | `ctx.Language != "java"` (`play_routes.go:145`) | dispatcher gate `patterns_dispatch.go:255` + hardcoded `Language: "java"` at `:277` — **the cited gate is inert** |
| go / revel | `!= "go"` | registration under `custom_go_revel` — **`revelExtractor.Extract` has no language gate at all before its `conf/routes` branch**; `file.Language != "go"` at `revel.go:148` guards only the Go-source half *below* it |

The Revel one is worth stating plainly: `isRevelRoutesFile(file.Path)` accepts bare basename `routes` (`revel.go:66`), and the routes branch at `:110` runs before any language check. If that extractor were ever dispatched a `conf/routes` file, it would parse it. Only the registered token stops it.

This is the fourth instance this cycle of **protection living somewhere other than where you'd look**, and it has a specific consequence for #6959: if anyone changes dispatch, the inner gates provide **zero** backstop — java's is inert and go's is absent. Ownership between the producers rests entirely on one layer.

## Not a blocker

Nothing here is a defect in your diff, and I am not asking for a change. Two follow-ups, both mine to carry:

- I will add this mechanism table to **#6957**. Its diagnosis is correct — the java producer is unreachable *because* `conf/routes` now classifies as `"scala"` and the dispatcher demands `"java"` — but the table records which layer actually decides that, which the issue does not currently say.
- **#6959** gets the "separation rests on dispatch alone" note, since it changes how that issue should be priced.

One thing I would take if you are inclined: a comment at `play_routes.go:145` and at `revel.go:110` naming the dispatcher as the thing that actually gates them. That is the fix for this defect class — name the guard next to the thing it guards — and it is two lines. Say the word and I will land it separately rather than hold this PR.

Merging once `test (windows-latest)` lands.

PR #6961, opened 2026-09-07
fix(#6940): the polling change-detector never converged on a path git reports and the walker refuses — observe the manifest's verdict instead of re-deriving four gates

### LINKED ISSUE #6940: poll mode never converges on a path git reports and the walker refuses via the extension, ignore or sparse gate — a reindex every interval, forever

cajasmota:
Closes #6940.

`walk.WalkRepo`'s file branch has **four** gates; the poller can cheaply reproduce **one** (`walk.IndexableEntryType`, the entry-type axis). A path git reports that clears that gate and is then refused by the extension filter, the file-level ignore layer or the sparse filter entered the candidate set, was stamped by nothing, and read as new on **every cycle, forever** — one full reindex enqueued per repo per interval. `photo.png` and a tracked-but-gitignored `gen.go` were both driven at 5/5 cycles in #6940.

## The mechanism, and why it needs no gate of its own

The three unreproducible gates were the reason this was left open: two of them (the inherited ignore stack, `opts.Sparse`) are not answerable from a single path. **They do not have to be.** The poller does not need to *predict* the walker's verdict — it can *observe* it.

`declineTracker` records a path as refused when all three hold:

1. the poller **submitted** a reindex request naming it,
2. an index pass has since **completed** — `Manifest.IndexedAt` is stamped by every `SaveManifestAtCommit`, whether or not the file set changed, so it is proof a pass *ran*, not merely that something changed,
3. the manifest **still does not hold** the path.

That is exactly "reported, offered to the indexer, and not stamped". It is gate-agnostic — it converges the extension filter, the ignore layer, the sparse filter and any gate added later — and it **duplicates no predicate**, so there is no second copy to drift out of agreement with the walker. That was #6940's option 4 and the mistake MA-1 already made once. `existsOnDisk`'s doc comment keeps promising exactly the one axis it decides; nothing was promised away.

Alternatives priced and rejected: re-walking per discovered path (option 1) pays the walk this poller exists to avoid, every cycle, for every such path; `git check-ignore` (option 3) closes one of the three gates and adds a process per cycle.

## Both directions are graded (#6902)

An assertion that the loop stopped cannot see a change that is now **missed** — the mirror defect is blocker 1 of #6935 (a path the indexer *will* stamp being refused by the poller).

- **A decline needs a completed pass.** A file whose reindex was merely debounced or circuit-broken is re-reported every cycle, level-triggered, and never dropped.
- **A decline is forgiven** the moment the manifest holds the path, and the forgiveness chain is tested end to end rather than stipulated: un-ignore `gen.go` → the `.gitignore` edit is *itself* a manifest key → the poller reports it → the reindex re-walks the repo → `gen.go` is stamped → the decline is dropped and the next edit is detected.
- **Declines are per repo** and die with `RemoveRepo`; a re-registered repo re-learns them.
- Nothing is persisted: a daemon restart costs one extra cycle per refused path. The set is capped (`maxDeclinedPaths`), past which the poller **over-fires** rather than dropping a path — the direction this file chooses everywhere.

## Ordering, so the two guards are not mutually masking

Forgiveness runs **before** promotion. Promote-then-forgive lets the promotion loop refuse a freshly stamped path and the forgiveness loop erase the evidence in the same call: two guards that only ever fail together, so neither is graded. Forgive-then-promote leaves a stamped path visibly refused. The mutant that deletes promotion's stamped-check was **ALIVE** under the original order and is **DEAD** under this one — the reorder is the grading.

## Tests — convergence as convergence, per gate

`internal/daemon/watch/poll_convergence_6940_test.go`, all through the real poller, real git, real `diff` manifest round-trip, with an index pass between cycles. Every case asserts a **bounded number of cycles** after which the path stops being reported, via the existing `cpConverges` helper — not first-cycle detection, which is what hid the two earlier instances.

| case | gate |
|---|---|
| `ExtensionFilteredPathConverges` | 1 — `shouldSkipFileByExt` (`photo.png`) |
| `TrackedButGitignoredPathConverges` | 3 — `igStack.MatchFile` (`gen.go`, tracked via `add -f`) |
| `SparseFilteredPathConverges` | 4 — sparse filter, via an untracked subtree outside the pattern set |
| `NewIndexableFileIsNeverDeclined` | mirror: indexed file stays detectable after an edit |
| `NoDeclineWithoutACompletedIndexPass` | mirror: 3 cycles, no pass, still reported, zero declines |
| `DeclineIsForgivenWhenTheIndexerChangesItsMind` | mirror: un-ignore → stamped → detected again |
| `DeclinesAreScopedToOneRepo` / `RemoveRepoDropsTheDeclineSet` | scope and lifetime |

Each case pins its own premise first (git reports it; the walker refuses it; the manifest does not hold it) so it cannot go vacuous if a gate moves. The sparse case **fails rather than skips** if the path stops reaching the candidate set — that state is "untested", not "unreachable".

**Two adjacent gaps found while making the sparse case reachable, both real and neither #6940** — recorded here, to be filed separately:

- `gitmeta.ProbeRepo` reads `git config --local`, but git ≥ 2.32 writes `core.sparseCheckout` into the **worktree-scoped** config (`.git/config.worktree`). A sparse checkout created by a current git is invisible to the probe. The test mirrors the flag into `--local` to keep the gate reachable, and says why.
- `gitmeta.IsPathIncluded` treats a bare `*` pattern as match-all, and git's **cone** patterns are `/*` + `!/*/` + `/src/` — so a cone checkout makes grafel's own sparse filter include everything. The test uses `--no-cone` deliberately, and says why.

## Verification

- `go test -count=1` green on `./internal/daemon/watch/` (159.9 s), `./internal/daemon/` (31.3 s), `./cmd/grafel/` (156.6 s). macOS/APFS.
- **Mutants — 6 scored, 6 DEAD**, each edit proved landed by an exact occurrence count against a shasum-verified backup, with a green baseline immediately before:
  1. drop the decline guard on git-reported files → DEAD (4 tests, incl. both repro shapes)
  2. decline without requiring a completed pass (the permissive direction) → DEAD (`NoDeclineWithoutACompletedIndexPass`)
  3. delete forgiveness → DEAD (`DeclineIsForgivenWhenTheIndexerChangesItsMind`)
  4. promotion ignores whether the manifest stamped the path → **ALIVE under promote-then-forgive, DEAD after the reorder** (`NewIndexableFileIsNeverDeclined`)
  5. one tracker shared across repos → DEAD (scope + lifetime)
  6. drop the decline guard on subtree-walked paths → DEAD (`SparseFilteredPathConverges`)

Nothing here is `runtime.GOOS`-conditional; `ci:full` is added for the macOS leg, which otherwise never runs these tests. All measurement is macOS/APFS — the target is Linux containers on overlayfs and nothing has been measured there.

## Scope

Poll mode only. fsnotify mode is untouched (its sink is repo-level, so it has no candidate set to poison), `PollingEnabled()` is untouched, and `auto` still resolves to fsnotify — this unblocks #6932 arm B part 2 rather than being part of it.

---

## Review round 2 — all four asks addressed (`281210545`)

1. **M1b — condition (1), first half.** `TestChangePoller_RecreatedManifestKeyIsNeverRefused`: delete a manifest key → pass prunes it → recreate it → must be reported, with `cpDeclineCount == 0`. Mutant (pending recorded from `changed ∪ discovered` — the naive `changed`-only form does not compile, as the reviewer also found) is **DEAD** on exactly that test.
2. **M7 — condition (1), second half: ALIVE and UNREACHABLE, reported as such rather than graded away.** Every discovered path is absent from the manifest, and `diff.isChanged` returns true unconditionally for a path the manifest lacks, so `discovered ⊆ changed` always and a cycle with a non-empty discovered set always submits. There is no input that reaches the guard today. `TestChangePoller_EveryDiscoveredPathIsSubmitted` pins that premise (both discovery shapes) and asserts the reachable half — a quiescent cycle records no pending. If `Filter` ever gains a way to drop a discovered path, that test fails first, which is the moment the guard becomes load-bearing.
3. **The cap — bounded, not re-worded.** `pending` is now capped too: it drains *only* on a completed pass, so it was unbounded in exactly the debounced / circuit-broken case this design is built around. `TestDeclineTracker_CapOverflowOverFires` asserts the **direction** (past the cap, paths keep re-reporting rather than being dropped) and drives the two caps in **separate rounds** — with `pending` capped, a promotion loop ignoring its own cap never sees more than the cap in one pass, so scoring them together grades neither. Mutants **M6 (refused cap) DEAD, M6b (pending cap) DEAD**.
4. **Sparse `t.Skip` → `t.Fatal`.** Scored the reviewer's own mutation: deleting the two `config --local` mirror lines gave `SKIP … PASS, exit 0` before; it is now **DEAD** (`--- FAIL`), so the only sparse leg cannot vanish unseen on a `-v`-less CI run.
5. **Comments narrowed, no behaviour change.** Forgiveness names only the trigger that self-triggers (`.gitignore` is itself a manifest key); the sparse-cone case is recorded as what it is — no manifest key changes, the declined path is suppressed, so it heals on an unrelated pass or a daemon restart, a real bounded behaviour change against pre-#6940 — and the extension-list case heals by construction, since growing it means an upgrade that restarts the daemon and drops the unpersisted set.

**Recorded, not fixed here** (being filed separately): production convergence turns on the manifest save at `incremental.go:716`, whose own comment invites its deletion — and deleting it resurrects #6940 with every test in this PR still green, because `cpIndexPass` performs the save itself. Also confirmed by the reviewer against git 2.50.1: the two `gitmeta` sparse gaps above are one issue — grafel's sparse support is inert on any git ≥ 2.32 and would over-include in cone mode if the probe ever fired.

Verification for this round: `go test -count=1` green on `./internal/daemon/watch/` (156.2 s), `./internal/daemon/` (30.9 s), `./cmd/grafel/` (148.5 s), macOS/APFS.

---

## Review round 3 — MP-2 (`e425573d7`)

**Both halves of the ask, not either/or.** `noteSubmitted` keeping the FIRST submission time was documented and unobserved, and it is only *safe* because of a call ordering nothing wrote down — `pollRepo` reconciles against the manifest **before** the caller submits, so at most one `noteSubmitted` lands between any two `reconcile`s. The comment now names that ordering as the dependency it is, **and** `TestDeclineTracker_KeepsTheFirstSubmissionTime` pins the sentence directly, asserting the stored value and its consequence separately (they are different claims): `pending` still holds `T1` after a second submission at `T2`, and a manifest stamped **between** `T1` and `T2` is proof of a completed pass and must decline. **MP-2 is DEAD** — RED on the overwriting tracker, GREEN restored from a SHA-256-verified `cp`.

**Two mutually-masking pairs came out of one design, and that is a property of the shape.** First promotion/forgiveness (`M3`/`M4` — fixed by reordering to forgive-then-promote), then the two caps (`M6` came back ALIVE because the `pending` cap meant a promotion loop ignoring its own cap never saw more than the cap in one pass — fixed by driving them in separate rounds). Anyone adding a guard to `declineTracker` should assume the same and **score compound guards part by part**; a single mutant over both halves grades neither.

Verification for this round: `go test -count=1` green on `./internal/daemon/watch/` (161.0 s), `./internal/daemon/` (31.8 s), `./cmd/grafel/` (148.7 s), macOS/APFS.

### DIFF
--- a/internal/daemon/watch/change_poller.go
+++ b/internal/daemon/watch/change_poller.go
@@ -447,13 +447,21 @@ func (p *ChangePoller) pollRepo(abs string) (changedOut, discoveredOut []string)
 	return changed, discovered
 }
 
-// maxDeclinedPaths caps the per-repo decline set. The set is bounded in
-// practice by what `git status -unormal` reports as FILES (an untracked subtree
-// arrives as one collapsed `dir/` line), but a cap keeps a pathological repo
-// from turning a convergence fix into unbounded daemon memory. Past the cap the
-// poller stops recording and reverts to the pre-#6940 behaviour for the
-// overflow — it OVER-fires rather than dropping a path, which is the direction
-// this file chooses everywhere else.
+// maxDeclinedPaths caps BOTH per-repo maps a decline passes through — pending
+// and refused. Both sets are bounded in practice by what `git status -unormal`
+// reports as FILES (an untracked subtree arrives as one collapsed `dir/` line),
+// but a cap keeps a pathological repo from turning a convergence fix into
+// unbounded daemon memory.
+//
+// pending is capped for a specific reason (#6961 review): it drains ONLY on a
+// completed index pass, so it is exactly the debounced / circuit-broken repo —
+// the scenario condition (2) is built around — where it would otherwise grow
+// without limit while nothing ever removed an entry.
+//
+// Past the cap the poller stops recording and reverts to the pre-#6940
+// behaviour for the overflow — it OVER-fires rather than dropping a path, which
+// is the direction this file chooses everywhere else. That direction is
+// asserted, not asserted-about: see TestDeclineTracker_CapOverflowOverFires.
 const maxDeclinedPaths = 4096
 
 // declineTracker is the #6940 convergence state for one repo.
@@ -519,6 +527,9 @@ func (t *declineTracker) noteSubmitted(discovered []string, at time.Time) {
 		if _, ok := t.pending[rel]; ok {
 			continue // keep the FIRST submission time
 		}
+		if len(t.pending) >= maxDeclinedPaths {
+			continue // overflow: this path keeps re-reporting, as it did before #6940
+		}
 		t.pending[rel] = at
 	}
 }
@@ -532,9 +543,25 @@ func (t *declineTracker) reconcile(m *diff.Manifest, logger *slog.Logger, abs st
 
 	// Forgiveness FIRST. A refusal is an observation about one past index
 	// pass, not a permanent verdict: once the manifest holds the path the
-	// indexer has plainly changed its mind (the .gitignore rule was dropped,
-	// the sparse cone widened, the extension list grew), and half 1 sweeps it
-	// from here on anyway.
+	// indexer has plainly changed its mind, and half 1 sweeps it from here on
+	// anyway.
+	//
+	// Only ONE of the ways the indexer can change its mind SELF-TRIGGERS, and
+	// it is the one driven end to end by
+	// TestChangePoller_DeclineIsForgivenWhenTheIndexerChangesItsMind: dropping
+	// a .gitignore rule, because the .gitignore is itself a manifest key, so
+	// half 1 reports the edit and the reindex it asks for re-walks the repo.
+	// The other two do NOT (#6961 review):
+	//
+	//   - Widening a sparse cone edits .git/info/sparse-checkout, which is no
+	//     manifest key, and the newly included path is suppressed — so it
+	//     cannot request the pass that would forgive it. It stays invisible
+	//     until an unrelated change triggers a stamping pass, or the daemon
+	//     restarts. That is bounded and self-healing, but it IS a behaviour
+	//     change against pre-#6940, where the next cycle picked it up.
+	//   - Growing the indexed-extension list means upgrading grafel, which
+	//     restarts the daemon — and nothing here is persisted, so the whole
+	//     decline set is gone. That one heals by construction.
 	//
 	// The order is deliberate and it is what makes the two "is it stamped?"
 	// tests in this function INDEPENDENTLY graded. Promote-then-forgive would
--- a/internal/daemon/watch/change_poller.go
+++ b/internal/daemon/watch/change_poller.go
@@ -65,6 +65,22 @@
 // bounded number of cycles, not first-cycle detection, because first-cycle
 // detection is exactly what hid the bug.
 //
+// CONVERGENCE ON A PATH THE WALKER REFUSES (#6940). WalkRepo's file branch has
+// FOUR gates — extension filter, entry-type gate, file-level ignore layer,
+// sparse filter — and a single path is not enough to evaluate two of them (the
+// ignore layer needs the stack the walk accumulates while descending; sparse
+// needs opts.Sparse). So the poller cannot ask "would the indexer stamp this?"
+// by re-deriving the gates, and the first version of this file looped forever
+// on an untracked `photo.png` or a tracked-but-gitignored `gen.go`.
+//
+// It does not have to ask. The MANIFEST already answers it, after the fact: a
+// path grafel reported, that an index pass has since completed WITHOUT
+// stamping, is by definition a path the indexer declined — whichever gate did
+// it, including gates that do not exist yet. See declineTracker: no gate is
+// duplicated, so there is no second copy of any predicate to drift out of
+// agreement with the walker (the mistake #6940 lists as option 4, and the one
+// MA-1 already made once).
+//
 // WARM-UP (arm D of #6932). A fresh worktree's first `git status` costs
 // 2.4-9.0 s until git's untracked cache exists; every subsequent one is ~60 ms.
 // AddRepo therefore sets core.untrackedCache=true and runs one throwaway
@@ -148,6 +164,10 @@ type ChangePoller struct {
 
 	mu    sync.Mutex
 	repos map[string]struct{}
+	// declines holds one declineTracker per repo — the #6940 convergence
+	// state. It is keyed by the same absolute path as repos and dropped by
+	// RemoveRepo.
+	declines map[string]*declineTracker
 
 	cycles  uint64 // atomic — completed poll cycles
 	submits uint64 // atomic — sink invocations
@@ -176,6 +196,7 @@ func NewChangePoller(cfg ChangePollerConfig, sink EventSink, logger *slog.Logger
 		sink:          sink,
 		logger:        logger,
 		repos:         make(map[string]struct{}),
+		declines:      make(map[string]*declineTracker),
 		stopCh:        make(chan struct{}),
 		doneCh:        make(chan struct{}),
 	}
@@ -226,6 +247,7 @@ func (p *ChangePoller) RemoveRepo(repoPath string) {
 	}
 	p.mu.Lock()
 	delete(p.repos, abs)
+	delete(p.declines, abs)
 	p.mu.Unlock()
 }
 
@@ -297,7 +319,7 @@ func (p *ChangePoller) PollOnce() map[string][]string {
 
 	out := make(map[string][]string, len(repos))
 	for _, abs := range repos {
-		changed := p.pollRepo(abs)
+		changed, discovered := p.pollRepo(abs)
 		if len(changed) == 0 {
 			continue
 		}
@@ -307,6 +329,13 @@ func (p *ChangePoller) PollOnce() map[string][]string {
 		p.logger.Info("change-poller: changes detected — enqueuing reindex",
 			"repo", abs, "changed", len(changed), "bulk", bulk)
 		p.sink(abs, bulk)
+		// Only paths we ACTUALLY asked for a reindex of become pending, and
+		// only after the request went out: "the indexer declined it" is a
+		// claim about a pass that had the chance to stamp it. Recording
+		// pending on a cycle that submitted nothing would let an unrelated
+		// index pass decline a path grafel never asked about — the mirror
+		// defect (#6902), a change silently missed instead of over-reported.
+		p.trackerFor(abs).noteSubmitted(discovered, time.Now().UTC())
 	}
 	return out
 }
@@ -314,13 +343,17 @@ func (p *ChangePoller) PollOnce() map[string][]string {
 // pollRepo computes the changed set for one repo. Both halves of hybrid B live
 // here: git for DISCOVERY of paths the manifest has never heard of, and the
 // manifest's own key set for the CHANGE DECISION.
-func (p *ChangePoller) pollRepo(abs string) []string {
+//
+// It also returns the DISCOVERED paths — the candidates that are not manifest
+// keys. PollOnce hands those to the decline tracker once it has actually
+// submitted them (#6940).
+func (p *ChangePoller) pollRepo(abs string) (changedOut, discoveredOut []string) {
 	if p.stateDir == nil {
-		return nil
+		return nil, nil
 	}
 	stateDir := p.stateDir(abs)
 	if stateDir == "" {
-		return nil
+		return nil, nil
 	}
 	m := diff.LoadManifest(stateDir)
 	if len(m.Files) == 0 {
@@ -328,8 +361,15 @@ func (p *ChangePoller) pollRepo(abs string) []string {
 		// unreadable). Every file would read as new and the poller would ask
 		// for a reindex on every cycle forever. Initial indexing belongs to the
 		// scheduler, not here.
-		return nil
+		return nil, nil
 	}
+
+	// #6940: reconcile the decline tracker against the manifest we just read,
+	// BEFORE the candidate set is built. This is where "reported, and an index
+	// pass has since declined to stamp it" is decided, and where a path the
+	// indexer has since changed its mind about is let back in.
+	tr := p.trackerFor(abs)
+	tr.reconcile(m, p.logger, abs)
 
 	cand := make(map[string]struct{}, len(m.Files)+16)
 	// Half 1 — the manifest's key set. This is what converges edit-then-revert
@@ -358,14 +398,40 @@ func (p *ChangePoller) pollRepo(abs string) []string {
 	if !ok {
 		p.logger.Warn("change-poller: git status failed — sweeping the manifest key set only", "repo", abs)
 	}
+	//
+	// A path the tracker has recorded as DECLINED is not taken either: an
+	// index pass has already had it in front of it and did not stamp it, so
+	// it can only read as new for as long as the poller keeps offering it.
+	// That is #6940, and it is the same non-convergence as the deletion case
+	// entering through a different gate.
+	var discovered []string
 	for _, f := range files {
-		if _, known := cand[f]; known || existsOnDisk(abs, f) {
+		if _, known := cand[f]; known {
+			continue
+		}
+		if tr.declined(f) {
+			continue
+		}
+		if existsOnDisk(abs, f) {
 			cand[f] = struct{}{}
+			discovered = append(discovered, f)
 		}
 	}
 	for _, d := range dirs {
 		for _, rel := range walkUntrackedSubtree(abs, d) {
+			if _, known := cand[rel]; known {
+				continue
+			}
+			// The subtree walk applies the gates it can see, but it is rooted
+			// BELOW the repo root, so it does not inherit the root ignore
+			// stack and knows nothing of sparse state — it can hand back a
+			// path the real, root-rooted index walk then refuses. Same
+			// tracker, same reason.
+			if tr.declined(rel) {
+				continue
+			}
 			cand[rel] = struct{}{}
+			discovered = append(discovered, rel)
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


cajasmota:
## Coordinator mutant on `c90e0088d`

Scored in my own detached worktree at the PR head, `git status --porcelain` empty, restored by `cp` from a shasum-verified backup (`5e020db7…` both ways). Baseline verified green immediately before (`./internal/daemon/watch` 152.7 s, exit 0).

I deliberately picked a target outside everything the independent reviewer was briefed to attack (dropping conditions wholesale, first-cycle declines, never-forgive, the cap, the `IndexedAt`-is-always-stamped premise), so this is one nobody else ran.

**MP-1 — weaken the *boundary* of condition (2) rather than removing it.**

`change_poller.go:557` — `if !m.IndexedAt.After(submittedAt)` → `if m.IndexedAt.Before(submittedAt)`. Strict "strictly after" becomes ">=", so a manifest stamped at *exactly* the submission instant now counts as a completed pass. Edit proven landed: `pre=1`, `post_old=0`, `post_new=1`. `go vet` clean.

**Result: ALIVE** (`-count=1`, exit 0, 168.0 s).

## Verdict: distinguishable and ungraded — and cheap to grade

Production reachability is essentially nil, and I checked rather than assumed: `IndexedAt` is `time.Time` with `json:"indexed_at"` (`diff.go:118`), stamped by `time.Now().UTC()` (`:180`), and Go marshals it as RFC3339 **Nano**, so no precision is lost across the disk round-trip. `submittedAt` is an independent `time.Now().UTC()` at `:338`. Two independent nanosecond-resolution clock reads landing on the same instant is not a thing that happens, and nothing derives one from the other or truncates either.

So I am **not** claiming a live defect. What I am claiming is that the strict `After` is a **deliberate correctness choice** — "a pass that completed strictly after I asked for it" — carrying the whole weight of condition (2), and **nothing in the suite would notice if it were relaxed.**

Normally that is where I would record "too expensive to kill, equivalent under the current suite" and move on. My own standing rule is that *"too expensive" is a claim, not evidence*, so I priced it — and the seam already exists:

```go
func (t *declineTracker) noteSubmitted(discovered []string, at time.Time)   // :512
func (t *declineTracker) reconcile(m *diff.Manifest, ...)                   // :304
```

Both take their time source **as a parameter**. A test can build a tracker, `noteSubmitted(paths, T)`, then `reconcile(&diff.Manifest{IndexedAt: T})` with the two exactly equal and assert **zero declines** — no clock injection, no filesystem, no fake time package. That is on the order of a dozen lines against functions your test file already exercises directly.

**Please add it.** Not because the boundary will drift on its own, but because this PR's entire safety argument is "a decline requires proof that a pass actually ran", and the `>` that makes it *proof* rather than *coincidence* is the one part of that argument nothing observes. A future refactor that "simplifies" the comparison would be green.

## Why this direction specifically

Every serious defect this milestone came from a mutant that made a predicate MORE permissive, and suppression mechanisms are the worst case of it: a decline that should not have happened is invisible by construction — the path simply stops being reported and the graph quietly goes stale. Your own assertions cannot see it, which is exactly why the boundary wants a positive test rather than a comment.

I note you already caught one of these yourself — the promotion mutant that came back ALIVE because the forgiveness loop masked it, which you fixed by reordering `reconcile` to forgive-then-promote so the two guards fail independently. That was the right call and it is the same defect class as this one, one level down.

## Status

CI is fully green on `c90e0088d` (the skipped `python ratchet tests` is path-filtered and this PR is Go-only). **Holding the merge for the independent review**, which is still running. Once it lands I will merge on green with this one addition, unless the reviewer surfaces something larger.

cajasmota:
## Independent adversarial review — **REQUEST-CHANGES** (two ungraded guards + one prose overclaim; no defect in shipped behaviour)

Reviewed `c90e0088d` in my own detached worktree at
`/private/tmp/claude-501/-Users-jorgecajas-Projects-archigraph/c0fefed4-0195-4c81-9a94-07a7145710a4/scratchpad/review-6961-qz9/wt`
(and a second read-only checkout `.../review-6961-qz9/aux` for the gitmeta probes). Every number below is
**macOS 25.6 / APFS, go test -count=1, git 2.50.1 (Apple Git-155)**. Nothing was measured on Linux/overlayfs.

The mechanism is right and the four guards that carry it are all DEAD. What I am asking for is small: the
one condition the PR names first — *"the poller **submitted** a reindex naming it"* — is ungraded in **both**
of its halves, and one of them is a reachable silent miss in exactly the direction #6902 says an
"it converged" assertion cannot see. That makes "**Both directions are graded**" the one claim in the body
I could not reproduce.

### Mutant table — 11 scored (9 in the final serial campaign + 2 aborted/recovered)

Each edit proved landed by an **exact occurrence count** against a `shasum -a 256`-verified backup, restored
by `cp` (never `git checkout --`), with a **green full-package baseline run immediately before every score**.
Baseline: `ok internal/daemon/watch 150.8s`. Driver + all 18 run logs kept.

| # | mutation (permissive direction first) | verdict | killed by |
|---|---|---|---|
| M2 | drop condition (2) — decline with no completed pass | **DEAD** | `NoDeclineWithoutACompletedIndexPass` |
| M3 | drop condition (3) — promotion ignores `m.Files[rel]` | **DEAD** | `NewIndexableFileIsNeverDeclined`, `DeclinesAreScopedToOneRepo` |
| M4 | never forgive (delete the forgiveness loop) | **DEAD** | `DeclineIsForgivenWhenTheIndexerChangesItsMind` (**only** that test) |
| M5 | decline on the first cycle (`pending[rel] = time.Time{}`) | **DEAD** | `NoDeclineWithoutACompletedIndexPass` |
| M6 | ignore the cap (`if false && len(t.refused) >= maxDeclinedPaths`) | **ALIVE** — equivalent under the current suite | — (zero tests name `maxDeclinedPaths`) |
| M1b | pending recorded from `changed`, not `discovered` | **ALIVE — distinguishable and ungraded** | see below |
| M7 | pending recorded on a cycle that submitted nothing | **ALIVE — distinguishable and ungraded** | see below |
| M9 | reorder only: promote-then-forgive, nothing else | **ALIVE** — behaviour-preserving (see "the reorder") | — |
| M8 | promote-then-forgive **+** M3 | **ALIVE w.r.t. the poller suite** — reproduces your finding | only the flaky `TestInterleavedDirectoryFills…` fired |
| M1 | `noteSubmitted(changed…)` naive form | DID-NOT-COMPILE (`discovered` unused) — re-run as M1b | — |
| — | first campaign aborted: baseline went red under my own concurrent `go test` | recovered, re-run serially | — |

### The two ungraded halves of condition (1)

**M1b — recording *manifest keys* as pending is a reachable missed change.** Distinguishing input, no test has it:

1. `alpha.go` is a manifest key; delete it. The poller reports it (`DeletionConverges` covers this far).
2. The pass prunes it from the manifest and advances `IndexedAt`. Under M1b `alpha.go` is now **refused**.
3. Recreate `alpha.go`. It is no longer a manifest key, so it can only arrive through discovery — and
   discovery skips refused paths. **The recreated file is silently never indexed** until some unrelated
   change triggers a pass that stamps it (or the daemon restarts).

The shipped code is correct here *because* `discovered` excludes anything already in `cand` — but nothing in
the suite fails when that restriction is removed. The missing assertion is ~12 lines: delete → pass →
recreate → assert reported, and assert `cpDeclineCount == 0`.

**M7 — the "only after the request went out" guard is ungraded too.** `change_poller.go:332-338` justifies it
by name ("would let an unrelated index pass decline a path grafel never asked about — the mirror defect").
Moving `noteSubmitted` above the `len(changed) == 0` guard passes the whole package. It is reachable only
through a race (a path `git status` reports and `diff.Filter` then does not return — e.g. a build artefact
deleted between the two), so it is narrow; but the comment states it as a designed guard and nothing holds it.

**M6 — the cap's *direction* is prose.** `maxDeclinedPaths` appears in **zero** test files, so
"past the cap the poller **over-fires** rather than dropping a path" is unobserved in either direction.
Related: the cap comment says it "keeps a pathological repo from turning a convergence fix into unbounded
daemon memory", but only `refused` is capped. **`pending` is unbounded**, and it is drained *only* by a
completed pass — i.e. it grows without limit in precisely the debounced/circuit-broken scenario this design
is built around. Low severity, one-line fix, but the comment currently claims more than the code does.

### The `IndexedAt` premise — **holds, by enumeration, and it is observed**

- `IndexedAt` has exactly **one** assignment site in the whole tree: `internal/indexer/diff/diff.go:180`,
  unconditional, first statement of `SaveManifestAtCommit`. `SaveManifest` (diff.go:157) delegates to it.
  There is no second writer: the only other way a `file-index.json` reaches disk is the **byte copy** in
  `worktree_seed.go:290-310`, which carries the source manifest's own (older) stamp rather than minting one.
- More importantly for condition (3): **every** production save site is preceded by a full repo walk feeding
  `allFiles` — `incremental.go:643` (too-many-changed reject), `:716` (zero-change sweep), `:1761` (normal
  pass), `cmd/grafel/index.go:1389` (full index). So "the manifest still lacks the path" really is the walk's
  verdict and not an artefact of a partial write. A conditional stamp ("only when the file set changed")
  would be caught: the pass between cycles in `ExtensionFilteredPathConverges` changes no manifest entry.
- **The one thing I would flag.** In production the convergence turns on the manifest save at
  `incremental.go:716` — the zero-change pass — and that call's own comment invites its deletion
  ("#6199 lists the sweep on a zero-change pass as pure waste and the cheapest remaining win… this is why the
  deletion is still worth doing. Just replace the healing when you do: sweep here but skip the SaveManifest").
  Take that advice and **#6940 comes straight back, with every test in this PR still green**, because
  `cpIndexPass` performs the save itself — no test drives the real incremental pass through the poller.
  One sentence at `incremental.go:716` naming #6940 as a second reason that save is load-bearing would close it.

### The reorder — correct on its own merits, and it is genuinely what grades M3

- **M9 (reorder alone) is ALIVE**, and that is the expected answer: `refused ∩ pending = ∅` always
  (promotion deletes from `pending` when it inserts into `refused`, and `declined()` keeps a refused path out
  of `discovered`, so it can never be re-added to `pending`). The two loops therefore **commute** on correct
  code — forgive-then-promote is behaviour-preserving, not a behaviour change, and the "forgiven and
  re-promoted within one `reconcile`" case cannot arise. Nothing the old order handled is lost.
- **M8 confirms your claim for the stated reason**: promote-then-forgive **plus** the M3 deletion leaves the
  poller suite entirely green (its only failure was the flaky test below), while M3 alone under the shipped
  order dies on `NewIndexableFileIsNeverDeclined`. So the reorder is what makes the guard observable, and it
  is not "a different assertion running first".

### Claims 1-5

1. **`cpConverges` asserts a bound**, not first-cycle detection — `t.Fatalf` on the last iteration
   (`change_poller_6932_test.go:615-628`); the #6940 cases require `n == 0` after the declining pass. Confirmed.
2. **The sparse premise pin is real for the axis it claims — but the leg can still vanish silently.** The
   `t.Fatalf` on "did not surface `outside/gen.go`" is genuine, and the test **PASSES (does not skip) here**.
   However the two earlier `t.Skip`s are silent, and they are load-bearing: I removed only the two
   `config --local core.sparseCheckout*` mirror lines (exact anchor count 1, restored by `cp`, shasum
   verified) and the test came back `--- SKIP … PASS, exit 0`. CI runs without `-v`, so a skip is invisible
   there. Since the mirror makes the outcome deterministic, that second `t.Skip` should be a `t.Fatal` too.
3. **Debounced/circuit-broken never drops a path** — reproduced: 3 cycles, 3 submissions, 0 declines,
   still reported (`NoDeclineWithoutACompletedIndexPass`, run verbosely).
4. **`SkipEntry` cannot serve as the decline signal — confirmed, with one wording nit.** Gate 1
   (`walker.go:266`) and gate 4 (`walker.go:329`) `return nil` with **no** `SkipEntry` appended; only gate 3
   emits one. So it can answer at most one of the three gates. The nit: the poller *does* receive `SkipEntry`
   and discards it at `change_poller.go:683` (`found, _, err := walk.WalkRepo(sub, nil)`), so "not reachable"
   is loosely worded — the substantive conclusion stands.
5. **No predicate is duplicated — confirmed by independent grep.** The poller's only gate call is
   `walk.IndexableEntryType` (`change_poller.go:617`), the walker's own function. `shouldSkipFileByExt` has
   exactly one non-test caller (`walker.go:266`) and `gitmeta.IsPathIncluded` exactly one (`walker.go:329`).

### Forgiveness — reachable, but only one of the three triggers your comment names is

`reconcile`'s comment offers three: "the `.gitignore` rule was dropped, **the sparse cone widened**, **the
extension list grew**". Only the first self-triggers, and that is why it is the one you could drive end to
end: the `.gitignore` is itself a manifest key. Widening a sparse cone changes **no manifest key** (it edits
`.git/info/sparse-checkout`), and the declined path is suppressed, so it cannot request the pass that would
forgive it — the newly-included file stays invisible until an unrelated change triggers a pass, or the daemon
restarts. Before this PR it was picked up on the next cycle. Bounded and self-healing, but it is a real
behaviour change in the suppression direction and the comment reads as if all three were symmetric. (The
"extension list grew" case does self-heal, because growing it means upgrading grafel, which restarts the
daemon and drops the unpersisted set — worth saying explicitly.) Also: **M4 is killed by exactly one test**,
the one written for it; forgiveness has no second grader.

### Both adjacent gaps CONFIRMED — independently, against **git 2.50.1 (Apple Git-155)**

Driven through the real `gitmeta` API in the aux checkout:

- **`ProbeRepo` misses a modern sparse checkout.** After `git sparse-checkout set src`,
  `git config --local --get core.sparseCheckout` **exits 1 with no output**; the value is in
  `.git/config.worktree` (`extensions.worktreeConfig=true`). `ProbeRepo` returns `IsSparse=false`,
  `patterns=[]`. Same for `--no-cone`. So on any git ≥ 2.32 grafel's sparse filter never engages at all.
- **Cone patterns include everything.** The pattern file git writes is exactly `/*`, `!/*/`, `/src/`.
  `IsPathIncluded` normalises the first line to `*`, treats it as match-all and returns `true` before it
  ever reaches the negation — `out/b.go`, `anything/at/all.go` and `node_modules/x.go` were all reported
  **included**. (Non-cone `src` behaves correctly: `out/b.go` false, `src/a.go` true.)

Together these are one issue, not two: **grafel's sparse-checkout support is inert on current git, and
would over-include if the probe ever did fire in cone mode.** Happy for it to be filed on this confirmation.

### One environmental observation, not attributable to this PR

`TestInterleavedDirectoryFillsAcrossReposAreNotDoubleCharged` (`fdbudget_6268_test.go:1290`, fsnotify
fd-ledger, untouched by this diff) failed in **2 of 18** clean serial runs here, plus once more under CPU
contention: `ledger read 109, want it drained to at most 60`. I did not bisect it to `main`, so I claim only
that it is unrelated to the changed code — but it is load-sensitive enough to poison a "green baseline"
claim on a busy machine, which is how my first campaign aborted.

### What I am asking for

1. Grade condition (1): the delete → pass → recreate case (kills M1b), and an assertion that a cycle which
   submits nothing records no pending (kills M7).
2. Make the second `t.Skip` in the sparse case a `t.Fatal`, or say in the body that the leg may be silently
   absent on CI.
3. Narrow the forgiveness comment to the trigger that self-triggers, and note the restart path for the other two.

Nice-to-have, not blocking: cap `pending` as well as `refused` (or correct the cap comment), and add the
#6940 cross-reference at `incremental.go:716`.

*Not merged, nothing pushed. Both review worktrees removed; `git status --porcelain` clean.*

cajasmota:
## Second coordinator mutant, on the final head (`281210545`)

My first mutant (MP-1, the `IndexedAt` equality boundary) was scored on `c90e0088d` and you killed it. The head has moved twice since and **gained new code** — the `pending` cap and its surrounding logic — so a kill on the old head is not a kill on this one. Scored again in a fresh detached worktree, `git status --porcelain` empty, restored by `cp` from a SHA-256-verified backup (`5bc66978…` both ways), baseline green immediately before (162.7 s).

**MP-2 — overwrite the pending timestamp instead of keeping the first.**

`change_poller.go:527-530`:

```go
if _, ok := t.pending[rel]; ok {
    continue // keep the FIRST submission time
}
```

I made the re-submission path overwrite with the latest `at`. Edit proven landed (`pre=1`, `post_old=0`, `post_new=1`), `go vet` clean.

**Result: ALIVE** — full `./internal/daemon/watch` green, 156.0 s, exit 0.

## Verdict: I could not build a distinguishing production trace, and I am not claiming there is none

This is the direction that matters — keeping the *first* time makes a pass more likely to count as "strictly after", so it is the choice that lets a decline happen at all. Overwriting makes declines harder and, carried far enough, reintroduces #6940.

Tracing it against the real call order, though, the two agree at every reconcile point I could construct. Within one cycle the order is fixed:

- `pollRepo` runs `tr.reconcile(m, …)` (`:372`)
- the caller then submits and calls `noteSubmitted(discovered, now)` (`:338`)

So there is **exactly one `noteSubmitted` between any two `reconcile` calls**, and at the instant each reconcile reads `pending[rel]`, the mutant's value is the one written by the *previous* cycle's submit while the shipped value is the first-ever submit. Those differ only from the third cycle onward, and only while a path stays pending with no completed pass in between — and in the traces I worked through, a pass that completes before a cycle's manifest read is already promoted by that same cycle's reconcile using the older timestamp, so the difference never becomes observable.

I want to be explicit that this is **reasoning, not a measurement**. I did not construct the interleaving, and "I could not build one" is a statement about my effort, not a proof of equivalence. The concurrency here is real — the indexer completes asynchronously against a poll loop — and that is exactly the shape where hand-tracing is least trustworthy.

## What I would like, and it is not a bug fix

**Pin the documented behaviour directly.** The comment says "keep the FIRST submission time"; nothing observes it. Both functions are directly callable, so this needs no production trace and no concurrency:

```
noteSubmitted(["a.go"], T1)
noteSubmitted(["a.go"], T2)   // T2 > T1
// assert the tracker still holds T1 for "a.go"
```

Roughly eight lines. It pins the sentence rather than a speculative failure, and it is the assertion that would fail the day the reconcile-before-submit ordering changes — which is the only thing currently making the choice safe. Today that ordering is an undocumented dependency of a documented behaviour, which is the fragile combination.

If you would rather record MP-2 as *equivalent under the current call ordering* with that reasoning in the comment, and skip the test, say so and I will merge on that — but then the comment should name the ordering as the reason, because "keep the FIRST" on its own does not explain why it is safe to keep it.

## The rest

Your four fixes all check out from here, and two are worth calling out:

- **M7 reported as unreachable rather than claimed as a kill** — `discovered ⊆ changed` always, so no input reaches the guard — and then grading the *premise* instead, so the test fails the moment `Filter` gains a way to drop a discovered path. That is the right handling of an unreachable mutant.
- **M6 first came back ALIVE because the pending cap masked the refused cap.** That is the second mutually-masking pair in this PR, after the promotion/forgiveness one. Splitting them into separate rounds is the fix; it is worth a line in the body that this design has now produced two of them, because it is a property of the shape rather than an accident.

Merging once CI is green on the final head.

cajasmota:
Heads-up for when this lands — #6964 is fixed in #6967, which changes what the two sparse mirror lines in `poll_convergence_6940_test.go` do:

- `config --local core.sparseCheckout true` — **now unnecessary**. `gitmeta.ProbeRepo` passes no scope flag, so it reads the worktree scope git actually writes. Harmless if left (the probe reads `--local` too), but it is the line that made a green sparse test not evidence of working sparse support.
- `config --local core.sparseCheckoutCone true` — **actively wrong**, please delete: the fixture uses `--no-cone` deliberately, and this line makes the probe report `ConeMode=true` for a non-cone checkout. It changes no verdict now (one matcher serves both modes after #6967), but it is a fixture whose label contradicts its content.
- The `--no-cone` choice and its comment can go too: cone patterns are handled correctly now, so this test can use either mode on its own merits.

`./internal/daemon/watch/` is green with the #6967 fix in place (139.1 s, `-count=1`, macOS/APFS, git 2.50.1) — but that was against `main`, which does not yet have this branch, so it is not a check of the interaction. Whoever merges second should re-run it.

The convergence axis this test pins is unaffected either way.

Issue #6962, opened 2026-09-07
A UTF-8 BOM defeats razor's (?m)^@using / ^@inject anchors — latent, not live: 0 of 105 corpus .razor files carry one


cajasmota:
Found during #6343 Arm 0. The finding is real; **its severity is lower than it was first reported to me, and the correction is the useful part of this issue.**

## The defect, structurally certain

`internal/extractors/razor/extractor.go`:

```go
reInject = regexp.MustCompile(`(?m)^@inject\s+(\S+)\s+(\S+)`)   // :56
reUsing  = regexp.MustCompile(`(?m)^@using\s+([\w\.]+)`)        // :60
```

With `(?m)`, `^` matches at start-of-text and after each `\n`. A UTF-8 BOM (`EF BB BF`) occupies bytes 0-2, so a file whose **first line** is `@using Foo.Bar` has that directive at byte 3, not byte 0 — `^@using` cannot match it. The directive is silently dropped; every later line is unaffected because `^` matches fine after `\n`.

**There is no BOM stripping anywhere in the read path.** Verified: zero matches for BOM handling under `internal/` (non-test), excluding `SBOM` substring false positives in `cross/manifest/`. So nothing upstream normalises it away.

## Severity — measured, and NOT what I was first told

It was reported to me as "a live defect in shipped support, since it also affects BOM-prefixed `.razor` files today". I measured the shipped side before filing:

| extension | files in corpora | with BOM | BOM **and** line 1 is `@using`/`@inject` |
|---|---|---|---|
| **`.razor`** (supported today) | 105 | **0** | **0** |
| `.cshtml` (unsupported) | 1,119 | 1,021 | 137 |

**Zero of 105 corpus `.razor` files carry a BOM**, so there is no measured instance of this firing in shipped support. The 137 affected files are all `.cshtml`, which grafel does not index at all — and per #6343's Arm 0 measurement, likely never will via this extractor.

So the honest classification is **latent, not live**: structurally certain if it occurs, zero measured occurrences in the supported extension. It is worth fixing because it is three lines, not because anything is currently broken.

Two caveats in the other direction, so this is not read as "cannot happen":
- **A zero measurement is corpus-relative.** These are Microsoft sample repos. Visual Studio does emit BOMs on Windows for some file templates, so a real Blazor project authored in VS is a plausible source of BOM-prefixed `.razor` that the corpus simply does not contain.
[...]

PR #6963, opened 2026-09-07
fix(#6370): razor emitted no hierarchy edge at all — `@inherits`/`@implements` appeared nowhere in the package, so "what extends this Blazor component" was empty

### LINKED ISSUE #6370: DONE (8/8): erlang, graphql, nim, pony, lisp, clojure, ocaml, razor all ship hierarchy edges — the nine-language silence audit is complete

cajasmota:
Closes #6370

Eighth and **last** arm of #6370. Seven shipped (erlang, graphql, nim, pony, lisp, clojure, ocaml); razor was blocked on #6343 and is unblocked by its Arm 0 verdict — `.cshtml` will not be routed into this package, so this arm handles `.razor` only and designs for one dialect.

`@inherits` → **EXTENDS**, `@implements` → **IMPLEMENTS**.

## The value, stated rather than implied

**Do not read this as a volume fix.** Measured 2026-09-07 on the read-only corpora at `/Users/jorgecajas/Projects/archigraph-corpora` (macOS 25.6.0 / APFS / arm64):

| | files | `^@inherits` | `^@implements` |
|---|---|---|---|
| `.razor` (all corpora) | **105** | **4 files / 4 occurrences** (all `LayoutComponentBase`) | **3 files / 3 occurrences** (2 `IDisposable`, 1 `IAsyncDisposable`) |
| `.cshtml`, aspnetcore-mvc (**not routed**, out of scope) | 480 | 3 files | 1 file |
| `.cshtml`, aspnetcore-docs-samples (**not routed**) | 639 | 0 | 0 |

That is the broadest ASP.NET sample I have: **7 occurrences over 105 routed files**, and 4 more in 1,119 files this package does not see. Zero occurrences carry leading whitespace; zero of the 105 `.razor` files carry a BOM (independently re-confirming #6343 Arm 0). `@inherits LayoutComponentBase` and `@implements IDisposable` are idiomatic Blazor and real applications plausibly carry more, but **that extrapolation is left unmade** — I could not find a broader routed sample to measure.

The justification is the **silence**, not the volume: before this, `@inherits` and `@implements` appeared **zero times** anywhere in `internal/extractors/razor/`, and razor is absent from `cross/hierarchy`'s `supportedLanguages`. "What extends this Blazor component" returned empty, which is indistinguishable from "nothing does" — the #6321 class.

**No orphan figure is quoted anywhere in this arm as evidence an edge is real.** Per #6947, `analytics/scan.go` keys `touched[from]` unconditionally, so a dangling edge de-orphans its source and the metric would have improved whether or not these edges bound. The FROM/TO endpoints below are the evidence instead.

## Where the edges come from, and why not `supportedLanguages`

Emitted from `internal/extractors/razor/hierarchy.go`, modelled on `internal/extractors/groovy/hierarchy.go` (**not** csharp's). The cross pass `addEntity`s a `SCOPE.Component` for the type *and* for each parent; this package already emits exactly one component per file, so registering razor there would mint a duplicate component per file with the edges anchored on the pass's node instead of the one `@inject` / `@using` / the `CONTAINS` edges already use. `TestRazorHierarchy_NoComponentMintedForHierarchyTarget` and three `forbidden_entities` rows guard the **shape** that route produces — an entity minted from a hierarchy target rather than from a file path. **Neither observes the route itself**: both drive this extractor directly, so adding `"razor"` to `supportedLanguages` would not move either by a byte. The test was called `…NoDuplicateComponents` and advertised as guarding the route; an assertion whose name claims a wider population than it covers is a defect this board keeps re-finding, so it is renamed and both prose sites are corrected.

**The razor-specific judgement calls:**

- **FROM side.** Razor entities are not classes — the only per-file type-like node is the `SCOPE.UIComponent` whose name `componentNameFromPath` computes from the *file path without reading the file*. Confirmed that is what the rest of the razor graph anchors on (the `CONTAINS` edges to event handlers are embedded on the same record), so the hierarchy edges are embedded there too, with **`FromID` empty** so assembly stamps the owning record's id (#6295 / #6298 / #6365 / #6367).
- **TO side.** Bare written name, generic arguments erased — the Solidity / Crystal / F# / Groovy convention. A Blazor base or interface is essentially always declared elsewhere, so a file-pinned structural ref would never bind.
- **Multiplicity.** Razor spells several interfaces as repeated `@implements` **lines**, not a comma list. Every match is read and deduped.

### The qualified-target decision, measured rather than argued

`@inherits MyApp.Components.BaseComponent` is emitted **as written**, and that is the form that binds *worse*. Measured through the production resolver (`TestRazorHierarchy_QualifiedTargetDoesNotBind` keeps the table from going stale):

| ToID form | binds to `Components/BaseComponent.razor` |
|---|---|
| `MyApp.Components.BaseComponent` (as written) | **0 of 1** |
| `BaseComponent` (bare last segment) | **1 of 1** |

The ocaml arm's lesson (#6948) is that "no mis-binding" is not "sufficient" — ambiguity drops the edge rather than pointing it wrong. I chose the losing form anyway.

**The original reason I gave was half-true and the review probed it past n=1.** "A leaf name binds to any same-named type in any namespace" over-claims:

| candidates named `BaseComponent` | bare form |
|---|---|
| exactly **one**, and not the intended one | **BINDS — to the wrong node.** This is #6369's hazard, and it is the shape that justifies the decision |
| **two or more** | **DANGLES.** Ambiguity drops the edge; it does not mis-point it — the #6948 outcome |

So the decision stands on the single-candidate shape alone, and razor components carry **no namespace at all** to tell the two apart. The distinction is worth stating precisely because the ambiguous shape is the one a reader assumes — getting it backwards is the correction #6948 had to make publicly. Narrowed in both `hierarchy.go` and the test comment. All 7 corpus occurrences are unqualified, so either direction would be widened on no evidence. **The qualified form therefore dangles — a recorded gap of the same class as ocaml's deliberately-unemitted module-qualified parents (#6947), not an accident.**

### BOM (#6962) — decided, and said out loud

#6962 records that a UTF-8 BOM defeats `(?m)^@using` at line 1 in this package. My anchors would have inherited the hole verbatim, so **they tolerate a leading BOM explicitly** and `TestRazorHierarchy_BOMAtLineOne` pins it *with the directive on line 1*, not line 2 — line 2 is preceded by a `\n` that `^` matches with or without the handling, so a line-2 fixture would prove nothing.

**This is not a fix for #6962.** `reUsing` and `reInject` are untouched and still carry the defect; #6962 owns the package-wide fix. What is claimed is only that this arm does not ship a third anchor with the same hole.

## Evidence — endpoints, not counts

`TestRazorHierarchy_ResolvedEndpoints` drives the real pipeline (`ResolveImports` → `ReferencesEmbedded`) exactly as graph assembly does. Before this arm the test could not be written: the component carried zero hierarchy relationships, so there was no endpoint to measure.

```
FROM=MainLayout@Components/Layout/MainLayout.razor EXTENDS    TO=LayoutComponentBase@Components/Layout/LayoutComponentBase.razor
FROM=MainLayout@Components/Layout/MainLayout.razor IMPLEMENTS TO=<UNRESOLVED:IDisposable>
```

The unresolved row is asserted deliberately: `IDisposable` is a BCL interface in no indexed tree, and pinning the dangle is what stops the fixture from silently starting to bind to something else.

## New golden fixture: `razor-blazor-mini`

A Blazor tree (4 components + one C# file) that grades both directions (#6902):

- `MainLayout` **EXTENDS** `AppLayoutBase` — razor → razor, cross-file.
- `NavMenu` **IMPLEMENTS** `IAppDisposable` and `IAppTrackable` — **cross-language razor → C#**, which is where a Blazor component's interfaces actually live. The second row is the only one that grades multiplicity.
- Two pre-existing `CONTAINS` rows held, so "dropped all embedded relationships" stays distinguishable from "dropped only the hierarchy kinds".
- Five forbidden relationship rows and three forbidden entity rows, each annotated with whether it FIRES under a named mutant or is an unreachable fence, stated plainly.

**Grade: entities 6/6 (100%), relationships 5/5 (100%), forbidden hits 0, forbidden entity hits 0.**

### Re-baseline, deliberate

`baseline.json` gains **one** entry and no existing figure was touched:

| | entity_found / expected | rel_found / expected | entity_extracted_total | rel_extracted_total |
|---|---|---|---|---|
| `razor-blazor-mini` **before** | — (fixture did not exist) | — | — | — |
| `razor-blazor-mini` **after** | 6 / 6 | 5 / 5 | 25 | 35 |

**0 removed, 0 re-kinded, 0 renamed** — the entry is additive and no other fixture's figures move, because no other fixture contains a `.razor` file (the only producer this change touches). The three golden-corpus size constants move **37 → 38** in `absent_relationships_6490_test.go`, `subtype_assertion_6488_test.go` and `zero_relationships_6490_test.go`; the delta is **re-derived from disk** (`ls -d internal/quality/golden/*/ | wc -l` = 38) and the one fixture member producing it is `razor-blazor-mini`. Two corpus ledgers also needed a real entry rather than a bump:

- `commentPrefixes6949` gains `".razor": {"@*", "//"}` — **two** prefixes, because a `.razor` file is two languages stacked (markup comments `@* … *@`, the `@code` half is C# and comments `//`). Without it all four razor sources were **silently unscanned**, which is the exact false-negative that test exists to prevent; it caught this itself.
- The `TestOnlyTheIntendedGoldenRowsAssertSubtype_6488` ledger gains this fixture's six subtype-asserting rows. The subtypes are load-bearing, not decoration: `SCOPE.UIComponent` means **two** things in this fixture (a component, and an `@inject` service — `Js` in the *same file* as the `Counter` row), and `SCOPE.Component` means **four** in `Shared/Abstractions.cs` (file carrier, class, interface, `interface:*` extraction record). A kind-only row is satisfiable by a record that is not the thing the row names.

## Mutation scoring

Twelve mutants, preferring the **more permissive** direction. Every edit was applied by a script that asserts an **exact occurrence count** before writing and aborts otherwise (`M6` legitimately rewrites 2), the baseline was verified green immediately before each score, and each was restored by `cp` from a shasum-verified backup — never `git checkout` — with the checksums re-verified after every restore. `git diff HEAD~1` was checked before pushing.

|---|---|---|---|
| M1 | drop the `^` anchor | **DEAD** | `TestRazorHierarchy_DirectiveMustBeLineLeading` **and** fixture forbidden `Counter --[EXTENDS]--> WraithBase` |
| M2 | disable `scrubRazorComments` | **DEAD** | `TestRazorHierarchy_RazorCommentsIgnored` **and** fixture forbidden `Counter --[IMPLEMENTS]--> IPhantom` |
| M12 | M1 **and** M2 together | **DEAD** | 3 forbidden hits — `WraithBase`, `GhostBase`, `IPhantom`. Run because the `GhostBase` decoy sits mid-line inside a one-line comment, so neither M1 nor M2 alone reaches it; the row's own note says so rather than implying a kill it never had |
| M3 | accept `@inherit` (`@inherits?`) | **DEAD** | `TestRazorHierarchy_NearMissDirectiveNamesRejected` |
| M4 | also emit **EXTENDS** for `@implements` | **DEAD** | `TestRazorHierarchy_ImplementsMultipleDirectives` **and** forbidden `NavMenu --[EXTENDS]--> IAppDisposable`. Recall stays 5/5 under it — the forbidden row is the only fixture-side instrument |
| M5 | also emit **IMPLEMENTS** for `@inherits` | **DEAD** | `TestRazorHierarchy_InheritsAndImplements` **and** forbidden `MainLayout --[IMPLEMENTS]--> AppLayoutBase`. Same: recall unmoved |
| M6 | drop the BOM tolerance | **DEAD** | `TestRazorHierarchy_BOMAtLineOne` |
| M7 | read only the first match per directive | **DEAD** | `TestRazorHierarchy_ImplementsMultipleDirectives` **and** fixture relationships 4/5 |
| M8 | drop the self-edge guard | **DEAD** | `TestRazorHierarchy_SelfReferenceDropped` |
| M9 | drop generic erasure | **DEAD** | `TestRazorHierarchy_GenericsErased` |
| M10 | set `FromID` to the owner (file-anchored shape) | **DEAD** | 9 unit tests via `hierTargets`' `FromID` assertion |
| M11 | `directiveTarget` returns the whole trimmed line | **was ALIVE → now DEAD** | see below |

**M11 is the one worth reading.** Returning `strings.TrimSpace(stripGenerics(arg))` instead of the first token survived the *entire* suite and the golden fixture at 100%/0-forbidden. It is **distinguishable but was ungraded**: the two readings agree on every well-formed directive (generics are erased before the split, and a same-line `@* … *@` is blanked to spaces TrimSpace removes) and differ only on a malformed line — `@inherits Foo Bar` emits `Foo`, which binds, versus `Foo Bar`, which can never bind. Rather than declare it equivalent or not-production-reachable, it is **graded**: `TestRazorHierarchy_FirstTokenOfArgumentWins` pins the binding reading, and `directiveTarget`'s doc comment now names that test and says why it exists. Re-scored after: **DEAD**.

## Review round (second commit)

An independent adversarial review **APPROVED** and reproduced every table above — corpus counts, the 37→38 re-derivation, the fixture grade, and all suites including `./cmd/grafel/`. It left one ungraded mutant and four wording defects. **Three of the four are the same underlying defect in three directions: an annotation asserting something no test observes** — one over-claiming a mechanism (the qualified-target rationale), one under-claiming a guard (the `forbidden_entities` notes), one naming a population it does not cover (the duplicate-component test).

| # | Note | Fix |
|---|---|---|
| 1 | **leading whitespace ungraded** | decided (rejected) and asserted — see Known gaps |
| 2 | **qualified-target reason half-true** | narrowed to the single-candidate shape — see above |
| 3 | **`forbidden_entities` notes falsified in my favour** | corrected; the rows are graded, not aspirational |
| 4 | **test name claims a route it cannot see** | renamed + both prose sites corrected |
| 5 | **`rebuildInjectEntities` phantom-ALIVE trap** | named beside the code |

**On (3), I reproduced the falsifying mutant rather than relaying it.** M14 — mint a `SCOPE.Component` per hierarchy target, the shape the rejected route produces — fires `forb_ent 1` on `AppLayoutBase [SCOPE.Component] in Components/Layout/MainLayout.razor` and fails the renamed test. Combined with M12+M2 (so the comment decoys match at all), **all three** entity fences fire. Running it turned up one more thing reading could not: `GhostBase`'s row named `kind: SCOPE.UIComponent`, a kind the route never mints, so that row alone sat **unfireable beside two that fired** — a fixture label claiming more than its content. Corrected to `SCOPE.Component`, re-verified, and the reason recorded in the row itself.

**On (5)**: `rebuildInjectEntities` retains only `entities[:1]`, so an entity appended between step 1 and step 3 is silently discarded — an entity-side mutant placed there is a no-op that scores as a phantom ALIVE and argues a guard is missing. The hierarchy edges are unaffected (they ride on `entities[0]`). Named in a comment beside the code, which is the standing fix for this class; M14 is placed after step 3a for exactly this reason.

| # | New mutant | Verdict | Killed by |
|---|---|---|---|
| M13 | accept leading horizontal whitespace (2 occurrences) | **DEAD** (was ALIVE before this round) | `TestRazorHierarchy_DirectiveMustBeLineLeading` — `indented directives produced [EXTENDS:IndentedBase IMPLEMENTS:IIndented], want none`. Applied → RED, restored by `cp` from a shasum-verified backup → GREEN |
| M14 | mint a `SCOPE.Component` per hierarchy target | **DEAD** | `TestRazorHierarchy_NoComponentMintedForHierarchyTarget` (+2 more) **and** fixture `forb_ent 1` |
| M15 | M14 + M12 + M2 | **DEAD** | `forb_ent 3` — all three entity fences named |

The clean fixture grade is **unchanged** at 6/6, 5/5, 0 forbidden, 0 forbidden entities, so `baseline.json` needs no edit. Suites re-run after the round: `quality`, `quality/{analytics,audit,fitness}`, `extractors/razor`, `extractors`, **`cmd/grafel`** — `EXIT=0`; `gofmt -l` clean, `go vet` clean.

### Third commit — MR-1, and the pattern named

The coordinator's own mutant on the second head found a **fourth** unheld claim, same class again. `scrubRazorComments` blanks a comment's bytes but deliberately leaves its **newlines**, because `lineOf` counts newlines in the *scrubbed* string. Removing the `if out[i] != '\n'` guard passed the entire package:

```
src = "@*\ncomment\n*@\n@inherits BasePage\n"   // @inherits is on line 4
mutant:   EXTENDS -> BasePage line=2
restored: EXTENDS -> BasePage line=4
```

Nothing observed it because the scrubber's two halves were covered by tests that **could not see each other**: `TestRazorHierarchy_LineStamped` asserted a line over a source with no comment in it, and the comment tests exercise a multi-line `@* … *@` but assert only the *absence* of edges, never a line. The function's doc comment stated the claim outright — that sentence was the thing nothing held.

`TestRazorHierarchy_LineStamped` now stamps **through** a two-line comment header (an ordinary Blazor file shape, not a contrived input) and asserts exactly one EXTENDS edge, so an unstamped edge cannot vacuously satisfy the line check.

| # | New mutant | Verdict | Killed by |
|---|---|---|---|
| MR-1 | `scrubRazorComments` blanks newlines too | **DEAD** (was ALIVE) | `TestRazorHierarchy_LineStamped` — `line = "4", want "6"`. Applied → RED, restored by `cp` from a shasum-verified backup → GREEN |

`hierarchy.go` is unchanged by this commit; only the test moved, so `./cmd/grafel/` did not need re-running (it was green on the identical non-test tree one commit earlier).

### The pattern, named for whoever reads this package next

**Four instances in one arm** of prose running ahead of the assertions — the dominant defect class on this board, and worth stating as a pattern rather than as four separate fixes:

1. the qualified-target rationale **over-claimed a mechanism** (mis-binds vs. dangles);
2. three `forbidden_entities` notes **under-claimed a guard** ("unreachable" rows that fire);
3. a test **named a population it does not cover** (`…NoDuplicateComponents` cannot see the route it advertised);
4. the scrubber's doc comment **asserted an invariant nothing measured** (newlines preserved for line stamping).

They were found by three different people in three passes, and each was found by *running something*, never by re-reading. The transferable instruction: for every clause in a doc comment, ask whether there is an input where **that clause alone** decides the outcome — and if the answer is yes, the input belongs in a test. My finding the unfireable `GhostBase` row is what prompted the search for a fourth; there may be a fifth.

One note I am **not** acting on, recorded rather than dropped: the review found that deleting just `"//"` from the `commentPrefixes6949` razor entry leaves the suite green — the shipped fixture has no framework marker in a `//` comment, so that half of the entry is correct but unpinned. False-negative direction only, and pinning it would mean planting a framework marker in the fixture for no other reason.

## Verification

macOS (Darwin 25.6.0) / APFS / arm64, in a dedicated worktree; the repo root was never walked, and every corpus figure above names its tree by absolute path. `go test -count=1` on every touched package **plus `./cmd/grafel/`** (it digest-pins the whole graph over a mixed-language fixture and nothing under `internal/extractors/...` points at it):

```
ok  internal/quality            43.072s
ok  internal/quality/analytics   2.027s
ok  internal/quality/audit       1.235s
ok  internal/quality/fitness     1.355s
ok  internal/extractors/razor    0.562s
ok  internal/extractors         45.142s
ok  cmd/grafel                 155.892s
ok  internal/resolve             1.687s
EXIT=0
```

`gofmt -l` clean, `go vet` clean. The quality run used an isolated `HOME` + `GRAFEL_HOME` + `GRAFEL_DAEMON_ROOT`; the real `~/.grafel` was never written and no daemon was started, stopped or killed.

## Known gaps, filed rather than hidden

- **Namespace-qualified `@inherits`/`@implements` dangle** (0 of 105 corpus files use one) — measured above, same class as #6947.
- **Leading whitespace before a directive is not accepted — decided and pinned.** Originally shipped "deliberately undecided", which the review correctly rejected: the permissive form (`^[^\S\n]*@inherits…`) passed the *entire* suite and the fixture at 6/6, 5/5, 0 forbidden, so the direction was distinguishable, production-reachable and observed by nothing. A zero over 105 files is corpus-relative and cannot defend a choice. Now REJECTED and asserted (M13 below), matching `reUsing`/`reInject`; accepting indented directives may well be right, but it is a widening that needs its own evidence, a coordinated change to those two anchors, and an edit to this assertion.
- **The BOM defect in `reUsing`/`reInject` is untouched** — #6962 owns it.

Refs #6343, #6321, #6369, #6947, #6948, #6902, #6962, #6437


cajasmota:
## Independent adversarial review — **APPROVE** (four non-blocking notes)

Independent checkout, detached at `4a30e8868`, **absolute path**
`/private/tmp/claude-501/-Users-jorgecajas-Projects-archigraph/c0fefed4-0195-4c81-9a94-07a7145710a4/scratchpad/review-6963-vx4/wt`.
Corpus figures measured read-only in `/Users/jorgecajas/Projects/archigraph-corpora`.
macOS Darwin 25.6.0 / APFS / arm64. `GRAFEL_HOME` + `GRAFEL_DAEMON_ROOT` redirected into my own scratch;
the real `~/.grafel` was never written, no daemon touched, no process killed. `go vet` clean, `gofmt -l` empty,
`-count=1` throughout, restores by `cp` from a shasum-verified backup, final `git status --porcelain` empty.

The body's tables are falsifiable and **every one of them reproduced**.

### Reproduced, independently

| Claim | My measurement | Verdict |
|---|---|---|
| 105 routed `.razor` files | 105 | ✅ |
| 4 `^@inherits` (all `LayoutComponentBase`) | 4 files / 4 occ, all `LayoutComponentBase` | ✅ |
| 3 `^@implements` (2 `IDisposable`, 1 `IAsyncDisposable`) | exactly that | ✅ |
| 0 occurrences with leading whitespace | 0 | ✅ |
| 0 of 105 files carry a BOM | 0 | ✅ |
| all 7 occurrences unqualified | 0 dotted targets | ✅ |
| golden corpus 37 → 38, sole new member `razor-blazor-mini` | `git ls-tree -d HEAD~1` = 37, `HEAD` = 38, diff = one line, `razor-blazor-mini`; no other fixture contains a `.razor` file | ✅ re-derived, not bumped |
| fixture 6/6, 5/5, 0 forbidden, totals 25/35 | my own `grafel quality` run: `ent 6/6 rel 5/5 forbidden 0 forb_ent 0`, `entity_extracted_total=25`, `relationship_extracted_total=35` | ✅ exactly `baseline.json` |
| suites green incl. `./cmd/grafel/` | `razor`, `quality`, `quality/{analytics,audit,fitness}`, `resolve`, `cmd/grafel` — **EXIT=0** (cmd/grafel 144s) | ✅ digest pin holds |

**No orphan figure is used as evidence anywhere in this PR** — I looked. Correct per #6947.

### My own mutants (7, restored + re-verified green after each)

| # | Mutant | Occurrences edited | Verdict |
|---|---|---|---|
| R1 | drop the `^` anchor (both patterns) | 2 | **DEAD** — `TestRazorHierarchy_DirectiveMustBeLineLeading`; fixture exit 2 with exactly `Counter --[EXTENDS]--> WraithBase`, recall unmoved at 5/5 |
| R2 | **allow leading horizontal whitespace** (`^[^\S\n]*\x{FEFF}?@…`) | 2 | **ALIVE — distinguishable and ungraded** (see below) |
| R3 | remove the dedupe (`if seen[key]` → `if false`) | 1 | **DEAD** — `TestRazorHierarchy_DuplicateTargetDeduped` |
| R4 | drop the BOM tolerance (`\x{FEFF}?` removed) | 2 | **DEAD** — `TestRazorHierarchy_BOMAtLineOne`: `EXTENDS = [], want [LayoutComponentBase]` |
| R5 | extra `add("EXTENDS", reImplements)` | 1 | **DEAD** — 6 unit tests **and** fixture forbidden `NavMenu --[EXTENDS]--> IAppDisposable` **while recall stayed 5/5**. The forbidden row is genuinely the only fixture-side instrument, and it is not decoration |
| R6 | `directiveTarget` → `strings.TrimSpace(stripGenerics(arg))` | 1 | **DEAD**, killed by `TestRazorHierarchy_FirstTokenOfArgumentWins` **and nothing else** — your M11 narrative is exactly right |
| R7 | mint a `SCOPE.Component` per hierarchy target (the cross/hierarchy-shaped duplicate) | 1 | **DEAD** — `TestRazorHierarchy_NoDuplicateComponents` + fixture `forb_ent 1`. See note 3 |

Every edit was proved to land by an exact occurrence count before writing; R7 was scored twice, in two positions,
which is what turned up note 3.

### (a) The qualified-target decision — **sound, but the stated reason is half-true**

I widened it past n=1. Three probes through the production resolver
(`ResolveImports` → `ReferencesEmbedded`), same shape as your test:

| Probe | Setup | Result |
|---|---|---|
| C | qualified `MyApp.Components.BaseComponent`, one `Components/BaseComponent.razor` | **DANGLES** — reproduces your 0-of-1 |
| A | bare `BaseComponent`, **two** `BaseComponent.razor` in different folders | **DANGLES** — the ambiguous case **drops**, it does not point at a wrong node |
| D | bare `BaseComponent`, **exactly one** unrelated `ThirdPartyArea/BaseComponent.razor` | **BINDS** to `ThirdPartyArea/BaseComponent.razor` — the wrong node |

So: the #6369 wrong-node hazard is real, but only in the D shape (one same-named candidate that is not the intended
one). In the A shape the resolver degrades to exactly the ocaml #6948 outcome — ambiguity drops the edge. The code
comment's "a leaf-name `ToID` binds to any same-named type in any namespace" overstates it; D is the case that
justifies the decision, and it is worth naming precisely, because A is the shape a reader assumes. Ship as is — with
0 of 105 corpus files qualified, both directions are unevidenced, which the body already says plainly.

### (b) The BOM fixture — **it really carries the bytes**

Not asserted from intent: `﻿` in a Go literal is `EF BB BF`, and R4 proves it behaviourally — removing the
tolerance makes `TestRazorHierarchy_BOMAtLineOne` fail with an empty result. A test that only *believed* it had a BOM
would have stayed green under R4. Directive on line 1, so the `\n`-before-line-2 escape hatch does not apply. ✅

### (c) `commentPrefixes6949` — **non-vacuous, in three independent ways; one half unpinned**

- Delete the `".razor"` entry → `TestGoldenFixtureMarkersAreNotCommentOnly_6949` **fails**:
  *"4 fixture source files have an extension commentPrefixes6949 does not know, so they were NOT scanned"*, naming
  the razor files. The coverage limb is real, and your "it caught this itself" is verified.
- Detection, not just coverage: I planted a real `import_marker` in an `@* … *@` markup comment → flagged; and one in
  a `//` comment inside `@code` → flagged. **Both prefixes function.**
- **Note 4 (minor):** dropping just `"//"` from the entry leaves the whole suite green — the shipped fixture has no
  framework marker in a `//` comment, so that half of the entry is correct but unpinned. False-negative direction
  only; not worth a fixture edit, worth knowing.

### (d) Are any guards decoration? No — but one annotation is falsified

1. **`forbidden_relationships` kind-exclusivity rows** — verified live under R5: recall stays 100%, the row is the
   sole detector, exit 2. Non-vacuous.
2. **`TestRazorHierarchy_NoDuplicateComponents`** — fires under R7 (mutant minting an entity per hierarchy target).
   Not decoration **within this package**. But it cannot observe the route it is advertised as guarding: `extract()`
   calls `extractor.Get("razor")` only, so registering razor in `cross/hierarchy`'s `supportedLanguages` would not
   move it at all. The body and `hierarchy.go`'s package comment both say the test "guards" that route; it guards
   the *symptom* in this extractor, not the route. One-line wording fix.
3. **The `AppLayoutBase / SCOPE.Component / MainLayout.razor` `forbidden_entities` row is NOT unreachable.** Its own
   note says *"NOT FIRED by any mutant of this extractor — it cannot be, since this extractor mints entities only
   from file paths."* R7 falsifies that: a mutant that mints entities from directive **arguments** produced
   `forb_ent 1` — `AppLayoutBase [SCOPE.Component] in Components/Layout/MainLayout.razor subtype=component`. The row
   is graded, which is strictly better than advertised; the annotation is the thing that is wrong. Same for the
   `GhostBase` / `IPhantom` rows' "unreachable" wording — reachable by the same mutant class.
4. Worth recording while I was in there: `rebuildInjectEntities` **retains only `entities[:1]`**, so an entity
   appended between step 1 and step 3 is silently discarded — R7 placed there was a *no-op that read as a passing
   mutant*. The hierarchy edges are unaffected (they ride on `entities[0]`), but anyone scoring a future entity-side
   mutant in this package must place it after step 3a or they will score a phantom.

### Note 2 (the one live mutant): leading whitespace is ungraded

`(?m)^[^\S\n]*\x{FEFF}?@inherits…` passes the **entire** razor suite, the whole `internal/quality` package, and the
golden fixture at 6/6, 5/5, 0 forbidden. It is distinguishable and production-reachable (an indented `    @inherits
Foo`), and nothing observes the choice in either direction. The body calls it "deliberately undecided" — fair, but a
zero is corpus-relative (#6437's lesson) and "matches `reUsing`/`reInject`" is a claim no test holds. Cost to grade:
**two lines** — one indented-directive case appended to `TestRazorHierarchy_DirectiveMustBeLineLeading`, pinning
whichever direction you mean. Not a merge blocker; a decision that is documented and unpinned is one edit from being
silently reversed.

### Other checks

- **`FromID` empty ⇒ owner's id, and the hierarchy edges anchor where the rest of the razor graph does.**
  Verified structurally, not by prose: `extractor.go:164` (hierarchy) and `extractor.go:225` (CONTAINS) both append to
  `entities[0]`; `hierTargets` errors on any non-empty `FromID`; `TestRazorHierarchy_ResolvedEndpoints` resolves FROM
  to `MainLayout@Components/Layout/MainLayout.razor`, TO to `LayoutComponentBase@…LayoutComponentBase.razor`, with the
  `IDisposable` row deliberately `<UNRESOLVED:…>`. All confirmed. ✅
- **Can the FROM side collide?** (`componentNameFromPath` never reads the file.) Probe with two `MainLayout.razor` in
  different folders: distinct ids `05c68706…` / `b962a8b2…`, one edge each, `FromID=""`. `graph.EntityID` keys on the
  source file, so **no FROM-side collision**. ✅
- **Fixture read against its names:** `AppLayoutBase.razor` genuinely declares no directive; `MainLayout.razor` has
  exactly the one `@inherits`; `NavMenu.razor` has two `@implements` on separate lines plus a real `@code` handler;
  `Counter.razor` carries all three decoys as described (mid-line markup, one-line `@*…*@`, multi-line block) and no
  directive; `Abstractions.cs` really declares both interfaces plus the `JsHelper` class and `IJsHelper` that make the
  subtype rows load-bearing. Each file carries what its name implies, and the multiplicity axis varies while the
  owning component is held constant. ✅

**APPROVE.** The four notes are wording/pinning, none of them changes an edge. Recommended follow-ups, all small:
correct the "unreachable fence" annotations on the three `forbidden_entities` rows (they are reachable and fire),
narrow the "guards the route" wording on `TestRazorHierarchy_NoDuplicateComponents`, tighten the qualified-target
rationale to the one-candidate case, and either pin the leading-whitespace direction or say in the comment that it is
unpinned.

cajasmota:
## Coordinator mutant on the final head (`a9c987af4`)

Scored in my own detached worktree, `git status --porcelain` empty, restored by `cp` from a shasum-verified backup (`ee42c734…` both ways), baseline verified green immediately before.

I picked a target outside the reviewer's seven and outside your own set: not the anchors, the dedupe, BOM tolerance, `directiveTarget`, entity-minting, or whitespace — the **comment scrubber's newline preservation**.

**MR-1 — blank newlines inside `@* … *@` along with everything else.**

`scrubRazorComments` currently keeps newlines where they are:

```go
for i := loc[0]; i < loc[1]; i++ {
    if out[i] != '\n' {
        out[i] = ' '
    }
}
```

I removed the guard so every byte in a comment becomes a space. Edit proven landed (`pre=1`, `post_old=0`, `post_new=1`), `go vet` clean.

**Result: ALIVE** — full `./internal/extractors/razor` green, exit 0.

## Verdict: distinguishable, production-reachable, and ungraded

The function's own comment states the claim: *"keeping every byte offset (and every newline) where it was so line stamping stays exact."* Line numbers are computed as `lineOf(scrubbed, m[0])` — off the **scrubbed** string — so collapsing a comment's newlines shifts every directive that follows it.

I demonstrated it rather than arguing it, with a throwaway probe (since removed; the worktree is clean and back at `a9c987af4`):

```
src = "@*\ncomment\n*@\n@inherits BasePage\n"     // @inherits is on line 4

mutant:   PROBE: EXTENDS -> BasePage line=2   (true line is 4)
restored: PROBE: EXTENDS -> BasePage line=4   (true line is 4)
```

A two-line comment header above a directive is ordinary Blazor, so this is reachable on real files, and the edge would point two lines off.

**Why nothing caught it.** `TestRazorHierarchy_LineStamped` is the only line assertion in the package, and its source is `"@page \"/x\"\n@using System\n@inherits BasePage\n"` — **no razor comment at all**. The comment tests at `hierarchy_6370_test.go:254-255` do exercise multi-line `@* … *@`, but they assert the *absence* of edges, never a line. So the two halves of the scrubber are each covered by a test that cannot see the other: one file has comments and checks no edges, the other checks a line and has no comments. Neither observes the interaction the doc comment is about.

That is the same shape as three things you already corrected this round — an annotation asserting something no test observes — which is why I think it belongs with them rather than as a follow-up issue.

**The fix is one character of test input:** put a multi-line `@* … *@` above the directive in `TestRazorHierarchy_LineStamped` and assert the line is still correct. Prove it goes RED under MR-1 first, per the usual rule.

## Not a defect in your diff

The shipped code is correct. This is the fourth instance in this PR of a true claim that nothing holds — and the fact that you found the `GhostBase` unfireable-row yourself while checking the third one is the reason I looked for a fourth. Worth a line in the body noting the pattern: this arm's prose was consistently ahead of its assertions, in both directions.

Merging once CI is green on this head, with MR-1 graded.

Issue #6957, opened 2026-09-07
Play Java and Revel routes producers are still unreachable — each hard-gates on its own base language, and one file has one token


cajasmota:
#6956 routed `conf/routes` / `*.routes` to `"scala"`, which reaches
`internal/custom/scala`'s `rePlayRoute`. It does NOT reach the other two producers, because
each hard-gates on its own base language:

```
internal/custom/java/play_routes.go:145   ctx.Language != "java" || !playFrameworks[ctx.Framework]
internal/custom/golang/revel.go           language     != "go"
```

One file has one language token, so no classification can satisfy all three. A Play **Java** repo
and a **Revel** repo now get their routes extracted by the *scala* producer (which content-sniffs
and stamps `framework: "play"` regardless, so the routes do land) — but
`internal/custom/java`'s richer `Route` + `Handler` pair, and Revel's `module:` mounts and
controller-action attribution, stay unreachable in production exactly as before.

Measured on `play-samples/play-java-rest-api-example` (via #6956's harness): +8 entities, all from
the scala arm. `internal/custom/java`'s `extractPlayRoutes` contributed zero.

Two shapes to price:

1. **Widen the three gates** to accept the routes language in addition to their own, and dispatch
   the file to all three prefixes via `extraCustomPrefixesForLanguage` (which exists for exactly
   this — "files whose content is owned by more than one language family"). Then all three fire and
   the dedup problem moves downstream.
2. **Pick the owner per repo** from the co-present language (a `build.sbt` + `app/**.scala` repo is
   Play Scala; a `pom.xml` + `app/**.java` repo is Play Java). The classifier has no repo context,
   so this belongs in the framework-detection pass, not in `detectLanguage`.

Note that (1) makes THREE producers of the same route live at once, which is the #6370 / #6927
shape one step worse than what #6952 already flagged. Whichever is taken, it needs the
`Route` / `SCOPE.Operation` ownership question in the sibling issue settled first.

Refs #6952, #6956

cajasmota:
Grounding note from a coordinator mutant on #6956 (`d170d6863`). The diagnosis here is correct; this adds **which layer actually decides it**, because the answer is not where the issue points and it changes what a fix has to touch.

## Measured

[...]

Issue #6276, opened 2026-08-08
A Django Manager subclass is emitted as SCOPE.Schema, which is not what a manager is


cajasmota:
Found while enabling the `internal/custom/**` extractors for the golden benchmark (#6260). Not caused by that change — the change only made it observable, because before it the Django custom pass never ran during the benchmark.

## Symptom

`internal/quality/golden/python-django-mini` loses 1 must-have entity (26/28 → 25/28) and 2 must-have relationships (7/12 → 5/12).

The entire delta is one class. Before and after, the per-fixture report's missing lists:

```
BEFORE  ent 26/28  rel 7/12          AFTER  ent 25/28  rel 5/12
  SCOPE.Component UserListView         SCOPE.Component ActiveUserManager   <- NEW
  SCOPE.Component UserDetailView       SCOPE.Component UserListView
                                       SCOPE.Component UserDetailView
  REL UserListView   CONTAINS .get     REL UserListView       CONTAINS .get
  REL User           CONTAINS .full_label   REL User          CONTAINS .full_label
  REL User           CONTAINS .Meta    REL User               CONTAINS .Meta
  REL UserDetailView CONTAINS .get     REL ActiveUserManager  CONTAINS .by_email      <- NEW
  REL main           CALLS  execute_…  REL ActiveUserManager  CONTAINS .get_queryset  <- NEW
                                       REL UserDetailView     CONTAINS .get
                                       REL main               CALLS  execute_…
```

**`ActiveUserManager` and its two methods are the whole regression.** Everything else on this fixture was already missing.

### Note for anyone triaging from the #6260 summary

`UserListView` / `UserDetailView` are **not** part of this drop. They were already absent as `SCOPE.Component` with the custom pass off, and they are absent with it on. An earlier summary of #6260 described this fixture's regression as a views-reclassification question; that is wrong. Whether Django class-based views should be `SCOPE.Component`, `SCOPE.Operation`/`endpoint`, or `View` is a real question, but it is a **pre-existing residual**, unrelated to this change, and it should not be bundled here.

(For the record, `expected.json` already expects both facets for these classes — `{"name":"UserListView","kind":"SCOPE.Component"}` *and* `{"name":"UserListView","kind":"View","note":"Framework-level View entity."}` — so the fixture's model is that the language-level node and the framework-level node coexist. Only the `View` half currently resolves.)

## Mechanism

`internal/quality/golden/python-django-mini/src/users/models.py`:

```python
[...]

Issue #6275, opened 2026-08-08
A Hibernate @Entity facet displaces the class node it is a facet OF, and an ORM sentinel silently stands in for it


cajasmota:
Found while enabling the `internal/custom/**` extractors for the golden benchmark (#6260). Not caused by that change — the change only made it observable, because before it the custom pass never ran during the benchmark.

## Symptom

`internal/quality/golden/java-spring-mini` loses 6 must-have relationships, 21/21 → 15/21:

```
SCOPE.Component User -CONTAINS-> SCOPE.Operation User.getEmail
SCOPE.Component User -CONTAINS-> SCOPE.Operation User.getName
SCOPE.Component User -CONTAINS-> SCOPE.Schema    User.email
SCOPE.Component User -CONTAINS-> SCOPE.Schema    User.name
SCOPE.Component User -CONTAINS-> SCOPE.Schema    User.id
SCOPE.Component com/example/demo/repository/UserRepository.java -IMPORTS-> SCOPE.Component User
```

All six report `from_resolved=true, to_resolved=true` — both endpoints exist, the edge between them does not.

Entity recall stays at **25/25**, and that is the more alarming half: the fixture's `SCOPE.Component User` expectation is satisfied, but by a stub. See "The false green" below.

## Mechanism

`internal/quality/golden/java-spring-mini/src/com/example/demo/model/User.java` is a plain `@Entity` class with five members.

**With the custom pass off**, the graph holds exactly one entity named `User`:

```json
{"id":"516f4c7014077766","kind":"SCOPE.Component","name":"User",
 "subtype":"orm_model_sentinel",
 "qualified_name":"scope:ormmodel:com/example/demo/model/User.java#User",
 "source_file":"com/example/demo/model/User.java","start_line":7}
```

Note `start_line: 7` — the real class's line. This node is ormlink's sentinel, which has **absorbed the base Java class node** in dedup and taken over its identity. The six CONTAINS/IMPORTS edges hang off it, and the fixture passes.

**With the custom pass on**, there are three:

```json
{"id":"516f4c7014077766","kind":"SCOPE.Component","name":"User",
 "subtype":"orm_model_sentinel","start_line":0}                       <- sentinel, now empty

{"id":"f6cde4f36281e32a","kind":"SCOPE.Schema","name":"User",
 "qualified_name":"com.example.demo.model.User","signature":"@Entity class User",
 "start_line":7,
[...]

PR #6965, opened 2026-09-07
fix(#6960): the daemon's incremental re-extract never dispatched a custom extractor — the gate lived in cmd/grafel, which internal/extractors cannot import

### LINKED ISSUE #6960: CONFIRMED + FIXED: the incremental path reached neither RunCustomExtractors call site, so a changed file lost every custom entity unchanged files kept (gate-ON only)
Found while grounding #6952's classifier change (PR #6956). **Filed as a question with a verified fact under it, not as a diagnosis** — the fact is certain, the consequence is not, and the difference decides whether this is a large defect or a non-issue.

## The verified fact

`RunCustomExtractors` is the **only** dispatcher that selects custom extractors — `cmd/grafel/inproc_custom.go:18`, `cmd/grafel/quality.go:30` and `cmd/grafel/index.go:4093` all say so in their own comments, because a custom pass registers under a synthetic language token (`python_dramatiq`, `scala_play`, …) that nothing else matches.

It has exactly **two** non-test call sites:

- `cmd/grafel/index.go:4142`
- `internal/daemon/extract/subproc.go:364`

**`internal/extractors/incremental.go` contains zero occurrences of it.** Verified by count in the repo root checkout at `59de3205d`.

## The open question — this is what someone needs to answer

**Does the incremental re-extraction path reach either call site?**

That is the whole issue, and it is not answered by the grep above. `subproc.go` is a dispatcher too, and if incremental work funnels through it, custom extractors run on incremental updates and there is nothing here to fix. If incremental work instead calls into `incremental.go`'s own extraction without reaching either site, then **a file changed after the initial index loses every custom-extractor entity it had** — and that would be universal across all ~340 custom passes, not specific to any language.

Answer it by **tracing the call chain**, not by grepping a filename. A prior finding on this board nearly became a phantom issue because "run by nothing" was concluded from one scoped grep without following the hop chain.

## Why it is worth the trace

If the consequence is real, the shape is nasty in a familiar way: the **initial** index is correct and the **incremental** one silently degrades, so the graph is right when you first look at it and wrong after you edit. That is the same silent-degradation class as #6921, #6923, #6928 and #6940 — none of which announced themselves.

It also bears on **#6959**: the Play `Route` ownership question assumes both the custom lane and the rule pack produce on the same file. If the custom lane drops out on incremental updates, ownership changes depending on *how* the file was indexed, which is a worse problem than two producers agreeing to disagree.

## What a resolution has to establish

1. **The call chain, stated hop by hop**, from an incremental file change to either `RunCustomExtractors` call site or to the point where it provably is not reached.
2. **Driven, not read**: change one file that a custom extractor produces entities for, run an incremental update, and compare the entity set before and after. `internal/custom/scala`'s Play routes producer is a convenient probe now that `conf/routes` is classified (#6952) — ~900 entities on a real repo, so a drop would be unmissable.
3. If it is real: whether the fix belongs at the call site or in the dispatcher, and **whether the ratchet or the `cmd/grafel` digest pin can see it at all** — both were structurally blind to #6946 and the digest pin is blind to #6952, so neither should be assumed to cover this.
4. If it is **not** real, say which hop carries it and put that in a comment at `incremental.go`, so the next person greping for `RunCustomExtractors` does not re-file this.

Refs #6952, #6959, #6940.

cajasmota:
Closes #6960.

## The answer to the question the issue asked

**No — the incremental path reaches neither `RunCustomExtractors` call site.** Traced hop by hop, then driven.

```
fs watcher / scheduler tick
  → cmd/grafel/daemon.go:1293  daemonSchedulerIncremental
  → internal/extractors.TryIncremental                 (incremental.go:393)
  → tryIncremental Step 6 "extract-changed-files" loop (incremental.go:~1000)
  → cls.ClassifyWithSize → Get(cr.Language)            (EXACT-KEY registry lookup)
  → safeExtract(ctx, ext, input)
  → [Pass 2.5 engine.Detector / classfold / scoped resolve / merge / write]
```

`Get` is an exact-key lookup. Every custom/framework extractor registers under a **prefixed** key — `custom_scala_di`, `python_django` — so that lookup can never select one. `RunCustomExtractors` (`custom_dispatch.go:208`) is the only dispatcher that does the prefix fan-out, and its two non-test call sites are `cmd/grafel/index.go:4142` and `internal/daemon/extract/subproc.go:364`, neither of which is on this path.

**Checked for indirect reference, per the #6957 lesson.** `RunCustomExtractors` is not held as a function value, not in a dispatch table, and not behind a registry indirection: `CustomExtractorsFor` has no non-test caller outside `custom_dispatch.go` itself, `MergeWithCustom`'s only production caller is `index.go:4154`, and `Get(cr.Language)` returns the single exactly-keyed extractor, not a composite that fans out. `internal/extractors/custom_registry.go` blank-imports all 22 `internal/custom/**` packages, so the extractors *are* registered in this very process — nothing structural prevented the call; there simply was no call.

## Driven, not read — and the incrementality proof

Mechanism: `extractors.TryIncremental` invoked directly, which **is** the production incremental entry point (`daemonSchedulerIncremental` is its only non-test caller). The proof that the pass was incremental rather than a silent full reindex is `res.Done == true` **and** `res.ChangedFiles == 1`: `TryIncremental` returns `Done=false` with a `FallbackReason` on every path where it declines and the caller runs a full index instead. Both assertions are `t.Fatalf` in the harness, so a fallback fails the test rather than passing it for the wrong reason. Log line from the run: `incremental: done changed=1 entities=8 rels=3 … flows_recomputed=true`.

Probe: a Play controller. Seed a graph containing its `SCOPE.DI|singleton:HomeController` (the "initial index ran with the gate on" state), edit the file, run one incremental pass.

| | before | after |
|---|---|---|
| `SCOPE.Operation\|about` (the edit, base extractor) | present | present |
| `SCOPE.DI\|singleton:HomeController` (custom-only) | **gone** | present |

Step 5 evicted the file's entities and the loop re-added only the base half.

## Why the consequence is narrower than the issue assumed — and why that decided the fix

Both full-index paths dispatch custom extractors only behind an **opt-in**:

- `cmd/grafel/index.go:4138` — `GRAFEL_INPROC_CUSTOM_EXTRACTORS`, default OFF;
- `internal/daemon/extract/subproc.go:364` — inside the `GRAFEL_SUBPROC_EXTRACT` branch, itself default OFF and described at `inproc_custom.go:20` as "an env var with zero writers anywhere in the tree".

So on a **default** configuration the initial index emits no custom entities either, and an incremental pass loses nothing — the graph was never wrong. With the gate **on**, it is exactly the silent-degradation shape #6921/#6923/#6928/#6940 share: a *changed* file loses what every *unchanged* file keeps, so the graph is right when you first look at it and wrong after you edit.

That is why the fix **reads the same gate** instead of dispatching unconditionally. The property this path owes its caller is "same graph as a full reindex", and a pass that runs custom extractors when the full path did not is as divergent as one that skips them when it did — plus it would silently pay the +17.5% wall / +18.2% alloc the gate exists to withhold.

## The fix: at the call site, with the gate lifted one package down

- `internal/extractors/custom_gate.go` (new) — `InProcCustomExtractorsEnabled()` now owns the env read. `incremental.go` cannot import `cmd/grafel`, so without this the incremental path would need a **second literal reading** of `GRAFEL_INPROC_CUSTOM_EXTRACTORS`, free to drift from the first. One reader, two callers.
- `cmd/grafel/inproc_custom.go` — `inProcCustomExtractors()` delegates. Its existing table test (`inproc_custom_extractors_5989_test.go`) now exercises the shared function.
- `internal/extractors/incremental.go` — the dispatch, placed **after** `safeExtract` and **before** `input.TSTree.Close()`, mirroring `index.go:4138`. Seam placement is load-bearing. **Correction from review:** moving it past the `Close()` is a silent **no-op**, not a use-after-free — `input.TSTree = nil` follows immediately, so the moved dispatch fails the `!= nil` guard and quietly emits nothing. The comment in the file said "use-after-free" and now says what was measured. The guard stays for the *other* reason (content-only extractors emitting with no tree, which the dart test grades), and `MergeWithCustom` — not `append` — applies the #6104 non-destructive (SourceFile, Kind, Name) merge.

Default behaviour is unchanged: with the gate off the added branch is not taken.

**Observed side effect, gate-on only, and it is the intended behaviour rather than a regression** (the facet it turns on is also what the #6275 blocker below was about): with the merge in place `class_kind_folds` goes 1 → 0 on the scala fixture and both `SCOPE.Component|HomeController` and `Controller|HomeController` survive. That is #6276's `twinAnchors` rule (`internal/engine/classfold.go:527`) firing correctly — the custom merge emits a `grafel.twin_of` facet, and a record some other record anchors is never a fold source. `cmd/grafel`'s `foldClassHierarchyShadows` holds the same exclusion, so the two paths agree.

## Can the ratchet or the digest pin see this?

**No, neither — by construction, exactly as the issue warned.** `cmd/grafel`'s digest pin runs with `GRAFEL_INPROC_CUSTOM_EXTRACTORS=""`, i.e. custom passes **off**, so it is structurally blind to this entire class; the full `./cmd/grafel/` suite is green here and would have been green with the defect in place. The extraction-quality ratchet grades a full index, not an incremental one, and there is no incremental leg at all. This is why the coverage added is four package-level tests rather than a threshold bump.

## Verification (macOS 15.6 / APFS, arm64), measured in `…/scratchpad/impl-6960-k7q2/wt`

- `go test -count=1 ./internal/extractors/` — **exit 0**, 37.2s.
- `go test -count=1 ./cmd/grafel/` — **exit 0**, 147.7s. Run despite nothing in `internal/extractors/...` pointing at it, because it digest-pins the whole graph.
- `gofmt -l ./cmd ./internal` — clean. `go vet` on both packages — clean.
- No entity/count constant moved, so nothing to re-enumerate.

## Mutants — scored with an exact occurrence count on every edit, restored by `cp` from a `shasum -a256`-verified backup (never `git checkout`), with a green baseline re-verified before each score

| # | Mutant | Verdict |
|---|---|---|
| M1 | `if false && InProcCustomExtractorsEnabled() && …` (the fix disabled — i.e. the pre-fix state) | **DEAD** — `…GateIsOn6960` fails: `SCOPE.DI entities present: []` |
| M2 | drop the gate, dispatch unconditionally (**more permissive**) | **DEAD** — `…GateIsOff6960` fails |
| M3 | drop the `input.TSTree != nil` guard (**more permissive**) | **ALIVE at first.** Production-distinguishable, not equivalent: `dart`/`clojure`/`erlang`/`fsharp`/`nim`/`graphql` have custom extractors and **no** tree-sitter grammar, and their base extractors are source scanners that emit records anyway, so the pass does not fall back — the custom entities would land where the full path emits none. **Graded** with `…HonoursTheNilTreeGuard6960` (dart, `SCOPE.UIComponent|HomePage`), then **DEAD** |
| M4 | `append(records, customEnts…)` in place of `MergeWithCustom` (**more permissive**) | **ALIVE against the first version of the test, which asserted only cardinality.** This is the finding worth carrying forward: **the `n != 1` count assertion could not see it and never fired** — the downstream entity dedupe collapses the duplicate pair, so the cardinality check held on the mutant while the defect was live. What is distinguishable is the *content*: the merge keeps the java method's real span `8-11` where append leaves the annotation line `8-8`, and the `SCOPE.Process` flow entity derived from it disappears. **Graded** by asserting the #6104 span invariant + the flow entity, then **DEAD**. The independent reviewer reproduced this on the first attempt, cardinality blindness included |

M2 also grades the gate function itself: removing the gate is indistinguishable from `InProcCustomExtractorsEnabled()` always returning true, and the two incremental tests differ **only** in the env value.

**Both directions are graded.** `…GateIsOn6960` sees entities wrongly *dropped*; `…GateIsOff6960` and `…HonoursTheNilTreeGuard6960` see entities wrongly *added* — which no "the entities survive" assertion can.

Every test carries its premise as a `t.Fatalf`: that the re-extraction happened at all (the base entity from the edit is present), that `RunCustomExtractors` emits the asserted record for that content (else the absence assertions are vacuous), that the **base** extractor does *not* also emit it (else the gate-on assertion does not distinguish custom dispatch from base extraction), and — for M4's fixture — that the base and custom records genuinely collide on `(SourceFile, Kind, Name)`.

---

### BLOCKER (fixed): the merge orphaned its own `grafel.twin_of` anchor — #6275

`entityRecordToGraphEntity`'s comment predicted this **by name**: the missing pre-stamp-id → final-id remap was safe *"only because … `MergeWithCustom` has no caller on this incremental path"*, and *"if anything ever starts stamping `twin_of` from code reachable here, the #6275 orphaned-anchor bug comes back silently"*. This PR added that caller. Reproduced on the PR's own scala fixture, gate ON: `SCOPE.Type|HomeController` carried `twin_of="861a490cf84f2640"`, matching **no** entity id in the graph.

The cause is mechanical: `enrichFromTwin` stamps `types.EntityRecord.ComputeID()` — `sha256(OrgID+ProjectID+SourceFile+Kind+Name)` — while this path mints `graph.EntityID(repoTag, Kind, Name, SourceFile)`. Different preimages, same record. The full path repoints the facet in `stampEntityIDs`; this path did not, so *"same graph as a full reindex"* was false for merged entity **properties** — the PR's central claim, and exactly where a content-keyed parity comparison is blind (it deliberately does not compare ids).

Fixed as the comment prescribed: **`remapTwinOfAnchors`**, applied in `convertExtractedRecords` before any id is derived. Batch scope is sufficient and stated in the code: the facet and its anchor are the two halves of *one* `MergeWithCustom` decision over *one* file's records. An anchor outside the batch keeps its pre-stamp value rather than acquiring a wrong one — the lookup simply misses. The cheap `recordsCarryTwinOf` pre-scan means a batch with no facet pays no second hash, mirroring `recordsHaveTwinOf` on the full path.

**And the comment is replaced.** It asserted a precondition this change invalidated, which is how the next reader gets misled; it now says what happened, what was measured, and why the remap lives one level up (the mapping is batch-wide; a per-record function cannot express it).

### Gate parity, both halves (was half-done)

The full path's gate is `i.customExtractors || inProcCustomExtractors()`, and the first cut read **only the env half** — leaving the original #6960 defect alive in the `WithCustomExtractors` lane, which `grafel quality` uses and which is the only lane on by default anywhere. The asymmetry was backwards.

The programmatic opt-in now travels in **`ExtractorConfig.InProcCustomExtractors`** (tri-state, `nil` = fall through to env) — the config `TryIncremental` already receives — and **both** dispatch sites evaluate one expression, `extractors.CustomExtractorsEnabled(cfg)`. Config wins over env in **both** directions, per ExtractorConfig's own #2320 precedence, which is why an explicit `false` is expressible and graded.

### R4 graded (~4 lines, as priced)

`MergeWithCustom(customEnts, records)` — arguments swapped — survived the entire suite while moving the facet and narrowing `SCOPE.Type|HomeController` **6-10 → 7-7**. Now asserted in the gate-on test, both halves: the span (which record won) *and* the anchor's target (which record the facet describes), since either alone survives the swap in some configuration.

### `custom_gate.go` now has a grader in its own package

A vocabulary table test next to the helper. Previously the only thing checking the exported gate was the delegating table test in `cmd/grafel`, so a diff touching only `custom_gate.go` produced a package set that never ran its grader — the same shape as the digest-pin trap. It pins the accepted vocabulary (`1`/`true`/`yes`, trimmed) rather than mere truthiness, and asserts `CustomExtractorsEnabled(nil)` and `CustomExtractorsEnabled(&ExtractorConfig{})` agree with it, or the two sites do not in fact share one gate.

### Round-2 mutants — same protocol (exact occurrence count on every edit, green baseline re-verified before each score, restore by `cp` from a `shasum -a256`-verified backup)

| # | Mutant | Verdict |
|---|---|---|
| N1 | `remapTwinOfAnchors` → early `return` (the blocker, un-fixed) | **DEAD** — `…DoesNotOrphanMergeFacetAnchors6960` fails with the reviewer's exact hash `861a490cf84f2640`; `…GateIsOn6960` fails too |
| N2 | `MergeWithCustom(customEnts, records)` — R4 | **DEAD** — span `7-7` (want `6-10`) **and** `grafel.twin_of` empty (want the base record's id) |
| N3 | gate reads the env half only | **DEAD** — both subtests of `…ReadsTheProgrammaticGateHalf6960` |
| N4 | `CustomExtractorsEnabled` ORs config with env instead of config-first (**more permissive**) | **DEAD** — the `config off, env set` subtest. Without this a config could never suppress |
| N5 | `InProcCustomExtractorsEnabled` → `return v != ""` (**more permissive**) | **DEAD** — `…GateVocabulary6960`, now inside `internal/extractors`, where R3 previously found nothing |

### Re-verification (macOS 15.6 / APFS, arm64, in `…/scratchpad/impl-6960-k7q2/wt`)

`go test -count=1 ./internal/extractor/ ./internal/extractors/` — **exit 0** (1.9s / 39.4s). `go test -count=1 ./cmd/grafel/` — **exit 0** (156.9s). `go build ./...` — OK. `go vet` on all three packages — clean. `gofmt -l ./cmd ./internal` — clean.

Unchanged from round 1, and the review confirmed both halves independently: the hop chain holds with no function-value or dispatch-table indirection, and **neither the ratchet nor the digest pin can see any of this** — eight `cmd/grafel` test files set `GRAFEL_INPROC_CUSTOM_EXTRACTORS` and none of them contains the string `Incremental`. These tests are the only protection.

---

## Round 3 — the coordinator's MI-1 found a hazard, and the remap is the first code exposed to it

**MI-1 itself (deleting the `r.Name == ""` guard) is left ALIVE deliberately** — not production-reachable on any constructible input, defensive rather than load-bearing, and manufacturing a fixture to force it would cost more than it protects. The guard stays; no test spent on it.

**What chasing it surfaced is real.** The two id schemes disagree about field boundaries:

| | preimage |
|---|---|
| `types.EntityRecord.ComputeID` | `OrgID + ProjectID + SourceFile + Kind + Name` — concatenated, **no separators** |
| `graph.EntityID` | every field **NUL-separated** |

So `{SourceFile:"x.go", Kind:"A", Name:"BC"}` and `{SourceFile:"x.go", Kind:"AB", Name:"C"}` are **one key and two entities**. `remapTwinOfAnchors` is the first code in the tree to key a map on the **coarser** id while storing the **finer** one, which is what makes the disagreement reachable here and nowhere else. Both records wrote the same slot; the survivor was whichever index came last — incidental, not intended — so a `twin_of` naming that preimage was repointed at an **arbitrary one of two distinct entities**.

That is the #6369 wrong-node hazard, and it is **strictly worse than the dangling anchor this PR fixes**: a dangling `twin_of` names an id no entity carries, so it is inert and detectable, while a confidently-wrong one reads as valid to `classfold`'s `twinAnchors` and to the symbol index's facet-alias rule, which will then alias a name onto the wrong definition.

**Resolution: an ambiguous key is SKIPPED, not resolved, and warned about** (`slog` Warn carrying the pre-stamp id and both final ids). Skipping degrades to exactly the pre-remap behaviour for that one facet — the honest, detectable failure — and it is the only outcome that does not require the function to invent an answer it has no evidence for. Silently picking one, the behaviour before this check, is the option that is not defensible. The comment states which and why.

The underlying `ComputeID` separator problem is broader than this function and is being filed separately; `ComputeID` is deliberately **not** touched here — many call sites already depend on that identity.

### Grading — a direct unit test on `remapTwinOfAnchors`, no fixture, no repo, no corpus

The colliding pair is *constructed from the hash property*, not discovered in a codebase, and **both names are non-empty** so the test exercises the reachable path rather than the `Name == ""` guard MI-1 targets. Premises are `t.Fatalf`: that `ComputeID` really collides (if it ever gains separators this test must fail loudly, not pass vacuously) and that `EntityID` really does not (otherwise there is no wrong node to point at). **Both collision orders are driven** — the entire hazard is that the survivor is decided by arrival order, so a single order cannot distinguish "skipped" from "happened to pick the one I asserted".

| # | Mutant | Verdict |
|---|---|---|
| MI-2 | disable the collision check (the behaviour before this commit) | **DEAD** — the facet is remapped to `e422d03bc35b0818`, i.e. the *second* record's id, demonstrating the arrival-order pick |
| MI-4 | skip **every** key, not just ambiguous ones (would silently revert the #6275 fix) | **DEAD** — the positive control fails, and so do `…GateIsOn6960` and `…DoesNotOrphanMergeFacetAnchors6960` |

Re-verified after restore: `./internal/extractor/` + `./internal/extractors/` **exit 0** (1.6s / 37.5s), `./cmd/grafel/` **exit 0** (144.5s), `go vet` on all three clean, `gofmt -l ./cmd ./internal` clean.

### DIFF
--- a/cmd/grafel/index.go
+++ b/cmd/grafel/index.go
@@ -3881,6 +3881,17 @@ func (i *Indexer) classifyAndReadWithProgress(ctx context.Context, absRepo strin
 	// can consult feature toggles via the Config channel (Config wins; env var
 	// is the backward-compat fallback). Future work: merge a config file on top.
 	extractorCfg := extractor.ConfigFromEnv()
+	// #6960 — carry the PROGRAMMATIC half of the custom-extractor gate
+	// (WithCustomExtractors) in the config rather than only in Indexer state, so
+	// that one expression — extractors.CustomExtractorsEnabled — decides it on
+	// this path and on the daemon's incremental path, which has no Indexer and
+	// receives nothing but an *ExtractorConfig. Only the opt-in is stamped: a
+	// false here would OVERRIDE a set env var, which is not what an unset option
+	// means.
+	if i.customExtractors {
+		on := true
+		extractorCfg.InProcCustomExtractors = &on
+	}
 
 	// Use a shared atomic counter so all workers contribute to the same
 	// tick cadence without needing an additional mutex acquisition per file.
@@ -4138,7 +4149,7 @@ func (i *Indexer) classifyAndReadWithProgress(ctx context.Context, absRepo strin
 				// the guard a file that FAILED to parse would still produce
 				// custom entities. See
 				// TestInProcCustomExtractorsNilTreeGuardIsLoadBearing.
-				if (i.customExtractors || inProcCustomExtractors()) && file.TSTree != nil {
+				if extractors.CustomExtractorsEnabled(&extractorCfg) && file.TSTree != nil {
 					customEnts, customErrs := extractors.RunCustomExtractors(ctx, file)
 					if len(customErrs) > 0 && verbose() {
 						for _, ce := range customErrs {
--- a/cmd/grafel/inproc_custom.go
+++ b/cmd/grafel/inproc_custom.go
@@ -1,8 +1,7 @@
 package main
 
 import (
-	"os"
-	"strings"
+	"github.com/cajasmota/grafel/internal/extractors"
 )
 
 // inProcCustomExtractors reports whether the DEFAULT in-process indexing path
@@ -30,7 +29,10 @@ import (
 // silently disables ~12 framework post-passes, and on a Django fixture it is a
 // NET LOSS (entities 201→186, http_endpoint_definition 18→4). That hole is
 // tracked separately as #6102 and is out of scope here.
+// #6960 — the env read itself now lives in internal/extractors
+// (InProcCustomExtractorsEnabled) so the daemon's incremental re-extract path
+// can consult the SAME gate. It cannot import cmd/grafel, and a second literal
+// reading of the variable here would be free to drift from that one.
 func inProcCustomExtractors() bool {
-	v := strings.TrimSpace(os.Getenv("GRAFEL_INPROC_CUSTOM_EXTRACTORS"))
-	return v == "1" || strings.EqualFold(v, "true") || strings.EqualFold(v, "yes")
+	return extractors.InProcCustomExtractorsEnabled()
 }
--- a/internal/extractor/extractor.go
+++ b/internal/extractor/extractor.go
@@ -54,6 +54,22 @@ type ExtractorConfig struct {
 	// Tri-state: nil means "not set in Config; fall through to env var".
 	JSEmitDestructureDetail *bool
 
+	// InProcCustomExtractors controls whether an in-process indexing pass
+	// dispatches the internal/custom/** framework extractors (issue #5989).
+	// Tri-state: nil means "not set in Config; fall through to the
+	// GRAFEL_INPROC_CUSTOM_EXTRACTORS env var".
+	//
+	// #6960 — this field is how the PROGRAMMATIC half of the full path's gate
+	// (cmd/grafel's WithCustomExtractors, i.customExtractors) becomes visible to
+	// a pass that has no Indexer. The daemon's incremental re-extraction is
+	// handed an *ExtractorConfig and nothing else; without this field its gate
+	// could read only the env half, so a full index opted in through
+	// WithCustomExtractors would emit custom entities that the next incremental
+	// pass silently dropped — #6960 surviving in the other lane. Read it through
+	// extractors.CustomExtractorsEnabled, never directly, so both paths evaluate
+	// one expression.
+	InProcCustomExtractors *bool
+
 	// IncrementalReindexSet records whether IncrementalReindex was explicitly
 	// set via Config (as opposed to being the zero value). This lets callers
 	// distinguish "Config says false" from "Config not consulted".
--- a/internal/extractors/custom_gate.go
+++ b/internal/extractors/custom_gate.go
@@ -3,6 +3,8 @@ package extractors
 import (
 	"os"
 	"strings"
+
+	"github.com/cajasmota/grafel/internal/extractor"
 )
 
 // InProcCustomExtractorsEnabled reports whether an in-process indexing path
@@ -26,4 +28,28 @@ import (
 func InProcCustomExtractorsEnabled() bool {
 	v := strings.TrimSpace(os.Getenv("GRAFEL_INPROC_CUSTOM_EXTRACTORS"))
 	return v == "1" || strings.EqualFold(v, "true") || strings.EqualFold(v, "yes")
+}
+
+// CustomExtractorsEnabled is the WHOLE gate — the config half and the env half —
+// and is what every dispatch site should call.
+//
+// The full in-process path's condition was `i.customExtractors ||
+// inProcCustomExtractors()`. A pass with no Indexer cannot evaluate the first
+// half, and #6960's first cut read only the second, which left the original
+// defect alive in the WithCustomExtractors lane: a full index opted in
+// programmatically emits custom entities, and an incremental pass that saw only
+// an unset env var would drop them on the next edit. Routing the programmatic
+// opt-in through ExtractorConfig.InProcCustomExtractors — which TryIncremental
+// already receives — makes the two paths evaluate the same expression over the
+// same inputs.
+//
+// CONFIG WINS OVER ENV, deliberately, and in BOTH directions: an explicit
+// `false` in Config suppresses a set env var. That is ExtractorConfig's own
+// established precedence (#2320, "Config-first/env-fallback") and the tri-state
+// pointer is what distinguishes "Config says no" from "Config did not say".
+func CustomExtractorsEnabled(cfg *extractor.ExtractorConfig) bool {
+	if cfg != nil && cfg.InProcCustomExtractors != nil {
+		return *cfg.InProcCustomExtractors
+	}
+	return InProcCustomExtractorsEnabled()
 }
--- a/internal/extractors/custom_gate.go
+++ b/internal/extractors/custom_gate.go
@@ -1,0 +1,29 @@
+package extractors
+
+import (
+	"os"
+	"strings"
+)
+
+// InProcCustomExtractorsEnabled reports whether an in-process indexing path
+// should additionally dispatch the custom/framework extractors registered under
+// internal/custom/** (issue #5989) via RunCustomExtractors.
+//
+// ONE READER, TWO CALLERS. The env var used to be read only by
+// cmd/grafel/inproc_custom.go, which is why the daemon's incremental path could
+// not consult it without importing cmd/grafel. #6960: the gate lives here, in
+// the package that owns the dispatcher, and cmd/grafel/inproc_custom.go
+// delegates. A second literal reading of the same variable would be free to
+// drift from this one — that drift is the failure this placement prevents, and
+// it is exactly the failure that made incremental and full disagree.
+//
+// The default stays OFF; see cmd/grafel/inproc_custom.go for the cost rationale
+// and internal/extractors/incremental.go's dispatch site for why the incremental
+// path must read the SAME gate rather than choosing its own answer: the property
+// incremental owes its caller is "same graph as a full reindex", and a pass that
+// runs custom extractors when the full path did not is as wrong as one that
+// skips them when the full path did.
+func InProcCustomExtractorsEnabled() bool {
+	v := strings.TrimSpace(os.Getenv("GRAFEL_INPROC_CUSTOM_EXTRACTORS"))
+	return v == "1" || strings.EqualFold(v, "true") || strings.EqualFold(v, "yes")
+}
--- a/internal/extractors/incremental.go
+++ b/internal/extractors/incremental.go
@@ -1077,6 +1077,47 @@ func tryIncremental(ctx context.Context, repoPath, stateDir string, logger *log.
 		// daemon runs this on a watcher tick, and a panic here would take down a
 		// path whose whole job is to not lose the user's graph.
 		records, extErr := safeExtract(ctx, ext, input)
+
+		// #6960 — Pass 2: the custom/framework extractors under internal/custom/**.
+		//
+		// THIS PATH HAD NO CUSTOM DISPATCH AT ALL. The ~340 framework extractors
+		// register under PREFIXED registry keys ("python_django", "custom_scala_di"),
+		// and the exact-key `Get(cr.Language)` above can never match one, so
+		// RunCustomExtractors is the only thing that selects them —
+		// custom_dispatch.go:192. It had two non-test call sites and neither was
+		// on this path, so a re-extraction emitted the base records only.
+		//
+		// THE GATE IS THE FIX, NOT THE CALL. By DEFAULT the full in-process path
+		// does not dispatch these either (cmd/grafel/inproc_custom.go), so the
+		// default-config graph was never wrong — the initial index emitted no
+		// custom entities to lose. Under GRAFEL_INPROC_CUSTOM_EXTRACTORS=1 it
+		// WAS: the full index emitted them, Step 5 above evicted them on the next
+		// edit, and this loop re-added nothing, so a changed file silently lost
+		// every custom entity while every unchanged file kept its own. Reading
+		// the same gate makes the two paths agree in BOTH directions — including
+		// the one an "entities survive" assertion cannot see, where incremental
+		// would ADD entities a default full reindex never produced.
+		//
+		// SEAM PLACEMENT IS LOAD-BEARING, exactly as at cmd/grafel/index.go:4138:
+		// after safeExtract (so base records exist to merge into) and BEFORE the
+		// Close() below, because custom extractors walk the same parse tree.
+		// Moving it after the Close is a use-after-free, not merely a no-op. The
+		// `input.TSTree != nil` guard is carried for the same reason it is there:
+		// a subset of custom extractors work on CONTENT and emit entities with no
+		// tree at all, so without it a file that FAILED to parse would still
+		// produce custom entities the full path would not.
+		if InProcCustomExtractorsEnabled() && input.TSTree != nil {
+			customEnts, customErrs := RunCustomExtractors(ctx, input)
+			for _, ce := range customErrs {
+				logger.Printf("incremental: custom extractor on %s: %v", rel, ce)
+			}
+			if len(customEnts) > 0 {
+				// MergeWithCustom, not append — the same #6104 non-destructive
+				// (SourceFile, Kind, Name) merge the full path applies.
+				records = MergeWithCustom(records, customEnts)
+			}
+		}
+
 		if input.TSTree != nil {
 			// Release the CGo allocation now rather than deferring to the end of
 			// the loop: peak RSS must stay at one live tree, not len(reallyChanged).
--- a/internal/extractors/incremental.go
+++ b/internal/extractors/incremental.go
@@ -1090,23 +1090,38 @@ func tryIncremental(ctx context.Context, repoPath, stateDir string, logger *log.
 		// THE GATE IS THE FIX, NOT THE CALL. By DEFAULT the full in-process path
 		// does not dispatch these either (cmd/grafel/inproc_custom.go), so the
 		// default-config graph was never wrong — the initial index emitted no
-		// custom entities to lose. Under GRAFEL_INPROC_CUSTOM_EXTRACTORS=1 it
-		// WAS: the full index emitted them, Step 5 above evicted them on the next
-		// edit, and this loop re-added nothing, so a changed file silently lost
-		// every custom entity while every unchanged file kept its own. Reading
-		// the same gate makes the two paths agree in BOTH directions — including
-		// the one an "entities survive" assertion cannot see, where incremental
-		// would ADD entities a default full reindex never produced.
+		// custom entities to lose. With the gate ON it WAS: the full index
+		// emitted them, Step 5 above evicted them on the next edit, and this loop
+		// re-added nothing, so a changed file silently lost every custom entity
+		// while every unchanged file kept its own. Reading the same gate makes
+		// the two paths agree in BOTH directions — including the one an "entities
+		// survive" assertion cannot see, where incremental would ADD entities a
+		// default full reindex never produced.
 		//
-		// SEAM PLACEMENT IS LOAD-BEARING, exactly as at cmd/grafel/index.go:4138:
-		// after safeExtract (so base records exist to merge into) and BEFORE the
-		// Close() below, because custom extractors walk the same parse tree.
-		// Moving it after the Close is a use-after-free, not merely a no-op. The
-		// `input.TSTree != nil` guard is carried for the same reason it is there:
-		// a subset of custom extractors work on CONTENT and emit entities with no
-		// tree at all, so without it a file that FAILED to parse would still
-		// produce custom entities the full path would not.
-		if InProcCustomExtractorsEnabled() && input.TSTree != nil {
+		// BOTH HALVES OF THE GATE, not just the env var. The full path's
+		// condition is CustomExtractorsEnabled over the same ExtractorConfig, so
+		// the programmatic opt-in (cmd/grafel's WithCustomExtractors, which
+		// `grafel quality` uses and which is the only lane on by default
+		// anywhere) reaches this decision too. An env-only read here left the
+		// original defect alive in exactly that lane.
+		//
+		// SEAM PLACEMENT IS LOAD-BEARING, exactly as at cmd/grafel/index.go: after
+		// safeExtract (so base records exist to merge into) and BEFORE the Close()
+		// below, because custom extractors walk the same parse tree.
+		//
+		// MOVING IT AFTER THE Close() IS A SILENT NO-OP, NOT A CRASH, and an
+		// earlier version of this comment claiming a use-after-free was wrong —
+		// measured: the block below sets `input.TSTree = nil` immediately after
+		// Close(), so a dispatch moved past it fails the `!= nil` guard and simply
+		// never fires. It runs to completion and quietly emits nothing, which is
+		// worse to debug than a crash, not better.
+		//
+		// The `input.TSTree != nil` guard is carried for the reason the full
+		// path's is: a subset of custom extractors work on CONTENT and emit
+		// entities with no tree at all, so without it a file that FAILED to parse
+		// — or a language with no grammar at all, which is the case the dart test
+		// drives — would produce custom entities the full path never emits.
+		if CustomExtractorsEnabled(cfg) && input.TSTree != nil {
 			customEnts, customErrs := RunCustomExtractors(ctx, input)
 			for _, ce := range customErrs {
 				logger.Printf("incremental: custom extractor on %s: %v", rel, ce)
@@ -2050,18 +2065,21 @@ func entityRecordToGraphEntity(r types.EntityRecord, repoTag string) graph.Entit
 	// with the wrong id, so nothing downstream of it dangles. It surfaced only
 	// because a flow property happens to embed the id as text.
 	id := graph.EntityID(repoTag, r.Kind, r.Name, r.SourceFile)
-	// #6275 — r.Properties (and therefore any grafel.twin_of a #6104 merge
-	// facet carries) is copied VERBATIM here, with no equivalent of
-	// cmd/grafel/index.go's stampEntityIDs remap (old ComputeID() -> this
-	// freshly computed `id`). That is harmless TODAY only because the sole
-	// twin_of writer, internal/extractors/custom_dispatch.go's
-	// enrichFromTwin, is reachable exclusively from the full-index path
-	// (MergeWithCustom has no caller on this incremental path — see
-	// classfold.go's FoldFrameworkClassKinds doc comment, "TryIncremental
-	// runs no cross extractors at all"). If anything ever starts stamping
-	// twin_of from code reachable here, the #6275 orphaned-anchor bug comes
-	// back silently: this function would need the same pre-stamp-id ->
-	// final-id remap stampEntityIDs performs.
+	// #6275 — r.Properties (and therefore any grafel.twin_of a #6104 merge facet
+	// carries) is copied VERBATIM here. This function performs no remap of its
+	// own, and it must not: the old comment here said the omission was safe
+	// because "MergeWithCustom has no caller on this incremental path", and
+	// #6960 ADDED that caller (the custom dispatch in the Step 6 re-extract
+	// loop), so enrichFromTwin now writes twin_of from code reachable here. The
+	// predicted failure was real and was measured: SCOPE.Type|HomeController
+	// landed carrying twin_of="861a490cf84f2640" — a ComputeID() hash — while
+	// the anchor's actual graph id is the graph.EntityID above, a different
+	// preimage entirely, so the facet named nothing.
+	//
+	// The remap is done ONE LEVEL UP, in convertExtractedRecords via
+	// remapTwinOfAnchors, and it has to be: the mapping is pre-stamp id ->
+	// final id over the WHOLE batch, and this function sees one record. A
+	// per-record fix is not expressible here.
 	return graph.Entity{
 		ID:            id,
 		Name:          r.Name,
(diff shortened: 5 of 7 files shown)


cajasmota:

The diagnosis is right, the hop chain checks out, the incrementality pin is real, and M4 is the genuinely
good finding it is claimed to be — I reproduced it. But the merge this PR newly makes reachable
**resurrects #6275**, and the code comment that predicted exactly that is still sitting five hundred lines
below the change asserting the opposite.

Measured in my own checkout, macOS 15.6 / APFS / arm64:
`…/scratchpad/review-6965-zx9/wt` @ `8d26a041f`. `go vet ./internal/extractors/ ./cmd/grafel/` clean;
`go test -count=1 ./internal/extractors/ ./cmd/grafel/` **exit 0** (41.3s / 147.1s); `gofmt -l ./cmd ./internal`
clean. Restores by `cp` from a `shasum -a256`-verified backup; `git status --porcelain` empty at the end.

---

### BLOCKER — the merge reintroduces #6275: an orphaned `grafel.twin_of` anchor

`internal/extractors/incremental.go:2053-2064` says, in the entity-stamping function this path returns through:

> `r.Properties` (and therefore any `grafel.twin_of` a #6104 merge facet carries) is copied VERBATIM here,
> with no equivalent of `cmd/grafel/index.go`'s `stampEntityIDs` remap … That is harmless TODAY only because
> the sole `twin_of` writer … **is reachable exclusively from the full-index path (`MergeWithCustom` has no
> caller on this incremental path** …). **If anything ever starts stamping `twin_of` from code reachable
> here, the #6275 orphaned-anchor bug comes back silently.**

This PR adds that caller at `incremental.go:1117`. The PR body already notes the merge emits a `twin_of`
facet (that is what it credits the `class_kind_folds` 1 → 0 change to) — so the precondition the comment
names is now met, and the comment is now a false statement about its own file.

It is not theoretical. Probe on the PR's own scala fixture, gate ON:

```
DUMP SCOPE.Type|HomeController id=0735f83d47278f5e span=6-10 twin="861a490cf84f2640"
ORPHANED ANCHOR: grafel.twin_of="861a490cf84f2640" matches NO entity id in the incremental graph
```

Cause is mechanical: `enrichFromTwin` stamps `effectiveID(&be)`, which for an unstamped base record is
`types.EntityRecord.ComputeID()` — `sha256(OrgID+ProjectID+SourceFile+Kind+Name)[:16]`. The incremental path
then mints the real id at `incremental.go:2052` with `graph.EntityID(repoTag, Kind, Name, SourceFile)` —
a different preimage entirely (repo tag, NUL separators, different field order). The full path repoints the
facet (`cmd/grafel/index.go:4702`); this path does not. So the very property the PR sets out to establish —
*"same graph as a full reindex"* — does not hold for the merged entity's properties.

`internal/engine/classfold.go`'s `twinAnchors` and the symbol index both read `twin_of`; on this path it now
names nothing. Please either apply the `stampEntityIDs`-equivalent remap on the incremental side, or (if
that is a bigger job than this PR wants) file it and **replace the `incremental.go:2053` comment**, which a
future reader will otherwise trust.

---

### Mutants I scored (all with an exact occurrence count, green baseline re-verified before each)

| # | Mutant | Verdict |
|---|---|---|
| R1 | drop `input.TSTree != nil` from the new guard (**permissive**) | **DEAD** — `…HonoursTheNilTreeGuard6960` fails. Author's M3 reproduced |
| R2 | `append(records, customEnts...)` for `MergeWithCustom` (**permissive**) | **DEAD** — and the claim about *why* is confirmed: the `n != 1` cardinality `Fatalf` **did not fire**; what failed was the span invariant (`start=8 end=8`) and the missing `SCOPE.Process` flow entity. Author's M4 reproduced exactly, cardinality blindness included |
| R3 | `InProcCustomExtractorsEnabled` → `return v != ""` (**permissive**: accepts `0`/`no`/`false`/`garbage`) | **DEAD**, but *only* in `cmd/grafel` (`TestInProcCustomExtractorsGateParsing`). `./internal/extractors/` is **exit 0** on this mutant — see the note below |
| R4 | `MergeWithCustom(customEnts, records)` — arguments swapped | **ALIVE — distinguishable and ungraded.** Survives the *entire* `./internal/extractors/` suite (exit 0, 37.9s) |
| R5 | move the whole dispatch block **after** `input.TSTree.Close()` | **DEAD** — `…GateIsOn6960` fails. Also disproves a prose claim; see below |

**R4 is the one to fix.** Differential dump of the produced graph, baseline vs mutant, same scala fixture:

```
- SCOPE.Component|HomeController span=6-10 twin=""
+ SCOPE.Component|HomeController span=6-10 twin="8d5b506bda65519a"
- SCOPE.Type|HomeController      span=6-10 twin="861a490cf84f2640"
+ SCOPE.Type|HomeController      span=7-7  twin=""
```

The facet lands on the wrong node **and** `SCOPE.Type|HomeController` narrows 6-10 → 7-7 — the *same*
span-narrowing hazard M4 exists to catch, one fixture over, with nothing watching. Pricing the grade: one
line in `…GateIsOn6960` asserting that entity's span is the class body (6-10), ~3 lines with the message.
That is well under the "too expensive" bar.

**Note on R3.** `internal/extractors/custom_gate.go` has **no test in its own package**; the only thing
grading the new exported helper is the delegating table test in `cmd/grafel`. That is fine while the
delegation stands, but it means a diff touching only `custom_gate.go` yields a package set that never runs
its grader — the same shape as the `cmd/grafel` digest-pin trap the PR body already respects. Consider a
three-line table test next to the helper.

**Note on R5.** The comment says *"Moving it after the `Close` is a use-after-free, not merely a no-op."*
**Measured: it is a no-op, not a use-after-free** — the block immediately below sets `input.TSTree = nil`
after `Close()`, so the carried `input.TSTree != nil` guard makes the moved dispatch simply not fire. R5
runs to completion and fails the gate-on assertion cleanly, no crash. The seam placement is still correct
and still graded; only the stated reason is wrong. Worth fixing, since it is the kind of sentence someone
later reasons from.

---

### The four verdicts asked for

**(a) The hop chain — CONFIRMED, independently, and not by a filename grep.** Every non-test reference to
`RunCustomExtractors` in the tree is either a comment, its own definition (`custom_dispatch.go:208`), or one
of the two call sites (`cmd/grafel/index.go:4142`, `internal/daemon/extract/subproc.go:364`); pre-PR,
`incremental.go` had none. It is never taken as a function value, never entered in a dispatch table, and
`CustomExtractorsFor` has no non-test caller outside `custom_dispatch.go` itself, so there is no registry
indirection to hide behind. The #6957-shaped escape is not present here. R5 is the positive control: putting
the dispatch where it cannot see a tree reproduces the original symptom exactly.

**(b) The incrementality premise pin — REAL, and not vacuous.** `res.Done` and `res.ChangedFiles == 1` are
both `t.Fatalf`, and `Done=false` is precisely the fallback-to-full-index signal. Observed in my run:
`incremental: done changed=1 entities=13 rels=4 … flows_recomputed=true`. A silent full reindex fails these
tests rather than passing them.

**(c) The call site is graded, not just the helper — YES.** R1, R2 and R5 all mutate the call site with
`custom_gate.go` untouched, and all three die. The helper's own parsing is graded separately (R3) and only
from `cmd/grafel` — flagged above, not a blocker.

**(d) Severity, against #6966 — accurately stated, not overstated, with one omission.** "With the gate off
there is nothing to lose" is correct and correctly narrows this. One thing the PR body does not mention:
`cmd/grafel/index.go:4141` is `i.customExtractors || inProcCustomExtractors()`, and the new incremental
dispatch reads **only** the env half. A full index run via `WithCustomExtractors` (per #6966, its sole
non-test caller is `grafel quality`) with the env unset would emit custom entities that a subsequent
incremental pass drops — the original #6960 defect, surviving in the `i.customExtractors` lane. **Alive but
not production-reachable today**, because `grafel quality` never takes the incremental path. Worth one
sentence in the comment so the parity claim is not read as broader than it is.

**Blindness claim — both halves CONFIRMED.** Eight `cmd/grafel` test files set
`GRAFEL_INPROC_CUSTOM_EXTRACTORS`; **zero** of them contain the string `Incremental`. So no `cmd/grafel`
test combines the gate with an incremental pass, the digest pin runs gate-off by construction, and
`./cmd/grafel/` would indeed have been green with the defect in place. These four tests really are the only
thing protecting this — which is why R4 and the #6275 anchor matter more here than they would elsewhere.

---

**Blocking on the #6275 orphaned anchor.** R4's grading and the two comment corrections (the false
`incremental.go:2053` premise, the false use-after-free claim) are cheap and I would take them in the same
pass. Everything else here is solid work — the M4 cardinality finding in particular is exactly right and
reproduced on the first attempt.

*No merge, no push to the branch. Worktree removed; `git status --porcelain` clean.*

cajasmota:
All four addressed in `582974596`. Thanks — the blocker was right, and the comment that predicted it made it findable in one read.

**BLOCKER — #6275 orphaned anchor.** Reproduced your `861a490cf84f2640` exactly. Fixed with `remapTwinOfAnchors`, the pre-stamp-id → final-id remap `entityRecordToGraphEntity`'s comment prescribed, applied in `convertExtractedRecords` before any id is derived. Batch scope is sufficient and now says why in the code: the facet and its anchor are the two halves of one `MergeWithCustom` decision over one file's records, and an anchor outside the batch keeps its pre-stamp value rather than acquiring a wrong one. `recordsCarryTwinOf` gates the second hash, so a facet-free batch pays nothing. **The comment is replaced** — it asserted the precondition this PR invalidated, which is the part that would have misled the next reader.

**Gate parity.** You were right that the asymmetry was backwards. The programmatic half now travels in `ExtractorConfig.InProcCustomExtractors` (tri-state) — the config `TryIncremental` already receives — and both sites evaluate `extractors.CustomExtractorsEnabled(cfg)`. Config wins over env in **both** directions per #2320, so an explicit `false` is expressible; N4 (OR instead of config-first) dies on that.

**R4 graded**, at the price you quoted: the gate-on test now asserts the span **and** the anchor's target, since either alone survives the swap in some configuration. Scored: `7-7` (want `6-10`) plus an empty `twin_of`.

**R5 comment corrected.** You're right that it is a no-op, not a use-after-free — `input.TSTree = nil` follows the `Close()`. The comment now says that, and says it is worse to debug than a crash rather than better. The guard stays for the reason the dart test grades.

**R3 note taken**: `custom_gate.go` has a vocabulary table test in its own package now, pinning the accepted vocabulary rather than truthiness, and asserting `CustomExtractorsEnabled(nil)` / `(&ExtractorConfig{})` agree with it.

Round-2 mutants, same protocol: N1 (remap disabled) DEAD, N2 (R4) DEAD, N3 (env-only gate) DEAD, N4 (OR not config-first) DEAD, N5 (`v != ""`, now graded inside `internal/extractors`) DEAD. Re-verified: `./internal/extractor/` + `./internal/extractors/` exit 0, `./cmd/grafel/` exit 0 (156.9s), `go build ./...` OK, vet/gofmt clean.

Also folded your two confirmations into the PR body: the cardinality assertion **not firing** on R2 is now stated as the point of that row rather than a footnote, and the ratchet/digest-pin blindness is stated with your eight-files-zero-`Incremental` evidence.

cajasmota:
## Coordinator mutant on `582974596`

Own detached worktree, `git status --porcelain` empty, restored by `cp` from a SHA-256-verified backup (`b0086435…` both ways), baseline green immediately before (53.0 s) and after (36.7 s).

I targeted the new remap, outside everything you and the reviewer scored (nil-tree guard, swapped merge args, gate parsing, dispatch-after-`Close`, remap-disabled, env-only gate, OR-vs-config-first, `v != ""`).

**MI-1 — delete the `if r.Name == "" { continue }` guard in `remapTwinOfAnchors`**, so unnamed records also enter `preStampToFinal`. Edit proven landed (guard count 1 → 0), `go vet` clean.

**Result: ALIVE** — `./internal/extractors/` green, exit 0, 39.6 s.

## The verdict, and the finding underneath it

Chasing reachability turned up something worth more than the mutant.

**`ComputeID` and `EntityID` disagree about where one field ends and the next begins.**

```go
// types.EntityRecord.ComputeID — bare concatenation, NO separators
h.Write([]byte(e.OrgID + e.ProjectID + e.SourceFile + e.Kind + e.Name))

// graph.EntityID — NUL-separated
h.Write([]byte(repo)); h.Write([]byte{0})
h.Write([]byte(kind)); h.Write([]byte{0})
h.Write([]byte(name)); h.Write([]byte{0})
h.Write([]byte(sourceFile))
```

Demonstrated with a throwaway probe (since removed; tree clean at `582974596`):

```
ComputeID  A=06e77f0326cac892  B=06e77f0326cac892  collide=true
EntityID   A=8cb5cdf8e9141db1  B=96318bcb45369427  collide=false
```

— where `A = {SourceFile:"x.go", Kind:"AB", Name:""}` and `B = {SourceFile:"x.go", Kind:"A", Name:"B"}`.

**Why this lands on your PR specifically.** `remapTwinOfAnchors` builds a map **keyed by the coarser id** (`ComputeID`) whose values are the **finer** id (`EntityID`). Two records that collide under the key therefore write one entry, and the survivor is whichever the range hits last — map iteration over `records` is index-ordered here, so it is the later index, but that is incidental rather than intended. A `twin_of` naming that preimage is then repointed to **one of two different entities**, chosen arbitrarily.

That is the #6369 wrong-node hazard, and it is strictly worse than the dangling anchor this PR fixes: a dangling `twin_of` is detectably broken, a confidently-wrong one is not.

**MI-1's honest classification: distinguishable only through such a collision.** Removing the guard widens the population entering the map and so widens the collision surface — but I could not construct a *realistic* corpus input where an unnamed record's preimage equals a named anchor's, and I am not going to manufacture a fixture to force one. So: **alive, not production-reachable on any input I could build**, with the distinguishing input named precisely (a `Kind`/`Name` boundary collision). Keep the guard — it narrows the surface — but it is defensive, not load-bearing, and I would not spend a test on it.

## What I would like instead, and it is cheap

The collision is pre-existing and not yours to fix. But **your remap is the first code to key a map on `ComputeID` and store `EntityID`**, which is exactly where a boundary collision turns into a wrong node rather than a duplicate.

Add the collision check while you are in there — a few lines:

```
if prior, seen := preStampToFinal[preStampID]; seen && prior != finalID {
    // two records collide under ComputeID's separator-free preimage but are
    // distinct under EntityID — repointing would pick one arbitrarily
}
```

Then either skip the remap for that key or log it. Either is defensible; **silently picking one is not**, and today that is what happens. Say in the comment which you chose and why.

I am filing the underlying `ComputeID` separator problem separately — it is broader than this PR and touches anything that treats `ComputeID` as an identity.

## The rest checks out

Your round-2 work is sound from here. The remap is applied in `convertExtractedRecords` **before** any id is derived, which is the ordering the old comment prescribed, and you replaced that comment rather than leaving it asserting the precondition this PR invalidated — that was the actual defect the reviewer found, and it is properly closed.

Two things worth keeping in the body, which I see you added: the `n != 1` cardinality assertion **never fired** while the defect was live, and the ratchet/digest-pin blindness is now evidenced (eight `cmd/grafel` files set the gate, zero mention `Incremental`). Those are the two facts a future reader needs.

Merging once CI is green, with the collision check.

cajasmota:
Done in `b765d6fc3`. MI-1 left alive as you asked — guard kept, no test spent on it.

**Resolution chosen: skip the remap for an ambiguous key, and warn.** Reasoning, and it is in the comment: skipping degrades to exactly the pre-remap behaviour for that one facet — a `twin_of` naming an id no entity carries, which is inert and detectable — whereas resolving it would require the function to invent an answer it has no evidence for. A dangling anchor announces itself; a confidently-wrong one is read as valid by `classfold`'s `twinAnchors` and the symbol index's facet-alias rule. The `slog` Warn carries the pre-stamp id and *both* final ids, so the collision is diagnosable without a repro.

The comment also records why this function is where the hazard lands: it is the first code in the tree to key a map on the coarser id (`ComputeID`, no separators) while storing the finer one (`EntityID`, NUL-separated).

**Graded** with a direct unit test on `remapTwinOfAnchors` — no fixture, no corpus. Two notes on its construction:

- The pair is `{Kind:"A",Name:"BC"}` / `{Kind:"AB",Name:"C"}`, **both names non-empty**. Your probe pair uses `Name:""`, which the `r.Name == ""` guard skips — a collision built on it would be testing the guard MI-1 targets rather than the reachable path.
- **Both orders are driven.** The hazard *is* arrival order, so a single ordering cannot distinguish "skipped" from "happened to pick the one I asserted".

Premises are `Fatalf`: that `ComputeID` still collides (if it ever gains separators this test fails loudly rather than passing vacuously) and that `EntityID` still does not.

Mutants: **MI-2** (check disabled) **DEAD** — remaps to `e422d03bc35b0818`, the *second* record's id, showing the arrival-order pick outright. **MI-4** (skip every key, not just ambiguous ones — the silent revert of the #6275 fix) **DEAD** on the positive control plus both round-2 anchor tests.

`./internal/extractor/` + `./internal/extractors/` exit 0, `./cmd/grafel/` exit 0 (144.5s), vet/gofmt clean, `git diff HEAD~1` reviewed. `ComputeID` untouched — scoped to the collision check as instructed.

PR #6967, opened 2026-09-07
fix(#6964): sparse-checkout support was inert — the probe read a scope git stopped writing in 2.32, and cone's `/*` was read as match-everything

### LINKED ISSUE #6964: Sparse-checkout support is inert on current git: the probe reads --local (git ≥2.32 writes worktree scope) and cone's leading /* is read as match-everything
Found during #6940 (PR #6961) by the implementation agent, independently confirmed by the adversarial reviewer on **git 2.50.1 (Apple Git-155)**, and re-verified from source by me. Three parties, and — importantly — the source read is a *different* method from the runtime probe, so these are not one confirmation wearing three hats.

**Two independent defects, either of which alone makes grafel's sparse-checkout support inert on current git. Together they mean a sparse repo is indexed in full, silently.**

## Defect 1 — the probe reads the wrong config scope

`internal/gitmeta/sparse.go:82` and `:88`:

```go
enabled := RunGit(repoPath, "config", "--local", "--get", "core.sparseCheckout")
cone    := RunGit(repoPath, "config", "--local", "--get", "core.sparseCheckoutCone")
```

Since **git 2.32**, `git sparse-checkout init/set` writes `core.sparseCheckout` to the **worktree-scoped** config (`.git/config.worktree`), not `.git/config`. `--local` does not read worktree scope, so on any repo made sparse by a modern git the probe returns empty and grafel concludes **"not sparse"**.

Confirmed at runtime on git 2.50.1: `--local` returns nothing; the value is in `.git/config.worktree`.

## Defect 2 — cone patterns are read as match-everything

`sparse.go:157-165`:

```go
p = strings.TrimPrefix(p, "/")
...
if p == "" || p == "*" {
    match = true // wildcard / root
}
```

Git's **cone mode** writes exactly this pattern set into `.git/info/sparse-checkout`:

```
/*
!/*/
/src/
```

`TrimPrefix(p, "/")` turns `/*` into `*`, which hits `match = true`. So the very first pattern of every cone-mode sparse file makes `IsPathIncluded` return true for **every path**, and the `/src/` restriction that expresses the user's actual intent is never reached.

The adjacent comment is also wrong, and it is why this survived:

```go
// Negation patterns (!) are not implemented here — cone mode doesn't
// use them; non-cone repos with negation patterns are rare.
```

**Cone mode does use them** — `!/*/` is the second line git itself writes. The comment asserts the exact thing that would have made someone look.

## Why it went unnoticed

Both defects fail toward **inclusion**, which is invisible: a sparse repo indexes fully, producing a *superset* of the right answer. Nothing looks broken — there is no error, no missing entity, only extra work and extra graph. The failure mode is silence, the same class as #6321, #6940 and #6946.

This also explains why #6961's sparse convergence test needed a `config --local` mirror line to work at all: the test writes the config where the probe looks, rather than where git puts it. **The test is therefore not evidence that sparse works in production** — it is evidence that the code is self-consistent with its own fixture. Worth stating plainly, because a green sparse test is exactly what would stop someone re-checking this.

## What a fix must establish

1. **Read the scope git actually writes.** `git config --get` without `--local` (or explicitly `--worktree` with a `--local` fallback) — and establish which git versions each spelling supports, since the whole defect is a version-scope mismatch. Do not swap one hardcoded scope for another without saying what happens on older git.
2. **Implement cone semantics properly**, including negation. `/*` + `!/*/` + `/src/` must include `src/**` and exclude everything else at the root. The current "treat negation as an excluder" comment needs to become real behaviour or an explicit, tested limitation.
3. **Grade it against real git**, not against a hand-written config. Drive `git sparse-checkout set src` in a `t.TempDir()` repo, let git write whatever it writes, and assert grafel's view matches `git ls-files`. A fixture that hand-writes the config reproduces defect 1 by construction — that is precisely how this survived.
4. **Assert the exclusion direction.** Both defects fail toward inclusion, so any test that only checks "included paths are included" passes against fully-broken code. Assert that an *excluded* path is excluded.
5. **State the git version measured.** These are version-dependent by nature.

## Scope note

Not fixed in #6961 — correctly out of scope there, and the implementation agent flagged rather than quietly patched it. #6961's sparse test remains valid for the axis it pins (poll convergence), with the caveat above.

Refs #6940, #6961.

cajasmota:
Closes #6964.

grafel's sparse-checkout support did nothing on any current git. Two independent defects, either of which alone makes it inert; together, a sparse repo was indexed **in full, silently**.

Both were re-verified at runtime here before being fixed, on **git 2.50.1 (Apple Git-155), macOS/APFS** — and every version-dependent claim below is now re-derived by the tests at runtime (`rgSparseScopes` prints the scopes *this* git used into any failure), so a different git reports what it actually does instead of being assumed.

## Defect 1 — the probe read a scope git stopped writing in 2.32

`ProbeRepo` asked `git config --local --get core.sparseCheckout`. Since git 2.32, `git sparse-checkout init/set` turns on `extensions.worktreeConfig` and writes `core.sparseCheckout` / `core.sparseCheckoutCone` into the **worktree** scope (`<git-dir>/config.worktree`), which `--local` cannot read. Measured in a `t.TempDir()` repo:

```
--local=<unset/error>   --worktree=true   (no scope flag)=true
```

**The fix passes no scope flag at all**, which is not "swapping one hardcoded scope for another" — it is git's own resolution order (worktree → local → global → system), so the probe agrees with the working tree git actually produced:

| case | `--local` (before) | `--get` (after) |
|---|---|---|
| git ≥ 2.32, `sparse-checkout set` | **misses it** (worktree scope) | reads it |
| git < 2.32, `sparse-checkout init` | reads it (local scope) | reads it — `--get` covers local too |
| value set by hand in `--local` | reads it | reads it |
| value inherited from `--global`/`--system` | misses it | reads it — **and git honours it**: driving `git read-tree -mu HEAD` under `GIT_CONFIG_GLOBAL` with `core.sparseCheckout=true` really does produce a sparse working tree |

So the new spelling is **strictly wider** than the one it replaces: nothing that used to be detected stops being detected, on any version.

`--worktree --get` was considered and rejected: it is a hard error (exit 1) on any repo that has not enabled `extensions.worktreeConfig` — i.e. on every non-sparse repo and every repo made sparse by git < 2.32.

`--bool` is added alongside, so `core.sparseCheckout = 1` / `yes` / `on` is read the way git reads it rather than falling through the old literal `== "true"` comparison.

## Defect 2 — cone's `/*` was read as match-everything

`IsPathIncluded` did `p = strings.TrimPrefix(p, "/")` and then `if p == "" || p == "*" { match = true }`. Git's cone mode writes exactly:

```
/*
!/*/
/src/
```

`/*` normalised to `*`, matched everything, returned "included" on the **first** line, and the `/src/` restriction that expresses the user's actual intent was never reached. The adjacent comment — *"Negation patterns (!) are not implemented here — cone mode doesn't use them"* — asserted the exact thing that would have made someone look: `!/*/` is the **second line git itself writes**. That comment is gone, replaced by behaviour.

### What the matcher does now

Sparse patterns are gitignore syntax with the sense inverted: a match **includes**, a `!` match **excludes**, and the **last** match wins. A pattern matching a *directory* carries its verdict down until something deeper matches.

So the decision walks the path prefix by prefix — `a`, `a/b`, `a/b/c.go` — evaluating the whole list against each prefix, last match setting a running verdict that the next prefix inherits. Patterns are parsed properly: negation, directory-only (`/` suffix), anchoring (leading slash *or* an inner slash, per gitignore), `**` as zero-or-more components, and `path.Match` for the rest (so `*` never crosses `/`, and `?`, `[...]`, `\` escapes work).

`ConeMode` **no longer selects an algorithm**, and its doc comment now says so. Cone mode is a restriction on what git *writes*, not on how those lines are read — cone patterns are a subset of the general syntax, so one matcher answers both. The flag is still reported, because callers surface it.

One rule that gitignore and sparse-checkout do **not** share, and which is measured rather than read off the docs: an excluded directory does **not** prune its subtree. `/*` + `!/*/` + `/services/payments/` checks out `services/payments` while leaving `services/orders` absent. A matcher that prunes gets this wrong and still matches git on the simple cone case — hence a test of its own.

## Why nobody noticed

Both defects fail toward **inclusion**. A sparse repo indexed in full is a *superset*: no error, no missing entity, just extra work and extra graph. Same silent class as #6321, #6940, #6946.

## Grading — against real git, never a hand-written config

`internal/gitmeta/sparse_realgit_6964_test.go`. **Nothing in it hand-writes a config value or a pattern file** — that is the single rule that matters here, because a hand-written fixture reproduces defect 1 *by construction*, and that is precisely how this survived. Every case drives `git sparse-checkout set …` in a `t.TempDir()` repo and lets git write whatever it writes.

**Ground truth is git's own per-path verdict.** `git ls-files -v` tags every index entry; the `S` tag is skip-worktree, i.e. *not* checked out. `assertMatchesGit` compares grafel's answer to git's for **every tracked path**, so the **exclusion direction is asserted for every path in every case** — and it `t.Fatal`s if either direction turns out to be unexercised, because an all-included comparison is satisfied by exactly the broken behaviour.

Cone and non-cone are **separate axes**, and no fixture stands in for another:

| test | axis |
|---|---|
| `ProbeRepo_SeesASparseCheckoutMadeByRealGit` | defect 1 alone — probe sees what real git just created |
| `ProbeRepo_DisableIsSeenAsNotSparse` | the other direction — a wider scope read must not make a repo look sparse forever |
| `Sparse_ConeMatchesGit` | cone, one directory — the `/*` + `!/*/` + `/src/` shape defect 2 broke |
| `Sparse_NestedConeMatchesGit` | cone, nested — git adds `!/services/*/`, so a matcher that mishandles the second-level negation passes the flat case and fails here |
| `Sparse_NonConeMatchesGit` | non-cone, anchored pattern written verbatim |
| `Sparse_NonConeUnanchoredMatchesGit` | non-cone, no leading slash |
| `Sparse_NonConeDoubleStarMatchesGit` | `**` spanning **two** components |
| `Sparse_NonConeNegationMatchesGit` | negation where a later include takes back an earlier exclude |
| `Sparse_ConeRootFilesStayIncluded` | the surprising cone rule stated on its own: root **files** stay, root **directories** go |
| `Sparse_ReInclusionUnderAnExcludedDirectory` | re-inclusion below an excluded directory (sparse ≠ gitignore) |
| `Sparse_BooleanSpelling` | `1`/`yes`/`on` are sparse; `0` is not |

## Mutants (round 1) — 13 scored, 13 DEAD

Preferring the **more permissive** direction, since that is the direction both shipped defects already fail in. Each edit was proved landed by an exact occurrence count, with a green baseline immediately before and a restore by `cp` from a SHA-256-verified backup (never `git checkout --`).

Defects 1 and 2 were scored **separately**, with the other fix kept in place — scored together, either one masks the other and neither is graded.

| # | mutant | verdict |
|---|---|---|
| M1 | probe back to `--local --get` (**defect 1 alone**, defect-2 fix kept) | DEAD — 8 tests |
| M2 | old trim-prefix matcher restored (**defect 2 alone**, defect-1 fix kept) | DEAD — 5 tests |
| M3 | negation ignored, every match includes (*permissive*) | DEAD |
| M4 | first match wins instead of last (*permissive* — cone's `/*` is first) | DEAD |
| M5 | directory-only gate dropped, so `!/*/` also matches root files | DEAD |
| M6 | anchoring dropped, every pattern floats (*permissive*) | DEAD |
| M7 | parent verdict not inherited by deeper prefixes | DEAD |
| M8 | no prefix walk, leaf path only | DEAD |
| M9 | `--bool` dropped from the probe | DEAD (**see below**) |
| M10 | `**` loses its zero-or-more-components meaning | DEAD (**see below**) |
| M11 | empty pattern set includes everything (*permissive*) | DEAD |
| M12 | `ConeMode` short-circuits to include-all — the shape of the original bug's excuse (*permissive*) | DEAD |
| M13 | a pattern need not consume the whole path (*permissive*) | DEAD |

**M9 and M10 were ALIVE on the first scoring, and both were vacuous fixtures rather than acceptable gaps.** Recording them because each is an instance of a pattern worth expecting:

- **M9** — the boolean-spelling case ran `sparse-checkout set` and *then* wrote the spelling under test into `--local`. Scope precedence is system < global < local < **worktree**, so the `true` git had just written into the worktree scope outranked it and the subtest passed no matter what the probe did. Fixed by clearing the worktree-scoped flag first (`rgClearWorktreeFlag`, which asserts the clear worked), leaving git's own pattern file untouched.
- **M10** — the `**` case matched a directory **one** level down, where `path.Match("**", "src")` answers true on its own and a matcher with no `**` support looks correct. Fixed by moving the target **three** levels down so `**` must swallow two components.

## The `--local` mirror lines in #6961 — deleted here

#6961 **merged** as `6fbea8186` (an earlier draft of this body said it was still open — stale). This branch is rebased onto it and both mirror lines in `internal/daemon/watch/poll_convergence_6940_test.go` are **deleted in this PR** rather than left as a follow-up:

- `config --local core.sparseCheckout true` — unnecessary: the probe reads the scope git writes.
- `config --local core.sparseCheckoutCone true` — **inert, not "actively wrong"** as an earlier draft said. `sparse-checkout set --no-cone` writes `core.sparseCheckoutCone=false` into the *worktree* scope, which outranks local, so the no-scope-flag probe already resolves `false`. It was actively wrong only under the old `--local` probe.

The comment that justified `--no-cone` by pointing at defect 2 is rewritten: `--no-cone` stays because a verbatim pattern is the simplest shape for that test's subject (poll convergence), and sparse semantics are graded against real git here instead.

`TestChangePoller_SparseFilteredPathConverges` **passes and does not skip** on the merged tree with this change: `--- PASS (2.92s)`, exit 0.

---

## Review round 2 — blocker fixed, two ungraded mutants killed (`f50379c3c`)

Rebased onto `6fbea8186`. Full independent review: comment 5566980699.

### BLOCKER — the flag without a pattern file would have indexed a repo to **zero files**

Dropping the scope flag is right and strictly wider, and that is exactly the problem: it made the **global/system** scope reachable for the first time, and `git config --global core.sparseCheckout true` is the **pre-2.25 manual sparse workflow** — still in real dotfiles. Re-measured here on git 2.50.1, in a repo with no `<git-dir>/info/sparse-checkout`:

| | |
|---|---|
| git | `ls-files -v` tags **every** path `H`; the working tree is complete |
| grafel, as reviewed | `IsSparse=true, Patterns=nil` → `IsPathIncluded` short-circuits to `false` → the walker drops **every file** |

and it drops them **silently**: sparse is the one walk layer that skips with no `SkipEntry` and nothing on `PrintSkipped`, so there would not even be a `[skip]` line to explain an empty graph. Total loss, in the exclusion direction, newly reachable.

**The gate belongs in `ProbeRepo`, not `IsPathIncluded`.** The constructed-value contract `IsSparse=true, Patterns=nil → exclude` is pinned deliberately by `sparse_test.go` and `sparse_walker_test.go` and is correct; what was wrong is `ProbeRepo` *manufacturing* that value. And it keys on **the file, not the pattern count** — those are different states in git, measured:

| state | git checks out | grafel now |
|---|---|---|
| pattern file **absent** | **everything** | not sparse → full index |
| pattern file present but **empty** | **nothing** | sparse, excludes everything |
| pattern file present, comment-only | nothing | sparse, excludes everything |

An existing-but-*unreadable* file (the #6416 FIFO case) also lands on "not sparse" — the over-inclusive direction, which is the direction this file chooses everywhere. `parseSparsePatternFile` now returns that distinction, and `gitmeta_fifo_6416_test.go` gains an assertion that a refused pattern file reads as *no file* rather than as an empty pattern set, because that is now the difference between a superset and a zero-file graph.

Neither new test hand-writes anything: the global flag is written **by git** into a `GIT_CONFIG_GLOBAL` the test process sets with `t.Setenv` (so `ProbeRepo`'s own git child resolves the same config `git ls-files -v` does), and the empty pattern file is written **by git** via `sparse-checkout set --no-cone --stdin` with empty input.

### Mutants — 4 more scored, 4 DEAD, each on exactly the test it targets

Green baseline immediately before each, edit proved landed by an exact occurrence count, restore by `cp` from a SHA-256-verified backup, re-scored a second time against the final head.

| # | mutant | verdict |
|---|---|---|
| B1 | the reviewed code: no pattern-file gate | DEAD — `ProbeRepo_GlobalFlagWithoutAPatternFileIsNotSparse` |
| B2 | the **wrong** fix: gate on the pattern **count** too | DEAD — `ProbeRepo_EmptyPatternFileStaysSparse` |
| RX1 | `anchored = true` always | DEAD — `Sparse_NonConeTrulyUnanchoredMatchesGit` |
| RX2 | `**` means one-or-more instead of zero-or-more | DEAD — `Sparse_NonConeDoubleStarMatchesGit` |

*(B2's first form left `haveFile` unused and failed to compile. A build break is not a kill; it was re-scored in a compiling form.)*

### Two more vacuous fixtures — and the lesson, since this is now four

RX1 and RX2 were ALIVE because of fixtures whose names claimed more than their bodies reached:

- `TestSparse_NonConeUnanchoredMatchesGit` used `"services/payments"` and `"src/sub"` — **both contain an inner slash, so gitignore anchors both**. The unanchored branch of `sparsePattern.matches` had no fixture anywhere in the tree. Renamed to `..._NonConeInnerSlashIsAnchoredMatchesGit` (a real axis, just not the one the name promised), and a genuine bare-name case added.
- `TestSparse_NonConeDoubleStarMatchesGit` only ever had `deep/` at depth ≥ 2, so `**` was never asked to match **zero** components. A root-level `deep/e.go` is now in the fixture, so one pattern exercises zero *and* more-than-one in one comparison.

**Four of these in one change — two I found myself (M9, M10), two found by review. The generalisable point: driving real git does not by itself guarantee a fixture reaches the axis its name claims.** My own M10 fix moved the `**` target from one level to three, which fixed too *few*: it closed "no `**` support at all" and left "`**` means one-or-more" open. A fixture needs its *named branch* proved reachable, and the cheapest proof is the mutant that ought to kill it.

### Smaller corrections

- Removed the `\!` / `\#` de-escaping in `parseSparsePattern`: `path.Match` reads the backslash as an escape too, so it was a branch that could never change an answer, and therefore could never be graded.
- Documented the one known divergence from git, pre-existing and over-inclusive: a pattern with a **trailing space** (`/src/ `) is stripped by `parseSparsePatternFile`'s `TrimSpace` and read as `/src/`, while git keeps the space and matches nothing.
- `ConeMode`'s doc no longer claims callers surface it — it has **zero non-test readers** in the tree.

### Not changed, and why

The matcher itself. The review drove 22 adversarial pattern shapes through real `git sparse-checkout set` against `git ls-files -v`; 21 matched exactly and the 22nd is the pre-existing trailing-space divergence above. I have no counter-example either.

## Verification

`go test -count=1`, macOS/APFS, git 2.50.1 (Apple Git-155), on the final head `f50379c3c` rebased onto `6fbea8186`:

- `./internal/gitmeta/` — 39.2 s
- `./internal/daemon/walk/` — 13.4 s (the only consumer of `IsPathIncluded`)
- `./internal/extractors/` — 52.6 s (calls `ProbeRepo`)
- `./internal/daemon/` — 32.4 s
- `./internal/daemon/watch/` — 160.7 s, and `TestChangePoller_SparseFilteredPathConverges` re-run alone: `--- PASS (2.92s)`, not a skip
- `./cmd/grafel/` — 157.5 s (calls `ProbeRepo`; also carries the whole-graph digest pin)

`go vet` and `gofmt -l` clean on every touched package.

Nothing here is `runtime.GOOS`-conditional. `ci:full` is applied and re-applied after every push (a push fires `synchronize`, not `labeled`): all measurement is macOS, the macOS leg does not run without the label, and this change is exactly the kind a platform difference in git behaviour would break.

## Behaviour change — smaller than the first draft of this body claimed

The original wording promised cone-mode users a smaller graph. **That is wrong, and the review measured it.** On a real cone-sparse `django` (cone on `django/db`): 7043 tracked, 6897 skip-worktree, 146 files on disk, and `WalkRepo` returns **146 files with sparse and 146 without — delta 0**. git had already removed the out-of-cone files from the working tree, so the walker never saw them; the graph was cone-limited before this PR, by git rather than by grafel.

What actually changes:

1. **`CoverageStatus` flips `full` → `partial`** on a genuinely sparse repo. Cosmetic: written at `cmd/grafel/index.go`, carried through `graph.Document`, surfaced as a dashboard field. No consumer degrades or treats it as a problem.
2. **Untracked files outside the cone are now excluded.** git does not delete what it never tracked, so these are the paths the filter actually removes — and this is exactly the #6940/#6961 case, where such a path was re-reported by the poller forever.
3. `ConeMode` is now honestly documented as having **no non-test reader anywhere in the tree**, so nothing depended on it selecting an algorithm.

The blocker case above is the one place where the delta would have been enormous, and in the wrong direction.

Refs #6940, #6961.

### DIFF
--- a/internal/daemon/watch/poll_convergence_6940_test.go
+++ b/internal/daemon/watch/poll_convergence_6940_test.go
@@ -185,28 +185,22 @@ func TestChangePoller_SparseFilteredPathConverges(t *testing.T) {
 	cpWriteFile(t, repo, "src/keep.go", "package src\n")
 	cpGitRun(t, repo, "add", "-A")
 	cpGitRun(t, repo, "commit", "-q", "-m", "src")
-	// --no-cone deliberately: git's CONE patterns are `/*` + `!/*/` + `/src/`,
-	// and gitmeta.IsPathIncluded treats a bare `*` as match-all, so a cone
-	// checkout makes its own filter include everything. That is a real gap in
-	// the sparse implementation and it is not #6940; a non-cone pattern set is
-	// the shape grafel's filter actually applies, so it is the one that puts
-	// the SPARSE GATE under test here rather than a no-op.
+	// --no-cone deliberately, but no longer for the reason first recorded here.
+	// #6964 fixed both gaps this fixture used to work around, and the two
+	// `config --local` mirror lines that came with them are GONE (#6967): the
+	// probe now reads the scope git writes, so nothing needs mirroring, and
+	// cone patterns are matched properly, so a cone checkout no longer makes
+	// grafel's own filter include everything. --no-cone is kept because a
+	// verbatim pattern is the simplest shape for THIS test's subject, which is
+	// poll convergence, not sparse semantics — those are graded against real
+	// git in internal/gitmeta/sparse_realgit_6964_test.go.
 	if out, err := cpGitTry(repo, "sparse-checkout", "set", "--no-cone", "src"); err != nil {
 		t.Skipf("git sparse-checkout unavailable here: %v\n%s", err, out)
 	}
-	// git >= 2.32 writes core.sparseCheckout into the WORKTREE-scoped config
-	// (.git/config.worktree, via extensions.worktreeConfig), and
-	// gitmeta.ProbeRepo reads `config --local` only — so the probe does not see
-	// a sparse checkout this git just created. That gap is real and is not
-	// #6940; mirroring the flag into --local here keeps the SPARSE GATE, which
-	// is what this test is about, reachable rather than silently untested.
-	cpGitRun(t, repo, "config", "--local", "core.sparseCheckout", "true")
-	cpGitRun(t, repo, "config", "--local", "core.sparseCheckoutCone", "true")
 	if si := gitmeta.ProbeRepo(repo); !si.IsSparse {
-		// NOT a skip (#6961 review): the two config lines above make this
-		// outcome deterministic, and CI runs without -v, so a skip here would
-		// delete the only sparse leg in silence — removing just those two
-		// lines turned the whole test into "SKIP ... PASS, exit 0".
+		// NOT a skip (#6961 review): with the probe reading git's own scope
+		// resolution this outcome is deterministic, and CI runs without -v, so
+		// a skip here would delete the only sparse leg in silence.
 		t.Fatal("the probe does not see the sparse checkout this fixture just created — the sparse gate is UNTESTED, not unreachable")
 	}
 
--- a/internal/gitmeta/gitmeta_fifo_6416_test.go
+++ b/internal/gitmeta/gitmeta_fifo_6416_test.go
@@ -172,11 +172,18 @@ func TestParseSparsePatternFileFIFODoesNotHang(t *testing.T) {
 	p := mkfifoInGitTemp(t, dir, "info", "sparse-checkout")
 
 	var got []string
+	var haveFile bool
 	mustReturnGit(t, "parseSparsePatternFile with a FIFO sparse-checkout", func() {
-		got = parseSparsePatternFile(p)
+		got, haveFile = parseSparsePatternFile(p)
 	})
 	if len(got) != 0 {
 		t.Errorf("parseSparsePatternFile = %v, want none for a refused pattern file", got)
+	}
+	// A refused pattern file must read as NO FILE, not as an empty pattern
+	// set: ProbeRepo turns "no file" into "not sparse" (full index, a
+	// superset) and an empty pattern set into "exclude everything" (#6967).
+	if haveFile {
+		t.Error("parseSparsePatternFile reported the FIFO as a readable pattern file — that would index the repo to zero files")
 	}
 }
 
@@ -205,9 +212,12 @@ func TestGitMetaReadersStillReadRegularFiles(t *testing.T) {
 	}
 
 	sp := writeGitFile(t, dir, "# comment\nsrc/\n\napps/web/\n", "info", "sparse-checkout")
-	pats := parseSparsePatternFile(sp)
+	pats, haveFile := parseSparsePatternFile(sp)
 	if len(pats) != 2 || pats[0] != "src/" || pats[1] != "apps/web/" {
 		t.Errorf("parseSparsePatternFile = %v, want [src/ apps/web/]", pats)
+	}
+	if !haveFile {
+		t.Error("parseSparsePatternFile: a readable pattern file must report haveFile=true")
 	}
 }
 
@@ -221,7 +231,7 @@ func TestGitMetaSkipIsReported(t *testing.T) {
 
 	dir := t.TempDir()
 	p := mkfifoInGitTemp(t, dir, "info", "sparse-checkout")
-	mustReturnGit(t, "parseSparsePatternFile with a FIFO", func() { _ = parseSparsePatternFile(p) })
+	mustReturnGit(t, "parseSparsePatternFile with a FIFO", func() { _, _ = parseSparsePatternFile(p) })
 
 	out := buf.String()
 	if !strings.Contains(out, p) {
@@ -281,7 +291,7 @@ func TestGitMetaSkipNotReportedForAbsentFile(t *testing.T) {
 
 	dir := t.TempDir()
 	_ = readGitdirFile(filepath.Join(dir, ".git"))
-	_ = parseSparsePatternFile(filepath.Join(dir, "info", "sparse-checkout"))
+	_, _ = parseSparsePatternFile(filepath.Join(dir, "info", "sparse-checkout"))
 	_ = contentToken(filepath.Join(dir, "refs", "heads", "main"))
 	if buf.Len() != 0 {
 		t.Errorf("absent files reported %q, want silence", buf.String())
--- a/internal/gitmeta/sparse.go
+++ b/internal/gitmeta/sparse.go
@@ -64,9 +64,11 @@ type SparseInfo struct {
 	// It does NOT select a matching algorithm. Cone mode is a restriction on
 	// what git WRITES — directory prefixes expressed as `/*`, `!/*/`, `/dir/`
 	// — and those lines are ordinary gitignore-syntax patterns, so
-	// IsPathIncluded answers both modes with one matcher (#6964). The flag is
-	// reported because callers surface it, and because it is the only way to
-	// tell a cone repo from a non-cone one after the fact.
+	// IsPathIncluded answers both modes with one matcher (#6964).
+	//
+	// It is reported for callers, and consumed by none: there is no non-test
+	// reader of this field anywhere in the tree today. Said plainly so nobody
+	// re-derives behaviour from it.
 	ConeMode bool
 }
 
@@ -128,7 +130,30 @@ func ProbeRepo(repoPath string) SparseInfo {
 	coneMode := cone == "true"
 
 	// Parse the sparse-checkout pattern file.
-	patterns := parseSparsePatternFile(filepath.Join(gitDir, "info", "sparse-checkout"))
+	//
+	// NO PATTERN FILE MEANS NOT SPARSE, whatever the flag says (#6967 review).
+	// The flag alone is reachable — `git config --global core.sparseCheckout
+	// true` is the pre-2.25 manual workflow and still sits in real dotfiles —
+	// and in that state git populates the FULL tree: measured on git 2.50.1,
+	// `ls-files -v` tags every path 'H' and the working tree is complete.
+	// Reporting IsSparse=true with no patterns would hand the walker a
+	// SparseInfo that excludes every path (that is the deliberate contract for
+	// a constructed value), so the repo would index to ZERO files — silently,
+	// because the sparse layer is the one walk layer that skips without
+	// emitting a SkipEntry. This gate is here rather than in IsPathIncluded
+	// precisely so that contract is left alone.
+	//
+	// A file that EXISTS but yields no patterns is a different state and stays
+	// sparse: git checks out NOTHING for an empty (or comment-only) pattern
+	// file, also measured. So the test is the file, not the pattern count.
+	//
+	// An existing-but-unreadable file lands on "not sparse" too, which is the
+	// over-inclusive direction — a superset, the same direction every other
+	// failure in this file chooses.
+	patterns, haveFile := parseSparsePatternFile(filepath.Join(gitDir, "info", "sparse-checkout"))
+	if !haveFile {
+		return SparseInfo{}
+	}
 
 	return SparseInfo{
 		IsSparse: true,
@@ -138,16 +163,20 @@ func ProbeRepo(repoPath string) SparseInfo {
 }
 
 // parseSparsePatternFile reads the sparse-checkout pattern file and returns
-// non-empty, non-comment lines. Returns nil when the file is absent or
-// unreadable (a missing file is not an error — it just means no patterns).
-func parseSparsePatternFile(path string) []string {
+// its non-empty, non-comment lines plus whether the file was read at all.
+//
+// The second return exists because "no patterns" and "no file" are different
+// states in git and must not be collapsed: an empty pattern file checks out
+// nothing, an absent one checks out everything (#6967 review). Callers that
+// only want the lines can ignore it; ProbeRepo cannot.
+func parseSparsePatternFile(path string) ([]string, bool) {
 	// readGitMetaFile, not os.Open: open(2) is what blocks on a FIFO, so
 	// scanning rather than slurping made no difference to #6416. The pattern
 	// file is name-chosen ("info/sparse-checkout" under the git dir) and is
 	// read before any walk, so no entry-type gate sits in front of it.
 	data, err := readGitMetaFile(path, maxSparsePatternBytes)
 	if err != nil {
-		return nil
+		return nil, false
 	}
 
 	var patterns []string
@@ -159,7 +188,7 @@ func parseSparsePatternFile(path string) []string {
 		}
 		patterns = append(patterns, line)
 	}
-	return patterns
+	return patterns, true
 }
 
 // IsPathIncluded reports whether relPath (a repo-relative, forward-slash path
@@ -202,6 +231,11 @@ func parseSparsePatternFile(path string) []string {
 // ConeMode is informational only. Cone patterns are a subset of the general
 // syntax, so one matcher answers both; the flag is still reported for the
 // coverage badge and for callers that want to know how the repo was made.
+//
+// Known divergence, pre-existing and in the over-inclusive direction: a
+// pattern with a trailing space (`/src/ `) is stripped by
+// parseSparsePatternFile's TrimSpace and read as `/src/`, while git keeps the
+// space and matches nothing. grafel indexes a superset there.
 //
 // Cost: the pattern list is parsed on each call rather than cached, because
 // SparseInfo is a plain value that callers construct and mutate directly. The
@@ -274,10 +308,11 @@ func parseSparsePattern(raw string) (sparsePattern, bool) {
 	if strings.HasPrefix(line, "!") {
 		p.negate = true
 		line = line[1:]
-	} else if strings.HasPrefix(line, `\!`) || strings.HasPrefix(line, `\#`) {
-		// gitignore escape for a literal leading '!' or '#'.
-		line = line[1:]
 	}
+	// gitignore's `\!` / `\#` escapes need no handling here: path.Match reads
+	// a backslash as an escape too, so `\!foo` matches the literal "!foo"
+	// either way. Stripping the backslash would be a branch that cannot change
+	// an answer, and so could never be graded.
 	if strings.HasSuffix(line, "/") {
 		p.dirOnly = true
 	}
--- a/internal/gitmeta/sparse.go
+++ b/internal/gitmeta/sparse.go
@@ -12,7 +12,10 @@
 // Detection strategy (zero git-process overhead on non-sparse repos):
 //  1. Read the git-dir via `git rev-parse --git-dir` (already available via
 //     RunGit — 2-second timeout, returns "" on failure).
-//  2. Check git-config boolean `core.sparseCheckout`. If false/absent → not sparse.
+//  2. Check git-config boolean `core.sparseCheckout` with git's own scope
+//     resolution (worktree → local → global → system). Since git 2.32 the
+//     value lives in the WORKTREE scope, which `--local` cannot see (#6964).
+//     If false/absent → not sparse.
 //  3. Parse <git-dir>/info/sparse-checkout for the active pattern list.
 //
 // SparseInfo is intentionally lightweight — it is computed once per index
@@ -26,6 +29,7 @@ package gitmeta
 import (
 	"bufio"
 	"bytes"
+	"path"
 	"path/filepath"
 	"strings"
 )
@@ -55,9 +59,14 @@ type SparseInfo struct {
 	// IsSparse is false or the pattern file is unreadable.
 	Patterns []string
 
-	// ConeMode is true when core.sparseCheckoutCone is enabled. In cone mode
-	// every pattern is treated as a directory prefix (recursive); in non-cone
-	// mode standard gitignore-style globs apply.
+	// ConeMode is true when core.sparseCheckoutCone is enabled.
+	//
+	// It does NOT select a matching algorithm. Cone mode is a restriction on
+	// what git WRITES — directory prefixes expressed as `/*`, `!/*/`, `/dir/`
+	// — and those lines are ordinary gitignore-syntax patterns, so
+	// IsPathIncluded answers both modes with one matcher (#6964). The flag is
+	// reported because callers surface it, and because it is the only way to
+	// tell a cone repo from a non-cone one after the fact.
 	ConeMode bool
 }
 
@@ -78,14 +87,44 @@ func ProbeRepo(repoPath string) SparseInfo {
 		gitDir = filepath.Join(repoPath, gitDir)
 	}
 
-	// Check core.sparseCheckout config flag.
-	enabled := RunGit(repoPath, "config", "--local", "--get", "core.sparseCheckout")
+	// Check the core.sparseCheckout config flag.
+	//
+	// SCOPE (#6964): this deliberately passes NO scope flag. `git config --get`
+	// resolves worktree → local → global → system, which is exactly how git
+	// itself resolves core.sparseCheckout, so the probe agrees with the working
+	// tree git actually produced:
+	//
+	//   * git >= 2.32 — `git sparse-checkout init/set` turns on
+	//     extensions.worktreeConfig and writes core.sparseCheckout /
+	//     core.sparseCheckoutCone into the WORKTREE scope
+	//     (<git-dir>/config.worktree). `--local` does NOT read that scope, so
+	//     the previous `--local --get` returned empty and every sparse repo
+	//     created by a modern git was probed as "not sparse" — measured on git
+	//     2.50.1 (Apple Git-155), where `--local --get` exits 1 with no output
+	//     while `--get` returns "true".
+	//   * git < 2.32 — the same keys are written to the LOCAL scope
+	//     (<git-dir>/config). `--get` reads local too, so this spelling is
+	//     strictly wider than the one it replaces: nothing that used to be
+	//     detected stops being detected.
+	//   * a value set by hand in --local (or inherited from --global/--system)
+	//     is honoured, because git honours it — a core.sparseCheckout=true in
+	//     the global config really does give a sparse working tree, verified by
+	//     driving `git read-tree -mu HEAD` under GIT_CONFIG_GLOBAL.
+	//
+	// `--worktree --get` was rejected: it is a hard error (exit 1) on any repo
+	// that has not enabled extensions.worktreeConfig, i.e. on every non-sparse
+	// repo and on every repo made sparse by git < 2.32.
+	//
+	// --bool normalises git's boolean spellings, so `core.sparseCheckout = 1`
+	// (or "yes"/"on") is read as true the way git reads it, instead of falling
+	// through the old literal `== "true"` comparison.
+	enabled := RunGit(repoPath, "config", "--bool", "--get", "core.sparseCheckout")
 	if enabled != "true" {
 		return SparseInfo{}
 	}
 
-	// Detect cone mode.
-	cone := RunGit(repoPath, "config", "--local", "--get", "core.sparseCheckoutCone")
+	// Detect cone mode. Same scope reasoning as above.
+	cone := RunGit(repoPath, "config", "--bool", "--get", "core.sparseCheckoutCone")
 	coneMode := cone == "true"
 
 	// Parse the sparse-checkout pattern file.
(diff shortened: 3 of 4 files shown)


cajasmota:

Measured in my own checkout, `/private/tmp/claude-501/-Users-jorgecajas-Projects-archigraph/c0fefed4-0195-4c81-9a94-07a7145710a4/scratchpad/review-6967-zq8v/wt`, detached at `d4ec29ee`, and at `bb6e1a3bd` (that head merged with `origin/main` = `6fbea8186`, which carries #6961). **git 2.50.1 (Apple Git-155), macOS/APFS**, `go vet` clean, `gofmt -l` clean, `-count=1` throughout, every restore by `cp` from a SHA-256-verified backup, final `git status --porcelain` empty.

The matcher rewrite is the best part of this change and I could not break it. The blocker is not in the matcher — it is one newly-reachable input to `ProbeRepo`.

---

### BLOCKER — a global `core.sparseCheckout=true` with no pattern file now indexes **zero files**

Dropping the scope flag is strictly wider, as claimed. That is the problem: it makes the **global/system** scope reachable for the first time, and there is a shape in that scope which git and grafel now disagree about totally.

`git config --global core.sparseCheckout true` is the **pre-2.25 manual sparse workflow** — still all over old blog posts and dotfiles. On a machine carrying that line, in any repo with no `.git/info/sparse-checkout`:

```
git ls-files -v  ->  H cmd/m.go / H cmd/n.go / H root.txt / H src/a.go / H src/b.go
                     (all H — git checks out EVERYTHING; git status clean)
```

and grafel, measured:

```
IsSparse=true  Patterns=[]  Cone=false  Coverage=partial
IsPathIncluded("root.txt") = false   (git checks it out: true)
IsPathIncluded("src/a.go") = false   (git checks it out: true)
```

`ProbeRepo` returns `IsSparse=true, Patterns=nil`; `IsPathIncluded` short-circuits `len(si.Patterns) == 0 → false`; `walker.go:329` drops every file. **The whole repo silently indexes to nothing**, and the drop is the one walk layer that emits no `SkipEntry` and nothing on `PrintSkipped` (`walker.go:323-332`), so there is not even a `[skip]` line to explain it.

Before this PR `--local --get` could not see the global scope, so this was unreachable. It is a new failure mode, in the exclusion direction, and it is total.

**Fix (one place, `ProbeRepo`):** the pattern file being absent is exactly the state in which git populates the full tree, so it should not be reported as sparse. Not in `IsPathIncluded` — `internal/gitmeta/sparse_test.go:114` and `internal/daemon/walk/sparse_walker_test.go:136` deliberately pin the constructed-value contract `IsSparse=true, Patterns=nil → exclude`, and that contract is fine. It is `ProbeRepo` that must not manufacture that value. Please grade it with a fixture that sets the flag via `GIT_CONFIG_GLOBAL` and asserts against `git ls-files -v`, since the current file's "never hand-write a config value" rule (right rule) means no fixture can currently reach the global scope at all.

---

### Mutant table — scored in the **exclusive** direction, per the inverted risk

| # | mutant | verdict |
|---|---|---|
| RX1 | `p.anchored = true` — unanchored (no-slash) patterns can no longer float | **ALIVE — distinguishable and ungraded** |
| RX2 | `**` matches **one**-or-more components instead of zero-or-more (`i := 1`) | **ALIVE — distinguishable and ungraded** |
| RX3 | an excluded directory prunes its subtree (`if !included && isDir { return false }`) | DEAD — 5 tests |
| RX4 | every pattern forced `dirOnly` | DEAD — 5 tests |

Each edit proved landed by an exact occurrence count (marker count 1, residual count of the original expression 0), against a green baseline (`ok internal/gitmeta 37.698s`) taken immediately before.

**RX1** is ALIVE across **all three** consumer packages on the merged tree — `./internal/gitmeta` ok, `./internal/daemon/walk` ok, `./internal/daemon/watch -run TestChangePoller_SparseFilteredPathConverges` ok. Production-reachable: I drove five bare-name pattern shapes through real `git sparse-checkout set --no-cone` and compared to `git ls-files -v`. Under RX1, all five disagree **in the exclusion direction**:

```
payments   -> MISMATCH services/payments/h.go, services/payments/deep/d.go, vendor/payments/x.go  grafel=false git=true
h.go       -> MISMATCH services/payments/h.go, services/orders/h.go                                grafel=false git=true
payments/  -> MISMATCH services/payments/h.go, services/payments/deep/d.go, vendor/payments/x.go   grafel=false git=true
*.go       -> MISMATCH 8 paths                                                                     grafel=false git=true
deep/      -> MISMATCH src/sub/deep/c.go, services/payments/deep/d.go                              grafel=false git=true
```

The reason it is ungraded is a **fixture whose label claims more than its content**: `TestSparse_NonConeUnanchoredMatchesGit` uses `"services/payments"` and `"src/sub"` — both contain an inner slash, so per gitignore both are **anchored**. The test named "Unanchored" never reaches the unanchored branch of `sparsePattern.matches`. Its own doc comment says so out loud. That branch has no fixture anywhere in the repo.

**RX2** is ungraded because `TestSparse_NonConeDoubleStarMatchesGit`'s `/**/deep/` only ever has `deep` at depth ≥ 2 — `**` is never asked to match **zero** components. Add a root-level `deep/` and real git disagrees: `MISMATCH deep/e.go grafel=false git=true`. (Same family as the M10 you already found and fixed — moving the target from one level to three fixed *too few*, not *too many*.)

**I priced both kills rather than asserting they are cheap, and both are trivial.** Applied to your own file, unmutated suite stayed green (`ok 36.003s`), and each mutant then died on exactly one test:

- RX2 — **one line**: add `"deep/e.go"` to `rgFixtureFiles`. `TestSparse_NonConeDoubleStarMatchesGit` already runs `assertMatchesGit` over every tracked path, so it needs nothing else. → `--- FAIL: TestSparse_NonConeDoubleStarMatchesGit`
- RX1 — **five lines**: a case with a genuinely unanchored pattern.
  ```go
  func TestSparse_NonConeTrulyUnanchoredMatchesGit(t *testing.T) {
      dir := rgFixture(t)
      si := probeAfterSparseSet(t, dir, "--no-cone", "payments")
      assertMatchesGit(t, dir, si)
  }
  ```
  → `--- FAIL: TestSparse_NonConeTrulyUnanchoredMatchesGit`

  Please also rename the existing case to say what it measures (an anchored pattern reached via an *inner* slash) — that is a real axis, it is just not the one the name promises.

---

### (a) The no-scope-flag choice — **correct, and every claim in the body reproduces**

Measured on a `t.TempDir()` repo, git 2.50.1:

- fresh repo, `--worktree --get` → **exit 1**. `--worktree` rejection confirmed.
- after `sparse-checkout set src`: `--local` **exit 1 / empty**, `--worktree` = `true`, no-flag = `true`, `extensions.worktreeConfig` = `true`. Defect 1 reproduces exactly.
- **conflicting scopes** (worktree `true` vs local `false`): no-flag resolves `true`, and git's own behaviour agrees — `ls-files -v` still shows the skip-worktree entry. The probe follows git, which is the right winner.
- **`--bool` garbage** (`core.sparseCheckout = banana`): `git config --bool --get` **exits 128**, `RunGit` maps a non-zero exit to `""` (`gitmeta.go:79-82`), so the probe reports **not sparse**. Git itself refuses to run any command in that repo. So a garbage value cannot newly exclude files — the one direction the review brief flagged is safe. `false`/`0` are graded by the existing `BooleanSpelling/false` subtest, whose `rgAnySparseOnDisk` premise check is real. `off`/`no` are ungraded but are the same `--bool` code path as `on`/`yes`, which are graded.

The only thing "strictly wider" costs you is the blocker above — the widening itself is right, the handling of what it newly reaches is not.

### (b) Matcher fidelity — **excellent; 21 of 22 shapes match real git exactly**

I built a differential harness driving real `git sparse-checkout set` and comparing grafel's verdict to `git ls-files -v` for every tracked path, over shapes your fixtures do not have. Every one of the brief's attacks lands on the correct answer:

`payments` (bare name) · `h.go` (bare filename) · `payments/` · `*.go` · `/services/**/h.go` · `/src/**` · `**/deep/` · `/**/deep/` · `/src/?.go` · `/src/[ab]*.go` · `/src/[ab` (unterminated bracket — git also matches nothing, so `path.Match`'s `ErrBadPattern → false` agrees) · `/services/*/h.go` (`*` correctly does not cross `/`) · `deep/` · `/*` `!/*/` `/services/` `!/services/orders/` (negation re-including then re-excluding a subtree) · `/src/` `!/src/a.go` (file-level re-include) · `/*` `!/*/` `/services/payments/deep/` (re-inclusion two levels under an excluded dir) · `/*` `!/*/` `/src/` `!/src/sub/` `/src/sub/deep/` (three-level alternation) · `/*` `!/src/sub/deep/`.

The prefix-walk with an inherited verdict is the right model, and the "excluded directory does not prune" rule is genuinely reproduced rather than argued — RX3 kills 5 tests including `TestSparse_ReInclusionUnderAnExcludedDirectory`, which proves that test is executing past its `t.Skipf` on this git rather than silently skipping.

One divergence, and it is **pre-existing and in the safe direction**: pattern `/src/ ` (trailing space). `parseSparsePatternFile`'s `strings.TrimSpace` (unchanged by this PR) turns it into `/src/`; git does not strip it and checks out nothing. grafel is a superset there. Not a blocker, worth a line in the doc comment since the file now documents its gitignore fidelity precisely.

Also: `\!`/`\#` de-escaping in `parseSparsePattern` is a **no-op** — `path.Match` already treats `\!` as an escaped literal `!`, so stripping the backslash reaches the same verdict either way. Harmless, but it is ungraded and cannot be graded, because it cannot change an answer.

### (c) Vacuous fixtures — **two more, both found and both cheap**

Beyond your own M9/M10: `TestSparse_NonConeUnanchoredMatchesGit` does not exercise unanchored matching (RX1), and `TestSparse_NonConeDoubleStarMatchesGit` does not exercise `**`-matches-zero (RX2). Both are the same family as the two you already caught. Every other case body I read does reach the path its name claims, and `assertMatchesGit`'s both-directions guard is **real, not decorative** — RX3 and RX4 each die on it, and the `wantOut == 0` / `wantIn == 0` fatals are what make an all-included comparison impossible to pass on.

### (d) The #6961 interaction — **it merged; benign, and the write-up is stale in two places**

#6961 landed as `6fbea8186`, one commit ahead of this PR's base, so `poll_convergence_6940_test.go` **is on main now** and the "still open, cannot be updated in this change" framing no longer holds. On the merged tree the test **passes and does not skip**: `--- PASS: TestChangePoller_SparseFilteredPathConverges (2.98s)`, exit 0. No silent skip — the worst outcome does not occur.

But the recorded reasoning about it is wrong on the point that matters. `core.sparseCheckoutCone true` in `--local` is **not** "actively wrong" after this PR — it is **inert**. `sparse-checkout set --no-cone` writes `core.sparseCheckoutCone=false` into the **worktree** scope, and worktree outranks local (I measured that precedence above), so the no-scope-flag probe resolves `false` and `ConeMode` is already correct. It was actively wrong only under the *old* `--local` probe. It should still be deleted, and so should the `--local core.sparseCheckout true` mirror — please do that in this PR now that the file is on main, rather than leaving a follow-up. Note its `--no-cone src` fixture uses a bare `src`, which is the very unanchored shape RX1 leaves ungraded; it does not happen to distinguish RX1, but it is one more reason to land the RX1 fixture.

### Blast radius of `full` → `partial` — smaller than the body claims, and worth restating

I cloned a corpus repo (`django`) and made it cone-sparse on `django/db`:

```
patterns=[/* !/*/ /django/ !/django/*/ /django/db/]   cone=true   coverage=partial
tracked=7043   skip-worktree=6897   files on disk=146
WalkRepo no-sparse=146 files ; with-sparse=146 files ; delta=0
```

**The file-count delta on a genuine sparse checkout is zero.** git already removed the out-of-cone files from the working tree, so the walker never saw them — the graph was *already* limited to the cone before this PR. So "a cone-mode user gets a smaller graph" is not what happens; what actually changes is (1) the `CoverageStatus` badge, and (2) **untracked** files sitting outside the cone are now excluded, which is exactly the #6961 case. I would restate that paragraph — as written it promises users a change they will not observe, while under-selling the untracked-file effect.

On `partial` itself: no consumer degrades or reports it as a problem. It is written at `cmd/grafel/index.go:911`, carried through `graph.Document` / flatbuffers, and surfaced as a dashboard field (`internal/dashboard/v2_group_settings.go:97`). Cosmetic. `ConeMode` has **zero non-test consumers anywhere in the tree** — so (4) in the brief is answered: nothing depended on it selecting an algorithm, because nothing depends on it at all. The doc comment's "The flag is reported because callers surface it" is not true today; "reserved for callers" would be honest.

---

### Summary

Blocking: the global-scope / no-pattern-file total wipeout. Requested with it: the two ungraded exclusive-direction mutants (6 lines of fixture between them, both kills demonstrated), the `TestSparse_NonConeUnanchoredMatchesGit` rename, deleting the two now-mergeable `--local` mirror lines in `poll_convergence_6940_test.go`, and correcting the blast-radius paragraph. The matcher itself I would merge as-is — 21/22 against real git on shapes chosen to break it is the strongest evidence I could produce, and I could not produce a counter-example.

<sub>Isolated worktree; nothing under `.claude/worktrees/` or any other agent's tree touched; scratch harnesses deleted; `git status --porcelain` empty at `d4ec29ee`.</sub>

cajasmota:
## Coordinator mutant on the rebased head (`f50379c3c`)

Own detached worktree, `git status --porcelain` empty, restored by `cp` from a SHA-256-verified backup (`9dd9f400…` both ways), baseline green immediately before (36.9 s) and after (35.5 s).

I went looking for a **silent-exclusion** path, since that is the direction this diff newly opens and the direction the review's blocker came from. Target chosen outside everything you and the reviewer scored (pattern-file gate, pattern-count gate, `anchored` always, `**` arity, excluded-dir pruning, dirOnly).

**MS-1 — shrink `maxSparsePatternBytes` from `8 << 20` to `24`**, forcing the pattern file to truncate. Edit proven landed (`pre=1`, `post_old=0`, `post_new=1`), `go vet` clean.

**Result: DEAD** — five tests fail: `TestGitMetaReadersStillReadRegularFiles`, `TestSparse_NestedConeMatchesGit`, `TestSparse_NonConeInnerSlashIsAnchoredMatchesGit`, `TestSparse_ReInclusionUnderAnExcludedDirectory`, `TestProbeRepo_sparseEnabled`.

Good — the truncation direction is graded, and by five tests rather than one.

## The observation the mutant surfaced, which is not a blocker

`readGitMetaFile` → `safeio.ReadFile` → `io.LimitReader(f, maxBytes)` + `io.ReadAll`. **A file over the cap is truncated silently — no error, no report.**

For the sparse pattern file that lands in the one state your fix is otherwise careful to avoid:

- an **absent** file → `haveFile=false` → not sparse → full index (over-inclusive, safe, and now correct per the blocker fix)
- an **unreadable** file → `haveFile=false` → not sparse (over-inclusive, safe, deliberately chosen and stated in the comment)
- an **oversized** file → `haveFile=true` with a **partial pattern set** → still sparse, fewer patterns → **files silently excluded**

That is the one input class that lands on the dangerous side, and a mid-line cut can also leave a corrupted final pattern that matches something unintended rather than nothing.

**I am not claiming this is reachable.** 8 MiB of sparse patterns is enormous — a non-cone monorepo with per-file patterns is the only shape I can imagine approaching it, and I have not measured one. `reportGitMetaSkip` deliberately stays quiet here too, but that is correct as written: it reports only `ErrNotRegular`/`ErrWouldBlock`, and a truncation is neither.

**The cheap guard, if you want it:** read `maxBytes+1` and treat "hit the cap" as **not sparse** rather than as a partial set — the same over-inclusive direction every other failure in this file already chooses, and consistent with the reasoning you wrote for the unreadable case. A few lines, and it closes the last input that falls the wrong way.

Entirely your call. It is a remote input and the PR is already better than what it replaces; I would take it as a follow-up issue just as happily, and if you would rather leave it, say so and I will file it rather than hold this.

## The rest checks out from here

The blocker fix is keyed on the **file**, not the pattern count, which is the distinction that matters — and your measurement of the counter-case is what proves it: an empty pattern file makes git check out nothing, so it must stay sparse, while an absent one must not. Gating on pattern count would have looked identical and been wrong, which is exactly what your B2 mutant pins.

Two things worth keeping in the body, which I see are there: the django **delta-0** measurement replacing the overstated blast-radius claim, and the note that four vacuous fixtures appeared in one change — two you found, two the reviewer found — with the lesson that driving real git does not prove a fixture reaches the branch its name claims. That second one is the more valuable of the two.

Merging once CI is green.

Issue #6968, opened 2026-09-07
ComputeID concatenates fields with no separators while EntityID NUL-separates them — 12 entity-kind prefix pairs (plus a 13th the vocabulary cannot see), but ZERO collisions measured over 464k entities


cajasmota:
Found while scoring a coordinator mutant on #6965. Demonstrated, not argued.

## The defect

The repo has two entity-id functions and they disagree about where one field ends and the next begins.

```go
// internal/types — EntityRecord.ComputeID: bare concatenation, NO separators
h.Write([]byte(e.OrgID + e.ProjectID + e.SourceFile + e.Kind + e.Name))

// internal/graph — EntityID: every field NUL-separated
h.Write([]byte(repo));       h.Write([]byte{0})
h.Write([]byte(kind));       h.Write([]byte{0})
h.Write([]byte(name));       h.Write([]byte{0})
h.Write([]byte(sourceFile))
```

`EntityID` was written with the boundary problem in mind — the NUL bytes are there precisely so field contents cannot bleed across boundaries. `ComputeID` was not.

## Demonstrated

Two records, identical `SourceFile`, differing only in how `Kind` and `Name` split the same characters:

```
A = {SourceFile:"x.go", Kind:"AB", Name:""}
B = {SourceFile:"x.go", Kind:"A",  Name:"B"}

ComputeID  A=06e77f0326cac892  B=06e77f0326cac892  collide=true
EntityID   A=8cb5cdf8e9141db1  B=96318bcb45369427  collide=false
```

So `ComputeID` treats two entities as one, while every other part of the system treats them as two.

## Why it matters, in order of severity

**1. A map keyed on `ComputeID` with `EntityID` values silently picks a winner.** This is live in the fix for #6960 (PR #6965): `remapTwinOfAnchors` builds `preStampToFinal` keyed by `ComputeID`, valued by `EntityID`. Colliding records write one entry, the survivor decided by iteration order, and a `grafel.twin_of` naming that preimage is repointed to an **arbitrary one of two distinct entities**. That is the #6369 wrong-node hazard, and it is worse than a dangling reference: a dangling `twin_of` is detectably broken, a confidently-wrong one is not. A collision check is being added there as a local mitigation — this issue is the underlying cause.

[...]

Issue #6966, opened 2026-09-07
OWNER DECISION: the custom-extractor default-off gate is now measured (+20.2% wall, heap +40% on large repos) — stale #6105 clause fixed; blocked on heap concentration and #6973's edge losses, not on the old reasons


cajasmota:
Found while verifying a claim in #6965 (for #6960). The claim checked out, and following it produced something larger.

## The verified facts

**No default path runs custom extractors.** The gate at `cmd/grafel/index.go:4141` is:

```go
if (i.customExtractors || inProcCustomExtractors()) && file.TSTree != nil {
```

- `i.customExtractors` is set only by `WithCustomExtractors`, whose **sole non-test caller is `cmd/grafel/quality.go:45`** — `grafel quality` over a golden fixture.
- `inProcCustomExtractors()` reads `GRAFEL_INPROC_CUSTOM_EXTRACTORS`, unset by default.
- The other call site, `internal/daemon/extract/subproc.go:364`, sits inside the `GRAFEL_SUBPROC_EXTRACT` branch — also opt-in.

The code says so plainly at `index.go:530`: *"The default stays OFF … The one caller that opts in is `grafel quality` over a golden fixture."*

So on a default `grafel index`, and in the daemon, **none of the ~340 custom extractors execute.** This is deliberate and documented, not a bug. What follows is about the *justification* and the *consequences*, both of which have drifted.

## Justification 1 is stale: #6105 is CLOSED

The gate comment (`index.go:4131-4133`) gives two reasons:

> its wall/alloc cost (+17.5% wall, +18.2% TotalAlloc) has never been measured at corpus scale, and **#6105 (synthetic `Class:<Name>` endpoints resolving to nothing) is still open**. Flipping the default is a separate decision.

**#6105 is closed.** One of the two stated blockers no longer exists, and the comment still asserts it. Anyone deciding whether to flip this reads that sentence and concludes the blocker stands.

## Justification 2 admits it was never measured

The remaining reason is cost — and the comment itself says the figures have **never been measured at corpus scale**. So the default-off decision currently rests on one stale blocker and one acknowledged non-measurement. That is not an argument for flipping it; it is an argument that **nobody currently knows what flipping it costs**, which is a different and more fixable problem.

The +17.5% / +18.2% figures also deserve a provenance check: on what corpus, at what size, with which extractors registered.

## The consequence the board has been missing

[...]

PR #6970, opened 2026-09-07
test(#6968): the ComputeID collision scan came back zero — over 464k entities, and the prefix-pair surface is 12, not 8

cajasmota:
## The measurement came back zero, and here is precisely what the zero is over

This is the **measurement arm** of #6968. `ComputeID` is **not changed** — the decision to
attempt that depends on this number, and this number does not argue for paying a migration today.

The scan groups every emitted entity by `(SourceFile, ComputeID)` and reports every group holding
**more than one distinct `graph.EntityID`** — i.e. exactly the shape where the bare-concatenation
`ComputeID` folds two records that the NUL-separated `EntityID` keeps apart.

| custom-extractor gate | entities | `(SourceFile, ComputeID)` groups | colliding groups |
|---|---:|---:|---:|
| **OFF** (production default, the user-facing number) | 464,651 | 464,280 | **0** |
| ON (`grafel quality` only, #6966) | 539,291 | 538,921 | **0** |

59 repos: the 58 repos of `archigraph-corpora` plus grafel itself. macOS/APFS, `darwin 25.6.0`.
Every repo produced entities; no repo errored; no repo was skipped.

`OrgID`/`ProjectID` are a constant prefix within one index run, so omitting them from the grouping
key cannot change which records land in the same group — a constant prefix is collision-invariant.

### The zero is a measurement, not a no-op

The detector carried a **positive control**: injecting the two records
`{Kind:"SCOPE.State", Name:"MachineIdle"}` and `{Kind:"SCOPE.StateMachine", Name:"Idle"}` into a
scanned document, in one file, made it report exactly one group with two distinct `EntityID`s
(`f235d7936fd496bc` vs `a5df3da3a732b17b`, one shared `ComputeID` `f17b1cd3d49d3c9d`) while the
same repo reported zero without the injection. The scan ran the production `Index()` in-process
from this branch — not through the installed daemon.

### The reachable surface is twelve pairs, not eight

The issue's enumeration filtered to `SCOPE.*` identifiers in `kinds.go`. Enumerating
`AllEntityKinds()` — the **emit** vocabulary — gives **12** ordered proper-prefix pairs. The four
the `SCOPE.*` filter dropped are real emittable kinds:

```
Test          | TestClass                  <- "Test" named "ClassFoo" vs "TestClass" named "Foo"
Test          | TestConfig
http_endpoint | http_endpoint_call
http_endpoint | http_endpoint_definition
```

`Test|TestClass` is the most ordinary-looking of the twelve — considerably less contrived than the
`SCOPE.Event`/`SCOPE.EventType` case the issue nominated.

### Which pairs actually appeared, and how close they got

Corpus-wide kind counts (gate OFF; the ON numbers differ only in `SCOPE.External`, 9760 vs 9849):

| pair | member A | member B | both members in one repo |
|---|---:|---:|---|
| `SCOPE.External` / `SCOPE.ExternalEndpoint` | 9849 | 68 | **yes** — actix-examples, aspnetcore-docs-samples, awesome-compose, flask, grafel, nextjs, rails-actionpack, requests, sidekiq |
| `Test` / `TestClass` | 1584 | 356 | **yes** — django, flask-realworld, flask, grafel, requests |
| `Test` / `TestConfig` | 1584 | 7 | **yes** — click, flask-realworld, flask, requests |
| `SCOPE.Channel` / `SCOPE.ChannelBinding` | 9 | 0 | no |
| `SCOPE.Event` / `SCOPE.EventBusEvent` | 0 | 7 | no |
| `SCOPE.Event` / `SCOPE.EventFlow` | 0 | 22 | no |
| `SCOPE.Event` / `SCOPE.EventType` | 0 | 4 | no |
| `SCOPE.External` / `SCOPE.ExternalService` | 9849 | 0 | no |
| `SCOPE.Model` / `SCOPE.ModelEvent` | 0 | 5 | no |
| `SCOPE.State` / `SCOPE.StateMachine` | 0 | 2 | no |
| `http_endpoint` / `http_endpoint_call` | 0 | 83 | no |
| `http_endpoint` / `http_endpoint_definition` | 0 | 848 | no |

**Three of twelve pairs have both members alive inside one repo.** So the zero does not rest on
"these kinds never meet" — it rests on the narrower "they never met in the same file with names
that complete the boundary". That is a thinner margin than "latent by construction", and it is why
this PR adds a guard rather than closing the issue.

`SCOPE.Event`, `SCOPE.Model`, `SCOPE.State`, `SCOPE.ExternalService`, `SCOPE.ChannelBinding` and
`http_endpoint` produced **zero** entities across all 59 repos. Per the board's own rule, a zero
count is corpus-relative: it means these corpora do not exercise those producers, not that the
producers cannot fire.

### A thirteenth pair the vocabulary cannot see

The corpus scan observed **79 distinct kinds**, of which **ten are not declared by
`AllEntityKinds()`** — all emitted by `internal/custom/**`:

```
ChannelEvent  File  SCOPE.DI  SCOPE.Interface  SCOPE.Middleware
SCOPE.Observability  SCOPE.Router  SCOPE.Type  Stream  Subscription
```

`AllEntityKinds()`' own doc comment says it exists so tests can "validate that no producer leaks a
free-form kind string". Ten leak. One of them matters here directly: **`SCOPE.Router` forms a
thirteenth prefix pair with the declared `SCOPE.Route`, and both members are emitted on the corpus
today.** The guard in this PR cannot see it, because a free-form string is not enumerable from the
vocabulary. That is recorded as a named blind spot in the test rather than papered over; closing it
means making those producers declare their kinds, which is a separate change.

(`SCOPE.Type` is also on that list — the exact kind the issue body nominated as the motivating
example. It is emitted and undeclared. `SCOPE.TypeAlias` was not observed, so the pair is not live,
but it is outside the guard either way.)

### The consumer that is still undefended

Two call sites key a map on `ComputeID` and value it with `EntityID`:

- `internal/extractors/incremental.go` `remapTwinOfAnchors` — **guarded**. It builds an `ambiguous`
  set, warns, and leaves the `twin_of` dangling rather than picking a winner (#6960/#6965).
- `cmd/grafel/index.go` `stampEntityIDs` (~L4757) — **not guarded**. Same `preStampToFinal` map,
  same two hashes, and no ambiguity check: on a collision the last writer wins by iteration order
  and a `grafel.twin_of` naming that preimage is repointed at an arbitrary one of two distinct
  entities. That is the full-index path, i.e. the common one.

The measured collision count is zero, so nothing is wrong on today's corpus — but the defence that
#6965 added to the incremental path was never mirrored onto the full-index path, and only one of
the two is protected if a colliding pair ever appears. Worth its own issue; deliberately not fixed
here, since this arm must not change identity behaviour.

## What this PR adds

`TestComputeIDKindPrefixPairs6968` in `internal/types`. It enumerates `AllEntityKinds()` for
ordered proper-prefix pairs and pins the result at the twelve above, failing in **both** directions:

- a **new** pair — someone added a kind that widened the collision surface; the failure names the
  completing prefix and tells them to rename or to add the pair with a justification;
- a **stale** pair — the roster outlived the vocabulary and must shrink.

It enumerates the emit vocabulary rather than source-scanning `kinds.go`, so a declared-but-
unemittable constant cannot inflate it and a kind moved behind a conversion cannot vanish from it.

### Verification

- `go test -count=1 ./internal/types/` — **green** (full package suite, not a `-run` subset).
- No production code is touched — the diff is one new `_test.go` file — so no extractor is reached
  and `./cmd/grafel/` is not implicated. (Its digest hashes fields, not ids, so it could not have
  seen an id change in any case.)

### Mutants scored

Green baseline immediately before each; the edit's landing proved by an exact occurrence count;
`kinds.go` restored by `cp` from a `shasum`-verified backup (`0db31847…`, byte-identical before and
after), never by `git checkout --`.

| mutant | occurrences asserted | verdict |
|---|---|---|
| add `EntityKindOperationSynthetic = "SCOPE.OperationSynthetic"` to `AllEntityKinds()` | 2 | **DEAD** — fails naming `SCOPE.Operation\|SCOPE.OperationSynthetic` |
| drop `EntityKindStateMachine` from `AllEntityKinds()` | 1 removed | **DEAD** — fails naming the now-stale `SCOPE.State\|SCOPE.StateMachine` |

The first mutant's initial attempt did not compile; a build failure is not a kill, so it was
rewritten and re-scored against a compiling tree. Both were re-scored against the final version of
the test.

Refs #6968.

### DIFF
--- a/internal/types/computeid_prefix_pairs_6968_test.go
+++ b/internal/types/computeid_prefix_pairs_6968_test.go
@@ -42,12 +42,22 @@ import (
 // than AllEntityKinds(). The corpus scan observed ten kinds that no
 // AllEntityKinds() entry declares — ChannelEvent, File, SCOPE.DI,
 // SCOPE.Interface, SCOPE.Middleware, SCOPE.Observability, SCOPE.Router,
-// SCOPE.Type, Stream, Subscription — all from internal/custom/** extractors.
+// SCOPE.Type, Stream, Subscription — all from internal/custom/** extractors,
+// and therefore all behind the default-OFF custom gate (#6966).
+//
 // One of them, SCOPE.Router, forms a THIRTEENTH prefix pair with the declared
-// SCOPE.Route, and both members are emitted on the corpus today. This test
-// cannot see that pair, because a free-form kind string is not enumerable from
-// the vocabulary; closing the gap means making those producers declare their
-// kinds, which is a separate change. Refs #6968.
+// SCOPE.Route. State its gate precisely, because the two members do not share
+// one: SCOPE.Route is emitted in BOTH gate states (36 entities, in
+// aspnetcore-docs-samples, awesome-compose and grafel), while SCOPE.Router
+// appears ONLY gate-ON (2 entities, in grafel and play-scala-starter; its sole
+// producer is internal/custom/scala/frameworks.go). So the pair has both
+// members only under gate-ON, and on a default index SCOPE.Router is not
+// emitted at all — it is not a live pair for users today.
+//
+// This test cannot see that pair in either gate state, because a free-form
+// kind string is not enumerable from the vocabulary; closing the gap means
+// making those producers declare their kinds, which is a separate change.
+// Refs #6968.
 func TestComputeIDKindPrefixPairs6968(t *testing.T) {
 	want := map[string]bool{
 		// The eight SCOPE.* pairs enumerated on the issue.
@@ -116,9 +126,6 @@ func TestComputeIDKindPrefixPairs6968(t *testing.T) {
 	}
 	for _, p := range removed {
 		t.Errorf("stale ComputeID prefix pair %s in this test's roster: it is no longer produced by AllEntityKinds(). Remove it from `want`.", p)
-	}
-	if len(added) == 0 && len(removed) == 0 && len(got) != len(want) {
-		t.Errorf("roster size %d != enumerated size %d", len(want), len(got))
 	}
 	if t.Failed() {
 		var all []string
--- a/internal/types/computeid_prefix_pairs_6968_test.go
+++ b/internal/types/computeid_prefix_pairs_6968_test.go
@@ -1,0 +1,131 @@
+package types
+
+import (
+	"fmt"
+	"sort"
+	"strings"
+	"testing"
+)
+
+// EntityRecord.ComputeID hashes OrgID+ProjectID+SourceFile+Kind+Name as a BARE
+// concatenation, with no separators, while graph.EntityID NUL-separates every
+// field. Two records that split the same characters differently across the
+// Kind/Name boundary therefore share one ComputeID and hold two distinct
+// EntityIDs — the #6968 disagreement.
+//
+// The Kind/Name boundary can only be crossed when one kind is a PROPER PREFIX
+// of another: kind A named B+X collides with kind A+B named X. That makes the
+// reachable surface a pure property of the kind vocabulary, and this test pins
+// it: the set of ordered prefix pairs over AllEntityKinds() must equal the
+// roster below, exactly.
+//
+// A NEW pair means a newly-added kind widened the surface — the guard fails and
+// whoever added the kind decides whether the pairing is safe. A REMOVED pair
+// means the roster has gone stale and must shrink. Neither direction is a
+// silent pass, and the test is deliberately over AllEntityKinds() (the emit
+// vocabulary) rather than over a source scan of kinds.go, so a constant that is
+// declared but never emittable cannot inflate it and a kind moved behind a
+// conversion cannot vanish from it.
+//
+// Scanning the corpus for a live instance (#6968) found none: 0 colliding
+// groups over 464,651 entities / 464,280 (SourceFile, ComputeID) groups with
+// the custom-extractor gate OFF, and 0 over 539,291 / 538,921 with it ON,
+// across 58 archigraph-corpora repos plus grafel itself (macOS/APFS). That
+// zero is corpus-relative and this test does NOT assert that no collision can
+// occur: a pair on the roster is reachable the moment two entities of the
+// paired kinds land in one file with completing names, and three of the twelve
+// pairs (SCOPE.External/SCOPE.ExternalEndpoint, Test/TestClass, Test/TestConfig)
+// already co-occur inside a single corpus repo. What this test does is convert
+// "we checked once" into "we keep checking".
+//
+// KNOWN BLIND SPOT, deliberately not papered over: the emit surface is WIDER
+// than AllEntityKinds(). The corpus scan observed ten kinds that no
+// AllEntityKinds() entry declares — ChannelEvent, File, SCOPE.DI,
+// SCOPE.Interface, SCOPE.Middleware, SCOPE.Observability, SCOPE.Router,
+// SCOPE.Type, Stream, Subscription — all from internal/custom/** extractors.
+// One of them, SCOPE.Router, forms a THIRTEENTH prefix pair with the declared
+// SCOPE.Route, and both members are emitted on the corpus today. This test
+// cannot see that pair, because a free-form kind string is not enumerable from
+// the vocabulary; closing the gap means making those producers declare their
+// kinds, which is a separate change. Refs #6968.
+func TestComputeIDKindPrefixPairs6968(t *testing.T) {
+	want := map[string]bool{
+		// The eight SCOPE.* pairs enumerated on the issue.
+		"SCOPE.Channel|SCOPE.ChannelBinding":    true,
+		"SCOPE.Event|SCOPE.EventBusEvent":       true,
+		"SCOPE.Event|SCOPE.EventFlow":           true,
+		"SCOPE.Event|SCOPE.EventType":           true,
+		"SCOPE.External|SCOPE.ExternalEndpoint": true,
+		"SCOPE.External|SCOPE.ExternalService":  true,
+		"SCOPE.Model|SCOPE.ModelEvent":          true,
+		"SCOPE.State|SCOPE.StateMachine":        true,
+		// Four more that the issue's SCOPE.*-only enumeration missed: the emit
+		// vocabulary also holds unprefixed and snake_case kinds, and they pair
+		// among themselves. "Test" + name "ClassFoo" against "TestClass" + name
+		// "Foo" is the most ordinary-looking of the twelve.
+		"Test|TestClass":                         true,
+		"Test|TestConfig":                        true,
+		"http_endpoint|http_endpoint_call":       true,
+		"http_endpoint|http_endpoint_definition": true,
+	}
+
+	kinds := AllEntityKinds()
+	if len(kinds) < 2 {
+		t.Fatalf("AllEntityKinds() returned %d kinds — the enumeration below would be vacuous", len(kinds))
+	}
+
+	got := map[string]bool{}
+	for _, a := range kinds {
+		for _, b := range kinds {
+			if a == b {
+				continue
+			}
+			as, bs := string(a), string(b)
+			if as == "" || bs == "" {
+				t.Errorf("empty entity kind in AllEntityKinds(): %q / %q", as, bs)
+				continue
+			}
+			// a is a PROPER prefix of b: an entity of kind a named
+			// strings.TrimPrefix(bs, as)+X has the same Kind+Name
+			// concatenation as one of kind b named X.
+			if len(as) < len(bs) && strings.HasPrefix(bs, as) {
+				got[as+"|"+bs] = true
+			}
+		}
+	}
+
+	var added, removed []string
+	for p := range got {
+		if !want[p] {
+			added = append(added, p)
+		}
+	}
+	for p := range want {
+		if !got[p] {
+			removed = append(removed, p)
+		}
+	}
+	sort.Strings(added)
+	sort.Strings(removed)
+
+	for _, p := range added {
+		parts := strings.SplitN(p, "|", 2)
+		t.Errorf("NEW ComputeID prefix pair %s: an entity of kind %s named %q+X shares a ComputeID with one of kind %s named X in the same file, while their graph.EntityIDs differ (#6968). "+
+			"Either rename the kind so neither is a prefix of the other, or add the pair to the roster in this test with a note on why the collision is acceptable.",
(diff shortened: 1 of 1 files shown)


cajasmota:
## Independent adversarial review — **APPROVE** (two nits, neither blocking)

Reviewed in my own detached worktree at the PR head `2dd7843c7`:
`/private/tmp/claude-501/-Users-jorgecajas-Projects-archigraph/c0fefed4-0195-4c81-9a94-07a7145710a4/scratchpad/review-6970-zq7v/wt`.
Nothing was measured in the repo root or in anyone else's tree. Final `git status --porcelain` empty;
`internal/types/kinds.go` byte-identical to my pre-mutation backup (`0db31847…`, the same digest you
report), restored by `cp` throughout, never by `git checkout --`.

### (a) Does the positive control make the zero real? — **Yes, replicated independently**

I did not take your harness (it is not in the branch — see nit 2). I wrote my own detector against a
**real in-process `Index()`** run from this branch (`WithExportJSON(true)`, `GRAFEL_DAEMON_ROOT`
isolated to a temp dir), grouping by `(SourceFile, ComputeID)` and counting groups holding more than
one distinct `graph.EntityID(repoTag, kind, name, sourceFile)`, then re-running the same detector
over the same entity slice with `{SCOPE.State,"MachineIdle"}` + `{SCOPE.StateMachine,"Idle"}`
injected into one existing file:

| repo | entities | groups | colliding | colliding after injection |
|---|---:|---:|---:|---:|
| click | 2868 | 2835 | **0** | **1** |
| flask | 2689 | 2654 | **0** | **1** |
| requests | 1675 | 1645 | **0** | **1** |
| express | 1151 | 1151 | **0** | **1** |
| play-scala-starter (gate OFF) | 306 | 306 | **0** | **1** |
| play-scala-starter (gate ON) | 349 | 349 | **0** | **1** |

Independently at the hash level: those two records share `ComputeID` and hold two distinct
`EntityID`s (`f139e443f889d14c` for both `ComputeID`s; `fb792581f7fb70df` vs `cb8f7c09b66a33e0`).
So the zero is a zero from a detector that demonstrably fires. Your control stands.

**The entities-vs-groups gap is duplicate records, confirmed, not a grouping bug.** My detector
also counts records that share a group *and* an `EntityID`: click 2868−2835 = 33 and dupSameID = 33;
flask 35/35; requests 30/30 — an exact reconciliation in every repo. They are `SCOPE.External`
twins with an empty `SourceFile` (`ext:glob` vs `ext:glob:glob`). Your ~371 is the same phenomenon.

**The key cannot be coarser than stated:** `ComputeID` already hashes `SourceFile`, so
`(SourceFile, ComputeID)` is exactly as fine as `ComputeID` alone — there is no coarsening to worry
about. And no repo I ran produced zero entities or errored.

### (b) Is twelve the right number? — **Yes; the issue's eight was wrong**

Enumerated from `AllEntityKinds()` myself, in a throwaway `main` against this branch: **94 kinds, no
duplicates, exactly 12 ordered proper-prefix pairs**, character-for-character the roster in `want`.
The four the issue's `SCOPE.*` filter dropped (`Test|TestClass`, `Test|TestConfig`,
`http_endpoint|http_endpoint_call`, `http_endpoint|http_endpoint_definition`) are real. 12 it is.

**Co-occurrence confirmed at the entity level, not just from a table:** one flask index emits
`Test`=392, `TestClass`=7, `TestConfig`=3, `SCOPE.External`=255, `SCOPE.ExternalEndpoint`=1 — five
members of three pairs alive in one repo. The zero really does rest on "never in the same file with
completing names", exactly as the PR says.

### (c) Mutants — the stale half works, and the comparison is an exact set

Green `-run TestComputeIDKindPrefixPairs6968` baseline immediately before each; every edit proved by
an exact occurrence count (the script aborts on a miscount); `go vet` clean on every mutant before
scoring; `-count=1`; restore + `shasum` after each.

| # | mutant | landing | verdict |
|---|---|---|---|
| M1 | `AllEntityKinds()` returns `nil` | 1 | **DEAD** — `t.Fatalf` "returned 0 kinds — the enumeration below would be vacuous" |
| M2 | `AllEntityKinds()` returns 2 kinds that pair with nothing | 1 | **DEAD** — 12 stale-pair errors (a non-empty-but-vacuous vocabulary is also caught) |
| M3 | **rename** `SCOPE.StateMachine` → `SCOPE.StateEngine` (same count, 12→12) | 1 | **DEAD** — fails in **both** halves at once: NEW `SCOPE.State\|SCOPE.StateEngine` + stale `SCOPE.State\|SCOPE.StateMachine`. This is the membership-vs-exact-set question, and it settles it: the comparison is a genuine two-way set difference, so a same-count substitution cannot slip through |
| M4 | drop `EntityKindStateMachine` from the list (stale half alone) | 1 | **DEAD** — stale-pair error, unaccompanied. The stale half is not decorative |
| M5 | add `EntityKindRevSynth = "SCOPE.RouteSynth"` + list entry (new half alone) | 1+1 | **DEAD** — NEW `SCOPE.Route\|SCOPE.RouteSynth`, message names the completing suffix correctly |
| M6a | declare a constant but do **not** list it | 1 | **GREEN, correct by design** — positive confirmation of the PR's claim that a declared-but-unemittable constant cannot inflate the roster |
| M6b | add an **empty-string** kind to the list | 1+1 | **DEAD** — the `as == "" \|\| bs == ""` arm fires; an empty kind cannot silently pair with everything |
| M7 | replace the `len(got) != len(want)` fallback body with `t.Fatalf("REVIEW PROBE")` | 1 | **GREEN ⇒ the branch is unreachable.** `added` and `removed` both empty over two string sets implies equal cardinality, so this arm is structurally dead code. Harmless, but it is not a third safety net — delete it or keep it knowing it can never fire |

Both of your own mutants reproduce. I found no vacuity path: empty, truncated, renamed,
substituted, added, dropped and empty-string vocabularies all go RED.

### (d) Is the custom-extractor blind spot stated honestly? — **Yes, with one precision nit**

Verified, not taken on trust. Indexing `play-scala-starter` **with the gate ON** emits `SCOPE.DI`(18),
`SCOPE.Interface`(1), `SCOPE.Middleware`(1), `SCOPE.Observability`(1), **`SCOPE.Router`(1)**,
`SCOPE.Type`(10) — six of your ten, every one failing `types.IsValidEntityKind`. Same repo **gate
OFF**: zero undeclared kinds. `SCOPE.Router` has exactly one producer,
`internal/custom/scala/frameworks.go:657` (`PlayRouter`), and it is behind `WithCustomExtractors`,
which defaults OFF. The doc comment names the blind spot, lists the kinds, names `SCOPE.Router`,
and says why the guard cannot see it — that is honest, and the guard's stated population matches its
name.

**Nit 1 (doc precision).** "`SCOPE.Router` … and both members are emitted on the corpus today" reads
as the production default. Under the production default `SCOPE.Router` is emitted **zero** times —
it needs the `grafel quality` gate. Worth six words: "…emitted on the corpus today (under the
custom-extractor gate)". Separately, I did **not** observe `SCOPE.Route` in any of the five repos I
indexed, so the *pair's* liveness is a claim I could not confirm in my sample; your corpus-wide
number may well carry it, but the sentence would be safer scoped to the gate.

**Nit 2 (reproducibility).** The scan harness is not in the branch, so the headline zero cannot be
re-run from this PR by anyone — I had to re-implement it to check you. That is defensible for a
throwaway measurement arm, but it is the reason the guard, not the number, is the deliverable, and
it makes the number un-auditable after this review. A caveat while re-implementing, in case it saves
someone an hour: in the exported graph, an entity `id` is **not** always `graph.EntityID(...)` —
`SCOPE.External` records carry address-shaped ids (`ext:glob`). My first-pass detector keyed on the
exported id and reported **33 phantom collisions on click**; keying on `graph.EntityID` gives 0. A
future re-run that keys on the wrong field will fabricate collisions rather than miss them, but it
will still be wrong.

Nothing here blocks. No production code is touched, `./internal/types/` is green
(`go test -count=1`, full package, exit 0), the guard fails in both directions, and it survives every
vacuity attack I could construct. Not merged, nothing pushed to the branch.

cajasmota:
## Coordinator mutant on `2dd7843c7`

Own detached worktree, `git status --porcelain` empty, restored by `cp` from a SHA-256-verified backup (`59d64d6e…` both ways), baseline green immediately before (12.6 s) and after (4.9 s).

I picked the one mutant that matters most for a guard like this: **does it catch the exact mistake it exists to prevent — the one I actually made?**

**MG-1 — reproduce my own enumeration error inside the guard.** I filtered `kinds` to `SCOPE.*`-prefixed entries only, which is precisely how I derived "8 prefix pairs" in the issue.

**Result: DEAD**, and it names the four pairs my filter dropped, individually:

```
stale ComputeID prefix pair Test|TestClass … no longer produced by AllEntityKinds()
stale ComputeID prefix pair Test|TestConfig …
stale ComputeID prefix pair http_endpoint|http_endpoint_call …
stale ComputeID prefix pair http_endpoint|http_endpoint_definition …
```

That is the strongest evidence this guard is worth having. My enumeration was wrong in exactly the way a hand-rolled vocabulary scan goes wrong — a filter whose population is narrower than the name I gave it — and the guard both catches it and tells the reader which four went missing. It is also a real test of the **stale** half specifically, which is the half that usually rots first.

Combined with the reviewer's seven (empty vocabulary, 2-kind vocabulary, rename in both halves, stale-only, new-only, empty-string kind) and your two, I am satisfied the roster is an **exact set** rather than a membership check.

## Two things before I merge, both small

**1. The precision nit the reviewer raised is worth taking, because it is a gate claim.** The doc comment says `SCOPE.Router`'s pair members are "both emitted on the corpus today", which reads as the production default. The reviewer found `SCOPE.Router`'s only producer (`internal/custom/scala/frameworks.go:657`) sits behind the **default-OFF custom gate** (#6966), and saw no `SCOPE.Route` at all in five repos. So the honest phrasing is *"both observed gate-ON; neither on a default index."*

I care about this one more than its size suggests: I relayed the same imprecise claim into #6968 myself and am correcting it there. Every custom-lane statement on this board needs its gate state attached — that is the whole point of #6966.

**2. Delete the unreachable fallback, or say why it stays.** The reviewer proved `len(got) != len(want)` is structurally unreachable by swapping its body for `t.Fatalf` and watching the suite stay green. Harmless dead code, but in a file whose entire job is to fail loudly, an unreachable branch reads as a third safety net that does not exist. Either drop it or comment it as belt-and-braces.

Nothing else. The measurement, the roster, and the blind-spot documentation are all sound, and the scan's positive control is what makes the zero meaningful rather than decorative.

Issue #6105, opened 2026-08-02
Custom extractors emit synthetic Class:<Name> endpoint IDs that never resolve to any entity


cajasmota:
With in-process custom extraction enabled (#5989, default-off flag), dangling relationship endpoints rise from **1481 to 2081 occurrences (+40.5%)**, with roughly **240 new distinct dangling IDs**.

Most are attributable to the supersede defect in #6104 — an edge pointing at a base entity that was destroyed. But a subset is different in kind, and is a separate defect:

```
40 ×  FROM  Class:Widget<N>
40 ×  FROM  Class:Owner<N>
```

These are **synthetic IDs emitted by custom extractors that resolve to no entity that ever existed in either arm** — not gate-off, not gate-on. They are not orphaned references to something that was removed; they reference something that was never created.

That is distinct from #6104's re-keying gap. Fixing supersede to preserve or re-key destroyed entities will not make these resolve, because there is no target to re-key to.

Two plausible shapes, not yet distinguished:
- the extractor emits an edge to a placeholder it expects a later pass to materialise, and no such pass runs on this path
- the extractor constructs the ID with a different scheme than `graph.EntityID(repoTag, kind, name, sourceFile)`, so it can never match the entity it means to reference

The second is worth checking first, given that these extractors have never run against the live graph and so have never had their ID construction validated against the current scheme.

Note on measurement: raw "unresolved endpoint" counts are meaningless as an absolute here — grafel intentionally retains `ext:` and `scope:` placeholders on the base path. Only the **delta** the gate introduces is meaningful, which is what the figures above report.

Found while reviewing #5989. Not caused by that change; surfaced by it.

Refs #5989, #6104

Issue #5989, opened 2026-07-25
custom extractors (all of internal/custom/**, incl. Django) never run: only caller is behind default-off GRAFEL_SUBPROC_EXTRACT


cajasmota:
## Summary

Every extractor under `internal/custom/**` — the whole Pass 2 class, including the entire Django extractor — is reachable from **exactly one** call site, which sits behind an opt-in env flag that nothing sets. In the default configuration these extractors never run, so the entities and relationships they produce are simply absent from the graph.

This was found while reviewing #5988 (a memory optimisation to a Django *resolver* pass) and is unrelated to that change. It is being filed separately because it is a product-behaviour question, not a performance one.

## Evidence

`extractors.RunCustomExtractors` has exactly one non-test caller:

```
internal/daemon/extract/subproc.go:367:   customEnts, _ := extractors.RunCustomExtractors(ctx, file)
```

That path is reached only from the `subprocExtract()` branch at `cmd/grafel/index.go:1227`. The flag is opt-in with no default:

```go
// cmd/grafel/extract.go:21-24
func subprocExtract() bool {
	v := strings.TrimSpace(os.Getenv("GRAFEL_SUBPROC_EXTRACT"))
	return v == "1" || strings.EqualFold(v, "true") || strings.EqualFold(v, "yes")
}
```

A repo-wide grep for `GRAFEL_SUBPROC_EXTRACT` finds only reads and comments — **nothing in the daemon, the scheduler, the launchd plist, or any config sets it.** The in-process branch (`cmd/grafel/index.go:1277-1300`) runs Pass 1 (per-language AST), Pass 2.5 (YAML framework rules) and Pass 3 (cross-language), and never calls `RunCustomExtractors`.

The gate is documented as temporary — `extract.go:18` says *"Gated on GRAFEL_SUBPROC_EXTRACT=1 during the rollout"* — so this looks like a rollout that stalled with the flag still off, rather than a deliberate permanent default.

Note this is a **different** flag from the one `grafel status` reports. `status` prints `indexing: incremental=… subprocess=…` where `subprocess` is `sched.SubprocessIndexEnabled()` (`internal/cli/status.go:121`) — out-of-daemon *indexing*. An operator reading `subprocess=on` would reasonably conclude subprocess extraction is enabled. It is not; they are unrelated switches.

## Confirmed effect

On a real Django repo (confidential corpus, verified locally, no names here):

[...]

Issue #5954, opened 2026-07-23
epic(index): reduce indexing time + RSS + worktree-derived incremental indexing


cajasmota:
## Goal

Reduce **indexing wall time** (~19m51s on the monorepo corpus) and **peak RSS**
(~3.3 GB whole-machine, ~3277 MB in the `index-internal` subprocess), with the
primary driver being **concurrent agent worktrees** — today each worktree indexes
fully and independently, so N agents cost N x 3.3 GB and N x 19 min.

Full roadmap + 4 grounded research passes:
`docs/superpowers/specs/2026-07-24-indexing-optimization-roadmap.md`
Memory-observability spec: `docs/superpowers/specs/2026-07-24-index-memory-observability-design.md`

## Relationship to #5890

#5890 (write-side streaming graph.fb) is the **structural RSS** track and remains the
home for streaming-write / segmented-FB work (#5903/#5904/#5905/#5907/#5914). This epic
is the **broader program**: measurement, indexing-TIME wins, the memory-bounded
incremental foundation, and the worktree-derived indexing that is the actual end goal.
Structural RSS items link into #5890 rather than duplicating it.

## Baseline (monorepo corpus: 427,301 entities / 1,852,622 rels / 31,820 files)

- Wall ~= 1191s: extraction ~268s (22%), **Tier B link passes ~600s (50%, fully serial)**, flows+write+group-algo ~150-250s.
- Peak RSS ~= 3.3 GB; structural floor ~= 0.8-1.0 GB, so **~2.3 GB is slack**.

## Phases

- **Phase 0 - Measure.** Land the memory-observability instrumentation; convert confidence-gated RSS estimates into ranked byte attribution. (Prereq for the profiling-gated wins.)
- **Phase 1 - Certain quick wins (no format change).** Parallelize independent Tier B link passes; share a source-content/token cache across the ~7 sniffer passes that re-read all files; `MarshalIndent`->`Marshal` on large sidecars (time). Drop the Bazel full-slice copy; free `merged`/scope resolver `idx` in `buildDocument` (RSS ~250-450 MB, structurally certain).
- **Phase 2 - mmap-prev-graph + memory-bounded incremental.** Swap `LoadGraphFromDir` -> `MultiReader` on the prev-graph read; convert the 50-file incremental bail to a memory/time budget; per-repo hash-gate the incremental-friendly link passes. Unblocks drifted worktrees + drops single-index RSS.
[...]

PR #6972, opened 2026-09-07
docs(#6966): the custom-extractor gate cited a blocker that closed a month earlier, and quoted its cost with no denominator

cajasmota:
Comment-only. The gate at `classifyAndReadWithProgress` gave two reasons for staying default-OFF; one was false and the other was unfalsifiable as written.

## #6105 closed a month before this comment stopped saying so

The comment asserted #6105 (synthetic `Class:<Name>` endpoints resolving to nothing) was **"still open"**. It closed 2026-08-04, fixed in #6127 — which gave `internal/custom/java` structural refs the language segment the resolver parses. Anyone weighing the flip read that sentence and concluded a blocker stood. It is removed and replaced with the closure and its fixing PR.

## "+17.5% wall, +18.2% TotalAlloc" had no denominator

The figures are real. Their provenance, which the comment never carried, is PR #6106: one arm per process, alternated externally, 10 reps each, over the **#5989 fixtures — not a corpus**. Full denominator now in the comment:

| metric | OFF | ON |
|---|---|---|
| wall | 1.82 s | 2.14 s |
| TotalAlloc | 805.5 MB | 952.3 MB |
| peak HeapAlloc | 162.3 MB | 169.7 MB (+4.6%) |
| entities | 15360 | 18583 |

Every fixture peaked under ~170 MB, so none of it reaches the GB-scale regime epic #5954 targets. The comment now says to quote the figures with that scale or not at all. The `+17.5%` reference in `incremental_custom_extractors_6960_test.go` is qualified the same way.

## A stale line number

`WithCustomExtractors`' doc comment pointed at `index.go:3944-3945` for those figures. The gate has been past line 4100 for some time. Replaced with a searchable anchor instead of a number that will drift again.

## What this does NOT do

**It does not flip the gate.** The default stays OFF. The corpus-scale measurement the comment conceded was never taken is posted on #6966 (58 repos, gate OFF vs ON); the flip itself is the owner's decision and is not made here.

## Verification

`go test -count=1 ./cmd/grafel/ ./internal/extractors/` — both `ok`.

**No test can observe a comment.** That green suite is evidence nothing else was disturbed; it is not evidence the corrected comment is right. That rests on #6105's closed state, commit `a9c25f0` (`fix(#6105): …`, PR #6127), and the #6106 PR body quoted above — all checkable independently of this branch.

Refs #6966

### DIFF
--- a/cmd/grafel/index.go
+++ b/cmd/grafel/index.go
@@ -527,8 +527,10 @@ func WithExportJSON(export bool) IndexOption {
 // tests running in the same binary.
 //
 // The default stays OFF. inProcCustomExtractors in inproc_custom.go gives the
-// qualitative rationale; the cost figures are at the gate itself, index.go:3944-3945
-// (+17.5% wall, +18.2% TotalAlloc, never measured at corpus scale). The one
+// qualitative rationale; the cost figures and their provenance are at the gate
+// itself, in classifyAndReadWithProgress — search for "The gate stays
+// default-OFF" rather than trusting a line number (this reference said
+// index.go:3944-3945 long after the gate had moved past line 4100). The one
 // caller that opts in is `grafel quality` over a golden fixture — see
 // qualityIndexOptions in quality.go.
 func WithCustomExtractors(enabled bool) IndexOption {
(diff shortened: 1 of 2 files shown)


cajasmota:
## Coordinator verification

**This diff is comment-only, so there is no mutant to score and I am not going to pretend otherwise.** Mutating a comment produces a compiling program with identical behaviour by construction — the verdict would be "equivalent" for every possible edit, which is a fact about comments, not evidence about this change. For a comment correction the meaningful check is whether the new text is *true*, so that is what I did, independently:

| claim | verified |
|---|---|
| #6105 is closed | **CLOSED 2026-08-04T13:58:29Z** |
| fixed by PR #6127 / `a9c25f0` | **PR #6127 MERGED**, commit `a9c25f02a`, titled `fix(#6105): emit the missing language slot in java structural refs` |
| the `+17.5% / +18.2%` figures come from PR #6106 over fixtures | **PR #6106 MERGED**, titled `feat(#5989): wire the 340 dead custom extractors into the in-process path, behind…` — the gate-introducing PR, scoped to #5989 |

All three hold. The stale clause was stale, and the denominator is now stated.

**The provenance finding is the valuable part of this arm** and deserves to outlive the diff: the gate's cost justification was **never a corpus number**. `+17.5% / +18.2%` came from the #5989 fixtures, each peaking under ~170 MB, while the comment presented them alongside "never measured at corpus scale" in a way that read as a measured cost awaiting confirmation. It was a fixture rate awaiting a denominator.

That the corpus number then **reproduced** the fixture rate (+20.2% wall, +16.6% alloc) is a pleasant result, but it was not knowable in advance and is not what the comment claimed.

**Agreed on the green suite caveat, which you stated yourself:** no test can observe a comment, so `./cmd/grafel/` and `./internal/extractors/` passing is evidence that nothing *else* moved — not that the wording is right. The evidence for the wording is the table above.

Merging on green. The measurement arm's own output — the heap concentration on large repos, and #6973's edge losses — is what should drive the default-gate decision now, and both are recorded where that decision will be made rather than in this diff.

PR #6974, opened 2026-09-07
fix(#6971): stampEntityIDs picked a twin_of anchor by arrival order — the guarded path was the rare one

### LINKED ISSUE #6971: stampEntityIDs builds the same ComputeID→EntityID map with NO ambiguity check — #6965 guarded the incremental path and left the full-index path (the default) undefended
Found during #6968's collision scan, **verified in source by me before filing**.

## The asymmetry

Two functions build the same map — pre-stamp `ComputeID()` as the key, final `graph.EntityID()` as the value — and only one of them checks for collisions.

| path | function | ambiguity check |
|---|---|---|
| incremental | `internal/extractors/incremental.go` → `remapTwinOfAnchors` | **yes** — added by #6965: detects two records sharing a `ComputeID` preimage with distinct `EntityID`s, skips the remap for that key, warns |
| **full index** | `cmd/grafel/index.go:4742` → `stampEntityIDs` | **none** |

`stampEntityIDs` at `:4756-4765`:

```go
// preStampToFinal maps a record's pre-stamp ComputeID() to the final
// graph.EntityID() this function assigns it, but only when they differ.
preStampToFinal := make(map[string]string)
...
preStampID := r.ComputeID()
finalID := graph.EntityID(i.repoTag, r.Kind, r.Name, r.SourceFile)
if preStampID != finalID {
    preStampToFinal[preStampID] = finalID
}
```

No `prior`, no `seen`, no ambiguity set. Two records colliding under the key write one entry and the survivor is whichever index comes last, so a `grafel.twin_of` naming that preimage is repointed to an **arbitrary one of two distinct entities** — the #6369 wrong-node hazard, and worse than a dangling anchor because a dangle announces itself while a confidently-wrong id is read as valid by `classfold`'s `twinAnchors` and the symbol index's facet-alias rule.

**The guarded path is the one that runs rarely; the unguarded one is the default.** #6965's fix was correctly scoped to the defect it was fixing, but the reasoning it established — "silently picking one is the only option that is not defensible" — applies here with more force.

## Severity: latent today, and measured to be so

#6968's corpus scan found **zero** colliding `(SourceFile, ComputeID)` groups across 59 repos — 464,651 entities gate-OFF, 539,291 gate-ON — with a positive control proving the detector fires. So nothing is currently mis-pointed. This is a missing guard, not a live corruption.

Two reasons it still matters:

- **The collision surface is real and larger than first thought.** 12 entity-kind prefix pairs exist (not the 8 I originally enumerated — I filtered `SCOPE.*` source strings rather than reading `AllEntityKinds()`), three of them already co-occur inside a single repo, and a 13th (`SCOPE.Router`/`SCOPE.Route`) is invisible to any vocabulary guard because 10 emitted kinds are undeclared. The zero rests on "never in the same *file* with completing names", which is a thinner guarantee than it sounds.
- **A zero is corpus-relative.** Several of the paired kinds emitted nothing at all in this corpus, so their pairs were never exercised.

## The fix

Mirror #6965's guard: detect `seen && prior != finalID`, then **skip that key rather than resolving it**, plus a warn carrying both final ids. Skipping degrades to a dangling anchor — inert and detectable — while resolving invents an answer the function has no evidence for. That reasoning is already written out in `remapTwinOfAnchors`; this is the same decision in the same shape.

**Grade it directly**, as #6965 did: a unit test on `stampEntityIDs` with two records whose `Kind`/`Name` split the same characters (`{Kind:"A",Name:"BC"}` and `{Kind:"AB",Name:"C"}` — **both names non-empty**, since an empty name is skipped by a different guard and would test that instead), driving **both arrival orders**, since arrival order is the hazard. Pin the premises with `Fatalf`: `ComputeID` must still collide and `EntityID` must still not.

## Do not fix `ComputeID` here

Adding separators to `ComputeID` is the root fix and it changes every id that function produces — a migration, not an edit. That decision belongs to #6968. This issue is the local guard for the unguarded path.

Refs #6968, #6965, #6369.

cajasmota:
## The asymmetry, closed

Two functions build the same map — pre-stamp `types.EntityRecord.ComputeID()` as the key, final `graph.EntityID()` as the value — and only one of them checked for collisions. The guarded one was the **incremental** path; the unguarded one was `cmd/grafel/index.go`'s `stampEntityIDs`, i.e. the **full index**, the path that runs by default.

`ComputeID` concatenates `OrgID+ProjectID+SourceFile+Kind+Name` with **no separators**; `graph.EntityID` NUL-separates every field. So `{SourceFile:"x.go", Kind:"A", Name:"BC"}` and `{SourceFile:"x.go", Kind:"AB", Name:"C"}` are **one key and two entities**. Without a check, both records wrote the same slot and the survivor was whichever index came last — a `grafel.twin_of` naming that preimage was repointed at an **arbitrary one of two distinct entities**. That is the #6369 wrong-node hazard, and it is strictly worse than a dangling anchor: a dangle names an id no entity has and is inert and detectable, while a confidently-wrong one reads as **valid** to `classfold`'s `twinAnchors` and to the symbol index's facet-alias rule.

## The fix

Mirrors `internal/extractors/incremental.go`'s `remapTwinOfAnchors` (#6965) — same detection (`seen && prior != finalID`), same verdict (**skip the ambiguous key, do not resolve it**), same warn carrying both final ids. The comment says it mirrors that function, so the two stay recognisably one decision. `EntityRecord.ID` stamping is untouched by the guard: both colliders still get their own distinct final id; only the remap map is governed.

**Scope is the guard.** `ComputeID` is **not** changed here. Adding separators is the root fix, it rewrites every id that function produces, and it is a migration — it belongs to #6968.

## Grading

A unit test on `stampEntityIDs` directly:

- **Both names non-empty** (`{Kind:"A",Name:"BC"}` / `{Kind:"AB",Name:"C"}`) — an empty `Name` is skipped by a *different* guard and would have tested that instead.
- **Both arrival orders driven**, because arrival order *is* the hazard; fixing the order proves half of it.
- **Premises pinned with `Fatalf`**: `ComputeID` must still collide, `EntityID` must still not, and the shared pre-stamp id must differ from both final ids. If #6968 ever lands, this test fails **loudly** instead of passing vacuously.
- An **ordinary, non-colliding** facet/anchor pair rides in the same batch as a built-in positive control.

### Mutants scored — all DEAD

macOS/APFS, in the branch worktree, `go test -count=1`, each edit proved landed by an exact occurrence count and restored by `cp` from a shasum-verified backup.

| mutant | result |
|---|---|
| disable the collision check (`false && seen && …`) | **RED** — and the wrong id **differs between the two arrival orders**, which is the hazard itself |
| skip **every** key — the silent no-op that "fixes" the bug by disabling the remap entirely | **RED** on the positive control, and on both #6275 tests |
| resolve to `prior` instead of skipping (the more-permissive direction) | **RED**, again order-dependent |
| move `r.ID = finalID` after the ambiguity branch | **RED** — the ambiguous record would ship an empty ID |

## Review rounds — two further mutants killed, and two results worth recording

**Silent no-op ruled out from both independent directions.** The primary risk here is a guard that looks correct on a corpus where #6968 measured **zero** collisions over 464k entities. There are exactly two ways to build one: *detect-but-don't-act* (delete the `ambiguous[twin]` consult) and *act-but-don't-record* (never write into `ambiguous`, i.e. first-writer-wins). Both are **DEAD** on the new test, **in both arrival orders**, and the failure output prints two *different* wrong ids depending on order — the hazard verbatim. Together with the `skip every key` mutant dying on the in-batch positive control *and* both #6275 tests, the no-op is closed from every side.

**Skip-not-resolve matches `remapTwinOfAnchors` exactly.** Read side by side: same detection (`seen && prior != finalID`), same lazily-made `ambiguous map[string]bool` at cap 1, same `continue` without storing, same warn keys (differing only in the `index:` / `incremental:` prefix), same single consult site with the same "LEAVE IT DANGLING, DELIBERATELY" verdict. The **only** intended delta is the `r.ID = finalID` hoist — necessary because this loop also stamps and the incremental one does not — and mutant 4 grades exactly that.

### `MS-2` — the two paths must stamp the same set of records

Removing the `Name == ""` skip from the **slow path only** survived the whole suite. Under it an empty-name record acquires an ID only when some *unrelated* record in the batch happens to carry a `twin_of` — action at a distance. `TestStampEntityIDs_BothPathsSkipTheSameRecords` now compares the two paths directly; MS-2 is **DEAD** (`fast="" slow="58e1ead07da096b0"`). The invariant is stated next to the fast-path guard, cross-referenced from the slow one.

### `R6` — the docstring asserted what nothing observed

Removing the skip from **both** paths kept the parity assertion satisfied (both stamped `58e1ead07da096b0`), kept the positive control passing, and left the entire package **green**. The lockstep was pinned; the rule the loops are in lockstep *about* was not — the claim that they "agree because both SKIP, not because both are inert" lived only in a comment. One assertion (the empty-name ID must be `""` from both paths) closes it: green unmutated, **R6 DEAD**. The named-record control is kept and relabelled as the control for the *other* direction.

**Recorded, not fixed** (non-blocking): the `Fatalf` premise pins call `recordsHaveTwinOf` on each batch, so they pin the **predicate**, not the **dispatch** — they never observe which loop `stampEntityIDs` entered. Forcing the gate always-fast therefore passes this test in isolation; it dies at suite level on the #6275 tests, which can only pass if the slow path runs. The comment now says so.

**Also recorded**: the one real detection gap — a pair missed when the *first* record's `preStampID` equals its own `finalID` — is unreachable by construction (two different preimages under two different constructions agreeing in the same 64 bits; ~2.5e-14 over the 464k corpus). No test, because there is no input to write. `remapTwinOfAnchors` has the identical gap for the identical reason.

## Verification

`go test -count=1 ./cmd/grafel/` — **green, exit 0**, on every round (146.8s / 139.9s / 161.3s). `cmd/grafel` is the only package touched. Note its digest pin hashes **fields, not ids**, so it cannot see an id change: the green digest is **not** evidence here — the unit test is.

Closes #6971
Refs #6968, #6965, #6369

### DIFF
--- a/cmd/grafel/index.go
+++ b/cmd/grafel/index.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"log/slog"
 	"os"
 	"path/filepath"
 	"runtime"
@@ -4776,6 +4777,28 @@ func (i *Indexer) stampEntityIDs(records []types.EntityRecord) {
 	// preStampToFinal maps a record's pre-stamp ComputeID() to the final
 	// graph.EntityID() this function assigns it, but only when they differ.
 	preStampToFinal := make(map[string]string)
+	// ambiguous holds every pre-stamp key that TWO records with DIFFERENT final
+	// ids both claim. Issue #6971: this is the same decision, in the same
+	// shape, as internal/extractors/incremental.go's remapTwinOfAnchors — the
+	// incremental twin of this loop — and its doc comment carries the full
+	// reasoning. In short: types.EntityRecord.ComputeID concatenates
+	// OrgID+ProjectID+SourceFile+Kind+Name with NO separators while
+	// graph.EntityID NUL-separates every field, so {SourceFile:"x.go",
+	// Kind:"A", Name:"BC"} and {SourceFile:"x.go", Kind:"AB", Name:"C"} are ONE
+	// key and TWO entities. Keying a map on the COARSER id while storing the
+	// FINER one lets both records write the same slot, and the winner is
+	// whichever index comes last — so a twin_of naming that preimage would be
+	// repointed at an ARBITRARY one of two distinct entities (#6369's wrong-node
+	// hazard). Such a key is therefore SKIPPED, not resolved: skipping degrades
+	// to the pre-remap behaviour for that one facet — a dangling twin_of, which
+	// names an id no entity has and is thus detectable and inert — while
+	// resolving invents an answer this function has no evidence for, and a
+	// confidently-wrong anchor reads as VALID to classfold's twinAnchors and to
+	// the symbol index's facet-alias rule. (The underlying ComputeID separator
+	// problem is broader than this function and is filed as #6968; do not "fix"
+	// it here by changing ComputeID, which is an identity many other call sites
+	// already depend on.)
+	var ambiguous map[string]bool
 	for k := range records {
 		r := &records[k]
 		if r.Name == "" {
@@ -4783,10 +4806,26 @@ func (i *Indexer) stampEntityIDs(records []types.EntityRecord) {
 		}
 		preStampID := r.ComputeID()
 		finalID := graph.EntityID(i.repoTag, r.Kind, r.Name, r.SourceFile)
+		r.ID = finalID
+		if prior, seen := preStampToFinal[preStampID]; seen && prior != finalID {
+			if ambiguous == nil {
+				ambiguous = make(map[string]bool, 1)
+			}
+			ambiguous[preStampID] = true
+			slog.Default().Warn("index: two records share one pre-stamp entity id — "+
+				"a grafel.twin_of naming it will be left unremapped rather than pointed at "+
+				"an arbitrary one of them",
+				"pre_stamp_id", preStampID,
+				"source_file", r.SourceFile,
+				"kind", r.Kind,
+				"name", r.Name,
+				"final_id", finalID,
+				"other_final_id", prior)
+			continue
+		}
 		if preStampID != finalID {
 			preStampToFinal[preStampID] = finalID
 		}
-		r.ID = finalID
 	}
 	if len(preStampToFinal) == 0 {
 		return
@@ -4798,6 +4837,10 @@ func (i *Indexer) stampEntityIDs(records []types.EntityRecord) {
 		}
 		twin, ok := r.Properties[types.EntityTwinOfProperty]
 		if !ok || twin == "" {
+			continue
+		}
+		if ambiguous[twin] {
+			// LEAVE IT DANGLING, DELIBERATELY. See the note above.
 			continue
 		}
 		if final, remapped := preStampToFinal[twin]; remapped {
--- a/cmd/grafel/index.go
+++ b/cmd/grafel/index.go
@@ -4766,6 +4766,14 @@ func (i *Indexer) stampEntityIDs(records []types.EntityRecord) {
 	if !recordsHaveTwinOf(records) {
 		for k := range records {
 			r := &records[k]
+			// INVARIANT, PINNED BY TestStampEntityIDs_BothPathsSkipTheSameRecords:
+			// this skip and the identical `r.Name == ""` skip in the slow path
+			// below must stay in lockstep. The two paths differ ONLY in whether
+			// they build the twin_of remap; they must stamp exactly the same SET
+			// of records. Drop this guard from one path alone and a record's
+			// identity starts depending on whether some UNRELATED record in the
+			// same batch happens to carry a grafel.twin_of — action at a
+			// distance, and a heisenbug when it bites. Edit one, edit both.
 			if r.Name == "" {
 				continue
 			}
@@ -4798,9 +4806,23 @@ func (i *Indexer) stampEntityIDs(records []types.EntityRecord) {
 	// problem is broader than this function and is filed as #6968; do not "fix"
 	// it here by changing ComputeID, which is an identity many other call sites
 	// already depend on.)
+	//
+	// ONE DETECTION GAP EXISTS AND IS UNREACHABLE BY CONSTRUCTION, so do not
+	// spend time rediscovering it: a colliding pair goes unnoticed if the FIRST
+	// record's preStampID happens to equal its own finalID, because the
+	// `preStampID != finalID` store below never runs and `seen` stays false for
+	// the second. That requires two DIFFERENT preimages under two DIFFERENT
+	// constructions — sha256(OrgID+ProjectID+SourceFile+Kind+Name) versus
+	// sha256(repo \0 kind \0 name \0 sourceFile), both truncated to 16 hex — to
+	// agree in the same 64 bits: a 2^-64 coincidence, not an input anyone can
+	// supply. There is no test here because there is no input to write.
+	// remapTwinOfAnchors has the identical gap for the identical reason; the two
+	// stay mirrored.
 	var ambiguous map[string]bool
 	for k := range records {
 		r := &records[k]
+		// Lockstep with the fast path's identical skip — see the INVARIANT note
+		// there.
 		if r.Name == "" {
 			continue
 		}
--- a/cmd/grafel/stamp_entity_ids_ambiguous_6971_test.go
+++ b/cmd/grafel/stamp_entity_ids_ambiguous_6971_test.go
@@ -156,6 +156,13 @@ func TestStampEntityIDs_AmbiguousPreStampKeyIsSkippedNotResolved(t *testing.T) {
 // So the assertion is deliberately a COMPARISON between the two paths, not a
 // restatement of either one's rule: the same empty-name record is stamped in
 // two batches identical but for a twin_of carrier, and the results must agree.
+//
+// KNOWN LIMIT OF THE Fatalf PINS BELOW: they call recordsHaveTwinOf on each
+// batch, i.e. they pin the PREDICATE, not the DISPATCH — they do not observe
+// which loop stampEntityIDs actually entered. Reviewer mutant R4 (force the
+// gate always-fast) therefore passes THIS test in isolation; it dies at suite
+// level on the #6275 twin_of tests, which can only pass if the slow path runs.
+// The coverage exists, just not where a reader of this test would look.
 func TestStampEntityIDs_BothPathsSkipTheSameRecords(t *testing.T) {
 	const (
 		repo         = "test_repo"
@@ -205,8 +212,18 @@ func TestStampEntityIDs_BothPathsSkipTheSameRecords(t *testing.T) {
 			fastBatch[0].ID, slowBatch[0].ID)
 	}
 
-	// Positive control: the paths agree because both SKIP, not because both are
-	// inert. A normally-named record must be stamped, identically, by each.
+	// The paths must agree by SKIPPING the record, not by both stamping it.
+	// Reviewer mutant R6 — delete the `Name == ""` skip from BOTH loops — kept
+	// the parity assertion above satisfied (both stamped 58e1ead07da096b0) and
+	// left the ENTIRE ./cmd/grafel/ package green. The lockstep was pinned; the
+	// rule the two loops are in lockstep ABOUT was not. This observes it.
+	if fastBatch[0].ID != "" || slowBatch[0].ID != "" {
+		t.Errorf("the empty-name record was STAMPED by both paths (fast=%q slow=%q); the paths must agree by SKIPPING it, not by both stamping it", fastBatch[0].ID, slowBatch[0].ID)
+	}
+
+	// Positive control for the other direction: agreement must not come from the
+	// loops being globally inert. A normally-named record must be stamped,
+	// identically, by each.
 	wantNamed := graph.EntityID(repo, "SCOPE.Component", namedRecName, file)
 	if fastBatch[1].ID != wantNamed {
 		t.Errorf("fast path did not stamp the named record: got %q want %q", fastBatch[1].ID, wantNamed)
--- a/cmd/grafel/stamp_entity_ids_ambiguous_6971_test.go
+++ b/cmd/grafel/stamp_entity_ids_ambiguous_6971_test.go
@@ -1,0 +1,141 @@
+package main
+
+import (
+	"testing"
+
+	"github.com/cajasmota/grafel/internal/graph"
+	"github.com/cajasmota/grafel/internal/types"
+)
+
+// Issue #6971 — stampEntityIDs builds preStampToFinal keyed by the record's
+// pre-stamp types.EntityRecord.ComputeID() and valued by the final
+// graph.EntityID(). ComputeID concatenates OrgID+ProjectID+SourceFile+Kind+Name
+// with NO separators; graph.EntityID NUL-separates every field. So two records
+// whose Kind/Name split the same characters are ONE key and TWO entities.
+//
+// Before the fix the map had no ambiguity check at all: both records wrote the
+// same slot and the survivor was whichever index came last, so a grafel.twin_of
+// naming that preimage was repointed at an ARBITRARY one of two distinct
+// entities — the #6369 wrong-node hazard, and strictly worse than a dangle
+// because classfold's twinAnchors and the symbol index's facet-alias rule read
+// a confidently-wrong anchor as valid.
+//
+// internal/extractors/incremental.go's remapTwinOfAnchors (#6965) already made
+// this decision for the INCREMENTAL path; #6971 is the same decision mirrored
+// onto the FULL-INDEX path, which is the one that runs by default. An ambiguous
+// key is SKIPPED, not resolved — the twin_of is left dangling deliberately.
+//
+// ARRIVAL ORDER IS THE HAZARD, so both orders are driven. An ordinary,
+// non-colliding facet/anchor pair rides along in the same batch as a positive
+// control: skipping EVERY key would "fix" the collision by disabling the remap
+// entirely, and that no-op must fail here.
+func TestStampEntityIDs_AmbiguousPreStampKeyIsSkippedNotResolved(t *testing.T) {
+	const (
+		repo          = "test_repo"
+		collidingFile = "x.go"
+		ordinaryFile  = "y.go"
+	)
+
+	// Both names NON-EMPTY: stampEntityIDs skips Name=="" via a DIFFERENT
+	// guard, so an empty-name pair would exercise that guard instead of the
+	// collision this test is about.
+	recA := types.EntityRecord{Kind: "A", Name: "BC", SourceFile: collidingFile}
+	recB := types.EntityRecord{Kind: "AB", Name: "C", SourceFile: collidingFile}
+
+	sharedPreStamp := recA.ComputeID()
+	finalA := graph.EntityID(repo, recA.Kind, recA.Name, recA.SourceFile)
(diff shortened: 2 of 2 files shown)


cajasmota:
## Coordinator mutant on `962c7123a`

Own detached worktree, `git status --porcelain` empty, restored by `cp` from a SHA-256-verified backup (`8ae65915…` both ways), baseline green immediately before (148.4 s).

Your four mutants cover the guard's logic well, so I went after the thing the refactor *created* rather than the thing it fixes: **`stampEntityIDs` now has two paths, and nothing pins that they agree about which records they stamp.**

**MS-2 — remove the `r.Name == ""` skip from the SLOW path only**, leaving the fast path's intact. Edit proven landed (two `if r.Name == ""` skips remain, from three). `go vet` clean.

**Result: ALIVE** — `./cmd/grafel/` green, exit 0, 137.8 s.

## Why this one is worth a test rather than a shrug

The two paths are supposed to be indistinguishable except for the remap. Under MS-2 they are not: an empty-name record gets an ID **only when some other record in the same batch happens to carry a `twin_of`**. That is action at a distance — a record's identity depending on an unrelated record's properties — and it is exactly the kind of thing that reads as a heisenbug when it eventually bites.

It is also **not hypothetical as a regression path**: you introduced the fast path in this PR, so the invariant "both paths stamp the same set" is new, unstated, and currently held by nothing. The next person optimising either loop has no signal.

**Cheap to grade**: one record with `Name: ""` plus one with a `twin_of`, asserting the empty-name record's ID is the same whichever path runs. Roughly ten lines, no fixture. Please add it, and state the invariant in a comment next to the fast-path guard so the pairing is visible.

## Your flagged edge case — I checked it, and it is unreachable

You flagged that if the *first* record of a colliding pair has `preStampID == finalID` it is never stored, so `seen` is false and the collision goes undetected. Correct reading of the code, and mirroring `remapTwinOfAnchors` rather than diverging was the right call.

It is also **unreachable in practice**, and I think that is worth stating so it stops being an open worry: `ComputeID` is `sha256(OrgID+ProjectID+SourceFile+Kind+Name)` truncated to 16 hex, `EntityID` is `sha256(repo \0 kind \0 name \0 sourceFile)` truncated to 16 hex. `preStampID == finalID` requires two *different* preimages under two *different* constructions to collide in the same 64 bits. That is a 2⁻⁶⁴ coincidence, not an input anyone can supply.

So: real in the code, not reachable by construction. Worth one sentence in the comment saying so — otherwise the next reader rediscovers it and has to redo this analysis.

## The rest is sound

Mirroring `remapTwinOfAnchors` and saying so in the comment is right, and hoisting `r.ID = finalID` above the branch so an ambiguous record still gets stamped is the detail I would most expect someone to miss — your mutant 4 pins it. Mutant 2 (skip *every* key) is the one that matters most, because it is the silent no-op that would "fix" the bug by disabling the remap; killing it on both #6275 tests plus the positive control is the right standard.

Merging once CI is green, with MS-2 graded.

cajasmota:
## MS-2 killed, and one correction to the framing

Pushed `144026682`. Tree: my own worktree, `git status --porcelain` empty after restore, `grep -c MUTANT` = 0 in the committed file.

### 1. The parity test

`TestStampEntityIDs_BothPathsSkipTheSameRecords` stamps the **same** empty-name record in two batches that differ in exactly one thing — a `twin_of` carrier whose anchor is not in the batch, so it only flips `recordsHaveTwinOf` — and asserts the two results agree. The assertion is a **comparison between the paths**, not a restatement of either one's rule, so it fails whichever path drifts, in either direction.

Two things guard it against going vacuous:

- **Both path selections are pinned with `Fatalf`** on `recordsHaveTwinOf`. If either batch ever stopped taking the path it is meant to take, the test would silently be comparing one path with itself and would pass on a mutant; it now dies instead.
- **A positive control**: a normally-named record present in both batches must be stamped, identically, by each. The paths must agree because both *skip*, not because both are inert — deleting the stamping from both loops would satisfy the parity assertion alone.

**MS-2 scored: DEAD.** Applied to the slow path only (edit proven landed — `if r.Name == ""` count 3 → 2, `go vet` clean), it goes RED with `fast="" slow="58e1ead07da096b0"`. Restored by `cp` from a SHA-256-verified backup (`6f738ce6…` both ways), green again. `go test -count=1 ./cmd/grafel/` green, exit 0, **139.9 s** (macOS/APFS).

The invariant is now stated **next to the fast-path guard**, with the slow-path skip carrying a one-line cross-reference back to it, so whoever edits either one sees the pairing.

### 2. One correction: I did not introduce the fast path

The reasoning holds and the test is worth having, but the premise given for *why* is wrong, and it changes what the test is protecting against. `git show HEAD~1:cmd/grafel/index.go` has the `if !recordsHaveTwinOf(records)` fast path at **line 4765** — it was added by #6275, and its own `PERFORMANCE:` paragraph in the doc comment explains why. My commit `962c7123` did not touch that loop at all; the diff there is entirely inside the slow path.

So this is **not** a new invariant that my refactor created and left unstated — it is a **pre-existing** unpinned invariant, live on `main` since #6275, that your mutant found while looking at my change. Practically that makes it slightly more worth having than the "new invariant" framing suggests, since it has been unheld for longer than one PR. It also means the regression MS-2 models could already have been introduced by anyone editing either loop at any point since #6275, not just by the next person optimising mine.

### 3. Edge case recorded as unreachable

Your 2⁻⁶⁴ argument is right and I have written it into the comment, with the reason spelled out (two different preimages under two different constructions — `ComputeID`'s bare concatenation vs `EntityID`'s NUL-separated fields, both truncated to 16 hex — agreeing in the same 64 bits), an explicit "no test, because there is no input to write", and a note that `remapTwinOfAnchors` has the identical gap for the identical reason so the two stay mirrored. That should stop the next reader redoing the analysis.

Not merging — yours to land.

cajasmota:
**You are right and I was wrong — verified.** I wrote that you introduced the fast path in this PR and used that to justify the test. `git show 02dde28ef:cmd/grafel/index.go` (main, before this branch) has `if !recordsHaveTwinOf(records)` at line 4765, and `git log -S recordsHaveTwinOf` traces it to `fb4c6fd4b fix(#6275)`.

So the invariant is **pre-existing and has been unpinned on main since #6275** — not something this refactor created.

That cuts in your favour twice. The test is worth more than my rationale claimed: MS-2's regression has been reachable by anyone editing either loop for the whole life of that fast path, not just since this branch. And it means I picked the mutant for a reason that was factually wrong while the mutant itself was finding something real — the right conclusion from a wrong premise, which is worth flagging as luck rather than judgement.

Thank you for checking rather than accepting the framing. The correction is now the more accurate half of the PR body.

**MS-2 DEAD** and the grading is better than what I asked for: pinning both path selections with `Fatalf` on `recordsHaveTwinOf` so the test cannot degenerate into comparing one path with itself, plus a positive control establishing the two agree *because both skip*, not because both are inert. That second point is the difference between a comparison and a restatement, and it is exactly where this kind of test usually goes vacuous.

Merging on green.

cajasmota:
## Independent adversarial review — **REQUEST-CHANGES** (one line; everything else holds)

Measured in my own detached worktree at `144026682`:
`/private/tmp/claude-501/-Users-jorgecajas-Projects-archigraph/c0fefed4-0195-4c81-9a94-07a7145710a4/scratchpad/review-6974-zq7v/wt`.
`go vet ./cmd/grafel/` exit 0. Clean-tree full `go test -count=1 ./cmd/grafel/` **exit 0, 149.2s**. Every mutant edit proved landed with an exact occurrence count (`die` unless `n==1`), every restore by `cp` from a shasum-verified backup (never `git checkout --`), final `git status --porcelain` empty. Fast scoring suite = `-run 'TestStampEntityIDs|TestFoldClassHierarchyShadows'` (1.2s), escalated to the full package where a mutant survived it.

### Mutants

| # | mutant | verdict |
|---|---|---|
| A2 | skip **every** key (`seen \|\| !seen`) — the silent no-op | **DEAD** — `_AmbiguousPreStampKeyIsSkippedNotResolved` (positive control) **and** `_RewritesTwinOfToFinalID` (#6275). Author's claim reproduced. |
| A4 | sink `r.ID = finalID` below the ambiguity branch | **DEAD** — assertion 3, ambiguous record ships an empty ID. Author's claim reproduced. |
| R1 | keep detection, delete the **consult** site (`if ambiguous[twin] { continue }`) — guard fires but nothing acts | **DEAD**, and the failure output is the hazard verbatim: `twin_of` lands on `57291cb8959610f5` in one arrival order and `70998089d3d7f64a` in the other. |
| R2 | detect + `continue` but **never record into `ambiguous`** (first-writer-wins) | **DEAD** — same assertion, same order-split. This is the purest "guard silently fails to fire" shape and it is graded. |
| R3 | **widen** to `seen` alone (over-firing direction) | **DEAD** — `_RewritesTwinOfToFinalID`'s fixture happens to contain a duplicate anchor (`SCOPE.Component\|User`, same final id), so over-firing drops a legitimate remap and is caught. Incidental, but real. |
| R4 | break the path gate so the **slow path never runs** (`if true \|\| !recordsHaveTwinOf(...)`) | **DEAD at suite level** (#6275 + the new test). Note below. |
| R5 | coordinator's MS-2: remove the `r.Name == ""` skip from the **slow path only** | **DEAD** — `_BothPathsSkipTheSameRecords` fails with `fast="" slow="58e1ead07da096b0"`. The new test does its job. |
| R6 | remove the `r.Name == ""` skip from **both** paths | **ALIVE under the full `./cmd/grafel/` suite** (exit 0, 143.2s). See below. |

### (a) Can the guard silently fail to fire? No.

R1 and R2 are the two independent ways to build a guard that looks identical on a collision-free corpus — detect-but-don't-act, and act-but-don't-record — and both die on the new test, in both arrival orders. A2 (the "protect by disabling the remap wholesale" shape) dies on the in-batch positive control. The latency of the collision is handled: the fixture *manufactures* the collision rather than waiting for one, and the three `Fatalf` premise pins mean that if #6968 lands and `ComputeID` gains separators, this test dies loudly instead of passing empty.

### (b) Is the parity test a real comparison? Yes — with one gap, which is the change I'm asking for.

It is genuinely a comparison (R5 dies on it), and the `Fatalf` pins on `recordsHaveTwinOf` do stop it degenerating into one path against itself.

But the docstring's claim — *"the paths agree because both **SKIP**, not because both are inert"* — is not what the test observes. The named-record positive control proves the paths are **not globally inert**; nothing asserts the empty-name record was **not stamped**. **R6** is the witness: delete the `Name == ""` skip from *both* loops and the paths still agree (both stamp `58e1ead07da096b0`), the positive control still passes, and the **entire `./cmd/grafel/` package is green**. The lockstep is pinned; the rule the paths are in lockstep *about* is not.

Price of grading it: **one assertion**, which I wrote and scored:

```go
	if fastBatch[0].ID != "" || slowBatch[0].ID != "" {
		t.Errorf("the empty-name record was STAMPED by both paths (fast=%q slow=%q); the paths must agree by SKIPPING it, not by both stamping it", fastBatch[0].ID, slowBatch[0].ID)
	}
```

Green on the unmutated branch (exit 0), and it turns R6 **DEAD** (`fast="58e1ead07da096b0" slow="58e1ead07da096b0"`). `gofmt` clean. That is the whole of my REQUEST-CHANGES — the guard itself needs nothing.

Minor, not blocking: R4 shows the parity test alone **passes** when the path gate is forced always-fast, because its pins call `recordsHaveTwinOf(batch)` rather than observing which loop `stampEntityIDs` actually entered. It is covered transitively by the #6275 tests, so the suite kills R4; worth knowing that the pin is on the predicate, not on the dispatch.

### (c) Skip-not-resolve vs the incremental implementation: identical.

Read side by side against `internal/extractors/incremental.go`'s `remapTwinOfAnchors`: same detection (`seen && prior != finalID`), same `ambiguous map[string]bool` lazily made at cap 1, same `continue` without storing, same warn keys (`pre_stamp_id`/`source_file`/`kind`/`name`/`final_id`/`other_final_id`, differing only in the `index:`/`incremental:` prefix), same single consult site with the same "LEAVE IT DANGLING, DELIBERATELY" verdict. The only structural difference is the `r.ID = finalID` hoist, which is correct and necessary here because this loop also stamps and the incremental one does not — and A4 grades exactly that. The comments cross-reference each other and the reasoning matches the code in both.

### Claim 4 (the 2⁻⁶⁴ gap) — reasoning holds.

`EntityRecord.ComputeID` (`internal/types/entity.go:137`) hashes `OrgID+ProjectID+SourceFile+Kind+Name` as one bare concatenation; `graph.EntityID` (`internal/graph/graph.go:259`) writes `repo \0 kind \0 name \0 sourceFile`. Different preimages under the same digest, both `[:16]` hex = 64 bits, so a first record with `preStampID == finalID` is a 64-bit coincidence *per record* (~2.5e-14 over the 464k corpus), not an input anyone can construct. Agreed: there is no test to write, and writing the gap down beats rediscovering it.

### Scope: clean.

`git diff --stat main...144026682` is `cmd/grafel/index.go` (+67/-1) and the new test file only. `ComputeID` is untouched — correctly deferred to #6968. Agreed the digest pin hashes fields not ids and is not evidence here.

cajasmota:
## R6 killed — taking the reviewer's assertion as written

Pushed `06ee4c56a`. Test-only commit; `cmd/grafel/index.go` is byte-identical to `144026682` (SHA-256 `6f738ce6…` before and after the mutant round).

The assertion is theirs, verbatim, and it does what they said:

```
--- FAIL: TestStampEntityIDs_BothPathsSkipTheSameRecords
    the empty-name record was STAMPED by both paths (fast="58e1ead07da096b0" slow="58e1ead07da096b0");
    the paths must agree by SKIPPING it, not by both stamping it
```

**R6 scored DEAD.** Applied to *both* loops, edit proven landed (`if r.Name == ""` count 3 → 1, each removal confirmed separately, `go vet` clean) → RED. Restored by `cp` from a SHA-256-verified backup → green. `go test -count=1 ./cmd/grafel/` **exit 0, 161.3s** (macOS/APFS).

I kept the named-record control and relabelled it as what it actually is — the control for the *other* direction, that agreement does not come from the loops being globally inert. The two assertions are now a matched pair: one forbids both paths stamping, the other forbids neither path stamping. Neither alone is sufficient, which is the thing I got wrong the first time.

The R4 note is written into the test's docstring: the `Fatalf` pins observe the **predicate** (`recordsHaveTwinOf`), not the **dispatch**, so forcing the gate always-fast passes this test in isolation and dies at suite level on the #6275 tests instead. Recorded rather than fixed, per the review.

PR body updated with both results you asked to keep — the no-op ruled out from both independent directions, and the line-by-line correspondence with `remapTwinOfAnchors` with the `r.ID` hoist as the sole intended delta.

### On the acknowledgement

Worth being precise about what failed, because the shape is slightly narrower than "prose asserting what no test observes" and I think that matters for catching the next one. The control **was** real — it observes something, it dies to a real mutant, and it is not decorative. What was unobserved was the *reason* it was there: I wrote a comment naming the failure mode it defends against (`both are inert`) and it only defended against half of that space. A control that is genuinely load-bearing in one direction reads, to a reader and to me, as though it covers the direction the comment claims.

So the tell is not "is this assertion vacuous" — it wasn't — but **"does the comment name a set of failures, and does the assertion cover all of them?"** `not because both are inert` names two ways to be inert: both stamping, and neither stamping. I asserted one. R6 is exactly the other.

That is also the honest reading of your review comment: singling the control out as "the difference between a comparison and a restatement" was true, and still didn't check the claim attached to it. Same tell, one level up.

Not merging — yours to land.

Issue #6976, opened 2026-09-07
MEASURED: the byName default arm encodes a premise that stopped holding — 'is a declaration' is not knowable there, and the gate-ON regression is 100% ranking, 0% policy


cajasmota:
Traced during #6973's diagnosis. This is the **general** mechanism; `blazor.go`'s missing file gate is one instance of it, filed separately.

## The mechanism

`refs.go`'s `indexByName` builds a repo-wide name → entity index. When two entities share a name, it takes the default arm: `delete(byName, name)` and `ambigName = true`. The name is then unresolvable for everyone.

**Nothing weighs the two claimants.** A reference-shaped custom entity — one minted from a *call site* rather than a declaration — takes the slot from, and then destroys the slot of, the **sole real declaration** of that name. The declaration is the thing every consumer wants; the call-site entity is a mention of it.

Downstream, the testmap stub's global `byName` fallback fails, and anything else relying on that fallback fails with it.

## Measured consequence

On `aspnetcore-mvc` with custom extractors ON, this chain destroyed **6,639 `TESTS` edges** (#6973) while the entity count *rose* 89%. Gating the worst producer recovered 59% of them; **the residual ~2,700 come from other producers via this same mechanism** — `INFERRED_FROM_CODE_IF`, `HTTP_AWAIT_CALL`, and DI passes. So fixing `blazor.go` alone leaves ~41% of the loss in place, which is why this is filed on its own.

## Why the existing locality machinery does not cover it

This is the **#6369 / #6467 locality-tier family, one tier short**. Those issues established that resolution needs to prefer the more local, more authoritative candidate rather than treating all same-named entities as equally valid. The tiers implemented so far do not distinguish **a declaration from a reference to a declaration**, so both land in the same bucket and annihilate each other.

Related but distinct: #6104's facet rule rescues same-name pairs that are two views of one thing in one file. It does not apply here because the claimants live in **different files**, which is precisely the case that goes to the destructive default arm.

## What a fix has to establish

1. **A declaration outranks a reference.** Whatever the tier is called, an entity minted from a call site must not be able to evict a declaration. Establish how "is a declaration" is known at that point — if the record does not carry it, that is the first thing to fix and the rest follows.
2. **Deleting the entry is the harshest available option.** Ambiguity currently makes a name resolvable by *nobody*. Consider whether an ambiguous name should keep a best candidate with a lower-confidence marking. **Do not assume the current behaviour is deliberate** — check whether `delete` was chosen or inherited.
[...]

Issue #6973, opened 2026-09-07
DIAGNOSED (hypothesis falsified): blazor.go's missing .razor gate mints 25k false-positive name duplicates that evict real declarations from byName, collapsing DeriveTestsWalkUp's seeds — not a dangling-edge prune


cajasmota:
Surfaced by the corpus-scale gate measurement on #6966 (58 repos, gate OFF vs `GRAFEL_INPROC_CUSTOM_EXTRACTORS` ON, in-process `Index()`, one arm per process). Not caused by that measurement — exposed by it.

## The loss

With the gate ON, three relationship kinds go **down**, concentrated almost entirely in one repo:

| kind | corpus delta | `aspnetcore-mvc` | rest of corpus |
|---|---|---|---|
| `TESTS` | **-4,828** | 43,621 → 36,982 (**-6,639**) | +1,811 across 21 repos |
| `ENTRY_POINT_OF` | -13 | 73 → 37 (**-36**) | +23 |
| `STEP_IN_PROCESS` | -62 | 221 → 113 (**-108**) | +46 |

`SCOPE.External` is also -88 corpus-wide and `SCOPE.Process` -13.

On `aspnetcore-mvc` the same run **adds 35,406 entities** (39,766 → 75,172, +89.0%). So the gate simultaneously produces the largest entity gain in the corpus and **deletes 15.2% of that repo's `TESTS` edges and half its `ENTRY_POINT_OF`**.

`aspnetcore-realworld` shows the same signature at small scale: entities +78.1%, `ENTRY_POINT_OF` 17 → 12, `STEP_IN_PROCESS` 56 → 37.

## Why this is not obviously "refinement"

#6104 fixed the merge to be non-destructive **for entities** — identity-keyed, "a merge never loses an entity, a merge never narrows a span". Both invariants are entity-side. Nothing in that fix constrains the **edges** attached to a superseded, re-owned or re-subtyped entity.

The shape here is consistent with an endpoint being re-keyed or re-owned by a custom C# extractor while its incoming `TESTS` / `ENTRY_POINT_OF` / `STEP_IN_PROCESS` edges keep pointing at the pre-merge identity and are then dropped as dangling. That is a hypothesis, not a diagnosis — I have not confirmed it, and the alternative (the base path was over-emitting `TESTS` and the custom path is correctly narrower) is not excluded.

Two things make it worth resolving before anyone weighs the gate flip:

[...]

Issue #6343, opened 2026-08-19
MEASURED: routing .cshtml to the Blazor extractor yields incidental noise (0.1%/1.6%, zero endpoint→template edges) — option (i) dead, open on (ii) build one vs (iii) won't-fix


cajasmota:
Found by running the #6338 unsupported-language report over real ASP.NET corpora rather than fixtures.

## The gap

`.razor` is supported. **`.cshtml` is not** — and `.cshtml` is the majority form in real ASP.NET code:

| corpus | `.cshtml` files |
|---|---|
| aspnetcore-mvc | **473** |
| aspnetcore-docs-samples | **639** |

Both are public Microsoft sample repositories, so this is not an unusual codebase shape — it is what mainstream ASP.NET MVC looks like.

## Why it matters

This is the same class of blind spot that prompted #6338 in the first place: a user indexes an ASP.NET MVC application, gets a graph containing the controllers and models but **none of the views**, and nothing anywhere says why. The views are where the endpoint-to-template wiring lives, so what goes missing is precisely the cross-layer topology grafel exists to show.

Razor syntax in `.cshtml` and `.razor` overlaps heavily — both are Razor — so this may be closer to a routing/registration gap than a new extractor. `internal/extractors/razor/` already exists. Worth establishing first whether the existing extractor parses `.cshtml` correctly and simply is not routed to it, or whether the dialects diverge enough to need real work:

- `.cshtml` (MVC/Pages) uses `@model`, `@Html.*`, `@Url.*`, layout/section directives, and `@functions`.
- `.razor` (Blazor components) uses `@code`, `@page`, component tags, parameter binding.

If the existing extractor handles the shared Razor grammar, the fix may be an entry in `internal/classifier/classifier.go`'s extension table plus whatever `.cshtml`-specific directives matter.

## How it was found, which is the transferable part

It was **not** found by reading code. #6338's author had considered `.cshtml`, excluded it from the report's language allow-list as "risky" on reasoning, and only reversed that after running the report over real corpora and seeing 473 and 639 files. The data overruled the reasoning.

That is the second time in this cycle that running against real trees rather than curated fixtures changed a conclusion — #6329 is the other, where a confidently-argued mechanism measured to a zero-edge delta.

## Note

`.cshtml` is now in #6338's report allow-list, so once that lands, a user indexing an ASP.NET MVC repo will at least be **told** their views are unindexed rather than silently missing them. That makes this a visible gap rather than an invisible one — but it is still a gap.

[...]

PR #6977, opened 2026-09-07
fix(#6975): blazor.go ran the whole Blazor rule set over every C# file — 72% of the gate's entities on aspnetcore-mvc were its false positives

### LINKED ISSUE #6975: blazor.go has no .razor file gate and its code-method regex makes the access modifier optional — 72% of gate-ON entities on aspnetcore-mvc are Blazor false positives in a repo with no Blazor
Traced during #6973's diagnosis. **This is the proximate cause of #6973's 6,639 lost `TESTS` edges**, and it is a defect in its own right independent of the gate.

## The defect

`internal/custom/csharp/blazor.go` has **no file gate**. Its sibling `blazor_deep.go:179` has one. And `reBlazorCodeMethod` makes the access modifier optional, so the pattern matches ordinary C# constructor calls.

Net effect: in a repo with **no Blazor at all**, every `new X(` in ordinary C# mints a `SCOPE.Operation/function` named `X`.

**Measured on `aspnetcore-mvc`** (gate ON, in-process `Index()`): **25,469 of the 35,566 added entities — 72% — are Blazor false positives.** That repo contains no Blazor.

## Why it matters beyond the noise

The false positives are *name duplicates*, and that is what does the damage. They collide in the repo-wide `byName` index (`refs.go` `indexByName`), which on collision takes the default arm — `delete(byName)` and `ambigName = true`. #6104's facet rule does not rescue them because they live in different files from the real class.

The testmap stub `scope:operation:<file>#<member>` then fails its `byName` global fallback, so explicit `TESTS` edges stop resolving (resolved targets 8,496 → 5,515, **-35%**), `DeriveTestsWalkUp` loses its seeds, and derived edges collapse 8,639 → 1,992. **6,601 of #6973's 6,639 losses are exactly "seed edge gone".**

**Causal control**: gating `blazor.go` to `.razor` recovers **3,906 of the 6,639 edges (59%)** while still keeping a **+26% entity gain**. The residual ~2,700 is the same mechanism from other producers (`INFERRED_FROM_CODE_IF`, `HTTP_AWAIT_CALL`, DI) — filed separately as the `byName` slot-theft issue.

## Why this is not filed as "just add the gate"

The fix is small — a `.razor` file gate matching `blazor_deep.go:179`, plus tightening the two over-firing regexes — but **it changes graph output for every C# repo**, which is why the diagnosing agent stopped rather than shipping it. That is the right instinct, and it makes this a scoped decision rather than a one-liner:

1. **Which files should `blazor.go` see?** `blazor_deep.go` uses `.razor`. Confirm that is right rather than copying it — Blazor code-behind lives in `.razor.cs`, and a gate that excludes it would trade one silent loss for another. **Measure which extensions actually carry the constructs this file matches** before choosing.
2. **The regexes need tightening on their own merits.** An optional access modifier on a "code method" pattern is what makes `new X(` match; a file gate would hide that without fixing it. Both, or the next dialect re-opens it.
3. **Grade the loss direction, not just the gain.** The current tests cannot see this: the entity count went *up* by 89% while edges were destroyed. Assert the specific `TESTS` edges survive, not a count — #6973's count table showed explicit `TESTS` at **+8 net while 7,221 edges were substituted**.
4. **Note the gate state.** Per #6966, `blazor.go` only runs with custom extractors ON, which is not the default. So this is not currently breaking users' graphs — it is breaking the configuration every custom-lane measurement on this board is taken in, which is its own problem.

## Verification a fix must carry

- Reproduce the 25,469 figure before changing anything, and report it after.
- The 59% edge recovery is the causal control; re-derive it rather than citing it.
- `go test -count=1` on the touched packages plus `./cmd/grafel/`, and the golden ratchet — a new file gate changes fixture output if any fixture has C#.
- Mutants preferring the MORE PERMISSIVE direction: remove the new gate, widen it to all `.cs`, make the regex modifier optional again.

Refs #6973, #6966, #6104.

cajasmota:
`internal/custom/csharp/blazor.go` had **no file gate**, and `reBlazorCodeMethod` made the access modifier optional with no line anchor. Together those meant `var repo = new OrderRepository(_db);` in ordinary C# minted a `SCOPE.Operation/function` named `OrderRepository` — a name duplicate of a real class, in a different file, which evicts the real entry from the repo-wide `byName` index and takes the `TESTS` resolution path down with it (#6973).

> **Revised after review** ([reviewer](https://github.com/cajasmota/grafel/pull/6977#issuecomment-5574400334), [coordinator](https://github.com/cajasmota/grafel/pull/6977#issuecomment-5574428354)). Three mutants were ALIVE and two prose claims were false; a third false claim turned up while re-measuring. All are addressed in `f872add0d` and described below. **Every number in this body was measured by me, not carried over from the review.**

## Measured, before and after

`aspnetcore-mvc` from `archigraph-corpora`, **macOS/APFS**, `GRAFEL_INPROC_CUSTOM_EXTRACTORS=1`, in-process `Index()`, `GOMAXPROCS=3`. **Gate state: custom extractors ON**, which is *not* the default (#6966) — it is the configuration every custom-lane measurement on this board is taken in.

| arm | entities | `TESTS` |
|---|---|---|
| gate OFF | 39,765 | 43,621 |
| gate ON, before this change | 75,174 | 36,982 |
| gate ON, after this change | 50,266 | 40,888 |

- `blazor.go`'s own output on that repo: **25,479 → 0** (21,320 `INFERRED_FROM_BLAZOR_CODE_METHOD` + 4,159 `INFERRED_FROM_BLAZOR_COMPONENT_REF`) — 72.0% of the 35,409 entities the gate added, in a repo containing no Blazor. The issue's 25,469 figure reproduced within 10.
- The gate still keeps a **+26.4%** entity gain over the OFF arm.
- **3,906 of the 6,639 lost `TESTS` edges recovered — 58.8%.** That is the issue's causal control, **re-derived here**, not cited.
- The residual ~2,700 is the general declaration-vs-reference eviction in `refs.go` `indexByName`. That is **#6976** and is deliberately **not touched**.

`aspnetcore-mvc` has zero `.razor.cs` files and its 7 `.razor` files never reach this extractor, so these figures are invariant under the regex work below.

## The gate was measured, not copied

#6975 says explicitly not to copy `blazor_deep.go:179`'s `.razor` gate, because code-behind lives in `.razor.cs`. Two measurements decided it:

1. **Which extensions carry the constructs.** Over the whole `archigraph-corpora` tree — 3,726 `.cs`, 105 `.razor`, **0** `.razor.cs`, 1,129 `.cshtml` — the five `@`-directive rules score **zero** in `.cs`. They fire in `.razor` (25/119/10/4/0 = **158 sites**) and, smaller, in `.cshtml` (9/39/0/0/1 = **49 sites**). Only the code-method and component-tag rules fire in `.cs` — 39,009 and 12,994 sites, all false positives.
2. **What can actually reach this function.** `.razor` classifies as language `"razor"` (`classifier.go:345`), so the `file.Language != "csharp"` check rejects it outright. Independently, there is no tree-sitter grammar for razor, so a `.razor` file also has `TSTree == nil`. Either alone is sufficient: **the only extension that both carries Blazor and reaches this code today is `.razor.cs`.**

The gate admits `.razor || .razor.cs`, and the **markup** rules are further restricted to `.razor` — in code-behind `<Foo>` is a generic type argument, not a component reference.

> Finding, out of scope: `blazor_deep.go:179`'s `.razor` arm is dead for the same reason. Worth its own issue.

## The regex, fixed on its own merits

**Three tightenings, each independently graded.**

1. **Anchor** `^[ \t]*` — excludes a **mid-expression** `new X(`.
2. **Type-token keyword rejection** (`blazorNonTypeKeywords`) — the anchor does **not** exclude a **line-leading** `new X(`, and `.razor.cs` is ordinary C#, so that shape survived *inside* the gate. Measured over the corpus C#: **2,503** line-leading `new X(`, **1,666** `return F(`, **293** `else if (`. The pattern now captures the type token as group 1 and `Extract` rejects statement keywords. Go's RE2 has no lookahead, so this cannot live in the pattern — **that map is the mechanism**, which is why all 40 of its entries are now graded individually (see below). Only the type slot is filtered — a companion name-slot filter would only ever fire together with this one and grade neither.

> **Population, reconciled.** My `return F(` count was **1,661** and the reviewer's **1,666**; both are right over different walks. The whole `archigraph-corpora` tree has 3,726 `.cs` files and **1,666** sites; restricting to the 7 C#-bearing repos by name gives 3,721 files and 1,661. `new X(` (2,503) and `else if (` (293) are identical either way. **Every figure in this PR and in the code comments is now the whole-tree one, and says so** — two numbers for one thing in a review record is worse than either.
3. **`\n` in the generic class** `(?:<[^;\n]*>)?` — without it the class is greedy across lines and invents a method out of two unrelated constructs.

Effect on `.cs`: **39,009 → 23,490 (anchor) → 18,853 (anchor + keyword rejection)**. On `.razor` the rejection removes 13 of 110 matches, and **enumerating** them rather than counting shows all 13 are non-declarations — 9 `else if (` and 4 `await <Method>(` call sites, the collider shape again. No declaration lost.

## What review changed

| item | was | now |
|---|---|---|
| **R5** — line-leading `new X(` survived the anchor **inside the gate** | prose claimed the anchor covered it | `blazorNonTypeKeywords`; fixture carries both forms. The reviewer's diagnosis reproduced exactly: moving the old fixture's constructor call onto its own continuation line made the assertion pass with the defect intact |
| **R3** — `Language != "csharp"` guard | **ungraded** (deleting it passed this package *and* all of `cmd/grafel`) | `TestBlazorLanguageGuardRejectsNonCSharp6975` over `razor`/`vbnet`/`java` with a `csharp` control on identical content |
| **MB-1** — `\n` in the generic class | **ungraded** (`[^;]*` left both packages green) | `TestBlazorCodeMethodGenericDoesNotSpanLines6975` — mints `Sum`, declared nowhere, under the mutant; `Render` is the positive control |
| **Prose** — "directive rules fire only in `.razor`" | **false** over its own population (`.cshtml` has 49 sites) | both populations stated; `.cshtml` is unsupported (#6343) |
| **Prose** — `.razor` kept for "a future razor grammar" | **inert**, as R3 proves — the language check would reject it anyway | says the suffix pair names the Blazor file family, as `blazor_deep.go:179` has it |
| **Prose** — ratchet coverage | implied it covered this change | states it **cannot** — see below |

**A third false claim, found while re-measuring.** The comment said a required-modifier variant "drops exactly the bare `@code` methods", citing 110 vs 97. With the keyword rejection in place the two variants select the **identical 97 sites** over the corpus's 105 `.razor` files — 28 distinct names, **zero** found by either alone in either direction. Every real method there carries a modifier. The anchor is now justified on **mechanism** (it rejects a match for its *position*, which is the defect) rather than on a margin that does not exist, and the modifier-less case is stated as pinned **synthetically**, not by the corpus.

**A process note, because it nearly produced a fabricated result.** `(?:<[^;\n]*>)?` occurs **twice** — line 71 (doc comment) and line 90 (the regex). Editing the comment is a no-op that reads as ALIVE. The MB-1 mutant script asserts the fragment is on exactly `[71, 90]`, that line 90 is the regex, that line 90 changed, and that **line 71 still carries `[^;\n]`**.

## Grading the LOSS direction

Every assertion names a **specific entity or edge**. A cardinality assertion is blind here by construction: entities rose 89% while edges were destroyed, and #6973 measured explicit `TESTS` at +8 net while 7,221 edges were substituted.

- `cmd/grafel/blazor_gate_tests_edges_6975_test.go` indexes with the gate **OFF** first as a control, then requires **every** `TESTS` edge from that arm to survive the gate being **ON**, reporting missing edges by name.
- `internal/custom/csharp/blazor_file_gate_6975_test.go` names individual entities in both directions of the gate, plus a `.razor` vacuity guard.

## Mutants — 8/8 DEAD

Re-scored against the revised code. Each edit proven to land by an **occurrence count** and a **changed sha**; green baseline immediately before each; restored by `cp` from a shasum-verified backup.

| | mutant | verdict |
|---|---|---|
| m1 | delete the gate | **DEAD** |
| m2 | widen the gate to any `.cs` | **DEAD** |
| m3 | narrow the gate to `.razor` only | **DEAD** (3 tests) |
| m4 | restore the pre-#6975 unanchored regex | **DEAD** |
| m5 | run markup rules over code-behind | **DEAD** |
| m6 | drop the `blazorNonTypeKeywords` rejection (R5) | **DEAD** — names all four line-leading shapes |
| m7 | delete the `Language != "csharp"` guard (R3) | **DEAD** |
| m8 | widen the generic class to `[^;]*` (MB-1) | **DEAD** |
| m9 | delete any ONE of the 40 `blazorNonTypeKeywords` entries (MB-2) | **DEAD, 40/40** |
| m10 | add a key to the map that the test's literal does not list | **DEAD** |

## MB-2 — the keyword map is graded as an exact set

MB-2 deleted one entry (`"throw"`) and left the other 39; the package stayed green. `throw`'s own reachability is thin and **no live defect is claimed**: `throw new ArgumentException(x)` is already suppressed by the `"new"` key (the pattern aligns on `new` as the type token and `ArgumentException` as the name), and what `"throw"` uniquely guards is `throw Foo(...)`, legal only when `Foo()` returns an exception.

The finding is about the **map** — a **40-entry** membership list (the review said 39; counted, it is 40) graded at some members with nothing saying which. Since RE2 has no lookahead, that map *is* the mechanism. `TestBlazorNonTypeKeywordsIsGradedAsAnExactSet6975` now asserts per key that it is **reachable** (the pattern aligns on it as the type token of a line-leading `<kw> Target(x);`) and **suppressed** (`Extract` mints no `Target`), plus **set equality against an independent literal list in both directions**. It is an internal (`package csharp`) test because the map is unexported and the point is to iterate the real thing.

**Two of my own assertions were wrong first, and scoring is what said so.**

1. The first revision built its subtests by **ranging over `blazorNonTypeKeywords`**. Re-scoring MB-2 against it: **still ALIVE**. Deleting a key removes its subtest and nothing else notices — a test driven by the thing under test cannot detect a deletion. The expectation is now a hand-maintained literal, which is the whole point of "exact set".
2. The reachability check used `FindStringSubmatch`, which returns the **first** match in the source — the fixture's own `void Anchor()` line — so it failed every key for the wrong reason.

**40/40 single-key deletions DEAD**, plus the addition direction. 39 were scored by loop and **the 40th separately**: `while read` drops a final line with no trailing newline, and reporting 39/40 as "all" would have been the same class of error as the two above. One aborted edit worth recording: the first loop pass fed the entire keyword list to the mutant script as a single token, because **zsh does not word-split unquoted parameters** — the occurrence assertion rejected it rather than letting a green run read as ALIVE.

## How the prose errors were found

All three false claims in this PR's prose — "directive rules fire only in `.razor`", the ratchet line, and the 110-vs-97 margin — were found by **re-measuring, never by re-reading**. The 110-vs-97 figure had been relayed by the coordinator *and* independently verified by the reviewer, and it was still wrong once the keyword rejection changed what the variants select. Re-reading a sentence confirms it is well-formed; only re-running the measurement behind it can falsify it.

## Verification

- `go test -count=1` green on `internal/custom/csharp`, `cmd/grafel`, `internal/resolve`, `internal/extractors`.
- **Golden ratchet**: 38 gated fixtures held baseline (29 at 100% must-have recall), no baseline moved. **What that does and does not show:** the golden set has 8 `.cs` files (which the gate now excludes) and 4 `.razor` files (which never reached this extractor), and **zero `.razor.cs`**. So the ratchet confirms only that removing the noise moved no baseline — it **cannot see the extension this gate newly admits**, and no fixture exercises it. `grafel quality` does run with custom extractors ON (`quality.go:45`), so the `.cs` half is real coverage.
- All numbers macOS/APFS, measured in `.../scratchpad/impl-6975-r7k/wt`.

Closes #6975
Refs #6973, #6966, #6976, #6104

### DIFF
--- a/cmd/grafel/blazor_gate_tests_edges_6975_test.go
+++ b/cmd/grafel/blazor_gate_tests_edges_6975_test.go
@@ -1,0 +1,174 @@
+package main
+
+import (
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+
+	"github.com/cajasmota/grafel/internal/graph"
+)
+
+// #6975 / #6973 — the EDGE side of the un-gated Blazor extractor.
+//
+// WHY A COUNT ASSERTION CANNOT SEE THIS DEFECT. On the `aspnetcore-mvc` corpus
+// repo, turning the custom-extractor gate ON (GRAFEL_INPROC_CUSTOM_EXTRACTORS=1,
+// in-process Index(), macOS/APFS) moved entities 39,765 -> 75,174 (+89.0%)
+// while TESTS edges moved 43,621 -> 36,982 (-6,639). #6973 measured explicit
+// TESTS at +8 net while 7,221 edges were substituted underneath. Any assertion
+// of the form "count went up" or "count did not change" is satisfied by that
+// by construction. This test therefore names ONE SPECIFIC EDGE and requires it
+// to be present with the gate ON.
+//
+// THE MECHANISM. `internal/custom/csharp/blazor.go` had no file gate, so
+// `var repo = new OrderRepository(_db);` in ordinary C# minted a
+// SCOPE.Operation/function literally named `OrderRepository`, in a DIFFERENT
+// file from the real class of that name. The two collide in the repo-wide
+// byName index (internal/resolve/refs.go indexByName), whose collision arm
+// deletes the entry and sets ambigName — so the surviving name resolves to
+// nothing and the testmap stub `scope:operation:<file>#<member>` loses its
+// byName global fallback. The seed TESTS edge disappears and DeriveTestsWalkUp
+// loses its starting point.
+//
+// SCOPE NOTE. The general declaration-vs-reference eviction in indexByName is
+// #6976 and is deliberately NOT addressed here; it accounts for the residual
+// edges the file gate does not recover. This test pins only what the gate is
+// responsible for.
+func TestBlazorGateKeepsTestsEdges6975(t *testing.T) {
+	repo := t.TempDir()
+	write := func(rel, body string) {
+		t.Helper()
+		p := filepath.Join(repo, rel)
+		if err := os.MkdirAll(filepath.Dir(p), 0o755); err != nil {
+			t.Fatal(err)
+		}
+		if err := os.WriteFile(p, []byte(body), 0o644); err != nil {
+			t.Fatal(err)
+		}
+	}
+
+	// The production type. Its NAME is the one the false positive duplicated.
+	write("src/OrderRepository.cs", `
+namespace Shop.Data;
+
+public class OrderRepository
+{
+    public int Count()
+    {
+        return 0;
+    }
+}
+`)
+	// An ordinary consumer that constructs it. `new OrderRepository(` here is
+	// the exact site that used to mint the colliding SCOPE.Operation.
+	write("src/OrderService.cs", `
+namespace Shop.Services;
+
+public class OrderService
+{
+    public int Total()
+    {
+        var repo = new OrderRepository();
+        var mapper = new OrderMapper();
+        return repo.Count();
+    }
+}
+`)
+	// The test that must keep its TESTS edge.
+	write("tests/OrderRepositoryTests.cs", `
+namespace Shop.Tests;
+
+public class OrderRepositoryTests
+{
+    [Fact]
+    public void CountReturnsZero()
+    {
+        var repo = new OrderRepository();
+        Assert.Equal(0, repo.Count());
+    }
+}
+`)
+
+	testsEdgesFor := func(arm string) map[string]bool {
+		t.Helper()
+		state := t.TempDir()
+		t.Setenv("GRAFEL_DAEMON_ROOT", t.TempDir())
+		switch arm {
+		case "off":
+			t.Setenv("GRAFEL_INPROC_CUSTOM_EXTRACTORS", "")
+		case "on":
+			t.Setenv("GRAFEL_INPROC_CUSTOM_EXTRACTORS", "1")
+		}
+		if err := Index(repo, filepath.Join(state, "graph.json"), "blazor_gate_6975", nil, false, false); err != nil {
+			t.Fatalf("Index(arm=%s): %v", arm, err)
+		}
+		doc, err := graph.LoadGraphFromDir(state)
+		if err != nil {
+			t.Fatalf("LoadGraphFromDir(arm=%s): %v", arm, err)
+		}
+		byID := map[string]*graph.Entity{}
+		for i := range doc.Entities {
+			byID[doc.Entities[i].ID] = &doc.Entities[i]
+		}
+		out := map[string]bool{}
+		for i := range doc.Relationships {
+			r := &doc.Relationships[i]
+			if r.Kind != "TESTS" {
+				continue
+			}
+			from, ok1 := byID[r.FromID]
+			to, ok2 := byID[r.ToID]
+			if !ok1 || !ok2 {
+				continue
+			}
+			out[from.Name+" -> "+to.Name] = true
+		}
+		return out
+	}
+
+	// The gate-OFF arm establishes that the edge exists at all, so a
+	// gate-ON failure below is attributable to the custom extractors and not
+	// to the fixture never having produced the edge. Without this control the
+	// assertion could pass vacuously in the wrong direction.
+	off := testsEdgesFor("off")
+	if len(off) == 0 {
+		t.Fatal("control arm (gate OFF) produced no TESTS edges at all; the fixture cannot grade anything")
+	}
+	on := testsEdgesFor("on")
+
+	// Every TESTS edge present with the gate OFF must survive the gate being
+	// turned ON. Named individually — the failure message lists the missing
+	// edges, not a delta.
+	var lost []string
+	for e := range off {
+		if !on[e] {
+			lost = append(lost, e)
+		}
+	}
+	if len(lost) > 0 {
+		t.Errorf("custom extractors destroyed %d TESTS edge(s) that exist without them:\n  %s",
+			len(lost), strings.Join(lost, "\n  "))
+	}
+
+	// And the specific false positive must be gone: no SCOPE.Operation named
+	// `OrderRepository` may be minted from OrderService.cs. This is what makes
+	// the edge assertion above non-accidental.
+	state := t.TempDir()
+	t.Setenv("GRAFEL_DAEMON_ROOT", t.TempDir())
+	t.Setenv("GRAFEL_INPROC_CUSTOM_EXTRACTORS", "1")
+	if err := Index(repo, filepath.Join(state, "graph.json"), "blazor_gate_6975", nil, false, false); err != nil {
+		t.Fatal(err)
+	}
+	doc, err := graph.LoadGraphFromDir(state)
+	if err != nil {
+		t.Fatal(err)
+	}
+	for i := range doc.Entities {
+		e := &doc.Entities[i]
+		if p, _ := e.PropLookup("provenance"); strings.HasPrefix(p, "INFERRED_FROM_BLAZOR_") &&
+			!strings.HasSuffix(e.SourceFile, ".razor") && !strings.HasSuffix(e.SourceFile, ".razor.cs") {
+			t.Errorf("blazor provenance %q on non-razor file %s (entity %s %q)",
+				p, e.SourceFile, e.Kind, e.Name)
+		}
+	}
+}
--- a/cmd/grafel/custom_extractor_dangling_6105_test.go
+++ b/cmd/grafel/custom_extractor_dangling_6105_test.go
@@ -204,6 +204,30 @@ public class Address {
     private String id;
 }
 `,
+		// #6975 — THE COLLISION TARGET, AND WHY IT LIVES IN ITS OWN FILE NOW.
+		// TestCustomExtractorDependsOnServiceDoesNotMisbind6123's non-vacuity
+		// guard needs an UNQUALIFIED SCOPE.Operation named
+		// `PostgreSqlContainer` to exist, or a mis-bind through the
+		// operation-family byName hint would be undetectable. That entity used
+		// to appear by accident: internal/custom/csharp/blazor.go had no file
+		// gate and its code-method pattern had no line anchor, so the
+		// `new PostgreSqlContainer()` call site below minted one. #6975 closed
+		// both holes, so the fixture now DECLARES the colliding name, in a
+		// `.razor.cs` code-behind file — the one extension that legitimately
+		// reaches that extractor. The base C# extractor cannot supply it: it
+		// qualifies method names (`OrderTests.Setup`), so a method of this name
+		// in an ordinary class would be `Fixtures.PostgreSqlContainer` and
+		// would not collide.
+		"cs/Fixtures.razor.cs": `namespace Shop.Tests
+{
+    public partial class Fixtures
+    {
+        void PostgreSqlContainer()
+        {
+        }
+    }
+}
+`,
 		"cs/OrderTests.cs": `using Testcontainers.PostgreSql;
 
 namespace Shop.Tests
@@ -688,9 +712,17 @@ func TestCustomExtractorSchemaFieldRefsStayInTheirOwnFile6105(t *testing.T) {
 // ExternalServiceTargetID at external_service.go:102), not by Name. So
 // `service:X` could never bind to a service node even where one existed, and
 // on a leaf-name collision it binds to whatever else is called X — here the
-// SCOPE.Operation the base C# extractor made for the `new PostgreSqlContainer()`
-// call site, reached through the DEPENDS_ON_SERVICE operation-family hint
-// (refs.go:1782).
+// SCOPE.Operation named `PostgreSqlContainer`, reached through the
+// DEPENDS_ON_SERVICE operation-family hint (refs.go:1782).
+//
+// #6975 CORRECTED THIS COMMENT'S ATTRIBUTION. That collider was never made by
+// the base C# extractor, which qualifies method names (`OrderTests.Setup`).
+// It was made by internal/custom/csharp/blazor.go, which had no file gate and
+// an unanchored code-method pattern, so `new PostgreSqlContainer()` in
+// ordinary C# minted an unqualified SCOPE.Operation of that name. That is the
+// defect #6975 fixed, so the fixture now declares the colliding name
+// explicitly in cs/Fixtures.razor.cs rather than relying on the bug to
+// supply it. See the note there.
 //
 // THE FIX. test_doubles.go now mints the canonical target ref rather than the
 // unaddressable Name. Where a matching service node exists the edge binds to it
--- a/internal/custom/csharp/blazor.go
+++ b/internal/custom/csharp/blazor.go
@@ -44,31 +44,67 @@ var (
 	// GRAFEL_INPROC_CUSTOM_EXTRACTORS, in-process Index()) this rule alone
 	// produced 21,320 entities in a repo with no Blazor at all.
 	//
-	// THE FIX IS AN ANCHOR, NOT A REQUIRED MODIFIER. A C# declaration begins
-	// its own line (only indentation precedes it), so `^[ \t]*` excludes every
-	// mid-expression `new X(` while still admitting the modifier-less methods
-	// that are idiomatic inside a `@code` block (`void Increment()`).
-	// Requiring an access modifier would have been the obvious tightening and
-	// is the WRONG one: measured over the 105 `.razor` files in
-	// archigraph-corpora, the old pattern finds 126 declaration sites, this
-	// one finds 110 (87%), and a required-modifier variant finds only 97
-	// (77%) — it drops exactly the bare `@code` methods.
+	// TIGHTENING 1 — AN ANCHOR, NOT A REQUIRED MODIFIER. A C# declaration
+	// begins its own line (only indentation precedes it), so `^[ \t]*`
+	// excludes every MID-EXPRESSION `new X(` while still admitting the
+	// modifier-less members that are legal inside a `@code` block
+	// (`void Increment()`), which a required access modifier would drop.
 	//
-	// The return-type alternation is `[\w\.\?]+(?:<[^;\n]*>)?(?:\[\])?`
+	// THE CORPUS DOES NOT ADJUDICATE THAT CHOICE, and an earlier revision of
+	// this comment claimed it did — it cited 110 sites for the anchor against
+	// 97 for a required-modifier variant and concluded the modifier "drops
+	// exactly the bare `@code` methods". Once tightening 2 lands, the two
+	// variants select the IDENTICAL 97 sites over the 105 `.razor` files in
+	// archigraph-corpora: 28 distinct names, zero sites found by either one
+	// alone, in either direction. Every real method in that corpus carries an
+	// access modifier. The anchor is preferred on mechanism, not on a
+	// measured margin: it rejects a match for being in the wrong POSITION,
+	// which is the actual defect, where a required modifier is a proxy that
+	// merely correlates with it on code that follows one convention. The
+	// modifier-less case is therefore pinned synthetically
+	// (TestBlazorGateAdmitsRazorCodeBehind6975), not by the corpus.
+	//
+	// TIGHTENING 2 — THE ANCHOR IS NOT ENOUGH ON ITS OWN, AND AN EARLIER
+	// REVISION OF THIS COMMENT CLAIMED IT WAS. An anchor only excludes `new X(`
+	// when something precedes it on the line. A LINE-LEADING one — the second
+	// line of `var svc =\n    new OrderService(db);`, or a bare statement — is
+	// still parsed as "type `new`, method `OrderService`" and mints exactly
+	// #6973's collider. Measured over the same corpus C#: 2,503 line-leading
+	// `new X(` sites, plus 1,661 `return F(` (type `return`, method `F`) and
+	// 293 `else if (` (type `else`, method `if`). The file gate keeps those out
+	// of `.cs`, but `.razor.cs` IS ordinary C#, so inside the gate the shape
+	// survives untouched.
+	//
+	// It is not only a `.cs` problem. On the corpus `.razor` files this
+	// rejection removes 13 of the anchor's 110 matches, and ENUMERATING them
+	// rather than counting shows all 13 are non-declarations: 9 `else if (`
+	// and 4 `await <Method>(` call sites — the latter minting operations named
+	// after a method being CALLED, which is the collider shape again. No
+	// declaration is lost on this corpus. The pattern therefore CAPTURES the type token as
+	// group 1 and Extract rejects the match when that token is a statement or
+	// expression keyword rather than a type — see blazorNonTypeKeywords. Go's
+	// RE2 has no lookahead, so this cannot be expressed in the pattern itself.
+	//
+	// TIGHTENING 3 — the return type is `[\w\.\?]+(?:<[^;\n]*>)?(?:\[\])?`
 	// rather than a fixed `Task|void|…` list, and the generic argument is
 	// `[^;\n]*` rather than `[^>\n]+`, because NESTED generics are common
 	// (`Task<List<Order>>`) and an inner-`>`-terminated class loses them
-	// silently — a mutant run caught that regression here before it shipped.
-	// Greediness plus the `\s+(\w+)\s*\(` tail makes it settle on the last
-	// `>` that still leaves a name and an open paren on the line.
+	// silently — a mutant run caught that regression before it shipped.
+	// The `\n` in that class is load-bearing in the OTHER direction and is
+	// graded separately (TestBlazorCodeMethodGenericDoesNotSpanLines6975):
+	// without it, `[^;]*` is greedy across newlines, so a `<` on one line
+	// reaches a `>` many lines below and invents a method out of two unrelated
+	// declarations. Greediness plus the `\s+(\w+)\s*\(` tail then makes it
+	// settle on the last `>` that still leaves a name and an open paren.
 	//
-	// Over the 3,721 `.cs` files in the same corpora the old pattern found
-	// 38,983 sites and this one still finds 23,474: the anchor alone does NOT
-	// solve the false-positive problem, the file gate in Extract does. What
-	// the anchor buys is that the `new X(` class does not return the moment
-	// another dialect widens that gate.
+	// WHAT EACH TIGHTENING IS WORTH, over the 3,721 `.cs` files in the corpora:
+	// the old pattern found 38,983 sites, the anchor alone 23,474, and the
+	// anchor plus the keyword rejection 18,842. The regex alone does NOT solve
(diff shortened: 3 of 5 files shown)


cajasmota:
## Independent adversarial review — **APPROVE** (with 4 non-blocking findings, one worth a follow-up issue)

Independent checkout, detached at PR head `b6ccf91`, in
`/private/tmp/claude-501/-Users-jorgecajas-Projects-archigraph/c0fefed4-0195-4c81-9a94-07a7145710a4/scratchpad/review-6977-qz9/wt`.
macOS/APFS, `GOMAXPROCS=3`, isolated `GRAFEL_HOME`/`GRAFEL_DAEMON_ROOT`, `-count=1` throughout,
corpus copied to scratch. `go vet` clean, `gofmt -l` clean, `git status --porcelain` empty at end.
Every restore was `cp` from a shasum-verified backup; final shasums re-verified against the backups.

### Every headline number re-derived, not cited

`aspnetcore-mvc` copied to scratch, in-process `Index()`, **gate ON = `GRAFEL_INPROC_CUSTOM_EXTRACTORS=1`** (not the default, #6966):

| arm | entities (yours / mine) | `TESTS` (yours / mine) | `blazor.go` provenance |
|---|---|---|---|
| gate OFF | 39,765 / **39,789** | 43,621 / **43,621** | 0 / **0** |
| ON, before (parent `blazor.go`) | 75,174 / **75,193** | 36,982 / **36,982** | 25,479 / **25,479** (21,320 CODE_METHOD + 4,159 COMPONENT_REF) |
| ON, after | 50,266 / **50,294** | 40,888 / **40,888** | 0 / **0** |

- **58.8% recovery reproduces exactly**: (40,888 − 36,982) / (43,621 − 36,982) = 3,906 / 6,639 = **58.83%**.
- Entity gain retained: 50,294 / 39,789 = **+26.4%**.
- Entity totals differ from yours by +19…+28 (0.06%); `TESTS` and the `blazor.go` figures are byte-identical. I did not chase the entity delta — it does not touch any claim.

**Extra control you did not run — the loss direction at corpus scale, as edge *sets*.** I dumped every
resolvable `TESTS` edge as `from.Name -> to.Name` per arm and diffed the sets:

- edges the fix removes relative to ON-before: **2,057**
- of those, edges that also exist in the **gate-OFF** arm: **0**

Every edge the fix destroys is one the OFF arm never had — i.e. purely spurious chains minted by the
false positives (`…Test -> ActionContext -> CreateMockUrlHelper`). That is the strongest available
evidence the recovery is not itself a substitution, and it is the check #6973 says a count cannot do.

**Regex measurement re-run** over `archigraph-corpora` with the exact patterns (my census: 3,726 `.cs`,
105 `.razor`, **0 `.razor.cs`**, 1,129 `.cshtml`):

| | `@page` | `@inject` | `[Parameter]` | `@layout` | `@inherits` | tag | oldRE | newRE | required-modifier |
|---|---|---|---|---|---|---|---|---|---|
| `.cs` | 0 | 0 | 0 | 0 | 0 | 12,994 | 39,009 | 23,490 | 16,472 |
| `.razor` | 25 | 119 | 10 | 4 | 0 | 427 | **126** | **110 (87%)** | **97 (77%)** |
| `.cshtml` | 9 | 39 | 0 | 0 | 1 | 56 | 132 | 66 | 37 |

126 / 110 / 97 confirmed on the nose, denominator = the 105 `.razor` files. The anchor is a restriction,
not a modifier requirement — confirmed. Set-diffing old vs new sites on `.razor`: the 13 "new-only" hits
are line-number artefacts of the old pattern starting on the *previous* line, not genuinely new admissions;
the drops are `new EditContext(`, `throw new InvalidOperationException(`, `new StringBuilder(` — correct.

### Mutants (mine, independent of yours)

| # | mutant | verdict |
|---|---|---|
| **R1** | `test_doubles.go`: `ref := containerServiceRef(name)` → `ref := name` (revert #6123, i.e. break the thing the repaired fixture *guards*) | **DEAD** — and it fails with the real mis-bind, `SCOPE.Component:OrderTests -> SCOPE.Operation:PostgreSqlContainer`, not merely a changed string |
| **R2** | gate widened to any `.cs` (`HasSuffix(".razor.cs")` → `HasSuffix(".cs")`) | **DEAD** (`TestBlazorGateRejectsOrdinaryCSharp6975`) |
| **R3** | delete the `file.Language != "csharp"` guard | **ALIVE** — `internal/custom/csharp` **and** the full `cmd/grafel` package both green. Equivalent in production (the `TSTree != nil` dispatch guard already blocks `.razor`), but see finding 2 |
| **R4** | generic class narrowed back, `<[^;\n]*>` → `<[^>\n]+>` | **DEAD** (`TestBlazorGateAdmitsRazorCodeBehind6975`) — the nested-generic positive control is real |
| **R5** | fixture-strengthening, not code: move `new PriceCalculator(_rates)` in `TestBlazorCodeMethodIgnoresConstructorCalls6975` to line-leading position (collection initializer) | **the shipped code FAILS it** — see finding 1 |

Independent regression-pin check: with the **parent** `blazor.go` restored,
`TestBlazorGateKeepsTestsEdges6975` fails naming **2 destroyed `TESTS` edges** plus 6 blazor-provenance
entities on non-razor files. It is a genuine pin, not a passing decoration.

**Golden ratchet** (`scripts/quality/run.sh --ratchet`, isolated home): **exit 0 — "38 gated fixtures held
their baseline (29 at 100% must-have recall)"**. Confirmed, no baseline moved.

---

### (a) Can `.razor` reach a custom extractor? No — and the markup half is dead by construction

Driven, not read: indexed a temp repo with `Pages/Counter.razor` (`@page`/`@layout`/`@inherits`/`@inject`/
`<MyCustomCard/>`/`@code`) and `Pages/Counter.razor.cs`, gate ON. Result: **0 `INFERRED_FROM_BLAZOR_*`
entities from the `.razor` file**, 3 from the `.razor.cs` one. Your claim holds.

So `isRazorMarkup` is **never true in production**, and rules 1, 2, 4, 6, 7 are unreachable. I read this as
**intended dead code, correctly labelled** — the comment says so — so it is not a mistake. Two caveats:

- The stated rationale is **inert as written**: *"`.razor` is kept in the gate so this extractor becomes
  correct rather than newly wrong if a razor grammar is ever registered."* Registering a grammar would not
  be enough — `.razor` classifies as `"razor"` (classifier.go:345) and `Extract` returns at
  `file.Language != "csharp"` before the gate is consulted. My mutant **R3** deleting exactly that guard is
  ALIVE, which is the proof. Suggest amending to "…if a razor grammar is registered *and* the language
  guard is relaxed", so the next reader does not act on a one-step plan that takes two.
- `TestBlazorMarkupStillExtractedFromRazor6975` is described as "the vacuity guard for the gate as a whole:
  … Without this, gating to nothing at all would pass every test above." In production the gate *does* gate
  the markup rules to nothing; the test only passes because it hands a `.razor` path the fabricated language
  `"csharp"`. Worth saying so in the comment — the test is fine as a spec of intent, it just is not the
  vacuity guard it claims to be.

### (b) 58.8% — reproduces exactly. See table above.

### (c) The fixture repair **strengthened** the test. Verified in all three directions

- **(a) the base extractor really does qualify.** Driven: on `cs/Fixtures.razor.cs` and `cs/Ordinary.cs` the
  base C# extractor emits `Fixtures.PostgreSqlContainer`, `OrderTests.Setup`, `Resid.Index`. The old
  comment's attribution was wrong; your correction is right.
- **(b) the fixture genuinely supplies the entity.** The unqualified `SCOPE.Operation:PostgreSqlContainer`
  now carries `provenance=INFERRED_FROM_BLAZOR_CODE_METHOD` and `SourceFile=cs/Fixtures.razor.cs` — a real
  declaration in a file the gate legitimately admits. The guard was not weakened, nothing was deleted, no
  baseline moved.
- **(c) it still fails when the guarded thing breaks.** Mutant **R1** — the collider still attracts the
  mis-bind from its new file, and the test reports it as a mis-bind. This was the failure mode most worth
  worrying about (a "repair" that quietly makes the collider stop colliding), and it did not happen.

### (d) No assertion is a count where it should name edges

`cmd/grafel` names edges and reports the missing ones individually; the unit tests name entities in both
directions. The one cardinality assertion, `len(ents) != 0` in `TestBlazorGateRejectsOrdinaryCSharp6975`,
is a *zero*-count on the reject side, which is strictly stronger than naming. No objection.

---

### Findings

**1 — (worth a follow-up issue) the `new X(` collider class is NOT closed inside the files the gate admits.**
The comment claims *"what the anchor buys is that the `new X(` class does not return the moment another
dialect widens that gate."* That is not what the anchor buys. `^[ \t]*` excludes *mid-expression*
`new X(` only; a **line-leading** one still matches, because `new` is consumed as the return type. Driven
against the shipped binary in a `.razor.cs` file:

```csharp
        private static readonly object[] Attrs = new object[]
        {
            new TagHelperAttribute("a", "b"),   // -> SCOPE.Operation "TagHelperAttribute"  (unqualified)
        };
        public IActionResult Index()
        {
            else if (y) { }                     // -> SCOPE.Operation "if"
            return View();                      // -> SCOPE.Operation "View"
        }
```

Of the 23,490 sites the new pattern still finds in the corpus's ordinary C#, **2,503 are line-leading
`new X(`** — the exact #6973 collider shape, an unqualified operation duplicating a real class name —
plus 1,666 `return F(`, 293 `else if (`, 172 `await`. The gate keeps all of that out of plain `.cs`, so
this is **not a blocker and not a regression** (the old pattern had it too, everywhere). But it fires
verbatim in `.razor.cs`, which is ordinary C#, so a real Blazor repo re-creates #6973 at code-behind scale.

**It is distinguishable and currently ungraded, and grading it costs 4 lines**: mutant **R5** — moving the
existing fixture's `new PriceCalculator(_rates)` onto its own line inside a collection initializer — makes
your own assertion fail against HEAD. A fix is ~5 lines (RE2 has no lookahead, so skip a match whose first
token is `new`/`return`/`else`/`await`/`yield`). Please either take the 4-line fixture line + the guard, or
file it and link it here so the prose claim and the code stop disagreeing.

**2 — the `.razor` future-proofing rationale needs one more clause** (see (a) above; evidence is mutant R3).

**3 — the admit side of the gate has zero coverage outside the new unit test.** The corpus contains **0
`.razor.cs` files** (my census; you report the same 0, but the PR body then leans on measurement 1 to
justify admitting that extension — measurement 1 has nothing to say about a population of size zero). And
`internal/quality/golden/razor-blazor-mini` contains four `.razor` files and one `.cs`, **no `.razor.cs`** —
so "razor-blazor-mini included" in the ratchet section is vacuous *for this change*: `blazor.go` contributed
nothing to that fixture before or after. The ratchet result is still a valid no-regression signal; it is just
not evidence that the gate preserves Blazor extraction. Consider adding a `.razor.cs` file to that fixture.

**4 — one PR-body sentence overreaches its own population.** *"the five `@`-directive rules score zero in
`.cs` and fire only in the 105 `.razor` files."* Over the stated population, which explicitly includes the
1,129 `.cshtml` files, they also fire there: `@page` 9, `@inject` 39, `@inherits` 1. Harmless in effect —
`.cshtml` is not in the classifier extension map at all, so it never reaches any custom extractor — but the
sentence should say "only `.razor` *among the extensions that reach this code*", or name `.cshtml` and why
it does not matter.

### Verdict

**APPROVE.** The diagnosis is right, the gate is the correct fix, the 58.8% is real and reproduces to the
edge, the edge-set control I added shows the fix destroys no edge the OFF arm has, the tests are genuine
regression pins, and the fixture repair strengthened the test it touched. Findings 2–4 are comment/coverage
accuracy. Finding 1 is a real residual that the PR's prose currently denies — I would like the 4-line
fixture change or a linked follow-up before merge, but I am not blocking on it.

cajasmota:
## Coordinator mutant on `b6ccf9155`

Own detached worktree, `git status --porcelain` empty, restored by `cp` from a SHA-256-verified backup (`4a966d87…` both ways), baseline green immediately before.

**MB-1 — remove the newline exclusion from the generic group.** `reBlazorCodeMethod`'s return type is `[\w\.\?]+(?:<[^;\n]*>)?(?:\[\])?`; I changed `[^;\n]` to `[^;]`, so `<…>` can now span lines.

**Result: ALIVE** — `./internal/custom/csharp/` green (0.8 s) and `./cmd/grafel/` green (142.9 s), both exit 0.

**Process note, because it nearly produced a false result.** The pattern `(?:<[^;\n]*>)?` occurs **twice** — once at line 57 in the doc comment describing the regex, once at line 71 in the regex itself. My first attempt asserted a single occurrence, so it aborted and applied nothing; the suite then came back green against an **unmutated file**, which would have read as ALIVE had I not been asserting the count. I re-applied to line 71 only and verified the comment still carries `[^;\n]` while the regex carries `[^;]`.

My own `grep -c` verification then returned `0` for both, which is shell-escaping noise rather than a failed edit — I confirmed with a direct read instead. Two instruments disagreeing is the moment to stop trusting the cheap one; this is the same family as the `bfs`/`find` error I already recorded.

**The verdict.** `\n` in that character class is doing real work — it stops a `<` on one line from swallowing to a `>` many lines later and inventing a method from two unrelated declarations. Nothing observes it. The class is the one my notes already flag: **Go character classes match newlines unless you exclude them**, and the exclusion here is deliberate but ungraded.

Cheap to grade: one fixture with a `<` on a line whose matching `>` is several lines below, asserting no operation is minted. A handful of lines.

## The reviewer's R5 residual is the one I would act on first

They found that `^[ \t]*` only excludes **mid-expression** `new X(` — a **line-leading** one still mints an unqualified `SCOPE.Operation` named after a real class. 2,503 such sites in the corpus's C#, plus 1,666 `return F(` and 293 `else if (`.

That is gated out of `.cs` by this PR, but **`.razor.cs` is ordinary C#**, so #6973's collider shape survives inside the gate. The PR's prose says the anchor addresses this; it addresses part of it. Their fix is ~5 lines with 4 to grade, and moving the existing fixture's constructor call onto its own line already makes your assertion fail.

Combined with **R3 (delete the `Language != "csharp"` guard) coming back ALIVE**, the gate currently has two unpinned edges and one ungraded regex detail. None of them is a defect in what shipped; all three are places where the next edit gets no signal.

## Two prose corrections worth making

- **"the five `@`-directive rules fire only in `.razor`" is false over its own stated population** — `.cshtml` carries 49 directive sites. Harmless, since `.cshtml` is unsupported (#6343), but the sentence is the evidence for the gate and should be true.
- **The ratchet line is vacuous for this change**: `.razor.cs` has zero files in both the corpus and the golden set, so "38 fixtures held baseline" says nothing about the extension this gate newly admits. Worth stating, because it is the difference between "the ratchet covers this" and "the ratchet cannot see this".

Everything else reproduced exactly — 58.83% recovery, the regex census, `.razor` unreachability driven rather than argued, and the fixture repair strengthened rather than gutted, with R1 proving the collider still attracts the mis-bind from its new file. That last one is the check I most wanted, because a bug being load-bearing for a test is where a repair usually guts it.

cajasmota:
## Second coordinator mutant, on `f872add0d`

My first was scored on `b6ccf9155`; this round changed production code (capture group, keyword map, call-site guard), so a verdict on the old head does not carry. Own detached worktree, `git status --porcelain` empty, restored by `cp` from a SHA-256-verified backup (`81247823…` both ways), baseline green immediately before.

**MB-2 — delete ONE entry from `blazorNonTypeKeywords`.** I removed `"throw"`, leaving the other 38. Edit proven landed: the hit set was exactly `[line 147]`, the line changed, and `"throw": true` is absent from the whole file afterwards.

**Result: ALIVE** — `./internal/custom/csharp/` green, exit 0.

## The verdict, stated narrowly

**`throw`'s own reachability is thin, and I want to be precise rather than alarming.** The common form `throw new ArgumentException(x)` is already caught by `"new"`, not by `"throw"` — the regex aligns on `new` as the type token and `ArgumentException` as the name. What `"throw"` uniquely guards is `throw Foo(...)`, which is legal C# when `Foo()` returns an exception but is uncommon. So this particular deletion is **alive and only narrowly production-reachable**.

**The finding is about the map, not about `throw`.** It is a 39-entry membership list, and at least one member can be removed with no signal. That means the map is graded at *some* members and not others, and nothing tells you which — so a future edit tidying the list, or a merge dropping a line, loses a guard silently. Given the list exists precisely because RE2 has no lookahead, it is the whole mechanism.

**Cheap to grade as an exact set**: a table test iterating every key in `blazorNonTypeKeywords` and asserting each one individually suppresses a synthetic line-leading `<kw> Foo(`. Roughly ten lines, it grades all 39 rather than the handful the corpus happens to exercise, and it fails when someone adds a keyword without a case. That converts membership into exact-set coverage — the distinction that has bitten this board repeatedly.

Your call whether to take it here or file it; I would take it, because the map is new in this PR and this is the moment it is cheapest.

## The self-correction you made unprompted is the best thing in this round

You went back and falsified your own justification: with the keyword rejection in place, the required-modifier and anchor variants select the **identical 97 sites** — 28 distinct names, zero found by either alone in either direction. **The 110-vs-97 margin does not exist**, and it was the evidence I relayed and the reviewer verified. Re-justifying the anchor on mechanism, and marking the modifier-less case as pinned **synthetically rather than by corpus**, is exactly right.

That is the third false claim found in this PR's own prose — after "fire only in `.razor`" and the ratchet line — and the only one found by the person who wrote it. Worth noting in the body that all three were found by *re-measuring*, never by re-reading.

Also noted: your `return F(` count is **1,661** against the reviewer's 1,666. Small, but say in the body that the two runs differ and which population you used, rather than letting two numbers for one thing sit in the record.

## Process

Your mutant script asserting the hit set is `[71, 90]`, that line 90 is the regex, and that **line 71 still carries `[^;\n]`** is the right shape. Mine caught an off-by-one on this very file for the same reason — the assertion aborted the edit and nothing ran, which is the failure mode I want, twice now.

Merging once CI is green, with MB-2 either graded or filed.

cajasmota:
**Settling the question I raised on #6978: this PR's near-miss record was correct, and I am recording that with the evidence rather than leaving it open.**

#6995's enumeration reported that `blazor.go`'s `isRazorMarkup` arm is dead — and since the file restricts its markup rules to that arm, its five `@`-directive rules and the component-tag rule never run. It summarised that as "the near-miss landed halfway", which would have meant this PR shipped a partial fix and I merged it.

**An independent check says otherwise, and the distinction is the whole point.** #6977 measured the five `@`-directive rules at **zero** in `.cs` files; only the component-tag rule fired there, and **every hit was a measured false positive**. So nothing that ever produced a true positive was lost.

**That makes this latent capability, not a regression.** The markup rules cannot fire today because `.razor` does not route to the csharp custom package — which is #6996's question, not this PR's. Copying `blazor_deep.go`'s gate here would still have disabled a working producer; that near-miss was real and it was avoided.

**One thing #6995 got wrong that matters for anyone reading this later:** routing `razor` → `custom_csharp_` would **not** by itself make the markup rules run. Both extractors early-return on `file.Language != "csharp"`, pinned by `TestBlazorLanguageGuardRejectsNonCSharp6975`, and `blazor.go`'s own comment twenty lines above the new one already said so. So the path from "the arm is dead" to "the rules would work if routed" has a second gate in it.

**Why I chased this at all**, since the answer turned out to be "no change": I merged this PR and recorded its near-miss as avoided. A later enumeration said it landed halfway. Those cannot both be right, and the one I had a stake in was the one I had least standing to adjudicate — so it went to someone else, with the two possible answers spelled out in advance so the finding could not be shaped by which I preferred. The record stands because it was checked, not because I defended it.

Refs #6995, #6978, #6996, #6975.

Issue #6979, opened 2026-09-07
MEASURED: 85% of gate-OFF byName evictions are genuine ambiguity and correctly refused — the real defect is the 262 same-construct pairs (#6981) and #4428's const-group naming (174 CALLS edges)


cajasmota:
Surfaced by #6976's measurement and split out because it is a **different problem with a different audience**: #6976 is about the custom-extractor lane, which is off by default (#6966). This one affects every user today.

## The measurement

On `aspnetcore-mvc` with the custom gate **OFF** — the default configuration, the one users actually run — **1,893 names are evicted from the repo-wide `byName` index**.

Each eviction takes the destructive default arm in `refs.go`'s `indexByName`: `delete(byName, name)` and `ambigName = true`, so the name becomes unresolvable for everyone. Anything relying on the global `byName` fallback — the testmap stub among them — silently fails to resolve.

**262 of the 1,893 are same-file pairs describing one construct**, which is the subset most likely to be wrong rather than genuinely ambiguous: two records for one thing, annihilating each other.

## Why this is filed separately from #6976

#6976 established that the gate-**ON** regression is *100% ranking, 0% policy* — 0 of its 3,236 ON-only evictions are declaration-vs-declaration. **That finding does not transfer here.** These 1,893 are base-path evictions with no custom lane involved, and nobody has yet classified them. The obvious hypotheses are all untested:

- genuine ambiguity, correctly refused (in which case this is working as designed and the issue closes);
- overloads or partial classes producing several records for one declaration;
- the 262 same-file pairs being two *views* of one construct, which #6104's facet rule rescues in some cases and not others.

**Classify them before designing anything.** The ratio decides whether this is a defect at all.

## What it shares with #6976, and what it does not

Shares the mechanism and the history: `delete` is original to `82d72fcff` (PORT-2-FIX), whose commit message states the unambiguous-only policy explicitly. It was correct when written, because the index then held only extractor declarations. **The premise was never revisited** as producers multiplied.

Does **not** share the diagnosis. #6976's answer — reference-shaped entities outranking declarations — is specific to the custom lane. Do not assume it applies; measure.

## What a resolution needs

1. **Classify the 1,893**, and the 262 same-file pairs specifically. Report how many are genuine ambiguity versus multiple records for one construct.
[...]

Issue #6467, opened 2026-08-21
resolve: the #6427 byName precedence has no locality tier, so a function-local const outranks an import placeholder and captures the package name repo-wide


arthurgeron:
I index a multi-repository corpus with grafel and hit this when the #6427 retarget landed: IMPORTS edges that had been pointing at `ext:<pkg>` started pointing at in-repo entities.

The `byName` precedence #6427 introduced ranks a record on one property, whether `isImportPlaceholderKind` says it is a per-import placeholder. Any record for which that is false counts as a declaration of the name, and against a placeholder it takes the repository-wide slot without either arm of the precedence looking at where it was declared. A `const` emitted inside a function body is not a placeholder, so it wins the slot for every file in the repository, and an import of a same-named package resolves to it instead of staying unresolved for the external-library binder.

## Example

For example, given a repository of two files, where `toolkit` is a declared dependency in `package.json` and `src/rows.tsx` imports nothing:

```tsx
// src/consumer.tsx
import { useToolkit } from "toolkit";

export function Panel() {
  const value = useToolkit();
  return <div>{value}</div>;
}
```

```tsx
// src/rows.tsx
export function Rows({ ids }: { ids: number[] }) {
  const toolkit: Array<{ id: number }> = [];
  ids.forEach((id) => toolkit.push({ id }));
  return <span>{toolkit.length}</span>;
}
```

I indexed that repository twice, on `2872b201a` (#6427's parent) and on `da80569f2` (main), with:

```
grafel group add repro --repo repro=<path> \
  --install=false --rules=false --mcp=false --watchers=false --git-hooks=false
grafel rebuild repro
grafel export cypher --group repro --out graph.cypher
```

The two exports differ by exactly one line. On the parent:

```
src/consumer.tsx -IMPORTS-> ext:toolkit
```

On main:

```
src/consumer.tsx -IMPORTS-> 8dc7162e164ff6ea
```

`8dc7162e164ff6ea` is the `const toolkit` on line 2 of `src/rows.tsx`. The cypher export shows it as a `SCOPE.Component` named `toolkit`; the entity record carries `subtype: "const"` and `properties["local_scope"]: "true"`. The edge should have stayed on `ext:toolkit`, which is what the parent produces. Nothing else moved: `ext:toolkit` still exists in the same export, the file's `DEPENDS_ON` still points at it, and the `useToolkit` call still binds to `ext:toolkit:useToolkit`. Only the module's own IMPORTS edge left.

[...]

PR #6980, opened 2026-09-07
fix(#6976): a mention of a name evicted the sole declaration of it — six producers now say so, and one tier reads it

### LINKED ISSUE #6973: DIAGNOSED (hypothesis falsified): blazor.go's missing .razor gate mints 25k false-positive name duplicates that evict real declarations from byName, collapsing DeriveTestsWalkUp's seeds — not a dangling-edge prune

### LINKED ISSUE #6976: MEASURED: the byName default arm encodes a premise that stopped holding — 'is a declaration' is not knowable there, and the gate-ON regression is 100% ranking, 0% policy

cajasmota:
Closes #6976. Option A as recommended on the issue, sequenced after #6975 (`448d7bb38`), which this branches from.

**Tree measured in:** `/private/tmp/claude-501/-Users-jorgecajas-Projects-archigraph/c0fefed4-0195-4c81-9a94-07a7145710a4/scratchpad/impl-6976b-k7v2r/wt`. `aspnetcore-mvc` copied to scratch; `/Users/jorgecajas/Projects/archigraph-corpora` untouched. In-process `Index()`, one arm per process, `GOMAXPROCS=3`, `nice -n 10`, isolated `GRAFEL_HOME`/`GRAFEL_DAEMON_ROOT`. macOS/APFS, Apple Silicon, 16 GB shared box.

**The gate, on every figure below (#6966):** OFF = `GRAFEL_INPROC_CUSTOM_EXTRACTORS` unset, ON/TIER = `=1`. This is not what users run. #6979 owns the gate-OFF population and is **not** folded in — with the gate off this arm is inert, verified below.

---

## The fix, both halves

**Upstream first, because the discriminator is not on the record.** The issue's diagnosis established that `indexByName` cannot derive "declaration or mention?" from a `types.EntityRecord`: no field means it, `provenance` is a producer id rather than a semantic class, `custom/csharp/helpers.go:15` sets `StartLine == EndLine` for the entire C# custom lane so span ranks the lane and not the record, and the same `makeEntity` stamps `QualityScore` 1.0 uniformly. So the emit site stamps it.

**1. `Properties["reference_shaped"] = "true"`, at six emit sites.** The property convention mirrors #6467's `local_scope` — an extractor-stamped key with one named predicate in `internal/resolve` — rather than a new `EntityRecord` field or a `Subtype` convention, because `Subtype` is load-bearing for denoise, docs and the #6104 facet rule and overloading it would move consumers that have nothing to do with resolution.

| producer | what it reads |
|---|---|
| `INFERRED_FROM_DOTNET_DI_PROVIDER` | a constructor **parameter type**; the class is declared in its own file |
| `INFERRED_FROM_ASPNET_RETURN_TYPE` | an `ActionResult<T>` **return annotation** |
| `INFERRED_FROM_ASPNET_FROM_BODY` | a `[FromBody]` **parameter annotation** |
| `INFERRED_FROM_FROM_BODY` | the validation lane's `[FromBody]` rule, same shape |
| `INFERRED_FROM_BLAZOR_COMPONENT_REF` | a `<Foo>` **markup tag** — a use of the component declared in `Foo.razor` |
| `INFERRED_FROM_BLAZOR_INJECT` | `@inject IFoo Foo` — a service declared elsewhere |

Auditing the ~340-producer custom lane is explicitly **out of scope**; an unmarked producer keeps its pre-#6976 behaviour exactly, and `TestMarkedProducersAreExactlyTheMeasuredSix6976` fails if the stamp leaks past these six.

**One substitution against the issue's recommended list, on the merits.** The issue named `BLAZOR_CODE_METHOD` among the four. `reBlazorCodeMethod` matches a **method declaration** in an `@code` block — its own comment says so — so marking it would be a false label, and its whole population (959 of the 1,222 sole-declaration evictions) is already dead with #6975's file gate. It is in the negative control instead, alongside `BLAZOR_PAGE` and `BLAZOR_PARAMETER`. `BLAZOR_INJECT` takes the sixth slot: same reference shape as `DOTNET_DI_PROVIDER`, same lane.

**2. One tier in `indexByName`, the #6369/#6467 shape, three clauses.** A reference neither displaces an unmarked incumbent nor blanks it; an unmarked record **reclaims** the slot from a reference extracted first (extraction order is not stable, so both directions are needed); two references still collide, and `nameAmbigRef` remembers that so a declaration arriving later can still reclaim. A reference is also barred from reclaiming placeholder-only ambiguity and from displacing a placeholder incumbent, for the reason #6467 barred a local from both. **Ambiguity is not traded away** — this narrows *which records compete*, it does not soften the default arm, so `lookupBareWithLocality` and `external.Synthesize` still get every case they exist to rescue.

**Scoped to CROSS-FILE pairings**, which the golden ratchet forced and which the issue body already argued for: *"#6104's facet rule rescues same-name pairs that are two views of one thing in one file. It does not apply here because the claimants live in different files."* See the ratchet section — this was a real regression, caught, diagnosed and fixed, not a baseline bump.

---

## Measured

Edge **sets**, not counts — `TESTS` `(from, to, source)` triples:

| | OFF | ON | **TIER** |
|---|---|---|---|
| `TESTS` | 43,620 | 40,887 | **44,192** |
| lost vs OFF | — | 6,353 | **1,227** |
| gained vs OFF | — | 3,620 | 1,799 |
| **targets that resolve to a real entity id** | 17,135 | 13,737 | **18,863** |
| dangling targets | 60.7% | 66.4% | **57.3%** |
| entities | 39,766 | 50,260 | **50,260** |
| `ENTRY_POINT_OF` | 73 | 69 | **73** |
| `STEP_IN_PROCESS` | 221 | 219 | **226** |
| `CONTAINS` | 71,674 | 90,118 | 90,118 |

**5,126 specific lost edges come back** — 3,305 `tests-walkup` and 1,821 explicit — and **5,126 of 5,126 point at an entity that exists in the tier graph.** The entity gain is untouched (50,260 = 50,260). **The earlier claim that "the only three kinds that move are TESTS, STEP_IN_PROCESS and ENTRY_POINT_OF" was wrong, and wrong in this PR's own diagnosed way — it was a per-kind CARDINALITY reading, and a cardinality reading cannot see a substitution.** The corrected set-level result is below.

(The ON figures differ from the issue's because #6975 landed in between: the Blazor lane no longer runs over every `.cs` file, so entities are 50,260 rather than 75,191 and the ON-arm damage is 6,353 rather than 14,070. Both arms re-measured on this tree; the OFF arm reproduces the issue's 39,766 entities exactly.)

### Named recovered edges, and their producers

A test whose entire fan-out was destroyed, `Endpoints_AccessParameters_InitializedFromProvider -> ActionDescriptorCollection` (`0037057ffef3c61c`):

```
explicit  OFF  -> 9c285b91edbadfc0  ActionDescriptorCollection [SCOPE.Component]
                  src/Mvc/Mvc.Core/src/Infrastructure/ActionDescriptorCollection.cs
          ON   -> scope:operation:src/Mvc/Mvc.Core/test/Routing/
                  ActionEndpointDataSourceBase.cs#ActionDescriptorCollection   UNRESOLVED
          TIER -> 9c285b91edbadfc0                                             restored

walk-up   OFF  -> 36e49e60b72304fc  DefaultActionDescriptorCollectionProvider.UpdateCollection
                  6b8fb52810da88ac  MockActionDescriptorCollectionProvider…(Benchmark)
                  ac1c4b1358b8e0c6  MockActionDescriptorCollectionProvider…(ActionSelector)
          ON   -> (none — the whole walk-up fan-out is gone)
          TIER -> all three restored, all three targets exist
```

Top recovered target names: `ValidationVisitorBenchmarkBase.GetActionContext` (338 edges), `RuntimePerformanceBenchmarkBase.IterationSetup` (254), `ActionEndpointFactory.CreateRequestDelegate` (254), `MvcAttributeRouteHandler.RouteAsync` (254), `MvcRouteHandler.RouteAsync` (254), `ActionContext` (254), `ModelStateDictionary` (205), `MvcOptions` (172), `ActionDescriptor` (166), `TagBuilder` (94).

Attribution — the custom-lane entities that share a name with a recovered target, by provenance: `DOTNET_DI_PROVIDER` 121, `ASPNET_RETURN_TYPE` 36, `FROM_BODY` 1, `ASPNET_FROM_BODY` 1, `DATA_ANNOTATION` 1 (declaration-shaped, correctly unmarked). **The producers doing the evicting are the producers that got marked.**

### The reverse direction, graded

Confident (hex-target) `TESTS` edges present in the arm but **not** in OFF: **ON 1,777, TIER 1,777 — identical.** The tier does not add a single confidently-wrong binding; it stops destroying correct ones. That is the #6369 hazard this could most easily have traded into, and it did not.

### ON → TIER, as a SET (this is the corrected finding 2)

**Gate ON throughout (#6966).** Reproduced on my own tree this round, not cited: parent
`448d7bb38` ON arm and this branch's TIER arm, in-process `Indexer.Run`, `GOMAXPROCS=3`,
`nice -n 10`, isolated `GRAFEL_HOME`/`GRAFEL_DAEMON_ROOT`, corpus copied to scratch.

```
ON -> TIER whole-graph set diff:   only_on 5,636 rows    only_tier 8,952 rows
kinds whose in and out counts are EQUAL (pure substitution, invisible to a count table):
  CALLS 2,989  USES 257  IMPLEMENTS 133  RETURNS 108  INJECTED_INTO 70  CONTAINS 55
  EXTENDS 35  BINDS 17  ACCEPTS_INPUT 12  REGISTERS 9  RESOLVES_TO 5     = 3,690 rows
```

So **~3,690 non-`TESTS` edges change target with zero net count movement**, across eleven
kinds the old sentence did not mention. The three kinds that move in COUNT (`TESTS` +3,305,
`STEP_IN_PROCESS` +7, `ENTRY_POINT_OF` +4) were the only three that a per-kind count could
ever have revealed. This is #6973's blindness reappearing inside the PR that fixes #6973,
and it was the reviewer who caught it.

Classifying every `only_tier` row against the `only_on` rows with the same
`(kind, from, source)`:

| | rows |
|---|---|
| **UNRES → HEX** — the intended win | **5,050** (CALLS 2,629, TESTS 1,821, USES 257, RETURNS 108, IMPLEMENTS 94, INJECTED_INTO 66, EXTENDS 32, BINDS 17, ACCEPTS_INPUT 12, REGISTERS 9, RESOLVES_TO 5) |
| **HEX → HEX** — a confident target replaced by a DIFFERENT confident target | **406** (CALLS 360, IMPLEMENTS 39, INJECTED_INTO 4, EXTENDS 3) |
| → still unresolved | 45 |
| no `only_on` counterpart (net new edges, not substitutions) | 3,451 |

**The HEX→HEX population is the #6369 hazard class and it shipped ungraded.** The PR's
reverse-direction check cannot see it by construction: that check is `TESTS`-only and covers
only edges ABSENT from OFF, whereas every one of these is present in all three arms. Graded
against the OFF arm, which is the closest thing to ground truth available here:

| | count |
|---|---|
| TIER's new target is among OFF's targets for that `(kind, from, source)` — a **correction** | **402** |
| the pairing is absent from OFF entirely — ungradeable | 4 |
| **OFF endorsed the ON target and TIER moved AWAY** | **0** |

Under a stricter rule that demands OFF have exactly ONE target for the pairing before it is
allowed to vote: **69 corrections, 0 disagreements, 337 pairings where OFF is multi-target
and the comparison is not 1:1** (TIER's pick is inside OFF's set in all of them). Both
gradings agree on the number that matters: **not one substitution moves an edge off a target
the no-custom-lane arm endorsed.**

**Credit where it is due.** This population, its size, and the favourable direction of the
grading were all found by the independent reviewer, not by me — my own measurement asserted a
per-kind count and stopped there. My partition differs from theirs in the pairing rule (they
report UNRES→HEX 5,104 / HEX→HEX 271 with 85 corrections, 119 `ext:`→local and 56
uncorroborated third-target picks; I group by `(kind, from, source)` and cover every row,
which moves rows between the buckets). **The two independent partitions agree on the only
claim being made from them: zero edges move off an OFF-endorsed target.**

### Inert with the gate off

The gate-OFF dump is **byte-identical** between this branch and its parent `448d7bb38`: `cmp` exit **0** on **250,928 rows** (39,766 entity rows + 211,162 relationship rows) for `aspnetcore-mvc`. **Re-established after the round-2 fix below**, not carried over — that fix touches the reclaim branch, so the property was re-measured rather than assumed. No user's graph changes.

The reviewer added the stronger argument, which holds for every repo and not just the one measured: `reference_shaped` has exactly ONE writer in the tree, so with the gate off `isRef` is universally false, both new `case` arms are unreachable, every new conjunct is an identity, and the trailing `delete` runs on an always-empty map. **Inert by construction, not merely measured-inert.**

---

## Verification

- `go test -count=1` green on `./internal/resolve/`, `./internal/custom/csharp/` and `./cmd/grafel/` (the last because an extractor change moves `cmd/grafel`'s whole-graph digest pin). `gofmt -l` clean, `go vet` clean.
- **Golden ratchet** (`scripts/quality/run.sh --ratchet`): **OK — 38 gated fixtures held their baseline, 29 at 100% must-have recall.**

**The ratchet caught a real regression, and the enumerated fixture member is the fix's shape.** The first revision failed it: `csharp-aspnet-core-mini` `relationship_found` 13 → 12, on exactly one member — `UserService -IMPLEMENTS-> IUserService`. `IUserService` is declared **nowhere** in that fixture; it has two claimants in **one** file, `Controllers/UsersController.cs`: the .NET DI pass's `SCOPE.Class` (now marked) and `internal/engine/detector.go`'s YAML rule-pack `Dependency` (unmarked, and *also* a mention — that producer is out of scope). Unmarked-beats-marked made the pair stop colliding, the `Dependency` record took the slot outright, and the edge moved off the `SCOPE.Class` node the fixture pins. Ambiguity had been carrying that edge to a better answer through the kind-hint path. Scoping the tier to cross-file pairings restores it, costs **nothing** on the corpus (all figures above are post-scoping and identical to pre-scoping), and is pinned by `TestSameFilePairingIsUnchanged_6976` with `TestCrossFileTierStillFiresWhenSameFileIsScopedOut_6976` as its control. No baseline constant was touched.

## Round 2 — the cross-file scoping missed the reclaim path (`991f901f9`)

**Tree measured in:** `/private/tmp/claude-501/-Users-jorgecajas-Projects-archigraph/c0fefed4-0195-4c81-9a94-07a7145710a4/scratchpad/impl-6976b-k7v2r/wt`; parent arm in a detached worktree at `448d7bb38`. macOS/APFS, Apple Silicon.

The tier is scoped to CROSS-FILE pairings on two of the three places that can change a
same-file outcome. **The reclaim path out of `ambigName` did not get the guard.** So three
claimants in ONE file — two reference-shaped, one unmarked — bound to the **unmarked** record
in claimant order `ref, ref, decl` and stayed ambiguous in the other two rotations, against a
parent that is ambiguous in all three. Two things wrong with that:

- it is **order-DEPENDENT binding** in exactly the shape the cross-file scoping exists to
  prevent — in a file whose own comments say *"extraction order is not stable, so both
  directions are needed"*; and
- the winner is the **unmarked** record, which is the `IUserService` mechanism that regressed
  `csharp-aspnet-core-mini` 13 → 12. That fixture escapes only because it has **two**
  claimants rather than three.

**The guard has to live where `nameAmbigRef` is RAISED, not where it is read.** By the time
the reclaim branch runs, `nameHolderFile` has already been deleted, so the reviewer's
suggested `&& idx.nameHolderFile[name] != srcFile` **at the reclaim site** would compare
against an empty string and fire unconditionally — a no-op guard. Raising the flag only for a
cross-file reference pair is what makes the reclaim it authorises cross-file. That is their
second suggested form, and it is the one that works.

**Priced at ~4 production lines + ~20 test lines; the actual diff is one conjunct and 135
test lines.** The production change is:

```go
-  if isRef && idx.nameHolderRef[name] {
+  if isRef && idx.nameHolderRef[name] &&
+      idx.nameHolderFile[name] != srcFile {
```

The test side came in over the estimate on purpose: the cases are **enumerated, not sampled**.

### The test now observes what its name claims, across the whole space it names

`TestSameFilePairingIsUnchanged_6976` covered same-file **pairs** in both orders and its name
was read as covering same-file claimants generally — the reviewer exhibited a same-file set of
three that changes. Three rounds of hand-picked cases have now missed two holes on this
branch, so the new cases enumerate:

- `allOrders6976` runs a case in **every** claimant order and **fails if the enumeration is
  not n!** — a permutation helper that silently ran 4 of 6 orders would reproduce the class of
  defect it exists to catch.
- `TestSameFileThreeClaimantsStayAmbiguous_6976` — all **3! = 6** orders of two same-file
  references plus one same-file unmarked record; all six must stay unresolved.
- `TestSameFilePairingIsEnumerated_6976` — the 2-claimant set over its full 2! space, so the
  same-file claim is observed at both sizes it can have.
- `TestCrossFileThreeClaimantsStillReclaim_6976` — the **control**: the same three-record shape
  spread across three files must land on the declaration in all six orders, including
  `ref, ref, decl`, which is the one order that goes through the narrowed reclaim path.

### Mutants on the new conjunct — both directions, 2/2 DEAD

Each edit proven landed with an exact occurrence count **and** an assertion on the hit set
(anchor unique, count 1, mutated line re-grepped); scored with `-count=1`; restored by `cp`
from a sha256-verified backup (`c91d1972…d930c` → `694b0f2c…e0ab6`), zero `MUTANT` text
remaining afterwards.

| # | mutant | verdict |
|---|---|---|
| **M1** | drop the conjunct entirely — the **MORE PERMISSIVE** direction, i.e. the shipped bug | **DEAD** — `TestSameFileThreeClaimantsStayAmbiguous_6976`, and **only** the two `ref,ref,decl` orders fail (`d…02,d…03,d…04` and `d…03,d…02,d…04`). That is the reviewer's exhibit reproduced to the order. |
| **M2** | invert it to `== srcFile` — the restrictive direction | **DEAD** — `TestCrossFileThreeClaimantsStillReclaim_6976` (both `ref,ref,decl` orders) **and** `TestDeclarationReclaimsFromReferenceOnlyAmbiguity_6976`. |

The two halves of the conjunct are therefore graded **part by part with no masking pair**: M1
fails only the same-file test, and the cross-file reclaim tests are what catch M2. That keeps
true of the new line what the reviewer's V3/V4 established for the reference arm.

### Cost on the corpus: none

The gate-ON TIER dump for `aspnetcore-mvc` is **identical row for row before and after this
change** — 283,371 rows, `cmp` exit 0 on the sorted dumps — so every figure in the *Measured*
section above stands unchanged and was not re-derived from a stale file. Golden ratchet
re-run: **OK, 38 gated fixtures held their baseline, 29 at 100% must-have recall.**
`go test -count=1` green on `./internal/resolve/`, `./internal/custom/csharp/` and
`./cmd/grafel/`; `gofmt -l` and `go vet` clean.

## `e50c8653e` — "six MEASURED producers" was one measurement short

The reviewer's wording nit, fixed rather than argued: `wantMarked6976` and the scope note on
`resolve.ReferenceShapedProp` both said "the SIX **measured** producers in the
`aspnetcore-mvc` eviction population". **Five** were measured there. The sixth,
`INFERRED_FROM_BLAZOR_INJECT`, was justified by **shape** — `@inject IFoo Foo` names a service
declared elsewhere, the same shape as `DOTNET_DI_PROVIDER`, in the same lane — and #6976 names
it nowhere in that population. Both sites now say five measured plus one classified by shape.
Comment-only; `./cmd/grafel/` re-run anyway because `refs.go` is touched and it digest-pins
the whole graph.

## Mutation score — 16/16 DEAD, more-permissive direction first

Every mutant applied with an exact occurrence assertion, scored with `-count=1`, file restored and re-diffed after each.

| # | mutant | verdict |
|---|---|---|
| R1 | ref arm deleted — a reference may evict a declaration | DEAD `TestDeclarationOutranksReference_6976` |
| R2 | reclaim arm deleted — a declaration cannot take the slot back | DEAD same, `b-then-a` order |
| R3 | ref arm drops `!nameHolderRef` — reference outranks reference | DEAD `TestTwoReferencesStillCollide_6976` |
| R4 | `ambigName` guard drops `isRef` — a reference reclaims ambiguity | DEAD `TestReferenceDoesNotReclaimPlaceholderOnlyAmbiguity_6976` |
| R5 | import-displacement arm drops `!isRef` | DEAD `TestReferenceDoesNotOutrankImportPlaceholder_6976` |
| R6 | placeholder-yield arm drops `!nameHolderRef` | DEAD same, other order |
| R7 | `nameAmbigRef` never recorded | DEAD `TestDeclarationReclaimsFromReferenceOnlyAmbiguity_6976` |
| R8 | tail AND-ing dropped — a trailing mention re-raises reference status | DEAD `TestSameIDReindexDoesNotReRaiseReferenceStatus_6976` |
| R9 | cross-file scoping dropped from the ref arm | DEAD `TestSameFilePairingIsUnchanged_6976` |
| R10 | cross-file scoping dropped from the reclaim arm | DEAD same, other order |
| R11 | `nameHolderFile` not updated when a declaration reclaims | DEAD `TestHolderFileFollowsTheSlot_6976` |
| R12 | `nameHolderFile` not updated when a declaration displaces a placeholder | DEAD `TestHolderFileFollowsTheSlotAfterPlaceholderDisplacement_6976` |
| R13 | `nameHolderFile` not updated on the tail insert | DEAD `TestSameFilePairingIsUnchanged_6976` |
| P1 | one `markReferenceShaped` call deleted | DEAD that producer's row, by name |
| P2 | stamp applied to the whole C# lane in `makeEntity` | DEAD all three negative-control rows |
| P3 | stamped value is not `"true"` | DEAD every marked row |

R11 and R12 were **ALIVE** on the first scoring — the stale-holder-file cases — and were graded rather than argued away; each cost one ~20-line test. R13's first mutant text was not unique in the file and was reported unscored rather than silently passing; it is scored above with a unique anchor.

## What the tests assert, and why not counts

This defect survived a measurement that reported explicit `TESTS` at **+8 net while 7,221 edges were substituted** (#6973) — a cardinality assertion is blind to a substitution by construction. So every case in `reference_decl_tier_6976_test.go` names the id the edge must land on, runs the production `BuildIndex` + `ReferencesEmbedded` path, fails loudly if the fixture stops carrying exactly one `TESTS` edge, and runs in **both extraction orders**. `wantMarked6976` / `wantUnmarked6976` are hand-written **independent literals**, not derived from any production table, so deleting a `markReferenceShaped` call turns a row red by name.

Both directions are graded: `TestTwoDeclarationsStillCollide_6976` pins that two real declarations of one name stay ambiguous, and `TestUnmarkedRecordIsUnchanged_6976` is the control that the tier reads the marker and nothing derivable from the record.

Refs #6973, #6975, #6369, #6467, #6104, #6966, #6979.

### DIFF
--- a/internal/custom/csharp/aspnet_request_response.go
+++ b/internal/custom/csharp/aspnet_request_response.go
@@ -78,6 +78,9 @@ func (e *aspnetReqRespExtractor) Extract(ctx context.Context, file extractor.Fil
 		ent := makeEntity(dtoType, "SCOPE.Component", "", file.Path, file.Language, lineOf(src, m[0]))
 		setProps(&ent, "framework", "aspnet_core", "provenance", "INFERRED_FROM_ASPNET_FROM_BODY",
 			"dto_kind", "request")
+		// #6976 — a `[FromBody] Foo foo` parameter annotation mentions the DTO
+		// type; the DTO class is declared in its own file.
+		markReferenceShaped(&ent)
 		add(ent)
 	}
 
@@ -102,6 +105,9 @@ func (e *aspnetReqRespExtractor) Extract(ctx context.Context, file extractor.Fil
 		ent := makeEntity(typeName, "SCOPE.Component", "", file.Path, file.Language, lineOf(src, m[0]))
 		setProps(&ent, "framework", "aspnet_core", "provenance", "INFERRED_FROM_ASPNET_RETURN_TYPE",
 			"dto_kind", "response")
+		// #6976 — an `ActionResult<Foo>` return annotation mentions `Foo`; the
+		// declaration is elsewhere.
+		markReferenceShaped(&ent)
 		add(ent)
 	}
 
--- a/internal/custom/csharp/blazor.go
+++ b/internal/custom/csharp/blazor.go
@@ -296,6 +296,8 @@ func (e *blazorExtractor) Extract(ctx context.Context, file extractor.FileInput)
 			ent := makeEntity(serviceType, "SCOPE.Component", "", file.Path, file.Language, lineOf(src, m[0]))
 			setProps(&ent, "framework", "blazor", "provenance", "INFERRED_FROM_BLAZOR_INJECT",
 				"service_type", serviceType)
+			// #6976 — `@inject IFoo Foo` names a service DECLARED elsewhere.
+			markReferenceShaped(&ent)
 			add(ent)
 		}
 	}
@@ -327,6 +329,9 @@ func (e *blazorExtractor) Extract(ctx context.Context, file extractor.FileInput)
 			}
 			ent := makeEntity(name, "SCOPE.UIComponent", "component", file.Path, file.Language, lineOf(src, m[0]))
 			setProps(&ent, "framework", "blazor", "provenance", "INFERRED_FROM_BLAZOR_COMPONENT_REF")
+			// #6976 — a `<Foo>` markup tag is a USE of the component declared
+			// in Foo.razor, not a second declaration of the name `Foo`.
+			markReferenceShaped(&ent)
 			add(ent)
 		}
 	}
--- a/internal/custom/csharp/dotnet_di.go
+++ b/internal/custom/csharp/dotnet_di.go
@@ -186,6 +186,10 @@ func (e *dotnetDIExtractor) Extract(ctx context.Context, file extractor.FileInpu
 			ent := makeEntity(provider, "SCOPE.Class", "", file.Path, "csharp", c.line)
 			setProps(&ent, "framework", "dotnet_di", "provenance", "INFERRED_FROM_DOTNET_DI_PROVIDER",
 				"di_role", "provider")
+			// #6976 — the carrier is a CONSTRUCTOR PARAMETER TYPE; the class
+			// it names is declared in its own file. Largest single producer
+			// in the measured eviction residual (403 claimants).
+			markReferenceShaped(&ent)
 			ent.Relationships = append(ent.Relationships, types.RelationshipRecord{
 				ToID: "consumer:" + c.name,
 				Kind: string(types.RelationshipKindInjectedInto),
--- a/internal/custom/csharp/helpers.go
+++ b/internal/custom/csharp/helpers.go
@@ -197,3 +197,36 @@ func csSharedDispatchVerbOwner(src string) string {
 	}
 	return ""
 }
+
+// referenceShapedProp is the #6976 producer-side marker: this record was
+// minted from a MENTION of a name — a constructor parameter type, an
+// `ActionResult<T>` return annotation, an `@inject` service type, a `<Foo>`
+// markup tag — and NOT from the declaration of that name, which lives in
+// another file and is extracted separately.
+//
+// WHY IT IS STAMPED HERE AND NOT DERIVED DOWNSTREAM. internal/resolve's
+// byName index has the whole types.EntityRecord and still cannot tell the two
+// apart: no field means it, `provenance` is a producer id rather than a
+// semantic class, makeEntity above stamps StartLine == EndLine for the ENTIRE
+// C# custom lane (genuine declarations included) so span is a lane signal, and
+// the same makeEntity stamps QualityScore 1.0 uniformly. The emit site is the
+// only place that knows. Without the marker, a mention of `JsonResult` in one
+// controller evicted the sole real declaration of `JsonResult` from the
+// repository-wide name index and took every bare-name edge to it down with it.
+//
+// The spelling is shared with resolve.ReferenceShapedProp, which is the only
+// reader. The two are literal strings on both sides — the same arrangement
+// `local_scope` (#6467) uses between the JS extractor and the same reader —
+// and reference_shaped_marker_6976_test.go pins them equal so they cannot
+// drift.
+//
+// SCOPE (#6976). Stamped on the six measured reference-shaped producers in the
+// `aspnetcore-mvc` eviction population, NOT across the ~340-producer custom
+// lane. An unmarked producer keeps its pre-#6976 behaviour exactly.
+const referenceShapedProp = "reference_shaped"
+
+// markReferenceShaped stamps referenceShapedProp on a record. One helper so
+// every stamped emit site is greppable through a single symbol.
+func markReferenceShaped(e *types.EntityRecord) {
+	setProps(e, referenceShapedProp, "true")
+}
--- a/internal/custom/csharp/reference_shaped_marker_6976_test.go
+++ b/internal/custom/csharp/reference_shaped_marker_6976_test.go
@@ -1,0 +1,230 @@
+package csharp_test
+
+import (
+	"context"
+	"testing"
+
+	_ "github.com/cajasmota/grafel/internal/custom/csharp"
+	extreg "github.com/cajasmota/grafel/internal/extractor"
+	"github.com/cajasmota/grafel/internal/resolve"
+	"github.com/cajasmota/grafel/internal/types"
+)
+
+// #6976 — the producer side of the declaration-outranks-reference tier.
+//
+// internal/resolve cannot derive "is this record a declaration or a mention?"
+// from the record: no field means it, `provenance` is a producer id rather
+// than a semantic class, helpers.go's makeEntity stamps StartLine == EndLine
+// for the ENTIRE C# custom lane so span ranks the lane and not the record, and
+// the same makeEntity stamps QualityScore 1.0 uniformly. The emit site is the
+// only place that knows, so the marker is stamped there and this file pins it.
+//
+// WHY THE EXPECTATION IS AN INDEPENDENT LITERAL. wantMarked6976 below is
+// hand-written, NOT derived from any production table. A test that ranges over
+// the collection under test cannot detect a deletion from it — that exact
+// mistake shipped once in this area and was caught only by re-scoring (#6975
+// follow-up). Deleting a markReferenceShaped call must turn a row RED.
+//
+// THE MUTANTS THESE ARE WRITTEN TO KILL, more-permissive direction first:
+//
+//	m1 — delete any one of the six markReferenceShaped calls
+//	     => that producer's row in TestReferenceShapedProducersAreMarked6976
+//	        fails by name.
+//	m2 — stamp markReferenceShaped unconditionally in makeEntity (the
+//	     "mark the whole lane" shortcut)
+//	     => TestDeclarationShapedProducersAreNotMarked6976 fails: an @page
+//	        route, an [ApiController] route and a Blazor @code method are
+//	        declaration-shaped and must stay unmarked.
+//	m3 — change the stamped VALUE to anything but "true"
+//	     => every marked row fails, because resolve.isReferenceShaped tests
+//	        for the exact string.
+//	m4 — rename the property on one side only
+//	     => TestReferenceShapedPropSpelling_6976 in internal/resolve fails,
+//	        and so does every row here, because the rows read the constant
+//	        through resolve.ReferenceShapedProp rather than a local copy.
+
+// wantMarked6976 is the closed scope of #6976: the SIX measured producers in
+// the `aspnetcore-mvc` eviction population that mint an entity from a MENTION
+// of a name declared elsewhere. Written out by hand.
+//
+// Deliberately NOT a repo-wide taxonomy of the ~340 custom producers — see the
+// scope note on resolve.ReferenceShapedProp. Adding a seventh producer means
+// adding a row here.
+var wantMarked6976 = []struct {
+	provenance string
+	why        string
+}{
+	{"INFERRED_FROM_DOTNET_DI_PROVIDER", "constructor parameter type; the class is declared in its own file"},
+	{"INFERRED_FROM_ASPNET_RETURN_TYPE", "ActionResult<T> return annotation mentions T"},
+	{"INFERRED_FROM_ASPNET_FROM_BODY", "[FromBody] parameter annotation mentions the DTO type"},
+	{"INFERRED_FROM_FROM_BODY", "the validation lane's [FromBody] rule, same shape"},
+	{"INFERRED_FROM_BLAZOR_COMPONENT_REF", "a <Foo> markup tag is a USE of the component declared in Foo.razor"},
+	{"INFERRED_FROM_BLAZOR_INJECT", "@inject IFoo names a service declared elsewhere"},
+}
+
+// wantUnmarked6976 is the negative control, and it is what stops the fix from
+// being implemented as "mark the whole C# custom lane". Each of these
+// producers reads a DECLARATION: it is entitled to the repository-wide byName
+// slot and marking it would cost its own resolution.
+var wantUnmarked6976 = []string{
+	"INFERRED_FROM_BLAZOR_PAGE",        // @page "/counter" declares the route
+	"INFERRED_FROM_BLAZOR_CODE_METHOD", // a method DECLARATION in an @code block
+	"INFERRED_FROM_BLAZOR_PARAMETER",   // a [Parameter] property declaration
+}
+
+// csharpFixtures6976 are the sources that make every producer above fire. One
+// map, consumed by both directions, so the marked and unmarked verdicts are
+// read off the SAME extraction run and cannot drift apart.
+func csharpFixtures6976() map[string]extreg.FileInput {
+	return map[string]extreg.FileInput{
+		"custom_csharp_dotnet_di": {
+			Path: "src/Controllers/HomeController.cs", Language: "csharp",
+			Content: []byte(`
+public class HomeController {
+    public HomeController(ActionContext context, IWrapperProvider wrapper) { }
+}
+`),
+		},
+		"custom_csharp_aspnet_reqresp": {
+			Path: "src/Controllers/OrderController.cs", Language: "csharp",
+			Content: []byte(`
+[ApiController]
+public class OrderController : ControllerBase {
+    [HttpPost]
+    public ActionResult<OrderResponse> Create([FromBody] OrderRequest body) { return null; }
+}
+`),
+		},
+		"custom_csharp_validation": {
+			Path: "src/Controllers/PayController.cs", Language: "csharp",
+			Content: []byte(`
+public class PayController : ControllerBase {
+    public IActionResult Pay([FromBody] PaymentRequest body) { return null; }
+}
+`),
+		},
+		"custom_csharp_blazor": {
+			Path: "src/Pages/Counter.razor", Language: "csharp",
+			Content: []byte(`@page "/counter"
+@inject IWeatherService Weather
+
+<WeatherTable Rows="@rows" />
+
+@code {
+    [Parameter] public int Start { get; set; }
+
+    private void IncrementCount() { }
+}
+`),
+		},
+	}
+}
+
+// extractByProvenance6976 runs every fixture through its extractor and returns
+// provenance -> whether the emitted record carries the reference marker, plus
+// the entity name, so a row that never fired is distinguishable from a row
+// that fired unmarked.
+func extractByProvenance6976(t *testing.T) map[string]struct {
+	marked bool
+	name   string
+} {
+	t.Helper()
+	got := map[string]struct {
+		marked bool
+		name   string
+	}{}
+	for key, in := range csharpFixtures6976() {
+		e, ok := extreg.Get(key)
+		if !ok {
+			t.Fatalf("%s not registered", key)
+		}
+		recs, err := e.Extract(context.Background(), in)
+		if err != nil {
+			t.Fatalf("%s extract: %v", key, err)
+		}
+		for i := range recs {
+			collect6976(got, &recs[i])
+		}
+	}
+	return got
+}
+
+func collect6976(got map[string]struct {
+	marked bool
+	name   string
+}, r *types.EntityRecord) {
+	p := r.Properties["provenance"]
+	if p == "" {
+		return
+	}
+	// A provenance that fires more than once must agree with itself; OR-ing
+	// would let one marked record vouch for an unmarked sibling.
+	if prev, seen := got[p]; seen && prev.marked != (r.Properties[resolve.ReferenceShapedProp] == "true") {
+		got[p] = struct {
+			marked bool
+			name   string
+		}{false, prev.name + "|DISAGREES:" + r.Name}
+		return
+	}
+	got[p] = struct {
+		marked bool
+		name   string
+	}{r.Properties[resolve.ReferenceShapedProp] == "true", r.Name}
+}
+
+func TestReferenceShapedProducersAreMarked6976(t *testing.T) {
+	got := extractByProvenance6976(t)
+	for _, w := range wantMarked6976 {
+		t.Run(w.provenance, func(t *testing.T) {
+			v, fired := got[w.provenance]
+			if !fired {
+				t.Fatalf("fixture is vacuous: %s emitted no record, so this row asserts "+
+					"nothing. Fix the fixture, not the expectation.", w.provenance)
+			}
+			if !v.marked {
+				t.Errorf("%s emitted %q WITHOUT %s=true — %s. Unmarked, a mention of a name "+
+					"still evicts the sole declaration of that name from the repository-wide "+
+					"index and takes every bare-name edge to it down with it (#6976).",
+					w.provenance, v.name, resolve.ReferenceShapedProp, w.why)
+			}
+		})
+	}
+}
+
+func TestDeclarationShapedProducersAreNotMarked6976(t *testing.T) {
+	got := extractByProvenance6976(t)
+	for _, p := range wantUnmarked6976 {
+		t.Run(p, func(t *testing.T) {
+			v, fired := got[p]
+			if !fired {
+				t.Fatalf("fixture is vacuous: %s emitted no record, so this control asserts "+
+					"nothing", p)
+			}
+			if v.marked {
+				t.Errorf("%s emitted %q WITH %s=true, but it reads a DECLARATION. Marking it "+
+					"costs it the repository-wide slot it is entitled to; #6976 is scoped to "+
+					"records minted from a MENTION, not to the C# custom lane.",
+					p, v.name, resolve.ReferenceShapedProp)
+			}
+		})
+	}
+}
+
+// TestMarkedProducersAreExactlyTheMeasuredSix6976 closes the scope in the
+// other direction: nothing in the fixtures above may acquire the marker
+// without a row in wantMarked6976. Without this, a future "stamp it in
+// makeEntity" shortcut would pass both tests above.
+func TestMarkedProducersAreExactlyTheMeasuredSix6976(t *testing.T) {
+	want := map[string]bool{}
+	for _, w := range wantMarked6976 {
+		want[w.provenance] = true
+	}
+	for p, v := range extractByProvenance6976(t) {
+		if v.marked && !want[p] {
+			t.Errorf("%s emitted %q with %s=true but is not one of the six measured "+
+				"producers #6976 scopes the marker to. Either it is genuinely "+
+				"reference-shaped — then add a row to wantMarked6976 with the measurement "+
+				"that justifies it — or the stamp leaked.", p, v.name, resolve.ReferenceShapedProp)
+		}
+	}
+}
--- a/internal/custom/csharp/reference_shaped_marker_6976_test.go
+++ b/internal/custom/csharp/reference_shaped_marker_6976_test.go
@@ -43,9 +43,12 @@ import (
 //	        and so does every row here, because the rows read the constant
 //	        through resolve.ReferenceShapedProp rather than a local copy.
 
-// wantMarked6976 is the closed scope of #6976: the SIX measured producers in
-// the `aspnetcore-mvc` eviction population that mint an entity from a MENTION
-// of a name declared elsewhere. Written out by hand.
+// wantMarked6976 is the closed scope of #6976: six producers that mint an
+// entity from a MENTION of a name declared elsewhere. FIVE were measured in
+// the `aspnetcore-mvc` eviction population; the sixth, BLAZOR_INJECT, is
+// classified by SHAPE and is not in that population at all — #6976 names it
+// nowhere. Saying "six measured" would assert a measurement that was never
+// taken. Written out by hand.
 //
 // Deliberately NOT a repo-wide taxonomy of the ~340 custom producers — see the
 // scope note on resolve.ReferenceShapedProp. Adding a seventh producer means
--- a/internal/custom/csharp/validation.go
+++ b/internal/custom/csharp/validation.go
@@ -288,6 +288,9 @@ func (e *csharpValidationExtractor) Extract(ctx context.Context, file extractor.
 			"validation_framework", "FromBody",
 			"provenance", "INFERRED_FROM_FROM_BODY",
 		)
+		// #6976 — same shape as the aspnet [FromBody] rule: a parameter type
+		// annotation naming a DTO declared in another file.
+		markReferenceShaped(&dtoEnt)
 		add(dtoEnt)
 	}
 
--- a/internal/resolve/reference_decl_tier_6976_test.go
+++ b/internal/resolve/reference_decl_tier_6976_test.go
@@ -1,0 +1,459 @@
+// Package resolve — #6976: a record minted from a MENTION of a name evicted
+// the sole real DECLARATION of that name from the repository-wide byName
+// index, and took every bare-name edge to it down with it.
+//
+// THE SHAPE, from `aspnetcore-mvc` with the custom-extractor gate ON. The
+// .NET DI extractor reads a constructor parameter `ActionContext context` in
+// HomeController.cs and mints a SCOPE.Class named `ActionContext` in THAT
+// file. The class `ActionContext` is declared in ActionContext.cs and
+// extracted there. Two entities, one name, different files — so indexByName
+// took its default arm: `delete(byName, "ActionContext")` + ambiguous, for the
+// whole repository. Every bare-name TESTS edge naming `ActionContext` then
+// dangled, in files that had never heard of HomeController.
+//
+// WHY THE ASSERTIONS ARE ON EDGES AND NOT ON COUNTS. This defect survived a
+// measurement that reported explicit TESTS at +8 net while 7,221 edges were
+// SUBSTITUTED (#6973): a cardinality assertion cannot see a substitution. So
+// every case below names the id the edge must land on.
+package resolve_test
+
+import (
+	"testing"
+
+	"github.com/cajasmota/grafel/internal/resolve"
+	"github.com/cajasmota/grafel/internal/types"
+)
+
+const (
+	declFile6976 = "src/Mvc/ActionContext.cs"
+	ctrlFile6976 = "src/Mvc/Controllers/HomeController.cs"
+	apiFile6976  = "src/Mvc/Api/ApiController.cs"
+	testFile6976 = "test/Mvc.Test/ActionContextTest.cs"
+
+	// Stand-ins for the production sha256 ids — distinct by construction so
+	// no case can pass by ID collision.
+	declID6976     = "d000000000000001" // the real `class ActionContext`
+	ctrlRefID6976  = "d000000000000002" // DI provider mention in HomeController.cs
+	apiRefID6976   = "d000000000000003" // return-type mention in ApiController.cs
+	otherDecl6976  = "d000000000000004" // a SECOND real declaration of the name
+	placeholder976 = "d000000000000005" // a per-import placeholder for the name
+	localBind6976  = "d000000000000006" // a function-body local named the same
+	testerID6976   = "d000000000000009" // the test entity carrying the edge
+
+	sharedName6976 = "ActionContext"
+)
+
+// declRecord6976 is the real `public class ActionContext` — unmarked, the
+// thing every consumer of the name wants.
+func declRecord6976(id, file string) types.EntityRecord {
+	return types.EntityRecord{
+		ID: id, Kind: "SCOPE.Class", Name: sharedName6976,
+		SourceFile: file, Language: "csharp", StartLine: 12, EndLine: 40,
+	}
+}
+
+// refRecord6976 is a reference-shaped record: the entity a custom extractor
+// mints when it reads a MENTION of the name. `reference_shaped` is stamped at
+// the emit site (internal/custom/csharp) — see resolve.ReferenceShapedProp.
+func refRecord6976(id, file, provenance string) types.EntityRecord {
+	return types.EntityRecord{
+		ID: id, Kind: "SCOPE.Class", Name: sharedName6976,
+		SourceFile: file, Language: "csharp", StartLine: 7, EndLine: 7,
+		Properties: map[string]string{
+			"framework":                  "dotnet_di",
+			"provenance":                 provenance,
+			resolve.ReferenceShapedProp:  "true",
+			"grafel_fixture_marker_only": "6976",
+		},
+	}
+}
+
+// testerRecord6976 carries the bare-name TESTS edge the mechanism destroyed.
+func testerRecord6976() types.EntityRecord {
+	return types.EntityRecord{
+		ID: testerID6976, Kind: "SCOPE.Pattern", Name: "ActionContextTest",
+		Subtype: "unit", SourceFile: testFile6976, Language: "csharp",
+		Relationships: []types.RelationshipRecord{
+			{FromID: testerID6976, ToID: sharedName6976, Kind: "TESTS"},
+		},
+	}
+}
+
+// resolveTESTS6976 runs the production index + rewrite over recs and returns
+// the ToID the single TESTS edge ended up on. It fails the test if the fixture
+// stops carrying exactly one such edge, so no case can pass vacuously.
+func resolveTESTS6976(t *testing.T, recs []types.EntityRecord) string {
+	t.Helper()
+	cp := append([]types.EntityRecord(nil), recs...)
+	for i := range cp {
+		cp[i].Relationships = append([]types.RelationshipRecord(nil), cp[i].Relationships...)
+	}
+	idx := resolve.BuildIndex(cp)
+	resolve.ReferencesEmbedded(cp, idx)
+
+	got, seen := "", 0
+	for i := range cp {
+		for _, r := range cp[i].Relationships {
+			if r.Kind == "TESTS" {
+				seen++
+				got = r.ToID
+			}
+		}
+	}
+	if seen != 1 {
+		t.Fatalf("fixture is vacuous: found %d TESTS edge(s), want exactly 1", seen)
+	}
+	return got
+}
+
+// bothOrders6976 runs one case in both extraction orders. Extraction order is
+// not stable across runs, so a tier that only works in one direction is a bug
+// that reports itself as a flake.
+func bothOrders6976(t *testing.T, a, b types.EntityRecord, check func(t *testing.T, got string)) {
+	t.Helper()
+	tester := testerRecord6976()
+	for _, o := range []struct {
+		tag  string
+		recs []types.EntityRecord
+	}{
+		{"a-then-b", []types.EntityRecord{a, b, tester}},
+		{"b-then-a", []types.EntityRecord{b, a, tester}},
+	} {
+		t.Run(o.tag, func(t *testing.T) {
+			check(t, resolveTESTS6976(t, o.recs))
+		})
+	}
+}
+
+// TestDeclarationOutranksReference_6976 is the reported defect. The bare-name
+// TESTS edge must land on the DECLARATION's id, in either extraction order.
+func TestDeclarationOutranksReference_6976(t *testing.T) {
+	bothOrders6976(t,
+		declRecord6976(declID6976, declFile6976),
+		refRecord6976(ctrlRefID6976, ctrlFile6976, "INFERRED_FROM_DOTNET_DI_PROVIDER"),
+		func(t *testing.T, got string) {
+			if got != declID6976 {
+				t.Errorf("TESTS -> %q, want the declaration %q (%s): a record minted from a "+
+					"constructor-parameter MENTION of %q evicted the sole declaration of that "+
+					"name from the repository-wide index, leaving the edge unresolvable (#6976)",
+					got, declID6976, declFile6976, sharedName6976)
+			}
+		})
+}
+
+// TestTwoDeclarationsStillCollide_6976 is the OTHER direction, and it is the
+// one that matters more: the tier must not start resolving names that are
+// GENUINELY ambiguous between two real declarations. A confidently-wrong bind
+// reads as valid and is worse than a dangling reference (#6369).
+func TestTwoDeclarationsStillCollide_6976(t *testing.T) {
+	bothOrders6976(t,
+		declRecord6976(declID6976, declFile6976),
+		declRecord6976(otherDecl6976, apiFile6976),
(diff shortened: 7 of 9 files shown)


cajasmota:
## Coordinator mutant MD-1 on `0a5a57f68` — DEAD

Own detached worktree, `git status --porcelain` empty at start and after restore, restored by `cp` from a SHA-256-verified backup (`0e37ebce…` both ways), zero `MUTANT` text remaining.

**MD-1 — drop the marker from ONE producer only.** I removed the single `markReferenceShaped(&ent)` call at `internal/custom/csharp/dotnet_di.go:192` (`DOTNET_DI_PROVIDER`), replacing it with an inert `_ = &ent`, and left the other five stamps intact. Edit proven landed: the hit set was exactly `[191]` (0-based), and zero `markReferenceShaped` calls remained in that file afterwards.

**Result: DEAD** — `TestReferenceShapedProducersAreMarked6976` failed; `./internal/resolve/` stayed green.

### Why I chose this one, and what it settles

This is the mutant that speaks to the concern I opened the review with: **a replay is a reconstruction, and its failure mode is a silent partial one.** An edit that never got replayed leaves one producer unmarked, which looks identical to a producer deliberately excluded. MD-1 *is* that failure, injected deliberately — and the suite catches it. So each of the six stamps is graded **individually**, not collectively, and the reconstruction risk on this axis is closed by evidence rather than by the disclosure being credible.

I also checked the trap that shipped in #6975: `TestReferenceShapedProducersAreMarked6976` ranges over `wantMarked6976`, a `[]struct{...}` **independent literal** of six entries, with `wantUnmarked6976` as a negative control. It is not derived from the production collection, so a deletion cannot hide in it. That is the correct shape and it is why MD-1 died rather than passing quietly.

One thing MD-1 does **not** establish: it grades that each *listed* producer is marked. It cannot tell you the list is the right six — that judgement is the `BLAZOR_CODE_METHOD`-vs-`BLAZOR_INJECT` question, which is the reviewer's item 4 and not something a mutant on this file can reach.

cajasmota:
## Independent adversarial review — **REQUEST-CHANGES** (two narrow items; the fix itself is sound and every headline number reproduced)

**Tree I measured in (absolute):** `/private/tmp/claude-501/-Users-jorgecajas-Projects-archigraph/c0fefed4-0195-4c81-9a94-07a7145710a4/scratchpad/review-6980-qz9m/wt` @ `0a5a57f68`, parent arm in `.../review-6980-qz9m/parent` @ `448d7bb38`. Corpus copied to `.../review-6980-qz9m/corpus/aspnetcore-mvc`; `/Users/jorgecajas/Projects/archigraph-corpora` untouched. In-process `Indexer.Run` via `runIndexerOn`, one arm per process, `GOMAXPROCS=3`, `nice -n 10`, isolated `GRAFEL_HOME`/`GRAFEL_DAEMON_ROOT`. macOS/APFS, Apple Silicon. **The gate is stated on every figure.**

Worktree left byte-identical to `0a5a57f68` (`git status --porcelain` empty; `refs.go` sha256 `c91d1972…d930c` restored from a shasum-verified `cp` backup after every mutant — no `git checkout` was used).

---

### (a) All six marked sites survived the reconstruction — **YES**, and they are the six claimed

`grep -rn markReferenceShaped internal/ cmd/` → exactly **6 call sites + 1 definition**, no more, no fewer. Named, each read against its surrounding emit block (not just the call line):

| # | site | provenance set 3 lines above the stamp |
|---|---|---|
| 1 | `internal/custom/csharp/dotnet_di.go:192` | `INFERRED_FROM_DOTNET_DI_PROVIDER` |
| 2 | `internal/custom/csharp/validation.go:293` | `INFERRED_FROM_FROM_BODY` |
| 3 | `internal/custom/csharp/blazor.go:300` | `INFERRED_FROM_BLAZOR_INJECT` |
| 4 | `internal/custom/csharp/blazor.go:334` | `INFERRED_FROM_BLAZOR_COMPONENT_REF` |
| 5 | `internal/custom/csharp/aspnet_request_response.go:83` | `INFERRED_FROM_ASPNET_FROM_BODY` |
| 6 | `internal/custom/csharp/aspnet_request_response.go:110` | `INFERRED_FROM_ASPNET_RETURN_TYPE` |

No half-applied edit found: the helper (`helpers.go:226/230`) has 6 callers, the reader (`refs.go:1789/1795`) has 3 callers (flat `BuildIndex`, `BuildModuleSymbols`, `buildModuleSymbolsOrderedPos`), **both** index writers were updated (`symbol_index.go:217, 457, 954`), and the `srcFile` argument is `normalizePath`d on both paths (`refs.go:1216`, `symbol_index.go:195/435`) so the cross-file comparison cannot disagree between the flat and module-partitioned builders. No mutant text anywhere in the changed files. `gofmt -l` clean, `go vet` clean on all three packages. `internal/quality/golden/baseline.json` untouched (`git diff HEAD~1 --stat -- internal/quality/` is empty).

### (b) Gate-OFF byte-identical — **YES**, and it is provable by construction, which is stronger

Reproduced empirically on `aspnetcore-mvc`, gate OFF (`GRAFEL_INPROC_CUSTOM_EXTRACTORS` unset), branch vs parent:

```
branch_OFF.tsv  211,162 rows   entities 39,766
parent_OFF.tsv  211,162 rows   entities 39,766
cmp branch_OFF.tsv parent_OFF.tsv           -> exit 0
cmp branch_OFF.tsv.entities parent_OFF.tsv.entities -> exit 0
```

(39,766 also reproduces the issue's OFF entity count exactly.) My row count is lower than the PR's 250,954 because I diffed one repo, not the full corpus — but the by-construction argument covers every repo: `reference_shaped` has **exactly one writer in the tree** (`markReferenceShaped`, in `internal/custom/csharp`), so with the gate off `isRef` is universally false and `nameHolderRef`/`nameAmbigRef` are never populated. Walking each new clause under those constants: the `ambigName` guard collapses to the pre-#6976 expression, the two `!idx.nameHolderRef[name]`/`!isRef` conjuncts are identities, both new `case` arms are unreachable, the `default`-arm `if isRef` is dead, and the trailing switch takes `case !isRef: delete(…)` on an always-empty map. `nameHolderFile` is written but read only from the two dead arms. **Inert, not merely measured-inert.**

### (c) Cross-file scoping — **principled, but the scoping is incomplete** (finding 1)

The principle is right and it is #6976's own argument, not a post-hoc patch: #6104's facet rule already owns same-file same-name pairs as "two views of one construct", so the declaration/reference distinction has no business ranking them. The `csharp-aspnet-core-mini` regression is a correct diagnosis of a real hazard.

**But the scoping was applied to two of the three places that can now change a same-file outcome.** The `ambigName` reclaim branch (`refs.go:1885-1892`) has no `nameHolderFile` guard, so reference-only ambiguity raised by two **same-file** references can be reclaimed by a **same-file** unmarked record. Probed with three claimants all in `src/Controllers/OneFile.cs` (2 marked refs + 1 unmarked declaration, distinct kinds → distinct ids), production `BuildIndex` + `ReferencesEmbedded`, single `TESTS` edge, vacuity-checked:

| extraction order | parent `448d7bb38` | this branch |
|---|---|---|
| `ref, ref, decl` | `"ActionContext"` (unresolved) | **`d0000000000000c3`** (the declaration) |
| `decl, ref, ref` | `"ActionContext"` | `"ActionContext"` |
| `ref, decl, ref` | `"ActionContext"` | `"ActionContext"` |

Two consequences. First, `TestSameFilePairingIsUnchanged_6976` asserts more than it covers — same-file **pairings** are unchanged, same-file **sets of three** are not, and the test's name and the PR body both make the wider claim. Second, and worse, the branch turns an order-**independent** same-file outcome into an order-**dependent** one, in an area where this PR's own body says *"extraction order is not stable, so both directions are needed"*. The winning record in the firing order is the **unmarked** one — which is exactly the mechanism that produced the `IUserService` 13→12 ratchet failure; that fixture only escapes because it has two claimants rather than three.

Cheap to close: add `&& idx.nameHolderFile[name] != srcFile` to the reclaim condition in the `ambigName` branch (or scope `nameAmbigRef` to cross-file at the point it is raised), and extend the same-file test with the 3-claimant case in all three orders. I priced it at ~4 production lines + ~20 test lines — the same price #6980 already paid for R11/R12.

### (d) Reverse direction adds no confident bindings — **true as stated, but it is not the claim the PR leans on** (finding 2)

Reproduced exactly, and stronger than reported: the two sets are not merely equal in size, they are **byte-identical** (`cmp` exit 0).

```
gate ON, TESTS edges with a hex target present in the arm but NOT in OFF:
  parent ON  1,777
  branch TIER 1,777     cmp parent_ON.newconf branch_TIER.newconf -> exit 0
```

Every other headline figure reproduced to the row, gate ON except the OFF column:

| | OFF | ON (parent) | TIER (branch) |
|---|---|---|---|
| `TESTS` `(from,to,source)` triples | 43,620 | 40,887 | 44,192 |
| lost vs OFF | — | 6,353 | 1,227 |
| gained vs OFF | — | 3,620 | 1,799 |
| entities | 39,766 | 50,260 | 50,260 |

**5,126** of the edges lost in ON come back in TIER, and **0 of their 132 distinct target ids are missing from the TIER entity set**. TIER exceeding OFF is fully explained and is *not* over-binding: TIER's 1,799 gains vs OFF are 1,777 hex + 22 unresolved, and those 1,777 are the **same set** the parent's ON arm already produced. The tier contributes **zero** new confident targets vs OFF; the surplus over OFF is the gate's own entity gain, which the tier neither adds to nor redirects.

**The problem is the claim built on top of it.** The PR states *"no relationship kind regresses against ON; the only three that move are `TESTS` +3,305, `STEP_IN_PROCESS` +7, `ENTRY_POINT_OF` +4."* That is a **cardinality** reading, and the set diff ON→TIER says otherwise — this is #6973's substitution blindness, in this PR, in the direction its own body warns about:

```
ON -> TIER, whole graph:   only_on 5,636 rows   only_tier 8,952 rows
kinds with EQUAL in/out counts (pure substitution, invisible to a per-kind count):
  CALLS 2,989   USES 257   IMPLEMENTS 133   RETURNS 108   INJECTED_INTO 70   CONTAINS 55   EXTENDS 35
```

Classified by whether the target was resolved before and after:

* **UNRES→HEX 5,104** (CALLS 2,683, TESTS 1,821, USES 257, RETURNS 108, IMPLEMENTS 94, INJECTED_INTO 66, EXTENDS 32, …) — the intended win, and it reaches seven kinds the PR does not mention.
* **HEX→HEX 271** — a *confident* target that changed to a *different* confident target: CALLS 225, IMPLEMENTS 39, EXTENDS 3, INJECTED_INTO 4. This is the #6369 hazard class, and the PR's reverse-direction check cannot see it by construction: that check looks only at `TESTS`, and only at edges **absent from OFF**, whereas every one of these is present in all three arms.

Graded against the OFF arm as the closest thing to ground truth here:

| | CALLS | IMPLEMENTS | EXTENDS | INJECTED_INTO |
|---|---|---|---|---|
| TIER now agrees with OFF (a correction) | 46 | 36 | 3 | — |
| OFF fell to `ext:` synthesis, TIER picks a local entity | 119 | 0 | 0 | — |
| OFF picked a **third** hex entity — uncorroborated | 54 | 2 | 0 | — |
| **OFF agreed with ON and TIER moved away** | **0** | **0** | **0** | **0** |

So the direction is genuinely good — **not one** substitution moves an edge off a target the no-custom-lane arm endorsed — and I would not block on the substitutions themselves. What I am asking for is that the body stop asserting a per-kind cardinality result the set diff contradicts: **~3,647 non-`TESTS` edges change target**, 271 of them confidently, and none of that is currently graded or even disclosed. Replacing the "only three kinds move" sentence with the UNRES→HEX / HEX→HEX split above (and a line on the 56 uncorroborated third-target picks) would make the PR's evidence match its own stated standard.

### Claim 3, claim 4, R11/R12

* **Golden ratchet reproduced independently**: `scripts/quality/run.sh --ratchet` → exit **0**, *"38 gated fixtures held their baseline (29 at 100% must-have recall)"*. `go test -count=1` exit 0 on `./internal/resolve/`, `./internal/custom/csharp/`, and `./cmd/grafel/` (144s — correctly run, since the digest pin lives there).
* **`BLAZOR_CODE_METHOD` → negative control: correct.** `reBlazorCodeMethod` matches `private void IncrementCount()` — a method **declaration** — so marking it would be a false label independent of whether its population is dead after #6975. The substitution is right on the merits, not on the cost. One wording nit: `wantMarked6976`'s doc calls all six *"the SIX **measured** producers in the `aspnetcore-mvc` eviction population"*, but `BLAZOR_INJECT` was justified by **shape** ("same reference shape as `DOTNET_DI_PROVIDER`"), not by a measurement in that population — #6976 named it nowhere. Say "five measured plus one classified by shape", or cite its count.
* **R11/R12 kills verified** by re-deriving them from the code: `setNameHolderFile` is called on **every** `byName` write (5 sites) and `nameHolderFile` is deleted on both ambiguity paths, so no arm can read a stale holder file. I found the third stale-holder case they imply — the tail insert — already scored as R13.

### My mutants — 4 scored, more-permissive direction first, none of them the author's

Each applied with an asserted occurrence count (`grep -c` on a unique anchor, refused if ≠1), landing proved by re-grepping the mutated line, scored with `-count=1`, restored by `cp` from a sha256-verified backup and re-diffed. Green baseline immediately before each.

| # | mutant | verdict |
|---|---|---|
| **V1** | reader accepts any truthy value: `props[ReferenceShapedProp] == "true"` → `!= ""` (`refs.go:1795`, 1 occurrence) | **ALIVE, not production-reachable.** The only writer of the key is `markReferenceShaped`, which writes the literal `"true"`. I probed the masking pair — V1 **plus** P3 (`setProps(e, referenceShapedProp, "yes")`) — and the compound is **DEAD**: `collect6976` in the producer test compares against `"true"` directly rather than going through the reader, so the value contract is pinned independently on the producer side. Reader-side strictness is nonetheless ungraded; one table row in `reference_decl_tier_6976_test.go` feeding `reference_shaped:"false"` and asserting declaration treatment closes it (~5 lines). P3 nit, not a blocker. |
| **V2** | mark a **declaration-shaped** producer: insert `markReferenceShaped(&ent)` on the `@page` emit site (`blazor.go`, anchor `"route_path", route)`, 1 occurrence; call count 2→3) | **DEAD** — `TestDeclarationShapedProducersAreNotMarked6976/INFERRED_FROM_BLAZOR_PAGE` **and** `TestMarkedProducersAreExactlyTheMeasuredSix6976`, both by name. The single-site variant of P2 is caught as cleanly as the whole-lane one. |
| **V3** | ref arm drops `!idx.nameHolderLocal[name]` — a reference no longer collides with a **function-local** incumbent (not in the PR's R-table) | **DEAD** — `TestReferenceDoesNotOutrankLocalBinding_6976/b-then-a`. The test exists and is order-covered even though the mutant table does not list it. |
| **V4** | ref arm drops `!idx.nameHolderImport[name]` — a reference no longer collides with a **placeholder** incumbent (the mirror of R5, which mutates the *import* arm) | **DEAD** — `TestReferenceDoesNotOutrankImportPlaceholder_6976/b-then-a`. |

Both compound guards in the new `case isRef` arm are therefore graded part by part, in both orders — no mutually-masking pair. The one gap my mutants exposed is not in the tier's conjuncts at all: it is the missing cross-file guard on the `ambigName` reclaim path, which no mutant can reach because there is no test asserting that behaviour to kill (finding 1 — *distinguishable and ungraded*, assertion named, grading priced at ~24 lines).

---

### Summary

The upstream/downstream split is the right design, the reconstruction is complete and verifiable, the safety property holds by construction as well as by measurement, the tier is well-tested in the direction it was tested, and every number in the body reproduced on an independent tree. Two things to fix before merge:

1. **Close (or pin) the same-file 3-claimant reclaim** — it introduces order-dependent binding in exactly the shape the cross-file scoping exists to prevent, and the current same-file test's name over-claims.
2. **Correct the "only three kinds move" sentence** with the set-level ON→TIER numbers — ~3,647 non-`TESTS` edges are substituted, 271 of them hex→hex, and the PR's own standard (#6973) is that a per-kind count cannot see that.

Neither reaches a user today (#6966: the gate is off, and (b) is proven inert), so this is a hold on evidence quality and on one latent hole, not on blast radius. Happy to re-review on push. Did not merge; did not push to the branch.

cajasmota:
## Confirmation pass — **APPROVE** (`e50c8653e`)

**Tree I measured in (absolute):** `/private/tmp/claude-501/-Users-jorgecajas-Projects-archigraph/c0fefed4-0195-4c81-9a94-07a7145710a4/scratchpad/review-6980-conf-r4x8/wt` @ `e50c8653e`, with a second checkout of the **pre-fix** `0a5a57f68` at `.../review-6980-conf-r4x8/prefix` used only to test my own suggestion. Corpus copied to `.../review-6980-conf-r4x8/corpus/aspnetcore-mvc`; `archigraph-corpora` untouched. `GOMAXPROCS=3`, `nice -n 10`, isolated `GRAFEL_HOME`/`GRAFEL_DAEMON_ROOT`. macOS/APFS. Both worktrees left byte-identical to their commits (`git status --porcelain` empty; `refs.go` sha256 `326b6412…938f7` restored by `cp` from a shasum-verified backup after each mutant — no `git checkout` used).

### Your pushback is right, and I confirmed it by running my own suggestion

Not accepted on the argument — reproduced. `ambigName[name] = true` has **exactly one** production write site (`refs.go:1993`), in the `default` collision arm, and that same arm deletes `nameHolderFile` before returning (`refs.go:2036`). So by the time the reclaim branch at `refs.go:1888` runs, `nameHolderFile[name]` is always `""`.

I applied my suggested guard to the **pre-fix** tree (`0a5a57f68`, one occurrence asserted, landing re-grepped) and ran my exhibit:

```
+ if isImport || isLocal || isRef ||
+     idx.nameHolderFile[name] == srcFile ||          <- my suggestion
      !(idx.nameAmbigImport[name] || idx.nameAmbigRef[name]) {

PREFIX+MYGUARD  order ref,ref,decl -> "d0000000000000c3"   (still binds — guard never fires)
```

**My guard would have been a no-op.** Raising `nameAmbigRef` only for a cross-file pair is the correct site, and it is the only site that still has the information. Thank you for checking it instead of implementing it.

### The fix closes my exhibit — verified with my own records, not yours

Your `TestSameFileThreeClaimantsStayAmbiguous_6976` uses the `Dependency` shape. I re-ran **my original exhibit verbatim** (2 marked refs + one unmarked `declRecord6976`, distinct kinds → distinct ids, all three in one file) through `allOrders6976` on `e50c8653e`:

```
all 6 orders -> "ActionContext"  (unresolved, matching parent 448d7bb38 in all orders)
```

Order-dependence is gone, in the harder direction (an unmarked *declaration*, not just an unmarked mention).

**The enumeration is real, not a comment.** `allOrders6976` increments `seen` per leaf and `t.Fatalf`s unless `seen == n!` computed in code; sub-test names are the concatenated ids, so no permutation can be silently merged by a name collision. Observed: 6 / 6 / 2 sub-tests for the three tests, 48 passing `6976` sub-tests in total. **The control genuinely lands on the declaration** — `TestCrossFileThreeClaimantsStillReclaim_6976` asserts `got == declID6976` (not merely "not ambiguous") and passes in all six orders, so the narrowing did not disarm the reclaim.

### Two mutants of my own on the new conjunct — neither is your M1 or M2

| # | mutant (occurrence asserted = 1, landing re-grepped, `-count=1`, `cp`-restored) | verdict |
|---|---|---|
| **C1** | **Hoist `delete(idx.nameHolderFile, name)` above the `nameAmbigRef` raise** and drop it from the arm's tail — a plausible "group the deletes together" refactor. The conjunct then reads `"" != srcFile`, always true, silently restoring pre-fix behaviour. Verified landed: delete count 2, sited at `refs.go:2002`, immediately above the guard; `gofmt` clean. | **DEAD** — `TestSameFileThreeClaimantsStayAmbiguous_6976`, failing **exactly the two `ref,ref,decl` orders** (`002,003,004` and `003,002,004`), the same signature as your M1. So the guard's **position relative to the delete** is graded, not just its presence — which matters, because that ordering is the whole reason the guard works here and not at the reclaim site. |
| **C2** | `setNameHolderFile` becomes write-once (`if _, ok := …; ok { return }`) — a stale holder file feeding the new conjunct, the R11/R12 hazard aimed at the fix's input rather than at the fix. | **DEAD** — `TestHolderFileFollowsTheSlot_6976` **and** `TestHolderFileFollowsTheSlotAfterPlaceholderDisplacement_6976`. |

### Finding 2 — the two partitions are consistent, and **mine was the wrong one**

You asked for a second opinion on the divergence. I re-derived the partition from **my own dumps** with the one thing my first pass got wrong — a `(from, kind, source)` key can carry **more than one target**, and my `awk` hash kept only the last one. Redone with target **sets**:

```
only_on 5,636   only_tier 8,952               (identical to yours)
UNRES->HEX 5,050    HEX->HEX 406              (identical to yours)
HEX->HEX keys graded vs OFF:
  TIER agrees with OFF        284 keys  (= your 402 edges)
  absent from OFF               4 keys  (= your 4 ungradeable)
  OFF agreed with ON, TIER moved away    0
  OFF picked a THIRD real entity         0
  OFF fell to ext: synthesis             0
```

So the divergence is **not** two measurements of different things — it is my first pass measuring something other than what it said. Once multiplicity is handled, my numbers *are* yours: 406 edges across 288 keys, which is exactly your `402 corrections + 4 ungradeable` counted at edge level instead of key level. **My "119 `ext:`" and "56 uncorroborated third-target picks" were artefacts of the key collapse** — comparing one arbitrarily-retained target against one arbitrarily-retained OFF target manufactured both buckets. There are no uncorroborated third-target picks. Please treat your figures as the correct ones and mine as superseded; the body already credits both, and the honest version is that the reviewer's split was the defective one.

The claim both agree on — **zero** edges moved off a target the no-custom-lane arm endorsed — survives under either method, and now with a partition that can actually support it.

### Nothing else moved

| check | result |
|---|---|
| gate-**OFF** dump, `e50c8653e` vs parent `448d7bb38` | `cmp` **exit 0**, 211,162 rows, entities 39,766 |
| gate-**ON** TIER dump, `e50c8653e` vs pre-fix `0a5a57f68` | `cmp` **exit 0**, 233,103 rows, entities 50,260 — **the fix costs nothing on the corpus**, row for row |
| `go test -count=1` `./internal/resolve/` `./internal/custom/csharp/` `./cmd/grafel/` | exit 0 (cmd/grafel 149s) |
| `gofmt -l`, `go vet` | clean |
| `scripts/quality/run.sh --ratchet` | exit 0 — **38 gated fixtures held, 29 at 100%** |

(My row counts are one repo, `aspnetcore-mvc`, against your full-corpus 250,928 / 283,371 — same verdict, smaller denominator.)

### Doc fix

`wantMarked6976` and the `ReferenceShapedProp` scope note now say five measured plus `BLAZOR_INJECT` classified by shape. That is the accurate claim.

---

**APPROVE.** The reclaim hole is closed at the only site that could close it, my exhibit is gone in all six orders with the harder record shape, the cross-file reclaim is unaffected, the enumeration is enforced in code rather than asserted in a comment, and two mutants I invented on the new conjunct — including the delete-ordering one that the fix's own reasoning depends on — are DEAD. Gate-OFF remains byte-identical and inert by construction. My remaining first-pass nits (reader-side `== "true"` ungraded; `TestMarkedProducersAreExactlyTheMeasuredSix6976` closing only over provenances the four fixtures fire) are P3 and do not block. Did not merge; did not push to the branch.

cajasmota:
## Coordinator mutant CE-1 on `e50c8653e` — DEAD

My earlier MD-1 was scored on `0a5a57f68`; this round added a production conjunct, so a verdict on the old head does not carry. Own detached worktree, `git status --porcelain` empty at start and after restore, restored by `cp` from a SHA-256-verified backup (`326b6412…` both ways), zero `MUTANT` text remaining, baseline green immediately before.

**CE-1 — drop the cross-file scoping from the *reference* arm at `refs.go:1955`.**

Six mutants have now been scored on this PR and **every one of them landed on a different site than this conjunct**: M1/M2 and C1/C2 all attack the `nameAmbigRef` raise at `:2002` (dropping it, inverting it, hoisting the delete above it, staling its input), and V3/V4 attack this arm's *other* two conjuncts — `!nameHolderLocal` and `!nameHolderImport`. The `!= srcFile` at `:1955` was the one left, and it is not an arbitrary choice: **it is the same cross-file scoping the `csharp-aspnet-core-mini` ratchet regression forced**, just at the arm that refuses the displacement rather than the arm that raises ambiguity.

```go
-  case isRef && idx.nameHolderFile[name] != srcFile &&
+  case isRef && // MUTANT CE-1: cross-file scoping dropped from the ref arm
       !idx.nameHolderRef[name] && !idx.nameHolderImport[name] &&
       !idx.nameHolderLocal[name]:
```

Edit proven landed: exactly 1 occurrence, hit set asserted `[1955]`, plus an assertion that the following line is `!idx.nameHolderRef[name]` so the edit could not land on a same-shaped conjunct elsewhere (`:1977` is one). 0 occurrences of the original afterwards, `go vet` clean before scoring.

**Result: DEAD** — `./internal/resolve/` fails three tests, `./internal/custom/csharp/` stays green:

- `TestSameFilePairingIsUnchanged_6976`
- `TestSameFileThreeClaimantsStayAmbiguous_6976`
- `TestSameFilePairingIsEnumerated_6976`

### What I expected, and why being wrong is the point

I picked this conjunct expecting to find it graded **only by the golden ratchet** — the regression that produced it was a fixture-level `relationship_found 13 → 12`, and a scoping justified by a fixture is usually pinned by that fixture and nothing faster. That would have been a real finding: the ratchet is the slow, coarse gate, and a conjunct reachable only through it is one refactor away from silent removal.

It is not what happened. The **fast unit suite** kills it, in under two seconds, and the three tests that fire are exactly the enumerated same-file ones added this round for finding 1. So the enumeration written to close the reviewer's exhibit at `:2002` turns out to grade the cross-file scoping at `:1955` as well — the same-file space is enumerated, so it does not matter which arm breaks the property; any arm that starts binding a same-file pair is caught. **That is the durable result: the enumeration is a property test over the same-file space, not a regression pin on one site**, and it covers a conjunct nobody wrote it for.

Taken with the reviewer's C1/C2 and V3/V4, all four conjuncts of the reference arm and the position, input, presence and polarity of the raise-site guard are now individually graded, with no masking pair found.

**APPROVING and merging on green.** Both review findings are closed, the reviewer verified the author's pushback rather than accepting it — reproducing that their own suggested guard would have been a no-op — and independently withdrew their partition of finding 2 as an artefact of a key collapse, which is the harder half of a review to get right.

PR #6982, opened 2026-09-07
fix(#6969): an oversized sparse pattern file was truncated silently, and truncation was the one input class failing toward EXCLUSION

### LINKED ISSUE #6969: readGitMetaFile truncates silently at its cap, so an oversized sparse pattern file stays sparse with a PARTIAL pattern set — the one input class that fails toward exclusion
Found by a coordinator mutant on #6967 (for #6964). **Not reachable on anything I measured** — filed so the last input class that falls the wrong way is on the record rather than rediscovered.

## The mechanism

`internal/gitmeta` reads git metadata through `readGitMetaFile` → `safeio.ReadFile`, which is:

```go
if maxBytes > 0 {
    r = io.LimitReader(f, maxBytes)
}
return io.ReadAll(r)
```

**A file over the cap is truncated silently — no error, no report.** `reportGitMetaSkip` announces only `ErrNotRegular` / `ErrWouldBlock`, and a truncation is neither. That is correct as written; there is simply nothing to announce.

## Why it matters for the sparse pattern file specifically

After #6967, `ProbeRepo` distinguishes three states, and two of them deliberately fail toward inclusion:

| input | result | direction |
|---|---|---|
| **absent** pattern file | `haveFile=false` → not sparse → full index | over-inclusive (safe) |
| **unreadable** pattern file | `haveFile=false` → not sparse | over-inclusive (safe, stated in the comment) |
| **oversized** pattern file | `haveFile=true` with a **partial pattern set** → still sparse, fewer patterns | **under-inclusive — files silently excluded** |

That last row is the one input class landing on the dangerous side of a change whose whole review was about not silently dropping files. A mid-line cut can also leave a **corrupted final pattern** that matches something unintended rather than nothing, which is worse than dropping the line.

`maxSparsePatternBytes = 8 << 20`.

## Reachability — honestly, low

**I have not demonstrated a repo anywhere near 8 MiB of sparse patterns and I am not claiming one.** A non-cone monorepo with per-file patterns is the only shape I can imagine approaching it. Cone mode writes a handful of lines and cannot get close.

So this is filed as a latent hole, not a live defect. The reason it is worth recording anyway is that its failure mode is **silent exclusion**, which is invisible by construction: no error, no `SkipEntry`, no missing-entity signal — just a smaller graph.

## The fix, if taken

Read `maxBytes+1` and treat "hit the cap" as **not sparse**, matching the over-inclusive direction every other failure in that file already chooses and consistent with the reasoning already written for the unreadable case. A few lines.

**Grade it in the exclusion direction**: a test asserting that an oversized pattern file yields a *full* index, not a partial one. Note that a coordinator mutant shrinking the cap to force truncation is already **DEAD across five tests**, so the truncation path is exercised — what is ungraded is specifically the *decision* about what an over-cap read should mean.

## Wider question this raises

`maxSparsePatternBytes` is one of several caps in `internal/gitmeta/saferead.go`. **The same silent-truncation semantics apply to every file read through `readGitMetaFile`**, and each one deserves the same question: if this file is truncated, is the resulting state a safe superset or a silent subset? For `HEAD`/`commondir` a truncation likely produces an unparseable value that fails loudly; for a *list*-shaped file like `packed-refs` it would produce a silently short list, which is the same shape as this issue. Worth one pass over the callers rather than fixing only the sparse one.

Refs #6964, #6967.

cajasmota:
Closes #6969.

`safeio.ReadFile` is `io.LimitReader` + `io.ReadAll`: **a file over the cap comes back truncated with a nil error**. For every other file `internal/gitmeta` reads that is loud — HEAD, `commondir`, the `.git` gitdir-file and a loose ref are each one short line, every parser here demands a shape, and a 64 KiB prefix of something else fails that shape and falls through to the "unknown" the callers already handle.

`info/sparse-checkout` is the one **list**-shaped reader in the package, and a prefix of a list is a valid-looking shorter list. The repo stayed sparse with fewer includes and files left the index with **no error, no `SkipEntry`, no missing-entity signal** — just a smaller graph.

It was also the last of #6967's three input states landing the wrong way:

| input | before | after |
|---|---|---|
| absent pattern file | not sparse → full index | unchanged |
| unreadable pattern file | not sparse → full index | unchanged |
| **oversized pattern file** | **sparse with a partial pattern set → files silently excluded** | **not sparse → full index** |

## The shape, and why

`safeio.ReadFileLimited(path, policy, maxBytes) ([]byte, bool, error)` — a distinct function returning the one bit `ReadFile` throws away.

**Truncation is observed, not inferred.** The read asks for `maxBytes+1`; truncation is the observation that a `(maxBytes+1)`'th byte existed. The obvious alternative — `len(b) == maxBytes` at the call site — is wrong in the direction that matters: a file whose size is *exactly* the cap is complete, and rejecting it would turn a perfectly good sparse repo into a full index while looking identical to correct code on every other input. The extra byte is never handed to a caller; the slice is capped either way. `maxBytes+1` is guarded against overflow at `math.MaxInt64`, where the wrap goes negative and `io.LimitReader` with a negative `n` reads **zero** bytes — the exact silent under-read this function exists to prevent.

`truncated` is false whenever `err != nil` or `maxBytes <= 0`: an unlimited read has no cap to hit, and a failed read has no claim to make about how much of the file it saw.

### Wholesale rejection, not "drop the corrupted final pattern"

Stated explicitly because the two are different safety properties and only one is right. A truncated read carries **two** hazards:

1. Every pattern **past the cut** is gone — arbitrarily many, not one. The surviving prefix is a syntactically perfect shorter list, so the repo stays sparse with fewer includes and the index silently shrinks.
2. The **cut line itself** is a pattern git never wrote. `!/src/sub/` severed at the cap is `!/src` — not a pattern that fails to match, a pattern that matches **wider** than intended.

Trimming the final line addresses hazard 2 and leaves hazard 1 entirely — and hazard 1 is the one that drops files. Rejecting the file lands both on "not sparse", i.e. a **full index**: a superset, the same over-inclusive direction the absent and unreadable cases already chose. Graded as **M4** below, which is that rejected design implemented and run.

The rejection is **reported** (`reportGitMetaTruncation`, sharing `reportGitMetaSkip`'s dedup map and 8-line cap). "Not sparse" on a repo the user made sparse is otherwise indistinguishable from grafel ignoring their sparse-checkout — #6338's shape.

## Callers: what changed and what deliberately did not

`ReadFile` keeps its signature **and its contract** (a prefix, nil error), now delegating to `ReadFileLimited` and discarding the bit. So the answer is short: **one call site changed.**

- **Changed** — `internal/gitmeta` `parseSparsePatternFile`, via a new `readGitMetaFileLimited`.
- **Deliberately unchanged, same package** — `readGitMetaFile`'s four other call sites in `cache.go` (`commondir`, HEAD, the gitdir-file, `contentToken`'s loose ref), all under `maxGitPointerBytes = 64 KiB`. Each is one short line with a demanded shape; a 64 KiB prefix of anything else fails to parse and reaches the existing "unknown" path loudly.
- **Deliberately unchanged, elsewhere** — every other `safeio.ReadFile` / `ReadFileReported` caller: `internal/licenses`, `internal/extractors/golang/gomod`, `internal/extractors/yaml/helm`, `internal/extractors/javascript/aliases`, `internal/install/detect`, `internal/links/safe_source_read` (which documents inheriting truncation on purpose), `internal/daemon/walk/saferead`, `internal/engine`'s substrate/wrapper readers and `orm_queries_python_mongo_agg.go:1312`, and the ~20 `MaxConfigFileBytes` config readers across `internal/install`, `internal/mcp`, `internal/registry`, `internal/dashboard`, `internal/agents`, `internal/agentpatterns`, `internal/cli`, `internal/quality/fitness`, `internal/daemon/worktree`, `internal/daemon/watch`. For all of these a truncated read costs a *field*, not a *decision about what gets indexed*. `TestReadFile_StillTruncatesSilently` pins that non-change: if `ReadFile` ever starts erroring on a capped read, the change reached callers it was not meant to reach.

Is it reachable? **No — but state that precisely: this is an absence of counter-example, not a corpus measurement.** No repo I could produce comes near 8 MiB of sparse patterns, and cone mode writes a handful of lines. I did not enumerate a corpus and measure pattern-file sizes, so the honest claim is that nothing reachable to me exhibits it, not that nothing does. This closes a latent hole whose failure mode is invisible by construction, not a live defect — same conclusion the issue reached, now with the hole shut.

## Grading — a named file, not a counter

The claim is "an oversized pattern file no longer silently narrows the index", so the artefact is **a specific file that would have been excluded and now is not, by name**.

`internal/gitmeta/sparse_truncation_6969_test.go` (package `gitmeta_test`, real git):

- **Premise asserted first.** `trSparseFixture` drives `git sparse-checkout set src` and `t.Fatal`s unless `services/orders/h.go` is *excluded* and `src/a.go` is *included* right now. Without that every assertion below is satisfied by a fixture that was never sparse.
- **`..._LineBoundaryCut_IndexesInFull`** — hazard 1. git's own bytes padded to exactly the cap, then `/services/` appended past it. The cut lands on a newline (**asserted**, byte `cap-1`). Observed: `services/orders/h.go` is **in** the index, and `si.Patterns` is empty.
- **`..._MidLineCut_DropsTheCorruptedPattern`** — hazard 2, a *separate* case, not a variant. The cut lands inside `!/src/sub/` leaving exactly `!/src` (**asserted**, both that the fragment is `"!/src"` and that byte `cap-1` is *not* a newline, so the two geometries cannot collapse into one). Observed: `!/src` never reaches `si.Patterns`, and `src/a.go` is **in** the index.
- **`..._ExactlyAtTheCapStaysSparse`** — the counter-case that makes the signal load-bearing rather than a size threshold. A file of exactly `maxSparsePatternBytes` is complete: the repo must stay sparse and `services/orders/h.go` must stay **out**. This asserts the exclusion direction survives at the boundary.

The external file hard-codes `8 << 20` as an **independent literal** rather than importing the constant — a fixture built relative to the constant under test cannot detect that constant changing, and these geometries are byte-exact. `TestSparseCapConstantMatchesTheTestsLiteral` (internal package) is the single place the two are compared and fails loudly if they diverge.

`internal/gitmeta/sparse_truncation_report_6969_test.go` covers the unit directly: truncated → `haveFile=false`, no patterns, and a report line naming the file, saying which way the decision fell, and carrying the issue ref — plus the other direction, that a complete pattern file prints **nothing** (this package is entered on every `whoami`/`ResolveCWD`; a reporter that fires on healthy repos buries its own signal).

`internal/safeio/readfile_limited_6969_test.go` sits at `cap-1` / `cap` / `cap+1` rather than "small" and "huge", with the overflow, unlimited, error and non-change cases.

## Mutants — 6 scored, 6 DEAD

Every one in the **more permissive** direction. Each edit proved landed by an exact occurrence count with the hit line printed; green baseline immediately before each score; restore by `cp` from a SHA-256-verified backup, never `git checkout --` (the three production files were re-hashed to their pre-mutation digests after every round, and the final tree matches).

| # | mutant | verdict |
|---|---|---|
| M1 | `ReadFileLimited` returns `truncated=false` over the cap — the signal never fires | **DEAD** — `TestReadFileLimited_AtTheBoundary` + both oversized cases + the report test |
| M2 | the sparse path ignores `truncated` (`if false && truncated`) — the #6969 pre-state | **DEAD** — both oversized cases + the report test |
| M3 | truncation inferred as `len(b) >= maxBytes`, probe back to `maxBytes` — the ambiguous form the design rejects | **DEAD** — `AtTheBoundary` *and* `ExactlyAtTheCapStaysSparse`, which is the case that exists for it |
| M4 | **the rejected design**: trim the corrupted final line, keep the rest, stay sparse | **DEAD** — both oversized cases + the report test. Hazard 1 is what kills it |
| M5 | reject, but silently — the truncation report never fires | **DEAD** — the report test |
| M6 | `readGitMetaFileLimited` always answers `truncated=false` | **DEAD** — both oversized cases + the report test |

M3 and M4 are the two that matter: they are the two designs a reasonable person would have written instead, and each dies on the case written specifically to distinguish it.

## Verification

`go test -count=1` green on `./internal/safeio/`, `./internal/gitmeta/` and **`./cmd/grafel/`** (the whole-graph digest pin — 165s, and no diff-derived package set reaches it). `gofmt -l` and `go vet` clean on every touched package. `git status --porcelain` clean, no `MUTANT` text in the tree.

Measured on **macOS/APFS, git 2.50.1 (Apple Git-155)** — the real-git fixtures log the version they ran on into every failure.

### DIFF
--- a/internal/gitmeta/saferead.go
+++ b/internal/gitmeta/saferead.go
@@ -25,8 +25,15 @@ const (
 
 	// maxSparsePatternBytes covers info/sparse-checkout, which is a pattern
 	// list and legitimately large in a big monorepo. 8 MiB truncates only a
-	// file that is already pathological, and truncation degrades to the same
-	// "some patterns not applied" outcome an unreadable file already produces.
+	// file that is already pathological.
+	//
+	// A truncated pattern file is NOT the same outcome as an unreadable one and
+	// the two must not be conflated (#6969): unreadable lands on "not sparse"
+	// and indexes the repo in FULL, while a silently-truncated read used to
+	// leave the repo sparse with a partial pattern set, so files were dropped
+	// from the index with no error, no SkipEntry and no missing-entity signal.
+	// parseSparsePatternFile now reads the truncation signal and lands
+	// truncated on "not sparse" too.
 	maxSparsePatternBytes = 8 << 20
 )
 
@@ -78,11 +85,57 @@ func setGitMetaSkipOutput(w io.Writer) func() {
 // read failure as unknown" behaviour; the skip is reported here so that
 // behaviour stops being silent.
 func readGitMetaFile(path string, maxBytes int64) ([]byte, error) {
-	b, err := safeio.ReadFile(path, safeio.FollowSymlinks, maxBytes)
+	b, _, err := readGitMetaFileLimited(path, maxBytes)
+	return b, err
+}
+
+// readGitMetaFileLimited is readGitMetaFile plus the truncation signal.
+//
+// The two exist side by side on purpose. For this package's POINTER files —
+// HEAD, commondir, the .git gitdir-file, a loose ref — a truncated read is
+// already loud: each is one short line, every parser here demands a shape
+// ("ref: ", "gitdir: ", 40 hex bytes), and a 64 KiB prefix of something else
+// fails that shape and falls through to the same "unknown" the callers already
+// handle. Those callers keep readGitMetaFile and are deliberately unchanged.
+//
+// The sparse pattern file is the one LIST-shaped reader here, and a prefix of
+// a list is a valid-looking shorter list. That is the whole of #6969: it fails
+// silently and it fails toward EXCLUSION, the one direction nothing downstream
+// can notice.
+func readGitMetaFileLimited(path string, maxBytes int64) ([]byte, bool, error) {
+	b, truncated, err := safeio.ReadFileLimited(path, safeio.FollowSymlinks, maxBytes)
 	if err != nil {
 		reportGitMetaSkip(path, err)
 	}
-	return b, err
+	return b, truncated, err
+}
+
+// reportGitMetaTruncation announces a git metadata file that was read only in
+// part because it exceeded its byte cap.
+//
+// It is a SEPARATE report from reportGitMetaSkip and not a new branch inside
+// it, because the two carry different facts: a skip says the file was not read
+// at all and names the entry kind; this says the file was read and the tail is
+// missing. It shares the dedup map and the cap, so the combined output is
+// still bounded the way a report nobody scrolls past has to be.
+func reportGitMetaTruncation(path string, maxBytes int64) {
+	gitMetaSkipMu.Lock()
+	if gitMetaSkipSeen == nil {
+		gitMetaSkipSeen = map[string]bool{}
+	}
+	if gitMetaSkipSeen[path] || len(gitMetaSkipSeen) >= maxGitMetaSkipReports {
+		gitMetaSkipMu.Unlock()
+		return
+	}
+	gitMetaSkipSeen[path] = true
+	last := len(gitMetaSkipSeen) == maxGitMetaSkipReports
+	w := gitMetaSkipOut
+	gitMetaSkipMu.Unlock()
+
+	fmt.Fprintf(w, "grafel: %s exceeds %d bytes and was read only in part; treating this repo as NOT sparse so no file is silently excluded (#6969)\n", path, maxBytes)
+	if last {
+		fmt.Fprintf(w, "grafel: further git-metadata skips suppressed after %d\n", maxGitMetaSkipReports)
+	}
 }
 
 // reportGitMetaSkip says out loud that a git metadata file was refused for
--- a/internal/gitmeta/sparse.go
+++ b/internal/gitmeta/sparse.go
@@ -149,7 +149,9 @@ func ProbeRepo(repoPath string) SparseInfo {
 	//
 	// An existing-but-unreadable file lands on "not sparse" too, which is the
 	// over-inclusive direction — a superset, the same direction every other
-	// failure in this file chooses.
+	// failure in this file chooses. So does an OVERSIZED one, whose read is
+	// truncated: before #6969 that was the single input class landing the
+	// other way, staying sparse with a partial pattern set.
 	patterns, haveFile := parseSparsePatternFile(filepath.Join(gitDir, "info", "sparse-checkout"))
 	if !haveFile {
 		return SparseInfo{}
@@ -169,13 +171,40 @@ func ProbeRepo(repoPath string) SparseInfo {
 // states in git and must not be collapsed: an empty pattern file checks out
 // nothing, an absent one checks out everything (#6967 review). Callers that
 // only want the lines can ignore it; ProbeRepo cannot.
+//
+// It reports false for an ABSENT file, an UNREADABLE one, and an OVERSIZED one
+// whose read was truncated (#6969) — three inputs, one verdict, chosen so that
+// every way of failing to obtain the complete pattern set indexes the repo in
+// FULL rather than in part.
 func parseSparsePatternFile(path string) ([]string, bool) {
 	// readGitMetaFile, not os.Open: open(2) is what blocks on a FIFO, so
 	// scanning rather than slurping made no difference to #6416. The pattern
 	// file is name-chosen ("info/sparse-checkout" under the git dir) and is
 	// read before any walk, so no entry-type gate sits in front of it.
-	data, err := readGitMetaFile(path, maxSparsePatternBytes)
+	data, truncated, err := readGitMetaFileLimited(path, maxSparsePatternBytes)
 	if err != nil {
+		return nil, false
+	}
+	// A TRUNCATED pattern file is rejected WHOLESALE, not trimmed to its last
+	// intact line (#6969).
+	//
+	// Two distinct hazards live in a truncated read and only one of them is
+	// fixed by dropping the cut line:
+	//
+	//  1. Every pattern PAST the cut is gone — arbitrarily many, not one. The
+	//     surviving prefix is a valid-looking shorter list, so the repo stays
+	//     sparse with fewer includes and files vanish from the index with no
+	//     error and no SkipEntry. Dropping the cut line leaves this untouched.
+	//  2. The cut line itself is a pattern git never wrote. "!/src/deep/"
+	//     severed at the cap is "!/src", which does not fail to match — it
+	//     matches something WIDER than intended.
+	//
+	// So trimming addresses hazard 2 and leaves hazard 1, which is the one
+	// that silently shrinks the graph. Rejecting the file lands both on "not
+	// sparse", i.e. a FULL index — a superset, the same over-inclusive
+	// direction the absent and unreadable cases already choose above.
+	if truncated {
+		reportGitMetaTruncation(path, maxSparsePatternBytes)
 		return nil, false
 	}
 
--- a/internal/gitmeta/sparse_truncation_6969_test.go
+++ b/internal/gitmeta/sparse_truncation_6969_test.go
@@ -1,0 +1,259 @@
+// Truncation of the sparse-checkout pattern file (#6969).
+//
+// safeio.ReadFile caps a read and returns the PREFIX with a nil error. For
+// every other file this package reads that is loud — a 64 KiB prefix of HEAD
+// is not "ref: ..." and every parser rejects it. The sparse pattern file is
+// the one LIST-shaped reader, and a prefix of a list is a valid-looking
+// shorter list: the repo stayed sparse with fewer includes and files left the
+// index with no error, no SkipEntry and no missing-entity signal.
+//
+// It was also the only input class landing on the EXCLUSION side. #6967
+// settled the other two deliberately toward inclusion (absent → not sparse,
+// unreadable → not sparse); oversized went the other way.
+//
+// WHAT THESE TESTS OBSERVE. Not a counter and not "IsSparse is false" alone —
+// a NAMED FILE that the truncated pattern set would have dropped from the
+// index and that is now indexed. Two cut geometries are separate cases,
+// because the two hazards in the comment on parseSparsePatternFile are
+// different failures:
+//
+//   - a cut on a LINE BOUNDARY silently deletes every pattern after it;
+//   - a cut MID-LINE additionally leaves a pattern git never wrote, which
+//     matches something WIDER than intended rather than nothing.
+//
+// The pattern file here is git's own bytes with padding APPENDED. The config
+// is still written by real git, so #6964's rule ("never hand-write a config
+// value") holds where it is load-bearing — an over-cap pattern file is a shape
+// no git will ever produce, so it has to be constructed.
+//
+// Measured on macOS/APFS. Every case reports the git it ran on.
+package gitmeta_test
+
+import (
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+
+	"github.com/cajasmota/grafel/internal/gitmeta"
+)
+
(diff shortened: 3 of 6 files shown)


cajasmota:

Measured in my own worktree at `44315b203` (detached), plus a second worktree at the parent `448d7bb38`, on **macOS/APFS, git 2.50.1 (Apple Git-155)**. Clean-tree final run, `-count=1`: `./cmd/grafel/` (152.3s) + `./internal/safeio/` + `./internal/gitmeta/` all **ok, exit 0**. Tree `git status --porcelain` empty and the three production files re-hashed to their pre-mutation SHA-256 after every round.

### (d) The named file — confirmed across versions, not just asserted

I built the padded fixture myself and printed the artefact on **both** commits (same fixture, same git):

| commit | `IsSparse` | `patterns` | `IsPathIncluded("services/orders/h.go")` |
|---|---|---|---|
| `448d7bb38` (parent) | `true` | `[/* !/*/ /src/]` | **false** |
| `44315b203` (this PR) | `false` | `[]` | **true** |

So a named file — **`services/orders/h.go`** — that the parent silently drops from the index is indexed by this PR, and the truncation line is printed. The author's three fixtures copied verbatim onto the parent: both oversized cases **FAIL** there, `ExactlyAtTheCapStaysSparse` **PASSES** there (correctly — it is a counter-case, not a distinguishing one). The all-DEAD table is backed by a real behaviour difference, not only by pinning.

### (b) The boundary — verified independently

Not by re-reading the author's table: my own probe over caps `{1,7,64,1000}` × sizes `{cap-1, cap, cap+1, cap+2}` plus `maxBytes ∈ {0, -1, MinInt64, MaxInt64-1, MaxInt64}`. Exactly-at-the-cap is **complete**, `cap+1` is **truncated**, the probe byte never reaches the caller, `maxBytes <= 0` is unlimited and never truncated on either `ReadFileLimited` or `ReadFile`, and `MaxInt64` does not wrap. Correct.

### (a) "Exactly one call site changed" — holds

I enumerated every `safeio.ReadFile` / `ReadFileReported` / `ReadFileLimited` reference in non-test code. `ReadFileLimited` has exactly **one** caller (`internal/gitmeta/saferead.go:106`); `ReadFile` keeps signature and contract and its 24-odd `ReadFileReported` + 8 direct callers are untouched. One completeness nit in the *prose*, not the code: `internal/engine/orm_queries_python_mongo_agg.go:1312` is a `MaxConfigFileBytes` reader that falls outside both "engine's substrate/wrapper readers" and the listed `~20` packages — the enumeration is one member short of the set it claims. No behavioural consequence.

One thing I checked before deciding it is *not* a finding: the read now issues **one extra `Read` past the cap**, which would be a hang risk on a FIFO/character device that supplies exactly `maxBytes` then never EOFs. `nonBlockingOpen` **fstats the opened fd** and refuses anything non-regular (`open_unix.go:50`, `open_other.go:37`), so that read can only ever land on a regular file. Closed.

### (c) Wholesale rejection is the right property

Agreed on the merits, not by deference. Trimming the cut line fixes hazard 2 (a fragment matching *wider*) and leaves hazard 1 (every pattern past the cut) fully intact — and hazard 1 is the one that removes files. "Not sparse → full index" is a superset with no missing entity, the same direction #6964/#6967 chose for absent and unreadable; truncation was the one input class landing the other way. I confirmed hazard 2 is real and not hypothetical: on the parent the mid-line fixture produces `patterns=[/* !/*/ /src/ !/src]` — the fabricated `!/src` genuinely reaches the matcher.

### My mutants — 4 scored, none of M1–M6

Each edit proved landed by an exact occurrence count (1 each) with the hit line printed; green baseline immediately before each score; restored by `cp` from a SHA-256-verified backup, never `git checkout --`.

| # | mutant (all in the permissive direction) | verdict |
|---|---|---|
| N1 | `sparse.go`: `if truncated` → `if truncated && len(data) == 0` — truncation only rejects an *empty* read, so every real truncation stays sparse | **DEAD** — both oversized cases + the report test |
| N2 | `saferead.go`: `maxSparsePatternBytes = 8<<20` → `64<<20` — the signal fires only on a much larger file | **DEAD** — and critically `TestSparseCapConstantMatchesTheTestsLiteral` **is the first failure**, so the independent-literal guard is real, not decoration |
| N3 | `safeio.go`: overflow guard `probe < math.MaxInt64` → `math.MaxInt32` — no truncation ever reported for a cap ≥ 2 GiB | **ALIVE, not production-reachable** — every production cap is 64 KiB…8 MiB. The distinguishing input is a file of `maxBytes+1` bytes under a ≥ 2 GiB cap, i.e. a >2 GiB fixture; I priced it and it is not worth a test |
| N4 | `sparse.go`: report the truncation but **fall through and keep parsing** — the most plausible real slip, and distinct from the author's M5 (reject silently) | **DEAD** — both oversized cases + the report test |

N2 answers the brief's "verify the guard fails when the constant moves" directly: it does.

### Nits (non-blocking)

1. The unchanged-caller enumeration omits `internal/engine/orm_queries_python_mongo_agg.go:1312` (prose only).
2. "**Measured** answer: no repo I can produce comes near 8 MiB of sparse patterns" — that is an absence of counter-example, not a measurement over a corpus. The conclusion is plainly right (8 MiB ≈ 400k pattern lines), but "measured" overstates the evidence by one notch.

Neither changes the verdict. **APPROVE** — not merging and not pushing to the branch.

cajasmota:
## Coordinator mutant CD-1 on `44315b203` — DEAD

Own detached worktree at `/private/tmp/.../scratchpad/coord-mut-6982/wt`, `git status --porcelain` empty at start and after restore, restored by `cp` from a SHA-256-verified backup (`bbf21bb5…` both ways), zero `MUTANT` text remaining. Baseline green immediately before (`./internal/gitmeta/` 37.1s, `./internal/safeio/` 12.4s, exit 0). macOS/APFS.

**CD-1 — make the middle hop swallow the signal.** Both of you attacked the endpoints: M3 attacks how `safeio` decides truncation, N1/N4 attack what the consumer does with it. Neither mutated the **wiring between them**. `readGitMetaFileLimited` (`internal/gitmeta/saferead.go:110`) is the only thing carrying the bool across the package boundary, and it sits one line away from `readGitMetaFile`, which discards that same bool **deliberately** — so a future edit "tidying" the two into one is a live risk and it would look entirely reasonable.

```go
-  return b, truncated, err
+  _ = truncated // MUTANT CD-1
+  return b, false, err // MUTANT CD-1: plumbing swallows the truncation signal
```

Edit proven landed: exactly 1 occurrence of the original line, hit set asserted to be `[110]`, plus an assertion that line 110 is inside `readGitMetaFileLimited` and not its one-line neighbour — 0 occurrences of the original afterwards.

**Result: DEAD**, on three tests:

- `TestParseSparsePatternFile_TruncatedIsReportedAndRejected`
- `TestSparse_OversizedPatternFile_LineBoundaryCut_IndexesInFull`
- `TestSparse_OversizedPatternFile_MidLineCut_DropsTheCorruptedPattern`

### The part worth reading

**`./internal/safeio/` stayed green.** That is the result, not a footnote. The kill is located entirely in the consumer package, which means the signal is graded **as a delivered fact at the point of use**, not merely as a return value `safeio`'s own tests happen to observe. A bool that is correct at its source and dropped in transit is indistinguishable from one that was never computed, and this is the assertion that separates them.

It also independently corroborates the reviewer's item (d) from the other side: they showed a named file (`services/orders/h.go`) that the parent excludes and this includes; CD-1 shows the *only* wire carrying that verdict cannot be cut without two of those same named-file tests going red.

### Process note, since it bit me again

My first attempt asserted the hit set was `[111]`. It is `[110]`. **The assertion aborted the edit and the file was never modified** — had I skipped it, the suite would have come back green against an unmutated file and I would have reported CD-1 ALIVE and argued that the plumbing was ungraded. That is the fourth time this week the hit-set assertion has caught an off-by-one that would have inverted a verdict. My second attempt then failed `go vet` (`declared and not used: truncated`) — a non-compiling mutant proves nothing, so it needed `_ = truncated` before it was worth scoring.

Issue #6906, opened 2026-09-06
DECIDED against TYPED_AS (#6984 adopted REFERENCES+ref_kind=field_target_type): now a cleanup — delete or reserve TYPED_AS and HAS_TYPE, and kill the comments describing a producer that will never exist


cajasmota:
`TYPED_AS` is a valid, documented relationship kind with **zero producers** — every `SCOPE.Enum` value-set node in every language is a terminal orphan by construction

Found while grounding contributor report #6726, which measures **`SCOPE.Enum`: 721 entities, 100% terminal orphan** over a real 73k-entity csharp/fsharp/java/protobuf corpus. That number is not a resolver regression and not a corpus artefact — it is the exact, expected behaviour of the code as written.

## The defect

`internal/extractor/enum_valueset.go` documents an inbound edge that nothing emits.

- `:23` — "which fields are constrained to enum X?" — **the inbound `TYPED_AS` edges**
- `:70` — "This value is stored as the entity's `QualifiedName`, so a `TYPED_AS` edge whose `ToID` equals it binds via the resolver's `byQualifiedName` exact-match tier."

Both describe a binding whose producer side does not exist. `TYPED_AS` appears in exactly three places in the tree:

```
internal/types/kinds.go:1753   RelationshipKindTypedAs RelationshipKind = "TYPED_AS"   ← declaration
internal/types/kinds.go:2009   RelationshipKindTypedAs,                                ← valid-kinds slice
internal/extractor/enum_valueset.go:23, :70                                            ← the two comments
```

`RelationshipKindTypedAs` has **zero** references outside those two `kinds.go` lines — no extractor, no enricher, no test. `EnumQualifiedName` has exactly **one** production caller (`enum_valueset.go:139`, the entity's own `QualifiedName` field); no edge builder anywhere constructs a `ToID` from it, and every other `"scope:enum:"` literal in the tree is a **test expectation**, never an endpoint:

```
internal/extractors/{python,golang,java,javascript,ruby}/…_test.go, internal/mcp/literal_parity_tool_test.go
```

So the kind is *valid* (it passes `IsValidRelationshipKind`), *documented*, and *never produced*. Every `SCOPE.Enum` node carries `Module CONTAINS` and nothing else, in every language that emits one.

## Reproduced on two entities

A purpose-built fixture (csharp, java, protobuf, fsharp — each an enum plus a field **and** a parameter typed as that enum), indexed end-to-end in-process, produced 38 entities / 64 relationships:

[...]

Issue #6726, opened 2026-08-27
Feedback report (csharp/fsharp/java/protobuf, 73k entities): Schema 99.3% / Operation 94.6% orphan, 1.0% annotation coverage


Bvalenzuela:
# grafel feedback report

Generated: 2026-08-27T21:05:18Z
grafel version: v0.3.0 (commit b037c3f, built 2026-08-21T18:26:38Z)
Group profile: 9 language(s) (csharp, fsharp, markdown, protobuf, java, css, yaml, c, html), 73000-74000 entities, 221000-222000 relationships
Confidence: 81% (29/36 sanity checks passed)

### Entities by language

| Language | Entity count (range) |
|---|---|
| c | 6-20 |
| csharp | 100+ |
| css | 21-100 |
| fsharp | 100+ |
| html | 6-20 |
| java | 100+ |
| markdown | 100+ |
| protobuf | 100+ |
| yaml | 6-20 |

### Entity kind distribution

| Kind | Language | Count (range) |
|---|---|---|
| Config | csharp | 6-20 |
| Controller | csharp | 6-20 |
| Middleware | csharp | 21-100 |
| Model | csharp | 100+ |
| Operation | csharp | 21-100 |
| SCOPE.Component | csharp | 100+ |
| SCOPE.Enum | csharp | 100+ |
| SCOPE.ExceptionType | csharp | 21-100 |
| SCOPE.Operation | csharp | 100+ |
| SCOPE.Pattern | csharp | 100+ |
| SCOPE.Process | csharp | 21-100 |
| SCOPE.Schema | csharp | 100+ |
| http_endpoint_call | csharp | 6-20 |
| http_endpoint_definition | csharp | 21-100 |
| SCOPE.Stylesheet | css | 21-100 |
| SCOPE.Component | fsharp | 100+ |
| SCOPE.Operation | fsharp | 100+ |
| SCOPE.Schema | fsharp | 100+ |
| SCOPE.Enum | java | 21-100 |
| SCOPE.Operation | java | 21-100 |
| SCOPE.Schema | java | 21-100 |
| SCOPE.CodeBlock | markdown | 100+ |
| SCOPE.Document | markdown | 21-100 |
| SCOPE.Operation | protobuf | 21-100 |
| SCOPE.Schema | protobuf | 100+ |

### Source-window completeness

Entities with valid start/end line: **89.2%** (65247 of 73182)

### Annotation coverage

Framework-annotated entities: **1.0%** (696 of 73182)

[...]

PR #6984, opened 2026-09-07
feat(#6912): C# fields carried their declared type as a property and never as an edge — emit the field→type REFERENCES edge, for same-file types only

cajasmota:
C# `SCOPE.Schema`/field entities recorded their declared type in the `field_type` **property** and never in `Relationships`, so every field was a leaf on its outbound side and "which fields point at `Customer`?" returned empty. This adds the producer, for C# only, and only where the edge is guaranteed to bind.

**Arm A of a multi-language defect — `Refs #6912`, the issue stays open.** Java does not retain the type string at all and several other core extractors don't either; capturing it is per-language cost and a different arm.

## The standing gate was cleared FIRST, and it is now a committed test

#6906/#6912 exist because a comment described a binding nobody built, so nothing was implemented until a synthetic field→type edge had been driven through the production resolver. `TestCsharpFieldTypeRefs_ResolverBindsEveryEdge` is that probe, kept: it runs the real path (`resolve.BuildIndex` → `resolve.ReferencesEmbedded`, exactly as graph assembly does) and asserts all 10 emitted edges are **rewritten to real entity IDs**, naming each `(field ⇒ target entity)` pair. Re-confirmed after rebasing onto `bd03cdfeb` — #6980 reworked `indexByName` and the `nameAmbigRef` raise in `refs.go`, so the pre-rebase result was re-measured rather than carried over.

## The three decisions, and the evidence for each

**Kind + discriminator — adopted, not minted.** `REFERENCES` with the property `ref_kind: "field_target_type"`, which is what `internal/custom/{python,golang,java,javascript,ruby}` already emit through the shared `referencesClassEdge`. `TYPED_AS` and `HAS_TYPE` stay producerless (#6906, #5828); a third spelling for one relation would split every "which fields point at X?" query (#6902's family).

**Address — deliberately NOT that lane's `Class:<Target>`, and the reason is measured.** Both `Customer` and `Class:Customer` reach the resolver's bare-name tier, which returns **AMBIGUOUS** as soon as a second file declares a type of the same name (`Result`, `Options`, `Settings` are routine in C#). Probed both ways:

| ToID form | one file declares `Customer` | two files declare `Customer` |
|---|---|---|
| `Customer` | rewritten | **ambiguous** |
| `Class:Customer` | rewritten | **ambiguous** |
| `scope:component:class:csharp:<file>:Customer` | rewritten | **rewritten** |
| `scope:enum:<file>:OrderStatus` | rewritten | **rewritten** |
| `OrderStatus` (bare) | **ambiguous** — the `SCOPE.Schema`/enum twin shares the Name | ambiguous |

So targets are addressed by location: `extractor.BuildComponentStructuralRef` for class/interface/struct/record, `extractor.EnumQualifiedName` for enums. Once resolved the `ToID` is rewritten to the target's entity ID, so the stub dialect is a **binding mechanism only** — no downstream query sees it, and the kind/`ref_kind` consistency the custom lane established is untouched. `TestCsharpFieldTypeRefs_SameNameInTwoFiles` holds this, and `TestCsharpFieldTypeRefs_ResolverBindsEveryEdge` carries a positive control asserting the bare name really is ambiguous in that fixture — otherwise the structural address would be ungraded.

The enum target is the `SCOPE.Enum` value-set node, which gives that population its **first inbound edge** (#6906 measures it at 100% terminal orphan). That is the "widen `REFERENCES` to carry QN-addressed enum targets" option #6912's own comment argues for; `ref_kind` expresses the distinction as a property.

**A non-binding edge is worse than no edge, so: same-file types only.** An unresolved stub is kept verbatim and dangling and classified `bug-extractor`, with no masking branch and no drop pass, and each emitted edge adds exactly one endpoint to the disposition denominator. Being declared in this file is the only condition a pass-1 per-file extractor can *know*, so it is the only condition under which an edge is emitted.

### The rule, branch by branch — and what grades each branch

Every bare type name written in the declared type is a candidate (`nullable_type`, `array_type`, `pointer_type`, tuple and generic-argument syntax are all unwrapped by one recursive walk); a candidate becomes an edge **iff a type of that name is declared in this same file**.

| input | result | graded by |
|---|---|---|
| `Customer Buyer` (declared here) | edge → `…:Customer` | `_EmittedEdges` |
| `OrderStatus Status` (enum here) | edge → `scope:enum:…:OrderStatus` | `_EmittedEdges` |
| `IShipper` / `Money` / `Address` (interface / struct / record) | edge | `_EmittedEdges` |
| `Address? ShipTo`, `Customer[] History` | edge to the element type | `_EmittedEdges` |
| `List<Customer>`, `Dictionary<string, Customer>` | edge to **`Customer`**, never to `List`/`Dictionary` | `_EmittedEdges` + `_ForbiddenTargets` (the wrappers are banned targets **by name**) |
| `Order Parent` (self-reference) | edge → itself | `_EmittedEdges` |
| `int` / `string` | **no edge** | `_ForbiddenTargets`, by field name |
| `System.Net.Http.HttpClient` | **no edge** | `_ForbiddenTargets`, by field name |
| `App.Other.Customer` where this file also declares `Customer` | **no edge** — a qualified name is not descended into, so the rightmost segment cannot bind to the wrong entity | `_QualifiedTypeIsNotResolved` |
| type declared in **another** file | **no edge** | `_CrossFileTypeGetsNoEdge` (+ positive control that the field exists) |
| name declared **twice in one file** | **no edge** — removed from the index rather than guessed | `_SameNameDeclaredTwiceInOneFile` (+ positive control that both declarations and the field exist) |
| open type parameter `T` | no edge — not declared here | same single check |

The in-file check is deliberately the **only** guard. A primitive blocklist in front of it would fire only where it already fires, leaving both ungraded.

**The cost of the same-file rule, stated as an assertion and not as prose:** a field whose type lives in another file gets nothing. Closing that needs a cross-file type view and `FileInput` has none in pass 1 (`CrossFileFields` is attached for pass-2.5 detectors only). Separate arm, not a widening of this one.

## Measurement — and the golden-corpus number is zero

All figures below are from the **core pass-1 extractor**, which runs regardless of the custom gate (#6966); this pass consults no detector output, so **custom-gate ON and OFF are the same number**. macOS/APFS, measured in `…/scratchpad/impl-6912a-q8w3z/wt`.

**Golden corpus: 0 edges added, 0 `*/field` orphans cleared.** Its 8 `.cs` files contain **3 C# field entities in total**, enumerated rather than counted:

- `csharp-aspnet-core-mini` `UsersController._userService : IUserService` — interface declared in another file
- `csharp-hangfire-mini` `JobSelector.Ascending : string` — primitive
- `csharp-hangfire-mini` `JobSelector.Descending : string` — primitive

None has a same-file declared type, so none gains an edge. The issue's "86 field orphans" ceiling is dominated by other languages; **none of the 86 clears here**, and that is why the `cmd/grafel` digest and the golden ratchet are both unmoved rather than silently re-baselined.

**Real C# corpora — where the mechanism actually has reach** (read-only clones under `archigraph-corpora`, extractor run directly):

| corpus | `.cs` files | field entities | fields gaining an edge | edges |
|---|---|---|---|---|
| aspnetcore-mvc | 2677 | 7637 | 490 (6.4%) | 493 |
| aspnetcore-realworld | 75 | 140 | 9 (6.4%) | 9 |
| WakeOnLAN | 22 | 133 | 6 (4.5%) | 6 |

21 of aspnetcore-mvc's 493 target an enum. Sampled edges are semantically real — `Family.Dad → Person`, `Parent.Children → Child`, `Order11.Customer → Person11`, `ResourceInvoker._exceptionContext → ExceptionContextSealed`.

**#6726's 99.3% of 35,353 is the contributor's corpus, not ours. The mechanism transfers; the number does not, and nothing here claims it moves.**

## Known over-fires: namespace scope is NOT consulted (added after review)

The same-file check is **file scope and nothing else**. Two cases emit a wrong edge, both found in review:

```csharp
namespace A { class Customer { … } }
namespace B { class Order { public Customer Buyer { get; set; } } }   // one file → WRONG edge
class Box<Customer> { public Customer Item { get; set; } }            // beside a same-file class Customer → WRONG
```

**These are worse than a dangling edge, by the same logic that chose this PR's address.** The edge *binds*, so it never lands in `bug-extractor` and no disposition figure will ever surface it — unlike an unresolved stub, which is at least counted.

**Measured incidence: zero.** aspnetcore-mvc has 12 `.cs` files declaring two or more namespaces and emits **no** field-type edge inside any of them; aspnetcore-realworld and WakeOnLAN have no multi-namespace file at all. Latent, not live — which is why this arm records the limitation instead of fixing it (a follow-up issue is being filed).

Recorded in three places rather than only on the board: the `field_type_refs.go` header (at the code someone reads), and two tests — `TestCsharpFieldTypeRefs_KnownOverFire_NamespaceScopeIsNotConsulted` and `..._TypeParameterShadowsSameFileType` — which **pin the wrong behaviour on purpose** and are expected to fail when the follow-up fixes it, so no fix can change this silently.

**The duplicate-name fixture no longer implies coverage it does not assert.** It was written with its two declarations in two namespaces of one file, which made its *shape* suggest namespace scope was part of the collide branch. It now uses partial classes — same branch, no implication.

**One prose over-claim removed:** "that one check drops … open type parameters `T`" was false in general; `T` is dropped only because nothing in the file happens to be named `T`. The `Box<Customer>` case above is the counter-example.

## Verification

- `go test -count=1 ./internal/extractors/csharp/` — green (the only package touched). `./internal/extractors/...` green.
- **`./cmd/grafel/` green** — the `custom_extractor_spans_6118_test.go` digest pin is **unmoved**, and the reason was enumerated, not assumed: its 7 inline `.cs` fixtures contain exactly two field entities — `OrdersController._svc : IOrderService` and `ShopContext.Orders : DbSet<Order>` — and neither type is declared in the file that uses it, so this pass emits nothing there. No constant was bumped.
- **Golden ratchet** (`scripts/quality/run.sh --ratchet`, isolated `HOME`/`GRAFEL_HOME`/`GRAFEL_DAEMON_ROOT`): 38 gated fixtures held their baseline, 29 at 100% must-have recall. No baseline moved.
- Both whole-graph gates were re-run **after** the rebase onto `bd03cdfeb`; the pre-rebase runs are not quoted as evidence for the merged state.
- Re-run again after the review commit: `./internal/extractors/...` exit 0, `./cmd/grafel/` exit 0 (148s, digest still unmoved), ratchet exit 0 (38 held, 29 at 100%). That commit is comment-and-test only — no production behaviour changed.

### Mutants, all in the MORE PERMISSIVE direction — 4 applied, 4 DEAD

Each edit was applied from a shasum-verified backup with an asserted occurrence count of exactly 1 **and** an assertion on the enclosing function (and, for M3, on the neighbouring `case` line); `go vet` passed before every score, and the tree was restored by `cp` and shasum-compared afterwards.

| mutant | site | verdict |
|---|---|---|
| **M1** drop the in-repo check — fall back to a bare-name `ToID` for any unresolved candidate | `attachCsharpFieldTypeRefs` | **DEAD** — `_EmittedEdges`, `_CrossFileTypeGetsNoEdge`, `_SameNameDeclaredTwiceInOneFile`, `_ResolverBindsEveryEdge` all fail |
| **M2** descend into qualified names | `csTypeRefCandidates` | **DEAD** — `_QualifiedTypeIsNotResolved` |
| **M3** address targets by bare name instead of structurally | `csInFileTypeTargets` | **DEAD** — `_EmittedEdges`, `_QualifiedTypeIsNotResolved`, `_SameNameInTwoFiles`, `_ResolverBindsEveryEdge` |
| **M4** keep a name declared twice in one file and bind the first | `csInFileTypeTargets` | **DEAD** — `_SameNameDeclaredTwiceInOneFile` |
| **RM3** drop the per-field dedup *(review-found, ALIVE → now graded)* | `attachCsharpFieldTypeRefs` | **DEAD** — `(Customer, Customer) Pair` yields one `want` row, so a duplicate fails `_EmittedEdges` and `_ResolverBindsEveryEdge` |
| **RM4** drop `alias_qualified_name` from the not-descended case *(review-found, ALIVE → now graded)* | `csTypeRefCandidates` | **DEAD** — `global::Customer Aliased` is a named forbidden field; fails `_EmittedEdges`, `_ForbiddenTargets`, `_ResolverBindsEveryEdge` |

The review's RM1 and RM2 stay ALIVE and are **not** production-reachable / equivalent, with the reviewer's reasons: `Extract` is per-file so the `SourceFile != filePath` guard can never fire, and `csFieldTypeRefsMetaKey` has exactly one write site and it is on a field record.

## Not done here

- Cross-file targets (arm B; needs a pass-1 cross-file type view).
- Any other language — no second extractor was touched.
- The two over-fires above (namespace scope, type-parameter shadowing) — pinned, not fixed; zero measured incidence.
- The corollary the reviewer drew and I did not: if `Class:<Target>` goes ambiguous under duplicate type names, the shipped **custom lane** dangles the same way in golang/javascript/ruby/java (`refs.go:4574` masks `lang == "" || lang == "python"` only). Filed separately by the coordinator; out of scope for this arm.
- `internal/extractor/enum_valueset.go:23,70` still describe the `TYPED_AS` plan. Now that a real producer exists those comments are misleading and should be corrected on **#6906**, which owns them; left alone to keep this change to one package.

Refs #6912, #6906, #5828, #6726, #6966.

### DIFF
--- a/internal/extractors/csharp/csharp.go
+++ b/internal/extractors/csharp/csharp.go
@@ -91,6 +91,12 @@ func (e *Extractor) Extract(_ context.Context, file extractor.FileInput) ([]type
 	// ones declared in this same file (which is all #4854 ever emitted, and
 	// always as EXTENDS).
 	entities = attachCsharpHierarchy(entities)
+	// Issue #6912 — field→declared-type REFERENCES edges, for the types
+	// DECLARED IN THIS FILE only. Runs after the walk because the set of
+	// in-file declarations is only complete once every declaration is on the
+	// table; see field_type_refs.go for why the target is addressed
+	// structurally rather than by bare name.
+	entities = attachCsharpFieldTypeRefs(entities, file.Path)
 	// Issue #90 — language tag for resolver dynamic-pattern dispatch.
 	extractor.TagRelationshipsLanguage(entities, "csharp")
 	extractor.TagEntitiesLanguage(entities, "csharp")
--- a/internal/extractors/csharp/field_members.go
+++ b/internal/extractors/csharp/field_members.go
@@ -41,7 +41,7 @@ func emitFieldMembers(
 	var fields []types.EntityRecord
 	seen := make(map[string]bool)
 
-	add := func(name, typ string, startNode ts.Node) {
+	add := func(name, typ string, typeNode, startNode ts.Node) {
 		if name == "" || seen[name] {
 			return
 		}
@@ -72,6 +72,14 @@ func emitFieldMembers(
 			Metadata:           map[string]interface{}{"subtype": "field", "owner": ownerName},
 			EnrichmentRequired: false,
 		})
+		// Issue #6912 — stash every bare type name written in the declared
+		// type for attachCsharpFieldTypeRefs, which turns the ones DECLARED IN
+		// THIS FILE into REFERENCES edges once the walk has seen every
+		// declaration. Nothing is decided here: a field may be declared before
+		// the type it names.
+		if cands := csTypeRefCandidates(typeNode, src); len(cands) > 0 {
+			fields[len(fields)-1].Metadata[csFieldTypeRefsMetaKey] = cands
+		}
 	}
 
 	// Record positional parameters: `record Foo(int Id, string Name)`. The
@@ -83,9 +91,10 @@ func emitFieldMembers(
 				if p == nil || p.Type() != "parameter" {
 					continue
 				}
-				typ := leafTypeName(p.ChildByFieldName("type"), src)
+				typeNode := p.ChildByFieldName("type")
+				typ := leafTypeName(typeNode, src)
 				name := childFieldText(p, "name", src)
-				add(name, typ, p)
+				add(name, typ, typeNode, p)
 			}
 		}
 	}
@@ -98,15 +107,17 @@ func emitFieldMembers(
 			}
 			switch ch.Type() {
 			case "property_declaration":
-				typ := leafTypeName(ch.ChildByFieldName("type"), src)
+				typeNode := ch.ChildByFieldName("type")
+				typ := leafTypeName(typeNode, src)
 				name := childFieldText(ch, "name", src)
-				add(name, typ, ch)
+				add(name, typ, typeNode, ch)
 			case "field_declaration":
 				vd := findChildByType(ch, "variable_declaration")
 				if vd == nil {
 					continue
 				}
-				typ := leafTypeName(vd.ChildByFieldName("type"), src)
+				typeNode := vd.ChildByFieldName("type")
+				typ := leafTypeName(typeNode, src)
 				for j := 0; j < int(vd.ChildCount()); j++ {
 					d := vd.Child(j)
 					if d == nil || d.Type() != "variable_declarator" {
@@ -122,7 +133,7 @@ func emitFieldMembers(
 							}
 						}
 					}
-					add(name, typ, ch)
+					add(name, typ, typeNode, ch)
 				}
 			}
 		}
--- a/internal/extractors/csharp/field_type_refs.go
+++ b/internal/extractors/csharp/field_type_refs.go
@@ -49,10 +49,31 @@ import (
 // IN THE SAME FILE — the single condition under which a pass-1 per-file
 // extractor can know an entity exists at all. That one check is what drops
 // primitives (`int`, `string`), framework and BCL types (`HttpClient`), generic
-// wrappers (`List`), open type parameters (`T`) and every unmodelled name: none
-// of them is declared in the file, so none of them gets an edge. It is
-// deliberately the ONLY guard — a redundant primitive blocklist in front of it
-// would fire only where this check already fires, leaving both ungraded.
+// wrappers (`List`) and every unmodelled name: none of them is declared in the
+// file, so none of them gets an edge. It is deliberately the ONLY guard — a
+// redundant primitive blocklist in front of it would fire only where this check
+// already fires, leaving both ungraded.
+//
+// WHAT THE CHECK IS NOT. It is FILE scope and nothing else: it consults neither
+// the C# namespace nor type-parameter scope, so it over-fires twice, and both
+// over-fires produce an edge that BINDS — which makes them worse than a
+// dangling edge, because `bug-extractor` never sees a bound edge and no
+// disposition figure will surface them.
+//
+//	namespace A { class Customer … }
+//	namespace B { class Order { Customer Buyer … } }   ← one file: WRONG edge
+//	class Box<Customer> { Customer Item … }            ← beside a same-file
+//	                                                     class Customer: WRONG
+//
+// So an open type parameter `T` is dropped only because nothing in the file
+// happens to be named `T`, NOT because the guard understands type parameters.
+// Measured incidence is zero — aspnetcore-mvc has 12 .cs files declaring two or
+// more namespaces and emits no field-type edge inside any of them, and
+// aspnetcore-realworld and WakeOnLAN have no multi-namespace file at all — which
+// is why this arm records the limitation instead of fixing it. Both cases are
+// PINNED as known-wrong behaviour by the two
+// TestCsharpFieldTypeRefs_KnownOverFire_* cases, which a fix is expected to
+// break.
 //
 // THE COST OF THAT RULE, stated plainly: a field whose type is declared in
 // ANOTHER file gets no edge, which on a one-type-per-file C# codebase is most
--- a/internal/extractors/csharp/field_type_refs.go
+++ b/internal/extractors/csharp/field_type_refs.go
@@ -1,0 +1,210 @@
+package csharp
+
+import (
+	"github.com/cajasmota/grafel/internal/treesitter/ts"
+
+	"github.com/cajasmota/grafel/internal/extractor"
+	"github.com/cajasmota/grafel/internal/types"
+)
+
+// field_type_refs.go — the field→declared-type edge for C# (issue #6912).
+//
+// THE GAP. emitFieldMembers records a field's declared type as the `field_type`
+// PROPERTY and never as a relationship, so every SCOPE.Schema/field entity is a
+// leaf on its outbound side: `Order.Buyer` knows it is a `Customer` and the
+// graph cannot answer "which fields point at Customer?". #6912 measured
+// SCOPE.Schema/field at 86 of the golden corpus's 109 SCOPE.Schema orphans.
+//
+// THE VOCABULARY IS NOT NEW. The custom lane already emits exactly this edge in
+// five languages through one shared helper (internal/custom/{python,golang,
+// java,javascript,ruby}, `referencesClassEdge`): Kind REFERENCES with the
+// property `ref_kind: "field_target_type"`. Core adopts that spelling rather
+// than minting a third name — `TYPED_AS` and `HAS_TYPE` are both declared in
+// internal/types/kinds.go with zero producers (#6906, #5828), and a third
+// spelling would split every "which fields point at X?" query.
+//
+// THE ADDRESS IS NOT the custom lane's `Class:<Target>`, and that divergence is
+// measured, not stylistic. Both `Customer` and `Class:<Customer>` reach the
+// resolver's bare-name tier, which returns AMBIGUOUS the moment a second file
+// declares a type of the same name — `Result`, `Options` and `Settings` are
+// routine in C#. This pass therefore addresses the target by its exact
+// location:
+//
+//	class / interface / struct / record → extractor.BuildComponentStructuralRef
+//	                                      "scope:component:class:csharp:<file>:<Name>"
+//	enum                                → extractor.EnumQualifiedName
+//	                                      "scope:enum:<file>:<Name>"
+//
+// Both forms are file-scoped and bind exactly — proven against the production
+// resolver (resolve.BuildIndex → LookupStatusHint / ReferencesEmbedded) in
+// field_type_refs_6912_test.go, including the two-files-one-name case where the
+// bare-name forms go ambiguous. Once resolved the ToID is rewritten to the
+// target entity's ID, so the stub dialect is a binding mechanism only and no
+// downstream query sees it.
+//
+// A NON-BINDING EDGE IS WORSE THAN NO EDGE. An unresolved stub is kept verbatim
+// and dangling and classified `bug-extractor`; there is no masking branch and no
+// drop pass, and every emitted edge adds exactly one endpoint to the disposition
+// denominator (#6906). So the pass emits ONLY when the target type is DECLARED
+// IN THE SAME FILE — the single condition under which a pass-1 per-file
+// extractor can know an entity exists at all. That one check is what drops
+// primitives (`int`, `string`), framework and BCL types (`HttpClient`), generic
+// wrappers (`List`), open type parameters (`T`) and every unmodelled name: none
+// of them is declared in the file, so none of them gets an edge. It is
+// deliberately the ONLY guard — a redundant primitive blocklist in front of it
+// would fire only where this check already fires, leaving both ungraded.
+//
+// THE COST OF THAT RULE, stated plainly: a field whose type is declared in
+// ANOTHER file gets no edge, which on a one-type-per-file C# codebase is most
+// of them. Closing that needs a cross-file type view, and FileInput has none in
+// pass 1 (CrossFileFields is attached for pass-2.5 detectors only). That is a
+// separate arm, not a widening of this one.
+
+// csFieldTypeRefsMetaKey is the per-field stash written by emitFieldMembers and
+// consumed — and deleted — by attachCsharpFieldTypeRefs. The candidates cannot
+// be turned into edges during the walk: `class Order { Customer Buyer; }` may be
+// declared BEFORE `class Customer` in the same file, so the set of in-file
+// declarations is only complete once the walk has finished.
+const csFieldTypeRefsMetaKey = "field_type_refs"
+
+// csFieldTargetRefKind is the value of the `ref_kind` edge property, matching
+// the custom lane's shared referencesClassEdge helper verbatim.
+const csFieldTargetRefKind = "field_target_type"
+
+// csTypeRefCandidates returns every bare type name written in a field's declared
+// type expression, in source order and without deduplication.
+//
+// It descends the type node and collects `identifier` leaves, which unwraps
+// nullable (`Order?`), array (`Order[]`), pointer (`Order*`) and tuple
+// (`(int, Customer)`) syntax for free, and for a generic returns the CONSTRUCTOR
+// as well as its arguments — `Dictionary<string, Order>` yields
+// [Dictionary, Order] and `List<Order>` yields [List, Order]. Emitting to `List`
+// would be wrong and emitting to `Order` is right; the caller's in-file check is
+// what separates them, since `Dictionary` and `List` are never declared in the
+// file and `Order` may be. A user-defined generic (`Box<T>` declared here) is
+// picked up by the same mechanism.
+//
+// A qualified name (`App.Other.Customer`, `System.Net.Http.HttpClient`) is NOT
+// descended into, and that is the one case where the rule is stricter than the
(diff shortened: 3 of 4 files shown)


cajasmota:
## Independent adversarial review — **APPROVE**, with two follow-ups to file (neither blocking)

Measured in my own worktree at `/private/tmp/claude-501/-Users-jorgecajas-Projects-archigraph/c0fefed4-0195-4c81-9a94-07a7145710a4/scratchpad/review-6984-zx7/wt`, detached at `bb79923b9` (parent `bd03cdfeb`). macOS/APFS. Every figure below is from the **core pass-1 extractor**, which is custom-gate-independent; the gate-independence is not merely asserted here — the two `cmd/grafel` tests my permissive control turned red are the **GateOff** ones.

### (a) The `Class:<Target>` ambiguity claim is TRUE — and the custom lane IS therefore defective

Reproduced independently: two files each declaring `class Customer`, synthetic probe stubs injected onto the same field record, driven through `resolve.BuildIndex` → `resolve.ReferencesEmbedded`.

| ToID | 1 file declares `Customer` | 2 files declare `Customer` |
|---|---|---|
| `Customer` | bound | **NOT bound** (kept verbatim) |
| `Class:Customer` | bound | **NOT bound** (kept verbatim) |
| this PR's structural ref | bound | **bound** |

The deviation is forced, not stylistic. **And the same measurement condemns the shipped custom lane**: `referencesClassEdge` in `internal/custom/{python,golang,java,javascript,ruby}` emits `Class:<Target>`, so any repo where two files declare a same-named model produces a dangling `field_target_type` edge there today. Worse, only *one* language is masked: `internal/resolve/refs.go:4574` routes `Class:` stubs to `DispositionDynamic` **only under `lang == "" || lang == "python"`**, and the Go/JS/Ruby helpers stamp `language` explicitly — so golang/javascript/ruby/java dangle straight into `bug-extractor`. That is a live defect, worth its own issue; it is a reason to keep this PR's address, not to block it.

### (b) NOT a #6451 repeat

I searched for readers rather than accepting the argument. The only non-producer reference to `ref_kind == "field_target_type"` tree-wide is one custom-lane test (`internal/custom/java/issue4367_livedrepro_test.go:134`). Nothing downstream reads the pre-resolution ToID, and the core stub always resolves inside its own file's record batch, so there is no stub-survival path. `SCOPE.ExternalAPI` was two *surviving* dialects on one kind; this is two binding mechanisms with a single post-resolution representation. Note the resolver is itself `Class:`-prefix-aware (the salvage arm above), so the dialects are **not** interchangeable inside `resolve` — which again argues for the deviation.

### (c) Golden coverage is zero, but the producer is NOT ungated

The digest hashes `R %s -[%s]-> %s` (`cmd/grafel/custom_extractor_spans_6118_test.go:440`), so a real edge would move it — the "unmoved because zero edges" explanation is structurally sound. I then tested whether the pin is *load-bearing* rather than assuming it: applied a permissive bare-name-fallback control (emit `ToID: <cand>` for every unresolved candidate) and re-ran the whole package —

```
./cmd/grafel/  FAIL  (142s)
  --- FAIL: TestCustomExtractorGateOffGraphIsUnchanged
  --- FAIL: TestCustomExtractorGateOffDeltaIsExactlyTheDocumentedSpanGain
```

So `cmd/grafel` is a genuine over-firing gate for this producer. The golden **ratchet** has no reach (3 C# fields, none same-file-typed — I confirm the enumeration), but "zero golden fixture" ≠ "ungated". **I would not require a new fixture.**

Suites re-run by me, as shipped: `./internal/extractors/...` exit 0 (75 ok) · `./cmd/grafel/` exit 0 (145.5s) · `scripts/quality/run.sh --ratchet` exit 0, **38 gated fixtures held, 29 at 100% must-have recall** — matching the PR body exactly.

### Corpus numbers reproduced — all three, exactly

Walked each read-only corpus, ran the extractor per file, counted by name.

| corpus | files | fields | fields w/ edge | edges | enum targets | PR claim |
|---|---|---|---|---|---|---|
| aspnetcore-mvc | 2677 | 7637 | 490 | **493** | **21** | matches |
| aspnetcore-realworld | 75 | 140 | 9 | **9** | 0 | matches |
| WakeOnLAN | 22 | 133 | 6 | **6** | 4 | matches |

**A named field the parent leaves unwired and this wires:** `InputFormatterResult._failure` in `src/Mvc/Mvc.Abstractions/src/Formatters/InputFormatterResult.cs` → `scope:component:class:csharp:…/InputFormatterResult.cs:InputFormatterResult`. **One it wires wrongly: none found on any of the three corpora** (see below).

### (d) Over-firing — two real cases, both zero-incidence on the corpora

**1. Namespace scope is not consulted (the serious one).** One file, one declaration, no ambiguity — and a wrong edge that *binds*, so `bug-extractor` will never surface it:

```csharp
namespace App.A { public class Customer { … } }
namespace App.B { public class Order { public Customer Buyer { get; set; } } }
```
→ emits `F.cs:Order.Buyer -> scope:component:class:csharp:F.cs:Customer`, though `App.B.Order` cannot see `App.A.Customer` without a `using`.

The PR's `_SameNameDeclaredTwiceInOneFile` fixture is built on **exactly this two-namespace shape** and grades only the collide branch; the single-declaration-wrong-namespace neighbour is left open. Classic "fixture pins one axis, leaves its neighbour open".

**Incidence, measured, not guessed:** aspnetcore-mvc has 12 `.cs` files declaring ≥2 namespaces and **0 field-type edges are emitted inside any of them**; aspnetcore-realworld and WakeOnLAN have 0 multi-namespace files. Latent, not live. File as a follow-up.

**2. Type-parameter shadowing.** `public class Box<Customer> { public Customer Item { get; set; } }` beside a same-file `class Customer` → false edge `Box.Item -> …:Customer`. Rarer (single-letter parameters dominate), same follow-up.

Cases I attacked that came back **clean**: nested types (`Outer.Inner` — the extractor names the nested class `Inner` and the edge is correct), a same-file class shadowing a BCL name (correct — the local type really does win), `global::Customer` (no edge), `App.Other.Customer` (no edge), jagged arrays/tuples/nullable (correct element type), `Box<Customer>` where `Box` is also declared here (two edges, both defensible).

**One prose over-claim:** *"That one check is what drops … open type parameters `T`"* — it drops `T` only because no same-file class happens to be named `T`. Case 2 is the counter-example; the sentence claims a property the guard does not have.

### Mutant table — 5 of my own, all in the MORE PERMISSIVE direction, none overlapping M1–M4

Each applied from a `shasum`-verified backup, occurrence count asserted `== 1` (RM3: the guard is a 2-occurrence symbol, both inside `attachCsharpFieldTypeRefs`), enclosing function asserted, `go vet` clean before every score, `-count=1`, green baseline immediately before each, restored by `cp` with sha re-compared (`98fd6489…` before and after every score).

| mutant | site | verdict |
|---|---|---|
| **RM1** delete the `r.SourceFile != filePath` guard | `csInFileTypeTargets` | **ALIVE — not production-reachable.** `Extract` is per-file and every record carries `file.Path`; the guard can never fire. Dead defensive code. |
| **RM2** delete `r.Kind != "SCOPE.Schema" \|\| r.Subtype != "field"` | `attachCsharpFieldTypeRefs` | **ALIVE — equivalent.** `csFieldTypeRefsMetaKey` has exactly one write site (`field_members.go:81`) and it is on a field record. |
| **RM3** delete the `emitted[t.toID]` dedup | `attachCsharpFieldTypeRefs` | **ALIVE — distinguishable and ungraded.** `public (Customer, Customer) Pair` → 2 identical edges vs 1 (verified against the parent binary on the same input). **Grading cost: one field line in `fieldTypeRefsSrc`; the `want` literal is unchanged** — a duplicate is a length mismatch. ~1 line. |
| **RM4** drop `alias_qualified_name` from the early return | `csTypeRefCandidates` | **ALIVE — distinguishable and ungraded.** `public global::Customer G` gains `Order.G -> …:Customer` (verified). **Grading cost: 1 field line + 0 want lines** (it must produce nothing). |
| **RM5** *(sensitivity control, not scored)* bare-name fallback for unresolved candidates | `attachCsharpFieldTypeRefs` | **DEAD ×2 layers** — 4 package tests + 2 `cmd/grafel` GateOff tests. This is the evidence for (c). |

RM3/RM4 are ~2 lines total to close and I'd take them in this PR; I am not blocking on them.

### Verdict

**APPROVE.** The three decisions (kind, address, same-file-only) are each backed by a measurement I reproduced rather than relayed; all three corpus figures are exact; the resolver gate is a real production-path probe; the negative direction is asserted by name; and I could not construct an over-firing case that occurs on any of the three real corpora. Follow-ups to file, none blocking: **(i)** the custom lane's `Class:<Target>` dangles under duplicate model names in golang/javascript/ruby/java (Python alone is masked at `refs.go:4574`); **(ii)** namespace scope and type-parameter shadowing are not consulted by the same-file check; **(iii)** RM3/RM4 are 2 lines of grading.

cajasmota:
## Coordinator mutant CF-1 on `8f7cdd12b` — ALIVE, but not production-reachable

Own detached worktree, `git status --porcelain` empty at start and after restore, restored by `cp` from a SHA-256-verified backup (`d4464d85…` both ways), zero `MUTANT` text remaining, baseline green immediately before. macOS/APFS.

**CF-1 — remove the same-file ENFORCEMENT, not the same-file lookup.** Nine mutants have been scored here and all of them attack the *consumer* side of the rule: whether an unresolved name still emits, whether qualified names are descended, which address is used, whether collisions dedup. **Nobody attacked the line that decides which records are eligible to be targets in the first place** — `field_type_refs.go:173`, inside `csInFileTypeTargets`:

```go
-  if r.SourceFile != filePath {
-      continue
-  }
+  _ = filePath // MUTANT CF-1: same-file enforcement removed
```

Edit proven landed: exactly 1 occurrence, hit set asserted to a single line, and the **enclosing function asserted to be `csInFileTypeTargets`** rather than `attachCsharpFieldTypeRefs` — which matters, because both take `(records, filePath)` and a naive replace could have drifted between them.

**Result: `./internal/extractors/csharp/` green, exit 0 — ALIVE.**

## Why the verdict is *not production-reachable*, and how I established it

`csharp.go:99` calls `attachCsharpFieldTypeRefs(entities, file.Path)` from inside
`func (e *Extractor) Extract(_ context.Context, file extractor.FileInput)`. `entities` is local to that call and is populated only by walking **that one file**, so every record already satisfies the guard.

The one way that could be false is a *shared* node minted with a foreign `SourceFile` — this package emits several (`emitConfigConsumerEdges`, `emitExceptionFlowEdges`). So I checked rather than assumed: **every `SourceFile:` assignment in the package is `file.Path` / `filePath` / `f.Path`**, with no exceptions, and the two shared-node kinds (`SCOPE.Config`, `SCOPE.ExceptionType`) do not match the switch anyway, which admits only `SCOPE.Enum` and `SCOPE.Component` with subtype class/interface/struct/type.

So the guard is **defensive against a caller that does not exist**. That is a legitimate reason for a mutant to live, and it is not a defect in this PR.

## What I am recording rather than requesting

The grading price is about **five lines** — `csInFileTypeTargets` is a plain function taking `([]types.EntityRecord, string)`, so a direct unit call with a two-file record slice grades it without any fixture machinery. I am **not** asking for it and I am merging without it, because pinning a guard against a caller that does not exist is worth less than a third round-trip on a PR that has already had two.

**But the price is on the record for the moment it stops being defensive.** If anyone ever moves this pass to a batch caller — one that hands it records from more than one file — this guard becomes the only thing standing between the pass and cross-file emission, and today nothing would fail if it were removed. The address makes that worse rather than better: `BuildComponentStructuralRef("csharp", filePath, r.Name)` is built from the **field's** file, so a leaked cross-file target would not merely be wrong, it would be addressed to a file that never declared it.

## On the rest of the round

RM3 and RM4 graded and now DEAD, taking the reviewer's two-line pricing rather than arguing it — the tuple case pins the dedup by a length mismatch, and `global::Customer` pins `alias_qualified_name` as a named forbidden field. The fixture no longer implies coverage it lacks: `_SameNameDeclaredTwiceInOneFile` moved to **partial classes**, which keeps the collide branch and drops the namespace implication entirely, and both over-fires are now pinned as **known-wrong** with comments saying a fix is expected to break them. That is the right way to carry a known limitation — the test tells the next person what the state of the world is, instead of a comment claiming a property nothing checks.

Merging on green.

Issue #6983, opened 2026-09-07
REFUTED (by me, after filing): 1 MiB source truncation is unreachable — classifier.maxIndexableBytes excludes those files with the same constant; the real loss is exclusion, and SkipReason=too_large has no consumer


cajasmota:
Found while merging #6982 (for #6969), **verified in source in the main checkout at `bd03cdfeb` before filing**.

#6969 fixed silent truncation for one 8 MiB git-metadata file that nothing reachable actually exceeds. **The same silence sits on eleven call sites that read SOURCE files at 1 MiB**, where the population is not exotic at all — and #6982 just built the mechanism to observe it.

## The population

`internal/links/safe_source_read.go:74`'s `readSourceFile` is `safeio.ReadFile`, which truncates at the bound and returns a nil error. Ten passes call it with `maxSourceFileBytes = 1 << 20`:

```
complexity_pass.go:69      dataflow_pass.go:119, :343   taint_flow.go:124
def_use_pass.go:110        payload_drift.go:355         effect_propagation.go:117
constant_propagation.go:365  template_pattern_pass.go:80  reachability.go:319
```

and the engine twin does the same at `internal/engine/http_endpoint_substrate_fold.go:302` with `substrateMaxFileBytes = 1 << 20` (deliberately the same value — #6823 exists because the two resolvers diverged once already).

`string_pass.go:357` is **not** in this population and should not be lumped in: `stringScanMaxFileBytes` is `4 MiB + 1` precisely so an over-length file still arrives over-length and is still discarded by the pass's existing check. That one is reasoned correctly and is a pure memory saving.

## This is stated, not hidden — and that is the point

`safe_source_read.go:44-52` says it outright:

> The cost is real and is inherited whole. `safeio.ReadFile` TRUNCATES at the bound (`io.LimitReader`) rather than failing, so a declaring file larger than 1 MiB is parsed from its first megabyte only: a constant declared past that point is **silently not found, with no telemetry**.

The trade was accepted for a stated reason — a bound is not optional, and base-URL constants and handler bodies live in small modules. **I am not disputing the bound.** The claim here is narrower: *"with no telemetry"* was a constraint of the API at the time, and as of `bd03cdfeb` it no longer is. `safeio.ReadFileLimited` returns an observed truncation bool, and `internal/gitmeta/saferead.go` shows the wrapper shape for adopting it in one line.

## Why this is worth more than #6969 was

[...]

PR #6987, opened 2026-09-07
fix(#6985): an oversized file was dropped from the graph with no name and no reason — the loss is exclusion, and the comment that said otherwise caused a phantom issue

### LINKED ISSUE #6985: An oversized file is dropped from the graph with no user-visible name or reason — the tally filter is correct, the missing piece is a gitmeta-style bounded report
Successor to #6983, which I filed and then refuted. That issue chased a **truncation** that cannot happen; the measurement that refuted it found the real loss on the same axis, and this is it.

**Verified in source in the main checkout at `bd03cdfeb` before filing** — including the parts that make the claim *narrower* than it first looks.

## The loss

`internal/classifier/classifier.go:172` skips any file over `maxIndexableBytes` (`= 1 MiB`, `classifier.go:27`) with `SkipReason: "too_large"`. The file **mints no entity and appears nowhere in `graph.json`**. Confirmed end-to-end: `big.js` at 2,097,293 B with a declaration at byte 1,572,928 → `skipped=1`, zero entities, zero mentions of `big.js` in the graph; the control `small.js` at 962,718 B with the identical declaration → found.

This is a real capability boundary and I am **not** proposing to move it. The question is whether a user can tell it happened to a file they care about.

## What is NOT the defect — stating this because I got the neighbouring issue wrong

- **`UnsupportedTally` ignoring `too_large` is deliberate and documented.** `unsupported.go:278-281` says so in terms: *"Everything that is not an unsupported-LANGUAGE skip is ignored: indexed files, vendored files, binaries, oversized files…"*. That filter is correct for what the tally is for. **It is not an oversight and should not be "fixed" by widening it** — that would change what the unsupported-extension report means.
- **The reason is not thrown away entirely.** `classifier.go:89` and `:117` attach `classifier.skip_reason` as an OTel span attribute. So there is telemetry; it just requires tracing to be enabled and is not a thing a user reads after an index.

## What the defect actually is

**There is no user-visible surface that names an oversized file.** `grep` for the reporting shape finds skip messages for commit-couple (`index.go:2707`), coupling (`:2729`), shared-db (`:2772`) and coverage (`:2865`) — but nothing equivalent for a classifier size skip. The user gets an aggregate `skipped=N` with no reason and no filename, for a file that has been silently dropped from the graph.

Compare `internal/gitmeta`, which after #6982 has exactly the shape this wants: `reportGitMetaSkip` and `reportGitMetaTruncation`, sharing a dedup map and a cap so the output stays bounded. That is a working precedent in-tree for "announce a bounded list of dropped inputs".

## Why this is worth doing and also worth keeping small

The population is **not latent** — unlike #6983's, it has named instances today. Across all 60 corpora at `/Users/jorgecajas/Projects/archigraph-corpora` (27,335 files, macOS/APFS): **20 files over 1 MiB, 10 of them with a recognised language** — 8 vendored/compiled JS bundles plus `pnpm-lock.yaml` in `nextjs`, and `openapi-stripe`'s `openapi.yaml`. `django`, `etcd`, `staxrip` and `aspnetcore-mvc` have none at all.

**Flagged as corpus-relative, not a base rate** — a zero on our corpora never means "cannot occur", and the interesting case is precisely the user with a large generated file they expected to be indexed. But it also means this is a reporting improvement, not an emergency.

1. **Which surface.** A stderr line per file does not scale to a repo of vendored bundles; `gitmeta`'s dedup-map-plus-cap is the shape that survives. Decide deliberately and say why.
2. **That it observes the artefact.** A test must assert the **emitted report for a named oversized file**, not a counter the code keeps about itself. A diagnostic signature is not a regression pin.
3. **That the `UnsupportedTally` filter is left alone**, and that whatever is added does not change what the unsupported-extension report counts. Assert the non-change.
4. **The `skipped=N` aggregate should still reconcile.** If a file is now named, it must not also be double-counted.

## The other residual from #6983, recorded here so it is not lost

**TOCTOU**: a file that grows past 1 MiB between the index-time `os.Stat` and a later group-link read. Nobody has characterised it, it is genuinely not covered by the size-gate argument that closed #6983, and it is the one path on which a real (not hypothetical) truncation could still occur. Out of scope for this issue; noted so it stays on the record.

Refs #6983, #6982, #6969, #6966.

cajasmota:
Closes #6985.

Two deliverables on one axis: the oversized file gets a bounded, user-visible report, and the comment that made #6983 look like a live defect is rewritten to say what is true.

Branched from `bd03cdfeb`. Everything below was run in the worktree at `/private/tmp/claude-501/-Users-jorgecajas-Projects-archigraph/c0fefed4-0195-4c81-9a94-07a7145710a4/scratchpad/impl-6985-k4m7p/wt` (macOS Darwin 25.6 / APFS), never in the main checkout.

**`3e266c957` → `d86b5fba6` answers the REQUEST-CHANGES review; every item was confirmed and taken.** Summary of what changed is at the bottom.

## Part 1 — the report

`internal/classifier/oversized_report.go`: an `oversizedReporter` (mutex, `seen` map capped at `maxOversizedReports = 8`, a `suppressed` counter, an optional writer) hangs off the `Classifier`, and `classifyWithSizeInner`'s size gate calls `c.oversized.report(...)`. The `ClassifyResult` is byte-identical to before.

```
grafel: <path> is <n> bytes, over the 1048576-byte limit, and was NOT indexed — it mints no entity and appears nowhere in the graph (#6985)
grafel: further oversized-file skips are not named; the total is reported at the end of the index   # once, on the 8th
grafel: 32 further oversized-file skips were not named (8 named, 40 over the 1048576-byte limit in total) (#6985)
```

**Which surface, and why.** stderr, always on, dedup + cap, shaped after `internal/gitmeta`'s `reportGitMetaSkip` / `reportGitMetaTruncation` (`internal/gitmeta/saferead.go:121,157`). Bounded output is the requirement: the population is a repo full of vendored bundles (8 of the 10 recognised-language oversized files across the corpora are compiled/vendored JS, 9 of 10 in one repo — **quoted from #6983's measurement, not re-measured here**), so a line per file is the shape to avoid.

**Lifetime: per-Classifier, i.e. per index.** The first cut kept the state process-global, and the reviewer measured the consequence: run 2 in the same process printed **nothing at all**, for a different repo too, not even the suppression notice — a user who fixed nothing and re-ran saw a clean report, which is worse than no report. `gitmeta` is not a precedent for that: its inputs are read once per process, while this code re-enters per index **and** per daemon incremental reindex (`internal/extractors/incremental.go:1003`). Every construction site (`cmd/grafel/index.go:808`, `internal/daemon/extract/coordinator.go:662`, `subproc.go:105`, `internal/extractors/incremental.go:886`) calls `classifier.New` per run, so per-Classifier state gets the right lifetime **with no reset call for a caller to forget** — chosen over an exported `ResetOversizedReport()` for exactly that reason.

**Why at the mint point.** Within this package `"too_large"` is minted only in `classifyWithSizeInner`, while **four** call sites consume it (the four above). Reporting at each consumer is the shape that let eleven copies of one read diverge in #6823. (Scoped as the reviewer asked: `internal/secrets/secrets.go:410` declares its own unrelated `SkipTooLarge`; the claim is package-local, not repo-wide.)

**The cap now carries a count.** `ReportOversizedSummary` prints the hidden total and is called from **one** place — `Indexer.Run`, immediately before the `processed=… skipped=…` line — because that is the surface a user reads after an index. The daemon's coordinator/subproc/incremental paths print no run summary at all, so a trailing count there would land nowhere; their named lines are unaffected. Nothing is printed when nothing was suppressed. Cost stated in the source: past the cap a path is not remembered, so a repeated unnamed file would count twice; on the index path each file is classified once, and the alternative is an unbounded map keyed by exactly the population this report exists because it can be large.

**`UnsupportedTally` is untouched** — no line of `internal/classifier/unsupported.go` changed. **`skipped=N` still reconciles** — the report is additive and returns.

### Tests — all observe the emitted report

`internal/classifier/oversized_report_6985_internal_test.go` (each test builds its own `Classifier`, so a fresh instance *is* the starting state — the previous version's order-coupling on a shared map is gone):

| test | asserts |
|---|---|
| `…ReportNamesTheFile` | the line carries the **path**, size, limit, consequence |
| `…SilentAtAndUnderTheLimit` | 1 MiB exactly and 1 KiB → **no output**, both indexed |
| `…NamesEachDistinctFile` | two distinct paths → two named lines |
| `…DedupsRepeatsWithoutChangingTheSkip` | one path twice → one line, two skips |
| `…IsCapped` | 9 files → 8 named, 9th absent, the notice **exactly once**, and total lines **== cap + 1** |
| `…SummaryCountsWhatTheCapHid` | 40 files → one summary line naming 32 hidden / 8 named / 40 total |
| `…SummarySilentWhenNothingWasSuppressed` | cap never reached → no summary line |
| `…IsPerIndexNotPerProcess` | two classifiers, **one writer, no reset** → the second index names a never-seen path *and* a re-indexed one |
| `…IsSafeUnderConcurrentClassification` | 32 goroutines over 4 paths, locking writer → each named exactly once, 4 lines total (run under `-race`) |
| `…LeavesUnsupportedTallyUnchanged` | `.pas` control `{".pas":1}`; three oversized files leave `Counts()` `DeepEqual` to it |
| `…SkipReasonIsUnchangedFor…` | oversized `.pas` is still `too_large`, `Language == ""` |

`cmd/grafel/oversized_report_6985_test.go` is new and observes the **production index path**: a temp repo with 9 oversized `.js` files run through `Indexer.Run` with `os.Stderr` captured, asserting 8 named files plus `1 further oversized-file skips were not named` and `9 over the 1048576-byte limit in total`.

### Mutants — 13 applied, 13 DEAD

Each applied with an asserted exact pre-edit occurrence count, post-edit counts printed, and the enclosing function named; `go vet` exit 0 for every one; scored `-count=1`; restored by `cp` from `shasum`-verified backups, never `git checkout`. The reviewer's four ungraded seams are N1/N2/N8/N13 below.

| # | mutation (enclosing func) | result |
|---|---|---|
| M1 | delete the report call (`classifyWithSizeInner`) | **DEAD** — 8 tests |
| M2 | line without the filename (`report`) | **DEAD** — 7 tests |
| M3 | constant dedup key, distinct files suppressed (`report`, 7 anchors) | **DEAD** — 5 tests |
| M4 | `maxOversizedReports = 0` (file scope) | **DEAD** — 7 tests |
| M5 | report before the size gate (`classifyWithSizeInner`) | **DEAD** — `…SilentAtAndUnderTheLimit` |
| M6 | widen `UnsupportedTally.Observe` to fold in `too_large` (`Observe`) | **DEAD** — the new non-change test **and** the pre-existing `TestUnsupportedTally_OtherSkipReasonsNotCounted/too-large` |
| M7 | cap boundary `>=` → `>` (`report`) | **DEAD** — 2 tests |
| **N1** | print the notice **once per suppressed file** (`report`; notice sites 1→2) | **DEAD** — `…IsCapped` (line-count assertion) |
| **N2** | **delete the mutex entirely** (`report`, `summary`, `setOversizedReportOutput`; locks 2→0, unlocks 4→0, field and `sync` import removed) | **DEAD under `-race`** — 4 DATA RACE reports + `…IsSafeUnderConcurrentClassification`. **Honest caveat: WITHOUT `-race` it survived 5/5 runs** — the concurrent-map-write fatal is scheduling-dependent. The kill's premise is `-race`, and `.github/workflows/test.yml:404-408` runs `-race` only on a tag push, a dispatch with `race=true`, or the **`ci:full` label** — so this PR wants that label (which is also what runs the macOS leg). |
| **N8** | share `seen`/`suppressed` process-wide again — the reviewed defect (`report`, `summary`; `r.seen` 7→0) | **DEAD** — `…IsPerIndexNotPerProcess` + 4 others |
| N9 | `summary` returns unconditionally (`summary`) | **DEAD** — `…SummaryCountsWhatTheCapHid` |
| N10 | `summary` prints a zero row (`summary`) | **DEAD** — 3 tests |
| N11 | summary reports `named+hidden` where it should report `hidden` (`summary`) | **DEAD** — `…SummaryCountsWhatTheCapHid` |
| N12 | ignore the writer seam, always stderr (`writer`) | **DEAD** — 8 tests |
| **N13** | delete `i.classifier.ReportOversizedSummary()` (`Indexer.Run`) | **ALIVE before the new cmd/grafel test** (whole `./cmd/grafel/` suite green, 152s) → **DEAD after it**. That is why the end-to-end test was written rather than priced. |

N3/N4 from the review (the two mutually-masking `oversizedSeen = nil` resets) **no longer exist**: per-Classifier state has no global to reset, so there is no compound guard left to grade part by part.

## Part 2 — the `safe_source_read.go` comment

Rewritten, and **a comment-only diff has no mutant — every mutation of it is equivalent by construction — so each sentence was fact-checked against source instead.** Verified, with locations:

- `safeio.ReadFile` truncates: it is `ReadFileLimited` minus the bool — `internal/safeio/safeio.go:265-268`, `:309-311`. The stale *"with no telemetry"* clause is gone (`ReadFileLimited` exists at `:285`).
- `classifier.maxIndexableBytes` is `1 * 1024 * 1024` (`internal/classifier/classifier.go:27`), the same value as `maxSourceFileBytes` (`internal/links/safe_source_read.go`) and `substrateMaxFileBytes` (`internal/engine/http_endpoint_substrate_fold.go:92`). **None of the three constants is changed here.**
- The deciding gate: `classifier.go:172-174`, before language detection.
- **Four** `ClassifyWithSize` consumers, not three — the review caught the undercount, and #6983 had listed the fourth: `cmd/grafel/index.go:3987`, `coordinator.go:673`, `subproc.go:277`, `internal/extractors/incremental.go:1003`.
- **Two residuals, not one.** Alongside TOCTOU, all three index sites initialise `size := int64(-1)` and leave it there when `os.Stat` fails (`cmd/grafel/index.go:3980`, `coordinator.go:669`, `subproc.go:273`); `-1 > maxIndexableBytes` is false, so an oversized file whose stat fails passes the gate, mints an entity, reaches `readSourceFile` and **is** truncated — deterministic, no concurrency needed, no report. The comment now says the exclusion argument holds **on the normal path** and names both residuals as uncharacterised. Writing "THE ONE RESIDUAL" was the same defect class the rewrite exists to remove.
- The per-pass claim ("every pass here iterates a file set built from entity `SourceFile`") is still **carried over from #6983's enumeration, not independently re-derived** — not a second confirmation.

## Verification

Custom-gate state (#6966): **gate-OFF** for every figure — `go test` and `scripts/quality/run.sh --ratchet` neither enter `WithCustomExtractors`.

- `go vet ./internal/classifier/ ./internal/links/ ./cmd/grafel/` — exit 0; `gofmt -l` on the touched dirs — empty.
- `go test -count=1 ./internal/classifier/ ./internal/links/ ./cmd/grafel/ ./internal/cli/ ./internal/daemon/extract/` — **exit 0**, all five `ok` (`cmd/grafel` 149s, whole-graph digest pin).
- `go test -count=1 -race ./internal/classifier/ ./internal/links/` — **exit 0**.
- `scripts/quality/run.sh --ratchet` (isolated `GRAFEL_HOME` + `GRAFEL_DAEMON_ROOT`) — **exit 0**: *"quality ratchet OK — 38 gated fixtures held their baseline (29 at 100% must-have recall)"*. No fixture moved; no baseline constant touched.

**What I did NOT measure:** no corpus index — the population figures are quoted from #6983. Neither residual is characterised: I did not measure how often a stat fails on an oversized file, nor the TOCTOU window. I did not measure daemon behaviour end-to-end; the per-index lifetime claim rests on the four `classifier.New` construction sites plus the unit test above, not on a running daemon.

## What changed after the review (`3e266c957` → `d86b5fba6`)

| review item | disposition |
|---|---|
| (a) process-global dedup+cap goes permanently, globally silent | **fixed** — state is per-Classifier; pinned by `…IsPerIndexNotPerProcess` (and N8) |
| (b) `"the single place too_large is minted"` is package-local, not repo-wide | **scoped** in the source comment |
| (c) the cap never says how much it hid | **fixed** — `ReportOversizedSummary` + the `cmd/grafel` end-to-end test |
| (d1) "all three extraction entry points" — there are four | **fixed** in both comments |
| (d2) "THE ONE RESIDUAL IS TOCTOU" — the `size = -1` stat bypass is a second | **fixed**; named as a residual with its three sites |
| N1 suppression line ungraded | **graded** — DEAD |
| N2 mutex ungraded, ALIVE even under `-race` | **graded** — DEAD under `-race` (see the caveat in the table; wants the `ci:full` label) |
| N3/N4 mutually-masking resets | **dissolved** — no global state, no resets |
| the `reflect.DeepEqual` tally assertion passes #6975 | unchanged |

### DIFF
--- a/cmd/grafel/index.go
+++ b/cmd/grafel/index.go
@@ -2904,6 +2904,10 @@ func (i *Indexer) Run(ctx context.Context, absRepo string) (*graph.Document, err
 	trk.Done(len(files), doc.Stats.Entities)
 
 	dur := time.Since(start)
+	// #6985 — close the oversized-file report before the aggregate line, so a
+	// user who saw eight names also learns how many more there were. Prints
+	// nothing unless the cap actually hid something.
+	i.classifier.ReportOversizedSummary()
 	fmt.Fprintf(os.Stderr,
 		"grafel: processed=%d extracted=%d skipped=%d failed=%d "+
 			"entities=%d relationships=%d "+
--- a/cmd/grafel/oversized_report_6985_test.go
+++ b/cmd/grafel/oversized_report_6985_test.go
@@ -1,0 +1,99 @@
+package main
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+)
+
+// #6985 — the oversized-file report and its end-of-index total, observed on
+// the PRODUCTION index path rather than in a classifier unit test.
+//
+// This test exists because the summary call is one line of wiring in
+// Indexer.Run: deleting it left every classifier-package test green, so
+// nothing graded whether a user actually gets the total. The named lines are
+// asserted here too, for the same reason — they are emitted from inside the
+// worker pool, which no unit test exercises.
+//
+// NOTE: this test swaps the process-global os.Stderr handle and therefore MUST
+// NOT call t.Parallel(), for the reason spelled out on
+// TestExternalSynthesis_VerboseCounter.
+func TestOversizedFilesAreNamedAndTotalledByTheIndexer_6985(t *testing.T) {
+	repo := t.TempDir()
+
+	// One indexable file so the run has real work, plus enough oversized ones
+	// to overflow the report's cap.
+	if err := os.WriteFile(filepath.Join(repo, "app.js"),
+		[]byte("export function handler(req, res) { return res.send('ok'); }\n"), 0o644); err != nil {
+		t.Fatalf("write app.js: %v", err)
+	}
+	const oversizedCount = 9
+	padding := strings.Repeat("x", 1<<20) // one byte of prelude puts it over 1 MiB
+	for i := range oversizedCount {
+		name := fmt.Sprintf("bundle_%d.js", i)
+		body := fmt.Sprintf("// %s\n/*%s*/\n", name, padding)
+		if err := os.WriteFile(filepath.Join(repo, name), []byte(body), 0o644); err != nil {
+			t.Fatalf("write %s: %v", name, err)
+		}
+		if st, err := os.Stat(filepath.Join(repo, name)); err != nil || st.Size() <= 1<<20 {
+			t.Fatalf("fixture degenerate: %s is %v bytes (err %v), want > 1 MiB", name, st.Size(), err)
+		}
+	}
+
+	r, w, err := os.Pipe()
+	if err != nil {
+		t.Fatalf("pipe: %v", err)
+	}
+	origStderr := os.Stderr
+	os.Stderr = w
+	done := make(chan string, 1)
+	go func() {
+		var buf []byte
+		tmp := make([]byte, 4096)
+		for {
+			n, err := r.Read(tmp)
+			if n > 0 {
+				buf = append(buf, tmp[:n]...)
+			}
+			if err != nil {
+				break
+			}
+		}
+		done <- string(buf)
+	}()
+
+	idx := newTestIndexer(t, "oversized_repo", nil, "")
+	_, runErr := idx.Run(context.Background(), repo)
+
+	os.Stderr = origStderr
+	w.Close()
+	got := <-done
+	if runErr != nil {
+		t.Fatalf("Run: %v\nstderr:\n%s", runErr, got)
+	}
+
+	// Exactly the cap's worth of files is named…
+	named := 0
+	for i := range oversizedCount {
+		if strings.Contains(got, fmt.Sprintf("bundle_%d.js is ", i)) {
+			named++
+		}
+	}
+	if named != 8 {
+		t.Errorf("indexer named %d oversized files, want 8 (the cap); stderr:\n%s", named, got)
+	}
+	if !strings.Contains(got, "was NOT indexed") {
+		t.Errorf("no oversized-skip line on the production path; stderr:\n%s", got)
+	}
+	// …and the end-of-index line says how many it did not name. Without the
+	// ReportOversizedSummary call in Run this is the assertion that fails.
+	if !strings.Contains(got, "1 further oversized-file skips were not named") {
+		t.Errorf("no end-of-index total for the suppressed oversized files; stderr:\n%s", got)
+	}
+	if !strings.Contains(got, "9 over the 1048576-byte limit in total") {
+		t.Errorf("the total does not say how many files were over the limit; stderr:\n%s", got)
+	}
+}
--- a/internal/classifier/classifier.go
+++ b/internal/classifier/classifier.go
@@ -168,8 +168,13 @@ func (c *Classifier) classifyWithSizeInner(filePath string, sizeBytes int64) Cla
 		return ClassifyResult{Skip: true, SkipReason: "empty_path"}
 	}
 
-	// Size check first — cheapest guard.
+	// Size check first — cheapest guard. The skip is announced (#6985): an
+	// oversized file mints no entity and appears nowhere in graph.json, and
+	// before the report the only trace of it was the aggregate `skipped=N`.
+	// See oversized_report.go for why the report lives here and why the
+	// UnsupportedTally is not the surface for it.
 	if sizeBytes > maxIndexableBytes {
+		reportOversizedSkip(filePath, sizeBytes)
 		return ClassifyResult{Skip: true, SkipReason: "too_large"}
 	}
 
--- a/internal/classifier/classifier.go
+++ b/internal/classifier/classifier.go
@@ -53,6 +53,10 @@ type ClassifyResult struct {
 type Classifier struct {
 	// tracer is the OTel tracer used for classify spans.
 	tracer trace.Tracer
+
+	// oversized carries this classifier's oversized-skip report state. It is
+	// per-Classifier, i.e. per index, on purpose — see oversized_report.go.
+	oversized oversizedReporter
 }
 
 // New constructs a Classifier.
@@ -174,7 +178,7 @@ func (c *Classifier) classifyWithSizeInner(filePath string, sizeBytes int64) Cla
 	// See oversized_report.go for why the report lives here and why the
 	// UnsupportedTally is not the surface for it.
 	if sizeBytes > maxIndexableBytes {
-		reportOversizedSkip(filePath, sizeBytes)
+		c.oversized.report(filePath, sizeBytes)
 		return ClassifyResult{Skip: true, SkipReason: "too_large"}
 	}
 
--- a/internal/classifier/oversized_report.go
+++ b/internal/classifier/oversized_report.go
@@ -25,71 +25,136 @@ import (
 // report counts, which is a different report about a different fact. This is a
 // separate surface for a separate disposition.
 //
-// WHY HERE rather than at the extraction entry points. Three call sites
-// classify with a size (cmd/grafel/index.go, internal/daemon/extract's
-// coordinator and subproc). Reporting at each of them is the shape that let
-// eleven copies of one read diverge in #6823; classifyWithSizeInner is the
-// single place "too_large" is minted, so guarding it there is by construction.
+// WHY HERE rather than at the consuming call sites. Within this package
+// "too_large" is minted in exactly one place, classifyWithSizeInner, while
+// FOUR call sites consume it: cmd/grafel/index.go, internal/daemon/extract's
+// coordinator and subproc, and internal/extractors/incremental.go:1003 on the
+// daemon's watcher-driven incremental path. (internal/secrets declares its own
+// unrelated SkipTooLarge constant; that is a different subsystem and not this
+// mint point.) Reporting at each consumer is the shape that let eleven copies
+// of one read diverge in #6823; guarding the mint point is by construction.
+//
+// WHY THE STATE IS PER-CLASSIFIER. It was process-global in the first cut of
+// this file, and that made the report go permanently and globally silent after
+// the first eight oversized files: the second index in the same process — a
+// different repo included — printed nothing at all, not even the suppression
+// line, so a user who fixed nothing and re-ran saw a clean report. internal/
+// gitmeta can hold process-global state because its inputs are read once per
+// process; this code re-enters per index AND per daemon incremental reindex.
+// Every construction site (cmd/grafel/index.go:808,
+// internal/daemon/extract/coordinator.go:662, subproc.go:105,
+// internal/extractors/incremental.go:886) calls classifier.New per run, so
+// hanging the state off the Classifier makes the lifetime per-index with no
+// reset call for a caller to forget.
 
-// oversized* back the always-on report below. Shape taken from
-// internal/gitmeta's reportGitMetaSkip / reportGitMetaTruncation: a dedup map
-// plus a cap, so the output stays bounded.
-var (
-	oversizedMu   sync.Mutex
-	oversizedSeen map[string]bool
-	oversizedOut  io.Writer = os.Stderr
-)
-
-// maxOversizedReports caps the report the way maxGitMetaSkipReports caps its
-// own: a warning long enough to scroll past reports nothing.
+// maxOversizedReports caps the NAMED lines, the way maxGitMetaSkipReports caps
+// internal/gitmeta's: a warning long enough to scroll past reports nothing.
 //
 // The cap is load-bearing here rather than a backstop. The population is
 // exactly a repo full of vendored bundles — of the ten oversized files with a
 // recognised language across grafel's 60 corpora, eight are compiled/vendored
 // JS and nine of the ten are in a single repo — so a line per file is the
-// output shape this report has to avoid. A bounded list plus the existing
-// `skipped=N` aggregate is what a reader can act on.
+// output shape this report has to avoid.
 const maxOversizedReports = 8
 
-// setOversizedReportOutput redirects the report for tests and returns a
-// restore func. Test-only helper.
-func setOversizedReportOutput(w io.Writer) func() {
-	oversizedMu.Lock()
-	prev := oversizedOut
-	oversizedOut = w
-	oversizedSeen = nil
-	oversizedMu.Unlock()
-	return func() {
-		oversizedMu.Lock()
-		oversizedOut = prev
-		oversizedSeen = nil
-		oversizedMu.Unlock()
-	}
+// oversizedReporter is one index's worth of oversized-skip reporting state.
+//
+// The mutex is not decoration: ClassifyWithSize runs inside the indexer's
+// worker pool (cmd/grafel/index.go), so an unguarded map write here is a
+// concurrent map write — a Go runtime fatal, not a warning.
+type oversizedReporter struct {
+	mu sync.Mutex
+	// seen holds the paths already NAMED, and is capped at
+	// maxOversizedReports so its memory is O(cap) rather than O(oversized
+	// files in the repo).
+	seen map[string]bool
+	// suppressed counts the skips that were not named because the cap was
+	// already full. It is what lets the end-of-index line say how much was
+	// cut: "8 of 40" and "8 of 9" are indistinguishable without it.
+	suppressed int
+	// out is nil in production and is os.Stderr then; tests redirect it.
+	out io.Writer
 }
 
-// reportOversizedSkip names a file that was dropped for exceeding the size
-// gate, and says what the consequence is.
+// writer resolves the destination. Callers hold r.mu.
+func (r *oversizedReporter) writer() io.Writer {
+	if r.out == nil {
+		return os.Stderr
+	}
+	return r.out
+}
+
+// setOversizedReportOutput redirects one classifier's report. Test-only
+// helper: production never sets it, so the zero value writes to stderr.
+func setOversizedReportOutput(c *Classifier, w io.Writer) {
+	c.oversized.mu.Lock()
+	c.oversized.out = w
+	c.oversized.mu.Unlock()
+}
+
+// report names a file that was dropped for exceeding the size gate, and says
+// what the consequence is.
 //
 // It reports and returns; it does not change the ClassifyResult, so the
 // `skipped=N` aggregate still counts a named file exactly once, as a skip. The
-// dedup map is keyed on the file path, so two distinct oversized files are two
+// dedup is keyed on the file path, so two distinct oversized files are two
 // lines and the same file classified twice is one.
-func reportOversizedSkip(filePath string, sizeBytes int64) {
-	oversizedMu.Lock()
-	if oversizedSeen == nil {
-		oversizedSeen = map[string]bool{}
+func (r *oversizedReporter) report(filePath string, sizeBytes int64) {
+	r.mu.Lock()
+	if r.seen == nil {
+		r.seen = map[string]bool{}
 	}
-	if oversizedSeen[filePath] || len(oversizedSeen) >= maxOversizedReports {
-		oversizedMu.Unlock()
+	if r.seen[filePath] {
+		r.mu.Unlock()
 		return
 	}
-	oversizedSeen[filePath] = true
-	last := len(oversizedSeen) == maxOversizedReports
-	w := oversizedOut
-	oversizedMu.Unlock()
+	if len(r.seen) >= maxOversizedReports {
+		// Past the cap the path is not remembered, so a repeat of an unnamed
+		// file counts twice. On the index path each file is classified once,
+		// and the alternative is an unbounded map keyed by a population this
+		// report exists because it can be large.
+		r.suppressed++
+		r.mu.Unlock()
+		return
+	}
+	r.seen[filePath] = true
+	last := len(r.seen) == maxOversizedReports
+	w := r.writer()
+	r.mu.Unlock()
 
 	fmt.Fprintf(w, "grafel: %s is %d bytes, over the %d-byte limit, and was NOT indexed — it mints no entity and appears nowhere in the graph (#6985)\n", filePath, sizeBytes, maxIndexableBytes)
 	if last {
-		fmt.Fprintf(w, "grafel: further oversized-file skips suppressed after %d\n", maxOversizedReports)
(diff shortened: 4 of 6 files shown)


cajasmota:
# Independent adversarial review — **REQUEST-CHANGES**

Tree measured: `/private/tmp/claude-501/-Users-jorgecajas-Projects-archigraph/c0fefed4-0195-4c81-9a94-07a7145710a4/scratchpad/review-6987-adv9/wt` at `3e266c957`, plus a parent checkout at `bd03cdfeb` in `.../scratchpad/parent-wt`. macOS Darwin 25.6 / APFS. Never the main checkout. `git status --porcelain` empty at the end; every mutant restored by `cp` from a `shasum`-verified backup (`d65de41403efcbeae609cb6dec98af0e04919ca0`), never `git checkout`.

**Custom-gate state (#6966):** every `go test` / `go vet` / ratchet figure below is **gate-OFF**. The two end-to-end `grafel quality` runs are **gate-ON** (`quality` is the one non-test caller of `WithCustomExtractors`); that does not qualify the result, because the size skip is at `cmd/grafel/index.go:3987`, upstream of the `index.go` custom gate.

The shape of the change is right — one report at the one mint point, additive, classification byte-identical, `skipped=N` reconciled. Two things block it: the report **goes permanently silent** in the process it was written for, and Part 2 replaces two wrong sentences with two new wrong sentences of the same class.

---

## (a) The dedup lifetime across two indexes in one process — **CONFIRMED, and worse than the PR body says**

Measured, not reasoned. Probe (in-package, one writer, no reset between passes — production has no reset; `setOversizedReportOutput` is test-only):

```
PROBE A run1 (8 distinct oversized files) lines=9
        run2 (SAME repo re-indexed, unchanged) lines=0   output=""
        run3 (8 BRAND-NEW paths, a DIFFERENT repo) lines=0   output=""
```

The PR body's statement — *"the dedup map and cap are process-lifetime, so in a long-lived daemon only the first 8 distinct oversized files are ever named"* — is accurate but stops one step short of the consequence, and the consequence is the harm:

- `if oversizedSeen[filePath] || len(oversizedSeen) >= maxOversizedReports` short-circuits on **either** condition. Once 8 distinct paths have been seen **anywhere in the process**, the cap is saturated **permanently and globally** — run3 above is a different repo with paths never seen before, and it gets nothing.
- The suppression line does not fire either: `last := len(oversizedSeen) == maxOversizedReports` is true exactly once, on the 8th insert, forever. So run 2+ is not "8 lines and a suppression note" — it is **total silence, with no marker that anything was suppressed**.
- Net effect: a user who fixes nothing and re-runs sees a **clean report**. That is the failure mode the brief named, and it is worse than no report at all.

`internal/gitmeta` is not a precedent for this. Its inputs are read once per process. The classifier runs **per index**, and on the daemon also per *incremental* reindex — `internal/extractors/incremental.go:1003` is a `ClassifyWithSize` consumer on the watcher-driven path, so a long-lived daemon re-enters this code every time a file changes.

**Ask:** reset `oversizedSeen` at index start (an exported `ResetOversizedReport()` called where the indexer constructs its classifier — `cmd/grafel/index.go`, `coordinator.go`, `incremental.go`), and pin it with a test that runs two passes under one writer and asserts the second pass is non-empty. This is small; the current behaviour is not shippable as a "user-visible surface".

## (b) Is the size gate the only `"too_large"` mint? — **YES for the graph classifier**

Production-wide `grep 'too_large'` (excluding `_test.go`): the only mint is `internal/classifier/classifier.go:178`. `Classifier.Classify` (the no-size path) has **no** size gate, but it has **no production caller** — the only hits for `.Classify(` are `treesitter.Classify`, `probe.Classify` and the package doc comment. `internal/secrets/secrets.go:410` declares its own `SkipTooLarge = "too_large"`, a different subsystem with a different struct; `oversized_report.go:32`'s "the single place `too_large` is minted" is true within the classifier and false repo-wide — please scope that sentence.

**But the gate is bypassable without TOCTOU, and that undercuts Part 2 (see (d)):** all three index sites initialise `size := int64(-1)` and leave it there when `os.Stat` fails — `cmd/grafel/index.go:3980`, `internal/daemon/extract/coordinator.go:669`, `internal/daemon/extract/subproc.go:273`. `-1 > maxIndexableBytes` is false, so an oversized file whose stat fails **passes the gate**, mints an entity, reaches `readSourceFile`, and *is* truncated — silently, and with no report. Non-concurrent, deterministic, in-tree.

## (c) Does the cap convey its own truncation? — **partly on the first index, not at all afterwards**

Measured, 40 oversized files in one pass: `named=8 suppression=1 totalLines=9`. So a user with 40 vendored bundles sees 8 names and `further oversized-file skips suppressed after 8`. It conveys **that** the list was cut, never **how much** was cut — "8 of 40" and "8 of 9" read identically. That is acceptable-ish on its own (gitmeta precedent), but combined with (a) the real answer is: on the second index the cap conveys *nothing*, because nothing is printed. An "and N more" count, or a single end-of-run line carrying the total, would fix both halves cheaply.

## (d) Part 2's prose — **TOCTOU is not overclaimed; two other sentences are wrong**

Good: *"THE ONE RESIDUAL IS TOCTOU, and it is uncharacterised… Nobody has measured how often that happens"* — correctly flagged as unmeasured, matching #6983's closing comment. No overclaim there. The `safeio.ReadFile`/`ReadFileLimited` facts check out (`internal/safeio/safeio.go:265-268`, `:285`, `:309-311`), the three constants are genuinely unchanged, and the carried-over per-pass claim is honestly labelled as carried over.

Two defects, both the same class the comment exists to correct:

1. **"All three extraction entry points gate on that result"** — there is a **fourth** `ClassifyWithSize` consumer: `internal/extractors/incremental.go:1003`. It does gate identically (`if cr.Skip || cr.Language == ""`), so the *conclusion* survives — but the enumeration is false, and #6983's own measurement comment listed that site explicitly. `oversized_report.go:29-30` repeats the same undercount ("Three call sites classify with a size").
2. **"THE ONE RESIDUAL IS TOCTOU"** — false. The `size = -1`-on-stat-failure bypass in (b) is a second residual, it is not concurrent, and it produces exactly the truncation the paragraph above it says cannot happen. A universal "the one" claim that source contradicts is precisely what produced #6983.

Rewrite as "the residuals are TOCTOU (uncharacterised) and a stat failure that hands the gate `size = -1`", and say "four call sites".

## The `reflect.DeepEqual` tally assertion (#6975) — **passes**

`control` is derived from the thing under test, but it is pinned to the **independent literal** `map[string]int{".pas": 1}` before it is used as an expectation, and `Counts()` returns a fresh copy (`unsupported.go:329-335`), so `control` is a snapshot and not an alias of the live map. Non-vacuous. No change needed.

---

## My mutants — 4 applied, **0 DEAD**

None of the author's 7 re-run. Each applied with an exact pre-edit occurrence assertion and the enclosing function printed; `go vet` exit 0 for every one; baseline re-verified green immediately before each; scored `-count=1`.

| # | mutation (enclosing func) | result |
|---|---|---|
| **N1** | print the suppression line **once per suppressed file** instead of once (`reportOversizedSkip`; anchor count 1→1) | **ALIVE — distinguishable and ungraded, production-reachable.** 40 bundles → 33 suppression lines: output unbounded again, the exact failure the cap exists to prevent. Nothing observes that the line fires once. **Priced: 2 lines** in `TestOversizedReportIsCapped` — `strings.Count(out,"suppressed after")==1` and `strings.Count(out,"grafel:")==maxOversizedReports+1`. |
| **N2** | **delete the mutex entirely** — `oversizedMu.Lock()` 3→0, `oversizedMu.Unlock()` 4→0, 7 markers (`setOversizedReportOutput`, its restore closure, `reportOversizedSkip`) | **ALIVE — including under `-race`** (`go test -count=1 -race ./internal/classifier/` exit 0). No test in the package classifies concurrently, so the mutex is 100% ungraded. **Production-reachable and fatal:** `ClassifyWithSize` runs inside the `i.workers` goroutine pool at `cmd/grafel/index.go:3972-3987`, so unguarded `oversizedSeen[filePath] = true` is a concurrent map write → Go runtime fatal, not a warning. **Priced: ~12 lines** (a `sync.WaitGroup` over 32 goroutines on 4 oversized paths + a locking writer), run under `-race`. I wrote and ran exactly that probe; it passes on HEAD and is what N2 needs to die. |
| **N3** | drop `oversizedSeen = nil` from `setOversizedReportOutput`'s set path (anchor count 1) | **ALIVE** |
| **N4** | make `setOversizedReportOutput`'s restore closure a no-op (anchor count 1) | **ALIVE** |
| **N3+N4** | both resets removed | **DEAD** — `TestOversizedReportIsCapped` fails |

N3/N4 are **mutually-masking guards**: two redundant resets, each fully shadowed by the other, so *neither* is graded on its own. It also means `TestOversizedReportIsCapped` silently depends on a clean map (4 paths accumulate ahead of it) rather than on the cap — an order-coupling worth breaking by having the cap test assert its own starting state.

Note the shape: the seams the author did not mutate — the writer indirection, the mutex, the reset, the suppression line's own bound — are **all** ungraded. The 7-DEAD table pins the report's *content* and nothing about the state machine that decides whether it is emitted at all, which is the same machinery that produces the (a) defect.

## Verification I ran

- `go vet ./internal/classifier/ ./internal/links/ ./cmd/grafel/` — exit 0.
- `go test -count=1 ./internal/classifier/ ./internal/links/` — exit 0. `go test -count=1 -race ./internal/classifier/` — exit 0. `go test -count=1 ./internal/classifier/ ./cmd/grafel/` — exit 0 (`cmd/grafel` 158.6s, whole-graph digest pin).
- `scripts/quality/run.sh --ratchet` (isolated `GRAFEL_HOME` + `GRAFEL_DAEMON_ROOT`) — exit 0: *"quality ratchet OK — 38 gated fixtures held their baseline (29 at 100% must-have recall)"*.

## End-to-end difference, exhibited (brief item 5)

Fixture: a copy of `internal/quality/golden/typescript-react-mini` with one added `src/big_bundle.js` of **2,000,080 bytes** declaring `truncMarkerHandler` past the 1 MiB cut. Run through the production `Index()` entry point via `grafel quality <fixture-dir>`, isolated `HOME`/`GRAFEL_HOME`/`GRAFEL_DAEMON_ROOT`, with binaries built from each commit:

```
bd03cdfeb (parent):  occurrences of "big_bundle" on stdout+stderr = 0
3e266c957 (this PR): grafel: big_bundle.js is 2000080 bytes, over the 1048576-byte limit,
                     and was NOT indexed — it mints no entity and appears nowhere in the graph (#6985)
skipped=1 on BOTH   (classification unchanged, as claimed)
```

So the report does fire on a real index, on the real path, and the `ClassifyResult` really is unchanged. That part is confirmed by observation rather than by unit test — and it is why (a) matters: this is the one line the user gets, and on their second run it is not there.

## Summary

Ship-blockers: **(a)** the process-lifetime dedup+cap going permanently, globally silent — including for repos never seen — with no suppression marker; **(d)** "all three extraction entry points" (there are four) and "THE ONE RESIDUAL IS TOCTOU" (the `size = -1` stat-failure bypass is a second, non-concurrent one). Worth fixing in the same pass: grade N1 and N2 (2 and ~12 lines), and give the cap a count so "8 of 40" is distinguishable from "8 of 9".

cajasmota:
## Coordinator mutant CG-1 on `d86b5fba6` — DEAD, and the *way* it died is the interesting part

Own detached worktree, `git status --porcelain` empty at start and after restore, restored by `cp` from a SHA-256-verified backup (`819c676d…` both ways), zero `MUTANT` text remaining. Green baseline immediately before on the same two packages. macOS/APFS, gate-OFF.

**CG-1 — make the PRODUCTION default write nowhere.** Seventeen mutants have been scored on this PR and every one attacks the reporter's *logic*: whether it fires, what it names, the dedup key, the cap, the mutex, the lifetime, the summary wiring. **Nobody attacked the destination.** `oversized_report.go:80-85` resolves it:

```go
 func (r *oversizedReporter) writer() io.Writer {
 	if r.out == nil {
-		return os.Stderr
+		_ = os.Stderr // MUTANT CG-1 keeps the import live
+		return io.Discard // MUTANT CG-1: production default writes nowhere
 	}
 	return r.out
 }
```

`out` is nil in production and set only by `setOversizedReportOutput`, a **test-only** helper. So the question is whether anything observes the branch production actually takes — a report that is perfectly correct and goes to `io.Discard` is indistinguishable from no report at all, and every test that redirects the writer would still pass.

Edit proven landed: exactly 1 occurrence, hit set asserted `[82]`, **enclosing function asserted** to be `writer()`. First attempt failed `go vet` with `"os" imported and not used` — a non-compiling mutant proves nothing, so it needed `_ = os.Stderr` before it was worth scoring.

**Result: DEAD.**

- `./internal/classifier/` — **green.**
- `./cmd/grafel/` — **FAIL**, on exactly one test: `TestOversizedFilesAreNamedAndTotalledByTheIndexer_6985`.

## Why that split matters

Every test in the classifier package redirects the writer, so **none of them can see this**. The production default is graded by **one test, in a different package** — the one you wrote in this round *because* you found the summary wiring was ALIVE against the whole `./cmd/grafel/` suite and chose to write a test rather than price the gap.

So that decision was load-bearing twice over. Without it, `writer()`'s production branch would be entirely ungraded, and this PR would ship a reporter whose output could be redirected into a hole by a one-line edit with a green suite. **That is the single most valuable thing in this round**, and it came from refusing to leave an ALIVE mutant as a priced note.

## On the rest

The lifetime fix is the right shape: hanging the state off the `Classifier` makes it per-index **by construction**, so there is no `Reset` for a caller to forget — a stronger fix than resetting correctly, and it dissolved N3/N4's mutually-masking pair rather than grading them, which is the better outcome.

**Your N2 caveat is exactly the disclosure I want and I am recording it as a good call, not a weakness.** Deleting the mutex is DEAD under `-race` with 4 DATA RACE reports and **survives 5/5 without it**, because the map-write fatal is scheduling-dependent. Stating that the kill's premise is `-race`, then noticing that `test.yml` only runs `-race` on tag push / dispatch / `ci:full`, then adding the label and dispatching the run — that is the difference between a kill and a kill you can rely on. A mutant whose death depends on scheduling luck is not graded, and you said so instead of banking it.

Merging once CI and the dispatched `-race` run are green.

Issue #6986, opened 2026-09-07
MEASURED + DECIDED: python arm shipped (#6994); ruby/java WON'T-PORT — the bare Class: stub already binds 30/32 there and the structural address would dangle all of them; golang/javascript blocked on a corpus that exercises GORM and decorator ORMs


cajasmota:
Surfaced by the independent review of **#6984** (arm A of #6912), which drew a corollary the implementation did not. **Verified in source in the main checkout at `bd03cdfeb` before filing**, including the parts that make it narrower than it first reads.

## The mechanism

#6984 set out to adopt the custom lane's shipped shape for a field→declared-type edge — `REFERENCES` + `ref_kind: "field_target_type"` + `ToID: "Class:<Target>"`. It kept the kind and the `ref_kind` and **deviated on the address**, because measurement showed `Class:<Target>` **goes AMBIGUOUS as soon as a second file declares a same-named type**, producing a dangling edge.

The reviewer reproduced that independently through `resolve.BuildIndex` → `ReferencesEmbedded`: with two files declaring `Customer`, both bare `Customer` and `Class:Customer` stay unbound, while the structural address binds in both cases.

**The corollary: the custom lane ships that address in five languages today.** All five confirmed to mint the same `Class:<Name>` form:

| language | helper |
|---|---|
| python | `helpers.go:62` `pyClassRef` |
| golang | `helpers.go` `goClassRef` |
| javascript | `helpers.go:68` — literal `"Class:" + targetClass` |
| ruby | `helpers.go` `rbClassRef` |
| java | `orm_helpers.go:12` `javaClassRef` |

shared by the Django / SQLAlchemy / DRF / Pydantic, GORM, TypeORM / MikroORM / Drizzle / Mongoose / Prisma / Sequelize / Objection, ActiveRecord and JPA-Hibernate passes.

## What the Python guard actually does — read this before assuming Python is fine

`internal/resolve/refs.go:4574` guards on `(lang == "" || lang == "python") && strings.HasPrefix(originalStub, "Class:") && isSimplePythonIdentifier(name) && idx.nameExists(name)` and returns `DispositionDynamic`.

**That is a DISPOSITION guard, not a resolution.** It does not bind the edge. The Python edge dangles exactly like the others; it is merely **reclassified from `bug-resolver` to `dynamic`**, so it stops appearing as a defect in the disposition table. The comment says as much — it exists for Django's runtime metaclass dispatch (#529).

[...]

PR #6989, opened 2026-09-07
fix(#6988): GenericForeignKey's ct_field was read as a target model — 5 of the 21 wrong edges were BOUND, not dangling

### LINKED ISSUE #6988: django: GenericForeignKey's ct_field STRING is treated as a target model — 21 wrong edges on django, and 5 of them BIND to the sibling field's Constraint node
Found while measuring #6986. **I verified both admits and the emit site in the main checkout at `bd03cdfeb` before filing**, after first refusing the finding as ungrounded and sending it back — the mechanism in the first report did not match the regex, and the corrected one does.

## The bug

`GenericForeignKey("content_type", "object_id")` takes the names of **two sibling fields on the same model**. Neither is a model name. The Django pass treats argument 1 as a target model and emits a `REFERENCES` edge to it.

**Two compounding admits, and fixing either one closes it:**

1. `isDjangoRelationalField` (`internal/custom/python/django.go:903`) gates on
   `strings.HasSuffix(rhs, "ForeignKey")` — and **`GenericForeignKey` ends with `ForeignKey`**.
2. `djangoModelRelTargetRe` (`django.go:152`) is
   `(?:ForeignKey|OneToOneField|ManyToManyField)\s*\(\s*(?:to\s*=\s*)?(?:["']([^"']+)["']|([A-Z][A-Za-z0-9_]*))`
   — **no left word boundary**, so it matches the `ForeignKey(` *inside* `GenericForeignKey(`.

Emitted at `django.go:705` via `referencesClassEdge`.

Note for anyone reading the regex first: `GenericForeignKey` is not in the alternation, and I initially concluded from that the mechanism could not be what was reported. It is the **missing boundary**, not an alternation entry.

## Named instance

`tests/delete/models.py:279`, `class GenericB1`:

```python
generic_delete_top = GenericForeignKey("content_type", "object_id")
```

→ `GenericB1.generic_delete_top` —`REFERENCES`→ `Class:content_type`, `framework=django`, `provenance=INFERRED_FROM_MODEL_FIELD_TARGET`.

`object_id` is **not** affected — only the first argument is captured. `GenericForeignKey()` with no arguments emits nothing, and `GenericRelation("X")` fails the `HasSuffix` gate. Both are consistent with the measured output rather than merely asserted.

## Why this is worse than a dangling edge — the part that matters

On `django`, gate-**ON**: **21 `field_target_type` edges are GFK-sourced. 16 dangle. 5 BIND.**

The 5 bind to the **sibling `ForeignKey` field's `Constraint` node** — e.g. `Book.created_by` → `created_by_ct` — not to a model, not to anything the source meant. That is **#6369's wrong-node hazard live**: a confidently-wrong edge that reads as valid, and it is strictly worse than the dangling case because nothing flags it. Only 16 of the 21 wrong edges are detectable as unresolved; the other 5 are invisible.

**File this at 21-wrong-edges, not 11-danglers.** The first report said 11, which was only the subset targeting the literal string `content_type`; classifying by whether the *source field* is a GFK declaration gives 21.

## The cheap explanation, ruled out

There is **no `class content_type`** anywhere in the corpus. The 12 same-named graph entities are `Constraint` nodes minted at the sibling `content_type = models.ForeignKey(ContentType, …)` lines. All 11 literal-`content_type` edges are the same shape — 9 read at their own file and line, and the 2 with an unresolved FROM endpoint match the identical `content_object = GenericForeignKey("content_type", "object_id")` form. So this is not #6986's duplicate-name or untwinned-record population wearing a different hat.

## Scope

**Gate-OFF — the default users run (#6966) — these edges do not exist at all.** This is a custom-lane defect and its user-facing incidence today is zero. That caps the urgency and should be stated on every figure rather than left implicit.

## What a fix must do

- **Close it at the gate, the regex, or both** — say which and why. A left word boundary on the regex is the smaller change; excluding `GenericForeignKey` in `isDjangoRelationalField` is the more legible one. Doing both is defensible if each is asserted separately: **two guards that only fire together grade neither.**
- **Assert the negative by name.** A named GFK field must produce **no** edge — `generic_delete_top` above is the ready-made case. Recall cannot detect over-firing; only a forbidden row catches a too-broad predicate, and over-firing is the entire defect here.
- **Assert the 5 wrong-but-bound edges specifically**, not a count. A count table cannot see a substitution (#6973: +8 net `TESTS` while 7,221 edges were substituted) — and here the wrong edges are *bound*, so a total will not move at all when they are removed. **The `Book.created_by` → `created_by_ct` shape is the one to pin.**
- **Do not break the legitimate cases**: `ForeignKey`, `OneToOneField`, `ManyToManyField` in both string (`'app.Model'`, `'self'`) and symbol form, and `to=` kwarg form.
- Check whether `GenericRelation` and any other `*ForeignKey`-suffixed constructor deserve the same treatment, rather than fixing only the one name the corpus happened to surface.

Refs #6986, #6912, #6369, #6966.

cajasmota:
Closes #6988. Refs #6986, #6369, #6966, #6973.

## The defect

`GenericForeignKey("content_type", "object_id")` names two **sibling fields on the same model**. Neither argument is a model. `isDjangoRelationalField` (`internal/custom/python/django.go`) gated on `strings.HasSuffix(rhs, "ForeignKey")`, and `GenericForeignKey` ends in `ForeignKey`, so argument 1 was captured as a target model and a `field_target_type` `REFERENCES` edge was emitted to `Class:<ct_field>`.

**Gate state on every figure below (#6966): custom-Python-lane gate ON.** Gate-OFF — the default users run — these edges were never produced at all, so the user-facing incidence of this defect today is **zero**. The measured population is 21 GFK-sourced `field_target_type` edges on the `django` corpus: 16 dangle, **5 BIND** to the sibling `ForeignKey` field's `Constraint` node (`Book.created_by` → `created_by_ct`). Those 5 are #6369's wrong-node hazard live — confidently wrong and invisible, because nothing flags a bound edge.

## Which admit I closed, and why

The issue named two compounding admits, either of which alone closes the bug. **I closed only the gate.**

**Shipped — the constructor gate.** `isDjangoRelationalField` now strips the dotted prefix and rejects `GenericForeignKey` by final segment, so `GenericForeignKey`, `fields.GenericForeignKey` and `contenttypes.fields.GenericForeignKey` are all covered. It is a **deny-list, not an exact-match allow-list**, on purpose: third-party `*ForeignKey` **subclasses** (django-mptt's `TreeForeignKey`, and the `ForeignKey` subclasses apps define) *do* take a model as argument 1, and an allow-list would silently drop them. Pinned by `TestIssue6988_GateKeepsRelationalConstructors`.

**Rejected on measured merit — the left word boundary on `djangoModelRelTargetRe`.** I implemented it first. It closes #6988, and it also **breaks `TreeForeignKey(Category, ...)`**: the missing boundary is exactly what lets a `*ForeignKey` subclass reach its target argument, and `\b` refuses those with the same stroke. This was not an argument — the `\b` version turned `TestIssue6988_LegitimateRelationalFieldsSurvive` red on `Book.category → Class:Category`, an edge that **does** exist on `bd03cdfeb`. Independently reproduced end-to-end by the reviewer on the parent. The trade-off is now pinned in the suite: `TestIssue6988_RegexBoundaryRejected_WouldDropSubclasses` compiles the rejected variant and asserts **both** halves — that it would have closed #6988, and that it would have dropped the subclass. If that cost claim is ever wrong, that test fails and the decision gets revisited.

**This is a trade, not a clean win, and the scope of the claim is the measured corpus.** One guard is sufficient for **every form observed on the `django` corpus** — not in general. The two residues below would both have been closed by the second guard; it costs all `*ForeignKey` subclass recall to do it, which is why the trade went this way. Shipping one guard also keeps it **graded**: two guards that only fire together grade neither, and there is no masking pair here.

## Known residues — disclosed, zero-incidence today, filed as follow-up

Both are #6988's own shape and neither is fixed here. On the `django` corpus, gate ON, `field_target_type` edges whose target is snake_case (so cannot be a model class) went **18 on the parent → 0 on HEAD**, with 956 `field_target_type` edges total on both sides. Both residues are therefore **real hazards that the corpus does not currently exhibit**.

1. **A GFK subclass.** `class CTForeignKey(GenericForeignKey)` used as `content_object = CTForeignKey("content_type", "object_id")` still yields `Tag.content_object → Class:content_type`. That is the deny-list's open-endedness by construction — a new non-target constructor goes in the list.
2. **The 400-char forward scan.** `fullRHS := body[fIdx[0]:min(fIdx[0]+400, len(body))]` hands `djangoRelTarget` the *following lines* as well, so a legitimate `models.ForeignKey(settings.AUTH_USER_MODEL, ...)` whose own argument the regex cannot parse reads a *later* GFK's `ct_field` string instead: `Note.user → Class:content_type` survives on HEAD (and exists on the parent).

Residue 2 is why an earlier draft of the code comment — "`isDjangoRelationalField` is the sole guard on which constructors get here" — was **wrong and is now corrected in the source**. The gate decides which constructors are *processed*; it does not bound what the regex *reads*. That sentence was prose asserting what no test observes, in the PR that fixes an over-firing predicate, and it does not ship.

## Assertions

- **The negative, by name.** `TestIssue6988_NamedGFKFieldEmitsNoTargetEdge` uses the issue's ready-made case verbatim — `generic_delete_top = GenericForeignKey("content_type", "object_id")` from `tests/delete/models.py:279` — and asserts **zero `field_target_type` edges from `GenericB1.generic_delete_top`**, not merely "not to `Class:content_type`", so a re-broadened predicate cannot pass by picking a different wrong target. Recall cannot detect over-firing, and over-firing is the whole defect.
- **Two positive controls on the same fixture**, so the negative is not a vacuous "this file produced nothing" pass: the sibling real `ForeignKey` (`GenericB1.content_type → Class:ContentType`) still resolves, and the GFK field keeps its `CONTAINS` membership — the fix removes the wrong `REFERENCES` edge, **not the field node**.
- **The 5 wrong-but-BOUND edges as a SHAPE, not a count.** `TestIssue6988_WrongButBoundEdgeShapeIsGone` pins `Book.created_by → Class:created_by_ct` gone while `Book.created_by_ct → Class:ContentType` stays. Those edges are *bound*, so a cardinality assertion is blind to their removal by construction (#6973: +8 net `TESTS` while 7,221 edges were substituted). **Layer, stated in the docstring:** this is a pre-resolution emission test, so it pins the emitted shape. The "5 of 21 BIND to a `Constraint` node" figure is corpus-measured, **not** observed by this test, which never runs the resolver.
- **The legitimate set, one assertion per member** (`TestIssue6988_LegitimateRelationalFieldsSurvive`, end to end): `ForeignKey` string `'app.Model'`, `ForeignKey` string `'self'` → owner, `ForeignKey` symbol, `OneToOneField` symbol, `OneToOneField` `to=` kwarg, `ManyToManyField` symbol, `ManyToManyField` `to=` kwarg, plus the `TreeForeignKey` subclass.
- **The symbol group's uppercase-initial anchor** (`TestIssue6988_RegexRejectsLowercaseInitialSymbol`, added on review): `models.ForeignKey(settings.AUTH_USER_MODEL, ...)` and `models.ManyToManyField(auth.User, ...)` must yield **no** target. A lowercase-initial identifier is a module path, never a model name.
- **The neighbour.** `TestIssue6988_GenericRelationEmitsNoTargetEdge` pins today's `GenericRelation` behaviour: it wears neither relational suffix, so it has never been admitted here. Its argument *is* a model, so admitting it is a **recall** change, not this fix — pinning it now makes that a visible change rather than a silent one.

## Verification — macOS 15.6 / APFS / arm64

Measured in `…/scratchpad/impl-6988-z7v2n/wt`, a dedicated worktree off `bd03cdfeb`.

- **Red first.** Before the fix, `TestIssue6988_NamedGFKFieldEmitsNoTargetEdge` failed with `first ToID="Class:content_type"` and `TestIssue6988_WrongButBoundEdgeShapeIsGone` with `first ToID="Class:created_by_ct"` — the two shapes the issue predicted, reproduced exactly.
- `go test -count=1 ./internal/custom/python/` — **PASS** (exit 0), the only package touched.
- `go test -count=1 ./cmd/grafel/` — **PASS** (exit 0). Whole-graph digest pin; no diff-derived package set reaches it.
- `scripts/quality/run.sh --ratchet` — **exit 0**, 38 gated fixtures held their baseline (29 at 100% must-have recall). **No fixture moved, and none could**: `grep -rn "GenericForeignKey\|GenericRelation" internal/quality/golden/` returns nothing, so no golden fixture exercises this path. No baseline constant was touched. Gating therefore rests on the named forbidden row plus the corpus delta (18 → 0), independently re-derived by the reviewer.
- `gofmt -l internal/custom/python/` clean; `go vet` clean. All three suites re-run after the review changes.

Green baseline immediately before each; `go vet` before each score; `-count=1` throughout; each edit proven with an exact occurrence count (`assert n==1`) and the enclosing function or the rewritten line printed after the edit; restored each time by `cp` from a shasum-verified backup, never `git checkout`.

| # | Mutation | Verdict | Killed by |
|---|---|---|---|
| A | Delete the deny-list check in `isDjangoRelationalField` — restore the bare suffix admit | **DEAD** | `GateRejectsGenericForeignKey`, `NamedGFKFieldEmitsNoTargetEdge`, `WrongButBoundEdgeShapeIsGone` |
| B | Key the deny-list on the full dotted `rhs` instead of the final segment — dotted `fields.GenericForeignKey` slips through | **DEAD** | `GateRejectsGenericForeignKey` |
| C | Apply the rejected fix: add `\b` to `djangoModelRelTargetRe` | **DEAD** | `RegexBoundaryRejected_WouldDropSubclasses`, `RegexKeepsEveryTargetForm`, `LegitimateRelationalFieldsSurvive` |
| D | Widen the symbol capture `([A-Z][A-Za-z0-9_]*)` → `([A-Za-z][A-Za-z0-9_]*)` (1 of the 9 sites of that pattern in the file; the other 8 verified untouched) | **DEAD** *(was ALIVE and ungraded — graded on review)* | `RegexRejectsLowercaseInitialSymbol` |
| E | Forward-scan window `+400` → `+4000` in `(*DjangoExtractor).Extract` | **ALIVE — equivalent on the `django` corpus** (956 `field_target_type` edges either way), not equivalent in principle. Ungraded on purpose: it is residue 2 above, filed as follow-up | — |
| F | `strings.LastIndexByte(ctor, '.')` → `strings.IndexByte` — `contenttypes.fields.GenericForeignKey` misses the deny-list | **DEAD** | `GateRejectsGenericForeignKey` |

A–C are mine; D–F are the reviewer's. C is the one that matters for the design decision: the choice *not* to add the boundary is itself graded, and the recall it would have cost is asserted rather than claimed.

### DIFF
--- a/internal/custom/python/django.go
+++ b/internal/custom/python/django.go
@@ -156,7 +156,13 @@ var (
 	// but it drops `TreeForeignKey(Category, ...)` with it — measured, not
 	// assumed: see TestIssue6988_RegexBoundaryRejected_WouldDropSubclasses.
 	// This regex extracts a target; it is NOT the constructor filter.
-	// `isDjangoRelationalField` is the sole guard on WHICH constructors get here.
+	// `isDjangoRelationalField` decides which constructors are PROCESSED — but it
+	// does not bound what this regex READS. The caller hands it a 400-char
+	// forward window (`fullRHS` below) that can span later lines, so a
+	// legitimate `ForeignKey` whose own argument this regex cannot parse
+	// (`models.ForeignKey(settings.AUTH_USER_MODEL, ...)`) can still match a
+	// string belonging to a constructor the gate REJECTED. That residue survives
+	// #6988 and is zero-incidence on the measured corpus; it is filed separately.
 	djangoModelRelTargetRe = regexp.MustCompile(
 		`(?:ForeignKey|OneToOneField|ManyToManyField)\s*\(\s*(?:to\s*=\s*)?(?:["']([^"']+)["']|([A-Z][A-Za-z0-9_]*))`)
 
--- a/internal/custom/python/django.go
+++ b/internal/custom/python/django.go
@@ -148,6 +148,15 @@ var (
 	// target model — string form ('app.Model' / 'self') or symbol form (Model).
 	// Group 1 = relation kind, group 2 = quoted target (if string form),
 	// group 3 = identifier target (if symbol form).
+	// NOTE (#6988): this alternation has NO left word boundary, and that is
+	// deliberate — it is what lets a third-party `*ForeignKey` SUBCLASS
+	// (django-mptt's `TreeForeignKey`, and the `ForeignKey` subclasses apps
+	// define) reach its target argument. A leading `\b` would also have closed
+	// #6988, by refusing to match the `ForeignKey(` inside `GenericForeignKey(`,
+	// but it drops `TreeForeignKey(Category, ...)` with it — measured, not
+	// assumed: see TestIssue6988_RegexBoundaryRejected_WouldDropSubclasses.
+	// This regex extracts a target; it is NOT the constructor filter.
+	// `isDjangoRelationalField` is the sole guard on WHICH constructors get here.
 	djangoModelRelTargetRe = regexp.MustCompile(
 		`(?:ForeignKey|OneToOneField|ManyToManyField)\s*\(\s*(?:to\s*=\s*)?(?:["']([^"']+)["']|([A-Z][A-Za-z0-9_]*))`)
 
@@ -898,12 +907,43 @@ func (e *DjangoExtractor) Extract(ctx context.Context, file extractor.FileInput)
 	return out, nil
 }
 
+// djangoNonTargetRelCtors names field constructors that carry a relational
+// SUFFIX but whose first argument is NOT a model (issue #6988).
+// `GenericForeignKey(ct_field, fk_field)` takes the names of two sibling fields
+// on the same model, so treating argument 1 as a target model produced a wrong
+// edge — dangling when the sibling name matched no entity, and worse,
+// confidently BOUND to the sibling ForeignKey's Constraint node when it did
+// (#6369's wrong-node hazard).
+//
+// Keyed on the FINAL dotted segment, so `GenericForeignKey`,
+// `fields.GenericForeignKey` and `contenttypes.fields.GenericForeignKey` are all
+// covered. It is a deny-list rather than an exact-match allow-list on purpose:
+// third-party `*ForeignKey` SUBCLASSES (django-mptt's `TreeForeignKey`, and the
+// `ForeignKey` subclasses apps define) do take a model as argument 1, and an
+// allow-list would silently drop them. `GenericForeignKey` is the one such
+// constructor this list currently carries — the corpus surfaced no other, and a
+// new name goes here rather than into a re-broadened suffix rule.
+// `GenericRelation` is the adjacent case and is NOT listed: its argument IS a
+// model, but it wears neither suffix, so it has never been admitted here at all
+// (pinned by TestIssue6988_GenericRelationEmitsNoTargetEdge — admitting it is a
+// recall change, not this fix).
+var djangoNonTargetRelCtors = map[string]bool{
+	"GenericForeignKey": true,
+}
+
 // isDjangoRelationalField reports whether a field constructor RHS (the matched
 // `models.XField` head) is a relational field carrying a target-model argument.
 func isDjangoRelationalField(rhs string) bool {
-	return strings.HasSuffix(rhs, "ForeignKey") ||
-		strings.HasSuffix(rhs, "OneToOneField") ||
-		strings.HasSuffix(rhs, "ManyToManyField")
+	ctor := rhs
+	if dot := strings.LastIndexByte(ctor, '.'); dot >= 0 {
+		ctor = ctor[dot+1:]
+	}
+	if djangoNonTargetRelCtors[ctor] {
+		return false
+	}
+	return strings.HasSuffix(ctor, "ForeignKey") ||
+		strings.HasSuffix(ctor, "OneToOneField") ||
+		strings.HasSuffix(ctor, "ManyToManyField")
 }
 
 // djangoRelTarget extracts the bare target-model class name from a relational
--- a/internal/custom/python/django_gfk_6988_test.go
+++ b/internal/custom/python/django_gfk_6988_test.go
@@ -134,6 +134,26 @@ func TestIssue6988_RegexKeepsEveryTargetForm(t *testing.T) {
 	}
 }
 
+// TestIssue6988_RegexRejectsLowercaseInitialSymbol grades the symbol group's
+// UPPERCASE-INITIAL anchor — `([A-Z][A-Za-z0-9_]*)`. Nothing else observes it,
+// and widening it to `[A-Za-z]` is production-reachable: on the django corpus it
+// mints `LogEntry.user -> Class:settings` (from
+// `models.ForeignKey(settings.AUTH_USER_MODEL, ...)`, django/contrib/admin/models.py)
+// and `User.friends -> Class:auth`. A lowercase-initial identifier is a module or
+// package path, never a Django model name, so it must yield no target.
+func TestIssue6988_RegexRejectsLowercaseInitialSymbol(t *testing.T) {
+	cases := []string{
+		`    user = models.ForeignKey(settings.AUTH_USER_MODEL, on_delete=models.CASCADE)`,
+		`    friends = models.ManyToManyField(auth.User, blank=True)`,
+	}
+	for _, rhs := range cases {
+		if got := djangoRelTarget(rhs, "LogEntry"); got != "" {
+			t.Errorf("djangoRelTarget(%q) = %q, want \"\" — a lowercase-initial identifier is a "+
+				"module path, not a model; the symbol group must stay anchored to [A-Z] (#6988)", rhs, got)
+		}
+	}
+}
+
 // ---------------------------------------------------------------------------
 // End-to-end, through the real Django extractor.
 // ---------------------------------------------------------------------------
@@ -235,6 +255,11 @@ func TestIssue6988_NamedGFKFieldEmitsNoTargetEdge(t *testing.T) {
 // `created_by_ct` — so they read as valid and a cardinality assertion is blind
 // to their removal by construction (#6973). This asserts the exact substitution
 // shape instead.
+//
+// Layer note: this is a PRE-RESOLUTION emission test, so what it pins is the
+// emitted edge shape. The "5 of 21 BIND to a sibling Constraint node" figure is
+// corpus-measured (gate ON), NOT observed here — this test does not run the
+// resolver and makes no claim about binding.
 func TestIssue6988_WrongButBoundEdgeShapeIsGone(t *testing.T) {
 	rels := djangoEdges6988(t, gfkSrc6988)
 
--- a/internal/custom/python/django_gfk_6988_test.go
+++ b/internal/custom/python/django_gfk_6988_test.go
@@ -1,0 +1,302 @@
+package python
+
+import (
+	"context"
+	"regexp"
+	"testing"
+
+	"github.com/cajasmota/grafel/internal/extractor"
+	"github.com/cajasmota/grafel/internal/types"
+)
+
+// Issue #6988 — `GenericForeignKey(ct_field, fk_field)` names two SIBLING FIELDS
+// on the same model. Neither argument is a model. The Django model pass used to
+// treat argument 1 as a target model and emit a `field_target_type` REFERENCES
+// edge to `Class:<ct_field>`.
+//
+// Two compounding admits let it through, and EITHER ALONE closes the bug:
+//
+//	(1) `isDjangoRelationalField` gated on `strings.HasSuffix(rhs, "ForeignKey")`
+//	    and `GenericForeignKey` ends in `ForeignKey`.
+//	(2) `djangoModelRelTargetRe` has no LEFT word boundary, so the alternation's
+//	    `ForeignKey(` matches INSIDE `GenericForeignKey(`. (`GenericForeignKey`
+//	    is not in the alternation — the missing boundary is the whole story.)
+//
+// ONLY (1) was fixed. (2) was rejected on measured merit: the missing boundary
+// is also what lets third-party `*ForeignKey` SUBCLASSES (django-mptt's
+// `TreeForeignKey`) reach their target, and a `\b` drops those with it —
+// TestIssue6988_RegexBoundaryRejected_WouldDropSubclasses compiles the rejected
+// variant and shows both halves of that trade. Shipping one guard also keeps it
+// GRADED: two guards that only fire together grade neither.
+//
+// Scope note (#6966): these edges only ever existed with the custom Python lane
+// gate ON. Gate-OFF — the default — they were never produced, so the
+// user-facing incidence of the defect was zero.
+
+// ---------------------------------------------------------------------------
+// The shipped guard: the constructor gate.
+// ---------------------------------------------------------------------------
+
+// TestIssue6988_GateRejectsGenericForeignKey grades the `HasSuffix` gate
+// on its own. It never reaches the regex.
+func TestIssue6988_GateRejectsGenericForeignKey(t *testing.T) {
+	rejected := []string{
+		"GenericForeignKey",
+		"fields.GenericForeignKey",
+		"contenttypes.fields.GenericForeignKey",
+		"django.contrib.contenttypes.fields.GenericForeignKey",
+	}
+	for _, rhs := range rejected {
+		if isDjangoRelationalField(rhs) {
+			t.Errorf("isDjangoRelationalField(%q) = true, want false: "+
+				"GenericForeignKey's first argument is a sibling field name, not a model (#6988)", rhs)
+		}
+	}
+}
+
+// TestIssue6988_GateKeepsRelationalConstructors pins the ACCEPT side of
+// the same gate, member by member — a set needs an assertion per member. The
+// third-party `*ForeignKey` SUBCLASSES are deliberately still admitted: they do
+// take a model as argument 1, so a whole-suffix ban would be a recall loss.
+func TestIssue6988_GateKeepsRelationalConstructors(t *testing.T) {
+	accepted := []string{
+		"models.ForeignKey",
+		"ForeignKey",
+		"models.OneToOneField",
+		"models.ManyToManyField",
+		"mptt.models.TreeForeignKey", // django-mptt subclass: arg 1 IS a model
+	}
+	for _, rhs := range accepted {
+		if !isDjangoRelationalField(rhs) {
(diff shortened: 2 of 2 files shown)


cajasmota:
## Independent adversarial review — **APPROVE** (two ungraded residues to follow up, neither a blocker)

Measured in two dedicated worktrees, macOS 15.6 / APFS / arm64:
`…/scratchpad/review-6989-q9x4/wt` (parent `bd03cdfeb`) and `…/review-6989-q9x4/head` (`58341a61b`).
Backup shasum of `django.go` = `634b69a6…` (same file the author reports). **Gate ON throughout (#6966);
gate-OFF these edges never existed, so user-facing incidence today is zero.**

### (a) The `\b` rejection is CORRECT — my issue body is wrong and I will correct it

Verified on the parent, **end to end through the registered `python_django` extractor**, not by reading the regex:

- `Item.category = TreeForeignKey(Category, on_delete=…)` on `bd03cdfeb` emits
  `Item.category -> Class:Category`, `REFERENCES`, `ref_kind=field_target_type`,
  `provenance=INFERRED_FROM_MODEL_FIELD_TARGET`. **The edge exists on the parent.**
- Patching the package var to `regexp.MustCompile("\\b" + djangoModelRelTargetRe.String())` and re-running the
  same extraction: `TreeForeignKey edge present = false`, `GFK wrong edge present = false`.

So the `\b` variant does close #6988 **and** does drop every `*ForeignKey` subclass with it. My issue body's
literal claim ("either alone closes the bug") is true; its *implication* — that the two admits are
interchangeable fixes — is what is wrong, and the author is right to have shipped the gate instead.
`TestIssue6988_RegexBoundaryRejected_WouldDropSubclasses` genuinely **compiles and runs** the rejected regex
(`FindStringSubmatch` on both inputs), so it is a measurement, not prose. My independent end-to-end run agrees
with it.

### (b) The deny-list does not under-fire, but it still admits two GFK spellings

No under-firing found: the only entry is `GenericForeignKey`, no legitimate constructor carries that final
segment, and `TestIssue6988_GateKeepsRelationalConstructors` pins `TreeForeignKey` explicitly. An allow-list
would be the wrong shape — measured, per (a). **Aliased imports are safe**: `import GenericForeignKey as GFK`
never reaches the gate at all, because `djangoModelFieldRe` requires the callee head to end in
`Field|ForeignKey|…`. Two spellings do survive, both reproduced on `58341a61b`:

1. **GFK subclass.** `class CTForeignKey(GenericForeignKey)` used as `content_object = CTForeignKey("content_type", "object_id")`
   → `Tag.content_object -> Class:content_type`. #6988's exact shape. This is the deny-list's open-endedness by construction.
2. **The 400-char forward scan.** `fullRHS := body[fIdx[0]:min(fIdx[0]+400, …)]` in `(*DjangoExtractor).Extract`
   (line ~711) hands `djangoRelTarget` the *following lines* too, so a legitimate `ForeignKey` whose own
   argument the regex cannot read reads the GFK's `ct_field` string instead:
   ```
   user = models.ForeignKey(settings.AUTH_USER_MODEL, on_delete=models.CASCADE)
   note_object = GenericForeignKey("content_type", "object_id")
   ```
   → **`Note.user -> Class:content_type` survives on HEAD** (it also exists on the parent). `settings.AUTH_USER_MODEL`
   next to a GFK is ordinary Django. The PR comment's "`isDjangoRelationalField` is the sole guard on WHICH
   constructors get here" is therefore not quite true — the *gate* is, but the *regex still reads text from
   constructors the gate rejected*. Prose asserting what no test observes.

**Incidence check (django corpus, gate ON):** I scanned every `.py` under `archigraph-corpora/django` with the
real extractor and counted `field_target_type` edges whose target is snake_case (cannot be a model class):
**parent = 18** (incl. `GenericB1.generic_delete_top -> Class:content_type` and `TaggedItem.created_by ->
Class:created_by_ct`, the issue's two named shapes), **HEAD = 0**. Total `field_target_type` edges 956 on both.
So both residues are **zero-incidence on the measured corpus** — real hazards, not present today. Follow-up issue,
not a blocker.

### (c) One guard is sufficient *on the measured corpus*, not in general

Both residues in (b) would have been closed by the second guard. The "no masking pair / keep it graded" reasoning
is sound and I'd ship the same way, but the claim as written in the PR body is stronger than the evidence. Please
soften it to "sufficient for every form observed on the corpus" and file the two residues.

### (d) Bound-substitution shape: **yes, asserted as a shape**

`TestIssue6988_WrongButBoundEdgeShapeIsGone` asserts `Book.created_by -> Class:created_by_ct` absent, zero
`field_target_type` from `Book.created_by`, **and** `Book.created_by_ct -> Class:ContentType` still present. That
is a substitution shape, not a cardinality. One honest limit worth stating in the comment: this is a
pre-resolution emission test, so it pins the *emitted* shape; the "5 of 21 BIND to a Constraint node" claim is
corpus-measured, not test-pinned. That is the right layer to pin at — just don't let the docstring imply the test
observes the binding.

### Mutants — mine, permissive direction, none of A/B/C

Green baseline (`exit 0`) immediately before each; `go vet` before each score; `-count=1`; each edit applied with
an asserted `count==1` and the enclosing symbol printed; restored by `cp` from the `634b69a6…` backup, never
`git checkout`.

| # | Mutation | Verdict |
|---|---|---|
| **D** | `djangoModelRelTargetRe` symbol group `([A-Z][A-Za-z0-9_]*)` → `([A-Za-z]…)` (1 of 9 sites in the file; other 8 left intact) | **ALIVE — distinguishable and ungraded.** Production-reachable: on the django corpus it mints `LogEntry.user -> Class:settings` (`django/contrib/admin/models.py`) and `User.friends -> Class:auth`. **Grading cost ≈ 2 lines**: `TestIssue6988_RegexKeepsEveryTargetForm` pins only an accept set; add one reject case `models.ForeignKey(settings.AUTH_USER_MODEL, …)` → want `""`. Cheap — please add it. |
| **E** | forward-scan window `fIdx[0]+400` → `+4000` in `(*DjangoExtractor).Extract` | **ALIVE — equivalent on the django corpus** (956 `field_target_type` edges either way). Not equivalent in principle. The sharper point is that the *existing* 400-char window already leaks across ~6 lines, which is exactly residue (b)(2), and nothing pins it. |
| **F** | `strings.LastIndexByte(ctor,'.')` → `strings.IndexByte` (first dot), so `contenttypes.fields.GenericForeignKey` → `fields.GenericForeignKey` misses the deny-list | **DEAD** — `TestIssue6988_GateRejectsGenericForeignKey`. The member-per-spelling list is load-bearing, not decorative. |

### Suites (all in `…/review-6989-q9x4/head`, tree clean, `git status --porcelain` empty)

- `go test -count=1 ./internal/custom/python/` — **exit 0**
- `go test -count=1 ./cmd/grafel/` — **exit 0**, 154.3 s
- `scripts/quality/run.sh --ratchet` (isolated `GRAFEL_HOME`/`GRAFEL_DAEMON_ROOT`, `GOMAXPROCS=3`) — **exit 0**,
  38 gated fixtures held baseline, 29 at 100 % must-have recall
- Zero-golden-coverage claim **verified**: `grep -rc GenericForeignKey internal/quality/golden/` has no non-zero
  file. The ratchet genuinely could not move. Gating rests on the named forbidden row plus the corpus delta above
  (18 → 0), which I re-derived independently — that is adequate.

### Asks before merge (all small, none blocking my APPROVE)

1. Add the mutant-D reject case (~2 lines) to `TestIssue6988_RegexKeepsEveryTargetForm`.
2. Soften "the sole guard on WHICH constructors get here" / "one guard is sufficient" to the corpus-scoped claim,
   and file a follow-up for the two residues in (b) — the GFK subclass and the 400-char forward scan. Both are
   0-incidence today; both are #6988's shape.
3. I will correct #6988's body: admit (2) closes the bug but costs all `*ForeignKey` subclass recall.

Not merged, nothing pushed. Both review worktrees removed.

cajasmota:
## Coordinator mutant CH-1 on `2710740ea` — DEAD

Own detached worktree, `git status --porcelain` empty at start and after restore, restored by `cp` from a SHA-256-verified backup (`68872d2b…` both ways), zero `MUTANT` text remaining, green baseline immediately before. macOS/APFS, gate-ON population.

**CH-1 — unanchor the START of the forward window.** Six mutants have been scored here and the window has been attacked once, at its **end** (E: 400 → 4000, equivalent on the corpus). Nobody attacked its **start** — which is the half that decides whose text you are reading, and the half residue 2 in #6990 is about.

```go
-  fullRHS := body[fIdx[0]:min(fIdx[0]+400, len(body))]
+  fullRHS := body[0:min(fIdx[0]+400, len(body))] // MUTANT CH-1: window start unanchored
```

Now every field's target extraction searches from the top of the class body, so the first relational constructor anywhere above it can supply the target. Edit proven landed: exactly 1 occurrence, hit set asserted `[717]`, enclosing function printed and checked, 0 occurrences of the original afterwards, `go vet` clean before scoring.

**Result: DEAD** — `./internal/custom/python/`, on `TestIssue6988_LegitimateRelationalFieldsSurvive`.

## The part worth noting

It died on the **positive** set, not the forbidden row.

The eight legitimate-forms assertions were added to prove the fix does not *under*-fire — that `ForeignKey`/`OneToOneField`/`ManyToManyField` in string, symbol and `to=` forms all still resolve. They turn out to also pin **which declaration each target came from**, because a mis-anchored window gives a field the wrong neighbour's target and the expected pairs stop matching. So the set that exists to protect recall is simultaneously the thing grading the window's anchoring — a property nobody wrote it for, and a second instance this cycle of an enumerated positive set outperforming the pin it was written as.

That is worth stating because it changes how #6990 should be read: **the window's start is graded; its width is not.** E showed 400 → 4000 is equivalent on the corpus, and residue 2 needs only a GFK within the window of a legitimate relational field. So the open residue is genuinely about *width and content*, not about anchoring, and a fix that re-anchors the window to the declaration it started in would close it at the cause.

## On the round

Mutant D is graded and DEAD, with the mutation applied to exactly **1 of the 9** `([A-Z][A-Za-z0-9_]*)` sites in the file and the other 8 asserted intact — that is the right way to mutate a pattern that appears many times, and it is the failure mode that has produced false ALIVE verdicts on this board more than once.

The prose fix landed **in the source comment, not only in the PR body**: it now says the gate decides which constructors are *processed* but does not bound what the regex *reads*, and names the residue. That is the correction that mattered most — it was the board's dominant defect class about to ship inside the fix for an over-firing predicate.

Merging on green.

Issue #6599, opened 2026-08-25
terminality is derived per KIND, so property and field leaves land in the orphan defect bucket — 75.2% of SCOPE.Operation orphans and 90.8% of SCOPE.Schema orphans are leaves


cajasmota:
The orphan bucket's terminal-by-construction derivation is **per KIND** (`internal/feedback/report.go`, the #6346 derivation around `:400`). A kind is exempted as terminal only if the whole kind is terminal. `SCOPE.Operation` participates in semantics because methods call things — so every *leaf* sibling inside that kind lands in the defect bucket, and there is no way to exempt them.

This is documented in that comment block as "known limitation (2)". #6583 has now measured it, on a language where the leaves are the **majority of the kind**.

## Measured (#6583)

3 real VB.NET trees, 302 `.vb` files, through extract + `ResolveImports` + `ReferencesEmbedded`, replicating the `internal/feedback` orphan definition exactly.

`SCOPE.Operation`: **4,820 / 9,394 orphan = 51.3%** — reproducing the shape reported in #6535.

Split by subtype, the orphan pool is dominated by one thing:

| subtype | orphans | share of pool |
|---|---|---|
| `property` | 3,626 | **75.2%** |
| `function` | 496 | 10.3% |
| `sub` | 521 | 10.8% |
| `constructor` | 170 | 3.5% |
| `event` | **2** | 0.04% |

And the decisive cut — properties by declaration shape:

| shape | total | orphan | rate |
|---|---|---|---|
| auto-property (no accessor body) | 3,070 | 2,896 | 94.3% |
| full property, body has **no** call site | 749 | 730 | 97.5% |
| full property, body **has** call sites | 485 | **0** | **0.0%** |

**Every property that has something to link is linked.** The orphans are exactly the properties that declare storage and nothing else — the field-leaf shape, one kind over.

Excluding property leaves: **51.3% → 23.5%**, roughly the `function` rate.

## Not VB.NET-specific

The same measurement found `SCOPE.Schema` `field` leaves are **1,862 of 2,050** `SCOPE.Schema` orphans (90.8%). So the artefact accounts for most of two different kinds, in two different languages' worth of shape.

## What this is not

- **Not an extractor gap.** Accessor call sites are harvested correctly — 0 of 485 properties with call sites are orphaned.
- **Not a resolver gap.** Consistent with #6535's own Resolution Disposition vector putting `bug-extractor` at 8.41%, which could never account for a 60.6% rate.
[...]

PR #6991, opened 2026-09-07
feat(#6912): protobuf fields carried their declared type only as a property and in the Signature — emit the field→type REFERENCES edge, same-file targets only

cajasmota:
Arm B of #6912 — the field→declared-type edge for **protobuf**. Arm A (C#) is #6984 (`a3e0e8e41`), the template this follows. One language; other languages remain, so `Refs`, not `Closes`.

## What the gap actually was here, and why it is not the same gap as C#

`buildField` records a proto field's declared type as the `type` PROPERTY and inside `Signature`, and never as a relationship — the #6912 shape, and the population #6726 measures at 99.3% orphan on their corpus (protobuf is one of their four languages).

But protobuf was **not** a language with no field-type edge at all. `buildMessage` already emits one REFERENCES edge per named field type — anchored on the **MESSAGE**, with the field named only in a `via_field` property:

```
message User -[REFERENCES {type: Profile, via_field: profile}]-> message Profile
```

So the message was wired and the `SCOPE.Schema`/`field` record was a leaf on its outbound side. *"Which fields point at `Profile`?"* — answerable for C# after arm A, and for protobuf answerable only by unpacking a property on an edge anchored one level up.

**This adds a second edge from the field rather than moving the existing one.** Moving it would break the message→message topology the .proto dependency view is built on (#6419, #6422). Duplicating the *anchor* is the point; the information is not duplicated, because the two edges are distinguished by `ref_kind` and answer different questions.

## Decisions, all inherited rather than re-litigated

| axis | choice | why |
|---|---|---|
| kind | `REFERENCES` | arm A + the custom lane's `referencesClassEdge`; `TYPED_AS` (#6906) and `HAS_TYPE` (#5828) are declared with zero producers, and a third spelling splits every "which fields point at X?" query (#6902) |
| property | `ref_kind: "field_target_type"` | verbatim from arm A and the five custom-lane languages. Plus `field_name` and `target_type`, as arm A — **both values asserted**, after review found them claimed here and read by nothing (R1/R2) |
| address | `messageTypeRef` → `scope:schema:message:proto:<file>:<Name>` | **not** `Class:<Name>`, whose bare-name tier #6986 measured at 91.6% dangling. This is already this package's type address (message→type, file→message #6422, rpc request/response #6359), it is file-scoped, and post-resolution the ToID is an entity ID so no query sees the dialect |
| anchor | `FromID` empty | the Django precedent (`python/django_relational.go:246-252`). Hibernate sets an explicit FromID because it hangs its edge off its own parallel node — a different shape, and #6912's body was corrected on exactly this |
| enum targets | same dialect | protobuf enums are `SCOPE.Schema`/`enum`, addressed by `messageTypeRef` like a message. Arm A needed a second dialect (`scope:enum:`) for `SCOPE.Enum`; this arm has **one** address form |

## The same-file rule is REUSED, not reimplemented

`dropUnresolvableTypeRefs` (#6357) already runs in `Extract` after the walk and removes every `REFERENCES` edge — **on any entity, it walks them all** — whose target is not a message/enum declared in this file. Emitting the field edge into the same address space means the same-file rule has one implementation and one spelling for both anchors. Its doc comment is updated to say the property is now load-bearing for two callers rather than incidental for one.

A second private copy of the check inside `field_type_refs.go` would fire only where that pass already fires, leaving both ungraded — the trap arm A called out for its primitive blocklist.

### A guard was deleted here on a false premise, and review caught it — the reasoning is worth keeping

The first cut of `protoFieldTypeCandidates` carried a `protoScalars` blocklist. Mutant M1 (delete it) was killed **only by the unit test written for it**: every end-to-end assertion stayed green. That was read as the redundant guard arm A refused — the in-file gate already drops every scalar, because no `.proto` declares `message string` — and the blocklist was deleted.

**The premise was false in fact, and review disproved it by constructing the case.** "No `.proto` declares `message string`" is a *corpus-relative zero*, not a property of the pass:

```proto
syntax = "proto3";
message string { int32 v = 1; }
enum bool { B_ZERO = 0; }
message Holder { string name = 1; bool flag = 2; map<string, Order> by_key = 3; }
```

`message string` **is** in the gate's local set, so without the blocklist `Holder.name -> …:string` survives **and binds** — it never reaches `bug-extractor` and no disposition figure surfaces it. That is exactly the failure this PR's own header calls out for `namedTypeRefs`' qualifier-stripping, reintroduced two paragraphs later.

The file is legal proto3 and protoc agrees with the graph's *silence*, not with its edge. Verified independently in this worktree (`grpcio-tools` in a scratch venv, `--descriptor_set_out`, **RC 0**):

```
Holder.name: type=TYPE_STRING type_name=''
Holder.flag: type=TYPE_BOOL   type_name=''
Holder.real: type=TYPE_MESSAGE type_name='.Order'
```

Empty `type_name` on the two shadowed fields: the grammar matches `string` in field-type position as a scalar before it ever considers a message type, so the shadowing declaration can never be the referent. `map<string, Order>` gained a bogus key edge beside the correct value edge, and because `namedTypeRefs` kept **its** scalar check, the two anchors disagreed in the wrong direction — the field edge over-firing exactly where the message edge correctly stayed silent.

**M1's verdict is therefore ALIVE and production-reachable, not redundant.** The guard is restored and is now graded **end to end** by `TestProtoFieldTypeRefs_ScalarShadowedByASameFileMessage` (with a positive control that the shadowing entities really exist, so the gate's local set really contains them), which answers the objection that killed it in the first place.

#### The general lesson, now in the code and not only here

**A mutant killed only by its own unit test does not establish that a guard is redundant.** It is equally consistent with the end-to-end fixture space simply *lacking the case*. Distinguishing the two requires **constructing the case** — the step that was skipped, and the whole cost of the review round.

Two other mutants in this PR have that same shape and are correctly left as-is: R3 (qualified-name pass-through) and R4 (the empty-name guard) are DEAD by the unit table and masked end to end, because `…:demo.Profile` and an empty-named target are genuinely not in `local`. The unit table is the right grader for those. The difference is that for the scalar rule the masking was *not* real, and only the constructed fixture could show it. `TestProtoFieldTypeCandidates_Rules` and `TestProtoFieldTypeRefs_EveryScalarProducesNoEdge` both carry notes saying which of the two situations each row is in.

## The rules, each stated and each graded

- **Scalars** → nothing. Enumerated, not sampled: `TestProtoFieldTypeRefs_EveryScalarProducesNoEdge` **generates** one field per member of `protoScalarLiteral`, asserts the field entity exists and carries that scalar as its type (positive control on the fixture), asserts no edge, and asserts a non-scalar field in the same message **did** get its edge (positive control on the pass). `TestProtoScalars_MatchesTheSpecList` pins the production `protoScalars` map against that independent literal (#6975), in both directions.
- **`map<K, V>`** → **the wrapper is never a target; each of K and V is a candidate, subject to the same qualified-name rule.** proto3 constrains a key to an integral/string scalar, so in practice only V ever survives; both are walked rather than assuming the spec. All three branches graded end to end by `TestProtoFieldTypeRefs_MapValueIsTheTarget` (`map<string, Order>` → exactly one row → `Order`; `map<int32, string>` → nothing; `map<string, Missing>` → nothing), with a positive control asserting all three fields really carry a `map<…>` type.
- **`repeated` / `optional` / `required`** never become the target. The label is a separate return value from `fieldTypeAndLabel` and is prepended only to `Signature`; this pass reads the type. `TestProtoFieldTypeRefs_LabelIsNeverTheTarget` asserts all three labelled/unlabelled forms reach the *same* target, with a positive control that the fixture's labels are actually present, and the three label strings are in the forbidden-target list.
- **Qualified and imported types** → nothing, **asserted by name**: `User.address` (imported from `common.proto`), `User.qualified` (`demo.Profile`), `User.fq` (`.demo.Profile`), `User.gone` (`Missing`, declared nowhere), `Outer.deep` (`Outer.Inner` — a nested reference is a qualified name).
- **In-file enums** are targets: `User.role → enum Role`.
- **oneof members** are edged like any other field (`User.alt → Profile`), so #6358's recovered population participates.

### The one place this arm is deliberately NARROWER than the existing edge

`namedTypeRefs` strips a qualifier to its trailing segment, so `demo.Profile` in a file that also declares its own `Profile` makes the **message-level** edge bind to the wrong entity — silently, because a bound edge never reaches `bug-extractor`. `protoFieldTypeCandidates` does not strip: any name carrying a `.` is rejected outright.

`TestProtoFieldTypeRefs_QualifiedTypeIsNotStrippedToASameFileName` pins **both halves**: the field edge is absent, and the pre-existing message edge is byte-for-byte what it was. That second assertion is how a future narrowing of `namedTypeRefs` (which belongs with #6357's cross-file package index) is forced to come here and say so.

## Binding — for protobuf kinds and the protobuf address, not in general

Arm A established that the resolver binds a field→type `REFERENCES` edge at all. `TestProtoFieldTypeRefs_ResolverBindsEveryEdge` drives **protobuf** entity kinds (`SCOPE.Schema`/`message`, `SCOPE.Schema`/`enum`) and the **protobuf** dialect through the production resolver (`resolve.BuildIndex` → `resolve.ReferencesEmbedded`, exactly as assembly does) and asserts every ToID comes back rewritten to a real entity ID, named endpoint by named endpoint (9 rows). Zero dangling is the assertion, not a ratio.

It carries a positive control on the address choice itself: the **bare** name `Profile` must be AMBIGUOUS in the two-file fixture, or the file-scoped address is ungraded. `TestProtoFieldTypeRefs_SameNameInAnotherFileChangesNothing` asserts a rival `.proto` declaring `Profile`/`Order`/`User` leaves this file's edge set unchanged.

The golden run is an independent confirmation through the **real indexer** rather than the unit resolver: proto-mini scores 18/18 relationships with the new row **found**, and 0 forbidden hits.

## Numbers, and the gate on each

**Pass-1 only, gate-independent** — every figure below is this extractor's own output, measured by running the registered `protobuf` extractor over each file; no detector output is read, so no detector gate applies. macOS/APFS, worktree `…/scratchpad/impl-6912b-q9x4t/wt`, corpora read from `/Users/jorgecajas/Projects/archigraph-corpora` (read-only, not indexed).

| corpus | `.proto` files | field entities | fields gaining an edge | edges | dangling |
|---|---|---|---|---|---|
| etcd | 5 | 28 | 3 (10.7%) | 3 | 0 |
| apollo-server | 1 | 125 | 48 (38.4%) | 48 | 0 |
| grpc-go-examples | 3 | 4 | 0 | 0 | 0 |
| actix-examples | 1 | 2 | 0 | 0 | 0 |
| gin | 1 | 0 | 0 | 0 | 0 |
| **total** | **11** | **159** | **51 (32.1%)** | **51** | **0** |

Dangling was checked, not assumed: the probe asserted every emitted ToID is a same-file message/enum address.

etcd's 10.7% is the honest floor made visible — its field types are overwhelmingly `etcdserverpb.X` cross-file references, which is #6357's deferred package index, not a widening of this arm.

**Do not read #6726's 99.3% as moving.** That is their corpus, measured on v0.3.0; the mechanism transfers and the number does not. Also note #6599's `terminalLeafSubtypes` already routes `field` out of the orphan defect count on v0.3.1+, so the orphan-rate arithmetic that made this a P1 is separately in question (see the issue thread) — what this PR claims is the *edge*, which `internal/quality/analytics/scan.go` has no leaf exemption for.

**Golden corpus: +1 relationship, enumerated.** `relationship_extracted_total` for proto-mini moved **55 → 56**, and the whole of that delta is:

```
User.profile (SCOPE.Schema/field, user.proto)
  -[REFERENCES {ref_kind: field_target_type, target_type: Profile}]->
    scope:schema:message:proto:user.proto:Profile
```

Enumerated by extracting the two golden `.proto` files directly and listing every `ref_kind=field_target_type` edge — one row. `User.address` (cross-file) and every scalar field produce nothing, which is why the delta is 1 and not 4.

The baseline was **hand-edited for proto-mini only** (`relationship_found`/`_expected` 17→18, `relationship_extracted_total` 55→56; entity counts unchanged — this arm adds an edge, not an entity). A full `--update-baseline` regeneration was run first and **discarded**, and review proved that was right for a stronger reason than originally given. Two back-to-back regenerations — one at this branch's head, one at the untouched parent `a3e0e8e41` — differ **only** in proto-mini's `17/17 → 18/18` and `55 → 56`, so the committed hand-edit is byte-identical to a regeneration, and `proto-mini +1` is genuinely the only delta attributable to this change.

The other movement is **eight fixtures across four languages** — C# ×3, docker-compose, Python ×3, Razor — *not* "several C#", and every one is a `*_extracted_total` **ceiling**: no `_found`/`_expected` recall figure moved anywhere. They appear identically at the parent, so they are environmental. Crucially, `--ratchet` does **not** fail on a ceiling drop, so committing the regeneration would have been a **silent weakening** of eight ceilings rather than a visible failure.

expected.json gains **two** rows, so the fixture grades the arm in both directions:
- expected: `User.profile -REFERENCES-> Profile`, with a note stating it is not a duplicate of the message-anchored row and how the two are told apart.
- forbidden **F5**: `User.address -REFERENCES-> scope:schema:message:proto:user.proto:Address` — the field-anchored twin of F1, and the row that catches over-firing on the imported-type case. Like F1 it does **not** bless the cross-file gap: when a package index makes the target resolve, delete it and add an expected row.

## Verification

macOS 25.6.0 / APFS, in the worktree above.

- `go test -count=1 ./internal/extractors/proto/... ./internal/quality/... ./internal/resolve/...` — **green**
- `go test -count=1 ./cmd/grafel/` (155s) — **green**. Worth recording precisely: it was green *before* expected.json changed and red *after*, failing `TestProtoEnumValueSubtypeIsGraded_6488` on its **absolute** `17/17` pin (updated to `18/18`, comments updated to say why). So the whole-graph digest tests in `cmd/grafel` did **not** see the new edge — the ones under `worktree_seed_parity_test.go` are A/B parity digests, not absolute pins, so an unchanged result there is not evidence of an unchanged graph. The #6488 absolute count is what caught it. (Review confirmed and generalised this: `semanticDigest` is computed twice at runtime and compared to itself via `assertParity`, with no hardcoded digest literal anywhere in the file.)
- `scripts/quality/run.sh --ratchet` — **exit 0**, 38 gated fixtures held their baseline (29 at 100%). It failed first with `proto-mini: relationship_extracted_total GREW 55 -> 56`, which is the movement enumerated above.
- `gofmt -l internal/ cmd/` clean; `go vet` clean.
- `fieldEnt.Relationships` is an **append**, not an assignment (review, non-blocking): `buildField` sets none today so both are correct now, but an assignment silently clobbers whatever a future `buildField` adds.
- Corpus figures re-measured after restoring the guard: **unchanged** (159 fields, 51 edged, 32.1%, 0 dangling) — no corpus file shadows a scalar name, which is precisely why the corpus could not have found the bug.
- Run under `GRAFEL_HOME` / `GRAFEL_DAEMON_ROOT` pointed at scratch; the real `~/.grafel` was not written and no daemon was started, stopped or killed.

## Mutation score

Preferring the **more permissive** direction. Each edit was applied by a script asserting an **exact occurrence count of 1**, the enclosing function was named and the hit printed, `go vet` ran before every score, the baseline was green immediately before each, always `-count=1`, and each restore was `cp` from a shasum-verified backup (`git checkout` was never used).

| # | mutant (enclosing function) | verdict |
|---|---|---|
| M1 | delete the scalar blocklist (`protoFieldTypeCandidates`) | **ALIVE AND PRODUCTION-REACHABLE — input named.** Originally scored "redundant" on a false premise and the guard was deleted; review constructed `message string` / `enum bool` (protoc RC 0) and the edge survives and BINDS. Guard restored, and now **DEAD end to end** via `TestProtoFieldTypeRefs_ScalarShadowedByASameFileMessage` as well as the unit table. See above |
| M2 | make `dropUnresolvableTypeRefs` skip `field` records, i.e. drop the in-file check for this arm only | **DEAD** — 12 tests, including `TestProtoFieldTypeRefs_CrossFileTypeGetsNoEdge` and `_EveryScalarProducesNoEdge`, plus pre-existing `TestProto_NoRelationshipTargetsAPhantomNode` |
| M3 | do not unwrap `map<>`, target the wrapper (`protoFieldTypeCandidates`) | **DEAD** — `_MapValueIsTheTarget`, `_EmittedEdges`, `_ResolverBindsEveryEdge`, `_Rules` |
| M4 | read `Signature` instead of the type, so the label becomes part of it (`buildMessage`/`addMember`) | **DEAD** — 7 tests incl. `_LabelIsNeverTheTarget` |
| M5 | strip the qualifier to its trailing segment instead of rejecting it (`protoFieldTypeCandidates`) | **DEAD** — 5 tests incl. `_QualifiedTypeIsNotStrippedToASameFileName` and `_ForbiddenTargets` |
| M6 | anchor the edge on the message by setting a non-empty `FromID` (`protoFieldTypeRefs`) | **DEAD** — 6 tests, via `ftrEdges`' FromID and carrier-kind assertions |
| R1 | `field_name` property value → a constant (`protoFieldTypeRefs`) | **found DISTINGUISHABLE AND UNGRADED by review → now DEAD** (8 tests). Nothing read the value; the PR claimed it. `ftrEdges` now cross-checks it against the record's dotted Name on **every** edge in the file |
| R2 | `target_type` property value → a constant (`protoFieldTypeRefs`) | **found DISTINGUISHABLE AND UNGRADED by review → now DEAD** (8 tests). Cross-checked against the ToID's trailing segment, plus two full property lists asserted against an independent literal in `TestProtoFieldTypeRefs_EdgePropertiesAreSpelledOut` |
| R3 | weaken the qualified-name guard so a qualified name passes through unstripped | **DEAD by the unit table alone; masked end to end** — `…:demo.Profile` is genuinely not in `local`. Correctly left at the unit; noted rather than acted on, and the test says so |
| R4 | drop the `p == ""` guard (`protoFieldTypeCandidates`) | **DEAD by the unit table alone; masked end to end.** Reachable only through degenerate `map<, >` text. Same disposition as R3 |
| M7 | drop the per-field dedupe (`protoFieldTypeRefs`) | **ALIVE at first → graded, then DEAD.** The only expression naming one target twice is `map<Order, Order>`, which is invalid proto3 and so unreachable through `Extract`; rather than claim it cannot happen, it is graded at the unit in 12 lines by `TestProtoFieldTypeRefs_DedupesPerTarget`, which carries a `map<Key, Order>` positive control so it cannot pass for a pass that emits at most one edge per field |

## What this arm cannot do

In the code, in a header block, so it is read where the decision lives:

- **File scope only.** Unlike arm A there is no namespace over-fire to pin: one `.proto` declares exactly one package, so there is no in-file shadowing to get wrong.
- **No cross-file / imported targets at all**, which for real protobuf is the majority (etcd: 3 of 28). Needs #6357's package index; separate arm.
- **A nested message referenced as `Outer.Inner` gets nothing** (bare `Inner` does). Same qualified-name rule, asserted by name.

Refs #6912, #6726, #6906, #6984, #6986, #6357.

### DIFF
--- a/cmd/grafel/quality_subtype_assertion_6488_test.go
+++ b/cmd/grafel/quality_subtype_assertion_6488_test.go
@@ -100,7 +100,8 @@ func TestProtoEnumValueSubtypeIsGraded_6488(t *testing.T) {
 	// enum-value subtype flip was live and undetected — the fixture's size AT
 	// THAT TIME. #6518 has since added the per-file carrier entity and the
 	// file -> message CONTAINS row that carrier made gradeable, so the figures
-	// the assertion below demands are 19/19 and 17/17. The #6488 property is
+	// the assertion below demands are 19/19 and 18/18 (#6912 arm B added the
+	// eighteenth relationship row). The #6488 property is
 	// unchanged: absolute counts, re-derived here rather than read from
 	// baseline.json.
 	rep := quality.Evaluate(fix, doc)
@@ -115,9 +116,11 @@ func TestProtoEnumValueSubtypeIsGraded_6488(t *testing.T) {
 				r.SubtypeMismatch, r.GotSubtype, "enum_value")
 		}
 	}
-	// 19 since #6518 added the per-file SCOPE.Component carrier, and 17/17
-	// relationships since the file -> message CONTAINS row it made gradeable
-	// (see proto-mini/NOTICE.md, which asked for exactly that row).
+	// 19 since #6518 added the per-file SCOPE.Component carrier, and 18/18
+	// relationships since #6912 arm B added the field->declared-type row
+	// (User.profile -> Profile) on top of the file -> message CONTAINS row
+	// #6518 made gradeable (see proto-mini/NOTICE.md, which asked for that one).
+	// The entity count is unchanged: arm B adds an edge, not an entity.
 	if rep.EntityExpected != 19 || rep.EntityFound != 19 {
 		var missing []string
 		for _, r := range rep.EntityResults {
@@ -128,8 +131,8 @@ func TestProtoEnumValueSubtypeIsGraded_6488(t *testing.T) {
 		t.Errorf("proto-mini entities %d/%d want 19/19; missing: %v",
 			rep.EntityFound, rep.EntityExpected, missing)
 	}
-	if rep.RelExpected != 17 || rep.RelFound != 17 {
-		t.Errorf("proto-mini relationships %d/%d want 17/17", rep.RelFound, rep.RelExpected)
+	if rep.RelExpected != 18 || rep.RelFound != 18 {
+		t.Errorf("proto-mini relationships %d/%d want 18/18", rep.RelFound, rep.RelExpected)
 	}
 	if n := len(rep.ForbiddenHits); n != 0 {
 		t.Errorf("proto-mini forbidden hits = %d, want 0", n)
--- a/internal/extractors/proto/field_type_refs.go
+++ b/internal/extractors/proto/field_type_refs.go
@@ -119,28 +119,57 @@ const protoFieldTargetRefKind = "field_target_type"
 // integral or string scalar, so in practice only V ever yields a candidate — but
 // both are walked rather than assuming the spec, exactly as namedTypeRefs does.
 // `map<string, Order>` therefore yields [Order]: one candidate, not two, and not
-// a `map` target. All three branches are graded by
-// TestProtoFieldTypeRefs_MapValueIsTheTarget.
+// a `map` target — the `string` key is dropped by the scalar rule below, which
+// is load-bearing here and not decorative (see it). All branches are graded by
+// TestProtoFieldTypeRefs_MapValueIsTheTarget and, for the shadowed-scalar key,
+// by TestProtoFieldTypeRefs_ScalarShadowedByASameFileMessage.
 //
-// THERE IS NO SCALAR BLOCKLIST HERE, and its absence is the design, not an
-// omission. `string`, `int32` and the rest are the largest field population in
-// any .proto file and they must produce zero edges — but they already do,
-// because no .proto declares `message string`, so the in-file gate drops every
-// one of them. A protoScalars check in front of that gate would fire only where
-// the gate already fires, leaving BOTH ungraded (arm A refused the identical
-// blocklist for the identical reason). It was written and then removed after
-// mutation scoring made the masking explicit: deleting the check changed no
-// end-to-end behaviour on any fixture, which is what "redundant" means.
-// TestProtoFieldTypeRefs_EveryScalarProducesNoEdge enumerates the whole
-// production scalar set through Extract and asserts the observable consequence
-// rather than the mechanism.
+// Two rejections:
 //
-// (namedTypeRefs, the older message-anchored path, does keep a protoScalars
-// check. It is left alone: re-deciding it is not this arm's business, and
-// TestProtoScalars_MatchesTheSpecList pins that set against an independent
-// literal either way.)
+//   - A SCALAR (the full protoScalars set: double float int32 int64 uint32
+//     uint64 sint32 sint64 fixed32 fixed64 sfixed32 sfixed64 bool string bytes)
+//     yields nothing. This is the largest field population in any .proto file
+//     and the one that must produce zero edges.
 //
-// One rejection:
+//     THIS CHECK IS NOT REDUNDANT WITH THE IN-FILE GATE, and the first cut of
+//     this arm deleted it on the argument that it was. The argument was that no
+//     .proto declares `message string`, so the gate drops every scalar anyway
+//     and a blocklist in front of it would fire only where the gate already
+//     fires. **That is a corpus-relative zero, not a property of the pass**, and
+//     it is FALSE IN FACT — review of #6991 constructed the case:
+//
+//     message string { int32 v = 1; }
+//     enum bool { B_ZERO = 0; }
+//     message Holder { string name = 1; bool flag = 2; }
+//
+//     `message string` is in the file, so it IS in the gate's local set: without
+//     this check `Holder.name` gets an edge to it and the edge BINDS, so it
+//     never reaches `bug-extractor` and nothing surfaces it — the exact failure
+//     the qualified-name rule below exists to prevent. The file compiles under
+//     protoc at RC 0, and protoc's own descriptor binds `Holder.name` to
+//     TYPE_STRING with an EMPTY type_name: the grammar matches `string` in
+//     field-type position as a scalar before it ever considers a message type,
+//     so the shadowing message can never be the referent and the edge asserts
+//     something protoc says does not exist. `map<string, Order>` under that
+//     declaration gains a bogus second key edge on top of the right one.
+//
+//     Graded end to end, not at the unit, by
+//     TestProtoFieldTypeRefs_ScalarShadowedByASameFileMessage — because "an
+//     edge case that compiles under protoc" is what separates a live guard from
+//     a redundant one, and only a fixture containing the case can tell them
+//     apart. TestProtoFieldTypeRefs_EveryScalarProducesNoEdge enumerates the
+//     whole production scalar set separately, against an independent literal
+//     (#6975).
+//
+//     THE GENERAL LESSON, recorded here because it cost a review round: a
+//     mutant killed ONLY by its own unit test does not establish that a guard is
+//     redundant. It is equally consistent with the end-to-end population simply
+//     lacking the case — and telling those two apart requires CONSTRUCTING the
+//     case, which is the step that was skipped. (`namedTypeRefs`, the older
+//     message-anchored path, keeps its own protoScalars check; had this one
+//     stayed deleted the two anchors would have disagreed in the wrong
+//     direction, the field edge over-firing exactly where the message edge
+//     correctly stays silent.)
 //
 //   - A QUALIFIED name — anything containing a `.`, including the
 //     fully-qualified leading-dot form `.foo.bar.Order` — yields nothing, and is
@@ -162,7 +191,7 @@ func protoFieldTypeCandidates(ftype string) []string {
 	}
 	var out []string
 	for _, p := range parts {
-		if p == "" || strings.Contains(p, ".") {
+		if p == "" || protoScalars[p] || strings.Contains(p, ".") {
 			continue
 		}
 		out = append(out, p)
--- a/internal/extractors/proto/field_type_refs.go
+++ b/internal/extractors/proto/field_type_refs.go
@@ -1,0 +1,206 @@
+package proto
+
+import (
+	"strings"
+
+	"github.com/cajasmota/grafel/internal/types"
+)
+
+// field_type_refs.go — the field→declared-type edge for protobuf (issue #6912,
+// arm B; arm A is internal/extractors/csharp/field_type_refs.go, #6984).
+//
+// THE GAP, AND HOW IT DIFFERS FROM C#. buildField records a proto field's
+// declared type as the `type` PROPERTY and in the Signature, and never as a
+// relationship, so every SCOPE.Schema/field entity this package emits is a leaf
+// on its outbound side. That is the #6912 population: #6726 measures
+// SCOPE.Schema at 99.3% orphan over their corpus, and protobuf is one of their
+// four languages.
+//
+// What protobuf already had, and what it did NOT have, is the whole design
+// question here. buildMessage already emits a REFERENCES edge per named field
+// type — but it hangs it off the MESSAGE (`message User` → `message Order`,
+// properties `type` / `via_field`). The message is therefore already wired; the
+// FIELD is not. "Which fields point at Order?" is answerable for C# after arm A
+// and was not answerable for protobuf, because the only edge that knew about
+// `Order` was anchored one level up and named its field in a property instead
+// of in its endpoints.
+//
+// So this pass adds a SECOND edge, from the field, rather than moving the
+// existing one. Moving it would break the message→message topology the .proto
+// dependency view is built on (#6419, #6422); duplicating the ANCHOR is the
+// point, not duplicating the information.
+//
+// VOCABULARY AND ADDRESS ARE ARM A'S, NOT NEW. Kind REFERENCES with the
+// property `ref_kind: "field_target_type"` — the spelling the custom lane's
+// shared `referencesClassEdge` helper has used in five languages, and the one
+// arm A adopted for core. `TYPED_AS` (#6906) and `HAS_TYPE` (#5828) are both
+// declared in internal/types/kinds.go with zero producers; a third spelling
+// would split every "which fields point at X?" query (#6902).
+//
+// The address is messageTypeRef — `scope:schema:message:proto:<file>:<Name>` —
+// which is this package's existing structural type address, already used by the
+// message→type edge, by the file→message CONTAINS edge (#6422) and by rpc
+// request/response types (#6359). It is FILE-SCOPED, so it does not go
+// AMBIGUOUS when a second .proto declares a same-named message — the failure
+// that #6986 measured at 91.6% dangling for the custom lane's `Class:<Name>`
+// form. Post-resolution the ToID is the target entity's ID, so no query sees
+// the dialect. Note the SCOPE.Enum arm of arm A has no analogue here: protobuf
+// enums are SCOPE.Schema/enum entities and take the SAME messageTypeRef
+// address, so there is one address dialect on this arm rather than two.
+//
+// A NON-BINDING EDGE IS WORSE THAN NO EDGE. An unresolved stub is kept verbatim
+// and dangling and classified `bug-extractor`, with no masking branch and no
+// drop pass, so every miss is a full hit and each emitted edge adds one
+// endpoint to the disposition denominator (#6906). The edge is therefore
+// emitted ONLY where the target is declared IN THE SAME FILE.
+//
+// THAT GATE IS NOT NEW EITHER, and reusing it is deliberate. #6357 already
+// installed dropUnresolvableTypeRefs, which runs in Extract after the walk and
+// removes every REFERENCES edge — on ANY entity — whose ToID is not a
+// message/enum declared in this file. Emitting the field edge into that same
+// address space means the same-file rule has ONE implementation and ONE
+// spelling for both anchors. A second private copy of the check inside this
+// file would fire only where that pass already fires, leaving both ungraded
+// (the redundant-guard trap arm A called out for its primitive blocklist).
+// The consequence is asserted end-to-end rather than described: see
+// TestProtoFieldTypeRefs_CrossFileTypeGetsNoEdge.
+//
+// WHAT THIS PASS IS NOT.
+//
+//   - It is FILE scope and nothing else. It does not consult the proto
+//     `package` declaration, and it cannot: one .proto file declares exactly one
+//     package, so there is no in-file namespace shadowing to get wrong. That is
+//     why arm A's namespace over-fire has no counterpart here.
+//   - It is DELIBERATELY STRICTER THAN THE MESSAGE-LEVEL EDGE on qualified
+//     names, and this is the one place the two anchors disagree. namedTypeRefs
+//     strips a qualifier to its trailing segment (`.foo.bar.Order` → `Order`),
+//     so if THIS file also happens to declare an unrelated `Order`, the
+//     message-level edge binds to the wrong entity — silently, because a bound
+//     edge never reaches `bug-extractor`. protoFieldTypeCandidates does not
+//     strip: any name carrying a `.` is skipped outright. A same-file type is
+//     written bare in practice, so the recall this costs is small and the false
+//     edge it prevents is invisible. Pinned in both directions by
+//     TestProtoFieldTypeRefs_QualifiedTypeIsNotStrippedToASameFileName, which
+//     also pins that the message-level edge's behaviour is UNCHANGED by this
+//     arm. Widening the field edge is not the fix; narrowing namedTypeRefs is,
+//     and it belongs with the cross-file package index #6357 defers.
+//   - A NESTED message referenced as `Outer.Inner` therefore gets nothing from
+//     this pass, by the same rule. That is the honest floor, asserted by name
+//     rather than described.
+//   - It cannot reach an IMPORTED type at all, which for protobuf is the
+//     majority case: `import "common.proto"; ... Address address = 3;` gets no
+//     edge, because pass 1 sees one file and the import directives carry file
+//     paths, not package names (#6357). Closing that needs the cross-file proto
+//     package index, which is a separate arm.
+
+// protoFieldTargetRefKind is the value of the `ref_kind` edge property. It
+// matches arm A's csFieldTargetRefKind and the custom lane's referencesClassEdge
+// verbatim — the discriminator is what makes the edge queryable as a
+// field→declared-type edge across languages, so it must be one string.
+//
+// It is NOT the spelling the pre-existing message→type edge uses (`type` /
+// `via_field`). That edge is a different relation with a different anchor and is
+// left exactly as it was; re-spelling it would be a graph-wide change with no
+// bearing on #6912.
+const protoFieldTargetRefKind = "field_target_type"
+
+// protoFieldTypeCandidates returns the bare type names a field's declared type
+// may address, in source order and without deduplication.
+//
+// The input is the rendered type string from fieldTypeAndLabel /
+// mapFieldTypeAndLabel — the TYPE only. The proto label (`repeated`,
+// `optional`, `required`) is returned by those helpers as a SEPARATE value and
+// prepended only to the Signature, so `repeated Order orders = 3;` arrives here
+// as "Order" and never as "repeated Order". Pinned by
+// TestProtoFieldTypeRefs_LabelIsNeverTheTarget.
+//
+// map<K, V> IS UNWRAPPED, and the rule is: the wrapper is never a target and
+// each of K and V is considered independently. proto3 constrains a map key to an
+// integral or string scalar, so in practice only V ever yields a candidate — but
+// both are walked rather than assuming the spec, exactly as namedTypeRefs does.
+// `map<string, Order>` therefore yields [Order]: one candidate, not two, and not
+// a `map` target. All three branches are graded by
+// TestProtoFieldTypeRefs_MapValueIsTheTarget.
+//
+// THERE IS NO SCALAR BLOCKLIST HERE, and its absence is the design, not an
+// omission. `string`, `int32` and the rest are the largest field population in
+// any .proto file and they must produce zero edges — but they already do,
+// because no .proto declares `message string`, so the in-file gate drops every
+// one of them. A protoScalars check in front of that gate would fire only where
+// the gate already fires, leaving BOTH ungraded (arm A refused the identical
+// blocklist for the identical reason). It was written and then removed after
+// mutation scoring made the masking explicit: deleting the check changed no
+// end-to-end behaviour on any fixture, which is what "redundant" means.
+// TestProtoFieldTypeRefs_EveryScalarProducesNoEdge enumerates the whole
+// production scalar set through Extract and asserts the observable consequence
+// rather than the mechanism.
+//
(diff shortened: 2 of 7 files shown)


cajasmota:

One blocking finding: **deleting the scalar blocklist opens a silent mis-bind on legal, protoc-clean proto3.** Everything else in this PR I could check, I checked, and it holds — including the two things I most expected to break (the hand-edited baseline, and the 0-dangling claim).

All figures below measured in my own worktree at `7d8bcc803` (`…/scratchpad/review-6991-k4m8p/wt`) and a second at the parent `a3e0e8e41` (`…/review-6991-k4m8p/parent`), macOS/APFS, `HOME`+`GRAFEL_HOME`+`GRAFEL_DAEMON_ROOT` all redirected to scratch, `GOMAXPROCS=3`. Corpora read-only from `/Users/jorgecajas/Projects/archigraph-corpora`.

---

### BLOCKING — the deleted `protoScalars` blocklist was not redundant

The header block states the design premise: *"no `.proto` declares `message string`, so the in-file gate drops every one of them."* That is a **corpus-relative zero**, not a property of the pass, and the gate does **not** drop them. Probe through the real `Extract`, in my worktree:

```proto
syntax = "proto3";
message string { int32 v = 1; }
enum bool { B_ZERO = 0; }
message Holder { string name = 1; bool flag = 2; int32 n = 3; }
```
```
Holder.flag -> scope:schema:message:proto:shadow.proto:bool
Holder.name -> scope:schema:message:proto:shadow.proto:string
```

`message string` is in the local set, so `local[ToID]` is true and the edge **survives and binds**. It never reaches `bug-extractor`. This is precisely the failure mode the PR's own header calls out for `namedTypeRefs`' qualifier-stripping — *"silently, because a bound edge never reaches bug-extractor"* — reintroduced two paragraphs later under the redundant-guard argument.

**It is legal proto3, and protoc disagrees with the graph.** I installed `grpcio-tools` in a scratch venv and compiled it:

- `protoc --descriptor_set_out` on that file: **RC 0** (both the `message string` and the `enum bool` shapes).
- The descriptor for `Holder.name` is `type=TYPE_STRING, type_name=''` — protoc binds the **scalar**, with no message reference at all. The grammar matches `"string"` in field-type position as a scalar before it ever considers `messageType`, so the shadowing message can never be the referent.

So grafel now emits an edge protoc says does not exist, and it binds. Two aggravations:

- `map<string, Order>` under that declaration yields **two** edges — `-> Order` and a bogus `-> string` key edge.
- `namedTypeRefs`, the message-anchored path, **keeps** its `protoScalars` check, so the two anchors now disagree in the wrong direction: the field edge over-fires exactly where the message edge correctly stays silent.

`TestProtoFieldTypeRefs_EveryScalarProducesNoEdge` passes only because its generated fixture declares no shadowing message — the enumeration is over the scalar set, but the fixture space around it was hand-picked. Cross-file shadowing (`message string` in an imported file) *is* correctly silent; I checked that separately.

M1's verdict should be re-read as **alive, production-reachable, input named** rather than redundant. The fix is to restore the check in `protoFieldTypeCandidates` — and it is then gradeable end-to-end, which answers the objection that killed it: the three probes above are the grading, roughly 25 lines, and they kill M1 through `Extract` rather than at the unit.

---

### (a) The hand-edited baseline is **exactly right**, and the discarded deltas **are** environmental — proven

I ran `scripts/quality/run.sh --runs 1 --update-baseline` twice on this machine, back to back, in two clean worktrees: once at the PR head, once at the untouched parent `a3e0e8e41`. **The entire diff between the two generated baselines:**

```
     "proto-mini": {
-      "relationship_found": 17,   "relationship_expected": 17,
+      "relationship_found": 18,   "relationship_expected": 18,
-      "relationship_extracted_total": 55
+      "relationship_extracted_total": 56
```

Nothing else. `proto-mini +1` is genuinely the only delta attributable to this change, and the committed hand-edit is byte-identical to what a regeneration produces (proto-mini does not appear in the generated-vs-committed diff at all).

The ceiling drops are present **identically at the parent**, so the discard was correct:

| fixture | key | committed | gen@parent | gen@HEAD |
|---|---|---|---|---|
| csharp-aspnet-core-mini | entity/rel_extracted_total | 57 / 110 | 49 / 94 | 49 / 94 |
| csharp-hangfire-mini | " | 36 / 66 | 30 / 60 | 30 / 60 |
| csharp-quartz-net-mini | " | 23 / 45 | 19 / 41 | 19 / 41 |
| docker-compose-services-mini | " | 48 / 68 | 47 / 65 | 47 / 65 |
| python-django-mini | rel_extracted_total | 255 | 251 | 251 |
| python-fastapi-mini | " | 28 / 57 | 27 / 54 | 27 / 54 |
| python-rq-mini | " | 22 / 43 | 20 / 39 | 20 / 39 |
| razor-blazor-mini | " | 25 / 35 | 22 / 32 | 22 / 32 |

Two corrections to the PR body, neither material: it is **eight** fixtures across four languages, not "several unrelated C# fixtures" — C#, docker-compose, Python ×3, Razor. And every one of them is a `*_extracted_total` ceiling; no `_found`/`_expected` recall figure moved anywhere. `--ratchet` does not fail on a ceiling drop (I ran it: **exit 0, 38 gated fixtures, 29 at 100%**, reproducing the PR's figure), which is what makes committing the regeneration a silent weakening rather than a visible failure. Discarding it was right.

### (b) The second edge does not duplicate or inflate anything I can find

Orphan is per-entity (`internal/quality/analytics/scan.go`), so the message was already non-orphan and the *field* moves from leaf to non-leaf — the population #6726 measures. No edge-per-message counter double-counts: the two rows differ in `From` (field vs message) and in `ref_kind`. Disposition gains one endpoint per field edge, and — same-file only — every one of them resolves, so it moves the resolved numerator and denominator together and cannot inflate `bug-extractor`.

### (c) Safe under `message string`? **No.** See the blocking finding.

### (d) The rule is **same-file**, not resolvable-anywhere — by construction

`dropUnresolvableTypeRefs` builds `local` from the `entities` slice of the **current** `Extract` call, and `Extract` sees one file. There is no repo-wide index in reach, so a cross-file type cannot be kept. Observed as well as read: `TestProtoFieldTypeRefs_CrossFileTypeGetsNoEdge` passes both files but extracts each independently, and my own cross-file shadow probe is silent. Confirmed also: **one** address dialect — `User.role` resolves to `SCOPE.Schema/enum:Role` through the same `messageTypeRef` form.

### (e) F5 is a real forbidden row, not a comment

Probed through the production indexer on the proto-mini golden. At HEAD: `rel 18/18, ent 19/19, forbiddenHits=0`. With the in-file gate bypassed for `field` records:

```
forbiddenHits=1
User.address -REFERENCES-> scope:schema:message:proto:user.proto:Address | note="F5, the FIELD-anchored twin of…"
```

It fires. So does the expected row: 18/18 is the graded figure, not an asserted constant.

### (f) The corpus table and 0-dangling reproduce exactly

Walked the five corpora with the registered extractor and asserted every emitted `ToID` is in the same file's message/enum address set:

```
etcd             files=5  fields=28  edged=3  (10.7%)  edges=3   dangling=0
apollo-server    files=1  fields=125 edged=48 (38.4%)  edges=48  dangling=0
grpc-go-examples files=3  fields=4   edged=0  (0.0%)   edges=0   dangling=0
actix-examples   files=1  fields=2   edged=0  (0.0%)   edges=0   dangling=0
gin              files=1  fields=0   edged=0  (0.0%)   edges=0   dangling=0
TOTAL            files=11 fields=159 edged=51 (32.1%)  edges=51  dangling=0
```

Every figure matches the PR body. Note that 0-dangling is structural here (the gate keeps only what is in `local`), which is why it cannot see the scalar-shadow bug — the wrong edge binds.

### (g) The `worktree_seed_parity_test.go` characterisation is **correct**, and it does generalise

`semanticDigest` is computed twice at runtime and compared to itself via `assertParity(want, got)`. There is **no** hardcoded digest literal anywhere in the file (`grep -nE '"[0-9a-f]{16,}"'` returns nothing). An unchanged result there is not evidence of an unchanged graph. Worth stating somewhere more durable than this PR.

---

### My mutants (4 new; M1–M7 not re-run)

Each applied by a script asserting **exactly one** occurrence, with the enclosing function named and the hit line printed; `go vet` before every score; green baseline immediately before each (`./internal/extractors/proto/ ./internal/resolve/ ./internal/quality/...` green, `./cmd/grafel/` green at 149.6s); always `-count=1`; each restore by `cp` from a `shasum -a 256`-verified backup (`git checkout` never used).

| # | mutant (enclosing function) | verdict |
|---|---|---|
| R1 | `field_name` property value → a constant (`protoFieldTypeRefs`) | **DISTINGUISHABLE AND UNGRADED.** Full proto + resolve suite green. No test reads this property's value; `ftrEdges` filters on `ref_kind` and formats `name -> ToID` only. The PR body claims "Plus `field_name` and `target_type`, as arm A" — prose asserting what no test observes. Priced: ~2 lines, fold both values into `ftrEdges`' formatted string |
| R2 | `target_type` property value → a constant (`protoFieldTypeRefs`) | **DISTINGUISHABLE AND UNGRADED.** Same evidence, same 2-line fix |
| R3 | qualified-name guard weakened: `Contains(p, ".")` → `Contains(p, "..")`, so a qualified name passes through **unstripped** (`protoFieldTypeCandidates`) | **DEAD**, but by `TestProtoFieldTypeCandidates_Rules` alone — every end-to-end assertion stayed green, because `…:demo.Profile` is not in `local` and the gate drops it. The `.`-rejection is end-to-end redundant in the *pass-through* direction; only M5's *stripping* direction is observable. Not a defect — the unit test is the correct guard here. Noting it because it is the same masking shape that was used to justify deleting the scalar blocklist, where the conclusion does not hold |
| R4 | drop the `p == ""` guard, admitting an empty-named target (`protoFieldTypeCandidates`) | **DEAD** by `TestProtoFieldTypeCandidates_Rules`; masked end-to-end for the same reason. Reachable only through degenerate `map<, >` text |

Minor, non-blocking: `fieldEnt.Relationships = protoFieldTypeRefs(...)` in `buildMessage` is an assignment, not an append. `buildField` sets no `Relationships` today so it is correct now, but it silently clobbers anything a future `buildField` adds. `append` costs nothing.

---

**To unblock:** restore `protoScalars` in `protoFieldTypeCandidates` and grade it with the shadowing fixture (an edge case that compiles under protoc is not a redundant guard), and assert the two edge property values. Everything else — the baseline edit, the environmental discard, the same-file rule, F5, the corpus table, the parity-digest observation — I verified independently and it stands.

Did not merge, did not push.

cajasmota:
Blocker accepted in full — you were right, and the premise I built on was mine, not the tooling's.

**`protoScalars` is restored** (`e8d5020f3`), and I reproduced your external fact independently rather than taking it on report: `grpcio-tools` in a scratch venv, `--descriptor_set_out` on your fixture, **RC 0**, and the descriptor reads

```
Holder.name: type=TYPE_STRING type_name=''
Holder.flag: type=TYPE_BOOL   type_name=''
Holder.real: type=TYPE_MESSAGE type_name='.Order'
```

Empty `type_name` on both shadowed fields — protoc binds the scalar and never considers the shadowing declaration, so the edge I was emitting asserts a reference protoc says does not exist, and it binds, so nothing surfaces it.

The guard is now graded **through `Extract`**, not only at the unit table: `TestProtoFieldTypeRefs_ScalarShadowedByASameFileMessage` uses your fixture (plus `enum bool` and the `map<string, Order>` key row), with a positive control asserting the shadowing entities really exist so the gate's local set really contains them. Deleting `protoScalars[p]` now turns it red as well as the unit table — re-scored, both fail.

**The lesson is in the code, not just the body.** `field_type_refs.go` records that a mutant killed only by its own unit test does not establish redundancy — it is equally consistent with the end-to-end fixture space lacking the case, and only constructing the case tells them apart. Your R3/R4 are the honest other half of that: both are DEAD by the unit table and genuinely masked end to end, so the unit table is the right grader there and I have left them alone. `TestProtoFieldTypeCandidates_Rules` and `_EveryScalarProducesNoEdge` now each say which of the two situations their rows are in, and the enumeration test says explicitly that it enumerates the scalar *set* and not the fixture space around it — which is the hole you walked through.

**R1/R2 — both DEAD** (8 tests each). `ftrEdges` now cross-checks `field_name` against the record's dotted Name and `target_type` against the ToID's trailing segment on **every** edge in the file, so the invariant travels with every test rather than one; `TestProtoFieldTypeRefs_EdgePropertiesAreSpelledOut` additionally pins two full property lists against an independent literal, including `User.by_id`, the one case where `target_type` is not the declared type string.

**`append`, not assign** — done, with the reason in a comment.

**Body updated with your two corrections**: eight fixtures across four languages (C# ×3, docker-compose, Python ×3, Razor), all `*_extracted_total` ceilings with no recall figure moving. And your parent-vs-head double regeneration is now the stated reason the discard was right — that committing it would have been a *silent* weakening, since `--ratchet` does not fail on a ceiling drop, is a better argument than the one I had. Your `semanticDigest` generalisation is in the body too.

Re-verified after the change: `./internal/extractors/proto/`, `./internal/resolve/`, `./internal/quality/...` green; `./cmd/grafel/` green (152.6s); ratchet **exit 0**, 38 gated / 29 at 100%, proto-mini 18/18, total 56, 0 forbidden. Corpus figures re-measured with the guard back: **unchanged** at 159 fields / 51 edged / 32.1% / 0 dangling — which is itself the point, since no corpus file shadows a scalar name and the corpus therefore never could have found this.

Not merged.

cajasmota:
## Coordinator mutant CI-1 on `e8d5020f3` — DEAD

Own detached worktree, `git status --porcelain` empty at start and after restore, restored by `cp` from a SHA-256-verified backup (`7cc2d22d…` both ways), zero `MUTANT` text remaining, green baseline immediately before. macOS/APFS.

**CI-1 — convert the qualified-name REJECTION into DESCENT.** Eleven mutants have been scored here. The qualified-name guard has been attacked twice, both by the reviewer, and both times *within* rejection semantics: R3 weakened it to `".."`, R4 dropped the empty-part guard. **Nobody changed what the guard does** — and descent is the specific thing arm A refused for C#, so it is also the cross-language consistency question.

```go
-  if p == "" || protoScalars[p] || strings.Contains(p, ".") {
-      continue
-  }
+  if i := strings.LastIndex(p, "."); i >= 0 { // MUTANT CI-1: descend qualified names instead of rejecting
+      p = p[i+1:]
+  }
+  if p == "" || protoScalars[p] {
+      continue
+  }
```

So `etcdserverpb.Member` becomes `Member` and binds if any same-file message is named `Member` — which is exactly etcd's population, the corpus where you measured 3/28 (10.7%) *because* its types are overwhelmingly cross-file. I chose it expecting the honest floor of a same-file rule to be graded by prose and by a corpus number rather than by an assertion. The scalar and empty-part disjuncts were left intact so the mutant cannot be masked by them.

Edit proven landed: exactly 1 occurrence, hit set asserted `[194]`, **enclosing function asserted** to be `protoFieldTypeCandidates`, 0 occurrences of the original afterwards, `go vet` clean before scoring.

**Result: DEAD** — `./internal/extractors/proto/` fails five tests, `./internal/resolve/` stays green:

- `TestProtoFieldTypeRefs_QualifiedTypeIsNotStrippedToASameFileName`
- `TestProtoFieldTypeCandidates_Rules`
- `TestProtoFieldTypeRefs_EmittedEdges`
- `TestProtoFieldTypeRefs_ForbiddenTargets`
- `TestProtoFieldTypeRefs_ResolverBindsEveryEdge`

The first of those is a **forbidden row built for precisely this mutation**, by name. So the floor is not prose: "a qualified type is not stripped to a same-file name" is asserted as behaviour, at the emission layer and again through the resolver, and it is graded end-to-end rather than only in the unit table. That is the distinction this PR's blocking finding was about, applied to the neighbouring guard.

## On the round

Restoring `protoScalars` was right and the way it was re-graded is what matters: through `Extract`, with the shadowing entities asserted to exist so the gate's local set genuinely contains them, plus the `enum bool` and `map<string, Order>` key rows. Deleting the guard now fails that test **and** the unit table — the two situations the earlier mutant could not distinguish are now distinguished by construction.

Reproducing the reviewer's protoc result independently (RC 0, `Holder.name: TYPE_STRING type_name=''`) rather than accepting it was the right call. protoc denying the reference is the fact that makes this a wrong edge rather than a debatable one, and it is an external check neither of you had to take on trust.

**The corpus figure being unchanged with the guard restored — 159 fields, 51 edged, 32.1%, 0 dangling — is the point, not a footnote.** No corpus file shadows a scalar, so the corpus never could have found this, and a number that does not move is exactly what a corpus-relative zero looks like from the inside.

Merging on green, once the re-triggered macOS leg reports.

Issue #6779, opened 2026-09-02
an entity-kind rename silently orphans every indexed graph — BLOCKS the #6776 migration; SchemaVersion does not move and nothing reports the mismatch


cajasmota:
Found while writing the v0.3.1 release notes. Not a defect introduced by any one change — a missing mechanism that two changes in this release are the first to actually need.

## The situation

v0.3.1 renames entity kinds and entity names:

- **#6451** retired `SCOPE.ExternalAPI` and split it into `SCOPE.ExternalEndpoint` and `SCOPE.IngressHost`. Verified on `main`: `internal/types/kinds.go:52-53` declares the two new kinds, `:417-419` registers them, and no Go producer emits the old string any more — only comments reference it.
- **#6499** made Kotlin `SCOPE.Operation` names class-qualified (`<Class>.<method>` instead of a bare method name).

**Nothing migrates or invalidates an already-indexed graph.** A group indexed under v0.3.0 keeps entities kinded `SCOPE.ExternalAPI` and Kotlin operations under bare method names, indefinitely, until something causes those files to be re-extracted. A query filtering on `SCOPE.ExternalEndpoint` returns nothing against that graph, and there is no signal anywhere that the graph is speaking an older vocabulary.

## What I checked, and where I stopped

- `graph.SchemaVersion` (`internal/graph/graph.go:22`) is **1** and was not bumped by either change. It is the on-disk `graph.json` schema version and is not consulted on a kind rename.
- Staleness is content-driven: `internal/daemon/stale_reindex.go` guards *how often* a stale repo may be re-admitted, not *whether* a version change makes one stale.
- I searched `internal/` and `cmd/` for `needsReindex` / `isStale` / `shouldReindex` and for version-keyed invalidation in `internal/registry` and `internal/daemon`, and found none.

**Boundary:** those are scoped searches, and I have been caught twice this week by concluding "nothing does X" from one. If a version-keyed invalidation exists somewhere I did not look, this issue is wrong and should be closed with the pointer. **Please verify before implementing** — the check is cheap: index a group, rename a kind, restart, and see whether the old kind survives.

## Why it is worth a mechanism rather than a release note

The release note ("reindex your groups") is the right stopgap and is in the v0.3.1 notes. It is not a fix, because:

1. **It is silent.** A user who misses the note gets empty query results with no error, from a graph that looks healthy. That is the same shape as the defects this milestone keeps finding — a state indistinguishable from "there is nothing there".
[...]

Issue #6757, opened 2026-09-01
the relationship-kind vocabulary is enforced by nothing — IsValidRelationshipKind has zero non-test callers, so an undeclared PRODUCES reaches the graph unchallenged


cajasmota:
Found while grounding #6741.

## `IsValidRelationshipKind` has zero non-test callers

```
$ grep -rn "IsValidRelationshipKind" internal cmd | grep -v _test.go
internal/types/kinds.go:1613:// IsValidRelationshipKind reports whether s is one of the typed RelationshipKind values.
internal/types/kinds.go:1614:func IsValidRelationshipKind(s string) bool {
```

Its own definition and its doc comment. Nothing else.

So the relationship-kind vocabulary is **enforced by nothing**. `AllRelationshipKinds()` is a list that describes an intention rather than a constraint, and any string can be written to the graph as a relationship kind.

This is the most complete instance of this repo's most-repeated defect class. The usual shape is a gate too weak to catch something. Here the gate is **not invoked at all** — it cannot fail because it never runs.

## The instance that exposed it

`internal/custom/java/quartz.go:274` emits `RelationshipType: "PRODUCES"`, plumbed to a real graph edge via `patterns_dispatch.go:363-394`. **`PRODUCES` is not in `AllRelationshipKinds()`**, so `IsValidRelationshipKind("PRODUCES")` returns `false` — and nothing asks.

Four consumers hard-code the same undeclared string and would traverse it: `internal/mcp/flow_tools.go:2024-2025`, `internal/links/reachability.go:79-80`, `internal/mcp/dead_code.go:182`, `internal/dashboard/topology_compound.go:235`.

`PRODUCES` is very unlikely to be the only one.

## A second contradiction in the same area

`docs/adrs/0003-scope-entity-taxonomy.md:29` lists `CONSUMES_QUEUE` inside what it calls a **closed enum**. `internal/types/kinds_test.go:98-99` asserts the opposite in as many words:

> `"CONSUMES_QUEUE has no producer; must NOT be valid"`

One of those has been wrong since it was written, and nothing forced the question because the vocabulary is unenforced either way.

## Scope to establish FIRST

**Enumerate every relationship-kind string actually written to the graph and diff it against `AllRelationshipKinds()`** before wiring anything. The size of that undeclared set decides what this issue is:

- a short list → wiring the validator is close to a one-line fix plus a few declarations;
[...]

PR #6992, opened 2026-09-07
fix(#6906): delete TYPED_AS and HAS_TYPE — two names for one edge #6984 already ships as REFERENCES/ref_kind=field_target_type, and a comment describing the producer that was never built

### LINKED ISSUE #6906: DECIDED against TYPED_AS (#6984 adopted REFERENCES+ref_kind=field_target_type): now a cleanup — delete or reserve TYPED_AS and HAS_TYPE, and kill the comments describing a producer that will never exist

cajasmota:
Closes #6906. Refs #5828, #6912, #6984, #6757, #6451, #6726, #6975.

`TYPED_AS` and `HAS_TYPE` were both declared in `internal/types/kinds.go`, both registered in `AllRelationshipKinds()`, and both named the same concept: a field or parameter → its declared type. #6984 shipped that edge for C# as `REFERENCES` carrying `ref_kind: "field_target_type"` — the spelling the custom lane had already shipped in five languages via `referencesClassEdge`. So both kinds became **deliberately** producerless, which is worse than an oversight: nothing a reader of `kinds.go` could see distinguished "decided against" from "someone forgot", and three separate issues (#5828, #6906, #6912) were filed rediscovering the same gap.

**No producer is added here.** The decision against both kinds is already made.

## Outcome chosen: DELETE both, identically

Both constants and both `AllRelationshipKinds()` entries are gone. The alternative — keeping them marked "reserved" — was rejected because `TestAllRelationshipKinds_CoversEveryDeclaredConstant` (`internal/types/kinds_enum_completeness_6757_test.go:157-160`) already states the rule in as many words: *"There is no legitimate reason to declare a RelationshipKind and withhold it from the vocabulary: if a kind is to be retired, delete the constant."* A reserved marker would have to be either registered (still valid, still a third name a producer could pick up) or declared-and-unregistered (which that guard fails). Deletion is the only state the existing guards accept.

The decision is recorded in prose immediately above `AllRelationshipKinds()` as a **RETIRED KINDS** note, so the next person does not re-derive it, and pinned by a test so the note is not the only thing holding it.

## Step 1 — what removal would break, established BEFORE removing anything

Measured in the worktree at `.../scratchpad/impl-6906-k3p8w/wt`, branched from `origin/main` = `3c7d2c5ec`. macOS/APFS.

**The #6757 "no non-test caller" claim, re-verified on current `main` rather than inherited.** `grep -rn IsValidRelationshipKind` over the whole tree returns callers in **test files only**, plus its own definition (`kinds.go:2093`) and its call from `IsDeclaredRelationshipKind` (`derived_kinds.go:91`). `IsDeclaredRelationshipKind` in turn has **no non-test caller** either — only comments in `derived_kinds.go` and `graph/fbwriter/non_enum_kinds.go` reference it. So the predicate rejects nothing at runtime, and dropping two entries cannot drop an edge.

**Every consumer of the valid-kinds slice, enumerated.** `AllRelationshipKinds()` has exactly one non-test production consumer: `enumKindSet` in `internal/graph/fbwriter/non_enum_kinds.go:229-236`, the #6757 arm-C write-path counter. It reports how many written edges carry a kind *absent* from the enum. Removing two entries changes that report only if an edge of either kind is written — nothing produces one, so the counter is unaffected. Every other reference to the accessor in non-test code is a comment.

**"Zero producers" is pinned by a test, not by my grep** — the strongest evidence here, surfaced by review. `TestProducerRelationshipKindLiterals_AreValid` (via `internal/relkinds`) scans producer source for emitted kind literals and machine-checks them against the vocabulary. Review demonstrated it: unregistering a kind that *does* have a producer (`RelationshipKindSupervises`) goes red on that test. So a produced kind cannot be deleted silently, and the suite — not the author — is what certifies these two had no producer.

**Compile-time proof nothing referenced the constants:** `go build ./...` and `go vet ./...` both exit 0 across the whole module with both deleted.

### On-disk safety: deleting an enum member cannot renumber persisted kinds

Established at runtime by review, not by inference — and `go build`/`go vet` could not have supplied it, so it is recorded here as the load-bearing fact.

`internal/graph/schema/graph.fbs` declares **zero** `enum` types; `Entity.kind` and `Relationship.kind` are both `kind: string`. `fbwriter`'s `enumKindSet` is an in-memory `map[string]struct{}` from `sync.OnceValue`, never serialised, so it has no ordinal at all. Kinds persist **by name**.

Proven end-to-end: a `graph.fb` written from the **parent** tree with 112 relationships — one per parent `AllRelationshipKinds()` entry, including `HAS_TYPE` at index 54 and `TYPED_AS` at index 100 — read back with **this branch's** `fbreader.Open` + `IterateRelationships` gives an empty index+kind diff, exit 0. Every kind string survives byte-for-byte at every position, including the two deleted ones. Old graphs read correctly.

`types.KindVocabularyVersion` (#6779) is correctly not bumped: its rule is scoped to **entity** kinds, and with zero producers no stored graph can contain either edge anyway.

**`cmd/grafel` whole-graph digest and the golden ratchet: both held.** `go test -count=1 ./cmd/grafel/` → `ok, 152.0s`. `scripts/quality/run.sh --ratchet` → *"38 gated fixtures held their baseline (29 at 100% must-have recall)"*, exit 0. Nothing moved, so there is no delta to enumerate — consistent with the removal touching no producer.

### The enumeration was incomplete on its first pass — corrected

The first version of this body said "no rule YAML, no dashboard, MCP or webui reference". **That was false**, and review caught it. `site/src/scripts/coverage.js` listed `"has_type"` and `"typed_as"` in the public coverage page's kind roster, so the page would have advertised two kinds that no longer exist — the exact false advertisement #6906 was filed to end, shipping on this PR's own Cloudflare Pages deploy.

The miss has one cause worth naming: I grepped the **uppercase** literals and the Go constant names; the site uses the **lowercase display form**. A name has more than one spelling — constant, string literal, lowercase display, snake_case, serialised alias — and an enumeration that covers one spelling has not covered the name. Both items are removed, and a guard now scans that file so the next spelling-mismatch fails a test instead of a review.

The stale `(60)` / `(106)` counts in that file's comment are gone rather than restated. They were presented as the sizes of `AllEntityKinds()` / `AllRelationshipKinds()` but actually described the arrays themselves (the accessors return 94 and 110 on this branch), and an unchecked number is how the drift went unnoticed. The comment now says plainly that the arrays are hand-maintained and a subset, with the accessors as the only authority. Fixing the wider drift is out of scope for #6906.

**Complete surface list, after the correction:** the four Go files in this diff plus `site/src/scripts/coverage.js`. A case-insensitive `--binary-files=text` grep over the whole tree finds nothing else — no golden fixture, no `.json`/`.yaml`, no binary artefact, no reflection or string-keyed kind lookup.

## Step 3 — the comments, fact-checked sentence by sentence

`internal/extractor/enum_valueset.go` described a `TYPED_AS` binding whose producer half never existed. Rewritten to say what is true. Each claim and the file it was checked against:

| Claim now in the file | Verified against |
|---|---|
| the inbound edge is `REFERENCES` with `ref_kind="field_target_type"` | `internal/extractors/csharp/field_type_refs.go` — `Kind: string(types.RelationshipKindReferences)` and `{K: "ref_kind", V: csFieldTargetRefKind}` in `attachCsharpFieldTypeRefs`; `csFieldTargetRefKind = "field_target_type"` |
| of the six `field_target_type` producers, only C#'s addresses its target by QualifiedName | the other five: `custom/python/helpers.go:104` `pyClassRef`, `custom/golang/helpers.go:103` `goClassRef`, `custom/java/orm_helpers.go:52` `javaClassRef`, `custom/javascript/helpers.go:71` `"Class:" + targetClass`, `custom/ruby/helpers.go:89` `rbClassRef` — every one a `Class:<target>` stub |
| the producer reads the `SCOPE.Enum` record's `QualifiedName` directly | `csInFileTypeTargets`: `case r.Kind == "SCOPE.Enum": put(r.Name, r.QualifiedName)` |
| the QN binds via the `byQualifiedName` exact-match tier, ahead of the other tiers | `internal/resolve/refs.go` — the `idx.byQualifiedName[stub]` probe at `:2292`, at the top of `LookupStatusHint` (`:2268`) |
| `TYPED_AS` no longer exists | this diff |

Line numbers were deliberately **not** written into the new comments — the ones cited by the source issue had already rotted (`refs.go:2103` is no longer the probe on `3c7d2c5ec`). The comments name functions; the line numbers live here, in a PR that does not drift.

**One sentence was withdrawn on review.** The first version said *"in every other language a `SCOPE.Enum` node still has no inbound edge"* — an unqualified absolute over every language and every edge kind, which nothing observes and which I had not checked. It is replaced by the bounded claim in the table above (the six `field_target_type` producers, all five alternatives enumerated). Deleted rather than hedged: that sentence is the exact class of defect this PR is cleaning up.

**A newly-false comment was also fixed.** `internal/extractors/csharp/field_type_refs.go:22` said `TYPED_AS` and `HAS_TYPE` *"**are** both declared in internal/types/kinds.go"* — true when #6984 wrote it, false the moment this branch merges. The first version of this body called that line accurate; it is now past tense and names #6906 as where they went. A PR whose thesis is "a comment described a binding nobody built" must not leave a comment describing constants nobody declares.

**A comment-only diff has no mutant** — every mutation of a comment is equivalent by construction, so none was manufactured for that part. The table above is the verification instead.

## The guards, and proof they fire

`internal/types/kinds_retired_6906_test.go` holds two tests.

`TestRetiredRelationshipKindsStayRetired` fails if either name reappears on **all three Go surfaces** a re-addition could use: a const declaration in `kinds.go` (read from source via the existing AST walk, so a constant declared under *any* name is caught by its value), membership of `AllRelationshipKinds()`, and `IsValidRelationshipKind`.

`TestRetiredRelationshipKindsAreGoneFromTheCoveragePage` extends it to the surface a Go-only enumeration cannot see: a substring scan of `site/src/scripts/coverage.js` for the lowercase display forms. It is deliberately a raw substring scan, not a JS parse — the spelling to catch **is** the raw token, and a parser would add a way to be silently wrong about a file with no Go meaning.

Both compare against **independent string literals** (`retired6906Kinds = []string{"TYPED_AS", "HAS_TYPE"}`), never against the production slice — a guard that ranges over `AllRelationshipKinds()` and asks whether it contains what `AllRelationshipKinds()` contains cannot observe a re-addition (#6975). The file says so, so nobody "simplifies" it into a constant reference.

**Non-vacuity is asserted before any absence is believed**, on both. The AST guard requires a non-empty parse, two independent controls (`CALLS`, `REFERENCES`) **present** in it, and `IsValidRelationshipKind` **true** for both controls. The page guard requires three quoted controls (`"references"`, `"inherits"`, `"contains"`) present in the bytes it read, and treats a missing file as a hard failure rather than a skip.

**Proven to fire, five times, each mutant isolating one surface.** `gofmt`/`go vet` clean before each score, green baseline (`--- PASS`) immediately before, `-count=1`, occurrence count and enclosing scope asserted for every edit, and restore by `cp` from `shasum`-verified backups (`kinds.go` `10bdf67744…`, `coverage.js` `b6e8905eca…`, `kinds_retired_6906_test.go` `bfd1f648f5…`), never `git checkout`. All in the more-permissive direction:

| # | Mutant (permissive) | Edit landed | Result |
|---|---|---|---|
| M1 | re-declare `RelationshipKindTypedAs = "TYPED_AS"` **and** register it | 2 occurrences: `kinds.go:1893` in the `const` block, `:1912` inside `AllRelationshipKinds` | **DEAD** — all three assertions fired (declaration, registration, predicate) |
| M2 | declare `RelationshipKindHasType = "HAS_TYPE"`, do **not** register it | 1 occurrence, `kinds.go:1893`, in the `const` block | **DEAD** — the source-declaration assertion fired alone |
| M3 | register `RelationshipKind("TYPED_AS")` as a raw literal with **no** constant | 1 occurrence, `kinds.go:1911`, inside `AllRelationshipKinds` | **DEAD** — registration and predicate fired; the AST surface correctly did not |
| N5 | put `"typed_as"` back in the site roster | 1 occurrence, `coverage.js:34`, inside `REL_GROUPS`' "Structural" item array; `node --check` clean | **DEAD** — the page guard named the file and the item |
| N6 | point `coverageJSPath` at `site/astro.config.mjs` — a file that **exists and reads fine** | 1 occurrence, the `const coverageJSPath` declaration | **DEAD** — control `"references"` absent from 206 bytes read. This is the "reads the wrong file" no-op a count floor cannot catch |

M2/M3 are the pair that matters for the Go guard: each defeats one surface and is caught by another, so no assertion is carried by its neighbour. N5/N6 are the same pair for the page guard — one grades the content check, one grades the non-vacuity check.

Review scored four more (N1–N4) on the parent-vs-branch axis, including `N4` (reorder `AllRelationshipKinds()`) → **equivalent**, correctly so given that nothing persists an ordinal, and `N2` (re-add a kind *and* shrink the guard literal) → **ALIVE**, expected because it edits the pin itself. `N2` named the honest limit — the retirement rested on one string list with nothing cross-checking the non-Go surfaces — and `TestRetiredRelationshipKindsAreGoneFromTheCoveragePage` is the hardening it asked for.

## Verification

- `go test -count=1` on every touched package and its neighbours: `internal/types`, `internal/extractor`, `internal/graph/...` (incl. `fbwriter`), `internal/relkinds`, `internal/extractors/csharp` — 0 FAIL lines, exit 0.
- `go test -count=1 ./cmd/grafel/` — `ok 152.0s` (whole-graph digest pin).
- `scripts/quality/run.sh --ratchet` — 38 fixtures held, exit 0. No baseline constant was touched.
- `go build ./...`, `go vet ./...`, `gofmt -l internal/` — all clean, exit 0.
- `site/src/scripts/coverage.js` is a plain IIFE with no lint or test leg: no eslint/prettier config in `site/`, and no workflow in `.github/workflows/` builds it — only the Cloudflare Pages `astro build` does. Checked with `node --check` (exit 0) before and after every edit.
- Ran under `GRAFEL_HOME` / `GRAFEL_DAEMON_ROOT` pointed at a scratch directory; the real `~/.grafel` was not written.

`REFERENCES`, `ref_kind`, and everything #6984 shipped are untouched.

### DIFF
--- a/internal/extractor/enum_valueset.go
+++ b/internal/extractor/enum_valueset.go
@@ -22,11 +22,15 @@ import (
 //     value-for-value by a NestJS `enum Status { ACTIVE = 1, ARCHIVED = 2 }`.
 //   - "which fields are constrained to enum X?" — inbound REFERENCES edges
 //     carrying the property ref_kind="field_target_type", emitted from a field
-//     record to this node's QualifiedName. C# is the only language that emits
-//     them (internal/extractors/csharp/field_type_refs.go, #6984); in every
-//     other language a SCOPE.Enum node still has no inbound edge. No TYPED_AS
-//     edge exists: that kind was declared, never produced, and deleted by
-//     #6906 — see the RETIRED KINDS note in internal/types/kinds.go.
+//     record to this node's QualifiedName. Of the six
+//     ref_kind="field_target_type" producers, only C#'s
+//     (internal/extractors/csharp/field_type_refs.go, #6984) addresses its
+//     target by QualifiedName; the five in internal/custom/{python,golang,
+//     java,javascript,ruby} emit a `Class:<target>` stub instead, so C# is the
+//     only one that names a value-set node the way this helper spells it.
+//     No TYPED_AS edge exists: that kind was
+//     declared, never produced, and deleted by #6906 — see the RETIRED KINDS
+//     note in internal/types/kinds.go.
 //
 // Each language extractor detects the enum shape (Python Enum subclass, TS
 // `enum` / string-literal union, Java enum, Go iota const block, Ruby
--- a/internal/extractor/enum_valueset.go
+++ b/internal/extractor/enum_valueset.go
@@ -20,8 +20,13 @@ import (
 //     the full member list AND each member's literal value, so a Django
 //     Python `class Status(IntEnum): ACTIVE = 1; ARCHIVED = 2` is reproduced
 //     value-for-value by a NestJS `enum Status { ACTIVE = 1, ARCHIVED = 2 }`.
-//   - "which fields are constrained to enum X?" — the inbound TYPED_AS edges
-//     from fields/params declared with the enum type.
+//   - "which fields are constrained to enum X?" — inbound REFERENCES edges
+//     carrying the property ref_kind="field_target_type", emitted from a field
+//     record to this node's QualifiedName. C# is the only language that emits
+//     them (internal/extractors/csharp/field_type_refs.go, #6984); in every
+//     other language a SCOPE.Enum node still has no inbound edge. No TYPED_AS
+//     edge exists: that kind was declared, never produced, and deleted by
+//     #6906 — see the RETIRED KINDS note in internal/types/kinds.go.
 //
 // Each language extractor detects the enum shape (Python Enum subclass, TS
 // `enum` / string-literal union, Java enum, Go iota const block, Ruby
@@ -67,8 +72,14 @@ type constMember struct {
 //	scope:enum:<sourceFile>:<EnumName>
 //
 // File-scoped so distinct same-named enums in different files stay distinct.
-// This value is stored as the entity's QualifiedName, so a TYPED_AS edge whose
-// ToID equals it binds via the resolver's byQualifiedName exact-match tier.
+// This value is stored as the entity's QualifiedName (see EnumEntity below), so
+// an edge whose ToID equals it binds via the resolver's byQualifiedName
+// exact-match tier — the first probe in resolve.Index.LookupStatusHint, ahead
+// of the structural, kind and bare-name tiers. The one producer that uses it is
+// C#'s field→declared-type pass (csInFileTypeTargets in
+// internal/extractors/csharp/field_type_refs.go), which reads the SCOPE.Enum
+// record's QualifiedName directly and emits REFERENCES with
+// ref_kind="field_target_type".
 func EnumQualifiedName(sourceFile, enumName string) string {
 	return "scope:enum:" + sourceFile + ":" + enumName
 }
--- a/internal/extractors/csharp/field_type_refs.go
+++ b/internal/extractors/csharp/field_type_refs.go
@@ -19,9 +19,10 @@ import (
 // five languages through one shared helper (internal/custom/{python,golang,
 // java,javascript,ruby}, `referencesClassEdge`): Kind REFERENCES with the
 // property `ref_kind: "field_target_type"`. Core adopts that spelling rather
-// than minting a third name — `TYPED_AS` and `HAS_TYPE` are both declared in
-// internal/types/kinds.go with zero producers (#6906, #5828), and a third
-// spelling would split every "which fields point at X?" query.
+// than minting a third name — `TYPED_AS` and `HAS_TYPE` were both declared in
+// internal/types/kinds.go with zero producers (#6906, #5828) and were DELETED
+// by #6906 once this pass shipped; a third spelling would have split every
+// "which fields point at X?" query.
 //
 // THE ADDRESS IS NOT the custom lane's `Class:<Target>`, and that divergence is
 // measured, not stylistic. Both `Customer` and `Class:<Customer>` reach the
--- a/internal/types/kinds.go
+++ b/internal/types/kinds.go
@@ -1236,10 +1236,6 @@ const (
 	// citizens via this edge kind.
 	RelationshipKindUnresolvedFetch RelationshipKind = "UNRESOLVED_FETCH"
 
-	// #1343: TypeScript type extraction edges.
-	//   HAS_TYPE     : entity → SCOPE.Schema  (e.g. a variable or field whose declared type is the target schema)
-	RelationshipKindHasType RelationshipKind = "HAS_TYPE"
-
 	// #1344: Post-rebuild rename / move / split detection.
 	// Emitted from a NEW entity to the OLD entity's ID after the indexer
 	// detects that a function, method, or class was renamed between rebuilds.
@@ -1794,18 +1790,6 @@ const (
 	// internal/engine/ws_channel_grouping.go.
 	RelationshipKindJoinsChannel RelationshipKind = "JOINS_CHANNEL"
 	RelationshipKindBroadcastsTo RelationshipKind = "BROADCASTS_TO"
-	// #3628 data-model: field/param → enum value-set edge. Emitted from a
-	// model field, struct field, or parameter whose declared type is a
-	// SCOPE.Enum value-set node, to that node. Lets the graph answer "which
-	// fields are constrained to enum X's values?" — the inbound side of the
-	// enum-parity contract. ToID is the SCOPE.Enum entity's QualifiedName so
-	// the resolver binds it via the byQualifiedName exact-match tier. Properties:
-	//   "enum"  : the bare enum type name.
-	//   "field" : the field/param name carrying the type (when known).
-	// Honest-partial: only statically-resolvable enum-typed declarations emit
-	// this edge; dynamic / computed types do not. See
-	// internal/extractor/enum_valueset.go.
-	RelationshipKindTypedAs RelationshipKind = "TYPED_AS"
 	// [sbom] DEPENDS_ON_PACKAGE points a manifest's project-anchor entity at a
 	// synthetic SCOPE.Package node (Name "package:<ecosystem>:<name>") it
 	// declares as an external dependency. Distinct from the file-scoped
@@ -1908,6 +1892,18 @@ const (
 	RelationshipKindDeploys RelationshipKind = "DEPLOYS"
 )
 
+// RETIRED KINDS — do not re-add. TYPED_AS (#6906) and HAS_TYPE (#5828) were
+// declared and registered here for months with zero producers and zero
+// consumers. Both named the same concept: a field / parameter → its declared
+// type. #6984 shipped that edge for C# as RelationshipKindReferences carrying
+// the property ref_kind="field_target_type" (internal/extractors/csharp/
+// field_type_refs.go), which is the spelling the custom lane had already
+// shipped in five languages via referencesClassEdge. A third name for one
+// concept is the defect #6451 was merged to remove for SCOPE.ExternalAPI, so
+// both constants were deleted rather than kept as reserved words. The deletion
+// is pinned by TestRetiredRelationshipKindsStayRetired in
+// kinds_retired_6906_test.go.
+//
 // AllRelationshipKinds returns every RelationshipKind producers may emit.
 func AllRelationshipKinds() []RelationshipKind {
 	return []RelationshipKind{
@@ -1980,8 +1976,6 @@ func AllRelationshipKinds() []RelationshipKind {
 		RelationshipKindCloudEventFlows,
 		// #1217:
 		RelationshipKindUnresolvedFetch,
-		// #1343:
-		RelationshipKindHasType,
 		// #1344 rename detection:
 		RelationshipKindRenamedFrom,
 		// #1374 Django signal/admin connectivity:
@@ -2060,8 +2054,6 @@ func AllRelationshipKinds() []RelationshipKind {
 		// [realtime] WS room/channel grouping: JOINS_CHANNEL / BROADCASTS_TO.
 		RelationshipKindJoinsChannel,
 		RelationshipKindBroadcastsTo,
-		// #3628 data-model: field/param → enum value-set edge.
-		RelationshipKindTypedAs,
 		// #3834 (epic #3829) member-granularity inheritance edge:
 		RelationshipKindInherits,
 		// #4306 deterministic markdown doc ingestion (opt-in):
--- a/internal/types/kinds_retired_6906_test.go
+++ b/internal/types/kinds_retired_6906_test.go
@@ -1,6 +1,10 @@
 package types
 
-import "testing"
+import (
+	"os"
+	"strings"
+	"testing"
+)
 
 // Issue #6906 — TYPED_AS and HAS_TYPE are RETIRED. This guard fails if either
 // name comes back.
@@ -67,6 +71,57 @@ func TestRetiredRelationshipKindsStayRetired(t *testing.T) {
 		if IsValidRelationshipKind(retired) {
 			t.Errorf("IsValidRelationshipKind(%q) = true; the retired kind is back in the "+
 				"vocabulary (#6906/#5828)", retired)
+		}
+	}
+}
+
+// coverageJSPath is the public coverage page's hand-maintained kind roster.
+// Relative to this package directory, which is where `go test` runs.
+const coverageJSPath = "../../site/src/scripts/coverage.js"
+
+// TestRetiredRelationshipKindsAreGoneFromTheCoveragePage extends the retirement
+// to the one surface a Go-only enumeration cannot see.
+//
+// The first version of this PR grepped the UPPERCASE literals and the Go
+// constant names, and concluded "no dashboard or webui reference". It was
+// wrong: site/src/scripts/coverage.js listed "has_type" and "typed_as" in its
+// LOWERCASE display form, so the public page would have advertised two kinds
+// that no longer exist. A name has more than one spelling — constant,
+// string literal, lowercase display form — and an enumeration that covers one
+// spelling has not covered the name.
+//
+// This is deliberately a plain substring scan rather than a JS parse: the
+// spelling to catch IS the raw token, and a parser would add a way to be
+// silently wrong about a file that has no Go meaning.
+func TestRetiredRelationshipKindsAreGoneFromTheCoveragePage(t *testing.T) {
+	raw, err := os.ReadFile(coverageJSPath)
+	if err != nil {
+		t.Fatalf("read %s: %v (the coverage page is checked in; if it moved, "+
+			"repoint this guard rather than deleting it)", coverageJSPath, err)
+	}
+	page := string(raw)
+
+	// Non-vacuity: the scan must be looking at a file that really carries the
+	// lowercase roster, in the spelling the retired names would have used.
+	// Without this a moved, emptied or restructured file reads as "absent".
+	for _, control := range []string{`"references"`, `"inherits"`, `"contains"`} {
+		if !strings.Contains(page, control) {
+			t.Fatalf("control %s is absent from %s (%d bytes read); this scan is not looking at "+
+				"the relationship-kind roster and its absence checks are vacuous",
+				control, coverageJSPath, len(page))
+		}
+	}
+
+	for _, retired := range retired6906Kinds {
+		// The display form: lowercased, quoted as a roster item. The prose
+		// comment in that file names the retired kinds in UPPERCASE to explain
+		// why they are missing, which is why this matches the quoted lowercase
+		// token and not the bare name.
+		item := `"` + strings.ToLower(retired) + `"`
+		if strings.Contains(page, item) {
+			t.Errorf("%s still lists %s; %s was retired by #6906/#5828 and the public coverage "+
+				"page would advertise a kind grafel cannot emit",
+				coverageJSPath, item, retired)
 		}
 	}
 }
--- a/internal/types/kinds_retired_6906_test.go
+++ b/internal/types/kinds_retired_6906_test.go
@@ -1,0 +1,72 @@
+package types
+
+import "testing"
+
+// Issue #6906 — TYPED_AS and HAS_TYPE are RETIRED. This guard fails if either
+// name comes back.
+//
+// Both were declared in kinds.go and registered in AllRelationshipKinds() with
+// zero producers and zero consumers, and both named one concept: a field or
+// parameter → its declared type. #6984 shipped that edge for C# as
+// RelationshipKindReferences with the property ref_kind="field_target_type",
+// adopting the spelling the custom lane had already shipped in five languages.
+// Re-declaring either constant would put a third name on one concept.
+//
+// WHY THE LITERALS ARE WRITTEN OUT HERE. Every assertion below compares against
+// an INDEPENDENT string literal, never against AllRelationshipKinds() or any
+// other production value (#6975): a guard that ranges over the production slice
+// and asks whether it contains what the production slice contains cannot
+// observe a re-addition. "TYPED_AS" and "HAS_TYPE" appear in this file as
+// hard-coded strings for that reason, and must not be replaced by a reference
+// to a constant.
+var retired6906Kinds = []string{"TYPED_AS", "HAS_TYPE"}
+
+// TestRetiredRelationshipKindsStayRetired asserts the retirement on all three
+// surfaces a re-addition could use: the const declarations in kinds.go (read
+// from source, so a constant declared under ANY name is caught by its value),
+// the registered vocabulary, and the IsValidRelationshipKind predicate.
+func TestRetiredRelationshipKindsStayRetired(t *testing.T) {
+	declared := declaredRelationshipKindsFromSource(t)
+
+	// Non-vacuity, ahead of every absence check. A parse that read the wrong
+	// file, or read the right file and matched nothing, would let all three
+	// absence assertions below pass by examining nothing. Two independent
+	// literals must be PRESENT before any absence is believed.
+	if len(declared) == 0 {
+		t.Fatalf("no RelationshipKind constants parsed from %s; every absence check below "+
+			"would be vacuous", kindsSourceFile)
+	}
+	byValue := map[string]string{}
+	for _, d := range declared {
+		byValue[d.Value] = d.Name
+	}
+	for _, present := range []string{"CALLS", "REFERENCES"} {
+		if _, ok := byValue[present]; !ok {
+			t.Fatalf("control %q is not among the %d constants parsed from %s; the extraction is "+
+				"not reading the declarations and the absence checks are vacuous",
+				present, len(declared), kindsSourceFile)
+		}
+		if !IsValidRelationshipKind(present) {
+			t.Fatalf("control IsValidRelationshipKind(%q) = false; the predicate under test is "+
+				"not answering, so a false on a retired kind proves nothing", present)
+		}
+	}
+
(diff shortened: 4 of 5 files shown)


cajasmota:
## Independent adversarial review — **REQUEST-CHANGES** (two one-line prose fixes; the code is sound)

Measured in my own worktree at `/private/tmp/claude-501/-Users-jorgecajas-Projects-archigraph/c0fefed4-0195-4c81-9a94-07a7145710a4/scratchpad/review-6992-adv7/wt` (PR head `f545442`) and a parent checkout at `.../review-6992-adv7/parent` (`3c7d2c5ec`). macOS/APFS, isolated `GRAFEL_HOME`/`GRAFEL_DAEMON_ROOT`.

### (a) Can deleting an enum member renumber persisted kinds? **No — kinds persist by NAME, demonstrated at runtime, not inferred.**

The premise that these kinds are "a serialised enum" is false. `internal/graph/schema/graph.fbs` declares **zero** `enum` types (`grep -c '^enum ' → 0`); both `Entity` and `Relationship` carry `kind: string`. `fbwriter`'s `enumKindSet` is an in-memory `map[string]struct{}` built by `sync.OnceValue` from `AllRelationshipKinds()`, used only as the #6757 arm-C write-path counter — it is never serialised, so it has no ordinal at all.

Proven end-to-end rather than argued. In the **parent** tree I wrote a `graph.fb` via `fbwriter.WriteAtomic` containing **112** relationships, one per parent `AllRelationshipKinds()` entry — including `HAS_TYPE` at index 54 and `TYPED_AS` at index 100 — then read that same file with the **PR** tree's `fbreader.Open` + `IterateRelationships`:

```
diff parent-written-kinds.txt pr-read-kinds.txt   →  no differences, exit 0
pr read 112 rels;  54  HAS_TYPE   100  TYPED_AS   (read back verbatim by the PR binary)
```

An old graph reads correctly: every kind string, at every position, survives byte-for-byte, including the two deleted ones. No shift, no misread. My mutant **N4** (swap `JOINS_CHANNEL`/`BROADCASTS_TO` in `AllRelationshipKinds()`) is correspondingly **equivalent** — nothing in the suite observes slice order, which is the right answer given the above.

I also checked the adjacent on-disk mechanism: `types.KindVocabularyVersion` (#6779). Its bump rule is explicitly scoped to **entity** kinds ("Bump this whenever an entity kind is RENAMED or RETIRED"), and its guard `kind_vocabulary_bump_guard_6779_test.go` enumerates entity kinds only. Correctly **not applicable** here, and doubly so: with zero producers no stored graph can contain either edge. Worth stating in the PR body, since a reviewer will ask.

### (b) Is the "no non-test caller" enumeration complete? **The Go half yes — and it is machine-checked, not grep-checked. The non-Go half is NOT: one live surface was missed.**

Better than grep: my mutant **N1** (delete `RelationshipKindSupervises,` from `AllRelationshipKinds()`, keep the constant — 1 occurrence, inside `AllRelationshipKinds`) is **DEAD**, failing both `TestAllRelationshipKinds_CoversEveryDeclaredConstant` **and `TestProducerRelationshipKindLiterals_AreValid`**. That second one is the real evidence for this PR: `internal/relkinds` scans producer source for emitted kind literals and validates them against the vocabulary. If either retired kind had a producer anywhere, deleting it would have gone red. "Zero producers" is therefore pinned by a test, not by the author's grep — say so in the PR body, it is the strongest argument you have.

**MISSED SURFACE — required fix 1.** The author's whole-tree grep searched the **uppercase** literals and the Go constant names. The public site uses the lowercase display form, and a case-insensitive grep finds a **fifth** file:

```
site/src/scripts/coverage.js:26
  { label: "Structural", items: [..., "references","has_props","has_type","typed_as","inherits"] },
```

with the file comment at `:11-13` calling these lists *"Full authoritative kind lists, sourced from internal/types/kinds.go … every kind grafel extractors are permitted to emit"*. After this PR the public coverage page advertises two kinds that do not exist — the exact false-advertisement #6906 was filed to end — and it ships on this PR's own Cloudflare Pages deploy. The PR body's sentence **"no rule YAML, no dashboard, MCP or webui reference"** is therefore false as written. Delete `"has_type","typed_as"` from that array (the stale `(106)` count in the same comment is pre-existing — parent is 112, PR is 110 — worth correcting while you are in there).

Nothing else: `--binary-files=text` case-insensitive grep over the whole tree finds only the 4 Go files plus that one. No golden fixture, no `.json`/`.yaml`, no binary artefact, no reflection or string-keyed kind lookup.

### (c) Is the guard non-vacuous on all three surfaces? **Yes — verified by breaking the controls, not by reading them.**

- **Positive control fires on the "reads the wrong file" path** (the one a count floor cannot catch): pointing `kindsSourceFile` at `derived_kinds.go` — which parses fine and yields a **non-empty** set of 1 constant — fails with `control "CALLS" is not among the 1 constants parsed from derived_kinds.go; … the absence checks are vacuous`. Exactly right.
- **Predicate surface graded independently** — my mutant **N3**: `IsValidRelationshipKind` returns true for `"TYPED_AS"`/`"HAS_TYPE"` before the scan (1 occurrence, inside `IsValidRelationshipKind`; `gofmt`/`go vet` clean). **DEAD** on both kinds, and — the point — the AST and registration assertions did **not** fire, so that surface carries its own weight.
- Independent literals per #6975: confirmed, `retired6906Kinds` never touches a production value.
- **N2** (asymmetry: re-add `HAS_TYPE` const + registration *and* shrink the guard literal to `{"TYPED_AS"}`) is **ALIVE**, green across `internal/types`. Not a defect — it edits the pin itself, and any guard dies that way — but it names the honest limit: the retirement rests on exactly one string list, and nothing cross-checks it against the non-Go surfaces. Cheap hardening if you want it: have the same test grep `site/src/scripts/coverage.js` for the lowercase forms. That is the assertion that would have caught (b).

### (d) Are the rewritten comments factually correct? **Five of six claims yes. One pre-existing comment is falsified by this diff — required fix 2.**

Verified against source: `Kind: string(types.RelationshipKindReferences)` + `{K: "ref_kind", V: csFieldTargetRefKind}` with `csFieldTargetRefKind = "field_target_type"`; `csInFileTypeTargets`'s `case r.Kind == "SCOPE.Enum": put(r.Name, r.QualifiedName)`; `"scope:enum` appears in exactly two non-test files (`enum_valueset.go` constructor, `field_type_refs.go` comment), so C#-only holds; the `idx.byQualifiedName[stub]` probe does sit ahead of the structural/kind/name tiers in `LookupStatusHint`. Naming functions instead of line numbers is applied consistently and is the right call.

**Required fix 2** — `internal/extractors/csharp/field_type_refs.go:22-24`, left untouched and described in the PR body as *"accurate"*:

> `TYPED_AS` and `HAS_TYPE` **are both declared in** internal/types/kinds.go with zero producers (#6906, #5828)

Present tense, and false the moment this merges. A PR whose entire thesis is "a comment described a binding nobody built" must not leave behind a comment describing constants nobody declares. One-line fix: *"were declared … and were deleted by #6906"*.

**One nit, not blocking.** *"in every other language a `SCOPE.Enum` node still has no inbound edge"* is an unqualified absolute over every language, and nothing observes it — `EnumEntity` has callers in python/golang/dart/rust/… and the claim covers edges of **any** kind, not just field→type. I found no golden fixture asserting an edge into `SCOPE.Enum` (`"to_kind": "SCOPE.Enum"` → 0 hits), so it is probably true, but it is the one sentence in the new comments with no evidence behind it. Suggest narrowing to "no inbound **field→declared-type** edge", which is what you actually verified.

### Mutants scored (M1–M3 not re-run)

| # | Mutant | Edit landed | Result |
|---|---|---|---|
| N1 | unregister a **produced** kind (`RelationshipKindSupervises,` removed from `AllRelationshipKinds`, constant kept) | 1 occurrence → 0, inside `AllRelationshipKinds` | **DEAD** — `TestAllRelationshipKinds_CoversEveryDeclaredConstant` **and** `TestProducerRelationshipKindLiterals_AreValid` |
| N2 | asymmetric: re-add `HAS_TYPE` const+registration **and** shrink guard literal to `{"TYPED_AS"}` | 2 occurrences in `kinds.go` (`:1239` const, `:1981` registration) + 1 in the guard | **ALIVE**, expected — it edits the pin; documents the single-literal limit (see (c)) |
| N3 | permissive predicate: `IsValidRelationshipKind` returns true for both retired names | 1 occurrence, inside `IsValidRelationshipKind` | **DEAD** — predicate surface only; AST + registration correctly silent |
| N4 | reorder `AllRelationshipKinds()` (`JOINS_CHANNEL`↔`BROADCASTS_TO`) | 1 occurrence, inside `AllRelationshipKinds` | **equivalent** under `internal/types` + `fbwriter` + `relkinds` — and correctly so: nothing persists an ordinal (see (a)) |

`gofmt -l` and `go vet` clean before each score, green `internal/types` baseline immediately before, `-count=1`, restore by `cp` from `shasum`-verified backups (`kinds.go` `10bdf67744…` — matching the author's — `kinds_retired_6906_test.go` `3083561b42…`, `kinds_enum_completeness_6757_test.go` `16d3f2c1e1…`), never `git checkout`. Tree confirmed clean after every restore.

### Suites re-run independently (all exit 0, `-count=1`)

`./internal/types/... ./internal/extractor/ ./internal/extractors/csharp/ ./internal/relkinds/ ./internal/graph/... ./cmd/grafel/` — **0 FAIL lines**, `cmd/grafel` `ok 145.7s` (whole-graph digest pin held). `scripts/quality/run.sh --ratchet` → *"38 gated fixtures held their baseline (29 at 100% must-have recall)"*, exit 0. The author's verification reproduces exactly.

### Verdict

The deletion itself is correct and safe: no on-disk hazard, no runtime reachability, no producer, guards genuinely non-vacuous on all three surfaces, ratchet and digest pin held. **REQUEST-CHANGES only for the two prose falsehoods** — `site/src/scripts/coverage.js` (user-facing, and contradicts the PR body's own "no webui reference"), and the now-false present tense in `csharp/field_type_refs.go:22`. Both are one-liners; happy to re-approve on sight.

cajasmota:
## Coordinator mutant CJ-1 on `a877c6b42` — ALIVE, not production-reachable, cheap to grade

Own detached worktree, `git status --porcelain` empty at start and after restore, restored by `cp` from a SHA-256-verified backup (`1135c08a…` both ways), zero `MUTANT` text remaining, green baseline immediately before (9 packages ok). macOS/APFS.

**CJ-1 — the deletion's one RUNTIME consequence, on the write path.** Nine mutants have been scored here and every one lands on a *declaration* surface: the constant, the registration, the predicate, the coverage page, the enum order. The reviewer's read-back experiment covers the **read** path. Nobody touched what changes on the **write** path.

`nonEnumKindTally.observe` (`internal/graph/fbwriter/non_enum_kinds.go:278`) skips a kind that is in the enum and tallies one that is not. **Deleting these two kinds moves them across that line**: before this PR a `TYPED_AS` edge was `inEnum` and silently skipped; after it, the same edge falls through and is counted as an **unknown kind**. CJ-1 restores the old behaviour for exactly those two names, leaving everything else alone:

```go
+  if kind == "TYPED_AS" || kind == "HAS_TYPE" { // MUTANT CJ-1: retired kinds still treated as in-enum
+      return
+  }
   if _, inEnum := enumKindSet()[kind]; inEnum {
       return
   }
```

Edit proven landed: exactly 1 occurrence, hit set asserted `[278]`, **enclosing function asserted** to be `observe(kind string)`, `go vet` clean before scoring.

**Result: ALIVE** — `./internal/graph/...` and `./internal/types/...` both green, exit 0.

## The verdict, stated narrowly

**Not production-reachable, and the PR is the reason why.** The distinguishing input is a relationship whose kind string is literally `TYPED_AS` or `HAS_TYPE`. Nothing can produce one: `TestProducerRelationshipKindLiterals_AreValid` machine-checks producer literals against the vocabulary, and the retirement guard's raw-literal surface (M3) blocks re-introduction without a constant. So this is not a defect and it does not block.

**But it is the deletion's only observable runtime behaviour, and it is undocumented.** The change is not merely "two identifiers vanish" — it silently reclassifies a kind from *skipped* to *unknown-tallied*. That is arguably the **desirable** behaviour: if a retired kind ever reappears in a written graph, surfacing it in the unknown-kind tally is exactly what you would want, and it is a second line of defence behind the source-scan guard. Right now nothing says so, and nothing would notice if a later refactor of `observe` took it away.

**Priced: about ten lines** — a unit test on `observe` asserting a `TYPED_AS` edge lands in `t.counts` rather than being skipped, with an in-enum control that still is. That is cheap enough that "too expensive" is not available, and it converts an accident into a documented property. **I am not blocking on it**; your call whether to take it here or let me file it. I would take it, because it is the moment the intent is clearest and because the tally comment already explains its own design carefully enough that this belongs beside it.

## On the round

Both required fixes are right, and the wider fix on `coverage.js` is better than what I asked for: deleting the stale `(60)`/`(106)` counts rather than restating them, having found the accessors actually return 94 and 110 while the arrays hold 59 and 106. **An unchecked number is how that drift went unnoticed**, and the same reasoning is why the fix was to remove the claim rather than to correct it.

`N6` — repointing `coverageJSPath` at a file that exists and parses fine — is the right control, because it is the "reads the wrong file" no-op that a count floor passes. Two independent guards now each fail for a different reason.

Declining the reviewer's suggested phrasing on the enum comment, on the grounds that a `Class:<X>` stub can reach the bare-name tier and a `SCOPE.Enum` node carries the enum's `Name`, is the right call: substituting one unverified absolute for another would have missed the point of the nit.

Issue #6990, opened 2026-09-07
django: the 400-byte forward scan reads a FOLLOWING declaration's target — LIVE on django (LogEntry.user -> ContentType, binds to a real model); general population unmeasured, my zero-incidence claim was wrong


cajasmota:
Successor to #6988, which shipped in #6989. Both residues were **reproduced on the fix commit `58341a61b`** by the independent reviewer, and both are **zero-incidence on our corpus today**, so this is a correctness-of-shape issue rather than a live loss.

#6988 fixed the case where `GenericForeignKey("content_type", "object_id")`'s first argument — the name of a sibling field — became a target model. The fix is a gate on `isDjangoRelationalField`. These are the two ways the same bad target still gets produced.

## Residue 1 — a GFK subclass

```python
class CTForeignKey(GenericForeignKey):
    ...
```

`Tag.content_object` → `Class:content_type`. The gate rejects the exact name `GenericForeignKey` by final dotted segment; a subclass has a different name and passes.

This is the mirror image of the trade #6988 settled: the deny-list is open-ended by construction, exactly as the `HasSuffix(rhs, "ForeignKey")` admit was.

## Residue 2 — the 400-character forward scan reads text belonging to a REJECTED constructor

`internal/custom/python/django.go:702`:

```go
fullRHS := body[fIdx[0]:min(fIdx[0]+400, len(body))]
```

The window starts at a matched field and runs forward 400 bytes **regardless of what is in it**. So a legitimate

```python
user = models.ForeignKey(settings.AUTH_USER_MODEL, ...)
```

sitting above a `GenericForeignKey("content_type", "object_id")` reads the GFK's `ct_field` string out of the *following* declaration: `Note.user → Class:content_type`, still present on the fix commit.

**This corrects a natural reading of #6989** and is the more interesting of the two: the gate decides *which constructors are processed*, but the regex still reads text belonging to constructors the gate rejected. "The gate is the sole control over which constructors reach target extraction" would be prose that no test observes.

## Measured

django corpus, gate **ON** (#6966 — the custom lane is off by default, so **user-facing incidence today is zero**): snake_case `field_target_type` targets **18 on the parent → 0 on the fix commit**, with 956 total edges on both sides. Both residues need a shape our corpus does not currently contain.

**A zero here is corpus-relative and is not "cannot occur"** — residue 2 in particular needs only a GFK declared within 400 bytes after a real relational field, which is an ordinary way to write a Django model.

## Also open, from the same review

[...]

PR #6994, opened 2026-09-08
fix(#6986): the Python custom lane's field->model edge was addressed by a name that could not bind — 863 edges on django, 39 bound, and re-addressing it silently stripped django_rel from 760 more

cajasmota:
## Gate state, first

**Everything below is gate-ON** (`WithCustomExtractors(true)`, the in-process entry point `grafel quality` uses). The custom lane is **OFF by default** (#6966), so **user-facing incidence on a default `grafel index` today is ZERO** — none of these edges are produced at all. That caps the urgency; it does not make the address correct.

Measured in `/private/tmp/.../scratchpad/impl-6986a-k7m3q/wt` at `64ffa3d22`, on the `django` corpus repo copied to scratch (`/Users/jorgecajas/Projects/archigraph-corpora` untouched). macOS/APFS, Apple Silicon, `GOMAXPROCS=3`, `nice -n 10`, isolated `GRAFEL_HOME` / `GRAFEL_DAEMON_ROOT`.

## Part 1 — the three questions the arm was gated on

**1. Can the Python producer build a structural address? YES, and more cheaply than the C# arm.** Every `referencesClassEdge` call site has `file.Path`, and it does not need the target's file: the Python dialect addresses by the **consumer's** file. `internal/resolve/refs.go` `lookupStructural` resolves `scope:component:class:python:<file>:<Name>` in two tiers — same-file `lookupLocationKind`, then, for `component` scope with `lang=="python"` **only**, cross-file `lookupUniqueRealComponentByName`. The Python hierarchy extractor has emitted EXTENDS targets in exactly this shape with exactly this convention since the flask-realworld wave, and the Django class anchor already carries `ref: scope:component:class:python:tests/admin_views/models.py:ReferencedByParent` as a property.

**2. Same-file is not the constraint here.** Because of that cross-file tier, this pass is **not** same-file-only the way #6984 had to be. Measured: `dup` (≥2 declaring files) 439 newly bound, `uniq-same` 305, `uniq-other` 3.

**3. The untwinned-`Model` subset binds — measured, not assumed.** 724 of 747 newly-bound edges bind to the untwinned rule-pack `Model` node (the resolver's real-tier-first preference), 20 to `SCOPE.Component`; same file, same line, same qualified name. `ParentWithFK.fk → ReferencedByParent` and `B3.restrict → R` both bind.

## The result

| | before | after |
|---|---|---|
| `field_target_type` edges | 863 | 949 |
| bound | **39 (4.5%)** | **886 (93.4%)** |
| bound `(field, file, target)` keys | 39 | 802 |
| **keys that LOST a binding** | — | **0** |
| **bound to a file that does NOT declare the name** | — | **0** |

The +86 edges are edges whose *FROM* endpoint is itself an unresolved stub (`Person.friends` exists in four files): they previously collapsed into one because their `Class:` ToIDs were byte-identical, and now stay distinct. Key set is 863 before and after; `onlyBefore` = `onlyAfter` = 0.

8 edges were **upgraded**, not substituted: `Request.profile`, `SessionEvent.session`, `ColorfulItem.colors`, `SimulationRun.start`/`.end`, `Thing.color`, `Product.image`, `FooImage.my_image` moved from a synthesised `SCOPE.External` node with no source file to the real in-tree model.

## The part that would have shipped as a silent substitution

`internal/extractors/python/django_relational.go:246` already emits a REFERENCES edge **from the same field entity to the same target**, addressed `scope:component:ref:python:<file>:<Name>` and stamped `django_rel` / `self_ref` / `django_fk_string`. On django, **766 field entities carry both edges; the core edge binds 708 times while the custom edge binds 8.**

While the custom ToID never bound, the two edges could not collide. Once it binds, graph assembly **dedupes the pair and keeps exactly one property set** — and it keeps the custom one. Measured on a full re-index of the naive change:

```
REFERENCES edges carrying django_rel:   1018  ->  258      (-760)
```

`GRAPH_RELATES` (922) was unaffected, so the model↔model cardinality survived — but the **field-level** relation type and self-reference flag vanished from 760 edges, invisibly to any recall count. That is the #6973 class.

So the producer now stamps those three properties itself, mirroring `parseTargetExpr`'s semantics exactly. Re-measured on a third full index:

```
REFERENCES edges carrying django_rel:   1018  ->  1207
django_rel (from,to) PAIR SET:  before 1018, after 1207, LOST 0
REFERENCES self_ref:            1018  ->  1207
django_fk_string:                169  ->   210
```

**The count does not come back to exactly 1018, and it should not.** 1018 counted only the edges the *core* producer emitted. The custom lane emits a relational-field edge for 97 django fields the core pass produces nothing for, plus the 86 previously-collapsed ones; those now carry `django_rel` too. The load-bearing control is the **set** control — every `(from, to)` pair that carried `django_rel` before still carries it, `LOST 0` — which a count could not have shown. Named example, asserted in the suite: `Order.watchers = models.ManyToManyField('self')` carries `django_rel=ManyToManyField`, `self_ref=true`, `django_fk_string=self`; `Order.buyer = models.ForeignKey(Customer)` carries `django_rel=ForeignKey`, `self_ref=false`, no `django_fk_string`.

**Vocabulary correction.** An earlier version of the `djangoRelCtorLeaf` comment said `django_rel` "carries the same vocabulary as the core extractor's `djangoFieldTypes` key". It does not, and the difference is a **gain**: core gates on `djangoRelationalFieldTypes`, an exact three-name allow-list, while this lane gates on `isDjangoRelationalField`, a `*ForeignKey`/`*OneToOneField`/`*ManyToManyField` **suffix** rule. `TreeForeignKey` reaches this lane and is stamped `django_rel: TreeForeignKey`, a value core never emits — because core emits no edge for that field at all. That direction is what keeps the pair-set claim safe: this lane can only ADD a `django_rel`-carrying edge where core had none, never rename one core already produced. The comment now says that, and `TestIssue6988_LegitimateRelationalFieldsSurvive`'s `Book.category` row pins it.

**Semantic agreement, verified per form** (reviewer's check, stronger than the corpus set-diff): core `parseTargetExpr` and custom `djangoRelTargetDetail` produce identical `django_rel` / `self_ref` / `django_fk_string` on `ForeignKey('auth.User')`, `ForeignKey('Customer')` and `OneToOneField(Customer)`. **The app-labelled string form `'auth.User'` is not covered by my suite and it agrees** — that is the case that would have silently broken `ResolveDjangoStringFKRefs`, and its agreement is what makes the `LOST 0` pair-set result safe rather than lucky.

Full relationship-profile diff, before → after, is exactly the intended edges and nothing else:

```
+847  REFERENCES INFERRED_FROM_MODEL_FIELD_TARGET  bound
-761  REFERENCES INFERRED_FROM_MODEL_FIELD_TARGET  dangling
-708  REFERENCES (core django_rel)                 bound      (absorbed by the dedupe)
 -52  REFERENCES (core django_rel)                 dangling   (absorbed)
  -1  SCOPE.External/function entity + its CONTAINS (an edge that now binds in-tree)
```
**On the +86 — legitimate on the TO side, with a caveat that belongs in #6726's arithmetic.** The reviewer named the population in `archigraph-corpora/django`: `class Person(models.Model)` with `friends = models.ManyToManyField("self")` is declared in **four distinct apps** (`tests/test_runner/models.py:4`, `tests/m2m_recursive/models.py:22`, `tests/queries/models.py:690`, `tests/m2m_signals/models.py:24`). Four declarations, four relationships — four edges is right, not inflation. **But the FromID is the bare `Person.friends`, identical across all four files and still unresolved**, so the graph gains four outbound edges hanging off **one** collapsed source node. That is pre-existing FROM-side under-resolution, not something this PR introduces, and it means "+86 distinct edges" is only half true: **86 distinct targets, from fewer distinct sources.**

They are not new references. A field whose own entity is an unresolved stub (`Person.friends` exists in four files) used to collapse with its namesakes because their `Class:<Target>` ToIDs were byte-identical; a file-qualified ToID no longer collapses. The producer emits the **same number of records** — one per relational field. The `referencesClassEdge` doc comment originally read "never more edges than before", which is true of records and false of the graph; it now states the measured version and names the test behind "never a guessed one".

(`RENAMED_FROM` also moves between runs — a git-history pass sensitive to the state dir, not to this change; the same noise appears on a repo where no `field_target_type` edge changed at all.)

## Both directions are asserted, and the negative is by name

- `TestPythonFieldTargetType_BindsWhenTheNameIsDeclaredInTwoFiles` — `Customer` declared in `shop/models.py` AND `billing/models.py`. A **positive control** first asserts the bare `Class:Customer` form is ambiguous in the fixture, or the structural address would be ungraded. Then `Order.buyer` must bind, and bind to **shop's** Customer.
- `TestPythonFieldTargetType_NeverBindsToANonDeclaringFile` — the over-fire direction, named: no bound `field_target_type` edge may point at an entity in a file that does not declare that name, and specifically `Order.buyer` may not reach `billing/models.py`'s `Customer`. Guarded against vacuity (`checked == 0` fails).
- `TestPythonFieldTargetType_SelfRefIsTheStringFormOnly` — `self_ref` is true for the literal `'self'` and **false** for a symbol target naming the enclosing class, which is `parseTargetExpr`'s rule. A permissive reading flips `Node.parent`.
- **`TestPythonFieldTargetType_CrossFileTier_BindsUniqueRefusesAmbiguous` (new)** — the cross-file tier is the entire difference between this arm and #6984/#6991, and nothing in this package graded it: `NeverBindsToANonDeclaringFile`'s consumer declares `Customer` itself, so the same-file tier resolves first and the cross-file tier never fires. The new fixture's consumer (`orders/models.py`) declares **neither** target, and all three outcomes are asserted together — ambiguous (`Customer` in two other files) must **DANGLE**, unique (`Vendor` in one) must **BIND to that file**, absent (`Ghost`) must **DANGLE** — with fixture preconditions checked first so the ambiguity cannot silently disappear, and per-edge existence checked so no branch passes vacuously.

**A vacuity bug in my own first draft, found while writing that test and fixed here.** `resolve.ReferencesEmbedded` rewrites `FromID` in place once the source endpoint binds, so every post-resolution `e.Rel.FromID == "Order.buyer"` match in the original file matched **nothing** and the assertions under it passed vacuously. All matching now goes through `edgeKey6986`, built from the owning entity's `Name` plus the `field_name` property — both untouched by resolution — and each such loop carries an explicit `checked == 0` failure. This affected the binding half of `BindsWhenTheNameIsDeclaredInTwoFiles` and the `Order.buyer`-specific clause of `NeverBindsToANonDeclaringFile`.

## The 5 updated tests — what each was pinning, and why the new assertion pins the same thing

None was weakened; each asserted `pyClassRef(X)`, i.e. **the old address literal**, as the way of saying "this edge exists and points at X".

| test | what it pinned | new assertion |
|---|---|---|
| `TestIssue6988_NamedGFKFieldEmitsNoTargetEdge` | its forbidden row is `len(edges) != 0` and is **untouched**; only the *positive control* (`GenericB1.content_type` still emits a real edge to `ContentType`) named the address | same edge, addressed via `fieldTargetRef6988("ContentType")` |
| `TestIssue6988_WrongButBoundEdgeShapeIsGone` | the wrong-but-bound shape `Book.created_by → created_by_ct` must NOT be emitted, and the sibling `Book.created_by_ct → ContentType` must survive | both, re-addressed. The forbidden shape is still asserted as *this exact ToID absent* **and** as *zero `field_target_type` edges from that field* |
| `TestIssue6988_LegitimateRelationalFieldsSurvive` | all 8 legitimate forms (FK/O2O/M2M × string/`'self'`/symbol/`to=`/third-party subclass) still emit an edge naming the right target | same 8 rows, same targets, structural ToID |
| `TestIssue4366_DjangoModelFields_AreMembers` | both relational fields emit a target edge naming `Group` and `User` | same two, structural ToID |
| `TestIssue4366_SQLAlchemyFields_AreMembers` | (a) an edge to a related model exists **and** (b) its address RESOLVES — asserted by looking `Class:Order` up in the symbol table | (a) unchanged. (b) is now pinned **one layer lower and more strictly**: drive the real production path (`BuildIndex` → `ReferencesEmbedded`) and require the emitted ToID to come back rewritten to a real entity ID, with a `boundTargets == 0` vacuity guard. Dropping (b) would have been the weakening; it is kept |

`pyClassRef` is untouched and still correct for the CONTAINS FromID, which was never re-addressed.

## Mutants

Both scored with `-count=1` off a green baseline, edit proven applied by exact occurrence count **inside the named function**, restored by `cp` from a shasum-verified backup.

- **M1 — permissive `self_ref`** (`return m[2], "", m[2] == ownerClass` in `djangoRelTargetDetail`, 1 occurrence, inside `djangoRelTargetDetail`): **DEAD** — `TestPythonFieldTargetType_SelfRefIsTheStringFormOnly`.
- **M2 — file segment dropped from the address** (`BuildComponentStructuralRef("python", "", targetClass)`, 1 occurrence, inside `referencesClassEdge`): **DEAD** — 7 tests, including `TestPythonFieldTargetType_NeverBindsToANonDeclaringFile`. That is the interesting kill: a fileless address falls through to the cross-file unique tier, which is precisely the over-fire the negative test exists to catch.
- **M3 (reviewer's) — the cross-file tier accepts an ambiguous name** (`continue` on collision in `lookupUniqueRealComponentByName`, refs.go, 1 occurrence, inside that function): was **ALIVE in this package** — the tier this PR routes 847 edges through was graded only by `internal/resolve` helper-equivalence tests. `TestPythonFieldTargetType_CrossFileTier_BindsUniqueRefusesAmbiguous` now kills it: re-scored after adding the test, `./internal/custom/python/` goes **RED on that test alone**. Restored by `cp` from a shasum-verified backup (`786d695b…` both ways), zero `MUTANT` text remaining, green after.
- **CK-1 (reviewer's, posted on this PR)** — **DEAD**, and worth reading for *which* tests killed it: **three of the six are the #6988 GFK tests**, written for an unrelated defect. They caught a corrupted language segment because they assert **resolved targets** rather than emitted strings, so a broken address breaks them even though nothing about GFKs changed. That is evidence the address is graded more broadly than by the tests added for it here.

### Correction carried from review

The ambiguous-name case was called unasserted during review; it is not. `TestPythonFieldTargetType_BindsWhenTheNameIsDeclaredInTwoFiles` covers it and asserts it **binds** — which is the correct expectation, because the consumer's file declares `Customer` too and the same-file `lookupLocationKind` tier resolves before the cross-file one. The genuinely-ambiguous direction (the consumer's file does NOT declare the name and two other files do) is the one that must dangle, and it is `TestPythonFieldTargetType_NeverBindsToANonDeclaringFile` that forbids guessing there.

## Gates

- `go test -count=1 ./internal/custom/python/` — ok
- `go test -count=1 ./internal/resolve/` — ok
- `go test -count=1 ./cmd/grafel/` — **PASS, 153.5s, 0 failures**, re-run on the final head under a FRESH isolated `GRAFEL_HOME` + `GRAFEL_DAEMON_ROOT`. Two earlier attempts on this box wedged in `TestRebuild_MarksTheGroupForeground` → `internal/gitmeta.runGitReal`, on a `git status` subprocess that hung and left a `(git)` process behind (goroutine dump captured). That is environmental — a stale state dir or a concurrent Go test run reproduces it — not this change: filtered to non-comment lines, `git diff` of the non-test source since the previously green run is **empty**.
- `scripts/quality/run.sh --ratchet` — **OK, 38 gated fixtures held their baseline (29 at 100% must-have recall)**. **No fixture moved**, so no baseline was regenerated and no hand-edit was needed; `git status` shows no `internal/quality/**` change.

Second-repo control: `django-realworld` — 3 DRF `field_target_type` edges, bound before and after, same targets.

## Two things this deliberately does NOT fix

- **`LogEntry.user` is a live instance of #6990, and this PR CAN turn it into a bound wrong edge. My earlier claim that it could not was false of the mechanism.** `user = models.ForeignKey(settings.AUTH_USER_MODEL, …)` at `django/contrib/admin/models.py:63` emits a target of `ContentType`: the 400-char `fullRHS` window reads past the dotted `settings.X` (which the regex's `[A-Z]` symbol alternative correctly rejects) into the **next** field's `ForeignKey(ContentType`.

  Minimal repro, now pinned as `TestPythonFieldTargetType_KnownWrong_6990_OverReadBindsToTheNextFieldsTarget`:

  ```
  parent address (Class:ContentType):   LogEntry.user  DANGLING
  A′     address (structural):          LogEntry.user  BOUND -> ContentType@contenttypes/models.py
  ```

  A dangling wrong edge promoted to a **bound** wrong edge is #6369's hazard, and it is worse than a dangle because nothing flags it. Three things must be said plainly:

  1. **On django it stays dangling, but the guarantee is not this address's.** Re-measured under A′ (fresh index, gate-ON): the ToID is `ext:django:ContentType` and it is **DANGLING** — the external-synthesis pass intercepts first. `class ContentType(models.Model)` is declared **exactly once** in the corpus (`django/contrib/contenttypes/models.py:134`), so absent that interception the cross-file unique tier *would* bind it.
  2. **My over-fire controls cannot detect this class.** Both "0 bound to a non-declaring file" and `NeverBindsToANonDeclaringFile` grade the **file** the resolver picked. Here the file is right — `ContentType` really is declared there. The wrongness is in the **name the extractor chose**, a dimension neither control observes.
  3. **A′ enlarges the population that could exhibit it**, from 39 bound edges to 886 — more than twenty-fold.

  **Re-measured #6990 corpus incidence under A′** (the original zero-incidence figure was taken against the old, never-binding address and no longer describes the shipped graph): of 949 `field_target_type` edges, the over-read population — custom target disagreeing with the core producer's target for the same field — is **1**, `LogEntry.user`, and it is dangling. So bound-wrong incidence on django under A′ is **0**, population **1**. The wrong *name* is #6990's defect, not this PR's; the promotion path is this PR's and is now pinned. Deleting that test is the expected consequence of fixing #6990.
- **The duplicate-producer question.** 765 of 863 of these edges duplicate a core edge that already binds. Reconciling the two producers — stamping `ref_kind: field_target_type` on the core edge and dropping the custom Django duplicate — is the better end state and is being filed separately; it does not replace this change, because 97 django fields have no core twin and the seven non-Django Python ORM frameworks (SQLAlchemy, peewee, pony, beanie, mongoengine, tortoise, DRF) have no core producer at all.

`Refs #6986` rather than `Closes` — four languages (golang, javascript, ruby, java) still ship `Class:<Name>`, and `internal/resolve/refs.go:4574`'s Python disposition guard is untouched: 86.5% of what it routes `dynamic` is the ORM-queries family it was written for, so it must not be narrowed on this evidence.

### DIFF
--- a/internal/custom/python/django.go
+++ b/internal/custom/python/django.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"regexp"
+	"strconv"
 	"strings"
 
 	"go.opentelemetry.io/otel"
@@ -505,7 +506,7 @@ func (e *DjangoExtractor) Extract(ctx context.Context, file extractor.FileInput)
 			if strings.HasSuffix(callee, "Serializer") && !strings.Contains(callee, ".") &&
 				callee != "Serializer" && callee != "ModelSerializer" {
 				serEnt.Relationships = append(serEnt.Relationships,
-					referencesClassEdge(className+"."+attr, callee, "drf", attr))
+					referencesClassEdge(file.Path, className+"."+attr, callee, "drf", attr))
 			}
 		}
 
@@ -715,9 +716,38 @@ func (e *DjangoExtractor) Extract(ctx context.Context, file extractor.FileInput)
 				continue
 			}
 			fullRHS := body[fIdx[0]:min(fIdx[0]+400, len(body))]
-			if target := djangoRelTarget(fullRHS, className); target != "" {
-				modelEnt.Relationships = append(modelEnt.Relationships,
-					referencesClassEdge(className+"."+attr, target, "django", attr))
+			if target, rawFKString, isSelf := djangoRelTargetDetail(fullRHS, className); target != "" {
+				rel := referencesClassEdge(file.Path, className+"."+attr, target, "django", attr)
+				// #6986 — carry the RELATION-SHAPE properties the core
+				// extractor's parallel edge carries. This is not decoration.
+				// internal/extractors/python/django_relational.go emits a
+				// REFERENCES edge from the SAME field entity to the SAME target,
+				// stamped `django_rel` / `self_ref` / `django_fk_string`. While
+				// this pass's ToID was the never-binding `Class:<Target>` the two
+				// edges could not collide, so both survived. Now that the
+				// structural ToID resolves to the same entity ID, graph assembly
+				// DEDUPES the pair and keeps exactly one property set — measured
+				// on django gate-ON: REFERENCES edges carrying `django_rel` fell
+				// 1018 → 258 when the address changed and these three properties
+				// were not carried across. Restoring them here is what makes the
+				// address change property-preserving rather than a silent
+				// substitution (#6973).
+				//
+				// The values mirror the core extractor's semantics EXACTLY:
+				// `django_rel` is the constructor's leaf name (ForeignKey /
+				// OneToOneField / ManyToManyField, and the *ForeignKey suffix
+				// forms isDjangoRelationalField admits); `self_ref` is true ONLY
+				// for the literal string target `'self'` — NOT for a symbol
+				// target that happens to name the enclosing class, which is
+				// parseTargetExpr's rule; `django_fk_string` is set only for the
+				// string form and carries the raw value BEFORE app_label
+				// stripping.
+				rel.Properties.Set("django_rel", djangoRelCtorLeaf(rhs))
+				rel.Properties.Set("self_ref", strconv.FormatBool(isSelf))
+				if rawFKString != "" {
+					rel.Properties.Set("django_fk_string", rawFKString)
+				}
+				modelEnt.Relationships = append(modelEnt.Relationships, rel)
 			}
 		}
 		out = append(out, modelEnt)
@@ -958,22 +988,50 @@ func isDjangoRelationalField(rhs string) bool {
 // form (`ForeignKey(Model, ...)` / `ForeignKey(to=Model, ...)`). For 'self' it
 // returns the enclosing model class so the edge self-references the owner.
 // Returns "" when no recognizable target is present (e.g. lazy callables).
-func djangoRelTarget(rhs, ownerClass string) string {
+// djangoRelCtorLeaf returns the leaf constructor name of a relational field's
+// RHS — `models.ForeignKey` → `ForeignKey`. Mirrors the leaf-stripping
+// isDjangoRelationalField already performs, so the `django_rel` property this
+// pass stamps carries the same vocabulary as the core extractor's
+// `djangoFieldTypes` key (#6986).
+func djangoRelCtorLeaf(rhs string) string {
+	ctor := rhs
+	if dot := strings.LastIndexByte(ctor, '.'); dot >= 0 {
+		ctor = ctor[dot+1:]
+	}
+	return ctor
+}
+
+// djangoRelTargetDetail is djangoRelTarget plus the two facts the `self_ref`
+// and `django_fk_string` edge properties need: whether the target was written
+// as the literal string `'self'`, and the raw string-literal value before the
+// app_label segment is stripped. Returns ("", "", false) exactly where
+// djangoRelTarget returns "".
+//
+// `isSelf` is deliberately NOT "the target resolves to the enclosing class":
+// `ForeignKey(Person)` written inside `class Person` returns isSelf=false, the
+// same as internal/extractors/python/django_relational.go's parseTargetExpr,
+// so the two producers cannot disagree on the property.
+func djangoRelTargetDetail(rhs, ownerClass string) (target, rawFKString string, isSelf bool) {
 	m := djangoModelRelTargetRe.FindStringSubmatch(rhs)
 	if m == nil {
-		return ""
+		return "", "", false
 	}
 	if m[1] != "" { // string form
 		raw := m[1]
 		if raw == "self" {
-			return ownerClass
+			return ownerClass, "self", true
 		}
 		if dot := strings.LastIndexByte(raw, '.'); dot >= 0 {
-			return raw[dot+1:] // strip app_label
+			return raw[dot+1:], raw, false
 		}
-		return raw
+		return raw, raw, false
 	}
-	return m[2] // symbol form
+	return m[2], "", false // symbol form
+}
+
+func djangoRelTarget(rhs, ownerClass string) string {
+	target, _, _ := djangoRelTargetDetail(rhs, ownerClass)
+	return target
 }
 
 // extractBalancedBrackets returns the content of a [...] list starting at openPos.
--- a/internal/custom/python/django.go
+++ b/internal/custom/python/django.go
@@ -982,17 +982,21 @@ func isDjangoRelationalField(rhs string) bool {
 		strings.HasSuffix(ctor, "ManyToManyField")
 }
 
-// djangoRelTarget extracts the bare target-model class name from a relational
-// field declaration's argument blob. Handles the string form
-// (`ForeignKey('app.Model', ...)` / `ForeignKey('self', ...)`) and the symbol
-// form (`ForeignKey(Model, ...)` / `ForeignKey(to=Model, ...)`). For 'self' it
-// returns the enclosing model class so the edge self-references the owner.
-// Returns "" when no recognizable target is present (e.g. lazy callables).
 // djangoRelCtorLeaf returns the leaf constructor name of a relational field's
 // RHS — `models.ForeignKey` → `ForeignKey`. Mirrors the leaf-stripping
-// isDjangoRelationalField already performs, so the `django_rel` property this
-// pass stamps carries the same vocabulary as the core extractor's
-// `djangoFieldTypes` key (#6986).
+// isDjangoRelationalField already performs.
+//
+// The resulting `django_rel` vocabulary is a strict SUPERSET of the core
+// extractor's, not the same set (#6986). Core gates on
+// `djangoRelationalFieldTypes`, an exact three-name allow-list; this lane gates
+// on `isDjangoRelationalField`, a `*ForeignKey` / `*OneToOneField` /
+// `*ManyToManyField` SUFFIX rule. So a third-party subclass such as
+// `TreeForeignKey` reaches here and is stamped `django_rel: TreeForeignKey`, a
+// value core never emits — because core emits no edge for that field at all.
+// That direction is a gain and is what keeps the property-preservation claim
+// safe: this lane can only ADD a `django_rel`-carrying edge where core had
+// none, never rename one core already produced. Pinned by
+// TestIssue6988_LegitimateRelationalFieldsSurvive's `Book.category` row.
 func djangoRelCtorLeaf(rhs string) string {
 	ctor := rhs
 	if dot := strings.LastIndexByte(ctor, '.'); dot >= 0 {
@@ -1029,6 +1033,12 @@ func djangoRelTargetDetail(rhs, ownerClass string) (target, rawFKString string, 
 	return m[2], "", false // symbol form
 }
 
+// djangoRelTarget extracts the bare target-model class name from a relational
+// field declaration's argument blob. Handles the string form
+// (`ForeignKey('app.Model', ...)` / `ForeignKey('self', ...)`) and the symbol
+// form (`ForeignKey(Model, ...)` / `ForeignKey(to=Model, ...)`). For 'self' it
+// returns the enclosing model class so the edge self-references the owner.
+// Returns "" when no recognizable target is present (e.g. lazy callables).
 func djangoRelTarget(rhs, ownerClass string) string {
 	target, _, _ := djangoRelTargetDetail(rhs, ownerClass)
 	return target
--- a/internal/custom/python/django_gfk_6988_test.go
+++ b/internal/custom/python/django_gfk_6988_test.go
@@ -194,6 +194,18 @@ func fieldTargetEdgesFrom6988(rels []types.RelationshipRecord, from string) []ty
 	return out
 }
 
+// fieldTargetRef6988 is the ToID a field_target_type edge carries for a target
+// declared in the fixture file these tests extract ("models.py"). #6986 moved
+// that address off the bare-name `Class:<Target>` stub (which never bound once
+// a second entity shared the name) and onto the structural
+// `scope:component:class:python:<file>:<Target>` form. The assertions below are
+// unchanged in INTENT — "this field emits a target edge pointing at <Target>" —
+// only the dialect the target is named in has moved. `pyClassRef` is still the
+// right helper for the CONTAINS FromID, which was never re-addressed.
+func fieldTargetRef6988(target string) string {
+	return extractor.BuildComponentStructuralRef("python", "models.py", target)
+}
+
 func hasEdge6988(rels []types.RelationshipRecord, from, to, kind string) bool {
 	for _, r := range rels {
 		if r.FromID == from && r.ToID == to && r.Kind == kind {
@@ -239,8 +251,8 @@ func TestIssue6988_NamedGFKFieldEmitsNoTargetEdge(t *testing.T) {
 	// Positive control: the sibling REAL ForeignKey on the same class still
 	// resolves, so the negative above is not a vacuous "this file produced
 	// nothing" pass.
-	if !hasEdge6988(rels, "GenericB1.content_type", pyClassRef("ContentType"), string(types.RelationshipKindReferences)) {
-		t.Error("GenericB1.content_type -> Class:ContentType missing: the fix removed a LEGITIMATE edge (#6988)")
+	if !hasEdge6988(rels, "GenericB1.content_type", fieldTargetRef6988("ContentType"), string(types.RelationshipKindReferences)) {
+		t.Error("GenericB1.content_type -> ContentType missing: the fix removed a LEGITIMATE edge (#6988)")
 	}
 	// The GFK field itself must still be a CONTAINS member of its model — the
 	// fix removes the wrong REFERENCES edge, not the field node.
@@ -263,8 +275,8 @@ func TestIssue6988_NamedGFKFieldEmitsNoTargetEdge(t *testing.T) {
 func TestIssue6988_WrongButBoundEdgeShapeIsGone(t *testing.T) {
 	rels := djangoEdges6988(t, gfkSrc6988)
 
-	if hasEdge6988(rels, "Book.created_by", pyClassRef("created_by_ct"), string(types.RelationshipKindReferences)) {
-		t.Error("Book.created_by -> Class:created_by_ct still emitted: the wrong-but-BOUND edge shape " +
+	if hasEdge6988(rels, "Book.created_by", fieldTargetRef6988("created_by_ct"), string(types.RelationshipKindReferences)) {
+		t.Error("Book.created_by -> created_by_ct still emitted: the wrong-but-BOUND edge shape " +
 			"(#6369's wrong-node hazard) survives (#6988)")
 	}
 	if got := fieldTargetEdgesFrom6988(rels, "Book.created_by"); len(got) != 0 {
@@ -272,8 +284,8 @@ func TestIssue6988_WrongButBoundEdgeShapeIsGone(t *testing.T) {
 			len(got), got[0].ToID)
 	}
 	// The sibling that MINTS the `created_by_ct` node keeps its own real edge.
-	if !hasEdge6988(rels, "Book.created_by_ct", pyClassRef("ContentType"), string(types.RelationshipKindReferences)) {
-		t.Error("Book.created_by_ct -> Class:ContentType missing (#6988)")
+	if !hasEdge6988(rels, "Book.created_by_ct", fieldTargetRef6988("ContentType"), string(types.RelationshipKindReferences)) {
+		t.Error("Book.created_by_ct -> ContentType missing (#6988)")
 	}
 }
 
@@ -320,8 +332,8 @@ class Book(models.Model):
 		"Book.category":    "Category",  // third-party *ForeignKey subclass
 	}
 	for from, target := range want {
-		if !hasEdge6988(rels, from, pyClassRef(target), string(types.RelationshipKindReferences)) {
-			t.Errorf("%s -> Class:%s missing — #6988 broke a legitimate relational field", from, target)
+		if !hasEdge6988(rels, from, fieldTargetRef6988(target), string(types.RelationshipKindReferences)) {
+			t.Errorf("%s -> %s missing — #6988 broke a legitimate relational field", from, target)
 		}
 	}
 }
--- a/internal/custom/python/field_target_address_6986_test.go
+++ b/internal/custom/python/field_target_address_6986_test.go
@@ -1,0 +1,301 @@
+package python_test
+
+import (
+	"context"
+	"sort"
+	"testing"
+
+	extreg "github.com/cajasmota/grafel/internal/extractor"
+	"github.com/cajasmota/grafel/internal/extractors"
+	"github.com/cajasmota/grafel/internal/graph"
+	"github.com/cajasmota/grafel/internal/resolve"
+	"github.com/cajasmota/grafel/internal/types"
+
+	_ "github.com/cajasmota/grafel/internal/custom/python"
+	_ "github.com/cajasmota/grafel/internal/extractors/python"
+)
+
+// field_target_address_6986_test.go — the Python custom lane's field→declared-
+// type address (issue #6986).
+//
+// MEASURED PREMISE, gate-ON (`WithCustomExtractors(true)`; the custom lane is
+// OFF by default, so user-facing incidence today is zero — #6966). On the
+// `django` corpus repo at 64ffa3d22: 863 `ref_kind: field_target_type` edges,
+// 39 bound (4.5%). Re-addressed from `Class:<Target>` to
+// `scope:component:class:python:<consumer file>:<Target>`: 949 edges, 886 bound
+// (93.4%), zero previously-bound key lost, zero bound to a file that does not
+// declare the name.
+//
+// The address is built from the CONSUMER's file, not the target's — which a
+// per-file pass-1 extractor cannot know. That is the dialect the resolver
+// already speaks for Python: internal/resolve/refs.go lookupStructural resolves
+// this shape same-file via lookupLocationKind and then, for `component` scope
+// with lang=="python" only, cross-file via lookupUniqueRealComponentByName.
+
+const modelsPath6986 = "shop/models.py"
+
+// twoFilesSrc6986 is the shape the whole issue turns on: a target model name
+// declared in TWO files. `Class:Customer` goes AMBIGUOUS here and dangles; the
+// file-scoped structural address binds to the RIGHT one.
+const ownerSrc6986 = `from django.db import models
+
+
+class Customer(models.Model):
+    name = models.CharField(max_length=100)
+
+
+class Order(models.Model):
+    buyer = models.ForeignKey(Customer, on_delete=models.CASCADE)
+    watchers = models.ManyToManyField('self', blank=True)
+`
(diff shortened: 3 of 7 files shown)


cajasmota:
## Coordinator mutant CK-1 on `2d9d1debf` — DEAD. Plus one prose claim to fix before merge.

Own detached worktree, `git status --porcelain` empty at start and after restore, restored by `cp` from a SHA-256-verified backup (`eb54d59d…` both ways), zero `MUTANT` text remaining, green baseline immediately before. macOS/APFS, gate-ON.

**CK-1 — corrupt the address's LANGUAGE segment.** M1 attacked a property value and M2 dropped the file segment; nobody touched the dialect itself. The whole design rests on the address being the dialect `internal/resolve` already speaks, so I changed one word of it:

```go
-  ToID: extractor.BuildComponentStructuralRef("python", filePath, targetClass),
+  ToID: extractor.BuildComponentStructuralRef("go", filePath, targetClass), // MUTANT CK-1
```

Edit proven landed: exactly 1 occurrence, hit set `[131]`, **enclosing function asserted** to be `referencesClassEdge`, `go vet` clean before scoring.

**Result: DEAD** — six tests in `./internal/custom/python/`, `./internal/resolve/` green:
`TestIssue6988_NamedGFKFieldEmitsNoTargetEdge`, `TestIssue6988_WrongButBoundEdgeShapeIsGone`, `TestIssue6988_LegitimateRelationalFieldsSurvive`, `TestPythonFieldTargetType_BindsWhenTheNameIsDeclaredInTwoFiles`, `TestIssue4366_{Django,SQLAlchemy}ModelFields_AreMembers`.

Worth noting **which** tests died: three of them are the #6988 GFK tests, which exist for a different defect entirely. They caught this because they assert *resolved* targets rather than emitted strings — so the address change is graded by tests nobody wrote for it. That is the second time this cycle an enumerated positive set has outperformed the pin it was written as.

**A correction to my own review brief**, since I got this wrong when scoping the reviewer: I told them "never binds to a non-declaring file" and "never binds an ambiguous name" are different claims and only the first is asserted. `TestPythonFieldTargetType_BindsWhenTheNameIsDeclaredInTwoFiles` shows the ambiguous case **is** asserted — and asserts it *binds*, which is right, because the consumer's file declares it and the same-file tier wins before the uniqueness tier is consulted. The reviewer should still verify it binds to the **same-file** target rather than merely binding, but the axis is covered and I said it wasn't.

## The one thing to fix before I merge

`helpers.go`'s doc comment ends:

> "…an ambiguous or absent target still dangles — **never more edges than before**, and never a guessed one."

**The PR's own measurement says 863 → 949 edges.** The two readings of that sentence come apart: the producer emits the same number of *records*, but the graph gains 86 *edges*, because previously-identical `Class:` ToIDs collapsed and distinct structural ToIDs no longer do. Under the graph reading — the one a reader of a resolver comment will take — the sentence is false, and it is contradicted by a number in the same PR.

Please qualify it to what you measured: same records, more distinct edges, and say why (the collapse). **This is the dominant defect class on this board and it is in the doc comment of the change that fixes an address** — exactly where the next person will trust it. Everything else in that comment is unusually good: it states the measured reason, names the two tiers, and explicitly flags that this is *not* same-file-only unlike #6984. That is why the one unqualified absolute stands out.

Not blocking on the reviewer's verdict — fix the sentence and I will merge on green.

cajasmota:
## Independent adversarial review — **APPROVE with two required follow-ups**

Head reviewed: **`148e65a13`** (code identical to `2d9d1debf`; the delta is comment-only, `helpers.go` +17/−2 — verified by `git diff 2d9d1debf..148e65a13`, one file, doc block only). All mutants and suites were scored at `2d9d1debf`; no re-score needed.

Measured in my own worktree, absolute path
`/private/tmp/claude-501/-Users-jorgecajas-Projects-archigraph/c0fefed4-0195-4c81-9a94-07a7145710a4/scratchpad/review-6994-zx9/wt`, macOS/APFS, `GOMAXPROCS=3`, `nice -n 10`, `./cmd/grafel/` and the ratchet under a **fresh isolated `GRAFEL_HOME` + `GRAFEL_DAEMON_ROOT`** per run. **Every figure below is gate-ON; the custom lane is OFF by default (#6966), so user-facing incidence on a default `grafel index` today is ZERO.**

Gates reproduced, all green, exit code checked (not tailed):
`./internal/custom/python/` ok · `./internal/resolve/` ok · `./cmd/grafel/` ok **157.2s** · `scripts/quality/run.sh --ratchet` **OK — 38 gated fixtures held their baseline (29 at 100% must-have recall)**, `git status --porcelain` **empty** → **confirmed: no fixture moved, nothing regenerated.**

---

### (a) Can the cross-file tier bind an ambiguous name? **No — constructed and confirmed.**

I built the case your suite does not: consumer `orders/models.py` declares nothing, `Customer` declared in **both** `shop/models.py` and `billing/models.py`.

| fixture | result |
|---|---|
| cross-file **ambiguous** (2 declaring files, consumer declares neither) | **DANGLING** `scope:component:class:python:orders/models.py:Customer` ✅ |
| cross-file **unique** (1 declaring file) | **BOUND** → `Customer@shop/models.py` ✅ |
| target **absent** | **DANGLING** ✅ |

The tier refuses correctly. But — and this is the point your own comment now leans on — **`TestPythonFieldTargetType_NeverBindsToANonDeclaringFile` does not grade that refusal.** Its fixture has `Order` and `Customer` in the *same* file, so the same-file tier resolves first and the cross-file tier never fires. Proof: mutant **M3** below makes the cross-file tier accept an ambiguous name and `./internal/custom/python/` stays **green**. "Non-declaring file" and "not ambiguous" are different claims, and only the first is asserted here.

### (b) Does the `django_rel` pair set lose zero members? **Semantics verified; the corpus set-diff I did not re-run.**

I did not re-index django (cost), so the `LOST 0` number is yours, not mine. What I *did* do is check the only way it could be false — a property-value divergence between the two producers on a form your suite does not cover. Core `parseTargetExpr` vs custom `djangoRelTargetDetail`, same fixture, per field:

| field | core `django_rel`/`self_ref`/`django_fk_string` | custom |
|---|---|---|
| `ForeignKey('auth.User')` | `ForeignKey` / `false` / **`auth.User`** | identical ✅ |
| `ForeignKey('Customer')` | `ForeignKey` / `false` / `Customer` | identical ✅ |
| `OneToOneField(Customer)` | `OneToOneField` / `false` / *(unset)* | identical ✅ |
| `TreeForeignKey('Customer')` | **core emits nothing** | `TreeForeignKey` / `false` / `Customer` |

The **app-labelled string form is not in your suite** and it agrees — that is the case that would have silently broken `ResolveDjangoStringFKRefs`. Good.

One prose correction: the comment says `django_rel` "carries the same vocabulary as the core extractor's `djangoFieldTypes` key". `djangoRelationalFieldTypes` is an exact 3-name allow-list; `isDjangoRelationalField` is a `*ForeignKey` suffix deny-list. `TreeForeignKey` is emitted by the custom lane with a `django_rel` value the core vocabulary never produces. It is a **gain**, not a loss, so the set claim survives — but the sentence asserts more than is true and nothing observes it.

### (c) Are the +86 edges distinct relationships? **Yes on the TO side; the FROM side is still one collapsed stub.**

Named, from `archigraph-corpora/django` (read-only, not copied — grep only): `class Person(models.Model)` with `friends = models.ManyToManyField("self")` is declared in **four distinct apps** — `tests/test_runner/models.py:4`, `tests/m2m_recursive/models.py:22`, `tests/queries/models.py:690`, `tests/m2m_signals/models.py:24` (plus `Person`-shaped namesakes in `aggregation`, `annotations`, `aggregation_regress`, `m2m_through`). Four declarations, four relationships — **four edges is right, not inflation.**

Caveat for #6726's arithmetic: the FromID is the bare `Person.friends`, identical across all four files and still unresolved, so the graph gains four outbound edges hanging off **one** collapsed source node. That is pre-existing FROM-side under-resolution, not something this PR introduces — but "+86 distinct edges" is only half true: 86 distinct *targets*, from fewer distinct *sources*.

The new comment's replacement claim — "the producer emits the SAME NUMBER OF RECORDS, one per relational field, unchanged" — **is true.** Every `referencesClassEdge` call site in the diff is a 1:1 `-`/`+` replacement adding the `filePath` argument; no call site was added or removed, and the django loop still emits at most one record per `djangoModelFieldRe` match.

### (d) Does any of the five updated tests pin less? **No.**

The three #6988 tests swap `pyClassRef(X)` → `fieldTargetRef6988(X)` inside the same `hasEdge6988(from,to,kind)` predicate; the forbidden row in `WrongButBoundEdgeShapeIsGone` is unchanged **and** still carries the independent `len(fieldTargetEdgesFrom6988(...)) != 0` check. `TestIssue4366_DjangoModelFields_AreMembers` keeps both targets. `TestIssue4366_SQLAlchemyFields_AreMembers` is **strictly stronger, claim verified**: the old form asserted resolution for `Class:Order` only, via a tier the new address does not use; the new form drives `BuildIndex` → `ReferencesEmbedded` and requires **every** `field_target_type` edge in the fixture to come back as a real entity ID, with a `boundTargets == 0` vacuity guard.

---

### Mutants (4 of my own; M1/M2 not re-run)

Each proven applied by exact occurrence count **inside the named enclosing function**, `go vet` clean, green baseline immediately before, `-count=1`, restored by `cp` from a shasum-verified backup (`git checkout` never used; tree confirmed clean after each).

| # | mutant | enclosing func | verdict |
|---|---|---|---|
| **M3** | cross-file tier **accepts an ambiguous name** (`uniqueMatchInFamily(..., true)` + `continue` on collision) | `lookupUniqueRealComponentByName` (refs.go, 1 occ) | **DEAD — but only in `internal/resolve`** (`TestKindIDs_EntryPointBattery`, `TestLocKindMerge_EntryPointEquivalence`, both direct-helper equivalence tests). **`./internal/custom/python/` GREEN.** This PR's own suite does not grade the refusal it depends on. |
| **M4** | the three property stamps dropped | `(*DjangoExtractor).Extract` (django.go, 1 occ) | **DEAD** — `..._CarriesTheRelationShapeProperties`, `..._SelfRefIsTheStringFormOnly` |
| **M5** | file segment points at a **non-declaring** file (`"zzz_nondeclaring/"+filePath`) — the inverse of your M2 | `referencesClassEdge` (helpers.go, 1 occ) | **DEAD** — 6 tests, incl. `..._BindsWhenTheNameIsDeclaredInTwoFiles` behaviourally (not just address-literal pins) |
| **M6** | `lang == "python"` guard removed → cross-file component tier widened to **every language** | `lookupStructural` (refs.go, 1 occ) | **ALIVE — distinguishable and ungraded.** `./internal/custom/python/` + `./internal/resolve/` green **and the full 38-fixture golden ratchet green.** Pre-existing code, untouched by this PR, so not this PR's debt — but it is the tier this PR now routes 847 edges through. |

---

## The one finding that needs an answer before gate-ON

**The PR body's `LogEntry.user` sentence — "this PR neither fixes it nor turns it into a bound wrong edge" — is true of the django corpus and false of the mechanism.**

Minimal faithful reproduction of #6990's over-read (`settings.AUTH_USER_MODEL` target, 400-char `fullRHS` window reads into the next field's `ForeignKey(ContentType`), with `ContentType` declared uniquely in one other in-tree file:

```
PARENT address (Class:ContentType):        LogEntry.user  DANGLING
A′     address (structural):               LogEntry.user  BOUND -> ContentType@contenttypes/models.py
```

That is a dangling edge traded for a **confidently-wrong bound** edge — #6369's hazard, demonstrated, not argued. On the django corpus it does not fire only because `ext:django:ContentType` intercepts first (I did not reproduce that; it is your measurement). In the corpus, `class ContentType(models.Model)` is declared **exactly once** (`django/contrib/contenttypes/models.py:134`), so the cross-file unique tier *would* bind it — the guarantee rests entirely on the external-synthesis pass, not on this address logic.

Critically, **your over-fire controls cannot see this**: both "0 bound to a file that does not declare the name" and `NeverBindsToANonDeclaringFile` are satisfied, because `ContentType` *is* declared in the file it bound to. The wrongness is in the target **name** the extractor chose, not the **file** the resolver picked. A′ takes that class of defect from 39 bound edges to 886.

### Required follow-ups (not blocking this merge — lane is gate-OFF, ratchet green, net 4.5% → 93.4%)

1. **Scope the sentence and pin the shape.** Add the two-file `LogEntry.user` fixture above as a forbidden row (`LogEntry.user` must not bind to `ContentType`), and re-measure #6990's corpus incidence *under A′* — the original zero-incidence measurement was taken against the old, never-binding address and no longer describes the graph this ships.
2. **Grade the cross-file ambiguity refusal in this package.** M3 shows it is currently graded only by `internal/resolve` helper-equivalence tests. Add the fixture I constructed — consumer file declaring neither, `Customer` in two others, edge must **dangle** — as a `field_target_address_6986_test.go` case. It is ~20 lines and it is the only thing that would grade the tier this PR newly depends on. Optionally file M6 (`lang == "python"` guard ungraded by anything, ratchet included) as its own issue.

Nothing merged, nothing pushed. Worktree removed.

cajasmota:
## Coordinator mutant CL-1 on `6a38db207` — DEAD. Approving; merging on green.

Own detached worktree, `git status --porcelain` empty at start and after restore, restored by `cp` from a SHA-256-verified backup (`563d0c1e…` both ways), zero `MUTANT` text remaining, green baseline immediately before. macOS/APFS.

**First, the reason CK-1 still counts.** I verified rather than assumed: the non-test diff `148e65a13..6a38db207` restricted to `helpers.go` and `django.go`, with comment and blank lines filtered out, is **empty**. Every production change since the head I scored is inside a comment, so CK-1's DEAD verdict on the address dialect carries and does not need re-running.

**CL-1 — attack the vacuity FIX, not the production behaviour.** The new value in this round is the keying you introduced after finding that `ReferencesEmbedded` rewrites `FromID` in place, so post-resolution `FromID == "Order.buyer"` matched nothing and two test halves were passing vacuously. `edgeKey6986` keys on owner Name + `field_name` precisely because resolution touches neither. **So `field_name` is now load-bearing for the tests, not just a property** — and nothing had graded it in that new role.

I blanked it in `referencesClassEdge` only:

```go
-  {K: "field_name", V: fieldName},
+  {K: "field_name", V: ""}, // MUTANT CL-1: field_name blanked in referencesClassEdge only
```

**The assertion caught a real hazard before the edit landed.** My first attempt asserted one occurrence; there are **two** — `containsFieldEdge` at `:83` stamps the identical literal. A naive replace would have hit the wrong producer, and a green result would then have read as ALIVE and argued the keying was ungraded. I re-applied by line with the enclosing function asserted to be `referencesClassEdge` **and** asserted not to be `containsFieldEdge`. Fifth time this cycle the hit-set assertion has prevented an inverted verdict.

**Result: DEAD** — four tests, including both new ones:
`TestPythonFieldTargetType_BindsWhenTheNameIsDeclaredInTwoFiles`, `TestPythonFieldTargetType_CrossFileTier_BindsUniqueRefusesAmbiguous`, `TestPythonFieldTargetType_CarriesTheRelationShapeProperties`, `TestPythonFieldTargetType_SelfRefIsTheStringFormOnly`.

So the vacuity fix is itself graded: break the key and the tests fail loudly rather than passing on zero matches. That is the property that was missing before, and it is now the property that would catch its own regression.

## On the round

**Finding your own vacuous tests is the best thing in this PR.** Two assertions were passing on zero matches because resolution rewrites the field they matched on — that is the failure mode where a suite reports success for doing nothing, and it is very hard to see from the outside. You found it while writing an unrelated test, said so, and fixed the whole class with `checked == 0` guards rather than patching the two you noticed.

The `LogEntry.user` correction is right and is now pinned as **known-wrong** — DANGLING under the parent, **BOUND to `ContentType@contenttypes/models.py`** under A′ — with a `t.Skip` escape naming #6990. That is the honest shape: the mechanism can produce a bound wrong edge, the django outcome rests on the `ext:django:` synthesis pass rather than on this address logic, and neither over-fire control can see the class because both grade the *file* while the wrongness is in the *name*. Recording that as a test rather than a sentence is what stops it being rediscovered.

The `django_rel` vocabulary correction went the right way too: a strict **superset** of core's three-name allow-list, with `TreeForeignKey` as the witness — a gain, and the direction that makes the pair-set claim safe.

**Merging on green** once `ci:full`'s legs report; the label is re-triggered.

PR #6995, opened 2026-09-08
fix(#6978): 26 custom-extractor path gates can never fire — the enumeration, and a guard that fails when a 27th appears

### LINKED ISSUE #6978: MEASURED: 15 unreachable custom-extractor gates across 9 files / 6 packages — the criterion is LANGUAGE ROUTING via CustomExtractorsFor, not TSTree==nil as I originally wrote (.sql has a grammar and is still dead)
Found during #6975's gate work, **verified in source by me before filing**.

## The trap

A `.razor` suffix check inside a **C# custom extractor is unreachable**, and looks entirely reasonable.

The chain:

1. `.razor` classifies as language **`"razor"`**, not `"csharp"` (`internal/classifier/classifier.go`).
2. There is **no tree-sitter grammar for razor** — verified, nothing registers one.
3. So `file.TSTree == nil` for every `.razor` file.
4. `cmd/grafel/index.go`'s custom dispatch requires `file.TSTree != nil`, so a `.razor` file **never reaches `RunCustomExtractors` at all**.

A gate written as `HasSuffix(path, ".razor")` therefore selects nothing, forever, in silence. There is no error, no skip line, and the extractor simply produces nothing — which is indistinguishable from "this repo has no Blazor".

## The live instance

`internal/custom/csharp/blazor_deep.go:179`:

```go
isRazor := strings.HasSuffix(file.Path, ".razor") ||
    strings.HasSuffix(file.Path, ".razor.cs")
```

**Only the first disjunct is dead.** The gate still works via `.razor.cs`, so this is unreachable code rather than a broken extractor — worth stating precisely, because "the `.razor` arm is dead" reads like the extractor is dead, and it is not.

## Why it is worth an issue rather than a comment

**It nearly caused a real defect this week.** #6975 had to gate `blazor.go`, and the obvious move — copy the sibling's `.razor` check — would have made that extractor **entirely dead code** while looking like a correct fix, and would have "resolved" #6973's 6,639 lost edges by silently disabling the producer instead of correcting it. The measurement is what caught it.

The `TSTree != nil` guard is itself correct and load-bearing for a different reason (`index.go:4169` documents it: content-only extractors would otherwise emit entities for files that failed to parse). This issue is not a request to change it.

## What a fix should establish

1. **Make the dead arm impossible to write by accident.** Options, unranked: a comment at the `TSTree`
   guard naming the grammarless extensions it silently excludes; a helper the custom lane calls instead
   of hand-rolled `HasSuffix` gates; or a test asserting that no custom extractor's gate references an
   extension whose language has no grammar. The third is the only one that cannot rot.
2. **Enumerate the affected extensions, not just `.razor`.** Any classified language without a grammar
   has this property. **That list is the deliverable** — I have only established `.razor`, and a fix
   aimed at one extension will leave the rest.
3. **Delete or document `blazor_deep.go:179`'s first disjunct.** It is harmless today; leaving it
   unmarked is how the next person concludes `.razor` gating works.

## Grading

A test enumerating custom-extractor gates against grammarless languages must be shown to **fail** when a
dead gate is introduced — add a synthetic one and confirm RED. A guard over a list that never fires is
the same class of defect as the thing it is guarding against.

**Reachability**: only matters with custom extractors ON, which is not the default (#6966). No user's
graph is affected; the configuration every custom-lane measurement is taken in is.

Refs #6975, #6973, #6966.

cajasmota:
#6978 asked for **the enumeration** — one dead gate is a curiosity; the class is the finding. Derived from the classifier and the dispatch map, **not hand-listed**: the population is **26 unreachable gate literals across 15 files in 8 language packages**, plus 2 extension-shaped literals adjudicated as content matches.

> **Revised after review** ([review](https://github.com/cajasmota/grafel/pull/6995#issuecomment-review), REQUEST-CHANGES). The first revision said 15. It was **under-inclusive in two independent ways**, both real, and its two new comments asserted something the suite already contradicted. All of it is fixed below, and **every number in this body was re-measured against the current scanner, not carried over.**

Measured in `.../scratchpad/impl-6978-q9x4v/wt`, macOS/APFS, on `2d6c82ba8`.

## The criterion is LANGUAGE ROUTING, not `TSTree`

The issue's chain routes through `TSTree == nil`. That is true but **not the load-bearing step** — the reviewer verified all three call sites independently and reached the same conclusion:

- `internal/daemon/extract/subproc.go:364` calls `RunCustomExtractors` with **no `TSTree != nil` guard** (the per-file loop simply continues when `parser.Parse` errors). Only the in-proc (`cmd/grafel/index.go:4179`) and incremental (`incremental.go:1126`) paths carry it.
- Conversely `.sql` **has** a grammar, so `TSTree` is non-nil, and three `.sql` gates are dead anyway.

The step that actually decides it: `RunCustomExtractors` is the **only** dispatcher that selects a prefixed custom key (ordinary dispatch is an exact-key `Get(file.Language)` at `registry.go:95`, which cannot match `custom_csharp_blazor`), and it selects **purely on `file.Language`** via `CustomExtractorsFor`. So:

> A suffix gate in package `P` is unreachable **iff** the classifier's language for that suffix does not route to a prefix any of `P`'s registered keys carries.

Path-independent, repo-independent, **constructive**. **No corpus number appears in this PR.** The one zero I state is labelled as corpus-relative.

## What the review changed

| | was | now |
|---|---|---|
| **H1** — `extractor.Register` matched the **identifier** `extractor` | `internal/custom/javascript` (34 registrations) and `internal/custom/java` (7) import it as `extreg` ⇒ zero keys ⇒ `gateIsReachable` returned `true` for an empty key set. **Two whole packages silently exempt** | the scan resolves the import **alias** per file, and a scanned package yielding **zero keys is a FAILURE**, not an excused skip. Every `internal/custom/<lang>` package registers something, so zero is only ever a scanner bug |
| **H2** — `pathShaped()` was a marker list of subject NAMES | dropped every gate whose subject was `low` / `lower` / `fp`; `php/di_graph.go` was invisible **inside a package the scan covered** | the subject filter is **deleted, not extended**. It was the same hand-picked-enumeration failure the derivation exists to avoid |
| **Polarity** | a candidate the filter dropped disappeared silently | **fail-closed**: every extension-shaped literal is a candidate, and every unreachable one must be adjudicated by hand into `knownUnreachableGates` **or** `knownContentMatchLiterals`. Unlisted ⇒ FAIL |
| **Shapes seen** | `strings.HasSuffix`/`EqualFold` only | + equality comparisons (which is how `php/test_data.go`'s `.feature` slice compare surfaces), + `switch` cases, + an identifier bound exactly once to a literal in the same file |
| **Population** | 15 | **26 gates + 2 content matches** |
| **blazor comments** | "routing `razor` to `custom_csharp_` is what would make these rules run" | **false, and contradicted `blazor.go`'s own comment 20 lines above.** Corrected — see below |
| **"The near-miss landed halfway"** | a heading | **dropped.** See "blazor is latent capability" |
| **Windows** | `test (windows-latest)` RED on `acc936aa3` | fixed and pinned — see below |

## The Windows failure, and why it belongs in this PR

`test (windows-latest)` was red on the first revision, and the failure was the **guard**, not the code it guards. `filepath.Walk` yields `..\custom\csharp\blazor.go` on Windows; the prefix strip was written in forward slashes, so it was a **silent no-op**, every key kept a `../custom/` prefix the hand-written lists do not have, and **both directions of the set comparison failed at once** — every gate reported as new, all four live-gate controls reported as missing.

Two things beyond making it pass:

1. **The normalisation is asserted, not just the outcome.** `gateKey` is a pure function; it reports a key that still contains a backslash, and a strip that did nothing. A key that merely lines up on the author's platform is one refactor from reverting.
2. **It is pinned from any platform.** `filepath.ToSlash` is a **no-op on Unix**, so a test built on it could not have caught this from the macOS or Linux legs — which is exactly how the bug reached CI. `gateKey` replaces backslashes unconditionally, and `TestGateKeyNormalisesWindowsPaths6978` feeds it a literal `..\custom\csharp\blazor.go`. **Scored: reverting `gateKey` to the pre-fix `ToSlash`-then-strip makes that test RED on macOS** (W1 below).

Worth recording rather than fixing quietly: this PR is *about* gates that select nothing because of a path-shaped assumption, and it shipped one. That is evidence the class is easy to fall into, not an aside. No `ci:full` was added — the Windows leg runs by default and caught it.

## The enumeration — 26 sites

| site | literal | whole gate or one disjunct? | why unreachable | anything else break if removed? |
|---|---|---|---|---|
| `cpp/ros_extractor.go` `Extract` | `.xml` | **whole alternative** (`isXML`); `isCPP` live | `.xml` classifies as **nothing**; `file.Language == "xml"` beside it names a language the classifier never emits | no — ROS `package.xml` manifests are simply not parsed |
| `cpp/unreal_extractor.go` `Extract` | `.Build.cs` | **whole alternative** (`isCSharp`); `isCPP` live | `.cs` → `csharp` → `custom_csharp_`, not `custom_cpp_` | no |
| `csharp/blazor.go` `Extract` | `.razor` | **one disjunct** — but it gates the entire markup rule set | `.razor` → language `razor`, no custom prefix | no; `.razor.cs` keeps the two code rules alive |
| `csharp/blazor_deep.go` `Extract` | `.razor` | **one disjunct** | same | no |
| `javascript/vue.go` `isVueFile` | `.vue` | **one disjunct** | `.vue` → language `vue`, no custom prefix | no — `lang != "typescript" && lang != "javascript" && !vueFile` keeps the ts/js arm live |
| `javascript/svelte.go` `Extract` | `.svelte` | **one disjunct** | `.svelte` → `svelte`, no prefix | no — identical shape to vue.go |
| `javascript/nuxt.go` `Extract` | `.vue`, `.server.vue`, `.client.vue` | **three disjuncts** | same | no — `.server.ts`/`.client.ts`/`.js` arms are live, and `.vue` only feeds `isPagesFile` beside a live `isServerAPI` |
| `java/play_routes.go` `isPlayRoutesFile` | `.routes` | **one disjunct** | `.routes` → `scala` → `custom_scala_`; this is `custom_java_patterns` | no — falls through to a live source-marker path |
| `java/struts_routes.go` `ExtractStruts`, `extractStrutsRequestValidation` | `.xml` ×2 | **one disjunct / one branch** | `.xml` classifies as nothing; and the adapter stamps `ctx.Language` `"java"` literally, so `ctx.Language != "java"` rejects it a second time | no |
| `kotlin/orm_query.go` `Extract` | `.sq`, `.sqm` | **one disjunct** (`isSq`) | SQLDelight extensions classify as **nothing** | no; the `language == "kotlin"` arm is live |
| `kotlin/orm_schema.go` `isSQLDelightFile` | `.sq`, `.sqm` | **first branch** | same | no; the kotlin-import branch is live |
| `lua/middleware.go`, `lua/routing.go` `Extract` | `.moon`, `.conf` ×2 each | **two disjuncts each** | `.moon` classifies as nothing; `.conf` too, except basename `nginx.conf` → language `nginx`, also without a prefix | no — and `strings.Contains(ext, "nginx")` beside them is **live**: it fires for a `.lua` file under `nginx/` |
| `nim/ormin_orm.go` `Extract` | `.sql` | **whole branch** | `.sql` → `sql` → `custom_js_` **only** | removing it deletes ormin's SQL-DSL half outright |
| `rust/diesel.go` `Extract` | `.sql` | **whole branch** | same | removing it deletes `extractSQLMigration` outright |
| `rust/sqlx_rbatis.go` `isSqlxMigrationFile` | `.sql` | **whole predicate** — always `false` in production | same | same |
| `php/di_graph.go` `Extract` | `.yaml`, `.yml` | **whole branch** | `.yaml` → `yaml`, no custom prefix; `di_graph.go` has **no language guard**, so routing alone decides it | removing it deletes the Symfony `services.yaml` DI alias→implementation `BINDS` edges outright |
| `php/test_data.go` `Extract` | `.feature` | **whole predicate** | `.feature` classifies as nothing | removing it deletes Behat feature-file handling |

Adjudicated as **content matches**, not gates: `javascript/inngest.go`'s `.inngest` and `.step`, both matched against a `receiver` identifier parsed out of the source. Extension-shaped by coincidence.

**The distinction the issue asked for**, and I checked it at each site rather than inheriting it — the review's summary called `vue.go`/`svelte.go` whole gates; they are disjuncts beside a live ts/js arm, and the difference is exactly the bug-vs-tidy-up line:

- **19 are unreachable code beside a working path** — a tidy-up. What is unreachable in the Vue/Svelte/Nuxt case is SFC-specific extraction (the `<template>` half), not the extractors.
- **5 are latent capability** — diesel/sqlx migration parsing, ormin's SQL DSL, php's Symfony DI graph, php's Behat handling: written, unit-tested, and never run on a real index. Each of the five is exercised **only** by a direct-call unit test, which is what creates the impression of coverage.

A blanket "delete unreachable gates" rule would destroy the second group. **This PR deletes nothing.**

**One zero, labelled.** `.razor.cs` is claimed live by routing (`.cs` → `csharp` → `custom_csharp_`), not by a corpus hit — #6975 measured **zero** `.razor.cs` files across `archigraph-corpora`. That is a corpus-relative zero and says nothing about reachability.

## blazor is latent capability, not a regression #6977 introduced

The first revision's heading "the near-miss landed halfway" is **dropped**. Pre-#6977 `blazor.go` had no file gate, so its markup rules ran over every `.cs` file; #6977 measured the five `@`-directive rules at **zero** matches in `.cs`, and the component-tag rule's `.cs` output as false positives. No `.razor` file has ever reached either extractor on any path. So **nothing that ever produced a true positive was lost** — #6977 removed measured false positives and left a capability that could not fire before it either. The "avoided" record stands.

## What would have to become true — all three conditions, not one

Both new comments originally said routing would be enough. **It would not**, and `blazor.go`'s own pre-existing comment ~20 lines above already said so — the new comment contradicted it in the same function, which is this PR's own defect class written into its fix. Corrected at both sites and here:

- **`.razor`** needs **three** changes, in order: (1) an entry routing `"razor"` to `custom_csharp_` in `customPrefixForLanguage`; (2) the `file.Language != "csharp"` guard widened — it rejects the file a second time and is **deliberately pinned** by `TestBlazorLanguageGuardRejectsNonCSharp6975`; (3) a razor tree-sitter grammar, for the two dispatch paths that also require `TSTree != nil`. Routing alone would light up nothing.
- **`.sql` → rust/nim** needs **prefix routing only**: `sql` has a grammar and those extractors have no language guard on that branch. Cheapest of the set, and the one with real output behind it.
- **`.vue` / `.svelte`** need a prefix entry for `vue`/`svelte`, plus the ts/js language checks widened at each site — same two-part shape as razor, minus the grammar (they would still need one for the two `TSTree` paths).
- **`.yaml` → php** needs prefix routing only — `di_graph.go` has no language guard at all.
- **`.sq` / `.sqm` / `.moon` / `.xml` / `.feature`** need a **classifier** entry first; they name no language today.

Follow-ups are filed rather than bundled: #6996 (`.sql` routing) and #6997 (the three dispatch sites disagree on the `TSTree` guard). Routing changes graph output for whole ecosystems and is a measured decision, not a side effect of an enumeration PR.

## The guard

`internal/extractors/custom_gate_reachability_6978_test.go`. AST-scans `internal/custom/**`, resolves each file's registry import **alias** to read that package's `Register` keys, and asks the **real dispatch predicate** — `classifier.LanguageForExtension` + `extractors.CustomExtractorsFor` — whether any file matching each literal could reach any of them.

Both lists are **hand-written literals** keyed `pkg/file.go|func|literal`, per #6975: not derived from the scan they grade, no line numbers (they rot), set equality **in both directions**, so repairing a site requires deleting its line.

**Two design choices worth stating, because the cheaper version of each was available and wrong:**

- **The `pathShaped` name allowlist was deleted, not extended.** Adding `low`/`lower`/`fp` to it would have made the reviewer's M2 pass while leaving the mechanism — a hand-picked list of subject names, silently dropping whatever it did not anticipate — which is the exact failure the derivation exists to avoid. Deleting it made the scan fail-closed instead.
- **A zero-key package is a FAILURE, not an excused skip.** That branch is what hid `javascript` and `java`: an empty key set read as "out of scope" and returned reachable. Turning it into an assertion closed the hole at the precise point that concealed it — and it is the check most likely to be silently inert, so it is scored (**N5**, 21 keyless-package failures; and the reviewer's K1 independently confirmed it fires and names the package).

### The adjudication is checked against a property, not taken on trust

The reviewer's **CM-1** moved a genuine dead gate — `csharp/blazor.go|Extract|.razor` — out of `knownUnreachableGates` and into `knownContentMatchLiterals`, and the test **passed**. `adjudicated[g] != ""` is satisfied identically by either list; the duplicate check catches an entry in *both*, the reverse-direction check catches one in *neither*, and nothing checked an entry was in the **right** one.

It is reachable in the only sense that matters for a guard: **the failure message invites the edit** — it names both lists and gives no way to choose — and someone facing a red build picks one. The misfile is not cosmetic: `gate` means *a defect or latent capability someone should look at*, `content` means *a false positive to ignore*, so filing a gate as content does not mislabel a finding, it **retires it, permanently and silently**. That is this PR's own failure mode, reproduced inside its guard.

The scan already knows the subject expression — that is how the two content entries were identified as matching against a parsed `receiver` rather than a path — so the difference is now **asserted**: a gate's literal must be compared against a path-derived expression; a content match's must not.

Path taint is a **may-analysis** over single-file bindings, plus parameters that receive a path-derived argument at a same-file call site (`isVueFile`, `isPlayRoutesFile`, `isSqlxMigrationFile` — every helper this matters for is called from its own file). **Its polarity is the point:** taint that fails to follow a subject makes the **gate** list fail loudly ("listed as a gate but its subject is not path-derived"); it cannot become a silent pass. Requiring a *single* binding first reported two genuine gates as content matches, because `fp` is bound in several functions of one file — literal resolution wants MUST, taint wants MAY, and using one polarity for both was wrong.

### What it cannot see — stated because a guard presented as complete is a trap

Each was **scored ALIVE**, not assumed. The scan reaches a literal in `strings.HasSuffix`/`EqualFold`, an equality comparison, or a `switch` case, and resolves an identifier bound **exactly once** to a literal **in the same file**. It does **not** see: a suffix built by concatenation or returned by a function; an identifier bound in another file or imported; an identifier assigned more than once (dropped, not guessed); `strings.Contains`/`HasPrefix`/a regexp/a path-**segment** test — the lua `Contains(ext, "nginx")` disjunct is exactly this shape, and it happens to be live; or a gate in a helper package outside `internal/custom/**`.

What it does guarantee: the shapes that produced all 26 instances cannot grow a 27th without a failure, and **no package is exempt from being asked**.

### Mutants — 13 scored, 13 DEAD

Every mutant the reviewer scored ALIVE — M1/M2/M3 in round 2, CM-1 in round 3 — was re-scored against the current revision, plus controls for each failure mode the fixes introduced.

| | mutant | verdict |
|---|---|---|
| **M1** | dead `.razor` gate in `javascript/vue.go` `isVueFile` — the `extreg` alias package (was **ALIVE**) | **DEAD** — names the exact site |
| **M2** | dead `.razor` gate in `php/laravel.go` with subject `low := strings.ToLower(file.Path)` (was **ALIVE**) | **DEAD** |
| **M3** | `const razorSuffix = ".razor"` + `HasSuffix(file.Path, razorSuffix)` in `kotlin/orm_query.go` (was **ALIVE**) | **DEAD** — same-file constant resolution added; it was cheaper to close than to document |
| **N1** | add a NEW dead gate (`.templ` in `custom/golang/nethttp.go`) | **DEAD** |
| **N1b** | add a **live** gate (`.go`, same site, same shape) | **PASSES** — the guard discriminates on reachability, not on "any new `HasSuffix`" |
| **N2b** | repoint `customGateRoot` at `../extractors` — 311 non-test files, parses fine, clears the file floor | **DEAD** via 4 live-gate controls + 28 reverse-direction failures. This is the "reads the wrong file" no-op a count floor does not catch |
| **N3** | repair a listed site (delete `blazor_deep`'s `.razor` disjunct) without delisting | **DEAD** — reverse direction |
| **N4** | `gateIsReachable` always returns `true` (match, then do not act) | **DEAD** |
| **N5** | point `registryImportPath` at a non-existent package — simulating exactly the H1 alias-resolution failure | **DEAD**, 21 keyless-package failures. Under the old code this was **silent** |
| **W1** | revert `gateKey` to the pre-fix `filepath.ToSlash`-then-strip | **DEAD on macOS** — the Windows bug is now catchable from a Unix leg |
| **CM-1** | move a real gate (`csharp/blazor.go` `.razor`) into `knownContentMatchLiterals` (was **ALIVE**) | **DEAD** — fails on the subject being path-derived |
| **CM-2** | the opposite misfile: a content match (`inngest` `.step`) listed as a gate | **DEAD** |
| **CM-3** | pin every subject verdict to `false` | **DEAD** on all **26** gates — the gate branch is not inert |
| **CM-4** | pin every subject verdict to `true` | **DEAD** on both content entries — the content branch is not inert |

Each edit proven to land with an exact occurrence count (`assert s.count(old)==1` before replacing, then `grep -c` on the result) and the enclosing function named; `go vet` on the touched package before scoring; every restore `cp` from a shasum-verified backup, **never `git checkout`**, with `git status --porcelain` checked afterwards.

**One restore trap worth recording:** restoring `blazor_deep.go` from a backup taken *before* the round-1 comment landed silently reverted that committed comment. `shasum` matched the backup and said nothing — `git diff` against HEAD is what caught it, which is why the check is against HEAD and not against the backup.

## Verification

- `go test -count=1` green: `internal/extractors` (37.1s), `internal/custom/csharp`, `internal/custom/javascript`, `internal/custom/java`, `internal/custom/php`.
- **`./cmd/grafel/`** with a **fresh isolated** `GRAFEL_HOME` + `GRAFEL_DAEMON_ROOT` and `GOMAXPROCS=3`: **ok, exit 0, 175.8s**. Run after the last non-test source edit; every later change is inside `internal/extractors`' `_test.go`, which `cmd/grafel` does not compile.
- **Golden ratchet** (`scripts/quality/run.sh --ratchet`, its own fresh isolated home, `GOMAXPROCS=3`): **38 gated fixtures held their baseline, 29 at 100% must-have recall**, exit 0. **No baseline was regenerated** — the only non-test edits in this PR are comments, so a moved baseline would itself have been the bug.
- `gofmt -l` clean, `go vet` clean. Round 3 touches one `_test.go` in `internal/extractors` and nothing else (`git diff HEAD` against the round-2 commit), so the `cmd/grafel` and ratchet runs above still hold.
- No corpus index was run: another agent was indexing django on the same box, and the question is answerable by construction.

Closes #6978
Refs #6975, #6977, #6973, #6966, #6996, #6997

### DIFF
--- a/internal/custom/csharp/blazor.go
+++ b/internal/custom/csharp/blazor.go
@@ -260,14 +260,27 @@ func (e *blazorExtractor) Extract(ctx context.Context, file extractor.FileInput)
 	// below are further restricted to `.razor`, and only the two code rules
 	// (methods, [Parameter]) run over code-behind.
 	// #6978 — isRazorMarkup is UNREACHABLE, so the markup rules below have
-	// never run on any file. `.razor` classifies as language "razor", which
-	// has no entry in customPrefixForLanguage, so no custom_csharp_* extractor
-	// is ever dispatched for a `.razor` file (internal/extractors/
-	// custom_dispatch.go). Only the `.razor.cs` code-behind arm is live, and
-	// that is the arm #6975 measured. Kept and listed in
-	// TestCustomExtractorGatesAreReachable6978 rather than deleted: routing
-	// "razor" to custom_csharp_ is what would make these rules run, and this
-	// is where that change lands.
+	// never run on any file, and could not have before #6975 either. This
+	// EXTENDS point 2 above rather than restating it: there is a THIRD gate,
+	// earlier than both of the two named there. Dispatch itself is keyed on
+	// language — RunCustomExtractors selects via CustomExtractorsFor, and
+	// "razor" has no entry in customPrefixForLanguage
+	// (internal/extractors/custom_dispatch.go) — so a `.razor` file is never
+	// handed to a custom_csharp_* extractor in the first place, on ANY of the
+	// three dispatch paths.
+	//
+	// SO ROUTING ALONE WOULD NOT MAKE THESE RULES RUN. Point 2 above is right
+	// and this comment does not contradict it: reaching the markup rules needs
+	// all three, in order — (1) an entry routing "razor" to `custom_csharp_`;
+	// (2) the `file.Language != "csharp"` check above widened, which
+	// TestBlazorLanguageGuardRejectsNonCSharp6975 pins deliberately; and (3) a
+	// razor tree-sitter grammar, for the two dispatch paths that also require
+	// file.TSTree != nil (cmd/grafel/index.go, incremental.go — the daemon
+	// subprocess path has no such guard).
+	//
+	// Kept, not deleted, and listed in
+	// TestCustomExtractorGatesAreReachable6978 so it cannot be mistaken for a
+	// working `.razor` gate; that test fails if a NEW unreachable gate appears.
 	isRazorMarkup := strings.HasSuffix(file.Path, ".razor")
 	isRazorCodeBehind := strings.HasSuffix(file.Path, ".razor.cs")
 	if !isRazorMarkup && !isRazorCodeBehind {
--- a/internal/custom/csharp/blazor.go
+++ b/internal/custom/csharp/blazor.go
@@ -259,6 +259,15 @@ func (e *blazorExtractor) Extract(ctx context.Context, file extractor.FileInput)
 	// generic type argument, not a component reference. So the markup rules
 	// below are further restricted to `.razor`, and only the two code rules
 	// (methods, [Parameter]) run over code-behind.
+	// #6978 — isRazorMarkup is UNREACHABLE, so the markup rules below have
+	// never run on any file. `.razor` classifies as language "razor", which
+	// has no entry in customPrefixForLanguage, so no custom_csharp_* extractor
+	// is ever dispatched for a `.razor` file (internal/extractors/
+	// custom_dispatch.go). Only the `.razor.cs` code-behind arm is live, and
+	// that is the arm #6975 measured. Kept and listed in
+	// TestCustomExtractorGatesAreReachable6978 rather than deleted: routing
+	// "razor" to custom_csharp_ is what would make these rules run, and this
+	// is where that change lands.
 	isRazorMarkup := strings.HasSuffix(file.Path, ".razor")
 	isRazorCodeBehind := strings.HasSuffix(file.Path, ".razor.cs")
 	if !isRazorMarkup && !isRazorCodeBehind {
--- a/internal/custom/csharp/blazor_deep.go
+++ b/internal/custom/csharp/blazor_deep.go
@@ -176,6 +176,19 @@ func (e *blazorDeepExtractor) Extract(ctx context.Context, file extractor.FileIn
 	}
 
 	src := string(file.Content)
+	// #6978 — THE `.razor` DISJUNCT IS UNREACHABLE, and kept deliberately.
+	// `.razor` classifies as language "razor" (internal/classifier), and
+	// "razor" has no entry in customPrefixForLanguage
+	// (internal/extractors/custom_dispatch.go), so CustomExtractorsFor never
+	// returns a custom_csharp_* extractor for such a file. No `.razor` file
+	// reaches this function on ANY dispatch path — that is a routing fact, not
+	// a corpus observation. (A razor grammar would also be needed for the two
+	// paths that additionally require file.TSTree != nil.)
+	//
+	// It is NOT deleted: the suffix pair names the Blazor file family, and the
+	// day razor is routed here this is the line that must already be right.
+	// TestCustomExtractorGatesAreReachable6978 lists it, so it cannot be
+	// mistaken for a working `.razor` gate — and it fails if a NEW one appears.
 	isRazor := strings.HasSuffix(file.Path, ".razor") ||
 		strings.HasSuffix(file.Path, ".razor.cs")
 	var entities []types.EntityRecord
--- a/internal/custom/csharp/blazor_deep.go
+++ b/internal/custom/csharp/blazor_deep.go
@@ -177,18 +177,28 @@ func (e *blazorDeepExtractor) Extract(ctx context.Context, file extractor.FileIn
 
 	src := string(file.Content)
 	// #6978 — THE `.razor` DISJUNCT IS UNREACHABLE, and kept deliberately.
-	// `.razor` classifies as language "razor" (internal/classifier), and
-	// "razor" has no entry in customPrefixForLanguage
-	// (internal/extractors/custom_dispatch.go), so CustomExtractorsFor never
-	// returns a custom_csharp_* extractor for such a file. No `.razor` file
-	// reaches this function on ANY dispatch path — that is a routing fact, not
-	// a corpus observation. (A razor grammar would also be needed for the two
-	// paths that additionally require file.TSTree != nil.)
+	// Three independent gates stop a `.razor` file, and ALL THREE would have
+	// to change for this arm to matter:
 	//
-	// It is NOT deleted: the suffix pair names the Blazor file family, and the
-	// day razor is routed here this is the line that must already be right.
-	// TestCustomExtractorGatesAreReachable6978 lists it, so it cannot be
-	// mistaken for a working `.razor` gate — and it fails if a NEW one appears.
+	//  1. Dispatch is keyed on language. RunCustomExtractors selects via
+	//     CustomExtractorsFor, and "razor" has no entry in
+	//     customPrefixForLanguage (internal/extractors/custom_dispatch.go), so
+	//     no custom_csharp_* extractor is offered the file at all — on any of
+	//     the three dispatch paths.
+	//  2. The `file.Language != "csharp"` check 5 lines above rejects it a
+	//     second time. That check is live and load-bearing, pinned by
+	//     TestBlazorLanguageGuardRejectsNonCSharp6975, so ROUTING ALONE IS NOT
+	//     ENOUGH — the guard would have to be widened too.
+	//  3. There is no razor tree-sitter grammar, so file.TSTree is nil and the
+	//     in-proc (cmd/grafel/index.go) and incremental (incremental.go) paths
+	//     skip custom extractors entirely. The daemon subprocess path
+	//     (subproc.go) carries no such guard, so this one is path-specific.
+	//
+	// None of that is a corpus observation; it is how routing is wired. The
+	// arm is NOT deleted because the suffix pair is what names the Blazor file
+	// family (blazor.go uses the same pair), and
+	// TestCustomExtractorGatesAreReachable6978 lists it so it cannot be
+	// mistaken for a working `.razor` gate.
 	isRazor := strings.HasSuffix(file.Path, ".razor") ||
 		strings.HasSuffix(file.Path, ".razor.cs")
 	var entities []types.EntityRecord
--- a/internal/extractors/custom_gate_reachability_6978_test.go
+++ b/internal/extractors/custom_gate_reachability_6978_test.go
@@ -29,15 +29,27 @@ package extractors_test
 // WHY A GUARD RATHER THAN A FIX. The sites below are NOT one defect. Some are
 // unreachable code beside a live path (blazor_deep's `.razor` arm); some are a
 // whole capability that has never run (rust's diesel/sqlx `.sql` migration
-// parsing). Deleting them uniformly would throw away the second kind. This
-// test freezes the population so a NEW one cannot be written in silence.
+// parsing, php's Symfony `services.yaml` DI graph, the whole Vue/Svelte/Nuxt
+// family). Deleting them uniformly would throw away the second kind. This test
+// freezes the population so a NEW one cannot be written in silence.
+//
+// POLARITY: FAIL-CLOSED. The scan applies NO filter to the gate's subject
+// expression. Every extension-shaped literal compared against anything is a
+// candidate, and every candidate the reachability predicate calls unreachable
+// must appear in one of the two hand-written lists below — as a real gate, or
+// as a literal that is matched against file CONTENT rather than a path. An
+// unreachable candidate in neither list FAILS. The first version of this test
+// filtered candidates by a marker list of subject NAMES ("path", "rel", "ext",
+// …) and silently dropped every gate whose subject was called `low`, `lower`
+// or `fp` — that is the hand-picked-enumeration failure this derivation exists
+// to avoid, so the filter is gone rather than extended.
 
 import (
 	"go/ast"
 	"go/parser"
-	"go/printer"
 	"go/token"
 	"os"
+	"path"
 	"path/filepath"
 	"sort"
 	"strconv"
@@ -51,6 +63,14 @@ import (
 // customGateRoot is the tree scanned. Relative to this package's directory,
 // which is where `go test` runs.
 const customGateRoot = "../custom"
+
+// registryImportPath is the package whose Register() calls name a package's
+// dispatch keys. The scan resolves the LOCAL name of this import per file
+// rather than matching the identifier `extractor`: internal/custom/javascript
+// (34 registrations) and internal/custom/java (7) import it as `extreg`, and an
+// identifier-matching scan silently found zero keys for both — which the
+// empty-key branch then read as "out of scope", exempting two whole packages.
+const registryImportPath = "github.com/cajasmota/grafel/internal/extractor"
 
 // knownUnreachableGates is the frozen population of extension gates that
 // cannot fire, as `<pkg>/<file>|<enclosing func>|<literal>`.
@@ -75,11 +95,35 @@ var knownUnreachableGates = []string{
 	"cpp/unreal_extractor.go|Extract|.Build.cs",
 	// `.razor` classifies as "razor" (classifier.go), and "razor" has no
 	// custom prefix. `.razor.cs` classifies as "csharp" and IS live, so both
-	// extractors still run — over code-behind only. In blazor.go this means
-	// the MARKUP rule set, which is further restricted to isRazorMarkup, has
-	// never run on any file.
+	// extractors still run — over code-behind only.
 	"csharp/blazor.go|Extract|.razor",
 	"csharp/blazor_deep.go|Extract|.razor",
+	// Vue SFCs classify as language "vue", which has no custom prefix, so no
+	// custom_js_* extractor can see one. All four are DISJUNCTS, not whole
+	// gates — checked at each site rather than assumed: vue.go and svelte.go
+	// both read `lang != "typescript" && lang != "javascript" && !vueFile`, so
+	// the ts/js arm keeps each extractor live; nuxt's `.server.vue`/
+	// `.client.vue` sit beside live `.server.ts`/`.client.ts`/`.js` arms, and
+	// its `.vue` feeds isPagesFile only, beside a live isServerAPI path. What
+	// is unreachable is SFC-specific extraction (the `<template>` half), not
+	// the extractors.
+	"javascript/vue.go|isVueFile|.vue",
+	"javascript/nuxt.go|Extract|.vue",
+	"javascript/nuxt.go|Extract|.server.vue",
+	"javascript/nuxt.go|Extract|.client.vue",
+	// `.svelte` classifies as "svelte", also without a custom prefix. Same
+	// disjunct shape as vue.go.
+	"javascript/svelte.go|Extract|.svelte",
+	// The java pattern adapter (patterns_dispatch.go) registers
+	// `custom_java_patterns`, so only `.java` files reach these functions —
+	// and the adapter stamps ctx.Language "java" literally, so the
+	// `ctx.Language != "java"` guard rejects anything else a second time.
+	// `.routes` classifies as "scala" (routing to custom_scala_) and `.xml` as
+	// nothing. Both are disjuncts: each falls through to a live source-marker
+	// path when false, so the extractors still work on `.java`.
+	"java/play_routes.go|isPlayRoutesFile|.routes",
+	"java/struts_routes.go|ExtractStruts|.xml",
+	"java/struts_routes.go|extractStrutsRequestValidation|.xml",
 	// SQLDelight `.sq`/`.sqm` are classified as nothing, so no FileInput with
 	// that path ever exists. The kotlin-language arm of both gates is live.
 	"kotlin/orm_query.go|Extract|.sq",
@@ -102,14 +146,43 @@ var knownUnreachableGates = []string{
 	"nim/ormin_orm.go|Extract|.sql",
 	"rust/diesel.go|Extract|.sql",
 	"rust/sqlx_rbatis.go|isSqlxMigrationFile|.sql",
+	// `.yaml`/`.yml` classify as "yaml", which has no custom prefix. di_graph
+	// has NO language guard, so routing alone decides it: the whole Symfony
+	// `services.yaml` DI alias→implementation branch is latent capability,
+	// exercised only by its own unit test. Subject is `lower`, which is why
+	// the first version of this scan could not see it.
+	"php/di_graph.go|Extract|.yaml",
+	"php/di_graph.go|Extract|.yml",
+	// Behat `.feature` files classify as nothing. Written as a slice compare
+	// (`file.Path[len(file.Path)-8:] == ".feature"`) rather than a HasSuffix
+	// call, which is why the scan also inspects equality comparisons.
+	"php/test_data.go|Extract|.feature",
 }
 
-// liveGateControls are gates in the SAME files that the scan must find and
-// must NOT report as dead. This is the positive control for the "read the
-// wrong file / matched nothing" no-op: if the scan is pointed at a tree that
-// parses fine but contains no gates, or its matcher stops recognising gate
-// shapes, these disappear and the test fails on them rather than passing
-// vacuously on an empty dead set.
+// knownContentMatchLiterals are extension-shaped literals that are NOT file
+// gates: they are matched against file CONTENT (a source line, an identifier
+// fragment, a config value), so "unreachable as an extension" says nothing
+// about them. Same id shape, same hand-written discipline, same both-direction
+// set equality — an entry here is an explicit claim that the site is not a
+// path gate, and a reviewer can check it in one line.
+//
+// This list exists because the scan deliberately applies no subject filter: a
+// filter that dropped these would also drop real gates, as the first version
+// of this test did. Adjudicating them by hand is the price of fail-closed.
+var knownContentMatchLiterals = []string{
+	// Both are matched against `receiver` — an identifier parsed out of the
+	// SOURCE (`step.run(...)`, `inngest.createFunction(...)`) — not against a
+	// path. Extension-shaped by coincidence.
+	"javascript/inngest.go|Extract|.inngest",
+	"javascript/inngest.go|inngestStepReceiverAttributed|.step",
+}
+
+// liveGateControls are gates that the scan must find and must NOT report as
+// dead. This is the positive control for the "read the wrong file / matched
+// nothing" no-op: if the scan is pointed at a tree that parses fine but
+// contains no gates, or its matcher stops recognising gate shapes, these
+// disappear and the test fails on them rather than passing vacuously on an
+// empty dead set.
 var liveGateControls = []string{
 	"csharp/blazor.go|Extract|.razor.cs",
 	"csharp/blazor_deep.go|Extract|.razor.cs",
@@ -118,8 +191,8 @@ var liveGateControls = []string{
 }
 
 // minGateLiterals / minFilesParsed are floors on the scan's raw work. They
-// catch only the crudest no-op (read nothing), which is why the two set
-// assertions above carry the real weight.
+// catch only the crudest no-op (read nothing), which is why the set assertions
+// above carry the real weight.
 const (
 	minGateLiterals = 25
 	minFilesParsed  = 300
@@ -133,19 +206,30 @@ type gateSite struct {
 }
 
 // scanCustomGates walks internal/custom/** and returns every extension-shaped
-// literal tested against a path-shaped expression, split into unreachable and
-// reachable, plus the number of non-test files parsed.
-func scanCustomGates(t *testing.T) (dead, live []gateSite, filesParsed int) {
+// literal compared against something, split into unreachable and reachable,
+// plus the number of non-test files parsed and the packages whose Register
+// keys came back empty.
+func scanCustomGates(t *testing.T) (dead, live []gateSite, filesParsed int, keylessDirs []string) {
 	t.Helper()
 
 	keysByDir := map[string][]string{}
+	dirsSeen := map[string]bool{}
 	var all []gateSite
 
 	err := filepath.Walk(customGateRoot, func(p string, info os.FileInfo, err error) error {
 		if err != nil {
 			return err
 		}
-		if info.IsDir() || !strings.HasSuffix(p, ".go") || strings.HasSuffix(p, "_test.go") {
+		if info.IsDir() {
+			// testdata/ holds Go SOURCE FIXTURES, not packages: they register
+			// nothing, and judging them would make every scan report a keyless
+			// package. `go build` ignores them for the same reason.
+			if info.Name() == "testdata" {
+				return filepath.SkipDir
+			}
+			return nil
+		}
+		if !strings.HasSuffix(p, ".go") || strings.HasSuffix(p, "_test.go") {
 			return nil
 		}
 		filesParsed++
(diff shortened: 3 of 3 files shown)


cajasmota:

Measured in `/private/tmp/claude-501/-Users-jorgecajas-Projects-archigraph/c0fefed4-0195-4c81-9a94-07a7145710a4/scratchpad/review-6995-z8k2p/wt` (detached at `acc936aa3`), macOS/APFS. Nothing pushed, nothing merged.

The premise correction is **right** and the reachability model is the correct one. But the population is **not 15**, the guard does **not** fail when a 16th appears — a 16th and a 17th are already in the tree today — and two of the new comments assert something the suite already contradicts.

---

### (a) Language routing, not `TSTree` — the author is RIGHT, and #6978's body is wrong

Verified all three call sites in the worktree:

- `internal/daemon/extract/subproc.go:364` — the per-file loop continues when `parser.Parse` errors (`file.TSTree` simply stays nil, line ~313), and the Pass-2 dispatch is gated on `runExtract` **only**. No `TSTree` guard. A `.razor` file *does* reach `RunCustomExtractors` on this path.
- `cmd/grafel/index.go:4179` and `internal/extractors/incremental.go:1126` — both carry `&& file.TSTree != nil`.
- `RunCustomExtractors` → `CustomExtractorsFor(file.Language)` (`custom_dispatch.go:209`, `:156`) — prefix selection keyed on **language alone**.

So step 4 of #6978's chain is true of one path out of three and is not the criterion. **Please correct the issue body** — the "iff the classifier's language for that suffix does not route to a prefix P's keys carry" formulation is the right one and holds on all three paths.

### (b) The enumeration is INCOMPLETE — two independent holes in the scan, each with real missed sites

I re-derived from scratch: a broader AST walk over `internal/custom/**` taking **every** extension-shaped literal regardless of call shape, per-package `Register` keys with an **alias-agnostic** matcher, and the same real predicates (`classifier.LanguageForExtension` + `extractors.CustomExtractorsFor`). 487 files, 176 distinct sites.

**H1 — the `extractor.Register` matcher is alias-sensitive.** The scan requires the selector's package ident to be literally `extractor`. `internal/custom/javascript` (34 registrations) and `internal/custom/java` (7) import the registry as **`extreg`**. Their `keysByDir` is therefore empty, and `gateIsReachable` returns `true` for an empty key set ("out of this test's scope") — **two whole packages are exempt**. Real dead gates hiding there, by exactly the `.razor` argument (`vue` and `svelte` have no entry in `customPrefixForLanguage`):

- `javascript/vue.go:60` `.vue` — the **whole `isVueFile` predicate**
- `javascript/svelte.go:160` `.svelte` — whole gate
- `javascript/nuxt.go:191, 340, 343` `.vue` / `.server.vue` / `.client.vue`

**H2 — `pathShaped()`'s marker list misses ordinary subject names.** It scans for `Path/path/rel/base/file/name/ext`. It does **not** match `lower`, `low`, `fp`, `p`, `s`:

- `php/di_graph.go:102` — `isYAML := strings.HasSuffix(lower, ".yaml") || strings.HasSuffix(lower, ".yml")`, where `lower = strings.ToLower(path)`. `.yaml`/`.yml` classify as `yaml`, which has **no** custom prefix, so `custom_php_` never sees one. This is a **fourth latent-capability whole branch** — Symfony `services.yaml` DI alias→impl `BINDS` edges — sitting in a package the scan *does* cover, and it is not listed. `di_graph.go` has no language guard, so routing alone decides it; and `di_graph_test.go:103` drives it directly with `config/services.yaml`, which is the same "unit test creates a false impression of coverage" shape you correctly called out for the three `.sql` sites.
- `php/test_data.go:80` — `isFeatureFile := len(file.Path) > 8 && file.Path[len(file.Path)-8:] == ".feature"`. Not a `HasSuffix` call at all, so invisible to the matcher; `.feature` classifies as nothing → dead (Behat feature files).

So: the population is at least **19-21**, not 15, and there are at least **four** latent-capability sites, not three. The title's "a guard that fails when a 16th appears" is not what the guard does.

### (c) blazor.go — **latent capability, NOT a regression #6977 introduced**. But two new comments are wrong.

Checked `git show 448d7bb38 -- internal/custom/csharp/blazor.go` against the current file:

- Pre-#6977 `blazor.go` had no file gate, so the markup rules ran over every `.cs` file. #6977 measured the five `@`-directive rules at **zero** matches in `.cs` (they fire only in `.razor`), and the component-tag rule's `.cs` output as false positives — the one fixture that depended on a collider entity was repaired by declaring the name, not by a baseline bump.
- No `.razor` file has ever reached either extractor on any path.

⇒ **nothing that ever produced a true positive was lost.** #6977 removed measured false positives and left a capability that could not fire before it either. The near-miss recorded as *avoided* stands. Please **drop the heading "The near-miss landed halfway"** — it contradicts the precise sentence in your own paragraph and is the line a reader will quote.

**The substantive defect is elsewhere, and it is the dominant class on this board.** Both new comments say routing is sufficient:

> blazor.go: "routing `razor` to `custom_csharp_` is what would make these rules run"
> blazor_deep.go: "the day razor is routed here this is the line that must already be right"

It is **not**. Both extractors early-return on `file.Language != "csharp"` (`blazor_deep.go:174`; the same check above `blazor.go`'s gate), pinned by `TestBlazorLanguageGuardRejectsNonCSharp6975`. A `.razor` file carries language `"razor"` and would be rejected after routing. `blazor.go`'s **pre-existing comment ~20 lines above the new one says exactly this** ("registering a grammar would not help, because the language check would still reject the file") — the new comment contradicts it in the same function. The PR body's "What would have to become true" bullet has the same omission. It needs all three conditions: prefix routing **+ widening the language guard** + a grammar for the two `TSTree` paths.

### (d) Vacuity / over-breadth

**Not over-broad** — `gateIsReachable` calls the real dispatch predicate, and N1b is the right control. **Under-inclusive**, and a new dead gate is easy to write without the guard noticing. Three mutants of my own, none of them a re-run of N1–N4:

| | mutant | verdict | why |
|---|---|---|---|
| **M1** | add a dead `.razor` gate to `javascript/vue.go` `isVueFile`, subject `filepath.ToSlash(fp)` (path-shaped, `HasSuffix`, extension-shaped literal) | **ALIVE** (guard exit 0) | H1 — `extreg` alias ⇒ `dirKeys` empty ⇒ treated as reachable |
| **M2** | add a dead `.razor` gate to `php/laravel.go` `Extract` with subject `low := strings.ToLower(file.Path)`, in a package the scan *does* cover | **ALIVE** (exit 0) | H2 — `pathShaped("low")` is false |
| **M3** | add `const razorSuffix = ".razor"` + `strings.HasSuffix(file.Path, razorSuffix)` to `kotlin/orm_query.go` `Extract` | **ALIVE** (exit 0) | argument is an `*ast.Ident`, not a `*ast.BasicLit` |

Each edit proved to land: `assert s.count(old)==1` before replacing, `grep -c` on the result, enclosing function printed, `go vet` clean on the touched package. Each restored by `cp` from a shasum-verified backup (hashes re-verified after restore), never `git checkout`; `git status --porcelain` empty afterwards. Green guard baseline re-established before each score, `-count=1`.

### Cheapest fixes

1. Match `Register` on **any** selector package ident, and make an empty key set for a scanned package a **failure**, not a pass — every `internal/custom/<lang>` package registers something, so "this dir yielded 0 keys" is only ever a scanner bug.
2. Add `low`, `lower`, `fp`, `p`, `s` to `pathShaped`, or invert it: take every extension-shaped literal and carry an explicit allowlist for the content-match false positives (`.route`, `.Get`, `.transaction`, …). The allowlist is auditable; the marker list silently drops sites.
3. Then re-derive `knownUnreachableGates` and re-state the count. Add the newly-surfaced latent-capability sites (`php/di_graph.go` yaml, `javascript/vue.go`/`svelte.go`/`nuxt.go`) to the "what would have to become true" section.
4. Fix the two blazor comments and the PR-body bullet to name the language guard.

### Verification I ran

Tree above; `go test -count=1` **green, exit 0** on `./internal/extractors/... ./internal/custom/...` (96 packages ok). `./cmd/grafel/` run with fresh isolated `GRAFEL_HOME`/`GRAFEL_DAEMON_ROOT` and `GOMAXPROCS=3` — **ok, exit 0, 171.8s**. **Confirmed no baseline was regenerated**: the whole diff vs `2d6c82ba8` is three files (`blazor.go`, `blazor_deep.go`, the new test), so the golden ratchet was not re-run — the diff contains no behavioural change to move it.

cajasmota:
## Re-check at `bf6a7a06a` — **APPROVE**

Fresh worktree, detached at `bf6a7a06a`: `/private/tmp/claude-501/-Users-jorgecajas-Projects-archigraph/c0fefed4-0195-4c81-9a94-07a7145710a4/scratchpad/review-6995-r2/wt`, macOS/APFS. Nothing pushed, nothing merged. Every mutant restored by `cp` and then checked with **`git diff HEAD`**, not shasum alone.

### Mutants — my three re-run, three new

| | mutant | verdict |
|---|---|---|
| **R1** | dead `.razor` gate in `javascript/vue.go` `isVueFile` (the `extreg`-alias hole) | **DEAD** — `UNADJUDICATED unreachable literal "javascript/vue.go\|isVueFile\|.razor"`, exit 1 |
| **R2** | dead `.razor` gate in `php/laravel.go` with subject `low` (the `pathShaped` hole) | **DEAD** — names `php/laravel.go\|Extract\|.razor`, exit 1 |
| **R3** | `const razorSuffix = ".razor"` + `HasSuffix(file.Path, razorSuffix)` in `kotlin/orm_query.go` | **DEAD** — names `kotlin/orm_query.go\|Extract\|.razor`, exit 1 |
| **K1** *(new)* | delete `internal/custom/clojure`'s sole `extractor.Register` so the package yields zero keys | **DEAD** — `package ../custom/clojure yielded zero extractor.Register keys …`, plus 4 unadjudicated `.clj*` literals, exit 1. **The fail-closed branch genuinely fires and names the package** — this is the one I was most prepared to find silently inert, and it is not |
| **N1** *(new)* | suffix by concatenation: `strings.HasSuffix(file.Path, "."+"razor")` in `php/laravel.go` | **ALIVE**, exit 0 — matches documented blind spot 1 verbatim |
| **N2** *(new)* | cross-file ident: `const razorSuffixX = ".razor"` in `kotlin/orm_schema.go`, used in `kotlin/orm_query.go` | **ALIVE**, exit 0 — matches documented blind spot 2 verbatim |
| **W1** *(re-verified)* | revert `gateKey` to the forward-slash-only strip | **RED on macOS**: `gateKey("..\\custom\\csharp\\blazor.go") rejected its own input: contains a backslash …`. The Windows pin works from a Unix box, which is the whole point of making it a pure function |

So the documentation of what the scan cannot see **matches its behaviour** — both blind-spot mutants are alive exactly where the comment says they will be, and neither is presented as covered.

### The three things you asked me to attack

1. **Zero-keys = failure.** Verified live (K1). `gateIsReachable` now returns `false` on an empty key set and `keylessDirs` is reported per package. The `testdata` `SkipDir` is the one exemption and it is justified — those files are `package sample`/`package routers` fixtures that register nothing (82 files, none containing a `Register` call), so without the skip every scan would report four phantom keyless packages and the rule would be unusable.
2. **The adjudication list is not a hiding place.** Both `knownContentMatchLiterals` entries check out in one line each: `inngest.go:124` `receiver == "step" || strings.HasSuffix(receiver, ".step")` and `inngest.go:191` `!strings.HasSuffix(receiver, ".inngest")` — `receiver` is a substring cut out of the source by regexp (`src[m[2]:m[3]]`), not a path. Genuine content matches, correctly filed. Misfiling in the other direction is structurally impossible: a live gate listed as unreachable fails the reverse-direction assertion.
3. **You corrected me and you are right — stated plainly.** `vue.go:81` and `svelte.go:164` both read `if lang != "typescript" && lang != "javascript" && !vueFile { return }`, so the ts/js arm keeps each extractor live and the SFC test is a disjunct. `nuxt.go`: `.vue` is a *conjunct* of `isPagesFile` sitting beside a live `isServerAPI` path, and `.server.vue`/`.client.vue` are switch-case disjuncts beside live `.server.ts`/`.server.js`. My "whole dead predicate" framing was wrong at the site level; what is unreachable is SFC-specific extraction, not the extractors. The list comments say exactly that.

Also confirmed: the blazor comments now name all three conditions in order (routing → the `file.Language != "csharp"` guard → the grammar for the two `TSTree` paths), so the contradiction with the pre-existing point 2 above them is gone.

### One non-blocking observation, for a follow-up rather than this PR

`java/play_routes.go|isPlayRoutesFile|.routes` is listed as a disjunct, which is accurate for the literal. But its two siblings are unreachable from `internal/custom/java` for the same reason: `classifier.go:750` maps the **basename** `routes` → `"scala"`, which routes to `custom_scala_`, so `base == "routes"`, `HasPrefix(base, "routes.")` and `Contains(filePath, "conf/routes")` cannot fire either. The predicate as a whole never returns true in production — a sixth latent capability. Neither form is extension-shaped, so this sits squarely in the declared blind-spot class (path-segment / non-extension equality) and is not a defect in the guard; it is worth a line in that entry's comment.

### Verification

`go test -count=1` **green, exit 0**: `./internal/extractors/... ./internal/custom/...` (96 packages ok) and `./cmd/grafel/` with a **fresh** `GRAFEL_HOME`+`GRAFEL_DAEMON_ROOT` and `GOMAXPROCS=3` — **ok, exit 0, 162.9s**. Baseline re-established green before each mutant score; `go vet` clean on every touched package; tree confirmed identical to `bf6a7a06a` afterwards (`git diff HEAD` empty). No baseline regenerated — the diff is still three files.

Both round-1 blockers are closed and the round-2 work found more than I did. **APPROVE.**

cajasmota:
## Coordinator mutant CM-1 on `bf6a7a06a` — ALIVE, production-reachable, ~10 lines to close

Own detached worktree, restored by `cp` from a SHA-256-verified backup (`c94fbde9…` both ways) **and verified with `git diff HEAD`** — clean, per the trap you logged, where a pre-comment backup silently reverted a committed comment and the shasum said nothing. (The 23 `MUTANT` hits under `internal/extractors` are pre-existing in `main` — prior mutants documented in test comments — not residue.) Baseline green immediately before. macOS/APFS.

**CM-1 — silence a real dead gate by MISFILING it.** Twelve mutants have been scored here. They attack the scanner (alias, subject shape, constants, concatenation, cross-file idents), the fail-closed branch (K1 deletes a `Register`), the controls, and the Windows key. **Nobody attacked the adjudication itself** — which is the one thing the design rests on, because the whole guard reduces to "an unreachable literal must appear in one of two lists".

I moved a genuine dead gate from `knownUnreachableGates` into `knownContentMatchLiterals`:

```go
-  // in knownUnreachableGates
-  "csharp/blazor.go|Extract|.razor",
+  // in knownContentMatchLiterals
+  "csharp/blazor.go|Extract|.razor", // MUTANT CM-1: real dead gate MISFILED as a content match
```

**Result: ALIVE** — `TestCustomExtractorGatesAreReachable6978` passes, exit 0.

## Why this one matters more than its size suggests

The adjudication check is `if adjudicated[g] == ""` — **both lists satisfy it identically**. The duplicate check catches an entry in *both* lists, and the reverse-direction check catches an entry in *neither*-detected state. Nothing checks that an entry is in the **right** list.

And this is production-reachable in the only sense that applies to a guard: **the failure message itself invites the edit.** It tells the next person "list it in `knownUnreachableGates` … or list it in `knownContentMatchLiterals`", and offers no way to tell which is correct. Someone under time pressure, facing a red build, picks one. The classification is not decoration — `gate` means *a defect or a latent capability someone should look at*, `content` means *a false positive, ignore it* — so a misfile does not merely mislabel, it **retires a real finding into the ignore bucket, permanently and silently**. That is the exact failure mode this PR exists to prevent, reproduced inside its own guard.

**The scan already has what it needs to close this.** Your two content-match entries were identified because they match against `receiver` — an identifier parsed out of the source — rather than against a path, and the scanner necessarily knows the subject in order to have classified them at all. So assert the property instead of trusting the list: **an entry in `knownContentMatchLiterals` must have a non-path subject, and an entry in `knownUnreachableGates` must have a path subject.** Then a misfile fails on the semantic difference rather than passing on set membership.

Roughly ten lines, and it converts the two lists from *assertions of intent* into *checked classifications*. **I would like this before merge** — a third round is more than I would normally ask, but the guard's whole value is that a dead gate cannot appear without adjudication, and right now adjudication is one honest mistake away from silence. If you think the price is wrong, show me the diff rather than arguing it.

If the subject is genuinely not available at that point, say so and I will take a count-pin instead — pinning each list's length against an independent literal at least forces a misfile to be a deliberate two-number edit.

## On the round

Round 2 is a much stronger PR than round 1, and two choices are why. **Deleting the `pathShaped` name allowlist rather than extending it** refused the hand-picked-enumeration failure that the derivation existed to avoid. And **making a zero-key package a FAILURE rather than an excused skip** closed the hole at exactly the point it hid `javascript` and `java` — K1 confirms that branch fires and names the package, which was the check most likely to be silently inert.

Correcting the reviewer on `vue.go`/`svelte.go` — disjuncts beside a live ts/js arm, so what is unreachable is SFC-specific extraction rather than the extractors — is the right kind of precision, and they accepted it. The `gateKey` pin is the right shape too: a Windows-only bug that can only be caught on Windows is a bug that reaches CI, and W1 being RED on macOS means this one is now catchable from a Unix box.

cajasmota:
## CM-1 replayed on `d2d3b934a` — now DEAD. Approving; merging on green.

Fresh worktree at the round-3 head, green baseline first, restored by `cp` and verified with **`git diff HEAD`** (clean). I re-ran the *same* misfile rather than trusting the description, because the whole finding was that the two lists were interchangeable — a fix for that has to be checked by repeating the exact edit that passed before.

Moving `csharp/blazor.go|Extract|.razor` from `knownUnreachableGates` into `knownContentMatchLiterals` now fails with:

> `"csharp/blazor.go|Extract|.razor" is listed as a CONTENT MATCH but its literal IS compared against a path-derived expression, so it is a file gate. Listing a real gate here retires a live finding into the ignore bucket, which is the failure this whole test exists to prevent.`

That is the right diagnostic: it names the mechanism, not just the mismatch.

**You took the harder half and it was the right call.** I offered a count-pin fallback; you established the subject was available and asserted the property instead. A count-pin would have made a misfile a two-number edit — annoying rather than impossible. The property assertion makes the two lists mean *different things*, so a misfile fails on the semantics.

**CM-3/CM-4 are the part I would have missed.** A two-branch assertion where only one branch ever fires grades neither, so pinning the subject verdict `false` (DEAD on all 26 gates) and `true` (DEAD on both content entries) is what proves both branches are live. That is the mutually-masking-guards lesson applied before anyone had to learn it here.

**And the assertion caught a real error in itself.** Requiring a *single* binding — MUST polarity, copied from the literal-resolution path — misreported `java/play_routes.go` `.routes` and `java/struts_routes.go` `.xml` as content matches, because `fp` is bound in several functions of one file. Literal resolution wants MUST; taint wants MAY. Two genuine gates would have been silently retired into the ignore bucket by the very assertion added to prevent that. It surfaced only because the polarity was chosen so that a taint failure makes the **gate** list fail loudly rather than pass quietly — the design decision doing real work, not the test doing it by luck.

Merging.

PR #7000, opened 2026-09-08
feat(#6912): rust fields carried their declared type only as a property and in the Signature — emit the field→type REFERENCES edge, same-file targets only

cajasmota:
Arm C of #6912 — Rust. Arm A is C# (#6984), arm B protobuf (#6991). Languages remain, so **Refs #6912**, not Closes.

## Arm B's question, answered before anything was designed

Arm B corrected #6912's central claim for the second time: the right question per language is not *"does a field→type edge exist"* but **"is the FIELD record the anchor, or is something else"** — an edge anchored on the parent leaves the field a leaf and leaves #6726's orphan rate exactly where it was.

For Rust the answer is **neither**. Every relationship the package emits was enumerated (`internal/extractors/rust/struct_fields.go:224`, `rust.go:156/208/294/552/918`):

| edge | anchor |
|---|---|
| `CONTAINS` struct/enum → field (#4854) | owner component |
| `CONTAINS` trait/impl → method (#144) | trait / impl |
| `CALLS` (#616) | operation |
| `IMPORTS` | file |
| `EXTENDS` supertrait | trait |

None carries a field's declared type at **any** anchor — unlike protobuf, where `buildMessage` already wired the message. So this arm **adds** the relation rather than duplicating an anchor: there is no existing topology to preserve (#6419/#6422 has no Rust counterpart) and nothing to double-count against. The type was retained only as the `field_type` property and inside the `Signature`.

## Design — inherited, not re-litigated

`REFERENCES` + `ref_kind: "field_target_type"`; empty `FromID` so assembly anchors the edge on the field record (Django precedent); target addressed by `extractor.BuildComponentStructuralRef("rust", <file>, <Name>)`, **not** the custom lane's `Class:<Name>` (#6986: 91.6% dangling). One address dialect, not arm A's two — a Rust `enum` is a `SCOPE.Component`, so it takes the same component address as a struct. The `SCOPE.Enum` value-set nodes from #4431 are a parallel population and are deliberately not targeted; targeting both would put two edges on one field for one declaration.

Emitted **only** where the named type is declared in the same file. Same-file rule, no cross-file tier — the Python arm's cross-file behaviour comes from a tier its resolver already had; Rust has none, so none was invented.

## Rust specifics, each with its own assertion

- **Primitives** — no blocklist, and the reason is the grammar, not our corpus. In tree-sitter-rust a primitive in type position is the node kind `primitive_type`, never `type_identifier`, at every depth (`i32`, `Vec<u8>`, `Box<str>`, `(Order, i32)`, `fn(Order) -> u8`). The collector takes `type_identifier` only, so no primitive is ever a candidate. **Arm B's counter-example is constructed rather than argued away**: the fixture declares `struct <p>;` for all 16 primitive names — proven present in the in-file target set by a positive control — beside a field of each type, and asserts per member that no edge appears. Where proto's missing blocklist **emitted** an edge protoc denies, Rust's absent one can only **miss** one, which never dangles.
- **Generic wrappers** — `Vec<Order>`, `Option<Box<Order>>`, `Arc<Mutex<Customer>>`, `HashMap<String, Customer>`: the constructor and the arguments are both candidates and the in-file check separates them, so the wrapper is never a target. A **user-defined** generic declared in the file (`Holder<Customer>`) legitimately yields two edges, which is why the constructor is not blanket-dropped.
- **References, lifetimes, pointers, arrays, tuples, `dyn`, fn-types** — `&'a Customer`, `&mut Order`, `*const Customer`, `[Customer; 3]`, `(Customer, Customer)`, `Box<dyn Shipper>`, `fn(Customer) -> i32` all reach the type and not the borrow. The lifetime `'a` and an array-length const are `identifier` nodes, not `type_identifier`.
- **Paths** — `crate::models::Order`, `super::Order`, `self::Order`, `<Order as Trait>::Assoc` are `scoped_type_identifier` and are skipped **whole**. Never stripped to the last segment: arm B carries a forbidden row named for exactly that mutation, and so does this arm.
- **Type parameters** — an open `T` is dropped only because nothing is named `T`; a parameter shadowing a same-file type over-fires. **Pinned as known-wrong.**
- **Enums and traits** are legitimate same-file targets; so is a `type` alias.
- **`impl Order` is not a type declaration**, and this one is load-bearing and Rust-specific: `buildImpl` emits a `SCOPE.Component` named after the implementing type, so `struct Order` + `impl Order` — the commonest shape in the language — puts two components called `Order` in one file. Counting the impl as a declaration would trip the duplicate-name rule and silently delete the edges this pass exists to add. Asserted by `TestRustFieldTypeRefs_ImplBlockIsNotATypeDeclaration`, with a positive control that both impl components exist.

A `WHAT THE CHECK IS NOT` header block carries all of this at the code, arm A's shape.

## A guard that was deleted, and the evidence that it was inert rather than merely redundant

**Review settled this with the compiler.** rustc **1.98.1** was installed into scratch and the case enumerated: `struct i32;` beside `pub f: i32` compiles **RC=0 for all 17 primitive names**, and rustc really does bind the field to the shadowing struct — `S { f: i32 }` (the name used as a *value*) compiles only with the shadow present, and fails `error[E0423]` without it. Then 17 primitives × 16 nesting shapes = **272 enumerated cases**, each carrying two positive controls (the shadowing `struct P` asserted present in the in-file target set, **and** the identical shape written with `Customer` required to produce an edge). All 272 pass; no primitive edges at any depth.

So the case arm B taught us to construct **is** constructible in Rust — and that *strengthens* the deletion rather than undermining it, because grafel's failure mode here is a **miss**, never a wrong emission. proto's absent blocklist emitted an edge protoc denies; Rust's absent one can only fail to emit one, and a miss never dangles.

**The harness note is part of the result, not a footnote.** The reviewer's first rustc pass returned RC=1 for all 17 names — which would have supported the *opposite* finding, that the case cannot be written at all. It was a broken harness (`-o /dev/null` failing to make a temp dir), and it was caught by the **positive control**, not by inspection. A compiler-verified claim is only as trustworthy as the control that proves the compiler was actually invoked; without it this PR would have carried a confident, backwards conclusion.

The first cut also carried `primitive_type`, `lifetime` and `scoped_identifier` skip cases. A mutant removing `primitive_type` came back **ALIVE**, and arm B's lesson says an alive-or-dead unit result settles nothing on its own — so the mechanism was checked instead: all three node kinds are leaf tokens or carry only `identifier` children, so the collector returns nothing for them **whether the case is present or not**. That is a stronger claim than arm B's deleted blocklist could make (it does not rest on "our corpus has no such file"), and it was checked in both directions anyway: deletion leaves every fixture green **and** leaves the corpus edge counts below byte-identical over 789 real `.rs` files. The one **live** case, `scoped_type_identifier`, is kept and graded.

## Verification

`go test -count=1` green on `./internal/extractors/rust/...`, `./internal/custom/rust/...` and **`./cmd/grafel/`** (fresh isolated `GRAFEL_HOME` + `GRAFEL_DAEMON_ROOT`, 148s), `gofmt`/`go vet` clean. Golden ratchet green: **38 gated fixtures held**.

**The golden baseline moved by exactly +1**, enumerated: `rust-tokio-mini` `relationship_extracted_total` 63 → 64, which is `MemoryStore.users: HashMap<String, User>` → `User` in `src/store.rs`. Hand-edited as ONE number and **proven identical to a regeneration**: `--update-baseline` was run at HEAD *and* at the untouched parent (`2d6c82ba8`) and the two generated baselines diffed — the entire difference is that single line. Committing a blanket regeneration would have **lowered** `*_extracted_total` ceilings on six other fixtures (this machine's detector env differs from the recorded one), and `--ratchet` does not fail on a ceiling drop, so it would have been a silent weakening (#6991).

### Measured, gate-independent, macOS/APFS (pass-1 extractor only; no detector output is read)

| corpus | .rs files | field entities | fields gaining an edge | edges |
|---|---|---|---|---|
| tokio | 373 | 972 | 212 (21.8%) | 213 |
| just | 236 | 760 | 29 (3.8%) | 29 |
| actix-examples | 152 | 234 | 16 (6.8%) | 16 |
| mini-redis | 28 | 58 | 7 (12.1%) | 7 |
| **total** | **789** | **2,024** | **264 (13.0%)** | **265** |

Every emitted edge's target is declared in the same file **by construction**, so that check is a re-derivation of the gate rather than an independent confirmation of it; the independent evidence is `TestRustFieldTypeRefs_ResolverBindsEveryEdge`, which drives every edge through the production resolver (`BuildIndex` → `ReferencesEmbedded`) and asserts each ToID comes back rewritten to a real entity ID, with a positive control that the bare-name form is AMBIGUOUS in the same fixture.

**#6726's 99.3% is their corpus and Rust is not in it — no claim is made that it moves.** The `just` figure (3.8%) is the honest floor of a same-file rule on a codebase whose types live one per module file.

One incidental finding, recorded in the resolver test rather than in prose: a struct and its `impl` block are **one graph node**. `graph.EntityID` (`internal/graph/graph.go:259`) hashes `repo ∥ Kind ∥ Name ∥ SourceFile` — **`Subtype` is not in the hash** — and both records are `SCOPE.Component`/`Order`/same file, so there is no second node to mis-bind to and the `[impl+struct]` subtype set cannot mask one. (Confirmed in review by reading the hash rather than this sentence. The shape is not hypothetical: the gated fixture `rust-tokio-mini/src/store.rs` is `struct MemoryStore` + `impl MemoryStore` in one file.) The test therefore reports the sorted **set** of subtypes behind each resolved ID (`Order` comes back as `impl+struct`); keying a map by ID and storing the subtype reads as a mis-binding when it is nothing of the kind.

## Mutants, scored in the MORE PERMISSIVE direction

Each edit was applied by exact-occurrence count with the enclosing function asserted, `go vet` first, a green baseline immediately before, and restored by `cp` from a shasum-verified backup.

| mutant | verdict |
|---|---|
| descend a qualified path (`crate::x::Order` → `Order`) | **DEAD** — 5 tests, incl. the forbidden row named for it |
| drop the same-file check (emit to any name) | **DEAD** — 10 tests |
| count `impl` as a type declaration | **DEAD** — 2 tests |
| drop the duplicate-name removal | **DEAD** |
| drop the per-field target dedup | **DEAD** — 3 tests |
| assign `Relationships` instead of appending | **DEAD** — 3 tests |
| collect bare `identifier` as well as `type_identifier` | **ALIVE, production-reachable** (`&'a Customer` beside a `struct a`; `[Order; SIZE]` beside a `struct SIZE`) → graded by the new `TestRustFieldTypeRefs_NonTypeIdentifiersAreNotCandidates`, then **DEAD** |
| delete the `primitive_type` skip case | **equivalent** — the case is inert (see above); the guard was removed rather than left ungraded |
| *(review R1)* collect primitives: `case "primitive_type", "type_identifier"` | **DEAD** — the 16-primitive shadow test. The behaviour the deleted guard described is graded even though the guard is gone |
| *(review R2)* drop `\|\| r.Subtype != "field"` from the compound guard | **equivalent under the current producer** — the stash has exactly one writer (`struct_fields.go`) on records hardcoded `Kind: "SCOPE.Schema", Subtype: "field"`, so no record can reach the guard failing it. Defence-in-depth against a second future writer; recorded as equivalent, not as DEAD |
| *(review R3)* delete `delete(r.Metadata, rustFieldTypeRefsMetaKey)` | was **ALIVE, production-reachable and ungraded** — survived all three suites incl. `./cmd/grafel/` → graded in `2d0b05705`, now **DEAD** |

## The one item review left open, and it is now closed (`2d0b05705`)

Deleting `attachRustFieldTypeRefs`'s `delete(r.Metadata, rustFieldTypeRefsMetaKey)` was **ALIVE through every suite, `./cmd/grafel/` included**, and it is not cosmetic: `internal/graph/graph.go:176` copies `EntityRecord.Metadata` straight onto the JSON-serialised `Entity.Metadata` (`json:"metadata,omitempty"`), so without that line **every Rust field entity in every indexed repo would persist a `field_type_refs` array into the graph**. The doc comment said "and clears the stash it consumed" and nothing observed it — the dominant defect class, prose asserting what no test observes.

`TestRustFieldTypeRefs_EmittedEdges` now asserts the key is absent on every field record, with a positive control on the record count so the assertion cannot pass by inspecting nothing. Re-scored after the change: **DEAD** — and the failure names `Customer.name` leaking `[String]`, i.e. a field that gets *no* edge leaks the stash too, which is why the assertion is over the whole field population rather than over the edge-bearing subset. Test-only change; no production file was touched (`git diff HEAD` clean on `field_type_refs.go` after the mutant, restored by `cp` from a shasum-verified backup).

## Remaining on #6912

java, kotlin, scala, go, php, swift, cpp, ruby. Java still retains no type string at all, only a `Signature` — a capture problem before an edge problem.

Refs #6912, #6984, #6991, #6986, #6726, #6906, #6975.

### DIFF
--- a/internal/extractors/rust/field_type_refs.go
+++ b/internal/extractors/rust/field_type_refs.go
@@ -1,0 +1,287 @@
+package rust
+
+import (
+	"github.com/cajasmota/grafel/internal/treesitter/ts"
+
+	"github.com/cajasmota/grafel/internal/extractor"
+	"github.com/cajasmota/grafel/internal/types"
+)
+
+// field_type_refs.go — the field→declared-type edge for Rust (issue #6912,
+// arm C; arm A is internal/extractors/csharp/field_type_refs.go (#6984), arm B
+// internal/extractors/proto/field_type_refs.go (#6991)).
+//
+// WHAT RUST ALREADY HAD, ASKED THE WAY ARM B TAUGHT US TO ASK IT. #6912's body
+// says no core extractor emits this edge; that sentence has now been corrected
+// twice, so the question this arm answered FIRST is arm B's: is the FIELD
+// record the anchor of some existing type edge, or is something else? For Rust
+// the answer is NEITHER. Every relationship this package produces was
+// enumerated (struct_fields.go:224 and rust.go:156/208/294/552/918):
+//
+//	struct/enum → field         CONTAINS   (attachRustFieldContains, #4854)
+//	trait/impl  → method        CONTAINS   (#144)
+//	operation   → callee        CALLS      (#616)
+//	file        → module        IMPORTS
+//	trait       → supertrait    EXTENDS
+//
+// None of them carries a field's declared type at any anchor — unlike protobuf,
+// where buildMessage already wired the MESSAGE and left the field a leaf. So
+// this arm ADDS the relation rather than duplicating an anchor: there is no
+// existing topology to preserve, and nothing to double-count against. The type
+// was retained only as the `field_type` PROPERTY and inside the Signature
+// ("<type> <wire>"), which is exactly the shape #6912 describes: `Order.buyer`
+// knows it is a `Customer` and the graph cannot answer "which fields point at
+// Customer?".
+//
+// VOCABULARY AND ADDRESS ARE INHERITED, NOT RE-DECIDED. Kind REFERENCES with
+// the property `ref_kind: "field_target_type"` — the custom lane's spelling in
+// five languages (`referencesClassEdge`), adopted by arms A and B. `TYPED_AS`
+// and `HAS_TYPE` were DELETED in #6992 precisely so a third spelling could not
+// be minted. The address is the structural, file-scoped
+// extractor.BuildComponentStructuralRef("rust", <file>, <Name>) —
+// `scope:component:class:rust:<file>:<Name>` — and NOT the custom lane's
+// `Class:<Name>`, which #6986 measured at 91.6% dangling because the bare-name
+// resolver tier goes AMBIGUOUS as soon as a second file declares the same name
+// (`Config`, `Error`, `Client` in any Rust workspace). Post-resolution the ToID
+// is the target entity's ID, so no query sees the dialect. There is ONE address
+// dialect on this arm, not arm A's two: a Rust `enum Foo` is a SCOPE.Component
+// (subtype "enum"), so it takes the same component address as a struct. The
+// separate SCOPE.Enum value-set nodes emitted by emitRustConstValueSets (#4431)
+// are a parallel population and are deliberately NOT targeted — targeting both
+// would put two edges on one field for one declaration.
+//
+// A NON-BINDING EDGE IS WORSE THAN NO EDGE. An unresolved stub is kept verbatim
+// and dangling and classified `bug-extractor`; there is no masking branch and
+// no drop pass, so every miss is a full hit and each emitted edge adds one
+// endpoint to the disposition denominator (#6906). The edge is therefore
+// emitted ONLY where the named type is DECLARED IN THE SAME FILE — the single
+// condition a pass-1 per-file extractor can verify. That one check is what
+// drops `String`, `HashMap`, `serde_json::Value` and every unmodelled name.
+//
+// WHAT THE CHECK IS NOT.
+//
+//   - IT IS NOT A PRIMITIVE BLOCKLIST, AND IT DOES NOT NEED ONE — this is the
+//     one guard arm B had to keep, and the reason Rust does not is a property
+//     of the grammar rather than of our corpus. In tree-sitter-rust, a
+//     primitive in TYPE POSITION is the node kind `primitive_type`, never
+//     `type_identifier`, at every depth: `i32`, `Vec<u8>`, `Box<str>`,
+//     `(Order, i32)`, `fn(Order) -> u8`. rustTypeRefCandidates collects
+//     `type_identifier` only, so no primitive is ever a candidate and there is
+//     nothing for a blocklist to catch. arm B's counter-example was
+//     CONSTRUCTED here rather than argued away (`struct i32;` beside a field
+//     `x: i32`, for all 16 primitive names): the shadowing struct IS in the
+//     in-file target set, and the field still yields no candidate because the
+//     grammar tokenised its type as `primitive_type`. If rustc resolves such a
+//     shadow to the struct, this pass MISSES an edge — the safe direction, a
+//     miss never dangles — where proto's missing blocklist EMITTED one protoc
+//     denies. Enumerated per member by
+//     TestRustFieldTypeRefs_EveryPrimitiveShadowedByASameFileStructProducesNoEdge.
+//
+//     The first cut of this arm ALSO carried explicit `primitive_type`,
+//     `lifetime` and `scoped_identifier` skip cases in rustTypeRefCandidates,
+//     and they were deleted after a mutant showed the primitive one to be
+//     INERT rather than redundant-but-live: all three node kinds are either
+//     leaf tokens or carry only `identifier` children, so the collector
+//     returns nothing for them whether the case is present or not. That is a
+//     stronger claim than arm B's deleted blocklist could make — it does not
+//     rest on "our corpus has no such file" — and it was checked in both
+//     directions anyway: the deletion leaves every fixture green AND leaves
+//     the corpus counts below byte-identical over 789 real .rs files. The one
+//     case that is LIVE, `scoped_type_identifier`, is kept and is graded by
+//     TestRustFieldTypeRefs_QualifiedPathIsNotStrippedToASameFileName.
+//
+//   - IT IS FILE SCOPE AND NOTHING ELSE. It consults neither Rust's module
+//     scope nor type-parameter scope, so it over-fires twice, and both
+//     over-fires produce an edge that BINDS — worse than a dangling edge,
+//     because `bug-extractor` never sees a bound edge:
+//
+//     mod a { pub struct Customer; }
+//     mod b { pub struct Order { buyer: Customer } }   ← one file: WRONG edge
+//     struct Box<Order> { item: Order }                ← beside a same-file
+//     struct Order: WRONG
+//
+//     So an open type parameter `T` is dropped only because nothing in the file
+//     happens to be named `T`, NOT because the guard understands type
+//     parameters. Both are PINNED as known-wrong by the two
+//     TestRustFieldTypeRefs_KnownOverFire_* cases, which a fix is expected to
+//     break.
+//
+//   - IT DOES NOT DESCEND A QUALIFIED PATH. `crate::models::Order`,
+//     `super::Order`, `self::Order` and `<Order as Trait>::Assoc` are
+//     `scoped_type_identifier` and are skipped WHOLE — the trailing segment is
+//     never taken. Writing `crate::other::Order` in a file that also declares
+//     its own `Order` names the OTHER type, and stripping to the last segment
+//     would bind the edge to the wrong entity, silently. Arm B carries a
+//     forbidden row named for exactly that mutation; so does this arm
+//     (TestRustFieldTypeRefs_QualifiedPathIsNotStrippedToASameFileName).
+//
+//   - THE WRAPPER IS NEVER THE TARGET, but not by a wrapper list. A
+//     `generic_type` contributes its constructor AND its arguments as
+//     candidates; `Vec`, `Option`, `Box`, `Arc`, `Mutex` and `HashMap` are not
+//     declared in the file, so only the element type survives the in-file
+//     check. A user-defined generic declared here (`struct Holder<T>`, used as
+//     `Holder<Order>`) is a legitimate target by the same rule, which is why
+//     the constructor is not blanket-dropped.
+//
+//   - `impl Order { … }` IS NOT A TYPE DECLARATION and is excluded from the
+//     target set. buildImpl emits a SCOPE.Component named after the implementing
+//     type, so a Rust file that declares `struct Order` and `impl Order` holds
+//     TWO components named `Order` — the overwhelmingly common shape. Treating
+//     the impl as a declaration would collide every such name and silently
+//     delete the edges this pass exists to add.
+//
+// THE COST OF THE SAME-FILE RULE, stated plainly: a field whose type is declared
+// in another module file gets no edge, and Rust's one-type-per-file convention
+// is weaker than C#'s but real. Closing that needs a cross-file type view, which
+// FileInput has none of in pass 1. Separate arm, not a widening of this one.
+
+// rustFieldTypeRefsMetaKey is the per-field stash written by rustFieldsFromList
+// and consumed — and deleted — by attachRustFieldTypeRefs. The candidates
+// cannot become edges during the walk: `struct Order { buyer: Customer }` may be
+// declared BEFORE `struct Customer` in the same file, so the set of in-file
+// declarations is complete only once the walk has finished.
+const rustFieldTypeRefsMetaKey = "field_type_refs"
+
+// rustFieldTargetRefKind is the value of the `ref_kind` edge property, matching
+// arm A's csFieldTargetRefKind, arm B's protoFieldTargetRefKind and the custom
+// lane's referencesClassEdge verbatim. The discriminator is what makes the edge
+// queryable as a field→declared-type edge across languages, so it is one string.
+const rustFieldTargetRefKind = "field_target_type"
+
+// rustTypeRefCandidates returns every bare type name written in a field's
+// declared type expression, in source order and without deduplication.
+//
+// It descends the type node and collects `type_identifier` leaves, which unwraps
+// reference (`&'a Customer`, `&mut Order`), pointer (`*const Order`), array
+// (`[Order; 3]`), tuple (`(Order, i32)`), dyn (`Box<dyn Shipper>`), function
+// (`fn(Order) -> u8`) and generic syntax for free. A lifetime carries its name as
+// an `identifier`, not a `type_identifier`, so `'a` is dropped by the same rule
+// that drops nothing else.
+//
+// ONE node kind is skipped WHOLE, without descending: `scoped_type_identifier`
+// — a qualified path. `crate::models::Order`, `super::Order`, `self::Order` and
+// `<Order as Trait>::Assoc` therefore yield NOTHING rather than `Order`, which
+// is the point (see the header block).
+//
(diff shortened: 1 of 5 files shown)


cajasmota:

Measured in my own worktree at `7d45af712`:
`/private/tmp/claude-501/-Users-jorgecajas-Projects-archigraph/c0fefed4-0195-4c81-9a94-07a7145710a4/scratchpad/review-7000-k7q2m/wt`
(macOS 25.6.0 / APFS / arm64). Suites clean at HEAD: `./internal/extractors/rust/...` ok, `./internal/custom/rust/...` ok, `./cmd/grafel/` ok (162s, fresh isolated `GRAFEL_HOME` + `GRAFEL_DAEMON_ROOT`), ratchet **RC=0, 38 gated fixtures held**. `go vet` clean.

### (a) Deleting the three skip cases — **SAFE, and now with the compiler's verdict**

I did not take the namespace argument. I installed **rustc 1.98.1 (48a229cea 2026-09-01)** into scratch and compiled the case.

**The case is real — arm B's shape is genuinely present here.** `struct i32;` beside `pub f: i32` compiles, `RC=0`, for **all 17** primitive names (`i8 i16 i32 i64 i128 isize u8 u16 u32 u64 u128 usize f32 f64 bool char str`), enumerated, not sampled. Harness controls: `struct Foo` + `f: Foo` → RC=0; an undeclared type → RC=1. (My first pass reported RC=1 for all 17 — the *positive control* caught it as `-o /dev/null` failing to make a temp dir, not a language fact. Worth recording: without that control I would have filed the opposite finding.)

And rustc **binds the field to the struct, not the primitive**:

```rust
pub struct i32;
pub struct S { pub f: i32 }
pub fn mk() -> S { S { f: i32 } }   // RC=0 — `i32` used as a VALUE
```
Counter-control, same file without the shadow: `error[E0423]: expected value, found builtin type i32`. So the shadow wins, and `str` compiling as a *sized* field is the same fact from the other side.

**So the case exists, and grafel MISSES it — the safe direction.** That is the load-bearing difference from #6991: proto's absent blocklist **emitted** an edge protoc denies; Rust's absent one can only fail to emit one. A miss never dangles.

**Checked at every depth, not top level.** I wrote my own probe (`TestReview_PrimitiveNeverEdgesAtAnyDepth`) — **17 primitives × 16 nesting shapes = 272 enumerated cases**: top, `Vec<>`, `Option<>`, `Box<>`, `&'static`, `&'static mut`, `*const`, `[P; 3]`, `&'static [P]`, `(P, u8)`, `Vec<Option<Box<P>>>`, `HashMap<u8, P>`, `fn(P) -> u8`, `fn(u8) -> P`, `Vec<(Option<P>, [P; 2])>`, `Holder<Vec<&'static P>>`. Each case carries two controls: `struct P` is asserted present in the in-file target set, **and** the identical shape with `Customer` must produce an edge — so "no edge" can never come from the shape failing to parse. **All 272 pass; no primitive edged at any depth.**

`lifetime` / `scoped_identifier` likewise, with all shadow names declared (`struct a; struct b; struct SIZE; struct other; struct models;`): `&'a Customer`, `&'b mut Order`, `Vec<&'a Customer>`, `Holder<&'b Order>`, `Box<dyn Iterator<Item = &'a Order> + 'b>`, `[Order; SIZE]`, `Vec<[Customer; SIZE]>` bind only the real types; `crate::other::Order`, `super::models::Customer`, `Vec<crate::other::Order>`, `Option<Box<self::models::Customer>>` yield **nothing** at every depth — the qualified path is skipped whole, not stripped, nested as well as bare.

**Removing the three cases is right.** Arm B's lesson was *construct the case*, and the case was constructed correctly here.

### (b) The `impl` pin and the one-node claim — **both hold**

`graph.EntityID` (`internal/graph/graph.go:259`) hashes `repo ∥ Kind ∥ Name ∥ SourceFile` — **`Subtype` is not in the hash**. `struct Order` and `impl Order` in one file are both `SCOPE.Component`/`Order`/same file, so they are literally one ID. The `[impl+struct]` subtype *set* in `TestRustFieldTypeRefs_ResolverBindsEveryEdge` is therefore not masking a mis-binding — there is no second node to mis-bind to. Read the hash, did not take the prose.

The pin is real and the shape is not hypothetical: the gated fixture itself (`rust-tokio-mini/src/store.rs`) is `struct MemoryStore` + `impl MemoryStore` in one file.

### (c) Over-firing — **none survived**

Every §3 attack landed where the PR says: `crate::`/`super::`/`self::` paths get nothing (at depth too); wrappers are never the target (only the element survives the in-file check — `Holder` *is* targeted, correctly, as a same-file user-defined generic); the lifetime name and the array-length const are both graded with their shadowing structs declared. The two known over-fires (module scope, type-parameter shadow) are pinned as known-wrong rather than described.

### (d) The +1 baseline — **exactly right, reproduced**

`git diff 2d6c82ba8..7d45af712 -- baseline.json` is **exactly two lines** (`63` → `64`), so a hand-edit cannot have silently lowered another ceiling — the file has no other change. I then built `cmd/grafel` at **both** commits and ran `quality` on the fixture:

| | entity_extracted_total | relationship_extracted_total |
|---|---|---|
| parent `2d6c82ba8` | 31 | **63** |
| HEAD `7d45af712` | 31 | **64** |

`+1`, and `src/store.rs` declares `pub struct User` alongside `MemoryStore.users: HashMap<String, User>` — the enumerated edge. `rust-tokio-mini` is the only rust fixture, and the ratchet held all 38.

### My mutants (permissive direction; none of the author's 8 re-run)

| # | mutant (enclosing function) | edit landed | verdict |
|---|---|---|---|
| R1 | `rustTypeRefCandidates`: `case "type_identifier"` → `case "primitive_type", "type_identifier"` — i.e. *collect* primitives | 1/1 occurrence in the file (2 repo-wide, sibling untouched), confirmed inside `rustTypeRefCandidates` by `awk` | **DEAD** — killed by `TestRustFieldTypeRefs_EveryPrimitiveShadowedByASameFileStructProducesNoEdge`. The inverse of the deletion: the *behaviour* the deleted guard described is graded even though the guard is gone. |
| R2 | `attachRustFieldTypeRefs`: drop `\|\| r.Subtype != "field"` from the compound guard | 1/1, confirmed in-function | **equivalent under the current producer** — `rustFieldTypeRefsMetaKey` has exactly one writer (`struct_fields.go:168`) on records hardcoded `Kind: "SCOPE.Schema", Subtype: "field"`, so no record can reach the guard failing it. Defence-in-depth against a second future writer; correctly ungraded. |
| R3 | `attachRustFieldTypeRefs`: delete the `delete(r.Metadata, rustFieldTypeRefsMetaKey)` line | 1/1 → 0 post-edit, confirmed in-function | **ALIVE, distinguishable, production-reachable, ungraded** — see below. |

**R3, the one finding.** Survives `./internal/extractors/rust/...`, `./internal/custom/rust/...` **and `./cmd/grafel/` (171s, mutant applied)**. It is reachable, not cosmetic-only: `graph.go:176` copies `EntityRecord.Metadata` straight onto `Entity.Metadata`, which is `json:"metadata,omitempty"` — so without the `delete`, every Rust field entity in every indexed repo would persist a `field_type_refs: [...]` array into the graph. The doc comment says *"and clears the stash it consumed"*; nothing observes it. Classic **prose asserts what no test observes**.

Not a blocker — the `delete` is present and correct on this branch — but it is ungraded, and I priced the grading rather than calling it expensive: **~4 lines** in the existing `TestRustFieldTypeRefs_EmittedEdges`, asserting `recs[i].Metadata["field_type_refs"]` is absent on the field records. Worth adding here or as a follow-up.

---

**APPROVE.** The deletion is safe and, unusually, is now backed by the compiler saying the case *is* constructible — which strengthens rather than weakens the claim, because grafel's failure mode there is a miss. Not merging, not pushing.

*Reviewer probe (`zzreview_depth_test.go`, 272 + 1 cases) was written in my scratch worktree only and is not part of this branch; the tree was restored by `cp` from a shasum-verified backup and `git diff HEAD` confirmed empty after every mutant.*

cajasmota:
## Coordinator mutant CN-1 on `2d0b05705` — DEAD. Approving; merging on green.

Own detached worktree, green baseline first, restored by `cp` from a SHA-256-verified backup (`6b632375…`) **and verified with `git diff HEAD`** — clean. macOS/APFS.

**CN-1 — score the compound guard part by part.** Ten mutants have been scored here. The eligibility guard in `rustInFileTypeTargets` is a three-way disjunction:

```go
if r.SourceFile != filePath || r.Kind != "SCOPE.Component" || r.Name == "" {
```

Your "drop the same-file check" mutant grades the **first** disjunct. **Nobody graded the second.** Two guards that only ever fire together grade neither, so I removed the `Kind` disjunct alone, leaving the other two intact so the mutant cannot be masked by them:

```go
-  if r.SourceFile != filePath || r.Kind != "SCOPE.Component" || r.Name == "" {
+  if r.SourceFile != filePath || r.Name == "" { // MUTANT CN-1: Kind disjunct dropped
```

Edit proven landed: exactly 1 occurrence, hit set `[221]`, **enclosing function asserted** to be `rustInFileTypeTargets`, `go vet` clean before scoring.

**Result: DEAD** — `TestRustFieldTypeRefs_EmittedEdges`, `_EdgeProperties`, `_ResolverBindsEveryEdge`.

Worth noting *why* it dies, because it is the same hazard your `impl` finding is about: without the `Kind` check, any record whose `Subtype` happens to be `struct`/`enum`/`trait`/`type_alias` becomes a declaration candidate regardless of its kind — a second claimant for a name, which the collide rule then **deletes**. So the two halves of that guard defend the same property from opposite directions, and both are now graded.

**Nothing else is left ungraded that I can find.** The subtype allow-list, the collide removal, dedup, append-not-assign, qualified descent, `impl`-as-declaration and the same-file check are all DEAD; `primitive_type` was proven inert and removed rather than left ungraded; R2 is correctly recorded as **equivalent under the current producer** rather than DEAD.

## On the round

**R3's failure output is better than the fix.** It names `Customer.name` leaking `[String]` — a field that gets *no* edge still leaked the stash — which is why the assertion had to range over the whole field population rather than the edge-bearing subset. A test written only against edge-bearing records would have passed while the leak shipped.

**And the rustc verification is the strongest evidence any arm in this family has produced.** 17/17 compiling at RC=0, `E0423` as the counter-control proving rustc really binds the field to the shadow, 272 depth cases with two positive controls each. The distinction that matters is the one the body now states: grafel's failure there is a **miss**, never a wrong emission — the opposite of arm B's blocklist, which emitted an edge protoc denies. That is why deleting here is safe and deleting there was not.

The harness note belongs exactly where it now is. The first rustc pass returned RC=1 for all 17 and would have produced the **opposite** conclusion; the positive control caught it as a broken invocation rather than a language fact. A compiler-verified claim is only as good as the harness, and that is the sentence that will save the next person.

Issue #6722, opened 2026-08-27
MCP client doesn't reconnect after `grafel restart` / `doctor --kill-stale` — persistent 'Failed to connect to MCP server'


auxmedrano:
### Anonymization verification

- [x] I have verified this report contains no real source code, file paths, or identifier names (all entity names look like ent_a3f7 or op_92c1; all paths look like `<go>/<seg-1>/<seg-2>.go`).
- [ ] I generated this report with `grafel feedback` v0.9+ (running v0.3.0, so leaving unchecked)

### Feedback report

<details>
<summary>Full report (paste, could not drag-and-drop the file from this environment)</summary>

# grafel feedback report

Generated: 2026-08-27T20:34:59Z
grafel version: v0.3.0 (commit b037c3f, built 2026-08-21T18:26:14Z)
Group profile: 18 language(s) (go, typescript, python, vue, avro, terraform, yaml, javascript, sql, shell, markdown, dockerfile, html, scss, env, java, makefile, json), 139000-140000 entities, 660000-661000 relationships
Confidence: 85% (51/60 sanity checks passed)

### Entities by language

| Language | Entity count (range) |
|---|---|
| avro | 100+ |
| dockerfile | 100+ |
| env | 21-100 |
| go | 100+ |
| html | 21-100 |
| java | 21-100 |
| javascript | 100+ |
| json | 6-20 |
| makefile | 6-20 |
| markdown | 100+ |
| python | 100+ |
| scss | 21-100 |
| shell | 100+ |
| sql | 100+ |
| terraform | 100+ |
| typescript | 100+ |
| vue | 100+ |
| yaml | 100+ |

### Entity kind distribution

| Kind | Language | Count (range) |
|---|---|---|
| SCOPE.Schema | avro | 100+ |
| Component | dockerfile | 21-100 |
| Config | dockerfile | 100+ |
| Dependency | dockerfile | 21-100 |
| Operation | dockerfile | 21-100 |
| SCOPE.Component | dockerfile | 21-100 |
| SCOPE.Config | dockerfile | 21-100 |
| SCOPE.Config | env | 21-100 |
| Config | go | 6-20 |
| Route | go | 100+ |
| SCOPE.Component | go | 100+ |
| SCOPE.Config | go | 100+ |
| SCOPE.DataAccess | go | 100+ |
| SCOPE.Enum | go | 100+ |
| SCOPE.ExceptionType | go | 100+ |
| SCOPE.Operation | go | 100+ |
[...]

Issue #5633, opened 2026-06-25
daemon restart: drain in-flight MCP calls + guarantee single mcp-bridge (no orphan, clean reconnect)


cajasmota:
Restarting the daemon surfaces `connection is shut down to the client` for in-flight MCP calls; the bridge reconnects but in-flight calls hard-fail, and two `mcp-bridge` processes were seen at once (orphan). Make restart graceful: drain/await in-flight calls before close (or return a retryable error the bridge auto-retries), ensure exactly one bridge survives a restart, and a clean reconnect window.

cajasmota:
## Plan

Investigated the bridge (`internal/cli/mcp_bridge.go`), the daemon RPC accept loop + shutdown (`internal/daemon/server.go`), and the MCP RPC methods (`internal/daemon/mcp_rpc.go`).

**Root causes**
1. **Hard-fail on restart.** When the daemon closes the socket mid-call, the client-side `rpc.Call` returns `rpc.ErrShutdown` / `io.EOF`. `handleToolsCall` resets the cached client but *surfaces the error to the MCP client* as `grafel daemon error: connection is shut down` — it never reconnects+retries.
2. **No bounded drain.** Shutdown closes the listener then `connWG.Wait()` (unbounded), but in-flight MCP dispatch is not explicitly awaited with a cap before the process exits, so a slow in-flight call can still get its socket killed.
3. **Orphan bridges.** Nothing reaps a prior `grafel mcp-bridge` for the same socket, so two can coexist after a restart.

**Fix (three parts)**
1. **Graceful drain (daemon):** add an explicit in-flight MCP WaitGroup + `draining` flag on `Service`. On shutdown set draining (new MCP RPCs get a clean retryable error instead of being half-served), then `Wait` for in-flight MCP calls with a bounded timeout before closing the socket / exiting.
2. **Bridge reconnect+retry:** on `rpc.ErrShutdown` / `io.EOF` / `io.ErrUnexpectedEOF` mid-call, reset the client, reconnect, and retry (bounded attempts + small backoff) for both `tools/list` and `tools/call`. All grafel MCP tools are read-mostly graph queries — safe to retry; no non-idempotent call exists, so no special-casing needed. Only a persistent failure surfaces an error to the client.
3. **Single bridge:** reap any orphaned prior `grafel mcp-bridge` on startup (mirroring the `internal/process` reaping approach from #5632), keyed by the target socket, so a second bridge can't linger.

[...]

PR #7001, opened 2026-09-08
fix(#6999): the bridge's singleton was keyed on a machine-global socket path, so every new mcp-bridge SIGTERMed every other session's — the reap is deleted, not rescoped

### LINKED ISSUE #6999: FIXED (82396b1ca): MCP 'Transport closed' — mcp-bridge keyed its singleton pidfile on the per-user daemon socket, so every new bridge SIGTERMed every other SESSION's bridge (v0.3.2 since #5633; unix/macOS only, Windows was never affected)
## Summary
During a heavy automated run (grafel **v0.3.2**, worktree-aware `--track-worktrees`, a React/TS/Vite project), **13 of 113 grafel MCP calls (≈11.5%) failed with an identical, tool-agnostic error** — the MCP server dropped the transport connection mid-call:

```
tool call error: tool call failed for `grafel/<tool>`
Caused by:
    Transport closed
```

## Distribution (all same "Transport closed" signature)
grafel_find ×7 · grafel_impact_radius ×2 · grafel_get_source ×1 · grafel_index_status ×1 · grafel_orient ×1 · grafel_endpoints ×1. (grafel_related / test_analysis / diff had 0 errors.)

**Key signal that it's server-side liveness, not bad args/query cost:** even `grafel_index_status` (a trivial health call) hit it. The failure is independent of the tool or arguments — whatever call is in flight when the transport drops dies. Suggests the grafel MCP server **crashes / restarts / OOMs** intermittently, taking down the stdio connection.

## Environment
- grafel **v0.3.2** (commit de290e8), background daemon mode, incremental indexing on, mem cap 2560MB (768 serve + 1792 engine).
- Worktree-aware: calls issued with `cwd` pointing at git worktrees under `/home/agent/repo/.worktrees/tsk_*` (per-task worktrees, cap ~10).
- Project: React 19 + Vite + TS + Tailwind (`the-factory-web`), ~small codebase, single repo (+ a related specs repo).

## Triggering calls (all well-formed, verbatim-sampled)
- `grafel_find {cwd:".worktrees/tsk_30b5fa11", query:"realtimeFrameSchema", search:"bm25", limit:20, token_budget:2000}`
- `grafel_find {cwd:".worktrees/tsk_8f5cb947", query:"briefing", search:"substring", cross_repo:true, limit:100, token_budget:10000}`  ← heaviest; a plausible OOM repro (cross_repo + limit 100 + 10k budget)
- `grafel_impact_radius {cwd:".worktrees/tsk_14fd0be8", scope:"changeset", base:"origin/main", head:"HEAD", detail:"full", token_budget:6000}`
- `grafel_get_source {cwd:".worktrees/tsk_22d53d1c", entity_id:"the-factory::fb1fe9282ac6e498", max_lines:1000}`
- `grafel_orient {cwd:".worktrees/tsk_1fce9975", view:"overview", token_budget:3000}`
- `grafel_index_status {cwd:".worktrees/tsk_a0aad136"}`  ← trivial call, still hit "Transport closed"
- `grafel_endpoints {cwd:".worktrees/tsk_8f5cb947", action:"definitions", path_contains:"briefing", limit:20}`

## Ask
Check the grafel daemon/server logs for a crash / panic / restart / OOM-kill correlated with worktree-scoped requests (the `cwd=.worktrees/tsk_*` pattern). Two likely angles: (1) a memory spike on a heavy `cross_repo` find exceeding the engine mem cap → OOM → transport drop; (2) a liveness bug in worktree mode where a per-worktree graph rebuild/reconcile races the serving path and drops the connection. A retry-on-transport-drop in the bridge would mask it, but the root crash is the fix. Happy to provide more telemetry.

cajasmota:
Closes #6999.

## What was wrong

`internal/cli/mcp_bridge_singleton.go` implemented #5633's "exactly one bridge per daemon socket" with a pidfile keyed on the **daemon socket path**. `daemon.DefaultLayout()` makes that `<per-user root>/sockets/daemon.sock` — no cwd, repo, worktree or session component. So **one pidfile named the single bridge of an entire machine**, and every newly started bridge read it and `process.Kill`ed the pid it found: the incumbent bridge of some *other* live session, across projects and across worktrees.

The victim had **no SIGTERM handler**, so it died on the default disposition (exit 143) with stdin still open and wrote nothing anywhere. The reaper's own line went to the *client's* stderr, so `~/.grafel/logs/` recorded nothing either. Client-visible result: a tool-agnostic `Transport closed` killing whatever call was in flight.

## First question: is the reap needed at all?

**No — established, and pinned by a test.**

`bridge.run` reads stdin and returns `nil` on `io.EOF`. When an MCP client dies, its end of the pipe closes and the bridge falls out of its loop unaided. `TestBridge_ExitsWhenStdinCloses` asserts exactly that, and names it as the premise the fix rests on.

It follows that the orphan #5633 observed had a **live client** (a daemon restart does not touch stdin). And a bridge with a live client that has lost the *daemon* is, from the outside, **indistinguishable from a second concurrent session's perfectly healthy bridge** — same executable, same socket, same everything a pidfile can see. There is no key that unifies "an orphan and its replacement" while separating "two concurrent sessions". Choosing to kill inside that ambiguity is precisely what produced #6999, so the fix is not a narrower key for the kill: **the kill is deleted.** The remedy for a bridge that has lost the daemon is owned by that bridge — reconnect (#6722) or exit — not by a stranger with a signal.

## What the change does

1. **No bridge signals any process, ever.** `reapPriorBridge` and `bridgeReapGrace` are gone.
2. **The record is per-session**, keyed on `(socket, session id)`. The session id is the explicit `GRAFEL_MCP_SESSION`, else the device/inode of the client's stdin pipe, else this process's pid. It scopes a filename and nothing else; a colliding id costs one overwritten record, because no signal is derived from it.
3. **A signalled bridge names its own death** — one line to stderr *and* to `~/.grafel/logs/daemon.log` giving the pidfile, the socket, the signal, and #6999 — then exits 128+signum, so a supervisor sees the same status as before.
4. #5633's other two parts (graceful drain, reconnect+retry) are untouched; they are what actually fixed its reported symptom.

## Secondary hazard — verified, and eliminated

Confirmed real on `main`: `isLiveBridge` returned **true** when `process.PidIsGrafel` errored, and `PidIsGrafel` matches **any** grafel process by executable basename (`exeBaseNameIsGrafel`, substring match). A stale pidfile whose pid had been recycled by the daemon or the engine was therefore a valid SIGTERM target — **the bridge could kill the daemon.**

It is eliminated by construction (nothing signals on that answer any more), and additionally fixed: the unverifiable case now answers **false**. Graded — see M3.

## Tests

`#6999` is a cross-process defect; no in-process assertion can observe it. The new tests run **real bridge processes**, and the children are copies of the test binary named `grafel`, because the old reap was gated on `PidIsGrafel`: with a child called `cli.test` the reap would have been skipped and the test would have passed against the bug.

- `TestBridge_ConcurrentSessionsAllSurviveAndServe` — the headline. Three bridges on **one** socket from a project, a **second project**, and a **worktree of the first** (each member asserted individually, not one representative). After each start, every earlier session must still be alive **and still answer an `initialize`**. Then: three **distinct** ownership records, each naming its own pid.
- `TestBridge_DoesNotSignalAnotherLiveGrafelProcess` — the negative. A live grafel-named process is seeded into both the legacy per-socket path and this session's path; starting a bridge must leave it running.
- `TestBridgeSingleton_NeverSignalsPriorBridge` — the unit-level pin: acquiring over a record naming a live grafel process leaves it alive and logs instead.
- `TestBridge_SignalledBridgeNamesItselfInTheDaemonLog` — end-to-end through the production wiring: a real bridge is SIGTERMed and must leave exactly one line in `daemon.log` naming its pidfile, its socket and #6999, while still exiting 143.
- Plus: stdin-close self-exit, per-session vs per-socket key, session-id sources, and `isLiveBridge` on an unverifiable executable.

## Mutants (permissive direction), `go vet` first, green baseline before each, `-count=1`

| # | Mutant | Verdict |
|---|---|---|
| M1 | restore the reap (`process.Kill(prior)`) | **DEAD** — `NeverSignalsPriorBridge`: "prior bridge (pid N) was terminated by acquire" |
| M2 | widen the key back to the socket alone | **DEAD** — `ConcurrentSessions…` ("has no ownership record of its own") + `PathIsPerSessionNotPerSocket` |
| **M1+M2** | **both — i.e. the exact pre-#6999 code** | **DEAD** — `ConcurrentSessions…`: *session "project-a" DIED when session "project-b" started*; also `DoesNotSignalAnotherLiveGrafelProcess` and 2 more. This is the production repro. |
| M3 | `isLiveBridge` returns true when `PidIsGrafel` errors | **DEAD after grading.** Initially *distinguishable and ungraded* — no seam made the error branch reachable. Priced at 4 lines (a `pidIsGrafel` package var), taken; `TestIsLiveBridge_UnverifiableExecutableIsNotABridge` now kills it. |
| M4 | drop the daemon-log mirror | **DEAD** — both daemon-log assertions |
| M5 | never arm the signal logger in `bridge.run` | **DEAD** — signalled bridge exits −1 (killed) instead of 143 |
| M6 | collapse every signal to exit 143 | **DEAD** — SIGINT must be 130 |

Each edit was applied with an asserted occurrence count of 1, its enclosing function printed, and restored by `cp` from a shasum-verified backup with `git status` confirming a clean tree afterwards (`grep -rn MUTANT internal/cli/` finds only a pre-existing unrelated comment).

## Verification

Tree: `…/scratchpad/fix-6999-v8n3p/wt`, branched from `c6a99d75f`. Every run under a fresh isolated `GRAFEL_HOME` + `GRAFEL_DAEMON_ROOT`; the developer's `~/.grafel` was never written to.

- `go test -count=1 ./internal/cli/...` — ok (58.6s) + wiztui ok
- `go test -count=1 ./cmd/grafel/` — ok (149.2s), fresh isolated root
- `go test -count=1 ./internal/quality/ ./internal/process/ ./internal/daemon/` — all ok
- `gofmt -l internal/cli` clean, `go vet ./internal/cli/` clean

**Not run: the full `scripts/quality/run.sh --ratchet`.** The diff touches only the MCP bridge's process lifecycle (`internal/cli/mcp_bridge*`) — no extractor, resolver or graph surface — and the always-on structural gate `go test ./internal/quality -run TestBaseline` is green as part of the package run above. Say the word if you want the ~40-minute indexing ratchet run anyway.

**This PR needs the `ci:full` label** — it is `runtime.GOOS`-conditional territory (signals, pidfiles, unix vs windows session identity), and macOS tests only run with that label.

---

### New finding: **#6999 never affected Windows, and neither did the fix's diagnostic**

The Windows CI leg on `d766d630a` was red, and it was a real platform gap, not a flake:

```
--- FAIL: TestBridge_ConcurrentSessionsAllSurviveAndServe
    session "project-a" (pid 2380) has no ownership record of its own; records=map[]
```

`records=map[]` — **empty**. Every earlier assertion passed, so all three bridges started, survived and kept serving; only the record enumeration failed, because **no record was written by anybody**.

Cause, verified along the whole chain (`bridge.run` → `defaultSocketPath` → `daemon.DefaultLayout().SocketPath` → `bridgeSingletonPath` → `filepath.Dir`): on Windows `SocketPath` is a **named pipe** (`\\.\pipe\grafel-<hash>`) and `SocketDir` is deliberately `""` (`internal/daemon/paths_windows.go`: *"named pipes have no filesystem directory"*). So `filepath.Dir(SocketPath)` is `\\.\pipe`, `os.WriteFile` always fails there, and `acquireBridgeSingleton`'s error is **non-fatal by design** — the bridge logged `bridge singleton: … (continuing)` and served on. The record has never existed on Windows.

**The pre-#6999 code derived the pidfile the same way** (`bridgeSingletonPath(socketPath) { dir := filepath.Dir(socketPath) … }`, `d766d630a^`). The reap was gated on `readBridgePID(path)` returning a prior pid, and that file was never written — so **the reap never fired on Windows. Windows users were never affected by #6999.**

Its limits, stated plainly: the reap evidence in the issue (83 reap lines across 10 project dirs) is **macOS-only**. It is *consistent with* this conclusion but is **not independent confirmation** of it — the conclusion rests on the code path above, not on the absence of Windows log lines, which nobody collected.

### Fix: derive the record dir from `Layout.Root`, not from the socket

`bridgeRecordDir()` returns `<Layout.Root>/sockets` on every platform, and `acquireBridgeSingleton` `MkdirAll`s it (nothing binds a socket there on Windows, so nothing else creates it). Rationale for that choice over the alternatives:

- `Layout.Root` is a real, writable directory on **both** platforms; `SocketDir` is empty on Windows and `filepath.Dir(SocketPath)` is a pipe namespace.
- It **keeps the unix location unchanged** under `GRAFEL_DAEMON_ROOT` (already `<root>/sockets`), so nothing moves for the tests or for the common case. It additionally pins the record to grafel's own state when the socket is in `XDG_RUNTIME_DIR` (a per-boot tmpfs).
- `sockets/` is already one of grafel's scratch dirs (`internal/install.grafelScratchDirs`), so `uninstall --purge` still cleans records up.
- The session id **stays in the hash**, and so does the socket path.

New pin, unskipped on every platform: `TestBridgeRecordDir_IsAWritableDirectoryOnEveryPlatform` removes the dir, then writes through the production `acquireBridgeSingleton` against the real `DefaultLayout().SocketPath`. Its scope is stated in its doc comment: the *derivation* is only distinguishable where root and socket dir differ (Windows); what it grades everywhere is that the dir is created and written through — the half that never happened on Windows. `TestBridge_ConcurrentSessionsAllSurviveAndServe` now enumerates `bridgeRecordDir()` rather than a hand-built `filepath.Join(root, "sockets")`.

**Why this was invisible until now:** the old singleton tests used fabricated paths under `t.TempDir()` (`filepath.Join("x","y","daemon.sock")`), so they never exercised the real layout. The multisession test found it precisely because its children run `bridge.run` with the **production** `DefaultLayout`. That is the argument for the test.

### Two ALIVE mutants graded — both now DEAD

Each mutant applied with an asserted occurrence count of **1** and its hit set printed, `go vet` clean, green `-count=1` baseline immediately before, restored by `cp` from a shasum-verified backup with `git diff HEAD` empty.

| # | Mutant | Before | After |
|---|---|---|---|
| **M-C** | `appendDaemonLog` drops `os.O_APPEND` (occurrence 1, line 81; `grep -c O_APPEND` = 0 after) | ALIVE | **DEAD** — `TestBridgeSignal_NoticeNamesItsCause` now seeds prior content and asserts **`strings.HasPrefix`**. `Contains(prior)` still passes when the notice is written *in front of* it, and a wrong write offset is exactly the failure mode. `daemon.log` is the **daemon's** log, shared with concurrent writers. |
| **M-A** | release closure removes the record unconditionally (drop `cur == self`, occurrence 1) | ALIVE | **DEAD** — `TestBridgeSingleton_ReleaseLeavesANewerOwnersRecord` seeds a *different* pid as the newer owner, calls `release()`, and requires the file to still exist and still name that pid. The artefact, not a counter. |
| **new** | drop `acquireBridgeSingleton`'s `MkdirAll` | — | **DEAD** — `TestBridgeRecordDir_IsAWritableDirectoryOnEveryPlatform`. |

### The self-collection claim is now scoped to what is actually pinned

`mcp_bridge_singleton.go`'s header said a bridge whose client is gone *"exits on its own"* and named `TestBridge_ExitsWhenStdinCloses` as the pin. That test runs an **idle in-process bridge over an `io.Pipe`**: it pins EOF-while-idle, not "whose client is gone" — the same defect class this PR fixes one level down.

The header now says which shapes self-collect (client SIGKILL, client SIGTERM, client closes stdin while alive — all three verified with real processes by the reviewer), and names the one that does not: a bridge **blocked in a daemon call** is not in its read loop and never observes the EOF, because `callDaemon` → `net/rpc` `Call` has **no deadline**. Filed as **#7003**, and deliberately not fixed here — the old reap only collected that case when a *new* bridge started, i.e. the #6999 event itself. `TestBridge_ExitsWhenStdinCloses`'s doc comment now states its scope.

### Verification (this round)

Fresh isolated `GRAFEL_HOME` + `GRAFEL_DAEMON_ROOT` per run; the developer's `~/.grafel` was never written to; no process I did not start was signalled.

- `go test -count=1 ./internal/cli/ ./internal/daemon/ ./internal/process/` — all ok (58.5s / 32.3s / 1.0s)
- `go test -count=1 ./cmd/grafel/` — ok (177.3s), fresh isolated root
- `gofmt -l internal/cli` clean, `go vet ./internal/cli/ ./internal/daemon/` clean
- Cross-`vet` for `GOOS=windows` is not possible in this tree (the tree-sitter cgo bindings exclude all files under a cross build) — a pre-existing limitation, unrelated to this diff. **The Windows leg in CI is the check that matters here**, which is why the `ci:full` label was removed and re-added after the push.

### DIFF
--- a/internal/cli/mcp_bridge.go
+++ b/internal/cli/mcp_bridge.go
@@ -500,16 +500,20 @@ func (b *bridge) defaultSocketPath() (string, error) {
 func (b *bridge) run(r io.Reader, w io.Writer) error {
 	defer b.closeRPCClient()
 
-	// #5633: guarantee exactly one bridge per daemon socket. Reap an orphaned
-	// prior bridge (e.g. one left attached after a daemon restart) and claim a
-	// per-socket pidfile. Best-effort: a failure here must not stop us serving,
-	// so we log and continue. Skipped when the socket path cannot be resolved.
+	// #5633, rescoped by #6999: record this process as the bridge serving this
+	// socket FOR THIS SESSION, and arm the signal logger so an externally
+	// terminated bridge says so in the daemon log. Nothing here signals another
+	// process. Best-effort: a failure must not stop us serving, so we log and
+	// continue. Skipped when the socket path cannot be resolved.
 	if socketPath, serr := b.defaultSocketPath(); serr == nil && socketPath != "" {
-		if release, aerr := acquireBridgeSingleton(socketPath, b.log); aerr != nil {
+		release, pidfile, aerr := acquireBridgeSingleton(socketPath, b.log)
+		if aerr != nil {
 			b.log("bridge singleton: %v (continuing)", aerr)
 		} else {
 			defer release()
 		}
+		stopSignalLogger := installBridgeSignalLogger(pidfile, socketPath, b.log)
+		defer stopSignalLogger()
 	}
 
 	br := bufio.NewReaderSize(r, 64*1024)
--- a/internal/cli/mcp_bridge.go
+++ b/internal/cli/mcp_bridge.go
@@ -506,11 +506,20 @@ func (b *bridge) run(r io.Reader, w io.Writer) error {
 	// process. Best-effort: a failure must not stop us serving, so we log and
 	// continue. Skipped when the socket path cannot be resolved.
 	if socketPath, serr := b.defaultSocketPath(); serr == nil && socketPath != "" {
-		release, pidfile, aerr := acquireBridgeSingleton(socketPath, b.log)
-		if aerr != nil {
-			b.log("bridge singleton: %v (continuing)", aerr)
+		// The record dir comes from the layout ROOT, never from the socket's
+		// own directory: on Windows the socket is a named pipe and has no
+		// directory, which is why no record was ever written there (#6999).
+		var pidfile string
+		if recordDir, derr := bridgeRecordDir(); derr != nil {
+			b.log("bridge singleton: %v (continuing)", derr)
 		} else {
-			defer release()
+			release, path, aerr := acquireBridgeSingleton(recordDir, socketPath, b.log)
+			pidfile = path
+			if aerr != nil {
+				b.log("bridge singleton: %v (continuing)", aerr)
+			} else {
+				defer release()
+			}
 		}
 		stopSignalLogger := installBridgeSignalLogger(pidfile, socketPath, b.log)
 		defer stopSignalLogger()
--- a/internal/cli/mcp_bridge_multisession_test.go
+++ b/internal/cli/mcp_bridge_multisession_test.go
@@ -1,0 +1,429 @@
+package cli
+
+import (
+	"bufio"
+	"crypto/sha256"
+	"encoding/hex"
+	"encoding/json"
+	"fmt"
+	"io"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"runtime"
+	"strconv"
+	"strings"
+	"syscall"
+	"testing"
+	"time"
+
+	"github.com/cajasmota/grafel/internal/process"
+)
+
+// mcp_bridge_multisession_test.go — the #6999 artefact test.
+//
+// #6999 is a cross-process defect: a bridge SIGTERMed OTHER processes. No
+// in-process assertion can observe it, so these tests run real bridge processes
+// and assert on what survives. The children are copies of this test binary
+// named "grafel", because process.PidIsGrafel gates the (now deleted) reap on
+// the executable's name: with a child called "cli.test" the old code would have
+// skipped the reap and the test would have passed against the bug.
+
+const envBridgeChildMode = "GRAFEL_TEST_BRIDGE_CHILD"
+
+// TestHelperBridgeChild is the child entrypoint. It is inert unless the parent
+// sets envBridgeChildMode, so a normal `go test` run never enters it.
+func TestHelperBridgeChild(t *testing.T) {
+	switch os.Getenv(envBridgeChildMode) {
+	case "":
+		t.Skip("child entrypoint; not run directly")
+	case "bridge":
+		// A real bridge over the parent's pipes, using the DEFAULT socket path
+		// (GRAFEL_DAEMON_ROOT points it at the parent's temp root) so the
+		// production singleton path is the one under test.
+		b := &bridge{}
+		_ = b.run(os.Stdin, os.Stdout)
+		os.Exit(0)
+	case "idle":
+		// A live, grafel-named process that is NOT a bridge: the reap target
+		// stand-in. Exits when its stdin closes or after a bounded wait.
+		done := make(chan struct{})
+		go func() { _, _ = io.Copy(io.Discard, os.Stdin); close(done) }()
+		select {
+		case <-done:
+		case <-time.After(60 * time.Second):
+		}
+		os.Exit(0)
+	}
+}
+
+// ── harness ──────────────────────────────────────────────────────────────────
+
+type childProc struct {
+	t    *testing.T
+	cmd  *exec.Cmd
+	pid  int
+	in   io.WriteCloser
+	out  *bufio.Reader
+	wait chan error
+}
+
+// isolateGrafelRoot points the daemon layout at a per-test temp root. Every
+// test in this file MUST call it: without it the bridge resolves the real
+// ~/.grafel and would write (pre-#6999, signal) into the developer's live
+// daemon directory.
+func isolateGrafelRoot(t *testing.T) string {
+	t.Helper()
+	root := t.TempDir()
+	if err := os.MkdirAll(filepath.Join(root, "sockets"), 0o755); err != nil {
+		t.Fatalf("mkdir sockets: %v", err)
+	}
+	if err := os.MkdirAll(filepath.Join(root, "logs"), 0o755); err != nil {
+		t.Fatalf("mkdir logs: %v", err)
+	}
+	t.Setenv("GRAFEL_DAEMON_ROOT", root)
+	t.Setenv("GRAFEL_HOME", root)
+	return root
+}
+
+// grafelNamedTestBinary copies this test binary to a file named "grafel" so
+// children are classified by process.PidIsGrafel exactly as a real bridge is.
+func grafelNamedTestBinary(t *testing.T) string {
+	t.Helper()
+	self, err := os.Executable()
+	if err != nil {
+		t.Skipf("cannot locate test binary: %v", err)
+	}
+	src, err := os.ReadFile(self)
+	if err != nil {
+		t.Skipf("cannot read test binary: %v", err)
+	}
+	name := "grafel"
+	if runtime.GOOS == "windows" {
+		name = "grafel.exe"
+	}
+	// Not t.TempDir(): the binary must outlive per-subtest cleanup ordering.
+	dir, err := os.MkdirTemp("", "grafel-bin")
+	if err != nil {
+		t.Fatalf("tempdir: %v", err)
+	}
+	t.Cleanup(func() { _ = os.RemoveAll(dir) })
+	dst := filepath.Join(dir, name)
+	if err := os.WriteFile(dst, src, 0o755); err != nil {
+		t.Skipf("cannot copy test binary: %v", err)
+	}
+	return dst
+}
+
+// startChild launches a grafel-named child in the given mode and cwd.
+func startChild(t *testing.T, exe, mode, cwd string, extraEnv ...string) *childProc {
+	t.Helper()
+	cmd := exec.Command(exe, "-test.run=^TestHelperBridgeChild$", "-test.timeout=120s")
+	cmd.Dir = cwd
+	cmd.Env = append(os.Environ(),
+		envBridgeChildMode+"="+mode,
+		"GRAFEL_DAEMON_ROOT="+os.Getenv("GRAFEL_DAEMON_ROOT"),
+		"GRAFEL_HOME="+os.Getenv("GRAFEL_HOME"),
+	)
+	cmd.Env = append(cmd.Env, extraEnv...)
+	cmd.Stderr = os.Stderr
+	in, err := cmd.StdinPipe()
+	if err != nil {
+		t.Fatalf("stdin pipe: %v", err)
+	}
+	out, err := cmd.StdoutPipe()
+	if err != nil {
+		t.Fatalf("stdout pipe: %v", err)
+	}
+	if err := cmd.Start(); err != nil {
+		t.Skipf("cannot start child: %v", err)
+	}
+	c := &childProc{t: t, cmd: cmd, pid: cmd.Process.Pid, in: in, out: bufio.NewReader(out), wait: make(chan error, 1)}
+	go func() { c.wait <- cmd.Wait() }()
+	t.Cleanup(func() {
+		_ = in.Close()
+		_ = cmd.Process.Kill()
+		<-c.wait
+	})
+	return c
+}
+
+// serves sends one MCP `initialize` and returns true if a well-formed response
+// comes back — i.e. this bridge is alive AND still doing its job.
+func (c *childProc) serves(id int) bool {
+	c.t.Helper()
+	req := fmt.Sprintf(`{"jsonrpc":"2.0","id":%d,"method":"initialize","params":{}}`+"\n", id)
+	if _, err := io.WriteString(c.in, req); err != nil {
+		c.t.Logf("pid %d: write: %v", c.pid, err)
+		return false
+	}
+	type resp struct {
+		ID     float64         `json:"id"`
+		Result json.RawMessage `json:"result"`
+	}
+	line := make(chan string, 1)
+	go func() {
+		s, err := c.out.ReadString('\n')
+		if err != nil && s == "" {
+			line <- ""
+			return
+		}
+		line <- s
+	}()
+	select {
+	case s := <-line:
+		var r resp
+		if err := json.Unmarshal([]byte(strings.TrimSpace(s)), &r); err != nil {
+			c.t.Logf("pid %d: bad response %q: %v", c.pid, s, err)
+			return false
+		}
+		return int(r.ID) == id && len(r.Result) > 0
+	case <-time.After(10 * time.Second):
+		c.t.Logf("pid %d: no response within 10s", c.pid)
+		return false
+	}
+}
+
+func (c *childProc) exited() bool {
+	select {
+	case err := <-c.wait:
+		c.wait <- err
+		return true
+	default:
+		return false
+	}
+}
+
+// TestBridge_ConcurrentSessionsAllSurviveAndServe is the headline assertion of
+// #6999: N bridges started against ONE daemon socket from different working
+// directories all stay alive and all keep serving. Before the fix, each new
+// bridge SIGTERMed the incumbent and only the newest survived.
+//
+// Each member of the set is asserted individually — a second project and a
+// git worktree of the first are distinct members, not one representative,
+// because those are the two shapes #6999 was reported in.
+func TestBridge_ConcurrentSessionsAllSurviveAndServe(t *testing.T) {
+	root := isolateGrafelRoot(t)
+	exe := grafelNamedTestBinary(t)
+
+	work := t.TempDir()
+	projectA := filepath.Join(work, "project-a")
+	projectB := filepath.Join(work, "project-b")
+	worktreeA := filepath.Join(projectA, ".worktrees", "tsk_1fce9975")
+	for _, d := range []string{projectA, projectB, worktreeA} {
+		if err := os.MkdirAll(d, 0o755); err != nil {
+			t.Fatalf("mkdir %s: %v", d, err)
+		}
+	}
+
+	sessions := []struct {
+		name string
+		cwd  string
+	}{
+		{"project-a", projectA},
+		{"project-b (different project, same socket)", projectB},
+		{"worktree of project-a", worktreeA},
+	}
+
+	var started []*childProc
+	for _, s := range sessions {
+		c := startChild(t, exe, "bridge", s.cwd)
+		// Each bridge must be serving before the next one starts, so that the
+		// next one's startup is what we are testing.
+		if !c.serves(1) {
+			t.Fatalf("session %q did not come up", s.name)
+		}
+		// Every session started BEFORE this one must still be serving.
+		for i, prev := range started {
+			if prev.exited() {
+				t.Fatalf("session %q (pid %d) DIED when session %q started — a bridge terminated another session's bridge (#6999)",
+					sessions[i].name, prev.pid, s.name)
+			}
+			if !prev.serves(100 + i) {
+				t.Fatalf("session %q (pid %d) stopped serving when session %q started (#6999)",
+					sessions[i].name, prev.pid, s.name)
+			}
+		}
+		started = append(started, c)
+	}
+
+	// Every session must own a DISTINCT ownership record naming its own pid.
+	// Under the pre-#6999 per-socket key there is exactly one such file for the
+	// whole machine.
+	socketDir := filepath.Join(root, "sockets")
+	entries, err := os.ReadDir(socketDir)
+	if err != nil {
+		t.Fatalf("read socket dir: %v", err)
+	}
+	recorded := map[int]string{}
+	for _, e := range entries {
+		if !strings.HasPrefix(e.Name(), "mcp-bridge-") || !strings.HasSuffix(e.Name(), ".pid") {
+			continue
+		}
+		p := filepath.Join(socketDir, e.Name())
+		pid, ok := readBridgePID(p)
+		if !ok {
+			t.Fatalf("unreadable bridge record %s", p)
+		}
+		if other, dup := recorded[pid]; dup {
+			t.Fatalf("pid %d recorded twice (%s and %s)", pid, other, p)
+		}
+		recorded[pid] = p
+	}
+	for i, c := range started {
+		if _, ok := recorded[c.pid]; !ok {
+			t.Fatalf("session %q (pid %d) has no ownership record of its own; records=%v — the key is not per-session (#6999)",
+				sessions[i].name, c.pid, recorded)
+		}
+	}
+	if len(recorded) != len(started) {
+		t.Fatalf("want %d distinct bridge records, got %d: %v", len(started), len(recorded), recorded)
+	}
+}
+
+// TestBridge_DoesNotSignalAnotherLiveGrafelProcess is the negative: starting a
+// bridge must leave every other grafel process alone, including one whose pid a
+// stale record names. Pre-#6999 that record was machine-global AND
+// PidIsGrafel matched any grafel process, so this pid was SIGTERMed.
+func TestBridge_DoesNotSignalAnotherLiveGrafelProcess(t *testing.T) {
+	root := isolateGrafelRoot(t)
+	exe := grafelNamedTestBinary(t)
+	cwd := t.TempDir()
+
+	victim := startChild(t, exe, "idle", cwd)
+	if !process.IsAlive(victim.pid) {
+		t.Fatalf("victim pid %d not alive", victim.pid)
+	}
+	if ok, err := process.PidIsGrafel(victim.pid); err != nil || !ok {
+		t.Skipf("cannot classify victim as a grafel process (ok=%v err=%v); "+
+			"without that this test cannot observe a reap", ok, err)
+	}
+
+	// Seed the victim's pid into EVERY plausible record path: the per-socket
+	// path the pre-#6999 code used, and this process's own session path.
+	socket := filepath.Join(root, "sockets", "daemon.sock")
+	seed := []string{
+		legacyPerSocketPidfilePath(socket),
+		bridgeSingletonPath(socket, bridgeSessionID()),
+	}
+	for _, p := range seed {
+		if err := os.WriteFile(p, []byte(strconv.Itoa(victim.pid)+"\n"), 0o600); err != nil {
+			t.Fatalf("seed %s: %v", p, err)
+		}
+	}
+
+	b := startChild(t, exe, "bridge", cwd)
+	if !b.serves(1) {
+		t.Fatal("bridge did not come up")
+	}
+	time.Sleep(300 * time.Millisecond) // the pre-#6999 reap grace was 500ms; SIGTERM itself is instant
+	if victim.exited() {
+		t.Fatalf("a starting bridge terminated another live grafel process (pid %d) — #6999", victim.pid)
+	}
+}
+
+// legacyPerSocketPidfilePath reproduces the pre-#6999 per-socket key so a test
+// can seed the exact file the old code read. It is test-only: production must
+// never derive a path from the socket alone again.
+func legacyPerSocketPidfilePath(socketPath string) string {
+	sum := sha256.Sum256([]byte(socketPath))
+	return filepath.Join(filepath.Dir(socketPath), "mcp-bridge-"+hex.EncodeToString(sum[:6])+".pid")
+}
+
+// TestBridge_SignalledBridgeNamesItselfInTheDaemonLog asserts the second half
+// of the fix end-to-end, through the production wiring: a REAL bridge process
+// (bridge.run, which is what arms the handler) is SIGTERMed the way a
+// pre-#6999 bridge was, and must leave one line in ~/.grafel/logs/daemon.log
+// naming its pidfile, its socket and #6999 — while still exiting 143.
+//
+// This is the artefact that turns the next occurrence into a five-minute
+// triage instead of an excavation: before the fix the reaped bridge died on the
+// default SIGTERM disposition and wrote nothing anywhere.
+func TestBridge_SignalledBridgeNamesItselfInTheDaemonLog(t *testing.T) {
+	if runtime.GOOS == "windows" {
+		t.Skip("SIGTERM delivery is unix-only")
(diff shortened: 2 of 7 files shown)


cajasmota:
## Coordinator mutant on `d766d630a` — one ALIVE, and it is worth grading

Own detached worktree at `/private/tmp/.../coord-mut-7001-q9x/wt` (absolute, not a shared tree — the reviewer has their own). Baseline `go test -count=1 ./internal/cli/` exit 0 immediately before. Restored by `cp` from a SHA-256-verified backup (`34809e9f…` both ways), `git diff HEAD` empty, `git status --porcelain` empty afterwards.

I picked a seam nobody mutated: `appendDaemonLog`, the new writer this PR adds to a file it does **not** own.

### CO-1 — drop `os.O_APPEND` from `appendDaemonLog`

`os.OpenFile(path, os.O_CREATE|os.O_APPEND|os.O_WRONLY, 0o600)` → `os.OpenFile(path, os.O_CREATE|os.O_WRONLY, 0o600)`.

Edit proven landed: the pattern occurred **exactly once**, the hit set was **`[line 81]`**, and `O_APPEND` is absent from the whole file afterwards (`grep -c` = 0). `go vet` exit 0, so this is not a non-compiling mutant.

**Result: ALIVE.** `go test -count=1 ./internal/cli/` exit 0, `ok … 58.878s`.

### It is distinguishable, and the distinguishing input is the production one

I did not stop at the argument. Seeding three realistic prior lines into `logs/daemon.log` and then calling `appendDaemonLog` once, under the mutant:

```
2026/09/08 10:05:49 grafel-mcp-bridge: mcp-bridge (pid 1) terminating on signal terminated: #6999
n v0.3.2
2026/09/02 10:00:02 daemon: pidfile-acquire begin
```

Prior content preserved: **false**. Without `O_APPEND` the write lands at offset 0: one prior line is destroyed outright, the next is left as the corrupt fragment `n v0.3.2`, and only what the new line did not reach survives.

**Why that is not a cosmetic loss here specifically.** `daemonLogPath()` resolves to `daemon.DefaultLayout().LogPath` — the **live daemon's own log**, not a bridge-private file. It is the file I read to classify 251 daemon startups while diagnosing this very issue, and #7002 asks for more instrumentation to be written into it. So this PR's whole justification for `appendDaemonLog` — *"a record that outlives the client's stderr"* — is the same property the mutant destroys, and it destroys the daemon's record, not just the bridge's. A bridge is also not the only writer: the daemon appends concurrently, so offset-0 writes corrupt in both directions.

### Verdict: ungraded, and cheap to grade

Every existing assertion is `strings.Contains` over a log the test wrote from empty, plus one `strings.Count(... ) != 1`. All of those hold when the file's prior contents are destroyed, **by construction** — a `Contains` on a file you alone wrote cannot observe that you clobbered somebody else's line. This is the "assert the artefact, not the counter" shape again, one level down: the artefact asserted is *the new line*, and the property that matters is *the old lines survive*.

**Grading it is about six lines**, inside `TestBridgeSignal_NoticeNamesItsCause`, which already owns this file:

```go
prior := "2026/09/02 10:00:00 daemon: graceful shutdown complete\n"
if err := os.WriteFile(filepath.Join(root, "logs", "daemon.log"), []byte(prior), 0o600); err != nil {
    t.Fatal(err)
}
// … existing handleBridgeSignal call and assertions …
if !strings.HasPrefix(string(body), prior) {
    t.Fatalf("the bridge overwrote pre-existing daemon log content; the log is shared with the daemon:\n%s", body)
}
```

`HasPrefix` rather than `Contains` is the point — `Contains(prior)` still passes if the notice is written *before* it, and the failure mode is precisely a write at the wrong offset.

I'd take it here rather than file it: the writer is new in this PR, so this is the moment it is cheapest, and the file it writes into is the one the next #6999 triage will read.

Nothing else I probed was worth reporting: the `bridgeSessionID` precedence chain and the blank-env case are pinned by `TestBridgeSessionID_DistinctPerSession`, and the `cur == self` guard in the release closure is genuinely equivalent under this suite for a reason I accept — with a per-session key, the overwrite it guards against needs two colliding session ids, and the only consequence is a missing diagnostic record.

Otherwise this is the right fix and the right shape of evidence: `TestBridge_ExitsWhenStdinCloses` is the load-bearing premise and it is pinned directly, and the multisession test asserts each session individually rather than picking a representative.

cajasmota:
## Independent adversarial review — **REQUEST-CHANGES** (small, listed; the code change itself is sound)

Tree: `…/scratchpad/review-7001-k2q7z/wt`, detached at `d766d630a`. Every run under a fresh isolated
`HOME`/`GRAFEL_HOME`/`GRAFEL_DAEMON_ROOT`; the owner's `~/.grafel` was never written to, and no process I
did not start was signalled.

Green before every score, `-count=1`: `./internal/cli/...` ok (58.9s) · `./internal/process/` ok ·
`./internal/daemon/` ok (32.8s) · `./cmd/grafel/` ok (153.7s, fresh isolated root). `go vet ./internal/cli/`
clean. Tree restored by `cp` from a shasum-verified backup, `git diff HEAD` empty, `git status --porcelain`
empty. (One unrelated flake seen once and not reproduced: `TestStatusTotalsCountADoubleRegisteredRepoOnce`,
"fixture broken: row reports 0 rels" — pre-existing, not yours.)

### (a) Does a client-less bridge really always exit? **No — the header comment overstates it.**

I tested it with real processes rather than reading it (`grafel mcp-bridge` built from this branch, spawned
by a Python client holding the stdin pipe):

| scenario | result |
|---|---|
| client `SIGKILL`ed | bridge exited |
| client `SIGTERM`ed | bridge exited |
| client closes bridge stdin, stays alive | bridge exited (zombie until the parent reaps) |
| client leaks the pipe **write end** into a grandchild, then dies | **bridge alive indefinitely** (state `S`) |
| bridge blocked in a daemon call, client `SIGKILL`ed | **bridge alive indefinitely** |

The last row is the one that matters, and it is production-reachable. `callDaemon` →
`rpcClient.Call` (`net/rpc`) has **no deadline**: against a daemon that accepts the connection and never
replies — a wedged/swapping daemon, which this project has on record — the bridge never returns to its read
loop, so it never observes the EOF. Repro: unix listener that `accept()`s and never answers, one `tools/call`,
`kill -9` the client → the bridge was still `S` 30 s later, and would be there forever.

So `mcp_bridge_singleton.go:28-31` — *"A bridge whose client is gone exits on its own… Nothing external is
needed to collect it, and TestBridge_ExitsWhenStdinCloses pins that"* — asserts more than the test observes.
`TestBridge_ExitsWhenStdinCloses` runs an **idle** in-process bridge over an `io.Pipe`; it pins EOF-while-idle,
not "whose client is gone". This is the dominant defect class on this board, and it is load-bearing here: the
same over-general premise in a header comment is what governed the old `kill`.

That said, the trade is **not** kill-for-leak in the general case, and I do not think it blocks the direction:
in every ordinary client-death shape the bridge does collect itself, and in the wedged-daemon shape the old
reap only cleaned up when a *new* bridge started — the very event that caused #6999. **Ask:** scope the comment
to what is pinned (EOF while idle / not blocked in a daemon call), name the blocked-in-`Call` exception, and
file the follow-up for a deadline or stdin-close cancellation on `callDaemon` (natural neighbour of #6722).

### (b) Does anything still signal? **No.** Confirmed by construction.

`grep` over the non-test bridge sources finds no `process.Kill`, no `.Signal(`, no `Process.Kill`; the only
signal API left is `signal.Notify` (**receive**). `reapPriorBridge` and `bridgeReapGrace` are gone repo-wide.
The record is read only to decide whether to emit a log line, and nothing branches to an action on it.
Secondary hazard **confirmed real on `c6a99d75f`**: `isLiveBridge` returned `true` on a `PidIsGrafel` error,
and `exeBaseNameIsGrafel` is `strings.Contains(b, "grafel")` — the daemon's own basename matches, so a recycled
pid held by `grafel serve` was a valid SIGTERM target. Eliminated by construction here, and additionally fixed.

Live three-session artefact check with real processes on one socket (not just the test): a, then b, then c —
after each start every earlier bridge was still alive **and still answered `initialize`**, and the sockets dir
held **three distinct** records naming exactly the three bridge pids. The headline holds outside the test binary.

### (c) Session-id fallback collision-safe? **Yes.**

`env: → stdin:<dev>-<ino> → pid:<n>`. Two sessions that both fall through to *pid* are distinct by
construction. A collision (shared `GRAFEL_MCP_SESSION`, both stdins on `/dev/null`, a shared tty) costs one
overwritten record and nothing else, because no signal is derived from it — verified in (b). One nit below (M-A).

### (d) Skipping the ratchet? **Justified — agree.**

The diff is confined to `internal/cli/mcp_bridge*`: process lifecycle, pidfile naming, signal handling. Nothing
in it is reachable from extraction, resolution or graph assembly, so the ratchet has no surface to move. I ran
`./cmd/grafel/` myself (the whole-graph digest pin) — green — which is the check that would actually have caught
a smuggled extractor effect. Do not spend the ~40 minutes.

### Mutants — mine, permissive direction, seams the author did not touch

Each applied with an asserted occurrence count of **1**, enclosing function printed, `go vet` clean, green
baseline immediately before, `-count=1`, restored by `cp` from a shasum-verified backup with `git diff HEAD` empty.

| # | Mutant (enclosing func) | Verdict |
|---|---|---|
| **M-A** | `acquireBridgeSingleton` — release closure removes the record **unconditionally** (drop the `cur == self` guard) | **ALIVE — distinguishable and ungraded.** Reachable whenever two bridges share a session id: bridge A's exit deletes bridge B's live record. The comment *"Only remove the record if it still names us — a newer bridge for the same session may already own it"* is asserted by nothing. Assertion: acquire A, acquire B on the same (socket, session), release A, require the file to still exist **and read B's pid**. I priced it at ~8 lines. |
| **M-B** | `stdinIdentity` — return device only, dropping the inode | **DEAD** — `TestBridge_ConcurrentSessionsAllSurviveAndServe`: *session "project-a" (pid …) has no ownership record of its own … the key is not per-session (#6999)*. |
| **M-C** | `appendDaemonLog` — drop `os.O_APPEND` (`O_CREATE\|O_WRONLY`) | **ALIVE — distinguishable and ungraded, and this one has teeth.** The notice is then written at **offset 0 of `~/.grafel/logs/daemon.log`**, overwriting the daemon's own log head, and concurrent bridge deaths overwrite each other. Reachable on any bridge SIGTERM. The existing assertions count lines in a log the test itself created empty, so append-ness is unobserved. Assertion: seed `daemon.log` with a prior line, call `appendDaemonLog`, require the prior line to **still be present** and the notice to follow it. ~6 lines. |
| **M-D** | `(b *bridge).run` — arm the signal logger only when `acquireBridgeSingleton` succeeded | **ALIVE but barely production-reachable** — needs an unwritable sockets dir. Worth a line of comment at most; I would not gate on it. |

I did not re-run M1–M6.

### Also checked, no finding

- Signal handler cannot hang on the log write in any realistic shape: `appendDaemonLog` opens a regular file
  and returns on any error; the only blocking shape is a FIFO/hung mount at `logs/daemon.log`, which nothing
  creates. It does not fire on a normal exit — `stopSignalLogger` disarms via `signal.Stop` + `close(done)`.
- `os.Exit` in the handler skips `release()` and `closeRPCClient()`, leaving a stale record — but the pre-change
  default disposition skipped them too, so this is not a regression.
- Concurrent writers to `daemon.log` are fine **as written** (a single short `O_APPEND` write) — see M-C for
  why that is currently an accident rather than a pinned property.

### What I need before merge

1. Scope the `mcp_bridge_singleton.go` self-collection claim to what is pinned, name the blocked-in-`Call`
   exception, and file the `callDaemon` deadline follow-up.
2. Grade **M-C** (`O_APPEND`) — ~6 lines.
3. Grade **M-A** (release guard) — ~8 lines.

None of these change the fix. The direction — delete the kill rather than narrow it — is right, nothing signals
any more, the per-session key demonstrably holds with real processes, and the secondary daemon-kill hazard was
real and is gone.

cajasmota:
## Second coordinator mutant, on `36da881db`

My first was scored on `d766d630a`; this round added production code (`bridgeRecordDir`, a changed `acquireBridgeSingleton` signature, a `MkdirAll`), so a verdict on the old head does not carry. Own detached worktree at `/private/tmp/.../coord-mut-7001-q9x/wt` (absolute). Baseline `go test -count=1 ./internal/cli/` exit 0 immediately before. Restored by `cp` from a SHA-256-verified backup (`0f6d9458…` both ways), `git diff HEAD` empty and `git status --porcelain` empty afterwards.

### CP-1 — change the record directory name: `filepath.Join(layout.Root, "sockets")` → `"bridges"`

I picked this because it targets a **prose claim that is load-bearing for cleanup**, which is the defect class this board keeps shipping:

> `sockets/` is already one of grafel's scratch dirs (internal/install: `grafelScratchDirs`), so `uninstall --purge` still cleans these up.

I verified the claim itself first rather than relaying it — `internal/install/uninstall.go:154` is `var grafelScratchDirs = []string{"store", "docs", "backups", "logs", "sockets"}`, so the sentence is true. The question was whether anything *pins* it: if the directory name were free to drift, records would silently stop being purged and nothing would fail.

Edit proven landed: exactly **1** occurrence, hit set **`[line 108]`**, `go vet` clean.

**Result: DEAD**, and killed three times over:

- `TestBridgeRecordDir_IsAWritableDirectoryOnEveryPlatform` — `record dir = ".../bridges", want ".../sockets" (it must come from Layout.Root)`
- `TestBridge_ConcurrentSessionsAllSurviveAndServe` — same comparison at `mcp_bridge_multisession_test.go:262`
- `TestBridge_SignalledBridgeNamesItselfInTheDaemonLog` — caught it through the **artefact**: the daemon-log line no longer names a `pidfile=.../sockets/mcp-bridge-` path

**The reason it died is the part worth naming.** All three assert against an **independent literal** `"sockets"`, not against `bridgeRecordDir()`'s own return value. A test that ranged over the production value would have followed the mutation and passed — that exact mistake shipped here before and was caught only by re-scoring. This is the right construction, and `internal/install/` stayed green under the mutant, which confirms the purge coupling is real but is graded on the `cli` side only.

I expected this one to be ALIVE. It was not, and the code is better than my hypothesis.

### On the round's other work

`M-A` and `M-C` are graded and re-scored DEAD by the author; `M-C` is the mutant the reviewer and I had each found independently. Ask 3's narrowing is the right outcome — the header now names the three verified self-collecting shapes and the one exception with its cause, and `TestBridge_ExitsWhenStdinCloses`'s doc comment says it pins EOF-while-idle only. An honest narrow claim is worth more than a broad one.

Ask 4 is answered without a Windows skip, which is what I cared about: `TestBridgeRecordDir_IsAWritableDirectoryOnEveryPlatform` deletes the directory and writes through production `acquireBridgeSingleton` against the real `DefaultLayout().SocketPath`, and its doc comment states the honest scope — the *derivation* is only distinguishable where root and socket dir differ, i.e. Windows; what it grades everywhere is that the directory is created and written through. That distinction being written down is the difference between a pin and a decoration.

**Merging on green.** The Windows leg is the only real check on the record-dir fix — `GOOS=windows` cross-`vet` fails in this tree because the tree-sitter cgo bindings exclude every file under a cross build (pre-existing, unrelated to this diff, and worth its own issue). So I am waiting for that leg specifically, not for the rollup.

cajasmota:
## Delta review `d766d630a..36da881db` — **APPROVE**

Independent second pass, **delta only**. Tree: `…/scratchpad/review-7001-d2/wt`, detached at `36da881db`.
Every run under a fresh isolated `HOME`/`GRAFEL_HOME`/`GRAFEL_DAEMON_ROOT`; the owner's `~/.grafel` was never
written to; no process I did not start was signalled. Green baseline before each score, `-count=1`:
`./internal/cli/` ok (63.3s) · `./internal/process/` ok · `./internal/daemon/` ok (33.9s) ·
`./cmd/grafel/` ok (155.1s, fresh isolated root). `go vet ./internal/cli/` and `gofmt -l internal/cli` clean.
Restored by `cp` from a shasum-verified backup after every mutant, `git diff HEAD` and `git status --porcelain`
both empty.

### `bridgeRecordDir()` — correct on both platforms, and failure-tolerant

`Layout.Root` is a real writable directory in every branch of both `DefaultLayout`s:
`paths_unix.go` → `$GRAFEL_DAEMON_ROOT` (non-empty by its own `root != ""` check) else `$HOME/.grafel`;
`paths_windows.go` → `$GRAFEL_DAEMON_ROOT` else `%APPDATA%\grafel` else `UserHomeDir()\grafel`, with the
`UserHomeDir()` error returned rather than swallowed. Neither can return a `Layout` with an empty `Root` and a
nil error. When the layout itself fails (unix `selectSocketPath` over the 104-byte limit) `bridgeRecordDir`
returns the error — and `b.defaultSocketPath()` has already failed by then, so the whole block is skipped
consistently.

`MkdirAll` failure verified with a real bridge process against a `0500` root:

```
mode=readonly  served=yes
stderr: bridge singleton: create bridge record dir …/sockets: permission denied (continuing)
```

The bridge answered `initialize` normally. Happy path also verified end to end: the record appears at
`<root>/sockets/mcp-bridge-<hash>.pid` naming the live bridge pid, and is removed when the client closes stdin.

### The Windows claim — **confirmed, two ways**

Per **#7005** I cannot cross-compile or execute Windows in this tree, so I reasoned from source **and** from CI,
and I am stating that limitation explicitly.

*Source.* `paths_windows.go` sets `SocketDir: ""` and `SocketPath: transport.WindowsPipeName(root)`, i.e.
`\\.\pipe\grafel-<hash>`. `d766d630a^` (`c6a99d75f`) derived the record path as
`filepath.Dir(socketPath)`. I walked Go's own `filepathlite.volumeNameLen` (windows): the `\\.` branch treats
the next component as part of the volume, so the volume of `\\.\pipe\grafel-abc` is `\\.\pipe` and `Dir` yields
`\\.\pipe\`. The record path was therefore `\\.\pipe\mcp-bridge-<hash>.pid` — inside the pipe namespace, where
`os.WriteFile` cannot create anything and `os.ReadFile` finds nothing. So `readBridgePID` **never** returned a
prior pid on Windows, `isLiveBridge` was never consulted, and `reapPriorBridge` could never fire there.

*CI, independently.* The Windows leg at `d766d630a` — which is exactly this code — reports
`records=map[]` (**empty**, not wrong) while having already passed the survival-and-serve assertions for all
three sessions. That is a direct observation of the write failing on Windows, and of nothing being reaped. It
does not share a premise with my source reading.

**Conclusion: true — Windows never wrote the record, and Windows users were never hit by #6999.** Worth carrying
in the PR body: the blast radius of #6999 is unix/macOS only.

### `TestBridgeRecordDir_IsAWritableDirectoryOnEveryPlatform` — not vacuous, scope claim accurate

It is not vacuous on unix: **M-F below kills it there.** It `RemoveAll`s the dir and then writes through
production `acquireBridgeSingleton` against the real `DefaultLayout().SocketPath`, so the create-and-write-through
half is graded on every platform, unskipped.

The stated scope is accurate, and I checked it rather than took it: under `GRAFEL_DAEMON_ROOT` the unix socket
already lives at `<root>/sockets`, so `filepath.Dir(SocketPath)` and `filepath.Join(Root, "sockets")` coincide
and no unix assertion can separate the two derivations. The comment says exactly that. See M-E.

I did not re-derive the `grafelScratchDirs`/`uninstall --purge` question — you covered it.

### Mutants — new code only

Applied with an asserted occurrence count of **1**, enclosing function named, hit set asserted after the edit,
`go vet` clean, green baseline immediately before, `-count=1`, restored by `cp` from a shasum-verified backup
with `git diff HEAD` empty.

| # | Mutant (enclosing func) | Verdict |
|---|---|---|
| **M-E** | `bridgeRecordDir` → `return filepath.Dir(layout.SocketPath), nil` (revert the derivation) | **ALIVE on unix — DEAD on the Windows leg.** `./internal/cli/` ok (62.0s) with it applied, precisely as the test's scope comment predicts. It is not ungraded, though: this mutant **is** `d766d630a`, and that commit's Windows leg failed with `records=map[]`. The kill is real, it just lives on a platform this machine cannot run — which is why `ci:full` is load-bearing on this PR. |
| **M-F** | `acquireBridgeSingleton` — delete the `os.MkdirAll(recordDir, 0o700)` block | **DEAD** — `TestBridgeRecordDir_IsAWritableDirectoryOnEveryPlatform`: *acquire against the real layout socket … failed: … no such file or directory — the ownership record does not exist on this platform*. This is the proof the new test is not vacuous on unix. |
| **M-G** | `bridgeRecordDir` — delete the `strings.TrimSpace(layout.Root) == ""` guard | **ALIVE, essentially not production-reachable.** No `DefaultLayout` branch on either platform returns an empty `Root` with a nil error. The only input that reaches it is a whitespace-only `GRAFEL_DAEMON_ROOT` (`"   "`), a test/agent-isolation variable; without the guard the record would go to a relative `sockets/` under the cwd. Cheap defensive guard, not worth an assertion. |

### The round-1 asks

All three land, and I re-checked the two gradings are not decorative:
`TestBridgeSingleton_ReleaseLeavesANewerOwnersRecord` asserts the **artefact** (the file exists **and** still
reads the newer pid), not a counter. The `O_APPEND` grade correctly uses `HasPrefix(prior)` rather than
`Contains` — `Contains` cannot distinguish a notice written *in front of* the prior line, and the comment says
why. The header comment now names the three verified self-collecting shapes and the blocked-in-`Call` exception
with its cause and #7003, and `TestBridge_ExitsWhenStdinCloses`'s doc states its narrow scope. Nothing overclaims.

**APPROVE on the delta.** No new production risk found; ship it with `ci:full` so the Windows leg that kills M-E
actually runs. Not merged, not pushed.

Issue #6996, opened 2026-09-08
sql-language custom gates in rust/ and nim/ are unreachable: `sql` routes to `custom_js_` only, so diesel/sqlx/ormin migration parsing has never run


cajasmota:
Found while enumerating #6978 (see #6995 for the full table and the derivation).

`customPrefixForLanguage` routes language `"sql"` to `custom_js_` **only**. Three extractors gate on `.sql` from packages that prefix cannot reach, so their code has never run on a real index — only in their own unit tests, which call `Extract` directly:

- `internal/custom/rust/diesel.go` `Extract` — the whole `.sql` migration branch (`extractSQLMigration`), i.e. `up.sql`/`down.sql` CREATE TABLE + REFERENCES parsing.
- `internal/custom/rust/sqlx_rbatis.go` `isSqlxMigrationFile` — the predicate is **always false** in production, so every sqlx/rbatis migration entity is unreachable.
- `internal/custom/nim/ormin_orm.go` `Extract` — ormin's SQL-DSL half.

Unlike the `.razor` sites in the same enumeration, **this needs no grammar work**: `.sql` has a tree-sitter grammar, so `TSTree` is non-nil and every dispatch path would deliver the file. The only missing piece is routing — an `extraCustomPrefixesForLanguage["sql"]` entry listing `custom_rust_` and `custom_nim_`.

**Why it is not bundled into #6995.** Turning these on changes graph output for every Rust and Nim repo that carries migrations, and the other direction is real: `custom_js_` already claims `.sql` for Prisma/Drizzle, so a `.sql` file would then be seen by three extractor families at once. That is a measured decision (entity/edge deltas on a corpus repo with migrations), not a side effect of an enumeration PR.

`TestCustomExtractorGatesAreReachable6978` lists all three; repairing them requires deleting their lines from `knownUnreachableGates`, which is the intended workflow.

Refs #6978, #6995.

Issue #2320, opened 2026-05-26
[tech-debt][PR-2312-followup] Extractor interface lacks Config channel — env-var toggles accumulating ad-hoc


cajasmota:
## Context
PR #2312 (issue #2284) gated markdown heading emission behind `ARCHIGRAPH_MARKDOWN_EMIT_HEADINGS`. This is the THIRD env-var toggle on an extractor (mirroring `ARCHIGRAPH_INCREMENTAL_REINDEX`, `ARCHIGRAPH_INCREMENTAL_MAX_FILES` in `incremental.go`, plus JavaScript measure-mode). The extractor interface (`internal/extractor/extractor.go`) has no Config channel.

## Fix
Add `Config` field on `FileInput` (or a registry-level config) so per-extractor toggles can be set via the indexer's config file rather than process env. Migrate the three existing env-var toggles to the new mechanism — keep env-var read as a fallback for backward compat.

Effort: M. Priority: medium — pattern is accumulating; the next extractor that needs a toggle will be the fourth ad-hoc env var.

Refs: PR #2312, also affects `internal/extractors/incremental.go` and JS measure-mode.

PR #7004, opened 2026-09-08
fix(#6997): the daemon's extract subprocess dispatched custom extractors under neither guard — align it with the other two sites and pin the agreement

### LINKED ISSUE #6997: The three RunCustomExtractors call sites disagree: subproc.go has neither the TSTree guard nor the custom-extractor gate, so 6 grammarless custom packages run on exactly one path
Found while enumerating #6978 (see #6995).

There are three call sites of `extractors.RunCustomExtractors`, and they do not agree on the `TSTree` guard:

- `cmd/grafel/index.go:4179` — `CustomExtractorsEnabled(&cfg) && file.TSTree != nil`
- `internal/extractors/incremental.go:1126` — `CustomExtractorsEnabled(cfg) && input.TSTree != nil`
- `internal/daemon/extract/subproc.go:364` — **neither guard**: it dispatches under `runExtract` alone, with no `TSTree` check and no `CustomExtractorsEnabled` check.

Consequence: six `internal/custom/` packages target languages with **no tree-sitter grammar** — `clojure`, `crystal`, `dart`, `erlang`, `fsharp`, `nim` — so their files always have `TSTree == nil`. Those packages therefore run on the daemon subprocess path and **never** on the in-proc or incremental paths. The same asymmetry means the subprocess path also ignores the custom-extractor gate that #6966 keeps default-OFF.

The `TSTree != nil` guard is load-bearing where it exists (`index.go:4173` documents why: content-only extractors would emit entities for files that failed to parse), so the fix is not "delete it" — it is to decide which of the two behaviours is correct and make all three sites agree, then pin the agreement.

Not a user-facing defect today, since the in-proc gate is default-OFF; it is a divergence between the paths measurements are taken on.

Refs #6978, #6995, #6966.

cajasmota:
Closes #6997.

Measured in `.../scratchpad/impl-6997-k7m2p/wt`, macOS/APFS, on `81f99d691`.

## The fork the issue implies does not exist, and the measurement is why

"Make all three sites agree" reads like a choice between removing custom
entities from every user's daemon graph and leaving the divergence documented.
Both halves of that framing are wrong, and tracing the call chain rather than
the call site is what shows it:

- `internal/daemon/extract/subproc.go` is reached only via `extract.Coordinate`,
  whose **single caller** is `cmd/grafel/index.go` behind `subprocExtract()` —
  `GRAFEL_SUBPROC_EXTRACT`, **default-OFF**.
- The daemon's default path is `GRAFEL_SUBPROCESS_INDEXER` (default-ON), which
  forks `grafel index-internal` and extracts **in-process** — through the
  *guarded* site. It never enters this file.

So the unguarded dispatch was never on any default graph. It was a disagreement
between an opt-in path and the two paths every custom-lane measurement is taken
on — and `GRAFEL_SUBPROC_EXTRACT`'s own stated acceptance criterion
(`cmd/grafel/extract.go`) is *"the in-process path stays the default until
benchmarks + quality fixtures confirm byte-identical output."* Aligning is not a
policy choice here; it is that rollout's contract.

**This PR does not touch #6966.** It makes this path *obey* the gate instead of
answering it locally.

## Blast radius, gate stated on every figure

Three arms per repo: the `81f99d691` binary (which ignores the gate), this
branch with the gate OFF, this branch with `GRAFEL_INPROC_CUSTOM_EXTRACTORS=1`.
All arms `GRAFEL_SUBPROC_EXTRACT=1` (nothing here is reachable otherwise),
`GOMAXPROCS=3`, `GRAFEL_SUBPROC_CONCURRENCY=3`, isolated `HOME`/`GRAFEL_HOME`/
`GRAFEL_DAEMON_ROOT`, corpora copied out of the read-only tree.

| repo | `81f99d691` | this branch, gate OFF | this branch, gate ON |
|---|---|---|---|
| django-realworld | 456 ent / 1463 rel | **371 / 1298** | 456 / 1463 |
| laravel-routing | 1159 / 3715 | **1108 / 3609** | 1159 / 3715 |
| dart-samples | 884 / 1877 | 884 / 1877 | 884 / 1877 |

**Gate-ON is identical to the old behaviour on all three**, so the entire delta
is the gate; the `TSTree != nil` guard removed **zero** entities here. That zero
is corpus-relative and not a claim that the guard is inert — see the nim
fixture below, where it removes real output.

What the gate removes, by kind:

| repo | kinds |
|---|---|
| django-realworld | `SCOPE.Operation` −51, `SCOPE.Datastore` −19, `SCOPE.Schema` −13, `SCOPE.Config`/`External`/`Pattern` −1 each, `SCOPE.Component` +1 (a merge that no longer supersedes) — 85 net |
| laravel-routing | `SCOPE.Pattern` −28, `SCOPE.Operation` −17, `SCOPE.Component` −5, `SCOPE.Process` −1 — 51 net |

Anyone on `GRAFEL_SUBPROC_EXTRACT=1` who wants those entities back sets
`GRAFEL_INPROC_CUSTOM_EXTRACTORS=1` — the same switch the in-process path has
always required.

## The named pin did not observe what it was named for

The issue and the in-proc comment both cite
`TestInProcCustomExtractorsNilTreeGuardIsLoadBearing`. It asserts only that
`RunCustomExtractors` *emits with a nil tree*, and `t.Skip`s if it does not. It
never observes any call site applying the guard, so **every one of the three
sites could have dropped it and that test would still be green.** The premise it
pins is true and worth keeping; the guard was unpinned. Both guards are now
graded separately, at all three sites.

## What changed

- `subproc.go` dispatches under `runCustom && file.TSTree != nil`, with
  `runCustom := runExtract && extractors.CustomExtractorsEnabled(customCfg)` —
  the same expression as the other two sites.
- The **programmatic** half of the gate (`WithCustomExtractors`) now crosses the
  process boundary: `CoordinatorConfig.CustomExtractors` → `--custom-extractors`
  → `SubprocessOptions.CustomExtractors`. The **env** half needs no plumbing —
  children get `os.Environ()`. Only the opt-in is forwarded, mirroring the
  in-proc stamp: a forwarded `false` would *suppress* a set env var in the
  child, since config beats env in both directions (#2320).

## Pinning the agreement

`TestRunCustomExtractorsCallSitesCarryBothGuards6997` (in `internal/extractors`,
the package that owns the dispatcher) parses **every non-test `.go` file in the
repo**, collects the enclosing `if` conditions that actually dominate each
`RunCustomExtractors` call, and expands single assignments within the enclosing
function so `subproc.go`'s `runCustom` resolves rather than reading as
unguarded (which would push the next author to inline a long expression into a
hot loop to satisfy a test). Shape borrowed from #6995's `gateKey` scan:

- Set equality on the three sites **in both directions** — a fourth site fails,
  and deleting or renaming one fails until the frozen list is edited. The list
  is hand-written, not derived from the scan it grades.
- Each guard is asserted and reported **separately**; `A && B` scored as a unit
  grades neither half.
- An empty scan is a `Fatal`, not a pass.
- **Coverage (review finding (b)).** The site walk descends only into
  `*ast.FuncDecl` and matches only a direct callee, so two fourth sites slipped
  past it with the whole suite green: a package-level `var x = func(...)` (R1)
  and a call through a function **value** (R1b). R1b is the shape this repo
  already uses — `extract.Run` reaches production only as
  `Hooks{RunExtract: runExtractSubprocess}`. Every mention of the dispatcher is
  now counted by an **independent** traversal (its own walk, its own parse, a
  flat ident visitor) and must be accounted for by a collected site; a count
  derived from the collection it grades could not detect an omission in it. The
  failure names the offending `file:line`. **R1, R1b — now DEAD** (4 mentions
  vs 3 sites), along with V1 (mention scan sees no files → the vacuity `Fatal`
  fires) and V2 (stop excluding the declaration's own name → the count is real).
  M4 and M5 re-scored DEAD.

Enclosing-function names, not line numbers, so unrelated edits above a site
never force a re-baseline.

## Mutants — all DEAD, scored with `-count=1`, each edit proved by an exact
occurrence count and the line it landed on

| # | mutant | result |
|---|---|---|
| M1 | the true pre-fix state: `runCustom := runExtract` + `if runCustom {` | **DEAD** — scanner reports *both* guards missing; all three behavioural tests RED |
| M2 | drop **only** `TSTree != nil` | **DEAD** — scanner reports only the TSTree finding; only `...RequireAParseTree6997` RED |
| M3 | drop **only** the gate | **DEAD** — scanner reports only the gate finding; only the gate + propagation tests RED |
| M4 | delete one entry from the frozen site list | **DEAD** |
| M5 | add a fourth production dispatch site, unguarded | **DEAD** — named in the failure output |
| M6 | `subprocCustomExtractorsOptIn` always returns `nil` | **DEAD** |
| M7 | *(permissive)* it always returns `&true` | **DEAD** |
| M8 | coordinator stops forwarding an explicit `false` | **DEAD** |
| M9 | *(permissive)* `index.go` forwards `subprocCustomExtractorsOptIn(true)` | **initially ALIVE**, then graded — see below |

M2 and M3 are the part-by-part grading: each fails a **disjoint** set of
assertions, so neither guard is riding on the other.

**M9 was alive and production-reachable.** Its distinguishing input is measured
above (456 vs 371 on django-realworld). Nothing in this repo exercises the
`subprocExtract()` branch of `Index()` — it forks `os.Executable()`, which
inside a test binary is the test binary, and that path takes no `BinaryPath`
override. Rather than plumb an override into production for testability, the
forwarded *expression* is asserted with the same AST instrument used on the
dispatch sites (`TestSubprocCoordinatorGetsTheIndexersOwnOptIn6997`, ~40 lines);
it also fails if a second `CoordinatorConfig` literal appears in `index.go`.

## Fixtures: what varies, what is held constant

- `TestSubprocCustomExtractorsHonourTheGate6997` — one file, four arms varying
  **only** (env half × config half): unset/nil, `1`/nil, unset/`true`,
  `1`/`false`. Same fixture, same path, so nothing but the gate can move the
  output. The negative arms are checked on **identity against the gate-off
  baseline**, not on a count (a swap keeps the count), and every arm must be a
  superset of the base entities.
- `TestSubprocCustomExtractorsRequireAParseTree6997` — gate held ON, so only the
  `TSTree` guard can suppress. The fixture is a **real production shape**: nim
  has a registered `custom_nim_` prefix and **no tree-sitter grammar**, so
  `TSTree` is nil for every `.nim` file in every repo, and
  `internal/custom/nim/norm_orm.go` is content-only. Same for clojure, crystal,
  dart, erlang, fsharp. A **positive control** fails the test if direct dispatch
  ever stops emitting for that content, so the assertion cannot go vacuous.
- `TestCoordinateForwardsTheProgrammaticCustomGate6997` — forks a built binary;
  the two tests above call `Run` in-process and would pass just as well if
  nothing were ever forwarded.

## Why CI is the only opinion on a diff of this shape

Per **#7005** this tree cannot be cross-compiled for Windows at all —
`GOOS=windows go vet ./internal/...` fails because cross-builds default to
`CGO_ENABLED=0` and every tree-sitter grammar package drops out. For a change
that forks a binary and plumbs a flag across a process boundary, CI is not a
second opinion, it is the only one; hence `ci:full` on this PR.

## Verification

- `go test -count=1 ./internal/daemon/extract/ ./internal/extractors/` — green
  (green baseline taken before each mutant score).
- `go test -count=1 ./cmd/grafel/` — green, with a **fresh** `HOME` +
  `GRAFEL_HOME` + `GRAFEL_DAEMON_ROOT`.
- `scripts/quality/run.sh --ratchet` — *"38 gated fixtures held their baseline
  (29 at 100% must-have recall)"*. **No baseline moved**, so nothing was
  re-baselined.
- `gofmt -l`, `go vet` on all touched packages — clean.

## Not in scope

#6996 (routing `sql` to `custom_rust_`/`custom_nim_`) is untouched. #6966's
default is untouched — this PR gives that decision one place to be made instead
of two.

### DIFF
--- a/cmd/grafel/custom_gate_optin_6997_test.go
+++ b/cmd/grafel/custom_gate_optin_6997_test.go
@@ -1,0 +1,101 @@
+package main
+
+import (
+	"go/ast"
+	"go/parser"
+	"go/printer"
+	"go/token"
+	"strings"
+	"testing"
+)
+
+// #6997 — subprocCustomExtractorsOptIn is what decides whether the
+// custom-extractor gate crosses into the extract subprocesses, and it must
+// forward ONLY the programmatic opt-in.
+//
+// Both error directions are real and neither is caught anywhere else:
+//
+//   - Always returning a non-nil false would send `--custom-extractors false`
+//     to every child on a default index. Config beats env, so that would
+//     SUPPRESS a set GRAFEL_INPROC_CUSTOM_EXTRACTORS in the child while the
+//     in-process path honoured it — the same disagreement between paths this
+//     issue exists to close, reintroduced from the other end.
+//   - Always returning a non-nil true would run ~340 custom extractors for
+//     operators who opted into GRAFEL_SUBPROC_EXTRACT and nothing else,
+//     deciding #6966's default-OFF question for them locally.
+func TestSubprocCustomExtractorsOptIn6997(t *testing.T) {
+	if got := subprocCustomExtractorsOptIn(false); got != nil {
+		t.Errorf("no programmatic opt-in must forward NOTHING (nil), so the child "+
+			"falls back to the inherited env exactly as the in-process path does; got %v", *got)
+	}
+	got := subprocCustomExtractorsOptIn(true)
+	if got == nil {
+		t.Fatal("WithCustomExtractors must survive the process boundary; got nil")
+	}
+	if !*got {
+		t.Errorf("opt-in forwarded as %v, want true", *got)
+	}
+}
+
+// TestSubprocCoordinatorGetsTheIndexersOwnOptIn6997 grades the one line the
+// helper test above cannot reach: WHAT Index() hands the coordinator.
+//
+// The subprocExtract() branch of Index() is not exercised by any test in this
+// repo — it forks `grafel extract` via os.Executable(), which inside a test
+// binary is the test binary, and there is no BinaryPath override on that path.
+// So `CustomExtractors: subprocCustomExtractorsOptIn(true)` — an unconditional
+// opt-in that would run ~340 custom extractors for every operator who set
+// GRAFEL_SUBPROC_EXTRACT=1 — survives every behavioural suite. Measured
+// distinguishing input, macOS/APFS: django-realworld indexes to 456 entities
+// with the gate on and 371 with it off.
+//
+// Rather than plumb a binary override into production for testability, the
+// forwarded EXPRESSION is asserted, the same instrument
+// TestRunCustomExtractorsCallSitesCarryBothGuards6997 uses on the dispatch
+// sites. A constant here fails; so does forwarding some other flag.
+func TestSubprocCoordinatorGetsTheIndexersOwnOptIn6997(t *testing.T) {
+	fset := token.NewFileSet()
+	f, err := parser.ParseFile(fset, "index.go", nil, 0)
+	if err != nil {
+		t.Fatalf("parse index.go: %v", err)
+	}
+
+	const want = "subprocCustomExtractorsOptIn(i.customExtractors)"
+	found := 0
+	ast.Inspect(f, func(n ast.Node) bool {
+		lit, ok := n.(*ast.CompositeLit)
+		if !ok {
+			return true
+		}
+		sel, ok := lit.Type.(*ast.SelectorExpr)
+		if !ok || sel.Sel.Name != "CoordinatorConfig" {
+			return true
+		}
+		found++
+		var got string
+		for _, el := range lit.Elts {
+			kv, ok := el.(*ast.KeyValueExpr)
+			if !ok {
+				continue
+			}
+			if key, ok := kv.Key.(*ast.Ident); ok && key.Name == "CustomExtractors" {
+				var sb strings.Builder
+				if err := printer.Fprint(&sb, fset, kv.Value); err != nil {
+					t.Fatalf("print: %v", err)
+				}
+				got = sb.String()
+			}
+		}
+		if got != want {
+			t.Errorf("extract.CoordinatorConfig.CustomExtractors = %q, want %q.\n"+
+				"The custom-extractor gate forwarded to the extract subprocesses must be "+
+				"the INDEXER's own opt-in; a constant decides #6966's default-OFF question "+
+				"for whoever set GRAFEL_SUBPROC_EXTRACT.", got, want)
+		}
+		return true
+	})
+	if found != 1 {
+		t.Fatalf("found %d extract.CoordinatorConfig literals in index.go, want exactly 1; "+
+			"a second call site would need the same forwarding and this test would not see it", found)
+	}
+}
--- a/cmd/grafel/extract.go
+++ b/cmd/grafel/extract.go
@@ -86,6 +86,21 @@ func subprocBatchSize() int {
 	return n
 }
 
+// subprocCustomExtractorsOptIn mirrors, for the subprocess-extract path, the
+// stamp classifyAndReadWithProgress applies to ExtractorConfig: only the
+// programmatic OPT-IN travels. An unset WithCustomExtractors must not be
+// forwarded as an explicit `false`, because Config beats env in both
+// directions — a forwarded false would suppress a set
+// GRAFEL_INPROC_CUSTOM_EXTRACTORS in the child while the in-process path
+// honoured it, which is the very disagreement #6997 is closing.
+func subprocCustomExtractorsOptIn(optedIn bool) *bool {
+	if !optedIn {
+		return nil
+	}
+	on := true
+	return &on
+}
+
 // skipPassNames flattens the indexer's skip-set into the slice the
 // coordinator forwards to each subprocess via --skip-pass.
 func skipPassNames(set map[string]bool) []string {
@@ -120,6 +135,10 @@ func runExtractSubprocess(argv []string) error {
 		skipPasses = fs.String("skip-pass", "", "comma-separated pass names to skip")
 		drfNames   = fs.String("drf-names", "", "path to coordinator-written DRF register-name file (#1292)")
 		ormFields  = fs.String("orm-fields", "", "path to coordinator-written ORM field-name file (#2505)")
+		// #6997 — tri-state: unset means "the parent said nothing", and the
+		// child then resolves the custom-extractor gate from the inherited
+		// environment exactly as the in-process path does.
+		customExtr = fs.String("custom-extractors", "", "programmatic custom-extractor opt-in (true/false); empty means fall back to GRAFEL_INPROC_CUSTOM_EXTRACTORS")
 	)
 	if err := fs.Parse(argv); err != nil {
 		return err
@@ -140,6 +159,18 @@ func runExtractSubprocess(argv []string) error {
 		}
 	}
 
+	var customExtractors *bool
+	switch strings.ToLower(strings.TrimSpace(*customExtr)) {
+	case "":
+		// nil — env fallback.
+	case "1", "true", "yes":
+		on := true
+		customExtractors = &on
+	default:
+		off := false
+		customExtractors = &off
+	}
+
 	return extract.Run(context.Background(), extract.SubprocessOptions{
 		RepoRoot:      *repo,
 		Language:      *lang,
@@ -149,5 +180,7 @@ func runExtractSubprocess(argv []string) error {
 		SkipPasses:    skipSet,
 		DRFNamesPath:  *drfNames,
 		ORMFieldsPath: *ormFields,
+
+		CustomExtractors: customExtractors,
 	})
 }
--- a/cmd/grafel/index.go
+++ b/cmd/grafel/index.go
@@ -1806,6 +1806,13 @@ func (i *Indexer) Run(ctx context.Context, absRepo string) (*graph.Document, err
 			// #5135: foreground rebuilds run at the higher rebuild cap;
 			// background scheduler reindexes stay throttled.
 			Interactive: i.interactive,
+			// #6997 — the same programmatic half of the custom-extractor gate
+			// classifyAndReadWithProgress stamps onto ExtractorConfig, carried
+			// to the children so this path evaluates the identical gate. Only
+			// the opt-in is stamped, for the same reason as there: a false
+			// would OVERRIDE a set env var, which is not what an unset option
+			// means.
+			CustomExtractors: subprocCustomExtractorsOptIn(i.customExtractors),
 		})
 		if cerr != nil {
 			trk.Fail(cerr.Error())
--- a/internal/daemon/extract/coordinator.go
+++ b/internal/daemon/extract/coordinator.go
@@ -77,6 +77,14 @@ type CoordinatorConfig struct {
 	// foreground rebuild runs fast; only the throttled background path keeps
 	// the conservative GRAFEL_EXTRACT_GOMAXPROCS default (1 since #5960).
 	Interactive bool
+
+	// CustomExtractors forwards the PROGRAMMATIC half of the
+	// custom-extractor gate to every child via --custom-extractors
+	// (#6997). nil (the production case) forwards nothing and lets each
+	// child resolve the gate from the inherited environment, which is the
+	// same answer the in-process path computes. See
+	// SubprocessOptions.CustomExtractors.
+	CustomExtractors *bool
 }
 
 // runtimeCaps is the process-wide runtime-reloadable cap store (#5137). The
@@ -526,6 +534,17 @@ func Coordinate(ctx context.Context, repoRoot string, files []string, cfg Coordi
 			}
 			if skip != "" {
 				args = append(args, "--skip-pass", skip)
+			}
+			// #6997 — carry the programmatic custom-extractor opt-in across
+			// the process boundary. Only forwarded when the parent actually
+			// said something; otherwise the child falls back to the env half,
+			// which it inherits through childEnv.
+			if cfg.CustomExtractors != nil {
+				if *cfg.CustomExtractors {
+					args = append(args, "--custom-extractors", "true")
+				} else {
+					args = append(args, "--custom-extractors", "false")
+				}
 			}
 
 			cmd := exec.CommandContext(ctx, bin, args...)
--- a/internal/daemon/extract/custom_gate_6997_test.go
+++ b/internal/daemon/extract/custom_gate_6997_test.go
@@ -1,0 +1,287 @@
+package extract
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"os"
+	"path/filepath"
+	"sort"
+	"strings"
+	"testing"
+
+	"github.com/cajasmota/grafel/internal/extractors"
+)
+
+// #6997 — the subprocess extract path used to dispatch custom extractors under
+// `runExtract` alone: no custom-extractor gate, no `TSTree != nil`. The two
+// other dispatch sites (cmd/grafel/index.go, internal/extractors/incremental.go)
+// carry both. This pins the gate half BEHAVIOURALLY, at this path's own
+// entrypoint, so a source-level scan is not the only thing standing between the
+// paths and a fresh divergence.
+//
+// Measured consequence of the fix, macOS/APFS, GRAFEL_SUBPROC_EXTRACT=1 (the
+// only way to reach this file in production; default-OFF), against 81f99d691:
+// django-realworld 456 -> 371 entities with the gate OFF, and 456 -> 456 with
+// GRAFEL_INPROC_CUSTOM_EXTRACTORS=1 — i.e. byte-for-byte the old behaviour when
+// the gate is on, and the in-proc path's behaviour when it is not.
+
+// runCustomGateArm runs one extraction of a Django-shaped python file and
+// returns the emitted entity identities as "<kind>|<name>".
+func runCustomGateArm(t *testing.T, custom *bool) []string {
+	t.Helper()
+	repo := t.TempDir()
+	src := "from django.db import models\n" +
+		"\n" +
+		"\n" +
+		"class Widget(models.Model):\n" +
+		"    name = models.CharField(max_length=10)\n"
+	if err := os.WriteFile(filepath.Join(repo, "models.py"), []byte(src), 0o644); err != nil {
+		t.Fatal(err)
+	}
+	batch := filepath.Join(t.TempDir(), "batch.txt")
+	if err := os.WriteFile(batch, []byte("models.py\n"), 0o644); err != nil {
+		t.Fatal(err)
+	}
+
+	var buf bytes.Buffer
+	if err := Run(context.Background(), SubprocessOptions{
+		RepoRoot:         repo,
+		BatchPath:        batch,
+		BatchID:          "gate-6997",
+		Output:           &buf,
+		CustomExtractors: custom,
+	}); err != nil {
+		t.Fatalf("Run: %v", err)
+	}
+
+	var got []string
+	dec := json.NewDecoder(strings.NewReader(buf.String()))
+	for dec.More() {
+		var env Envelope
+		if err := dec.Decode(&env); err != nil {
+			t.Fatalf("decode: %v\n---\n%s", err, buf.String())
+		}
+		if env.Type == KindEntity && env.Entity != nil {
+			got = append(got, string(env.Entity.Kind)+"|"+env.Entity.Name)
+		}
+	}
+	sort.Strings(got)
+	return got
+}
+
+func TestSubprocCustomExtractorsHonourTheGate6997(t *testing.T) {
+	ptr := func(b bool) *bool { return &b }
+
+	cases := []struct {
+		name string
+		env  string // "" means unset
+		cfg  *bool
+		want bool // custom entities expected
+		why  string
+	}{
+		{"default: env unset, no programmatic opt-in", "", nil, false,
+			"this is what every production caller passes; #6966 keeps the gate default-OFF"},
+		{"env opt-in", "1", nil, true,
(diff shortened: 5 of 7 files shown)


cajasmota:
Added `ci:full` so the macOS leg runs. The diff has no `runtime.GOOS` branch, so it does not meet that rule literally — I added it for the shape rather than the letter.

`TestCoordinateForwardsTheProgrammaticCustomGate6997` **forks a built binary**, `index.go` forks `os.Executable()`, and the gate now crosses a process boundary as a `--custom-extractors` flag. Process-forking plus argv plumbing is precisely where this repo breaks on other platforms and nowhere else: #7001's Windows leg failed this week on a path that was invisible on macOS and Linux (a named-pipe socket path has no filesystem directory, so a file the code wrote everywhere else was never written at all).

It is also worth knowing while reviewing this that per **#7005** you cannot check any of that locally: `GOOS=windows go vet ./internal/...` does not compile in this tree, because cross-builds default to `CGO_ENABLED=0` and every tree-sitter grammar package drops out. So for a diff of this shape, CI is not a second opinion — it is the only one.

No re-trigger dance needed here: the label went on before any push, so `labeled` fires on its own.

cajasmota:

Measured in `/private/tmp/claude-501/-Users-jorgecajas-Projects-archigraph/c0fefed4-0195-4c81-9a94-07a7145710a4/scratchpad/review-7004-z3q/wt` (own worktree at `e37b08afd`), plus a detached `81f99d691` tree for the base binary. macOS/APFS, Apple Silicon, `GOMAXPROCS=3`, `GRAFEL_SUBPROC_CONCURRENCY=3`, isolated `HOME`/`GRAFEL_HOME`/`GRAFEL_DAEMON_ROOT` per run, corpus copied out of the read-only tree. **Every figure below is macOS/APFS only (#6966).**

Two findings, both non-blocking, both priced. Neither is a regression in this PR.

---

### (a) The reachability claim is TRUE. The "owner-level fork" framing I was briefed with was wrong.

I did not read the call sites — I broke them.

**Rename-to-break** (survives every indirection a grep misses — function values, interface methods, registry entries, converted constants):

| renamed symbol | non-test references that stopped compiling |
|---|---|
| `extract.Coordinate` | `cmd/grafel/index.go:1801` — **one**, inside `if subprocExtract()` |
| `extract.Run` (subproc.go) | `cmd/grafel/extract.go:174` — **one**, inside `runExtractSubprocess` |

`runExtractSubprocess` reaches production only as a function **value**: `cmd/grafel/main.go:142` `RunExtract: runExtractSubprocess` → `internal/cli/extract.go:28` — i.e. the hidden `grafel extract` subcommand, which only `Coordinate` fork-execs. That is exactly the indirection a scoped grep would have missed, and the rename caught it.

**Runtime proof, not inference.** I built a probe binary with one `fmt.Fprintln(os.Stderr, "ZZPROBE-subproc-Run-executed")` at the top of `extract.Run` and indexed django-realworld:

| env | probe hits in `extract.Run` |
|---|---|
| default (nothing set) | **0** |
| `GRAFEL_SUBPROC_EXTRACT=1` | **2** |

**And the decisive user-facing measurement — the default graph does not move.** `grafel index-internal` (the child the daemon forks on its default `GRAFEL_SUBPROCESS_INDEXER` path) on django-realworld:

| binary | env | entities / rels | entity identity set vs base |
|---|---|---|---|
| `81f99d691` | default | 378 / 1329 | — |
| this branch | default | 378 / 1329 | **0 differing lines** |

So no default user's graph loses a single entity. The fork I posited does not exist.

### (b) Claim 2 reproduced independently, on identity rather than counts

All arms `GRAFEL_SUBPROC_EXTRACT=1`:

| arm | entities / rels | vs `81f99d691` |
|---|---|---|
| `81f99d691` | 456 / 1463 | — |
| branch, gate OFF | **371 / 1298** | 88 removed, 3 added |
| branch, `GRAFEL_INPROC_CUSTOM_EXTRACTORS=1` | 456 / 1463 | **0 differing lines — identical entity SET, not just count** |
| `81f99d691`, env gate ON | 456 / 1463 | 0 — confirms the old binary ignores the gate in both directions |

By-kind removals: `SCOPE.Operation` −51, `Datastore` −19, `Schema` −13, `Component` +3/−2, `Pattern`/`External`/`Config` −1 each. **Net −85, matching the PR's table member for member**, including the `SCOPE.Component +1` merge artefact. A substitution could not have hidden here: the gate-ON arm is set-identical, so nothing was swapped underneath the total.

The PR's framing of its own zero (the `TSTree` guard removing zero entities on these three repos is *corpus-relative*, with nim named as the shape where it is not) is honest and correctly hedged.

### (c) Claim 5 confirmed — the named pin did not pin

`TestInProcCustomExtractorsNilTreeGuardIsLoadBearing` observes only `RunCustomExtractors`' own output with a nil tree; it names no call site. Run with `-v`: it **PASSes and logs** (`still emits 1 entity(ies)`) — the `t.Skip` does **not** fire, so the test exists but grades nothing about any dispatch site. All three sites could have dropped both guards with it green.

---

## My own mutants — 6 scored, `-count=1`, `go vet` clean before each, each edit proved by an exact occurrence count and the line it landed on, restored by `cp` from a shasum-verified backup with `git diff HEAD` empty afterwards

| # | mutant (permissive direction preferred) | seam | result |
|---|---|---|---|
| **R1** | fourth production dispatch site, unguarded, as a **package-level `var x = func(...)`** | scanner's node selection | **ALIVE — distinguishable and ungraded** |
| **R1b** | fourth site inside a real `func`, called through a **function value** (`d := extractors.RunCustomExtractors; d(ctx, file)`) — the `Hooks{RunExtract: …}` shape this repo already uses in `main.go` | `isDispatcherCall6997` | **ALIVE — distinguishable and ungraded** |
| **R5** | **two sites in one function**: a second, unguarded `RunCustomExtractors` inside `subproc.go`'s `Run` | enclosing-function-name resolution | **DEAD** — duplicate key breaks set equality *and* the per-site guard check fires |
| **R2** | **negate** the gate: `runCustom := runExtract && !CustomExtractorsEnabled(customCfg)` (permissive: custom extractors then run on the default) | gate semantics vs the scan's substring test | **DEAD** behaviourally (`HonourTheGate6997` + `CoordinateForwards…`). Scanner passes it — the substring check is a *shape* check, not a semantics check |
| **R3** | child-side `--custom-extractors` parsing: `default:` no longer sets `&false` (permissive: an explicit false stops suppressing a set env var in the child) | flag parsing on the child side | **DEAD** — `CoordinateForwardsTheProgrammaticCustomGate6997`, cfg-false arm |
| **R4** | coordinator strips `GRAFEL_INPROC_CUSTOM_EXTRACTORS` from `childEnv` | the `os.Environ()` inheritance assumption | **DEAD** — same test, env-on arm |
| **R6** | negate the gate at `incremental.go` (a site this PR grades only by scan) | third site | **DEAD** — four pre-existing `…6960` tests |

**R1/R1b is the answer to "can the scanner be defeated by a differently-shaped fourth call site": yes.** `scanDispatchSites6997` only descends into `f.Decls` `*ast.FuncDecl`s, so a dispatch inside a package-level function literal is invisible; and `isDispatcherCall6997` matches only a direct `Ident`/`SelectorExpr` callee, so a call through a function value is invisible. Both mutants left the whole suite green.

I want to be precise about severity: this is a hole in a **new safety net**, not a regression, and the three real sites are each *also* graded behaviourally (R2/R3/R4/R6 all died on behaviour, not on the scan). But R1b's shape is not hypothetical here — this codebase already routes `runExtractSubprocess` through exactly that indirection.

**Priced fix, ~10 lines, no restructuring:** after collecting sites, count every `*ast.Ident` named `RunCustomExtractors` across the same non-test file set and assert it equals the number of collected call sites. Both R1 and R1b break that equality (the ident exists; the site was never collected). That converts "the scanner did not see it" from a silent pass into a failure. Worth a follow-up issue rather than a block.

### (d) No baseline moved — verified, not trusted

`scripts/quality/run.sh --ratchet` on the branch, fresh isolated home, exit **0**, `38 gated fixtures held their baseline (29 at 100% must-have recall)`. Because that line prints even when a ceiling falls, I compared **all 38 fixtures' reports field-by-field against the committed `internal/quality/golden/baseline.json`**, including the `entity_extracted_total` / `relationship_extracted_total` **ceilings**:

```
baseline fixtures: 38   reports: 38
MISSING reports: []   DROPS: []   RISES: []
```

Zero drops, zero rises, zero unmeasured fixtures. The diff also touches no baseline file, so nothing was re-recorded.

### M9's mitigation — it works, but a behaviour pin was affordable

I applied M9 (`subprocCustomExtractorsOptIn(true)` in `index.go`) and scored it two ways:

- The author's AST pin **kills it**: `--- FAIL: TestSubprocCoordinatorGetsTheIndexersOwnOptIn6997`.
- The **artefact** is also distinguishing and I measured it: the M9 binary indexes django-realworld to **456 / 1463** with the env gate unset and `GRAFEL_SUBPROC_EXTRACT=1`, where the honest branch gives **371 / 1298**.

So the mutant is dead either way, and the PR's reasoning about `os.Executable()` is correct — but the conclusion "a behaviour pin was not affordable" does not hold. `internal/daemon/extract/coordinator_test.go:295` already has `buildGrafel(t)`, and `cmd/grafel` already has four tests that build and fork the binary. A test that builds `grafel`, runs `index-internal` twice under `GRAFEL_SUBPROC_EXTRACT=1` (env gate unset vs set) and asserts the gate-off arm emits no `SCOPE.*` custom entities is roughly the same ~40 lines as the AST assertion, needs **no** `BinaryPath` override in production, and pins the emitted graph instead of the code's description of itself — so it survives a rename or an inline of `subprocCustomExtractorsOptIn`, which the current pin does not.

Recommendation, not a blocker: keep the AST test (it also guards the second-`CoordinatorConfig`-literal case, which a behavioural test would not see) and add the end-to-end arm in follow-up.

### Verification on the restored tree

`git status --porcelain` empty, `git diff HEAD` empty, `gofmt -l` clean, `go vet` clean.

- `go test -count=1 ./internal/daemon/extract/ ./internal/extractors/` — **ok** (51.3s / 52.9s)
- `go test -count=1 ./cmd/grafel/` — **ok** (155.5s), fresh `HOME` + `GRAFEL_HOME` + `GRAFEL_DAEMON_ROOT`
- `scripts/quality/run.sh --ratchet` — exit 0, 38/38 figures identical to baseline

**APPROVE.** The reachability claim holds under a method that survives indirection *and* at runtime; the blast-radius table reproduces on identity; the default graph is unchanged entity-for-entity. Two follow-ups worth filing: the ident-count hardening of the scanner (R1/R1b), and the end-to-end arm for M9.

cajasmota:
## Coordinator: taking finding (b) here rather than deferring it

The review is APPROVE with two non-blocking recommendations. I am taking **(b)** before merge and leaving **(c)** as a follow-up. Reasoning, so the split is on the record:

**(b) — the AST scanner is defeatable, and by a shape this repo already uses.** `scanDispatchSites6997` only descends into `*ast.FuncDecl` and `isDispatcherCall6997` only matches a direct callee, so a fourth dispatch site written as a package-level `var x = func(...)`, or reached through a function value, is invisible — the reviewer confirmed both with the whole suite green.

That is not a hypothetical shape here. **`extract.Run` is itself reached only as a function value** (`Hooks{RunExtract: …}`), which the reviewer established while proving the reachability claim. So the one mechanism this PR adds to stop the three sites diverging is blind to the calling convention the surrounding code already uses.

The reviewer priced the fix at **~10 lines** — assert that the count of `RunCustomExtractors` identifiers across the scanned file set equals the collected site count. "Too expensive to grade" is a claim rather than evidence, and this one has been priced. A safety net is cheapest to close while it is new and its author still has the context; deferred, it becomes an issue nobody sizes. Taking it.

**(c) — the M9 behaviour pin — I am deferring, and I want to be honest that this is the weaker half of the decision.** The reviewer is right that `buildGrafel(t)` already exists, that four `cmd/grafel` tests already build and fork the binary, and that an end-to-end arm is roughly the same ~40 lines while surviving a rename or inline that the current expression-pin would not. That is a better test than what is here.

I am deferring it because it replaces a working pin with a better one rather than closing a hole, and because the current pin does kill M9 — verified twice, by the author and by the reviewer's own artefact measurement (456/1463 vs the honest 371/1298). **Deferring is not a judgement that it is unnecessary**, and per the standing rule I am recording it as a follow-up rather than letting an approval quietly absorb it.

### On the reachability claim

The review settled it by **rename-to-break**, which survives indirection where a grep does not — the method matters, because "single caller" is exactly the class of claim a scoped grep gets wrong. `extract.Coordinate` has one non-test reference, `extract.Run` one, reachable only as a function value; the runtime probe showed `extract.Run` executing **0** times under default env and 2 under `GRAFEL_SUBPROC_EXTRACT=1`; and base vs branch on the default path produced **378/1329 with a 0-line entity-identity diff**.

**My fork framing was wrong.** I briefed this expecting an owner-level decision that would strip entities from every user's daemon graph, and the measurement says no default user loses an entity. Recording that plainly: the PR's reversal is correct and it was established by measurement, not by argument.

Gate-ON being **set-identical** (0 differing lines, not merely an equal count) is the assertion that matters on claim 2 — an equal count cannot see a substitution, which is how a previous defect on this board hid while a total moved by +8 and 7,221 edges were swapped underneath it.

cajasmota:
Finding (b) is taken in `1e145e60b`. R1 and R1b were real: I reproduced both, and both left the whole suite green before the change.

The scan now counts every mention of the dispatcher across the same non-test file set and requires each one to be accounted for by a collected site. The count comes from an **independent** traversal — its own walk, its own parse, a flat ident visitor instead of the guard-aware statement walk — because a count derived from the collection it grades cannot detect an omission in that collection. The declaration's own name is excluded: `func RunCustomExtractors(...)` is the thing being called, not a call of it. The failure prints every mention as `file:line:col`, so an unattributable one says *where* to teach the site walk rather than only that something is wrong.

Scored `-count=1`, green baseline immediately before each, every edit proved by an exact occurrence count and its asserted hit set, restored by `cp` from a shasum-verified backup with `git diff HEAD` clean afterwards:

| # | mutant | result |
|---|---|---|
| R1 | fourth site as a package-level `var x = func(...)` | **DEAD** — 4 mentions, 3 sites, names `zz_r1_mutant.go:10:24` |
| R1b | fourth site called through a function value | **DEAD** — 4 mentions, 3 sites, names `zz_r1b_mutant.go:10:18` |
| V1 | the mention scan matches no files at all | **DEAD** — the vacuity `Fatal` fires ("either would make the coverage check below pass trivially") rather than the equality passing 0 == 0 |
| V2 | stop excluding the declaration's own name | **DEAD** — 4 vs 3, i.e. the count is a real measurement and not tuned to agree |
| M4 | frozen-list entry deleted | **DEAD** (re-scored) |
| M5 | direct fourth site | **DEAD** (re-scored) |

V1 and V2 are there because the two ways this check could have been worthless are an empty file set and a count that agrees by construction; each now has its own dead mutant.

Verification: `go test -count=1 ./internal/extractors/ ./internal/daemon/extract/` green (44.9s / 50.9s), `go vet` and `gofmt -l` clean. `./cmd/grafel/` and the ratchet were not re-run — this commit touches one test file in `internal/extractors` and nothing else (`git diff HEAD --stat`: 1 file, +100).

On **(c)**: agreed the end-to-end arm is the better pin, and the reasoning about `os.Executable()` does not extend to "unaffordable" now that `buildGrafel(t)` is on the table. Deferring per the coordinator, as a preference rather than a judgement that it is unnecessary — the current pin does kill M9, twice measured.

Also recorded in the body, from #7005: this tree cannot be cross-compiled for Windows at all (`CGO_ENABLED=0` drops every tree-sitter grammar), so for a diff that forks a binary and plumbs a flag across a process boundary, CI is the only opinion available — which is why `ci:full` is on this PR.

cajasmota:
## Coordinator mutant on `1e145e60b` — **CR-1 DEAD**

Own detached worktree at `/private/tmp/.../coord-mut-7004-r7t/wt` (absolute). Green baseline immediately before (`./internal/daemon/extract/` 53.5s, `./internal/extractors/` 58.6s, both exit 0). Restored by `cp` from a SHA-256-verified backup (`70ee4b12…` both ways); `git diff HEAD` and `git status --porcelain` empty afterwards.

### CR-1 — flip the compound guard from `&&` to `||`

`subproc.go:399`: `if runCustom && file.TSTree != nil {` → `if runCustom || file.TSTree != nil {`

Edit proven landed: exactly **1** occurrence, hit set **`[line 399]`**, `go vet` clean.

I picked this because it is the one rewrite that **keeps both guard identifiers textually present** while destroying the conjunction. A scan that checks "both guards appear at this site" is satisfied by it, so it separates a real behavioural grading from a presence check — and a compound guard scored only as a unit grades neither half. M2 and M3 already showed each half fails a disjoint assertion set; this asks the complementary question: is the *combination* itself pinned, or only the two members?

**Result: DEAD**, killed on behaviour by three tests, not by the scan:

- `TestSubprocCustomExtractorsHonourTheGate6997/env_opt-in`
- `TestSubprocCustomExtractorsHonourTheGate6997/programmatic_opt-in,_env_unset`
- `TestSubprocCustomExtractorsRequireAParseTree6997`
- `TestCoordinateForwardsTheProgrammaticCustomGate6997`

`./internal/extractors/` stayed green, which is the right locality — the guard lives in `daemon/extract` and only that package's behavioural arms should notice.

So the conjunction is pinned as a conjunction, and the AST scan is not carrying weight it cannot bear.

### On finding (b)

The closure is the right shape. Deriving the mention count from an **independent** traversal — its own walk, its own parse, a flat ident visitor rather than the guard-aware statement walk — is what makes it a real measurement rather than a tautology, and `V2` (stop excluding the declaration's own name) is the mutant that proves it. `V1` covers the empty-file-set path. Those are the two ways this class of check goes vacuous, and each now has its own dead mutant.

Reproducing `R1` and `R1b` before changing anything, rather than fixing from the description, is also the right order — it means the fix is known to close the reported hole and not an adjacent one.

I am satisfied on my own account. **Merging on green.**

Issue #6916, opened 2026-09-06
Rule patterns using `name_group: 0` mint entities named after the whole regex match — Tier A + Tier C COMPLETE (20 of 39 sites); Tier B x4 and Tier D remain


cajasmota:
Rule patterns using `name_group: 0` mint entities named after the whole regex match — `"gin.Default()"` and `"WebApplication.CreateBuilder("` are entity Names in the graph today

Found while grounding contributor report #6726's four unchecked 100%-terminal kinds. This is the **`Config`** half of that grounding, and it is the reason "wire up the missing edges" is the wrong first move for that kind.

## The defect

Both the csharp and go `Config` patterns use **`name_group: 0`** — capture group 0 is the entire regex match, so the whole matched text becomes the entity's `Name`.

Measured entity names from a purpose-built fixture, indexed in-process:

```
Config  "WebApplication.CreateBuilder("
Config  "gin.Default()"
```

These are **marker nodes naming a call site, not a thing.** There is no referent for an edge to point at, and no user-meaningful identity: `"gin.Default()"` is not a config object anyone can navigate to, and the trailing `(` in the csharp one is a bare artefact of where the regex happened to stop.

The csharp `Middleware` names have the same problem from the other direction — `Authentication`, `Authorization`, `Routing` are **framework pipeline stages**, not entities in the repo being indexed.

## Why this blocks the obvious fix

#6726 reports `Config` at 16/16 terminal orphan and `Middleware` at 49/49. The obvious remedy is "add the missing relationship rules". **That is premature here**, because an edge into a node named `"gin.Default()"` is not more useful than no edge — it is a dangling-by-construction endpoint dressed up as a resolved one, and per the resolver grounding on #6906 a non-binding endpoint is *worse* than no edge, since it lands as `bug-extractor`.

**Decide whether these should be entities at all before wiring anything into them.**

There is a related structural problem recorded on #6914: `*/Middleware/*.cs` with `name_from: class_name` emits **no** entity (`detector.go:543` default branch), so the real `LoggingMiddleware` class is a `SCOPE.Component`. The kind-specific nodes are a **parallel shadow population** beside the real class entities — the same shape #6912 found for Hibernate, where a wired parallel population stands beside an unwired main one.

## What a fix has to establish

[...]

Issue #6914, opened 2026-09-06
P2: Go `.Use` middleware rules — STEP 1 SHIPPED (#7006, the four fabricating relationship rules deleted); step 2 (REGISTERED_ON) blocked on #6916, step 4 (stop minting csharp/sinatra Middleware entities) is a separate decision


cajasmota:
The Go `.Use(...)` middleware rule never fires on the normal single-argument form, and fabricates a wrong edge on the two-argument form — replicated identically in gin, chi, echo and fiber

Found while grounding contributor report #6726's four unchecked 100%-terminal kinds. **`Middleware` is the one of the four that is a genuine defect**, and it has two independent halves.

## Half 1 — the Go rule is broken in both directions

gin, chi, echo and fiber all carry the **identical** rule:

```
\.Use\s*\(\s*(\w[\w.]*)\s*(?:,\s*(\w[\w.]*))?\s*\)
source_group: 1, target_group: 2
```

Group 2 is **optional**. So on the normal form:

```go
r.Use(middleware.Logger)     // targetName == "" -> detector.go:556 `continue` -> NO EDGE, EVER
```

and on the two-argument form:

```go
r.Use(AuthMiddleware, listOrders)
  -> Middleware/AuthMiddleware --ROUTES_TO--> Controller/listOrders
```

which is **semantically fabricated**: `Use(A, B)` registers two middlewares. It does not make A route to B.

Both halves measured on a purpose-built go/gin fixture indexed in-process: single-arg → 0 edges; two-arg → the wrong edge, with both endpoints binding to real entity IDs through the `byKind` tier.

**So the rule is dead on the common form and wrong on the rare one.** Neither direction produces a correct edge.

**The golden corpus already contains the proof.** `internal/quality/golden/go-chi-mini/expected.json:28` asserts the `middleware.Logger` **entity** exists, with the note *"r.Use(...) middleware binding"* — and asserts **no edge on it**, because the rule cannot produce one for that line. The fixture documents the symptom without naming it.

This is close kin to the **#6809** self-loop family: a rule whose capture-group wiring produces something other than what its name claims.

## Half 2 — no csharp producer at all

[...]

PR #7006, opened 2026-09-08
fix(#6914): delete the four Go `.Use` middleware relationship rules — dead on 1 arg, fabricated on 2, silent on 3+

cajasmota:
## What

Deletes the four Go `.Use` middleware **relationship** rules (gin, chi, echo, fiber) and pins their
absence. This is **step 1 only** of #6914 — steps 2–4 stay open, hence `Refs`, not `Closes`.

## Why

All four files carried the identical rule, whose second capture group is optional:

```
pattern:      '\.Use\s*\(\s*(\w[\w.]*)\s*(?:,\s*(\w[\w.]*))?\s*\)'
source_type:  Middleware
target_type:  Controller
relationship: ROUTES_TO
source_group: 1
target_group: 2
```

Measured through the real `LoadAllRules()` + `Detector.Detect()` path on `src/main.go`, language `go`
(worktree at `.../scratchpad/impl-6914-k7m2/wt`, macOS/APFS):

| form | before | after |
|---|---|---|
| `r.Use(middleware.Logger)` | ents `[Middleware:middleware.Logger]`, rels `[]` | **identical** |
| `r.Use(middleware.Logger, middleware.Recoverer)` | rels = **4 ×** `Middleware:middleware.Logger --ROUTES_TO--> Controller:middleware.Recoverer` (one per rule file) | rels `[]` |
| `r.Use(middleware.RequestID, middleware.RealIP, middleware.Logger)` | ents `[]`, rels `[]` | **identical** |
| `r.Use(gin.Recovery())` | ents `[]`, rels `[]` | **identical** |

The only difference the deletion makes is the four fabricated edges disappearing. `Use(A, B)`
registers two middlewares on the router; it does not make A route to B. And it fires on
`r.Use(middleware.Logger, middleware.Recoverer)` — a line in essentially every chi service — so
reachability is settled without a corpus survey.

**No correct edge is lost**, demonstrated rather than argued: the one-arg, three-arg and
call-expression forms emitted **zero** relationships *while the rules were still present*, so the
deletion cannot have taken anything from them. `TestIssue6914_NoCorrectEdgeWasLost` asserts exactly
that, and it is unchanged by restoring the rules — which is the point.

**Entity rules untouched.** `Middleware:middleware.Logger` is still minted, held in place by
`TestIssue6914_UseEntityRuleStillMintsMiddleware` so an over-broad "fix" that rips out the whole
`.Use` block cannot pass the main pin for the wrong reason. Whether that shadow population should
exist at all is step 4, and `go-chi-mini/expected.json:28` asserts it today.

## Pinning the absence

The always-on gate is blind here. Verified by parsing every `expected.json` (not by grep): **zero
`expected_relationships` rows in any of the 38 golden fixtures name `Middleware` at either end** —
all `Middleware` rows are `expected_entities`/`forbidden_entities`. Nothing would catch this rule
returning.

The new test asserts the **emitted artefact** — no relationship with a `Middleware` endpoint from a
two-argument `.Use` line — not a count of rules and not a property of the YAML text. A
count-of-rules assertion passes the moment someone re-adds the edge in a different spelling. The
assertion reads the endpoint **kind** off `FromID`/`ToID` (the detector builds both as
`<type>:<name>`), so it cannot be fooled by an entity merely *named* something middleware-ish — the
name/kind confusion `nim-objects-mini`'s `from_name: "Config"` rows set up.

## Mutants

Each edit's landing was proved with an exact occurrence count **and** the hit set across all four
files, since the same rule text appears in four places. Green baseline immediately before each
score; `-count=1` throughout.

| # | mutant | verdict |
|---|---|---|
| M1 | restore the deleted rule verbatim in **chi.yaml only** (`target_type: Middleware` count: chi 1, gin/echo/fiber 0) | **DEAD** — 1 offending edge, both assertions fired |
| M2 | re-add to **gin.yaml only**, *reversed and renamed*: `source_type: Controller`, `target_type: Middleware`, `relationship: USES`, groups swapped | **DEAD** — `Controller:middleware.Recoverer --USES--> Middleware:middleware.Logger`. The kind-based assertion survives a direction flip, a group swap and a different relationship kind |
| M3a | delete the `Middleware` **entity** rule from **all four** files | **DEAD** — the scope guard fires |
| M3b | delete the `Middleware` entity rule from **chi.yaml only** (entity-rule count: chi 0, gin/echo/fiber 1) | **ALIVE — equivalent mutant by construction.** The four `Middleware` entity rules are byte-identical, none carries `requires_framework`, and `detector.go` sets the entity's `framework` property from **`lang`**, not from the producing file — so the emitted record is identical in Kind, Name, SourceFile, StartLine and every property whichever rule minted it, and the dedup key `entityType:name:file` collapses them. **No `.go` input can distinguish it.** Not "expensive to grade": ungradeable. Underlying defect filed as #7007 |

## Review round 2 — M-R5 and the missing positive control

Both blockers from the review are fixed in `08e8db3a7` (test-only, no rule YAML touched).

**M-R5** — re-add to `gin.yaml` alone with identical `source_type: Middleware` / `target_type: Controller` /
`ROUTES_TO` / groups 1,2, pattern narrowed to `'\.Use\s*\(\s*(\w+)\s*,\s*(\w+)\s*\)'` — was **ALIVE**
because the only two-argument fixture used **dotted** names, pinning arity and leaving identifier shape open. The
two-argument pin is now table-driven over both shapes (dotted chi stack + `v1.Use(AuthRequired, Logger)`).
Re-scored: **DEAD** — `Middleware:AuthRequired --ROUTES_TO--> Controller:Logger`, failing on the **new
bare-identifier leg** while the dotted leg passes, so the widening is what does the killing. Landing proved with
an asserted hit set: the mutant's fixed pattern string and `source_type: Middleware` each count gin 1,
chi/echo/fiber 0.

**Positive control** — each two-argument fixture emits **zero** relationships in total at HEAD, so the absence
assertion could not fail. Both now assert a control edge (`Route:/health --ROUTES_TO--> Controller:healthHandler`,
`Route:/v1/items --ROUTES_TO--> Controller:listItems`). **Verified able to fail:** deleting the route line from
either constant (1 occurrence each, asserted) drops that fixture to zero relationships and the control fires on
**both** legs. Recorded caveat, in the test comment: the control is *liveness only*, not a pin on a specific
route rule — several Go rule files match the same route line, so flipping `gin.yaml`'s `ROUTES_TO` to `LINKS_TO`
leaves the control edge standing and the suite green.

**M-R2** (re-add with renamed kinds, `Handler`/`CALLS`) is left **uncovered on purpose and named in the test
comment** with the reason: killing it means asserting a `.Use` line emits no relationship *at all*, true today but
in direct conflict with step 2's intended `Middleware REGISTERED_ON <app>` edge once #6916 unblocks a
per-variable router anchor.

Re-verified after the change (worktree `.../scratchpad/impl-6914-k7m2/wt`, macOS/APFS, `-count=1`): `gofmt -l`
clean, `go vet` clean, `go test ./internal/engine/` **ok** 49.2s, `scripts/quality/run.sh --ratchet` exit 0, 38
fixtures held, no baseline moved. `./cmd/grafel/` not re-run — nothing outside the test file changed. YAMLs
restored by `cp` from a shasum-verified backup after every mutant, and the test file likewise (shasum
`7e77cdae4`); `git status` confirmed test-file-only before committing.

## Review round 3 — CQ-1 (multi-line layout)

Fixed in `851b40b29` (test-only, no rule YAML touched).

**CQ-1** — re-add to `chi.yaml` alone with the same `source_type: Middleware` / `target_type: Controller` /
`ROUTES_TO` / groups 1,2, pattern requiring a newline between the arguments
(`'\.Use\s*\(\s*(\w[\w.]*)\s*,\s*\n\s*(\w[\w.]*)'`) — was **ALIVE**. Both existing legs put the
arguments on one line, so the table varied identifier shape while holding **line layout** constant: M-R5's class,
one axis over. The original rules *did* match this layout (they used `\s*`, and RE2's `\s` matches `\n`), so the
pin's multi-line coverage was incidental and never asserted.

Added a third table leg — the same two-argument content in gofmt's layout, with its own control
(`Route:/orders --ROUTES_TO--> Controller:listOrders`). **CQ-1 re-scored: DEAD** —
`Middleware:middleware.RequestID --ROUTES_TO--> Controller:middleware.Logger`, failing **only** on the new
multi-line leg while the dotted and bare legs pass. Landing proved with an asserted hit set: `MUTANT CQ-1` and
`source_type: Middleware` each count chi 1, gin/echo/fiber 0; chi restored by `cp` from a shasum-verified backup
(`cf1c1c6c2` both ways), `git status` test-file-only afterwards.

**Three axes, and it stops there**: arity (`TestIssue6914_NoCorrectEdgeWasLost`), identifier shape, line layout.
Stated in the test comment — the goal is a pin that survives the obvious rewrites, not one that anticipates every
possible regex.

Re-verified: `gofmt -l` clean, `go vet` clean, `go test -count=1 ./internal/engine/` **ok** 43.6s,
`scripts/quality/run.sh --ratchet` exit 0, 38 fixtures held, no baseline moved. `./cmd/grafel/` not re-run —
nothing outside the test file changed.

## Ratchet — the three claims, verified here rather than taken on trust

1. **One fixture source line** touches `.Use`: `go-chi-mini/src/main.go:19`, `r.Use(middleware.Logger)` — single-arg, which already produced no edge.
2. **One expected row** names `Middleware`: `go-chi-mini/expected.json:28`, an *entity* row.
3. **No `expected_relationships` row anywhere** names `Middleware` (parsed, per above).

`scripts/quality/run.sh --ratchet` → `quality ratchet OK — 38 gated fixtures held their baseline (29
at 100% must-have recall)`, exit 0. **No baseline file moved** (`git status` clean apart from the
four YAMLs and the new test).

## Verification

All in `.../scratchpad/impl-6914-k7m2/wt`, macOS/APFS, `-count=1`, exit codes read from redirected
files (never a pipe):

- `gofmt -l internal/engine/` — clean; `go vet ./internal/engine/` — clean
- `go test -count=1 ./internal/engine/` — **ok**, 44.4s
- `go test -count=1 ./cmd/grafel/` — **ok**, 149.2s, with a **fresh** `GRAFEL_HOME` + `GRAFEL_DAEMON_ROOT`. This package digest-pins the whole graph, so a rule change must run it; it passed, i.e. the deletion moves no digest.
- `scripts/quality/run.sh --ratchet` — OK, no baseline movement

The four YAMLs were restored by `cp` from a shasum-verified backup after each mutant, never by `git
checkout`, and `git diff HEAD` was confirmed to hold only the intended deletion before committing.

Refs #6914

### DIFF
--- a/internal/engine/go_use_middleware_rule_6914_test.go
+++ b/internal/engine/go_use_middleware_rule_6914_test.go
@@ -87,6 +87,40 @@ func main() {
 }
 `
 
+// use6914TwoArgMultiline is the SAME two-argument content in gofmt's own
+// layout. Coordinator mutant CQ-1 re-added the defect with the arguments
+// separated by a newline —
+// `'\.Use\s*\(\s*(\w[\w.]*)\s*,\s*\n\s*(\w[\w.]*)'`, same
+// source_type/target_type/kind/groups — and a table whose every leg put both
+// arguments on ONE line slept through it. The deleted rules used `\s*`, and
+// RE2's `\s` matches `\n`, so they DID fire on this layout; the pin covered it
+// only incidentally, never by assertion, which is exactly what a narrower
+// re-add walks through.
+//
+// Three axes are enumerated here and that is where this stops: arity (the
+// NoCorrectEdgeWasLost test), identifier shape, and line layout. The goal is a
+// pin that survives the obvious rewrites, not one that anticipates every
+// possible regex.
+const use6914TwoArgMultiline = `package main
+
+import (
+	"net/http"
+
+	"github.com/go-chi/chi/v5"
+	"github.com/go-chi/chi/v5/middleware"
+)
+
+func main() {
+	r := chi.NewRouter()
+	r.Use(
+		middleware.RequestID,
+		middleware.Logger,
+	)
+	r.Get("/orders", listOrders)
+	http.ListenAndServe(":8080", r)
+}
+`
+
 const use6914TwoArgBare = `package main
 
 import "github.com/gin-gonic/gin"
@@ -201,6 +235,11 @@ func TestIssue6914_TwoArgUseEmitsNoMiddlewareEdge(t *testing.T) {
 			name:    "bare identifiers (gin group middleware)",
 			src:     use6914TwoArgBare,
 			control: "Route:/v1/items --ROUTES_TO--> Controller:listItems",
+		},
+		{
+			name:    "gofmt multi-line layout (chi stack of three)",
+			src:     use6914TwoArgMultiline,
+			control: "Route:/orders --ROUTES_TO--> Controller:listOrders",
 		},
 	} {
 		t.Run(tc.name, func(t *testing.T) {
--- a/internal/engine/go_use_middleware_rule_6914_test.go
+++ b/internal/engine/go_use_middleware_rule_6914_test.go
@@ -3,6 +3,7 @@ package engine
 import (
 	"context"
 	"fmt"
+	"slices"
 	"sort"
 	"strings"
 	"testing"
@@ -53,9 +54,23 @@ import (
 // (the `name_group: 0` router anchor); if it ever lands it will legitimately
 // need this pin updated — deliberately, with the new edge named here.
 
-// use6914TwoArg is the form that produced the fabricated edge. It is the
-// idiomatic chi middleware stack, not a contrivance.
-const use6914TwoArg = `package main
+// use6914TwoArgDotted and use6914TwoArgBare are the two-argument form in its
+// TWO identifier shapes. Both are needed, and the first alone is not enough.
+//
+// The deleted pattern captured `(\w[\w.]*)` — dotted OR bare — so a fixture
+// using only `middleware.Logger` pins the ARITY axis and leaves its neighbour,
+// the IDENTIFIER SHAPE, wide open. Review mutant M-R5 exploited exactly that:
+// re-adding the rule with the pattern narrowed to the house style already used
+// by chi's own route rule, `'\.Use\s*\(\s*(\w+)\s*,\s*(\w+)\s*\)'`, is the
+// same defect with the same source_type/target_type/kind/groups, fires on
+// textbook gin middleware — and a dotted-only fixture sleeps through it.
+//
+// Each carries a route line as its POSITIVE CONTROL. Both files emit ZERO
+// relationships in total once the rules are gone, so without a control edge an
+// assertion of the form "no relationship touches a Middleware endpoint" would
+// keep passing even if the constant stopped producing anything at all — a
+// vacuous absence. The control edge is named per-fixture in the table below.
+const use6914TwoArgDotted = `package main
 
 import (
 	"net/http"
@@ -67,7 +82,21 @@ import (
 func main() {
 	r := chi.NewRouter()
 	r.Use(middleware.Logger, middleware.Recoverer)
+	r.Get("/health", healthHandler)
 	http.ListenAndServe(":8080", r)
+}
+`
+
+const use6914TwoArgBare = `package main
+
+import "github.com/gin-gonic/gin"
+
+func main() {
+	r := gin.Default()
+	v1 := r.Group("/v1")
+	v1.Use(AuthRequired, Logger)
+	r.GET("/v1/items", listItems)
+	r.Run(":8080")
 }
 `
 
@@ -137,27 +166,71 @@ func rels6914(res *DetectResult) []string {
 //
 // The assertion is on the endpoint KIND, not on the exact fabricated edge
 // string, so it also kills a re-add that swaps the capture groups, reverses the
-// direction, or renames the captured symbols.
+// direction, renames the captured symbols, or narrows the identifier pattern.
+//
+// KNOWN UNCOVERED, deliberately: a re-add that also RENAMES THE KINDS — say
+// `source_type: Handler`, `target_type: Controller`, `relationship: CALLS` —
+// slips past this, because nothing in the emitted edge still says "Middleware".
+// Killing it would mean asserting that a `.Use` line emits NO relationship at
+// all. That is true today, but it would fight #6914 step 2, whose intended
+// `Middleware REGISTERED_ON <app>` edge comes off exactly these lines once
+// #6916 unblocks a per-variable router anchor. Naming the gap here beats
+// closing it in a way that a later, correct edge has to fight.
 func TestIssue6914_TwoArgUseEmitsNoMiddlewareEdge(t *testing.T) {
-	got := rels6914(detect6914Go(t, use6914TwoArg))
+	for _, tc := range []struct {
+		name string
+		src  string
+		// control is an edge the fixture must STILL emit. Without it an
+		// absence assertion cannot fail, and a fixture that quietly stopped
+		// producing anything would read as a pass. Verified able to fail:
+		// deleting the route line from either constant drops that fixture to
+		// zero relationships and this assertion fires.
+		//
+		// It is a liveness control, NOT a pin on any one framework's route
+		// rule — several Go rule files match the same route line, so flipping
+		// gin.yaml's ROUTES_TO to LINKS_TO leaves the control edge standing.
+		// Do not read a passing control as evidence about a specific rule.
+		control string
+	}{
+		{
+			name:    "dotted identifiers (chi middleware stack)",
+			src:     use6914TwoArgDotted,
+			control: "Route:/health --ROUTES_TO--> Controller:healthHandler",
+		},
+		{
+			name:    "bare identifiers (gin group middleware)",
+			src:     use6914TwoArgBare,
+			control: "Route:/v1/items --ROUTES_TO--> Controller:listItems",
+		},
+	} {
+		t.Run(tc.name, func(t *testing.T) {
+			got := rels6914(detect6914Go(t, tc.src))
 
-	var offenders []string
-	for _, r := range got {
-		if strings.HasPrefix(r, "Middleware:") || strings.Contains(r, "--> Middleware:") {
-			offenders = append(offenders, r)
-		}
-	}
-	if len(offenders) > 0 {
-		t.Errorf("a two-argument `.Use(A, B)` line emitted %d edge(s) touching a Middleware endpoint; "+
-			"`Use(A, B)` registers two middlewares, it does not relate A to B:\n  %s",
-			len(offenders), strings.Join(offenders, "\n  "))
+			var offenders []string
+			for _, r := range got {
+				if strings.HasPrefix(r, "Middleware:") || strings.Contains(r, "--> Middleware:") {
+					offenders = append(offenders, r)
+				}
+			}
+			if len(offenders) > 0 {
+				t.Errorf("a two-argument `.Use(A, B)` line emitted %d edge(s) touching a Middleware endpoint; "+
+					"`Use(A, B)` registers two middlewares, it does not relate A to B:\n  %s",
+					len(offenders), strings.Join(offenders, "\n  "))
+			}
+
+			if !slices.Contains(got, tc.control) {
+				t.Errorf("positive control missing: this fixture no longer emits %q, so the "+
+					"absence assertion above proves nothing. Relationships emitted:\n  %s",
+					tc.control, strings.Join(got, "\n  "))
+			}
+		})
 	}
 
 	// Name the exact edge the four deleted rules produced, so a reader of a
 	// future failure can tell this defect returning from some other rule
 	// growing a Middleware endpoint.
 	const fabricated = "Middleware:middleware.Logger --ROUTES_TO--> Controller:middleware.Recoverer"
-	for _, r := range got {
+	for _, r := range rels6914(detect6914Go(t, use6914TwoArgDotted)) {
 		if r == fabricated {
 			t.Errorf("the #6914 fabricated edge is back verbatim: %s", fabricated)
 		}
--- a/internal/engine/go_use_middleware_rule_6914_test.go
+++ b/internal/engine/go_use_middleware_rule_6914_test.go
@@ -1,0 +1,207 @@
+package engine
+
+import (
+	"context"
+	"fmt"
+	"sort"
+	"strings"
+	"testing"
+
+	"github.com/cajasmota/grafel/internal/extractor"
+)
+
+// #6914 (step 1): gin, chi, echo and fiber each carried the SAME relationship
+// rule
+//
+//	pattern:      '\.Use\s*\(\s*(\w[\w.]*)\s*(?:,\s*(\w[\w.]*))?\s*\)'
+//	source_type:  Middleware
+//	target_type:  Controller
+//	relationship: ROUTES_TO
+//	source_group: 1
+//	target_group: 2
+//
+// whose second capture group is optional. Driven through the real
+// LoadAllRules() + Detector.Detect() path on the fixtures below, the rule was
+// measured to be wrong in every direction it could take:
+//
+//	r.Use(middleware.Logger)                     -> group 2 empty, detector.go
+//	                                                `continue`s: no edge, ever.
+//	r.Use(middleware.Logger, middleware.Recoverer)
+//	                                             -> FOUR copies (one per rule
+//	                                                file) of
+//	                                                Middleware:middleware.Logger
+//	                                                --ROUTES_TO-->
+//	                                                Controller:middleware.Recoverer
+//	r.Use(a, b, c)                               -> matches nothing.
+//	r.Use(gin.Recovery())                        -> `(\w[\w.]*)\s*\)` cannot
+//	                                                match `()`: invisible.
+//
+// The two-argument edge is FABRICATED: `Use(A, B)` registers two middlewares
+// on the router; it does not make A route to B. And it fires on ordinary chi
+// source, so it is not a theoretical case. The four rules were deleted.
+//
+// These tests pin the ARTEFACT — the edges the shipped rules actually emit for
+// a `.Use` line — not the rule count and not the YAML text. A count-of-rules or
+// grep-the-YAML assertion passes the moment someone re-adds the edge in a
+// different spelling; asserting the emitted graph does not.
+//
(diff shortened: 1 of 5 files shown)


cajasmota:
## Independent adversarial review — **REQUEST-CHANGES** (small: ~6 lines of test)

Tree: `/private/tmp/claude-501/-Users-jorgecajas-Projects-archigraph/c0fefed4-0195-4c81-9a94-07a7145710a4/scratchpad/review-7006-z3q/wt`, detached at `612da2c5d`, macOS/APFS. All numbers below are from that tree, `-count=1`, exit codes read from redirected files.

The **deletion is correct and I could not find a correct edge it removes.** What I am asking for is the pin: I have a mutant that re-adds this exact defect, fires on idiomatic gin source, and the whole `./internal/engine/` suite stays green.

### (a) Does any `.Use` shape produce a correct edge this deletion removes? — **No.**

Not by reading: I enumerated **20 `.Use` shapes** through the real `LoadAllRules()` + `Detect()` path and diffed the full `(entities, relationships)` output at `612da2c5d` against the same probe with the four rules restored from `612da2c5d^`. Shapes covered: 1/2/3-arg, call expression, method value (`r.Use(s.Auth)` and 2-arg), chained (`r.Group("/x").Use(A, B)`), multi-line, slice spread (`mws...`), generic instantiation (`Wrap[T]`), `Use` on a **non-router** (`db.Use(tracing.Plugin, otel.Plugin)`), fiber's `app.Use("/api", h)` string-prefix form, echo/fiber/gin idioms, 3-level qualified names, two `.Use` lines in one file, and a realistic chi `main.go` with a route.

Every single difference between before and after is a `Middleware:A --ROUTES_TO--> Controller:B` edge from a **two-argument** `.Use`, ×4 (relationships are **not** deduped, only entities are — `detector.go` `seenEntities` covers entities only). Nothing else changed anywhere. Two extras worth recording:

- `db.Use(tracing.Plugin, otel.Plugin)` — a **GORM plugin registration**, no router in sight — produced `Middleware:tracing.Plugin --ROUTES_TO--> Controller:otel.Plugin` ×4 before the deletion. The rule was not merely wrong on routers, it fired on any `.Use` in any Go file. Extra support for the deletion.
- Out of scope but noting for step 4 / #7007: the surviving **entity** rule mints `Middleware:mws...` from `r.Use(mws...)` — a garbage name. Unchanged by this PR.

### (b) Does the pin assert the artefact, and survive a rewritten rule? — **Artefact yes; survives *some* rewrites, sleeps through two.**

It reads `FromID`/`ToID` off the emitted relationships, so it is a real artefact assertion, it is scoped (only edges touching a `Middleware` endpoint, not a whole-output grep), and it does **not** over-reach (M-R4 below). But:

| # | mutant (landing proved by exact count + hit set across all four YAMLs) | verdict |
|---|---|---|
| **M-R1** | re-add the rule **verbatim to two files** (chi + echo; `source_type: Middleware` counts chi 1 / echo 1 / fiber 0 / gin 0) | **DEAD** — 2 offending edges, both assertions fired |
| **M-R2** | re-add to **gin only**, same pattern/groups, but `source_type: Handler`, `target_type: Controller`, `relationship: CALLS` (counts `source_type: Handler` gin 1, others 0) | **ALIVE — distinguishable and ungraded.** Green on `-run TestIssue6914` *and* on the full `./internal/engine/` package (0 failures). The kind-based assertion is exactly what makes it blind to a re-add that renames the kinds |
| **M-R3** | re-add to **fiber only** with `target_group: 5` (out of range; count fiber 1, others 0) | **equivalent** — `extractGroup` returns `""`, detector `continue`s; the 20-shape probe output is byte-identical to HEAD. No input can distinguish it |
| **M-R4** | flip the **surviving** chi route rule `relationship: ROUTES_TO` → `LINKS_TO` (chi LINKS_TO 1 / ROUTES_TO 1, others 0) | **ALIVE and correct** — the pin does **not** fire on an unrelated edge. No over-reach |
| **M-R5** | re-add to **gin only**, identical types/kind/groups, pattern narrowed to bare identifiers: `'\.Use\s*\(\s*(\w+)\s*,\s*(\w+)\s*\)'` (fixed-string count gin 1, others 0) | **ALIVE and production-reachable** — see below |

**M-R5 is the blocker.** It is the same defect, same `source_type: Middleware` / `target_type: Controller` / `ROUTES_TO` / groups 1,2 — only the identifier pattern narrowed to the house style already used by chi's own route rule (`\s*,\s*(\w+)`). It emits

```
Middleware:AuthRequired --ROUTES_TO--> Controller:Logger
```

on `v1 := r.Group("/v1"); v1.Use(AuthRequired, Logger)` — textbook gin middleware, in the framework whose file I mutated. `go test -count=1 -run TestIssue6914 ./internal/engine/` → **ok**. The pin sleeps because its only two-argument fixture uses **dotted** names (`middleware.Logger`), so the single fixture pins one axis (arity) and leaves its neighbour (identifier shape) open.

**Also: the main pin has no positive control.** `use6914TwoArg` emits **zero relationships in total** at HEAD (measured — `ENTS=[Config:chi.NewRouter()] RELS=[]`). If that constant ever stopped producing anything, `TestIssue6914_TwoArgUseEmitsNoMiddlewareEdge` still passes. `TestIssue6914_UseEntityRuleStillMintsMiddleware` is a positive control for *the load path*, but on a different source constant.

**Asked-for change (~6 lines, no new file):** table-drive the two-arg pin over two fixtures — the existing dotted one plus `v1.Use(AuthRequired, Logger)` (kills M-R5) — and give it a positive control, e.g. assert the realistic-chi variant still yields its `Route:/health --ROUTES_TO--> Controller:healthHandler` edge so the "no Middleware edge" claim cannot go vacuous. M-R2 (renamed kinds) I would leave ungraded and say so in the comment: killing it means asserting "this `.Use` line emits no relationship at all", which is true today but would fight step 2's intended `Middleware REGISTERED_ON <app>` edge. Naming it beats silently not covering it.

### (c) Declining to grade M3b — **right call, and the reason is stronger than stated.**

The four `Middleware` **entity** rules are byte-identical (`pattern: '\.Use\s*\(\s*(\w[\w.]*)\s*\)'`, `entity_type: Middleware`, `name_group: 1`, `scope: file`), none carries `requires_framework`, and `detector.go` sets the entity's `framework` property to **`lang`** (`framework: lang` in `compile()`), not to the producing file — so the emitted record is identical in Kind, Name, SourceFile, StartLine and every property regardless of which of the four files produced it, and the dedup key `entityType:name:file` collapses them. This is not "no observable I could find"; it is an **equivalent mutant by construction — no `.go` input can distinguish it.** Right to decline here; #7007 is the right home.

### (d) Did any ratchet baseline move? — **No, and I checked it independently of the runner's line.**

`scripts/quality/run.sh --ratchet` (fresh `GRAFEL_HOME` + `GRAFEL_DAEMON_ROOT`) → exit **0**, "38 gated fixtures held their baseline (29 at 100%)". `git status --porcelain --untracked-files=no` → **0 tracked modifications**, so no baseline file was rewritten. Then, rather than trusting that line, I parsed all 38 baseline entries against the 38 fresh reports: **228 field comparisons**, with a positive control confirming the comparison can fail. `entity_found`, `entity_expected`, `relationship_found`, `relationship_expected` — **zero diffs everywhere**.

Your three ratchet claims reproduce exactly:
1. one fixture source line touches `.Use` — `go-chi-mini/src/main.go:19` (grep over the whole golden tree returns that line and `expected.json:28`, nothing else);
2. one expected row names `Middleware` — `go-chi-mini/expected.json:28`, an entity row;
3. parsed, not grepped: **527 `expected_relationships` rows across 38 fixtures, zero name `Middleware`** in any of `from_kind/from_name/to_kind/to_name/to_bare_name`. The golden gate really is blind here, which is why the unit pin has to carry the weight.

One thing to flag that is **not** yours: 15 of the 228 comparisons differ on `entity_extracted_total` / `relationship_extracted_total` (all **drops**) in 8 fixtures — csharp-aspnet-core / hangfire / quartz-net, docker-compose-services, python-django / fastapi / rq, razor-blazor. `--ratchet` does not gate those fields, so it prints OK anyway. Every one of those fixtures has **0 `.go` files** and no `.Use` line, and Go rule files load only for `language: go`, so this PR cannot be the cause; `go-chi-mini` (the only Go fixture with `.Use`) held all six fields exactly (71 / 207). Pre-existing baseline staleness in an ungated field — worth its own issue.

### Verification run here (all at `612da2c5d`, worktree restored by `cp` from shasum-verified backups after every mutant; `git status --porcelain` empty before each verification run)

- `gofmt -l internal/engine/` — 0 lines; `go vet ./internal/engine/` — exit 0
- `go test -count=1 ./internal/engine/` — **ok**, 42.4s
- `go test -count=1 ./cmd/grafel/` — **ok**, 169.2s, fresh isolated `GRAFEL_HOME` + `GRAFEL_DAEMON_ROOT`
- `scripts/quality/run.sh --ratchet` — exit 0, plus the independent 228-comparison check above

Flip to APPROVE as soon as the two-argument pin covers the bare-identifier shape.

cajasmota:
## Coordinator mutant on `08e8db3a7` — **CQ-1 is ALIVE and production-reachable**

Own detached worktree at `/private/tmp/.../coord-mut-7006-w4k/wt` (absolute). Baseline `go test -count=1 ./internal/engine/` exit 0 immediately before (`ok 48.4s`). Restored by `cp` from a SHA-256-verified backup (`791a02db…` both ways), `git diff HEAD` and `git status --porcelain` both empty afterwards.

The widened pin is a real improvement and M-R5 is genuinely dead. But the fix moved the fixture along **one** axis — identifier shape — and I went looking for the next neighbour, because that is the class this PR has now been caught by twice.

### CQ-1 — re-add the defect matching only the MULTI-LINE form

Appended to `chi.yaml` (anchor count asserted 1 before the edit; `MUTANT CQ-1` occurrence count 1 after; diff confirms an 8-line addition and nothing else):

```yaml
  - pattern: '\.Use\s*\(\s*(\w[\w.]*)\s*,\s*\n\s*(\w[\w.]*)'
    source_type: Middleware
    target_type: Controller
    relationship: ROUTES_TO
    source_group: 1
    target_group: 2
```

Identical `source_type` / `target_type` / `relationship` / groups to the four deleted rules. The only change is that the pattern **requires a newline between the arguments**, so it fires on the multi-line call and not on a single-line one.

**Result: ALIVE.** `go test -count=1 ./internal/engine/` exit 0, zero `--- FAIL` lines.

### It is reachable on the canonical form, demonstrated not argued

```go
r.Use(
    middleware.RequestID,
    middleware.Logger,
)
```

emits, under the mutant:

```
Middleware:middleware.RequestID --ROUTES_TO--> Controller:middleware.Logger
```

That is **gofmt's own output** for a chi middleware stack of more than one or two entries — the form every real chi service with a non-trivial stack is written in. It is not an exotic shape.

### Why the pin misses it

Both legs of the two-argument table — `use6914TwoArgDotted` and `use6914TwoArgBare` — put the two arguments **on one line**. The table now varies identifier shape (dotted vs bare) while holding **line layout** constant, so a re-add that only matches across a newline is invisible. Same defect class as M-R5, one axis over.

Worth noting *why* this is not covered by the original rules' own behaviour: the deleted patterns used `\s*` between arguments, and in RE2 `\s` **matches `\n`** — so the original rules did fire on the multi-line form, and the two-argument fixture happens to catch them only because it also fires on one line. The pin's coverage of multi-line was incidental, never asserted, and a narrower re-add drops straight through it.

### Asked change — ~2 lines

Add a third leg to the existing table: the same two-argument content in gofmt's multi-line layout, with its own positive control. It reuses the machinery that is already there. I would not go further than that — enumerate layout as a third axis and stop; the goal is a pin that survives the obvious rewrites, not one that anticipates every regex.

### One incidental observation

The probe output shows the control edge twice:

```
Route:/health --ROUTES_TO--> Controller:healthHandler
Route:/health --ROUTES_TO--> Controller:healthHandler
```

That is **#7007** again — duplicate rules across the Go framework files emitting the same relationship more than once, since relationships are not deduped while entities are. Not this PR's problem, and it does not affect the assertions here (`slices.Contains`), but it is a second live instance of that finding and worth having on the record.

### On the rest of the round

The positive controls are right, and the comment stating that the control is **liveness only, not a pin on a specific route rule** — because several Go rule files match the same route line — is exactly the kind of honest scoping that stops a passing control being misread later. Naming M-R2 as knowingly uncovered, with the reason that killing it would fight step 2's intended `REGISTERED_ON` edge, is better than silently leaving the gap.

Restating M3b as an **equivalent mutant by construction** rather than "hard to observe" is the accurate claim: `detector.go` sets `framework` from `lang`, not the producing file, so no `.go` input can distinguish which of the four rules minted the entity.

cajasmota:
## CQ-1 re-scored on `851b40b29` by me — **DEAD**

I found this one, so I confirmed the kill myself rather than taking the re-score on report.

Same worktree (`/private/tmp/.../coord-mut-7006-w4k/wt`), checked out at `851b40b29`, `chi.yaml` verified byte-identical to my earlier backup (`791a02db…`) before starting — so the mutant went onto clean, unmutated code. Green baseline immediately before, exit 0.

Re-applied CQ-1 verbatim (the multi-line-only re-add, occurrence count 1):

**Result: DEAD.**

```
--- FAIL: TestIssue6914_TwoArgUseEmitsNoMiddlewareEdge
    --- FAIL: TestIssue6914_TwoArgUseEmitsNoMiddlewareEdge/gofmt_multi-line_layout_(chi_stack_of_three)
```

Failing on **exactly** the new leg and nothing else — the dotted and bare legs pass, so the new axis is doing the killing rather than some incidental breakage. That is the shape I wanted: a kill attributable to the assertion that was added for it. Restored by `cp`, `git diff HEAD` and `git status --porcelain` empty afterwards.

The pin now covers three axes — arity, identifier shape, line layout — each with its own positive control. Stopping there is right: the goal is a pin that survives the obvious rewrites, not one that anticipates every regex, and `M-R2` stays deliberately and visibly uncovered because killing it would fight step 2's intended `REGISTERED_ON` edge once #6916 lands.

Worth keeping in the record: **the gap existed because the original rules' multi-line coverage was incidental.** They used `\s*` between arguments and RE2's `\s` matches `\n`, so they always fired on the multi-line form — but no fixture ever asserted it, which is exactly how a narrower re-add walked through. Coverage that arrives as a side effect of a regex detail is not coverage.

**Merging on green.**

PR #7008, opened 2026-09-08
test(#6998): pin lookupStructural's cross-file unique-name tier to python — deleting the lang guard was ALIVE against the whole golden ratchet

### LINKED ISSUE #6998: lookupStructural's cross-file unique-name fallback is restricted to lang==python by an assertion nobody wrote — removing the guard is ALIVE, including against the full golden ratchet
Found by the independent review of #6994 (#6986 arm 1). **Pre-existing code — not that PR's debt**, filed so it is on the record rather than rediscovered.

## The gap

`internal/resolve/refs.go`'s `lookupStructural` resolves a `scope:component:...` address in two tiers: same-file `lookupLocationKind`, then — **for `component` scope with `lang == "python"` only** — `lookupUniqueRealComponentByName`, which binds when exactly one entity in the whole graph declares the name.

**Removing that `lang == "python"` guard is ALIVE.** Nothing fails: not the resolver package, not the touched extractor packages, and **not the full golden ratchet**. So the restriction of a cross-file, whole-graph, unique-name fallback to a single language is asserted by nothing.

## Why it matters

This is the widest tier in the structural lookup. It is the one thing that makes the Python address dialect *not* same-file-only, which is exactly what distinguishes it from the C# (#6984) and protobuf (#6991) ports — both of which are deliberately same-file because a cross-file guess is #6369's wrong-node hazard.

If the guard were removed or drifted, **every other language's structural addresses would silently gain a whole-graph unique-name fallback**. That is a resolution-semantics change affecting the entire graph, and it would ship green.

The direction matters: this is the **permissive** direction. A binding that should not happen produces a confidently-wrong edge, which reads as valid and is worse than a dangling one.

## Scope, honestly stated

- The mutant is **ALIVE and distinguishable**; whether it is *reachable* in the sense of anyone actually removing the guard is a question about future edits, not about today's graph. **Today's behaviour is correct** — this is a missing assertion, not a defect.
- It is **not** gated on the custom lane: `lookupStructural` is core resolution, so unlike #6986's other findings this one is **not** capped by the #6966 default-off gate. Any regression here would be user-facing.
- Nobody has measured what would actually change if the fallback applied to other languages. **That measurement is worth having before anyone "fixes" the guard in either direction** — it may be that some other language wants it, and the right outcome is a deliberate decision rather than a pin on the status quo.

## What a fix has to do

1. **Assert the restriction by behaviour**, not by reading the source: a non-Python structural address whose name is globally unique but declared in another file must **not** bind. Name the entities.
2. **Assert the Python case still does bind** — a positive control, so the test cannot pass by the tier never firing at all.
3. Do not weaken it into a source-scan for the string `lang == "python"`. **A scan-and-assert guard has several no-op modes**; this needs the behavioural pair.
4. If the measurement in "Scope" shows another language should have the tier, that is a separate, deliberate change with its own both-directions grading (#6369) — not a side effect of adding a test.

Refs #6994, #6986, #6984, #6991, #6369, #6966.

cajasmota:
Closes #6998.

**Behaviour is unchanged. This is the missing assertion, not a fix.** The `lang == "python"` guard at `internal/resolve/refs.go:3098` is exactly as it was.

## What was ungraded

`lookupStructural` resolves a `scope:component:...` address in two tiers: same-file `lookupLocationKind`, then — for `component` scope with **`lang == "python"` only** — `lookupUniqueRealComponentByName`, which binds when exactly one entity in the whole graph declares the name. That second tier is the widest one in the structural lookup, and the only thing making the Python address dialect *not* same-file-only — which is precisely what separates it from the C# (#6984) and protobuf (#6991) ports, both deliberately same-file because a cross-file guess is #6369's wrong-node hazard.

Removing the guard was **ALIVE**: the resolver package, the touched extractor packages and the full golden ratchet all stayed green. So if the guard drifted, every other language would silently gain a whole-graph unique-name fallback and it would ship green. That is the **permissive** direction — a binding that should not happen is a confidently-wrong edge, which reads as valid and is worse than a dangling one. Not capped by the #6966 default-off gate: `lookupStructural` is core resolution.

## The pin — a behavioural pair, in `internal/resolve/structural_component_lang_guard_6998_test.go`

`TestStructuralComponentCrossFileTier_IsPythonOnly6998`, table-driven over one fixture builder:

| case | address lang | expected |
|---|---|---|
| `python` (**positive control**) | `python` | binds to `SharedBase` in `app/base.py`, `statusRewritten` |
| `Python-mixed-case` (**second positive control**) | `Python` | **binds** — see below |
| `csharp` | `csharp` | no binding, `statusUnmatched` |
| `protobuf` | `protobuf` | no binding, `statusUnmatched` |
| `php` | `php` | no binding, `statusUnmatched` |
| `python3` | `python3` | no binding, `statusUnmatched` |
| `ruby` | `ruby` | no binding, `statusUnmatched` |
| `java` | `java` | no binding, `statusUnmatched` |
| `ruby-on-python-paths` | `ruby`, with the python case's **exact file paths** | no binding, `statusUnmatched` |
| `config-scope` | scope `config`, lang `python` | no binding, `statusUnmatched` |
| `operation-scope` | scope `operation`, lang `python` | no binding, `statusUnmatched` |

**Round 2 — the negatives are an enumeration, not a sample.** The first version graded `ruby` and `java`, which pins "not ruby, not java", not "python only". A negative set of two members cannot detect a predicate that got *broader* — #6998's own shape. The review scored three widenings ALIVE against it: `HasPrefix(lang, "py")`, `HasPrefix(lang, "p")` (**one character**, admitting `php`/`protobuf`/`proto`, all live language tokens), and the guard extended with `csharp`/`typescript`/`go`. `csharp` and `protobuf` are exactly the ports this PR cites as deliberately same-file (#6984, #6991), so a drift admitting either is #6369's wrong-node hazard in the permissive direction, arriving green. Those four rows close it.

**Round 3 — SCOPE is now a graded axis too.** Round 2's nine rows all varied the language and all held the scope segment at `component`, so nothing asserted the **other** guard on that line, `refs.go:3094`'s `strings.EqualFold(scopeKind, "component")`. Widening it to `strings.HasPrefix(strings.ToLower(scopeKind), "co")` (CS-1) was ALIVE — and reachable on an address production emits today: `internal/extractors/python/config_consumer.go:333` mints `scope:config:ref:python:<file>:<Name>`, and `config` starts with `co`. Under the mutant that address **bound to a class in another file** — #6369's wrong-node hazard in the permissive direction, arriving green. Two rows close it, with `lang` held at `python` so a failure is unambiguously about scope. Deliberately two, not the scope vocabulary.

The subtype segment stays `class` rather than production's `ref`: `lookupStructural` never reads it — the only segments it indexes are `stubScopeKindIndex(1)`, `stubScopeLangIndex(3)`, `stubScopeFileIndex(4)`, `stubScopeTailIndex(5)`, and there is no `parts[2]` in the function — so `scope:config:ref:…` and `scope:config:class:…` take an identical path. Holding it constant keeps scope the only varied axis at no cost in fidelity.

**Round 2 — the invariant is case-INSENSITIVE python-only.** Mixed-case `Python` binds, via the `strings.ToLower(parts[stubScopeLangIndex])` on the line above the guard. That was ungraded, and a reader would reasonably have assumed an exact byte match. `Python-mixed-case` is a second positive control and it is what kills the narrowing mutant R6.

The entities are named in the assertions (`SharedBase` in `app/base<ext>`, consumer `Child` in `app/child<ext>`); no count is asserted. The positive control is what stops the negatives passing by the tier never firing at all.

Explicitly **not** a source scan for the string `lang == "python"` — a scan-and-assert guard has several independent no-op modes and would survive the very rewrite it exists to catch.

### Axes

**Varied:** the address's language segment, and the entity `Language` / file extension that go with it.
**Held constant:** entity kind (`Class`) and subtype, the declared name (`SharedBase`), its global uniqueness, the cross-file relationship (the consumer's file declares nothing of that name), the directory (so `pkgDirOf` is identical), and the address's scope-kind and subtype segments.
The `ruby-on-python-paths` case holds the **file paths** constant too, so the only thing separating it from the binding case is the language segment itself — the guard cannot be mistaken for a path- or extension-keyed effect.

Of that joint variation the **address's** language segment is the load-bearing half: the entity's `Language` field is not read by this tier at all. The review probed it both ways — entity `Language=ruby` with a python address binds; entity `Language=python` with a ruby address does not. The fixture varies both only because that is what a realistic record looks like. Also enumerated and confirmed: an **empty** language segment dangles honestly, taking no third path (the `lang == ""` conditions near `refs.go:4574` are disposition routing in a different function, not a binding tier). Both facts are now recorded in the test comment.

### Both premises are asserted per case, before the outcome

1. `lookupUniqueRealComponentByName("SharedBase")` returns the base class in **every** case, binding and non-binding alike. So a negative can never pass because the candidate was absent or ambiguous: the only thing withholding the binding is the language guard.
2. `lookupLocationKind(consumerFile, "SharedBase", componentKindFamily)` must **miss**, so the same-file tier is not what the case is measuring.

## Mutants

Both proven applied by an **exact occurrence count of 1** at `refs.go:3098`, with the enclosing function asserted to be `lookupStructural` — `lang == "python"` appears **12 times** in that file, so a naive replace lands on the wrong conjunct. Green baseline immediately before each score, `go vet` and `gofmt -l` clean, `-count=1` always, restored by `cp` from a shasum-verified backup (`326b6412…` before and after), zero `MUTANT` text remaining, `git diff HEAD` empty on the production file.

- **M1 — the issue's mutant: delete the guard** (`if true {` in place of `if lang == "python" {`). **DEAD.** `./internal/resolve/` goes RED on `TestStructuralComponentCrossFileTier_IsPythonOnly6998` — subtests `ruby`, `java` and `ruby-on-python-paths` fail; `python` still passes.
- **M2 — complementary: the python tier made same-file-only** (`lookupLocationKind(filePath, tail, componentKindFamily)` in place of `lookupUniqueRealComponentByName(tail)`, guard kept). **DEAD, and killed by the `python` subtest ALONE** — the three negatives still pass. So the pin is not satisfiable by an unrelated failure, and the positive control is load-bearing rather than decorative.

- **R2 — `strings.HasPrefix(lang, "py")`**. Was ALIVE; now **DEAD**, killed by `python3`.
- **R3 — `strings.HasPrefix(lang, "p")`**, the one-character widening. Was ALIVE; now **DEAD**, killed by `protobuf`, `php` and `python3`.
- **R4 — guard extended with `csharp`, `typescript`, `go`** (`go` even pre-empts the package-scoped go tier ten lines below). Was ALIVE; now **DEAD**, killed by `csharp`.
- **R6 — narrowing: `parts[stubScopeLangIndex] == "python"`**, i.e. the case fold dropped. **DEAD**, killed by `Python-mixed-case` **alone** — so that row is load-bearing, not decoration.

- **CS-1 — the SCOPE guard at `refs.go:3094` widened to a `co` prefix**. Was ALIVE; now **DEAD**, killed by `config-scope`.
- **CS-2 — the scope guard extended with `operation`**. **DEAD**, killed by `operation-scope`.

No row is decoration: `python3` kills R2 and R3, `protobuf` and `php` kill R3, `csharp` kills R4, `Python-mixed-case` kills R6, `config-scope` kills CS-1, `operation-scope` kills CS-2, and `ruby`/`java`/`ruby-on-python-paths` kill M1. M1 and R3 were re-scored after the round-3 fixture change to confirm the language axis did not regress. The two guards sit one line apart and were addressed by distinct line numbers — scope at `:3094` (that token occurs once), language at `:3098` (`lang == "python"` occurs 12 times).

Round-2 mutants were applied by a script that asserts, and prints, all of: `lang == "python"` occurs exactly **12** times in `refs.go`; the target line is `refs.go:3098`; its enclosing `func` is `lookupStructural`; and **exactly one** line differs from the backup (the hit set itself is printed).

Worth recording, and **independently reproduced in review**: under M2 **no other test in `./internal/resolve/` failed** — notably not the Django FK path, which `django_fk.go:172` documents as depending on this tier. So before this PR the cross-file tier's *firing* was ungraded in that package, not merely its language restriction. This PR closes the firing half outright. That gap is being filed separately and is not chased here.

## Verification

All in `/private/tmp/claude-501/-Users-jorgecajas-Projects-archigraph/c0fefed4-0195-4c81-9a94-07a7145710a4/scratchpad/impl-6998-q4v9x/wt` (worktree at `82396b1ca`). macOS/APFS, Apple Silicon, `GOMAXPROCS=3`, `nice -n 10`, isolated `GRAFEL_HOME` + `GRAFEL_DAEMON_ROOT` per run.

- `go vet ./internal/resolve/` — clean; `gofmt -l internal/resolve/` — empty.
- `go test -count=1 ./internal/resolve/` — **ok** (0.9s), every round.
- `go test -count=1 ./cmd/grafel/` — **ok** (163s), fresh isolated home, on the round-1 tree. **Not re-run for round 2, deliberately**: nothing outside the `_test.go` changed, and a test file is not linked into the `grafel` binary the ratchet grades.
- `bash scripts/quality/run.sh --ratchet` — **exit 0** in all three rounds, `38 gated fixtures held their baseline (29 at 100% must-have recall)`. No baseline moved; no constant touched.

## Out of scope, on purpose

Nobody has measured what would change if the fallback applied to other languages, and #6998 says that measurement is worth having **before** anyone moves the guard in either direction. I did not move it and I did not produce that measurement — scoring M1 against `./internal/resolve/` alone is not it. If some language wants the tier, that is a deliberate change with its own both-directions grading (#6369), not a side effect of adding a test.

Refs #6994, #6986, #6984, #6991, #6369, #6966.

### DIFF
--- a/internal/resolve/structural_component_lang_guard_6998_test.go
+++ b/internal/resolve/structural_component_lang_guard_6998_test.go
@@ -1,0 +1,147 @@
+package resolve
+
+import (
+	"testing"
+
+	"github.com/cajasmota/grafel/internal/types"
+)
+
+// #6998 — lookupStructural's cross-file, whole-graph unique-name fallback
+// (lookupUniqueRealComponentByName) fires for `component`-scope structural
+// addresses with lang == "python" and for no other language. Removing that
+// guard was ALIVE: the resolver package, the touched extractor packages and
+// the full golden ratchet all stayed green, so the restriction of the widest
+// tier in the structural lookup to one language was asserted by nothing.
+//
+// Today's behaviour is correct; this file is the missing assertion. The tier
+// is what makes the Python address dialect NOT same-file-only, which is
+// exactly what separates it from the C# (#6984) and protobuf (#6991) ports —
+// both deliberately same-file because a cross-file guess is #6369's
+// wrong-node hazard. Widening it would give every language a whole-graph
+// unique-name fallback: the permissive direction, where a binding that should
+// not happen becomes a confidently-wrong edge that reads as valid.
+//
+// The pin is behavioural, not a source scan for the literal `lang ==
+// "python"` — a scan-and-assert guard has several independent no-op modes and
+// would survive the very rewrite it exists to catch.
+//
+// AXES. Varied: the address's language segment (and the entity Language /
+// file extension that go with it). Held constant: the entity kind and
+// subtype, the declared name, the global uniqueness of that name, the
+// cross-file relationship (the consumer's file declares nothing), the
+// directory (so pkgDirOf is identical), and the address's scope-kind and
+// subtype segments. Case `ruby-on-python-paths` additionally holds the FILE
+// PATHS constant too, so the only thing separating it from the binding case
+// is the language segment itself — the guard cannot be mistaken for a
+// path- or extension-keyed effect.
+
+// langGuardFixture6998 builds the one shape both halves of the pair share:
+// a class `SharedBase` declared in `app/base<ext>`, globally unique, and a
+// consumer file `app/child<ext>` that declares NOTHING. The structural
+// address is minted with the CONSUMER's file — the Python dialect's
+// convention (the hierarchy extractor addresses `class Child(SharedBase):`
+// by the file the subclass declaration lives in), so the same-file
+// lookupLocationKind tier necessarily misses and only the cross-file tier
+// can bind.
+func langGuardFixture6998(lang, ext string) (base types.EntityRecord, consumer types.EntityRecord, ref string) {
+	baseFile := "app/base" + ext
+	childFile := "app/child" + ext
+	base = types.EntityRecord{
+		ID: "c0de6998ba5e0001", Kind: "Class", Subtype: "class",
+		Name: "SharedBase", QualifiedName: "SharedBase",
+		SourceFile: baseFile, Language: lang,
+	}
+	// The consumer entity exists only so the consumer file is a real file in
+	// the index with a real, DIFFERENT name in it — the same-file tier has
+	// something to miss on rather than nothing at all.
+	consumer = types.EntityRecord{
+		ID: "c0de6998c41d0002", Kind: "Class", Subtype: "class",
+		Name: "Child", QualifiedName: "Child",
+		SourceFile: childFile, Language: lang,
+	}
+	ref = "scope:component:class:" + lang + ":" + childFile + ":SharedBase"
+	return base, consumer, ref
+}
+
+func TestStructuralComponentCrossFileTier_IsPythonOnly6998(t *testing.T) {
+	cases := []struct {
+		name     string
+		lang     string
+		ext      string
+		wantBind bool
+	}{
+		// The positive control. Without it the three negatives below could
+		// all pass by the tier never firing at all, which is the commonest
+		// way an absence assertion here turns out to be decoration.
+		{name: "python", lang: "python", ext: ".py", wantBind: true},
+		{name: "ruby", lang: "ruby", ext: ".rb", wantBind: false},
+		{name: "java", lang: "java", ext: ".java", wantBind: false},
+		// Language varied, file paths held identical to the python case.
+		{name: "ruby-on-python-paths", lang: "ruby", ext: ".py", wantBind: false},
+	}
+
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
+			base, consumer, ref := langGuardFixture6998(tc.lang, tc.ext)
+			idx := BuildIndex([]types.EntityRecord{base, consumer})
+
+			// Premise 1 — the cross-file tier's own precondition holds in
+			// EVERY case, binding and non-binding alike. `SharedBase` is
+			// globally unique, so lookupUniqueRealComponentByName would
+			// return the base class if it were reached. This is what makes
+			// the negatives sharp: the ONLY thing withholding the binding is
+			// the language guard, not an absent or ambiguous candidate.
+			if id, ok := idx.lookupUniqueRealComponentByName("SharedBase"); !ok || id != base.ID {
+				t.Fatalf("premise: lookupUniqueRealComponentByName(SharedBase) = (%q,%v), "+
+					"want (%q,true) — the cross-file tier could not have bound in this "+
+					"fixture for a reason unrelated to the language guard (#6998)",
+					id, ok, base.ID)
+			}
+
+			// Premise 2 — the same-file tier misses. The consumer's file does
+			// not declare `SharedBase`, so nothing but the cross-file tier
+			// can produce a binding here.
+			if id, ok := idx.lookupLocationKind(consumer.SourceFile, "SharedBase", componentKindFamily); ok {
+				t.Fatalf("premise: lookupLocationKind(%q, SharedBase) = (%q,true); the "+
+					"consumer file must not declare the name or the same-file tier, not "+
+					"the cross-file one, is what this case measures (#6998)",
+					consumer.SourceFile, id)
+			}
+
+			id, status, handled := idx.lookupStructural(ref)
+			if !handled {
+				t.Fatalf("lookupStructural did not claim %q (handled=false)", ref)
+			}
+
+			if tc.wantBind {
+				if id != base.ID {
+					t.Fatalf("lookupStructural(%q) = %q (status=%d), want the class "+
+						"%q declared in %q — the Python cross-file unique-name tier "+
+						"must still bind (#6998 positive control)",
+						ref, id, status, base.ID, base.SourceFile)
+				}
+				if status != statusRewritten {
+					t.Fatalf("lookupStructural(%q) status = %d, want statusRewritten=%d (#6998)",
+						ref, status, statusRewritten)
+				}
+				return
+			}
+
+			if id == base.ID {
+				t.Fatalf("lookupStructural(%q) bound the class %q declared in %q — a "+
+					"lang=%q structural address reached the whole-graph unique-name "+
+					"fallback, which is python-only. A cross-file guess in another "+
+					"language's dialect is a confidently-wrong edge (#6998, #6369)",
+					ref, base.ID, base.SourceFile, tc.lang)
+			}
+			if id != "" {
+				t.Fatalf("lookupStructural(%q) = %q, want no binding (#6998)", ref, id)
+			}
+			if status != statusUnmatched {
+				t.Fatalf("lookupStructural(%q) status = %d, want statusUnmatched=%d — the "+
+					"address must dangle honestly rather than resolve or report an "+
+					"ambiguity it did not find (#6998)", ref, status, statusUnmatched)
+			}
+		})
+	}
+}
--- a/internal/resolve/structural_component_lang_guard_6998_test.go
+++ b/internal/resolve/structural_component_lang_guard_6998_test.go
@@ -25,7 +25,11 @@ import (
 // "python"` — a scan-and-assert guard has several independent no-op modes and
 // would survive the very rewrite it exists to catch.
 //
-// AXES. Varied: the address's language segment (and the entity Language /
+// AXES. Two axes are graded here: the address's LANGUAGE segment and its
+// SCOPE segment (rows `config-scope` / `operation-scope`), each varied while
+// the other is held at the binding value.
+//
+// Varied: the address's language segment (and the entity Language /
 // file extension that go with it). Held constant: the entity kind and
 // subtype, the declared name, the global uniqueness of that name, the
 // cross-file relationship (the consumer's file declares nothing), the
@@ -59,7 +63,7 @@ import (
 // by the file the subclass declaration lives in), so the same-file
 // lookupLocationKind tier necessarily misses and only the cross-file tier
 // can bind.
-func langGuardFixture6998(lang, ext string) (base types.EntityRecord, consumer types.EntityRecord, ref string) {
+func langGuardFixture6998(scopeKind, lang, ext string) (base types.EntityRecord, consumer types.EntityRecord, ref string) {
 	baseFile := "app/base" + ext
 	childFile := "app/child" + ext
 	base = types.EntityRecord{
@@ -75,13 +79,14 @@ func langGuardFixture6998(lang, ext string) (base types.EntityRecord, consumer t
 		Name: "Child", QualifiedName: "Child",
 		SourceFile: childFile, Language: lang,
 	}
-	ref = "scope:component:class:" + lang + ":" + childFile + ":SharedBase"
+	ref = "scope:" + scopeKind + ":class:" + lang + ":" + childFile + ":SharedBase"
 	return base, consumer, ref
 }
 
 func TestStructuralComponentCrossFileTier_IsPythonOnly6998(t *testing.T) {
 	cases := []struct {
 		name     string
+		scope    string
 		lang     string
 		ext      string
 		wantBind bool
@@ -89,13 +94,13 @@ func TestStructuralComponentCrossFileTier_IsPythonOnly6998(t *testing.T) {
 		// The positive control. Without it every negative below could pass
 		// by the tier never firing at all, which is the commonest way an
 		// absence assertion here turns out to be decoration.
-		{name: "python", lang: "python", ext: ".py", wantBind: true},
+		{name: "python", scope: "component", lang: "python", ext: ".py", wantBind: true},
 		// Second positive control: the guard reads `strings.ToLower(...)`,
 		// so the invariant is case-INSENSITIVE python-only. A reader would
 		// otherwise reasonably assume an exact byte match, and a future
 		// edit that dropped the ToLower would be a silent narrowing that
 		// no other row here can see.
-		{name: "Python-mixed-case", lang: "Python", ext: ".py", wantBind: true},
+		{name: "Python-mixed-case", scope: "component", lang: "Python", ext: ".py", wantBind: true},
 		// The negatives are an ENUMERATION of the guard's neighbours, not a
 		// sample. Two sampled members cannot detect a predicate that got
 		// BROADER (#6998's own shape), and each of the following was scored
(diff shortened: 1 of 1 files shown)


cajasmota:

Measured in my own worktree at `/private/tmp/claude-501/-Users-jorgecajas-Projects-archigraph/c0fefed4-0195-4c81-9a94-07a7145710a4/scratchpad/review-7008-zx3/wt` (detached at `eb10faa29`). macOS/APFS, Apple Silicon, `GOMAXPROCS=3`, `nice -n 10`, `-count=1`. `go vet ./internal/resolve/` clean; baseline `ok` immediately before every score. Production file restored by `cp` from a shasum-verified backup (`786d695b…` before and after); `git diff HEAD` empty and `git status --porcelain` empty at the end.

Every edit to `refs.go` was applied by a script that asserts (a) `lang == "python"` occurs exactly **12** times in the file, (b) the target line is at the asserted index, (c) the enclosing `func …` line contains `lookupStructural` (or `lookupUniqueRealComponentByName`), and (d) **exactly one** line differs from the backup. No edit was scored without that print.

### Mutant table (mine, on top of the author's M1/M2)

| # | Edit at `refs.go:3098` (or as noted) | Result |
|---|---|---|
| R1 | `if lang != "python" {` (invert) | **DEAD** — all four subtests fail |
| R2 | `if strings.HasPrefix(lang, "py") {` | **ALIVE** — `./internal/resolve/` green |
| R3 | `if strings.HasPrefix(lang, "p") {` | **ALIVE** — green, and this admits `php`, `protobuf`, `proto` (all three are live language tokens in `refs.go`) |
| R4 | `if lang == "python" \|\| lang == "csharp" \|\| lang == "typescript" \|\| lang == "go" {` | **ALIVE** — green, including `go`, which pre-empts the package-scoped `lang == "go"` tier 10 lines below |
| R5 | `:3231` `return "", false` → `continue` in `lookupUniqueRealComponentByName` (bind on a **non-unique** name) | **DEAD**, but killed by the pre-existing `TestLocKindMerge_EntryPointEquivalence`, not by this PR |

R2–R4 are **distinguishable and ungraded**, and they are the permissive direction. Grading them is cheap — the table already exists; each is one more row (`csharp`, `protobuf`, `php`, `python3`). **I am not blocking on it**: R2–R4 are strictly weaker widenings than the author's M1 (`if true`), which this PR kills, so the PR moves the needle in the right direction from "nothing at all". Worth a follow-up row set given #6984 (C#) and #6991 (protobuf) are the exact ports the PR body cites as deliberately same-file — a drift that added `csharp` or `protobuf` to this guard would still ship green today.

### (a) Can a negative pass for the wrong reason? — **No.**

Both premises hold and I tried to break them. Premise 1 (`lookupUniqueRealComponentByName` returns the base class) is asserted per case and closes candidate-absent/ambiguous; premise 2 closes same-file-tier-shadowing; `handled` is asserted so a parse bail cannot masquerade as `statusUnmatched`; no other tier in the `component` branch is reachable for a non-`go`, non-`python` lang. I could not construct a wrong-reason pass.

### (b) Is language really the only varied axis? — **Substantively yes**, with one declared caveat that I verified is harmless.

I read the fixture builder, not the description. Kind/subtype/name/uniqueness/cross-file relation/directory/scope segments are genuinely held constant, and `ruby-on-python-paths` really does reuse the python case's exact paths (`app/base.py`, `app/child.py`) — the label matches its content.

The caveat: the entity `Language` field is set from the same `lang` in every case, so no case isolates the **address** language segment from the **entity** `Language`. I probed this directly (throwaway test, since deleted):

- entity `Language=ruby`, address lang `python` → **binds** (`status=1`)
- entity `Language=python`, address lang `ruby` → **no bind** (`status=3`)

So entity `Language` is provably irrelevant and the joint variation costs nothing. The body's phrase "the only thing separating it from the binding case is the language segment itself" is technically an overstatement for `ruby-on-python-paths`; it is true in effect.

### (c) Enumeration beyond the hand-picked arms

- **empty language** (`scope:component:class::…`) → no bind, `statusUnmatched`. It does **not** take a third path; the `(lang == "" || lang == "python")` guards at `:4574`/`:4588` are *disposition* routing in a different function, not a binding tier. Confirmed, not assumed.
- **mixed case** `Python` → **binds**, via the `strings.ToLower` on the line above. So the real invariant is *case-insensitive* python-only. Not a defect, but it is ungraded — no case covers it.
- `python3` → no bind (which is why R2 needs R3 to be clearly production-reachable; R3 is).

### (d) Is the M2 observation true? — **Confirmed, independently reproduced.**

I applied M2 myself (`:3099` → `lookupLocationKind(filePath, tail, componentKindFamily)`) and ran `go test -count=1 -v ./internal/resolve/`. The **only** failure is `TestStructuralComponentCrossFileTier_IsPythonOnly6998/python`. Nothing else in the package fails — notably not the Django FK path, which `django_fk.go:172` documents as depending on `lookupStructural → lookupUniqueRealComponentByName`. So yes: before this PR the cross-file tier's *firing* was ungraded in `./internal/resolve/`, not merely its language restriction. This PR closes the firing half outright (the positive control is load-bearing, exactly as claimed) and closes the language half for `ruby` and `java` only.

### (e) Did any ratchet baseline move? — **No, and it cannot have.**

`git diff --stat HEAD~1 HEAD` is exactly **one file, 147 insertions**: the new `_test.go`. A `_test.go` file is not linked into the `grafel` binary that `scripts/quality/run.sh` grades, so no fixture measurement and no `baseline.json` entry can change. That is a stronger statement than a green `--ratchet` run (which, per the brief, would not surface a ceiling drop anyway), so I did not spend the corpus run.

Likewise I did **not** re-score M1 against the ratchet: R4 and R3 are strict *subsets* of M1's widening, and M1 (`if true`) was already reported ALIVE there, so the ratchet cannot distinguish any of them. That is consistent, not a new gap.

### Verdict

**APPROVE.** The pin pins: it is behavioural rather than a source scan, the positive control is load-bearing (M2 dies on it alone), the negatives cannot pass vacuously, and the axis claims survive reading the fixture bodies. The guard was correctly left untouched and the "what would other languages gain" measurement was correctly not attempted here.

One follow-up worth filing (not blocking): extend the table with `csharp`, `protobuf`, `php` and a `python3`/`Python` pair, so the guard's *shape* (exact, case-folded match) is graded and not just two sampled negatives.

cajasmota:
## Coordinator mutant on `16cbc09cd` — **CS-1 is ALIVE and production-reachable**

Own detached worktree at `/private/tmp/.../coord-mut-7008-p2m/wt` (absolute). Green baseline immediately before (`./internal/resolve/` exit 0). Restored by `cp` from a SHA-256-verified backup (`326b6412…` both ways); `git diff HEAD` and `git status --porcelain` empty afterwards.

Round 2 is a real improvement — R2/R3/R4 are dead, each row kills something, and adding R6 unasked so the new `Python` positive control is itself graded is exactly right: a positive control nothing can kill is decoration.

But the enumeration went deep on **one** axis. Every fixture is `scope:component:...`. So I widened the *other* guard on that line.

### CS-1 — widen the SCOPE guard, not the language guard

`refs.go:3094`: `if strings.EqualFold(scopeKind, "component") {` → `if strings.HasPrefix(strings.ToLower(scopeKind), "co") {`

Edit proven landed: exactly **1** occurrence, hit set **`[line 3094]`**, one line differs from the backup, `go vet` clean.

**Result: ALIVE.** `./internal/resolve/` exit 0, zero failures.

### The distinguishing input is a shape production emits today

`config` starts with `co`. `internal/extractors/python/config_consumer.go:333` builds exactly:

```go
return "scope:config:ref:python:" + filepath.ToSlash(configName) + ":" + configName
```

Driving that address through `lookupStructural` under the mutant:

```
lookupStructural("scope:config:ref:python:app/child.py:Settings")
  = id="c0de7008ba5e0001" status=1(statusRewritten) handled=true
CONFIG ADDRESS BOUND TO THE CLASS "Settings" (app/base.py)
```

A **config reference binds to a class in another file**. That is #6369's wrong-node hazard in its worst form — a confidently-wrong edge, which reads as valid and is worse than a dangling one — and it is the precise failure mode this issue exists to prevent. `scope:config:` is live (4 producer sites, one of them python), so this is not a hypothetical scope kind.

### Why the current pin cannot see it

`langGuardFixture6998` hard-codes `ref = "scope:component:class:" + lang + ...`. Language varies across nine rows; **scope varies across none**. The negatives establish "not ruby, not java, not csharp, not protobuf, not php, not python3" — all statements about the *language* guard. Nothing on that line's *scope* guard is asserted at all.

This is the same shape as the widenings the reviewer found, one axis over — the third time on this board this week that a pin enumerated one axis thoroughly and left its neighbour open.

### Asked change — a few rows, not machinery

Give `langGuardFixture6998` a scope segment parameter and add negative rows holding `lang: python` constant while varying scope:

- **`config`** — kills CS-1 specifically, and pins a real production address shape.
- **`operation`** — the largest scope kind in the codebase (244 sites vs component's 214), and the comment immediately above this block says the *operation* path "only fires for operation scope", so the two are already entangled in the reader's mind. Worth pinning that they stay separate.

Two rows is enough. **Stop there** — I am not asking for the whole scope vocabulary; the goal is a pin that survives an obvious widening, not one that enumerates 20 scope kinds. And keep `lang: python` fixed in these rows so a failure is unambiguously about scope.

Re-score CS-1 and show it DEAD, naming which row kills it.

### One thing I checked so nobody repeats it

I verified `scope:config` is genuinely live before claiming reachability — `internal/extractor/config_key.go:54`, `internal/extractors/python/config_module.go:175`, `config_consumer.go:333`, `internal/extractors/config/discover.go:388`. The python one is what makes `lang == "python"` and a `co`-prefixed scope coincide in the same address, which is what the mutant needs.

cajasmota:
## CS-1 re-scored on `b8eedf707` by me — **DEAD**

I found this one, so I confirmed the kill myself rather than taking the re-score on report.

Same worktree (`/private/tmp/.../coord-mut-7008-p2m/wt`), checked out at `b8eedf707`, `refs.go` verified byte-identical to my backup (`326b6412…`) before starting — so the mutant went onto clean, unmutated code. Green baseline immediately before, exit 0. Occurrence count for the scope guard asserted **1**, hit set `[3094]`, one line differs.

**Result: DEAD.**

```
--- FAIL: TestStructuralComponentCrossFileTier_IsPythonOnly6998
    --- FAIL: TestStructuralComponentCrossFileTier_IsPythonOnly6998/config-scope
```

Failing on **exactly** the `config-scope` row and nothing else — the kill is attributable to the row added for it, not to incidental breakage. Restored by `cp`, `git diff HEAD` and `git status --porcelain` empty afterwards.

### On round 3

Adding **CS-2** unasked, because `operation-scope` otherwise killed nothing, is the right call and the second time this round you have made it. A row that grades no mutant is the decoration we have been removing everywhere else, and noticing it about your *own* new row is harder than noticing it about someone else's.

The judgement call you flagged is also the right one to have flagged: keeping the fixture's subtype segment as `class` rather than production's `ref`, on the grounds that `lookupStructural` reads only indices 1, 3, 4 and 5 and never `parts[2]`, so `scope:config:ref:…` and `scope:config:class:…` take an identical path. That keeps scope the only varied axis, and you verified it against the function rather than assuming it. **Recording it in the test comment is what makes it safe** — the next reader would otherwise reasonably suspect the fixture of not matching production, and would either "fix" it or re-derive the argument.

Re-scoring M1 and R3 under the new fixture shape, to confirm the language axis did not regress when scope became a parameter, is the step that is easy to skip and would have been the expensive one to miss.

### Where this pin stands now

Two guards on adjacent lines, both graded, on their own axes:

- **language** (`:3098`) — nine rows, two positive controls, killing `HasPrefix(lang,"py")`, `HasPrefix(lang,"p")`, an explicit widening to `csharp`/`typescript`/`go`, guard inversion, guard deletion, and the case-fold narrowing.
- **scope** (`:3094`) — two rows with `lang: python` held fixed, killing a `co` prefix widening and an explicit widening to `operation`.

That is the shape I wanted: the widest tier in the structural lookup, previously restricted by nothing an assertion could see, is now restricted by something that fails when either conjunct gets broader. **Merging on green.**

PR #7011, opened 2026-09-08
fix(#6909): the link pass never seeded SCOPE.ChannelBinding — and the sidecar it writes is what grafel_dead_code reads

### LINKED ISSUE #6909: The two reachability seed maps still diverge after #6907 — `SCOPE.ChannelBinding` seeds in `mcp` but not `links`, and the sidecar preference defeats ADR-0025 in the normal case
The two reachability seed maps still diverge after #6907 — `SCOPE.ChannelBinding` is a seed in `mcp` and not in `links`, and because the handler prefers the sidecar, ADR-0025 is defeated in the normal case

Found by the independent review of PR #6907 (#6902). **Not caused by that PR** — #6907 fixed the `Route` divergence between these two maps; this is a *second, pre-existing* divergence in the same pair, which the fix did not touch and which the review found by diffing the maps rather than by grepping for the kind under repair.

## The defect

Two hand-maintained seed sets are documented as mirrors of each other:

- `internal/links/reachability.go:104` — `frameworkEntryKinds`, the link pass's BFS seed set. Its verdict is **persisted** to the `<group>-links-reachability.json` sidecar.
- `internal/mcp/dead_code.go:228` — `frameworkEntryKindsMCP`, whose own comment says it *"mirrors the link-pass framework seeds (`internal/links/reachability.go frameworkEntryKinds`)"*.

`frameworkEntryKindsMCP` contains `"SCOPE.ChannelBinding"`, added by #5782 with an explicit rationale from ADR-0025:

```go
// #5782 (ADR-0025): a ChannelBinding is a config-side messaging
// declaration with no callers by design — never report it as dead code.
"SCOPE.ChannelBinding": true,
```

`links`' `frameworkEntryKinds` does **not** contain it.

## Why the divergence is not symmetric — the sidecar decides

This is the part that makes it a live defect rather than a tidiness complaint.

`handleDeadCode` **prefers the on-disk sidecar** and only recomputes when `from != ""` or the sidecar fails to load — confirmed by reading the handler, and independently by a coordinator mutant: forcing `loadReachabilitySidecar` to never load is DEAD, killed by #6839's degraded-repo tests.

So in the normal case the MCP map is **not consulted at all**. The verdict comes from the sidecar, which was written by the link pass using the map that lacks `SCOPE.ChannelBinding`. The protection ADR-0025 asks for exists only on the fallback path — precisely the path that does not run when a group has had link passes.

**A ChannelBinding is therefore reported as dead code in the normal case, which is the exact outcome ADR-0025 forbids.**

## What is and is not established

- **Structurally real**: both maps read, the divergence confirmed, the sidecar preference confirmed twice (by reading and by mutant).
- **No volume evidence**: the measured local index contains **zero** `SCOPE.ChannelBinding` entities, so nobody has observed this happening. It is a latent defect, not a reported one.
- **Not established**: whether any other member diverges between the two maps in either direction. Only `Route` (fixed by #6907) and `SCOPE.ChannelBinding` were compared member-by-member — **do not assume the rest match**; the whole point of this issue is that a comment claiming "mirrors" is not a guarantee.
- **Not established**: whether ADR-0025's intent extends to the persisted path at all, or whether the author of #5782 knowingly scoped it to the recompute path. **Read #5782 and ADR-0025 before choosing a fix** — if the seed belongs in `links` too, the fix is one line plus a sidecar re-run; if it was deliberately MCP-only, the comment is wrong and should say so.

## The shape this belongs to

#6846 found **eleven** hand-maintained copies of one exclusion list, two of them live bugs. #6894 and #6902 were the same defect one site apart, and **each of their enumerations was short by one**. This is now the third consecutive instance of "a hand-copied kind list diverged and nothing asserted equality".

The review of #6907 also notes a **third** hand-copy in the same family: `reachabilityEdgeKindsMCP` mirrors `reachabilityEdgeKinds` with **nothing asserting they are equal**.

**The durable fix is an assertion, not another sync.** #6907's author recommends sharing the predicates via `internal/types` (with accepting **and** forbidden rows) as a follow-up, noting the seed maps are the harder half: site D exists specifically to avoid an `mcp → links` dependency, so sharing them means moving a link-pass-owned concept into a leaf package. That is a design call worth making deliberately — but even without sharing, **a test asserting the two maps are equal would have caught both this and the `Route` bug**, and is cheap.

Refs #6907, #6902, #6894, #6846, #5782, #6839.

cajasmota:
Fixes #6909.

`grafel_dead_code` could report a `SCOPE.ChannelBinding` as dead code — the exact
outcome ADR-0025 forbids — because the protection was installed in the wrong one of
two hand-copied seed sets.

## The defect

Two sets decided the same question in two packages:

- `internal/links/reachability.go` `frameworkEntryKinds` — the link pass's BFS seed
  set. Its verdict is **persisted** to `<group>-links-reachability.json`.
- `internal/mcp/dead_code.go` `frameworkEntryKindsMCP` — a hand-copy whose comment
  claimed it "mirrors the link-pass framework seeds".

#5782 added `"SCOPE.ChannelBinding": true` to the **MCP** copy, citing ADR-0025 §3
(a ChannelBinding is a config-side messaging declaration with no callers by design).
`links` never got it. And `handleDeadCode` **prefers the on-disk sidecar**, consulting
its own map only when `from` is set or the sidecar fails to load — so in the normal
case, once a group has had link passes, the MCP map is not consulted at all and the
verdict comes from the sidecar written without the seed.

This is the **second** divergence in this same pair; #6907 fixed `Route`.

**Latency, honestly stated:** this is **latent**, not observed. The measured local
index contains **zero** `SCOPE.ChannelBinding` entities, so nobody has hit it. A zero
measurement is corpus-relative and does not mean it cannot occur — but it equally does
not license a claim of incidence, and none is made here.

## The fix — one set, no copy

1. `SCOPE.ChannelBinding` is added to `internal/links`' `frameworkEntryKinds`, so the
   persisted sidecar honours ADR-0025.
2. **`frameworkEntryKindsMCP` and `reachabilityEdgeKindsMCP` are deleted.** They existed
   "to avoid an mcp → links dependency" — a dependency that already exists two functions
   above them (`reachabilitySidecarPath` calls `links.PassSidecarPath`). `internal/mcp`
   now calls `links.IsFrameworkEntryKind` / `links.IsReachabilityEdgeKind`. Eliminating
   the duplication is preferred over asserting equality between two things that can
   drift together.

`internal/coverage` keeps its own edge-kind set: it answers a different question
(test reachability, `coverage.PropTestReachable`), and folding it in here would be a
behaviour change out of this issue's scope. It is called out in the new comment.

## Grading — both directions, on the defect's path

`internal/mcp/channelbinding_sidecar_seed_6909_test.go`:

- `TestDeadCode_ChannelBindingIsSeededOnTheSidecarPath_6909` writes a one-repo fixture
  graph **plus real source files** (#6839 degrades a repo whose sources it cannot read,
  which would make every assertion vacuous), runs the real `links.RunAllPasses`, and then
  calls `handleDeadCode`. A test that only builds an in-memory graph exercises the
  **fallback**, which already had the seed and passes with the defect fully present —
  that is the shape of #6902's existing test. The `source == "sidecar"` assertion is the
  guard that keeps this one on the defect's path.
  - recall row: `mp.messaging.outgoing.orders-6909` (`SCOPE.ChannelBinding`) must **not**
    be reported.
  - forbidden row: `reconcileLedger6909` — unexported, uncalled, no inbound edge, not a
    sniffable Go entry-point — must **still** be reported. Adding a seed is a widening,
    and a widening that silences genuinely dead code is a regression.
- `TestReachabilitySeedSetsAreNotDuplicated_6909` pins the durable half by
  **enumerating the whole seed set** — `links.FrameworkEntryKinds()` (sorted) asserted
  against a **hand-written literal of all ten kinds**, deliberately not derived from the
  set under test, since a want list computed from the thing it checks agrees with any
  deletion. Sampling loops are kept alongside for diagnosis (`SCOPE.Class` /
  `SCOPE.Function` / `SCOPE.Variable` must never be seeds; `CALLS`/`CONTAINS` propagate
  reachability and `SAME_AS` does not).

**Why a NEW test rather than an extended one.** Under the seed-removal mutant, the
pre-existing `TestDeadCode_BareRouteIsAFrameworkEntryPoint_6902` — the fallback-shaped
test for the *first* divergence in this same pair — stayed **GREEN**. That is direct
evidence this defect could not have been caught by extending the existing coverage.
Symmetrically, a mutant in `internal/links` leaves `./internal/links/` green: the only
thing grading that package's seed set lives in `internal/mcp`, which is where the
consumer is. Both facts are recorded in the code comments.

## Mutants (each applied by exact-match count, restored by `cp` from a shasum-verified
backup, scored with `-count=1` off a green baseline)

| Mutant | Verdict |
|---|---|
| Remove `"SCOPE.ChannelBinding"` from `links.frameworkEntryKinds` | **DEAD** — and it reproduces the issue verbatim: `dead_code = [mp.messaging.outgoing.orders-6909 reconcileLedger6909]` |
| Force `handleDeadCode` to never prefer the sidecar (`ok = false`) | **DEAD** — `source = "live_recompute"`, i.e. the vacuity check fires rather than the test quietly passing on the fallback |
| Add an unrelated kind (`"SCOPE.Function"`) to the seed set — the **more permissive** direction | **DEAD** twice: the forbidden row goes empty (`dead_code = []`) and the seed-set pin rejects the kind |
| Add an empty-string key `""` to the seed set | **DEAD** — `FrameworkEntryKinds()` drift |
| **Delete a pre-existing seed**, `"SCOPE.GrpcMethod"` | **DEAD** — `FrameworkEntryKinds()` drift. `./internal/links/` stays green under it, which is why the grading test lives in `internal/mcp` and says so |
| Add a *plausible* new seed `"SCOPE.Interface"` | **DEAD** — `FrameworkEntryKinds()` drift |

The last three were **ALIVE** against the first revision's sampled pin (5 of 10 kinds) and
are what the second commit fixes; one enumeration kills all three.

The third commit applies the same enumeration to the **other** exported set,
`reachabilityEdgeKinds` — the BFS **traversal filter** (`if !reachabilityEdgeKinds[e.Kind]`),
which the second revision exported and did not pin:

| Mutant | Verdict |
|---|---|
| **Delete `"IMPORTS"`** from the traversal filter | **DEAD** — `ReachabilityEdgeKinds()` drift. Was **ALIVE** against both packages before this commit |
| Add `"SAME_AS"` | **DEAD** — drift, plus the pre-existing `SAME_AS` row |
| Add `""` | **DEAD** — drift |

A member leaving that set is the **false-positive** direction and is worse for a user than
the false negative this PR fixes: an entity reachable only through, say, an import edge stops
being reachable and the tool names **live code** for deletion. Under-reporting is quietly
useless; over-reporting sends people to delete working code. That reasoning is recorded at the
definition, so the next reader does not have to re-derive it.

## Verification (macOS 15 / APFS, arm64; worktree at `scratchpad/impl-6909-n4x8p/wt`)

- `go vet ./internal/links/ ./internal/mcp/` — clean; `gofmt -l` — clean.
- `go test -count=1 ./internal/links/ ./internal/mcp/` — **ok** (13.3s / 140.9s), run with
  the working tree byte-identical to the committed state (shasums checked).
- `go test -count=1 ./cmd/grafel/` under a fresh `GRAFEL_HOME` + `GRAFEL_DAEMON_ROOT` — **ok**
  (147.2s), run on the exact committed tree.
- `scripts/quality/run.sh --ratchet` — exit 0, "38 gated fixtures held their baseline
  (29 at 100% must-have recall)"; `internal/quality/golden/baseline.json` unmodified.
- The follow-up commit adds one exported enumerator with no caller in the pass and edits the
  test file, so production behaviour is unchanged by it and `./cmd/grafel/` + the ratchet were
  not re-run for it; `./internal/links/` + `./internal/mcp/` re-run green (8.5s / 99.1s).

- The third commit is likewise one exported enumerator with no caller in the pass plus the test
  file, so `./cmd/grafel/` and the ratchet are not re-run for it either; `./internal/links/` +
  `./internal/mcp/` green (9.3s / 99.9s).

`internal/coverage` is left alone deliberately: its `reachabilityEdgeKinds` is `{TESTS, CALLS}`
built from `types` constants and answers test-reachability — a name collision, not a duplicate
of this set. Its line 208, `if !reachabilityEdgeKinds[rel.Kind]`, is syntactically identical to
the links one, and that resemblance alone produced an issue filed on a wrong premise (#7012,
closed invalid). A note at the links definition now says which question each set answers, so the
next reader does not make the same round trip.

Refs #6907, #6902, #6846, #5782, #6839, ADR-0025.

### DIFF
--- a/internal/links/reachability.go
+++ b/internal/links/reachability.go
@@ -110,7 +110,14 @@ var reachabilityEdgeKinds = map[string]bool{
 // an HTTP route — and #6776 arm B7 added both kinds together because both are
 // live. Seeding only the prefixed one persisted every Java/Spring/Django
 // route into the sidecar as unreachable, along with everything only it reaches.
-// internal/mcp/dead_code.go's frameworkEntryKindsMCP mirrors this map.
+// #6909: internal/mcp/dead_code.go used to hand-copy this map (and
+// reachabilityEdgeKinds) into frameworkEntryKindsMCP. The two copies diverged
+// twice — on "Route" (#6902) and on "SCOPE.ChannelBinding" — and the second
+// divergence was live, because handleDeadCode PREFERS the sidecar this pass
+// writes and only consults its own map when the sidecar fails to load. There
+// is now no copy: mcp consumes IsFrameworkEntryKind / IsReachabilityEdgeKind
+// below. (internal/coverage has its own edge-kind set for a different
+// question — test reachability — and is deliberately not folded in here.)
 var frameworkEntryKinds = map[string]bool{
 	"http_endpoint_definition": true,
 	"http_endpoint":            true,
@@ -118,10 +125,27 @@ var frameworkEntryKinds = map[string]bool{
 	"SCOPE.Route":              true,
 	"Route":                    true, // #6902
 	"SCOPE.MessageTopic":       true,
+	// #5782 (ADR-0025), added here by #6909: a ChannelBinding is a
+	// config-side messaging declaration with no callers by design — never
+	// report it as dead code. ADR-0025 §3 named only the MCP map because
+	// that was the only seed set at the time; the sidecar this pass writes
+	// is what the tool actually reads, so the seed has to be here too.
+	"SCOPE.ChannelBinding":     true,
 	"SCOPE.GrpcMethod":         true,
 	"SCOPE.ServerlessFunction": true,
 	"SCOPE.EventBusEvent":      true,
 }
+
+// IsFrameworkEntryKind reports whether an entity of this kind is a
+// framework-managed entry-point, i.e. a BFS seed that needs no inbound edge.
+//
+// Exported for internal/mcp's grafel_dead_code fallback, which used to keep a
+// hand-copied mirror of frameworkEntryKinds. #6909: one set, one consumer path.
+func IsFrameworkEntryKind(kind string) bool { return frameworkEntryKinds[kind] }
+
+// IsReachabilityEdgeKind reports whether an edge of this kind propagates
+// reachability. Exported for the same reason as IsFrameworkEntryKind (#6909).
+func IsReachabilityEdgeKind(kind string) bool { return reachabilityEdgeKinds[kind] }
 
 // reachabilityEntry is one persistent reachability fact for the sidecar.
 type reachabilityEntry struct {
--- a/internal/links/reachability.go
+++ b/internal/links/reachability.go
@@ -147,6 +147,33 @@ func IsFrameworkEntryKind(kind string) bool { return frameworkEntryKinds[kind] }
 // reachability. Exported for the same reason as IsFrameworkEntryKind (#6909).
 func IsReachabilityEdgeKind(kind string) bool { return reachabilityEdgeKinds[kind] }
 
+// FrameworkEntryKinds returns every seed kind, sorted.
+//
+// The predicate above cannot be tested by sampling: a positive/negative list of
+// SOME members grades only those members, so deleting an unlisted seed — say
+// "SCOPE.GrpcMethod", which would report every gRPC method as dead code — is
+// silent. This exists so a test can assert the WHOLE set against an independent
+// hand-written literal, which is the only shape that catches a member being
+// added, removed or renamed. Review of #6909 scored three such mutants ALIVE
+// against a sampled pin.
+//
+// The grading test lives in internal/mcp (channelbinding_sidecar_seed_6909_test.go)
+// because that is where the consumer is: an edit to this map that breaks the
+// dead-code tool leaves ./internal/links/ green. Do not delete it as "not about
+// links".
+//
+// Seeding a kind makes its whole TRANSITIVE CLOSURE reachable, not just the
+// entity itself — that is the intended semantics (a route's handlers are live
+// because the route is), and it is the blast radius of adding a member here.
+func FrameworkEntryKinds() []string {
+	out := make([]string, 0, len(frameworkEntryKinds))
+	for k := range frameworkEntryKinds {
+		out = append(out, k)
+	}
+	sort.Strings(out)
+	return out
+}
+
 // reachabilityEntry is one persistent reachability fact for the sidecar.
 type reachabilityEntry struct {
 	Repo         string   `json:"repo"`
--- a/internal/links/reachability.go
+++ b/internal/links/reachability.go
@@ -147,6 +147,34 @@ func IsFrameworkEntryKind(kind string) bool { return frameworkEntryKinds[kind] }
 // reachability. Exported for the same reason as IsFrameworkEntryKind (#6909).
 func IsReachabilityEdgeKind(kind string) bool { return reachabilityEdgeKinds[kind] }
 
+// ReachabilityEdgeKinds returns every reachability-propagating edge kind,
+// sorted, for the same reason FrameworkEntryKinds exists: this set cannot be
+// graded by sampling.
+//
+// It is the BFS TRAVERSAL FILTER (see the `if !reachabilityEdgeKinds[e.Kind]`
+// below), so a member leaving it is worse for a user than a missing seed:
+// an entity reachable only through, say, an IMPORTS edge stops being reachable
+// and the dead-code tool reports LIVE code as dead. Under-reporting is a
+// quietly useless tool; over-reporting sends people to delete working code.
+// Deleting "IMPORTS" was scored ALIVE against every test in this repo before
+// the enumeration in internal/mcp existed.
+//
+// NOT to be confused with internal/coverage's identically-named
+// reachabilityEdgeKinds, whose line 208 is syntactically identical to the one
+// below. That set is {TESTS, CALLS} built from types.RelationshipKind*
+// constants and answers "which tests reach this entity" (coverage.PropTestReachable),
+// not "is this dead". It is a name collision, not a copy of this set, and must
+// not be folded in here — the resemblance has already produced one issue filed
+// on a wrong premise (#7012, closed invalid).
+func ReachabilityEdgeKinds() []string {
+	out := make([]string, 0, len(reachabilityEdgeKinds))
+	for k := range reachabilityEdgeKinds {
+		out = append(out, k)
+	}
+	sort.Strings(out)
+	return out
+}
+
 // FrameworkEntryKinds returns every seed kind, sorted.
 //
 // The predicate above cannot be tested by sampling: a positive/negative list of
--- a/internal/mcp/channelbinding_sidecar_seed_6909_test.go
+++ b/internal/mcp/channelbinding_sidecar_seed_6909_test.go
@@ -260,6 +260,32 @@ func TestReachabilitySeedSetsAreNotDuplicated_6909(t *testing.T) {
 				"and silences the dead-code tool.", kind)
 		}
 	}
+	// The SECOND exported predicate, enumerated the same way and for a sharper
+	// reason: this set is the BFS TRAVERSAL FILTER, so a member leaving it makes
+	// entities reachable only through that edge kind report as DEAD — the tool
+	// names live code for deletion. Review of this PR scored deleting "IMPORTS"
+	// ALIVE against every test in the repo. Hand-written, not derived from the
+	// map under test, for the same reason as wantSeeds above.
+	wantEdgeKinds := []string{
+		"CALLS", "CONSUMES", "CONTAINS", "DEPENDS_ON", "DISCRIMINATES_ON",
+		"ENTRY_POINT_OF", "EXTENDS", "FETCHES", "HANDLES", "HANDLES_SIGNAL",
+		"IMPLEMENTS", "IMPORTS", "NAVIGATES_TO", "PRODUCES", "REFERENCES",
+		"REGISTERS", "RENDERS", "RESOLVES_TO", "ROUTES_TO", "STEP_IN_PROCESS",
+		"TESTS", "UNRESOLVED_FETCH", "USES", "USES_HOOK",
+	}
+	gotEdgeKinds := links.ReachabilityEdgeKinds()
+	if !slices.Equal(gotEdgeKinds, wantEdgeKinds) {
+		t.Errorf("links.ReachabilityEdgeKinds() has drifted from the set this repo "+
+			"intends to propagate reachability along.\n got:  %q\n want: %q\n"+
+			"A kind REMOVED here makes every entity reachable only via that edge "+
+			"report as dead code — the false-positive direction, which sends users "+
+			"to delete working code. A kind ADDED lights up entities nothing really "+
+			"reaches. If the change is intended, edit wantEdgeKinds in the same "+
+			"commit. (internal/coverage has an identically-named set answering test "+
+			"reachability — it is a name collision, not this set.)",
+			gotEdgeKinds, wantEdgeKinds)
+	}
+
 	if !links.IsReachabilityEdgeKind("CALLS") || !links.IsReachabilityEdgeKind("CONTAINS") {
 		t.Errorf("links.IsReachabilityEdgeKind must accept CALLS and CONTAINS; "+
 			"got CALLS=%v CONTAINS=%v",
--- a/internal/mcp/channelbinding_sidecar_seed_6909_test.go
+++ b/internal/mcp/channelbinding_sidecar_seed_6909_test.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"os"
 	"path/filepath"
+	"slices"
 	"testing"
 
 	"github.com/cajasmota/grafel/internal/graph"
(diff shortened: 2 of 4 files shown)


cajasmota:
## Independent adversarial review — REQUEST-CHANGES (one small, priced fix)

The production change is right and the test is genuinely on the defect's path — I reproduced both of
those independently. What I am asking for is 8 lines: the new pin protects **5 of the 10 seeds** and
nothing else, so removing a pre-existing seed is silent.

All numbers below: macOS 15.6 / APFS / arm64, Go toolchain in-repo, detached worktree at
`.../scratchpad/review-7011-qz7k/wt` @ `f6ecfc7f7`. Tree byte-identical to the commit before and after
every score (`git status --porcelain` empty; `reachability.go` `b544f9ae…`, `dead_code.go` `62bcf016…`).
Restores by `cp` from a shasum-verified backup, never `git checkout`.

### Baseline (green, immediately before each score)

| Gate | Result |
|---|---|
| `gofmt -l internal/links internal/mcp` | clean |
| `go vet ./internal/links/ ./internal/mcp/` | exit 0 |
| `go test -count=1 ./internal/links/ ./internal/mcp/` | **ok** 10.6s / 100.7s |
| `go test -count=1 ./cmd/grafel/` (fresh `GRAFEL_HOME` + `GRAFEL_DAEMON_ROOT`) | **ok** 146.3s |
| `scripts/quality/run.sh --ratchet` | exit 0 — "38 gated fixtures held their baseline (29 at 100% must-have recall)" |
| `internal/quality/golden/baseline.json` modified? | **no** — verified from `gh pr diff 7011 --name-only` (4 files, none is the baseline), not from the "38 held" line |

### Mutants I scored (mine, not the author's — each applied by exact occurrence count, hit set asserted)

| # | Mutant | Verdict |
|---|---|---|
| M1 | `IsFrameworkEntryKind` → `return true` (1 occ) | **DEAD** ×3 — `TestReachabilitySeedSetsAreNotDuplicated_6909`, `TestDeadCode_BareRouteIsAFrameworkEntryPoint_6902`, `TestDeadCodeKindFilter` |
| M2 | add `"": true` to `links.frameworkEntryKinds` — empty kind seeds on **both** the sidecar and fallback paths | **ALIVE** — distinguishable, ungraded (production-reachability unproven: I found no `Kind == ""` guard in `internal/graph`/`links`/`mcp`, but no extractor proof either) |
| M3 | **delete a pre-existing seed** — `"SCOPE.GrpcMethod"` (1 occ, `after=0` asserted) | **ALIVE** — `links` + `mcp` + `coverage` all green. This is the one that matters. |
| M4 | force `handleDeadCode` never to prefer the sidecar (`ok = false`) | **DEAD** — `source = "live_recompute"`, path guard fires with the intended message. Author's claim reproduced verbatim. |
| M5 | remove `"SCOPE.ChannelBinding"` from `links.frameworkEntryKinds` | **DEAD** — `dead_code = [mp.messaging.outgoing.orders-6909 reconcileLedger6909]`. **And `TestDeadCode_BareRouteIsAFrameworkEntryPoint_6902` stayed GREEN under it** — direct proof the pre-existing fallback-shaped test could not have caught #6909. |
| M6 | add a *plausible* new seed `"SCOPE.Interface"` — the permissive direction a future maintainer might genuinely take | **ALIVE** — the negative list names only `SCOPE.Class` / `SCOPE.Function` / `SCOPE.Variable` |

Note on M5: the mutant is **in `internal/links`** and `./internal/links/` stayed green — the only thing
grading that package's seed set lives in `internal/mcp`. Worth a line in the comment so the next person
doesn't delete the mcp test as "not about links".

### The four verdicts asked for

**(a) Is the `mcp → links` dependency claim true? YES, verified at the parent commit.**
`git show f6ecfc7f7^:internal/mcp/dead_code.go` — `reachabilitySidecarPath` already reads
`links.PassSidecarPath("", group, "reachability")`, and the file already imported `links`. The comment
"kept in this file … to avoid an mcp → links dependency" was describing a dependency that had existed
since #6178 round 3. Deleting the copies adds nothing. **This is the strongest part of the PR.**

I also checked the replacements are behaviourally identical rather than sampling them: both predicates
are literal one-line map reads of the same maps, and the key sets match exactly —
`reachabilityEdgeKindsMCP` vs `links.reachabilityEdgeKinds` **24 = 24, zero diff**; the framework sets
differ pre-fix by exactly one key (`SCOPE.ChannelBinding`) and post-fix by none. Absent key, empty
string and case behave the same because it is the same map lookup.

**(b) Is the test genuinely on the sidecar path? YES.** M4 confirms the `source == "sidecar"` guard is
load-bearing, and M5 confirms the fallback-shaped #6902 test does not fire. It drives real
`links.RunAllPasses` over real on-disk sources, and #6839 degradation is covered by construction: a
degraded repo yields no verdicts, which fatals assertion (1) rather than making (2) vacuous.

**(c) Is removing a pre-existing seed caught? NO — M3 is ALIVE.** The positive literal list pins
`SCOPE.ChannelBinding`, `Route`, `SCOPE.Route`, `SCOPE.MessageTopic`, `http_endpoint_definition`.
Unpinned: `http_endpoint`, `SCOPE.Endpoint`, `SCOPE.GrpcMethod`, `SCOPE.ServerlessFunction`,
`SCOPE.EventBusEvent`. The PR body says the test "pins the durable half"; it pins half of the half. A
future edit that drops `SCOPE.GrpcMethod` reports every gRPC method as dead code and no test says so.

**(d) Can the widening mask real dead code? Not this widening; yes in general.** The `SCOPE.ChannelBinding`
addition is graded in both directions on the defect's path — `reconcileLedger6909` is still reported
(and I saw it in M5's own output alongside the ChannelBinding, so the forbidden row is not vacuous).
But M6 shows a *different* plausible widening is invisible. Also worth stating explicitly in the code
comment: seeding a kind makes its whole **transitive closure** reachable, not just the entity — that is
correct semantics here (same as `SCOPE.MessageTopic`), but it is the actual blast radius and the test's
forbidden row has no inbound edge from the ChannelBinding, so the transitive direction is untested.

### The fix I am asking for (~8 lines, kills M2, M3 and M6 together)

Sampling cannot detect a predicate that got broader or narrower — enumerate instead. Export the set,
not just the predicate, and assert it exactly:

```go
// links
func FrameworkEntryKinds() []string { /* sorted keys */ }
```
then in `TestReachabilitySeedSetsAreNotDuplicated_6909` compare against an independent sorted literal of
all ten kinds. That single assertion is DEAD against M2 (extra `""` key), M3 (missing `SCOPE.GrpcMethod`)
and M6 (extra `SCOPE.Interface`), and an intended future change just edits the literal. If you'd rather
not export a second symbol, adding the five missing kinds to the existing positive loop kills M3 for
~5 lines and leaves M2/M6 open — I'd take the enumeration.

### Scope (Claim 4) — leaving `internal/coverage` alone is CORRECT, and it is not a third copy

`internal/coverage`'s `reachabilityEdgeKinds` is `{TESTS, CALLS}` — **2 kinds, built from
`types.RelationshipKind*` constants**, with zero overlap in intent and no framework-entry seed set at
all. Diff against `links`' 24-kind set: 24 lines added, 0 shared structure. It answers "which tests
reach this entity" (`PropTestReachable`), not "is this dead". It is a name collision, not a divergence.
Folding it in would be a behaviour change. Filing it separately is fine but I'd expect that issue to
close as "different question".

### Honesty check on the PR body — PASSES

The body states the defect is **latent**, that the measured local index contains **zero**
`SCOPE.ChannelBinding` entities, and explicitly that a zero measurement is corpus-relative and does not
license a claim of incidence. No incidence is claimed anywhere. The `dashboard_tools.go` comment update
is accurate too.

Re-run `./internal/links/ ./internal/mcp/` after the enumeration and I'll re-score M2/M3/M6 — everything
else here is already verified and I have no other objection.

cajasmota:
## Coordinator mutant on `d86adb810` — **CU-1 is ALIVE**, and it is the sibling of the hole you just closed

Own detached worktree at `/private/tmp/.../coord-mut-7011-t3n/wt` (absolute). Green baseline immediately before (`./internal/links/` 10.6s, `./internal/mcp/` 84.5s, both exit 0). Restored by `cp` from a SHA-256-verified backup (`26c6105b…` both ways); `git diff HEAD` and `git status --porcelain` empty afterwards.

The round-2 fix is right and does what it says: the ten-kind hand-written literal kills M2, M3 and M6 together, and dropping `SCOPE.GrpcMethod` is now caught. The doc comment recording *why* the grading test lives in `internal/mcp` — that a seed-set mutant leaves `./internal/links/` green, so a future reader must not relocate or delete it as "not about links" — is the kind of note that stops a correct test being removed for a plausible-sounding reason.

**But this PR exports two predicates and enumerates one.**

### CU-1 — delete one kind from `reachabilityEdgeKinds`

Removed `"IMPORTS"` from the 24-kind edge set at `reachability.go:76`. Hit set asserted **`[76]`**, exactly one occurrence, zero `"IMPORTS"` lines remaining afterwards, `go vet` clean.

**Result: ALIVE.** `./internal/links/` and `./internal/mcp/` both exit 0.

### Why it matters

`reachabilityEdgeKinds` is the BFS traversal set — `reachability.go:207` is `if !reachabilityEdgeKinds[e.Kind] { … }`, so it decides which edges propagate reachability at all. Removing `IMPORTS` means an entity reachable **only** through an import edge stops being reachable, and a dead-code tool then reports live code as dead. That is the **false-positive** direction, which is worse for a user than the false-negative this PR fixes: an under-reporting tool is quietly useless, but an over-reporting one sends people to delete working code.

The new literal covers `FrameworkEntryKinds()` — the ten **seeds**. The 24 **edge kinds** have no equivalent, so `IsReachabilityEdgeKind` is in exactly the state `IsFrameworkEntryKind` was in before this round: exported, load-bearing, and pinned by nothing that would notice a member leaving.

This is the same shape as the review's own finding, one predicate over — and the same shape as #6998 (language enumerated, scope left open) and #6914 (arity enumerated, layout left open). **Enumerating one set thoroughly is what makes its sibling look covered.**

### Asked change — the same ~8 lines, applied to the other set

Export the edge kinds the way you exported the seeds, and assert them against a second hand-written literal of all 24. Reuse the machinery; do not build a second mechanism. Then score three mutants on it — delete a member, add an implausible member, add the empty string — and show each DEAD.

**Two constraints:**

1. **The literal must be hand-written**, not derived from the map under test — the same reason your seed literal is. A derived want-list agrees with any deletion.
2. **Do not extend this to `internal/coverage`.** Its `reachabilityEdgeKinds` is `{TESTS, CALLS}` from `types` constants answering *test* reachability — a name collision, not a copy. I filed that as #7012 on a wrong premise and have closed it as invalid; your original scope call was correct and I was wrong to second-guess it.

### One thing worth stating in the PR body

`internal/coverage/reachability.go:208` contains a **syntactically identical line** to the one in links (`if !reachabilityEdgeKinds[rel.Kind]`). That resemblance is what led me to file #7012 wrongly. A one-line note in the links comment saying the coverage set is a different question would have saved that round trip, and will save the next reader the same one.

cajasmota:
## CU-1 re-scored on `86a19a0df` by me — **DEAD**

I found this one, so I confirmed the kill myself rather than taking the re-score on report. Same worktree, checked out at `86a19a0df`, green baseline immediately before. `"IMPORTS"` map-key occurrence asserted **1** before the edit and **0** after; `go vet` clean.

**Result: DEAD** — `TestReachabilitySeedSetsAreNotDuplicated_6909`. Restored, `git diff HEAD` and `git status --porcelain` both empty.

Both exported predicates are now enumerated against **hand-written** literals — ten seeds and twenty-four edge kinds — so a member leaving either set fails rather than passing quietly.

### The doc comment is the part I would keep

> *it is the traversal filter, so a member leaving makes entities reachable only via that edge report as dead — the false-positive direction, which sends users to delete working code*

That records the asymmetry rather than the mechanism. The two sets look interchangeable and are not: losing a **seed** under-reports dead code, which is quietly useless; losing an **edge kind** over-reports it, which is actively harmful. A future reader deciding how much care each deserves now has the answer in front of them instead of having to derive it.

Recording that deleting `"IMPORTS"` was ALIVE *before* the pin existed is the other half — it dates the coverage, so nobody re-derives whether the pin was ever load-bearing.

### On the coverage note

Adding it at the links definition is right, and it earns its place: `internal/coverage/reachability.go:208` is **syntactically identical** to the links line, and that resemblance is exactly what led me to file #7012 on a wrong premise. The note is there to stop the next person spending the round trip I spent.

### Where this PR ended up

It started as "add one kind to a map". It ends as: the duplicate maps deleted rather than asserted (after checking that the stated reason for their existence was false), the defect pinned **on the sidecar path** where it actually lives — with the pre-existing #6902 fallback test demonstrated to stay green under the seed-removal mutant, which is the evidence a new test was needed rather than an extended one — and both exported sets enumerated.

Three rounds, each one finding the next unenumerated thing: five of ten seeds pinned → all ten → the other set entirely unpinned. Same shape three times, which is worth naming: **enumerating one set thoroughly is what makes its sibling look covered.**

**Merging on green.**

PR #7015, opened 2026-09-08
fix(#6916): the aws_cdk `Config CALLS Component` edge could never bind — the entity was named `new cdk.App(`, the edge sources on the variable

cajasmota:
## Tier B, aws_cdk arm — 2 of the 39 `name_group: 0` sites. The other 37 are untouched.

This is the proving arm for #6916: it is the one site pair where the narrowing does not merely
improve a name, it makes an **already-shipped, already-emitted edge start binding**.

Measured in a detached worktree at `3199981ad`,
`.../scratchpad/impl-6916b-q7k2v/wt`, **macOS/APFS, Darwin 25.6.0** (#6966 gate — every figure
below is that platform only).

### The defect

Both aws_cdk `Config` source_patterns used `name_group: 0`, so the Name is the whole match:

| file | shipped pattern | shipped Name |
|---|---|---|
| `javascript_typescript/frameworks/aws_cdk.yaml:50` | `new\s+(?:cdk\.)?App\s*\(` | `new cdk.App(` |
| `python/frameworks/aws_cdk.yaml:47` | `(?:cdk\.)?App\s*\(\s*\)` | `cdk.App()` |

The `Config CALLS Component` relationship rule **in the same file** sources on `source_group: 2` of
`new (\w+Stack)\s*\(\s*(\w+)` — the **variable**, `app`. The entity is named `new cdk.App(`. Those
two strings can never be equal, so the edge dangles at its source end by construction. That is the
#6906 `bug-extractor` shape, and it has already shipped.

### The change

Both patterns now capture the **assignment target**, following `express.yaml:91`. The two languages
are **not** the same regex — Python has no declarator keyword:

```yaml
# javascript_typescript
- pattern: "(\\w+)(?:\\s*:\\s*[\\w.]+)?\\s*=\\s*new\\s+(?:cdk\\.)?App\\s*\\("
  name_group: 1
# python
- pattern: "(?m)^[ \\t]*(?:\\w+\\.)?(\\w+)\\s*=\\s*(?:cdk\\.)?App\\s*\\(\\s*\\)"
  name_group: 1
```

**Both patterns changed after review** — see the round-2 comment below for the blocker (Python needed
a left anchor or it bound the CALLS edge to the *wrong* variable), the type-annotation gap, and the
over-firing fixture.

**Correction to an earlier claim in this body:** it said the JS `(?:(?:const|let|var)\s+)?`
declarator group was "found by a failing test". That was wrong. Review mutant RM3 measured the group
**inert** — `(\w+)` alone already backtracks to the assignment target on `const`/`let`/`var`,
multi-declarator and attribute lines alike — and it has been **deleted** rather than described. What a
failing test did find was that the *first* JS spelling, which required the declarator, could not
reach a multi-declarator line.

**Exact edit landing**: the replacement was applied with a scripted `assert count == 1` on each
file, and `grep -c 'name_group: 0'` over both files is now `0` / `0`. Two files carry near-identical
rules; the hit set is exactly those two lines, `git diff --stat` = 2 files, 10 insertions, 6
deletions.

### 1. The edge binds — named, both endpoints

Driven through the real `LoadAllRules()` + `Detector.Detect()`. The detector builds a relationship
endpoint ID as `"<type>:<captured name>"` (`detector.go:562`); the source end binds exactly when an
emitted entity has that Kind and Name.

Fixture (`bin/app.ts`, and its Python twin) declaring both endpoints in one file:

```
Config:app --CALLS--> Component:ApiStack
```

- **Before**: entities were `Config:"new cdk.App("` + `Component:ApiStack`. The edge's source
  endpoint `Config:app` **named no emitted entity**.
- **After**: entities are `Config:app` + `Component:ApiStack`. **Both endpoints of that edge name an
  entity emitted by the same run.**

`TestIssue6916_CDKCallsEdgeBindsAtItsSource` asserts the edge string and then each endpoint against
the emitted entity set — not a count. (#6973: `TESTS` moved +8 net while 7,221 edges were
substituted underneath.)

### 2. What it costs: the unassigned construction

A bare `new cdk.App()` / `cdk.App()` with no assignment now matches nothing and mints no entity.
That is pinned deliberately by `TestIssue6916_UnassignedAppMintsNoConfig`, each leg carrying
`Component:ApiStack` as a positive control so the absence cannot pass vacuously.

**How many corpus occurrences is that? Zero — because the corpus contains zero CDK at all.**
Measured over `/Users/jorgecajas/Projects/archigraph-corpora` (60 repos, `node_modules`/`dist`/
`build`/`.next`/`venv` excluded): **0** matches of `new\s+(cdk\.)?App\s*\(` in `.ts/.js/.tsx/.jsx`
and **0** matches of `(cdk\.)?App\s*\(\s*\)` in `.py`. That is consistent with the #6916 measurement
comment, which attributed **0 entities** to both aws_cdk sites.

**So this bullet is honestly under-measured and I am flagging it rather than claiming it.** A
corpus zero is corpus-relative, never a capability claim. What I can put on the record:

- The **only** CDK App construction anywhere in this repo's trees,
  `testdata/fixtures/real-world/cdk/api_lambda_dynamodb_stack.ts:130`, is `const app = new App();` —
  **assigned**.
- Structurally, an unassigned App is **unusable**: every stack takes the app as its scope argument
  (`new MyStack(app, 'Id')`), and the relationship rule's own `source_group: 2` is `(\w+)` — a
  variable. An App that is never bound to a name can never appear in that edge, so the node it used
  to mint was a per-file boolean marker with nothing to attach to.

If you want this measured against a real CDK population before Tier A copies the shape, that is a
fair hold and I will take it.

### 3. Migration: **no `fbversion.Version` bump.** Deliberate, and here is why.

`internal/types/kinds.go:26` already settles this class of question explicitly: *"A rename must
therefore NOT be expressed as an fbversion bump. That would reindex every registered group behind
the user's back."* `fbversion.Version` (currently `6`) versions the FlatBuffers **payload format**;
a bump makes `internal/graph`'s min-version gate reject the stored graph, which `internal/daemon`
turns into an automatic full reindex of **every** registered group.

This change alters no format, no schema, and no entity **kind** — it changes one entity **Name**, in
one framework, in two languages. `KindVocabularyVersion` is also not in scope: it is bumped when a
kind is renamed or retired, and `Config` is neither.

The consequence, stated rather than hidden: `ComputeID` hashes
`OrgID+ProjectID+SourceFile+Kind+Name` (`internal/types/entity.go:139`), so this changes the entity
ID. A **full re-index is clean** (whole-document FlatBuffers generation, no rows). The two
**incremental** paths carry no rule digest in the manifest, so a rule edit there yields a mixed-ID
graph plus dangling edges the merge never cleans — **#6085, pre-existing, not created by this PR and
not fixed by it**. Forcing a fleet-wide reindex to paper over it costs every user a full index for a
two-line rule change in one framework; the #6757-arm-C call was that the user gets to choose when to
pay that.

### Verification (all in the worktree above, `-count=1`, `go vet` first, `gofmt -l` clean)

- `go test ./internal/engine/` — **ok, 43.996s**
- `go test ./cmd/grafel/` — **ok, 148.029s**, in a **fresh** `GRAFEL_HOME` + `GRAFEL_DAEMON_ROOT`
  (+ `HOME`) sandbox under the scratchpad. The real `~/.grafel` was never touched.
- `scripts/quality/run.sh --ratchet` — **exit 0**, `quality ratchet OK — 38 gated fixtures held
  their baseline (29 at 100% must-have recall)`. **No baseline moved**, so no fixture-member
  enumeration is owed; no constant was bumped and `--update-baseline` was never run.

### Mutants — scored in the permissive direction, `-count=1`, green baseline immediately before each

| # | mutant | verdict |
|---|---|---|
| M1a | JS `name_group: 1` → `0` | **DEAD** — `Config:app` gone; edge reported dangling; `Config:"app = new cdk.App("` flagged non-identifier |
| M1b | Python `name_group: 1` → `0` | **DEAD** — same two assertions |
| M2a | JS variable capture `(\w+)` → `(.+)` | **DEAD** — minted `Config:"region = 'eu-west-1', app"` |
| M2b | Python variable capture `(\w+)` → `(.+)` | **DEAD** — minted `Config:"self.app"` |
| M4 | full revert of **both** sites to the shipped state | **DEAD** — both marker Names named in the failure |
| M5 | JS: assignment prefix made **optional**, `name_group: 1` | **ALIVE — equivalent, with a mechanism** |

**M2b was ALIVE on the first score and that is why the Python leg exists.** The single-assignment
fixture cannot grade `(.+)`: forced to stop at the `=`, it captures `app` too, so it is
indistinguishable there. The shape that separates them is **not shared between the languages** —
JS/TS needs a multi-declarator `const` line, Python has no declarator keyword and needs an attribute
target (`self.app = cdk.App()`). Two fixtures, one per language; grading one and assuming the other
is exactly the "fixtures pin one axis and leave the neighbour open" failure.

**M5 is equivalent under the detector, not merely under the suite.** With the assignment optional
and `name_group: 1`, a bare `new cdk.App()` matches but capture group 1 is empty; `detector.go` only
falls back to the full match when `nameGroup == 0`, so an empty name `continue`s and no entity is
emitted. Every input therefore produces the identical entity set. That is not an argument —
`TestIssue6916_UnassignedAppMintsNoConfig` passing under M5 **is** the demonstration that the
optional branch emits nothing.

### Known uncovered, deliberately

A re-add that renames the `Config` kind itself slips past these tests, exactly as noted on the #6914
pin. Closing it would mean asserting a whole-graph absence, which fights the Tier A work that will
legitimately add `Config` anchors.

Refs #6916, #6914, #6906, #6912, #6726, #6085.

### DIFF
--- a/internal/engine/aws_cdk_app_name_6916_test.go
+++ b/internal/engine/aws_cdk_app_name_6916_test.go
@@ -79,10 +79,25 @@ new ApiStack(app, 'ApiStack');
 
 // cdk6916PyAttrApp is the Python analogue and grades the SAME widening in the
 // Python rule. Python has no declarator keyword, so the multi-declarator line
-// above has no Python spelling; the shape that separates `(\w+)` from `(.+)`
-// here is an ATTRIBUTE target, `self.app = cdk.App()`, which is idiomatic when
-// the App is built inside a class. `(\w+)` captures "app"; `(.+)` captures
-// "self.app", which the CALLS rule's `(\w+)` source group can never equal.
+// above has no Python spelling; the shape that separates `(\w+)` from `(.+)` and
+// from `([\w.]+)` here is an ATTRIBUTE target, `self.app = cdk.App()`, which is
+// idiomatic when the App is built inside a class. `(\w+)` (with the rule's
+// `(?:\w+\.)?` prefix consuming `self.`) names it "app"; the widened captures
+// name it "self.app".
+//
+// CORRECTED (review ask 4): this fixture grades the NAME ONLY. It emits no
+// relationship at all — the Python CALLS rule is `(\w+Stack)\s*\(\s*(\w+)\s*,`
+// and `ApiStack(self.app, "ApiStack")` has no `,` directly after a bare `(\w+)`,
+// so NEITHER spelling produces an edge here. The earlier comment implied "app"
+// would bind and "self.app" would not; measured, neither does. The binding is
+// graded by TestIssue6916_CDKCallsEdgeBindsAtItsSource, not here.
+//
+// The second method carries a MULTI-LEVEL attribute chain, `self.inner.app`. The
+// rule's `(?:\w+\.)?` prefix is one level deep, so that line mints nothing —
+// correct, because a dotted name could never equal the CALLS rule's `(\w+)`
+// source anyway. It grades the dotted-capture mutant `(\w+)` -> `([\w.]+)`,
+// which the single-level `self.app` line cannot: there the `(?:\w+\.)?` prefix
+// already eats `self.` and both spellings name it "app".
 const cdk6916PyAttrApp = `import aws_cdk as cdk
 
 class ApiStack(cdk.Stack):
@@ -93,6 +108,79 @@ class Deployment:
     def build(self):
         self.app = cdk.App()
         ApiStack(self.app, "ApiStack")
+
+    def build_nested(self):
+        self.inner.app = cdk.App()
+`
+
+// cdk6916TSAnnotated carries a TYPE ANNOTATION, which is ordinary TypeScript and
+// which no fixture covered (review ask 2 / mutant RM4). Deleting the rule's
+// `(?:\s*:\s*[\w.]+)?` group makes `(\w+)` land inside the annotation and mint
+// `Config:App` — the TYPE name — so the edge dangles again.
+const cdk6916TSAnnotated = `import * as cdk from 'aws-cdk-lib';
+
+export class ApiStack extends cdk.Stack {
+}
+
+const app: cdk.App = new cdk.App();
+new ApiStack(app, 'ApiStack');
+`
+
+// cdk6916PyMultiTarget is the review BLOCKER made into a fixture. Python's
+// `App()` is a plain call, so the text of a multi-target assignment contains the
+// literal substring `env = cdk.App()`. An unanchored `(\w+)\s*=` captures `env`
+// — which is the string "prod" — and the CALLS edge then BINDS TO THE WRONG
+// VARIABLE. Before #6916 that edge merely dangled; a confidently wrong endpoint
+// is worse than a known-bad one.
+//
+// The rule's `(?m)^[ \t]*` anchor makes this mint nothing instead. That is the
+// deliberate cost and it is pinned as such: no Config entity, and no relationship
+// sourced on a Config.
+const cdk6916PyMultiTarget = `import aws_cdk as cdk
+
+class EnvStack(cdk.Stack):
+    pass
+
+app, env = cdk.App(), "prod"
+EnvStack(env, "EnvStack")
+`
+
+// cdk6916TSDestructured is the nearest JS/TS spelling of the same hazard. It is
+// here to SHOW the JS rule does not have the seam rather than to assume it: the
+// JS pattern requires `= new ... App(` immediately, so `(\w+)` can only ever be
+// the token directly left of that `=`, which in JS is the assignment target.
+// A destructuring form matches nothing at all.
+const cdk6916TSDestructured = `import * as cdk from 'aws-cdk-lib';
+
+export class EnvStack extends cdk.Stack {
+}
+
+const [app, env] = [new cdk.App(), 'prod'];
+new EnvStack(env, 'EnvStack');
+`
+
+// cdk6916TSPlainAppCall / cdk6916TSNewAppCall are the OVER-FIRING pair (review
+// ask 3 / coordinator mutant CV-1). They are the same file except for the word
+// `new`. `App()` as a plain call is an ordinary React component invocation in a
+// repo with no CDK anywhere; `new` is the only thing keeping this rule off it.
+//
+// The `new` leg is the POSITIVE CONTROL: it proves the rule can reach this file
+// at all, so the absence asserted on the other leg is not vacuous. Both are
+// deliberately free of any CDK marker.
+const cdk6916TSPlainAppCall = `import App from './App';
+
+function bootstrap() {
+  const app = App();
+  return app;
+}
+`
+
+const cdk6916TSNewAppCall = `import App from './App';
+
+function bootstrap() {
+  const app = new App();
+  return app;
+}
 `
 
 // cdk6916TSBareApp / cdk6916PyBareApp are the construction with NO assignment.
@@ -166,6 +254,7 @@ func TestIssue6916_CDKAppConfigIsNamedAfterVariable(t *testing.T) {
 		forbidden string
 	}{
 		{"typescript", "bin/app.ts", "typescript", cdk6916TSApp, "new cdk.App("},
+		{"typescript with type annotation", "bin/app.ts", "typescript", cdk6916TSAnnotated, "new cdk.App("},
 		{"python", "app.py", "python", cdk6916PyApp, "cdk.App()"},
 	} {
 		t.Run(tc.name, func(t *testing.T) {
@@ -190,6 +279,7 @@ func TestIssue6916_CDKAppConfigIsNamedAfterVariable(t *testing.T) {
 func TestIssue6916_CDKCallsEdgeBindsAtItsSource(t *testing.T) {
 	for _, tc := range []struct{ name, path, lang, src string }{
 		{"typescript", "bin/app.ts", "typescript", cdk6916TSApp},
+		{"typescript with type annotation", "bin/app.ts", "typescript", cdk6916TSAnnotated},
 		{"python", "app.py", "python", cdk6916PyApp},
 	} {
 		t.Run(tc.name, func(t *testing.T) {
@@ -284,5 +374,99 @@ func TestIssue6916_UnassignedAppMintsNoConfig(t *testing.T) {
 					strings.Join(ids, "\n  "))
 			}
 		})
+	}
+}
+
+// TestIssue6916_MultiTargetAssignmentDoesNotMisattribute is the review blocker,
+// pinned. The failure mode it guards is NOT "an edge goes missing" — it is "an
+// edge binds to the wrong variable", which the rest of this file cannot see
+// because the wrong name (`env`) is identifier-shaped and passes every other
+// assertion here.
+//
+// Both legs assert the same two things: no Config entity is minted from a
+// multi-target / destructuring assignment, and the `Config:env --CALLS-->` edge
+// the relationship rule emits from `EnvStack(env, …)` therefore does NOT BIND.
+//
+// That edge is emitted either way — the relationship rule is independent of the
+// entity rule and fires on the `EnvStack(env,` text alone. It dangled before
+// #6916 and it must go on dangling: the whole point of this issue is that a
+// dangling endpoint is a known-bad one, while an endpoint that resolves to the
+// WRONG variable is a wrong answer stated confidently. Asserting "the edge is
+// absent" would be false; asserting "the edge does not bind" is the real
+// guarantee. Component:EnvStack is the positive control on each leg.
+func TestIssue6916_MultiTargetAssignmentDoesNotMisattribute(t *testing.T) {
+	for _, tc := range []struct {
+		name, path, lang, src string
+		// wrong is the name the unanchored capture donated. Named so a future
+		// failure reads as this defect rather than as a general absence.
+		wrong string
+	}{
+		{"python tuple assignment", "app.py", "python", cdk6916PyMultiTarget, "env"},
+		{"typescript array destructuring", "bin/app.ts", "typescript", cdk6916TSDestructured, "env"},
+	} {
+		t.Run(tc.name, func(t *testing.T) {
+			res := detect6916(t, tc.path, tc.lang, tc.src)
+			ids := entityIDs6916(res)
+			rels := rels6916(res)
+
+			for _, id := range ids {
+				if strings.HasPrefix(id, "Config:") {
+					t.Errorf("a multi-target assignment minted %q. The App is bound to `app`; "+
+						"any other name here makes the CALLS edge resolve to the WRONG variable "+
+						"(%q is a string literal), which is worse than the dangling edge it replaced.",
+						id, tc.wrong)
+				}
+			}
+			for _, r := range rels {
+				src, _, ok := strings.Cut(r, " --")
+				if !ok || !strings.HasPrefix(src, "Config:") {
+					continue
+				}
+				if slices.Contains(ids, src) {
+					t.Errorf("the edge %q BINDS: its source %s names an entity emitted by this "+
+						"same run. It dangled before #6916 and must go on dangling — resolving it "+
+						"to %q (a string literal, not the App) is a confidently wrong answer.",
+						r, src, tc.wrong)
+				}
+			}
+
+			// Positive control — without it both loops pass on a fixture that
+			// stopped producing anything at all.
+			if !slices.Contains(ids, "Component:EnvStack") {
+				t.Errorf("positive control missing: this fixture no longer emits "+
+					"Component:EnvStack, so the absence assertions above prove nothing. Entities:\n  %s",
+					strings.Join(ids, "\n  "))
+			}
+		})
+	}
+}
+
+// TestIssue6916_PlainAppCallIsNotACDKApp grades the rule in the OVER-FIRING
+// direction — the direction every mutant on the first round of this change
+// missed, and the one #7013 measured at 24.5% of minted framework entities
+// landing in repos that do not use the framework (Config at 60.5%).
+//
+// The interaction is worth stating: making the Name useful also makes an
+// over-firing hit HARDER to spot. Before #6916 a stray match was named `App(`,
+// visibly junk; now it would be named `app`, which reads as a real CDK
+// application.
+func TestIssue6916_PlainAppCallIsNotACDKApp(t *testing.T) {
+	// Negative: `const app = App()` in a file with no CDK anywhere.
+	ids := entityIDs6916(detect6916(t, "src/bootstrap.tsx", "typescript", cdk6916TSPlainAppCall))
+	for _, id := range ids {
+		if strings.HasPrefix(id, "Config:") {
+			t.Errorf("a plain `App()` call in a non-CDK file minted %q. `new` is the only thing "+
+				"separating this rule from every JS file that calls a function named App().", id)
+		}
+	}
+
+	// Positive control: the SAME file with `new` added must mint Config:app.
+	// Without this the assertion above would also pass if the rule had stopped
+	// firing on this path, this language, or this content entirely.
+	ctrl := entityIDs6916(detect6916(t, "src/bootstrap.tsx", "typescript", cdk6916TSNewAppCall))
+	if !slices.Contains(ctrl, "Config:app") {
+		t.Errorf("positive control missing: `const app = new App()` at the same path no longer "+
+			"mints Config:app, so the absence asserted above proves nothing about `new`. Entities:\n  %s",
+			strings.Join(ctrl, "\n  "))
 	}
 }
--- a/internal/engine/aws_cdk_app_name_6916_test.go
+++ b/internal/engine/aws_cdk_app_name_6916_test.go
@@ -1,0 +1,288 @@
+package engine
+
+import (
+	"context"
+	"fmt"
+	"regexp"
+	"slices"
+	"sort"
+	"strings"
+	"testing"
+
+	"github.com/cajasmota/grafel/internal/extractor"
+)
+
+// #6916 (Tier B, aws_cdk arm): the two aws_cdk `Config` source_patterns used
+// `name_group: 0`, so the entity's Name was the whole regex match —
+// `"new cdk.App("` (JS/TS) and `"cdk.App()"` (Python).
+//
+// The relationship rule in the SAME file could therefore never bind:
+//
+//	source_type: Config
+//	target_type: Component
+//	relationship: CALLS
+//	source_group: 2   // the VARIABLE in `new MyStack(app, 'Id')` -> "app"
+//
+// The edge's source name is the variable `app`; the entity's name was the
+// marker string `new cdk.App(`. Those two strings can never be equal, so the
+// edge shipped dangling at its source end — the #6906 `bug-extractor` shape.
+//
+// Narrowing the capture to the assigned variable (the express.yaml:91 shape)
+// makes an ALREADY-EMITTED edge start binding. These tests pin the ARTEFACT:
+// the entity's Name and the edge with BOTH endpoint IDs, driven through the
+// real LoadAllRules() + Detector.Detect() path. A count is not evidence.
+//
+// The detector builds a relationship endpoint ID as "<type>:<captured name>"
+// (detector.go:562) and an entity carries Kind + Name; the source end BINDS
+// exactly when some emitted entity has Kind/Name equal to the edge's FromID
+// halves. That equality is what these tests assert, in both directions.
+
+// cdk6916TSApp is the canonical CDK entry point (`bin/app.ts`). The Stack class
+// is declared in the same file so BOTH endpoints of the CALLS edge are emitted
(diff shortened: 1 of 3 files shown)


cajasmota:
## Coordinator mutant on `7fe46670f` — **CV-1 is ALIVE and production-reachable**

Own detached worktree at `/private/tmp/.../coord-mut-7015-h5r/wt` (absolute). Green baseline immediately before (`./internal/engine/` exit 0). Restored by `cp` from a SHA-256-verified backup (`e788e96c…` both ways); `git diff HEAD` and `git status --porcelain` empty afterwards.

The fix is right and the demonstration is the right one — the edge genuinely binds now, asserted as the edge plus endpoint set rather than a count. And finding `M2b` ALIVE on your own first score, because Python has no declarator keyword and the JS fixture could not reach it, is the fifth instance of that pattern this week and **the first caught by an author rather than a reviewer.**

Every mutant scored so far — yours, five of them — moves the rule in the **restrictive** direction: revert `name_group`, widen the capture so it breaks binding, make the assignment optional so nothing matches. **None asks whether the rule can start minting things it should not.** That is the direction #7013 was re-ranked **P1** on two hours ago.

### CV-1 — make `new` optional

`javascript_typescript/frameworks/aws_cdk.yaml`, the App-instantiation pattern:

```
\s*=\s*new\s+(?:cdk\.)?App\s*\(     →     \s*=\s*(?:new\s+)?(?:cdk\.)?App\s*\(
```

Exactly one occurrence, asserted before the edit. **Result: ALIVE** — `./internal/engine/` exit 0.

### The distinguishing input is ordinary React, in a repo with no CDK

```tsx
import App from './App';

function bootstrap() {
  const app = App();
  return app;
}
```

Under the mutant, driven through the real `LoadAllRules()` + `Detect()` path at `src/bootstrap.tsx`:

```
Config entities minted from a file with NO CDK: [Config:app]
```

`App()` as a plain call is a React component invocation, not a CDK constructor. **`new` is the only thing separating this rule from firing on any JS file that calls a function named `App`** — and after this PR the entity is named `app`, which reads as a perfectly plausible CDK application. Before this PR it would at least have been named `App(`, which is visibly junk.

So there is a real interaction worth naming: **making the name useful also makes an over-firing hit harder to spot.** That is not an argument against the change — it is an argument for pinning the boundary while the rule is being touched.

### Why this specific direction matters right now

#7013 measured **1,757 of 7,172 minted framework entities (24.5%) landing in repos that do not use the framework**, with `Config` at **60.5% polluted** — and `Config` is exactly the kind this rule mints. `pulumi.yaml:45` is the top offender at 668 wrong `Config` entities from an `export const x =` pattern, which is the same shape of over-broad match.

This PR is the **template for 37 more sites** (Tier A's eight are the #6914 anchor). If the template's test suite grades only the restrictive direction, every site copied from it inherits an ungraded over-firing boundary — into the kind that is already 60% noise.

### Asked change — one negative fixture

Add a case asserting that a **non-CDK** `App()` call mints **no** `Config`. It is a few lines against the harness you already have, and it makes `new` a graded part of the pattern rather than an incidental one. `TestIssue6916_UnassignedAppMintsNoConfig` is the right neighbour to put it beside.

I am not asking for a general over-firing sweep here — that is #7013's job, and it needs the `requires_framework` mechanism that #7014 shows is currently inexpressible for 22 of 35 rule files. One fixture, for the boundary this rule actually has.

### On the two decisions you flagged

**The unassigned-App hold is right, and I am taking it.** The corpus contains zero CDK, so the structural argument is all there is. Tier A should not copy this shape until the assigned-vs-unassigned split is measured against a real CDK population — that is 8 sites and the #6914 anchor riding on an unmeasured assumption. Recorded on #6916.

**The no-`fbversion` reading is sound** and `internal/types/kinds.go:26` is the right authority: a rename must not force-reindex every group. Saying so in the body, with the #6085 mixed-ID consequence stated rather than buried, is what that question needed.

cajasmota:
## Independent adversarial review — **REQUEST-CHANGES** (small, concrete asks)

Measured in my own detached worktree at `7fe46670f`,
`/private/tmp/claude-501/-Users-jorgecajas-Projects-archigraph/c0fefed4-0195-4c81-9a94-07a7145710a4/scratchpad/review-7015-zq8m/wt`,
**macOS/APFS, Darwin 25.6.0** (#6966 gate — every figure below is that platform only). Rule files
restored by `cp` from a shasum-verified backup after every mutant; `git diff HEAD` empty and re-shasummed each time.

The core claim holds. I am asking for changes because **this is the Tier A template**, and the shape has
a seam the fixtures do not see that turns a *dangling* edge into a *confidently wrong bound* edge.

### Verification (re-run independently, not taken from the body)

| leg | result |
|---|---|
| `go test ./internal/engine/ -count=1` (`go vet` clean first) | **ok, 48.2s** |
| `go test ./cmd/grafel/ -count=1`, fresh `HOME`+`GRAFEL_HOME`+`GRAFEL_DAEMON_ROOT` under the scratchpad | **ok, 152.6s** |
| `scripts/quality/run.sh --ratchet` | **exit 0**, `38 gated fixtures held their baseline (29 at 100% must-have recall)` |
| no baseline moved — **independent of that line**: `git status --porcelain` after the ratchet | **empty** |

### (a) Does the edge genuinely bind? **Yes — and the pin asserts the binding, not just the name.**

Reproduced through the real `LoadAllRules()` + `Detect()`. TS and Python both emit
`Config:app --CALLS--> Component:ApiStack` with **both** endpoints present in the same run's entity set;
at `HEAD^` the entity set is `Config:"new cdk.App("` / `Config:"cdk.App()"` and the source end names nothing.

I attacked the binding half directly rather than the entity rule (RM1/RM1b below) and both are DEAD, so
the pin is not asserting the wrong half.

**But one fixture is name-only, and its stated rationale is false.** On `cdk6916PyAttrApp`
(`self.app = cdk.App()` + `ApiStack(self.app, "ApiStack")`) the emitted relationship set is **empty** —
the Python CALLS rule is `(\w+Stack)\s*\(\s*(\w+)\s*,` and `self.app,` has no `,` after `(\w+)`. The test
comment says `(.+)` captures `self.app`, "which the CALLS rule's `(\w+)` source group can never equal" —
implying `app` *can*. It cannot: **no edge exists there at all, under either pattern.** The JS twin has the
same property (`this.app = new cdk.App()` yields `Config:this --CALLS--> …`, still dangling). Please correct
that comment; the mutant kill is real, the reason given for it is not.

### (b) What the unassigned forms lose — **nothing that ever bound. Not a Tier A blocker.**

Tested, not read. Six unassigned shapes driven through the detector, before (`HEAD^`) vs after:

| form | before | after |
|---|---|---|
| `new cdk.App();` | `Config:"new cdk.App("`, no edge | nothing, no edge |
| `new ApiStack(new cdk.App(), 'X')` | `Config:"new cdk.App("` + `Config:new --CALLS--> …` (dangling) | marker gone, same dangling edge |
| `new cdk.App().synth()` / `return new cdk.App()` | marker, no edge | nothing, no edge |
| `cdk.App()` / `cdk.App().synth()` / `ApiStack(cdk.App(), "X")` (py) | `Config:"cdk.App()"`, no edge | nothing, no edge |

Your structural argument survives the empirical check: in every unassigned shape the marker node had
**no edge to anchor**, and the one edge that is emitted (`Config:new`) dangled before and dangles now.
The only thing lost is a per-file boolean marker. I confirm the hold you offered is **not needed** on this ground.

Two *pre-existing*, unchanged gaps found while measuring, worth knowing before Tier A quotes a recall number:
`app = cdk.App(outdir="cdk.out")` matches neither the shipped nor the new Python pattern (`App\s*\(\s*\)`
is empty-parens-only), yet the CALLS rule still emits `Config:app --CALLS--> …` — a dangling edge on a very
common real idiom. `if (app := cdk.App()):` is the same. Neither is caused by this PR; neither is pinned.

### (c) fbversion: **your reading is correct.** KindVocabularyVersion: correct mechanically, with one doc snag.

`fbversion.Version` gates the FlatBuffers payload format, `internal/graph`'s min-version check rejects the
stored graph on a bump and `internal/daemon` turns that into a fleet-wide reindex — `kinds.go:36-40` rules
that class of change out explicitly. This PR changes no format, schema or kind. **Agreed: no bump.**

`KindVocabularyVersion` is also correctly untouched — its guard
(`internal/types/kind_vocabulary_bump_guard_6779_test.go:275`) keys on the **EntityKind constant set**
(added/removed), and `Config` is a YAML `entity_type`, not a member of that set.

The snag is in `kinds.go` itself, not in this PR: its worked example for v1 cites **#6499, "made Kotlin
`SCOPE.Operation` names class-qualified"** — a *Name*-spelling change — as one of the "two such renames".
By that example #6916 is the same class; by the definition and by the guard it is not. Worth a one-line
issue against `kinds.go` so the next Tier A arm does not have to re-litigate it. Not a blocker here.

### Mutants (mine, all `-count=1`, green baseline immediately before each; exact-occurrence assert + hit set printed per apply — `source_group: 2` appears once in **each** file, so both were confirmed individually)

| # | mutant | verdict |
|---|---|---|
| RM1 | **JS CALLS `source_group: 2` → `1`** (attack the binding, not the name) | **DEAD** — `TestIssue6916_CDKCallsEdgeBindsAtItsSource/typescript` |
| RM1b | Python CALLS `source_group: 2` → `1` | **DEAD** — same test, python leg, names the missing edge |
| RM2 | Python `App\s*\(\s*\)` → `App\s*\(` (permissive: allow args) | **ALIVE — distinguishable and ungraded.** Full `./internal/engine/` green. Distinguisher: `app = cdk.App(outdir="cdk.out")` mints `Config:app` under the mutant, nothing under HEAD. The mutant is *more* correct; the empty-parens restriction is pinned by nothing |
| RM3 | JS: delete `(?:(?:const\|let\|var)\s+)?` | **ALIVE — equivalent under the detector.** `(\w+)` alone already backtracks to the variable on `const`/`let`/`var`, multi-declarator and attribute lines included; I could construct no input where the group changes the emitted set. The body says the optional declarator "was found by a failing test" — the group is inert |
| RM4 | JS: delete `(?:\s*:\s*[\w.]+)?` | **ALIVE — distinguishable and ungraded, production-reachable.** Full `./internal/engine/` green. On `const app: cdk.App = new cdk.App();` the mutant mints **`Config:App`** (matched inside the annotation) and the edge dangles; HEAD mints `Config:app` and binds. The group is load-bearing and **no fixture carries a type annotation** — in TypeScript CDK that is not an exotic shape |
| RM5 | Python `(\w+)` → `([\w.]+)` (dotted capture) | **DEAD** — identifier assertion, `"self.app"` |

### The blocking finding — a false edge that **binds**

The new Python capture has no left anchor, so on a multi-target assignment it locks onto the **wrong**
variable. Driven through the real detector:

```python
app, env = cdk.App(), "prod"
EnvStack(env, "EnvStack")
```
→ `entities=[Component:EnvStack Config:env …]`, `rels=[Config:env --CALLS--> Component:EnvStack]`

`env` is the string `"prod"`. **Before this PR that edge dangled** (entity was `Config:"cdk.App()"`); after it,
it resolves — a wrong answer stated confidently, which is worse than the marker it replaces. The suite cannot
see it: every assertion checks `Config:app` is present and identifier-shaped, and `env` is identifier-shaped.
`(.+)` → `(\w+)` was graded; `^` was not. Same axis, unguarded neighbour — the fifth-instance pattern again.

### Asks (all small)

1. **Anchor the capture** so it cannot cross a `,` — e.g. Python `(?m)^\s*(?:\w+\.)?(\w+)\s*=\s*(?:cdk\.)?App\s*\(\s*\)`,
   and the JS equivalent — **or** pin the mis-attribution as a known cost the way you pinned the unassigned form.
   Either way it needs a fixture; the current set grades neither. ~10 lines.
2. **Add a type-annotated TS fixture** (`const app: cdk.App = new cdk.App();`) to kill RM4. ~5 lines.
3. **Fix the `cdk6916PyAttrApp` comment** — that fixture emits no relationship at all, so it grades the Name, not the binding.
4. Optional / not blocking: file the `App(**kwargs)` gap (RM2) and the `kinds.go` #6499-example inconsistency as their own issues.

(1) and (2) are what I would want settled *before* this shape is copied to the other 37 sites; (3) is a comment fix.
Everything else in the PR — the binding claim, the cost analysis, the migration call, the ratchet — I confirm.

cajasmota:
## Round 2 — `3fbf4f668`. Blocker fixed, all three ALIVE mutants now DEAD on the **full** `./internal/engine/` package.

Same worktree, `/private/tmp/claude-501/-Users-jorgecajas-Projects-archigraph/c0fefed4-0195-4c81-9a94-07a7145710a4/scratchpad/impl-6916b-q7k2v/wt`, **macOS/APFS, Darwin 25.6.0** (#6966 gate). Rule files restored by `cp` from shasum-verified backups after every mutant (`f79012a9…` JS, `83a026b9…` Python, re-shasummed each time); `git status --porcelain` shows only the three intended files.

### Blocker — anchored, and the pin asserts the right thing

You were right and the fixture proves it: unanchored, `app, env = cdk.App(), "prod"` minted `Config:env` and `Config:env --CALLS--> Component:EnvStack` **resolved**.

Python is now `(?m)^[ \t]*(?:\w+\.)?(\w+)\s*=\s*(?:cdk\.)?App\s*\(\s*\)`.

**One correction to the ask, found by the fixture failing.** My first attempt asserted "no relationship is sourced on a Config here". That assertion is **false** — the relationship rule is independent of the entity rule and fires on the `EnvStack(env,` text alone, so the edge is emitted either way. The real guarantee, and what is now asserted, is that **the edge does not BIND**: its source `Config:env` names no entity emitted by the same run. It dangled before #6916 and it goes on dangling. Asserting absence would have been a pin on something untrue.

`TestIssue6916_MultiTargetAssignmentDoesNotMisattribute` — Python tuple assignment + the JS destructuring twin, `Component:EnvStack` as positive control on each.

**JS needs no anchor, and that is now demonstrated rather than assumed.** The JS pattern requires `= new … App(` immediately, so `(\w+)` can only ever be the token directly left of that `=` — in JS, always the assignment target. `const [app, env] = [new cdk.App(), 'prod']` matches nothing. That is the second leg of the test above, not a paragraph.

### Ask 2 — annotated TS fixture added

`const app: cdk.App = new cdk.App()` now appears in **both** the Name table and the binding table.

### Ask 3 — CV-1 graded, with a control

`TestIssue6916_PlainAppCallIsNotACDKApp`: your exact React file (`const app = App()` at `src/bootstrap.tsx`, no CDK marker anywhere) must mint no `Config`. Its positive control is **the same file with `new` added**, which must mint `Config:app` — so the absence cannot pass because the rule stopped reaching that path, language or content. The interaction you named (a useful name makes an over-firing hit *harder* to spot) is written into the test comment, since that is the reason the fixture has to exist rather than being obvious from the rule.

### Ask 4 — both prose errors fixed

- `cdk6916PyAttrApp`'s comment now says what it actually pins: **the Name only**. Measured, it emits **no relationship at all** under either spelling — the old comment's "which the CALLS rule's `(\w+)` source group can never equal" implied `app` *would* bind there, and it does not.
- The body claim that the JS declarator group "was found by a failing test" is **corrected in the body**. RM3 is right: the group is inert, so I **deleted it** rather than keeping a group nothing grades.

### Mutants — re-scored, exact-occurrence assert per edit, green baseline immediately before each, `go vet` first, `-count=1`

The three you flagged were confirmed against the **full** `./internal/engine/` package, not just the 6916 tests:

| # | mutant | before | now |
|---|---|---|---|
| **B1** | Python anchor `(?m)^[ \t]*(?:\w+\.)?` deleted (the pre-review pattern) | — | **DEAD** — names both the minted `Config:env` and the edge that binds |
| **CV-1** | JS `new\s+` → `(?:new\s+)?` | ALIVE | **DEAD** — `a plain App() call in a non-CDK file minted "Config:app"` |
| **RM4** | JS annotation group deleted | ALIVE | **DEAD** — `Config:app` absent, edge reported dangling |
| RM5 | Python `(\w+)` → `([\w.]+)` | DEAD | **DEAD (regraded)** |
| M1a / M1b | `name_group: 1` → `0`, per language | DEAD | **DEAD (regraded)** |
| M2a / M2b | variable capture → `(.+)`, per language | DEAD | **DEAD (regraded)** |
| M4 | full revert of both sites | DEAD | **DEAD (regraded)** |
| RM3 | JS declarator group deleted | ALIVE, equivalent | **N/A — group deleted** |

**Two things worth flagging from the re-score, both self-caught:**

1. **RM5 went ALIVE under my own fix and I nearly shipped it that way.** The new `(?:\w+\.)?` prefix greedily eats `self.`, so on `self.app = cdk.App()` both `(\w+)` and `([\w.]+)` name it `app` — the mutant the reviewer had killed became indistinguishable on the only fixture that could see it. A **multi-level** attribute chain (`self.inner.app = cdk.App()`) separates them again: HEAD mints nothing, the mutant mints `Config:inner.app`. **DEAD.** A fix that anchors a capture can silently un-grade a neighbouring mutant; re-scoring the whole set rather than only the new mutants is what caught it.

2. **M1a/M1b first came back ALIVE from an edit that never landed.** `name_group: 1` occurs **twice** per file — the App rule and the CfnOutput rule — so the `assert count == 1` refused to apply and the scoring run measured unmodified HEAD. The assert did its job; the "ALIVE" was mine to discard, not a finding. Re-scored with the pattern line included in the match anchor: both **DEAD**.

### Verification (all re-run at `3fbf4f668`)

| leg | result |
|---|---|
| `go vet ./internal/engine/` + `gofmt -l internal/engine/` | clean |
| `go test ./internal/engine/ -count=1` | **ok, 44.4s** |
| `go test ./cmd/grafel/ -count=1`, fresh `GRAFEL_HOME` + `GRAFEL_DAEMON_ROOT` under the scratchpad | **ok, 145.8s** |
| `scripts/quality/run.sh --ratchet` | **exit 0**, `38 gated fixtures held their baseline (29 at 100% must-have recall)` — **no baseline moved**, nothing bumped, `--update-baseline` never run |
| `git status --porcelain` after the ratchet | only the three intended files |

### On the settled decisions

Recorded, not re-litigated: the unassigned-App hold is lifted on your six-shape measurement (every unassigned form lost only a marker with no edge to anchor, and `Config:new` dangled before and after), and the `fbversion` / `KindVocabularyVersion` reading stands.

Filed your RM2 finding as #7016 rather than folding it in — `app = cdk.App(outdir="cdk.out")` matches neither the shipped nor the new Python pattern, yet the CALLS rule still emits a dangling `Config:app` edge on a very common real idiom. It is pre-existing, it is a *recall* gap rather than a correctness one, and widening `App\s*\(\s*\)` on this PR would be an ungraded change riding on a graded one.

Refs #6916, #7013.

cajasmota:
## CV-1 re-scored on `3fbf4f668` by me — **DEAD**

I found it, so I confirmed the kill myself. Same worktree at `3fbf4f668`, green baseline immediately before, occurrence count asserted **1**.

**Result: DEAD** — `TestIssue6916_PlainAppCallIsNotACDKApp`, failing on exactly the fixture added for it. Restored, `git diff HEAD` and `git status --porcelain` empty.

### You corrected my ask, and you were right

I asked you to assert that a multi-target assignment emits **no relationship sourced on a Config**. That is false, and your own fixture failing is what surfaced it: the CALLS rule is independent of the entity rule and emits `Config:env --CALLS--> Component:EnvStack` from the `EnvStack(env,` text either way.

The guarantee that actually matters — and the one you pinned — is that the edge **does not bind**: its source names no emitted entity. It dangled before this PR and it goes on dangling. That is the correct property, and it is a sharper statement than the one I asked for.

I would rather be corrected by a failing fixture than have an approximate instruction implemented literally. Two rounds now where the arm has been better than its brief.

### The two self-caught traps are the part I want on the record

**RM5 went ALIVE under your own fix.** The new `(?:\w+\.)?` prefix eats `self.`, which made `([\w.]+)` indistinguishable on the only fixture that could see it — so a mutant that had been dead was resurrected *by the repair*, silently. Catching that required re-scoring the whole prior set against the new code rather than assuming earlier verdicts carry. They do not: **a mutant verdict is only valid against the code it was scored on.**

**M1a/M1b first read ALIVE from an edit that never landed.** `name_group: 1` occurs twice per file, the count assertion refused to apply, and the run measured unmodified HEAD. That is precisely the failure mode the occurrence-count discipline exists for — an unapplied edit reads as ALIVE and argues the opposite of the truth. The assertion did its job; discarding the run rather than reporting it was right.

### Where the arm ended up

Three defects found on **2 sites** that would otherwise have shipped across **10** — the unanchored capture turning a dangling edge into a wrong-binding one, type-annotated TS naming the type instead of the variable, and the over-firing boundary. Plus the inert declarator group **deleted** rather than re-described, which is the right resolution when prose and code disagree about whether something is load-bearing.

Filing RM2 as **#7016** yourself rather than folding it in was also right: it is pre-existing, recall-direction, and would have been an ungraded widening on `Config` — the kind #7013 measures at 60.5% polluted.

**Merging on green.** Tier A can proceed from this template once it lands; I have recorded on #6916 that the unassigned-App hold is lifted and why.

PR #7017, opened 2026-09-08
docs(#6911): the `SCOPE.` prefix is an accident for 11 kind pairs and load-bearing for the 12th — and only a comment said so

### LINKED ISSUE #6911: The `SCOPE.` prefix became semantically load-bearing for one kind pair, while arms B4–B8 still document it as an accident
The `SCOPE.` prefix became semantically load-bearing when #6776 admitted bare `Endpoint` as a distinct member — and arms B4–B8's comments still say the prefix is an accident

Found by the independent review of PR #6908, the final arm of #6776. **Not a defect in that PR** — the review CONCURS with its disposition and the arm is correct. This is the documentation debt the decision leaves behind, filed so the older comments do not mislead the next reader.

## What changed

#6776's whole framing, across arms B4–B8, was that the missing `SCOPE.` prefix on rule-declared kinds is an **accident, not a second namespace**. The issue body says so explicitly, and the evidence given was that nothing in the code treats `Route` differently from `SCOPE.Route`.

Arms B5–B8 acted on that reading: every pair they migrated was a **true synonym**, folded into the prefixed spelling.

The final arm could not. Bare `Endpoint` is **Electron IPC**; `SCOPE.Endpoint` is the **HTTP** concept — settled by #6820 and shipped as `9706a5cde` / `fa05d3e2c`. So it entered `types.AllEntityKinds()` as `EntityKindEndpointBare`, a **distinct member alongside** `SCOPE.Endpoint`.

That is the right call — see below — but it has a consequence the series' own comments now contradict: **there now exists a pair of kinds whose only distinction is the `SCOPE.` prefix, and that distinction carries meaning.** The prefix is load-bearing for exactly one pair, in a codebase whose comments say it is meaningless.

## Why the arm is still right, and why this is debt rather than a defect

The honest alternative is to rename the Electron kind to something like `IPCChannel`, making the two names differ by more than a prefix. That was **priced and declined by the owner** in #6820, whose ruling was to fix the ten consumer sites with **no rename and no stored-graph migration**.

That leaves exactly two options for the last ledger row: admit it as a member, or leave a kind grafel itself produces permanently invalid. The arm chose membership, and the reviewer's independent verification strengthens the argument for it: **`IsValidEntityKind` has zero non-test production callers** anywhere in `internal/` or `cmd/`. `detector.go:466` writes `Kind: sp.entityType` verbatim, so membership never gates production — it only decides whether grafel's own validator rejects an entity grafel emitted.

So this is not a back-door namespace introduced by stealth. It is a naming debt created by a deliberate, reasoned owner decision, and the only thing wrong is that nothing says so where a reader will look.

## What to do

Cheap, and no code behaviour changes:

1. **Correct the stale framing.** #6776's body and the B4–B8 comments assert the prefix is an accident. That is now true of every kind **except** the `Endpoint` / `SCOPE.Endpoint` pair. Note the exception where the claim is made.
2. **Say it at the declaration.** `internal/types/kinds.go` should record, at `EntityKindEndpointBare`, that the bare spelling is Electron IPC, that the prefixed one is HTTP, that they are **not** synonyms, and that the rename was considered and declined in #6820 — so the next person to tidy up "an obviously missing prefix" does not merge them and silently break the Electron surface.
3. **Decide whether the rename is worth revisiting later.** Not now, and not without the owner: it is a rename plus a stored-graph migration, which is the exact cost #6820 declined. Recording the option is the point, not reopening it.

## Not established

- Whether any **other** pair in `AllEntityKinds()` differs only by the `SCOPE.` prefix while meaning different things. **Only the `Endpoint` pair was examined.** Given #6776 spent eight arms discovering that this population was larger and stranger than it looked each time, do not assume this is the only one — a sweep for prefix-only pairs is cheap and would settle it.
- Whether any consumer keys on the prefix itself (as opposed to on the full kind string) in a way that would make the distinction load-bearing in code rather than only in naming.

Refs #6776, #6820, #6908, #6893, #6894.

cajasmota:
Fixes #6911.

## The debt

#6776's B-series settled that a missing `SCOPE.` prefix is an **accident, not a second namespace**, and arms B5–B8 acted on that reading by folding true synonyms into the prefixed spelling. Arm B9 could not: bare `Endpoint` is an **Electron IPC channel**, `SCOPE.Endpoint` is the **HTTP entrypoint**, and #6820 **priced and declined** the rename that would make the two names differ by more than a prefix.

So there is now exactly one pair of kinds whose only distinction is the `SCOPE.` prefix, and that distinction carries meaning — in a codebase whose comments say the prefix is meaningless.

## Why this is more than comments

The issue asks for documentation. A comment does not fail a build, and the hazard is build-shaped: someone tidying up "an obviously missing prefix" merges the two kinds and silently re-labels every Electron IPC channel as HTTP. Comments **and** a pin.

- **`internal/types/kinds.go`, at `EntityKindEndpointBare`** — a DO-NOT-FOLD paragraph: the eleven pairs where the prefix really is an accident, the one where it is not, and the load-bearing clause that the rename **was considered and declined by the owner in #6820**. Without that clause the next reader re-derives the decision or overturns it.
- **The two sites that make the accident claim** — `internal/entkinds/rule_declared_kinds_sweep_guard_6744_test.go`'s "The namespace question, answered", and `producer_entity_kinds_6776_test.go`'s header + `goUnprefixedKindsDeferred` — now record the exception where the claim is made.
- **`internal/types/prefix_only_pairs_6911_test.go`** — the pin.

## The issue's open question, answered

"Whether any **other** pair differs only by the `SCOPE.` prefix while meaning different things" is not left open. **Twelve** prefix-only pairs exist in `AllEntityKinds()`; eleven are the `*Bare` synonym constants (Component, Model, Schema, View, Config, Operation, Route, Service, Constraint, Plugin, Template) admitted by arms B6/B7/B8 as spellings, and `Endpoint` is the twelfth. The test carries that classification as a table, so a **thirteenth pair cannot appear without an author saying which of the two readings it is**.

## Mutants

Scored in the worktree, `-count=1`, `go vet` first, off a green baseline, each edit proved landed by an exact occurrence count.

| Mutant | Verdict |
|---|---|
| Fold by repointing — `EntityKindEndpointBare = "SCOPE.Endpoint"` | **DEAD**, both new tests, with the Electron-IPC-vs-HTTP reason in the failure text |
| …same mutant, with the new test file moved aside | **still DEAD**, via four pre-existing tests |
| Add an unclassified prefix-only pair — bare `Queue` beside `SCOPE.Queue`, with the vocabulary roster and both declaration counts re-pinned as a real author would | **DEAD**, and `TestPrefixOnlyPairs6911_PopulationIsClassified` is the **sole** failure |

**Reported rather than glossed:** for the fold mutant the new assertion is a second lock on a door already locked — `endpoint_bare_kind_6776_test.go` (×2), the vocabulary version pin and the declaration-match test all fire. It is kept because it names #6820 in its failure text, but the coverage this PR genuinely **adds** is the third row: nothing in the tree observed an unclassified prefix-only pair before.

## Scope

No behaviour change. Nothing renamed, no `IsValidEntityKind` caller touched, no stored graph migrated. `KindVocabularyVersion` does **not** move — no kind is added or retired.

## Verification

`gofmt -l`, `go vet`, and `go test -count=1 ./internal/types/ ./internal/entkinds/` green. `./cmd/grafel/` also run green under a **fresh** `GRAFEL_HOME`/`GRAFEL_DAEMON_ROOT`, because `kinds.go` is not a `_test.go` file and holds the whole-graph digest pin; the change there is comment-only, so no extraction output can move.

## Reported, not fixed here (per brief)

`kinds.go`'s own `KindVocabularyVersion` v1 example cites #6499 — a *Name*-spelling change — as one of "two such renames", which contradicts the constant's own definition. Left alone deliberately; needs its own issue.

---

## Review follow-up (`f14e4648b`) — comments only, no assertion added or removed

**1. The "silently" claim in `kinds.go` was wrong, and in the direction that matters.** It credited only the new file for a fold that four pre-existing tests already catch — a reader could have deleted one of the four as redundant. The paragraph now names all four (`TestEndpointBare6776_*` x2, `TestEntityKindVocabularyIsPinnedToItsVersion`, `TestAllEntityKinds6830_ListsEveryDeclaredKindExactlyOnce`), plus the fifth that catches re-spelling the `electron.yaml` producers, and says the new file is a sixth lock kept only because its failure text names #6820. The fifth citation was re-verified here rather than taken on report: re-spelling all three `entity_type: Endpoint` sites fails `TestRuleDeclaredLedger6776_IsRetiredBecauseThePopulationIsEmpty`. All five names confirmed present at the files cited.

**2. Known limit, now written down: the table forces a classification, not a correct one.** Classifying a 13th pair `false` is a hard failure; classifying it `true` is free and green, and the permissive half is the one this repo keeps getting bitten in.

**Tying a `true` row to the `*Bare` constant family was built and scored, not argued about — and it does not close the gap.** Cost was not the problem (~18 lines reusing `declaredEntityKindsFromSource`):

| Mutant | Verdict |
|---|---|
| 13th pair `Queue` declared `EntityKindQueueBare`, classified `true` — the `Endpoint` situation recurring, named the way anyone would name it | **ALIVE, green** |
| ...declared `EntityKindQueue` instead | **does not compile** — the prefixed kind already owns the identifier, so the compiler *forces* the `*Bare` name |
| ...declared `EntityKindQueueUnprefixed`, classified `true` | **DEAD** — the only thing the check discriminates is orthography |
| The 11 genuine synonyms | pass |

And `EntityKindEndpointBare` is itself in that family: the one pair that is **not** a synonym carries the suffix, because the suffix names the spelling, not the concept. A check that reads as if it gates honesty while gating identifier spelling is worse than the recorded limit, so it is left unbuilt and the limit is documented in the test's own doc comment.

**Verification for this commit:** `gofmt -l`, `go vet`, `go test -count=1 ./internal/types/ ./internal/entkinds/` green; restores by `cp` from shasum-verified backups with `git diff HEAD` empty between scores. `./cmd/grafel/` skipped: the only changes are one `_test.go` and a comment in `kinds.go`, and it was already green on the parent commit. The extraction gate is `scripts/quality/run.sh --ratchet` (always-on in CI); it measures recall against fixtures and cannot move for a comment-plus-test change, so it was not run locally.

### DIFF
--- a/internal/entkinds/rule_declared_kinds_sweep_guard_6744_test.go
+++ b/internal/entkinds/rule_declared_kinds_sweep_guard_6744_test.go
@@ -65,6 +65,18 @@ package entkinds_test
 // is why `Module` validates by coincidence rather than by design. Nothing
 // documents an un-prefixed namespace, and nothing implements one.
 //
+// ONE EXCEPTION, ADDED BY #6911. The verdict above is true of every rule-YAML
+// kind except one: bare `Endpoint` (electron.yaml:41,46,52) is an Electron IPC
+// channel, while SCOPE.Endpoint is the HTTP entrypoint. #6776 arm B9 therefore
+// admitted it as a DISTINCT member rather than folding it, and #6820 priced and
+// DECLINED the rename that would have made the two names differ by more than a
+// prefix. For that one pair the `SCOPE.` prefix is load-bearing. Read the
+// paragraphs above as "an accident everywhere it has not been examined and
+// found otherwise", not as a licence to merge a prefixed and un-prefixed kind
+// on sight. See internal/types/kinds.go at EntityKindEndpointBare, and
+// internal/types/prefix_only_pairs_6911_test.go, which fails if the pair is
+// folded.
+//
 // This verdict stands and is now the repository's only one. The competing note
 // in internal/types/producer_kinds_test.go, which called the un-prefixed names
 // "intentionally outside the validator set", was retracted by #6776 arm B3; the
--- a/internal/types/kinds.go
+++ b/internal/types/kinds.go
@@ -774,10 +774,25 @@ const (
 	// OWNER IN #6820 (a rename plus a stored-graph migration); that ruling
 	// stands, so the confusable spelling is deliberate and is carried here
 	// instead. Deleting this constant, or pointing it at "SCOPE.Endpoint",
-	// silently re-labels every Electron IPC channel as an HTTP endpoint.
-	// internal/types/prefix_only_pairs_6911_test.go fails if either happens,
-	// and pins the twelve-pair population so a thirteenth pair cannot be added
-	// without someone saying which of the two readings it is.
+	// re-labels every Electron IPC channel as an HTTP endpoint.
+	//
+	// THAT FOLD IS NOT SILENT, AND THE GUARDS ARE OLDER THAN THIS COMMENT — so
+	// do not delete any of them as redundant. Both spellings of the fold
+	// (repointing this constant, or dropping it from AllEntityKinds()) were
+	// scored on #6911 and each is caught by FOUR pre-existing tests:
+	// TestEndpointBare6776_IsADistinctValidEntityKind and
+	// TestEndpointBare6776_MembershipIsNotHTTPMembership
+	// (endpoint_bare_kind_6776_test.go), TestEntityKindVocabularyIsPinnedToItsVersion
+	// (kind_vocabulary_bump_guard_6779_test.go) and
+	// TestAllEntityKinds6830_ListsEveryDeclaredKindExactlyOnce
+	// (all_entity_kinds_roster_6830_test.go). Re-spelling the three
+	// electron.yaml producers is caught by a fifth,
+	// TestRuleDeclaredLedger6776_IsRetiredBecauseThePopulationIsEmpty.
+	// internal/types/prefix_only_pairs_6911_test.go adds a sixth lock on the
+	// fold, kept only because its failure text names #6820; what it adds that
+	// nothing else observed is the NEXT pair — it pins the twelve-pair
+	// population, so a thirteenth cannot be added without someone saying which
+	// of the two readings it is.
 	EntityKindEndpointBare EntityKind = "Endpoint"
 )
 
--- a/internal/types/kinds.go
+++ b/internal/types/kinds.go
@@ -760,6 +760,24 @@ const (
 	//
 	// Spelling unchanged — nothing is renamed or retired — so
 	// KindVocabularyVersion does not move.
+	//
+	// #6911 — DO NOT FOLD THIS INTO SCOPE.Endpoint. THE MISSING PREFIX IS NOT
+	// AN ACCIDENT HERE. Twelve kinds in AllEntityKinds() have a `SCOPE.`-less
+	// twin; eleven of them (Component, Model, Schema, View, Config, Operation,
+	// Route, Service, Constraint, Plugin, Template — the `*Bare` constants
+	// above) really are one concept written two ways, which is why the #6776
+	// B-series and internal/entkinds' sweep guard call the prefix an accident.
+	// THIS PAIR IS THE ONE EXCEPTION: `Endpoint` is an Electron IPC channel,
+	// `SCOPE.Endpoint` is an HTTP entrypoint, and the prefix is the only thing
+	// telling them apart. Renaming the bare kind to something like `IPCChannel`
+	// so they differed by more than a prefix was PRICED AND DECLINED BY THE
+	// OWNER IN #6820 (a rename plus a stored-graph migration); that ruling
+	// stands, so the confusable spelling is deliberate and is carried here
+	// instead. Deleting this constant, or pointing it at "SCOPE.Endpoint",
+	// silently re-labels every Electron IPC channel as an HTTP endpoint.
+	// internal/types/prefix_only_pairs_6911_test.go fails if either happens,
+	// and pins the twelve-pair population so a thirteenth pair cannot be added
+	// without someone saying which of the two readings it is.
 	EntityKindEndpointBare EntityKind = "Endpoint"
 )
 
--- a/internal/types/prefix_only_pairs_6911_test.go
+++ b/internal/types/prefix_only_pairs_6911_test.go
@@ -20,7 +20,15 @@ package types_test
 //
 //   - TestPrefixOnlyPairs6911_EndpointPairIsNotFolded — the two kinds are
 //     present, distinct, and neither is spelled as the other. This is the one
-//     that fires when someone "fixes the obviously missing prefix".
+//     that fires when someone "fixes the obviously missing prefix". IT IS NOT
+//     THE ONLY ONE, AND IT WAS NOT FIRST: both spellings of that fold were
+//     scored on #6911 and each is already caught by four pre-existing tests
+//     (TestEndpointBare6776_IsADistinctValidEntityKind,
+//     TestEndpointBare6776_MembershipIsNotHTTPMembership,
+//     TestEntityKindVocabularyIsPinnedToItsVersion,
+//     TestAllEntityKinds6830_ListsEveryDeclaredKindExactlyOnce). This one is
+//     kept because its failure text names #6820, which none of theirs does —
+//     not because the fold was unguarded.
 //   - TestPrefixOnlyPairs6911_PopulationIsClassified — the SET of prefix-only
 //     pairs is exactly the twelve classified below. #6911 asks whether any
 //     OTHER pair means two things while differing only by the prefix; the
@@ -34,6 +42,36 @@ package types_test
 // pair SHRINKS the population, which the set assertion would also catch but
 // would report as "a pair went missing" rather than naming the concept; and a
 // new unclassified pair leaves the Endpoint pair untouched.
+//
+// # KNOWN LIMIT: only the STRICT direction is gated, and this was measured
+//
+// The table forces a thirteenth pair to be CLASSIFIED, not to be classified
+// HONESTLY. Classifying a newcomer `false` (semantically distinct) is a hard
+// failure until the twoConcept roster below moves with it; classifying it
+// `true` (synonym) is free and green, and nothing observes whether that is
+// true. The permissive half — the `Endpoint` situation recurring — is the
+// unguarded one.
+//
+// Tying a `true` row to the `*Bare` constant family was built and scored on
+// #6911 rather than argued about, and it does NOT close this. Two reasons, in
+// order of weight:
+//
+//   - The compiler forces the name. For any prefix-only pair the identifier
+//     `EntityKind<X>` is already taken by the PREFIXED kind (declaring a second
+//     one does not compile — scored), so the bare constant is pushed to
+//     `EntityKind<X>Bare` whatever the author means by it. A thirteenth pair
+//     that means two things, declared the way anyone would declare it and
+//     marked `true`, stayed GREEN under the check.
+//   - `EntityKindEndpointBare` is itself in that family. The one pair on this
+//     list that is NOT a synonym carries the `Bare` suffix, and kinds.go says
+//     why: the suffix names the SPELLING, not the concept. So family
+//     membership cannot mean "synonym" without contradicting the row this file
+//     exists for.
+//
+// The check only ever fired on a deliberately unconventional identifier, so it
+// would have graded orthography while reading as if it graded honesty. Left
+// unbuilt on purpose; there is no oracle for "same concept", and the failure
+// texts below state the intent instead.
 
 import (
 	"sort"
@@ -119,7 +157,7 @@ func TestPrefixOnlyPairs6911_EndpointPairIsNotFolded(t *testing.T) {
 			"These are two members naming two concepts — an Electron IPC channel and an HTTP "+
 			"entrypoint. The `SCOPE.` prefix looks like the accident #6776 describes and is not "+
 			"one here: the rename that would separate the names properly was declined in #6820, "+
-			"so folding them silently re-labels every Electron IPC channel as HTTP.",
+			"so folding them re-labels every Electron IPC channel as HTTP.",
 			bare, sawBare, prefixed, sawPrefixed)
 	}
 }
--- a/internal/types/prefix_only_pairs_6911_test.go
+++ b/internal/types/prefix_only_pairs_6911_test.go
@@ -1,0 +1,181 @@
+package types_test
+
+// prefix_only_pairs_6911_test.go — #6911.
+//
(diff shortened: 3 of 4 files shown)


cajasmota:

Measured in my own worktree at `/private/tmp/claude-501/-Users-jorgecajas-Projects-archigraph/c0fefed4-0195-4c81-9a94-07a7145710a4/scratchpad/review-7017-vx8/wt`, detached at `a36a5a7e1`, macOS/APFS. Fresh `GRAFEL_HOME` + `GRAFEL_DAEMON_ROOT` per run, `-count=1`, `go vet` and `gofmt -l` clean, green baseline before every score, every mutant edit proved landed by an exact occurrence count, every restore by `cp` from a shasum-verified backup with `git status --porcelain` empty afterwards.

Baseline: `go test -count=1 ./internal/types/ ./internal/entkinds/` → **ok / ok**. `./cmd/grafel/` → **ok (147s)**.

### The four questions

**(a) Did the fold pin genuinely pre-exist? YES — confirmed exactly as the author reported.** With `prefix_only_pairs_6911_test.go` moved aside and `EntityKindEndpointBare` repointed at `"SCOPE.Endpoint"`, `./internal/types/` fails on precisely four pre-existing tests: `TestEntityKindVocabularyIsPinnedToItsVersion`, `TestAllEntityKinds6830_ListsEveryDeclaredKindExactlyOnce`, `TestEndpointBare6776_IsADistinctValidEntityKind`, `TestEndpointBare6776_MembershipIsNotHTTPMembership`. The author's self-report is accurate and the PR body already says so. Credit for reporting it rather than counting it.

**(b) Does the classification test add real coverage, and can it go vacuous? Real coverage: YES. Vacuous: NO.** I reproduced the 13th-pair mutant independently (bare `Queue` beside the existing `SCOPE.Queue`, roster + both declaration counts re-pinned as a real author would) and `TestPrefixOnlyPairs6911_PopulationIsClassified` is the **sole** failure. Nothing else in the tree observes an unclassified prefix-only pair. On vacuity: the expected side is an independent literal (`classifiedPrefixOnlyPairs`), not derived from `AllEntityKinds()`; `len(got) == 0` is a real positive control that `t.Fatal`s; and truncation-by-one is demonstrably caught — mutant 3 below removes one roster member and the `want`-loop fires.

**(c) Is 12 right? YES.** I enumerated it myself out of `types.AllEntityKinds()` (94 kinds, 94 unique) with my own `strings.CutPrefix` walk, not the test's helper: `[Component Config Constraint Endpoint Model Operation Plugin Route Schema Service Template View]` — 12, and identical to the table. The 11 names in the `kinds.go` comment also match exactly.

**(d) Would anything catch the comment becoming false? Partly — better than I expected.** The *factual* half is pinned by a pre-existing test I did not know about: re-spelling all three `entity_type: Endpoint` sites in `electron.yaml` to `SCOPE.Endpoint` kills `TestRuleDeclaredLedger6776_IsRetiredBecauseThePopulationIsEmpty`, whose failure text names the three sites *and* names #6820 as the rejected rename. I also verified independently that those three sites (lines 41, 46, 52) are the **only** `entity_type: Endpoint` producers in the whole `internal/engine/rules/` tree, so the comment's site citation is exact. What is **not** graded is the English gloss: swapping which spelling is Electron and which is HTTP, in `kinds.go`, leaves `./internal/types/ ./internal/entkinds/` green. I'm not asking for that to be graded — the pattern text (`ipcMain\.handle`, `ipcRenderer\.invoke`) sits at the pinned line numbers and reads for itself.

### Mutants I scored

| # | Mutant | Verdict |
|---|---|---|
| 1 | `EntityKindEndpointBare = "SCOPE.Endpoint"`, **new test file moved aside** | **DEAD** — 4 pre-existing tests. Confirms the author's honesty finding. |
| 2 | Drop `EntityKindEndpointBare` from `AllEntityKinds()` (constant kept), new file aside | **DEAD** — 4 pre-existing tests. With the new file present: 6 failures, incl. both new tests. |
| 3 | `EntityKindTemplateBare = "Templates"` — a pair silently stops being prefix-only | **DEAD** — vocabulary pin (mechanical, an author re-pins it) **plus** `PopulationIsClassified`, which is the only *semantic* failure. |
| 4 | Add a 13th prefix-only pair (`Queue`), vocabulary roster + both declaration counts re-pinned | **DEAD**, `PopulationIsClassified` the **sole** failure. This is the coverage the PR genuinely adds. |
| 5 | …then classify the 13th as `true` (synonym) | **ALIVE — green.** See finding below. |
| 6 | …then classify the 13th as `false` (distinct) | **DEAD** — `twoConcept = [Endpoint Queue], want exactly [Endpoint]`. |
| 7 | Swap which spelling the `kinds.go` comment calls Electron vs HTTP | **ALIVE, ungraded** (prose). Priced below; I do not think it is worth grading. |
| 8 | `electron.yaml` × 3: `entity_type: Endpoint` → `SCOPE.Endpoint` | **DEAD** — `TestRuleDeclaredLedger6776_...`, pre-existing, names the sites and #6820. |

### Findings (none blocking)

1. **The table "forces a classification" only in the strict direction (mutants 5 + 6).** A future author who adds a 13th pair can go green by typing `true`, and nothing observes whether that is honest — no assertion ties a `true` row to anything. Typing `false`, i.e. classifying honestly when the pair *is* distinct, is a hard failure that forces `kinds.go` and the test to move together. So the permissive direction is the unguarded one, which is the direction this repo keeps getting bitten in. I do **not** think this is fixable — there is no oracle for "same concept" — and the failure text is explicit about the intent, so I would merge as-is. Worth naming in the PR body as a known limit rather than leaving the reader to infer "forces a classification" means "forces a correct one".

2. **The `kinds.go` comment says "silently" and credits only the new file.** "Deleting this constant, or pointing it at `SCOPE.Endpoint`, silently re-labels every Electron IPC channel as an HTTP endpoint. `internal/types/prefix_only_pairs_6911_test.go` fails if either happens" — mutants 1 and 2 show it is **not** silent and never was: four pre-existing tests fire, and the new file is the fifth. A reader who later deletes the new file will think they removed the only guard; a reader auditing the guard will over-value it. One clause naming `endpoint_bare_kind_6776_test.go` and `rule_ledger_retired_6776_test.go` alongside it would make the comment true. This is a comment-accuracy fix in a PR whose whole purpose is comment accuracy, so I'd take it — but it does not block.

3. **"No `golden`/`ratchet` Makefile target exists" is literally true and slightly misleading.** There is no such Makefile target, but `scripts/quality/run.sh --ratchet` is the gate this repo uses and it is now always-on in CI (52aaa84f1's predecessor #6750). It measures extraction recall against fixtures, so a comment-only change plus a `_test.go` cannot move it and running it locally was **not** needed — but the body should say "the gate is `scripts/quality/run.sh --ratchet` and it cannot move here", not "no target exists".

### Scope checks

- Diff is comments in three files plus one new `_test.go`. **No behaviour change** — confirmed.
- `KindVocabularyVersion` **not** bumped. Correct: no kind added or retired.
- **`./cmd/grafel/` was a reasonable precaution and it passed** (147s, green). `kinds.go` is non-test and digest-pinned, and the repo's own rule is that any change to it must run that package; nothing moved, which is the expected outcome for a comment-only edit. No finding.
- `gofmt -l internal/types internal/entkinds` clean.

Nothing I attacked came back false. The PR body's mutant table is accurate, including the row that undercuts its own value. Approve.

cajasmota:
## Coordinator mutant on `f14e4648b` — **CW-1 DEAD**

Own detached worktree at `/private/tmp/.../coord-mut-7017-k4v/wt` (absolute). Green baseline immediately before. Restored by `cp` from a SHA-256-verified backup (`cad64a87…` both ways); `git diff HEAD` and `git status --porcelain` empty afterwards.

Every mutant scored so far — yours and the reviewer's, thirteen between you — moves a **constant** or a classification row. None asks the boundary question: **can a new prefix-only pair enter through the rule vocabulary instead of the constant vocabulary?** That matters here because #6776's whole subject is the relationship between those two vocabularies, and the classification test reads `AllEntityKinds()`, which rules do not go through.

### CW-1 — introduce a bare kind via rule YAML, not via a constant

`electron.yaml:58`, `entity_type: Service` → `entity_type: Queue`. `EntityKindQueue = "SCOPE.Queue"` already exists at `kinds.go:67`, so this creates a **new prefix-only pair** — bare `Queue` beside `SCOPE.Queue` — without touching a single Go constant. Occurrence asserted, resulting `entity_type: Queue` count 1.

**Result: DEAD**, and by two independent tests rather than one:

- `TestNoUndeclaredRuleEntityKinds`
- `TestRuleDeclaredLedger6776_IsRetiredBecauseThePopulationIsEmpty`

So the constant/rule boundary is genuinely closed, and the ledger test earns its retirement claim: it does not merely assert the population is empty today, it fails when something re-populates it. That was worth confirming rather than assuming, because "the ledger is empty" reads like a statement about the past.

### On the gap you were asked to close

**Pricing it, building it, scoring it, and then not shipping it is the right outcome**, and the reasoning is better than the fix would have been:

> The compiler *forces* a `*Bare`-style name for every prefix-only pair, so the check only ever fires on a deliberately unconventional identifier — it grades orthography while reading as if it graded honesty.

And the second half is the one that settles it: **`EntityKindEndpointBare` is itself in that family**. The one non-synonym carries the suffix, because the suffix names the spelling rather than the concept. So family membership cannot mean "synonym" without contradicting the exact row this file exists for.

I asked for a measured answer rather than a guess, and "cheap but invalid" is a more useful result than either "too expensive" or a shipped check that grades the wrong property. Writing the limit into the test's doc comment **with the measurement** is the correct disposal — the next person to notice the asymmetry will find the attempt already made and the reason recorded, instead of rebuilding it.

### The prose correction went further than asked, correctly

You named all four pre-existing locks, found a **fifth** (`TestRuleDeclaredLedger6776_IsRetiredBecauseThePopulationIsEmpty`), and **re-verified it yourself** by re-spelling all three `entity_type: Endpoint` sites in `electron.yaml` rather than taking it on report. That is the same test my own mutant hit from the other direction, which is a useful independent corroboration of what it covers.

Dropping "silently" from both the comment and the new test's failure text matters for the reason the correction was asked for: a reader who believes the new file is the only lock may delete one of the five real ones as redundant.

### Process note worth keeping

Your `gh pr view --json body > file && cat >> file` ran from a non-repo cwd, `gh` failed, `&&` skipped the append, and `gh pr edit -F <empty file>` **blanked the PR body**. That is the compound-command trap in a third shape this session — the first two were a hook blocking a heredoc, and a transient 401 doing the same. The general rule is now: **never let a `gh` call and a file write share an invocation, in either order.**

It also left a failed `closure-keyword` run in the record. The check has since re-run green — three runs exist, latest SUCCESS — but it is worth knowing that a rollup query which looks for *any* FAILURE will still report this PR red. Only the "latest run per check name" form gives the right answer.

**Merging on green.**

Issue #7014, opened 2026-09-08
MEASURED: the rule loader silently drops key paths under unmodelled containers — the 14-spelling half SHIPPED (#7024); the metadata half and the top-level guard (193 red, not 0) remain


cajasmota:
Found by the #7013 sweep, which needed the framework-detection metadata and discovered most of it is either absent or written in a dialect nothing reads.

## The finding

**27 rule files declare `detection_signals.package_json_deps`. Nothing reads that key.**

The field that *is* read is `import_markers`, consumed by `requires_framework`. So 27 files carry framework-detection configuration that has no effect, while the mechanism it looks like it feeds sits unusable elsewhere.

Surrounding numbers from the same sweep, for scale:

- Only **3 of 154** source_patterns carry `requires_framework`.
- **22 of 35** rule files with source_patterns declare **no `import_markers` at all** — so for those, `requires_framework` has nothing to evaluate and cannot be turned on.

## Why it matters beyond tidiness

This is the *blocking* cause of **#7013**, where 24.5% of minted framework entities land in repos that do not use the framework (`Constraint` 98% polluted, `TestClass` 90%, `Config` 61%). The obvious remedy there — "set `requires_framework` on the over-firing rules" — **cannot be applied to most of them**, because the markers it needs do not exist. The gate is not merely switched off; it is inexpressible.

The two dialects make that worse than a plain gap: a rule author who writes `detection_signals.package_json_deps` has every reason to believe they have scoped their rule. Nothing tells them otherwise — no parse error, no warning, no test. **This is configuration asserting what no code observes**, which is the config-shaped version of the defect class this board keeps finding in prose.

## What is NOT established

- **Whether `package_json_deps` was ever read.** No git archaeology was done. It may be a removed consumer, a never-finished feature, or a copy-paste from another tool's schema. **That history changes the fix** — a removed consumer argues for restoring it, a never-finished one argues for deleting the key.
- Whether the 27 files' declared deps are *correct*, i.e. whether they would work if the key were read. Nobody has checked them against reality; they may have been written carelessly precisely because they never did anything.
- Whether other unread keys exist in the rule schema. **Only this one was found, and only incidentally.** A schema-vs-consumer audit is the measurement that would settle it, and it has not been done.

## Suggested order

1. **Establish the history of `package_json_deps`** — was there a consumer? That decides restore vs. delete.
[...]

Issue #7013, opened 2026-09-08
P1: 1,757/7,164 (24.5%) framework entities in non-using repos — variant D SHIPPED (762 removed, 43%, net +753); 920 INEXPRESSIBLE (pulumi=675) + 75 ungated remain


cajasmota:
Found by the #6916 measurement arm while enumerating `name_group: 0` sites. **These are a different defect from #6916** — the entity's *name* being the whole regex match is that issue; this is the pattern matching a construct from an unrelated framework entirely, so it would remain wrong after any `name_group` fix.

## Two measured instances

- **`internal/engine/rules/javascript_typescript/frameworks/langchain.yaml:77`** mints a `Service` entity from **`createHttpProxyAgent(`** in a **nextjs** repo. That is a proxy-agent constructor from `http-proxy-agent`; it has nothing to do with LangChain.
- **`internal/engine/rules/javascript_typescript/frameworks/electron.yaml:61`** fires on **Express's `app.on(`**. `app.on` is an EventEmitter method on any Express application; Electron's `app` object is a different `app` in a different package.

Both were observed on the 60-repo corpus, not reasoned about from the regexes.

## Why this is worth its own issue

An over-broad *entity* rule is not the same failure as a badly-named one. A wrongly-named entity is a marker nobody can point an edge at (#6916). A **wrongly-minted** entity asserts that a repo uses a framework it does not use — it pollutes the kind's population, inflates any per-framework count taken from the graph, and gives a `Service`/`Config` node to a repo that has neither.

It also means measurements over these kinds are contaminated by an unknown amount. The #6916 arm found **12 of 39** `name_group: 0` sites fire at all, with **4 sites producing 80%** of the entities — so a couple of over-firing rules can be a large fraction of what a kind looks like corpus-wide.

## The general shape, which is the part worth fixing

Neither rule is scoped to the framework it is named for. `langchain.yaml` runs on every `.js`/`.ts` file in every repo, as does `electron.yaml`; nothing requires the repo to import langchain or electron. This is the same architecture gap measured on **#7007**: `RelationshipRule` has no `requires_framework` field at all — it exists only on `SourcePattern` — and framework rule files are not scoped to repos that use the framework. These two are the *entity*-side instances of it.

**So a per-rule regex tightening fixes these two and leaves the class open.** Say which you are doing.

## What is NOT established

[...]

Issue #7007, opened 2026-09-08
CORRECTED: duplicated Go framework rules do NOT reach the graph (index.go:6463 dedupes on RelationshipID) — the cost is wasted regex work plus a first-record-wins Properties collapse that is safe only because `framework` comes from `lang`


cajasmota:
Found while deleting the four `.Use` middleware relationship rules for #6914 (PR #7006), by scoring a mutant that was expected to die and did not.

## The mechanism

The Go framework rule files — `gin.yaml`, `chi.yaml`, `echo.yaml`, `fiber.yaml` — carry **byte-identical** copies of several rules. They are named per framework, but **every Go rule file runs on every `.go` file**: nothing scopes `gin.yaml` to a repo that imports gin.

`detector.go` then dedupes entities on `entityType:name:file`. So four identical entity rules produce one entity, and **deleting any one of the four leaves the output byte-identical**, because the other three still mint it.

Measured (PR #7006, mutant M3b): deleting the `Middleware` entity rule from **chi.yaml alone** is **ALIVE** — every test still passes. Deleting it from **all four** (M3a) is DEAD. The rules are mutually masking, exactly the way two guards that only ever fire together grade neither.

**Nothing in `Detect()` output attributes an entity to the rule file that produced it**, so there is no assertion a test could make about which of the four minted it.

## Why it matters

1. **Per-file rule deletions are ungraded across this whole family.** A rewrite, a bad merge, or a tidy-up that drops one framework's copy is invisible — the suite stays green and the graph is unchanged *until the last copy goes*. Then it changes all at once, and the commit that broke it is three commits back.
2. **Relationship rules are NOT deduped the same way.** The same duplication amplified the #6914 defect: the fabricated two-argument edge `Middleware:middleware.Logger --ROUTES_TO--> Controller:middleware.Recoverer` was emitted **four times**, once per rule file, on a single source line. The issue recorded the wrong edge but not that it arrives ×4. So the duplication is not benign for edges: it multiplies them.
3. It makes the framework naming misleading. A reader of `chi.yaml` reasonably believes it governs chi repos; it governs every Go file in every repo.

## What is NOT established

- **Whether this extends beyond these four files.** Only the Go framework set was measured. The same shape is plausible wherever one language has several framework files with overlapping patterns — python, javascript_typescript and rust all have that structure — but **it was not checked, and one language is not a population.**
[...]

Issue #6393, opened 2026-08-20
eleven GraphQL rule files parse into zero-value rules that no Go code reads, so they look like engine config and are not


cajasmota:
## What

Eleven files under `internal/engine/rules/graphql/` are keyed on a top-level `server_frameworks:`, which **nothing in Go reads**:

```
graphql/frameworks/{apollo_server,hasura,gqlgen_go,pothos_jsts,strawberry_python,
                    graphql_yoga,relay_server,absinthe_elixir,aws_appsync,
                    type_graphql_jsts}.yaml
graphql/_manifest.yaml
```

- `grep -rn "server_frameworks|ServerFrameworks|package_json_dependencies" internal/ cmd/ --include='*.go'` → **zero hits**.
- `FrameworkRule` (`internal/engine/schema.go:13`) maps exactly four keys: `frameworks`, `file_conventions`, `source_patterns`, `relationship_rules`.

So each of these parses successfully into a **zero-value rule** — `name=""`, no source patterns, no relationship rules, no file conventions. They sit in the rules tree, look like rules, and contribute nothing to the engine.

## How it surfaced

While fixing #6391. `type_graphql_jsts.yaml` had failed to parse for 79 days, and I initially reported that as "TypeGraphQL detection has never worked". That was wrong, and finding out why is what exposed this: the file never drove detection because **no file with this shape does**. GraphQL extraction is implemented in Go and is green.

## Severity — measured, and lower than it first looked

My first concern was that `docs/coverage/registry.json` cites these files as evidence for capability claims. It does, but the exposure is small:

- Capabilities citing a catalog file at all: **1** (`protocol.graphql` / `service_extraction`).
- Capabilities whose cites are **only** catalog files: **0**.

That one capability also cites a non-catalog path, so **no published capability claim rests solely on a non-functional file.** Scope of that measurement: `records[].capabilities[].cites` in `registry.json`. Raw string occurrences of these paths elsewhere in the registry are higher (15, with duplicates) and I have not classified those.

So this is **not** a false-capability-claim bug. It is dead weight with two costs:

1. **Misleading to readers.** A file containing `detection.package_json_dependencies: type-graphql` reads exactly like live detection config. It fooled me for an hour, and I wrote a wrong impact assessment from it.
[...]

PR #7019, opened 2026-09-08
fix(#7014): a typo inside source_patterns/relationship_rules/file_conventions was silently dropped — strict-unmarshal the executable half only

cajasmota:
## What

Unknown keys inside `source_patterns[]`, `relationship_rules[]` and `file_conventions[]` now **fail at load**, naming the file, the line, the key and the container. Unknown keys anywhere else load exactly as before.

Three `UnmarshalYAML` methods (`SourcePattern`, `RelationshipRule`, `FileConvention`) check the mapping's keys against the set the struct models — derived by reflection from the `yaml:` tags, so adding a field to the struct admits the key automatically and the guard cannot drift from what it guards. No rule YAML changed.

## The strongest evidence: the loaded rule set is byte-identical with and without the guard

From the independent review — `LoadAllRules()` dumped to canonical JSON at `5d09def56`, then again with `schema_strict.go` removed from the package:

```
with guard:    34 langs / 520 rules / 233094 bytes   sha 98703af3602bc951b9a7d0874565f68f15a3acbd
without guard: 34 langs / 520 rules / 233094 bytes   sha 98703af3602bc951b9a7d0874565f68f15a3acbd
```

That is stronger than "the ratchet stayed green": **no baseline *can* have moved**, because the extractor's input is the same bytes. The ratchet ran green anyway (see Verification).

## Why this scope and no wider

The audit on #7014 measured three prices over the 520 loader-scoped rule files:

| scope | files red |
|---|---|
| full strict `KnownFields(true)` | 515 / 520 — a migration |
| top-level key allowlist | 254 |
| **strict scoped to the three list containers** | **0** |

The executable half of the schema is exactly modelled — zero unmodelled keys occur inside those three containers — which is why the guard is free today. I re-measured the 0-red figure myself before building (520 candidates, 0 red) and it is now a test rather than a one-off.

## What this buys

A typo in an **executable** position — `name_group` misspelled, `source_group` written where `target_group` was meant — becomes a load error instead of a silently-ignored key. That is the defect class that produced #7014: `detection_signals.package_json_deps` sat unread in **59 files** for the whole life of the tree because nothing objects to a key the loader does not model.

It already earned its keep: the guard immediately failed `TestInertRuleFiles_RealRuleIsNotFlagged`, whose "relationship rules only" fixture writes **`relationship_type:`** where the modelled key is `relationship`. That fixture has been asserting "here is a real relationship rule" with a rule that decodes to an empty `Relationship`. Fixed in this PR (test-only).

To be precise about what that incident does and does not say: **#6393's guard was doing its job.** `isZeroRule` checks `len(r.RelationshipRules) == 0`, so the assertion held on list length and only the fixture's *label* over-claimed; the one-line fix neither weakens nor strengthens the leg. What it exposes is a different, pre-existing gap — **`isZeroRule` cannot tell a real rule from a present-but-garbage one** — and that is a second benefit of this PR: for the three executable containers, a garbage entry can no longer be present at all.

## What this does NOT buy

**Nothing about the metadata half.** `frameworks.name` (266 files), `source_patterns[].scope` (329 entries, declared and unread), `file_conventions[].pattern` (0 occurrences), `file_conventions[].description` (1), the 59 `package_json_deps` declarations and the **14 misspelled spellings of the top-level container** all stay exactly as they are, silently dropped as before. Extending the guard to them turns 254 files red; that is a migration with its own decision, and it is #7013's ground.

Per the issue: `source_patterns[].scope` is **not** deleted even though it has no reader. It sits inside a strict container, so it has to stay modelled to keep loading. It stays modelled and stays unread (#6916).

## Pinned

`internal/engine/strict_containers_7014_test.go`

- **Fires**: one leg per container, plus a leg putting `package_json_deps` itself into an executable position. Each asserts the file is absent from `rules` **and** from `rep.Loaded` (not merely that a failure was recorded), and that the failure names the file, the key and the container.
- **Boundary**: five real metadata shapes — top-level `category`/`github_stars_2025`/`notes`, a misspelled top-level container, `detection_signals.package_json_deps`, unmodelled keys under `frameworks` and under `frameworks.detection` — must all still load, with their modelled fields intact. This is the scoping decision; without it the next tidy-up extends the guard and turns 254 files red.
- **0-red as a test**: all 520 embedded rule files still load, zero failures.

### Vacuity — which of the five no-op modes each assertion closes

A scan-and-assert-absence guard has five independent ways to be a no-op and a count floor closes only the first:

1. *inspects no files* → candidate floor (`>= expectedMinRuleFiles`).
2. *inspects files but never reaches the containers* → **entry floor**: the real tree must yield >= 350 executable-container entries (currently **562** = 329 `source_patterns` + 155 `file_conventions` + 78 `relationship_rules`; an earlier revision of that comment said "~445 … 38 file_conventions" — 38 is the audit's *file* count, not the entry count, and is corrected). Those entries exist only because the strict `UnmarshalYAML` built them — it is the only decode path into these types — so a guard off the real decode path cannot produce them.
3. *reaches them but does not detect a bad key* → the four positive-control legs.
4. *detects but does not act* → asserting absence from `rules` and `rep.Loaded`, not just a recorded failure.
5. *acts, but the assertion is too broad* → asserting the exact key name and file path, plus the boundary test asserting the guard does **not** fire outside the three containers.

## Recorded, deliberately not fixed here

- **The reflection walk is `t.Field(i)` only**, so an embedded / `,inline` field would allow the embedded type's lowercased *name* while rejecting the keys yaml.v3 actually inlines. Nothing is embedded in the three structs today, and it **fails closed** — rule files go red, nothing is silently accepted, and `Test7014_AllEmbeddedRuleFilesStillLoad` reddens CI before it could ship. A landmine for whoever adds an embedded struct, not a hazard now; named in a comment at the walk.
- **A YAML merge key (`<<: *base`) inside a guarded entry is now rejected**, because `<<` is not a modelled field name. **Zero** rule files use one today. Recording it so the first person to hit it knows it was a decision, not an oversight.
- **YAML aliases are checked correctly** — this is the non-obvious half. An aliased entry is resolved before the guard sees it, so a typo reached through an alias *does* fire; the reviewer verified this directly.

## Mutants

Scored in a **separate checkout** (`.../impl-7014b-q9z3m/mut`, detached at `353753a31` + these three files) so the verification tree was never mutated; green baseline immediately before each score, `go vet` first, `-count=1`, each edit applied by a script that **asserts an exact occurrence count of 1** and prints `APPLIED`, restored by `cp` from a shasum-verified backup (never `git checkout --`). `git diff --stat` back to the single intended test-file change after every leg.

| mutant | result |
|---|---|
| drop strictness from `source_patterns` only | **DEAD** — `.../source_patterns_typo` + `.../unmodelled_metadata_key_inside_source_patterns`; the other two legs stayed green, so the legs are independent |
| drop strictness from `relationship_rules` only | **DEAD** — `.../relationship_rules_typo` only |
| drop strictness from `file_conventions` only | **DEAD** — `.../file_conventions_typo` only |
| make the guard vacuous (`rejectUnknownContainerKeys` returns nil before the check) | **DEAD** — all four positive-control legs |
| extend the guard to the top level (`FrameworkRule.UnmarshalYAML`) | **DEAD, loudly** — the boundary test fails on 3 legs *and* `Test7014_AllEmbeddedRuleFilesStillLoad` reports **247 real rule files RED**, **247** real rule files RED. This is the boundary assertion doing its job. |
| **case-insensitive key matching** (`allowed[key]` → `allowed[strings.ToLower(key)]`) | **DEAD** — `.../case_variant_of_a_real_key`, and *only* that leg, so the kill is precisely attributed. Was **ALIVE** on the first revision: the case sensitivity was documented and ungraded, and a future edit lowercasing keys "helpfully" would have accepted `Name_Group` / `Source_Group` — a widening in exactly the direction this guard exists to prevent. Graded now, in both directions: the variant is rejected **and** the exactly-spelled key still decodes (`NameGroup == 1`) in the same run, so a guard that rejected both would not pass. |
| take the strict decoder off the real decode path (loader nils the three containers after unmarshal) | **DEAD** — entry floor: `produced only 0 executable-container entries ... want >= 350`. This is vacuity mode 2, which the candidate floor alone does not close. |

### 247 vs 254 — different populations, both correct

Review settled this: **247** = files carrying an unmodelled top-level key, which is the cost of extending the guard to the top level and is this PR's figure. **254** = files with no top-level `frameworks:` key at all (520 − 266). The gap is exactly 7 files that have neither — `go/frameworks/{echo,gin,chi,fiber,gorm}.yaml`, `ruby/frameworks/{sidekiq,rspec}.yaml` — pure `source_patterns`/`file_conventions` files. The audit's §4 table labels 254 as the allowlist cost; that label is the error, and #7014/#7013 are being corrected to carry **247**.

## Verification

Tree measured: `/private/tmp/claude-501/-Users-jorgecajas-Projects-archigraph/c0fefed4-0195-4c81-9a94-07a7145710a4/scratchpad/impl-7014b-q9z3m/wt`, branch `lane/7014-strict-containers` off `353753a31`. macOS 25.6.0 / arm64. All `go test` runs `-count=1`; `go vet ./internal/engine/` clean; `gofmt -l` empty.

- `go test -count=1 ./internal/engine/` — **ok** (43.6s)
- `go test -count=1 ./cmd/grafel/ ./internal/quality/` with a **fresh `GRAFEL_HOME` + `GRAFEL_DAEMON_ROOT`** — **ok** (176.5s / 34.0s)
- `scripts/quality/run.sh --runs 1 --ratchet` — **exit 0**, `quality ratchet OK — 38 gated fixtures held their baseline (29 at 100% must-have recall)`. **No baseline moved in either direction**, which is the check that matters here: a loader change that shifted extraction output would mean a rule stopped loading.

Review round (case-sensitivity leg, the reflection-walk and merge-key comments, and the floor-comment arithmetic): `go vet` clean, `gofmt -l` empty, `go test -count=1 ./internal/engine/` **ok** (51.7s). `./cmd/grafel/` and the ratchet were **not** re-run and do not need to be — the change is the strict decoder, its tests and comments, and the loaded rule set is byte-identical to the pre-guard decode, so neither can move. The reviewer independently ran both green at `5d09def56` (`./cmd/grafel/` 155.5s, `./internal/quality/` 34.6s, ratchet `exit 0`, 38 gated fixtures held).

Independently re-measured the 0-red claim before building, with a throwaway program outside the repo: **520 loader-scoped rule files, 0 red**. The test now reports the same from inside: `0 red: 520/520 rule files loaded, 562 executable-container entries decoded strictly`.

Refs #7014, #7013, #7007, #6916.

### DIFF
--- a/internal/engine/graphql_inert_rules_guard_6393_test.go
+++ b/internal/engine/graphql_inert_rules_guard_6393_test.go
@@ -197,8 +197,10 @@ func TestInertRuleFiles_RealRuleIsNotFlagged(t *testing.T) {
 		"detection markers only": "frameworks:\n  name: Yoga\n  detection:\n    import_markers:\n    - \"graphql-yoga\"\n",
 		"file conventions only":  "file_conventions:\n- glob: \"**/*.graphql\"\n  entity_type: Model\n",
 		"source patterns only":   "source_patterns:\n- pattern: \"type (\\\\w+)\"\n  entity_type: Model\n",
+		// `relationship`, not `relationship_type`: the latter is not a modelled
+		// key and, since #7014, does not load at all.
 		"relationship rules only": "relationship_rules:\n- pattern: \"implements (\\\\w+)\"\n  " +
-			"relationship_type: IMPLEMENTS\n",
+			"relationship: IMPLEMENTS\n",
 	}
 	for name, body := range cases {
 		t.Run(name, func(t *testing.T) {
--- a/internal/engine/schema_strict.go
+++ b/internal/engine/schema_strict.go
@@ -1,0 +1,160 @@
+package engine
+
+import (
+	"fmt"
+	"reflect"
+	"sort"
+	"strings"
+
+	"gopkg.in/yaml.v3"
+)
+
+// This file makes the EXECUTABLE half of the rule schema strict (#7014).
+//
+// The rule schema has two halves. The metadata half — `frameworks:`,
+// `category`, `github_stars_2025`, the fourteen different spellings of the
+// top-level container — is 95% unmodelled: 465 of 488 distinct key paths in the
+// rule tree are dropped by yaml.Unmarshal without a word. The executable half —
+// the entries of `source_patterns`, `relationship_rules` and `file_conventions`
+// — is by contrast exactly modelled: the audit on #7014 found ZERO unmodelled
+// keys inside those three list containers across all 520 loader-scoped files.
+//
+// That asymmetry is what makes this guard free. A typo in an executable
+// position (`name_group` misspelled, `source_group` written where
+// `target_group` was meant) is today indistinguishable from a key the loader
+// simply does not model: yaml.Unmarshal drops both in silence. That is the
+// defect class that produced #7014 in the first place —
+// `detection_signals.package_json_deps` sat unread in 59 files for the entire
+// life of the tree because nothing objects to a key nobody models.
+//
+// So: unknown keys INSIDE the three list containers are a load error. Unknown
+// keys ANYWHERE ELSE keep loading exactly as before.
+//
+// The boundary is deliberate, not an unfinished job. Full strict
+// `KnownFields(true)` unmarshalling turns 515 of 520 rule files red; a
+// top-level-key allowlist turns 254 red. Both are migrations with their own
+// decision to make (see #7013). Scoped to the three containers the cost today
+// is ZERO files. Widening it is a migration, not a tidy-up — see
+// Test7014_UnknownKeyOutsideExecutableContainersStillLoads, which exists to
+// make that widening fail loudly rather than land by accident.
+//
+// Key matching is exact and case-sensitive, which is marginally stricter than
+// yaml.v3's own field matching. No rule file in the tree relies on the
+// difference (measured: 0 red).
+
+// strictContainerFields returns the set of YAML keys a struct models, derived
+// from its `yaml:"..."` tags. Deriving it by reflection rather than hand-listing
+// the keys means a NEW field added to SourcePattern et al. is accepted the
+// moment it is declared — the guard cannot drift out of sync with the struct it
+// is guarding.
+func strictContainerFields(t reflect.Type) map[string]bool {
+	fields := make(map[string]bool, t.NumField())
+	for i := 0; i < t.NumField(); i++ {
+		f := t.Field(i)
+		if f.PkgPath != "" { // unexported
+			continue
+		}
+		name := strings.Split(f.Tag.Get("yaml"), ",")[0]
+		switch name {
+		case "-":
+			continue
+		case "":
+			name = strings.ToLower(f.Name)
+		}
+		fields[name] = true
+	}
+	return fields
+}
+
+// rejectUnknownContainerKeys fails if the mapping node carries any key the
+// struct type does not model. container names the list container for the error
+// message ("source_patterns", …).
+//
+// A non-mapping node is passed through untouched: that is a plain type error
+// and yaml.v3 reports it better than this function could.
+func rejectUnknownContainerKeys(value *yaml.Node, t reflect.Type, container string) error {
+	if value.Kind != yaml.MappingNode {
+		return nil
+	}
+	allowed := strictContainerFields(t)
+	var unknown []string
+	line := 0
+	for i := 0; i+1 < len(value.Content); i += 2 {
+		key := value.Content[i].Value
+		if allowed[key] {
+			continue
+		}
+		unknown = append(unknown, key)
+		if line == 0 {
+			line = value.Content[i].Line
+		}
+	}
+	if len(unknown) == 0 {
+		return nil
+	}
+	known := make([]string, 0, len(allowed))
+	for k := range allowed {
+		known = append(known, k)
+	}
+	sort.Strings(known)
+	return fmt.Errorf(
+		"line %d: unknown key %s in %s[]: the engine does not model it, so it would be silently ignored (known keys: %s)",
+		line, strings.Join(quoteAll(unknown), ", "), container, strings.Join(known, ", "))
+}
+
+func quoteAll(in []string) []string {
+	out := make([]string, len(in))
+	for i, s := range in {
+		out[i] = fmt.Sprintf("%q", s)
+	}
+	return out
+}
+
+// The three aliases below exist only to break the recursion: decoding into a
+// type that does NOT have an UnmarshalYAML method is what lets each
+// UnmarshalYAML delegate the real work back to yaml.v3.
+
+type fileConventionFields FileConvention
+
+// UnmarshalYAML rejects unknown keys inside a `file_conventions:` entry (#7014).
+func (c *FileConvention) UnmarshalYAML(value *yaml.Node) error {
+	if err := rejectUnknownContainerKeys(value, reflect.TypeOf(FileConvention{}), "file_conventions"); err != nil {
+		return err
+	}
+	var raw fileConventionFields
+	if err := value.Decode(&raw); err != nil {
+		return err
+	}
+	*c = FileConvention(raw)
+	return nil
+}
+
+type sourcePatternFields SourcePattern
+
+// UnmarshalYAML rejects unknown keys inside a `source_patterns:` entry (#7014).
+func (p *SourcePattern) UnmarshalYAML(value *yaml.Node) error {
+	if err := rejectUnknownContainerKeys(value, reflect.TypeOf(SourcePattern{}), "source_patterns"); err != nil {
+		return err
+	}
+	var raw sourcePatternFields
+	if err := value.Decode(&raw); err != nil {
+		return err
+	}
+	*p = SourcePattern(raw)
+	return nil
+}
+
+type relationshipRuleFields RelationshipRule
+
+// UnmarshalYAML rejects unknown keys inside a `relationship_rules:` entry (#7014).
+func (r *RelationshipRule) UnmarshalYAML(value *yaml.Node) error {
+	if err := rejectUnknownContainerKeys(value, reflect.TypeOf(RelationshipRule{}), "relationship_rules"); err != nil {
+		return err
+	}
+	var raw relationshipRuleFields
+	if err := value.Decode(&raw); err != nil {
+		return err
+	}
+	*r = RelationshipRule(raw)
+	return nil
+}
--- a/internal/engine/schema_strict.go
+++ b/internal/engine/schema_strict.go
@@ -40,13 +40,29 @@ import (
 //
 // Key matching is exact and case-sensitive, which is marginally stricter than
 // yaml.v3's own field matching. No rule file in the tree relies on the
-// difference (measured: 0 red).
+// difference (measured: 0 red), and the case-variant leg of
+// Test7014_UnknownKeyInsideExecutableContainerFailsLoad observes it.
+//
+// One further deliberate narrowing: a YAML MERGE KEY (`<<: *base`) inside a
+// guarded entry is rejected, because `<<` is not a modelled field name. Zero
+// rule files use one today. YAML ALIASES are unaffected — an aliased entry is
+// resolved before the guard sees it, so a typo reached through an alias fires
+// correctly (verified in review).
 
 // strictContainerFields returns the set of YAML keys a struct models, derived
 // from its `yaml:"..."` tags. Deriving it by reflection rather than hand-listing
 // the keys means a NEW field added to SourcePattern et al. is accepted the
 // moment it is declared — the guard cannot drift out of sync with the struct it
 // is guarding.
+//
+// One limit on that claim, recorded because it is a landmine rather than a
+// hazard: the walk is t.Field(i) only, so it does NOT understand an embedded or
+// `,inline` field. None of the three guarded structs has one today. If one is
+// added, this walk would allow the embedded type's own lowercased NAME and
+// reject the keys yaml.v3 actually inlines — which fails CLOSED (rule files go
+// red, nothing is silently accepted) and is caught by
+// Test7014_AllEmbeddedRuleFilesStillLoad, but it is a surprise worth naming
+// here rather than rediscovering.
 func strictContainerFields(t reflect.Type) map[string]bool {
 	fields := make(map[string]bool, t.NumField())
 	for i := 0; i < t.NumField(); i++ {
--- a/internal/engine/strict_containers_7014_test.go
+++ b/internal/engine/strict_containers_7014_test.go
@@ -34,10 +34,15 @@ import (
 
 // strict7014MinContainerEntries is the non-vacuity floor for mode 2: the number
 // of executable-container entries the strict decoder must actually have built
-// from the real embedded tree. The tree currently yields ~445
-// (329 source_patterns + 78 relationship_rules + 38 file_conventions); the floor
-// is set below that so ordinary rule authoring does not trip it, but far enough
-// above zero that a guard which never runs on real data cannot pass.
+// from the real embedded tree. The tree currently yields 562
+// (329 source_patterns + 155 file_conventions + 78 relationship_rules); the
+// floor is set below that so ordinary rule authoring does not trip it, but far
+// enough above zero that a guard which never runs on real data cannot pass.
+//
+// (An earlier revision of this comment said "~445 … 38 file_conventions". 38 is
+// the audit's FILE count for file_conventions, not its ENTRY count. Corrected
+// in review — a floor comment that misstates the population it floors is the
+// same defect class this file closes one layer down.)
 const strict7014MinContainerEntries = 350
 
 // goodRuleYAML is a rule file exercising all three executable containers with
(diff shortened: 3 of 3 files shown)


cajasmota:

Reviewer's tree: `/private/tmp/claude-501/-Users-jorgecajas-Projects-archigraph/c0fefed4-0195-4c81-9a94-07a7145710a4/scratchpad/review-7019-k4v8t/wt`, detached at **`5d09def56`**; mutants scored in a **separate** checkout `.../review-7019-k4v8t/mut` at the same commit, restored by `cp` from shasum-verified backups (both files back to `8da36cf06`/`f923079cc` after every leg, `git status` clean). macOS 25.6.0 / arm64, APFS. All `go test` `-count=1`, `go vet ./internal/engine/` clean.

---

### The strongest verification I can offer: the loaded rule set is **byte-identical** with and without the guard

Dumped `LoadAllRules()` to canonical JSON at `5d09def56`, then again with `schema_strict.go` removed from the package (i.e. pre-guard decode behaviour), same binary otherwise:

```
with guard:    34 langs / 520 rules / 233094 bytes   sha 98703af3602bc951b9a7d0874565f68f15a3acbd
without guard: 34 langs / 520 rules / 233094 bytes   sha 98703af3602bc951b9a7d0874565f68f15a3acbd
```

Identical. That is a stronger statement than "the ratchet did not move": no baseline *can* have moved, because the extractor's input is the same bytes. The ratchet ran green anyway (`exit 0`, `38 gated fixtures held their baseline (29 at 100% must-have recall)`), as did `./internal/engine/` (48.7s), `./cmd/grafel/` (155.5s) and `./internal/quality/` (34.6s) with a **fresh** `GRAFEL_HOME` + `GRAFEL_DAEMON_ROOT`.

### (a) 0-red holds, and the test cannot go vacuous in modes 1, 2 or 5

Re-measured independently with a throwaway program outside the repo (hand-retyped struct key sets, no dependency on this PR's code), over the 520 loader-scoped files:

```
loader-scoped candidates: 520
scoped-strict RED files: 0
executable-container entries: 562  (source_patterns 329, file_conventions 155, relationship_rules 78)
entries that are not mappings: 0
files containing a YAML merge key '<<:': 0
container keys differing only in case: 0
```

**Both floors exist and both fire, independently** — I did not take this on report:

- candidate floor: loader mutated to `len(parts) != 4` → `only 0 rule files found, want >= 517: the measurement is vacuous`.
- entry floor: loader nils the three containers after unmarshal → `strict decoder produced only 0 executable-container entries from the real rule tree, want >= 350`. The candidate floor stays satisfied in that leg, which is the point.

### (b) The reflection-derived key set — no live hole, one latent one

| shape | status |
|---|---|
| `yaml:"-"` | handled; no field uses it (branch dead today) |
| `omitempty` / any tag with a comma | handled by `Split(...)[0]`; **no field in the three structs has one**, so the branch is unexercised — mutant M2 below is equivalent |
| untagged field | falls back to `ToLower(f.Name)`, matching yaml.v3; no untagged field exists — mutant M3 equivalent |
| unexported field | skipped; none exist |
| YAML **alias** entry (`- *anchor`) | **checked correctly** — I fed an anchor carrying `name_grup:` through an alias into `source_patterns[]` and the guard fired, naming the anchor's line. Not a hole. |
| YAML **merge key** (`<<: *base`) inside an entry | **rejected**, with `unknown key "<<"`. New narrowing vs yaml.v3, 0 occurrences in the tree, undocumented. |
| **embedded / `,inline` struct** | **latent hole, fails closed.** `strictContainerFields` walks `t.Field(i)` only. An embedded `Foo` with no tag yields the allowed key `"foo"` while yaml.v3 inlines `Foo`'s *own* keys — so every real key of the embedded struct becomes "unknown" and every file using them goes red. |

That last one is the only thing I would ask for, and it is cheap: the "cannot drift from what it guards" claim is true for plain tagged fields and **false** for an embedded/inline one. One line in `strictContainerFields` — `if f.Anonymous { panic/t.Fatal "embedded fields are not modelled by this walk" }`, or a reflection test over the three types asserting `!f.Anonymous` — converts a future red-tree surprise into a compile-time-ish stop. Non-blocking.

Two smaller notes, both non-blocking:

- **Case sensitivity is documented and ungraded** (mutant M1 below is ALIVE). `schema_strict.go` asserts "exact and case-sensitive, marginally stricter than yaml.v3's own field matching" and nothing observes it. Grading it costs one row in the existing table: `Name_Group: 1`. I confirmed such a leg does fail on M1 and pass on HEAD.
- **A wrong number in a comment.** `strict7014MinContainerEntries`'s comment says the tree yields "~445 (329 source_patterns + 78 relationship_rules + **38** file_conventions)". `38` is the audit's *file* count; the entry count is **155**, and the real total is **562**, which is what the test itself logs. The floor is fine; the arithmetic beside it is not.

### (c) The boundary is pinned strongly — five is enough, because the guard is attached to *types*, not paths

I did not trust that structurally; I ran **all fourteen** misspelled top-level containers the audit named (`orms_database`, `orms`, `database_libraries`, `orms_and_database`, `template_engines`, `orm_db`, `frameworks_and_dialects`, `orm_database`, `database`, `orms_databases`, `framework_templates`, `mobile`, `framework`, `queues`), each carrying `name`/`category`/`github_stars_2025`/`detection.import_markers`/`detection.package_json_deps`/`detection_signals.package_json_deps`, alongside a full valid executable half. **All 14 load, 0 failures.** So do: an executable container *nested under* a misspelled top-level container with a bogus key inside it (correctly untouched — it never reaches the guarded types), and the singular near-misses `source_pattern:` / `relationship_rule:`.

### (d) 247 vs 254 — **both are right; they measure different populations**, and the gap is exactly 7 named files

- **247** = files carrying an unmodelled **top-level key**. I get exactly 247, matching the PR. The 14 spellings sum to 247 with no file carrying two of them.
- **254** = files with **no top-level `frameworks:` key at all** — the audit's §2b figure, which is also exactly right (520 − 266).
- The audit's §4 table then labels 254 as "top-level allowlist: 254 red". **That label is the error**: an allowlist only reddens a file that has an unmodelled key, and **7 files have neither `frameworks:` nor any unmodelled top-level key** — they are pure `source_patterns`/`file_conventions` files:

```
go/frameworks/{echo,gin,chi,fiber,gorm}.yaml
ruby/frameworks/{sidekiq,rspec}.yaml
```

**247 is the number for "cost of extending the guard to the top level"** — the PR's figure — and #7013 should carry it, not 254.

### (e) Claim 5 — the day-one catch, and whether #6393's guard was doing its job

Confirmed the mechanism: pre-fix the fixture decoded to `RelationshipRules = [{Pattern: "implements (\w+)", Relationship: ""}]`, and `isZeroRule` checks **`len(r.RelationshipRules) == 0`**, so the leg passed on list length while the entry it named was half-empty.

Was it passing for the wrong reason? **The assertion was true, the fixture's label was not.** The leg pins "a file carrying relationship_rules is not reported inert", which held with an empty `Relationship`; the one-line fix does not weaken or strengthen it, it just makes the fixture mean what its name says. So **#6393's guard was doing its job** — it detects the zero-value `FrameworkRule`, which is what it promised. What the incident exposes is a different, pre-existing gap in `isZeroRule`: it cannot tell a real rule from a *present but garbage* one. This PR closes that gap for exactly the three containers, which is a nice second-order argument for the scope. Worth a line in the PR body; not a change request.

### Mutants (mine, 6 scored; all in `.../review-7019-k4v8t/mut`, each edit asserted at exactly 1 occurrence and printed `APPLIED`)

| # | mutant | result |
|---|---|---|
| M1 | `allowed[key]` → `allowed[strings.ToLower(key)]` (accept case variants) | **ALIVE — distinguishable and ungraded.** Killed by one extra row (`Name_Group: 1`) in `Test7014_UnknownKeyInsideExecutableContainerFailsLoad`; ~6 lines. Priced, not deferred. |
| M2 | `strings.Split(f.Tag.Get("yaml"), ",")[0]` → `f.Tag.Get("yaml")` | **Equivalent under the current suite** — no field in the three structs carries `omitempty` or any tag option, so no input distinguishes it. Grading it requires a struct change, i.e. it is graded by construction the day one is added. |
| M3 | untagged-field fallback `name = ToLower(f.Name)` → `continue` | **Equivalent** — every field in all three structs is tagged. |
| M4 | `SourcePattern.UnmarshalYAML` logs the error and **continues decoding** (no-op mode 5: detect-but-do-not-act) | **DEAD** — `Test7014_UnknownKeyInsideExecutableContainerFailsLoad/{source_patterns_typo, unmodelled_metadata_key_inside_source_patterns}`. This is the mode the author's own list did not name, and it *is* graded, because the legs assert absence from `rules` **and** `rep.Loaded`. |
| M5 | loader nils the three containers after unmarshal (guard off the real decode path) | **DEAD** — entry floor: `produced only 0 executable-container entries ... want >= 350`. Independently reproduces the author's mode-2 kill. |
| M6 | loader path filter `len(parts) != 3` → `!= 4` (inspect nothing) | **DEAD** — candidate floor: `only 0 rule files found, want >= 517`. |

### Verdict

**APPROVE.** The scope is the right one and it is *pinned* rather than merely stated; 0-red reproduces exactly; the test is non-vacuous in modes 1, 2, 3, 4 and 5, verified by mutation rather than by reading; the rule set is byte-identical before and after, so nothing downstream can have shifted. The only real production risk of a guard like this — a false positive taking a whole rule file out of the engine — is bounded here, because every production caller (`cmd/grafel/index.go:811`, `internal/daemon/extract/subproc.go:122`, `internal/extractors/incremental.go:358`) loads the **embedded** FS only, and `Test7014_AllEmbeddedRuleFilesStillLoad` reddens CI before any such file could ship.

Follow-ups, none blocking: guard `strictContainerFields` against embedded/inline fields (the one place the "cannot drift" claim is false); grade the documented case sensitivity (M1); fix `445`/`38` → `562`/`155` in the floor comment; note the `<<:` merge-key narrowing; correct 254 → **247** where #7013 quotes the cost of widening.

cajasmota:
## Coordinator mutant on `598490d2f` — **CX-1 DEAD**, and one design fact worth stating in the body

Own detached worktree at `/private/tmp/.../coord-mut-7019-m8p/wt` (absolute). Restored by `cp` from a SHA-256-verified backup (`2f43349d…` both ways); `git diff HEAD` and `git status --porcelain` empty afterwards.

Twelve mutants have been scored between you and the reviewer, all on the **decoder**: strictness removed per container, made vacuous, extended to the top level, taken off the decode path, made case-insensitive, tag-split, untagged-fallback, inspect-nothing, detect-but-don't-act. That set is thorough.

I went one layer up instead, because this session's most repeated defect is a mechanism graded on one path while a sibling path stays open.

### First: there is no sibling path

`loader.go` exposes four entry points (`LoadAllRules`, `LoadAllRulesReport`, `LoadAllRulesFromFS`, `LoadAllRulesFromFSReport`), but they all funnel to a **single** `yaml.Unmarshal(data, &rule)` at `loader.go:142`, and the only non-test callers are the two wrappers at `:27` and `:36`. The other `yaml.Unmarshal` calls in `internal/engine/` parse **user source files** — docker-compose, k8s, openapi — not rule files. **One decode point, and the guard is on it.** Worth stating in the body; it is the question a reader of a new decoder will ask.

### CX-1 — a typo in an executable container of a REAL shipped rule file

Inserted `nmae_group: 1` (a typo of `name_group`) as a sibling inside the first `source_patterns[]` entry of `internal/engine/rules/go/frameworks/gin.yaml`. Occurrence asserted 1.

**Result: DEAD**, by two tests — and the composition matters:

- `Test7014_AllEmbeddedRuleFilesStillLoad` — yours.
- **`TestRuleInventory_EveryEmbeddedRuleFileParses` — pre-existing.**

That second one is the reassuring part, and it answers the question I actually set out to ask.

### The design fact: load failures are non-fatal, so this guard's failure mode is "the whole file vanishes"

`loader.go:142` appends to `report.Failures` and `return nil` — a rejected file is **skipped**, its rules removed from the engine, and loading continues. The comment records that as deliberate and long-standing (#6341, cf. #6330), and notes failures are no longer discarded.

So the practical effect of strictness is a **change in blast radius, not just in detection**: before this PR a typo in an executable container silently ignored **one key**; after it, the same typo silently drops **every rule in that file**. That is a much larger consequence for the same input.

It is the right trade — a rule file that cannot be decoded as written should not half-load, which is exactly the `isZeroRule` gap you identified — and **the pre-existing inventory test means a shipped rule file cannot reach users in that state**. But the sentence "0-red today, so this is free" is only true because CI catches it; the runtime behaviour is not a hard failure. **Say that in the body.** A reader who assumes strict decoding means "loud error at startup" would be wrong.

### What I did NOT establish

Whether any **user-supplied** rule path exists. `LoadAllRulesFromFS` takes an arbitrary `fs.FS`, which is the shape a user rule-pack feature would use, but its only non-test callers today are the two embedded wrappers. **If rules are embedded-only, the non-fatal path is fully covered by CI and there is nothing further to do.** If a user rule-pack path is ever added, this becomes a real hazard and the non-fatal decision should be revisited then — not now.

I am not filing that as an issue: it is a conditional about a feature that does not exist, and the board does not need a speculative row.

**Merging on green.**

PR #7020, opened 2026-09-08
fix(#6916): Tier A — 8 `Config` rules named their entity after the whole regex match, so a file with two routers had one node

cajasmota:
Tier A of #6916, all eight sites, following the Tier B (aws_cdk) template proven in #7015. **29 of the 39 sites remain after this** — `Refs`, not `Closes`.

| rule site | construction | was named | is named |
|---|---|---|---|
| `go/frameworks/chi.yaml:39` | `chi.NewRouter()` | `"chi.NewRouter()"` | `r` |
| `go/frameworks/echo.yaml:33` | `echo.New()` | `"echo.New()"` | `e` |
| `go/frameworks/fiber.yaml:33` | `fiber.New()` | `"fiber.New()"` | `app` |
| `go/frameworks/gin.yaml:33` | `gin.Default()` / `gin.New()` | `"gin.Default()"` | `r` |
| `go/frameworks/gorm.yaml:15` | `gorm.Open(...)` | `"gorm.Open("` | `db` |
| `rust/frameworks/axum.yaml:82` | `Router::new()` | `"Router::new()"` | `app` |
| `csharp/.../asp_net_core.yaml:90` | `WebApplication.CreateBuilder(...)` | `"WebApplication.CreateBuilder("` | `builder` |
| `csharp/.../net_maui.yaml:124` | `MauiApp.CreateBuilder()` | `"MauiApp.CreateBuilder("` | `builder` |

## Why this is the #6914 anchor

All eight used `name_group: 0`, so the Name was the whole matched text. Combined with `scope: file`, a constant Name meant every construction in a file collapsed onto **one** node — chi's 50 `chi.NewRouter()` calls measured **30** `Config` entities, actix's sibling shape **83 → 73**. There was nothing for a per-router `Middleware REGISTERED_ON <router>` edge to point at, and adding the edge rule would not have created one. Naming the variable is what makes the anchor exist.

`TestIssue6916_TierAEachConstructionGetsItsOwnEntity` pins that on the six sites where a second construction in one file is idiomatic. The two C# sites are excluded on purpose and the test says why: a `WebApplication`/`MauiApp` builder is the single entry point of its file, so a second one there would be a fixture shaped to the test rather than to the language.

## All three Tier B cases, carried per LANGUAGE

Tier B measured its Python mutant ALIVE after the JS axis was fully graded. That repeated here: **56 mutants scored across the 8 sites**, and both ALIVE families were language-shaped rather than tier-shaped.

**1 — the anchor.** Tier B's blocker was that an unanchored capture is not "the assignment target" but "whatever token sits left of the `=`". Python's reachable spelling was a multi-target assignment; in Go, Rust and C# it is a **commented-out example line**, which every real codebase carries. Unanchored, `// r := chi.NewRouter()` mints `Config:r` in a file that builds no router. Each leg's positive control is the same file with the comment marker removed.

**2 — the type annotation.** Go `var r *chi.Mux =`, Rust `let app: Router =`, C# `WebApplicationBuilder builder =` — three different slots in three different patterns, so three fixtures. Measured and now recorded in the rule comments rather than assumed from Tier B: all three put the type *after* the declarator, so deleting the group makes an annotated declaration mint **nothing**, it does not capture the type the way the TypeScript sibling did.

**3 — over-firing, and this is the one that found a live mutant on seven of the eight sites.** Making the qualifier optional (`(?:chi\.)?NewRouter`, `(?:WebApplication\.)?CreateBuilder`) survived every fixture in the first round, because those spell the negative with a *different* qualifier — `router.NewRouter()` — which such a mutant still refuses. The input that separates them is the **bare call**: `r := NewRouter()`, `db, err := Open(dsn)`, `var builder = CreateBuilder(args)`. That is what a package's own constructor call looks like from inside that package, in any repo, with no framework anywhere — Tier B's `const app = App()` in React, in three more languages. `TestIssue6916_TierABareConstructorCallMintsNoConfig` derives each leg from that site's own control by deleting exactly the qualifier, and asserts the derivation landed.

Stating the interaction #7015 asked for: **making the name useful also makes an over-firing hit harder to spot.** A junk entity called `gin.Default()` is visibly a marker; one called `r` reads as a real router. These land in `Config`, which #7013 measures at **60.5% polluted**.

## gorm is not a copy of the template, and the reason is measured

`gorm.Open` returns `(*gorm.DB, error)`, so `db, err := gorm.Open(...)` is *the* idiom. The Tier B Python pattern ported verbatim — unanchored, no comma list — mints **`Config:err`**: the error value, named as though it were the database. Identifier-shaped, so it passes every other assertion in the file. That is the Tier B blocker (#6369) in Go, and it is why the pattern carries `(?:\s*,\s*[\w.]+)*`: Go's multi-return is *positional*, so the first name is the first return and the list can be read safely — which Python's `app, env = App(), "prod"` could not.

The two guards are graded by **different** inputs and neither masks the other: the comma list by that fixture, the anchor by the commented-out line. Removing the anchor alone still names `db`. The rule comment and the test comment both say exactly that now — an earlier draft of both claimed the anchor alone produced `err`, which measurement disproved.

## Fixed in passing

The old patterns were unanchored substring matches, so axum's `Router\s*::\s*new` fired on **`MyRouter::new()`** — a `Config` minted in a crate with no axum. Pinned as `axum`'s over-firing leg.

## The deliberate cost, and the digest re-pin

An **unassigned** construction now mints nothing. `TestIssue6916_TierAUnassignedConstructionMintsNoConfig` pins it on all eight so it is a decision on the record. #6916 lifted the hold on this ground already: six unassigned shapes were measured on the Tier B arm and every one lost only a marker with no edge to anchor.

It costs exactly one entity on the `cmd/grafel` semantic-digest fixture, whose `rs/routes.rs` returns `Router::new().route(...)` as a bare expression. Re-pinned as `gateOffDigest6916`, with the delta **enumerated member by member from dumped before/after graphs, not derived from the counts**:

```
-E Config "Router::new()" rs/routes.rs      the marker itself; no relationship
                                            had it as an endpoint, before or after
-R Module:span6918 -[CONTAINS]-> that entity   the only edge incident on it
```

141/283 → 140/282. Nothing positioned moved.

**One thing to flag for the rest of the tier:** that fixture is the repo's only axum sample and it is the *unassigned* form. `Router::new().route(...)` as a function's tail expression is idiomatic axum, so axum is the site of the eight where the assignment idiom may not be the dominant one. Nothing that ever bound is lost either way — the marker had no edges — but if a per-router anchor is wanted for axum specifically, a second shape (the `-> Router` function name) is likely needed, and that is not this PR.

## Verification

Isolated `GRAFEL_HOME` + `GRAFEL_DAEMON_ROOT` + `HOME` triple, fresh per run. macOS/APFS only (#6966); no Linux or Windows figure is claimed here.

- `go vet` + `gofmt -l` clean on both touched packages.
- `./internal/engine/` full suite green (`-count=1`).
- `./cmd/grafel/` full suite green (`-count=1`).
- `scripts/quality/run.sh --ratchet` — **38 gated fixtures held their baseline**, none moved. `go-chi-mini` builds `r := chi.NewRouter()` and its recall is unchanged; no golden `expected.json` pins any of these eight marker Names (the three pinned marker names on #6916 are all sinatra, i.e. Tier C).
- **Every leg of the new file fails on the parent commit** — 8/8 sites on the naming test, and the commented-out and bare-call negatives fail there too, which is what establishes that the over-firing they assert against is real and pre-existing rather than hypothetical.
- **56 mutants, all DEAD**, scored per site with a green baseline before each and the file restored from a shasum-verified backup after each (no `git checkout --`). Restoration byte-verified at the end of the sweep. The set: `name_group: 1 → 0`, widen the capture to `(.+)`, delete the anchor, drop the constructor qualifier, delete the type/declarator group, dotted capture `([\w.]+)`, drop the `var` declarator, plus gorm's comma list and axum's `let`. Two ALIVE families (the bare call on 7 sites, axum's `let` on 1) were graded with new fixtures and the **whole set re-scored afterwards**, since a Tier B mutant went ALIVE under its author's own fix.

Refs #6916, #6914, #7013, #6906, #6369.

### DIFF
--- a/cmd/grafel/custom_extractor_spans_6118_test.go
+++ b/cmd/grafel/custom_extractor_spans_6118_test.go
@@ -734,6 +734,37 @@ const (
 	// components, the two /orders endpoints, the six C# hierarchy edges and the
 	// language histogram above are all unchanged.
 	gateOffDigest6809 = "acfd2bc242dc7eadee89bf94de824776d5e8fc8c60078397e569d86ba3ebeb4d"
+	// gateOffDigest6916 is the current pin. #6916 Tier A narrowed eight `Config`
+	// source_patterns from `name_group: 0` — the entity Name was the whole regex
+	// match — to the ASSIGNED VARIABLE. One of the eight is axum's
+	// `Router::new()`, and this fixture's rs/routes.rs builds its router as an
+	// unassigned expression:
+	//
+	//	pub fn app() -> Router {
+	//	    Router::new().route("/items", get(list_items))
+	//	}
+	//
+	// With no variable there is no name to give the entity, so it is no longer
+	// minted. The delta is 141/283 -> 140/282, enumerated member by member (the
+	// gate-OFF entity+relationship sets dumped before and after and diffed; NOT
+	// derived from the counts):
+	//
+	//	-E Config "Router::new()" rs/routes.rs
+	//	     the marker node itself. Its Name was the matched text, so nothing
+	//	     could ever reference it: no relationship in this fixture had it as
+	//	     an endpoint, before or after.
+	//	-R Module:span6918 -[CONTAINS]-> Config:"Router::new()"
+	//	     the module containment edge carrying the entity above; it goes with
+	//	     it, and it was the ONLY edge incident on it.
+	//
+	// That is the whole delta: a marker with no referent plus its own
+	// containment. Every positioned entity, both /orders endpoints, the six C#
+	// hierarchy edges and the language histogram are unchanged. The loss is the
+	// deliberate cost recorded on #6916 — six unassigned shapes were measured on
+	// the Tier B arm and every one lost only a marker with no edge to anchor —
+	// and it is pinned in internal/engine by
+	// TestIssue6916_TierAUnassignedConstructionMintsNoConfig.
+	gateOffDigest6916 = "9415edfea35f8254755f3b0e8505b01e43706918f201bc473d6c5fd0283537a5"
 )
 
 func TestCustomExtractorGateOffGraphIsUnchanged(t *testing.T) {
@@ -741,8 +772,16 @@ func TestCustomExtractorGateOffGraphIsUnchanged(t *testing.T) {
 	t.Setenv("GRAFEL_INPROC_CUSTOM_EXTRACTORS", "")
 	off := persistAndReload(t, runIndexerOn(t, fixture, "span6118", nil))
 	got := semanticDigest6118(off)
-	if got == gateOffDigest6809 {
+	if got == gateOffDigest6916 {
 		return
+	}
+	if got == gateOffDigest6809 {
+		t.Fatalf("gate-OFF graph reverted to the pre-#6916 baseline — an axum " +
+			"`Router::new()` with no assignment is minting a `Config` entity named " +
+			"after the whole regex match again. rs/routes.rs returns the router as a " +
+			"bare expression, so there is no variable to name it after; check that " +
+			"rust/frameworks/axum.yaml still anchors its Config pattern on a `let` " +
+			"binding and still uses name_group 1")
 	}
 	if got == gateOffDigest6862 {
 		t.Fatalf("gate-OFF graph reverted to the pre-#6809 baseline — a relationship_rule " +
@@ -793,8 +832,8 @@ func TestCustomExtractorGateOffGraphIsUnchanged(t *testing.T) {
 		t.Fatalf("gate-OFF graph reverted to the pre-fix 2f0175dfc baseline — the #6118 " +
 			"span donation is no longer reaching the default path")
 	}
-	t.Fatalf("gate-OFF graph changed against ALL pinned digests\n got  %s\n post-#6809 %s\n post-#6862 %s\n post-#6742 %s\n post-#6601 %s\n post-#6485 %s\n post-#6152 %s\n post-#6138 %s\n post-#6118 %s\n 2f0175dfc %s\n"+
-		"(entities=%d relationships=%d)", got, gateOffDigest6809, gateOffDigest6862,
+	t.Fatalf("gate-OFF graph changed against ALL pinned digests\n got  %s\n post-#6916 %s\n post-#6809 %s\n post-#6862 %s\n post-#6742 %s\n post-#6601 %s\n post-#6485 %s\n post-#6152 %s\n post-#6138 %s\n post-#6118 %s\n 2f0175dfc %s\n"+
+		"(entities=%d relationships=%d)", got, gateOffDigest6916, gateOffDigest6809, gateOffDigest6862,
 		gateOffDigest6742, gateOffDigest6601, gateOffDigest6485, gateOffDigest6152,
 		gateOffDigest6138, gateOffDigest6118, gateOffDigest6118Base,
 		len(off.Entities), len(off.Relationships))
@@ -924,9 +963,16 @@ func TestCustomExtractorGateOffDeltaIsExactlyTheDocumentedSpanGain(t *testing.T)
 	// `Response` stub plus its Module:_external CONTAINS edge, which existed
 	// only to terminate that third one. Nothing that came out of a file was
 	// removed, which is why the file-component positions below are untouched.
+	//
+	// Then #6916 Tier A narrowed axum's `Config` pattern to the assigned variable
+	// (141/283 -> 140/282). rs/routes.rs builds its router as an unassigned
+	// expression, so the marker entity `Config:"Router::new()"` and the single
+	// Module CONTAINS edge that carried it are gone — enumerated member by member
+	// in the gateOffDigest6916 comment. Nothing that came out of a file was
+	// removed, so the file-component positions below are again untouched.
 	const (
-		wantEntities = 141
-		wantRels     = 283
+		wantEntities = 140
+		wantRels     = 282
 	)
 	if len(off.Entities) != wantEntities || len(off.Relationships) != wantRels {
 		t.Fatalf("gate-OFF graph size moved: entities=%d (want %d) relationships=%d (want %d) — "+
--- a/internal/engine/rules/csharp/frameworks/asp_net_core.yaml
+++ b/internal/engine/rules/csharp/frameworks/asp_net_core.yaml
@@ -87,9 +87,18 @@ source_patterns:
     scope: file
 
   # WebApplication.CreateBuilder(...) entry point
-  - pattern: "WebApplication\\.CreateBuilder\\s*\\("
+  # Named after the ASSIGNED VARIABLE (#6916): var builder = WebApplication.CreateBuilder(...)
+  # name_group 0 named this entity "WebApplication.CreateBuilder(" — a marker string with a trailing
+  # paren that no edge could bind to, one per FILE.
+  #
+  # The `(?m)^[ \\t]*` ANCHOR is load-bearing and graded: unanchored, the capture is
+  # just "the token left of the `=`", which is reachable from a commented-out
+  # example line. `(?:[\\w.<>\\[\\]]+\\s+)?` is the declarator/TYPE slot, so both
+  # `var builder =` and `WebApplicationBuilder builder =` name `builder` and neither
+  # names the type; `(?:\\w+\\.)?` keeps `this.builder =` naming `builder`.
+  - pattern: "(?m)^[ \\t]*(?:[\\w.<>\\[\\]]+\\s+)?(?:\\w+\\.)?(\\w+)\\s*=\\s*WebApplication\\.CreateBuilder\\s*\\("
     entity_type: Config
-    name_group: 0
+    name_group: 1
     scope: file
 
   # builder.Services.AddXxx(...) — DI service registration
--- a/internal/engine/rules/csharp/frameworks/net_maui.yaml
+++ b/internal/engine/rules/csharp/frameworks/net_maui.yaml
@@ -121,9 +121,18 @@ source_patterns:
     scope: file
 
   # MauiApp.CreateBuilder() — app entry point
-  - pattern: "MauiApp\\.CreateBuilder\\s*\\("
+  # Named after the ASSIGNED VARIABLE (#6916): var builder = MauiApp.CreateBuilder(...)
+  # name_group 0 named this entity "MauiApp.CreateBuilder(" — a marker string with a trailing
+  # paren that no edge could bind to, one per FILE.
+  #
+  # The `(?m)^[ \\t]*` ANCHOR is load-bearing and graded: unanchored, the capture is
+  # just "the token left of the `=`", which is reachable from a commented-out
+  # example line. `(?:[\\w.<>\\[\\]]+\\s+)?` is the declarator/TYPE slot, so both
+  # `var builder =` and `MauiAppBuilder builder =` name `builder` and neither
+  # names the type; `(?:\\w+\\.)?` keeps `this.builder =` naming `builder`.
+  - pattern: "(?m)^[ \\t]*(?:[\\w.<>\\[\\]]+\\s+)?(?:\\w+\\.)?(\\w+)\\s*=\\s*MauiApp\\.CreateBuilder\\s*\\("
     entity_type: Config
-    name_group: 0
+    name_group: 1
     scope: file
 
   # .UseMauiApp<TApp>() — application class registration
--- a/internal/engine/rules/go/frameworks/chi.yaml
+++ b/internal/engine/rules/go/frameworks/chi.yaml
@@ -36,9 +36,23 @@ source_patterns:
     scope: file
 
   # chi.NewRouter() init
-  - pattern: 'chi\.NewRouter\s*\(\s*\)'
+  # chi.NewRouter() init, named after the ASSIGNED VARIABLE (#6916): r := chi.NewRouter()
+  # name_group 0 named this entity "chi.NewRouter()" — a marker string, one per FILE, that no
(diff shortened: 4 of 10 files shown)


cajasmota:
## Coordinator probe on `9e75b4eeb` — **CY-1: the decision was right, and it is now verified rather than argued**

Own detached worktree at `/private/tmp/.../coord-mut-7020-r6t/wt` (absolute). Probe removed; `git diff HEAD` and `git status --porcelain` empty afterwards.

You scored 56 mutants and the review will score more, so rather than add a 57th on the same axes I went after **the PR's value claim** — the thing #6914 step 2 will build on.

That claim is *per-router multiplicity*: one marker per **file** becomes one per **variable**. Measured on the corpus as chi 50 constructions → 30 entities, actix 83 → 73.

### Where it is graded, and where it is not

`twoVars` fixtures exist for **6 of the 8 sites** — chi, echo, fiber, gin, gorm, axum. The two C# sites (`asp_net_core.yaml:90`, `net_maui.yaml:124`) fall back to the annotated single-construction fixture, which grades the **name** but not the **multiplicity**.

**That is a documented decision, not an oversight**, and I want to be clear I found the comment before I found the gap:

> *Two builders in one C# file is legal but not idiomatic — a WebApplication has one entry point per Program.cs — so the per-variable multiplicity is graded on the six sites where a second construction is realistic, not here.*

The reasoning is sound. But it is an **argument about idiom**, and the property it is standing in for is a fact about the regex. So I measured the fact.

### CY-1 — two constructions in one C# file, both sites

```
asp_net_core (Program.cs):      Config names = [first second]
net_maui     (MauiProgram.cs):  Config names = [firstApp secondApp]
```

**Both mint two distinct entities, correctly.** So the scoping decision is safe: it is not covering for a defect, and the C# patterns behave exactly as the six graded ones do.

### The ask — ~6 lines, non-blocking

Promote that probe into the two site entries as their `twoVars` fixture. The content above is the whole thing; the existing table already does the work.

The argument for taking it is not that the current state is wrong — I have just shown it is right — but that **an ungraded property drifts, and an argument about idiom does not fail a build.** If someone later tightens either C# pattern in a way that reintroduces per-file collapsing, six sites would catch it and these two would not. That asymmetry is the shape this PR family has been bitten by five times, most recently in your own first over-firing fixture, which used a *different qualifier* rather than an unqualified call and left the mutant ALIVE on 7 of 8 sites.

I would take it. I am not blocking on it — the review may surface something that matters more, and this can ride with whatever else comes back.

### On the rest of the round

Three things I want on the record because they are the reason this PR reads as trustworthy:

- **You found your own too-weak fixture and re-scored the whole 56-mutant set afterwards**, rather than patching the one case. A mutant verdict is only valid against the code it was scored on, and Tier B proved that the hard way when a previously-DEAD mutant went ALIVE under its own fix.
- **You corrected your own prose twice against measurement** — that removing the anchor alone still names `db` (the comma list masks it, so the two guards are graded by *different inputs*), and that in Go/Rust/C# deleting the type-annotation group mints *nothing* rather than capturing the type as TypeScript did. Both corrections are load-bearing and both went the direction that made your own claims smaller.
- **The digest delta was enumerated, not inferred** — `Config:"Router::new()"` plus the single `Module CONTAINS` edge that carried it, 141/283 → 140/282, with the golden ratchet unmoved. A re-pinned digest resting on an inferred cause is how a substitution ships.

**On axum**: invoking the report-don't-force clause was right. My read is that shipping it here is correct — the marker had no edges, so nothing that ever bound is lost, and a variable anchor is strictly better than a whole-match one even if the idiomatic unassigned builder chain needs a second shape later. The function-name anchor is a #6914 step-2 question, not a Tier A one.

cajasmota:
## Independent adversarial review — **APPROVE** (one ungraded over-fire filed as follow-up, not a blocker)

Measured in my own checkouts, macOS/APFS only (#6966), no Linux/Windows figure claimed:
head `9e75b4eeb` at `.../scratchpad/review-7020-zq9/wt`, parent `353753a31` at `.../review-7020-zq9/parent`,
mutants scored in a third tree `.../review-7020-zq9/mut` so no verification tree was ever mutated.
Isolated `GRAFEL_HOME` + `GRAFEL_DAEMON_ROOT` (+ isolated `HOME` for the `cmd/grafel` leg), fresh per run.

### (a) The digest delta IS exactly the one entity claimed — reproduced, not accepted

I did not take the 141/283 → 140/282 total. I added a throwaway dump test to **both** trees that writes the
sorted gate-OFF entity and relationship sets, ran it on parent and on head with separate fresh `GRAFEL_HOME`s,
and diffed the two files:

```
1d0
< E	Config	Router::new()	rs/routes.rs
280d278
< R	5692ad2a437e629d	CONTAINS	e45937144e40a72b
```

**Two deletions, zero additions, zero modifications** — so it is a set difference of one member on each side,
not a net of two changes (#6973's failure mode). Incidence confirmed: `e45937144e40a72b` (the removed Config)
occurs **exactly once** in the parent dump and that once is the edge above; the `from` id `5692ad2a437e629d`
is the `span6918` module, whose CONTAINS count goes 103 → 102. Enumeration verified member by member.

`./cmd/grafel/` full suite green (`-count=1`, 160.9s, exit 0), so `gateOffDigest6916` is the digest the code
actually produces. **Golden ratchet green independently**: `scripts/quality/run.sh --ratchet` exit **0** —
38 gated fixtures held, 29 at 100%. The ratchet fails on a *rise* as well as a drop, so exit 0 (not the
printed line) is the evidence the ceiling did not move.

### The "fails on parent" claim — exact, not a spot-check

I copied the new test file onto `353753a31` unchanged and ran it: **63 subtests, 63 FAIL, 0 PASS**, all
11 top-level tests red. Same file on head: **63 PASS, 0 FAIL**. Every leg pins the change; none passes
regardless.

### (b) The over-firing negatives DO grade all eight sites

`TierABareConstructorCallMintsNoConfig` covers 7 (axum excluded with a stated reason — stripping its
qualifier leaves `::new()`, not Rust) and axum's leg is the `MyRouter::new()` substring instead. Both
directions fail on parent, so the over-firing they assert against is real and pre-existing. I probed the
neighbours on that axis against the shipped regexes and then re-confirmed the interesting ones through the
real detector:

| neighbour | result |
|---|---|
| `chi2.NewRouter()` (differently-qualified pkg) | no match ✅ |
| `myhelper.NewRouter()` | no match ✅ |
| aliased import `chi2 "…/chi/v5"` → `chi2.NewRouter()` | mints nothing (recall miss, **not a regression** — parent's `chi\.` never matched it either) |
| `x.chi.NewRouter()`, `a.b.router = chi.NewRouter()` | no match |
| `let app = my::Router::new()` | no match ✅ |
| `db, err := Open(dsn)` bare | no match ✅ |

The **`MyRouter::new()` bonus fix is real**: on parent that fixture emits `Config:Router::new()` in a crate
with no axum (parent failure text confirms it). It is now refused. Note the fix comes from the `^[ \t]*let`
anchor, not from a qualifier — worth saying out loud, because it means *dropping axum from this PR would
leave that live #7013 over-fire in place*.

### (c) gorm's comma-list deviation is safe — I tested the argument rather than reading it

| input | group 1 |
|---|---|
| `db, err := gorm.Open(dsn)` | `db` ✅ |
| `db, _ := gorm.Open(dsn)` | `db` ✅ |
| `db , err := gorm.Open(dsn)` (spaced) | `db` ✅ |
| `s.db, err = gorm.Open(dsn)` | `db` ✅ |
| `a, b, c := gorm.Open(dsn)` (3-value) | `a` — first return, correct by positionality |
| `err, db := gorm.Open(dsn)` (reversed) | `err` — **but not compilable Go**; `*gorm.DB` cannot bind to `err`. Not production-reachable. |
| `if db, err := gorm.Open(dsn); err != nil` | **no match** (recall miss, see below) |

The positional argument holds: the only way group 1 is the wrong value is a source file that does not
compile. Python's `app, env = App(), "prod"` had no such protection — the deviation is justified.

**The two guards are graded by different inputs, confirmed by construction.** Removing the anchor alone on
`\tdb, err := gorm.Open(dsn)` still yields `db` (leftmost match starts at `db`), so the anchor's only killing
input is the commented line — matches the author's corrected prose. Removing the **comma list alone** makes
that same line match *nothing at all* (the `,` cannot be absorbed by the type slot), so it is killed by the
naming test. Neither masks the other; this is not a mutually-masking pair.

The type-annotation self-correction also checks out per language: `var r *chi.Mux =`, `let app: Router =`,
`WebApplicationBuilder builder =` all keep the **variable**; deleting the slot makes the annotated form mint
**nothing** in all three (Go/Rust/C# put the type after the declarator). Verified by mutant M3/M5 below and
by direct probe.

### (d) axum SHOULD ship in this PR

Verdict: **ship it, and open the `-> Router` function-name shape as the follow-up** — do not drop it.

1. Nothing that could ever bind is lost: I proved the removed entity had **exactly one** incident edge and
   that edge was its own `Module CONTAINS`. #6914 step 2 needs a *name*; the unassigned form never had one.
2. Dropping axum keeps `Router\s*::\s*new` as an unanchored substring rule, i.e. it *preserves* a measured
   #7013 over-fire. Removing axum from this PR is strictly worse than landing it.
3. The residual cost is recall on the `fn app() -> Router { Router::new()… }` shape, which is exactly what
   the function-name shape would recover. That is additive to this pattern, not a rewrite of it.

### Mutants — 5 scored by me, all **DEAD**, all killed only at their own site

Green baseline before each, `go vet` clean, `-count=1`, full `./internal/engine/` suite per score. Each edit
asserted to land with an **exact occurrence count of 1** in exactly one of the eight files, and each restored
by `cp` from a shasum-verified backup with `git status --porcelain` empty after (never `git checkout --`).
Two of the five (M1, M4) are re-scores of families the author declared DEAD — both confirmed.

| # | site | mutation (permissive direction) | verdict | killed by |
|---|---|---|---|---|
| M1 | chi | delete `(?m)^[ \t]*` anchor on **one** site only | **DEAD** | `TierACommentedConstructionMintsNoConfig/chi` — and *only* chi, proving fixtures are site-scoped and no site's fixture silently covers another |
| M2 | gorm | move the comma list **before** the capture: `(?:[\w.]+\s*,\s*)*(\w+)` → names `err` | **DEAD** | 5 gorm legs incl. `TierAConfigIsNamedAfterItsVariable` |
| M3 | echo | widen the capture `(\w+)` → `(.+)` | **DEAD** | 7 echo legs |
| M4 | axum | make `let` optional: `let\s+` → `(?:let\s+)?` | **DEAD** | `TierAReassignmentMintsNoConfig/axum` only — that fixture is load-bearing and not redundant |
| M5 | asp_net_core | capture the receiver: `(?:\w+\.)?` → `(\w+\.)?` (group renumber) | **DEAD** | 7 asp legs incl. `TierAFieldTargetNamesTheFieldNotTheReceiver` |

Every mutant's kill set is confined to its own site — no cross-site leakage, which was the "make two sites
identical" seam I was asked to probe.

### One finding: `Config:if` — distinguishable and ungraded (NOT a blocker, NOT a regression)

The four Go router sites (`chi`, `echo`, `fiber`, `gin`) name the entity after a **keyword** when the
construction is an if/for/switch init statement. Confirmed through the real detector, not just the regex:

```
if r := chi.NewRouter(); r != nil {   =>  Config:if
if r := gin.New(); r != nil {         =>  Config:if
```

Cause: the optional type slot `(?:\s+[\w.*\[\]]+)?` is available even when `var` did not match, so it eats
` r` and the keyword becomes group 1. This is the same class the PR fixes (`Config:err`) wearing a keyword.

Why it does not block:
- **Not a regression** — parent minted `Config:"chi.NewRouter()"` on the identical line.
- Low reachability: none of `chi.NewRouter` / `echo.New` / `fiber.New` / `gin.Default` returns an error, so
  an if-init around them is contrived. gorm — where `if db, err := gorm.Open(dsn); err != nil {` *is* the
  idiom — is **immune** (I checked: it mints nothing there, a recall miss rather than a wrong name).
- `if` is not identifier-plausible the way `err` was, so it does not read as a real router.

Priced: RE2 has no lookahead, so the clean fix is splitting each Go pattern into two `source_patterns`
entries — a `var`-form (declarator + type slot) and a short-form (no type slot) — roughly 2 added lines per
Go rule file plus one forbidden-name leg. Worth a Tier-A follow-up issue together with the two recall misses
this review found and the record should carry:
`if db, err := gorm.Open(dsn); err != nil` (gorm, mints nothing) and C# `private static readonly x = WebApplication.CreateBuilder(args)`
(only one token is allowed before the name).

Also noted, harmless but asymmetric and undocumented for the Go sites: a plain rebinding `r = chi.NewRouter()`
**does** mint on Go while the same shape is deliberately refused on axum (`let` required, pinned by
`TierAReassignmentMintsNoConfig`). The axum choice is argued in the test; the Go choice is not stated
anywhere. A sentence in the rule comments would close it.

### Verification run here

- `go vet ./internal/engine/` clean; `./internal/engine/` green (`-count=1`, 50.3s, exit 0).
- `./cmd/grafel/` green (`-count=1`, 160.9s, exit 0) under a fully isolated `HOME`/`GRAFEL_HOME`/`GRAFEL_DAEMON_ROOT`.
- `scripts/quality/run.sh --ratchet` exit 0, 38 gated fixtures held.
- Parent 353753a31 with the new test file: 63/63 subtests FAIL. Head: 63/63 PASS.
- All mutation-tree files byte-restored (shasum-matched) and `git status --porcelain` empty; my throwaway
  dump/probe test files deleted from both trees.

**APPROVE.** The digest move is exactly the one entity and its one edge, the over-firing negatives grade all
eight sites, gorm's comma list is safe on positional grounds I tested rather than read, and axum belongs in
this PR.

PR #7023, opened 2026-09-08
fix(#7022): a statement keyword reached the `Config` capture — three entries per Go site, so no type slot sits behind a bare `\w+`

### LINKED ISSUE #7022: Tier A residuals: `if r := chi.NewRouter(); …` mints `Config:if` on 4 Go sites (the #6916 defect class it was meant to remove), and the 2 C# sites have no twoVars fixture so per-variable multiplicity is ungraded there
Two residuals from #6916 Tier A (PR #7020, `87106f12c`). Both were found during review, both were judged non-blocking, and both live in the same eight rule sites — so whoever picks one should look at the other.

## 1. `if r := chi.NewRouter(); …` mints **`Config:if`** — the keyword becomes the capture

On `chi`, `echo`, `fiber` and `gin`, Go's if-with-initialiser form puts a keyword where the pattern expects a variable, and the optional type slot is available without `var`, so `if` is captured as the entity name. Confirmed through the real detector.

**Why it was not blocking:**
- **Not a regression.** The parent minted a marker on the same line; Tier A changed *which* junk name appears, not whether one does.
- **Low reachability.** None of those four constructors returns an error, so the if-init idiom has no reason to appear with them. `gorm` — where `if db, err := gorm.Open(...); err != nil` *is* the real idiom — is **immune**, because its comma-list handling takes a different path.

**Why it is worth fixing anyway:** an entity named `if` is precisely the defect class #6916 exists to remove — a node named after a syntax token rather than a thing. It lands in `Config`, which **#7013 measures at 60.5% polluted**. Shipping a fix for "named after the whole match" that can still produce "named `if`" leaves the issue's own premise half-standing.

**Priced by the reviewer** at splitting each Go pattern into var-form and short-form entries — four sites, so a real change rather than a one-liner. That is the reason it is filed rather than folded in.

## 2. The two C# sites have no `twoVars` fixture, so per-variable multiplicity is ungraded there

Tier A's value claim is that one marker per **file** becomes one per **variable** — the #6914 anchor. `twoVars` fixtures grade that on **6 of 8** sites; `asp_net_core.yaml:90` and `net_maui.yaml:124` fall back to a single-construction fixture that grades the **name** but not the multiplicity.

**This is a documented decision, not an oversight** — the comment argues two builders in one C# file is legal but not idiomatic, since a `WebApplication` has one entry point per `Program.cs`.

**And the decision is safe: I measured it.** Two constructions in one file, both sites, through the real detector:

```
asp_net_core (Program.cs):      Config names = [first second]
net_maui     (MauiProgram.cs):  Config names = [firstApp secondApp]
```

Both mint two distinct entities correctly. So this is **not covering for a defect** — it is a correct behaviour with no standing check.

**Why fix it anyway:** an argument about idiom does not fail a build. If either C# pattern is later tightened in a way that reintroduces per-file collapsing, six sites catch it and two do not. That asymmetry — one axis graded, its sibling left open — is the single most repeated defect shape in this PR family, and it caught Tier A's own first over-firing fixture, which used a differently-*qualified* call rather than an unqualified one and left the mutant ALIVE on 7 of 8 sites.

**~6 lines**: promote the probe above into the two site entries' `twoVars` slots. The table already does the work.

## Not established

- **Whether the `Config:if` shape exists in the other Tier A languages.** Only the four Go sites were checked. Rust's `if let` and C#'s pattern forms were not examined, and **one language family is not a population.**
- Whether any corpus repo actually contains the if-init form with these four constructors. Reachability was argued from the constructors' signatures, not measured against a corpus.

Refs #6916 (29 sites remain there), #6914, #7013, #7020.

cajasmota:
Closes #7022. Both Tier A residuals from #6916 / PR #7020, in the same eight rule sites.

**Updated after review (`209d60c9d`)** — the first cut had a regression and two ALIVE mutants. Both are fixed and every mutant has been re-scored against **this** head. Changes since `644b3f0e5` are called out inline.

## 1. A statement keyword reached the capture

The four Go `Config` patterns carried an optional `var` **and** an optional trailing type slot in **one** entry, so a keyword matched in the name slot and the real variable in the type slot. Measured at `87106f12c` through the real detector — this is **wider than the first version of this body claimed**:

```
if r := chi.NewRouter(); …     -> Config:if
if r = chi.NewRouter(); …      -> Config:if
if s.router = chi.NewRouter()  -> Config:if      (qualified LHS)
switch r = chi.NewRouter(); …  -> Config:switch
go r = chi.NewRouter()         -> Config:go
for r := chi.NewRouter(); …    -> Config:for
const r = chi.NewRouter()      -> Config:const
```

All seven mint **nothing** at head, on chi, echo, fiber and gin. `gorm` was already immune (its comma list takes a different path) and is deliberately left as a single entry — it is the control for the split.

### Three entries, not two

Each of the four sites is now three entries, carved so that **no entry offers a free type slot after a bare `\w+`** — that slot is what a keyword walks into:

1. **short form** — `r := chi.NewRouter()`, `s.router = chi.NewRouter()`. No type slot at all.
2. **`var` form** — `var mux *chi.Mux = chi.NewRouter()`. The type slot exists only behind a literal `var`, which no keyword can spell.
3. **grouped form** — `var (` … `r *chi.Mux = chi.NewRouter()` … `)`. **New since review**; see below.

### The blocker review caught: the split lost a working binding

Inside `var (…)` the keyword is on its own line, so entry 2's literal `var` is unreachable and entry 1 has no type slot. A grouped declaration **with an explicit type** matched the pre-split pattern and matched neither new entry:

| fixture | `87106f12c` | first cut | head |
|---|---|---|---|
| `var (\n\tr *chi.Mux = chi.NewRouter()\n)` | `[r]` | **`[]`** | `[r]` |
| `var (\n\tr chi.Router = chi.NewRouter()\n)` | `[r]` | **`[]`** | `[r]` |
| `var (\n\trs []*chi.Mux = …\n)` | `[rs]` | **`[]`** | `[rs]` |
| two declarators in one block | `[api r]` | **`[]`** | `[api r]` |

Losing a binding that worked is worse than the junk name the split removed. Entry 3 restores it.

The obvious fix — making `var` merely optional on entry 2, as review suggested — was **measured and rejected**: it reopens the hole for every plain-`=` keyword head above (review's own MU2 shows this, which is why entry 3 takes a different route). Entry 3 instead requires the type to be **package-qualified** (`[*\[\]]*chi\.\w+`), which no statement keyword can spell, so it is keyword-proof by construction rather than by absence of a counter-example.

**Measured residual, accepted and documented in the test:** a grouped declarator whose type is *unqualified* — `var ( r Mux = chi.NewRouter() )`, reachable only via a dot-import or a local alias — minted `[r]` at the parent and mints nothing now. That is the price of keeping keywords out. Every other real binding shape is preserved; the earlier claim that the union "covers every Go shape that is legal at statement level" is **withdrawn** and replaced by this measurement.

### Neighbours re-probed around the new boundary

Parent vs head, same probe, through the real detector: grouped typed (pointer / qualified / slice), grouped untyped, grouped multi-declarator, grouped unqualified, single-line `var` typed and untyped, short declaration, field assignment, `const` grouped and single, six keyword heads, three commented spellings, bare constructor, foreign package. Only the two rows above changed direction, and only in the intended one.

### Immunity of the other two language families (unchanged, independently reproduced)

`if let Some(app) = Router::new()` → `[]`, `if x { let app = Router::new(); }` → `[]` (rust's pattern requires a line-initial `let`); `if (x) builder = WebApplication.CreateBuilder(args)` → `[]`, same for MauiApp (C# has no if-initialiser). Recorded in the test's doc comment. Two of three families measured, not assumed.

## 2. The two C# sites had no `twoVars` fixture

Tier A's per-variable multiplicity claim was graded on 6 of 8 sites. The behaviour was already correct — `[first second]` and `[firstApp secondApp]` — but the axis had no direct pin. Both shapes are now in the sites' `twoVars` slots, and the two comments arguing the multiplicity was deliberately ungraded there are rewritten rather than left contradicting the fixtures.

**Claim corrected:** the earlier body said the new fixtures "are what kills" a per-file collapse. That does not follow — review showed MU5 is DEAD at `87106f12c` too, via five pre-existing legs. The fixtures add a **direct** pin on an axis that genuinely was ungraded on those two sites; they are not the only thing catching a collapse.

## 3. Two ALIVE mutants closed, one dead assertion removed

- **The anchor was graded on the short form only**, while all four files claimed it was graded on "BOTH" entries — a prose claim no test observed, inside a PR whose parent issue is about exactly that. The `commented` fixture now carries the short, the single-line `var` **and** the grouped spelling, so all three anchors are graded, and the control asserts all three names.
- **`ifInit` pinned only the `:=` spelling.** Renamed `keywordInit` and extended with the plain-`=` `if` and `switch` heads, so the legal sibling of the shape being fixed is graded too.
- **`ifInitWant` is gone.** Head mints nothing for these shapes and its check was a `t.Logf`, not an assertion. The fixture now asserts an empty `Config` set outright, backed by the site's own plain-assignment positive control.

## Mutants — 9 applied, 9 DEAD, all re-scored against `209d60c9d`

Separate detached checkout, green baseline (`./internal/engine/` `ok` 49.6s) immediately before, `-count=1`, `go vet` clean, each edit an exact-occurrence-count replacement (`assert count == 1`) against a **specific one of the three similar entries**, each restore by `cp` from a shasum-verified backup with `git status --porcelain` empty.

| # | Mutant | Verdict |
|---|---|---|
| MU1 | chi entry 1: restore the optional type slot (the original defect) | **DEAD** — `KeywordInitialiser…/chi.yaml:79` |
| MU2 | chi entry 2: `var\s+` → `(?:var\s+)?` (permissive; was ALIVE) | **DEAD** — `KeywordInitialiser…/chi.yaml:79` |
| MU3 | gin: fold all three entries back into the pre-split single entry | **DEAD** — `KeywordInitialiser…/gin.yaml:73` |
| MU4 | chi entry 2: drop the anchor (permissive; was ALIVE) | **DEAD** — `CommentedConstruction…/chi.yaml:79` |
| MU5 | both C# sites: capture the constant qualifier → per-file collapse | **DEAD** — incl. both new `twoVars` legs |
| MU6 | chi entry 3: drop the anchor (permissive) | **DEAD** — `CommentedConstruction…/chi.yaml:79` |
| MU7 | chi entry 3: delete it — i.e. re-introduce the regression | **DEAD** — `GroupedVarDeclaration…` + `CommentedConstruction…` |
| MU8 | chi entry 3: allow ANY type, not just a qualified one (permissive) | **DEAD** — `KeywordInitialiser…/chi.yaml:79` |
| MU9 | chi entry 3: match only the FIRST declarator of a block | **DEAD** — `GroupedVarDeclaration…` (the second declarator) |

MU1/MU2/MU8 score the **over-firing** direction, MU3/MU7 the **regression** direction, MU4/MU6 the anchors, MU9 the per-line grain. Every leg that grades an absence re-runs the site's own plain-assignment control, so a rule that simply stopped firing cannot pass.

## Verification — macOS/APFS (#6966)

- `./internal/engine/` — `ok` 53.1s, `-count=1`
- `./cmd/grafel/` — `ok` 167.2s, `-count=1`, under a **fresh** `GRAFEL_HOME` + `GRAFEL_DAEMON_ROOT`. The whole-graph digest pin did not move.
- `scripts/quality/run.sh --runs 1 --ratchet` — exit 0, *38 gated fixtures held their baseline (29 at 100% must-have recall)*, and `git status --porcelain` empty afterwards, so **no baseline moved** (not inferred from the summary line).
- `gofmt -l` clean, `go vet ./internal/engine/` clean.

Refs #6916, #6914, #7013, #7020.

### DIFF
--- a/internal/engine/rules/go/frameworks/chi.yaml
+++ b/internal/engine/rules/go/frameworks/chi.yaml
@@ -40,32 +40,53 @@ source_patterns:
   # edge could ever bind to. Naming the variable is also what makes a PER-ROUTER
   # anchor possible (#6914): a file building two routers now mints two entities.
   #
-  # TWO entries, not one (#7022). A SINGLE entry with an optional `var` AND an
-  # optional trailing type slot let `if r := chi.NewRouter(); …` match with the keyword `if`
-  # in the name slot and the real variable in the type slot — minting `Config:if`,
-  # a node named after a syntax token, which is the exact defect class #6916
-  # exists to remove. Confirmed through the real detector on all four Go sites;
-  # gorm was immune because its comma list takes a different path. Splitting the
-  # forms puts the type slot BEHIND a literal `var`, so nothing but a declared
-  # name can reach the capture.
+  # THREE entries, not one (#7022). A SINGLE entry with an optional `var` AND an
+  # optional trailing type slot let `if r := chi.NewRouter(); …` match with the keyword
+  # `if` in the name slot and the real variable in the type slot — minting
+  # `Config:if`, a node named after a syntax token, which is the exact defect
+  # class #6916 exists to remove. `for`/`switch`/`go` initialisers minted
+  # `Config:for` / `Config:switch` / `Config:go` the same way. All confirmed
+  # through the real detector; gorm was immune because its comma list takes a
+  # different path, and is deliberately left as one entry.
   #
-  # The `(?m)^[ \t]*` ANCHOR is load-bearing and graded on BOTH entries. Without it
-  # the capture is just "the token left of the `=`", which in Go is reached from
-  # inside comments and doc examples — `// r := chi.NewRouter()` would mint a
-  # plausible-looking `Config:` in a file that builds nothing at all. `(?:\w+\.)?`
-  # on the short form keeps `s.router = chi.NewRouter()` naming the field rather than the
-  # receiver.
+  # The three entries carve the Go declaration forms so that NO entry offers a
+  # free type slot after a bare `\w+` — that slot is what a keyword walks into:
+  #
+  #   1. short form   `r := chi.NewRouter()`, `s.router = chi.NewRouter()` — no type slot at all.
+  #   2. `var` form   `var mux *chi.Mux = chi.NewRouter()` — the type slot exists
+  #      only BEHIND a literal `var`, which no keyword can spell.
+  #   3. grouped form `var (` … `r *chi.Mux = chi.NewRouter()` … `)` — inside a
+  #      grouped declaration the `var` is on its own line, so entry 2 cannot
+  #      reach it and entry 1 has no type slot. This entry restores it (it
+  #      matched the pre-split pattern, and dropping it would have been a
+  #      REGRESSION), and it stays keyword-proof by requiring the type to be
+  #      chi-QUALIFIED: `if r = chi.NewRouter()` cannot match, because `r` is not
+  #      spelled `chi.Something`. The cost is a grouped declaration whose type is
+  #      a dot-imported or aliased name; that is narrower than the pre-split
+  #      pattern and is the deliberate price of keeping keywords out.
+  #
+  # The `(?m)^[ \t]*` ANCHOR is load-bearing on ALL THREE entries and graded on
+  # all three by the `commented` fixture, which carries the short, the `var` and
+  # the grouped spelling. Without it the capture is just "the token left of the
+  # `=`", which in Go is reached from inside comments and doc examples — a
+  # commented `r := chi.NewRouter()` would mint a plausible-looking `Config:` in a file
+  # that builds nothing at all. `(?:\w+\.)?` on the short form keeps
+  # `s.router = chi.NewRouter()` naming the field rather than the receiver.
+  #
+  # Measured: Go puts the type AFTER the name, so entries 2 and 3 are what let an
+  # explicitly typed declaration match at all — neither captures the TYPE the way
+  # the TypeScript sibling did (#6916 Tier B, review ask 2).
   - pattern: '(?m)^[ \t]*(?:\w+\.)?(\w+)\s*:?=\s*chi\.NewRouter\s*\(\s*\)'
     entity_type: Config
     name_group: 1
     scope: file
 
-  # Declaration form: `var mux *chi.Mux = chi.NewRouter()`. Measured: Go puts the
-  # type AFTER the name, so deleting this entry's type slot makes an explicitly
-  # typed declaration mint NOTHING — it does not capture the type the way the
-  # TypeScript sibling did (#6916 Tier B, review ask 2). The type slot lives here
-  # rather than on the short form so it cannot swallow a keyword (#7022).
   - pattern: '(?m)^[ \t]*var\s+(\w+)(?:\s+[\w.*\[\]]+)?\s*=\s*chi\.NewRouter\s*\(\s*\)'
+    entity_type: Config
+    name_group: 1
+    scope: file
+
+  - pattern: '(?m)^[ \t]*(\w+)\s+[*\[\]]*chi\.\w+\s*=\s*chi\.NewRouter\s*\(\s*\)'
     entity_type: Config
     name_group: 1
     scope: file
--- a/internal/engine/rules/go/frameworks/chi.yaml
+++ b/internal/engine/rules/go/frameworks/chi.yaml
@@ -35,22 +35,37 @@ source_patterns:
     name_group: 1
     scope: file
 
-  # chi.NewRouter() init
   # chi.NewRouter() init, named after the ASSIGNED VARIABLE (#6916): r := chi.NewRouter()
   # name_group 0 named this entity "chi.NewRouter()" — a marker string, one per FILE, that no
   # edge could ever bind to. Naming the variable is also what makes a PER-ROUTER
   # anchor possible (#6914): a file building two routers now mints two entities.
   #
-  # The `(?m)^[ \t]*` ANCHOR is load-bearing and graded. Without it the capture is
-  # just "the token left of the `=`", which in Go is reached from inside comments
-  # and doc examples — `// r := chi.NewRouter()` would mint a plausible-looking
-  # `Config:r` in a file that builds no router at all. `(?:\w+\.)?` keeps
-  # `s.router = chi.NewRouter()` naming `router` rather than `s`; the optional
-  # `var` + type group is what lets `var r *chi.Mux = chi.NewRouter()` match at all.
-  # Measured: Go puts the type AFTER the name, so deleting that group makes an
-  # explicitly typed declaration mint NOTHING — it does not capture the type the
-  # way the TypeScript sibling did (#6916 Tier B, review ask 2).
-  - pattern: '(?m)^[ \t]*(?:var\s+)?(?:\w+\.)?(\w+)(?:\s+[\w.*\[\]]+)?\s*:?=\s*chi\.NewRouter\s*\(\s*\)'
+  # TWO entries, not one (#7022). A SINGLE entry with an optional `var` AND an
+  # optional trailing type slot let `if r := chi.NewRouter(); …` match with the keyword `if`
+  # in the name slot and the real variable in the type slot — minting `Config:if`,
+  # a node named after a syntax token, which is the exact defect class #6916
+  # exists to remove. Confirmed through the real detector on all four Go sites;
+  # gorm was immune because its comma list takes a different path. Splitting the
+  # forms puts the type slot BEHIND a literal `var`, so nothing but a declared
+  # name can reach the capture.
+  #
+  # The `(?m)^[ \t]*` ANCHOR is load-bearing and graded on BOTH entries. Without it
+  # the capture is just "the token left of the `=`", which in Go is reached from
+  # inside comments and doc examples — `// r := chi.NewRouter()` would mint a
+  # plausible-looking `Config:` in a file that builds nothing at all. `(?:\w+\.)?`
+  # on the short form keeps `s.router = chi.NewRouter()` naming the field rather than the
+  # receiver.
+  - pattern: '(?m)^[ \t]*(?:\w+\.)?(\w+)\s*:?=\s*chi\.NewRouter\s*\(\s*\)'
+    entity_type: Config
+    name_group: 1
+    scope: file
+
+  # Declaration form: `var mux *chi.Mux = chi.NewRouter()`. Measured: Go puts the
+  # type AFTER the name, so deleting this entry's type slot makes an explicitly
+  # typed declaration mint NOTHING — it does not capture the type the way the
+  # TypeScript sibling did (#6916 Tier B, review ask 2). The type slot lives here
+  # rather than on the short form so it cannot swallow a keyword (#7022).
+  - pattern: '(?m)^[ \t]*var\s+(\w+)(?:\s+[\w.*\[\]]+)?\s*=\s*chi\.NewRouter\s*\(\s*\)'
     entity_type: Config
     name_group: 1
     scope: file
--- a/internal/engine/rules/go/frameworks/echo.yaml
+++ b/internal/engine/rules/go/frameworks/echo.yaml
@@ -34,32 +34,53 @@ source_patterns:
   # edge could ever bind to. Naming the variable is also what makes a PER-ROUTER
   # anchor possible (#6914): a file building two routers now mints two entities.
   #
-  # TWO entries, not one (#7022). A SINGLE entry with an optional `var` AND an
-  # optional trailing type slot let `if e := echo.New(); …` match with the keyword `if`
-  # in the name slot and the real variable in the type slot — minting `Config:if`,
-  # a node named after a syntax token, which is the exact defect class #6916
-  # exists to remove. Confirmed through the real detector on all four Go sites;
-  # gorm was immune because its comma list takes a different path. Splitting the
-  # forms puts the type slot BEHIND a literal `var`, so nothing but a declared
-  # name can reach the capture.
+  # THREE entries, not one (#7022). A SINGLE entry with an optional `var` AND an
+  # optional trailing type slot let `if e := echo.New(); …` match with the keyword
+  # `if` in the name slot and the real variable in the type slot — minting
+  # `Config:if`, a node named after a syntax token, which is the exact defect
+  # class #6916 exists to remove. `for`/`switch`/`go` initialisers minted
+  # `Config:for` / `Config:switch` / `Config:go` the same way. All confirmed
+  # through the real detector; gorm was immune because its comma list takes a
+  # different path, and is deliberately left as one entry.
   #
-  # The `(?m)^[ \t]*` ANCHOR is load-bearing and graded on BOTH entries. Without it
-  # the capture is just "the token left of the `=`", which in Go is reached from
-  # inside comments and doc examples — `// e := echo.New()` would mint a
-  # plausible-looking `Config:` in a file that builds nothing at all. `(?:\w+\.)?`
-  # on the short form keeps `s.engine = echo.New()` naming the field rather than the
-  # receiver.
+  # The three entries carve the Go declaration forms so that NO entry offers a
+  # free type slot after a bare `\w+` — that slot is what a keyword walks into:
+  #
+  #   1. short form   `e := echo.New()`, `s.engine = echo.New()` — no type slot at all.
+  #   2. `var` form   `var srv *echo.Echo = echo.New()` — the type slot exists
+  #      only BEHIND a literal `var`, which no keyword can spell.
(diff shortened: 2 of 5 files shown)


cajasmota:

Measured in a **separate detached worktree of `644b3f0e5`** at
`…/scratchpad/review-7023-q7v/wt`, with a second worktree of the parent `87106f12c` at
`…/scratchpad/review-7023-q7v/base` for before/after differencing. macOS/APFS only (#6966).
Green baseline before every score (`./internal/engine/` `ok` 54.1s), `-count=1`, `go vet` clean,
every mutant applied by an exact-occurrence-count replacement (assert `count == 1`) and every
restore verified by `shasum` against a backup + `git diff HEAD` empty (never `git checkout --`).

The fix itself is real and it works in **both** directions. Two things block it: a shape that now
falls **between** the two split entries and is a behaviour regression against `87106f12c`, and a
new prose claim in all four files that no test observes.

### (a) Is any Tier A mutant resurrected? — **No.**

The new union is a **subset** of the old one (short form = old minus the type slot; `var` form =
old restricted to a literal `var`), so no over-firing property can be widened. Confirmed through
the real detector — identical at parent and head: `chi.NewRouter()` bare `[]`, `chi2.NewRouter()`
`[]`, `// r := chi.NewRouter()` `[]`, `s.router = chi.NewRouter()` `[router]`,
`r, ctx := chi.NewRouter(), context.Background()` `[]`. `gorm.yaml` is untouched and still names
`db` for `db, err :=`, `db, _ :=`, `s.db, err =` and `var db *gorm.DB =`, and is still immune to
the if-initialiser (`[]`). Re-scored two Tier A families against **this** head: collapsing the
entries back (MU1) and reverting one of four files (MU3) both **DEAD**.

The fix is also **wider than the PR claims**: `for` and `switch` initialisers minted
`Config:for` / `Config:switch` at `87106f12c` and mint nothing now. Good, but unmentioned and
ungraded.

### (b) Does a form fall between the two entries? — **Yes. This is the blocker.**

A **grouped `var (…)` declaration with an explicit type** matched the old single entry and matches
**neither** new entry. Measured, same probe, both trees:

| fixture | `87106f12c` | `644b3f0e5` |
|---|---|---|
| `var (\n\tr *chi.Mux = chi.NewRouter()\n)` | `Config=[r]` | **`Config=[]`** |
| `var (\n\tr chi.Router = chi.NewRouter()\n)` | `Config=[r]` | **`Config=[]`** |
| `var (\n\te *echo.Echo = echo.New()\n)` | `Config=[e]` | **`Config=[]`** |
| `var (\n\tapp *fiber.App = fiber.New()\n)` | `Config=[app]` | **`Config=[]`** |
| `var (\n\tr *gin.Engine = gin.Default()\n)` | `Config=[r]` | **`Config=[]`** |
| `var (\n\tr = chi.NewRouter()\n)` (untyped) | `[r]` | `[r]` — unaffected |
| `var r *chi.Mux = chi.NewRouter()` (single-line) | `[r]` | `[r]` — unaffected |

Inside a `var (…)` block the `var` keyword is on its own line, so the `var` form's literal `var`
can never be reached, and the short form has no type slot any more. All four sites lose it; `gorm`
keeps it only because it was not split. The PR body's "the union still covers every Go shape that
is legal at statement level" is not true of a grouped declaration, and each file's new comment
advertises the type slot with the single-line example only. Either add `(?m)^[ \t]*(?:var\s+)?…`
to the *declaration* entry (it cannot swallow a keyword there — the `=` is plain, not `:=`, so
`if r := …` still cannot reach it) or state the loss explicitly and pin it. Corpus reachability is
zero *in the gated fixtures* — the `cmd/grafel` digest did not move — but a zero is
corpus-relative, not "cannot occur".

### (c) Do the rust/C# immunity findings hold? — **Yes, both.**

Reproduced through the real detector at head:
`if let Some(app) = Router::new()` → `[]`; `if x { let app = Router::new(); }` → `[]`, with the
positive control `let app = Router::new();` → `[app]`, so the absence is not vacuous.
`if (x) builder = WebApplication.CreateBuilder(args);` → `[]` and
`if (x) builder = MauiApp.CreateBuilder();` → `[]`, with `var builder = WebApplication.CreateBuilder(args);`
→ `[builder]`. Both recorded reasons are correct.

### (d) Do the new C# fixtures kill a per-file collapse? — **They fire, but the credit claim is not established.**

`[first second]` and `[firstApp secondApp]` reproduce, and MU5 (the PR's own M3: capture the
constant qualifier so the file collapses) **DEAD** on the new leg —
`"first" is missing. Config names emitted: [WebApplication]`. But the same mutant applied to
**`87106f12c`, which has no `twoVars` fixture for that site, is also DEAD** — on five pre-existing
legs (`ConfigIsNamedAfterItsVariable`, `AnnotatedDeclarationNamesTheVariable`,
`CommentedConstructionMintsNoConfig`, `NonFrameworkCallMintsNoConfig`,
`ChainedBuilderNamesTheAssignedVariable`). So "M3/M4 land on subtests that did not exist before
this PR, **so the new fixtures are what kills them**" does not follow: a mutant the parent already
kills does not price the new leg. The legs are still worth having (the multiplicity axis on those
two sites genuinely was ungraded), but the PR body should not claim they are what makes the mutant
die. If the axis is worth 30 lines, score a mutant that only they kill.

The rewritten comments no longer contradict the fixtures — that part is clean.

### Mutants — 5 applied, 3 DEAD, 2 ALIVE and distinguishable

| # | Mutant (exact-count replacement, 1 occurrence each) | Verdict |
|---|---|---|
| MU1 | chi short form: restore the optional type slot `(?:\s+[\w.*\[\]]+)?` | **DEAD** — `TierAIfInitialiserMintsNoKeywordConfig/…chi.yaml:58`: *minted `Config:"if"`* |
| MU2 | chi **`var` form**: `var\s+` → `(?:var\s+)?` (permissive) | **ALIVE** — whole `./internal/engine/` `ok` 47.6s |
| MU3 | gin only: fold both entries back into the pre-fix single entry | **DEAD** — `TierAIfInitialiserMintsNoKeywordConfig/…gin.yaml:52` |
| MU4 | chi **`var` form**: drop the anchor, `(?m)^[ \t]*var` → `\bvar` (permissive) | **ALIVE** — whole `./internal/engine/` `ok` 48.4s |
| MU5 | asp_net_core: capture the constant qualifier (per-file collapse) | **DEAD** at head — *and DEAD at `87106f12c` too*, see (d) |

**MU4 — distinguishable and ungraded, and it falsifies new prose.** All four files now say *"The
`(?m)^[ \t]*` ANCHOR is load-bearing and graded on **BOTH** entries."* It is graded on the short
form only. With the anchor removed from the declaration entry alone, the entire engine suite stays
green while:

```
// var r *chi.Mux = chi.NewRouter()          -> Config=[r]      (head: [])
//	var mux *chi.Mux = chi.NewRouter()        -> Config=[mux]    (head: [])   // doc example
// r := chi.NewRouter()                      -> Config=[]       (the graded one)
```

That is exactly the defect the anchor exists to prevent — a plausible `Config` in a file that
builds no router — and it is reachable from any doc comment showing a typed declaration, which is
the normal way that shape appears in a comment. Either grade it (a commented-out `var` declaration
in the existing `commented` fixture is a one-line addition) or delete the word "BOTH".

**MU2 — distinguishable, production-reachable, ungraded.** The new `ifInit` fixture pins only the
`:=` form. With `var` made optional on the declaration entry, the suite stays green while the
plain-`=` keyword forms come back:

```
if r = chi.NewRouter(); r != nil { }   -> Config=[if]      (head: [])
switch r = chi.NewRouter(); { }        -> Config=[switch]  (head: [])
go r = chi.NewRouter()                 -> Config=[go]      (head: [])
```

All three are legal Go with a pre-declared `r`, and `if x = f(); cond` is ordinary Go. The
regression this PR fixes has a plain-`=` twin that nothing grades. One extra `ifInit` variant using
`=` instead of `:=` closes MU2 and would have caught it.

### Two smaller notes

- **`ifInitWant` is not asserted.** Head mints **nothing** for the if-initialiser on all four Go
  sites, so `ifInitWant: "r"` is never satisfied and the "must still name the variable" direction is
  a `t.Logf`, not a check. The struct comment calls it "the variable that a correct rule may still
  name (empty where the shape mints nothing at all)" — the measured answer is "nothing at all", so
  the four sites should carry `ifInitWant: ""` and the field should be dropped, or the log should be
  a real assertion. As written it reads as a graded expectation that is neither met nor enforced.
  (The forbidden-name loop is vacuous on an empty result, but `assertControl6916` plus the separate
  positive control cover that — that part is fine.)
- The line references in the test doc comment (`chi.yaml:58`, `echo/fiber/gin.yaml:52`) all land on
  the short-form entry. Verified.

### Verification I ran independently

- `./internal/engine/` — `ok` 54.1s, `-count=1`, `go vet` clean.
- `./cmd/grafel/` — `ok` 193.8s, `-count=1`, under a **fresh** `GRAFEL_HOME` + `GRAFEL_DAEMON_ROOT`
  created for this run. The whole-graph digest pin did **not** move: confirmed.
- `scripts/quality/run.sh --runs 1 --ratchet` — *quality ratchet OK — 38 gated fixtures held their
  baseline (29 at 100% must-have recall)*. Independently confirmed **no baseline moved** by
  `git status --porcelain` being empty afterwards, not by the summary line (`--ratchet` prints it
  anyway on a ceiling drop).
- Both worktrees left clean (`git status --porcelain` empty, `git diff HEAD` empty) after the last
  restore.

### What I would need to flip to APPROVE

1. Close or explicitly document-and-pin the grouped-`var`-with-type gap in (b).
2. Fix the "graded on BOTH entries" claim — grade the declaration entry's anchor, or drop the word.
3. Add the plain-`=` if/switch variant that kills MU2, or say why it is out of scope.
4. Soften the (d) claim in the PR body, and either resolve `ifInitWant` to `""` or make it assert.

cajasmota:
## Coordinator mutant on `209d60c9d` — **CZ-1 DEAD**, and the round is right

Own detached worktree at `/private/tmp/.../coord-mut-7023-w2k/wt` (absolute). Green baseline immediately before. Restored by `cp` from a SHA-256-verified backup (`a1c56247…` both ways); `git diff HEAD` and `git status --porcelain` empty afterwards.

Fourteen mutants have been scored between you and the reviewer, all on **one site's** patterns. The structure is now **three entries × four files = twelve places**, so I went after the property none of those fourteen tests: **cross-site independence.**

### CZ-1 — delete `fiber`'s entry 3 only, leaving chi/echo/gin intact

Removed the four lines of fiber's grouped-typed entry (pattern + `entity_type` + `name_group` + `scope`), asserted against the exact line content first, `entity_type: Config` count 3 → 2 in that file.

**Result: DEAD**, and precisely attributed:

```
--- FAIL: TestIssue6916_TierACommentedConstructionMintsNoConfig/go/frameworks/fiber.yaml:73
--- FAIL: TestIssue6916_TierAGroupedVarDeclarationNamesEachVariable/go/frameworks/fiber.yaml:73
```

**Only the fiber subtests fail.** No fixture for chi, echo or gin covers for its neighbour, and both the positive property (grouped var names each variable) and the negative one (a commented construction mints nothing) are graded per site. That is the property I most wanted, because twelve near-identical regex entries across four files is exactly where one fixture starts standing in for four.

### On rejecting the reviewer's fix — you were right, and you proved it

This is the part of the round worth naming. The reviewer proposed `(?:var\s+)?` on the declaration entry, with a parenthetical that it "cannot swallow a keyword there — the `=` is plain, not `:=`". **You measured it instead of taking it, and it reopens `if r = ctor()` → `Config:if` — which is the reviewer's own MU2.** So the suggestion contradicted the reviewer's own finding two sections earlier.

Requiring a **package-qualified type** (`[*\[\]]*chi\.\w+`) is keyword-proof *by construction* rather than by enumeration, which is a stronger property than the one asked for — no keyword can spell `chi.Mux`. And it closes `if s.router = ctor()`, the qualified-LHS case the proposed route would have left open.

**An implementer disproving a reviewer's suggested fix with a measurement is the outcome the review loop exists for**, and it is the second time this round that measuring beat arguing.

### The residual is documented honestly

`var ( r Mux = ctor() )` — an unqualified type inside a grouped block — was `[r]` and is now `[]`. That needs the type dot-imported while the constructor stays qualified, which is an inconsistent import style, and you recorded it in the test rather than in a commit message. I checked the neighbour: an **aliased import** (`c "…/chi"` → `c.NewRouter()`) matches **none** of the three entries, but that is **pre-existing and uniform** — every entry requires a literal `chi.`, so aliasing was already unsupported before this PR and is not a residual it introduced.

### The rest of the asks

`MU4` and `MU2` **killed rather than prose-softened** — the `commented` fixture now carries all three spellings so every anchor is graded, and `ifInit` became `keywordInit` with the plain-`=` twin. The dead `ifInitWant` is gone rather than repaired into something vacuous. And the body now carries the measurement in place of the statement-level-coverage claim, the softened C# credit, and the `for`/`switch`/`go`/`const`/qualified-LHS wins — **seven keyword shapes, more than either the reviewer or I had.**

**Merging on green.**

PR #7024, opened 2026-09-08
fix(#7013): 52 rule files' import_markers were written correctly and unreadable — the container key was misspelled 14 ways

cajasmota:
Step 1 of #7013's fix path, as re-ordered by the #7014 audit: make the 52 files' existing `import_markers` visible before anyone writes new ones.

**Tree measured, by absolute path:** `/private/tmp/claude-501/-Users-jorgecajas-Projects-archigraph/c0fefed4-0195-4c81-9a94-07a7145710a4/scratchpad/impl-7013a-k9x4t/wt`, branch `lane/7013-container-spellings`, parent `87106f12c`. Reference tree for the before/after comparison: a second checkout of `87106f12c` at `.../impl-7013a-k9x4t/ref`. Gate on every figure: macOS 25.6.0 / arm64 / APFS, host Go toolchain (#6966).

## Decision: reconcile the files, do not alias the loader — and pin that choice

The brief offered both. Measured before choosing, and the measurement makes the trade one-sided:

- **The rename is sufficient.** All 247 misspelled containers are YAML mappings, and **246 of 247** have a sub-shape `FrameworkMeta` already models (`name` a string in 246, `detection` a mapping in 209, `import_markers` a list of strings in all 52 that have one). Nothing else has to change for the markers to load. The one exception is **`objective_c/frameworks/platform.yaml`**, whose `mobile:` block carries neither `name` nor `detection` (it has `platform:`, `category:`, `entry_point_signals:`, `structural_signals:`): after the rename it decodes to an empty `FrameworkMeta`, which is exactly what it decoded to before — no name, no markers, nothing gated — so the rename is harmless there, and the file is now a `frameworks:` block with no modelled field in it, worth a line on #7014's metadata pass. The other unmodelled keys those files carry (`category`, `notes`, `npm_packages`, `detection_signals.*`, …) are dropped exactly as they already are under `frameworks:` in the other 266 files — that is #7014's metadata migration, untouched here.
- **No file needed merging.** **0 of 520** loader-scoped files carry both `frameworks:` and a misspelling, so this is a key rename per file, never a merge of two blocks.
- **Aliasing legitimises 14 spellings permanently** and is the one direction that cannot be walked back. `TestLoader_DoesNotAdoptAnUnmodelledContainer` now records that decision, because without it the choice is unpinned: a loader that quietly adopts markers from whatever container it finds passes every other assertion here (after the reconciliation, no rule file offers it one to adopt).

## What changed

247 files, one line each: the top-level container key → `frameworks:`. The full diff is `247 -<misspelling>:` / `247 +frameworks:` and nothing else; **0 files touched outside `internal/engine/rules/`**, **0 `source_patterns` entries touched**, and none of `go/frameworks/{chi,echo,fiber,gin}.yaml` (another lane owns those) — they carry no container key at all.

| was | files | | was | files |
|---|---:|---|---|---:|
| `orms_database` | 121 | | `orms_databases` | 6 |
| `orms` | 35 | | `framework_templates` | 5 |
| `database_libraries` | 14 | | `mobile` | 4 |
| `template_engines` | 11 | | `framework` | 3 |
| `orms_and_database` | 11 | | `queues` | 1 |
| `orm_db` | 10 | | | |
| `orm_database` | 9 | | | |
| `frameworks_and_dialects` | 9 | | | |
| `database` | 8 | | | |

After: **one spelling in the tree** — `frameworks:` in 513 files, `0` files with markers under anything else. The 5 remaining files with no container at all (`go/frameworks/{chi,echo,fiber,gin,gorm}.yaml`) and the 2 empty ruby files are out of scope; they hide nothing.

## 1. How many markers became live: 52 files, 95 → 147

**147 of 520** loader-scoped files now declare readable `frameworks.detection.import_markers`, up from **95**. The 52 recovered sit under the five misspellings that carried markers: `orms_database` 18, `orms` 15, `database_libraries` 13, `database` 4, `orm_database` 2. Named examples, with the markers that were unreadable: `csharp/orms/mongodb_csharp.yaml` (`using MongoDB.Driver`, `using MongoDB.Bson`), `go/orms/bun.yaml` (`"github.com/uptrace/bun"` + 2 dialects), `lua/orms/lua_resty_mysql.yaml` (7 markers), `zig/orms/zig_sqlite.yaml` (`@import("sqlite")`), `csharp/orms/redis_stackexchange.yaml`, `clojure/orms/datomic.yaml` (18 markers).

## 2. What regresses: nothing, and the reason is structural before it is empirical

`importMarkers` reaches the engine at `detector.go:153` and is read by exactly one thing — `frameworkPresent` — which is called only when some `source_patterns` entry **in the same rule file** sets `requires_framework` (`detector.go:394-410`). So a file's markers can only change behaviour through its own patterns.

- **All 247 renamed files compile to a wholly inert rule set.** Decoding them through `FrameworkRule` gives `sourcePatternEntries=0`, `relationshipRuleEntries=0`, `fileConventionEntries=0`, `requiresFrameworkEntries=0` — no patterns of any kind, so `frameworkGateNeeded` is false, the recovered markers are never consulted, and such a rule set cannot emit an entity or an edge whatever its markers say. (Read precisely: **89 of the 247 do carry a top-level `source_patterns:` key** — always the empty list `source_patterns: []`. "No entries" is the true statement; a `grep -l source_patterns` over the 247 will not return nothing.)
- The **10 live `requires_framework` sites in 7 files** (`ansible/frameworks/ansible_core.yaml` ×2, `cicd/frameworks/github_actions.yaml` ×3, `docker/frameworks/docker_compose.yaml`, `python/frameworks/cherrypy.yaml`, `python/frameworks/falcon.yaml`, `python/frameworks/sqlalchemy.yaml`, `vbnet/frameworks/winforms.yaml`) are **disjoint from the 247** — all seven already keyed on `frameworks:`, all ten already evaluating against visible markers. Their gating is byte-identical before and after.
- `frameworks.name` is the only other key the rename exposes, and it has no reader (#7014).

**This is a property of the 247 files, not a sample of their behaviour**, and it is checkable from the diff alone: inert rule sets cannot move extraction, whatever corpus you point them at. A before/after corpus comparison of the real `Detector` over both rule trees was also run during development and agreed — but that harness is not in the diff, so nobody can inspect its comparison key or re-run it, and an uncheckable measurement is not evidence. The structural result above is what this claim rests on.

So step 1 recovers the metadata and moves no extraction. The recall trade #7013 is holding on (`requires_framework`'s per-file gate) is untouched and still unmeasured; this change only makes it expressible for 52 more files.

## 3. Golden ratchet: unmoved

`scripts/quality/run.sh --ratchet` → exit **0**, `38 gated fixtures held their baseline (29 at 100% must-have recall)`. No constant bumped, no `--update-baseline`.

## The pin

`internal/engine/rule_container_key_7013_test.go`, asserting artefacts rather than counts:

- **`TestRuleTree_RecoveredImportMarkersLoad`** — five named files, one per formerly-wrong spelling that carried markers, must decode their **exact** marker list and framework name through the loader's own `yaml.Unmarshal`-into-`FrameworkRule` step.
- **`TestRuleTree_NoImportMarkersHiddenUnderAnUnmodelledContainer`** — no loader-scoped file anywhere may declare `import_markers` under a container the schema does not model. Non-vacuity both ways: ≥500 files examined, and ≥147 with readable markers, so "reconciled by deleting the markers" cannot read as clean.
- **`TestRuleTree_NoReconciledContainerSpellingReturns`** — the 14 spellings are a closed, named set and none may return, marker-bearing or not. This is what makes the fix non-recurring; it is **not** the strict top-level guard #7014 priced at 247 red files, and it costs 0 red today.
- **`TestLoader_DoesNotAdoptAnUnmodelledContainer`** — the loader must not learn the misspellings (see Decision).
- Two positive controls on the checker itself (it must flag a hidden container, and must not flag an empty `import_markers` list).

## Mutants, permissive direction first

Scored in a **separate checkout** (`.../impl-7013a-k9x4t/mut`, detached at the branch commit) so the verification tree was never mutated. Green baseline immediately before each, `go vet` first, `-count=1`, each edit proved landed by an exact occurrence count (`assert n==1` on the replacement, then a `grep -c` on the mutated text), and each restored by `cp` from a shasum-verified backup — never `git checkout --` — with `git diff HEAD` confirmed empty after every one.

| # | mutant | verdict |
|---|---|---|
| M1 | re-break one spelling on a marker-bearing file (`csharp/orms/mongodb_csharp.yaml` → `orms_database:`) | **DEAD** — 3 tests, by file and key name |
| M2 | reconciliation silently skips a file that has **no** markers (`clojure/frameworks/amazonica_dynamodb.yaml` → `framework:`) | **DEAD** — only by `NoReconciledContainerSpellingReturns`; without that test this mutant is alive, which is why it exists |
| M3 | make the checker permissive — treat `orms_database`/`database` as visible containers | **DEAD** — the positive control names the two files it stopped flagging |
| M4 | make the **loader** adopt markers from any container when `frameworks:` is absent (the aliasing option, applied silently) — implemented as a pre-`Unmarshal` adoption pass in `loader.go` | **ALIVE** against `./internal/engine -count=1` (ok, 77.6s) and against all three of this PR's tree guards — nothing in the tree feeds it, since the reconciliation removed every input. Killed by the new `TestLoader_DoesNotAdoptAnUnmodelledContainer` (`the loader adopted 1 import_markers from a container it does not model`). ~20 lines. |

M4 is the one worth naming: the guard and the reconciliation were mutually masking — each made the other's failure invisible — so the loader control had to be written on a synthetic FS rather than on the tree.

**Scoped precisely, because "the whole suite was green" is too strong a claim to carry unqualified.** "ALIVE" above is about *this* adopting-loader implementation against `./internal/engine` at this commit. The independent review reproduced the mutant as an `UnmarshalYAML` method on `FrameworkRule` instead, and **that variant is additionally caught by the pre-existing `TestInertRuleFiles_DetectsAnInertFile` (#6393)** — a file keyed on an unmodelled container stops parsing to the zero rule once the loader adopts its contents. So the new test is the only guard that names this defect, and the mutual-masking pair is real; it is not the only net under every possible adopting loader.

## Verification

- `go vet ./internal/engine/` clean; `gofmt -l` clean on the new file.
- `go test ./internal/engine/ -count=1` → **ok, 52.9s**, exit 0.
- `go test ./cmd/grafel/ -count=1` → **ok, 171.7s**, exit 0, in a fresh three-variable sandbox (`HOME`/`GRAFEL_HOME`/`GRAFEL_DAEMON_ROOT` all under scratch); the real `~/.grafel` was never touched.
- `scripts/quality/run.sh --ratchet` → exit 0, same sandbox.
- **The other consumers of the rules tree**, which my original verification set omitted and the independent review ran: `go test ./internal/entkinds/ ./internal/relkinds/ ./tools/coverage/ ./internal/types/ -count=1` → all **ok** (22.2s / 16.7s / 5.2s / 14.0s). These matter here because `entkinds` and `relkinds` walk the YAML **at any depth** rather than through the schema, and the rename does shift a key path prefix for 5 files — `html_templates/frameworks/*.yaml` carry `scope_entity_mapping.entity_type: Component`, formerly under `template_engines:`. `tools/coverage` is directory- and filename-driven and never reads the container key. Green, so this is a gap in the *stated* verification rather than a defect — and a rule-tree change should name every consumer of the tree it checked (cf. "package selection is not suite selection").
- The commit is 248 files: 247 one-line renames + the new test. `git status` clean in `wt`, `mut` and the primary checkout.

## Stated limits

- **The 465 unmodelled metadata key paths are untouched**, as are the 59 `package_json_deps` declarations and every `requires_framework` setting. This change writes no new markers and sets no gate — deliberately (#7013 steps 2 and 3).
- **The no-regression result holds only while those 247 rule sets stay inert.** It is a structural property of the files as they are, not a promise about the future: a `source_patterns` entry added to one of them could change behaviour — which is the point of recovering the markers.
- **The closed 14-item spelling list has a known gap**: a *fifteenth* spelling on a *marker-less* file is invisible to both new tree guards (no markers to hide, not on the list) and stays alive. The review priced the fix — after this PR the loader-scoped tree holds exactly four distinct top-level keys, all modelled (`frameworks` 513, `source_patterns` 162, `relationship_rules` 158, `file_conventions` 154), so the strict top-level guard #7014 measured at 247 red files now **costs 0 red** and subsumes the hand-maintained list. Filed as an immediate follow-up rather than widened into this PR.
- The 5 container-less `go/frameworks/*.yaml` files still declare no detection metadata at all; #7013 step 2 has to write theirs, and now has one spelling to write it under.

Refs #7013, #7014.

### DIFF
--- a/internal/engine/rule_container_key_7013_test.go
+++ b/internal/engine/rule_container_key_7013_test.go
@@ -1,0 +1,427 @@
+package engine
+
+import (
+	"io/fs"
+	"path/filepath"
+	"strings"
+	"testing"
+	"testing/fstest"
+
+	"gopkg.in/yaml.v3"
+)
+
+// This file is the guard for step 1 of #7013.
+//
+// FrameworkRule models exactly one top-level container, `frameworks:`, and
+// that container is the only route by which `detection.import_markers` reaches
+// the engine (loader.go unmarshals into FrameworkRule; detector.go:153 copies
+// fr.Frameworks.Detection.ImportMarkers into the compiled rule set, and
+// frameworkPresent is the sole consumer). yaml.Unmarshal drops unknown
+// top-level keys silently, so a file that spells the container `orms_database:`
+// parses successfully and hands the engine no markers at all.
+//
+// The #7014 audit measured that: 247 of the 520 loader-scoped rule files keyed
+// their container on one of 14 misspellings — `orms_database` (121), `orms`
+// (35), `database_libraries` (14), `template_engines` (11), `orms_and_database`
+// (11), `orm_db` (10), `orm_database` (9), `frameworks_and_dialects` (9),
+// `database` (8), `orms_databases` (6), `framework_templates` (5), `mobile` (4),
+// `framework` (3), `queues` (1) — and 52 of those 247 carried real, correct
+// `detection.import_markers` the engine could not see. The markers were written;
+// they were unreadable.
+//
+// The 247 files were reconciled onto `frameworks:`. The two assertions here are
+// the artefact, not the count:
+//
+//   - specific named files' markers load through the loader's own decode path,
+//     with their exact contents; and
+//   - no loader-scoped rule file anywhere in the tree keeps `import_markers`
+//     under a container the schema does not model.
+//
+// The second is what makes the fix non-recurring: re-introducing any of the 14
+// spellings — or inventing a fifteenth — fails here by file name and key name,
+// whether it arrives as a regression or as a new rule file written to the old
+// habit.
+//
+// Deliberately out of scope: the rest of the metadata half. 465 unmodelled key
+// paths remain (`category`, `notes`, `npm_packages`, `detection_signals.*`, …)
+// and are still dropped at load; that is the tracked migration on #7014, not
+// this guard. This guard is about the ONE key that gates a live mechanism.
+
+// hiddenMarkerFile names one rule file that declares import_markers under a
+// top-level container the schema does not model.
+type hiddenMarkerFile struct {
+	path string
+	key  string
+}
+
+// importMarkerVisibility walks the loader-scoped rule files under rootDir and
+// reports, for each, whether its declared import_markers are reachable by the
+// engine.
+//
+// hidden lists files whose markers sit under an unmodelled top-level container
+// (the #7013 defect). visible counts files whose markers sit under
+// `frameworks:` and therefore load. examined counts loader-scoped files looked
+// at, so a caller can tell "nothing hidden" from "nothing scanned" — a walk
+// that finds no files would otherwise report a clean tree.
+//
+// It is a free function so the checker can be run against a synthetic tree; a
+// checker that always answers "all visible" is worth nothing on the real one.
+func importMarkerVisibility(fsys fs.FS, rootDir string) (hidden []hiddenMarkerFile, visible, examined int, err error) {
+	walkErr := fs.WalkDir(fsys, rootDir, func(path string, d fs.DirEntry, err error) error {
+		if err != nil {
+			return err
+		}
+		if d.IsDir() {
+			return nil
+		}
+		if ext := filepath.Ext(path); ext != ".yaml" && ext != ".yml" {
+			return nil
+		}
+		rel, relErr := filepath.Rel(rootDir, path)
+		if relErr != nil {
+			return nil
+		}
+		// Mirror the loader: only <lang>/<subdir>/<file>.yaml, supported subdir.
+		parts := strings.Split(filepath.ToSlash(rel), "/")
+		if len(parts) != 3 || !ruleSubdirs[parts[1]] {
+			return nil
+		}
+		data, readErr := fs.ReadFile(fsys, path)
+		if readErr != nil {
+			return nil
+		}
+		var raw map[string]any
+		if yaml.Unmarshal(data, &raw) != nil {
+			// A parse failure is a different defect, caught by the #6341
+			// inventory assertion. Not hidden — absent.
+			return nil
+		}
+		examined++
+		for key, val := range raw {
+			if !declaresImportMarkers(val) {
+				continue
+			}
+			if key == "frameworks" {
+				visible++
+				continue
+			}
+			hidden = append(hidden, hiddenMarkerFile{path: filepath.ToSlash(path), key: key})
+		}
+		return nil
+	})
+	return hidden, visible, examined, walkErr
+}
+
+// declaresImportMarkers reports whether a top-level container value carries a
+// non-empty detection.import_markers list.
+func declaresImportMarkers(val any) bool {
+	container, ok := val.(map[string]any)
+	if !ok {
+		return false
+	}
+	detection, ok := container["detection"].(map[string]any)
+	if !ok {
+		return false
+	}
+	markers, ok := detection["import_markers"].([]any)
+	return ok && len(markers) > 0
+}
+
+// TestRuleTree_NoImportMarkersHiddenUnderAnUnmodelledContainer is the assertion
+// over the real embedded tree. Every rule file's import_markers must sit under
+// `frameworks:`, the one container FrameworkRule models.
+func TestRuleTree_NoImportMarkersHiddenUnderAnUnmodelledContainer(t *testing.T) {
+	hidden, visible, examined, err := importMarkerVisibility(rulesFS, "rules")
+	if err != nil {
+		t.Fatalf("walking the embedded rules tree: %v", err)
+	}
+
+	// Non-vacuity, both halves. A walk that scanned nothing, or a tree whose
+	// markers had been "reconciled" by deleting them, must not read as clean.
+	if examined < 500 {
+		t.Fatalf("examined %d loader-scoped rule files, want >= 500: the guard is "+
+			"scanning a fraction of the tree, so its verdict is not about the tree", examined)
+	}
+	if visible < 147 {
+		t.Errorf("%d rule files declare readable frameworks.detection.import_markers, want >= 147: "+
+			"95 were readable before #7013 step 1 and 52 more were recovered from misspelled "+
+			"containers; a drop means markers were deleted rather than reconciled", visible)
+	}
+
+	for _, h := range hidden {
+		t.Errorf("%s declares detection.import_markers under top-level key %q, which "+
+			"FrameworkRule does not model — the markers are dropped at load and "+
+			"requires_framework cannot evaluate against them (#7013, #7014). "+
+			"The container is spelled `frameworks:`.", h.path, h.key)
+	}
+}
+
+// reconciledContainerSpellings are the 14 top-level keys the 247 files were
+// keyed on before #7013 step 1, measured over the loader-scoped tree. They are
+// listed as the closed set they are: a file using any of them again is a
+// regression of this change, whether or not it happens to carry markers today.
+//
+// This is narrower than the strict top-level guard #7014 priced (247 red files,
+// a migration): it forbids the spellings that were actually removed, and costs
+// nothing.
+var reconciledContainerSpellings = []string{
+	"orms_database", "orms", "database_libraries", "template_engines",
+	"orms_and_database", "orm_db", "orm_database", "frameworks_and_dialects",
+	"database", "orms_databases", "framework_templates", "mobile",
+	"framework", "queues",
+}
+
+// TestRuleTree_NoReconciledContainerSpellingReturns keeps the tree at one
+// spelling. TestRuleTree_NoImportMarkersHiddenUnderAnUnmodelledContainer only
+// sees a file once it carries markers; a file re-keyed on `orms_database:` with
+// no markers yet is invisible to it, and is exactly how the next 52 would
+// accumulate.
+func TestRuleTree_NoReconciledContainerSpellingReturns(t *testing.T) {
+	banned := make(map[string]bool, len(reconciledContainerSpellings))
+	for _, k := range reconciledContainerSpellings {
+		banned[k] = true
+	}
+
+	examined := 0
+	err := fs.WalkDir(rulesFS, "rules", func(path string, d fs.DirEntry, err error) error {
+		if err != nil || d.IsDir() {
+			return err
+		}
+		if ext := filepath.Ext(path); ext != ".yaml" && ext != ".yml" {
+			return nil
+		}
+		rel, relErr := filepath.Rel("rules", path)
+		if relErr != nil {
+			return nil
+		}
+		parts := strings.Split(filepath.ToSlash(rel), "/")
+		if len(parts) != 3 || !ruleSubdirs[parts[1]] {
+			return nil
+		}
+		data, readErr := fs.ReadFile(rulesFS, path)
+		if readErr != nil {
+			return nil
+		}
+		var raw map[string]any
+		if yaml.Unmarshal(data, &raw) != nil {
+			return nil
+		}
+		examined++
+		for key := range raw {
+			if banned[key] {
+				t.Errorf("%s keys its detection metadata on %q — one of the 14 spellings "+
+					"reconciled onto `frameworks:` in #7013 step 1. The engine models "+
+					"`frameworks:` and nothing else, so everything under %q is dropped at "+
+					"load, including any import_markers added later.",
+					filepath.ToSlash(path), key, key)
+			}
+		}
+		return nil
+	})
+	if err != nil {
+		t.Fatalf("walking the embedded rules tree: %v", err)
+	}
+	if examined < 500 {
+		t.Fatalf("examined %d loader-scoped rule files, want >= 500", examined)
+	}
+}
+
+// TestRuleTree_RecoveredImportMarkersLoad pins the artefact this change exists
+// to produce: the named files whose markers were unreadable before the
+// reconciliation now decode, with their exact contents, through the same
+// yaml.Unmarshal-into-FrameworkRule step the loader performs.
+//
+// One file per formerly-wrong spelling that carried markers.
+func TestRuleTree_RecoveredImportMarkersLoad(t *testing.T) {
+	cases := []struct {
+		path        string // embedded path
+		wasKeyedOn  string // the container spelling it used before #7013 step 1
+		name        string // frameworks.name, also unreadable before
+		wantMarkers []string
+	}{
+		{
+			path:       "rules/csharp/orms/mongodb_csharp.yaml",
+			wasKeyedOn: "orms_database",
+			name:       "MongoDB.Driver (C# official driver)",
+			wantMarkers: []string{
+				"using MongoDB.Driver",
+				"using MongoDB.Bson",
+			},
+		},
+		{
+			path:       "rules/csharp/orms/redis_stackexchange.yaml",
+			wasKeyedOn: "orms",
+			name:       "StackExchange.Redis",
+			wantMarkers: []string{
+				"using StackExchange.Redis",
+				"using ServiceStack.Redis",
+			},
+		},
+		{
+			path:       "rules/go/orms/bun.yaml",
+			wasKeyedOn: "database_libraries",
+			name:       "bun",
+			wantMarkers: []string{
+				`"github.com/uptrace/bun"`,
+				`"github.com/uptrace/bun/dialect/pgdialect"`,
+				`"github.com/uptrace/bun/dialect/mysqldialect"`,
+			},
+		},
+		{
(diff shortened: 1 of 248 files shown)


cajasmota:
## Coordinator mutants on `bf26a9f82` — **DA-1 and DA-2 both DEAD**, and the layering is worth naming

Own detached worktree at `/private/tmp/.../coord-mut-7024-x3d/wt` (absolute). Green baseline immediately before each. Restored by `cp` from SHA-256-verified backups (`c355e32c…` for `schema.go`, `d5d70eed…` for the rule file — both matched after restore); `git diff HEAD` and `git status --porcelain` empty afterwards.

A 247-file rename invites mutants on the *mechanism* — the spellings, the guard, the closed set. Your four and the reviewer's cover that. I went after the **payload** instead: the claim that **52 files' `import_markers` are now readable**. That is what this PR is for, and it is the thing a mechanism test can pass without.

### DA-1 — make `import_markers` unreadable globally

`schema.go:46`, `yaml:"import_markers"` → `yaml:"import_markers_DA1"`. Occurrence asserted 1, `go vet` clean. Note this sits under `frameworks:` (metadata), so #7019's strict-container guard does **not** fire on it — a clean mutant, not a decode error.

**Result: DEAD** — but by `TestIssue6927_DockerAnsible_GatesAreDeclaredAndLoaded` and `TestIssue6927_Compose_ServiceRuleNeedsBothFences`, which are **pre-existing**.

That is worth stating precisely rather than as a win: those tests cover `docker_compose` and `ansible_core`, which are among the **7 files whose markers were already readable**. So #6927 pins that markers load *at all*; it says nothing about the 52 this PR rescued.

### DA-2 — regress exactly one of the 52

Reverted `rules/clojure/orms/datascript.yaml`'s top-level key from `frameworks:` back to `database_libraries:` — one file, one line, the precise inverse of what this PR did to it.

**Result: DEAD**, by **both tests this PR adds**:

```
--- FAIL: TestRuleTree_NoImportMarkersHiddenUnderAnUnmodelledContainer
--- FAIL: TestRuleTree_NoReconciledContainerSpellingReturns
```

So the payload **is** pinned per-file, and by this PR's own work. The first is the artefact assertion I asked for — markers not hidden — and the second is the closed-set spelling guard. They fail on different grounds for the same input, which is the right redundancy: one would still catch a *new* misspelling that hides markers, the other a *reintroduced* known one that does not.

### The layering, stated

- **Markers load at all** → pinned by #6927, pre-existing, on 2 of the 7 originally-readable files.
- **These 52 files' markers are not hidden** → pinned by this PR.
- **No reconciled spelling returns** → pinned by this PR.
- **The loader does not adopt an unmodelled container** → pinned by your `TestLoader_DoesNotAdoptAnUnmodelledContainer`, which exists because you found the guard and the reconciliation were **mutually masking** — with every file renamed, nothing exercised the guard.

That last one is the part I want on the record. Two mechanisms that only fire together grade neither, and it is the defect class this board files most often. **Finding it in your own work, and fixing it with a synthetic FS rather than declaring the mutant equivalent, is the first time this session an author has caught that shape unprompted.**

### On the decision to rename rather than alias

Measuring first — all 247 misspelled containers are mappings already matching `FrameworkMeta`, and **0 of 520 files carry both keys**, so a rename needs no merge — is what made this a rename rather than a migration. I had a mild preference and said to measure before choosing; the measurement is what settles it, and aliasing would have legitimised 14 spellings permanently.

**Merging on green**, pending the independent review.

cajasmota:
## Independent adversarial review — **APPROVE** (with three follow-ups, none blocking)

**Trees measured, absolute paths.** Branch: `/private/tmp/claude-501/-Users-jorgecajas-Projects-archigraph/c0fefed4-0195-4c81-9a94-07a7145710a4/scratchpad/review-7024-q7z/wt` @ `bf26a9f82`. Parent reference: `.../review-7024-q7z/ref` @ `87106f12c`. Mutants scored in a third checkout `.../review-7024-q7z/mut` @ `bf26a9f82`, never in the verification tree. Gate on every figure: macOS 25.6.0 / arm64 / APFS, host Go toolchain (#6966). I did not push and did not merge.

### Diff shape — confirmed exactly as described

248 files. 247 of them are one changed line each (`247 -<misspelling>` / `247 +frameworks:` and literally nothing else); the 248th is the new test. **0 files outside `internal/engine/rules/`** except that test, **0 `source_patterns` edits**, and **none of `go/frameworks/{chi,echo,fiber,gin}.yaml`**. The 14-spelling histogram in the body matches the diff line-for-line. No production Go file changed — the "loader guard" is a test, not a loader change.

Independent yaml.v3 census of the loader-scoped tree (520 files both sides): **before** 95 files with readable markers, **52 hidden** across 5 spellings; **after** 147 readable, **0 hidden**, **0 misspellings remaining**. Reproduces the body's numbers.

### (b) Did any file carry both keys? **No — 0 of 520.**

Decoded every loader-scoped file at `87106f12c` and counted files carrying `frameworks:` alongside a container-shaped sibling key: **0**. Confirmed post-rename as well. This is a rename, not a merge.

One correction to the sub-shape claim: it is **246 of 247**, not 247. `objective_c/frameworks/platform.yaml`'s `mobile:` block carries neither `name` nor `detection` (it has `platform:`, `category:`, `entry_point_signals:`, `structural_signals:`). It renames harmlessly — the file now has a `frameworks:` with no modelled field in it — but it is a `frameworks:` block with no name, which is worth a line on #7014's metadata pass.

### (a) Is the 0-differing-keys comparison sound and reproducible? **Not reproducible as written — and not load-bearing.**

The harness is not in the diff, so I could not inspect the comparison key, and per the brief that is exactly the failure mode to worry about (#6973). I did not need it: the conclusion is established **structurally**, and more strongly than the body states.

Decoding all 247 renamed files through `FrameworkRule` gives:

```
files=247 sourcePatternEntries=0 relationshipRuleEntries=0 fileConventionEntries=0
withMarkers=52 requiresFrameworkEntries=0
```

Not just "no `source_patterns`" — **no `relationship_rules` and no `file_conventions` either**. Every one of the 247 compiles to a rule set with no patterns of any kind, so it cannot emit an entity or an edge whatever its markers say. `grep -l requires_framework` over the 247 returns **nothing**. Extraction cannot move; the corpus run confirms that rather than establishing it, which the body's "Stated limits" already concedes.

Two notes: (1) 89 of the 247 **do** carry a top-level `source_patterns:` key — always `source_patterns: []`. "Carries no entry" is true; "carries no `source_patterns`" is what a reader will grep for and will appear false. Worth one clarifying word. (2) Consider committing the harness, or replacing that paragraph with the inert-rule-set statement above, which is checkable from the diff alone.

### (c) "markers are consumed only by `frameworkPresent`" — **survives a method that defeats indirection.**

Rename-to-break, not grep. In `mut` I renamed `FrameworkRule.Frameworks`, `FrameworkMeta.Name` and `FrameworkDetection.ImportMarkers` (each edit proved landed by an exact `grep -c` == 1) and compiled the whole module with `-gcflags=all=-e` so nothing truncated at "too many errors":

- `go build -gcflags=all=-e ./...` → **exactly one** non-test error: `internal/engine/detector.go:153`.
- `go test -gcflags=all=-e -run XXXNONE ./...` → the only non-test file named is `internal/engine/detector.go`.
- No reflection or `yaml.Marshal` path over `FrameworkRule` anywhere in the module.

So the field has one reader, and `frameworkPresent` is its only consumer. `schema.go` restored by `cp` from a shasum-verified backup (`b476c98a…` matched after restore), `git status` empty.

**Follow-on — what else reads the rules tree, and a gap in the stated verification.** Three packages read `internal/engine/rules` from disk outside the engine: `tools/coverage/discover.go` (directory/filename driven — never looks at the container key), `internal/entkinds/scan.go` and `internal/relkinds/scan.go` (walk the YAML **at any depth**, not through the schema). The rename does change a YAML path prefix for 5 files — `html_templates/frameworks/*.yaml` carry `scope_entity_mapping.entity_type: Component`, formerly under `template_engines:`. Those packages were **not** in the PR's verification list. I ran them:

```
ok internal/entkinds 22.2s   ok internal/relkinds 16.7s   ok tools/coverage 5.2s   ok internal/types 14.0s
```

Green, so no regression — but please add them to the stated verification set; the current set was under-scoped for a change to the rule tree (cf. "package selection is not suite selection").

### (d) Does the guard's closed set have gaps? **Yes, one — and it just became free to close.**

| # | mutant | verdict |
|---|---|---|
| **MR1** | a **15th** spelling on a **marker-bearing** file (`go/orms/ent.yaml`: `frameworks:` → `orm:`, a singular near-miss absent from the 14) | **DEAD** — `NoImportMarkersHiddenUnderAnUnmodelledContainer` names file *and* key; the `visible >= 147` floor fires too. The open-set guard genuinely covers a fifteenth spelling. |
| **MR2** | the same 15th spelling on a **marker-less** file (`python/orms/sqlite_py.yaml`: `frameworks:` → `orm:`) | **ALIVE** — whole `./internal/engine` suite `ok 47.8s`. Invisible to the hidden-marker guard (no markers) and to the closed 14-item list (`orm` not in it). **Distinguishable and ungraded.** Not production-reachable today (`frameworks.name` has no reader and that file has 0 patterns), but it silently re-opens the exact defect class this PR closes, which is the stated purpose of `NoReconciledContainerSpellingReturns`. |
| **MR3** | markers under a modelled container but a **misspelled sub-key** (`python/orms/django_orm.yaml`: `frameworks.detection:` → `frameworks.detection_signals:`) | **DEAD**, but only via the `visible >= 147` **counter**, and the message misdiagnoses it — "markers were deleted rather than reconciled" when they were neither. A whole class (wrong sub-key under a right container) is graded by a count and by nothing that names an artefact. |
| **MR4** | **reproduce M4** — an `UnmarshalYAML` on `FrameworkRule` that adopts `name`/`import_markers` from any container when `frameworks:` yields nothing (the aliasing option, applied silently) | **DEAD by `TestLoader_DoesNotAdoptAnUnmodelledContainer` only** among the new guards: `NoImportMarkersHidden…`, `NoReconciledContainerSpellingReturns` and `RecoveredImportMarkersLoad` all **PASS** under it. The mutual-masking claim reproduces exactly, and the synthetic-FS control is the right instrument. **Correction:** my variant is *also* caught by pre-existing `TestInertRuleFiles_DetectsAnInertFile`, so "ALIVE against the whole `./internal/engine` suite" does not reproduce for this implementation — the new test remains the one that names the defect, but it is not the sole net for every adopting loader. |

**Pricing MR2, which is the useful part.** After this PR the loader-scoped tree has exactly **four distinct top-level keys**, all four modelled by `FrameworkRule`:

```
513 frameworks   162 source_patterns   158 relationship_rules   154 file_conventions
```

So the **strict top-level guard #7014 priced at 247 red files now costs 0 red**, and it subsumes the hand-maintained 14-item list — MR1 *and* MR2 both die under it, and no fifteenth spelling can ever be invented past it. This PR is what made that guard free; that argues for taking it now (here or as an immediate follow-up) rather than shipping a closed set that is one un-listed word away from silent.

### Verification I ran myself

- `go vet ./internal/engine/` exit 0; `go test ./internal/engine/ -count=1` → **ok 48.7s**, exit 0 (green baseline re-established immediately before each mutant).
- `go test ./cmd/grafel/ -count=1` → **ok 205.7s**, exit 0, in a fresh three-variable sandbox (`HOME`/`GRAFEL_HOME`/`GRAFEL_DAEMON_ROOT` all under scratch). The real `~/.grafel` was never written.
- `go test ./internal/entkinds/ ./internal/relkinds/ ./tools/coverage/ ./internal/types/ -count=1` → all ok (see above).
- **Golden ratchet, verified the strong way**: `scripts/quality/run.sh --ratchet` in its own sandbox → exit 0, `38 gated fixtures held their baseline (29 at 100% must-have recall)`, **and `git status --porcelain` in `wt` is empty after the run** — so no ceiling was silently rewritten, which the summary line alone would not have shown.
- Every mutant restored by `cp` from a shasum-verified backup, never `git checkout --`; `git status --porcelain | wc -l` == 0 in `mut` confirmed after each.

### Summary

The payload is real and correctly scoped, the rename is provably a rename and not a merge, the "one consumer" claim holds under a compile-level probe that indirection cannot defeat, and the no-regression result is structurally guaranteed rather than merely measured. **APPROVE.**

Non-blocking follow-ups, in order of value:
1. Take the strict top-level container guard now that it costs 0 red — it closes MR2 and retires the 14-item list (#7014).
2. Add `internal/entkinds`, `internal/relkinds`, `tools/coverage` to the verification set for rule-tree changes.
3. Fix the two small prose slips: "no `source_patterns`" (89 files have the key, always `[]`) and "247 sub-shapes match `FrameworkMeta`" (246 — `objective_c/frameworks/platform.yaml` is the exception).

PR #7026, opened 2026-09-08
fix(#6990): the relational-target scan ran 400 bytes forward regardless of where the declaration ended, so a field bound to the NEXT field's model

### LINKED ISSUE #6990: django: the 400-byte forward scan reads a FOLLOWING declaration's target — LIVE on django (LogEntry.user -> ContentType, binds to a real model); general population unmeasured, my zero-incidence claim was wrong

cajasmota:
Fixes #6990 (residue 2 only — see Scope).

## What was wrong

`internal/custom/python/django.go` handed `djangoModelRelTargetRe` a **raw 400-byte forward slice** from the matched field, unbounded by the declaration it belonged to:

```go
fullRHS := body[fIdx[0]:min(fIdx[0]+400, len(body))]
```

`models.ForeignKey(settings.AUTH_USER_MODEL, ...)` — dotted and lowercase-initial — is *correctly* rejected by the regex's `[A-Z]` symbol alternative. The regex then kept scanning inside that oversized window and matched the **next field's** `ForeignKey(ContentType`. django/contrib/admin/models.py's `LogEntry.user` came out targeting `ContentType`: a real model, a bound edge, the wrong thing.

## What bounds the window now

`djangoDeclEnd(body, fIdx[1]-1)` — `djangoModelFieldRe` ends `\s*\(`, so `fIdx[1]-1` is the constructor's own `(`. The scan follows **parenthesis depth**, not bytes and not lines, so:

- multi-line declarations are covered *by construction*, not incidentally;
- it skips the two places a `)` can appear without closing anything: string literals (single, double, **triple**-quoted, honouring `\` escapes) and `#` comments — and does not mistake a `#` inside a string for a comment, nor an apostrophe inside a comment for a string;
- on an **unbalanced** source it returns `-1` rather than guessing, and the caller falls back to the pre-#6990 400-byte window.

**Why the fallback is kept.** On every `-1` input it is byte-identical to `main`, so it cannot regress anything — and one `-1` path is reachable from **valid** Python, because `extractClassBody`'s indent cut can truncate a class body mid-declaration. That is a pre-existing condition, and preserving the old window is the correct behaviour there. (Better justification than the "never wider" one this body originally gave; thanks to the review for it.) The fallback width is now pinned by **magnitude**, not just sign: a target ending at byte 400 must be captured and one ending at byte 401 must not, which together admit exactly one window width.

**The regex is unchanged.** A left `\b` would close the same case but drops `TreeForeignKey(Category, ...)` with it (#6988). Shipping the bound alone keeps it graded — two guards that only fire together grade neither. `TestIssue6990_TheBoundIsTheSoleGuard` shows the distinguishing input directly: handed the old two-declaration text the unchanged regex still captures `ContentType`; handed the bounded text it captures nothing.

## Axes: varied vs held constant

`TestIssue6990_TargetFormsSurviveEveryDeclarationLayout` is a full 3x4 cross.

**Varied**
- **Declaration layout**: single-line · multi-line · multi-line with a trailing comma.
- **Target form**: `settings.AUTH_USER_MODEL` (dotted lowercase → *no* target) · bare CamelCase symbol `Profile` · quoted `"catalog.Author"` · quoted `'self'`.
- **Window bound vs byte count**: declaration-end (normal path) and the 400-byte fallback (unbalanced path), each with its own test.
- **`)` placement**: flat · nested call · dotted kwarg with no nesting · doubly nested · inside `"`/`'`/`"""` strings · inside a `#` comment · `#` inside a string · escaped quote · unterminated (all six ways, including a quote closed only on a later line).
- **Fallback width**: byte 400 (captured) vs byte 401 (not), plus a field's own target right after its own `(`.

**Held constant, and graded elsewhere**
- The constructor: `models.ForeignKey` throughout. The constructor gate is #6988's — `TestIssue6988_GateKeepsRelationalConstructors` / `TestIssue6988_GateRejectsGenericForeignKey`.
- The regex's left boundary — `TestIssue6988_RegexBoundaryRejected_WouldDropSubclasses` (unchanged, still green).
- The `[A-Z]` symbol anchor — `TestIssue6988_RegexRejectsLowercaseInitialSymbol`.

**Every** row, including every negative one, carries a **sentinel sibling** declared immediately after the field under test with a different target, asserted to resolve. That is what stops "want no target" passing vacuously: if the window still over-read, the sentinel's target is what it would steal.

Named forbidden rows: `TestIssue6990_NamedResidue2_RealFieldFollowedByGFK` (the issue's shape — a real FK directly above a `GenericForeignKey`, asserting *no* edge at all rather than "not `content_type`", so a re-broadened window cannot pass by picking a different wrong target).

## The known-wrong pin

`TestPythonFieldTargetType_KnownWrong_6990_OverReadBindsToTheNextFieldsTarget` asserted today's wrong behaviour and carried a `t.Skip` telling the fixer to delete it. Confirmed it fires with the fix in place, then **converted it** into `TestPythonFieldTargetType_6990_NoOverReadIntoTheNextFieldsTarget` rather than deleting it: the correct behaviour for `LogEntry.user` is a **negative** (the declaration names a settings indirection this extractor does not resolve, so it emits nothing), and nothing else in that file grades it. It keeps the same fixture, the same end-to-end merge+resolve path, and gains `LogEntry.content_type` as the positive control — the specific sibling whose target was being stolen. Its stale prose is rewritten.

## Mutants

Applied to `django.go` one at a time; each edit asserted to land by exact occurrence count + shasum before/after, restored by `cp` from a byte-for-byte backup with the shasum re-verified and `git diff HEAD` empty. Scored with `-count=1`.

| # | Mutant | Verdict |
|---|---|---|
| 1 | Revert the bound: `end = min(fIdx[0]+400, len(body))` unconditionally | **DEAD** — 5 tests, incl. all three layouts of the `settings.AUTH_USER_MODEL` row, the named GFK row, and the converted pin |
| 2 | Widen the fallback: `end = len(body)` on `-1` | **DEAD** — `TestIssue6990_UnbalancedSourceFallsBackToTheOldByteWindow` |
| CU-2 | Empty the fallback: `end = fIdx[0]` | **DEAD** (was ALIVE; killed in `aa220a250`) |
| C-120 / C-600 / C-700 | Fallback constant `400` → `120` / `600` / `700` | **DEAD** (120 and 600 were ALIVE — the constant was graded by sign, not magnitude) |
| 8 | Delete `case '\n': return -1` in the string skipper | **DEAD** (was ALIVE; a quote closed only on a later line) |
| 3 | Delete the string-literal skip | **DEAD** — 4 subtests |
| 4 | Delete the `#`-comment skip | **DEAD** — 2 subtests |
| 5 | Off-by-one on the closer: `return i` | **DEAD** — 12 subtests |
| 6 | Delete the triple-quote branch | **DEAD** — 1 subtest |
| 7 | Delete the backslash-escape case | **DEAD** — 1 subtest |

12/12 dead, all re-scored against the final tests. No equivalent-under-the-suite and no ungraded survivors.

**Non-vacuity of the negatives.** Every "want no target" row now asserts *both* that no `field_target_type` edge exists **and** that the field itself still does (a `CONTAINS` edge, or the entity in the merged set for the converted pin) — a negative that passes because the field vanished is passing for the wrong reason.

## Scope — residue 1 is deliberately NOT touched

`class CTForeignKey(GenericForeignKey)` still passes `isDjangoRelationalField`'s suffix check at `django.go`. That is a **policy decision** (deny-list vs allow-list) that #6988/#6989 settled once on measured merit, and it is pinned by `TestIssue6988_RegexBoundaryRejected_WouldDropSubclasses`. It is left alone here on purpose; #6990 stays open for it if the board wants it.

## Incidence — what is and is not claimed

- **Extractor-level behaviour is proven**, end to end through merge + `resolve.BuildIndex` / `ReferencesEmbedded`.
- **Graph-level incidence on the real django corpus is NOT ESTABLISHED and is not claimed.** An in-repo comment asserted that this edge stays dangling anyway because the `ext:django:` external-synthesis pass intercepts first; nobody has verified that, so it is neither repeated nor contradicted. No "fixes N edges" number appears anywhere in this PR — a number nobody measured is worse than none.
- **Gate (#6966)**: these edges exist only with the custom Python lane ON. It is off by default (`WithCustomExtractors`, `cmd/grafel/index.go`), so default-index baselines are untouched.

## Stale prose fixed

`django.go`'s comment above `djangoModelRelTargetRe` asserted this residue was *"zero-incidence on the measured corpus"*. The issue's own correction comment contradicts that: the measurement covered the **snake_case** target population and was stated over a wider one — `ContentType` is CamelCase and a genuine model, invisible to any shape filter. The comment now says the general population is **unmeasured, not zero**, and records what actually bounds the read. That sentence is how the bug survived once already.

## Verification

- `go test -count=1 ./internal/custom/python/` — green.
- `go test -count=1 ./internal/custom/... ./internal/extractors/python/... ./internal/resolve/...` — green.
- **`./cmd/grafel/` was NOT run, deliberately — and the original reason given here was wrong.** Two `cmd/grafel` tests *do* drive the django extractor: `inproc_custom_extractors_5989_test.go:93` and `generated_seam_6329_test.go:51`. The conclusion holds for a different reason: **both fixtures are single-line, balanced, with the target as the first argument**, so `djangoDeclEnd` returns the same slice the 400-byte window did and the bound is a no-op on them. Verified by reading both, not inferred from the diff's package set. "No consumer reaches this" has been wrong on this board before.

### DIFF
--- a/internal/custom/python/django.go
+++ b/internal/custom/python/django.go
@@ -157,13 +157,30 @@ var (
 	// but it drops `TreeForeignKey(Category, ...)` with it — measured, not
 	// assumed: see TestIssue6988_RegexBoundaryRejected_WouldDropSubclasses.
 	// This regex extracts a target; it is NOT the constructor filter.
-	// `isDjangoRelationalField` decides which constructors are PROCESSED — but it
-	// does not bound what this regex READS. The caller hands it a 400-char
-	// forward window (`fullRHS` below) that can span later lines, so a
-	// legitimate `ForeignKey` whose own argument this regex cannot parse
-	// (`models.ForeignKey(settings.AUTH_USER_MODEL, ...)`) can still match a
-	// string belonging to a constructor the gate REJECTED. That residue survives
-	// #6988 and is zero-incidence on the measured corpus; it is filed separately.
+	// `isDjangoRelationalField` decides which constructors are PROCESSED. It does
+	// not bound what this regex READS — the CALLER's window does, and until
+	// #6990 that window was a raw 400-byte forward slice from the matched field,
+	// unbounded by the declaration it belonged to. A legitimate `ForeignKey`
+	// whose own argument this regex cannot parse
+	// (`models.ForeignKey(settings.AUTH_USER_MODEL, ...)` — dotted and
+	// lowercase-initial, so the `[A-Z]` symbol alternative correctly rejects it)
+	// therefore ran on into the NEXT field's declaration and captured ITS target.
+	// django/contrib/admin/models.py's `LogEntry.user` did exactly that and came
+	// out targeting `ContentType`: a real model, a bound edge, the wrong thing.
+	//
+	// CORRECTION (#6990): an earlier version of this comment called that residue
+	// "zero-incidence on the measured corpus". That was a measurement of the
+	// snake_case target population (chosen for #6988's `content_type` shape)
+	// stated over a wider one. `ContentType` is CamelCase and a genuine model
+	// name, so no shape filter could see it. The general population was never
+	// counted — it is UNMEASURED, not zero.
+	//
+	// The window is now bounded by the field's OWN call parentheses
+	// (`djangoDeclEnd`), so this regex can no longer read text belonging to a
+	// constructor the gate rejected — or to any other declaration. The regex
+	// itself is UNCHANGED: adding a left `\b` here would close the same case but
+	// drop `TreeForeignKey(Category, ...)` with it (#6988), and two guards that
+	// only fire together grade neither.
 	djangoModelRelTargetRe = regexp.MustCompile(
 		`(?:ForeignKey|OneToOneField|ManyToManyField)\s*\(\s*(?:to\s*=\s*)?(?:["']([^"']+)["']|([A-Z][A-Za-z0-9_]*))`)
 
@@ -715,7 +732,19 @@ func (e *DjangoExtractor) Extract(ctx context.Context, file extractor.FileInput)
 			if attr == "" || !isDjangoRelationalField(rhs) {
 				continue
 			}
-			fullRHS := body[fIdx[0]:min(fIdx[0]+400, len(body))]
+			// #6990 — bound the target scan to THIS field's own declaration.
+			// fIdx[1] is one past the match, whose last byte is the `(` of the
+			// constructor call (djangoModelFieldRe ends `\s*\(`), so the slice
+			// runs from the start of the assignment line to the matching `)`.
+			// Multi-line declarations are covered by construction: the scan
+			// follows parens, not bytes or lines. The 400-byte cap survives ONLY
+			// as the unbalanced-source fallback, where it is exactly the old
+			// behaviour — never wider.
+			end := djangoDeclEnd(body, fIdx[1]-1)
+			if end < 0 {
+				end = min(fIdx[0]+400, len(body))
+			}
+			fullRHS := body[fIdx[0]:end]
 			if target, rawFKString, isSelf := djangoRelTargetDetail(fullRHS, className); target != "" {
 				rel := referencesClassEdge(file.Path, className+"."+attr, target, "django", attr)
 				// #6986 — carry the RELATION-SHAPE properties the core
@@ -1042,6 +1071,79 @@ func djangoRelTargetDetail(rhs, ownerClass string) (target, rawFKString string, 
 func djangoRelTarget(rhs, ownerClass string) string {
 	target, _, _ := djangoRelTargetDetail(rhs, ownerClass)
 	return target
+}
+
+// djangoDeclEnd returns the index one past the `)` that closes the call whose
+// opening `(` sits at openPos, or -1 when the source is unbalanced within body.
+//
+// #6990 — this is what bounds a relational field's target scan to the field's
+// OWN declaration. A byte window cannot do that: Django field declarations
+// legitimately span lines, so any fixed count is either too short for a real
+// multi-line `ForeignKey(\n  "app.Model",\n  on_delete=...\n)` or long enough to
+// reach the NEXT field's constructor. Following the parentheses is exact in
+// both directions.
+//
+// It counts parenthesis depth only — a `)` cannot appear inside `[...]` or
+// `{...}` without its own `(` — and it skips the two places a `)` can appear
+// without closing anything: string literals (single, double, and triple-quoted,
+// honouring backslash escapes) and `#` comments. An unterminated single-quoted
+// string or comment yields -1 rather than a guess, so the caller falls back to
+// the pre-#6990 byte window instead of over-reading further than it used to.
+func djangoDeclEnd(body string, openPos int) int {
+	if openPos < 0 || openPos >= len(body) || body[openPos] != '(' {
+		return -1
+	}
+	depth := 0
+	for i := openPos; i < len(body); i++ {
+		switch body[i] {
+		case '(':
+			depth++
+		case ')':
+			depth--
+			if depth == 0 {
+				return i + 1
+			}
+		case '#':
+			nl := strings.IndexByte(body[i:], '\n')
+			if nl < 0 {
+				return -1
+			}
+			i += nl
+		case '\'', '"':
+			j := skipPyStringLiteral(body, i)
+			if j < 0 {
+				return -1
+			}
+			i = j - 1
+		}
+	}
+	return -1
+}
+
+// skipPyStringLiteral returns the index one past the string literal that starts
+// at i (which must hold a quote byte), or -1 if it is unterminated. Triple
+// quotes are handled first so a `)` inside a docstring-style default cannot be
+// mistaken for a closing paren.
+func skipPyStringLiteral(s string, i int) int {
+	q := s[i]
+	if i+2 < len(s) && s[i+1] == q && s[i+2] == q {
+		quote := s[i : i+3]
+		if end := strings.Index(s[i+3:], quote); end >= 0 {
+			return i + 3 + end + 3
+		}
+		return -1
+	}
+	for j := i + 1; j < len(s); j++ {
+		switch s[j] {
+		case '\\':
+			j++
+		case '\n':
+			return -1
+		case q:
+			return j + 1
+		}
+	}
+	return -1
 }
 
 // extractBalancedBrackets returns the content of a [...] list starting at openPos.
--- a/internal/custom/python/django_decl_window_6990_test.go
+++ b/internal/custom/python/django_decl_window_6990_test.go
@@ -1,0 +1,306 @@
+package python
+
+import (
+	"fmt"
+	"strings"
+	"testing"
+
+	"github.com/cajasmota/grafel/internal/types"
+)
+
+// Issue #6990, residue 2 — the relational-target scan ran forward a raw 400
+// BYTES from the matched field, unbounded by the declaration it belonged to. A
+// field whose own target argument the regex cannot parse therefore captured the
+// NEXT field's target. `LogEntry.user`, declared
+// `models.ForeignKey(settings.AUTH_USER_MODEL, ...)`, came out targeting
+// `ContentType` — a real model, a bound edge, the wrong thing.
+//
+// The window is now bounded by the field's OWN call parentheses
+// (`djangoDeclEnd`). The regex is UNCHANGED: this ships ONE guard, so it stays
+// graded (#6988's lesson — two guards that only fire together grade neither).
+//
+// SCOPE. Residue 1 of #6990 — `class CTForeignKey(GenericForeignKey)` passing
+// `isDjangoRelationalField`'s suffix check at django.go — is NOT touched here.
+// It is a policy decision (deny-list vs allow-list) that #6988/#6989 settled
+// once, and it is pinned by TestIssue6988_RegexBoundaryRejected_WouldDropSubclasses.
+//
+// GATE (#6966): every edge below exists only with the custom Python lane ON.
+// It is off by default (`WithCustomExtractors`), so default-index baselines are
+// untouched. Graph-level incidence of the defect on the real django corpus is
+// NOT ESTABLISHED and is not claimed anywhere in this file; what is graded here
+// is extractor-level behaviour.
+
+// ---------------------------------------------------------------------------
+// The bound itself: djangoDeclEnd.
+// ---------------------------------------------------------------------------
+
+// TestIssue6990_DeclEnd_FindsTheCallsOwnCloser grades every place a `)` can
+// appear without closing the call. Only the unit level can observe these: the
+// target is the FIRST argument, so an EARLY close still contains it and is
+// invisible end-to-end. A LATE close is the bug.
+func TestIssue6990_DeclEnd_FindsTheCallsOwnCloser(t *testing.T) {
+	cases := []struct {
+		name string
+		body string // `|` marks the expected end (one past the closing paren)
+	}{
+		{"flat", `f(A)|`},
+		{"nested call", `f(A, on_delete=models.CASCADE)|`},
+		{"double nested", `f(A, default=(1, (2, 3)))|`},
+		{"paren in double-quoted string", `f(A, help_text="see (the docs)")|`},
+		{"paren in single-quoted string", `f(A, help_text='see (the docs)')|`},
+		{"paren in triple-quoted string", `f(A, help_text="""see (the
+docs)""")|`},
+		{"paren in comment", `f(A,  # a ) here
+    on_delete=CASCADE)|`},
+		{"hash inside string is not a comment", `f(A, default="#)")|`},
+		{"apostrophe inside comment is not a string", `f(A,  # don't ) do this
+    x=1)|`},
+		{"escaped quote inside string", `f(A, default="a\")b")|`},
+		{"multi-line with trailing comma", `f(
+    "catalog.Author",
+    on_delete=models.CASCADE,
+)|`},
+		{"stops at its OWN closer, not a later one", `f(A)| g(B)`},
+	}
+	for _, c := range cases {
+		t.Run(c.name, func(t *testing.T) {
+			want := strings.IndexByte(c.body, '|')
+			if want < 0 {
+				t.Fatalf("fixture %q has no `|` end marker", c.body)
+			}
+			body := strings.Replace(c.body, "|", "", 1)
+			if got := djangoDeclEnd(body, 1); got != want {
+				t.Errorf("djangoDeclEnd(%q, 1) = %d, want %d (slice %q, want %q) (#6990)",
+					body, got, want, body[:max(got, 0)], body[:want])
+			}
+		})
+	}
+}
+
+// TestIssue6990_DeclEnd_RefusesToGuess pins the -1 side. The caller falls back
+// to the pre-#6990 400-byte window there, which is EXACTLY the old behaviour —
+// never wider. A guess (end-of-body, say) would be strictly worse than the bug
+// this fixes, so "unbalanced" must be reported, not papered over.
+func TestIssue6990_DeclEnd_RefusesToGuess(t *testing.T) {
+	cases := []struct{ name, body string }{
+		{"never closed", `f(A, on_delete=CASCADE`},
+		{"unterminated string", `f(A, help_text="oops)`},
+		{"unterminated triple-quoted string", `f(A, help_text="""oops)`},
+		{"unterminated trailing comment", `f(A,  # oops`},
+		{"depth never returns to zero", `f((A)`},
+	}
+	for _, c := range cases {
+		t.Run(c.name, func(t *testing.T) {
+			if got := djangoDeclEnd(c.body, 1); got != -1 {
+				t.Errorf("djangoDeclEnd(%q, 1) = %d, want -1 — an unbalanced source must fall back "+
+					"to the byte window, not guess an end (#6990)", c.body, got)
+			}
+		})
+	}
+	// openPos must actually name a `(`.
+	for _, pos := range []int{-1, 0, 99} {
+		if got := djangoDeclEnd(`f(A)`, pos); got != -1 {
+			t.Errorf("djangoDeclEnd(%q, %d) = %d, want -1", `f(A)`, pos, got)
+		}
+	}
+}
+
+// ---------------------------------------------------------------------------
+// End to end. Two axes crossed: DECLARATION LAYOUT x TARGET FORM.
+// ---------------------------------------------------------------------------
+
+// declLayout6990 renders one field declaration three ways. The argument text is
+// identical in all three; only the line structure varies. RE2's `\s` matches
+// `\n`, so multi-line coverage in this package has been INCIDENTAL and never
+// asserted — which is exactly how a narrowing walks through a pin.
+type declLayout6990 struct {
+	name   string
+	render func(attr, ctor, args string) string
+}
+
+var declLayouts6990 = []declLayout6990{
+	{"single-line", func(attr, ctor, args string) string {
+		return fmt.Sprintf("    %s = %s(%s, on_delete=models.CASCADE)", attr, ctor, args)
+	}},
+	{"multi-line", func(attr, ctor, args string) string {
+		return fmt.Sprintf("    %s = %s(\n        %s,\n        on_delete=models.CASCADE)", attr, ctor, args)
+	}},
+	{"multi-line-trailing-comma", func(attr, ctor, args string) string {
+		return fmt.Sprintf("    %s = %s(\n        %s,\n        on_delete=models.CASCADE,\n    )", attr, ctor, args)
+	}},
+}
+
+// TestIssue6990_TargetFormsSurviveEveryDeclarationLayout crosses the two axes.
+//
+// VARIED: declaration layout (single-line / multi-line / multi-line with a
+// trailing comma) x target form (dotted lowercase `settings.AUTH_USER_MODEL`,
+// bare CamelCase symbol, quoted `"app.Model"`, quoted `'self'`).
+//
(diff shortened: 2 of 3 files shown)


cajasmota:
## Coordinator mutants on `067686eb2` — **CU-1 DEAD, CU-2 ALIVE and cheap to grade**

Own detached worktree at `/private/tmp/.../coord-mut-7026-z5r/wt` (absolute), private script dir. Green baseline immediately before each. Restored by `cp` from a SHA-256-verified backup (`92ceb64b…` both ways); `git diff HEAD` and `git status --porcelain` empty afterwards.

Your eight mutants are all on `djangoDeclEnd`'s internals and on the fallback's **width**. So I went after the two things that sit either side of that: whether the bound is *reached* at all, and whether the fallback's **positive** behaviour is graded.

### CU-1 — make the bound silently inert (the anchor offset)

`djangoDeclEnd(body, fIdx[1]-1)` → `djangoDeclEnd(body, fIdx[0])`. Occurrence asserted 1 before and after, `go vet` clean.

`fIdx[0]` is the start of the assignment line, not the constructor's `(`, so the guard at the top of `djangoDeclEnd` (`body[openPos] != '('`) returns **-1 on every field**, and every declaration falls through to the 400-byte window. The whole PR becomes a no-op while still compiling, still calling the new function, and still carrying all its unit tests.

**Result: DEAD**, by three tests:

```
--- FAIL: TestIssue6990_TargetFormsSurviveEveryDeclarationLayout
--- FAIL: TestIssue6990_NamedResidue2_RealFieldFollowedByGFK
--- FAIL: TestPythonFieldTargetType_6990_NoOverReadIntoTheNextFieldsTarget
```

This is the one I most wanted, because extracting a decision into a helper and unit-testing the helper does **not** pin the behaviour — the call site stays free to ignore it, and that exact shape has shipped here before. It does not: the 3×4 cross and the converted pin both observe the bound end-to-end, not just `djangoDeclEnd` in isolation. Converting the known-wrong pin into a positive assertion rather than deleting it is what makes the third failure line possible.

### CU-2 — narrow the fallback to nothing

`end = min(fIdx[0]+400, len(body))` → `end = fIdx[0]`. Occurrence asserted 1, `go vet` clean.

**Result: ALIVE.** `./internal/custom/python/` fully green.

So on unbalanced source, a build that finds **no relational target at all** is indistinguishable from the shipped one. Every Django field in an unbalanced file silently loses its target edge and nothing notices.

**This is the sibling of your own mutant 2, and the test name says so.** `TestIssue6990_UnbalancedSourceFallsBackToTheOldByteWindow_NotWider` grades exactly one direction — it places a target beyond 400 bytes and asserts it is **not** captured. It even carries a good premise control (`djangoDeclEnd = -1`, so the fixture provably exercises the fallback branch). What no test asserts is that a target **within** 400 bytes on unbalanced source is still found. Widening is graded; narrowing to zero is free.

**Consequence for the PR body's justification.** The stated reason to keep the fallback is that it is "exactly the old behaviour, never wider". The first half of that claim is **not currently pinned** — under this suite, "the old byte window" and "nothing at all" are the same fallback. That matters more than a usual ungraded axis, because the fallback's whole reason to exist is that it preserves prior behaviour rather than dropping edges.

**Priced at ~6 lines**, and the fixture already does the work: add a second, *near* target inside the existing unbalanced fixture (within 400 bytes of the field) and assert it **is** captured, alongside the far one that must not be. Same test, one more row and one more assertion. I would take it rather than defer it — "distinguishable and ungraded" on the exact path a reviewer is independently questioning is not a good thing to ship.

### On the layering

- **The bound is reached, and reached from the right anchor** → CU-1.
- **The bound stops in the right place** → your eight, across the layout × target-form cross.
- **The fallback is never wider** → your mutant 2, killed rather than deferred.
- **The fallback is never narrower** → **nothing**. CU-2.

**You wrote the killing test rather than deferring an ALIVE mutant.** Your mutant 2 was alive and distinguishable and you graded it in the same pass. That is the behaviour this board asks for and rarely gets — "too expensive to kill" is a claim needing a number, and you never made it.

**You verified the pin fired its `t.Skip` before converting it.** That is the step that separates "the fix works" from "the test no longer looks". A pin that silently starts skipping reads exactly like a pin that passes.

Independent review is running and will reach the fallback from the other side — whether it should exist at all, given `-1` returns the code to the behaviour this PR removes. CU-2 sharpens that question rather than answering it: if the fallback's positive behaviour is worth keeping, it needs an assertion; if it is not, the honest fallback is to emit nothing deliberately.

cajasmota:
## CU-2 killed — `aa220a250`

You were right, and the diagnosis was exactly the useful one: the PR body justified keeping the fallback as *"exactly the old behaviour, never wider"*, and only the **never-wider** half was pinned. `end = fIdx[0]` — an empty window that finds nothing at all on any unbalanced file — was indistinguishable from what I shipped. The half going unobserved was the half the fallback exists for.

**Extended the existing fixture, no new test.** `TestIssue6990_UnbalancedSourceFallsBackToTheOldByteWindow` now runs two genuinely-unbalanced declarations:

- **FAR** (unchanged): `Owner.subject` unterminated, >400 bytes of filler, then a parseable `ForeignKey(Sentinel)` — asserted **not** captured.
- **NEAR** (new): `NearOwner.near = models.ForeignKey(Nearby,`, equally unterminated so it takes the same `end < 0` branch, with its own target immediately after its own `(` — asserted **captured**, exactly one edge, to `Nearby`.

The `djangoDeclEnd == -1` premise control is now a loop over **both** declarations, so neither half can quietly be grading the normal path. The FAR half keeps its `Owner.sentinel` positive control.

Renamed off `..._NotWider` — that name now under-describes what it grades. Nothing else moved; the fix itself is byte-identical to `067686eb2`.

## Re-scored, both against the new test

| Mutant | Verdict | Dies on |
|---|---|---|
| **CU-2** — `end = min(fIdx[0]+400, len(body))` → `end = fIdx[0]` | **DEAD** | `NearOwner.near emitted 0 field_target_type edge(s), want exactly 1 -> ...:Nearby` |
| **My mutant 2** — same line → `end = len(body)` | **DEAD** (re-scored, still) | `Owner.subject emitted 1 ... first ToID=...:Sentinel ... never WIDER` |

Each dies on its **own** half and on a different assertion, so the two directions are graded separately rather than by one compound row. Applied one at a time, occurrence count asserted `== 1` before running, `cp`-restored from a shasum-verified backup with `git diff HEAD` empty after. `-count=1` throughout.

`go test -count=1 ./internal/custom/... ./internal/extractors/python/... ./internal/resolve/...` green; `go vet` and `gofmt -l` clean.

## Held to scope

Residue 1 untouched. No graph-level incidence number claimed. I did **not** reopen whether the fallback should exist at all — the test grades the behaviour as shipped, so if the independent review comes back arguing it should deliberately emit nothing, that flips one assertion in one place and is a clean second round.

cajasmota:
# APPROVE

Independent adversarial review. Own worktree, scored against `067686eb2` and re-scored against the current head `aa220a250` (`django.go` is byte-identical between them — shasum `1057eade…` — so nothing below is stale on the production file; every test verdict is from `aa220a250`).

Verification I ran myself, `-count=1` throughout: `gofmt -l` clean, `go vet` clean, `./internal/custom/python/` green, `./internal/custom/... ./internal/extractors/python/... ./internal/resolve/...` green. Backup shasum re-verified and `git status --porcelain` empty after every mutant.

**No blocking defect.** Four non-blocking findings, each with a price.

---

## 1. The fallback to the known-buggy path — I think keeping it is right

I enumerated the `-1` paths straight off the code rather than off the PR body: (a) invalid `openPos`; (b) depth never returns to 0 before end-of-body; (c) `#` with no trailing `\n`; (d) unterminated triple-quote; (e) single-quoted string hitting `\n`; (f) single-quoted string hitting end-of-body. The author's five map onto these. **The list is complete.**

Every one of them requires source that is not valid Python — with one exception worth recording: (e)/(b) are reachable from *valid* Python through `extractClassBody` (django.go:1296), which cuts the class body at the first line whose indent ≤ the class indent. A field with a triple-quoted default containing a column-0 line truncates the body mid-declaration. That is pre-existing and unchanged by this PR, and the fallback is the correct behaviour there.

On that input the fallback is **byte-identical to `main`**, so it cannot regress anything; it merely declines to improve. The alternative — emit nothing — would be a behaviour change on a path this fix does not otherwise touch, and would drop legitimately-parseable targets (exactly the `NearOwner.near` shape now pinned). **Keep it as shipped.**

I scored the empty-window mutant (`end = fIdx[0]`) at `067686eb2`: **ALIVE**. At `aa220a250`: **DEAD**, on `NearOwner.near`, on a different assertion and a different declaration from the widening half. That gap is closed.

## 2. The constant `400` is graded by direction but not by magnitude — non-blocking

New mutants on the fallback constant, all scored at `aa220a250`:

| Fallback | Verdict |
|---|---|
| `+700` | DEAD |
| `+600` | **ALIVE** |
| `+120` | **ALIVE** |
| `fIdx[0]` (0) | DEAD |

So anything in roughly `(120, 700)` passes. The FAR target sits ~640 bytes out and the NEAR target ~30 bytes in, leaving ~300 bytes of slack on each side. The *directions* are graded; the sentence "deliberately **exactly** the old behaviour" is not — a build with a 600-byte fallback is indistinguishable from this one, and 600 is not the old behaviour.

**Price to tighten: ~4 lines** — shrink the FAR filler so its target starts at ~405 bytes and pad the NEAR declaration so its target starts at ~390. That pins 400 to ±15 in both directions with no new test.

## 3. Mutants the author did not run — the too-early direction is genuinely graded

Seven new mutants beyond the author's eight. Each edit asserted to land by exact occurrence count before running; restored by `cp` with shasum re-verified.

| Mutant | Verdict |
|---|---|
| Drop paren-depth tracking (`case ')': return i+1`) | **DEAD** — `double nested`, `depth never returns to zero`, and the fallback premise control |
| `i = j` instead of `i = j - 1` after a string (swallows a `)` immediately after a literal) | **DEAD** |
| Triple-quote closer off-by-3 (`return i+3+end`) | **DEAD** |
| Drop the `body[openPos] != '('` guard | **DEAD** |
| Fallback `+700` / `fIdx[0]` | **DEAD** (see §1, §2) |
| Fallback `+600` / `+120` | **ALIVE** (see §2) |
| Delete `case '\n': return -1` in `skipPyStringLiteral` | **ALIVE** |

On the last one: every `RefusesToGuess` string fixture is *also* unterminated at end-of-body, so removing the newline guard changes nothing on any of them. Distinguishing input: `f(A, x="oops` + newline + `)")` — shipped returns `-1`, mutant returns an index. Verdict: **distinguishable and ungraded, but not production-reachable on valid Python** (a single-quoted literal cannot span a line). Price to grade: **one row** in `TestIssue6990_DeclEnd_RefusesToGuess`.

The too-early direction I was asked to weight hardest is covered: decorators between fields don't touch paren balance and the field regex is per-line; `on_delete=models.CASCADE` has no nesting and is subsumed by the double-nested row; `related_name="x)y"` is the `paren in double-quoted string` row; `#` inside a string and an apostrophe inside a comment each have their own row.

## 4. The skipper against real Python — one gap, and it fails benign

- **Raw strings**: `\` still terminates-escapes for the tokenizer inside `r"…"`, and `r"…\"` is a Python syntax error, so the escape branch cannot produce a false ending. Correct as written.
- **Implicit concatenation** (`"a" "b"`): each literal is skipped independently. Correct.
- **`b`/`f`/`rb` prefixes**: only the quote byte triggers the branch, so prefixes are inert. Correct.
- **Empty string `""` adjacent to a quote**: the triple-quote test is `s[i+1]==q && s[i+2]==q`, which I walked for `x=""`, `x="" "y"`, `x="""` at end-of-body — all correct.
- **Gap: PEP 701 f-strings reusing the outer quote** (Python 3.12+), e.g. `help_text=f"{d["k"]}"`. The skipper ends the literal at the inner quote and paren counting resumes over string text. I traced the consequence: the window ends *early*, or `-1`. Since the target is always the first positional / `to=` argument, an early end still contains it — this **cannot** reproduce the #6990 over-read. Benign; worth one sentence in the doc comment, not a fix.

## 5. Test bodies read against their names

The "every row carries a sentinel sibling" claim **holds** — I checked it row by row: the 3×4 cross asserts the sentinel with `t.Fatalf` on every row including the negatives; `NamedResidue2` has `Note.content_type`; the converted pin has `sawContentType`; the fallback test has `Owner.sentinel`.

Two things to fix:

- **One label overclaims.** `{"nested call", "f(A, on_delete=models.CASCADE)"}` contains no nested call and no nested parens — it is structurally identical to the `flat` row. Single-level nesting is therefore never exercised (double-level is, and subsumes it), so this is a naming defect rather than a coverage hole. Fix: `f(A, on_delete=models.SET(get_sentinel))`.
- **Nothing asserts the field under test still exists** on the negative rows. `Owner.subject`, `Note.user` and `LogEntry.user` are all asserted to emit zero `field_target_type` edges; the sibling control proves the extractor ran, but a change that dropped `subject`/`user` as a field entity outright would pass every one of them. The conversion also removed the old "lone field" premise control. `NamedResidue2` has a CONTAINS assertion for `content_object` but not for `user`. **Price: ~3 lines** — one `hasEdge6988(rels, pyClassRef(...), "<owner>.<field>", Contains)` per negative test.

## 6. Both deliberate non-actions check out

- **Residue 1 untouched.** `isDjangoRelationalField` is byte-identical to `main` — the diff touches only the comment block above `djangoModelRelTargetRe`, the caller's window, and two new functions. No partial behaviour change.
- **No graph-level incidence number** appears in the PR body, the commit messages, or any new test comment. The unverified `ext:django:` interception claim is explicitly retracted rather than repeated or contradicted. Correct handling.

## 7. The `./cmd/grafel/` argument: right conclusion, wrong reason

The PR says "nothing here reaches the digest-pinned graph test". That is not accurate — `cmd/grafel/inproc_custom_extractors_5989_test.go:93` and `cmd/grafel/generated_seam_6329_test.go:51` both drive this django custom extractor through the CLI path.

The conclusion still holds, for a different reason I checked directly: both fixtures are **single-line, balanced declarations whose target is the first argument**, so the bound is a no-op on them. The repo's only multi-line django fixture, `testdata/fixtures/real-world/python/django_models.py` (lines 71, 126, 175), is likewise balanced with the target first, and is consumed by `internal/extractors/python`, which I ran green. Please correct the sentence rather than the decision.

---

**Safe to merge.** Findings 2, 3 (`\n` guard), and 5 are cheap and worth taking; none of them blocks.

cajasmota:
## All four items done — `ffad6e6e8`. Fix unchanged; tests and prose only.

### 1. The `400` is now pinned by MAGNITUDE, not sign

You were right that this is CU-2 one level down. `TestIssue6990_UnbalancedSourceFallsBackToTheOldByteWindow` now carries two **byte-exact** rows built by a `boundaryClass(class, field, matchEnd)` helper that pads a class so the *following* field's target ends at a chosen offset:

- `FarOwner.subject` — target ends at byte **401** → must **not** be captured.
- `EdgeOwner.subject` — target ends at byte **400** → **must** be captured.

Together those admit **exactly one** window width. `NearOwner.near` (a field's own target right after its own `(`) is kept as the simplest not-narrower row.

One implementation note worth recording: the target is the single character **`Q`** on purpose. A longer name still matches as a *truncated prefix* when the window cuts through it — `[A-Z][A-Za-z0-9_]*` is perfectly happy with `Boundar` — which would have made the "boundary" fuzzy by the length of the name and quietly reintroduced the slack I was trying to remove. One character makes "the match fits" and "does not fit" adjacent.

### 2. `case '\n': return -1` — graded, not argued

Added to `RefusesToGuess`: `f(A, x="oops\ny = ")`. The quote *is* closed, but only on a later line, so the literal is unterminated and the scan must refuse. Without the rule the skipper runs to line 2's quote and then reports the `)` after it as this call's closer. Mutant 8 **DEAD**. I did not take the "unreachable, so exempt" route — it was one line.

### 3. Fixture label, and the missing half of the vacuity guard

- `"nested call"` now contains real nesting — `f(A, default=make())`. Its old body survives under an accurate name, `"call with a dotted kwarg, no nesting"`, so nothing is lost.
- **Every negative row now also asserts the field still exists.** `Owner.subject` (all three layouts of the `settings.AUTH_USER_MODEL` row), `Note.user`, `FarOwner.subject` via a `CONTAINS` edge; `LogEntry.user` in the converted pin via the entity in the merged set. You are right that this was the sentinel-sibling guard missing on its other half.

### 4. `./cmd/grafel/` — corrected, and verified rather than restated

Confirmed both by reading: `cmd/grafel/inproc_custom_extractors_5989_test.go:93` and `cmd/grafel/generated_seam_6329_test.go:51` are the two `models.ForeignKey` fixtures, and both are single-line, balanced, target-first — so `djangoDeclEnd` returns the same slice the byte window did and the bound is a **no-op** on them. That is now the stated reason, replacing "the diff is contained to `internal/custom/python/`". The PR body also now carries your justification for keeping the fallback (byte-identical to `main` on every `-1`, and one `-1` path reachable from valid Python via `extractClassBody`'s indent cut) in place of the weaker one I wrote, in both the body and the test's doc comment.

### Re-scored — 12/12 DEAD

Every mutant re-run against the **edited** tests, since I changed what grades the fallback.

| Mutant | Verdict | Dies on |
|---|---|---|
| 1 — revert the bound to an unconditional 400 bytes | DEAD | 5 tests |
| 2 — `end = len(body)` | DEAD | `FarOwner.subject ... never WIDER than 400` |
| CU-2 — `end = fIdx[0]` | DEAD | `NearOwner.near ... want exactly 1 -> Nearby` |
| **C-120** — constant `400`→`120` (was ALIVE) | **DEAD** | `EdgeOwner.subject emitted 0 ... not a narrower one` |
| **C-600** — `400`→`600` (was ALIVE) | **DEAD** | `FarOwner.subject emitted 1 ... ToID=...:Q` |
| C-700 — `400`→`700` | DEAD | same row |
| 3 / 4 / 5 / 6 / 7 — skipper, comment skip, off-by-one, triple-quote, escape | DEAD | `DeclEnd` subtests |
| **8** — delete `case '\n': return -1` (was ALIVE) | **DEAD** | `quote closed only on a later line` |

Each dies on its own row and on a different assertion — no compound row is doing the grading. Applied one at a time, occurrence count asserted `== 1` before each run, `cp`-restored from shasum-verified backups, `git diff HEAD` empty after every one, `-count=1` throughout.

`go test -count=1 ./internal/custom/... ./internal/extractors/python/... ./internal/resolve/...` green; `go vet` and `gofmt -l` clean.

Residue 1 untouched. No incidence number claimed. Fallback-should-exist not reopened — recorded as settled, keep.

PR #7027, opened 2026-09-08
fix(#7025): the 10 files hiding 49 import_markers are unreachable by PATH and by SHAPE — #7024's rename is a regression there, not a rescue

### LINKED ISSUE #7025: CORRECTED: 10 rule files carry 49 inert import_markers because they are OUTSIDE loader scope (2-segment path), not because the container is misspelled — and the #7024 rename would REGRESS them (sequence vs mapping)
#7024 fixed 14 misspellings of the `frameworks:` container and made 52 files' `import_markers` readable. **The same defect is still live in 10 more files carrying 50 `import_markers` blocks**, under container spellings the closed set does not contain.

This is the review's MR2 finding, and it is substantially larger than MR2 described it. MR2 said "a 15th spelling on a marker-**less** file". Measured on merged `main` (`e2b54fef0`), the 15th-and-beyond spellings are on marker-**bearing** files.

## The 10 files

| file | top-level key(s) | `import_markers` blocks |
|---|---|---|
| `clojure/test_patterns.yaml` | `testing_frameworks` | 4 |
| `elixir/extras.yaml` | `domain_roots`, `orms_db` | 5 |
| `elixir/test_patterns.yaml` | `testing` | 6 |
| `go/test_patterns.yaml` | `testing_frameworks` | 7 |
| `lua/test_patterns.yaml` | `testing` | 3 |
| `protobuf/extras.yaml` | `detection`, `domain_roots`, `extraction_targets`, `package_patterns`, `toolchains` | 3 |
| `python/complexity.yaml` | `complexity_signals` | 6 |
| `python/test_patterns.yaml` | `testing_frameworks` | 7 |
| `rust/test_patterns.yaml` | `test_conventions`, `testing` | 8 |
| `zig/test_patterns.yaml` | `testing` | 1 |

**Not one of these files carries a single modelled top-level key** (`frameworks`, `source_patterns`, `relationship_rules`, `file_conventions`). So the markers are not sitting beside readable content — the whole file is outside the schema, exactly as the 247 were.

**What is established:** these `import_markers` cannot reach `frameworkPresent`. #7024 established by rename-to-break — renaming `Frameworks`/`Name`/`ImportMarkers` and compiling with `-gcflags=all=-e` — that markers have exactly one non-test consumer, `detector.go:153`, reached only through the `frameworks:` container. Markers under `testing:` or `complexity_signals:` cannot get there.

**What is NOT established, and blocks a naive fix:** whether these containers are inert. #7024 proved its 247 were (`sourcePatternEntries=0` etc. across all four modelled entry types). **I have not measured that here, and I do not expect it to hold** — `testing_frameworks` and `complexity_signals` read like keys some other pass consumes off disk. `internal/entkinds`, `internal/relkinds` and `tools/coverage` all walk this tree at arbitrary YAML depth. **Renaming these to `frameworks:` without first establishing who reads them would be a behaviour change, not a rescue.** That is the difference between this and #7024, and it is why this is a separate issue rather than a second commit on the same one.

## Correction: the top-level strict guard is not free

The #7024 review concluded that after the rename the tree has "exactly four distinct top-level keys, all modelled", making #7014's top-level strict guard cost **0 red files** and subsume the hand-maintained 14-item closed list. **That does not reproduce.** On `e2b54fef0`, across all 714 rule files:

- **66 distinct top-level keys**, not 4.
- **193 files carry at least one unmodelled top-level key** — 122 ordinary rule files, 31 `_manifest.yaml`, 40 under `_engine/`.

So the guard costs **193 red**, and the `_manifest.yaml` and `_engine/` files are a genuinely different schema that a rules-tree guard should probably not be judging at all. The "0 red, take it now while it's free" framing is wrong; the guard is still worth building, but it needs a scope decision first (which file classes it governs), and it is not a cheap opportunistic win.

I measured the four-key claim because a count in an issue body is an assertion. The reviewer's number was most likely taken over the `frameworks:`-bearing subset (538 of the 714 files) rather than the whole tree.

## Why this matters beyond the markers

The defect class is **an unmodelled top-level key silently swallowing authored content, with no error**. #7024 removed 14 instances of it. There are 66 top-level keys in this tree and 4 are modelled. Until the loader refuses what it does not model, "the gate is inexpressible" (#7013) and "the gate is expressible and unused" stay indistinguishable — which is the exact measurement #7013 step 2 needs to make.

## Suggested order

1. Establish who reads `testing_frameworks` / `testing` / `complexity_signals` / `domain_roots` / `extraction_targets` off disk — rename-to-break, the method #7024 used, not grep.
2. Only then decide rescue-by-rename vs. model-the-container-properly, per key.
3. Scope and land the top-level strict guard (#7014) so the 15th spelling cannot be introduced silently again.

Refs #7013, #7014, #7024, #6927.

cajasmota:
**Tree measured, by absolute path:** `/private/tmp/claude-501/-Users-jorgecajas-Projects-archigraph/c0fefed4-0195-4c81-9a94-07a7145710a4/scratchpad/impl-7025-q7m/wt`, branch `lane/7025-hidden-markers`, parent `e2b54fef0`. macOS 25.6.0 / arm64 / APFS, host Go toolchain `go1.26.4 darwin/arm64`.

## The headline: the rescue #7025 specifies is not available, and applying it would be a regression

The brief's step 2 was "rename the container to `frameworks:` so the markers become readable, exactly as #7024 did". Measured before doing it. It cannot work here, for **two** reasons, and **neither is the container spelling**:

1. **THE PATH.** `LoadAllRulesFromFS` (`loader.go:113-127`) accepts exactly `<lang>/<subdir>/<file>.yaml` for `subdir ∈ ruleSubdirs = {frameworks, orms, queues}`. All ten #7025 files sit at `<lang>/<file>.yaml` — **two** path segments. The loader never opens them, never lists them in `RuleLoadReport.Candidates`, and never decodes them at all. **Demonstration:** renaming `go/test_patterns.yaml`'s `testing_frameworks:` → `frameworks:` and re-running a harness that dumps `engine.LoadAllRules()` as JSON produced a **zero-byte diff**. The correctly-spelled container is still invisible. The container key was never the blocker.

2. **THE SHAPE.** `FrameworkRule.Frameworks` is a single `FrameworkMeta` **mapping** (`schema.go:19`). All ten #7025 containers are **sequences** of framework entries (`- name: clojure.test` …). So inside loader scope the same rename does not decode — it *fails* to decode: `yaml.Unmarshal` errors, `loader.go:141` records a parse failure, and the file is dropped **along with any rules it does carry**. On these files the #7024 move is not a no-op, it is a regression.

That is the difference #7025 was opened to find, and it is a different one from the one it predicted.

## Step 1 — who reads these containers, per key

**Method, per the brief: rename-to-break, not grep.** All **16** top-level keys on the ten files were renamed in place to `zzprobe_<key>:` (each edit proved landed by an exact `assert n == 1` occurrence count before substitution, 16/16), then every consumer that reaches the rule tree from disk or embed was re-run and its **output** compared, not just its exit code.

**How the consumer set was established** (this is the part a grep only lower-bounds): the only code that opens files under `internal/engine/rules/` is `internal/engine/loader.go` (`//go:embed all:rules`) and `tools/coverage/discover.go` (`filepath.Join(repoRoot, "internal", "engine", "rules", …)`). `internal/entkinds` and `internal/relkinds` walk whatever root they are handed and are pointed at the repo root by their own tests. Non-Go: nothing under `scripts/`, `.github/` or `Makefile` reads these keys; `docs/coverage/registry.json` *cites* `<lang>/test_patterns.yaml` as evidence in 8 records but is data, not a decoder (see "what this exposes" below).

| container key | file(s) | carries markers? | decoded by | evidence |
|---|---|---|---:|---|
| `testing_frameworks` | clojure, go, python `test_patterns.yaml` | yes (4+7+6) | **nothing** | rename → engine rule-set, entkinds, relkinds, coverage all identical |
| `testing` | elixir, lua, rust, zig `test_patterns.yaml` | yes (6+3+8+1) | **nothing** | same |
| `complexity_signals` | `python/complexity.yaml` | yes (6) | **nothing** | same |
| `orms_db` | `elixir/extras.yaml` | yes (5) | **nothing** | same |
| `toolchains` | `protobuf/extras.yaml` | yes (3) | **nothing** | same |
| `test_conventions` | `rust/test_patterns.yaml` | **no** | **nothing** | same |
| `domain_roots` | elixir + protobuf `extras.yaml` | **no** | **nothing** | same |
| `detection` (top-level) | `protobuf/extras.yaml` | **no** | **nothing** | same |
| `extraction_targets` | `protobuf/extras.yaml` | **no** | **nothing** | same |
| `package_patterns` | `protobuf/extras.yaml` | **no** | **nothing** | same |

Per-consumer result under the 16-key rename:

- **`internal/engine`** (embedded loader): `LoadAllRules()` serialised to JSON — **byte-identical**. Structural reason: the files are not loader-scoped, so the key is never even looked at.
- **`internal/entkinds`** (walks the YAML at **any depth**, and computes `Site.Bound` **from the YAML path**, so a top-level rename is exactly the kind of change that could move it): full `Site` list (`file|line|kind|origin|path|bound`) + `UnresolvedSites` + parsed-file counts — **identical**. None of the ten declares `entity_type:` or `entity_mapping:` anywhere.
- **`internal/relkinds`**: reads top-level `relationship_rules:` only; none of the ten has one. Site list **identical**.
- **`tools/coverage`**: `Discover(repoRoot, "")` marshalled to JSON — **identical** once its own map-iteration nondeterminism is normalised away (see below). Its only use of `test_patterns.yaml` is `strings.Contains(strings.ToLower(fileBytes), needle)` for a framework **name** (`discover.go` `testFrameworkWalker`), which is blind to key names by construction.
- Package suites, `-count=1`: `./internal/engine ./internal/entkinds ./internal/relkinds ./tools/coverage ./internal/types` — all **ok** with the mutant applied.

**Positive controls, because "no diff" from an insensitive probe is worthless:**

- Appending `probe_block: {entity_type: ProbeKind7025}` to `go/test_patterns.yaml` **did** appear in the entkinds output (`ENT internal/engine/rules/go/test_patterns.yaml|151|ProbeKind7025|rule_yaml|probe_block.entity_type|false`). So entkinds **does** read these files at arbitrary depth and its verdict **does** depend on their content — it simply does not decode the container key.
- **`tools/coverage.Discover` is nondeterministic.** Its first run under the mutant differed from baseline by 24 lines (a `synthesizeSlim` / PHP evidence block reordering between two candidates). Two further runs of the *same* mutated tree were byte-identical to baseline, and **two runs of the same tree differed from each other** — so the diff is map ordering, not signal. Comparing sorted line multisets: baseline == mutant, exactly. Recorded because a 24-line diff read at face value would have been a false positive on a language (PHP) not among the ten.

**Verdict: all 10 keys are inert**, exactly as #7024's 247 were. What is not available here is #7024's remedy.

## Correcting the issue's inventory

#7025 listed each file's **full** top-level key set rather than the marker-bearing one. Re-derived over all 714 rule-tree YAML files by counting non-empty `import_markers:` lists at any depth:

| file | marker-bearing key | blocks |
|---|---|---:|
| `clojure/test_patterns.yaml` | `testing_frameworks` | 4 |
| `elixir/extras.yaml` | `orms_db` | 5 |
| `elixir/test_patterns.yaml` | `testing` | 6 |
| `go/test_patterns.yaml` | `testing_frameworks` | 7 |
| `lua/test_patterns.yaml` | `testing` | 3 |
| `protobuf/extras.yaml` | `toolchains` | 3 |
| `python/complexity.yaml` | `complexity_signals` | 6 |
| `python/test_patterns.yaml` | `testing_frameworks` | 6 |
| `rust/test_patterns.yaml` | `testing` | 8 |
| `zig/test_patterns.yaml` | `testing` | 1 |

**49 blocks, not 50.** `rust`'s 8 are all under `testing` (`test_conventions` has none); `protobuf`'s 3 are all under `toolchains` (`detection`, `domain_roots`, `extraction_targets`, `package_patterns` have none); `elixir/extras.yaml`'s 5 are all under `orms_db` (`domain_roots` has none); `python/test_patterns.yaml` has 6 non-empty blocks — a 7th `import_markers:` key there is an empty list, and an empty list carries nothing to lose (same convention as #7024's `declaresImportMarkers`). The ten-file set itself reproduces exactly.

## What landed: one test file, no rule file touched

`internal/engine/hidden_import_markers_7025_test.go` (471 lines). **Zero non-test lines changed; zero YAML changed.** The diff is one new file.

- **`TestRuleTree_HiddenImportMarkersAreOutsideLoaderScope`** — the whole-tree assertion #7024's guard is *structurally unable* to make. `importMarkerVisibility` walks only `<lang>/<subdir>/<file>.yaml`, so all ten of these files are invisible to it; this one walks **every** YAML in the tree and reports each top-level container other than `frameworks:` carrying markers at any depth. It asserts the closed inventory above **per file, per key, per block count** (not a total — a total is blind to substitution), plus the two load-bearing properties: every entry is **not loader-scoped**, and every container is **sequence-shaped**. Either being false would mean the #7024 remedy applies and the file should have been rescued, not inventoried. Non-vacuity: `examined >= 700` (the whole tree, not the loader-scoped fifth) and `totalBlocks == 49`.
  - `LoaderScoped` is taken from **`LoadAllRulesFromFSReport`'s own `Candidates` list**, not re-derived from the path here. First draft re-derived it, and mutant M3 (below) showed why that is worthless: a local copy of the path rule agrees with a loader that has changed. Two "independent" checks sharing a premise are one check.
- **`TestLoader_IgnoresRuleFilesOutsideARuleSubdir`** — reason 1, on a synthetic FS, **exercised alone**. A perfectly-spelled `frameworks:` mapping with markers at `rules/clojure/test_patterns.yaml` loads nothing; the byte-identical file at `rules/clojure/frameworks/clojure_test.yaml` loads and its markers are asserted by content, so "no markers" is about the path and not about a loader that reads nothing.
- **`TestLoader_SequenceShapedFrameworksContainerFailsToParse`** — reason 2, on a synthetic FS, exercised alone: a sequence-shaped `frameworks:` inside loader scope produces exactly one `parse` failure and takes the file's `source_patterns` with it.
- **`TestHiddenImportMarkerInventory_Controls`** — grades the checker: five synthetic files covering the #7025 shape (unscoped + sequence), the #7024 shape (scoped + mapping, which must be reported *as scoped* — that is what tells the two defects apart), a correctly-keyed file, an empty marker list, and a file with no markers under a suspicious key.
- **`TestCountImportMarkerBlocks_Depth`** — the one axis the fixtures above hold constant: markers nested at two different depths, an empty list, and a key that merely *contains* `import_markers`.

## The mutual-masking trap, explicitly

#7024's author hit it: with every file renamed, nothing exercised the strict-container guard. Here the pairing would have been worse, because **no data changed at all** — a tree-only guard would be graded solely by a tree that already satisfies it. That is why both mechanism claims (path, shape) are pinned on **constructed** inputs rather than on the real tree, and why the checker has its own five-file control. M3 and M6 below are the mutants those synthetic tests exist to kill, and M6 is killed by *nothing else in the repository*.

## Mutants, permissive direction first

Scored in the branch worktree with a byte-for-byte `cp` backup of every mutated file (never `git checkout --`), each edit proved landed by an exact occurrence count *before* the run, `git status --porcelain` confirmed clean after each restore, `-count=1` throughout, and every verdict re-scored against the final code (M3 was re-scored after the `LoaderScoped` coupling fix).

| # | mutant | verdict |
|---|---|---|
| M1 | apply the fix #7025 asked for — rename `go/test_patterns.yaml`'s `testing_frameworks:` → `frameworks:` | **DEAD** — inventory entry named, total 42 ≠ 49. The failure message says what a *real* rescue would have to look like. |
| M2 | an 11th hidden-marker file (`dart/lint_patterns.yaml`, `linters:` + 1 block) | **DEAD** — new entry named by path and key, total 50 ≠ 49 |
| M3 | **permissive loader**: accept `<lang>/<file>.yaml` by synthesising a `frameworks` segment (~3 lines in `loader.go`) | **DEAD twice** — by `IgnoresRuleFilesOutsideARuleSubdir`, and by the tree test's `LoaderScoped` assertion on 9 of 10 files. **Against the first draft it was DEAD only once**, because `LoaderScoped` was re-derived from the path instead of read from the loader's report. That is the fix described above; the mutant is what found it. |
| M4 | **permissive checker**: count empty `import_markers:` lists as blocks | **DEAD** — 3 tests (tree total, controls, depth) |
| M5 | **checker no-op**: `hiddenImportMarkerInventory` returns `nil, 714, nil` (a plausible-looking `examined` so the non-vacuity floor is satisfied) | **DEAD** — tree test + controls |
| M6 | **make the rename work**: `FrameworkMeta.UnmarshalYAML` adopts a sequence's first entry (~15 lines in `schema.go`) | **DEAD** — by `SequenceShapedFrameworksContainerFailsToParse` **only**. The rest of `./internal/engine` (`-count=1`, 47.0s) is **green** under it. This is the one mutant worth naming: nothing in the repository before this PR observed that the schema refuses a sequence, so "the shape blocks the rename" was an unpinned claim. |
| M7 | **re-scope the walk** to loader-scoped files — i.e. reintroduce exactly #7024's blind spot | **DEAD** — all ten entries vanish; tree test + controls |

## Varied vs. held constant

The repeated defect here is enumerating one set thoroughly and leaving its sibling ungraded. Stated explicitly:

**Varied**
- *Path scope*: unscoped `<lang>/<file>.yaml` **and** scoped `<lang>/<subdir>/<file>.yaml` — both appear as fixtures, and the scoped one must be reported *as scoped* (controls), not merely reported.
- *Container shape*: sequence **and** mapping, both in fixtures and asserted per real-tree entry.
- *Container key*: all 16 real keys probed, not only the 5 marker-bearing ones — including the 5 that carry no markers and would look "covered" by the marker-bearing enumeration.
- *Marker list*: non-empty, empty, and absent; nested at one level and at three; plus a decoy key `import_markers_legacy`.
- *Correctness direction*: a correctly-keyed `frameworks:` file must **not** be reported (controls), and the rescue mutant M1 must fail — so the guard is graded in both directions, not only "hidden must be flagged".
- *Consumers*: four independent packages, each with its own artefact compared, plus a positive control that one of them actually reads the files.

**Held constant, and therefore ungraded here — say so rather than let the enumeration imply coverage**
- **The other 112 ordinary rule files carrying unmodelled top-level keys with no markers.** #7025 measured 193 files with at least one unmodelled top-level key (122 ordinary + 31 `_manifest.yaml` + 40 `_engine/`). This guard sees a file only once it carries markers, exactly as #7024's did. A file at `<lang>/<file>.yaml` acquiring markers under a new key **is** caught (M2); a file acquiring an unmodelled *key* with no markers is **not**. That is #7014's strict top-level guard, still unbuilt, and its scope decision (which file classes it governs) is untouched here.
- **`_manifest.yaml`, `_engine/` and `database_index/`** are walked by the inventory (they are YAML in the tree) but none carries markers, so no assertion here is about them and none of their schemas is judged.
- **The 465 unmodelled metadata key paths** (#7014) and every `requires_framework` setting: untouched.
- **Multi-document YAML**: `yaml.Unmarshal` reads the first document only, as does `loader.go:141` and `entkinds`. A container after a `---` is invisible to this guard, as it is to the engine. No file in the tree currently uses one.
- **Per file, the keys I did not rename**: nothing was renamed at all in the committed diff — the probe renames were reverted and `git status` verified clean — so this axis is vacuous by construction rather than by argument.

## What this exposes, and does not fix

`docs/coverage/registry.json` cites `internal/engine/rules/<lang>/test_patterns.yaml` as evidence in 8 records, with notes like *"PHPUnit test detection via test_patterns.yaml"* and *"embedded and loaded by engine/loader.go (`go:embed all:rules`)"*. The **embed** is real — the bytes are in the binary. The **load** is not: for the flat `<lang>/test_patterns.yaml` files the loader skips the path, so nothing in those files reaches `FrameworkRule`. `tools/coverage`'s own needle scan is a `strings.Contains` over the raw bytes, which is what actually backs the citation. Not touched here; worth a line on #7014.

## The follow-up the rescue actually needs

Rescuing these 49 blocks means **relocating** each catalogue entry into a loader-scoped `<lang>/frameworks/<framework>.yaml` in the mapping shape the schema models — roughly 49 new rule files, each of which the engine then compiles into a rule set. That is a schema-and-behaviour decision (what a `test_patterns.yaml` entry *means* as a `FrameworkRule`, and whether `python/complexity.yaml`'s `complexity_signals` is a framework at all), not a key rename, and it is the wrong thing to smuggle into a guard PR. The alternative — teaching `FrameworkRule` a sequence-shaped container and widening `ruleSubdirs`/the path rule — is M3 and M6 combined, i.e. two of the mutants this PR exists to kill, and needs to be argued on #7014's schema thread rather than applied silently.

Either way the guard above is what makes the change visible when it happens: the inventory fails the moment a file leaves it, and the message says what a real rescue has to look like.

## Verification

- `gofmt -l internal/engine/` clean; `go vet ./internal/engine/` clean.
- `go test ./internal/engine/ ./internal/entkinds/ ./internal/relkinds/ ./tools/coverage/ ./internal/types/ -count=1` → **all ok**, exit 0 (48.1s / 16.0s / 11.5s / 5.6s / 7.8s). The last four are the consumers named in the brief; `internal/types` is included because #7024's review flagged its omission.
- `go test ./cmd/grafel/ -count=1` → **ok, 156.7s**, exit 0, in a fresh three-variable sandbox (`HOME`/`GRAFEL_HOME`/`GRAFEL_DAEMON_ROOT` all under scratch); the real `~/.grafel` was never touched. **Reported honestly: the first run FAILED** — `TestRebuildFiveSequentialAlwaysComplete`, *"Rebuild RPC 5 hung after 5s"*. It passed 3/3 in isolation and the full package passed on re-run in a second fresh sandbox. This PR changes no non-test code and no rule YAML and touches nothing in `cmd/grafel`, so it cannot be the cause; recorded as a pre-existing timing flake rather than dropped. `cmd/grafel` was run despite being a test-only change because it digest-pins the whole graph.
- `git status --porcelain` clean in the worktree; the commit is one file.

Refs #7025, #7013, #7014, #7024.

### DIFF
--- a/internal/engine/hidden_import_markers_7025_test.go
+++ b/internal/engine/hidden_import_markers_7025_test.go
@@ -1,10 +1,14 @@
 package engine
 
 import (
+	"bytes"
+	"errors"
 	"fmt"
+	"io"
 	"io/fs"
 	"path/filepath"
 	"sort"
+	"strings"
 	"testing"
 	"testing/fstest"
 
@@ -116,13 +120,15 @@ type hiddenMarkerInventoryEntry struct {
 // sibling look covered; this function exists to look at the sibling.
 //
 // examined counts files parsed, so "nothing hidden" is distinguishable from
-// "nothing scanned".
-func hiddenImportMarkerInventory(fsys fs.FS, rootDir string) (entries []hiddenMarkerInventoryEntry, examined int, err error) {
+// "nothing scanned". skipped lists every file the walk reached but could not
+// read as a mapping, so "nothing hidden" is also distinguishable from "the
+// file was never decoded" — see the block comment at the skip sites.
+func hiddenImportMarkerInventory(fsys fs.FS, rootDir string) (entries []hiddenMarkerInventoryEntry, examined int, skipped []string, err error) {
 	// Ask the loader which files it opens, rather than re-deriving its path
 	// rule. See hiddenMarkerInventoryEntry.LoaderScoped.
 	_, report, loadErr := LoadAllRulesFromFSReport(fsys, rootDir)
 	if loadErr != nil {
-		return nil, 0, loadErr
+		return nil, 0, nil, loadErr
 	}
 	candidates := make(map[string]bool, len(report.Candidates))
 	for _, c := range report.Candidates {
@@ -139,17 +145,41 @@ func hiddenImportMarkerInventory(fsys fs.FS, rootDir string) (entries []hiddenMa
 		if ext := filepath.Ext(path); ext != ".yaml" && ext != ".yml" {
 			return nil
 		}
+		// Every early return from here on is recorded. A scan-and-assert-absence
+		// guard has more ways to be a no-op than to work, and "the decoder could
+		// not read this file" is the one that looks healthiest: examined stays
+		// plausible, the entry set stays exactly right, and a file carrying
+		// hidden markers is simply never looked at. Two real shapes do it — a
+		// top-level SEQUENCE and a MULTI-DOCUMENT file — and neither is
+		// hypothetical: the loader and internal/entkinds have the same
+		// first-document-only limit. So a skip must be loud, not impossible.
 		data, readErr := fs.ReadFile(fsys, path)
 		if readErr != nil {
+			skipped = append(skipped, filepath.ToSlash(path)+": read: "+readErr.Error())
 			return nil
 		}
+		//
+		// The two shapes fail differently and both have to be caught. A
+		// top-level sequence fails the decode outright. A multi-document file
+		// does NOT: yaml.Unmarshal reads document 1 and returns nil, so a second
+		// document carrying markers is dropped with no error at all — the
+		// quieter of the two. A decoder loop is what tells them apart.
+		dec := yaml.NewDecoder(bytes.NewReader(data))
 		var raw map[string]any
-		if yaml.Unmarshal(data, &raw) != nil {
+		if uerr := dec.Decode(&raw); uerr != nil && !errors.Is(uerr, io.EOF) {
+			skipped = append(skipped, filepath.ToSlash(path)+": decode into map[string]any: "+uerr.Error())
+			return nil
+		}
+		var extra any
+		if derr := dec.Decode(&extra); derr == nil {
+			skipped = append(skipped, filepath.ToSlash(path)+
+				": multi-document: only document 1 is decoded, later documents are not scanned")
 			return nil
 		}
 		examined++
 		rel, relErr := filepath.Rel(rootDir, path)
 		if relErr != nil {
+			skipped = append(skipped, filepath.ToSlash(path)+": relative path: "+relErr.Error())
 			return nil
 		}
 		relSlash := filepath.ToSlash(rel)
@@ -180,7 +210,8 @@ func hiddenImportMarkerInventory(fsys fs.FS, rootDir string) (entries []hiddenMa
 		}
 		return entries[i].Key < entries[j].Key
 	})
-	return entries, examined, walkErr
+	sort.Strings(skipped)
+	return entries, examined, skipped, walkErr
 }
 
 // countImportMarkerBlocks counts non-empty `import_markers:` lists anywhere
@@ -243,9 +274,18 @@ var knownHiddenMarkerInventory = []hiddenMarkerInventoryEntry{
 // assertion #7024's guard cannot make, because its walk is loader-scoped and
 // these ten files are not.
 func TestRuleTree_HiddenImportMarkersAreOutsideLoaderScope(t *testing.T) {
-	got, examined, err := hiddenImportMarkerInventory(rulesFS, "rules")
+	got, examined, skipped, err := hiddenImportMarkerInventory(rulesFS, "rules")
 	if err != nil {
 		t.Fatalf("walking the embedded rules tree: %v", err)
+	}
+	// A file the walk could not decode is a file this guard did not judge. All
+	// 714 rule files parse as mappings today, so this costs nothing and closes
+	// the hole permanently: a rule file added as a top-level sequence, or as a
+	// multi-document file, must fail here rather than pass silently.
+	for _, sk := range skipped {
+		t.Errorf("the hidden-marker walk skipped %s. This guard asserts an ABSENCE, so a "+
+			"file it cannot decode is a file it did not check — markers hidden there would "+
+			"read as a clean tree (#7025). Fix the file, or teach the walk its shape.", sk)
 	}
 	// Non-vacuity: the whole tree, not the loader-scoped fifth of it.
 	if examined < 700 {
@@ -421,9 +461,12 @@ func TestHiddenImportMarkerInventory_Controls(t *testing.T) {
 			"toolchains:\n- name: cargo\n  detection:\n    files:\n    - Cargo.toml\n")},
 	}
 
-	got, examined, err := hiddenImportMarkerInventory(fsys, "rules")
+	got, examined, skipped, err := hiddenImportMarkerInventory(fsys, "rules")
 	if err != nil {
 		t.Fatalf("hiddenImportMarkerInventory: %v", err)
+	}
+	if len(skipped) != 0 {
+		t.Errorf("skipped = %v, want none: every fixture here is a mapping", skipped)
 	}
 	if examined != 5 {
 		t.Errorf("examined = %d, want 5 (every yaml file, not only loader-scoped ones)", examined)
@@ -440,6 +483,66 @@ func TestHiddenImportMarkerInventory_Controls(t *testing.T) {
 		if got[i] != want[i] {
 			t.Errorf("inventory[%d] = %+v, want %+v", i, got[i], want[i])
 		}
+	}
+}
+
+// TestHiddenImportMarkerInventory_UndecodableFilesAreReported is the positive
+// control on the skip counter. Without it, `skipped == 0` on the real tree is
+// satisfied just as well by a walk that never appends to skipped at all — the
+// same vacuity the counter exists to remove, moved one level up.
+//
+// Both fixtures are real shapes, not invented ones, and both carry hidden
+// markers so that a silent skip would be a MISSED FINDING rather than a
+// harmless omission:
+//
+//   - a top-level SEQUENCE, which does not decode into map[string]any; and
+//   - a MULTI-DOCUMENT file, whose second document is where the markers are.
+//
+// The third fixture is an ordinary mapping, so "everything was skipped" cannot
+// pass either.
+func TestHiddenImportMarkerInventory_UndecodableFilesAreReported(t *testing.T) {
+	fsys := fstest.MapFS{
+		// Top-level sequence: yaml.Unmarshal cannot fit it into a map.
+		"rules/go/seq_patterns.yaml": {Data: []byte(
+			"- name: testing\n  detection:\n    import_markers:\n    - '\"testing\"'\n")},
+		// Multi-document: the markers live after the separator.
+		"rules/go/multidoc_patterns.yaml": {Data: []byte(
+			"testing_frameworks: []\n---\ntesting:\n- name: testify\n  detection:\n" +
+				"    import_markers:\n    - '\"github.com/stretchr/testify\"'\n")},
+		// Decodable, and genuinely hidden: the walk must still do its job.
+		"rules/lua/test_patterns.yaml": {Data: []byte(
+			"testing:\n- name: busted\n  detection:\n    import_markers:\n    - \"require('busted')\"\n")},
+	}
+
+	got, examined, skipped, err := hiddenImportMarkerInventory(fsys, "rules")
+	if err != nil {
+		t.Fatalf("hiddenImportMarkerInventory: %v", err)
+	}
+
+	if len(skipped) != 2 {
+		t.Fatalf("skipped = %v, want 2 entries (the sequence file and the multi-document "+
+			"file): if either now decodes, the walk grew a shape it did not have and the "+
+			"tree assertion's skipped==0 is checking something different", skipped)
+	}
+	for _, want := range []string{"rules/go/multidoc_patterns.yaml", "rules/go/seq_patterns.yaml"} {
+		found := false
+		for _, sk := range skipped {
+			if strings.HasPrefix(sk, want+":") {
+				found = true
+			}
+		}
+		if !found {
+			t.Errorf("skipped = %v, want an entry naming %s", skipped, want)
+		}
+	}
+	// The skip is loud, not fatal: the rest of the walk still runs and still
+	// reports. A counter that fired by aborting the walk would make the tree
+	// assertion pass for the wrong reason.
+	if examined != 1 {
+		t.Errorf("examined = %d, want 1 (only the decodable fixture)", examined)
+	}
+	if len(got) != 1 || got[0].Path != "lua/test_patterns.yaml" || got[0].Key != "testing" || got[0].Blocks != 1 {
+		t.Errorf("inventory = %+v, want the one decodable hidden entry", got)
 	}
 }
 
--- a/internal/engine/hidden_import_markers_7025_test.go
+++ b/internal/engine/hidden_import_markers_7025_test.go
@@ -1,0 +1,471 @@
+package engine
+
+import (
+	"fmt"
+	"io/fs"
+	"path/filepath"
+	"sort"
+	"testing"
+	"testing/fstest"
+
+	"gopkg.in/yaml.v3"
+)
+
+// #7025: the ten rule files that still declare import_markers the engine cannot
+// see — and the measurement that says renaming their container is NOT the fix.
+//
+// # What #7024 fixed, and why the same move does not apply here
+//
+// #7024 found 247 loader-scoped rule files spelling the `frameworks:` container
+// 14 other ways. Those files WERE opened by the loader, WERE unmarshalled into
+// FrameworkRule, and their content was dropped only because yaml.Unmarshal
+// discards unknown top-level keys. Renaming the key to `frameworks:` was
+// therefore sufficient: 52 files' markers became readable.
+//
+// The remaining ten files fail for two different reasons, and neither is the
+// container spelling:
+//
+//  1. THE PATH. LoadAllRulesFromFS accepts exactly <lang>/<subdir>/<file>.yaml
+//     for subdir in ruleSubdirs (frameworks, orms, queues). All ten sit at
+//     <lang>/<file>.yaml — two path segments — so the loader never opens them,
+//     never adds them to RuleLoadReport.Candidates, and never decodes them at
+//     all. Spelling the container `frameworks:` changes nothing, because
+//     nothing reads the file.
+//
+//  2. THE SHAPE. FrameworkRule.Frameworks is a single FrameworkMeta MAPPING.
+//     Every one of the ten containers is a SEQUENCE of framework entries
+//     (`- name: clojure.test` …). Renaming such a container to `frameworks:` in
+//     a file the loader does open makes the file stop parsing: the decode
+//     errors, the loader records a parse failure, and the file's rules —
+//     including any it currently contributes — are dropped. The rename is not a
+//     no-op there, it is a regression.
+//
+// Both mechanisms are pinned below on a synthetic FS, so each is exercised on
+// its own rather than only through the real tree (a guard that fires only when
+// the tree happens to feed it grades nothing).
+//
+// # Who reads these containers today: nobody (measured, #7025 step 1)
+//
+// Established by rename-to-break, not by grep. All 16 top-level keys on the ten
+// files — testing_frameworks, testing, test_conventions, complexity_signals,
+// domain_roots, orms_db, detection, extraction_targets, package_patterns,
+// toolchains — were renamed to `zzprobe_<key>:` and every disk-walking consumer
+// of the rule tree was re-run:
+//
+//   - internal/engine (the embedded loader): compiled rule set byte-identical.
+//   - internal/entkinds (walks the YAML at ANY depth): site list identical.
+//   - internal/relkinds (top-level relationship_rules only): identical.
+//   - tools/coverage (Discover over the repo root): identical once its own
+//     map-order nondeterminism is normalised away.
+//
+// Positive control on that measurement: appending `probe_block: {entity_type:
+// ProbeKind7025}` to go/test_patterns.yaml DID appear in the entkinds site
+// list, so the scanners do read these files — they simply do not decode the
+// container key. tools/coverage's only use of test_patterns.yaml is a
+// whole-file `strings.Contains` needle for a framework NAME
+// (discover.go testFrameworkWalker), which is blind to key names by
+// construction.
+//
+// So the containers are inert, exactly as #7024's 247 were. What is NOT
+// available here is #7024's remedy.
+//
+// # What this guard therefore asserts
+//
+// The inventory is a closed set: these ten files, these container keys, these
+// block counts, and every one of them outside loader scope. An eleventh such
+// file fails here. One of the ten moving INTO loader scope while still keyed on
+// an unmodelled container fails here (and in #7024's guard). Rescuing one — by
+// relocating its entries under <lang>/frameworks/ in the mapping shape the
+// schema models — fails here too, which is the point: the count moves only when
+// something real happens, and the test names what changed.
+//
+// Deliberately NOT done: the rescue. Relocating these catalogues into
+// loader-scoped rule files turns each entry into a rule set the engine
+// compiles, one per framework, which is a schema and behaviour decision, not a
+// key rename. Left to the #7025 follow-up.
+
+// hiddenMarkerInventoryEntry is one (file, top-level container key) pair that
+// declares import_markers the engine cannot reach.
+type hiddenMarkerInventoryEntry struct {
+	// Path is relative to the rules root, slash-separated.
+	Path string
+	// Key is the top-level container key the markers sit under.
+	Key string
+	// Blocks is the number of non-empty import_markers lists under Key, at any
+	// depth.
+	Blocks int
+	// LoaderScoped reports whether the loader opens this file at all. It is
+	// taken from LoadAllRulesFromFSReport's own Candidates list rather than
+	// re-derived from the path here: a local copy of the path rule would agree
+	// with a loader that had changed, and the whole point of this field is to
+	// track what the loader actually does.
+	LoaderScoped bool
+	// SequenceShaped reports whether the container is a YAML sequence.
+	// FrameworkRule.Frameworks is a mapping, so a sequence cannot be rescued by
+	// renaming the key — the decode would fail.
+	SequenceShaped bool
+}
+
+// hiddenImportMarkerInventory walks EVERY yaml file under rootDir — not only
+// the loader-scoped ones — and reports each top-level container other than
+// `frameworks:` that carries non-empty import_markers anywhere beneath it.
+//
+// The whole-tree walk is the difference from #7024's importMarkerVisibility,
+// which is scoped to <lang>/<subdir>/<file>.yaml and is structurally unable to
+// see any of these ten files. Enumerating one set thoroughly is what made its
+// sibling look covered; this function exists to look at the sibling.
+//
+// examined counts files parsed, so "nothing hidden" is distinguishable from
+// "nothing scanned".
+func hiddenImportMarkerInventory(fsys fs.FS, rootDir string) (entries []hiddenMarkerInventoryEntry, examined int, err error) {
+	// Ask the loader which files it opens, rather than re-deriving its path
+	// rule. See hiddenMarkerInventoryEntry.LoaderScoped.
+	_, report, loadErr := LoadAllRulesFromFSReport(fsys, rootDir)
+	if loadErr != nil {
+		return nil, 0, loadErr
+	}
+	candidates := make(map[string]bool, len(report.Candidates))
+	for _, c := range report.Candidates {
+		candidates[filepath.ToSlash(c)] = true
+	}
+
+	walkErr := fs.WalkDir(fsys, rootDir, func(path string, d fs.DirEntry, err error) error {
+		if err != nil {
+			return err
+		}
+		if d.IsDir() {
+			return nil
+		}
+		if ext := filepath.Ext(path); ext != ".yaml" && ext != ".yml" {
+			return nil
+		}
+		data, readErr := fs.ReadFile(fsys, path)
+		if readErr != nil {
+			return nil
+		}
+		var raw map[string]any
+		if yaml.Unmarshal(data, &raw) != nil {
+			return nil
+		}
+		examined++
+		rel, relErr := filepath.Rel(rootDir, path)
+		if relErr != nil {
+			return nil
+		}
+		relSlash := filepath.ToSlash(rel)
+		scoped := candidates[filepath.ToSlash(path)]
+
+		for key, val := range raw {
+			if key == "frameworks" {
+				continue
+			}
+			n := countImportMarkerBlocks(val)
+			if n == 0 {
+				continue
+			}
+			_, isSeq := val.([]any)
+			entries = append(entries, hiddenMarkerInventoryEntry{
+				Path:           relSlash,
+				Key:            key,
+				Blocks:         n,
+				LoaderScoped:   scoped,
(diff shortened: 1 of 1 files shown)


cajasmota:
# APPROVE

Independent adversarial review. Own worktree at `/private/tmp/claude-501/-Users-jorgecajas-Projects-archigraph/c0fefed4-0195-4c81-9a94-07a7145710a4/scratchpad/rev-7025-m2q/wt`, head `5129aa5ba`, macOS arm64. Every claim below was re-derived here; nothing was taken from the PR body. Nothing was pushed; the branch is untouched.

**The author's replacement account is correct on every load-bearing point, and the two synthetic controls each fail alone.** One latent gap in the guard's walk is worth a follow-up line but does not block.

## 1. The PATH rule — as described, exactly

`loader.go:119-127`: `parts := strings.Split(...)`, `if len(parts) != 3 { return nil }`, then `if !ruleSubdirs[subdir] { return nil }`. `ruleSubdirs` (`loader.go:68-72`) is exactly `{frameworks, orms, queues}` — three entries, no more. `report.Candidates` is appended *after* both rejections, so a non-conforming file is never a candidate, never read, never decoded. Confirmed hard 3-segment.

Other entry points into the tree: `internal/entkinds`, `internal/relkinds` and `tools/coverage/discover.go` do walk it at arbitrary depth, and `internal/engine/loader.go` is the only thing that decodes into `FrameworkRule`. I did not re-run the author's four-consumer rename-to-break (they ran it with a positive control that entkinds *does* read these files); the structural point that matters for this PR — nothing but the loader decodes container keys — holds by inspection of the decode sites.

**Measured, not asserted:** loader-scoped YAML files in the tree = **520**; total YAML = **714**. The 10 files sit in the 194-file remainder.

## 2. The SHAPE rule — as described, exactly

`schema.go:19`: `Frameworks FrameworkMeta` — a plain struct, i.e. mapping-only. Grep for `UnmarshalYAML` across `internal/` returns methods on `FileConvention`, `SourcePattern` and `RelationshipRule` only (`schema_strict.go:136/151/166`). **There is no custom unmarshaller on `FrameworkMeta` or `FrameworkRule`**, so a sequence hits yaml.v3's default struct decode and errors. Confirmed.

## 3. The inventory — re-derived independently, all 10, and it reproduces

Not by reading the Go checker: with my own PyYAML walker over `internal/engine/rules` (independent parser, independent recursion, independent path rule), counting non-empty `import_markers` lists at any depth under every non-`frameworks` top-level key:

```
clojure/test_patterns.yaml  testing_frameworks  4  scoped=False seq=True
elixir/extras.yaml          orms_db             5  scoped=False seq=True
elixir/test_patterns.yaml   testing             6  scoped=False seq=True
go/test_patterns.yaml       testing_frameworks  7  scoped=False seq=True
lua/test_patterns.yaml      testing             3  scoped=False seq=True
protobuf/extras.yaml        toolchains          3  scoped=False seq=True
python/complexity.yaml      complexity_signals  6  scoped=False seq=True
python/test_patterns.yaml   testing_frameworks  6  scoped=False seq=True
rust/test_patterns.yaml     testing             8  scoped=False seq=True
zig/test_patterns.yaml      testing             1  scoped=False seq=True
ENTRIES 10  TOTALBLOCKS 49
```

**49, not 50** — the author's correction to #7025 stands. All 10 unscoped, all 10 sequence-shaped, **checked across 10 not a sample**. `test_conventions`, `domain_roots`, `detection`, `extraction_targets`, `package_patterns` carry zero markers, so #7025 attributing markers to them was wrong. `python/test_patterns.yaml` = 6 non-empty (one further `import_markers:` is `[]`).

## 4. The zero-byte-diff proof — reproduced, and NOT vacuous

Renamed `go/test_patterns.yaml`'s `testing_frameworks:` → `frameworks:` (edit proved landed: `assert count == 1` before substitution) and compared the compiled rule set. Identical: same JSON byte length, same SHA-256 prefix.

Non-vacuity is the part worth stating, because two empty things also compare equal. The compared object is:

```
candidates=520 loaded=520 failures=0 langs=34 rulesets=520 markers=772 jsonbytes=174942
```

175 KB, 520 rule sets, 772 import markers — a large, populated object that did not move by one byte. The rename is inert on these files because the file is never opened, and the comparison is capable of seeing a change.

## 5. The mutual-masking trap — THE ONE THAT MATTERED — clears

Both mutants applied to my own worktree, each edit proved landed by an exact occurrence count, restored by `cp` from a shasum-verified backup, `git status --porcelain` empty after each, `-count=1` throughout. **Every test run individually, so nothing rides on a sibling.**

**M3 — permissive loader** (accept `<lang>/<file>.yaml` by synthesising a `frameworks` segment, 3 lines in `loader.go`):

| test, run alone | result |
|---|---|
| `TestLoader_IgnoresRuleFilesOutsideARuleSubdir` | **FAIL** |
| `TestRuleTree_HiddenImportMarkersAreOutsideLoaderScope` | **FAIL** — 10 of 10 `AND is loader-scoped` errors |
| `TestHiddenImportMarkerInventory_Controls` | **FAIL** |
| `TestLoader_SequenceShapedFrameworksContainerFailsToParse` | **ok** |

**M6 — sequence-tolerant `FrameworkMeta.UnmarshalYAML`** (adopts the first entry, ~20 lines in `schema.go`):

| test, run alone | result |
|---|---|
| `TestLoader_SequenceShapedFrameworksContainerFailsToParse` | **FAIL** |
| `TestLoader_IgnoresRuleFilesOutsideARuleSubdir` | **ok** |
| `TestRuleTree_HiddenImportMarkersAreOutsideLoaderScope` | **ok** |
| `TestHiddenImportMarkerInventory_Controls` | **ok** |
| whole `./internal/engine` package | **FAIL, and only on that one test** (45.9s) |

Both **DEAD**, and the kill sets are disjoint on the two synthetic controls. This is not a pair that only fires together: the path control fails alone under a path-only mutation with the shape control green, and the shape control fails alone under a shape-only mutation with the path control green. Given that the PR changes zero data, this was the whole review, and it holds.

M6's second half is worth repeating: with the sequence-tolerant unmarshaller applied, **the entire rest of the package is green**. Before this PR nothing in the repository observed that the schema refuses a sequence. That claim is now pinned by exactly one test, which is one more than it had.

**M3's coupling fix is real and verifiable from here.** `hiddenMarkerInventoryEntry.LoaderScoped` is populated from `LoadAllRulesFromFSReport`'s own `Candidates` (`hidden_import_markers_7025_test.go:124-131`). Proof it cannot drift: I mutated *only* `loader.go` and all ten entries flipped to `LoaderScoped=true`. A locally re-derived path rule would have agreed with itself and stayed silent. (Minor: the PR body says "9 of 10"; I measure 10 of 10 — stronger than claimed.)

## 6. The walk really is whole-tree

Not a floor argument — a count. The inventory examines **714** YAML files. #7024's `importMarkerVisibility` (and `rule_container_key_7013_test.go:86`, `graphql_inert_rules_guard_6393_test.go:91`) filter on `len(parts) != 3 || !ruleSubdirs[parts[1]]` and therefore see **520**. The 10 files are in the 194 a loader-scoped walk is structurally unable to reach. The `examined >= 700` floor is not decorative: re-scoping the walk to loader scope yields 520, which trips it.

## 7. The follow-up claim — tested, and the guard does go red

The claim "the guard fails the moment anyone does the real fix" is the one that decides whether the next person deletes it. I did the real fix: lifted `clojure.test` out of `clojure/test_patterns.yaml`'s `testing_frameworks:` sequence and wrote it to `clojure/frameworks/clojure_test.yaml` in the mapping shape the schema models. Result:

```
--- FAIL: TestRuleTree_HiddenImportMarkersAreOutsideLoaderScope
    expected hidden-marker entry "clojure/test_patterns.yaml:testing_frameworks=4" is gone.
    If it was rescued — the entries relocated under <lang>/frameworks/ in the mapping shape
    the schema models — update knownHiddenMarkerInventory and say so. ...
    total unreachable import_markers blocks = 48, want 49
```

It fails, it names the file, and the message tells the person doing the work what to do. It does not silently tolerate the good state. Restored; `git status` clean.

## 8. One mutant the author did not run — ALIVE, disclosed only in half

`hiddenImportMarkerInventory` decodes into `map[string]any` and, on error, **returns silently without incrementing `examined`**. Two shapes therefore fall out of the guard entirely. I added both as real rule files:

- `dart/topseq_7025rev.yaml` — top-level **sequence** carrying `import_markers`.
- `dart/multidoc_7025rev.yaml` — a second YAML document after `---` carrying `testing:` + markers.

`TestRuleTree_HiddenImportMarkersAreOutsideLoaderScope` → **ok**. Two new rule files with unreachable `import_markers` and the guard says nothing, while its own doc comment says "an eleventh such file fails here". Edits proved landed (`ls | grep 7025rev` showed both), removed after.

**Verdict: distinguishable and ungraded.** Not equivalent (the guard's stated contract is exactly this), not unreachable (a catalogue file that drops its wrapping key *is* a top-level sequence; these ten files are catalogues). The multi-document half is disclosed in the PR's held-constant list; **the top-level-sequence half is not** — the held-constant section names multi-doc but not "top level is not a mapping", so the enumeration reads more complete than it is.

**Price: ~4 lines, and it is currently free.** I measured the tree: 714 YAML files, **714 unmarshal into `map[string]any`, 0 skipped**. So `if skipped != 0 { t.Errorf(...) }` — or asserting `examined == totalYAMLSeen` — costs zero red files today and closes the whole hole including the malformed-file case. Worth a follow-up commit or a line on #7014; not a merge blocker, since it is latent and the guard is strictly better than what existed.

Two loader-scoped files, `ruby/frameworks/sidekiq.yaml` and `ruby/frameworks/rspec.yaml`, are comment-only (882 and 817 bytes, zero keys) and load as empty rule sets. Unrelated to this PR; noting it since it surfaced from the same walk.

## 9. The `cmd/grafel` flake — pre-existing timing, from evidence

`TestRebuildFiveSequentialAlwaysComplete` (`cmd/grafel/daemon_rebuild_test.go:134-160`) fires five sequential `daemonRebuildFuncCore` calls against a **mock** `mockIndexFn` and a temp group, each with a hard `time.After(5 * time.Second)` wall-clock deadline. It never calls `engine.LoadAllRules`, never touches the rules tree, and reaches no code this PR modifies — the PR is one new `_test.go` file in `internal/engine`, zero non-test lines, zero YAML. A wall-clock deadline plus a machine running several agent worktrees concurrently is a sufficient explanation, and it passed 3/3 in isolation and on a full re-run.

**Read: pre-existing, environmental, not caused by this PR.** The right follow-up is not this PR — it is that a 5s hard deadline on a mocked path is a flake generator on a loaded runner. Worth its own issue rather than a rerun next time it bites.

## 10. Merge readiness

The branch's parent is `e2b54fef0`; `main` has since advanced to `8bf2ad2c8` (`go/frameworks/{chi,echo,fiber,gin}.yaml` + `tier_a_config_variable_6916_test.go`). I merged `main` locally in my own worktree to check the guard survives it:

- merge is clean, no conflicts;
- inventory on the merged tree is **unchanged: 10 files, 49 blocks**;
- `go test ./internal/engine/ -count=1` on the merged tree → **ok, 46.8s**.

So the 49 is not stale against current `main`.

Also verified on the branch as shipped: `gofmt -l internal/engine/` clean, `go vet ./internal/engine/` clean, `go test ./internal/engine/ -count=1` → **ok, 45.9s**. Worktree `git status --porcelain` empty at the end; every mutant restored by `cp` from a shasum-verified backup, never `git checkout --`.

---

**APPROVE.** Safe to merge. The premise of #7025 was wrong and the replacement account is right on all four load-bearing points (path, shape, inventory of 49, inertness), each re-derived here from source rather than from the PR body. The controls that make a zero-data-change guard meaningful each fail alone, which is the specific thing this PR had to get right. The one gap — a top-level-sequence or multi-document rule file slipping past the walk unexamined — is latent, costs ~4 lines to close, and belongs in a follow-up.

cajasmota:
## Follow-up pushed: `de601e2e2`. Merged current `main` (`8bf2ad2c8`) first — clean, inventory still 49.

The vacuity finding is right and it was the one that mattered, because this PR's whole value is the guard. Fixed, plus one correction to the diagnosis.

### The two shapes fail differently, and only one of them was a "skip"

- **Top-level sequence** — `yaml.Unmarshal` into `map[string]any` errors, the walk returned early, nothing recorded. A genuine silent skip. Undisclosed, as you said.
- **Multi-document** — **does not error at all.** `yaml.Unmarshal` decodes document 1 and returns `nil`, so the walk proceeded normally with a perfectly valid mapping and the markers after the separator were simply never in the object it looked at. `skipped == 0` alone would **not** have closed this one; the file is not skipped, it is truncated. It needs a `yaml.NewDecoder` loop and a second `Decode` to notice a document 2 exists.

So the ask as written (~4 lines, count the skips) would have closed the shape you flagged and left the shape I disclosed exactly as open as before — while now *looking* closed, which is worse than the disclosure. Both are recorded now.

### What landed

`skipped []string` on `hiddenImportMarkerInventory`, appended at all four early returns (read failure, decode failure, multi-document, rel-path failure), sorted, and asserted empty by `TestRuleTree_HiddenImportMarkersAreOutsideLoaderScope` with the offending path named. `examined` is unchanged, per the ask. All 714 rule files are single-document mappings today, so the cost is zero and the hole closes permanently.

**`TestHiddenImportMarkerInventory_UndecodableFilesAreReported`** is the positive control, and it is not optional: `skipped == 0` on a tree that has nothing to skip is satisfied just as well by a walk that never appends to `skipped` at all — the same vacuity, moved one level up. It uses your two fixtures, both **carrying hidden markers**, so a silent skip there is a missed finding rather than a harmless omission, plus a third decodable-and-hidden file so "everything was skipped" cannot pass either. It also asserts the walk **continues** after a skip (`examined == 1`, the one real entry still reported) — a counter that fired by aborting would make the tree assertion pass for the wrong reason.

### Mutants: 9/9 DEAD, all re-scored against the edited walk

Nothing was carried over; the walk they were scored on changed, so every verdict was re-derived. Same discipline as before — exact occurrence count asserted before each run, `cp`-restore from shasum-verified backups, `git status` clean after each, `-count=1`.

| # | mutant | verdict (re-scored) |
|---|---|---|
| M1 | the naive #7025 rename (`go/test_patterns.yaml` → `frameworks:`) | **DEAD** — tree test |
| M2 | an 11th unscoped marker-bearing file | **DEAD** — tree test (new-entry set + total) |
| M3 | permissive loader accepting 2-segment paths | **DEAD ×3** — path control, tree test, checker controls |
| M4 | permissive checker: count empty `import_markers` lists | **DEAD ×3** |
| M5 | inventory returns `nil, 714, nil, nil` | **DEAD ×3** — including the new control |
| M6 | `FrameworkMeta.UnmarshalYAML` adopts a sequence's first entry | **DEAD ×1** — shape control **only**; rest of `./internal/engine` green (50.6s) |
| M7 | re-scope the walk to loader-scoped files (#7024's blind spot) | **DEAD ×3** |
| **M8** | **the skip counter never appends** (`_ = append(...)`, ×4, the new assertion's own vacuity) | **DEAD ×1 — by the new control only.** Every other test in the package is green under it. |
| **M9** | **revert the decoder loop to plain `yaml.Unmarshal`** so multi-document goes quiet again | **DEAD ×1 — by the new control only**, and it is what proves the multi-doc half is a real assertion rather than a comment |

M6, M8 and M9 each die by exactly one test and nothing else in the repository, which is the disjointness that keeps the guard non-vacuous.

### Confirmations of your points

- **`LoaderScoped` / 10 vs 9** — you are right and my number was wrong. Re-run of M3 against the current test fails the tree assertion on all ten, plus the path control and the checker controls. My "9" was read off a truncated `head -8` of the failure output.
- **Your widening probe and the reviewer's rescue probe are the two directions**, and both now also run past the skip counter without tripping it — worth noting, since a probe file that failed to decode would have produced a false green in either direction.

### Not done, per your instruction

No rescue, no scope widening; tests only, one file. And noted for #7025: activating the 49 gates is now an open question, not an assumed yes — they have been inert for their whole life, so switching them on newly fires 49 gates that have never run, against #7013's 24.5% over-fire finding. That needs measuring before anyone relocates a file.

### Verification

- `gofmt -l internal/engine/` clean; `go vet ./internal/engine/` clean.
- `go test ./internal/engine/ ./internal/entkinds/ ./internal/relkinds/ ./tools/coverage/ ./internal/types/ -count=1` → all **ok**, exit 0 (47.8 / 15.1 / 7.8 / 4.6 / 5.6s).
- `go test ./cmd/grafel/ -count=1` → **ok, 143.0s**, exit 0, fresh three-variable sandbox; the real `~/.grafel` untouched. Green first time in this round (the earlier `TestRebuildFiveSequentialAlwaysComplete` hang did not recur).
- `git status --porcelain` clean; branch merged onto `8bf2ad2c8`; the PR is still one file.

Issue #7028, opened 2026-09-08
CLASS: 13 identical-pattern groups shadow 1,560 entities under a silently-dead twin — 2 resolved/unblocked (#7029, #7032), 11 remain (~1,338); gating one twin alone removes ZERO


cajasmota:
Measured while pricing #7013's `requires_framework` gates. **Gating one of a pair of byte-identical patterns in two different rule files removes nothing — it silently reassigns the entities to the other file's framework.**

## The mechanism

`seenEntities` (`internal/engine/detector.go:367`) is **per file and shared across rule sets**, and rule sets load in `fs.WalkDir` lexical order (`loader.go:103`). So when two rule files carry the same pattern, whichever sorts first wins every mint, and the second is silently dead — it looks like a second producer but has never produced anything on any file the first also matched.

Gate the first, and the mints do not disappear. They transfer.

## The measured instance

`python/frameworks/flask.yaml` #1 and `python/frameworks/quart.yaml` #1 carry a **byte-identical** pattern — same regex `@(\w+)\.(get|post|put|patch|delete)…`, same `entity_type`, same `name_group`. `flask` sorts first.

Adding `requires_framework: true` to `flask.yaml` #1 alone:

- **before**: `flask#1 {django: 75, flask: 5}`
- **after**: `flask#1 {flask: 1}`, `quart#1 {django: 75, flask: 4}`

**Zero entities removed. And 4 genuine Flask routes are relabelled Quart** — a repo that uses Flask and not Quart now reports Quart routes. That is strictly worse than making no change at all, and every summary count stays flat while it happens.

## The second pair, unfixed

`flask.yaml` #0 and `quart.yaml` #0 are the same identical twin, covering **178 mints**. Not touched by the #7013 work, and the same trap waits for anyone who gates or tightens either one.

## Why this is worth its own issue

Three properties make it dangerous beyond the instance:

1. **It is invisible to a count.** Total mints, per-kind totals and the over-fire rate are all unchanged. Only a per-*site* diff of the mint set shows the transfer, which is why #7013's step-2 pass — which ranked by site — did not surface it and the recall pass did.
2. **It fails toward a wrong label, not a missing one.** A dropped entity is a recall loss someone eventually notices. A *relabelled* entity is a confident wrong answer, and `bug-extractor` will not flag it because the entity binds fine.
3. **The dead twin looks like a working producer.** Any audit that asks "does this pattern have a producer" answers yes for both.

## What is not established

[...]

PR #7029, opened 2026-09-08
fix(#7013): variant D — gate the 5 worst over-firing python patterns, and gate the flask/quart route TWIN together or it removes nothing

cajasmota:
Ships **variant D** as priced by the two measurement passes on #7013: **762 over-fired entities removed, 9 true positives lost, net +753 — 43% of this P1's 1,757.** Six one-line rules changes, no new mechanism and no new metadata.

`Refs #7013, #7028.` This closes 43% of the issue, not the issue.

## Per pattern, before → after

| pattern | kind | minted | bad removed | good removed | net |
|---|---|---:|---:|---:|---:|
| `pytest.yaml` #1 | TestClass | 358 | 322 | 4 (**3 real**, 1 is itself an FP) | +318 |
| `sqlalchemy.yaml` #5 | Constraint | 251 | 246 | 4 | +242 |
| `flask.yaml` #6 | Model | 78 | 76 | 2 → **1** after the widening | +75 |
| `flask.yaml` #1 + `quart.yaml` #1 | Route | 80 | **75** (0 if gated alone) | 4 → **0** after the widening | +75 |
| `quart.yaml` #5 | Middleware | 43 | 43 | 0 *here* | +43 |
| **total** | | 810 | **762** | **9** | **+753** |

Plus: `flask.yaml`'s `import_markers` widened with `from flask import` and `import flask`.

Every index was re-verified against the pattern that actually produces the named `entity_type` rather than trusted as an ordinal, and each assertion in the new tests identifies its pattern by a fragment of the regex, not by position — an ordinal slides silently when a pattern is inserted above it.

## The coupling — why `quart.yaml` #1 is in this commit and not a follow-up

`flask.yaml` #1 and `quart.yaml` #1 are a **byte-identical** pattern in two rule files: same regex, same `entity_type`, same `name_group`. `seenEntities` (`detector.go:367`) is per file and **shared across rule sets**, and rule sets load in `fs.WalkDir` lexical order (`loader.go:103`), so `flask` wins every mint and `quart`'s twin has never produced anything on any file `flask` also matched.

Gating `flask.yaml` #1 alone therefore removes **zero** entities and hands all 80 mints to Quart:

- before: `flask#1 {django: 75, flask: 5}`
- after: `flask#1 {flask: 1}`, `quart#1 {django: 75, flask: 4}`

Four genuine Flask routes get **relabelled Quart** in a repo with no Quart in it — strictly worse than making no change. And it is invisible to every count: totals, per-kind totals and the over-fire rate all stay flat. It is invisible in the emitted entity too — `detector.go:451` fills the `framework` property from `sp.framework`, which `compile()` sets to the **language** (`python`, `detector.go:171`), not the rule file, so a transfer between two python rule sets cannot be observed in `DetectResult` at all.

It is pinned twice instead:

- **behaviourally** — a marker-free file mints a Route if *either* twin is left ungated, so the forbidden-direction Route rows go red for either half;
- **structurally** — `TestFlaskAndQuartRouteTwinsGatedTogether_7013` asserts the general property: any `source_pattern` appearing in two python rule files with the same regex, `entity_type` and `name_group` must agree on `requires_framework`. There are four such pairs today, all flask/quart (#0 Route `.route`, #1 Route, #2 Controller `Blueprint`, #4 Middleware `errorhandler`); the test also fails if the #1 pair itself disappears, so it cannot go vacuous.

## `frameworkPresent` is per FILE

`detector.go:96 func (c compiledRuleSet) frameworkPresent(content string)`, called at `:402` with `content := string(file.Content)`. A gate kills every match in any file that does not **itself** carry a marker, even in a repo that plainly uses the framework. That is the entire cost profile, and it is why the losses concentrate in alias- and re-export-heavy Python.

Each of the six gated patterns is graded in both directions with the **same construct on both sides**.

**Varied:**
- marker presence (forbidden arm vs recall arm) — the only axis separating the two tables;
- for the Route twin, **which of the two identical rule files** can see its framework (a Flask-marked row and a Quart-marked row over the identical construct);
- marker **spelling** across three widening fixtures: `Flask(__name__)` (pre-existing), `from flask import Blueprint` (widened), `import flask` (widened, qualified-import style);
- the surrounding module for the over-fire check: a Django-shaped module rather than a bare one.

**Held constant:** the construct text, the entity name, the entity kind, the file path, the language, and the rule tree (the real `LoadAllRules()`, the same one Pass 2.5 runs).

**Premises asserted, never assumed:** every forbidden fixture is checked against the on-disk `import_markers` of *all six* relevant rule files (`pytest`, `sqlalchemy`, `flask`, `quart`, plus `falcon`/`cherrypy`, whose own #6152 bare-class gates reach the class fixtures); every recall fixture is checked to contain one. Markers are read from the YAML **files**, not from the compiled Go structs under test, so a loader bug cannot make a premise vacuously true. An empty marker list is a hard failure.

**Construct isolation was enumerated, not assumed.** Each fixture was run against every `source_pattern` in `rules/python/**` to confirm only the intended gated pattern reaches it. That is why the Middleware fixture uses `async def`: `flask.yaml` #5 is a near-twin of `quart.yaml` #5 and is **ungated**, but its `\ndef` cannot match an async hook, so the fixture isolates `quart` #5 instead of being masked by flask's mint.

## Accepted losses — recorded in the YAML, not fixed

- **`sqlalchemy` #5 loses 4.** `flask-realworld`'s `conduit/database.py` re-exports `Column = db.Column` and imports `relationship` centrally, so the model modules contain **no sqlalchemy token at all**. Adding `flask_sqlalchemy` as a marker was tried and rescued **zero**. Structural under per-file semantics. Accepted at 246:4.
- **`pytest` #1 loses 4, of which 3 are real** — pytest classes in `tests/test_*.py` whose fixtures come from `conftest.py` and which legitimately never name pytest. Structural. The 4th, `conduit/settings.py:54 class TestConfig(Config)`, is itself a false positive that the repo-level classifier miscredits; removing it is a win.
- **`flask` #6 loses 1** after the widening — it reaches Flask only via `from flask_jwt_extended import …` and a local re-export, not rescuable without a marker loose enough to over-fire.
- **`quart` #5 loses nothing *here*.** The corpus contains **no Quart user at all** — zero repos classify as using Quart and all 43 bad mints are in Flask repos. This is "loses nothing here", not "loses nothing"; recall for real Quart code is untested by that corpus and rests on the fixture in this PR.
- The widened `import flask` marker matches by substring, so it also reaches `import flask_sqlalchemy`, `import flask_admin` and friends. Those are Flask ecosystem packages, and the widening was measured at **zero** cost to bad-removal on the corpus — but it is a widening, and `TestWidenedFlaskMarkers_DoNotReopenOverfire_7013` is what will catch it if a future marker goes too far.

## Direction of error, carried forward from the measurement

- **762 is a lower bound** — step 2's "uses" classifier is a repo-wide scan, so a vendored blob or a comment marks a repo as *using* and hides over-firing behind it.
- **9 is an upper bound in kind** — a mint counts as "good" only because some unrelated file mentioned the framework, demonstrably so for `TestConfig`.
- **9 is a lower bound in corpus breadth** — only 48 true-positive mints exist across all five patterns in this corpus, across 3 repos.
- **The 7 structural losses come from one repo** (`flask-realworld`). The per-file gate's cost on alias-heavy Python is measured at n=1.

## Independent grounding that the change is not inert

Read-only scan of the `django` corpus repo alone, counting distinct-name matches per file for each gated pattern and asking whether the file carries any marker for the framework it is credited to:

| pattern | matches in django | of which in files with NO marker |
|---|---:|---:|
| `sqlalchemy` #5 `ForeignKey(` | 244 | **244** |
| `pytest` #1 `class Test*` | 322 | **322** |
| `flask` #6 `class X(db.Model\|Model)` | 76 | **76** |

642 mints, **none** of which survives the gate, from one repo — matching the coordinator's per-pattern bad counts (246 / 322 / 76) almost exactly. The removal is real and concentrated exactly where the sweep said it was.

## Mutant scores — 11 applied, 11 DEAD

Each mutant was applied with an **exact occurrence-count assertion** before and after the edit, so a silently-unapplied edit could not read as ALIVE; each was scored with `-count=1`; the tree was restored from a shasum-verified backup between runs (never `git checkout --`).

|---|---|---|---|
| M1 | `pytest` #1 ungated | **DEAD** | forbidden `pytest#1_TestClass` |
| M2 | `sqlalchemy` #5 ungated | **DEAD** | forbidden `sqlalchemy#5_Constraint` |
| M3 | `flask` #6 ungated | **DEAD** | forbidden `flask#6_Model` |
| M4 | **`flask` #1 ungated ALONE** (the #7028 trap) | **DEAD** | both forbidden Route rows + the twin-parity test + the over-fire test |
| M5 | **`quart` #1 ungated ALONE** (the trap, mirrored) | **DEAD** | same four |
| M6 | `quart` #5 ungated | **DEAD** | forbidden `quart#5_Middleware` |
| M7 | widening reverted (both markers dropped) | **DEAD** | `RescueBlueprintModule` |
| M8 | widening narrowed to `from flask import` only | **DEAD** *(was ALIVE — see below)* | `RescuePlainImportModule` |
| M9 | `frameworkPresent` → always `false` (gate kills everything) | **DEAD** | all six recall rows + both rescue tests |
| M10 | `frameworkPresent` → always `true` (gate wired but inert) | **DEAD** | all six forbidden rows + the over-fire test |
| M11 | flask markers widened over-broadly (`"def "` added) | **DEAD** | four forbidden rows + the over-fire test |

**M8 was ALIVE on the first pass and was killed rather than argued away.** Dropping `import flask` and keeping only `from flask import` left every other test green — the second widened marker was carried by nothing, because the blueprint fixture happens to use a `from …` import. It cost ~20 lines to add `TestWidenedFlaskMarkers_RescuePlainImportModule_7013`, a module reaching Flask through `import flask` + `flask.Blueprint(...)`, with a premise asserting neither the pre-widening markers *nor the other widened marker* can see it. That is now the only test that grades `import flask`.

M4 and M5 are the pair that matters: they are the parent's behaviour for their half of the twin, and they turn the forbidden fixture red — positive proof that this change removes entities rather than only pinning what was already true.

## Verification

- `go test -count=1 ./internal/engine/...` — **green** (`internal/engine` 52s, `internal/engine/httproutes`). `gofmt -l internal/engine/` clean, `go vet ./internal/engine/` clean.
- `go test -count=1 ./cmd/grafel/` — **green** (150s), run with a **fresh** `GRAFEL_HOME` and `GRAFEL_DAEMON_ROOT`. Its `Kind|Name|QualifiedName|Subtype|SourceFile|StartLine|EndLine` digest **did not move**, so there is no delta to enumerate: the mixed-language fixture's Python content is framework-honest and every gated pattern that fires there fires on a file carrying its own marker.
- `scripts/quality/run.sh --runs 1 --ratchet` — 38 fixtures held (29 at 100% must-have recall), and, which is the check that actually means something, **`git status --porcelain` is empty afterwards**. No baseline was rewritten and none needed to be. `python-pytest-sqlalchemy-mini` in particular held, which is the recall direction on the two heaviest gates.

### One fixture legitimately moved, hand-edited

`TestZeroProducingKinds6776_EverySiteFires` / `..._EveryNegativeHolds` — the `TestClass` site. Its positive body was `class TestOrders:\n    pass\n` with no pytest marker anywhere, so it asserted the pattern fires on a file the gate now correctly suppresses. Both bodies gained `import pytest`.

The **negative** body gained it too, deliberately. Without the marker the negative would hold for two reasons at once — wrong class name **and** absent framework marker — and would keep passing even if the `Test\w+` half of the pattern were widened to anything at all. With the marker present, the class name is the only thing separating the two bodies, which is what that case's `negWhy` claims. The site's label was also corrected from `pytest.yaml:60` to `:96`.

## Not established

- Recall outside the Python bucket; cross-bucket dedup interactions were not exercised.
- How many identical-twin pairs exist **outside** `rules/python/frameworks` — the new parity test enumerates that directory only. The general sweep is #7028's deliverable, and `flask`#0/`quart`#0 (178 mints) is the same trap still waiting there for whoever tightens either one.
- Whether the widened flask markers over-fire outside this corpus.

### DIFF
--- a/internal/engine/python_framework_gates_7013_test.go
+++ b/internal/engine/python_framework_gates_7013_test.go
@@ -1,0 +1,569 @@
+// Package engine — #7013 variant D.
+//
+// Six one-line rules changes, measured on the 60-repo corpus at 762 over-fired
+// entities removed against 9 true positives lost:
+//
+//	pytest.yaml     #1  TestClass   requires_framework   322 bad : 3 real
+//	sqlalchemy.yaml #5  Constraint  requires_framework   246 bad : 4 real
+//	flask.yaml      #6  Model       requires_framework    76 bad : 1 real
+//	flask.yaml      #1  Route       requires_framework  \  75 bad : 0 real
+//	quart.yaml      #1  Route       requires_framework  /  (one coupled change)
+//	quart.yaml      #5  Middleware  requires_framework    43 bad : 0 here
+//	flask.yaml          import_markers widened with `from flask import`,
+//	                    `import flask` — rescues 5 of the losses, costs 0 bad
+//
+// This file grades BOTH directions for every one of them, because each
+// direction is blind to the other's failure:
+//
+//   - FORBIDDEN: the gated pattern no longer mints in a file that does not
+//     itself carry a marker. A recall-only test cannot see over-firing.
+//   - RECALL: the gated pattern still mints in a file that does carry one. A
+//     forbidden-only test is passed perfectly by a gate that kills everything.
+//
+// ...plus two properties that no count of either direction can observe:
+//
+//   - THE TWIN COUPLING (#7028). flask.yaml #1 and quart.yaml #1 are a
+//     byte-identical pattern in two rule files. seenEntities (detector.go:367)
+//     is per file and shared ACROSS rule sets, and rule sets load in fs.WalkDir
+//     lexical order (loader.go:103), so `flask` wins every mint and `quart`'s
+//     twin has never produced anything. Gating flask's copy alone removes zero
+//     entities and relabels 4 genuine Flask routes as Quart. Totals, per-kind
+//     counts and the over-fire rate all stay flat while that happens, and the
+//     emitted entity cannot show it either: detector.go:451 sets the
+//     `framework` property from `sp.framework`, which compile() fills with the
+//     LANGUAGE (`python`, detector.go:171) and not the rule file — so a
+//     transfer between two python rule sets is invisible in DetectResult by
+//     construction. It is pinned twice below instead: behaviourally, because a
+//     marker-less file mints a Route if EITHER twin is ungated; and
+//     structurally, because the two files' flags must agree.
+//   - THE WIDENING, in both of its halves: that it rescues the blueprint-module
+//     case, and that it does not re-open the over-fire it was added alongside.
+//
+// Fixture discipline: every fixture below carries the SAME construct in the
+// forbidden and the recall arm, so the only thing varied is marker presence.
+// The varied and held-constant axes are listed on each test.
+package engine
+
+import (
+	"context"
+	"os"
+	"strings"
+	"testing"
+
+	"gopkg.in/yaml.v3"
+
+	"github.com/cajasmota/grafel/internal/extractor"
+)
+
+// ---------------------------------------------------------------------------
+// Premise helpers — grounded in the rule FILES on disk, never in the compiled
+// Go structs under test, so a loader bug cannot make a premise vacuously true.
+// ---------------------------------------------------------------------------
+
+// pgRuleFile is a minimal on-disk view of a python framework rule file.
+type pgRuleFile struct {
+	Frameworks struct {
+		Detection struct {
+			ImportMarkers []string `yaml:"import_markers"`
+		} `yaml:"detection"`
+	} `yaml:"frameworks"`
+	SourcePatterns []struct {
+		Pattern           string `yaml:"pattern"`
+		EntityType        string `yaml:"entity_type"`
+		NameGroup         int    `yaml:"name_group"`
+		RequiresFramework bool   `yaml:"requires_framework"`
+	} `yaml:"source_patterns"`
+}
+
+// pgFrameworks is every rule file this change touches. Both flask AND quart are
+// named everywhere they can both matter — enumerating one thoroughly is what
+// makes its twin look covered.
+var pgFrameworks = []string{"pytest", "sqlalchemy", "flask", "quart"}
+
+// pgLoadRuleFile parses one python framework rule file straight off disk.
+func pgLoadRuleFile(t *testing.T, framework string) pgRuleFile {
+	t.Helper()
+	path := "rules/python/frameworks/" + framework + ".yaml"
+	data, err := os.ReadFile(path)
+	if err != nil {
+		t.Fatalf("read %s: %v", path, err)
+	}
+	var doc pgRuleFile
+	if err := yaml.Unmarshal(data, &doc); err != nil {
+		t.Fatalf("parse %s: %v", path, err)
+	}
+	return doc
+}
+
+// pgMarkers returns a framework's on-disk import_markers, failing if the list
+// is empty — an empty list makes every "no marker present" premise trivial and
+// every "marker present" premise unsatisfiable.
+func pgMarkers(t *testing.T, framework string) []string {
+	t.Helper()
+	m := pgLoadRuleFile(t, framework).Frameworks.Detection.ImportMarkers
+	if len(m) == 0 {
+		t.Fatalf("%s declares no frameworks.detection.import_markers — the gate under test has "+
+			"no signal, and every premise here would be vacuous", framework)
+	}
+	return m
+}
+
+// pgAssertNoMarkers asserts src carries no import marker for ANY framework
+// whose gated pattern could mint from it. Without this the forbidden direction
+// could pass for the wrong reason. `falcon` and `cherrypy` are included because
+// their own gated bare-class patterns (#6152) reach the class fixtures too.
+func pgAssertNoMarkers(t *testing.T, src string) {
+	t.Helper()
+	for _, fw := range append(append([]string{}, pgFrameworks...), "falcon", "cherrypy") {
+		for _, m := range pgMarkers(t, fw) {
+			if strings.Contains(src, m) {
+				t.Fatalf("fixture premise broken: source contains %s marker %q, so it is not a "+
+					"marker-free file and cannot grade the forbidden direction", fw, m)
+			}
+		}
+	}
+}
+
+// pgAssertHasMarker asserts src carries at least one marker for framework, the
+// recall direction's premise.
+func pgAssertHasMarker(t *testing.T, framework, src string) {
+	t.Helper()
+	for _, m := range pgMarkers(t, framework) {
+		if strings.Contains(src, m) {
+			return
+		}
+	}
+	t.Fatalf("fixture premise broken: source contains none of %s's import_markers %v, so a "+
+		"mint here would not prove the gate lets real %s code through",
+		framework, pgMarkers(t, framework), framework)
+}
+
+// pgAssertGated asserts on disk that the named framework file carries exactly
+// one source_pattern producing entityType whose regex contains patternSubstr,
+// and that it carries requires_framework.
+//
+// It identifies the pattern by a distinguishing fragment of its regex rather
+// than by entity_type alone, because these files carry several patterns per
+// kind (flask.yaml alone mints Route twice) and by ordinal, because an ordinal
+// silently slides when a pattern is inserted above it.
+//
+// This pins the WIRING: without it, a forbidden-direction pass is equally
+// consistent with "the gate holds" and "someone deleted the pattern".
+func pgAssertGated(t *testing.T, framework, entityType, patternSubstr string) {
+	t.Helper()
+	doc := pgLoadRuleFile(t, framework)
+	found := 0
+	for _, sp := range doc.SourcePatterns {
+		if sp.EntityType != entityType || !strings.Contains(sp.Pattern, patternSubstr) {
+			continue
+		}
+		found++
+		if !sp.RequiresFramework {
+			t.Errorf("%s.yaml pattern %q produces %s but does NOT carry requires_framework — "+
+				"#7013's gate is not wired for it", framework, sp.Pattern, entityType)
+		}
+	}
+	switch {
+	case found == 0:
+		t.Fatalf("%s.yaml has no source_pattern producing %s whose regex contains %q: the "+
+			"pattern this test grades was rewritten or deleted, and the forbidden direction "+
+			"below would pass vacuously", framework, entityType, patternSubstr)
+	case found > 1:
+		t.Fatalf("%s.yaml has %d source_patterns producing %s and matching %q — the fragment no "+
+			"longer identifies one pattern, so this assertion is ambiguous",
+			framework, found, entityType, patternSubstr)
+	}
+}
+
+// pgDetect runs the REAL embedded rule set, the same one Pass 2.5 runs.
+func pgDetect(t *testing.T, path, src string) *DetectResult {
+	t.Helper()
+	rules, err := LoadAllRules()
+	if err != nil {
+		t.Fatalf("LoadAllRules: %v", err)
+	}
+	res, err := New(rules).Detect(context.Background(), extractor.FileInput{
+		Path:     path,
+		Language: "python",
+		Content:  []byte(src),
+	})
+	if err != nil {
+		t.Fatalf("detect: %v", err)
+	}
+	if res == nil {
+		t.Fatal("detect returned nil result")
+	}
+	return res
+}
+
+// pgHas reports whether the detector emitted an entity with this kind and name.
+func pgHas(res *DetectResult, kind, name string) bool {
+	for i := range res.Entities {
+		if res.Entities[i].Kind == kind && res.Entities[i].Name == name {
+			return true
+		}
+	}
+	return false
+}
+
+// pgDump renders every emitted entity, for failure messages.
+func pgDump(res *DetectResult) string {
+	var b []string
+	for i := range res.Entities {
+		b = append(b, res.Entities[i].Kind+":"+res.Entities[i].Name)
+	}
+	if len(b) == 0 {
+		return "(no entities)"
+	}
+	return strings.Join(b, ", ")
+}
+
+// ---------------------------------------------------------------------------
+// The constructs. Each appears verbatim in both a marker-free and a marked
+// fixture, so marker presence is the ONLY axis varied between the two arms.
+// ---------------------------------------------------------------------------
+
+const (
+	// pytest.yaml #1 — `(?m)^class\s+(Test\w+)\s*`.
+	pgTestClassConstruct = "class TestPgProbe:\n    def test_pg_probe(self):\n        assert True\n"
+	// sqlalchemy.yaml #5 — `ForeignKey\s*\(\s*["']([^"']+)["']`.
+	pgConstraintConstruct = "pg_probe_col = ForeignKey(\"pg_probe_table.id\")\n"
+	// flask.yaml #6 — `class\s+(\w+)\s*\(\s*(?:db\.Model|Model)\s*\)`.
+	pgModelConstruct = "class PgProbeModel(db.Model):\n    pass\n"
+	// flask.yaml #1 / quart.yaml #1 — the byte-identical twin.
+	pgRouteConstruct = "@pg_probe_bp.get(\"/pg-probe-route\")\ndef pg_probe_view():\n    return \"ok\"\n"
+	// quart.yaml #5 — `async def` deliberately: flask.yaml #5's near-twin
+	// requires `\ndef` and cannot reach an async hook, so this construct
+	// isolates quart #5 instead of being masked by flask's ungated pattern.
+	pgMiddlewareConstruct = "@pg_probe_app.before_request\nasync def pg_probe_hook():\n    return None\n"
+)
+
+// pgGateCase is one gated pattern, graded in both directions with the same
+// construct on both sides.
+type pgGateCase struct {
+	name string
+	// frameworks whose rule file must carry the gate for this kind. Two for
+	// the Route twin — flask AND quart, named together everywhere.
+	frameworks []string
+	kind       string
+	entity     string
+	construct  string
+	// patternSubstr identifies WHICH pattern in those files is under test, by
+	// a fragment of its regex — see pgAssertGated.
+	patternSubstr string
+	// recallMarker is prepended to construct for the recall arm. It is a
+	// marker of recallFramework and of nothing else.
+	recallFramework string
+	recallMarker    string
+}
+
+var pgGateCases = []pgGateCase{
+	{
+		name:            "pytest#1_TestClass",
+		patternSubstr:   "^class\\s+(Test\\w+)",
+		frameworks:      []string{"pytest"},
+		kind:            "TestClass",
+		entity:          "TestPgProbe",
+		construct:       pgTestClassConstruct,
+		recallFramework: "pytest",
+		recallMarker:    "import pytest\n\n",
+	},
+	{
+		name:            "sqlalchemy#5_Constraint",
+		patternSubstr:   "ForeignKey",
+		frameworks:      []string{"sqlalchemy"},
+		kind:            "Constraint",
+		entity:          "pg_probe_table.id",
+		construct:       pgConstraintConstruct,
+		recallFramework: "sqlalchemy",
+		recallMarker:    "from sqlalchemy import ForeignKey\n\n",
+	},
+	{
+		name:            "flask#6_Model",
+		patternSubstr:   "db\\.Model",
+		frameworks:      []string{"flask"},
+		kind:            "Model",
+		entity:          "PgProbeModel",
+		construct:       pgModelConstruct,
+		recallFramework: "flask",
+		recallMarker:    "app = Flask(__name__)\n\n",
+	},
+	{
+		// The twin, graded with a FLASK marker: flask sorts first and mints.
+		name:            "flask#1_Route_flask_marked",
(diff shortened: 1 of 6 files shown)


cajasmota:
# APPROVE

Independent adversarial review of `a3c66a354` against `main` = `1b115b73b`. Own worktree, tree restored from shasum-verified backups after every mutant, `git status --porcelain` empty at the end. Full `go test -count=1 ./internal/engine/...` green at HEAD (47s + 1.5s), `gofmt -l internal/engine/` clean, `go vet ./internal/engine/` clean.

Nothing here blocks the merge. Two things worth acting on: one prose claim that no artefact carries, and one ALIVE mutant that is a guard against the *next* arm rather than a defect in this one.

---

## 1. Does the quality ratchet grade this? Fixtures genuinely unaffected — and I can say why

Not taken on trust. I dumped every entity the real `LoadAllRules()` detector mints from all 25 `.py` files under `internal/quality/golden/`, once with this branch's four rule files and once with `1b115b73b`'s (`git show` of each, backups restored after). **Byte-identical, zero diff, 28 entities each side.** So the author's "no baseline needed rewriting" is real, not an artefact of the ratchet's known inability to fail on a ceiling drop.

The golden set touches four of the six gated patterns' sites:

| site | pattern | marker in the file | fate |
|---|---|---|---|
| `python-pytest-sqlalchemy-mini/src/tests/test_invoices.py:21` `class TestInvoiceTotals` | pytest#1 | `import pytest` | survives → **recall graded** |
| `.../tests/helpers.py:16` `class TestFromADocstring` | pytest#1 | `import pytest` | survives → **recall graded** |
| `.../app/models.py:35` `ForeignKey("customers.id")` | sqlalchemy#5 | `from sqlalchemy import …` | survives → **recall graded** |
| `python-fastapi-mini/src/app/routers/things.py:8` `@router.get("/things")` | flask#1/quart#1 shape | none | **never belonged to flask#1** — `fastapi` sorts before `flask` in `fs.WalkDir` order, so `fastapi.yaml` owns the `Route:/things` `seenEntities` key either way |

**But the net is partly blind, and this is a real finding about the safety net rather than about the PR:**

- **The forbidden direction is graded by zero golden fixtures.** There is no over-firing Python site anywhere in the golden set — every gated pattern that fires there fires on a file carrying its own marker. The 762 removals are invisible to the ratchet by construction.
- **flask#6 (Model), flask#1/quart#1 (Route) and quart#5 (Middleware) are ungraded in *both* directions** by the ratchet — no golden site exercises them at all.

So `python_framework_gates_7013_test.go` is not a supplement to the ratchet here; for four of six patterns it is the *only* grader. That is an argument for the test file being as thorough as it is, and an argument for a python-flask/quart golden fixture at some point. Filing that is out of scope for this PR.

**Same answer for `cmd/grafel`.** Its one candidate fixture is `cmd/grafel/testdata/audit2678_python/fastapi_app/main.py` — three `@app.<verb>("/…")` decorators, no flask/quart marker. It loses nothing for the same lexical-order reason. The unmoved digest is genuine.

## 2. The coupling — verified from both sides, on the corrected framing

Both halves scored with an exact occurrence-count assertion before and after the edit (a silently-unapplied edit would have read ALIVE):

|---|---|---|
| **MUT-B** — `flask`#1 gated alone (ungate `quart`#1) | **DEAD** | `DoNotMintWithoutMarkers/flask#1_Route_flask_marked`, `…/quart#1_Route_quart_marked`, `TestFlaskAndQuartRouteTwinsGatedTogether_7013`, `TestWidenedFlaskMarkers_DoNotReopenOverfire_7013` |
| **MUT-C** — `quart`#1 gated alone (ungate `flask`#1) | **DEAD** | the same four |

Both kills are the *right* kill under the corrected framing: the forbidden Route rows go red because the entity is **still minted** from a marker-free file — i.e. the tests observe that gating one half removes **zero entities**, which is the hazard, and which no grand-total count can see. Confirmed, not inferred.

## 3. Finding — the "relabelled Quart" claim is asserted by prose and carried by nothing

`detector.go:171` fills `sp.framework` with the **language**. My golden dump confirms it empirically: every source-pattern entity emits `framework=python`; only synthesizer entities carry a real framework (`fw=django` on `SCOPE.ModelEvent`, `fw=fastapi` on `http_endpoint_definition`). A flask→quart twin transfer emits the identical entity. **Nothing is relabelled "Quart" in any artefact a consumer can read.**

The PR body states this correctly in its own "invisible in the emitted entity too" paragraph — but three durable places assert the opposite as if it were observable:

- `flask.yaml`: *"That relabels 4 genuine Flask routes as Quart in a repo with no Quart"*
- `quart.yaml`: *"mislabels 4 real Flask routes as Quart"*
- `TestFlaskAndQuartRouteTwinsGatedTogether_7013`'s failure message: *"relabels the survivors with the other framework … 4 real Flask routes reported as Quart"*

Those are the sentences the next maintainer reads. What is true and sufficient is the measured half: **gating one alone removes zero entities; the mints transfer instead of disappearing.** Please reword the three comments (and `#7028`'s body) to say that. Not a merge blocker — the behaviour and the tests are right; only the justification overclaims.

One genuinely observable delta of the transfer, unmeasured either way: `flask.yaml` declares 3 `file_conventions` entries and `quart.yaml` 1, so entities in files matching flask's globs would see a different `file_convention` property after a transfer. Worth a line in the comment, not a test.

## 4. Recall direction — graded

**MUT-A** — `frameworkPresent` → `return false` (gate kills everything). **DEAD**, and comprehensively: all **six** recall rows red, plus `RescueBlueprintModule`, `RescuePlainImportModule`, and the `6776` TestClass site. A forbidden-only suite would have survived this; this one does not. The recall side is real.

Incidental, worth knowing: under MUT-A the `Route` entity vanishes but `http_endpoint_definition:http:GET:/pg-probe-route` still appears — the httproutes synthesizer is a separate path the gate does not touch, so endpoint coverage does not follow the Route entity down.

## 5. Mutants the author did not run

- **MUT-E — wrong entry, right file.** Moved `flask`#6's `requires_framework` onto `flask`'s `Service` (`Flask(__name__)`) pattern. **DEAD** — `pgAssertGated`'s regex-fragment identification plus its `found > 1` / `found == 0` fatals catch it, and the forbidden `flask#6_Model` row goes red. The "indices verified against the producing `entity_type`, patterns identified by regex fragment not ordinal" claim **holds under test**, not just in prose.

- **MUT-D — scope creep in the gating direction. ALIVE. Distinguishable and ungraded.** Added `requires_framework: true` to **both** halves of the `flask`#0 / `quart`#0 `.route` twin (parity preserved, so the parity test is satisfied). Full `./internal/engine/` suite **green, 53.7s**. Distinguishing input: a marker-free file containing `@app.route("/x")` — mints `Route` today, mints nothing under D. #7028 prices that twin at **178 mints**, and it is precisely the next thing someone will touch on this issue. Nothing in the suite asserts that an *ungated neighbour* still mints without a marker, so this PR pins "the six named patterns are gated" but not "and nothing else is."
  **Price: ~10 lines.** One negative-control row over the `@app.route("/x")` construct in a marker-free file asserting the `Route` **is** minted. Follow-up, not a blocker — D is not this PR's behaviour, and the parity test already stops the asymmetric version.

- **MUT-F — gate `pytest`'s `Fixture` pattern too.** ALIVE, but **equivalent under any input**: that pattern's own regex requires the literal `@pytest.fixture`, which is itself one of pytest's `import_markers`, so the gate can never change its outcome. Self-marking pattern. Recorded to distinguish it from D, which is not.

## 6. The hand-edited fixture — reasoning and edit both check out

`import pytest` is a genuine `pytest.yaml` `import_markers` entry (`- import pytest`, first in the list). Adding it to the **negative** body is right for the stated reason: without it the negative held for two reasons at once and would have survived widening the `Test\w+` half to anything; with it, the class name is the only discriminator, which is what `negWhy` claims. The `:60` → `:96` correction is **right** — `pytest.yaml:96` is the `(?m)^class\s+(Test\w+)\s*` pattern line (verified on disk). And the site is not vacuous: MUT-A turned it red.

## 7. Accepted losses — re-derived independently on the read-only corpus

- **`sqlalchemy`#5's 4.** The two loss-bearing modules `flask-realworld/conduit/{articles,profile}/models.py` contain exactly one framework import line each — `from flask_jwt_extended import current_user` — and **no sqlalchemy token of any kind**. `conduit/database.py` holds the only `from sqlalchemy.orm import`. Not marker-fixable without a marker that means nothing. **Structural. Confirmed.**
- **`pytest`#1's 4.** Exactly four files across the flask repos carry `^class Test*` with no pytest marker: `tests/test_{authentication,profile,articles}.py` plus `conduit/settings.py:54 class TestConfig(Config)`. Reproduces the author's 4-of-which-3-real **exactly**. I also priced the one rescue the write-up did not name: a `test_*.py` basename/convention signal would recover all 3, but **76** django files match `test_*.py`, carry `^class Test*` and no pytest marker — **76:3** is a clearly bad trade, and `frameworkPresent` reads file content only, so it is not expressible anyway. Correctly accepted, not under-explored.
- **The widening's zero cost is independently confirmed.** No `.py` file in `django`, `pandas`, `requests`, `click` or `django-realworld` contains `from flask import` or `import flask` at all — so the widened markers cannot reach any of the over-firing repos. The "rescues 5, costs 0 bad-removal" claim survives an outside check.

## 8. `quart`#5's zero is corpus-relative — correctly hedged

Confirmed: `^\s*(from quart|import quart)` matches **nothing** in the corpus; every `quart` hit is `quarter`/`Quarter`. And the hedge is present everywhere it matters — the PR body says *"loses nothing **here**"*, the table says *"0 *here*"*, and `quart.yaml`'s own comment says *"'loses nothing here', not 'loses nothing' — recall for real Quart code … rests on the fixture below."* No overclaim.

---

## Verdict

**APPROVE.** The measurement reproduces, the losses are structural rather than under-explored, the coupling is pinned in both directions and for the right reason, and the recall direction is genuinely graded (MUT-A kills all six rows).

Two asks, neither blocking:

1. **Reword the three "relabelled Quart" claims** (`flask.yaml`, `quart.yaml`, the twin test's failure string, and #7028's body) to the measured fact — *gating one alone removes zero entities; the mints transfer*. Nothing emits a Quart label.
2. **File the MUT-D follow-up** (~10 lines): a control asserting an ungated `.route` Route still mints in a marker-free file, so the next arm on #7028's `flask`#0/`quart`#0 twin cannot silently widen the gate's scope.

And one thing for the board, not for this PR: **the extraction-quality ratchet grades the forbidden direction of this change at zero fixtures**, and four of the six patterns at zero fixtures in either direction. The gate's entire safety net is the Go test file in this PR.

cajasmota:
## Coordinator mutant on `a3c66a354` — **CV-1 DEAD, and precisely attributed**

Own detached worktree at `/private/tmp/.../coord-mut-7029-y2v/wt` (absolute), private script dir. Green baseline immediately before. Restored by `cp` from a SHA-256-verified backup (`57f04d9b…` both ways); `git diff HEAD` and `git status --porcelain` empty afterwards.

Between you and the reviewer, sixteen mutants have been scored — the coupling from both sides, the recall direction (`frameworkPresent → false`), the wrong-entry-of-the-right-file case, the marker widening. What none of them tests is the property that becomes available only *because* this PR ships six near-identical one-line changes across four files: **cross-pattern independence.**

Six gates spelled the same way in four files is exactly the shape where one fixture quietly stands in for six, and every mutant so far would still pass if that were true.

### CV-1 — remove one gate, leave the other five

Deleted `requires_framework: true` from `sqlalchemy.yaml`'s **`Constraint`** entry only (line 123), leaving its pre-existing `Model` gate and all of pytest/flask/quart intact. Gate count in that file asserted 2 → 1 before running.

**Result: DEAD**, and the attribution is exact:

```
--- FAIL: TestGatedPythonPatterns_DoNotMintWithoutMarkers_7013
    --- FAIL: .../sqlalchemy#5_Constraint
```

**One subtest. Nothing else moved.** No pytest, flask or quart row covers for its neighbour, and the pre-existing `Model` gate in the same file does not cover for the new `Constraint` one either — which is the sharper half, since a file-level fixture would have caught the deletion while grading nothing about *which* entry carries the gate.

The table-driven shape with a regex fragment per row rather than an ordinal is what makes this hold, and it is the same discipline that killed the reviewer's MUT-E.

### The layering, now complete

- **The gate fires at all** → MUT-A (`frameworkPresent → false`), reviewer.
- **The gate is on the right entry of the right file** → MUT-E, reviewer.
- **The twins are coupled** → MUT-B / MUT-C, yours, and they kill for the right reason: the forbidden Route rows see the mint *survive*, not merely a count move.
- **Each gate is graded by its own fixture** → CV-1.
- **The `#0` twin one entry over** → **MUT-D, ALIVE.** Being fixed in this round.

### On the two things I sent back

Neither is a defect in the work. The **"relabelled as Quart"** wording is **my** error, propagated into your YAML and test strings from my brief — the `framework` property comes from the bucket directory name at `detector.go:171`, so both twins emit `framework=python`. The half your mutants actually prove — gating one twin alone removes **zero** entities — is intact and is the real justification. **MUT-D** is a genuinely ungraded neighbour worth ~10 lines.

### What I am filing rather than asking you to fix

The reviewer established the golden fixtures are **partly blind** here: the forbidden direction is graded by **zero** goldens, and `flask`#6, `flask`#1/`quart`#1 and `quart`#5 are ungraded in both directions by them. Your verification was correct — all 25 golden `.py` files are byte-identical at 28 entities each, and the `cmd/grafel` digest genuinely did not move because the fastapi `@router.get` site never belonged to `flask`#1 (`fastapi` sorts before `flask`). The gap is in the safety net, not in this PR, and I would rather have that written down than quietly patched here.

This is the first change on #7013 that **removes** over-firing rather than describing it — 762 entities, 43% of the P1 — and the reason I am comfortable with a change whose whole purpose is to make the graph smaller is that the recall direction was measured before it was written, not defended after.

cajasmota:
## Both items addressed — `2f9c94119`. No gate, marker or coupling changed: the rules diff is comments-only and the four gate counts are unchanged (flask 2, quart 2, pytest 1, sqlalchemy 2).

### 1. The "relabelled as Quart" claim is gone from all three durable places

I re-derived it before rewording rather than taking it on report, and it is false as stated. `compile()` sets `framework: lang` at `detector.go:171`, where `lang` is the key of `d.rules` — the rule bucket — so both twins emit `framework=python`. The only field read out of the `frameworks:` block in the mint path is `Detection.ImportMarkers` (`detector.go:153`); `frameworks.name` is parsed by `schema.go` and read by nothing. **Nothing is relabelled, and there is no per-framework field for a test to assert on.**

`flask.yaml`, `quart.yaml` and `TestFlaskAndQuartRouteTwinsGatedTogether_7013`'s failure string now say what the mutants actually prove: **gating one twin alone removes ZERO entities — the mints TRANSFER to the other rule file rather than disappearing, and no count moves** (80 before, 80 after, 0 over-firing removed). The justification for the coupling is unchanged and intact; it never needed the relabelling claim.

The test header now also states *why* the coupling can only be pinned behaviourally-plus-structurally: the transfer is invisible in `DetectResult` by construction, so there is nothing else to assert on.

**One correction to the correction:** the observable difference between the two files is `file_convention`, and it is **flask 6 / quart 0**, not 3 / 1 — `quart.yaml` declares no `file_conventions:` block at all. Mentioned in the YAML as unmeasured, and deliberately not tested, per your instruction.

### 2. MUT-D killed — `TestRouteDecoratorTwinIsUngated_7013`

Confirmed ALIVE first: gating **both** halves of the `flask.yaml`#0 / `quart.yaml`#0 `.route` twin passed the entire suite.

The new test records the current state rather than changing it, in both halves:

- **structural** — both `.route` Route patterns exist (asserted `seen == 2`, so it cannot go vacuous if one is edited away) and both are ungated on disk;
- **behavioural** — a marker-free `@pg_probe_bp.route("/pg-probe-decorated")` still mints, which is what "ungated" means.

Its doc comment tells whoever turns it red what to do: measure recall first, gate **both** halves, then replace it with a forbidden+recall pair like the #1 rows — rather than deleting it. **#0 is not gated here.** That stays #7028's call, and the 178 mints stay unmeasured for recall.

| arm | verdict | killed by |
|---|---|---|
| MUT-D — both #0 halves gated | **DEAD** | `TestRouteDecoratorTwinIsUngated_7013` |
| MUT-D1 — flask#0 gated alone | **DEAD** | that test + the twin-parity test |
| MUT-D2 — quart#0 gated alone | **DEAD** | that test + the twin-parity test |

### Re-scored after the wording change

M4, M5, M9 and M10 all assert through failure strings I edited. Re-run with `-count=1`, exact occurrence counts asserted before and after each edit, `cp`-restored from shasum-verified backups: **all four still DEAD**, with the same killing tests as before. `git diff HEAD` empty afterwards.

### Verification of this commit

- `go test -count=1 ./internal/engine/...` — green (48.9s). `gofmt -l internal/engine/` clean, `go vet ./internal/engine/` clean.
- `go test -count=1 ./cmd/grafel/` — green (158.6s), fresh `GRAFEL_HOME` + `GRAFEL_DAEMON_ROOT`, digest unmoved. Run despite the rules change being comments-only, rather than claiming a comment cannot move the digest.
- The ratchet was not re-run: the rules diff since the green ratchet pass contains no non-comment line (verified by filtering the diff), and the only other change is a new test file.

Thanks for catching the prose defect — it was load-bearing-looking and would have outlived the PR in three files.

Issue #12, opened 2026-05-08
PORT-10: live verification + migration runbook


cajasmota:
End-to-end verification + post-mortem documentation.

Tasks:
- Install archigraph from a fresh machine (verify install.sh / install.ps1)
- Run `archigraph wizard` with a representative 3-repo group
- Verify watchers, hooks, MCP register correctly
- Run cross-repo benchmark questions:
  - How does feature X flow from mobile UI to API endpoint?
  - How does the frontend aggregate counts from a backend endpoint?
  - How does a background data sync job work?
- Compare token economy vs grep (target: archigraph wins or ties)
- Write post-mortem at `docs/migration/v1.md`
- Write 5+ ADRs covering major decisions
- Update README with usage examples
- Tag v1.0.0 release

Acceptance: cross-repo questions return useful answers under 1k tokens; install flow is one-liner clean.

Effort: 1 day

Depends on: all previous PORT issues

cajasmota:
Implementation in PR #417 (branch `chore/issue-12-runbook`, SHA `0e0cc39`).

**Files written:**
- `docs/migration/v1.md` — v1 post-mortem & migration runbook
- `docs/adrs/0010-structural-refs-format.md`
- `docs/adrs/0011-per-language-bare-name-allowlists.md`
- `docs/adrs/0012-receiver-type-stdlib-dispatch.md`
- `docs/adrs/0013-cross-file-import-aware-resolution.md`
- `docs/adrs/0014-corpus-expansion-strategy.md`
- `README.md` — usage examples + corpus/coverage section

**Verification:**
- `go build ./...` clean
- Forbidden-term grep over diff: clean
- Install-script smoke: read `install.sh` / `install.ps1`; `make build` + `archigraph doctor` + `archigraph --version` exercised locally. End-to-end `curl | bash` flow gated on the v1.0.0 release tag (documented as such in the runbook section 4).

**Not done (per task instructions):**
- v1.0.0 tag — explicitly deferred to after #44 closes
- Hard token-economy numbers — qualitative table in the runbook; the harness is a post-1.0 follow-up

**Issue body cleanup:** updated #12 body to remove the legacy 3-repo group reference and rephrased the cross-repo benchmark questions to be domain-generic.

PR #7032, opened 2026-09-08
fix(#7030): dormant-bucket consultation order was a Go map range, so the framework stamped on a shared yaml entity was randomised per process

### LINKED ISSUE #7030: Go map iteration decides which rule bucket claims a .yaml entity, so the emitted `framework` property flips docker<->kubernetes between runs on identical input (142 entities measured)
Found while enumerating #7028. **On a `.yaml` file matched by two dormant-bucket rule sets, which one wins is decided by Go map iteration order — so the emitted `framework` property changes from run to run on identical input.**

## Mechanism

`dormantBucketAliases` (`internal/engine/detector.go:121`) is a **map**. `compile()` appends buckets onto `d.compiled["yaml"]` under `range` over it. Go randomises map iteration order per process, so the order in which those rule sets are consulted is randomised per process.

`seenEntities` (`detector.go:367`) is per-file and shared across rule sets, so **whichever bucket is consulted first claims the entity** and the other is shadowed. The winner is therefore random.

## Measured instance

`docker/frameworks/docker_compose.yaml#1` and `kubernetes/frameworks/kubernetes_manifests.yaml#12` carry the **byte-identical** pattern `image:\s+(\S+)` (`entity_type: Dependency`), and co-occur on **92 files** for **142 entities**.

Because the two rule sets sit in **different buckets**, and the emitted `framework` property is set from `framework: lang` — the bucket directory name (`detector.go:171`) — the winner determines the property. So those 142 entities are stamped `framework=docker` or `framework=kubernetes` **depending on which way the map iterated that run**.

This is not a stable-but-arbitrary choice. It is a different answer each time the process starts.

## Why this is worth its own issue rather than a note on #7028

- **#7028 is about a dead twin producing nothing.** This is about *non-deterministic output on identical input*, which is a different and worse property: two indexes of an unchanged repo can disagree.
- **It blocks #7028's fix for that pair.** You cannot sensibly pick a winner, or write a regression test asserting one, while the current winner is random. The determinism fix has to land first.
- **It is very likely not the only site.** The same `range`-over-a-map shape appears wherever bucket aliasing is applied; only the `docker`/`kubernetes` pair was measured, because that is the pair the twin enumeration happened to surface. **Nobody has enumerated the others.**

## Related, and probably the same root cause

**#6821** reports entity counts non-deterministic across runs of an identical tree, with `SCOPE.Process` and `SCOPE.External` named. Whether that is this mechanism has **not** been established — but a map-ordered producer race is exactly the shape that would produce it, and anyone picking up #6821 should look here first. Do not merge the two issues until someone has actually traced it; two non-determinism reports sharing a plausible cause is not evidence they share an actual one.

## Suggested first step

Replace the map with an **ordered slice** so bucket consultation order is fixed and reviewable, then decide the `docker` vs `kubernetes` winner deliberately. The natural disambiguator for that specific pair is `file_conventions` (`docker-compose.yml`, … vs `k8s/*.yaml`, …) rather than an import marker — only `docker_compose.yaml` carries a marker at all.

A regression test needs to assert the **order**, not just that some winner exists. A test that runs once and checks "one of them won" passes under the bug.

## Not established

- **How many other pairs are affected.** One measured; the aliasing shape is general.
- **Whether the non-determinism is observable in a shipped graph** as opposed to in the detector's output — the mint path is unconditional (`cmd/grafel/index.go:4403`, no `WithCustomExtractors` branch), which argues yes, but I have not run two full indexes and diffed them. **That is the check that would settle it, and it has not been done.**
- Whether `#6821` is this. See above.

Refs #7028, #6821, #7013.

cajasmota:
Fixes #7030.

> **Revised after review.** The first revision ordered the aliases purely lexically and disclosed only one of the two output flips it shipped. The reviewer was right: I *measured* the second flip and filed it under "#6821 evidence" — as a signal about entity counts — instead of as an impact of my own change. That is the difference between a disclosed trade and a silent regression. The order changed to `cicd, ansible, docker, kubernetes`, the impact section below now names **both** pairs, and the ungraded pair (M8, ALIVE) is now pinned.

## What changed

`dormantBucketAliases` (`internal/engine/detector.go`) was a `map[string][]string`, and `compile()` appended each dormant bucket's compiled rule sets onto the shared target language key under `range` over it. `Detect`'s per-file `seenEntities` lets the **first** rule set that matches an entity claim it, and the emitted `framework` property is the bucket directory name (`framework: lang`) — so on a `.yaml` file matched by two aliased buckets, the property flipped run to run on identical input.

It is now an ordered `[]dormantBucketAlias`, plus a `dormantAliasTargets(bucket)` lookup for the two existing tests that indexed the map. The mechanism is otherwise untouched.

## The order, and why — per pair

**`cicd, ansible, docker, kubernetes`: lexical with one documented exception.** One global order has to serve two shared-target pairs, and no single rule earns both, so the declaration comment argues them separately.

**`docker` vs `kubernetes` — lexical, and a genuine non-choice.** `docker_compose.yaml`#1 and `kubernetes_manifests.yaml`#12 carry the byte-identical `image:\s+(\S+)`, **142 entities**. `image:` in a compose file is docker and in a manifest is kubernetes, so *either* fixed winner is wrong for roughly half of them. The winner was random, so there is no status quo worth preserving and no evidence favouring either. Lexical buys exactly one thing: the pair *has* a stable winner, which is the precondition for choosing one on the merits. **#7028 owns that choice** and wants a `file_conventions`-scoped gate, not a reordering here.

**`cicd` vs `ansible` — `cicd` is out of lexical position, deliberately.** `ansible_core.yaml`#0 and `github_actions.yaml`#3 differ only by an anchor (`(?m)^\s+` vs `\s+`) and shadow each other on **~999 entities** — 7× the docker/kubernetes pair, and **#7028's largest row**. Unlike that pair, **one label here is simply correct**: `archigraph-corpora` holds **90 `.github/workflows/*.y*ml` files and 0 Ansible playbooks**, so every entity in the measured population is CI, not Ansible. Lexical would have pinned `ansible` on all of them, every run — trading a coin-flip mislabel for a permanent one.

Putting `cicd` first is **not** a status-quo reflex. It won ~86–88% of the time under the map and that frequency is noise, not a property. It is the only label anyone has evidence for. The correct long-term fix is to disambiguate or gate the two patterns, exactly as for docker/kubernetes; until then a determinism fix must not ship a new mislabel class as a side effect.

## Impact on output — both pairs

Pinning a winner where there was none moves output by construction. Measured on `main` (`1b115b73b`) with fresh compiles, then on this branch:

| pair | population (#7028) | `main` | this branch | assessment |
|---|---:|---|---|---|
| `docker` / `kubernetes` on `image:\s+(\S+)` | 142 | `map[docker:23 kubernetes:177]` /200 — `kubernetes` **~88%** | `docker` **200/200** | Changes the *typical* output. Accepted: both labels were wrong ~half the time, so no correctness is lost and #7028 decides on the merits. |
| `cicd` / `ansible` on `- name:` (pure GH Actions workflow, zero Ansible content) | ~999 | `map[ansible:48 cicd:352]` /400 — `cicd` **88%** | `cicd` **400/400** | Preserves the correct label for the entire measured population. Under the previous revision this was `ansible` 400/400 — the regression this revision fixes. |

Both `main` figures are my own runs in a detached `main` worktree; the reviewer reproduced both independently at n=400 (`map[docker:57 kubernetes:343]`, and `cicd` 346/400 on the workflow).

## How the tests fail on main — with the evidence

The obvious test does not work: a single `Detect` reports one winner under the bug too, and `-count=N` reruns in the same process, so it is no defence against map-order randomisation. What breaks it is that Go re-randomises the start offset on **every `range` over a map**, so each fresh `compile()` drew an independent order. `TestDormantAliasConsultationOrderIsDeterministic_7030` therefore builds **200 fresh `Detector`s per case** from the real embedded rules and requires all of them to agree.

**Verified by running it on `main`, not asserted.** Probe in a detached `main` worktree:

```
--- FAIL: TestProbe7030MainDeterminism
    NONDETERMINISTIC: 2 distinct framework values across 200 compiles: map[docker:23 kubernetes:177]

--- FAIL: TestZZProbe7030Workflow
    MAIN framework distribution: map[ansible:48 cicd:352] ; entity-count distribution: map[8:400]
    NONDETERMINISTIC on main: map[ansible:48 cicd:352]
```

Both cases fail on `main`; both are single-valued on this branch.

## Three axes, and no guard masks another

`internal/engine/dormant_alias_order_7030_test.go`:

1. **`TestDormantAliasOrderIsExactAndStable_7030`** — the order as an **exact sequence** (index by index, bucket *and* target order, plus length and no duplicate bucket). "Contains these buckets" would pass under the map, which is precisely the bug.
2. **`TestDormantAliasAppendsEveryBucket_7030`** — the other direction: an ordered slice that silently dropped a bucket, or a `compile()` that stopped appending onto the second target, would sail through an order assertion. It checks `len(compiled[target]) == len(rules[target]) + Σ len(rules[bucket])` over every alias naming that target (`compile()` appends exactly one `compiledRuleSet` per `FrameworkRule`, unconditionally). Deliberately **order-independent**, so it cannot mask the order test and vice versa.
3. **`TestDormantAliasConsultationOrderIsDeterministic_7030`** — the behavioural consequence, now **parameterised over both pairs** with a per-pair fixture, deriving the expected winner *from the declared order* rather than hard-coding it. So test 1 grades the order and test 3 grades its consequence, including for a call site that reverses or re-maps a correctly-declared slice.

Distinguishing inputs: **M1** fails only the order test (completeness passes); **M2** fails only completeness among the two structural guards (the order test passes); **M5/M7/M8** fail only determinism.

## Mutants — re-scored against the reordered data

Every edit asserted at an **exact occurrence count** before running and confirmed to differ from the backup (an unapplied edit reading ALIVE is the trap). `detector.go` `cp`-restored from a `shasum`-verified backup after each; final `shasum 97759efd…` matches and `git diff HEAD` is empty. `-count=1` throughout; fresh compiles, never `-count=N`.

| # | mutant | order | complete | determ. | verdict |
|---|---|---|---|---|---|
| M1 | swap `cicd`/`ansible` in the **declaration** | **FAIL** | pass | pass | **DEAD** — order test alone. `[0].bucket = "ansible", want "cicd"`. |
| M2 | `range alias.targets[:1]` | pass | **FAIL** | FAIL | **DEAD** — `compiled["yaml"] holds 14 rule sets, want 16`. Order test passes, so the two structural guards are disjoint. |
| M3 | delete the `kubernetes` entry | **FAIL** | pass | pass | **DEAD** — `has 3 entries, want 4`. Completeness passes (bucket absent from both sides of the count), so only the exact-sequence assertion catches it. |
| M4 | swap `docker`/`kubernetes` in the declaration | **FAIL** | pass | pass | **DEAD** — `[2].bucket = "kubernetes", want "docker"`. Determinism passes *by design* (it derives the winner), so a deliberate winner change is **reported, not silently accepted** — the review gate #7028 needs. |
| M5 | **call site** iterates the slice in reverse | pass | pass | **FAIL** | **DEAD** — determinism alone. `map[kubernetes:200], want "docker"`. |
| M7 | **call site** re-maps the slice into a map and ranges it (the refactor-back regression) | pass | pass | **FAIL** | **DEAD** — determinism alone, and it re-exhibits the bug: `map[docker:175 kubernetes:25]`, `map[ansible:35 cicd:165]`. |
| M8 | **call site** swaps only indices 0/1 (`cicd`⇄`ansible`); declaration untouched | pass | pass | **FAIL** | **DEAD** — was **ALIVE against the whole package** before this revision. Now `map[ansible:200] across 200 compiles, want "cicd"`. This is the reviewer's §4 gap, closed. |
| — | the original defect (map + `range`) | — | — | **FAIL** | **DEAD**, measured on `main` directly (both tables above). Not runnable as an in-branch mutant: reverting the type stops the new tests compiling. M7 is its in-branch stand-in. |

**No survivors.** M8 was the only ALIVE mutant and is now graded.

## Verification

- `go test -count=1 ./internal/engine/...` — **exit 0**, 51s. `gofmt -l` clean, `go vet` clean.
- `go test -count=1 ./cmd/grafel/` with a **fresh** `GRAFEL_HOME` and `GRAFEL_DAEMON_ROOT` — **exit 0**, 169s. **Digest did not move**, and it *cannot*: `entityTupleKey` hashes **eight** fields (`Kind|Name|QualifiedName|Subtype|Language|SourceFile|StartLine|EndLine` — `Language` per #6862, which my first body omitted) and states outright that *"Entity Properties are still NOT hashed."* A property-only change is invisible to it.
- `scripts/quality/run.sh --runs 1 --ratchet` — "38 gated fixtures held (29 at 100% must-have recall)", and the check that actually counts: **`git status --porcelain` empty afterwards**. No fixture moved, so no baseline was hand-edited and no `--update-baseline` was run.
- **No flaky-baseline evidence.** Nothing in the golden set moved in either direction, consistent with neither observer seeing the property: the reviewer confirmed `scripts/quality/ratchet.py` mentions `framework` only in a comment and gates on must-have entity recall. **Nothing in the repo would have caught this defect** — the new determinism test is the only observer of a `framework` property anywhere. That is a repo coverage finding; the coordinator is filing it.

## On #6821 — still a weak signal, still not asserted

Three samples now, all pointing away from this mechanism being #6821's cause, none conclusive:

1. Swapping `ansible`/`cicd` on a mixed yaml changed **only** the `framework` property on two `Task` entities. Count 9 both ways, `Kind|Name` set identical.
2. The `main` workflow probe above: **entity-count distribution `map[8:400]`** — constant across all 400 compiles while the property took 2 distinct values.
3. No `source_pattern` in any of the four dormant buckets declares `scope: Process` or `scope: External`, the scopes #6821 names (`scope:` values there are `file`, `Service`, `Operation`, `Schema`, `Component`).

The reviewer's independent probes agree (counts constant at 3, 13 and 8 while up to 3 property signatures appeared). Three samples of hand-written yaml is not a corpus, and the `ansible`/`cicd` pair is near-identical rather than identical, so on other content their name *sets* could differ and counts could move. **Do not merge #6821 into this until someone traces it.** Not widened here.

## Not established

- **How many other pairs cross an aliased-bucket boundary on a shared pattern.** #7028's census names the two fixed-order pairs above; I pinned the order for all four buckets and did not enumerate further pairs.
- **Whether either pinned winner is *right*.** For `docker`/`kubernetes`, out of scope by construction — #7028. For `cicd`/`ansible`, `cicd` is right for the *measured* population (90 workflow files, 0 playbooks); a corpus containing real playbooks would mislabel them `cicd`, and the real fix is still to disambiguate the two patterns.

### DIFF
--- a/internal/engine/detector.go
+++ b/internal/engine/detector.go
@@ -118,11 +118,59 @@ func (c compiledRuleSet) frameworkPresent(content string) bool {
 //     docker_compose.yaml rules target compose files → `yaml`. The bucket is
 //     indivisible at load time, so it is aliased onto both; non-matching
 //     patterns are inert.)
-var dormantBucketAliases = map[string][]string{
-	"cicd":       {"yaml"},
-	"ansible":    {"yaml"},
-	"kubernetes": {"yaml"},
-	"docker":     {"dockerfile", "yaml"},
+//
+// # This is an ORDERED SLICE, not a map, and the order is load-bearing (#7030)
+//
+// compile() APPENDS each bucket's compiled rule sets onto the shared target
+// key, and Detect's per-file `seenEntities` (see below) lets the FIRST rule set
+// that matches an entity claim it — the later one is silently shadowed. The
+// emitted `framework` property comes from the bucket directory name
+// (`framework: lang`), so for two buckets aliased onto the SAME target the
+// consultation order decides the property that gets stamped.
+//
+// While this was a `map[string][]string`, `range` over it randomised that order
+// per process. `docker/frameworks/docker_compose.yaml`#1 and
+// `kubernetes/frameworks/kubernetes_manifests.yaml`#12 carry the byte-identical
+// pattern `image:\s+(\S+)` (entity_type Dependency) and co-occur on 92 files /
+// 142 entities, so those entities came out `framework=docker` or
+// `framework=kubernetes` depending on the map seed: two indexes of an unchanged
+// tree could disagree.
+//
+// The order below is LEXICAL BY BUCKET NAME. That is a deliberate non-choice:
+// it is the same rule the loader already uses to order rule FILES
+// (fs.WalkDir, loader.go), it is mechanically re-derivable by anyone reading
+// this list, and it does not smuggle in a claim that `docker` deserves to beat
+// `kubernetes` on the shared `image:` pattern. It only guarantees that the pair
+// HAS a stable winner, which is the precondition for choosing one on the
+// merits (#7028 owns that choice, and wants a file_conventions-scoped gate
+// rather than a reordering).
+//
+// So: reordering these entries CHANGES EXTRACTION OUTPUT for any two buckets
+// sharing a target language. Do it deliberately, with a recall measurement, and
+// update TestDormantAliasOrderIsExactAndStable_7030 — which pins this exact
+// sequence — in the same change.
+var dormantBucketAliases = []dormantBucketAlias{
+	{bucket: "ansible", targets: []string{"yaml"}},
+	{bucket: "cicd", targets: []string{"yaml"}},
+	{bucket: "docker", targets: []string{"dockerfile", "yaml"}},
+	{bucket: "kubernetes", targets: []string{"yaml"}},
+}
+
+// dormantBucketAlias is one bucket→concrete-language(s) alias entry.
+type dormantBucketAlias struct {
+	bucket  string
+	targets []string
+}
+
+// dormantAliasTargets returns the concrete languages a bucket is aliased onto,
+// or nil when the bucket is not aliased at all.
+func dormantAliasTargets(bucket string) []string {
+	for _, a := range dormantBucketAliases {
+		if a.bucket == bucket {
+			return a.targets
+		}
+	}
+	return nil
 }
 
 // Detector applies YAML-driven framework extraction rules to source files.
@@ -312,12 +360,16 @@ func (d *Detector) compile() {
 	// frameworks/*.yaml files carry engine schema keys (source_patterns /
 	// file_conventions / relationship_rules) — they are documentation-only
 	// descriptors — so aliasing it onto `html` would add zero extraction.
-	for bucket, targets := range dormantBucketAliases {
-		sets, ok := d.compiled[bucket]
+	//
+	// Iterated in the fixed slice order declared at dormantBucketAliases — see
+	// the comment there for why the order decides output and must not drift
+	// (#7030).
+	for _, alias := range dormantBucketAliases {
+		sets, ok := d.compiled[alias.bucket]
 		if !ok || len(sets) == 0 {
 			continue
 		}
-		for _, target := range targets {
+		for _, target := range alias.targets {
 			d.compiled[target] = append(d.compiled[target], sets...)
 		}
 	}
--- a/internal/engine/detector.go
+++ b/internal/engine/detector.go
@@ -136,22 +136,47 @@ func (c compiledRuleSet) frameworkPresent(content string) bool {
 // `framework=kubernetes` depending on the map seed: two indexes of an unchanged
 // tree could disagree.
 //
-// The order below is LEXICAL BY BUCKET NAME. That is a deliberate non-choice:
-// it is the same rule the loader already uses to order rule FILES
-// (fs.WalkDir, loader.go), it is mechanically re-derivable by anyone reading
-// this list, and it does not smuggle in a claim that `docker` deserves to beat
-// `kubernetes` on the shared `image:` pattern. It only guarantees that the pair
-// HAS a stable winner, which is the precondition for choosing one on the
-// merits (#7028 owns that choice, and wants a file_conventions-scoped gate
-// rather than a reordering).
+// The order below is LEXICAL BY BUCKET NAME, WITH ONE DOCUMENTED EXCEPTION.
+// Two pairs of buckets share the `yaml` target and share a pattern, so the
+// order decides a label for each; they are argued separately on purpose,
+// because one global order has to serve both and no single rule earns both.
+//
+//   - docker vs kubernetes — LEXICAL, and genuinely a non-choice. `image:` in
+//     a compose file is docker and in a manifest is kubernetes, so EITHER fixed
+//     winner is wrong for roughly half the 142 entities. There is no status quo
+//     worth preserving (the winner was random) and no evidence favouring
+//     either, so lexical picks one without smuggling in a claim. All it buys is
+//     that the pair HAS a stable winner, which is the precondition for choosing
+//     one on the merits. #7028 owns that choice and wants a
+//     file_conventions-scoped gate (`docker-compose.yml` vs `k8s/*.yaml`),
+//     NOT a reordering here.
+//
+//   - cicd vs ansible — `cicd` IS OUT OF LEXICAL POSITION, DELIBERATELY.
+//     `ansible/…/ansible_core.yaml`#0 and `cicd/…/github_actions.yaml`#3 differ
+//     only by an anchor (`(?m)^\s+` vs `\s+`) and shadow each other on ~999
+//     entities — 7x the docker/kubernetes pair, and #7028's largest row. Unlike
+//     that pair, ONE label here is simply correct: the measured population is
+//     90 `.github/workflows/*.y*ml` files and ZERO Ansible playbooks, so every
+//     one of those entities is CI, not Ansible. Lexical would pin `ansible` on
+//     all of them, every run — trading a coin-flip mislabel for a permanent
+//     one. Putting `cicd` first is NOT a status-quo reflex (it happened to win
+//     ~86% of the time under the map, and that frequency is noise, not a
+//     property); it is the only label anyone has evidence for.
+//
+//     The correct long-term fix is to disambiguate the two patterns or gate
+//     them on file_conventions, exactly as for docker/kubernetes. Until then a
+//     determinism fix must not ship a new mislabel class as a side effect.
 //
 // So: reordering these entries CHANGES EXTRACTION OUTPUT for any two buckets
 // sharing a target language. Do it deliberately, with a recall measurement, and
-// update TestDormantAliasOrderIsExactAndStable_7030 — which pins this exact
-// sequence — in the same change.
+// update TestDormantAliasOrderIsExactAndStable_7030 (which pins this exact
+// sequence) and TestDormantAliasConsultationOrderIsDeterministic_7030 (which
+// pins the observable consequence for BOTH pairs) in the same change.
 var dormantBucketAliases = []dormantBucketAlias{
-	{bucket: "ansible", targets: []string{"yaml"}},
+	// cicd before ansible: out of lexical order on purpose — see above.
 	{bucket: "cicd", targets: []string{"yaml"}},
+	{bucket: "ansible", targets: []string{"yaml"}},
+	// lexical from here.
 	{bucket: "docker", targets: []string{"dockerfile", "yaml"}},
 	{bucket: "kubernetes", targets: []string{"yaml"}},
 }
--- a/internal/engine/docker_ansible_multiline_anchor_6927_test.go
+++ b/internal/engine/docker_ansible_multiline_anchor_6927_test.go
@@ -87,7 +87,7 @@ func rules6927(t *testing.T, frameworkName, wantBucket string) map[string][]Fram
 		t.Fatalf("expected exactly one %q rule set in the %q bucket, got %d",
 			frameworkName, wantBucket, len(out[wantBucket]))
 	}
-	if targets := dormantBucketAliases[wantBucket]; !containsStr6927(targets, "yaml") {
+	if targets := dormantAliasTargets(wantBucket); !containsStr6927(targets, "yaml") {
 		t.Fatalf("bucket %q is no longer aliased onto yaml (%v); these tests hand Detect "+
 			"Language=yaml and would grade nothing", wantBucket, targets)
 	}
--- a/internal/engine/dormant_alias_order_7030_test.go
+++ b/internal/engine/dormant_alias_order_7030_test.go
@@ -1,0 +1,216 @@
+package engine
+
+import (
+	"context"
+	"sort"
+	"testing"
+
+	"github.com/cajasmota/grafel/internal/extractor"
+)
+
+// #7030 — dormant-bucket consultation order.
+//
+// dormantBucketAliases used to be a map, and compile() appended each bucket's
+// compiled rule sets onto the shared target language key under `range`. Go
+// randomises map iteration order, so the order in which those buckets were
+// consulted was random per range. Detect's per-file seenEntities lets the FIRST
+// rule set that matches an entity claim it, and the emitted `framework`
+// property is the bucket directory name — so on a yaml file matched by two
+// aliased buckets, the property flipped from run to run on identical input.
+//
+// The three tests below grade the fix on three independent axes, and each fails
+// on an input the others survive:
+//
+//   - Order: the exact declared sequence. Reordering two entries fails this and
+//     nothing else — the completeness test is order-independent by construction.
+//   - Completeness: every listed bucket's sets still reach every one of its
+//     targets. Making compile() append only the first target (or skip a bucket)
+//     while leaving the declaration untouched fails this and nothing else.
+//   - Determinism: the observable consequence. This is the one that fails on
+//     main; the other two are structural pins that main would pass.
+
+// wantDormantAliasOrder7030 is the exact expected sequence. Written out here
+// rather than derived from the production value so that a change to the
+// production value is what the test reports.
+var wantDormantAliasOrder7030 = []dormantBucketAlias{
+	{bucket: "ansible", targets: []string{"yaml"}},
+	{bucket: "cicd", targets: []string{"yaml"}},
+	{bucket: "docker", targets: []string{"dockerfile", "yaml"}},
+	{bucket: "kubernetes", targets: []string{"yaml"}},
+}
+
+// TestDormantAliasOrderIsExactAndStable_7030 pins the consultation order as an
+// EXACT SEQUENCE, not as a set. "contains these buckets" would pass under the
+// map, which is precisely the bug.
+func TestDormantAliasOrderIsExactAndStable_7030(t *testing.T) {
+	if got, want := len(dormantBucketAliases), len(wantDormantAliasOrder7030); got != want {
+		t.Fatalf("dormantBucketAliases has %d entries, want %d: %v",
+			got, want, dormantBucketAliases)
+	}
+	for i, want := range wantDormantAliasOrder7030 {
+		got := dormantBucketAliases[i]
+		if got.bucket != want.bucket {
+			t.Errorf("dormantBucketAliases[%d].bucket = %q, want %q — the order is "+
+				"load-bearing (it decides which bucket's `framework` property wins on a "+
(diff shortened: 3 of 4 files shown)


cajasmota:
# REQUEST-CHANGES — independent review

The mechanism, the fix and the three-axis test design are all correct, and I reproduced every determinism claim from scratch. What blocks is not the code: **the PR discloses one of the two output flips it ships, and the one it omits is the larger one and the one where the pinned winner is demonstrably the wrong label.**

Scored on `9c0a0efe5`, in my own detached worktree; `main` = `1b115b73b` in a second one. `internal/engine` green (`go test -count=1`, 56.8s), `gofmt -l` and `go vet` clean, `git status --porcelain` empty after every mutant (`detector.go` shasum `877a3ef4…` restored each time).

## 1. The defect and the fix — reproduced, both directions

400 fresh `Detector`s per run, my own probe, not the author's:

| input | `main` (1b115b73b) | branch |
|---|---|---|
| `image: nginx:1.25` compose yaml | `map[docker:57 kubernetes:343]` — **2 distinct signatures** | `map[docker:400]`, 1 signature |
| mixed compose/k8s/actions/ansible yaml | `docker:42 kubernetes:358`, **3 distinct signatures** | 1 signature |
| pure GitHub Actions workflow | **2 distinct signatures** (see below) | 1 signature |

So: the bug is real, `-count=N` genuinely would not have caught it, and the fix genuinely pins it. **kubernetes won 85.75% and 89.5% in my two samples** — the reviewer's "~88%" and the author's 177/200 are the same number. The 142-entity figure from #7028 is the population; I did not re-derive it, but nothing depends on the exact count.

## 2. BLOCKING — the undisclosed flip: `.github/workflows/*` is now stamped `framework=ansible`

`ansible` and `cicd` are also both aliased onto `yaml`, and lexical order puts `ansible` **first**. On a **pure GitHub Actions workflow with no Ansible content at all** (`name:`, `on:`, `jobs:`, `runs-on:`, `uses: actions/checkout@v4`, two `- name:` steps):

```
main,   346/400 runs:   Task|Install deps|cicd     Task|Run unit tests|cicd     Config|ci|cicd
main,    54/400 runs:   Task|Install deps|ansible  Task|Run unit tests|ansible  Config|ci|ansible
branch, 400/400 runs:   Task|Install deps|ansible  Task|Run unit tests|ansible  Config|ci|ansible
```

This is #7028's largest row — `ansible_core#0 >> github_actions#3`, **999 entities**, seven times the docker/kubernetes pair — and this PR pins it to `ansible`, permanently, having been `cicd` ~86% of the time. `archigraph-corpora` contains **90 `.github/workflows/*.y*ml` files and 0 Ansible playbooks or `roles/*/tasks/*`**. Every entity in the measured population comes from a GitHub Actions workflow, and they are now all labelled `ansible`.

The PR body did observe this flip — it is the "2 Tasks changed `framework` ansible→cicd" sample — but reports it **only as evidence about #6821**, never in the impact section. The body's impact paragraph names the docker/kubernetes pair and stops. So "a deliberate non-choice" is doing exactly the work the board's most-filed defect does: the prose asserts neutrality the code does not have.

For the record, "non-choice" *is* defensible for docker/kubernetes — `image:` in a compose file is docker, in a k8s manifest is kubernetes, either fixed winner is wrong half the time, and there is no status quo worth preserving. It is not defensible for ansible/cicd, where one label is right for the entire measured population and lexical order picks the other one.

**My view on lexical-vs-majority.** One global order has to serve both pairs, so treat them separately rather than reaching for one rule:

- docker/kubernetes: keep the lexical winner. Nothing is lost, #7028 owns the real fix (a `file_conventions` gate).
- ansible/cicd: **preserve the majority — put `cicd` before `ansible`.** Not because 86% is a property worth keeping (it is not), but because it is the label that is correct on the only population anyone has measured, and pinning determinism should not also ship a new mislabel class to ~999 entities. The cost is one documented exception to "lexical everywhere", which is cheaper than the regression.

Order `cicd, ansible, docker, kubernetes` gets both, with a comment saying why `cicd` is out of lexical position. If you would rather not decide here, the minimum I can approve is: state the ansible/cicd flip in the body as an impact, file the follow-up, and add the pin in §4 — but shipping the flip silently is what I am declining.

## 3. Mutants — the disjointness claim holds, and the call site is pinned

Every edit asserted at exactly 1 occurrence before running, `perl -0pi`, verified applied by `diff` against the backup (an unapplied edit reading ALIVE was the trap here), each of the three new tests run **alone**:

| # | mutant | order | complete | determinism | verdict |
|---|---|---|---|---|---|
| M1 | swap `ansible`/`cicd` in the declaration | **FAIL** | PASS | PASS | **DEAD** — order test alone. Author's claim confirmed. |
| M2 | `range alias.targets[:1]` | PASS | **FAIL** (`14 … want 16`) | FAIL | **DEAD**. Order test passes, so the two *structural* guards are disjoint exactly as the body claims — the body scopes it correctly ("among the two structural guards"). Determinism also fails; harmless, it does not create masking. |
| M5 | **call site iterates the slice in reverse**; declaration untouched | PASS | PASS | **FAIL** (`kubernetes:200, want docker`) | **DEAD** — determinism alone. |
| M7 | **call site re-maps the slice into a map and ranges it** (the refactor-back regression) | PASS | PASS | **FAIL** (`map[docker:174 kubernetes:26]`) | **DEAD** — determinism alone. |

So the reviewer's §4 concern is answered: because `TestDormantAliasConsultationOrderIsDeterministic_7030` reads the expected winner *off the declaration* and asserts it against *observed output*, a caller that reverses or re-maps a correctly-declared slice is caught. That is a genuinely good test design and worth keeping.

## 4. ALIVE, distinguishable and ungraded — the ansible/cicd behaviour

| M8 | call site swaps **only** indices 0/1 (`ansible`⇄`cicd`); declaration untouched | PASS | PASS | PASS |

**ALIVE against the entire `internal/engine` package** (`go test -count=1 ./internal/engine/` → `ok`, 56.9s, mutant confirmed applied: 3 `aliases8` occurrences). It changes real output — the 999-entity population above flips label — and nothing observes it. The determinism test hard-selects `docker`/`kubernetes`; the largest pair is pinned by *declaration* only.

**Price: ~15 lines.** Parameterise the determinism test over `{pair, fixture}` — add `{"ansible","cicd"}` with the pure-workflow yaml above and assert the `Task` framework equals the first of the two in declared order. Same shape as the existing case, kills M8, and makes the ansible/cicd choice reviewable instead of incidental.

## 5. The digest argument — correct, and nothing else covers it either

Read, not guessed. `entityTupleKey` (`cmd/grafel/custom_extractor_spans_6118_test.go:408`) is **eight** fields, not the seven the PR body lists — `Language` is in there (#6862) and the body omits it. Immaterial to the conclusion, which the code states outright two lines above: *"Entity Properties are still NOT hashed."* A property-only change cannot move the digest. Confirmed.

I checked the other candidate: `scripts/quality/ratchet.py` mentions `framework` exactly once, in a comment. The gate measures must-have entity recall and does not observe the property either. **So nothing in the repo would have caught this defect**, and the new determinism test is the only observer of a `framework` property anywhere. That is a coverage finding about the repo, not about this PR, and worth its own issue.

## 6. The #6821 hedge is honest

Every probe I ran held entity **count** constant across all 400 compiles (3, 13 and 8 respectively) while up to 3 distinct property signatures appeared. That is a second, independent sample agreeing with the author's: this mechanism moves properties, not counts, on the content tried. It remains a *weak* signal — two samples of hand-written yaml is not a corpus — and the body says so and declines to assert. No change needed; do not let anyone read it as a conclusion in either direction.

## Summary

Fix: correct. Tests: well designed, disjoint, and they pin the call site rather than just the data. Blocking: the ansible/cicd flip is unmentioned in the impact section, pins the wrong label for the only population measured, is seven times the size of the pair that *is* disclosed, and is ungraded (M8 ALIVE). Reorder it or disclose it, and add the pin either way.

cajasmota:
## Coordinator mutant on `b78e7f6c6` — **CW-1 DEAD, against an independent literal**

Own detached worktree at `/private/tmp/.../coord-mut-7032-r7q/wt` (absolute), private script dir. Green baseline immediately before. Restored by `cp` from a SHA-256-verified backup (`853a52e8…` both ways); `git diff HEAD` and `git status --porcelain` empty afterwards.

Twelve mutants have been scored across you and the reviewer — order swaps at the declaration and the call site, truncation, bucket drop, call-site reverse, call-site re-map. So I went after the property that decides whether *any* of those verdicts mean anything: **is the expected order an independent literal, or is it derived from the thing under test?**

That matters here specifically because your report said the determinism test "derives the winner from the declared order". If that were the only check, swapping the declaration would swap the expectation with it and the mutant would pass — the self-referential shape that has bitten this board before, where a table test driven by the production collection cannot observe a change to that collection.

### First, a hazard that is closed by construction

`dormantAliasTargets()` (`detector.go:193`) was added for the two tests that previously indexed the map. It **ranges over `dormantBucketAliases` itself** rather than carrying its own copy, so a test using the helper and production using the slice cannot disagree. Worth stating explicitly: extracting a lookup for tests is normally exactly how the two views drift apart, and this one cannot.

### CW-1 — swap `cicd`/`ansible` in the production declaration only

Asserted applied by line position (`ansible` at 177, `cicd` at 178 — i.e. `ansible` genuinely first) rather than by a diff I eyeballed, since a silently-unapplied edit reads as ALIVE and argues the opposite of the truth.

**Result: DEAD**, and by the right mechanism:

```
--- FAIL: TestDormantAliasOrderIsExactAndStable_7030
    dormantBucketAliases[0].bucket = "ansible", want "cicd" — the order is load-bearing
    dormantBucketAliases[1].bucket = "cicd", want "ansible" — the order is load-bearing
```

`dormant_alias_order_7030_test.go:37` holds an **independent literal copy** of the expected slice. So the expectation does not follow the declaration, and both positions are checked rather than just the pair's membership. The determinism test's derived winner is a second, weaker layer on top of a hard literal — not the only check. Your report undersold it.

The failure message also names *why* the order matters and instructs the reader to change it deliberately. That is the thing that makes an order assertion survive contact with the next person: without it, a future author reads a red literal as a stale test and updates it.

### The layering, complete

- **The order is exactly this, positionally** → CW-1, against an independent literal.
- **The call site consults the slice in its declared order** → reviewer's M5 (reverses) and M7 (re-maps into a map), both DEAD by the determinism test alone.
- **Every bucket is still consulted** → M2 (`targets[:1]`), DEAD by completeness alone with order passing.
- **The specific `ansible`/`cicd` pair is graded** → M8, ALIVE in round 1, now DEAD.
- **The defect itself** → measured on `main`, not argued: 400 fresh compiles, `map[docker:57 kubernetes:343]`.

Disjoint kill sets throughout — order alone, completeness alone, determinism alone — which is what makes this set non-vacuous rather than merely all-green.

### On the round

The blocking finding was that lexical order pinned **999 workflow entities** to `ansible` on a corpus with **zero** Ansible playbooks — 7× the docker/kubernetes population you disclosed — and that you had measured that flip and filed it under "#6821 evidence" rather than as your own impact. You accepted it squarely, said so in the body, and your own re-measurement (`map[ansible:48 cicd:352]`) independently matches the reviewer's 346/400.

Two things I want on the record:

- **The per-pair argument now lives in the declaration comment, not the PR body.** `cicd` is out of lexical position because it is the correct label for the only measured population — explicitly *not* because 86% is a property. Whoever reorders this next reads the reasoning at the line they are editing, which is the only place it reliably gets read.
- **You held the scope line.** You did not pick the docker/kubernetes winner (still #7028's, and it needs its own recall measurement), did not touch the alias mechanism further, and kept the #6821 hedge weak and unasserted through a third sample where the entity-count distribution stayed constant at `map[8:400]`. A third consistent sample is a reason to keep hedging carefully, not to upgrade the claim.

**Merging on green** — Windows leg outstanding.

PR #7034, opened 2026-09-08
fix(#6916): Tier C — 9 name_group:0 rules mint a marker, not an entity: delete them (79 corpus entities, 0 edges), and correct the "already dropped downstream" claim

cajasmota:
## Tier C of #6916 — 9 sites deleted, per-site measurement first

Nine `name_group: 0` source_patterns are **deleted**, not renamed. Each minted an
entity whose `Name` was the whole regex match: a per-file boolean marker naming a
call site, with no referent for any edge to point at. `name_group: 0` sites in
`internal/engine/rules/`: **29 → 20** (`grep -cE '^ +name_group: 0$'`, so the count
is of rule lines, not of comment mentions).

Instrument: the real `engine.Detector` (`LoadAllRules` + `Detect`, i.e. *post*
`applyPrecisionDedup`) over **12,605 files** of `archigraph-corpora` (60 repos,
read-only, 3-way concurrency), run twice — once from a detached worktree at the
parent `1b115b73b` and once at this branch — with the two binaries' shasums
differing, so the rule change is provably in the embedded FS of the second.

### Per-site: what it mints, and whether anything binds

| site | kind | marker Name | entities on corpus | edges incident | disposition |
|---|---|---|---:|---:|---|
| `rust/.../actix_web.yaml:90` | Config | `#[actix_web::main]` | **69** | **0** | deleted |
| `kotlin/.../kmp.yaml:83` | Config | `sourceSets {` | **10** | **0** | deleted |
| `ruby/.../sinatra.yaml:73` | Service | `helpers do` | 0\* | 0 | deleted |
| `ruby/.../sinatra.yaml:93` | Middleware | `before '/admin/*' do` | 0\* | 0 | deleted |
| `ruby/.../sinatra.yaml:99` | Middleware | `after do` | 0\* | 0 | deleted |
| `ruby/.../sinatra.yaml:117` | Config | `Sinatra::Application.run!` | 0\* | 0 | deleted |
| `js_ts/.../langchain.yaml:42` | Operation | `RunnableSequence.from(` | 0\* | 0 | deleted |
| `js_ts/.../langchain.yaml:66` | Operation | `tool(async (` | 0\* | 0 | deleted |
| `js_ts/.../langchain.yaml:71` | Operation | `.bindTools(` | 0\* | 0 | deleted |

**\* A zero here is CORPUS-RELATIVE, never a capability claim.** The corpus holds no
Sinatra app and no LangChain-JS repo. All seven zero-rows are **proven live** on a
fixture driven through the real detector at the parent commit — that is what the
new test's parent-run failure (below) demonstrates.

**Whole-corpus delta, parent → HEAD:** total entities **13,725 → 13,646 (−79)**;
`Config` **1,655 → 1,576**; entity-name deltas **exactly two** (`#[actix_web::main]`
69→0, `sourceSets {` 10→0); **relationship-endpoint deltas: zero**. Nothing else
moved. Bind status is therefore **unbound for all nine** — every one is dead weight
in the edge sense, and two of them are dead weight at volume.

Bind status was also settled statically, since a bind requires exact Name equality:
only three rule files in these four language buckets name any of these kinds as a
relationship endpoint, and each sources on something the marker Name can never
equal — sinatra `Config ROUTES_TO Controller` on a mount **path**, `Controller
REGISTERED_ON Middleware` on a **module constant**, `Route CALLS Service` on a
`<name>_helper` **method**, express `Operation ROUTES_TO Controller` on a **path**.
`javascript_typescript/frameworks/langchain.yaml` has `relationship_rules: []`.

### Correction to the issue: "already dropped downstream" is FALSE

Tier C's stated reason for the four langchain sites was that they are already
suppressed as Operation noise, so the rules are dead weight. **Verified rather than
relayed, and it does not hold for the three JS sites.**
`dropStatementNoiseOperations` keys on a leading `@`, a bare `=`, or a leading
statement keyword — a trailing-paren call fragment has none of those. All three
names sail through `Detect` and reach the graph. `precision_dedup.go`'s own doc
comment in fact listed them as names that **"must be KEPT"**. So deleting these
three is a **deliberate recall reduction**, labelled as such in the rule comments
and graded here; the site the pass was actually written for is the **python**
`@tool` decorator, which *is* dropped (case 1).

`internal/engine/precision_dedup.go` is updated for that: three of the four
examples its comment cited no longer exist. Doc-only; the predicate is unchanged
and `precision_dedup_test.go`'s rows still pass untouched — `{"RunnableSequence.from(",
false}` and `{".bindTools(", false}` still pin the filter's **narrowness** even though
no rule mints those names now, so the comment edit un-grades nothing.

**How the review adjudicated "must be KEPT", which is better than how this PR put it:**
that sentence is a **blast-radius instruction for the filter**, not an endorsement of the
entities. `dropStatementNoiseOperations` classifies by name shape alone with no idea
which rule minted a name, so widening it to catch trailing-paren fragments would also
have killed trpc's `createTRPCClient<` and every other legitimate call-idiom Operation.
"Must be KEPT" constrains the *filter*; deleting the rule is the surgical alternative the
filter could not perform. Keeping a rule alive because a filter was told not to catch it
inverts the dependency.

### The python langchain site was NOT touched — it is out of my lane

`python/frameworks/langchain.yaml:43` (`@tool`) is the tenth line the Tier C list
names, and `internal/engine/rules/python/frameworks/**` is off-limits to this arm
(unmerged PR). It is also the one site where "already dropped downstream" is TRUE,
so deleting it is free whenever the file is available. **1 site remains in Tier C.**

### What was deleted, and what was deliberately kept (varied vs held constant)

**Varied** (one axis: the nine marker rules, deleted). **Held constant**: every
other pattern in the four files, every `name_group` on a surviving rule, all four
`frameworks:` detection blocks, and all fixture *sources* — the golden fixture's
`src/` tree is untouched, which is what keeps the new forbidden rows reachable.

Each file keeps siblings, pinned by name in `TestIssue6916_TierCSurvivingSiblings`:

- `actix_web.yaml` — **keeps** `Config "HttpServer::new("` and `Config "App::new()"`.
  Both are still `name_group: 0` and still ugly: they are **Tier D** (a thing could
  be captured), not Tier C. Only the attribute macro, which has no name to capture
  at all, is deleted. The fact is not lost either: `internal/substrate/
  entry_points_rust.go` lists `actix_web::main` in `rustLifecycleAttrs`.
- `kmp.yaml` — keeps `Module <name>Main` from `val commonMain by getting`.
  **Measured limit, recorded not glossed:** all 10 deleted markers came from
  ktor/ktor-samples `build.gradle.kts` files spelling their source sets
  `jvmMain.dependencies { … }`, which the Module rule does not capture, so those six
  files lose their only source-set node. What is lost is a boolean that could not
  say *which* source sets exist; a rule for the property-access spelling is Tier D
  work, not a reason to keep a marker.
- `sinatra.yaml` — keeps Route, Controller, `configure :<env>` Config, `error <code>`
  Middleware, and the `helpers <Module>` include form. Ruby also has properly named
  producers for the deleted constructs, **on the gated lane only** — see the next
  section, which states that plainly rather than disclaiming it.
- `langchain.yaml` — keeps `.pipe(` Operation, `new DynamicTool(`/`new
  DynamicStructuredTool(` Operation, the prompt-template Schemas, and the Tier D
  `create\w+Agent(` Service.

### Stated plainly: the replacement producers are OFF by default

Every "there is a properly named producer instead" argument in this PR and in the new
rule comments points at the **custom-extractor lane, and `InProcCustomExtractorsEnabled`
(`internal/extractors/custom_gate.go`) says "The default stays OFF"**. So in a default
index none of these run:

- `internal/custom/ruby/middleware.go` — `SCOPE.Pattern "sinatra_filter:before[:/admin/*]"`
- `internal/custom/ruby/sinatra_deep.go` — `SCOPE.Pattern "sinatra_helpers_block"`
- `internal/custom/javascript/langchain.go:89` — `SCOPE.Operation "RunnableSequence"`
  with `chain_type`/`provenance` properties, for the exact `RunnableSequence.from([`
  construct site 42 deleted (found by the review; site 42 is the easiest of the nine
  to justify and this PR originally cited only `.pipe(`)

**In a default index these facts have no substitute at all — the entities are simply
gone, not produced better elsewhere.** The shadow-population framing is true only with
the gate ON, and the earlier wording read as "deleted here, produced properly over
there", which for a default index is false.

That does **not** change the decision, and the justification never depended on the
replacements: what is deleted is a marker named after the matched line carrying **zero
edges** (measured: zero relationship-endpoint deltas across 12,605 files). A default
index loses **noise, not information** — which is #6916's entire purpose — and that
holds whether or not any replacement producer runs. `internal/substrate/
entry_points_rust.go`'s `rustLifecycleAttrs` reading of `actix_web::main` is the one
cited mechanism that is *not* on the gated lane.

### One recall gap this PR creates, named rather than left to be rediscovered

After this change **`tool(...)` and `.bindTools(` have no producer anywhere** — not in
the YAML, not on the gated lane (`internal/custom/javascript/langchain.go`'s tool regex
requires `new`). And the spelling that survives, `new DynamicTool(`, is the
**deprecated** one, while `tool(...)` is the current `@langchain/core` idiom. So the
file now recognises the old spelling of a tool declaration and not the new one.

The fix is a real pattern capturing the tool's `name:` from the options object — Tier D
work, not a name group — so it is **filed against #6916's Tier D list** rather than
folded in here or left in a rule comment. Recorded here because it is obvious today and
invisible in three months.

### The sinatra golden fixture, re-baselined by hand

`internal/quality/golden/ruby-sinatra-modular-mini/expected.json` pinned three of
these markers as `must_exist` with the note *"pinned so a later cleanup is a
deliberate change rather than an accident."* This is that change:

- the three rows move from `expected_entities` to **`forbidden_entities`** — the only
  direction that can catch a deleted rule coming back — with the reasoning recorded
  in each row's note;
- must-have recall falls **13 → 10 by design**, `helpers AuthHelpers` staying a
  must-have as the held-constant control;
- `baseline.json` is edited **by hand, one fixture, four numbers**:
  `entity_found`/`entity_expected` 13 → 10, `entity_extracted_total` 61 → 58,
  `relationship_extracted_total` 118 → 115 (the three markers plus the three
  containment edges that carried them).

**Two-run diff evidence** (`--runs 1 --update-baseline` at HEAD *and* at the untouched
parent in a second worktree, both with a fresh isolation triple, then the two
generated baselines diffed against each other):

```
249,250c249,250      < entity_found: 13            > entity_found: 10
                     < entity_expected: 13         > entity_expected: 10
253,254c253,254      < entity_extracted_total: 61  > entity_extracted_total: 58
                     < relationship_extracted_total: 118  > ...: 115
```

One fixture, four numbers — that is what MY change did. The control matters: the
**parent-generated** baseline already differs from the committed one on **8 unrelated
fixtures** (extracted totals only, e.g. `entity_extracted_total` 57→49, 36→30, 23→19,
relationship 255→251 …). A blanket `--update-baseline` would have committed those
lowered ceilings as if they were this PR's. They are not, and they are not here.

`scripts/quality/run.sh --runs 1 --ratchet` at HEAD: **`quality ratchet OK — 38 gated
fixtures held their baseline (29 at 100% must-have recall)`**, with `git status
--porcelain` showing only this PR's own edits — no generated churn. The fixture's own
report reads `entity_found 10 / entity_expected 10 / entity_recall 1.0 /
forbidden_entity_hits 0`.

### Grading — deletions are ungradeable by a recall test

New `internal/engine/tier_c_marker_deletion_6916_test.go`: a table of **11 legs over
the 9 sites** (sinatra `before`/`after` get both the path-carrying and the **bare**
spelling), each asserting the marker's `<Kind>:<Name>` is absent from an input that
**provably produced it**, each with a positive control from the same file so an
unloadable rule file cannot pass it. Plus `TestIssue6916_TierCSurvivingSiblings` (the
neighbour axis) and `TestIssue6916_TierCLangchainMarkersWereNotSuppressedDownstream`
(the corrected claim, with `@tool` as its control).

The bare-spelling legs are there because of the trap recorded on the issue: `before`/
`after` already carry an optional group 1, `extractGroupFromIndex` returns `""` for an
unmatched optional group and the detector then skips the match — so `name_group: 1`
would have silently stopped extracting bare `before do` / `after do` while leaving the
path-carrying spelling alive. Deletion is the honest version of that, and both shapes
are now pinned.

**Mutant scores — 13 mutants, 13 DEAD.** Every edit was proven to land (exact
occurrence count asserted before running; a wrong count = SKIP, never a silent
ALIVE), and every mutant was reverted from a shasum-verified copy with byte equality
re-asserted.

| mutant | verdict |
|---|---|
| M1 restore `helpers do` | **DEAD twice** — `internal/engine` (only its own leg fails), **and the always-on quality gate**: `1 forbidden entity row(s) fired — always fatal`, naming `Service "helpers do" app.rb`, plus both extracted ceilings 58→59 / 115→116. Exit 2. |
| M2 restore sinatra `before` | DEAD — fails `:93` **and** `:93 (bare)` |
| M3 restore sinatra `after` | DEAD — fails `:99` **and** `:99 (bare)` |
| M4 restore `Sinatra::Application.run!` | DEAD — fails `:117` |
| M5 restore `#[actix_web::main]` | DEAD — fails `actix_web.yaml:90` |
| M6 restore `sourceSets {` | DEAD — fails `kmp.yaml:83` |
| M7 restore `RunnableSequence.from(` | DEAD — fails `langchain.yaml:42` |
| M8 restore `tool(async (` | DEAD — fails `langchain.yaml:66` |
| M9 restore `.bindTools(` | DEAD — fails `langchain.yaml:71` |
| M10 delete the KEPT `configure` rule | DEAD — sibling test fails (sinatra leg) |
| M11 delete the KEPT `App::new()` Tier D marker | DEAD — sibling test fails (actix leg) |
| M12 delete the KEPT kmp `Module` rule | DEAD — sibling test fails (kmp leg) |
| M13 delete the KEPT langchain `.pipe(` rule | DEAD — sibling test fails (langchain leg) |

Two independent confirmations landed after this table, both at the **gate** rather than
the package: restoring only the `after ... do` rule (coordinator) and restoring only the
`before ... do` rule (reviewer, single-fixture run with an unmutated positive control)
each fired `1 forbidden entity row(s) fired — always fatal`. So all three new forbidden
rows are proven reachable by an actual gate run rather than by inspection. Both also
recorded that **`go test ./internal/quality/...` stays GREEN under those mutants** — only
`scripts/quality/run.sh --ratchet` catches them, which is worth knowing for anyone
verifying a Tier C/D deletion by the package suite alone.

M2–M9 are single-site mutants, each failing **only its own leg**, so per-site
attribution is not inferred from a compound run. Independently, the whole test file
run at the untouched parent worktree (`1b115b73b`, all nine rules present) fails
**all 11 legs**, each naming its own marker — the deletion is pinned, not assumed.

### Verification

- `go test -count=1 ./internal/engine/` — ok (47.9s)
- `go test -count=1 ./internal/quality/...` — ok (all four packages)
- `go test -count=1 ./cmd/grafel/` with a **fresh** `GRAFEL_HOME` + `GRAFEL_DAEMON_ROOT`
  — **ok (153s), the semantic digest did NOT move.** Predicted and then confirmed: the
  #6118 fixture contains no actix/KMP/sinatra/langchain construct, so no fixture member
  produces a delta. Nothing was bumped.
- `gofmt -l internal/ cmd/` — clean
- Three pre-existing tests asserted the deleted behaviour and were **inverted, not
  deleted**: `TestDetect_LangChain_PromptsChainsTools`, and in
  `sinatra_multiline_anchor_6917_test.go` both the `helpers do` recall assertion (now
  a named absence plus an explicit `error 404` control, replacing a `len(Service) != 0`
  that would have passed with the wrong rule alive) and the heredoc over-fire row,
  which its own comment asked the next person to "move into the forbidden list" — done.

Refs #6916

### DIFF
--- a/internal/engine/langchain_detect_test.go
+++ b/internal/engine/langchain_detect_test.go
@@ -73,34 +73,38 @@ func TestDetect_LangChain_PromptsChainsTools(t *testing.T) {
 		t.Errorf("expected >=2 Schema (prompt) entities, got %d", c)
 	}
 
-	// chain_composition: .pipe() stages (captures upstream runnable name) +
-	// RunnableSequence.from.
+	// chain_composition: .pipe() stages (captures upstream runnable name).
 	if !names["prompt"] {
 		t.Errorf("expected .pipe() chain stage named 'prompt'; got %v", keys(names))
 	}
-	if !names["RunnableSequence.from("] {
-		t.Errorf("expected RunnableSequence.from composition entity; got %v", keys(names))
-	}
 
-	// tool_use_detection: the three tool declaration idioms + bindTools.
+	// tool_use_detection: the two `new`-constructor tool declaration idioms.
 	if !names["DynamicTool"] {
 		t.Errorf("expected DynamicTool entity; got %v", keys(names))
 	}
 	if !names["DynamicStructuredTool"] {
 		t.Errorf("expected DynamicStructuredTool entity; got %v", keys(names))
 	}
-	if !names[".bindTools("] {
-		t.Errorf("expected bindTools entity; got %v", keys(names))
-	}
-	// tool(async (...) factory.
-	var sawToolFactory bool
-	for n := range names {
-		if n == "tool(async (" || n == "tool((" {
-			sawToolFactory = true
+
+	// #6916 Tier C. Three `name_group: 0` rules in this file were DELETED, and
+	// this test used to require the entities they minted:
+	//
+	//	Operation "RunnableSequence.from("   RunnableSequence.from([...])
+	//	Operation "tool(async ("             the tool() factory
+	//	Operation ".bindTools("              model.bindTools([...])
+	//
+	// Each Name was the whole regex match — a call-site fragment, trailing paren
+	// included — so nothing could ever reference it. The assertions are inverted
+	// here rather than deleted: an absence with no assertion behind it is how a
+	// deleted rule comes back unnoticed, and this test drives the same source as
+	// the surviving assertions above, so it is the cheapest place to grade it.
+	// The full site table and the per-site controls live in
+	// tier_c_marker_deletion_6916_test.go.
+	for _, gone := range []string{"RunnableSequence.from(", ".bindTools(", "tool(async (", "tool(("} {
+		if names[gone] {
+			t.Errorf("#6916 Tier C: the deleted marker rule is back — Operation %q "+
+				"is in the graph again; got %v", gone, keys(names))
 		}
-	}
-	if !sawToolFactory {
-		t.Errorf("expected tool() factory entity; got %v", keys(names))
 	}
 }
 
--- a/internal/engine/precision_dedup.go
+++ b/internal/engine/precision_dedup.go
@@ -279,9 +279,18 @@ func symbolicID(kind, name string) string {
 //
 // This classifier is deliberately NARROW. The codebase legitimately uses
 // non-identifier Operation names for call-idiom detection — e.g.
-// "RunnableSequence.from(", ".bindTools(", "tool(async (", "createTRPCClient<".
-// Those must be KEPT. So we drop ONLY three concrete statement shapes that carry
-// no architectural signal:
+// "createTRPCClient<" (javascript_typescript/frameworks/trpc.yaml). Those must
+// be KEPT. So we drop ONLY three concrete statement shapes that carry no
+// architectural signal:
+//
+// #6916 Tier C note: three of the four examples this comment used to list —
+// "RunnableSequence.from(", ".bindTools(", "tool(async (" — no longer exist.
+// They were `name_group: 0` marker rules in
+// javascript_typescript/frameworks/langchain.yaml and were DELETED at the rule
+// level rather than filtered here. Their presence in this list was also the
+// evidence that #6916's "already dropped downstream as Operation noise" claim
+// about them was wrong: this pass never dropped them, and the langchain site
+// it WAS written for is the python `@tool` decorator, case (1) below.
 //
 //  1. Bare decorator text — name starts with `@` and the remainder is a plain
 //     identifier with no call/args (e.g. "@tool", "@property"). A decorator
--- a/internal/engine/rules/javascript_typescript/frameworks/langchain.yaml
+++ b/internal/engine/rules/javascript_typescript/frameworks/langchain.yaml
@@ -38,11 +38,10 @@ source_patterns:
     entity_type: Operation
     name_group: 1
     scope: file
-  # RunnableSequence.from([...]) — the explicit composition constructor.
-  - pattern: "RunnableSequence\\.from\\s*\\("
-    entity_type: Operation
-    name_group: 0
-    scope: file
+  # #6916 Tier C: `RunnableSequence.from(` used to mint an Operation entity
+  # literally NAMED "RunnableSequence.from(" via `name_group: 0`, trailing
+  # paren included. The `.pipe()` rule above is the bindable form of the same
+  # LCEL composition fact (it captures the upstream runnable variable).
   # --- prompt_template_extraction (#2865) ---------------------------------
   # ChatPromptTemplate.fromMessages(...) / .fromTemplate(...) and the bare
   # PromptTemplate.fromTemplate(...) factory. Captures the template-class +
@@ -63,15 +62,15 @@ source_patterns:
     entity_type: Operation
     name_group: 1
     scope: file
-  - pattern: "\\btool\\s*\\(\\s*(?:async\\s*)?\\("
-    entity_type: Operation
-    name_group: 0
-    scope: file
-  # .bindTools([...]) — attaches a tool set to a model (function-calling).
-  - pattern: "\\.bindTools\\s*\\("
-    entity_type: Operation
-    name_group: 0
-    scope: file
+  # #6916 Tier C: the `tool(async (` and `.bindTools(` rules used to mint
+  # Operation entities literally NAMED "tool(async (" and ".bindTools(" via
+  # `name_group: 0` — call-site fragments with no referent. Both survive
+  # dropStatementNoiseOperations (measured: its noise predicate keys on a
+  # leading "@", a bare "=", or a leading statement keyword, none of which a
+  # trailing-paren fragment has), so this IS a deliberate removal of entities
+  # that reach the graph, not the deletion of an already-suppressed rule. The
+  # `new DynamicTool(...)` / `new DynamicStructuredTool(...)` rule above keeps
+  # the named form of the tool-declaration fact.
   # --- agents -------------------------------------------------------------
   # Agent factory (createReactAgent, createToolCallingAgent, …).
   - pattern: "create\\w+Agent\\s*\\("
--- a/internal/engine/rules/kotlin/frameworks/kmp.yaml
+++ b/internal/engine/rules/kotlin/frameworks/kmp.yaml
@@ -79,11 +79,21 @@ source_patterns:
     name_group: 1
     scope: class
 
-  # kotlin { sourceSets { ... } } — source set configuration block
-  - pattern: "\\bsourceSets\\s*\\{"
-    entity_type: Config
-    name_group: 0
-    scope: file
+  # #6916 Tier C: `sourceSets {` used to mint a Config entity literally NAMED
+  # "sourceSets {" via `name_group: 0` — a Gradle DSL block opener, not a thing
+  # in the repo. 10 entities on the 60-repo corpus, the 7th-biggest name_group:0
+  # producer, all unbound (no relationship_rule in the kotlin bucket names
+  # Config as an endpoint). The named, bindable form of the same information is
+  # the `val <name>Main by getting` rule below (5 Module entities on the same
+  # corpus: backendMain, frontendMain, jsMain).
+  #
+  # Measured limit, recorded rather than glossed: all 10 deleted markers came
+  # from ktor/ktor-samples build.gradle.kts files that spell their source sets
+  # `jvmMain.dependencies { … }`, which the Module rule below does NOT capture —
+  # so those six files are left with no source-set node at all. What is lost
+  # there is a per-file boolean that could not name WHICH source sets exist; a
+  # rule for the property-access spelling is the fix if that fact is wanted, and
+  # it belongs with the other Tier-D captures, not here.
 
   # val commonMain by getting / val androidMain by getting — source set variable
   - pattern: "val\\s+(\\w+Main)\\s+by\\s+(?:getting|creating)"
--- a/internal/engine/rules/ruby/frameworks/sinatra.yaml
+++ b/internal/engine/rules/ruby/frameworks/sinatra.yaml
@@ -69,11 +69,19 @@ source_patterns:
     name_group: 1
     scope: class
 
-  # Helper block: helpers do ... end
-  - pattern: "(?m)^\\s*helpers\\s+do"
-    entity_type: Service
-    name_group: 0
-    scope: file
+  # #6916 Tier C: `helpers do` used to mint a Service entity literally NAMED
+  # "helpers do" via `name_group: 0` — the matched line, not a helper. Nothing
+  # could bind to it: the only rule targeting Service (Route CALLS Service,
+  # below) takes its target from `(\\w+_helper|authenticate!|authorize!)`, a
+  # method name, which can never equal "helpers do". The helper METHODS inside
+  # the block are already extracted by the Ruby extractor as SCOPE.Operation
+  # (that is what the Route CALLS edge binds to), and the `helpers <Module>`
+  # include form below keeps its real, bindable name. Ruby also has a properly
+  # named producer for the block itself — internal/custom/ruby/sinatra_deep.go
+  # emits SCOPE.Pattern "sinatra_helpers_block" — so this rule was a parallel
+  # shadow population beside it (whether that extractor runs on a given index is
+  # gated separately by GRAFEL_INPROC_CUSTOM_EXTRACTORS; no claim is made here
+  # about which indexes carry it).
 
   # Helper include: helpers HelperModule
   #
@@ -89,17 +97,21 @@ source_patterns:
     name_group: 1
     scope: file
 
-  # Before filter: before "/path" do or before do
-  - pattern: "(?m)^\\s*before\\s+(?:[\"']([^\"']+)[\"']\\s+)?do"
-    entity_type: Middleware
-    name_group: 0
-    scope: file
-
-  # After filter: after "/path" do or after do
-  - pattern: "(?m)^\\s*after\\s+(?:[\"']([^\"']+)[\"']\\s+)?do"
-    entity_type: Middleware
-    name_group: 0
-    scope: file
+  # #6916 Tier C: the `before`/`after` filter rules used to mint Middleware
+  # entities literally NAMED "before '/admin/*' do" / "after do" via
+  # `name_group: 0` — the whole matched line. Deleted rather than renamed to
+  # the optional path group: `extractGroupFromIndex` returns "" for an
+  # unmatched optional group and the detector then skips the match, so
+  # `name_group: 1` would silently stop extracting bare `before do` / `after
+  # do` altogether — a recall change wearing a rename's clothes. Nothing bound
+  # to either name: the only rule targeting Middleware (Controller
+  # REGISTERED_ON Middleware, below) takes its target from `use
+  # Rack::Session::Cookie`, a module constant. `error <code> do` below is the
+  # one filter-family rule that already carries a real name and it stays. As
+  # with `helpers do` above, Ruby has a properly named producer for the same
+  # construct — internal/custom/ruby/middleware.go emits SCOPE.Pattern
+  # "sinatra_filter:before" / "sinatra_filter:before:/admin/*" — so what is
+  # deleted here is the shadow, not the fact.
 
   # Configure block: configure :production do
   - pattern: "(?m)^\\s*configure\\s*(?::(\\w+))?\\s*do"
@@ -113,11 +125,13 @@ source_patterns:
     name_group: 1
     scope: file
 
-  # Sinatra::Application classic-style app boot
-  - pattern: "Sinatra::Application\\.run!"
-    entity_type: Config
-    name_group: 0
-    scope: file
+  # #6916 Tier C: `Sinatra::Application.run!` used to mint a Config entity
+  # literally NAMED "Sinatra::Application.run!" via `name_group: 0` — a boot
+  # call site, not a configuration object. It made the sinatra `Config
+  # ROUTES_TO Controller` rule dangling BY CONSTRUCTION: that rule sources on
+  # `map "/api" do`, i.e. a mount PATH, so the only Config entity the framework
+  # minted could never be the edge's endpoint. The `configure :<env> do` rule
+  # above is the real, named Config producer and is kept.
 
 # relationship_rules: regex patterns that produce edges between entities.
 relationship_rules:
--- a/internal/engine/rules/rust/frameworks/actix_web.yaml
+++ b/internal/engine/rules/rust/frameworks/actix_web.yaml
@@ -86,11 +86,19 @@ source_patterns:
     name_group: 0
     scope: file
 
-  # #[actix_web::main] entry point
-  - pattern: "#\\[actix_web\\s*::\\s*main\\s*\\]"
-    entity_type: Config
-    name_group: 0
-    scope: file
+  # #6916 Tier C: the `#[actix_web::main]` attribute-macro marker used to mint a
+  # Config entity literally NAMED "#[actix_web::main]" via `name_group: 0` — a
+  # per-file boolean ("this file has the actix main macro") with no referent for
+  # any edge to point at, and no relationship_rule in the rust bucket names
+  # Config as an endpoint, so it carried zero edges. It was also the THIRD
+  # marker on the same concept: `HttpServer::new(` and `App::new()` below fire
+  # on the same main.rs and mean the same thing ("server entry point"). Those
+  # two are kept (Tier D still owes them a real name); this one is deleted
+  # rather than renamed because a bare attribute macro has no name to capture.
+  # The attribute is not lost as a FACT either: internal/substrate/
+  # entry_points_rust.go lists `actix_web::main` in rustLifecycleAttrs and reads
+  # it for entry-point classification, which is the modelled form of exactly
+  # what this marker gestured at.
 
   # App::new() — application builder (middleware config point)
   - pattern: "App\\s*::\\s*new\\s*\\(\\s*\\)"
--- a/internal/engine/sinatra_multiline_anchor_6917_test.go
+++ b/internal/engine/sinatra_multiline_anchor_6917_test.go
@@ -157,8 +157,40 @@ func TestIssue6917_ModularSinatraAppYieldsRoutes(t *testing.T) {
 		t.Errorf("configure :production not extracted as Config; Configs: %v",
 			entity6917Names(res, "Config"))
 	}
-	if len(entity6917Names(res, "Service")) == 0 {
-		t.Errorf("expected the `helpers do` block as a Service, got none")
+	// #6916 Tier C deleted the `helpers do` BLOCK rule (a Service named after the
+	// whole matched line), so the Service entities this fixture yields now come
+	// from the `helpers <Module>` INCLUDE rule alone. Naming that expectation is
+	// what keeps the assertion honest: `len(Service) != 0` would have gone on
+	// passing if the include rule had died and the block rule survived.
+	// This fixture carries only the BLOCK form (`helpers do`), not the include
+	// form (`helpers AuthHelpers`), so after the deletion it yields no Service at
+	// all. Both directions are asserted: the name that must not come back, and
+	// the set that must stay empty. The positive controls for "this file is still
+	// being extracted" are the Route / Controller / Config / Middleware
+	// assertions above; the include form's own recall is graded by
+	// internal/quality/golden/ruby-sinatra-modular-mini (Service AuthHelpers,
+	// must_exist) and by tier_c_marker_deletion_6916_test.go.
+	if got := entity6917Names(res, "Service"); len(got) != 0 {
+		t.Errorf("expected no Service entities from a file whose only helpers construct "+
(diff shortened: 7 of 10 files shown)


cajasmota:
## Coordinator mutant on `6d2f8f78b` — **CX-1 DEAD, and it grades the claim the whole re-baseline rests on**

Own detached worktree at `/private/tmp/.../coord-mut-7034-v8n/wt` (absolute), private script dir. Restored by `cp` from a SHA-256-verified backup (`ddeabe24…` both ways); `git diff HEAD` and `git status --porcelain` empty afterwards.

Your thirteen cover per-site attribution (M2–M9, each failing only its own leg) and the kept siblings (M10–M13). What none of them tests is the load-bearing sentence of the re-baseline itself, which you wrote into the fixture:

> *"src/app.rb still contains the bare `after do` filter, so this row is reachable: restoring the deleted rule turns the fixture red."*

That is a claim about the **forbidden row**, not about the rule. And a forbidden row that is *unreachable* is a no-op that reads exactly like protection — which matters more here than usual, because moving three `must_exist` rows into `forbidden_entities` is the entire mechanism by which this deletion stays deleted.

### CX-1 — restore the deleted `after ... do` rule, nothing else

Reinstated the exact deleted four lines into `sinatra.yaml` with a `CV-MUT` marker, occurrence asserted 1, and confirmed the file still parses as YAML before running (a rule file that fails to load is silently *skipped* by the loader, which would have produced a green run for the wrong reason).

**First result — and this is the part worth reading:**

```
go test -count=1 ./internal/quality/...   →  exit 0, GREEN
```

The Go package suite does not see it. Then:

```
scripts/quality/run.sh --runs 1 --ratchet →  exit 2
==> quality ratchet FAILED
  - ruby-sinatra-modular-mini: 1 forbidden entity row(s) fired — always fatal
```

**DEAD.** The row is genuinely reachable, your sentence is true, and the failure is **"always fatal"** rather than a ceiling comparison — which is the right severity, since `--ratchet` does not fail on a ceiling drop and a growth-gated ceiling could never catch a rule coming back.

I am recording the green package run alongside the red ratchet deliberately: *"`go test ./internal/quality/...` is green"* and *"the extraction-quality gate is green"* are different statements, and anyone verifying a Tier C or Tier D deletion by the former will conclude the forbidden rows work when they have not been exercised at all.

### The layering, complete

- **Each deleted site is attributed** → M2–M9, single-leg failures.
- **The kept siblings still fire** → M10–M13.
- **The deletion cannot be silently reverted** → CX-1, via the always-fatal forbidden row.
- **The whole test file at the parent fails all 11 legs** → your own control, which is what rules out the tests passing vacuously on either side.

### Three things I want on the record

**You verified the issue's premise instead of implementing it.** #6916 said the three JS langchain sites were "already dropped downstream as Operation noise, so the rules are dead weight". You found that false — they reach the graph, and `precision_dedux`'s own comment names those three as ones that **"must be KEPT"**. So three of the nine are a deliberate recall reduction against another component's stated intention. Labelling that rather than quietly shipping it is the correct call, and the independent review is adjudicating whether the comment is stale or whether this over-reaches by three sites. **I am not merging until it does** — that is the one open question on this PR.

**The two-run baseline diff caught a standing hazard in the act.** The parent-generated baseline already differs from the committed one on **8 unrelated fixtures**, which is exactly what a blanket `--update-baseline` would have committed: silently lower ceilings on eight fixtures nobody touched. Hand-editing one fixture's four numbers and proving it against a two-run diff is the only method that separates "my change did this" from "this machine does this", and it paid for itself here.

**Seven of the nine minted zero on this corpus, and you said so as a corpus-relative zero rather than a capability claim** — then proved each live on a fixture instead. "Produced nothing here" never means "cannot produce", and that distinction has cost this board real time when it was skipped.

### One thing to carry, not to fix here

The kmp limit you recorded rather than glossed: all 10 markers came from files spelling source sets `jvmMain.dependencies {}`, which the surviving Module rule does not capture, so six files lose their only source-set node. You called it Tier D. Recording it in the PR rather than in a commit message is what makes it findable when Tier D is picked up.

**Merging on green, once the review returns on the langchain three.**

cajasmota:
## APPROVE

Independent adversarial review. Own worktree at `6d2f8f78b`, parent `1b115b73b`, fresh isolation triple on every gate run, `git status --porcelain` empty at the end of every mutant. I did not write this change.

Everything I could check independently held. Four mutants of my own, all DEAD except one that is ALIVE in a direction the ratchet documents as allowed — and that one turns out to be the same hole that produced the 8-fixture drift, which is the most useful thing in this review.

---

### 1. The langchain question — the deletion is RIGHT, and the "must be KEPT" comment was never an endorsement

**The author's finding is confirmed, exactly as reported.** `isStatementNoiseOperationName` (`precision_dedup.go`) drops a name only when it (1) starts with `@` and the remainder is bare identifier chars, (2) contains a `=` that is not part of `==`/`!=`/`<=`/`>=`, or (3) opens with one of twelve statement keywords plus a space. `RunnableSequence.from(`, `tool(async (` and `.bindTools(` satisfy none of the three. And the pre-edit comment did list those three names verbatim under "Those must be KEPT". Read, not relayed.

**Adjudication: the comment is right about the filter and says nothing about the rules, so there is no real tension and the PR is not over-reaching.**

That sentence is a *blast-radius* instruction for a blunt instrument. `dropStatementNoiseOperations` classifies by name shape alone, with no idea which rule minted the name; widening it to catch trailing-paren call fragments would also take out `createTRPCClient<` (`trpc.yaml`) and every other legitimate call-idiom Operation in the tree. "Must be KEPT" means *this pass must not be broad enough to remove them* — it is a constraint on the filter, not a claim that the three entities carry value. Deleting the rule is precisely the surgical alternative the comment could not perform. Keeping a rule alive because a filter was told not to catch it inverts the dependency.

The deciding question you named — what does a consumer do with a langchain `Operation` today — settles it the same way. `Operation` is the kind real functions and methods land in, and `langchain.yaml` has `relationship_rules: []`, so these three nodes were unbindable strings sitting in the middle of the population a consumer traverses for actual callables. A node named `.bindTools(` is not a smaller version of a useful node; it is a different kind of thing wearing the same kind label. Rendering it is a cost, not a benefit.

**Two facts that strengthen the case beyond what the PR body claims**, and one that weakens it:

- `internal/custom/javascript/langchain.go:89-93` already emits `SCOPE.Operation "RunnableSequence"` with `chain_type`/`provenance` properties for the very same construct (`RunnableSequence\s*\.\s*from\s*\(\s*\[`). The PR cites only `.pipe(` as the surviving form; the properly-named producer for this exact site exists and is not mentioned. Site 42 is the easiest of the nine to justify.
- `precision_dedup_test.go:205-206` still carries `{"RunnableSequence.from(", false}` and `{".bindTools(", false}`, so the filter's *narrowness* stays pinned even though no rule mints those names now. The comment edit does not un-grade the thing the comment was protecting.
- **The weakening fact, stated plainly because the PR only disclaims it:** `InProcCustomExtractorsEnabled` in `internal/extractors/custom_gate.go` says "The default stays OFF". So every "properly named producer" cited in the rule comments — `internal/custom/ruby/middleware.go`, `sinatra_deep.go`, and the JS `langchain.go` I found above — does **not** run in a default index. In the default path these facts now have no substitute, not a better-named one. That does not change my verdict, because an unnameable unbindable node is not a substitute either, but it does mean the shadow-population framing is only true with the gate on.

**One genuine recall gap nobody has named**, and the only thing I'd ask be filed: `.bindTools(` and `tool(async (` have **no producer anywhere now** — not in the YAML, not in `internal/custom/javascript/langchain.go` (whose `reLangchainTool` requires `new`), not on the gated path. And `tool(...)` is the *current* idiom in `@langchain/core`; `new DynamicTool(` — the form the PR keeps — is the deprecated one. So the file now recognises the old spelling of tool declaration and not the new one. The right fix is a real pattern that captures the tool's `name:` from the options object, i.e. Tier D, not a name group. Please file it against #6916's Tier D list rather than leaving it in a rule comment.

---

### 2. The sinatra baseline hand-edit — both halves verified independently

Regenerated `baseline.json` at HEAD (`--runs 1 --update-baseline`, fresh isolation, shared Go caches only) and diffed the result against the committed file.

**Half one — the hand-edit is exactly what a correct regenerate produces.** `ruby-sinatra-modular-mini` **does not appear in the diff at all**. Its own report reads `entity_found 10 / entity_expected 10 / entity_recall 1.0 / forbidden_entity_hits 0 / entity_extracted_total 58 / relationship_extracted_total 115` — all four hand-edited numbers correct.

**Half two — the 8-fixture drift is real, and I can now name the fixtures.** Regenerated vs committed, extracted totals only, recall figures untouched on every one:

| fixture | entity_extracted | relationship_extracted |
|---|---|---|
| csharp-aspnet-core-mini | 57 → 49 | 110 → 94 |
| csharp-hangfire-mini | 36 → 30 | 66 → 60 |
| csharp-quartz-net-mini | 23 → 19 | 45 → 41 |
| razor-blazor-mini | 25 → 22 | 35 → 32 |
| python-django-mini | 98 → 98 | 255 → 251 |
| python-fastapi-mini | 28 → 27 | 57 → 54 |
| python-rq-mini | 22 → 20 | 43 → 39 |
| docker-compose-services-mini | 48 → 47 | 68 → 65 |

**Not one of the eight is ruby, rust, kotlin or a langchain-JS fixture**, so the drift cannot be this PR's — the attribution is now positive, not inferred from a two-run diff. The author's decision not to blanket-regenerate was correct, and a blanket regenerate would have buried eight unexplained ceilings under a Tier C label. (The committed header still reads `measured_on: fc9124eef / 2026-09-06`; my regenerate stamps the base ref, so the drift accumulated since that sha. `18748824b` — the `noiseLocalScope` tier fix, an intended entity reduction — is the obvious suspect. Worth filing so the eight get re-recorded deliberately.)

`--ratchet` at HEAD with the committed baseline: `quality ratchet OK — 38 gated fixtures held their baseline (29 at 100% must-have recall)`, exit 0, `git status --porcelain` empty.

**My mutants on the four hand-edited numbers** (`ratchet.py check` against the fresh stamped reports; each edit asserted to land before scoring):

| mutant | verdict |
|---|---|
| RM-recall: `entity_found`/`entity_expected` back to 13 | **DEAD** — exit 2, `entity_found REGRESSED 13 -> 10` plus `entity_expected changed … re-record with --update-baseline` |
| RM-tight: `entity_extracted_total` 58 → 55 | **DEAD** — exit 2, `entity_extracted_total GREW 55 -> 58` |
| RM-loose: ceilings left at the pre-PR 61 / 118 | **ALIVE** — exit 0, `38 gated fixtures held their baseline` |
| RM0 control (committed baseline) | exit 0 |

So two of the four hand-edited numbers are graded in both directions and two are graded in one. RM-loose is **distinguishable and ungraded, in the loose direction only**: `observed()` records extracted totals "in the growth direction only" by design, and its docstring argues for that. Price: zero here — the author picked the tight number and my regenerate agrees — but it is worth naming that RM-loose and the 8-fixture drift are *the same hole*. A gate that cannot see a fall is why eight ceilings went stale for two days without a word. I am not asking for that to change in this PR.

---

### 3. Seven zero-mint sites — the framing holds

"Corpus-relative zero, not a capability claim" is the correct reading and the PR does not lean on the zeros for anything. Liveness is established the only way it can be: `tier_c_marker_deletion_6916_test.go` carries an input per site through the real `LoadAllRules` + `Detect` (post-`applyPrecisionDedup`), and the whole file fails all 11 legs at the untouched parent. I re-derived the inputs myself — `#[actix_web::main]` in a `main.rs`; `sourceSets {` in a `build.gradle.kts`; the four sinatra constructs, indented, in a `Sinatra::Base` subclass, plus the two **bare** spellings; the three JS call fragments in a `.ts` file. Every one is idiomatic, not contrived, and the sinatra four would fire four times per file in any real Sinatra app. So the honest statement is not "these never fire" but "these fire a lot in repos the corpus does not contain, and what they mint there is still an unbindable per-file boolean". The golden fixture's deliberate 13 → 10 is that admission made in the graded direction. Deleting them stays right.

The unbound claim I re-derived statically rather than trusting: `grep` for `from_type:`/`to_type:` in the four language buckets returns nothing, and reading all three relationship rules in `sinatra.yaml` confirms the endpoints source on a mount **path**, a `use <Constant>` **module**, and a `\w+_helper` **method** — none can equal a marker Name. `actix_web.yaml` and `kmp.yaml` name no `Config` endpoint at all; `langchain.yaml` is `relationship_rules: []`.

---

### 4. The kmp limit — acceptable, with a filing ask

Acceptable. `val\s+(\w+Main)\s+by\s+(?:getting|creating)` genuinely cannot see `jvmMain.dependencies { … }`, so the six ktor-samples files do lose their only source-set node. But what they lose is a node named `sourceSets {` that no edge could reach — a file with no entity and a file with one unnameable entity are the same file to a consumer, and the second is worse in the population it pollutes (`Config`, measured 60.5% polluted on #7013). Blocking a deletion on the absence of a capture that never existed would make Tier C unshippable by construction.

The ask is procedural: the property-access spelling is currently recorded **only in a YAML comment**. Add it to #6916's Tier D list on the issue, where the tier is tracked, so it survives the next person's `git log`.

---

### 5. The inverted tests — read each one, and scored two of them

All three inversions assert something specific, not merely an absence:

- `TestDetect_LangChain_PromptsChainsTools` keeps `.pipe()`→`prompt`, `DynamicTool`, `DynamicStructuredTool` and `>=2 Schema` as live controls, and forbids all four deleted names including the `tool((` variant the old loop accepted. Nothing became a no-op.
- `TestIssue6917_ModularSinatraAppYieldsRoutes` replaces `len(Service) != 0` with an exact-empty assertion **plus** a named absence **plus** a new `error 404 do` Middleware control. Your instinct was right that `len(Service) != 0` was the dangerous one — it would have passed with the include rule dead and the block rule alive. The replacement is strictly stronger. Its one cost: exact-empty means any future legitimate ruby `Service` producer turns this red. That is the correct failure (someone reads the comment) rather than a silent pass, so I would leave it.
- `TestIssue6917_MultilinePatternsDoNotOverFire` moves `{"Service","helpers do"}` from the known-and-reachable list to `forbidden`, which is what its own comment asked the next person to do.

Scored two inversions with my own mutants (occurrence count asserted before each run; restored from a shasum-verified copy with `git status --porcelain` empty after):

| mutant | verdict |
|---|---|
| IM1 — re-add the `helpers do` Service rule to `sinatra.yaml` | **DEAD three ways**: `TestIssue6917_ModularSinatraAppYieldsRoutes` FAIL, `TestIssue6917_MultilinePatternsDoNotOverFire` FAIL, and **only** the `sinatra.yaml:73` subtest of the Tier C table |
| IM2 — re-add the `.bindTools(` rule to `langchain.yaml` | **DEAD**: `TestDetect_LangChain_PromptsChainsTools` FAIL and **only** the `langchain.yaml:71` subtest |

Single-site attribution reproduced independently, not taken from the PR table.

---

### 6. The mutant nobody ran — and a null result the PR is right to call null

**IM3, the one I most wanted: is the `before '/admin/*' do` forbidden row actually reachable?** The author demonstrated the gate firing for `helpers do` (M1) but only the engine legs for M2/M3, so two of the three new forbidden rows rested on "the source still contains the construct" — and a row whose name must match exactly, with the quotes, is precisely where a fixture claims more than its content. Restored *only* the `before` rule, rebuilt the binary, and ran the single fixture:

```
forbidden_entity_hits 1
  {"name":"before '/admin/*' do","kind":"Middleware","got_name":"before '/admin/*' do",
   "got_kind":"Middleware","got_source_file":"app.rb"}
entity_extracted_total 58 -> 59      exit 2
```
with the unmutated binary on the same fixture as the positive control: `forbidden_entity_hits 0`, `58`, exit 0. **DEAD.** Two of the three rows are now proven reachable by an actual gate run rather than by inspection; the third (`after do`, bare, line 46 of `src/app.rb`) is the same rule shape through the same unnarrowed `(kind, name)` matcher.

**On "did a deletion take a neighbour with it":** the strongest available evidence is not a mutant. Diffing each of the four rule files parent→HEAD **with comments and blank lines stripped** yields nothing but the 36 deleted lines — nine `- pattern:` / `entity_type:` / `name_group:` / `scope:` blocks. Zero surviving lines modified, zero added. Combined with `TestIssue6916_TierCSurvivingSiblings` (which pins all five sinatra survivors, both remaining actix `Config` markers, the kmp `Module`, and four langchain survivors) and a green 38-fixture ratchet, there is no unpinned neighbour a block deletion could plausibly have reached.

**The `cmd/grafel` digest could not have graded this, and the PR says so — correctly.** I checked rather than accepted the prediction: `entityTupleKey` hashes `Kind|Name|QualifiedName|Subtype|SourceFile|Language|StartLine|EndLine`, so a *deleted entity absolutely would* move it — but the #6118 fixture's rust file is **axum** (`Router::new().route(...)`, no `#[actix_web::main]`), its TS/JS files are **express**, and it contains no ruby and no kotlin at all. An unmoved digest here is a null result, and the grading came entirely from the engine table plus the golden fixture. Worth stating so nobody later cites the digest as evidence.

---

### Verification I ran

- `go vet ./internal/engine/ ./internal/quality/...` — clean
- `go test -count=1 ./internal/engine/` — ok 51.4s
- `go test -count=1 ./internal/quality/...` — ok (all four packages)
- `gofmt -l internal/ cmd/` — clean
- `scripts/quality/run.sh --runs 1 --ratchet` at HEAD — exit 0, 38 held, `git status --porcelain` empty
- `scripts/quality/run.sh --runs 1 --update-baseline` at HEAD — diffed against the committed baseline (results above)
- 4 mutants (RM-recall, RM-tight, RM-loose, IM1, IM2, IM3 — 6 in total), each with a landing assertion and a shasum-verified restore

### Follow-ups I'd file, none blocking

1. **`tool()` / `.bindTools(` have no producer at all now** — and `tool()` is the current spelling while the kept `new DynamicTool(` is the deprecated one. Tier D pattern work.
2. **The 8 drifted baselines** (4 C#/razor, 3 python, docker-compose) — unrelated to this PR, confirmed twice, invisible because the ratchet allows a fall (proven by RM-loose).
3. **kmp `jvmMain.dependencies {`** — promote from a YAML comment to #6916's Tier D list.
4. Drive-by, cosmetic: `scripts/quality/run.sh:161` blanks `_iso_home` whenever `GRAFEL_HOME == $HOME/.grafel`, so the **exact incantation the warning itself recommends** (`export HOME=$(mktemp -d); export GRAFEL_HOME=$HOME/.grafel; export GRAFEL_DAEMON_ROOT=$HOME/.grafel`) prints `WARNING: PARTIALLY ISOLATED environment` on a fully isolated run. Advisory-only, but it cries wolf on the one setup every agent is told to use.

Safe to merge.

PR #7036, opened 2026-09-08
feat(#6912): go fields carried their declared type only at the struct anchor — emit the field→type REFERENCES edge, same-file targets only

cajasmota:
Arm D of #6912. Arms A (C#, #6984), B (proto, #6991) and C (rust, #7000) shipped `REFERENCES` with `ref_kind=field_target_type`, same-file targets only. This is the Go arm, and it is the root cause of contributor issue #6726 (`Schema` 99.3% orphan of 35,353 entities).

`Refs #6912` — one arm of several, so not a closure keyword.

## Why Go is arm B's shape, not arm C's

Go is the only language on this issue that **already** shipped a same-file-gated declared-type edge: `extractStructFieldDependencies` emits `DEPENDS_ON` per struct field type, gated on `knownTypeNames`. But it anchors on the **struct**, leaving the field a leaf on its outbound side — and `DEPENDS_ON` names no field at all, not even in a property. So "which fields point at `Customer`?" was unanswerable.

This adds a **second** edge anchored on the field and leaves the existing one untouched. The duplicated thing is the anchor, not the information.

## The two decisions the brief asked me to make deliberately

**1. Address — `BuildComponentStructuralRef`, not the neighbour's bare name.** The adjacent `DEPENDS_ON` uses `ToID: t` (bare). #6986 measured that form at 91.6% dangling for the custom lane. The new edge uses `scope:component:class:go:<file>:<Name>`, which is file-scoped. Adjacency was not treated as a reason to inherit.

**2. Allow-list — I admit `SCOPE.Schema`, and it is load-bearing.** Read from Go's own `extractTypes` rather than inherited:

| Kind | Subtype | Admitted | Why |
|---|---|---|---|
| `SCOPE.Component` | `struct` | yes | `type T struct{…}` |
| `SCOPE.Component` | `interface` | yes | `type T interface{…}` |
| `SCOPE.Schema` | `type_alias` | **yes** | `type Celsius float64`, `type Handler func()`, `type T = U` |
| everything else | | no | |

Arm C's guard is `Kind != "SCOPE.Component"`, which is right for Rust and **wrong here** — in Go a type alias is a `SCOPE.Schema`. Inheriting it would have silently dropped **32 of the 168 corpus edges (19%)**. Measured, not argued: mutant **M4** applies exactly that guard and is DEAD.

## A Go-specific hazard arm C's rule would not have caught

`type Status int` beside a typed const block produces a `SCOPE.Enum` value-set named `Status` **in the same file**, alongside the `SCOPE.Schema/type_alias`. Two kinds at one (file, name) is two graph nodes; the resolver's `ambigLocation` check runs *before* the `byLocation` fallback this address depends on, so the edge would dangle. Arm C's collision check looked only **within its own admitted set** — the alias is admitted, the value-set is not — so it would not have seen this.

The rule here spans **every** same-file record, and counts **kinds, not records**, mirroring `internal/resolve/refs.go:1308` (`existing != e.ID`) exactly. Counting records instead would delete a legitimate edge: `import _ "embed"` beside `type embed struct{…}` is two records but one node. Both directions graded (M5, M6 both DEAD).

## Candidate collector is not `resolveTypeReferences`

The brief pointed at `resolveTypeReferences` as the existing unwrapper. I did **not** reuse it: it collects every descendant `type_identifier`, which for a `qualified_type` means it **strips `other.Order` to `Order`** — binding to a same-file `Order` silently, since a bound edge never reaches `bug-extractor`. Arms B and C both forbid exactly that. `goFieldTypeCandidates` skips `qualified_type` whole, and also skips anonymous `struct_type`/`interface_type` (an anonymous struct's member is not the field's declared type).

The pre-existing `DEPENDS_ON` keeps its strip, **pinned as pre-existing** in `TestGoFieldTypeRefs_QualifiedTypeIsNotStrippedToASameFileName`. Narrowing it is a separate change with its own blast radius, not a rider.

No primitive blocklist, and that is deliberate: Go's predeclared identifiers are `type_identifier` in the grammar and are legally shadowable, so `type rune struct{…}` means a field written `rune` genuinely names that type. Arm B's `protoScalars` guard is correct there and would be **wrong** here — both directions asserted.

## Corpus yield — `archigraph-corpora`, read-only, sequential

| | |
|---|---|
| Go files scanned | 686 |
| Files emitting edges | 73 |
| `field_target_type` edges | **168** |
| **BOUND** | **168** |
| **Dangling** | **0** (100% bind) |

Target kinds: `Component/struct` 107, **`Schema/type_alias` 32**, `Component/interface` 29. Top repos: etcd 87, grpc-go-examples 44, gin 25, chi 12.

Per-file resolution is faithful here because `byLocation`/`ambigLocation` are keyed by file path and the pass is same-file only.

## The double-count question (arm B's open question — answered)

**Measured by reading the arithmetic, not reasoned.** No metric double-counts, because **no metric in `internal/quality` counts endpoints**, and there is **no edge dedup anywhere** in the counting path. But two rates move, both flattering:

- **`audit.go:424` OrphanRate is DEFLATED** — `touched` (`audit.go:363-368`) admits any non-`CONTAINS`/`DECLARES` edge in either direction, so a field flips to non-orphan on being a `FromID`. That is the #6912 gap closing and is intended. Worth naming: `fitness/rule.go:511-526` computes a *second*, inbound-only orphan rate that will **not** move, so the two published orphan numbers diverge further after this arm.
- **`audit.go:433` ReferencesPerFunction is INFLATED — a genuine distortion.** Numerator counts REFERENCES **edges** (`audit.go:382-384`), denominator counts function **entities**; a `SCOPE.Schema/field` is not function-like (`audit.go:480-493`), so every edge raises the ratio with no denominator movement. `DEPENDS_ON` is counted by nothing, so the old anchor contributed 0 and the new one contributes 1. It feeds `refHealth`/`RiskScore` (`audit.go:584`) and the "REFERENCES emission missing" gate at `audit.go:680`, **which it can therefore mask.**

**Price of the fix:** nothing in `internal/quality` reads `ref_kind` today. Excluding `field_target_type` from that one numerator is a ~5-line change at `audit.go:382-384` plus a test — but it affects **arms A–D at once**, so it belongs in its own issue rather than smuggled into the Go arm. Reported, not absorbed.

## Digest and ratchet movement — enumerated

- **`go test ./cmd/grafel/`**: green, fresh `GRAFEL_HOME` + `GRAFEL_DAEMON_ROOT`; parity tests confirmed **run, not skipped**.
  **The unmoved digest is NOT evidence for this change and is not cited as such.** That digest is a same-binary **parity** set (`worktree_seed_parity_test.go:161`) keyed `(FromID, ToID, Kind)` with properties discarded — it cannot see `ref_kind`, could not move for this arm by construction, and would not see a duplicate edge either. Running it is still correct; reading it as coverage is not. (My brief described it as a cross-run pin; that was wrong.)
- **`scripts/quality/run.sh --runs 1 --ratchet`**: caught exactly one fixture — `go-chi-mini`, `relationship_extracted_total` 207 → 208.

**Delta enumerated, not bumped blind.** I extracted the real fixture tree and listed the new edges: exactly one, `MemoryStore.users -> User` (`store/store.go`, `users map[string]User`, `User` declared in the same file). That accounts for +1 precisely. Note `UsersHandler.Store store.Store` correctly gets **nothing** — a qualified cross-package type. Hand-edited that **one number**; `git diff` on the baseline is a single line. No blanket `--update-baseline`. Re-run: **38 gated fixtures held**, `git status --porcelain` empty.

## Verification

- `go test -count=1` green: `./internal/extractors/golang/...`, `./internal/resolve/...`, `./internal/quality/...` (incl. `analytics`, `audit`, `fitness`), `./internal/extractor/...`, `./internal/graph/...`, `./cmd/grafel/`.
- `go test ./internal/quality/...` **and** the extraction-quality gate run separately — **both green, both reported.**
- `gofmt -l internal/` clean.

## Axes — varied vs held constant

Main fixture **varies**: wrapper form (bare, pointer, slice, array, map key *and* value, channel, func param, generic argument), target entity kind (`Component/struct`, `Component/interface`, `Schema/type_alias`), wire-name form (plain vs json-tag renamed). **Holds constant**: file path, owning struct, and target name (`Customer`) — so a wrapper regression reads as a missing row, not a changed target.

Axes needing a different file or declaration shape get **their own fixture**, because they cannot vary inside the main one without also varying something else: cross-file, enum shadow, blank-import name sharing, shadowed predeclared identifier, qualified-name strip, type-parameter over-fire.

Every forbidden row is asserted **by name** (`Qty`, `Label`, `Foreign`, `Nested`, `Anon`, `Self`, `St`), not by a count. Each negative assertion carries a **positive control in the same fixture** — `Status` (shadowed, no edge) beside `Tier` (identical form, no const block, edge present); `Holder.Foreign` beside `Holder.Local`; cross-file `Order.Buyer` beside same-file `Order.Local` — so an absence can never pass by the pass simply being off.

## Per-conjunct mutant verdicts — RE-SCORED after review

Every edit proven to land by an **exact occurrence count**; `cp`-restore from a backup taken **from the tree being mutated**, shasum-verified back afterwards.

| Mutant | Verdict |
|---|---|
| M1 — pass 2 `SourceFile != filePath` | **DEAD** |
| N2 — pass 1 `SourceFile != filePath` (M1's twin) | **DEAD** *(new)* |
| M2 — drop `Name == ""` | equivalent-under-the-suite |
| **M3 — allow-list** | **DEAD** *(was mis-scored; see below)* |
| M4 — arm C's `Kind != "SCOPE.Component"` | **DEAD** ×3 |
| M5 — ambiguity guard | **DEAD** ×3 |
| M6 — ambiguity counts records, not kinds | **DEAD** ×3 |
| M7 — collector skip: `qualified_type` | **DEAD** |
| M8a — collector skip: `struct_type` | **DEAD** |
| **M8b — collector skip: `interface_type`** | **DEAD** *(was vacuous)* |
| M9 — self-reference guard | **DEAD** |
| N1 — per-field ToID dedup | **DEAD** *(new)* |
| M10 — field-record guard in attach | equivalent-under-the-suite |
| M11 — stash uses `resolveTypeReferences` | **DEAD** ×3 |

**12 DEAD, 2 equivalent.**

### B1 — I was wrong about the allow-list, and the experiment I offered proved nothing

I claimed the allow-list was belt-and-braces and cited an M3+M5 compound as evidence. **M5 is independently lethal, so the compound killed exactly the tests M5 alone kills** — it carried no information about M3. A compound mutant grades its second half only if its first half is *not* lethal alone. Checking the compound was the right instinct; the conclusion was not available from it.

The claim was also false. Varying the one axis `BlankImportSharingATypeNameBinds` **holds constant** — *where* the type is declared — gives the distinguishing input:

```go
import _ "embed"   // SCOPE.Component, Subtype "", Name "embed"
type Order struct{ Blob embed }   // `type embed struct{…}` is in a SIBLING file
```

`import _` does not bind the identifier, and a Go package is a directory, so this is ordinary legal Go. The import placeholder is then the **only** same-file record named `embed`: one node (rule 2 silent), bare name ("refused by name shape" inapplicable). Without the allow-list, `Order.Blob` binds **to an import placeholder** — a wrong binding, which never reaches `bug-extractor` precisely because it binds.

Graded by `TestGoFieldTypeRefs_ImportPlaceholderIsNeverATarget`. **The allow-list is load-bearing, not belt-and-braces**, and the header now says so — and records the reasoning error, since that part is the more portable lesson.

### B2 — the `interface_type` skip was vacuous

`Anon interface{ Do() }` contains no `type_identifier`, so deleting `interface_type` from the skip list survived the whole package. Now `Get() Customer`. Scored as two mutants: **M8a and M8b both DEAD**, where the earlier "M8 DEAD ×2" was really DEAD once and vacuous once.

### N1 / N2 — the two cheap ungraded conjuncts, taken

- **N1**: added `Both map[Customer]Customer` (one declared type named twice) to grade the ToID dedup. DEAD. Matters because the audit pass has **no edge dedup**, so a duplicate would count twice in `ReferencesTotal` and `relationship_extracted_total`.
- **N2**: `goInFileTypeTargets` carries the `SourceFile` conjunct in **both** loops; I had paid to kill it in pass 2 and left pass 1 open. Added a `b.go` record sharing a name with an `a.go` type under a *different kind* — which is what makes a pass-1 leak observable at all. DEAD.

### The two survivors, stated rather than glossed

- **M2** is a dead store: a candidate is a `type_identifier`'s text, never empty, so `targets[""]` is never consulted.
- **M10** is masked because only field records receive the stash; if that changed, `goFTEdges` asserts no non-field record carries one of these edges.

### N4 — invariant documented

Rule 2 spans every record **at step 4b-bis**, not every record `Extract` returns. Steps 4c–6b append more that `buildSymbolIndex` sees. It holds today only because those producers emit *prefixed* names (`error_handling:`, `exception:`, `config:`) that no bare `type_identifier` can equal — now written down as an invariant to preserve rather than an accident relied on.

## Known-wrong, pinned deliberately

`TestGoFieldTypeRefs_KnownOverFire_TypeParameterShadowedByASameFileType` — the pass does not model type parameters, so `type Box[T any]` beside `type T struct{}` over-fires. Pinned so a real fix breaks a test rather than changing behaviour silently.

## Honest floor

Same-file only. For Go — where a package is a *directory* of files — this is the dominant miss, larger than for C# or Rust. Closing it needs a cross-file package type view pass 1 does not have; binding bare type names across files is the exact hazard #6976/#6369 are about, and I did not open it.

Also recorded honestly: I could **not** construct a Go `type_spec` form that names a type without emitting an entity (all of `Stack[int]`, `[...]int`, `func(...int)`, `(int)`, `interface{ int | string }` etc. emit one). The "knownTypeNames is a superset" argument is therefore recorded as **unconstructed**, and the real reason for reading records instead is stated instead: a `map[string]bool` carries no Kind, so it cannot express either rule this pass needs.

### DIFF
--- a/internal/extractors/golang/extractor.go
+++ b/internal/extractors/golang/extractor.go
@@ -191,6 +191,17 @@ func (g *GoExtractor) Extract(ctx context.Context, file extractor.FileInput) ([]
 	records = attachClassContains(records, file.Path)
 
 	// ----------------------------------------------------------------
+	// 4b-bis. Issue #6912 — field→declared-type REFERENCES. Runs after
+	//     every primary-pass record for the file is in place, because the
+	//     pass refuses any target name carried by more than one same-file
+	//     record and the SCOPE.Enum value-sets (step 2) and import
+	//     placeholders (step 3) are part of that count. See
+	//     field_type_refs.go. The struct-anchored DEPENDS_ON emitted by
+	//     extractStructFieldDependencies is deliberately left untouched.
+	// ----------------------------------------------------------------
+	records = attachGoFieldTypeRefs(records, file.Path)
+
+	// ----------------------------------------------------------------
 	// 4c. Track A (analog of #641/#650/#670 for Go) — REFERENCES-edge
 	//     emission. Runs after every primary-pass entity is in place so
 	//     the file-scope symbol table covers functions, methods,
--- a/internal/extractors/golang/field_type_refs.go
+++ b/internal/extractors/golang/field_type_refs.go
@@ -1,0 +1,418 @@
+package golang
+
+import (
+	"strings"
+
+	"github.com/cajasmota/grafel/internal/extractor"
+	"github.com/cajasmota/grafel/internal/treesitter/ts"
+	"github.com/cajasmota/grafel/internal/types"
+)
+
+// field_type_refs.go — the field→declared-type edge for Go (issue #6912, arm D).
+// Arm A is internal/extractors/csharp/field_type_refs.go (#6984), arm B is
+// internal/extractors/proto/field_type_refs.go (#6991), arm C is
+// internal/extractors/rust/field_type_refs.go (#7000).
+//
+// THE GAP, AND WHY GO IS ARM B'S SHAPE RATHER THAN ARM C'S. Go is the ONLY
+// language on this issue that already ships a same-file-gated declared-type
+// edge: extractStructFieldDependencies emits DEPENDS_ON per struct field type,
+// gated on collectDeclaredTypeNames. But it anchors that edge on the STRUCT
+// (`FromID: ownerName`), so the struct is wired and the FIELD is not. Every
+// SCOPE.Schema/field entity extractStructFieldEntities emits (#4850) is a leaf
+// on its outbound side, which is the #6912 population: #6726 measures
+// SCOPE.Schema at 99.3% orphan across their corpus, and Go is one of their
+// languages.
+//
+// So this pass adds a SECOND edge, anchored on the field, and leaves the
+// struct-anchored DEPENDS_ON exactly as it was. That is arm B's ruling: the
+// duplicated thing is the ANCHOR, not the information. "Which FIELDS point at
+// Order?" is what the new anchor answers and the old one cannot — DEPENDS_ON
+// names no field at all, not even in a property, so after this pass the two
+// edges carry different facts rather than the same fact twice.
+//
+// DOES THE SECOND ANCHOR DOUBLE-COUNT IN internal/quality? This was arm B's
+// open question and it lands here because Go is the only language shipping both
+// edges. MEASURED, by reading the arithmetic rather than reasoning about it —
+// answer: no metric counts the same fact twice, because no metric counts
+// ENDPOINTS; but two rates move, both in the flattering direction, and neither
+// dedups.
+//
+//   - internal/quality/audit/audit.go:424 OrphanRate is DEFLATED. `touched`
+//     (audit.go:363-368) admits any non-CONTAINS/DECLARES edge in EITHER
+//     direction, so a field that was previously reachable only by its CONTAINS
+//     edge flips to non-orphan the moment it is the FromID here. That is the
+//     #6912 gap closing and is the intended effect — but it is worth naming that
+//     the connectivity is derived from a fact the graph already recorded at the
+//     struct anchor. Note internal/quality/fitness/rule.go:511-526 computes a
+//     SECOND, inbound-only orphan rate that will NOT move for the field, so the
+//     two published orphan numbers diverge further after this arm.
+//   - audit.go:433 ReferencesPerFunction is INFLATED, and this one is a genuine
+//     distortion rather than a gain. Its numerator counts REFERENCES EDGES
+//     (audit.go:382-384) while its denominator counts FUNCTION ENTITIES, and a
+//     SCOPE.Schema/field is not function-like (audit.go:480-493) — so every edge
+//     added here raises the ratio with no denominator movement. DEPENDS_ON is
+//     counted by nothing, so the old anchor contributed zero and the new one
+//     contributes one. It feeds refHealth/RiskScore (audit.go:584) and the
+//     "REFERENCES emission missing" gate at audit.go:680, which it can therefore
+//     mask. The honest fix is for that metric to exclude ref_kind=field_target_type
+//     — nothing in internal/quality reads `ref_kind` today — but that is a
+//     quality-metric change touching every arm A–D at once, not a Go-only one,
+//     so it is reported rather than smuggled in here.
+//
+// THREE PLACES THIS ARM DELIBERATELY DOES NOT COPY ITS NEIGHBOUR
+//
+//  1. THE ADDRESS IS STRUCTURAL, NOT A BARE NAME. The adjacent DEPENDS_ON uses
+//     `ToID: t` — the bare type name. That is not the address arms A–C settled
+//     on, and adjacency is not a reason to inherit it: a bare name is resolved
+//     by name-family lookup across the whole graph and is what #6986 measured at
+//     91.6% dangling for the custom lane's `Class:<Name>` form. This pass uses
+//     BuildComponentStructuralRef — `scope:component:class:go:<file>:<Name>` —
+//     which is FILE-SCOPED, so a second .go file declaring a same-named type
+//     cannot make it ambiguous. Post-resolution the ToID is the target entity's
+//     ID, so no query sees the dialect. The existing DEPENDS_ON keeps its bare
+//     name; re-spelling it is a graph-wide change with no bearing on #6912.
+//
+//  2. THE ALLOW-LIST ADMITS SCOPE.Schema, BECAUSE IN GO A TYPE ALIAS IS ONE.
+//     Arm C's guard is `Kind != "SCOPE.Component"`, which is correct for Rust
+//     (buildRustTypeAlias emits a Component) and WRONG here: extractTypes emits
+//     `type Celsius float64`, `type Handler func()`, `type UserID string` and
+//     the Go-1.9 `type T = U` form as Kind SCOPE.Schema, Subtype type_alias
+//     (extractor.go:11-13, and the two type_alias branches). Inheriting arm C's
+//     kind check verbatim would have silently dropped every alias target — the
+//     single largest non-struct declared-type population in idiomatic Go. So the
+//     allow-list is written on Go's own buildComponent behaviour:
+//
+//     SCOPE.Component + struct     → admitted (type T struct{…})
+//     SCOPE.Component + interface  → admitted (type T interface{…})
+//     SCOPE.Schema    + type_alias → admitted (type T U / type T = U)
+//
+//     Everything else is refused. Being HONEST about what that refusal is worth:
+//     the excluded records are refused by NAME SHAPE before the subtype check
+//     ever sees them, so the subtype conjunct is a belt-and-braces guard rather
+//     than a live filter, and the PR body reports it as such rather than
+//     claiming a kill it does not have.
+//
+//     - IMPORT ENTITIES are SCOPE.Component with an EMPTY Subtype and Name set
+//     to the whole import path (extractImportEntities), never truncated to a
+//     segment. A single-segment path IS a bare name sharing this file's
+//     namespace, and it is reachable — `import _ "embed"` does not bind the
+//     identifier, so a file may legally BLANK-import "embed" and also declare
+//     `type embed struct{…}`. But both records are SCOPE.Component/embed in this
+//     file, hence ONE graph node under rule 2, so admitting the import record
+//     would address the same node by the same ToID and change no output. The
+//     case is still pinned — TestGoFieldTypeRefs_BlankImportSharingATypeNameBinds
+//     — because it is the case that proves rule 2 must count KINDS and not
+//     records: under a record count this edge disappears.
+//     - SCOPE.Schema + field is refused so a field never targets a field. Also
+//     unreachable by name shape: a field entity's Name is dotted
+//     (`Order.Buyer`) and a candidate is a single type_identifier, which cannot
+//     contain a dot.
+//     - SCOPE.Component + file (#577) is refused; its Name is the file path.
+//
+//  3. THE CANDIDATE COLLECTOR IS NOT resolveTypeReferences, and this is the one
+//     correctness fix in the arm. resolveTypeReferences collects EVERY
+//     descendant type_identifier, which for Go's `qualified_type` node
+//     (`other.Order`) is the TRAILING SEGMENT — so it strips `other.Order` to
+//     `Order`. In a file that also declares its own `Order`, the struct-anchored
+//     DEPENDS_ON therefore binds to the WRONG entity today, silently, because a
+//     bound edge never reaches `bug-extractor`. Arms B and C both carry a
+//     forbidden row for exactly that mutation. goFieldTypeCandidates skips
+//     `qualified_type` WHOLE rather than taking its tail, so the new edge is
+//     deliberately narrower than its neighbour on qualified names. Pinned in
+//     both directions by TestGoFieldTypeRefs_QualifiedTypeIsNotStrippedToASameFileName,
+//     which also pins that DEPENDS_ON's behaviour is UNCHANGED by this arm —
+//     narrowing resolveTypeReferences is a separate fix with its own blast
+//     radius, not a rider on this one.
+//
+// A NON-BINDING EDGE IS WORSE THAN NO EDGE, AND GO HAS A PARALLEL POPULATION
+// THAT MAKES THIS SHARPER THAN IT WAS FOR RUST. An unresolved stub is kept
+// verbatim and dangling and classified `bug-extractor`, so every miss is a full
+// hit. The Go-specific hazard is extractGoEnums: a `type Status int` beside a
+// const block typed by it produces a SCOPE.Enum value-set named `Status`
+// (extractor.EnumEntity) IN ADDITION to the SCOPE.Schema/type_alias of the same
+// name in the same file. The resolver's byLocation fallback is what binds a
+// component-space ref to a Schema entity, and that fallback consults
+// ambigLocation FIRST — so a name carried by two same-file entities returns
+// statusAmbiguous and the edge dangles.
+//
+// Arm C's collision check only looked for duplicates WITHIN its own admitted
+// set and would not have seen this: the alias is admitted and the enum
+// value-set is not. So goInFileTypeTargets counts across EVERY same-file record
+// and drops any name denoting more than one graph node, whether or not the
+// second node is a legal target. That is strictly the resolver's own ambiguity
+// rule, mirrored at the emit site so the pass declines to emit rather than
+// emitting a stub that cannot bind.
+//
+// The rule is graded by a POSITIVE CONTROL IN THE SAME FIXTURE rather than by
+// the shadowed case alone: `type Status int` carries a typed const block and
+// gets no edge, while `type Tier int` — the same declaration form with no const
+// block, hence no value-set and one node — DOES get one. Without that pair the
+// test cannot tell "the enum shadow is respected" from "a type_alias target
+// never works at all", which is the failure mode a recall-only assertion has.
+// See TestGoFieldTypeRefs_TypeShadowedByAnEnumValueSetGetsNoEdge.
+//
+// WHAT THIS PASS IS NOT
+//
+//   - SAME FILE ONLY. A field whose type is declared in another file of the
+//     same package gets nothing, and for Go — where a package is a directory of
+//     files and one-type-per-file is not the convention — that is the dominant
+//     miss, larger than it was for C# or Rust. Closing it needs the cross-file
+//     package type view that pass 1 has none of; binding a bare type name
+//     across files is the exact hazard #6976/#6369 are about. Separate arm, not
+//     a widening of this one.
+//   - IT DOES NOT UNDERSTAND TYPE PARAMETERS. `type Holder[T any] struct { Item
+//     T }` yields candidate `T`, which is dropped only because nothing in the
+//     file happens to be DECLARED `T`. A file that also declares `type T
+//     struct{}` gets a wrong edge. Known-wrong and pinned as such by
+//     TestGoFieldTypeRefs_KnownOverFire_TypeParameterShadowedByASameFileType, a
+//     case a real fix is expected to break.
+//   - AN ANONYMOUS NESTED STRUCT OR INTERFACE IS NOT A TARGET AND CONTRIBUTES
+//     NONE. `Inner struct { A Order }` yields nothing: `Order` is the declared
+//     type of `Inner.A`, not of `Inner`, and emitting `Inner → Order` with
+//     target_type=Order would assert something false. struct_type and
+//     interface_type are therefore skipped whole. This costs no reachable
+//     recall — `topLevelFieldDeclarations` already refuses to treat nested
+//     struct members as the outer struct's fields (#4850), so the nested field
+//     has no entity of its own to carry the edge either.
+
+// goFieldTypeRefsMetaKey is the per-field stash written by
+// extractStructFieldEntities and consumed — and deleted — by
+// attachGoFieldTypeRefs.
+//
+// The candidates cannot become edges during the walk. Two independent reasons,
+// either sufficient: `type Order struct { Buyer Customer }` may be declared
+// BEFORE `type Customer struct{}` in the same file, so the in-file declaration
+// set is complete only once the walk has finished; and the collision rule above
+// must see the enum value-sets and import records, which are appended to
+// `records` after extractTypes returns.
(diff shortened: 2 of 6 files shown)


cajasmota:
# REQUEST-CHANGES

Independent adversarial review, scored at head `1ea9bb32532260bd151b1d2e340aa7468321fd6a` in a private worktree. Every mutant edit was applied from a backup taken **from this tree** with an exact-occurrence-count assertion, restored by `cp`, and the tree verified back to `4ca5d3bf0393b2aee162f591806c81248cbcc04c` with an empty `git status --porcelain` after each round. Never `git checkout --`.

The change is correct and the reasoning is unusually careful. What I am blocking on is small and cheap: **one claim in the header comment is false, the experiment offered as its evidence proves nothing, and one forbidden row in the main fixture is vacuous.** Total cost of the fix is about fifteen lines and one fixture character.

---

## What reproduced exactly

I rebuilt the corpus measurement independently (own harness, `extractor.Get("go")` over `archigraph-corpora`, read-only, sequential) rather than trusting the table:

| | reported | I measured |
|---|---|---|
| Go files scanned | 686 | **686** |
| files emitting edges | 73 | **73** |
| `field_target_type` edges | 168 | **168** |
| `Component/struct` targets | 107 | **107** |
| **`Schema/type_alias` targets** | **32 (19%)** | **32 (19.0%)** |
| `Component/interface` targets | 29 | **29** |
| top repos | etcd 87, grpc-go-examples 44, gin 25, chi 12 | **identical** |

Also confirmed green here: `./internal/extractors/golang/...`, `./internal/quality/...` (incl. `analytics`/`audit`/`fitness`), `gofmt -l internal/extractors/golang/`, `go vet`, and `./cmd/grafel/` (154s, fresh `GRAFEL_HOME` + `GRAFEL_DAEMON_ROOT`, exit 0). The seed/parity digest tests **ran, not skipped** (`TestWorktreeSeedParity_SeededGraphMatchesAFullIndex`, `..._PathAOnASeededBaseMatchesPathAOnASelfBuiltBase`, `TestContentParity_PathA_6129` all PASS).

---

## The three design departures — all three are right

**1. Admitting `Schema/type_alias` — CORRECT, and the 19% is real.** Arm C's guard is verified verbatim at `internal/extractors/rust/field_type_refs.go:221` (`r.Kind != "SCOPE.Component"`), and my independent count says it would drop exactly 32 of 168. Not argued, measured, and it reproduces. On "can admitting `Schema` admit something that should not be a target": the subtype conjunct (`type_alias`) is what carries that, and `SCOPE.Schema/field` — the only other Schema subtype in play — is unreachable by name shape anyway (dotted Name vs. a single `type_identifier`). No new hazard.

**2. Counting KINDS, not records — CORRECT, and the mirror genuinely holds.** I checked the mirror rather than crediting it. `graph.EntityID` (`internal/graph/graph.go:259`) hashes `(repo, kind, name, sourceFile)` — **Subtype is not an input** — so within one file two records collide into one node exactly when they share a Kind, which is precisely what `existing != e.ID` at `internal/resolve/refs.go:1308` tests. The mirror is exact.

Both directions reproduced independently:
- **M5** (drop the ambiguity guard) → **DEAD**, kills `TypeShadowedByAnEnumValueSetGetsNoEdge` + `ResolvesToEntityIDs`.
- **M6** (count records instead of kinds) → **DEAD**, kills `EdgeSetIsExact` + `BlankImportSharingATypeNameBinds` + `ResolvesToEntityIDs`.

`TestGoFieldTypeRefs_ResolvesToEntityIDs` is the strongest thing in the PR: it runs the real `resolve.BuildIndex` / `ReferencesEmbedded`, asserts the resolved endpoint set exactly, and carries a positive control proving the bare name really is ambiguous in that fixture.

**3. Declining `resolveTypeReferences` — CORRECT; your brief's suggestion would have shipped a wrong binding.** Verified at `internal/extractors/golang/extractor.go:2154`: it is `findAll(node, "type_identifier")` over all descendants, and tree-sitter-go's `qualified_type` carries its `name:` as a `type_identifier` — so `other.Order` does become candidate `Order`, and in a file that declares its own `Order` the struct-anchored `DEPENDS_ON` binds to the wrong entity today. Reusing it here would have reproduced that on the new anchor. Pinning it as pre-existing rather than riding a narrowing on this PR is the right call. (Same helper also descends anonymous `struct_type`, so `DEPENDS_ON` over-fires there too — also pre-existing, also correctly out of scope.)

---

## Blocking findings

### B1 — the M3+M5 compound proves nothing about M3, and M3 is a live filter

**M5 alone is already DEAD.** I scored the compound M3+M5 and it kills the *identical two tests* M5 alone kills — not one test more. A compound mutant whose one half is independently lethal cannot grade the other half. So the experiment offered as evidence for "the allow-list is belt-and-braces" carries no information about the allow-list at all. This is the exact shape of the trap ("two guards that only fire together grade neither") — the instinct to check the compound was right, but the conclusion drawn from it is not available from that experiment.

**And the underlying claim is false.** The header says the excluded records "are refused by NAME SHAPE before the subtype check ever sees them" and that admitting the import record "would address the same node by the same ToID and change no output". That holds only when the type is declared **in the same file** — which is exactly what `TestGoFieldTypeRefs_BlankImportSharingATypeNameBinds` holds constant. Vary that one axis and the claim breaks:

```go
package models

import _ "embed"          // SCOPE.Component, Subtype "", Name "embed"

type Order struct {
	Blob embed            // `type embed struct{…}` lives in a SIBLING file
}
```

- HEAD: `EDGES=[]` — correct, the same-file rule refuses it.
- M3 (allow-list removed): `EDGES=[Order.Blob -> scope:component:class:go:models.go:embed]` — a field bound to an **import placeholder**.

Both runs captured directly. `import _` does not bind the identifier, so this is legal Go, and a package is a directory of files, so the sibling-file declaration is the ordinary case.

I enumerated the reachable space rather than hand-picking: of the records the Go extractor emits at steps 1–4b with a bare (non-dotted, non-path) Name, package-level funcs and the file entity are excluded by Go's own package-scope uniqueness or by name shape; enum value-sets require a same-file named type and therefore always present two kinds; **the import placeholder is the only one that can stand alone.** So the hole is narrow — but the allow-list is load-bearing, not belt-and-braces, and a false "this guard has no kill" note in a header is the defect class this board keeps getting caught by.

**Verdict: distinguishable and ungraded. Price ~12 lines** — the fixture above as a forbidden row (`Order.Blob` carries no edge), plus correcting the header paragraph to say the allow-list *does* have a kill and naming the input.

### B2 — the `interface_type` half of the collector's skip conjunct is vacuous

The main fixture's forbidden row is `Anon interface{ Do() }`. That body contains **no `type_identifier` at all**, so removing `"interface_type"` from the skip list changes nothing and survives the entire package. I scored it: **ALIVE across the full `./internal/extractors/golang/` suite.** The distinguishing input is one character-class away:

```go
Anon interface{ Get() Customer }
```

- HEAD: `EDGES=[]` (verified).
- mutant: `EDGES=[Order.Anon -> …:Customer]` — asserting `Customer` is `Anon`'s declared type, which is false, and is the same class of error the anonymous-struct row exists to forbid.

So M8's "DEAD ×2" is really DEAD once and a vacuous row once: `Order.Nested` grades `struct_type`, `Order.Anon` grades nothing. **Price: change the fixture's `Do()` to `Do() Customer`.** The expected set does not move.

---

## Non-blocking, priced

**N1 — the `emitted[t.toID]` dedup is ungraded.** `map[Customer]Customer` and `func(Customer) Customer` each yield the same candidate twice; with the dedup removed both emit two identical edges (captured). The fixture varies the map axis for *both halves judged* (`map[Tier]Customer`) but leaves the same-name-twice neighbour open. It matters because the audit pass has **no edge dedup** (confirmed at `audit.go:355-388`), so a duplicate would count twice in `ReferencesTotal` and in `relationship_extracted_total`. Zero occurrences in the 686-file corpus, so low urgency — but corpus-absence is corpus-relative. **Verdict: distinguishable and ungraded. Price ~2 lines** (one fixture field + one expected row).

**N2 — pass 1's `SourceFile != filePath` is the ungraded twin of M1.** `goInFileTypeTargets` has that conjunct in *both* loops. You paid an 82-line internal test to kill it in pass 2 — correctly, and I confirm that kill is real and not manufactured: `field_type_refs_scope_6912_test.go` calls the function directly with a genuine two-file slice and asserts both directions (`b.go`'s types absent from `a.go`, and `b.go`'s own set non-empty so the first assertion is not vacuous). But deleting the conjunct from **pass 1** survives the whole suite. Distinguishing input: the same multi-file slice with one name declared in two files under different kinds — the cross-file count then over-suppresses a legitimate same-file target. **Verdict: alive but not production-reachable** (identical status to M1 pre-fix). Grading one twin and not the other is the inconsistency; **price ~4 lines** in the file that now exists.

**N3 — "digest unmoved" is not evidence, and should not be cited as such.** The `cmd/grafel` digest is a **parity** digest between two runs of the same binary (`worktree_seed_parity_test.go:161`), keyed `(FromID, ToID, Kind)` as a **set**, with properties discarded. It cannot move for this change by construction; it could not see `ref_kind`, and it would not see N1's duplicate edge either. Running the suite is still right (I ran it, green, parity tests not skipped) — just retire the sentence that reads as a signal.

**N4 — rule 2 does not actually span *every* same-file record.** `attachGoFieldTypeRefs` runs at step 4b-bis, so records appended at 4c–6b (`SCOPE.Pattern`, `SCOPE.ExceptionType`, config keys, constant sets) are invisible to it while `buildSymbolIndex` sees them. Today every one of those carries a prefixed name (`error_handling:…`, `exception:…`, `config:…`), so no collision is constructible — and I verified it empirically: re-running all 168 corpus edges against the **post-`Extract`** record set, **0** target names carry more than one kind and **0** targets lack a same-file declaration. Not a bug; the invariant is just undocumented, and a future producer emitting a bare-named entity would silently reintroduce dangling. One sentence.

**N5 — for #7000, not this PR.** `rustInFileTypeTargets` counts **records**, not kinds, and since every admitted rust record is `SCOPE.Component`, two same-named records there are one node — so arm C is over-strict in exactly the direction Go's M6 grades. Likely unreachable in Rust, but worth a note on that issue.

---

## The other things you asked me to check

**Ratchet hand-edit — enumeration is COMPLETE and equals a correct regenerate.** I ran my harness over `internal/quality/golden/` independently: exactly **one** new edge across all five Go fixture files, `MemoryStore.users -> scope:component:class:go:…/store/store.go:User` (`field_name=users`). `go-chi-mini` is the **only** golden fixture containing `.go` files, so no other baseline number can move from this arm. +1 accounts for `207 → 208` precisely; the `baseline.json` diff is a single line; no blanket `--update-baseline`. Confirmed.

**Double-count answer — verified line by line, and the `ReferencesPerFunction` distortion is exactly as described.** `touched` (`audit.go:363-368`) admits both endpoints of any non-structural edge, so the field flips to non-orphan on being a `FromID` — the intended #6912 effect. `ReferencesTotal` counts REFERENCES **edges** (`audit.go:381-383`), `totalFuncs` sums `FunctionsByLanguage`, and `isFunctionLike` (`audit.go:480-493`) matches only function/method/operation — a `SCOPE.Schema/field` is not function-like, so every edge here raises the ratio with zero denominator movement. It feeds `riskScore` and the gate in `recommend()` (`lr.Functions > 100 && lr.ReferencesPerFunction < 0.5`), so **yes, it can suppress a genuine "REFERENCES emission missing" recommendation.** And `fitness/rule.go:511-526` is inbound-only (`hasInbound[…ToID]`), so it will **not** move for the field and the two published orphan numbers do diverge further. Your read is correct on every point; filing it separately is the right call, since it spans arms A–D.

**Retracted claims.** Both corrections landed in the header: the "declared but unemitted type" argument is now explicitly recorded as `UNSUPPORTED` with the real reason given (a `map[string]bool` carries no Kind), and the blank-import paragraph was rewritten. Nothing downstream rests on the retracted *type_spec* version. The blank-import correction, however, is **incomplete** — B1 is precisely its residue.

---

## Summary

Nothing here emits a wrong edge in production today, the corpus numbers reproduce to the unit, the ratchet delta is exact, and all three departures from arm C's template are correct and better-reasoned than the template. What blocks is that the PR states a guard has no kill when it has one, offers a compound experiment that cannot support that statement, and carries a forbidden row that grades nothing. B1 (~12 lines) and B2 (one fixture line) and I'll re-approve on sight.

cajasmota:
## Coordinator mutants on `b070a6832` — **CZ-1 DEAD, and a hypothesis of mine disproved**

Own detached worktree at `/private/tmp/.../coord-mut-7036-t9x/wt` (absolute), private script dir. Green baseline immediately before. Both files restored by `cp` from SHA-256-verified backups (`bad5ae2b…`, `6b632375…`); `git diff HEAD` and `git status --porcelain` empty afterwards.

Fourteen mutants have been scored between you and the reviewer, and they cover the guards thoroughly — the allow-list, the ambiguity rule in both directions, the interface skip, the ToID dedup, both pass-1 conjuncts. So I went after the **payload's vocabulary** rather than its guards, because this P1 is shipping the *same property spelling* across eight languages and a wrong spelling is a silently unconsumable edge rather than a missing one.

### CZ-1 — near-miss the ref_kind spelling

`goFieldTargetRefKind = "field_target_type"` → `"field_target_types"` (`field_type_refs.go:222`). Occurrence asserted 1.

**Result: DEAD — 9 failing tests.** The spelling is pinned by independent literals in the test file, not derived from the constant under test, so a table driven by the production value could not have hidden this.

### CZ-2 — the hypothesis, and it was wrong

There are **four independent hardcoded constants**, one per shipped arm:

```
csharp/field_type_refs.go:94    csFieldTargetRefKind    = "field_target_type"
proto/field_type_refs.go:105    protoFieldTargetRefKind = "field_target_type"
rust/field_type_refs.go:149     rustFieldTargetRefKind  = "field_target_type"
golang/field_type_refs.go:222   goFieldTargetRefKind    = "field_target_type"
```

No shared constant, no cross-arm agreement test, and no test anywhere references any of the four constants by name. My hypothesis was that **arm-to-arm agreement is therefore ungraded** — that one arm could drift and only its own package would notice, leaving consumers keying on `ref_kind` silently unable to read that language.

I scored it on the **shipped** rust arm rather than on yours, since that is where a latent gap would already be live: `rustFieldTargetRefKind` → `"field_target_types"`.

**Result: DEAD — 10 failing tests in rust's own package.**

**So the hypothesis is disproved and the current shape is fine.** Every arm asserts the literal string `"field_target_type"` in its own tests, independently of its own constant. Four independent literals all naming the same string enforce agreement transitively: any single arm's drift is caught by that arm, so no combination can diverge silently. A shared constant would be tidier, but it would be *one* grader instead of four, and an independent literal per arm is the stronger pattern — the same reason `dormant_alias_order_7030_test.go` keeps its own copy of the expected slice.

Recording the negative result deliberately. "I checked whether the vocabulary could drift across arms; it cannot, and here is why" is worth more to the next person than silence, and it removes a plausible-sounding concern from the backlog rather than leaving it to be re-raised.

### On round 2

**B1 is the round's real work, and you generalised it correctly.** The compound proved nothing because M5 is independently lethal, so it killed exactly M5's tests — and you put *"a compound grades its second half only if its first half isn't lethal alone"* in the header, which is the portable form of the lesson rather than the local fix. The claim itself was also backwards: with `type embed` in a **sibling** file, the import placeholder is the only same-file record of that name, so **only the allow-list** stops a field binding to it. `TestGoFieldTypeRefs_ImportPlaceholderIsNeverATarget` closes it and M3 is now DEAD.

Three more things worth naming:

- **B2 was DEAD once and vacuous once.** Splitting it into M8a/M8b turned a misleading "DEAD ×2" into two real kills. That is the same fixture-label-overclaims-its-body defect this board keeps finding, caught in your own mutant table.
- **You retracted the digest sentence rather than defending it.** It is a same-binary **parity** set keyed `(from,to,kind)` with properties discarded — it cannot see `ref_kind` and could not have moved. Running it stays right; reading it as coverage was not, and my brief told you the wrong thing there.
- **You corrected an existing assertion whose stated reason your own new fixture had invalidated** (absence now true for two independent reasons). Nobody would have caught that but you, and leaving it would have been the exact defect B1 was about.

### N5 — I am filing it

`rustInFileTypeTargets` counts **records, not kinds** — the same over-strictness your M6 grades for go, in an arm that has already shipped. That is a latent wrong-answer in arm C found from arm D, which is the sort of thing only a second implementation surfaces.

**Merging on green.**

Issue #7038, opened 2026-09-08
rustInFileTypeTargets counts RECORDS not KINDS, so arm C refuses field->type edges that would bind — the exact over-strictness go's arm D grades as a killing mutant


cajasmota:
Found while implementing arm D (go, #7036) of #6912. **`rustInFileTypeTargets` counts records, not kinds — the same over-strictness go's ambiguity rule was deliberately written to avoid.** Arm C (rust) has already shipped, so this is latent in `main`.

## The mechanism

Both arms refuse to emit a field→declared-type edge when the target name is ambiguous within the file. The question is what "ambiguous" counts.

**Go counts kinds.** Two *records* sharing a name can still be **one node**, because `graph.EntityID` hashes `(repo, kind, name, file)` with **Subtype excluded** — mirroring `resolve/refs.go:1308`. So `import _ "embed"` plus `type embed struct{}` is two records but one node, and the edge is legitimate.

**Rust counts records** (`rustInFileTypeTargets`). On the equivalent Rust shape it will refuse an edge that would have bound correctly.

## Why this is a real defect and not a style difference

Go's arm graded **both** directions deliberately — a mutant that counts records instead of kinds is DEAD there, precisely because counting records **deletes a legitimate edge**. That mutant is the demonstration that record-counting is wrong, not merely different, and it was scored three times.

Arm C is the only shipped arm with the record-counting form. Arms A (C#) and B (proto) were not checked for this shape and **should be**, rather than assumed clean — see below.

## Direction of the error

**Under-emission, not wrong binding.** Rust refuses edges it could safely emit; it does not emit wrong ones. So this is a recall loss against #6912's own goal — closing `Schema` orphans for contributor **#6726** — rather than a correctness hazard. That makes it lower urgency than a mis-binding would be, and it is exactly the direction that a recall-only test suite cannot see (**recall cannot detect over-firing, and an over-strict guard produces no failing assertion at all**).

## Not established

- **How many rust edges are actually lost.** Zero measurement. The corpus yield for the equivalent go shape was 168 edges total with 32 (19%) depending on the `Schema/type_alias` admission, but **nothing about that transfers to rust** — different language, different subtype vocabulary, different corpus footprint. Anyone picking this up should measure before assuming it matters.
- **Whether arms A (C#) and B (proto) share the shape.** Not checked. They were written before go's kind-vs-record distinction was articulated, so they are equally likely to have it — and equally likely not to, since each arm made its own allow-list decisions.
[...]

Issue #7037, opened 2026-09-08
ReferencesPerFunction divides edges by FUNCTION entities, so every #6912 field->type edge inflates it — and it feeds the audit.go:680 gate meant to detect a MISSING extractor


cajasmota:
Found while measuring #7036 (the go arm of #6912). **`ReferencesPerFunction` divides an edge count by a *function-entity* count, so every field→declared-type edge inflates it with no denominator movement — and it feeds the gate that is supposed to detect a missing extractor.**

## The mechanism

`ReferencesPerFunction` has a `REFERENCES` edge numerator and a function-entity denominator. A **field** is not function-like, so a field-anchored `REFERENCES` edge raises the ratio while contributing nothing to the denominator.

That ratio is not decorative:

- it feeds **`RiskScore`**, and
- it feeds the **"REFERENCES emission missing" gate at `audit.go:680`**.

So the arms of #6912 — whose entire purpose is to *add* field→declared-type `REFERENCES` edges — **push the metric in the direction that says "REFERENCES emission is healthy"**. A language whose extractor genuinely emits no references could be masked by another language's field edges, because the ratio is not per-producer.

**This is a gate that gets quieter as the thing it watches for gets more likely.**

## Scope: all four arms at once, not just go

Arms A (C#, #6984), B (proto), C (rust, #7000) already shipped field→type `REFERENCES` edges; the go arm (#7036) adds 168 more. So the distortion is already live and grows with each remaining arm (java, swift, php, cpp, kotlin, scala are all queued on #6912). **That is why this is filed rather than absorbed into #7036** — fixing it inside one arm would leave the same defect credited to three others and make the arm's own numbers incomparable with its predecessors'.

The author priced the fix at **~5 lines** (`audit.go:382-384`) plus a test.

## A second, milder effect: two published numbers now diverge

`OrphanRate` **deflates** as these arms land — which is intended and is exactly the gap closing — but `fitness`'s **separate inbound-only orphan rate does not move**. So two orphan numbers that used to track each other now drift apart, and a reader comparing them will see a discrepancy that is an artefact of measurement rather than of the graph.

This matters for contributor **#6726** specifically: their report is expressed in orphan rates, and four of their six headline numbers were already defects in *our* measurement (#6599, #6536, #6826, #6827). Shipping a fifth measurement artefact into the same conversation is worth avoiding.

## Suggested shape

[...]

PR #7039, opened 2026-09-08
fix(#6912): java fields carried their declared type at no anchor at all — emit the field→type REFERENCES edge, same-file targets only

cajasmota:
Arm E of #6912 (A=C#/#6984, B=proto/#6991, C=rust/#7000, D=go/#7036), and the remaining language of contributor **#6726**'s corpus (csharp/fsharp/java/protobuf) that this issue covers.

Java field entities — from **both** emit sites, class `field_declaration` and record header component — were leaves on their outbound side. This adds one field-anchored `REFERENCES` edge with `ref_kind=field_target_type`, same-file targets only, matching the shipped vocabulary and address.

**The issue's long-standing claim that java "retains NO type string at all" is false**, as the grounding pass at `663f6c50a` found and as this arm re-verified: `typeName := nodeText(typeNode, …)` is a live local at the record-component site, and `buildFieldEntity` holds the `field_declaration` node whose `type` child `buildFieldSignature` already spans. No capture step was needed.

**Arm B's question, asked rather than inherited: is the FIELD the anchor, or is something else?** Unlike Go, java shipped **no declared-type edge at any anchor**. Three neighbouring populations look like one and are each left untouched — `javaInjectFieldTypes` (class-anchored, DI-annotation-gated, bare/`ext:` target), `emitReferences` (operation → field *usage*, opposite direction), and `internal/custom/java/{hibernate,orm_helpers}.go` (already spells `field_target_type`, but off a parallel `SCOPE.Component` population, and `internal/custom/**` defaults OFF; #6986 measured that lane's address at 91.6% dangling).

> **Second commit (`3609dfc`) is review follow-up.** It corrects **two claims that were invented rather than observed** and two reasons that were stated wrongly, and closes one mirror hole. Each is called out inline below with what was wrong. Corpus output is byte-identical across the two commits (147/147/0, same 125/18/4 breakdown).

---

## The allow-list, and why

Written on java's own `buildComponent` rather than inherited. Java routes `class_declaration`, `interface_declaration`, `enum_declaration` and `record_declaration` through **one** `buildComponent` call with only the Subtype varying (`java.go:324-355`, `:1374-1396`) — the grounding pass could not confirm the class subtype string because it is set through a variable; it is the literal `"class"` at `java.go:326`. So the allow-list is a **subtype** list:

| record | verdict |
|---|---|
| `SCOPE.Component` + `class` | admitted |
| `SCOPE.Component` + `interface` | admitted |
| `SCOPE.Component` + `enum` | admitted |
| `SCOPE.Component` + `record` | admitted |
| `SCOPE.Component` + `file` (#577) | refused — Name is the file path; **also** unreachable by name shape (always carries `.java`, and a `type_identifier` cannot contain a dot). Claimed as nothing more. |
| `SCOPE.Component` + empty subtype | refused — no producer emits one today; this grades the default arm against the producer not yet written |
| `SCOPE.Schema` + `schema` (`nosql_model.go:187`) | refused — **a live competitor**: `@Document class Customer` emits this named after the *class*, in the same file |
| `SCOPE.Schema` + `field` | refused — reachable in principle (a module-scope field keeps a **bare** name), so refused on Kind, not on name shape |
| `SCOPE.Enum` value-sets, `SCOPE.Operation`, `SCOPE.ExceptionType`, `SCOPE.Template`, config keys | refused |

The subtype universe reaching a java file's records is `{class, interface, enum, record, file}` — walk's four, `extractor.FileEntity`, `synthComp` (`"class"`), panache (`"interface"`, synthetic path). The four admitted are exhaustive and the two refusals are complete.

**Arm C's `Kind != "SCOPE.Component"` guard would have cost nothing here** — every java target is a Component. Arm D measured that same guard dropping 19% of Go's edges; the shape simply does not recur. That is exactly why this arm's real departure is on the *other* rule.

## The ambiguity rule — counting kinds, and the java-specific scope

Kinds, never records: `graph.EntityID` hashes `(repo, Kind, Name, SourceFile)` with **Subtype excluded**, mirroring `resolve/refs.go:1308`'s `existing != e.ID`, so two same-file records sharing a Kind are **one node**. Counting records refuses edges that would have bound — the defect filed as **#7038** against arm C.

> ⚠️ **Corrected in `3609dfc` — the original write-up's example was fabricated.** It claimed `class Order` could sit beside "Lombok's synthesized `SCOPE.Component` `Order`". It cannot: `synthComp`'s only call site (`lombok.go:438`) passes `className + "Builder"`, **never the class's own name** — and the test file then promised an `@Builder` fixture that was never written. Rather than delete the example, the actual reachable shape was found and pinned: **`@Builder class Order` synthesizes `OrderBuilder` into this file**, so a file that *also* declares `class OrderBuilder` by hand carries two `SCOPE.Component/class` records of one name — one graph node. `TestJavaFieldTypeRefs_LombokBuilderDuplicateComponentIsOneNode` now grades this **from java source**, asserts the two-record premise before the edge so it cannot pass vacuously, and kills M4 alongside the unit test. #7038's defect is now graded on a production-reachable shape rather than only a hand-built record set.

**The count is restricted to the component address family, not to every same-file record, and that is a deliberate departure from arm D.**

`buildJavaEnumValueSet` (`enum_valueset.go:25`) emits a `SCOPE.Enum` value-set named **identically** to the `SCOPE.Component/enum` walk emits from the same node in the same file. Under arm D's all-kinds rule *every* java enum is two kinds and **every enum target is refused**.

It would also be wrong on the resolver's own terms. `lookupStructural` (`refs.go:2982`) resolves a `scope:component:…` ref through `lookupLocationKind` filtered to `componentKindFamily` (`refs.go:2341`) **first**; `ambigLocation` — the kind-agnostic index arm D's rule mirrors — is only consulted **after that misses** (`refs.go:3004`). `SCOPE.Enum` is not in that family, so it never enters the tier that decides this address. Go's `type_alias` targets were `SCOPE.Schema`, *outside* the family, which is precisely why arm D fell through to `ambigLocation` and had to mirror it.

**Measured, not argued:** arm D's rule applied unchanged emits **129** edges here instead of 147 — it drops **18, 12.24%, and every one of them is an enum target**.

**Proved, not read:** `TestJavaFieldTypeRefs_ResolvesToEntityIDs` drives the **real resolver** over an extracted file carrying both records and asserts `Order.state` binds to the Component's ID. It carries class-, interface- and record-typed rows as positive controls, so a rule that emitted nothing could not pass it, and a bare-name-ambiguity precondition, so the structural address is not ungraded.

> ⚠️ **Corrected in `3609dfc`.** The original header cited `TestJavaFieldTypeRefs_SameFileEnumTypeBinds` — in detail, as the grader for this arm's headline departure — and **no such test was ever written**. The substance *was* graded (`ResolvesToEntityIDs` + `EdgeSet` + `Unit_EnumValueSetDoesNotShadowItsComponent`), so there was no coverage gap, but a named grader that does not exist is worse than none: the next reader greps for it, fails, and distrusts the rest. Fixed, along with three citations that dropped the `_Unit_` infix.

Both directions of the record-vs-kind question are graded: `LombokBuilderDuplicateComponentIsOneNode` (source) and `Unit_DuplicateComponentRecordsAreOneNode` (record level) kill a record-counting mutant; `Unit_TwoDistinctFamilyKindsAreRefused` kills a no-ambiguity mutant; `Unit_AnotherFilesCollisionDoesNotShadowThisFile` kills the pass-1 scope in the *permissive-in-reverse* direction.

### The one behaviour change in the follow-up commit: `SCOPE.Class`

The reviewer's mutant **MZ-1** — deleting `"Class"` from `javaComponentAddressFamily` — was **ALIVE**, and it exposed a genuine asymmetry. The comment claimed the map carried "both the bare and SCOPE.-prefixed spellings", yet **`SCOPE.Class` was absent**. `BuildIndex` dual-keys a `SCOPE.Class` entity under *both* `"SCOPE.Class"` and the trimmed `"Class"` (`refs.go:1208-1211`), and `"Class"` **is** in `componentKindFamily` — so such an entity blanks `uniqueMatchInFamily`, falls through to `ambigLocation`, and **dangles**, while this pass would happily emit for it.

The map is **not** a copy of `componentKindFamily`. It is that slice **closed under the index's trim alias** — `K ∈ family || TrimPrefix(K, "SCOPE.") ∈ family` — and the eight entries are now exactly that set. Unreachable from java source today (no java producer emits a `Class`- or `SCOPE.Class`-kinded entity); fixed anyway, because a mirror with a hole in it is worse than no mirror. Both spellings graded by `Unit_ScopeClassParticipatesViaTheTrimAlias`; **deleting either entry is now DEAD.**

## Qualified names — skipped whole, never reduced

`scoped_type_identifier` is skipped without descending. Arm D found Go's `resolveTypeReferences` stripping `other.Order` → `Order` and binding it to a same-file `Order` — silently wrong, since a bound edge never reaches `bug-extractor`. **Java's `Outer.Inner` form is worse**: *both* segments plausibly name a same-file type, so any reduction is a guess between two wrong answers.

This costs no argument recall: in `java.util.List<Order>` the `type_arguments` node is a **sibling** of the `scoped_type_identifier` under `generic_type`, so `Order` is still collected while `java`/`util`/`List` are not.

## Generics

The constructor **and** the arguments are collected and judged independently — arm A's and arm C's rule. `List<Order>` yields `[List, Order]`; `List` is refused by the in-file declaration check unless the file genuinely declares `class List`, in which case the edge is correct. `Map<String, Order>` yields both halves; `List<? extends Order>` unwraps the wildcard; `Order[]` unwraps the array; `Map<Customer, Customer>` emits **one** edge. No wrapper blocklist to keep in sync. A type **parameter** (`class Holder<T> { T item; }`) is a known over-fire, pinned as such with a control.

## Primitives, boxed types, java.lang

**Java primitives can never be candidates** — `int`/`long`/`short`/`byte`/`char` parse as `integral_type`, `float`/`double` as `floating_point_type`, `boolean` as `boolean_type`; none is a `type_identifier`, and `long[]` yields nothing at all. This is the inverse of Go, where `int` is a shadowable identifier and arm D had to pin the shadowing case. Enumerated against an independent literal list of all eight, each paired with a declared `class Int` positive control.

**Boxed and `java.lang` names ARE candidates** and are refused solely by the in-file declaration check — no name blocklist, which would be actively wrong: a file declaring its own `class Integer` genuinely means *that* type in field position. The same source text yields no edge in one file and an edge in the other; that pair is what proves the refusal is the declaration check.

## Corpus yield (read-only, capped at 3 cores)

| | |
|---|---|
| `.java` files | **645** — `kafka` 471, `kafka-streams-examples` 115, `spring-petclinic` 47, `awesome-compose` 11, `ratpack-example-books` 1 |
| repos contributing **edges** | 2 (`kafka` 137, `kafka-streams-examples` 10) |
| edges emitted | **147** |
| **bound** | **147** |
| **dangling** | **0** |
| bound targets | 125 `SCOPE.Component/class`, **18 `/enum`**, 4 `/interface` |

Independently reproduced by the reviewer with a separate in-process harness over the same 645 files, including the 18-dropped/129-remaining counterfactual.

### What 147/147 does *not* establish

- **There is no recall denominator.** Nothing here counts fields whose declared type *is* same-file-declared and got nothing, so the under-firing direction is unaudited **on the corpus**; over-firing is graded only by fixtures.
- **Zero `/record` targets** is corpus-relative, not a capability claim — `RecordComponentIsAnAnchor` and `AllFourComponentSubtypesAreTargets` both produce them.
- **The corpus is java-thin** and nothing here transfers to #6726's. The mechanism transfers; the number does not.
- **Recall ceiling** (added in `3609dfc`, observed by probe, all pre-existing and none introduced here): a **multi-declarator field emits only its first name** (`class M { Customer a, b; }` → `M.a` only, so `b` can never carry an edge); an **interface constant emits no field entity at all** (`interface Consts { Customer DEFAULT = null; }` → no record, no edge); and an **`@interface` type gets no Component**, so an in-file annotation type can never be a target. These cap the arm's ceiling and are worth a follow-up.

## Gate movement

- **`scripts/quality/run.sh --runs 1 --ratchet`** — green, and **`git status --porcelain` shows no baseline drift**, which is the real check.

  > ⚠️ **Enumeration corrected in `3609dfc`.** The original body said "all **6** `java-spring-mini` files declare exactly one type each". The golden set has **8** java files across **two** fixtures, and the omitted one holds **the only multi-type java file in the entire golden set** — `java-quartz-mini/src/jobs/SendEmailJob.java` — i.e. exactly the shape the claim turns on. The conclusion survives: **both quartz files declare no fields at all**, so there is still nowhere an edge could appear. Same defect class as arm D's "go-chi-mini is the only golden fixture with `.go` files"; the fix is to enumerate the *set*, not one repo in it.

- **`go test ./internal/quality/...`** — green, separately from the above; they are not the same gate.
- **`go test -count=1 ./cmd/grafel/`** — green in an isolated `GRAFEL_HOME`/`GRAFEL_DAEMON_ROOT` (140s, then 157s after the follow-up), digest tests run not skipped. **No digest movement is claimed as coverage.**

  > ⚠️ **Reason corrected in `3609dfc`, and it was wrong in the flattering direction.** The original said the digest "could not have moved for this change". `semanticDigest6118` keys relationships as `(fromTuple, Kind, toTuple)` with properties discarded, so it is blind to **`ref_kind`** — but it is **not** blind to a *new* `REFERENCES` edge appearing, which is what this arm adds. It could not move because its java fixture (`java/OrdersController.java` in `custom_extractor_spans_6118_test.go`) declares one class and **no fields**, so there is no field entity to carry an edge. **"Nothing to move", not "the instrument is blind."** Declining to cite it as coverage was right; the reason was not.

- `go test -count=1` green on `./internal/extractors/java/...`, `./internal/resolve/...`, `./internal/extractor/...`, `./internal/quality/{,analytics,audit,fitness}`, `./internal/graph/...`.

**`ReferencesPerFunction` will inflate** and is *not* fixed here — it spans arms A–E and is filed as **#7037**.

## Axes — varied vs held constant

The main fixture **varies**: wrapper form (bare, array, generic argument, qualified-generic argument, map key *and* value, wildcard, user-generic argument), the target's **declaration form** (class / interface / enum / record), and the **anchor** (class `field_declaration` vs record header component).
It **holds constant**: the file path, the declaring type (`Order`), and the target name (`Customer`), so a wrapper regression surfaces as a *missing row* rather than a *changed target*.

Axes that cannot vary inside it without also varying something else get their own fixture, each stated: cross-file targets need a second **file**; the type-parameter over-fire needs a file that **declares** the parameter's name; the nosql collision needs an `@Document` annotation; the boxed-type shadowing needs a second file with the local declaration; **the Lombok duplicate-Component case needs a `@Builder` beside a hand-declared `<Class>Builder`** — and that fixture now exists, which it did not when the sentence was first written.

The whole-set assertion is an **equality**, so it grades over-firing as loudly as under-firing, and every forbidden row is named in a comment against the decision it grades.

## Mutants — per conjunct

**22 of 22 DEAD**, re-scored in full after the follow-up commit. Every edit's landing proved by an exact occurrence count before the run and re-derived after (`EDIT-DID-NOT-LAND` / `EDIT-VERIFY-FAILED` are distinct harness outcomes). Private `GOCACHE` throughout — the shared cache was purged mid-run and produced phantom `build failed` / `package net/http is not in std` errors; three runs were discarded and re-run rather than scored.

| # | mutation (permissive direction unless noted) | verdict |
|---|---|---|
| M1 | allow-list **subtype** switch deleted | **DEAD** — 2 |
| M2 | allow-list **Kind** check deleted | **DEAD** — `Unit_NonComponentKindWithAnAdmittedSubtypeIsRefused`. Alive at first pass; *not* deferred as equivalent. Unreachable from java **source** (every same-file record with an admitted subtype is a Component, or a `SCOPE.Enum` value-set that **shares its name with one** and so computes the identical toID), but the distinguishing input is one line and was constructed. |
| M3 | ambiguity rule deleted | **DEAD** — 2 |
| M4 | ambiguity counts **records** not kinds (#7038's defect) | **DEAD** — now **2**: `LombokBuilderDuplicateComponentIsOneNode` (source) + `Unit_DuplicateComponentRecordsAreOneNode` |
| M5 | ambiguity spans **all** kinds (arm D's rule verbatim) | **DEAD** — 5, incl. `ResolvesToEntityIDs` |
| M6 | `scoped_type_identifier` descended | **DEAD** — 5 |
| M7 | self-reference skip deleted | **DEAD** — 3 |
| M8 | per-field target dedupe deleted | **DEAD** — 2 |
| M9 | `ref_kind` near-miss → `field_target_types` | **DEAD** — **15**, all against independent literals |
| M10 | address dialect → bare name | **DEAD** — 13, incl. `ResolvesToEntityIDs` |
| M11 | record-component stash removed | **DEAD** — 3. Grades the **second** emit site. |
| M12 | `buildField` stash removed | **DEAD** — 14 |
| M13 | emit guard deleted | **DEAD** — also alive at first pass, also constructed rather than deferred |
| M14 | owner prefix not stripped from `field_name` | **DEAD** — `EdgeProperties` |
| M15 | …guard's **Kind** conjunct alone | **DEAD** — fails *only* subtest `wrong_kind, field_subtype` |
| M16 | …guard's **Subtype** conjunct alone | **DEAD** — fails *only* subtest `schema_kind, wrong_subtype`. Complementary with M15, so neither conjunct masks the other. |
| M17 | stash not cleared | **DEAD** |
| M18 | pass-2 same-file filter dropped | **DEAD** |
| M19 | pass-1 same-file filter dropped (**restrictive** direction) | **DEAD** |
| M20 | hook call removed | **DEAD** — 15 |
| **MZ-1** | `"Class"` deleted from the address family | **DEAD** — was ALIVE at review; closed by adding `"SCOPE.Class"` and the grader |
| **MZ-3** | `"SCOPE.Class"` deleted from the address family | **DEAD** — the new entry is itself graded, so it cannot be removed unnoticed |
| M21 | hook **moved** to immediately after `walk` | **ALIVE — equivalent, with a complete enumeration.** ⚠️ The original enumeration named five passes and **omitted two**, including `stampClassLevelTransactional` — the only other pass that takes `&entities` **by pointer**, i.e. precisely the one an incomplete enumeration should not have dropped. All **seven** are now listed with what each does to the slice: `stampClassLevelTransactional` mutates `op.Properties` only, `attachImportRelationships` is handed `&entities[0]` and cannot append, `synthesizePanacheDSLEntities` is the only appender and stamps a **synthetic** source path, and the remaining four attach relationships only. Conclusion unchanged, and the reviewer confirmed it beyond the suite: moving the hook produces a byte-identical 147/147/0 on all 645 corpus files. |

No compound mutant was used as evidence anywhere. M15/M16 split M13's two conjuncts precisely because a single record violating both would leave either free to be deleted.

## Still not established

- **Cross-file targets are out of scope.** Java's one-public-type-per-file convention makes same-file a better fit than it was for Go, but a package-private sibling type and every imported type still get nothing. Binding a bare type name across files is the hazard #6976/#6369 are about.
- The **type-parameter over-fire** is known-wrong and pinned, not fixed. A real fix must thread the enclosing declaration's `type_parameters` list to the field and is *expected* to break `KnownOverFire_TypeParameterShadowedByASameFileType`.
- The **pass-1 record-set invariant is stated, not enforced**: it holds because every producer running after the hook emits a **prefixed** name no bare `type_identifier` can equal. A future producer emitting a bare-named java entity would reintroduce dangling.
- **Convention, not taken:** every other post-`walk` pass in `Extract` is wrapped in a `recover()`; `attachJavaFieldTypeRefs` is called bare. No panic path was found (nil type node and nil `Metadata` are both handled), and the coordinator scoped this follow-up to non-behavioural changes, so it is reported rather than changed.

Refs #6912

### DIFF
--- a/internal/extractors/java/field_type_refs.go
+++ b/internal/extractors/java/field_type_refs.go
@@ -1,0 +1,453 @@
+package java
+
+import (
+	"strings"
+
+	"github.com/cajasmota/grafel/internal/extractor"
+	"github.com/cajasmota/grafel/internal/treesitter/ts"
+	"github.com/cajasmota/grafel/internal/types"
+)
+
+// field_type_refs.go — the field→declared-type edge for Java (issue #6912, arm
+// E). Arm A is internal/extractors/csharp/field_type_refs.go (#6984), arm B is
+// internal/extractors/proto/field_type_refs.go (#6991), arm C is
+// internal/extractors/rust/field_type_refs.go (#7000), arm D is
+// internal/extractors/golang/field_type_refs.go (#7036).
+//
+// THE GAP. Java emits a SCOPE.Schema/field entity from two places — record
+// header components (walk's record_declaration arm) and class field
+// declarations (buildField) — and BOTH leave it a leaf on its outbound side.
+// The declared type was never lost: it is a live local at each emit site
+// (`typeName := nodeText(typeNode, …)` for a record component; buildField holds
+// the field_declaration node and already reads its `type` child through
+// buildFieldSignature). The issue body long claimed Java "does not even retain
+// the type string" and that a capture step was needed first; that claim was
+// FALSE, and the grounding pass at 663f6c50a corrected it. Java is structurally
+// the same arm as its predecessors: the datum was in hand and simply never
+// became a relationship.
+//
+// Java is one of the four languages in contributor #6726's corpus (csharp,
+// fsharp, java, protobuf), which measures SCOPE.Schema at 99.3% orphan across
+// 35,353 entities. Arms A and B closed csharp and protobuf; this closes java.
+//
+// WHERE THE ANCHOR IS. Nothing in this package emitted a declared-type edge at
+// ANY anchor before this pass — that is arm B's question, asked here rather than
+// inherited, and the answer differs from Go's. Go already shipped a
+// struct-anchored DEPENDS_ON that this arm had to sit beside; Java has no such
+// edge. What Java does have is three NEIGHBOURING populations that are easy to
+// mistake for one, and none of them is a field→type relation:
+//
+//   - javaInjectFieldTypes (java.go:390-395) emits class → REFERENCES for
+//     @Inject/@Autowired field types. The anchor is the CLASS, the target is
+//     an `ext:`/bare name rather than a structural ref, and it fires only under
+//     a DI annotation. It is untouched.
+//   - emitReferences (references.go) emits operation → REFERENCES → field for
+//     field USAGE inside a method body. Opposite direction, different fact.
+//   - internal/custom/java/hibernate.go and orm_helpers.go emit a
+//     `field_target_type` ref_kind already — but off a PARALLEL
+//     SCOPE.Component population, and internal/custom/** defaults OFF
+//     (internal/extractors/custom_gate.go). #6986 measured that lane's
+//     `Class:<Name>` address at 91.6% dangling. It is not a core producer and
+//     it does not reduce this arm's scope.
+//
+// THE ALLOW-LIST, WRITTEN ON JAVA'S OWN buildComponent RATHER THAN INHERITED.
+// Arm C's guard is `Kind != "SCOPE.Component"`, and arm D measured that exact
+// guard dropping 19% of Go's edges because a Go type_alias is a SCOPE.Schema.
+// Java's shape is the opposite: EVERY declared type this package emits as a
+// nameable target is a SCOPE.Component, because walk's single
+// class/interface/enum/record arm routes all four through buildComponent with
+// only the Subtype varying (java.go:324-355, :1374-1396). So the allow-list is
+// a SUBTYPE list, and it admits all four:
+//
+//	SCOPE.Component + class      → admitted (class Order {…})
+//	SCOPE.Component + interface  → admitted (interface Shipper {…})
+//	SCOPE.Component + enum       → admitted (enum Status {…}) — see below
+//	SCOPE.Component + record     → admitted (record Money(…) {…})
+//
+// Refused, and each for a reason that is stated at the strength it has:
+//
+//   - SCOPE.Component + "file" (#577, extractor.FileEntity) — its Name is the
+//     FILE PATH. At repo root that is a bare-looking `Order.java`, but it
+//     always carries the `.java` extension and a `type_identifier` cannot
+//     contain a dot, so this one is ALSO unreachable by name shape. Refused
+//     anyway; claimed as nothing more.
+//   - SCOPE.Component with an EMPTY subtype — synthComp (lombok.go:332) and the
+//     Panache DSL interfaces (panache.go:866-896) always set one, but a future
+//     synthesizer that does not would otherwise be admitted silently. This is
+//     the java analogue of Go's import placeholder: an unsubtyped Component is
+//     not a declared type.
+//   - SCOPE.Operation, SCOPE.Schema (field entities AND nosql_model.go:187's
+//     `schema` model node, which is named after the CLASS and so genuinely
+//     shares a name with an admitted target), SCOPE.Enum value-sets,
+//     SCOPE.ExceptionType, SCOPE.Template, config keys.
+//
+// The allow-list is a LIVE FILTER, not belt-and-braces: nosql_model.go emits
+// `SCOPE.Schema`/`schema` named `Order` in the SAME FILE as `class Order`, and
+// SCOPE.Schema/`field` records carry `Order.buyer`-shaped names. Neither is a
+// legal target. Graded by TestJavaFieldTypeRefs_NoSqlModelSchemaIsNeverATarget
+// and TestJavaFieldTypeRefs_FieldIsNeverATarget.
+//
+// THE AMBIGUITY RULE, AND WHY JAVA'S IS NOT ARM D'S. Arm D counts distinct
+// KINDS across EVERY same-file record and drops any name denoting more than
+// one graph node. Copying that verbatim here would have been WRONG IN THE
+// EXPENSIVE DIRECTION, and the reason is specific to Java:
+//
+// buildJavaEnumValueSet (enum_valueset.go:25) emits a SCOPE.Enum value-set for
+// every non-empty `enum` declaration, named IDENTICALLY to the
+// SCOPE.Component/enum that walk emits from the same node in the same file. So
+// under arm D's rule EVERY Java enum is two kinds, and EVERY enum target would
+// be refused — and a field typed by a same-file enum is one of the most
+// valuable edges this arm can produce. That is exactly the over-strictness
+// filed as #7038 against arm C, reached by a different route.
+//
+// It would also have been WRONG ON THE RESOLVER'S OWN TERMS, which is what
+// decides it. A component-space structural ref is resolved by
+// Index.lookupStructural (internal/resolve/refs.go:2982) through
+// lookupLocationKind FIRST, filtered to componentKindFamily (refs.go:2341:
+// Component/Class/View/Model and their SCOPE.* spellings). ambigLocation — the
+// kind-AGNOSTIC index arm D's rule mirrors — is only consulted AFTER that
+// lookup misses (refs.go:3004). A SCOPE.Enum sibling is not a member of
+// componentKindFamily, so it never enters the tier that decides this address,
+// and the enum's SCOPE.Component record binds uniquely. Go's type_alias
+// targets were SCOPE.Schema, outside that family, which is why arm D fell
+// through to ambigLocation and had to mirror it; Java's targets never do.
+//
+// So javaInFileTypeTargets counts distinct kinds RESTRICTED TO THE COMPONENT
+// ADDRESS FAMILY — the set lookupLocationKind will actually weigh — and drops a
+// name only when two DISTINCT family kinds carry it. Within one file two records
+// sharing a Kind are ONE graph node (graph.EntityID hashes (repo, Kind, Name,
+// SourceFile) with Subtype EXCLUDED, mirroring refs.go:1308's `existing != e.ID`
+// test), so `class Order` beside Lombok's synthesized `SCOPE.Component`
+// `Order` — or beside a second `SCOPE.Component`/`record` of the same name — is
+// one node and one ToID, and counting RECORDS instead would delete an edge that
+// binds perfectly well. That is #7038's defect and it is graded in BOTH
+// directions here:
+//
+//   - TestJavaFieldTypeRefs_SameFileEnumTypeBinds drives the REAL RESOLVER over
+//     an extracted `enum Status` + `Status status;` file and asserts the edge is
+//     rewritten to the Component's ID — a mutant restoring arm D's all-kinds
+//     count fails it. It carries a POSITIVE CONTROL in the same fixture (a
+//     class-typed field in the same file) so it cannot pass by the edge simply
+//     never being emitted.
+//   - TestJavaFieldTypeRefs_DuplicateComponentRecordsAreOneNode gives a name two
+//     admitted records of the same Kind; a mutant counting records drops it.
+//   - TestJavaFieldTypeRefs_TwoDistinctFamilyKindsAreRefused constructs the
+//     SCOPE.Component + SCOPE.Model collision the rule exists for.
+//   - TestJavaFieldTypeRefs_Unit_AnotherFilesCollisionDoesNotShadowThisFile
+//     grades the pass-1 SCOPE of the count, which fails in the same direction:
+//     the resolver's index is keyed by (file, name), so a collision in a
+//     DIFFERENT file must not refuse a target here.
+//
+// MEASURED, not argued: across the 645 .java files in the read-only corpus at
+// archigraph-corpora, this pass emits 147 edges, of which 147 bind and 0 dangle.
+// Arm D's all-kinds rule, applied unchanged, emits 129 — it drops 18 edges,
+// 12.2%, AND EVERY ONE OF THEM IS AN ENUM TARGET. The bound targets break down
+// as 125 SCOPE.Component/class, 18 /enum, 4 /interface. Arm C's
+// `Kind != "SCOPE.Component"` guard, by contrast, would have cost nothing here:
+// every Java target is a Component, which is the whole reason this arm's
+// departure is on the ambiguity rule rather than on the kind check.
+//
+// A NON-BINDING EDGE IS WORSE THAN NO EDGE. An unresolved stub is kept verbatim,
+// dangles, and is classified `bug-extractor`, so every miss is a full hit. That
+// is why this pass emits ONLY for a type DECLARED IN THIS FILE and refuses
+// everything else outright — see javaFieldTypeCandidates for how java.lang,
+// primitives, generic wrappers and imported types are all refused by the same
+// single check.
+//
+// WHAT THIS PASS IS NOT
+//
+//   - SAME FILE ONLY. Java's one-public-type-per-file convention makes this a
+//     far better fit than it was for Go — but a package-private helper type in a
+//     sibling file, and every imported type, still gets nothing. Binding a bare
+//     type name across files is the exact hazard #6976/#6369 are about, and it
+//     needs the resolver rather than the extractor. Separate arm.
+//   - QUALIFIED NAMES ARE SKIPPED WHOLE, NEVER REDUCED. See
+//     javaFieldTypeCandidates.
+//   - IT DOES NOT UNDERSTAND TYPE PARAMETERS. `class Holder<T> { T item; }`
+//     yields candidate `T`, dropped only because nothing in the file happens to
+//     be DECLARED `T`. A file that also declares `class T {}` gets a wrong edge.
+//     Known-wrong and pinned as such by
+//     TestJavaFieldTypeRefs_KnownOverFire_TypeParameterShadowedByASameFileType,
+//     a case a real fix is expected to break. Java's grammar gives no help here:
+//     a type parameter and a class reference are the same `type_identifier`
+//     node, and separating them needs the enclosing declaration's
+//     `type_parameters` list threaded to the field — the same shape Go's arm
+//     deferred.
+//   - IT DOES NOT INFLATE PAST ONE EDGE PER (field, target). A field written
+//     `Map<Order, Order>` yields the candidate `Order` twice and emits ONE edge.
+
+// javaFieldTypeRefsMetaKey is the per-field stash written at the two emit sites
+// (walk's record_declaration arm and buildField) and consumed — and deleted — by
+// attachJavaFieldTypeRefs.
+//
+// The candidates cannot become edges at the emit site. Two independent reasons,
+// either sufficient: `class Order { Customer buyer; }` may be declared BEFORE
+// `class Customer` in the same file, so the in-file declaration set is complete
+// only once walk has finished; and the target index must see records the
+// synthesizers append AFTER walk returns (Lombok builders, Panache interfaces,
+// the nosql model node), which is what makes the allow-list's nosql case
+// decidable at all.
+const javaFieldTypeRefsMetaKey = "field_type_refs"
+
+// javaFieldTypeRefsOwnerKey stashes the field's declaring type name alongside
+// the candidates. It is NOT re-derived by splitting the field entity's Name on
+// '.', because a field's leaf name and its owner are both free-form: a nested
+// type's members are emitted bare (walk's parentType does not propagate through
+// unrelated nodes), so `A.b` cannot be assumed to mean owner `A`. The emit site
+// knows the owner exactly; carrying it is cheaper than guessing it.
+const javaFieldTypeRefsOwnerKey = "field_type_refs_owner"
+
+// javaFieldTargetRefKind is the value of the `ref_kind` edge property. It
+// matches arm A's csFieldTargetRefKind, arm B's protoFieldTargetRefKind, arm C's
+// rustFieldTargetRefKind and arm D's goFieldTargetRefKind verbatim — the
+// discriminator is what makes the edge queryable as a field→declared-type edge
+// across languages, so it must be one string. There is deliberately no shared
+// constant: arm D scored that as a mutant (CZ-2) and found four independent
+// literals grade the agreement transitively, where one shared constant would be
+// a single grader.
+const javaFieldTargetRefKind = "field_target_type"
+
+// javaFieldTypeCandidates returns every bare type name written in a field's
+// declared type expression, in source order and without deduplication.
+//
+// It descends the type node and collects `type_identifier` leaves, which unwraps
+// array (`Order[]`), generic (`List<Order>` → BOTH `List` and `Order`), nested
+// generic (`Map<String, Order>` → `Map`, `String`, `Order`) and wildcard
+// (`List<? extends Order>` → `List`, `Order`) syntax for free, with no wrapper
+// list to keep in sync.
+//
+// A JAVA PRIMITIVE NEEDS NO BLOCKLIST, AND — UNLIKE GO — CANNOT NEED ONE. Go's
+// `int` is a shadowable universe-scope identifier that tree-sitter-go emits as a
+// plain `type_identifier`, so arm D had to reach the primitive through its
+// in-file declaration check and pin the shadowing case. Java's primitives are
+// KEYWORDS: the grammar gives `int`/`long`/`short`/`byte`/`char` the node type
+// `integral_type`, `float`/`double` `floating_point_type`, and `boolean`
+// `boolean_type`. None is a `type_identifier`, none can be redeclared, and
+// `long[]` descends to `integral_type` + `dimensions` and yields nothing at all.
+// So a primitive is not merely refused, it is never a candidate — verified
+// against the grammar rather than assumed, and pinned by
+// TestJavaFieldTypeRefs_PrimitiveFieldsProduceNoEdge, which enumerates all eight
+// against an independent literal list.
+//
+// THE BOXED TYPES, java.lang AND EVERY IMPORTED TYPE are ordinary
+// `type_identifier`s and ARE collected — then refused by the single in-file
+// declaration check in javaInFileTypeTargets, which is the only gate this pass
+// has and the only one it needs. `String`, `Integer`, `List`, `HttpClient` and
+// `com.acme.Widget` are none of them declared in the file, so none produces an
+// edge. Writing a name blocklist instead would be actively wrong: a file that
+// declares its own `class String` means THAT type in field position, and Java
+// permits it (the unqualified name resolves to the local declaration, shadowing
+// java.lang). Pinned in both directions by that primitive test and by
+// TestJavaFieldTypeRefs_BoxedAndJavaLangTypesProduceNoEdge.
+//
+// ONE node kind is skipped WHOLE, without descending:
+//
+//   - `scoped_type_identifier` — `com.example.Order`, `java.util.List`,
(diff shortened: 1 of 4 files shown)


cajasmota:
# APPROVE — independent adversarial review of #7039 (arm E, java) at `678d07a9a` vs `main` `a16fd6d61`

Own worktree, own **private `GOCACHE`** (`…/rev-6912java-v2p/gocache`) exported for every `go` invocation — no phantom build failures, and no mutant scored as "not applicable". Backup of each mutated file taken from **this** tree, restored by `cp` and shasum-verified; `git status --porcelain` empty at the end. Every mutant edit's landing proved by an exact occurrence count before and after.

Nothing blocks. Six prose/enumeration defects and one ungraded-but-unreachable conjunct, all priced below.

---

## The departure: verified, and I think it is right

**Every corpus number reproduces independently.** I extracted all 645 `.java` files under `archigraph-corpora` in-process (single test, `GOMAXPROCS=3`), ran `resolve.BuildIndex` + `ReferencesEmbedded` per file, and recomputed arm D's rule from the same record sets rather than by mutating source:

```
.java files: 645
per repo: kafka=471 kafka-streams-examples=115 spring-petclinic=47 awesome-compose=11 ratpack-example-books=1
edges=147 bound=147 dangling=0
bound target kinds: SCOPE.Component/class=125 SCOPE.Component/enum=18 SCOPE.Component/interface=4
arm-D all-kinds rule would drop: 18 (enum=18) leaving 129
```

147 / 147 / 0, the 125+18+4 breakdown, and **18 dropped, all enums, 129 remaining (12.24%)** — all confirmed.

**The resolver argument holds at source, and the premise is observed rather than read.** `lookupStructural` calls `lookupLocationKind(filePath, tail, structuralKindFamilies("component"))` at `refs.go:2982`, and `ambigLocation` is only reached at `:3004`; `componentKindFamily` (`:2982`-adjacent decl) is `{Component, Class, View, Model, SCOPE.Component, SCOPE.View, SCOPE.Model}` — no `SCOPE.Enum`. `graph.EntityID` (`graph/graph.go:259`) hashes `(repo, kind, name, sourceFile)` with Subtype excluded, so same-Kind duplicates are one node and `base.set(kind, id)` never blanks. I dumped the main fixture's records and **the `SCOPE.Enum/enum Status` competitor really is there beside `SCOPE.Component/enum Status`** in the same file, and `ResolvesToEntityIDs` binds `Order.state => SCOPE.Component:Status [enum]` through the real resolver. The departure is graded end-to-end, not just at unit level.

**Judgement on three arms, three rules: principled variation of ONE rule that has never been written down — not drift.** The invariant every arm actually needs is *"refuse a name whose graph node is not uniquely determined by the tier that resolves the address you emit."* Arm C counts only within its admitted set (wrong — #7038). Arm D counts all kinds, which is right **for go** because go's `type_alias` targets are `SCOPE.Schema`, outside `componentKindFamily`, so go's refs genuinely fall through to the kind-agnostic `ambigLocation`. Java's targets are always in the family, so the family-restricted count is the faithful mirror — and copying arm D would cost 12.2%, all enums, as measured above. So the three rules are three instantiations of one principle against three resolver tiers. What is drifting is not the code, it is that the principle exists only as per-arm prose: five more arms are queued (swift/php/cpp/kotlin/scala, plus fsharp), and swift and php both have parallel value-set / non-Component-kind shapes that will need this reasoning again. **Recommend a one-paragraph statement of the rule on #6912 (or a small shared doc comment) that each remaining arm derives its count from, plus a re-check that arm C's within-set count is corrected under it** — otherwise arm F copies whichever arm it read last.

## Also-pushed-on items

1. **Allow-list is a subtype list, and it is exhaustive.** `java.go:326` is the literal `subtype := "class"` inside the shared `class_declaration/interface_declaration/enum_declaration/record_declaration` arm — confirmed. The full `SCOPE.Component` subtype universe reaching a java file's records is `{class, interface, enum, record, file}`: walk's four, `extractor.FileEntity` (`extractor.go:348`, `"file"`), `synthComp` (only call site `lombok.go:438`, subtype `"class"`), panache (`"interface"`). The four are exhaustive and the two refusals are complete. Arm C's `Kind != "SCOPE.Component"` guard indeed costs nothing here — it is in fact already the pass-2 filter, and all 147 bound targets are Components.
2. **Generics / qualified names — verified empirically, not read.** With a probe fixture: `java.util.List<Customer>` → `Customer`; `java.util.Map<java.lang.String, Customer>` → `Customer`; `java.util.List<java.util.List<Customer>>` → `Customer`; `Outer.Inner<Customer>` → `Customer`; and `List<Customer>` in a file that declares `class List` → **both** `List` and `Customer`. The `type_arguments`-is-a-sibling claim is exactly right. Primitives: the eight-primitive enumeration passes with its `class Int` positive control and no edge, consistent with the grammar's `integral_type`/`floating_point_type`/`boolean_type` node types — never candidates, confirmed.
3. **M2 / M13 / M15 / M16 re-scored by me — all DEAD, kills real.** Each edit landed exactly once. M2 (pass-2 Kind check deleted) fails `Unit_NonComponentKindWithAnAdmittedSubtypeIsRefused`. M15 (Kind conjunct alone) fails **only** subtest `wrong_kind,_field_subtype` while `schema_kind,_wrong_subtype` PASSES; M16 (Subtype conjunct alone) fails **only** `schema_kind,_wrong_subtype` while the other PASSES. Complementary, so neither conjunct masks the other — this is a correct split, not a cosmetic one. The fixtures are not manufactured to force a kill: each carries an admitted-Component positive control in the same record set, so a rule that emitted nothing fails them too.
4. **M21 is genuinely equivalent, and I checked it beyond the suite.** I moved the hook to immediately after `walk`: whole java package green **and** the 645-file corpus output identical (147/147/0, same 125/18/4 breakdown). But **the equivalence enumeration in the comment is incomplete**: five passes are named, and **two between `walk` and the hook are omitted** — `stampClassLevelTransactional` (`java.go:83`, which does take `*[]types.EntityRecord`) and `attachImportRelationships`. I closed both myself: the former only mutates `op.Properties` through `findJavaOp`, the latter is handed `&entities[0]` and cannot append; the only `append` to `entities` between the two points is `synthesizePanacheDSLEntities`, which stamps `panacheDSLSyntheticSourceFile` (`panache.go:857`). Conclusion stands; **add the two missing names**, because an equivalence argument is only as good as its enumeration and this one omitted the only other pass taking the slice by pointer.
5. **Corpus narrowness — what 147/147 does and does not establish.** Per-repo split above: three real codebases contribute (kafka + kafka-streams-examples + spring-petclinic), not two. It establishes that the structural address and the same-file gate produce **zero** dangling stubs on real java at ~600-file scale, and that the enum decision is worth 18 edges. It does not establish **recall** — there is no denominator anywhere in this PR (no count of fields whose declared type *is* same-file-declared and got nothing), so "147" is unaudited in the under-firing direction on the corpus; over-firing is graded only by fixtures; `/record` targets are 0; and nothing here transfers to #6726's corpus.
6. **A mutant nobody ran — MZ-1: `"Class": true` deleted from `javaComponentAddressFamily` → ALIVE. Verdict: distinguishable and ungraded, not production-reachable today.** No java producer emits a bare-`Class`-kinded entity, so nothing distinguishes it from java source. It exposes a real asymmetry the comment mis-states, though: it claims the map carries "both the bare and SCOPE.-prefixed spellings", but **`SCOPE.Class` is absent**, while `BuildIndex` dual-keys a `SCOPE.Class` entity under *both* `"SCOPE.Class"` and the trimmed `"Class"` (`refs.go:1265-1296`) and `componentKindFamily` contains `"Class"`. So a same-file `SCOPE.Class` named as an admitted Component would blank the family match (`uniqueMatchInFamily` → `("", false)`), fall through to `ambigLocation`, and **dangle**, while this pass would emit it. Unreachable today; the mirror invariant the comment itself states is the thing worth keeping true. **Price: 1 line (`"SCOPE.Class": true`) plus one row in the existing unit table, ~6 lines.**
   Second mutant, MZ-2: `fieldName = strings.TrimPrefix(r.Name, owner+".")` → `r.Name` → **DEAD** (`EdgeProperties`).

## Prose that asserts what no test observes (all non-blocking, all cheap)

- **`field_type_refs.go:126` cites `TestJavaFieldTypeRefs_SameFileEnumTypeBinds`, which does not exist** — `grep -rn` across `internal/` finds it only in that comment. It is described in detail (real resolver, positive control) as the grader for this arm's headline departure. The substance *is* graded (`EdgeSet` + `ResolvesToEntityIDs` + `Unit_EnumValueSetDoesNotShadowItsComponent`), so there is no coverage gap — but a future reader grepping the named grader finds nothing. Three other citations drop the `_Unit_` infix (`TwoDistinctFamilyKindsAreRefused`, `FieldIsNeverATarget`, `DuplicateComponentRecordsAreOneNode`).
- **The Lombok example is fabricated.** The header says "`class Order` beside Lombok's synthesized `SCOPE.Component` `Order` … is one node". `synthComp`'s only call site (`lombok.go:438`) passes `builderClass`, i.e. `<Class>Builder` — never the class's own name — so that production shape does not exist. The test file's header (l.26-27) then promises "the Lombok duplicate-Component case needs a `@Builder`. Each is below with its own comment" and **there is no `@Builder` fixture below**. The duplicate-same-Kind case is graded only by a hand-built record set. Both statements should be corrected to say the case is defensive and unreachable from java source today — precisely the honesty the M2/M13 write-ups already show.
- **The ratchet enumeration omits a fixture, and the omitted one is the only interesting one.** The body says "all **6** `java-spring-mini` `.java` files declare exactly one type each … there was nowhere it could have moved." `internal/quality/golden/java-quartz-mini` also has `.java` files, and `src/jobs/SendEmailJob.java` is **the only multi-type java file in the entire golden set** — exactly the shape the claim turns on. The conclusion survives: I extracted **all 8** golden `.java` files and got **0** `field_target_type` edges (both quartz files declare no fields at all). Same defect class as arm D's "go-chi-mini is the only golden fixture with `.go` files"; the fix is to enumerate the *set*, not one repo in it.
- **`cmd/grafel` digest: right to decline, slightly wrong reason.** `semanticDigest6118` keys relationships as `(fromTuple, Kind, toTuple)` with properties discarded, so it is blind to `ref_kind` — but **not** blind to a *new* `REFERENCES` edge appearing. It could not move here because the fixture's only java file, `java/OrdersController.java`, declares one class and **no fields**. Say that rather than "could not have moved for this change".

## Two smaller things

- **Recall ceiling nobody stated.** 147 is bounded by java's *field-entity* population, which has two pre-existing holes I hit while probing: a multi-declarator field emits only the first name (`class M { Customer a, b; }` → `SCOPE.Schema/field M.a` only, no `M.b`), and an **interface constant emits no field entity at all** (`interface Consts { Customer DEFAULT = null; }` → no field record, no edge). Neither is introduced here and neither is this arm's job, but they cap the arm's ceiling and are worth a follow-up. (`@interface` annotation types also get no Component, so an in-file annotation type can never be a target — negligible.)
- **Convention nit.** Every other post-`walk` pass in `Extract` is wrapped `func(){ defer func(){ _ = recover() }(); … }()`; `attachJavaFieldTypeRefs` is called bare, so a nil-deref there would lose **every** entity for the file rather than one pass. I found no panic path (nil type node and nil `Metadata` are both handled), so this is convention, not a bug — ~3 lines if you want it.

## Suites run (private GOCACHE, `-count=1`)

`./internal/extractors/java/...` green (baseline and after every restore); `./internal/quality/...`, `./internal/quality/{analytics,audit,fitness}`, `./internal/resolve/...`, `./internal/extractor/...` all green. I did not run `./cmd/grafel/` (author's isolated-env green + the digest analysis above make it uninformative for this change). Corpus harness read-only, 3 cores, no `grafel` subcommand, nothing written to `~/.grafel`.

**Verdict: APPROVE.** The one departure this PR rests on is correct, measured, and graded through the real resolver; the mutant table survives independent re-scoring including the two first-pass-ALIVE conjuncts and the split pair; the one ALIVE mutant is genuinely equivalent, now demonstrated on 645 files and not only under the suite. The findings are documentation accuracy and one unreachable mirror hole.

cajasmota:
## Coordinator mutants on `3609dfca0` — **DA-1 and DA-2 both DEAD, by different tests**

Own detached worktree at `/private/tmp/.../coord-mut-7039-w6k/wt` (absolute), **private `GOCACHE`** in that lane, backup taken from the post-edit tree. Green baseline immediately before. Restored by `cp` from a SHA-256-verified backup (`fd698791…` both ways); `git diff HEAD` and `git status --porcelain` empty afterwards.

22 mutants have been scored between you and the reviewer, and they cover the ambiguity family scoping, the allow-list, generics, primitives, the address-family mirror, and the hook placement. What none of them touches is the constraint that every arm of this P1 is built on and that no arm may break: **same-file targets only.** Cross-file binding of a bare type name is the #6976/#6369 hazard, so this is the one conjunct where a silent regression is worst.

There are **two** same-file conjuncts in `field_type_refs.go`, at `:371` and `:384` — one on the ambiguity family scan, one on the target scan. Grading one is exactly what would make the other look covered, which is the most repeated defect shape on this board.

### DA-1 — drop the same-file conjunct from the **family scan** (`:371`)

`r.SourceFile != filePath || r.Name == "" || !javaComponentAddressFamily[r.Kind]` → drop the first conjunct. Occurrence count asserted 2 → 1, `go vet` clean.

**DEAD:**
```
--- FAIL: TestJavaFieldTypeRefs_Unit_AnotherFilesCollisionDoesNotShadowThisFile
```

### DA-2 — drop the same-file conjunct from the **target scan** (`:384`)

`r.SourceFile != filePath || r.Name == "" || r.Kind != "SCOPE.Component"` → drop the first conjunct. Occurrence asserted 1 after the edit.

**DEAD:**
```
--- FAIL: TestJavaFieldTypeRefs_Unit_OtherFileRecordsAreNeverTargets
```

### The point: different tests, different properties

The two conjuncts do genuinely different jobs and each has its own grader:

- `:371` — a **collision in another file must not shadow this file's** resolution (a cross-file record wrongly *suppressing* a legitimate edge)
- `:384` — **another file's records are never targets** (a cross-file record wrongly *becoming* a target)

Those are opposite failure directions — one costs recall, the other produces a wrong binding — and they are graded separately rather than by one test that happens to fail for both. The test names say which is which, which is what makes them survive the next refactor.

### On the round

The two invented claims are gone, and the way you handled the Lombok one is the part worth naming. **The example was fabricated, but you did not delete it** — you found the collision that *is* reachable (`@Builder class Order` synthesising `OrderBuilder` beside a hand-declared `class OrderBuilder`, two `SCOPE.Component/class` records, one node), **probed it before writing it**, and turned it into a test that grades the count-kinds-not-records rule **from java source** rather than from a hand-built record set. M4 now dies to two tests instead of one, and #7038's direction is graded on a production shape. Replacing an invention with the real thing it gestured at is strictly better than removing it.

Three more things:

- **You corrected the reviewer, and you were right.** `cmd/grafel` has java fixtures across several test files; the digest fixture specifically (`java/OrdersController.java`) is one class with **no fields** — so the honest reason is *nothing to move*, not *the instrument is blind*. **The digest is not blind to a new `REFERENCES` edge** — relationships are hashed as `R <from> -[Kind]-> <to>`; only properties are unhashed. I verified that from source myself, because I had relayed the wrong framing into two other briefs. Your "wrong in the flattering direction" self-assessment was the right call and it caught my error, not just yours.
- **MZ-1 closed rather than recorded**, and the framing is better than the fix: the address family is not a copy of `componentKindFamily`, it is **that slice closed under the trim alias**, and the 8 entries are now exactly `{K : K ∈ family ∨ trim(K) ∈ family}`. MZ-3 makes the new entry unremovable too.
- **The M21 enumeration named 5 passes and omitted 2** — including `stampClassLevelTransactional`, the only *other* pass taking `&entities` by pointer, i.e. the one most able to invalidate the equivalence. An equivalence argument resting on an incomplete enumeration is worthless; all 7 are now listed.

### Merge status

This is **held, not blocked**. v0.3.3 is publishing and `main` is frozen at the tagged SHA; merging would move it off the only commit ever validated as a sequence. It goes in as soon as the release lands.

PR #7040, opened 2026-09-08
feat(#6912): F# record fields carried their declared type only as a string property — emit the field→type REFERENCES edge, same-file targets only

cajasmota:
Arm E of #6912 — the field→declared-type `REFERENCES` edge for F#, `ref_kind=field_target_type`, field-anchored, **same-file targets only**. Arms A (C#/#6984), B (proto/#6991), C (rust/#7000), D (go/#7036) precede it.

F# is one of the four languages in contributor **#6726**'s corpus (csharp / fsharp / java / protobuf) and `du_record_members.go:77` mints F# record members as `Kind: "SCOPE.Schema"` — so they are part of the exact 99.3%-orphan population that issue reports. F# was never assessed by #6912's grounding pass, which covered php/cpp/swift/java/kotlin/scala/go/ruby.

---

## 1. What `member_type` actually contains — probed, and it changes the design

`parseRecordFields` (`du_record_members.go:140`) takes everything after the first `:` in the field segment and only `TrimSpace`s it. Probed output, pinned against literals in `TestFSharpFieldTypeRefs_MemberTypeIsVerbatimSourceText`:

| written | `member_type` |
|---|---|
| `Plain: Acct` | `Acct` |
| `Opt: Acct option` | `Acct option` |
| `Lst: Acct list` | `Acct list` |
| `Deep: Acct list option` | `Acct list option` |
| `Mp: Map<string, Acct>` | `Map<string, Acct>` |
| `Arr: Acct[]` | `Acct[]` |
| `Tup: int * Acct` | `int * Acct` |
| `Fn: int -> Acct` | `int -> Acct` |
| `Qual: Domain.Acct` | `Domain.Acct` |
| `Gen: 'T` | `'T` |
| `Mut: mutable Acct` | `mutable Acct` |

So it is the declared type **as written, character for character** — not swift's lossy first-pre-order-identifier, and not absent. **It is the one shipped field-type property on this issue that loses nothing**, which is why this arm reads the property where arm C stashed AST candidates.

**Arm C's alternative is not available here anyway:** the F# extractor is regex- and text-driven end to end (`moduleRE`/`typeRE`/`memberRE` + `extractIndentBody`); there is no tree-sitter grammar for F# in this tree. "Stash from the AST" is not a choice declined, it is a thing that does not exist.

## 2. Postfix generics — why the pass tokenises instead of unwrapping

`Acct list` / `Acct option` put the type constructor **after** its argument. No shipped arm has faced that, and a wrapper-list unwrapper would have to enumerate FSharp.Core and would still miss a user-defined postfix constructor.

So the pass **does not unwrap at all.** It scans the type expression into bare identifiers and judges each independently against the same-file declaration set — arm D's rule, reached by a different route. It works because every named type in an F# type expression is a bare identifier, whatever composes them: `<>`, postfix juxtaposition, `[]`, `*`, `->`, units of measure. `option`, `list`, `seq`, `Map` fall out because the file does not declare them — **no blocklist**.

`TestFSharpFieldTypeRefs_CandidateScannerEnumeratesTheShapeSpace` enumerates 28 rows rather than hand-picking three.

### Qualified names: skipped **whole**, never reduced

`Domain.Acct` is scanned as **one token** (`.` is an identifier-continuation char precisely so the whole qualified name arrives together) and dropped. Taking the trailing segment is the wrong binding arm D found in go's `resolveTypeReferences`. The fixture writes `Domain.Acct` in a file that **also declares its own `Acct`**, with a positive control, so the two behaviours give different output. Three depths tested (`A.B`, `A.B.C`, `System.Collections.Generic.List<Acct>`), because one dot cannot distinguish "dropped whole" from "the scanner stopped at the first dot".

### Generic parameters: closed, not pinned as a known over-fire

`'T` is consumed apostrophe-first so it can never re-enter the scanner as bare `T`. Arm D **shipped** this as a known over-fire because Go's `T` is indistinguishable from a type name; F#'s leading apostrophe makes it distinguishable. The fixture declares `type T` so the row grades the consumption rather than the absence of a target.

## 3. The target allow-list, and its live kill

Enumerated from this package's own emit sites, keyed `Kind/Subtype`:

**Admitted** — every subtype `classifyTypeSubtype` can return, plus `detectCEBuilder`'s and `applyElmishFeliz`'s re-labels:
`SCOPE.Component/{record, discriminated_union, interface, class, struct, alias, type, computation_builder}`, `SCOPE.Model/elmish_model`, `SCOPE.Event/elmish_msg`.

**Refused, three distinct hazards:**

- **`SCOPE.Component/import` — THE LIVE KILL, and sharper than Go's.** `buildImportEntities` (`extractor.go:678`) names each `open` placeholder `importDisplayName(mod)` — the **LAST SEGMENT** — so `open Acme.Acct` puts a `SCOPE.Component` literally named `Acct` in the file. `resolve/refs.go` filters import placeholders out of `indexByName` (`:1611`) but **not** out of `byLocation`/`byLocationKind` (`:1271-1317`), and `byLocation` is exactly the fallback a component-space structural ref lands in. Without the refusal a field typed `Acct` **binds to the import placeholder** — a *wrong binding*, not a dangle, so nothing downstream reports it. Graded by a fixture that declares **no** type of that name, so the ambiguity rule cannot mask the allow-list.
- **`SCOPE.Component/module` and `/namespace`** — also bare-named. Costs nothing today (same Kind + Name + file ⇒ one node), but a module with no companion type is a bare name with nothing behind it.
- **`SCOPE.Operation` / `SCOPE.Pattern` / `SCOPE.UIComponent` / `Component/file` / `Schema/{field,du_case,active_pattern_case}`** — not types. The three `SCOPE.Schema` subtypes are additionally unreachable by name shape (dotted names; a candidate never contains a dot), claimed as nothing more than belt.

Arm C's `Kind != "SCOPE.Component"` guard would drop `SCOPE.Model` and `SCOPE.Event` — i.e. `State: Model`, the commonest declared-type relation in an Elmish app. On the realistic probe below, 3 of 19 edges (15.8%) target a re-kinded record.

### An honest correction: the allow-list rows and the call placement are **mutually masking**

The first draft of the header claimed the after-`applyElmishFeliz` placement was load-bearing. Measured, it is not:

| mutant | verdict |
|---|---|
| drop the two Elmish rows, keep the placement | **DEAD** (1 test) |
| move the call before `applyElmishFeliz`, keep the rows | **ALIVE** |
| both | **ALIVE** |

Before the re-kind, `Model` is `SCOPE.Component/record` and `Msg` is `SCOPE.Component/discriminated_union` — both already admitted — so the two arrangements produce identical output. **The rows' kill holds only while the placement is held constant, and the placement is not independently graded.** Recorded in the header, because a pair of mutually-masking guards grades neither and the failure mode is claiming both are load-bearing.

The after-placement is kept on an argument that does **not** depend on the Elmish rows: read at the end, the pass sees every record's **final** Kind and Subtype, so a future re-kind cannot silently invalidate the allow-list or the ambiguity rule — an invariant `applyElmishFeliz` already breaks for the earlier placement.

## 4. Ambiguity counts KINDS, not records — both directions graded

`graph.EntityID` hashes `(repo, Kind, Name, SourceFile)` with **Subtype excluded**, and `resolve/refs.go:1308` flips `(file, name)` ambiguous on `existing != e.ID`. So within a file, two records collide into **one node** exactly when they share a Kind.

- **No-rule mutant** → `let Acct = 42` beside `type Acct` is two Kinds, two nodes, `statusAmbiguous`, a guaranteed dangle. DEAD, with a positive control (`Ok: Money`) in the *same* fixture so it cannot pass by the pass being dead.
- **Record-count mutant** → `module Acct` heading an `Acct.fs` that also declares `type Acct` is two records, **one node**, an edge that binds. DEAD.

(The nested companion-module form `module Acct =` produces no second record at all: `moduleRE` anchors on `\s*$` and never matches a header ending in `=`. Pre-existing gap, noted because it is why the fixture uses the top-level form.)

**#7038** is the same over-strictness in the already-shipped rust arm.

## 5. DU decision: **records only**, deliberately

`du_record_members.go` emits DU cases through the same function, and `| Placed of Acct` does put a type name in `member_type`. Refused on `Subtype`, for two independently sufficient reasons:

- **Different relation.** A record field *is* a member with a declared type. A DU case is a **constructor** — `Status.Placed of Acct` means a `Status` may be *built from* an `Acct`, in one alternative only. Emitting it under `ref_kind=field_target_type` would make one query return two different facts (#6902's family). If it is worth having it wants its own `ref_kind` and its own arm.
- **Lossier datum.** `splitDUCase` takes everything after ` of ` verbatim, so `| Named of who: Acct * qty: int` arrives **with the field labels still in it** (probed). Harmless under the declaration gate, but it is inference on a parse never built to carry types.

Graded, not asserted: the fixture's payload type **is** declared in the file and **is** admitted, so removing the `Subtype` guard emits and the row fails.

## 6. Varied vs held-constant axes

**Varied:** type-expression shape (bare / postfix `option`,`list` / prefix `Map<,>` / array / tuple / function / qualified / generic param / nested `list option`); target Kind+Subtype (record, DU, class, interface, alias, computation_builder, `SCOPE.Model`, `SCOPE.Event`); the *rival* record sharing the name (import placeholder, module, `SCOPE.Operation`, active pattern); where the target is declared (same file vs sibling file); member Subtype (field vs du_case vs active_pattern_case); uppercase vs lowercase type names.

**Held constant:** one repo, one language, the `.fs` extension (`.fsi` differs only in hierarchy emission, untouched here); the edge Kind (`REFERENCES`) and the anchor (the field record).

## 7. Mutants — 18 scored, per conjunct, permissive direction first

| # | mutation | verdict |
|---|---|---|
| M1 | allow-list removed entirely | **DEAD** (2) |
| M2 | arm C's `Kind != "SCOPE.Component"` verbatim | **DEAD** (2) |
| M2b | only the two Elmish rows removed | **DEAD** (1) |
| M3 | ambiguity rule removed | **DEAD** (1) |
| M4 | ambiguity counts records, not kinds | **DEAD** (1) |
| M5 | qualified name reduced to its trailing segment | **DEAD** (2) |
| M6 | `'T` skips only the apostrophe (re-enters as `T`) | **DEAD** (2) |
| M7 | records-only `Subtype` guard removed | **DEAD** (1) |
| M8 | self-reference guard neutered | **DEAD** (1) |
| M9 | per-field ToID dedup removed | **DEAD** (2) |
| M10 | `ref_kind` → `field_target_types` (near-miss) | **DEAD** (16) |
| M12 | explicit `FromID` (Hibernate's shape) | **DEAD** (1) |
| M13 | `SourceFile` conjunct removed from pass 1 (`nameKinds`) | **DEAD** (1) |
| M14 | `SourceFile` conjunct removed from pass 2 (admission) | **DEAD** (1) |
| M15 | `SourceFile` conjunct removed from the field loop | **DEAD** (1) |
| M11 | call moved before `PrependFileCarrier` | **ALIVE — equivalent** |
| M11b | call moved before `applyElmishFeliz` | **ALIVE — equivalent** |
| M16 | `declared == ""` early continue removed | **ALIVE — equivalent** |
| M17 | `len(targets) == 0` early return removed | **ALIVE — equivalent** |

Each conjunct was scored **separately**: the three `SourceFile` sites are three mutants (M13/M14/M15), not one; the allow-list and the ambiguity rule are M1/M2/M2b and M3/M4. No compound was offered as evidence for a half — and the one pair that looked compound (M2b/M11b) is reported above as *mutually masking* rather than as two kills.

**Why the four ALIVE ones are equivalent, with reasons rather than "too expensive":**
- **M11** — `FileEntity` names the carrier `file.Path` (`extractor/extractor.go:344`) and every F# path contains a `.`, which no bare candidate token ever does. Unkillable; the placement is defence-in-depth that removes the need for the invariant.
- **M11b** — measured above; the two arrangements produce identical output.
- **M16 / M17** — pure short-circuits. An empty `member_type` yields zero candidates; an empty target map misses every lookup.

M13/M14/M15 are unreachable through `Extract` (one file per call), so `field_type_refs_scope_6912_test.go` calls the functions directly with the multi-file input the guards exist for — the whole cost of grading them is one file.

**Every mutant edit was proved to land** with an exact occurrence count before the run, and the tree was restored by `cp` from a shasum-verified backup taken from *this* tree, with `git diff HEAD` empty afterwards.

## 8. Measurement — and the corpus has no F# at all

**`/Users/jorgecajas/Projects/archigraph-corpora` contains ZERO `.fs`/`.fsi`/`.fsx`/`.fsproj` files.** So this arm **produced nothing there** — a fact about the corpus, not about the producer. Stating it plainly rather than reporting a zero as success, and it is why the binding gate is a test rather than a corpus bind-rate.

**The probe is sufficient to ship the arm and NOT sufficient to claim yield or orphan clearance for #6726.** It establishes that the address dialect binds for the shapes its author imagined; it says nothing about yield on real F# or about clearing #6726's `Schema` orphan population. Review found two defects living in shapes it omits (an import placeholder colliding with a re-kinded `Model`; a `let` shadowing a type). Since #6726 is expressed in orphan rates and four of its six headline numbers were already defects in our own measurement, "19/19 bound" must not be read as a yield figure for it.

Substitute measurement, labelled as synthetic: a **4-file realistic probe** (Elmish client with `Model`/`Msg`/`Filter`, a shared domain module with records + a DU + an alias, a server wiring module with a class, plus the repo's own `testdata/elmish_counter.fs`) — 92 entities, **39 record-field entities** — driven through the **production resolver** (`BuildIndex` → `ReferencesEmbedded`):

**19 edges, 19 BOUND, 0 dangling.**

Also refused, correctly, on that probe: `Map<string, string>` (no in-file target), `Map<string, Shared.Order>` (qualified skipped), `Shared.Customer` (qualified + cross-file). `TestFSharpFieldTypeRefs_ResolverBindsEveryEdge` carries the same gate in-suite over a two-file fixture, with a **positive control on the address dialect** — the bare name is ambiguous there, which is why a bare-name ToID (`#6986`: 91.6% dangling) is not what we emit.

## 9. Gates

- `go test -count=1` green on `./internal/extractors/fsharp/...`, `./internal/resolve/...`, `./internal/quality/...` (incl. `analytics`, `audit`, `fitness`), `./internal/extractor/...`, `./internal/graph/...`.
- **`go test -count=1 ./cmd/grafel/`** green (148s) in a **fresh** `GRAFEL_HOME` + `GRAFEL_DAEMON_ROOT`, with the parity tests verified **run, not skipped** (`-run Parity`, 15 PASS; the 3 Django skips are pre-existing and unrelated). Not cited as coverage of this change — that digest is a same-binary parity set keyed `(from,to,kind)` with **properties discarded**, so it cannot see `ref_kind`.
- `scripts/quality/run.sh --runs 1 --ratchet` — 38 fixtures held, and the real check passes: **`git status --porcelain` shows only this arm's own files.** `baseline.json` is untouched, and that is enumerable rather than lucky: `find internal/quality/golden -name '*.fs*'` returns **0** — there is no F# golden fixture, so no baseline could move. No `--update-baseline`, no hand-edit needed.
- **`ReferencesPerFunction` will inflate** (these edges are `REFERENCES` and a `SCOPE.Schema/field` is not function-like). Left alone per **#7037**, deliberately, so this arm's numbers stay comparable with arms A–D.

## 10. Environment finding worth recording

Mid-run a build failed with `no space left on device`; the data volume was genuinely at **1.6 GiB free** (the shared scratchpad is ~67 GB). I ran `go clean -cache`, which reclaimed ~190 GB — but it ran for many minutes **concurrently with other agents' builds**, and that is what produced the phantom `build failed` / missing-cache-object errors a parallel agent reported. Not disk pressure at the time they saw it: **my `go clean`**. Every `go` invocation from that point on, including **every mutant in the table above**, ran under a **private `GOCACHE`** in this lane's scratchpad, and every suite in §9 was re-run under it. No verdict in this PR rests on a shared-cache build.

Refs #6912

### DIFF
--- a/internal/extractors/fsharp/extractor.go
+++ b/internal/extractors/fsharp/extractor.go
@@ -534,7 +534,32 @@ func extractFSharp(src, filePath string) []types.EntityRecord {
 	// `Core.fs` declaring `module Core.fs` already emits a record named
 	// exactly the path, and graph.EntityID does not hash Subtype, so a second
 	// SCOPE.Component would land under the module record's id (#6369/#6480).
-	return extractor.PrependFileCarrier(filePath, "fsharp", entities)
+	// #6912 arm E: the field→declared-type REFERENCES edge. Placement is
+	// load-bearing in BOTH directions and the call therefore wraps the carrier
+	// rather than preceding it:
+	//
+	//   - It runs AFTER applyElmishFeliz, which RE-KINDS the `Model` record to
+	//     SCOPE.Model and the `Msg` DU to SCOPE.Event. This is NOT because the
+	//     earlier placement would lose those edges — measured, it would not:
+	//     before the re-kind both records are already-admitted
+	//     SCOPE.Component subtypes, and moving the call produces identical
+	//     output. The reason is that reading at the END means the allow-list
+	//     and the ambiguity rule see every record's FINAL Kind and Subtype, so
+	//     a future pass that re-kinds something cannot silently invalidate
+	//     either. See the fsharpTypeDeclKinds comment for the three-row
+	//     experiment behind that claim.
+	//   - It must run after EVERY producer that can add a same-file record,
+	//     because its ambiguity rule counts the graph nodes a name denotes in
+	//     this file and a collision it cannot see reaches the resolver as a
+	//     dangling edge. Passing the post-PrependFileCarrier slice makes that
+	//     unconditional rather than resting on the carrier's Name always being
+	//     a path. That half is defence-in-depth and is scored as such: moving
+	//     the call to BEFORE PrependFileCarrier is an EQUIVALENT mutant (ALIVE,
+	//     and unkillable) because FileEntity names the carrier file.Path
+	//     (extractor/extractor.go:344) and every F# path contains a `.`, which
+	//     no bare candidate token ever does.
+	return attachFSharpFieldTypeRefs(
+		extractor.PrependFileCarrier(filePath, "fsharp", entities), filePath)
 }
 
 // classifyTypeSubtype determines the F# type subtype from the declaration context.
--- a/internal/extractors/fsharp/extractor.go
+++ b/internal/extractors/fsharp/extractor.go
@@ -539,15 +539,18 @@ func extractFSharp(src, filePath string) []types.EntityRecord {
 	// rather than preceding it:
 	//
 	//   - It runs AFTER applyElmishFeliz, which RE-KINDS the `Model` record to
-	//     SCOPE.Model and the `Msg` DU to SCOPE.Event. This is NOT because the
-	//     earlier placement would lose those edges — measured, it would not:
-	//     before the re-kind both records are already-admitted
-	//     SCOPE.Component subtypes, and moving the call produces identical
-	//     output. The reason is that reading at the END means the allow-list
-	//     and the ambiguity rule see every record's FINAL Kind and Subtype, so
-	//     a future pass that re-kinds something cannot silently invalidate
-	//     either. See the fsharpTypeDeclKinds comment for the three-row
-	//     experiment behind that claim.
+	//     SCOPE.Model and the `Msg` DU to SCOPE.Event — and this ordering is
+	//     LOAD-BEARING, via the AMBIGUITY RULE rather than the allow-list.
+	//     `open Foo.Model` beside an Elmish `type Model` is the distinguishing
+	//     input: read after the re-kind the name denotes two kinds and the pass
+	//     declines (0 edges); read before, both records are SCOPE.Component,
+	//     one kind, and the pass emits a stub that DANGLES because Component
+	//     and Model share a kind family. It is not, as an earlier draft of this
+	//     comment claimed, that the allow-list would lose the re-kinded rows —
+	//     before the re-kind those records are already admitted. See the
+	//     fsharpTypeDeclKinds block for the full experiment, and
+	//     TestFSharpFieldTypeRefs_PlacementAfterTheElmishRekindIsLoadBearing
+	//     for the grader.
 	//   - It must run after EVERY producer that can add a same-file record,
 	//     because its ambiguity rule counts the graph nodes a name denotes in
 	//     this file and a collision it cannot see reaches the resolver as a
--- a/internal/extractors/fsharp/field_type_refs.go
+++ b/internal/extractors/fsharp/field_type_refs.go
@@ -36,16 +36,25 @@ import (
 //     (extractor.go's moduleRE/typeRE/memberRE + extractIndentBody). There is
 //     no tree-sitter grammar for F# in this tree, so "stash from the AST" is
 //     not an available option, not a choice declined.
-//  2. `member_type` IS THE VERBATIM SOURCE TEXT, NOT A LOSSY PROJECTION.
+//  2. `member_type` IS THE VERBATIM SEGMENT TEXT, NOT A LOSSY PROJECTION.
 //     parseRecordFields takes everything after the first `:` in the field
 //     segment and only TrimSpace's it (du_record_members.go:140). Probed
 //     against `Order`, `Customer option`, `Customer list`,
 //     `Map<string, Customer>`, `Customer[]`, `int * Customer`,
-//     `int -> Customer`, `Domain.Customer` and `'T`, the property is character
-//     for character the declared type as written. It is the one shipped
-//     field-type property on this issue that loses nothing, which is why this
-//     arm needs no stash and why the fixture in
-//     field_type_refs_6912_test.go pins all nine shapes against literals.
+//     `int -> Customer`, `Domain.Customer` and `'T`, it is character for
+//     character what was written. It is the one shipped field-type property on
+//     this issue that loses nothing — swift's is the first pre-order
+//     type_identifier and yields `Dictionary` for `Dictionary<String, Order>` —
+//     which is why this arm needs no stash.
+//
+//     THE PRECISE CLAIM IS "THE SEGMENT AFTER THE COLON", NOT "THE TYPE".
+//     Nothing strips a trailing comment, so `Anno: Customer // why` arrives as
+//     `"Customer // why"` and scans to candidates [Customer why]. It is
+//     harmless — `why` is refused by the declaration gate and the right edge
+//     still emits — but it is a real edge of the verbatim claim rather than a
+//     rounding of it, and a file that happened to declare `type why` would get
+//     a spurious edge. Pinned by the trailing-comment row in
+//     TestFSharpFieldTypeRefs_MemberTypeIsVerbatimSourceText.
 //
 // F#'s POSTFIX GENERICS ARE NOT A SPECIAL CASE — THEY ARE WHY TOKENISING WORKS
 //
@@ -230,29 +239,51 @@ type fsharpFieldTypeTarget struct {
 // dropped there. Graded by TestFSharpFieldTypeRefs_ElmishModelAndMsgAreTargets,
 // which fails when those two rows alone are deleted.
 //
-// THESE TWO ROWS AND THE CALL PLACEMENT ARE ONE DECISION, NOT TWO GUARDS, AND
-// SCORING THEM SEPARATELY IS MISLEADING. Measured rather than argued:
+// THE CALL PLACEMENT IS LOAD-BEARING TOO — BUT THROUGH RULE 2, NOT THIS
+// ALLOW-LIST. This paragraph has been wrong twice in opposite directions and
+// the sequence is worth keeping, because the second error is the rarer one.
+//
+// The first draft said the after-applyElmishFeliz placement was load-bearing
+// BECAUSE the allow-list needs the re-kinded rows. A three-row experiment
+// refuted that justification:
 //
 //	drop the two rows, keep the placement          -> DEAD (1 test)
 //	move the call before applyElmishFeliz,
-//	  keep the two rows                            -> ALIVE
-//	move the call AND drop the two rows            -> ALIVE
+//	  keep the two rows                            -> ALIVE  <- fixture-bound
+//	move the call AND drop the two rows            -> ALIVE  <- fixture-bound
 //
-// Before the re-kind, `Model` is still SCOPE.Component/record and `Msg` still
-// SCOPE.Component/discriminated_union — both already admitted — so the earlier
-// placement needs no extra rows and produces identical output. The two
-// arrangements are equivalent, which means the middle row above is the honest
-// verdict for the placement: it is NOT independently graded, and the two rows'
-// kill holds only while the placement is held constant. Recorded because a pair
-// of mutually-masking guards grades neither, and the failure mode is claiming
-// both are load-bearing when one arrangement of the pair is what matters.
+// Before the re-kind, `Model` is SCOPE.Component/record and `Msg` is
+// SCOPE.Component/discriminated_union — both already admitted — so on THAT
+// input the two arrangements agree, and the allow-list explanation is indeed
+// wrong. The second draft then RETRACTED THE CLAIM ALONG WITH ITS REASON. That
+// over-corrected: the claim survives on a different mechanism, and withdrawing
+// it left a reachable dangling-edge mutant ungraded.
 //
-// The after-placement is nonetheless the one kept, on an argument that does not
-// depend on the Elmish rows at all: read at the END, the pass sees the FINAL
-// Kind and Subtype of every record, so a future producer that re-kinds or adds
-// a same-file record cannot silently invalidate either the allow-list or the
-// ambiguity rule. Reading earlier would rest on the invariant "no later pass
-// changes a Kind" — an invariant applyElmishFeliz already breaks.
+// The distinguishing input is the commonest module name in an Elmish app:
+//
+//	module App
+//	open Elmish
+//	open Foo.Model      // import placeholder, SCOPE.Component, named "Model"
+//	type Model = { Count: int }
+//	type Env = { State: Model }
+//
+//	as shipped (read AFTER the re-kind)  -> 0 edges
+//	call moved BEFORE the re-kind        -> 1 edge, DANGLING through
+//	                                        BuildIndex -> ReferencesEmbedded
+//
+// The separator is RULE 2. Read after the re-kind, `Model` denotes two kinds
+// (SCOPE.Model/elmish_model + SCOPE.Component/import), so rule 2 refuses. Read
+// before, both records are SCOPE.Component — one kind — so rule 2 is silent,
+// the allow-list admits, and the emitted stub cannot resolve: Component and
+// Model are BOTH in componentKindFamily, so lookupLocationKind finds no unique
+// answer and ambigLocation blanks it. Graded by
+// TestFSharpFieldTypeRefs_PlacementAfterTheElmishRekindIsLoadBearing.
+//
+// So: the two allow-list rows are graded (each individually — deleting either
+// one alone kills), AND the placement is graded, and they are graded by
+// DIFFERENT rules. Recorded at length because "the justification failed, so the
+// claim falls" is the wrong inference — re-derive whether the claim still holds
+// before retracting it.
 //
 //	REFUSED, and each of the three refusals is a distinct hazard:
 //
@@ -308,14 +339,35 @@ var fsharpTypeDeclKinds = map[string]bool{
 //
 //  2. A name that denotes MORE THAN ONE DISTINCT GRAPH NODE in this file is
 //     dropped, and the count spans EVERY record in the file rather than only
-//     the admitted ones. This is the resolver's own ambiguity rule mirrored at
-//     the emit site: buildSymbolIndex marks (file, name) ambiguous on meeting a
-//     second distinct entity ID (`existing != e.ID`, internal/resolve/refs.go:1308),
-//     and the byLocation fallback this address depends on consults
-//     ambigLocation first — so a name shared with a NON-target node cannot bind
-//     either, and emitting would guarantee a dangling stub. Arm C's rule (count
-//     only within the admitted set) would not see such a collision; #7038 is
-//     filed against it for that.
+//     the admitted ones. Arm C's rule (count only within the admitted set)
+//     would not see such a collision; #7038 is filed against it for that.
+//
+//     IT IS A CONSERVATIVE APPROXIMATION OF THE RESOLVER'S RULE, NOT A MIRROR
+//     OF IT, and the earlier draft of this comment claimed otherwise: it said
+//     that a second kind means statusAmbiguous so "emitting would GUARANTEE a
+//     dangling stub". That premise is false in the direction the rule is
+//     graded. Measured by hand-emitting the refused edge on this arm's own
+//     shadow fixture and driving it through BuildIndex -> ReferencesEmbedded:
+//
+//     rival SCOPE.Operation/let beside SCOPE.Component/record  -> BINDS, correctly
+//     rival SCOPE.Component/import beside SCOPE.Model          -> DANGLES
+//
+//     The separator is the KIND FAMILY. refs.go:2982 tries
+//     lookupLocationKind(file, name, structuralKindFamilies("component"))
+//     BEFORE the kind-agnostic byLocation tier that ambigLocation guards, so a
+//     rival OUTSIDE componentKindFamily is simply not weighed and the edge
+//     resolves. The rule is therefore genuinely load-bearing only when the
+//     rival kind is INSIDE that family — Component/View/Model, the case rule 2
+//     shares with the call-placement experiment above — and it silently costs
+//     recall for every out-of-family rival (SCOPE.Operation, SCOPE.Pattern, the
+//     SCOPE.Schema sub-populations).
+//
+//     The over-strictness is KEPT here rather than narrowed: mirroring the
+//     family table at the emit site is real coupling to the resolver's internal
+//     tiering and deserves its own decision across arms C/D/E, not a rider on
+//     this one. What is fixed is the stated reason. The general form — count
+//     the kinds the resolver TIER THAT RESOLVES YOUR ADDRESS actually weighs —
+//     is recorded on #6912.
 //
 //     THE UNIT IS THE KIND, NOT THE RECORD, and that too is the resolver's own
 //     rule rather than a convenience: graph.EntityID hashes
--- a/internal/extractors/fsharp/field_type_refs.go
+++ b/internal/extractors/fsharp/field_type_refs.go
@@ -1,0 +1,434 @@
+package fsharp
+
+import (
+	"strings"
+
+	"github.com/cajasmota/grafel/internal/extractor"
+	"github.com/cajasmota/grafel/internal/types"
+)
+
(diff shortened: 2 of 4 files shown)


cajasmota:
# REQUEST-CHANGES — independent adversarial review of arm E (F#)

Reviewed `78396a50d` against `a16fd6d61` in my own worktree, with a **private `GOCACHE`** in this lane's scratchpad exported for every `go` invocation (the shared cache was purged mid-run today by a concurrent `go clean -cache`; a phantom build failure scores as *not applicable*, not ALIVE/DEAD, so it removes mutants silently rather than announcing itself). Every mutant below applied with an asserted **exact occurrence count**, restored by `cp` from a shasum-verified backup taken from **this** tree; `git diff HEAD` empty afterwards, both mutated files back to their exact SHA-256.

Nothing here is a correctness hazard in the emitted graph. What blocks is **one equivalence claim that is false**, and **one load-bearing premise that is observably false on the very fixture that grades it** — and they are the same mechanism seen from opposite sides.

---

## BLOCKING 1 — M11b is not equivalent. It is distinguishable, production-reachable, and ungraded; and the retraction over-corrected

The header reports `M11b — call moved before applyElmishFeliz → ALIVE, equivalent`, on the measured argument that before the re-kind `Model`/`Msg` are already-admitted `SCOPE.Component` subtypes so "the two arrangements produce identical output". That measurement is fixture-bound. The distinguishing input is not exotic — it is the commonest module name in an Elmish app:

```fsharp
module App
open Elmish
open Foo.Model          // import placeholder, SCOPE.Component, named "Model"
type Model = { Count: int }
type Env = { State: Model }
```

| arrangement | emitted | through the production resolver |
|---|---|---|
| **as shipped** (after `applyElmishFeliz`) | **0 edges** | — |
| **M11b** (before `applyElmishFeliz`) | **1 edge** | **DANGLING** `scope:component:class:fsharp:App.fs:Model` |

Mechanism: the allow-list is not what separates them — the **ambiguity rule** is. Read *after* the re-kind, `Model` denotes two kinds (`SCOPE.Model/elmish_model` + `SCOPE.Component/import`) → refused. Read *before*, both are `SCOPE.Component` → one kind → admitted → an edge that `lookupLocationKind` cannot resolve uniquely (both `Component` and `Model` are in `componentKindFamily`) and that `ambigLocation` then blanks. The shipped suite is **green under M11b**, so this is *distinguishable and ungraded*, not equivalent.

The consequence for the header: **the first draft's claim that the placement is load-bearing was right**; the three-row experiment retracted a true claim because its fixture had no in-family rival. The replacement argument ("reading at the END protects against a *future* re-kind") understates it — `applyElmishFeliz` already breaks it *today*, via the ambiguity rule rather than the allow-list.

**Price to close: ~15 lines** — the fixture above asserting zero edges, plus a `t.Log`-free resolver assertion if you want the dangle direction pinned. Please also correct the two prose sites (`field_type_refs.go` `fsharpTypeDeclKinds` block, `extractor.go:541-548`) so no claim in the diff still rests on the retracted-then-wrongly-retracted version.

## BLOCKING 2 — rule 2's stated premise ("emitting would guarantee a dangling stub") is false in the direction it is graded

`field_type_refs.go` (rule 2) and `TestFSharpFieldTypeRefs_TypeShadowedByAnotherKindGetsNoEdge` both assert that a name denoting two kinds means `statusAmbiguous`, so "any edge would **DANGLE**". I hand-emitted exactly the refused edge on the author's own fixture (`let Customer = 42` beside `type Customer`, premise verified: `SCOPE.Operation/let Customer` **and** `SCOPE.Component/record Customer` both present) and drove it through `BuildIndex` → `ReferencesEmbedded`:

```
SHADOW-PROBE: BOUND to Domain.fs:SCOPE.Component/record:Customer
```

It **binds, correctly**, via `refs.go:2982` `lookupLocationKind(file, tail, structuralKindFamilies("component"))` — the kind-aware tier that runs *before* `ambigLocation`. Same result with a `SCOPE.Operation` rival beside a re-kinded `SCOPE.Model`.

So rule 2 is a **conservative approximation** of the resolver's rule, not a mirror of it: it is genuinely load-bearing only when the rival kind is **inside `componentKindFamily`** (`Component`/`View`/`Model` — the case in BLOCKING 1), and it silently costs recall for every out-of-family rival (`SCOPE.Operation`, `SCOPE.Pattern`, the `SCOPE.Schema` sub-populations). I am not asking for the rule to be narrowed in this PR — mirroring the family table at the emit site is real coupling and deserves its own decision. I am asking that the prose stop claiming a dangle that does not happen, since the same false premise is what #7038 is about in arm C and the go arm carries the same wording. Worth a follow-up issue across arms C/D/E.

---

## Verified, and stronger than claimed

- **No F# tree-sitter grammar** — confirmed. Nothing F# in `go.mod` (ocaml is present; it is not F#), nothing in `third_party/` (swift only), and `extractor.go:20` already said so pre-existing. The "different technique because there is no AST" justification stands.
- **`member_type` round-trip** — reproduced all nine shapes verbatim, and five the fixture does not carry: the **real** F# mutable position (`mutable Bal: Acct` → `Acct`; note the table's `Mut: mutable Acct` is not legal F#, so that row grades a shape that cannot occur), `float<kg>`, `Acct'`, `Result<Acct, string>`, all correct. One caveat on the wording: it is the source text **after the first colon**, not strictly the type — `Anno: Acct // trailing comment` yields `"Acct // trailing comment"` and candidates `[Acct trailing comment]`. Harmless under the declaration gate, and the right edge still emits; "character for character the declared type as written" should read "the segment text after the colon".
- **The live kill is real, and it is a wrong binding** — hand-emitting the refused edge for `open Acme.Customer` in a file declaring no `Customer`: `IMPORT-PROBE: BOUND to App.fs:SCOPE.Component/import:Customer`. The `indexByName`(`:1611`)/`byLocation`(`:1271-1317`) asymmetry is read correctly. This is the strongest justification on the arm and it is under-sold, not over-sold.
- **The two Elmish rows are individually graded** — I scored them separately rather than trusting the pair (this is the shape that grades neither): deleting **only** `SCOPE.Event/elmish_msg` → **DEAD**; deleting **only** `SCOPE.Model/elmish_model` → **DEAD**, both on `..._ElmishModelAndMsgAreTargets`. M2b is not a masked compound. Re-kinding confirmed live (`SCOPE.Model/elmish_model`, `SCOPE.Event/elmish_msg`), and both re-kinded targets bind through the resolver — `SCOPE.Model` via the family tier, `SCOPE.Event` only via the kind-agnostic `byLocation` fallback (`SCOPE.Event` is not in `componentKindFamily`), which is worth a sentence somewhere.
- **Three `SourceFile` conjuncts, three rows** — confirmed by reading `field_type_refs_scope_6912_test.go`: pass 1 gets its own cross-file `SCOPE.Operation` row, pass 2 the foreign-target rows, the field loop the `Foreign.Buyer` row. No compound offered.
- **DU refusal** — re-scored: removing only `r.Subtype != "field"` → **DEAD** on `..._DUCasePayloadGetsNoEdge`. The "two independent reasons" are rationale, not two guards, so there is no masking pair here.
- **Self-reference guard** — re-scored (`&& false`, since deleting it stops compiling): **DEAD**. `parent_class` is the bare `typeRE` name, so the `t.name == owner` comparison is on the right dialect.
- **Corpus / ratchet** — `find` over `archigraph-corpora` returns **0** `.fs/.fsi/.fsx/.fsproj`; `internal/quality/golden` holds **0** `*.fs*`. The baseline could not have moved, and declining `cmd/grafel` as coverage (properties discarded, keyed `(from,to,kind)`) is right.
- **`'T` closed vs go's shipped over-fire** — confirmed: `TestGoFieldTypeRefs_KnownOverFire_TypeParameterShadowedByASameFileType` pins a go field typed by a type *parameter* binding to a same-file `type T`. That is a live wrong binding in a shipped arm, and go's AST *can* distinguish a `type_parameter_list`. Worth filing against arm D rather than leaving pinned.
- Baseline green under my private cache: `./internal/extractors/fsharp/... ./internal/resolve/... ./internal/extractor/... ./internal/graph/...`, `gofmt -l` clean, `go vet` clean.

## A mutant nobody ran

**RM3 — drop the `r.Kind != "SCOPE.Schema"` conjunct from the field loop, keeping `Subtype != "field"`: ALIVE.** Verdict: *equivalent for any input `extractFSharp` can produce* — `memberSubtype := "field"` at `du_record_members.go:46` is the **only** site in the package that mints Subtype `field`, and it always sets `SCOPE.Schema`. But it is **distinguishable through the direct-call path the arm already uses** (a hand-built `SCOPE.Component/field` record), so by this arm's own standard for M13/M14/M15 it is ungraded. **Price: ~4 lines** in `field_type_refs_scope_6912_test.go`, next to the `Foreign.Buyer` row.

Also DEAD, for the record: hoisting `emitted` out of the per-field loop (cross-field dedup, distinct from M9's per-field dedup) → **DEAD, 3 tests**.

## On the hand-built probe

Substituting a 4-file probe for an empty corpus is the right response and the "produced nothing there" framing is honest. But it establishes only that **the address dialect binds for the shapes its author imagined**: 19/19 bound says nothing about yield on real F#, nothing about orphan clearance for #6726's `Schema` population, and — as both blocking items show — nothing about the shapes the author did not think of. Both defects above live in inputs the probe omits (an import placeholder colliding with a re-kinded `Model`; a `let` shadowing a type). The probe's shapes are representative of the *happy* path and convenient on the collision path. Fixing the two prose/grading items is enough for me; I would not ask for a corpus that does not exist.

---

Re-request review after the M11b fixture + the two prose corrections and I will re-score only those.

cajasmota:
## Revision pushed — `342176d1f`. Both blocking items reproduced in my own harness before fixing, both closed. Comments and tests only; **the non-comment diff in `field_type_refs.go` and `extractor.go` is empty**, so rules 1 and 2 are behaviourally untouched and no baseline can move.

### BLOCKING 1 — you were right, and the retraction was the error, not the original claim

Reproduced independently before touching anything, on your fixture:

| arrangement | emitted | through `BuildIndex` → `ReferencesEmbedded` |
|---|---|---|
| as shipped (after `applyElmishFeliz`) | **0 edges** | — |
| call moved before | **1 edge** | **DANGLE** `scope:component:class:fsharp:App.fs:Model` |

…with the shipped suite green under the mutant. Distinguishable, production-reachable, ungraded — exactly as you scored it.

`TestFSharpFieldTypeRefs_PlacementAfterTheElmishRekindIsLoadBearing` closes it: **M11b is DEAD.** Two premise guards (the re-kind happened; the placeholder is named `Model`) and a positive control (`Cfg: Config`), so a failure separates "the ordering held" from "this file emits nothing".

Both prose sites corrected. The header now keeps the whole sequence rather than just the endpoint, because the transferable part is the second error, not the first:

> "the justification failed, so the claim falls" is the wrong inference — re-derive whether the claim still holds before retracting it.

Worth naming what actually went wrong in my reasoning: my three-row experiment varied the *placement* while holding the *rival* constant, and there was no in-family rival in any of the three fixtures. So it measured the allow-list explanation correctly and was blind to the rule-2 one. That is the same fixture-axis failure I listed in my own PR body under "varied vs held constant" — I varied the rival record for the *allow-list* rows and not for the *placement*.

### BLOCKING 2 — reproduced, and the false half is exactly where you put it

Hand-emitted the refused edge on my own `let Customer = 42` fixture and on the in-family case, both through the production resolver:

```
OUT-OF-FAMILY (SCOPE.Operation rival):  BOUND to SCOPE.Component/record:Customer
IN-FAMILY (SCOPE.Component/import vs SCOPE.Model):  DANGLE
```

So "would guarantee a dangling stub" is false for every out-of-family rival. Rule 2 is now documented as a **conservative approximation**: load-bearing only when the rival is inside `componentKindFamily`, silently costing recall for `SCOPE.Operation` / `SCOPE.Pattern` / the `SCOPE.Schema` sub-populations, because `refs.go:2982` tries `lookupLocationKind(…, structuralKindFamilies("component"))` **before** the `ambigLocation`-guarded tier.

Behaviour unchanged, per your instruction — the narrowing is real coupling to the resolver's tiering and belongs to a decision across arms C/D/E, not a rider here. The comment points at your general rule on #6912. I have not touched go's arm or #7037.

The test's own comment carried the same false premise and is corrected too.

### RM3 — closed rather than excused

`TestFSharpFieldTypeRefs_OnlySchemaKindedFieldsAreAnchors`, 4 lines beside the `Foreign.Buyer` row, with a positive control. **DEAD.** Your framing is the one I adopted: equivalent for anything `extractFSharp` can produce, but ungraded *by this arm's own standard for M13/M14/M15*, so it gets the same treatment.

### Re-scored after the prose edits (exact occurrence count asserted pre-write, postcondition asserted post-write)

| mutant | verdict |
|---|---|
| M11b — call moved before `applyElmishFeliz` | **DEAD** (was ALIVE) |
| RM3 — `Kind != "SCOPE.Schema"` dropped | **DEAD** (was ALIVE) |
| M3 — ambiguity rule removed | **DEAD**, now 2 tests |
| M2b-model — only the `SCOPE.Model` row | **DEAD** |
| M2b-event — only the `SCOPE.Event` row | **DEAD** |

I re-ran your individual-row scoring rather than citing it: both Elmish rows kill alone, so M2b was never a masked compound.

### Fixture corrections you found

- **`Mut: mutable Customer` is not legal F#** — `mutable` prefixes the *field*. That row graded a shape that cannot occur. Replaced with `mutable Bal: Customer`, which is what `parseRecordFields`' `TrimPrefix("mutable ")` actually exists for.
- **Trailing comments ride along.** `Anno: Customer // why this field exists` → `member_type` is `"Customer // why this field exists"`, candidates `[Customer why this field exists]`. Now a pinned row, and the header's claim is restated precisely: **the segment after the colon, not the type**. Harmless under the declaration gate and the right edge still emits — but a file declaring `type why` would get a spurious edge, so it is a real limitation rather than a rounding.

### On the probe — stated, not defended

Added to the PR body: the 4-file probe is **sufficient to ship the arm and not sufficient to claim yield or orphan clearance for #6726**. It establishes that the address dialect binds for the shapes its author imagined, and both of your blocking items lived in shapes it omits — an import placeholder colliding with a re-kinded `Model`, and a `let` shadowing a type. Since #6726 is expressed in orphan rates and four of its six headline numbers were already defects in our own measurement, "19/19 bound" must not be read as a yield figure for that issue.

Also worth a sentence, from your notes: `SCOPE.Model` binds via the family tier while `SCOPE.Event` binds only via the kind-agnostic `byLocation` fallback (`SCOPE.Event` is not in `componentKindFamily`) — so the two admitted re-kind rows do not resolve by the same route.

### Gates (private `GOCACHE` for every `go` call)

`./internal/extractors/fsharp/...`, `./internal/resolve/...`, `./internal/quality/...` (+ `analytics`/`audit`/`fitness`), `./internal/extractor/...` green. `go test -count=1 ./cmd/grafel/` green (158s) in a fresh `GRAFEL_HOME`/`GRAFEL_DAEMON_ROOT` — run because the original brief mandates it for an extractor change, cited as nothing more, since the non-comment diff is empty and the digest discards properties anyway. Ratchet: 38 held, `baseline.json` untouched, `git status` shows only this arm's four files. `gofmt -l` clean, `go vet` clean. Tree restored by `cp` from a backup taken from **this** tree; both mutated files back to their exact SHA-1, `git diff HEAD` empty.

Not filed by me, but I agree they should be, and they are yours to place: go's `TestGoFieldTypeRefs_KnownOverFire_TypeParameterShadowedByASameFileType` pins a live wrong binding that Go's AST *can* fix (`type_parameter_list`), and the rule-2 wording is shared with arms C and D.

Re-requesting review.

PR #7042, opened 2026-09-08
fix(#6912): arm H — PHP fields carried their declared type only as a string property; emit the field→type REFERENCES edge (and the corpus yield is ZERO — read §12)

cajasmota:
## Arm H (php) of #6912 — and the headline is a ZERO

PHP fields now emit the field→declared-type `REFERENCES` edge (`ref_kind=field_target_type`), same-file targets only, matching arms A (C#), B (proto), C (rust), D (go), F (java) and E (F#).

**Measured yield on the available corpus: 0 edges from 331 field entities across 390 `.php` files.** Not an implementation failure — a structural fact about PHP, quantified below. **Read the "Is this arm worth merging?" section before merging; the decision is yours, not mine.**

The mechanism itself is exercised and correct: **23 edges across the 9 fixture files, 23 bound / 0 dangling**, driven through the real resolver (`BuildIndex` → `ReferencesEmbedded`) rather than asserted at the emit site.

**Revised after independent review** — F1–F4 fixed in `f84fc8e92`; see §13.

---

## 1. The resolver tier that resolves a PHP field's address — driven, not inferred

The ToID is `extractor.BuildComponentStructuralRef` → `scope:component:class:php:<file>:<Name>`. `resolveStructuralRef` (`internal/resolve/refs.go:2982`) tries, in order:

| tier | site | what it weighs |
|---|---|---|
| 1 | `lookupLocationKind(file, name, structuralKindFamilies("component"))` — `refs.go:2982` | `componentKindFamily` only (`refs.go:2341`), plus each entity's SCOPE-trimmed alias (`refs.go:1290-1296`) |
| 2 | `ambigLocation[file][name]` — `refs.go:3004` | **every kind** — set whenever two distinct entity IDs share (file, name) (`refs.go:1302-1315`) |
| 3 | `byLocation[file][name]` — `refs.go:3007` | kind-agnostic |

**Every target this arm admits is a `SCOPE.Component`, so every edge it emits is answered at tier 1 and never reaches tiers 2/3.** Established by driving `BuildIndex` → `ReferencesEmbedded` over records the real extractor produced, not by reading `refs.go`.

### Which kinds that tier weighs, and why

`componentKindFamily` = `{Component, Class, View, Model, SCOPE.Component, SCOPE.View, SCOPE.Model}`, closed under the index's trim alias — so the membership test is `K ∈ family ∨ TrimPrefix(K,"SCOPE.") ∈ family`, and `SCOPE.Class` is in because `Class` is. Eight entries, in `phpComponentAddressFamily`. (Arm F shipped this table with `SCOPE.Class` missing in a first revision while its comment claimed both spellings; the hole is not re-introduced, and `TestPhpFieldTypeRefs_Unit_ScopeClassParticipatesViaTheTrimAlias` asserts every entry can actually suppress, so no entry is decoration.)

**And the honest part: rule 2 is VACUOUS as shipped — for a stronger reason than the first revision gave.** `SCOPE.Component` is the only `componentKindFamily` kind *any* PHP producer emits — core `internal/extractors/php` and custom `internal/custom/php` were both enumerated, neither emits `SCOPE.Model`, `SCOPE.View`, `SCOPE.Class`, nor a bare `Component`/`Class`/`View`/`Model`. **The reviewer pointed out that the custom half of that enumeration is irrelevant**: this pass runs inside core `Extract`, *before* `MergeWithCustom`, so no `internal/custom/php` record ever reaches it. The core half alone settles it, *a fortiori* — the reason has been tightened rather than the claim weakened. (Enumerating the custom lane was still the right check: it is where the future producer would live.) So `len(nameKinds[name]) > 1` cannot fire from PHP source today. It is kept as a resolver mirror for the producer that adds an in-family kind later (an Eloquent-model detector is the obvious candidate) and, being unreachable from `Extract`, it is graded by direct call rather than claimed as fixture-covered.

## 2. Measured counterfactuals

Harness: the real extractor over 390 corpus `.php` files, computing the shipped edge set and each rival rule's edge set per file, diffing named rows.

| rule | edges (corpus, 390 files) | edges (9 fixture files) | rows changed vs shipped |
|---|---|---|---|
| **shipped** (family-scoped kind count) | **0** | **23** | — |
| **arm D** (count ALL same-file kinds) | 0 | **22** | **drops `Order.price -> Money`** (`twotables.php`), −4.3% |
| **arm F** (family-scoped kind count) | 0 | 23 | **0 — this IS the shipped rule**, reached from php's own tier, not borrowed |
| **arm C** (count RECORDS in the admitted set) | 0 | 23 | 0 on both — unreachable in PHP: two admitted same-name declarations in one file is a fatal redeclaration |
| **no allow-list** (any same-file `SCOPE.Component`) | **1** | **26** | **adds 3 WRONG BINDINGS** — `Order.carrier -> Ship` (import placeholder), `Order.ns -> App` (**namespace record**, added for F3), `Middleware.middleware -> Closure` (the corpus one); see §3 |
| **enums admitted as targets** | 0 | **24** | adds `Order.status -> Status`, which **dangles**; see §4 |

The corpus deltas are all 0 because the corpus total is 0 — a delta over an empty set separates nothing. The counterfactuals are therefore **also** measured over every fixture file this arm ships, where they are non-vacuous, using the same harness against the real extractor. That is the honest form: arm D's rule is not refused because it cost 19% here, it is refused because PHP keeps **classes and functions in separate symbol tables**, so

```php
class Money {}
function Money() { return 1; }   // legal PHP — different symbol table
class Order { public Money $price; }
```

puts a `SCOPE.Operation` beside the `SCOPE.Component`. Arm D's rule refuses that target. Tier 1 binds it without hesitation, because `SCOPE.Operation` is not in the family it weighs — asserted through the real resolver in that test, not argued. **Arm C's rule is not copied and neither is its stated reason** (#7038).

## 3. The wrong-binding kill, reproduced from the corpus rather than invented

`useImportRecord` (`php.go:650`) mints one `SCOPE.Component` per `use`, named the **top namespace segment**, with an **empty Subtype**. For an *unqualified* use there is no segment to strip, so the placeholder is named exactly what a field types:

```
laravel-routing/src/Illuminate/Routing/Controllers/Middleware.php
  use Closure;
  public function __construct(public Closure|string|array $middleware, ...)
```

Without `phpTypeDeclSubtypes` this pass emits `Middleware.middleware -> Closure` **and it binds** — to an import placeholder. It is the *one and only* edge the whole 390-file corpus produces without the allow-list, and it is wrong. A wrong binding never reaches `bug-extractor` because it binds. `TestPhpFieldTypeRefs_UnqualifiedUseImportIsNeverATarget` is that file's shape, and it asserts the placeholder record exists first so it cannot pass vacuously; `TestPhpFieldTypeRefs_ImportPlaceholderIsNeverATarget` covers the namespace-root spelling in a file that declares **no** class of that name, so rule 2 cannot mask the allow-list.

`buildNamespace` (`php.go:483`) mints the same shape for the file's own `namespace`, and is refused by the same list.

## 4. PHP enums are refused, and it is a measurement

`enum Status: string { case Open = 'open'; }` is minted **twice in one file** — `SCOPE.Schema/enum` by `buildEnum` (`php.go:384`) and a `SCOPE.Enum` value-set of the **same Name** by `buildEnumValueSet` (`const_valueset.go:181`). Two kinds, two nodes, and **neither is in `componentKindFamily`**. Driven through the real resolver:

```
enum Status: string { case Open = 'open'; }  ->  DANGLE (statusAmbiguous, tier 2)
enum Empty2 {}   (no cases, so no value-set) ->  binds
```

So admitting enums would emit a guaranteed dangling stub for every enum anybody actually writes. `TestPhpFieldTypeRefs_EnumTargetWouldDangle` hand-emits the refused edge and asserts the dangle, with a class target in the same harness that binds — so a failure cannot be the harness.

**This is a named recall loss.** `public Status $status` is ordinary PHP and stays orphan. Closing it needs a **second address dialect**: a schema-space ref would hit `lookupLocationKind` with `schemaKindFamily` `{Schema, Field, Property, SCOPE.Schema}`, where the `SCOPE.Enum` value-set is *not* a member and the `SCOPE.Schema/enum` resolves uniquely. Deliberately not taken here — two dialects behind one `ref_kind` is #6451's shape and should be decided across arms, not invented by this one.

## 5. Fixture axes

**VARIED** (one per property of one class, so a failure names the axis): type composition (bare / `?T` / `A|B` / `A&B` / `(A&B)|null` / union-with-primitive); name qualification (bare / `Ns\T` / `\Ns\T` / `\T`); target existence (this file / another file / nowhere); target kind (class, interface, trait, enum, import placeholder, namespace, file carrier); PHP builtin (`int`, `string`, `array`, `iterable`, `callable`, `mixed`, `self`); declaration site (property, **promoted constructor parameter**, multi-declarator `public Money $m1, $m2;`, untyped); identifier case (declared spelling vs lower-cased use).

**HELD CONSTANT**: one file per fixture (the cross-file direction is graded by the unit test, deliberately); one declaring class where a second would confuse the self-reference rule; visibility `public` except where promotion requires otherwise; no framework/ORM annotations (the custom lane is gated off by default and is not this pass's input).

Every field entity the fixture emits **must** have a row in the shape table and every row must match a real entity — a property added without a row fails, and a row naming no field fails as vacuous. The scanner gets its own enumeration (30 rows) over the same syntax space, which is what makes union/intersection/qualification graded as properties rather than as three lucky fixtures.

**Promoted constructor parameters are covered** — `emitPhpFieldMembers` emits them through the same `add` closure, so they are one more `SCOPE.Schema/field` with a `field_type`. Asserted (`Order2.owner`, `Order2.amount`, `Order2.count`), and the corpus wrong-binding above is itself a promoted parameter.

## 6. Two same-file conjuncts, graded in opposite directions

There are **three** `SourceFile != filePath` sites (rival scan, target scan, field loop) and none is reachable through `Extract`, since it is called once per file. Graded by direct call, in **opposite failure directions** so grading one does not make the other look covered:

- a cross-file `SCOPE.Model` named `Customer` wrongly **suppressing** `a.php`'s own unambiguous `Customer` (rival scan)
- a cross-file `interface Shipper` wrongly **becoming** a target for `a.php` (target scan)
- a `b.php` field wrongly picking up `a.php`'s target set (field loop)

Plus the mirror — asking for `b.php` returns `b.php`'s own declarations — so "absent" is never "the function returned nothing".

## 7. No primitive blocklist

`int`, `string`, `array`, `callable`, `iterable`, `object`, `mixed`, `never`, `void`, `false`, `null`, `true`, `float`, `bool`, `self`, `static`, `parent` are collected as candidates like any other identifier and refused by the **declaration gate**. No blocklist, and none is needed: PHP **reserves** every one of those as a class name (`class int {}` is a fatal error), so unlike F#'s shadowable type abbreviations there is no legal same-file declaration a blocklist would have to out-rank. `self`/`static` are refused a second time by the self-reference rule.

## 8. Qualified names vanish whole

`\` is scanned as an identifier continuation character **and** may start a token, so `Ns\Order`, `\Ns\Order` and `\Order` each arrive as one token and are rejected as one. Reducing to the last segment is the failure arm D found in Go's `resolveTypeReferences`. A **leading** separator is refused for a second, independent reason: in a file with `namespace App;`, bare `Money` means `App\Money` while `\Money` means the *global* `Money` — different types, so binding one to the other would be wrong. Conservative in the namespace-less case, which is the direction to be conservative in.

## 9. Grading — 25 mutants, 23 DEAD, 2 equivalent with the reason stated

Every mutant `go vet`-compiled, applied through a script that **asserts an exact occurrence count before and after** (three mutants aborted as "pattern occurs 0 times, edit NOT applied" rather than being scored as ALIVE), reverted by `cp` from a shasum-verified byte backup, with `git status --porcelain` checked clean after each. Final tree shasums are byte-identical to the baseline.

| # | mutant (call site, not an extracted helper) | verdict |
|---|---|---|
| M1 | allow-list disabled | DEAD (2) — both import-placeholder tests |
| M2 | ambiguity rule disabled | DEAD (2) |
| M3 | ambiguity count widened to ALL kinds (**arm D's rule**) | DEAD (2) |
| M4 | `SCOPE.Class` dropped from the family table | DEAD (1) |
| M5 | leading `\` skipped instead of consumed whole | DEAD (4) |
| M6 | `\` no longer an identifier continuation | DEAD (1) |
| M7 | return the FIRST candidate only (**swift's exact defect**) | DEAD (3) |
| M8 | store folded, look up unfolded | DEAD (15) — *plumbing half, not evidence about folding* |
| M8b | store folded, look up folded → unfolded | DEAD (10) — *ditto* |
| M8c | **fold neither side** — the coherent case-sensitive alternative | DEAD (6), incl. both behavioural tests |
| M9 | self-reference rule disabled | DEAD (1) |
| M10 | `ref_kind` near-miss (`field_target_types`) | DEAD (11) |
| M11 | `trait` row removed | DEAD (2) |
| M12 | `interface` row removed | DEAD (4) |
| M13 | `SourceFile` conjunct, rival scan | DEAD (1) |
| M14 | `SourceFile` conjunct, target scan | DEAD (1) |
| M15 | `SourceFile` conjunct, field loop | DEAD (1) |
| M16 | field-loop `Kind` conjunct | DEAD (1) |
| M17 | field-loop `Subtype` conjunct | DEAD (1) |
| M18 | folded-spelling collision branch | DEAD (1) |
| M19 | **the call site in `Extract` removed** | DEAD (11) |
| M20 | per-target dedup removed | **ALIVE → graded → DEAD** |
| M21 | **call moved to immediately after `walk()`** | ALIVE — *equivalent*, see below |
| M22 | `Relationships` assigned instead of appended | DEAD (2) |
| M23 | empty-`field_type` guard disabled | ALIVE — *equivalent*: `phpFieldTypeCandidates("")` returns nil, so the loop body never runs |
| M25 | `class` row removed | DEAD (15) |

**M8/M8b are not offered as evidence about case folding.** They are inconsistent halves of one mechanism and are lethal for that reason alone; M8c is the coherent alternative implementation and is the row that grades the behaviour.

**M20 was ALIVE and was closed rather than recorded.** Two candidates fold to one target only for `Money|money`, which PHP rejects as a duplicate union member — so no `.php` file reaches it. But `attachPhpFieldTypeRefs` takes **records**, and "one edge per target, never per candidate" should not rest on an upstream parser's error handling. Graded by unit call; M20 is now DEAD.

**M21's ALIVE is an equivalence, and the enumeration behind it is complete** (an equivalence resting on a partial enumeration is worthless). The seven passes between `walk()` and the call site, and what each emits: `emitConfigConsumerEdges` → `SCOPE.Config`; `emitTemplateRenderEdges` → `SCOPE.Template`; `emitTranslationKeyEdges` → `SCOPE.TranslationKey`; `emitExceptionFlowEdges` → `SCOPE.ExceptionType` (observed: `SourceFile "<exception>"`, `Name "exception:X"`); `emitConstValueSets` → `SCOPE.Enum`; `emitPHPTestScopeOwner` → `SCOPE.Operation` (`tests.go:83`); `attachPhpExtends` → **relationships only, no records**. **None** is in `phpComponentAddressFamily` and none is a `SCOPE.Component`, so neither placement can differ. The comment therefore claims placement is *not* load-bearing rather than inventing a reason for it — and names what would make it load-bearing (a future pass minting a bare-named Component).

Two things I deliberately did **not** copy: arm C's record-counting rule (#7038), and the "*emitting would guarantee a dangling stub*" justification carried by arms C, D and E — it is false in general, a conservative approximation load-bearing only for in-family rivals. This arm's refusals each state their own measured reason.

## 10. Gates

- `go test -count=1 ./internal/extractors/php/ ./internal/custom/php/ ./internal/resolve/ ./internal/extractor/ ./internal/quality/` — green. `gofmt -l` clean.
- **`go test -count=1 ./cmd/grafel/`** — green in 169s with a fresh `GRAFEL_HOME` + `GRAFEL_DAEMON_ROOT`, no skips.
- **The digest did not move, and the reason is that its fixture has no PHP file at all** — all 18 files enumerated (`cs/*` ×7, `java/OrdersController.java`, `ts/routes.ts`, `web/server.js`, `myproj/*` ×2, `myapp/*` ×4, `svc/handler.go`, `rs/routes.rs`). A new `REFERENCES` edge **would** move it (`R <from> -[Kind]-> <to>` is hashed; only properties are not), so this is a real negative result, not a blind instrument.
- **`scripts/quality/run.sh --ratchet`** — "38 gated fixtures held", and `git status --porcelain` empty afterwards, which is the actual check. **But it is vacuous for this change and I am not offering it as evidence:** the golden set's only `.php` file, `php-slim-mini/src/restapi/v1/index.php`, declares **zero** classes and **zero** typed properties. There was nowhere for a php edge to appear.

## 11. Recall ceiling and corpus caveat, stated where the numbers are

23/23 bound is a **bind rate, not completeness**. Observed ceilings, each by probe:

- **Same-file only.** The dominant one for PHP — see below.
- **PHP enums** — refused, §4.
- **PHP 8.2 DNF types** (`(A&B)|null`) — `phpDeclaredType` captures **nothing**: `field_type` is `""` and the **Signature loses the type too**. A capture defect that predates this arm and degrades the Signature whether or not this arm exists, so it is pinned as a known ceiling (`TestPhpFieldTypeRefs_DNFTypeIsAKnownCaptureCeiling`, which fails loudly if the capture is fixed) rather than fixed in passing. **Filed as #7044**; the test references the issue instead of re-describing it.
- **Untyped properties** (`public static $untyped`) — no datum, nothing to bind.

**Corpus**: 390 `.php` files in **5 repos, effectively 4** — symfony-routing 224, laravel-routing 60, laravel-quickstart 53, symfony-demo 52, awesome-compose 1. Vendor directories excluded. A narrow corpus, and a clean result on it.

## 12. Is this arm worth merging? The zero, explained

| repo | .php files | field entities | files declaring >1 type | …of those, with fields | edges |
|---|---|---|---|---|---|
| symfony-routing | 224 | 130 | 7 | 3 | 0 |
| laravel-routing | 60 | 119 | 0 | 0 | 0 |
| laravel-quickstart | 53 | 16 | 0 | 0 | 0 |
| symfony-demo | 52 | 66 | 1 | 0 | 0 |
| awesome-compose | 1 | 0 | 0 | 0 | 0 |
| **total** | **390** | **331** | **8** | **3** | **0** |

**PSR-4 mandates one class per file.** 8 of 390 files declare more than one addressable type, 3 of those have any field, and none of those 3 names a same-file type. So the same-file constraint — which cost arms D/F/E a slice of recall — costs PHP **essentially all of it**. This is not a corpus accident; it is the language's dominant convention.

Where the PHP prize actually is, measured with the same harness: **43 field candidates resolve to a class declared in a *different* file of the same repo** (symfony-routing 29, symfony-demo 14, and 0 in both Laravel repos, whose fields are typed with framework classes outside the tree). That is an **upper bound for a naive cross-file rule** — it ignores namespaces and `use` aliasing, so some of the 43 would be wrong bindings — and it is the number that should size a cross-file arm.

**Why I still think it should merge** (the call is the coordinator's):

1. The refusals are the durable part and they are now graded from PHP source: the `Closure` import-placeholder wrong binding is **live in the corpus today** and would be the first thing a cross-file arm walked into.
2. The enum dangle is measured and recorded, so nobody re-derives it.
3. Vocabulary parity across seven languages, with a zero-edge corpus meaning zero regression risk.
4. It becomes non-zero on the multi-declaration files that do exist, and immediately meaningful when the cross-file arm lands.

**Why it might not**: it clears **no orphan** on the corpus that motivated #6726, and a producer that adds no measured edges is code with no measured benefit. If the board's standard is measured orphan reduction, the honest answer is that **PHP needs the cross-file arm, and this is its groundwork rather than its fix.** I am not going to dress a zero up as a win.

## 13. Independent review — what it broke, and what it could not

The review re-derived the headline from its own probe and its own scanner re-implementation: `FILES=390 FIELDS=331 EDGES=0 MULTITYPE=8 MULTITYPE_WITH_FIELDS=3 CROSSFILE_CANDS=43`, every cell of §12 and the per-repo split matching, the `Closure` wrong binding reproduced as **exactly one** corpus edge, the rule-2 vacuity enumeration confirmed complete, M21's seven passes confirmed (six take `&entities` and mutate in place — the shape most able to invalidate that equivalence — and none appends a `SCOPE.Component`), and the digest fixture's 18 keys confirmed. It notes the **43 is one well-corroborated measurement, not two**: their method shares an input shape with mine.

It broke four things, all sub-behavioural, none able to change a corpus edge. All four are fixed in `f84fc8e92`:

**F1 — a fixture row the language does not permit.** `public callable $cb;` is **not legal PHP**: typed properties exclude `callable` (context-dependent) and `void`. It sat under a comment claiming every row was *"legal PHP 8.1, checked shape by shape against the grammar"* — so the comment was false as well as the row, and the version label was wrong too (`(A&B)|null` is **8.2** DNF). Replaced with `public object $obj;`, same axis. The comment now records the correction rather than quietly applying it: this is the **third** instance of this shape on the board, and the pattern is that the *claim of having checked* is what goes unchecked.

**F2 — a distinguishable, ungraded conjunct (the real find).** Pass 2's `r.Kind != "SCOPE.Component"` is **mutually masked** by the subtype allow-list: no core PHP producer mints Subtype ∈ {class, interface, trait} under any other Kind, so the two guards only ever fire together and the allow-list alone carried every refusal the suite observed — *including the enum one*, which `TestPhpFieldTypeRefs_EnumIsNotATarget` therefore grades through the allow-list, not through this conjunct. Two guards that only fire together grade neither.

The distinguishing input is `{Kind: "SCOPE.Model", Subtype: "class"}` — **exactly the future Eloquent-model producer this PR cites to justify keeping rule 2**, which is graded against that producer by direct call. Grading one and not the other was an inconsistency, not a considered scope. `TestPhpFieldTypeRefs_Unit_OnlyComponentsAreTargets` closes it, **enumerating** seven non-Component Kinds (`SCOPE.Model`, `SCOPE.View`, `SCOPE.Class`, `SCOPE.Schema`, `SCOPE.Enum`, `SCOPE.Pattern`, `SCOPE.Service`) each carrying an **admitted** subtype so the allow-list cannot be what refuses them, each with a positive control in the same input, plus the mirror direction (the same Kinds with a refused subtype) so the two guards are shown independent rather than one standing in for the other.

|---|---|---|---|
| **N1** | pass-2 `r.Kind != "SCOPE.Component"` **deleted** | ALIVE | **DEAD** — `Unit_OnlyComponentsAreTargets` |
| **N8** | same conjunct **widened to admit `SCOPE.Schema`** | ALIVE | **DEAD** — `Unit_OnlyComponentsAreTargets` |

**F3 — an axis row claiming more than its content.** The target-kind axis listed `namespace`, and §3 asserted `buildNamespace` is refused by the same allow-list, but **no test drove a field typed at a namespace root**. `imports.php` now declares `public App $ns;` under `namespace App;`, with the bare `SCOPE.Component`/`""` record asserted to exist first so the refusal is doing work. It is a *different producer* from `useImportRecord` with the same shape, which is why the import tests did not already cover it — and the no-allow-list counterfactual now shows it as a third wrong binding.

**F4 — an untested interaction already sitting in the fixture.** `models.php` declared **both** `use Ship\Thing;` and `class Ship {}` and never typed a field `Ship` — the one case where the allow-list must **admit** rather than refuse, and it binds only because `graph.EntityID` hashes `(repo, Kind, Name, SourceFile)` with **Subtype excluded**, so the two records collapse to one node instead of tripping tier 2's kind-agnostic `ambigLocation`. Nothing pinned that. `Order2.carrier` is now typed `Ship`, and `TestPhpFieldTypeRefs_ImportPlaceholderSharingAClassNameIsOneNode` asserts the two-record / one-ID premise **before** the binding assertion, so it fails at the premise if `EntityID` ever starts hashing Subtype rather than the arm quietly beginning to dangle.

Also from the review, and accepted: **N2** (`strings.EqualFold(t.name, owner)` → `==`) is genuinely **equivalent** — both derive from the same class-declaration identifier text, and the fold-collision guard already refuses the only shape that could separate them. Left as `EqualFold` because it is the correct predicate for a case-insensitive language, not because a mutant demanded it.

**Not re-verified by the review, so stated as mine alone**: the fixture aggregate (now 23 edges / 23 bound), which the reviewer explicitly did not recount.

Gates re-run after the fixes: `./internal/extractors/php/ ./internal/custom/php/ ./internal/resolve/` green; `./cmd/grafel/` green in 154s in a fresh isolated `GRAFEL_HOME`/`GRAFEL_DAEMON_ROOT` with 0 skips; ratchet "38 gated fixtures held" with `git status --porcelain` empty afterwards (still vacuous for PHP, as stated in §10).

Refs #6912, #6726, #7038, #7044, #6451, #6976, #6369.

### DIFF
--- a/internal/extractors/php/field_type_refs.go
+++ b/internal/extractors/php/field_type_refs.go
@@ -1,0 +1,484 @@
+package php
+
+import (
+	"strings"
+
+	"github.com/cajasmota/grafel/internal/extractor"
+	"github.com/cajasmota/grafel/internal/types"
+)
+
+// field_type_refs.go — the field→declared-type edge for PHP (issue #6912, arm H).
+//
+// Shipped predecessors: arm A csharp/field_type_refs.go (#6984), arm B
+// proto/field_type_refs.go (#6991), arm C rust/field_type_refs.go (#7000), arm D
+// golang/field_type_refs.go (#7036), arm F java/field_type_refs.go (#7039), arm E
+// fsharp/field_type_refs.go (#7040).
+//
+// THE GAP. emitPhpFieldMembers (field_members.go:60) mints one SCOPE.Schema/field
+// entity per typed class property and per promoted constructor parameter, and
+// records the declared type ONLY as Properties["field_type"] and inside Signature.
+// The single edge attached to it is the owner→member CONTAINS, which the orphan
+// definition excludes by design, so every PHP field is a LEAF on its outbound
+// side. Contributor #6726 measures SCOPE.Schema at 99.3% orphan of 35,353
+// entities; PHP fields are minted as SCOPE.Schema, so they are part of that
+// population.
+//
+// ============================================================================
+// 1. THE PROPERTY IS LOSSLESS — VERIFIED BY PROBE, NOT BY READING
+// ============================================================================
+//
+// phpDeclaredType (field_members.go:135) returns the VERBATIM source span of the
+// first primitive_type | named_type | optional_type | union_type |
+// intersection_type | type_list child. Probed through the real extractor on 18
+// property shapes, every one arrives character for character as written:
+//
+//	?Customer   Customer|Money   Customer&Shipper   \Other\Order   ?\Money
+//	string|int  self             array              iterable       int
+//
+// So unlike swift — whose field_type is the FIRST pre-order type_identifier and
+// yields `Dictionary` for `Dictionary<String, Order>` — PHP's needs no stash from
+// the AST. Arm C's technique is declined because it is unnecessary here, not
+// because it is wrong. Pinned by TestPhpFieldTypeRefs_FieldTypeIsVerbatim.
+//
+// PROMOTED CONSTRUCTOR PARAMETERS ARE COVERED FOR FREE and are asserted rather
+// than assumed: emitPhpFieldMembers emits them through the SAME `add` closure
+// with the same Properties, so `public function __construct(public readonly
+// Customer $owner)` is one more SCOPE.Schema/field with field_type="Customer".
+// Graded by the promoted rows of TestPhpFieldTypeRefs_ShapeSpace.
+//
+// ============================================================================
+// 2. THE RESOLVER TIER THAT RESOLVES THIS ADDRESS — DRIVEN, NOT INFERRED
+// ============================================================================
+//
+// The ToID is extractor.BuildComponentStructuralRef, i.e.
+// `scope:component:class:php:<file>:<Name>`. resolveStructuralRef
+// (internal/resolve/refs.go:2982) tries, in order:
+//
+//  1. lookupLocationKind(file, name, structuralKindFamilies("component"))
+//     — i.e. componentKindFamily {Component, Class, View, Model, SCOPE.Component,
+//     SCOPE.View, SCOPE.Model} (refs.go:2341), plus each entity's SCOPE-trimmed
+//     alias (refs.go:1290-1296), which is what makes a bare `Class`-kinded
+//     entity a family member too.
+//  2. ambigLocation[file][name] (refs.go:3004) — set when TWO DISTINCT ENTITY
+//     IDS share (file, name), REGARDLESS OF KIND (refs.go:1302-1315).
+//  3. byLocation[file][name] — kind-agnostic.
+//
+// Every target this pass admits is a SCOPE.Component, so every edge it emits is
+// answered at TIER 1 and never reaches tiers 2/3. That is the whole basis of the
+// ambiguity rule below, and it was established by driving the real resolver
+// (BuildIndex → ReferencesEmbedded) over records produced by the real extractor,
+// not by reading refs.go. TestPhpFieldTypeRefs_ResolvesToEntityIDs is that drive.
+//
+// ============================================================================
+// 3. WHY PHP ENUMS ARE NOT TARGETS — A MEASUREMENT, NOT A PREFERENCE
+// ============================================================================
+//
+// buildEnum (php.go:384) mints `enum Status: string { case A = 'a'; }` as
+// SCOPE.Schema/enum named "Status", and emitConstValueSets → buildEnumValueSet
+// (const_valueset.go:181) ADDITIONALLY mints a SCOPE.Enum value-set with the
+// SAME Name in the SAME file. Two kinds, two graph nodes (graph.EntityID hashes
+// (repo, Kind, Name, SourceFile)), and NEITHER is in componentKindFamily.
+//
+// A component-space ref to such a name therefore misses tier 1, and hits tier 2
+// as AMBIGUOUS. Driven through the real resolver:
+//
+//	enum Status: string { case A = 'a'; }   ->  DANGLE (statusAmbiguous)
+//	enum Empty2 {}   (no cases, no value-set) ->  binds
+//
+// So admitting enums would emit a dangling stub for every enum that has at least
+// one case — i.e. for every enum anybody writes. A non-binding edge is WORSE than
+// no edge: it is kept verbatim and classified `bug-extractor`, a full hit on the
+// disposition denominator. Enums are refused, and the refusal is graded in both
+// directions by TestPhpFieldTypeRefs_EnumIsNotATarget (the enum field gets no
+// edge) with a same-fixture positive control (a class field in the same class
+// does get one), so the test cannot pass by the pass being broken.
+//
+// THIS IS A RECALL LOSS AND IT IS NAMED AS ONE. `public Status $status` is
+// ordinary PHP and stays orphan. Closing it needs a SECOND ADDRESS DIALECT — a
+// schema-space ref, which would hit lookupLocationKind with schemaKindFamily
+// {Schema, Field, Property, SCOPE.Schema}, where the SCOPE.Enum value-set is NOT
+// a member and the SCOPE.Schema/enum resolves uniquely. That is a real option
+// with a measured basis, and it is deliberately NOT taken here: shipping two
+// address dialects behind one ref_kind is #6451's defect shape, and it should be
+// decided across arms rather than invented by this one.
+//
+// ============================================================================
+// 4. THE AMBIGUITY RULE — DERIVED FROM THE TIER, NOT COPIED FROM AN ARM
+// ============================================================================
+//
+// The standing rule on #6912 is: COUNT THE KINDS THAT THE RESOLVER TIER WHICH
+// RESOLVES YOUR ADDRESS ACTUALLY WEIGHS. Per section 2 that tier is
+// lookupLocationKind filtered to componentKindFamily, so this pass counts kinds
+// WITHIN THAT FAMILY — arm F's scope, reached from php's own tier rather than
+// borrowed. The UNIT is the KIND and not the RECORD, mirroring refs.go:1308:
+// within one file two records sharing a Kind are ONE node, so counting records
+// would refuse edges that bind perfectly well. Arm C counts records and is wrong
+// under the rule (#7038); its rule is not copied and neither is its stated
+// reason.
+//
+// AND THE HONEST PART: SCOPE.Component IS THE ONLY componentKindFamily KIND ANY
+// PHP PRODUCER EMITS. Core (internal/extractors/php) and custom
+// (internal/custom/php) were both enumerated: neither emits SCOPE.Model,
+// SCOPE.View, SCOPE.Class, nor a bare Component/Class/View/Model. So
+// len(nameKinds[name]) can never exceed 1 from PHP source today and rule 2 is
+// VACUOUS AS SHIPPED. It is kept as a mirror of the resolver for the producer
+// that adds an in-family kind later — exactly the shape arm F's SCOPE.Class hole
+// was — and because it is unreachable from Extract it is graded by a direct-call
+// unit test (TestPhpFieldTypeRefs_Unit_InFamilyRivalSuppresses...) rather than
+// claimed to be covered by the fixture suite.
+//
+// The counterfactual of ARM D's rule (count ALL same-file kinds) is not vacuous
+// here, and it is the reason this arm does not inherit it: PHP has separate
+// symbol tables for classes and functions, so `class Money {}` beside
+// `function Money() {}` in one file is LEGAL PHP and puts a SCOPE.Operation
+// beside the SCOPE.Component. Arm D's rule refuses that target; tier 1 binds it
+// without hesitation, because SCOPE.Operation is not in the family it weighs.
+// Graded by TestPhpFieldTypeRefs_SameNamedFunctionDoesNotSuppressTheClass, whose
+// edge is additionally driven through the real resolver.
+//
+// ============================================================================
+// 5. THE TARGET ALLOW-LIST — THE LIVE WRONG-BINDING KILL
+// ============================================================================
+//
+// buildUseImports/useImportRecord (php.go:622-660) mints ONE SCOPE.Component per
+// `use` statement whose Name is the TOP namespace segment, and buildNamespace
+// (php.go:483) does the same for the file's own `namespace`. Both carry an EMPTY
+// Subtype. So `use Ship\Thing;` puts a Component literally named `Ship` in this
+// file — and a component-space ref to `Ship` BINDS TO IT (driven, section 2's
+// harness). Without the subtype allow-list a field typed `Ship` in that file
+// binds to an IMPORT PLACEHOLDER: a confident wrong answer rather than a dangle,
+// and a wrong binding is the direction `bug-extractor` cannot flag. Graded by
+// TestPhpFieldTypeRefs_ImportPlaceholderIsNeverATarget, whose file declares NO
+// class of that name so rule 2 cannot mask the allow-list.
+//
+// SCOPE.Component/file (extractor.FileEntity) is refused by the same list. It is
+// unreachable by name shape — its Name is the file path and a candidate token
+// can never contain `/` or `.` — and that is claimed as belt, not as the reason.
+//
+// ============================================================================
+// 6. NO PRIMITIVE BLOCKLIST
+// ============================================================================
+//
+// `int`, `string`, `array`, `callable`, `iterable`, `object`, `mixed`, `never`,
+// `void`, `false`, `null`, `true`, `float`, `bool`, `self`, `static`, `parent`
+// are collected as candidates like any other identifier and refused by the
+// DECLARATION GATE: the file does not declare a type of that name. No blocklist
+// exists, and none is needed — PHP reserves every one of those words as a class
+// name (`class int {}` is a fatal error), so unlike F#'s shadowable type
+// abbreviations there is no legal same-file declaration a blocklist would have to
+// out-rank. Graded by the primitive rows of TestPhpFieldTypeRefs_ShapeSpace.
+//
+// `self`/`static`/`parent` additionally name the OWNER, which is refused a second
+// time by the self-reference rule in attachPhpFieldTypeRefs.
+//
+// ============================================================================
+// 7. SAME-FILE TARGETS ONLY
+// ============================================================================
+//
+// A field whose type is declared in another .php file gets NOTHING. Binding a
+// bare type name across files is exactly the #6976/#6369 hazard, and PHP makes it
+// sharper than most: `use App\Models\Order;` means the bare `Order` in this file
+// denotes a class in a DIFFERENT file, and the same bare name is declared in
+// dozens of files across a Symfony/Laravel tree. Closing it needs the cross-file
+// namespace view pass 1 has none of. Separate arm, not a widening of this one.
+//
+// There are TWO independent same-file conjuncts and they fail in OPPOSITE
+// directions, so they are graded separately rather than one paying for the other:
+// the rival scan in phpInFileTypeTargets (a cross-file record wrongly SUPPRESSING
+// an edge) and the target scan (a cross-file record wrongly BECOMING a target).
+// TestPhpFieldTypeRefs_Unit_TargetsAreScopedToTheRequestedFile covers both.
+
+// phpFieldTargetRefKind is the value of the `ref_kind` edge property. It matches
+// arm A's csFieldTargetRefKind, arm B's protoFieldTargetRefKind, arm C's
+// rustFieldTargetRefKind, arm D's goFieldTargetRefKind, arm F's
+// javaFieldTargetRefKind and arm E's fsharpFieldTargetRefKind verbatim — the
+// discriminator is what makes the edge queryable as a field→declared-type edge
+// across languages, so it must be one string. Arm D established that the six
+// independent literals are deliberate: six graders enforcing agreement
+// transitively beat one shared constant.
+const phpFieldTargetRefKind = "field_target_type"
+
+// phpFieldTypeCandidates returns every UNQUALIFIED identifier written in a PHP
+// type expression, in source order and without deduplication.
+//
+// It is a scanner, not a parser. It models neither `?` nor `|` nor `&` nor
+// `(A&B)|null` DNF grouping, and that is the point: every named type in a PHP
+// type expression appears as a bare identifier whatever composes them, so there
+// is no type-syntax table to keep in sync. A UNION names several types and EACH
+// gets its own edge (`Customer|Money` → two), because each alternative genuinely
+// is a declared type the field may hold; an INTERSECTION likewise (`A&B` — the
+// value is both). Nullability contributes nothing: `?Customer` names exactly the
+// one type `Customer`.
+//
+// ONE token shape is consumed WHOLE and discarded, and consuming it WHOLE is what
+// makes it safe — a scanner that merely SKIPPED the backslash would re-enter
+// mid-token and produce exactly the bare name the skip exists to prevent:
+//
+//   - A NAMESPACE-QUALIFIED name, in either spelling: relative (`Other\Order`) or
+//     fully-qualified with a leading separator (`\Other\Order`, `\Money`). `\` is
+//     scanned as an identifier continuation character, AND a token may START with
+//     it, so the whole qualified name arrives as one token and is rejected as one.
+//
+// Reducing `Other\Order` to `Order` is precisely the failure arm D found in Go's
+// resolveTypeReferences, where `other.Order` bound to a same-file `Order` — a
+// wrong binding, which never reaches `bug-extractor` because it binds. A LEADING
+// separator is refused for a second and independent reason: in a file that
+// declares `namespace App;`, the bare `Money` means `App\Money` while `\Money`
+// means the GLOBAL `Money`, so they are different types and binding one to the
+// other would be wrong in the namespaced case. It is conservative in the
+// namespace-less case (where the two coincide) and that is the direction to be
+// conservative in. Both spellings are graded by the qualified rows of
+// TestPhpFieldTypeRefs_ShapeSpace and by
+// TestPhpFieldTypeRefs_QualifiedTypeIsNotStrippedToASameFileName.
+//
+// A field type is at most a few dozen characters, so the scan is linear in the
+// property length with no allocation beyond the result slice.
+func phpFieldTypeCandidates(typ string) []string {
+	var out []string
+	r := []rune(typ)
+	i := 0
+	for i < len(r) {
+		switch {
+		case r[i] == '\\':
+			// Fully-qualified name — consume the separator AND every segment
+			// after it, so no segment can re-enter the scanner on its own.
+			i++
+			for i < len(r) && (isPhpIdentRune(r[i]) || r[i] == '\\') {
+				i++
+			}
+		case isPhpIdentStartRune(r[i]):
+			j := i
+			for j < len(r) && (isPhpIdentRune(r[j]) || r[j] == '\\') {
+				j++
+			}
+			if tok := string(r[i:j]); !strings.ContainsRune(tok, '\\') {
+				out = append(out, tok)
+			}
+			i = j
+		default:
+			i++
+		}
+	}
+	return out
+}
+
+// isPhpIdentStartRune / isPhpIdentRune implement PHP's identifier grammar,
+// `[a-zA-Z_\x80-\xff][a-zA-Z0-9_\x80-\xff]*`. The high-byte range is why
+// non-ASCII letters are admitted rather than treated as separators: `class Ünité`
+// is a legal PHP class and a field typed `Ünité` must produce the same edge as an
+// ASCII one.
+func isPhpIdentStartRune(r rune) bool {
+	return r == '_' || (r >= 'a' && r <= 'z') || (r >= 'A' && r <= 'Z') || r >= 0x80
+}
+
+func isPhpIdentRune(r rune) bool {
+	return isPhpIdentStartRune(r) || (r >= '0' && r <= '9')
+}
+
+// phpFieldTypeTarget is one in-file type declaration a field-type edge may
+// address: the structural ToID that binds to it, and the DECLARED spelling of the
+// name for the `target_type` property and the self-reference check.
+type phpFieldTypeTarget struct {
+	toID string
+	name string
+}
+
+// phpTypeDeclSubtypes is the target ALLOW-LIST over SCOPE.Component subtypes.
+//
+// It is an allow-list rather than a denylist because PHP emits several BARE-NAMED
+// SCOPE.Components that are not type declarations (section 5), and a future
+// producer adding another one must be refused BY DEFAULT. Enumerated from this
+// package's own emit sites rather than inherited from an arm:
+//
+//	ADMITTED — every subtype buildComponent (php.go:304) is ever called with:
+//	  class      php.go:127
+//	  interface  php.go:181
+//	  trait      php.go:214
+//
+//	REFUSED:
+//	  ""     — buildNamespace (php.go:483) and useImportRecord (php.go:650).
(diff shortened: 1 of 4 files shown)


cajasmota (commented):
## Independent adversarial review — arm H (php), head `1037307e5`

I did not write this code and I went looking for the load-bearing claims to break. **I broke three, all small; the headline survived every attack I could mount.** Everything below was re-derived in my own worktree with a private `GOCACHE`, from an independently written probe (my own scanner re-implementation, my own counting) driven through the real `Extract`, not through the author's harness.

---

### 1. THE ZERO IS REAL — and every number in §12 reproduces exactly

This was the claim most likely to be an artefact (a corpus-relative zero has produced two confident false readings on this board). It is not.

My probe: walk `archigraph-corpora`, skip `/vendor/`, run the registered php extractor per file, count `SCOPE.Schema/field` records, count relationships with `ref_kind=field_target_type`, and independently re-derive the multi-type and cross-file numbers.

```
FILES=390 FIELDS=331 EDGES=0 MULTITYPE=8 MULTITYPE_WITH_FIELDS=3 CROSSFILE_CANDS=43
REPO awesome-compose        files=  1 fields=   0 multi= 0 multiWithFields= 0 edges=0
REPO laravel-quickstart     files= 53 fields=  16 multi= 0 multiWithFields= 0 edges=0
REPO laravel-routing        files= 60 fields= 119 multi= 0 multiWithFields= 0 edges=0
REPO symfony-demo           files= 52 fields=  66 multi= 1 multiWithFields= 0 edges=0
REPO symfony-routing        files=224 fields= 130 multi= 7 multiWithFields= 3 edges=0
```

Every cell of the §12 table matches: 390 / 331 / 0 / 8 / 3 / 43, and the per-repo split 130 / 119 / 16 / 66 / 0. A separate `find -name '*.php' -not -path '*/vendor/*'` over the *whole* corpora tree returns 390 in exactly those 5 repos, so the file set is not a filtered subset. The three multi-type-with-fields files are named (`UrlGeneratorTest.php`, `ObjectLoaderTest.php`, `CompiledUrlMatcherDumperTest.php`) and none names a same-file type.

**Caveat on the 43**: I got 43 too, but by re-implementing the same scanner over the same property. Two methods sharing one input shape is ONE confirmation, not two. Treat 43 as a well-corroborated single measurement.

### 2. THE LIVE WRONG-BINDING KILL REPRODUCES — exactly one edge, exactly that file

I disabled the allow-list (`if false && !phpTypeDeclSubtypes[...]`) and re-ran the corpus probe:

```
EDGE laravel-routing/src/Illuminate/Routing/Controllers/Middleware.php Middleware.middleware -> Closure
FILES=390 FIELDS=331 EDGES=1 ...
```

One edge, wrong, on the file §3 names, and it is a promoted constructor parameter. This is the durable part of the arm and it is not a story.

### 3. The rule-2 vacuity enumeration — complete, but the stated reason is over-broad

Core: over 390 real files the extractor emits `[SCOPE.Component, SCOPE.Config, SCOPE.Enum, SCOPE.ExceptionType, SCOPE.Operation, SCOPE.Schema, SCOPE.Template]`. `SCOPE.Component` is the only `componentKindFamily` member. Component subtypes observed: `class:278 interface:18 trait:12 file:389 "":1370`. Custom: every `makeEntity` call site in `internal/custom/php` passes `SCOPE.Pattern|Operation|Component|Schema|Service|Config|UIComponent|Datastore` — no `Model`/`View`/`Class`. **The enumeration is complete and the conclusion holds.**

But the *reason* is over-broad: `attachPhpFieldTypeRefs` runs inside core `Extract`, before `MergeWithCustom`, so **`internal/custom/php` records never reach this pass at all**. Enumerating them was not wasted (it is the right check for the future producer), but rule 2 is vacuous for a stronger reason than the comment gives. Worth one sentence in the header so the next arm does not inherit a weaker argument than it is entitled to.

### 4. M21's seven passes — the count is right and the pointer shape does not invalidate it

`walk()` → call site: `emitConfigConsumerEdges`, `emitTemplateRenderEdges`, `emitTranslationKeyEdges`, `emitExceptionFlowEdges`, `emitConstValueSets`, `emitPHPTestScopeOwner`, `attachPhpExtends` = **seven**. Six take `&entities` and mutate in place — the shape most able to invalidate this kind of equivalence — but none appends a `SCOPE.Component`, and `attachPhpExtends` appends relationships only. The corpus kind census above is the empirical backstop: over 390 files, no in-family kind other than `SCOPE.Component` ever appears. **M21's equivalence stands.**

### 5. The digest, and the ratchet

`writeSpanFixture6118`'s map has **18 keys**, `cs/*` ×7 (the author's list is right; a naive line-scan gives 6 and misses `cs/OrdersController.cs`), and **no `.php`**. The negative result is a blind instrument, correctly declared as such. `internal/quality/golden/php-slim-mini/src/restapi/v1/index.php` has **zero** `class|interface|trait` declarations — the ratchet is vacuous here, as stated.

### 6. The arm D counterfactual premise is true

`class Money {} function Money() {}` in one file is legal PHP (separate symbol tables), and `TestPhpFieldTypeRefs_SameNamedFunctionDoesNotSuppressTheClass` asserts the two-record premise before asserting the edge, then drives it through the resolver. Not argued — demonstrated.

---

## What I broke

### F1 — a fixture row that is not legal PHP (`callable`)

`phpFTSrc` line: `public callable $cb;`

**`callable` is not permitted as a property type in any PHP version** (typed properties exclude `void` and `callable`, because `callable` is context-dependent). `class Order2 { public callable $cb; }` is a fatal error, not a program. The fixture comment says *"every row in it is legal PHP 8.1 (checked shape by shape against the grammar, not eyeballed)"* — that claim is false, and this is the third instance of this shape on the board (F#'s `Mut: mutable Customer`).

Same comment, smaller: `public (Customer&Shipper)|null $dnf;` is **PHP 8.2** DNF, not 8.1. The header says 8.1; §5's own prose correctly says 8.2.

Behavioural impact: none — `callable` produces no edge either way. The defect is the claim, and the claim is the thing this board keeps getting bitten by. Replace with `iterable`/`object`, or move the row to a comment that says "this shape cannot appear in real source".

### F2 — pass 2's `r.Kind != "SCOPE.Component"` conjunct is ungraded (two of my mutants)

Deleting it outright (**N1**) and widening it to admit `SCOPE.Schema` (**N8**) both survive `./internal/extractors/php/ ./internal/resolve/` in full. It is **mutually masked** by the subtype allow-list: nothing in core php mints a record with `Subtype ∈ {class, interface, trait}` under any other Kind, so the allow-list alone carries the refusal — including the enum refusal, which `TestPhpFieldTypeRefs_EnumIsNotATarget` grades through the allow-list, not through this conjunct.

Not production-reachable today, so this is not a bug. But the distinguishing input is `{Kind: "SCOPE.Model", Subtype: "class"}` — **precisely the Eloquent-model detector the PR names as the reason rule 2 is kept**. Rule 2 is graded against that future producer by direct call; this conjunct, which the *same* producer would exercise, is not. One row in `TestPhpFieldTypeRefs_Unit_...` closes it. Two guards that only fire together grade neither.

### F3 — the `namespace` row of the target-kind axis is claimed, not tested

The fixture header lists *"target kind: class, interface, trait, enum, import placeholder, namespace, file carrier"*, and §3 says *"`buildNamespace` (php.go:483) mints the same shape for the file's own `namespace`, and is refused by the same list."* I confirmed `buildNamespace` mints `Name = top segment of the namespace`, `Kind = SCOPE.Component`, `Subtype = ""` — so `namespace App; class Order { public App $a; }` is the shape. **No test in either file drives a field typed at the namespace root.** The refusal is real (it is the same allow-list, and my N5 mutant kills the empty-subtype widening via the *import* tests) but the axis row is a label claiming more than its content. One property in the placeholder fixture fixes it.

### F4 — an untested interaction that is already in the fixture

`phpFTSrc` declares **both** `use Ship\Thing;` and `class Ship {}` in `models.php` — a name carried by an admitted target *and* a refused placeholder in one file — and then never types a field `Ship`. That is the one case where the allow-list must *admit* rather than refuse, and it is only safe because `graph.EntityID` hashes `(repo, Kind, Name, SourceFile)` and **not** `Subtype`, so the two records collapse to one node instead of tripping tier 2's kind-agnostic `ambigLocation`. Nothing in this arm pins that premise. Cheap to close: type one property `Ship` in `models.php`, add the row, and let `phpFTAssertAllBind` do the rest.

---

## Mutant table — 8 mutants, none previously run

All on the **call site**, biased permissive. Every edit applied through a script asserting an exact occurrence count before and after (`BEFORE count(old)=N want=N` / `AFTER count(old)=0`), `go vet`-clean, reverted by `cp` from a shasum-verified backup (`bebb5ef1c697…68d6a` restored and re-verified after every run), `git status --porcelain` empty at the end. `-count=1` throughout, against `./internal/extractors/php/ ./internal/resolve/`.

| # | mutant (all permissive except N7) | verdict |
|---|---|---|
| **N1** | pass-2 `r.Kind != "SCOPE.Component"` conjunct **deleted** | **ALIVE — distinguishable and ungraded** (F2). Not reachable from today's producers; input `{Kind:"SCOPE.Model", Subtype:"class"}` |
| **N2** | self-reference `strings.EqualFold(t.name, owner)` → `t.name == owner` | ALIVE — **equivalent**: `t.name` (declared spelling) and `parent_class` both derive from the *same* class-declaration identifier text, so they are byte-identical whenever they denote the same type; the fold-collision guard already refuses the only shape that could separate them |
| **N3** | rule 2 `len(nameKinds[r.Name]) > 1` → `> 2` | **DEAD (2)** — `Unit_InFamilyRivalSuppressesTheTarget`, `Unit_ScopeClassParticipatesViaTheTrimAlias` |
| **N4** | scanner ident branch: relative-qualified token reduced to its **last segment** instead of rejected | **DEAD (2)** — `Unit_CandidateScannerEnumeratesTheShapeSpace`, `QualifiedTypeIsNotStrippedToASameFileName` |
| **N5** | allow-list widened to admit **empty-Subtype** records (`r.Subtype != "" && !phpTypeDeclSubtypes[...]`) | **DEAD (2)** — both import-placeholder tests |
| **N6** | leading-`\` branch **also emits** the last segment (`\Money` → `Money`) | **DEAD (4)** — incl. `ShapeSpace` (`globalMoney`) and `ResolvesToEntityIDs` |
| **N7** | fold-collision keeps the **first** spelling instead of refusing both | **DEAD (1)** — `Unit_FoldedSpellingCollisionRefusesBoth` |
| **N8** | pass-2 Kind conjunct widened to admit `SCOPE.Schema` | **ALIVE — mutually masked** by the subtype allow-list; confirms F2 and shows the enum refusal rests on the allow-list, not on this conjunct |

5 DEAD / 2 ALIVE-and-masked (one finding, one confirming it) / 1 genuinely equivalent.

**Not verified**: the "22 edges across the 9 fixture files, 22 bound / 0 dangling" total — I checked the per-fixture assertions and re-derived the corpus numbers, but did not recount the fixture aggregate.

---

## The merge judgement, argued both ways

**Against merging.** A producer with zero measured output is code with no measured benefit. #6726's complaint is an orphan rate, and this clears zero orphans on the only PHP corpus we have — not "few", *zero*, across 331 field entities. Every line of it (484 source + 1,106 test) is maintenance surface whose behaviour no corpus observes, so the next person to touch `phpDeclaredType`, `useImportRecord` or `componentKindFamily` pays a review cost against a pass that has never emitted anything real. And there is a specific way merging it banks a claim the corpus does not support: seven-language "vocabulary parity" on `ref_kind=field_target_type` reads, in any dashboard or roll-up, as *PHP field→type edges exist* — when what exists is a mechanism that has produced none. The measured prize is the **43 cross-file candidates**, and this arm is architecturally excluded from all of them; shipping it makes PHP look done.

**For merging, and this is where I land.** The zero is not the whole product. Three things this PR produces are *not* zero and would otherwise be re-derived from scratch by the cross-file arm:

1. The `Closure` import-placeholder wrong binding is **live in the corpus today** — reproduced above, one edge, exactly one — and it is the first thing a cross-file rule walks into, because a cross-file rule enlarges the candidate set without changing the placeholder shape. Catching it now costs nothing; catching it after a cross-file arm ships costs a wrong-binding regression that `bug-extractor` cannot flag *because it binds*.
2. The enum double-mint (`SCOPE.Schema/enum` + `SCOPE.Enum`, neither in family) is measured, driven through the real resolver, and pinned by a test that **fails loudly if the resolver ever starts binding it**. That is the standing gate from #6912's own comment thread ("only emit when the target is known to exist"), discharged for PHP.
3. The DNF capture ceiling — `field_type` *and* `Signature` both empty for `(A&B)|null` — is a pre-existing defect in the property producer that this arm found by enumerating the shape space, pinned so that fixing it fails visibly. **That one deserves its own issue and it is not a field-edge issue.**

Zero edges also means zero regression risk, which is the rare case where "merge the groundwork" is not a hand-wave: there is no corpus behaviour to break. And the arm is *not* permanently zero — it fires on the 8 multi-declaration files the moment one of them grows a same-file-typed property, and the same-file target scan is the exact component a cross-file arm keeps as its inner loop.

**Where I think the author is over-correcting.** §12's "I am not going to dress a zero up as a win" is the right instinct, but it undersells the actual deliverable. The load-bearing output of this PR is not 0 edges — it is **one reproduced wrong binding, one measured dangle class, and one capture ceiling**, none of which required the edge producer to exist but all of which are now pinned by tests that fail when the premise changes. Merge it as *evidence with a mechanism attached*, and let the merge note say so — but do not close #6726's PHP slice, and do not count PHP in any "N languages shipped" line without the qualifier.

**Recommendation: approve.** F1–F4 are all sub-behavioural (a fixture row that cannot appear in real source, two conjuncts that are unreachable-but-ungraded, one axis label ahead of its content) and none of them can change a single corpus edge, because there are none to change. Fix F1 and F2 in a follow-up (or in this branch — F1 is a one-word edit and F2 is one table row in the existing unit test), and **file the DNF capture ceiling as its own issue** before it gets forgotten inside a php-field-edge PR.

PR #7043, opened 2026-09-08
feat(#6912): swift's field_type property threw the type away before an edge could use it — emit the field→type REFERENCES edge from the AST, same-file targets only

cajasmota:
Arm G of #6912. Six arms shipped before this one: A=C# (#6984), B=proto (#6991), C=rust (#7000), D=go (#7036), F=java (#7039), E=fsharp (#7040).

Swift fields are minted as `SCOPE.Schema` (`internal/extractors/swift/field_members.go`), so they are **in** the population contributor #6726 measures at 99.3% orphan, not adjacent to it.

**19 edges, 19 bound, 0 dangling** over 234 `.swift` files in 3 corpora.

> **Review correction — the first cut of this PR emitted 38 edges and 19 of them were WRONG BINDINGS.** An independent review found that half the corpus edges bound to an `extension` carrier rather than to a declaration. `swiftDeclSubtype` has no `extension` case and falls through to `return "class"`, so `extension String { … }` mints `SCOPE.Component/class:String` in that file and pass 2 could not tell it from a real declaration. It **binds**, so it never reaches `bug-extractor` — the exact direction this arm already refused for type parameters. Fixed here; the subtype fallthrough itself is filed as **#7047** and deliberately not fixed on this PR (§2a). Section 4 records what the structural cause was: the fixture pair varied the cross-file axis and the extension axis **separately and never their intersection**, which is where half the corpus lives.

---

## 1. The trap this arm exists to close — probed, not read

`field_type` is `firstDescendantText(type_annotation, "type_identifier")`, i.e. the **first pre-order** identifier. I dumped the real CST for 21 property shapes before writing anything:

| declaration | grammar shape | `field_type` |
|---|---|---|
| `var d: Dictionary<String, Order>` | `user_type[ tid "Dictionary", type_arguments[…"Order"…] ]` | **`"Dictionary"`** |
| `var j: Set<Order>` | same | **`"Set"`** |
| `var k: Result<Order, MyErr>` | same | **`"Result"`** |
| `var h: Foundation.Data` | `user_type[ tid "Foundation", ".", tid "Data" ]` | **`"Foundation"`** |
| `var c: [Order]` | `array_type[ user_type "Order" ]` | `"Order"` (by luck of ordering) |
| `var b: Order?` | `optional_type[ user_type "Order" ]` | `"Order"` (same) |

So for **every** generic wrapper the argument is already gone, and for a module-qualified type the property holds the **module**. #6912's remaining-work list described php/cpp/swift as *"read the property, emit the edge"*; doing that in Swift ships the `List<Customer>` → `List` binding arm A went out of its way to forbid, and binds `Foundation.Data` to a same-file `struct Foundation`.

This arm therefore takes **arm C's technique** (stash AST candidates at the emit site) and **not arm E's** (tokenise the string) — F#'s `member_type` is the declared type verbatim, Swift's is not.

`field_type` is **left exactly as it is**: fixing it changes entity Properties and `Signature` for every Swift field, which is a separate change with its own blast radius. Its lossiness is now a live expectation — `TestSwiftFieldTypeRefs_FieldTypePropertyIsLossyForGenerics` asserts `Dictionary`/`Result` *and* the edge that does **not** go to the wrapper, so the next reader meets the trap as a failing test rather than as prose.

Two more things the probe settled, each turned into a test rather than an argument:

- `@Published var s: Order` shapes `Published` as a `user_type` under a `modifiers > attribute` **sibling** of the `type_annotation`. Candidates are collected from the `type_annotation` subtree only, so the wrapper is out of reach by construction. `TestSwiftFieldTypeRefs_PropertyWrapperTypeIsNotACandidate` declares a same-file `struct Published` so the assertion can actually fail.
- `Order.Type` (metatype), `Foundation.Data` (module-qualified) and `Order.Inner` (nested) are **one `user_type` node with two `type_identifier` children** around an anonymous `.`. Head and tail are both refused; type **arguments** are still descended into, so `Swift.Array<Order>` yields `Order`.

---

### (1) Which resolver tier resolves a Swift field's address?

The address is `extractor.BuildComponentStructuralRef` → `scope:component:class:swift:<file>:<Name>`, so `lookupStructural` (`internal/resolve/refs.go:2982`) calls

```go
idx.lookupLocationKind(filePath, tail, structuralKindFamilies("component"))   // = componentKindFamily
```

and, on a miss, falls through to `ambigLocation` / `byLocation` (`refs.go:3004-3012`).

**Swift is the first arm where the answer is not a single tier.** It emits admissible targets in two kind spaces, and the *same* address resolves them through *different* tiers:

| target | tier that resolves it | why | ambiguity count |
|---|---|---|---|
| `SCOPE.Component` (class / struct / enum / actor / protocol) | `lookupLocationKind`, **before** `ambigLocation` | `buildSymbolIndex` indexes each entity under its Kind **and** its SCOPE-trimmed Kind (`refs.go:1207-1210`), so `SCOPE.Component` lands under `"Component"`, which `componentKindFamily` lists | **within `componentKindFamily`** (arm F's rule) |
| `SCOPE.Schema/type_alias` (`typealias Money = Double`) | `lookupLocationKind` **misses entirely**; `ambigLocation` decides | `"SCOPE.Schema"` trims to `"Schema"`, which is in no family | **every kind** (arm D's rule) |

Driven through the real resolver, not inferred: `TestSwiftFieldTypeRefs_ResolvesToEntityIDs` builds the index and runs `resolve.ReferencesEmbedded`, and `TestSwiftFieldTypeRefs_AliasShadowedByAnOperationWouldDangleUnderArmFsRule` probes the *shadowed alias address itself* and asserts it does **not** bind, with the unshadowed one asserted to bind so the negative result is about the collision and not about aliases in general.

### (2) Which kinds does each tier weigh?

- **Component tier** weighs `componentKindFamily` = `Component, Class, View, Model, SCOPE.Component, SCOPE.View, SCOPE.Model`. Membership is **derived**, not hand-listed: `swiftInComponentAddressFamily` also trims a `SCOPE.` prefix before testing, because the index does — a `SCOPE.Class` entity lands under the key `Class`. (Arm F hand-listed `"SCOPE.Class"`; deriving it keeps this in step with the resolver rather than with a snapshot of it.)
  - **`SCOPE.Enum` is invisible to it**, and that is load-bearing: `swift.go:116-120` emits a `SCOPE.Enum` value-set beside **every** `enum` Component, same name, same file. Arm D's all-kinds rule would refuse **every Swift enum target**.
- **Alias tier** weighs nothing — it is the kind-agnostic `ambigLocation`, set on any second distinct entity ID (`refs.go:1301-1316`). Arm F's rule here admits `typealias Tier` beside a top-level `func Tier()` (Schema + Operation, neither in the component family) and the edge dangles.

**The unit is the KIND, not the RECORD, in both scans.** `graph.EntityID` hashes `(repo, Kind, Name, SourceFile)` with Subtype **excluded**, and `buildSymbolIndex` marks a collision only on `existing != e.ID` (`refs.go:1308`). Arm C counts records (#7038) and is wrong; that rule is not copied and its justification comment is not reproduced.

### (3) Measured counterfactuals — same corpus, same candidates, three rules

234 `.swift` files, 3 repos (vapor, vapor-api-template, ktor-samples), 295 field entities. **Re-measured against the corrected 19-edge base** — the first cut's table was computed against the 38-edge base and its percentages were therefore wrong.

| rule | edges | delta | dropped rows |
|---|---|---|---|
| **this arm** (split by target kind) | **19** | — | — |
| **arm D** (all kinds, everywhere) | 14 | **−5, −26.3%** | `OTP.swift: HOTP.digest→OTPDigest`, `HOTP.digits→OTPDigits`, `TOTP.digest→OTPDigest`, `TOTP.digits→OTPDigits`, `URLEncodedFormData.swift: URLEncodedFormData.values→URLQueryFragment` — **all five enum targets**, each refused by its own value-set |
| **arm F** (component family, Components only) | 19 | **0** | none |
| **arm C** (record count, Components only) | 14 | **−5, −26.3%** | `HTTPMediaType.swift: HTTPMediaTypeSet.compressible→HTTPMediaTypeSet`, `…incompressible→HTTPMediaTypeSet`, `…mediaTypeLookup→HTTPMediaType`, `Middleware.swift: HTTPMiddlewareResponder.middleware→Middleware`, `Session.swift: Session._id→SessionID` |

**The dropped row sets are identical to the first cut's, row for row.** The 19 wrong bindings the extension fix removed were disjoint from the rows these three rules discriminate on, so the *deltas* survive the correction unchanged and only the percentages move. That is the result, not an assumption — each counterfactual was re-run against the fixed tree.

Three things I want stated plainly rather than folded into the win:

- **Arm F's rule costs nothing measurable here, and I am not claiming go's 19%.** The corpus emits exactly **5 `SCOPE.Schema/type_alias` records across 4 distinct names**, and **none** of them is a same-file field target. Re-derived: there are 45 `typealias` source lines, but 40 are nested inside a type body and #4854's walk emits no entity for those; of the 5 top-level ones, two share the name `VaporSendableMetatype` under `#if`/`#else` (see M6b below). The alias tier is therefore justified from the resolver and graded by fixture, and is a **corpus-relative zero** — exactly the shape arm B's `protoScalars` episode warned about, so I kept it and said why rather than deleting it on the strength of a corpus that cannot see it.
- **Arm C's −5 is a new, idiomatic corroboration of #7038.** Every one of those five is `struct Foo` beside `extension Foo` **in the declaring file**: the walk emits a second `SCOPE.Component` for the extension, so the name carries two records and **one** graph node. That is the case the extension marker must *not* break, and it is the other half of §2a: when the declaration is in the same file the two records share an `EntityID`, so refusing the carrier per-record leaves the declaration record admitted and the `ToID` identical. `TestSwiftFieldTypeRefs_ExtensionBesideItsTypeIsStillOneNode` asserts the two-record/one-kind premise first so it cannot pass vacuously.
- Arm D's −5 is the **same shape** arm F measured at 12.24% in java. Third independent confirmation that the rule is per-tier.

---

## 2a. The extension defect, and why the marker rather than the subtype

`tree-sitter-swift` routes `extension` through **`class_declaration`** — the same node type as class/struct/enum/actor — and `swiftDeclSubtype` (`swift.go:236`) switches on four declaration keywords with **no `extension` case**, falling through to `return "class"`. So an extension carrier is a real graph node with the same Kind, the same Name and the same file as the type it extends, and a **different** `EntityID` from the real declaration when that declaration lives elsewhere.

Measured, by reproducing the first cut (`swiftIsExtensionDecl` disabled at the emit site) and diffing the corpus edge set row for row:

| | edges | targets |
|---|---|---|
| first cut | **38** | — |
| this revision | **19** | — |
| **removed — every one a wrong binding** | **19** | `BaseNEncoding` ×7, `String` ×6, `RoutesBuilder` ×2, `HTTPClient`, `PasswordHasher`, `Request`, `tm` ×1 each |

`tm` is `extension tm: @retroactive Sendable` in `RFC1123.swift` — a **libc struct**, bound to a per-file Swift carrier. `Request` is `FileIO.swift`'s `extension Request`, while the real `struct Request` in `Request.swift` got **no inbound edge at all**.

**The fix marks the carrier at the emit site instead of changing the subtype**, and that is a choice, not an oversight:

1. Minting `Subtype "extension"` changes an **entity field** on every Swift extension in every indexed repo. `entityTupleKey` hashes `Subtype`, so it moves the `cmd/grafel` digest and every golden expectation naming a subtype. That is a graph-wide change riding on an edge PR — what arm D declined for `resolveTypeReferences`.
2. The deciding datum, the anonymous `extension` keyword, is **already in hand** at the emit site — the same "it is in hand, so refuse it for free" argument this arm used for type parameters.
3. **The two fixes compose.** When #7047 lands and `swiftDeclSubtype` mints `"extension"`, pass 2's allow-list (`class|struct|enum|actor|protocol`) already refuses it and the marker becomes redundant rather than wrong. Pinned by the `Subtype: "extension"` row in `ExtensionCarrierIsNeverATarget`.

**The marker is read by pass 2 ONLY.** The carrier stays in pass 1's collision scans, because it *is* a real node the resolver sees. Over-correcting — hiding it from the scans as well, which reads as the tidier change — produces a **new** wrong binding, and that is graded: `TestSwiftFieldTypeRefs_AliasExtendedInItsOwnFileGetsNoEdge` (`typealias Money = Int` beside `extension Money`) fails under exactly that over-correction (mutant M19).

### What the bind-rate instrument could NOT see

Worth stating plainly, because it is the reason this survived the first round: the corpus resolver harness reports **`total=38 bound=38 dangling=0 wrongName=0 wrongFile=0`** on the *unfixed* code. The carrier has the right name, in the right file, with an admissible kind and subtype. **No bind-rate, dangle-rate or `bug-extractor` instrument can detect this defect class.** Only the declaration-vs-extension distinction detects it, which is why the grading is a fixture intersection and a corpus row-diff rather than a count.

---

## 3. Two deliberate departures beyond the ambiguity rule

**A type parameter shadowed by a same-file type is REFUSED, not pinned as known-wrong.** Arms A and D both ship this over-fire with a comment saying a real fix is expected to break their test. Swift can refuse it for free: `emitSwiftFieldMembers` already holds the owning declaration node, so `struct Box<T>`'s `type_parameters` are in hand at the collection site. It is a fix in the **too-broad** direction, which is the direction this board keeps getting caught by, and it binds when it is wrong — so it never reaches `bug-extractor`.

**A self-reference IS emitted** (`class Node { var next: Node? }` → `Node.next → Node`). Arm D suppresses the equivalent, but its reason does not transfer: Go already ships a struct-anchored `DEPENDS_ON` that declines the self case and arm D matched it. Swift has no adjacent declared-type edge, and arms A and C — which likewise have none — do not suppress it. `CONTAINS` relates owner→field, not field→owner, so this is not a restatement of an existing edge.

---

## 4. Fixture axes

**Varied.**

- **Type-expression shape** — bare, optional, implicitly-unwrapped, array, dictionary-literal, generic ×1 and ×2 args, nested generic, tuple, function type, existential `any`, dotted, duplicate-target tuple. Each asserted against **one** target name so the shape is the only moving part.
- **Target kind** — struct / class / protocol / enum-with-value-set / typealias.
- **Collider kind** — `SCOPE.Enum` must *not* suppress; `SCOPE.Operation` *must* suppress an alias. Each with a positive control of the same declaration form and no collider.
- **Owner declaration form** — `struct`, `class`, generic `struct Box<T>` / `Bag<Element>` / `Pair<A, B>`.
- **Type-parameter name length** — 1 char (`T`) **and** multi-char (`Element`), plus parameter **position** (first and second of `Pair<A, B>`). *Added this round (reviewer's R6): held at one character, a refusal restricted to short names — the obvious wrong way to write this — survived the whole suite.*
- **`SCOPE.`-prefixed vs trimmed kind spelling** — a `SCOPE.Class` collider row. *Added this round (reviewer's R3): the trim in `swiftInComponentAddressFamily` was dead and ungraded, so the claim that deriving family membership beats arm F's hand-list was unexercised.*
- **Which half of a compound guard is wrong** — `{SCOPE.Schema, type_alias}` and `{SCOPE.Component, field}` impostors, one per half of `Kind != "SCOPE.Schema" || Subtype != "field"`. *Added this round (reviewer's R1/R2): the single original impostor differed on **both** halves, so each masked the other and neither was graded.*
- **PRESENCE OF AN `extension` IN THE FIELD'S FILE, CROSSED WITH WHERE THE TYPE IS DECLARED.** ***This is the cell the first cut left empty and where half its edges were wrong.*** All four combinations are now fixtures:

  | | no `extension` here | `extension` here |
  |---|---|---|
  | **declared here** | most tests (bind) | `ExtensionBesideItsTypeIsStillOneNode` — **must still bind** |
  | **declared elsewhere** | `ForeignFileTypeIsNeverATarget` control | `ForeignFileTypeIsNeverATarget` — **must not bind** |
  | **declared in the stdlib** | `PrimitiveFieldsProduceNoEdge` | `ExtensionOfAStdlibTypeIsNeverATarget` — **must not bind** |

- **Duplicate-record origin** — `extension Foo` beside `struct Foo` (Component branch) and a `#if`/`#else` `typealias` pair (alias branch): two records, one kind, one graph node, on each branch.

**Held constant, and named because holding one axis is what makes its sibling look covered.**

- **File of the rival.** Every external fixture is single-file except `ForeignFileTypeIsNeverATarget`, which is a **recall statement, not a grading of the same-file conjuncts** — `Extract` sees one file at a time, so from the external package neither conjunct can ever fire and a mutant deleting either survives that whole file. Both are graded internally in `field_type_refs_scope_6912_test.go`, **in opposite directions**: a foreign-file record must not *suppress* an edge (pass 1) and must not *become* one (pass 2), with rows chosen so a leak in one changes nothing about the rows for the other.
- **The third collider** — a second *component-family* kind, which must suppress a Component — has **no source-level fixture**, because no Swift emit site produces one today and writing Swift that "would" is fabricating an example. Graded at the function level instead, in `TheTwoCollisionScansAreIndependent`, which also pins that the two scans suppress *different* things.
- Repo (one id; cross-repo collision is not modelled and not graded). Language segment. `let` / `weak` / `lazy` / access modifiers are **not** varied — the argument that they cannot reach a decision is in the header, and the fixtures do **not** demonstrate it. Nesting: `walkBody` emits no entity for a type declared inside a class body, so a nested type is not a candidate target in any fixture; that is a property of #4854's walk, not of this pass.

---

## 5. Mutation table — 27 mutants, all at the call site, all compiling, all DEAD

Applied by exact-string replacement with an **asserted occurrence count before and after**; reverted by `cp` from a shasum-verified byte backup, never `git checkout`. Three first-round mutants did not compile (`declared and not used`, `imported and not used`) and were rewritten until they did — a non-compiling mutant proves nothing.

| # | mutation | verdict | killed by |
|---|---|---|---|
| M1 | dotted `user_type` refusal disabled | DEAD | `DottedTypeBindsNeitherHeadNorTail` |
| M2 | do not descend into `type_arguments` | DEAD | `TypeExpressionShapes`, `FieldTypePropertyIsLossy…`, `Dotted…`, `ResolvesToEntityIDs` |
| M3 | Component tier uses **all kinds** (arm D's rule) | DEAD | `EnumValueSetDoesNotSuppressItsComponent`, `TheTwoCollisionScansAreIndependent`, `ResolvesToEntityIDs` |
| M4 | alias tier uses **component family** (arm F's rule) | DEAD | `AliasShadowedByAnOperationWouldDangle…`, `TheTwoCollisionScansAreIndependent` |
| M5 | alias tier removed entirely | DEAD | 6 tests incl. `UnshadowedAliasIsATarget` |
| M6 | component scan counts **records** (arm C's rule) | DEAD | `ExtensionBesideItsTypeIsStillOneNode`, `TwoRecordsOneKindAreOneNodeUnit` |
| M6b | alias scan counts **records** | DEAD | `TwoRecordsOneKindAreOneNodeUnit` (alias branch) |
| M7 | pass-1 `SourceFile != filePath` deleted | DEAD | `TargetsAreScopedToTheRequestedFile` |
| M8 | pass-2 `SourceFile != filePath` deleted | DEAD | `TargetsAreScopedToTheRequestedFile` |
| M9 | type-parameter filter ignored | DEAD | `TypeParameterShadowedByASameFileTypeGetsNoEdge` |
| M10 | subtype allow-list widened to any Component | DEAD | `AllowListRefusesNonTypeComponents` |
| M11 | bare-name ToID instead of the structural address | DEAD | `ResolvesToEntityIDs`, `TargetsAreScoped…` |
| M12 | per-field target dedupe removed | DEAD | `TypeExpressionShapes`, `ResolvesToEntityIDs` |
| M13 | `Kind/Subtype == field` guard removed in attach | DEAD | `OnlyFieldRecordsCarryTheEdge` |
| M14 | stash not deleted from Metadata | DEAD | `StashIsClearedFromMetadata` |
| M15 | candidates taken from the `field_type` **property** | DEAD | 5 tests — the arm's central premise |
| M16 | candidates scoped to the whole `property_declaration` | DEAD | `PropertyWrapperTypeIsNotACandidate` |

**Round 2 — the four the reviewer scored ALIVE, plus M19 and the new fix's own mutants.** Every row re-verified with an asserted occurrence count of exactly 1 before and after, `go vet` clean (a guard is disabled with `if false && <cond>` so no variable goes unused), `-count=1`, and reverted by `cp` from a shasum-verified backup.

| # | mutation | verdict | killed by |
|---|---|---|---|
| **R1** | **Kind half alone** disabled — `false && r.Kind != "SCOPE.Schema"` | **DEAD** | `OnlyFieldRecordsCarryTheEdge` (via the new `KindImpostor` row) |
| **R2** | **Subtype half alone** disabled — `false && r.Subtype != "field"` | **DEAD** | `OnlyFieldRecordsCarryTheEdge` (via the new `AliasImpostor` row) |
| M13 | the whole compound guard disabled | DEAD | `OnlyFieldRecordsCarryTheEdge` |
| **R3** | `SCOPE.`-trim branch disabled in `swiftInComponentAddressFamily` | **DEAD** | `TheTwoCollisionScansAreIndependent` (new `SCOPE.Class` rows) |
| **R6a** | type-parameter refusal restricted to 1-char names | **DEAD** | `TypeParameterShadowedByASameFileTypeGetsNoEdge` (`Bag<Element>`) |
| **R6b** | only the **first** type parameter collected | **DEAD** | `TypeParameterShadowedByASameFileTypeGetsNoEdge` (`Pair<A, B>.second`) |
| **M19** | **over-correction** — carrier also hidden from pass 1's collision scans | **DEAD** | `AliasExtendedInItsOwnFileGetsNoEdge` |
| **EXT** | pass-2 extension refusal disabled | **DEAD** | `ExtensionCarrierIsNeverATarget`, `ForeignFileTypeIsNeverATarget`, `ExtensionOfAStdlibTypeIsNeverATarget`, `AliasExtendedInItsOwnFileGetsNoEdge` |
| **MARK** | `swiftIsExtensionDecl` never consulted at the emit site | **DEAD** | `ForeignFileTypeIsNeverATarget`, `ExtensionOfAStdlibTypeIsNeverATarget`, `AliasExtendedInItsOwnFileGetsNoEdge` |
| **MARK2** | extension marker not deleted from `Metadata` | **DEAD** | `ExtensionCarrierMarkerIsClearedFromMetadata` |

**R1 and R2 are the mutually-masking finding, scored part by part.** The first revision's single `{SCOPE.Component, class}` impostor differs from a field record on *both* halves, so deleting the whole guard died while deleting *either half alone* survived — a compound grades its second half only if its first half is not lethal alone. Two impostors, one wrong on each half, is the fix; both halves are now independently lethal.

**M19 is the over-correction mutant, and it is the reason the marker's scope is narrow.** Hiding the carrier from pass 1 as well as pass 2 is the tidier-looking change and it introduces a *new* wrong binding: `typealias Money = Int` beside `extension Money` would stop being ambiguous, the alias would be admitted, and its component-space address would resolve through `lookupLocationKind` to the **extension carrier** rather than to the alias. It binds. Graded.

**Three of these were ALIVE on the first round and were graded rather than excused** — M6b, M13 and M10 — which is exactly the "too expensive to kill / belt-and-braces" verdict arm D recorded as a mistake. Each cost a handful of lines at the function level.

**Correction: M6b's reachability claim was wrong, and the retraction is narrower than it looks.** I wrote that the only Swift producing two `SCOPE.Schema` records under one name in one file is a duplicate `typealias`, "which swiftc rejects". The corpus this arm measures falsifies the *conclusion*: `vapor/Sources/Vapor/Utilities/VaporSendableMetadataType.swift` is

```swift
#if compiler(>=6.2)
public typealias VaporSendableMetatype = SendableMetatype
#else
public typealias VaporSendableMetatype = Any
#endif
```

— two top-level `typealias` declarations of one name in one file, which the **extractor** sees as two `SCOPE.Schema/type_alias` records. So M6b's alias branch is **source-reachable**, which makes grading it *more* important than the old comment implied, not less. `TestSwiftFieldTypeRefs_ConditionalTypealiasIsStillOneNode` is the source-level form.

What is **not** retracted: swiftc really does reject a genuine duplicate `typealias` in one scope. The `#if`/`#else` form compiles precisely because only one branch is active — it is a *conditional-compilation* shape the extractor flattens, not a duplicate the compiler accepts. The false step was inferring source-unreachability from a true fact about swiftc while the extractor sees both branches. The comment in `field_type_refs_scope_6912_test.go` now says exactly that, and no more.

**All-DEAD is pinning, not truth — and this PR is the proof of that, because the first cut was also all-DEAD while half its edges were wrong.** The truth claim therefore rests on the corpus row-diff in §2a, not on the mutation table: the 38-edge and 19-edge target sets were enumerated and differenced by name and site, and every removed row was confirmed to be a name with **no declaration in that file, only an extension**.

Driving the corpus through `resolve.BuildIndex` → `resolve.ReferencesEmbedded` still reports 19/19 bound and 0 dangling, but §2a records that this instrument reported **38/38 bound, 0 dangling** on the unfixed code too, so it grades address construction and nothing else. `ResolvesToEntityIDs` asserts the **bare** name `Customer` is ambiguous in its fixture, so a bare-name ToID could not pass it and the structural address is genuinely graded.

---

## 6. Gates

- `go test -count=1 ./internal/extractors/swift/` — green (24 new tests; 4 added this round).
- **`go test -count=1 ./cmd/grafel/`** — green (305s), with a **fresh `GRAFEL_HOME` + `GRAFEL_DAEMON_ROOT`** inside the worktree scratch dir; the real `~/.grafel` was not touched. **The digest did not move, and the honest reason is that the digest fixture (`custom_extractor_spans_6118_test.go`) contains no `.swift` file at all** — re-verified this round by grep, 0 matches, not assumed. A new `REFERENCES` edge *does* move `semanticDigest6118` (relationships are hashed as `R <from> -[Kind]-> <to>`); entity **properties and metadata** are not hashed, so the `Metadata` marker this PR adds moves nothing by construction and the net edge change (−19) would have, had any fixture held Swift. A real negative result about the fixture, not a blind instrument.
- `scripts/quality/run.sh --ratchet` — prints *38 gated fixtures held (29 at 100% must-have recall)*, and **`git status --porcelain` is empty of `baseline.json` afterwards**, which is the check that actually distinguishes a hold from a ceiling drop: the script **exits 0 and prints "held" even on a ceiling DROP**, so its exit code is not evidence.
  **The golden set is VACUOUS for this pass and I am not offering the ratchet as evidence.** Re-measured this round: its 6 `.swift` files hold **17 field entities and gain 0 edges** — none declares its field's type in the same file. It even contains the relevant shape (`extension User` beside `struct User` in `swift-swiftui-mini/.../User.swift`) and still gains nothing, because no field there is typed `User`. So the ratchet could not have caught the extension defect and cannot confirm the fix.
- `gofmt -l` clean; `go vet` clean.

## 7. Recall ceiling, honestly

**6.44% of Swift fields gain an edge (19 of 295). A bind rate of 19/19 is not completeness — and §2a shows it is not even soundness.** The floor is the same-file rule, and for Swift it bites harder than for C# or Java: a Swift module is a directory of files and one-type-per-file **is** the convention. Cross-file binding of a bare type name is the #6976/#6369 hazard and this arm does not open it. Also unclaimed: inferred types (`var x = Order()` has no `type_annotation`), computed properties (not stored members, no field entity at all), and every dotted type including the metatype.

**Corpus caveat:** 234 `.swift` files in **3** repos, and 219 of them are one repo (vapor). The mechanism transfers; the 6.44% is vapor's number more than Swift's. #6726's 99.3% is their corpus and no claim is made that it moves — Swift is not one of their four languages.

## 8. Known, unclosed, and left as an invariant rather than a guarantee

`internal/custom/swift/swiftui.go:288` mints a **bare-named `SCOPE.Component`** from an `@ObservedObject`/`@StateObject`/`@EnvironmentObject` **property name**, in the same file, on a lane this pass cannot see.

**Correction: the hazard is CONFLATION, not ambiguity — the earlier version of this section had the mechanism wrong.** That site emits `SCOPE.Component`, the **same** Kind as the declaration it could collide with, and entity IDs are re-derived downstream as `graph.EntityID(repo, Kind, Name, SourceFile)` (`internal/extractors/incremental.go:2068`, which deliberately ignores the emitted `r.ID`). A same-file, same-name `SCOPE.Component` from the custom lane therefore carries the **identical id**; `buildSymbolIndex` marks a collision only on `existing != e.ID` (`refs.go:1308`). So there is **no `ambigLocation` entry, no blanked `.base` key and no dangle** — the two records **merge into one node**. The residual hazard is that a property and a type would share a node, which is a different and lesser complaint than the edge failing to bind.

**This also settles why the fixture refusal in §4 was right, and the refusal stands.** The genuinely unseen hazard is a custom-lane record carrying a **second component-family KIND** under a declared type's name — and that site does not produce one. Writing Swift that "would" produce one is fabricating an example, so there is still no source-level fixture for it and it is graded at the function level in `TheTwoCollisionScansAreIndependent` instead. Fixing the derivation does not create a fixture obligation.

**Measured incidence of the conflation shape: 0** on the corpus and 0 on the golden set (whose one `@StateObject var viewModel: UserListViewModel` shares no name with a same-file declaration). That is a corpus fact, not a language fact, and the header states the invariant to preserve rather than relying on the accident.

Refs #6912, #6726, #7038, #7047, #6976, #6369, #492.

### DIFF
--- a/internal/extractors/swift/field_members.go
+++ b/internal/extractors/swift/field_members.go
@@ -38,6 +38,10 @@ func emitSwiftFieldMembers(
 
 	var fields []types.EntityRecord
 	seen := make(map[string]bool)
+	// #6912 arm G — the owner's generic parameters shadow any same-named type
+	// declared in the file. Collected here because `node` (the owning
+	// declaration) is only in hand at this site.
+	typeParams := swiftTypeParameterNames(node, src)
 
 	if body != nil {
 		for i := 0; i < int(body.ChildCount()); i++ {
@@ -56,8 +60,17 @@ func emitSwiftFieldMembers(
 			}
 			seen[name] = true
 			typ := ""
+			// #6912 arm G — `cands` is the field→declared-type edge's input, and
+			// it is deliberately NOT derived from `typ`. firstDescendantText
+			// returns the FIRST pre-order type_identifier, so `typ` is
+			// "Dictionary" for `Dictionary<String, Order>` and "Foundation" for
+			// `Foundation.Data`: the information the edge needs is already gone
+			// by the time the property is written. See field_type_refs.go's
+			// header for the probe and for what the property is left as.
+			var cands []string
 			if ta := firstChildOfType(ch, "type_annotation"); ta != nil {
 				typ = firstDescendantText(ta, src, "type_identifier")
+				cands = swiftFieldTypeCandidates(ta, src, typeParams)
 			}
 			dotted := ownerName + "." + name
 			sig := name
@@ -80,7 +93,10 @@ func emitSwiftFieldMembers(
 					"field_type":   typ,
 					"parent_class": ownerName,
 				},
-				Metadata:           map[string]interface{}{"subtype": "field", "owner": ownerName},
+				Metadata: map[string]interface{}{
+					"subtype": "field", "owner": ownerName,
+					swiftFieldTypeRefsMetaKey: cands,
+				},
 				EnrichmentRequired: false,
 			})
 		}
--- a/internal/extractors/swift/field_type_refs.go
+++ b/internal/extractors/swift/field_type_refs.go
@@ -1,0 +1,501 @@
+package swift
+
+import (
+	"strings"
+
+	"github.com/cajasmota/grafel/internal/extractor"
+	"github.com/cajasmota/grafel/internal/treesitter/ts"
+	"github.com/cajasmota/grafel/internal/types"
+)
+
+// field_type_refs.go — the field→declared-type edge for Swift (issue #6912,
+// arm G). Arm A is internal/extractors/csharp/field_type_refs.go (#6984),
+// arm B internal/extractors/proto (#6991), arm C internal/extractors/rust
+// (#7000), arm D internal/extractors/golang (#7036), arm F
+// internal/extractors/java (#7039), arm E internal/extractors/fsharp (#7040).
+//
+// THE GAP. emitSwiftFieldMembers (#4854) mints one SCOPE.Schema/field record
+// per stored property and records the declared type ONLY as the unhashed
+// `field_type` property. Nothing links the field to the ENTITY of that type,
+// so every Swift field is a leaf on its outbound side — the population
+// contributor #6726 measures at 99.3% orphan for SCOPE.Schema. Swift fields are
+// minted as SCOPE.Schema (field_members.go), so they are IN that population,
+// not adjacent to it.
+//
+// THE TRAP THIS ARM EXISTS TO CLOSE, AND WHY THE PROPERTY IS NOT THE INPUT.
+// `field_type` is firstDescendantText(type_annotation, "type_identifier") —
+// the FIRST PRE-ORDER type_identifier under the annotation. Probed against the
+// real grammar rather than read:
+//
+//	var d: Dictionary<String, Order>   user_type[ type_identifier "Dictionary",
+//	                                              type_arguments[…"Order"…] ]
+//	                                   → field_type == "Dictionary"
+//	var j: Set<Order>                  → field_type == "Set"
+//	var k: Result<Order, MyErr>        → field_type == "Result"
+//	var c: [Order]                     array_type[ user_type "Order" ]
+//	                                   → field_type == "Order"  (by luck of order)
+//	var b: Order?                      → field_type == "Order"  (by luck of order)
+//	var h: Foundation.Data             user_type[ tid "Foundation", ".", tid "Data" ]
+//	                                   → field_type == "Foundation"
+//
+// So for every generic wrapper the argument is ALREADY GONE by the time the
+// property is written, and for a module-qualified type the property holds the
+// MODULE. Reading `field_type` and emitting an edge would ship exactly the
+// `List<Customer>` → `List` binding arm A went out of its way to forbid, and a
+// `Foundation.Data` field would bind to a same-file `struct Foundation`.
+//
+// This arm therefore takes arm C's technique — stash the candidates FROM THE
+// AST at the emit site — and not arm E's (tokenise the type string), because
+// F#'s `member_type` is the declared type verbatim and Swift's `field_type` is
+// not. `field_type` is left exactly as it is: fixing it would change entity
+// Properties and Signature for every Swift field in the graph, which is a
+// separate change with its own blast radius. Its lossiness is pinned by
+// TestSwiftFieldTypeRefs_FieldTypePropertyIsLossyForGenerics so the next reader
+// meets the trap as a failing expectation rather than as prose.
+//
+// ==========================================================================
+// THE AMBIGUITY RULE — SWIFT NEEDS BOTH SHIPPED RULES, SPLIT BY TARGET KIND
+// ==========================================================================
+//
+// The standing instruction on #6912 is: COUNT THE KINDS THAT THE RESOLVER TIER
+// WHICH RESOLVES YOUR ADDRESS ACTUALLY WEIGHS. Arm D (go) counts all kinds,
+// arm F (java) counts within the component address family, arm C (rust) counts
+// records and is wrong (#7038). Swift is the first arm where the answer is not
+// one of those three, because SWIFT EMITS TARGETS IN TWO DIFFERENT KIND SPACES
+// and the same address resolves them through two DIFFERENT tiers.
+//
+// The address is BuildComponentStructuralRef → `scope:component:class:swift:
+// <file>:<Name>`, so lookupStructural (internal/resolve/refs.go:2982) calls
+// lookupLocationKind(file, name, structuralKindFamilies("component")) =
+// componentKindFamily. Driven through the real resolver, not inferred:
+//
+//   - TARGET IS SCOPE.Component (class / struct / enum / actor / protocol).
+//     byLocationKind indexes an entity under BOTH its Kind and its SCOPE-
+//     trimmed Kind (refs.go:1207-1210), so a SCOPE.Component lands under
+//     "SCOPE.Component" AND "Component", and "Component" is in
+//     componentKindFamily. lookupLocationKind's tier-2 pass therefore MATCHES
+//     and returns before ambigLocation is ever consulted. Only a second kind
+//     THAT IS ALSO IN THAT FAMILY can make uniqueMatchInFamily see two distinct
+//     ids and fail. This is arm F's rule, and Swift needs it for the same
+//     reason java did: an `enum Status` emits a SCOPE.Component AND a
+//     SCOPE.Enum value-set of the SAME NAME IN THE SAME FILE (swift.go:116-120,
+//     types.go buildEnumValueSet). SCOPE.Enum trims to "Enum", which is in NO
+//     family, so the value-set cannot shadow the Component — and arm D's
+//     all-kinds rule would refuse EVERY Swift enum target.
+//
+//   - TARGET IS SCOPE.Schema/type_alias (`typealias Money = Double`,
+//     types.go:154-165). "SCOPE.Schema" trims to "Schema", which is NOT in
+//     componentKindFamily, so lookupLocationKind MISSES ENTIRELY and the ref
+//     falls through to the kind-agnostic pair at refs.go:3004-3012:
+//     ambigLocation first, then byLocation. ambigLocation is set whenever a
+//     (file, name) meets a SECOND DISTINCT ENTITY ID (refs.go:1301-1316) —
+//     ANY kind at all. So for an alias target the count must span every kind.
+//     This is arm D's rule, and arm F's rule applied here would emit an edge
+//     that DANGLES: `typealias Tier` beside a top-level `func Tier()` is
+//     Schema + Operation, invisible to a component-family count, and
+//     statusAmbiguous at the resolver. Graded by
+//     TestSwiftFieldTypeRefs_AliasShadowedByAnOperationWouldDangleUnderArmFsRule,
+//     which drives the real resolver rather than asserting the emit decision.
+//
+// The unit is the KIND rather than the RECORD in both halves, mirroring the
+// resolver: graph.EntityID hashes (repo, Kind, Name, SourceFile) with Subtype
+// EXCLUDED, and buildSymbolIndex marks a collision only on `existing != e.ID`
+// (refs.go:1308). Two same-file records sharing a Kind are ONE node and must
+// not be counted twice. Arm C counts records and is wrong for exactly this
+// reason (#7038) — its rule is not copied here.
+//
+// WHY NOT SIMPLY DROP type_alias TARGETS AND USE ARM F'S RULE UNCHANGED. That
+// is the arm-F counterfactual and it is a pure recall loss: arm D measured 19%
+// of Go's edges on exactly this shape after nearly inheriting arm C's
+// Component-only guard. The measured Swift figure is in the PR body. A Swift
+// typealias is a first-class nameable type and a field declared with one is
+// not a different proposition from a field declared with a struct.
+//
+// ==========================================================================
+// WHAT THIS PASS IS NOT
+// ==========================================================================
+//
+//   - SAME FILE ONLY. A field whose type is declared in another file gets
+//     nothing. Binding a bare type name across files is the #6976/#6369
+//     hazard and every arm on this issue rests on refusing it. For Swift this
+//     is the dominant miss: a Swift module is a directory of files and
+//     one-type-per-file IS the convention, so the recall ceiling here is lower
+//     than C#'s or Java's. Both directions of the same-file conjuncts are
+//     graded — a foreign-file record must not SUPPRESS an edge (pass 1) and
+//     must not BECOME one (pass 2) — because grading one is precisely what
+//     makes the other look covered.
+//
+//   - A DOTTED TYPE GETS NO EDGE AT ALL, head or tail. tree-sitter-swift gives
+//     `Foundation.Data` and `Order.Inner` and the metatype `Order.Type` one
+//     `user_type` node carrying TWO type_identifier children around an
+//     anonymous ".". Taking the tail would bind `Order.Inner` to a same-file
+//     `Inner`; taking the head would bind `Order.Type` to `Order` but also
+//     `Foundation.Data` to a same-file `Foundation`. Both are wrong bindings,
+//     and a wrong binding never reaches bug-extractor precisely because it
+//     binds. So the head-and-tail of a dotted user_type are both refused —
+//     which costs the metatype and the module-qualified same-module type, the
+//     honest price of not guessing. Type ARGUMENTS are still descended into,
+//     so `Swift.Array<Order>` yields `Order`. Pinned by
+//     TestSwiftFieldTypeRefs_DottedTypeBindsNeitherHeadNorTail.
+//
+//   - AN INFERRED TYPE GETS NOTHING. `var p = Order()` has no type_annotation;
+//     the initializer expression is not read. A computed property is not a
+//     stored member and has no field entity at all (#4854).
+//
+//   - PROPERTY-WRAPPER AND ATTRIBUTE TYPES ARE NOT CANDIDATES. `@Published var
+//     s: Order` shapes `Published` as a `user_type` under a `modifiers >
+//     attribute` sibling of the type_annotation. Candidates are collected from
+//     the type_annotation subtree ONLY, so the wrapper is out of reach by
+//     construction rather than by a name filter. Graded by
+//     TestSwiftFieldTypeRefs_PropertyWrapperTypeIsNotACandidate, which declares
+//     a same-file `struct Published` so the assertion can fail.
+//
+//   - RECORDS THIS PASS CANNOT SEE. The collision scan spans the records this
+//     extractor has built, which is every record `Extract` returns. It does NOT
+//     span the custom lane: internal/custom/swift/swiftui.go:288 mints a
+//     BARE-NAMED SCOPE.Component from an @ObservedObject/@StateObject property
+//     NAME in the same file. That kind IS in the component address family, so a
+//     property named after a same-file type would make the target ambiguous at
+//     the resolver and dangle an edge this pass believed safe. Measured
+//     incidence on the corpus is in the PR body; the invariant to preserve is
+//     that the custom lane does not mint a bare Component sharing a declared
+//     type's name, not that today's corpus happens not to.
+//
+// A NON-BINDING EDGE IS WORSE THAN NO EDGE (#6912's hard requirement): an
+// unresolved stub is kept verbatim, dangles, and is classified bug-extractor,
+// so every miss is a full hit on the disposition denominator. That is why the
+// rules above refuse rather than guess, and why
+// TestSwiftFieldTypeRefs_ResolvesToEntityIDs drives every emitted edge through
+// resolve.BuildIndex → resolve.ReferencesEmbedded and names the entity each one
+// lands on.
+
+// swiftFieldTypeRefsMetaKey is the per-field stash written by
+// emitSwiftFieldMembers and consumed — and deleted — by
+// attachSwiftFieldTypeRefs.
+//
+// The candidates cannot become edges during the walk: `struct Order { var b:
+// Customer }` may be declared BEFORE `struct Customer` in the same file, so the
+// in-file declaration set is complete only once the walk has finished, and the
+// enum value-sets the collision scan must see are appended as the walk goes.
+const swiftFieldTypeRefsMetaKey = "field_type_refs"
+
+// swiftFieldTargetRefKind is the value of the `ref_kind` edge property. It
+// matches arm A's csFieldTargetRefKind, arm B's protoFieldTargetRefKind, arm
+// C's rustFieldTargetRefKind, arm D's goFieldTargetRefKind, arm F's
+// javaFieldTargetRefKind and the custom lane's referencesClassEdge verbatim —
+// the discriminator is what makes the edge queryable as a field→declared-type
+// edge across languages, so it must be one string.
+const swiftFieldTargetRefKind = "field_target_type"
+
+// swiftComponentFamilyKeys mirrors internal/resolve's componentKindFamily
+// (refs.go:2341-2348) exactly — including the deliberate ABSENCE of
+// SCOPE.Service (#6459/#6492) and of any Schema/Enum spelling.
+var swiftComponentFamilyKeys = map[string]bool{
+	"Component": true,
+	"Class":     true,
+	"View":      true,
+	"Model":     true,
+	// The SCOPE.* spellings the family lists literally.
+	"SCOPE.Component": true,
+	"SCOPE.View":      true,
+	"SCOPE.Model":     true,
+}
+
+// swiftInComponentAddressFamily reports whether an entity of this Kind is
+// WEIGHED by lookupLocationKind under componentKindFamily.
+//
+// The trim step is not defensive padding: buildSymbolIndex indexes every entity
+// under BOTH its Kind and its SCOPE-trimmed Kind (refs.go:1207-1210), so a
+// hypothetical "SCOPE.Class" entity lands under the key "Class", which the
+// family lists. Deriving membership instead of hand-listing the spellings is
+// what keeps this in step with the resolver rather than with a snapshot of it.
+func swiftInComponentAddressFamily(kind string) bool {
+	if swiftComponentFamilyKeys[kind] {
+		return true
+	}
+	if t := strings.TrimPrefix(kind, "SCOPE."); t != kind && swiftComponentFamilyKeys[t] {
+		return true
+	}
+	return false
+}
+
+// swiftTypeParameterNames returns the names bound by a declaration's
+// `type_parameters` child — `struct Box<T>` → {"T"}.
+//
+// Arms A and D both ship a KNOWN-WRONG over-fire here: a type parameter named
+// after a same-file type binds the field to that type. Swift can refuse it for
+// free because emitSwiftFieldMembers already holds the owning declaration node,
+// so the shadowing scope is in hand at the collection site rather than needing
+// a scope tracker. Doing it is a deliberate departure from those arms and it is
+// in the direction this board keeps getting caught by — too broad, and binding,
+// so invisible. Graded by
+// TestSwiftFieldTypeRefs_TypeParameterShadowedByASameFileTypeGetsNoEdge.
+func swiftTypeParameterNames(node ts.Node, src []byte) map[string]bool {
+	if node == nil {
+		return nil
+	}
+	var out map[string]bool
+	for i := 0; i < int(node.ChildCount()); i++ {
+		tp := node.Child(i)
+		if tp == nil || tp.Type() != "type_parameters" {
+			continue
+		}
+		for j := 0; j < int(tp.ChildCount()); j++ {
+			p := tp.Child(j)
+			if p == nil || p.Type() != "type_parameter" {
+				continue
+			}
+			if n := firstDescendantText(p, src, "type_identifier"); n != "" {
+				if out == nil {
+					out = make(map[string]bool)
+				}
+				out[n] = true
+			}
+		}
+	}
+	return out
+}
+
+// swiftFieldTypeCandidates returns every bare type name written in a field's
+// type_annotation, in source order and without deduplication, minus the owner's
+// type parameters.
+//
+// Node handling, each shape confirmed against the real grammar by CST probe:
+//
+//   - `user_type` — the nominal type. When it is DOTTED (an anonymous "." child
+//     stands between two type_identifier children) neither segment is taken;
+//     see the header block. Otherwise its single direct type_identifier child is
+//     the candidate. Either way any `type_arguments` child IS descended into, so
+//     `Dictionary<String, Order>` yields String and Order and NOT Dictionary,
+//     and `Swift.Array<Order>` yields Order.
+//   - everything else recurses. That covers `optional_type` (`Order?`),
+//     `array_type` (`[Order]`), `dictionary_type` (`[String: Order]`),
+//     `tuple_type` (`(Int, Order)`), `function_type` (`(Order) -> Void`),
+//     `existential_type` (`any Shape`) and `opaque_type` (`some Shape`) with no
+//     wrapper list to keep in sync.
+//
+// A SWIFT PRIMITIVE NEEDS NO BLOCKLIST, and — as in arm D and unlike arm B — a
+// blocklist would be actively wrong. `Int`, `String` and `Bool` are ordinary
+// stdlib nominal types, not grammar keywords: the parser gives them the same
+// `user_type > type_identifier` shape as a user type. They are refused by the
+// in-file declaration check instead, which is the correct order of operations,
+// because a file that declares `struct Int` genuinely does mean THAT type in
+// field position. Both directions graded:
+// TestSwiftFieldTypeRefs_PrimitiveFieldsProduceNoEdge and
+// TestSwiftFieldTypeRefs_ShadowedStdlibNameIsATarget.
+func swiftFieldTypeCandidates(ta ts.Node, src []byte, typeParams map[string]bool) []string {
+	var out []string
+	var walkType func(n ts.Node)
+	walkType = func(n ts.Node) {
+		if n == nil {
+			return
+		}
+		if n.Type() == "user_type" {
+			dotted := false
+			for i := 0; i < int(n.ChildCount()); i++ {
+				if c := n.Child(i); c != nil && c.Type() == "." {
+					dotted = true
+					break
+				}
+			}
+			for i := 0; i < int(n.ChildCount()); i++ {
+				c := n.Child(i)
+				if c == nil {
+					continue
+				}
+				switch c.Type() {
+				case "type_identifier":
+					if dotted {
+						continue
+					}
+					name := nodeTextSwift(c, src)
+					if name != "" && !typeParams[name] {
+						out = append(out, name)
+					}
+				case "type_arguments":
+					walkType(c)
+				}
+			}
+			return
+		}
+		for i := 0; i < int(n.ChildCount()); i++ {
+			walkType(n.Child(i))
+		}
+	}
+	walkType(ta)
+	return out
+}
+
+// nodeTextSwift returns a node's source text.
+func nodeTextSwift(n ts.Node, src []byte) string {
+	if n == nil {
+		return ""
+	}
+	return string(src[n.StartByte():n.EndByte()])
+}
+
+// swiftFieldTypeTarget is one in-file type declaration a field can point at:
+// the structural ToID that binds to it and the bare name for `target_type`.
+type swiftFieldTypeTarget struct {
+	toID string
+	name string
+}
+
+// swiftInFileTypeTargets indexes every type DECLARED IN THIS FILE that a
+// field-type edge may address, keyed by bare name, applying the split
+// ambiguity rule derived in the header block.
+//
+// Pass 1 builds two independent collision counts because the two admitted
+// target kinds resolve through two different resolver tiers:
+//
+//	famKinds[name]  — distinct Kinds in the COMPONENT ADDRESS FAMILY. Governs a
+//	                  SCOPE.Component target, which lookupLocationKind resolves
+//	                  before ambigLocation is consulted (arm F's rule).
+//	allKinds[name]  — distinct Kinds, full stop. Governs a SCOPE.Schema/
+//	                  type_alias target, which lookupLocationKind cannot see at
+//	                  all, so it falls through to ambigLocation (arm D's rule).
+//
+// Pass 2's allow-list is written on Swift's own emit sites rather than
+// inherited:
+//
+//	SCOPE.Component + class|struct|enum|actor|protocol → admitted
+//	SCOPE.Schema    + type_alias                       → admitted
+//
+// Everything else is refused. What that actually excludes, and how reachable
+// each is, stated as measured rather than as reassurance:
+//
+//   - SCOPE.Enum (the value-set beside every `enum`) — REACHABLE AND LOAD-
+//     BEARING. Admitting it would address the value-set with a component-space
+//     ref that the component family cannot resolve to it, so the edge would
+//     have to fall through to byLocation, where the same-named Component makes
+//     it ambiguous. It is refused as a TARGET while still being invisible to
+//     the component-family COUNT — the two halves of arm F's rule.
+//   - SCOPE.Schema/field — refused so a field never targets a field. This one
+//     is ALSO unreachable by name shape: a field record's Name is dotted
+//     (`Order.buyer`) and a candidate is a single type_identifier, which cannot
+//     contain a dot. Refused anyway, claimed as nothing more.
+//   - SCOPE.Component/module (the import carrier) — refused by subtype, and
+//     ALSO unreachable by name shape, because buildImport namespaces the
+//     carrier as `<file>::import::<module>` precisely so it cannot match a bare
+//     Swift type identifier (#492). Go had no such protection and its import
+//     placeholder was the one record its allow-list genuinely had to stop; the
+//     Swift equivalent is already stopped upstream. Pinned by
+//     TestSwiftFieldTypeRefs_ImportCarrierNameCannotCollide so a rename of the
+//     carrier surfaces here.
+//   - SCOPE.Component/file (#577) — refused by subtype; its Name is the file
+//     path.
+//   - SCOPE.Operation — refused by kind. Not reachable as a wrong TARGET (its
+//     kind is not admitted) but very much reachable as a COLLIDER: a top-level
+//     `func Tier()` is what makes a same-named typealias dangle, which is the
+//     whole reason the alias arm counts all kinds.
+func swiftInFileTypeTargets(records []types.EntityRecord, filePath string) map[string]swiftFieldTypeTarget {
+	famKinds := make(map[string]map[string]bool)
+	allKinds := make(map[string]map[string]bool)
+	for i := range records {
+		r := &records[i]
+		if r.SourceFile != filePath || r.Name == "" {
+			continue
+		}
+		if allKinds[r.Name] == nil {
+			allKinds[r.Name] = make(map[string]bool)
+		}
+		allKinds[r.Name][r.Kind] = true
+		if swiftInComponentAddressFamily(r.Kind) {
+			if famKinds[r.Name] == nil {
+				famKinds[r.Name] = make(map[string]bool)
+			}
+			famKinds[r.Name][r.Kind] = true
+		}
+	}
+
+	targets := make(map[string]swiftFieldTypeTarget)
+	for i := range records {
+		r := &records[i]
+		if r.SourceFile != filePath || r.Name == "" {
+			continue
+		}
+		switch {
+		case r.Kind == "SCOPE.Component":
+			switch r.Subtype {
+			case "class", "struct", "enum", "actor", "protocol":
+			default:
+				continue
+			}
+			// Component tier: only a rival IN THE SAME ADDRESS FAMILY can
+			// break uniqueMatchInFamily. A SCOPE.Enum value-set cannot.
+			if len(famKinds[r.Name]) > 1 {
+				continue
+			}
+		case r.Kind == "SCOPE.Schema" && r.Subtype == "type_alias":
+			// Alias tier: the component family cannot see this entity at all,
+			// so the ref falls through to ambigLocation, which counts EVERY
+			// kind. A same-named Component is included in that count, which is
+			// also what keeps a Component target from being clobbered here.
+			if len(allKinds[r.Name]) > 1 {
+				continue
+			}
+		default:
+			continue
+		}
+		targets[r.Name] = swiftFieldTypeTarget{
+			toID: extractor.BuildComponentStructuralRef("swift", filePath, r.Name),
+			name: r.Name,
+		}
+	}
+	return targets
+}
+
+// attachSwiftFieldTypeRefs appends one REFERENCES edge per (field, in-file
+// declared type) pair and clears the stash it consumed.
+//
+// FromID is left empty so graph assembly anchors the edge on the field record
+// that carries it — the Django precedent at
+// internal/extractors/python/django_relational.go:246-252, and arms B, C, D
+// and F. (Hibernate sets an explicit structural FromID because it hangs its
+// edge off its own parallel SCOPE.Component node; that is a different shape and
+// #6912's body was corrected on the point.)
+//
+// A SELF-REFERENCE IS EMITTED. `class Node { var next: Node? }` yields
+// `Node.next → Node`. Arm D suppresses the equivalent, but its reason does not
+// transfer: Go already ships a struct-anchored DEPENDS_ON that declines the
+// self case, and arm D matched it. Swift has no adjacent declared-type edge to
+// stay consistent with, and arms A and C — which likewise have none — do not
+// suppress it. The CONTAINS edge relates owner→field, not field→owner, so the
+// self edge is not a restatement of one that exists. Graded by
+// TestSwiftFieldTypeRefs_SelfReferentialFieldGetsAnEdge.
+//
+// Relationships is APPENDED to, never assigned: a Swift field record carries no
+// outbound edge today, but assigning would silently clobber one added later.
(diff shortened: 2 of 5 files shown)


cajasmota (commented):
## Adversarial review — arm G (swift), head `0977ee1`

Independent worktree, own `GOCACHE`, own copy of the swift corpus. I re-derived every number rather than reading it. Almost everything in this body held up; one thing did not, and it is 50% of the shipped edges.

---

# BLOCKING — `extension Foo` mints a `SCOPE.Component` for a type declared in ANOTHER file, and the allow-list admits it. **19 of the 38 corpus edges (50%) are this.**

`swiftDeclSubtype` (swift.go:240-259) switches on the anonymous declaration keyword and **falls through to `"class"` for anything it does not recognise — including `extension`**. So `extension String { … }` in `DirectoryConfiguration.swift` mints

```
SCOPE.Component/class : String @ Sources/Vapor/Utilities/DirectoryConfiguration.swift
```

and `swiftInFileTypeTargets`'s pass-2 allow-list (`SCOPE.Component` + `class|struct|enum|actor|protocol`) cannot tell it from a real `class String` declared there. `TestSwiftFieldTypeRefs_ExtensionBesideItsTypeIsStillOneNode` uses exactly this conflation to make its record-vs-kind point — but it only ever exercises the case where the type **is also declared in the file**. The far larger population is `extension Foo` with **no** `Foo` declaration in that file.

Measured on the same corpus this PR measures (my own copy, 234 files / 295 fields / **38 edges** — reproduced exactly). For each of the 38 rows I asked whether the target name has any `struct|class|enum|protocol|actor|typealias <Name>` in the same file:

```
19 of 38 edges have NO declaration of the target in that file — only `extension <Name>`:

  DirectoryConfiguration.publicDirectory   -> String        (extension String)
  DirectoryConfiguration.resourcesDirectory-> String
  DirectoryConfiguration.viewsDirectory    -> String
  DirectoryConfiguration.workingDirectory  -> String
  FileMiddleware.defaultFile               -> String        (fileprivate extension String)
  FileMiddleware.publicDirectory           -> String
  RFC1123DateCache.cachedTimestampAndComponents -> tm       (extension tm: @retroactive …  — a C struct)
  Base32.`default` / canonical / lowercasedCanonical / relaxed -> BaseNEncoding   (extension BaseNEncoding)
  Base64.`default` / bcrypt / canonical    -> BaseNEncoding
  AsyncPasswordHasher.hasher               -> PasswordHasher
  EventLoopHTTPClient.http                 -> HTTPClient
  FileIO.request                           -> Request
  HTTPRoutesGroup.root                     -> RoutesBuilder
  MiddlewareGroup.root                     -> RoutesBuilder
```

And it is not merely cosmetic — the edge lands on the **wrong node**. Two-file probe, real resolver, no fabrication (this is `FileIO.swift`'s literal shape):

```
Sources/App/Request.swift : public struct Request { … }
Sources/App/FileIO.swift  : extension Request { … }
                            public extension String { … }
                            struct FileIO { var request: Request; var name: String }

records:  Request SCOPE.Component/struct @ Sources/App/Request.swift
          Request SCOPE.Component/class  @ Sources/App/FileIO.swift     <- the extension carrier
          String  SCOPE.Component/class  @ Sources/App/FileIO.swift

BOUND FileIO.request -> 90111c4a678d3857 [Sources/App/FileIO.swift:SCOPE.Component/class:Request]
BOUND FileIO.name    -> 62f5a1c84571f9d9 [Sources/App/FileIO.swift:SCOPE.Component/class:String]
```

The field binds to the **extension carrier**, a distinct entity ID from the real declaration. The real `Request` gets no inbound edge.

What this falsifies, in the PR's own words:

1. **"SAME FILE ONLY. A field whose type is declared in another file gets nothing."** False for 19 of 38. `PasswordHasher`, `BaseNEncoding`, `HTTPClient`, `Request`, `RoutesBuilder`, `String`, `tm` are all declared elsewhere (or in the stdlib / libc).
2. **`ForeignFileTypeIsNeverATarget`** passes only because its rival file's type is not *extended* in the field's file. Add `extension Customer {}` to that fixture's `Models.swift` and `Order.buyer` starts binding. That is the fixture-neighbour defect this board keeps hitting: the *cross-file* axis is varied while the *extension* axis is held constant, and vice-versa in `ExtensionBesideItsTypeIsStillOneNode`. Their intersection is the uncovered cell, and it is where half the corpus lives.
3. **"A SWIFT PRIMITIVE NEEDS NO BLOCKLIST … refused by the in-file declaration check, which is the correct order of operations, because a file that declares `struct Int` genuinely does mean THAT type."** `extension String` is not a declaration of `String`, and the justification does not reach it. 6 of 38 edges are `-> String`. The same `var name: String` gets an edge in `DirectoryConfiguration.swift` and nothing in the file next door. `PrimitiveFieldsProduceNoEdge` holds only because its fixture has no `extension Int`; `ShadowedStdlibNameIsATarget` varies the *shadowing* axis using `struct Int` and holds the *extension* form constant.
4. **"12.88% of Swift fields gain an edge"** — half of that is this path, undisclosed.
5. This is a **binding** over-fire, so `bug-extractor` never sees it — the exact direction the PR refuses to ship for type parameters ("it is a fix in the too-broad direction, which is the direction this board keeps getting caught by, and it binds when it is wrong — so it never reaches `bug-extractor`"). Applied consistently, that argument decides this case too.

It also reframes §2's arm-C counterfactual. The 5 arm-C rows are `struct Foo` **beside** `extension Foo`; the dominant extension population — 19 rows — is `extension Foo` **without** `struct Foo`, which a record count would not have caught either. The counterfactual is correct; the reading of the corpus around it missed the bigger shape.

**Ask (either is fine, but not silence):**
- **(a) Refuse it.** The declaration keyword is anonymous-child data you already read at the emit site — the same "it is in hand, so refuse it for free" argument you used for `type_parameters`. Either mint `Subtype: "extension"` in #4854 and drop it from the allow-list, or build the target set from non-extension declaration names. Then re-run the 38 and state the new number.
- **(b) Keep it deliberately** — but then say so in the header and body ("a same-file extension of a foreign type IS an admitted target, and here is why"), correct the three claims above, put the 19/38 figure in §7 next to the 12.88%, and add a fixture with `extension Request` in a file that does not declare `Request`. Note that (b) means a field binds to a per-file extension node rather than to the type, which I think you will not want.

---

# What I could not break

Everything else in this PR survived, including the parts I was briefed to attack hardest. Stated explicitly so the next reader does not re-do it.

**The two-tier claim — verified in both directions, from the resolver source.**
- `swiftComponentFamilyKeys` matches `componentKindFamily` (refs.go:2341-2348) verbatim, including the deliberate absence of `SCOPE.Service`.
- Component tier: a `SCOPE.Component` entity writes `.base` under `"SCOPE.Component"` **and** `"Component"` (refs.go:1207-1210). `lookupLocationKind`'s tier-1 (`.real`, `includePlaceholders=false`) skips every `SCOPE.*` family member and misses; tier-2 over `.base` hits both keys with **one** ID, so `uniqueMatchInFamily` returns and `refs.go:2982` returns **before** `ambigLocation` at 3004. Only a second family kind can break it. Arm F's rule, correctly derived.
- `SCOPE.Enum` trims to `"Enum"`, which is in no family — confirmed against `structuralKindFamilies`. It cannot shadow an enum's own Component. Confirmed end-to-end: `Order.status` binds to `SCOPE.Component/enum:Status` with the value-set present (premise asserted first in the fixture).
- Alias tier: `"SCOPE.Schema"` → `"Schema"`, in no family, `lookupLocationKind` misses entirely, falls to `ambigLocation` (set on any second distinct ID at (file, name) — any kind) then `byLocation`. `ResolvesToEntityIDs` shows `Order.tier => SCOPE.Schema/type_alias:Tier` binding through the real resolver, so the tier is reachable, not decorative.

**All four counterfactuals re-derived independently — exact, row for row.**

| rule | my edges | PR |
|---|---|---|
| this arm | **38** | 38 |
| arm D (component tier → all kinds) | **33 (−5, −13.2%)** | −5, same 5 OTP/URLEncodedFormData enum rows |
| arm C (family **records**, Components only) | **33 (−5, −13.2%)** | −5, same 5 rows |
| arm F (alias tier removed) | **38 (0)** | 0 |

The arm-F zero is **genuinely corpus-relative**, not a tier that cannot fire: the corpus emits exactly **5** `SCOPE.Schema/type_alias` records (`CodingKeyRepresentable`, `HTTPStatus`, `ValidationKey`, `VaporSendableMetatype` ×2) across 4 distinct names, none of them a same-file field target. Keeping the tier and saying why was the right call.

**The `field_type` lossiness — real in both directions**, measured on real CST:
`Order.h: Foundation.Data` → `field_type == "Foundation"`; `Order.m: Order.Type` → `field_type == "Order"`. And the arm's AST path does **not** reintroduce the bug from the other side: neither yields an edge — no bare `Data`, no `Foundation`, no `Order`. That was the wrong-binding trap I was sent to find and it is not there.

**The type-parameter refusal is complete and does not over-refuse.** Beyond the shipped fixture:

```
struct Element{}  struct Box<Element>{ var item: Element }  -> no edge     (multi-char param)
                  struct Other       { var e: Element }     -> [Element]   (sibling unaffected)
struct A{}        struct Pair<A,B>   { var a: A; var b: B } -> no edge / no edge
struct T{} protocol Shipper{} struct Box<T: Shipper>{ var item:T; var s:Shipper } -> no edge / [Shipper]
```

**The mutability/access axis really is inert** — the PR calls this an argument its fixtures do not demonstrate, which is honest; I demonstrated it: `let` / `private` / `public private(set)` / `weak` / `lazy` / `static let` / `var` all yield `[Customer]`.

**"The `SourceFile` conjuncts are ungradable from `Extract`" is TRUE, not merely harder** — `attachSwiftFieldTypeRefs` is called inside `Extract` with a single `file.Path`, so no external fixture can put a foreign record in the slice. You did not use that as an excuse; you paid ~60 lines for the internal test, and it grades finely (see R9/R10 below).

**The digest and the ratchet, verified rather than accepted.** `cmd/grafel/custom_extractor_spans_6118_test.go` contains **zero** occurrences of the string `swift` — a real negative result. The golden set is **6 `.swift` files, 17 field entities, 0 edges** — I measured it; the ratchet is vacuous for this change, exactly as stated.

**Corpus caveat holds.** 234 `.swift`, 219 in vapor, 10 vapor-api-template, 5 ktor-samples. No conclusion in the body is stated more broadly than that base supports; §7 already says the 12.88% is vapor's number.

---

# My mutants — 10, none previously run, all at the call site, all compiling, biased permissive

Each applied by exact-string replacement with an asserted occurrence count before and after; each reverted by `cp` from a shasum-verified backup with `git status --porcelain` checked empty after; `go vet` before every run; `-count=1` always.

| # | mutation (all permissive unless noted) | verdict |
|---|---|---|
| R1 | attach anchor guard: drop **only** `r.Subtype != "field"` | **ALIVE** |
| R2 | attach anchor guard: drop **only** `r.Kind != "SCOPE.Schema"` | **ALIVE** |
| R3 | neuter the `SCOPE.`-trim derivation in `swiftInComponentAddressFamily` | **ALIVE** |
| R4 | component tier: `len(famKinds) > 1` → `> 2` | DEAD (`TheTwoCollisionScansAreIndependent`) |
| R5 | alias tier: `len(allKinds) > 1` → `> 2` | DEAD (`TheTwoCollisionScans…`, `AliasShadowed…`) |
| R6 | type-param refusal only for **single-character** parameter names | **ALIVE** |
| R7 | dotted `user_type` binds its **tail** instead of refusing | DEAD (`DottedTypeBindsNeitherHeadNorTail`) |
| R8 | component collision scan counts only exact `SCOPE.Component` | DEAD (`TheTwoCollisionScansAreIndependent`) |
| R9 | pass-1 file conjunct leaks for the **allKinds** map only | DEAD (`TargetsAreScopedToTheRequestedFile`) |
| R10 | pass-2 file conjunct leaks for **alias** targets only | DEAD (`TargetsAreScopedToTheRequestedFile`) |

R9/R10 are the good news: I tried to leak each conjunct for one map / one target kind at a time, and the internal test caught every partial. The two-directions-in-one-file design does what it claims.

### R1 + R2 — **M13 is a compound guard graded only as a conjunction.** Non-blocking, ~2 lines.

`OnlyFieldRecordsCarryTheEdge`'s impostor is `{Kind: "SCOPE.Component", Subtype: "class"}` — it differs from a field record on **both** halves, so each half masks the other and neither is graded. Your M13 deleted the whole guard and died; each half deleted alone is ALIVE. Fix: add a `{Kind: "SCOPE.Schema", Subtype: "type_alias"}` impostor carrying the stash, and a `{Kind: "SCOPE.Component", Subtype: "field"}` one. (Mutually-masking guards, the #6912-era finding.)

### R6 — the type-parameter fixture holds the parameter **name length** constant at one character. Non-blocking, ~3 lines.

The code is correct (I proved `Box<Element>` above). But the only fixture uses `T`, so a regression restricting the guard to short names survives the suite. This is the arm's headline departure from arms A/D — worth one more row, `struct Box<Element>` beside `struct Element`.

### R3 — the `SCOPE.`-trim derivation is currently dead and ungraded.

The header sells deriving membership over arm F's hand-list as staying "in step with the resolver". Only `SCOPE.Class` reaches that branch, nothing emits it for Swift, and no fixture exercises it. Keep it — but it is an unexercised claim, not a demonstrated improvement. A one-line row in `TheTwoCollisionScansAreIndependent` with a `SCOPE.Class` collider would make it real.

---

# Two prose corrections (no code change)

**§8's SwiftUI hazard is mis-derived — and the correction *supports* your refusal to write the fixture.**

You wrote that `internal/custom/swift/swiftui.go:288` minting a bare-named `SCOPE.Component` from a property name "would make the target ambiguous at the resolver and dangle an edge this pass believed safe". It cannot. That site emits **`SCOPE.Component`** — the *same* kind as the declaration it would collide with. Entity IDs are re-derived downstream as `graph.EntityID(repoTag, Kind, Name, SourceFile)` (`internal/extractors/incremental.go:2068`, deliberately ignoring `r.ID`), so a same-file same-name `SCOPE.Component` from the custom lane has the **identical ID**. `buildSymbolIndex` marks a collision only on `existing != e.ID` (refs.go:1308) — no `ambigLocation`, no blanked `.base` key, no dangle. The two records merge into one node. That is a *conflation* risk (a property and a type sharing a node), not an ambiguity risk.

This resolves the apparent tension a reader will spot between §8 and the fixture refusal: the fixture refusal is about a second component-family **KIND** in one file, and swiftui.go does not produce one. Your refusal to fabricate that Swift stands. Please just restate §8's mechanism.

**M6b's reachability claim is falsified by your own corpus.** You wrote that the only Swift producing two `SCOPE.Schema` records under one name in one file is a duplicate `typealias`, "which swiftc rejects". `Sources/Vapor/Utilities/VaporSendableMetadataType.swift` emits **two** `SCOPE.Schema/type_alias` records for `VaporSendableMetatype`, via `#if` / `#else` conditional compilation. You graded that branch anyway, which was right — but it is source-reachable, not contract-only, and the corpus proves it.

**One more small gap:** `FieldTypePropertyIsLossyForGenerics` pins only the generic direction. The qualified-name direction (`Foundation.Data` → `"Foundation"`) is claimed in the header and the body and asserted nowhere. Two rows.

---

## Verdict

**Request changes**, on the extension finding alone. The resolver derivation, the split tier, the four counterfactuals, the type-parameter fix, the same-file conjunct grading, the digest and ratchet negatives — all of it re-derived and all of it correct. But "same-file targets only" is the arm's central promise and half its shipped edges do not meet it, on the idiom (`extension` on a foreign or stdlib type) that is the reason vapor is in the corpus at all. Refuse it or document and grade it; either way the 38 needs recomputing and §7 needs the number.

Nothing here needed a claim to be taken on trust. Everything I disputed, I ran.

cajasmota (commented):
## Second adversarial pass — head `6ede8a9`

Same isolated worktree, private `GOCACHE`, own copy of the corpus. Every number below is re-derived, not read. **The finding is fully addressed and I could not break the fix.**

---

# 1. The correction, re-derived independently

I reproduced the first cut on *this* tree by disabling the pass-2 refusal (`false && ext`, occurrence count asserted 1→1) and differenced the corpus target sets by name and site:

```
THIS REVISION   FILES=234 FIELDS=295 EDGES=19
EXT DISABLED    FILES=234 FIELDS=295 EDGES=38
REMOVED 19    ADDED 0
removed by target: {BaseNEncoding: 7, String: 6, RoutesBuilder: 2,
                    PasswordHasher: 1, HTTPClient: 1, Request: 1, tm: 1}
```

Exactly the PR's split, exactly the 19 rows I named last round, and **zero added** — the fix is a pure subtraction.

**Precision check on what remains.** For each of the 19 surviving edges I asked whether the target name has a real declaration keyword (`struct|class|enum|protocol|actor|typealias`) in that same file:

```
edges whose target has NO declaration in the same file: 0 of 19
```

No residue. "Same file" now means *declared there*.

---

# 2. The marker's scope — attacked in both directions, and it holds

The coordinator flagged the placement as non-obvious. It is, and it is right.

**A carrier must not become a target.** `EXT` (refusal disabled) — **DEAD**, four tests: `ExtensionCarrierIsNeverATarget`, `ForeignFileTypeIsNeverATarget`, `ExtensionOfAStdlibTypeIsNeverATarget`, `AliasExtendedInItsOwnFileGetsNoEdge`. Corpus −19.

**A carrier must still be able to make a name ambiguous.** `M19` (hidden from both pass-1 maps) — **DEAD**. I then split it, because a compound grades its second half only if the first is not lethal alone:

- **M19a** — hidden from the **allKinds** map only → **DEAD** (`AliasExtendedInItsOwnFileGetsNoEdge`).
- **M19b** — hidden from the **famKinds** map only → **ALIVE, and genuinely equivalent.** A carrier's Kind is always `SCOPE.Component`, and `famKinds` is read only on a `SCOPE.Component` *target* record, which by construction is a same-name non-carrier in the same file contributing that identical key. Excluding the carrier therefore cannot change `len(famKinds[name])` for any input this extractor can build. So M19's lethality comes entirely from its allKinds half — which is fine, because the other half is a no-op, not an ungraded branch. Worth stating out loud rather than leaving the compound unsplit.

**Placement between the allow-list and the famKinds check.** `S7` — refusal moved *above* the subtype `switch` → **ALIVE, equivalent**: both arms `continue`, so the ordering within loop 2 is immaterial. The only placement that carries meaning is loop 2 vs loop 1, and that is exactly what M19/M19a/S3 grade. Demonstrated rather than argued.

**Can a carrier escape the marker?** Eight real-world extension forms in one fixture — bare, `public`, `fileprivate`, `@available(...)`, `: Sendable`, `@retroactive @unchecked Sendable`, a `where` clause, `@objc` — each with a same-file field of that type, and with the premise asserted that each really produced a `SCOPE.Component` record. **All eight refused; none escaped.** `S1` (scan only the first child, which misses `public extension` because `modifiers` occupies index 0) is **DEAD** on `ExtensionOfAStdlibTypeIsNeverATarget`, so the access-modifier axis is genuinely graded and not merely correct by luck.

**Can a real declaration be marked?** Six declaration forms — `struct`, `final class`, `public actor`, `protocol`, `enum`, `@objc public final class ... : NSObject` — all still bind. No over-marking.

**Does the marker reach the graph?** No record retains it, on a fixture carrying both an `extension String` (refused) and an `extension Customer` beside `struct Customer` (still binds). `S4` (marker cleared *before* the target set is computed) and `S5` (marker written only when `Metadata` is already non-nil) are both **DEAD** — so the ordering inside `attachSwiftFieldTypeRefs` and the nil-map init are load-bearing and graded, not incidental.

**Collateral check on the new `Metadata` map.** Carriers previously had `Metadata == nil` (proved by S5 dying). `attachSwiftExtends` (`field_members.go:169`) guards on `Metadata == nil` and would now fall through for a carrier — but `base_candidates` is absent, so the loop body never runs and the only effect is a no-op `delete`. Inert, checked rather than assumed.

---

# 3. Counterfactuals against the 19-base, and the disjointness claim

| rule | my edges | delta | PR |
|---|---|---|---|
| this arm | **19** | — | 19 |
| arm D (component tier → all kinds) | **14** | −5, −26.3% | 14 |
| arm C (family records, Components only) | **14** | −5, −26.3% | 14 |
| arm F (alias tier removed) | **19** | 0 | 19 |

**The disjointness claim is true, and I checked it the only way that settles it** — by set membership, not by count. arm D's five dropped rows (`HOTP/TOTP.digest/digits → OTPDigest/OTPDigits`, `URLEncodedFormData.values → URLQueryFragment`) and arm C's five (`HTTPMediaTypeSet` ×2, `HTTPMediaType`, `Middleware`, `SessionID`) are *all members of the surviving 19* and share no row with the 19 that were removed. So the deltas are unchanged from the first cut and only the denominators move. Identical row sets, re-derived.

The arm-F zero remains a genuine **corpus-relative** zero, not an unreachable tier: 5 `SCOPE.Schema/type_alias` records across 4 distinct names, none a same-file field target, while `ResolvesToEntityIDs` shows an alias binding end-to-end.

---

# 4. My round-1 ALIVE mutants — all four now DEAD, verified

| # | mutation | verdict | killed by |
|---|---|---|---|
| R1 | attach guard, **Kind half alone** disabled (`false && …`) | **DEAD** | `OnlyFieldRecordsCarryTheEdge` |
| R2 | attach guard, **Subtype half alone** disabled | **DEAD** | `OnlyFieldRecordsCarryTheEdge` |
| R3 | `SCOPE.`-trim branch disabled | **DEAD** | `TheTwoCollisionScansAreIndependent` |
| R6a | type-param refusal restricted to 1-char names | **DEAD** | `TypeParameterShadowedByASameFileTypeGetsNoEdge` |
| R6b | only the **first** type parameter collected | **DEAD** | same |

R1 and R2 are the part that mattered: each half is now **independently lethal**, so the compound is scored part by part rather than as a pair. That is the correct fix for a mutually-masking guard and it was done with two impostors rather than by weakening the assertion.

---

# 5. My new mutants this round — 9, call site, permissive-biased

All applied by exact-string replacement with an asserted occurrence count (1 before, 1 after, re-read from disk), `go vet` clean before every run (guards disabled with `if false && …` so nothing goes unused), `-count=1`, reverted by `cp` from a shasum-verified backup with `git status --porcelain` **and** `git diff HEAD --name-only` checked empty after each.

| # | mutation | verdict |
|---|---|---|
| **EXT** | pass-2 carrier refusal disabled | **DEAD** (4 tests; corpus 19→38) |
| **M19a** | carrier hidden from the **allKinds** map only | **DEAD** |
| **M19b** | carrier hidden from the **famKinds** map only | ALIVE — *equivalent*, reason above |
| **S1** | `swiftIsExtensionDecl` scans only child 0 (misses `public extension`) | **DEAD** |
| **S2** | subtype allow-list widened with `"extension"` | **DEAD** (`ExtensionCarrierIsNeverATarget`) |
| **S3** | carrier refused only when it is the sole kind for that name | **DEAD** |
| **S4** | marker cleared **before** the target set is computed | **DEAD** (3 tests) |
| **S5** | marker written only when `Metadata` is already non-nil | **DEAD** (3 tests) |
| **S6** | carrier refusal gated on `Subtype == "class"` | ALIVE — not production-reachable, and harmless |
| **S7** | refusal moved above the subtype allow-list | ALIVE — *equivalent* (ordering immaterial) |

**S6** is the only one worth a sentence: the marker check is graded only against records whose subtype is `"class"`, which is every carrier today. Post-#7047 the carrier becomes `"extension"` and the allow-list refuses it one line earlier, so the gate would be dead code rather than a hole. No change asked for — recorded so the #7047 PR knows the marker's own row goes vacuous at that moment and `ExtensionSubtypeWouldAlsoBeRefused` is what carries the contract.

**S2 is the one I most expected to survive and did not.** The §2a composition claim ("if `swiftDeclSubtype` later mints `extension`, the allow-list already refuses it") is not prose — widening the allow-list to admit `"extension"` dies immediately. The claim is graded.

---

# 6. THE INSTRUMENT-BLINDNESS CLAIM — **it survives, with one material qualification**

Measured, on this tree, by toggling only the refusal:

```
FIXED     total=19  bound=19  dangling=0  wrongName=0  wrongFile=0
UNFIXED   total=38  bound=38  dangling=0  wrongName=0  wrongFile=0
```

**Confirmed.** Nineteen wrong bindings pass every field-level check the bind-rate harness performs, because the carrier has the right name, in the right file, with an admissible kind and subtype.

Surveying the rest of the instrument surface, the blindness is broader than the harness:

- **`bug-extractor` disposition** — blind by construction. It classifies *unresolved* stubs; a wrong edge that binds is never a candidate.
- **#6726 orphan rate** — **worse than blind: it rewards the defect.** Nineteen extra Swift fields stop being orphans. An arm optimising that number is paid to make this mistake.
- **`semanticDigest6118`** — detects *change*, never *wrongness*, and holds no `.swift` fixture (re-verified: **0** matches for `swift` in `custom_extractor_spans_6118_test.go`).
- **The quality ratchet** — vacuous here. I re-measured the golden set on this head: **6 `.swift` files, 17 field entities, 0 edges**, and it *does* contain `extension User` beside `struct User` in `swift-swiftui-mini/.../User.swift` — it still gains nothing, because no field there is typed `User`. Zero recall to move in either direction.

**The qualification, and I think it changes the issue you want to file.** The defect class is **not structurally inexpressible**. `internal/quality/expected.go:41` already carries `forbidden_relationships`, keyed on `from_name/from_kind/from_file/kind/to_name/to_kind/to_file` — a **`to_file`-keyed negative row can express exactly this**, e.g. "`DirectoryConfiguration.publicDirectory` must not `REFERENCES` `String` in `DirectoryConfiguration.swift`", and the swift fixture **already uses two such rows**. So the gap is coverage, not expressiveness.

There is, however, a real second-order gap, and the swift fixture documents it against itself in `known_wrong_edges_not_expressible`:

> `forbidden_relationships` is the right shape but is scored as always-fatal … so a currently-emitted wrong edge cannot be recorded there without turning the gate red.

I verified that: `diff.go:738` appends a `ForbiddenHit` on any match and `scripts/quality/ratchet.py:539-541` fails on `forbidden_hits > 0` — "always fatal", verbatim. So the harness can **prevent** a wrong edge it has a row for, but cannot **record** one that already ships, which is precisely the state a newly-found defect is in.

If it helps the issue text, the accurate statement is: *every aggregate/automatic instrument we have is blind to a wrong edge that binds — and one of them pays for it — while the single instrument that can express it needs a hand-written negative row per case and has no non-fatal tier for recording a known-wrong edge.* Two concrete asks fall out: (a) golden coverage of the declaration-vs-extension cell, (b) a non-fatal known-wrong tier for `forbidden_relationships`.

---

# 7. Prose corrections — both landed, neither over-corrected

**§8 / SwiftUI.** The mechanism is now stated correctly (same Kind → `graph.EntityID(repo, Kind, Name, SourceFile)` re-derived at `incremental.go:2068` → identical id → `existing != e.ID` never trips → the records merge; conflation, not ambiguity). It does not over-claim in the other direction either: it keeps the residual complaint, and it correctly concludes that **the fixture refusal stands**, because the unseen hazard is a second component-family *KIND* and that site does not produce one. Correct, and correctly scoped.

**M6b.** The retraction is scoped exactly as far as it should go and no further: the swiftc fact is kept, only the *unreachability inference* is withdrawn, and the mechanism is named — `#if`/`#else` compiles because one branch is inactive while the extractor sees both. The comment says that and stops. `ConditionalTypealiasIsStillOneNode` is the source-level form, and the function-level row is kept beside it rather than replaced.

---

# 8. Gates — confirmed, not accepted

- `go test -count=1 ./internal/extractors/swift/` — **green** on a pristine tree at `6ede8a9`.
- `go test -count=1 ./cmd/grafel/` — **green, 269s**, with a fresh `GRAFEL_HOME` **and** `GRAFEL_DAEMON_ROOT` inside my own scratch. Digest unmoved; the reason is verified, not assumed (0 `swift` matches in the 6118 fixture).
- `gofmt -l` clean; `go vet ./internal/extractors/swift/ ./cmd/grafel/` clean.
- Ratchet: I did **not** run it, and it is not evidence here — I measured the golden set directly at 0 edges, which makes it necessarily unmoved for this pass. The PR says exactly that rather than offering the "38 fixtures held" line as a safety result, which is the right handling.
- Tree left pristine: `git status --porcelain` empty, `git diff HEAD` empty, file shasums identical to the checkout.

---

## Verdict — **approve**

The finding was taken at full strength and fixed in the narrow direction. The marker's scope is the interesting part and it is correct in both directions, for a reason the PR states and a mutant (M19/M19a/S3) enforces — refusing the carrier as a *target* while keeping it in the *node counts*, because it is a node the resolver really sees. Choosing the marker over the subtype change, and filing #7047, is the right call: I confirmed the two compose, and confirmed it by mutant rather than by reading the comment.

Everything I re-derived matched: 19/19, the 19-row removal split, zero residue, three counterfactuals row-for-row, the disjointness by set membership, all four of my previous ALIVE mutants now dead, and the bind-rate blindness on the unfixed tree. My nine new mutants left nothing ungraded — the two survivors are equivalent with stated reasons, one of them a demonstration of the ordering claim rather than a gap.

Two items for elsewhere, neither blocking: the `forbidden_relationships` known-wrong tier (§6) and S6's note for the #7047 PR (§5).

Not merging — that is yours.

PR #7049, opened 2026-09-14
feat(java): add cross-repo Dubbo contracts and call map

### LINKED ISSUE #7048: Add cross-repository Apache Dubbo contract extraction and call map
## Problem

Grafel does not currently model Apache Dubbo consumer/provider relationships across repositories, which makes distributed Java request flows and change impact difficult to understand.

Dubbo projects may declare contracts through multiple styles:

- `@DubboReference` / `@DubboService`
- Legacy `@Reference` / `@Service`
- `ReferenceBean<T>` / `ServiceBean<T>`
- Spring Dubbo XML `<dubbo:reference>` / `<dubbo:service>`

## Proposed capability

- Extract normalized Dubbo consumer and provider contracts from annotations, bean APIs, and XML.
- Preserve interface, group, version, protocol, and resolution metadata.
- Link consumers to providers across repositories when the contract is uniquely compatible.
- Emit method-level links when interface method and arity can be resolved.
- Use lower-confidence inferred links only when omitted or unresolved metadata has no known conflict and identifies one external provider.
- Provide a dedicated Dubbo call-map dashboard so large distributed graphs remain understandable.

## Validation target

Validate extraction and linking with unit tests, the quality recall ratchet, coverage-matrix updates, and a real multi-repository graph.

0nlyl:
## Summary

- extract Apache Dubbo consumer and provider contracts from modern and legacy annotations
- support `ReferenceBean<T>`, `ServiceBean<T>`, and Spring Dubbo XML configuration
- link compatible Dubbo contracts and methods across repositories with resolved/inferred confidence
- add a dedicated Dubbo call-map API and dashboard route
- register and document the Java Dubbo coverage capabilities

## Closes / Refs

Closes #7048

## Testing

- [x] Focused Go tests for Java extraction, XML discovery, linking, and dashboard handlers
- [x] Focused `go vet` for changed Go packages
- [x] Frontend unit tests: 30 files, 316 tests
- [x] TypeScript type check and Vite production build
- [x] Coverage registry generation and validation
- [x] Quality recall ratchet: 38 gated fixtures held their baseline
- [x] `git diff --check`

## Real graph verification

Validated against the local `ossac-platform` multi-repository graph:

- 151 Dubbo services/providers
- 309 consumers and 309 cross-repository links
- 13 repositories represented
- 92 resolved links and 217 inferred links

## Notes

The complete Go test suite has existing Windows symlink-privilege failures unrelated to this change; the relevant packages and repository quality gates pass.


cajasmota:
Hey — thank you, this is a really generous PR to just show up with. Cross-repo Dubbo is genuinely one of the harder things to model here: the interface lives in one repo and the impl in another, so the whole thing hinges on getting the contract identity right across that boundary. Not a small ask, and not something I'd have got to on my own for a long while.

I've approved the CI runs (fork PRs park until a maintainer waves them through — nothing was wrong on your end, it just sits there looking idle). They should be going now.

Next step on my side is a proper review rather than a skim — 34 files deserves that, and I'd rather give you real feedback than a rubber stamp. I'll come back with specifics.

One thing that'd help me in the meantime, whenever you get a chance: what did you test it against? Even "I ran it on these two repos and eyeballed the output" is useful — mostly I want to know which shapes you already know it handles so I don't waste your time asking about cases you've covered.

Thanks again. 🙏

cajasmota (commented):
Alright — I've been through this properly, and the short version is: this is good work and I'd like it to land. Below is the real feedback I promised, plus a handful of things I fixed on your branch directly so you don't have to round-trip on them. Nothing here is a surprise-me-later item; the design is sound.

## What I liked, specifically

A few decisions here are better than the obvious thing, and I want to name them because they're the reason the cross-repo part actually works:

- **Deferring contract identity to a link pass instead of resolving it at extraction time.** The comment on `ExtractDubbo` says it outright — "unresolved group/version values are never guessed" — and the whole design follows from that. Extraction records what the file says *and how sure it is* (`interface_fqn_resolved`, `interface_candidates`, `group_resolved`), and the pass decides. That's the right seam.
- **`dubboMetadataCompatible` keying on resolvedness, not on the raw text.** An unresolved `${dubbo.group}` makes no claim about the group, so it isn't a conflict. This is the non-obvious call and it's the one that makes property-driven Dubbo configs link at all. I mutated it four different ways and the interesting ones all died on your own fixtures.
- **The ambiguity guard in `newDubboHit`** — two surviving wildcard candidates and you drop the hit entirely rather than picking one. Same for two external providers. Dangling beats guessing, and you chose dangling every time.
- **`hasDubboAnnotationImport`.** Gating legacy `@Reference` / `@Service` on the actual `org.apache.dubbo` / `com.alibaba.dubbo` annotation import is exactly the right discriminator — `@Service` is Spring's most common annotation and a looser gate would have been a disaster. I threw a Spring-only file at it that mentions `org.apache.dubbo` in a *comment*: zero entities emitted. That held.
- Running `maskJavaComments` before any of the regexes, and giving the call map its own paginated endpoint rather than pushing tens of thousands of edges through `/links`.

## Cross-repo identity: it holds

This was the thing I most wanted to check, so I drove the real extractor rather than reading the diff — a consumer file in one repo, a provider file in another, each resolving the interface through its own imports:

```
consumer repo: dubbo:consumer:com.example.api.OrderFacade   rpc_arity=2 rpc_method=submit
provider repo: dubbo:provider:com.example.api.OrderFacade   rpc_arity=2 rpc_method=submit
```

Both sides mint the same address at the contract level *and* the method level. I also checked what happens when the interface repo isn't in the group at all: both sides still resolve from their own import statement, so the link is still made — which I think is the right answer, and it's nice that it falls out of the design rather than needing a special case. The wildcard-import case degrades to a dangle, never a wrong binding. That's the property that matters.

## One real defect — fixed on your branch

`discoverDubboXML` gates at the file level (`<dubbo:` appears somewhere) but then matches elements by local name only, so in a mixed Spring context every *other* namespace's `<service>` / `<reference>` gets harvested as a Dubbo contract:

```xml
<beans xmlns:dubbo="http://dubbo.apache.org/schema/dubbo"
       xmlns:osgi="http://www.eclipse.org/gemini/blueprint/schema/blueprint">
  <dubbo:reference id="orderFacade" interface="com.example.api.OrderFacade"/>
  <osgi:service   interface="com.other.NotDubbo" ref="notDubboBean"/>
  <osgi:reference id="alsoNotDubbo" interface="com.other.AlsoNot"/>
</beans>
```

That yields 4 Dubbo entities instead of 1. It's not just noise: an `<osgi:service interface="com.other.X">` in one repo and an `<osgi:reference interface="com.other.X">` in another are a valid consumer/provider pair as far as the link pass is concerned, so they get joined as a cross-repo Dubbo call. Gemini Blueprint / CXF / Camel sitting in the same Spring context as Dubbo is pretty normal in the legacy Java shops that still run Dubbo, so I think this one is reachable.

Go's decoder already resolves the prefix for us, so the fix is a namespace check — I pushed it rather than making you round-trip:

- `isDubboElementNamespace` accepts `http://dubbo.apache.org/schema/dubbo`, `http://code.alibabatech.com/schema/dubbo`, anything ending `/schema/dubbo`, and the bare literal `dubbo` (which is what `encoding/xml` hands back when the prefix is used without an `xmlns` declaration — I checked, and your existing fixture still passes either way).
- Two tests: one for the mixed-namespace document above, one for the undeclared-prefix case.

If you'd rather widen or narrow that namespace set, go ahead — you know the deployed schema URLs better than I do.

## A smaller one, also fixed

`implements Handler<String, Order>` was being split on every comma, so the interface came out as `Handler<String`, got qualified with the *implementation's* own package, and was marked resolved — a contract address that can never match anything. I pulled the split into `firstImplementedInterface`, which splits only on top-level commas and drops the type arguments, so it resolves to `com.example.api.Handler`. Single-interface classes behave exactly as before.

## One I'd rather ask about than change

Related, but this one looks like a genuine design choice rather than a slip, so I left it alone: `implements Serializable, OrderFacade` takes the *first* entry, so the provider's contract is minted as `java.io.Serializable` and the real `OrderFacade` contract is never emitted at all — the consumer then silently dangles. The obvious fixes both have downsides (prefer the entry that has an import? skip `java.*` / known marker interfaces? prefer whichever one some consumer in the group actually references?) and I didn't want to pick one on your behalf. What do Dubbo providers in the wild you've looked at actually look like here — is a marker interface first common enough to be worth handling, or is this a corner you'd rather leave dangling?

## Test coverage — two conjuncts nothing was watching

Your fixtures are well built against "the fix stopped working". They're thinner on "it fires too broadly", which is the usual shape. Two concrete gaps, and I've pushed tests for both:

- **Same-repo.** The pass is cross-repo only, and that's enforced by two separate conjuncts — the `provider.repo != repo` filter in `externalDubboProviders` and the `provider.repo != consumer.repo` in the inferred arm. I removed each independently and the suite stayed green both times: a repo with a provider module and a consumer module (common enough in Dubbo monorepos) would have started linking to itself with nothing to catch it. `TestDubboPassSkipsSameRepoConsumerProvider` pins both.
- **The `known` half of `dubboMetadataCompatible`.** `TestDubboPassLinksUniqueUnresolvedConfigAsInferred` has the consumer unresolved but with the *same* literal value as the provider, so the `value != ""` conjunct masks the `known` one — dropping `left.known && right.known` left the suite green. `TestDubboPassUnresolvedPlaceholderIsNotAConflict` gives the placeholder a different literal (`${dubbo.group}` vs `orders`) so the resolvedness check is what's doing the work.

With those two in, 9 of the 10 mutations I ran on the pass die. The survivor is the `!strings.Contains(interfaceName, ".")` in the final guard of `newDubboHit` — I'm fairly sure it can't fire today, since the candidate block above it already blanks anything without a dot and every candidate is built as `pkg + "." + type`. It's a harmless belt and I left it; mentioning it only so it doesn't look like a gap I missed.

## Minor — genuinely take or leave

- `Discover` now does a bounded read of every unclassified `.xml` in every repo to test for `<dubbo:`. Bounded, so probably fine, but it's a new per-file cost on large repos — did you happen to measure it on the 13-repo graph you validated against?
- `dubboAttribute`'s `[^,]+` and the annotation regex's `([^)]*)` both stop at the first `,` / `)`, so `@DubboService(methods = @Method(name = "x"))` or a `parameters = {...}` map will truncate. I'd guess that's rare enough not to matter — just flagging in case you've hit it.
- Consumer `rpc_arity` counts call-site arguments, provider counts declared parameters, so a varargs method called with 3 args won't method-match a 2-parameter declaration. It degrades cleanly to the service-level link, so this reads deliberate to me — confirming rather than asking you to change it.
- `extractDubboConsumerCalls` scans the whole file for `fieldName.method(`, so a same-named local in an unrelated class in the same file would mint a phantom method contract. Rare, and cheap to live with.

## Gates I ran

Everything below is green on your branch with my three commits on top (`6b63d40`, `f28a838`, `2aaf7e6` — two fixes and one test-only commit, all pushed to `feat/java-dubbo-contracts`; shout if you'd rather I hadn't and I'll drop them):

- `go build ./...`, `go vet`, `gofmt -l` — clean
- `go test -count=1` on `internal/custom/java`, `internal/links`, `internal/extractors/config`, `internal/dashboard` — all pass
- `go test -count=1 ./cmd/grafel/` (the whole-graph digest pin — easy to miss, since nothing in `internal/extractors/...` points at it) — pass, digest unmoved
- `scripts/quality/run.sh --ratchet` — 38 fixtures held, working tree clean afterwards (that last part is the check that actually matters; the script exits 0 even on a ceiling drop)

The only thing I couldn't check locally is the frontend suite — no `node_modules` in my worktree and I didn't want to install one just for this. CI covers it and you reported it green, so I'm taking that as read.

## Where that leaves us

I'd like to merge this. The remaining open item is the `implements Serializable, OrderFacade` question above — happy to land as-is and follow up separately if you'd rather, or if you want to take it now I'll hold.

And the earlier question still stands whenever you have a moment, no rush: which Dubbo shapes did you already run this against? Mostly so I know which ground you've already covered and don't send you re-testing things you've checked.

Thanks again for this — the uncertainty-carrying design in particular is something I'd want to copy for the next RPC framework we model. 🙏

0nlyl:
I pulled the three follow-up commits (`6b63d40`, `f28a838`, and `2aaf7e6`) and reviewed them locally. The namespace scoping and generic raw-type handling match the intended behavior.

### What I tested against

The implementation was exercised against the local `ossac-platform` graph and focused fixtures covering:

- Apache Dubbo `@DubboReference` / `@DubboService`
- - legacy Apache/Alibaba `@Reference` / `@Service`, gated by the actual Dubbo annotation import
- - `ReferenceBean<T>` and `ServiceBean<T>`
- - Spring XML `<dubbo:reference>` and `<dubbo:service>` configuration
- - literal and placeholder-driven `group` / `version` metadata
- - explicit imports, wildcard imports, unresolved candidates, and ambiguity/dangling behavior
- - service-level linking plus method name/arity matching
- - resolved and inferred cross-repo links
The real graph validation behind the PR description covered 13 repositories and produced 151 provider/service contracts, 309 consumers and 309 cross-repository links: 92 resolved and 217 inferred. I also visually checked representative consumer → contract → provider paths in the call-map page rather than relying only on aggregate counts.

After pulling your commits, the focused Windows tests for `internal/custom/java`, `internal/links`, and `internal/extractors/config` pass locally. The full dashboard package still reaches the existing Windows symlink-privilege failure mentioned in the PR description.

### `implements Serializable, OrderFacade`

I searched the current 16-repository local corpus for marker-first and multi-interface implementations. There are ordinary classes with `Serializable` first, but I did not find a Dubbo provider with a marker interface before its service contract. The multi-interface service implementations I found put the business interface first and lifecycle/support interfaces second.

My preference is therefore to land this PR as-is and handle marker-first providers in a focused follow-up. I would rather not make the current extractor guess based on whichever interface has an import or a matching consumer in the group. A follow-up can define explicit semantics—likely known marker/lifecycle exclusion or multiple candidates with the same ambiguity safeguards—and pin them with dedicated fixtures.

### XML discovery cost

I did not run an isolated before/after benchmark for the bounded XML read, so I cannot give a defensible delta. The latest full 16-repository rebuild on my machine completed the group rebuild/link/analytics pipeline in about 4m11s; one 357-file Java repository extracted in about 1.4s. That does not show the XML scan as a visible bottleneck, but it is not an A/B measurement.

The currently failing Windows compile check appears unrelated to the change: it failed during `go mod download` while fetching `github.com/knights-analytics/hugot@v0.7.3` from `proxy.golang.org` with an HTTP/2 `INTERNAL_ERROR`, before compilation or vet ran. A rerun should distinguish the transient download failure.

For the other edge cases you listed, the current behavior is intentional for this PR: nested annotation arguments may truncate, varargs can fall back to the service-level link, and same-named locals can theoretically over-attribute a field call. I agree those are better handled as separately scoped hardening work rather than expanding this PR further.

cajasmota:
Pushed one more test-only commit to your branch — nothing about the implementation changed.

Your `hasDubboAnnotationImport` guard is exactly right, and it turns out to be the single thing standing between a plain Spring `@Service` class and being minted as a Dubbo provider (which would then feed the cross-repo link pass). It just didn't have any test observing it yet — normal for new code, and squarely a maintainer chore rather than yours. So I added two tests, one per call site, so a future refactor can't quietly delete it:

- `TestDubboProviderAnnotationRequiresDubboImport`
- `TestDubboReferenceAnnotationRequiresDubboImport`

Thanks again for this one — it's in good shape.

cajasmota:
**Merged** → `5391629da`. Thank you — this is now grafel's first cross-repo RPC contract support, and it arrived from outside the project, which is not a small thing.

The design choice I want to name, because it's the reason this went in smoothly: **you deferred identity to the link pass and carried the uncertainty forward** (`interface_fqn_resolved`, `interface_candidates`, `group_resolved`) instead of guessing at extraction time. Every ambiguity case you wrote chooses a dangle over a binding. That's exactly the right instinct here — a missing edge is a gap someone can find, a *wrong* edge is a lie the graph tells confidently, and our own tooling can't flag the second one. Several of our internal arms have gotten that backwards.

It held up under attack: driven through the real resolver, consumer and provider sides mint the same address at both contract and method level, and the interface repo being absent from the group still links correctly — because each side resolves from its own import. That falls out of your design rather than being patched in.

### What we changed on your branch

Four commits, all on top of your work:

1. **`6b63d40`** — `discoverDubboXML` gated at file level (`<dubbo:` anywhere in the file) but matched elements by *local name*, so in a mixed Spring context `<osgi:service>`, `<osgi:reference>` and `<jaxws:service>` all became Dubbo contracts. Worse, an osgi service in one repo plus an osgi reference in another got **joined as a cross-repo Dubbo call**. Added a namespace gate, plus the undeclared-prefix case where `encoding/xml` hands back the bare literal.
2. **`f28a838`** — `implements Handler<String, Order>` was split on every comma, so the contract became `Handler<String`, qualified with the impl's own package and marked resolved. Now splits top-level commas only and strips type args.
3. **`2aaf7e6`** — tests for two ambiguity conjuncts that were reachable but ungraded.
4. **`9575e12`** — tests pinning your Dubbo-import guard (see below).

That last one is worth explaining, because **your guard was right and we just gave it teeth.** The annotation regexes match `@(DubboService|Service)` and `@(DubboReference|Reference)` — and `@Service` is Spring's. Your `hasDubboAnnotationImport` check is the only thing standing between every Spring `@Service` class in every Java repo and a minted Dubbo provider. It was doing its job, but nothing observed it: we removed it at each call site in turn and the whole suite stayed green, and a plain Spring class then produced two Dubbo entities that would have flowed straight into the cross-repo link pass.

There was even an existing test that *looked* like it covered this — Spring `@Service` fixture, asserts zero entities — but it was vacuous: its class had no `implements` clause, so the interface resolved empty and the extractor bailed before the guard ever mattered. Adding `implements OrderFacade` is what makes the fixture actually bite. Easy thing to miss, and not a mark against you — it's the normal state of new code, and catching it is our job, not yours.

### One open question, not a blocker

`implements Serializable, OrderFacade` currently mints the contract as `java.io.Serializable` and never emits the real one. That's a genuine policy call and you know Dubbo better than we do — prefer the first *imported* entry? Skip `java.*` marker interfaces? Something else? Happy to implement whichever you'd pick, or leave it to you if you'd rather.

Also flagged for later, no action needed from you: `Discover` now bounded-reads every unclassified `.xml` in every repo, which is worth measuring on a large corpus at some point.

Closing #7048 as delivered. If you want to take the `Serializable` question or anything else here, you're very welcome — and thanks again for the care that went into this. 🙏

Issue #7051, opened 2026-09-14
Windows: daemon scheduled task can fail silently on its first run right after `grafel install`/`start`, with no automatic recovery within the operation's own budget


mota-mota:
## Summary

On Windows, `grafel install` / `grafel start` (and any internal call to
`ensureLoaded`) register the `com.grafel.daemon` scheduled task and then fire
it once via `schtasks /run`, deliberately ignoring that command's exit code
(`schtasks_windows.go:289`, comment: *"A /run failure is non-fatal — the
readiness poll is the real success signal"*). In practice this run can fail
on the very first attempt immediately following `/create`, with **zero**
output written to `daemon.log` (the failure happens before the daemon's
structured logger is reached). Because Task Scheduler's own
`RestartOnFailure` (`PT1M` / 3 attempts) has a 1-minute interval — longer
than the 60 s `waitReady` budget `ensureLoaded` allows — and does not appear
to trigger for this on-demand `/run` in the observed case, nothing brings the
daemon up before `ensureLoaded` gives up. The install/start operation then
reports the daemon as unreachable, `grafel status`/`grafel doctor` keep
saying "daemon not running" / "daemon unreachable", and **the daemon stays
down indefinitely** — until something manually retries the task (or the user
logs off/on again, if the logon trigger still fires).

A plain retry of the *exact same, unmodified* task
(`Start-ScheduledTask -TaskName com.grafel.daemon`, no reinstall, no code
change) succeeds immediately and produces a completely normal startup
sequence.

## Environment

- grafel `v0.3.3` (commit `a16fd6d`, built `2026-09-08T19:58:38Z`)
- Windows 11 Pro (x86_64), official installer, binary at `C:\Users\<user>\.grafel\bin\grafel.exe`
- Task registered at `LeastPrivilege` / `InteractiveToken`, no elevation

## Impact

- After an install/repair cycle, the daemon can end up **silently dead**
  with no user-visible indication that a retry would fix it — `grafel
  status` just says "not running", `grafel doctor` says "daemon unreachable
  via RPC socket" and "engine degraded: stale heartbeat", neither of which
  hints that the fix is "run the task again."
- Every MCP tool call fails (`/mcp` → "Reconnected to grafel, but fetching
  tools failed") until a human notices and manually re-triggers the task.
- `grafel doctor`'s own stale-daemon detection is a no-op on this platform
  (`stale-daemon scan: process scan: process introspection unsupported on
[...]

Issue #7050, opened 2026-09-14
Windows: the daemon can die silently within ~1 minute of a confirmed-healthy start, regardless of launch method, with zero trace anywhere


mota-mota:
## Summary

On Windows, a `grafel serve` daemon process that has **already been confirmed
alive and RPC-responsive** (via `grafel status`, showing a real pid, uptime,
rss, and dashboard URL) can disappear entirely within under a minute — with
**no line written to `daemon.log`**, no Windows Event Log entry, no WER
crash report, and no Windows Defender detection. The process is simply gone,
as if cleanly terminated by something external, not crashed. This is
independent of how the daemon was started: it was reproduced via
`Start-ScheduledTask` and via a fully detached `Start-Process`, neither of
which involves the other.

This matches (and now reproduces + narrows down) a previously-noted, unresolved
observation from an earlier session: *"Daemon inestable: tras arrancarlo con
`Start-ScheduledTask` muere solo a los pocos minutos sin escribir nada en
`logs/daemon.log`... Pendiente de investigar."*

## Environment

- grafel `v0.3.3` (commit `a16fd6d`, built `2026-09-08T19:58:38Z`)
- Windows 11 Pro (x86_64), official installer
- Windows Defender real-time protection **disabled** on this machine
  (`Get-MpComputerStatus` → `RealTimeProtectionEnabled: False`,
  `AntivirusEnabled: False`) — ruled out as the cause here, though a
  third-party AV/EDR was not checked

## Impact

- Makes the daemon fundamentally unreliable on Windows: even a *successful*
  start (task or manual) is not a durable state. A user/tool that sees
  `grafel status` report `Daemon: running` cannot trust that it will still be
  running a minute later.
- Compounds `issue-windows-daemon-first-run-after-install.md`: even the
  documented workaround there ("just retry the task/process") is not
  actually reliable, since the retried instance can itself die silently
  shortly after.
- `grafel doctor`'s stale-daemon detection cannot see this either
  (`stale-daemon scan: process scan: process introspection unsupported on
  windows`), so there is no built-in signal, even after the fact, that this
  happened — a user just finds the daemon down again with no explanation.

## Steps to reproduce (observed three times, independently, in one session)

1. **Via `Start-ScheduledTask`:** started `com.grafel.daemon` from a fully
   stopped state. `grafel status` confirmed `Daemon: running pid=32804…`
   moments later. Left alone (not stopped or touched) for a few minutes while
[...]

Issue #7005, opened 2026-09-08
Windows-conditional code cannot be compiled or vetted locally: cross-builds default to CGO_ENABLED=0 and every tree-sitter grammar drops out, so `GOOS=windows go vet ./internal/cli/` fails before type-checking


cajasmota:
Surfaced while reviewing #7001, which fixes a bug that **only manifests on Windows** (the mcp-bridge ownership record was never written there, because the Windows socket path is a named pipe and `filepath.Dir(`\\.\pipe\…`)` is not a writable directory).

## The gap

`GOOS=windows GOARCH=amd64 go vet ./internal/cli/` does not compile. Verified myself on `36da881db`, not relayed:

```
github.com/tree-sitter/tree-sitter-bash/bindings/go: build constraints exclude all Go files in .../tree-sitter-bash@v0.23.3/bindings/go
github.com/cajasmota/grafel/internal/treesitter/ts/grammars/dockerfile: build constraints exclude all Go files in .../grammars/dockerfile
… (every grammar, vendored and in-tree)
```

Cross-compilation defaults to `CGO_ENABLED=0`, and the tree-sitter bindings are cgo-only, so every grammar package drops out and nothing that transitively imports them can be type-checked for another platform. `internal/cli` does, so **no Windows-conditional code in this repo can be compiled, vetted or type-checked locally.**

## Why it matters more than it looks

This repo has a standing rule to add the `ci:full` label for any change with `runtime.GOOS`-conditional behaviour. That rule exists because CI is the only place the other platforms are exercised — but the gap is worse than "tests only run in CI". **The code cannot even be compiled locally for Windows.** So for a Windows-conditional change:

- a local `go vet` proves nothing about the Windows build;
- a typo in a `//go:build windows` file is invisible until CI;
- and the round-trip to find out is a full CI run, which is where #7001 spent one iteration.

#7001's `mcp_bridge_session_windows.go` is exactly this shape: a whole file that no local command in this tree can check.

## What is NOT claimed

- I have not established that a `CGO_ENABLED=1` cross-build works given a mingw toolchain — that is the obvious next thing to test, and it may be the whole fix.
- I have not measured how many packages are affected; I checked `./internal/cli/` because that is where the Windows-conditional bridge code lives. **The rate is unmeasured — one package is not a population.**
- This is pre-existing and unrelated to #7001's diff. It is not a regression.

## Possible shapes, unranked

[...]

Issue #6333, opened 2026-08-18
the Windows daemon wrapper writes no log, so any launch failure is completely silent and Task Scheduler blames wscript


cajasmota:
Split out of #6325, where it was correctly judged out of scope. It is the highest-value remaining item from that list.

## The problem

The Windows daemon is now launched through a hidden WScript wrapper (#6320, hardened in #6328). The wrapper writes **no log**, and it runs under `//B` (batch mode), which suppresses error dialogs.

So when `shell.Run` fails to launch the daemon, the user gets:

- no console (that is the point of the wrapper),
- no log line,
- no error dialog,
- and Task Scheduler's last-result reflects **`wscript`**, not `grafel` — so the one signal a user *can* see points at the wrong process.

The daemon is simply dead, and nothing anywhere says why.

## Why this is not hypothetical

#6325's F2 is the proof. The `.vbs` was written UTF-8 with no BOM while WSH decodes `.vbs` as ANSI, so **any user with an accented or non-Latin username** got a mis-decoded path, a failed launch, `RestartOnFailure` burning its three attempts, and a dead daemon. That defect was found by static reasoning about WSH's encoding behaviour — not by a bug report — because there was no diagnostic for a user to send us.

Everything else that makes `shell.Run` fail lands in exactly the same silent hole:

- the binary moved or was renamed
- execute permission revoked
- antivirus quarantine (common for a freshly written `.vbs` on a corporate image)
- `%LOCALAPPDATA%` redirected by policy
- a partially written wrapper (the case #6328's atomic write closes)

None of those are exotic, and none of them would produce anything a user could act on or report.

## Fix

Redirect the wrapper's output to `opts.LogDir`. The wrapper already knows the paths it needs; this is a small change to the generated VBScript plus the `Options` it renders from.

Worth deciding at the same time: whether a launch failure should also leave a breadcrumb that `grafel doctor` can find and surface, so the diagnosis does not depend on the user knowing where to look.

## Testing note

The renderer is testable — #6328 moved it into an untagged `internal/daemon/service/schtasks_wrapper.go` precisely so its output can be asserted on any platform. The *behaviour* (does WSH actually write the log) is not testable without a real Task Scheduler, so the honest split is: assert the emitted script contains the redirection, and state plainly that end-to-end verification is manual.

Refs #6317, #6320, #6325, #6328

cajasmota:
[...]

PR #7052, opened 2026-09-14
fix(#7050): a dead stderr handle silenced daemon.log, and a failed force-kill still took the pidfile

cajasmota:
## What

Three changes on the Windows silent-daemon investigation, in descending order of importance.

### 1. One dead stderr handle could silence `daemon.log` for an entire run

`cmd/grafel/daemon.go` built the daemon logger over `io.MultiWriter(os.Stderr, logFile)`. `io.MultiWriter` **stops at the first sink that errors and never reaches the rest**, and `slog` discards the handler's write error. So a single failing write to stderr silences `daemon.log` for the whole process lifetime, with nothing anywhere to say why.

That sink is not hypothetical on Windows: `detachConsole()` (FreeConsole) runs ~80 lines *before* the logger is built, and a daemon spawned detached (`DETACHED_PROCESS` / `CREATE_NO_WINDOW` / a hidden `wscript` wrapper) may hold no valid stderr handle at all.

This is a **candidate explanation** for the observation in #7050 that otherwise looks impossible — a provably-running, RPC-responsive daemon with *zero* lines in an append-only `daemon.log`, not even the `startup: pidfile-acquire begin` emitted milliseconds after start. It also gates this PR's whole value: a WARN nobody can read buys nothing.

`newFanoutWriter` writes to every sink regardless of what the others do, and reports failure only when they *all* fail. The combining happens inside `buildDaemonSlogLogger`, not at the call site, so one test holds the decision.

**Not claimed:** that this is what happened on the reporter's machine. Windows is unobservable here, and `daemon.log` staying empty has other candidate causes (a `GRAFEL_DAEMON_ROOT` in that shell would redirect log *and* pipe together, which `grafel status` would follow). The verified part is that the mechanism exists — pinned by a test that also documents `io.MultiWriter`'s behaviour so the premise cannot go stale.

### 2. A failed force-kill still took the pidfile (review M7)

`process.ForceKill` is `OpenProcess(PROCESS_TERMINATE)` + `TerminateProcess` on Windows and can be refused (`ERROR_ACCESS_DENIED` across sessions / integrity levels). The code ignored `killErr` and overwrote the pidfile anyway: the unresponsive owner stays alive, our pid is recorded as the owner, and every later reclaim/reap decision keys off a pid that is not the one serving — two live daemons, silently, in exactly the case this log line exists for. It now refuses, returning an error naming the pid it could not clear, and still logs the attempt with `kill_err` set.

### 3. The WARN reports the probe budget it actually spent

`socketIsHealthy` returns the attempts it made; the WARN reports that instead of re-deriving `socketHealthProbeRetries+1`. A field that re-derives a constant keeps claiming "3 attempts" after someone edits the constant to 0.

Plus the original change: the reclaim's `TerminateProcess` is logged at WARN by the surviving process, naming the pid it killed. **No reclaim decision changed** — no timeout moved, no retry added.

## Grading

Mutants on the call site, permissive-biased, each asserted to apply at exactly 1 occurrence, restored by `cp` from shasum-verified backups (`git diff HEAD` and `git status --porcelain` empty after).

| # | Mutant | Verdict |
|---|---|---|
| M1 | silence the WARN | **DEAD** — `Reclaim_LogsTheForceKilledPID` |
| M2 | log a reclaim on the healthy-owner path too | **DEAD** — `HealthyOwner_LogsNoReclaim` |
| M7r | drop the failed-kill refusal, proceed as before | **DEAD** — `ReclaimKillFails_RefusesAndLogsKillErr` |
| M3 | `socketHealthProbeRetries` 2 → 0 | **DEAD** — `SocketIsHealthy_ProbeBudget` + `ReclaimWARN_FieldsMatch…` |
| M4 | probe timeout 300ms → 1ns | **DEAD** — `SocketIsHealthy_ProbeBudget` |
| M8 | `"socket"` field → `pidPath` | **DEAD** — `ReclaimWARN_FieldsMatch…` |
| M9b | `probe_attempts` → hardcoded 1 | **DEAD** — `ReclaimWARN_FieldsMatch…` |
| M10 | fanout aborts on the first failing sink | **DEAD** — 3 fanout tests |
| M11 | `buildDaemonSlogLogger` uses `io.MultiWriter` | **DEAD** — `BuildDaemonSlogLogger_SurvivesADeadConsoleSink` |

M9 as a bare literal does not compile (`attempts` goes unused), so it was re-run as M9b, which compiles — a build failure is not a kill. M5/M6 from the review (identity check, always-reclaim) remain DEAD and untouched.

## Verification

`go test -count=1 ./internal/daemon ./cmd/grafel` green (39s / 214s). `go vet` green natively for both. `gofmt` clean.

**Corrected from the first version of this body**, both errors flagged in review:

- **`ci:full`** — not needed, but not for the reason first given. "No build tag changed" does not follow: `AcquirePIDFile` dispatches straight into `//go:build windows` code (`pidExeBaseName`, `ForceKill` → `TerminateProcess`) and both logged paths come from `paths_windows.go`. What settles it is `test.yml:160` — ubuntu + windows run on every PR, so `internal/daemon` and `cmd/grafel`, including every new test, already compile and execute on Windows. `ci:full` adds only macOS tests and `-race`, and this diff adds neither macOS-specific code nor concurrency.
- **`GOOS=windows go vet`** — the first body reported it green for `./internal/daemon/service/...` and `./internal/process/...`, **neither of which contains a changed file**. The changed packages are `internal/daemon` and `cmd/grafel`, and #7005 blocks vetting them under `GOOS=windows` (`CGO_ENABLED=0` excludes the tree-sitter bindings). So: the changed packages **could not be vetted for Windows here**; Windows coverage comes from the `test (windows-latest)` leg.

## Not in this PR

The `#7051` readiness-budget argument published on the contributor's thread was wrong and has been retracted by the maintainer; that decision is now decoupled from this diff. #7053's cost estimate is corrected on the issue itself.

Refs #7050, #7051, #6333.


cajasmota (commented):
## Independent adversarial review — head `8383a52f`, own worktree, macOS

The change itself is right and I want it in. What I am blocking on is (a) three ungraded mutants on the exact property this PR exists to establish, and (b) one argument published to a contributor's thread that I believe is **wrong** and will freeze a real fix if it stands.

Baseline before any mutation, private `GOCACHE`, `-count=1`:

```
ok  internal/daemon           46.6s
ok  internal/process           4.0s
ok  internal/daemon/service    8.7s
ok  internal/daemon/watchreg   1.0s
```

`test (windows-latest)` is still **pending** at the time of writing. Not mergeable until it lands.

---

## 1. The load-bearing negative does not survive

From the #7051 reply:

> If we raise the readiness budget past `PT1M` … then we are deliberately arranging for a **second `grafel serve` process to be launched while `install`/`start` is still waiting**.

This is the reason given to the reporter for not fixing the defect they correctly identified. I think it is unsound in two independent ways, either of which is sufficient.

**(a) The budget does not cause the retry.** `RestartOnFailure` is a Task Scheduler property of the registered task. If an instance ends in failure, Task Scheduler fires the retry at `PT1M` whether `defaultReadiness.budget` is 60 s, 10 s, or 300 s. Raising the budget changes exactly one thing: whether `install` is still *blocked in `waitReady`* when the retry lands. It cannot make the second `serve` more or less likely. So the hazard being named as the *cost of the fix* is a hazard that exists identically **today**, under the 60 s budget — `ensureLoaded` just returns `errNotReady` and walks away a few seconds before the retry fires unobserved. `ensureLoaded` does no teardown on failure (it returns `st.Installed = true` and the error; `manager.go`), so nothing about the current budget suppresses anything.

**(b) A live daemon means a live task instance, so the retry cannot fire against an incumbent.** From this repo's own `schtasks_windows.go`, the wrapper is a supervisor, by design:

> `wscript` wrapper — launches grafel without a console, **waits for it, and propagates its exit code so Task Scheduler remains the process supervisor**

`RestartOnFailure` fires when an instance *ends* in failure. While a healthy daemon runs, the instance is *running*, not failed — no retry. And `<MultipleInstancesPolicy>IgnoreNew</MultipleInstancesPolicy>` independently blocks a second instance while one is running. Both of those are in the same file as the `PT1M` we quoted at the reporter.

So the task-launched second `serve` can only exist once the previous instance has already exited — i.e. in exactly the state where there is no incumbent from that lineage to kill. In #7051's own scenario (`/run` fails before the structured logger is reached) the retry meets a **stale** pidfile, and a stale pidfile is overwritten silently with no kill at all. The path we said we were avoiding is not reachable from the scenario we invoked to avoid it.

There is a residual real risk, but it is a *different* one and the budget does not gate it: a manual `Start-Process … serve` bypasses `IgnoreNew` entirely and can race a task instance. That is #7050 repro 2/3, and raising the readiness budget neither causes nor cures it.

**Ask:** retract this on #7051 and decouple the two decisions. "We want to understand the death before we move a timeout" is a perfectly good reason to defer. "Raising the budget would manufacture the kill" is not, and as written it tells a contributor their correct finding is unfixable. A false *don't fix it that way* is more expensive than a slow fix.

### A second objection to Hypothesis 1 itself, while we are here

A **successful** reclaim leaves the killer alive: `AcquirePIDFile` returns `release`, `run()` continues to `socket-listen`, and the killer *becomes* the daemon under a new pid. #7050's reporter observed `Daemon: not running` / "daemon unreachable via RPC socket" after every death. So H1 as stated does not explain the observed end state — it needs a second, unexplained event in which the killer also dies. Worth saying out loud before the reporter spends an evening on the experiment; it does not sink H1, but it is a step we have not accounted for, and the WARN this PR adds is what would expose it.

### The diagnostic promise has an unchecked premise

The WARN uses the same `logger` that emits `startup: pidfile-acquire begin` two lines earlier (`server.go:618-619`), and `daemon.log` is shared and append-only by design (`paths.go:79-81`). So the field placement is right and the killer's line will land where we are telling the reporter to look. **But** #7050 repro 2 reports zero new lines in `daemon.log` for a run whose daemon was demonstrably up (`uptime=1s rss=117.5MB`, dashboard URL served) — including no `pidfile-acquire begin`. That process provably executed the line immediately above our new one. Until we know why its output did not reach `daemon.log`, we do not know that the new WARN will reach it either.

**Ask:** add one question to the experiment — "for the `Start-Process` run, does `startup: pidfile-acquire begin` appear in `daemon.log` at all?" If it does not, `daemon.log` is not the observation surface for that launch method and this PR will not be visible to the reporter regardless of whether H1 is correct. That is a one-line addition and it protects the whole experiment.

---

## 2. Is a log line the right ship? Yes — but grade it

For: it is the honest minimum, it changes no behaviour, it makes a previously unrecorded `TerminateProcess` findable, and it is worth having under #6333 whether or not H1 survives. It fires on the correct branch (after the `socketIsHealthy` early-return, before the pidfile write), at WARN, from the process that survives to write it. I verified all of that.

Against: it closes neither issue, and a P1 with a merged PR against it looks attended-to. Mitigated by the PR body being explicit that no behaviour changed — keep that.

---

## 3. Mutant table — 7 more on the call site, permissive-biased

Each applied by `cp` from a shasum-verified backup, asserted to match at **exactly 1 occurrence before and after**, `go vet ./internal/daemon` green, `go test -count=1 ./internal/daemon`. Tree confirmed clean (`git status --porcelain` and `git diff HEAD` both empty) after the last restore.

| # | Mutant | Verdict |
|---|---|---|
| M3 | `socketHealthProbeRetries` 2 → **0** (condemn after a single 300 ms probe) | **ALIVE** — ungraded |
| M4 | `socketHealthProbeTimeout` 300 ms → **1 ns** | **ALIVE** — ungraded |
| M5 | `ok && pidIsLiveDaemonFunc(existing)` → `ok && (pidIsLiveDaemonFunc(existing) \|\| pidAlive(existing))` (kill any live pid — #7054's hole) | **DEAD** — `TestAcquirePIDFile_LiveNonGrafelPID_TreatedStale` is the distinguishing input; the test binary `SIGKILL`s itself, `signal: killed` |
| M6 | `if socketIsHealthy(socketPath)` → `if false && …` (always reclaim, kill a healthy incumbent — literally the #7050 scenario) | **DEAD** — `…HealthySocket_RefusesNoReclaim` + the new `…HealthyOwner_LogsNoReclaim` |
| M7 | gate the WARN on `killErr == nil` — a **failed** kill logs nothing | **ALIVE** — ungraded, and production-reachable |
| M8 | `"socket", socketPath` → `"socket", pidPath` | **ALIVE** — ungraded |
| M9 | `"probe_attempts", socketHealthProbeRetries+1` → `"probe_attempts", 1` | **ALIVE** — ungraded |

Good news first: **M5 and M6 are dead**, so the two decisions that actually matter — *is this pid ours* and *is the incumbent healthy* — are genuinely pinned. That is more than I expected.

**M7 is the blocking one.** No test stubs `forceKillFunc`; all three new tests spawn a real child and kill it for real, so on darwin/linux `killErr` is always `nil` and the `killErr != nil` case is never entered. On Windows `ForceKill` → `Kill` → `OpenProcess(PROCESS_TERMINATE)`, which can fail with `ERROR_ACCESS_DENIED` (cross-session, different integrity level) — and the code then **still overwrites the pidfile and proceeds**, producing two live daemons. That is the single most dangerous state on this path and it is the state the log line exists to make visible; under M7 it is silent and nothing notices. Fix is ~15 lines: stub `forceKillFunc` to return an error in one new test, assert the WARN still fires and carries a non-nil `kill_err`.

**M3/M4** are pre-existing (`withFakeSocketHealth` replaces `socketHealthProbe` wholesale, so `socketIsHealthy`'s loop is never exercised by anything), but this PR is the one that *advertises* `probe_attempts` and `probe_timeout` as the evidence that condemned a process. Shipping a field whose underlying constant can be silently halved is a diagnostic that lies by omission. A direct test on `socketIsHealthy` counting probe invocations is ~15 lines and kills M3, M4 and M9 at once.

**M8** is minor on its own but confirms the shape: of the five fields, only `reclaimed_pid` (and the WARN level) are asserted. Note that `socketPath` comes from `paths.go` / `paths_windows.go`, so the value in that field is genuinely platform-dependent — worth one assertion that it is the socket, not some other path.

---

## 4. `ci:full` — agree it is not needed, but the stated reason is not the reason

The conclusion is right; the argument is not. "No build-tagged code changed, so the fix is platform-neutral" does not follow — `AcquirePIDFile` dispatches straight into `//go:build windows` code (`pidExeBaseName`, `ForceKill` → `TerminateProcess`), and both paths whose values it now logs come from `paths_windows.go`. Absence of a build tag in the changed *files* says nothing about the platform-conditionality of the *path*.

What actually settles it: `test` runs `ubuntu-latest` + `windows-latest` on every PR (`test.yml:160`), so `internal/daemon` — including all three new tests — is compiled and executed on Windows already. `ci:full` buys macOS *tests* (macOS only compiles by default) and `-race` everywhere. This diff adds no macOS-specific code, and the new test file adds no concurrency and no `t.Parallel`, so neither is load-bearing here. **No label needed.** (Noting for the record that if it ever is, a push fires `synchronize`, not `labeled` — the label has to come off and go back on.)

## 5. `GOOS=windows go vet` — the vetted set does not intersect the diff

The PR reports vet green under `GOOS=windows` for `./internal/daemon/service/...` and `./internal/process/...`. **Neither contains a changed file.** The diff touches `internal/daemon/pidfile.go`, `internal/daemon/server.go` and two test files — all in `internal/daemon`, which I confirmed is blocked by #7005 exactly as described:

```
GOOS=windows CGO_ENABLED=0 go vet ./internal/daemon   → exit 1
  internal/treesitter/ts/official/official.go:77:19: undefined: tsofficial.Language
  … build constraints exclude all Go files in …tree-sitter-rust/bindings/go
```

Practical risk is nil — the change uses only `io` and `log/slog`, and the Windows CI leg compiles and runs the package with CGO on. But the sentence as written reads as "the changed code was vetted for Windows", and it was not. Please say what is true: *#7005 blocks vetting the changed package; Windows coverage comes from the `test (windows-latest)` leg.*

## 6. #7053 / #7054 deferred — right call on #7054, mispriced on #7053

**#7054** (basename can't separate four roles behind one `grafel.exe`) — correctly deferred. The proposed fix is a pidfile *format* change (pid + creation time) touching every `engine.pid` consumer. Not this PR.

**#7053** (`watchreg.Sweep` kills with no identity check) — "too expensive here" is not what the code says. `SweepDeps` already exists as the injection point, already carries `Alive`, `Kill`, `LiveDaemonPID`, and already exists precisely "so the staleness/reaping logic can be unit-tested with fake PIDs". The fix is one more field (`IsGrafel func(int) (bool, error)`), one guard in the orphan branch, and a `pid != live` skip — I count roughly 12 lines of production code, with the fake-pid harness already built. That is the same order as this PR's log line. Deferring it is a perfectly legitimate *scheduling* choice — one change per review round is good discipline — but let us record it as that, not as cost.

One correction for #7053's text while it is open: `SweepResult.Orphaned`'s doc says the pids are "SIGTERM'd", and `Sweep`'s doc says "live-but-orphaned entries are SIGTERM'd and dropped". `deps.Kill` is `process.Kill`, which on Windows is `TerminateProcess` — a hard kill that runs no defers, i.e. the same silent-death signature as the reclaim path. Worth fixing in the comments so the next reader does not rule this path out on the strength of a word that is false on the platform where it matters.

---

## Could not verify (macOS only — these are UNRESOLVED, not diagnoses)

- Whether any second `serve` started in the reporter's windows. **Diagnostic:** the `Disable-ScheduledTask` experiment already proposed — plus the `pidfile-acquire begin` question from §1.
- Whether `RestartOnFailure` covers an on-demand `/run` instance at all. **Diagnostic:** enable `Microsoft-Windows-TaskScheduler/Operational` before the repro.
- Whether `ForceKill` can fail with `ACCESS_DENIED` against a daemon in another session on Windows. **Diagnostic:** M7's missing test would at least pin our *response* to that failure, which is the half we control.
- Why `daemon.log` received nothing from a provably-running `Start-Process` daemon. **Diagnostic:** §1's added question. This one gates the value of the whole PR on the reporter's machine.

---

## Verdict

**Request changes** — on the tests, not the behaviour. The behaviour is right and I would take it today with:

1. a test stubbing `forceKillFunc` to fail, asserting the WARN still fires with `kill_err` set (kills M7);
2. a direct test on `socketIsHealthy` counting probe invocations (kills M3, M4, M9);
3. the vet sentence in the PR body corrected to say `internal/daemon` could not be vetted for Windows and why.

And separately, off the diff and more important than any of it: the "raising the budget manufactures the second `serve`" argument on #7051 needs retracting. Do not merge before `test (windows-latest)` reports.

cajasmota (commented):
## Round 2 — head `95c7852c`, own worktree, macOS

The `io.MultiWriter` finding is real and is the best thing in this PR. The corrections to the `ci:full` and `GOOS=windows` claims are accurate readings — I re-checked both. CI is green on all legs including `test (windows-latest)` (20m6s).

**Not mergeable yet.** Two items, one of them a behaviour regression I reproduced here, both small.

Baseline on this head, private `GOCACHE`, fresh `GRAFEL_HOME` + `GRAFEL_DAEMON_ROOT`, `-count=1`:

```
ok  internal/daemon   41.6s
ok  cmd/grafel       206.9s
```

`cmd/grafel` green covers the #6118 digest pins (`custom_extractor_spans_6118_test.go` and the other seven digest-bearing files) — the logger change moved nothing. Verified, not assumed.

---

## BLOCKER 1 — the M7 refusal cannot tell "already dead" from "access denied", and blocks startup on the #5710 scenario itself

You asked whether the refusal can strand a pid it cannot clear. It can, and not only in the exotic case. I reproduced it on macOS against the **real** `process.ForceKill`:

```
err = pidfile reclaim: force-kill of unresponsive owner pid 82575 failed,
      refusing to take the pidfile: signal(82575, SIGKILL): os: process already finished
release = nil    pidfile still = 82575
```

The sequence `AcquirePIDFile` actually performs is:

1. `pidIsLiveDaemonFunc(existing)` → true — the pid **is** alive at this instant
2. `socketIsHealthy(...)` → ~900 ms of probes plus 100 ms of sleeps
3. `forceKillFunc(existing)`

If the incumbent finishes its graceful shutdown during step 2, then at step 3 the pid is gone and `ForceKill` returns `ESRCH` / `os.ErrProcessDone` — a **non-nil error** — and the new `if killErr != nil` refuses the acquire. `grafel serve` does not start.

That window is not a corner case: it is exactly #5710's scenario, a daemon stuck in graceful shutdown, which is the situation the reclaim was built for. And this is not Windows-specific — `process_darwin.go` and `process_linux.go` both go through `proc.Signal(SIGKILL)`, so it fires on all three platforms.

It is transient across *attempts* (the next `serve` sees a dead pid, treats it as stale, proceeds). But nothing retries: `ensureLoaded` does not re-fire `serve`, and per #7051 the `/run` exit code is still ignored. So on Windows one unlucky race leaves the daemon down until a human intervenes — which is the exact user-visible failure this PR exists to serve. Trading a silent double-daemon for "the daemon does not come up" is the wrong direction, and it is a regression against `8383a52f`, where the kill error was ignored and this path proceeded.

**Fix, ~3 lines:** after a failed kill, re-probe liveness and treat "gone" as cleared.

```go
if killErr != nil && pidAlive(existing) {
    return nil, fmt.Errorf(...)
}
```

`pidAlive` is already in this file. That keeps the ACCESS_DENIED refusal you want (the owner is still there) and drops the refusal when the kill "failed" because the process had already exited (nothing to protect). Please also pin it: a test with a reaped pid, `pidIsLiveDaemonFunc` faked true, real `forceKillFunc`, asserting the acquire **succeeds**.

Worth logging which of the two happened, too — `owner_still_alive=true/false` next to `kill_err` is the field that tells a user "we could not kill it" apart from "it beat us to it".

---

## BLOCKER 2 — the fanout's central promise is completely ungraded (N2 ALIVE)

This is the one that worried me most and it is worse than I expected. `fanoutWriter` exists to guarantee *every* sink receives the bytes. Adding a `break` after the first **successful** sink survives the entire `cmd/grafel` suite:

```go
		delivered = true
		break        // <- mutant
	}
```

`go vet` clean, `go test -count=1 ./cmd/grafel` **green (full package, not a subset)**.

In production the sinks are `(os.Stderr, logFile)` in that order. On any healthy machine stderr succeeds first, the loop breaks, and **`daemon.log` is never written for the whole run** — the identical symptom this PR was written to fix, now reachable on every platform rather than only after `FreeConsole`. Nothing notices.

The cause is the shape the board keeps hitting: all four fanout tests vary *which* sink fails and hold one axis constant — **no test has a first sink that succeeds together with a later sink whose content is then verified**.

- `FailingSinkDoesNotSilenceTheOthers` — `(errWriter, file)`, first sink fails
- `FailingSinkLast_StillReportsSuccess` — `(file, errWriter)`, checks only the file
- `AllSinksFail_ReportsError` — no sink succeeds
- `BuildDaemonSlogLogger_SurvivesADeadConsoleSink` — `(errWriter, file)`, first sink fails

**Fix, ~8 lines:** two working buffers, one write, assert **both** received it. That is the test that makes the type mean what its doc comment says.

---

## N1 ALIVE — `errNoSinks` is asserted in prose and observed by nothing

```go
if len(f.sinks) == 0 { return 0, errNoSinks }   →   { return len(p), nil }
```

`go vet` clean, **full `./cmd/grafel` green**. The comment says *"A silently-successful write to zero sinks is the same lie this type exists to stop telling"* — and no test constructs `newFanoutWriter()` with no sinks. Four lines to close. Low severity (no zero-sink caller exists today: `daemon.go:745` passes two, `selftest.go:353` passes one), but it is a documented guarantee with no observer, and `newFanoutWriter` is variadic precisely so a future caller can get it wrong.

---

## New mutant table — 6 rows, call site, permissive-biased

Each applied by `cp` from a shasum-verified backup (`fan`/`dae`/`pid` backups taken and printed), asserted to add **exactly one** occurrence of the mutated text, `go vet` green before scoring. Every ALIVE verdict is from the **full** package, never a `-run` subset. `git status --porcelain` and `git diff HEAD` both empty afterwards, tree back at `95c7852c`.

| # | File | Mutant | Verdict |
|---|---|---|---|
| **N2** | `logfanout.go` | `break` after the first **successful** sink (on a healthy box: stderr wins, `daemon.log` never written) | **ALIVE** — full `./cmd/grafel` green. Blocker 2. |
| **N1** | `logfanout.go` | zero sinks reports `len(p), nil` instead of `errNoSinks` | **ALIVE** — full `./cmd/grafel` green |
| N3 | `logfanout.go` | `newFanoutWriter` drops the last sink (`sinks[:len(sinks)-1]`) | **DEAD** — `FailingSinkDoesNotSilenceTheOthers`, `BuildDaemonSlogLogger_SurvivesADeadConsoleSink` |
| N5 | `logfanout.go` | mark `delivered` even when the sink errored (report success with nothing written) | **DEAD** — `AllSinksFail_ReportsError` |
| N6 | `daemon.go` | builder wraps only the first sink (`newFanoutWriter(sinks[0])`) | **DEAD** — `BuildDaemonSlogLogger_SurvivesADeadConsoleSink` |
| N4 | `pidfile.go` | unhealthy return re-derives the constant: `return false, socketHealthProbeRetries + 1` | **ALIVE but equivalent under the suite** — see below |

**N4 is not a finding.** It is indistinguishable only because `socketHealthProbeRetries` is 2, which makes the re-derived constant numerically equal to the count on the all-miss path. The constant itself is pinned independently by `TestSocketIsHealthy_ProbeBudget`'s literal `wantAttempts = 3` plus the per-probe timeout assertions, so the M9 regression genuinely cannot come back: setting `socketHealthProbeRetries = 0` fails that test whether or not N4 is applied. Recording it as equivalent, with the reason, rather than as a hole.

**M9b checked as you asked — it is stronger than M9 intended, not weaker.** M9 would only have proved "the field is not a hardcoded literal". What shipped pins the value twice and independently: `ProbeBudget` asserts the loop's own behaviour against literals, and `FieldsMatchWhatActuallyHappened` compares `probe_attempts` against a probe count the *test* accumulates in its own stub. Re-running rather than scoring a non-compiling mutant was the right call.

---

## Does the `io.MultiWriter` mechanism survive scrutiny? Yes — with one qualification that should go on #7050

I tried to break it and could not. Reporting the parts separately, since the thread is already published:

**Verified here.** `io.MultiWriter` stops at the first erroring sink — and rather than only asserting that in a comment, `TestIOMultiWriter_StopsAtTheFirstFailingSink` pins the stdlib premise itself, so the fix's justification cannot silently go stale. That is the right way to depend on someone else's behaviour. The ordering claim also holds: `detachConsole()` is at `daemon.go:662` inside `if !foreground`, the logger is built at `daemon.go:745`, 83 lines later. `slog`'s handler discards the writer's error. So one failing stderr write did silence `daemon.log` for the entire process lifetime, and the daemon served normally throughout. **That is a real bug on its own terms**, independent of whether it is #7050's bug.

**The mechanism cleanly explains repro 2 and 3.** `Start-Process -WindowStyle Hidden` allocates a console (hidden is still a console), so `os.Stderr` is a valid console handle at start; `FreeConsole` then invalidates it; every subsequent stderr write fails; `io.MultiWriter` aborts before reaching the file. That predicts precisely "provably running, zero lines in an append-only `daemon.log`, not even `pidfile-acquire begin`".

**It does not explain repro 1.** Repro 1 was `Start-ScheduledTask` — the wscript `//B` action, which has no console for `FreeConsole` to invalidate — and #7051 step 5 reports a task-launched run writing *a full normal startup sequence* to `daemon.log`. Both cannot be true of the same launch path. That tension predates this PR and I cannot resolve it from macOS, but the #7050 comment should not read as though the mechanism covers all three repros. Please qualify it to repro 2/3.

**A cheaper, stronger confirming diagnostic than "try the new build".** This one needs no new binary and tests the mechanism against the *shipped* `v0.3.3`:

```powershell
Start-Process -FilePath "$env:USERPROFILE\.grafel\bin\grafel.exe" -ArgumentList "serve" `
  -WindowStyle Hidden -RedirectStandardError "$env:TEMP\grafel-stderr.txt"
```

Redirecting stderr to a **file** gives the process a handle `FreeConsole` does not invalidate. If `daemon.log` then fills where it previously stayed empty — with the *old* binary — the mechanism is confirmed directly, which is far better evidence than a new build merely working. The negative control is redirecting only `-RedirectStandardOutput`: stderr stays on the console, and `daemon.log` should still be silent.

### Smaller notes on the fanout, none blocking

- **Short writes.** `Write` ignores each sink's returned `n`, so a sink that accepts 3 of 30 bytes with a nil error counts as delivered and the call reports `len(p), nil`. Documented as deliberate, and unreachable today because every production sink is an `*os.File`, whose `Write` converts a short write into `io.ErrShortWrite` itself. Alive but not production-reachable; the distinguishing input is any non-`*os.File` sink, and `buildDaemonSlogLogger` is variadic. Worth one clause in the comment saying *why* it is safe (all sinks are `*os.File`) rather than only that it is intended.
- **Return value.** Reporting `len(p)` on partial delivery technically breaks the `io.Writer` contract. No caller depends on it (`slog`'s handler discards `n` and the error), and `selftest.go` is the only other caller. Fine as documented.
- **No retry loop, no unbounded behaviour.** Each `Write` tries each sink exactly once. Nothing is retried forever.
- **The inverse hole is silent.** If `daemon.log` is the failing sink and stderr succeeds, the write reports success and nothing anywhere records that the *durable* sink is dead. That is the mirror image of the bug being fixed. Not a blocker, but a one-shot WARN to the surviving sinks naming the sink that failed would close it cheaply.
- **All-succeed path is unchanged** — every sink written, `len(p), nil`, same as `io.MultiWriter`. Confirmed.

---

## Corrections re-checked

- **`ci:full`** — the new wording cites `test.yml:160` and the reasoning now matches the file. Correct, and correctly no longer resting on "no build tag changed". No label needed.
- **`GOOS=windows go vet`** — now states plainly that `internal/daemon` and `cmd/grafel` could not be Windows-vetted here because of #7005, with Windows coverage coming from the test leg. That matches what I reproduced in round 1. Accurate.

## Could not verify (macOS — UNRESOLVED, not diagnoses)

- Whether `os.Stderr` writes actually fail after `FreeConsole` on the reporter's Windows. **Diagnostic:** the `-RedirectStandardError` experiment above, against the old binary.
- Why repro 1 (task-launched) wrote nothing while #7051 step 5's task-launched run wrote everything.
- Whether `ForceKill` can return ACCESS_DENIED against a same-user daemon in another session. Blocker 1's fix makes this moot for the benign case either way.

## Verdict

**Hold.** Two changes, both small, neither touching the design:

1. `if killErr != nil && pidAlive(existing)` plus a test that a pid which died during the probe window still lets the acquire succeed (Blocker 1 — reproduced, and a regression against the previous head);
2. one test asserting **both** sinks of a two-working-sink fanout receive the bytes (Blocker 2 — N2), and optionally the four lines for `errNoSinks` (N1).

With those, I would merge. The `io.MultiWriter` root cause is good work and I would like it in front of the reporter quickly — but not with a startup path that refuses to start when the incumbent exits a moment early.

cajasmota:
Round-3 push: `117f6e17`. Both blockers fixed, plus N1 and the two non-blocking items.

## BLOCKER 1 — the refusal was a regression; thank you for reproducing it

You are right, and the reproduction is what made it obvious: a non-nil `killErr` does not mean the owner is still there. The liveness check that gates the reclaim and the kill are separated by `socketIsHealthy`'s ~900 ms of probes and sleeps, and an incumbent finishing its shutdown inside that window makes the real `ForceKill` return `ESRCH` on every platform. Refusing there blocks startup in #5710's own scenario, with nothing to retry it.

Now `killErr != nil && pidAlive(existing)`, exactly as you suggested, and the WARN carries `owner_still_alive` so the two cases are distinguishable in the log rather than only in the code. The `ACCESS_DENIED` refusal that motivated M7 is unchanged.

The M7 test now uses a genuinely live child (`spawnLiveChild`) — with a fake pid it would have silently moved to the "it beat us to it" branch and stopped testing the refusal at all.

## BLOCKER 2 — N2, and the shape behind it

`break` after the first successful sink is a better mutant than any of mine, because it is the *production* ordering that makes it fatal: `(os.Stderr, logFile)`, stderr wins on a healthy box, `daemon.log` never written. Your diagnosis of the cause is right and it is the one that keeps recurring here — all four tests varied *which* sink fails and none had a succeeding first sink with a later sink's content verified.

`TestFanoutWriter_EverySinkReceivesTheBytes` is that test: three buffers, one write, all three contents asserted.

N1 closed too — `newFanoutWriter()` with no sinks is now constructed by a test and its error asserted.

## Both non-blocking items addressed

- **Short writes**: the comment now says *why* ignoring them is safe (every production sink is an `*os.File`, whose `Write` converts a short write to `io.ErrShortWrite` and returns it as an error, so a partial write cannot reach the success path unnoticed) and flags that a future non-file sink would need this revisited.
- **The inverse hole**: the first time a sink fails, the failure is announced **on the surviving sinks**, once — not once per record. If `daemon.log` dies while stderr lives, stderr says so, and the error names the sink for free (`os.File` write errors are `*PathError`). That is what #7050's reporter would have found on the other side.

## Round-3 mutants

Call site, permissive-biased, each asserted to apply at exactly 1 occurrence, `go vet` green before scoring, restored by `cp` from shasum-verified backups (`git status --porcelain` and `git diff HEAD` empty after).

| # | Mutant | Verdict |
|---|---|---|
| N8 | refuse on **any** kill error (the regression you found) | **DEAD** — `ReclaimKillFailsBecauseOwnerAlreadyExited_Proceeds` |
| N9 | never refuse — proceed after a failed kill | **DEAD** — `ReclaimKillFails_RefusesAndLogsKillErr` |
| N2 | `break` after the first successful sink | **DEAD** — full `./cmd/grafel` |
| N1 | zero sinks report `len(p), nil` | **DEAD** — `NoSinks_ReportsError` |
| N7 | announce the failure on every record, not once | **DEAD** — `FailedSinkIsAnnouncedOnTheSurvivors_Once` |

N4 left exactly as you scored it: equivalent under the suite, with `ProbeBudget`'s literals pinning the constant independently.

## Verification

`go test -count=1 ./internal/daemon ./cmd/grafel` green (69.7s / 226.8s) with fresh `GRAFEL_HOME` + `GRAFEL_DAEMON_ROOT`. `go vet` green natively for both. `gofmt` clean.

**One thing that changed since your `ci:full` assessment**, flagging rather than deciding: the fanout now takes a `sync.Mutex` (the one-shot notice needs state, and it makes each record atomic across sinks), so the diff is no longer concurrency-free. `go test -race` is green on the fanout tests locally, but if you want `-race` on it in CI that is now a real argument for the label — and the label would have to come off and go back on, since a push fires `synchronize`.

Everything about repro 1 stays as you qualified it publicly; nothing further posted to #7050 or #7051 from here.

cajasmota:
Round-4 push: `00177d23`. CM-2 is dead, and your reading of why `-race` was useless here was exactly right.

## The concurrent test

`TestFanoutWriter_ConcurrentWriters_SerialisedAndNoticedOnce` — 8 goroutines x 40 records through one fanout whose **middle sink always fails**, so `reported` / `emittingNotice` are exercised concurrently rather than just existing.

Assertions are interleaving-independent by construction, per your warning about racy premises:

- the **set** of records each surviving sink holds, each exactly once — never their order, which is not an invariant;
- the one-shot notice appears **exactly once** per surviving sink, however many goroutines raced into the failure path together;
- an **overlap counter**: a sink that records whether two writers were ever inside it at the same time. While the mutex is held across the whole `Write` that is impossible, so this assertion can only fail in one direction — it never fires on correct code, and a non-zero count is proof rather than luck.

The surviving sinks keep their bytes in a deliberately **unsynchronised** `bytes.Buffer`. That is the part that gives `-race` something to observe: the detector was not weak here, it was simply pointed at single-threaded code.

## CM-2 re-scored

Mutex deleted from `Write` (`f.mu.Lock()` occurrences 1 → 0, asserted), `go vet` green before scoring:

| Run | Verdict | Signal |
|---|---|---|
| `go test -count=1` (no `-race`) | **DEAD** | 278 overlapping writes observed, **and** records lost — `"rec-0-4"` present 0 times of 320 |
| `go test -race -count=1` | **DEAD** | 19 `DATA RACE` reports, at `logfanout.go:91-92` (`s.reported` read-modify-write) and `:121` (`s.reported` read inside the notice) — precisely the fields the lock protects |

**It dies without `-race` too**, which is the better of the two outcomes you offered: the ubuntu and windows legs that run on every PR now catch this, not only the `ci:full` `-race` leg. That also means the label's value here does not rest on the mutex — your assessment that it buys little stands, and I'd keep it for the same reason you gave (right in principle, cheap) rather than for this.

Two independent signals kill it without `-race` — lost records and observed overlap — so the kill does not depend on the buffer corrupting in any particular way.

## Verification

`go test -count=1 ./cmd/grafel ./internal/daemon` green (220.9s / 46.8s) with fresh `GRAFEL_HOME` + `GRAFEL_DAEMON_ROOT`; `-race` green on the fanout tests. `go vet` and `gofmt` clean. Tree restored from a shasum-verified backup, `git status --porcelain` and `git diff HEAD` empty.

**Production code unchanged this round — tests only.** Nothing posted to #7050 or #7051.

cajasmota (commented):
## Round 3 (final) — head `00177d23`, own worktree, macOS

**Verdict: MERGE**, once the in-flight `ci:full` matrix reports green. I have no outstanding objection to the code. Both round-2 blockers are properly fixed — not patched to satisfy my wording — and the one mutant that survived is a defensive guard on an unreachable path, which is a comment nit, not a defect.

Baseline on this head, private `GOCACHE`, fresh `GRAFEL_HOME` + `GRAFEL_DAEMON_ROOT`, `-count=1`:

```
ok  internal/daemon    51.8s
ok  cmd/grafel        255.8s
```

At the time of writing, `test (ubuntu-latest)` has passed on both runs; `test (macos-latest)` and `test (windows-latest)` are still pending on the `-race` matrix. That is the only gate left.

---

## 1. Blocker-1 fix reviewed, not just accepted

I prescribed the clause; here is the review of what shipped.

**`ownerStillAlive := killErr != nil && pidAlive(existing)` — correct, and the predicate choice is right.**

The asymmetry with the earlier `pidIsLiveDaemonFunc` is deliberate and it is the right call, for three reasons:

- **Identity is already settled.** The is-it-grafel question was answered before the probe window. The only open question after a failed kill is "did our kill land", and bare liveness is exactly that question. Re-running the identity check would answer a question nobody asked.
- **On Windows the two predicates read different access rights, and the split is precisely what makes this work.** `ForceKill` needs `PROCESS_TERMINATE`; `pidAlive` → `process.IsAlive` opens `PROCESS_QUERY_LIMITED_INFORMATION` and — importantly — **returns true on `ERROR_ACCESS_DENIED`**, on the stated reasoning that the kernel resolved the pid in order to deny us. So in the ACCESS_DENIED scenario the refusal exists for, `pidAlive` correctly reports the owner as alive. The predicate is right in the one case that matters, for a reason the code already documents.
- **It is not stubbable, and that is a feature.** `pidIsLiveDaemonFunc` is faked by `withFakePidIsLiveDaemon`; `pidAlive` is not. Using the indirection here would let a test's fake mask real liveness and quietly re-open the branch. **P2 proves the suite catches exactly that substitution** — swapping `pidAlive(existing)` for `pidIsLiveDaemonFunc(existing)` is DEAD. The asymmetry is graded, not accidental.

**Can it strand a live-but-unkillable owner?** Yes — deliberately, and I think correctly. A genuinely alive, genuinely unkillable, genuinely unresponsive owner now blocks startup indefinitely. That is the intended trade against two-daemons-one-pidfile, and the mitigation is that the error names the pid and now says *why*: "…failed **and it is still alive**, refusing to take the pidfile". That is actionable (kill it elevated) where the previous silent divergence was not. I am satisfied this is the right side of the trade.

**One residual I convinced myself is not worth changing:** between `forceKillFunc` returning an error and `pidAlive` running, a pid could in principle be recycled onto an unrelated process, producing a false refusal. The window is microseconds, it requires the kill to have failed *because* the process vanished and the OS to reuse that exact number immediately, and it self-heals on the next attempt (`pidIsLiveDaemonFunc` then rejects the recycled pid as not-grafel and the pidfile is treated as stale). Not worth a guard.

**`owner_still_alive` reports the branch taken, verified.** It is the same variable the `if` tests, computed once. **P4** — replacing the logged value with a re-derived `killErr != nil` — is DEAD. Not re-derived.

One small honesty note, not blocking: when `killErr == nil` the `&&` short-circuits and `pidAlive` is never called, so `owner_still_alive=false` there means "we did not check" rather than "we observed it gone". The inference is sound (the kill succeeded), but the field reads as an observation in all four field-combinations. A clause in the comment would close the gap between what it says and what it measured.

## 2. The M7 test genuinely exercises the refusal, and cannot go vacuous the same way

This was the right thing to be worried about — it is the #7055 shape exactly, and the fake pid `424242` would have drifted silently into the "already gone" branch under the new condition, leaving a test named `...RefusesAndLogsKillErr` that never reached the refusal.

What shipped is sound on both counts:

- `spawnLiveChild` gives a genuinely live pid, so `pidAlive` returns true and the refusal branch is the one executed. `forceKillFunc` is stubbed, so the child is never actually killed and stays alive for the duration.
- **It carries its own anti-vacuity pin**: `if alive, _ := found["owner_still_alive"].(bool); !alive { t.Fatalf(...) }`. If this test ever drifts into the "it beat us to it" branch again, `owner_still_alive` is false and the test fails rather than passing for the wrong reason. That is the assertion that makes the regression self-announcing, and it is the part I would have asked for.
- The mirror, `...OwnerAlreadyExited_Proceeds`, uses a reaped pid with the **real** `forceKillFunc` and asserts both the success and `owner_still_alive=false`. The two tests pin opposite branches through the same code with different real-world inputs.

**P3** (never refuse — full M7 revert) is DEAD via the refusal test; **P1** (drop the liveness re-probe — the round-2 regression) is DEAD via the already-exited test. Both directions are now held.

## 3. The failure notice — attacked, holds

Walked every path you named:

- **Lost?** No. A sink's first failure is captured into `newlyFailed` and `reported` is set in the same locked section, and the notice is emitted before `Write` returns. The one case where no notice is emitted is when `delivered == false` — every sink failed — and there is by definition nowhere to write it. Correct, not a loss.
- **Duplicated?** No. `reported` is per-sink and set on first failure, so a permanently dead sink yields one notice for the life of the process, not one per record. **P6** (append to `newlyFailed` unconditionally, announcing on every record) is DEAD.
- **Written to a sink already known dead?** No. `noticeLocked` skips `s.reported`. **P7** (remove that skip) is DEAD.
- **Recurse?** It cannot, but not for the reason the comment gives — see P5 below.
- **A sink that fails *during* the announcement?** Handled acceptably: the write error is discarded and `reported` is not set, so that sink is picked up as newly-failed on the next record and announced then. Nothing is lost, and nothing loops.
- **Deadlock?** `noticeLocked` is called with `f.mu` held and never re-acquires it. Clean.
- **Does the injected plain-text line break a `daemon.log` consumer?** I checked rather than assumed: `internal/cli/feedback_timeline.go:443-444` does `json.Unmarshal(...); if err != nil { continue }`, so a non-JSON notice line is skipped, not fatal. Tolerated under `GRAFEL_DAEMON_LOG_JSON=1` as well as the default text handler.

## 4. New mutant table — 7 rows, call site, permissive-biased

`cp` restore from shasum-verified backups (`r3_fan`, `r3_pid`, both printed on first run); each edit asserted to add **exactly one** occurrence of the mutated text, with the anchor asserted at exactly one occurrence first; `go vet` green before every score; `-count=1`; fanout mutants screened under `-race`; every verdict from the **full** package, never a `-run` subset. `git status --porcelain` and `git diff HEAD` both empty afterwards, tree back at `00177d23`.

| # | File | Mutant | Verdict |
|---|---|---|---|
| P1 | `pidfile.go` | drop the liveness re-probe (`killErr != nil` alone — the round-2 regression) | **DEAD** — `...OwnerAlreadyExited_Proceeds` |
| P2 | `pidfile.go` | re-probe with the **stubbed** `pidIsLiveDaemonFunc` instead of real `pidAlive` | **DEAD** — `...OwnerAlreadyExited_Proceeds` |
| P3 | `pidfile.go` | never refuse (`if false && ownerStillAlive`) | **DEAD** — `...RefusesAndLogsKillErr` |
| P4 | `pidfile.go` | log a re-derived `killErr != nil` as `owner_still_alive` | **DEAD** — `...OwnerAlreadyExited_Proceeds` |
| P6 | `logfanout.go` | announce on every record, not once per sink | **DEAD** — `...AnnouncedOnTheSurvivors_Once`, `...ConcurrentWriters...` |
| P7 | `logfanout.go` | write the notice to sinks already known dead | **DEAD** — `...FailingSinkDoesNotSilenceTheOthers`, `...ConcurrentWriters...` |
| **P5** | `logfanout.go` | delete the `emittingNotice` recursion guard | **ALIVE — equivalent; the guard is unreachable** |

**P5 is not a hole, and I do not want a test written for it.** `noticeLocked` writes directly to `s.w`, never through `f.Write`, so there is no path that re-enters it — the protection against recursion is structural, not the flag. And the one construction that could recurse (a fanout passed as its own sink) would deadlock on the non-reentrant `f.mu` *before* `emittingNotice` was ever consulted, so the flag would not save that case either. The correct resolution is to fix the comment — "cannot recurse **because notices bypass Write**" — and either keep the flag as cheap belt-and-braces or drop it. Deleting it changes no behaviour and adding a test would pin a mechanism that does nothing. Flagging it only because the comment currently claims a protection the flag does not provide.

## 5. Is the concurrency test's premise deterministic? Yes — by construction, and it holds up under stress

This was worth checking and it survives. Every assertion is interleaving-independent **for correct code**:

- **`overlaps == 0`** is one-directional: with `f.mu` held across the whole of `Write` plus `noticeLocked`, two writers can never be inside a sink together, so zero is guaranteed, and non-zero is proof rather than luck. It cannot false-fail.
- **The record-set assertion is exact, not probabilistic.** Serialised writes mean each record reaches each surviving sink exactly once, so `strings.Count(got, rec) == 1` is deterministic. I checked the one thing that could make it accidentally wrong — substring collisions between `rec-3-1` and `rec-3-11` — and the trailing `\n` in the payload disambiguates them. The counts are sound.
- **Exactly-one notice** is deterministic under the mutex: the first Write that touches the dead sink sets `reported`, every later Write finds `newlyFailed` empty.
- **`dead.n == 320`** is exact: every record attempts every sink once, and notices never target a reported sink.
- **`t.Errorf` in the goroutines, `t.Fatalf` only on the main goroutine** — the right way round.

The only scheduling-dependent property is *sensitivity* when the mutex is removed, and that does not rest on the overlap counter alone: lost and duplicated records are a far louder signal, which matches your own no-`-race` DEAD result. The unsynchronised `bytes.Buffer` is the right instrument — under correct code every access is ordered by `f.mu`, so a happens-before edge exists and `-race` is clean, while a removed mutex gives the detector something real to see.

I stress-ran it to check practice matched theory: **`go test -count=30 -race -run TestFanoutWriter_ConcurrentWriters ./cmd/grafel` — green, 30/30.** Not scheduling luck.

Your CM-2 observation is the durable lesson here and deserves to outlive this PR: a `-race` leg over single-threaded tests reports nothing by construction, so "green under `-race`" was never evidence that the mutex was graded. Worth carrying forward.

## 6. The `ci:full` mechanics you hit

Your read is right and matches `test.yml`: the job's `if:` gates on `pull_request_target` **with the label**, and a `synchronize` push does not re-evaluate that the way a `labeled` event does — which is why the label sitting on the PR produced `compile (macos-latest)` but no macOS *test* leg. Remove-and-re-add is the correct remedy. I flagged this in round 1 as a footnote; it is worth promoting into the workflow's own comment block so the next person does not have to rediscover it, and that is the one follow-up I would file off the back of this PR.

## Could not verify (macOS — UNRESOLVED, not diagnoses)

- Whether `os.Stderr` writes actually fail after `FreeConsole` on the reporter's Windows. The `-RedirectStandardError` experiment against the **old** binary is still the cheapest confirmation, and it is the one that would settle the mechanism rather than merely showing the new build works.
- Why repro 1 (task-launched) wrote nothing while #7051 step 5's task-launched run wrote everything. Agreed the thread should carry the repro-2/3-only qualification.
- Whether `ForceKill` returns ACCESS_DENIED against a same-user daemon in another session. The fix is correct under either answer.
- The `-race` matrix on macOS and Windows was still running when I finished.

## Verdict

**MERGE** when the in-flight `ci:full` matrix is green. No code change is required and I am not asking for one. Optional follow-ups, none of which should hold this: the `emittingNotice` comment (P5), the `owner_still_alive` short-circuit clause, and promoting the `labeled`-vs-`synchronize` gotcha into `test.yml`'s comments.

Three rounds in, the thing I would actually note for the record: every behaviour change in this PR now has a mutant on both sides of the decision it makes, and the two tests that pin the refusal fail loudly if they ever drift into each other's branch. That is the property that was missing in round 1.

Issue #7056, opened 2026-09-15
P1: every quality instrument reported a PERFECT result on code where half the edges were wrong — and forbidden_relationships, the one mechanism that could express it, is always-fatal so a shipping defect cannot be recorded


cajasmota:
Every quality instrument we have reported a **perfect** result on code where **half the edges were wrong**. Measured on the swift arm of #6912 (PR #7043) by toggling only the carrier refusal:

```
FIXED     total=19  bound=19  dangling=0  wrongName=0  wrongFile=0
UNFIXED   total=38  bound=38  dangling=0  wrongName=0  wrongFile=0
```

Nineteen of those 38 edges pointed at an `extension Foo` carrier instead of the declaration of `Foo` — `BaseNEncoding`×7, `String`×6, `RoutesBuilder`×2, `HTTPClient`/`PasswordHasher`/`Request`/`tm`×1. A field typed `Request` in `FileIO.swift` bound to that file's extension carrier while the real `struct Request` in `Request.swift` received no inbound edge at all.

**Nothing flagged it**, because the carrier has the right name, the right file, an admissible kind and an admissible subtype. It is wrong in the only dimension nothing measures: **it is not a declaration.**

## Why each instrument missed it

| Instrument | Why it is blind |
|---|---|
| bind rate / `dangling` | The edge resolves. A wrong target is still a target. |
| `wrongName` / `wrongFile` | Both correct — the carrier genuinely has that name in that file. |
| `bug-extractor` | Classifies **unresolved** stubs. A confidently-wrong edge is not a stub. |
| `semanticDigest6118` | Detects *change*, not *wrongness* — and holds **0** `.swift` files. |
| golden ratchet | Vacuous for swift: 6 files / 17 fields / **0 edges**. It even contains `extension User` beside `struct User` and still gains nothing. |
| **#6726's orphan rate** | **Actively rewards the defect** — the 19 wrong edges make 19 fields stop being orphans. The metric improves. |

That last row is the one that should worry us: the contributor-facing number this milestone is being steered by **goes up** when this bug is present.

## The important correction: this is a COVERAGE gap, not an expressiveness gap

My first instinct was to file this as "the harness cannot express the class." **That is wrong**, and the distinction changes what to build.

[...]

Issue #7047, opened 2026-09-08
swift: swiftDeclSubtype has no 'extension' case and defaults to "class", so 'extension String {}' is indistinguishable from a same-file class declaration — 19 of 38 field->type edges bind to an extension carrier


cajasmota:
`swiftDeclSubtype` recognises four declaration keywords and has no case for `extension`, so an `extension Foo { … }` mints a `SCOPE.Component` with `Subtype: "class"` named `Foo` — indistinguishable, to every downstream consumer, from `class Foo { … }` declared in that file.

## The code

`internal/extractors/swift/swift.go:236`

```go
func swiftDeclSubtype(node ts.Node, src []byte) string {
	for i := 0; i < int(node.ChildCount()); i++ {
		ch := node.Child(i)
		if ch == nil || ch.IsNamed() {
			continue
		}
		switch string(src[ch.StartByte():ch.EndByte()]) {
		case "struct":  return "struct"
		case "enum":    return "enum"
		case "actor":   return "actor"
		case "class":   return "class"
		}
	}
	return "class"
}
```

`extension` matches no case, so the loop completes and the function returns its `"class"` default. The default is doing double duty: it is both "a `class` declaration" and "anything unrecognised", and those are not the same claim.

## Why it matters beyond a mislabelled subtype

An extension is **not a declaration**. `extension String { … }` in `Util.swift` does not mean `String` lives in `Util.swift` — it means someone added members to a type declared elsewhere, very often in another module entirely. But the emitted record says `String` is a class declared in `Util.swift`, so **any consumer asking "is type T declared in this file?" gets the wrong answer**, and gets it as a confident yes rather than as a miss.

Measured consequence, on the swift arm of #6912 (PR #7043) over 234 corpus `.swift` files:

- **19 of 38 field→type edges — exactly half — bind to an extension carrier rather than a declaration.**
- Six target `String` (`extension String`), one targets `tm` (`extension tm: @retroactive`, a libc struct), the rest `BaseNEncoding`, `Request`, `HTTPClient`, `PasswordHasher`, `RoutesBuilder`.
- A two-file probe through the real resolver: a field typed `Request` in `FileIO.swift` binds to that file's **extension carrier**, a different entity ID from the real `struct Request` in `Request.swift` — which receives no inbound edge at all.

[...]

Issue #7046, opened 2026-09-08
12 cpp custom gates admit file.Language == "c", which is routed to no prefix — and the #6978 reachability scan is blind to language-equality gates by construction


cajasmota:
Twelve gates in `internal/custom/cpp` admit `file.Language == "c"`. **The language `"c"` is routed to no custom-extractor prefix at all**, so every one of those disjuncts is dead — and the test built to catch exactly this class of defect cannot see them.

## The two halves, both verified from source

**1. `"c"` is routed nowhere.** `customPrefixForLanguage` and `extraCustomPrefixesForLanguage` (`internal/extractors/custom_dispatch.go`) have no `"c"` key. `RunCustomExtractors` selects purely on `file.Language`, so a file classified `c` is handed to no custom extractor in the repository.

**2. Twelve gates admit it anyway.** All in `internal/custom/cpp`, across 8 files and 12 keys:

```
driver_schema.go:123      if file.Language != "cpp" && file.Language != "c" {
nlohmann_json.go:78       ... || (file.Language != "cpp" && file.Language != "c") {
orm.go:119, :296, :434    if file.Language != "cpp" && file.Language != "c" {
orm_sql_wrappers.go:282, :341, :398
redis_query.go:101
type_alias.go:81
ros_extractor.go:94       isCPP := file.Language == "cpp" || file.Language == "c"
unreal_extractor.go:105   isCPP := file.Language == "cpp" || file.Language == "c"
```

**No pass is disabled by this.** Every one keeps a live `"cpp"` arm, satisfied by `.cpp/.cc/.cxx/.hpp/.hxx`. The `"c"` half is simply inert.

## Why it is worth an issue rather than a tidy-up

`.h` classifies as `c` (`classifier.go:389`), so **header-only C/C++ shipped as `.h` is invisible to the entire cpp custom lane** — the twelve authors who wrote `|| "c"` were reaching for precisely that case, and got no error telling them it does nothing.

The question this raises is a real fork, and it should be decided rather than defaulted:
- **If the disjuncts are intent**, `"c"` needs a routing entry — and that newly exposes 12 passes to every `.c` and `.h` file in every corpus, which must be measured before it lands, not after. Over-firing is the failure direction this board keeps paying for (#7013).
- **If they are copy-paste**, they should be deleted, because a dead disjunct reads as coverage.

Either way, twelve sites currently assert a capability that does not exist.

## The blind spot is the more durable defect

[...]

Issue #7041, opened 2026-09-08
COMPLETE — all four arms (go, java, rust, C#) no longer emit a WRONG BINDING edge for a field typed by a type parameter; four languages, four incompatible scoping rules, none portable


cajasmota:
Shipped in **v0.3.3** via #7036 (arm D of #6912). The go field→declared-type pass **does not understand type parameters**, so a generic type parameter whose name collides with a same-file declared type produces a **wrong binding** — an edge asserting a field's declared type is something it is not.

## The shape

```go
package models

type T struct{ Unrelated string }

type Box[T any] struct {
	Item T
}
```

`Box.Item`'s candidate is `T`. Normally that is dropped because nothing in the file is *declared* `T` — but here `type T struct{…}` is, so the candidate matches and the edge is emitted. It asserts `Box.Item`'s declared type is that struct. **It is not; `T` is the type parameter, which shadows the outer declaration inside `Box`.**

## Direction: wrong binding, not under-emission

This is the worse of the two failure directions and the one arms A–E were all written to avoid. A dangling stub gets kept verbatim and classified `bug-extractor`, so it is visible; a **bound** edge to the wrong entity looks correct to every consumer and to `bug-extractor` alike. It is the exact hazard that made arm D skip `resolveTypeReferences` (which would have reduced `other.Order` → `Order`) and made arm E skip qualified names whole.

## It is pinned, deliberately and well

`TestGoFieldTypeRefs_KnownOverFire_TypeParameterShadowedByASameFileType` (`internal/extractors/golang/field_type_refs_6912_test.go:516`) asserts the wrong edge **is** present, and carries a `t.Skipf` that fires when it disappears:

> *"KNOWN-WRONG edge is gone … If type parameters are now tracked as a shadowing scope, DELETE this test; it exists only to make the over-fire visible."*

So the defect cannot be fixed silently, and the fixer is told what to do with the pin. That is the same expected-to-break pattern used on #6990, and it is why this is a filed known limitation rather than a latent surprise.

## F# closed its equivalent, which is the argument that this is fixable

Arm E (#7040) hit the same question with F#'s `'T` and **closed it** rather than shipping an over-fire. Go's type-parameter list is available on the AST — the pass simply does not consult it. So this is a gap in the pass, not a limit of the language or of the extractor's inputs.

## Suggested fix

[...]

PR #7057, opened 2026-09-15
feat(#6912): arm I — C/C++ fields get a REFERENCES edge to the same-file type they declare

cajasmota:
## Arm I (C/C++) — #6912. **1 edge on the corpus as the pipeline actually classifies it, and that number is the finding, not the headline.**

Ninth arm: C#, proto, rust, go, java, F#, php, swift, **cpp**.

C/C++ fields now carry the field-anchored `REFERENCES` edge with
`ref_kind=field_target_type`, same-file targets only, addressed by
`scope:component:class:<lang>:<file>:<Name>` where `<lang>` is the FILE's
language — one extractor serves both `c` and `cpp`, so the segment is a variable
and that is graded, not assumed.

---

### 1. The numbers, with the ceiling stated where the count is

| measurement | value |
|---|---|
| fixture suite | **19 edges / 19 bound / 0 dangling**, every endpoint NAMED (`..._ResolvesToEntityIDs`) |
| corpus, **as the pipeline classifies it** | **1 edge / 1 bound / 0 dangling**, 172 C/C++ files, 51 field entities (5 function-shaped, refused) |
| corpus, `.h` parsed as C++ (**a hypothetical, labelled as one**) | **6 edges / 6 bound / 0 dangling**, 254 field entities (68 function-shaped, refused) |

**Revised after review — the hypothetical was 23 and is now 6.** 17 of those 23
hung off a `SCOPE.Schema/field` that is a member FUNCTION declaration or a
pointer-to-member-function member, and are refused as edge SOURCES (§3b). The
real-corpus 1 is unchanged.

**A bind rate is not completeness, and here it is not even the interesting
number.** The recall ceilings, each asserted rather than described: cross-file
types get nothing; `using`-aliases and `typedef`s mint no record at all so a field
typed after one gets nothing; a **templated class's members emit no field entity**
at all; enums are refused (§4).

**Corpus caveat, and it is severe.** Three repos hold any C/C++: **spdlog (149
files), staxrip (17), dart-samples (6)** — effectively ONE repo, and spdlog is a
header-only template library. This is the narrowest corpus any arm on this issue
has had. No claim is made that #6726's number moves.

### 2. Why the real corpus number is 1 — measured, not reasoned

`classifier.go:389` maps **`.h` → language `c`**, so 109 of spdlog's 149 files and
14 of staxrip's 17 are parsed with the **C grammar**, where `class`, `template`
and `namespace` do not parse. The effect is not marginal:

| | field entities | edges |
|---|---|---|
| `.h` → `c` (shipped) | 51 | **1** |
| `.h` → `cpp` (hypothetical) | 254 | **6** |

staxrip alone goes from 19 field entities / 0 edges to 137 / 5. **Header-only C++
is largely invisible to this lane**, and this arm does not change that — it is a
classifier question (adjacent to #7046), stated here because otherwise the `1`
reads as a property of this pass rather than of its input. I have not filed it as
its own issue; flagging for the owner.

### 3. THE NOT-A-DECLARATION AUDIT — the #7047 / #7056 hazard is LIVE in cpp

`extractClassLike` (`extractor.go:825`) emits a `SCOPE.Component` for a
`class_specifier` **with or without a body**. So `class Order;` mints a record
with the **right name, right file, admissible Kind and admissible Subtype** that
declares nothing — arm G's `extension` carrier exactly. Every same-file record
able to carry a type's bare name was enumerated from this package's emit sites
and **probed**, not assumed:

| carrier | minted? | refused by |
|---|---|---|
| **forward declaration** `class Order;` | **yes — Component/class, bare name** | `Metadata["definition"]` (new) |
| **elaborated type specifier** `struct Item i;` | **yes — a second bodyless Component** | same marker |
| `#include <vector>` | **yes — Component/import named `vector`** | subtype allow-list |
| `using namespace std;` | **yes — Component/import named `std`** | subtype allow-list |
| `using std::string;` | yes — named `std::string` | allow-list; also unreachable (`::`) |
| namespace, incl. **re-opened** | yes — Component/namespace | subtype allow-list |
| **explicit specialisation** `template<> struct Box<int>` | yes — Schema/template named `Box<int>` | Kind; also unreachable (`<`) |
| `friend class Buddy;` | **no record** | — (pinned so it stays that way) |
| `using Alias = Order;` / `typedef` | **no record** | — (pinned as a capture ceiling) |
| file carrier | yes | subtype allow-list |

**With and without the refusal, measured by toggling only it:**

- corpus as classified: 1 → 1 (**+0**) — nothing to catch, because §2.
- `.h`-as-cpp hypothetical: **6 → 11 (+5, i.e. 45% of the unrefused total would be
  wrong)**. Named: `Device`×2, `AVSMap`×1, `IFunction`×1, `sockaddr_in`×1.

**This figure was +16 before review and is now +5, and the correction matters:**
11 of those 16 hung off a function-shaped SOURCE (§3b) and are refused on that
ground now, so counting them here was double-counting one edge's two defects.
The refusal is still load-bearing — 5 of 11 — and `AVSMap` is still live.

**The `AVSMap` row is real code, not a fixture.**
`staxrip/Source/FrameServer/avisynth.h:193` is literally `class AVSMap;`, and 11
fields in that same header are typed `AVSMap*`. Its real definition is not in the
repo at all. Without the marker those 11 edges bind to a carrier that declares
nothing, and **every instrument reports bound / not-dangling / right-name /
right-file** — which is precisely why #7056 exists.

**I did not treat the 100% bind rate as evidence.** What I diffed is the TARGET
SET row for row with the refusal toggled, which is the only instrument that can
see this class.

### 4. The resolver tier and the kinds it weighs — DRIVEN, not read

Every target this pass admits is a `SCOPE.Component`, so every edge is answered at
**tier 1**, `lookupLocationKind(file, name, componentKindFamily)`
(`refs.go:2982`), and never reaches `ambigLocation` (tier 2) or `byLocation`
(tier 3). Established by driving `resolve.BuildIndex` → `resolve.ReferencesEmbedded`
over records from the real extractor (`..._ResolvesToEntityIDs`,
`..._ConstructorDoesNotSuppressItsClass`), not by reading `refs.go`.

So the ambiguity count is scoped to **componentKindFamily closed under the index's
trim alias** (8 entries; arm F shipped this table with `SCOPE.Class` missing — that
hole is not re-introduced), and the **unit is the KIND, not the RECORD**.

**Counterfactuals, measured:**

| rule | corpus (as classified) | corpus (`.h`-as-cpp) | dropped rows |
|---|---|---|---|
| **arm F / this arm** (kinds ∈ family) | 1 | 6 | — |
| **arm D** (all same-file kinds) | 1 | **5 (−1)** | `PClip.p -> IClip`, carried by `SCOPE.Component+SCOPE.Operation` |
| **arm C** (records) | **0 (−1)** | **4 (−2)** | `Place.coordinate -> Coordinate`, `PClip.p -> IClip` |

(Re-measured against the 6-edge base after the §3b source refusal; the earlier
table was computed against 23, so its rows included edges that no longer exist.
The remaining rows are the same defect, on the surviving denominator.)

**Arm D's rule is wrong for C++ for a reason that is a property of the language,
not of this corpus: a constructor is named after its class.** An inline
`IClip() { }` makes `extractFunction` emit `SCOPE.Operation` named `IClip` in the
same file as the Component — so arm D's rule refuses a target for **every class
that defines a constructor inline**, which in header-heavy C++ is most of them.
`SCOPE.Operation` is not in `componentKindFamily`, so tier 1 binds it without
hesitation. Graded through the REAL resolver by
`TestCppFieldTypeRefs_ConstructorDoesNotSuppressItsClass`, whose premise (the
same-named Operation exists) is asserted first so it cannot pass vacuously.

Arm C's record rule additionally deletes the **1** edge the shipped pipeline
actually produces — `dart-samples` declares `Coordinate` twice in one header —
which is #7038 reproduced from a second language.

**I do not repeat arms C/D/E's justification** that a refusal "would guarantee a
dangling stub". It is false in general, and for cpp it is demonstrably false in
the enum case below.

### 5. Enums refused — and the PHP reason does NOT transfer

A cpp enum is `SCOPE.Schema/enum`, and `Color c;` is ordinary C++. Arm H refused
PHP enums because a component-space ref to one hits an **ambiguous** tier 2 and
dangles. **That reason does not hold here**: this package never calls
`extractor.EnumEntity`, so there is no `SCOPE.Enum` value-set beside the enum, and
a component-space ref would fall through to the **kind-agnostic byLocation tier
and BIND**. Refused anyway, for two reasons I can defend:

1. an address whose scope segment says `component` should not be answered by a
   non-component — admitting them would put this arm's targets on **two tiers with
   two ambiguity rules** (arm G's shape, which arm G took only because swift left
   it no choice);
2. the honest alternative is a **schema-space dialect** (`schemaKindFamily`
   resolves a cpp enum at tier 1) — two dialects behind one `ref_kind`, #6451's
   shape, which should be decided across arms rather than invented here.

**Measured recall loss: 0 edges on the corpus as classified; 2 on the
`.h`-as-cpp hypothetical** (`async_msg_type`, `pad_side`). Named as a loss, not
papered over.

### 6. Fixture axes — VARIED vs HELD CONSTANT

**The cpp trap, stated because it is what would make this fixture look thorough
while grading one axis.** In the tree-sitter C++ grammar the pointer / reference /
array decoration lives in the **declarator, not the type**, so `Order o;`,
`Order* p;`, `Order& r;` and `Order arr[4];` all arrive at the pass as the
**identical string `"Order"`**. Those rows are labelled in the fixture as what
they actually grade — that each decorated member gets its own entity and its own
edge — and the type-expression axis is swept separately.

**VARIED**
- *type expression* (declarator held constant): bare, `std::vector<Order>`,
  `std::map<Item, Order>`, `std::unique_ptr<Order>`, `std::pair<Order, Item>`
  (two targets), `Box<Order>`, `struct Item` (C elaborated form), `unsigned int`,
  `const int`, `::Order`, `Ns::Order` (own test), `struct { int inner; }`.
- *declarator* (type held constant): value, `*`, `&`, `[4]`, multi-declarator.
- *target kind*: definition class / struct / union; **forward declaration**;
  `#include` placeholder; `using namespace` placeholder; `using X::Y` placeholder;
  Schema/enum; Schema/template; explicit specialisation; alias; typedef; the owner
  itself; a name nothing declares.
- *language*: `cpp` and, in its own test, `c` (the `.h` row, asserting the ref's
  language segment).
- *source-endpoint shape*: plain data member vs member function declaration vs
  pointer-to-member-function member vs bare `void f();` (§3b).
- *template nesting*: outside any template, before the template, inside it, and
  two templates deep (§6).
- *scanner shape space*: **enumerated** (34 rows) rather than sampled —
  `..._CandidateScannerEnumeratesTheShapeSpace`. Arm E's strongest test is this
  one and the reason holds here: through `Extract`, a scanner returning junk and
  one returning nothing are indistinguishable on every row that does not bind.

**HELD CONSTANT**: one declaring class per fixture (so the self-reference rule has
one meaning); one file per fixture — the same-file conjuncts need a second file's
records and are graded by their own unit tests, in opposite directions.

### 7. Same-file conjuncts — THREE, each graded separately

Counted rather than assumed, because php had three where every table said two:

| # | site | failure direction | grader |
|---|---|---|---|
| (a) | rival scan, pass 1 | a cross-file record wrongly **SUPPRESSING** an edge | `..._Unit_TargetsAreScopedToTheRequestedFile` (rival block) |
| (b) | target scan, pass 2 | a cross-file record wrongly **BECOMING** a target | same test (target block) |
| (c) | field scan, `attachCppFieldTypeRefs` | a foreign field record walked at all | same test (field block) |

Each is a separate mutant (M9 / M10 / M11), each DEAD.

### 7b. Corrections made after review (#7057 round 2, commit `9a155cfc9`)

| finding | what was wrong | what it is now |
|---|---|---|
| **F4** | the #7047 audit covered only the TARGET endpoint; 17 of 23 edges hung off a member FUNCTION declaration or a pointer-to-member-function member, whose `field_type` is the RETURN type | such records carry `Metadata["function_shaped"]` and are refused as edge SOURCES; capture defect filed as **#7061**; `cppFieldNames`' doc comment, which asserted something the code does not do, corrected |
| **F1** | §6 claimed #7041's template-parameter over-fire was UNREACHABLE — true of a templated class's DIRECT members, false for a class NESTED in one | fixed, not shipped known-wrong: parameter names in scope are stamped per field and refused by name, scoped to the template's own records |
| **F2** | 6 of 8 `cppComponentAddressFamily` entries were ungraded — arm F's hole's re-entry path | every entry driven individually (8 mutants, all DEAD), negative half enumerated, trim aliases driven through the real resolver |
| **F3** | the self-reference comparison was ungraded on case | `struct money{}; struct Money { money m; };` must still bind — C++ does not fold |

### 8. Mutation table — 25 mutants, 23 DEAD, 2 equivalent with a verified reason

Every mutant compiles (`go vet`), the exact occurrence count of its anchor is
asserted before and after so a silently-unapplied edit cannot read ALIVE, each is
applied at the **call site** rather than to an extracted helper, and the tree is
restored by `cp` from a shasum-verified backup (`git diff HEAD` and
`git status --porcelain` empty afterwards; `-count=1` throughout).

| # | mutant | verdict |
|---|---|---|
| M1 | `ref_kind` → `field_target_types` (near-miss) | DEAD (16) |
| M2 | **definition refusal deleted** (call site) | DEAD (4) |
| M3 | subtype allow-list deleted | DEAD (1 — see below) |
| M4 | allow-list widened to admit `import` | DEAD (1) |
| M5 | Kind test deleted from the target scan | DEAD (1) |
| M6 | ambiguity rule deleted | DEAD (1) |
| M7 | ambiguity rule → ALL kinds (arm D) | DEAD (2) |
| M8 | ambiguity rule → RECORDS (arm C) | DEAD (2) |
| M9 / M10 / M11 | the three same-file conjuncts | DEAD (1 each) |
| M12 | qualified names DESCEND to the last segment | DEAD (2) |
| M13 | brace rule deleted | DEAD (2) |
| M14 | scanner returns only the FIRST candidate (**swift's defect reproduced inside it**) | DEAD (5) |
| M15 | self-reference refusal deleted | DEAD (3) |
| M16 | per-field dedup deleted | DEAD (1) |
| M17 | `FromID` set (edge re-anchored) | DEAD (1) |
| M18 | `target_type` echoes the declared text | DEAD (5) |
| M19 | `Relationships` assigned, not appended | DEAD (3) |
| M20 | ref language hardcoded to `cpp` | DEAD (1) |
| M21 | leading-`::` branch deleted | DEAD (4) |
| M22 | definition marker never stamped | DEAD (16) |
| M23 | definition marker stamped **unconditionally** | DEAD (3) |
| M24 | the pass called twice | DEAD (14) |
| M25 | pass moved BEFORE `attachCppFieldMembership` | **ALIVE — equivalent** |
| M26 | pass moved AFTER `emitExceptionFlowEdges` | **ALIVE — equivalent** |

**M25/M26 are reported as equivalent with a reason, not as a gap, and the reason
is verified rather than asserted**: `attachCppFieldMembership` only appends
`CONTAINS` to existing Components and deletes a metadata key — it mints no record
and changes no Kind / Subtype / Name / SourceFile / `definition`;
`emitExceptionFlowEdges` mints `SCOPE.ExceptionType` on the synthetic
`<exception>` path (`internal/extractor/exception_flow.go:48,125`), neither this
file nor a component-family kind. The header claims placement is **not**
load-bearing, and these two mutants are what that claim is worth. (Arm E's lesson
cuts the other way here: I checked whether the claim holds before making it.)

**Round 2 adds 17 more, all DEAD** (commit `9a155cfc9`): N1 source refusal
deleted · N2 marker never stamped · N3 marker stamped on every field · N4
descendant scan narrowed to the node itself (the pre-existing blind spot) · N5
template-parameter refusal deleted · N6 parameters never stamped · N7 stamp
overwrites instead of appending · N8 stamp range starts at 0 · N9 self-reference
folds case · and **one per `cppComponentAddressFamily` entry (8)**, each killed by
`..._Unit_EveryAddressFamilyEntryIsGraded` alone.

**Two of those were ALIVE mid-round and both were closed rather than recorded.**
N8 (stamp range from 0) needed a field declared BEFORE the template — the fixture
had only an AFTER row, so the range bound was ungraded; N7 (overwrite vs append)
needed a nested template. Both fixtures were added and both mutants now die. All
26 first-revision mutants were re-scored against the corrected tree: unchanged.

**M3 dies by exactly one test, and that is the finding it was written for.**
Through `Extract` the subtype allow-list and the definition marker are
**mutually masked** — an `#include` placeholder is refused by BOTH — so deleting
the allow-list changes nothing the fixture suite can see. Its only killer is
`..._Unit_OnlyTypeSubtypesAreTargets`, which supplies the distinguishing input the
fixture suite cannot construct: a refused subtype that **does** carry the marker.
The same reasoning produced M5 (Kind test vs allow-list, distinguishing input
`{Kind: "SCOPE.Model", Subtype: "struct"}`). Two guards that only fire together
grade neither.

### 9. Gates

- **`go test -count=1 ./cmd/grafel/` — GREEN**, in an isolated `GRAFEL_HOME` /
  `GRAFEL_DAEMON_ROOT`. **The digest did not move, and the honest reason is
  "nothing to move", not "the instrument is blind":** `writeSpanFixture6118`
  contains cs / java / ts / js / py / go / rs files and **no `.c`, `.h`, `.cpp` or
  `.hpp` at all**. A new `REFERENCES` edge WOULD move it (`R <from> -[Kind]-> <to>`
  is hashed); `ref_kind` would not (properties are unhashed), and neither would
  the new `Metadata["definition"]`.
- **`scripts/quality/run.sh --ratchet` — exit 0, "38 gated fixtures held", and
  `git status --porcelain` empty afterwards** (the real check, since `--ratchet`
  exits 0 on a ceiling DROP). **It is VACUOUS for this change**: the golden set
  contains **zero** `.c`/`.h`/`.cpp`/`.hpp` files. Reported as a no-op, not
  offered as evidence.
- `go test -count=1 ./internal/extractors/cpp/` green; `gofmt -l` clean. All three
  re-run after the round-2 changes (`struct_fields.go` now stamps a marker on
  field records — `Metadata` is unhashed, so no baseline moves).

### 10. What I refuse, and why

- **Enums as targets** (§5) — a named recall loss, measured, with the alternative
  spelled out rather than invented.
- **Cross-file targets** — the #6976/#6369 hazard; C++ makes it sharper, and it
  needs an include-graph view pass 1 has none of.
- **Aliases and typedefs** — they mint no record; closing that is a capture change
  and is pinned so it fails loudly rather than silently changing these answers.
- I did **not** change `extractClassLike`'s Subtype to distinguish a forward
  declaration, which would move entity subtypes and the #6118 digest. The marker
  is `Metadata`, which is unhashed — arm G's targeted-fix precedent.

### 11. Not established

- **Whether #6726's cpp slice moves. It does not, on this corpus** — and that is a
  statement about our corpus (§2), not about C++.
- Whether a `.h`-as-cpp classification is the right fix, or whether it would
  regress genuine C headers. The hypothetical is a measurement of THIS pass's
  input, not a proposal.
- The `.h`-as-cpp figures come from parsing `.h` with the C++ grammar in a
  harness; the shipped pipeline never does this.
- Whether `internal/custom/**` holds a cpp field-target producer. No custom lane
  exists for cpp and the core emit sites were enumerated completely; the custom
  tree was not swept.

Refs #6912, #6726, #7047, #7056, #7061, #7038, #7041, #7046, #6451, #6976, #6369.


cajasmota (commented):
## Adversarial review — arm I (cpp), #7057

Independent worktree at `d4a9dde41`, own `GOCACHE`, own `GRAFEL_HOME`/`GRAFEL_DAEMON_ROOT`, corpora read-only. I re-derived the headline numbers from scratch rather than reading them, and every mutant below is applied at the call site with an exact occurrence count asserted before and after, reverted by `cp` from a shasum-verified backup (`f6e618e…`, restored and re-verified after each).

**Verdict: REQUEST CHANGES.** Not because the arm is wrong — almost everything it claims I reproduced to the row, and the forward-declaration result is the best not-a-declaration work on this issue so far — but because one shipped source comment says *unreachable* about something I made happen in four lines (F1), six of the eight entries in the table whose arm-F hole it promises not to re-introduce are ungraded (F2), and the exhaustive #7047 audit is done on one endpoint only: on the other endpoint, 8 of the 23 edges hang off something that is not a field (F4).

---

### What survived, independently re-derived

**The corpus numbers are exact.** Walking `archigraph-corpora` myself and picking the grammar the way `classifier.go:389` does:

| | files | field entities | edges |
|---|---|---|---|
| as classified (`.h`→`c`) | 172 | 51 | **1** (`Coordinate`, dart-samples) |
| `.h`-as-cpp hypothetical | 172 | 254 | **23** |

staxrip 19 → 137 field entities and 0 → 22 edges. 109 spdlog `.h` + 14 staxrip `.h` + 3 dart-samples `.h` = 126 `.h` + 3 `.c` on the C grammar. Every figure in §1/§2 matches.

**The forward-declaration number survives, row for row.** Toggling *only* the refusal (`if !cppIsTypeDefinition(r)` → `if false && …`, anchor count 1 before and 1 after):

```
as classified:      1 -> 1   (+0)
.h-as-cpp:         23 -> 39  (+16)
new targets: AVSMap x11, Device x2, IFunction x2, sockaddr_in x1
```

Identical to the PR, including the named rows. And `staxrip/Source/FrameServer/avisynth.h:193` is literally `class AVSMap;` — I read it; `INeoEnv`, `IFunction`, `Device` are forward-declared on the lines above it. The same mutant is DEAD with 4 killers, matching M2. **This is the strongest not-a-declaration result on the issue so far, and the toggle-the-target-set method is the right instrument.**

**The constructor claim survives, and it is a fact about what the extractor emits, not an argument about C++.** Driving the real extractor on `class IClip { public: IClip() {} };`:

```
SCOPE.Component class     IClip   definition=true
SCOPE.Operation function  IClip            <-- same name, same file
```

so arm D's all-same-file-kinds rule refuses `IClip` as a target for every class with an inline constructor. The counterfactual holds. `SCOPE.Operation` is indeed outside `componentKindFamily`, so tier 1 binds without hesitation.

**Three same-file conjuncts, and there really are three.** I counted them in source (pass-1 rival scan, pass-2 target scan, the field scan in `attachCppFieldTypeRefs`) and mutated each *alone* with `if (false && r.SourceFile != filePath) || …`: each is DEAD on its own, each killed by `..._Unit_TargetsAreScopedToTheRequestedFile`. Isolation is real.

**M3's isolation is real.** `comp()` stamps `definition: true`, so `..._Unit_OnlyTypeSubtypesAreTargets` genuinely supplies "a refused subtype that *does* carry the marker" — the one input the fixture suite cannot build. And it drives `attachCppFieldTypeRefs`, the pass entry point, not an extracted helper. The compound-masking argument is sound.

**M25/M26's equivalence rests on a complete enumeration.** I enumerated `Extract` myself: `FileEntity` → `walkStructural` → `attachOutOfLineContains` → `attachCppFieldMembership` → **this pass** → `emitExceptionFlowEdges` → `TagRelationshipsLanguage` → `TagEntitiesLanguage`. The two moves cross exactly the two passes named. `attachCppFieldMembership` deletes only `base_candidates` (`struct_fields.go:233`) and mints nothing; `emitExceptionFlowEdges` mints `SCOPE.ExceptionType` on the synthetic `<exception>` path. No omission — unlike the java arm.

**The enum reasoning checks out.** `internal/extractors/cpp` never calls `extractor.EnumEntity` (the only occurrence of the string is the comment saying so), and `refs.go:2982+` confirms the order — `lookupLocationKind` → `ambigLocation` → kind-agnostic `byLocation`. A lone cpp enum would fall to tier 3 and **bind**, so arms C/D/E's "would guarantee a dangling stub" wording is false here and was right to drop. Both refusal reasons given instead are defensible.

**Both weak gates are honestly reported.** `writeSpanFixture6118` holds cs/java/ts/js/py/go/rs and **no** `.c`/`.h`/`.cpp`/`.hpp` — "nothing to move" is the true reason, not a blind instrument. The golden set holds **zero** C/C++ files across all 38 fixtures, so `--ratchet` is genuinely vacuous. `EntityRecord.ComputeID` hashes `OrgID+ProjectID+SourceFile+Kind+Name` — `Metadata` is not hashed, so the marker cannot move an entity ID. Reporting two gates as no-ops rather than as evidence is the right call and I confirmed both.

---

### Findings

#### F1 — "in cpp it is unreachable, which is asserted rather than assumed" is FALSE (#7041 is live here)

`field_type_refs.go` §6:

> a TEMPLATED class's members emit no field entity at all … so `template<typename T> struct Box { T v; }` beside a real `struct T` **cannot over-fire** … in cpp it is unreachable, which is asserted rather than assumed

It is reachable. `walkStructural`'s template arm recurses into the templated class's body with `walkStructural` (extractor.go:341-355), so a class **nested inside** a template goes through the class arm and *does* get field entities. Driving the real extractor:

```cpp
struct T { int v; };
template<typename T> struct Outer4 { struct Inner4 { T val; }; };
```
```
SCOPE.Schema field Inner4.val ft="T"
  [REFERENCES -> scope:component:class:cpp:p/probe.cpp:T  ref_kind=field_target_type target_type=T]
```

`Inner4::val` has the *template parameter* `T`, which shadows `::T`. The edge binds to the wrong type — #7041's shape exactly, in the arm whose comment says it cannot happen. The named grader `..._TemplatedClassMembersEmitNoFieldEntity` pins only the template's **direct** members: the fixture grades one axis and the prose claims the whole shape. This is the defect class the board keeps paying for.

Not observed in this corpus (no single-letter targets in the 23/39 sets), so I am not claiming a corpus regression — but "unreachable" is the claim, and it is wrong. Either narrow the sentence to direct members and file the nested case, or refuse a candidate that names a template parameter in scope.

#### F2 — 6 of the 8 `cppComponentAddressFamily` entries are ungraded, which is arm F's hole's re-entry path

The comment says the closure under the trim alias is "load-bearing rather than pedantic", that "these eight entries are exactly that set", and that arm F's missing-`SCOPE.Class` hole "is not re-introduced here". True today — and nothing would notice if it were. Removing one entry at a time, full cpp suite, `-count=1`:

| entry removed | verdict |
|---|---|
| `SCOPE.Class` | **ALIVE** |
| `Class` | **ALIVE** |
| `Component` | **ALIVE** |
| `View` + `SCOPE.View` | **ALIVE** |
| bare `Model` only | **ALIVE** |
| `SCOPE.Model` + `Model` | DEAD (`..._Unit_InFamilyRivalSuppressesTheTarget`) |
| `SCOPE.Component` | DEAD (same) |

Only `SCOPE.Model` is pinned — it is the one kind the rival test happens to use — and `SCOPE.Component` is pinned only because it is the target's own kind. The trim-alias claim specifically (bare `Model`) is asserted and unobserved. `..._Unit_OnlyComponentsAreTargets` already loops over kinds; making `..._Unit_InFamilyRivalSuppressesTheTarget` loop over all eight family kinds is a handful of lines and grades the rule instead of one sample.

#### F3 — the self-reference comparison is ungraded on the axis the file argues about

The header argues at length that the target key must not be case-folded because C++ is case-sensitive. The *other* name comparison — `owner != "" && t.name == owner` — is not graded on that axis. `strings.EqualFold(t.name, owner)` is **ALIVE** against the whole suite, and I confirmed it is distinguishable on ordinary C++:

```cpp
struct money { int a; };
struct Money { money m; };
```
shipped: `Money.m -> money`. Mutant: no edge, silently. Distinguishable and ungraded.

#### F4 — the #7047 audit is exhaustive on the TARGET endpoint and entirely absent on the SOURCE endpoint, and the source is not a field 35% of the time

I dumped the source record of every one of the 23 edges. 21 are in `avisynth.h`; I read all 21 declaration lines. They split three ways:

| source line shape | count | is it a field? |
|---|---|---|
| `IClip* p;`, `PClip child;`, `VideoInfo vi;`, … | **5** | yes — the intended case |
| `VideoFrameBuffer* (VideoFrame::*GetFrameBuffer)() const;` — pointer-to-member-function **data** members | **8** | a data member, but its declared type is a function-pointer type; the edge names the pointed-to function's **return** type |
| `virtual const VideoInfo& __stdcall GetVideoInfo() = 0;`, `IClip* GetPointerWithAddRef() const;`, … — member **function declarations** | **8** | **no** |

So 8 of 23 edges (and 8 of the 21 in the only real C++ file the arm has) hang off a `SCOPE.Schema/field` that is a *method*, and the edge asserts

```
field_name=GetVideoInfo  ref_kind=field_target_type  target_type=VideoInfo
```

about `virtual const VideoInfo& __stdcall GetVideoInfo() = 0;`. There is no such field.

The capture defect is pre-existing — `emitClassFieldMembers` (#4854) mints field entities for member function declarations — and this PR did not cause it. But this arm is the first pass to **propagate** it into an edge whose `ref_kind` makes a claim about a field, and the headline "23 edges / 23 bound / 0 dangling, every endpoint NAMED" conceals it: this is #7056's blindness mirrored onto the other endpoint. Right file, right name, admissible kind, binds cleanly — and the source is not a field. Given how carefully §3 audits the target side, the absence of the symmetric audit is the gap that matters most in this PR. It needs a count in §11 at minimum; better, refuse a field record whose captured shape is a function declarator (`Signature` here strips the parameter list, so the AST is the place to test).

---

### My mutants (17 + 2 controls, none previously run)

All at the call site, permissive-biased, `go vet` clean, exact anchor counts asserted, `-count=1`, tree restored to `f6e618e…` and `git status --porcelain` empty after every row.

| # | mutant | verdict |
|---|---|---|
| R1 | family: drop `SCOPE.Class` | **ALIVE — distinguishable and ungraded (F2)** |
| R2 | family: drop `Class` | **ALIVE — ungraded (F2)** |
| R3 | family: drop `Model` + `SCOPE.Model` | DEAD (1) |
| R4 | family: drop `View` + `SCOPE.View` | **ALIVE — ungraded (F2)** |
| R5 | family: drop `Component` | **ALIVE — ungraded (F2)** |
| R6 | family: drop `SCOPE.Component` | DEAD (1) |
| R18 | family: drop bare `Model` only (the trim-alias claim) | **ALIVE — ungraded (F2)** |
| R7 | rival scan additionally requires a type subtype (permissive) | DEAD (1) |
| R8 | brace refusal → leading brace only (permissive) | DEAD (2) |
| R9 | allow-list drops `union` | DEAD (3) |
| R12 | ambiguity threshold `> 1` → `> 2` (permissive) | DEAD (1) |
| R13 | dedup keyed by target name, not ToID | DEAD (1) — and equivalent in production; the key is derived from the name |
| R14 | self-reference check case-folded | **ALIVE — distinguishable, input named (F3)** |
| R15/R16/R17 | the three same-file conjuncts, one at a time | DEAD (1 each) |
| M2 (re-run) | definition refusal deleted at the call site | DEAD (4) — matches |

`..._Unit_ADefinitionMarkerIsRequired` already sweeps `nil` / `{}` / missing key / `false`, so the two marker mutants I planned (nil→true, key-presence-not-value) are graded; I dropped them rather than pad the table.

---

### The `.h` finding vs #7046 — sibling, not the same defect

Common upstream cause (`classifier.go:389`), two disjoint consequences:

- **#7046** is the **custom** lane: `"c"` is a language no `customPrefixForLanguage` entry routes, so the 12 `|| file.Language == "c"` disjuncts in `internal/custom/cpp` are dead.
- **This** is the **core** lane: the cpp extractor *is* registered for `"c"` (`extractor.go:63`) and does run — but `Extract` selects `cGrammar()` for it (`extractor.go:88-92`), so `class` / `template` / `namespace` never parse. That is the 51-vs-254 field entities.

Routing `"c"` to a custom prefix — #7046's proposed fix — recovers **none** of the 203 missing field entities, because they are lost at grammar selection, upstream of custom dispatch. Only reclassifying `.h` would move both. So: independent fixes, one shared cause. I agree it should not be filed inside this PR — but it should be filed, not just flagged in a PR body; it is a larger number than the arm itself and it will otherwise be rediscovered by arm J.

On honesty of labelling: I checked every occurrence of `23`/`39` in the PR body and source comments, and the hypothetical is labelled at each one. No conclusion leans on it as measured. The 41% in §3 is a share of the hypothetical total and is stated as such.

### Gates

Run in my own worktree with a private `GOCACHE` and a **fresh** `GRAFEL_HOME` + `GRAFEL_DAEMON_ROOT`:

```
go vet ./internal/extractors/cpp/ ./cmd/grafel/     exit 0
gofmt -l internal/extractors/cpp cmd/grafel         (no output)
go test -count=1 ./internal/extractors/cpp/         ok   0.620s
go test -count=1 ./cmd/grafel/                      ok   188.072s
git status --porcelain                              clean
```

`--ratchet` is vacuous as the author says — the golden set's 38 fixtures contain zero `.c`/`.h`/`.cpp`/`.hpp`, so `git status --porcelain` being empty is the only thing it establishes. Same for the #6118 digest: `writeSpanFixture6118` holds no C/C++ file, so `./cmd/grafel/` passing is a compile-and-no-collateral result here, not evidence about this pass.

### Not verified

- The arm-C (19) and arm-D (21) counterfactual *magnitudes* — I verified the mechanism behind arm D's (the same-file `SCOPE.Operation` named after the class) and the `Coordinate` double-declaration behind arm C's `-1`, but did not re-derive the totals.
- `internal/custom/**` was not swept for a cpp field-target producer; the author says the same.
- The Windows test leg was still pending when I looked.

cajasmota:
## Round 2 — `9a155cfc9`. All four findings addressed; the body is updated rather than appended to, so its numbers are the corrected ones.

**F4 — the important one, and the review is right that it is #7056 mirrored.**
I audited what an edge may point AT and never what it may hang OFF.

Reproduced and measured: **68 of 254** C++-parsed corpus field entities (26.8%)
are minted from a construct that is not a plain data member — a member function
declaration wrapped in pointer/reference decoration, or a
pointer-to-member-function member — and in both the record's `field_type` is the
**return** type. **17 of the arm's 23 edges** hung off one.

- Such records now carry `Metadata["function_shaped"]` and are **refused as edge
  SOURCES**. Corpus: **23 → 6** on the `.h`-as-cpp hypothetical; **1 → 1** as the
  pipeline actually classifies (that one edge was never function-shaped).
- **I chose BOTH options you offered, and the reason is that they cover different
  things.** The false comment is corrected in place, because it is what misleads
  the next reader of `cppFieldNames`; and the capture bug is filed as **#7061**,
  because correcting the comment does not stop the bogus ENTITY being minted, and
  68 of 254 is a number that should not live only in a PR body. Neither the entity
  nor its `CONTAINS` edge is touched here — that changes entity counts, the #6118
  digest and the #4854 contract.
- I did **not** re-derive your 8/8 per-shape split; my harness measured the
  aggregate (17). The source comment says so and attributes the split to the
  review rather than restating it as mine.

**A consequence worth flagging, because it cuts against my own headline.** The
forward-declaration refusal's figure drops from **+16 to +5** (6 → 11, 45% of the
unrefused total). 11 of those 16 hung off a function-shaped source, so counting
them there was double-counting one edge's two defects. The refusal is still
load-bearing and `AVSMap` is still live — but the inflated number is corrected in
the body, not defended.

**F1 — fixed, not shipped known-wrong.** You are right that it is four lines away:
`walkStructural` recurses into a template's body, so a class NESTED in one gets
field entities and `T val;` would have bound to a same-file `struct T`. The
parameter names in scope are now stamped per field and refused by name, scoped to
the template's own records. Graded in four positions — outside any template,
**before** it, inside it, and two templates deep. The "unreachable" claim is gone;
what remains is the narrower true statement (a templated class's DIRECT members
emit no field entity), and it is labelled as covering only that.

**F2 — every entry now individually graded, one mutant each, all DEAD.** One
structural note: `SCOPE.Component` cannot be graded by a rival of its own kind,
because the target IS one and a same-kind rival is one node — which is the
kind-not-record rule. Its entry is instead what makes the target's own kind
countable, so all seven other rows fail without it; the mutant confirms that. The
negative half is enumerated too, so the table cannot quietly become "every kind",
and the trim aliases (`Class` / `SCOPE.Class` — arm F's actual hole) are driven
through the real resolver rather than argued.

**F3 — graded.** `struct money{}; struct Money { money m; };` must still produce
an edge; the `EqualFold` mutant is DEAD.

**Grading.** 17 new mutants (9 + 8 family entries), all DEAD; the 26 from the
first revision re-scored against the corrected tree, unchanged (24 DEAD, M25/M26
equivalent). **Two were ALIVE mid-round and both were closed rather than
recorded:** N8 (stamp range starting at 0) needed a field declared BEFORE the
template — my fixture had only an AFTER row, so the range bound was ungraded — and
N7 (overwrite instead of append) needed a nested template. Both fixtures were
added. That is the same varied/held-constant defect this board keeps finding,
caught in my own round-2 work rather than in review.

`./cmd/grafel/` green in an isolated env, `--ratchet` exit 0 with
`git status --porcelain` empty, `gofmt` clean — all re-run after the changes,
since `struct_fields.go` now stamps a marker (`Metadata` is unhashed, so no
baseline moves).

Not merging. Refs #7061.

cajasmota (commented):
## Adversarial review — round 2, `9a155cfc9`

Same isolated worktree, private `GOCACHE`, fresh `GRAFEL_HOME`/`GRAFEL_DAEMON_ROOT`, corpora read-only. Every number below I re-derived; every mutant is applied at the call site with an **exact string-occurrence** count (not a line count) asserted before and after, and the tree is restored by `cp` from shasum-verified backups after each — all three files end at their exact baseline hashes.

**Verdict: HOLD.** All four round-1 findings are genuinely fixed, the self-correction downward is accurate rather than an over-correction, and the source-endpoint work is the best thing in the PR. What holds it is one new defect of exactly the class this round just fixed, one level down: **`cppTemplateParams` — the function the new refusal now depends on — is wrong in *both* directions across five parameter forms**, and the recursion the marker depends on is ungraded because the mutant written for it was compound with a lethal first half.

---

### Everything claimed, re-derived

| claim | my measurement |
|---|---|
| 68 of 254 function-shaped (26.8%) | **68 / 254 = 26.77%** ✔ |
| 23 → 6 on the hypothetical | **23 → 6** ✔ (toggling only `cppIsFunctionShaped`: 6 → 23) |
| 1 → 1 as classified | **1 → 1** ✔ (5 function-shaped refused there) |
| forward declarations **6 → 11 (+5, 45%)** | **6 → 11**, +5, 45.45% ✔ |
| named rows `Device`×2, `AVSMap`×1, `IFunction`×1, `sockaddr_in`×1 | **exactly those five** ✔ |

**The corrected forward-declaration figure survives, and it is right for the reason given.** `AVSMap` is still live (11 → 1, because 10 of the 11 `AVSMap*` fields were themselves function-shaped) and the refusal is still load-bearing at 5 of 11. Dropping +16 to +5 is not an over-correction: I toggled each refusal independently and the two populations do overlap by exactly 11, so counting them in both places really was double-counting one edge's two defects.

**The 23 → 6 drop lands on precisely the right six.** The survivors are `Coordinate`, `IClip`, `PClip`, `VideoFrame`, `VideoFrameBuffer`, `VideoInfo` — which is exactly my round-1 per-shape split (5 genuine data members in `avisynth.h`, plus dart-samples' `Coordinate`) arrived at from the opposite direction. Two independent methods, and they agree.

**The attribution is honest.** §3b says the per-shape split is "NOT re-derived here", attributes 8 + 8 to the review, and states it "accounts for 16 of the 17" rather than rounding it up to 17. Nothing downstream leans on it — every figure in §1, §2, §3 and the counterfactual table is the aggregate the pass measured itself. The counterfactual table was correctly re-measured against the new 6-edge base instead of being left stale.

**F2 — fixed and now genuinely graded.** I re-ran one mutant per entry: all eight are **DEAD**, each killed by `..._Unit_EveryAddressFamilyEntryIsGraded`. The `SCOPE.Component` argument is **not** a rationalisation — dropping it means `nameKinds` never records the target's own kind, so `len(nameKinds[name])` can no longer exceed 1 and the rival rows collapse; it dies with two killers. `Class` / `SCOPE.Class` — arm F's actual hole — die on their own. All six entries that were ALIVE in round 1 are dead.

**F3 — fixed.** `EqualFold` is DEAD via `..._SelfReferenceComparisonIsCaseSensitive`.

**F4 — fixed and the marker is clean in both directions.** I swept the shape space myself: `void greet();` mints no record at all; `Order* getP()`, `Order& getR()`, `const Order& getCR() const`, `virtual Order* getV() = 0`, a pointer-to-member-function member and a plain function-pointer member are all marked; and **no** plain data member is marked — value, `*`, `&`, `[4]`, multi-declarator, and initialised members all stay unmarked and keep their edges. No false positives found.

**N7/N8 closures are real.** N8: `stampCppTemplateParams(out, 0, …)` (made to compile with `_ = before`) is **DEAD** via `..._TemplateParameterIsNeverATarget`, so the fixture genuinely has a field declared *before* the template. I confirmed the behaviour directly too — a field before the template binds, one after binds, only fields inside are refused. N7: overwrite-instead-of-append is **DEAD** via `..._NestedTemplateParametersAreAllRefused`, and two templates deep really does carry both lists (`tparams=[U T]`).

**The `cppFieldNames` doc correction is accurate and does not overclaim.** `cppFirstChildOfType(ch, "function_declarator")` is direct-child only, so "covers `void greet();`, does not cover a function declarator wrapped in pointer/reference decoration" is exactly right — I verified both halves.

---

### Findings

#### H1 (holds the PR) — `cppTemplateParams` is wrong in BOTH directions, so F1 is fixed for four parameter forms and broken for two

The new refusal reads `Metadata["template_params"]`, which comes from `cppTemplateParams` (extractor.go:1289). That function collects **every** `type_identifier`/`identifier` child of a parameter declaration and has no case for two parameter kinds. I swept the `template_parameter_list` shape space against a same-file `struct Order`:

| form | `tparams` | `In.o` (declared `Order`) | |
|---|---|---|---|
| `template<typename Order>` | `[Order]` | refused | correct |
| `template<class Order>` | `[Order]` | refused | correct |
| `template<typename... Order>` | `[Order]` | refused | correct |
| `template<int Order>` | `[Order]` | refused | correct |
| **`template<template<typename> class Order>`** | **`[]`** | **binds → `struct Order`** | **WRONG BINDING** |
| **`template<int Order = 4>`** | **`[]`** | **binds → `struct Order`** | **WRONG BINDING** |
| **`template<typename T = Order>`** | **`[T, Order]`** | **refused** | **over-refusal** |
| **`template<typename T = Order, typename U = Item>`** | **`[T, Order, U, Item]`** | **both refused** | **over-refusal ×2** |
| **`template<Cc Order>`** (C++20) | **`[Cc, Order]`** | refused, but `Cc` is harvested | over-refusal if a type is named `Cc` |

Two consequences:

1. **#7041's shape is still live** for template-template parameters and for defaulted non-type parameters. That is the wrong-binding direction — the one F1 was filed for and the one `bug-extractor` cannot see, because it binds.
2. **Correct edges are silently deleted** whenever a default type argument is a bare identifier naming a same-file type. `template<typename Alloc = MyAllocator>` is ordinary C++, and this arm's whole motivating corpus is header-only template code.

Root cause is one rule: only the **last** identifier before any `=` is the parameter name, and `template_template_parameter_declaration` / the optional non-type form are missing from the case list. `cppTemplateParams` predates this PR, but this PR is the first to make a **refusal** decision from it — at extractor.go:1271 it only decorates a template entity — so the PR owns the consequence.

And the same pattern as the three round-1 findings: the source comment says "both directions graded by `..._TemplateParameterIsNeverATarget`". That fixture grades the **positional** axis (outside / before / inside / two deep) thoroughly. It holds the **parameter-list contents** axis constant at a single bare `typename T`.

#### H2 (holds the PR) — the recursion the marker depends on is ungraded, and N4 is a compound whose first half is lethal alone

N4 narrowed `cppHasDescendantOfType` "to the node itself", i.e. deleted the whole child loop. That is lethal on its own — the decorator node is never a `function_declarator` — so it kills without ever exercising the recursion. Keeping the loop and removing only the recursive call:

```go
for i := 0; i < int(n.ChildCount()); i++ {
    if c := n.Child(i); c != nil && c.Type() == typ { return true }
}
```

is **ALIVE** against the whole suite, and **ALIVE** against the corpus (`functionShaped=68`, `edges=6`, unchanged). It is nonetheless production-distinguishable, and not marginally:

| source | shipped | depth-1 mutant |
|---|---|---|
| `Order** getPP();` | marked | **unmarked → emits `getPP -> Order`** |
| `Order*** getPPP();` | marked | **unmarked → wrong edge** |
| `Order*& getPR();` | marked | **unmarked → wrong edge** |
| `Order** (*fpp)(int);` | marked | **unmarked → wrong edge** |

Every existing fixture and every one of the 68 corpus records has its `function_declarator` at depth 1. One row — `Order** getPP();` — closes it. This is exactly the trap the board keeps naming: a compound mutant grades its second half only if its first half is not lethal alone, and N4's is.

#### H3 (minor, not holding) — the `function_shaped` marker is not graded where its sibling is

Reading it as key presence rather than value (`_, ok := r.Metadata["function_shaped"]`) is **ALIVE**. Equivalent in production today — the marker is only ever stamped `true` — but the sibling `definition` marker *is* graded for a `false` value by `..._Unit_ADefinitionMarkerIsRequired`, which sweeps `nil` / `{}` / missing / `false`. One row in the function-shaped unit test closes the asymmetry.

#### H4 (cosmetic) — the shape-scan prose undercounts, and two case entries are unreachable

`cppFunctionShapedNames` says "TWO distinct source shapes land here". There are three: a plain function-pointer data member (`Order* (*fp)(int);`) is also marked, correctly, and for the same reason. Separately, dropping `init_declarator` or `array_declarator` from the case list is ALIVE with no corpus or fixture change, and I could not construct a distinguishing input for either — I report them as defensive entries I could not show to matter, not as a defect.

---

### My mutants this round (24 scored, all new)

Exact occurrence counts asserted; `go vet` clean on every scored row; `-count=1`; restored to `4fbfa4f9…` / `67ebbad8…` / `2e8881f6…` after each.

| # | mutant | verdict |
|---|---|---|
| S1 | source refusal deleted at the call site | DEAD (1) |
| T7 | `function_shaped` read as key presence, not value | **ALIVE — equivalent in production, ungraded (H3)** |
| S3 | nil Metadata ⇒ function-shaped | DEAD (8) |
| S9 | marker stamped unconditionally | DEAD (20) |
| T20 | drop `reference_declarator` from the shape scan | DEAD (1) — corpus 68→57, edges 6→13 |
| T18 | drop `pointer_declarator` from the shape scan | DEAD (1) — corpus 68→11, edges 6→16 |
| T15 | drop `init_declarator` from the shape scan | ALIVE — no distinguishing input found (H4) |
| T16 | drop `array_declarator` from the shape scan | ALIVE — no distinguishing input found (H4) |
| **T5** | **descendant scan → depth 1 (recursion only)** | **ALIVE — distinguishable, input named (H2)** |
| S10 | template-parameter refusal deleted | DEAD (2) |
| T6 | stamp range from 0 (N8, made to compile) | DEAD (1) |
| S12 | stamp overwrites instead of appending (N7) | DEAD (1) |
| S13 | stamp only the first field in range | DEAD (1) |
| S14 | parameter refusal case-folded | DEAD (2) |
| S15 | definition refusal deleted (isolated from the new source refusal) | DEAD (4) |
| S16–S23 | one per `cppComponentAddressFamily` entry (8) | **all DEAD** |
| S24 | self-reference check case-folded | DEAD (1) |

Plus three toggle experiments on the corpus (definition refusal, source refusal, template-parameter refusal) reported above. The template-parameter refusal costs and gains **nothing** on the corpus — 11 fields carry `template_params` and none of them would have over-fired — so that guard is fixture-graded only, which is worth stating since H1 says the fixture's axis is too narrow.

### Gates

Own worktree, private `GOCACHE`, fresh `GRAFEL_HOME` + `GRAFEL_DAEMON_ROOT`:

```
gofmt -l internal/extractors/cpp cmd/grafel     (no output)
go vet ./internal/extractors/cpp/ ./cmd/grafel/ exit 0
go test -count=1 ./internal/extractors/cpp/     ok   0.643s
go test -count=1 ./cmd/grafel/                  ok   185.282s
git status --porcelain                          clean, before and after
```

Re-confirmed both weak gates are still vacuous: the golden set's 38 fixtures contain **zero** `.c`/`.h`/`.cpp`/`.hpp`, and `writeSpanFixture6118` contains none either, so neither `--ratchet` nor the #6118 digest can observe this change. Unchanged from round 1. CI on this head is green on 15 checks with the Windows test leg still pending.

### What clears the hold

1. Fix `cppTemplateParams` to take only the parameter **name** (the last identifier before any `=`) and add the two missing parameter kinds; add fixture rows for the five forms in H1's table — the two wrong-binding rows are the ones that matter.
2. One fixture row with a depth-≥2 function declarator (`Order** getPP();`) so the recursion in `cppHasDescendantOfType` is graded, and re-score N4 as two separate mutants.

H3 and H4 are worth a line each but do not hold it.

### Not verified

- Whether a template-template or defaulted-non-type parameter shadows a same-file type anywhere in the corpus — the corpus has no such collision either way, so H1's over- and under-refusals are both zero-incidence *on this corpus*, which per the board's own rule is a corpus-relative zero and not a bound on the language.
- The arm-C/arm-D counterfactual magnitudes on the new base (I verified the mechanisms, not the totals).
- `internal/custom/**` was not swept.
- I did not execute `scripts/quality/run.sh --ratchet` itself this round. Its vacuity is established structurally — the golden set contains no C/C++ file, so the gate cannot observe this change in either direction — and running a 38-fixture index concurrently with the `./cmd/grafel/` leg on this shared 16GB machine is the thing that thrashes swap. Stated as not-run rather than reported as green.

cajasmota:
## Round 3 — `8274ef6cc`. H1–H4 fixed. Both headline numbers re-measured and unchanged.

### H1 — the parameter forms, enumerated, and what each now collects

`cppTemplateParams` is replaced by one uniform rule: **the name is the LAST name
node in a pre-order walk of the parameter declaration, stopping at the first `=`
at any depth, and skipping a nested `template_parameter_list` whole.** Taking the
last is what survives a parameter whose type or constraint is itself a named
entity; stopping at `=` is what keeps a default argument from reading as a
binding; skipping the inner list is what keeps a template-template parameter's own
parameters out of scope.

| form | old | new |
|---|---|---|
| `typename A` | `[A]` | `[A]` |
| `class B` | `[B]` | `[B]` |
| `typename C = Order` | **`[C, Order]`** — deletes a correct edge | `[C]` |
| `typename A = Order, typename B = Item` | **`[A, Order, B, Item]`** — deletes two | `[A, B]` |
| `int D` | `[D]` | `[D]` |
| `int E = 4` | **`[]`** — #7041 live | `[E]` |
| `Item* Order` | **`[Item, Order]`** — deletes edges to `Item` | `[Order]` |
| `Cc Order` (constrained) | **`[Cc, Order]`** — harvests the concept | `[Order]` |
| `template<typename> class F` | **`[]`** — #7041 live | `[F]` |
| `template<typename> class G = Tmpl` | **`[]`** | `[G]` |
| `typename... H` | `[H]` | `[H]` |
| `int... I` | **`[]`** — #7041 live | `[I]` |
| `typename` (unnamed) | `[]` | `[]` |
| `template<typename Order> class` (unnamed outer, **named inner**) | **`[Order]`** — deletes a correct edge | `[]` |

**The last row is not from the review — it came out of enumerating.** An unnamed
template-template parameter whose inner parameter is named is the only input that
grades the nested-list skip on its own; without it, that skip was a guard with no
distinguishing input. Legal C++, and it fails in the over-refusal direction.

Graded twice over: `TestCppTemplateParams_EnumeratesTheParameterFormSpace` reads
the names off the field metadata #6912 actually consumes (so the whole path is
graded, not the helper in isolation), and
`TestCppFieldTypeRefs_TemplateParameterFormsAreAllRefused` drives the same forms
to the edge level with a real-type control in the same nested class. The member
expression differs per form because C++ requires it — a type parameter bare, a
non-type parameter only inside a template argument, a template-template parameter
only when applied — and getting that wrong is how a row ends up grading nothing.

**None of these forms occurs in our corpus.** Stated as a corpus-relative zero,
and not used as a reason to skip any of them.

### H2 — the recursion, re-scored as its own row

| mutant | edit | verdict |
|---|---|---|
| **N4a** | scan narrowed to the node ITSELF (child loop **and** recursion deleted) | DEAD |
| **N4b** | **only** the recursion removed, child loop kept | **DEAD** (was ALIVE) |

Four depth-≥2 rows added, each one layer deeper than the last and all legal C++:
`Item** getPP()` (depth 2), `Item*** getPPP()` (depth 3), `Item*& getPR()`
(pointer under reference), `Item** (*fpp)(int)` (function pointer returning a
pointer). Each escapes and emits a wrong edge with only the recursion removed.

You were right that this is my own rule turned on me: I applied "a compound
grades its second half only if its first half is not lethal alone" to M3 and to
the three markers and then wrote N4 as a compound whose first half is lethal. The
rule is now in the source comment next to the recursion, not only in a PR body.

### H3 / H4

`function_shaped` is graded on **both** values — the distinguishing input is an
explicit `false`, which is what a producer stamping the key unconditionally would
write; a reader testing key PRESENCE (`_, ok :=`) now dies. That closes the
asymmetry with `definition`, which was graded on both from the start. The
"TWO shapes" sentence says **three** and enumerates them by node kind.

### T15 / T16 — not chased, and one of them has a *reason*

I did try to construct a distinguishing input, and report the result rather than
adding a fixture:

- **`array_declarator`: no distinguishing input, for a language reason, not a
  corpus one.** An array of functions is ill-formed in C++, so a
  `function_declarator` can never nest *under* an `array_declarator`; an array of
  function pointers puts the pointer inside parentheses, which makes the outermost
  declarator a `function_declarator` (skipped by the direct-child guard) or a
  `pointer_declarator`. Probed `Order (*fp[4])(int)`, `Order* (*fp[4])(int)`,
  `Order* (S::* mp[3])(int)`, `Order& (*fr[2])() = {}` — every one presents as
  function/pointer/reference at the top, never array.
- **`init_declarator`: no distinguishing input, but the reason is weaker** — in
  this grammar a member's initializer is a *sibling* of the declarator inside
  `field_declaration` (`Order* p = nullptr;` → `pointer_declarator` + `= nullptr`),
  so nothing wraps. That is an observation about this grammar version, not about
  the language, and I am not claiming more.

Both entries stay, unexercised, because they are in a switch whose default is to
refuse — the safe direction.

### Numbers — re-measured after the changes, all unchanged

- **1 edge** as the pipeline classifies (172 files, 51 fields, 5 function-shaped).
- **6 edges** on the `.h`-as-cpp hypothetical (254 fields, **68 function-shaped =
  26.77%**); forward declarations **6 → 11 (+5, 45.45%)** with exactly `Device`×2,
  `AVSMap`×1, `IFunction`×1, `sockaddr_in`×1; armD −1, armC −2, enums 2.
- New datum: **11 corpus fields now carry a template-parameter stamp** and none of
  their candidates names a same-file type, so the H1 fix changes no corpus edge —
  it closes a language-level gap the corpus cannot see.

### Grading

**10 new mutants, all DEAD** — P1 (the original collector restored verbatim,
killing 2 tests, so the enumeration is not vacuous), P2–P4 (each missing
parameter kind), P5 (the `=` stop), P6 (the nested-list skip), P7 (first name
instead of last), N4a/N4b, H3M. **All 43 from rounds 1–2 re-scored against the
corrected tree: unchanged** (41 DEAD, M25/M26 equivalent).

On your question about the two ALIVE rows with no distinguishing input: in **my**
table those are M25 and M26, the two call-placement moves, and **I still cannot
distinguish either** — `attachCppFieldMembership` appends `CONTAINS` to existing
Components and deletes one metadata key, minting no record and changing no Kind /
Subtype / Name / SourceFile / `definition`; `emitExceptionFlowEdges` mints
`SCOPE.ExceptionType` on the synthetic `<exception>` path, neither this file nor a
component-family kind. Both remain equivalent-with-a-verified-reason, not
ungraded.

`./cmd/grafel/` green in an isolated env, `--ratchet` exit 0 with
`git status --porcelain` empty, `gofmt` clean.

Not merging.

cajasmota (commented):
## Adversarial review — round 3, `8274ef6cc`. **MERGE.**

Same isolated worktree, private `GOCACHE`, fresh `GRAFEL_HOME`/`GRAFEL_DAEMON_ROOT`, corpora read-only. Exact string-occurrence counts on every mutant (not line counts), `cp` restore from shasum-verified backups after each — all three files end at their exact baseline hashes.

H1–H4 are all closed. I attacked the rewritten collector from four directions — the grammar itself, the mutation table, the over-refusal direction, and the corpus — and found nothing that holds the PR.

---

### The under-collecting `default:` — I went at the grammar, not the comment

The right way to judge "is there an unmodelled node kind that binds a name" is to ask the grammar, so I swept **26 template forms** through the real parser and enumerated every direct child of every `template_parameter_list`:

```
type_parameter_declaration               modelled
optional_type_parameter_declaration      modelled
variadic_type_parameter_declaration      modelled
parameter_declaration                    modelled
optional_parameter_declaration           modelled
variadic_parameter_declaration           modelled
template_template_parameter_declaration  modelled
<  >  ,                                  tokens
ERROR                                    parse failure, from `template<Cc Order = int>`
```

**Nothing else exists.** The seven modelled kinds are exactly the grammar's declaration kinds; the `default:` catches only punctuation and parse errors. So the under-collecting choice is not merely defensible in principle — in the current grammar it has nothing to under-collect. The direction is also the right one for *this* refusal: an unmodelled kind that binds no name costs recall on one guard, whereas over-collecting deletes arbitrary correct edges with no symptom. And it is pinned: making `default:` return the declaration's text is **DEAD**.

The one form the parser fails on is worth a line. `template<Cc Order = int>` (constrained parameter carrying a default) produces an `ERROR` node — and the collector still binds `[Order]` correctly, same as the no-default control. So even the parse-failure case does not leak. That ERROR is a pre-existing tree-sitter-cpp limitation, not this PR's.

### All 14 forms are graded, not tabulated

A table in a comment is not a test, so I mutated the rule itself and every entry in its case list:

| mutant | verdict |
|---|---|
| **U1** FIRST name wins instead of LAST | DEAD (2) |
| **U2** the `=` stop removed | DEAD (2) |
| **U3** the nested-list skip removed | DEAD (1) |
| **U4** `default:` over-collects (returns the decl text) | DEAD (1) |
| **U6 × 7** one mutant per modelled declaration kind | **all 7 DEAD** |

Every one of the three rule clauses — last-not-first, stop-at-`=`, skip-the-nested-list — dies on its own, and so does every kind in the case list. That is the same discipline the family table got in round 2, applied to the new surface. Your own check of the skip branch is confirmed: `U3` dies to `TestCppTemplateParams_EnumeratesTheParameterFormSpace`, and the 14th form (`template<typename Order> class`, unnamed outer / named inner) is the only row that can kill it — finding that one by enumerating rather than by working the review list is the good kind of result.

**P1 shows what it is offered as showing.** Splicing the round-2 collector back in verbatim kills exactly two tests — the form-space enumeration and its edge-level half — which is what "the enumeration is not vacuous" means: it catches the real historical defect, not a synthetic one. On its own P1 would not prove each row load-bearing; U1–U6 do, and together the claim stands.

### Over-refusal — pressed hard, no leak

This is the direction nobody had pressed and the one where a defect has no symptom. Every construction I could build behaves correctly:

| construction | result |
|---|---|
| `Sibling` class *between* two shadowing templates | binds `-> Order` ✔ |
| `Sibling2` *after* both | binds ✔ |
| class inside a `namespace` following a template | binds ✔ |
| **two sibling templates with different parameter names** (`T6<Order>`, `T7<Item>`) | each stamps only its own; `I6.i -> Item` and `I7.o -> Order` both bind ✔ |
| function template followed by a class | no stamp leak ✔ |
| parameter that shadows nothing (`T`) | `I5.o -> Order` binds, `I5.t` refused ✔ |
| class nested *deeper* inside a template's inner class | stamped — correct, the parameter is still in scope ✔ |

The sibling-templates row is the sharpest: no cross-contamination between two adjacent parameter scopes.

### Corpus — unchanged, and the new datum is honest

Re-derived independently on this head: 172 files; **51 fields / 1 edge** as classified; **254 fields / 68 function-shaped (26.77%) / 6 edges** hypothetical, same six targets. Toggling each refusal alone: definition **6 → 11** (+5, 45.45%, `AVSMap`×1 `Device`×2 `IFunction`×1 `sockaddr_in`×1), function-shaped **6 → 23**, template-parameter **6 → 6**.

The "11 stamped, none shadowing" claim is **accurate, and it is not "the fix does nothing"**. The 11 are real fields in spdlog's bundled fmt and Qt sinks, with parameters `Context`, `Char`, `N`, `Mutex` — the stamping is live; what is absent is a same-file type of those names, so the refusal fires zero times here. That is a corpus-relative zero on the *collision*, not on the mechanism, and the PR labels it that way.

Worth recording: `field_type_refs.go` is **byte-identical to round 2** (`4fbfa4f9…`), so every refusal I verified last round carries over unchanged rather than needing re-derivation.

### H2, H3, H4 — closed

**H2 confirmed closed.** N4b — the recursion removed while the child loop is kept, the mutant that was ALIVE in round 2 — is now **DEAD**, killed by `..._FunctionShapedMemberIsNeverASource`. The four depth-≥2 fixtures do the work, and the rule now sits in the source comment beside the recursion, which is where the next reader will be. Re-scoring N4 as N4a/N4b instead of leaving the compound is the right correction.

**H3** closed — the presence-test reader now dies. **H4** — the prose enumerates three shapes by node kind.

### M25 / M26 — confirmed, not refuted

Both placement moves are still **ALIVE** and I could not distinguish either. I re-enumerated `Extract` on this head: `FileEntity` → `walkStructural` → `attachOutOfLineContains` → `attachCppFieldMembership` → **this pass** → `emitExceptionFlowEdges` → `TagRelationshipsLanguage` → `TagEntitiesLanguage`. The two moves cross exactly the two passes the reason names, no pass was added this round, and neither mints a same-file component-family record. Equivalent with a reason that still holds.

### My mutants this round (17 scored, all new)

| # | mutant | verdict |
|---|---|---|
| U1 | FIRST name wins instead of LAST | DEAD (2) |
| U2 | `=` stop removed | DEAD (2) |
| U3 | nested-list skip removed | DEAD (1) |
| U4 | `default:` over-collects | DEAD (1) |
| U5 | name switch also accepts `field_identifier` | ALIVE — **equivalent**, a `field_identifier` cannot occur in a `template_parameter_list` (grammar sweep); no distinguishing input |
| U6.1–U6.7 | one per modelled declaration kind | **all DEAD** |
| U7 | N4b: recursion removed, child loop kept | **DEAD** (was ALIVE in round 2) |
| U8 | M25 placement move | ALIVE — equivalent, reason re-verified |
| U9 | M26 placement move | ALIVE — equivalent, reason re-verified |
| P1 | round-2 collector restored verbatim | DEAD (2) — as claimed |

Plus three corpus toggle experiments and a 26-form grammar sweep.

### Gates — and one red I chased down rather than waved away

```
git diff HEAD                                   empty
gofmt -l internal/extractors/cpp cmd/grafel     (no output)
go vet ./internal/extractors/cpp/ ./cmd/grafel/ exit 0
go test -count=1 ./internal/extractors/cpp/     ok  0.705s
go test -count=1 ./cmd/grafel/  (run 1)         FAIL -> chased, see below
go test -count=1 ./cmd/grafel/  (run 2, quiet)  ok  175.779s
git status --porcelain                          clean, before and after
```

**The first `./cmd/grafel/` run went red**, on `TestRebuild_SubprocessReroute_TimeoutCancelsChild`. It is not this PR, and I did not settle that by re-running until it went green:

1. **Structurally impossible.** That test replaces `runRebuildSubprocess` with a stub that only does `<-ctx.Done()`. No repo is indexed, no file is parsed, **no extractor is invoked at all** — so a change confined to `internal/extractors/cpp` cannot reach the code under test. Same shape as the vacuity argument for the other two gates.
2. **The log shows a scheduling stall, not a logic failure.** The 100 ms per-repo timeout fired exactly as designed (`rebuild wedged STALLED — no result after 100ms; surfacing as timeout`); what missed was the test's **5 s wall-clock** deadline — `rebuild exit group=subproc-timeout-group took=5.88s`. I had a mutation runner in flight and another agent compiling in `internal/daemon/service`.
3. **5/5 PASS** re-run in isolation on this head.
4. **CI `test (ubuntu-latest)` is green on this exact head** (`8274ef6cc`), and that leg runs this package.
5. The same package was green on the round-2 head on this same machine.

**Full-package re-run on a quiet machine: `ok github.com/cajasmota/grafel/cmd/grafel 175.779s`, exit 0.** Six independent lines of evidence, five of them not re-runs.

**Worth its own issue, independent of this PR:** `daemon_rebuild_subprocess_test.go` gates a 100 ms-timeout behaviour behind two 5 s wall-clock `time.After` deadlines. That is a load-sensitive gate which can red a PR it has no relationship to — exactly what happened here — and I found no existing issue for it. Flagging for the owner rather than filing inside this lane.

Both weak gates re-confirmed structurally unchanged rather than re-run: the golden set's 38 fixtures still hold **zero** `.c`/`.h`/`.cpp`/`.hpp`, and `writeSpanFixture6118` still holds none, so neither `--ratchet` nor the #6118 digest can observe this change in either direction.

### Not verified

- I did not run `scripts/quality/run.sh --ratchet`, for the same reason as round 2: it is provably vacuous here, and a 38-fixture index alongside the `./cmd/grafel/` leg is what thrashes swap on this machine. Stated as not-run rather than reported green.
- The arm-C / arm-D counterfactual magnitudes on the 6-edge base (mechanisms verified across rounds, totals not re-derived).
- `internal/custom/**` was not swept — no cpp field-target producer is claimed there.
- What else the `template<Cc Order = int>` ERROR node costs elsewhere in that file's extraction. The parameter itself binds correctly; the wider blast radius of that grammar limitation is out of this PR's scope.
- The Windows test leg was pending at 15/16 green when I looked.

### Verdict

**MERGE.** Three rounds in, every number in the PR body has been re-derived independently and matches; the two claims that were false are gone; the tables that were decorative are now each pinned by their own mutant; and the direction with no symptom — over-refusal — holds under everything I could construct. The remaining ALIVE mutants are equivalences with reasons I checked rather than accepted.

Issue #6325, opened 2026-08-18
the hidden-console wrapper misses a third console source, is written non-atomically, and a non-ASCII username silently kills the daemon


cajasmota:
Follow-up to #6320 (merged). The approach there is right — hidden WScript host for the scheduled task, `CREATE_NO_WINDOW` for the supervised engine child — and the root-causing was accurate. These are the on-disk-artifact items found in review afterwards. All ours to fix.

Note up front, because it shapes the priority: **#6320's macOS test coverage is zero in substance.** The only line that compiles on macOS is `executil.NoWindow(cmd)`, which is a no-op (`internal/executil/nowindow_other.go:8`). The Windows tests do run automatically in `test.yml`'s `windows-latest` leg (unscoped `go test ./...`, line 381) — but `.github/workflows/windows.yml` greps out `/internal/daemon` entirely, so that workflow never exercises any of it.

## 1. Third console source, not covered — the fix is incomplete without this

`internal/dashboard/handlers_updates.go:215-221` — `defaultUpdateRunner` spawns the console-subsystem `grafel.exe` from **inside the daemon process** (reachable from `handleUpdatesApply` / `handleUpdatesRefreshRules`) with no `executil.NoWindow`. Now that the daemon correctly has no console, Windows allocates a fresh one for this child: clicking "update" in the dashboard pops a console window.

Every other daemon-resident spawn in the repo already calls `executil.NoWindow` — `gitmeta.go:83,117,224`, `worktree.go:413`, `subprocess_runner.go:546,715,904`, `extract/coordinator.go:535`, `commit_coupling_edges.go:361,372,412`, `schtasks_windows.go:175`, `install/gitignore.go:79`, `skilllink/linkdir_windows.go:74`, `watchers/loader_windows.go:61`. This is the lone gap. (CLI-invoked spawns like `cli/update.go` are correctly excluded — they run from a real console.)

One line.

## 2. The `.vbs` is written UTF-8 with no BOM; WSH decodes `.vbs` as ANSI

`internal/daemon/service/schtasks_windows.go:248` writes the wrapper as raw UTF-8, no BOM. Windows Script Host decodes `.vbs` using the system ANSI codepage unless the file carries a UTF-16LE BOM.

[...]

PR #7058, opened 2026-09-15
fix(#7051): the /run that starts the Windows daemon threw its exit code away, and the retry meant to cover it landed after we had stopped waiting

### LINKED ISSUE #7051: Windows: daemon scheduled task can fail silently on its first run right after `grafel install`/`start`, with no automatic recovery within the operation's own budget

cajasmota:
Fixes #7051. Reported by @mota-mota, with the `file:line` already in the report.

## The defect

Three constants, three files, nothing comparing them:

```go
schtasks_windows.go:289   _ = schtasksCmd("/run", "/tn", taskName).Run()
schtasks_windows.go:66    <RestartOnFailure><Interval>PT1M</Interval><Count>3</Count>
manager.go                defaultReadiness{budget: 60 * time.Second}
```

The first `/run` after `/create` can fail, write nothing anywhere, and leave the
daemon down indefinitely. `RestartOnFailure` is the declared safety net for
exactly that — and at `PT1M` against a flat 60 s budget it cannot land inside
the window `install`/`start` actually waits on, so the net never catches
anything the reporter was waiting on. A plain retry of the same unmodified task
succeeds immediately.

## Approach, and why

**Both candidate 1 and candidate 2, because they fix different halves.**

Retrying `/run` ourselves is the half that is entirely within our control and
depends on no Task Scheduler behaviour we cannot observe from here — including
the still-UNRESOLVED question of whether `RestartOnFailure` covers an on-demand
`/run` instance at all. If it does not, candidate 2 alone fixes nothing.

Aligning the constants is still worth doing on its own terms, and it is the
half that keeps the bug from coming back.

### 1. `/run` is retried, bounded, and its exit code is no longer discarded

`runAttemptPolicy` (3 attempts, 750 ms apart, 5 s per invocation). Every field
is clamped by `normalized()` rather than by callers behaving — an uncapped retry
is the same class of defect as the one being fixed. Each attempt gets its own
deadline, so a wedged `schtasks` cannot turn a transient failure into a hang.

It stays **non-fatal** after the last attempt (the LogonTrigger is still armed),
but it is no longer **silent**.

### 2. The two numbers are no longer independent

The budget is now a function of the interval:

```go
var schtasksReadiness = readinessConfig{
    budget:   restartOnFailureInterval + defaultReadiness.budget,
    interval: defaultReadiness.interval,
}
```

Task Scheduler documents one minute as the minimum for `RestartOnFailure`, so
that constant is the fixed end of the relationship and the budget is the side
that moves. The task XML's `<Interval>` is **injected from that same constant**
instead of carrying its own `PT1M` literal — so a future edit to either one
cannot desynchronise them, which is the failure mode that produced this bug.
Windows `install`/`restartService` use it; `stopConverge` is unchanged.

A comment saying "keep these in sync" is what was there before, in effect. It
did not work.

### 3. The observability half (#6333's territory, cheap here)

`ServiceManager` gains an **optional** `loadDiagnostics` capability: a backend
whose `Load()` succeeds while a sub-step it treats as non-fatal failed says so.
`ensureLoaded` emits those warnings as progress and appends them to the
`socket not ready` error. The reporter's complaint — `status` said "not running"
with no error at all — now has something in it. Optional because only Windows
has such a sub-step today.

### 4. Testability was the deliverable

The old code was untestable *because* it called `schtasksCmd(...).Run()` inline
and dropped the result. The policy — XML renderer, constants, derivation, retry
loop — now lives in `schtasks_policy.go` **with no build tag**, for the same
reason `schtasks_wrapper.go` has none (#6320): keeping it behind
`//go:build windows` is precisely why no contributor's test binary ever compiled
the two constants together. `retryRun` takes a `sleep` seam and hands each
attempt a context, so the whole retry policy is driven from a test on macOS
without spending real time.

## What is verified, and what is not

**I cannot run Windows.** Stated plainly because @mota-mota will be the one
testing this.

| | |
|---|---|
| `go test -count=1 ./internal/daemon/service/` (+`-race`) | **GREEN** on macOS |
| `GOOS=windows go vet ./internal/daemon/service/...` | **exit 0** |
| Does that prove the fix works on the reporter's machine? | **No** |

`GOOS=windows go vet` proves the Windows-tagged files — production **and** test
— compile and vet clean under the Windows build constraint. It executes
nothing. The vet run is scoped to this one package deliberately: per #7005 a
wider cross-build defaults to `CGO_ENABLED=0` and fails on the tree-sitter
bindings, which has nothing to do with this change.

**UNRESOLVED, and not claimed:**

- Whether `RestartOnFailure` covers an on-demand `/run` instance at all. The
  in-process retry is deliberately independent of the answer.
- Why the first `/run` fails. Nothing here diagnoses that; it makes the failure
  survivable and visible.
- Whether the widened budget changes anything about #7050. It should not — the
  retraction on #7051 applies — but it is not something I can observe.

`Load`, `runTaskNow`, `install` and `restartService` are behind
`//go:build windows`, so no macOS test binary executes them. They are pinned at
the **source** level instead (the technique
`TestWriteUnit_UsesAtomicArtifactWrites` already uses in this package). That is
a weaker instrument than running the code, and it is labelled as such in the
test file — but it is strictly stronger than the nothing that guarded those
lines before, and it is what kills M11/M12 below.

## Fixture axes

**Varied:** the `/run` outcome sequence (succeed-first, fail-then-succeed,
fail-always); the policy (default, over-large, per-attempt-timeout-only);
the injected `<Interval>`/`<Count>` (`PT1M` vs a sentinel `PT7M`/`9` — the axis
that distinguishes "injected" from "hardcoded", since the real value equals the
old literal); context state (live vs pre-cancelled); backend load cleanliness
(warning vs none).

**Held constant:** the task name, wrapper host and wrapper path in every render;
`time.Sleep` replaced by a recorder in every retry test except the deliberate
anti-hang one; the readiness poll interval.

## Mutation table

16 mutants. Each edit proven landed by exact occurrence count before/after, each
compiled under **both** `go vet` and `GOOS=windows go vet` (a non-compiling
mutant proves nothing; guard deletions use `if false && <cond>`), each scored
with `-count=1`, each reverted by `cp` from a shasum-verified backup.

| # | Mutant (call-site and permissive-biased) | Verdict | Killed by |
|---|---|---|---|
| M1 | `retryRun` ignores `normalized()` — uncapped attempts | **DEAD** | `TestRunAttemptPolicyIsClampedNoMatterWhatItIsGiven` |
| M2 | `maxElapsed` drops the attempt-timeout term (under-reports worst case) | **DEAD** | `TestMaxElapsedCountsTheAttemptTimeouts` |
| M3 | budget reverts to the flat 60 s — **the original defect** | **DEAD** | `TestSchtasksReadinessOutlastsRestartOnFailure`, `…IsDerivedNotChosen` |
| M4 | budget outlasts `PT1M` by 1 s, no cold-start slack | **DEAD** | both readiness tests |
| M5 | task XML re-hardcodes `PT1M` instead of consuming the constant | **DEAD** | `TestTaskXMLCarriesWhateverIntervalItIsGiven` |
| M6 | `retryRun` returns nil after every attempt failed | **DEAD** | `TestRetryRunGivesUpAtTheCapAndReportsWhy` (+2) |
| M7 | `retryRun` ignores context cancellation | **DEAD** | `TestRetryRunStopsOnContextCancellation` |
| M8 | `retryRun` drops the per-attempt deadline (**hang risk**) | **DEAD** | `TestRetryRunCannotHangOnAWedgedAttempt` (5.00 s → fail) |
| M9 | `ensureLoaded` stops appending warnings to the error | **DEAD** | `TestEnsureLoadedSurfacesSwallowedLoadWarnings` |
| M10 | `loadWarningsOf` always reports a clean load | **DEAD** | same |
| M11 | **CALL SITE:** `Load` goes back to `_ = schtasksCmd("/run"…).Run()` | **DEAD** | `TestLoadNoLongerDiscardsTheRunExitCode` (source-level) |
| M12 | **CALL SITE:** `install` reverts to the platform-neutral budget | **DEAD** | `TestWindowsEntryPointsUseTheDerivedReadinessBudget` (source-level) |
| M13 | upper attempts clamp removed | **DEAD** | clamp test + `TestRetryRunNeverOutlastsTheSafetyNetItPrecedes` |
| M14 | backoff clamp removed | **DEAD** | same two |
| M15 | sleeps after the final attempt too | **DEAD** | `TestRetryRunGivesUpAtTheCapAndReportsWhy` |
| M16 | `attemptTimeout` clamp accepts anything | **DEAD** | `TestRetryRunNeverOutlastsTheSafetyNetItPrecedes` |

The retry bound was attacked specifically (M1, M13, M14, M15, M16) because
"a retry loop with no cap, or one that can exceed the budget it lives inside"
is this bug's own shape. The clamp bound is asserted for the **worst policy the
clamp permits**, not only for the default anyone can edit:
`5 × 5 s + 4 × 5 s = 45 s < PT1M`. M8 covers the inverse hazard — a transient
failure becoming a hang.

M11 and M12 die to a **source** pin, not to executed behaviour. That is "not
graded here", **not** "ungradable": `Load` is Windows-only only because it calls
`schtasksCmd` directly, and hoisting the `/run` step behind a func field would
let an untagged test drive it for real. The review priced that at ~20 lines and
did not ask for it in this PR; the framing is corrected rather than defended.
What the source pin buys — and does not — is shown by R5 below.

## CI

**This needs the `ci:full` label** — the behaviour is Windows-conditional and
PRs otherwise run ubuntu + windows only, with macOS compiled but never tested.

⚠️ A label sitting on the PR during a `synchronize` does **nothing** — only the
`labeled` event fires the leg. If anything is pushed after labelling, the label
must be removed and re-added.

---

# Review round 2 — HOLD cleared

The review found the **production change correct** (it could not make the retry
hang, exceed its budget, or retry unbounded, and re-derived the clamp arithmetic
independently), and held on **grading**: the PR's headline claim — *"a future
edit to either one cannot desynchronise them"* — was **false as graded**. R1, R2
and R3 each restored a hardcoded literal at a different seam of the derivation
and the suite stayed green. The M5 tautology recurred twice more, on the
mechanism the PR is actually about.

No production behaviour changed in this round. One small production
**extraction** (`loadWarningLog`) was made to close R5 behaviourally rather than
by text.

## The sentinel, and why a literal can no longer coincide with it

`restartOnFailureInterval` and `defaultReadiness.budget` are **both 60 s**. Any
assertion comparing the budget against their sum is satisfied by
`120 * time.Second`, `2 * defaultReadiness.budget`, or
`defaultReadiness.budget * 2`. So the derivation is driven with inputs that are
equal **neither to each other nor to any constant in the package**:
`derivedReadiness(7*time.Minute, {budget: 13s, interval: 41ms}).budget == 7m13s`,
plus a second unrelated pair (`3m` + `2s`). One point can be hit by coincidence;
two cannot be hit by any constant. `intervalXML` is graded the same way — `PT7M`,
`PT23M`, plus the production value and two floor cases.

**R1 survived even after that**, and the reason is worth stating: the production
value and a literal `120 * time.Second` **are the same value**, so no value
assertion can separate *"was produced by the derivation"* from *"happens to match
it"*. Only the assignment can. It is pinned at the source, the same instrument
that closes R3, with the same stated limit — the two value tests prove the
function does the right thing; the pin proves it is the thing being called.

## Re-scored

| # | Mutant | Round 1 | Now | Killed by |
|---|---|---|---|---|
| **R1** | budget → literal `120 * time.Second` | ALIVE | **DEAD** | `TestSchtasksReadinessIsAssignedFromTheDerivation` |
| R1b | `derivedReadiness` body ignores its inputs | — | **DEAD** | `TestDerivedReadinessIsAFunctionOfItsInputs` |
| R1c | derivation doubles the base instead of adding the interval | — | **DEAD** | same |
| R1d | budget → `2 * defaultReadiness.budget` (equals derived value) | — | **DEAD** | assignment pin |
| R1e | assignment keeps the call, swaps the inputs | — | **DEAD** | assignment pin |
| **R2** | `intervalXML` ignores its input, returns `"PT1M"` | ALIVE | **DEAD** | `TestRestartOnFailureIntervalXMLIsValidISO8601` (PT7M/PT23M cases) |
| **R3** | `generateTaskXML` re-hardcodes `"PT1M"` / `3` | ALIVE | **DEAD** | `TestGenerateTaskXMLInjectsTheConstantsRatherThanLiterals` |
| R4 | `retryRun` hands `attempt` the unbounded parent ctx | DEAD | **DEAD** | `TestRetryRunCannotHangOnAWedgedAttempt` |
| **R5** | recorder body gutted (`_ = msg`) | ALIVE | **DEAD** | `TestLoadWarningLogActuallyRecords` (+2) |
| **R6** | per-load `reset` gutted | ALIVE | **DEAD** | `TestLoadWarningLogResetDropsThePreviousLoad` |
| **R7** | `if len(warnings) > 0` made permissive | ALIVE | **DEAD** | `TestEnsureLoadedDoesNotDecorateACleanNotReadyError` |
| R-cap | warning truncation removed | — | **DEAD** | `TestLoadWarningLogCapsAnOverlongWarning` |
| M1–M16 | original table, re-run against the reworked code | DEAD | **DEAD** | unchanged |

**27 mutants, 0 ALIVE.** Each edit proven landed by exact occurrence count
before/after, each compiling under **both** `go vet` and
`GOOS=windows go vet`, each scored `-count=1`, each reverted by `cp` from a
shasum-verified backup with `git diff HEAD` and `git status --porcelain` empty
afterwards.

## R5 is the answer to "is a source pin enough"

**No.** M11 asserts that `Load`'s source *contains* `m.note(...)`. Gutting the
recorder's body leaves that text exactly where it is, records nothing, and
defeats the entire observability deliverable with every test green. A source pin
proves a call is **written**, not that it **does** anything.

So the recorder is now `loadWarningLog`, a small **untagged** type embedded in
the Windows manager. Its recording, its per-load reset and its truncation are
**executed** by tests on every platform. The residual — that `Load` wires up
*that* recorder — is still source-level, and is stated as such.

The truncation also addresses the review's "unbounded, but unbudgeted" note: a
`/run` warning carries up to `maxRunAttempts` untruncated `schtasks`
`CombinedOutput` blobs into an error a caller may print. Capped at 512 bytes,
with the pass-through of a short warning asserted too.

> **Corrected in round 4.** As written above, that cap was **ungraded in the
> direction it exists for** — see CN-2 below. This paragraph overclaimed.

## Two residual risks, flagged rather than left to be rediscovered

Both raised by the review, both judged low-risk, **neither testable from macOS —
so both stay UNRESOLVED**:

1. **`CombinedOutput` can outlive a context kill.** It waits for the output pipes
   to close, not merely for the killed child, so a grandchild inheriting those
   pipes would keep `retryRun`'s attempt alive past its deadline. Judged low risk
   (Task Scheduler launches the task from its own service context, so the wrapper
   should not inherit `schtasks`' handles) — but it is a judgement, not a
   measurement.
2. **`retryRun` has no total-elapsed guard of its own.** The 45 s ceiling holds
   *given an attempt that returns once its context is done*. The test's fake
   respects the context by construction, so the test cannot see case 1 either.

Also unchanged and worth repeating: `install()` and `restartService()` pass `nil`
`onProgress` (pre-existing), so on the real Windows path the new progress line
goes nowhere and **only the returned error carries the warnings**.

## Deliberately not done

- **The ~20-line seam** to make `Load` itself drivable from an untagged test. The
  review explicitly did not ask for it in this PR, and it is a production change
  to a Windows path I cannot execute. The framing above is corrected instead.
- **Untagging the `<UserId>` tests.** `schtasks_userid_windows_test.go` is still
  `//go:build windows` even though everything it renders is now untagged — but
  `TestCurrentUserSID_Shape` in the same file genuinely is Windows-only, so it is
  a file split, not a tag removal. Out of scope for a hold that is about grading.
- `internal/install/watchers/watchers.go:412` — a different task with no budget
  paired to it. Tracked separately as **#7059**.

---

# Round 4 — the cap was ungraded in the direction it exists for

**CN-2, raised by the maintainer, and it is the R1 tautology a third time.**
Raising `maxLoadWarningLen` from `512` to `1 << 30` left the whole suite
**green**.

The round-3 test scaled **both** its input (`maxLoadWarningLen*3`) and its
ceiling (`maxLoadWarningLen + marker`) with the constant, so **no value of the
cap could fail it**. I had guarded the over-aggressive direction — short
warnings pass through untouched — and left its sibling, *that a long warning is
actually cut*, open. One axis graded, its sibling ungraded, in the fix for
exactly that. It is reachable: `schtasks_windows.go` does `m.note(err.Error())`,
that error carries `CombinedOutput`, and the warnings are appended to the
user-facing "socket not ready" error a stuck user is trying to read.

**The cap is now pinned against literals rather than against itself**
(`64 <= maxLoadWarningLen <= 4096`), plus a truncation assertion built from
fixed literals — a 100 000-byte input against a `4096 + marker` ceiling — so it
stays meaningful however the constant moves.

## Two more holes, found by re-scoring rather than by assuming the fix worked

| | |
|---|---|
| **CN-2f** (keeps the **tail** instead of the head) was **ALIVE** | The fixture was a homogeneous run of `x`. A filler identical at both ends cannot tell *"kept the front"* from *"kept the back"* — every assertion over it is satisfied by either. The fixture now carries **distinct head and tail markers**, and the head is what is demanded: it is the half that names the failure. |
| **RUNE-2** (unbounded backoff) was **ALIVE** | On **valid** UTF-8 the bounded and unbounded loops stop in the same place, so the rune-boundary test cannot see the difference. Only input that is **not UTF-8 at all** distinguishes them — which is precisely the case the bound exists for. There is now a raw-continuation-byte fixture for it. |

## The rune-boundary question — answered, with its reachability left UNRESOLVED

**The mechanism is real, so the fix is in; whether it is reachable is not
something I can determine, so it is not claimed.**

`maxLoadWarningLen` counts **bytes**. `msg[:cap]` can split a UTF-8 sequence and
append `… (truncated)` to a broken rune. The cut now backs off to a rune start,
**bounded by `utf8.UTFMax-1`** so that input which is not valid UTF-8 at all — a
raw OEM-codepage blob, where continuation-shaped bytes are common — loses at
most three bytes instead of unwinding arbitrarily far.

What made this worth doing rather than arguing about: this repo **already
documents** that `schtasks` speaks the user's locale (the `//nolint:localematch`
English text-match in `Unload`) and that a non-ASCII `%USERPROFILE%` is routine
enough to have caused #6325. That is grounded in our own source.

**UNRESOLVED:** whether Windows hands Go *valid UTF-8* for `schtasks` output on
a non-English locale at all. Console output arrives in the OEM codepage, and Go
returns the raw bytes; if they are not UTF-8, rune-boundary truncation neither
helps nor hurts. That cannot be determined from macOS and is not asserted. The
fixtures are **synthetic and say so in the test** — they assert the recorder is
correct *if* handed multi-byte text, and are not a claim about what `schtasks`
emits.

## Round-4 scores

|---|---|---|---|
| **CN-2** | cap → `1 << 30` (the maintainer's mutant) | **DEAD** | `TestLoadWarningLogCapIsActuallyABudget` + truncation test |
| CN-2b | cap → `65536` (a plausible-looking widening) | **DEAD** | same |
| CN-2c | cap → `8` (over-aggressive) | **DEAD** | `…ActuallyRecords` (+2) |
| CN-2d | truncation removed entirely | **DEAD** | `…TruncatesAnOverlongWarning` (+2) |
| CN-2e | truncates but drops the marker | **DEAD** | `…TruncatesAnOverlongWarning` (+1) |
| **CN-2f** | keeps the **tail** instead of the head | ALIVE → **DEAD** | head/tail-marked fixture |
| CN-2g | mangles a short warning too | **DEAD** | `…ActuallyRecords` |
| CN-2h | marker emptied (truncation says nothing) | **DEAD** | `…TruncatesOnARuneBoundary` |
| RUNE-1 | rune backoff disabled | **DEAD** | `…TruncatesOnARuneBoundary` |
| **RUNE-2** | backoff unbounded | ALIVE → **DEAD** | `…BoundsTheBackoffOnInvalidUTF8` |
| RUNE-3 | cut overshoots by one byte | **DEAD** | truncation + backoff tests |
| RUNE-4 | backoff walks **forward**, past the cap | **DEAD** | both |

**12 round-4 mutants, 0 ALIVE. The earlier 27 re-run, 0 ALIVE.** Same protocol
throughout: exact occurrence count before/after, compiling under **both**
`go vet` and `GOOS=windows go vet`, `-count=1`, reverted by `cp` from a
shasum-verified backup, `git status --porcelain` empty.

> **Process note, recorded because it nearly cost something.** The round-4
> driver crashed on its own landed-assertion **after writing a mutant and before
> restoring**, leaving `RUNE-3` applied in the worktree. It surfaced as two test
> failures that looked like bugs in the new tests; the tests were right and the
> tree was dirty. The driver now wraps every mutation in `try/finally` so a
> crash cannot leave a mutant behind. The tests detecting the stray mutant
> precisely is, incidentally, the best evidence they work.

---

## ⚠️ CI: a green tick on this PR is not three-platform coverage

Confirmed empirically **twice on this PR** by the maintainer: a push fires
`synchronize`, and the `ci:full` label **sitting on the PR during a
`synchronize` does nothing**. Only the `labeled` event starts the macOS test
leg, so the label has to be **removed and re-added after every push**.

Until that is done the run contains `compile (macos-latest)` and **no
`test (macos-latest)`**. Do not read the resulting green as macOS having been
tested. And none of it is Windows execution regardless — see "What is verified,
and what is not" above.


cajasmota (commented):
# Adversarial review — HOLD (narrow; ~30 test-only lines clears it)

Independent reviewer; I did not write this. Own worktree at `b37d12ea`, private `GOCACHE`.
Baseline `go test -count=1 ./internal/daemon/service/` **GREEN**, `-race` **GREEN**,
`go vet` and `GOOS=windows go vet ./internal/daemon/service/...` both **exit 0**,
`gofmt -l` clean, `go build ./cmd/grafel` clean.

**The production change is right.** The `/run` exit code is acted on, the retry is
structurally bounded, the per-attempt deadline genuinely prevents a hang, and the
derivation runs in the correct direction. I could not make the retry loop hang, exceed
its budget, or retry unbounded. **The hold is entirely about grading, not about
behaviour** — and it lands on the one claim the PR is built on.

---

## 1. The derivation is real in the code, and ungraded at every seam but one

The `PT1M` literal is gone from `daemonTaskXMLTemplate`, the budget is genuinely
`restartOnFailureInterval + defaultReadiness.budget`, and `generateTaskXML` genuinely
injects `restartOnFailureIntervalXML()`. The direction is right:
`TestRestartOnFailureIntervalXMLIsValidISO8601` pins `restartOnFailureInterval >= time.Minute`,
so the interval is the fixed end and the budget is the side that moves. That matches
what the PR body claims.

No surviving `PT1M` on the daemon path. One does survive at
`internal/install/watchers/watchers.go:412` — a *different* schtasks task with no budget
paired to it, so out of scope; noting it only so it is not a surprise later.

**But the mechanism that is supposed to stop the two desynchronising again is not tested
at either end.** Three mutants, each proven landed by exact occurrence count, each clean
under `go vet` *and* `GOOS=windows go vet`, each scored `-count=1`, each reverted by `cp`
from a `git show HEAD:`-derived backup with `git diff HEAD` and `git status --porcelain`
empty afterwards:

| # | Mutant | Landed | Verdict |
|---|---|---|---|
| **R1** | `schtasksReadiness.budget` → literal `120 * time.Second` | 1→0 / 0→1 | **ALIVE** |
| **R2** | `restartOnFailureIntervalXML()` → `_ = minutes; return "PT1M"` | 1→0 / 0→1 | **ALIVE** |
| **R3** | `generateTaskXML`: `RestartInterval: restartOnFailureIntervalXML()` → `"PT1M"`, `RestartCount: restartOnFailureCount` → `3` | 1→0 / 0→1 | **ALIVE** |

**R1 is the headline one.** `TestSchtasksReadinessIsDerivedNotChosen` computes
`want := restartOnFailureInterval + defaultReadiness.budget` and compares the value
against it — but `restartOnFailureInterval == defaultReadiness.budget == 60s` today, so a
flat `120 * time.Second`, or `2 * defaultReadiness.budget`, or
`defaultReadiness.budget + defaultReadiness.budget`, all satisfy it. The test named
*IsDerivedNotChosen* cannot distinguish derived from chosen. That is the exact trap M5
was written to escape, sitting on the PR's primary claim.

**R3 is the sharper one, because M5 is specifically claimed to close it.** The sentinel
fixture (`PT7M`/`9`) is genuine and it does break the tautology — but only for
`renderTaskXML`, i.e. "the template honours its parameter". Nothing grades the *caller
that fills the parameter*. Put the literal back at the injection site and the whole suite
stays green, `TestTaskXMLCarriesWhateverIntervalItIsGiven` included. The template is
guarded; the injection is not. That is half a mechanism, one level up from where it was
fixed.

**R2 is the same shape a third time.** `TestRestartOnFailureIntervalXMLIsValidISO8601`
asserts `got != "PT1M"` → fail: it pins the output to the literal a correct
implementation happens to produce, so a function that ignores `restartOnFailureInterval`
entirely passes it.

So — to answer the question directly — **M5's tautology does recur, twice, and on the
headline claim.** The sentinel technique the author invented for the template is the
right instrument; it just needs applying to the two seams either side of it.

## 2. The retry bound holds. Re-derived from the clamp code

`normalized()` clamps `attempts` to `[1, 5]`, `backoff` to `[0, 5s]`, and `attemptTimeout`
to `(0, 5s]` (the `<= 0 || > max` form covers both ends). `retryRun` normalizes its own
argument, so the worst permitted wall time is `5×5s + 4×5s = 45s < PT1M`. Confirmed.
`maxElapsed` reproduces that arithmetic from the same clamp constants and compares against
a *different* constant (`restartOnFailureInterval`), so
`TestRetryRunNeverOutlastsTheSafetyNetItPrecedes` is not tautological. I could not
construct an input that exceeds the bound: every field is clamped and nothing is read
unnormalized on the hot path.

**M8's watchdog does bound the case it names.** I checked the permissive variant the table
does not cover — keep `context.WithTimeout`, but hand `attempt` the *parent* `ctx` instead
of `attemptCtx` (`_ = attemptCtx; err := attempt(ctx, i)`; 1→0 / 0→1; both vets clean):

| **R4** | `retryRun` passes the unbounded parent ctx to `attempt` | **DEAD** — `TestRetryRunCannotHangOnAWedgedAttempt` (5.00s) |

That is a real seam, really graded.

**Residual, UNRESOLVED (not a blocker):** `retryRun` has no total-elapsed guard of its own
— the 45s ceiling holds *given an attempt that returns once its context is done*.
Production's attempt is `schtasksCmdContext(ctx, …).CombinedOutput()`, and `CombinedOutput`
waits for the output pipes to close, not merely for the killed child. I judge the risk low
(Task Scheduler launches the task from its own service context, so the wrapper should not
inherit schtasks' pipe handles), but I cannot test it from macOS and the test's fake
respects ctx by construction.

## 3. The observability half is ungraded at its call site

| **R5** | `noteLoadWarning` body → `_ = msg` (records nothing) | 1→0 / 0→1 | **ALIVE** |
| **R6** | `Load()` drops the `m.loadWarnings = nil` per-load reset | 1→0 / 0→1 | **ALIVE** |
| **R7** | `ensureLoaded`: `if len(warnings) > 0` → `if true \|\| len(warnings) > 0` | 1→0 / 0→1 | **ALIVE** |

**R5 is the direct answer to "is the source pin enough".** M9/M10 grade `ensureLoaded` and
`loadWarningsOf` — both untagged, both genuinely executed, both fine. M11 grades that
`Load`'s *source text* contains `noteLoadWarning(`. R5 leaves that text exactly where it
is and makes the function record nothing: the entire observability deliverable is
defeated and every test passes. The source pin proves the call is written, not that it
does anything.

R6 is low-reachability (each entry point builds a fresh manager, so `Load` runs once per
instance today) — alive, but I could not name a production input that distinguishes it.
R7 is cosmetic: every non-Windows "socket not ready" error would gain a trailing ` ()`.
Neither blocks.

On the error path itself: no duplication bug — warnings deliberately go to both progress
and the error, the original error is wrapped with `%w`, and `errors.Is` still works. The
text is not unbounded, but it is *unbudgeted*: the warning is the full `retryRun` error,
carrying up to 3 × untruncated `schtasks` `CombinedOutput`. Worth a cap. Also note that
`install()` and `restartService()` pass `nil` onProgress (pre-existing), so on the real
Windows install path the new progress line goes nowhere and only the returned error
carries the warnings.

## 4. On M11/M12 — "ungradable on macOS" is overstated

`install` and `restartService` are one-line wrappers, so a source pin there is defensible.
`Load` is not: the only thing keeping it Windows-only is that it calls `schtasksCmd`
directly. Hoisting the `/run` step behind a `func(context.Context) error` field on
`schtasksManager`, defaulted in `newServiceManager`, would let an untagged test drive
`Load`'s warning-recording behaviour for real — and would have killed R5. I price that at
~20 lines, most of it test. I am **not** asking for it in this PR, but the table should
say "harder", not "ungradable", because R5 shows what the source pin actually buys.

## 5. Build tags and the split — clean

`schtasks_policy.go` with no build tag is correct and consistent with
`schtasks_wrapper.go`/#6320. Nothing Windows-only leaked in: no syscall, no
`%LOCALAPPDATA%`, no SID. `schtasksReadiness` is now a package-level var on every platform
and is harmless there. No test was orphaned — `schtasks_userid_windows_test.go` was
correctly repointed at the production `renderTaskXML` instead of re-parsing the template
itself, which is a real improvement. One free win missed: that file is still
`//go:build windows`, so the `<UserId>` conditional tests still never run on macOS/Linux
even though everything they touch is now untagged.

On the retraction: I checked it independently and **it is correct**. `RestartOnFailure` is
a Task Scheduler property and fires irrespective of our budget; `MultipleInstancesPolicy:
IgnoreNew` is in the rendered XML and is now pinned by
`TestTaskXMLKeepsIgnoreNewInstancePolicy`. Widening the budget does not manufacture a
second `serve`. Nothing to resurrect.

---

## Verdict: HOLD

Not because the fix is wrong — it is right, and I want it in the reporter's hands. The
hold is that the PR's own headline deliverable ("a future edit to either one cannot
desynchronise them") is **demonstrably false as graded**: R1, R2 and R3 each restore a
hardcoded literal at a different seam of that derivation, and the suite stays green.
Merging as-is means the next edit reintroduces #7051 silently — precisely what this PR
exists to prevent.

### What clears the hold — test-only, ~30 lines, no production behaviour change

1. **Grade the budget derivation with a sentinel**, the way M5 grades the template.
   Extract `func derivedReadiness(interval time.Duration, base readinessConfig) readinessConfig`,
   point `schtasksReadiness` at it, and assert
   `derivedReadiness(7*time.Minute, base).budget == 7*time.Minute + base.budget`.
   Kills R1. The current value-equality test cannot, because the two constants it is meant
   to keep distinct are both 60s.
2. **Grade `restartOnFailureIntervalXML` as a function of its input**, not against
   `"PT1M"`: `intervalXML(7*time.Minute) == "PT7M"`, plus the existing sub-minute floor
   case. Kills R2.
3. **Pin the injection call site** with the instrument already used for
   `install`/`restartService`:
   `funcBodySource6325svc(t, "schtasks_windows.go", "generateTaskXML")` must contain
   `restartOnFailureIntervalXML()` and must not contain `"PT1M"`. Kills R3 — and it is the
   site M5 is claimed to protect.

Optional, not blocking: extend the same source pin to `noteLoadWarning` (or take the
~20-line seam above) for R5; cap the warning text length; correct the M11/M12 framing from
"ungradable" to "not graded here".

Everything else — the retry bound, the per-attempt deadline, the non-fatal-but-not-silent
policy, the build-tag split, the error wrapping — I tried to break and could not.

**Cannot verify from here:** anything that executes on Windows. `GOOS=windows go vet`
compiles the tagged production and test files and runs none of them; the author says so and
is right to. Whether the retry actually fixes @mota-mota's machine is unresolved, and
correctly not claimed.

PR #7063, opened 2026-09-15
fix(#6912): kotlin fields recorded no declared type at all — capture it, then bind it (91 edges, 91 bound)

cajasmota:
## Arm: **kotlin** — the first arm on #6912 that is genuinely **capture, then bind**

`91 edges / 91 bound / 0 dangling` over **447 `.kt` files in 3 repos** (ktor 195, ktor-samples 151, exposed 101), 1490 field entities. Nine arms: C#, proto, rust, go, java, F#, php, swift, **kotlin**. (cpp is in review at #7057 and is untouched here.)

---

### 1. The no-type-datum premise HELD — and it was re-verified, not inherited

`kotlin.go`'s two field producers — `buildProperty` and `emitPrimaryConstructorFields` — built their `SCOPE.Schema`/`field` records with **no `Properties` map, no `Signature`, no type datum of any kind**. Probed on `e51d69fce` by extracting a file with six property shapes and printing every record, rather than trusting the issue text: **#6912 asserted exactly this about JAVA and arm F found it false.**

So capture and bind land **together**. The capture writes to `Metadata` **scratch** that the attach pass deletes, and when the map is left empty it is set back to `nil`, so a field whose type named nothing addressable is byte-identical to what it was before this pass existed. **No entity field changes; the graph gains edges and nothing else.** A capture shipping alone would have been one more unhashed property that nothing reads — the state this issue exists to fix.

### 2. Where the type lives, and the two things beside it that must not be read

```
class_parameter       [binding_pattern_kind] [modifiers] [simple_identifier] [:] TYPE [= default]
property_declaration  [binding_pattern_kind] [user_type RECEIVER] [.]
                      [variable_declaration [simple_identifier] [:] TYPE] [getter|property_delegate]
```

- **An extension property's RECEIVER is a sibling of `variable_declaration`, outside it.** A capture that walked the whole `property_declaration` would emit `p -> Recv`: right name, right file, admissible kind, admissible subtype — **and not a declaration of the field's type**. This is kotlin's instance of the shape that cost arm G half its edges (#7047) and arm cpp 16 of 39 (#7057). Scoping the capture to the variable_declaration excludes it structurally. **Graded: M17.**
- **A `class_parameter`'s annotations sit before the colon and a default value after it.** tree-sitter-kotlin spells an annotation's name with a `user_type` — the very node the scanner collects from — so without the post-colon window, `@Marker val o: Order` in a file that also declares `annotation class Marker` emits `o -> Marker`. **Graded: M11, M10.**

### 3. Resolver tier(s) and the kinds each weighs — **driven, not read**

| target | tier | counts |
|---|---|---|
| `SCOPE.Component` (class / data_class / interface / enum / object) | `Index.lookupStructural` → `lookupLocationKind` filtered to `componentKindFamily`, which **returns before `ambigLocation` is consulted** | family kinds only (arm F's rule) |
| `SCOPE.Schema`/`type_alias` | invisible to that filter, so the ref **falls through to `ambigLocation`**, which is kind-agnostic | every kind (arm D's rule) |

Both tiers are exercised end-to-end against `resolve.BuildIndex` + `resolve.ReferencesEmbedded`: `..._ResolverBindsEveryEdge` (5 targets: class, interface, enum, object, typealias) and `..._ResolverBindsTheTypeAliasTierToo`, which grades the alias tier **on its own** so it cannot look covered by the four component rows.

**Why kotlin needs the split more than any predecessor.** Kotlin emits **three** same-file, same-name rivals *outside* the family:

| source | Component record | rival kind (outside family) |
|---|---|---|
| `enum class Status` | `SCOPE.Component/enum` | `SCOPE.Enum` value-set |
| `object Pages { const val … }` | `SCOPE.Component/object` | `SCOPE.Enum` **const-group** value-set, named after the object |
| `@Service class Billing` | `SCOPE.Component/class` | `SCOPE.Service`, named after the class (`SCOPE.Service` is deliberately not in `componentKindFamily`, refs.go:2347) |

The unit of the count is the **KIND**, not the record (`graph.EntityID` excludes Subtype, so two same-file records sharing a Kind are one node — #7038's defect, graded by `Unit_DuplicateComponentRecordsAreOneNode`), and the family list is `componentKindFamily` **closed under the index's trim alias** (8 entries; `SCOPE.Class` included, graded by `Unit_ScopeClassParticipatesViaTheTrimAlias`).

### 4. The measured counterfactuals, with the dropped rows named

Same corpus, same harness, one rule changed at a time, restored from a shasum-verified backup each time.

| rule | edges | delta |
|---|---|---|
| **shipped (split)** | **91** | — |
| **arm D** — all kinds, both tiers | **66** | **−25 (27.5%), every one an enum target** |
| **arm C** — `Kind != "SCOPE.Component"` | **88** | **−3, every one a typealias target** |
| **arm F** — family kinds, both tiers | **91** | 0 on this corpus — **but wrong, see below** |

Arm D's 25 dropped rows are `ForeignKeyConstraint.{onDelete,onUpdate,deleteRule,updateRule} -> ReferenceOption`, `JoinPart.joinType -> JoinType`, `EngineOption.engine -> TableEngine`, `{Offset,Unbounded}WindowFrameBound.direction -> WindowFrameBoundDirection`, `WindowFrameClause.unit -> WindowFrameUnit`, `VectorDistance.metric -> VectorDistanceMetric`, `Clause.{action,type} -> Clause{Action,Condition}`, `Statement.type -> StatementType`, `StatementType.group -> StatementGroup`, `IdentifierManagerApi.oracleVersion -> OracleVersion`, `H2Dialect.{h2Mode,majorVersion} -> H2{CompatibilityMode,MajorVersion}` (exposed), and `ApplicationConfigValue.type -> Type`, `RoutingPathSegment.kind -> RoutingPathSegmentKind`, `EnumConfig.{enum,enumList,enumMap} -> TestEnum`, `CompressedResource.compression -> CompressedFileType`, `EnumContainer.{enumValue,enumList} -> TestEnum` (ktor).

Arm C's 3: `JoinPart.conditions -> JoinCondition` (exposed), `Config.digestProvider` and `DigestAuthenticationProvider.userNameRealmPasswordDigestProvider -> DigestProviderFunction` (ktor-samples).

**Arm F's rule is a corpus-relative tie and still wrong** — no corpus file has a typealias colliding with a non-family kind. It dies in the suite (`..._TypeAliasTargetIsRefusedWhenAnyKindCollides`), and the refusal's **reason is proved rather than asserted**: `..._RefusedAliasEdgeWouldActuallyDangle` hand-emits the declined edge through the **real resolver** and watches it dangle, with a Component row carrying the *same* rival (`fun Handler()`) binding in the same test. Arms C/D/E carry a "would guarantee a dangling stub" justification that arm E found false in general; this arm measured it for kotlin instead of repeating it.

### 5. Same-file conjuncts — there are **THREE**, and each is graded separately

php's author found three where every table said two; grading one is what makes its sibling look covered.

| # | site | direction it fails in | grader | mutant |
|---|---|---|---|---|
| A | pass 2, the **target** scan | a foreign record wrongly **becomes** a target | `Unit_TargetsAreScopedToTheRequestedFile` | **M4 DEAD** |
| B | pass 1, the **ambiguity** scan | a foreign collision wrongly **suppresses** an edge | `Unit_AnotherFilesCollisionDoesNotShadowThisFile` | **M5 DEAD** |
| C | attach, the **anchor** scan | a non-field record wrongly grows an edge | `Unit_OnlySchemaFieldsAreAnchors` (with a positive control so the two refusals cannot pass vacuously) | **M6 DEAD** |

### 6. The not-a-declaration audit

A 100% bind rate is not evidence, so each kotlin construct that is *not* a type declaration was enumerated and checked from source:

| construct | outcome |
|---|---|
| **companion object** (named or not) | mints **no entity at all** — `walk` has no `companion_object` case. Unreachable as a target; the test **asserts the no-entity premise** so it cannot pass vacuously. |
| **root-package `import Order`** | a **bare-named `SCOPE.Component/import`** — kotlin's import placeholder, exactly arm D's. **Refused by the subtype allow-list**, graded from source with the carrier's existence asserted. Removing the allow-list (M3) makes the field bind to it. |
| **dotted `import com.acme.Order`** | Name keeps the full dotted path, which no bare `type_identifier` can equal; pinned so a rename of that rule surfaces here. |
| **file carrier (#577)** | refused by subtype; Name is the file path. |
| **`expect` / `actual` / `external` class** | **ADMITTED** — these *are* declarations, routed through the ordinary `class_declaration` arm. Said out loud rather than left implicit. |
| **`object X` vs `class X`** | **ADMITTED** — an object declaration declares a type (`val r: Registry` names it). |
| **typealias** | **ADMITTED**, via its own tier. |
| **extension property receiver** | excluded structurally (§2). |
| **supertype list** (`class Sub : Base()`) | mints no entity. **Kotlin has no extension-carrier construct at all**, which is the single reason this arm needs no `Metadata` marker like arm G's. |
| **`SCOPE.Enum` value-sets, `SCOPE.Service` twins, `SCOPE.Schema/field`** | refused as targets while staying invisible to the family count — the two halves of the rule, graded separately. |

Corpus cross-check: **all 91 bound targets' source spans carry a real `class` (77) / `interface` (11) / `typealias` (3) keyword**, with 0 in an "other" bucket.

### 7. The SOURCE endpoint audit (cpp's F4)

cpp found 68 of 254 "fields" were member **functions** with the return type captured. Kotlin's field records come from exactly two producers, and **both are genuine property declarations**: a class/object/interface-body `property_declaration`, and a primary-constructor `class_parameter` carrying a `binding_pattern_kind`. A function is a `function_declaration` → `SCOPE.Operation` and is **never** a field. The three shapes worth naming are all real properties with a real declared type, and all three are graded in `..._AnchorShapeSpace`:

- getter-only (`val x: Order get() = …`) — **3 corpus edges**
- delegated (`val x: Order by lazy { … }`) — **5 corpus edges**
- function-typed (`val cb: (Order) -> Unit`) — collected through the function type's parameters and return

A **non-`val` constructor parameter** is not a property and mints no field entity; asserted in the same test so the anchor list is exhaustive rather than illustrative.

### 8. Fixture axes — VARIED vs HELD CONSTANT

**VARIED across the file** (each in its own test, so no row moves two axes at once):
- **type expression**: bare · nullable · generic arg · nested generic · array element · `out` variance · function-type parameter · function-type return · `suspend` function type · dotted-qualified carrying an argument · dotted-qualified whole · star projection · function-type receiver · type parameter (13 rows, enumerated)
- **anchor**: class-body property · primary-constructor `val` · object-body property · interface property · getter-only · delegated · extension property in a class body · non-`val` ctor param
- **target kind/subtype**: class · data_class · interface · object · enum · annotation class · expect class · external class · typealias · undeclared · top-level function · import carrier · file carrier · `SCOPE.Enum` value-set · const-group value-set · `SCOPE.Service` · `SCOPE.Operation` · `SCOPE.Schema/field`
- **file**: same-file vs other-file, in **both** failure directions
- **rival kind**, held against a **fixed** target kind and vice versa (`..._TypeAliasTargetIsRefusedWhenAnyKindCollides` / `..._ComponentTargetSurvivesTheSameCollision` differ in **one** token)

**HELD CONSTANT** inside each test: the edge Kind and `ref_kind` spelling, the owner type name, the file layout. The axes deliberately **crossed** rather than varied separately: *extension receiver × same-file target of that name*, and *annotation × same-file annotation class* — those intersections are where arms G and cpp lost their first cuts.

### 9. Mutation testing — 21 mutants, **21 DEAD**

Call-site, permissive-biased, each proving its edit landed with an exact occurrence count, each `go vet`-clean (`if false && …` where deleting a guard would leave a variable unused), each reverted by `cp` from a shasum-verified backup with the driver's `trap` guaranteeing restore even on a crash.

|---|---|---|---|
| M1 | component tier counts ALL kinds (arm D) | DEAD | 6 tests incl. all three rival-kind tests + resolver |
| M2 | alias tier counts family only (arm F) | DEAD | `TypeAliasTargetIsRefusedWhenAnyKindCollides` |
| M3 | no Component subtype allow-list | DEAD | `ImportCarrierIsNeverATarget` |
| M4 | pass 2 ignores SourceFile | DEAD | `Unit_TargetsAreScopedToTheRequestedFile` |
| M5 | pass 1 ignores SourceFile | DEAD | `Unit_AnotherFilesCollisionDoesNotShadowThisFile` |
| M6 | any record is an anchor | DEAD | `Unit_OnlySchemaFieldsAreAnchors` |
| M7 | dotted type takes its last segment | DEAD | 3 tests |
| M8 | collect `type_identifier` outside `user_type` | DEAD | shape-space row (function-type receiver) |
| M9 | no type-parameter filter | DEAD | `TypeParameterIsNeverATargetEvenWhenShadowed` |
| M10 | scan past the `=` default value | DEAD | `ConstructorParameterDefaultValueIsNotScanned` |
| M11 | no post-colon window | DEAD | `AnnotationOnAConstructorParamIsNotItsDeclaredType` |
| M12 | no per-target dedup | DEAD | `OneEdgePerTargetEvenWhenNamedTwice` |
| M13 | stash not deleted | DEAD | `StashIsDeletedFromEveryRecord` |
| M14 | family drops `SCOPE.Class` | DEAD | `Unit_ScopeClassParticipatesViaTheTrimAlias` |
| M15 | count RECORDS not KINDS (#7038's defect) | DEAD | `Unit_DuplicateComponentRecordsAreOneNode` |
| M16 | `ref_kind` near-miss (`field_target_types`) | DEAD | 19 tests, all against independent literals |
| M17 | capture the whole `property_declaration` (the #7047 shape) | DEAD | 16 tests incl. `ExtensionPropertyReceiverIsNotTheDeclaredType` |
| M18 | no typealias tier | DEAD | 3 tests incl. both resolver tests |
| M19 | `Relationships` assigned, not appended | **was ALIVE** → **GRADED** → DEAD | see below |
| M20 | suppress the self-reference (arm D/F's rule) | DEAD | `SelfReferentialFieldGetsAnEdge` |
| M21 | leave an empty `Metadata` map behind | DEAD | `StashIsDeletedFromEveryRecord` |

**M19 is the one worth reading.** No kotlin producer gives a field record an outbound edge today, so an assigning mutant is invisible from source and the honest first verdict was *alive, not production-reachable*. "Too expensive to kill" is a claim, not evidence: the distinguishing input is a field record pre-seeded with one relationship, it cost **four lines** (`Unit_AnExistingOutboundEdgeIsPreserved`), and M19 re-scored **DEAD**.

**M11's kill is narrower than its name.** The post-colon window is load-bearing on the **class_parameter** anchor only — on the property anchor the `variable_declaration` scope already excludes annotations, so the property-side annotation test is killed by M17, not M11. Both rows are kept; the attribution is stated rather than implied.

### 10. Gates

- **`go test -count=1 ./cmd/grafel/` — PASS (180s)**, fresh `GRAFEL_HOME` + `GRAFEL_DAEMON_ROOT` under `/private/tmp`, never the real `~/.grafel`. The **digest did not move, and the reason is "nothing to move", not "the instrument is blind"**: `writeSpanFixture6118` writes 18 files — cs ×7, java, ts, js, py ×6, go, rs — and **no `.kt` at all**. Relationships *are* hashed as `R <from> -[Kind]-> <to>`, so a new REFERENCES edge would move it; there is simply no kotlin file for one to appear in. The capture step cannot move it either, since it touches no hashed entity field (`Metadata` is not among the eight, and is deleted regardless).
- **`scripts/quality/run.sh --runs 1 --ratchet` — OK, 38 fixtures held**, run **separately** from the `cmd/grafel` leg (never concurrently — 16GB machine). The real check, `git status --porcelain`, shows **`baseline.json` unmodified**, so no ceiling drop hid behind the exit code. **This gate is VACUOUS for this arm and is offered as nothing more**: `kotlin-spring-mini` holds three `.kt` files and none can produce an edge — `User`/`CreateUserRequest`'s fields are `Int`/`String`, `UserController`'s two properties carry no type annotation, and both `Platform*.kt` declare no properties. Enumerated file by file rather than inferred from the unchanged number.
- `go test -count=1 ./internal/extractors/kotlin/` green; `./internal/resolve/ ./internal/treesitter/ ./internal/generated/ ./internal/patterns/` green (every package outside the arm whose tests mention kotlin); `gofmt -l` clean; `go vet` clean.

### 11. Honest limits

- **Recall ceiling: 91 of 1490 field entities (6.1%).** `91/91 bound` is a **bind rate, not completeness.** Same-file targets only (#6976/#6369) — every cross-file and every imported type gets nothing, and in kotlin that is most of them. The pre-existing extractor ceilings on top of that: a **destructured property** (`val (a, b) = …`) mints no field entity at all; a **companion-object property** mints none either (the walk's companion body never reaches `buildProperty`); a **top-level property** mints none. None of those three is introduced here and none is fixed here.
- **The function-type RECEIVER is not descended.** `Order.() -> Unit` yields `Unit` and not `Order`, because the receiver occupies the same position a dotted qualifier does and `Outer.Inner.() -> Unit` makes the two indistinguishable without a second rule. A stated recall floor, **not** a claim that the receiver names nothing.
- **Corpus caveat:** 447 files but effectively **3 repos, two of them the same project family** (ktor + ktor-samples = 346 of 447). A clean result and a narrow one.
- **A corpus-relative zero is not a bound on the language.** Arm F's rule tying at 91 says nothing about whether the shape it mishandles occurs in Kotlin — it does, it is `typealias X` beside `fun X()`, and it is graded from source.
- **No claim is made that #6726's orphan rate moves.** Their corpus is csharp/fsharp/java/protobuf; kotlin is not in it.

---

Refs #6912, #6984, #6991, #7000, #7036, #7039, #7040, #7042, #7043, #7038, #7047, #7056, #6976, #6369.


cajasmota (commented):

Reviewed on my own checkout of `7c667974e` (worktree `agent-a191f726171129aa3`, `git log --oneline -1` = `7c667974e`, tree clean before and after every mutant). I did not write this code and I went looking for the permissive direction first.

This is one of the strongest arms on #6912. §3's two-tier claim is **true at source** (I re-derived it rather than reading the table), §5's conjunct count is honest, §6's companion-object premise is **true** (I probed it), and the three rival-kind tests are not vacuous (each asserts its premise, and `ConstGroupValueSetSiblingDoesNotRefuseTheObjectTarget` really does mint the value-set — I checked, it needs ≥2 `const val`, and the fixture has two). `RefusedAliasEdgeWouldActuallyDangle` is the right shape: both rows carry the *same* rival and differ in one token.

It still ships one **wrong bound edge** and one **ungraded half of a shipped guard**, both in the same guard, and both contradicted by prose in the PR body and the source.

---

### F1 (blocking) — the type-parameter shadow filter only covers the IMMEDIATE declaration; an ENCLOSING declaration's type parameters leak and bind

`field_type_refs.go`'s doc-comment:

> TYPE PARAMETERS ARE EXCLUDED BY NAME. […] the over-fire arm F had to pin as KNOWN-WRONG […] Kotlin's does too, **but the declaration's `type_parameters` list is right there at the emit site, so the shadow is refused instead of documented.**

and `TestKotlinFieldTypeRefs_TypeParameterIsNeverATargetEvenWhenShadowed`:

> Java's arm pinned this as a KNOWN OVER-FIRE; **Kotlin refuses it**, because the declaration's `type_parameters` list is in hand at the emit site.

`kotlinTypeParameterNames(node, …)` reads the type parameters of the declaration whose body/constructor is being walked, and nothing above it. An `inner class` — legal Kotlin, and the one nested form that *does* capture the outer class's type parameters — gets an empty shadow set. Run against the **shipped** binary, on this branch:

```
=== inner-class ctor param, OUTER typeparam ===
package p
class T
class Outer<T> { inner class Inner(val x: T) }
  EDGE Inner.x -> scope:component:class:kotlin:F.kt:T        <-- WRONG

=== inner-class body property, OUTER typeparam ===
package p
class Key
class Cache<Key> { inner class Entry { val k: Key = q } }
  EDGE Entry.k -> scope:component:class:kotlin:F.kt:Key      <-- WRONG
```

`Inner.x` is typed by `Outer`'s type parameter. The edge points at `class T`, which is not its declared type — and it **binds**, so it never reaches bug-extractor. This is Java's known-over-fire, present here, in the arm whose comments say it is refused rather than documented. `class Cache<Key>` beside `class Key` is not an exotic spelling.

Per the standing rule, "it does not occur in ktor/exposed" is not a reason to accept the rule; and the `91/91 bound` figure cannot see this, because a wrong edge that binds counts as bound.

Two sub-claims in §11 are therefore also wrong as written: the honest-limits list names three pre-existing extractor ceilings and does not name this one, and §6's audit is presented as an enumeration of everything that is not a declaration.

### F2 (blocking, same guard) — the shadow filter is graded on ONE of the two anchors; the other half is ALIVE and production-reachable

`TypeParameterIsNeverATargetEvenWhenShadowed` uses `class Holder<T>(val item: T, val real: Order)` — **the primary-constructor anchor only**. The body-property anchor (`class Holder<T> { val item: T }`) is nowhere in the suite.

Scored, line-targeted so exactly one occurrence changes, `go vet`-clean, `cp`-restored, shasum-verified:

| mutant | edit | landed | verdict |
|---|---|---|---|
| **MR-1b** | `kotlin.go:272` (class-arm body property) `buildProperty(ch, file, rec.Name, classTypeParams)` → `…, nil)` | 1/1 | **ALIVE** |
| MR-2 | `emitPrimaryConstructorFields(…, classTypeParams)` → `…, nil)` | 1/1 | DEAD — `TypeParameterIsNeverATargetEvenWhenShadowed` |

MR-1b's distinguishing input is `class T` + `class Holder<T> { val item: T }`, which is ordinary Kotlin and which the shipped code correctly refuses (I confirmed both sides: shipped → no edge, mutant → edge). M9 in §9's table reports "no type-parameter filter — DEAD", but M9 removes *both* halves and only one is graded. This is exactly §8's own stated failure mode — the fixture varies "a type parameter is present" while holding "which anchor" constant — applied to the guard §8 was written to protect.

### F3 (ALIVE, not production-reachable — reported for the record, not blocking)

| mutant | edit | verdict | why |
|---|---|---|---|
| MR-3 | anchor guard drops its Kind half: `r.Kind != "SCOPE.Schema" \|\| r.Subtype != "field"` → `r.Subtype != "field"` | **ALIVE** | *not production-reachable* — the only two `Subtype: "field"` sites in the package (`kotlin.go:972`, `:1036`) both carry `SCOPE.Schema`. `Unit_OnlySchemaFieldsAreAnchors`'s two refusals are `SCOPE.Operation/function` and `SCOPE.Schema/type_alias`; neither has Subtype `field`, so the compound is graded jointly, not part by part. Distinguishing input: a record with Kind ≠ `SCOPE.Schema` and Subtype `field`. |
| MR-4 | alias branch drops its Subtype half: `r.Kind == "SCOPE.Schema" && r.Subtype == "type_alias"` → `r.Kind == "SCOPE.Schema"` | **ALIVE** | *not production-reachable* — this is the mutually-masking pair the source comment itself names ("SCOPE.Schema/field — refused by subtype; **also** unreachable by name shape"). `FieldIsNeverATarget` only exercises the name-shape half. Distinguishing input: a bare-named `SCOPE.Schema/field` record; no Kotlin producer emits one (both call sites qualify with the owner). |
| MR-5 | `kotlinComponentAddressFamily` drops the bare `"Component"`, `"View"`, `"Model"` entries | **ALIVE** | *not production-reachable* — the Kotlin extractor emits no bare-kinded records. The source comment says "the eight entries below are **exactly** that set"; three of the eight are ungraded. I re-derived the closure by hand and the set **is** correct (`componentKindFamily` = 7 entries without `SCOPE.Class`; closure under the trim alias adds it → 8). |

### Mutants I ran that came back DEAD (re-scores / new)

|---|---|---|
| MR-6 (= §9's M14) family drops `SCOPE.Class` | DEAD | `Unit_ScopeClassParticipatesViaTheTrimAlias` |
| MR-7 Component subtype allow-list widened with `"import"` | DEAD | `ImportCarrierIsNeverATarget` |
| MR-2 ctor-anchor type-param filter | DEAD | `TypeParameterIsNeverATargetEvenWhenShadowed` |

### Claims I verified at source rather than taking on faith

- **§3, the two tiers.** `BuildComponentStructuralRef` yields `scope:component:…` for *both* target kinds → `structuralKindFamilies("component")` → `componentKindFamily` (refs.go:2341, 7 entries, `SCOPE.Service` deliberately absent per :2429). `lookupLocationKind` (refs.go:3420) runs at refs.go:2982 and **returns before** the `ambigLocation` probe at refs.go:3004. Tier 1 (`.real`, `includePlaceholders=false`) skips every `SCOPE.*` family member, so a pure-Kotlin file always lands on tier 2 (`.base`, dual-indexed raw + trimmed). A `SCOPE.Schema/type_alias` matches no family key at either tier → falls to `ambigLocation`, which is kind-agnostic. **The split is real and the direction of each count is right.** Arm rust's records-vs-kinds defect is not repeated: `graph.EntityID` excludes Subtype, and `locKindEnt.base.set` only blanks on a *different* ID, so two same-Kind records never blank each other.
- **§5, the conjunct count.** Exactly two `r.SourceFile != filePath` guards (field_type_refs.go:365 target scan, :383 ambiguity scan) plus the kind/subtype anchor guard in `attachKotlinFieldTypeRefs`. Three, as claimed — no fourth. (Nit: the anchor guard is not a *same-file* conjunct at all; same-file there is a consequence of `Extract` being per-file, not of a check. The table's framing is looser than the code.)
- **§6, companion objects.** Probed: `class Order { companion object Factory }` mints **no** entity, and `val f: Factory` produces no edge. `interface Repo { companion object { val o: Order } }` mints no field. Premise holds.
- **§11, the rival kinds.** Probed on the real extractor: `enum class Status` → `SCOPE.Enum/enum Status` + `SCOPE.Component/enum Status`, edge emitted; `object Pages { const val ×2 }` → const-group value-set present, edge emitted; `@Service class Billing` → `SCOPE.Service Billing`, edge emitted; `typealias Handler` + `fun Handler()` → correctly refused; root-package `import Order` → correctly refused. All four hold.

### Two smaller things, non-blocking

- **§8's target-kind axis double-counts.** `annotation class Tgt` is extracted as `SCOPE.Component` **Subtype `class`** — byte-identical to the plain `class` row. The `TargetKindSpace` row labelled "annotation class" varies nothing the "class" row did not already vary. Same for `expect class` / `external class` / `sealed class` / `value class` / `abstract class` — I probed all of them and every one is `SCOPE.Component/class`. The list reads as six distinct subtypes; there is one. (Label vs content.)
- **`fun interface` is not in the audit and is not in the limits list.** `fun interface Cb` is extracted as `SCOPE.Operation/function Cb`, so `val c: Cb` produces no edge — and, more interestingly, it is a *collider* that would make a same-named typealias ambiguous. A pre-existing extractor defect, not introduced here, but §6 presents its enumeration as exhaustive and §11 presents its ceiling list as complete. Local classes (`fun f() { class Local(val o: Order) }`) mint no entity either, also unlisted.
- `attachKotlinFieldTypeRefs` is inserted **inside** Track B's comment block in `Extract`, so that comment now reads as documentation of the #6912 call rather than of `resolveImportToIDs`. Cosmetic. (The "runs LAST of the entity-producing passes" claim itself is true — `resolveImportToIDs` produces no entities.)

### Gates I ran myself

| gate | result |
|---|---|
| `go vet ./internal/extractors/kotlin/ ./cmd/grafel/` | exit 0 |
| `gofmt -l internal cmd` | empty |
| `go test -count=1 ./internal/extractors/kotlin/` | ok 0.636s |
| `go test -count=1 ./internal/resolve/` | ok 1.194s |
| `GRAFEL_HOME=$(mktemp -d) GRAFEL_DAEMON_ROOT=$(mktemp -d) go test -count=1 ./cmd/grafel/` | **ok 184.4s**, exit 0 — digest unmoved, and §10's reason for that ("`writeSpanFixture6118` writes no `.kt`") is the right reason, so this leg is a compile gate here and not a behavioural one |

---

**Verdict: REQUEST-CHANGES**, on F1 and F2 only. Everything else above is recorded, not demanded.

F1 is a wrong edge that binds, in the shape the PR's own comments and the PR body both say Kotlin refuses. Either the code covers enclosing declarations, or both prose claims get downgraded to a stated limit in §11 — but the current pairing (unqualified claim + narrower code) is the defect class this milestone has been losing to. F2 needs the body-property anchor graded whichever way F1 lands, since MR-1b is ALIVE on ordinary Kotlin today.

I am naming what I could not observe being defended; **I have not implemented or verified any fix, and prescribing one is not reviewing it** — the shape of the correction is the author's call, and whatever lands needs its own mutant score.

cajasmota:
## REQUEST-CHANGES addressed — `948c08dfc` + `d3382b2ef`

Both blocking findings are fixed, and one thing in the review itself turned out not to be what the extractor does. Everything below replaces the corresponding section of the PR body.

---

### F1 — fixed by REFUSING, not by pinning

`kotlinVisibleTypeParameterNames` replaces the immediate `type_parameters` read. It seeds with the declaration's OWN list and then ascends the enclosing-declaration chain **while the current declaration is `inner`**, adding each enclosing CLASS's list.

The reviewer's two repros now refuse, and the `91/91 bound` figure could not have seen either:

```
class Key;  class Cache<Key> { inner class Entry { val k: Key = q } }   -> no edge (was: -> Key)
class T;    class Outer<T>   { inner class Inner(val x: T) }            -> no edge (was: -> T)
```

A pin was not needed, so it was not used — and #7041's two-way-`Skipf` shape is not reproduced here.

**The over-refusal direction is where the work went.** `TestKotlinFieldTypeRefs_NestingFormSpace` enumerates the nesting-form space and grades every row in BOTH directions, each fixture carrying `val real: Order` as a positive control so an absence is never a silent no-op:

| nesting form | captures? | expected | mutant that gets it wrong |
|---|---|---|---|
| `inner class` in a generic outer — **ctor anchor** | yes | edge REFUSED | MR-2 |
| `inner class` in a generic outer — **body anchor** | yes | edge REFUSED | **MR-1b** |
| `inner class` inside an `inner class`, generic grandparent | yes (2 hops) | edge REFUSED | N3 |
| `inner class` with its OWN shadowing list (`B<T, W>` in `A<T>`) | yes | edge REFUSED | N4 |
| `inner class` in a **non-generic** outer | no | **edge KEPT** | N1 |
| **plain nested** `class` in a generic class | no | **edge KEPT** | N1 |
| **plain nested** `class` in a generic **interface** | no | **edge KEPT** | N1 |
| `object` declaration inside a generic class | no | **edge KEPT** | — |
| `inner class` inside an **`object`** inside a generic class | no | **edge KEPT** | N2 |
| `inner class` inside a **plain nested** class inside a generic class | no | **edge KEPT** | N1 |
| local class in a generic `fun` | — | mints no entity; no edge either way | — |

The last two "KEPT" rows are the ones an ascent to the *nearest enclosing CLASS* — rather than the *nearest enclosing DECLARATION* — gets wrong, silently deleting a correct edge. That is arm cpp's failure shape (#7057) and it is why the table exists rather than three fixtures.

### F2 — the two anchors are now graded disjointly

`TypeParameterIsNeverATargetFromABodyProperty` grades the body-property call site on its own, holding the shadow name, the rival and the control constant and varying only the anchor. Scored both ways, line-targeted, `go vet`-clean, `cp`-restored, shasum-verified, exact-match-count asserted:

| mutant | edit | landed | kills | verdict |
|---|---|---|---|---|
| **MR-1b** | class-body arm `buildProperty(…, classTypeParams)` → `…, nil` | 1/1 | `…FromABodyProperty`, NestingFormSpace/body-anchor | **DEAD** |
| MR-2 | `emitPrimaryConstructorFields(…, classTypeParams)` → `…, nil` | 1/1 | `…EvenWhenShadowed`, 3 ctor-anchor rows | **DEAD** |

The two kill sets are disjoint — neither anchor's mutant touches the other's test.

### New mutants on the new code (permissive direction first)

| mutant | edit | landed | verdict |
|---|---|---|---|
| N1 | `kotlinDeclarationIsInner` returns true for every `class_declaration` (over-refuse) | 1/1 | **DEAD** — 3 "KEPT" rows |
| N2 | boundary set reduced to `class_declaration` (ascend to nearest CLASS, jumping over an object) | 1/1 | **DEAD** — inner-inside-object row |
| N3 | ascend at most one level | 1/1 | **DEAD** — inner-inside-inner row |
| N4 | ascent REPLACES the own list instead of unioning | 1/1 | **DEAD** — own-shadowing row (`w: W`) |
| N5 | drop the `encl.Type() != "class_declaration"` half of the stop | 1/1 | **ALIVE — equivalent for valid Kotlin.** No `object` / `companion_object` can carry `type_parameters`, so "add nothing, then stop because an object is never `inner`" is the same computation as "stop". The only distinguishing input is a `class_declaration` carrying `inner` whose nearest enclosing declaration is `fun <Q> g()` — illegal Kotlin (`inner` is rejected on a local class) AND a local class mints no entity, so the shadow set is never consulted for one. The check is kept as a guard against a future widening of the boundary set, not as live behaviour. |
| P1 | `sealed class` gets its own subtype | 1/1 | **DEAD** — TargetKindSpace's new per-row kind assertion |
| P2 | admit `SCOPE.Operation` as a target kind | 1/1 | **DEAD** — TargetKindSpace/top-level-function, alias-collision test |
| MR-5 | drop bare `Component` / `View` / `Model` from the address family | 1/1 | **DEAD** (was ALIVE) — see below |

### §5 — corrected conjunct table

The review's nit is right and the old framing was wrong. There are **two** same-file conjuncts, not three:

| # | guard | what it is |
|---|---|---|
| 1 | `field_type_refs.go` target scan — `r.SourceFile != filePath` | a genuine same-file check |
| 2 | `field_type_refs.go` ambiguity scan — `r.SourceFile != filePath` | a genuine same-file check |
| — | `attachKotlinFieldTypeRefs`'s `r.Kind != "SCOPE.Schema" \|\| r.Subtype != "field"` | **an ANCHOR guard, not a same-file check.** Same-file holds there as a consequence of `Extract` being per-file, never because anything tests it. |

### §6 / §11 — the limits, as they now actually stand

Added to the honest-limits list, both **recall** floors (a missing edge, never a wrong one), both pre-existing and neither introduced nor fixed here:

- **`fun interface Cb` mints NO entity for the interface itself.** I probed this rather than relaying it: the review described it as a `SCOPE.Operation` named `Cb` and therefore a collider for a same-named `typealias`, and **neither half is what the extractor does** — nothing named `Cb` is emitted at all, so it cannot collide with anything. Only its MEMBERS surface, and they surface BARE (`go`, not `Cb.go`) because no parentType is in scope for them. `val c: Cb` produces no edge. (Separately: the *body-less* form `fun interface Cb` derails the parse and swallows the rest of the file. Out of scope here, noted for whoever fixes the walk.)
- **A local class mints no entity** (`fun f() { class Local(val o: Order) }`) — neither a target nor a source, so the enclosing generic function's type parameters are never a shadow question.

Both are graded by `UnmintedDeclarationForms`, which asserts the ABSENCE by name plus a floor on the member actually surfacing, so it cannot go vacuous if the fixture stops parsing.

Added to §11 as a **stated behaviour**, not a limit: the shadow set is the one VISIBLE at the declaration, and the nesting-form table above is the statement of where that stops.

### §8 — the target-kind axis, relabelled AND made observable

The review is right that six labels were one axis. `TargetKindSpace` now carries a per-row `wantRecords` column naming the Kind/Subtype the extractor **actually** mints, asserted before the edge assertion:

- Genuinely distinct extracted subtypes: `class`, `data_class`, `interface`, `object`, `enum`, `type_alias` — **six**, not eleven.
- Source-spelling-only rows, relabelled `… (extracted as class)`: `annotation` / `expect` / `external` / `sealed` / `value` / `abstract`, all `SCOPE.Component/class`.
- The `enum class` row is the one that mints **two** records (`SCOPE.Enum/enum` then `SCOPE.Component/enum`) and the column now says so — which is the component-family half of the ambiguity rule, observed rather than described.

Mutant P1 (a distinct `sealed_class` subtype) dies on that column, so the labels are now pinned to content.

### MR-3 / MR-4 — alive, not production-reachable, recorded not manufactured

| mutant | verdict | distinguishing input |
|---|---|---|
| MR-3 | ALIVE, not production-reachable | a record with `Kind != "SCOPE.Schema"` **and** `Subtype == "field"`. Both `Subtype: "field"` producers in the package carry `SCOPE.Schema`; `Unit_OnlySchemaFieldsAreAnchors`'s two refusals differ on Subtype, so the compound is graded jointly. |
| MR-4 | ALIVE, not production-reachable | a **bare-named** `SCOPE.Schema/field` record. Both call sites qualify with the owner, so the name-shape half masks the subtype half — the mutually-masking pair the source comment already names. |

No fixture was manufactured for either.

### MR-5 — was ALIVE, now DEAD

The map's comment called its eight entries "exactly" the trim-alias closure while three (bare `Component` / `View` / `Model`) were ungraded. They are unreachable from Kotlin SOURCE but perfectly reachable through the synthetic-record harness `field_type_refs_units_6912_test.go` exists for, and grading them there is that file's whole design. `Unit_EveryAddressFamilyEntryMakesARefAmbiguous` drives one rival record per entry, asserts the target is blanked, and carries a **non-family** rival (`SCOPE.Enum`) as the table's positive control so no row can pass on a pass that emitted nothing.

`SCOPE.Component` is the one entry absent from that table, and its absence is a consequence of the rule: a same-named `SCOPE.Component` record is the target's OWN graph node and must NOT blank anything — graded in the opposite direction by `Unit_DuplicateComponentRecordsAreOneNode`. Said out loud rather than left as a silent omission.

Still unchecked, and now stated in the source rather than implied away: **nothing asserts this duplicate and `resolve`'s `componentKindFamily` stay in sync.** That is the price of the deliberate non-import.

### Corpus re-measurement

Re-measured on 447 `.kt` files (ktor 195, ktor-samples 151, exposed 101), same harness, before and after:

| | files | field entities | edges | bound | dangling |
|---|---|---|---|---|---|
| shipped `7c667974e` | 447 | 1490 | 91 | 91 | 0 |
| **`d3382b2ef`** | 447 | 1490 | **91** | **91** | **0** |

**Nothing moved, and the full edge SET is byte-identical** (`diff` of the sorted `file\|field\|toID` list: 0 lines). No rows changed, so there are none to name. ktor / ktor-samples / exposed declare no `inner class` whose field is typed by a captured parameter — which is exactly why a corpus count could never have been the instrument for F1, and why the review had to find it from source.

### Gates

| gate | result |
|---|---|
| `gofmt -l internal cmd` | empty |
| `go vet ./internal/extractors/kotlin/ ./cmd/grafel/` | exit 0 |
| `go test -count=1 ./internal/extractors/kotlin/` | ok 0.563s |
| `go test -count=1 ./internal/resolve/` | ok 1.146s |
| `GRAFEL_HOME=$(mktemp -d) GRAFEL_DAEMON_ROOT=$(mktemp -d) go test -count=1 ./cmd/grafel/` | **ok 193.4s**, exit 0 — digest unmoved (no `.kt` in the span fixture, so this stays a compile gate here) |

cajasmota:
## CK-1 addressed — `0789c6423` (test-only)

Confirmed the finding, reproduced it, and it is exactly as described: the code is correct and the suite did not defend it. Over-refusal direction, production-reachable, no symptom when it breaks.

**Why the table missed it, restated because it is the generalisable part:** every row varied the nesting RELATIONSHIP and the GENERICITY of the enclosing declaration, and every row held the nested declaration's own modifier constant at bare `class` or `inner class`. `inner` is one member of a node type that has siblings, and not one sibling appeared nested inside a generic class anywhere in the file.

### Rows added

All in the **KEPT** direction with the `real: Order` positive control, except the last:

| row | direction | kills CK-1 |
|---|---|---|
| `data class M(val y: T, val real: Order)` in `class A<T>` | KEPT | **yes** |
| `sealed class M(…)` in `class A<T>` | KEPT | **yes** |
| `annotation class M(…)` in `class A<T>` | KEPT | **yes** |
| `value class M(val y: T) { val real: Order get() = z }` in `class A<T>` | KEPT | **yes** |
| `enum class M { A, B; val y: T …; val real: Order … }` in `class A<T>` | KEPT | no — see below |
| `sealed class A<T> { inner class N(val y: T, val real: Order) }` | REFUSED | no — see item 4 |

**I enumerated rather than took `data` alone.** A hand-picked representative is precisely how this axis came to be held constant in the first place, and each row costs four lines. Probed against the grammar: the non-`inner` `class_modifier` members a nested declaration can carry are exactly **`data`, `sealed`, `annotation`, `value`** — all four are rows, and all four kill CK-1.

**One correction to my own first draft, and to the framing in your message.** I wrote `enum` into the class_modifier list and then probed it: **`enum` is NOT a `class_modifier`.** `enum class` puts a bare `enum` node directly under `class_declaration`, not under `modifiers` — which is why `classDeclarationSubtype` reads it as a direct child. So the enum row sits **outside CK-1's blast radius and does not kill it**, and the comment now says so rather than implying a fifth kill. It stays for a different reason, stated: it is the only nested row whose target mints **two** records, so it shows the component-family half of the ambiguity rule holding up under a nested declaration.

**Checked, not assumed, before writing rows** (your "do not manufacture rows for forms that mint no entity"):
- `abstract` / `open` are an `inheritance_modifier`, `private` is a `visibility_modifier` — different CST node types, outside the blast radius, and nothing in this pass reads them. **No rows.**
- A bare `enum class M { A, B }` mints **no field at all**, and a `value class` takes exactly **one** constructor property. Written the way the other rows are, both would have asserted nothing — a vacuous no-op row apiece. Both use body properties instead, which moves the anchor, and the table says so.

### 4. The enclosing declaration's modifier — checked, and it is NOT the same blind spot

`sealed class A<T> { inner class N(val y: T) }` still CAPTURES and still refuses — verified. But it is not the same gap: **the code never reads the enclosing declaration's modifier.** A `sealed` / `abstract` / `data` outer is still a `class_declaration` carrying a `type_parameters` list, so no mutant distinguishes that row. I added it anyway as one row, because "the outer is always a bare `class`" was on the same held-constant list that hid CK-1, and I say in the comment that it grades nothing — an ungraded row that claims to grade something is the defect this PR keeps being about.

### Updated mutant table row

| mutant | edit | landed | verdict |
|---|---|---|---|
| **CK-1** | `kotlinDeclarationIsInner`: `cm.Child(k).Type() == "inner"` → `!= "___never___"` (treat any `class_modifier` as `inner`) | 1/1, vet-clean | **DEAD** — `NestingFormSpace/{data,sealed,annotation,value}_class_nested_in_a_generic_outer` |

Re-scored **verbatim** on the final tree, byte-level `cp` backup, shasum-verified restore, `git diff HEAD` empty after.

### VARIED / HELD CONSTANT block

Rewritten. It now names the modifier axis explicitly, records which node types are outside the blast radius and why, and flags the two fixtures whose form forced the anchor to move. The old block listing only the held-constant axes it happened to think of is what made this invisible on reading.

### Gates (re-run)

| gate | result |
|---|---|
| `gofmt -l internal cmd` | empty |
| `go vet ./internal/extractors/kotlin/ ./cmd/grafel/` | exit 0 |
| `go test -count=1 ./internal/extractors/kotlin/` | ok 0.654s |
| `go test -count=1 ./internal/resolve/` | ok 1.159s |
| `GRAFEL_HOME=$(mktemp -d) GRAFEL_DAEMON_ROOT=$(mktemp -d) go test -count=1 ./cmd/grafel/` | **ok 180.7s**, exit 0 — digest unmoved, as expected for a test-only change |

Corpus not re-measured: the diff touches one `_test.go` file and no production byte, so the 447 files / 91 edges / 91 bound / 0 dangling figure from `d3382b2ef` stands unchanged.

PR #7064, opened 2026-09-16
fix(#6912): arm K — scala fields recorded no declared type at all; capture it, then bind it

cajasmota:
Arm K of #6912, and the **last** one. Scala fields now carry a `REFERENCES` edge with `ref_kind=field_target_type` to the same-file declaration of their declared type. Ruby stays scoped out — `attr_accessor :buyer` names no type.

Ten arms shipped before this: C# (#6984) · proto (#6991) · rust (#7000) · go (#7036) · java (#7039) · F# (#7040) · php (#7042) · swift (#7043) · C/C++ (#7057) · kotlin (#7063, in review).

---

## 1. The premise, probed rather than inherited

#6912 asserted that java "retains NO type string at all" and **that was false** — arm F found the type already a local at `java.go:425`. So Scala's absence was re-established before anything was built.

A 10-shape file was extracted and **every record printed**:

```
SCOPE.Schema   field   Svc.a    sig="" props=map[] meta=map[]
SCOPE.Schema   field   Svc.b    sig="" props=map[] meta=map[]
...  (9 more, all identical)
```

**No `Properties`, no `Signature`, no type datum of any kind** — at both anchors (`buildScalaField`, `emitScalaCaseClassFields`). Scala is a genuine **capture-then-bind** arm, and the capture ships WITH the bind: a capture alone would be one more unhashed property that nothing reads, which is the state #6912 exists to fix.

`TestScalaFieldTypeRefs_PremiseFieldRecordsCarryNoTypeDatum` pins it, so a future producer that starts writing a `field_type` property has to come past this test and decide whether this arm should read it.

**Arm B's question, asked rather than assumed** — *is the FIELD the anchor, or is something else?* Scala has a container-level type map (`collectContainerFieldTypes`, `scala.go:270`) but it is a **call-receiver resolution table**, not an edge producer: it feeds `extractCallRelationships` so `repo.find()` emits `Repo.find`. No type edge exists at any anchor. Unlike go (`DEPENDS_ON` on the struct) and proto (message-anchored), there is nothing here to avoid re-anchoring.

## 2. Where the type lives in the grammar, and what sits beside it

Probed against the real grammar, both anchors:

```
val_definition   [modifiers] [annotation] val identifier ":" TYPE ["=" INITIALIZER]
var_definition   [modifiers] [annotation] var identifier ":" TYPE ["=" INITIALIZER]
class_parameter  [modifiers] [val|var]    identifier ":" TYPE ["=" DEFAULT]
```

**The neighbour that must not be read is the INITIALIZER, and it is inside the same node:**

```
val c = new Repo()
  val_definition [val, identifier c, "=",
                  instance_expression[new, type_identifier "Repo", arguments]]
```

A capture that walked the whole `val_definition` emits `c -> Repo` from an **inferred** type: right name, right file, admissible kind, admissible subtype — and #7056 established no instrument we own can see that it is wrong. The capture is scoped to the window **after `:` and before `=`**, which excludes it structurally, along with a class parameter's default value and a leading modifier list.

### The type-expression shape space, ENUMERATED

Arm I's template-parameter refusal was wrong in **both** directions and only enumeration found it. Every row below was printed from a real parse and is a row in `TestScalaFieldTypeRefs_TypeExpressionShapeSpace`:

| written | CST | candidates |
|---|---|---|
| `Repo` | `type_identifier` | Repo |
| `Option[Repo]` | `generic_type[tid, type_arguments]` | Option, Repo |
| `List[List[Repo]]` | nested `generic_type` | List, List, Repo |
| `List[_]` | `type_arguments` holds `wildcard` | List |
| `Repo with Named` | `compound_type` | Repo, Named |
| `Repo & Named` | `infix_type[tid, operator_identifier, tid]` | Repo, Named |
| `(Repo, Named)` | `tuple_type` | Repo, Named |
| `Repo => Named` | `function_type[parameter_types, tid]` | Repo, Named |
| `Repo @unchecked` | `annotated_type[tid, annotation[tid]]` | **Repo only** |
| `com.acme.Repo` | `stable_type_identifier` (DOTTED) | — |
| `_root_.com.acme.Repo` | `stable_type_identifier` | — |
| `Registry.Inner` | `stable_type_identifier` | — |
| `Registry.type` | `singleton_type[identifier, ".", type]` | — |
| `T` (a type parameter) | `type_identifier` | — |
| `val c = new Repo()` | no colon at all | — |

**The annotation trap is Scala's own**, and it is the over-collection direction: an `annotated_type` holds the annotation's name as a `type_identifier` **inside the declared type's own subtree**, and a Scala annotation *is* a class. `r: Repo @Inject` beside `class Inject` would otherwise be a fully-binding, fully-wrong edge. Graded with the carrier class declared in the fixture so a leak becomes a visible edge rather than a silent drop.

**Dotted is skipped WHOLE**, not reduced to its last segment — arm D found Go's `resolveTypeReferences` stripping `other.Order` to `Order` and binding it wrongly. Scala is worse than Go here: `Registry.Inner` is a **type projection** whose qualifier is *itself* a same-file declaration, so both segments name something in this file and any reduction picks one of two wrong answers.

**Over-refusal is graded too.** `TestScalaFieldTypeRefs_ShapeSpaceHasAPositiveControlPerRefusal` pairs each refusal that names a same-file type with a row differing ONLY in the refused dimension that DOES emit — otherwise a refusal row is indistinguishable from a degraded parse.

## 3. The resolver tier table — DRIVEN, not read

| question | answer | where |
|---|---|---|
| Which tier resolves my address? | `Index.lookupStructural` → `lookupLocationKind(file, name, structuralKindFamilies("component"))`, `refs.go:2982` | returns FIRST |
| Is a second tier ever reached? | **No.** `ambigLocation` (`refs.go:3004`) is never consulted | one target kind only |
| Which kinds does it weigh? | `componentKindFamily` (`refs.go:2341`) closed under the index's trim alias — 8 entries | `scalaComponentAddressFamily` |
| Kinds it filters out | everything else, incl. `SCOPE.Enum`, `SCOPE.Operation`, `SCOPE.Service` | |

**Scala needs exactly ONE tier, and the reason is a fact about the extractor, not the resolver.** Arms G and J needed both because they had a `SCOPE.Schema/type_alias` target kind. Scala's `type Alias = X` parses as `type_definition`, `walkNode` has no case for it, and **it mints no entity at all** — so there is no second target kind. That premise is asserted in `TestScalaFieldTypeRefs_TheOnlyTierIsTheComponentFamilyTier`, which fails loudly if a `type_definition` arm is ever added.

Every claim above was driven through `resolve.BuildIndex` → `resolve.ReferencesEmbedded`:

- `..._ResolverBindsEveryEdge` — the standing gate: 5 edges, 5 bound, 0 dangling.
- `..._OutOfFamilyRivalsDoNotDangle` — **the refusal arms C/D/E justified as "emitting would guarantee a dangling stub" is FALSE here.** Hand-emitting the edge arm D's rule would refuse **binds**. Three rivals, each with its existence asserted first.
- `..._InFamilyRivalWouldActuallyDangle` — an in-family rival (`SCOPE.Model`) genuinely dangles, an out-of-family one (`SCOPE.Enum`) does not, and the no-rival control binds. Three rows, the rival held constant except for its family membership.

## 4. The ambiguity rule, re-derived — and why it differs from its neighbours

> **Count the KINDS the resolver tier that resolves YOUR address actually weighs.**

For Scala that is **family kinds only** — arm F's answer, reached from Scala's own tier rather than inherited. Two Scala facts make the alternatives wrong, and they are wrong in *different* ways:

**(a) Arm D's all-kinds rule is a pure recall loss.** Scala mints three same-file, same-name rivals OUTSIDE the family:

| construct | rival |
|---|---|
| `sealed trait Shape` + `case object Circle extends Shape` | `SCOPE.Enum` value-set **named Shape** (`constantset.go` sealed-trait arm) |
| `object Status { val A = …; val B = … }` | `SCOPE.Enum` const-group value-set **named after the object** |
| `class Order` + `def Order(): Int` (a factory) | `SCOPE.Operation` named Order — legal and idiomatic Scala |

None can blank a component-space ref. Refusing on them costs edges and prevents nothing.

**(b) Arm C's record count is wrong on the single most idiomatic shape in the language.** `case class Order(…)` beside its **companion `object Order`** is two `SCOPE.Component` records with the same Name in the same file — and **ONE graph node**, because `graph.EntityID` hashes `(repo, Kind, Name, SourceFile)` with **Subtype excluded**. This is the Scala mirror of arm I's constructor (`class Order { Order(); }`), and it is *more* common: every companion object in the language. Counting records deletes a correct edge, which is #7038 in a new place.

The unit of the count is the **KIND**. Graded both ways: `..._CompanionObjectIsOneNodeNotAnAmbiguity` (asserts the two-record premise first), `..._CompanionPairResolvesToOneNode` (drives the resolver and checks the edge landed on the merged id), `..._Unit_SecondRecordOfTheSameKindIsOneNode`.

### Measured counterfactuals — with the dropped rows NAMED

On the 5-file probe (47 field entities), each rule applied ONE AT A TIME to the shipped code and re-measured:

| rule | edges | Δ | wrong bindings |
|---|---|---|---|
| **shipped** (family kinds + subtype allow-list) | **26** | — | **0** |
| arm D — ambiguity counts every kind | 10 | **−16** | 0 |
| arm C — ambiguity counts RECORDS | 11 | **−15** | 0 |
| arm C — `Kind != "SCOPE.Component"` (no subtype allow-list) | 28 | **+2** | **2** |
| no allow-list and no ambiguity guard | 28 | +2 | 2 |
| arm F — family kinds | 26 | 0 | 0 |

**Arm F's row is 0 because for Scala it IS the shipped rule** — stated rather than presented as independent validation.

**Arm D's 16 dropped rows** (every one has an out-of-family same-name rival): `Billing.{decorated,fallback,history,latest,pair}→Invoice`, `Line.unit→Money`, `Money.Zero→Money`, `Order.total→Money`, `InMemoryOrders.{index,last}→Order`, `Order.Empty→Order`, `OrderController.order→Order`, `Order.status→Status`, `Carriers.all→Carrier`, `Shipment.{alternates,carrier}→Carrier`.

**Arm C's 15 dropped rows are a DIFFERENT set** — every companion-object pairing (Invoice ×5, Money ×3, Order ×4, Carrier ×3) — and it **keeps** `Order.status→Status`, which arm D drops. The two rules are independent, not two strengths of one rule.

**The +2 rows are the ones that matter**, because a missing edge has no symptom and a wrong one has none either: dropping the subtype allow-list makes `Report.order→Order` and `Report.amount→Money` in `Reporting.scala` bind to **import placeholders**, flagged by the probe harness's declaration-span audit (`NOKEYWORD 2`).

## 5. The not-a-declaration audit

A 100% bind rate is not evidence — #7056 established that no instrument we own can see an edge that binds to a non-declaration. Every same-file, same-name carrier Scala can mint, each with its **existence asserted so the refusal cannot pass vacuously**:

| carrier | reachable? | refused by | graded |
|---|---|---|---|
| `SCOPE.Component` + `""` — an **import placeholder** | **YES, and more easily than any prior arm** | subtype allow-list | `..._ImportPlaceholderIsNeverATarget` + `..._RefusedImportPlaceholderEdgeWOULDBIND` |
| `SCOPE.Component/object` | yes (a term, not a type) | subtype allow-list | `..._ObjectIsNeverATarget` + `..._RefusedObjectEdgeWOULDBIND` |
| `SCOPE.Component/file` (#577), `/twirl` (#501) | name is the file path | subtype allow-list | `..._FileCarrierIsNeverATarget` |
| `SCOPE.Enum` value-sets | yes | Kind check | `..._Unit_TargetKindMustBeComponent` |
| `SCOPE.Operation` | yes | Kind check | same |
| `SCOPE.ExceptionType` | **no** — sentinel `SourceFile`, not this file's path | same-file conjunct | stated at that strength |
| `SCOPE.Schema/field` | no — field Names are dotted | Kind check | claimed as nothing more |

**The import placeholder is the load-bearing one, and Scala reaches it more easily than Go or F# did.** `buildImports` names the placeholder after the **FIRST dotted segment** of the path, so an ordinary sibling-package import — `import Order.Status`, idiomatic for pulling a member off a companion — mints a bare-named `SCOPE.Component` called `Order` in this file. No root-package import is required.

And the refused edge **does not dangle — it BINDS**, confidently and wrongly, which is the F# shape (arm E) rather than the Go one. Proved by hand-emitting it through the real resolver and checking the id it landed on. `bug-extractor` classifies unresolved stubs, so nothing we own could ever have surfaced it.

**The mirror is graded too**: `..._ImportPlaceholderBesideARealClassStillBinds` — the refusal must not cost the edge when a real declaration of that name is also present (over-refusal, no symptom).

**`object` is refused because an object declares a TERM, not a type** — `val r: Registry` does not compile for `object Registry`; only `Registry.type` does, and `singleton_type` is refused by the scanner. **That fixture is deliberately not legal Scala and the test says so out loud** rather than claiming it was grammar-checked (arm H's send-back). There is no legal input that distinguishes the refusal; the legal reachable shape is the companion, where the refusal costs nothing because the class supplies the identical ToID. Measured at **zero delta on the probe**.

### The SOURCE endpoint, audited too

Arm I found 68 of 254 "fields" were member *functions*. Scala's field records come from exactly two producers, both genuine value members. A `def` is a `SCOPE.Operation` and can never be an anchor (`..._AbstractValMintsNoFieldSoItIsNoAnchor` asserts both halves).

### Recall floors, pinned as tests rather than described

- **Scala 3 `enum Color`** mints a `SCOPE.Enum` value-set and **no `SCOPE.Component`** — `walkNode` has no `enum_definition` case — so it can never be a target. Pinned; fails loudly if that case is added.
- **Abstract `val abs: Repo`** in a trait parses as `val_declaration`, which `emitContainerWithMembers` does not intercept — no field entity exists to anchor on.
- **Second parameter list** (`class Multi(a: Repo)(implicit b: Named)`) — `emitScalaCaseClassFields` breaks after the first `class_parameters` block.
- **Type parameters are scoped to the immediately enclosing declaration** — `walkNode` carries no lexical stack, so a field of a NESTED class typed by the OUTER class's `T` keeps that shadow.
- **`Registry.type` and the function-type receiver** are recall floors, stated as such, not a claim that they name nothing.

## 6. The three same-file conjuncts

Every table on this issue said two; php's author found three. Scala has three, each failing in a **different direction**, each graded on its own — grading one is what makes its sibling look covered.

| # | site | failure direction | mutant | verdict | killer |
|---|---|---|---|---|---|
| 1 | `scalaInFileTypeTargets` pass 1 (family COUNT) | a cross-file record wrongly **SUPPRESSES** a correct edge — no symptom | M11 | **DEAD** | `..._Unit_ForeignRecordNeverSuppressesAnEdge` |
| 2 | `scalaInFileTypeTargets` pass 2 (TARGET scan) | a cross-file record wrongly **BECOMES** a target — confident wrong binding | M12 | **DEAD** | `..._Unit_ForeignRecordIsNeverATarget` |
| 3 | `attachScalaFieldTypeRefs` field loop (ATTACH ANCHOR) | a field from **another file** receives an edge into this one | M13 | **DEAD** | `..._Unit_ForeignFieldIsNeverAnAnchor` |

All three are unreachable through `Extract` (which sees one file), so they are graded by calling **the call site** with hand-built multi-file slices — not by extracting a predicate and unit-testing it, which does not pin behaviour (#6533).

### Golden fixtures — unmoved, and enumerated rather than asserted

`scala-play-mini` is the only `.scala` golden fixture: **3 files, 8 field entities, 0 edges.** Every declared type (`User`, `Future`, `Map`, `ExecutionContext`) is cross-file or external; `UserController.scala` declares `case class User` but its own fields are typed `ControllerComponents` / `UserService`, both imported. So `baseline.json` and the #6118 digest are unmoved **because the producer does nothing there** — stated as a fact about the fixture, not presented as a safety result. `cmd/grafel` was nonetheless run green in an isolated env (`GRAFEL_HOME` / `GRAFEL_DAEMON_ROOT`), because **the digest is NOT blind to a new `REFERENCES` edge** — relationships are hashed as `R <from> -[Kind]-> <to>`; only properties are not.

`testdata/fixtures/**` (not baseline-pinned) gains **7 edges**, all inspected against source and all correct: `UserService.users→User`, `InMemoryUserRepository.store→User`, `TaskController.taskForm→TaskForm`, `Task.taskFormat→Task`, `TaskForm.reads→TaskForm`, `Report.format→Report`, `ProcessReport.replyTo→Result`. `internal/verify` green.

### Corpus — a ZERO, and it is STRUCTURAL

**20 `.scala` files / 2 repos** (kafka-streams-examples 10, play-scala-starter 10), **29 field entities, 0 edges, 0 dangling.**

That zero is characterised rather than reported: **24 declared-type candidates were captured** (so the capture works), and **all 24 name types declared nowhere in the corpus** — `KStream`, `KTable`, `KafkaStreams`, `Properties`, `Array`, `Byte`, `String`, `Long`, `Instant`, `EmbeddedSingleNodeKafkaCluster`. **Same-file 0, cross-file 0, external 24.** Only **1 of 20 files** declares more than one type.

So unlike arm H (php), where 43 cross-file candidates were the real prize, **this corpus has no Scala prize at either distance** — it is 20 sample files from two demo repos. Do **not** read this zero as a bound on the language, and do **not** count Scala in any "N languages shipped" line unqualified: the mechanism ships, the corpus yield does not exist.

### The substitute, stated as a substitute

A **5-file idiomatic probe** driven through the production resolver: sealed traits with case objects, companion objects, case classes, generic wrappers, tuples, annotated types, qualified types, an inferred val, a type parameter shadowing a class, and sibling-package imports.

```
FILES 5  FIELDS 50  EDGES 26  BOUND 26  DANGLING 0  NOKEYWORD 0
```

`NOKEYWORD 0` is arm G's audit: every target's span begins with a real `class` / `trait` / `case class` / `sealed` keyword. **A hand-built probe grades only the shapes its author imagined** — arm E said this and both its blocking defects lived in shapes it omitted. It is sufficient to ship the arm and **not** sufficient to claim any yield or orphan clearance for #6726.

## 8. Mutants — 30 scored

**25 DEAD.** Five carry a stated non-DEAD verdict, and **all four ALIVE ones were additionally checked to produce a byte-identical edge set on both the probe and the corpus**, so the verdicts rest on a measurement as well as an argument.

| # | mutation | verdict | killers |
|---|---|---|---|
| M1 | annotation subtree no longer skipped | DEAD | `..._AnnotationNameIsNeverATarget` |
| M2 | dotted `stable_type_identifier` descended | DEAD (2) | shape space + positive control |
| M3 | `singleton_type` descended | *equivalent* | see below |
| M4 | type-parameter shadow no longer refused | DEAD | `..._TypeParameterNeverShadowsASameFileClass` |
| M5 | window no longer stops at `=` | DEAD (2) | `..._InitializerIsNeverTheDeclaredType`, `..._ClassParameterDefaultIsNeverTheDeclaredType` |
| M6 | window no longer requires `:` | *alive, mutually masked* | see below |
| M7 | **arm D's rule** — count every kind | DEAD (5) | sealed-trait, const-group, operation, resolver drive, unit |
| M8 | **arm C's rule** — count records | DEAD (6) | companion ×2, import-beside-class, unit |
| M9 | subtype allow-list removed | DEAD (5) | import placeholder ×2, object ×2, unit enumeration |
| M10 | target `Kind` check removed | DEAD | `..._Unit_TargetKindMustBeComponent` |
| M11 | conjunct 1 not same-file scoped | DEAD | `..._Unit_ForeignRecordNeverSuppressesAnEdge` |
| M12 | conjunct 2 not same-file scoped | DEAD | `..._Unit_ForeignRecordIsNeverATarget` |
| M13 | conjunct 3 not same-file scoped | DEAD (2) | `..._Unit_ForeignFieldIsNeverAnAnchor` |
| M14 | anchor kind/subtype check removed | DEAD (2) | `..._Unit_OnlySchemaFieldsAreAnchors` |
| M15 | `ref_kind` near-miss (`field_target_types`) | DEAD (35) | pinned by independent literals |
| M16 | `append` → assign | DEAD (3) | `..._Unit_ExistingRelationshipsSurvive` |
| M17 | per-ToID dedup removed | DEAD | `..._DuplicateCandidatesEmitOneEdge` |
| M18 | capture no longer deleted | DEAD (3) | premise + leak tests |
| M19 | empty `Metadata` map left behind | DEAD (2) | premise + unit |
| M20 | `field_name` carries the qualified name | DEAD | `..._EdgeShapeAndProperties` |
| M21 | `target_type` from the candidate | *equivalent* | see below |
| M22 | empty-Name records admitted | DEAD | `..._Unit_EmptyNameRecordsAreIgnored` |
| M23 | structural-ref language segment wrong | DEAD | `..._EdgeShapeAndProperties` |
| M24 | attach runs BEFORE the value-set passes | *equivalent under the shipped rule* | see below |
| M25 | emit site stops passing the type-param set | DEAD | `..._TypeParameterNeverShadowsASameFileClass` |
| M26 | case-class parameter anchor stops capturing | DEAD (6) | |
| M27 | template-body val/var anchor stops capturing | DEAD (24) | |
| M28 | attach runs AFTER `TagRelationshipsLanguage` | DEAD | `..._EdgeShapeAndProperties` (language tag) |
| M29 | ambiguity guard deleted outright | DEAD | `..._Unit_InFamilyRivalRefusesTheEdge` |
| M30 | dedup keyed on the candidate, not the ToID | *equivalent* | `targets` is keyed by name and ToID is a pure function of the name |

### The non-DEAD verdicts, each with its reason

**M21 — equivalent by construction.** `targets` is keyed by the candidate and `t.name` is set from the record `Name`, which *is* the key. `cand == t.name` always.

**M30 — equivalent by construction**, same reason.

**M24 — equivalent under the shipped rule, and the header says so.** Moving the attach call before `emitConstantSets` / `emitExceptionFlowEdges` changes nothing, because every record those passes add is `SCOPE.Enum` (out of family, not an admissible target) or carries a sentinel `SourceFile`. The placement is **defensive**, and it is what makes the arm-D counterfactual measurable at all. **The one direction in which the placement IS load-bearing is graded** — M28, moving the call after `TagRelationshipsLanguage`, drops the `language` property off every edge and is DEAD.

**M3 — `singleton_type` is redundant today, and that is recorded in the header rather than quietly kept.** Probed on `X.type` and `Outer.Inner.type`: the node's children are `identifier` / `stable_identifier` plus the `type` keyword and **never** a `type_identifier`, so the scanner collects nothing from it either way. Kept because it states the intent at the node the reader looks at.

**M6 — ALIVE, and the honest reading is that it and M1 are MUTUALLY MASKING.** Removing either alone changes no output; removing **both** is caught by two tests (`..._AnnotationNameIsNeverATarget`, `..._ModifierAnnotationIsNeverATarget`). The compound is offered as evidence of **nothing**, per arm D's rule: *a compound grades its second half only if its first half is not lethal alone*, and M1 is lethal alone.

The distinguishing input for M6 would be a **pre-colon `type_identifier` that is not inside an `annotation` node**. The `val_definition` / `class_parameter` child space was enumerated by probe — `modifiers → access_modifier → access_qualifier → identifier`; the name is `identifier` / `identifiers`; `val`/`var` and punctuation are anonymous — and the only such node is the `annotation`, whose subtree M1's guard refuses. **No fixture was manufactured to kill it.** The colon requirement is therefore stated in the header as a *defensive* structural scope, and **not** claimed to be independently load-bearing. Both facts are in the production comment, so the next reader does not have to rediscover them.

## 9. Gates

- `go test -count=1 ./internal/extractors/scala/` — green, 48 test rows.
- `GRAFEL_HOME=$(mktemp -d) GRAFEL_DAEMON_ROOT=$(mktemp -d) go test -count=1 ./cmd/grafel/` — green (176s), isolated.
- `go test -count=1 ./internal/quality/...` — green (baseline unmoved).
- `go test -count=1 ./internal/verify/` — green.
- `go test -count=1 ./internal/extractors/...` — green.
- `go vet ./...` — clean. `gofmt -l internal cmd` — empty.

## 10. Not established

- **No orphan-clearance figure for #6726.** The corpus yields zero and the golden set yields zero; the 26 edges are from a hand-built probe. The mechanism ships; no yield claim is made.
- **Whether a cross-file Scala arm is worth doing.** For php the answer was measured (43 candidates). Here it is **0 cross-file candidates on 20 files** — but that is a statement about a 2-repo demo corpus, not about Scala.
- **Whether Scala has a same-name `SCOPE.Service` / pattern-lane twin.** `internal/patterns` mints `SCOPE.Model`, which IS in the family, and the in-family guard is graded against it — but no survey of what the pattern lane actually emits for Scala files was run. The guard is correct either way; its *incidence* is unmeasured.
- **The multi-declarator naming defect is pre-existing and untouched.** `val tg, th: Repo` mints a single field record named `Holder.Repo` (`extractName` falls through to the `type_identifier`). Observed by probe; not this arm's to fix.

Refs #6912, #7038, #7047, #7056, #6726.


cajasmota (commented):
## REQUEST-CHANGES — independent adversarial review of #7064 (scala, arm K)

Reviewed on my own checkout of `0bd0c083f` (`review-7064`), all probes driven through the real
`Extract` and, where stated, through `resolve.BuildIndex` + `resolve.ReferencesEmbedded`. Probe
files were deleted and the tree restored (`shasum` of `field_type_refs.go` unchanged,
`git status --porcelain` empty) before posting.

The arm is strong. Every premise I probed at source held. But there is **one real defect** — a
confidently-wrong, fully-binding edge — in the exact guard this arm added, and it is ungraded for
precisely the reason the kotlin arm was sent back a day ago.

---

### FINDING 1 (blocking) — a variance-annotated type parameter bypasses the shadow guard and mints a wrong, binding edge

`scalaTypeParameterNames` collects the declaration's type-parameter names from the direct children
of `type_parameters`, matching `identifier` / `type_identifier` and — per the comment — the wrappers
`variant_type_parameter` / `type_parameter`.

**Neither wrapper node exists in this grammar.** Probed CST:

```
class Box[+A](val a: A)
  type_parameters "[+A]"
    covariant_type_parameter "+A"        <-- not variant_type_parameter
      + ; identifier "A"

class Box[-A](val a: A)
    contravariant_type_parameter "-A"
```

So `+A` / `-A` are **never added to the type-parameter set**, and the guard is a no-op for them:

```
class A
class Box[+A](val a: A)      →  EDGE Box.a -> A          <-- WRONG
class Box[A](val a: A)       →  (no edge)                <-- the shipped control, correct
class Box[-A] { val a: A }   →  EDGE Box.a -> A          <-- WRONG
class T; class R
class Fn[-T, +R](val in: T, val out: R)  →  Fn.in -> T, Fn.out -> R   <-- both WRONG
```

Driven through the real resolver, the wrong edge **binds**, it does not dangle:

```
VARIANCE-SHADOW: edges=[Box.a -> A] bound=1 dangling=0 boundTo=[87b5ddabefe7a441]
  class A id = 87b5ddabefe7a441
```

`bound=1 dangling=0` is the #7056 signature exactly: nothing we own can see it. This is the same
over-fire arm F had to pin as KNOWN-WRONG for Java, and the header's claim —

> Scala's does too, but the declaration's `type_parameters` list is right there at the emit site,
> so the shadow is **refused** instead of documented.

— is false for `[+A]` / `[-A]`, which is the *dominant* declaration form for the collection-shaped
generics Scala is written in (`List[+A]`, `Option[+A]`, `Function1[-T, +R]`).

Two corroborating facts:

- I scored a mutant that renames both switch arms to names that can never match
  (`variant_type_parameter` → `zz_never_…`, same for `type_parameter`): **ALIVE, 0 failing tests.**
  They are dead code, and the comment asserting they exist "in some grammar revisions" is prose
  asserting what no test observes.
- `TestScalaFieldTypeRefs_TypeParameterNeverShadowsASameFileClass` varies the parameter's **name**
  (`[T]` vs `[U]`) and the **anchor** (class parameter vs template body) while holding the
  parameter's **declaration form** constant at bare-invariant. That is the kotlin-arm shape from
  #7063 restated: one axis enumerated thoroughly makes its sibling look covered.

I am flagging, not prescribing — I have verified the hole, not any fix for it.

### FINDING 2 (blocking, same guard) — the nested-generic shadow is a WRONG BINDING, is filed as a "recall floor", and is pinned by nothing

PR §5 lists, under *"Recall floors, pinned as tests rather than described"*:

> **Type parameters are scoped to the immediately enclosing declaration** — `walkNode` carries no
> lexical stack, so a field of a NESTED class typed by the OUTER class's `T` keeps that shadow.

Both halves of that framing are wrong:

```
class T
class Outer[T] { class Inner(val x: T) }   →  EDGE Inner.x -> T
NESTED-SHADOW: bound=1 dangling=0 boundTo=[7c103fc2442ff656]  ( == class T's id )
```

Inside `Outer[T]`, `T` is the type parameter; the top-level `class T` is shadowed. The pass emits an
edge to the class anyway, and it **binds**. That is not a recall floor — a recall floor loses an
edge and is safe; this gains a wrong one and is invisible. And it is not "pinned as a test": grep
for `Outer` / `Nested` / nesting across all three new test files returns nothing. The record set is
byte-identical to the top-level control where the same edge is correct, so no instrument in this PR
or downstream distinguishes them.

The author decided this over-fire was worth a guard and a test at the top level. The same over-fire
one nesting level down is neither guarded nor graded nor honestly labelled.

### FINDING 3 (minor) — the ambiguity guard is unreachable through `Extract`, which is a stronger statement than the PR makes

PR §10 says only that the in-family rival's *incidence* is unmeasured. It is more than that:
`attachScalaFieldTypeRefs` is handed `entities`, the slice this extractor alone built, and this
extractor mints exactly one family Kind (`SCOPE.Component`). `len(famKinds[name]) > 1` therefore
**cannot fire through `Extract`** — pattern-lane `SCOPE.Model` records are not in that slice. M29 /
my MR14 are DEAD only against hand-built record sets. The guard is defensive, which is fine; the
claim that its unit is the KIND is measured only counterfactually. Worth saying out loud in the
header so the next arm does not inherit "graded" from this one.

### Nit — `scHandEmit` hardcodes `"scope:component:class:scala:Test.scala:"+target`

Production uses `extractor.BuildComponentStructuralRef`. They agree today (I checked
`internal/extractor/structural_ref.go:31-33`), but the `WOULDBIND` tests would keep passing if the
builder's format moved. Not blocking.

---

### Premises I probed at source rather than accepting — all TRUE

| claim | verdict |
|---|---|
| `type Alias = X` parses `type_definition`, `walkNode` has no case, mints **no entity** | **TRUE** — verified top-level, inside an `object`, and for Scala 3 `opaque type`. The one-tier conclusion stands. |
| `buildImports` names the placeholder after the FIRST dotted segment | **TRUE** — `import Order.Status` mints bare `SCOPE.Component` `Order`, subtype `""`; golden fixture shows `javax` / `play` / `models` / `scala` placeholders |
| the refused import-placeholder edge **binds**, it does not dangle | **TRUE** — harness drives the real `BuildIndex`/`ReferencesEmbedded`; the "would guarantee a dangling stub" justification really is false here |
| the three out-of-family rivals exist | **TRUE** — `sealed trait Shape` + `case object` → `SCOPE.Enum` *named Shape*; `object Order { val … }` → `SCOPE.Enum` named `Order`; `def Order()` → `SCOPE.Operation` named `Order`. None blocks the edge. |
| companion `case class Order` + `object Order` = two records, one Kind, one node, edge emitted | **TRUE** (this is *not* #7038 repeated) |
| Scala 3 `enum Color` mints `SCOPE.Enum` only, no Component | **TRUE** |
| abstract `val` in a trait mints no field; second parameter list ignored | **TRUE** |
| golden `scala-play-mini` = 3 files, 8 fields, **0 edges** | **TRUE**, measured, not assumed — so the #6118 digest is unmoved because the producer does nothing there |
| a bounded `[T <: Repo]` / context-bound `[T: Ordering]` does **not** over-refuse a field genuinely typed `Repo`/`Ordering` | **TRUE** — I went looking for this over-refusal and it is not there |

### Mutants I scored independently (16, restored by `cp`, `go vet` before each score, edit-landed asserted)

| # | mutation (permissive direction first) | verdict |
|---|---|---|
| MR1 | `variant_type_parameter` / `type_parameter` arms made unmatchable | **ALIVE — dead code → FINDING 1** |
| MR2 | subtype allow-list admits `object` | DEAD (3) |
| MR3 | subtype allow-list removed entirely | DEAD (5) |
| MR4 | ambiguity threshold widened `>1` → `>2` | DEAD (1) |
| MR5 | pass 1 (family count) not same-file scoped | DEAD (1) |
| MR6 | `annotation` no longer skipped | DEAD (1) |
| MR7 | dotted `stable_type_identifier` descended | DEAD (2) |
| MR8 | colon requirement removed (author's M6) | *alive, mutually masked* — I confirmed the author's own verdict independently. I enumerated the `val_definition` / `class_parameter` child space by CST dump and found no legal pre-colon `type_identifier` outside an `annotation`. The one shape that puts a colon **below** the top level (`val b: Repo*` → `repeat_pattern > typed_pattern`) mints no field record at all, so it distinguishes nothing. Verdict accepted as stated. |
| MR9 | window no longer stops at `=` | DEAD (2) |
| MR10 | type-parameter shadow guard removed | DEAD (1) — *for invariant parameters only; see FINDING 1* |
| MR11 | target `Kind` check removed | DEAD (1) |
| MR12 | anchor same-file conjunct removed | DEAD (2) |
| MR13 | pass 2 (target scan) not same-file scoped | DEAD (1) |
| MR14 | ambiguity guard deleted outright | DEAD (1) — but see FINDING 3 |
| MR15 | empty-`Name` records admitted | DEAD (1) |
| MR16 | per-ToID dedup removed | DEAD (1) |

14/16 DEAD, including **every** widening I threw at the target-side allow-list, the Kind check, the
same-file conjuncts and the capture window. The suite is genuinely good in the permissive direction
everywhere except the type-parameter set, which is where the one hole is.

### Gates re-run on my checkout

- `go vet ./internal/extractors/scala/ ./cmd/grafel/` — clean (exit 0)
- `go test -count=1 ./internal/extractors/scala/ ./internal/resolve/` — green
- `go test -count=1 ./internal/quality/... ./internal/verify/ ./internal/extractors/...` — green, 80 packages
- `GRAFEL_HOME=$(mktemp -d) GRAFEL_DAEMON_ROOT=$(mktemp -d) go test -count=1 ./cmd/grafel/` — green, 191s, isolated

### What I did not establish

- I did not check whether `[+A]` / `[-A]` beside a same-file class of the same name occurs in any
  real corpus. It does not need to: the invariant case doesn't either, and it was guarded.
- I did not attempt a fix for either finding. **Prescribing is not reviewing** — I verified the
  holes, not any repair of them.
- The corpus zero: I accept it as characterised (24 candidates, all external). I have no evidence it
  hides a producer that never fires — the 5-file probe and my own 30-odd probe files fire it
  constantly. But the hand-built probe grades only the shapes its author imagined, and FINDING 1 is
  a shape it did not imagine, which is the point.

cajasmota:
## REQUEST-CHANGES addressed — `a0c9e2e0b`

Both blocking findings confirmed at source and fixed; F3 stated at its real strength. Nothing the review confirmed TRUE was regressed.

---

### F1 — the two switch arms were dead code, and the header's claim was false

Confirmed exactly as reported. The CST dump gives `covariant_type_parameter` / `contravariant_type_parameter`; `variant_type_parameter` / `type_parameter` do not exist in this grammar.

**The direct-child space of `type_parameters`, enumerated by CST dump rather than recalled:**

| child node | written | what it names |
|---|---|---|
| `identifier` | `[A]` | THE PARAMETER — refuse |
| `covariant_type_parameter` | `[+A]` | `+`, `identifier`, bounds — refuse the identifier |
| `contravariant_type_parameter` | `[-A]` | `-`, `identifier`, bounds — refuse the identifier |
| `upper_bound` | `[A <: Repo]` | a REAL type — keep bindable |
| `lower_bound` | `[A >: Repo]` | a REAL type |
| `view_bound` | `[A <% Repo]` | a REAL type |
| `context_bound` | `[A : Ord]` | a REAL type |
| `type_parameters` (nested) | `[F[_]]`, `[F[X]]` | F's OWN params — not in scope for fields |
| `annotation` | `[@specialized A]` | an annotation CLASS — keep bindable |
| `wildcard` | `[_ <: Repo]` | no name at all |
| `[ ] , :` | | anonymous |

**A `type_identifier` is NEVER a direct child** — every one sits inside a bound, an annotation or a nested list. So the name arm now matches `identifier` alone; if the grammar ever minted a direct one it would be a bound's RHS, i.e. a type that must stay bindable.

**Graded in both directions, with a positive control in every fixture** (`TestScalaFieldTypeRefs_TypeParameterFormSpace`, 17 rows + a template-body/trait repeat on the other anchor). Every row's source is legal Scala — a contravariant parameter sits in a function-type argument (`class Sink[-A](val f: A => Real, …)`), never directly on a `val`.

**The over-refusal direction is now PINNED, not merely absent.** The reviewer went looking for it on bounded and context-bound parameters and correctly found it was not there; it is now graded so it cannot arrive silently. Bound / context-bound / annotated / higher-kinded-inner rows all carry the RHS as a same-file class that MUST still bind (`Box.r -> Repo`, `Box.o -> Ord`, `Box.x -> X`). Two mutants exist purely for that direction — MR1e and MR3b — both DEAD.

The header's "refused instead of documented" sentence is corrected in place rather than narrowed, and says out loud that it was false for the dominant form, that a mutant proved the arms dead, and why the shipped test's axis choice hid it.

### F2 — the nesting rule, derived from Scala, not ported from Kotlin

Confirmed: `class T; class Outer[T] { class Inner(val x: T) }` emitted `Inner.x -> T` and it bound.

**Scala's rule is the OPPOSITE of arm J's.** Kotlin nests statically by default, so a nested class does not see the outer's type parameters unless declared `inner` — arm J must NOT inherit. Scala has no `inner` keyword and no static nesting at all: every member class, trait and object of a template is an inner member of it, and the template is evaluated in a scope already containing the declaration's type parameters. **`object` is not Kotlin's `companion object` exception** — an object nested inside a class is a per-instance member, only a TOP-LEVEL object is static-like — so it does not block.

So the scope **accumulates**: `walkNode` carries the enclosing set and each declaration merges its own into a **copy**, so nothing leaks sideways into a sibling or forward out of a nested body.

Enumerated and graded (`TestScalaFieldTypeRefs_NestedDeclarationInheritsTheOuterTypeParameters`):

| row | direction |
|---|---|
| nested in a generic class / generic trait | refuse |
| nested inside an `object` inside a generic class | refuse (Scala ≠ Kotlin) |
| an `object`'s own `val` inside a generic class | refuse |
| two levels down through a NON-generic intermediate | refuse |
| nested in a COVARIANT generic class (the two fixes compose) | refuse |
| template-body `val` of a nested class (the other anchor) | refuse |
| nested class with its OWN shadowing parameter | refuse (by its own) |
| **nested in a NON-generic class** | **MUST BIND** — nesting alone must never refuse |
| **outer parameter named differently** | **MUST BIND** |
| **sibling of a generic class** | **MUST BIND** — no sideways leak |
| **sibling under a GENERIC outer** | **MUST BIND** — the only row that distinguishes merge-into-a-copy from merge-into-the-parent's-map (MR6 was ALIVE until it existed) |

**A class inside a generic `def` mints nothing** — `walkNode`'s `function_definition` arm returns without recursing. Checked before writing anything; no fixture was manufactured for a form that mints nothing. The premise is asserted instead (`…_PremiseAClassInsideADefMintsNoField`) and gains a row the day that arm starts recursing.

**Two further routes were probed and found reachable**, and mutants dropping the scope on either were ALIVE until rows existed: `walkNode`'s DEFAULT recursion (a class under an `if`/`try` block in a generic template body really does mint `Inner.x`) and the MALFORMED-declaration arm. The last two of those fixtures are **deliberately not legal Scala** and the test says so out loud.

### F3 — stated at its real strength, in the source

Agreed and adopted: the guard **cannot fire through `Extract` at all**, which is stronger than "incidence unmeasured". `records` is the slice this extractor alone builds and it mints exactly one family Kind.

**Decision: the guard stays**, as the defence against a second family-Kind producer landing in this package — the day a `SCOPE.Class` / `SCOPE.View` arm arrives, the count starts mattering and the unit (the KIND, not the record) is already right. The reason is now **in the source**, not only here, and a new test (`…_Unit_TheAmbiguityGuardCannotFireThroughExtract`) observes the one-Kind premise on a broad source and fails loudly the day it moves — verified non-vacuous by a positive control.

### Not regressed

All nine confirmed premises stand untouched, including `type Alias = X` minting no entity (so the **one-tier rule stands**), the import-placeholder first-segment naming, the refused-edge-**binds** claim, companion-object-is-one-node, and the three out-of-family rivals.

**MR8 kept as-is**, with the stronger reason recorded in the header: the one shape putting a `:` below the top level — `val b: Repo*` — mints no field record, so it distinguishes nothing. No fixture was manufactured to kill an unreachable mutant.

The `scHandEmit` nit is noted and left (non-blocking, and changing it would weaken the WOULDBIND tests' independence from the builder).

---

### Re-measurement

**5-file idiomatic probe, driven through `resolve.BuildIndex` + `resolve.ReferencesEmbedded`:**

```
parent 0bd0c083f : FILES 5  FIELDS 57  EDGES 39  BOUND 39  DANGLING 0  NOKEYWORD 0
fixed  a0c9e2e0b : FILES 5  FIELDS 57  EDGES 37  BOUND 37  DANGLING 0  NOKEYWORD 0
```

**Δ −2, and both dropped rows were WRONG BINDINGS, named:**

| row | why it was wrong |
|---|---|
| `Domain.scala \| Fn.f -> T` | `class Fn[-T, +R](val f: T => R, …)` — **F1**, contravariant parameter bound to the top-level `class T` |
| `Domain.scala \| Boxed.x -> T` | `class Outer[T] { class Boxed(val x: T, …) }` — **F2**, nested class bound to the `class T` the outer shadows |

No correct edge was lost: all 37 survivors are byte-identical to the parent's, and `Holder.item -> T` (the invariant form the shipped guard already refused) is unchanged.

Note the probe declares only one collidable name (`class T`), so **−2 is a floor on the defect's incidence, not an estimate of it** — the real rate depends on how often a file declares a class named like a variance parameter.

**Corpus:** 20 `.scala` files / 2 repos, **29 fields, 0 edges** — unchanged, and structurally so (all declared types are external). The fix can only remove edges, and there were none.

**Golden digest unmoved** — `cmd/grafel` green in an isolated env (183s).

### Mutants — 13 scored, 11 DEAD

| # | mutation | verdict | killers |
|---|---|---|---|
| **MR1** | **the reviewer's, VERBATIM — both variance arms unmatchable** | **DEAD** | 11 rows (was 0 before) |
| MR1a | covariant arm only unmatchable | DEAD | 10 |
| MR1b | contravariant arm only unmatchable | DEAD | 4 |
| MR1c | delete the variance arm's `break` | *equivalent under an OBSERVED grammar fact* | see below |
| MR1e | a variance wrapper's bounds join the shadow set (over-refusal) | DEAD | 6 |
| MR2 | re-add `type_identifier` to the name arm | *equivalent under an OBSERVED grammar fact* | see below |
| MR3 | bound wrappers matched as names at the top level | DEAD | 3 |
| MR3b | bound wrappers DESCENDED at the top level (over-refusal) | DEAD | 7 |
| MR4 | the nesting fix reverted (arm J's kotlin rule) | DEAD | 8 |
| MR5 | members get the ENCLOSING set only, not their own | DEAD | 8 |
| MR6 | the merged set mutates the parent's map (sideways leak) | DEAD | 2 |
| MR7 | the MALFORMED-declaration arm drops the scope | DEAD | 2 |
| MR8 | the DEFAULT recursion drops the scope | DEAD | 5 |

**The two non-DEAD verdicts rest on a MEASUREMENT, not an argument.** `TestScalaFieldTypeRefs_PremiseTypeParametersChildShapes` walks 21 declaration forms and asserts (1) no direct child of `type_parameters` is a `type_identifier` — so MR2 changes nothing, and (2) a variance wrapper has exactly ONE direct `identifier` child — so MR1c changes nothing. It refuses to pass on a vacuous walk (floors on both counters) and it pins the child-node-type set, so a NEW child type is a visible failure rather than a silent gap. If either fact moves, that test fails **before** the mutant becomes live — and for MR2 the failure message says explicitly that it has become an over-refusal.

MR1d was scored, found unreachable (the `break` fires before any bound child), **rebuilt as MR1e so the mask could not hide it**, and MR1e is DEAD.

### Gates

- `go test -count=1 ./internal/extractors/... ./internal/resolve/ ./internal/verify/ ./internal/quality/...` — green
- `GRAFEL_HOME=… GRAFEL_DAEMON_ROOT=… go test -count=1 ./cmd/grafel/` — green, 183s, isolated
- `go vet ./...` — clean · `gofmt -l internal cmd` — empty

cajasmota (commented):
## APPROVE — independent adversarial RE-review of the fix (`0bd0c083f..a0c9e2e0b`)

Reviewed on my own checkout (`rereview-7064`, head `a0c9e2e0b`), in my own worktree. I attacked the
**fix**, not the premises the first review already confirmed. Probe files were deleted, every mutant
restored by `cp` from a byte-level backup, and the tree verified clean (`git status --porcelain`
empty, `shasum` of `field_type_refs.go` / `scala.go` / both touched test files unchanged) before
posting.

**Both blocking findings are genuinely fixed, and — the part that mattered — the over-refusal
direction is really pinned, not merely absent.** I went at it with six independent over-refusal
mutants and the table killed every reachable one. I found no blocking defect. Two nits below.

---

### 1. The enumerated direct-child space of `type_parameters` — re-dumped, not accepted

I dumped the CST myself over **32 declaration forms**, a superset of the premise test's 21, chosen
to break the enumeration: `[A >: Low <: High]`, `[A : Ord : Eq]`, `` [`A`] ``, `[A <: B[A]]`,
`[M[_[_]]]`, `[@specialized(Int) A]`, `[F[+X]]`, `[A <: Repo with Named]`, `[A <: Repo & Named]`,
and the same lists behind `sealed` / `final` / `implicit` / `abstract` / `case` prefixes.

**The enumeration holds exactly.** The child set observed across all 32 forms is precisely the
`known` map in the premise test — `[ ] , :` · `identifier` · `wildcard` · `annotation` ·
`covariant_type_parameter` · `contravariant_type_parameter` · `upper_bound` · `lower_bound` ·
`view_bound` · `context_bound` · nested `type_parameters` — with **no new node type and no direct
`type_identifier`**. Matching `identifier` alone is licensed.

FACT 2 re-derived independently on 9 variance forms (`[+A]` · `[-A]` · `[+A <: Repo]` ·
`[-A >: Repo]` · `[+A : Ord]` · `[+A <% Ord]` · `[@specialized +A]` · `[+A >: Low <: High]` ·
`[+F[X]]`): **exactly one direct `identifier` child in every one**, and every bound stays wrapped in
its own node inside the wrapper. So MR1c/MX11 really is equivalent, by measurement.

**The premise test is non-vacuous in the detect-and-act sense, not just the count-floor sense.** A
count floor catches only "didn't read", so I broke the premise three ways and confirmed each one
fails loudly:

| positive control | result |
|---|---|
| FACT 1 detector retargeted to a child type that DOES occur (`identifier`) | **FAILS**, `FACT 1 GONE` |
| FACT 2 threshold flipped to `ids != 2` | **FAILS**, `FACT 2 GONE` |
| `upper_bound` removed from the `known` child-type pin | **FAILS**, `NEW type_parameters child node type` |

### 2. Over-refusal — attacked hardest, and it holds

Driven through the real `Extract`, with the RHS declared as a same-file class so a silent drop
becomes a visible missing row. **All twelve bind:**

`[A <: Repo]` · `[A >: Repo]` · `[A <% Ord]` · `[A : Ord]` · `[@specialized A]` · `[F[X]]` (inner
`X` → binds, correct: `X` is scoped to `F`, not to the class) · `[F[+X]]` · `[+A <: Repo]` ·
`[-A >: Repo]` · `[+A : Ord]` · `[_ <: Repo]`. And the one that must NOT bind — a bound whose RHS is
**itself a type parameter**, `class B[A, C <: A](val f: A)` — is correctly refused.

Six over-refusal mutants, each targeting **one** part of the compound so the guard is scored part by
part rather than only as a whole:

| # | mutation | verdict | killers |
|---|---|---|---|
| MX1 | `upper_bound` descended at the top level | **DEAD** | 2 |
| MX5 | `lower_bound` descended | **DEAD** | 1 |
| MX6 | `view_bound` descended | **DEAD** | 1 |
| MX2 | `context_bound` descended | **DEAD** | 2 |
| MX3 | `annotation` descended | **DEAD** | 1 |
| MX4b | nested `type_parameters` descended for `identifier` | **DEAD** | 1 |
| MX12 | the variance wrapper reads its OWN bounds (grades "deliberately not read") | **DEAD** | 5 |
| MX4 | nested `type_parameters` descended for `type_identifier` | *equivalent under an observed grammar fact* — a nested list holds `identifier`/`wildcard`, never `type_identifier`; the reachable form is MX4b, DEAD | 0 |
| MX11 | the variance arm's `break` deleted (= MR1c, verbatim) | *equivalent under an observed grammar fact* — one direct `identifier` child, re-derived above | 0 |

### 3. The Scala nesting rule — the language claim checked, then attacked from the other side

The claim is right. Scala has no `inner` keyword and no static nesting; a member `object` of a class
is a per-instance member, so it does **not** block the way Kotlin's `companion object` does, and the
scope accumulates. The inverse question — *is there a Scala form where the parameter is genuinely
NOT in scope, which this now over-refuses?* — I probed the four candidates named and found none:

| probe | result |
|---|---|
| class inside a generic `def` | **mints nothing** — verified, so nothing to over-refuse (`function_definition` returns without recursing) |
| class inside an `object` inside a generic class | refuses — correct, `object` is an inner member |
| class inside a generic `trait` | refuses — correct |
| nested declaration REDECLARING the same name | refuses by its OWN parameter — the inner binding wins |
| **generic class inside a TOP-LEVEL object** | refuses its own param only; the object's later `val x: T` **BINDS** — no leak up |

Leak directions, all clean: **sideways** (`class Outer[U] { class Mid[T](…); class Later(val x: T) }`
→ `Later.x -> T` binds), **downward-then-back-up** (`class Wrap { class Inner[T](…); val x: T }` →
`Wrap.x -> T` binds), **forward in the same body**, and top-level siblings. MX7 (merge into the
parent's map instead of a copy) is **DEAD**, killed by exactly the row the author says is the only
one that can see it. MX8 (ignore `enclosing` entirely, i.e. revert F2) is **DEAD**, 14 rows.

Both `walkNode` routes behave as claimed — a class under an `if`/`try` block in a generic template
body really does mint `Inner.x`, and the two malformed fixtures are labelled *not legal Scala* in
the test text. Nothing else depends on malformed input.

### 4. Anchor / declaring-form / modifier axes — the shape that produced a defect on the last three arms

The form-space table holds the declaring form at `class` and carries no modifier prefixes, so I ran
that axis myself, 14 rows: class-parameter `val` / `var` / plain case-class param / `val` on a
`case class` · template-body `val` / `var` / `lazy val` / `override val` · `trait` · `sealed trait` ·
`abstract class` · `final class` · `implicit class`. **Every one refuses the variance parameter and
emits the control.** No held-constant axis on this arm hides a behavioural difference.

Two anchor mutants confirm the two producers are graded separately: MX10 (template-body anchor loses
its own params) **DEAD**, MX9b (class-parameter anchor loses them) **DEAD**, 16 rows.

### 5. F3 and the measurements

`…_Unit_TheAmbiguityGuardCannotFireThroughExtract` is **non-vacuous in the acting sense**: adding
`SCOPE.Enum` to `scIsComponentFamilyKind` makes it fail with `PREMISE MOVED: Extract now mints 2
family Kinds for "Status"`. The retention reason is true — `records` is this extractor's own slice,
`typeParams` is read-only downstream, and one family Kind is minted.

Verified rather than accepted: **golden digest unmoved** — `GRAFEL_HOME=… GRAFEL_DAEMON_ROOT=…
go test -count=1 ./cmd/grafel/` exit **0**, 184.7s, fresh temp dirs. **Corpus shape** — 20 `.scala`
files across exactly 2 repos, as stated. `./internal/extractors/scala/ ./internal/resolve/` exit 0;
`gofmt -l internal cmd` empty. The −2 caveat is stated in the conservative direction and is correct:
one collidable name is a floor, and more collidable names can only remove more wrong edges.

---

### Nit 1 (non-blocking) — one of the two class-parameter call sites is graded by an incidental row

`emitScalaCaseClassFields` is called twice, and which one runs depends on whether the class has a
template body. Every row of the 17-row form-space table is a bodyless `class Box[+A](…)`, so the
table grades only the `body == nil` site (MX9b: 16 killers). The other site — the one that runs for
**every generic class that has a body**, which is most of them — is killed by exactly one
assertion, `Sink.f0 -> Real`, a field the fix's own comment describes as "pre-existing and not this
arm's". Distinguishable and graded, so not a finding; but it is graded by accident. One row of the
form `class Box[+A](val a: A, val r: Real) { }` would grade it on purpose. *Prescribing is not
reviewing — I have verified the thinness, not that fix.*

### Nit 2 (non-blocking, pre-existing) — the template-body anchor is only reachable with an initializer

An abstract member declaration (`val a: A` with no `=`) mints **no field record at all** in this
extractor. The fix's template-body fixtures all use `= null` / `= a0`, so none of them is vacuous —
I checked each. Worth recording so a future arm does not write a template-body row without an
initializer and grade nothing.

---

**Verdict: APPROVE.** The two blocking defects are fixed at the root rather than patched at the
symptom, the direction with no symptom is the one that is now pinned hardest, and the two non-DEAD
mutants rest on grammar facts I re-derived independently on a wider form space than the test walks.
I did not merge and did not push.

Issue #7060, opened 2026-09-15
C++ headers are parsed with the C grammar: .h classifies as 'c' and the cpp extractor hard-selects cGrammar(), so 123 of 172 corpus files lose every class/template/namespace — 51 field entities instead of 254


cajasmota:
`.h` classifies as language `c`, and the cpp extractor hard-selects the **C grammar** for that language. So a header containing C++ — `class`, `template`, `namespace` — is parsed by a grammar in which none of those constructs exist, and the entities inside them are never emitted.

## The three lines

```go
internal/classifier/classifier.go:389     ".h": "c",

internal/extractors/cpp/extractor.go:53   extractor.Register("c",   &CppExtractor{lang: "c"})
internal/extractors/cpp/extractor.go:91   grammar := cppGrammar()
internal/extractors/cpp/extractor.go:92   if lang == "c" { grammar = cGrammar() }
```

The extractor **is** registered for `"c"` and **does** run. It simply parses with `tsc.Language()` instead of `tscpp.Language()`.

## Measured

On the corpus, counted while implementing the cpp arm of #6912 (PR #7057) and reproduced independently by its reviewer:

| | as classified today | with `.h` parsed as C++ |
|---|---|---|
| C/C++ files | 172 | 172 |
| **field entities** | **51** | **254** |
| field→type edges | **1** | 23 |

**123 of the 172 files are `.h`.** One repo alone (staxrip) goes from 19 field entities to 137.

So the cpp field→type arm has a real-corpus yield of **one edge**, and that is not because C++ headers are rare — it is because we parse them with the wrong grammar.

## This is a SIBLING of #7046, not the same defect

Worth stating precisely, because the two share an upstream cause and have disjoint consequences — and fixing one does nothing for the other.

- **#7046** is the **custom** lane: `"c"` is routed to no custom-extractor prefix, so 12 `file.Language == "c"` disjuncts in `internal/custom/cpp` are dead.
- **This** is the **core** lane: the core extractor runs happily for `"c"` and picks the C grammar.

**Routing `"c"` to `custom_cpp_` would recover none of the 203 missing field entities**, because the core parse has already thrown the C++ syntax away before any custom pass sees the file. Anyone fixing #7046 and assuming header coverage is now handled would be wrong.

## What is NOT claimed

[...]

PR #7066, opened 2026-09-16
fix(#7041): go treated a type parameter as a same-file type, so `type Box[T any]` beside `type T struct{}` emitted an edge that BOUND — the last arm of #6912 still shipping this known-wrong

### LINKED ISSUE #7041: COMPLETE — all four arms (go, java, rust, C#) no longer emit a WRONG BINDING edge for a field typed by a type parameter; four languages, four incompatible scoping rules, none portable

cajasmota:
Go closes its arm of #6912's type-parameter over-fire. **It is not the last** — **three arms still ship it: C#, Rust and Java** (F2 of the review on this PR; verified at source, not relayed). The arms that have closed theirs are F#, swift (`96bee2ece`), cpp (#7057) and now go. **The first line of this body previously said go was the last, and that was false** — #7041's own body had already recorded arms A (C#) and C (rust) as unchecked, and this body converted an unchecked status into a settled one. Corrected here, and reported for #7041/#6912, whose titles and comments carry the same claim.

| arm | still over-fires? | evidence, read at source on this branch |
|---|---|---|
| **C#** | yes | `internal/extractors/csharp/field_type_refs_6912_test.go:352` — `TestCsharpFieldTypeRefs_KnownOverFire_TypeParameterShadowsSameFileType`, `class Box<Customer>` pinned to `Box.Item -> …:Customer` with a **`t.Fatalf`** |
| **Rust** | yes | `internal/extractors/rust/field_type_refs_6912_test.go:593` — same shape, `struct Box2<Customer>` → `Box2.item -> …:Customer`, **`t.Fatalf`** |
| **Java** | yes | prose claim at `internal/extractors/java/field_type_refs.go:175` (*"IT DOES NOT UNDERSTAND TYPE PARAMETERS … a case a real fix is expected to break"*) plus the pin at `field_type_refs_6912_test.go:450`, which is **also a hard `t.Fatalf`** (via the `javaFTWantEqual` helper, `:167`) — the review called Java prose-only; it is a real gate too |
| F#, cpp, swift, go | no | `fsharp/field_type_refs.go` skips a leading-`'` token whole; cpp emits no field entity for a template's direct members (#7057); swift's `TestSwiftFieldTypeRefs_TypeParameterShadowedByASameFileTypeGetsNoEdge`; go — this PR |
| php, proto | n/a | neither language has generics in field-type position |

**Asymmetry worth stating**: C#'s, Rust's and Java's pins all `t.Fatalf`, so each is a **real gate a fix would break** — unlike go's deleted pin, which `t.Skipf`'d when the over-fire was gone and `t.Logf`'d when it was present, and graded nothing in either direction.

`type Box[T any] struct { Item T }` beside a same-file `type T struct{}` emitted an edge asserting `Box.Item`'s declared type is that struct. It is not — `T` is the type parameter, which shadows the outer declaration inside `Box`. This is the **wrong-binding** direction: #7056 established that no instrument we own can see it, and #6726's orphan rate actively *improves* when it fires.

`goTypeParameterNames` now reads the declaration's type-parameter list and makes it a shadowing scope; the #6912 candidate stash refuses any candidate the set contains.

---

## 1. CST dump evidence for the node names

Trap (a), from the scala arm (#7064): that arm shipped a guard switching on `variant_type_parameter` / `type_parameter` — **neither node exists in that grammar** — and it was dead code no mutant could kill. So the spellings here were **read off a dump of a real parse**, not recalled. The dump (a throwaway `zz_cstdump_7041_test.go`, not committed) over the whole enumerated form space gave:

```
type_declaration  "type Box[T any] struct { Item T }"
  type_spec
    type_identifier               "Box"
    type_parameter_list           "[T any]"
      type_parameter_declaration  "T any"
        identifier                "T"        <- COLLECTED
        type_constraint           "any"      <- NEVER DESCENDED
          type_identifier         "any"
    struct_type                   "struct { Item T }"
```

The forms that matter for the collector's shape, verbatim from the dump:

| source | structure under `type_parameter_list` |
|---|---|
| `type Pair[K comparable, V any]` | **two** `type_parameter_declaration`, one `identifier` each |
| `type M[K, V any]` | **one** `type_parameter_declaration`, **two** `identifier` children, one `type_constraint` |
| `type Con[T Order]` | `identifier "T"` + `type_constraint "Order"` → `type_identifier "Order"` |
| `type Uni[T int｜string]` | `type_constraint` → two `type_identifier` siblings |
| `type Appr[T ~int]` | `type_constraint` → `negated_type` → `type_identifier` |
| `type Lit[T interface{ Ship() }]` | `type_constraint` → `interface_type` → `method_elem` |
| `func Gen[T Order](x T) T` | `function_declaration` carries a `type_parameter_list` directly |
| `func (b Box[T]) Do(...)` | the receiver's `[T]` is **`type_arguments`**, not a parameter list |
| `type Al[T any] = Box[T]` | **the grammar emits an `ERROR` node for the `=`** — the generic-alias form is not supported by this grammar version |

**Every one of those three strings is graded.** `TestGoTypeParameterNames_FormSpace` opens with a premise check that the parse produced a non-zero count of `type_parameter_list`, `type_parameter_declaration` and `type_constraint` nodes, so a "want none" row can never pass vacuously — and mutants M4/M5 below corrupt a matcher string and each kill 18 tests.

## 2. The enumerated form space, graded in both directions

Trap (b). cpp's first refusal was wrong in **both** directions, found only by enumeration. The table is `TestGoFieldTypeRefs_7041_FormSpace`; each row asserts the file's **complete** field→declared-type edge set, so the present and absent directions are the same assertion.

| # | form | refused | **must still bind (over-refusal control)** |
|---|---|---|---|
| 1 | `type Box[T any]`, `type T` **before** | `Box.Item` | `Box.Ok → Ctl` |
| 2 | same, `type T` **after** | `Box.Item` | `Box.Ok → Ctl` |
| 3 | `[K comparable, V any]`, both collide | `Pair.Key`, `Pair.Val` | `Pair.Ok → Ctl` |
| 4 | `[K, V any]` — one decl, two names | `Multi.A`, `Multi.B` | `Multi.Ok → Ctl` |
| 5 | `[K comparable, V any]`, only `K` collides | `Pair.Key` | `Pair.Ok → Ctl` |
| 6 | **`[T Order]` — constraint is a same-file type** | `Con.It` | **`Con.Ord → Order`** |
| 7 | **`[T Order \| Ctl]` — union of same-file types** | `Uni.U` | **`Uni.Ord → Order`, `Uni.Ok → Ctl`** |
| 8 | **`[T ~Alias]` — `negated_type` over a same-file alias** | `Appr.A` | **`Appr.Al → Alias`** (a `type_alias` target, not a struct) |
| 9 | `[T interface{ Ship() }]` | `Lit.L` | `Lit.Ok → Ctl` |
| 10 | **`[T interface{ Order }]` — literal embedding a named type** | `Emb.E` | **`Emb.Ord → Order`** |
| 11 | `[string any]` — parameter named after a predeclared identifier | `Box.S` | `Box.Ok → Ctl`, **`Sibling.S → string`** |
| 12 | wrappers: `*T []T [3]T map[T]T map[T]Holder[T] chan T func(T) error Holder[T]` | all 9 `T` candidates | **`Box.MapH → Holder`, `Box.Wrap → Holder`** |
| 13 | scope reach: sibling declaration | `Box.Item` | **`Sibling.Val → T`, `Sibling.Ptr → T`** |
| 14 | declaring form: generic **interface** | (emits no field entity) | `Use.Ok → Ctl`, `Use.Val → T` |
| 15 | declaring form: generic **function** | (emits no field entity) | `Use.Ord → Order`, `Use.Val → T` |
| 16 | declaring form: method on a **generic receiver** | `Box.Item` | `Use.Val → T` |
| 17 | **control: generic struct, no collision at all** | — | `Box.Ok → Ctl`, `Box.Wrap → Ctl`, `Box.Wrap → Holder` |

Rows 6–10 are the direction cpp got wrong: `Order` / `Alias` there are **constraints, not parameter names**, and harvesting them would silently delete a correct edge. Only `identifier` children **directly** under a `type_parameter_declaration` are collected; everything else in the list lives under a `type_constraint` and is never descended. Mutant M6 (descend the whole declaration) is killed by exactly rows 6, 7, 8 and 10 — the four rows that exist for it.

Row 17 is the control the deleted pin could never have been. It is killed by M7, so it is load-bearing, not decorative.

## 3. VARIED / HELD-CONSTANT, per table

Trap (c), from the kotlin arm: a table that sweeps one axis exhaustively makes its neighbour look covered. Both tables carry the block in-source; summarised:

**`TestGoTypeParameterNames_FormSpace`** (metadata level)
- **VARIED**: constraint form (7 kinds); parameter-list arity (one decl / two decls / one decl binding two names); declaring form (generic struct, generic interface, generic function, method on generic receiver, generic type definition); parameter-name shape (ordinary, predeclared identifier). Three **non-generic** declarations are in the table so "returns empty" is not carried by one row. A closing loop fails if any declaration in the fixture is **not** in the table, so the enumeration cannot silently drift.
- **HELD CONSTANT**: the file (one parse) and package scope — **attacked at the edge level**, where position is swept before/after.

**`TestGoFieldTypeRefs_7041_FormSpace`** (edge level)
- **VARIED**: constraint form; arity; parameter-name shape; **declared-type wrapper form** crossed with the parameter axis; **declaring form**; **position** of the colliding declaration (before/after, rows 1–2 — live because the pass documents that the in-file declaration set is complete only after the walk); **scope reach** (own declaration vs sibling vs file).
- **HELD CONSTANT**, each attacked elsewhere rather than assumed:
  - **the ANCHOR** is always a named struct field. The other two anchors a Go struct body offers get their own test: `TestGoFieldTypeRefs_7041_EmbeddedFieldAndDependsOnAreUntouched` pins the **embedded field** (EXTENDS) and the **struct-anchored DEPENDS_ON** with `t.Error`, both deliberately unchanged.
  - **the TARGET KIND** is mostly `SCOPE.Component/struct`; row 8 carries a `SCOPE.Schema/type_alias` target so the refusal is not graded against one kind only; interface targets are already swept by `TestGoFieldTypeRefs_EdgeSetIsExact`.
  - **the FILE** is one per row; cross-file is unchanged and covered by `TestGoFieldTypeRefs_CrossFileTypeGetsNoEdge`.

## 4. Mutants — permissive direction first

Call-site mutants (not extracted helpers) where the decision is consumed. Each: `cp` backup, exact match-count assertion before applying, `go vet` before scoring, `-count=1`, restore by `cp` + `shasum`.

| # | dir | mutation | verdict | killed by |
|---|---|---|---|---|
| M1 | permissive | delete the refusal at the stash site (`if typeParams[c] { continue }` → `_ = typeParams`) | **DEAD (16)** | 14 form rows + both other tests |
| M2 | permissive | pass `nil` for `typeParams` at the `extractTypes` call site | **DEAD (16)** | same |
| M3 | permissive | collect only the **first** `identifier` per declaration | **DEAD (3)** | row 4 (`[K, V any]`) + the `Multi` metadata row |
| M4 | permissive | `"type_parameter_declaration"` → `"type_parameter"` (a name that does not exist — #7064's exact bug) | **DEAD (18)** | both metadata tests + 14 form rows |
| M5 | permissive | `"type_parameter_list"` → `"type_parameters"` | **DEAD (18)** | same |
| M6 | **over-refusal** | descend the whole declaration, harvesting constraints too (`findAll(d, "identifier", "type_identifier")`) — cpp's shipped bug | **DEAD (7)** | rows 6, 7, 8, 10 + both metadata tests |
| M7 | **over-refusal** | refuse every candidate in any generic declaration (`typeParams[c]` → `len(typeParams) > 0`) | **DEAD (14)** | 12 rows **including the no-collision control, row 17** |
| M8 | — | `if len(out) == 0 { return nil }` → `return out` | **EQUIVALENT UNDER ANY SUITE** — the only consumer is `typeParams[c]`, and a lookup in an empty non-nil map and in a nil map both return `false`. Not "ungraded": no input can distinguish them. |

No fixture was manufactured to kill a mutant; every row exists for a form in the enumeration.

## 5. Corpus delta — a **named** zero, and no row moved

682 Go files (`etcd` 391, `grpc-go-examples` 118, `gin` 99, `chi` 74), copied to scratch, driven through the real extractor and `resolve.BuildIndex` + `ReferencesEmbedded`:

|---|---|---|---|
| **before** (refusal neutralised at the call site) | 168 | 168 | 0 |
| **after** | 168 | 168 | 0 |

The **row sets are identical** — not just the counts; every row was emitted and diffed, and no row is added, removed or re-pointed. Per row 5 of the standing rules, nothing dropped, so nothing needs proving wrong.

The zero is named rather than assumed. A second harness measured the incidence directly:

```
files=682  genericTypeSpecs=8  genericStructs=2  paramNames=14
fieldsInGenericStructs=2  fieldsNamingAParam=0  paramCollidingWithSameFileType=0
```

Eight generic declarations in the whole corpus, **two** of them generic structs, **two** struct fields inside a generic struct, **none** typed by its own parameter, and **no** parameter name anywhere colliding with a same-file declared type. These repos largely predate generics adoption, so the corpus cannot see this defect — exactly cpp's situation (#7057, zero occurrences, fixed anyway). **A corpus-relative zero is not a bound on the language.**

`cmd/grafel` digest: **unchanged**, run with fresh `GRAFEL_HOME` / `GRAFEL_DAEMON_ROOT` (179s, green). The mixed-language fixture's Go members contain no generic declaration, so no digest constant was touched or bumped.

## 6. What happened to the three false prose claims

The pin `TestGoFieldTypeRefs_KnownOverFire_TypeParameterShadowedByASameFileType` **graded nothing**: `t.Skipf` when the over-fire was gone, `t.Logf` when present, **no `t.Error`/`t.Fatal` on any path**. Three places said a real fix was "EXPECTED to break" it.

1. **The test's own doc-comment** (`:514`) — deleted with the test. In its place stands a comment recording what the pin was worth and why it could not have been this fix's safety net (deleting the candidate handling outright would have made it skip too).
2. **`field_type_refs.go:190`** — the `WHAT THIS PASS IS NOT` bullet, rewritten from "IT DOES NOT UNDERSTAND TYPE PARAMETERS … a case a real fix is expected to break" to "IT DOES UNDERSTAND TYPE PARAMETERS, SINCE #7041", naming the graded replacement, stating plainly that the old claim was false, and adding the still-out-of-scope boundary (DEPENDS_ON and EXTENDS).
3. **The comment on #7041** — the owner had already corrected it on the issue; this PR is its resolution.

No narrowed-but-still-generous version of any of the three was left behind. The `goFieldTypeRefsMetaKey` doc was also amended, since it asserted that *nothing* is decided during the walk and one thing now is — with the reason the two arguments for deferring do not apply to a lexical scope.

## 6b. Review round 2 — three claims corrected (commit `425a753c6`)

The reviewer could not break the refusal; all three findings were about claims.

**F1 — the EXTENDS half of the "graded fact" boundary graded nothing, and was only true of invalid Go.**
`field_type_refs.go` said the struct-anchored `DEPENDS_ON` *and* the embedded-field `EXTENDS` both still
treat a type parameter as a same-file type, and cited one test as making the pair a graded fact. Only the
`DEPENDS_ON` half is graded — **M13** (narrowing `extractStructFieldDependencies` at the call site) reds that
test on the `want [Base T]` row. The cited EXTENDS row embeds `Base`, a **real** type, so it grades
*"EXTENDS still emits for a real type"* and never touches the type-parameter claim; **M12**
(`&& !typeParams[base]` on the embedded-field gate) kills **0** tests.

Checked with the compiler rather than argued:

| embedded form | compiles? | emitted |
|---|---|---|
| `type Box[T any] struct { T }` | **no** — *embedded field type cannot be a (pointer to a) type parameter* | `Box -EXTENDS-> T` (the claimed over-fire, reachable only here) |
| `type Box[T any] struct { *T }` | **no**, same error | — |
| `type Box[T any] struct { Holder[T] }` | yes | **no EXTENDS at all** — `innermostTypeName` rejects the `[` whole |

**Decision: narrow the claim, do not manufacture a kill.** No legal, reachable input distinguishes M12, so
forcing it dead would require a fixture of Go that does not compile — which the standing rules forbid. The
header now scopes the graded half to `DEPENDS_ON`, states plainly that the EXTENDS path's behaviour on a type
parameter is **unreachable in compilable Go**, names both illegal forms and the legal one, and records that the
legal form's missing EXTENDS is a *separate pre-existing recall gap*, not this arm's wrong-binding.

**F3 — over-attribution (M10).** The SCOPE paragraph said the list is read as a DIRECT child *"so"* the scope is
exactly Go's. A whole-subtree `findAll` kills 0 tests; what delivers the scope is the **call site** handing over
one `type_spec`, and that IS graded (**M9**, file-global union, 6 leaves). Attribution corrected; no defect.

**F4 — no-op guard (M11).** `if len(kept) > 0` was a no-op: `attachGoFieldTypeRefs` deletes the key *before*
reading it and guards on `len(cands) == 0`, and grep over `internal` + `cmd` confirms it is the key's only
reader, called unconditionally from `extractor.go:202`. **Removed**, with the reason recorded in its place.

### Re-scored verbatim (own `cp` backups, exact match-count assertion, `go vet` before scoring, restore by `cp` + `shasum`, `git status --porcelain` empty after each)

| # | mutation | leaves | verdict |
|---|---|---|---|
| **M12** | `&& !typeParams[base]` on the embedded-field EXTENDS gate (`struct_fields.go`) | **0** — on `./internal/extractors/golang/ ./internal/resolve/` **and** on the wider `./internal/extractors/... ./internal/extractor/...` | ***alive but not production-reachable*** — distinguishing input `type Box[T any] struct { T }` beside `type T struct{}`, **rejected by the compiler** |
| **M13** | narrow `extractStructFieldDependencies` at the call site through `goTypeParameterNames(typeSpec, src)` | **1** | **DEAD** — `TestGoFieldTypeRefs_7041_EmbeddedFieldAndDependsOnAreUntouched` |

Both reproduce the reviewer's numbers exactly.

## 7. Verification

- `go test -count=1 ./internal/extractors/golang/ ./internal/resolve/` — green
- `go test -count=1 ./internal/extractors/... ./internal/extractor/...` — green
- `GRAFEL_HOME=$(mktemp -d) GRAFEL_DAEMON_ROOT=$(mktemp -d) go test -count=1 ./cmd/grafel/` — green, digest unmoved
- `go vet` green on every mutant before scoring

Refs #6912
Closes #7041


cajasmota (commented):
## REQUEST-CHANGES — the refusal itself is correct in both directions; two *claims* are not

Independent review, own worktree (`review-7066` @ `fb52aed5f`), own CST dump, own mutants. **I could not break the code.** Every attack I built — including eleven forms not in either table — behaves correctly, and the author's mutant table reproduces *exactly* (16/16/3/18/18/7/14 leaves, same test names). That is a better result than the last three arms.

The two changes I am asking for are prose, not code. I am flagging them because this PR's own thesis is #7056 — a claim nothing measures is invisible — and one of them is the sentence that says "the boundary is a graded fact rather than a claim".

---

### F1 — the EXTENDS half of the "graded fact" boundary is **not graded**, and the behaviour it describes is unreachable in compilable Go

`field_type_refs.go:199-206`:

> STILL OUT OF SCOPE, deliberately: the STRUCT-ANCHORED DEPENDS_ON (`extractStructFieldDependencies`) **and the embedded-field EXTENDS** both resolve a bare name against knownTypeNames and so still treat a type parameter as a same-file type. … **The boundary is a graded fact rather than a claim**: see `TestGoFieldTypeRefs_7041_EmbeddedFieldAndDependsOnAreUntouched`.

Half of that is true. The DEPENDS_ON half is genuinely graded — I narrowed `extractStructFieldDependencies` the same way #7041 narrowed the field edge (**M13**, filtering the returned rels through `goTypeParameterNames(typeSpec)`) and the named test goes **RED**, on the `want [Base T]` assertion. Good pin.

The EXTENDS half is not. **M12** — add `&& !typeParams[base]` to the embedded-field gate in `struct_fields.go:99`, i.e. make EXTENDS refuse type parameters exactly as this arm made the field edge refuse them — **kills 0 tests** across `./internal/extractors/golang/ ./internal/resolve/`. The cited test's EXTENDS row embeds `Base`, a real same-file type, so it grades *"EXTENDS still emits for a real type"* and never *"EXTENDS still treats a type parameter as a same-file type"*.

The test's own doc-comment is honest about why (`Embedding a type parameter is not legal Go … so the embedded row here uses a REAL same-file type`). The header comment is the one that promotes it to a graded fact, and it does so by citing a test that structurally cannot grade it.

Worse, I checked whether the claim is even *true*, and it is only true of code that does not compile:

| embedded form | legal Go? | what the extractor emits today |
|---|---|---|
| `type Box[T any] struct { T }` beside `type T struct{}` | **no** (*embedded field type cannot be a type parameter*) | `Box -EXTENDS-> T` — the claimed over-fire, reachable only here |
| `type Box[T any] struct { Holder[T] }` | yes | **no EXTENDS at all** — not to `Holder`, not to `T` |
| `type Box[T any] struct { *Holder[T] }` | yes | **no EXTENDS at all** |

So the sentence generalises an invalid-Go-only behaviour into a boundary statement, and the single legal way a type parameter reaches an embedded field produces a *different* pre-existing gap (generic instantiation loses its EXTENDS entirely) that the sentence does not describe. This is the narrowed-but-generous remnant the PR body says none of was left behind — on the one axis the PR explicitly named as separately pinned.

I am not prescribing the wording; **prescribing is not reviewing**, and I have not verified any fix. Either the claim gets scoped to what the test observes, or it gets a row that observes it.

---

### F2 — "Go is the last arm of #6912 that still shipped this known-wrong" is false

PR body, first line. C#, Rust and Java all still ship the identical over-fire, and unlike go's deleted `t.Skipf` pin, **C#'s and Rust's are hard `t.Fatalf` pins that assert the wrong edge is present**:

- `internal/extractors/csharp/field_type_refs_6912_test.go:352` — `TestCsharpFieldTypeRefs_KnownOverFire_TypeParameterShadowsSameFileType`, `class Box<Customer>` → `Box.Item -> …:Customer`, `t.Fatalf` if it changes shape.
- `internal/extractors/rust/field_type_refs_6912_test.go:593` — `TestRustFieldTypeRefs_KnownOverFire_TypeParameterShadowsSameFileType`, `struct Box2<Customer>` → `Box2.item -> …:Customer`, same shape.
- `internal/extractors/java/field_type_refs.go:175` — *"IT DOES NOT UNDERSTAND TYPE PARAMETERS … a case a real fix is expected to break"*, plus `field_type_refs_6912_test.go:450`.

#7041's own body already recorded this (*"arms A (C#), B (proto) and C (rust) have not been checked"*). The PR body converts an unchecked status into a settled one. Given that this PR exists because three comments asserted something no test observed, a status claim in its own body that four files contradict should not ship.

---

## Mutant table — my runs, not the author's

`cp` backup, exact match-count assertion before each edit (`AssertionError` if the count is not 1, so a silently-unapplied edit cannot read as ALIVE), `go vet` before scoring, `-count=1`, restore by `cp` + `shasum` + `git status --porcelain` empty after every single one.

| # | dir | mutation | leaves | verdict |
|---|---|---|---|---|
| M1 | permissive | delete the refusal at the stash site | 16 | **DEAD** |
| M2 | permissive | pass `nil` for `typeParams` at the call site | 16 | **DEAD** |
| M3 | permissive | `break` after the first `identifier` per declaration | 3 | **DEAD** (row 4 + the `Multi` metadata row) |
| M4 | permissive | `"type_parameter_declaration"` → `"type_parameter"` | 18 | **DEAD** |
| M5 | permissive | `"type_parameter_list"` → `"type_parameters"` | 18 | **DEAD** |
| M6 | **over-refusal** | also harvest `findAll(type_constraint, "type_identifier")` | 7 | **DEAD** (rows 6,7,8,10 + both metadata tests) |
| M7 | **over-refusal** | `typeParams[c]` → `len(typeParams) > 0` | 14 | **DEAD** (incl. the no-collision control) |
| **M9** *(mine)* | **over-refusal** | **file-global refusal set** — union every `type_spec`/`function_declaration`/`method_declaration`'s parameters and apply it to every declaration | **6** | **DEAD** — the `scope reach`, `generic function does not shadow the file` and predeclared-identifier rows all fire. The cross-declaration direction is genuinely graded. |
| M13 *(mine)* | over-refusal | narrow `extractStructFieldDependencies` the same way | 1 | **DEAD** — `EmbeddedFieldAndDependsOnAreUntouched` |
| M8 | — | `if len(out) == 0 { return nil }` → `return out` | 0 | *equivalent under any suite* — **I verified the "only consumer" claim myself** rather than accepting it: `goTypeParameterNames` has exactly one call site (`extractor.go:1902`), whose value reaches exactly one parameter (`struct_fields.go:57`), used at exactly one site (`typeParams[c]`). A lookup in a nil map and in an empty map are indistinguishable. Grep over `internal` + `cmd` finds no other reader. |
| **M12** *(mine)* | over-refusal | **EXTENDS refuses type parameters** | **0** | **distinguishable and ungraded → F1** (distinguishing input is invalid Go, so no behavioural defect — but the header's "graded fact" citation is false) |
| **M10** *(mine)* | — | direct-child scan → `findAll(decl, "type_parameter_list")[0]` | **0** | *alive; no production-reachable distinguishing input found* → **F3 (low)**. The doc says *"The list is looked up as a DIRECT child of the declaration node, **so** the scope is exactly Go's"*. The directness grades nothing; what delivers Go's scope is the call site handing it `typeSpec` (which M9 does grade). Over-attribution, not a defect. |
| **M11** *(mine)* | — | drop the `if len(kept) > 0` guard, always stash | **0** | *equivalent* → **F4 (informational)**. `attachGoFieldTypeRefs` deletes the key unconditionally and guards on `len(cands)==0`, so the new guard is a no-op. Harmless; just not one of the 8. |

---

## What I verified at source rather than on faith

**1. The three matcher strings.** Dumped the CST myself over ten forms. All three exist in this grammar version, spelled exactly as claimed. (`type_parameter_list` → `type_parameter_declaration` → `identifier` + `type_constraint`.) M4/M5 red 18 leaves each, so they are graded, not decorative. **No repeat of #7064.**

**2. "Only `identifier` children DIRECTLY under a `type_parameter_declaration`."** In every form I could construct the constraint is *always* wrapped in `type_constraint`; I never produced a constraint that surfaces as a bare `identifier` sibling of the parameter name. Including the nastiest nesting I could build, `[T interface{ ~[]E }, E any]`:

```
type_parameter_declaration  "T interface{ ~[]E }"
  identifier                "T"
  type_constraint           "interface{ ~[]E }"
    interface_type → type_elem → negated_type → slice_type → type_identifier "E"
```

`E` is four levels down and is a *different declaration's* parameter — collected from its own `type_parameter_declaration`, never harvested from `T`'s constraint. The load-bearing half holds.

**3. `[K, V any]`.** Dump confirms one `type_parameter_declaration` with two `identifier` children; both collected; M3 reds.

**4. The receiver.** Dump confirms `func (b Box[T]) Do` → `parameter_declaration → generic_type → type_arguments`, no `type_parameter_list`. Claim correct.

**5. Cross-declaration scope** — the item you flagged as most valuable. Correct, and **graded**: M9 (file-global union) dies on 6 leaves.

**6. The ERROR-node incident.** Reproduced. `type Al[T any] = Box[T]` parses as a `type_spec` with an `ERROR "="` *sibling*; the `type_parameter_list` is intact and read correctly, and there is no "first identifier" scan to go wrong. Two things I checked that the PR did not mention:
- the ERROR does **not** swallow the following declaration — `type After struct { Val T }` after it still emits `After.Val => T` correctly;
- the *reachable* variant `type Al[T any] = struct { Item T; Ok Ctl }` does produce a struct body and field entities (owner `Al`), and **the refusal works there**: `Al.Item` refused, `Al.Ok => Ctl` kept.

**7. Eleven forms outside both tables**, all correct in both directions: grouped `type ( … )` block; multi-name field `A, B T` (both refused) beside `C, D Ctl` (both kept); `json:"…"` wire-renamed field; nested anonymous struct inside a generic struct; blank `_` parameter; a parameter named after **its own owner** (`type Box[Box any]` — `Box.Item` refused, sibling `Sib.B => Box` still binds); constraint-collides-but-parameter-does-not; generic receiver method in a file with a colliding `type T`.

**8. The corpus zero — independently re-derived, and it is real.** File counts match to the unit (etcd 391 + gin 99 + chi 74 + grpc-go-examples 118 = 682, no vendor skew), and the generic-declaration grep lands on **exactly 8**, all in one file (`grpc-go-examples/stream_interfaces.go`). The two generic structs are `GenericServerStream` / `GenericClientStream`, whose only member is an **embedded** `ServerStream`/`ClientStream` — no named field, hence no anchor, hence 0 movement. Nothing upstream is hiding generics here (cf. #7060); these repos simply predate them. Per #7041's own instruction, that is not a bound on the language.

**9. The deleted pin and the rewritten prose.** The replacement is real — each absent edge sits in a fixture that also asserts a present one, which the old `t.Skipf` structurally could not do, and M7 confirms the no-collision control (row 17) is load-bearing. Grep for `expected to break` / `DOES NOT UNDERSTAND TYPE PARAMETERS` finds no go remnant. The only surviving hits are C#, Rust and Java — which is F2.

**10. Tables, VARIED vs HELD-CONSTANT.** Both tables carry the block in-source and both are honest about it. The metadata table's closing drift loop (*"type_spec %q is in the fixture but not in the table"*) is real and does fail on drift. Held-constant axes I attacked independently: declaring form (probed generic alias-with-struct-body, grouped block), position (already swept, rows 1–2), target kind (row 8 is a `type_alias`), anchor (F1 — half graded), file (cross-file unchanged).

## Verification

- `go test -count=1 ./internal/extractors/... ./internal/extractor/... ./internal/resolve/...` — exit 0
- `GRAFEL_HOME=$(mktemp -d) GRAFEL_DAEMON_ROOT=$(mktemp -d) go test -count=1 ./cmd/grafel/` — exit 0, 182s, **digest verified unmoved by running it, not by assuming it should be**
- `go vet` green on every mutant before scoring; `git status --porcelain` empty after each restore; final `shasum` on all three touched source files identical to HEAD

Nothing about the refusal logic blocks this. Fix the two claims and it is an approve from me.

cajasmota:
## Review round 2 — all four findings addressed in `425a753c6`. No behaviour changed except one no-op guard.

Thank you for the mutants; M12 and M13 both reproduce exactly, and M12 is the one that mattered.

### F1 — decision: **narrow the claim**, and record M12 as unreachable rather than kill it

You are right in both halves. **M13 DEAD (1 leaf)**, **M12 ALIVE (0 leaves)** — and I widened M12's scoring
beyond your two packages to `./internal/extractors/... ./internal/extractor/...`, still **0**. So nothing
anywhere grades "EXTENDS treats a type parameter as a same-file type".

I then asked whether it should be graded, and the answer is no, because no legal input reaches it. Checked
with the compiler rather than from the spec:

```
a.go:7:2:  embedded field type cannot be a (pointer to a) type parameter   // struct { T }
a.go:11:2: embedded field type cannot be a (pointer to a) type parameter   // struct { *T }
```

`struct { Holder[T] }` compiles and emits **no EXTENDS at all** — `innermostTypeName` returns `""` on the
first non-alphanumeric byte, so the `[` kills it before `knownTypeNames` is consulted. Your table is correct
to the byte.

So killing M12 would need a fixture of Go that does not compile, which the standing rules forbid, and which
would pin a behaviour no index can ever reach. **M12 is recorded as *alive but not production-reachable*,
with the distinguishing input named** (`type Box[T any] struct { T }` beside `type T struct{}`). The header
comment now:

- scopes the graded half to the struct-anchored `DEPENDS_ON` only, and says *what* grades it (the full-output
  `[Base T]` assertion, so a narrowing in either direction shows up);
- states that EXTENDS cannot make this mistake in compilable Go, naming both illegal forms **and** the legal
  one;
- records that `struct { Holder[T] }` losing its EXTENDS is a **separate pre-existing recall gap**, not this
  arm's wrong-binding — the thing the old sentence silently generalised over;
- records the portable reasoning error: *a mutant with no legal distinguishing input must be recorded as
  unreachable, never killed with a manufactured fixture.*

Your framing is the right one — a test that pins one half of a two-part claim makes the other half look
covered. That is exactly what happened, on the one axis the body advertised as separately pinned.

### F2 — the first line was false, and you undersold it by one

Verified all three at source on this branch rather than relaying, and found one correction **in the stricter
direction**: **Java's pin is also a hard `t.Fatalf`**, not prose-only. `javaFTWantEqual`
(`field_type_refs_6912_test.go:167`) fatals on mismatch, so
`TestJavaFieldTypeRefs_KnownOverFire_TypeParameterShadowedByASameFileType` (`:450`) is a real gate as much as
C#'s and Rust's.

| arm | over-fires | at source |
|---|---|---|
| C# | yes | `csharp/field_type_refs_6912_test.go:352`, `class Box<Customer>` → `Box.Item -> …:Customer`, **`t.Fatalf`** |
| Rust | yes | `rust/field_type_refs_6912_test.go:593`, `struct Box2<Customer>` → `Box2.item -> …:Customer`, **`t.Fatalf`** |
| Java | yes | prose at `java/field_type_refs.go:175` **and** the pin at `:450` — **`t.Fatalf` via the helper** |
| F#, cpp, swift, go | no | `'`-token skipped whole; template members emit no field entity (#7057); `…GetsNoEdge`; this PR |
| php, proto | n/a | no generics in field-type position |

**Three arms still ship it**, and all three pins are gates a fix would break — unlike go's deleted pin, which
`t.Skipf`'d in both directions and graded nothing. The body's first line is rewritten to say this, with the
table, and it names its own error rather than quietly replacing the sentence. Reported upward so #7041 and
#6912 can be corrected too.

### F3 — attribution fixed

The directness grades nothing; the **call site** handing over one `type_spec` is what delivers Go's scope, and
M9 grades it. The paragraph now says that, and says explicitly that a whole-subtree `findAll` kills no test
because a `type_parameter_list` cannot nest inside another declaration's `type_spec` here — tidiness, not
guarantee.

### F4 — guard removed

Confirmed a no-op independently: `attachGoFieldTypeRefs` is the key's only reader (grep over `internal` +
`cmd`), is called unconditionally at `extractor.go:202`, and `delete`s the key **before** the `len(cands) == 0`
check — so an empty stash can neither be observed nor leave the package. Removed, with the reason recorded in
its place so it does not get re-added.

### Verification

- `go test -count=1 ./internal/extractors/... ./internal/extractor/... ./internal/resolve/...` — exit 0
- `GRAFEL_HOME=$(mktemp -d) GRAFEL_DAEMON_ROOT=$(mktemp -d) go test -count=1 ./cmd/grafel/` — exit 0, 174s,
  **digest unmoved** (run, not assumed — the guard removal is a real code change)
- `go vet` green on both mutants before scoring; restore by `cp` + `shasum` identical to HEAD;
  `git status --porcelain` clean of the mutant after each

PR #7069, opened 2026-09-16
fix(#7041): arm java — a type parameter shadows a same-file type, and javac says it shadows even where the reference is illegal

cajasmota:
Closes the **java** arm of #7041. The other two live arms (C#, rust) are untouched.

`class Holder<T> { T item; }` in a file that also declares `class T {}` emitted a REFERENCES edge asserting `item`'s declared type was that class. It is not — `T` is the type parameter, which shadows the outer declaration. The edge **bound**, which is the direction #7056 established no instrument we own can see.

---

## 1. CST evidence for the node names (trap (a): verify, do not recall)

Dumped from `tsjava.Language()` via the same parser the tests use, with `Child()` (so anonymous tokens are visible), over the whole form space. The spellings this fix matches on, verbatim from the dump:

```
class_declaration  "class Bounded<T extends Order> { T t; }"
  class  "class"
  identifier  "Bounded"
  type_parameters  "<T extends Order>"
    <  "<"
    type_parameter  "T extends Order"
      type_identifier  "T"                 <-- the NAME: a DIRECT child
      type_bound  "extends Order"          <-- the BOUND: a SIBLING subtree
        extends  "extends"
        type_identifier  "Order"
    >  ">"
```

and the three other shapes the name-extraction has to survive:

```
type_parameter "T extends A & B"       -> [type_identifier T, type_bound[extends, tid A, &, tid B]]
type_parameter "T extends Comparable<T>" -> [type_identifier T,
                                             type_bound[extends, generic_type[tid Comparable,
                                               type_arguments[tid T]]]]
type_parameter "@NonNull T"            -> [marker_annotation[@, identifier NonNull], type_identifier T]
```

The list node is `type_parameters`, its members are `type_parameter`, and the name is a **direct** `type_identifier` child. Bounds and annotations are sibling subtrees, so a descendant walk would harvest `Order`, `A`, `B`, `Comparable` as shadowed names and silently delete correct edges. Those are exactly three matcher strings, and **each was corrupted and scored** — M4, M5, M6 in the table below, all DEAD. There is **no declaration-kind list** in this fix (no `class_declaration`/`interface_declaration`/… enumeration) precisely so there is nothing that can go stale the way the scala arm's two non-existent node types did.

`type_parameters` also appears on `method_declaration`, but that is **not** matched-and-dead — it is simply never on a field's ancestor chain, because no field entity is emitted inside a method body (probe: a local class `class Local { T f; }` in a method yields no `SCOPE.Schema/field` record and no component at all). Stated as a limit in the source rather than modelled.

## 2. The JLS rule, verified with javac 25.0.3 — and why Java's is NOT kotlin's or scala's

The brief's expectation (and kotlin's implemented rule) was: an inner class captures the outer's type parameters, a `static` nested class does not, so `T` inside a static nested class really does name the same-file `class T` and the edge must be **kept**. **That is wrong for Java.** Each form below was compiled in isolation, with a top-level type of the parameter's name present:

| # | source | javac 25.0.3 |
|---|---|---|
| 1 | `class Box<P> { class InnerC { P f; } }` | **rc=0** — inner class captures |
| 2 | `class Box<P> { static class NestedC { P f; } }` | rc=1 `non-static type variable P cannot be referenced from a static context` |
| 3 | `class Box<P> { record R(P x) {} }` | rc=1 same error (nested records are implicitly static, JLS 8.10.1) |
| 4 | `interface Box<P> { class C { P f; } }` | rc=1 same error (member types of an interface are implicitly static, JLS 9.5) |
| 5 | `record Rec<P>(int n) { class InnerC { P f; } }` | **rc=0** — an inner class of a generic **record** captures |
| 6 | `class Box<P> { class Mid { static class DeepStatic { P g; } } }` | rc=1 same error |
| 7 | `class Box<P> { static P f; }` | rc=1 same error (static member of C, JLS 8.1.2) |
| 8 | `class P {} class Box<P> { static class N { P f = new P(); } }` | rc=1 — **twice**, at the type and at the constructor |
| 9 | `class P5 {} class S { static class Box<P5> { P5 in; } static class Other { P5 out = new P5(); } }` | **rc=0** — a SIBLING outside the generic class uses the top-level type fine |
| 10 | `enum E<P> { A }` | rc=1 `enums cannot be generic` |
| 11 | `@interface A<P> { String value(); }` | rc=1 `annotation interface cannot be generic` |
| 12 | `class Q {} class S { static class Box<P> { class InnerC<Q> { Q f; } } }` | **rc=0** — an inner class may introduce its own list |

Row 8 is the decisive one. JLS 8.1.2 scopes a class type parameter over the **whole body** of its declaration and JLS 6.4.1 makes that a **shadow**; the static-context restriction makes the *reference* illegal, it does **not** reopen the outer name. javac does not fall back to `class P` — it errors. So:

* **kotlin** ascends only through `inner` declarations, because a plain nested Kotlin class genuinely does not see the outer parameter and `T` there really does name the same-file type;
* **scala** has no static nesting and always captures;
* **java** has both nesting forms, but the shadow is not conditioned on staticness at all — only the legality of the reference is.

**Implemented rule: ascend unconditionally through every enclosing declaration carrying a `type_parameters` list.** No `static` check, no implicitly-static carve-out. Over-refusing in rows 2/3/4/6/7 deletes no correct edge because **there is no valid program in which those names denote the same-file type**. Gating on `static` — the "obvious" port of kotlin's rule — would have *kept a wrong binding* on input that does not compile, for a rule Java does not have. Rows 1, 5, 9 and 12 are the compiling side and are the over-refusal controls.

## 3. The enumerated form space, both directions, with VARIED / HELD-CONSTANT per table

Five tables in `internal/extractors/java/field_type_refs_7041_test.go`. **Every fixture declares a real same-file type whose field edge must be present in the same equality assertion** — that is the deleted pin's positive-control idea, applied per-fixture rather than once.

### Table 1 — `..._TypeParameterFormSpace` (`Forms.java`, javac rc=0)

**VARIED:** list arity (1, 2); a bound (`extends Order`); an intersection bound (`extends A & B`); a recursive bound (`extends Comparable<T>`); a `TYPE_PARAMETER` annotation; the use position (bare, array `T[]`, generic argument `List<T>`, wildcard `List<? extends T>`, qualified-generic `java.util.List<? extends Order>`); and the no-parameter case.
**HELD CONSTANT:** declaring form `class`; anchor = class `field_declaration`; one file; colliding name `T` (except where arity forces `K`/`V`); control type `Order`.

| form | refused (no edge) | control that must remain |
|---|---|---|
| `class Box<T>` | `Box.item` | `Box.ok => Order` |
| `class Pair<K, V>` | `Pair.key`, `Pair.val` | `Pair.ok => Order` |
| `class Bounded<T extends Order>` | `Bounded.t` | `Bounded.ok => Order` (the **bound** stays addressable) |
| `class Multi<T extends A & B>` | `Multi.t` | `Multi.a => A`, `Multi.b => B` |
| `class Recur<T extends Comparable<T>>` | `Recur.t` | `Recur.ok => Order` |
| `class Anno<@NonNull T>` | `Anno.t` | `Anno.ok => Order` |
| `class Wrapped<T>` | `arr` (`T[]`), `list` (`List<T>`), `wild` (`List<? extends T>`) | `okWild => Order` |
| `class Plain` (no params) | — | **`Plain.t => T`** — the over-refusal control for the whole table |

`class T`, `class K`, `class V`, `class A`, `interface B` are all real same-file declarations, so every refused row would have bound before this change.

### Table 2 — `..._DeclaringFormSpace` (`Decls.java`, javac rc=0)

**VARIED:** the form carrying the parameters *and the anchor consuming them* — a `class` with a `field_declaration`, and a `record` with a **header component** (walk's `record_declaration` arm is a separate emit site; a fix wired only into `buildField` passes table 1 whole).
**HELD CONSTANT:** one parameter `TD`, one control `OrderD`, one file.

`BoxD.item` and `RecD.item` refused; `BoxD.ok`/`RecD.ok => OrderD` kept. The forms that **cannot** appear are stated rather than omitted: `enum` and `@interface` cannot be generic at all (javac rows 10/11), and a generic interface's constant emits no field entity (the recall ceiling already documented in the 6912 test header), so `IFD.CONST` yields no row in either direction — asserted by its absence from the equality.

### Table 3 — `..._NestingFormSpace_Compiling` (`NestOk.java`, javac rc=0)

**VARIED:** where the field sits relative to the binding declaration — the declaring class itself; an inner class; an **anonymous class in an instance-field initializer** (which this extractor *does* emit fields for, with bare names); an inner class that **redeclares** the name; an inner class with its **own, differently-named** list that still uses the outer's; a generic inner class of a non-generic outer; a non-generic inner class of a non-generic outer; and a **sibling** outside the generic class.
**HELD CONSTANT:** parameter name `TN`, control `OrderN`, one file, `class` throughout.

Refused: `OuterN.own`, `af` (anonymous), `InnerN.it`, `RedeclareN.rt`, `BothN.bt` + `BothN.bu`, `InnerGenN.gt`.
Controls kept: `OuterN.ok`, `aok`, `InnerN.iok`, `RedeclareN.rok`, `BothN.bok`, `InnerGenN.gok`, and the two that stop the ascent from over-refusing — **`InnerFlatN.ft => TN`** and **`SiblingN.st => TN`**, both outside any declaration binding `TN`.

### Table 4 — `..._NestingFormSpace_StaticContext` (`NestStatic.java`, javac rc=1 — deliberately)

**VARIED:** the static-context forms — explicit `static` nested class, nested record, member class of a generic interface, static class nested two levels down.
**HELD CONSTANT:** parameter name `TS`, control `OrderS`, one file.

This source does **not** compile, and that is the finding (javac rows 2/3/4/6 above). The table pins that the extractor's answer on such input is *refusal* rather than a wrong binding, and the four `OrderS` control rows prove the producer still runs inside every one of these forms — so a missing row means refusal, not silence.

### Table 5 — `..._CollidingDeclarationComesLast` (`Later.java`, javac rc=0)

**VARIED:** the **position of the colliding declaration**, which every table above holds constant at "before the generic one". `class TP {}` is declared *after* both the generic class that shadows it and the plain class that legitimately references it. This is the axis the stash-then-attach two-pass design exists for.
**HELD CONSTANT:** single parameter, class anchor, control `OrderP`.

`BoxP.item` refused; `PlainP.t => TP` and both `OrderP` controls kept.

## 4. Mutant table

Byte-level `cp` backup, `go vet` before scoring, `-count=1`, edit-landed asserted by shasum change, restored by `cp` with a shasum re-check on every row. Permissive direction first. The call site mutated is `stashJavaFieldTypeRefs` (M1), not only the helper.

| # | mutant | direction | verdict |
|---|---|---|---|
| M1 | call site: `javaFieldTypeCandidates(typ, src, nil)` — shadow set never computed | permissive | **DEAD** — all 5 tables |
| M2 | ascent returns as soon as the nearest list is non-empty (kotlin's "immediate list only" defect) | permissive | **ALIVE on first score → GRADED → DEAD.** See below |
| M3 | collect `type_identifier` **descendants** of `type_parameter` instead of direct children (harvests bounds) | over-refusing | **DEAD** — `Bounded.ok`, `Multi.a`, `Multi.b` |
| M4 | `"type_parameters"` → `"type_parametersZZ"` | permissive | **DEAD** — all 5 tables |
| M5 | `"type_parameter"` → `"type_parameterZZ"` | permissive | **DEAD** — all 5 tables |
| M6 | name child `"type_identifier"` → `"identifier"` | permissive | **DEAD** — all 5 tables |
| M7 | shadow set built by descending from the file ROOT (refuse a name file-wide) | over-refusing | **DEAD** — `Plain.t`, `InnerFlatN.ft`, `SiblingN.st`, `PlainP.t` |
| M8 | `decl.Child(i)` → `decl.NamedChild(i)` in the list lookup | — | **ALIVE — equivalent.** `type_parameters` is a *named* node, so the named-child view of a declaration always contains it; the mutant is behaviourally identical. Positive evidence rather than assertion: had it not been named, this mutant would have gone dark on every fixture exactly as M4 did, and M4 is DEAD on all five tables |
| M9 | collect only the FIRST parameter of a list | permissive | **DEAD** — `Pair.val => V` reappears |

**M2 is the trap-(c) finding and it was real.** Table 3 already varied nesting thoroughly, and the `RedeclareN<TN>` row *looks* like it grades the ascent — it does not, because the immediate list there already carries `TN`, so stopping at the first non-empty list passes it. The distinguishing input is an inner class with its **own, differently-named** list that still uses the outer's (`class OuterN<TN> { class BothN<UN> { TN bt; UN bu; … } }`, javac rc=0). It was added as `BothN` and M2 re-scored **DEAD**. One axis enumerated thoroughly made its sibling look covered, again.

## 5. Corpus delta

645 `.java` files across `archigraph-corpora` (kafka 471, kafka-streams-examples 115, spring-petclinic 47, awesome-compose 11, ratpack-example-books 1) — the same 645 the existing header measures. Extracted, `resolve.BuildIndex` + `resolve.ReferencesEmbedded`, edges counted by `ref_kind=field_target_type`. "Before" was produced from this same branch with the fix disabled at the call site (M1), so only the fix varies.

|---|---|---|---|
| before | 147 | 147 | 0 |
| after | **147** | **147** | **0** |

**Zero changed rows** — the two full `file:field => target [state]` listings are **byte-identical** (`diff` exit 0). No row to name, in either direction, so nothing was dropped and nothing needs proving-wrong.

**The mechanism is not dead there, and that is measured rather than assumed.** Instrumenting the refusal branch and re-running the corpus: it fires **127 times** — `K`×56, `V`×52, `T`×9, `Y`×4, `X`×4, `E`×2. Real Java code in the corpus does declare fields typed by a type parameter; none of those 127 names happens to also be a same-file declared type, which is why the emitted set is unchanged. So the zero delta is a property of the corpus's naming conventions, not of an unreachable filter — and per #7041's own guidance, a corpus-relative zero is not a bound on the language.

## 6. The deleted pin and the `:175` prose

* **Deleted** `TestJavaFieldTypeRefs_KnownOverFire_TypeParameterShadowedByASameFileType` (`field_type_refs_6912_test.go:450`). It was a real `t.Fatalf` gate and this fix breaks it, which is what it was for. A short note stands where it was, pointing at the replacement.
* **Its positive-control idea is kept and strengthened.** The pin's `H2.java` control (no colliding `class T`, edge must be absent) distinguished "the over-fire is still here" from "the producer stopped entirely". Every fixture in the new file carries that distinction *in the same assertion*: each source declares a real same-file type whose edge must be **present**. A pass that went silent fails all five tables; a pass that over-refuses fails a **named control row** rather than silently deleting an edge.
* **The three prose claims are corrected, not left behind** (#7056's point):
  * `field_type_refs.go:175` — the "IT DOES NOT UNDERSTAND TYPE PARAMETERS … known-wrong and pinned" bullet is replaced by what the pass now does, and by the one thing still not modelled (a *method's* own type parameters) stated as a limit with the reason it is unreachable.
  * `field_type_refs_6912_test.go` header — "the type-parameter over-fire needs a file that DECLARES the parameter's name" now points at the new file.
  * the `Holder.item` row comment in `TestJavaFieldTypeRefs_EdgeSet` — that row is unchanged (that fixture declares no `class T`) but it is now refused for the right reason, and says so.

## 7. Verification

* `go test -count=1 ./internal/extractors/java/ ./internal/resolve/` — **ok**
* `GRAFEL_HOME=$(mktemp -d) GRAFEL_DAEMON_ROOT=$(mktemp -d) go test -count=1 ./cmd/grafel/` — **ok** (186.8s). **The digest did not move**: that fixture's Java members contain no generic declaration whose parameter name collides with a same-file type, so there is no delta to enumerate and no constant was touched.
* `gofmt -l internal cmd` — empty. `go vet ./internal/extractors/java/` — clean.
* Red-before-green: all five tables failed on `cba3819d7`'s behaviour with every over-fire row present (`Forms.java` 20 edges → 10, `NestOk.java` 14 → 10, `NestStatic.java` 8 → 4, `Decls.java` 4 → 2, `Later.java` 4 → 3), and every control row was already present before the fix, so the controls grade the new direction rather than the old one.

Refs #7041, #6912, #7056, #7066.


cajasmota (commented):
## Adversarial review — REQUEST-CHANGES

Independent worktree, `1e318e437`, nothing from the PR body taken on faith. The headline
language claim is **correct** and I re-derived it with my own `javac 25.0.3`. The matcher
strings are **correct** and I dumped the CST myself. The M8 equivalence is **real** and I
have direct evidence for it. The corpus number **reproduces**. The digest **is** unmoved.

One finding blocks: **the over-refusal direction is graded at one of the two anchors and at
zero of the other**, and I have a compiling input plus a mutant that proves it.

---

### Finding 1 (blocking) — the record-component anchor has no over-refusal control, and the PR's central claim about that is false there

The PR states, twice (body §6, and the header of `field_type_refs_7041_test.go`):

> a pass that over-refuses fails a **named control row** rather than silently deleting an edge

That is true at the **class `field_declaration`** anchor. It is **false at the record header
component** anchor, which is the second, separate emit site (`java.go:492`) this fix exists
to cover.

Every row that grades the *scoping* of the shadow set — i.e. a name that IS a type parameter
somewhere in the file whose edge must nonetheless be **kept** — sits on a class field:

* `Forms.java:Plain.t => T`
* `NestOk.java:InnerFlatN.ft => TN`
* `NestOk.java:SiblingN.st => TN`
* `Later.java:PlainP.t => TP`

There is **no such row on a record component anywhere in the new file.** The record-anchor
controls are `Decls.java:RecD.ok => OrderD` and `NestStatic.java:NRecS.nrok => OrderS`, and
`OrderD`/`OrderS` are never a type parameter in their files. Those rows grade *producer
liveness* (the deleted pin's `H2.java` idea) — they do **not** grade whether the shadow set
is scoped to the field's position. Different property; the PR conflates them.

**Proof, by the symmetry of one mutation.** I applied the *identical* mutant at each anchor —
"compute the shadow set by descending from the file root instead of ascending from this
field", i.e. refuse a name file-wide once seen anywhere:

| mutant | anchor gated on | verdict |
|---|---|---|
| ME | `field_declaration` (class field) | **DEAD** — fails `Plain.t`, `InnerFlatN.ft`, `SiblingN.st`, `PlainP.t` |
| MA | `formal_parameter` (record component) | ***ALIVE*** — full `./internal/extractors/java/` **and** `./internal/resolve/` pass, `go vet` clean |

Same edit, same defect, one anchor kills it and the other cannot see it.

**Distinguishing input** (`javac 25.0.3`, **rc=0**):

```java
class OrderR {}
class TR {}
class GenR<TR> { TR inner; OrderR gok; }
record PlainR(TR t, OrderR ok) {}
class PlainC { TR t; OrderR ok; }
```

* HEAD (correct): emits `PlainR.t => TR` **and** `PlainC.t => TR`.
* Under MA: `PlainC.t => TR` survives, **`PlainR.t => TR` is silently deleted** — a *correct*
  edge, in the direction #7056 established no instrument we own can see, and the whole suite
  stays green.

This is trap (c) again, on the axis this PR's own table 2 was written to attack. Table 2
varies the anchor **for the refusal direction** (`RecD.item`, `NRecS.nr`) and holds it
constant **for the over-refusal direction**. Exactly the shape of M2/`RedeclareN`: one axis
enumerated thoroughly made its neighbour look covered.

*(Prescribing is not reviewing — but the gap is one row: give a record in one of the existing
fixtures a component whose type is a name bound as a type parameter elsewhere in that file,
outside the record's own ancestry, and assert the edge is kept. `Later.java` is the cheapest
home, and it also picks up the position axis for free.)*

Same shape, **not** claimed as a separate finding because I could not build a non-contrived
mutant for it: the **anonymous-class** field anchor also has no scoping control (`aok =>
OrderN` uses a name that is never a parameter). I verified by hand that HEAD gets it right —
`class C { I a = new I() { P f; Ok ok; }; }` with a top-level `class P` correctly keeps
`f => P` — so this is an observation, not a defect.

---

### Finding 2 (non-blocking) — prose asserting something the code contradicts

`field_type_refs.go` (the rewritten `:175` bullet):

> What is still **NOT modelled** … a METHOD's own type parameters (`<T> void f(T x)`) …
> **Modelling them would be unreachable code.**

Both halves are wrong about the code, in opposite directions. Because the ascent deliberately
carries **no declaration-kind list** and matches on `type_parameters` alone, it *does* collect
a method's and a constructor's parameters — I enumerated every node type in the grammar that
carries that child:

```
class_declaration  constructor_declaration  interface_declaration
method_declaration  record_declaration
```

So method type parameters **are** modelled; what is true is that they are **unreachable**, and
the code already contains that unreachable behaviour rather than omitting it. (I did confirm
the reachability premise independently: `class C { <P> void m() { class L { P f; } } }` and the
same with an anonymous class in a method body both emit **zero** field entities. The *fact* is
right; the sentence describing the code is not.) Suggest: "a method's own type parameters are
collected by the same ascent but can never fire, because …".

---

### Finding 3 (nit) — table 4's VARIED list names five forms; the fixture has four

`..._NestingFormSpace_StaticContext`'s comment claims it varies

> … **a static class nested two levels down**, and a nested class inside a static nested class.

`NestStatic.java` contains `static class DeepS { class DeeperS { … } }` — that is the *second*
of those (static-then-inner). The *first* (inner-then-static, the `class Mid { static class
DeepStatic { P g; } }` shape that is javac row 6 in your own evidence table) is **not in the
fixture**. Behaviourally both are refused by the union ascent so I do not believe it is
distinguishable, but the label claims more than the content — the repeat offender here.

---

## What I verified independently, and it all held

**Claim 1 — the headline language claim. Confirmed, with my own compiler.** I recompiled your
whole table plus new boundary forms. `javac` does **not** fall back to the top-level type; it
errors. The unconditional ascent is right, and gating on `static` would have kept a wrong
binding.

| form | my `javac 25.0.3` |
|---|---|
| `class P {} class Box<P> { class InnerC { P f; } }` | **rc=0** |
| `class P {} class Box<P> { class InnerC { P f = new P(); } }` | rc=1 `unexpected type` — **positive proof `P` denotes the variable, not `class P`** |
| `class P {} class Box { class InnerC { P f = new P(); } }` | **rc=0** — control, no generic ⇒ legal |
| `class P {} class Box<P> { static class N { P f = new P(); } }` | rc=1 `non-static type variable P cannot be referenced from a static context` ×2 |
| `class P {} class Box<P> { record R(P x) {} }` | rc=1 same |
| `class P {} interface Box<P> { class C { P f = new P(); } }` | rc=1 same |
| `class P {} class Box<P> { class Mid { static class Deep { P g = new P(); } } }` | rc=1 same |
| `class P {} class Box<P> { static P f = new P(); }` | rc=1 same |
| `class P {} record Rec<P>(int n) { class InnerC { P f; } }` | **rc=0** |
| `class P {} class S { static class Box<P> { P in; } static class Other { P out = new P(); } }` | **rc=0** |
| `enum E<P> { A }` / `@interface A2<P> { … }` | rc=1 cannot be generic |

Boundary forms you did **not** list, all consistent with the implemented rule:

| form | `javac` | note |
|---|---|---|
| `class C { <P> void m() { class L { P f; } } }` | rc=0 | generic **method** param shadows into a local class (`new P()` ⇒ `unexpected type`) |
| `class Box<P> { static void m() { class L { P f; } } }` | rc=1 | static method ⇒ static context |
| `class Box<P> { void m() { class L { P f; } } }` | rc=0 | instance method captures |
| `class Box<P> { I a = new I() { P f; }; }` | rc=0 | anonymous in **instance** field init |
| `class Box<P> { static I a = new I() { P f; }; }` | rc=1 | anonymous in **static** field init |
| `class Box<P> { class Inner<P> { P f; } }` | rc=0 | redeclaration legal |
| `class Box<P> { static class N<P> { P f; } }` | rc=0 | static nested may rebind the name |
| `class Box<T> { static class N { P f = new P(); } }` | **rc=0** | name not bound ⇒ top-level `class P`; the over-refusal boundary |
| `interface Box<P> { interface Inner { P f2(); } }` | rc=1 | implicitly static |
| `class Box<P> { enum E { A; P f; } }` | rc=1 | implicitly static |
| `interface Box<P> { class C<P> { P f; } }` | rc=0 | own param wins |

And the structural reason the unconditional ascent is *sound* rather than lucky: every one of
the five node types that carries `type_parameters` scopes those parameters over exactly its
own subtree, so the lexical parent chain **is** the scope chain. That is a stronger argument
than the one in the doc comment and worth putting there.

**Claim 3 — matcher strings. Confirmed, dumped myself,** including two forms you did not show
(`<@NonNull X extends Comparable<X> & java.io.Serializable>`, `<Z extends java.util.List<String>>`).
The name is a direct `type_identifier` child in every case; `type_bound`, `marker_annotation`,
`generic_type`, `type_arguments` and `scoped_type_identifier` are all siblings or deeper.
My own M3 (descendant walk) is **DEAD** — killed by `Bounded.ok`, `Multi.a`, `Multi.b`. The
over-refusal direction of the harvesting risk is properly graded.

**Claim 2 / M2 — the trap-(c) fix is load-bearing, not decorative.** I re-scored the "stop at
the nearest non-empty list" mutant: **DEAD**, and the diff names exactly one row —
`NestOk.java:BothN.bt => TN`, the row you added. Without it that mutant ships.

**Claim 7 / M8 — equivalence confirmed by evidence, not by the argument you gave.** I checked
namedness directly: `type_parameters` is a **named** child under all five carriers
(`class_declaration`, `constructor_declaration`, `interface_declaration`,
`method_declaration`, `record_declaration`). `Child` and `NamedChild` both reach it. Genuinely
equivalent.

**Claim 4 — the deleted pin's positive control.** Confirmed per-fixture at the class anchor: a
pass that goes silent fails all five tables, and a pass that over-refuses fails a *named* row
(that is what makes ME **DEAD**). The claim fails only at the record anchor — Finding 1.

**Claim 5 — the corpus instrumentation. Re-derived independently and it reproduces.** 645
`.java` files, restricted to the two real anchors: I count **131** refusals —
`K×58, V×54, T×9, X×4, Y×4, E×2` — against your **127** / `K×56, V×52, T×9, Y×4, X×4, E×2`.
Identical name set, identical shape, ±4 on `K`/`V`, which I attribute to my anchor predicate
being slightly broader than the production walk's, not to the guard. Unrestricted (all
`formal_parameter`s) it is 570. **The guard is emphatically not dead.** I also tested the
zero-delta claim head-on rather than by diffing listings: of every refused name, the number
that is **also a type declared in the same file** is **0**. So the byte-identical listings are
a corpus naming property, as you say — and a corpus-relative zero remains no bound on the
language.

**Claim 6 — the three prose corrections.** All three are corrected in place, none orphaned:
`field_type_refs.go:175`, the 6912 header, and the `Holder.item` row comment. (The `:175`
replacement introduces Finding 2.) The deleted pin leaves a pointer rather than a hole.

**Fixture labels vs content — compile-checked, all five.** `Forms.java` rc=0, `Decls.java`
rc=0, `NestOk.java` rc=0, `Later.java` rc=0, `NestStatic.java` **rc=1 with exactly four
errors, one per form** — deliberate and correctly labelled.

**Suites, on a clean tree at `1e318e437`:**
* `go test -count=1 ./internal/extractors/java/ ./internal/resolve/` — **ok**
* `GRAFEL_HOME=$(mktemp -d) GRAFEL_DAEMON_ROOT=$(mktemp -d) go test -count=1 ./cmd/grafel/` —
  **exit 0**. The digest is genuinely unmoved; I did not take that on faith.
* `gofmt -l internal cmd` — empty. `go vet ./internal/extractors/java/` — clean.
* Every mutant restored by `cp`; `shasum` back to `b4786ef7…` and `git status --porcelain`
  empty after each.

---

### Mutant table

| # | mutant | direction | verdict |
|---|---|---|---|
| MA | record anchor: shadow set descended from the file ROOT | over-refusing | ***ALIVE — distinguishable and ungraded → FINDING 1*** |
| ME | class-field anchor: same mutation | over-refusing | **DEAD** — `Plain.t`, `InnerFlatN.ft`, `SiblingN.st`, `PlainP.t` |
| MB | ascent stops at the nearest non-empty list (your M2) | permissive | **DEAD** — `NestOk.java:BothN.bt => TN`, and only that row |
| MC | `type_identifier` **descendants** of `type_parameter` (your M3) | over-refusing | **DEAD** — `Bounded.ok`, `Multi.a`, `Multi.b` |
| MD | collect only the FIRST parameter of a list (your M9) | permissive | **DEAD** |
| M8 | `Child` → `NamedChild` | — | **equivalent under the current suite** — verified by direct namedness probe, not by inference |

Everything about this arm's *reasoning* is sound — the language claim is right, contradicts
the brief for the right reason, and is backed by evidence that reproduces. The gap is that the
anchor axis is graded in one direction only, and the PR asserts otherwise in two places.
Close Finding 1 and this is a strong merge.

cajasmota:
## Review delta — `4e733bcdc`

Finding 1 accepted in full. It is the same defect as my own M2, one level up, and the diagnosis is exactly right: **a control row grades the property it VARIES, not the property it is filed under.** My table 2 varied the anchor for the *refusal* direction and held it constant for the *over-refusal* direction, then the body and the test header asserted coverage of both.

**Production code is unchanged in this commit** — `git diff` restricted to non-comment lines is empty — so the corpus numbers (147/147/0 before and after, byte-identical listings, 127 live refusals) stand unre-run by construction, and the digest cannot have moved for a new reason. Re-ran `./cmd/grafel/` anyway with fresh `GRAFEL_HOME`/`GRAFEL_DAEMON_ROOT`: **exit 0**, 181.0s.

---

### 1. Finding 1 — the missing record-anchor scoping control

Put in **`Later.java`**, as suggested. It is the cheapest home *and* the right one: table 5's subject is the *position* of the colliding declaration, so `class TP` is already declared last, and a record component naming it picks up the position axis and the anchor axis in the same row rather than needing a sixth fixture. The fixture now reads:

```java
class BoxP<TP> { TP item; OrderP ok; }
record BoxRP<TP>(TP item, OrderP ok) {}      // refusal, record anchor
class PlainP { TP t; OrderP ok; }
record PlainRP(TP t, OrderP ok) {}           // SCOPING control, record anchor
class TP {}
class OrderP {}
```

`javac 25.0.3` rc=0. `PlainRP` declares no parameters, so `TP` there is `class TP` and `PlainRP.t => TP` must be kept. I added the generic sibling `BoxRP<TP>` alongside so the record anchor carries **both** directions under the position axis, not just the one that was missing.

**Both mutants re-scored verbatim** — your gate, my tree, `cp` backup, `go vet` before scoring, exact-match-count assertion on the edit, `-count=1`, restored by `cp` with `shasum` back to `943fb0b1…` and `git status --porcelain` clean after each:

| mutant | anchor gated on | verdict | rows it drops |
|---|---|---|---|
| **ME** | `field_declaration` | **DEAD** | `Forms.java:Plain.t => T`; `NestOk.java:InnerFlatN.ft => TN`, `SiblingN.st => TN`, `pf => TN`; `Later.java:PlainP.t => TP` — five, and `PlainRP.t` **survives**, which is what makes the two anchors independently graded rather than jointly |
| **MA** | `formal_parameter` | **DEAD** | `Later.java:PlainRP.t => TP` — **exactly that one row**, nothing else |

MA is now a minimal kill: the row I added is the only thing standing between it and a green suite.

### 2. Audit for the same conflation — two more instances, both closed

I classified **every** row in the new file by what it varies, and the header now carries that taxonomy. Three kinds, and the third one I had not been distinguishing at all:

* **LIVENESS** — names a type that is *never* a parameter in its file (`Box.ok => Order`). Says the producer still runs at that anchor / in that form. Nothing about scoping. This is the deleted pin's `H2.java` idea and the weakest of the three.
* **SCOPING** — names a type that *is* bound as a parameter somewhere in the file, at a position where it is not in scope, so the edge must be KEPT. `Plain.t`, `InnerFlatN.ft`, `SiblingN.st`, `pf`, `PlainP.t`, `PlainRP.t`.
* **BOUND-HARVESTING** — names a type that appears *inside* a `type_parameters` subtree without being a parameter name. `Bounded.ok`, `Multi.a`, `Multi.b`, `RecD.ok`.

**Instance 2 — the anonymous-class field anchor**, which you flagged as an observation. You were right that it is not an independent grader, and I say so in the comment rather than implying otherwise: an anonymous class's field goes through the *same* emit site as a class field (`buildField`, java.go:1527), so there is no third call site to gate a mutant on. But the row that sat there (`aok => OrderN`) was liveness-only, so I added `pf => TN` — the same anonymous position inside a **non-generic** outer (`PlainOuterN`), javac rc=0. What it grades is that the ascent crosses `object_creation_expression` + `class_body` and still stops at the enclosing declaration: `pf` keeps its edge while `af`, in the identical position inside `OuterN<TN>`, does not. ME drops it, which is how I know it is graded and not decorative.

**Instance 3 — the bound-harvesting direction was graded at one anchor only.** `Bounded.ok`, `Multi.a`, `Multi.b` are all class fields; the record anchor had no bounded parameter at all. Rather than add a row I gave `RecD` a bound — `record RecD<TD extends OrderD>(TD item, OrderD ok) {}`, javac rc=0 — so the existing `RecD.ok => OrderD` becomes the bound-harvesting control there too. **M3 re-scored: now fails table 2 as well as table 1, dropping exactly `Decls.java:RecD.ok => OrderD`.** Before the bound, M3 was DEAD on class-anchor rows only.

**M2 re-scored** to keep the pair on the record: **DEAD**, failure diff names exactly `NestOk.java:BothN.bt => TN` — the row you independently confirmed as load-bearing.

Tables 2 and 4 now state plainly that their rows are liveness-only and point at tables 3 and 5 for the scoping direction, so the "one axis enumerated thoroughly makes its sibling look covered" reading is closed off in prose as well as in rows.

### 3. Finding 2 — the `:175` prose

Corrected, and you are right in both directions. Replaced with: the ascent *does* collect a method's and a constructor's parameters precisely because it carries no kind list; that behaviour is present and correct but can never fire, because a local class and an anonymous class declared in a method body both emit zero `SCOPE.Schema/field` records. The old sentence claimed the opposite of the code and then drew the wrong conclusion from it.

I also took your stronger soundness argument into the doc comment, **re-derived rather than relayed** — different method, different input from your CST dump. Parsed the pinned grammar's own `node-types.json` (`tree-sitter-java v0.23.5`, matching `go.mod`) for every named node with a `type_parameters` child or field:

```
class_declaration  constructor_declaration  interface_declaration
method_declaration  record_declaration
```

Exactly five, matching yours. Each scopes its parameters over precisely its own subtree, so the lexical parent chain **is** the scope chain and matching the list node alone can neither miss a binder nor invent one.

### 4. Finding 3 — table 4's label

Corrected to the four forms actually in `NestStatic.java` (`NestedS`, `NRecS`, `DeepS.DeeperS`, `CInIS`), counted against the source. The fifth I had listed — the inner-then-static shape, javac row 6 of my own evidence table — is named as omitted, with the note that it is refused by the same union ascent and is not separately distinguishable, so the label was claiming more than the content rather than hiding a gap.

### 5. Verification

* `go test -count=1 ./internal/extractors/java/ ./internal/resolve/` — **ok**
* `GRAFEL_HOME=$(mktemp -d) GRAFEL_DAEMON_ROOT=$(mktemp -d) go test -count=1 ./cmd/grafel/` — **exit 0** (181.0s)
* `gofmt -l internal cmd` — empty. `go vet ./internal/extractors/java/` — clean.
* Mutant table now: **ME DEAD · MA DEAD (was ALIVE) · M1 DEAD · M2 DEAD · M3 DEAD at both anchors · M4/M5/M6 DEAD · M7 DEAD · M9 DEAD · M8 equivalent**.

Thank you for the ±4 on the corpus refusals and for testing the zero-delta claim head-on — "refused names that are also same-file declared types = 0" is the assertion I should have made instead of diffing two listings, and it is the one that actually says what the byte-identical result means.

Not merging — over to you for final verification.

PR #7074, opened 2026-09-16
fix(#7041): rust treated a type parameter as a same-file type, so `struct Box2<Customer>` beside `struct Customer` emitted an edge that BOUND

cajasmota:
Refs #7041 (arm **rust**). go shipped in `b4d33570d` (#7066); java is in review (#7069); C# remains unassigned. This PR touches rust only.

`pub struct Box2<Customer> { pub item: Customer }` in a file that also declares `pub struct Customer` emitted `Box2.item -> Customer`. That edge **BINDS**, so no instrument we own can see it is wrong (#7056). `rustTypeParameterNames` now makes the declaration's own `type_parameters` list a shadowing scope, applied at the stash site in `struct_fields.go` where the owning `struct_item` / `enum_item` is already in hand — no scope tracker needed.

---

## 1. Grammar node names, verified by CST dump

The scala arm shipped a guard switching on two node types its grammar never produces (#7064), so nothing here is from memory. Dumped with `tree-sitter-rust` via the repo's own `ts` façade. The **list node is `type_parameters`, not go's `type_parameter_list`** — porting go's spelling verbatim would have been dead code.

```
===== struct Box<T> { item: T }
  struct_item
    type_identifier  "Box"
    type_parameters  "<T>"
      type_identifier  "T"

===== struct Pair<K, V> { k: K, v: V }
    type_parameters  "<K, V>"
      type_identifier  "K"
      type_identifier  "V"

===== struct Box<T: Order> { item: T }
    type_parameters  "<T: Order>"
      constrained_type_parameter  "T: Order"
        type_identifier  "T"
        trait_bounds  ": Order"
          type_identifier  "Order"          ← REAL type

===== struct Box<T: A + B> { item: T }
      constrained_type_parameter  "T: A + B"
        type_identifier  "T"
        trait_bounds  ": A + B"
          type_identifier  "A"
          type_identifier  "B"

===== struct Box<T> where T: Order { item: T }
    type_parameters  "<T>"
      type_identifier  "T"
    where_clause  "where T: Order"          ← SIBLING of type_parameters
      where_predicate
        type_identifier  "T"
        trait_bounds  ": Order"

===== struct Box<'a, T> { item: &'a T }
    type_parameters  "<'a, T>"
      lifetime  "'a"
        identifier  "a"                     ← identifier, NOT type_identifier
      type_identifier  "T"

===== struct Box<const N: usize> { item: [u8; N] }
    type_parameters  "<const N: usize>"
      const_parameter  "const N: usize"
        identifier  "N"                     ← value namespace
        primitive_type  "usize"

===== struct Box<T = Order> { item: T }
    type_parameters  "<T = Order>"
      optional_type_parameter  "T = Order"
        type_identifier  "T"
        type_identifier  "Order"            ← REAL type (cpp's shipped bug)

===== struct Box<T: Bnd = Order> { item: T }
      optional_type_parameter  "T: Bnd = Order"
        constrained_type_parameter  "T: Bnd"   ← NESTED, not a type_identifier
          type_identifier  "T"
          trait_bounds  ": Bnd"
        type_identifier  "Order"

===== struct Box<'a: 'b, 'b> { item: &'a u8 }
      constrained_type_parameter  "'a: 'b"
        lifetime  "'a"                      ← `left` is a LIFETIME here
        trait_bounds  ": 'b"

===== enum E<T> { V { item: T } }      enum_item   > type_parameters
===== union U<T> { item: T }           union_item  > type_parameters
===== struct Box<T>(T);                struct_item > type_parameters > ordered_field_declaration_list
===== impl<T> Box<T> { … }             impl_item   > type_parameters
===== trait Tr<T> { … }                trait_item  > type_parameters
===== type A<T> = Vec<T>;              type_item   > type_parameters
===== struct Box<T: ?Sized> { … }      trait_bounds > removed_trait_bound > type_identifier "Sized"
===== struct Box<T: Iterator<Item = Order>>  trait_bounds > generic_type > type_arguments > type_binding
===== struct Box<const N: usize = 3>   const_parameter, then an ERROR node "= 3"
                                       (this grammar version cannot parse const defaults)
```

Field names confirmed by probe, not assumed:

```
optional_type_parameter    field "name"          -> constrained_type_parameter "T: Bnd"
optional_type_parameter    field "default_type"  -> type_identifier "Order"
constrained_type_parameter field "left"          -> type_identifier "T"
constrained_type_parameter field "bounds"        -> trait_bounds ": Bnd"
constrained_type_parameter field "left"          -> lifetime "'a"         (for <'a: 'b>)
```

**Every matcher string in the shipped code is graded.** Corrupting `"type_parameters"` → 14 rows RED (M4); deleting or mis-fielding `"constrained_type_parameter"` → 3–4 rows RED (M5, M7); same for `"optional_type_parameter"` (M6, M8). There is no matcher that grades nothing. Conversely, ADDING a `"trait_bounds"` case is **inert** — the shipped switch never receives such a node (M9c, ALIVE) — which is the evidence that no branch is missing.

## 2. `rustc` output for each nesting form

`rustc 1.98.1`, `--crate-type=lib --edition=2021`, run in `rust:slim`. Two corrections have already been published on #7041 because a language rule was asserted from memory, so every claim below is a compiler transcript.

| program | rustc |
|---|---|
| `struct Customer{..}` + `struct Box2<Customer>{item:Customer}` + `fn mk()->Box2<i32>{Box2{item:5}}` | **compiles** — the parameter shadows the struct; `item` is an `i32` |
| `fn f<T>() { struct S { x: T } … }` | **error[E0401]** can't use generic parameters from outer item |
| `fn f<T>() { mod m { pub struct S { pub x: T } } }` | **error[E0425]** cannot find type `T` in this scope |
| `struct T{..}` + `fn f<T>() { struct S { x: T } … }` | **error[E0401] AGAIN** — rustc does **not** fall back to the top-level `struct T` |
| `impl<T> Box3<T> { fn get(self)->T { self.item } }` | **compiles** — an associated fn sees `T` |
| `impl<T> Box3<T> { fn g(){ struct S{x:T} … } }` | **error[E0401]** — a nested item does not |
| `impl<T> Tr for Box4<T> { type A = T; }` | **compiles** — an associated type sees `T` |
| `struct B1<T:Order>`, `B2<T=Real>`, `B3<T:Order+Clone>`, `B4<T> where T:Order`, `B5<'a,T>`, `B6<const N:usize>` | **all compile** |
| `enum E<Customer>{V{item:Customer}}`, `union U<Customer:Copy>`, `struct Tup<Customer>(Customer)` | **all compile** |
| `struct C<const Order: usize> { x: Order }` beside `struct Order` | **compiles** — `x` IS the struct; a const parameter binds in the value namespace |
| `struct D<const N: Alias> { item:[u8;N], ali: Alias }`, `type Alias = usize` | **compiles** — a const parameter's type can be a same-file alias |
| `struct C2<const N: Order>` | **error** — `Order` forbidden as a const-generic type (only integers, `bool`, `char`) |

rustc's own note on the E0401s: *"nested items are independent from their parent item for everything except for privacy and name resolution"*.

**Derived rule: do NOT ascend.** The pass reads the field's own owning declaration's list and nothing else. Row 4 of that table is what settles it: there is **no legal Rust** in which a nested item's bare name means the top-level type while an enclosing generic binds that name — the same shape as java's `static` correction, reached from the opposite direction. So "ascend" has **no legal distinguishing input**, which is why no fixture manufactures one (#7066's lesson: never build a mutant-killer out of code that does not compile). It is doubly unreachable here anyway: `walk()`'s `function_item` case does not descend into bodies, and an `impl` block cannot contain a struct.

Every fixture in the new table was concatenated into one file and compiled — it found a real defect in a draft row (`<Order: Bnd = Real>` needs `impl Bnd for Real {}`, E0277), which is now in the fixture.

## 3. The enumerated form space, graded in both directions

`internal/extractors/rust/field_type_refs_7041_test.go`, 22 rows. `want` is the **complete** sorted set of `ref_kind=field_target_type` edges for the fixture, so an extra edge fails as loudly as a missing one. Every row carries a live control (`ctl: Real`, or `b: Box<dyn Bnd>`, or `ali: Alias`) **inside the same fixture** — java's pattern, and the thing the deleted pin could not do.

| row | form | shadowed field | live control in the same fixture |
|---|---|---|---|
| `plain` | `<Order>` | `G1.item` dropped | `G1.ctl -> Real` |
| `two_params` | `<Order, Real>` | `G2.a`, `G2.b` dropped | `G2.c -> Alias` |
| `bounded` | `<Order: Bnd>` | `G3.item` dropped | `G3.b -> Bnd`, `G3.ctl -> Real` |
| `multi_bound` | `<Order: Bnd + Bnd2>` | `G4.item` dropped | `G4.b1 -> Bnd`, `G4.b2 -> Bnd2` |
| `where_clause` | `<Order> where Order: Bnd` | `G5.item` dropped | `G5.b -> Bnd`, `G5.ctl -> Real` |
| `lifetime` | `<'a, Order>` | `G6.item` dropped | `G6.y -> a` (a real `struct a`), `G6.ctl -> Real` |
| `const_param_type` | `<const N: Alias>` | — (no candidate: `N` is an `identifier`) | `G7.ali -> Alias`, `G7.ctl -> Real` |
| `const_param_name` | `<const Order: usize>` | — | **`G8.x -> Order` KEPT** (compiles; `x` is the struct) |
| `default` | `<Order = Real>` | `G9.item` dropped | **`G9.ctl -> Real` KEPT** (cpp's bug) |
| `default_with_bound` | `<Order: Bnd = Real>` | `G10.item` dropped | `G10.b -> Bnd`, `G10.ctl -> Real` |
| `enum_variant_field` | `enum G11<Order>` | `G11.V.item` dropped | `G11.V.ctl -> Real` |
| `tuple_struct_field` | `struct G12<Order>(Order, Real)` | **unreachable anchor** — no field records at all | asserted empty |
| `union_field` | `union G13<Order: Copy>` | **unreachable anchor** — `union_item` is not a `walk()` case | asserted empty |
| `nested_in_mod` | `mod m14 { struct G14<Order> }` | `G14.item` dropped | `G14.ctl -> Real` |
| `nested_in_fn` | `fn f15(){ struct G15<Order> }` | **unreachable anchor** — `walk()` does not enter fn bodies | asserted empty |
| `collider_trait` | `<Bnd>` shadowing a **trait** | `G16.item` dropped | `G16.ctl -> Real` |
| `collider_enum` | `<Ce>` shadowing an **enum** | `G17.item` dropped | `G17.ctl -> Real` |
| `collider_type_alias` | `<Alias>` shadowing a **type_alias** | `G18.item` dropped | `G18.ctl -> Real` |
| `collider_declared_after` | collider declared **after** the generic | `G19.item` dropped | `G19.ctl -> Real` |
| `sibling_declaration_keeps_edge` | `G20<Order>` + `Sib20 { buyer: Order }` | `G20.item` dropped | **`Sib20.buyer -> Order` KEPT** |
| `non_colliding_param` | `<T>`, `T` is not a same-file type | n/a | `G21.ctl -> Real` |

Three anchors turned out to be **unreachable**, and they are recorded as such rather than quietly omitted — each asserts the fixture is empty, so a future pass that starts emitting there turns the row RED and has to re-derive the rule.

## 4. VARIED / HELD-CONSTANT, per row

This is #7041 trap (c), which has bitten five consecutive PRs including two *fixes*. **A control row grades the property it VARIES, not the property it is filed under.**

**Anchor row** (every axis at its anchor value): `plain` — anchor = named-struct field; parameter form = bare `<Order>`; owner form = top-level `struct`; collider form = `struct`, declared **before** the generic.

| axis VARIED | rows |
|---|---|
| **parameter form** | `two_params`, `bounded`, `multi_bound`, `where_clause`, `lifetime`, `const_param_type`, `const_param_name`, `default`, `default_with_bound` |
| **anchor** (which producer makes the field record) | `enum_variant_field`, `tuple_struct_field`, `union_field` |
| **owner's nesting** | `nested_in_mod`, `nested_in_fn` |
| **collider's declaring form** | `collider_trait`, `collider_enum`, `collider_type_alias` |
| **collider's position** relative to the generic | `collider_declared_after` |
| **the declaration the parameter belongs to** (scope leakage) | `sibling_declaration_keeps_edge` |
| **nothing about scoping — producer LIVENESS only** | `non_colliding_param` |

`non_colliding_param` is called out explicitly because it is the row that *looks* like a scoping control and is not: its parameter name is never a same-file type, so it passes unchanged with the refusal deleted. It is not counted as a scoping control anywhere in this PR.

Attacking my own held-constant axes is where `collider_trait`/`collider_enum`/`collider_type_alias`, `collider_declared_after`, and the three anchor rows came from — and where the two **unreachable-anchor** findings came from, which I would have missed had I only varied the parameter form.

## 5. Mutant table

Every mutant: `cp` backup, exact-match edit with an **asserted match count** (a non-matching edit is reported as `EDIT-DID-NOT-LAND`, never scored), `go vet` before scoring, `-count=1`, restored by `cp`, verified by `shasum` **and** `git diff HEAD`. Permissive direction first. The call site is mutated, not just the helper (#6533, #7066).

| # | direction | mutation | verdict | killed by |
|---|---|---|---|---|
| M1 | permissive | **call site**: drop the refusal loop entirely | **DEAD** | 14 rows |
| M2 | permissive | **call site**: `emitRustStructFields` passes `nil` | **DEAD** | 14 rows |
| M3 | permissive | **call site**: `emitRustEnumVariantFields` passes `nil` | **DEAD** | `enum_variant_field` |
| M17 | permissive | **call site**: read `body` (field list) instead of `node` — go's exact analogue | **DEAD** | 14 rows |
| M4 | permissive | matcher `"type_parameters"` → go's `"type_parameter_list"` | **DEAD** | 14 rows |
| M5 | permissive | delete the `constrained_type_parameter` case | **DEAD** | `bounded`, `multi_bound`, `default_with_bound` |
| M6 | permissive | delete the `optional_type_parameter` case | **DEAD** | `default`, `default_with_bound` |
| M7 | permissive | field `"left"` → `"name"` on `constrained_type_parameter` | **DEAD** | `bounded`, `multi_bound`, `default_with_bound` |
| M8 | permissive | field `"name"` → `"left"` on `optional_type_parameter` | **DEAD** | `default`, `default_with_bound` |
| M9b | permissive | `constrained_type_parameter` returns its `bounds` instead of `left` | **DEAD** | `bounded`, `multi_bound`, `default_with_bound` |
| M9 | **over-refusal** | collect the **bound** alongside the name (cpp's shipped bug, faithfully) | **DEAD** | `bounded`, `multi_bound` |
| M10 | **over-refusal** | `optional_type_parameter` collects `default_type` (cpp's *other* shipped bug) | **DEAD** | `default`, `default_with_bound` |
| M11 | **over-refusal** | collect a `const_parameter`'s NAME | **DEAD** | `const_param_name` |
| M14 | **over-refusal** | collect a `const_parameter`'s TYPE | **DEAD** | `const_param_type` |
| M12 | **over-refusal** | collect a `lifetime`'s inner `identifier` | **DEAD** | `lifetime` |
| M13 | **over-refusal** | also read the `where_clause` | **DEAD** | `where_clause` |
| M18 | **over-refusal** | one type-parameter set for the whole FILE, applied to every declaration (the plausible wrong implementation) | **DEAD** | `sibling_declaration_keeps_edge` — the **only** row that kills it |
| M15 | permissive | whole-subtree lookup for `type_parameters` instead of direct child | **equivalent under legal stable Rust** | see below |
| M9c | over-refusal | add a `"trait_bounds"` case to the switch | **equivalent under the current suite (INERT)** | see below |
| — | permissive | **ascend** to an enclosing item's parameters | ***alive but not production-reachable*** | see below |

**M15 — equivalent, with the distinguishing input class named and shown not to exist.** I probed for nested `type_parameters` nodes under a `struct_item`: they DO occur, but only under `higher_ranked_trait_bound` — `struct S where for<'a> u8: Tr<'a>`, `struct S<T: for<'a> Fn(&'a u8)>`, `struct S { f: Box<dyn for<'a> Fn(&'a u8)> }`. In stable Rust `for<…>` binds **lifetimes only**, so such a list yields an empty name set and both lookups return `nil`. Distinguishing would need `for<T>`, which is not stable Rust. The direct-child form is kept as the stricter of the two.

**M9c — inert, not redundant-but-live.** Nothing in the shipped switch ever passes a `trait_bounds` node to `rustTypeParameterName` (only `left` and `name` are followed, and neither is ever a `trait_bounds`), so the case is unreachable whether present or absent. This is the positive evidence that the switch has no missing branch, not merely that the suite does not notice one.

**Ascend — alive but not production-reachable, with the input named.** The distinguishing input is "a nested declaration whose field names a same-file type that an enclosing generic item also binds". Every spelling of it is rejected by rustc (E0401 / E0425, table in §2), *including* the one with a matching top-level declaration present. Not killed with a manufactured invalid-Rust fixture.

## 6. Corpus: a named zero, and a refusal that fires 174 times

788 `.rs` files across four repos (`actix-examples`, `just`, `mini-redis`, `tokio`), copied to scratch; corpora untouched. Each file extracted through the production extractor and driven through the production resolver (`BuildIndex` → `ReferencesEmbedded`), exactly as graph assembly does.

|---|---|---|---|
| **before** (`origin/main`) | 265 | 265 | 0 |
| **after** | 265 | 265 | 0 |

The full row sets (`<file>:<field> -> <target_type>`, one line each, sorted) are **byte-identical** — `diff` exits 0. **No row changed, so none needs justifying as a correct drop.**

**That zero is non-vacuous, and the delta alone would have looked like a dead guard.** Instrumenting the refusal (temporary counter, removed before commit) shows it **fires 174 times** across those 788 files:

| refused name | count | | refused name | count |
|---|---|---|---|---|
| `T` | 125 | | `L` | 3 |
| `S` | 22 | | `Q`, `D` | 2 each |
| `W`, `U` | 6 each | | `R`, `I`, `F`, `E`, `B` | 1 each |
| `V` | 3 | | | |

Named examples: `just/src/alias.rs:Alias.target` refuses `T`; `just/src/binding.rs:Binding.value` refuses `V`; `just/src/recipe.rs:Recipe.dependencies` refuses `D`; `actix-examples/middleware/various/src/redirect.rs:CheckLoginMiddleware.service` refuses `S`; `tokio/tokio/src/io/async_fd.rs:AsyncFd.inner` refuses `T`. Not one of those 174 names is also declared as a struct/enum/trait/type_alias in its own file, which is exactly why the edge counts do not move: the guard is live, it is simply not yet load-bearing in this corpus. Per the #7057 precedent, a corpus-relative zero is not a bound on the language.

## 7. Interaction with #7038 (rust's ambiguity rule counts RECORDS, not KINDS)

**No interaction, and I did not touch it.** #7038 is on the **target** side (`rustInFileTypeTargets`'s collision map removes a name seen on two records, even when `graph.EntityID` would have made them one node). This change is entirely on the **candidate** side, at the stash in `struct_fields.go`; the two never meet. The corpus row set being byte-identical is consistent with that.

My form-space work did **not** incidentally demonstrate #7038 — every collider in the table is a single declaration, so no row exercises the collision path. One row adds `impl Bnd for Real {}` (rustc requires the default to satisfy the bound); subtype `impl` is already excluded from the target set, and `G10.ctl -> Real` is asserted present, confirming it does not collide.

## 8. The deleted pin, and the prose it left behind

- **Deleted**: `TestRustFieldTypeRefs_KnownOverFire_TypeParameterShadowsSameFileType` (`field_type_refs_6912_test.go:593`), a hard `t.Fatalf`. Its own message said *"if type-parameter scope is now consulted, delete this test and say so"*. Not worked around.
- **Corrected prose #1** — the surviving pin's doc-comment said "these two tests… both over-fires… expected to FAIL". Rewritten for the one over-fire that remains (module scope), with a paragraph naming the deleted sibling, what replaced it, and why a bare over-fire assertion cannot distinguish "the over-fire is still here" from "the producer died".
- **Corrected prose #2** — `field_type_refs.go`'s header bullet *"IT IS FILE SCOPE AND NOTHING ELSE… it over-fires twice"* with the `struct Box<Order>` example. Rewritten: module scope remains the one over-fire, the type-parameter one is closed, the derived Rust rule and its rustc evidence are recorded, and the **not-collected** set is spelled out with why each entry would delete a correct edge. A claim nothing measures is exactly what #7056 is about, so none was left standing.
- **Added prose**: the meta-key comment now records the one decision made during the walk and why it is safe there (the scope is lexical and belongs to a node already in hand), mirroring go's.

## 9. Verification

- `go test -count=1 ./internal/extractors/rust/ ./internal/resolve/` — PASS
- `GRAFEL_HOME=$(mktemp -d) GRAFEL_DAEMON_ROOT=$(mktemp -d) go test -count=1 ./cmd/grafel/` — PASS, **digest unmoved** (175s)
- `go build ./...`, `go vet ./internal/... ./cmd/...` — clean
- post-mutant-run `git diff --stat HEAD` matches the real change only; both production files verified by `shasum` against their pre-mutant backups


cajasmota (commented):

I did not write this code. Everything below was re-derived in my own worktree (`review-7074`, `d2761085b`), with my own `rustc 1.98.1` in `rust:slim` and my own corpus run. One finding.

**The shipped behaviour is correct.** The finding is a *grading* gap, in the over-refusal direction — the direction #7056 established has no symptom at all.

---

### FINDING 1 (over-refusal, distinguishable and ungraded) — the 22-row table holds the shadowed field's **candidate multiplicity** constant, so "the refusal is per-CANDIDATE, not per-FIELD" is asserted by no row

Applying the java lens to §4's own axis list: the table varies the parameter form, the anchor, the owner's nesting, the collider's form and the collider's position. It holds constant **the field-type expression of the shadowed field**. Every shadowed field in all 22 rows has exactly **one** candidate, and that one candidate is the refused one:

| row | shadowed field | candidates | refused | kept |
|---|---|---|---|---|
| `plain` | `item: Order` | `[Order]` | all | — |
| `lifetime` | `item: &'a Order` | `[Order]` | all | — |
| `bounded` | `item: Order` | `[Order]` | all | — |
| … | … | … | … | — |

**No row has a field whose declared type names a type parameter *and* a real same-file type.** So the loop in `struct_fields.go` is graded only on inputs where "drop the refused candidate" and "drop the whole field" are indistinguishable.

**Mutant M-R1 — over-refusal, whole-field instead of per-candidate.** `struct_fields.go`, exact-match edit, match count asserted = 1, `go vet` clean before scoring:

```go
 for _, c := range cands {
     if typeParams[c] {
-        continue
+        kept = kept[:0]
+        break
     }
     kept = append(kept, c)
 }
```

```
go test -count=1 ./internal/extractors/rust/ ./internal/resolve/   →  EXIT 0   ALIVE
```

Restored by `cp`; `shasum` back to `083e9ea4…`, `git diff HEAD` empty.

**It is distinguishable, on legal Rust.** Distinguishing input, compiled with `rustc 1.98.1 --crate-type=lib --edition=2021`, **exit 0**:

```rust
use std::collections::HashMap;
pub struct Order { pub n: u8 }
pub struct Real  { pub n: u8 }
pub struct G30<Order> { pub m: HashMap<Order, Real>, pub v: Vec<Order>, pub p: (Order, Real) }
```

Probed through the production extractor:

| | edges |
|---|---|
| shipped `d2761085b` | `G30.m -> …:Real`, `G30.p -> …:Real` |
| mutant M-R1 | *(none)* |

and the same at the **enum** anchor — `pub enum G31<Order> { V { p: (Order, Real) } }` → shipped `G31.V.p -> …:Real`, mutant nothing.

So a plausible wrong implementation of exactly this refusal — "if the field's type mentions a type parameter, the field's declared type is not a same-file type, drop it" — **silently deletes a correct edge at both anchors with the whole rust and resolve suites green.** That is cpp's shipped-bug direction, one level up from where §3 attacks it: §3 grades *which names enter the shadow set*; nothing grades *what the shadow set is allowed to remove from a field*.

**The same held-constant axis is ungraded in the PERMISSIVE direction too — mutant M-R2.** Apply the refusal only when the field's type is a single bare candidate:

```go
-        if typeParams[c] {
+        if typeParams[c] && len(cands) == 1 {
             continue
         }
```

```
go test -count=1 ./internal/extractors/rust/ ./internal/resolve/   →  EXIT 0   ALIVE
```

and on `pub struct G32<Order> { pub v: Vec<Order> }` (legal Rust, compiled) it **re-introduces this issue's own defect**: shipped emits nothing, M-R2 emits `G32.v -> scope:component:class:rust:src/g.rs:Order` — a wrong edge that **BINDS**, which is the exact thing #7041 was filed for. The fix is therefore ungraded in *both* directions at any field-type form other than a bare `type_identifier`.

It is not exotic. `HashMap<K, RealType>`, `Vec<T>`, `Result<T, MyError>`, `Option<T>`, `(T, Real)` are the ordinary shape of a generic container's fields — far more common in real Rust than `<Order: Bnd = Real>`. And the PR's own §6 names one: `just/src/recipe.rs:Recipe.dependencies` refusing `D` — its source line is `pub(crate) dependencies: Vec<D>,` (checked in the corpus), a **non-bare** field-type form, i.e. precisely the shape no row covers. If `just` had happened to declare a type named `D` in that file, M-R1 would have deleted a correct edge there and nothing would have noticed.

Prescribing is not reviewing, but for scale: one row with a mixed-candidate field (`pub m: HashMap<Order, Real>` in `G1`, `want` gaining `G1.m -> …Real`) closes it at the struct anchor, and the same on `G11.V` closes the enum anchor. The axis then wants naming in the VARIED/HELD-CONSTANT block, since that block is currently what makes the gap invisible.

---

### Everything else I attacked held up

**(2) The rustc-derived "do NOT ascend" rule — VERIFIED, and I pushed past the PR's table.** My own `rust:slim` / `rustc 1.98.1`, 26 programs. Every row of §2 reproduces, including the load-bearing one: adding a top-level `struct T` next to `fn f<T>(){ struct S{x:T} }` gives **E0401 again** (plus E0574), so rustc does not fall back. My extra boundary pushes:

| program | rustc |
|---|---|
| `impl<Order> B<Order> { struct Inner {…} }` | **error: struct is not supported in `trait`s or `impl`s** |
| `trait Tr<T> { struct Inner {…} }` | **same error** |
| `pub mod m<T> { … }` | **parse error** — a `mod` cannot be generic |
| `fn f<Order>(…) { fn g() { struct S { x: Order } } }` | **E0401** (two levels of nesting, still no capture) |
| `impl<Order> J<Order> { pub fn g() { struct S { x: Order } } }` | **E0401** |
| `impl<T> Tr for B4<T> { type A = T; }` / `impl<T> B<T> { fn get(self)->T }` | compile — associated items DO see `T` |
| `struct C<const Order: usize> { x: Order }` beside `struct Order` | compiles |
| `struct C2<const N: Order>` | error — forbidden as a const-generic type |
| `struct G10<Order: Bnd = Real>` without `impl Bnd for Real {}` | **E0277** — the author's own draft defect reproduces |

The two error rows in the first block are the ones that matter and were not in the PR: they close the only remaining way a *walked* node could sit inside a generic item. Combined with `walk()`'s `function_item` case returning before its body (`rust.go:223-231`, read at source), **"ascend" has no production-reachable input**, and no legal-Rust input at all. *Alive but not production-reachable*, with the input class named — the correct verdict, correctly reached.

**(1) Matcher spellings — graded, and the `trait_bounds` inertness argument is sound.** I did not take the DEAD column on faith:

| my mutant | result |
|---|---|
| delete the `constrained_type_parameter` case | **DEAD** — `bounded`, `multi_bound`, `default_with_bound` |
| `optional_type_parameter` returns `default_type` (cpp's bug) | **DEAD** — `default`, `default_with_bound` |
| `constrained_type_parameter` returns its `bounds`' first `type_identifier` | **DEAD** — `bounded`, `multi_bound`, `default_with_bound` |
| add a `trait_bounds` case returning its own text | **ALIVE** |
| add a `trait_bounds` case that **`panic`s** | **suite green — never reached** |

The panic probe is the version worth having: a green run on the text-returning variant only shows the harvested string matched nothing, whereas the panic shows the case is genuinely unreachable across the whole rust suite. That, plus the structural argument at source (only `left` and `name` are followed, and a `type_parameters` list has no `trait_bounds` direct child), makes "inert, therefore no branch is missing" hold.

**(3) The over-refusal direction inside the collector — no hole found.** `<'a: 'b>` routes through `constrained_type_parameter` with `left` = a `lifetime`, which falls to the default `return ""`; `[u8; N]`'s `N` is an expression identifier, not a `type_identifier`; a HRTB's nested `type_parameters` is under `where_clause`/a bound, never a direct child of the `struct_item`, so the direct-child lookup cannot reach it and the M15 equivalence argument stands.

**(4) The three "unreachable anchor" premises — all TRUE at source.** `function_item` in `walk()` returns without descending; `union_item` is not a `walk()` case and nothing under it produces records; `findRustFieldList` returns only a `field_declaration_list`, so an `ordered_field_declaration_list` yields no field records. Recording them as empty-asserting rows rather than omitting them is the right call.

**(6) Corpus — independently re-derived, and it matches on the number that matters.** Same four repos, read-only, driven through the production extractor in-process:

| | files | edges |
|---|---|---|
| shipped | 789 | **265** |
| refusal removed (= `origin/main`) | 789 | **265** |

`diff` of the two sorted row sets **exits 0 — byte-identical**, confirming §6. My independent refusal count, derived by a separate route (own CST walk + own `<…>` splitter, not the production helper), is **189**, against the PR's instrumented **174**; my file count is 789 against 788. The gap is in the expected direction and explainable — my counter has no access to the serde-`skip` / duplicate-wire-name exclusions that drop fields before the stash — but **174 is not reproduced to the digit**, so treat it as "order-of-magnitude live", as the java arm's 127-vs-131 was. The load-bearing question I asked head-on: **of every refused name, how many are also a same-file declared struct/enum/trait/union/type alias? Zero.** That is the actual reason the edge counts do not move, and it is a corpus-relative zero, not a bound on the language.

**(7) #7038 — no interaction, confirmed.** `rustInFileTypeTargets` is target-side and untouched; its `collide` map keys on records whose `Subtype` is one of struct/enum/trait/type_alias. Every collider in the table is a single declaration, and the `impl Bnd for Real {}` in `default_with_bound` is `Subtype == "impl"`, excluded before the collision map, which is why `G10.ctl -> Real` survives. My probing did not incidentally demonstrate #7038.

**(8) Prose — one inaccuracy, minor, and it points the permissive way.** `field_type_refs.go`, on `rustTypeParameterNames`:

> *"A lifetime and a const parameter contribute nothing, and **neither can be reached by a candidate anyway**: rustTypeRefCandidates collects only `type_identifier`, while a lifetime's name and a const parameter's name are both `identifier`."*

The shadow set is a set of **strings**, so a const parameter's name *can* collide with a candidate — which is precisely what the next sentence says (`struct C<const Order: usize> { x: Order }`, and the `const_param_name` row asserting `G8.x -> Order` is KEPT). The two sentences contradict each other, and the first is the one that would license a future reader to add a `const_parameter` case as "harmless". It is graded (M11 dies on `const_param_name`), so this is a wording defect rather than an ungraded claim — but it is the "prose asserting what no test observes" shape, inverted. Worth a one-line fix while the file is open.

**(9) The deleted pin and the two corrected comments** — read against the diff. The `t.Fatalf` pin is deleted, not worked around; the surviving `ModuleScopeIsNotConsulted` doc-comment no longer promises two over-fires; the `field_type_refs.go` header's "over-fires twice" bullet is rewritten with the not-collected set spelled out. The `non_colliding_param` row is correctly called out as grading liveness only and is not counted as a scoping control anywhere — the java trap is not repeated at the enum anchor either (`enum_variant_field` uses `Order`, a real same-file type, so it genuinely grades scoping there).

### Verification I ran myself

- `go test -count=1 ./internal/extractors/rust/ ./internal/resolve/` — **PASS** (baseline and after every restore)
- `GRAFEL_HOME=$(mktemp -d) GRAFEL_DAEMON_ROOT=$(mktemp -d) go test -count=1 ./cmd/grafel/` — **PASS, 174.187s — digest unmoved, verified not accepted**
- `gofmt -l internal cmd` — empty; `go vet ./internal/extractors/rust/` — clean
- every mutant: `cp` backup, asserted match count = 1, `go vet` before scoring, restored by `cp`, `shasum` + `git diff HEAD` verified; tree ends clean (`git status --porcelain` empty, both temporary probe test files deleted)

### Verdict

**REQUEST-CHANGES** on Finding 1. The rule is right, the derivation is genuinely done rather than asserted, the unreachability premises are true, the corpus zero is non-vacuous — but a natural wrong implementation of this exact refusal deletes correct edges at both anchors with a green suite, and by this issue's own standard that is the direction that has to be graded.

cajasmota:
## Review delta — `d266d2eb8`

**Finding 1 is correct and I am not going to argue any part of it.** The table varied nine things about *which names enter the shadow set* and nothing about *what the shadow set may remove from a field*. Every shadowed field in all 22 rows was one bare candidate, which is the one input shape on which "drop the refused candidate" and "drop the whole field" are the same function. M-R1 and M-R2 were both real, both distinguishable on legal Rust, and M-R2 in particular put **this issue's own wrong-and-BINDING edge back** with my suite green — the defect I was sent to fix, reachable through a `Vec<T>` field, ungraded by my own table.

The detail that should have caught me is the one you found: **my §6 refusal listing names `just/src/recipe.rs:Recipe.dependencies`, and its source line is `pub(crate) dependencies: Vec<D>,`.** I printed a non-bare field type in my own evidence section and did not read it as an axis.

---

### 1. Mixed-candidate rows at both anchors

| row | shadowed field(s) | refused | **kept** |
|---|---|---|---|
| `mixed_candidates_struct` | `m: HashMap<Order, Real>` | `Order` | **`G22.m -> Real`** |
| | `p: (Order, Real)` | `Order` | **`G22.p -> Real`** |
| | `q: (Real, Order)` | `Order` | **`G22.q -> Real`** — refused candidate in SECOND position |
| | `h: Holder<Order>` | `Order` | **`G22.h -> Holder`** — survivor is the generic CONSTRUCTOR, itself a same-file generic declaration |
| | `v: Vec<Order>` | `Order`, `Vec` unmodelled | **nothing** — the permissive half |
| | `ctl: Real` | — | **`G22.ctl -> Real`** |
| `mixed_candidates_enum_variant` | `m`, `p`, `v`, `ctl` as above on `enum G23<Order> { V { … } }` | | `G23.V.m -> Real`, `G23.V.p -> Real`, `G23.V.ctl -> Real` |

The enum row is not redundant with the struct row: the two anchors reach the stash site through different producers (`emitRustStructFields` vs `emitRustEnumVariantFields`), each computing its own shadow set from a different owner node, so a per-field refusal has to be graded at both.

Both sources compiled with `rustc 1.98.1 --crate-type=lib --edition=2021`, exit 0, before becoming rows — including instantiating `G22<i32>` / `G23<i32>` so the parameters are demonstrably parameters.

### 2. M-R1 and M-R2 re-scored verbatim

Transcribed exactly as you wrote them. `cp` backup, asserted match count = 1, `go vet` clean before scoring, `-count=1`, restored by `cp`.

```
baseline sha256(struct_fields.go) = ac9fe90ca82a0ca08ab6274a9127f6e6ca80298feefb7f9cb63d8409085bf667
M-R1 restored, sha256 == ac9fe90ca82a0ca08ab6274a9127f6e6ca80298feefb7f9cb63d8409085bf667
M-R2 restored, sha256 == ac9fe90ca82a0ca08ab6274a9127f6e6ca80298feefb7f9cb63d8409085bf667

M-R1   over-refusal   DEAD   mixed_candidates_enum_variant; mixed_candidates_struct
M-R2   permissive     DEAD   mixed_candidates_enum_variant; mixed_candidates_struct

git diff --stat HEAD -- struct_fields.go  →  '' (empty)
```

Both die at **both** anchors, and the failing-row lists show they are killed by the two new rows only — i.e. the rows are load-bearing, not incidentally red.

The full previous mutant set was re-run against the enlarged suite: **17 DEAD unchanged**, M15 and M9c still the two recorded equivalents. No verdict moved.

### 3. Multiplicity named, and the remaining held-constant axes named too

The VARIED/HELD-CONSTANT block now carries multiplicity as a first-class axis, lists it in the anchor row's values (`ONE candidate, and it is the refused one`), and — the part I think actually prevents a recurrence — names what is **still** held constant with the reason each cannot exercise the guard:

- **serde attributes on the shadowed field** — no row carries `#[serde(rename)]` / `#[serde(skip)]`. Verified at source that this axis *cannot* distinguish: both are handled in `rustFieldsFromList` **before** the stash and act on the field's WIRE NAME, never on its declared type; a skipped field `continue`s before any candidate is collected. Recorded as unable-to-distinguish, not as covered.
- **declaration count in the file beyond two** — `mixed_candidates_struct` now carries two generic declarations (`Holder`, `G22`) and `sibling_declaration_keeps_edge` one generic plus one not, which is what grades per-declaration scoping; a third varies nothing new.

The block also records *why* the omission mattered, in the terms this family keeps relearning: **an axis a block does not NAME is an axis nobody audits**, which is the same failure as filing a row under a property it does not vary — one level up.

### 4. Prose inaccuracy fixed

You are right that it pointed the permissive way. The shadow set is a set of **strings**, so both a const parameter's name and a lifetime's name *can* collide with a candidate — and both do, in rows that exist:

- `const_param_name` → `struct C<const Order: usize> { x: Order }` compiles, `x` IS the struct, `G8.x -> Order` asserted **KEPT**
- `lifetime` → `struct a` beside `<'a>`, `G6.y -> a` asserted **KEPT**

The comment now says both are reachable, both are graded, neither is collected, and records the deleted claim as the inverted "prose asserting what no test observes" shape rather than quietly editing it away.

### 5. Refusal count softened

Restating §6 as you asked: my instrumented count was **174**, your independently-routed count **189** (789 files vs my 788). The gap is in the expected direction — your counter has no access to the serde-`skip` / duplicate-wire-name exclusions that drop fields before the stash — but **174 is not reproduced to the digit and should be read as order-of-magnitude live**, as the java arm's 127-vs-131 was. The load-bearing figure is the one we agree on and which you asked head-on: **of every refused name, ZERO are also a same-file declared struct/enum/trait/union/type alias.** That, not the count, is why the edge totals do not move, and it is corpus-relative, not a bound on the language.

### 6. Kept from your review, not re-litigated

- The `trait_bounds` **panic probe** replaces my weaker text-returning mutant as the inertness evidence. I re-ran it myself on this branch: probe applied (match count 1, asserted), `go vet` clean, `go test -count=1 ./internal/extractors/rust/` **exit 0, zero panics** — the case is genuinely never reached, not merely harvesting a string that matches nothing. Restored by `cp`, sha back to `d6430ce9…`.
- Your two extra `rustc` boundary rows — `impl<T>`/`trait<T>` bodies reject a nested `struct` outright, `mod` cannot be generic, two-level fn nesting is still E0401 — close the last way a *walked* node could sit inside a generic item. Combined with `walk()` returning at `function_item`, "ascend" has **no production-reachable input and no legal-Rust input at all**; the *alive but not production-reachable* verdict stands on stronger ground than I had.

### Verification on `d266d2eb8`

- `go test -count=1 ./internal/extractors/rust/ ./internal/resolve/` — **PASS**
- `GRAFEL_HOME=$(mktemp -d) GRAFEL_DAEMON_ROOT=$(mktemp -d) go test -count=1 ./cmd/grafel/` — **PASS, 178.243s, digest unmoved** (re-run because the comment edits shift line numbers, which the digest could have seen)
- `gofmt -l internal cmd` — empty; `go vet ./internal/extractors/rust/ ./internal/resolve/` — clean
- production diff since `d2761085b` is **comment-only** (verified by filtering the diff for non-comment changed lines — none), so the corpus figures stand as measured rather than being re-asserted: 265 edges / 265 bound / 0 dangling, row sets byte-identical
- tree clean, `git status --porcelain` empty, all mutant backups verified by `shasum` + `git diff HEAD`

Not merging — over to you for final verification and your killing mutant.

PR #7070, opened 2026-09-16
fix(#7068): csharp matched `for_each_statement`, a node the grammar does not have, so no `foreach` loop variable was ever typed — and the arm that never ran was wrong in two more places

cajasmota:
Closes #7068
Refs #7065, #7071

> **Body revised after review round 3.** Sections 5, 6 and 7 below were written before two
> adversarial reviews changed what this PR does. Where a figure or a claim was superseded, the
> section says so inline rather than being silently rewritten — the delta comments on this PR
> carry the reasoning, and the code comments carry the current behaviour.

`internal/extractors/csharp/csharp.go` bound `foreach` loop variables under the node type `for_each_statement`. tree-sitter-c-sharp spells it **`foreach_statement`**. The literal was unpaired, so the loop had **never executed once** and every method call on a `foreach` variable emitted a bare-name `CALLS` target.

The one-word correction is the small part. The loop's *body* had also never executed, so it was not known to be correct — and it wasn't, in two more places.

## 1. CST dump evidence (the spelling was confirmed, not trusted)

Dumped from the same grammar object the package's tests parse with (throwaway test, not committed). Every row below is an observed parse, not a reading of the grammar source:

| source | node | `type` field | `left` field | `leafTypeName(type)` |
|---|---|---|---|---|
| `foreach (Order o in xs)` | `foreach_statement` | `identifier` | `identifier` | `Order` |
| `foreach (string s in xs)` | `foreach_statement` | `predefined_type` | `identifier` | `string` |
| `foreach (List<Order> g in xs)` | `foreach_statement` | `generic_name` | `identifier` | `List` |
| `foreach (Order[] a in xs)` | `foreach_statement` | `array_type` | `identifier` | `Order` |
| `foreach (Order? o in xs)` | `foreach_statement` | `nullable_type` | `identifier` | `Order` |
| `foreach (Ns.Order o in xs)` | `foreach_statement` | `qualified_name` | `identifier` | `Order` |
| `await foreach (Order o in Src())` | `foreach_statement` | `identifier` | `identifier` | `Order` |
| `foreach (ref Order o in span)` | `foreach_statement` | `ref_type` | `identifier` | `""` |
| `foreach (var o in xs)` | `foreach_statement` | **`implicit_type`** | `identifier` | **`var`** |
| `foreach (var (a, b) in pairs)` | `foreach_statement` | **`implicit_type`** | **`tuple_pattern`** | **`var`** |
[...]

Issue #7065, opened 2026-09-16
GATED: a tree-sitter matcher naming a node type the grammar does not have was a SILENT no-op — 3,322 sites now validated against the registered grammars on every CI run


cajasmota:
Every tree-sitter predicate in this repo is a **string literal compared against a grammar node type**, and a literal that names a node the grammar does not have is a **silent no-op**. Nothing in the tree — no build step, no test helper, no lint — checks that any of those strings corresponds to a real node type. A typo, a grammar upgrade, or a name recalled from a different language's grammar all produce the same outcome: the guard stops guarding and every test still passes.

## The instance that prompted this

The scala arm of #6912 (#7064, caught in review before merge) shipped a type-parameter shadow guard switching on:

```go
case "variant_type_parameter", "type_parameter":
```

**Neither node exists in tree-sitter-scala.** A CST dump gives `covariant_type_parameter` for `+A` and `contravariant_type_parameter` for `-A`. So the guard was dead for the *dominant* Scala generic form:

```
class A ; class Box[+A](val a: A)   →  EDGE Box.a -> A     WRONG, and it BINDS
class A ; class Box[A](val a: A)    →  (no edge)           the invariant control passed
```

The wrong edge **binds** to `class A`'s real id — the #7056 signature: right name, right file, admissible kind, wrong target. `26/26 bound` reported a perfect result.

**The detector that found it is the one worth stealing:** a mutant that made *both* switch arms unmatchable killed **zero** tests. That is the signature of dead code, and it is cheap to run deliberately.

## Scale

Scoped to `internal/extractors/` (not the repo root):

- **238 distinct node-type string literals** in `.Type() == "…"` / `.Type() != "…"` comparisons
- across **147 files**
- javascript 36 files, python 26, golang 12, java 10, php 9, csharp 8, ruby 7, kotlin 6, rust 5, hcl 5, swift 4, scala 3

That undercounts — it excludes `switch node.Type()` cases, helper calls like `firstChildOfType(n, "…")`, and any comparison not written with `.Type()` on the left.

**No validation exists.** `grep -rln "NodeTypeExists\|assertNodeType"` across `internal/` returns nothing.

## Why this is not just "write more tests"

[...]

PR #7075, opened 2026-09-16
fix(#7041): C# treated a type parameter as a same-file type, so `class Box<Customer>` beside `class Customer` emitted an edge that BOUND

cajasmota:
The last of four arms of #7041. `class Box<Customer> { public Customer Item { get; set; } }` in a file that also declares `class Customer` emitted a field→declared-type edge asserting `Box.Item`'s type was that class. It is not — `Customer` there is the type parameter, which shadows the declaration inside `Box` — and the edge **BINDS**, so no instrument we have can see it (#7056).

`csVisibleTypeParameterNames` makes the `type_parameter_list` of every lexically enclosing declaration a shadowing scope; `csTypeRefCandidates` drops **only the shadowed candidate**, never the field.

Production: `internal/extractors/csharp/field_type_refs.go`; the one-line call site is in `field_members.go`. **`csharp.go` is not touched** (PR #7070 is working there).

---

## 1. CST dump evidence — the grammar node names were read, not ported

Two things would have shipped silently without this.

**The list node is `type_parameter_list`.** Java's is `type_parameters`; rust's is `type_parameters`. Porting either spelling would have been dead code — the scala failure mode (#7065). Parsed from the pinned grammar's `node-types.json` (`tree-sitter-c-sharp v0.23.1`), **exactly seven** node kinds can carry a `type_parameter_list`:

```
class_declaration            (child)
struct_declaration           (child)
record_declaration           (child)
interface_declaration        (field: type_parameters)
delegate_declaration         (field: type_parameters)
method_declaration           (field: type_parameters)
local_function_statement     (field: type_parameters)
```

`enum_declaration` is absent — a C# enum cannot be generic. Each of the seven scopes its parameters over precisely its own subtree, so **the lexical parent chain IS the scope chain** and matching the list node alone can neither miss a binder nor invent one. No declaration-kind list is needed, and none is written.

**The last three can never fire, and that is a property of the walk, not a claim about C#.** `csharp.go`'s `walk` RETURNS at `method_declaration` / `constructor_declaration` without descending into the body, so a type declared in a method body emits no `SCOPE.Schema/field`; `local_function_statement` only occurs inside such a body; `delegate_declaration` has no body. Recorded as *matched but not production-reachable*, with the distinguishing input named — not killed by manufacturing a fixture.

**`type_parameter` carries a `name` FIELD.** Dumped (17 forms) and probed directly:

```
type_parameter "Order"                    -> name field: identifier=Order
type_parameter "in T"                     -> name field: identifier=T
type_parameter "out R"                    -> name field: identifier=R
type_parameter "[System.Obsolete] in T"   -> name field: identifier=T
```

so variance tokens (`in`/`out`, anonymous) and `attribute_list` (a **sibling** of the name) are both immune. A descendant-identifier walk would harvest attribute names — see mutant **M5**.

**A constraint is structurally out of reach.** `where T : Order` parses as a `type_parameter_constraints_clause`, a **sibling** of `type_parameter_list` under the declaration, never a child of it:

```
class_declaration "class Box<T> where T : Order { … }"
  identifier "Box"
  type_parameter_list "<T>"
    type_parameter "T" [ identifier "T" ]
  type_parameter_constraints_clause "where T : Order"   ← sibling
    identifier "T"
    type_parameter_constraint "Order"
```

That makes cpp's shipped bug (#7057 — harvesting `template<typename T = Order>`'s default argument and silently deleting a correct edge) unreproducible here. **Graded anyway rather than asserted:** row `G3.B` is the constraint's type in field position and must bind.

**Every shipped matcher was corrupted and confirmed to red** — M6, M7, M8 below.

---

### The rule: **ASCEND UNCONDITIONALLY.**

ECMA-334 **§7.7 (Scopes)**: the scope of a type parameter declared by a `type_parameter_list` on a `class_declaration` is "the *class_base*, *type_parameter_constraints_clauses*, and *class_body* of that *class_declaration*", with the same sentence for `struct_declaration`, `interface_declaration` and `delegate_declaration`; §7.7.1 makes the inner declaration **shadow** the outer name. A nested type declaration is part of the enclosing *class_body*, so a nested type's members sit inside the outer parameters' scope.

Corroborating, without a compiler:
- Microsoft's *Generic Classes* guidance states that a type nested in a generic class can depend on the enclosing type's type parameters.
- The compiler carries **CS0693** ("type parameter 'T' has the same name as the type parameter from outer type") — a diagnostic that exists *because* the outer parameter is visible in the nested declaration.
- At the CLR level `Outer<T>.Inner` is `` Outer`1+Inner ``, generic over the same `T`.

### Why this is not a port

Same implementation as java's arm, **different justification**, and neither is portable:

| arm | rule | reason |
|---|---|---|
| kotlin | ascend only through `inner` | a plain nested class does NOT capture; `T` there really names the same-file type |
| java | ascend unconditionally | the static-context reference is **illegal** — javac does not fall back to a top-level `P` |
| rust | do not ascend | E0401, and a top-level `struct T` does not make rustc fall back |
| **C#** | **ascend unconditionally** | the nested reference is **LEGAL and MEANS the outer parameter** |

The generalisation from the java correction applies and was checked here: *which names are in scope* and *which names this pass may bind to* are different questions. For C# they coincide, and the reason they coincide is stated rather than assumed from java's arm matching.

### What could not be verified — **no C# compiler on this machine**

`csc`, `dotnet`, `mono`, `mcs` all absent (verified). Two rows have load-bearing legality and are marked **UNVERIFIED-LEGALITY** in the test, with the reason and what would settle them:

| row | claim | why it is safe either way |
|---|---|---|
| `Sn4.A` — `class O<Order> { static class SN { public static Order A; } }` | §7.7 scopes the outer parameters over the whole *class_body*, and C# nested types are always static in the Java sense, so it should compile — **not demonstrated** | if legal, `Order` is the parameter and refusing is right; if illegal, no program exercises the row. Settled by: `csc` on that snippet. |
| `G13` — `class Mark : System.Attribute {}` + `class G13<[Mark] Order> { public Mark M {get;set;} }` | legal by the default `AttributeUsage` (all targets) and the `…Attribute`-suffix-optional lookup rule — **not demonstrated** | same reasoning; the KEEP half only gets *stronger* if the form is legal |

**Omitted by name rather than guessed:**
- a type parameter in **generic-constructor position** (`class B<D> { D<int,int> f; }`) — believed illegal, so no fixture asserts anything about it;
- a shadowing parameter on `delegate_declaration` / `method_declaration` / `local_function_statement` — omitted for the *verified* walk reason in §1, not for a legality reason;
- an **event** anchor. `event_field_declaration` is a distinct node type that `emitFieldMembers` does not handle at all (dumped), so no field entity exists there. That anchor does not exist in this pass; it is named here rather than left as an unexamined axis.

---

## 3. Enumerated form space — both directions, at every anchor

Three anchors exist: `property_declaration`, `field_declaration`, record positional `parameter`. All three carry **both** a REFUSE row and a scoping KEEP row.

### VARIED / HELD-CONSTANT

**`..._ParameterFormSpace`**
- **VARIES** — parameter form (plain · two-name · constrained · multi-constrained · variance-annotated · attributed); declaration kind carrying the list (class · struct · interface · record); **anchor** (property · field · record positional); **candidate multiplicity** (1 candidate · refused-first · refused-middle · refused-last); field-type syntax (bare · nullable · array · tuple · generic argument · generic constructor).
- **HELD CONSTANT** — nesting depth (top-level only; that axis is the whole of the next table) and namespace (one file-scoped namespace; namespace scope is a *separate* known over-fire, still pinned).

**`..._NestingFormSpace`**
- **VARIES** — nesting depth (1 and 2); whether the nested declaration carries its **own** non-empty list; nested declaration kind (class · record); the `static` modifier; **direction** (ancestor-generic → refuse; non-ancestor-generic → keep).
- **HELD CONSTANT** — parameter form (graded by the table above) and, at most rows, candidate multiplicity; the rows that vary it are marked.

### Per-row classification

Three kinds, stated in the test header:

- **[REFUSE]** — names a type parameter in scope → no edge. Grades the permissive direction.
- **[KEEP]** — names a **real** same-file type at a position where that name is **not** bound as a parameter → edge must survive. Grades the over-refusal direction, which has no symptom.
- **[LIVE]** — names a real same-file type whose name is never a parameter anywhere in the file. Grades producer **liveness only**, and is never counted as a control for scoping.

> *A control row grades the property it VARIES, not the property it is filed under.* Java's arm had two record rows that looked like anchor controls but whose names were never type parameters — they graded liveness. Here `PlainR.X` (record), `PlainS.X` (field), `Plain.X` (property), `N5.A` (descendant scope) and `Sib7.A` (sibling scope) all name `Order`, which **is** a parameter in nine declarations of their file.

**Parameter-form table** (`P.cs`) — 22 surviving edges asserted as an independent literal set, 13 refusals asserted by name:

| row | kind | varies |
|---|---|---|
| `G1.Shadowed` / `G1.Ok` | REFUSE / LIVE | plain generic class |
| `G2.A` `G2.B` / `G2.C` | REFUSE ×2 / LIVE | two parameters |
| `G3.A` / **`G3.B`** | REFUSE / **KEEP** | constraint naming a real same-file type; `Order` must bind in field position |
| `G4.A` / `G4.B` | REFUSE / KEEP | `where T : class, IThing, new()` |
| `G5.A` `G5.B` | REFUSE ×2 | **variance** (`in` / `out`) on an interface |
| `G6.A` / `G6.B` | REFUSE / KEEP | generic **struct** + **field** anchor |
| `G7.A` / `G7.B` | REFUSE / KEEP | generic **record**, positional-parameter anchor |
| `G8.A` / `G8.B` | REFUSE / KEEP | nullable `Order?` |
| `G9.A` / `G9.B` | REFUSE / KEEP | attributed parameter |
| **`G10.M` `G10.R`** | REFUSE+KEEP | **multiplicity**: `Dict<Order,Real>` / `Dict<Real,Order>` — refused in **middle** and in **last** position; both `Dict` and `Real` must survive |
| **`G11.M` `G11.R`** | REFUSE+KEEP | the same multiplicity at the **record** anchor |
| `G12.Pair` / `G12.Arr` | REFUSE+KEEP / REFUSE | tuple `(Order, Real)` — refused **first**; array with no survivor |
| `G13.A` / **`G13.M`** | REFUSE / **KEEP** | parameter attributed with a **same-file** attribute class; `Mark` in field position must bind |
| **`Plain.X`** | **KEEP** | over-refusal control, **property** anchor |
| **`PlainS.X`** | **KEEP** | over-refusal control, **field** anchor |
| **`PlainR.X`** | **KEEP** | over-refusal control, **record** anchor |

**Nesting table** (`N.cs`) — 10 surviving edges, 7 refusals by name:

| row | kind | varies |
|---|---|---|
| `Inner1.A` / `Inner1.B` | REFUSE / LIVE | depth 1 |
| `Deep2.A` / `Deep2.B` | REFUSE / LIVE | depth 2 — the ascent does not stop after one level |
| **`Inner3.A`** | REFUSE | **ascent past a NON-EMPTY nearest list** — the row java found only by scoring a mutant ALIVE |
| `Inner3.B` / `Inner3.C` | REFUSE / LIVE | nearest list, **unioned** with the outer, not picked over it |
| `Sn4.A` / `Sn4.B` | REFUSE / LIVE | `static` nested — **UNVERIFIED-LEGALITY**, see §2 |
| **`N5.A`** / `Inner5.B` / `Inner5.C` | **KEEP** / REFUSE / LIVE | **the ascent's over-refusal control**: `Order` is bound by N5's own **nested** class, i.e. DESCENDANT scope — a subtree search or a file-wide refusal deletes this edge |
| `R6.A` / `R6.B` | REFUSE / LIVE | ascent at the **record** anchor (nested record in a generic class) |
| **`Sib7.A`** / `Sib7.B` / `G7n.K` | **KEEP** / LIVE / LIVE | **sibling control** — the generic declaration is a sibling, not an ancestor |

**Replacement for the deleted pin** — java's two-fixture shape: a **collision** fixture (`class Box<Customer>` beside `class Customer`) and a **no-collision** fixture (`class Box2<T>`, no same-file `T`), each asserting the absence *and* a named control edge. "The over-fire is still here" fails the first fixture's absence assertion; "the producer stopped working entirely" fails **both** fixtures' control rows. They cannot be confused.

---

## 4. Mutant table

Byte-level `cp` backup, restored by `cp`, `shasum` + `git diff HEAD` verified after each. Every edit asserted landed with an exact match count. `go vet` clean before every score. `-count=1` throughout. **Permissive direction first; the call site, not the helper.**

| # | mutation | site | direction | verdict | killed by |
|---|---|---|---|---|---|
| **M1** | pass `nil` shadow set | **CALL SITE** (`field_members.go`) | permissive | **DEAD** | all 3 `_7041_` test functions |
| **M2** | shadow set **exempt at the record anchor only** (`typeNode.Parent().Type()=="parameter"`) | **CALL SITE** | permissive, anchor-asymmetric | **DEAD** | `G7.A`, `G11.M`, `G11.R`, `R6.A` — the record anchor is independently graded, which is the gap java shipped |
| **M3** | **whole-field refusal** (drop every candidate if any is shadowed) | **CALL SITE** | **over-refusal** | **DEAD** | exactly 9 rows: `G10.M/R`, `G11.M/R` (×2 targets each), `G12.Pair` — the rust-arm multiplicity gap, graded at **both** anchors |
| **M4** | ascent stops at the **nearest binder** | helper | permissive | **DEAD** | exactly `Inner3.A` |
| **M5** | collector uses a **descendant identifier walk** instead of the `name` field | helper | **over-refusal** | **DEAD** | exactly `G13.M`. Without the `G13` row this is **ALIVE** — `G9`'s `[System.Obsolete]` cannot grade it, because neither `System` nor `Obsolete` is declared in the file, so harvesting them is unobservable. The row was written for this mutant. |
| **M6** | match `"type_parameters"` (java's/rust's spelling) | helper | permissive / dead-code probe | **DEAD** | 3 test functions — the spelling is load-bearing |
| **M7** | match `"type_param"` instead of `"type_parameter"` | helper | dead-code probe | **DEAD** | 3 test functions |
| **M8** | `ChildByFieldName("type")` instead of `("name")` | helper | dead-code probe | **DEAD** | 3 test functions |
| **M9** | **file-wide**: ascend to root, then collect the whole subtree (java's "descend from the root" shape) | helper | **over-refusal** | **DEAD** | 23 rows across both tables, including every pure scoping control: `N5.A` (descendant), `Sib7.A` (sibling), `Plain.X` / `PlainS.X` / `PlainR.X` (all three anchors) and `G3.B` |

Not scored, and why: computing the shadow set from `startNode` instead of `typeNode` is **equivalent under any suite** — the parent chains differ by at most the one node between them, and neither carries a `type_parameter_list`.

---

## 5. Corpus — the delta **and** the firing count

**3726 `.cs` files** (`aspnetcore-mvc` 2677, `aspnetcore-docs-samples` 947, `aspnetcore-realworld` 75, `WakeOnLAN` 22, `awesome-compose` 5), copied read-only to scratch, driven through the production extractor and the production resolver (`BuildIndex` → `ReferencesEmbedded`).

|  | edges | bound | dangling |
|---|---|---|---|
| **before** (guard disabled by a byte-level mutant) | 643 | 643 | 0 |
| **after** | 643 | 643 | 0 |

**Row listings are byte-identical** (`shasum` `f73d91d4…` on both, 643 rows each). **No rows changed, so none needs justifying.**

**The zero is not a dead guard.** Instrumented at the refusal site, it fires **71 times** over **20 distinct names**:

```
TOTAL FIRINGS=71  DISTINCT NAMES=20
  17 T          13 TStartup    9 TModel      6 TValue    5 TKey
   4 TFloatingPoint          2 TDeclared    2 TEntryPoint   2 TItem
   1 each: TAttribute TElement TFilter TFilterAsync TMode TRequest
           TResult TTagHelper TType TUser TWrapped
```

**The load-bearing figure: refused names that are also same-file declared types = 0.** That is measured directly rather than by proxy — it is exactly what the empty before/after diff says. Had any refused name been a same-file declared type, the **before** run would have carried an extra row. The same figure was zero on java's and rust's arms.

Reporting only the delta would have looked like an unreachable guard.

---

## 6. What was done with the deleted pin and its prose

`TestCsharpFieldTypeRefs_KnownOverFire_TypeParameterShadowsSameFileType` (`field_type_refs_6912_test.go:352`) asserted the **wrong** edge was present behind a hard `t.Fatalf`. It is **deleted**, as its own message instructed, not worked around.

Its prose is corrected in three places rather than left standing (#7056: a wrong claim nothing measures is invisible):

1. **`field_type_refs.go`, the WHAT THE CHECK IS NOT block** — was "it consults neither the C# namespace nor type-parameter scope, so it over-fires **twice**"; now says file scope **plus** type-parameter scope, over-fires **once** (namespace), and carries a new block recording the type-parameter over-fire as **closed, not pinned**, pointing at `csVisibleTypeParameterNames`.
2. **The shared doc-comment above the two `KnownOverFire_` tests** — was written for two cases; rewritten for the one that remains, with an explicit paragraph saying the second is gone, where its replacement lives, and why the replacement has two fixtures.
3. **This issue** — will be updated on merge.

The **namespace** over-fire is untouched and still pinned by `..._KnownOverFire_NamespaceScopeIsNotConsulted`.

---

## 7. Verification

- `go test -count=1 ./internal/extractors/csharp/ ./internal/resolve/` — **ok / ok**
- `GRAFEL_HOME=$(mktemp -d) GRAFEL_DAEMON_ROOT=$(mktemp -d) go test -count=1 ./cmd/grafel/` — **ok** (191s, fresh dirs). Its fixture has 7 `.cs` files and **the digest did not move**: none of them declares a generic type whose parameter name collides with a same-file declared type, so the behaviour change is not reachable there. No constant was bumped.
- `gofmt -l internal cmd` — **empty**
- `go vet ./internal/extractors/csharp/` — clean

**No collision with #7070:** `internal/extractors/csharp/csharp.go` and `internal/quality/golden/csharp-aspnet-core-mini/` are **not touched**. The change is `field_type_refs.go` plus one call line (and its comment) in `field_members.go`.

Refs #7041


cajasmota (commented):
## Adversarial review — PR #7075 (C# arm of #7041), commit `52689f747`

Reviewed in an independent checkout of `refs/pull/7075/head` (`review-7075`). I did not write this code. `csharp.go` untouched, so no collision with #7070.

**Verdict: REQUEST-CHANGES** — one real defect, on exactly the axis this PR's own nesting block declares held constant, and whose header prose claims otherwise.

---

### FINDING 1 — the nesting table holds candidate multiplicity constant at EVERY row, and says it doesn't

`ft7041NestSrc` (`N.cs`): every one of the 17 field anchors has a **single bare-identifier** type. `Order A`, `Real B`, `Keep C`, `Keep K`, `record R6(Order A, Real B)` — not one field in the fixture has more than one candidate.

But the table's own VARIES/HELD-CONSTANT block (`field_type_refs_7041_test.go:44-47`) says:

> HELD CONSTANT — parameter form … and candidate multiplicity **at most rows**; the **rows that vary it are marked**.

and PR §3 says "**the two rows that vary it are marked**". There are no such rows, and none is marked. This is prose asserting what no test observes — the defect class #7056 was filed on.

**It is not only prose.** Multiplicity is graded *only* in `ParameterFormSpace`, where every binder is the field's **own declaring type**. Along the ascent — the whole subject of `NestingFormSpace` — "drop the refused candidate" and "drop the whole field" are still the same function, which is the rust arm's defect (#7041 rust) reproduced one table over.

**Mutant MX — ALIVE.** At the call site (`field_members.go`): *if a candidate is shadowed by an ancestor other than the field's own declaring type, drop the whole field.*

```
go vet ./internal/extractors/csharp/                       clean
go test -count=1 ./internal/extractors/csharp/ ./internal/resolve/   EXIT=0   (both ok)
```

Distinguishing input, run through the production extractor:

```csharp
public class Dict<A, B> { }
public class NOuter<Order> { public class Inner { public Dict<Order, Real> M { get; set; } } }
public class POuter<Order> {                      public Dict<Order, Real> M { get; set; }   }
```

| | baseline | under MX |
|---|---|---|
| `Inner.M -> …:Dict` | present | **gone** |
| `Inner.M -> …:Real` | present | **gone** |
| `POuter.M -> Dict`/`Real` | present | present |

Two correct edges silently deleted, with no symptom — the direction the PR body itself identifies as having none. A nested helper type inside a generic type is ordinary C#, so this is production-reachable, not a manufactured shape.

Corroborating: I re-scored **M3** (unconditional whole-field refusal). It is DEAD as claimed — but **only `ParameterFormSpace` fails; `NestingFormSpace` passes under M3.** The nesting table grades nothing about multiplicity at all, so MX (which is M3 restricted to the ascent) survives.

*Prescribing is not reviewing*, but the cost looks like one row: give any one of `Inner1.A` / `Deep2.A` / `R6.A` a `Dict<Order, Real>` type and add both survivors to the `want` literal set — then mark it, and correct the header sentence to match whatever ends up true.

---

### FINDING 2 (minor) — "exactly which rows are UNVERIFIED" enumerates one of the two

`field_type_refs.go:83` points the reader at `csVisibleTypeParameterNames` for "**exactly which rows are UNVERIFIED** and what would settle them". That block names **Sn4 only**. `G13`'s UNVERIFIED-LEGALITY lives in the test doc and the PR body and nowhere in production:

```
$ grep -rn UNVERIFIED internal/extractors/csharp/
field_type_refs.go:83                    ← the "exactly which rows" pointer
field_type_refs_7041_test.go:66,168,301,357
```

Same shape as round 3 of #7070: a caveat recorded in one place, missing where the claim is made. Either add G13 to the production block or soften "exactly".

### FINDING 3 (minor, citation) — the "Generic Classes" page does not say what it is cited for

Both the PR body and the production comment claim *"Microsoft's own **Generic Classes** guidance states it directly — a type nested in a generic class can depend on the enclosing type's type parameters."* I fetched that page: it covers inheritance from open/closed constructed types, constraint supersets, and invariance. **It never mentions nested types.**

The claim is true and your other citation carries it. The CS0693 reference page states it in so many words:

> "Use a different name for inner type parameters that shadow outer ones (CS0693, CS8387). When a generic member (such as a method or **nested type**) is inside a generic class or method, the inner type parameter isn't necessarily the same as the outer one."

Swap the citation.

---

### The load-bearing language claim — the author is right; the brief I was handed was wrong

I was briefed that a C# nested class must redeclare the outer's type parameters. **That is false.** ASCEND UNCONDITIONALLY is correct, and the CS0693 argument in §2 is sound as stated: a diagnostic whose text is "type parameter 'T' has the same name as the type parameter from **outer type**" is only coherent if the outer one is in scope in the nested declaration — and Microsoft's own prose for it names "nested type" explicitly (quoted above). No C# compiler exists here (confirmed), so this is settled from documents, which is what the PR claims and no more.

### Everything else I checked, and it held

**Grammar (`node-types.json`, tree-sitter-c-sharp v0.23.1, the pin in `go.mod:27`) — all four matcher claims exact.**
- Carriers of `type_parameter_list`: **exactly seven**, and the child/field split matches the PR's table line for line — `class_declaration`, `struct_declaration`, `record_declaration` (child); `interface_declaration`, `delegate_declaration`, `method_declaration`, `local_function_statement` (field `type_parameters`). `enum_declaration` absent.
- `type_parameter.name` is a **required** field of type `identifier`; its only child type is `attribute_list`; variance is an anonymous token, so it cannot be an identifier.
- `type_parameter_list`'s children are **only** `type_parameter` — so a `type_parameter_constraints_clause` is structurally unreachable, as claimed. Not taken on trust: **new mutant MC** (harvest every identifier under a sibling constraints clause — cpp's #7057 shape) is **DEAD**, killed by `G3.B` and `G4.B`. The constraint control discriminates.

**Unreachability claims — verified at source, not accepted.**
- `csharp.go:259` / `:274` `return` at `method_declaration` / `constructor_declaration` **before recursing into children**, so no field entity exists under a method body; `delegate_declaration` has no body child in the grammar. The three unreachable carriers are genuinely unreachable.
- **Event anchor:** `emitFieldMembers` handles exactly `property_declaration`, `field_declaration` → `variable_declaration`, and a record's `parameter_list` `parameter`. `event_field_declaration` is present in the grammar and handled nowhere. The omission premise is true. (`event_declaration` and `indexer_declaration` are likewise unhandled — pre-existing producer gaps, not this PR's.)

**Corpus — re-derived independently, and it matched exactly.** In-package probe over the same five repos, computing the shadow set and the raw candidates at all three anchors:

```
FILES=3726   TOTAL FIRINGS=71   DISTINCT=20
T 17 · TStartup 13 · TModel 9 · TValue 6 · TKey 5 · TFloatingPoint 4 ·
TDeclared 2 · TEntryPoint 2 · TItem 2 · 11 singletons
COLLIDING (refused name also a same-file declared type): map[]   ← the load-bearing zero
```

Byte-identical to §5. On java the reviewer got 131 vs 127 and on rust 189 vs 174; this is the first arm where the independent count matched the author's with no gap, and the zero reproduces.

**Mutants I re-scored myself** (byte-level `cp` backup, `shasum` + `git diff HEAD` verified after each restore, exact match count asserted on every edit, `go vet` before every score, `-count=1`):

| # | verdict | evidence |
|---|---|---|
| M1 `nil` shadow set | **DEAD** | all 3 `_7041_` functions, **including the replacement pin's `collision` subtest** — so the replacement is not vacuous, which was the thing #7041 exists to prevent |
| M2 record anchor exempt | **DEAD** | fails **both** tables |
| M3 whole-field refusal | **DEAD** | `ParameterFormSpace` only — **`NestingFormSpace` passes**, see Finding 1 |
| M5 descendant identifier walk | **DEAD** | by **exactly one row**, `G13.M -> Mark`, as claimed. Nothing else in either table moves. The claim is exact; the fragility is real and honestly disclosed — that row is load-bearing and is one line of a shared `want` literal, so a future edit can un-grade M5 without any test going red |
| **MX** ancestor-binder whole-field refusal | **ALIVE** | **FINDING 1** |
| **MC** harvest constraint-clause identifiers (new) | **DEAD** | `G3.B`, `G4.B` |

**Prose corrections** (§6): verified in all three places. The known-wrong pin is deleted, not skipped; the `KnownOverFire_` doc-comment is rewritten for the one case that remains and says where the replacement lives; the WHAT THE CHECK IS NOT block now says once, not twice. No surviving `t.Skipf`.

**Suites on the restored tree:** `go test -count=1 ./internal/extractors/csharp/ ./internal/resolve/` → ok/ok; `GRAFEL_HOME=$(mktemp -d) GRAFEL_DAEMON_ROOT=$(mktemp -d) go test -count=1 ./cmd/grafel/` → **ok, 186.9s** (fresh dirs — the digest really is unmoved, not assumed); `gofmt -l internal cmd` → empty; `go vet ./internal/extractors/csharp/` → clean.

---

This is the strongest arm of the four — the grammar claims are exact, the unreachability premises are true at source (java's arm shipped a false one), the corpus number reproduces to the row, and the two prior arms' defects (M2 record anchor, M3 multiplicity) are genuinely dead. Finding 1 is the same family defect once more: the block named its held-constant axis honestly in one table and then mis-described it in the other, and the ungraded direction is the silent one.

cajasmota:
## Review addressed — `85e98d3a1`. The finding was correct, and it was worse than an unnamed axis.

The block said the nesting table varied candidate multiplicity "at most rows; **the two rows that vary it are marked**". **No such rows existed.** All 17 field anchors in `ft7041NestSrc` had a single bare-identifier type, so along the **ascent** — the thing this arm is actually about — "drop the refused candidate" and "drop the whole field" were still the same function. Multiplicity was graded only in `ParameterFormSpace`, where **every binder is the field's own declaring type**, which structurally cannot reach that case.

I have no defence for the block. It was written to be the thing a reader trusts instead of re-deriving, and it pointed at rows I never wrote.

---

### 1. Mixed-candidate rows in the NESTING table — 4 rows, 3 anchors, 3 positions

| row | field type | refused position | anchor | depth |
|---|---|---|---|---|
| `Inner1.Mm` | `Dict<Order, Real>` | **MIDDLE** | property | 1 |
| `Inner1.Pf` | `(Order, Real)` | **FIRST** | **field** | 1 |
| `Deep2.Ml` | `Dict<Real, Order>` | **LAST** | property | 2 |
| `R6.Mr` | `Dict<Order, Real>` | **MIDDLE** | **record positional** | 1 |

`Inner1.Mm` is the reviewer's own distinguishing input. Each is asserted **per (field, TARGET)**, not per field — these fields *keep* edges, so the existing whole-field absence loop could not express them; a second loop was added for the (field → `Order`) pairs that must be absent while the neighbours survive.

### 2. MX re-scored verbatim — **DEAD**

Applied at the **call site** (`field_members.go`): compute the shadow set from the field's own declaring type only; if the full ascent refuses more than that, drop the whole field. Edit asserted landed (exact match count 1), `go vet` clean, `-count=1`.

Killed by **exactly the 7 new target rows** and nothing else:

```
Deep2.Ml -> …N.cs:Dict      Inner1.Mm -> …N.cs:Dict     R6.Mr -> …N.cs:Dict
Deep2.Ml -> …N.cs:Real      Inner1.Mm -> …N.cs:Real     R6.Mr -> …N.cs:Real
                            Inner1.Pf -> …N.cs:Real
```

**`ParameterFormSpace` passes under MX.** That is the direct evidence it was ALIVE before — the new rows are the only thing that reaches it, exactly as you said.

### 3. M3 re-scored — the pair is on the record

| mutant | scope of the whole-field refusal | verdict | which table kills it |
|---|---|---|---|
| **M3** | **any** binder | **DEAD** | **BOTH** — 9 rows in `ParameterFormSpace` (`G10.M/R`, `G11.M/R`, `G12.Pair`) **and** 7 rows in `NestingFormSpace`. Previously only `ParameterFormSpace` failed. |
| **MX** | an ancestor **other than** the declaring type | **DEAD** | `NestingFormSpace` **only** — 7 rows. `ParameterFormSpace` cannot reach it. |

MX is the strictly narrower mutant, and it is the one that proves the ascent's multiplicity is independently graded.

### 4. The block's claims are now true, **and the audit found three more**

You asked me to re-read every other VARIES/HELD-CONSTANT assertion against the rows that actually exist. I did, and the claim was aspirational in three further places:

1. **"Every form used below … whose legality is not in doubt"** — contradicted by the two `UNVERIFIED-LEGALITY` rows in the same file. Now: *exactly two* rows rest on an undemonstrated legality (`Sn4`, `G13`), named, with every other form still covered by the original sentence.
2. **Both fixture headers' lists of same-file types were stale** — `ft7041FormSrc` said "`Real`, `IThing` and `Keep`" after `Dict` and `Mark` had been added; `ft7041NestSrc` had no such list at all.
3. **A count of my own was wrong.** I wrote that `Order` is bound as a type parameter by "six declarations" in `N.cs`. It is **seven** (N1, N2, N3, N4, N6, Inner5, G7n) — now enumerated rather than counted, which is the only form of that claim that cannot drift.

Every axis in both blocks now names the rows that carry it. `ParameterFormSpace`'s block additionally records that generic-**constructor** position is varied only as a KEPT candidate (`Dict`), because a type parameter in constructor position is the omitted illegal form — previously that read as if both directions were covered.

### 5. Minors

- **`field_type_refs.go`** no longer promises "exactly which rows are UNVERIFIED" while naming `Sn4` alone. There are **two**, both named there now, and `G13` has its own bullet with the reason, the safe-either-way argument, and what would settle it (`csc` on that snippet).
- **The citation was wrong, and you were right to fetch it.** The MS *Generic Classes* page never mentions nested types. Replaced with the **CS0693** compiler-message page, which says it verbatim — *"a generic member (such as a method or **nested type**) … inside a generic class"* — plus the standing point that the diagnostic exists at all only because the outer parameter is visible in the nested declaration. The wrong citation is recorded in the comment rather than quietly dropped, per #7056.

---

### Verification

- `go test -count=1 ./internal/extractors/csharp/ ./internal/resolve/` — **ok / ok**
- `GRAFEL_HOME=$(mktemp -d) GRAFEL_DAEMON_ROOT=$(mktemp -d) go test -count=1 ./cmd/grafel/` — **ok** (187s, fresh dirs), digest unmoved
- `gofmt -l internal cmd` — empty; `go vet` clean before every mutant score
- Backups by `cp`, restored by `cp`, `shasum`-verified both ways; `git diff HEAD` confirms **the production diff this round is comment-only** — no behaviour changed, so the parent commit's corpus figures (3726 files, 643/643/0 before and after, 71 firings over 20 names, refused ∩ same-file-declared = ∅) stand unchanged and were not re-derived.
- `csharp.go` still untouched (#7070).

Nothing else surfaced. Not merging — yours to verify.

cajasmota (commented):
## Adversarial re-review — round 2, commit `85e98d3a1`

Independent checkout of `refs/pull/7075/head` (branch `r2-7075`), `85e98d3a1`. I did not write this code and did not re-audit what round 1 confirmed. `csharp.go` untouched — no collision with #7070.

**Verdict: REQUEST-CHANGES** — not for the thing round 1 found, which is genuinely fixed, but for the axis immediately next to it, and for an "EXACTLY N" claim written *by this round's own audit* that a compiler-message page contradicts.

First, the good news, verified verbatim rather than taken on report.

---

### Everything the round-2 comment claims about MX and M3 is exactly true

**MX — DEAD.** Applied at the call site (`field_members.go`): shadow set from the field's own declaring type only; if the full ascent refuses more, drop the whole field. Exact match count 1, `gofmt`, `go vet` clean, `-count=1`, `cp` backup/restore, `shasum` + `git status --porcelain` empty after.

Killed by **exactly the 7 new target rows**, and nothing else in the package:

```
Deep2.Ml -> N.cs:Dict    Inner1.Mm -> N.cs:Dict    R6.Mr -> N.cs:Dict
Deep2.Ml -> N.cs:Real    Inner1.Mm -> N.cs:Real    R6.Mr -> N.cs:Real
                         Inner1.Pf -> N.cs:Real
go test -count=1 ./internal/extractors/csharp/ ./internal/resolve/  -> 1 FAIL: _7041_NestingFormSpace only
```

**`ParameterFormSpace` passes under MX** — confirmed, and that is the load-bearing bit. I also ran the converse you did not: **MX against the parent's test file (`52689f747`) is GREEN (EXIT=0).** So the new rows are not merely *sufficient* to kill MX, they are the *only* thing that kills it. Round 1's finding and this fix are both exactly as described.

**M3 — DEAD at both tables**, by exactly 9 and exactly 7 rows, counted from the failure output:

| table | missing rows | extra |
|---|---|---|
| `ParameterFormSpace` | **9** — `G10.M/R`, `G11.M/R` (×2 targets), `G12.Pair` | 0 |
| `NestingFormSpace` | **7** — `Deep2.Ml`, `Inner1.Mm`, `R6.Mr` (×2), `Inner1.Pf` | 0 |

**The production diff really is comment-only.** `git diff -U0 52689f747..85e98d3a1 -- field_type_refs.go`, filtered to changed lines not beginning with `//`: **0 lines**. The parent's corpus figures stand.

**The counts and citations I re-derived independently:**

- **"SEVEN declarations bind `Order` in `N.cs` (N1, N2, N3, N4, N6, Inner5, G7n)"** — parsed the fixture: **7, and the same 7**. The six→seven correction is right, and enumerating rather than counting was the right call.
- **"All 17 field anchors had a single bare-identifier type"** (of the parent's `N.cs`) — 20 anchors, minus the three `string N` properties on `Order`/`Real`/`Keep` that carry no candidate = **17**. True.
- **"every declaration in this file is top-level, so EVERY binder here is the field's OWN declaring type"** — true structurally, and demonstrated by MX passing `ParameterFormSpace`.
- **"every list in this file is plain and unconstrained"** (`N.cs`) — true.
- **"Generic-CONSTRUCTOR position is varied only as a KEPT candidate (`Dict` in G10/G11)"** — true, and the omission is named in `field_type_refs.go`.
- **"`Dict`'s own `A`/`B` collide with nothing"** — true; no same-file type is named `A` or `B` in either fixture.
- **Every row name appearing in either VARIES/HELD-CONSTANT block resolves to a row that exists.** That was the round-1 defect and it is closed.
- **The CS0693 citation is accurate.** I fetched the page. It says verbatim: *"When a generic member (such as a method or **nested type**) is inside a generic class or method, the inner type parameter isn't necessarily the same as the outer one."* The quote is not paraphrased, and recording the wrong former citation instead of dropping it is the right disposition.
- **Both `UNVERIFIED-LEGALITY` rows (`Sn4`, `G13`) are named in both files** — `Sn4` ×4 and `G13` ×3 in `field_type_refs.go`, marked at the row in the test.

---

## FINDING 1 — the *number* of refused candidates is held at 1 at every row in both files, and no block names it. Mutant **MF** is **ALIVE**.

Round 1 forced multiplicity into the nesting table. Both blocks now enumerate that axis as:

> CANDIDATE MULTIPLICITY — one candidate (G1.Shadowed), refused-**first** (G12.Pair), refused-**middle** (G10.M, G11.M), refused-**last** (G10.R, G11.R)

Every member of that enumeration varies **where the one refused candidate sits**. Across all ~46 rows of `P.cs` and `N.cs` — including all four new ones — **no field ever has more than one refused candidate.** That axis is named nowhere, in either block.

**Mutant MF — ALIVE.** In `csTypeRefCandidates`, refuse only the *first* shadowed candidate and keep every later one:

```go
name := string(src[n.StartByte():n.EndByte()])
if !typeParams[name] || refused {
        out = append(out, name)
} else {
        refused = true
}
```

```
exact match count 1   gofmt -w   go vet -> clean
go test -count=1 ./internal/extractors/csharp/ ./internal/resolve/   EXIT=0   (ok / ok)
```

Distinguishing input, through the production extractor (`extractCSFiles` + `fieldTypeRefEdges`):

```csharp
public class Order { public string N { get; set; } }
public class Real  { public string N { get; set; } }
public class Dict<A, B> { }
public class Cache<Order, Real> { public Dict<Order, Real> M { get; set; } }
```

| | baseline | under MF |
|---|---|---|
| `Cache.M -> Z.cs:Dict` | present | present |
| `Cache.M -> Z.cs:Real` | — | **present** |

A wrong edge to a real same-file type, and it **BINDS** — the symptomless direction this whole issue exists for (#7056). This is not a manufactured shape: `Dictionary<TKey, TValue>`-shaped fields on two-parameter generics are ordinary C#, and the corpus firing table in §5 is full of `TKey`/`TValue`/`TModel` pairs. The collision requirement is the same rarity that applies to the guard as a whole.

**And `P.cs` already has the binder.** `G2<Order, Real>` binds both names; it just has no field naming both. The fix is two lines, and I priced it rather than asserting it was cheap:

```
G2:   … public IThing C { get; set; } public Dict<Order, Real> D { get; set; } }
want: "G2.D -> " + cls + "Dict",
```

```
with fix, unmutated production:  EXIT=0  ok
with fix, MF applied:            EXIT=1  FAIL _7041_ParameterFormSpace
                                 got  G2.D -> P.cs:Dict, G2.D -> P.cs:Real
                                 want G2.D -> P.cs:Dict
```

Prescribing is not reviewing — but the cost is one fixture field and one `want` row, and it also closes the axis in the block rather than leaving it unnamed.

The reason I am blocking on this rather than filing it as a nit: the enumeration above reads as *the* multiplicity space. It is the same shape as round 1's defect — a block naming an axis and pinning only part of it — one notch over, on the exact axis round 1 just forced open.

---

## FINDING 2 — "EXACTLY TWO rows rest on a legality that is not demonstrated" is false. `G5` is a third, and unlike the other two it is demonstrably **illegal**.

New this round, in the test header:

> EXACTLY TWO rows below rest on a legality that is not demonstrated — `Sn4` … and `G13` … EVERY OTHER form below is a plain generic class/struct/record/interface …, whose legality is not in doubt.

`field_type_refs_7041_test.go:125`:

```csharp
public interface G5<in Order, out Real> { Order A { get; set; } Real B { get; set; } }
```

From the **same Microsoft compiler-message page this PR now cites** for CS0693:

> **CS1961**: *Invalid variance: The type parameter must be covariantly valid on 'type'.* … A covariant (`out`) type parameter can only appear in output positions (return types), while a contravariant (`in`) type parameter can only appear in input positions (parameter types).

A `{ get; set; }` property is **both** positions. So `Order A { get; set; }` (contravariant `Order` in a getter) and `Real B { get; set; }` (covariant `Real` in a setter) are each CS1961. There is no compiler here, but this is the same evidentiary standard §2 sets for itself, and the document is unambiguous.

**Why this is more than a marker nit.** By the PR's own safe-either-way doctrine — *"if illegal, no program exercises the row"* — `G5.A` and `G5.B` grade nothing. And they are the sole occupants of two axes the VARIES block names:

- *the FORM of the parameter declaration — … **variance-annotated (G5)***
- *the DECLARATION KIND carrying the list — … **interface (G5)***

`G5` is the only `interface` and the only `in`/`out` row in `P.cs`. If the form cannot compile, both named axes are empty — which is precisely the class of unmet coverage claim this round's audit was supposed to close.

Priced: **two token deletions** make it legal without touching what it grades, and the suite stays green.

```
public interface G5<in Order, out Real> { Order A { set; } Real B { get; } }
go test -count=1 -run TestCsharpFieldTypeRefs_7041 ./internal/extractors/csharp/   EXIT=0  ok
```

`{ set; }` and `{ get; }` are still `property_declaration` nodes with a `type` field, so both rows keep grading exactly what they claim to.

---

## FINDING 3 (nit) — one more stale count, the sibling of the one you fixed

`field_type_refs_7041_test.go:221`, on the `Plain.X` control:

> `Order` is a parameter in **nine** declarations of this file

Parsed from `ft7041FormSrc`: **eleven** — G1, G2, G5, G6, G7, G8, G9, G10, G11, G12, G13. Present at the parent too, and not caught by the re-audit that fixed the identical sentence shape in `N.cs` from six to seven by enumerating. The control's validity does not depend on the number, so this is cosmetic — but §4.3 says "enumerated rather than counted … is the only form of that claim that cannot drift", and this is the other instance of that claim.

## FINDING 4 (nit) — the new per-`(field, TARGET)` loop can never execute

The §1 note frames the new loop as the mechanism for the mixed-candidate rows:

> a second loop was added for the (field → `Order`) pairs that must be absent while the neighbours survive

It cannot fire. The `want` comparison above it is `t.Fatalf` on exact join equality, so it runs first and aborts; if `got` ever contained `Inner1.Mm -> …Order` the joins would differ and the test would die there. If it does *not* contain it, the loop finds nothing. Demonstrated with a no-ascent mutant that leaks exactly that edge:

```
grep -c "edges mismatch"                 -> 1
grep -c "mixed-candidate row leaked"     -> 0
```

The **absence is graded** — by the exact-set `Fatalf`, which is why MX dies — so nothing is ungraded here. But the loop contributes a message it can never print, and the same is true of the pre-existing whole-field loop. Worth a one-line comment saying it is a diagnostic that only becomes load-bearing if the set comparison is ever relaxed, rather than prose implying it does the work.

---

### Verification (this head, pristine tree, `git status --porcelain` empty before and after every mutant)

- `go test -count=1 ./internal/extractors/csharp/ ./internal/resolve/` — **ok / ok**
- `GRAFEL_HOME=$(mktemp -d) GRAFEL_DAEMON_ROOT=$(mktemp -d) go test -count=1 ./cmd/grafel/` — **ok** (177.9s, fresh dirs, EXIT=0)
- `gofmt -l internal cmd` — empty; `go vet ./internal/extractors/csharp/` — clean before every score
- Every mutant: `cp` backup, exact match count asserted, `cp` restore, `shasum` + `git status --porcelain` verified. Never `git checkout --`.
- No C# compiler present; every legality claim above rests on the cited Microsoft compiler-message page, named as such.

| mutant | site | direction | verdict |
|---|---|---|---|
| **MX** (drop whole field when a non-declaring ancestor shadows) | call site | over-refusal | **DEAD** — exactly the 7 new rows; **ALIVE** against the parent test file |
| **M3** (whole-field refusal, any binder) | call site | over-refusal | **DEAD** — 9 rows + 7 rows, both tables |
| **MNA** (no ascent past the declaring type) | helper | permissive | **DEAD** — reported by the set `Fatalf`, not the new loop |
| **MF** (refuse only the FIRST shadowed candidate) | helper | **permissive (binds)** | **ALIVE → FINDING 1** |

Not merging, not pushing. Everything round 2 says about MX, M3 and the comment-only production diff checks out exactly; FINDING 1 is the neighbour of the axis you just opened, and FINDING 2 is in the sentence the audit itself wrote.

cajasmota:
## Round 3 addressed — `9d536cf4a`. Both blockers and both nits. One additional illegal row found by re-deriving.

---

## Blocker 1 — refusal COUNT was 1 at every row, and POSITION is a different axis

You are right that this is my own round-1 finding one turn further in, and the framing is the part I should have caught: round 1 I was pinned at *one candidate total*; I fixed it by varying **where the refusal sits**; that reads like "multiplicity is covered" and it is not. All ~46 rows still refused **exactly one** candidate, so *"refuse the shadowed candidates"* and *"refuse the FIRST shadowed candidate"* stayed the same function.

**Two rows, one per table** — I went one row past the priced fix, and the reason is below:

| row | field type | binders | refusals | survivor |
|---|---|---|---|---|
| `G2.M` | `Dict<Order, Real>` in `G2<Order, Real>` | own declaring type | **2** | `Dict` |
| `Inner3.Md` | `Dict<Order, Real>` in `Inner3<Real>` nested in `N3<Order>` | **two different levels** — `Order` from the ancestor, `Real` from the nearest list | **2** | `Dict` |

**MF re-scored — DEAD**, `go vet` clean, edit asserted landed (count 1), `-count=1`:

```
ParameterFormSpace:  a field naming TWO shadowed names leaked one of them:
                     "G2.M -> …P.cs:Real"
NestingFormSpace:    mixed-candidate row leaked the shadowed candidate:
                     "Inner3.Md -> …N.cs:Real"
```

`G2.M -> Real` is exactly the leak you predicted.

**Why the second row rather than just the priced one.** Under MF the nesting table fails on `Inner3.Md` **and nothing else** — so without it, that table would not grade refusal count at all, and the round-2 lesson (own-declaring-type binders cannot reach the ascent) would repeat verbatim one axis over. `Inner3.Md` is also the only row anywhere whose two refusals come from two *different levels*, so a "read one level" and a "refuse only the first" implementation leak different halves of it.

**Both blocks now name REFUSAL COUNT explicitly**, as a separate clause from refusal position, with the rows that carry it. The block also records why the round-2 fix did not reach it, so the next reader does not have to re-derive that distinction.

---

## Blocker 2 — the legality list was re-derived, and I found a second illegal row you did not flag

You are right, and the direction matters: this was a row asserted **legal** that is not, which is worse than an unmarked unknown because it looks graded.

`interface G5<in Order, out Real> { Order A { get; set; } Real B { get; set; } }` is **CS1961 twice** — a contravariant `in` parameter cannot appear in a getter (an output position), a covariant `out` parameter cannot appear in a setter (an input position). And G5 is the **sole** occupant of two axes my own block names — *variance-annotated* and *declaration kind … interface* — so under my own doctrine both were ungraded.

I re-read **every declaration in both fixtures** against the variance rules, the accessor rules, the attribute-target rules, the constraint-ordering rules and the nullable rules. That turned up **a second illegal row the review did not flag**:

> `class G9<[System.Obsolete] Order>` is **CS0592**. `ObsoleteAttribute`'s `AttributeUsage` lists Class, Struct, Enum, Interface, Constructor, Method, Property, Field, Event and Delegate — **not GenericParameter**. It cannot be applied to a type parameter at all.

### FIXED (illegal ⇒ graded nothing)

| row | was | now | diagnostic |
|---|---|---|---|
| `G5` | `Order A { get; set; } Real B { get; set; }` | `Order A { set; } Real B { get; }` | CS1961 ×2 |
| `G9` | `[System.Obsolete]` | `[App.Mark]` | CS0592 |

`[App.Mark]` keeps the **qualified-name CST shape**, which is G9's only reason to exist beside G13's simple name, and is valid on a type parameter because `Mark` declares no `AttributeUsage` and so defaults to `All`. **Probed after the fix:** `G5.A` and `G5.B` still emit as `SCOPE.Schema/field` entities with signatures `Order A` / `Real B`, so the accessor change does not cost the anchor and the two axes are now real rather than nominal.

### MARKED UNVERIFIED-LEGALITY — four, not two

| row | premise | what would settle it |
|---|---|---|
| `Sn4` | `static` class nested in a generic class | `csc` on the snippet |
| `G13` | same-file attribute class on a type parameter | `csc` |
| `G9` | **shares G13's premise** — recorded as a fourth row but explicitly **not** a second confirmation of the third; two rows on one premise are one premise | `csc` |
| `G8` | **new**: `Order?` on an *unconstrained* parameter needs C# 9+ (else CS8627); the sibling `Real?` warns CS8632 outside `#nullable enable`. The row grades the `nullable_type` CST shape, which parses at any LangVersion | `csc` with an explicit LangVersion |

Both the production block and the test header now carry this list, with the two superseded claims ("exactly which rows" → named Sn4 alone; "exactly two rows") recorded rather than folded in.

---

## Nit 3 — eleven, and enumerated

"nine declarations" was **eleven**: G1, G2, G5, G6, G7, G8, G9, G10, G11, G12, G13. Enumerated, not counted. This is the sibling of the sentence I corrected six→seven in round 2 by enumerating — I fixed one instance of the pattern and left the other standing, which is its own lesson.

## Nit 4 — the loop is now reachable, and demonstrated

Both absence loops (the new per-`(field, TARGET)` one **and** the pre-existing whole-field one, which had the same defect) now run **before** the exact-set `t.Fatalf`. Demonstrated by MF:

| | named `leaked` messages | `edges mismatch` lines |
|---|---|---|
| previous ordering | **0** | 2 |
| now | **2** | 2 |

The set comparison stays as the backstop for anything neither loop names. The `//§1` note is rewritten to say that absence was always graded *by the set* and that it was the **named diagnosis** that was dead — I have not left prose crediting a mechanism that never ran.

---

## Verification

- `go test -count=1 ./internal/extractors/csharp/ ./internal/resolve/` — **ok / ok**
- `GRAFEL_HOME=$(mktemp -d) GRAFEL_DAEMON_ROOT=$(mktemp -d) go test -count=1 ./cmd/grafel/` — **ok** (185s, fresh dirs), digest unmoved
- `gofmt -l internal cmd` — empty; `go vet` clean before every score; `cp` backups, `cp` restores, `shasum`-verified both ways
- **MX re-scored against the round-3 fixtures: still DEAD** (one failing test function, `NestingFormSpace`) — the new rows did not disturb it
- `git diff HEAD` on the production file: **0 non-comment changed lines**. The behaviour is untouched since `52689f747`, so the corpus figures (3726 files, 643/643/0 before and after, 71 firings over 20 names, refused ∩ same-file-declared = ∅) stand and were not re-derived
- `csharp.go` still untouched (#7070)

Nothing else surfaced. Not merging.

cajasmota:
## CY-1 addressed — `3ac2b330d`. You are right, and it is the fourth multiplicity axis in four rounds.

The diagnosis is exact: round 3 named REFUSAL COUNT, but every row it added counted **distinct shadowed names** — `Dict<Order, Real>` is two names with one occurrence each, `Inner3.Md` two names from two levels. **No row anywhere gave one shadowed name two occurrences**, so *"how many names are refused"* was graded and *"how many times one name occurs"* was not.

---

### 1. The rows — two, one per table

| row | field type | binder | occurrences of `Order` | survivor |
|---|---|---|---|---|
| `G1.Mo` | `Dict<Order, Order>` in `G1<Order>` | the field's **own declaring type** | **2** | `Dict` |
| `Inner1.Mo` | `Dict<Order, Order>` in `Inner1`, nested in `N1<Order>` | the **ancestor** — `Inner1` has no list of its own | **2** | `Dict` |

**Why the second one.** Same reasoning as `Inner3.Md`, which you endorsed: without it the nesting table would grade this axis **not at all**, and the round-2 lesson — own-declaring-type binders cannot reach the ascent — would repeat one axis over for the third time. Each fixture already carried its real-type positive control (`G1.Ok -> Real`, `Inner1.B -> Real`), so no new control was needed.

### 2. The axis is named in both blocks

Both VARIES blocks now carry **OCCURRENCES OF ONE NAME** as a clause distinct from refusal count and from refusal position, and the header states all **four** multiplicity axes explicitly rather than one:

> how many candidates a field has · **where** the refused one sits · how many **distinct names** are refused · how many **times** one refused name occurs

The block's running self-critique gains round 4, with the progression spelled out (one-candidate → position → count → occurrences) so the next reader inherits the distinction instead of re-deriving it.

### 3. CY-1 re-scored verbatim — **DEAD**

`go vet` clean, edit asserted landed (match count 1), `-count=1`, `cp` backup + `cp` restore, `shasum`-verified both ways.

```
ParameterFormSpace:  a shadowed name occurring TWICE in one field type leaked
                     an occurrence: "G1.Mo -> …P.cs:Order"
NestingFormSpace:    mixed-candidate row leaked the shadowed candidate:
                     "Inner1.Mo -> …N.cs:Order"
```

Killed by **exactly those two rows and nothing else** — `G2.M` and `Inner3.Md` pass under it, which is precisely why the count axis could not reach this.

### 4. The mirror — CY-2, refuse only the **LAST** occurrence: **DEAD, by the same two rows**

Implemented as a post-pass that collects every identifier and drops only each shadowed name's final occurrence. `go vet` clean, edit landed (count 1).

**This is one graded property, not two, and I am not claiming otherwise.** `attachCsharpFieldTypeRefs` dedups per `(field, target)`, so whichever occurrence leaks, **exactly one** wrong edge is emitted — the two mutants are observationally identical at the edge level. A third occurrence would not separate them either. That reasoning is now recorded at the row and beside the assertion, so nobody later reads "kills both directions" as two confirmations of one premise.

### 5. One thing I broke and fixed while there

My new row made an existing assertion message wrong: the refusal-count loop said *"a field naming TWO shadowed names leaked one of them"*, which is false for a row carrying **one** name twice. The loops are now split, each saying what it actually checks. Small, but it is the same defect class this PR has already fixed twice — prose describing a mechanism other than the one that ran.

---

### Verification

- `go test -count=1 ./internal/extractors/csharp/ ./internal/resolve/` — **ok / ok**
- `GRAFEL_HOME=$(mktemp -d) GRAFEL_DAEMON_ROOT=$(mktemp -d) go test -count=1 ./cmd/grafel/` — **ok** (181s, fresh dirs), digest unmoved
- `git diff HEAD` on the production file: **0 changed lines this round**; still 0 non-comment changed lines since `52689f747`. Behaviour untouched, so the corpus figures (3726 files, 643/643/0 before and after, 71 firings over 20 names, refused ∩ same-file-declared = ∅) stand and were not re-derived.
- `csharp.go` still untouched (#7070)

Nothing else surfaced. Not merging.

Issue #5729, opened 2026-07-10
epic: decouple MCP serving from the engine — two-process (serve + engine) architecture for fault isolation and independent deploy


cajasmota:
## Goal

Split the single `grafel daemon` into two cooperating processes so engine failures and engine deploys no longer drop the MCP connection:

- **serve** — MCP dispatch + dashboard + zero-copy `mmap` reads of `graph.fb`. Holds no heavy work; ~never crashes. The MCP bridge connects here.
- **engine** — scheduler, watcher, extraction (already subprocess-capable), enrichment, fbwriter. Produces `graph.fb` atomically. Crashes and deploys are isolated here.
- **Seam** = the atomic `graph.fb` FlatBuffer, already `mmap`'d zero-copy by the serve side (`internal/daemon/mcp/graph_cache.go`, ADR-0016/0017, mtime-reload).

## Why now

#5717's root cause: MCP dispatch runs **in-process in the daemon**, so every engine crash/restart severs all bridges at once. grafel is already decoupled at the **data layer** (FlatBuffers + `mmap` + atomic writes + subprocess extraction) — only the **process** is shared. Splitting at the `graph.fb` seam is therefore far cheaper than a shared-memory design would be.

## Hard requirements

- **Single install/lifecycle UX preserved.** The OS service layer provisions and manages **both processes as ONE unit** — **launchd** (macOS), **systemd** (Linux), **Windows Service/SCM** (Windows). `grafel install` / `update` / `start` / `stop` / `restart` stay single operations. **Update-path aware from day one** (`internal/install/*`, the plist/unit/service definitions).
- **Fault isolation:** an engine panic (e.g. #5726) must not drop MCP; serve keeps answering the last-good graph and surfaces a transient state.
- **Independent deploy:** deploying engine code restarts the engine process only; serve + the MCP connection are untouched.
- **Control/status plane:** mutating tools (`submit_repair` → `repair.json`, already file-based; `docgen_apply`; enrichment enqueue) and reindex triggers reach the engine via a small serve→engine channel; the engine writes a status/heartbeat file that serve reads (= #5725), including the indexed commit SHA (#5727).

## Phases

**Phase 0 — fault isolation (single-process, ships first, valuable alone):**
- #5726 — fbwriter fail-soft + `recover()` (on marshal failure keep the last-good `graph.fb`, don't abort the process).
- #5717 — `mcpCallTool` panic recovery + widened bridge retry/backoff + a transient "reconnecting" state.
[...]

PR #7085, opened 2026-09-16
fix(#7083): the engine child inherited the daemon's standard handles, so whether the daemon could run its own engine depended on its launcher

### LINKED ISSUE #7050: Windows: the daemon can die silently within ~1 minute of a confirmed-healthy start, regardless of launch method, with zero trace anywhere

### LINKED ISSUE #7083: the engine child inherits the parent's stdout/stderr, so whether the daemon can run its own engine depends on its launcher — on Windows that silently drops the child's output, or kills every spawn, depending on the handle value
`internal/daemon/supervise.go:108-109`, in `defaultEngineChildCommand`:

```go
cmd.Stdout = os.Stdout
cmd.Stderr = os.Stderr
```

The engine child **inherits the supervising process's standard handles**. That makes the daemon's ability to run its own engine depend on a property of whatever launched it, which it does not control and does not check.

On Windows this is not merely untidy. `os/exec` hands those `*os.File`s to `StartProcess`, which duplicates the underlying handles into the child; **duplicating an invalid handle fails**, so the spawn fails. A process launched with `Start-Process -WindowStyle Hidden` and **no** `-Redirect*` flag runs under `UseShellExecute = true` and gets no valid standard handles at all — and adding *any one* `-Redirect*` flag flips it to `UseShellExecute = false`, which creates real pipes for the **whole** stdin/stdout/stderr triad, not just the stream named.

That last detail is what makes this a live suspect rather than a style complaint. On #7050, @mota-mota ran the experiment with a negative control: redirecting **stdout only** — leaving stderr exactly as broken as in every prior silent repro — also made the symptom go away. The variable is not *which* stream is invalid. It is **whether the process has valid standard handles at all**.

## Why a failed spawn can end the daemon

`supervise.go:369` treats a failed spawn as a crash — back off and retry. Backoff starts at 500ms, doubles to a 30s ceiling, and `backoffAndMaybeGiveUp` gives up after **3 consecutive hits at the ceiling**, setting `fatalErr` and returning. An invalid inherited handle fails **deterministically**, so the retry loop never recovers, and summing the waits the first ceiling hit lands at **~61s**.

#7050's title is *"the daemon can die silently within ~1 minute of a confirmed-healthy start."*

It was **silent** because the log lines that would have named it —

```
engine supervisor: spawn failed
engine supervisor: relaunching engine after backoff
engine supervisor: giving up — engine unkeepable
```

— were going through the dead stderr sink fixed in `e51d69fce`. The logging bug was never the cause; it was the reason the cause was invisible.

## What is NOT claimed

- **Not confirmed as the cause of #7050.** The confirming evidence is a `daemon.log` from a run in the original no-`-Redirect*` shape on v0.3.3, containing those three lines with a handle error on the first. That has been requested; it has not come back.
- **It does not explain every death on that issue.** It fits the 55s one cleanly. One of their repros survived 15+ minutes, which needs either a successful spawn resetting the backoff counters or a second mechanism. One story is not being stretched over both.
- **Not a Windows-only concern.** Inheriting handles from an arbitrary launcher is fragile everywhere; Windows is just where it turns into a hard failure instead of a lost write.

## The fix, and how to grade it

Stop inheriting unconditionally. Give the child handles the daemon owns — the log sink it already opens, or `os.DevNull` — and fall back to that whenever the parent's are unusable rather than passing them through.

Grade it in **both** directions, since the permissive direction is the one that matters here:

- a parent with **valid** handles still gets child output where it goes today (no regression in the normal case), **and**
- a parent with **invalid/closed** handles still spawns the child successfully — the direction that is entirely ungraded now.

The second needs a test that actually closes or invalidates the parent's handles before spawning. A test that only asserts the assignment changed grades nothing: **assert the emitted artefact — that the child started — not the field the code set on itself.**

This is worth doing whether or not it turns out to be #7050's cause.

Refs #7050.

cajasmota:
Closes #7083. Related: #7050 (see "What this does NOT claim" below).

## The defect

`defaultEngineChildCommand` (`internal/daemon/supervise.go`) wired the engine child's streams to the supervising process's own:

```go
cmd.Stdout = os.Stdout
cmd.Stderr = os.Stderr
```

So the daemon's ability to run its own engine depended on a property of whatever launched it — one it does not control and never checks. `os/exec` passes those `*os.File`s to `StartProcess`, which **duplicates** the underlying descriptor/handle into the child. The daemon now **owns** the handles it hands down instead.

## What a Windows daemon with no standard handles actually hits

This is the load-bearing premise, so it is stated at the resolution the Go source supports. Read locally against go1.26.4; the Windows *runtime* behaviour behind it is **not** verified here (see "Not verified locally").

`GetStdHandle` has two distinct "no handle" answers, and they behave differently:

| `GetStdHandle` returns | `os.Stdout` is | old code's `Start()` | consequence |
|---|---|---|---|
| **NULL (0)** | non-nil `*os.File` wrapping handle 0 — `newFileFromNewFile` returns nil only for `InvalidHandle` (`os/file_windows.go:125-131`) | **succeeds** — `syscall/exec_windows.go:377` is `if attr.Files[i] > 0`, so handle 0 is **skipped**, never duplicated | engine output silently discarded; no crash loop |
| **INVALID_HANDLE_VALUE** | **nil** `*os.File` | **fails** — `writerDescriptor` takes the `*os.File` branch on a typed nil (`os/exec/exec.go:590`), `Fd()` yields `InvalidHandle`, which is `> 0`, so `DuplicateHandle` is called on it and errors | every spawn fails deterministically |

Only the second row produces the crash loop: `supervise.go`'s run loop treats a failed spawn as a crash, backs off 500ms → 30s, and `backoffAndMaybeGiveUp` gives up after 3 hits at the ceiling, with the first ceiling hit at ~61s. Nothing here establishes **which** of the two a `Start-Process -WindowStyle Hidden` daemon (`UseShellExecute = true`) gets — that is the gap, and it is the difference between "crash-loops in about a minute" and "runs but loses every engine log line". Both are defects this change removes; only one matches #7050's symptom.

## The fix — approach (a): own the sinks, keep the platform's split

The child's stdout goes to `<root>/logs/daemon.log` and its stderr to `<root>/logs/daemon.err` — the daemon's own two log files — both opened `O_APPEND`. When a sink cannot be opened the corresponding `cmd` field is **left nil**, which is os/exec's documented `os.DevNull`. Never the inherited handle, and never a nil `*os.File` (that is row 2 of the table above — the defect itself).

I chose (a) over (b) "probe the inherited handle and fall back":

- There is no portable way to ask whether an inherited handle is usable. The reliable test is to *use* it, which is the failure being avoided. The honest Windows probe is exactly the code I cannot compile or run here (#7005) — an unverifiable predicate for the one platform that needs it.
- (b) leaves two live paths where (a) leaves one, and the second is only taken on the launcher shape CI cannot reproduce.

**Why the split, and what changes where.** An earlier revision of this PR consolidated both streams into `daemon.log`. Review caught that as a regression, and it was: launchd's plist sets `StandardOutPath={{.LogDir}}/daemon.log` but `StandardErrorPath={{.LogDir}}/daemon.err` (`internal/daemon/service/launchd_darwin.go:58-62`); in split mode the indexing work lives in the engine child; and `daemon.err` is the surface the product points users at — `internal/cli/status_stats.go:786,806` and `internal/cli/doctor_summary.go:554` print "see daemon.err", and `rebuild_failure.go:17`, `internal/statusfile/statusfile.go:146,278` and `sched/scheduler.go:392` all record it as the only trace of prior incidents. Folding the child's stderr into `daemon.log` would have removed an engine panic from the file those messages send the user to.

Destination changes, stated rather than assumed:

| Launcher | child stdout | child stderr |
|---|---|---|
| launchd (`grafel install`) | `daemon.log` — **unchanged** | `daemon.err` — **unchanged** |
| `grafel start` (`defaultManualForkStart`) | `daemon.log` — **unchanged** | **moves** from `daemon.log` to `daemon.err` (which the daemon now creates) |
| systemd (`systemd_linux.go`, no `StandardOutput`/`StandardError`) | **moves** from the journal to `daemon.log` | **moves** from the journal to `daemon.err` |
| no valid handles (the #7083 shape) | was discarded or unspawnable → `daemon.log` | was discarded or unspawnable → `daemon.err` |

The two moves are the unavoidable price of owning the handles rather than inheriting them. They move output **towards** the files the product's own failure messages name, and on every launcher an engine-child panic now lands where `grafel doctor` says to look. On systemd, `journalctl` no longer shows engine-child output; that is the one real loss, and it is the same loss the fix requires on any launcher whose redirection we decline to inherit.

`root` was previously `_ = root`; it now locates the sinks. It is still **not** written into the child env — the store-root invariant (epic #5729) is intact, and `SetEngineChildCommandForTest` is unchanged. `internal/cli/watcher_ctl.go` is untouched.

Sinks are cached per path and held for the life of the process: the constructor has no completion hook (the seam returns only an `*exec.Cmd`), so opening one per spawn would leak a descriptor per relaunch — unbounded in a crash loop.

`logDirForRoot` / `logPathForRoot` / `errPathForRoot` in `paths.go` are now the **only** construction of those paths in production code: `paths_unix.go` (both branches of `DefaultLayout`) and `paths_windows.go` were re-spelling the shape and now call them, so the supervisor cannot drift from the layout the daemon itself uses. `paths_test.go` keeps its independent spelling as the check. The launchd plist template spells `daemon.log`/`daemon.err` a second time in its own package's template text; that coupling is named in the doc, not claimed to be shared.

## Grading

`internal/daemon/supervise_stdio_test.go`, portable (no build tag), standard os/exec helper-process pattern. Every test asserts **where the output actually landed**, not what field the constructor set — except test 3 and test 5, which must assert the field because "left nil" is itself the behaviour (and test 3 pairs it with an artefact check).

1. `..._HealthyParentKeepsTheLaunchdSplit` — no-regression, in launchd's actual shape. Child stdout reaches `daemon.log`, child stderr reaches `daemon.err`, neither marker crosses into the other file, and **sentinels written before the spawn survive** (the `O_APPEND` grading). Green before and after.
2. `..._InvalidParentHandlesStillSpawns` — the ungraded direction. Parent's `os.Stdout`/`os.Stderr` are a **closed** `*os.File`; the child must start and both streams must reach their own file, without truncation.
3. `..._UnopenableSinksFallBackAwayFromInherited` — sinks unopenable, inherited handle **valid**: `cmd.Stdout`/`cmd.Stderr` must be nil (not a nil `*os.File`) and the inherited sink must stay empty.
4. `..._NonAbsoluteRootYieldsNoSink` — table over `""`, `"./"`, `"relative-root"` from a `t.Chdir` temp cwd that *has* a `logs/` directory: no sink, and no `logs/daemon.log` created there.
5. `..._SinksAreSharedAcrossSpawns` — one sink per file across relaunches, and stdout's sink is not stderr's.

### Reds

The regression review caught (stderr folded into `daemon.log`), against the previous revision of this PR:

```
--- FAIL: TestEngineChildStdio_HealthyParentKeepsTheLaunchdSplit
    engine child stderr did not reach daemon.err; daemon.err="sentinel-already-in-daemon-err\n"
    engine child stderr was folded into daemon.log, losing the daemon.err split;
      daemon.log="sentinel-already-in-daemon-log\nengine-stdio-helper-stdout-marker\nengine-stdio-helper-stderr-marker\nPASS\n"
```

`O_TRUNC` in place of `O_APPEND`:

```
--- FAIL: TestEngineChildStdio_HealthyParentKeepsTheLaunchdSplit
    spawning the engine child truncated daemon.log; want it to still start with
      "sentinel-already-in-daemon-log\n", got "engine-stdio-helper-stdout-marker\nPASS\n"
    spawning the engine child truncated daemon.err; want it to still start with
      "sentinel-already-in-daemon-err\n", got "engine-stdio-helper-stderr-marker\n"
```

The original defect, on darwin:

```
--- FAIL: TestEngineChildStdio_InvalidParentHandlesStillSpawns
    engine child stdout did not reach daemon.log; log=""
```

Worth stating precisely: on Unix a closed handle does **not** make `Start` fail. `syscall.forkAndExecInChild` treats `fd == -1` as "close that descriptor in the child", so the child starts with no stdout and the output vanishes — the same shape as the NULL row of the Windows table. Asserting the artefact rather than the error is what makes this test grade the defect on every platform.

### Mutants

Each applied to a byte-level copy of the current file, restored by `cp`, `shasum -c` + `git diff` verified between runs.

| Mutant | Verdict |
|---|---|
| revert to `cmd.Stdout = os.Stdout; cmd.Stderr = os.Stderr` | **DEAD** (tests 2, 3, 4) |
| assign the sinks unconditionally, so an unopenable sink stores a nil `*os.File` | **DEAD** (tests 3, 4) |
| delete the absolute-path guard (and its now-unused import) | **DEAD** (test 4, rows `""` and `"./"`) |
| `O_TRUNC` instead of `O_APPEND` | **DEAD** (tests 1, 2) |
| consolidate stderr into `daemon.log` — the regression review found | **DEAD** (tests 1, 2, 5) |

The first four were ALIVE or absent before this round; all five are now DEAD.

## Verification

- `go test ./internal/daemon/ -count=1` → **exit 0** (`ok … 35.387s`).
- `go test ./internal/daemon/... ./internal/cli/ -count=1` → **exit 0**.
- `gofmt -l internal/daemon/` clean; `go vet ./internal/daemon/` clean.
- `./cmd/grafel/` not run: no extractor/resolver/assembly change.

### Not verified locally: Windows

Per #7005, Windows-conditional code cannot be compiled or vetted on this machine. The CI `test (windows-latest)` leg ran `internal/daemon` green on the previous revision, which shows the **fixed** code spawns correctly there. Still **unverified anywhere**:

- which `GetStdHandle` result a `Start-Process -WindowStyle Hidden` daemon actually gets — i.e. which row of the table above #7050's reporter is in;
- that the old code fails `Start` on Windows. The chain is read from Go's source (`os/file_windows.go`, `os/exec/exec.go:590`, `syscall/exec_windows.go:377`), not demonstrated on Windows.

`paths_windows.go` is touched (two helper substitutions) and is **unverified locally**; the `compile (windows-latest)` and `test (windows-latest)` legs are the check.

## What this does NOT claim

**This does not claim to fix #7050.** It is an unconfirmed candidate cause: it removes a fragility whose signature matches the symptom, *on the INVALID_HANDLE_VALUE branch only*. The confirming evidence — a `daemon.log` from a run in the original no-`-Redirect*` shape — has been requested and has not come back. It also does not explain every death on #7050: one repro survived 15+ minutes, which needs either a successful spawn resetting the backoff counters or a second mechanism.

The backoff/give-up policy is **untouched**, per scope. One observation, not acted on: a spawn that fails before the child exists is not the same event as a child that started and crashed, and the loop treats them identically — backoff cannot fix a deterministic construction failure, so it spends the whole crash-loop budget for nothing. If that is worth separating it deserves its own issue.


cajasmota:
**Round 2 verified, plus one coordinator mutant nobody ran. Merging.**

**CY-2 — swap the two sinks** (`cmd.Stdout` → `daemon.err`, `cmd.Stderr` → `daemon.log`), at `a0d4f574c`. This is distinct from the consolidate mutant you scored: both sinks still open, both files still exist and still append — only the **direction** flips. A fixture asserting merely that both files are non-empty would survive it. **DEAD**, killed by `TestEngineChildStdio_HealthyParentKeepsTheLaunchdSplit` and `TestEngineChildStdio_InvalidParentHandlesStillSpawns`. So the split's *direction* is graded, not just its existence.

Byte-level backup, `cp` restore, `shasum -c` OK, `git diff HEAD` empty.

**On F1 — you made the right call and did not hide its cost.** Consolidating would have silently broken the surface five product call sites point users at, precisely for the failures most worth diagnosing: in split mode the indexing work that panics lives in the child. Keeping two owned sinks preserves launchd's destinations byte-for-byte *and* makes "see `daemon.err`" true on every launcher rather than only on macOS — a net improvement, not a restoration. The per-launcher table is what makes it reviewable: the systemd journal loss is a real cost and it is stated as the price of not inheriting redirection, rather than omitted.

**On F4 — you re-derived it instead of relaying my summary, which is the behaviour I want.** The two-row table now says which `GetStdHandle` value produces which outcome, and states plainly that **which row a `Start-Process -WindowStyle Hidden` daemon lands in is unknown**. That distinction is load-bearing: it is the difference between "crash-loops in ~61s" and "runs but loses every engine log line", and only the first matches #7050. I had asserted the wrong one as the mechanism, on this PR, on #7083 and to the reporter on #7050; all three are corrected.

**F6 is stronger than asked.** Converting the production sites makes the one-definition claim true, and naming launchd's plist template as a **separate** spelling — rather than pretending it is shared — is the honest form. A claim that is checkable beats a claim that is absolute.

**Your process note is the most valuable thing in this round.** Restoring from a backup taken before the round's edits silently reverted your own work and scored a mutant against the wrong baseline. That it was caught by `shasum -c` against a checksum of the *intended* file is exactly why the checksum is taken of the version you mean to return to, not merely of "some backup". **Backups are per-round, and a restore is only proof if the checksum names the right target.** Recorded.

**What ships unproven, and stays labelled that way:** nothing anywhere demonstrates that the *old* code fails `Start` on Windows — CI only ever ran the fixed binary — and `paths_windows.go` cannot be compiled locally (#7005), so CI's Windows legs are the only check. Both are stated in the body. This still does not claim to fix #7050.

Follow-up already filed from your unacted observation: **#7087** — a spawn that never produced a child shares the crash-loop budget with a child that ran and crashed, and the recovery reset at `supervise.go:396-401` is gated on a child that actually ran, so it is unreachable for a construction failure.

Issue #7079, opened 2026-09-16
full vs incremental bind the same edge by DIFFERENT tiers — the import-aware passes have no candidates incrementally, so edges fall to the global bare-name guess tier (invisible until #7071)


cajasmota:
The full-index and incremental paths bind **the same edge to the same target by different resolution routes**, and until #7071's tier marker existed there was nothing that could see it. Three parity tests now diverge on the **tier**, not on the edge.

## Not a marker bug, and not a wrong edge

Stated up front because it decides the priority: **same edge, same target, all counts identical.** Bind rate, orphan rate, dangle count and the edge sets themselves agree exactly. The divergence is only in *how the edge was reached*.

The mechanism: the **import-aware passes have no candidates on the incremental path** (`import-aware rewrote=0/N`), so an edge that the full index resolves through an import-aware tier falls through to the **global bare-name tier** instead. Both produce the same id here — but one is a resolution and the other is a guess, and they were indistinguishable by construction until there was a name for the difference.

## Why it is worth an issue rather than an allow-list entry and silence

It is currently absorbed by **four allow-list entries keyed on tier value *and* direction sign** in the parity tests. That is the right immediate move — the tests should not be red on a known, understood difference — but it makes a real behavioural gap invisible again, which is the shape this milestone keeps finding.

The substantive question the allow-list defers: **on the incremental path, a population of edges is being bound by a lexical guess where the full path binds them by import resolution.** Today the guesses happen to land on the same targets in the fixtures we have. Nothing establishes that they must. The global bare-name tier is the coarsest one there is — on the django corpus it accounts for **27,903 of 32,003** guess-bound edges — and it is exactly the tier that #7071 exists to flag as untrustworthy.

So the incremental path is, for this population, **strictly less well resolved than the full path**, and a user indexing incrementally gets a different quality of graph than one indexing fully, with no signal that this is so.

## What is NOT claimed

- **No mis-bind observed.** I have not found an input where the two paths bind *different* targets. This is a difference in confidence, not (yet) a difference in result.
- **No incidence figure for the divergent population.** The four allow-list entries bound what the parity tests cover; the corpus-wide count of edges that change tier between full and incremental is **unmeasured**. That number is the first thing worth having, and #7071's new per-tier counters make it cheap to get — run both paths and diff the tier histograms.
[...]

Issue #7077, opened 2026-09-16
sresolver: nameToID is last-writer-wins with NO ambiguity sentinel, so the incremental path can emit a genuinely MIS-BOUND edge where the main resolver refuses


cajasmota:
`sresolver/scoped.go:402`'s `nameToID` map is **last-writer-wins with no ambiguity sentinel**. Where the main resolver *refuses* a bare name that two declarations could satisfy, the scoped resolver picks whichever it saw last — so the incremental path can emit a **genuinely mis-bound** edge, not merely an unmarked guess.

## Why this is worse than #7071, and why it needs its own issue

**#7071** is about a *single-candidate* lexical bind being written with no marker: one plausible match, bound with no way for any instrument to tell a guess from a real resolution. The edge may well be right; nothing records that nobody checked.

**This is the ambiguous case, silently resolved.** Two candidates, one arbitrarily chosen by map-write order. The main resolver has deliberate machinery to prevent exactly this — `byName` deletes on collision, `uniqueMatchInFamily` and `scanLeafMembers` return false on 2+, and the member indexes store blank collision sentinels — and existing tests pin that behaviour. The scoped resolver has none of it.

The resulting edge has a real name, a real file, an admissible kind, and it **binds**. That is the #7056 signature, and every instrument we own — bind rate, orphan rate, dangle count — scores it as a success.

## Provenance

Found during the read-only grounding of #7071, which enumerated every name-only `ToID` write in `internal/resolve` and then checked the out-of-package sites. Two turned up: `symbol_index.go:1027` is **dead** (no production callers), and this one, which is live on the incremental path. **The behaviour is already described in a comment at the site** — so it is known, and was never filed.

## What is NOT claimed

- **No corpus incidence measurement.** No counter exists on this path, so the frequency is unmeasured. It is not zero by construction — same-name declarations across scopes are ordinary in every language this repo indexes — but I am not asserting a number.
- **Not that the scoped resolver should copy the main resolver's machinery wholesale.** The two have different jobs and different inputs; the question is only whether an *ambiguous* bare name should bind arbitrarily or refuse. Refusing trades a wrong edge for a bare leaf, which is the safe direction and matches what the main resolver already does.
- **Not verified end-to-end.** I have read the site and its comment; I have not driven a mis-bind through `grafel` on the incremental path. **Whoever takes this should reproduce it first** — a defect argued from source and never observed is exactly what this milestone keeps finding.

## Suggested next step

[...]

Issue #7072, opened 2026-09-16
csharp: the per-method local-type map is flat and last-writer-wins, so two sibling-scope locals sharing a name emit a CALLS edge with the WRONG receiver type — and it binds


cajasmota:
The C# extractor's per-method local-variable type map is **flat and last-writer-wins**, so two locals of the same name in *sibling* scopes collide. The loser's method calls are then emitted with the **wrong** receiver type — and because the wrong type is a real same-file type, the resulting edge **binds**.

## Pre-existing, and reproducible with no `foreach` anywhere

Found while fixing #7068, and deliberately **not** folded into that PR. The implementing agent reproduced the identical mis-bind using only ordinary locals, so it is not a `foreach` defect and #7068's fix neither causes nor cures it:

```csharp
void M() {
    { Order o = GetOrder(); o.Process(); }      // o is an Order here
    { Customer o = GetCustomer(); o.Process(); } // …and a Customer here
}
```

Both calls are attributed to whichever declaration the walk saw last. One of the two `CALLS` edges names a type that never had that receiver.

## Why it is worth recording despite zero corpus incidence

**0 occurrences in the corpus** (3721 `.cs` files). That is a genuine measurement and also a weak defence: a corpus-relative zero is not a bound on the language, and same-name locals in disjoint blocks are ordinary C#, especially `item`, `e`, `x`, `result`, and — after #7068 — `foreach` loop variables, which are now typed and therefore now participate in this map.

The failure mode is the one this repo keeps finding and cannot see (#7056): the edge has a plausible name, a real file, an admissible kind, and it **binds** to a real entity. Bind rate, orphan rate and dangle count all score it as a success. Nothing distinguishes it from a correct edge except reading the source.

It is also **adjacent to #7071** but distinct. #7071 is about a *single-candidate lexical guess* being written with no confidence marker. This is about a *confidently wrong type* being handed to the resolver in the first place — the resolver behaves correctly given bad input. Fixing #7071 would make this edge's low confidence visible; it would not make the type right.

## What is NOT claimed

- **No claim that the flat map is wrong as a design.** A block-scoped symbol table is real work, and for the overwhelming majority of methods the flat map is correct and cheap. The question is whether the collision case should *refuse* rather than guess — dropping the entry on a same-name-different-type collision would trade a wrong binding for a bare name, which is the safe direction.
[...]

PR #7088, opened 2026-09-16
fix(#7084): aggregate forbidden hits with max — the 5-run median erased an intermittent wrong edge

### LINKED ISSUE #7084: the 5-run MEDIAN erases an intermittent forbidden hit — the one always-fatal precision signal is silently conditional on the defect being deterministic
`scripts/quality/run.sh:422-436` takes the **median** of `forbidden_hits` across `QUALITY_RUNS` indexes (default **5**), and it is the median that reaches `ratchet.py`. `ratchet.py:539-541` then fails only on `forbidden > 0`.

So a forbidden row that fires in **2 of 5** runs medians to **0**, and the gate never sees it.

This matters because `forbidden_relationships` is the repo's only mechanism for saying *"this edge is wrong"*, and it is documented as **always fatal** — the strongest guarantee in the quality system. That guarantee is silently conditional on the defect being deterministic. An intermittent false-positive edge — the output of a map-iteration order, a last-writer-wins collision, a scheduling-dependent pass ordering — is exactly the kind a median erases, and several open issues describe producers with precisely that shape (#7077's `nameToID` last-writer-wins, #7072's flat local map, #7079's full-vs-incremental divergence).

## Why the median is defensible for recall and wrong for precision

For `entity_found` / `relationship_found` the median is doing real work: it suppresses run-to-run jitter in a **recall** number so the ratchet does not flap. Losing one found entity in one run of five should not red the build.

Precision is not symmetric with recall here. `forbidden_hits` has exactly one defined good value — **0** — and any non-zero value is a *positive observation of a wrong edge*. There is no jitter to suppress: the row fired, which means the graph contained an edge a human wrote down as wrong. Taking the median of a set of positive observations and a set of non-observations discards evidence rather than smoothing noise.

**The correct aggregator for this signal is `max`, not `median`** — or, better, the union of offenders across runs, since `report.go` already emits the `forbidden` offender array and a flaky row is *more* interesting than a stable one, not less.

## What is NOT claimed

- **Not observed.** This is read from source (`run.sh:422-436`, `ratchet.py:539-541`), not reproduced. I have not constructed an intermittent forbidden hit and watched it vanish. **Whoever takes this should reproduce it first** — a defect argued from source and never observed is the exact class this milestone keeps producing, and I am not exempt from it.
- **No incidence figure.** Nothing records per-run forbidden counts, so how often a row has flaked below the median is **unmeasured and unmeasurable from existing artefacts**. It could be never.
- **Not that the median is wrong everywhere.** For the recall metrics it is likely right. This is specifically about the always-fatal precision signal.
- **Not that `ratchet.py`'s plumbing is broken.** It was checked: `run.sh --ratchet` propagates `ratchet.py`'s exit 2 through `|| EXIT=$?` and `exit "$EXIT"`, and nothing swallows it. (Note this is *not* the known `--ratchet` ceiling-drop trap, which is a deliberate `got > want` asymmetry at `ratchet.py:597-619`, not a plumbing bug.) The loss happens strictly upstream, in the aggregation.

## How to grade a fix

Both directions, and the second is the one that will be skipped:

1. A row firing in a **minority** of runs must fail the gate — assert on the **emitted failure text naming the row**, not on the aggregated integer, which is the thing being changed.
2. A row firing in **zero** runs must still pass, and a row firing in **all** runs must still fail with the same message as today — otherwise "make it stricter" quietly becomes "make it fire on everything", and a gate that always fires is as uninformative as one that never does.

Worth pairing with a positive control that the harness can actually produce a minority-firing row at all; otherwise direction 1 is graded by a fixture that cannot express it.

Found during read-only grounding of #7056 group 1.

Refs #7056, #7031.

cajasmota:
Fixes #7084.

## The defect

`scripts/quality/run.sh` merges the per-fixture runs with the **median** of every scalar — `forbidden_hits` and `forbidden_entity_hits` included. `QUALITY_RUNS` defaults to **5**, and `ratchet.py` fails only on `forbidden > 0`. So a forbidden row firing in **2 of 5** runs medianed to 0 and the gate never saw it — in the merged report, in the ratchet's verdict, or in strict mode's exit code.

`forbidden_relationships` is the repo's only mechanism for "this edge is wrong", and it is documented as **always fatal**. That guarantee was silently conditional on the defect being deterministic, which the producers most likely to flake are not (#7077 `nameToID` last-writer-wins, #7072 flat local map, #7079 full-vs-incremental divergence).

## Why max, and only here

The median is right for **recall**: `entity_found` jitters run to run, and smoothing it is what stops the ratchet flapping on noise. Losing one found entity in one run of five should not red the build.

Precision is not symmetric. `forbidden_hits` has exactly one defined good value — 0 — and any non-zero value is a **positive observation** that the graph contained an edge a human wrote down as wrong. There is no jitter to suppress, so taking the median of positive observations discards evidence rather than smoothing noise.

Both forbidden counts are now aggregated with `max`. **The recall metrics are untouched.**

## Visible, not merely fatal

Following the `entity_recall_min` / `entity_recall_max` idiom already in the same function, the merged report additionally carries:

- `forbidden_hits_runs` / `forbidden_entity_hits_runs` — how many runs saw the row fire. The aggregate alone cannot tell a row that fired once from one that fired five times, and the intermittent one is the more interesting.
- the **union** of the `forbidden` / `forbidden_entities` offender arrays across runs. The detail arrays are otherwise inherited from `base` = `reports[-1]`, one arbitrary run — so a row that did not fire in *that* run would be counted by the scalar and named by nothing.

## Grading

New tests live in `scripts/quality/test_ratchet.py`, which is what the **`python ratchet tests`** leg (`.github/workflows/test.yml`, job `python-quality-tests`, `python3 -m unittest -v scripts.quality.test_ratchet`) runs.

The merge program is a heredoc inside a shell script, so it cannot be imported. Copying it into the test file would grade a *copy* — run.sh could go back to the median and every case would stay green. The harness therefore **slices the real bytes out of run.sh** and executes them, asserting the slice found exactly one heredoc and that the extracted body still mentions `forbidden_hits`, `forbidden_entity_hits` and `sys.exit(2)`, so a wrong slice fails loudly instead of yielding a stub that passes everything.

Every fatal case asserts the **emitted failure text** out of `ratchet.check`, not the merged integer — the integer is the thing under edit, so asserting on it would grade the edit against itself.

- **Positive control** (`MinorityFiringRowIsConstructible`): proves the fixture shape is a genuine minority (2 of 5) *and* that its median is 0 — i.e. the harness can actually express the case the old code erased. Without it, direction 1 would be graded by a fixture that cannot express the defect.
- **Direction 1**: a row firing in 2 of 5 runs fails the gate, naming `demo-mini: 1 forbidden relationship hit(s) — always fatal`. Same case for the separate `forbidden_entity_hits` key (`... forbidden entity row(s) fired ...`) — one key fixed and the other left behind is exactly the shape that survives here.
- **Direction 2**: a row firing in **zero** runs still passes (`check` rc 0, no offender array in the merged doc, both `*_runs` counters 0); a row firing in **all** runs still fails **with the message it has today**. "Stricter" must not become "fires on everything".
- **The other grader**: `run.sh` reads the merge program's own exit status, and in strict mode that status alone decides the fixture. A separate case pins the aggregator's exit code (0 clean / 2 minority) with recall held at `found == expected`, so the recall branch cannot be the thing producing the code.
- **Scope guard**: `entity_found` is still the median — the odd run out is deliberately the *last* one, so `merged` inheriting `base` cannot satisfy it by accident.

### Mutants scored

| Mutant | Verdict |
|---|---|
| `max_int` → `med_int` for both forbidden counts (the pre-fix code) | **DEAD** — 4 failures: both direction-1 cases, the aggregator exit code, and the offender-union case. The zero-run and all-run controls stayed green, confirming they grade the controls and not the aggregator choice. |
| `max_int("forbidden_hits") + 1` (always fires) | **DEAD** — the clean case goes red, in both `check` and the aggregator's exit code, so direction 2 is not vacuous. |

Full suite at round 1: **52 tests, OK** (61 after round 2).

## Round 2 — review findings F1–F6

**F2 changed the issue, and it is fixed, not merely disclosed.** `run.sh` short-circuits the repeat loop after **3** runs once `entity_recall`/`relationship_recall` agree to within 0.5pp — a predicate that never looked at the forbidden counts. For a recall-stable fixture (most of them) `QUALITY_RUNS=5` was in practice **3**, so a row firing with p≈0.4 was missed **21.6%** of the time rather than 7.8%. `max` fixes what happens to the runs that were *taken*; it cannot recover a run the loop declined to take.

A forbidden hit in **any** completed run now vetoes the short-circuit. Chosen over "require the forbidden counts to agree too" because it is the weaker, safer predicate: the recall test is untouched, the veto can only make the loop run **longer**, and a green fixture is unaffected — it has no hit to veto with. The extra runs are spent only on a fixture that is already failing, where they buy offenders for the union.

| Finding | Fix |
|---|---|
| **F1** the forbidden-ENTITY strict-mode exit branch was ungraded (and the body claimed otherwise) | the aggregator exit-code case is parametrized over **both** keys |
| **F3** the union's de-duplication was unasserted | same offender in two runs is named once — including a **key-order-permuted** copy, which is what `sort_keys=True` buys |
| **F4** `max` + union made count and array independent | pinned with two *distinct* offenders (`forbidden_hits=1`, `len(forbidden)==2`). Which to believe: the **count** is "the most rows any single run saw" and is the gate's trigger only; the **array** is "every distinct wrong edge any run produced" and is the list to act on. The gate fires on either being non-empty, so the divergence can only under-state the headline number, never weaken the gate. |
| **F5** `forbidden_entity_hits_runs` was satisfied by a constant `0` | its non-zero direction and the offending entity's name are asserted |
| **F6** the positive control asserted about a constant no graded case consumed | `MINORITY_RUNS` / `UNANIMOUS_RUNS` / `CLEAN_RUNS` are module-level constants the graded cases read |

**Key/site pairs checked, per the one-key-behind pattern:** aggregator exit branch (edge ✔ / entity ✔), merged scalar (✔/✔), run counter (✔/✔), offender union (✔/✔), short-circuit veto (✔/✔), `ratchet.check` failure text (✔/✔ — round 1). Each pair is scored with the twin's own mutant, not a representative one.

**Also caught locally:** bash 3.2 — what macOS ships — scans command-substitution bodies for quotes even inside a `<<'PY'` heredoc, so a single apostrophe in a comment inside the short-circuit program made the whole script unparseable, with the error reported 200 lines away. `bash -n scripts/quality/run.sh` is now a test; nothing else in the suite executed run.sh as a shell script.

### Round-2 mutants — all DEAD

| Mutant | Verdict |
|---|---|
| M3b: delete the forbidden-**entity** exit branch | **DEAD** (was ALIVE) |
| M3a: delete the forbidden-**edge** exit branch (twin control) | DEAD |
| M5b: `forbidden_entity_hits_runs` → constant `0` | **DEAD** (was ALIVE) |
| M5a: `forbidden_hits_runs` → constant `0` (twin control) | DEAD |
| M7: union stops de-duplicating | **DEAD** (was ALIVE) |
| CTRL: flip `MINORITY_RUNS` to unanimous | **DEAD** (was ALIVE) — the control and two graded cases fail together |
| F2mut: remove the short-circuit veto (pre-round-2 behaviour) | DEAD — both veto cases |
| F2half: veto watches only the edge key | DEAD — the entity case only, so the twins are separably graded |
| UNION: drop the union, keep `base`'s inherited arrays | DEAD |
| MAX: `max_int` → `med_int` (round 1, re-scored) | DEAD — now 8 failing cases |

Full suite: **61 tests, OK**.

## Not verified

The 7.8% / 21.6% miss rates are arithmetic on an assumed p≈0.4 per-run firing probability, not a measured flake rate for any real fixture; they size the effect, they do not establish it. The mechanism — the short-circuit predicate ignoring the forbidden counts — is read off the code and pinned by test.

The `bash -n` test runs whatever bash is on the runner (bash 5 on ubuntu-latest), so it would not by itself have caught the bash-3.2 command-substitution quirk; it catches unterminated heredocs and ordinary syntax breaks on every platform, and the 3.2 case was caught locally on macOS.

The full quality gate was **not** run (compute-heavy, live daemon on the machine). Exit-code propagation in `run.sh` is untouched — `fixture_exit` handling, the `UNMEASURED` path and the `ratchet.py` `got > want` asymmetry are all unchanged — but that is argued from the diff, not measured against a real fixture corpus.


cajasmota:
**Round 2 verified, plus one coordinator mutant nobody ran. Merging.**

**CZ-1 — swap the two merged-report destinations** (`merged["forbidden_hits"] = max_forbidden_ent_hits` and vice versa), at `00ddfde49`. Both values still computed, both still `max` — only where they land changes. **DEAD, 7 failures.** That matters more than it looks: this whole round was about *one key graded and its twin left behind*, and a permutation is the one probe that shows the two keys carry genuinely **distinct fixture values**. If they had been equal, every assertion would have passed while the destinations were free — the fixture-level form of the same defect. Byte backup, `cp` restore, `shasum -c` OK, `git diff HEAD` empty.

I also verified the veto from the branch rather than from your report: `hits.append(... forbidden_hits > 0 or ... forbidden_entity_hits > 0)` watches **both** keys, and `if any(hits): print("no"); sys.exit(0)` runs **before** the recall comparison — so a forbidden hit cannot be overridden by stable recall. The round's own lesson applied to the round's own new code.

**On F2 — you fixed it rather than disclosing it, and chose the weaker predicate for the right reason.** "A forbidden hit in any completed run vetoes the short-circuit" can only make the loop run *longer*, and a green fixture has no hit to veto with, so the cost falls only on a fixture that is already failing — where the extra runs buy offenders for the union. That is strictly better than requiring the forbidden counts to agree, which could have stopped early on two matching non-zero readings.

**The bash 3.2 catch is the most valuable thing in this PR and it was not in anyone's findings.** macOS's bash scans command-substitution bodies for quotes even inside a `<<'PY'` heredoc, so one apostrophe in a comment made the whole script unparseable and reported it 200 lines away. Nothing in the suite ran `run.sh` as a shell script at all — `bash -n` is now a test. And you disclosed that CI runs bash 5, so that test would not itself have caught the 3.2 quirk: the guard is real but it is not the guard that found this.

**What ships labelled:** the 7.8% / 21.6% figures are arithmetic on an assumed p≈0.4, not a measured flake rate — they size the effect, the mechanism is what is pinned. No intermittent forbidden row has been driven against the real fixture corpus, since the full gate was deliberately not run.

The six key/site pairs table in the body — aggregator exit branch, merged scalar, run counter, offender union, short-circuit veto, `check` failure text, each ticked separately for edge and entity — is the format I want on anything with twinned keys from now on.

Issue #4682, opened 2026-06-10
Generalize local-variable receiver typing + test→CALLS coverage crediting — Java/Kotlin (JUnit/Mockito/MockMvc)


cajasmota:
Per-language child of #4672. Generalize the local-variable receiver typing + test→CALLS→handler coverage crediting landed for TS/JS in #4680 (PR) / #4671 to **Java/Kotlin**.

## Concrete construction + route-hit idioms (from #4672)
- **JUnit unit:** `XController c = new XController(dep); c.method();` — type the local `c` from the constructor (and DI fields) and resolve `c.method()` to the class method.
- **Mockito:** `@InjectMocks XController controller;` — the field is constructed/injected by the runner; type it from its declared type so `controller.method()` resolves.
- **Spring MockMvc:** `mockMvc.perform(get("/path"))` under `@WebMvcTest` — link the route-hit `get(...)` to the endpoint definition.
- **Receiver typing** for both local vars and DI fields (the existing Java receiverTypeName/extractCallRelationships pair in internal/extractors/java is the mirror of the TS path — extend it for the `new X()` local form if not already covered).
- **ORMs:** Spring Data repository methods, JPA entity manager reads.

## Acceptance
(a) `new X()` / `@InjectMocks` locals/fields resolve `var.method()` to the class method (CALLS edge from the test method scope); (b) MockMvc route-hit links to the endpoint; (c) ComputeCoverage credits via test→CALLS→handler (+ handler↔definition hop). Exact-mirror fixture + coordinator live-validation on a Java corpus (spring-petclinic). Honest exclusion preserved.

Part of #4672.

Issue #1935, opened 2026-05-23
Generalize "expand request body + response shape into a unified collapsible subtree" across all languages


cajasmota:
Image #123 + owner follow-up: when an endpoint parameter or response is a user-defined OBJECT, **walk the graph** to fetch its field list + types and display the shape inline. Render Parameters AND Response Shapes the SAME way — a collapsible indented subtree — instead of the current two-component split (Parameters table + Response shapes card).

## Vision
Both Parameters and Response sections share a single `<ShapeTree>` component. Each parameter / response row is its own top-level entry, and **clicking expands a subtree of fields** (children resolved from the graph).

```
Parameters (2)
  ▸ transferId       path    string                    Path segment.                    (no expand — primitive)
  ▾ request          body    TransferRequest           Request body inferred from method annotation.
     ├ transferId        String          @NotBlank
     ├ confirmedQty      BigDecimal      @Min(0)
     ├ note              String?         (nullable)
     └ items             List<ItemDTO>   ▸ (click to expand nested)

Response (1)
  ▾ PUT 200            TransferResponse
     ├ ok               Boolean
     ├ confirmedAt      Instant
     └ transfer         Transfer       ▸ (click to expand nested)
```

Collapsed by default for non-primitive shapes (avoid huge initial-render). User clicks the `▸` to expand. Nested types lazy-fetched via a `/api/.../shape?type=<id>&depth=1` endpoint.

## Unified `<ShapeTree>` component (frontend)
`webui-v2/src/components/ShapeTree.tsx` — takes `{rows: ShapeRow[]}` where each row is `{name, in?, type, type_entity_id?, description?, annotations[], nullable, has_children, children?}`. Renders top-level row + nested subtree when expanded. Used by BOTH Parameters and Response sections in `paths.tsx`.

The current "Response shapes" card visual (image #123: bigger PUT chip + "No body" inline) goes away — replaced by the unified subtree.

## Dashboard handler (backend)
`internal/dashboard/v2_paths.go` already injects `request_body_type` (post-#1908). Extend:
1. Each parameter row carries: `{type_entity_id, has_children: bool}` so the frontend knows if it can expand.
2. New endpoint `/api/v2/g/{group}/shape?type_entity_id=<id>&depth=1`: walks CONTAINS edges of the class to return field rows. Capped to depth 3 by default; client requests further on demand.

## Why a new endpoint
[...]

PR #7089, opened 2026-09-16
fix(#7073): a java @interface minted no entity at all, so a project's own annotations were structurally invisible

### LINKED ISSUE #7073: java: an `@interface` annotation type declaration mints NO entity at all — a project's own annotations are invisible in the graph, and a field typed by one can never bind
A Java **annotation type declaration** (`@interface Marker {}`) mints **no entity at all**. The declaration is walked, no error is raised, and the type simply does not exist in the graph — so nothing can reference it, and no field typed by it can bind.

## Probed, not inferred

```java
@interface Marker {}
class Plain {}
class Box { Marker m; Plain p; }
```

Every record the extractor emits:

```
A.java     SCOPE.Component  subtype=file
Plain      SCOPE.Component  subtype=class
Box        SCOPE.Component  subtype=class
Box.m      SCOPE.Schema     subtype=field
Box.p      SCOPE.Schema     subtype=field

field→type edges = [A.java:Box.p => Plain]
```

`Marker` has no record of any kind. `Box.m` is emitted as a field and gets **no edge**, while its sibling `Box.p` binds normally — so this is not a field-side defect, it is a missing target.

`annotation_type_declaration` appears **nowhere** in the java extractor's non-test source (`grep -n annotation_type_declaration internal/extractors/java/*.go`, excluding `_test`). There is no arm for it; it is not a refusal, it is an omission.

## How it was found, and what it is NOT

Found while running my own killing mutant against #7069 (the java arm of #7041). I was probing whether an **annotated** type parameter — `class Box<@Marker T>` — leaks the annotation's name into the type-parameter shadow set, which is exactly the shape that cost the scala arm its guard (`annotated_type` carries the annotation's name as a `type_identifier`, and a Scala annotation *is* a class).

**That probe came back clean**, and the isolation is worth recording because it is what makes this finding precise:

| fixture | edge |
|---|---|
| `class Box<@Deprecated T> { Marker m; … }`, `class Marker` | **`Box.m => Marker`** ✔ |
| `class Box<T> { Marker m; … }`, `class Marker` | **`Box.m => Marker`** ✔ |
| `class Box<T> { Marker m; … }`, `@interface Marker` | **none** |
| `class Box<@Marker T> { Marker m; … }`, `@interface Marker` | **none** |

So the annotation on the parameter is **not** harvested — #7069's guard is correct on that axis. The missing edge is caused solely by the *target* being an `@interface`.

**Not claimed:** that this is a regression, that #7069 caused it, or that it blocks anything. It is pre-existing and orthogonal. **No corpus incidence measured** — a field whose declared type is a custom annotation is legal Java but uncommon; the more likely cost is elsewhere (see below), and I have not quantified either.

## Why it is probably worth more than the field-edge case suggests

A custom `@interface` is a **declared type of the codebase**, and annotations are the backbone of every Java framework this repo cares about — Spring, JPA, the Dubbo work in #7048/#7049. Today, a project's own annotations are invisible as entities, which means:

- nothing can express "this class is annotated with **our** `@Marker`" as an edge to a real node, because the node does not exist;
- a same-file `@interface` cannot make a component-space reference ambiguous, since `javaComponentAddressFamily` never sees it — so an ambiguity rule derived by counting same-file declarations is counting a population with a hole in it;
- the recall-shaped instruments report nothing wrong, because a **missing entity has no dangling edge, no wrong name, and no orphan** — the #7060 shape (`.h` parsed with the C grammar) and #7056's thesis again.

## Suggested next step

Measure before building. Count `@interface` declarations across the Java corpus and, separately, count references to them — if a project's own annotations are as common as I expect in the Spring-flavoured repos, this is a target-kind gap worth its own arm rather than a curiosity. Then decide the Kind deliberately: an `@interface` is an interface in the JLS, and minting it as `SCOPE.Component/interface` would put it **inside** `javaComponentAddressFamily`, which changes ambiguity counts for every existing java edge. That is a real consequence to grade, in both directions, not a free addition.

Provenance: probed directly on `lane/7041-java-type-params` @ `4e733bcdc`; the entity listing above is verbatim extractor output, and the four-row isolation table was run as one table.

Refs #7041, #7069, #7056, #7060.

cajasmota:
Closes #7073.

> **Round 2.** Independent review returned REQUEST-CHANGES on four substantive
> points and two wording points. All six are addressed below, and the claims the
> first version of this body made that turned out to be **false** are recorded as
> false rather than quietly deleted — see *Corrections* at the end.
>
> **Round 3.** One further finding (DA-1): the signature whitespace collapse was
> ungraded because every signature this file asserted was single-line. Fixed with
> one assertion row; DA-1 re-scored DEAD. See *DA-1* below.

## The defect, measured

On `main`, this whole file:

```java
package com.example;

public @interface Audited {
    String value() default "";
    int level();
}
```

produces **exactly one record — the file entity**. No component for `Audited`,
no entities for its elements. A project's own annotations are structurally
invisible: nothing to find, nothing to point an edge at.

```
=== Audited.java: 1 entities
  SCOPE.Component | file | name="Audited.java"
```

`grep -rn "annotation_type_declaration" internal/extractors/java/` returned zero
hits on `main`.

## Grammar: derived, not assumed

Dumped the CST from the same `ts.Language` handle the daemon parses with
(`tsjava.Language()` via the official adapter), not from memory:

```
annotation_type_declaration          name=identifier   body=annotation_type_body
  modifiers
  @interface
  identifier                         "Audited"
  annotation_type_body
    annotation_type_element_declaration   type=type_identifier  name=identifier  value=string_literal
```

Two facts fall out, and both matter:

1. `annotation_type_declaration` exposes the **same `name` and `body` fields**
   `interface_declaration` does — probed explicitly with `ChildByFieldName`, not
   inferred from the child list. So `buildComponent` applies unchanged.
2. Its members are `annotation_type_element_declaration`, **not**
   `method_declaration`. That is the mechanical reason nothing downstream ever
   saw an annotation element: every member-minting arm in this package keys on
   `method_declaration`.

`go run ./tools/node-type-gate` exits **0** with `0 new` misses — both new
literals resolve against tree-sitter-java's symbol table.

## Decision 1 — what Kind is an annotation type?

**`SCOPE.Component` + `Subtype: "annotation"`.** No `SCOPE.*` vocabulary bump
(which `internal/types/kind_vocabulary_bump_guard_6779_test.go` gates); a new
subtype beside `class`/`interface`/`enum`/`record` is the existing pattern for
"same Kind, different declaration form".

**The decision stands; the reasons I first gave for it did not.** Both were
swept against the actual consumer surface and both are false:

- *"Reusing `interface` would pollute who-implements-this-interface answers."*
  It would not. Those answers are **edge-driven**
  (`docgen/llm_bundle.go:1505-1507`, `mcp/mro.go:438-456`), and an `@interface`
  emits neither `EXTENDS` nor `IMPLEMENTS` — so it contributes **zero rows under
  either subtype**.
- *"A distinct subtype keeps `list this project's annotations` answerable."* It
  does not, today. **`Subtype` is invisible to MCP entirely**: `serializeEntity`
  and `grafel_find` hit rows emit no `subtype` key, no tool filters on one, and
  Java does not dual-stamp `Properties["subtype"]`. That query is unanswerable
  over MCP right now under *any* subtype.

The reasons that **are** real are downstream subtype tables, and they cut both
ways. Stated with their costs:

| Consumer | Effect of `"annotation"` | Verdict |
|---|---|---|
| `engine.ClassLikeComponentSubtypes` (`classfold.go:303-314`) | **absent** → an annotation is never a class-fold *source* | inert today (no typed survivor exists for it to fold into), but a real exclusion, not a neutral one |
| `links/sameas_pass.go` `modelSubtypes` (`:94-104`) | **absent**, and `isDomainModelKind` returns `modelSubtypes[sub]` → excluded from cross-repo `SAME_AS` | deliberate — an annotation is not a shared domain model — but it *is* a cost if two repos ever share one |
| `feedback/report.go` `isFieldExtractionCandidate` (`:1130`) | **NOT** in `nonClassSubtypes` → **admitted** to the field-bearing denominator | correct side: with elements emitted as Subtype `"field"` children, an annotation genuinely *is* field-bearing and can pass |

The third row corrects the review note, which had it as *excluded*. I read the
predicate rather than relaying: `isClassLikeKind("SCOPE.Component")` is true,
`"annotation"` is not in `nonClassSubtypes` (`field`/`column`/`enum`/`const`/
`file`/`import`/`delegate`), `java` is not in `nonFieldBearingLanguages`, and the
`interface` special case does not apply — so it returns **true**.

Net: reusing `"interface"` would have put annotations into the first two tables
and left the third unchanged. A new subtype trades two rows worth nothing here
for a subtype that names what the thing is.

## Decision 2 — what about the elements?

**`SCOPE.Schema` / `field`.** An annotation element is a named, typed,
defaultable attribute that carries data — structurally the same thing as a
record header component, which this extractor already models as
`SCOPE.Schema`/`field` (#1935). It also inherits the #6912 field→declared-type
machinery, so `Tag tag();` inside an `@interface` reaches the project's own
`Tag`.

**The reason I first gave was false and is corrected.** I wrote *"no CALLS edge
can ever reach an annotation element, so an operation row would be uncallable by
construction."* That universal is wrong — see F1 below. The reason that holds is
the **dead-code surface**: `mcp/dead_code.go`'s `isLiveCodeKind` excludes
`"schema"` and admits `"operation"`, **keyed on Kind with no subtype escape
hatch**. Most annotation elements are read only by frameworks via reflection,
invisible to this graph, so as operations they would be reported as dead code
almost universally.

## F1 — the false universal, and the wrong-kind edge it was hiding

`a.value()` **is** a call — invokeinterface on the annotation proxy at runtime —
and the ordinary reflection idiom produces one:

```java
public @interface Audited { String value(); }
public class Reader {
    public String read(Class<?> c) {
        Audited a = (Audited) c.getAnnotation(Audited.class);
        return a.value();
    }
}
```

Measured: `SCOPE.Operation|Reader.read -[CALLS]-> Audited.value`, and it
**binds** — `LookupStatusHint("Audited.value","CALLS")` → status **1**, resolved
onto `Kind=SCOPE.Schema Subtype=field`. Driven by the pre-existing #4682
local-variable-receiver machinery, not by anything in this PR — but **on `main`
the stub was harmlessly unmatched because no entity of that name existed.** This
PR is what turns a dangling stub into a bound CALLS edge pointing at a
Schema/field.

**Decision: accepted and documented, not forbidden.** Reasons, in order of
weight:

1. **A forbidden row would be structurally vacuous where it matters.**
   Suppressing the stub is only possible when the annotation is declared in the
   **same file** as the reader. The dominant real shape has the annotation in
   another file, where the extractor cannot know the receiver's type is an
   annotation at all. A guard that holds in the graded case and is absent in the
   common one is worse than none — that is the #7065 dead-matcher shape pointed
   at my own fix.
2. **The edge is semantically right.** "Who reads the `value` element of
   `@Audited`" is a query a consumer wants, and this is the only edge that
   answers it.
3. **No production consumer keys a CALLS edge on its target's Kind.** Swept the
   `CALLS` readers in `internal/mcp`, `internal/docgen` and `internal/links`:
   they filter on *edge* kind, never on *target* kind. (`isLiveCodeKind` keys on
   the Kind of the node being judged, not of an edge's target.)

Pinned by `TestJavaAnnotationType_7073_ElementIsACallTarget`, which asserts the
edge, its binding, **and the Kind it lands on** — with the element located **by
name across all kinds**, deliberately, so that changing the element's kind fails
the *kind assertion* rather than failing as a *premise* and leaving the kind
claim ungraded. (First draft of this test did fail as a premise; caught by
re-scoring M17 against it, and fixed.) Selectivity control: `c.getAnnotation(…)`
in the same method must stay **unbound**, so "it bound" is not explained by
everything binding.

## F2 — the signature builder corrupted a `default` string literal

`buildAnnotationElementSignature` stripped modifiers with
`strings.ReplaceAll(raw, "public ", "")` over the **whole span**. Unlike
`buildFieldSignature` — which truncates at `=` and so never has a value in scope
— this signature deliberately **keeps** the `default <value>` clause, putting
arbitrary value text in range of the replace.

Measured before the fix:

```
Cfg.scope  sig="String scope() default \"api\""      ← want: default "public api"
```

My own test asserts `"int level() default 0"`, so the default clause is part of
the declared contract and this silently falsified it. Replaced with a
**prefix-trim loop**, which also makes the strip order-independent so
`abstract public` is handled as well as `public abstract`.

## F3 — element spans were ungraded

The *component* span was asserted; the new element builder had no such row.
Spans are persisted through the flatbuffer, hashed by `entityTupleKey`, and
drive `grafel_get_source`, so this is not equivalent to unasserted formatting.
`TestJavaAnnotationType_7073_ElementSpans` closes it, including a **multi-line**
element (`3-5`) so that `StartLine==EndLine` emitted for everything cannot pass.

Note the digest would have caught this **only if** the fixture contained an
`@interface` — it does not, which is exactly why the *enumeration* and not the
digest is what carries this section.

## DA-1 — the signature whitespace collapse was ungraded (round 3)

Deleting `raw = strings.Join(strings.Fields(raw), " ")` from
`buildAnnotationElementSignature` (`java.go:1679`) left
`./internal/extractors/java/` **green — exit 0**, `go vet` clean. Reproduced
here before fixing.

**Not equivalent, and the cause is a fixture shape, not a missing case.** Every
element whose `Signature` this file asserted was a **single-line** declaration —
`String value();`, `int level() default 0;`, `String scope() default "public api"`,
`public abstract String mode()`, `String plain();` — and the collapse is a no-op
on a single-line span, so no assertion could observe the call. The fixture *did*
contain a multi-line element (the one M8 kills on at span `3-5`), but only its
**span** was asserted; its **signature** never was.

So the fixture varied the span axis and held the signature axis constant at
single-line — **one axis pinned, its neighbour left open** — and it landed in the
round whose whole subject was signature correctness.

**Fix: one row.** `…_ElementSpans` now asserts `Signature` alongside the span on
all three rows, and the multi-line row is the only one in this file that can
observe the collapse at all.

**DA-1 re-scored: DEAD.**

```
Spans.two.Signature = "String\n        two()\n        default \"x\"",
                 want "String two() default \"x\""
```

**Occurrence count, and what was actually checked.**
`raw = strings.Join(strings.Fields(raw), " ")` appears **4 times** in `java.go`.
I probed all four; only `:1679` is mine.

| Site | Function | Deleted individually → | Status |
|---|---|---|---|
| `:1679` | `buildAnnotationElementSignature` (**added by this PR**) | suite **fails**, 1 failure | **graded — DEAD** |
| `:520` | `walk` (record-component signature) | `vet` 0, suite **exit 0**, 0 failures | **pre-existing, ungraded** |
| `:1741` | `buildMethodSignature` | `vet` 0, suite **exit 0**, 0 failures | **pre-existing, ungraded** |
| `:1757` | `buildClassSignature` | `vet` 0, suite **exit 0**, 0 failures | **pre-existing, ungraded** |

The three pre-existing sites are the same defect class for the same reason and
are **deliberately not fixed here** — out of scope for #7073, and worth their own
issue. Stated explicitly so nobody assumes a sweep that did not happen.

## F5 — the IMPORTS control did not grade what it claimed

Two problems, both real: nothing else in the fixture was named `Audited`, so
status 2 was unreachable and the "ambiguity control" graded nothing; and the
assertion (that an `IMPORTS` edge *naming* the annotation exists) is **true on
`main`**, since the ToID is raw import text emitted with no reference to any
entity.

Rewritten to grade the resolution step itself:

- **Pre-resolution premise:** the edge must hold the raw text `a.Audited`, and
  must *not* already hold the entity id — otherwise nothing about binding is
  being graded.
- **Post-resolution assertion:** the ToID equals the Audited component's entity
  id. Measured: it does.
- **Selectivity control:** a second import of a **non-existent** type
  (`import a.Ghost;`) must still hold its raw text, so a resolver that rewrote
  every import indiscriminately cannot pass.

## Usage sites — still a PARTIAL fix, stated plainly

- **Annotation APPLICATION** (`@Audited("x")` above a declaration) still emits
  **no edge at all**. This extractor reads annotation nodes for detection
  (`@Transactional`, tracing, Bean Validation) but has never modelled annotation
  *use* as a relationship, so minting the declaration cannot make a non-existent
  edge resolve. **#7073 does not close that.** The assertion fires if it ever
  changes, so the scope claim cannot go stale.
- **Annotation-element READ** (`a.value()`) *does* now produce a bound edge —
  F1. These are different things and the tests keep them apart.
- **The `import` edge** now resolves onto the new entity — F5.

## The red, first

**Round 1** (4 red):

```
--- FAIL: TestJavaAnnotationType_7073_MintsComponent
    want exactly 1 SCOPE.Component named "Audited", got 0. All entities:
      SCOPE.Component|file|com/example/audit/Decls.java
      SCOPE.Component|interface|Plain
      SCOPE.Operation|method|Plain.value
--- FAIL: TestJavaAnnotationType_7073_Elements          (Audited.value / Audited.level: got 0)
--- FAIL: TestJavaAnnotationType_7073_Nested            (Nested: got 0)
--- FAIL: TestJavaAnnotationType_7073_UsageSiteBinding  (premise failed: no Audited declaration to bind to)
```

**Round 2** — F2 was the only *behavioural* defect of the four, so it is the only
one with a red:

```
--- FAIL: TestJavaAnnotationType_7073_ElementSignatureKeepsDefaultLiteral
    Cfg.scope.Signature = "String scope() default \"api\"",
                    want "String scope() default \"public api\""
```

F1, F3 and F5 were **ungraded-but-correct** (F1, F3) or **wrongly-graded** (F5),
not broken behaviour, so their new tests are green on arrival. That is exactly
why each is paired with a mutant below — a test that is green on arrival and
kills nothing would be worthless.

**Round 3** — DA-1 is an ungraded-correct case like F1/F3, so its red is the
mutant's, shown above: with the collapse deleted, `Spans.two.Signature` comes
back carrying the raw newlines and indentation.

`TestJavaAnnotationType_7073_PlainInterfaceUnchanged` was green before and after
in all three rounds — it is the control, not a target.

| Fixture | Axis **varied** | Axes **held constant** |
|---|---|---|
| `javaAnn7073Src` (`Decls.java`) | the declaration keyword only: `@interface Audited` vs `interface Plain` | package, file, `public` modifier, member name (`value`), member declared type (`String`), member arity, nesting depth (both top-level) |
| `javaAnn7073NestedSrc` (`Outer.java`) | nesting depth only | the keyword (`@interface` throughout), package, modifier, member name/type/arity |
| `…_ElementTypeRefs` (`Meta.java`) | the **element's declared type**: same-file class / same-file annotation / primitive | enclosing annotation, file, modifier, element arity |
| `…_ElementSignatureKeepsDefaultLiteral` (`Cfg.java`) | the element's **modifier prefix** (absent/present) × whether the **default literal contains a modifier keyword** | enclosing annotation, file, declared type (`String`), arity |
| `…_ElementSpans` (`Spans.java`) | the element's **line extent** (single-line vs wrapped across three), graded on **both** span and signature (round 3) | enclosing annotation, file, declared type, arity, absence of modifiers |
| `…_ElementIsACallTarget` (`App.java`) | **receiver binding**: a typed local reading an element vs an unbindable JDK receiver | file, package, annotation, element |
| `…_UsageSiteBinding` (`a/`, `b/`) | **target existence**: an import of a real type vs of a non-existent one | consumer file, package, import syntax, annotation applications |
| `TestJavaFieldTypeRefs_AllFour…` (extended) | the target's **subtype**, one row each: class / interface / enum / record / **annotation** | file, host class, field arity, name shape |

## Forbidden rows (the negative direction)

- `Plain` stays **exactly one** `SCOPE.Component`/`interface`.
- `Plain.value` stays **exactly one** `SCOPE.Operation`/`method`, and mints
  **zero** `SCOPE.Schema` rows.
- Converse: `Audited.value` mints **zero** `SCOPE.Operation` rows.
- `Outer` (class enclosing a nested annotation) stays exactly one
  `SCOPE.Component`/`class`.
- `int order();` contributes **no** field-type edge, held in the same
  strict-equality assertion as the two rows that must be present.
- `import a.Ghost;` (non-existent type) must stay raw text after resolution.
- `c.getAnnotation(…)` must stay an **unbound** CALLS edge.

## Mutants scored

Each applied with an **exact substitution count** printed (`SUBS=1`), reverted by
`cp` from a byte-level backup **re-taken this round**, shasum re-verified and
`git diff HEAD` checked — never `git checkout --`.

| # | Mutant | Verdict |
|---|---|---|
| **M1** | the new subtype branch also fires on `interface_declaration` | **DEAD.** 8 failures, incl. this PR's `…_PlainInterfaceUnchanged`, `TestJavaExtractor_InterfaceEntity`, `TestRealClassEntityNotPruned`. |
| **M2** | `case "annotation_type_element_declaration", "method_declaration":` | **did not compile** (`duplicate case`). Not a verdict; recast as M2b. |
| **M2b** | the element builder also runs on every `method_declaration` | **DEAD.** 11 failures. Precise kill: `forbidden: "Plain.value" minted a SCOPE.Schema field (1)`. |
| **M3** | drop `"annotation"` from the field→type allow-list | **DEAD.** Kills at **both** emit sites: `…_ElementTypeRefs` and `TestJavaFieldTypeRefs_AllFour…`. |
| **M8** | collapse every element span to `StartLine:1, EndLine:1` | **was ALIVE → now DEAD.** `…_ElementSpans`: `Spans.one 1-1 want 2-2`, `Spans.two 1-1 want 3-5`, `Spans.three 1-1 want 6-6`. |
| **M13c** | make the modifier-stripping arm a no-op | **was ALIVE → now DEAD.** `Cfg.mode.Signature = "public abstract String mode()", want "String mode()"`. |
| **M16** | restore the F2 defect (whole-span `ReplaceAll`) | **DEAD.** `Cfg.scope.Signature = "…default \"api\"", want "…default \"public api\""`. |
| **M17** | element becomes `SCOPE.Operation`/`method` | **DEAD**, 7 failures. Re-scored after strengthening the F1 pin: it now dies on the **kind assertion** (`the CALLS target is "SCOPE.Operation\|method\|Audited.value", want "SCOPE.Schema\|field\|Audited.value"`) rather than on a premise. |
| **DA-1** | delete the whitespace collapse in `buildAnnotationElementSignature` (`:1679`) | **was ALIVE → now DEAD.** `Spans.two.Signature` comes back with raw newlines and indentation. |

## Digest (`cmd/grafel`)

`go test -count=1 ./cmd/grafel/` — with a **fresh `GRAFEL_HOME` and
`GRAFEL_DAEMON_ROOT`** — exits **0** (188s, round 2; 166s round 1). **The digest
did not move and the constant was not touched.** Round 3 is a **test-only**
change (no non-test file touched), so the digest cannot move and it was not
re-run.

The enumeration behind that: `writeSpanFixture6118`'s mixed-language fixture
contains **zero** `@interface` declarations (`grep -c "@interface"
cmd/grafel/custom_extractor_spans_6118_test.go` → `0`). The changed population
has no member in the fixture, so the delta is empty by enumeration rather than by
assumption.

**Correction (F6):** `entityTupleKey` is at `custom_extractor_spans_6118_test.go:408`,
not `:393`, and it hashes **eight** fields, not seven — I quoted the pre-#6862
tuple and omitted `Language`:

```go
fmt.Sprintf("%s|%s|%s|%s|%s|%s|%d|%d",
    e.Kind, e.Name, e.QualifiedName, e.Subtype, e.SourceFile, e.Language,
    e.StartLine, e.EndLine)
```

plus `R <from> -[Kind]-> <to>` keyed by endpoint content. Entity *Properties* are
still not hashed, so the property-only reasoning is unchanged — but the quoted
tuple is now the one that is actually in the file.

## Exit codes

| Command | Exit |
|---|---|
| `gofmt -l internal/extractors/java/` | **0** (no output) |
| `go vet ./internal/extractors/java/` | **0** |
| `go vet ./internal/... ./cmd/...` | **0** |
| `go test -count=1 ./internal/extractors/java/` | **0** |
| `go test -count=1 ./internal/extractors/... ./internal/custom/...` | **0** |
| `go test -count=1 ./cmd/grafel/` (fresh `GRAFEL_HOME` + `GRAFEL_DAEMON_ROOT`) | **0** |
| `go run ./tools/node-type-gate` | **0** — `0 new` misses |

## On the allow-list widening (kept, per review)

`field_type_refs.go` admitting `"annotation"` stays in this PR. It is **inert
without the entity-minting change**: `javaInFileTypeTargets` has exactly one
caller and gates nothing else, and `Subtype: "annotation"` has exactly one
producer — the change in this PR. Split out, it would be a provable no-op with no
way to write a non-vacuous test.

## Corrections to the first version of this PR body

Recorded rather than deleted, since three of them were stated as facts:

1. *"No CALLS edge can ever reach an annotation element."* **False.** F1.
2. *"Reusing `interface` would pollute who-implements answers."* **False** —
   those answers are edge-driven and an `@interface` emits no such edges.
3. *"A distinct subtype keeps `list this project's annotations` answerable."*
   **False** — `Subtype` is invisible to MCP; that query is unanswerable today
   under any subtype.
4. The `feedback` denominator effect is **admission**, not exclusion (this one
   corrects the *review note*, verified against the predicate).
5. `entityTupleKey` is `:408` and hashes **eight** fields including `Language`.
6. *"An annotation body holds no methods, no throw sites, no call sites."*
   **Overstated** — JLS 9.6 permits nested type declarations inside an
   annotation body, which can hold all three. Benign for the decision it
   supported (leaving `walkClassTx` / `exception_flow.go` / `references.go`
   declaration lists untouched: the generic recursion still descends into a
   nested type, and those passes key on the nested declaration itself), but it
   was another universal and should not have been written as one.

## Scope deliberately NOT taken

- **Annotation *application* is still edgeless.** Worth its own issue.
- `walkClassTx`, `exception_flow.go` and `references.go` keep their own
  declaration lists unchanged.
- **The three pre-existing ungraded whitespace collapses** (`:520`, `:1741`,
  `:1757`) — measured and reported in DA-1 above, not fixed here.


cajasmota:
**Round 3 verified independently. Merging.**

I re-scored **DA-1** myself at `82fd6b881` rather than taking the report: deleting the collapse at `java.go:1679` now fails with

```
Spans.two.Signature = "String\n        two()\n        default \"x\"", want "String two() default \"x\""
```

**DEAD.** `java.go`'s shasum is byte-identical to round 2's (`016180b2…`), which independently confirms this round is test-only and the digest cannot have moved. Restored from a round-3 backup, `shasum -c` OK, `git diff HEAD` empty.

**You did more than I asked and it turned a scope note into a measurement.** I said the other three occurrences were out of scope and only wanted to know which you checked. You probed all three — ~0.7s each — and reported exit codes: `:520`, `:1741`, `:1757` all **ungraded**, each deletable with the suite still at exit 0. Filed as **#7091**, with your priority call recorded: `buildClassSignature` first, because a wrapped `class Foo extends Bar implements Baz` is ordinary Java while a wrapped annotation element is rare, and `Signature` is what a reader sees in `grafel_inspect`.

The root cause is one thing rather than three, which is the part worth keeping: **nothing in this package asserts a multi-line signature anywhere**, so collapsing whitespace on a single-line span is a no-op that no number of `Signature` assertions can observe.

**Your account of the original miss is the clearest statement of this milestone's dominant defect I have read.** The multi-line element already existed — you wrote it in round 2 specifically so M8 could not pass on `StartLine == EndLine` — and you asserted its span and never its signature, in the round whose subject was signature correctness. Not a missing fixture: a fixture built for one axis and never read against its neighbour.

**Across three rounds this PR corrected two of my errors and one of the reviewer's.** The `nonClassSubtypes` inversion (an annotation is *admitted* to the field-bearing denominator, not excluded) I relayed without checking; the "list this project's annotations via Subtype" rationale was mine and is false, since `serializeEntity` emits no subtype key at all — now filed as **#7090**, because the general form of that assumption is load-bearing in more than one place. Both were caught by reading the predicate instead of the summary.

**F1 stands as accept-and-pin**, for the reason you gave: a forbidden row would be structurally vacuous where it matters, since suppression is only possible same-file while the dominant shape is cross-file — a guard that holds in the graded case and vanishes in the common one, aimed at your own fix. And your self-catch that the first F1 pin was dying as a *premise* failure, leaving the kind claim ungraded by its own test, is the kind of thing that otherwise ships looking green.

**The allow-list widening stays in**, per the reviewer's reasoning rather than mine: `javaInFileTypeTargets` has one caller and `Subtype: "annotation"` has one producer, so split out it would be a provable no-op with no way to write a non-vacuous test.

PR #7092, opened 2026-09-16
feat(#7056): a measured precision defect could only be fixed now or not written down — forbidden hits were unconditionally fatal

cajasmota:
Refs #7056 (group 1). **Mechanism only — no existing fixture row is converted.**

## The defect

`forbidden_relationships` is the only mechanism in this repo that can say *"this edge is wrong"*, and a hit was **unconditionally fatal** (`ratchet.py`, edges and entities alike). So a precision defect that had been found, measured, and could not be fixed today had exactly two fates: **fixed now, or not written down**. Findings evaporated instead of accumulating.

The tree already carried the receipts. `golden/solidity-mini/NOTICE.md` records two real, measured, shipping defects that were deliberately kept **out** of `expected.json` while they fired, because committing them would have broken the gate rather than recorded a gap — they lived as markdown prose, invisible to every instrument, until a fix happened to land. `baseline.json`'s own `measured_on_note` says the same of #6425's phantom `Vault.helper` Yul operation.

## The shape, and what in the code drove it

A **per-row `known_bad` flag on a forbidden row** (edge or entity), with a **mandatory `issue` and `note`**. The row still fires; what changes is where the hit is counted — out of `forbidden_hits` (fatal) and into `known_bad` (recorded, printed by name on every run **including green ones**).

### 1. Should a forbidden count go in `baseline.json`? **No — and that is the whole design.**

The brief flagged that there is no forbidden count in `baseline.json` today, so known-bad would default to "unmeasured" unless one were added. The counter-argument came from the code: `baseline.json` is **generated**, and `--update-baseline` rewrites it from observation. A known-bad count recorded there would be re-recorded as the new normal the first moment a row stopped firing — and that is not hypothetical, it is the **#6273 shape** this file already carries a scar from: `expectations_missing` was a flag in the generated baseline that documented a gap and then justified it, surviving every re-record for two whole fixtures.

So the floor is not recorded at all. It is **declared**, in the fixture's own `expected.json` — hand-written, reviewed, and a file that **no mode of `ratchet.py` ever writes to**. The gate compares the declaration against what the run observed. `--update-baseline` therefore has nothing to launder: the number it would have to weaken is not in a file it can touch, and deleting a `known_bad` row is a fixture edit a reviewer sees.

`ratchet.py` reads the fixture's `expected.json` itself and cross-checks the declared count against the report, so a **stale binary** that predates the channel cannot silently erase the finding either (the #6488 arm D reasoning about an absent extracted total, on this key). Pinned by `test_a_stale_binary_cannot_erase_a_declared_row`.

### 2. Stopping `--update-baseline` from re-recording a known-bad as the new normal

Two things, neither of which is a `carry_*` function — there is nothing to carry:

* nothing about the channel is in `baseline.json`, so there is no figure to weaken (`test_the_written_baseline_records_no_known_bad_figure` asserts the per-fixture entry did not grow a key);
* `build()` **refuses to re-record at all** while a declared row is silent, on the same terms it already refuses a report with no extracted total. A re-record taken in that state would leave the operator believing the tree is recordable.

## Grading — both directions

**Direction 1 (permissive).** A firing `known_bad` row passes, and the gate's **emitted output** names the row, the issue **and** the note — on the success path as well as the failure path, mirroring `format_known_regressions`. Asserted against the printed text, never the exit code or an internal counter. Backed by a positive control (`test_positive_control_the_same_row_without_known_bad_is_fatal`, `TestPositiveControl_TheSameRowsWithoutKnownBadAreFatal_7056`) so that a change which simply stopped reading `forbidden_hits` cannot score as a pass.

**Direction 2 — the one that would be skipped.** A `known_bad` row that **stops firing** goes **RED** and names the row, its issue, and both readings (fixed vs. dead producer). Graded by `KnownBadIsNotAOneWayValve` on the `check` side and by `KnownBadSurvivesTheMerge` on the aggregator side — **two graders**, because `run.sh`'s merge program is the *other* consumer and strict mode reads its exit code and nothing else.

**Validation.** A `known_bad` row with a missing `issue` or `note` is rejected at load, exactly as `TestKnownRegressionsAgreeWithRecordedFloor` rejects one today. So is `known_bad` on a recall row (honoured by nothing there) and `issue` on a row that is not `known_bad`.

## Aggregation — #7084 is respected, not re-opened

* **firing** rows are **unioned** across runs — a row seen by any run fired (#7084's reason);
* **silent** rows are **intersected** — only a row *no* run saw fire has stopped firing. A union there would red the gate on every intermittent producer; a median would be #7084 re-entered on the key that grades the other direction;
* a row observed **silent in any run vetoes the recall short-circuit**, so an intermittent producer cannot be declared dead on three samples. The veto can only make the loop run **longer**. No median, no early exit added.
* the counters follow the extracted-total treatment: absent in any run means the key is **dropped**, never defaulted to 0.

`extract_aggregator` / `extract_stability`'s token floors were widened to `known_bad_silent` / `silent_in_every_run`, so a slice that lost this code fails loudly rather than grading nothing.

## Candidate rows for the follow-up (deliberately untouched here)

* `golden/solidity-mini/NOTICE.md:208-215` — `Vault.ceiling -> type` and `Vault.fingerprint -> encode`, two measured shipping defects parked in prose.
* `baseline.json`'s `measured_on_note` — #6425's phantom `Vault.helper` Yul operation, recorded nowhere gradeable.
* `golden/vbnet-mini/expected.json:1` — the row whose own header says *"THIS ENTRY PINS A TRADEOFF, NOT A TRUTH"*. Note this one is an inverted case: it is a **must_exist** row pinning an edge the ideal extractor would *not* emit, so converting it is a re-authoring rather than a flag flip.

Each needs its own measurement; mixing them here would mean a red gate could not be attributed.

## Verification

| check | exit |
|---|---|
| `python3 -m unittest scripts.quality.test_ratchet` (79 tests) | **0** |
| `go test ./internal/quality/... -count=1` | **0** |
| `go vet ./...` (whole repo) | **0** |
| `go test ./cmd/grafel/ -run 'Quality\|Forbidden\|Subtype\|PlaceholderAnchor\|Vbnet'` | **0** |
| `bash -n scripts/quality/run.sh` | **0** |
| `gofmt -l` | clean |

`./cmd/grafel` was run despite the brief expecting it to be unnecessary: `cmd/grafel/quality_*_test.go` drive `grafel quality` end to end over the JSON report shape and `LoadFixture` validation, both of which this change moves. Only the quality-related tests were selected.

**Not verified, and stated plainly:** the full quality gate (`scripts/quality/run.sh`) was **not run** — it is compute-heavy and a live daemon on this machine is serving MCP. Everything asserted about the aggregator is asserted by executing `run.sh`'s **real bytes**, sliced out by `extract_aggregator` / `extract_stability`, not a copy; what is untested is the surrounding shell (the run loop, the mode dispatch, the exit-3 bail), which this change does not touch. Since no fixture declares a `known_bad` row yet — pinned by `TestNoGoldenFixtureDeclaresKnownBadYet_7056` — the gate's behaviour on the real corpus is unchanged by construction.

## Mutants (all DEAD; every occurrence of each mutated expression scored)

| # | mutation | occurrences | verdict |
|---|---|---|---|
| M1 | `ratchet.py`: silent-row loop made empty | 1 | **DEAD** ×4 — this is the direction-2 RED: without it the channel is a one-way valve |
| M2 | `run.sh`: silent intersection → union | 1 | **DEAD** — `test_a_row_firing_in_a_minority_of_runs_is_alive` |
| M3 | `run.sh`: strict-mode `_silent` arm removed | 1 | **DEAD** — the aggregator is the other grader |
| M4a/M4b | `diff.go`: `fb.KnownBad` / `fe.KnownBad` disabled | 1 each, **both arms scored** | **DEAD** ×4 each |
| M5 | `expected.go`: `note == ""` arm | 1 | **DEAD** |
| M6 | `expected.go`: `issue == ""` arm | 1 | **DEAD** |
| M7 | `expected.go`: `knownBad && !forbidden` arm | 1 | **DEAD** |
| M8 | `expected.go`: `!knownBad && issue != ""` arm | 1 | **DEAD** |
| M9b | `run.sh`: short-circuit veto disabled (token kept, so the slice guard is not what kills it) | 1 | **DEAD** — isolated to `test_a_silent_known_bad_row_vetoes_the_short_circuit` |
| M10a/M10b | `report.go`: `kb.Fired` → `true` | **2 occurrences** (`ToJSON`, `WriteHuman`), both scored | **DEAD** each, by different tests |
| M11 | `ratchet.py`: `build()` re-record refusal disabled | 1 | **DEAD** ×2 |
| M12 | `ratchet.py`: stale-binary cross-check disabled | 1 | **DEAD** ×2 |

The compound guards were scored **arm by arm** (M5–M8 are four arms of one `checkKnownBad`); M9's first attempt tripped the heredoc token floor and killed 7 tests including unrelated ones, so it was re-run as M9b with the token preserved, isolating the veto. Every mutant was applied by exact-string replacement with an asserted occurrence count, reverted by `cp` from a byte-level backup, and the working tree was re-verified against the backups by `diff -q` before committing.


cajasmota:
## Round 2 — both findings closed, plus three more twins the audit turned up

### F1 — the entity half of `declared_known_bad()` (fixed)

Reproduced exactly as reported: dropping `"forbidden_entities"` from the tuple at `ratchet.py:203` ran the **full suite green, exit 0, zero failures**. Root cause confirmed as diagnosed — `declare_known_bad()` wrote only `forbidden_relationships` and `kb_row()` defaulted to `cls="relationship"`, so no python test anywhere built an entity-class row.

Every known-bad python case is now parameterised over `KB_CLASSES = ("relationship", "entity")`, not just the stale-binary one — `check`, `build`, the aggregator→`check` end-to-end, and the short-circuit veto.

**Added a positive control first**, `DeclaredKnownBadReadsBothClasses`, because the parameterisation is worthless if the entity fixture does not actually declare anything: it asserts `declared_known_bad() == 1` for each class *and* that the two classes write **different fixture keys**. Without it every entity subTest would pass against a fixture that cannot express the defect — the same shape that hid this finding.

Re-scored: **`R1` entity-tuple → DEAD**, killed by 5 distinct cases, all on the `cls='entity'` subTest, including `test_a_stale_binary_cannot_erase_a_declared_row`.

**Veto scope**: `class == "relationship"` narrowing re-scored **DEAD**, killed solely by `test_a_silent_known_bad_row_vetoes_the_short_circuit (cls='entity')`. Saying it explicitly as asked: its blast radius is a **false RED** (an intermittent entity producer gets 3 samples instead of 5 and could be declared dead), never a false GREEN, and it is covered by the same entity-class case rather than one of its own. The test's docstring now says that too.

### F1 deliverable — the key-pair list

I grepped my own non-test diff for every site where the two classes are handled. **Six pairs. Three were ALIVE.**

| # | site | pair | verdict before | now |
|---|---|---|---|---|
| P1 | `ratchet.py` `declared_known_bad` tuple | `forbidden_relationships` / `forbidden_entities` | **ALIVE** (F1) | **DEAD** ×5 |
| P2 | `run.sh` union loop tuple | `forbidden` / `forbidden_entities` / `known_bad` | DEAD | DEAD — `test_the_firing_union_is_not_inherited_from_one_run` |
| P3 | `diff.go` `Class:` literals | `"entity"` / `"relationship"` | **ALIVE** | **DEAD** — class is now asserted per row, with two *different* expected values |
| P4 | `expected.go` `checkKnownBad` call sites | 2 recall arrays × 2 forbidden arrays | **ALIVE** on `forbidden_entities` | **DEAD** ×3 — validator table now runs over both forbidden arrays |
| P5 | `expected.go` `forbiddenTwin` | `expected_entities`→`forbidden_entities` / else | **ALIVE** | **DEAD** ×2 — the error must name the *right* target array |
| P6 | `report.go` `r.KnownBad` loops | `ToJSON` / `WriteHuman` | DEAD (round 1, disjoint killers) | unchanged |

P4 is the one worth flagging: **a known_bad *entity* row with no issue and no note was accepted at load** — the decorative-row state the validator exists to reject, on the class that had no test. P5 is lower severity but the same family as #6476: the diagnostic told an author to move an entity row to `forbidden_relationships`, i.e. it pointed the reader at the wrong thing in a message a confused author follows.

`TestLoadFixture_IssueWithoutKnownBadIsRejected_7056` was parameterised over both forbidden arrays for the same reason, even though its arm was already scored on the relationship side — an unexercised call site is a deletable call site.

### F2 — the vacuity floor (fixed)

Confirmed ALIVE: truncating the walk after the tenth fixture passed. Changed `scanned < 10` → `scanned != 38`, with a comment citing the sibling convention and its reasoning (a `<` floor accepts every number above it, including a truncated walk's). Re-scored **DEAD**, and the failure names the number: *"scanned 10 … want 38 … the walk is truncated, in which case every unscanned fixture is ungraded by this test"*.

Re-checked the planted-row control after the change: a `known_bad` row planted into `golden/proto-mini/expected.json` still reds the test and names the file; the fixture was restored and `git diff` over `golden/` is empty.

### The asymmetry, stated

Added to `declared_known_bad()`'s docstring and repeated here: the declaration **cannot be weakened by any machine** in this repo — that is the property the design rests on — but it **can be deleted by a human**, and code review is the only thing in the way. This is deliberately **weaker than the recall side**, where `baseline_test.go` carries `mustAnnotate`, a closed set in Go, precisely so the annotation and the thing it annotates cannot be removed in one edit. It is exactly as strong as an ordinary forbidden row, which anyone can also delete. The channel makes a finding **recordable, not undeletable** — the round-1 body over-claimed by implying otherwise.

### Verification

| check | exit |
|---|---|
| `python3 -m unittest scripts.quality.test_ratchet` (**81** tests, was 79) | **0** |
| `go test ./internal/quality/... -count=1` | **0** |
| `go vet ./internal/quality/` | **0** |
| `bash -n scripts/quality/run.sh` | **0** |
| `gofmt -l internal/quality/` | clean |

Backups were re-taken **this round** before any mutation and again after each test-file edit; every mutant was applied by exact-string replacement with an asserted occurrence count, reverted by `cp`, and verified by `shasum` plus a `git diff HEAD` showing only the three intended files. `git checkout --` was not used. The full quality gate was not run.

cajasmota:
## Round 3 — DB-1 closed, and the `:259` twin had the same exposure (two role pairs, not one)

### The red

DB-1 reproduced exactly: swapping `issue`/`note` on `ratchet.py`'s green-print line ran **all 81 tests green, exit 0**, and parsed cleanly — not a broken-mutant artefact. The emitted line, measured both ways:

```
HEAD     - solidity-mini: relationship Vault --[CALLS]--> helper  [#6425]  phantom yul operation minted from an assembly block label
MUTANT   - solidity-mini: relationship Vault --[CALLS]--> helper  [phantom yul operation minted from an assembly block label]  #6425
```

Observable, non-equivalent, and it degrades precisely the affordance print-on-success exists to give.

### What I found at `:259` — **yes, the same exposure, and worse**

You asked me to look rather than assume. The `STOPPED FIRING` line carries no note, so there is no issue/note pair to confuse — but it has **three** interpolations (`class`, `label`, `issue`) in one line and the suite asserted only that each string was *present*. **Two** role pairs were ALIVE:

| mutant | before | now |
|---|---|---|
| green line, `issue` ↔ `note` (DB-1) | **ALIVE**, 81 green | **DEAD** ×2 (both classes) |
| `:259` red line, `label` ↔ `issue` | **ALIVE**, 81 green | **DEAD** ×2 (both classes) |
| `:259` red line, `class` ↔ `label` | **ALIVE**, 81 green | **DEAD** ×2 (both classes) |

One process note, since it nearly cost a wrong verdict: my first attempt at the `class`↔`label` mutant cut across a string-literal boundary and left the file unparseable. That scored `exit 1` with no `FAIL:` lines — a **broken mutant reading as a kill**. I caught it on the output rather than the exit code, added an `ast.parse` check to the mutation step, and re-ran it as a valid mutant, which came back ALIVE. Reported here because "exit 1 means dead" is exactly the shortcut that would have hidden it.

### The fix

Both surfaces are now asserted on the **assembled substring, in order, with the brackets where they belong** — not on the presence of the parts:

```python
want = f"{cls} {kb_label(cls)}  [{KB_ISSUE}]  {KB_NOTE}"                      # green
want = f"known_bad {cls} {kb_label(cls)} STOPPED FIRING [{KB_ISSUE}]"         # red
```

plus an explicit negative — `assertNotIn(f"[{KB_NOTE}]", text)` — because the positive alone would still hold if the bracket contained the issue *and* the prose. The bracket is a tracking-id slot, and that is stated in the test rather than left to be inferred.

Both parameterised over `KB_CLASSES`, consistent with round 2. Each test asserts its own premise first (`rc == 0` for the green case, `rc == 2` for the red one) so neither can pass by grading the wrong run.

### The sentence you asked for, now in the test docstring

**`ratchet.py`'s green print and `internal/quality`'s `WriteHuman` are two different green-print surfaces and each needs its own content assertion.** One is what CI reads; the other is what a developer running `grafel quality` by hand reads. `TestKnownBadIsPrintedOnAGreenHumanReport_7056` covers the second and never touched the first — which is how this survived. Neither substitutes for the other.

The asymmetry that let it live is also worth naming: the python gate asserted this surface only for **absence** (`assertNotIn("known-bad", text)` on a no-rows fixture) and on the RED path. A surface graded only for absence is graded for one bit.

### Verification

| check | exit |
|---|---|
| `python3 -m unittest scripts.quality.test_ratchet` (**83** tests, was 81) | **0** |
| `go test ./internal/quality/... -count=1` | **0** |
| `bash -n scripts/quality/run.sh` | **0** |

Round-3 byte backup re-taken before any mutation and again after the test edit; every mutant applied by exact-string replacement with an asserted occurrence count **and an `ast.parse` validity check**, reverted by `cp`, and `diff -q` against the backup confirms `ratchet.py` is byte-identical to its pre-mutation state. `git diff HEAD` before committing showed **one file changed, +79, test-only**. No `git checkout --`. Full quality gate not run. Nothing widened beyond DB-1 and the `:259` twin you asked me to check.

cajasmota:
**Round 3 verified independently. Merging.**

I re-scored **DB-1** myself at `d2a037973`: swapping `issue`/`note` on the green-print line now fails with **2 real `FAIL:` lines**, one per class. `ratchet.py`'s shasum is byte-identical to round 2's, independently confirming this round is test-only. Restored by `cp`, `shasum -c` OK, `git diff HEAD` empty.

I scored it the way your own process note earned: **validated the mutant parses before running it, and counted `FAIL:` lines rather than trusting the exit code.**

**That process note is the most valuable thing in this PR, and it is going into the standing rules.** Your first `class`↔`label` swap cut across a string-literal boundary, left the file unparseable, and scored `exit 1` with **zero `FAIL:` lines** — a broken mutant reading as a kill. You caught it on the output rather than the exit code, added an `ast.parse` gate, and the valid re-run came back **ALIVE**. Trusting the exit code would have reported `:259` as fine and I would have merged with two live role swaps in the anti-rot output.

It is the exact mirror of a trap already recorded here: a *non-applied* edit reads as ALIVE, a *syntactically broken* edit reads as DEAD, and the gates are mirrors — count-assert on one axis, parse/vet on the other. Both are now written down together.

**My `:259` question was worth asking and I nearly did not.** It found **two more ALIVE** — `label`↔`issue` and `class`↔`label` — so three ungraded role swaps sat in the output whose only job is that a reader can see which issue tracks which known-bad row.

The fix shape is right: assert the **assembled substring in order, brackets included**, plus an explicit `assertNotIn` of the swapped form, because a positive-only check would still hold if the bracket contained the issue *and* the prose. Each test asserting its own premise (`rc == 0` green, `rc == 2` red) means neither can pass by grading the wrong run.

**Across three rounds the audits mattered more than the fixes.** Round 2's key-pair list — which I asked for instead of the single reported fix — turned up **three more ALIVE that nobody was looking for**, two of them real defects rather than grading gaps: a known_bad **entity** row with no issue and no note was *accepted at load* (the decorative-row state the validator exists to reject), and a load error told an author to move an entity row to `forbidden_relationships`. Neither would have surfaced from fixing F1 alone.

And adding `DeclaredKnownBadReadsBothClasses` **before** parameterising is the right instinct stated exactly right: otherwise every entity subTest passes against a fixture that cannot express the defect — the shape that hid F1 in the first place.

**The design decision holds, and it was verified rather than accepted.** Declaring the floor in the hand-written `expected.json` rather than the generated `baseline.json` is correct: `ratchet.py` has exactly one write site, `expected.json` is opened read-only, and no non-test Go file writes it — so "nothing writes it" holds unconditionally, stronger than the body claimed. The #6273 precedent you cited is the real thing: `expectations_missing` was a flag in the generated baseline that documented a gap and then justified it across every re-record.

The deletion asymmetry is now stated rather than implied: no machine here can weaken the declaration, but a human can delete the row and review is the only thing in the way — **weaker than `mustAnnotate` on the recall side**, equal to an ordinary forbidden row. Saying so is better than having it discovered.

Issue #7071, opened 2026-09-16
NINETEEN lexical-guess tiers bound a bare name with no marker, so a guess was indistinguishable from a real resolution on every instrument we own


cajasmota:
A `CALLS` edge that the resolver binds via the **bare-name locality tier** — the same-file / same-package-dir leaf-name suffix match — is written into the graph **byte-identical to a bind by exact qualified name**. No confidence, no property, no per-edge disposition. A downstream consumer cannot tell `Process → OrderService.Process (bound because it was the only same-named method in the caller's file, with the receiver's type never known)` from `OrderService.Process → … (bound on the qualified name the extractor actually resolved)`.

This came out of checking the downstream question #7068 raises and explicitly does **not** claim.

## What #7068 feared, and what is actually true

#7068 worried that a bare `CALLS -> Process` "silently binds to an unrelated same-named method". **On the multi-candidate path that is not what happens** — the resolver refuses to bind, at every tier, and the refusals are order-independent:

- `indexByName` (`internal/resolve/refs.go:1989-1994`) **deletes** the `byName` entry on a real collision and sets `ambigName[name]`, rather than keeping an arbitrary winner.
- `uniqueMatchInFamily` (`refs.go:2732`) returns `("", false)` the moment two distinct ids match, walking a fixed slice, not a map.
- `scanLeafMembers` (`refs.go:3839`) is the one function that ranges a map, and it is order-insensitive by construction: a blank sentinel within a scope, or two scopes disagreeing, both return `("", false)`.
- `byMember` / `byPackageMember` / `byNamespaceMember` all store a blank-string collision sentinel on duplicate keys (`refs.go:1381-1388`, `refs.go:1410-1414`), and every reader treats `""` as refuse.
- `ResolveCSharpCrossNamespaceCalls` (`imports.go:2524-2552`) bails on `resolved != id`.

Pinned by existing tests: `TestLeafNameTier_TwoOperationsStillAmbiguous_6141` (`internal/resolve/leafname_kind_filter_6141_test.go:234`), `TestLeafNameTier_SameScopeDuplicateStaysAmbiguous_6141` (`:456`), `TestReferences_OperationBareQualified_4554_Ambiguous` (`internal/resolve/operation_bare_qualified_4554_test.go:62`), and the `ambig-bare-no-hint` / `ambig-bare-hint-fail` subtests at `internal/resolve/refs_test.go:1056-1090`.

[...]

Issue #7068, opened 2026-09-16
csharp.go:967 matches `for_each_statement`; the grammar spells it `foreach_statement` — so NO foreach loop variable is ever typed, and every method call on one emits a bare-name CALLS edge


cajasmota:
`internal/extractors/csharp/csharp.go:967` matches the node type **`for_each_statement`**. The grammar spells it **`foreach_statement`**. The literal is unpaired — no arm anywhere matches the correct name — so `collectLocalVarTypes` **never types a `foreach` loop variable**, and every method call on one loses its receiver type.

## Proven by running the extractor, not by reading

```csharp
foreach (Order o in orders) { o.Process(); }   →  CALLS -> Process        // bare
Order p = GetOrder(); p.Process();             →  CALLS -> Order.Process  // typed
```

Correcting the single literal locally makes both typed. Nothing else changes.

## Why this is worth more than a one-character fix

The direction is **under-emission**, which is the quieter half of #7056 — but the residue is not "no edge", it is a **bare-name target**. `CALLS -> Process` is a well-formed edge pointing at a name with no owner, so it is exactly the late-bind shape one hop downstream: a resolver asked to bind `Process` has no type to disambiguate with, and whatever it picks will look like a success to every instrument we have. Bind rate, orphan rate and dangle count all report an edge was produced.

`foreach` is not a corner of C#. Any collection-iterating method call in the corpus has been losing its receiver type for as long as this literal has been wrong.

## How it was found, and the part that generalises

Found by the **#7065 audit**, which resolved every tree-sitter node-type literal in `internal/` and `cmd/` against the real grammar symbol tables. The audit's own positive control matters here: re-introducing scala's known dead pair (`variant_type_parameter` / `type_parameter`) raised scala's miss count 5 → 7 and flagged both at the right line, so a clean result would have meant something. It did not come back clean.

Surface measured: **2844 literal sites, 575 distinct, 156 files** in `internal/extractors` — 2.6× the grep-visible surface, because `switch node.Type()` (1521 sites) and derived helper sinks (223) dominate and grep sees neither. This defect lives in a `switch` form, which is why three years of grepping never surfaced it.

**42 genuine dead literals in total.** This is the only one with observable behaviour: 9 are in unreachable code (groovy's `buildMethod` path, proto `required`, python `identifier_pattern`), and 32 are dead but **paired with the correct name in the same arm** — harmless today, and a good argument for the baselined gate rather than a blanket one.

## What is NOT claimed

[...]

Issue #4685, opened 2026-06-10
Generalize local-variable receiver typing + test→CALLS coverage crediting — C#/.NET (xUnit/NUnit/Moq/WebApplicationFactory)


cajasmota:
Per-language child of #4672. Generalize the local-variable receiver typing + test→CALLS→handler coverage crediting landed for TS/JS in #4680 (PR) / #4671 to **C#/.NET**.

## Concrete construction + route-hit idioms (from #4672)
- **xUnit/NUnit unit:** `var c = new XController(dep); c.Method();` — type the local `c` from the constructor and resolve `c.Method()` to the class method.
- **Moq:** `new Mock<IService>()` deps passed into a constructed controller; type the controller local from its constructor.
- **Integration:** `WebApplicationFactory<Program>` + `HttpClient` — `client.GetAsync("/path")` links the route-hit to the endpoint.
- **ORMs:** EF Core `DbContext` / `DbSet` query reads.

## Acceptance
(a) `new X()` locals resolve `var.Method()` to the class method (CALLS edge from the `[Fact]`/`[Test]` method scope); (b) WebApplicationFactory/HttpClient route hits link to the endpoint; (c) ComputeCoverage credits via test→CALLS→handler. Exact-mirror fixture + coordinator live-validation on a C# corpus (aspnetcore-realworld / aspnetcore-mvc). Honest exclusion preserved.

Part of #4672.

PR #7093, opened 2026-09-16
fix(#7072): the C# local-type map was flat and last-writer-wins, so two same-name locals in sibling scopes gave one of them a confidently WRONG receiver

### LINKED ISSUE #7072: csharp: the per-method local-type map is flat and last-writer-wins, so two sibling-scope locals sharing a name emit a CALLS edge with the WRONG receiver type — and it binds

cajasmota:
## Round 2 — review findings F1/F2/F3

Review found the #7072 signature still live in the #7072 fix, plus a test whose
name claimed more than its body, plus a comment whose stated premise the same
function contradicted sixty lines down. All three addressed and graded.

F1 (blocker) — UNTYPED DECLARATIONS NOW POISON THE NAME.

Two paths reached the ambiguity ledger's `continue` before the ledger:
a `var` whose initialiser defeats inferImplicitLocalType, and a declared type
leafTypeName has no case for (a `ref` local's ref_type). Both skipped the
declarator outright, so the name was never recorded:

    { Order o = MakeOrder();    o.Ship(); }
    { var   o = MakeCustomer(); o.Ship(); }   // factory RHS => untyped

`o` stayed a clean single-typed entry, and `Order` was applied to the second
site as well — a wrong dotted receiver on a REAL same-file type, which binds
and scores as a success on every instrument. That is the exact defect this
change was opened to remove, reproduced inside it.

Decision: POISON, not pin-as-known-wrong. "We could not type this
declaration" is not "no declaration happened here" — the name IS bound at that
site by something, and a type we derived elsewhere provably does not describe
it. The ledger now records such names with an empty type, which disagrees with
every real type, so the refusal holds in either source order. This is the same
direction as the rest of the change and the cheaper of the two defensible
answers to keep honest; pinning it would have left a binding wrong edge in
place behind a test that says so.

The `foreach` twin reasons about empty types the other way on purpose, and it
can: `claimed[]` (csNamesBoundOutsideForeach, built from the SOURCE) is its
backstop. The locals pass has no backstop, which is why it must ledger the
name itself.

Corrected the false claim that made this hole readable as intended behaviour.
csharp.go's #4685 note said a factory RHS "stays UNTYPED so the call resolves
to its bare leaf rather than a fabricated receiver type". True of the
declaration in isolation, false of the method whenever a same-name sibling
existed — which is this issue's whole subject. The neighbouring sentence was
corrected in round 1 and this one was left; it now says what holds and at what
scope.

F2 (blocker) — `params win over locals` IS NOW PINNED.

Inverting the merge left the suite green. The test named for that precedence
used parameter `o` and locals named `p`: the maps never shared a key, so the
line was never exercised, and the test's own comment conceded as much. Classic
fixture-label-over-content.

TestCSharp7072_ParamsWinOverALocalOfTheSameName adds a parameter and a local
that genuinely collide, and the inverted-merge mutant is now DEAD. Whether such
a program compiles (CS0136) is marked DERIVED FROM THE SPEC AND UNDERIVED BY
EXECUTION — no C# compiler exists in this environment — but tree-sitter parses
it and the merge has to answer, so the answer is pinned where the code claims
it. The old test is renamed TestCSharp7072_RefusalLeavesAnUnrelatedParamAlone
and its comment now says what it does and does not grade.

RECORDED, NOT FIXED HERE: internal/extractors/java/java.go:948 carries the
identical `// params win over locals` line and is equally unprobed. Flagged for
its own issue.

F3 — THE TWO-STAGE FORM'S RECORDED REASON WAS WRONG.

It said a one-stage delete-on-collision was impossible "because the `foreach`
arm reads `out`". It does not — it WRITES `out` and gates on feAmbiguous plus
csNamesBoundOutsideForeach, having deliberately dropped its `out` lookup, which
the #7068 note in this same function states outright. The comment invented an
ordering dependency that does not exist and that nothing grades.

The form is kept, with the real reason recorded: the ambiguous SET is the only
thing that remembers a name ONCE disagreed. Deleting the entry from `out`
destroys that memory, so a third declarator agreeing with the first finds no
entry, sees no conflict, and RESURRECTS the wrong binding. That is not an
assertion — the one-stage delete-on-collision form was implemented as a mutant
and the stickiness test kills it.

ALSO REMOVED: a redundant guard of my own. The publish loop had
`|| typ == ""` beside the ambiguity check. It scored ALIVE, and so did
deleting it together with receiverTypeName's `t != ""` mask — an empty
receiver type is absorbed by csharpCallTarget's `recv == ""` fallback, which
the entire bare-leaf half of this suite grades. This file's own rule for that
case, quoted from the #7068 note: "a redundant guard fires only where the real
one already fires and leaves both ungraded." Deleted rather than left in front
of the real one.

GRADING. 12 mutants this round, every one `go vet`-gated BEFORE scoring and
scored by counting `--- FAIL` lines rather than the bare exit code, so a
non-compiling mutant cannot read as a kill. Two were found ALIVE and both were
acted on rather than reported.

  var-poisoning never fires        DEAD, 3   (F1, both source orders)
  var-poisoning always fires       DEAD, 4
  untypeable-declType skip back    DEAD, 2   (the `ref` local row)
  publish drops ambiguity arm      DEAD, 6
  publish drops `typ == ""` arm    ALIVE  -> guard deleted as redundant
  that arm + downstream mask both  ALIVE  -> confirmed the real guard is
                                             csharpCallTarget's `recv == ""`
  downstream mask alone            ALIVE  -> same finding, other direction
  params merge INVERTED            DEAD, 1   (F2, was ALIVE before this round)
  one-stage delete-on-collision    DEAD, 1   (F3's corrected reason, graded)
  collision guard never fires      DEAD, 7
  collision guard always fires     DEAD, 10
  drop the `prev != typ` arm       DEAD, 1   (same-name/SAME-type must bind)

Full ./internal/extractors/csharp/ suite -count=1 exit 0; go vet exit 0; gofmt
clean; tools/node-type-gate exit 0 with 0 new misses; ./cmd/grafel/ -count=1
exit 0 under a FRESH GRAFEL_HOME and GRAFEL_DAEMON_ROOT, 164s — the graph
digest did NOT move, which is expected: untyped-sibling collisions are a subset
of same-name collisions, measured at 0 in 3721 .cs files.

---

## Round 1

`collectLocalVarTypes` builds its name -> type map from a FLAT `findAllNodes`
walk over the whole method body, keyed by bare name. It has no model of block
scope, so two locals of the same name in sibling blocks both write the same
key and whichever declarator the walk reached last simply won:

    { Order o = GetOrder(); o.Process(); }
    { Customer o = GetCustomer(); o.Process(); }

The loser's call then carried a receiver type that name never had at that
site. Because the winner is a REAL same-file type, the dotted edge BINDS --
bind rate, orphan rate and dangle count all score it as a success. That is the
#7056 signature: the failure is invisible to every metric that would catch it.

REFUSE, DON'T GUESS. When a name maps to two DIFFERENT types the name is
dropped and its calls fall back to their bare leaf. The argument is
csNonBindableTypeKeyword's, verbatim and one case over: a dotted target leaves
the resolver's bare-name class and is scored as a confident bind (#7071), so a
WRONG dotted receiver is that same hazard rather than a lesser one -- a
fabricated `var.Process` at least dangles, while `Customer.Process` on a real
Customer does not. The `foreach` arm already reaches this answer by the same
route (feTypes/feAmbiguous), and csNamesBoundOutsideForeach exists only because
collision-refusal was already the convention there, implemented as a word list.
#7068's own note said what it needed -- "that needs scope, not a word list" --
and this is that refusal, with the honest caveat that it is still not scope.

Implemented as the same two-stage ledger the `foreach` arm uses: candidates
accumulate in localTypes/localAmbiguous and only unambiguous names publish to
`out`. It cannot be done in one stage against `out`, because the `foreach` arm
READS `out` and needs a refused name ABSENT from it, not present-then-deleted
partway through the walk. The `params win over locals` precedence at :629 is
untouched -- refusal removes a LOCAL entry, and a parameter's was never the
local map's to remove (pinned).

WHAT THE OTHER ANSWER WOULD COST, since it is not wrong: a block-scoped symbol
table means a recursive walk instead of findAllNodes, a frame pushed at every
scope-introducing node, and call sites resolved against the frame stack LIVE at
their position -- which means extractCallRelationships can no longer take a
prebuilt map and the params/locals merge becomes the stack's bottom frame. That
is a restructuring of two functions plus every binding form in
csNamesBoundOutsideForeach's enumeration. It is the right end state and nothing
here blocks it: a per-site table simply stops consulting the ledger.

SCOPE OF THE FIX. It reaches the local_declaration_statement path. The
`foreach` path already had its own ambiguity ledger and its own refusal against
names bound elsewhere (#7068), and is unchanged -- both paths now refuse, by
separate mechanisms over the same `out` map.

NESTED SHADOWING, which #7072 did not probe: treated identically to sibling
reuse, because a flat walk cannot tell them apart. An inner block redeclaring
an outer name at a different type is believed to be CS0136 and not a compilable
program -- DERIVED FROM THE SPEC AND UNDERIVED BY EXECUTION, as no C# compiler
exists in this environment (`csc`/`dotnet`/`mono`/`mcs` all absent, verified).
Recorded the way csNonBindableTypeKeyword records its own `dynamic` question,
rather than asserted. If the rule holds, refusing costs nothing real; if it
does not, the cost is an outer binding degraded to a bare leaf -- the honest
direction. What settles it: compile the fixture with `csc`.

Also corrected an over-claim in the #7068 comment, which stated that a
same-name/different-type collision "degrades to a bare leaf rather than to a
wrong dotted target". That was FALSE of the locals pass it appeared to
describe: only the `foreach` arm refused such a name. This change makes the
sentence true and says so where it stands.

GRADING. Corpus incidence is 0 in 3721 `.cs` files (measured), so a guard that
never fires and one that always fires are indistinguishable on the corpus --
fixtures are the only instrument, and both directions are graded. 9 mutants,
all DEAD, every occurrence of each mutated literal scored (`seen && prev !=
typ` appears twice -- the new locals ledger and the pre-existing `foreach` one)
and the compound predicate scored arm by arm:

  never-fires (locals)         DEAD, 4 tests
  always-fires (locals)        DEAD, 9 tests
  drop the `seen` arm          DEAD, 9 tests
  drop the `prev != typ` arm   DEAD, 1 test (same-name/SAME-type must bind)
  publish filter never-fires   DEAD, 4 tests
  publish filter always-fires  DEAD, 9 tests
  non-sticky flag              DEAD, 1 test (the stickiness the comment claims)
  foreach occurrence, never    DEAD, 1 test
  foreach occurrence, always   DEAD, 48 tests

Full ./internal/extractors/csharp/ suite -count=1 exit 0; go vet exit 0;
tools/node-type-gate exit 0 (0 new misses); ./cmd/grafel/ -count=1 exit 0 under
a fresh GRAFEL_HOME and GRAFEL_DAEMON_ROOT -- the graph digest did NOT move,
which is the expected result of a zero-incidence guard and is consistent with
the measured corpus count.

Closes #7072


cajasmota:

**My mutant DC-1 confirmed your riskiest decision rather than undoing it.** You *deleted* a guard of your own — the `|| typ == ""` arm that came back ALIVE — on the premise that `csharpCallTarget`'s `recv == ""` fallback is the load-bearing one. Deleting coverage on a false premise is silent, so I scored the premise: making that fallback never fire (`:724`, `go vet`-gated, single occurrence asserted) kills **15 tests**. The premise holds and the deletion was right.

Removing a redundant guard instead of adding a test that pins nothing is the rarer and better instinct, and applying this file's own rule to your own code — *"a redundant guard fires only where the real one already fires and leaves both ungraded"* — is exactly how that rule is supposed to be used.

**F1: poison rather than pin was the right call, for the reason you gave.** *"We could not type this declaration" is not "no declaration happened here"* — the name **is** bound at that site, and the type derived elsewhere provably does not describe it. Pinning would have left a *binding* wrong edge in place behind a test attesting to it, which is the single outcome this issue exists to eliminate.

**You found a second instance of the hole the review did not name** — an untypeable *declared* type (a `ref` local's `ref_type`, which `leafTypeName` has no case for) skipping the ledger one loop further out than the `var` path — and probed it before asserting it.

**And you corrected the review with evidence instead of deferring to it.** I forwarded F3 as a blocker, so that correction lands on me as much as on the reviewer: a one-stage form that deletes from `out` does **not** pass clean here — it dies on the sticky test. If the reviewer's version passed, it retained a refused-name set, which is the two-stage form under another name. The replacement reason is the true one and is now recorded: **the ambiguous set is the only thing that remembers a name once disagreed**, so deleting the entry destroys that memory and a third agreeing declarator resurrects the wrong binding. You demonstrated it as a mutant rather than arguing it.

F2 closed: the params/locals precedence now has a fixture where the names genuinely collide, so the line is exercised at all. The old test's name claimed more than its body did — parameter `o` against locals named `p` — which is fixture-label-over-content, and its own comment had already conceded it.

**Filed as #7094**, from your note: `internal/extractors/java/java.go:948` carries the identical unprobed `// params win over locals` line — and grounding it showed Java's `collectLocalVarTypes` at `:1296` is the **same flat last-writer-wins walk**, so the defect itself has an unfixed Java twin, not merely the comment. The issue records what is measured (the C# hole) separately from what is inferred (the Java one), and says to mutate before trusting the inference. It also warns against copying the two-stage ledger reflexively, since Java has no `foreach` arm of that kind.

Twelve mutants, each `go vet`-gated before scoring and counted by `--- FAIL` lines rather than exit code — the trap that cost a real finding on #7092 yesterday.

Issue #7091, opened 2026-09-16
java: 3 of 4 signature whitespace-collapse sites are ungraded — nothing in the package asserts a multi-line signature, so a wrapped class declaration would render with embedded newlines and no test would notice


cajasmota:
`internal/extractors/java/java.go` collapses a declaration's raw source span into a one-line `Signature` with `raw = strings.Join(strings.Fields(raw), " ")`. That call appears at **four sites**. Only one is graded.

| site | function | delete the collapse | status |
|---|---|---|---|
| `:1679` | `buildAnnotationElementSignature` | 1 failure | **graded** |
| `:520` | `walk`, record-component signature | `go vet` 0, suite **exit 0** | **ungraded** |
| `:1741` | `buildMethodSignature` | `go vet` 0, suite **exit 0** | **ungraded** |
| `:1757` | `buildClassSignature` | `go vet` 0, suite **exit 0** | **ungraded** |

`:1679` is graded only because #7089 just added the assertion. The other three are pre-existing.

## Why they are ungraded, and why that is one root cause rather than three

**Nothing in this package asserts a multi-line signature anywhere.** Collapsing whitespace on a single-line span is a no-op, so a fixture whose declarations all fit on one line cannot observe the call no matter how many assertions it makes about `Signature`. Every existing signature assertion is on a single-line declaration.

So this is not three independent oversights. It is one missing fixture shape — a declaration that wraps — replicated across every construct that renders a signature.

## Why it matters, in priority order

`buildClassSignature` is the one worth doing first. A wrapped

```java
public class OrderServiceImpl
        extends AbstractService
        implements OrderService, Auditable {
```

is ordinary Java and common in real code, whereas a wrapped annotation element is rare. `Signature` is persisted and is what a reader sees in `grafel_inspect` and `grafel_get_source` output, so an uncollapsed one is a visibly broken string with embedded newlines and source indentation, on exactly the declarations most likely to be inspected — the big classes at the top of a dependency graph.

`buildMethodSignature` has the same exposure for a method with a long parameter list or a `throws` clause on its own line.

## Provenance — measured, not declared

Found while closing #7073/#7089. My mutant DA-1 deleted the collapse at `:1679` alone and came back **ALIVE**; the author fixed it, and then — rather than simply reporting which of the other three they had checked — probed all three individually at ~0.7s each and reported the exit codes above. That turned a scope note into a measurement.

[...]

PR #7095, opened 2026-09-16
fix(#7094): Java's local-type map was flat and its enhanced-for arm overwrote it unconditionally, so a same-name local got a confidently WRONG receiver

### LINKED ISSUE #7094: java: collectLocalVarTypes is the same flat last-writer-wins map #7072 is fixing in C#, and 'params win over locals' is pinned by nothing in either extractor
#7072 is being fixed for C# in #7093. **Java has the same shape, unfixed**, and the precedence line both extractors rely on is pinned in neither.

## What I read at source

`internal/extractors/java/java.go:1296`, `collectLocalVarTypes`:

```go
out := map[string]string{}
for _, decl := range findAllNodes(body, "local_variable_declaration") {
```

A **flat walk over the whole method body** into one name-keyed map, with no scope tracking — structurally identical to the C# code #7093 is replacing. Two same-name locals in sibling blocks would collide, last writer wins, and the loser's calls would carry a receiver type that declaration never had.

`java.go:944-949`, the consumer:

```go
for k, v := range locals {
    merged[k] = v
}
for k, v := range paramTypes {
    merged[k] = v // params win over locals
}
```

Byte-identical in shape to the C# merge, including the comment.

## What is measured, and what is not — the distinction matters here

**Measured, on the C# side (#7093 review):** inverting that merge so locals overwrite params leaves the C# suite at **exit 0, zero FAIL lines**. The test that claims to pin it uses parameter `o` and locals named `p`, so the names never collide and the precedence line is never exercised — the test's own comment concedes as much. That is a demonstrated hole, not a suspicion.

**NOT measured, on the Java side.** I have not run a Java fixture with colliding sibling locals, and I have not mutated the Java merge. I read the code and grepped: no file under `internal/extractors/java/` other than `java.go` itself mentions params-precedence, which is suggestive but **is not proof** — a differently-named test could pin it incidentally. Whoever takes this should mutate first and find out, rather than inheriting my inference.

The reason I am filing on an unmeasured twin at all is that the C# original was also 0/3721 in corpus incidence and still worth fixing: the failure mode is an edge with a plausible name, a real file, an admissible kind, that **binds** — so bind rate, orphan rate and dangle count all score it as a success. Nothing distinguishes it from a correct edge except reading the source. A zero measurement on one corpus is not a bound on the language (#7056).

## A second, probably-shared hole

#7093's review found that C#'s fix is incomplete in a specific way: when `var` RHS inference yields no type, the declarator `continue`s **before** the ambiguity ledger, so an untyped sibling redeclaration does not poison the name and the original wrong-bind survives. Java's `collectLocalVarTypes` has the same `var`-with-inference structure (its comment at `:1303-1308` describes inferring from `new ClassName(...)`), so the same hole plausibly exists here. **Unprobed.** Check it in the same pass rather than discovering it a release later.

## Suggested shape

Mirror whatever #7093 lands, if the defect reproduces:

- refuse on a same-name/different-type collision rather than guessing, since a wrong dotted receiver is scored as a **confident** bind while a bare leaf is not — the argument C#'s own `var`/`dynamic` comment already makes;
- same-name/**same-type** must still bind;
- grade the collision check in **both** directions, because on a corpus with zero incidence a guard that never fires and one that always fires are indistinguishable;
- add a fixture where a parameter and a local genuinely **share a name**, so the precedence line is exercised at all. Mark the compilability question derived-not-executed unless you can run a compiler.

Do not copy C#'s two-stage ledger reflexively — #7093's review showed the stated reason for it (that the `foreach` arm reads the map mid-walk) is false, and a one-stage delete-on-collision passes clean there. Java has no `foreach` arm of that kind at all, so the simpler form is likely right here.

Refs #7072, #7093, #7056.

cajasmota:
Closes #7094.

## Measured before anything was changed

#7094 was opened from reading, so the first step was the colliding fixture run against the **unmodified** extractor. It reproduces:

```java
{ Order o = new Order();       o.a(); }   // -> Order.a   (right)
{ Customer o = new Customer(); o.b(); }   // -> Order.b   (WRONG, and it BINDS)
```

`collectLocalVarTypes` is a flat `findAllNodes` walk keyed by bare name with no model of block scope. Which declarator wins is a traversal artefact — `findAllNodes` is a stack DFS that pops siblings right-to-left, so the **earlier** declaration overwrites the later one (the opposite of C#'s last-writer-wins, same unsoundness). The winner is a real same-file type, so the wrong dotted edge binds and every metric scores it as a success.

## Where Java differs from C#, and why the policy differs

The issue's premise was that Java has no `foreach` arm of the kind C# has. **It has one, and it was the more dangerous half**: the `enhanced_for_statement` loop ran *after* the declaration walk and did a bare `out[name] = typ` with no ambiguity check at all. Measured:

```java
{ Order o = new Order(); o.a(); }
for (Customer o : cs)  { o.b(); }         // -> Customer.a AND Customer.b
```

So this is **not** C#'s two-mechanism shape. There is **one ledger fed by both arms**, and nothing reads the published map mid-walk. The two-stage form is kept for the reason #7072 ended up at: the ambiguous *set* is the only thing that remembers a name once disagreed, and Java has the third-declarator resurrection problem too (Order/Customer/Order all emitted `Order.*`). The one-stage delete-on-collision form was implemented as a mutant and the stickiness test kills it.

## The two extra holes

- **Hole 1 (untyped `var` sibling) REPRODUCES** — `{ Order o = ... } { var o = MyFactory.makeCustomer(); o.b(); }` emitted `Order.b` in both source orders. Untyped declarators now poison the name with an empty type.
- **Hole 2 (untypeable declared type) DOES NOT REPRODUCE** — recorded rather than papered over. `leafTypeName`'s switch covers every `_unannotated_type` the grammar produces, and a leading annotation is hoisted into the declaration's modifiers rather than producing an `annotated_type` type field. `@NonNull Customer o`, `final Customer o`, `Customer @NonNull [] o`, `List<@A Customer> o`, `@NonNull var o` were all measured reducing to a non-empty leaf. The empty-declType arm is defensive and **UNGRADED**; the comment says so. The fixture written for it was renamed to what it actually grades.

`params win over locals` was unexercised by any Java fixture, exactly as in C#. It is now pinned by a genuinely-colliding parameter/local and the inverted-merge mutant is DEAD.

## Recorded, not fixed here — needs its own issue

`leafTypeName`'s `scoped_type_identifier` case and `newExprClassName` both take `ids[len(ids)-1]` and both comment it as "the rightmost type_identifier". `findAllNodes` is a reverse DFS, so that index is the **leftmost**. Measured: `com.x.XController c = mk(); c.getCounts()` emits `com.getCounts`, and `var c = new com.x.XController(svc)` emits `com.com`. A fully-qualified type types its receiver to the package root. Left out to keep this diff disjoint from #7091 (signature builders, same file, untouched here).

## Round 2 — the JLS claim was false, and it was load-bearing

The comment said Java forbids an inner block from redeclaring a name in scope, so "a compilable program cannot contain that case". **False.** JLS §6.4 restricts redeclaration only within the *directly enclosing* method/constructor/initializer; a local or anonymous **class body** is a new class scope and may legally shadow. Calls inside such a body are attributed to the enclosing method entity, so the flat walk reaches across the class boundary:

```java
Order o = new Order();
o.a();                                   // was Order.a  ->  now bare "a"
Runnable r = new Runnable() {
  public void run() { Customer o = new Customer(); o.b(); }
};
```

Measured on compilable Java; the control with the inner variable renamed still emits `Order.a Customer.b`, so the loss is cross-class-boundary poisoning.

**Behaviour deliberately unchanged** — refusing across a class boundary is defensible and recall loss is the honest direction. What was wrong is the sentence asserting the case *cannot exist*, which is how a reader concludes there is nothing to grade. Recorded as a **fixture** rather than as prose: `TestJava7094_ClassBodyShadowingCostsRecall` asserts the cost in both the anonymous-class and named-local-class shapes plus the renamed control, so the cost moves when the behaviour does — a real symbol table stopping at the class boundary makes it fail, and it gets updated to assert the recovery.

Round 2's `java.go` diff is **comment-only**, verified mechanically (zero changed non-comment lines), so `./cmd/grafel/` was not re-run.

## Grading

11 new tests; **6 are RED** against the unmodified `java.go` (verified by byte-level `cp` restore, shasum-checked). 10 mutants, re-scored against the round-2 baseline rather than carried over, all **DEAD**, each `go vet`-gated *before* scoring and scored by counting `--- FAIL` lines rather than the exit code. Every mutated literal has exactly one occurrence in this package, verified by count before each patch. Both directions of the collision guard are graded, since the corpus incidence is zero.

| mutant | verdict |
|---|---|
| collision guard never fires | DEAD, 6 |
| collision guard always fires | DEAD, 7 |
| drop the `seen` arm | DEAD, 7 |
| drop the `prev != typ` arm | DEAD, 1 (same-name/same-type must bind) |
| publish filter never fires | DEAD, 6 |
| publish filter always fires | DEAD, 7 |
| untyped `var` skips the ledger | DEAD, 1 (hole 1) |
| foreach arm leaves the ledger | DEAD, 1 (the Java-specific half) |
| params merge INVERTED | DEAD, 1 (previously unexercised) |
| one-stage delete-on-collision | DEAD, 1 (stickiness) |

`./internal/extractors/java/` `-count=1` exit 0, 0 FAIL lines; `go vet` exit 0; gofmt clean; `tools/node-type-gate` exit 0, 0 new misses; `./cmd/grafel/` `-count=1` exit 0 in 184s under a fresh `GRAFEL_HOME` **and** `GRAFEL_DAEMON_ROOT` — graph digest unmoved, the expected zero-incidence result.


cajasmota:

**My mutant DD-1 — DEAD, and it probed a path nobody else took.** The reviewer's M6 made an empty type **cede** the ledger; mine made an empty type **agree with everything**, so the untyped declarator still enters the ledger and simply never conflicts. Same wrong outcome, different mechanism. `TestJava7094_UntypedVarSiblingPoisonsTheName` catches it — 1 FAIL, `go vet` clean, single occurrence asserted, restored by `cp` with the shasum re-verified.

That result is worth more than the kill: it shows the test grades the **outcome**, not the implementation route. A future refactor that reaches the same wrong answer by a different path still dies.

**Taking both options on F1 was the right call.** Striking the false JLS claim was necessary; adding `TestJava7094_ClassBodyShadowingCostsRecall` is what makes the cost **observed** instead of described. Your framing of it is exactly right — *"it fails when a real symbol table recovers `Order.a`, which is the point."* A known cost pinned as an observation is the honest form; prose asserting the case cannot arise is what we just removed.

And the underlying correction matters beyond this file: §6.4 restricts redeclaration only within the **directly enclosing** method/constructor/initializer, so a local or anonymous class body is a new class scope and may legally shadow — while the flat walk attributes calls inside it to the enclosing method entity, crossing the boundary.

**Re-scoring all ten mutants rather than carrying the counts over is the discipline that matters here.** A new test changes kill counts, and four of yours moved (5→6, 6→7, 6→7, 5→6). A carried-over count can hide a mutant whose only killer no longer exists.

**The `./cmd/grafel/` skip is mechanical, not asserted** — every changed non-context line in `java.go` begins with `//`, non-comment changed-line count **0**. That is checkable, which is the whole difference between a justification and a claim.

Corrected in the body: 11 tests, and the pre-fix RED count is **6** not 5, because the new fixture also fails against unmodified `java.go`.

**Filed from this work:**
- **#7097** — try-with-resources, catch-clause parameters and lambda parameters still cede the name, so the #7094 wrong-receiver bind survives in three more constructs. All three measured on compilable Java in both source orders; the standalone controls show they never type *themselves*, so they can only ever lose to a sibling. The reviewer's probes are ready-made fixtures. **Score each construct separately** — three additions that only fail together grade none of them.
- **#7096** now records the interaction you noted: this ledger compares the **leaf**, so `Customer` vs `com.x.Customer` reduces to `Customer` and `com` and the ledger refuses a pair that never disagreed. Harmless direction, but fixing #7096 will visibly change these collision fixtures, and whoever takes it should expect that rather than be surprised by it.

One more thing measured rather than assumed in this PR, worth keeping: Java's symptom is **inverted** from C#'s. `findAllNodes` pops right-to-left, so the *earlier* declaration wins here and the *later* one wins there. Same unsoundness, opposite fixture expectations — the kind of detail that makes a ported fixture silently wrong.

PR #7098, opened 2026-09-16
fix(#7096): three Java sites took the LEFTMOST type_identifier while calling it the rightmost, so a qualified type named its receiver after the package root

### LINKED ISSUE #7096: java: three sites take ids[len(ids)-1] calling it 'the rightmost type_identifier', but findAllNodes returns REVERSE source order — so a fully-qualified type names its receiver after the package root, and it binds
`internal/extractors/java/java.go` takes `ids[len(ids)-1]` from a `findAllNodes` result at **three** sites, each with a comment saying it is taking the **rightmost** `type_identifier`. It is taking the **leftmost**.

| site | context | comment |
|---|---|---|
| `:1044` | `newExprClassName` | *"the rightmost type_identifier"* |
| `:1369` | type-field leaf | — |
| `:1405` | `leafTypeName`, `scoped_type_identifier` | *"`com.foo.Bar` — leaf is the rightmost type_identifier"* |

## Why the index is inverted

`findAllNodes` (`:1481`) is a stack DFS that pushes children in index order and pops from the end:

```go
stack := []ts.Node{root}
for len(stack) > 0 {
    n := stack[len(stack)-1]
    stack = stack[:len(stack)-1]
    if set[n.Type()] { out = append(out, n) }
    for i := 0; i < int(n.ChildCount()); i++ { stack = append(stack, n.Child(i)) }
}
```

Children are pushed left-to-right and popped **right-to-left**, so `out` is in **reverse source order**. For `com.foo.Bar` the last element is `com`, not `Bar`.

## The measured consequence

Not inferred — produced by running the extractor:

- `com.x.XController c = mk(); c.getCounts();` → **`com.getCounts`**
- `var c = new com.x.XController(svc);` → **`com.com`** and **`com.getCounts`**

**A fully-qualified type names its receiver after the package root.** That contradicts the #4682 comment describing this path, and it is the #7056 signature again: `com.getCounts` is a plausible-looking dotted target that leaves the resolver's bare-name class and is scored as a **confident** bind, so bind rate, orphan rate and dangle count all read it as a success.

## What is NOT claimed

- **No corpus incidence figure.** Fully-qualified types in local declarations are ordinary Java but their frequency here is unmeasured. The `var c = new com.x.XController(...)` form is the more common idiom and is affected.
- **Not that the fix is a one-character change.** Taking `ids[0]` is correct *if* the reverse-order property holds for every input reaching these sites; that property is an artefact of `findAllNodes`, not a guarantee, and nothing currently pins it. A fix should either pin the ordering or stop depending on it — reading the node's own `name` field is likely better than index arithmetic over a flattened list.
- **Not audited beyond Java.** Other extractors have their own `findAllNodes`-shaped helpers; whether any makes the same index assumption is unchecked.

## Provenance, and a correction to the count

Found by the #7094 implementer while probing java's local-type map, and **measured** before reporting. They reported **two** sites; grounding it for this issue found **three** — `:1369` carries no comment and so was not caught by a comment grep. That is the occurrence-count lesson again: the third site was invisible to the search that found the first two.

Deliberately **not** fixed in #7095, whose subject is the collision map. Keeping the diffs disjoint means a red gate can be attributed.

## Suggested next step

Pin the ordering property first — a test asserting `findAllNodes` returns reverse source order — **or** stop relying on it. Then fix all three sites together and score them **individually**: three sites that only fail together grade none of them.

Grade in both directions: a fully-qualified type must resolve to its leaf, **and** an unqualified type must be unaffected. And correct the three comments, which are the reason this survived — each one asserts the opposite of what the line does, so a reader checking the code against its comment would conclude it was already right.

Refs #7094, #7095, #4682, #7056.

cajasmota:
Fixes #7096.

## The defect

Three sites in `internal/extractors/java/java.go` took `ids[len(ids)-1]` from a `findAllNodes` result and called it "the rightmost type_identifier". `findAllNodes` is a stack DFS that pushes children in index order and pops from the end, so its result is in **reverse source order** — `len-1` is the **leftmost** segment.

| site | function | what it did |
|---|---|---|
| 1 | `javaCallTarget`, `object_creation_expression` arm | `new com.x.XController(svc)` → CALLS `com.com` |
| 2 | `newExprClassName` | `var c = new com.x.XController(svc); c.getCounts()` → CALLS `com.getCounts` |
| 3 | `leafTypeName`, `scoped_type_identifier` arm | `com.x.XController c = mk(); c.getCounts()` → CALLS `com.getCounts` |

All three are genuinely distinct entry points (a constructor edge, a `var` local's type, a declared local's type); the earlier pass reported only two because site 1's comment names the leaf only indirectly.

A fully-qualified type named its receiver after the **package root**. That is the #7056 signature: a plausible dotted target that leaves the resolver's bare-name class and is scored as a **confident** bind, so bind rate, orphan rate and dangle count all read it as a success. Fixtures are the only instrument.

## The fix shape, and why this one

`ids[0]` is right only *if* the reverse-order property holds — and that property is an artefact of `findAllNodes`'s implementation that nothing pins. So the dependency is **removed, not inverted**:

- sites 1 and 2 reduce through `leafTypeName`, which descends the node's own children (generic → first named child, array → `element`, scoped → leaf);
- `leafTypeName`'s scoped arm reads the node's **last named child**. `scoped_type_identifier` declares **no fields** in tree-sitter-java 0.23.5 (`node-types.json`: `"fields": {}`), so `ChildByFieldName("name")` is not available; the grammar is `seq(qualifier, '.', repeat(annotation), type_identifier)`, which puts the trailing identifier last in every shape it can produce — including `com.x.@NonNull Bar` and `Outer<T>.Inner`.

No kind test and no recursion is written on that child: every child kind the grammar permits there is the identifier, so either would be an arm nothing can reach — and a mutant confirmed both were unobservable (see M6 below, dropped for that reason).

All three comments are corrected; each previously asserted the opposite of what its line did, which is why a reader checking code against comment concluded it was already right.

## Collateral (same arm, other callers)

`leafTypeName` also feeds field/param types, `javaSuperclassNames` and `javaSuperInterfaceNames`: `Outer.Inner` reduced to `Outer`, `java.util.List<T>` to `java`, `extends com.foo.Base` to `com`. All now reduce to the leaf.

## The #7095 interaction

#7095's ledger compares the **leaf**, so `Customer` and `com.x.Customer` — one type, two spellings — reduced to `Customer` and `com`, and the ledger refused both as a false collision. That false refusal is now gone; `TestJava7096_SameTypeTwoSpellingsIsNotACollision` is the direct regression pin and could not be written before this fix. The opposite direction — two **different** types, one spelled qualified — is pinned separately and still refuses both receivers.

**No existing #7094/#7095 fixture changed**: none of them spells a type qualified, so the whole `internal/extractors/java` suite is green without edits.

## Grading

9 fixtures, every assertion on the emitted edge (`Relationships[].ToID` for `CALLS`). On the parent commit, 7 of 9 are RED (the two always-fires/unqualified controls pass, as they must).

Mutants — `go vet` clean before scoring, `--- FAIL` lines counted, full package suite each time:

| mutant | change | verdict | killed by |
|---|---|---|---|
| M1 | site 1 back to `ids[len-1]` | DEAD | 1 test — `QualifiedConstructorEdgeBindsToLeaf` only |
| M2 | site 2 back to `ids[len-1]` | DEAD | 1 test — `VarLocalFromQualifiedNewBindsToLeaf` only |
| M3 | site 3 takes the FIRST named child | DEAD | 7 tests |
| M5 | site 3 back to `ids[len-1]` | DEAD | 7 tests |
| M7 | site 3 arm deleted | DEAD | 7 tests |
| M6 | site 3 reads the child recursively instead of as text | **ALIVE** → refinement dropped from the fix |

Sites 1 and 2 are each killed by exactly one distinct fixture, so they are graded individually rather than only in company. Site 3 is the shared reduction both delegate to, so its mutants take the whole set with them — that is the delegation, not a twinned fixture.

## Verification

- `go test -count=1 ./internal/extractors/java/` — exit 0, 0 `--- FAIL`
- `go run ./tools/node-type-gate` — exit 0 (`77 misses = 66 baseline + 11 alias-form + 0 new`)
- `go test -count=1 ./cmd/grafel/` with a fresh `GRAFEL_HOME` **and** `GRAFEL_DAEMON_ROOT` — exit 0, 0 `--- FAIL`
- `go vet ./internal/extractors/java/ ./cmd/grafel/` — exit 0

**The semantic digest did NOT move, and the corpus says why:** `semanticDigest6118` hashes relationships by resolved endpoint content, so a changed `CALLS` target would move it — but no Java fixture under `cmd/grafel` spells a type qualified, so **zero** members reach these sites. The constant is untouched (nothing was bumped).


cajasmota:
Round 2 — F1 and F2 addressed in `ae2d78473`. Nothing else touched.

**F1 — the five ungraded callers.** Confirmed exactly as reported: reverting `javaSuperclassNames` alone to `findAllNodes(...)[len-1]` is `go vet`-clean and passes the entire Java suite with **0 `--- FAIL`** lines while emitting `EXTENDS -> com`. Six fixtures added, **one surface per test on its own source** — six assertions in one test fail together, which grades none of them.

Reds on the parent commit (`4e70cb07f`): **13 of 15** RED; the two always-fires controls (unqualified types, and the real two-different-types collision) PASS, as they must.

Per-surface mutants — `go vet` 0 before scoring, `--- FAIL` lines counted, full package suite each time, backup re-taken this round and restored by `cp` with shasum + `git diff HEAD` verified:

| mutant | reverted caller | verdict | killed by |
|---|---|---|---|
| S4 | `javaSuperclassNames` | DEAD | 1 — `QualifiedSuperclassBindsToLeaf` only |
| S5 | `javaSuperInterfaceNames` | DEAD | 1 — `QualifiedInterfaceBindsToLeaf` only |
| S6 | `javaInjectFieldTypes` | DEAD | 1 — `QualifiedInjectedFieldBindsToLeaf` only |
| S7 | `collectFieldTypes` | DEAD | 1 — `QualifiedFieldTypeReceiverBindsToLeaf` only |
| S8 | `collectParamTypes` | DEAD | 1 — `QualifiedParamTypeReceiverBindsToLeaf` only |
| S9 | enhanced-for arm | DEAD | 1 — `QualifiedEnhancedForTypeBindsToLeaf` only |

Each surface is killed by its own fixture and by no other, so all eight callers are now graded individually. Round-1 mutants **re-scored** (not carried over) against the enlarged set: M1 and M2 still killed by exactly one fixture each, disjoint; M3/M5/M7 (the shared arm) now take all 13.

**F2 — reworded.** The sentence described the shipped implementation as a killed mutant. It now reads:

> even when the trailing segment carries a type annotation: the annotations sit BEFORE the identifier, so the identifier is still last. A reduction that took a FIXED index instead — `NamedChild(1)`, which is the identifier for `com.x.Bar` and the ANNOTATION the moment one appears — is killed here, and only here.

That claim is scored, not asserted: the `NamedChild(1)` mutant is **DEAD**, killed by `TestJava7096_AnnotatedQualifiedSegmentStillBindsToLeaf` and by no other test.

**Verification** — `./internal/extractors/java/` exit 0, 0 FAIL · `./cmd/grafel/` with fresh `GRAFEL_HOME` and `GRAFEL_DAEMON_ROOT` exit 0, 0 FAIL (200s) · `go run ./tools/node-type-gate` exit 0 (`77 = 66 baseline + 11 alias-form + 0 new`) · `go vet` 0 · `gofmt -l` clean. The digest still does not move.

Thanks for the three demonstrations — the 4433-node brute force, the equivalent-mutant proof for M6, and the `Outer.Inner` resolved-but-wrong analysis all land, and F1 was my own argument left half-applied.

cajasmota:

**My mutant DE-1 — DEAD, 12 FAIL lines.** I descended into the qualifier and took *its* last named child, so `com.x.XController` reduces to the **middle** segment `x` rather than the package root. That is distinct from both of the reviewer's qualifier mutants, which land on `com`. It dies, so the fixtures discriminate at **depth ≥ 3** — they grade "is the leaf", not merely "is not the first segment". `go vet` clean, exact match count asserted, restored by `cp` with the shasum re-verified.

**F1 is closed properly: each of the six collateral surfaces is killed by exactly one fixture and no other.** That is the same standard you set for sites 1/2/3, now applied to the five callers the first round left open. All eight `leafTypeName` callers are individually graded, and the round-1 mutants were **re-scored rather than carried** — which matters, because new fixtures move kill counts and a carried count can hide a mutant whose only killer no longer exists.

**F2 came back better than a correction.** The new sentence does not just remove the false claim, it explains *why* a fixed index breaks — `NamedChild(1)` is the identifier for `com.x.Bar` and becomes the **annotation** the moment one appears — and that claim is **scored**: the `NamedChild(1)` mutant is DEAD, killed by that test and only that test. A comment that earns its place by naming a mutant nobody can kill any other way.

**Three judgement calls from round 1 were confirmed by demonstration, not deference:**

- **The grammar claim is settled empirically.** My mid-flight note pointed out that `node-types.json` does not encode child *order*, so "last named child is the leaf" rested on a rule rather than evidence. Review brute-forced it instead: 19 seeds × every prefix truncation × every single-character deletion × stray tokens after every dot — **4433 `scoped_type_identifier` nodes, last named child was `type_identifier` in 100%**, including under error recovery, with `generic_type` and nested scoped nodes appearing only first, never last.
- **Deleting the recursive refinement was right, and M6 is provably equivalent** — zero disagreements across all 4433 nodes, because `leafTypeName` on a `type_identifier` returns exactly its trimmed text. A fixture for that arm is **unwritable**, not merely unwritten. That is the distinction that justifies deletion over pinning, and it is worth keeping as a rule.
- **The collateral breadth is an improvement, and `Outer.Inner` was the worst of the old cases** — it bound `i.ping()` to `Outer.ping`, a real same-file method, so the wrong edge *resolved* and scored as a confident success. Nothing failed when it changed because nothing ever asserted it.

**And the digest argument was upgraded from a grep to a demonstration** — `./cmd/grafel/` re-run under the most aggressive perturbation of these sites, digest still unmoved, with the two `.java` fixture files enumerated. "It did not move" is weak evidence on its own; "it did not move under a mutant that would have moved it" is not.

**On the scratchpad collision you flagged:** you were right to raise it, and the defence that held is worth naming — your script asserted its own worktree path, so the foreign invocation **failed loudly** instead of mutating your tree. That is the highest-value of the isolation mitigations. The omission was mine: roughly twelve briefs this week specified unique *worktree* paths and never once specified a private *scratchpad* subdirectory or unique script names. There is now a standing constraints block that every brief carries, with that rule first.

PR #7099, opened 2026-09-16
fix(#7097): three Java binders were missing from the local-type ledger, so each ceded its name to a sibling and emitted a wrong receiver that binds

### LINKED ISSUE #7097: java: try-with-resources, catch parameters and lambda parameters still cede the name to a sibling, so the #7094 wrong-receiver bind survives in three more constructs
#7094/#7095 made java's local-type ledger refuse on a same-name/different-type collision, and extended that to two name-binding forms: ordinary local declarations and `enhanced_for`. **Three more binder-shaped constructs still cede the name**, so the exact #7094 defect survives for each of them.

All three measured on the merged branch, all emitting a **wrong dotted receiver on a real same-file type** — the binding, metric-invisible failure #7094 exists to stop:

| construct | probe | emitted |
|---|---|---|
| try-with-resources | `{Order o … o.a();}` + `try (Customer o = new Customer()) { o.b(); }` | **`Order.b`** (both source orders) |
| catch-clause parameter | `{Order o … o.a();}` + `try { … } catch (MyEx o) { o.b(); }` | **`Order.b`** |
| lambda parameter | `{Order o … o.a();}` + `{ cs.forEach(o -> o.b()); }` | **`Order.b`** |

All three are **compilable Java** — the `Order o` is scoped to a sibling block, so there is no JLS §6.4 conflict.

## Why this is the same defect and not a new one

#7095's stated principle is: *"we could not type this declaration" is not "no declaration happened here."* A name bound at a site the extractor cannot type must **poison** the ledger, not be skipped — otherwise a sibling's type is handed to calls it never covered.

That principle is implemented for `var` (hole 1) and for `enhanced_for`. These three constructs bind a name and are not represented in the ledger at all, so they are the same hole in three more places.

Standalone controls confirm the shape precisely: `try (Customer o = …) { o.b(); }` **alone** emits the bare `b`, so these constructs never type *themselves*. They can only ever **lose to a sibling** — which is exactly the case that produces a confident wrong bind rather than an honest bare leaf.

## Why it matters despite low incidence

The failure has a plausible name, a real file, an admissible kind, and it **binds**. Bind rate, orphan rate and dangle count all score it as a success (#7056). A corpus-relative zero is not a bound on the language, and try-with-resources and lambda parameters are ordinary modern Java — `try (var conn = ...)`, `list.forEach(x -> ...)` — with `e`, `x`, `item`, `conn` being exactly the short names most likely to be reused across sibling blocks.

## What is NOT claimed

- **No corpus incidence measurement.** Not run. The #7072/#7094 family measured 0 in 3721 `.cs` files and was still worth fixing for the reason above; I am not asserting a number here.
- **Not a regression.** All three are pre-existing and were outside #7095's diff, which was deliberately kept narrow so a red gate stays attributable.
- **Not necessarily one fix.** They may share a `record(name, "")` call, or a catch parameter may want different treatment from a resource — decide per construct rather than sweeping.
- **Not audited in C#.** #7072's fix covers `var` and `foreach` there; whether C# has analogous unledgered binders (`using` declarations, `catch (E e)`, lambda parameters) is **unchecked** and worth the same probe.

## Suggested next step

The three probes above are ready-made fixtures. For each construct, add it to the ledger with an empty type and grade it **in both directions**:

- the colliding case must refuse (assert the **emitted edge**, not the map contents);
- the construct **alone** must still behave as it does today — `TestJava7094_EnhancedForAloneStillBinds` is the model for that control.

**Score each construct separately.** Three additions that only fail together grade none of them, which is the twinned-site pattern that has produced six findings in this milestone.

Found during the independent review of #7095, measured rather than argued.

Refs #7094, #7095, #7072, #7056.

cajasmota:
Closes #7097.

`collectLocalVarTypes` refuses on a same-name/different-type collision, but its ledger was fed by only two binder forms — `local_variable_declaration` and `enhanced_for_statement`. Three more constructs bind a name and were absent from it entirely, so each ceded its name to a sibling and emitted a **wrong dotted receiver on a real same-file type** — which binds, and which every instrument scores as a success (#7056).

## Measured before any change

| probe | emitted |
|---|---|
| `{Order o … o.a();}` + `try (Customer o = new Customer()) { o.b(); }` | `Order.b` (both source orders) |
| `{Order o … o.a();}` + `try { … } catch (MyEx o) { o.b(); }` | `Order.b` |
| `{Order o … o.a();}` + `cs.forEach(o -> o.b())` | `Order.b` |
| each construct **alone** | bare `b` — none types itself |

That last row is the shape of the defect: these constructs never type *themselves*, so they could only ever **lose to a sibling**, which is the case that produces a confident wrong bind rather than an honest bare leaf.

## Per-construct decision, not a sweep

All three record the name with an **empty type**, for three different reasons:

- **lambda parameter** — genuinely untypeable in the dominant form (`o -> …`); the grammar carries no type. `""` is the only honest entry.
- **catch parameter** — a declared type exists, but `catch_type` is a **union** (`catch (A | B e)`) with no single leaf to prefer. Picking one is exactly the guess #7094 refused to make, and `leafTypeName` has no `catch_type` case.
- **try-with-resources** — the one arm where a single declared leaf **is** derivable, so typing it would not be a guess. Still recorded as `""`, because typing it is a **recall addition with its own grading obligation**, not part of removing a wrong bind.

That last cost is **observed, not described**: `TestJava7097_TryWithResourcesSameTypeCost` pins the recall a *same-type* sibling loses to the poison — today `Order.a`/`Order.c` both bound by accident; now both fall back to bare. That fixture is what must move if someone later types the resource.

## Grammar spellings derived, not assumed

From a real parse tree: `try_with_resources_statement → resource_specification → resource{type,name,value}`; `catch_clause → catch_formal_parameter` with the type in an **unnamed `catch_type` child**, not a `type` field; `lambda_expression`'s `parameters` field across all three shapes (`identifier` / `inferred_parameters` / `formal_parameters`). A `try (existing)` resource has **no `name` field**, so it correctly binds nothing.

## Grading

7 `--- FAIL` against pre-change `java.go`; the 6 controls pass on both sides.

**11 mutants, all DEAD**, per construct and in both directions — arm-deleted and always-fires for each of the three, plus the three lambda parameter shapes scored individually and a `twr-type-instead-of-poison` variant.

**Two always-fires mutants were ALIVE in round 1 and that was a fixture defect, not a code one.** The negative controls called the outer `o.a()` *outside* the nested binder, where an over-broad walk cannot reach it — so they graded nothing. Capturing the outer receiver *inside* the catch body and the lambda body is what makes the over-refusal direction observable. Fixtures rewritten and **fully re-scored**, not carried.

## Verification

`./internal/extractors/java/` exit 0 · `./internal/extractors/ ./internal/resolve/ ./internal/links/` exit 0, 0 FAIL · `./cmd/grafel/` with fresh `GRAFEL_HOME` **and** `GRAFEL_DAEMON_ROOT` exit 0, digest **unmoved** (consistent with corpus incidence 0 for this family) · `go run ./tools/node-type-gate` exit 0, 0 new misses · `gofmt -l` and `go vet` clean.

## Not claimed

- **Corpus impact.** Incidence is 0, so the fixtures are the only instrument. No evidence this helps or costs anything in the wild beyond an unmoved digest.
- **The same-type recall cost beyond try-with-resources.** Graded for the resource arm only; the equivalent loss exists for catch and lambda but those shapes were judged contrived. A judgement call, not a measurement.
- **The C# twin**, which is unmeasured — see #7097.

## Found and deliberately NOT swept in

Two more constructs in the same family, both measured: `if (x instanceof Customer o) { o.b(); }` → **`Order.b`**, and `switch (x) { case Customer o -> o.b(); }` → **`Order.b`**. Filed separately with the grammar details rather than widened into this diff.


cajasmota:
**Round 2 pushed (`ddecc6f6b`).** Independent review returned REQUEST-CHANGES on one narrow item; the code itself was confirmed sound and is unchanged.

## What the review established, independently

Every grammar spelling was re-derived from the reviewer's own parse dumps rather than taken on the author's word, and all held — including the load-bearing one, that `try (o)` and `try (this.res)` carry **no `name` field**. Two things the author had not established: `collectLocalVarTypes` has exactly **one** caller, so there is no unfixtured consumer of the kind #7098 found five of; and `modifiers` (`final`) does not displace the `name` field in either construct.

12 mutants scored independently — 11 DEAD, and the single ALIVE is a **demonstrated equivalent**, not an ungraded arm: `catch_formal_parameter`'s only `identifier` descendant *is* its `name` field (the type sits under `catch_type`), so "poison every identifier" poisons the same set, shown with a distinguishing input rather than argued.

## The blocking item — a disclosed judgement call, then measured

The author's round-1 report said the same-type recall cost for catch and lambda was *"a judgement call, not a measurement"* and judged those shapes contrived. Measured, they are not:

| shape | pre-change | post-change |
|---|---|---|
| `{MyEx o=…; o.a();}` + `catch (MyEx o) { o.c(); }` | `MyEx.a`, `MyEx.c` | bare `a`, `c` |
| `{Customer o=…; o.a();}` + `cs.forEach(o -> o.c())` | `Customer.a`, `Customer.c` | bare `a`, `c` |
| `{Customer o=…; o.a();}` + `cs.forEach((Customer o) -> o.c())` | `Customer.a`, `Customer.c` | bare `a`, `c` |

That is exactly what `TestJava7097_TryWithResourcesSameTypeCost` exists to observe, in two more places with nothing observing it. The typed-lambda row is the least defensible — the type is literally in the source.

**The typed lambda got its own third fixture, and the reason is evidence rather than taxonomy:** `formal_parameters` is a different grammar node on a different switch arm, and — the point for a *cost* fixture — it has a different **move condition**, since its type is present in the source while the inferred row cannot move without functional-interface inference. `TestJava7097_LambdaTypedParameterSameTypeCost` is the **sole** killer of the new `lambda-formal-type-instead-of-poison` mutant.

**14 mutants, re-scored from scratch rather than carried** — which matters, because new fixtures move kill counts and a carried count can hide a mutant whose only killer no longer exists. Four counts did move (`catch-arm-deleted` 2→3, `lambda-arm-deleted` 3→5, `lambda-identifier-case-deleted` 1→2, `lambda-formal-case-deleted` 1→2). Three are new, including `empty-record-never-poisons` (11 kills), which grades the poison mechanism itself rather than any one arm.

## A false universal in the diff's own comment, corrected

The comment concluded *"All three bind, so all three poison."* That "all" is false and nothing observed it: `formal_parameters` can hold a **`spread_parameter`**, and the arm's `p.Type() == "formal_parameter"` guard skips it, so a varargs lambda parameter still cedes its name — `use((Customer... o) -> o.clone2())` beside a sibling `Order o` emits **`Order.clone2`**, before and after this PR. The comment now states what the arm actually does and cites #7102. **The hole itself is deliberately untouched here** so a red gate stays attributable.

## My own mutant, which nobody else ran — DG-1, DEAD (1 kill)

All 26 mutants scored by the author and the reviewer live inside `collectLocalVarTypes` or its node matchers. None touched the **call site** where the ledger meets `paramTypes`:

```go
for k, v := range locals     { merged[k] = v }
for k, v := range paramTypes { merged[k] = v } // params win over locals
```

I flipped the precedence so locals win — the permissive direction, in which a poisoned nested binder overrides a real parameter type. **DEAD, killed by `TestJava7094_ParamsWinOverALocalOfTheSameName` and only that test.** So the merge order is pinned.

Worth recording what the kill does *not* cover. That comment justifies itself with *"this only matters for nested blocks; conservative bias, no harm"* — and #7097 has just added three **nested-block** binder forms feeding `locals`, so the precedence is now load-bearing in a way it was not when the sentence was written. The one fixture that pins it uses a nested-block `local_variable_declaration`, i.e. the pre-existing form; the three new forms are not graded against this merge. Whether they *can* reach it depends on whether a resource, catch or lambda binder may legally collide with a parameter name, **which I am not asserting** — there is no Java compiler on this machine, and a JLS claim made from reading rather than compiling is how #7094's brief went wrong earlier in this milestone. Flagging it as an unmeasured cell, not as a defect.

## Verification (round 2, all `-count=1`)

`./internal/extractors/java/` 0 FAIL · `./internal/extractors/… ./internal/resolve/ ./internal/links/` 0 FAIL · `./cmd/grafel/` with fresh `GRAFEL_HOME`+`GRAFEL_DAEMON_ROOT`, **digest unmoved** · `node-type-gate` 0 new misses · `gofmt -l` and `go vet` clean.

## Disclosed, not hidden

`TestJava7097_LambdaInferredParameterSameTypeCost` is **never a sole killer** — every mutant it kills is also killed by the inferred collision fixture. It is kept as a direction observer rather than a unique grader, because the edit that would move it alone is not expressible as a local mutation. Fixture sources remain **derived-not-executed** as valid Java; there is no compiler here.

cajasmota:
## Correction to #7099's merged PR body, and to my own review note: there IS a Java compiler on this machine

#7099 shipped with the caveat *"fixture sources are **derived-not-executed** as valid Java; there is no compiler here."* I repeated a version of it when declining to assert a JLS claim on my DG-1 mutant note. **Both were false, and nobody had checked.**

```
$ which javac
/Users/…/.sdkman/candidates/java/current/bin/javac
$ javac -version
javac 25.0.3
```

`javac` 25 is installed via sdkman. (For contrast, and so this is not over-generalised: there genuinely is **no C# compiler** here — `csc`, `dotnet`, `mono`, `mcs` all verified absent. The two are not the same situation.)

## The open cell from my DG-1 note is now settled — by compilation

On #7099 I flagged an unmeasured cell rather than a defect. The call site merging the ledger with `paramTypes` justifies itself with *"Java forbids name-shadowing of parameters in the top-level method scope, so this only matters for nested blocks; conservative bias, no harm"*, and #7099 had just added three **nested-block** binder forms feeding that merge. Exactly one fixture pins the precedence, and it uses a nested-block `local_variable_declaration` — the pre-existing form. I wrote that whether the three *new* forms could reach that merge depended on whether a resource, catch or lambda binder may legally collide with a **parameter** name, and that I was **not asserting** it, for want of a compiler.

Three three-line files answer it. All three are rejected:

| probe | javac 25 |
|---|---|
| `void run(Order o) { o.a(); try (Cust o = new Cust()) { o.b(); } }` | `error: variable o is already defined in method run(Order)` |
| `void run(Order2 o) { … catch (MyEx2 o) { o.b(); } }` | `error: variable o is already defined in method run(Order2)` |
| `void run(Order3 o, List<Cust3> cs) { … cs.forEach(o -> o.b()); }` | `error: variable o is already defined in method run(Order3,List<Cust3>)` |

**So the cell is unreachable by construction and there is no defect.** None of #7099's three new binder forms can legally collide with a parameter, so the single fixture pinning the params-win precedence covers the only form that can reach it. The comment's "conservative bias, no harm" is correct as written, and now measured rather than reasoned.

That took under a minute, which is the point: the caveat cost more than the check would have.

## What this changes going forward

JLS legality and shadowing claims about Java fixtures are **testable here**, and this board's dominant defect class is prose asserting what nothing observes. Every Java extractor brief should carry *"compile every fixture snippet with `javac` and report zero errors"* — a fixture that does not compile is a fixture whose label may claim a shape Java cannot express.

Worth noting that the agent on #7100/#7102 did exactly this unprompted: **all 33 of its fixture snippets compiled clean with javac 25**, and it used the compiler to settle a mutant equivalence claim (legacy array notation on a variable-arity parameter is a compile error) rather than arguing it. That is the standard now.

Refs #7099, #7100, #7102.

Issue #7062, opened 2026-09-15
TestRebuild_SubprocessReroute_TimeoutCancelsChild gates a 100ms behaviour behind two 5s wall-clock deadlines, and its failure message blames a leak when what blew was the test's own budget


cajasmota:
`TestRebuild_SubprocessReroute_TimeoutCancelsChild` (`cmd/grafel/daemon_rebuild_subprocess_test.go`) gates a **100 ms** behaviour behind two **5 s wall-clock** deadlines. Under machine load it fails — and **the message it fails with is a false diagnosis**.

## The code

```go
// The rebuild must return promptly (bounded by the 100ms per-repo timeout),
// never blocking on the wedged child.
select {
case err := <-rebuildDone:
	…
case <-time.After(5 * time.Second):                                    // :223
	t.Fatal("rebuild did not return — the wedged child was not cancelled on timeout (leak)")
}

select {
case <-childReturned:
case <-time.After(5 * time.Second):                                    // :231
	t.Fatal("child goroutine never returned — ctx was not cancelled (process/claim leak)")
}
```

## What actually happened

It went red on an unrelated PR (#7057, a change confined to `internal/extractors/cpp`) while two other agents were compiling on this machine. Observed `took=5.88s`.

**The 100 ms per-repo timeout fired exactly as designed.** What missed was the test's own 5 s wall-clock budget — the goroutine simply did not get scheduled in time. The test then reported:

> `rebuild did not return — the wedged child was not cancelled on timeout (leak)`

That is the opposite of what occurred. A reader chasing this red goes looking for a cancellation leak in `daemonRebuildFuncCore`. There is none.

## Why it is worth an issue rather than a bump

**A 50× margin looks generous and is not a bound on anything.** The test asserts a *promptness* property — "returns within 100 ms, does not block on a wedged child" — but measures it with an absolute wall-clock deadline, so what it actually gates is *"this machine scheduled a goroutine within 5 s"*. On an unloaded box those coincide; under load they come apart, and when they do the test blames the code.

This is structurally the same defect the codebase keeps finding in its own assertions: **the thing asserted is not the thing measured.** #6700 is the same family on Windows (a 150 ms sleep raced against a 160 ms budget). Here the numbers are more comfortable and the failure mode is therefore rarer and more confusing.

[...]

PR #7103, opened 2026-09-16
fix(#7087): a spawn that never produced a child shared the crash budget whose recovery reset it can never reach — give it its own budget and a reset that fires

### LINKED ISSUE #7087: a spawn that never produced a child shares the crash-loop budget with a child that ran and crashed — and the recovery reset is unreachable for it, so a deterministic Start failure burns ~91s for nothing
`internal/daemon/supervise.go:368-374` treats a spawn that **never produced a child** identically to a child that **started and then crashed**:

```go
if err := cmd.Start(); err != nil {
    s.logger.Error("engine supervisor: spawn failed", "err", err)
    // Treat a failed spawn like a crash: back off and retry.
    if s.backoffAndMaybeGiveUp(ctx, &backoff, &ceilingHits) {
        return
    }
    continue
}
```

These are different failures with different recovery characteristics, and merging them makes the retry budget do the wrong thing for one of them.

## Why they are not the same

A **crashed child** is plausibly transient: a race, a bad input, transient memory pressure. Retrying with backoff is exactly right, and the loop knows it — `supervise.go:396-401` resets both `backoff` and `ceilingHits` once a child stays up past `healthyUptime`, on the reasoning that *"a child that stayed up long enough counts as recovered."*

A **construction failure** — `Start` returning an error before any process exists — is usually **deterministic**: a bad executable path, a missing binary, an unusable file handle. Nothing about waiting changes it. And critically, **the recovery reset can never fire for it**, because the reset is gated on a child that actually ran. So a deterministic construction failure walks the entire budget — 500ms doubling to a 30s ceiling, then 3 ceiling hits — and gives up, having spent roughly **~91 seconds** re-attempting something that failed identically every time.

The user-visible result is a daemon that appears healthy, then dies a minute and a half later, with the real cause named once in a log line at the very start of the sequence.

## Concrete instance

This is the mechanism behind #7083: the engine child inherited the daemon's standard handles, and on Windows an invalid inherited handle makes `StartProcess` fail at `DuplicateHandle` — deterministically, on every attempt. #7085 fixes that particular construction failure, but **not** the classification: the next deterministic `Start` error will burn the same budget the same way.

## What is NOT claimed

- **Not that spawn failures should not be retried at all.** Some are genuinely transient — EAGAIN under fork pressure, a briefly-locked binary during an upgrade. The claim is that they should not share a budget whose recovery rule is unreachable for them.
- **Not a live defect in #7085's fix.** That change is correct and this is orthogonal; it was noticed while implementing it and deliberately left unacted rather than scope-crept into an unrelated area.
- **No proposed threshold.** Whether the right answer is a separate (smaller) budget for construction failures, an immediate give-up on a repeated identical error, or simply a distinct log line and metric is a design question I am not settling here.
- **Not measured.** How often a construction failure occurs in the field is unknown; there is no counter distinguishing the two paths, which is itself part of the problem — a spawn failure and a crash are indistinguishable in every artefact the supervisor produces except one log message.

## Suggested shape, and how to grade it

Distinguish the two at the classification site — at minimum a separate counter and log verb, ideally a separate budget. **Grade it in both directions**, and note that the second is the one that will be skipped:

1. A **repeated identical** construction failure must stop sooner than a crash-loop does (the permissive direction is the status quo, so a test that only checks "it eventually gives up" passes today and grades nothing).
2. A **transient** construction failure followed by a successful spawn must still recover — otherwise "fail faster" quietly becomes "fail on the first hiccup", and a supervisor that gives up on a recoverable condition is a worse bug than the one being fixed.

Assert the **emitted log/metric** distinguishing the two paths, not the internal counter the supervisor keeps about itself.

Observed during #7083/#7085; the classification comment at `supervise.go:370` states the merge explicitly, so it is deliberate rather than accidental — which makes this a design question to reopen, not an oversight to correct.

Refs #7083, #7085, #7050.

cajasmota:
Closes #7087.

`cmd.Start()` returning an error means **no process ever existed**. The supervisor treated that identically to a child that ran and crashed, and the comment at the site said so deliberately: *"Treat a failed spawn like a crash: back off and retry."*

The asymmetry that makes the merge wrong is the **recovery reset**. `backoff` and `ceilingHits` reset only once a child stays up past `healthyUptime` — which requires a child to have existed. **For a construction failure that reset is structurally unreachable**, so a deterministic `Start` error walks the entire budget (500ms doubling to a 30s ceiling, then 3 ceiling hits, **~91s**) re-attempting something that fails identically every time. What the user sees is a daemon that looks healthy and dies ninety seconds later, with the real cause named once, in the first log line.

This is the mechanism behind #7083: on Windows an invalid inherited handle made `StartProcess` fail at `DuplicateHandle`, deterministically, on every attempt. #7085 fixed that particular construction failure and left the classification alone, so the next one would burn the same budget the same way.

## The fix — a reachable reset, not just a smaller number

The construction path gets its own budget **and its own reset that can actually fire**: a **successful `Start` clears the counter**. Consecutive construction failures cap at `maxSpawnFailures` (default 3): **~1.5s instead of ~91s**. That is what keeps the second direction honest — a transient `Start` failure (EAGAIN under fork pressure, a briefly-locked binary mid-upgrade) still recovers, because one successful spawn is enough to clear the count. "Fail faster" becoming "fail on the first hiccup" would be a worse defect than the one being fixed.

`backoffAndMaybeGiveUp` is split into `waitBackoff` + `giveUp`: **both paths share the wait, neither shares the verdict.**

**Give-up is on N consecutive failures regardless of error identity, not on an identical repeat** — and that is a deliberate departure from the issue's suggestion. An identity gate lets a stream of *differing* construction errors retry forever, because that path never touches the crash budget at all. The identity check reads as the tighter rule and is the weaker one.

## Graded on the emitted artefacts, not the counter the supervisor keeps about itself

- **Fatal error text**, returned verbatim to the user by `RunServe`: `engine child unspawnable: N consecutive spawn failures, no child process was ever created: …` versus the existing `engine child crash-looping: …`. Asserted in **both** directions for **both** arms.
- **Log stream**: `engine child spawn failed — no child process was created` present for construction failures and **absent** for crashes (where `engine child exited` appears instead).
- **Spawn attempts counted by the injected command factory** — an external observation, not an internal field: the construction arm stops strictly sooner than a crash arm under identical tuning.

No wall-clock deadline gates any behavioural assertion. #7062 in this same package has a 100ms behaviour sitting behind two 5s deadlines with a failure message that blames a leak when what blew was the test's own budget; the clock is injected here instead.

## Grading

**9 mutants, 8 DEAD**, each applied with an asserted match count of 1, `go vet` before scoring, `--- FAIL` lines counted, `-count=1`, restored by `cp` with shasum **and** `git diff HEAD` verified:

status-quo restore of the spawn branch (kills both directions) · no-reset-on-successful-`Start` (kills direction 2 specifically) · `>=`→`>` on the budget · drop the distinct log verb · unspawnable-fatal-text→crash-looping · init counter nonzero · reset-to-1 · and the **pair audit on the twinned verdict** — mutating the *crash* path's fatal text to "unspawnable" is also DEAD. Twinned keys have produced eight findings this milestone; scoring one twin says nothing about the other, so both were scored.

**1 ALIVE, recorded as equivalent under the current suite with a demonstration rather than as a kill:** adding `spawnFailures = 0` to the healthy-uptime crash block is a provable no-op — that block is reachable only after the successful-`Start` reset has already run, and the sole increment site `continue`s. A fixture for it is unwritable, not merely unwritten.

## Verification

`go test -count=1 -timeout 15m ./internal/daemon/...` exit 0, **0 `--- FAIL`** · `-race -run TestEngineSupervisor_` exit 0 · `gofmt -l` and `go vet` clean. No extractor or assembly path is reached (daemon supervision only), so `./cmd/grafel/` was not run.

## Not claimed

- **Windows and Linux behaviour.** The new test file is deliberately untagged and cross-platform — #7083 was Windows-only and this classification is not — but it has only *run* on macOS here. The ubuntu and windows legs are the real cross-platform signal. **Needs `ci:full` for a macOS leg.**
- **The production tuning (~91s vs ~1.5s) is derived from the constants, not measured.**
- **How often this occurs in the field**, which is still unknown — and was unknowable before this change, since a spawn failure and a crash were indistinguishable in every artefact the supervisor produced except one log message. That is now fixed, so the question becomes answerable.


cajasmota:
**Independent review: APPROVE.** Plus one correction to this PR body and one finding from my own mutants.

## What the review re-derived rather than took on trust

- **The arithmetic in this body is correct.** From the constants (500ms initial, ×2, 30s ceiling, 3 ceiling hits): 0.5+1+2+4+8+16+30+30 = **91.5s**, and the spawn budget's two waits are 0.5+1 = **1.5s**.
- **The ALIVE-equivalence claim is true.** `spawnFailures` is provably 0 at every point downstream of a successful `Start` — the sole increment site `continue`s or `return`s, and the crash-exit block is reachable only past the reset. The reviewer then added the *stronger* mutant in the same class (crash-exit clears `spawnFailures`) and it is equally equivalent.
- **Budget-interaction audit:** the spawn path calls `waitBackoff` directly and never `backoffAndMaybeGiveUp`, so a construction failure can never touch `ceilingHits`. The reverse is vacuous by the reachability argument. No wrong interaction.
- **Alternating failure/success terminates** — the successful spawns' children still crash, and the crash budget only resets on `healthyUptime`, so it still fires.
- **The design departure is sound as reasoned.** With an identity gate, a stream of *differing* construction errors would loop forever, because the spawn path touches the crash budget not at all and nothing else bounds it.

Six reviewer mutants: 4 DEAD (delete the post-`Start` reset — killed by the transient-recovery test, so direction 2 is genuinely graded; `>=`→`>`; successful `Start` also resetting the *crash* budget; resetting `spawnFailures` every iteration), 2 ALIVE and demonstrated equivalent.

## Correction to this PR body — the ~1.5s figure is conditional

`backoff` is **shared** between the two paths and resets only on healthy uptime. So "~1.5s instead of ~91s" holds **only from a fresh backoff**. If a crash loop has already driven `backoff` to the 30s ceiling without giving up, a following construction-failure sequence takes **~60s**, not 1.5s, to reach the unspawnable verdict.

Still bounded, still strictly better than 91.5s, and *"both paths share the wait, neither shares the verdict"* is the stated design rather than an oversight — but the unqualified 1.5s would become a wrong number in a changelog. **Read the figure as "~1.5s from a fresh backoff, ~60s worst case after a crash loop has already raised it."**

Related sharp edge, recorded not actioned: `maxSpawnFailures = 3` over a ~1.5s window could clip a genuine fork-pressure storm (3 EAGAINs inside 1.5s). Mitigated by give-up meaning an OS-unit recycle rather than data loss, and it is the sharpest edge of choosing no identity gate.

## My own mutants, which nobody else ran — DH-1 and DH-2, both ALIVE

All 15 mutants scored by the author and reviewer target the **mechanism** in the run loop. None touched the **production constants**:

| mutant | verdict |
|---|---|
| `defaultEngineMaxSpawnFailures` 3 → **1000** | **ALIVE** — `./internal/daemon/` exit 0, **0 `--- FAIL`** |
| `defaultEngineMaxCeilingHits` 3 → **1000** | **ALIVE** — same |

The supervisor's tests inject tuning, which is exactly what frees a shipped default from grading: the suite pins *"the budget mechanism works at whatever value you hand it"* and says nothing about the value users run. **Set `maxSpawnFailures` to 1000 and this PR's entire purpose is reverted with a green suite.**

**Not blocking, because I scored the twin.** `defaultEngineMaxCeilingHits` predates this PR and is equally free, so this is a pre-existing pattern that this change followed faithfully — and fixing only the new constant would leave the older, larger budget (~91s) just as unpinned. Filed as **#7106** covering both, with the kill priced rather than assumed: a no-tuning run asserting the verdict arrives within a bounded attempt count costs ~1.5s of wall clock.

## Disjoint failure, diagnosed not rerun

Full `./internal/daemon/...` showed **1 unrelated failure**: `TestAFileReplacingARemovedDirectoryIsStillReleasable` in `internal/daemon/watch` (`fdbudget_6293_test.go:212`), an fd-ledger settle flake. Structurally disjoint — `go list -deps ./internal/daemon/watch` does **not** contain `internal/daemon`, so no code this PR touches is reachable from it. `./internal/daemon/` itself is green, and `-race -run TestEngineSupervisor_` exits 0.

## Standing checks passed

Every new assertion targets the fatal error text, the log stream, or the injected factory's attempt count — no internal counter is asserted. No behavioural assertion is gated on a short wall clock; the 60s deadlines are outer bounds and their messages name what actually failed (*"never surfaced a fatal (spawn attempts so far: N)"*), which is not the #7062 shape. Fixture names match their bodies. No other test depended on the changed log text.

PR #7104, opened 2026-09-16
fix(#7090): Entity.Subtype decided graph identity while being invisible over MCP — emit it, and for 51% of entities it is the only route

### LINKED ISSUE #7090: Entity.Subtype is hashed by the digest and keyed on by engine passes, but serializeEntity never emits it — so whether an agent can see a subtype depends on which language produced the entity
`Entity.Subtype` is a first-class field. It is hashed by the graph digest (`cmd/grafel/custom_extractor_spans_6118_test.go:411` — `Kind|Name|QualifiedName|Subtype|SourceFile|Language|StartLine|EndLine`), it is what extractors use to distinguish a `class` from an `interface` from a `record` from an `annotation`, and several resolver and engine passes key on it.

**It never reaches an MCP consumer.** `internal/mcp/tools.go:2198-2225`, `serializeEntity` — the one function every entity payload goes through:

```go
out := map[string]any{
    "id":             id,
    "name":           e.Name,
    "qualified_name": e.QualifiedName,
    "kind":           stripScopePrefix(e.Kind),
    "file":           e.SourceFile,
    "line":           e.StartLine,
}
if wantVerbose {
    out["end_line"] = e.EndLine
    out["language"] = e.Language
    …
    if e.PropLen() > 0 { out["properties"] = e.PropsSnapshot() }
}
```

No `subtype` key, in either branch. `grafel_find`'s hit rows do not emit one either, and no MCP tool filters on it — `kind_filter` is bare-Kind only.

## The part that makes it a correctness problem rather than a missing feature

A subtype *can* reach a consumer, but only by an accident of two independent conditions:

1. the extractor **also** dual-stamped it into `Properties["subtype"]` — some do (`golang`, `python`, `svelte`, `config`, `vue`), many do not; **and**
2. the caller asked for **verbose**, since `properties` is inside the `wantVerbose` branch.

So what an agent can see about an entity's subtype depends on **which language produced it**, not on what the graph knows. Two entities with identical `Subtype` values are distinguishable through MCP if one came from Go and invisible if the other came from Java. `internal/mcp/flow_tools_test.go:472-484` shows the split is already known at test level — it describes an entity with "only the top-level Subtype set (no `Properties["subtype"]`)" as a case to handle — but nothing reconciles the two representations.

The canonical field is the invisible one. The visible one is an optional duplicate that some extractors remember to write.

## How it surfaced

Found during the adversarial review of #7089 (Java `@interface` support). That PR chose `SCOPE.Component` + `Subtype: "annotation"` partly to make *"list this project's own annotations"* answerable. It is not answerable — not because the entity is wrong, but because the field that distinguishes it does not cross the MCP boundary and Java does not dual-stamp. **I wrote that rationale into the brief, so the error is mine**, and it is corrected on that PR.

The narrower point generalises: **any design argument of the form "we encode it in the Subtype so a consumer can ask for it" is currently false** unless the extractor dual-stamps and the caller is verbose. That is a load-bearing assumption in more than one place.

## What is NOT claimed

- **Not that `Subtype` is unused.** It is hashed, and engine/resolver passes key on it (`engine.ClassLikeComponentSubtypes`, `links/sameas_pass.go`'s `modelSubtypes`). Inside the process it works. This is strictly about the MCP surface.
- **No incidence count.** How many entities carry a top-level `Subtype` with no `Properties["subtype"]` twin is unmeasured. It is cheap to get and is the first thing worth having, because it sizes whether this is a nuisance or a headline.
- **Not that emitting it is obviously free.** `serializeEntity` output is on the hot path for every MCP response; adding a key to every entity payload has a token cost that matters for the primary consumer. Whether it belongs in the default payload, in verbose only, or behind a filter parameter is a real design question and I am not settling it here.
- **Not a claim about the dashboard**, which I did not check.

## Suggested next step

Measure first: count entities with a non-empty top-level `Subtype` and no `Properties["subtype"]`, split by language. Then decide one representation and make the other derive from it, rather than leaving two fields that disagree by extractor.

**Grade a fix in both directions.** A subtype that is set must be visible; and — the direction that will be skipped — an entity with **no** subtype must not acquire an empty-string one that consumers then filter on. Assert the **emitted payload**, not the struct field, since the whole defect is that the struct field and the payload disagree.

Refs #7089, #7073, #7056.

cajasmota:
Closes #7090.

`Entity.Subtype` decides graph identity — `entityTupleKey` hashes `Kind|Name|QualifiedName|Subtype|SourceFile|StartLine|EndLine` — and engine passes key on it. `serializeEntity` (`internal/mcp/tools.go`) emitted `id / name / qualified_name / kind / file / line`, plus `end_line / language / repo` / algo fields / `properties` under verbose. **No `subtype` on either path.** So the field decided identity while being invisible to every agent reading the graph through MCP.

## Measured before choosing a shape

14 golden fixtures through the real indexer, 846 entities:

| | |
|---|---|
| entities with a non-empty `Subtype` | **663 / 846 (78%)** |
| of those, with **no** `Properties["subtype"]` twin | **430 — 51% of all entities** |
| payload cost where present | **+11.7%** on the six-key default row |

**For half the graph the new key is the only route to the distinction.** There was no fallback, not a less convenient one.

A fourth variant was measured and discarded: suppressing the key when it merely restates `Kind` buys nothing — only **3 of 663** subtypes equal their kind tail.

## Shape: emit on the **default** payload, only when non-empty

- **Default, not verbose** — `find`/`list` is where the distinction is actually asked for, and verbose-only would have left most traffic unable to see it.
- **Conditional** — the 183 subtype-less entities pay nothing and never acquire an empty-string key that consumers would have to special-case.
- The cost is **not** negligible at +11.7%, and that is stated rather than waved away; it buys the only route for 51% of entities.

## What each language actually emits — and it contradicts the framing the work started from

This was checked per producer rather than assumed, and the check inverted the premise. The issue (and my brief) framed this as *some languages encode the distinction in `Kind`, others only in `Subtype`*. **No fixture language encodes class-vs-interface in `Kind` at all:**

- **java** — `class` and `interface` both land on Kind `SCOPE.Component` with no dual-stamp, so the payload key is the only route.
- **go** — its class-likes are **also** Subtype-only (struct/interface under `SCOPE.Component`); only its *file* entity dual-stamps. Both shapes graded.
- **typescript** — does promote some distinctions into `Kind` (`SCOPE.Enum`, `SCOPE.Schema`, `SCOPE.UIComponent`) and dual-stamps most rows, so here the key is additive and must **agree** with `Kind` rather than replace it.

The language axis is real, but it is not the axis it was thought to be — and it is graded in all three shapes.

## Grading

The guard has exactly **1 occurrence** in `serializeEntity`. Each mutant `go vet`'d before scoring, applied with an asserted 1/1 match count, verdicts by `--- FAIL` line counts, `-count=1`, restored by `cp` from a sha-verified backup with the post-restore sha matched and `git diff --stat` showing only the fix.

**6 mutants, 6 DEAD:** delete the key **(3)** · drop the guard / always emit **(1 — killed by the negative-direction test, which is the point of having one)** · invert the guard **(4)** · value→`e.Kind` **(3)** · value→a constant **(3)** · move inside `wantVerbose` **(3)**.

Both directions are scored: an entity **with** a subtype must show it, and an entity **without** one must not gain an empty key.

## Verification

Full `./internal/mcp/` green, exit 0, **0 `--- FAIL`** (239s) — no golden or strict-equality payload expectation moved, which was the specific risk for an additive key. `SCHEMA.md` updated for the new key.

## Not claimed

- **`./cmd/grafel/`'s digest test was not run to completion.** No extractor changed, so the digest is untouched by this diff — **argued, not executed.**
- **The dashboard has its own separate `serializeEntity`**, left alone as out of scope. It still cannot see a subtype.
- **The measurement is fixture-corpus-relative.** No large real-repo index was available (`~/.grafel` holds no `graph.fb`), so the 78% / 51% / +11.7% figures are the golden corpus's, not a production repo's.

## Follow-up found, filed separately

`Kind` and `Subtype` overlap inconsistently across producers — java keeps class/interface under one `Kind` while typescript promotes enum to `SCOPE.Enum` — and `Properties["subtype"]` remains an optional duplicate rather than deriving from the canonical field. That is a modelling question, deliberately not touched here.


cajasmota:
**Round 2 pushed (`9af0adc64`). CI green on 16 real check-runs. Merging.**

Review returned REQUEST-CHANGES on one item, and it was the right one: `serializeEntity` has exactly **2 callers, both in `handleGetNode`** — `grafel_inspect` only. `grafel_find`'s rows come from `serializeHits`, untouched, so the PR's headline rationale (*"find/list is where the distinction is asked for"* — my wording, in the body I wrote) was undelivered, and the new `SCHEMA.md` bullet sat **inside the `grafel_find` section** asserting the key was on a shape that did not have it.

**Fixed by delivering the rationale, not by softening the doc.** `serializeHits` now emits `subtype` on the default `matches` row, same shape as `serializeEntity`'s — conditional on non-empty. `matches` is plain-JSON marshalled via `jsonResult` rather than tabular-encoded, so a heterogeneous key set across rows costs nothing; `generated` was already conditional in exactly that way.

## A fourth surface, found while fixing the third

Not named in my review brief: find's **default** (non-`full`) output renders through `renderCompact`/`renderPerRepoSummary` → `hitsToTOON` with a **fixed column schema** (`id[,repo],name,kind,file,line,score`). An optional per-row column does not fit a fixed tabular schema at all — a different shape of problem from the JSON path, and it is find's *default*, so the most common route still shows no subtype.

The `search=substring|name|literal` route was also left alone, with a reason rather than a shrug: its `item` struct already disagrees with find's rows on nearly every key (`entity_id`/`source_file`/`start_line`, and an **unstripped** `kind`), so appending one more field would add a third spelling of the same row instead of closing the gap. Both are now filed as **#7107**, and `SCHEMA.md` states where the key is **not** emitted so the doc under-claims rather than over-claims.

## Two prose claims corrected — one by execution

- *"No fixture language encodes class-vs-interface in `Kind`"* was **false for typescript** (`interface` → `SCOPE.Schema`, `class` → `SCOPE.Component`). Rather than edit the sentence, the typescript fixture **gained a class** so the leg now *asserts* the distinction. A correction that becomes a test. I have also corrected and retitled **#7105**, which had inherited the same false claim from me.
- *"java … with no dual-stamp"* was **overbroad** — java does dual-stamp its 8 **file** entities; true only of its class-likes, which is what the test's premise actually asserted.

## Grading

**16 mutants, all DEAD**, and round 1's eight were **re-scored rather than carried** — the counts moved (invert 5→9, delete 4→7), which is exactly why that rule exists. `serializeHits` and `serializeEntity` are now twins and every mutant was applied to **both**, which is the pair audit the last eight findings this milestone have argued for.

`TestFindAndInspectSubtypeTwinsAgree_7090` turns the audit into an assertion — presence *and* value, both verbosities, all three producers, vacuity-guarded in both directions.

## My own mutant, which nobody else ran — DI-1, DEAD (2 FAIL)

The author's 16 and the reviewer's 8 all suppress, relocate or corrupt the key uniformly. I ran a **per-language partial suppression** instead: emit the subtype unless `Kind == "SCOPE.Schema"`, which disables it for exactly typescript's `interface` shape and nothing else.

Two failures, and precisely the two that should fail:

```
--- FAIL: TestSubtypeReachesFindRow_TypeScript_7090
--- FAIL: TestFindAndInspectSubtypeTwinsAgree_7090
```

The java, go and csharp legs stayed green. **That is a positive verification of the per-language independence claim**, not just a kill — each leg grades its own producer, and no leg is standing in for another. It also proves the twin test is not vacuous: it detects a find/inspect disagreement that no single-leg value assertion would catch, which is what it exists for.

`go vet` clean before scoring, exact match count asserted (1/1) so a silently-unapplied edit could not read as ALIVE, verdict by `--- FAIL` line count, `-count=1`, restored by `cp` with the shasum re-verified and a clean tree confirmed.

## Verification

Full `./internal/mcp/` exit 0, 0 `--- FAIL` · `gofmt -l` and `go vet` clean · CI green on 16 check-runs including both Windows legs and the extraction-quality gate. The review settled the digest question by **execution** rather than argument — `./cmd/grafel/`'s three digest tests PASS with fresh `GRAFEL_HOME`/`GRAFEL_DAEMON_ROOT`.

## Carried forward, unresolved

The corpus figures were re-measured by review over **38** fixture dirs (the author used 14): 1075 entities, 94.2% with a non-empty `Subtype`, **74.6%** with no twin, +13.07% payload. Different population, same direction — every figure strengthens the design choice. Dual-stamped rows **never disagreed** with `Subtype` (0/211), so the duplicate is redundant where it exists today. The dashboard's separate `serializeEntity` still emits no subtype.

PR #7110, opened 2026-09-17
test(#7106): both supervisor give-up budgets shipped an ungraded production default — pin each behaviourally, in both directions

### LINKED ISSUE #7106: MEASURED: both engine-supervisor give-up budgets ship an UNGRADED production default — setting maxSpawnFailures or maxCeilingHits to 1000 leaves ./internal/daemon fully green
Found by mutation while reviewing #7103. **Both of the engine supervisor's give-up budgets ship a production default that no test observes.**

Measured on `fb369f0b1`, each mutant applied with an asserted match count of 1, `go vet` clean before scoring, `--- FAIL` lines counted, `-count=1`, restored by `cp` with the shasum re-verified:

| mutant | verdict |
|---|---|
| `defaultEngineMaxSpawnFailures` 3 → **1000** | **ALIVE** — `./internal/daemon/` exit 0, **0 `--- FAIL`** |
| `defaultEngineMaxCeilingHits` 3 → **1000** | **ALIVE** — `./internal/daemon/` exit 0, **0 `--- FAIL`** |

The mechanism around both is thoroughly graded — between them, #7103's author and its reviewer scored 15 mutants on the run loop, the resets, the budget comparison, the fatal-error text and the log verbs, and all but two provable equivalents died. **The production constants are the one thing none of them touched**, and they are free.

## Why this happens, and why it is the interesting part

The supervisor's tests **inject tuning**. That is good design — it is what lets them assert a behaviour in 1.5s instead of 91s. But injecting a knob so a control can drive it is precisely what frees every caller, including production, from the default's grading. The suite pins *"the budget mechanism works at whatever value you hand it"* and says nothing about the value users actually run.

The practical consequence is specific to what #7103 set out to fix. Its whole purpose is that a deterministic `Start` error should die in ~1.5s rather than walk a ~91s budget. **Set the default to 1000 and that regression is fully restored, with a green suite.** The fix is pinned only at tunings no user runs.

## Not a regression, and not #7103's fault

`defaultEngineMaxCeilingHits` predates #7103 and is equally ungraded, which is why this is filed rather than blocked on that PR: it is a pre-existing pattern that #7103 followed faithfully. **Both are named here deliberately** — scoring one twin says nothing about the other, and fixing only the new one would leave the older, larger budget (~91s) just as free. That twinned-site shape has produced eight findings this milestone.

## The cheap kill, so "too expensive" is not assumed

A constant-equality assertion (`if defaultEngineMaxSpawnFailures != 3`) is **not** the fix — it restates the value rather than observing a consequence, and it would survive any change that keeps the number while breaking the behaviour.

The behavioural pin is a supervisor run with **no tuning injected**, asserting the unspawnable verdict arrives within a bounded number of spawn attempts, counted by the injected command factory (the external observation #7103 already uses). At default tuning the two waits are 0.5s + 1s, so **the test costs ~1.5s of wall clock** — entirely affordable. The crash-budget equivalent is more expensive (~91s) and should probably be pinned by asserting the *attempt count* rather than waiting out the backoff; the wait is injectable separately from the budget.

**Do not gate either assertion on a wall clock.** #7062 in this same package has a 100ms behaviour behind two 5s deadlines with a failure message that blames a leak when what actually blew was the test's own budget.

## What is NOT claimed

- **Not that either default value is wrong.** 3 and 3 look reasonable. The defect is that nothing would notice if they changed.
- **Not that this is unique to the supervisor.** It is a general consequence of injectable tuning, and other packages very likely share it — but I measured these two, not a survey.
- **Not measured beyond `./internal/daemon/`.** A test elsewhere could in principle observe these constants; nothing in that package does, and that is where every supervisor test lives.

Refs #7103, #7087, #7062.

cajasmota:
Closes #7106.

Both engine-supervisor give-up budgets shipped a **production default that no test observed**. Measured before this change, each mutant applied with an asserted match count of 1, `go vet` clean, `--- FAIL` lines counted:

| mutant | verdict |
|---|---|
| `defaultEngineMaxSpawnFailures` 3 → 1000 | **ALIVE** — `./internal/daemon/` exit 0, 0 `--- FAIL` |
| `defaultEngineMaxCeilingHits` 3 → 1000 | **ALIVE** — same |

The mechanism around both was already graded to the hilt — 15 mutants across #7103's author and reviewer covered the run loop, the resets, the budget comparison, the fatal-error text and the log verbs, and all but two provable equivalents died. **Nothing touched the constants.** The supervisor's tests inject tuning, which is good design and is exactly what frees production from the default's grading: the suite pinned *"the mechanism works at whatever value you hand it"* and said nothing about the value users run. **Set `maxSpawnFailures` to 1000 and #7103's entire purpose reverted with a green suite.**

No production code changes here — one new test file, 195 lines.

## Pinned behaviourally, not by restating the constant

A constant-equality assertion (`if defaultEngineMaxSpawnFailures != 3`) would restate the value rather than observe a consequence, and would survive any change that kept the number while breaking the behaviour. Instead:

- **`defaultEngineMaxSpawnFailures`** — a supervisor with **no tuning injected at all**, child command pointing at a nonexistent binary so `cmd.Start` fails deterministically, attempts counted by the **injected factory** (an external observation, not an internal counter). Asserts the fatal text names `unspawnable` and attempts ≤ 5, with a positive control that the log stream really carries `no child process was created`.
- **`defaultEngineMaxCeilingHits`** — **the wait proved separable from the budget**: `waitBackoff` owns `backoffInitial`/`backoffMax` while `backoffAndMaybeGiveUp` owns `maxCeilingHits`. So the test shrinks only the two waits to 1ms/2ms and leaves `maxCeilingHits` and `healthyUptime` exactly as `newEngineSupervisor` set them. **The ~91s budget is graded in 0.6s and never waited out.** Asserts `crash-looping` in the fatal text and attempts ≤ 6.

No behavioural assertion is gated on a short wall clock — #7062 in this package has a 100ms behaviour behind two 5s deadlines with a failure message that blames a leak when what blew was the test's own budget.

## Both directions, and both twins

Each test also carries a **lower** bound: spawn ≥ 2 attempts (a single construction failure can be transient and must not kill serve) and crash ≥ 3 relaunches. A pin that only catches "too large" is half a pin.

| mutant | verdict | `--- FAIL` |
|---|---|---|
| `defaultEngineMaxSpawnFailures` 3 → 1000 | **DEAD** (was ALIVE) | 1, in 15.6s |
| `defaultEngineMaxSpawnFailures` 3 → 1 | **DEAD** | 1, in 0.01s |
| `defaultEngineMaxCeilingHits` 3 → 1000 | **DEAD** (was ALIVE) | 1, in 0.61s |
| `defaultEngineMaxCeilingHits` 3 → 1 | **DEAD** | 1, in 0.21s |

Each constant has exactly one use site (`newEngineSupervisor`). **Both twins are pinned and both are scored in both directions** — `defaultEngineMaxCeilingHits` predates #7103 and was equally free, so fixing only the newer constant would have left the older, larger budget just as unpinned. Scoring one twin says nothing about the other.

Restores were by `cp` from a byte-level backup with the shasum **and** `git diff HEAD` verified empty after each.

## Cost

The new pair costs **3.4s** wall clock when green — ~1.5s on the spawn arm from the real 0.5s+1s production waits, ~1.9s on the crash arm from four real subprocess spawns. That was priced before the work started rather than assumed affordable.

## Verification

`go test -count=1 -timeout 15m ./internal/daemon/...` → **exit 0, 0 `--- FAIL`**, all 19 packages ok · `go test -count=1 -race -run TestEngineSupervisor_ ./internal/daemon/` → ok, 38.8s · `gofmt -l internal/daemon/` clean · `go vet` clean.

## Not claimed

- **The 1000-mutant kill on the spawn arm takes 15.6s**, because the *production* backoff doubles (0.5+1+2+4+8s to reach attempt 6). Only the failing case pays that; the 60s outer bound covers it. Tightening the bound to 3 would restate the constant, so it stays at 5.
- **Four more production defaults in this same file remain ungraded, same shape:** `defaultEngineBackoffInitial`, `defaultEngineBackoffMax`, `defaultEngineHealthyUptime`, `defaultEngineDrainTimeout`. The crash test here injects the first two itself, so nothing observes their defaults either. **Not measured** — reported rather than swept in, and being filed separately.
- **Nothing ran on Windows or Linux.** Both tests are cross-platform by construction (they reuse `isolateSpawnFailEnv`/`instantExitCommand` from the untagged `supervise_spawnfail_test.go`), but off macOS that is derived-not-executed; the ubuntu and windows legs are the real signal.


cajasmota:
**Independent review: APPROVE.** Plus one factual correction to this body, and one finding of my own.

## What review re-derived rather than took on trust

- **Both kills are real and of the right thing.** `maxSpawnFailures` 3→1000 DEAD (1 FAIL, 15.5s, *"made 6 spawn attempts"*); `maxCeilingHits` 3→1000 DEAD (1 FAIL, 0.63s, *"7 attempts"*). Both fail on an **emitted** observation — injected-factory attempt count, fatal text, log stream — not an internal counter.
- **The bounds have genuine slack and are not the constant renotated.** Review scored the neighbours: `maxSpawnFailures` → 2 **ALIVE** and → 5 **ALIVE**; `maxCeilingHits` → 2 **ALIVE** and → 5 **ALIVE**. So each pin accepts a band of 2–5 against baselines of 3 and 4, and neither test names a constant outside a comment. No constant-equality coupling.
- **Separability holds, by source and by measurement.** `backoffAndMaybeGiveUp` compares `s.maxCeilingHits` while `waitBackoff` touches only `backoff`/`backoffMax` — and `defaultEngineBackoffMax` 30s→3ms is **ALIVE** across the whole package, which demonstrates the crash test's bound does not observe the injected waits, while a `maxCeilingHits` mutant moves attempts 1:1. That was the one way this fix could have reproduced #7106's own defect inside itself; it does not.
- **The positive controls are non-vacuous.** Deleting the `— no child process was created` suffix from the production log kills the spawn test; renaming the crash-path `engine child exited` warn kills the crash test.

## Correction to this body — the four-defaults claim is wrong on one of four

Review spot-checked all four against the full `./internal/daemon/` suite:

| default | verdict |
|---|---|
| `defaultEngineHealthyUptime` 60s → 1ms | **DEAD** — killed by **this PR's own crash test** |
| `defaultEngineBackoffInitial` 500ms → 1ns | ALIVE |
| `defaultEngineBackoffMax` 30s → 3ms | ALIVE |
| `defaultEngineDrainTimeout` 5s → 1ns | ALIVE |

`defaultEngineHealthyUptime` **is** now graded in the shrink direction: the crash test's children live ~122ms, so a tiny healthy-uptime resets the budget forever and attempts blow past the bound. So the follow-up is **three** ungraded defaults, not four, and one of them was incidentally fixed by this change. I asked for mutants rather than inference here precisely because I was about to file that issue — the claim in this body was reasoned, and reasoning got one of four wrong.

## Two honest weakenings worth stating plainly

- **The lower bounds forbid exactly the value 1.** spawn→1 DEAD, crash→1 DEAD, but →2 is ALIVE on both, so each lower bound is a **single-value guard**, not a band. That is sufficient for its stated purpose — no death on the first hiccup — but "both directions pinned" oversells it and this body should not.
- **The crash control's log grep is slightly broader than its comment.** It matches the substring `engine child exited`, which also matches the drain line `engine child exited after SIGTERM` (`supervise.go:632`). Secondary assertion, not blocking, but the comment claims more precision than the grep has.

## My own mutant, which nobody else ran — DK-1, DEAD (1 FAIL)

The author's 4 and the reviewer's ~10 all vary a budget's **value**. None tested whether the two budgets are **distinguishable from each other** — and they ship the same default, 3, which is exactly the twinned-value shape that has produced nine findings this milestone.

So I swapped the field rather than the value: `spawnFailures >= s.maxSpawnFailures` → `spawnFailures >= s.maxCeilingHits`. Under production defaults this is a **no-op**, because both are 3.

**DEAD, 1 FAIL — and the killer is the informative part:** `TestEngineSupervisor_RepeatedSpawnFailureStopsSoonerAndIsDistinguishable`, from #7103, not either of this PR's new tests. It catches the swap *because it injects different values for the two budgets*. This PR's default tests structurally cannot catch it, since both defaults are 3.

**The two test styles are complementary and neither is sufficient alone.** The injected-tuning test grades *field identity*; the no-tuning default test grades *the shipped value*. #7106 existed because the second was missing; this mutant confirms the first is doing work the second cannot do. Worth keeping as the reason not to "simplify" either away.

`go vet` clean before scoring, exact match count asserted (1/1), verdict by `--- FAIL` line count, `-count=1`, restored by `cp` with the shasum re-verified and a clean tree confirmed.

## Verification re-run by review

`go test -count=1 -timeout 15m ./internal/daemon/...` exit 0, 0 `--- FAIL`, 18 packages, 184s · `-race -run TestEngineSupervisor_` exit 0, 0 FAIL, **0 DATA RACE**, 38.2s · `gofmt -l` and `go vet` clean · **cost of the new pair measured at 1.94s** (1.52s spawn + 0.42s crash), so the 3.4s claim is accurate and arguably conservative · no FIFOs or sockets outside `t.TempDir()`.

## Not verified

Windows and Linux — macOS only locally; cross-platform-by-construction is derived-not-executed and the CI legs are the signal (`ci:full` is on, so the macOS leg runs too). Nobody re-scored #7103's 15 mechanism mutants. And nobody probed whether the crash test is sensitive on a machine where child uptime exceeds 60s — worth remembering if it ever flakes.

PR #7123, opened 2026-09-17
test(#7111): pin the three remaining ungraded supervisor defaults — and record backoffMax's growth direction as ungraded rather than fake it

### LINKED ISSUE #7111: MEASURED: three more supervisor production defaults are ungraded (backoffInitial, backoffMax, drainTimeout) — healthyUptime turned out DEAD, so injectable tuning freeing the default is a pattern in this file, not two incidents
#7106 established that the engine supervisor's two **give-up budgets** shipped production defaults nothing observed, and #7110 pinned both. The same shape holds for the file's remaining tuning constants. Measured, not inferred — each mutant run against the full `./internal/daemon/` suite:

| default | mutant | verdict |
|---|---|---|
| `defaultEngineBackoffInitial` | 500ms → **1ns** | **ALIVE** |
| `defaultEngineBackoffMax` | 30s → **3ms** | **ALIVE** |
| `defaultEngineDrainTimeout` | 5s → **1ns** | **ALIVE** |
| `defaultEngineHealthyUptime` | 60s → **1ms** | **DEAD** — killed by #7110's crash test |

**Three, not four — and the correction is the point.** #7110's author reported four, reasoning that their own crash test injects `backoffInitial`/`backoffMax` and therefore observes none of these defaults. That reasoning was right about three and wrong about `defaultEngineHealthyUptime`: the crash test's children live ~122ms, so shrinking healthy-uptime to 1ms resets the budget forever and attempts blow past the bound. It is now graded in the shrink direction **as a side effect** of pinning a different constant. The claim was filed only after mutants were run, because inference got one of four wrong.

## Why this keeps happening

The supervisor's tests **inject tuning**, which is good design — it is what lets #7110 grade a ~91s budget in 0.6s. But injecting a knob so a control can drive it is exactly what frees every caller, production included, from the default's grading. The suite pins *"the mechanism works at whatever value you hand it"* and says nothing about the values users actually run.

This is now a demonstrated **pattern in this file**, not two incidents: five constants examined, four of them ungraded at the time of examination.

## What a fix must not be

**Not a constant-equality assertion.** `if defaultEngineBackoffMax != 30*time.Second` restates the value rather than observing a consequence, and survives any change that keeps the number while breaking the behaviour.

#7110 set the usable pattern: run the supervisor with **no tuning injected**, and assert an **emitted** consequence — the injected command factory's attempt count, the fatal error text, or the log stream — with a **band** rather than an exact value. Its bounds accept 2–5 against baselines of 3 and 4, verified by scoring the neighbours (→2 and →5 both ALIVE), so the pin is not the constant in another notation.

## Per-constant notes

- **`defaultEngineBackoffInitial` and `defaultEngineBackoffMax`** are the wait shape, not a budget. Their observable consequence is elapsed time, which is the one thing that must **not** be asserted directly — #7062 in this package has a 100ms behaviour behind two 5s deadlines whose failure message blames a leak when what actually blew was the test's own budget. A usable observation may be the *number of attempts within a bounded window*, or the recorded backoff in the log line, rather than wall-clock duration. **This one needs thought before it needs code.**
- **`defaultEngineDrainTimeout`** bounds the graceful SIGTERM→exit wait before escalating to SIGKILL. Its consequence is observable in the drain path's emitted artefacts (`engine child exited after SIGTERM` versus a force-kill), which is a cleaner target than the two backoff constants.

## What is NOT claimed

- **Not that any default value is wrong.** 500ms / 30s / 5s look reasonable. The defect is that nothing would notice if they changed.
- **Not that this is unique to the supervisor.** It is a general consequence of injectable tuning and other packages very likely share it. Five constants in one file were measured; that is not a survey.
- **Not equal priority with #7106.** Those were *give-up* budgets, where a wrong value means the daemon dies or hangs. These are timing and drain constants — real, but a smaller blast radius.
- **Measured on macOS only**, against `./internal/daemon/` at `8381a9241`.

Refs #7106, #7110, #7103, #7087, #7062.

cajasmota:
Closes #7111.

> **Round 2 retracted this PR's central claim.** The original version declared `defaultEngineBackoffMax`'s **growth** direction unpinnable without a multi-minute wall-clock test. It is pinnable in **0.11s**. The first commit's message still carries the original wording and is superseded by this body.

Three supervisor production defaults were ungraded — raising or lowering them left `./internal/daemon/` fully green. Test-only change; no production code touched.

## All four mutants re-run, and the issue's own count was wrong

| default | mutant | before | after |
|---|---|---|---|
| `defaultEngineBackoffInitial` | 500ms → 1ns / → 30s | ALIVE | **DEAD** |
| `defaultEngineBackoffMax` | 30s → 3ms (shrink) | ALIVE | **DEAD** |
| `defaultEngineBackoffMax` | **30s → 10m (growth)** | **ALIVE, even against round 1's four pins** | **DEAD** |
| `defaultEngineDrainTimeout` | 5s → 1ns / → 60s | ALIVE | **DEAD** |
| `defaultEngineHealthyUptime` | 60s → 1ms | **DEAD** | untouched |

`healthyUptime` was already graded — killed by `TestEngineSupervisor_DefaultCrashBudgetIsBounded`, i.e. as a side effect of #7110 pinning a *different* constant. **Three, not the four this issue was filed with.**

## The growth direction: the seam, and why the first attempt missed it

`backoffAndMaybeGiveUp` tests `*backoff >= s.backoffMax` **before** `waitBackoff` sleeps. So injecting `backoffInitial` (*not* the constant under test) together with `maxCeilingHits = 1` (already graded by #7110) makes give-up arrive with **zero sleeping**, exactly when `backoffInitial >= backoffMax`. `backoffMax` is never injected.

Band proven by scoring the neighbours: **20s / 45s / 60s ALIVE; 15s / 1s / 3ms DEAD** → exactly `15s < backoffMax ≤ 60s`.

**Review's probe used fixed 1s/2s observation windows; those were rejected.** They **failed under `-race`**, where spawning and killing a child takes over a second, so the lower arm's own positive control fired. Replaced with an **event-driven** wait on the two mutually exclusive emitted lines — the crash-loop fatal versus `waitBackoff`'s `relaunching engine after backoff` — under a single 30s machinery bound that asserts nothing about any default. **Load now changes *when* the answer arrives, never *which*.** Cost fell from review's 1.11s to **0.11s** per arm.

## The other observations, and one deliberately rejected

- **`backoffInitial`** — spacing between spawn attempts 1 and 2, timestamped by the injected command factory, no tuning injected. Band 100ms–5s.
  **Not** the logged `backoff=` value: `waitBackoff` logs it *before* creating the timer, so that assertion would survive a break which keeps the number and stops sleeping. A value the code reports about itself is a diagnostic signature, not a regression pin.
- **`backoffMax` shrink** — with a 400ms first wait injected, the second wait must still be a real doubling (≥600ms) rather than clamped.
- **`drainTimeout`, both directions** — `terminateChild`'s two mutually exclusive lines, matched on **`exited after SIGTERM`** rather than the `engine child exited` prefix it **shares with the crash path**.

## Three non-floor bounds, not two

The original body claimed "every bound but two is a floor". Corrected — there are **three**, each recorded with its slack in the header of the file carrying it:

| bound | kind | slack |
|---|---|---|
| `gap < backoffInitialFloor` (100ms) | floor | load-proof |
| `gap > backoffInitialCeiling` | ceiling | **10×** |
| `gap < backoffCeilingFloor` | floor | load-proof |
| drain graceful `!contains(forceKillLine)` | **ceiling** | **12.5×** |
| drain escalation `waitFor(20s)` | ceiling | **4×** |

Both new growth-pin arms carry **no timing bound on any default** — that is the point of the event-driven wait.

## Grading

**18 mutants, re-scored from scratch, no carried counts** — full `./internal/daemon/` suite, `-count=1`, `go vet` per mutant, an exact match count asserted on each edit, `--- FAIL` lines counted, restore verified by shasum **and** `git diff HEAD` (ends 0 lines changed). Scored **twice**, before and after the `-race` fix; verdicts identical.

**Round 1's record is corrected, not carried:** its "backoffMax 1s/2s ALIVE" no longer holds — the new lower arm kills them.

## Cost and verification

New pins **7.58s** total (0.52 / 1.22 / 0.11 / 0.11 / 0.52 / 5.14); the 5.14s is the production drain window itself and is irreducible without injecting the constant under test.

`./internal/daemon/...` exit 0, 0 `--- FAIL` · `-race -run TestEngineSupervisor_` exit 0, 0 races · **3× `-count=5 -race` on the new pins** exit 0 · `gofmt -l` and `go vet` clean · zero stray helper processes · nothing written to the real `~/.grafel`.

Independently verified by review before round 2: independence confirmed **both ways** (mutating either backoff constant moves only its own pin), and a flake hunt of **8 repeat rounds** including 3 under 6-way CPU load (loadavg 7.5→11.1 on 12 cores) with max drift +0.09s.

## Not claimed

- **Windows and Linux**: macOS only. The drain file is `darwin || linux` by build tag — on Windows `signalTerminate` **is** `p.Kill()`, so a child cannot ignore the first signal and the drain window is not observable there at all. CI is the signal for the untagged backoff pins.
- **`backoffMax`'s second role is still ungraded** — filed as **#7126**. It is both a give-up threshold (now pinned both ways) **and** a clamp on the actual wait; deleting the clamp leaves the package green, and both arms here are blind to it by construction, the growth arm precisely because it never sleeps.
- Two **pre-existing** crash-path controls key on the shared `engine child exited` prefix (`supervise_defaults_test.go:189`, `supervise_spawnfail_test.go:173`). Both also assert the crash-loop fatal, so neither is vacuous. Left alone; this PR's own crash control matches `uptime=` instead and pins the log handler via `EnvDaemonLogJSON` for format stability.


cajasmota:
**Round 3 (`2519bd82a`): the ubuntu failure is diagnosed, and it was not this PR. My own diagnosis brief was wrong on three of the five facts it rested on — corrected below, because I handed those to the agent as evidence.**

## What actually broke

This PR carries `ci:full`, so **both of its runs are `-race`**. Under `-race` the `internal/daemon` test binary is ~126 MB, every crash cycle re-execs it, and `TestEngineSupervisor_UnkeepableEngineIsFatal` costs **6.5s against its own 15s budget on an idle machine**. A 2.3× margin, on a shared 4-core runner. I asserted in the brief that it "should surface the fatal in well under a second" — that is true without `-race` (0.52s) and wrong with it.

## My five "facts", each refuted with evidence

| my claim | what the data says |
|---|---|
| Round 1 passed with the same files, so package growth is not the cause | Round 1's `pull_request` run had `internal/daemon` at **215.3s and passed**; round 2's at **205.2s and failed**. Round 2's package was **faster**. |
| Six consecutive `test` greens today, round 2 is the first red | **Not the same regime.** Those runs have daemon 34–38s, `caps` 0.007s, `extract` 30s. Both of this branch's runs have 205–215s, 1.03s, 209–219s — the uniform ~7× of `-race`. **The comparable set is 2 runs, not 7.** |
| Windows passed with the same untagged growth arms | **`supervise_test.go` is `//go:build darwin \|\| linux`.** The failing test **does not exist** on Windows. That leg is silent, not exculpatory — and I read it as evidence in a message to the agent. |
| Local repro was clean | Because I ran it **without** `-race` (0.52s). With `-race`: 6.5s, every time. |
| File order + zero-sleep give-up | Consistent with added fork pressure, but not a distinct mechanism — and the proposed mechanism (binary size + exec cost) is **not POSIX-specific**, which is why it is `-race`-gated rather than platform-gated. |

Two lessons I am taking from that: **compare CI runs only within the same regime** (a `-race` run and a non-`-race` run of the same package are not comparable), and **check build tags before treating a passing platform as evidence**.

## The three hypotheses, refuted rather than dismissed

1. **Cleanup** — the probe does `defer sup.stop()` *inside* the helper, so the supervisor is joined before the test body asserts, strictly earlier than the pre-existing tests. `stop()` cannot hang: `run()` has already returned and closed `doneCh` on the give-up path, and every blocking point selects on `stopCh`.
2. **Leaked child** — the verdict is reached only *after* `<-waitCh`, i.e. after `cmd.Wait()` reaped it. Nothing is left to reap; zero stray processes measured.
3. **Interaction** — `-race -count=3` of the growth arms plus the failing test: **12/12 PASS, 0 `--- FAIL`**.

## And a real latent defect, found while diagnosing

`TestEngineSupervisor_UnkeepableEngineIsFatal` set `healthyUptime = 10 * time.Second` with the comment `// never "recovers"`. **That comment was prose nothing observed, and it was false**: `run()` resets **both** `backoff` and `ceilingHits` once a child's lifetime reaches `healthyUptime`, so a single slow child makes the fatal **structurally unreachable**. Forcing that branch reproduces the CI failure **byte for byte** — same line, ~15.0s.

Changed in the pre-existing test:

- `healthyUptime` 10s → **`time.Hour`**, so "never recovers" is now true by construction rather than by luck.
- A **positive control**: the fatal must arrive on ≥ `maxCeilingHits+1` spawns, so it cannot be #7087's *unspawnable* verdict in disguise.
- **Failure message** — names the budget as the **test's own**, and reports elapsed time plus spawn attempts counted from outside (the injected command factory), so "spawns are slow" and "the supervisor is not relaunching" are distinguishable. Demonstrated by forcing the branch: `"22 child spawn attempt(s) in 3.002s"`. Budget 15s → 90s (~14× idle `-race`, free on success).

That message was the #7062 shape — it blamed the supervisor for what was the test's own budget — and it is the reason this cost as much diagnosis as it did.

## Verification

`go test -count=1 ./internal/daemon/` exit 0, 0 FAIL, 47.3s · `go test -race -count=3 -run TestEngineSupervisor_` exit 0, 0 FAIL, 145.7s (Unkeepable 6.54 / 6.55 / 6.55s) · `gofmt -l` and `go vet` clean · no stray helpers.

**Not verified:** the `-race` margin figure is macOS-idle; the ubuntu number is inferred from the CI package timings above. Linux cannot be run here.

cajasmota:
**Merge gate: independent round-3 review APPROVE, CI green on all 7 workflows, and my own mutant found one thing the reviewer flagged but did not score.**

## Independent review (round 3 only, `506cc5697..2519bd82a`) — APPROVE

The load-bearing claim is confirmed independently, with a forced reproduction in both arms at the same slow-child load (child factory → `/bin/sleep 11`):

- `healthyUptime = 10s` (the old value): **1 `--- FAIL`**, no fatal ever, 9 spawns in 90.001s.
- `healthyUptime = time.Hour` (this PR): **PASS in 44.16s**, verdict = the crash-loop one.

Same input, opposite verdict — so the old test's give-up path was contingent on machine speed rather than on supervisor behaviour. The 44.16s is worth noting on its own: even *with* the fix, a slow-spawn machine blows past the old 15s budget, which justifies the budget raise as a machinery bound independently of the flake.

They also confirmed, from the repo rather than from my comment: `run()` resets **both** counters at `uptime >= healthyUptime` with `uptime` measured from *before* `cmd.Start()`; the crash-loop path is still what fires under `time.Hour` (4 spawns, `ceilingHits` 1→2→3); and `GOOS=windows go list -f '{{.TestGoFiles}}'` does not include `supervise_test.go` while `GOOS=linux` does. Their own `-race` timing: 6.59/6.59/6.60s vs 0.51s non-race — **12.9× for this test**, confirming the regime direction (the ~7× in my comment is a package-level figure and they did not re-measure it).

One observation of theirs I am recording rather than fixing: the new failure message's dichotomy is **incomplete** — a `healthyUptime` rewind presents as "the supervisor really is refusing to give up" (their 10s repro printed 9 spawns, which is > 4). That is a third cause the message folds into the second. Defensible, and `time.Hour` makes it unreachable here.

## My own mutant: DR-1, and the composite that matters → #7130

The reviewer flagged, but did not score, that the positive control's discrimination rests on an **uninjected** production default: it wants ≥ `maxCeilingHits+1` = 4 spawns, and the unspawnable verdict arrives at `defaultEngineMaxSpawnFailures = 3`. A margin of exactly one. I scored it.

**DR-1** — `defaultEngineMaxSpawnFailures = 3` → `4`: `go test -count=1 ./internal/daemon/` **EXIT=0, 0 `--- FAIL`. ALIVE.** #7110's pins accept the value (→2 and →5 were both measured ALIVE there), so nothing objects to 4.

Then the composite — force the #7087 unspawnable path and score it at each default:

| default | fatal arrives at | `--- FAIL` | control |
|---|---|---|---|
| `3` (today) | 3 spawns, wants ≥4 | **1** | **fires** |
| `4` (DR-1) | 4 spawns, wants ≥4 | **0** | **silent** |

So at 4 the unspawnable verdict is accepted *as a crash-loop verdict* and the package stays green. This is the mutually-masking shape: **#7110's band is too wide to see 4, and #7123's control needs the default to be under 4 but cannot say so.** Neither guard is wrong alone; together they leave the value ungraded while a test depends on it.

**Not blocking.** The control works at today's default — the reviewer's forced repro and my cell A both fire it, at the exact line and message. Filed as #7130 with three fix options (my preference: inject `maxSpawnFailures`, the one tuning knob this test leaves at the default).

## CI

All 7 workflows green on `2519bd82a` under `ci:full` (so `-race`): `test` (ubuntu + windows + python ratchet), node-type gate, quality, board-hygiene, module hygiene, cross-platform compile, windows installers. Note the `test (macos-latest)` leg did not appear in this run despite the label; `supervise_test.go` is `//go:build darwin || linux`, so ubuntu is the leg that matters for it, and I ran the darwin side locally myself (`-race -count=3`, 0 FAIL).

## Correction carried into the squash body

Round 1's commit message declares `backoffMax`'s growth direction **unpinnable**. That was refuted in round 2 — it is pinnable in 0.11s through a seam, and 30s→10m was ALIVE before it was pinned. The squash body is being written from a corrected file rather than concatenated commit messages, so the merged history does not carry the retracted claim.

Merging. Closes #7111; #7130 carries the residue.

cajasmota:
**Correction to the squash body — it names the wrong two defaults, and the error is now in `main`'s history where I cannot edit it.**

The first paragraph of `0f1011bd9` says:

> Three supervisor defaults were documented as pinned and were not: mutating `defaultEngineMaxSpawnFailures` and `defaultEngineMaxCeilingHits` left `./internal/daemon/` green…

**Wrong.** Those two are the DH-1/DH-2 mutants, which I filed as **#7106** and which **#7110** pinned. They are not what this PR touched.

The three defaults **this** PR pinned are the ones #7111 actually named — `backoffInitial`, `backoffMax`, `drainTimeout` — and the tests it added say so:

```
TestEngineSupervisor_DefaultBackoffInitialIsAHumanScaleWait
TestEngineSupervisor_DefaultBackoffCeilingIsReachedByAMinuteScaleWait
TestEngineSupervisor_DefaultBackoffCeilingLeavesRoomToGrow
TestEngineSupervisor_DefaultBackoffCeilingLeavesRoomBelowAMinute
TestEngineSupervisor_DefaultDrainTimeoutSurvivesASlowExit
TestEngineSupervisor_DefaultDrainTimeoutEscalatesPromptly
```

The rest of the body stands unchanged: the `backoffMax` growth-direction retraction, the `healthyUptime` latent defect, the failure-message rework, the budget raise, and the `-race` diagnosis are all this PR's and all correctly described.

**What does not change about #7130.** DR-1 mutated `defaultEngineMaxSpawnFailures` 3→4 and found it ALIVE. That is a statement about **#7110's** band being wide enough to accept 4 — which is how #7130 already attributes it — combined with this PR's positive control needing the value to be under 4. The composite finding is unaffected; only the paragraph above was misattributed.

This is the fourth time this milestone that my own *framing* of a change was wrong while the evidence under it was sound. The pattern is specific: I write the summary from what I remember measuring rather than from the diff in front of me. Grounding here would have cost one `git show --stat`.

Issue #7124, opened 2026-09-17
signature builders text-surgery family, item 2 only: grade buildMethodSignature + buildClassSignature STRUCTURALLY (item 1 shipped in e4dee929e; item 3 is a measured negative; siblings are #7133)


cajasmota:
#7117 and #7116 were filed as Java field-signature defects. Both are instances of one habit — **cutting or rewriting a raw source span by text** — and the habit is in at least **six more builders across four languages**. Verified at source, not inferred.

## 1. The same first-`=` cut, in three more extractors

`buildLetSig`, three near-identical copies:

- `internal/extractors/fsharp/extractor.go:604`
- `internal/extractors/reasonml/extractor.go:270`
- `internal/extractors/rescript/extractor.go:369`

```go
sig := strings.TrimSpace(decl)
if idx := strings.Index(sig, "="); idx >= 0 {
    sig = strings.TrimSpace(sig[:idx])
}
```

Identical to #7117's mechanism. An F# attribute with a named argument precedes the binding — `[<Foo(x = 1)>] let y = 2` — so the cut lands **inside the attribute** and the signature loses the binding entirely. ReasonML and ReScript have the same exposure via any `=` appearing before the binding's own.

**Not measured.** I confirmed the code, not the emitted output; whether attribute-before-`let` is idiomatic in each of the three needs a probe. #7117's Java case was measured before filing and this one is not — treat the three as "same shape, unmeasured" rather than three confirmed defects.

## 2. **The interesting one: the two Java siblings disagree about ordering, and one is right**

`buildMethodSignature` (`java.go:2168-2176`) strips annotation arguments **before** cutting at the body brace, and says why:

```go
// Strip annotation arguments FIRST to remove braces inside annotation args
// like @DeleteMapping("/{id}") → @DeleteMapping, so the body-brace search
// doesn't get confused by braces in annotation strings.
raw = stripAnnotationArgs(raw)
if idx := strings.Index(raw, "{"); idx >= 0 { raw = raw[:idx] }
```

`buildClassSignature` (`java.go:2186-2200`) does it **in the opposite order** — brace cut first, `stripAnnotationArgs` afterwards:

```go
if idx := strings.Index(raw, "{"); idx >= 0 { raw = raw[:idx] }
…
raw = stripAnnotationArgs(raw)   // too late
```

[...]

PR #7128, opened 2026-09-17
fix(#7124): buildClassSignature cut at a brace inside an annotation string, losing the class name — and the sibling whose comment documents that hazard was itself ungraded

cajasmota:
First of #7124's sequenced items. Refs #7124.

`buildMethodSignature` strips annotation arguments **before** cutting at the body brace, and its comment says exactly why:

```go
// Strip annotation arguments FIRST to remove braces inside annotation args
// like @DeleteMapping("/{id}") → @DeleteMapping, so the body-brace search
// doesn't get confused by braces in annotation strings.
```

**`buildClassSignature` did it in the opposite order** — brace cut first, `stripAnnotationArgs` afterwards, too late. One sibling documented the hazard; the other reproduced it.

## Measured before changing anything

| source | before | after |
|---|---|---|
| `@Table(name = "{weird}") class Foo` | **`@Table`** | `@Table class Foo` |
| `@Table(name = "{a{b}c}") class NestedBrace` | **`@Table`** | `@Table class NestedBrace` |
| `@Table(name = "{w}") class Wrapped extends Base implements Marker` | **`@Table`** | `@Table class Wrapped extends Base implements Marker` |
| `@Table(name = "plain") class Plainish` | `@Table class Plainish` | unchanged |
| `class Bare` | `class Bare` | unchanged |

**The class *name* is gone, not merely a trailing clause** — the signature collapses to the annotation. #7124 described this from source; it is now measured.

Line numbers had moved from the issue (`2783920c2` reworked this file): `buildMethodSignature` 2301–2318, `buildClassSignature` 2322–2336, `stripAnnotationArgs` 2501.

The fix is the sibling's order — `stripAnnotationArgs` to the top, brace cut after. **9 insertions, 2 deletions, no new logic.**

**Nested braces are handled, and for a reason worth recording:** `stripAnnotationArgs` balances **parentheses**, not braces, so brace nesting depth inside the argument list is irrelevant. Asserted as its own row rather than assumed.

## The pair audit found the sibling was ungraded too

Swapping `buildClassSignature`'s order back is **DEAD, 1 `--- FAIL`** — my class fixture, and under it exactly the two brace rows redden while `Bar` (one character from `Foo`), `Plain`, `Box<T>`, the wrapped `extends`/`implements` row, and all three #7091 wrapped tests stay green. So the fixture grades **the order**, not the presence of a signature.

Then the audit: **swapping `buildMethodSignature`'s order back was ALIVE.** The sibling whose comment documents the hazard — the one that was *correct* — had its correctness **graded by nothing**. It was right by comment, not by test, and any future edit could have reversed it silently.

Added a method fixture (`Repo.java`, `@DeleteMapping("/items/{id}")`) — test-only, that builder is untouched — and **re-scored both from scratch**: both now DEAD at 1 `--- FAIL` each, killed by disjoint tests.

That is the whole value of scoring every occurrence rather than a representative site.

## Verification

`javac` 25.0.3 on `Foo.java`, `Modified.java`, `Repo.java` under the names their declarations require — **zero diagnostics**. Annotation types are declared in-fixture, so no Spring/JPA classpath is needed; `public` requires its own compilation unit, which is why the modifier row is a separate file.

All `-count=1`, exit code **and** `--- FAIL` count checked: `./internal/extractors/java/` 0/0 · `./internal/extractors/… ./internal/resolve/ ./internal/links/ ./internal/docgen/` 0/0 · `./cmd/grafel/` 0/0 with fresh `GRAFEL_HOME`+`GRAFEL_DAEMON_ROOT` · `node-type-gate` OK and `-sites` 0 · `gofmt -l` empty · `go vet` 0.

The digest did **not** move. Expected, and **not** coverage — `entityTupleKey` does not hash `Signature`. Reported only because a *moved* digest would have meant something else changed.

## Deliberately not done

- **The structural read.** `class_declaration` exposes `name`, `type_parameters`, `superclass`, `interfaces` and `body` as named fields, so the raw-span text surgery could go entirely, taking #7124's unanchored-strip item with it. **This site is a good candidate** — but it shares a builder with `buildMethodSignature`, so doing both together (#7124 item 2) is cheaper than doing this one alone, and the diff was kept narrow rather than widened without checking back.
- **A latent defect that remains:** a `)` inside an annotation string literal still defeats `stripAnnotationArgs`. Same family, not fixed here.
- The reorder **narrows** #7124's unanchored-strip exposure at this site as a side effect, since the arguments are gone before the strip runs — **not graded**, so not claimed.

## Not claimed

**Incidence is unmeasured.** No corpus scan of how many real classes carry a `{` inside a class-level annotation literal, so the blast radius is unquantified.


cajasmota:
**Merge gate: independent review APPROVE with the pair audit reproduced from scratch, my own mutant DEAD, CI green on all 8 workflows. Two inaccuracies in the PR body, corrected below before they reach the squash message.**

## Independent review — APPROVE, and it re-derived the headline instead of accepting it

Both order-swap mutants scored in their own checkout, one occurrence each, killers **disjoint**:

| mutant | `--- FAIL` |
|---|---|
| `buildClassSignature` order swapped back | **1** (`TestJava7124_ClassSignature…`) |
| `buildMethodSignature` order swapped back | **1** (`TestJava7124_MethodSignature…`) |

Then the part I care about most: they **held the new test file out and re-scored both**. Both **ALIVE, 0 `--- FAIL`** — so before this PR *neither* site was graded. That confirms the pair-audit claim by measurement rather than by inheriting it, which is the right way to treat a PR's own headline.

They also checked the fixture grades the **order** and not merely the presence of a signature: under the class mutant exactly `Foo` and `NestedBrace` redden, while `Bar`/`Plain`/`Box<T>`/`Extending`/`Modified` stay green. And they compiled all three fixtures — `javac` exit 0, zero diagnostics — so the fixture is real Java, not something that merely parses.

## My own mutant: DT-1, the permissive direction

Nobody scored what happens if the brace cut is removed **entirely** — the signature then runs into the class body. That is the direction every serious defect this milestone has come from, so it is the one I took.

`buildClassSignature`: brace cut deleted, `stripAnnotationArgs` left in place. `go vet` 0, exact match count 1, `-count=1`.

**DEAD — 3 `--- FAIL`:**

```
--- FAIL: TestJava7091_WrappedClassSignatureIsCollapsed
--- FAIL: TestJava7124_ClassSignatureStripsAnnotationArgsBeforeBraceCut
--- FAIL: TestJava7124_ClassSignatureStillStripsVisibilityAfterReorder
```

Worth noting the killer set spans this PR's tests **and** a pre-existing one (#7091's): the brace cut's existence was already graded, and this PR's reorder did not weaken that. Restored by `cp`, shasum `e0e323df…`, `git diff HEAD` empty.

## Two corrections to the PR body

**1. The body tables a row the fixture does not contain.** It lists a `Wrapped` case carrying `{w}`, but the committed `Extending` row uses `"kin"` — **no brace**. So the body claims coverage of brace-in-literal × wrapped `extends`/`implements`, and the fixture does not have it. That combination is genuinely unobserved. This is the recurring shape here in a new place: not a fixture whose *name* overclaims, but a PR body that overclaims against the fixture it describes. The squash body is being written to state the real coverage.

**2. One remainder is undisclosed.** A `{` in a comment **between** the annotation and the keyword is the same family and is not in the body's disclosed-remainders list:

```
@Table(name="w") /* { */ class Cmt   →   "@Table /*"     (name lost)
```

The `)`-in-literal remainder *is* disclosed, and reproduces (`"@Tableb\") class Paren"`).

Both are coverage/disclosure corrections, not defects in the fix, so they do not block — but they go in the message rather than getting discovered later.

## Good news on two shapes I expected to be at risk

`{@link …}` in a javadoc comment is **never** at risk: tree-sitter puts a preceding doc comment *outside* the node span. That is the highest-incidence brace-in-text shape in real Java, and it cannot reach this code path at all. Same for a `'{'` char literal, plus interface/enum/record/nested-member/type-parameter-annotation — all correct post-fix.

## Sibling finding → #7133

The reviewer checked whether anything else shares the pattern. Java's own file is clean (`buildFieldSignature` reads the parse tree after #7122; `buildAnnotationElementSignature` has no brace cut). But **C# and Kotlin have the same inverted pair**, and C# has the same documented-sibling shape — `csharp.go:1805` strips attribute args first *with a comment naming the hazard*, `csharp.go:1825` cuts at the brace first. Both are worse than the java case:

```
[Route("api/{version}/items")] public class ItemsController  →  Signature ""
@Table(name = "{weird}") class Foo                           →  Signature ""
```

An **empty** signature rather than a truncated one, on `[Route("api/{version}/...")]` — the single most common attribute shape in ASP.NET. Filed as **#7133** with the requirement that both halves of each pair get scored, which is the lesson this PR earned.

## CI

All 8 workflows green on `52341ebc6`. `windows installers / install.bat (pinned)` failed first time on `curl: (35) Recv failure: Connection was reset` fetching the **v0.2.1 release asset** — an external download, and this diff is two files in the java extractor. Diagnosed before rerunning; rerun of failed jobs green. New known-disjoint failure class: the pinned-installer legs pull a real GitHub release asset.

Reviewer's own verification: `./internal/extractors/java/` 0/0 · `./internal/extractors/... ./internal/resolve/... ./internal/links/...` 0/0 · **`./cmd/grafel/` 0/0 (191s, fresh `GRAFEL_HOME` and `GRAFEL_DAEMON_ROOT`)** · `gofmt -l` empty · `go vet` 0. The PR correctly states the unmoved digest is **not** coverage here, since `entityTupleKey` omits `Signature`, and does not lean on it.

Merging. Closes #7124 item 1; item 2 remains, item 3 is a measured negative (see the issue), and #7133 carries the cross-language siblings.

Issue #6875, opened 2026-09-05
DIAGNOSED: the Windows rename denial is a rare cohort phase-locking tail in the test's own 8 writers — the budget is not undersized and the test does not overclaim


cajasmota:
`TestWriteFile_ConcurrentWritersSameDestination` (`internal/atomicfile`) failed on the Windows leg of an unrelated PR (#6873, which touches only `internal/extractor` test files and comments):

```
atomicfile_test.go:256: iteration 12 writer 4:
  rename …\.shared.json.tmp-1583407739 -> …\shared.json:
  still failing after 41 attempts over 200ms:
  rename …: Access is denied.
```

Rerunning the leg cleared it, so it is a flake rather than a regression. Filing because the *shape* is a real robustness question, not a scheduling accident to be re-run away.

## What the failure says

The retry budget — **41 attempts over 200ms** — was exhausted. On Windows a rename over an existing destination fails with `ERROR_ACCESS_DENIED` while another process or thread holds a handle to the destination, which under N concurrent writers to one path is not rare. So this is not "the machine hiccuped"; it is the retry budget being too small for the contention the test itself creates.

Two directions, and they need different answers:

1. **If the budget is wrong**, the production path has the same budget, and a real caller with concurrent writers hits the same wall — a genuine defect that happens to surface first in a test.
2. **If the test's concurrency exceeds what the API promises**, the test is asserting something `atomicfile` does not undertake to do, and the fix is to the test, not the retry loop.

Nothing in the failure distinguishes these, and that is the point of filing rather than re-running.

## What NOT to do

**Do not just raise the retry count.** A larger budget makes the flake rarer without answering which of the two cases above is true, and a rarer flake is worse than a frequent one — it gets re-run away by whoever meets it next, exactly as I did here, with no one learning anything.

## Suggested shape — measure first

1. **Establish the real contention profile on Windows.** How many attempts does a rename actually need under this test's concurrency, distributionally? If the 99th percentile sits near 41, the budget is simply undersized and the number should be derived rather than guessed. Point the budget at an absurd value once and read the printed distribution, rather than picking a new constant by eye.
2. Decide whether `atomicfile.WriteFile` **promises** safety under concurrent writers to one destination. If it does, the retry loop is the contract and needs to be sized to it. If it does not, say so in the doc and make the test model the supported usage.
[...]

Issue #6512, opened 2026-08-21
internal/daemon TempDir cleanup fails on Windows CI — TWO tests, TWO different Win32 errors: 'not empty' (delete-pending) and 'used by another process' (real handle hold)


cajasmota:
`internal/daemon` intermittently fails on `windows-latest` during test cleanup, not during the test itself:

```
--- FAIL: TestResolveMemLimitMB_DefaultClamped (0.01s)
    testing.go:1464: TempDir RemoveAll cleanup: unlinkat
    C:\Users\RUNNER~1\AppData\Local\Temp\TestResolveMemLimitMB_DefaultClamped797493729\001:
    The directory is not empty.
```

The test's assertions all pass. `t.TempDir()`'s deferred `RemoveAll` is what fails.

## What it means, and why it is probably a real leak

On Windows an **open handle blocks deletion**. "The directory is not empty" after `RemoveAll` means something inside was still open when cleanup ran — so the failure is evidence that a file handle outlives the test that opened it. POSIX allows unlinking an open file, which is exactly why the same leak is invisible on the Linux and macOS legs.

`TestResolveMemLimitMB_DefaultClamped` is a memory-limit resolver test and unlikely to be the owner. The surrounding log lines show `orphanroot` sweeps running concurrently against other tests' temp dirs, so the likelier owner is a background goroutine — a sweeper, watcher, or daemon plane — still holding something under a sibling test's directory when this one tears down.

## Why it is worth fixing rather than tolerating

It has intercepted an unrelated PR at least once this cycle, on a package that PR did not touch. Each occurrence costs a CI re-run and, worse, **trains the reader to dismiss red Windows legs as noise** — which is precisely how the genuine deadlock in #6493 nearly went unrecorded.

Two intermittent Windows-only failures now exist in this repo (this and #6493). Both fail in a way that looks like flake and is not.

## What a fix needs to establish

1. **Which goroutine holds the handle.** The failing test is probably a bystander; find the owner rather than adding a retry to this one.
2. **Is there a missing `Close` on a shutdown path?** If a daemon plane or sweeper starts in one test and is not fully drained before `TestMain` moves on, the handle can outlive the owning test entirely.
3. **Does `orphanroot`'s sweeper honour test lifetimes?** The concurrent sweep lines in the log make it a natural suspect — it walks other tests' temp roots by design.

[...]

PR #7129, opened 2026-09-17
feat(dashboard): add repository topology and bounded graph loading

### LINKED ISSUE #7127: Add repository-level topology for large multi-repository groups
## Problem

Entity-level graphs become difficult to load and navigate for large multi-repository groups. Streaming improves progress visibility, but it does not bound graph construction, transfer size, browser memory, or layout work.

## Proposed behavior

- Add a dedicated repository-level topology aggregated from persisted cross-repository links.
- - Support channel, evidence, repository, focus, direction, depth, search, and shortest-path filters.
- - - Provide bounded, paginated relationship evidence for repository edges.
- - - - Keep topology filter state in the URL for stable deep links.
- - - - - Apply finite node and edge caps to entity Graph LoD requests, including SSE, with explicit truncation metadata.
- - - - - - Avoid starting entity Graph requests for repository topology and module-overview surfaces.
## Acceptance criteria

- Repository topology scales with repository and cross-repository relationship counts rather than total entity count.
- - Existing graph and topology routes remain available.
- - - Filtered and truncated states are visible rather than silently omitting data.
- - - - Large graph requests remain bounded on both streaming and non-streaming endpoints.
- - - - - Backend, frontend, and performance regression tests cover the new behavior.

0nlyl:
## Summary

- Add a dedicated repository-level topology API and dashboard surface aggregated from persisted cross-repository links.
- Add channel, evidence, repository, focus, direction, depth, search, shortest-path, and paginated edge-detail workflows with URL-backed filters.
- Bound entity Graph node and edge output for every LoD, including SSE, and expose explicit truncation metadata.
- Avoid entity Graph requests when repository topology or module overview is active.
- Add backend, frontend, regression, and bounded-scale performance coverage.

## Closes / Refs

- Closes #7127

## Testing

- [ ] All tests pass: `go test ./...` — the complete suite includes environment-dependent tests that were not safe to run in this Windows session.
- [x] Relevant fixtures verified
- [x] No regression in cross-language parity

Additional verification:

- `go test ./internal/dashboard -skip 'TestV2FsList_DoesNotFollowProtectedHomeSymlink' -count=1`
- `go vet ./...`
- `npm test` — 36 files / 334 tests passed
- `npm run lint`
- `npm run build`
- `go run ./cmd/verify-dashboard -root .`

## Notes

- Repository topology responses are bounded and never embed entity-level graph arrays.
- Edge evidence is paginated and filtering happens server-side.
- Graph, stream, Dubbo, and message-topology routes have explicit regression coverage; the remaining established routes continue to be covered by their existing package tests.
- The branch contains only generic fixtures and documentation; no private repository names, paths, screenshots, or production data are included.


cajasmota:
Thank you — and thank you for filing #7127 alongside it, so the motivation is on the record separately from the change.

Reviewing now. Three things in this that I want to call out as the right instincts, because they are the parts that usually get skipped:

**Bounded output with explicit truncation metadata, for every LoD including SSE.** A graph endpoint that silently returns "some" of the data is the failure mode that is impossible to debug later — the consumer cannot tell a small graph from a truncated one. Saying *how* it was bounded is what makes it usable.

**Server-side filtering and paginated edge evidence.** Same reasoning: the cost stays where the data is.

**Avoiding entity-graph requests when repository topology or module overview is active.** That is the kind of thing that only shows up under real multi-repository load, which suggests you hit it rather than anticipated it.

I also appreciate that you left the `go test ./...` box unchecked and said why, rather than ticking it. **That is the right call** — the full suite has environment-dependent tests, some of which are genuinely unsafe outside a prepared state directory, and CI covers those legs on all three platforms anyway. An unchecked box with a reason is worth more to me than a checked one I have to verify.

## What happens next

CI is running (15 checks; the Windows test leg is the last one outstanding). Note for future pushes on a fork PR: **CI re-parks on every push here**, so if you push again and see no check-runs at all, that is `action_required` waiting on a maintainer rather than a green — ping the PR and I will approve the runs.

I am putting this through an independent adversarial review, which is the same treatment every change gets in this repo including my own — it is not a comment on your work. A 55-file change touching both the Go API surface and the frontend gets read carefully on a few specific axes:

- whether the bounds are **observable** (a test that would notice if truncation silently stopped happening, not just that it happens today);
- whether the truncation metadata is asserted rather than merely emitted;
- the interaction with the existing graph/Dubbo/message-topology/links/paths routes you list as unaffected — specifically whether anything **grades** that they are unaffected;
- and the usual house rule that a claim in a comment or a doc needs something observing it.

If review turns up anything I will bring it back here with the demonstration attached — a scored mutant or a reproduction, not an opinion. Anything I ask you to change, I will say exactly why and what evidence I have.

One question, purely to sequence my own review rather than a request for work: is the bounded-scale performance coverage you mention exercising a **synthetic** large group, or a real one? That changes where I would look hardest for the truncation boundary.

cajasmota:
Review is back. **The feature works and the code is good** — everything you said was green is green, plus the one test you had to skip, which passes here. The SSE bound in particular is done properly and I want to be specific about why below.

What I am asking for is narrow, and it is on the one axis the PR exists for: **the bounds and the truncation flags are almost entirely unobserved.** 11 mutants survived. I confirmed the headline myself before writing this, because I am not willing to tell you your tests do not grade your feature on someone else's word.

## The core finding

I mutated `repositoryTopologyMaxEdges` in both directions and scored each against the full `./internal/dashboard/` suite:

| mutant | `--- FAIL` | verdict |
|---|---|---|
| `2_000` → `1` | **6** | DEAD |
| `2_000` → `2_000_000` | **0** | **ALIVE** (41.5s, `ok`) |

That asymmetry is the whole thing. The suite notices if the cap drops **too much**, and cannot notice if it stops dropping **at all**. A cap that silently stops capping is invisible, which is the exact failure mode your PR body identifies as worth fixing — "the consumer cannot tell a small graph from a truncated one."

The structural cause is not a missing assertion, it is fixture size: the largest topology fixture in the diff is **5 repos / 30 links** (`repository_topology_filter_test.go:121`, `handlers_repository_topology_test.go:69`). Nothing anywhere approaches 500 nodes or 2000 edges, so as far as the suite is concerned both caps and the `truncated` flag are unreachable code.

Same result on the node cap (`500` → `500_000`: ALIVE), on `truncated := preLimitEdges > len(edges)` forced to `false` (ALIVE), and on deleting the node-cap arm's `truncated = true` (ALIVE).

## The entity-graph metadata: 7 mutants, 7 ALIVE

`EdgeTruncated` → constant `false`, → constant `true`, and each arm alone: all ALIVE, 1 occurrence each. `totalEdgeCount++` deleted in **both** branches (thinned and non-thinned): both ALIVE. So `total_edge_count` can report `0` and `edge_truncated` can report `false` while edges are genuinely being dropped, and the package stays green.

`API_V2.md` promises these fields "distinguish returned cardinality from source cardinality" and that "clients must not present a truncated result as complete". Nothing observes either sentence.

One thing worth flagging because it *looks* like coverage: `TestV2GraphStreamLoDMatchesFullPayload` compares stream meta against the full payload, but **both sides come from the same `buildV2GraphWithLimits` call**, so a mutation moves both together and the comparison still holds. It is a shared premise, not a second observation. That is a subtle trap and an easy one to walk into.

## What is graded, and graded well

I want this on the record because it is most of the PR: the node cap (threshold ×1000 → 6 FAILs), the LoD→cap wiring on the non-streaming path (6 FAILs), **the SSE path (5 FAILs each on two mutants)**, the edge cap at unit level (2 FAILs), and `NodeTruncated` (1 FAIL).

On SSE specifically: `v2_graph_stream.go:293` calls the identical `lodLimits` + `buildV2GraphWithLimits`, so it is genuinely the *same* bound rather than a parallel implementation, and it has its own test that kills its own mutants. Streaming paths usually get the limit re-applied at a different layer or after the first chunk. You avoided that.

## Frontend: the flag is ungraded end to end, with one graded link in the middle

`use-graph.ts` hard-coded to drop both backend flags (`nodeTruncated: false, edgeTruncated: false`): **332/332 still pass.** `graph.tsx:589` truncation banner forced off: **332/332 still pass.** Only the SSE reducer's pass-through is graded (`graph-stream-reducer.test.ts`). So backend→wire→hook→banner has one observed hop out of three.

## A memory regression, in a memory-bounding PR

`collectCappedGraphEdges` does `best := make(rankedGraphEdgeHeap, 0, cap)` — eager capacity proportional to the **cap**, not to the data. `rankedGraphEdge` is 64 bytes:

```
BenchmarkZZProbeTinyGraphFullLod-12   200   807220 ns/op   16007676 B/op   9 allocs/op
```

That is a **2-node, 1-edge graph at `?lod=full`: 16 MB and 0.8 ms per request**, per concurrent request. Default LoD costs 1.5 MB, `high` 7.7 MB. `make(rankedGraphEdgeHeap, 0, min(cap, 4096))` and letting it grow fixes it in one line.

## Smaller things

**The route-order comment is factually wrong.** `server.go:827` says the topology routes are *"registered before graph routes so the dedicated bounded payload never falls through to entity graph work."* Go's `ServeMux` matches by specificity, not registration order, and these patterns cannot conflict anyway. Demonstrated by moving both registrations *after* every graph route: `go vet` clean, `-run RepositoryTopology` still `ok`. The behaviour is fine; the stated mechanism does not exist. House rule here is that a comment claiming a mechanism needs that mechanism to be real, because the next person will rely on it.

**One perf test asserts the opposite of its name.** `webui-v2/src/lib/repository-topology-performance.test.ts` feeds 500 nodes / 2000 edges and asserts `toHaveLength(500)` / `toHaveLength(2_000)` — i.e. that *nothing was dropped* — with no timing and no cap assertion, under a "performance / bounded transformations" label. It is a determinism test, and it is also the slowest test in the suite at 1984 ms. Your Go side (`bench_repository_topology_test.go`, 100k links / 500 repos / 1M internal relationships) is the reasonable one; both are synthetic, which answers the question I asked earlier and is fine.

**Dead surface:** `capGraphEdges` has no production caller. `buildV2GraphWithNodeCap` is now test-only too and hard-codes `edgeCap: 0`, so `v2_graph_lod_compact_test.go` exercises an unbounded-edge path production never takes.

## Two things you got right that I want to name

**Privacy: PASS.** Every added line scanned for absolute paths, `C:\`, `/home/`, emails, hostnames, RFC1918 addresses and internal names — nothing. Fixtures are `repo-000`/`client`/`rule`/`a`–`e`. No binaries, no images. Your statement was accurate, and this is the one check where a miss is unrecoverable after merge.

**The skipped test was the right call.** `TestV2FsList_DoesNotFollowProtectedHomeSymlink` is pre-existing, untouched by your diff (`v2_fs_test.go:147`), and calls `os.Symlink`, which needs Developer Mode or admin on Windows. It passes here on macOS in 0.01s. Skipping it with a stated reason was correct, and so was leaving `go test ./...` unchecked.

## The ask

1. One fixture above each topology cap (>500 repos or >2000 edges) asserting the returned cardinality **and** `truncated`.
2. Assert `edge_truncated` / `total_edge_count` on a payload where edges actually get dropped — **separately per branch**, since the node-thinned and edge-capped paths compute them differently.
3. One test that is not the reducer for the frontend flag→banner path.
4. `min(cap, …)` on the heap preallocation.
5. Either correct the route-order comment or drop it; and either delete `capGraphEdges` or give it a caller.

(1) and (2) are the ones I would not merge without — they are what makes the rest of the PR checkable later. (3)–(5) I am happy to take as a follow-up if you would rather land the backend first; say which you prefer.

Also: the seven-route no-regression claim is covered for **4 of 7** by `TestRepositoryTopologyRegressionPreservesEstablishedRoutes` (graph, stream, dubbo, topology). Links, paths, settings and MCP rest on pre-existing package tests, which are green — so nothing is broken, the claim in the body is just broader than the new test. Narrowing the sentence is enough; no new tests needed. Same for "avoid entity-graph requests when topology **or** module overview is active": `graphRequestMode` models the module-overview half (graded, 2 tests), and the topology half is true by construction since it is a separate route.

Full disclosure on what could not be checked: every fixture is synthetic, so the review graded the code rather than production shapes, and real-corpus behaviour at the truncation boundary is unverified.

No rush on any of this, and push whenever — remember CI re-parks on a fork push here, so ping me if you see zero check-runs and I will approve the runs.

cajasmota:
**Quick note so you don't go looking: the red `test (windows-latest)` leg is not your change.** I should have flagged this when it went red rather than letting you find it.

The run's only failure is `internal/atomicfile`:

```
--- FAIL: TestWriteFile_ConcurrentWritersSameDestination (3.26s)
    atomicfile_test.go:256: iteration 36 writer 5: rename ...\.shared.json.tmp-2464819832
      -> ...\shared.json: still failing after 41 attempts over 200ms: Access is denied.
```

That is **#6875**, a long-standing Windows-only flake in our atomic-file writer, fully diagnosed and unrelated to anything here. The cause is the test's own eight concurrent writers contending for one destination — at that concurrency the runner denies 31–47% of renames, while a single-writer control did 160,000 writes with zero denials. It has nothing to do with the dashboard.

I checked before rerunning, rather than assuming, because that issue exists precisely because the check gets skipped: your diff is 55 files, **zero of them under `internal/atomicfile`**, and the only Go package it touches is `internal/dashboard` — which **passed on that same Windows leg in 112.9 seconds**. `test (ubuntu-latest)` passed, the python ratchet leg passed, and the other six workflows are all green. I have rerun the failed job and recorded the instance on #6875.

One thing your PR did change for me, though: every previous instance of this flake landed on one of my own PRs, where I know what it is at a glance. Landing on a contributor's PR makes it read as "my change broke Windows", with 55 files to go audit. That is a cost paid in confusion rather than CI minutes, and it has moved #6875 up my list — the diagnosis is complete, so keeping the test red in everyone's path buys no further information. I have flagged it there as an open decision rather than quietly changing the test, since what to do about the API's concurrency contract is a real call and not flake housekeeping.

No action needed from you on this. The review feedback in my previous comment is still the only thing waiting on you, and there is no rush on it.

cajasmota:
**CI is fully green now** — the rerun of `test (windows-latest)` passed, no code change, which is the confirmation that the failure was #6875 and not your branch. All 15 checks pass on `e0a45b65a`.

So nothing is outstanding on CI. The only thing waiting is the review feedback in my earlier comment, and specifically the two items I said I would not merge without:

1. a fixture above each topology cap (>500 repos or >2000 edges) asserting the returned cardinality **and** `truncated`;
2. `edge_truncated` / `total_edge_count` asserted on a payload where edges actually get dropped, separately per branch.

The other three (the frontend flag→banner test, `min(cap, …)` on the heap preallocation, and the route-order comment plus `capGraphEdges`) I am happy to take as a follow-up if you would rather land the backend first — just say which you prefer and I will treat that as the plan.

Take whatever time you need.

0nlyl:
Implemented all five review items in `d171a295c`:

- Added a 501-repository chain fixture that asserts the 500-node cap, returned edge cardinality, limits, and `truncated`.
- Added 2,001 unique directed repository pairs across 64 repositories that assert the 2,000 aggregated-edge cap and `truncated`.
- Added separate direct-limit tests for node-thinning and edge-cap-only paths. Both use payloads where edges are actually dropped and assert `total_edge_count`, `edge_truncated`, `node_truncated`, and served edge cardinality.
- Added a frontend wire-normalization → truncation-banner test, including the complete-graph no-banner case.
- Bounded heap preallocation with `min(cap, 4096)`, removed the unused `capGraphEdges` wrapper, and corrected the ServeMux registration-order comments.

I also narrowed the PR body’s route-regression statement to distinguish the explicit four-route regression coverage from the existing package tests.

Local verification:

- `go test ./internal/dashboard -skip 'TestV2FsList_DoesNotFollowProtectedHomeSymlink' -count=1`
- `go vet ./...`
- `npm test` — 36 files / 334 tests
- `npm run lint`
- `npm run build`
- `go run ./cmd/verify-dashboard -root .`

All added fixtures remain synthetic and generic.

cajasmota:
Thank you — that is all five items, including the two I said I would not merge without. Independent re-review is running now, and I will bring back the mutant verdicts either way.

**On the red `test (ubuntu-latest)` leg: it is in a package your diff does not touch, and I am establishing whether it is a flake before rerunning.** Flagging it now rather than after, so you are not left guessing.

The run's only failure is `internal/daemon/watch`:

```
--- FAIL: TestRemoveRepoDoesNotCallTheBackendUnderTheLock
    removerepo_6309_test.go:36: Remove(...) was called while holding w.mu
    removerepo_6309_test.go:77: RemoveRepo unwatched 1 directories, want 2
```

Your Go diff is `internal/dashboard` only — zero files under `internal/daemon` — and `test (windows-latest)` and the python ratchet leg both passed, as did the other six workflows. So I am confident it is not your change.

What I am **not** yet sure of is whether it is a flake or a real regression on `main`, and I would rather tell you the truth than the convenient version. This one is different from the two Windows flakes we have already hit on this PR (#6875's rename denial and #6512's TempDir cleanup): those are file-handle-lifetime problems that surface as cleanup or retry failures, whereas this is a **synchronisation assertion on Linux** — it claims a mutex was held across a backend call. That is the kind of failure that can be a genuine race rather than infrastructure noise, and it has no existing issue and no `main` baseline, because our `test` workflow does not run on a push to `main`.

So far: the test passes 3/3 in isolation on `main` at `c4af71f93`. I am now running the whole package, because CI runs it under load and this failure shape is order- and timing-sensitive. If it reproduces on `main` without your diff, it is a pre-existing defect and gets its own issue; if it does not, it is a flake and I will record the instance and rerun. Either way it is not a blocker for you and needs nothing from you.

One small thing for the re-review, purely so I can check the right thing: you ran `go test ./internal/dashboard -skip 'TestV2FsList_DoesNotFollowProtectedHomeSymlink'`. That is still the right call on your platform — it needs `os.Symlink`, which requires Developer Mode or admin on Windows — and CI runs it unskipped on Linux and macOS. I have asked the reviewer to run it unskipped too, so the skip is covered rather than assumed.

cajasmota:
**Diagnosed, and it is a false positive in the test, not a defect anywhere near your change. Reran; nothing needed from you.** Filed as **#7145**.

The detector is unsound. `TestRemoveRepoDoesNotCallTheBackendUnderTheLock` decides whether the backend was called while holding `w.mu` by calling **`w.mu.TryLock()`**, under this comment:

> `// TryLock is exact here: the only contended holder in this test is the caller itself.`

`TryLock` fails if **any** goroutine holds the mutex. That test's helper builds the watcher through `NewWatcherConfig`, which starts five background goroutines, and two of them take `w.mu` while `RemoveRepo` is running:

| goroutine → callee | `w.mu` acquisitions |
|---|---|
| `loop` | 1 |
| `reconcileLoop` → `reconcileOnce` → `nextReconcileBatch` | 1 |
| `reconcileLoop` → `reconcileOnce` → `scanDir` | **5** |

And the suite's own helper says the sweep is left running on purpose: *"The reconcile sweep is deliberately NOT pinned… every one of them runs it, at its production cadence."* So when `reconcileOnce` happens to be inside `scanDir`, `TryLock` returns false and the test reports that **the caller** held the lock. It did not.

**The second line of the failure is the same bug, not extra evidence.** On the false positive the check calls `t.Errorf` and returns early *without recording the call*, so the later count assertion sees a short list:

```
removerepo_6309_test.go:77: RemoveRepo unwatched 1 directories, want 2 —
  the backend calls were dropped rather than moved out of the critical section
```

Nothing was dropped; the recording was skipped. That message points a reader at a production defect that did not happen, which is most of why this took a careful look rather than a quick one.

On `main` at `c4af71f93` the test passes 3/3 in isolation and the full package passes (168.5s, 0 FAIL) — consistent with a scheduling-dependent false positive rather than anything deterministic.

**What this means for your PR: nothing.** Your Go diff is `internal/dashboard` only, zero files under `internal/daemon`, and every other leg was green. I have rerun the failed job.

For the record, this is the third distinct pre-existing test problem your PR has surfaced — #6875's Windows rename denial, #6512's TempDir cleanup, and now #7145. That is not bad luck on your part; a 55-file PR runs the full suite more times than most changes do, and each of these has been costing everyone a diagnosis pass. You have effectively been paying for our flakes, and #7145 is the one I am most glad to have found, because it was reporting a **false** production defect with a confident message.

Your five items are through independent re-review now — I will bring back the mutant verdicts either way, including whether the round-1 mutants that were ALIVE are now dead.

cajasmota:
**Both merge-blockers are genuinely fixed, independently verified by re-scoring the round-1 mutants. Merging.**

## The caps: 4/4 ALIVE → 4/4 DEAD

| mutant | round 1 | now |
|---|---|---|
| `repositoryTopologyMaxEdges` 2_000 → 2_000_000 | ALIVE | **DEAD** |
| edge-cap guard disabled | ALIVE | **DEAD** |
| node-cap guard disabled | ALIVE | **DEAD** |
| `truncated := preLimitEdges > len(edges)` → `false` | ALIVE | **DEAD** |
| node-cap arm's `truncated = true` deleted | ALIVE | **DEAD** |

And the fixtures do what their names claim: the tests assert `PreLimitNodes == 501` and `PreLimitEdges == 2001` **after** aggregation on `source+target+channel`, so the 2,001 pairs provably do not collapse to fewer. Both run in 0.01s, which is the part I did not expect — crossing a 2,000-edge boundary for free.

## The metadata: 7/7 ALIVE → 6/8 DEAD

`NodeTruncated` both directions, `EdgeTruncated → false`, the `edgeCapTruncated` arm alone, and `totalEdgeCount++` deleted in **each** branch separately — all dead. The two direct-limit tests genuinely separate the thinned and edge-capped paths, which was the specific thing I asked for.

## My own mutants: the aggregation key is built at two independent sites

`repositoryEdgeKey{Source, Target, Channel}` is constructed separately in `repository_topology_build.go:69` and `repository_topology_filter.go:49`. That is the twinned-key shape that bit us elsewhere this week, so I scored **both**, not a representative one:

- `Channel` dropped at the **filter** site only → **DEAD, 4 `--- FAIL`**
- `Channel` dropped at the **build** site only → **DEAD, 2 `--- FAIL`** (`…AggregatesPairAndChannel`, `…OrdersEdgesAndBoundsSamples`)

Disjoint killers, both directions covered. The two sites' agreement is graded.

## The allocation fix is real, measured

2-node/1-edge graph at `?lod=full`: **263,216 B/op** with the fix, **16,008,240 B/op** with only that line reverted. −98.4% bytes, ~97× faster, matching round 1's 16,007,676 figure. A one-line change that moves the number by two orders of magnitude.

Privacy re-checked on the generated fixtures — `repo-0000` / `dubbo:Contract0000`, nothing but `@/…` import aliases. `TestV2FsList_DoesNotFollowProtectedHomeSymlink` run **unskipped** and passing. `npm test` 334/334, lint and build clean. Go diff is `internal/dashboard` only for the whole PR, so no extractor or assembly exposure.

## Three residuals — filed, not held against you

I said in round 1 that items 1 and 2 were the ones I would not merge without, and that the rest could be follow-up. That still stands, so these go to issues rather than another round:

1. **`EdgeTruncated` → `true` is ALIVE** — nothing asserts `edge_truncated == false` on a complete graph. `NodeTruncated` has that negative; edges do not. One assertion.
2. **The route→banner hop is still unpinned.** Dropping both flags in `normalizeGraphPayload` is DEAD, but `<GraphTruncationBanner data={undefined} />` in `routes/graph.tsx` leaves 334/334 green, because the new test renders the component directly and nothing renders `GraphScreen`. Extracting the banner made it testable and moved the wiring hop rather than closing it — a genuinely easy thing to miss.
3. **`buildV2GraphWithNodeCap` is now test-only** (both callers in `_test.go`) while its doc comment still reads as a production rationale.

## One thing worth knowing about your own fixtures

They size themselves from the constants (`repositoryTopologyMaxNodes+1`), which is why they are robust — but it means the `500 → 500_000` mutant generates 500,001 repositories and turns `repositoryFromPrefixedEntityID`'s scan into ~2.5e11 operations. It spun at 100% CPU for four minutes rather than failing, so that arm had to be scored at the cap-application site instead. Not a defect, and self-sizing is still the right call; just be aware the suite pins "one past whatever the constant says" rather than the numeric boundary itself.

Thank you for this — it is a large change that came back tighter each round, and you fixed the two things that mattered without argument.

cajasmota:
Coming back to correct something in my own review conduct on this PR, now that it is merged.

**I should not have asked you for those two fixtures.** In [this comment](https://github.com/cajasmota/grafel/pull/7129#issuecomment-5713241230) I named two items as things I would not merge without — a fixture above each topology cap asserting cardinality and `truncated`, and `edge_truncated` / `total_edge_count` asserted per branch on a payload where edges actually drop. Both were test-writing chores in a package the maintainers know well. Neither was a design decision. I could have written them in less time than it took to write the comment asking for them, and the right move was to push one commit onto your branch and tell you what I had added — not to hold a correct fix waiting on you.

Two things made it worse, and I want to name them rather than let them pass as politeness:

- I handed you the **planning** as well: *"just say which you prefer and I will treat that as the plan."* That reads as courteous and is really a second ask, because it made you responsible for sequencing *our* follow-up work and blocked the merge on a reply. Deciding that sequence was my job.
- I closed with *"take whatever time you need"* while the PR sat unmerged. Patience is not a substitute for doing the work.

The standing rule here, which I failed to apply: **when a contributor PR needs a change, it is the maintainers' work, not the contributor's.** A missing fixture or a missing negative control is ours by default — we are the ones holding the mutation-scoring bar, and requiring a contributor to satisfy an internal grading standard before a correct fix lands is holding a good contribution hostage to our own process. What matters for the merge is whether the production code is right.

None of this reflects on your work, which was good and which you then did anyway. For the record of what actually shipped: caps went 4/4 ALIVE to 4/4 DEAD and truncation metadata 7/7 ALIVE to 6/8 DEAD between rounds, the fixtures assert cardinality *after* aggregation so 501 nodes and 2,001 pairs provably do not collapse, and the heap fix measured 263,216 B/op against 16,008,240 on a two-node graph. Your authorship is preserved in `36adc6edb`.

The three residuals I had offered as follow-up are on us and are being done now, not asked of you — they are tracked as #7146, and two of the three already have PRs open (#7154 for the `edge_truncated` negative and the test-only `buildV2GraphWithNodeCap`, #7149 for the route to banner wiring hop). One of them turned out to be a good deal more interesting than the write-up suggested: the frontend extraction had *moved* the untested hop up a level rather than closing it, and pinning it surfaced that the two truncation guard arms only ever moved together, so neither was graded. That was our gap to find, not yours.

Thanks for the contribution, and sorry for the round-trip I put in front of it.

PR #7134, opened 2026-09-17
fix(#7131): F# let modifiers were captured as the entity name — and two `let inline` bindings collapsed into one entity

### LINKED ISSUE #7131: MEASURED: F# 'let inline' / 'let private' / 'let internal' are indexed under the MODIFIER as the entity name — 5 of 8 modifier combinations record the wrong Name, and same-indent duplicates collapse
## What

`letRE` in `internal/extractors/fsharp/extractor.go:62` accommodates exactly two `let` modifiers:

```go
`(?m)^([ \t]*)let(?:\s+rec)?(?:\s+mutable)?\s+([a-zA-Z_][a-zA-Z0-9_']*)\s*(?:<[^>]*>)?\s*(?:[^=\n]*)=`
```

F# allows more than two. For every modifier it does not know, the **modifier itself is captured as the entity name** and the real name is lost — so `let inline distance p q = ...` is indexed as an operation called **`inline`**.

This is not a signature-cosmetics problem. `Name` and `QualifiedName` are both wrong, so the entity cannot be found by name, cannot be a call target, and collides with every other `let inline` in the file — `letSeen` is keyed `indent + ":let:" + name`, so **two `let inline` bindings at the same indentation silently collapse into one entity.**

## Measured — enumerated, not hand-picked

Probe run against the real extractor (`runFSharp`), source `let <modifiers>target (a: int) = a`, reading the `SCOPE.Operation` entity's `Name`:

| modifiers | recorded name | |
|---|---|---|
| *(none)* | `target` | OK |
| `rec` | `target` | OK |
| `mutable` | `target` | OK |
| `inline` | **`inline`** | WRONG |
| `private` | **`private`** | WRONG |
| `internal` | **`internal`** | WRONG |
| `private inline` | **`private`** | WRONG |
| `rec inline` | **`inline`** | WRONG |

**5 of 8 combinations wrong.** `let private` and `let inline` are both idiomatic, high-frequency F# — `inline` is how F# does static-resolution generics, and `private` is the default instinct for a module-local helper.

Note `rec inline` is wrong even though `rec` alone is handled: the regex's `(?:\s+rec)?(?:\s+mutable)?` is a fixed *sequence* of two optional literals, so it cannot absorb a third token in any position.

## Why nothing caught it

- No fixture exercises any modifier beyond `rec`/`mutable`. This is the fixture-space enumeration gap again: the axis (modifier) was varied once and its neighbours left open.
- `Name` **is** in `entityTupleKey`, so `./cmd/grafel/`'s digest would move — but only if the mixed-language fixture contained F#, and it does not.

## Suggested shape

Replace the fixed two-slot sequence with a repeated modifier group, e.g. an alternation of the known modifiers repeated zero-or-more times (`(?:\s+(?:rec|mutable|inline|private|internal))*`), then capture the name. The modifier list is closed and short, so an allowlist is preferable to "any identifier that is not the last one".

**Grading requirement:** the fix must be graded **per modifier and per combination**, not by one `let inline` fixture. A single-modifier fixture would leave `private inline` open, which is exactly how this got here. Enumerate the combinations in a table-driven test.

Also worth checking in the same pass, because the same fixed-sequence pattern may appear there: F# `type`, `module` and `member` scanners, which have their own modifier sets (`abstract`, `static`, `override`, `default`, `val mutable`).

Found while grounding #7124 item 3.

cajasmota:
Closes #7131.

`letRE` accommodated modifiers as a fixed **sequence** of two optional literals,
`(?:\s+rec)?(?:\s+mutable)?`, which cannot absorb a third token in any position. Every other
modifier landed in the **name** capture, so `let inline distance p q = ...` was indexed as an
operation called `inline` — `Name` and `QualifiedName` both wrong, entity unfindable by name and
unusable as a call target. And because `letSeen` is keyed `indent + ":let:" + name`, the mis-naming
**lost entities**: two same-indent `let inline` bindings both keyed on `"inline"` and the second was
dropped.

The fix replaces the two-slot sequence with a repeated allowlist over the closed modifier set:

```go
`(?m)^([ \t]*)let(?:\s+(?:rec|mutable|inline|private|internal|public)\b)*` +
    `\s+([a-zA-Z_][a-zA-Z0-9_']*)\s*(?:<[^>]*>)?\s*(?:[^=\n]*)=`
```

An allowlist rather than "any word before the name" is deliberate: the latter names
`let add x y = x + y` after its last parameter (mutant M7 below demonstrates exactly that).

## Strict TDD — the red output, before the fix existed

`go test -count=1 -run TestLetModifiers ./internal/extractors/fsharp/` at the test-only commit:
**15 of 18 table rows red**, plus 3 of the 4 standalone tests — 19 `--- FAIL` lines. Abridged:

```
--- FAIL: TestLetModifiers_NameIsNeverTheModifier/inline
    no `let` SCOPE.Operation named "target" for "let inline target (a: int) = a"; let names = [inline]
    modifier "inline" recorded as the entity name for "let inline target (a: int) = a" (line 3)
--- FAIL: TestLetModifiers_NameIsNeverTheModifier/private
    no `let` SCOPE.Operation named "target" for "let private target (a: int) = a"; let names = [private]
--- FAIL: TestLetModifiers_NameIsNeverTheModifier/internal          ... let names = [internal]
--- FAIL: TestLetModifiers_NameIsNeverTheModifier/public            ... let names = [public]
--- FAIL: TestLetModifiers_NameIsNeverTheModifier/private_inline    ... let names = [private]
--- FAIL: TestLetModifiers_NameIsNeverTheModifier/inline_private    ... let names = [inline]
--- FAIL: TestLetModifiers_NameIsNeverTheModifier/internal_inline   ... let names = [internal]
--- FAIL: TestLetModifiers_NameIsNeverTheModifier/rec_inline        ... let names = [inline]
--- FAIL: TestLetModifiers_NameIsNeverTheModifier/inline_rec        ... let names = [inline]
--- FAIL: TestLetModifiers_NameIsNeverTheModifier/private_rec       ... let names = [private]
--- FAIL: TestLetModifiers_NameIsNeverTheModifier/rec_private       ... let names = [private]
--- FAIL: TestLetModifiers_NameIsNeverTheModifier/private_mutable   ... let names = [private]
--- FAIL: TestLetModifiers_NameIsNeverTheModifier/mutable_private   ... let names = [private]
--- FAIL: TestLetModifiers_NameIsNeverTheModifier/inline_generic_params          ... [inline]
--- FAIL: TestLetModifiers_NameIsNeverTheModifier/private_inline_generic_params  ... [private]
--- FAIL: TestLetModifiers_Indented
    indented `let inline helper` not named "helper"; let names = [outer inline]
--- FAIL: TestLetModifiers_SameModifierNoCollapse
    two same-indent `let inline` bindings produced 1 let entities, want 2: [inline]
    missing let entity "alpha"; missing let entity "beta"
--- FAIL: TestLetModifiers_CurriedParamsAreNotTheName
    curried binding with a modifier not named "scale"; let names = [add inline]
FAIL	github.com/cajasmota/grafel/internal/extractors/fsharp	0.913s
```

Green after the fix: `ok github.com/cajasmota/grafel/internal/extractors/fsharp 0.540s`.

## Enumeration table — the modifier space, not a sample

| modifiers | declaration | pre-fix name | post-fix |
|---|---|---|---|
| *(none)* | `let target (a: int) = a` | `target` | `target` |
| `rec` | `let rec target … = if a > 0 then target (a-1) else 0` | `target` | `target` |
| `mutable` | `let mutable target = 0` | `target` | `target` |
| `inline` | `let inline target (a: int) = a` | **`inline`** | `target` |
| `private` | `let private target (a: int) = a` | **`private`** | `target` |
| `internal` | `let internal target (a: int) = a` | **`internal`** | `target` |
| `public` | `let public target (a: int) = a` | **`public`** | `target` |
| `private inline` | … | **`private`** | `target` |
| `inline private` | … | **`inline`** | `target` |
| `internal inline` | … | **`internal`** | `target` |
| `rec inline` | … | **`inline`** | `target` |
| `inline rec` | … | **`inline`** | `target` |
| `private rec` | … | **`private`** | `target` |
| `rec private` | … | **`private`** | `target` |
| `private mutable` | `let private mutable target = 0` | **`private`** | `target` |
| `mutable private` | `let mutable private target = 0` | **`private`** | `target` |
| `inline` + generics | `let inline target<'T> (a: 'T) = a` | **`inline`** | `target` |
| `private inline` + generics | `let private inline target<'T> (a: 'T) = a` | **`private`** | `target` |

Each row also asserts the complementary direction — **no** entity may be named `rec`, `mutable`,
`inline`, `private`, `internal` or `public`. Recall alone cannot detect over-firing.

### Legality, from the specification grammar (review round 2)

`public` is in the allowlist beyond the set the issue listed, and it is **correct**. The F#
Language Specification settles it (§ 14.6 *Function and Value Definitions*, grammar also under
§ 6.6 *Definition Expressions*; accessibility in § 10.5 *Accessibility Annotations*; corroborated
by MS Learn, [*Access Control in F#*](https://learn.microsoft.com/dotnet/fsharp/language-reference/access-control)):

```
function-defn := inline? access? ident-or-op typar-defns? argument-pats return-type? = expr
value-defn    := mutable? access? pat …
access        := public | private | internal
```

…with `rec` sitting *outside* the definition (`let rec function-or-value-defns`). So
`let public target (a: int) = a` is grammatical and M3's kill rests on a real shape.

The same grammar fixes the **order**: access *follows* `inline`/`mutable`, and `rec` precedes
everything. So `inline private`, `rec inline`, `rec private` and `mutable private` are legal, while
**`private inline`, `internal inline`, `inline rec`, `private rec`, `private mutable` and
`private inline`+generics are not legal F#.** An earlier revision of the test header claimed every
row was legal; that was false for those six, and it is exactly the prose-asserts-what-nothing-
observes class.

**Resolution — option (a): the six rows are kept, and relabelled.** They are now marked
`lenienceOnly` with the reason spelled out *at the row level*, not only in the header, so reading
one row in isolation cannot mislead:

```go
// lenienceOnly — NOT legal F#: the row reverses the spec's modifier
// order (access must FOLLOW inline/mutable, and `rec` precedes the
// definition). It pins the scanner's tolerance, not the language.
lenienceOnly fsGrammar = "NOT legal F# — pins scanner lenience to reversed modifier order"
```

The label is interpolated into every failure message, so a future red run says which kind of row
broke. Option (b) — deleting them — was verified free (all mutants stay DEAD without them), but the
lenience is *real scanner behaviour*: accepting a reversed order is a deliberate policy for a
scanner that must never lose a binding to a token order it did not expect, and deleting the rows
would leave that policy unpinned and free to regress silently. Pinning it deliberately, with an
honest label, is worth more than six fewer rows. **No scanner behaviour changed in this round —
prose and labels only.**

No F# toolchain exists in this environment (`dotnet`, `fsc`, `fsharpc`, `mono` all absent), so the
classification is read off the specification grammar and **was not executed against a compiler**.
Rows are otherwise shaped to be well-formed: `mutable` rows take no parameters, `rec` rows recurse.

### Axes varied vs held constant

**Varied:** modifier identity (6 words); modifier count (0, 1, 2); modifier order (both orders of
`private`/`inline`, `rec`/`inline`, `rec`/`private`, `mutable`/`private` — one order of each pair is
grammatical, the reverse is a labelled lenience probe); presence of value
parameters; presence of a generic parameter list; indentation (top level in the table,
nested-inside-a-binding in `TestLetModifiers_Indented`); number of same-modifier bindings in one
file (1 in the table, 2 in `TestLetModifiers_SameModifierNoCollapse`).

**Held constant:** the bound name (always `target`, so a wrong capture is unambiguous); entity kind
`SCOPE.Operation` and subtype `let`; one binding of interest per source; the enclosing `module M`;
the file path; parameter types (`int` / `'T`). Held-constant axes are called out in the test file's
header comment so the next fixture author sees which neighbours are still open. Grammatical status
is **not** an axis the table holds constant: it is labelled per row (`specLegal` / `lenienceOnly`).

### Duplicate-collapse test

`TestLetModifiers_SameModifierNoCollapse` — `let inline alpha` and `let inline beta` at the same
indentation must yield **two** entities. It asserts the count **and** both names; pre-fix it saw
`1: [inline]`.

### Negative controls (both would fail on an over-wide fix)

1. `TestLetModifiers_CurriedParamsAreNotTheName` — `let add x y = x + y` and
   `let inline scale factor value = factor * value`. F#'s dominant function form is curried,
   space-applied parameters, so this shape is genuinely present, and *either* over-widening
   ("last identifier before the params", or "any word is a modifier") names these `y` / `value`.
   Mutant M7 confirms it is not vacuous: it kills M7.
2. `TestLetModifiers_ModifierPrefixedNames` — `recompute`, `mutableState`, `inlineCache`,
   `privateKey`, `internalId`, `publicFacing`: legal F# identifiers that merely *begin* with a
   modifier word, so a boundary-less allowlist would truncate them. Honest accounting: this control
   passes both before and after the fix, and mutant M5 shows the `\b` it targets is **not**
   load-bearing (see below).

There is no control available for "a binding whose name *equals* a modifier word": `rec`,
`mutable`, `inline`, `private`, `internal` and `public` are all F# reserved keywords, so
`let inline = 1` is not a legal program. The only way to bind such a name is the double-backtick
quoted form, ``let ``inline`` = 1``, which the name capture (`[a-zA-Z_][a-zA-Z0-9_']*`) has never
matched before or after this change. Stated rather than fabricated — again derived, not executed.

## Mutants scored on my own fix

Applied with an asserted exact match count, `go vet` before every score, `-count=1`, restored by
`cp` from a byte backup with **both** the shasum and `git diff HEAD` verified (never
`git checkout --`). `--- FAIL` lines counted, not exit codes. The modifier group has exactly **one**
occurrence in the package (asserted: `grep -c` = 1); the sibling fixed-sequence patterns are listed
under "Pair audit" below and are deliberately out of scope.

| # | mutant | verdict | `--- FAIL` |
|---|---|---|---|
| M1 | drop `inline` from the allowlist | DEAD | 12 |
| M2 | drop `private` from the allowlist | DEAD | 9 |
| M3 | drop `public` from the allowlist | DEAD | 2 |
| M4 | modifier group non-repeating (`)*` → `)?`) | DEAD | 11 |
| M5 | remove `\b` after the allowlist | **ALIVE** | 0 |
| M6 | revert to the fixed two-slot sequence | DEAD | 19 |
| M7 | allowlist → any identifier (`(?:\s+[A-Za-z_][\w']*)*`) | DEAD | 9 |
| M8 | `letSeen` key drops `name` (`indent+":let:"`) | DEAD | 13 |

M3 dying on only 2 lines is expected and correct: `public` appears in exactly one table row plus its
parent.

M8 is the separable grading probe for the collapse consequence — the collapse fix itself is **not**
separable from the name fix (there is no second code change: correct names produce distinct
`letSeen` keys on their own), so M8 attacks the key instead and confirms the collapse assertion
observes something. It kills `TestLetModifiers_SameModifierNoCollapse` among 13 failures.

**M5 is recorded as equivalent — not as DEAD — and the equivalence is STRUCTURAL, not merely
suite-relative.** Removing `\b` changes no observable behaviour: consuming `rec` out of `recompute`
leaves `ompute` with no preceding `\s+`, so that alternative dies and Go's engine takes the
zero-modifier one. I demonstrated it by running both regexes side by side over all 8 inputs of the
prefix control plus two modifier-bearing bindings (identical captures); independent review then
brute-forced the two patterns over **~592k enumerated inputs** — modifier words, modifier-prefixed
names, whitespace including newline, `'`, `(`, `<'T>`, `=`, depths 1–6, two alphabets — and found
**0 distinguishing cases**. No fixture was manufactured to kill it. The `\b` is kept as
documentation of intent, and the code comment now records the brute-force result rather than only
"not independently graded".

## Pair audit — sibling fixed-sequence patterns, confirmed present, deliberately NOT fixed here

Same defect family, other scanners, one subsystem per change. Measured with the real patterns
(probe output, not inspection):

| pattern | input | result |
|---|---|---|
| `moduleRE` | `module private Foo` | **MISS** (module lost entirely) |
| `typeRE` | `type private Foo =` | **MISS** |
| `memberRE` | `static member Go () = 1` | **MISS** |
| `memberRE` | `member private this.Go () = 1` | name = **`private`** |
| `memberRE` | `member val Go = 1 with get, set` | name = **`val`** |
| `compexpr_active_patterns.go:49` (`let (|Even|Odd|)`) | `let private (|Even|Odd|) n = n` | **MISS** |

Two distinct failure modes: `memberRE` mis-names exactly as `letRE` did, while the others fail to
match at all and silently drop the entity. Reported for filing, not touched.

## Verification

- `go test -count=1 ./internal/extractors/fsharp/` → `ok … 0.540s` (whole package, not just the new tests)
- `GRAFEL_HOME=<fresh> GRAFEL_DAEMON_ROOT=<fresh> go test -count=1 ./cmd/grafel/` → `ok … 191.464s`,
  exit 0. **Digest unmoved**, as predicted: `Name` *is* hashed by `entityTupleKey`, but the
  mixed-language fixture contains no F# files, so no fixture member changes. No constant bumped.
- `gofmt -l internal/extractors/fsharp/` → empty; `go vet ./internal/extractors/fsharp/` → clean
- Full repo suite deliberately not run (out of scope per brief).


cajasmota:
**My own mutant, and it settles the `\b` question rather than leaving it at "equivalent". The guards are MUTUALLY MASKING, and the pair is jointly graded — which is the honest reading and better news than either the author's or the reviewer's wording.**

Nobody scored the **separator**. Both prior mutants attacked the modifier group's boundary (`\b`); I attacked the thing that boundary is supposedly redundant with — the `\s+` between the modifier group and the name capture.

Worktree at `f464b8871`, `go vet` 0 before every score, exact match count 1, `-count=1`, restored by `cp` with shasum `361c88c5…` and `git diff HEAD` empty.

| mutant | change | `--- FAIL` | verdict |
|---|---|---|---|
| **M5** (author's) | drop `\b` from the modifier group | 0 | **ALIVE** |
| **DU-1** (mine) | name separator `\s+` → `\s*` | 0 | **ALIVE** |
| **DU-2** (mine, compound) | **both** | **1** | **DEAD** |

DU-2's killer is `TestLetModifiers_ModifierPrefixedNames`, and the behavioural difference is unambiguous. Probing the compound mutant on names that begin with a modifier word:

```
recompute      → "ompute"
mutableState   → "State"
inlineCache    → "Cache"
privateKey     → "Key"
internalId     → "Id"
publicApi      → "Api"
```

All six are correct under either single mutant, and all six break under the pair.

## What that means for the two prior verdicts

**The reviewer's brute-force result stands but needs one qualifier.** ~592k enumerated inputs found 0 cases distinguishing `\b` from no-`\b` — true, and my DU-1 explains *why*: `\s+` is doing that work. Symmetrically, `\b` is doing `\s+`'s work, which is why DU-1 is ALIVE. So "structurally equivalent" is right **conditional on the other guard being present**, and neither guard is individually load-bearing.

**The author's disposition was correct.** Recording M5 as "equivalent under the suite, with a demonstration" rather than as a kill was the right call, and the reasoning they gave (the engine backtracks because a group match is only useful when followed by `\s+`) is exactly the mechanism — they identified the masking relationship without naming it as such.

**And there is no hole here.** This is the one time this milestone that a mutually-masking pair turned out to be *fine*: the two guards are genuinely redundant with each other, the redundancy is belt-and-braces rather than a gap, and `TestLetModifiers_ModifierPrefixedNames` grades the pair. My standing rule is to treat two guards that only fire together as *both ungraded* — that rule catches real defects, and here it correctly surfaced the relationship and then found the behaviour pinned. So nothing to file and nothing to change.

Worth one line in the comment, though, since the current wording says `\b` is "not independently graded… kept as documentation of intent": that is accurate but undersells it. `\b` is *redundant given* `\s+`, and vice versa — and the pair is graded by name. Saying that outright stops the next person deleting one on the grounds that a mutant survives it.

## Still outstanding

The prose fix from the review (the six ungrammatical table rows whose header calls them legal F#) is with the author. `public` stays — settled from the F# spec grammar, so M3's kill rests on a real shape.

Separately filed from the review's pair audit: the `indent+":let:"+name` dedup key is written as **two independent literals** (`extractor.go:340` and `:385`), and mutating only the first to `":LET:"` is **ALIVE at 0 FAIL** — the two sites' agreement is ungraded, so the let/member cross-dedupe can silently stop working. Pre-existing, not this PR's, and it is the score-every-occurrence shape rather than a representative-site one.

Issue #7105, opened 2026-09-16
DECIDED (measured): derive Properties[subtype] from Entity.Subtype — 796/1069 rows are invisible to the dashboard and docgen, which read ONLY the Properties spelling; zero identity/digest cost. Kind-enforcement on typescript's lone violation stays deferred


cajasmota:
Found while implementing #7090 (emitting `Subtype` over MCP), and measured there rather than inferred.

`Kind` and `Subtype` carry overlapping information, and **which of the two holds a given distinction is decided per producer with no stated rule.** Measured across 14 golden fixtures, 846 entities:

| producer | class-vs-interface distinction lives in | dual-stamps `Properties["subtype"]`? |
|---|---|---|
| **java** | `Subtype` only — `class` and `interface` both land on Kind `SCOPE.Component` | no |
| **go** | `Subtype` only — struct/interface both under `SCOPE.Component` | only for its *file* entity |
| **typescript** | partly `Kind` — promotes `SCOPE.Enum`, `SCOPE.Schema`, `SCOPE.UIComponent` | most rows |

**No fixture language encodes class-vs-interface in `Kind` at all.** That is worth stating because the opposite was assumed in #7090's own issue body and in the brief that dispatched it — the belief was that *some* languages put it in `Kind` and others in `Subtype`, so an agent could at least rely on `Kind` for some producers. It cannot. The axis is real but it is not that axis.

## Why this is a defect and not merely an inconsistency

**A consumer cannot write one query that works across languages.** "List this project's interfaces" needs `Subtype == "interface"` for java and go, and `Kind == "SCOPE.Enum"`-style reasoning for some typescript distinctions. Any consumer that picks one route is silently wrong for the other producers, and — the part that makes it hard to notice — it still returns *results*, just an incomplete set. That is the #7056 shape: a plausible answer that no bind-rate, orphan-rate or dangle instrument can flag.

**`Properties["subtype"]` is a third spelling of the same fact**, present on some producers' rows and absent on others, and it does not derive from the canonical field — it is stamped independently. So the same distinction has up to three homes with no invariant tying them together. #7090 measured the consequence: **430 entities (51% of the corpus) carry a `Subtype` with no `Properties["subtype"]` twin**, and only **3 of 663** subtypes equal their kind tail, so the three spellings are genuinely not redundant copies of each other.

[...]

Issue #1626, opened 2026-05-22
P0: move .archigraph OUT of repos into the daemon folder — fixes the re-index loop + keeps repos clean


cajasmota:
Owner directive + root cause of the runaway daemon (330% CPU, 1h+). The indexer writes `.archigraph/` (graph.fb, graph.json, enrichments, links, etc.) INSIDE each repo. The daemon watcher/drift-check then self-triggers on its OWN in-repo `graph.fb`/`graph.json` writes (the "mtime drift in <repo>/.archigraph" log) → rebuild → rewrite → rebuild → INFINITE LOOP (confirmed: persisted even with watchers=false, so the graph_cache.go drift-check is the trigger). It also pollutes the repos / creates commit noise. 

FIX — relocate all daemon-generated graph artifacts to the archigraph-managed store: `~/.archigraph/store/<group>/<repo>/` (graph.fb, graph.json, enrichments, links, .metadata.json, etc.). 
1. **Indexer** (cmd/archigraph/index.go + the write paths): write artifacts to the daemon store, NOT the repo. 
2. **Readers**: internal/graph/load.go, internal/mcp/state.go, internal/dashboard, internal/links — read from the store location. 
3. **Kill the loop**: with no in-repo graph.fb/json, the watcher/drift-check no longer self-triggers; ALSO fix/remove the graph_cache.go drift-rebuild trigger so it never rebuilds on .fb-vs-.json mtime drift (that is an internal artifact, not a source change). 
4. **Migrate**: move existing in-repo `.archigraph/` for upvate + polyglot into the store (so they do not need a full re-index), then the repos are clean. 
5. **Watchers**: re-enable safely — watch SOURCE only; the output is now external so no self-trigger. Re-enable watchers default after this is safe. 
NOTE: overlaps in-flight #1620 (internal/graph + daemon.go) — branch off main; expect to rebase. 
ACCEPTANCE: repos contain NO `.archigraph/`; the daemon reads graphs from `~/.archigraph/store/<group>/<repo>/`; a normal index + a reactive re-index do NOT loop (CPU settles); upvate/polyglot work after migration without a full re-index. Verify on an ISOLATED daemon (the live one is intentionally stopped).

Issue #481, opened 2026-05-18
Indexer: non-deterministic output from goroutine worker pools (causes measurement noise)


cajasmota:
PR #479's reported lua regression turned out to be measurement noise — 10 identical builds on same input produced bug-rates ranging 0.93% to 10.14%.

## Root cause
cmd/archigraph/index.go uses goroutine worker pools at lines 912, 1008, 1063. Combined with:
- Per-run `failed=1` (which file fails varies)
- Go map iteration order non-determinism in downstream dedup/classification

Small repos (<200 endpoints) see ±3-5pp swings on identical input.

## Fix
- Sort per-worker outputs by canonical entity ID before downstream passes
- Investigate the consistent `failed=1` file in kickstart.nvim (intermittent timeout or panic)
- Acceptance: same input → byte-identical graph.json across 10 runs

## Impact
Without this, measurement-based regression detection is unreliable on the small-repo half of the corpus. Blocks confident ship-gate trend tracking.

Refs #44 #479

cajasmota:
WIP — implementer dispatched (blocks wave 3 measurement).

cajasmota:
**Root-cause investigation + fix shipped — PR #486** (commit cace4a6)

Five distinct sources of non-determinism, all reproduced empirically on kickstart.nvim (10 runs, before fix: 10 distinct SHA256 hashes / bug-rate 0.93%–10.14%):

1. **Concurrent tree-sitter parsing.** The bundled smacker/go-tree-sitter ``*sitter.Language`` pointers in ``internal/treesitter.languageRegistry`` are shared across all parsers. Empirically the SAME Lua source file emitted between 0 and 17 entities depending on which workers were running in parallel. Per-file extraction is fully deterministic when serialised. Fix: ``parseMu`` around ``Parse``.
2. **Pass 1/2.5/3 worker-pool fan-in order** — ``BuildIndex`` first-writer-wins depended on goroutine scheduling. Fix: sort each fan-in slice on canonical fields.
3. **Lua extractor** iterated the module-table map without sorting, so emitted entity slice index and CONTAINS edge wiring varied.
4. **gonum graph algorithms** — articulation-point DFS root, ``IdentifyGodNodes``, ``ComputeSurpriseEdges``, ``ComputeCommunities`` all had non-deterministic tiebreakers. Fix: ID-based tiebreakers and sorted adjacency.
[...]

PR #7137, opened 2026-09-17
fix(#7105): 796 of 1069 entities carry a Subtype the dashboard and docgen cannot see — derive the Properties twin at the emission funnel

cajasmota:
Implements the decision recorded in the newest comment on #7105 (the issue **body's** two figures are refuted there; the comment's are the ones used): **derive `Properties["subtype"]` from the canonical `Entity.Subtype` at the emission chokepoint, unconditionally.** Nothing else from that issue is touched.

> **Review round 2 addressed (items 1–5).** The chokepoint is unchanged. The docstring's mechanism was wrong and is now scored rather than argued — and scoring it refuted the *replacement* claim too, which is the main finding of this round. See **Mechanism, measured** below.

## The gap, and why it is invisible

`Entity.Subtype` is the canonical home of every sub-distinction. `Properties["subtype"]` is a second spelling that 24 producers stamp independently — 22 of them only for their `file` entity. **Three consumers read the duplicate**, all re-verified on this commit:

- `internal/dashboard/graphstate.go` — `serializeEntity` emits no `subtype` key at all; `properties` (gated on `PropLen() > 0`) is the wire shape's only route.
- `internal/docgen/llm_bundle.go:1679` — `configEntryFromEntity` reads `get("subtype")`; `n.Subtype` is never read.
- `internal/docgen/tier0.go:711` — `renderSection`'s `**Properties:**` block, also gated on `PropLen() > 0`.

796/1069 (74.5%) of the golden corpus carries `Subtype` with no Properties twin, so three quarters of it is subtype-less to those surfaces. Deriving is behaviour-preserving on measured data: disagreement **0/211**, Properties-only **0/1069**.

## Chokepoint: `graph.SortDocumentForEmission`

One call, plus `internal/graph/subtype_property.go`. It is the repo's existing pre-serialization normaliser, reached by every production writer (`fbwriter` flat + segmented, the daemon's incremental rewrite, the determinism checker), idempotent, and in-place on the shared document.

**Rejected `fbwriter.buildEntity`** — the documented sole entity-serialization leaf, and otherwise the stronger candidate — because it is FlatBuffers-only, and `graph.json` (deliberately mtime-matched to `graph.fb` as "two encodings of the SAME index pass", #1626) would then disagree with the binary graph about the very field this issue is about. **Rejected read-time normalisation** in the `Prop*` accessors: `serializeEntity` and `renderSection` both gate on `PropLen() > 0`, so the key would never surface without also changing `PropLen`, and `len(props) == 0` is load-bearing around the indexer.

### Mechanism, measured — and two wrong statements of it on the way

This paragraph exists to explain why `graph.json` and `graph.fb` agree, which is the whole reason `buildEntity` was rejected. **It has now been wrong twice and is scored the third time.**

- v1 said *"every graph written to disk goes through `fbwriter.WriteGraphGenReport`, and both of its branches call it"*. **False** — `graph.json` is written by `graph.WriteAtomic` (`internal/graph/graph.go:864`), which normalises nothing and never goes near `fbwriter`. (Review item 1, correct.)
- v2 said `cmd/grafel/index.go:999` and `internal/extractors/incremental.go:1700` *"ARE the guarantee"*. **Also false.** I scored the mutant rather than assuming: **deleting `index.go:999` leaves the parity test GREEN.**

What is actually true is weaker and sufficient: **some normaliser runs on the shared `doc` pointer before `graph.json`'s encoder does, and there are two redundant, mutually-masking opportunities for it.** `index.go:1006`'s `.fb` write calls `SortDocumentForEmission(doc)` on the same pointer as `writeGraphGenFlat`'s *first statement* — so the document is normalised even when the marshal that follows fails, which matters because `graph.json` is still attempted after an `.fb` failure. `index.go:1094` then writes JSON from that already-normalised doc. Scored as a pair, because two guards that only fail together grade neither:

| pair-audit mutant | verdict |
|---|---|
| **F** — `index.go:999` removed (fbwriter's call left) | **ALIVE** |
| **G** — both `fbwriter` `SortDocumentForEmission` calls removed (`index.go:999` left) | **ALIVE** |
| **H** — both removed (the compound) | **DEAD**, 1 `--- FAIL` |

So `fbwriter`'s calls are **not** belt-and-braces for this property; they are one of the two guarantees, and either alone suffices. The docstring, the inline comment and the parity test's own header now say this, and the parity test's header warns explicitly not to read a green run as evidence that either single site is load-bearing.

## Strict TDD — red output before the fix existed

```
$ go test -count=1 -run 'SeesCanonicalSubtype' ./internal/dashboard/
--- FAIL: TestDashboardSerializeEntity_SeesCanonicalSubtype_7105 (0.02s)
    --- FAIL: TestDashboardSerializeEntity_SeesCanonicalSubtype_7105/gap_becomes_visible (0.00s)
        subtype_property_7105_test.go:103: #7105: serializeEntity emitted no `properties` at all for an entity whose Subtype is "interface"; the dashboard cannot see the subtype by any route (there is no top-level `subtype` key in the wire shape)
FAIL
FAIL	github.com/cajasmota/grafel/internal/dashboard	1.181s
```

```
$ go test -count=1 -run 'TestDocgenConfigEntry_SeesCanonicalSubtype_7105' ./internal/docgen/
--- FAIL: TestDocgenConfigEntry_SeesCanonicalSubtype_7105 (0.01s)
    --- FAIL: TestDocgenConfigEntry_SeesCanonicalSubtype_7105/gap_becomes_visible (0.00s)
        subtype_property_7105_test.go:70: #7105: ModuleConfigEntry.Subtype = "", want "go_module"; docgen reads props["subtype"] only and cannot see Entity.Subtype
FAIL
FAIL	github.com/cajasmota/grafel/internal/docgen	0.653s
```

Every test goes through the **production write path and the loader the consumers use** — `fbwriter.WriteGraphGenReport` into a `t.TempDir()`, then `graph.LoadGraphFromDir` (`internal/dashboard/store.go:247`) — and asserts on the **emitted artefact**, never on a hand-built Properties map. The preservation and forbidden subtests were green on `main` before the fix, as they must be.

### New in this round

- **`cmd/grafel/subtype_property_json_parity_7105_test.go`** (review item 2) — a real `Index()` of `testdata/crossfile_go` with `--export-json`, asserting the derived key in `graph.json`, the forbidden empty direction on real rows, and **entity-for-entity agreement between the two encodings**, with a non-vacuity floor. Observed: `10/13 graph.json entities carry a canonical Subtype; all 10 agree with graph.fb`. Before it, nothing in the repo exercised the JSON encoding of this field at all, and compound **H** was ALIVE.
- **`TestDocgenRenderSection_SubtypeWidening_7105`** (review item 3) — the reviewer is right that my disclosure was stronger than necessary. `renderSection` lists every `k = v`, so it **can** distinguish "gained nothing" from `subtype = `, and the empty direction is now graded at that surface. The comment in the `configEntryFromEntity` test is narrowed to what is genuinely ungraded: that reader, and only that reader, conflates absent with empty.
- **`TestSubtypeConflictIsReportedNotSilent_7105`** (review item 5) — see below.

### The forbidden direction

Deriving is a **widening** — it adds a key to 796 rows — and a recall-style suite is blind to one that fires too broadly. **Absent is correct, not empty-string.** Graded by: `forbidden_empty_subtype_gains_nothing`, `forbidden_property_less_entity_stays_property_less` (dashboard wire map), `forbidden_empty_subtype_renders_no_subtype_line`, `forbidden_property_less_entity_renders_no_properties_block` (tier-0 markdown), the empty-`Subtype` walk in the `graph.json` parity test, and `forbidden_existing_property_is_never_overwritten`.

## Disclosure: a real change to rendered output (review item 4)

**The 796 rows now render a `**Properties:**` block in tier-0 docs where they previously rendered none**, because the map is newly non-empty. It is the intended consequence of making the subtype visible to Properties-only readers, and `gap_renders_the_derived_property` is the row that pins it. **I agree with disclosing rather than suppressing it** — suppression would mean either a derive that hides from one reader (re-creating the split this issue is about) or a special case in `renderSection`'s gate, and the rendered line is accurate and useful. The reviewer's sweep of every other `PropLen() > 0` gate (`llm_bundle.go:809,1337,1474,1495,1777,1859`, `lombok_stub.go:39`) found only keyed `PropGet`s of unrelated keys, so nothing else changes.

## Nits (review item 5) — both taken

The dead `stamped` return is **gone**, and the silent resolution is **gone with it**: `DeriveSubtypeProperties` now returns `[]SubtypeConflict` and the funnel emits one stderr line naming the count and the first offending entity. I took the warn because a disagreement resolved in silence is exactly the shape that would be hardest to find later. It is graded on the **emitted stderr**, in both directions — including the two that matter more than the warning itself: the ordinary derive path (796 rows) must stay **silent**, and a second idempotent pass over a key this normaliser stamped itself must not read it back as a conflict, since the producers invoke the funnel more than once per write.

## Mutation scores — re-scored from scratch this round, never carried over

`go vet` before every score, exact match count asserted per mutant, `-count=1`, restore by `cp` from a byte backup with **both** `shasum -c` and `git diff HEAD` verified — no `git checkout --`. Occurrence count asserted for every mutated expression (1 each, except G's 2). The scripts assert my own worktree path *and* branch *and* a clean tree, and abort otherwise.

| mutant | `--- FAIL` lines | verdict |
|---|---|---|
| baseline (unmutated), both selections | 0 | — |
| **A** — derivation call + report removed entirely | **9** | DEAD |
| **B** — derivation made vacuous (`ok` → `!ok`) | **11** | DEAD |
| **C** — empty-`Subtype` guard removed | **6** | DEAD |
| **D** — existing-key guard removed (always overwrite, never report) | **4** | DEAD |
| **E** — conflict report suppressed (silent resolution) | **2** | DEAD |
| **F / G / H** — the normalisation pair audit | see table above | ALIVE / ALIVE / **DEAD** |

**A was reported as a kill last round on a mutant that did not compile.** Deleting the call site left `fmt` and `os` imported and unused; `go vet` caught it, it was not scored, and as a compiling mutant (imports fixed) it is DEAD with 9 `--- FAIL` lines. Flagging it because a non-compiling mutant exits non-zero with zero `--- FAIL` lines and reads as a kill to anything counting exit codes.

**F and G are recorded as ALIVE and equivalent under the current suite, with the demonstration** — H, their compound, is DEAD. They are not presented as kills.

## Blast radius

- **Entity IDs / edges: no movement.** `ComputeID` is `sha256(OrgID+ProjectID+SourceFile+Kind+Name)` (`internal/types/entity.go:137`).
- **Pinned `cmd/grafel` semantic digest: UNMOVED.** `go test -count=1 ./cmd/grafel/` with a fresh `GRAFEL_HOME` **and** a fresh `GRAFEL_DAEMON_ROOT`: `ok … 193.0s`, 0 `--- FAIL`. The three digest gates were then run by name to prove they were not skipped — `TestCustomExtractorGateOffGraphIsUnchanged`, `TestSemanticDigestGradesEntityLanguage6862`, `TestCustomExtractorGateOffDeltaIsExactlyTheDocumentedSpanGain` — all PASS. `entityTupleKey` hashes eight fields; `Properties` is not among them. **No pinned constant touched.**
- **Golden `expected.json`: not exposed** — `Properties` are not asserted by the quality harness.
- **Intended behavioural consequence:** `PropLen()` becomes non-zero for the 796 rows, which is what makes both `PropLen() > 0` gates fire. The empty-`Subtype` guard is what keeps that from reaching rows with nothing to say.

## Verification run (all `-count=1`, exit codes trusted, no `rtk` wrapper)

- `./internal/graph/...`, `./internal/dashboard/`, `./internal/docgen/`, `./internal/extractors/` — exit 0, **0 `--- FAIL`**
- `./cmd/grafel/` — `ok … 193.0s`, **0 `--- FAIL`**, fresh `GRAFEL_HOME` + `GRAFEL_DAEMON_ROOT`
- `gofmt -l cmd internal` — clean; `go vet` on `./internal/graph/ ./cmd/grafel/ ./internal/dashboard/ ./internal/docgen/` — clean

No full-repo suite was run.

## Explicitly out of scope (unchanged)

No `Subtype` **value** is changed anywhere. The enum value-set facet (`extractor.EnumEntity`) is untouched. typescript's `interface` → `SCOPE.Schema` is **not** normalised. The `SCHEMA.md` rule statement is a separate change needing its own forbidden-row grading.

Refs #7105.


cajasmota:
**My own mutant scores the gap the review could only argue from code reading — and it is wider than the subtype key.**

The reviewer established by enumeration that no production writer misses the derivation, but that `graph.json` is written by `graph.WriteAtomic`, which normalises **nothing** — it only carries the key because `cmd/grafel/index.go:999` normalised the shared `doc` in place first. They flagged, under "could not verify", that **no test exercises the JSON encoding for this field**. So I mutated the thing that claim rests on.

**DV-1: delete the unconditional `sortDocumentForEmission(doc)` at `cmd/grafel/index.go:999`**, leaving `fbwriter`'s own internal calls in place. Worktree at `99bb4f0ed`, `go vet` 0, exact match count 1, `-count=1`, restored by `cp` with shasum `61dee653…` and `git diff HEAD` empty.

| package | `--- FAIL` |
|---|---|
| `./internal/graph/...` + `./internal/dashboard/` + `./internal/docgen/` | **0** |
| `./cmd/grafel/` (fresh `GRAFEL_HOME` + `GRAFEL_DAEMON_ROOT`, 194.4s) | **0** |

**ALIVE everywhere.** The derivation's own tests cannot see it because they reach the writer through `fbwriter`, which still normalises; and `./cmd/grafel/`, which is where `index.go` and `--export-json` actually live, does not observe it either. So the reviewer's reading was right and it is now measured: the `graph.json` path carries this key **by luck of a shared in-place mutation**, and nothing would notice if that stopped.

## The part that is bigger than this PR

That call is not only the subtype derivation's guarantee. Its own comment says it is the **#481 belt-and-braces final sort** — *"external.Synthesize appends placeholders and Pass 4 attaches per-entity attributes via map lookups; resort by canonical IDs so the on-disk bytes are stable across runs of the SAME repo."*

DV-1 deletes that too, and **still nothing fails**. `fbwriter` sorts internally, so `graph.fb` stays byte-stable and the determinism checks that exist keep passing — but `graph.json` would be emitted unsorted, and no test observes it. So the #481 determinism guarantee for the JSON artefact is ungraded, and has been since long before this PR. **Filing that separately; it is not this change's to fix.**

## What this means for #7137

Not a defect in the change — the behaviour is correct and the call is unconditional. It means the **docstring names the wrong mechanism** (it credits `fbwriter.WriteGraphGenReport`, which `graph.json` never goes through) and the mechanism it *should* name is unobserved. Both are in the fix round I have sent back: correct the paragraph to say in-place normalisation before either encoder runs, and add the one parity row that makes it true-by-test rather than true-by-luck.

That row now has a scored justification rather than a stylistic one: **without it, deleting the line the whole guarantee rests on is invisible to every package in the repo.**

Also confirmed from the review, and worth recording: the four mutant counts reproduced exactly (A=4, C=3, D=2, each expression occurring exactly once), idempotence was demonstrated **by execution** rather than by reading — production already double-invokes, at `index.go:999` and `writeGraphGenFlat:108` — and both of your rejected alternatives were independently verified as correctly rejected.

cajasmota:
**My DV-1 comment above is wrong in its mechanism, and this round refuted it by scoring rather than by arguing. That is exactly what I want, so I am recording it prominently.**

I claimed the `graph.json` path carried the derived key *"by luck of a shared in-place mutation"* and that removing `cmd/grafel/index.go:999` would leave it unobserved. The measurement was right — DV-1 is ALIVE — but the reason was not. `writeGraphGenFlat` takes `doc *graph.Document`, a **pointer**, and its first statement is `graph.SortDocumentForEmission(doc)`. It normalises the **shared document in place**, and the `.fb` write at `:1006` runs before the JSON write at `:1094`. So `graph.json` is normalised either way.

The two sites are **mutually masking**, and the pair was scored properly:

| mutant | verdict |
|---|---|
| **F** — `index.go:999` removed | **ALIVE** |
| **G** — both `fbwriter` calls removed | **ALIVE** |
| **H** — both | **DEAD** |

Same shape as `\b`/`\s+` on #7134. And the line in my fix-round message calling `fbwriter`'s calls "belt-and-braces" is backwards — for this property they are a **primary** opportunity, not a fallback. Good call putting that in the code comment.

**So the new docstring is right to state the weaker claim.** "Some normaliser runs on the shared doc before `graph.json`'s encoder, via two redundant opportunities", with the pair scores attached and both earlier wrong versions recorded, is the accurate statement. The parity test grading the **compound** with a header warning not to read its green as evidence for either single site is the correct disposition — and it is the one that would have stopped me writing the DV-1 comment the way I did.

I have corrected #7140 as well; it is retitled to the narrower question that actually survives — whether anything asserts `graph.json`'s **row order**, which H's DEAD does not establish, since with both sorts gone the derivation stops running and the parity test fails for the key's absence rather than for ordering.

## The non-compiling mutant is the more important catch

Round 2's mutant A was scored on an edit that **did not compile** — deleting the call left `fmt`/`os` unused. It read as DEAD at 4 `--- FAIL`. As a compiling mutant it is DEAD at **9**.

The part worth dwelling on: the independent reviewer **also** reported A=4. Two parties, scoring separately, agreeing exactly — and both wrong, because they reproduced the same broken edit. That is not two confirmations; it is one premise confirmed twice. `go vet` before scoring is the standing rule precisely for this, and it is why the rule is "vet, then count `--- FAIL` lines", not "the exit code was non-zero".

Full re-score after fixing it: **A=9, B=11, C=6, D=4, E=2, all DEAD; F/G ALIVE-equivalent with H as the demonstration.** Every count moved, which is the other reason a broken mutant is dangerous — it does not merely mis-score itself, it makes the whole table look settled.

## Items 3 and 5

Closing the docgen gap via `renderSection` forbidden rows was the right call and the disclosure is now correctly narrowed to `configEntryFromEntity` alone. On item 5, taking **both** nits was more than I asked: replacing the dead `stamped` return with `[]SubtypeConflict` and a stderr line, then **grading the stderr in both directions** — including that the ordinary 796-row path stays silent and a second idempotent pass reports nothing — turns "should we warn?" from an opinion into observed behaviour. The silence assertion is the half that usually gets skipped.

Item 4 disclosed rather than suppressed, which I agree with: hiding the new `**Properties:**` block would need either a special case in `renderSection`'s gate or a derive that lies to one reader, and both are worse than a documented output change.

cajasmota:
**Merge gate closed. My own killing mutant is DEAD, and it answers the one question the review left open about the parity test's own strength.**

DV-1 turned out equivalent (masked by `fbwriter`'s in-place sort), so it did not satisfy the gate. Here is one that does.

**DX-1: stamp only the FIRST eligible entity** — `break` immediately after `e.PropSet("subtype", e.Subtype)`. This is the partial-application direction: the derivation still runs, still fires, still sets the right key with the right value, and simply stops after one row. It is the shape a non-vacuity floor exists to catch, and nobody had scored it.

Worktree at `120924122`, `go vet` 0, exact match count 1, `-count=1`, restored by `cp` with shasum `3b856053…` and `git diff HEAD` empty.

| suite | `--- FAIL` |
|---|---|
| `./internal/graph/...` + `./internal/dashboard/` + `./internal/docgen/` | **0 — ALIVE** |
| `./cmd/grafel/` (fresh `GRAFEL_HOME` + `GRAFEL_DAEMON_ROOT`, 187.0s) | **13 — DEAD** |

**DEAD, and the killer set is more interesting than the verdict.**

`TestGraphJSONAndFBAgreeOnDerivedSubtype_7105` is among the 13 — so the non-vacuity floor you added (10 of 13 rows carrying a canonical `Subtype`, all agreeing) **does** discriminate partial application, not merely presence. That was the thing I could not tell from reading it, and it is now measured.

The other **12** killers are all pre-existing and none of them are yours:

```
TestDiffReindex_CatchesInjectedMismatch      TestContentParity_PathA_6129
TestDiffReindex_FlowChainChange              TestContentParity_PathB_6129
TestDiffReindex_FlowDocsOnlyChange           TestMountParity_6461_RouteEdit_PathA / PathB
TestDiffReindex_NewCrossFileCall             TestMountParity_6461_MountEdit_PathA / PathB
TestIncrementalIndexPathB_ParityWithFullReindex   …_MountEdit_PathA_AmbiguousRouter
```

That is the incremental-vs-full parity infrastructure, and it catches this for a structural reason: a full reindex and an incremental reindex stamp **different subsets** under DX-1, so the two graphs diverge. Their diff output names the field directly — `properties {+subtype="function"}` — so a future partial-application regression will point straight at the derivation rather than reading as generic reindex churn.

**Why that matters beyond this PR.** Putting the derivation at a chokepoint that both the full and incremental paths traverse means the repo's existing parity harness grades it for free, in a dimension the PR never had to write a test for. That is a real argument for `SortDocumentForEmission` over `fbwriter.buildEntity` that neither of us made: the FlatBuffers-only site would have been invisible to every one of those 12 tests.

**The two ALIVE rows are also not a gap.** `internal/graph`, `internal/dashboard` and `internal/docgen` all stay green under DX-1 because each asserts on entities it constructs itself — they grade the *derivation's behaviour*, and cardinality across a real indexed corpus is `cmd/grafel`'s job. Worth knowing which suite owns which question, so nobody later "fixes" this by bolting a count assertion onto the unit tests.

Merging. Closes the decided part of #7105; the deferred typescript `Kind` normalisation and the SCHEMA.md rule statement stay open there, and #7140 keeps the narrower `graph.json` row-order question.

Issue #7133, opened 2026-09-17
KOTLIN ARM ONLY: kotlin.go:1112 buildClassSignature has the same inverted annotation/brace order — '@Table(name = "{weird}") class Foo' yields an EMPTY signature (C# arm shipped in 84964094e; where-clause split to #7139)


cajasmota:
## What

#7124/#7128 fixed `buildClassSignature` in the **java** extractor: it cut at the first `{` *before* stripping annotation arguments, so a brace inside an annotation string truncated the declaration inside the literal. Its sibling `buildMethodSignature` did the two operations in the correct order and carried a comment naming the hazard.

**The same inverted pair exists in C# and Kotlin, and in C# it is the same documented-sibling shape** — one function's comment names the hazard while the function next to it has the bug.

| file | correct | inverted |
|---|---|---|
| `internal/extractors/csharp/csharp.go` | `buildMethodSignature` (`:1805`) — strips attribute args first, with a comment naming the hazard | `buildClassSignature` (`:1825`) — cuts at the brace first |
| `internal/extractors/kotlin/kotlin.go` | — | `:1112` |

## Measured, with one-character controls

C#, canonical ASP.NET routing:

```
[Route("api/{version}/items")] public class ItemsController   →  Signature ""
[Route("api/v1/items")]        public class ItemsController   →  Signature "public class ItemsController"
```

Kotlin:

```
@Table(name = "{weird}") class Foo  →  Signature ""
@Table(name = "weird")   class Foo  →  Signature "class Foo"
```

Two things make these **worse than the java case they mirror**:

1. **The output is an empty signature, not a truncated one.** Java lost `class Foo` from the tail; here the entire signature is `""`.
2. **`[Route("api/{version}/...")]` is the single most common attribute shape in ASP.NET.** A brace inside a route template is not an exotic fixture — it is how the framework's own documentation writes it. Incidence is likely **higher** than the `@Table(name = "{weird}")` shape that prompted #7124.

## Java itself is clean — verified, not assumed

The java file's remaining signature builders do not share the pattern: `buildFieldSignature` reads the parse tree rather than cutting text (rewritten by #7122), and `buildAnnotationElementSignature` has no brace cut. The `SCOPE.Schema` twin record for a class uses a separate name-based builder and is unaffected. So this is a cross-language sibling, not a missed site in the file #7128 touched.

## Grading requirement

[...]

PR #7138, opened 2026-09-17
fix(#7133): C# buildClassSignature cut at a brace inside an attribute string — and BOTH halves of the pair were ungraded

cajasmota:
Refs #7133 — C# arm. Deliberately **not** `Closes`: `kotlin/kotlin.go` has the same inversion, is untouched here, and stays open on that issue.

## The defect

`buildClassSignature` (`internal/extractors/csharp/csharp.go`) cut the declaration at the first `{` **before** stripping attributes. A brace inside an attribute argument string is indistinguishable from the body brace to `strings.Index`, so the cut landed inside the literal — and then `stripCSharpAttributes` found no matching `]` for the unbalanced `[` remainder and swallowed it to end-of-string. The signature came out **empty**, not merely truncated.

## Red first

Test file written and run before the fix existed, on `e4dee929e`:

```
=== RUN   TestCSharp7133_ClassSignatureStripsAttributesBeforeBraceCut
    ItemsController.Signature = "", want "public class ItemsController"
    BraceItems.Signature = "", want "public class BraceItems"
    IRepo.Signature = "", want "public interface IRepo"
    ItemRec.Signature = "", want "public record ItemRec(int Id)"
    Point.Signature = "", want "public struct Point"
    Based.Signature = "", want "public class Based"
    MultiLine.Signature = "", want "public class MultiLine"
    Stacked.Signature = "", want "public class Stacked"
    TypeParamAttr.Signature = "public class TypeParamAttr<", want "public class TypeParamAttr<T>"
    Outer.Signature = "", want "public class Outer"
    Inner.Signature = "", want "public class Inner"
--- FAIL: TestCSharp7133_ClassSignatureStripsAttributesBeforeBraceCut (0.00s)
=== RUN   TestCSharp7133_ColonInAttributeArgumentIsNotTheDelimiter
    BraceColon.Signature = "", want "public class BraceColon"
--- FAIL: TestCSharp7133_ColonInAttributeArgumentIsNotTheDelimiter (0.00s)
=== RUN   TestCSharp7133_MethodSignatureStripsAttributeArgsBeforeBraceCut
--- PASS: TestCSharp7133_MethodSignatureStripsAttributeArgsBeforeBraceCut (0.00s)
FAIL	github.com/cajasmota/grafel/internal/extractors/csharp	0.885s
```

2 `--- FAIL`. Every control row — `PlainController`, `PlainItems`, `Plain`, `RouteAttribute`, `DisplayAttribute`, `ColonOnly`, `ColonAndBase`, `BareBase` — was green in the red run, which is what makes the defect rows grade the **order** and not the presence of a string. The method sibling passed, as expected: its order was already correct.

## The ordering chosen, and why

**collapse → strip → brace cut → `" :"` cut** — i.e. the brace cut moved **last**, not the strip moved first. Moving the strip to the front would also move it before `strings.Join(strings.Fields(raw), " ")`, handing `stripCSharpAttributes` raw multi-line source with original indentation, an input class it has never received.

**Corrected after review.** I first recorded mutant **d** (collapse moved after the strip) as ALIVE-but-equivalent, arguing that `stripCSharpAttributes` only ever branches on `[`, `]` and ` `. The alphabet claim is true; the consequence is not. The helper skips the run of plain **spaces** after the closing `]` — when what follows `]` is a **newline** that skip does not fire, and stripping before the collapse leaves an interior space the collapse can no longer remove. Two ordinary-C# shapes show it, and both are now fixture rows:

```
public class TPAttrNewline<[Route("{v}")]
        T> { }
  HEAD     -> "public class TPAttrNewline<T>"
  mutant d -> "public class TPAttrNewline< T>"

public record PersonRec([property: Route("first_{n}ame")]
        string First, string Last) { }
  HEAD     -> "public record PersonRec(string First, string Last)"
  mutant d -> "public record PersonRec( string First, string Last)"
```

Mutant **d** is now **DEAD at 1 `--- FAIL`**, failing on exactly those two rows and nothing else. So collapse-first is kept because it is **correct** — it is the ordering that yields `<T>` — not because the alternative is indistinguishable.

The brace cut stays **before** the `" :"` cut, exactly as today, so the base-list cut still sees a body-stripped string.

## The second delimiter: traced, not assumed

The issue asked whether the `" :"` base-list cut is a second, independent truncation path via a partial attribute left by the brace cut. It is **not reachable**. Traced step by step on `e4dee929e`:

```
raw            "[Display(Name = \"{x} : b\")] public class BraceColon { }"
after braceCut "[Display(Name = \""            (FIRED)
after collapse "[Display(Name = \""
after strip    ""                              <- unbalanced "[" swallowed to EOS
after colonCut ""                              (no " :" left to find)
```

The brace cut always wins and the strip empties the string, so `" :"` is handed `""`. And a colon inside an attribute argument has **never** been a delimiter in either order, because the attribute is stripped before that cut both before and after this change — `[Display(Name = "a : b")] public class ColonOnly` was already correct on `e4dee929e`. Graded in its own test, separate from the brace grid, in both directions (mutant **b** removes the `" :"` cut).

## Grading axes

**Varied:** whether the attribute argument's string literal contains `{`. For `ItemsController`/`PlainController` and `BraceItems`/`PlainItems` the only **deliberate** difference is the two brace characters in the route template. Their type names differ too, necessarily — two types cannot share a name in one file — and that difference is inert here: `Plain`, which carries no attribute at all, shows a distinct name emits a correct signature on its own.

**Held constant within each pair:** the file, the namespace, the attribute type, the attribute's argument shape, the modifiers, the declaration keyword, the body, and the enclosing scope.

**Second axis, varied independently:** declaration kind (class / interface / record / struct), attribute position (on the type / stacked on the type / on a type parameter / on a record positional parameter / on a nested type), presence of a base list, whether the attribute spans a line break, and — the axis the review found open — whether the character **after** the attribute's closing `]` is a space or a newline.

**Held constant across the whole grid:** the extractor entry point (`Extract` over one file), the assertion (exact `Signature` string equality on exactly one entity of the expected kind), and `Kind`/`Subtype`.

## Enumeration table

| shape | before | after | status |
|---|---|---|---|
| `[Route("api/{version}/items")]` class — the reported case | `""` | `public class ItemsController` | fixed |
| `[Route("api/v1/items")]` class — one-pair control | correct | correct | control, green both |
| `[Route("api/{v}/items")]` / `[Route("api/v/items")]` minimal pair | `""` / correct | correct / correct | fixed + control |
| brace-in-attribute × **interface** | `""` | `public interface IRepo` | fixed |
| brace-in-attribute × **record** (positional list + body) | `""` | `public record ItemRec(int Id)` | fixed |
| brace-in-attribute × **struct** | `""` | `public struct Point` | fixed |
| brace-in-attribute × **enum** | n/a | n/a | **not in grid** — `enum_declaration` takes `buildEnumEntity`, a different builder that packs the member list into `Signature` |
| brace-in-attribute + real base list | `""` | `public class Based` | fixed (base list still dropped) |
| attribute **spanning a line break** | `""` | `public class MultiLine` | fixed |
| **stacked** attribute lists, brace in the second | `""` | `public class Stacked` | fixed |
| attribute on a **type parameter**, `]` followed by a space | `public class TypeParamAttr<` | `public class TypeParamAttr<T>` | fixed (truncated, not emptied, before) |
| attribute on a **type parameter**, `]` followed by a **newline** | `public class TPAttrNewline<` | `public class TPAttrNewline<T>` | fixed — **and this row is what kills mutant d** |
| `[property: ...]` on a **record positional parameter**, on its own line | `public record PersonRec(` | `public record PersonRec(string First, string Last)` | fixed — second mutant-d killer |
| attribute on a **nested type** (outer + inner) | `""` / `""` | `public class Outer` / `public class Inner` | fixed |
| attribute on a **method parameter** | correct | correct | already correct — the method builder's path |
| `{` in a **preceding doc comment** (`/// <summary>see {brace}</summary>`) | correct | correct | **structurally unreachable** — verified for C#, not assumed from java: tree-sitter places the comment outside the node span |
| `{` in a **preceding line comment** | correct | correct | unreachable, same reason |
| colon inside an attribute argument | correct | correct | control — attribute stripped before the `" :"` cut in both orders |
| colon in an attribute argument **+ brace** | `""` | `public class BraceColon` | fixed via the brace path (see trace) |
| method: `[HttpDelete("items/{id}")]` + control + expression body | correct | correct | sibling half, now pinned |

### Disclosed remainders — still broken, unchanged, verified identical before and after

- `{` in a comment **between** the attribute and the keyword: `[Route("a")] /* { */ public class MidCmt` → `"/*"`. Java disclosed the same remainder on #7124.
- `{` in a char or string literal in the declaration **header**: `record CharDefault(char C = '{')` → `"public record CharDefault(char C = '"`. Not an attribute, so the reorder cannot reach it.
- generic constraint: `class Gen<T> where T : class` → `"public class Gen<T> where T"` — the `" :"` cut firing on a `where` clause.
- a `]` inside an attribute argument string would end `stripCSharpAttributes`' scan early (same unhandled shape as java's `)`-in-literal).

## Both halves of the pair — this is the headline

With the new test file **held out** of the package, **all six** ordering mutants were **ALIVE** in `./internal/extractors/csharp/`. All six were re-scored from scratch on the amended suite rather than carried over:

| mutant | held out | with this PR |
|---|---|---|
| **a** — class order swapped back (brace cut first) | ALIVE, 0 `--- FAIL` | **DEAD, 2 `--- FAIL`** |
| **b** — `" :"` base-list cut removed | ALIVE, 0 `--- FAIL` | **DEAD, 2 `--- FAIL`** |
| **c** — **SIBLING** `buildMethodSignature` order swapped back | ALIVE, 0 `--- FAIL` | **DEAD, 1 `--- FAIL`** |
| **e** — class brace cut removed entirely (permissive direction) | ALIVE, 0 `--- FAIL` | **DEAD, 2 `--- FAIL`** |
| **f** — method `"=>"` expression-body cut removed | ALIVE, 0 `--- FAIL` | **DEAD, 1 `--- FAIL`** |
| **d** — collapse moved after the strip | ALIVE, 0 `--- FAIL` | **DEAD, 1 `--- FAIL`** |

The two pair-half counts reported separately, as asked: **class half 2 `--- FAIL`**, **method half 1 `--- FAIL`** — disjoint killers, the class mutants never touch the method test and vice versa. #7128's lesson held exactly: the *correct* sibling, the one whose own comment documents the hazard, was itself ungraded, so aligning the broken half alone would have left it free to regress silently. `buildMethodSignature` is **not changed** by this PR — its test is test-only cover.

Every mutant was `go vet`-ed before scoring (a non-compiling mutant exits non-zero with zero `--- FAIL` and reads as a kill), applied with an **asserted exact match count** in the mutator, run with `-count=1`, and restored by `cp` from a byte backup with the shasum verified — never `git checkout --`. The harness asserts its own worktree path and branch before touching anything.

## Fixtures were NOT compiled

There is no C# compiler on this machine — `csc`, `dotnet`, `mono`, `mcs` all absent. Syntax was settled against **ECMA-334 6th edition** and the Microsoft C# reference, cited in the test file header:

- attribute on a type parameter is grammatical: `type_parameter ::= attributes? identifier` (ECMA-334 §15.2.3; `AttributeTargets.GenericParameter` in the Microsoft "Attributes" reference). The fixture's `[AttributeUsage(AttributeTargets.All, AllowMultiple = true)]` permits it.
- `[property: ...]` on a record positional parameter is grammatical: `attribute_section ::= "[" attribute_target_specifier? attribute_list "]"` with `property` among the targets (ECMA-334 §22.3); the C# 9 records specification names that target for positional parameters, and it is the conventional placement for `System.Text.Json` attributes on a record.
- a line break may appear wherever whitespace may (ECMA-334 §6.3.1), so an attribute section may be followed by a newline rather than a space — the shape mutant d turns on.
- a record declaration may carry both a positional parameter list and a body (Microsoft C# 9 records specification, "Record declarations").
- a comment is trivia and may sit between an attribute and the declaration keyword (ECMA-334 §6.3.3).

Every expected string in the tests is a **transcription of the extractor's own measured output**, not a compiler's.

## Verification run here

- `go test -count=1 ./internal/extractors/csharp/` → **ok**, 0 `--- FAIL`
- `go test -count=1 ./cmd/grafel/` with a **fresh `GRAFEL_HOME` and a fresh `GRAFEL_DAEMON_ROOT`** → **ok, 188.8s**, 0 `--- FAIL`, **digest unmoved**. Stated plainly: `Signature` is **not** among the eight fields `entityTupleKey` hashes, so an unmoved digest is **regression cover for this change, not coverage of it** — `./cmd/grafel/` cannot observe a signature defect at all.
- `gofmt -l internal/extractors/csharp/` → clean
- `go vet ./internal/extractors/csharp/` → clean

The `./cmd/grafel/` run above was made on the first commit; the review round that follows it touches **test files only** (`internal/extractors/csharp/issue7133_class_signature_order_test.go`), so the extractor binary and therefore the digest are unchanged by it.

No full-repo suite run. Not merged — review first.


cajasmota:
**My own mutant, on the one ordering claim the fix round added and nobody scored. It is equivalent — and that makes the new comment's last sentence an overclaim, not a defect.**

The amended comment says:

> The cut stays BEFORE the `" :"` cut, as it was, so the base-list cut still sees a body-stripped string.

That reads as a constraint. Six mutants were scored against the strip/brace ordering and the two cuts' *existence*, but nobody permuted the **two cuts relative to each other**. So I did.

**DW-1: run the `" :"` cut before the brace cut.** Worktree at `5127f0f06`, `go vet` 0, exact match count 1, `-count=1`, restored by `cp` with shasum `bde99cb4…` and `git diff HEAD` empty.

```
go test -count=1 ./internal/extractors/csharp/   →  EXIT=0, 0 `--- FAIL`
```

**ALIVE.** And — given how today has gone on this exact function — I did not stop at the verdict. Both cuts truncate to a **prefix**, so the result is the prefix ending at the *earlier* of the two positions regardless of which runs first: `min(b, c)` either way. Enumerated over an alphabet of `{`, `" :"`, `:`, `}`, `<`, `>`, space and a letter, all strings to length 5 — **37,449 inputs, 0 differing.**

So DW-1 is genuinely equivalent, and I am recording it as equivalent with the argument *and* the enumeration attached, not as a kill and not as an ungraded hole.

**What that means for the comment.** The rest of the paragraph is correct and load-bearing — the brace cut must come after `stripCSharpAttributes`, and mutants (a) and (e) kill at 2 each to prove it. But the final clause implies the brace-before-colon ordering is also load-bearing, and it is not: the base-list cut does **not** need to see a body-stripped string, because truncating to the earlier of two positions is order-independent.

That is a one-clause fix, not a blocker: say the two cuts are mutually order-independent (both truncate to a prefix) and that the constraint is only *strip before brace cut*. It is worth doing precisely because this function has now produced **four** wrong mechanism claims in one day — two of mine, one equivalence I relayed, and the reviewer's catch on (d) — and the pattern in every one is a sentence asserting a causal relationship that nothing observes. Leaving a fourth in place, in a comment written specifically to explain the ordering, invites the next person to preserve a constraint that does not exist.

**On the fix round itself: confirmed and good.** Mutant (d) is DEAD at 1 `--- FAIL` on exactly the two new rows, all six mutants were re-scored **from scratch** on the amended suite rather than carried over (a 2, b 2, c 1, d 1, e 2, f 1), and the false "equivalent on any input" sentence is gone from both the test header and the `MultiLine` comment — including the secondary correction that `MultiLine` never graded the ordering at all, since its line break is inside the parens while the orderings disagree about the text after `]`.

Worth noting for the record: the author caught an error in their own new comments **before committing** — they had written `"" before the fix` for both new rows, measured against the pre-fix file, found the real values were `"public class TPAttrNewline<"` and `"public record PersonRec("` (truncated, not emptied), and corrected it. That is the check that would have caught three of the four wrong claims above.

`closure-keyword` is green on the latest run (`105184633424`, 11:30:28Z). The earlier failure on this same sha (`105184472168`, 11:29:57Z) is the `synchronize` event firing on the push **before** the body edit; the `edited` event re-ran it against the fresh payload. Latest-per-name is the correct read.

PR #7142, opened 2026-09-17
fix(#7136): the F# letSeen dedup key was two independent ":let:" literals, and a one-sided change to either was invisible

### LINKED ISSUE #7136: MEASURED: the F# letSeen dedup key is built from two independent ":let:" literals (extractor.go:340 and :385) — mutating only the writer is ALIVE at 0 FAIL, so the let/member cross-dedupe can silently stop working
## What

The F# extractor dedups `let` bindings against members using a composite key, `indent + ":let:" + name`. That key is constructed from **two independent string literals at two sites**:

- `internal/extractors/fsharp/extractor.go:340` — the let-binding scanner, which *writes* the key into `letSeen`
- `internal/extractors/fsharp/extractor.go:385` — the member loop, which *reads* it to skip a member already emitted as a let

Their agreement is what makes the cross-dedupe work, and **nothing observes that they agree.**

## Measured

Mutating **only** site 340's literal from `":let:"` to `":LET:"`:

```
go test -count=1 ./internal/extractors/fsharp/   →  ALIVE, 0 `--- FAIL`
```

So the writer and the reader can silently disagree, the skip stops firing, and the suite stays green. The visible consequence is a duplicate entity — the same binding emitted once as a `let` operation and again by the member loop — which is precisely the class of defect `letSeen` exists to prevent.

## Why this is the score-every-occurrence shape, not a representative-site one

This is the recurring failure mode on this board: a DEAD verdict on one half of a twinned key says nothing about the other half. A test that exercises the dedupe **through** both sites passes as long as both literals are *whatever they both are* — it grades the mechanism, never the agreement. Only a mutant applied to **one** site can see it, and only if something asserts the outcome.

Note the asymmetry: mutating **both** sites together is almost certainly equivalent-under-the-suite (the key is internal, so a consistent rename is invisible by design). That is correct and not a defect. The defect is that a **one-sided** change is also invisible.

## Suggested shape

Extract the key construction into one function (`letSeenKey(indent, name string) string`) so there is a single literal and the two sites cannot drift. That removes the defect by construction rather than by assertion, which is the right move for an agreement between two call sites.

**But extraction alone does not pin it** — the standing lesson from #6533: extracting a decision and unit-testing the helper leaves the call sites free to ignore it. So the change also needs a test that observes the **consequence**: a fixture where a binding is reachable by both the let scanner and the member loop, asserting the entity **count** (not just that the names are right), so a failed skip shows up as a duplicate. Then mutate the single remaining literal and confirm the count assertion is what dies.

## Scope and provenance

Narrow, one file, and independent of #7131/#7134 (which fixed the modifier handling that produces the `name` half of this key) and of #7135 (the sibling scanners). It can be done in any order relative to those.

Found by the independent reviewer on #7134 as an extra mutant beyond the brief — the PR itself is sound and this is pre-existing, so it was correctly kept out of that change.

cajasmota:
Closes #7136

## The defect

The F# extractor's cross-dedupe — the thing that stops a member being indexed a
second time when the `let` scanner already emitted a binding of the same
name at the same indent — is an **agreement between two sites**:

- the let loop **writes** `indent + ":let:" + name` into `letSeen`
- the member loop **reads** the same string back to decide to skip

Each built the key from **its own string literal**, and nothing observed that
the two matched.

### Red baseline (measured on the parent commit, `c1e2c3008`)

Changing **only** the writer's literal to `":LET:"` (1 asserted match):

```
go vet ./internal/extractors/fsharp/    → clean
go test -count=1 ./internal/extractors/fsharp/
  exit=0   --- FAIL lines = 0      → ALIVE
```

The skip stops firing and the binding is emitted twice — once as a `let`
operation, once as a `member` — with the suite green.

## The fix

Both sites now go through one function:

```go
func letSeenKey(indent, name string) string {
	return indent + ":let:" + name
}
```

One literal exists, so the two sites cannot drift. That removes the defect **by
construction**, which is the right shape for an agreement between call sites.

## Why a COUNT and not names

Extraction alone does not pin it — the standing #6533 lesson: unit-testing the
helper grades the helper, and leaves either call site free to go back to
building the string inline. So the new test observes the **consequence**.

The two entities a failed skip produces **carry the same name** (they differ
only in subtype, `let` vs `member`). So "is there an operation called
`Compute`?" passes with the duplicate present. Only the count sees it.
`fsOpCountNamed` therefore deliberately does **not** filter on subtype.

## Mutant scores (all `go vet`-clean before scoring, exact match count asserted, `-count=1`, restored by `cp` from a byte backup with shasum **and** `git diff HEAD` verified)

| # | mutant | verdict | `--- FAIL` | which assertion died |
|---|---|---|---|---|
| a | the single remaining `":let:"` literal → `":LET:"` | **ALIVE — equivalent, on purpose** | 0 | none (see below) |
| b | **reader** site reverted to inline `indent+":LET:"+name` | **DEAD** | 2 | the count, in both `…MemberSharingLetNameIsDeduped` and `…StaticMemberSharingLetNameIsDeduped` |
| b2 | **writer** site reverted to inline `indent + ":LET:" + name` | **DEAD** | 2 | same two count assertions |
| d | `indent` dropped from the key | **DEAD** | 1 | `…DistinctIndentIsNotDeduped` (the over-broad direction) |
| c | both count assertions weakened to presence-only (`len(...) < 1`) | ALIVE alone (0 FAIL) | 0 | — |
| c+b | **c applied together with b** | **ALIVE, 0 FAIL** | 0 | nothing — proves the count is the load-bearing assertion |

Row **c+b** is the point: with the assertion weakened to presence-only, the very
drift mutant that row b kills survives at 0 `--- FAIL`. The count, not the name,
is what grades the dedupe.

Both call sites were scored (b and b2), not one representative site.

### Why (a) is ALIVE-and-equivalent rather than a hole

The brief predicted (a) would die. It **cannot**, and that is now correct
behaviour rather than a gap: after the fix there is exactly one literal, so
changing it moves the writer and the reader **together**. A consistent rename of
an internal key is invisible by design — the same asymmetry the issue calls out
for the two-sided case, which after this change is the *only* case that exists.
A one-sided change is what used to be invisible, and rows b/b2 are exactly that
change; both now die.

Equivalence argument, from the code's alphabet: the key is a key in a
function-local `map[string]bool` that is never emitted, never serialised and
never compared to anything but another key from the same builder (`grep`
confirms `letSeen` is touched at exactly four sites, all via `letSeenKey`).
`indent` is `[ \t]*` and `name` is `[a-zA-Z_][a-zA-Z0-9_']*` (capture groups 1
and 2 of `letRE`/`memberRE`), so **neither can contain `:`** and the separator
can never be forged out of the operands.

Enumeration backing it: over 558 `(indent, name)` pairs drawn from those exact
alphabets (6 indents × names up to length 3 over `aA_0'`), `":let:"` and
`":LET:"` are **both injective** (558 distinct keys each) and induce an
**identical partition** of the input space; 0 of the 558 inputs contain a `:`,
i.e. no input exists that could distinguish them. It is left untested **on
purpose**.

## Siblings found and deliberately NOT touched

I scanned every composite key and sentinel in `internal/extractors/fsharp/`.
**None has the twinned-literal shape.** Each of the others is built once, into a
local `key` variable that both the read and the write then use:

- `memberSeen` — `indent + ":member:" + name`, **one** site (`extractor.go:501`).
  #7141 (merged) added that literal and it has no twin. Explicitly checked,
  because this is the pair the issue's shape most invites.
- `"module:" + name` (`:398`) and `"namespace:" + name` (`:424`) — one site each,
  sharing one `seen` map under distinct prefixes.
- `kind + ":" + target` (`:1038`) — one site.
- `typeSeen`, and the `seen`/`caseSeen` maps in `compexpr_active_patterns.go`,
  `du_record_members.go`, `elmish_feliz.go`, `field_validations.go`,
  `validators.go` — all keyed on a bare name, no composite.

So the pair audit is: `letSeen` (fixed here) and `memberSeen` (checked, single
literal, nothing to fix). Nothing to file.

## Digest

`go test -count=1 ./cmd/grafel/` with a **fresh `GRAFEL_HOME` and a fresh
`GRAFEL_DAEMON_ROOT`** → `ok ... 185.546s`, exit 0, **0 `--- FAIL`**. The digest
is **unmoved**, as expected: I enumerated the digest fixture
(`writeSpanFixture6118`) and it contains **no `.fs`/`.fsx` file** — only `.cs`,
`.java`, `.py`, `.rs`, `.go`, `.ts`, `.js`. So this run is **regression cover,
not coverage** of the change: it proves nothing about the F# path, only that the
graph elsewhere did not move. The pin stayed at 140 entities / 282 relationships
and no constant was bumped.

The five digest gates were run **by name**, to prove they were not skipped:

```
--- PASS: TestCustomExtractorGateDoesNotRegressFileComponentStartLines (1.91s)
--- PASS: TestCustomExtractorGateOffGraphIsUnchanged (0.19s)
--- PASS: TestSemanticDigestGradesEntityLanguage6862 (0.19s)
--- PASS: TestCustomExtractorGateOffDeltaIsExactlyTheDocumentedSpanGain (0.20s)
--- PASS: TestFileComponentSpanDonationIsDeterministic (0.42s)
```

## Verification

```
gofmt -l internal/extractors/fsharp/ cmd/grafel/        → (no output)
go vet ./internal/extractors/fsharp/                    → clean
go test -count=1 ./internal/extractors/fsharp/          → ok  0.583s   (0 --- FAIL)
go test -count=1 ./cmd/grafel/  (fresh GRAFEL_HOME + GRAFEL_DAEMON_ROOT)
                                                        → ok  185.546s (0 --- FAIL)
```

No full-repo suite was run.

## What I could NOT verify

**No F# toolchain exists in this environment** (`dotnet`, `fsc`, `fsharpc`,
`mono` all absent), so whether a compiler accepts a `let` and a `member` sharing
the identifier `Compute` inside one type is read off the F# specification (§ 8.6
class type definitions: `let`-bindings are the private implementation, `member`s
the public surface) and is **derived-not-executed**. That question is also
immaterial to what is under test: this is a deliberately lenient line scanner
that reaches both lines regardless, which is precisely the condition the
cross-dedupe exists to handle. The fixture header says so on its own line.


cajasmota:
**Merge gate: independent review APPROVE with every number reproduced, and my own mutants close the one surface the extraction created.**

## My mutants — argument-order drift, the surface this fix introduced

Extracting into `letSeenKey(indent, name string)` removed the twinned **literal**, but it created a new twinned surface in its place: **two call sites passing two same-typed `string` arguments.** Go's type system cannot catch a swap. That is the shape worth attacking, and nobody had.

Worktree at `c4ff4f361`, `go vet` 0 before each, exact match count 1, `-count=1`, restored by `cp` with shasum `fac184b2…` and `git status` clean.

| mutant | site | `--- FAIL` |
|---|---|---|
| **DZ-1** — `letSeenKey(name, indent)` | reader (`:498`) | **2 — DEAD** |
| **DZ-2** — `letSeenKey(name, indent)` | writer (`:453`) | **2 — DEAD** |

Both killed by the same pair of count assertions (`…MemberSharingLetNameIsDeduped`, `…StaticMemberSharingLetNameIsDeduped`). I scored **both** sites rather than one, because a twinned surface graded at a representative site tells you nothing about its twin — that is the standing lesson, and it is the lesson this very PR exists to apply to the literal.

So the extraction is clean: it removed a drift surface without leaving an ungraded one behind. Worth stating explicitly, because "extract into a helper" is the reflex fix for a twinned literal and it silently converts a *textual* divergence into a *positional* one.

## Review outcome — APPROVE, and it re-derived rather than re-applied

Both per-site reverts scored **separately**: reader inline DEAD 2, writer inline DEAD 2. Plus their own `if false` mutant on the skip, DEAD 2.

The compound proof reproduced: both `!= 1` assertions weakened to `< 1` **plus** the drift mutant gives **ALIVE, 0 FAIL, vet clean** — and they identified *why* the count is the sole killer, which the body did not spell out: test 1's `fsFindLet` guard survives the drift, because the `let` entity is still present alongside the duplicate. So the count is genuinely load-bearing rather than incidentally redundant with the presence check.

The single-literal equivalence is upheld on a **wider** enumeration than the author's — 1395 pairs against 558, indents to length 3 over `{space, tab}`, across separators `":LET:"`, `""`, `" "`, `"\t"`, `"::"`, `"x"`, `"'"` — all injective, all inducing the identical partition, with the structural reason attached: `indent` is recoverable as the maximal whitespace prefix minus the separator's fixed prefix, so **any** fixed separator is injective. That is a stronger statement than the one I asked for.

Sibling audit extended to **five** keys rather than the three I asked for, all single-site. No twinned key anywhere in the package.

And they measured the prose against the parent rather than accepting it: the helper comment claims a one-sided change to **either** site left the package green, while the PR body had only measured the writer. They measured both on `c1e2c3008` — writer 0 FAIL, reader 0 FAIL — so the comment is true as written.

## Fixture legality settled empirically, not argued

No F# toolchain here, so legality is derived-not-executed. Rather than leave it there, they probed a shape that is unambiguously legal and needs **no** in-one-type collision — `module A =` with a 4-space `let Compute`, plus a separate `type T() =` with a 4-space `member this.Compute` — and got count=1, subtypes=`[let]`. So the dedupe path this test pins is reachable from legal source, and the fabricated-shape trap does not bite.

## That probe also found a real pre-existing defect → #7144

The same probe shows the collapse is **wrong** in that case: `A.Compute` and `T.Compute` are two distinct operations in unrelated enclosing scopes, and one is silently dropped, because the key is `indent + name` with **no enclosing scope**. Over-broad direction, invisible to recall assertions. Pre-existing, orthogonal to this change, and filed as **#7144** with the two things a fix must establish first — whether the scanner has the enclosing construct available at all, and that `DistinctIndentIsNotDeduped` must stay green, so indentation cannot simply be dropped.

Two non-blocking nits from the review, both accurate: the PR body transcribes the weakened assertion as `len(...) < 1` where the code is `fsOpCountNamed(...) != 1` (same semantics, cosmetic), and `…StaticMemberSharingLetNameIsDeduped` lacks test 1's `fsFindLet` guard, so an inversion keeping the `member` and dropping the `let` would still read count=1 there — test 1 covers that axis.

Merging once the last check lands.

PR #7147, opened 2026-09-17
fix(#7133): kotlin's signature builders cut at their delimiters BEFORE stripping annotations — BOTH halves, and the fun half broke with no brace at all

### LINKED ISSUE #7133: KOTLIN ARM ONLY: kotlin.go:1112 buildClassSignature has the same inverted annotation/brace order — '@Table(name = "{weird}") class Foo' yields an EMPTY signature (C# arm shipped in 84964094e; where-clause split to #7139)

cajasmota:
`buildClassSignature` cut the declaration at the first `{` **before** stripping annotations. A brace inside an annotation argument string is indistinguishable from the body brace to `strings.Index`, so the cut landed inside the literal, `stripKotlinAnnotations` then found no `)` matching the now-unbalanced `(` and consumed the remainder to end-of-string, and the emitted signature was **EMPTY**.

## Red output (this exact fixture, against `99bb52790`)

```
--- FAIL: TestKotlin7133_ClassSignatureStripsAnnotationsBeforeBraceCut (0.01s)
    Foo.Signature = "", want "class Foo"
    BraceItems.Signature = "", want "class BraceItems"
    IFace.Signature = "", want "interface IFace"
    Obj.Signature = "", want "object Obj"
    Data.Signature = "", want "data class Data(val a: Int)"
    Colors.Signature = "", want "enum class Colors"
    Sealed.Signature = "", want "sealed class Sealed"
    AnnoCls.Signature = "", want "annotation class AnnoCls"
    WithBase.Signature = "", want "class WithBase : Any()"
    Stacked.Signature = "", want "class Stacked"
    MultiLine.Signature = "", want "class MultiLine"
    CtorParam.Signature = "class CtorParam(", want "class CtorParam(val a: String)"
    CtorNL.Signature = "class CtorNL(", want "class CtorNL(val a: String)"
    TP.Signature = "class TP<", want "class TP<T>(val t: T)"
    TPNL.Signature = "class TPNL<", want "class TPNL<T>(val t: T)"
    Outer.Signature = "", want "class Outer"
    Inner.Signature = "", want "class Inner"
--- FAIL: TestKotlin7133_FunSignatureStripsAnnotationsBeforeBodyCuts (0.00s)
    topFun.Signature = "@Table(name", want "fun topFun(a: Int): Int"
    plainFun.Signature = "@Table(name", want "fun plainFun(a: Int): Int"
    spaceBraceFun.Signature = "@Ann(\"a", want "fun spaceBraceFun(a: Int): Int"
    Holder.namedArg.Signature = "@Table(name", want "fun namedArg(b: String): String"
```

17 of 23 class rows and 4 of 11 fun rows redden; 6 + 7 control rows and all 10 remainder rows stay green. (An earlier revision of this body said "4 of 10" and "6 + 6"; both were wrong. Re-measured against the pre-change file after the review round added two fun rows, and now recorded in the test file's own header so the two cannot drift apart again.)

## Pair or single site? A PAIR — and BOTH halves were inverted

The issue expected a single site. There are two, and this is where kotlin diverges from java (#7128) and C# (#7138): in both of those the sibling ran the strip first and carried a comment naming the hazard, so the pair audit only had to *grade* a correct function. Kotlin's `buildFunSignature` has no such comment and ran its cuts first too.

Its damage is narrower on the brace and much wider overall:

- **narrower on the brace** — it cuts at `" {"` (space-brace), so `@Table(name = "{weird}")` does not reach it: the character before the brace is a quote. `@Ann("a {b}")` does, and recorded as `@Ann("a`.
- **wider in general** — it also cuts at `" ="`, which **every named annotation argument contains**. No brace is needed anywhere:

  ```
  @Table(name = "weird") fun plainFun(a: Int): Int  ->  "@Table(name"
  ```

  That is the majority shape for an annotated Kotlin function, and it was losing the whole declaration and emitting the annotation instead.

So the "is the absence graded" question did not arise the way the brief anticipated — there was no correct sibling to grade, there was a second defect. Both are fixed here, and the two halves are graded by separate tests with **disjoint killers**.

## Ordering chosen, and why

`collapse → strip → cuts`, for both builders: the **cuts moved last** rather than the strip moved first. That is the smaller change — it keeps `stripKotlinAnnotations` on the single-line, single-spaced input class it has always received — and, unlike the reverse, it is also the **correct** one.

I read `stripKotlinAnnotations`'s own alphabet rather than carrying the C# reasoning across: it branches on exactly `@`, `[A-Za-z0-9_]`, `(`, `)` and the plain **space** `' '`. Its post-annotation skip therefore does not fire on a newline, so stripping before the collapse leaves an interior space the collapse can no longer remove. Two ordinary-Kotlin shapes pin it, and both are in the fixture:

```
class CtorNL(@Ann("{x}")\n    val a: String)
  collapse-then-strip -> "class CtorNL(val a: String)"    (chosen)
  strip-then-collapse -> "class CtorNL( val a: String)"   (mutant d)

class TPNL<@Ann("{v}")\n    T>(val t: T)
  collapse-then-strip -> "class TPNL<T>(val t: T)"
  strip-then-collapse -> "class TPNL< T>(val t: T)"
```

Same conclusion as the C# arm's corrected version: collapse-first is right because it is *correct*, not because the alternative is indistinguishable.

**VARIED (axis 1, the defect axis):** whether the annotation argument's string literal contains `{`. `Foo`/`Plain` and `BraceItems`/`PlainItems` are pairs whose only deliberate difference is the two brace characters. Their type names necessarily differ too — two types cannot share a name in one file — and that difference is inert: `Plain` carries the same annotation minus the braces and is green in both orders.

**VARIED (axis 2, independently):** declaration kind — class / interface / object / data class / enum class / sealed class / annotation class. All seven reach `buildClassSignature`; tree-sitter-kotlin maps interface, enum class and data class onto `class_declaration` and kotlin.go discriminates by child inspection, so there is no separate enum builder as there is in C#.

**VARIED (axis 3, independently):** annotation **position** — on the type / stacked on the type / spanning a line break / on a constructor parameter / on a type parameter / on a nested type — plus whether the declaration carries a supertype list.

**HELD CONSTANT:** file, package, annotation types (`Table`/`Ann`), argument shape (one named `String` argument), absence of modifiers other than the declaration keyword, the empty-or-one-member body, and the top-level enclosing scope.

**Controls** (green before *and* after, so the defect rows grade the ORDER and not the presence of a non-empty string) — class: `Plain`, `PlainItems`, `Bare`, `TypeParamNoAnn`, `Table`, `Ann` (6 of 23); fun: `ctrlFun`, `bareFun`, `exprBody`, `exprAnn`, `Holder.member`, `paramAnno`, `paramAnnoCtl` (7 of 11).

## Enumeration table

| shape | before | after |
|---|---|---|
| `@Table(name = "{weird}") class Foo` | `""` | `class Foo` |
| `@Table(name = "weird") class Plain` (control) | `class Plain` | `class Plain` |
| `@Table(name = "{w}") interface IFace` | `""` | `interface IFace` |
| `@Table(name = "{w}") object Obj` | `""` | `object Obj` |
| `@Table(name = "{w}") data class Data(val a: Int)` | `""` | `data class Data(val a: Int)` |
| `@Table(name = "{w}") enum class Colors { ... }` | `""` | `enum class Colors` |
| `@Table(name = "{w}") sealed class Sealed` | `""` | `sealed class Sealed` |
| `@Table(name = "{w}") annotation class AnnoCls` | `""` | `annotation class AnnoCls` |
| `@Table(name = "{w}") class WithBase : Any()` | `""` | `class WithBase : Any()` |
| `@Ann @Table(name = "{w}") class Stacked` | `""` | `class Stacked` |
| annotation spanning a line break (`MultiLine`) | `""` | `class MultiLine` |
| `class CtorParam(@Ann("{x}") val a: String)` | `class CtorParam(` | `class CtorParam(val a: String)` |
| `class CtorNL(@Ann("{x}")\n val a: String)` | `class CtorNL(` | `class CtorNL(val a: String)` |
| `class TP<@Ann("{v}") T>(val t: T)` | `class TP<` | `class TP<T>(val t: T)` |
| `class TPNL<@Ann("{v}")\n T>(val t: T)` | `class TPNL<` | `class TPNL<T>(val t: T)` |
| nested: `Outer` / `Inner` | `""` / `""` | `class Outer` / `class Inner` |
| `@Table(name = "{weird}") fun topFun(...)` | `@Table(name` | `fun topFun(a: Int): Int` |
| `@Table(name = "weird") fun plainFun(...)` — **no brace** | `@Table(name` | `fun plainFun(a: Int): Int` |
| `@Ann("a {b}") fun spaceBraceFun(...)` | `@Ann("a` | `fun spaceBraceFun(a: Int): Int` |
| `@Table(name = "{w}") fun Holder.namedArg(...)` | `@Table(name` | `fun namedArg(b: String): String` |
| `fun paramAnno(@Ann("{x}") a: Int): Int = a` (control) | correct | correct |
| `@Ann("ab") fun ctrlFun(...)` (control) | correct | correct |
| `fun exprBody(a: Int): Int = a` (control) | correct | correct |
| `@Ann("{x}") fun exprAnn(a: Int): Int = a` (control) | correct | correct |

### Disclosed remainders — measured byte-identical before and after, pinned by `TestKotlin7133_DisclosedRemainders`

| shape | records as | why it is not this arm's |
|---|---|---|
| `class DefStr(val s: String = "{x}")` | `class DefStr(val s: String = "` | brace is in a **header** literal, not an annotation; no reordering can help. Same remainder the C# arm disclosed. |
| `class DefChar(val c: Char = '{')` | `class DefChar(val c: Char = '` | same |
| `@Ann("x") /* { */ class BlockCmt` | `/*` | a comment **after an annotation** is inside the node span, because the annotation is |
| `@Ann("x") // { here` + `class LineCmt` | `//` | same |
| `/** doc { brace */ class KDocOnly` | `class KDocOnly` | **unreachable** — a preceding doc comment is *outside* the node span |
| `// line { comment` + `class LineOnly` | `class LineOnly` | same |
| `class CtorGet(@get:Ann("{w}") val a: Int)` | `class CtorGet(:Ann("` | `stripKotlinAnnotations` skips `@` then identifier chars then expects `(`; a `:` ends the scan and it falls through. A helper defect, not an ordering one. |
| `class CtorGetCtl(@get:Ann("w") val a: Int)` | `class CtorGetCtl(:Ann("w") val a: Int)` | the same helper defect without the truncation layered on top |
| `fun defaultArg(a: Int = 3): Int` | `fun defaultArg(a: Int` | the `" ="` cut on a **default parameter value**; annotation-independent, unaffected by the order |
| `fun paramSpaceBrace(@Ann("a {b}") a: Int)` | `fun paramSpaceBrace(@Ann("a` | the `" {"` cut inside a **parameter** annotation, which is kept by design; see the `paramAnno` guard below |

**Comment reachability, verified for kotlin rather than carried over:** a KDoc or line comment *preceding the whole declaration* is outside the node span and is structurally unreachable — same verdict as java and C#, and `KDocOnly`/`LineOnly` are the demonstration. But a comment sitting *between an annotation and the keyword* **is** inside the span, because the annotation is, so a `{` there does reach the cut. **That is a kotlin-specific difference from the other two arms**, and it is why verifying rather than assuming mattered a third time.

## Mutant scores

`go vet` before every score, exact match count asserted by the mutation script (which aborts unless it is running in this worktree on this branch), `-count=1`, restored by `cp` from a byte backup with the shasum and `git diff HEAD` verified between rounds. `--- FAIL` lines counted, never the exit code.

**With `issue7133_signature_order_test.go` held out, on the fixed tree — baseline 0 `--- FAIL`:**

| mutant | half | verdict | `--- FAIL` |
|---|---|---|---|
| (a) class order swapped back to HEAD | class | **ALIVE** | 0 |
| (a2) brace cut between collapse and strip | class | **ALIVE** | 0 |
| (b) class brace cut REMOVED | class | **ALIVE** | 0 |
| (c) class annotation strip REMOVED | class | **ALIVE** | 0 |
| (d) class collapse moved AFTER strip | class | **ALIVE** | 0 |
| (e) fun order swapped back to HEAD | fun | **ALIVE** | 0 |
| (f) fun `" {"` cut REMOVED | fun | **ALIVE** | 0 |
| (g) fun `" ="` cut REMOVED | fun | **ALIVE** | 0 |
| (h) fun annotation strip REMOVED | fun | **ALIVE** | 0 |
| (i) fun collapse moved AFTER strip | fun | **ALIVE** | 0 |
| (j) fun `" {"` cut relaxed to `"{"` | fun | **ALIVE** | 0 |

**All eleven ALIVE** — including *deleting the brace cut outright* and *deleting the annotation strip outright*. Nothing in `./internal/extractors/kotlin/` observed either builder's ordering, so this repeats #7128's and #7138's result in a third language, and goes further: neither half was even minimally graded.

**With the new test file present:**

| mutant | verdict | `--- FAIL` | killers |
|---|---|---|---|
| (a) | DEAD | 1 | `…ClassSignatureStripsAnnotationsBeforeBraceCut` |
| (a2) | DEAD | 1 | `…ClassSignatureStripsAnnotationsBeforeBraceCut` |
| (b) | DEAD | 2 | `…ClassSignature…`, `…DisclosedRemainders` |
| (c) | DEAD | 2 | `…ClassSignature…`, `…DisclosedRemainders` |
| (d) | DEAD | 1 | `…ClassSignatureStripsAnnotationsBeforeBraceCut` |
| (e) | DEAD | 1 | `…FunSignatureStripsAnnotationsBeforeBodyCuts` |
| (f) | DEAD | 2 | `…DisclosedRemainders`, `…FunSignature…` |
| (g) | DEAD | 2 | `…DisclosedRemainders`, `…FunSignature…` |
| (h) | DEAD | 1 | `…FunSignatureStripsAnnotationsBeforeBodyCuts` |
| (i) | **ALIVE** | 0 | — |
| (j) | DEAD | 1 | `…FunSignatureStripsAnnotationsBeforeBodyCuts` |

10 of 11 DEAD, and the killers are **disjoint by half**: no class mutant is caught by the fun test and no fun mutant by the class test, so these are two observations rather than one.

### Mutant (j): the fun half's `" {"` space guard — found in review, now DEAD

`buildFunSignature` strips only the prefix before `fun `, so a **parameter** annotation is kept by design and its content reaches the body cut. The single space in `" {"` is the only thing keeping the common unspaced shape out of it — and that guard was itself ungraded. Relaxing the cut to `"{"` was **ALIVE, 0 `--- FAIL`**, and is distinguishable:

```
fun paramAnno(@Ann("{x}") a: Int): Int = a
  shipped -> fun paramAnno(@Ann("{x}") a: Int): Int
  mutant  -> fun paramAnno(@Ann("
```

Two rows added — `paramAnno` and its brace-free control `paramAnnoCtl`. The mutant is now **DEAD, 1 `--- FAIL`**, killed by the fun test. Pre-existing guard, untouched by the reorder; this is cover, not a behaviour change. Its other side is disclosed as remainder (5) above: when the brace *does* follow a space inside a parameter annotation, the cut still fires.

Found by the independent reviewer on this PR. All eleven mutants were **re-scored from scratch** rather than carrying counts over, which earned its keep: adding the `paramSpaceBrace` remainder row moved mutant (f) from 1 `--- FAIL` to 2.

### Mutant (i): ALIVE, and recorded as a HOLE — deliberately **not** as an equivalence

The fun half's collapse-after-strip mutant survives. It would be convenient to call it equivalent by the same alphabet argument, and that is exactly the move the C# arm had to retract. So I tested it instead.

The two pipelines **are** distinguishable. Enumerating prefixes over `{@, A, (, ), space, newline, {, =, x, ", :}` to length 4, restricted to shapes where every `@` is followed by an identifier character, and appending a real function head: **48 of 35 271 inputs differ**. Every one of them needs a character sitting *immediately adjacent* to the `@` with no whitespace, plus a line break after the annotation — e.g. `A@A\nfun g()` → `"Afun g()"` vs `"A fun g()"`.

I could not construct a Kotlin source shape that reaches that set. Measured byte-identical under the mutant through the real extractor: `public@Ann("x")\nfun` (tree-sitter accepts it), `@Ann("x")@B\nfun`, `@Ann("x")\npublic fun`, `@Ann("x")\n\tinline fun`, and an annotated member function. A separate realistic sweep — 3 264 combinations of well-formed annotations × whitespace × modifiers × declaration tails — found 0 differences.

So it is left **ungraded on purpose for want of a reaching fixture**, and it is *not* recorded as equivalent: the mutant provably changes output on some input, I just could not get Kotlin to produce one. Recording it as equivalent would retire a real hole.

## Verification

- `go test -count=1 ./internal/extractors/kotlin/` → **ok** (full package, not just the new tests), re-run after the review round
- The review round is **test-only**: `kotlin.go` is byte-identical to the commit already verified against `./cmd/grafel/`, so that ~4-minute run is not repeated for it
- `go test -count=1 ./cmd/grafel/` with a **fresh `GRAFEL_HOME` and a fresh `GRAFEL_DAEMON_ROOT`** → **ok, 234.208s, 0 `--- FAIL`**
- Digest/parity gates re-run **by name** to prove they were not skipped — all PASS:
  `TestSemanticDigestGradesEntityLanguage6862`, `TestContentParity_PathA_6129`, `TestContentParity_PathB_6129`, `TestIncrementalIndexPathB_ParityWithFullReindex`, `TestWorktreeSeedParity_SeededGraphMatchesAFullIndex`
- **Does the `cmd/grafel` fixture corpus contain kotlin?** Enumerated, not assumed: **exactly one `.kt` file**, `cmd/grafel/testdata/spring_app/src/main/kotlin/com/example/ChannelRegistry.kt`. It is reached, and its signatures are unchanged by this commit (traced: `@Service class ChannelRegistry`, `register`, `unregister`, `getClients` all produce identical output in both orders). `Signature` is not among the eight fields `entityTupleKey` hashes, so an unmoved digest is the expected result either way — regression cover, not coverage.
- `gofmt -l internal/extractors/kotlin/` → empty; `go vet ./internal/extractors/kotlin/` → clean

## One fixture was not legal Kotlin — corrected in review

The use-site-target remainder was originally pinned on `@get:Ann("{w}") class UseSite`. **That is not legal Kotlin** — `get:` targets a property's getter and a class declaration has none — and the reviewer was right to call it. Nothing depended on its legality, and every mutant around it scored DEAD, which is precisely the point: **an all-DEAD score cannot detect a fabricated shape.** Only reading the fixture against the language could.

It is replaced by the shape kotlinlang.org uses in its own example ("Annotations" → "Annotation use-site targets": `class Example(@field:Ann val foo, @get:Ann val bar, @param:Ann val quux)`) — a constructor `val` parameter, where `get:` targets the generated getter:

```
class CtorGet(@get:Ann("{w}") val a: Int)   ->  class CtorGet(:Ann("
class CtorGetCtl(@get:Ann("w") val a: Int)  ->  class CtorGetCtl(:Ann("w") val a: Int)
```

Same helper defect, on source that can exist, plus a control that shows the defect without the truncation layered on top. The fixture's annotation now carries an explicit `@Target(AnnotationTarget.VALUE_PARAMETER, AnnotationTarget.PROPERTY_GETTER)`.

## Kotlin compiler

**There is none on this machine** — `kotlinc` and `kotlin` are both absent (`javac` is present, but it does not compile `.kt`). **The fixtures were not compiled.** Syntax is settled from the Kotlin language specification's syntax grammar (`typeParameter` / `typeParameterModifiers` for the annotation-on-a-type-parameter rows, `classParameter` for the annotation-on-a-constructor-parameter rows) and from kotlinlang.org (Classes, Sealed classes, Data classes, Enum classes, Annotations — including that an annotation with no explicit `@Target` is applicable to all targets bar a small excluded set, which is why `@Target` is stated explicitly in the fixture). Every citation is in the test file's header comment and marked **derived-not-executed**.

Everything asserted in the tests is the extractor's own observed output, not a compiler's: every expectation is a transcription of a measured run, and every "before" value quoted here and in the test comments was **measured against the pre-change file**, not predicted. One comment was corrected during this work for exactly that reason — `exprAnn` is a control, not a defect row, because the brace in `@Ann("{x}")` is preceded by a quote and so never reaches the `" {"` cut.

Closes #7133


cajasmota:
**My own mutant found one more ungraded guard, in the same family one level down — and it is distinguishable, not equivalent.**

Your fun half cuts at **`" {"`** (space-brace), not a bare `{`. That space requirement is what stops a brace inside a **parameter** annotation from truncating — and parameter annotations are deliberately kept intact, since the strip only applies to the prefix before `fun `. Nobody had scored it.

**EC-1: `strings.Index(raw, " {")` → `strings.Index(raw, "{")`** on the fun half only. Worktree at `f04c6b0d6`, `go vet` 0, `-count=1`, restored by `cp` with shasum `648af88b…` and `git diff HEAD` empty.

```
go test -count=1 ./internal/extractors/kotlin/   →  EXIT=0, 0 `--- FAIL`
```

**ALIVE.** And distinguishable — probed against both trees rather than reasoned about:

| | `fun paramAnno(@Ann("{x}") a: Int): Int = a` |
|---|---|
| shipped | `fun paramAnno(@Ann("{x}") a: Int): Int` |
| EC-1 | `fun paramAnno(@Ann("` |

So the space-brace requirement is load-bearing and unobserved. It is exactly the defect this PR fixes, one level down: the class half's brace lives in a *prefix* annotation and is handled by stripping; the fun half's surviving brace lives in a *parameter* annotation that is kept on purpose, and only the space stops it truncating.

**What I would like: one fixture row, no production change.** A `fun` with a brace inside a parameter annotation, asserting the full signature. Then confirm EC-1 goes DEAD and report its count. Behaviour is correct today — this is a pre-existing guard your reordering did not touch, and I am asking here only because the branch is open, the fixture table already exists, and CI has to re-run anyway.

One thing I noticed while probing, **not** an ask: `fun defaultVal(a: String = "{y}"): String` records as `fun defaultVal(a: String` on both trees, because the `" ="` cut takes the default value with it. Consistent, pre-existing, and arguably right for a signature — but if it is not in your disclosed remainders it probably should be, since a reader could reasonably expect the parameter's default to survive.

## On the rest of the round — this is the strongest of the three arms

**All 10 mutants ALIVE with the test file held out**, including deleting the brace cut outright *and* deleting the annotation strip outright. Neither half was graded at all. Java and C# each had one ungraded half; kotlin had two, plus a second defect nobody had reported.

The `" ="` finding is the one that matters most: `@Table(name = "weird") fun plainFun(a: Int): Int` → `"@Table(name"`, with **no brace anywhere**. That is the majority shape for an annotated Kotlin function, and it was emitting the annotation instead of the declaration. The issue and both my briefs framed this family as a brace hazard; it was never only that.

Two pieces of discipline worth naming, because both are things I have got wrong this week:

- **You read `stripKotlinAnnotations`'s alphabet rather than carrying my C# reasoning across.** I claimed on C# that moving the strip earlier hands the helper a new input class, and that was false there — but your space-only skip means collapse-first genuinely *is* different in kotlin, pinned by `CtorNL` and `TPNL`. Same conclusion, different reason, correctly derived.
- **You recorded the fun half's collapse-after-strip as a HOLE, not an equivalence.** 48 distinguishing inputs out of 35,271 means it is not equivalent, and you said so and left it ungraded and disclosed rather than taking the alphabet shortcut. A false equivalence retires a hole permanently; this is the right call and the harder one to make.

And verifying rather than assuming paid a third time: a preceding KDoc is outside the node span as in java and C#, but a comment **between the annotation and the keyword is inside it**, so `@Ann("x") /* { */ class C` records as `/*`. That is kotlin-specific and would have been wrong to carry over.

PR #7148, opened 2026-09-17
fix(#7145): the TryLock lock-holding detector fired on any holder, not the caller

### LINKED ISSUE #7145: MEASURED: TestRemoveRepoDoesNotCallTheBackendUnderTheLock detects lock-holding with w.mu.TryLock(), whose 'only contended holder is the caller' premise is FALSE — reconcileOnce->scanDir takes w.mu 5x concurrently, so the failure is a false positive
## What

`TestRemoveRepoDoesNotCallTheBackendUnderTheLock` (`internal/daemon/watch/removerepo_6309_test.go`) detects "the backend was called while holding `w.mu`" with **`w.mu.TryLock()`**:

```go
check := func(what, name string) error {
	if !w.mu.TryLock() {
		t.Errorf("%s(%s) was called while holding w.mu — ...", what, name)
		return nil
	}
	w.mu.Unlock()
	...
}
```

with this comment above it:

> `// TryLock is exact here: the only contended holder in this test is the caller itself.`

**That premise is false.** `TryLock` fails if **any** goroutine holds the mutex, and this test starts six of them that can.

## Measured

`newBudgetedWatcher` → `newBudgetedWatcherCfg` → `newBudgetedWatcherCfgNoCleanup` → **`NewWatcherConfig`**, which starts `loop`, `subscribeOwner`, `heartbeat`, `quarantineSweep` and `reconcileLoop`. Counting `w.mu.Lock()`/`RLock()` sites reachable from those goroutines:

| goroutine / callee | `w.mu` acquisitions |
|---|---|
| `loop` | **1** |
| `reconcileLoop` → `reconcileOnce` → `nextReconcileBatch` | **1** |
| `reconcileLoop` → `reconcileOnce` → `scanDir` | **5** |
| `heartbeat`, `quarantineSweep`, `subscribeOwner` | 0 |

And the suite's own helper says the reconcile sweep is deliberately left running:

> `// The reconcile sweep is deliberately NOT pinned. Debounce, bulk and heartbeat are... every one of them runs it, at its production cadence`

So while `RemoveRepo` is executing, `reconcileOnce` can be inside `scanDir` holding `w.mu`. `TryLock` then returns false and the test reports that **the caller** held it — a **false positive**. The production code may be entirely correct.

## The second failure is a cascade, not independent evidence

On the false positive, `check` calls `t.Errorf` and **returns early without appending to `seen`**. So the later count assertion sees a short list and fails too:

```
removerepo_6309_test.go:36: Remove(...) was called while holding w.mu — ...
removerepo_6309_test.go:77: RemoveRepo unwatched 1 directories ([.../src]), want 2 —
  the backend calls were dropped rather than moved out of the critical section
```

Both messages come from **one** racy `TryLock`. The second message's text — "the backend calls were dropped" — is actively misleading here: nothing was dropped, the recording was skipped. Anyone reading the failure is pointed at a production defect that did not occur.

## Observed

`test (ubuntu-latest)` on PR #7129 @ `d171a295c`, run `35225189461`, job `105214840502`. That PR's Go diff is `internal/dashboard` only — **zero files under `internal/daemon`** — and every other leg passed. It is an external contributor's PR, so the cost of this landed on someone with no context.

On `main` @ `c4af71f93`, macOS, non-`-race`: the test passes **3/3 in isolation** and the **full package passes** (168.5s, 0 FAIL). Consistent with a scheduling-dependent false positive rather than a deterministic failure — it needs the reconcile sweep to land inside the window.

## Not claimed

- **Not that the production code is correct.** This says the *detector* is unsound, so a failure from it is not evidence either way. #6309's underlying fix may be fine; this test cannot currently tell you.
- **Not reproduced on Linux.** No Linux here, and I did not reproduce the false positive locally on macOS — the CI failure is the only observation. What is established by reading is that the premise the detector rests on is false.
- **Not measured how often.** One CI instance.

## Suggested shape

The detector needs to distinguish "the caller holds `w.mu`" from "somebody holds `w.mu`". `TryLock` cannot. Options, in rough order of preference:

1. **Make the seam record the caller's lock state explicitly** — have `RemoveRepo` pass a flag, or set a goroutine-scoped marker before/after the critical section, so the assertion observes the property it means rather than inferring it from contention.
2. **Quiesce the background goroutines for this test** — pin `ReconcileInterval` the way `Debounce`, `BulkThreshold` and `HeartbeatInterval` already are. Cheapest, but note the helper's comment argues *against* pinning the sweep on purpose, so this trades one deliberate property for another and that trade should be made explicitly rather than silently.
3. **Assert on the emitted artefact** — if calling the backend under the lock has an observable consequence (it deadlocks on Windows per #6309), pin that consequence rather than the lock state.

Whichever is chosen, the comment asserting the premise must go or become true — it is the reason the unsoundness survived review.

**Grading requirement:** the replacement must still fail when the defect #6309 fixed is reintroduced. Re-apply that defect (call the backend inside the critical section) and confirm the new assertion dies; a detector that stops producing false positives by also stopping true ones is worse than what is there now.

Found while diagnosing a red leg on contributor PR #7129 before rerunning it.

cajasmota:
Closes #7145

Test-only change. `internal/daemon/watch/removerepo_6309_test.go` is the single file touched; production `watcher.go` is byte-identical to `main` (`shasum b638a7f4802a7ddbb0e4761cbb71e530039dda67`, verified after every mutant round).

> **Revision note (review round 2).** Two claims in the first version of this body were wrong and are corrected below, not quietly dropped: the "~3 orders of magnitude of headroom, including under `-race`" figure was **unmeasured and false at the tail** (real margin: 1.24x under `-race` + 96 spinners), and its stated mechanism (mutex hold time / `sync.Mutex` FIFO handoff) was **the wrong mechanism** (it is scheduler starvation of the probe goroutine). The option-2 rejection was also right for a wrong reason. Measurements in the tail table below are the reviewer's, reproduced in the test comment rather than re-derived here; everything I state as re-measured, I re-measured in this worktree.

## The unsoundness, verified from source

The detector was `w.mu.TryLock()`, under `// TryLock is exact here: the only contended holder in this test is the caller itself.`

- `newBudgetedWatcher` → … → `NewWatcherConfig` (`watcher.go:511`) starts `loop`, `subscribeOwner`, `heartbeat`, `quarantineSweep`, `reconcileLoop`.
- `reconcile.go` has 10 `w.mu.Lock()` sites: `scanDir` **5**, `nextReconcileBatch` **1**, `repairDir` **4**. My original breakdown named only the first two and so under-attributed `repairDir` (the reviewer flagged this; I measured 4, not the 3 stated in review). Every one of those sections is pure map work — re-measured here, `os.ReadDir` is outside the lock in `scanDir`, and `repairDir` captures `add := w.fsAdd` under the lock and calls it outside — so the transient-hold premise is stronger than I first claimed.
- Cadence: `reconcileLoop` ticks at `defaultReconcileInterval = 2s` (`reconcile.go:107`) — the helper pins `Debounce`, `BulkThreshold` and `HeartbeatInterval` and states in a comment that the sweep is **deliberately** left at production cadence (#6304).

`TryLock` fails on **any** holder, so a sweep landing in the window reported the caller. Demonstrated at detector level with a throwaway control (3 runs, identical):

```
foreign transient holder: OLD TryLock reports caller-holds=true, NEW probe reports caller-holds=false
```

**Correction on `loop`:** `loop`'s direct `w.mu.Lock()` is at `watcher.go:1484`, a startup-only capture **before** its `for` loop, so it cannot land in the `RemoveRepo` window; its reachable site is `handleEvent` → `chargeEventOpen:1667`, which needs an fs event this test does not generate. The `reconcile` paths are what make the old detector unsound here.

## Option chosen: 1 (make the property observable)

**Not** because pinning the sweep would leave another holder in the window — that argument rested on the `loop` miscount above and is withdrawn. The real reason: pinning `ReconcileInterval` would plausibly have closed the window **in this one test** while leaving `TryLock` **itself** unsound. A detector that happens not to misfire in one configuration is not a sound detector — the next helper change, or the next test that reuses it, re-opens it silently. Option 2 would also spend the helper's deliberate #6304 property (every ledger test runs the sweep) to buy that. Option 3 was unavailable: the #6309 consequence is a Windows deadlock with no observable artefact on this host, which is why the `fsAdd`/`fsRemove` seam exists at all.

The replacement asks a question `TryLock` cannot: **can a goroutine other than the caller acquire `w.mu` while this backend call is in flight?** — a blocking `Lock` from a spawned goroutine, now **3 attempts of 2s** (`muProbeAttempts` x `muProbeTimeout`).

### Why retrying is free

`check` runs **inside** the critical section, so the caller cannot release across any number of retries. Mutant A confirms it empirically: it burns the whole budget and still fails (2.03 s before, **6.04 s** now, still exactly 1 `--- FAIL`). So retries multiply the false-positive margin at **zero correctness cost**.

### The tail, measured — and the real mechanism

The governing mechanism is **not** mutex hold time and **not** `sync.Mutex`'s starvation-mode FIFO handoff: under load the sweep rate collapsed from 5,175/s to 1.25/s while the observed wait grew, i.e. the probe goroutine is losing the **CPU**, not the mutex. Max wait for one foreign blocking `Lock`, Apple Silicon macOS, two goroutines driving `reconcileOnce` at ~280x production cadence plus N spinners:

| config | probes | max wait | vs one 2s deadline |
|---|---|---|---|
| idle | 20.7M | 12.1 ms | 165x |
| 12 spinners | 5.3M | 113 ms | 18x |
| 48 spinners | 1.9M | 243 ms | 8x |
| 96 spinners | 981k | 432 ms | 5x |
| `-race`, 48 | 34k | 737 ms | 2.7x |
| **`-race`, 96** | 32k | **1.617 s** | **1.24x** |

Two things work in this test's favour and are stated in the comment: it draws **one** sample per run, not 10⁶; and this package's automatic CI legs run **without** `-race` (only a tag, a manual dispatch, or `ci:full` add it), where the same load peaked at **432 ms**. **Linux and Windows are unmeasured and no headroom figure is claimed for them.**

The failure message is now hedged accordingly: it names the expected reading (#6309) *and* the one alternative (a starved probe), with the signature that distinguishes them — so a starvation-induced timeout can no longer read as a flat assertion that `RemoveRepo` held the lock, which would be #7145 one layer down.

### Which direction of edit is dangerous

Added beside the constants: only **decreasing** them can break the detector. Larger values make a regression slower to report, never invisible — the seam sits inside the critical section, so the timeout is the only path by which detection can fire. (Reviewer measured `2s → 6s` **ALIVE**, and 1 h merely hangs; the smaller value is the one that breaks, and it is what a "reduce the flakiness" edit reaches for.)

## Also fixed: the cascade

`seen` is now appended on **both** branches. It used to be skipped on the failing branch, so one racy `TryLock` produced a second, independent-looking failure reading *"the backend calls were dropped rather than moved out of the critical section"* — naming a production defect that did not occur. Confirmed absent under mutant A: one failure, one defect.

## Mutants — all four re-scored against the final file state

Asserted exact match counts, `--- FAIL` lines counted (not exit codes), restored by `cp` with shasum + `git diff HEAD` verified each round.

| mutant | verdict | `--- FAIL` |
|---|---|---|
| **A** — #6309 reintroduced (`unwatchAll` moved inside the critical section) | **DEAD** (6.04 s) | **1** |
| **B** — vacuity (`_ = detached`; `RemoveRepo` unwatches nothing) | **DEAD** | **1** |
| **C** — `foreignGoroutineCanTakeMu` forced `return true` | **DEAD** | **2** |
| **D** — forced `return false` | **DEAD** | **3** |

Mutant A, post-retry:

```
    removerepo_6309_test.go:146: Remove(/var/.../001): no other goroutine could acquire w.mu in
    3 attempts of 2s while this backend call was in flight. The expected reading is that the caller
    holds w.mu across the backend call — the #6309 regression: ... The one alternative reading is a
    probe goroutine starved of CPU for 6s total without the lock ever being held by the caller; that
    was measured at a 1.6s tail only under -race plus heavy oversubscription, so if this test is the
    sole failure on an otherwise green, heavily loaded -race run, rule that out before reading it as
    a production defect
--- FAIL: TestRemoveRepoDoesNotCallTheBackendUnderTheLock (6.04s)
```

Mutant B:

```
    removerepo_6309_test.go:146: RemoveRepo unwatched 0 directories ([]), want 2 — RemoveRepo must
    still unwatch every directory it subscribed; one that unwatches nothing would satisfy the lock
    assertion above vacuously
```

C and D are what stop the new grader being vacuous: C kills only `caller-held_lock_is_reported`, D kills `foreign_transient_hold_is_not_reported` **and** the regression itself. Under A and B the grader stayed PASS, as intended — it grades the detector, not `RemoveRepo`.

Note on C and D: `go vet` reports `unreachable code` for both (the forced return precedes the loop). That is a lint, not a compile failure, and `unreachable` is not in the vet subset `go test` runs — both mutants compiled and produced real test failures, quoted above, rather than a tooling error masquerading as a kill.

## Verification

| run | result |
|---|---|
| `go test -timeout 60m -count=5 ./internal/daemon/watch/` | **ok, 886.888s, exit 0**, 0 `--- FAIL` |
| `go test -timeout 90m -race -count=1 ./internal/daemon/watch/` | **ok, 176.235s, exit 0**, 0 `--- FAIL` |
| `gofmt -l internal/daemon/watch/` | clean |
| `go vet ./internal/daemon/watch/` | clean |

The two runs are not comparable to each other (`-race` changes timings substantially here) and are not meant to be. `./cmd/grafel/` was **not** run: this package touches no extractor, so the graph digest cannot move.

## What I could not verify

- **Nothing run on Windows or Linux.** The #6309 deadlock is Windows-only; everything here is the platform-independent locking property, measured on macOS. The starvation tail above is macOS/Apple-Silicon only.
- **I did not reproduce the original CI false positive.** Its window is scheduling-dependent and it did not recur locally. What is established is the detector-level demonstration plus the source reading — not a reproduction of the CI leg.
- **`3 x 2s` is sized against a measured macOS tail, not a proven bound.** Nothing rules out a host that starves the probe for 6 s.
- **Not claimed: that the production code is correct.** This PR restores the detector's ability to answer that question; `RemoveRepo` is unchanged.


cajasmota:
**My own mutants: I attacked the deadline from both sides, and the result characterises it usefully — the lower bound is graded, the upper bound cannot break soundness.**

The obvious worry with replacing `TryLock` with a blocking `Lock` under a deadline is that the deadline becomes the new failure mode. Nobody had scored `muProbeTimeout` itself. My own worktree at `97f2dd14c`, `go vet` 0 before each, edit-landing asserted (not just the match count), `-count=1`, restored by `cp` with `git status` clean.

| mutant | `muProbeTimeout` | `--- FAIL` |
|---|---|---|
| **EL-2** | `2s` → **1ms** (starved) | **2 — DEAD** |
| **EL-1** | `2s` → **6s** (inflated) | **0 — ALIVE** |

**EL-2's killer is the right one**, and it is the direction that matters: `TestMuProbeDistinguishesAForeignHoldFromACallerHeldLock/foreign_transient_hold_is_not_reported`. That subtest pins exactly the #7145 defect — a foreign transient holder must **not** be reported as the caller — so an over-sensitive probe is caught by the control that exists for it. The false-positive direction is graded.

**EL-1 is ALIVE and is not a soundness hole**, which I established rather than assumed. When the seam is called *inside* the critical section — the #6309 defect — the caller cannot release until the seam returns, and the seam is waiting on the probe. So the timeout is the **only** mechanism by which detection can ever fire. A larger value makes detection **slower**, never absent. The evidence is in the run itself: under a 6s timeout the targeted tests took **6.9s**, meaning the positive control genuinely traverses the timeout path rather than short-circuiting.

So the deadline is asymmetric, and worth stating in the file:

- **Lower bound is the risk** — too small and a loaded runner produces the very false positive this PR removes. Graded by the foreign-transient-hold subtest.
- **Upper bound is only cost** — too large and the suite pays wall-clock on a real regression, but never misses one. Unguarded, and harmlessly so.

That asymmetry is the reason the `2s` choice is defensible in a way the old `TryLock` never was: `TryLock`'s failure mode was silent and wrong, whereas this one's failure mode in the dangerous direction is caught by a test. One line in the comment beside `muProbeTimeout` saying which direction is dangerous would make that durable — a future edit "reducing flakiness" is far more likely to reach for a *smaller* number, and that is the one that breaks.

Not asking for a change. Recording it, and the independent review has the load-margin measurement which is the complementary half.

## On the round itself

**Option 2 was rejected on a measurement, not a preference, and the measurement refutes my own brief.** I offered pinning the reconcile interval as the cheap route; quiescing `reconcileLoop` leaves **`loop`** still taking `w.mu`, so it narrows the window rather than closing it — and it would spend the helper's deliberate #6304 property (every ledger test runs the sweep at production cadence) to buy a partial fix. That is a better answer than the one I suggested.

**The detector-level demonstration is the right form of evidence.** `foreign transient holder: OLD TryLock reports caller-holds=true, NEW probe reports caller-holds=false` — a purely foreign hold, the caller holding nothing, the old detector blaming the caller. That shows the defect at the level of the mechanism rather than reproducing a CI flake, which is the honest thing to do when you cannot reproduce the original failure.

**And the cascade fix is the part that will save the most time.** `seen` is now appended on both branches, so a lock complaint can no longer manufacture a second, independent-looking count failure whose message blamed production for "dropping" backend calls that were in fact made. That message is what made #7145 cost a careful investigation instead of a glance.

Mutant A — #6309 reintroduced, `unwatchAll` back inside the critical section — DEAD at **1** `--- FAIL`, and notably only one: no cascade.

cajasmota:
**Correcting my own comment above: I relayed the option-2 refutation as sound, and its stated reason is false. The conclusion survives; the argument does not.**

I wrote that quiescing `reconcileLoop` "leaves **`loop`** still taking `w.mu`, so it narrows the window rather than closing it", and called it a better answer than the one I had suggested. The independent review checked the site:

- `loop`'s single **direct** `w.mu` acquisition is at `watcher.go:1485-1487` — **before** its `for` loop. A startup-only capture, which cannot land in the `RemoveRepo` window.
- Its reachable-during-the-window site is `handleEvent` → `chargeEventOpen:1667`, which needs an fs event this test never generates.

So pinning the sweep plausibly **would** have closed the window in this test. My repetition of the site-count argument was wrong, and I made it worse by endorsing it — a coordinator amplifying an agent's reasoning is how a bad argument acquires authority.

**Option 1 is still right, on the other argument the author gave:** pinning the sweep would have stopped this test misfiring while leaving `TryLock` **itself** unsound. A detector that happens not to misfire in one configuration is not a sound detector — the next test to build a watcher differently inherits the same false premise, and the comment asserting *"the only contended holder in this test is the caller itself"* would still be there, still false, still convincing. That is the reason to replace the mechanism rather than quiet the environment around it.

I have asked for the site-count justification to be dropped from the PR body in favour of that.

**The pattern, since it is the fourth time today:** I keep asserting what a callee does from its **name and position** rather than opening it. `writeGraphGenFlat` was nine lines and the answer was its first statement; `stripCSharpAttributes` branches on three characters; and here `loop`'s acquisition is three lines above its own `for`. Each time the fix was two minutes of reading. The new wrinkle is that this one was not even my own inference — I inherited it and stamped it, which is worse, because the agent had flagged its own reasoning as reasoning and I presented it as established.

One thing this does not change: the deadline findings stand on measurement, not argument. The reviewer's load table is the substantive result here — **1.24× margin under `-race` with 96 spinners**, against a comment claiming ~3 orders of magnitude, with the tail governed by **scheduler starvation of the probe goroutine** rather than mutex-hold time. That is now the blocking item, and the retry-the-probe fix is free in the true-defect case because the caller cannot release across retries while the seam waits inside the critical section.

Issue #7146, opened 2026-09-17
ALL THREE RESIDUALS from #7129 SHIPPED: item 2 in 95a4f9769 (#7149), items 1+3 in f4a1e6099 (#7154) — and item 1 uncovered a MASKED gate plus an ungraded re-filter


cajasmota:
Three residuals from the repository-topology work merged in `36adc6edb` (#7129). All three were found by re-scoring the round-1 mutants against the round-2 tests, and all three were explicitly **not** held against the contributor — the two merge-blocking items were fixed and verified, and these were in the group I had already offered as follow-up.

## 1. `EdgeTruncated → true` is ALIVE

Nothing asserts `edge_truncated == false` on a **complete** graph. `NodeTruncated` has that negative assertion; edges do not. So the flag can report truncation that did not happen and the suite stays green.

```
EdgeTruncated forced to `true`   →  ALIVE against the full ./internal/dashboard/ package
EdgeTruncated forced to `false`  →  DEAD, 2 --- FAIL
```

This is the over-firing direction, which is the documented blind spot here: a recall-style assertion is satisfied by the truncating case alone. One assertion on a payload that is *not* truncated closes it.

**Related, and deliberately not called a hole:** dropping the `edgeCapTruncated ||` arm so the flag rests only on `totalEdgeCount > len(edges)` is also ALIVE, and is believed **equivalent** — in all three paths `edgeCapTruncated` implies `totalEdgeCount > len(edges)` (non-thinned: `len(result) <= candidateCount <= inputCount == totalEdgeCount`; `cap <= 0`: same; thinned: `len(edges) <= inputCount <= totalEdgeCount`). That is **derived from the code's algebra and not fuzzed**, so it should be treated as a claim: a false equivalence retires a hole permanently, and the standard here is algebra *plus* an enumeration. Worth converting to a measurement while fixing item 1, since both touch the same expression.

## 2. The route→banner wiring hop is unpinned

```
normalizeGraphPayload with both flags dropped        →  DEAD, 1 failed / 334
<GraphTruncationBanner data={undefined} /> in
routes/graph.tsx                                     →  ALIVE, 334/334 green
```

The new frontend test renders `GraphTruncationBanner` **directly**, and nothing renders `GraphScreen`, so the route that wires the data into the banner is not exercised. Extracting the banner is what made it testable — and it also **moved** the untested hop rather than closing it, which is an easy thing to miss precisely because the extraction looks like the fix.

[...]

PR #7149, opened 2026-09-17
test(#7146): pin the route→banner wiring hop — the extraction moved the untested hop, it did not close it

cajasmota:
Refs #7146 — **item 2 only** (the route→banner wiring hop). Deliberately not `Closes`: items 1 (`EdgeTruncated` / the `edgeCapTruncated ||` arm) and 3 (`buildV2GraphWithNodeCap`) are owned by a separate Go-side change and stay open.

**Body rewritten after review round 2.** The independent review landed four attacks on the first cut; all four are fixed or recorded below, and every number here is re-measured against the final tree — no count is carried over from round 1.

## The hole

`#7129`'s re-score measured this pair:

```
normalizeGraphPayload with both truncation flags dropped          →  DEAD, 1 failed / 334
<GraphTruncationBanner data={undefined} /> in routes/graph.tsx    →  ALIVE, 334/334 green
```

`webui-v2/src/lib/graph-truncation-banner.test.tsx` renders `GraphTruncationBanner` **directly**, and nothing in the suite rendered `GraphScreen`. So the hop that actually carries the backend's truncation metadata to the user — the route deriving `data` and passing it as a prop — was ungraded. Extracting the banner is what made the component testable, and it **moved** the untested hop rather than closing it; the extraction looks like the fix, which is why this is easy to miss. Each extraction needs a control **one level up**.

## The fix

One new test file, `webui-v2/src/lib/graph-route-truncation-wiring.test.tsx`, which SSR-renders **`GraphScreen`** (the route) with the real `GraphTruncationBanner` underneath, plus one production line: a stable `data-testid="graph-truncation-banner"` anchor on the banner's root div so assertions can be scoped to the banner's own subtree.

Four cases:

| Case | `nodeTruncated` | `edgeTruncated` | `limits` | Delivery | Banner |
|---|---|---|---|---|---|
| 1 | **true** | false | present | stream (`phase: "done"`) | shown |
| 2 | false | **true** | absent | stream | shown |
| 3 | false | false | absent | stream | **absent** |
| 4 | **true** | false | present | **fallback** (`phase: "error"` → `fallback.data`) | shown |

## What review changed

**1. The absence case had no positive control.** The reviewer inserted `if (!data?.nodeTruncated) return null;` before the route's top-level `return (` — a route that renders *literally nothing* on a complete graph — and the absence-only assertion stayed **ALIVE at 336/336**. `not.toContain` cannot tell "the banner correctly stayed away" from "the route produced no output at all", and there are several more ways for it to be a no-op (threw into an error boundary, payload never arrived, string absent for an unrelated reason).

Case 3 now runs its positive control **first**: the markup is non-empty, contains the route's `Module overview` toolbar affordance, and contains the banner's own slot (`class="shrink-0 border-b border-border bg-bg px-4 py-2 space-y-2"` — the div the route renders the banner *into*, so it proves the banner's code path was reached). Only then does it assert the banner subtree is `null`. That mutant (`r1` below) is now **DEAD, 2 failing**, one of them case 3 itself.

**2. `edgeTruncated` was an ungraded arm.** Both the pre-existing fixture and my round-1 fixtures held `edge_truncated: false`, so the two arms of `if (!data || (!data.nodeTruncated && !data.edgeTruncated))` only ever moved together and neither was scored. Dropping the `edgeTruncated` arm was **ALIVE at 336/336** — the banner would silently vanish for an edge-cap-truncated graph with a green suite. New case 2 varies the flags **independently** (edges truncated, nodes NOT); that mutant (`b_node`) is now **DEAD, 1 failing**, killed by case 2 alone.

Scoped to the frontend on purpose: the reviewer notes the same blind spot spans the pre-existing banner test and the Go side, but that wider span is #7146 item 1 and another change owns it.

**3. Row `d` graded one arm of a two-arm ternary.** Mutating `data = streamFailed ? fallback.data : stream.state.payload` to `fallback.data` was DEAD, but the neighbour mutation to `stream.state.payload` — dropping the *fallback* arm — was **ALIVE at 336/336**, because `phase: "done"` was held constant and the fallback→banner delivery path was never exercised. New case 4 drives the payload through `phase: "error"` + a mocked `useGraph`, so both arms are now scored separately: `d_stream` **DEAD, 2** and `d_fallback` **DEAD, 1**.

**4. Rows c1–c3 were mis-attributed, and this is now settled by measurement rather than prose.** The route reaches `normalizeGraphPayload` only through `useGraph`'s `queryFn`; under SSR no fetch runs, and production *stream* payloads are built by `lib/graph-stream-reducer.ts`, which never calls the normalizer. So in round 1 those three kills came from the **test file's own** call to `normalizeGraphPayload`, upstream of the injection point — duplicates of the banner test's kills, not evidence about the route, and the fixture was composing a *stream* payload with the *fallback* branch's producer.

The fixtures are now plain `GraphPayload` literals, so the claim is checkable: c1–c3 each fail **exactly 1** test, and that test is the **banner test**, not either route case. `normalizeGraphPayload` is no longer listed as production plumbing on the route's path, because it is not on it.

**5. One assertion was not banner-scoped.** `0 / 12 nodes` occurs **twice** in the positive markup — once in the banner, once in the route's `Building graph… 0 / 12 nodes` progress affordance — so on its own it would survive deleting the very thing it checks. All four positive assertions now run against the banner subtree extracted via the `data-testid` anchor. Mutant `e_slot` (rename the anchor) is **DEAD, 3**, which is what makes the scoping load-bearing rather than decorative.

**Bonus, from the reviewer's "could not verify" list:** the `vi.mock` factory's shape was not type-checked against the real hook. Adding `satisfies StreamHook` caught **three** drifts in my hand-written mock — it invented `error` and `errorDetail` (the real `UseGraphStreamResult` has neither), omitted `errorMessage`, `state.totalEdges` and `state.done`, and typed `state.payload` as optional where `GraphStreamState` declares it non-optional. `useGraph` keeps a cast, not a `satisfies`: it returns a react-query `UseQueryResult` of ~20 fields and the route reads three, so constructing a whole one buys no grading power. That one mock is therefore **deliberately unpinned against shape drift**, stated in the file rather than left silent.

## What is mocked, and why the hop survives it

**Only the two data-source hooks the route consumes:** `@/hooks/use-graph-stream`, and `useGraph` from `@/hooks/use-graph` (via `importOriginal` so that module's other exports stay real). Both sit strictly **below** the hop. Real and unmocked on the path from the injected payload to the rendered bytes:

- the route's derivation `const data = streamFailed ? fallback.data : stream.state.payload;` (`graph.tsx:164`)
- the JSX pass-through `<GraphTruncationBanner data={data} />` (`graph.tsx:689`)
- the real `GraphTruncationBanner` — not stubbed, not shallow-rendered
- the route's whole top-level render, including the slot the banner sits in

Real `MemoryRouter` and a real `QueryClientProvider` are supplied; no fetch occurs because SSR does not run effects. The mutant table settles the mock-boundary question empirically rather than by argument: `a`, `d_stream`, `d_fallback`, `r1` and `e_slot` are killed **only** by the new file.

## Mutant table

Baseline **re-derived from scratch** on the final tree: **338 passed / 37 files, exit 0** (334 before this change; +4 tests, +1 file). Every row is a full `npm test` (= `vitest run src/lib`) — never a filtered run. Each mutant was applied through a harness that aborts unless `MATCH_COUNT == 1`, confirms the replacement text is on disk, runs `npx tsc --noEmit` (so a type error cannot masquerade as a kill), then restores by `cp` from a byte-level backup with `shasum` PRE==POST and `git diff HEAD -- webui-v2/src` empty. All 13 rows were **re-scored after the last source change**, not carried forward.

| # | Mutant | Verdict | Failing | Killed by |
|---|---|---|---|---|
| a | `data={data}` → `data={undefined}` at the wiring site | **DEAD** | 3 | new file only (was ALIVE 334/334) |
| r1 | route gets `if (!data?.nodeTruncated) return null;` — renders nothing on a complete graph | **DEAD** | 2 | new file only, incl. the absence case |
| b | banner guard → `if (!data)` (renders always) | **DEAD** | 2 | banner test + new absence case |
| b_node | banner guard → `if (!data \|\| !data.nodeTruncated)` (drop the **edge** arm) | **DEAD** | 1 | new file only — the EDGES case |
| b_edge | banner guard → `if (!data \|\| !data.edgeTruncated)` (drop the **node** arm) | **DEAD** | 3 | banner test + 2 new cases |
| d_stream | `data = streamFailed ? … : …` → `fallback.data` | **DEAD** | 2 | new file only |
| d_fallback | same site → `stream.state.payload` (drop the fallback arm) | **DEAD** | 1 | new file only — the FALLBACK case |
| e_edgecount | banner `(data.totalEdgeCount ?? data.edges.length)` → `data.edges.length` | **DEAD** | 3 | banner test + 2 new cases |
| e_limits | banner's `limits` clause deleted | **DEAD** | 3 | banner test + 2 new cases |
| e_slot | `data-testid` anchor renamed | **DEAD** | 3 | new file only |
| c1 | `nodeTruncated: w.node_truncated ?? false,` dropped from `normalizeGraphPayload` | **DEAD** | 1 | **banner test only** — not route evidence |
| c2 | `totalEdgeCount: …` dropped | **DEAD** | 1 | **banner test only** |
| c3 | `limits: …` dropped | **DEAD** | 1 | **banner test only** |

No ALIVE rows, so no open hole and no equivalence claim is made. The compound guard is scored **arm by arm** (`b_node` / `b_edge`), and the ternary is scored **arm by arm** (`d_stream` / `d_fallback`). Rows `b`, `b_edge`, `e_edgecount`, `e_limits` are honestly reported as already-graded by the pre-existing banner test; the new file adds independent route-level kills for them. Rows `a`, `r1`, `b_node`, `d_fallback`, `d_stream` and `e_slot` are what this change actually closes.

One harness note worth recording: a 2-minute tool timeout killed a batch mid-mutant and left `e_slot` **applied** in the worktree. It was restored from the byte-level backup and the row re-run individually; `git status` clean and `shasum` matched before any further measurement. That is the failure mode the `cp`-backup discipline exists for.

## Axes VARIED vs held CONSTANT

**VARIED** (see the case table above):
- `nodeTruncated` — true / false / false / true
- `edgeTruncated` — false / **true** / false / false — varied **independently** of `nodeTruncated`, which is the whole point of case 2
- `limits` — present / absent / absent / present
- delivery path — stream for cases 1-3, **fallback** for case 4

**Held CONSTANT:** `nodes: []`, `edges: []`, `communities: []`, `repos: []`, `totalNodeCount: 12`, `totalEdgeCount: 34`, the group id (`demo`, the route default), the route props (none), the router entry, the query-client config, and `hasMeta: true`.

Consequence: the difference between "banner shows" and "banner absent" is the truncation metadata alone — not a different render path, data source, or code branch. And the two axes the round-1 fixture conflated (`nodeTruncated` with `edgeTruncated`; success with fallback delivery) are now separated, each with its own mutant proving the separation is observed.

## Test file location

`src/lib/`, not beside the route. `package.json`'s test script is literally `vitest run src/lib`, so a test placed anywhere else is **never executed** by `npm test`. That is also why the banner's own test sits in `src/lib/` for a component under `src/components/graph/`. The convention followed is the runner's scope, not directory mirroring — recorded in the new file's header so the next reader does not "fix" the location into invisibility.

## Not verified / scope

- No Go code touched, no Go compiled; items 1 and 3 of #7146 are untouched by design.
- SSR (`renderToStaticMarkup`) is the repo's existing render harness — `webui-v2` has no jsdom and no `@testing-library/react`. So this grades render output, not click/effect behaviour; `useEffect`/`useLayoutEffect` do not run. Both hop expressions (`graph.tsx:164` and `:689`) are plain render-body code and the banner is a pure-render component, so SSR is sufficient *for this hop* — a test of effect-driven state in this route would need a different harness.
- The `useGraph` mock is not type-pinned against shape drift, deliberately (see above).
- `npx tsc --noEmit` exits 0 on the branch tip and on every one of the 13 mutants.


cajasmota:
## Independent adversarial review — #7149

Reviewer worktree: `/Users/jorgecajas/Projects/archigraph-wt-7149-rev` @ `ebc07a3d5` (detached on `origin/wt-7146-fe`), `webui-v2/node_modules` symlinked from the main checkout. Every row below is a **full** `vitest run src/lib` run (never a filtered one), each mutant applied with an asserted `MATCH_COUNT=1`, `npx tsc --noEmit` run per mutant, restored by `cp` from a byte-level backup with `shasum` PRE==POST and `git diff HEAD -- webui-v2/src` empty. No Go was compiled. Baseline re-derived from scratch: **336 passed / 37 files, exit 0**.

### What I confirmed (measured)

| Row | Mutant | Verdict | Failing | Failing test names |
|---|---|---|---|---|
| a | `data={data}` → `data={undefined}` (`routes/graph.tsx:689`) | **DEAD** | 1 | `routes/graph.tsx → GraphTruncationBanner wiring > renders the truncation banner…` — **new file only** |
| b | banner guard → `if (!data)` | **DEAD** | 2 | banner test's negative case **and** the new route-level negative case |
| c1 | `nodeTruncated: w.node_truncated ?? false` → `false` | **DEAD** | 2 | both files |
| d | `data = streamFailed ? fallback.data : stream.state.payload` → `fallback.data` | **DEAD** | 1 | **new file only** |

So rows a/b/c1/d reproduce exactly as claimed, including the two "new test only" rows: the hop **does** survive the mock, and the `data={undefined}` mutant that was ALIVE at 334/334 is now killed. The claim that the file is actually executed also holds — `package.json` `"test": "vitest run src/lib"`, the file is in `src/lib/`, and both test **names** appear in the runner output (baseline stderr labels, and as `×` rows under mutants a/b). Attack 1 failed to land.

SSR adequacy (attack 4) also failed to land: `const data = …` at `graph.tsx:164` and `<GraphTruncationBanner data={data} />` at `:689` are plain render-body expressions and the banner is a pure-render component, both outside the `isLoading`/`isError` ternaries at `:727`/`:747`. Nothing effect-gated is on the hop.

I dumped the actual markup in both cases: positive 5354 bytes, negative 5181 bytes, negative contains the real toolbar (`Module overview`, `Filters`, `Reset`) — so in the negative case the route genuinely renders. That is the good news, and it is precisely what the test itself does not assert.

### Attacks that landed

**1. The negative row has no positive control.** I inserted `if (!data?.nodeTruncated) return null;` immediately before the route's top-level `return (` — a route that renders **literally nothing** whenever the graph is complete:

```
MATCH_COUNT=1  TSC_EXIT=0  VITEST_EXIT=0
 Test Files  37 passed (37)
      Tests  336 passed (336)
```

**ALIVE.** `expect(html).not.toContain("Bounded result:")` cannot distinguish "the banner correctly stayed away" from "the route produced no output at all". The row is not *worthless* — it does kill mutant b, i.e. it catches banner over-firing, which is what it was added for — but as written it passes for the wrong reason under a whole class of regressions. One added line in the negative case (`expect(html).toContain("Module overview")`, or an assertion on a stable banner-container anchor) closes it.

**2. `edgeTruncated` is an ungraded arm, and the new test replicates the axis instead of varying it.** Banner guard `if (!data || (!data.nodeTruncated && !data.edgeTruncated))` → `if (!data || !data.nodeTruncated)`:

```
MATCH_COUNT=1  TSC_EXIT=0  VITEST_EXIT=0   Tests  336 passed (336)
```

**ALIVE + ungraded.** Not equivalent: an `edge_truncated: true / node_truncated: false` payload distinguishes them — the banner would silently vanish for an edge-cap-truncated graph and the suite would stay green. Both the pre-existing fixture and the new one hold `edge_truncated: false`, so the two guard arms only ever move together and neither is scored. The axes table declares `edge_truncated` held constant but not the consequence; given that item 1 of #7146 is *specifically* the `EdgeTruncated` / `edgeCapTruncated ||` arm, holding this axis constant is the one place it hurts most.

**3. Row d grades one arm of the ternary; its neighbour is open.** Same site, opposite direction — `data = streamFailed ? fallback.data : stream.state.payload` → `data = stream.state.payload` (drop the fallback arm):

```
MATCH_COUNT=1  TSC_EXIT=0  VITEST_EXIT=0   Tests  336 passed (336)
```

**ALIVE + ungraded.** `phase: "done"` is held constant in both cases, so `streamFailed` is never `true` and the fallback→banner delivery path is untested. Row d is real, but "the route's own derivation" is graded in the success direction only.

**4. Rows c1–c3 are mis-attributed — they say nothing about the route hop.** The route reaches `normalizeGraphPayload` only through `useGraph`'s `queryFn`, and the PR body itself states that no fetch occurs under SSR; the stream hook is mocked and its production payload is built by `src/lib/graph-stream-reducer.ts`, which never calls `normalizeGraphPayload`. So under this test `normalizeGraphPayload` is invoked **by the test file itself**, upstream of the injection point — the same way the banner test invokes it. Those three kills are therefore duplicates of the banner test's kills, not evidence about the route. Two consequences for the body: "the real `normalizeGraphPayload` … is on the path from the injected payload to the rendered bytes" is not true of the *route's* path, and the fixture composes a **stream** payload using the **fallback** branch's producer. Type-compatible, so not a defect in the hop grading — but it should not be listed as graded production plumbing.

**5. One of the four positive assertions is not banner-scoped.** `0 / 12 nodes` occurs **twice** in the positive markup: once in the banner, once in `Building graph… 0 / 12 nodes` (the route renders its `isLoading` canvas branch under SSR). The assertions match the whole document, not the banner subtree. `Bounded result:`, `0 / 34 edges` and `Caps: 500 nodes, 4,000 edges.` are each unique to the banner, so the row as a whole still fails when the banner disappears (verified by row a) — but `0 / 12 nodes` on its own would survive deleting the thing it is supposed to check. Attack 3 lands only on that sub-assertion.

### Could NOT verify

- No Go compiled or run (machine constraint); items 1 and 3 of #7146 untouched, as the PR says.
- No jsdom / `@testing-library/react` in `webui-v2`, so nothing about effect-driven or interactive behaviour — correctly scoped out in the body.
- The `vi.mock` factory's return shape is not type-checked against the real `useGraphStream` (`tsc --noEmit` is clean with the mock in place), so the mock is not pinned against future drift in the hook's shape. Minor, not blocking.
- Note on my own harness: the `MUTANT_APPLIED` line uses `grep -c -F` and over-counts for short replacement strings (81 for `false,`, 4 for a multi-line insert). The load-bearing assertion in every row is `MATCH_COUNT=1` from the applying script, which aborts otherwise.

### Verdict: **REQUEST-CHANGES**

The central claim is sound and the two headline kills are real — this genuinely closes the `data={undefined}` hole that #7129 measured ALIVE. The changes asked for are small:

1. **Add a positive control to the negative case.** It currently passes when the route renders nothing at all (mutant ALIVE at 336/336, shown above). Assert something the route always renders, or scope the absence assertion to the banner container.
2. **Vary `edge_truncated`.** Add a third case with `edge_truncated: true, node_truncated: false` so the two arms of the banner guard are scored independently. Without it, dropping the `edgeTruncated` arm is ALIVE at 336/336.
3. **Either grade the fallback arm or record it as an open hole.** Add a `phase: "error"` case whose payload arrives via `fallback.data`, or state in the body that `data = stream.state.payload` (fallback arm dropped) is ALIVE + ungraded — currently the body reads as if the derivation is graded.
4. **Correct the c1–c3 attribution in the body.** Say that `normalizeGraphPayload` runs in the test harness, upstream of the injection, and that those rows duplicate the banner test's kills rather than grading the route. Drop it from the "real production code on the path" list, and note that production stream payloads come from `graph-stream-reducer.ts`.
5. **Drop or re-scope `expect(html).toContain("0 / 12 nodes")`.** It matches `Building graph… 0 / 12 nodes` as well as the banner, so it is satisfied with the banner absent.

Items 1, 2 and 5 are one-to-three lines each. Item 3 may legitimately be deferred as long as it is *recorded* — an undeclared ungraded arm is what this PR exists to stop.

cajasmota:
## Verification round — `b3b496dc1`

Same reviewer worktree, moved to `b3b496dc1` (`git checkout --detach origin/wt-7146-fe`), `webui-v2/node_modules` symlinked, tree verified clean before and after every row (`git status --porcelain` empty, `git diff --stat HEAD` empty, `shasum` PRE==POST). Every row is a full `vitest run src/lib`, each mutant applied with an asserted `MATCH_COUNT=1` and `npx tsc --noEmit` run per mutant. Counts re-derived from scratch, not carried. No Go compiled.

**Baseline CONFIRMED: 338 passed / 37 files, exit 0, `tsc --noEmit` exit 0.** All four route cases execute — verified by name under `--reporter=verbose`, not by file count.

### Per-item verdicts

| Item | Claim | My measurement | Verdict |
|---|---|---|---|
| 1 | `r1` (route `return null`) ALIVE → **DEAD at 2**, one of them case 3 | **DEAD, 2 failing**: `…EDGES are truncated and nodes are not`, `…renders NO banner when the route holds a complete graph` | **CONFIRMED** |
| 2 | `b_node` ALIVE → **DEAD at 1**, killed by the edge-only case alone | **DEAD, 1 failing**: `…renders the banner when EDGES are truncated and nodes are not` | **CONFIRMED** |
| 3 | `d_fallback` ALIVE → **DEAD at 1** | **DEAD, 1 failing**: `…renders the banner when the payload arrives via the FALLBACK fetch` | **CONFIRMED** |
| 4 | c1/c2/c3 each fail **exactly 1**, and it is the **banner** test | all three: **1 failing**, and in every case the name is `GraphTruncationBanner > renders backend truncation metadata after hook normalization` — neither route case fails | **CONFIRMED** |
| — | `e_slot` (rename the `data-testid`) DEAD at 3 | **DEAD, 3 failing** — the three positive cases; the negative case correctly unaffected | **CONFIRMED** |

All five of the above are `tsc --noEmit` exit 0, so none is a fake kill from a non-compiling mutant.

### The harder questions, measured

**Does the item-1 control fail for the right reason?** Two distinct no-ops, scored separately:

- *Route renders but the banner's slot is absent* — `s1`, a purely cosmetic edit to the slot div's Tailwind classes (`py-2` → `py-1.5`, `routes/graph.tsx:688`): **all 4 cases fail**. So yes, the control genuinely detects a missing slot. The cost is that `BANNER_SLOT` is the literal class string, so a styling-only edit reddens the whole file for a reason unrelated to truncation. Non-blocking, but a `data-testid` on the slot div would decouple the control from the stylesheet the way the banner's own anchor already is.
- *Slot present but empty* — `a2`, deleting `<GraphTruncationBanner data={data} />` from the route entirely: **DEAD, 3 failing** (the three positive cases); case 3 passes. That residue is inherent to any absence assertion and is covered by the positive cases, so it is not a hole — but it does bound what the control proves: "the route and its slot rendered", not "the banner component was invoked".

**Item 2 — is the compound guard graded as a whole, not only arm-by-arm?** Pair audit, both arms and the connective:

- `b_node` (drop the `edgeTruncated` arm): **DEAD, 1** — edge-only case.
- `b_edge` (drop the `nodeTruncated` arm): **DEAD, 3** — node-only case, fallback case, and the pre-existing banner test.
- connective flip (`&&` → `||`), your EN-1: taken as reported **DEAD, 4** — not re-run.
- `!data ||` arm: **not a legal mutant** — removing it fails `tsc --noEmit` with 8× `TS18048: 'data' is possibly 'undefined'`. That arm is enforced by the compiler rather than by a test, which is a valid control; I am recording it as compiler-graded, **not** as DEAD and not as a hole.

So the two truncation arms are now scored independently *and* jointly. The `edgeTruncated: true / nodeTruncated: false` case is what carries it: before it existed both arms only ever moved together and neither was graded.

**Item 3 — does mocking `useGraph` put the mock at or above the hop?** No. The hop is the selector `streamFailed ? fallback.data : stream.state.payload`; the mock supplies `fallback.data`, i.e. the data *source*, strictly below the selector. The case is not vacuous, and I scored the selector's own input in both directions:

- `d_fallback` (drop the fallback arm): **DEAD, 1** — the fallback case only.
- `sf_true` (`streamFailed = true` unconditionally): **DEAD, 2** — the node-only and edge-only stream cases.

So the delivery-path axis is graded in both directions, which is what a single-direction ternary mutant could not establish last round. On the cast: the route reads exactly three `fallback` fields (`graph.tsx:164, 170/171, 174` — `data`, `isLoading`, `isError`), so the stated justification checks out, and the fallback case's discriminating power rests only on `data`. The cast therefore does not weaken this case *today*; it leaves that one mock unpinned against future shape drift, which the header says out loud. Acceptable as documented, given the stream mock — the one whose shape the hop actually reads through — now carries a real `satisfies StreamHook`.

**Item 4 — the re-attribution is now true of the code, not just of the prose.** The fixtures are plain `GraphPayload` literals and `normalizeGraphPayload` no longer appears in the route test; measured consequence is that c1/c2/c3 fall to the banner test alone (1 failing each, named above), where last round they each failed 2 including a route case. The three rows no longer read as evidence about the route hop.

### Points 1 and 5 of my previous list

Both closed and I verified the mechanism, not just the text: assertions now run against `bannerSubtree(html)` rather than the whole document, and `e_slot` DEAD at 3 shows that scoping is load-bearing rather than decorative — the `0 / 12 nodes` sub-assertion can no longer be satisfied by the route's `Building graph… 0 / 12 nodes` affordance.

### Residual notes (neither blocking nor a change request)

1. `BANNER_SLOT` couples all four cases to the slot div's Tailwind class string (`s1`, above). A `data-testid` on that div would fix it.
2. `bannerSubtree`'s regex is non-greedy to the first `</div>`; it is correct for today's flat banner and would silently truncate if the banner ever gained a nested element. A future nested banner would show up as a puzzling content-assertion failure rather than a clear one.
3. `a2` is the one row whose `tsc` is not clean, and only for `TS6133` (unused import left behind by the deletion); its kill is three assertion failures in the runner, not a compile break, so the verdict stands.

### Could NOT verify

No Go compiled or run; items 1 and 3 of #7146 remain out of scope by design. No jsdom in `webui-v2`, so nothing about effect-driven or interactive behaviour — correctly scoped out. EN-1 and EN-2 taken as reported, not re-derived.

All four items I raised are **CONFIRMED** closed, each by an independently re-derived mutant with named failing tests, and the two neighbours I checked opportunistically (`b_edge`, `sf_true`) are DEAD too. Nothing I could construct leaves a truncation-wiring mutant alive.

### **APPROVE**

Issue #7144, opened 2026-09-17
F# dedup keys carry no enclosing scope — typeSeen and the module key are bare names; letSeen/memberSeen use indent as a scope proxy. UNBLOCKED (#7176 closed), 4 arms + #7205


cajasmota:
## What

The F# extractor's `letSeen` dedupe key is `indent + ":let:" + name` — **indentation depth plus the bare name, with no enclosing scope**. So two bindings at the same *column* with the same name are treated as the same entity, no matter what encloses them.

That is wrong for ordinary, unambiguously legal F#:

```fsharp
module A =
    let Compute x = x

type T() =
    member this.Compute x = x
```

Both sit at four-space indentation and both are named `Compute`. Measured against the real extractor: **count = 1, subtypes = `[let]`** — `T.Compute` is dropped. They are two distinct operations in two unrelated enclosing scopes, and one of them silently disappears.

## Why indentation was ever a proxy for scope

Reasonably: in F# the offside rule makes indentation *carry* scope, so within one enclosing construct, column depth distinguishes nesting levels. The defect is that it distinguishes *depth* and not *identity* — two siblings at the same depth under **different parents** are indistinguishable to the key. The scanner is regex-based and does not track the enclosing `module`/`type`, so there is nothing in the key to separate them.

## Provenance and what is NOT claimed

Found by the independent reviewer on **#7142** while establishing that that PR's fixture pins a reachable shape. They probed the `module A` / `type T` form specifically because it needs no in-one-type name collision and is therefore unambiguously legal — and the collapse fired.

- **Pre-existing and orthogonal to #7136/#7142.** That work fixed a *twinned literal*: the key was built from two independent `":let:"` strings whose agreement nothing observed. This issue is about the key's **contents** being too coarse. #7142 is correct and complete for what it fixed; it neither caused nor worsened this.
- **Not measured for incidence.** `module`-plus-`type` files with a same-named member at the same indentation are plausible in idiomatic F# — a module of helper functions beside a type exposing the same operation is a normal shape — but I have not counted it in a corpus.
- **Derived-not-executed legality.** There is no F# toolchain on this machine (`dotnet`, `fsc`, `fsharpc`, `fsi`, `mono` all absent; `javac` is the only compiler). The probe shape needs no collision inside a single type, which is why it was chosen, but nothing compiled it.
[...]

Issue #7135, opened 2026-09-17
ALL THREE ARMS SHIPPED: memberRE (c1e2c3008), moduleRE+typeRE (0ed6767d7), active patterns (3bb6d28b8) — and arm 3's failure mode was NOT a silent total miss: letRE mis-names the head and mints a phantom (#7163)


cajasmota:
## What

#7131 fixed one instance of a pattern: `letRE` in the F# extractor accommodated a **fixed sequence** of two optional modifiers (`(?:\s+rec)?(?:\s+mutable)?`), so any third modifier landed in the name capture and `let inline f` was indexed as an operation named `inline`.

**The same fixed-sequence pattern is in four more F# scanners**, confirmed by probe during that work and deliberately left out of the fix to keep the change attributable. They split into two failure modes, and the first is worse:

### A. Mis-name — the modifier becomes the entity's `Name` (same shape as #7131)

`memberRE`:

- `member private this.Go` → indexed as **`private`**
- `member val Go` → indexed as **`val`**

Wrong `Name` and `QualifiedName`, so the member is unfindable by name and cannot be a call target. As with #7131, whatever keys the dedup map on the name will also collapse same-scope siblings — check that specifically rather than assuming it, since `letSeen`'s equivalent here may be keyed differently.

### B. Silent total miss — the declaration is not extracted at all

- `moduleRE` misses `module private Foo`
- `typeRE` misses `type private Foo =`
- `memberRE` misses `static member …`
- `compexpr_active_patterns.go:49` misses `let private (|Even|Odd|)`

These are arguably harder to notice than A: a mis-named entity at least shows up in a listing and looks odd, while a missed one leaves no trace and reads as "this file has no modules" — a plausible answer no bind-rate, orphan-rate or dangle instrument can flag. **`static member` is the one to be most concerned about**: it is how F# expresses a type's factory and operator members, so a type can lose its whole public surface.

## Not claimed

- **Not measured for incidence on a real corpus.** These are probe results against the extractors, on hand-written F# source. `let private` and `static member` are idiomatic, which is an argument, not a measurement.
[...]

PR #7150, opened 2026-09-17
fix(#7135): F# `module private Foo` and `type private Foo =` were not extracted at all

cajasmota:
`moduleRE` and `typeRE` in the F# extractor each accommodated a FIXED modifier
sequence — `(?:\s+rec)?` for one, nothing at all for the other — so an access
modifier did not mis-name the declaration, it made the declaration VANISH.

Measured against e96eb099f, before this change:

  module private Foo            -> 0 entities   (want 1)
  module internal Foo           -> 0 entities   (want 1)
  module rec private Foo        -> 0 entities   (want 1)
  type private Foo = { A: int } -> 0 type entities (want 1)
  type internal Foo() = ...     -> 0 type entities (want 1)
  module Foo / module rec Foo   -> 1 entity (the two accommodated shapes)

The two scanners miss for different reasons, and neither is a mis-name:
moduleRE anchors the line END (`\s*$`), so `module private Foo` left ` Foo`
behind after `[\w.]+` had taken `private` and the whole match was abandoned;
typeRE requires the name to start `[A-Z]`, so a lower-case modifier left the
name capture nothing to take.

This is the silent-miss mode, and it is the harder one to notice: a mis-named
entity at least appears in a listing looking odd, while a missed one leaves no
trace and reads as "this file has no modules" — an answer no bind-rate,
orphan-rate or dangle instrument can flag.

RED, and its decomposition, re-derived from scratch after the review rather
than carried: reverting BOTH patterns gives **88** `--- FAIL`. Per test (each
table's count includes its own parent line, and the parts below sum to 88,
checked by the script that produced them):

  TestTypeModifiers_TypeIsNeverMissed              49
  TestModuleModifiers_ModuleIsNeverMissed          25
  TestModuleTypeModifiers_AllowlistUpperBoundary    4
  TestModuleTypeModifiers_NoNameCollapse            3
  TestTypeModifiers_ModifiedTypeKeepsItsTopology    3
  TestTypeModifiers_IndentedType                    1
  TestModuleModifiers_IndentIsNotLoadBearing        1
  TestModuleTypeModifiers_OnlyRealDeclarations      1
  TestTypeModifiers_PrivateTypeKeepsItsSurface      1
  ----------------------------------------------- ---
  SUM                                              88

Reverting `moduleRE` alone gives 30 (25 + 2 + 2 + 1); `typeRE` alone gives 60
(49 + 3 + 3 + 2 + 1 + 1 + 1).

MODIFIER SETS, and the sources unioned for each. The memberRE arm paid for
this lesson (§ 8.13's `member-defn` genuinely omits `inline`), and it repeated
here in the opposite direction:

  module -> {rec, public, private, internal}
    1. F# spec § 10 "Namespaces and Modules":
       `module-defn := attributes? module access? ident = module-defn-body`
    2. § 10.5 accessibility: `access := private | internal | public`
    3. MS Learn "Modules": `module [accessibility-modifier]
       [qualified-namespace.]module-name`, "can be one of the following:
       public, private, internal" — and, on the SAME page, `module rec` under
       "Recursive modules", which the § 10 production does NOT carry. So
       deriving from that one production would have DROPPED a form the pattern
       already handled: a regression, not just an incomplete fix.
    4. Sibling letRE (#7131), which allowlists rec/mutable/inline/access;
       `mutable`/`inline` qualify a value, not a module, so only `rec` and
       `access` carry over.

  type -> {public, private, internal}, and nothing else
    1. F# spec § 8: `type-name := attributes? access? ident typar-defns?`,
       shared by every type-defn variant, and no production admits any other
       keyword between `type` and the ident
    2. § 10.5, same `access` production
    3. MS Learn "Access Control": specifiers apply "to modules, types,
       methods, value definitions, functions, properties, and explicit
       fields", "put in front of the name of the entity", with the examples
       `type private MyPrivateType()` and `type internal MyInternalType()`
    4. Siblings memberRE and letRE, whose access sets are the same three words

`protected` is deliberately NOT allowlisted: MS Learn "Access Control" states
it "is not used in F#" (its caveat concerns overriding an inherited member,
not a declaration-site modifier). There is no `type rec` either — recursive
types use `type A = ... and B`. BOTH exclusions are now GRADED rather than
merely stated; see the boundary test below. There is NO F# toolchain on this
machine (`dotnet`, `fsc`, `fsharpc`, `fsi`, `mono` all absent; `javac` is the
only compiler present), so every legality statement is derived-not-executed
and no fixture was compiled.

TESTS. The modifier space is enumerated as a CROSS PRODUCT rather than
sampled, and crossed with the type-body shapes because § 8 puts the modifier
slot in `type-name`, which record/DU/class/alias/interface/generic all share:

  - TestModuleModifiers_ModuleIsNeverMissed — 14 modifier phrases x 2 names
    = 28 rows (counted with `-v`, not derived)
  - TestTypeModifiers_TypeIsNeverMissed — 7 phrases x 8 body shapes = 56 rows
    (counted with `-v`)
  - TestModuleTypeModifiers_AllowlistUpperBoundary — NEW, see below
  - TestModuleTypeModifiers_NoNameCollapse
  - TestTypeModifiers_IndentedType
  - TestModuleModifiers_IndentIsNotLoadBearing
  - TestModuleTypeModifiers_ModifierPrefixedNames
  - TestModuleTypeModifiers_OnlyRealDeclarations
  - TestTypeModifiers_PrivateTypeKeepsItsSurface
  - TestTypeModifiers_ModifiedTypeKeepsItsTopology — NEW, see below
  - TestModuleModifiers_LocalModuleEqualsGapIsSeparate
  - TestModuleModifiers_BareKeywordUnchanged

What the cross product does and does NOT contain, stated so the label claims
no more than the fixtures hold: the empty phrase, each modifier alone, both
orders of every `rec`-with-access pair, and each ACCESS modifier repeated.
There is no `rec rec` row and no three-modifier row. Rows are labelled per row
and no header sentence claims every row is legal F#: `rec` with an access
modifier is attested by no source in EITHER order, so those rows carry
`lenienceOnly`, while a repeated access modifier carries a new
`lenienceRepetition` label, because calling it a reversed ORDER would claim
something the row does not test.

THE ALLOWLIST'S UPPER BOUNDARY IS NOW GRADED — this is what review added, and
it was a real hole. Must-have rows can only show that an accepted modifier is
not lost; nothing in them fails if the allowlist accepts a word F# does not.
On the first revision, three widening mutants were ALIVE at 0 `--- FAIL`
each: `protected` added to moduleRE's group, `protected` added to typeRE's,
and `rec` added to typeRE's. Both exclusions above existed only in prose, and
`fsModTypeModifierWords` does not help — it forbids an entity NAMED
`protected`, not an entity minted BECAUSE `protected` was accepted. All three
forms mint nothing today, so the forbidden rows are true and deterministic:

  module protected Target            -> 0 module entities
  type protected Target = { A: int }  -> 0 type entities
  type rec Target = { A: int }        -> 0 type entities

Each is paired with a POSITIVE CONTROL differing only in the modifier word
(`module private Target`, `type private Target`, `type internal Target`), which
must mint exactly 1 — so the row cannot pass because the harness sees nothing,
and a one-sided future widening fails instead of drifting.

DOWNSTREAM. The miss cost more than one node, and all five consequences are
now OBSERVED rather than asserted: subtype classification, type->member
CONTAINS edges and DU-case/record-field sub-entities
(TestTypeModifiers_PrivateTypeKeepsItsSurface), plus — added after review —
the hierarchy edges and collectRecordTypeNames
(TestTypeModifiers_ModifiedTypeKeepsItsTopology): `type private Derived()` with
`inherit Base()` and `interface IThing with` gets EXTENDS->Base and
IMPLEMENTS->IThing, and a PLAIN `type Person` record gets VALIDATES->Address
onto a `type internal Address`. That last direction is the one worth watching:
the entity that loses an edge is not the modified type, so an access modifier
on one declaration degraded a different, unmodified one.

Two PRE-EXISTING defects were measured while enumerating and are NOT fixed
here, because both reproduce with no modifier at all:

  - LOCAL MODULES ARE MISSED ENTIRELY: moduleRE's `\s*$` means `module Foo =`
    never matches. TestModuleModifiers_LocalModuleEqualsGapIsSeparate asserts
    the RELATION (the modifier makes no difference) rather than the gap, so it
    fails only if a future fix closes one form and not the other.
  - BLOCK COMMENTS AND TRIPLE-QUOTED STRINGS MINT PHANTOMS: `module Ghost`
    inside `(* ... *)` or `"""..."""` is extracted today. Line comments are
    correctly ignored and THAT direction is graded, by count, in
    TestModuleTypeModifiers_OnlyRealDeclarations.

Also pre-existing and deliberately untouched: lower-case type names are still
missed by typeRE's `[A-Z]`, unchanged before and after this commit; and
typeRE's twin `typeKindRE` was NOT widened, because it is unreferenced
anywhere under internal/ or cmd/ (classification runs through
classifyTypeSubtype instead), so the edit would be unobservable and no test
could grade it. The divergence is documented at the pattern, with the group to
carry over if it is ever wired up.

DEDUP KEYS — the pair audit, not just the fix. The module scan keys
`"module:" + name` with NO indent and typeSeen keys the BARE name, both
file-wide; neither has the indent+name shape of letSeen/memberSeen, so the
open #7144 finding does not apply to them and this commit does not touch
either key. A collapse was also not reachable through this defect: both
scanners MISSED the modified declaration rather than mis-naming it, so there
was never a modifier-shaped key for siblings to share.
TestModuleTypeModifiers_NoNameCollapse pins the direction the widening could
break, by count.

`\b` in each group is not load-bearing and is not claimed as graded. Both
variants were brute-forced against each other: 1,124,864 enumerated module
declaration lines and 512,000 type ones, 0 distinguishing captures each. The
equivalence is MASKED by the mandatory `\s+`, not structural — under a `\s*`
separator the same enumerations diverge on 27,648 and 22,128 inputs
respectively. An independent reviewer replaced that enumeration with their own
alphabet (identifiers whose leading substring is a modifier keyword: `recFoo`,
`recrec`, `privateThing`, `internals`, `recprivate`) over 933,120 module and
52,728 type lines and reached the same verdict: 0 divergences under `\s+`,
93,100 under `\s*`. Two alphabets, one conclusion.

MUTANTS, all re-scored from scratch after the review rows were added rather
than carried — several counts moved because the new tests also kill them.
`go vet` before each score, every edit applied with an asserted exact match
count, `-count=1`, restored by byte `cp` with the sha256 AND `git diff HEAD`
verified:

  moduleRE  drop rec 15, drop public 9, drop private 14, drop internal 11,
            non-repeating group 19, revert to the fixed sequence 30,
            drop the `^` line anchor 1,
            WIDEN by adding `protected` 2                     — all DEAD
  typeRE    drop public 17, drop private 26, drop internal 24,
            non-repeating group 25, revert to the fixed sequence 60,
            drop the `^` line anchor 1,
            WIDEN by adding `protected` 2, WIDEN by adding `rec` 2
                                                              — all DEAD
  both      drop the `\b` word boundary                          ALIVE

The two `\b` mutants are recorded as EQUIVALENT UNDER THE CURRENT SEPARATOR,
not as DEAD and not as a hole, with the enumerations above as the
demonstration. They are left untested on purpose — and both pattern comments
now say plainly that the disposition is CONDITIONAL on the mandatory `\s+`,
that nothing currently grades that separator, and that relaxing it diverges on
93,100 inputs of the independent enumeration. The separator's own hole is
pre-existing and filed as #7158, where `\s+` -> `\s*` is ALIVE at 0
`--- FAIL` in both patterns; it is not fixed here.

A NEGATIVE, measured rather than assumed. Deleting every COUNT assertion from
both tables and reverting the extractor produces 88 `--- FAIL` — the same as
the real red run — and deleting the count from
TestModuleTypeModifiers_OnlyRealDeclarations while dropping both anchors
produces 1, the same as the count-keeping control. So the counts killed
nothing the per-name checks did not also kill: a TOTAL miss is equally visible
to both, and a fixture's phantom names are authored, so they can be enumerated
in advance. The counts are kept because they do not require a test author to
predict a phantom's name, but they are NOT claimed as graded. The direction
that genuinely needed its own grading was the permissive one, and no count
could reach it — hence the boundary test above. Related hole that neither
direction closes: because both dedup maps are name-keyed, a phantom sharing a
real declaration's name is silently merged into it and is invisible to a count
and a presence check alike — the same family as the open #7144.

KNOWN AND DELIBERATELY LEFT: the modifier group accepts an arbitrary MULTISET,
so `module rec rec Target`, `module private public Target` and
`type public private Target` all mint one entity while the grammar admits at
most one access modifier and no `rec` on a type. This is deliberate lenience,
labelled per row, matching letRE and memberRE: an illegal multiset cannot
appear in source that compiles, so it can never mint a phantom from real F#,
and ranking the orders could only create a way to LOSE a real declaration.

VERIFICATION. `go test -count=1 ./internal/extractors/fsharp/` green (whole
package, 0 `--- FAIL`). `go test -count=1 ./cmd/grafel/` green with a FRESH
`GRAFEL_HOME` and a FRESH `GRAFEL_DAEMON_ROOT` — run twice, 212.9s then
198.1s, 0 `--- FAIL` both times — and the six digest gates also run BY NAME to
prove they were not skipped: TestSemanticDigestGradesEntityLanguage6862,
TestCustomExtractorAddedEntitiesCarryACompleteSpan,
TestCustomExtractorGateDoesNotRegressFileComponentStartLines,
TestCustomExtractorGateOffGraphIsUnchanged,
TestCustomExtractorGateOffDeltaIsExactlyTheDocumentedSpanGain,
TestFileComponentSpanDonationIsDeterministic — 6 PASS, 0 `--- FAIL`. The
digest is UNMOVED, and it is regression cover rather than coverage: `Name` is
one of `entityTupleKey`'s eight fields, but `writeSpanFixture6118`'s members
were re-enumerated and there are **18** of them — 7 `.cs`, 1 `.java`, 1 `.ts`,
1 `.js`, 6 `.py`, 1 `.go`, 1 `.rs` — and none is `.fs` or `.fsx`. (An earlier
revision of this body said 17 and omitted `rs/routes.rs`: the enumeration had
truncated the fixture function at a `func ` line occurring INSIDE one of its
own raw-string fixtures. Re-derived by scanning every map key at the literal's
indent.) `docs/grammar-canary-baseline.json` carries no F# entry — this
extractor is regex-based, not tree-sitter — so no baseline moves. `gofmt -l`
clean, `go vet` clean.

Refs #7135 — `moduleRE` + `typeRE` arm. Deliberately not Closes:
`compexpr_active_patterns.go:49` (`let private (|Even|Odd|)`) shares the same
fixed-sequence pattern and stays open.


cajasmota:
## Adversarial review — independent reviewer, did not write this code

Own worktree `/Users/jorgecajas/Projects/archigraph-wt-7150-rev` at `6b6ddf29e` (detached, never touched the author's tree). Every mutant: byte `cp` backup, exact-match-count assertion before running, `go vet` before scoring, `-count=1`, restored by `cp` with sha256 `2f4c9c8e…dfcfe` and `git diff HEAD` verified clean after each. No F# toolchain here (`dotnet`/`fsc`/`fsharpc`/`fsi`/`mono` absent), so every legality statement below is **derived-not-executed** and cited.

### What I re-derived and CONFIRMED (measured)

| Claim | My measurement |
|---|---|
| RED = 81 `--- FAIL` | **81**, both patterns reverted together. Breakdown: 25 `ModuleIsNeverMissed` (24 rows + parent), 49 `TypeIsNeverMissed` (48 + parent), 3 `NoNameCollapse` (2 + parent), 1 each `IndentIsNotLoadBearing` / `OnlyRealDeclarations` / `IndentedType` / `PrivateTypeKeepsItsSurface` |
| revert moduleRE only = 28 | **28** (`ModuleIsNeverMissed`, `IndentIsNotLoadBearing`, `NoNameCollapse`) |
| revert typeRE only = 54 | **54** (`TypeIsNeverMissed`, `IndentedType`, `PrivateTypeKeepsItsSurface`, `NoNameCollapse`, `OnlyRealDeclarations`) |
| package green | `go test -count=1 ./internal/extractors/fsharp/` — **0 `--- FAIL`**, `gofmt -l` clean |
| `./cmd/grafel/` green with fresh `GRAFEL_HOME` **and** `GRAFEL_DAEMON_ROOT` | **0 `--- FAIL`, 222.2s**, and all six digest gates PASS **by name** (`SemanticDigestGradesEntityLanguage6862`, `CustomExtractorAddedEntitiesCarryACompleteSpan`, `…DoesNotRegressFileComponentStartLines`, `…GateOffGraphIsUnchanged`, `…GateOffDeltaIsExactlyTheDocumentedSpanGain`, `FileComponentSpanDonationIsDeterministic`) |
| digest is regression cover, not coverage | Confirmed. `entityTupleKey` hashes **eight** fields (`Kind|Name|QualifiedName|Subtype|SourceFile|Language|StartLine|EndLine`) — `Signature`/`Properties` absent. `writeSpanFixture6118` members enumerated from the map keys: **zero `.fs`/`.fsx`** |
| no F# entry in `docs/grammar-canary-baseline.json` | Confirmed (0 hits) |
| `typeKindRE` unreferenced | Confirmed — `grep -rn typeKindRE internal cmd` returns **exactly one** line, its own definition |
| `protected` "not used in F#" | Verified verbatim from MS Learn *Access Control*: "The access specifier `protected` is not used in F#, although it is acceptable if you are using types authored in languages that do support `protected` access." The caveat is about **overriding** an inherited member, not about a declaration-site modifier, so excluding it from a `module`/`type` declaration allowlist is correct |
| §10's `module-defn` carries no `rec`; `module rec` comes from MS Learn | Verified both. fslang-spec §10: `module-defn := attributes? module access? ident = module-defn-body`, `access := private \| internal \| public`. MS Learn *Modules* documents `module rec` under "Recursive modules". Your citation is accurate and the regression argument for keeping `rec` is real |
| `module [accessibility-modifier] module-name =` local form | Verified on MS Learn — so the disclosed local-module gap is real and source-attested, not speculative |

### Attack 1 — the `\b` EQUIVALENCE, attacked at its ALPHABET. FAILED TO LAND; equivalence confirmed independently.

A false equivalence retires a hole permanently, so I did not take yours. I built my own enumeration with an alphabet that deliberately contains **identifiers whose leading substring is a modifier keyword** — the one class `\b` exists to separate: `recFoo`, `recrec`, `privateThing`, `internals`, `publicity`, `Recompute`, `recprivate`, `privateinternal`, plus `rec`/`public`/`private`/`internal`/`protected`, dotted long-idents (`Ns.Sub.Target`, `rec.Foo`), the empty token, four indents, five separators including empty/tab/double, and eight tails including `\r` and `=`.

- module: **933,120** lines, `\b` vs no-`\b` → **0 distinguishing captures** under `\s+`; **93,100** under `\s*` (e.g. `module rec rec recFoo` → `recFoo` with the boundary, `Foo` without).
- type: **52,728** lines → **0** divergences.
- Algebra that explains it: every repetition of the group is gated by a mandatory `\s+`, so an iteration that ends mid-identifier cannot be followed by whitespace; the engine backtracks to zero repetitions and the name capture takes the whole token. Dropping `\b` cannot change any capture while the separator is `\s+`.
- Both drop-`\b` mutants re-scored on the suite: **ALIVE, 0 `--- FAIL`** each. Recorded as ALIVE + **equivalent**, as you did. A different alphabet reached the same verdict, which is the confirmation an equivalence needs.

### Attack 2 — the arbitrary MULTISET. Confirmed present, judged deliberate, NOT a change request.

Measured through the extractor (not just the regex):

```
module rec rec Target              -> modules=[Target]
module private public Target       -> modules=[Target]
module internal private rec Target -> modules=[Target]
type public private Target = {...} -> types=[Target]
```

Grammar truth: §10 spells `access?` — **at most one** access modifier — and the spec production carries no `rec` at all, so every row above is illegal F#. There are **no forbidden rows** for illegal combinations, and `modifierPhrases` in fact *positively pins* `private private` as producing one entity (`lenienceRepetition`).

I am not asking for that to change. An illegal multiset cannot appear in source that compiles, so it can never mint a phantom from real F#; the rows are labelled per row, the lenience matches letRE/memberRE, and ranking orders could only create a way to lose a real module. The over-firing that *does* bite (block comments, `"""…"""`) is pre-existing and disclosed.

### Attack 3 — the ALLOWLIST'S UPPER BOUNDARY. **This one landed.** Three ALIVE + ungraded mutants.

The multiset is harmless because the *alphabet* is right. Nothing in the suite grades that alphabet. Three widening mutants, each `go vet`-clean, each applied with an asserted match count of 1:

| Mutant | Result |
|---|---|
| moduleRE group → `(?:rec\|public\|private\|internal\|protected)\b` | **ALIVE — 0 `--- FAIL`** |
| typeRE group → `(?:public\|private\|internal\|protected)\b` | **ALIVE — 0 `--- FAIL`** |
| typeRE group → `(?:public\|private\|internal\|rec)\b` | **ALIVE — 0 `--- FAIL`** |

Both of the PR's own boundary statements — "`protected` is deliberately NOT allowlisted" and "There is no `type rec`" — are asserted **only in prose**, and the body even says of the first that "an all-DEAD mutant score cannot detect one". That is true, and it is also the reason to close it: this is the repo's dominant defect class in the permissive direction, and recall rows are structurally blind to it. `fsModTypeModifierWords` does not help — it forbids an entity *named* `protected`, not an entity minted *because* `protected` was accepted.

It is cheap. I measured that all three forms mint **nothing today**, so a forbidden row is true, deterministic, and three lines:

```
module protected Target          -> modules=[]  types=[]
type protected Target = { A: int } -> types=[]
type rec Target = { A: int }       -> types=[]
```

### Attack 4 — downstream consequences asserted in prose but not observed.

The body lists five costs of the miss (subtype, CONTAINS, sub-entities, **hierarchy edges**, **collectRecordTypeNames**). The tests grade three. I measured the other two and they are **true** — but unobserved:

```
type private Derived() = inherit Base() / interface IThing with …
  -> Derived rels = [CONTAINS->…Go][EXTENDS->Base][IMPLEMENTS->IThing]
type internal Address = { City: string } ; type Person = { … Home: Address }
  -> Person rels = […][VALIDATES->Address]   (collectRecordTypeNames reached it)
```

Ten lines added to `TestTypeModifiers_PrivateTypeKeepsItsSurface` would make the claim observed rather than stated.

### Attack 5 — three quoted counts are wrong (evidence hygiene, no behaviour change).

1. "`TestModuleModifiers_ModuleIsNeverMissed` (26 rows x 2 names)". `modifierPhrases(fsAccess, ["rec"])` yields **14** phrases (1 empty + 3 access + 1 `rec` + 6 cross-pairs + 3 repeats), so the table is **28** subtests, which I counted with `-v`: `28`. (Type table: 7 × 8 = **56**, counted `56` — that one is right.)
2. The RED decomposition "24 + 48 + 2 + 4 + 7" sums to **85**, not 81. The real shape is 24 + 48 + 2 + 3 flat tests + 4 parents = 81.
3. `writeSpanFixture6118` has **18** members, not 17 — the enumeration omits `rs/routes.rs` (7 `.cs`, 1 `.java`, 1 `.ts`, 1 `.js`, 6 `.py`, 1 `.go`, **1 `.rs`**). The load-bearing conclusion (zero `.fs`/`.fsx`) survives.

Also: the table header says the space is a cross product including "each modifier repeated", but only the **access** modifiers are repeated — there is no `rec rec` row and no three-modifier row. That is a label claiming more than the fixture contains.

### Attack 6 — the `[A-Z]` anchor. Pre-existing, and NOT introduced or fixed here.

`type privateFoo =` and `type myLowercase = { A: int }` mint nothing both **before and after** this change, so lower-case type names remain missed — pre-existing, out of scope, stated here so it is not misread as this PR's doing. Worth recording: the *widening* direction of that anchor **is** graded — `[A-Z]` → `[A-Za-z]` is **DEAD**, 1 `--- FAIL`, `TestFSharpFieldTypeRefs_NoPrimitiveBlocklistIsNeeded_AndTheReasonIsNotTheObviousOne`.

### Attack 7 — "the counts did not pay off".

The reasoning is sound for *this* defect: a total miss is equally visible to a presence check, and authored phantom names can be enumerated in advance, and I confirm 81 is the red number with the counts present. But it must not be generalised. The direction counts exist for is over-firing, and over-firing is precisely what Attack 3 shows is still ungraded. Keeping the counts is right; the body's framing ("NOT claimed as graded") is honest.

### Could NOT verify

- Any legality claim by execution — no F# compiler on this host. All grammar statements above are derived from fslang-spec §10 / §10.5 and MS Learn *Modules* / *Access Control*, which I fetched and quoted rather than recalled.
- Your specific enumeration script (1,124,864 / 512,000 inputs) is not in the diff, so I could not audit its alphabet directly. I replaced it with my own rather than trusting it, and reached the same verdict.
- Whether any real-world `.fs` corpus file changes entity counts — the package suite and `./cmd/grafel/` both pass, and the digest fixture carries no F#, so no measurement exists either way. That is regression cover, exactly as the body says.

---

## REQUEST-CHANGES

1. **Grade the allowlist's upper boundary** — three forbidden rows, one per ALIVE mutant above, asserting **zero** entities for `module protected Target`, `type protected Target = { A: int }` and `type rec Target = { A: int }`. As shipped, adding `protected` to either group or `rec` to the type group survives the whole package suite with 0 `--- FAIL`, and both exclusions are stated only in prose. Re-score the three mutants after adding the rows (do not carry my counts over) and report the new `--- FAIL` numbers.
2. **Observe the two downstream halves the body asserts** — extend `TestTypeModifiers_PrivateTypeKeepsItsSurface` with the `EXTENDS`/`IMPLEMENTS` edges of a `type private Derived() = inherit Base()` + `interface IThing with`, and the `VALIDATES` edge from a plain record onto a `type internal` record (both measured working above). Right now the PR body claims five consequences and the suite observes three.
3. **Correct the three counts and the cross-product label in the PR body** — 28 module rows not "26 rows x 2 names"; the RED decomposition must sum to 81; the digest fixture has 18 members including `rs/routes.rs`; and either add `rec rec` / a three-modifier row or stop saying "each modifier repeated".

The fix itself is right, the sources are cited accurately, the equivalence claim survives an independent alphabet, and both suites are green including `./cmd/grafel/` with fresh roots. The change requests are about what the suite *notices*, not about what the patterns *do*.

cajasmota:
## Verification round — independent reviewer, did not write this code and did not write the prior review

Scope: confirm or refute that each of the three change requests is closed, plus the conditionality note. Not a re-review from scratch.

My own worktree `/Users/jorgecajas/Projects/archigraph-wt-7150-verify` at `c97f7856c` (detached). Never entered the author's tree or any other agent's. Every mutant: byte `cp` backup, **exact pre-edit match count asserted in-process** (a mismatch aborts without running), `go vet` before scoring, `-count=1`, full package suite (never a filtered run), restored by `cp` and verified by sha256 **and** `git diff HEAD`. Final state: `extractor.go` `058fbff3…c9749`, `module_type_modifiers_7135_test.go` `38eb7426…c542d7`, `validators.go` `62153e55…e0b861`, `git diff HEAD` empty, `gofmt -l` clean.

Baseline re-derived from scratch rather than carried: `go test -count=1 ./internal/extractors/fsharp/` → **0 `--- FAIL`**, exit 0.

---

### Ask 1 — the allowlist's upper boundary: **CONFIRMED**

All three widening mutants re-applied by me and re-scored. Each was `go vet` exit 0 (so no compile error masquerading as a kill), and each landed with an asserted match count of 1.

| Mutant | `--- FAIL` | Failing test names |
|---|---|---|
| `moduleRE` group → `(?:rec\|public\|private\|internal\|protected)\b` | **2 — DEAD** | `TestModuleTypeModifiers_AllowlistUpperBoundary` + `…/module:_`protected`_is_not_an_F#_access_specifier` |
| `typeRE` group → `(?:public\|private\|internal\|protected)\b` | **2 — DEAD** | `TestModuleTypeModifiers_AllowlistUpperBoundary` + `…/type:_`protected`_is_not_an_F#_access_specifier` |
| `typeRE` group → `(?:public\|private\|internal\|rec)\b` | **2 — DEAD** | `TestModuleTypeModifiers_AllowlistUpperBoundary` + `…/type:_there_is_no_`type_rec`_(rec_is_a_MODULE_modifier)` |

The failing assertion is the **forbidden** one (`module_type_modifiers_7135_test.go:709`, e.g. `"module protected Target…" minted 1 entities [Target]`), not an incidental one.

**Then I attacked the controls**, because a forbidden row whose control is broken passes for the wrong reason. I changed each control's modifier to a nonsense word (`qqzz`) on the unmutated extractor:

| Control broken | `--- FAIL` | Which assertion fired |
|---|---|---|
| `module private Target` → `module qqzz Target` | **2 — RED** | line **716**, `positive control … minted 0 entities [], want 1` |
| `type private Target` → `type qqzz Target` | **2 — RED** | line **716**, same |
| `type internal Target` → `type qqzz Target` | **2 — RED** | line **716**, same |

So the control half is itself load-bearing: the row cannot pass on a silent harness failure, in either direction. Both directions of the pair are graded, which is the specific thing the prior review asked for.

**Fixture name-collision check (the dedup-key hazard):** not applicable here, and I verified why rather than assuming. Each row runs the forbidden form and the control as **two separate `runFSharp` calls on two separate single-declaration sources** (`Forbidden.fs` / `Control.fs`), each containing exactly one declaration named `Target`. There is no shared fixture in which a phantom `Target` could merge into a live `Target` under `"module:" + name` or the bare-name `typeSeen` key. The rows grade what they claim.

### Ask 2 — the two prose-only downstream halves: **CONFIRMED**

`TestTypeModifiers_ModifiedTypeKeepsItsTopology` exists with the two subtests described (`hierarchy edges`, `nested-model VALIDATES onto a modified record`) and asserts `EXTENDS=[Base]`, `IMPLEMENTS=[IThing]` on `type private Derived()`, and `VALIDATES=[Address]` on the **plain** `type Person` over a `type internal Address`.

I did not take "the assertions exist" as evidence. I mutated the three **edge-producing paths** themselves and confirmed the new test dies in each case:

| Mutant (edge producer) | package `--- FAIL` | `…ModifiedTypeKeepsItsTopology` FAIL lines |
|---|---|---|
| `collectHierarchyEdges`: `add("EXTENDS", …)` → suppressed | 13 | **2 (parent + `hierarchy edges`) — DEAD** |
| `collectHierarchyEdges`: `add("IMPLEMENTS", …)` → suppressed | 16 | **2 (parent + `hierarchy edges`) — DEAD** |
| `validators.go` `collectRecordTypeNames`: record set forced empty (`if false && …`) | 3 | **2 (parent + `nested-model VALIDATES…`) — DEAD** |

All three `go vet` exit 0. The third is the important one: it kills only `TestFSharp_NestedModelValidates` and this new test, so the new test genuinely observes `collectRecordTypeNames` reaching a modifier-bearing record, rather than passing whether or not the edge is produced. Under a `typeRE`-only revert the test contributes **3** FAIL lines (parent + both subtests), consistent with the body's decomposition.

### Ask 3 — the corrected counts: **CONFIRMED**

- **The decomposition sums.** 49+25+4+3+3+1+1+1+1 = **88**, arithmetically, before measuring anything.
- **RED measured = 88.** Both patterns reverted together (`module(?:\s+rec)?`, bare `type`), `go vet` exit 0: **88 `--- FAIL`**, and my per-test decomposition is **identical to the body's, test for test**: `TypeIsNeverMissed` 49, `ModuleIsNeverMissed` 25, `AllowlistUpperBoundary` 4, `ModifiedTypeKeepsItsTopology` 3, `NoNameCollapse` 3, `PrivateTypeKeepsItsSurface` 1, `IndentedType` 1, `OnlyRealDeclarations` 1, `IndentIsNotLoadBearing` 1.
- **`moduleRE` alone = 30**, measured (`ModuleIsNeverMissed`, `NoNameCollapse`, `IndentIsNotLoadBearing`, `AllowlistUpperBoundary`). Body's `25+2+2+1` = 30. ✅
- **`typeRE` alone = 60**, measured (`TypeIsNeverMissed`, `NoNameCollapse`, `IndentedType`, `OnlyRealDeclarations`, `PrivateTypeKeepsItsSurface`, `ModifiedTypeKeepsItsTopology`, `AllowlistUpperBoundary`). Body's `49+3+3+2+1+1+1` = 60. ✅
- **Row counts, two independent derivations.** Statically from `modifierPhrases`: `(fsAccess, ["rec"])` = 1 empty + 3 access + (1 `rec` + 3×2 cross-pairs) + 3 repeats = **14** phrases × 2 names = **28**; `(fsAccess, nil)` = 1+3+0+3 = **7** × 8 body shapes = **56**. Measured with `-v`, counting `=== RUN` subtest lines: **module 28, type 56**. ✅
- **Digest fixture = 18 members**, re-derived by a method that cannot hit the trap the author described. I did **not** scan for `func ` boundaries. I scanned `writeSpanFixture6118`'s map literal for keys at the **two-tab map-key indent**, with a **backtick-parity tracker** so any line inside a raw-string body (including one beginning `func `) is skipped by construction. Result: **18** keys, function ending line 272 — 7 `.cs`, 1 `.java`, 1 `.ts`, 1 `.js`, 6 `.py`, 1 `.go`, 1 `.rs` (`rs/routes.rs` at line 251), **zero `.fs`/`.fsx`**. Exactly the body's enumeration.
- **Cross-product label.** Accurate in the header: "there is no `rec rec` row (only the access modifiers are repeated) and no three-modifier row", and the code bears that out — the repetition loop iterates `access` only, `rec` is never doubled, and no phrase has three words.

### Ask 4 — the conditionality note: **CONFIRMED**

I read the 14 inserted comment lines (and did not re-run the `\b` mutants or `./cmd/grafel/` — a comment-only edit cannot move either, and this round changed nothing but comments, so the `./cmd/grafel/` gate is genuinely not required here).

The note is present at **both** patterns: 8 lines at `moduleRE` and 5 at `typeRE` (the latter cross-referencing the former). It states (a) the disposition is **CONDITIONAL** on the mandatory `\s+`, (b) **nothing grades the separator**, (c) the independent enumeration's **93,100** divergences under `\s*`, (d) **#7158**, with the concrete behavioural consequence per pattern (`moduleLoader` → a module named `Loader`; `typeState() =` → a type named `State`), and (e) "re-read this equivalence before touching the separator". Every one of those matches what was measured, #7158 is open and its title is about exactly this, and both statements about which identifier would mint are the ones reported for EO-1/EO-2. Nothing in the note overclaims.

---

### One residual nit (non-blocking, not a change request)

`modifierPhrases`' own doc comment still reads "…both orders of every cross-kind pair, and **each modifier repeated**", while the loop repeats only `access`. The test header block two screens above now says the accurate thing ("only the access modifiers are repeated"), so the corrected label exists — but the helper's comment is a second copy of the old overclaim. Worth a word if the file is touched again; not worth a round.

### What I could NOT verify

- Any F# legality claim **by execution** — no F# toolchain on this host (`dotnet`/`fsc`/`fsharpc`/`fsi`/`mono` absent). I did not re-litigate the grammar; the prior review already fetched and quoted fslang-spec §10/§10.5 and MS Learn, and my scope was whether the exclusions are now **graded**, which they are. All legality statements in this PR remain derived-not-executed, as the body says.
- The author's enumeration scripts (1,124,864 / 512,000 inputs) are not in the diff; I did not re-audit them. The prior reviewer replaced them with an independent alphabet and reached the same verdict, and this round did not touch that code.
- I did not re-run `./cmd/grafel/` or the `\b` mutants, deliberately — comment-only round.

---

## APPROVE

All three change requests are closed, and closed with the grading rather than with prose: three widening mutants DEAD at 2 `--- FAIL` each with the right test names, the positive controls proven load-bearing by breaking all three, the two downstream halves observed by mutating the actual edge producers, the RED decomposition summing to and measuring 88 test-for-test, 28/56 row counts confirmed two ways, 18 fixture members re-derived by a raw-string-safe method, and the conditionality note present at both patterns saying only true things. The one residual is a duplicated stale phrase in a helper comment.

PR #7154, opened 2026-09-17
fix(#7146): edge_truncated could report truncation that never happened — nothing asserted the false case, and the test-only builder made a production claim

cajasmota:
Refs #7146 — items 1 and 3 only. Deliberately not `Closes`: item 2 (the frontend route→banner hop) is a separate PR and the issue stays open.

**Round 2 — addresses all four REQUEST-CHANGES asks from the adversarial review.** All four were comment/test-level; the substance of items 1 and 3 was reproduced and confirmed by the reviewer. The one behavioural change in this round is a new unit test closing the reviewer's own M9. Every count below is **re-derived from scratch** on the round-2 tree — none carried over, because adding a test can remove a mutant's only killer.

## Item 1 — the `EdgeTruncated → true` hole

`EdgeTruncated` could report truncation that never happened and the suite stayed green. Every pre-existing assertion on it checks the TRUE direction on a payload that really was truncated, which a stuck-true flag satisfies. `NodeTruncated` had its negative (`!got.NodeTruncated` on the edge-cap-only case); edges had none anywhere in `internal/dashboard/`.

**Added** `TestV2GraphCompleteGraphReportsNoEdgeTruncation` (new file `internal/dashboard/v2_graph_truncation_test.go`) — a 40-node / 60-edge graph served through the real `GET /api/v2/graph/{group}` handler at `lod=full`, asserting `total == served` on both axes and `node_truncated == edge_truncated == false`. The fixture is checked in-test against `lodLimits("full")`; if either cap is ever lowered below the fixture the test `t.Fatalf`s instead of quietly becoming a truncated-graph test that asserts the opposite. (Reviewer verified both size guards fire with their own messages when the constants are shrunk, and that the test carries its own positive control: `fetchV2Graph` fatals on any non-200, and the 40/60 count assertions run *before* the two `false` checks.)

### Axes VARIED vs held CONSTANT (the complete-graph test)
- **Varied:** nothing — this is a single point, on purpose: the one state the suite had no observation of.
- **Held constant:** `lod=full` (the largest cap pair); single repo; no cross-repo links; `filter_kind=""`; `include_external=false`; `include_modules=false`; 60 *distinct* `(from,to)` pairs so no two relationships collapse into one wire edge.
- The multi-axis coverage lives in the enumeration below, which varies all of those except the LoD name.

### The adjacent equivalence claim — landed **(b)**, no distinguishing state

The issue marked "dropping the `edgeCapTruncated ||` arm is equivalent" as a claim derived from algebra and never fuzzed. Converted to a measurement; **no reachable state distinguishes the two expressions.**

`TestEdgeTruncatedEqualsServedEdgeDeficit` (committed) enumerates the state space of `buildV2GraphWithLimits` and asserts `EdgeTruncated == (TotalEdgeCount > len(Edges))`:

| | |
|---|---|
| Alphabet | 6 entity counts {1,2,3,4,6,9} × 7 edge layouts {empty, chain, star, complete, half-chain-half-isolated, duplicated-chain, chain+3 dangling links} × 2 repo layouts {single, split-across-two-repos-so-cross-parity-edges-become-`grp.Links`} × 7 nodeCaps {0,1,2,3,5,9,1000} × 6 edgeCaps {0,1,2,3,5,1000} |
| Cases | **3,528**, assertion live in all of them |
| Differing | **0** |
| Buckets (now a **partition**, ask 3a) | `edgeless=1176 none=988 cap-only=444 thin-only=716 both=204`, **sum=3528** — asserted in-test, both that each bucket is non-empty and that the five sum to the case count |

**Ask 3a — fixed.** In round 1 four buckets summing to 2,352 were reported against 3,528 cases, leaving a reader to wonder about the missing 1,176. The reviewer instrumented it and found `unclassified == zeroTotalEdge == 1176` exactly — the edgeless states, which fell through because `none` is gated on `TotalEdgeCount > 0`. That reproduces here byte-for-byte. Rather than explain the remainder, the test now has an explicit `edgeless` arm plus a `sum != cases` assertion, so the buckets partition the space and nobody has to instrument it again. The invariant was always asserted in all 3,528 including the edgeless ones (M1 dies there first).

**Ask 3b — stated plainly.** The enumeration's invariant compares two quantities both derived from the same served payload, so it is **cap-value-agnostic by construction** and survives any mutant that only changes a cap's magnitude. It grades the flag's *definition*, not any cap value. This is now said in the test's doc comment and in the code note, next to the case count.

Non-vacuity of the arm itself (reviewer's measurement, reproduced in the code note): `edgeCapTruncated` is true in **578 of 3,528** states, all 578 via its `candidateCount > len(result)` disjunct. Its other disjunct, `inputCount != candidateCount`, fires **0 of 3,528** times — unreachable from this caller, because `visitEdges` has already dropped any edge with an invisible endpoint.

The disposition is written **into the code** at the `EdgeTruncated` site: ALIVE-and-equivalent *on purpose*, with the per-path algebra (non-thinned: `totalEdgeCount == inputCount` and `len(edges) == min(candidateCount, cap) <= inputCount`; thinned: `totalEdgeCount` counts the whole graph while the collector only sees kept-kept edges, so `totalEdgeCount >= inputCount`), the 578/3,528 reachability, the 0/3,528 unreachable disjunct, **and** both things the enumeration does not establish.

## Item 3 — `buildV2GraphWithNodeCap`

**Chose: delete it.** Justification, from what the code supports rather than what would make the comment true:

- Both production entry points (`v2_graph.go:195` and `v2_graph_stream.go:293`) derive `nodeCap, edgeCap := lodLimits(...)`, and `lodLimits` returns a non-zero pair on **every** branch including `default`. So `edgeCap == 0` — the value `buildV2GraphWithNodeCap` hard-coded — is a shape production **cannot** produce. Inventing a caller would have meant inventing a use for an unbounded-edge path that the LoD design deliberately does not have. (Reviewer independently verified all four `lodLimits` returns, that they are already pinned by the pre-existing `TestLodLimitsAreFinite` across 8 `lod` strings, and that the moved doc comment is true where it now sits.)
- Its doc comment was a production rationale ("for large groups this avoids materialising millions of…") for a function production did not call. That rationale is **true of `buildV2GraphWithLimits`**, which is where it now lives, extended with the invariant that makes the deletion safe.

**What the two callers became** (they still grade something real, and now grade a path production takes):
- `TestBuildV2GraphWithNodeCapMatchesLegacyThinning` → renamed `TestCompactNodeThinningMatchesLegacyThinning`, now calls `buildV2GraphWithLimits(..., cap, edgeCap)` with `_, edgeCap := lodLimits("normal")` — taken from `lodLimits` rather than hard-coded so it tracks the production constant. It still compares the compact integer-adjacency path node-for-node and edge-for-edge against the legacy build-then-thin path. A new in-test guard `t.Fatalf`s if the fixture's kept-kept edge count ever reaches `edgeCap`, so the test cannot silently stop grading node thinning and start grading edge capping.
- `BenchmarkBuildV2GraphLOD20K300K` → now `buildV2GraphWithLimits(..., normalLodNodeCap, normalEdgeCap)`. `normalLodNodeCap` is 3,000, the same node cap it used before, so the benchmark's node axis is unchanged and it now measures a bounded-edge run.

### Axes VARIED vs held CONSTANT (the compact-thinning test)
- **Varied by the change:** the edge cap only (0 → `normalEdgeCap`).
- **Held constant:** the 120-entity / 900-relationship fixture, `nodeCap = 30`, `filter_kind`, external/module flags, and the exact node-and-edge equality assertions. Edge ORDER is preserved by the change because with `candidateCount <= cap` the collector sorts by insertion `sequence`, which is the same order the `cap <= 0` branch produces — verified element-by-element, not just by length.

### Pair audit — the surviving sibling (ask 4, second half)

`buildV2Graph` (the 5-arg wrapper, `v2_graph.go:224`) is **also test-only-called** — I re-derived this independently: `grep -rn 'buildV2Graph(' internal/ cmd/ --include='*.go' | grep -v _test.go` returns **only its own definition**, and there are **5** call sites, all in `v2_graph_xrepo_test.go` and `v2_graph_lod_compact_test.go`. **It stays, and here is why:** it is a strictly weaker instance of the pattern than the function item 3 removed. It hard-codes `lodLimits("full")` — a cap pair production *does* serve — so the path its callers exercise is a real production path, and its doc comment makes no claim about production behaviour that the call graph contradicts. The defect item 3 fixed was a *false production rationale on an unreachable configuration*; neither half is true here. The coordinator is filing it separately; it is deliberately untouched in this PR so the diff stays confined to what was reviewed.

## Item 2

Not in this PR. The frontend route→banner hop is owned by another change; `Refs`, not `Closes`, for that reason.

## Mutants — round 2, all nine re-derived from scratch

Counts are **not** carried from round 1: adding `TestCollectCappedGraphEdgesDropsInvisibleEndpoints` could have removed some other mutant's only killer, so the whole set was re-scored on the round-2 tree. Every mutant: byte-level `cp` backup (never `git checkout -- <file>`), **exact per-part match count asserted before running** (the mutator refuses and restores on a mismatch), shasum compared before/after to prove the edit landed, **`go vet` per mutant** so a compile error cannot masquerade as a kill, full `go test -count=1 -timeout 1800s ./internal/dashboard/` — never a `-run` subset — and restore by `cp` with the shasum verified. Run in the background with an explicit long `-timeout`, after a 2-minute tool timeout left a mutant applied in round 1.

**Re-derived baseline: exit 0, 0 `--- FAIL`, 40.7s.**

| # | Mutant (as applied — reproducible) | `go vet` | `--- FAIL` | Killers | Disposition |
|---|---|---|---|---|---|
| M1 | `EdgeTruncated: edgeCapTruncated \|\| true` | clean | **2** | `TestV2GraphCompleteGraphReportsNoEdgeTruncation`, `TestEdgeTruncatedEqualsServedEdgeDeficit` | **DEAD** — the issue's ALIVE hole, closed |
| M1-ctl | M1 with the new test file moved out of the package | clean | **0** | — | **ALIVE** — positive control; reviewer confirmed with positive proof of the hold-out (`=== RUN` for both new tests = 0 against 1,183 total) |
| M2 | `EdgeTruncated: edgeCapTruncated && totalEdgeCount > len(edges) && false` | clean | **3** | 2 pre-existing + the enumeration | DEAD |
| M3 | **compilable form (ask 4):** `EdgeTruncated: totalEdgeCount > len(edges),` **plus `_ = edgeCapTruncated`** before `recomputeServedDegree` | **clean** | **0** | — | **ALIVE + EQUIVALENT** |
| M4 | `EdgeTruncated: edgeCapTruncated \|\| totalEdgeCount > len(edges) && false` | clean | **2** | `…ReportsEdgesDroppedByNodeThinning`, the enumeration | DEAD — the *other* arm is load-bearing |
| M5 | `NodeTruncated: totalNodeCount >= len(nodes)` (stuck true) | clean | **3** | incl. the new complete-graph test | DEAD |
| M6 | `collectCappedGraphEdges(nodes, edgeCap*2, …)` — cap **ENLARGED**, both sites (match count asserted = 2) | clean | **1** | `TestBuildV2GraphMetadataReportsEdgeCapWithoutNodeThinning` — **pre-existing** | DEAD, but see the caveat below |
| M7 | `len(nodes) > nodeCap*2` — node cap **ENLARGED** at the application site | clean | **6** | 1 pre-existing + 5 LoD tests | DEAD |
| M8 | `kept[from] \|\| kept[to]` (relax the thinned-branch gate) | clean | **0** | — | **ALIVE + MASKED** (corrected from "equivalent" — ask 1) |
| **M9** | **reviewer's:** delete **both** of `collectCappedGraphEdges`' `visible` re-filters (the two early returns in the `cap <= 0` branch and the `!sourceOK \|\| !targetOK` guard in the capped branch), compilably via `_, _ = sourceOK, targetOK` | clean | **1** | `TestCollectCappedGraphEdgesDropsInvisibleEndpoints` — **new in this round, its only killer** | **DEAD** (was ALIVE + ungraded) |

Round-1 → round-2 deltas: M3's row is now reproducible (it did not compile as tabled — `declared and not used: edgeCapTruncated` — which per the standing rule scores as a spurious kill; the disposition was unchanged but the row was not evidence). M9 went ALIVE → DEAD. Every other count reproduced identically. Tree ends at the exact pre-sweep checksum, `git status` showing only the two intended files.

### Ask 1 — M8 is **ALIVE + MASKED**, not equivalent

Round 1 called M8 equivalent on the strength of "0 of 4,116 whole-response digests differ". The *conclusion* holds, but the stated mechanism was half the story, and the reviewer's instrumentation quantified the missing half:

```
HEAD:  edgeCapTruncated=true 578/3528   ineq(cap<=0)=0    ineq(capped)=0     cand>len=578
M8:    edgeCapTruncated=true 1364/3528  ineq(cap<=0)=152  ineq(capped)=760   cand>len=578
```

M8 **flips `edgeCapTruncated` in 786 of 3,528 states**, firing the `inputCount != candidateCount` disjunct 912 times — a disjunct that fires **0** times at HEAD. The re-filter keeps the edge *list* identical (`cand>len` unchanged at 578); it does **not** keep the flag identical. The payload survives only because the `totalEdgeCount > len(edges)` arm is already true in every one of those 786 states and masks the flip. That is a third disposition — weaker than equivalent, and it is now recorded as one in the code, with both steps spelled out.

The reviewer also caught that the two new comments formed a **circular safety story** — the gate is safe because the collector re-filters; the collector's verdict is called redundant; the flag is right because of the arm the other comment calls load-bearing: three guards, one observation. The M8 comment now ends by naming the test that actually observes the mechanism and by saying explicitly that the function's two notes must not be read as each other's justification. Closing M9 is what makes that true rather than asserted.

### Ask 4, first half — the M6 caveat, said plainly

**M6's sole killer is a pre-existing test, and no new test in this PR grades a cap's value.** The 3,528-case enumeration structurally cannot: its invariant compares two quantities both derived from the same served payload, so it is cap-value-agnostic by construction and survives any cap-magnitude mutant. Cap *values* are graded by `TestBuildV2GraphMetadataReportsEdgeCapWithoutNodeThinning` and the LoD tests — all pre-existing. M6 DEAD 1 is a statement about the suite as a whole, **not** new coverage from this PR, and the "3,528 cases" headline should not be read as covering it. Per the issue's note, no mutant widens `repositoryTopologyMaxNodes`: that lives in the untouched topology path and would spin the self-sizing fixtures at 100% CPU rather than failing.

### Dispositions that are NOT kills, restated

- **M3 — ALIVE + equivalent.** Algebra path-by-path (reviewer verified) **plus** enumeration: 0 differing over 3,528, with the arm genuinely reached in 578, so the equivalence is non-vacuous. Left untested on purpose, said so in code.
- **M8 — ALIVE + masked.** Weaker than M3's disposition, deliberately: the mutant *does* change an intermediate value in 786 states, and only the masking arm hides it. Recorded with the numbers so nobody upgrades it to "equivalent" or fills it with a vacuous fixture.

## Verification

- `go test -count=1 -timeout 1800s ./internal/dashboard/` — **exit 0, 0 `--- FAIL` lines**, 40.7s. Full package, not a subset.
- `go vet ./internal/dashboard/` clean; `gofmt -l internal/dashboard/` empty.
- Buckets partition, asserted in-test: `edgeless=1176 none=988 cap-only=444 thin-only=716 both=204, sum=3528`.
- `./cmd/grafel/` **not** run and not required: the diff is three files, all under `internal/dashboard/` — no extractor, `internal/graph`, or assembly-path file is touched. Reviewer agreed.

### What I could NOT verify
- The round-1 uncommitted 4,116-case whole-response enumeration is still not in the diff, so it remains unreproducible from the PR alone. It is no longer load-bearing: M8's disposition now rests on the committed 3,528-case grid plus the reviewer's per-disjunct counters, and on `TestCollectCappedGraphEdgesDropsInvisibleEndpoints` observing the mechanism directly.
- M9's hold-out control is the reviewer's own measurement on the parent commit `feece307` (ALIVE, 0 `--- FAIL`), which is exactly the without-the-new-test state. I did not re-run it as a separate hold-out, so that control is the reviewer's run, not an independent second one.


cajasmota:
## Adversarial review — independent re-measurement

Worktree `/Users/jorgecajas/Projects/archigraph-wt-7154-rev` (detached at `feece3078`), asserted by absolute path inside every script. Every mutant: `cp` byte backup, exact-match-count assertion before running, `go vet` before scoring, full `go test -count=1 -timeout 1200s ./internal/dashboard/` (never a `-run` subset), restore by `cp` with `shasum` **and** `git diff HEAD` verified. Tree ends clean (`git diff --stat HEAD` empty, all three files at their pre-mutation checksums). Counts re-derived from scratch, none carried.

### Baseline (re-derived)

`./internal/dashboard/` — **exit 0, 0 `--- FAIL`, 43.9s**. `go vet` clean. Matches the claim.

### The bucket arithmetic — answered, and it is NOT a gap

The four reported buckets sum to 2,352 against 3,528 cases, leaving 1,176 unaccounted. I instrumented the enumeration with a `default:` arm plus a `TotalEdgeCount == 0` counter:

```
PROBE cases=3528 unclassified=1176 zeroTotalEdge=1176 sum=3528
3528 cases, buckets none=988 cap-only=444 thin-only=716 both=204, zero differing
```

`unclassified == zeroTotalEdge` **exactly**. The 1,176 are precisely the edgeless states — `none` is deliberately gated on `got.TotalEdgeCount > 0`, so a graph with no edges at all falls through every arm. The four reported numbers reproduce byte-for-byte. Crucially the **assertion is live in all 3,528**, edgeless cases included: there `deficit == false`, so a stuck-true flag fails there first (and `M1` does in fact die there). Nothing is unreached — but the PR body and the in-code note both say "all four truncation buckets non-empty" without saying the buckets partition only the edge-bearing 2,352, which invites exactly the arithmetic suspicion I arrived with. One clause fixes it.

### Did the walk reach the interesting state? (beyond the count floor)

A bucket floor cannot tell you whether the **mutated arm** was ever true. I instrumented `buildV2GraphWithLimits` and `collectCappedGraphEdges` with counters and re-ran the grid at HEAD:

```
HEAD:  edgeCapTruncated=true in 578/3528   inputCount!=candidateCount: 0 (cap<=0 branch)  0 (capped branch)   candidateCount>len(result): 578
```

So the `A` of `A || B` is genuinely reached in 578 states — the M3 equivalence is **not** vacuous over this grid. This is the measurement the equivalence claim actually needed and it holds. It also confirms the side note: `inputCount != candidateCount` is **0 of 3,528** in both branches, i.e. that disjunct is unreachable from `buildV2GraphWithLimits`.

### Mutants I re-derived

| # | Mutant | `go vet` | `--- FAIL` | Failing tests | vs claim |
|---|---|---|---|---|---|
| M1 | `EdgeTruncated: edgeCapTruncated \|\| true` | clean | **2** | `TestV2GraphCompleteGraphReportsNoEdgeTruncation`, `TestEdgeTruncatedEqualsServedEdgeDeficit` | **confirmed DEAD 2** |
| M1-ctl | M1, new test file moved out of the package | clean | **0** | — | **confirmed ALIVE**, with positive proof of the hold-out: `=== RUN` lines for both new tests = **0**, total `=== RUN` = 1,183 |
| M3 | drop the `edgeCapTruncated \|\|` arm | see below | **0** | — | **confirmed ALIVE + equivalent** |
| M6 | `collectCappedGraphEdges(nodes, edgeCap*2, …)`, both sites (match count asserted = 2) | clean | **1** | `TestBuildV2GraphMetadataReportsEdgeCapWithoutNodeThinning` | **confirmed DEAD 1** |
| M7 | `len(nodes) > nodeCap*2` at the application site | clean | **6** | `…ReportsEdgesDroppedByNodeThinning`, `TestV2GraphLoDMetadataReportsAppliedLimits`, `TestV2GraphLoD_Overview`, `_Normal`, `_LegacyNames`, `_DefaultIsNormal` | **confirmed DEAD 6** |
| M8 | `kept[from] \|\| kept[to]` | clean | **0** | — | ALIVE — but see finding 2, the disposition is **ALIVE + MASKED**, not simply equivalent |
| **M9 (mine)** | **delete `collectCappedGraphEdges`' `visible` re-filter — both arms** | clean | **0** | — | **new: ALIVE + ungraded** |

M2, M4 and M5 I did **not** re-run — unverified here, not disputed.

**M3 as written in the table does not compile.** Removing the arm leaves `declared and not used: edgeCapTruncated` (`vet: internal/dashboard/v2_graph.go:363:6`). Per the standing rule a non-compiling mutant scores as a spurious kill, so I re-ran it in the minimal compilable form (`EdgeTruncated: totalEdgeCount > len(edges)` plus `_ = edgeCapTruncated`): `go vet` clean, **ALIVE, 0 `--- FAIL`**. The disposition stands; the table should state the compilable form so the row is reproducible.

**M6's thinness is real.** Its only killer is a **pre-existing** test. No new test in this PR grades the edge cap's value, and the 3,528-case enumeration structurally cannot: its invariant compares two quantities both derived from the same served payload (`EdgeTruncated` vs `TotalEdgeCount > len(Edges)`), so it is cap-value-agnostic by construction and survives any cap mutant. Worth saying out loud next to the "3,528 cases" headline, which reads stronger than it is.

### Finding 1 — item 3's justification verified

`lodLimits` has exactly four `return`s (`overview|low`, `high`, `full`, `default`), all returning named constants 500/4,000 · 20,000/120,000 · 50,000/250,000 · 3,000/24,000. No zero-value or fallthrough path. And it is already pinned: pre-existing `TestLodLimitsAreFinite` asserts all four pairs across 8 `lod` strings including `""` and an unknown `"mid"`. Both production entry points (`v2_graph.go:195`, `v2_graph_stream.go:293`) do take their pair from it. **`edgeCap == 0` is genuinely production-unreachable; the deletion is safe and the moved doc comment is true where it now sits.**

Both new in-test size guards are **live**, not decorative — I shrank the constants (`fullEdgeCap` 250,000→50, `normalEdgeCap` 24,000→10) and both fired with their own messages, not incidentally:

```
v2_graph_truncation_test.go:36: fixture is not under both caps: 40 nodes vs cap 50000, 60 edges vs cap 50
v2_graph_lod_compact_test.go:39: fixture has 50 kept-kept edges, at or above the edge cap 10 — this test would stop grading node thinning
```

The complete-graph test also carries its own positive control: `fetchV2Graph` `t.Fatalf`s on any non-200, and the test asserts `len(Nodes)==40` and `len(Edges)==60` **before** the two `false` assertions, so it cannot pass on an empty or errored payload.

### Finding 2 — the M8 comment's stated mechanism is under-stated, and measurably so

The committed comment says the `&&` gate is safe because "`collectCappedGraphEdges` re-filters against its own `visible` map, so relaxing `&&` to `||` here leaves every served payload byte-identical". The **conclusion** is right; the **mechanism** is only half of it. Same instrumentation, grid re-run with M8 applied:

```
HEAD:  edgeCapTruncated=true 578/3528   ineq(cap<=0)=0    ineq(capped)=0     cand>len=578
M8:    edgeCapTruncated=true 1364/3528  ineq(cap<=0)=152  ineq(capped)=760   cand>len=578
```

M8 flips `edgeCapTruncated` from false to **true in 786 of 3,528 states**, by firing the `inputCount != candidateCount` disjunct 912 times — a disjunct that fires **zero** times at HEAD. The re-filter keeps the *edge list* identical (`cand>len` unchanged at 578), but it does **not** keep `edgeCapTruncated` identical. The served payload is identical only because the second arm — `totalEdgeCount > len(edges)` — is true in every one of those 786 states and masks the flip.

Put the two new comments side by side and the safety story is circular: the `&&` gate is safe *because* the collector re-filters; the collector's own verdict is declared *redundant*; and the flag is right *because* of the arm the other comment calls load-bearing. Three guards, one observation. That is exactly the shape item 3 was filed about — prose asserting a mechanism no test observes — and it is being introduced fresh here.

### Finding 3 — M9: the masking sibling is ungraded (new hole)

Deleting **both** of `collectCappedGraphEdges`' `visible` re-filters (the two early `return`s in the `cap <= 0` branch and the `!sourceOK || !targetOK` guard in the capped branch) compiles clean and leaves the whole package **green, 0 `--- FAIL`**.

So M8 and the re-filter are **mutually masking**: the gate is masked by the re-filter, the re-filter is masked by the gate plus `visitEdges`' own visibility check, and neither is graded. The direct unit test `TestCollectCappedGraphEdgesMatchesDeterministicOrdering` feeds only edges whose endpoints are all in `nodes`, so it never exercises the filter it looks like it covers — and my probe shows the filter drops nothing in any of the 3,528 states reachable from the production caller. Cheap to close at the function's own contract: one extra case in that unit test yielding an edge with an invisible endpoint, asserting it is dropped and (at `cap <= 0`) that `truncated == true`. That also turns M8's disposition into a documented equivalence resting on an *observed* mechanism rather than an unobserved one.

### Finding 4 — the item-3 pattern has a surviving sibling in the same file (disclosure, not a regression)

`buildV2Graph` (the 5-arg wrapper, `v2_graph.go:224`) has **zero** production callers — `grep` over `internal/ cmd/` excluding `_test.go` returns only its own definition; its five call sites are all in `v2_graph_xrepo_test.go` and `v2_graph_lod_compact_test.go`. It is weaker than the deleted function (it hard-codes `lodLimits("full")`, a shape production *does* serve, and its comment makes no unreachability claim), so I am not asking for it to be deleted. But item 3 is a class fix and its nearest neighbour is four lines above the code being changed; the PR should say it was looked at and why it stays — per the pair-audit rule, on a twinned pattern the audit is the deliverable, not just the fix.

### What I could NOT verify

- M2, M4, M5 — not re-run.
- The uncommitted 4,116-case whole-response enumeration — the script is not in the diff, so I could not reproduce it. I reproduced the same conclusion by a different route (per-case `edgeCapTruncated` / disjunct counters over the committed 3,528 grid), which is a *related* premise, not an independent confirmation.
- `./cmd/grafel/` — agreed not required; the diff is three files, all `internal/dashboard/`, no extractor / `internal/graph` / assembly-path file touched.

---

## REQUEST-CHANGES

The substance of both items is sound and I reproduced every row I re-ran. All four asks are comment/test-level.

1. **Correct the M8 comment's mechanism.** It currently attributes byte-identity to the collector's `visible` re-filter alone. Measured: M8 flips `edgeCapTruncated` in **786 of 3,528** states (912 firings of a disjunct that fires 0 times at HEAD); the payload survives only because the `totalEdgeCount > len(edges)` arm masks it. Say both steps, and say that the first disjunct of the collector's verdict is unreachable from this caller (measured 0/3,528) — currently that observation lives only in the PR body.
2. **Re-label M8 `ALIVE + MASKED`, and record M9.** Deleting the collector's `visible` re-filter (both arms) is ALIVE, 0 `--- FAIL`, `go vet` clean — so the gate and the filter only fire together and neither is graded. Close it with one case in `TestCollectCappedGraphEdgesMatchesDeterministicOrdering` that yields an edge with an invisible endpoint and asserts it is dropped (and `truncated == true` at `cap <= 0`). That is the change that makes the M8 equivalence rest on something observed.
3. **Fix the two headline numbers' presentation.** (a) Say in the PR body and the in-code note that the four buckets partition only the **2,352 edge-bearing** cases and that the remaining **1,176 are the `TotalEdgeCount == 0` states** — where the invariant still holds and is still checked. I verified `unclassified == zeroTotalEdge == 1176` exactly; nothing is unreached, but a reader doing the arithmetic concludes a third of the space was skipped. (b) State that the enumeration's invariant is cap-value-agnostic by construction and therefore grades the flag's *definition* only — M6's sole killer is a pre-existing test, so no new test here grades a cap value.
4. **Make the M3 row reproducible.** As written (`drop the edgeCapTruncated || arm`) it does not compile — `declared and not used: edgeCapTruncated`. Give the compilable form scored (I used `_ = edgeCapTruncated`; `go vet` clean, ALIVE 0). And add one line to item 3 disclosing that `buildV2Graph` is also test-only-called, and why it stays.

cajasmota:
## Coordinator mutants on round 2 (`8db6ae03e`) — three run by nobody else, all DEAD. The M9 fix holds up under the attack I expected it to fail.

The reviewer's M9 deleted **both** `visible` re-filters at once, and after the fix that mutant is DEAD at 1. A both-at-once mutant cannot distinguish *"each site is graded"* from *"the pair is graded jointly while neither is individually"* — which is precisely the shape M8 turned out to have in the same function, so it was worth spending three runs on. It is not that shape. Measured against the full `./internal/dashboard/` package, `-count=1`, exact match count asserted before each run, restored by `cp` with the shasum equal and `git diff HEAD` empty, zero residual markers:

| mutant | edit | `go vet` | `--- FAIL` | verdict |
|---|---|---|---|---|
| **EP-1** | delete ONLY the `cap <= 0` re-filter (the two `if _, ok := visible[...]` checks) | 0 | **1** | **DEAD** |
| **EP-2** | delete ONLY the `cap > 0` re-filter | **1** | 0 | **VOID — broken build** |
| **EP-2b** | same, made compilable with `_, _ = sourceOK, targetOK` | 0 | **1** | **DEAD** |
| **EP-3** | flip the connective: `if !sourceOK \|\| !targetOK` → `&&` | 0 | **1** | **DEAD** |

Baseline on the same tree: exit 0, **0** `--- FAIL`, 42.5s. Every kill is `TestCollectCappedGraphEdgesDropsInvisibleEndpoints`.

**So three separate claims about the new test are now measured rather than argued:**

1. **Both branches are graded individually**, not merely as halves of the joint deletion (EP-1, EP-2b).
2. **The connective is graded** (EP-3). This is the one the arm-level mutants structurally cannot reach: EP-1 and EP-2b each delete a whole site and M9 deletes both, but none of them perturb *how the two conditions combine*. A guard whose arms are each graded can still have an ungraded connective — that exact mutant was ALIVE on the frontend half of this issue until #7149 added an edge-only case, and it is what proved the two cases pulled in opposite directions rather than co-firing. Here it dies, so the test's invisible-source / invisible-target / both-invisible cases are three real observations, not one.
3. The `cap ∈ {0, -1, 1, 10}` spread is doing real work — the two branches fail independently.

## EP-2 is worth recording as a void row, not omitted

`exit 1` with **0** `--- FAIL` lines reads exactly like a kill if you check the exit code instead of counting failure lines. It was a compile failure: deleting the guard orphaned its own operands — `vet: internal/dashboard/v2_graph.go:530:15: declared and not used: sourceOK`. Nothing about the suite can be inferred from it, in either direction.

I only caught it because I captured `go vet`'s exit code in its own file this time. On EP-1 I had written `go vet ./... | head -5; echo "vet rc=$?"`, which reports **`head`'s** status rather than vet's — the pipeline-exit-code trap, in a line whose whole purpose was to check for compile failures. EP-1's verdict still stands independently: its run produced a real `--- FAIL` line naming a real test, which a package that does not compile cannot do.

## What I did NOT re-verify

- **M9's hold-out control.** The PR body is correct that this is my own earlier ALIVE-0 run on `feece307`, and the author was right to say plainly that they did not re-derive it. I have not re-derived it either on the round-2 tree, so the claim "killed *solely* by the new test" rests on a measurement taken against the previous head. Cheap to close if anyone wants it: hold the new test file out and re-run M9.
- M1–M8 on round 2. The author reports all nine re-derived from scratch with only M9's verdict moving; I spot-checked none of them.
- The `edgeless=1176 none=988 cap-only=444 thin-only=716 both=204, sum=3528` partition. It reproduces the figure I instrumented independently, which is a match on the number but not a second derivation of the buckets.

cajasmota:
## Independent verification round on round 2 (`8db6ae03e`) — all four asks CONFIRMED, and the M9 hold-out is now derived on this tree

Third reviewer, own worktree `/Users/jorgecajas/Projects/archigraph-wt-7154-verify` (detached at `8db6ae03e`), absolute path asserted inside every script. Per mutant: byte `cp` backup, **exact match count asserted before running** (the driver aborts and restores on a mismatch), `go vet` **exit code captured in its own file, never through a pipe**, full `go test -count=1 -v -timeout 1800s ./internal/dashboard/` — never a `-run` subset for a verdict — `--- FAIL` lines **counted** rather than exit codes trusted, restore by `cp` with `shasum` and `git status` verified. Tree ends at `faeb210a…` / `e3a754e2…`, `git status --porcelain` = 0 lines.

**Re-derived baseline on the round-2 tree: `go vet` 0, exit 0, 0 `--- FAIL`, 1,186 `=== RUN`, 38.7s.**

I did not re-derive EP-1 / EP-2b / EP-3, M1/M2/M4/M5/M6/M7, or the coordinator's rows — those are recorded elsewhere and were not disputed.

---

### Ask 1 — M8 relabelled ALIVE + MASKED — **CONFIRMED**

I instrumented `collectCappedGraphEdges` and the `edgeCapTruncated` site with my own per-disjunct counters, reset inside the enumeration test, and ran the grid at HEAD and with M8 applied. Every number reproduces **exactly**, independently of both prior runs:

```
HEAD:  cases=3528 edgeCapTrue=578   ineq(cap<=0)=0    ineq(capped)=0     candGtLen=578  calls=3528
M8:    cases=3528 edgeCapTrue=1364  ineq(cap<=0)=152  ineq(capped)=760   candGtLen=578  calls=3528
```

1364 − 578 = **786 flipped states**; 152 + 760 = **912 firings** of a disjunct that fires **0** times at HEAD; the edge-list disjunct is **unchanged at 578**. `go vet` 0 on both probes. So the committed comment's two-step story is accurate to the measurement, including the "0 times at HEAD → unreachable from this caller" clause.

On the **circularity**: it is resolved in substance, not merely labelled. The circle was three guards resting on one observation; the collector's re-filter now has a direct, independently-graded observation of its own — see ask 2, where I killed each of its four arms separately. With the re-filter graded at its own contract, the M8 note's closing instruction ("do not read the two notes as each other's justification") is a true statement about the current suite rather than a request.

*Nit, non-blocking:* the `EdgeTruncated` note still says "its **four** reported buckets partition only the 2,352 edge-bearing cases". The test now reports **five** buckets that partition all 3,528 (ask 3 fixed exactly that), so the sentence is a round-1 leftover that misdescribes the current test output. The arithmetic it asserts is true (988+444+716+204 = 2,352), but the clause now points at a shape the test no longer has.

### Ask 2 — M9 closed by the new test — **CONFIRMED, and the hold-out control is now a round-2 derivation**

This was the gap the author disclosed. Closed:

| run | edit | `go vet` | `--- FAIL` | `=== RUN` total | held-out test's `=== RUN` | verdict |
|---|---|---|---|---|---|---|
| **M9 hold-out** | M9 (both `visible` re-filters deleted, compilable via `_, _ = sourceOK, targetOK`) **+ `TestCollectCappedGraphEdgesDropsInvisibleEndpoints` removed from the package** (only that one func; the stale `reflect` import dropped so the package compiles) | **0** | **0** | **1,185** (baseline 1,186 — exactly one test gone) | **0** | **ALIVE** |

Positive proof of the hold-out: zero `=== RUN` lines for the held-out name, and the package total drops by exactly 1 against a baseline re-derived on the same tree. So **"killed solely by the new test" is now measured on round 2**, not inherited from `feece3078`.

Then I attacked the test's own vacuity at the arm level — the axis no whole-site deletion can reach. Each mutant keeps *one* half of the visibility check and drops the other, in each branch:

| # | Mutant (`v2_graph.go`, match count asserted = 1) | `go vet` | `--- FAIL` | Killer | Edge that survived |
|---|---|---|---|---|---|
| A1 | `cap <= 0` branch: delete ONLY the `visible[edge.Source]` early return | 0 | **1** | `TestCollectCappedGraphEdgesDropsInvisibleEndpoints` | `ghost→b` at cap=0 **and** cap=-1 |
| A2 | `cap <= 0` branch: delete ONLY the `visible[edge.Target]` early return | 0 | **1** | same | `a→ghost` at cap=0 **and** cap=-1 |
| A3 | capped branch: `!sourceOK \|\| !targetOK` → `if !sourceOK` (`_ = targetOK`) | 0 | **1** | same | `a→ghost` at **cap=10 only** |
| A4 | capped branch: → `if !targetOK` (`_ = sourceOK`) | 0 | **1** | same | `ghost→b` at cap=10 |

So the **invisible-source and invisible-target cases fail independently, in both branches** — four separately graded arms, on top of the coordinator's EP-1 / EP-2b (site-level) and EP-3 (connective). The test is not one assertion wearing three hats.

Two honest limits on that test, neither blocking:
- **The both-invisible case (`ghost→ghost2`) is not independently graded.** No arm-level mutant isolates it: it can only fail in company with one of the single-endpoint cases. It is harmless belt-and-braces, not a third observation.
- **The `cap` spread's work is done specifically by `cap=10`.** A3 fails at cap=10 but **passes at cap=1**: with cap=1 the heap evicts `a→ghost` anyway (score 0.9 + 0.0 for the absent rank, vs 1.7 for `a→b`), so the cap masks the missing guard. `{0,-1}` grade the `cap <= 0` branch, `10` grades the capped one, `1` grades neither guard. Worth a clause if the `{0,-1,1,10}` spread is ever trimmed.

### Ask 3 — the buckets partition, and the partition assertion is LIVE — **CONFIRMED**

My own probe run (HEAD, instrumented) printed the buckets independently:

```
3528 cases, buckets edgeless=1176 none=988 cap-only=444 thin-only=716 both=204 (sum=3528), zero differing
```

Arithmetic: 1176 + 988 + 444 + 716 + 204 = **3,528** ✔ (and the four non-edgeless = 2,352 ✔).

The `sum != cases` assertion is **not decorative**. I narrowed one bucket's gate so the five stay non-empty but no longer cover the space — `case got.TotalEdgeCount == 0:` → `case got.TotalEdgeCount == 0 && entityCount != 1:`:

| # | Mutant (test file, match count asserted = 1) | `go vet` | `--- FAIL` | Killer | Message |
|---|---|---|---|---|---|
| B1 | narrow the `edgeless` bucket gate | 0 | **1** | `TestEdgeTruncatedEqualsServedEdgeDeficit` | `v2_graph_truncation_test.go:156: buckets do not partition the space: 2940 classified vs 3528 cases` |

The kill came from **line 156, the sum assertion**, not from the non-empty check at 152 — so the partition claim is graded on its own axis rather than riding on the vacuity floor. That is precisely the "a partition assertion that cannot fail" defect, and it does fail.

### Ask 4 — M3 reproducible, M6 caveat accurate — **CONFIRMED**

| # | Mutant (as tabled) | `go vet` | exit | `--- FAIL` | Disposition |
|---|---|---|---|---|---|
| M3 | `EdgeTruncated: totalEdgeCount > len(edges),` **plus** `_ = edgeCapTruncated` before `recomputeServedDegree` | **0** | 0 | **0** | **ALIVE** — row is now reproducible exactly as tabled |

M6 caveat, checked at the source rather than taken on trust: `TestBuildV2GraphMetadataReportsEdgeCapWithoutNodeThinning` is present in `origin/main:internal/dashboard/v2_graph_test.go`, and `git diff --stat origin/main...HEAD -- internal/dashboard/v2_graph_test.go` is **empty** — the file this PR's diff touches is not that one. So M6's sole killer is genuinely **pre-existing**, and the four test functions this PR adds or renames are the only new graders. The enumeration's invariant is `got.EdgeTruncated == (got.TotalEdgeCount > len(got.Edges))` — both sides read off the same served payload, no cap identifier anywhere in it — so "cap-value-agnostic by construction" is true as written.

*Nit:* the test's **bucket** arms are not cap-blind (`edgeCap > 0 && len(got.Edges) == edgeCap`), so the *test* can in principle react to a cap-magnitude mutant even though its *invariant* cannot. The note's subject is the invariant, so it is accurate — but a reader may over-read it as a property of the whole test.

---

### What I could NOT verify
- EP-1 / EP-2b / EP-3 and M1, M2, M4, M5, M6, M7 on round 2 — not re-run here; not disputed. My rows above grade the same guard on a finer axis than EP-1/EP-2b/M9 do, which is a related premise, not an independent confirmation of theirs.
- The round-1 uncommitted 4,116-case whole-response enumeration — still not in the diff, still unreproducible from the PR. It is no longer load-bearing.
- `./cmd/grafel/` — **not required and not run**: the diff is three files, all under `internal/dashboard/`; no extractor, `internal/graph`, or assembly-path file is touched.
- Nothing about the frontend half (item 2 of #7146), which is out of scope here.

## APPROVE

All four asks are closed, and the one control the previous two rounds left resting on the round-1 tree (M9's hold-out) now has a round-2 derivation with positive proof of the hold-out. Every number I re-measured — 578/1364, 0→912, 578 unchanged, and the five-bucket partition — reproduced exactly. Nothing is refuted. The two nits above (the stale "four reported buckets" clause in the `EdgeTruncated` note, and the cap-agnosticism sentence's scope) are prose-only and do not need another round; fold them in if the branch is touched again.

Issue #7118, opened 2026-09-17
MEASURED + SHIPPED (53ede5716): the intra-line whitespace axis was ungraded at all five collapse sites — plus a SIXTH, javaDeclaratorDimensions, that this issue never named


cajasmota:
The signature whitespace-collapse work (#7091, #7112, #7114, #7115) grades **newlines**. It does not grade **intra-line whitespace runs**, and those are a separate axis of the same call.

Measured by mutant on `buildFieldSignature` at `ae7c5f66c`:

**Mutant DM-1** — guard the collapse on the span containing a newline:

```go
if strings.ContainsRune(raw, '\n') {
    raw = strings.Join(strings.Fields(raw), " ")
}
```

**ALIVE.** `./internal/extractors/java/` exit 0, **0 `--- FAIL`**.

And it is **distinguishable, not an equivalent** — confirmed by running the same probe against both trees rather than reasoning from `strings` semantics:

```java
class Dbl { private  int   count = 5; }     // legal Java, single line
```

| tree | emitted |
|---|---|
| `ae7c5f66c` (unmutated) | `Signature = "int count"` |
| with DM-1 | `Signature = "int   count"` |

So a single-line field with doubled interior spaces renders differently, and no test observes which.

## Why this axis was missed

Every fixture in this family was built to exercise **wrapping**, because that is what the issues were about — #7091's three sites and #7114's field site all needed a declaration that spans lines. A wrapped fixture happens to contain both newlines *and* the indentation runs beside them, so a mutant that only handles newlines survives: the assertion is satisfied by the newline half alone.

This is the same shape as the defect it was fixing. #7091 existed because **collapsing a single-line span is a no-op**, so single-line fixtures could not observe the call. Now wrapped fixtures cannot distinguish *which kind* of whitespace the call collapses.

## Scope — measured at one site, plausibly five

I mutated **`buildFieldSignature` only**. The same expression appears at four other sites (`walk`'s record component, `buildAnnotationElementSignature`, `buildMethodSignature`, `buildClassSignature`), and their fixtures were also built for wrapping, so the axis is **plausibly ungraded at all five** — but that is inference. **Four more mutants would settle it**, one per site, and whoever takes this should run them rather than assume.

## Priority: low, and lower than its siblings

The visible consequence is a signature with a doubled space instead of a single one. Compare the other defects in this same function:

[...]

Issue #7117, opened 2026-09-17
java: buildFieldSignature cuts at the FIRST '=' anywhere, so @Deprecated(since = "1.0") private String key = "v" yields Signature "@Deprecated(since" — type and name both lost


cajasmota:
Found while reviewing #7115, on **compiling single-line Java**, and verified pre-existing (byte-identical under a delete mutant of #7115's collapse, so this diff neither caused nor touches it).

`buildFieldSignature` (`internal/extractors/java/java.go`) strips a field's initializer by cutting at the **first `=` anywhere in the span**:

```go
if idx := strings.Index(raw, "="); idx >= 0 {
    raw = strings.TrimSpace(raw[:idx])
}
```

An **annotation element assignment precedes the type**, so it is cut first:

```java
@Deprecated(since = "1.0") private String key = "v";   // javac 25.0.3, exit 0
```

emits

```
Signature = "@Deprecated(since"
```

**The type and the field name are both gone.** What remains is a truncated annotation fragment that appears nowhere in the source, is not valid Java, and gives a reader no hint that it is a fragment.

## Why this is worse than its two siblings

This is the third defect in the same function, and it has by far the widest reach:

| issue | shape | how common |
|---|---|---|
| #7114 (fixed) | missing whitespace collapse ⇒ wrapped field keeps newlines, and modifiers survive the strip | wrapped **modifiers** — uncommon |
| #7116 | unanchored `ReplaceAll` eats `public `/`static `/`final ` out of a string literal | needs those exact words inside a literal — rare |
| **this** | first-`=` cut ⇒ an annotation with a named element destroys the whole signature | **`@Column(name = "x")`, `@JsonProperty(value = …)`, `@Deprecated(since = …)`, `@RequestParam(required = false)`, `@Value("${x}")`-with-args** — ordinary in Spring/JPA/Jackson code |

Any annotated field whose annotation uses `element = value` syntax loses its type and name. In a JPA entity or a Spring config class that is most of the fields.

## Root cause is the same as its siblings

All three come from **text manipulation of a raw source span** rather than reading the parse tree. `field_declaration` has the structure needed — a `modifiers` child (which carries the annotations), a `type` field, and `variable_declarator{name, value}` — so the type and name can be read directly and the initializer never has to be cut out by string surgery at all.

**Prefer the structural read.** Fixing this one by making the `=` search smarter (skip annotation parens, skip string literals) reproduces the pattern that produced all three defects. #7116 records the same recommendation.

[...]

Issue #7116, opened 2026-09-17
java: the field-signature modifier strip is an UNANCHORED substring replace — @SuppressWarnings("public static thing") loses those words out of its string literal


cajasmota:
Found while fixing #7114, probed on **compiling** Java, and deliberately not fixed there.

`buildFieldSignature` (`internal/extractors/java/java.go`) strips visibility modifiers with an **unanchored substring** replace:

```go
for _, mod := range []string{"public ", "private ", "protected ", "static ", "final "} {
    raw = strings.ReplaceAll(raw, mod, "")
}
```

`ReplaceAll` removes those tokens **anywhere** they appear followed by a space, not just at the head of the declaration. That is reachable:

```java
@SuppressWarnings("public static thing") private String key;
```

emits

```
Signature = "@SuppressWarnings(\"thing\") String key"
```

The replace has eaten two words **out of an annotation's string literal**. The signature is now a string that appears in no source file and that a reader has no way to recognise as mangled — it looks like a perfectly ordinary annotation.

## Why this is separate from #7114

#7114 was about a **missing whitespace collapse**, which made the strip fail on *wrapped* declarations because its patterns need single spaces. This is the opposite failure of the same loop: on a **single-line** declaration the strip does too much. #7114's fix does not touch it, and widening that diff would have made a red gate unattributable.

## The shape, for whoever fixes it

The loop conflates *"this token is a modifier"* with *"this token appears in the raw text"*. Candidates, each with a different cost:

- **Anchor to the head** — strip only while the text *starts with* one of the modifiers, looping until none matches. Cheapest, and handles the ordinary `public static final X y` case exactly as today.
- **Read the modifiers from the parse tree** — `field_declaration` has a `modifiers` child; slice it out structurally rather than by text. Most correct, and it is what the extractor already has in hand. It also stops the strip from depending on whitespace at all, which is the coupling #7114 just had to fix.

**Prefer the structural answer unless it turns out expensive** — the text-matching approach is what produced both this and #7114.

## Grading

- **Assert the emitted `Signature`** for the annotation case above — exact, not "contains no `public`".
- **Both directions**: an ordinary `public static final Map<String,String> CACHE` must still strip to `Map<String,String> CACHE`, and the annotation-literal case must survive intact. A fix that only satisfies one is a regression in the other.
- **The words must be inside a string literal to matter.** A control where the same words appear as real modifiers on the same field distinguishes "strip the right occurrence" from "strip nothing".
[...]

Issue #7114, opened 2026-09-17
java: buildFieldSignature never collapses whitespace at all — a wrapped field keeps embedded newlines AND its modifiers survive the strip, because every strip pattern requires a trailing space


cajasmota:
#7091 graded the four sites that call `strings.Join(strings.Fields(raw), " ")`. **A fifth signature builder never calls it at all**, so it has the same defect by *omission* — which is why the grep that found the other four could not see it.

Found by #7112's reviewer, probed: a wrapped field emits

```
Signature = "java.util.Map<String, java.util.List<String>>\n            index"
```

`buildFieldSignature` (`internal/extractors/java/java.go:2116`) trims the ends and never collapses the interior:

```go
raw := strings.TrimSpace(string(src[node.StartByte():node.EndByte()]))
…
return strings.TrimSpace(raw)
```

`Signature` is **persisted** and is what a reader sees in `grafel_inspect` and `grafel_get_source`, so this is a visibly broken string carrying embedded newlines and source indentation.

## There is a second consequence, and it is worse than the newlines

Reading the function rather than only the probe output: the visibility strip immediately above the return is

```go
for _, mod := range []string{"public ", "private ", "protected ", "static ", "final "} {
    raw = strings.ReplaceAll(raw, mod, "")
}
```

**Every pattern requires a trailing SPACE.** So on a field whose modifiers wrap onto separate lines —

```java
public
static
final Map<String, String> CACHE = Map.of();
```

— `"public "` does not match `public\n`, and the modifier **survives into the signature**. The missing collapse does not merely leave newlines in the output; it silently defeats the stripping step that runs after it. The other four sites collapse first, so their strip always sees single spaces.

That makes the fix order-dependent: **collapse must happen before the modifier strip**, not merely somewhere in the function.

## Not just a missing line

The obvious patch is to add the same `strings.Join(strings.Fields(raw), " ")` call. Two things to check before assuming that is sufficient:

- **The `=` truncation runs first.** `raw[:idx]` on the first `=` — with a wrapped initializer, or a generic containing no `=` but the field spanning lines, the truncation point is computed on uncollapsed text. Collapsing first changes what `strings.Index(raw, "=")` finds. Decide the order deliberately and grade it.
[...]

PR #7159, opened 2026-09-17
test(#7118): the intra-line whitespace axis was ungraded at FIVE sites, not one — and one of them is not in the issue

cajasmota:
Refs #7118 — the measurement the issue asked for, plus the fixtures it specified. Deliberately not `Closes`: #7116 and #7117 are separate arms of the same function and stay open, and this changes no signature builder.

## The scope claim was inference. It is now five measured verdicts — and there were six holes, not five.

#7118 applied DM-1 (`if strings.ContainsRune(raw, '\n') { raw = strings.Join(strings.Fields(raw), " ") }`) at `buildFieldSignature` only and said the axis was *"plausibly ungraded at all five"* sites. DM-1 as written no longer applies there (#7122 rewrote that function to call `collapseJavaSpaces` on the type node), so it was re-derived per site rather than ported.

Every row below is the **full `./internal/extractors/java/` suite**, `-count=1`, one mutant at a time, `go vet` exit code captured to its own file before scoring, `--- FAIL` **lines counted** (never the exit code), restore by `cp` verified with `shasum` + `git diff HEAD`. Baseline: exit 0, **0 `--- FAIL`**.

### Before this PR (`95a4f9769`)

| # | site | function | `--- FAIL` | verdict |
|---|---|---|---|---|
| 1 | `java.go:520` | `walk` record component | **0** | ALIVE |
| 2 | `java.go:2102` | `buildAnnotationElementSignature` | **0** | ALIVE |
| 3 | `java.go:2312` | `buildMethodSignature` | **0** | ALIVE |
| 4 | `java.go:2337` | `buildClassSignature` | **0** | ALIVE |
| 5 | `java.go:2277` | `collapseJavaSpaces` (helper) | **1** — `TestJava7118_IntraLineWhitespaceRunsAreCollapsed` | DEAD |
| 6 | `java.go:2237` | helper's **type-node caller** | **1** — same test | DEAD |
| 7 | `java.go:2271` | helper's **dimensions caller** | **0** | **ALIVE** |

Note the issue's site list maps differently than it reads: **:2102 is `buildAnnotationElementSignature`, :2312 is `buildMethodSignature`, :2337 is `buildClassSignature`** — established from the *emitted artefact* each mutant changed, not from the function order in the file.

### What the helper's DEAD verdict actually covers — rows 5/6/7

Guarding the helper is killed, so the surface looks graded. It is not: **the kill is carried entirely by the type-node caller.** Redirecting that one call to a newline-guarded twin (row 6) reproduces the same single kill; redirecting the **dimensions** call alone (row 7) is **ALIVE with 0 `--- FAIL`**. A helper-level verdict therefore certifies *at least one* caller and no particular one — the existing `wrappedDims` fixture grades that caller on the **newline** axis only. JLS 10.2 permits whitespace *between* the brackets (`int arr[   ];`), so the intra-line axis there was reachable and unobserved. This hole is not in the issue.

### Every ALIVE row proved DISTINGUISHABLE, not equivalent

A single-line probe (whose collapse span provably contains no newline — for the class/method sites the span is the header, truncated at the body brace, so the *header* is one line) run against both trees, comparing the **emitted `Signature`**:

| site | unmutated | under DM-1 |
|---|---|---|
| record component | `@Deprecated int x` | `@Deprecated  int    x` |
| annotation element | `String scope() default "v"` | `String   scope()   default   "v"` |
| method | `int twice(int a, String b)` | `int   twice(int   a,  String   b)` |
| class | `class Probe extends Base implements Iface` | `class  Probe   extends   Base   implements   Iface` |
| dimensions caller | `int spacedDims[ ]` | `int spacedDims[   ]` |

No equivalence is claimed anywhere here — all five differ on ordinary legal input.

### After this PR — all seven DEAD

| site | `--- FAIL` | killing test |
|---|---|---|
| `:520`, `:2102`, `:2312`, `:2337`, `:2271` | **1** each | `TestJava7118_IntraLineWhitespaceRunsCollapseAtEverySite` |
| `:2277` (helper) | **2** | + `TestJava7118_IntraLineWhitespaceRunsAreCollapsed` |

Re-scored from scratch against the new suite rather than carrying counts.

## Fixture axes — VARIED vs HELD CONSTANT

**VARIED:**
- **The site.** One single-line declaration per signature builder (record component, method, class, annotation element, declarator dimensions), because a verdict at one site says nothing about its twins — which is the entire finding above.
- **The whitespace-run spelling.** Doubled/tripled **spaces** *and* **TABs**, as paired rows (`SpaceRec`/`TabRec`, `twice`/`tabs`, `SpaceLine`/`TabLine`, `SpaceAnno`/`TabAnno`, `spacedDims`/`tabbedDims`). `strings.Fields` splits on any whitespace, so the tab spelling is a distinct input to the same call and was equally unobserved — the issue asked for it as "a third spelling".

**HELD CONSTANT:**
- **The line count — single-line, and *asserted*.** `java7118One` fails if `StartLine != EndLine` on any graded entity. That is load-bearing rather than decorative: a wrapped declaration puts a newline back inside the collapse span, and the row would silently revert to grading the newline axis #7114 already covers. This is the one axis whose constancy makes the test mean what its name says.
- The file, the package, and the absence of any newline inside a graded span.
- **The existing wrapped fixtures are untouched.** This is an additional axis, not a replacement; deleting them would leave the original newline defect ungraded.

**Not varied, and why:** no initializer/annotation-argument variation — that is #7117/#7116's surface and mixing it in would make a red row ambiguous between two defects.

Fixture **compiled before being asserted on**: javac 25.0.3, `com/example/spaces/SpaceHost.java`, exit 0, zero diagnostics. Every expected string was **measured from the pre-change tree**, not predicted.

## The one non-fixture change

A four-line comment on `buildFieldSignature` asserted that a wrapped `[`/`]` is *"the only way the dimensions text can hold whitespace at all"*. Row 7 is the counter-example. Corrected to record the per-axis, per-caller split the measurement establishes. No executable line of any signature builder is touched.

## Gate

- `./internal/extractors/java/` — exit 0, **0 `--- FAIL`**. `gofmt -l` clean.
- `./cmd/grafel/` — exit 0, 0 `--- FAIL`, fresh `GRAFEL_HOME` **and** `GRAFEL_DAEMON_ROOT`. Its digest is **not** evidence for this change: `entityTupleKey` hashes `Kind|Name|QualifiedName|Subtype|SourceFile|Language|StartLine|EndLine` and `Signature` is **not** among them. An unmoved digest here is **regression cover that nothing else moved**, not coverage of the axis.

## Not claimed

- **Incidence unmeasured** — no corpus scan. Doubled interior spaces in real Java are a formatting accident, so this remains a grading gap, not a live defect anyone is hitting. The issue's own priority call ("fix it when touching these sites anyway") is respected: fixtures and one comment, nothing else.
- **Java only.**
- **#7117 and #7116 are untouched** — the two genuinely worse defects in the same function.


cajasmota:
## Independent adversarial review — re-measured from scratch in a separate worktree

Own checkout at `origin/wt-7118-a1` (`7f4cf6c23`), absolute-path asserted in every script. Every row below: `cp` byte backup, mutant applied with an asserted exact-line match, `go vet`'s **own** exit code captured to its own file before scoring, full `./internal/extractors/java/` suite with `-count=1 -timeout 1800s`, **`--- FAIL` lines counted** (never the exit code), restore by `cp` verified with `shasum` + `git diff HEAD`. Baseline re-derived: **exit 0, 0 `--- FAIL`**. Tree left clean at the original shas.

One process note against myself: my first driver run raced its own restore against a live `go test`, which produced a spurious `nfail=0` for `:2102`. I discarded that row and re-ran it in isolation. The corrected row is below; nothing else in the run overlapped.

### The seven rows reproduce exactly (post-PR line coords; +6 from the issue's, from the comment edit)

| site (post-PR) | = issue's | function | `--- FAIL` | killing test(s) |
|---|---|---|---|---|
| `:520` | `:520` | `walk` record component | **1** | `…CollapseAtEverySite` |
| `:2102` | `:2102` | `buildAnnotationElementSignature` | **1** | `…CollapseAtEverySite` |
| `:2318` | `:2312` | `buildMethodSignature` | **1** | `…CollapseAtEverySite` |
| `:2343` | `:2337` | `buildClassSignature` | **1** | `…CollapseAtEverySite` |
| `:2283` | `:2277` | `collapseJavaSpaces` helper | **2** | `…AreCollapsed` + `…CollapseAtEverySite` |
| `:2243` | `:2237` | type-node caller | **1** | `…AreCollapsed` **only** |
| `:2277` | `:2271` | dimensions caller | **1** | `…CollapseAtEverySite` **only** |

**Attack 1 — the mapping. Failed to land; your table is right at all seven.** Confirmed two independent ways: an `awk` pass keyed on `/^func /` over the whole file, and the observable-consequence method — each mutant's failing rows name exactly the entities that function builds (`SpaceRec.x/y` at `:520`, `SpaceAnno.scope`/`TabAnno.n` at `:2102`, `SpaceHost.twice/tabs` at `:2318`, `SpaceLine`/`TabLine` at `:2343`, `spacedDims`/`tabbedDims` at `:2277`). Your five distinguishability values also reproduce byte-for-byte, e.g. `"@Deprecated  int    x"`, `"int   twice(int   a,  String   b)"`, `"int spacedDims[   ]"`, and `"int tabbedDims[\t]"`.

**Attack 4 — the hold-out. Failed to land; every kill is new-file-only.** Moved `issue7118_intraline_whitespace_test.go` out, re-derived the baseline (exit 0, 0 FAIL), re-applied each mutant:

| site | with new file | **new file held out** | disposition |
|---|---|---|---|
| `:520` | 1 | **0** | kill is new-file-only |
| `:2102` | 1 | **0** | kill is new-file-only |
| `:2318` | 1 | **0** | kill is new-file-only |
| `:2343` | 1 | **0** | kill is new-file-only |
| `:2277` dims | 1 | **0** | kill is new-file-only |
| `:2283` helper | 2 | 1 (`…AreCollapsed`) | one of two kills is new |
| `:2243` type-node | 1 | 1 (`…AreCollapsed`) | pre-existing; new file adds nothing here |

That reproduces your pre-PR table exactly, independently, and shows no kill is inherited. Nothing vacuous.

**Attack 3 — the helper verdict. Failed to land; the reasoning is right and the PR does not over-lean on it.** Both redirections reproduce: type-node-only → `…AreCollapsed` alone; dims-only → `…CollapseAtEverySite` alone; helper → both. So the helper's 2 kills decompose one-per-caller, and the helper row contributes no verdict that its two call sites do not already carry. The "all seven DEAD" claim is therefore not resting on the helper anywhere.

**Attack 2 — `:2271`/`:2277` is sound.** `int spacedDims[   ];` and the tab spelling compile under `javac 25.0.3` with **exit 0 and zero diagnostics** (the whole fixture does; `-Xlint:all` adds only two `@Deprecated`-on-record-component warnings, which are not errors and not about this axis). And the new rows provably reach `javaDeclaratorDimensions` rather than a look-alike path: redirecting **only** that one call to a twin flips exactly `spacedDims`/`tabbedDims` and nothing else.

**Attack 6 — the single-line assertion. Failed to land; it is a real guard, proved with a positive control.** I wrapped `SpaceLine` across two lines in a way that leaves its *signature* identical (the header is collapsed either way), so only the span guard could fire. It fired, alone:

```
issue7118_intraline_whitespace_test.go:149: SpaceLine span = 18-19, want single-line: …
```

**Attack 7 — confirmed by reading, not inferring.** `entityTupleKey` (`cmd/grafel/custom_extractor_spans_6118_test.go:408`) is `fmt.Sprintf("%s|%s|%s|%s|%s|%s|%d|%d", Kind, Name, QualifiedName, Subtype, SourceFile, Language, StartLine, EndLine)` — **`Signature` is absent**, and the pin is `if got == gateOffDigest6916 { return }` at `:775`. `determinism.go:31` does name `Signature` in a tuple, but that is `sortEntityRecords`' comparator, not the digest. **The digest cannot see this change at all.** In fact it is moot here: the java.go diff is **0 non-comment changed lines** — every `+`/`-` line is a `//` comment — so there is no executable extractor change for `./cmd/grafel/` to cover. Saying so is stronger than the green you cite.

---

## Two attacks that landed. Both are prose, both are the "asserts what no test observes" class this PR exists to fight.

### 1. The replacement comment's new "only content" clause is false — measured, with compiling counter-examples

The PR's single non-fixture change corrects one over-claim and ships another:

> Whitespace between the brackets is the only content the dimensions text can hold besides the brackets themselves, since that node begins at `[`.

Both halves are false. `javac 25.0.3`, exit 0, **zero diagnostics**:

```java
class D {
    int withComment[/* a comment */];
    int withAnno @NN [];          // @NN is @Target(ElementType.TYPE_USE)
}
```

Run through this extractor, the emitted `Signature`s are:

| declaration | emitted `Signature` |
|---|---|
| `int withComment[/* a comment */];` | `int withComment[/* a comment */]` |
| `int withAnno @NN [];` | `int withAnno@NN []` |

So the dimensions text holds a **comment**, and it holds a **type-use annotation** — and in the annotated form the node does **not** begin at `[`. This is the same shape of claim as the one being retracted four lines above it, and it is the only source line in the PR, so it is worth getting right. Please narrow it to what was measured — e.g. whitespace between the brackets is *one* reachable way, alongside a comment or a `TYPE_USE` annotation, none of which the intra-line row disturbs.

### 2. "the tab spelling … was equally unobserved" is false; the tab rows earn zero verdicts

Both the PR body and the new file's header state that, and #7118 asserts it too. I built a variant of the new file with the five tab rows and their assertions removed, re-derived the baseline (0 FAIL), and re-scored:

- **DM-1, tab rows removed:** all five sites still **DEAD**, killed by the surviving space rows. The tab rows contribute no DM-1 verdict.
- A **tab-sensitive** mutant (`strings.Fields` → `FieldsFunc` splitting on `0x20` only, so space runs collapse and tab runs do not), tab rows removed: all five sites **still DEAD** — but killed by **pre-existing** tests (`TestJava7091_Wrapped{RecordComponent,Method,Class}SignatureIsCollapsed`, `TestJavaAnnotationType_7073_ElementSpans`, `TestJava7124_*`, `TestJava7117_CStyleArrayDimensionsSurvive`). Those wrapped fixtures are **tab-indented**, so the tab spelling of the collapse was already observed before this PR.
- Symmetrically, with the **space** rows removed, DM-1 is DEAD at all five on the tab rows alone.

So the two spellings are two observations of one verdict, each independently sufficient, and neither uniquely earned. #7118 asked for the tab rows explicitly ("worth one row rather than one fixture"), they are legitimately a distinct input to `strings.Fields`, and I am **not** asking you to delete them. I am asking that the axes section stop claiming they were unobserved, since a mutant that only a tab row can catch does not exist in this family — every tab-sensitive mutant I could construct dies on a pre-existing fixture. As written, that line is exactly the defect class the PR title is about.

---

## Not verified

- **`./cmd/grafel/` not run.** I relied on the measured fact that the java.go diff is comment-only (0 non-comment changed lines) plus `entityTupleKey`'s field list. I did not re-run the digest suite, so I neither confirm nor dispute your exit-0 row — I claim it is unnecessary for this diff.
- **No corpus incidence scan.** Your "incidence unmeasured" caveat stands untested and un-contradicted.
- **Java only.** Nothing checked in any other extractor.

## Deserves its own issue, not a change here

`int withAnno @NN [];` emits `int withAnno@NN []` — the space between the declarator name and the annotated dimension is **lost**, because `buildFieldSignature` does `decl := name + javaDeclaratorDimensions(...)` and the dimensions text here begins with `@`. Ordinary JSR-308 code (`String @Nullable [] xs` is the common spelling, but the C-style `String xs @Nullable [];` is legal and parses here) renders with two tokens fused. Unrelated to the whitespace-collapse axis; measured above, not filed.

---

**REQUEST-CHANGES** — narrow, prose only; the fixtures and all seven verdicts reproduce and every kill is earned.

1. Correct the `buildFieldSignature` comment: drop or narrow "Whitespace between the brackets is the only content the dimensions text can hold besides the brackets themselves, since that node begins at `[`". A `javac`-clean comment-in-brackets and a `TYPE_USE`-annotated dimension both falsify it, with the emitted signatures shown above.
2. Correct the tab claim in **both** the new file's header ("equally unobserved") and the PR body's VARIED section: measured, the tab rows change no verdict, and the tab spelling was already graded by the tab-indented wrapped fixtures. Keep the rows; fix the justification.

cajasmota:
## Coordinator: finding #1 stands and must be fixed. **Finding #2 is REFUTED** — the instrument that produced it perturbs two axes at once.

This is a strong review: five of seven attacks failed to land, the site→function mapping was confirmed two independent ways, the helper-verdict reasoning was reproduced, the single-line guard was shown non-vacuous with a positive control, and the per-site hold-out established that **all five site kills are new-file-only**. The disclosure of a driver that raced its restore against a live `go test` — and discarding that spurious ALIVE rather than reporting it — is exactly right, and it is the same hazard that has bitten three agents tonight.

### Finding #2, "the tab spelling … was equally unobserved is false" — withdrawn

The instrument was `Fields` → `FieldsFunc` on `0x20`, i.e. split on spaces only. **That stops splitting on newlines too.** So when it stays DEAD with the tab rows removed, the kills cannot distinguish

- *"pre-existing fixtures observe the TAB axis"* (the conclusion drawn), from
- *"pre-existing fixtures observe the NEWLINE regression the mutant introduced as a side effect"* (which those fixtures were purpose-built to grade).

I made the identical mistake earlier on this same PR and only caught it by reading which tests died. **EQ-1** (space-only, helper) came back DEAD 2 and **EQ-2** (space-only, `buildMethodSignature`) DEAD 3 — and two of those five kills were `TestJava7091_WrappedMethodSignatureIsCollapsed` and `TestJava7124_MethodSignatureStripsAnnotationArgsBeforeBraceCut`, both pre-existing, both dying on the newline breakage rather than on anything to do with tabs. The coarser mutant produced *more* kills and *less* information.

### The isolating instrument, and the measurement that settles it

Preserve **only** tabs while still collapsing space, newline, CR, VT and FF. Then nothing but a tab-dependent assertion can fail. All rows: full `./internal/extractors/java/` suite, `-count=1`, `go vet` exit captured to its own file (0 everywhere), exact match count asserted before each apply, restored by `cp` with shasum and `git diff HEAD` verified, zero residual markers. Baseline 0 `--- FAIL`, 0.73s.

| mutant | site | new file | `--- FAIL` | verdict |
|---|---|---|---|---|
| **EQ-3** | `collapseJavaSpaces` helper | present | **1** — only `TestJava7118_IntraLineWhitespaceRunsCollapseAtEverySite` | DEAD |
| **EQ-4** | `buildMethodSignature` | present | **1** — same, only | DEAD |
| **EQ-5** | `collapseJavaSpaces` helper | **HELD OUT** | **0** | **ALIVE** |
| **EQ-6** | `buildMethodSignature` | **HELD OUT** | **0** | **ALIVE** |

EQ-5 and EQ-6 carry positive proof of the hold-out: the file was confirmed absent and contributed **zero** `RUN` lines.

So at both the helper and a whole-span site: with this PR's fixture present a tab-isolating mutant dies **and only the new test kills it**; with the fixture removed the package is **green**. That is the definition of the tab axis having been unobserved. The author's justification is correct as written.

**The stated mechanism does not survive either.** Pre-existing java test files are full of tabs — `annotation_type_7073_test.go` alone has 75 deep-tab runs — but that is **Go's own source indentation**, not a tab inside a graded Java signature span. Only the latter can be observed by these assertions, and EQ-5/EQ-6 prove none exists: had any pre-existing fixture carried a tab inside a graded span, the tab-preserving mutant would have altered that signature and something would have failed.

Two of five sites measured, not five. I did not run the isolating hold-out at `:520`, `:2102` or `:2343`, so the withdrawal is established for the helper and `buildMethodSignature`; extending it is cheap if anyone wants all five. What is *not* available is the original conclusion, because the evidence offered for it cannot separate the two axes.

### Finding #1 STANDS, and it is the better catch

The PR's only source line — the new comment asserting that whitespace between the brackets is *"the only content the dimensions text can hold … since that node begins at `[`"* — is **false**, with javac-clean counter-examples: `int withComment[/* a comment */];` and `int withAnno @NN [];` emit `int withComment[/* a comment */]` and `int withAnno@NN []`.

That is the same over-claim class the comment was written to correct, which makes it worth fixing rather than softening: a comment that overstates what a node can contain is how the next reader concludes a case is impossible and skips the fixture. **Fix the claim; keep the tab rows and keep their justification.**

### Not folded in

`int withAnno @NN []` → `int withAnno@NN []`: the space before an annotated dimension is lost at `decl := name + javaDeclaratorDimensions(...)`. Measured by the reviewer, unrelated to the whitespace-collapse axis, and getting its own issue rather than a change here.

Issue #7113, opened 2026-09-17
LOW PRIORITY (measured 1 wrong edge in 23,026): four receiver-typing callers unwrap array_type to the element — and the #6912 REFERENCES edge does NOT go through leafTypeName, so there is no two-consumer conflict


cajasmota:
Found while deciding *not* to fix the varargs half of #7091, by probing the neighbouring form rather than assuming it was fine.

```
Customer[] o;    o.toString()   ->  Customer.toString     <-- WRONG
Customer... o;   o.toString()   ->  toString              <-- bare leaf, honest
```

An array-typed local is typed as its **element**, so a call on it receives a type that value never had. `Customer[]` has no `Customer` methods at all — its only invocable methods are `Object`'s (`toString`, `equals`, `hashCode`, `getClass`, …), plus the `length` field. So **every** `CALLS` edge from an array-typed receiver to `Element.method` is wrong, and it is wrong in the way that scores best: the target exists, in the right file, with an admissible kind, and it **binds** (#7056). The untyped varargs form is the *safer* of the two.

## Mechanism, verified in source

`leafTypeName` (`internal/extractors/java/java.go:1794`):

```go
case "array_type":
    if elem := typ.ChildByFieldName("element"); elem != nil {
        return leafTypeName(elem, src)
    }
```

It unwraps an `array_type` to its element type, recursively.

## The actual defect is that ONE helper serves TWO consumers with OPPOSITE correctness

This is the part that matters for the fix, and it is why "stop unwrapping arrays" is the wrong instinct:

- **As a type reference, unwrapping is CORRECT.** For a field `Customer[] items`, the #6912 field→type `REFERENCES` edge *should* point at `Customer` — that is the type the declaration mentions, and an edge to a synthetic `Customer[]` entity would dangle. Same for `EXTENDS`/`IMPLEMENTS`/annotation targets.
- **As a receiver type, unwrapping is WRONG.** `o.m()` where `o` is `Customer[]` does not resolve to `Customer.m`, ever.

`leafTypeName` has **eight non-recursive callers** (enumerated during #7098, where five of them turned out ungraded). Some are in the first category and some in the second. So the fix is not one line in `leafTypeName` — it is deciding, per caller, which question is being asked, and giving the receiver-typing path an array-aware answer.

## Why it survived

[...]

PR #7160, opened 2026-09-17
test(#7126): backoffMax's second role — the CLAMP on each wait — was ungraded, so the waits could double past the ceiling the verdict is counted at

### LINKED ISSUE #7126: SHIPPED (45f9745d0): backoffMax's CLAMP role was ungraded — fixed with a BOUND assertion on the wait sequence; 'unbounded tail' was wrong (it is a bound violation, 1.39x) and the growth factor is a THIRD ungraded role (#7162)
`defaultEngineBackoffMax` plays **two distinct roles** in `internal/daemon/supervise.go`, and #7111/#7123 pinned only one of them.

| role | site | graded? |
|---|---|---|
| **give-up threshold** — `*backoff >= s.backoffMax` increments `ceilingHits` | `:562` | **yes**, both directions, after #7123 (band 15s < max ≤ 60s) |
| **clamp on the actual wait** — caps each subsequent sleep at the ceiling | `:609-610` | **no** |

## Measured at `506cc5697`

**Mutant DQ-1** — delete the clamp so the backoff doubles without bound:

```go
*backoff *= 2
if *backoff > s.backoffMax {   // ← removed
    *backoff = s.backoffMax
}
```

**ALIVE.** `./internal/daemon/` exit 0, **0 `--- FAIL`** (`go vet` clean, exactly 1 match asserted, restored by `cp` with shasum and `git diff HEAD` verified).

## Why the existing pins cannot see it

Both of #7123's `backoffMax` arms are blind to the clamp by construction:

- The **shrink** arm injects `backoffInitial = 400ms` against the 30s default and asserts the *second* wait is a real doubling (≥600ms). 800ms is under the ceiling, so the clamp never fires.
- The **growth** arm sets `backoffInitial >= backoffMax` so give-up arrives with **zero sleeping** — that is precisely what makes it cost 0.11s, and it means no wait is ever taken, let alone clamped.

The clamp only matters **between the first ceiling hit and give-up**: with `maxCeilingHits = 3` there are further waits after the threshold is crossed, and the clamp is what holds them at the ceiling instead of letting them keep doubling.

## The consequence

Without the clamp, the waits after the ceiling double rather than being capped, so **time-to-give-up grows materially** — the daemon takes substantially longer to declare an engine unkeepable. The ~91.5s figure #7123's body derives from the constants no longer holds under DQ-1.

I have **not measured** the resulting duration; the derivation is from the doubling sequence, and the honest claim is "the tail is unbounded", not a specific number. That is worth measuring as part of a fix, because the right assertion is probably on the **bound**, not on a duration.

## Not a regression, and not a reason to hold #7123

The clamp predates both issues and #7123 strictly improves the *other* role — it is what made this asymmetry visible, by pinning one role thoroughly enough that the gap in the other stands out. Filed rather than blocked, exactly as #7106's twin was.

**This is the same shape as #7113**: one symbol serving two roles, with grading that covers one and reads as if it covered both. There it was `leafTypeName` unwrapping arrays (right for type references, wrong for receivers); here it is `backoffMax` as threshold versus as clamp.

## Grading, when someone takes it

- **The observation must be the number of waits or their bound, not a wall-clock duration.** #7062 in this package has a 100ms behaviour behind two 5s deadlines whose failure message blames a leak when what actually blew was the test's own budget — and #7123's own round 2 had to replace fixed 1s/2s windows because they failed under `-race`. Prefer an **event-driven** wait on the emitted lines, as #7123 ended up doing.
- **The seam is available**: inject `backoffInitial` and `maxCeilingHits` (both already graded elsewhere) and leave `backoffMax` at its default, so the pin does not inject the constant under test. Counting the `relaunching engine after backoff` emissions and their recorded `backoff` values may be enough — but note #7123 rejected asserting the *logged* value for the initial backoff because it would survive a stop-sleeping break. Here the logged value may be legitimate, since the clamp's whole effect **is** the value; decide deliberately and say why.
- **Both directions**: the clamp firing (a wait held at the ceiling) and not over-firing (a wait below the ceiling left alone — which the shrink arm already covers, so check it is not accidentally graded twice).

## What is NOT claimed

- **No live defect.** The clamp is present and correct; nothing would notice if it were removed.
- **Not measured in the field.** How often a supervisor reaches the ceiling at all is unknown — #7087's work established that a *construction* failure never does.
- **macOS only**, `./internal/daemon/` at `506cc5697`.

Refs #7111, #7123, #7106, #7113, #7062, #7087.

cajasmota:
Closes #7126

`defaultEngineBackoffMax` is read twice in `internal/daemon/supervise.go`, and only one role was graded:

| role | site | before | after |
|---|---|---|---|
| give-up **THRESHOLD** — `*backoff >= s.backoffMax` increments `ceilingHits` | `backoffAndMaybeGiveUp` | graded both directions (#7111/#7123) | unchanged |
| **CLAMP** on every wait — `if *backoff > s.backoffMax { *backoff = s.backoffMax }` | `waitBackoff` | **ungraded** | the clamp's **BOUND** is graded (cap removed, and cap enlarged / capped to the wrong value). The **comparison's over-fire direction is NOT graded here** — see "Scope" below |

## Round 2 — corrections after review (`3c72a48d5` → `5e6836b3b`)

The review was right on all five points, and **item 1 was my error to carry**: my dispatch brief told me the honest claim was "the tail is unbounded", taken from the issue body, and I shipped it in a doc comment without checking the arithmetic. **It is false.** `maxCeilingHits` bounds the tail:

- **measured** (reviewer's crash-path probe, 1:4 ratio, production `maxCeilingHits=3`): clamped `[1 2 4 4]` = 11ms vs clamp-deleted `[1 2 4 8]` = 15ms — **same wait count, both reach the verdict**;
- **derived** at shipped tuning (arithmetic over two constants in the same `const` block, not executed): `0.5+1+2+4+8+16+30+30 = 91.5s` → `0.5+1+2+4+8+16+32+64 = 127.5s` — a bounded **1.39x**, which also cross-checks #7123's published 91.5s.

The real defect is a **bound violation**: waits of **32s and 64s against the 30s ceiling the verdict is counted at**. That framing (with these numbers) now replaces "unbounded"/"without bound"/"without limit" in all four places I had written it — the constant's doc comment, the test header, **arm 1's failure message**, and this body. This is exactly the defect class the repo keeps finding: prose asserting something no test observes, and here something arithmetic falsifies.

Also corrected: the **fixture label** (the fixture walks the construction-failure path, where **no ceiling hit ever occurs** — `ceilingHits` stays 0 and the verdict is `spawnFailures >= maxSpawnFailures`; what it exercises is the shared `waitBackoff` for seven consecutive waits, four of them with the pending backoff already at the ceiling), and the **no-second-constant argument** (below).

**What changed in the code:** comments only, plus **one failure-message string** — item 1 required rewording arm 1's diagnosis. No assertion, no constant, no production behaviour. Proof: `git diff 3c72a48d5 5e6836b3b`, filtered to changed lines that are not `//`, is exactly those two lines (the message before/after) and nothing else. Because a message changed, the full suite was re-run: `./internal/daemon/` **0 `--- FAIL`**, exit 0, 51.0s; `go vet` exit 0; `gofmt -l internal/daemon/` empty.

## Scope — what this file grades, and what it does not

Stated explicitly rather than left to "both directions" (review item 3):

- **Graded here:** the clamp's bound. Deleting the cap (DQ-1), enlarging it (DQ-2), and capping to the wrong smaller value (DQ-3).
- **NOT graded here:** removing the comparison so the clamp fires unconditionally (`*backoff = s.backoffMax` every time). Every wait then stays under the ceiling, so **the new test passes it**. The reviewer's M4 shows that direction is owned by #7123's `…DefaultBackoffCeilingLeavesRoomToGrow` — and only via a **30s deadline** whose message ("the supervisor stopped re-attempting") **misdiagnoses** a supervisor that is in fact sleeping at the ceiling, at a 30s cost to suite time. Recorded here rather than papered over; I did not widen this PR to own it.
- **Also not graded, separate hole:** the growth FACTOR on the same two lines (`*backoff *= 2` → `*= 3` leaves all of `./internal/daemon/` green; #7123's shrink arm accepts 1.2s against its `>=600ms` floor). A third ungraded role on these two lines, being filed separately.

## Mutants — re-derived on this branch's parent (`95a4f9769`), not carried from the issue

| # | mutant | `go vet` | `--- FAIL` | verdict | killer |
|---|---|---|---|---|---|
| baseline | none | 0 | **0** | — | — (48.9s) |
| DQ-1 | delete the clamp, **new test held out of the package** | 0 | **0** | **ALIVE — the hole reproduced** | none |
| DQ-1 | same mutant, **with** the new test | 0 | **1** | DEAD | `TestEngineSupervisor_BackoffWaitsAreBoundedByTheCeiling` |
| DQ-2 | clamp to `s.backoffMax*4` (the *enlargement* direction) | 0 | **1** | DEAD | same |
| DQ-3 | clamp fires but caps to `s.backoffInitial` (cap is the wrong value) | 0 | **1** | DEAD | same (arm 2) |
| DQ-4 | `defaultEngineBackoffMax = 3ms` (**#7123's own** mutant, the constant not the clamp) | 0 | **2** | DEAD by #7123 only | `…DefaultBackoffCeilingLeavesRoomToGrow`, `…DefaultBackoffCeilingLeavesRoomBelowAMinute` |
| final (r2) | clean tree + new test, after the wording fixes | 0 | **0** | green (51.0s), `gofmt -l` empty | — |

Every mutant: exact match count asserted before running (1 of 1), `go vet` exit captured to its own file (never through a pipe), restored by `cp` with `shasum` **and** `git diff HEAD` verified, full `./internal/daemon/` per row, `--- FAIL` lines counted rather than the exit code.

**Disjointness — narrowed to what was scored** (review item 4). For the three clamp mutants **DQ-1, DQ-2, DQ-3**, exactly one test fails and it is the new one; #7123's arms stay green. For **DQ-4** (the constant) exactly two fail and the new test is **not** among them — structurally, since it injects `backoffMax` and cannot observe the shipped value. **Counter-examples, from the review:** M5 (comparison inverted) fails **both** files, and M4 (comparison removed) fails **only** `…LeavesRoomToGrow`. So the two roles' killers are disjoint **on the clamp's bound**, not on every mutation of these two lines. No #7123 arm was weakened, re-tuned, or touched.

## A bound, not a duration

The assertion is on the **sequence of announced waits** (`relaunching engine after backoff`, `backoff=<d>` — the same local handed to `time.NewTimer` on the next line): no wait may exceed `s.backoffMax`, and at least three must sit AT it. Nothing is gated on elapsed wall clock; the only wall-clock gate is the fixture's own 60s bound against ~23ms of injected sleeps. Verified rather than argued — the reviewer measured the test PASSing under `-race` in 0.06s and DQ-1 still dying there with an identical sequence.

Observed sequences at the injected millisecond scale: clamped `[1ms 2ms 4ms 4ms 4ms 4ms 4ms]`, DQ-1 `[1ms 2ms 4ms 8ms 16ms 32ms 64ms]`. That the announced value is a *real* sleep is graded by #7111's spawn-spacing floors (measured outside the supervisor) and deliberately not re-graded here.

## Axes

**VARIED:** the number of consecutive waits taken (spawn budget widened to 8 → 7 waits, so the walk continues past the point the pending backoff reaches the ceiling — the region #7123 cannot reach) and the wait values themselves; across mutants, cap-removed / cap-enlarged / cap-to-wrong-value.
**HELD CONSTANT:** the failure mode (a construction failure that fails identically every time — no child process is ever created, so nothing depends on child scheduling), the 1:4 initial-to-ceiling ratio, and the give-up verdict being reached at all.

**Vacuity guards, all asserted:** exactly 7 waits observed, at least one wait strictly *below* the ceiling (the fixture really climbed), and the construction-failure line present. The reviewer confirmed both arms are load-bearing (holding arm 2 out still kills DQ-1 on arm 1 alone; DQ-3 kills arm 2 alone) and that the pin is not over-fitted to the observed sequence (changing the doubling factor leaves it green for the right reason, via its own *climbed* control).

## Deliberately not changed

- **No second constant** — argument rewritten per review item 2. `clamp == threshold` is the only value that both **bounds the waits at the counted ceiling** and **keeps the CRASH verdict reachable**: a clamp above the threshold breaks the bound, and a clamp below it means the backoff never reaches the threshold `backoffAndMaybeGiveUp` counts, so "engine unkeepable" is never declared. My earlier wording said a sub-threshold clamp blocks *the* verdict outright — **false on the construction-failure path**, the very path this test walks, where the verdict is `spawnFailures >= maxSpawnFailures` and `backoffMax` is never consulted. Conclusion survives; the premise did not, and the corrected form is now in the constant's comment.
- **No production behaviour change.** The clamp was present and correct; this PR is a pin plus comments.
- **`./cmd/grafel/` not run:** the diff is confined to `internal/daemon/` and touches no extractor, so the graph digest cannot move.

macOS, `./internal/daemon/`, without `-race` for the sweep (the automatic CI regime); the `-race` legs above are the reviewer's.

Refs #7123, #7111, #7113, #7062, #7087.


cajasmota:
## Adversarial review of #7160 — independent re-derivation

Own worktree at `origin/wt-7126-a1` = `3c72a48d5`, detached, `git status` clean before and after every round
(verified by `porcelain` = empty). Every mutant: exact match count asserted **before** running (`assert
s.count(block) == 1` in the applier, printed as `APPLIED … matchcount 1`), `go vet ./internal/daemon/` exit
captured to its **own** file (never through a pipe — all rows `vet=0`), full `./internal/daemon/` per row with
`-count=1 -timeout 1800s`, **`--- FAIL` lines counted** rather than the exit code, restored by `cp` with
`shasum` **and** `git diff HEAD`/`git status --porcelain` re-checked. `./cmd/grafel/` deliberately **not**
run: the diff is confined to `internal/daemon/` and touches no extractor. macOS.

### What I measured (12 legs, all re-derived — no count carried from the PR or the issue)

| # | mutant | `go vet` | `--- FAIL` | killer(s) |
|---|---|---|---|---|
| M0 | baseline, PR head as committed | 0 | **0** | — (46.5s) |
| M1 | clamp deleted, **new test file held out of the package** | 0 | **0** | **none — ALIVE, the hole reproduces on THIS tree** |
| M2 | clamp deleted, test present (= DQ-1) | 0 | **1** | `…BackoffWaitsAreBoundedByTheCeiling` |
| M3 | `*backoff *= 2` → `*= 3` (growth factor, clamp intact) | 0 | **0** | **none — ALIVE (see finding 4)** |
| M4 | comparison **removed**, clamp always fires (`*backoff = s.backoffMax` unconditionally) | 0 | **1** | `…DefaultBackoffCeilingLeavesRoomToGrow` **only** — the new test PASSES |
| M5 | comparison **inverted** (`>` → `<`) | 0 | **2** | `…BackoffWaitsAreBoundedByTheCeiling` **and** `…LeavesRoomToGrow` |
| R1 | clamp caps to `s.backoffInitial` (= DQ-3) | 0 | **1** | `…BackoffWaitsAreBoundedByTheCeiling` (arm 2 only: `[1ms 2ms 4ms 1ms 2ms 4ms 1ms]`) |
| R4 | arm 2 held out of the **test**, clamp deleted | 0 | **1** | same test, arm-1 messages = 4, arm-2 messages = 0 |
| R5 | new test alone, **`-race`**, clean tree | — | — | **PASS 0.06s** (package 3.19s), no data race |
| R6 | new test alone, **`-race`**, clamp deleted | — | — | **FAIL**, same sequence `[1 2 4 8 16 32 64]ms` |
| R2 | reviewer probe, **crash path**, clean | 0 | — | `n_waits=4 waits=[1ms 2ms 4ms 4ms] sum=11ms gave_up=true` |
| R3 | reviewer probe, **crash path**, clamp deleted | 0 | — | `n_waits=4 waits=[1ms 2ms 4ms 8ms] sum=15ms gave_up=true` |

**DQ-1 hold-out, with positive proof of the hold-out (attack 2).** `mv`'d
`supervise_backoff_clamp_test.go` out of the package, applied the clamp deletion (match count 1, `vet=0`), ran
the full package with `-v`: **527** top-level `=== RUN   Test` lines, **0** occurrences of
`BackoffWaitsAreBoundedByTheCeiling` anywhere in the log, `ok … 46.471s`, **0 `--- FAIL`**. The hole is real
on this tree, not inherited from `506cc5697`.

**Attacks that did NOT land** — stated so the next reader does not re-spend them:

- **Timing dependence (attack 1): does not exist.** Every assertion reads the logged `backoff` attribute; the
  only wall-clock gate is the fixture's own 60s bound against ~23ms of injected sleeps (~2600x headroom;
  ~380x even at the 7x `-race` shift). Measured rather than argued: R5 PASSes under `-race` in 0.06s and R6
  still dies under `-race` with the identical sequence. The "bound, not a duration" claim holds.
- **Over-fitting to `[1 2 4 4 4 4 4]` (attack 3): no.** M3 changes the doubling factor — the sequence becomes
  `1,3,4,4,…` — and the new test stays green, so it pins the bound and not the observed sequence. The
  production constants cannot reach it at all (it injects `backoffInitial`, `backoffMax`, `maxSpawnFailures`),
  and raising `clampProbeInitial` to the ceiling would fail it through its own *climbed* control, i.e. for the
  right reason.
- **Mutually-masking arms: no.** R4 holds arm 2 (`>=3 at the ceiling`) out of the test and DQ-1 still dies on
  arm 1 alone (4 arm-1 messages, 0 arm-2). Arm 2 has its own unique mutant (R1/DQ-3 fails arm 2 only). Both
  arms are load-bearing. Note for the record that on DQ-1 both arms fire (4x line 197 + 1x line 213), so the
  table's "1 `--- FAIL`" hides that DQ-1 is a two-arm kill — informational, not a defect.
- **Vacuity / positive control (attack 7): present and adequate.** Three asserted controls (construction-failure
  line, exactly 7 waits, at least one wait strictly below the ceiling), and `atCeiling >= 3` is itself a
  demonstration that the ceiling was reached. The DQ-3 row proves the controls are not carrying the pin.

### Findings — what I want changed

**1. "Unbounded" / "grows without limit" is false, and it is now shipping in a doc comment.** I measured the
crash path (the path the "between the first ceiling hit and give-up" window actually describes) at 1:4 with
production's `maxCeilingHits = 3`: clamped `[1 2 4 4]` sum 11ms, clamp deleted `[1 2 4 8]` sum 15ms — the
**same number of waits**, both reaching the give-up verdict. `maxCeilingHits` bounds the tail: deleting the
clamp buys `maxCeilingHits - 1` extra doublings and nothing more. At production tuning (derived arithmetic, not
executed — it cross-checks against #7123's published 91.5s): `0.5+1+2+4+8+16+30+30 = 91.5s` clamped versus
`0.5+1+2+4+8+16+32+64 = 127.5s` unclamped — **1.39x, bounded**. The real defect is a *bound violation* (waits of
32s and 64s against a ceiling the verdict counts at 30s), which is plenty to justify the pin; "unbounded" is
not. The word appears in the **`defaultEngineBackoffMax` doc comment** (shipped prose asserting something
falsifiable by arithmetic over two constants in the same `const` block), in the test header ("could double
without bound"), in **the arm-1 failure message** a future maintainer will read as a diagnosis, and in the PR
body. Please restate it as the bounded factor plus the bound violation.

**2. Premise (b) of the no-second-constant argument is false on the very path the new test runs.** The comment
says a clamp *below* the threshold "would stop the backoff ever reaching the threshold, so the verdict would
never be reached at all." On the **construction-failure** path the verdict is `spawnFailures >= s.maxSpawnFailures`
and `backoffMax` is never consulted: `run` calls `s.waitBackoff` directly there, and `ceilingHits` is touched
only inside `backoffAndMaybeGiveUp`, which only the crash branch calls (code read, plus the fixture itself —
all 7 of my spawn-path legs reach `engine child unspawnable` with `ceilingHits` never incremented). So a
sub-threshold clamp still reaches a give-up verdict on that path. Premise (a) is the same over-claim as
finding 1. **The conclusion survives** — clamp `>=` threshold is required for the *crash* verdict to be
reachable, and clamp `>` threshold breaks the bound, so `clamp == threshold` is the only value serving both —
but that is a different (and better) argument than the one written. I am **not** asking for a second constant.

**3. "The two roles' killers are disjoint, proven in both directions" is true only of the three mutants
scored.** Two further clamp mutants cross into #7123's arm: M5 (comparison inverted) fails **both** files, and
**M4 (comparison removed, clamp always fires) fails only `…LeavesRoomToGrow` while the new clamp test PASSES**.
So the clamp's own over-firing direction — the comparison being dropped, which is the direction a refactor
takes accidentally — is graded by the *threshold* file, not the clamp file, and the roles table's "graded, both
directions" over-reads what the new test does. Worse, that kill is a deadline kill: `…LeavesRoomToGrow`
exhausts its own 30s outer bound and reports *"only 2 spawn attempt(s) observed … the supervisor stopped
re-attempting"*, which misdiagnoses a supervisor that is sleeping 30s — and it costs 30s of suite time. A
~3-line arm in the new file would own it properly: **at least two waits strictly below the ceiling** (M4's
sequence is `[1 4 4 4 4 4 4]` — exactly one). If you consider that to be grading growth rather than the clamp,
say so explicitly instead of leaving "both directions" to cover it.

**4. Separate hole, not this PR's job: the doubling factor is ungraded.** M3 (`*backoff *= 2` → `*= 3`) leaves
the entire `./internal/daemon/` package green (0 `--- FAIL`). #7123's shrink arm asserts the second wait is
`>= 600ms` after a 400ms first wait, which `*= 3` satisfies at 1.2s. That is a *third* ungraded role on the
same two lines this PR documents. Worth its own issue in the #7126 family; I would not widen this PR for it.

**5. Minor, fixture-label accuracy.** The test header and the clamp-site comment both say the clamp bounds the
waits "between the first ceiling hit and the give-up verdict", but the fixture runs the construction-failure
path where **no ceiling hit ever occurs** (`ceilingHits` stays 0 and the verdict is the spawn budget's). Grading
shared code from that path is legitimate and the PR argues it well; the wording claims a window the fixture
does not enter.

### Not verified

- Linux/Windows behaviour (macOS only). Nothing in the fixture looks platform-sensitive, but I did not run it.
- The 91.5s/127.5s production figures are **derived from the constants, not executed** — deliberately, since the
  test injects a millisecond ceiling; my crash-path shape measurement (R2/R3) is the executed part.
- I did not re-derive DQ-2 (`backoffMax*4`) or DQ-4 (`defaultEngineBackoffMax = 3ms`); DQ-2's failing arms are
  predictable from R1/M2 and DQ-4's disjointness is structural (the fixture injects `backoffMax`).

The pin itself is good: the hole is real, the observation is the right artefact, it is genuinely timing-free
under `-race`, both arms are load-bearing, and the controls are not vacuous. What I am blocking on is claim
accuracy in text that is now shipping next to the constant.

**REQUEST-CHANGES**

1. Remove "unbounded"/"grows without limit"/"double without bound" from the `defaultEngineBackoffMax` doc
   comment, the test header, the arm-1 failure message and the PR body; replace with the bound violation plus
   the bounded factor (`maxCeilingHits - 1` extra doublings; 91.5s → 127.5s derived at production tuning).
2. Correct premise (b) of the no-second-constant paragraph: a sub-threshold clamp does **not** block the
   construction-failure verdict. Keep the conclusion, state it as "clamp == threshold is the only value that
   both bounds the waits at the counted ceiling and keeps the crash verdict reachable", and qualify (a) as a
   bound violation rather than unboundedness.
3. Either add the ~3-line "at least two waits strictly below the ceiling" arm (kills M4, which the new test
   currently survives), or narrow the roles table from "graded, both directions" and say in the body that the
   comparison's over-fire direction is owned by `…LeavesRoomToGrow` — noting that its kill there is a 30s
   deadline with a misdiagnosing message.
4. Soften the disjointness claim to the mutants actually scored, and record M4/M5 as the counter-examples.
5. Fix the "between the first ceiling hit and the give-up verdict" wording in the two comments to describe the
   path the fixture really walks.

(Items 1, 2, 4, 5 are text; item 3 is ~3 lines of test or one sentence of scope. No production behaviour change
is being asked for.)

cajasmota:
## Coordinator mutant ER-1 — the clamp's BOUNDARY, run by nobody else: **ALIVE + EQUIVALENT**, and the algebra is exhaustive

Every mutant scored so far perturbs the clamp's **existence** (DQ-1 deleted, M4 always-fires, M5 inverted) or its **value** (DQ-2 `*4`, DQ-3 `backoffInitial`). None touches its **boundary**, which is the off-by-one a refactor actually introduces — and there is a specific reason to look here: the two comparisons against `backoffMax` sit three lines apart with deliberately opposite strictness.

```go
if *backoff >= s.backoffMax {   // :584  give-up threshold
...
if *backoff >  s.backoffMax {   // :640  the clamp
```

Two tests of the same constant, one `>=` and one `>`, with nothing in the file saying why. That is exactly the shape a "make these consistent" cleanup changes silently.

**ER-1:** `>` → `>=` at the clamp. Full `./internal/daemon/`, `-count=1`, match count 1 asserted before the run, `go vet` exit captured to its own file (**0**), restored by `cp` with the shasum equal and `git diff HEAD` empty.

| | `--- FAIL` | |
|---|---|---|
| baseline on `5e6836b3b` | **0** | exit 0, 52.9s |
| **ER-1** | **0** | exit 0 — **ALIVE** |

### Why this is EQUIVALENT rather than a hole

The two forms differ on exactly one state: `*backoff == s.backoffMax`. In that state

- **HEAD** (`>`): condition false, assignment skipped, `*backoff` remains `s.backoffMax`.
- **ER-1** (`>=`): condition true, assignment performed, `*backoff` becomes `s.backoffMax`.

Identical result, because **the assignment is idempotent when the values are already equal.** And the branch body is only that assignment — verified by reading the whole block, which contains `*backoff = s.backoffMax` and nothing else, no counter, no logging, no early return. So no input distinguishes the two.

The distinguishing set here is a **single state**, analysed exhaustively, which is why this one needs no fuzz: the usual requirement for an equivalence — algebra PLUS an enumeration over the inputs that could distinguish — is satisfied by a one-element enumeration. That is the opposite of the situation on #7154, where an "equivalence" over a 3,528-state space turned out to be **masked** and needed 786 differing states found before anyone noticed.

### What this settles about the asymmetry, which is the useful part

The `>=` / `>` split is **not arbitrary, and not load-bearing either** — and the reason differs per site:

- **At the threshold (`:584`), strictness is behaviourally significant.** It decides whether `ceilingHits` increments when the backoff is exactly at the ceiling, which feeds the give-up verdict.
- **At the clamp (`:640`), strictness is behaviourally free**, by the idempotence above.

So a future "consistency" edit in either direction is harmless *at the clamp* and would be a real behaviour change *at the threshold*. Nothing currently records that, which is the one thing worth adding — one line at the clamp saying either comparison is correct here because the assignment is idempotent, so that

1. nobody "fixes" the inconsistency believing it matters, and
2. nobody writes a test pinning `>` at the clamp, which would be **vacuous by construction** and would read as coverage.

Requested as a comment-only addition. This does not block anything: the disposition is ALIVE + equivalent, recorded, not a gap.

Issue #7158, opened 2026-09-17
MEASURED (both twins ALIVE): the mandatory \s+ before the name is ungraded in moduleRE AND typeRE — and #7150's two \b equivalence claims rest on it, so relaxing it turns three holes live at once


cajasmota:
## What

The mandatory whitespace between the modifier prefix and the declared name is **ungraded in both F# declaration scanners**. Relaxing `\s+` to `\s*` leaves `./internal/extractors/fsharp/` completely green, and the relaxation is behaviourally live — it makes the scanners match identifiers that merely *start with* the keyword.

Measured by me on `0a7df19c1` (#7150's head), full package, `-count=1`, `go vet` clean, exact match count 1 asserted before each run, restored by `cp` with the shasum equal and `git diff HEAD` empty:

| mutant | edit | verdict |
|---|---|---|
| **EO-1** | `moduleRE`: `internal)\b)*\s+([\w.]+)\s*$` → `\s*` | **ALIVE, 0 `--- FAIL`** |
| **EO-2** | `typeRE`: `\s+([A-Z][a-zA-Z0-9_']*)` → `\s*` | **ALIVE, 0 `--- FAIL`** |

Baseline on the same tree: 0 `--- FAIL`, 0.52s.

**Both are distinguishable on ordinary input, so neither is equivalent** — they are ungraded holes. Divergence table (HEAD capture vs mutant capture):

```
'module Foo'            HEAD='Foo'     MUT='Foo'
'moduleFoo'             HEAD=None      MUT='Foo'
'moduleLoader'          HEAD=None      MUT='Loader'
'modulePrivateThing'    HEAD=None      MUT='PrivateThing'
'moduleX.Y'             HEAD=None      MUT='X.Y'

'type Foo ='            HEAD='Foo'     MUT='Foo'
'typeFoo ='             HEAD=None      MUT='Foo'
'typeAlias ='           HEAD=None      MUT='Alias'
'typeDef<T> ='          HEAD=None      MUT='Def'
'typeState() ='         HEAD=None      MUT='State'
```

So under the relaxation a bare line `moduleLoader` mints a module named `Loader`, and `typeState() =` mints a type named `State`. This is the **over-firing** direction, which recall-style must-have rows are structurally blind to: everything a fixture expects is still found and the phantoms are simply not counted.

## Both halves of the twin, deliberately

[...]

Issue #7152, opened 2026-09-17
fsharp: EDGE scanners read the comment/string-scrubbed copy but all four DECLARATION scanners read raw src — commented-out declarations mint phantoms, and name-keyed dedup hides them from count AND presence checks


cajasmota:
## What

The F# extractor maintains a **comment- and string-scrubbed** copy of the source — a blanking pass that replaces every byte of a suppressed region with a space while preserving `'\n'`, so line numbers survive (`extractor.go:1176-1200`, #6336). It handles nesting: F# block comments nest, `(*)` is excluded so `List.fold (*) 1 xs` is not read as a comment opener, and triple-quoted strings are covered.

**The declaration scanners do not use it.** All four run against the raw `src`:

```
extractor.go:393   moduleRE.FindAllStringSubmatchIndex(src, -1)
extractor.go:447   letRE.FindAllStringSubmatchIndex(src, -1)
extractor.go:491   memberRE.FindAllStringSubmatchIndex(src, -1)
extractor.go:552   typeRE.FindAllStringSubmatchIndex(src, -1)
extractor.go:574   memberRE.FindAllStringSubmatchIndex(src, -1)
```

The scrubbed copy's own doc comment names its consumers as `collectCalls`, `collectCEUsages` and `firstSignalLine` — the *edge* scanners. So within one extractor, **edges are scanned against scrubbed source and declarations against raw source.**

The consequence is phantom entities: a commented-out or string-embedded declaration still matches. `module Foo` on its own line inside a `(* ... *)` block satisfies `(?m)^([ \t]*)module(?:\s+rec)?\s+([\w.]+)\s*$` because nothing in that pattern knows it is inside a comment.

## What I verified myself, and what I did not

Verified by reading `e96eb099f`: the five call sites above all take `src`; the scrub exists, nests correctly, and its documented consumers are the edge scanners. That establishes the **inconsistency** and the **mechanism**.

Reported by the #7135 implementation agent and **not independently executed by me**: that `(* *)` blocks and `"""..."""` strings actually do mint phantom modules and types end-to-end, and that **line** comments are correctly ignored while that direction *is* graded. Those are their measurements, not mine — the phantom claim follows from the mechanism I checked, but I have not run the extractor over such a fixture, and the line-comment asymmetry in particular deserves confirming rather than assuming (`:763` does handle `//` for one substring in one context, which is not the same thing as the scanners being comment-aware).

[...]

Issue #7153, opened 2026-09-17
fsharp: typeKindRE is unreferenced repo-wide yet its comment claims it 'helps classify subtype' — and #7150 makes it typeRE's deliberately-divergent dead twin


cajasmota:
## What

`typeKindRE` is unreferenced. `grep -rn "typeKindRE"` across the whole repo (`--include='*.go'`, tests included) returns **exactly one line: its own declaration.**

```go
// internal/extractors/fsharp/extractor.go:204
// type kind after "=" — helps classify subtype
typeKindRE = regexp.MustCompile(
    `(?m)^([ \t]*)type\s+[A-Z][a-zA-Z0-9_']*\s*(?:<[^>]*>)?\s*(?:\([^)]*\))?\s*=\s*(\{|interface|class|\|)`,
)
```

Verified against `e96eb099f`.

Two things make it worth a ticket rather than a silent deletion:

**1. The doc comment states a production benefit for code production does not run.** *"helps classify subtype"* reads as a description of live behaviour. A reader auditing F# subtype classification finds this pattern, sees a comment saying it does the job, and concludes the path is covered — the same shape as #7146 item 3 (`buildV2GraphWithNodeCap`, test-only with a production rationale in its doc comment). That is how a gap survives an audit.

**2. It is `typeRE`'s twin, and #7150 has just moved one of them.** The two patterns share their entire prefix — `type` plus the same name/typar/parens sequence — differing only in what follows the `=`. #7150 (the #7135 `moduleRE`/`typeRE` arm) widens `typeRE` to accept access modifiers (`type private Foo =`) and deliberately leaves `typeKindRE` un-widened, on the correct ground that editing unreachable code is an edit nothing can grade. So after #7150 lands the twins **diverge by design**, and the divergence is documented at the pattern. That is the right call for that PR and it is also a standing hazard: a twinned surface where one half is live and the other is dead invites someone to "restore parity" by widening the dead half, or to copy the dead half as the reference.

## What is NOT claimed

- **Not claimed that F# subtype classification is broken.** Whatever classifies F# type subtypes today is doing it without this pattern; I have not established what that is, or whether the classification is right. That is the question worth answering before choosing between deletion and revival — if the classification is missing or wrong, this pattern is a clue to intent rather than dead weight.
- No corpus measurement, and no F# execution (no `dotnet`/`fsc`/`fsharpc`/`fsi`/`mono` on this machine).
[...]

PR #7164, opened 2026-09-17
fix(#7135): arm 3 — `let private (|Even|Odd|)` was extracted as NOTHING, and a missed definition takes its whole case set with it

### LINKED ISSUE #7135: ALL THREE ARMS SHIPPED: memberRE (c1e2c3008), moduleRE+typeRE (0ed6767d7), active patterns (3bb6d28b8) — and arm 3's failure mode was NOT a silent total miss: letRE mis-names the head and mints a phantom (#7163)

### LINKED ISSUE #7163: fsharp: letRE mis-names a modifier-carrying ACTIVE-PATTERN head — `let private (|Even|Odd|)` also mints a phantom operation called `private`
## What

`letRE` in `internal/extractors/fsharp/extractor.go` matches an **active-pattern** `let` head that carries a modifier, and mis-names it: the modifier becomes the entity's `Name`, minting a phantom `SCOPE.Operation`/`let` alongside the correct `SCOPE.Pattern`.

Probed against the extractor at `d3a144f7f` (i.e. **after** all three arms of #7135 landed; the phantom also reproduces at `45f9745d0`, before the third arm, so it is neither caused nor cured by that fix):

| source | entities minted |
|---|---|
| `let (\|Even\|Odd\|) n =` | `SCOPE.Pattern (\|Even\|Odd\|)` — correct, no phantom |
| `let rec (\|Even\|Odd\|) n =` | `SCOPE.Pattern (\|Even\|Odd\|)` **+ `SCOPE.Operation/let` named `rec`** |
| `let private (\|Even\|Odd\|) n =` | `SCOPE.Pattern (\|Even\|Odd\|)` **+ `SCOPE.Operation/let` named `private`** |
| `let inline private (\|Even\|Odd\|) n =` | `SCOPE.Pattern (\|Even\|Odd\|)` **+ `SCOPE.Operation/let` named `private`** |

Note the last row: the phantom is named `private`, not `inline` — the group matched ` inline`, then backtracked **one** repetition so the name capture could take `private`.

## Mechanism

```
letRE = (?m)^([ \t]*)let(?:\s+(?:rec|mutable|inline|private|internal|public)\b)*
        \s+([a-zA-Z_][a-zA-Z0-9_']*)\s*(?:<[^>]*>)?\s*(?:[^=\n]*)=
```

The modifier group is *optional/repeated*. On `let rec (|Even|Odd|) n =` the engine first takes ` rec`, then finds the name capture `[a-zA-Z_]…` facing `(` and cannot proceed, so it **backtracks the group to zero repetitions** — and then `\s+` takes the space and the NAME capture takes `rec`, with `(|Even|Odd|) n ` swallowed by the trailing `[^=\n]*`.

With **no** modifier there is no phantom, because the mandatory `\s+` then leaves the name capture facing `(` with nothing left to backtrack into. So the defect is **modifier-dependent**, which makes it an adjacent arm of the #7135 / #7131 family rather than an unrelated bug.

## Why it matters

Same consequence as #7131 and #7135 arm 1: wrong `Name` and `QualifiedName`, so the phantom is unfindable by any real name and cannot be a call target — and `letSeen` is keyed on `indent + ":let:" + name`, so **two same-indent active patterns with different modifiers collide on one key**: `let private (|A|B|)` and `let internal (|C|D|)` at the same indentation both key on their modifier, and two `let private` siblings both key on `"private"` with the second dropped. That is the collapse mode #7136 re-keyed for.

It is also the mode that a count assertion on the *active pattern* cannot see: the SCOPE.Pattern census is correct in every row above. Only a check on the operation census catches it.

## Not claimed

- **Not measured for incidence on a real corpus.** `let private (|Foo|_|)` is idiomatic, which is an argument, not a measurement.
- **Not verified against an F# compiler** — there is none on this machine (`dotnet`, `fsc`, `fsharpc`, `fsi`, `mono` all absent; `javac` is the only compiler present), so the legality of each probe source rests on the F# spec §10.5 / MS Learn "Access Control" and "Active Patterns", not on a compile.
- **No fix shape asserted.** RE2 has no negative lookahead, so "a name that is not a banana clip" cannot be written directly in `letRE`; the fix is probably a post-match reject in the let scanner (skip a match whose tail begins `(|`) or an explicit non-`(` guard. Whichever it is, the accepted-name shape needs its own enumeration — `letRE` is the *widest* of the four F# declaration scanners.

## Grading requirement

- Assert the **operation census**, not just presence of the active pattern: the phantom is an extra entity, and the existing SCOPE.Pattern assertions pass right through it.
- Enumerate the modifier space rather than hand-picking — a single `let private (|A|B|)` row leaves `let inline private` open, and that row is the one that proves the backtrack is *partial* rather than all-or-nothing.
- Include the **collision** direction: two same-indent modifier-carrying active patterns in one file, asserting the count, since `letSeen` will otherwise silently merge them.
- Score a mutant that reverts the guard and one that makes it too strict (rejecting a legitimate `let (x, y) = …` tuple head or a parenthesised operator definition such as `let (+.) a b = …`, if those are in `letRE`'s reach) — a too-strict guard is a silent-miss regression in the opposite direction.

Found while enumerating the modifier space for #7135's third arm (active patterns); out of that PR's scope — it is a different scanner, and its own arm.

cajasmota:
Third and last arm of #7135. `activePatternRE` accommodated a **fixed sequence of exactly two optional modifiers in exactly one order** (`(?:\s+rec)?(?:\s+inline)?`), so any access modifier made the whole active-pattern definition vanish.

```
- let(?:\s+rec)?(?:\s+inline)?\s+\(\|
+ let(?:\s+(?:rec|mutable|inline|private|internal|public)\b)*\s+\(\|
```

**The failure mode, corrected.** An earlier draft of this body and both file headers said "no entity, no edge, no diagnostic". That is **measurably false** and contradicted this PR's own #7163 section: pre-fix, `let private (|Even|Odd|) n =` minted a `SCOPE.Operation`/`let` named **`private`** on that very line. What was lost is *no active-pattern entity, no case sub-entity, no CONTAINS edge, no match-site USES edge and no diagnostic — only `letRE`'s phantom (#7163)*. The active-pattern census was 0, which is what every table asserts; the **line** was mis-attributed, not silent. The phrasing was inherited from #7135's title, where it is true of the module and type arms.

**Strictly worse here than at the `module`/`type` sites.** The definition is the *sole* producer of the case sub-entities, so a missed one takes its whole case set with it and `collectActivePatternCases` returns an empty map — every match arm `| Even ->` in the file resolves to nothing. Graded directly by `TestActivePatternModifiers_MatchSiteResolvesUnderAModifier`.

### The repeating group is RECALL first, lenience second — also corrected

`function-defn := inline? access? ident-or-op …`, with `rec` supplied by the `let rec` group ahead of it, makes **`inline private` and `rec private` the spec-legal order**. Measured: both match the new pattern and match *neither* slot order of the old one. So `)*` recovers **legal two-modifier F# that was missed outright**; the earlier draft called those 30 subtests "NOT legal F#", inheriting arm 2's `modifierPhrases` helper, which labels both orders of an `extra`+`access` pair as reversed. `fsAPLabel` now re-derives each label from the grammar's modifier **rank** (`rec` 0 → `inline` 1 → `access` 2; strictly increasing = legal, equal = repetition, decreasing = reversed). Correcting the shared arm-2 helper would relabel arm 2's tables too and wants its own commit; this arm just stops repeating the claim.

**What is genuinely lenience,** and it is load-bearing: the group still accepts an arbitrary multiset in any order (`private public`, `private rec`, `rec rec`). **16 subtests depend on it** — the three `a a` phrases × 5 clip shapes, plus `GroupIsRepeating/private private`, the file's only pure-multiset row.

### Blocker item 1: the central justification was prose-only, and the class was open

"Byte-for-byte parity with `letRE`" is asserted in ~11 places and was **observed by no test**. The six forbidden words were hand-picked, so every word not picked was ungraded — an unbounded class. The reviewer's demonstration: **adding `sealed` to the allowlist was ALIVE at 0 `--- FAIL`** with `go vet` exiting 0. I reproduced that on `b60b30614` before changing anything.

New in-package (`package fsharp`) test `TestActivePatternREModifierParityWithLetRE` asserts the two modifier prefixes are byte-identical, extracted from each compiled `Regexp.String()` — the artefact the engine uses, not the source bytes. It closes the whole class at once instead of one word at a time.

| mutant | before | after |
|---|---|---|
| `sealed` added to `activePatternRE` only | **ALIVE, 0** | **DEAD, 2** |
| `sealed` added to **both** patterns | — | ALIVE, 0 — **correct by design**: parity is the claim, not a frozen allowlist. A coordinated widening is a review conversation, not a test failure. |
| meta-regex broken so both extractions fail (vacuity attack) | — | **DEAD, 1** on the `t.Fatalf` guard |

The vacuity guard is the load-bearing half: a parity assertion over two *failed* extractions passes no matter what either pattern says. The test also asserts the extracted prefix still contains `)*`, `\s+`, `(?:` and at least one `|`. It deliberately does **not** pin the allowlist's contents — that would freeze the set and move a legitimate future widening from review into a test failure. The set's upper boundary stays graded by the forbidden rows.

### A wrong label is invisible on a green run — so the labeller is now graded too

Every `fsGrammar` label is consumed **only inside a `t.Errorf` format string**, which is exactly how 30 mislabelled subtests survived a full suite *and* a 17-mutant round. `TestActivePatternModifiers_LabelDerivation` pins 31 phrase→label rows against the grammar plus a mutual-distinctness check. **It immediately caught a real bug in my own `fsAPLabel`**: a `len(w) < 2` shortcut labelled a single non-allowlisted word (`sealed`) `specLegal`. Fixed by ranking every word before any shortcut; the mutant that reinstates the shortcut is **DEAD at 2**.

Reported honestly: the mutant reverting `NeverMissed` to the inherited `mp.grammar` is **ALIVE at 0**. A label read only by a format string cannot be observed on a green run without asserting on failure output. `fsAPLabel`'s *derivation* is graded; which label a given row *consumes* is not. That is a real limit of this correction, not a closed hole.

### Mutant table — the original 17 all reproduce exactly (reviewer re-scored independently)

| mutant | `--- FAIL` |
|---|---|
| drop `rec` / `mutable` / `inline` / `private` / `internal` / `public` | 48 / 36 / 43 / 65 / 50 / 47 — all DEAD |
| add `protected` / `static` / `val` / `abstract` / `override` | 2 each — all DEAD, each on its own row with its own reason string |
| allowlist → `[a-z]+` | 7 — DEAD |
| group non-repeating (`)*`→`)?`) / removed entirely | 123 / 172 — DEAD |
| indent `[ \t]*` → ` *` | 7 — DEAD |
| `\b` removed | 0 — **ALIVE + equivalent** |
| `\s+` → `\s*` | 0 — **ALIVE + ungraded, pre-existing (#7158)** |

RED re-derived at **161** (`NeverMissed 126 / Indentation 16 / UpperBoundary 7 / GroupIsRepeating 7 / MatchSite 5`; 126+16+7+7+5 = 161), all six UpperBoundary rows failing on the *positive control* assertion. Every mutant: `cp` backup, asserted exact pre-match count, `go vet`'s exit code captured into its own file (never through a pipe), restore by `cp`, `shasum` + `git diff HEAD` verified. `--- FAIL` lines counted, never exit codes.

**`\b` is equivalent here UNCONDITIONALLY, and #7158 does *not* apply to it at this site** — a correction to the brief I was given, which told me to reuse the siblings' *conditional* finding. `\b` can only bite when an allowlist word is followed by a *word* character, and the next obligatory token here is the literal `(` of the banana clip. 11,924,640 enumerated lines under each separator × 4 cells, 0 divergences, plus every single byte after each of the six words. The reviewer independently confirmed it with a proof from the continuation alphabet, 2,116,800 lines × 4 cells, 12,288 byte-suffix inputs, and the contrast: on `letRE`'s shape under `\s*`, `let privateFoo = 1` names `privateFoo` with `\b` and **`Foo`** without — because `letRE` continues with an identifier class and this pattern with a literal `(`.

### `mutable`: parity outweighs, with the reviewer's firmer citation

Illegal on this sub-form on three independent grounds, replacing my weaker "binds a pattern, not a storage location": the grammar admits `mutable` only in `value-defn := mutable? access? pat`, while a parameterised active-pattern head is a `function-defn := inline? access? ident-or-op …` with **no `mutable` slot at all**; `FSComp.txt` **874** ("Only record fields and simple, non-recursive 'let' bindings may be marked mutable"); and **831** ("Mutable function values should be written 'let mutable f = (fun args ...)'"). No compiling source produces the form, so excluding the word removes no real-world case and admitting it mis-extracts none. Kept for the `letRE` parity that is now asserted, and labelled `lenienceSubform`.

### This widening enlarges an open sibling defect's surface

`extractActivePatterns` is called with the **raw `src`** (`extractor.go:522`), not the comment/string-scrubbed copy the edge scanners use, so **#7152** applies here: a commented-out `let private (|A|B|)` inside a `(* … *)` block or a triple-quoted literal now mints a phantom where the two-slot form skipped it. The surface grows by exactly the declarations this arm newly recognises. #7152 is one shared call-site convention across all four F# declaration scanners and is not fixed here — but this arm is one of its causes.

### Found and filed, not fixed: #7163

The modifier-name guard, while still unscoped, caught a **different scanner's** defect. `letRE` also matches a modifier-carrying active-pattern head and mis-names it, because its optional group backtracks to zero repetitions when the name capture faces `(`: `let rec (|Even|Odd|) n =` → phantom named `rec`; `let inline private …` → `private`, so the backtrack is *partial*, one repetition. The reviewer confirmed the phantom set is **byte-identical before and after** this fix. So the guard is scoped to `SCOPE.Pattern`/`SCOPE.Schema` and **no row pins the phantom**, so a future fix to `letRE` will not have to delete a test that blessed it.

### Axes

**Varied:** modifier identity (6 allowlisted + `protected`/`static`/`val`/`abstract`/`override`/`qqzz`/`sealed`); count (0, 1, 2); order (both directions of every `extra`+`access` pair, `rec inline` vs `inline rec`); repeated access modifier; clip shape (2-case, 3-case, two partials); parameterisation (0, 1, 2 args); indentation (col 0, 4 spaces, tab); and the downstream consequence (case sub-entities, CONTAINS, match-site USES).

**Held constant:** case names, entity Kinds, one declaration of interest per file, file path.

**Deliberately not varied:** no three-modifier row; no `rec rec`/`inline inline` (only *access* words repeat); and **no two same-name active patterns in one file** — `extractActivePatterns`' `seen` map is keyed on the clip name alone with no enclosing scope, so a same-name sibling is dropped (open #7144, reproduces with no modifier, neither fixed nor asserted as correct).

Labels are now per row in **all four** enumerated tables; the first draft labelled only `NeverMissed` while `Indentation`, `MatchSite` and `GroupIsRepeating` carried unlabelled lenience rows. `AllowlistUpperBoundary` carries a per-row `reason` string instead, since every one of its rows is by construction not legal on a `let`.

### Verification

- `./internal/extractors/fsharp/` — **0 `--- FAIL`**; `go vet` exit 0; `gofmt -l` empty.
- `./cmd/grafel/` — **0 `--- FAIL`** (`ok … 221.657s`) with a fresh `GRAFEL_HOME` **and** `GRAFEL_DAEMON_ROOT`, digest unmoved; independently re-run by the reviewer at 212.158s. The pattern is **byte-identical** to the version that run tested — this round adds an in-package assertion about regex *text* plus comments and labels, and changes no extraction. Fixture has **18** members (re-derived with a backtick-parity tracker after a naive scan stopped at a `func ` line inside `svc/handler.go`'s raw string and reported 17), **zero `.fs`/`.fsx`**, so that gate is regression cover, not coverage.
- **No F# toolchain here** (`dotnet`, `fsc`, `fsharpc`, `fsi`, `mono` all absent; `javac` only) — every legality claim is *derived-not-executed* and cites the spec grammar, MS Learn, or `FSComp.txt`.

Refs #7152 — this widening enlarges its raw-`src` phantom surface.
Refs #7158 — the ungraded `\s+` separator. It does **not** cover the `\b` at this site; that equivalence is unconditional.
Refs #7163 — `letRE`'s phantom on a modifier-carrying active-pattern head, found here.
Refs #7144 — the enclosing-scope-free dedup key, why no same-name row exists.

Closes #7135 — arms 1 (`c1e2c3008`) and 2 (`0ed6767d7`) are on `main`; this is the last one.


cajasmota:
## Coordinator mutant ES-1 — the CLIP BODY, which nobody scored: **ALIVE + EQUIVALENT**. The `(?:\|_)?` partial-pattern suffix is dead weight.

Every mutant in the PR targets the **modifier prefix** — the six words dropped individually, five widening words added, `[a-z]+`, non-repeating, group-removed, `\b`, and the separator. Nothing touches the rest of the pattern. So I went after the half nobody probed:

```go
`\s+\(\|([A-Za-z0-9_'|]+(?:\|_)?)\|\)\s*([^=\n]*)=`
                             ^^^^^^^^^ ES-1 removes this
```

`(?:\|_)?` reads as the construct that makes **partial** active patterns (`(|Foo|_|)`) work, which is exactly why it is worth checking: a reader takes it as load-bearing.

**ES-1:** drop `(?:\|_)?`. Full `./internal/extractors/fsharp/`, `-count=1`, match count asserted, `go vet` exit captured to its own file (**0**), restored by `cp` with the shasum equal and `git diff HEAD` empty.

| | `--- FAIL` | |
|---|---|---|
| baseline on `b60b30614` | **0** | exit 0, 0.705s |
| **ES-1** | **0** | exit 0 — **ALIVE** |

### It is EQUIVALENT, and the group is strictly redundant

**Algebra.** The optional group's alphabet is `{|, _}`, and **both characters are already in the preceding class** `[A-Za-z0-9_'|]`. The class is `+` and greedy, so anything the optional group could match, the class has already consumed. Backtracking cannot create a difference either: the capture group encloses *both* the class and the optional suffix, so whether `|_` is consumed by the class or handed to the suffix, **the captured span is identical**.

**Enumeration**, over the alphabet that could distinguish them — `A | _ ' ) ( space tab =` — all clip bodies of length 0..6 inside `let (|…|) x =`: **597,871 inputs, 0 differing captures.**

So `(?:\|_)?` can be deleted with no behavioural change. Partial patterns work because `|` and `_` are in the character class, **not** because of that suffix.

### Why this is worth recording rather than just noting

This is the third disposition kind, and the least visible: not a hole, not a kill — a construct that **looks like it implements a feature and does not**. The risk is not a regression, it is a reader (or a future fix for #7163) concluding that partial-pattern support lives in that suffix and editing it expecting an effect. That is the same shape as the comment corrections in `53ede5716` and the `>`/`>=` note in `45f9745d0`: prose or syntax asserting something no behaviour depends on.

**Requested as a comment-only addition** — one line saying the suffix is redundant because the class subsumes it, that ES-1 is ALIVE and equivalent over 597,871 enumerated inputs, and that partial-pattern support comes from `|`/`_` being in the class. Deleting the suffix would also be defensible, but it is a live extractor pattern and a comment costs nothing; if you prefer deletion, the enumeration above is the evidence it is safe.

**Not a blocker.** Disposition is ALIVE + equivalent, recorded, not a gap.

### A method note against myself

My first attempt at this mutant was **broken and `go vet` passed it.** The replacement put a `// MUTANT-ES1` marker *inside the backtick regex literal*, so the compiled pattern contained the literal text `// MUTANT-ES1 partial-suffix dropped` — a completely different pattern from the one I meant to test, still valid Go, vet clean:

```
138:  `\s+\(\|([A-Za-z0-9_'|]+)\|\) // MUTANT-ES1 partial-suffix dropped\s*([^=\n]*)=`,
```

Caught by reading the mutated line back rather than trusting the applied-count. `go vet` proves a mutant **compiles**, never that it says what you meant — and a marker comment is safe in ordinary code and unsafe inside a raw string. The re-run asserts the resulting text directly (old count 0, new count 1) and carries no in-string marker. Same family as the raw-string trap that produced the 17-vs-18 fixture miscount twice tonight.

cajasmota:
## Independent adversarial review — arm 3 (`activePatternRE`)

Own worktree at `origin/wt-7135ap-a1` = `b60b30614`, detached, absolute-path asserted (`test "$PWD" = …-wt-7164-rev || exit 9`) inside every mutation script. Every mutant: byte-for-byte `cp` backup, an **asserted exact pre-match count** (four mutants **aborted as `PREMATCH cnt=2`** because the allowlist string also appears in a comment — re-applied against the regex-scoped group instead, never scored on the ambiguous match), `go vet`'s exit code captured into its own file, restore by `cp`, `shasum` + `git diff HEAD` verified after each. Final tree: `4cdd0c22…` / `afd5725d…`, `git status` clean, both equal to the recorded base. `--- FAIL` lines counted, never exit codes.

### The mutant table reproduces EXACTLY — 17/17, zero discrepancies

Re-derived from scratch, not carried over. Baseline 0 `--- FAIL` (0.53s).

| mutant | author | **mine** | verdict |
|---|---|---|---|
| drop `rec` | 48 | **48** | DEAD |
| drop `mutable` | 36 | **36** | DEAD |
| drop `inline` | 43 | **43** | DEAD |
| drop `private` | 65 | **65** | DEAD |
| drop `internal` | 50 | **50** | DEAD |
| drop `public` | 47 | **47** | DEAD |
| add `protected` / `static` / `val` / `abstract` / `override` | 2 each | **2 each** | DEAD |
| allowlist → `[a-z]+` | 7 | **7** | DEAD |
| non-repeating `)*`→`)?` | 123 | **123** | DEAD |
| modifier group removed | 172 | **172** | DEAD |
| indent `[ \t]*`→` *` | 7 | **7** | DEAD |
| `\b` removed | 0 | **0** | ALIVE + equivalent (verified below) |
| `\s+`→`\s*` | 0 | **0** | ALIVE + ungraded (#7158) |

I also re-derived **RED** by reverting only the regex to the pre-fix two-slot form: **161 `--- FAIL`**, decomposing to `NeverMissed 126 / Indentation 16 / UpperBoundary 7 / GroupIsRepeating 7 / MatchSite 5` — component-by-component identical to the body, and **all 6 UpperBoundary rows failed RED on the `positive control … minted 0, want 1` assertion** (6 occurrences of that string in the RED log). That claim is real, not inferred from the total.

### Attacks that FAILED to land

**1 — the five thin widening rows (the sibling arm's merge-blocker class).** Each of `protected`/`static`/`val`/`abstract`/`override` dies at exactly 2, and each dies **on its own row** with its own reason string (`AllowlistUpperBoundary/protected` … `/override`) — not on a neighbour. Then I broke the control (`const control = "private"` → `"qqzzcontrol"`): **7 `--- FAIL`, all six rows red on the control `Fatalf`**, so no forbidden row can pass on a silent harness failure. Name-collision hazard is N/A with a reason I verified rather than assumed: `runFSharp` constructs a fresh extractor per call, `extractActivePatterns`' `seen` map is function-local, and each row is two separate single-declaration sources (`ctl.fs`, `bad.fs`) with one clip name each — no shared fixture for a phantom to merge into.

One structural note, not a defect: the control is **one shared constant, not six** — each row runs it, but a single edit disarms all six at once. That is what my mutant measured.

**2 — the `\b` equivalence, and your correction of my instruction. Your correction is right; my instruction was wrong.** I told you to reuse the siblings' *conditional* finding; you measured it false here and re-derived it. I confirm it independently, both ways:

*Algebra.* Let the group be `(?:\s+W)*` with/without `\b`. `\b` ⊆ no-`\b`, so only ⊇ needs proof. In any successful no-`\b` derivation, if a repetition's `\b` would fail, the next byte is a word char; but every continuation is either the next repetition's `\s+` (needs whitespace) or the tail `\s+\(\|` / `\s*\(\|` (needs whitespace, or `(`). A word char is none of those. So every `\b` in a successful derivation already holds, and the two accept the same language **with identical submatches** — unconditionally, for any separator.

*Enumeration.* A 4-cell differential (`\b`/no-`\b` × `\s+`/`\s*`) over **2,116,800** constructed declaration lines — 21 words including `recFoo`, `privateThing`, `internalState`, `recrec`, `publicly`, `Rec`, `PRIVATE`, `rec_`, `public'`, `inline1`, 15 glue strings, 4 clip shapes, 5 argument forms, 4 indentations, 4 pre-clip variants — **0 divergences in all four cells**; plus 12,288 byte-suffix inputs (every byte after each of the six words, spaced and unspaced), **0 divergences**.

*The contrast is real.* I ran the same differential on `letRE`'s shape: `sep=\s+` → 0 divergences; `sep=\s*` → divergences, e.g. `let privateFoo = 1` gives name `privateFoo` with `\b` and **`Foo`** without. So the mechanism you name is exactly right: `letRE` continues with an identifier class and `activePatternRE` with the literal `(`. **Stated plainly: #7158 does NOT apply to the `\b` at this site, and the equivalence here is unconditional.** Leaving it untested on purpose is correct.

**4 — "matches `letRE` byte-for-byte".** True. Extracted both prefixes and compared: both are the identical 59-char string `let(?:\s+(?:rec|mutable|inline|private|internal|public)\b)*`, and the surrounding `^([ \t]*)` capture matches too.

**5 — shape-mutant over-fitting.** Not found. `grouprm`'s 172 decomposes cleanly (135+1 NeverMissed, 15+1 Indentation, 4+1 MatchSite, 7+1 GroupIsRepeating, 6+1 UpperBoundary-**on the control**, which is the correct behaviour for a total regression). No assertion encodes a literal expected entity list; the recall table's expectations are derived from `fsAPClipShapes`/`modifierPhrases`, so widening or narrowing an unrelated modifier cannot break an unrelated row. `fsAPNoModifierNamedEntity` is scoped to the two Kinds this scanner mints, so a future #7163 fix breaks nothing.

**7 — `./cmd/grafel/`.** Re-ran it myself, fresh `GRAFEL_HOME` **and** `GRAFEL_DAEMON_ROOT`: `ok … 212.158s`, **0 `--- FAIL`**, digest unmoved (`TestCustomExtractorGateOffGraphIsUnchanged` returns silently only on `gateOffDigest6916`; every other outcome is a `Fatalf` — it is a real pin). `entityTupleKey` hashes **8** fields — `Kind, Name, QualifiedName, Subtype, SourceFile, Language, StartLine, EndLine` — and **not** Properties. Fixture member count re-derived with a method that cannot be fooled by a `func ` line inside a raw string (path-literal extraction gated on backtick parity): **18 members, zero `.fs`/`.fsx`**. Your "regression cover, not coverage" framing is exactly right.

**8 — #7163.** Verified both halves empirically, by dumping every entity for six declarations with the fix and then with only the regex reverted to the pre-fix form. The phantom set is **byte-identical before and after**: `let rec (|Even|Odd|) n =` → `SCOPE.Operation/let` named `rec`; `let inline private …` → `private` (partial backtrack confirmed); and additionally `let private public …` → `public`, `let mutable …` → `mutable`. Neither caused nor cured here, and no row pins it. The table in the body is accurate.

**3 — `mutable`.** Your conclusion is right and I can cite it harder than the body does. `let mutable (|Even|Odd|) n = …` is illegal on three independent grounds (all **derived-not-executed** — no `dotnet`/`fsc`/`fsharpc`/`fsi`/`mono` here, `javac` only): the spec grammar puts `mutable` only in `value-defn := mutable? access? pat`, while a parameterised active-pattern head is a `function-defn := inline? access? ident-or-op …` with **no `mutable` slot at all**; `FSComp.txt` **874** — "Only record fields and simple, non-recursive 'let' bindings may be marked mutable"; and **831** — "Mutable function values should be written 'let mutable f = (fun args ...)'". So the word does admit a form the grammar forbids. **I agree the parity argument outweighs it**: `letRE` genuinely shares the `let` production (a real twinned surface, cf. #7153), the admitted form cannot occur in compiling F# so no real source is mis-extracted, and the row is labelled. Please swap the body's weaker justification ("binds a pattern, not a storage location") for the grammar/`FSComp` citation.

### Attacks that LANDED

**A. The PR's central justification is prose-only, and I have an ALIVE mutant to prove the class is open.** "Byte-for-byte parity with `letRE`" is asserted in ~11 places across the two files and **observed by no test**. The six enumerated forbidden words are graded; a **seventh is not**: adding `sealed` to the allowlist is **ALIVE at 0 `--- FAIL`**. That is the same defect class one layer up from the sibling arm's blocker — there the *exclusions* were prose, here the *principle that generates the exclusions* is prose. It is closable completely, not incrementally: `internal/extractors/fsharp/` already has in-package (`package fsharp`) test files, so ~8 lines asserting `activePatternRE.String()`'s modifier prefix is byte-identical to `letRE.String()`'s kills `sealed`, `protected` and every future fabricated word at once — and pins the claim the body rests on.

**B. Six recall rows are labelled "NOT legal F#" when they are the LEGAL order, and the body undersells the fix as a result.** `lenienceOnly`'s own doc comment says "access must **FOLLOW** inline/mutable", and the spec grammar agrees (`function-defn := inline? access? ident-or-op …`; `rec` comes from the `let rec` group ahead of it). But `modifierPhrases` labels **both** orders `lenienceOnly`, so `rec public|private|internal` and `inline public|private|internal` — 6 phrases × 5 shapes = **30 subtests** — are labelled as illegal F# while being the spec-legal order. Measured consequence: `let inline private (|Even|Odd|) n =` and `let rec private (|Even|Odd|) n =` match the new regex and **do not match the old one**. So the repeating group is required for **legal, real-world F#**, not merely tolerated as lenience — which is a stronger argument for `)*` than the body makes, and the body currently presents the repeat as graded only by lenience rows. (Derived-not-executed; the mislabel is inherited from the shipped arm-2 helper, so fixing it may want its own commit — but this arm should not repeat the claim.)

**C. "no entity, no edge, no diagnostic" is measurably false, and the PR contradicts itself.** Pre-fix, `let private (|Even|Odd|) n =` minted `SCOPE.Operation/let` named **`private`** on that very line (measured above). The active-pattern entity count was 0 — that part is right and is what the tests assert — but the line was not silent, and the body's own #7163 section says so. Reword to "no active-pattern entity, no case sub-entity, no edge, no diagnostic — only `letRE`'s phantom (#7163)".

**D. "labelled per row" holds in one table of five.** `NeverMissed` labels. `Indentation` (`rec private`, `private rec`), `MatchSite` (`inline private`) and `GroupIsRepeating` (all seven phrases, including **`private private` — the only pure multiset-lenience row in the PR**) carry lenience rows with no `fsGrammar` at all. And rows **do** depend on the multiset lenience: the three `a a` phrases (×5 shapes = 15 subtests) plus `GroupIsRepeating/private private` would fail against a grammar-faithful two-slot form. Label them, or narrow the claim.

**E. This arm enlarges the surface of an open sibling defect and does not say so.** `extractActivePatterns` is called with raw `src` (`extractor.go:522`), so **#7152** applies here too: a commented-out `let private (|A|B|)` at column 0 inside a `(* … *)` block now mints where the two-slot form would have skipped it. One sentence and a `Refs #7152` is enough. Related: the call-site comment claims the clip name is tracked "so the plain let-scanner does not also try to bind it" — my probe shows that tracking does not cover the modifier case, which is #7163.

### Could NOT verify

- **Any F# legality claim** — no F# toolchain on this machine. Everything above is derived from the spec grammar, MS Learn, and `FSComp.txt` message ids, and I say so at each point.
- Whether `(| Even | Odd |)` (spaces inside the clip) is legal F#; it does not match either regex. Out of this arm's scope (clip-body class, not the modifier group).

### Deserves its own issue, not a change here

**`and`-continued active patterns are never extracted.** `let rec (|A|B|) x = … and (|C|D|) y = …` yields **1** definition, not 2, under both the old and the new regex — both scanners anchor on `^[ \t]*let`, and F# uses `and` for mutually recursive bindings. Independent of the modifier axis and of #7144's name-keyed dedup. (Same probe: an attribute on the declaration line, `[<Literal>] let private (|A|B|) n =`, is also missed — same `^…let` anchor.)

---

**REQUEST-CHANGES.** No behaviour bug and no coverage hole in the six graded words; the fix is right and the mutant table is the most accurate I have re-scored on this issue. The items are cheap and all in the class this repo keeps paying for — a claim asserted in prose that nothing observes.

1. Add the ~8-line in-package assertion that `activePatternRE`'s modifier prefix is **byte-identical** to `letRE`'s. This is the blocker: it closes the unbounded widening direction (`sealed` is ALIVE at 0 today) and pins the parity claim the whole allowlist choice rests on.
2. Reword "no entity, no edge, no diagnostic" (body + test-file header) — pre-fix, `letRE` minted a `SCOPE.Operation` named `private` on that line.
3. Stop calling `extra access` order lenience in this arm's prose, and add the measured fact that `let inline private (|A|B|)` / `let rec private (|A|B|)` are spec-legal two-modifier forms the old regex missed — the repeating group is recall, not only lenience.
4. Label the lenience rows in `Indentation`, `MatchSite` and `GroupIsRepeating` (or narrow "labelled per row" to `NeverMissed`), and note that 16 subtests depend on the multiset lenience.
5. Add `Refs #7152` with one sentence: this widening enlarges the raw-`src` phantom surface for commented-out modifier-carrying active patterns. Replace the `mutable` justification with the grammar/`FSComp` 874/831 citation.

PR #7165, opened 2026-09-17
test(#7162): the backoff GROWTH FACTOR was ungraded — a floor on one wait cannot see it, only a ratio between two

### LINKED ISSUE #7162: SHIPPED (f9e7838e9): the backoff growth FACTOR was ungraded — pinned by a RATIO; 'enlargement is the direction nothing sees' was my claim and is FALSE (large enlargement was always visible; #7167 found two more sites)
## What

The engine supervisor's backoff **doubling factor** is ungraded. `*backoff *= 2` → `*backoff *= 3` leaves all of `./internal/daemon/` green: **0 `--- FAIL`**.

Measured by the independent adversarial reviewer on #7160 (`internal/daemon/supervise.go`), full package, `-count=1`, `go vet` clean.

**Why the nearest existing pin cannot see it:** #7123's shrink arm asserts the second wait is a real doubling with a floor of `>=600ms` against an injected `backoffInitial = 400ms`. Under a tripling factor the second wait is **1.2s**, which satisfies `>=600ms` comfortably. The assertion pins *"growth happened"*, not *"growth is by a factor of two"*, and the two are different claims.

## This is the THIRD ungraded role on the same two lines

The sequence is worth stating, because the pattern is the point:

```go
*backoff *= 2                      // <- role 3: the FACTOR (this issue, ungraded)
if *backoff > s.backoffMax {       // <- role 2: the CLAMP (#7126, fixed in #7160)
    *backoff = s.backoffMax
}
```

- **Role 1 — give-up threshold.** `*backoff >= s.backoffMax` increments `ceilingHits`. Graded both directions by #7111/#7123.
- **Role 2 — clamp on the actual wait.** Ungraded; found by #7126, fixed in #7160 with `TestEngineSupervisor_BackoffWaitsAreBoundedByTheCeiling`.
- **Role 3 — the growth factor.** Ungraded, this issue.

Each was found only after the previous one was pinned thoroughly enough for the next gap to stand out. **The general lesson: a single expression can carry several independent decisions, and grading one of them reads as grading the line.** #7126 called out the same shape in #7113 (`leafTypeName` unwrapping arrays — right for type references, wrong for receivers).

## Consequence, and it is deliberately modest

A wrong factor changes the *shape* of the retry curve, not whether the supervisor eventually gives up. With the clamp now pinned by #7160, waits still cannot exceed the counted ceiling, so a tripling factor reaches the ceiling sooner and spends fewer intermediate waits getting there. That is a tuning regression, not a correctness one.

**Not claimed:** no measured time-to-give-up under a changed factor, and **specifically not** that the tail is unbounded. That word was wrong when #7126 used it and wrong when I repeated it in a dispatch brief — the reviewer measured the crash path at production `maxCeilingHits=3` as clamped `[1 2 4 4]` vs deleted `[1 2 4 8]`, the **same wait count**, both giving up, 91.5s → 127.5s (1.39x). Any grading of this factor should assert the *sequence shape*, not a duration.

## Grading requirement

Assert the **ratio between consecutive announced waits**, not a floor on one of them. #7160 established the seam: the announced wait sequence is observable deterministically (it asserts no wait exceeds `s.backoffMax`, plus >=3 waits sitting at it), and that test PASSES under `-race` in 0.06s with an identical sequence — so this is gradable without any timing dependence.

- A row asserting consecutive ratios are exactly 2 until the clamp engages. `*= 3` must die, and so must `*= 1` (no growth) and `*= 4`.
- Do **not** assert the literal sequence `[1 2 4 4 4 4 4]` — that encodes `backoffInitial`, `maxCeilingHits` and the factor together, and would break on any legitimate change to the other two. #7160's review specifically checked that its test pins the bound rather than the sequence, and found it does; keep that property.
- Score the factor in both directions. Shrinking it is the direction an existing floor-style assertion already half-covers; **enlarging** it is the direction nothing sees, and enlargement is what a refactor reaches for.

Related: #7126 (the clamp, fixed in #7160), #7111/#7123 (the threshold), #7113 (the same one-symbol-two-roles shape in the java receiver typing).

cajasmota:
## What

`*backoff *= 2` → `*= 3` left all of `./internal/daemon/` green (**0 `--- FAIL`**). Re-derived on `main` (at `45f9745d0`) before touching anything: baseline 0 `--- FAIL`, exit 0, and the `*= 3` mutant with the new file held out is still 0 `--- FAIL`.

Why the nearest pin cannot see it: #7123's shrink arm (`TestEngineSupervisor_DefaultBackoffCeilingLeavesRoomToGrow`) asserts the second wait is `>= 600ms` against an injected `backoffInitial = 400ms`. Under a tripling factor that wait is **1.2s**, which clears the floor comfortably. That arm pins *"growth happened"*; this one pins *"growth is by a factor of two"*. A **floor on one wait cannot express the second claim** — only a **ratio between two consecutive waits** can.

## The assertion is a ratio, not a sequence

`TestEngineSupervisor_BackoffGrowsByExactlyTwoPerStep`:

```
for every consecutive pair (a, b) of announced waits with b < s.backoffMax:  b must equal 2*a
```

- **Not the literal sequence** (`[1 2 4 8 16 25 25]`). A sequence encodes `backoffInitial`, the budget and the factor together; #7126's review specifically verified its test pinned the **bound** rather than the sequence, and this keeps that property by pinning a **relation between neighbours**. Each checked pair is self-contained — it says only "this wait is twice the previous one" — so retuning the injected initial wait or the budget changes how *many* pairs are graded, never what any one of them asserts.
- The `b < s.backoffMax` guard keeps the clamp out of the ratio: once the clamp engages, `b` **is** the ceiling and the step ratio is the ceiling's, not the factor's.
- The injected tuning is 1ms against a 25ms ceiling — deliberately **not** a power-of-two ratio, so no doubling lands exactly on the ceiling and every pre-clamp step is visible to the guard.
- A **vacuity floor** requires at least two graded pairs, so a sequence that began at the ceiling cannot pass having asserted nothing.

## No timing dependence

Reuses the seam #7160 established: the artefact is the **announced** wait sequence (`"relaunching engine after backoff"`, `backoff=<d>` — the same local handed to `time.NewTimer`), read with #7126's existing `relaunchWaits`, unchanged. Nothing is gated on elapsed wall clock. Under `-race` the new test passes in **0.13s** with an identical sequence (`-race` moves this package's timings ~7x and the automatic CI legs run without it).

## Mutant table

Full `./internal/daemon/` per row, `-count=1`, `go vet` exit code captured to its own file (exit **0** on every row, so no row is a broken build), every mutant applied with an asserted exact match count of 1 and restored by `cp` with `shasum` + `git diff HEAD` verified.

| row | `--- FAIL` | killed by | disposition |
|---|---|---|---|
| baseline (`main`, no new file) | 0 | — | green |
| clean tree **with** new file | 0 | — | green |
| `*= 3`, **new file HELD OUT** | **0** | nothing | **ALIVE — positive control** |
| `*= 3` | 1 | this test only | DEAD |
| `*= 4` | 1 | this test only | DEAD |
| `*= 1` | 6 | this test + 5 pre-existing | DEAD (coarse — see below) |

Observed sequences: `*= 3` → `[1ms 3ms 9ms 25ms 25ms 25ms 25ms]`, `*= 4` → `[1ms 4ms 16ms 25ms 25ms 25ms 25ms]`, `*= 1` → `[1ms 1ms 1ms 1ms 1ms 1ms 1ms]`.

**Positive-control proof of the hold-out:** `go test -list '.*'` reports **484** compiled tests with the file moved aside versus **485** with it, and greps `0` for the new test's name in both the listing and the run log. So the DEAD verdicts are earned by this file, not inherited.

**Every kill came from the ratio assertion** — the vacuity floor fired in **zero** rows (`must grow by exactly`: 2/2/6 occurrences; `would pass vacuously`: 0/0/0).

**Confound, declared:** `*= 1` is a **coarse** mutant — it removes growth altogether, so five pre-existing tests also react to it. The **enlargement** rows (`*= 3`, `*= 4`) are the ones that isolate this role: nothing else in the package reacts to them at all, and enlargement is the direction nothing previously saw.

## Three roles, three disjoint killers — measured, not asserted

| role | mutant | `--- FAIL` | killer |
|---|---|---|---|
| 1 — give-up threshold | `defaultEngineBackoffMax` 30s → 10m | 1 | `TestEngineSupervisor_DefaultBackoffCeilingIsReachedByAMinuteScaleWait` |
| 2 — clamp on each wait | clamp block deleted | 1 | `TestEngineSupervisor_BackoffWaitsAreBoundedByTheCeiling` |
| 3 — growth factor | `*= 3` / `*= 4` | 1 | `TestEngineSupervisor_BackoffGrowsByExactlyTwoPerStep` |

Each mutant is killed by **exactly one** test, and the other two stay green — so the "three roles" framing is earned rather than claimed. The new file is blind to the clamp deletion by construction: that yields `1, 2, 4, 8, 16, 32, 64`, and the pairs the guard still checks all have ratio 2. Role 1's arms settle their verdict *before* any wait is taken, so they are factor-independent.

**No sibling pin is re-tuned, re-keyed or weakened.** The only change to `supervise_backoff_clamp_test.go` is its doc comment, which said the factor was "a third ungraded role, filed separately" and now points at this file plus the disjointness measurement. Both siblings' killers are green on the final tree.

## Characterisation, verified rather than repeated

A wrong factor changes the **shape** of the retry curve — how many short relaunch attempts happen before the cadence reaches the ceiling. It is **not** an unbounded tail: role 2's clamp caps each wait and `maxCeilingHits` ends the loop, so both the wait values and the wait **count** stay bounded. (That adjective was measured false for role 2 in #7160 and is equally false here — stated deliberately, since the predecessor's retracted wording is exactly what this line invites.) An enlarged factor is a **tuning** regression, not a correctness or bound one.

## Axes

**VARIED:** the growth factor (scored at 1, 3 and 4 as well as 2 — both directions), and the number of consecutive pre-clamp steps available to measure (four, versus the single step #7123's floor looks at).
**HELD CONSTANT:** the failure mode (a construction failure that fails identically every time, so no child process ever exists and nothing depends on child scheduling), the injected initial:ceiling ratio, and the spawn budget.

## Verification

- `./internal/daemon/` on the final tree: **0 `--- FAIL`**, exit 0, 47.9s.
- `go vet ./internal/daemon/` exit 0; `gofmt -l internal/daemon/` empty.
- `-race` on the new test and #7126's sibling: both PASS, 0 `--- FAIL`.
- **`./cmd/grafel/` deliberately not run:** the whole diff is inside `internal/daemon/` and touches no extractor, entity kind or digest surface, so the graph digest pin cannot move.

## Not changed, with reasons

- **#7123's `>= 600ms` floor arm** — left exactly as it is. It grades a *real slept interval*, which this file deliberately does not re-grade; re-tuning it to see the factor is what the issue warns against.
- **The `>` vs `>=` inconsistency in the clamp** — already recorded in-source as ALIVE + equivalent with its reasoning, and left untested on purpose. Untouched.
- **No assertion on how long the whole walk takes.** A duration would be scheduling luck under `-race`.

Closes #7162


cajasmota:
## Adversarial review of #7165 — independent re-derivation

Independent reviewer, own detached worktree at `origin/wt-7162-a1` (`b16d5106f`). Every number below is
**re-derived from scratch**, not read off the PR body. Full `./internal/daemon/` per row, `-count=1`,
`-timeout 1800s`, `go vet` exit code captured to its own file (exit **0** on every row, so no row is a
broken build), every mutant applied with an asserted exact match count and the mutated line **read back**
before running, restored by `cp` with `shasum` + `git diff HEAD` (0 on every row, `gofmt -l` empty).
`./cmd/grafel/` deliberately not run: the diff never leaves `internal/daemon/` and touches no extractor,
entity kind or digest surface — I agree with the PR's reasoning there.

**Every one of the seven attacks I was pointed at failed to land.** The claimed table reproduces exactly.

### Re-derived rows

| row | my `--- FAIL` | claimed | killed by | disposition |
|---|---|---|---|---|
| baseline (tree as pushed) | **0** | 0 | — | green |
| `*= 3`, new file **HELD OUT** | **0** | 0 | nothing | **ALIVE — positive control reproduced** |
| `*= 3` | **1** | 1 | new test only, via the **ratio** (line 238) | DEAD |
| `*= 4` | **1** | 1 | new test only, via the **ratio** | DEAD |
| `*= 1` | **6** | 6 | new test (via the **ratio**, not the floor) + 5 pre-existing | DEAD, confound honest |
| clamp block deleted | **1** | 1 | `…BackoffWaitsAreBoundedByTheCeiling` | role 2 only; new test GREEN |
| `defaultEngineBackoffMax` 30s→10m | **1** | 1 | `…DefaultBackoffCeilingIsReachedByAMinuteScaleWait` | role 1 only; new test GREEN |

The five pre-existing reactors on `*= 1`, named: `…BackoffWaitsAreBoundedByTheCeiling`,
`…DefaultBackoffCeilingLeavesRoomToGrow`, `…DefaultCrashBudgetIsBounded`,
`…RepeatedSpawnFailureStopsSoonerAndIsDistinguishable`, `…UnkeepableEngineIsFatal`. None of them is the new
test's killer in disguise: on `*= 1` the new test fails at **line 238 (the ratio)**, and the clamp pin fails
on its own at-ceiling floor (`only 0 of 7 relaunch waits sat at the 4ms ceiling`). Declaration is complete.

**My own hold-out proof** (attack 2, not reusing theirs): `go test -list '.*' ./internal/daemon/` →
**485** `^Test` lines with the file, **484** with it moved aside; the test name greps **1** in the with-listing,
**0** in the without-listing, and **0** occurrences in the entire held-out run log (so zero `=== RUN` for it).
Their 484/485 is correct.

### Attack 1 — the `b < s.backoffMax` guard (the crux)

**Pairs that actually survive the guard: 4 of 6.** Measured directly, not inferred: the sequence is
`[1ms 2ms 4ms 8ms 16ms 25ms 25ms]` (6 consecutive pairs), and forcing the guard to exclude **nothing**
(`if false && cur >= factorProbeMax`) fails on *exactly* waits 6 and 7 — `16ms→25ms` (1.56x) and `25ms→25ms`
(1x). So the guard removes precisely those two and grades `(1,2) (2,4) (4,8) (8,16)`. The "every consecutive
pair" framing is **four** observations, not one wearing a quantifier, and the guard is load-bearing in both
directions (removing it breaks a green tree).

The nastiest version — **a filter that empties its own input set** — does not land:
`if true || cur >= factorProbeMax` → **1 `--- FAIL`**, at line 248, the vacuity floor:
`only 0 of 6 consecutive-wait ratios sat below the 25ms ceiling … want at least 2`. It fails loudly instead
of passing vacuously. That floor is not decorative: it is the sole killer of `*= 100` (see below) and of a
retuned ceiling (next section).

### Attacks 1b/5 — retune sensitivity and the exactly-on-the-ceiling neighbour

I retuned `factorProbeMax` and ran the new test alone (vet 0 each time):

| `factorProbeMax` | sequence | graded pairs | result |
|---|---|---|---|
| 25ms (shipped) | `[1 2 4 8 16 25 25]` | 4 | PASS |
| **16ms (a step lands EXACTLY on the ceiling)** | `[1 2 4 8 16 16 16]` | 3 | **PASS** |
| 8ms | `[1 2 4 8 8 8 8]` | 2 (= the floor) | PASS |
| 4ms | `[1 2 4 4 4 4 4]` | 1 | **FAIL** (line 248, the floor) |

So the answer to the neighbour question: a step landing **exactly** on the ceiling is **skipped, never
misfired** (`cur >= factorProbeMax`), and over-skipping is caught by the floor. The "not a power of two"
tuning therefore **maximises the number of graded pairs** rather than being required for soundness — a
stronger position than the comment claims for it. Retuning down to a ceiling too close to the initial wait
fails loudly. Attack 5 does not land.

### Attack 6 — timing dependence

The new test contains exactly two time constructs: `sup.drainTimeout = 2 * time.Second` and a
`time.After(60 * time.Second)` outer bound whose message correctly disclaims being a statement about any
default. No sleep, no deadline, no elapsed-duration assertion; the artefact is the **announced** wait, read
through #7126's unchanged `relaunchWaits`. Under `-race` I ran the new test plus all three sibling killers:
`0 --- FAIL`, exit 0 — new test **0.13s** (their figure), clamp pin 0.07s, #7123's floor arm 1.22s, role 1's
arm 1.65s. No timing window to exploit.

### Attack 7 — #7123's floor arm

`gh pr diff` touches three files only; `supervise_defaults_backoff_test.go` is **not** among them, so nothing
this PR does changes what that arm observes. I read it: it grades `clock.gap(2)`, a **really slept** interval,
which is a genuinely different observable from the announced sequence. It passes on the final tree under both
regimes. Leaving it alone is the right call.

### The one thing that did land — attack 3's sharper question

> is there any *enlargement* mutant that a pre-existing test also kills?

**Yes.** `*backoff *= 100`: **2 `--- FAIL`** —

- `…BackoffGrowsByExactlyTwoPerStep`, via the **vacuity floor** (`only 0 of 6 … ratios sat below the 25ms
  ceiling`, sequence `[1ms 25ms 25ms 25ms 25ms 25ms 25ms]`), and
- `…DefaultBackoffCeilingLeavesRoomToGrow` (pre-existing), `only 2 spawn attempt(s) observed within the
  test's own 30s outer bound` — the second real wait becomes the 30s production ceiling and blows its bound.

This does **not** weaken the PR's per-row claim, which is correctly scoped: nothing else reacts to `*= 3` or
`*= 4`, measured. But it makes one **unqualified** sentence in the new file's AXES block measurably false:

> VARIED: the growth factor … *with ENLARGEMENT the direction nothing previously saw*

Enlargement is not a direction nothing previously saw — it is *small* enlargement (3x, 4x, i.e. everything
that keeps the injected/production waits inside the sibling pins' outer bounds) that nothing saw. A 100x
enlargement was already visible to #7123's floor arm, as a bound blow-out. This is the repo's named dominant
defect class (prose asserting what no test observes) sitting in a permanent source comment, so I am asking
for the clause, not for any code.

Incidentally this *strengthens* the case for the pin: the hole is exactly the realistic-regression window a
bound-based arm cannot reach.

### Two ungraded twins of the same role — separate issue, NOT this PR

Standing pair-audit: `*backoff *= 2` has **one** site in `supervise.go`, but the same package carries two
more growth-factor sites. I scored both with `*= 3`, full package, vet 0:

| site | mutant | `--- FAIL` | disposition |
|---|---|---|---|
| `internal/daemon/requests_drain.go:37` — `rebuildBackoff`'s `d *= 2` | `*= 3` | **0** | **ALIVE + ungraded** |
| `internal/daemon/server.go:979` — `acceptLoop` transient-error `backoff *= 2` | `*= 3` | **0** | **ALIVE + ungraded** |

Same role, same package, same blindness (both have a clamp and, in `rebuildBackoff`'s case, a cap-return, and
both are pinned only through bounds). This is out of scope for #7162 and should not grow this PR — filing it
is the right move.

### Not verified

- I did **not** score the announced-vs-actually-slept seam (`time.NewTimer(wait)` → `wait * 4`). Derived, not
  executed: an enlargement of the *real* sleep is invisible to every announcement-based pin here, while the
  *shrink* direction is covered by `clock.gap(2)` in #7123's arm. Skipped on the machine budget because a 4x
  real sleep confounds the 60s/90s legs; flagging it as a reading, not a measurement.
- `./cmd/grafel/` not run, by design (stated above).
- Windows/Linux legs: not run locally; CI's business.

## REQUEST-CHANGES

Everything measured in this PR is accurate, the positive control is genuine, the guard is non-vacuous and
retune-resilient, and the three-roles disjointness is earned. One required change, prose only:

1. In `supervise_backoff_factor_test.go`'s AXES block, qualify *"with ENLARGEMENT the direction nothing
   previously saw"* — measured false at `*= 100`, which yields **2 `--- FAIL`** including the pre-existing
   `…DefaultBackoffCeilingLeavesRoomToGrow` (bound blow-out at its own 30s outer bound). The true and
   stronger statement is that **small** enlargement — the factor changes that keep every wait inside the
   sibling pins' bounds, e.g. 3x and 4x — is the direction nothing previously saw, and that a large enough
   enlargement is caught (imprecisely, as a timeout) by a bound-based arm. Optionally record the `*= 100`
   row: it is the only row in which this file's **vacuity floor** is the killer, which is useful evidence
   that the floor is live.

No other change requested. With that clause fixed this is an approve from me.

cajasmota:
## Coordinator mutants ET-1/ET-2 — a NON-INTEGER factor, which nobody sampled: **DEAD 1**, and **ALIVE 0** with the new file held out

Every factor row so far lands on an integer: the author scored `*= 1`, `*= 3`, `*= 4`; the review added `*= 100`. The pin is an exact equality (`cur != factorProbeExpectedFactor * prev`), so the open question was never *"does 1.5x die"* — it was **whether the fix closes the whole band or only the points that were sampled**, and integer nanosecond arithmetic is exactly where a rounding artefact could put a step back onto an exact doubling.

Baseline re-derived on `4250e7f43`: **0 `--- FAIL`**, exit 0, 52.8s. Every row: full `./internal/daemon/`, `-count=1`, match count asserted before the edit, mutated line **read back** after applying it, `go vet` exit captured to its own file (0 on both rows), restored by `cp` with the shasum equal, `git diff HEAD` empty, marker count 0.

| mutant | new test | `--- FAIL` | verdict |
|---|---|---|---|
| **ET-1** `*backoff = *backoff * 3 / 2` (1.5x) | present | **1** — `…BackoffGrowsByExactlyTwoPerStep` | **DEAD** |
| **ET-2** same edit | **HELD OUT** | **0** | **ALIVE** |

ET-2's hold-out is positively asserted, not inferred: **zero** occurrences of the held-out test name anywhere in the run log.

The failing message reports **`grew by 1.5x`** at the first graded pair, twice — so the kill is attributable to the ratio itself rather than to a count or a downstream timeout.

### What this adds beyond the existing rows

1. **The band is closed continuously, not at sampled points.** No factor other than exactly 2 survives, integer or not. That matters for how the AXES paragraph reads: *"small enlargement was the direction nothing previously saw"* is now unambiguously a statement about the **pre-fix** state. A reader should not infer that some small-factor band still survives post-fix, and ET-2 is the measurement that rules it out.
2. **The non-integer band was ungraded before this PR, and this file alone closes it.** ET-2 ALIVE at 0 with the file removed means no pre-existing test — not #7123's `>=600ms` floor, not the clamp pin, not the ceiling-reachability test — reacts to a 1.5x factor. The author's hold-out control established that for `*= 3`; this extends it to a shape no row sampled.
3. **No rounding artefact.** `1e6 * 3 / 2` in integer nanoseconds produced a clean 1.5x at every step rather than drifting onto a doubling, which was the specific risk worth measuring instead of reasoning about.

### Method note

I read the mutated line back before running, rather than trusting the applied-count. Earlier tonight a marker comment of mine landed **inside a raw-string regex literal** on a different PR, producing a completely different pattern that `go vet` passed cleanly — `vet` proves a mutant compiles, never that it says what you meant. For ET-2 I also restored the held-out test file by `cp` and asserted its presence explicitly, because `git diff HEAD` is blind to a removed file and `git checkout --` silently no-ops on an untracked one.

Both rows are recorded; neither blocks. Combined with the review's `*= 100`, the factor's detectability is now mapped across the full range rather than at four integer points.

Issue #7166, opened 2026-09-17
ARM 1 SHIPPED (bfd1e6043). REMAINING — arm 2: a same-line attribute drops the declaration entirely, '[<SomeAttr>] let (|Even|Odd|) x =' yields no pattern AND no case entities; needs ^[ \t]* itself loosened


cajasmota:
## What

Two separate misses at the same anchor in `internal/extractors/fsharp/compexpr_active_patterns.go`. Both reproduce **before and after** #7164, so neither is caused by it.

`activePatternRE` is anchored `(?m)^([ \t]*)let…`, which requires the declaration to begin its own line with `let` as the first non-whitespace token.

**1. `and`-continued active patterns are never extracted.** F# continues a recursive group with `and`, not a second `let`:

```fsharp
let rec (|Even|Odd|) x = …
and (|Positive|Negative|) y = …
```

Measured: **1 entity, not 2** — the `and`-continued pattern is invisible, before and after #7164.

**2. A same-line attribute drops the declaration entirely.**

```fsharp
[<SomeAttr>] let (|Even|Odd|) x = …
```

The `^[ \t]*` prefix cannot reach a `let` that follows `[<…>]` on the same line, so the whole declaration is missed.

Both found by the independent adversarial reviewer on #7164 while verifying that PR's modifier widening; both explicitly **not** #7164's to fix.

## Direction, and how it differs from the modifier defect #7164 fixed

#7164 fixed a modifier-prefix miss where the head *also* minted a phantom — `letRE` caught `let private (|Even|Odd|)` and named it `private`. **Check whether the same fallback fires here**, because it changes the grading requirement:

- If `letRE` (or another scanner) catches the `and`-continued or attributed form and mis-names it, then a **must-have row alone is insufficient** — the phantom needs a forbidden row too, or the fix will look complete while leaving a wrong entity behind.
- If nothing catches it, an ordinary must-have row suffices.

I have **not** measured which; the reviewer measured the entity *count* (1 vs 2), not what else the tree emitted. That is the first thing to establish.

## Not claimed

- **No corpus incidence.** `let rec … and …` is the idiomatic way to write mutually recursive active patterns, and a same-line attribute is ordinary in Checker-style or serialisation-annotated code, but I have counted neither.
[...]

Issue #7156, opened 2026-09-17
fsharp: typeRE's [A-Z] anchor drops lower-case type names ('type myLowercase =') — but the WIDENING direction is already graded, so establish what the anchor is buying before touching it


cajasmota:
## What

Both F# type scanners require an **uppercase initial** for the type name:

```go
// internal/extractors/fsharp/extractor.go:200
typeRE = regexp.MustCompile(
    `(?m)^([ \t]*)type\s+([A-Z][a-zA-Z0-9_']*)\s*(?:<[^>]*>)?\s*(?:\([^)]*\))?\s*=`,
)
// :205 — typeKindRE, same [A-Z] anchor (unreferenced; see #7153)
```

F# does not require an uppercase initial on a type name. It is a strong convention, not grammar. So `type myLowercase = ...` is **never extracted**.

Verified by reading the patterns on `e96eb099f`, and independently confirmed by the #7150 reviewer.

## Provenance, and why it is filed separately

Raised by me while briefing the #7150 adversarial review, as an explicit out-of-scope item so it could not be mistaken for something that PR introduced or fixed. The reviewer confirmed it and agreed it belongs on its own ticket. **#7150 neither caused nor worsened it** — that PR widens the *modifier* prefix (`type private Foo =`), and the `[A-Z]` anchor governs the *name* that follows.

## What makes this one different from the neighbouring F# misses

The widening direction here is **already graded**. The reviewer measured that loosening the anchor is caught — `TestFSharpFieldTypeRefs_NoPrimitiveBlocklistIsNeeded…` dies on it. So unlike #7151 (local modules, where the over-broad direction has no forbidden row) this axis is pinned on one side. That changes the shape of the work: a fix has to widen the anchor **without** tripping the existing guard, which means establishing precisely which lowercase forms are legal type declarations and which are something else the anchor is currently, usefully, excluding.

That last point is the real question, and it should be answered before any pattern is touched: **what is `[A-Z]` currently buying?** A lowercase identifier after `type` also appears in shapes that are not a type declaration this extractor wants — and the guard that kills the widening mutant exists because *something* depended on the exclusion. Establish what, with a citation, rather than assuming the anchor is pure convention-encoding.

## Not claimed

- **No incidence measurement.** Lowercase F# type names are unconventional; I have not counted them in a corpus, and it is entirely possible the real-world rate is near zero, which would make this a correctness note rather than a recall gap worth spending on. **Measure before fixing.**
[...]

Issue #7151, opened 2026-09-17
fsharp: moduleRE is anchored \s*$, so LOCAL modules ('module Foo =') are never extracted at all — a missing form, not a missing modifier


cajasmota:
## What

`moduleRE` is anchored `\s*$`, so a module declaration with anything after the name does not match:

```go
// internal/extractors/fsharp/extractor.go:51
moduleRE = regexp.MustCompile(
    `(?m)^([ \t]*)module(?:\s+rec)?\s+([\w.]+)\s*$`,
)
```

F# has two module forms. A **top-level** module is `module Foo` on its own line — that matches. A **local** module is

```fsharp
module Foo =
    let helper x = x + 1
```

and the trailing ` =` means `\s*$` cannot match it. **Local modules are never extracted at all.** Verified by reading the pattern at `extractor.go:51` on `e96eb099f`; the `namespaceRE` sibling at `:56` has the same `\s*$` shape and is *correct* there, because a namespace declaration legitimately has nothing after the name.

## Provenance and what is NOT claimed

Found by the implementation agent on the #7135 `moduleRE`/`typeRE` arm while widening the modifier allowlist, and deliberately **not fixed there** — it is orthogonal to that arm (a missing *form*, not a missing *modifier*) and reproduces with no modifier at all.

- **Not measured for incidence.** `module Foo =` is an ordinary, idiomatic F# shape, but I have not counted it in a corpus.
- **Derived, not executed.** There is no F# toolchain on this machine (`dotnet`, `fsc`, `fsharpc`, `fsi`, `mono` all absent; `javac` is the only compiler here). The claim rests on reading the anchor, not on running F#.
- **Direction is a MISS, not a collapse.** Nothing is mis-named or merged; an entity that should exist simply never appears. That is the direction recall-style must-have assertions *can* see — so unlike a collapse, this one is gradable by an ordinary must-have row.

## Grading requirement

The #7135 agent pinned the modifier widening as a **relation** rather than as independent rows — i.e. the tests assert that the presence of a modifier makes *no difference* to whether a module is extracted. That has a useful consequence here: **a one-sided future fix fails.** If someone fixes the local-module form for `module Foo =` but not for `module private Foo =`, the relation rows go red rather than silently drifting apart. Keep that property when fixing this.

The fix needs, at minimum:
- a must-have row for `module Foo =` with a nested binding, and
- the same row with a modifier (`module private Foo =`), so the two forms stay tied together.

## Adjacent, do not conflate

[...]

PR #7169, opened 2026-09-17
fix(#7163): letRE minted a phantom operation named after a modifier — the repeated group gives back its LAST modifier and that word becomes the name

### LINKED ISSUE #7163: fsharp: letRE mis-names a modifier-carrying ACTIVE-PATTERN head — `let private (|Even|Odd|)` also mints a phantom operation called `private`

cajasmota:
Closes #7163

## The defect, re-derived from the pattern on disk

`letRE` in `internal/extractors/fsharp/extractor.go`:

```
(?m)^([ \t]*)let(?:\s+(?:rec|mutable|inline|private|internal|public)\b)*\s+([a-zA-Z_][a-zA-Z0-9_']*)\s*(?:<[^>]*>)?\s*(?:[^=\n]*)=
```

The modifier group is **repeated, therefore optional**. On an active-pattern head the
engine takes ` rec` as one repetition, then the mandatory `\s+` plus the name class
`[a-zA-Z_][a-zA-Z0-9_']*` face `(` and cannot proceed — so it **gives back the last
repetition**, `\s+` takes the space after `let`, and the NAME capture takes the modifier
word. `(|Even|Odd|) x ` is swallowed by the trailing `[^=\n]*` and the `=` matches.

Only the **last** repetition is given back, which is what makes the phantom name the
*last* modifier rather than the first. Measured on `f9e7838e9` (these are measured
values, not expectations — every row below is a line lifted from the RED log):

| source | phantom operation minted before the fix |
|---|---|
| `let (\|Even\|Odd\|) x =` | *none* — control; letRE correctly declines |
| `let rec (\|Even\|Odd\|) x =` | `let/rec` (census `[let/rec let/sentinelValue]`) |
| `let private (\|Even\|Odd\|) x =` | `let/private` |
| `let inline private (\|Foo\|_\|) y =` | `let/private` — **PARTIAL backtrack**: ` inline` was KEPT |

## Approach chosen: (b), skip offsets already claimed by `activePatternRE`

New `activePatternLetOffsets(src)` returns the set of line-start offsets at which
`activePatternRE` matched; the `let` loop declines a match starting at a claimed offset.
Both patterns anchor on `(?m)^([ \t]*)let`, so both `m[0]` are the line start and the
offsets are directly comparable — a premise that is **graded**, see M3.

Why not (a) — a post-filter on "the captured name is one of the six modifier words and
the remainder opens a banana clip": it duplicates `letRE`'s allowlist into a **third**
place. Those two allowlists are already pinned byte-for-byte equal by
`TestActivePatternREModifierParityWithLetRE` (#7164) *precisely because* a one-sided edit
to either is invisible; a third copy would need a third parity check, and it would embed
the clip syntax a second time as well. The offset set derives the rejection from the
scanner that already owns both, so a widening of either propagates for free.

Why not (c) — restructuring `letRE`: it is the approach most likely to break that same
parity test, and the issue's own note is right that RE2 has no lookahead, so "a name that
is not a banana clip" cannot be written in the pattern at all. **`letRE` is untouched by
this PR** and the parity test stays green (verified in isolation: 1 `=== RUN`, 1
`--- PASS`, exit 0).

### Residual — corrected, and larger than the first revision of this body said

**The first revision of this body named the one shape that has NO phantom, and that was
wrong in the reassuring direction.** The `and`-continued form (#7166) mints **zero**
phantom, before *and* after this guard: `and private (|On|Off|) y = …` leaves a `let`
census of `sentinelValue` alone, because `and` is not a `let` head and `letRE` never
reaches it. #7166 is a **missing entity**, not a surviving phantom. Re-measured on my own
tree, and now pinned by `TestActivePatternLetPhantom_AndContinuedMintsNoPhantom` so the
claim cannot be re-made from prose.

What genuinely survives is **24 (phrase, head) pairs** — 4 modifier phrases × 10 heads —
in two families. Every one measured **identical before and after this guard**, so they are
pre-existing and neither caused nor cured here:

| head | `let` census, post-fix | `SCOPE.Pattern`? |
|---|---|---|
| `let rec (\| Even \| Odd \|) x = 1` | `rec,sentinelValue` | **no** |
| `let private (\| Even \| Odd \|) x = 1` | `private,sentinelValue` | **no** |
| `let rec (\|Even\|Odd\| ) x = 1` | `rec,sentinelValue` | **no** |
| `let rec ( \|Even\|Odd\|) x = 1` | `rec,sentinelValue` | **no** |
| `let private (x, y) = (1, 2)` | `private,sentinelValue` | n/a |
| `let private (+.) a b = a + b` | `private,sentinelValue` | n/a |
| `let inline private (+.) a b = a + b` | `private,sentinelValue` | n/a |
| `let private { Foo = f } = r` | `private,sentinelValue` | n/a |
| `let private [a; b] = xs` | `private,sentinelValue` | n/a |
| `let private (Some v) = o` | `private,sentinelValue` | n/a |

- **Whitespace-bearing banana clips** (rows 1–4). `activePatternRE`'s clip class
  `[A-Za-z0-9_'|]+` excludes whitespace, so it declines, no offset is claimed, and the
  phantom stands. These are **real active patterns and therefore #7163's own class** — the
  fix is *incomplete for the defect it closes*, not merely adjacent to it — and **both
  halves are wrong on one input**: the phantom survives AND the `SCOPE.Pattern` is lost
  entirely. Whitespace tolerance looks intended: `strings.TrimSpace(c)` at
  `compexpr_active_patterns.go:312` is unreachable under that clip class. Whether such a
  clip is legal F# is **not settled** — no toolchain — so the rows pin the scanner, not
  the language. Being filed as its own issue.
- **Non-clip parenthesised / pattern heads** (rows 5–10). Not active patterns at all, so no
  offset is ever claimed; **outside approach (b)'s reach by construction**, and they need
  the other fix shape. Being filed as its own issue.

All ten are recorded in `TestActivePatternLetPhantom_ResidualHeads`, an **xfail ledger**
that fails loudly with a "move this row to the forbidden table" message the moment one is
fixed — see the mutant M8 below, which proves it is not vacuous. The
`activePatternLetOffsets` doc comment carries the same corrected text.

## Is a legitimate `let` named `rec`/`private`/… possible? No — derived, not executed

Verified against the reserved-word list rather than assumed. F# Language Specification
§ 3.4 "Identifiers and Keywords" lists `rec`, `mutable`, `inline`, `private`, `internal`,
`public` in `ident-keyword` (corroborated by MS Learn "Keyword Reference",
<https://learn.microsoft.com/dotnet/fsharp/language-reference/keyword-reference>). A
keyword is not an `ident`, and `function-defn` / `value-defn` (§ 14.6) bind an
`ident-or-op`, so `let rec = 1` cannot be written. The escaped form ``let ``rec`` = 1``
**is** legal, but its source text contains backticks and so never reaches `letRE`'s name
class — no row forbids it, and none blesses it.

**There is no F# toolchain in this environment** (`dotnet`, `fsc`, `fsharpc`, `fsi`,
`mono` all absent; `javac` is the only compiler present), so every legality claim here is
*derived-not-executed* and cited.

## RED log — 148 `--- FAIL` lines, with a decomposition that sums

```
4 top-level                                      4
  TestActivePatternLetPhantom_IssueRows          3   (of 4 rows; the no-modifier row PASSED)
  TestActivePatternLetPhantom_ModifierSpace    135   (of 140 rows; the 5 empty-phrase rows PASSED)
  TestActivePatternLetPhantom_Indented           3
  TestActivePatternLetPhantom_NoDedupCollapse    3
                                    subtests = 144
                                4 + 144      = 148  ✓
```

`TestActivePatternLetPhantom_PlainLetStillExtracted` contributes **0** — it is a pre-fix
control and was green before the change, which is the point of it.

The 5 passing `ModifierSpace` rows were confirmed by name to be exactly the empty-phrase
row of each clip shape (`-v` run: 141 `=== RUN`, 5 `--- PASS`, all five named `…/`), so
the modifier-dependence of the defect is measured, not argued.

## Grading

**FORBIDDEN rows (the over-firing direction — a must-have row cannot see this).** Every
row asserts that no `SCOPE.Operation` is named `rec`/`mutable`/`inline`/`public`/
`private`/`internal`/`let`, *and* that the `let` census is an **exact list**, not a floor.
The four issue shapes additionally name the specific word that was minted, per row.

**Forbidden names are UNIQUE in every fixture.** `letSeenKey(indent, name)` is
`indent + ":let:" + name`, so a phantom sharing a real declaration's name merges with it
and is invisible to *both* a presence and a count assertion (#7144, #7152). Every fixture
therefore names its one real binding `sentinelValue` — in no modifier allowlist — and
declares nothing whose name is a modifier word.

**Both halves asserted separately**, so a fix that suppresses the phantom by suppressing
the line is caught: `fsPhantomAssertNoPhantom` (forbidden) and
`fsPhantomAssertPatternMinted` (the `SCOPE.Pattern`, its subtype, and its full
`SCOPE.Schema` case set). M5 is the mutant that grades the second half.

**The partial backtrack is explicit**, both as a named row (`inline private (|Foo|_|)` →
expected phantom `private`, not `inline`) and across `ModifierSpace`, which enumerates
both orders of every cross-kind pair rather than hand-picking.

### Axes

**VARIED:** modifier identity (all six allowlisted words), modifier count (0, 1, 2),
modifier order (both orders of every cross-kind pair, plus each access modifier repeated
— 28 phrases via the shared `modifierPhrases` generator), banana-clip shape (two-case,
three-case, partial, parameterised-partial, no-argument — 5 shapes, 140 cross rows),
indentation (top level vs nested), declarations per file (one vs two same-indent), and
the too-strict direction (plain / `rec` / `inline private` / generic bindings, plus a
tuple head and an operator definition).

**HELD CONSTANT:** case names (`Even`/`Odd`, `Zero`/`Pos`/`Neg`, `Positive`,
`DivisibleBy`, `On`/`Off`); the real binding's name (`sentinelValue`); the enclosing
`module Patterns`; the asserted Kinds (`SCOPE.Operation` / `SCOPE.Pattern` /
`SCOPE.Schema`); the file path (`patterns.fs`).

## Mutants — full `./internal/extractors/fsharp/` suite per row, `-count=1`

Every mutant: `cp` backup, exact-match count asserted **before** running, the mutated
line **read back** by `grep -n`, `go vet` exit code captured into its own file (not
through a pipe), `--- FAIL` lines counted (never the exit code), restored by `cp` with
`shasum -c` and a marker count of 0.

|---|---|---|---|---|
| M1 | revert the guard (`if false && apClaimed[m[0]]`) | 0 | **148** | DEAD |
| M2 | too strict: `if len(apClaimed) > 0` — drop *every* let in a file containing any active pattern | 0 | **162** | DEAD |
| M3 | offset premise: claim `m[1]` (match end) instead of `m[0]` (line start) | 0 | **148** | DEAD |
| M5 | suppress the active patterns instead (`_ = apEntities`) | 0 | **338** | DEAD |
| M4 | **positive control** — M1 with this PR's test file held out | 0 | **0** | **ALIVE, ungraded before this PR** |
| M6 | *(built by the reviewer)* suppress **only** the case sub-entities (`_ = caseEnts`) | 0 | **303** | DEAD — **150 of this PR's 150 subtests** fail, so the required half is graded by a TARGETED mutant, not only by coarse M5 |
| M8 | widen `activePatternRE`'s clip class to admit whitespace — i.e. **apply a future fix** to the residual's first family | 0 | **3** | the **xfail ledger fires**, as designed |

**M8 grades the residual ledger itself**, added in review round 2: `[A-Za-z0-9_'|]+` →
`[A-Za-z0-9_'| \t]+` (1 match asserted, line read back as
`` `\s+\(\|([A-Za-z0-9_'| \t]+(?:\|_)?)\|\)\s*([^=\n]*)=` ``, vet rc 0). Result: **3**
`--- FAIL` = 1 top-level + 2 subtests, both in `TestActivePatternLetPhantom_ResidualHeads`,
with the intended message — `RESIDUAL LEDGER MOVED for "let rec (| Even | Odd |) x = 1":
let census="sentinelValue", recorded "rec,sentinelValue"`. Exactly 2 of the 4 whitespace
rows move (the *inner*-space ones); the leading/trailing-space rows stay residual because
that space falls outside the capture group — so the ledger is non-vacuous **and** finer
than the family it records. Restored by `cp`, verified by `diff` (rc 0) and a marker count
of 0.

**Recorded, not fixed** (reviewer's M7, and I agree it should be written down): dropping
only the `CONTAINS` edges is DEAD by the package (142) but **ALIVE w.r.t. all 150 of this
PR's subtests** — these rows grade the pattern and its case entities, *not* the edge.

Decompositions, each summing:

- **M1** = 4 top-level + (3 + 135 + 3 + 3 = 144) = **148** — identical to the RED log,
  which is the evidence the guard is the whole fix and nothing else in the package moved.
- **M2** = 7 top-level + (3 + 4 + 140 + 3 + 5 = 155) = **162**. The informative part is
  *which* rows move: `ModifierSpace` goes 135 → **140**, i.e. the five no-modifier control
  rows now fail too, because the `sentinelValue` binding is dropped. That is the
  too-strict direction being graded, in the opposite direction from M1.
- **M3** = the same 148 as M1, because claiming the wrong offset makes the guard never
  fire. This is the row that grades the premise "both patterns' `m[0]` is the line start":
  had it been ALIVE, the guard would have been decorative.
- **M5** = 12 top-level + (3 + 4 + 140 + 3 + 6 + 7 + 18 + 5 + 140 = 326) = **338**.
  **Confound, reported rather than hidden:** this mutant is coarser than the one behaviour
  it probes — it deletes the whole active-pattern emission, so it also fails five #7135
  tests and three `TestFSharp_ActivePattern_*` tests. It is kept because it is the only
  mutant that grades the *required* half, and my four tests fail under it **independently**
  (150 of the 326 subtest failures are in this PR's file, and `IssueRows` 3 → 4 /
  `ModifierSpace` 135 → 140 shows the required-half rows firing on exactly the rows the
  forbidden half leaves passing).

### Positive control (M4) — proof of the hold-out

Not inherited. The test file was **moved out of the package** (not `-run`-excluded) and
the fix reverted:

```
go test -list '.*'   with the file:  155 tests, 5 named TestActivePatternLetPhantom*
go test -list '.*'   held out:       150 tests, 0 named TestActivePatternLetPhantom*
go test -count=1 -v  held out:       exit 0, 0 '--- FAIL', 0 '=== RUN TestActivePatternLetPhantom'
```

So the phantom was ungraded by the other 150 tests in the package, and every DEAD verdict
above is attributable to this PR's rows.

Restored by `cp` afterwards; `shasum -c` OK on all three files and the mutant markers grep
to 0 — the committed tree is byte-identical to the measured one.

## Gates

- `./internal/extractors/fsharp/` — **0** `--- FAIL`, exit 0, `gofmt -l` clean.
- `TestActivePatternREModifierParityWithLetRE` (#7164) — **green**, run in isolation.
  `letRE` is untouched by this PR.
- `go test -count=1 ./cmd/grafel/` — exit 0, **0** `--- FAIL` (228.9s), with a fresh
  `GRAFEL_HOME` **and** `GRAFEL_DAEMON_ROOT`. **This is regression cover, not coverage.**
  The digest fixture is **`writeSpanFixture6118`** (`custom_extractor_spans_6118_test.go:32`
  — the only function feeding `semanticDigest6118`, at lines 280, 771, 873, 926, 1014). I
  re-enumerated it with a backtick-parity tracker: **18 members** (`cs/` ×7, `java/`, `ts/`,
  `web/server.js`, `myproj/` ×2, `myapp/` ×4, `svc/handler.go`, `rs/routes.rs`) and **zero**
  `.fs`/`.fsx`/`.fsi` — `grep -rn '\.fs"\|\.fsx"\|\.fsi"' cmd/grafel/` is 0 across the
  whole package. So the gate proves this change moved no other language's digest; it grades
  nothing about F#.
  *The "correction to my brief" in the first revision of this body is **withdrawn**.* It
  measured `writeDjangoFixture` (`inproc_custom_extractors_5989_test.go:63`, 10 members) —
  a different function feeding a different test — and used it to retract a true figure. 18
  reproduces exactly on `writeSpanFixture6118`. The conclusion above is unchanged.

## Review round 2 — what changed, and it is prose plus test labels only

Three REQUEST-CHANGES items, all on the honesty surface; the fix itself was found sound
(the reviewer's 9,600-case internal sweep — 15 phrases × 21 heads × 4 indents × 4
separators incl. newline-crossing × LF/CRLF — found **0** cases where the guard suppresses
a non-modifier name, which is a stronger statement of the premise than my M3). Corrected:
**(1)** the residual claim, in this body *and* in the `activePatternLetOffsets` doc comment
— the `and` form has no phantom, 24 pairs in two families do; **(2)** the labelling of
`PlainLetStillExtracted`'s two parenthesised rows, which hold the modifier axis constant,
plus a new xfail ledger and an `and`-form control; **(3)** the digest paragraph, which
measured the wrong fixture and retracted a true figure. No production behaviour changed in
round 2 — the only non-comment edits are test names, two new test functions, and the
fixture strings they need.

`./internal/extractors/fsharp/` after round 2: **exit 0, 0 `--- FAIL`**, `gofmt -l` clean,
`go test -list` 155 → **157** tests (the two new ones). `./cmd/grafel/` was not re-run: no
extraction path changed in round 2.

## Not changed, with reasons

- **`letRE` itself** — approach (c); see above. Touching its modifier group is the one
  edit `TestActivePatternREModifierParityWithLetRE` exists to catch.
- **#7166** (`and`-continued heads, same-line `[<Attr>]`) — the named residual of approach
  (b), filed and out of scope.
- **#7144 / #7152 / #7151 / #7156 / #7158** — out of scope per the brief. #7144 (the
  name-keyed dedup) is *worked around* here by making every forbidden name unique in its
  fixture, not fixed.
- **`let (x, y) = …` and `let (+.) a b = …`** — with **no modifier**, `letRE` never
  extracted these (its name class excludes `(`), measured on the pre-change file; their
  rows assert an **unchanged absence**, so the guard cannot become a fix for a different
  issue by accident. **Corrected labelling:** the first revision of this body presented
  them as covering "the two heads the issue named", but they hold the modifier axis — the
  axis this whole PR is about — CONSTANT at *absent*. Add a modifier and the same heads
  mint a phantom (`let private (x, y)` → census `private,sentinelValue`, measured). The
  rows are now named `tuple head/no modifier` and `operator defn/no modifier`, the test's
  header states what it does **not** grade, and the modifier-carrying counterparts are
  recorded in the residual ledger. A label claiming more than its content is the defect
  family this repo keeps finding.


cajasmota:
## On the fixture-count correction: neither figure is wrong — they are different fixtures, and the ambiguity is my brief's fault

The PR reports that `writeDjangoFixture` has **10** members, not the 18 my dispatch brief cited. Both numbers are correct, for different functions. Enumerated myself on `f9e7838e9` with a backtick-parity walk over each function body:

| fixture | file | members |
|---|---|---|
| **`writeSpanFixture6118`** — the one the DIGEST test builds | `cmd/grafel/custom_extractor_spans_6118_test.go:32` | **18** — 7 `.cs`, 6 `.py`, 1 `.java`, 1 `.ts`, 1 `.js`, 1 `.go`, 1 `.rs` |
| `writeDjangoFixture` | `cmd/grafel/inproc_custom_extractors_5989_test.go:63` | **10** |

**My brief never named the fixture**, so "the fixture has 18 members" was ambiguous in a file that contains several. That is the defect, and it is mine: a count is only checkable if the thing being counted is named. Fixed for future briefs — name the function, not "the fixture".

The conclusion is unchanged either way, and the PR states it correctly: **zero `.fs`/`.fsx` in either**, so for an F#-only change `./cmd/grafel/` is **regression cover, not coverage**. Worth being precise about *which* gate that applies to, though, since the digest pin lives on `writeSpanFixture6118` and that is the one whose stability is being cited.

Also noting the second correction, which stands and is more consequential: **`gh api user` read `cajasmota` on the first call and the account had drifted to the work account by push time.** The pre-write guard caught it. That is the first instance I have seen of the drift happening *inside* a single task rather than between sessions, and it is the case the rule exists for — a passing check earlier in a run does not carry forward. Recorded.

## On the approach

(b) is the right call and the justification is the one that matters: a name-blocklist would put `letRE`'s six-word allowlist in a **third** location, and the two existing copies are pinned byte-equal by `TestActivePatternREModifierParityWithLetRE` *precisely because* a one-sided edit to a duplicated allowlist is invisible. A third copy would need a third parity check plus a second copy of the clip syntax. Choosing the approach that does not add a twin is the lesson from `3bb6d28b8` applied rather than restated.

`letRE` untouched and the parity test green in isolation (1 `=== RUN`, 1 `--- PASS`) is the right thing to have verified, since that test existing at all is what makes approach (c) hazardous.

**M3 is the best row in the table** — claiming `m[1]` instead of `m[0]` grades the *offset premise* the whole guard rests on, not just the guard's presence. A guard keyed on the wrong offset would have looked identical to a working one under M1 alone.

**M5 is correctly declared a confound** (338 `--- FAIL` from suppressing active-pattern entities wholesale) and correctly noted as the only mutant grading the required half. Both halves separately is what the brief asked for; that it takes a coarse mutant to reach one of them is worth the disclosure rather than a quiet omission.

**#7166 named as the residual of (b)** is the disclosure I most wanted: an `and`-continued head is unclaimed by `activePatternRE`, so a phantom there survives — and `letRE` does not reach it either, because an `and` head is not a `let` head. So the fix is complete for what it claims and explicitly incomplete for a shape already filed. That is the honest shape of a scoped fix.

cajasmota:
## Independent adversarial review — #7169

Own detached worktree at `origin/wt-7163-a1` (`cb80dc600`), path asserted inside every script.
Nothing pushed, nothing committed. No F# toolchain here (`dotnet`/`fsc`/`fsharpc`/`fsi`/`mono`
absent), so every legality claim below is **derived-not-executed** and cited.

### What I re-derived and CONFIRMED

| claim | my measurement |
|---|---|
| `letRE` untouched | **confirmed** — whole diff is `416 insertions(+), 0 deletions`; `letRE`'s 4 lines are byte-identical to `origin/main` |
| parity test green in isolation | **confirmed** — 1 `=== RUN`, 1 `--- PASS`, exit 0 |
| baseline green | **confirmed** — `./internal/extractors/fsharp/` exit 0, **0** `--- FAIL`, `gofmt -l` clean |
| RED = 148, decomposition sums | **confirmed** — M1 (`if false && apClaimed[m[0]]`, vet rc 0, marker asserted, 1 match) gives **148** `--- FAIL`; 4 top-level + (IssueRows 3/4 + ModifierSpace 135/140 + Indented 3/3 + NoDedupCollapse 3/3 = 144) = 148, `PlainLetStillExtracted` 0/6 |
| M3 offset premise | **confirmed** — `apClaimed[m[1]]`, vet rc 0, **148** `--- FAIL`, DEAD |
| the 5 passing ModifierSpace rows are the no-modifier controls | **confirmed by name**: `total2/`, `total3/`, `partial/`, `partial/2args/`, `total2/noargs/` — exactly the empty-phrase row of each of the 5 clip shapes. (Nit: 140 subtest `=== RUN`, not the 141 in the body — the body appears to have counted the parent.) |
| **M4 hold-out, re-derived from scratch** | **confirmed, my own numbers**: file `mv`d out of the package + guard reverted → `go test -list '.*'` **155 → 150** tests, `^TestActivePatternLetPhantom` **5 → 0**, `go test -count=1 -v` **exit 0**, **0** `--- FAIL`, **0** `=== RUN TestActivePatternLetPhantom`. Independently reproduced; not inherited. |

### Attacks that FAILED to land (recorded with what I tried)

**1. Offset disagreement between the two scanners (the recall/over-firing direction).** I wrote an
internal-package sweep (`package fsharp`, using `activePatternLetOffsets` and `letRE` directly)
over the cross product of 15 modifier phrases × 21 head shapes × 4 indents (`""`, 4 spaces, tab,
mixed `" \t "`) × 4 separators (incl. `"\n    "`, which exercises `\s+` crossing a newline) × 2
line endings (LF, CRLF) = **9600 sources**, asserting that every `letRE` match at a *claimed*
offset captured a modifier word. **0 violations.** So the guard provably cannot suppress a real
binding on any of those inputs — the premise is stronger than M3 alone shows. The structural reason
is worth putting in the code: `activePatternRE` claiming an offset means the text is
`let <allowlisted-modifiers>* (|`, and from that text `letRE` can only reach its name class by
giving back a modifier, so its capture is *always* a modifier word (and with zero modifiers `letRE`
declines outright and the guard is a no-op).

Individually probed and consistent (no phantom, no divergence): CRLF; `let` at byte 0 with no
preceding newline; tab indent nested; a newline between `rec` and the clip; two clips of equal
line length in one file; same-line `[<Literal>]` (both scanners decline). **A leading BOM** makes
*both* scanners decline — no phantom, but also no active pattern at all; consistent, so not a guard
defect (a pre-existing recall gap, not this PR's).

**2. Forbidden-row name-collision vacuity.** Every fixture's real binding is `sentinelValue` /
`outerValue` / `target`; none is a modifier word, and the census assertions are exact lists. The
"nasty case" the brief asked for — a *legitimate* binding colliding with a forbidden expectation —
**cannot be constructed in legal F#**: `rec`/`private`/… are `ident-keyword`s (spec § 3.4), so
`let private = 0` is not writable, and the double-backtick form carries backticks that never reach
`letRE`'s name class. Attack does not land; the author's handling is correct.

**3. Both halves / targeted mutants (attack 4 — this is where M5's coarseness was the concern).**
I built the targeted mutants the brief asked for:
- **M6** — suppress *only* the case sub-entities (`out = append(out, caseEnts...)` → `_ = caseEnts`;
  vet rc 0, 1 match asserted, line read back): **303** `--- FAIL`, 8 top-level. **DEAD, and this
  PR's rows fail independently: IssueRows 4/4, ModifierSpace 140/140, Indented 3/3,
  NoDedupCollapse 3/3 = 150 of its 150 subtests.** So the required half (pattern + full case set)
  is graded by a *targeted* mutant, not only by coarse M5. This attack fails — M5's confound does
  not leave the direction ungraded.
- **M7** — drop only the `CONTAINS` edges (`Relationships: rels` → `nil`): **142** `--- FAIL`, but
  **0** of this PR's 150 subtests move. So the edge half is **ALIVE with respect to this PR** and
  killed only by pre-existing `TestActivePatternModifiers_NeverMissed` / `TestFSharp_ActivePattern_Total`.
  Graded by the package, not by these rows. Not a change request — recording it so nobody reads
  these rows as covering the edge.

**7. The #7166 residual claim.** Measured `let rec (|Even|Odd|) x = …` / `and private (|On|Off|) y = …`:
census `[let/sentinelValue]`, **zero phantom**, before *and* after the fix. So the second half of
the body's claim ("`letRE` does not reach it either") is right — and therefore the first half
("keeps its phantom") is **wrong**: the `and` form is a *missing entity* (#7166), not a surviving
phantom. See finding 1.

### Attacks that LANDED

#### 1. The named residual is the wrong one, and 8 real residual shapes are unnamed

The body and the `activePatternLetOffsets` doc comment both say: *"a modifier-carrying
active-pattern head that `activePatternRE` itself does NOT match keeps its phantom … The known such
shape is the `and`-continued form filed as #7166."* Measured, the `and` form carries **no** phantom
(above), while these do — enumerated by the same internal sweep, and each verified **identical
before and after the fix** (so pre-existing, neither caused nor cured, but *not* the shape named):

| head | phantom, pre-fix | phantom, post-fix | AP minted? |
|---|---|---|---|
| `let rec (\| Even \| Odd \|) x =` | `rec` | **`rec`** | no |
| `let private (\| Even \| Odd \|) x =` | `private` | **`private`** | no |
| `let rec (\|Even\|Odd\| ) x =` | `rec` | **`rec`** | no |
| `let rec ( \|Even\|Odd\|) x =` | `rec` | **`rec`** | no |
| `let private (x, y) = (1, 2)` | `private` | **`private`** | n/a |
| `let private (+.) a b = a + b` | `private` | **`private`** | n/a |
| `let inline private (+.) a b = a + b` | `private` | **`private`** | n/a |
| `let private { Foo = f } = r` / `let private [a; b] = xs` / `let private (Some v) = o` | `private` | **`private`** | n/a |

Full enumeration: 24 (phrase, head) pairs over 4 phrases × 10 heads survive. The first four rows are
**active-pattern heads** — squarely #7163's own defect statement — and the whitespace-in-clip gap is
**not filed anywhere** (I checked #7144/#7151/#7152/#7156/#7158/#7166 and searched open+closed).

Legality: `let private (x, y) = …`, `let private (+.) a b = …` and `let inline private (+.) …` are
ordinary F# (MS Learn "Access Control" permits `private` on a `let` binding; the parenthesised
operator definition is the documented form) — *derived-not-executed*. Whitespace inside a banana
clip (`(| Even | Odd |)`) I am **less** sure of and could not compile; note that
`compexpr_active_patterns.go:312` already does `strings.TrimSpace(c)` on each case, which is
**unreachable** given the clip class `[A-Za-z0-9_'|]+` excludes whitespace — that dead defensive
code is itself evidence someone intended whitespace tolerance. `let rec (x, y) = …` is probably
*not* legal (`rec` wants an identifier binding), so I excluded the `rec`+non-clip rows from the
claim.

#### 2. The test file varies the modifier axis everywhere EXCEPT on the two rows where the phantom lives

`TestActivePatternLetPhantom_PlainLetStillExtracted` has rows `let (x, y) = (1, 2)` and
`let (+.) a b = a + b` — the two heads #7163's grading requirement names — **with no modifier**,
asserting an unchanged absence. Add the one axis this entire PR is about and they mint a phantom.
Compiling counter-example, run on the merge candidate (`gofmt` clean, exit 1, **6** `--- FAIL`):

```
"let private (x, y) = (1, 2)":          let census="private,sentinelValue", want "sentinelValue"
"let private (+.) a b = a + b":         let census="private,sentinelValue", want "sentinelValue"
"let inline private (+.) a b = a + b":  let census="private,sentinelValue", want "sentinelValue"
"let rec (| Even | Odd |) x = …":       let census="rec,sentinelValue",     want "sentinelValue"
"let private (| Even | Odd |) x = …":   let census="private,sentinelValue", want "sentinelValue"
```

This is the repo's "fixture pins one axis and leaves its neighbour open" pattern, and here the
un-varied neighbour is not hypothetical — it holds live phantoms of the very defect under fix.

#### 3. The `cmd/grafel` digest-coverage paragraph measures the WRONG fixture

The body says: *"I enumerated `writeDjangoFixture` … it holds **10** members … **Correction to my
brief:** I was told the fixture has 18 members; measured, it has 10, and a naive grep agrees at 10
here — the 18 figure does not reproduce on this file."*

The digest fixture is not `writeDjangoFixture`. `semanticDigest6118` is called only from tests fed
by **`writeSpanFixture6118`** (`custom_extractor_spans_6118_test.go:32`, used at lines 280, 771,
873, 926, 1014). I enumerated it with a backtick-parity scanner: **18 members** —
`cs/OrdersController.cs`, `cs/ShopContext.cs`, `cs/OrderCreatedConsumer.cs`,
`cs/CreateOrderHandler.cs`, `cs/OrderValidator.cs`, `cs/CleanupJob.cs`, `cs/Startup.cs`,
`java/OrdersController.java`, `ts/routes.ts`, `web/server.js`, `myproj/settings.py`,
`myproj/urls.py`, `myapp/models.py`, `myapp/serializers.py`, `myapp/views.py`, `myapp/tasks.py`,
`svc/handler.go`, `rs/routes.rs`. Zero `.fs`/`.fsx`/`.fsi`.

So the *conclusion* ("regression cover, not coverage") is right, but it is supported by a
measurement of a different function, and the body publishes a "correction" that retracts a true
figure. The 18 reproduces exactly.

### Not verified

- Legality of whitespace inside a banana clip — no F# compiler here. If it is *illegal*, rows 1–4 of
  finding 1 collapse and only the non-clip heads remain (findings 1 and 2 still stand on those).
- Corpus incidence of any residual shape — not measured, as in the issue.
- Nothing about F# is covered by the `cmd/grafel` gate. I ran it myself with a fresh `GRAFEL_HOME`
  **and** `GRAFEL_DAEMON_ROOT`: **exit 0, 0 `--- FAIL`, 213.6s** — reproduced, and the body's
  "regression cover, not coverage" framing is the right one (see finding 3 for the fixture it
  should be citing).
- M2 and M5 I did not re-derive — M1/M3/M4 plus my own M6/M7 and the 9600-case sweep gave me the
  directions I needed.

### Verdict

The fix is **sound** (9600-case sweep: it can never suppress a real binding), the guard is the whole
delta, the grading is unusually strong, and the hold-out control reproduces exactly. What is wrong is
the **honesty surface**: the stated residual is the one shape that has no phantom, eight shapes that
do have one go unnamed, the two heads the issue itself asked about are pinned with the modifier axis
removed, and the digest paragraph measures the wrong fixture.

**REQUEST-CHANGES**

1. **Fix the residual statement** in the PR body *and* in the `activePatternLetOffsets` doc comment.
   `and`-continued heads mint **no** phantom (`letRE` does not reach them) — measured; delete the
   claim that they "keep their phantom". Replace it with the residual that is real: heads where
   `letRE` matches but `activePatternRE` does not — whitespace-bearing banana clips
   (`(| Even | Odd |)`, `(|Even|Odd| )`, `( |Even|Odd|)`) and non-clip parenthesised/pattern heads
   (`(x, y)`, `(+.) a b`, `{ … }`, `[ … ]`, `(Some v)`), each still minting the last modifier as an
   operation. Say it is pre-existing and unchanged by this PR (I measured that).
2. **Add the modifier axis to `TestActivePatternLetPhantom_PlainLetStillExtracted`** — or, if those
   rows must stay green, add an explicitly-named residual/xfail-style row set recording the measured
   census `"private,sentinelValue"` for `let private (x, y)` and `let private (+.) a b`, so the next
   reader cannot mistake the no-modifier rows for coverage of the modifier case. Leaving the two
   heads the issue named pinned only *without* a modifier is the gap that let me land finding 2.
3. **Correct the digest paragraph**: the digest fixture is `writeSpanFixture6118` (18 members, listed
   above, zero F#), not `writeDjangoFixture` (10). Withdraw the "correction to my brief" — 18
   reproduces. The "regression cover, not coverage" conclusion survives unchanged.
4. **File the two new findings as their own issues** (they are not this PR's to fix):
   (a) `activePatternRE`'s clip class `[A-Za-z0-9_'|]+` rejects whitespace, so a whitespace-bearing
   clip loses the `SCOPE.Pattern` *and* keeps the #7163 phantom — **both halves** wrong on one input;
   note the unreachable `strings.TrimSpace(c)` at `compexpr_active_patterns.go:312` as evidence of
   intended tolerance, and that legality is derived-not-executed;
   (b) the modifier-named phantom on non-active-pattern parenthesised/pattern `let` heads
   (`let private (x, y)`, `let private (+.) a b`, record/list/union patterns) — same `letRE`
   backtrack, outside approach (b)'s reach by construction.
5. **Optional but cheap, and I would take it**: promote the soundness property to a test. An
   internal-package assertion that *every* `letRE` match at a claimed offset captured a modifier word,
   over a generated head/indent/separator/EOL cross product, pins the offset premise in the
   **over-firing** direction (M3 only pins the under-firing one, and M2 pins it only coarsely). My
   sweep is 9600 cases, 0 violations, and runs in well under a second.

Recording for the record, not as a change request: M7 (drop only the `CONTAINS` edges) is DEAD by
the package but **ALIVE w.r.t. all 150 of this PR's subtests** — these rows grade the pattern and its
case entities, not the edge.

PR #7172, opened 2026-09-17
test(#7167): two more backoff growth factors were ungraded — and at these sites no test killed the extreme either

### LINKED ISSUE #7167: PAIR AUDIT (both ALIVE): two more ungraded backoff growth factors in internal/daemon — rebuildBackoff (requests_drain.go:37) and acceptLoop's transient path (server.go:979) both survive '*= 3' at 0 --- FAIL
## What

Two more backoff growth-factor sites in `internal/daemon/`, both **ALIVE + ungraded**. `*= 2` → `*= 3` at either site leaves the full `./internal/daemon/` package green: **0 `--- FAIL`**.

```
internal/daemon/requests_drain.go:37   rebuildBackoff   d *= 2
internal/daemon/server.go:979          acceptLoop, transient-error path   backoff *= 2
```

Measured by the independent adversarial reviewer on #7165 as a **pair audit** — i.e. after grading the factor at the supervisor site, they went looking for the same expression elsewhere rather than stopping at the one the issue named. That is the right instinct and it paid immediately.

## Why this is the same defect and not a new one

#7162 established that the supervisor's growth factor was ungraded because the nearest pin (#7123's shrink arm) asserts a **floor** on one wait — *"growth happened"* — and not a **ratio** between consecutive waits — *"growth is by two"*. A tripled 400ms is 1.2s, which clears a `>=600ms` floor comfortably.

These two sites have the same shape and, as far as the suite is concerned, the same blindness. Whether they have a floor-style pin at all is **not established here** — the only measurement is that `*= 3` survives at each.

## Not claimed

- **No corpus or runtime measurement of consequence.** A wrong factor changes the shape of a retry curve, not whether the loop terminates. For the supervisor site the cost was measured at 1.39x time-to-give-up at shipped tuning; **nothing equivalent is measured here**, and the two sites have different termination conditions, so do not carry that figure across.
- **Specifically NOT claimed to be unbounded.** That adjective was wrong when #7126 used it for the clamp and wrong when I repeated it in a brief: `maxCeilingHits` ends the supervisor loop, making it a bound violation (32s/64s waits against a 30s counted ceiling), not an unbounded tail. Whether *these* loops have a terminating counter is an open question — **establish it before describing the consequence**, and verify the adjective, not just the number.
- **Not claimed the fix is a copy of #7165's.** `rebuildBackoff` and `acceptLoop` may have no announced-wait seam to observe. #7165 could assert a ratio deterministically only because the supervisor **announces** each wait; if these sites just sleep, the observable may not exist and the honest answer could be "not gradable without a seam" rather than a test.

## Grading requirement, when someone takes it

- **Assert the RATIO between consecutive waits, not a floor on one of them.** A floor is precisely what failed at the supervisor site.
- **Do not assert a literal sequence** — that encodes the initial value, the cap and the factor together, and breaks on any legitimate change to the other two.
- **Score both directions, and split enlargement by magnitude.** The lesson from #7165: *large* enlargement (`*= 100`) was already caught at the supervisor site by a pre-existing 30s-bound test, while *small* enlargement (`*= 3`, `*= 4`) was invisible. I had written that "enlargement is the direction nothing sees" and the reviewer measured that false. Check both magnitudes at each site rather than inheriting either verdict.
- **No timing dependence.** `-race` shifts this package's timings roughly 7x and the automatic CI legs run without it, so a wall-clock assertion here is scheduling luck. Find the seam or report that there is none.
- **Include a hold-out positive control**: apply `*= 3` with your new test file removed and show ALIVE at 0, with positive proof of the hold-out (`go test -list` counts and zero `=== RUN` lines for the held-out names). Without it a DEAD verdict may be inherited from a pre-existing test.

## The generalisable point

`*backoff *= 2` / `if *backoff > max` turned out to carry **three** independent decisions — threshold (#7111/#7123), clamp (#7126, shipped `45f9745d0`), factor (#7162, #7165) — each found only after the previous was pinned well enough for the next gap to show. This issue extends that from *roles on one expression* to *occurrences of one expression across a package*: a verdict at one site says nothing about its twins, and the twins here were found by grepping for the shape rather than by reading the issue.

Related: #7162 and #7165 (the supervisor factor), #7126 (the clamp), #7111/#7123 (the threshold), #7113 (the same one-symbol-two-roles shape in java receiver typing).

cajasmota:
Closes #7167

Two more occurrences of `*= 2` in `internal/daemon/` carried an **ungraded growth factor**. Both are now pinned by a RATIO between consecutive waits. Both sites were measured **independently** — a verdict at one says nothing about the other — and they came out differently from the supervisor site in a way worth recording.

| | site A | site B |
|---|---|---|
| where | `internal/daemon/requests_drain.go:37` — `rebuildBackoff`, `d *= 2` | `internal/daemon/server.go:986` — `acceptLoop` transient-Accept path, `backoff *= 2` |
| seam | **yes** — `rebuildBackoff` is a pure function of `attempts`; the test calls it directly. No clock, no goroutine, no process. | **yes** — `logger.Warn("accept: transient error, backing off", …, "retry_in", backoff)` announces the very local about to be slept, one line before `time.Sleep`. |
| why nothing saw it | **one test calls it; no test observes it.** Corrected from an earlier "no observer at all", which was inferred from a grep and is wrong. **Three** injections in **three** tests replace the field with a constant — `requests_drain_unblock_test.go:167` (`return time.Minute`), `:230` and `:417` (`return 0`) — not four. And `TestDrain_KindRebuild_CrashLoopIsBounded` (`requests_drain_resume_test.go:33`) **does** reach the shipped function: `drainRequestsOnce` (`requests_drain.go:236`) builds a fresh `requestsDrainer` -> `newRebuildWorker` (`:172`) with the default `backoff: rebuildBackoff` (`:280`), and its panicking `rebuildFn` drives the `OutcomeCrashed` branch calling `w.backoff(rec.Attempts+1)` (`:392`). It cannot **observe** the factor because each of its 8 drain passes gets a fresh worker, so the `nextAttempt` gate map (`:265`) written at `:392` never survives to the gate read (`:364`). | a **floor on a total, on a wall clock**: `TestAcceptLoopSurvivesTransientErrors` asserts `elapsed >= 10ms` after two transient errors. Under tripling those cost 5ms+15ms=20ms — enlargement makes that floor *easier* to clear, so it can never fail on this direction. |
| new pin | `TestRebuildBackoff_GrowsByExactlyTwoPerAttempt` | `TestAcceptLoopBackoffGrowsByExactlyTwoPerStep` |

Neither test reads a wall clock. `-race` shifts this package's timings ~7x and the automatic CI legs run without it, so a timing window would be scheduling luck (#7062).

## Verdicts, per site, kept separate

Both sites: **ALIVE + ungraded**, re-derived on `f9e7838e9` before any test existed. Baseline 0 `--- FAIL`.

**The band in which a mutant survives is every magnitude, in both directions, at both sites** — measured, not assumed:

| pre-fix | `*= 1` | `*3/2` (non-integer) | `*= 3` | `*= 8` | `*= 100` |
|---|---|---|---|---|---|
| site A | ALIVE 0 | ALIVE 0 | ALIVE 0 | ALIVE 0 | ALIVE 0 |
| site B | ALIVE 0 | ALIVE 0 | ALIVE 0 | ALIVE 0 | ALIVE 0 |

The **non-integer** row is what makes this a *band* rather than a set of sampled points, and it was missing from the first version of this table. `*= 2` is integer-nanosecond arithmetic — precisely where a rounding artefact could park a step back onto an exact doubling and make a continuous claim false *between* the integers. Measured (`d = d * 3 / 2`, `backoff = backoff * 3 / 2`, both new files held aside, full `./internal/daemon/` per row): ALIVE at 0 `--- FAIL` at **both** sites, and a ratio-kill post-fix at both. So the band closes continuously.

This is the one place these sites **diverge from #7165**, and the divergence is structural rather than lucky. At the supervisor site `*= 100` was already DEAD, caught by an unrelated 30s bound, so detectability there was non-monotone in magnitude. Here **no test kills the extreme at either site**: site A's one caller cannot observe the value it computes, so there is no bound for a large factor to disturb, and at site B the clamp holds every wait at or below 1s, so even `*= 100` costs the six scripted errors only ~4.5s and no leg's timeout moves. So this is a *wider* hole than the supervisor's, not a narrower one — and there is no "the extreme was already covered by a bound" row to report, because there isn't one.

With the new files, **every row is exactly 1 `--- FAIL` and it is the new test**; nothing else in the package reacts at any magnitude.

| site A, with pin | killer | announced sequence |
|---|---|---|
| `*= 1` | ratio | `[30s 30s 30s 30s 30s 30s]` |
| `*3/2` | ratio (5 errors), floor **not** fired — 5 of 5 pairs graded | `[30s 45s 1m7.5s 1m41.25s 2m31.875s 3m47.8125s]` |
| `*= 3` | ratio | `[30s 1m30s 4m30s 5m0s 5m0s 5m0s]` |
| `*= 4` | ratio **and** floor (1 graded pair) | `[30s 2m0s 5m0s 5m0s 5m0s 5m0s]` |
| `*= 8` | ratio **and** floor (1 graded pair) | `[30s 4m0s 5m0s 5m0s 5m0s 5m0s]` |
| `*= 100` | **floor alone** (0 graded pairs) | `[30s 5m0s 5m0s 5m0s 5m0s 5m0s]` |

| site B, with pin | killer | announced sequence |
|---|---|---|
| `*= 1` | ratio | `[5ms 5ms 5ms 5ms 5ms 5ms]` |
| `*3/2` | ratio (5 errors), floor **not** fired — 5 of 5 pairs graded | `[5ms 7.5ms 11.25ms 16.875ms 25.3125ms 37.96875ms]` |
| `*= 3` | ratio | `[5ms 15ms 45ms 135ms 405ms 1s]` |
| `*= 4` | ratio | `[5ms 20ms 80ms 320ms 1s 1s]` |
| `*= 8` | ratio **and** floor (2 graded pairs) | `[5ms 40ms 320ms 1s 1s 1s]` |
| `*= 100` | ratio **and** floor (1 graded pair) | `[5ms 500ms 1s 1s 1s 1s]` |

The `*= 100` row at site A is positive proof the vacuity floor is **live rather than decorative**: the first step clamps, every pair is skipped, and without the floor the file would pass a factor of 100 having asserted nothing.

**Both** vacuity floors now have a named killer, which matters because a floor with none is indistinguishable from a dead guard. Site A's is the `*= 100` row above, on the factor axis. Site B's is on the **start-value** axis instead — no site-B *factor* row kills on the floor alone, but mutating `acceptBackoffStart` 5ms -> 900ms in production (both new files present) yields exactly **1 `--- FAIL`**, the new test, on the **floor alone** with **0** ratio errors: `only 0 of 5 consecutive-backoff ratios sat below the 1s ceiling (sequence [900ms 1s 1s 1s 1s 1s])`. That also shows the guard tracks the sequence production actually emits rather than being satisfied by the constant it names. Recorded in `acceptloop_backoff_factor_test.go`; the `acceptBackoffStart` value being otherwise ungraded is **not** this PR's arm — see #7174.

`*= 1` is **not a confound at either site** — 1 `--- FAIL` each — which is worth recording because it *was* one at the supervisor site (five unrelated tests reacted). At site B, `TestAcceptLoopSurvivesTransientErrors`' 10ms floor does not fire on it either: six ungrown 5ms waits still total 30ms.

## Guard direction — the property #7162 established, transferred and re-scored

A `cur >= max { continue }` guard must be load-bearing in **both** directions: a guard that can silently empty its own input set grades nothing. That was never carried across to these two sites. Scored one occurrence at a time, full `./internal/daemon/` per row:

| row | site | mutation | before the sub-probe (`62602d1b9`) | after |
|---|---|---|---|---|
| EV-1 | A | `if true \|\| cur >= rebuildBackoffMax` | DEAD 1 — graded-floor | DEAD 1 — graded-floor, `only 0 of 5 … want at least 2`, `[30s 1m0s 2m0s 4m0s 5m0s 5m0s]` |
| EV-2 | A | `if false && cur >= rebuildBackoffMax` | DEAD 1 — ratio | DEAD 1 — ratio, 2 errors, exactly the two pairs at the 5m ceiling |
| EV-3 | B | `if true \|\| cur >= acceptBackoffMax` | DEAD 1 — graded-floor | DEAD **2** — graded-floor in *both* site-B tests (`only 0 of 5`, `only 0 of 9`): one guard, two callers |
| EV-4 | B | `if false && cur >= acceptBackoffMax` | **ALIVE, 0** | **DEAD 1** — killed twice over: the new excluded-floor (`no consecutive pair reached the 1s ceiling (… 9 graded / 0 excluded)`) **and** the ratio (2 errors) |

**EV-4 was ALIVE + masked by the fixture, not equivalent** — calling it equivalent would retire a real hole permanently. Site B's probe scripts six errors, announcing `[5ms 10ms 20ms 40ms 80ms 160ms]` against a **1s** ceiling, so `cur >= acceptBackoffMax` is false at every pair and deleting the branch cannot change the graded set. The guard would matter the moment a sequence reached the ceiling.

**The two sites are not symmetric, and the prose no longer says they are.** Site A's shipped walk `[30s 1m0s 2m0s 4m0s 5m0s 5m0s]` reaches its 5m ceiling twice, so both directions of *its* guard are graded by the natural fixture with no extra work. Site B's shipped start (5ms) is 200x below its ceiling (1s), so it needs a longer, ceiling-crossing script before the guard is exercised at all. Both files now state which case they are.

**The new sub-probe**, `TestAcceptLoopBackoffClampSkipExcludesOnlyCeilingPairs`: ten scripted transient errors announcing `[5ms 10ms 20ms 40ms 80ms 160ms 320ms 640ms 1s 1s]` — nine pairs, **seven graded / two excluded**. Those counts are documented, **not asserted** (asserting 7/2 encodes start, ceiling and factor at once — the literal-sequence shape this PR avoids). What is asserted survives a retune: the excluded set must be **non-empty**, and the graded set must still clear `acceptFactorMinRatios`. Real sleeps ~3.3s; no assertion reads a clock.

Raising `acceptBackoffStart` to 100ms was the obvious alternative and was **rejected on two checked counts**: `acceptBackoffStart` is a **`const`**, so there is no test seam for it and using it would mean a production `const`->`var` change purely to serve a test — the earlier 900ms floor-control row was a production *source mutation*, not an injection; and its split is 3 graded / 2 excluded, sitting **exactly** on the floor of 3 with zero margin. Scripting more errors keeps the **shipped** start value and needs no production change.

The clamp-skip guard is also collapsed to a **single occurrence** (`gradeAcceptBackoffRatios`). It was about to exist twice, once per probe, and a guard with two copies is graded at whichever copy a mutant lands on. Each caller keeps its own floor, so the extraction did not free either from asserting.

This grades **this file's** exclusion rule, not the **production** clamp in `server.go` — that one stays ungraded and is #7174. The two are explicitly distinguished in the file.

## Hold-out positive controls

- site A: `d *= 3` with `requests_drain_backoff_factor_test.go` **moved aside** → **0 `--- FAIL`**.
- site B: `backoff *= 3` with `acceptloop_backoff_factor_test.go` **moved aside** → **0 `--- FAIL`**.
- Positive proof of each hold-out: `go test -list` counts **487** with both files, **486** with either one moved aside, and the moved file's test name is absent from that 486 (the one remaining match is the *other* new test).

So both kills are earned by the new files, not inherited.

## Consequence, with the adjective verified rather than borrowed

The supervisor site's **1.39x time-to-give-up is deliberately NOT carried across** — the two loops have different termination conditions:

- **Site A is bounded by a counter.** `ApplyAndAckBounded` dead-letters the survivor once `rec.Attempts` reaches `maxRebuildAttempts` (= 3), so a doomed rebuild takes at most two backoff waits. The factor changes the *width* of that fixed-length window, not whether it closes. From the shipped constants: `rebuildBackoff(2) + rebuildBackoff(3)` = 60s + 120s = **180s** at factor 2 versus 90s + 270s = **360s** at factor 3 — a 2.0x stretch of the span a crash-looping rebuild occupies its group's single in-flight slot. `rebuildBackoffMax` (5m) is never reached in production at all, because attempt 4 dead-letters instead of waiting.
- **Site B's loop has NO terminating counter** — it retries a transient Accept error forever, by design (returning makes `Run` unlink the socket and drop every MCP client). But the **wait** is clamped at `acceptBackoffMax`, so a wrong factor is still a shape change and **not an unbounded tail**: doubling from 5ms takes 8 growth steps to reach 1s and retries 9 times in the first ~1.3s, where tripling takes 5 steps and 6 retries. An EMFILE burst that clears in a few hundred ms is re-attempted fewer times before the daemon drops to once-a-second accepts.

Neither is a correctness defect; both are tuning regressions. No runtime or corpus measurement of either is claimed beyond the arithmetic above, which is read off the shipped constants.

## What the assertions are, and what they are not

A RATIO between consecutive waits, applied only where the clamp did not engage, plus a vacuity floor on how many pairs were actually graded. **Not** a literal sequence (that encodes the initial value, the cap and the factor at once and breaks on any legitimate retune of the other two). **Not** a floor on one wait — that is precisely the shape that failed at the supervisor site and, measured above, at site B.

Each file states what it is blind to **on purpose**, so the next reader does not fill the gap with a vacuous fixture: the clamp, the first-wait branch, the reset-on-success, the `attempts < 1` normalisation, and the shipped values of the four constants. Deleting either clamp leaves every pre-clamp ratio at 2 and both new files green — the arms stay disjoint from #7126's and #7123's rather than overlapping them. Site B's seam also observes the **announced** wait, not the slept one; a defect that announced X and slept 4X passes there, which is inherent to the design and the right trade in this package.

## The one production change, and why

`backoffStart` / `backoffMax` were **function-local** consts inside `acceptLoop`, so a test could not name the ceiling its clamp-skip guard tests against. They are hoisted to package scope as `acceptBackoffStart` / `acceptBackoffMax`, **unchanged in value**, so the assertion can reference the ceiling instead of hard-coding `1 * time.Second` into it — putting an absolute duration back into the assertion is the thing these files exist to avoid. Pure scope change; no behaviour, no value, no control flow touched.

## Axes

**VARIED:** the growth factor (1, **1.5**, 3, 4, 8, 100 as well as 2 — both directions, enlargement split by magnitude, and one **non-integer** factor so the band is closed continuously rather than only at sampled integers), the site (two independent occurrences of the same expression, graded and reported separately), and the number of consecutive growth steps measured (three below the ceiling at site A, five at site B — versus the single wait/total the nearest pins look at). **HELD CONSTANT:** `rebuildBackoffBase`, `rebuildBackoffMax`, `acceptBackoffStart`, `acceptBackoffMax` (referenced, never asserted — no absolute duration appears in any assertion; `acceptBackoffStart` is moved in exactly one row, as the site-B floor's positive control, and that row is reported as a control rather than as an arm), the clamp comparisons, the transient-error classification (the scripted errors are the ones `TestIsTransientAcceptErr` already pins), and the reset-on-success behaviour (no successful Accept is scripted mid-burst — that is `TestAcceptLoopBackoffResets`' role).

## Not changed, with the reason

- **The shipped factor, at either site.** Both are correct at 2; the defect was that nothing said so.
- **The sibling pins.** `supervise_backoff_factor_test.go`, `supervise_backoff_clamp_test.go`, #7123's arms and the two pre-existing `acceptLoop` tests are neither re-tuned nor re-keyed. All are green in the unmutated run and in all 10 mutant rows (each row's single `--- FAIL` is the new test).
- **`TestAcceptLoopSurvivesTransientErrors`' `elapsed >= 10ms` floor.** It is blind to enlargement and it is left alone: it is the only thing in the package that observes a *really-slept* interval at this site, which is the shrink-direction cover the announced-wait seam cannot give.
- **`acceptLoop`'s CLAMP** (`if backoff > acceptBackoffMax { backoff = acceptBackoffMax }`). Deleting it leaves all 487 tests green, and #7126 pinned the **supervisor's** clamp, not this twin. That is a real pre-existing hole; it is **filed as #7174 and deliberately not fixed here** — widening this PR would merge two arms that are better graded separately.
- **`./cmd/grafel/`** was NOT run, and is not required: the diff stays inside `internal/daemon/` and touches no extractor, so the graph digest cannot move.

## Measurement hygiene

Full `./internal/daemon/` per row, `-count=1`, `-timeout 1800s`; verdicts are counted `--- FAIL` **lines**, never an exit code; `go vet ./internal/daemon/` exit code captured to its own file on every row (0 everywhere); every mutant applied with an asserted exact match count, the mutated line read back, and restored by `cp` + `shasum` + a zero-count marker grep. `gofmt -l` clean. Baseline and final unmutated tree: **0 `--- FAIL`**, **488** tests (487 plus the new ceiling-crossing sub-probe). The three review rows added on the follow-up push (`*3/2` pre-fix at both sites, `*3/2` post-fix at both sites, `acceptBackoffStart` 900ms) were measured the same way on a re-derived baseline, and every mutated file was restored by `cp` and verified byte-identical by `shasum` afterwards.


cajasmota:
## Independent adversarial review — #7172

Reviewer did not write this code. Own detached worktree at `origin/wt-7167-a1` (`64e743bc2`), **29 rows**, every row the full `./internal/daemon/` suite, `-count=1 -timeout 1800s`, verdicts are counted `--- FAIL` **lines** (never exit codes), `go vet` exit captured to its own file per row (0 on all 29), every mutant applied with an asserted exact match count and the mutated line **read back** from disk, restored by `cp` + `shasum` + a zero-count marker grep. Tree verified byte-identical to HEAD at the end (`shasum` on all three mutated files), `gofmt -l internal/daemon/` clean, `git status` clean. `./cmd/grafel/` not run and not required — the diff stays inside `internal/daemon/` and touches no extractor. One test binary at a time throughout.

**Every quantitative claim in the PR reproduced.** Two prose claims did not. Details below, measured separated from not-verified.

---

### 1. The production hoist (attacked first) — safe

- **Values unchanged**: `acceptBackoffStart = 5 * time.Millisecond`, `acceptBackoffMax = 1 * time.Second`, byte-identical to the function-local consts they replace.
- **No collision, no shadowing**: after the hoist there is no other package-scope `acceptBackoff*`, `backoffStart` or `backoffMax` identifier anywhere in `internal/` — the only other `backoffMax` occurrences are the supervisor's **struct field** `s.backoffMax` (`supervise.go:292`), which is a different namespace and is untouched. Nothing else in the package reads the new package consts except the new test.
- **Is the guard satisfied by the constant rather than by behaviour?** No — measured, not argued. I mutated `acceptBackoffStart` 5ms → 900ms (production only, both new test files present): **1 `--- FAIL`**, the new test, on the floor alone — `only 0 of 5 consecutive-backoff ratios sat below the 1s ceiling (sequence [900ms 1s 1s 1s 1s 1s])`. The guard tracks the sequence production actually emits, so it is not self-satisfying.
- Side observation from that same row: **nothing else in the package reacts to a 180x change in `acceptBackoffStart`** — that value is otherwise ungraded. Not this PR's arm; noted below.

### 2. "Every magnitude survives at both sites" — falsification attempt failed, and I widened the band

Pre-fix = both new test files **moved aside** (not edited), production hoist retained (inert).

| pre-fix, full package | `*= 1` | `*= 3` | `*= 3/2` (mine) | `*= 8` | `*= 100` |
|---|---|---|---|---|---|
| **site A** `requests_drain.go:37` | ALIVE 0 | ALIVE 0 | ALIVE 0 | ALIVE 0 | ALIVE 0 |
| **site B** `server.go:986` | ALIVE 0 | ALIVE 0 | ALIVE 0 | ALIVE 0 | ALIVE 0 |

10/10 ALIVE at 0 `--- FAIL`, each read back from disk before running. The non-integer row — the one that closed the band continuously at the supervisor site — is ALIVE at **both** sites too, so the survival band is wider than the PR reports, not narrower. `*= 1` at site B specifically does **not** trip `TestAcceptLoopSurvivesTransientErrors`' `elapsed >= 10ms` floor, confirming that claim (note it is marginal by construction: two ungrown 5ms waits total exactly 10ms plus scheduling overhead — that floor is one scheduling hiccup from being a flake on that *hypothetical* tuning only, not on the shipped one).

### 3. Post-fix: 12 rows, each exactly 1 `--- FAIL`, and the sites are disjoint

Every row below is the full package with **both** new files present. In every row the single failing test is the one belonging to the mutated site — a site-A mutant never moves the site-B test and vice versa, so the "two independent verdicts" framing is earned and nothing is averaged.

| with pins | site A killer (graded pairs) | site B killer (graded pairs) |
|---|---|---|
| `*= 1` | ratio (5 errors) | ratio (5 errors) |
| `*= 3/2` (mine) | ratio (5 errors) | ratio (5 errors) |
| `*= 3` | ratio (2 errors) | ratio (4 errors) |
| `*= 4` | ratio **+ floor**, 1 of 5 graded | ratio only, 3 graded |
| `*= 8` | ratio **+ floor**, 1 of 5 graded | ratio **+ floor**, 2 of 5 graded |
| `*= 100` | **floor alone**, 0 of 5 graded | ratio **+ floor**, 1 of 5 graded |

Sub-counts read out of the failure text, not inferred, and they add up to the PR's decomposition exactly — including `[30s 5m0s 5m0s 5m0s 5m0s 5m0s]` at site A `*= 100` with 0 graded pairs.

**Attack 6 — is the vacuity floor live at site B too, or only at A?** On the *factor* axis it is live only at A: no site-B factor row is killed by the floor alone, exactly as the PR states. But the site-B floor is **not decorative** — the `acceptBackoffStart` 900ms row in §1 kills on that floor alone with 0 graded pairs. So both floors have a positive control; A's is on the factor axis, B's on the start-value axis. That row is worth adding to the file.

### 4. Hold-outs — re-derived independently, not reused

- `go test -list '.*'`, counted myself: **487** both files present · **486** with `requests_drain_backoff_factor_test.go` aside (`TestRebuildBackoff_GrowsByExactlyTwoPerAttempt` absent, sibling still listed) · **486** with `acceptloop_backoff_factor_test.go` aside (`TestAcceptLoopBackoffGrowsByExactlyTwoPerStep` absent, sibling still listed) · **485** with both aside. The PR's 487/486 figures reproduce, and the 485 row confirms the two files contribute one test each rather than one file contributing both.
- `d *= 3` with **only** site A's file aside (site B's file present): **0 `--- FAIL`**.
- `backoff *= 3` with **only** site B's file aside (site A's file present): **0 `--- FAIL`**.

Both kills are earned by their own file. Neither is inherited, and neither new file covers the other's site.

### 5. Sibling pins — disjoint in both directions

`*backoff *= 2` → `*= 3` in `supervise.go:664`, both new files present: **1 `--- FAIL`**, `TestEngineSupervisor_BackoffGrowsByExactlyTwoPerStep` and nothing else. The two new tests are silent on the supervisor's mutant, and (§3) the supervisor's pins are silent on these two sites' mutants. No single test kills factors at two sites; #7162's, #7126's and #7123's arms are green in all 29 rows.

### 6. Site B's log seam — vacuity attacked, limit confirmed and correctly disclosed

The seam reads the **announced** value (`retry_in=` out of a captured `slog` text stream), so a defect that announces X and sleeps 4X passes. The file says so in plain terms, which is the right disclosure and the right trade here; the shrink direction keeps its weak wall-clock cover from `TestAcceptLoopSurvivesTransientErrors`. The parse is guarded (`time.ParseDuration` failure is fatal with the logs attached), and a wrong/absent format would fail the `len(waits) != n` positive control rather than pass vacuously — I did not need to take this on faith: the 900ms row above changed the announced sequence and the assertions moved with it.

### 7. Clamp-deletion rows (disjointness claim, plus one finding that is NOT this PR's)

- **Site A**, in-loop `if d >= rebuildBackoffMax { return }` deleted: **0 `--- FAIL`** — but this is ALIVE + **plausibly equivalent**, not a hole: the trailing `if d > rebuildBackoffMax` still clamps, and for any non-decreasing growth the two forms return the same value. I did not enumerate this exhaustively, so I record it as *equivalent-by-algebra, not enumerated*.
- **Site B**, the whole `if backoff > acceptBackoffMax { backoff = acceptBackoffMax }` deleted: **0 `--- FAIL`**. This one is a real behavioural change (an unbounded 5ms·2^n accept wait under sustained fd pressure) and **nothing in the package catches it**. The PR is right that its file is blind by design and right that this keeps the arm disjoint — but no *other* arm covers it either: #7126 graded the **supervisor's** clamp, not `acceptLoop`'s. That is a pre-existing hole, out of scope here, and should be its own issue rather than a change to this PR.

---

## What I could NOT verify

- No runtime/corpus measurement of either consequence claim (180s→360s at site A; 9-vs-6 retries in the first ~1.3s at site B). Both are arithmetic off the shipped constants and the PR says exactly that; I re-derived the arithmetic and it is right, nothing more.
- Windows/Linux behaviour of the site-B fixture. Measured on macOS only, without `-race`. Nothing in either test reads a clock, so I expect no platform sensitivity, but I did not execute it elsewhere.
- Whether `clampAdel` is equivalent on *every* input — argued from the algebra, not enumerated.

## Two prose claims that did not survive

**(a) "No observer at all" is over-stated, and the count is wrong.** `requests_drain_backoff_factor_test.go:26-33` says *"every test in the package that exercises that path REPLACES the field with a constant of its own (… three times, and `return time.Minute` once)"*, and the PR body says *"all four tests on that path replace the field with a constant"*. Measured: there are **three** injections in **three** tests — `return 0` at `requests_drain_unblock_test.go:230` and `:417`, `return time.Minute` at `:167` — not four.

More importantly, **one test does exercise the shipped `rebuildBackoff`**: `TestDrain_KindRebuild_CrashLoopIsBounded` (`requests_drain_resume_test.go:33`) calls `drainRequestsOnce`, which builds a fresh `requestsDrainer` → `newRebuildWorker(…)` carrying the default `backoff: rebuildBackoff`, and its panicking `rebuildFn` drives the `OutcomeCrashed` branch that calls `w.backoff(rec.Attempts+1)` (`requests_drain.go:392`). It cannot *observe* the factor because each of its 8 passes gets a **new** drainer, so the in-memory `nextAttempt` gate map never survives to the pass that would be gated — which is why every magnitude is ALIVE. The grep sentence in the same paragraph ("grepping the package for `rebuildBackoff` outside its own declaration returns the single wiring line") is true; the conclusion drawn from it — that no test reaches the function — is not. The accurate statement is *one test calls it, no test observes it*, and the reason is worth recording because it is the actual mechanism.

**(b) The magnitude table is incomplete in a way the PR itself argues matters.** The non-integer row (`d = d*3/2`, `backoff = backoff*3/2`) is the row that proved the band closes *continuously* at the supervisor site. It is ALIVE pre-fix and dies on the ratio post-fix at **both** sites (§2, §3). Since the PR's central negative claim is about the whole band, that row belongs in the table.

Neither of these changes a verdict. Both are the repo's dominant defect class — prose asserting a mechanism nothing measured — in a file whose entire value is that its prose is exact.

---

## REQUEST-CHANGES

1. **Correct the site-A mechanism claim** in `requests_drain_backoff_factor_test.go` ("WHY NOTHING SAW IT") and the matching PR-body cell: it is **three** injections in **three** tests, not four, and `TestDrain_KindRebuild_CrashLoopIsBounded` **does** call the shipped `rebuildBackoff` through `drainRequestsOnce`'s fresh `requestsDrainer`. Replace "no observer at all" with the measured statement: one test calls it; no test observes it, because each drain pass gets a fresh worker whose `nextAttempt` gate map never survives.
2. **Add the non-integer magnitude** (`×3/2`) to both files' MEASURED blocks and to the PR-body table, at both sites: ALIVE 0 pre-fix, ratio-kill post-fix (site A `[30s 45s 1m7.5s …]`, site B `[5ms 7.5ms 11.25ms …]`). The band claim is the PR's headline; it should carry the row that closes it continuously.
3. **Record the site-B floor's positive control.** `acceptloop_backoff_factor_test.go` currently says no row is killed by the floor alone, which reads as "this floor is never exercised". It is exercised — on the *start-value* axis, not the factor axis: `acceptBackoffStart` 5ms → 900ms yields exactly 1 `--- FAIL`, the floor alone, `0 of 5` graded (sequence `[900ms 1s 1s 1s 1s 1s]`). One sentence.
4. **File the `acceptLoop` clamp as its own issue** (do not widen this PR): deleting `if backoff > acceptBackoffMax { backoff = acceptBackoffMax }` leaves all 487 tests green. #7126 covered the supervisor's clamp only. Same issue can note that `acceptBackoffStart`'s value is graded by nothing outside the new file's floor.

Everything else measured — the two seams, all 10 pre-fix rows, all 12 post-fix rows and their per-assertion decomposition, both hold-outs with my own `-list` counts, the cross-site and cross-sibling disjointness, and the production hoist — reproduces exactly as claimed. The pins themselves are sound and I would approve them on the next push with the four items above addressed.

PR #7173, opened 2026-09-17
fix(#7166) arm 1: `and`-continued active patterns were never extracted — activePatternRE now admits the keyword F# uses to continue a recursive group

cajasmota:
## What

`activePatternRE` was anchored `(?m)^([ \t]*)let…`, which requires `let` to be the first
non-whitespace token of the line. F# continues a recursive binding group with `and`, so

```fsharp
let rec (|Even|Odd|) x = …
and (|Positive|Negative|) y = …
```

produced **one** `SCOPE.Pattern`, not two — and a missed definition takes its whole case set with
it (it is the sole producer of the `active_pattern_case` sub-entities), so every match arm
`| Positive ->` in the file then resolves to nothing.

The anchor now admits a second head keyword. `Signature` names the keyword the declaration
actually carries (`and (|A|B|)`), read back off the matched text so no capture index moves.

### Why the alternation repeats the modifier group instead of factoring it out

The obvious shape is `(?:let|and)` + one modifier group. It cannot be used here.
`TestActivePatternREModifierParityWithLetRE` locates the shared prefix by the literal text
`let(?:\s+(?:…)\b)*` **in the compiled pattern**; `(?:let|and)(?:\s+…` separates `let` from its
group, `fsLetModifierGroupRE.FindString` returns `""`, and the test's own vacuity guard
`t.Fatal`s. **Verified, not assumed** — this was measured on the compiled `String()`, which is
also the answer to the brief's question: adding `and` to the keyword alternation *is* part of the
pinned prefix if it is placed in front of the group, and is *not* if each branch keeps its own
copy. So each branch carries the group, both from a new `fsLetModifiers` constant — one place to
edit, and the parity test stays **green** (it is in the package suite below, 0 `--- FAIL`).

## Axes

**Varied:** head keyword (`let` vs `and`); modifier phrase on the `and` head (none / `private` /
`internal` / `public` / `inline` / `inline private` / `private inline` / `rec` / `mutable` —
identity, count 0–2, and order); chain length (2 members **and** 3, so "handles `and`" cannot be
confused with "handles exactly one `and`"); clip shape (two-case total and partial `(|On|_|)`);
indentation of the `and` head (column 0 and nested inside `module Nested`); and the surrounding
text of a non-declaration `and` (line comment, indented line comment, inline string literal,
mid-line, `and` as a word **prefix** (`andThen`), `and` as a word **suffix** (`band`)).

**Held constant, deliberately:** case names (`Even`/`Odd`, `Positive`/`Negative`, `On`, `Up`/`Down`,
`Left`/`Right`, and `Ghost`/`Ghoul`/`Cee`/… for every forbidden row, so each forbidden name is
**unique in its fixture** — the F# dedup keys are name-keyed, so a ghost that collides with a real
name merges silently, #7144/#7152); the real binding's name (`sentinelValue`); the enclosing
`module Patterns`; the file path; the Kinds under assertion.

## The widening direction

Adding a keyword is a widening, and `extractActivePatterns` is handed the **raw `src`** (#7152),
so the `^([ \t]*)` line anchor is the only thing bounding it. Graded two ways:

1. **Forbidden rows** (`…_NotAtLineStartMintsNothing`): six fixtures in which the text
   `and (|…|) … =` is present and must mint **nothing** — census must stay exactly
   `[(|Even|Odd|)]` with exactly its two cases. All six pass.
2. **Measured, and reported rather than hidden:** a line-start declaration inside a `(* … *)`
   block or a `"""…"""` literal **is** extracted today — for `let`, on `main`, before this change.
   That is #7152, it is open, and this commit extends it by exactly its `and` mirror: no new class.
   `…_ScrubResidualParityWithLet` therefore pins **let/and parity**, not correctness: it asserts
   the two keywords agree in that context, so a future #7152 fix that reaches both stays green
   while one that reaches only one keyword fails. Pinning the gap as correct is what #7135 arm 3
   declined to do, and this does not do it either. Refs #7152.

## Mutants

Every mutant: `cp` backup, exact-match-count assertion on apply, the mutated line **read back**
and printed, `go vet` exit captured to its own file (never through a pipe), scored by counting
`--- FAIL` lines, restored by `cp` + `shasum` + `grep -c MUTANT_MARKER == 0`. Script asserts its
own worktree path and aborts otherwise.

|---|---|---|---|---|
| M1 | keyword alternation reverted to `^([ \t]*)let` + group | 0 | **15** | **DEAD** |
| M2 | **positive control** — M1 with this PR's test file held out | 0 | **0** | **ALIVE** (as required) |
| M3 | `and` branch unanchored (`\|[ \t]*and`), `let` branch untouched | 0 | **6** | **DEAD** |
| M4 | the factored form `(?:let\|and)` + ONE modifier group | 0 | **1** | **DEAD** — and this is the parity finding, measured |

**M1 decomposition (15):** 4 parent failures (`_BothPatternsExtracted`, `_ModifierAxis`,
`_ThreeMemberChain`, `_ScrubResidualParityWithLet`) + 11 subtests (9 `_ModifierAxis` rows +
2 `_ScrubResidualParityWithLet` rows). 4 + 9 + 2 = **15** ✔

**M2 positive control, proof the hold-out landed:** `go test -list '.*'` went **162 → 157** listed
tests, the diff being exactly the five `TestActivePatternAndContinued_*` names; the verbose run has
**zero** `=== RUN   TestActivePatternAndContinued` lines (the M1 and M3 runs have 22 each), and the
package exits 0 with 0 `--- FAIL`. So M1's kill rests on this PR's rows and on nothing that already
existed.

**M4 (1), the answer to "is `and` part of the pinned prefix?":** it depends on where you put it,
and this was measured rather than argued. The factored form extracts identically — every
`TestActivePatternAndContinued_*` row and the whole rest of the package stay green — but
`TestActivePatternREModifierParityWithLetRE` fails with *"could not locate a `let` +
repeated-modifier-group prefix in its compiled pattern, so the parity comparison below would be
vacuous"*. That is its **vacuity guard**, not its parity comparison: the modifier group itself is
untouched, but `let` is no longer adjacent to it, so the comparison would have degenerated to
`"" == ""` and silently graded nothing. 1 `--- FAIL`, and it is the only one in the package. The
shipped form keeps `let` adjacent to its group and the test green — so **adding `and` to the
keyword alternation is NOT part of the pinned prefix when each branch carries its own copy of the
group, and IS fatal to it when the keywords are factored in front.**

**M3 decomposition (6):** 1 parent (`_NotAtLineStartMintsNothing`) + 5 of its 6 subtests —
`line_comment`, `indented_line_comment`, `inline_string_literal`, `mid-line`, `` `and` SUFFIX ``.
1 + 5 = **6** ✔. The sixth row (`andThen`, the `and` **prefix**) correctly survives: it is rejected
by the mandatory separator after the keyword, not by the anchor, so M3 does not reach it — that is
the row's point, and it is the one M3 could not have graded. **Confound check:** M3 leaves the
`let` branch byte-identical and no test outside this PR's file fails under it, so the 6 failures
are attributable to the `and` anchor alone.

## Gate

- `go test -count=1 ./internal/extractors/fsharp/` — **PASS, 0 `--- FAIL`** (162 tests), including
  the pre-existing `TestActivePatternLetPhantom_AndContinuedMintsNoPhantom` (#7169) and
  `TestActivePatternREModifierParityWithLetRE` (#7164), both untouched and green.
- `go test -count=1 ./cmd/grafel/` with a **fresh `GRAFEL_HOME` and `GRAFEL_DAEMON_ROOT`** —
  **PASS**, exit 0, 0 `--- FAIL`, 174.195s. The `writeSpanFixture6118` digest fixture has 18 members and **zero** `.fs`/
  `.fsx`, so an F#-only change is expected to leave the digest **unmoved**, and it did.

## Not fixed here

**Arm 2 of #7166 — the same-line `[<SomeAttr>] let (|Even|Odd|) x = …` case — is NOT fixed.**
The two arms share an anchor and nothing else. Arm 1 adds a keyword *at* the anchor; arm 2 requires
the `^[ \t]*` prefix itself to be loosened so it tolerates preceding non-whitespace tokens. On a
scanner that reads raw `src`, that is a strictly larger widening of the #7152 surface — every
`[<…>]`-prefixed or otherwise-prefixed line in a comment or literal becomes reachable — and it
deserves its own forbidden-row grading rather than riding along here. Arm 2 remains open.

Also unclaimed: **corpus incidence** (neither form counted), and **legality is derived, not
executed** — there is no F# toolchain in this environment (`dotnet`, `fsc`, `fsharpc`, `fsi`,
`mono` all absent; `javac` is the only compiler present). The `and` group form rests on the spec's
recursive `let rec function-defn (and function-defn)*` production, MS Learn "Functions" →
*Recursive Functions*, and MS Learn "Active Patterns" for an active pattern being such a
`function-defn`. Rows whose legality those sources do not settle (`and rec`, `and mutable`,
`and private inline`) are labelled **per row** as lenience, not left silent.

Refs #7166 — arm 1 only; arm 2 (same-line attribute) stays open. Refs #7152.

---

## Round 2 — both REQUEST-CHANGES items closed

**Item 1 — the `and` copy of the modifier group was ungraded in the WIDENING direction.** The
reviewer is right and the mechanism is the one I half-found: `fsLetModifierGroupRE` required a
literal `let`, and `FindString` returns the first match, so the `and` branch's copy was compared to
nothing. `TestActivePatternREModifierParityWithLetRE` now extracts **every** head-keyword
occurrence (`(?:let|and)` + a capture on the group alone, `FindAllStringSubmatch`), compares **each**
against letRE's, and carries a **count guard** (`len(aps) == 2`, `len(lts) == 1`) so a branch whose
group is deleted or factored away fails too rather than passing vacuously. The vacuity substring
guards now run per occurrence as well.

**Item 2 — the `let` value of the two-valued `Signature` was pinned by nothing.**
`_BothPatternsExtracted` now asserts the opening `let rec` member's
`Signature == "let (|Even|Odd|)"` (plus its `StartLine`), beside the existing `and` assertion — both
values of the branch, in one fixture.

**Re-scored, all four DEAD** (`cp` backup, exact-match-count on apply, mutated line read back and
printed, `go vet` exit to its own file, restore verified by `shasum` + marker count 0):

| # | mutant | vet | failing tests | verdict | killed by |
|---|---|---|---|---|---|
| MR1 | `sealed` added to the **`and`** branch only | 0 | **1** | **DEAD** | `TestActivePatternREModifierParityWithLetRE` |
| MR2 | the identical widening on the **`let`** branch | 0 | **1** | **DEAD** | same |
| MR3 | **narrow** the `and` branch (drop `private`) | 0 | **5** | **DEAD** | parity test **+** `_ModifierAxis` parent + `{private, inline_private, private_inline}` (1+1+3 = 5) |
| MR4 | `head := "and"` unconditionally | 0 | **1** | **DEAD** | `_BothPatternsExtracted` (the new `let`-side signature row) |

**A counting trap found while scoring, and fixed.** The parity test's own `t.Errorf` text contained
the literal failure marker `--- FAIL`, so a score that greps the log for that marker read **2** for
MR1/MR2 and **6** for MR3 — one phantom failure each, in the test whose job is to be the last line
of defence against an ungraded widening. The message now says "zero failing tests" and explains why;
after the edit MR1 re-scores at **1**, which is the number above. Comments in the file still contain
the marker and are harmless — they never reach a test log.

**What changed since `e1c1a4a91`.** Three files. `compexpr_active_patterns.go` changed in
**comments only** — `git diff -U0 e1c1a4a91 -- internal/extractors/fsharp/compexpr_active_patterns.go`
filtered to non-comment, non-blank changed lines is **empty**; the regex and the `head`/`Signature`
logic are byte-identical to the reviewed revision. The correction there is factual: the old comment
claimed the two occurrences were the same bytes "by construction rather than by review", which MR1
disproved — the guarantee is the all-occurrences comparison plus its count guard, and the constant is
only convenience. The other two files are the parity test and this PR's test file.

Also taken from section E: `_ScrubResidualParityWithLet` now `t.Log`s the observed pair, recording
that it is currently `true/true`, so a future #7152 fix that silences both makes the row's
degeneration into vacuity visible instead of silent. Section E4 (a label-derivation cross-check) is
not taken — it was a suggestion, the reviewer hand-checked all 9 labels as correct, and adding it
would widen a round that is meant to be cheap.

**Not fixed, not mine:** the newline-spanning `\s+` separator the reviewer found (a bare line-start
`let`/`and` with the clip on a later line still mints). Pre-existing on **both** keywords, covered by
no fixture, and **filed separately** by the coordinator.

Gates re-run on the new head: `./internal/extractors/fsharp/` PASS, 0 failing tests;
`./cmd/grafel/` with fresh `GRAFEL_HOME` + `GRAFEL_DAEMON_ROOT` **PASS**, exit 0, 0 failing tests, 165.0s — digest again UNMOVED, as an F#-only change must leave it (`writeSpanFixture6118` has 18 members, zero `.fs`/`.fsx`).

---

## Round 3 — one prose correction (no behaviour change)

The reviewer measured that the count guard's own comment was **false about a third head keyword**,
and the truth is inverted in both halves. `fsLetModifierGroupRE` is itself keyed on `(?:let|and)`,
so a `|use` branch is **never extracted** and the count simply stays 2: **MC2** (`|use` + the shared
constant) and **MC2b** (`|use` + a `sealed`-widened group) are **both ALIVE at 0 failing tests,
`go vet` 0**. So:

- `aps == 2` catches a **deleted or factored-away** `let`/`and` branch — that is what it is for, and
  the reviewer isolated it with a **behaviour-neutral respelling** mutant (`(?:[\s]+…)` ≡ `\s`) that
  moved no behavioural row and failed **only** the count guard;
- it does **not** fire on an **added third keyword**, because the extractor's own `(?:let|and)`
  keying never reaches it;
- so a third keyword's modifier group would be **ungraded** — the same literal-keying hole one level
  up, in the extractor's hand-picked keyword list. Not a regression (the `let`-only era had it too),
  review-visible to reach, **filed separately**, not fixed here. The comment records the cheapest
  keyword-agnostic close: `strings.Count(activePatternRE.String(), fsLetModifiers) == 2`, which
  counts the **constant** and so cannot be evaded by adding a keyword.

**Prose only.** `git diff 23c5033d2 215085bae` filtered to non-comment, non-blank changed lines is
**0 lines** — one comment block in the parity test, nothing else. fsharp suite: exit 0, anchored
`^\s*--- FAIL` count **0**.

**Round 4 — the same correction inside the guard's `t.Fatalf` message** (authorized after I flagged
it, since it is a non-comment line). The message a reader meets mid-debugging now says:

> "A branch that lost its group accepts no modifier at all. A third head keyword is NOT caught
> here — fsLetModifierGroupRE never extracts it; see the comment above."

The reported phrase **`activePatternRE yielded %d modifier groups` is kept byte-identical**, so every
attribution quoted above and in the review still points at live text — no attribution update was
needed. **MC1 re-scored** (behaviour-neutral respelling of the `and` group, `(?:[\s]+…)` for
`(?:\s+…)`): **DEAD, 1 anchored failure, vet 0**, reporting
`activePatternRE yielded 1 modifier groups [...]` exactly as before. The new text contains **no**
test-failure marker; all 11 remaining occurrences in the package are on `//` lines, none in a string
literal. fsharp suite: exit 0, anchored `^\s*--- FAIL` **0**.

Also filed separately and not fixed here: the newline-spanning `\s+` separator (#7175).


cajasmota:
## Independent adversarial review — #7173 (arm 1)

Own worktree at `origin/wt-7166-a1` (`e1c1a4a91`), detached, path-asserted in every script. Every
mutant: `cp` backup, exact-match-count on apply, mutated line **read back and printed**, `go vet`
exit captured to its own file, scored by counting `--- FAIL` lines, restored by `cp` + `shasum` +
marker-count 0. Final tree: `git status` clean, file shasum `16c4df9b…` identical to the untouched
checkout, package green.

The change is correct and the `and`-branch behaviour is well graded. I found **two ungraded
directions the PR itself introduces**, both cheap to close, both in the class this repo treats as
dominant. Verdict at the bottom.

---

### A. Claims I re-derived and CONFIRM

| claim | my measurement |
|---|---|
| `./internal/extractors/fsharp/` PASS, 0 `--- FAIL`, 162 tests | exit 0, `--- FAIL` count **0**, `-list` `^Test` count **162** ✔ |
| M1 (alternation reverted) = 15 | reproduced: **15**, decomposition **4 parents + 9 `ModifierAxis` + 2 `ScrubParity` = 15** ✔ |
| M3 (`and` branch unanchored) = 6, **and `andThen` survives because of the SEPARATOR, not the anchor** | reproduced with `(?m)(?:^([ \t]*)let…\|[ \t]*and…)`: **6** `--- FAIL`, and the surviving row is exactly `word_with_an_'and'_PREFIX`. The attribution is **right** — I attacked this specifically and it held ✔ |
| "no capture index moves" | both branches are `(?:…)`, so groups stay 1/2/3. **Measured, not read:** 5 param shapes × 3 modifier phrases × {`let`,`and`} — `StartLine`, `active_pattern_cases`, `parameterised`, `partial` are **identical** on both heads in all 15 pairs; only `Signature` differs ✔ |
| `./cmd/grafel/` PASS, digest UNMOVED | ran **once**, fresh `GRAFEL_HOME` + `GRAFEL_DAEMON_ROOT`: exit 0, `--- FAIL` **0**, 166.9s. Independently: `writeSpanFixture6118` has **18 members, zero `.fs`/`.fsx`** (7×cs, java, ts, js, 6×py, go, rs), so an F#-only change *cannot* move it — the claim is structurally sound, not just observed ✔ |
| the `and` offsets `activePatternLetOffsets` now claims are a harmless superset | **measured, not argued**: over 6 adversarial sources (incl. `let v = 3 / and w = 4`, an `and` head at offset 0, an indented `and` between two `let`s, a newline-spanning `and`) → **0 collisions** between `and`-head match starts and `letRE` match starts. Collision is structurally impossible (both anchor at line start; one line cannot begin with both keywords) ✔ |
| axes varied; 3-member chain | `_ThreeMemberChain` asserts an **exact 5-name census**, not a floor, and adds the partial clip + nested indentation. `_ModifierAxis` genuinely varies 9 phrases **on the `and` head** ✔ |
| grammar labels | I re-derived all 9 against `fsAPLabel`'s rules: all agree except `rec`, deliberately overridden to the new `lenienceAndRec` — and correctly, since `rec` belongs to the `let rec` that OPENS the group and is not in `function-defn`. **No mislabelling found** ✔ |

### B. My own alphabet enumeration (25 candidate lines, not hand-picked around the fixtures)

**Mints nothing** (rejected): `andThen`, `and_`, `and'`, `andré`, `AND`, `and(|…`, `land`, `Land`,
`andand`, `and and`, `and sealed`, `and`+NBSP (Go `\s` is `[\t\n\f\r ]`, no U+00A0), `and // c` +
newline + clip.

**Newly reaches the `and` branch**: `\tand`, `    and`, `and\t`, `and rec`, `and inline`,
`and private rec`, `and (|X|_|)`, plus two lenience forms the fixtures don't name —
`type Foo = int` / `and (|A|B|) …` (recursive **type** group continuation) and `let ok = x` /
`and (|A|B|) …` (continuation of a non-`rec` binding). Same lenience class as `and rec`; harmless.

One form no fixture covers: **the separator `\s+` spans newlines**, so `and` alone on a line with
the clip on a *later* line mints — `and\n    (|A|B|) z =`, `and\r\n(|A|B|) z =` and
`and\n\n\n(|A|B|) z =` all produce the pattern, `StartLine` on the `and` line. This is a true
`let` **mirror** (`let` has the same `\s+`), so I am **not** calling it a new class — but a bare
line-start `and` is far more common in real F# than a bare line-start `let`, so the mirror is
materially louder on this branch. Noted, not a blocker.

---

### C. Finding 1 — the `and` copy of the modifier group is ungraded in the WIDENING direction

This is the `sealed` hole that `TestActivePatternREModifierParityWithLetRE` exists to close,
re-opened on one branch of the new alternation.

| mutant | vet | `--- FAIL` | verdict |
|---|---|---|---|
| **MR1** — add `sealed` to the **`and`** branch only | 0 | **0** | **ALIVE + ungraded** |
| **MR2** — *positive control*, the identical widening on the **`let`** branch | 0 | **2** | DEAD (`…ParityWithLetRE`) |
| **MR3** — *direction control*, **narrow** the `and` branch (drop `private`) | 0 | **4** | DEAD (`_ModifierAxis/{private,inline_private,private_inline}`) |

So the `and` copy is graded **only** in the narrowing direction, by the behavioural rows. The
widening direction — the unbounded one — is closed on `let` and open on `and`.

Mechanism, and it is not the constant's fault: `fsLetModifierGroupRE` is
``let\(\?:\\s\+\(\?:[a-z|]+\)\\b\)\*`` and `FindString` returns the **first** match, which is always
the `let` branch. The `and` branch's copy is never compared to anything. The in-source comment says
the two occurrences "must be the same bytes as letRE's **by construction rather than by review**" —
that is true of today's source, but MR1 is precisely a review-visible edit that construction does
not prevent, and it is ALIVE at 0 `--- FAIL` with `go vet` 0.

Cheapest closes (either):
- assert `strings.Count(activePatternRE.String(), <fsLetModifiers text>) == 2` — pins the count too,
  so deleting one branch's group is also caught; or
- widen `fsLetModifierGroupRE` to `(?:let|and)\(\?:…` and use `FindAllString`, comparing **every**
  occurrence against `letRE`'s, with a `len(all) == 2` guard.

### D. Finding 2 — the `let` value of the new two-valued `Signature` is pinned by nothing

`Signature` was a constant `"let " + name` and is now keyword-dependent. Only one of its two values
is asserted.

| mutant | vet | `--- FAIL` | verdict |
|---|---|---|---|
| **MR4** — `head := "and"` unconditionally | 0 | **0** | **ALIVE + ungraded** |

Not equivalent — demonstrated on the emitted artefact, under the mutant, with the whole package
green:

```
SIG (|Even|Odd|)          => "and (|Even|Odd|)"      <- the `let rec` head, mis-signed
SIG (|Positive|Negative|) => "and (|Positive|Negative|)"
```

`grep 'Signature' internal/extractors/fsharp/*_test.go` returns exactly **one** assertion on a
`SCOPE.Pattern` signature in the entire package, and it is this PR's `and` row. `let (|Even|Odd|)`
is asserted nowhere, before or after. Score-every-occurrence: the PR introduced the branch and
graded one side of it. One line in `_BothPatternsExtracted` — assert the **opening** `let rec`
member's `Signature == "let (|Even|Odd|)"` — closes it and makes the row's own claim ("the emitted
signature must name the head keyword the declaration actually carries") a two-sided one.

### E. Minor, not blocking

3. `_ScrubResidualParityWithLet` has no "currently both mint" guard. It grades **today** (M1 kills
   both its rows, so `gotLet=true, gotAnd=true` is live), but it passes equally on `false/false`.
   That is deliberate and stated — the risk is that a future #7152 fix silencing both leaves it
   green *and vacuous* with nothing flagging the transition. A `t.Log` of the observed pair, or a
   one-line note recording that the pair is currently `true/true`, would make the degeneration
   visible to whoever reads the next failure.
4. Grammar labels here are read **only** from `t.Errorf` and so are never evaluated on a green run.
   I hand-checked all 9 and they are right, but `active_pattern_modifiers_7135_test.go` has
   `TestActivePatternModifiers_LabelDerivation` for exactly this reason (it is what caught the
   30-subtest mislabelling) and this file has no equivalent. Rows are few; a cross-check against
   `fsAPLabel` with an explicit `rec`→`lenienceAndRec` override would cost ~6 lines.

### F. What I could NOT verify

- **Legality.** No F# toolchain here (`dotnet`, `fsc`, `fsharpc`, `fsi`, `mono` all absent; `javac`
  is the only compiler). Every legality judgement above, mine and the PR's, is
  **derived-not-executed**. The PR is explicit about this and labels per row, which I endorse.
- **Corpus incidence.** Neither the PR nor I counted how often `let rec … and (|…|)` occurs in real
  F#. The widening's *value* is unmeasured; only its *safety* is.
- **Arm 2.** Correctly out of scope; I did not attack it.

---

**REQUEST-CHANGES**

1. Close the widening direction on the **`and`** copy of the modifier group (MR1: ALIVE, 0
   `--- FAIL`, vet 0, while the identical mutant on the `let` copy is DEAD at 2). Either pin
   `strings.Count(activePatternRE.String(), fsLetModifiers) == 2`, or make
   `fsLetModifierGroupRE` match both keywords and compare **all** occurrences with a count guard.
   Re-score MR1 after the change — it must go DEAD.
2. Assert the **`let`** value of `Signature` (MR4: ALIVE, 0 `--- FAIL`, vet 0, and demonstrated to
   mis-sign every `let`-headed pattern). One assertion on the opening `let rec` member in
   `_BothPatternsExtracted`. Re-score MR4 — it must go DEAD.

Items 3 and 4 in section E are suggestions, not conditions.

cajasmota:
## Re-verification of the parity rewrite — `23c5033d2`

Narrow round: only the parity test was attacked. Same discipline (backup, exact-match-count on apply, line read back, `go vet` to its own file, `cp` restore + shasum + marker 0). Scored with an **anchored** `grep -cE '^\s*--- FAIL'`, which message text cannot spoof. Baseline: exit 0, anchored 0, 162 tests. Tree clean afterwards.

**Marker.** `--- FAIL` appears 3× in the parity file and 8× elsewhere in the package — **all 11 on `//` comment lines, zero in any string literal**, so none can reach a log. On MR1 unanchored and anchored counts now agree (1 = 1). The counting trap is genuinely closed.

**MR1 re-score (my own, anchored): DEAD 1.** Killing assertion named: the per-occurrence byte comparison, reporting `occurrence 1 and letRE`. Twin control MR2b (`sealed` on the `let` branch): **DEAD 1**, reporting `occurrence 0`. Both copies are now compared.

**1. Count guard — load-bearing, and isolated.** MC1 respells the `and` branch's group `(?:[\s]+…)` — behaviourally **identical** (`[\s]` ≡ `\s`), so no behavioural row moves, but the extractor no longer sees that occurrence. Result: **exactly 1 anchored `--- FAIL`**, and the only reporting assertion is `activePatternRE yielded 1 modifier groups`. The byte comparison passes in that run, so the count guard is not riding on it. Deletion/factoring direction: confirmed.

**2. Per-occurrence vacuity — confirmed, both ways.** MC4 (factored `(?:let|and)` + one group) → result set **empty** → `fsLetModifierGroups` `t.Fatal`s with *"could not locate a head-keyword…"*: an empty set cannot walk vacuously. MC3 (`and` group narrowed to a single word) → the bar guard fires as **`extracted group 1`** alongside `occurrence 1 and letRE`, so the guards really do run per occurrence and not just on the first.

**3. One residual, NOT blocking, and it is the cost question you asked about — measured, and the answer is the opposite of expected.** `aps == 2` does **not** fire on a third head keyword. `fsLetModifierGroupRE` is keyed on `(?:let|and)`, so a `use` branch's group is simply not extracted and the count stays 2.

- **MC2** — add `|use` + `fsLetModifiers`: **ALIVE, 0 `--- FAIL`**, vet 0.
- **MC2b** — add `|use` + a group widened with `sealed`: **ALIVE, 0 `--- FAIL`**, vet 0.

So the guard costs nothing for a legitimate third keyword (good), but the in-source claim *"a third branch needs its own row here"* is measurably false: a third branch with a **divergent, widened** allowlist is ungraded — the same hole one level up, now in the extractor's own hand-picked keyword list. Not a regression (the `let`-only era had it too) and it needs a review-visible source edit to reach, so it does not block. Cheapest close, if wanted: assert `strings.Count(activePatternRE.String(), fsLetModifiers) == 2` against the **constant itself**, which is keyword-agnostic — or reword the comment to say what the guard actually covers.

**APPROVE.** Both items from my first review are closed and I verified the fixes rather than the claims. The only follow-up is the comment wording (or the keyword-agnostic count) in point 3, and it is fine as a separate change.

PR #7179, opened 2026-09-18
fix(#7161): a TYPE_USE annotation on an array dimension lost the space before it — the declarator was name + dimensions, glued with no separator

### LINKED ISSUE #7161: java: a TYPE_USE annotation on an array dimension loses its leading space — 'int withAnno @NN []' emits 'int withAnno@NN []' (bare name+dimensions concatenation)
## What

A TYPE_USE annotation on an array dimension loses the space before it. Measured on the #7159 branch (`7f4cf6c23`, whose java.go diff is 0 non-comment lines, so this is `main`'s behaviour):

```java
int withAnno @NN [];
```

| | emitted `Signature` |
|---|---|
| expected | `int withAnno @NN []` |
| actual | `int withAnno@NN []` |

Compiles clean under javac 25.0.3 (`@NN` declared `@Target(ElementType.TYPE_USE)`), so this is legal Java, not a derived-legal construct.

**Mechanism:** `buildFieldSignature` assembles the declarator as `decl := name + javaDeclaratorDimensions(...)` — a bare concatenation with no separator. The dimensions text begins at the annotation rather than at `[`, so the space that separated the name from the annotation in the source is dropped.

## Provenance

Found by the independent adversarial reviewer on **#7159** while attacking that PR's claim that whitespace is *"the only content the dimensions text can hold"*. That claim is false — the dimensions text can also hold a comment (`int withComment[/* a comment */];`, emitted verbatim) and, as here, an annotation. #7159 is being corrected for the false claim; **this defect is not #7159's and is not caused by it**.

## Not claimed

- **No corpus incidence.** `int withAnno @NN [];` is legal and idiomatic in codebases using TYPE_USE annotations (Checker Framework style), but I have not counted it. Measure before spending on a fix.
- **Not measured beyond `buildFieldSignature`.** The same `name + dimensions` concatenation shape may exist at the record-component site or elsewhere; the reviewer probed the field path only. Four other whole-span collapse sites exist in this file, and a verdict at one says nothing about the others — that is the lesson #7118 just established by measurement.
- **Severity: low, and lower than its siblings.** The visible consequence is one missing space. Compare the open defects in the same function: **#7117** (an annotated field loses its **type and name** entirely) and **#7116** (the unanchored modifier strip eats words out of a string literal). Fix this when touching these sites anyway, not as its own errand.

## Grading requirement, when someone takes it

The axis is **over-firing-adjacent**: the output is wrong rather than missing, so a recall-style must-have row can see it — but only if a fixture contains an annotated dimension, and none does today.

- One must-have row asserting `int withAnno @NN []` with the space intact.
- A **comment-content** row too (`int withComment[/* a comment */]`), because both falsify the same over-claim and a fix that special-cases annotations would leave the comment case free to regress.
- Both rows must sit in a fixture where the declared names are **unique**, and the assertion must be on the exact `Signature` string rather than a substring: an over-broad assertion survives deleting the very thing it checks.

Related: #7118 (the intra-line whitespace axis, measured ALIVE at five sites — this is a sibling of that family but a *concatenation* defect rather than a *collapse* defect), #7117, #7116.

cajasmota:
## What

`buildFieldSignature` glued the declarator together as `name + javaDeclaratorDimensions(node, src)` — a bare concatenation with no separator. That is sound only while the dimensions text starts at `[`. JLS 10.2 spells the suffix `{Annotation} [ ]`, so a TYPE_USE annotation belongs to the `dimensions` node and its text can begin at `@`. The source's separating space then falls *between* the two operands, where nothing puts it back.

| declaration | before | after |
|---|---|---|
| `int withAnno @NN [];` | `int withAnno@NN []` | `int withAnno @NN []` |
| `int twoAnnoDims @NN [] @NN [];` | `int twoAnnoDims@NN [] @NN []` | `int twoAnnoDims @NN [] @NN []` |
| `int plainDims[];` | `int plainDims[]` | `int plainDims[]` (unchanged) |
| `int withComment[/* a comment */];` | `int withComment[/* a comment */]` | unchanged |
| `int normAnno@NN[];` (no source space at all) | `int normAnno@NN[]` | `int normAnno @NN[]` — a space **manufactured**, see *Policy* |
| `int plainTwoDims[][];` | `int plainTwoDims[][]` | unchanged |
| `int annoNotFirst[] @NN [];` | `int annoNotFirst[] @NN []` | unchanged |

Both "before" columns were **measured on the pre-change file** (the RED run), not written from expectation.

### Policy: NORMALISE, not replay (review finding F1)

The fix emits **exactly one space** before a dimensions suffix that opens at `@`, **whether or not the source had one**, and none before one that opens at `[`. On `int normAnno@NN[];` — javac-clean, no space anywhere — it therefore **manufactures** a space rather than restoring one.

That is deliberate and now stated as such in the doc block: a `Signature` is a rendered form, and this function already collapses every whitespace run inside it, so propagating incidental source spacing here would be the inconsistent choice. The earlier wording said the separator was *restored*, which was prose asserting what no test observed on that input. The word "restore" no longer stands anywhere over this behaviour, and `normAnno` grades it.

The predicate reads **character 0** rather than searching for `@`. `annoNotFirst` (`int annoNotFirst[] @NN [];`) pins that distinction.

**Scope: the normalisation is POSITION-0 ONLY, and #7180 is the asymmetry it leaves.** Only the first character of the dimensions text is examined, so an annotation sitting *later* in the same text with no space in front of it keeps none. One construct therefore renders three ways:

| source | emitted | |
|---|---|---|
| `int x@NN[];` | `int x @NN[]` | leading, no source space → **gains** one |
| `int x[]@NN[];` | `int x[]@NN[]` | inner, no source space → **keeps none** |
| `int x[] @NN [];` | `int x[] @NN []` | inner, with source space → keeps its own |

Inner positions are **not** untouched territory, and "inner positions are not normalised" would be a false way to say this: an inner whitespace **run** is already collapsed upstream by `collapseJavaSpaces` — `int x @NN[]   @NN   [];` emits `int x @NN[] @NN []` (measured on review). It is specifically the inner **zero-space** case that is left alone. Filed as **#7180**; deliberately not fixed here.

**Measured on review, NOT pinned by any row**: every committed row carries zero or one source space, so the "exactly one space" claim is graded only at that end. The multi-separator end was probed by the reviewer — three spaces, a TAB, and a NEWLINE before `@NN` each emit exactly one space (`int threeSpaces @NN []`, `int tabSep @NN []`, `int newlineSep @NN []`). The claim **holds** there; it is simply not pinned by this PR's rows, and a reader should not assume otherwise.

## This is a concatenation defect, not a collapse defect — verified, not assumed

#7118 (`53ede5716`) shipped the intra-line-whitespace family across six collapse sites, and two of its rows (`spacedDims`, `tabbedDims`) drive this very helper. They only ever reach it with text starting at `[`, so the missing separator is invisible to them and no amount of collapsing could restore it. Verified rather than argued: with the fix reverted (mutant **M1**) the entire `issue7118_intraline_whitespace_test.go` file stays GREEN and only this PR's two rows fail.

Two stale over-claims are corrected in passing — `issue7118_intraline_whitespace_test.go`'s "that text is the only thing the dimensions node can hold besides the brackets themselves", and the matching paragraph in `buildFieldSignature`'s doc block that recorded the annotation and comment cases as deliberately ungraded. They are graded now.

## Pair audit — every site where a name can be followed by a dimensions/suffix fragment

`name + <suffix>` is a **1-of-1** shape in `internal/extractors/java/java.go`. Verdicts are per site and each one is pinned by a row in the new test, so this is a standing witness rather than a one-off measurement.

| site | how it builds the declarator | verdict |
|---|---|---|
| `buildFieldSignature` (java.go:2251) | `name + javaDeclaratorDimensions(...)` — **concatenation** | **DEFECTIVE — fixed here** (`withAnno`, `twoAnnoDims`) |
| method parameters, via `buildMethodSignature` | replays the raw header span | correct already; pinned by `DimHost.viaParam` = `int viaParam(int paramAnno @NN [])` |
| record components, in `walk` (java.go:~519) | replays the raw component span | correct already; pinned by `DimRec.recComp` = `int recComp @NN []` |
| `buildClassSignature` | raw span; its `name` parameter is never concatenated | not applicable |
| `buildAnnotationElementSignature` | raw span, no `name` parameter | not applicable |
| `sig = typeName + " " + fieldName` (java.go:523) | record-component fallback; separator is **explicit**, no dimensions fragment | not applicable |

The record-component site is the one the issue body named as the likely twin. It **can** carry an annotated dimension — `record RecDims(int comp @TU []) { }` compiles clean under javac 25.0.3 with `@TU` declared `@Target(ElementType.TYPE_USE)` (with a non-TYPE_USE annotation javac rejects the *annotation*, not the dims, which is how I know the dims are syntactically fine there). It does not have the defect, because it replays a span rather than concatenating. Nothing is fixed there and nothing is filed for it.

## Axes

**VARIED**

- **Leading content of the dimensions text** — a TYPE_USE annotation (`@NN []`) vs the bracket itself (`[]`). *Both directions* are graded, so a fix that inserts the separator unconditionally fails on `plainDims`.
- **Non-whitespace content inside the brackets** — `[/* a comment */]`, the other falsifier of the "whitespace is the only content" claim. Its expected value is **unchanged** by this PR: it is a regression pin, asserted on exact equality so that it stays one, and it dies under the unconditional mutant.
- **Name-to-annotation spacing IN THE SOURCE** — `int withAnno @NN [];` (one space) vs `int normAnno@NN[];` (none). This is the axis the emitted form differs on between a replay policy and a normalising one. It was previously in **neither** list — the silent-neighbour shape — and is now varied, named, and pinned.
- **Dimension count** — one vs two, on **both** branches: `@NN [] @NN []` on the annotated side and `[][]` on the plain side. A plain multi-dimension declaration did not previously exist **anywhere in this package**, so the false branch had been pinned only at count 1.
- **Position of the annotation within the dimensions text** — at its head (`@NN []`) vs after a plain dimension (`[] @NN []`).
- **Site** — field vs method parameter vs record component (the pair audit above).

**HELD CONSTANT**, named so the next reader can see what is still open: the element type (`int` everywhere), the annotation identity (`@NN`, no arguments), the line count (every declaration single-line, and *asserted* single-line — a wrap would move the row onto #7114's newline axis), the file, the package, and the absence of any initializer.

**NOT varied here, and MEASURED-ON-REVIEW rather than pinned**: an annotation *with arguments* on a dimension, and whitespace or a comment between the name and a `[` that carries no annotation. The reviewer enumerated 13 javac-clean shapes with the raw `dimensions` span printed and established that the first character of that span is exactly `@` or `[` — never whitespace, a comment or a newline — and that a `[` inside an annotation argument (`@Sz(msg="[weird]") []`) is never at position 0. Read their absence from the row list as measured-but-unpinned: **neither** untested-and-unknown **nor** graded.

Declared names are unique across the fixture, and every assertion is on the **exact** `Signature` string, never a substring.

## javac

Compiled before being asserted on. `javac -Xlint:all`, javac **25.0.3**:

| fixture | result |
|---|---|
| `com/example/dims/DimHost.java` (the full test fixture, **recompiled after the three new declarations**, byte-identical to `java7161DimsSrc` — verified by `diff`) | **exit 0, ZERO diagnostics** |
| `RecDims.java` — record component with `@TU []` | **exit 0**, zero diagnostics |
| `RecDims.java` — same but `@Deprecated []` | exit 1: *"annotation @Deprecated not applicable in this type context"* — the annotation is rejected, the dims are not |

## Mutants

Every mutant: `cp` backup, `python3` edit asserting **exactly 1 match** before running, mutated line **read back and printed**, `go vet` exit captured into its own file (never through a pipe), `--- FAIL` counted with an anchored `grep -cE '^\s*--- FAIL'`, restore by `cp` + `shasum` + a marker grep asserting count 0. Restored file hashes match the pre-mutation baseline exactly (`f582a386…` / `7ddd2645…`) and the marker count is 0.

**Re-scored from scratch after the new rows landed** — kill counts are not carried over from round 1, because a carried count can hide a mutant whose only killer no longer exists.

| # | mutant | vet | `--- FAIL` | verdict | killing assertions in THIS PR's file |
|---|---|---|---|---|---|
| M1 | the fix reverted (`dims` used with no separator) | 0 | **1** | **DEAD** | `withAnno`, `twoAnnoDims`, **`normAnno`** |
| M2 | M1 **with this PR's test file held out** — positive control | 0 | **0** | **ALIVE 0** (as required) | — (none exist) |
| M3 | separator inserted **unconditionally** | 0 | **3** | **DEAD** | `withComment`, `plainDims`, **`plainTwoDims`**, **`annoNotFirst`** |
| M4 | predicate `!HasPrefix(dims,"[")` → `strings.Contains(dims,"@")` — fires on an `@` **anywhere** instead of at position 0 | 0 | **1** | **DEAD** | **`annoNotFirst`** — and nothing else in the whole package |
| M5 | predicate → `dims != "[]"` — an exact-match false branch instead of a prefix one, i.e. correct at dimension count 1 only | 0 | **3** | **DEAD** | `withComment`, **`plainTwoDims`**, **`annoNotFirst`** (note `plainDims` does **not** fail: it is exactly `[]`) |

M4 and M5 exist to grade the two claims the new rows make, and each is killed by a row that did not exist before this round. M5 is the one that shows the count-1-only pin was a real hole: without `plainTwoDims` and `annoNotFirst` it would have been scored on `withComment` alone.

**Decomposition, summed.**

- **M1** — 1 failing test = `TestJava7161_AnnotatedDimensionKeepsItsSeparator` (1). Row errors inside it: `withAnno` (1) + `twoAnnoDims` (1) + `normAnno` (1) = **3**. The 7118 file contributes **0**, which is the measurement behind the "collapse tests do not grade this" claim.
- **M2** — 0 failing tests, 0 row errors. Positive proof the hold-out landed: `go test -list '.*'` returned **369** test names with the file present and **368** without (delta 1 = the one `Test…` function the file defines; `java7161One` is a helper, not a `Test`), `grep -c TestJava7161` on the without-list = **0**, and a `-v` run contained **0** `=== RUN` lines matching `TestJava7161`. So M1 is killed by this PR's rows and by nothing else in the package.
- **M4** — 1 failing test, 1 row error: `annoNotFirst`. Isolated: it perturbs only inputs whose dimensions text contains an `@` away from position 0, of which this package has exactly one.
- **M5** — 3 failing tests = `TestJava7117…` (1) + `TestJava7118…` (1) + `TestJava7161…` (1) = **3**. Row errors: 1 + 2 + 3 = **6**. Same confound class as M3 and declared the same way.
- **M3** — 3 failing tests = `TestJava7117_CStyleArrayDimensionsSurvive` (1) + `TestJava7118_IntraLineWhitespaceRunsCollapseAtEverySite` (1) + `TestJava7161_AnnotatedDimensionKeepsItsSeparator` (1) = **3**. Row errors: 2 + 2 + 4 = **8**. This PR's own contribution is `withComment` + `plainDims` + `plainTwoDims` + `annoNotFirst` = **4**, which is the point: the new rows grade the **no-annotation** direction on their own, not only the annotated one.

**Confound, declared.** M3 is not isolated — it perturbs every C-style dimension in the package, so it also breaks #7117's and #7118's pre-existing rows. That does not weaken the verdict it was built for, because this PR's own two rows fail independently inside it; it is recorded here rather than left for a reader to notice.

## Gates

- `go test -count=1 ./internal/extractors/java/` — **exit 0, 0 `--- FAIL`** (full package, not a `-run` subset).
- `go test -count=1 ./cmd/grafel/` with a **fresh `GRAFEL_HOME` and `GRAFEL_DAEMON_ROOT`** — **exit 0, 0 `--- FAIL`**, 167s, run on `6150ce63a`. **Not re-run on `00dd78b8c`, and here is why that is legitimate rather than a skip**: `git diff 6150ce63a 00dd78b8c` touches `java.go` in **comments only** (filtered diff below: 12 changed lines, all additions, all in the test file's fixture and row list, zero in `java.go` outside comments). The compiled extractor is byte-identical to the sha that was gated. This is an extractor change and `cmd/grafel/custom_extractor_spans_6118_test.go` digest-pins the whole graph, so a diff-derived package set would never have reached it.
- `gofmt -l internal/extractors/java/` — empty.

**Digest status: UNMOVED, and that is correct rather than lucky.** `writeSpanFixture6118`'s 18 members include exactly one `.java` file, `java/OrdersController.java`. That file declares **no fields at all** — only `@RestController class OrdersController` and two `@GetMapping`/`@PostMapping` methods — so it contains zero `variable_declarator` nodes and therefore zero `dimensions` nodes. `buildFieldSignature` is never called on it, so the changed line cannot alter any entity it produces. Nothing was bumped; the constant is untouched.

## Filtered diff `6150ce63a` → `00dd78b8c`, non-comment lines only

```
+    int normAnno@NN[];
+    int plainTwoDims[][];
+    int annoNotFirst[] @NN [];
+        {"SCOPE.Schema", "DimHost.normAnno", "int normAnno @NN[]"},
+        {"SCOPE.Schema", "DimHost.plainTwoDims", "int plainTwoDims[][]"},
+        {"SCOPE.Schema", "DimHost.annoNotFirst", "int annoNotFirst[] @NN []"},
```

Twelve changed lines in total once the six interleaved blank lines are counted; **zero deletions, and zero non-comment changes in `java.go`**.

## What the reviewer did NOT re-verify, stated so it does not read as confirmed

- `./cmd/grafel/` and `gofmt -l` were **not** re-run by the reviewer — they verified the digest *reasoning*, not the run. Both are mine, from the runs recorded above.
- M2's `-list` **369 → 368** hold-out arithmetic was **not** reproduced by the reviewer. I re-derived it in round 2: WITH = 369, WITHOUT = 368, `grep -c TestJava7161` on the without-list = 0, `=== RUN` lines matching `TestJava7161` = 0.
- The reviewer's own `MB` finding (predicate rewritten to its contrapositive `HasPrefix(dims,"@")`) is **ALIVE + equivalent**, on their 13-shape enumeration. I have not re-derived that equivalence and am not claiming it as mine; it is recorded here because a later reader should not re-open it as a hole.

## Not claimed

**No corpus incidence.** The issue said so and I did not change it: `int withAnno @NN [];` is idiomatic in Checker-Framework-style code but I have not counted it anywhere. The severity is still one missing space in one signature.

Closes #7161


cajasmota:
## Adversarial review — independent checkout at `origin/wt-7161-a1` (`6150ce63a`)

I did not write this code. Own worktree, own mutants, baseline `java.go` = `f582a386…` (matches the author's), package baseline **exit 0 / 0 `--- FAIL`** before any mutation. Everything below is MEASURED unless marked otherwise.

---

### Attack 1+2 — the predicate is a proxy; enumerate what `dims` can actually begin with

This was my primary attack and it **did not land**. I built a 13-shape probe, compiled it with **javac 25.0.3 `-Xlint:all` → exit 0, zero diagnostics**, then printed the **raw `dimensions` span**, the collapsed text, and the emitted `Signature` from an in-package probe (since deleted; tree restored).

| source | raw `dimensions` span | emitted `Signature` |
|---|---|---|
| `int a1 @NN [];` | `@NN []` | `int a1 @NN []` |
| `int a2 @Sz(max=5) [];` | `@Sz(max=5) []` | `int a2 @Sz(max=5) []` |
| `int a3 @Sz(msg="[weird]") [];` | `@Sz(msg="[weird]") []` | `int a3 @Sz(msg="[weird]") []` |
| `int a4 @Sz(vals={1,2}) [];` | `@Sz(vals={1,2}) []` | `int a4 @Sz(vals={1,2}) []` |
| `int a5@NN[];` | `@NN[]` | `int a5 @NN[]` ← **see finding F1** |
| `int a6 /* c */ [];` | `[]` | `int a6[]` |
| `int a7/*c*/[];` | `[]` | `int a7[]` |
| `int a8\n        [];` | `[]` | `int a8[]` |
| `int a9 [];` | `[]` | `int a9[]` |
| `int b1[] @NN [];` | `[] @NN []` | `int b1[] @NN []` |
| `int b2 @NN [] [];` | `@NN [] []` | `int b2 @NN [] []` |
| `int d1 @NN /* mid */ [];` | `@NN /* mid */ []` | `int d1 @NN /* mid */ []` |

**Enumeration of the first character of `dims`: exactly `@` or `[`.** Never whitespace (`collapseJavaSpaces` = `strings.Join(strings.Fields(...))`, which trims, and the span starts at the token anyway), never a comment, never a newline. The two attacks I was pointed at both fail:

- **Annotation *with* arguments** — handled (`a2`). An argument containing a literal `[` (`a3`, `a4`) does **not** fool `HasPrefix(dims, "[")`, because the bracket is never at position 0.
- **Whitespace/comment between the name and a non-annotated `[`** — the author's stated reason ("outside the `dimensions` span") is **correct by measurement**, not just by reading: `a6`, `a7`, `a8` all produce a span of exactly `[]`. Their span boundary belief is right, so the predicate's domain is what they say it is.

**Mutant MB** — predicate rewritten to its contrapositive `strings.HasPrefix(dims, "@")`: vet 0, **0 `--- FAIL`, ALIVE**. I record this as **ALIVE + equivalent, deliberately untested**, with the enumeration above as the demonstration: over the measured alphabet `{@, [}` plus the `dims != ""` guard, `!HasPrefix("[")` and `HasPrefix("@")` are the same function. Algebra alone would not have been enough; the 13-shape javac-verified enumeration is what makes it an equivalence rather than a hole.

### Attack 3 — pair-audit "already correct" verdicts. **Did not land; the witnesses are real pins.**

I mutated each replay site to simulate the concatenation defect and checked the new rows actually fail:

- **MC** — record-component replay (`java.go:520`) mutated to drop the space before `@`: vet 0, **1 `--- FAIL`, 1 row**, and it is `DimRec.recComp.Signature = "int recComp@NN []"`. **DEAD, and killed by this PR's row and nothing else in the package.**
- **MD** — parameter replay (`java.go:2335`, scoped to `"Anno @"` so the method-annotation confound is excluded — declared): vet 0, **1 `--- FAIL`, 1 row**, `DimHost.viaParam.Signature = "int viaParam(int paramAnno@NN [])"`. **DEAD**, same attribution.

So both witness rows *can* fail, and both would catch a future rewrite from replay to assembly. This is the part of the PR I tried hardest to break and could not.

### Attack 4 — M3's confound declaration. **Honest; decomposition re-derived and it sums.**

`if dims != ""` (unconditional separator): vet 0, **3 `--- FAIL`**, and the row errors are **6**:

- `TestJava7117_CStyleArrayDimensionsSurvive` → `arr`, `wrappedDims` (2)
- `TestJava7118_IntraLineWhitespaceRunsCollapseAtEverySite` → `spacedDims`, `tabbedDims` (2)
- `TestJava7161_AnnotatedDimensionKeepsItsSeparator` → `withComment`, `plainDims` (2)

3 = 1+1+1, 6 = 2+2+2, and this PR's own two rows fail **independently** inside the confound. Exactly as stated.

### Attack: M1 reproduced

`if false` (fix reverted): vet 0, **1 `--- FAIL`**, 2 rows (`withAnno`, `twoAnnoDims`), and **zero failures from the #7118 file**. The "collapse tests do not grade this" claim is confirmed on my own tree. (I did not re-run M2's hold-out; M1's attribution is visible directly in the failure list.)

### Attack 5 — the digest explanation. **Confirmed.**

`writeSpanFixture6118`'s single `.java` member is `java/OrdersController.java` (grep for `.java` members returns exactly one hit, line 143). Read in full: `@RestController class OrdersController` with `list()` and `create()` and **no field declarations at all** — the only `[]` in it is inside the string literal `return "[]";`. Zero `variable_declarator`, zero `dimensions`, so `buildFieldSignature` is never reached and an unmoved digest is correct rather than lucky. *Not independently re-run:* I did not re-execute `./cmd/grafel/` (machine budget); I verified the stated reason, not the author's exit code.

### Attack 7 — exact-string assertions. **Confirmed.** Every row goes through `if got.Signature != want.sig` — exact equality, no `Contains`, no substring. `java7161One` additionally `t.Fatalf`s on a non-unique match and asserts single-line spans.

### Attack 6 — the two corrected over-claims. **Correct, and not over-corrected.** The new #7118 comment now says whitespace is *not* the only content the dimensions text can hold (comments and TYPE_USE annotations also), which matches the measured set; the `java.go` doc block likewise. Neither replaces the old guess with a new one.

---

## Findings

**F1 — an ungraded input on the new branch, where the emitted signature is NOT a source replay (minor, measured).**
`int a5@NN[];` is javac-clean (exit 0, zero diagnostics). Its `dims` is `@NN[]` — starts at `@`, so the new branch fires and emits **`int a5 @NN[]`**, inserting a space the source never had. The behaviour is defensible (this function already normalises whitespace), but the doc block describes the mechanism as *"The source's separating space then falls BETWEEN the two operands… **Restore** the separator"* — on this input there is no source space to restore, and the code **manufactures** one. That is the "prose asserts what no test observes" shape. It is one row and one clause.

**F2 — one axis neither varied nor named (minor).**
The Axes section carefully lists what is held constant, but the **name↔annotation spacing** axis — the one the predicate's output actually differs on between "replay" and "normalise" — appears in neither list. Related and smaller: the **dimension-count** axis is varied only on the annotated side (1 vs 2); there is **no plain multi-dimension row anywhere in the package** (`grep` for `[][]` across the 7114/7117/7118 fixtures returns nothing), so the false branch is pinned only at count 1. Current output on `int b1[] @NN [];` → `int b1[] @NN []` is correct (measured) but ungraded.

Neither finding is a correctness defect in the shipped code. Both are cheap.

---

## What I could NOT verify

- `./cmd/grafel/` and `gofmt -l` were not re-run here (I verified the digest *reasoning*, not the run).
- M2's `-list` 369→368 hold-out arithmetic was not reproduced; M1's own failure list already attributes the kill.
- No corpus incidence measurement — the PR does not claim one.

---

## REQUEST-CHANGES

Both items are small; the fix itself is correct and I found no way to make the predicate over- or under-fire on a javac-legal input.

1. **Add a row for `int a5@NN[];`** (name and `@` touching in the source) asserting the exact emitted `Signature`, so the new branch's normalising behaviour is graded rather than incidental — and **soften the doc block's "restore the separator" wording** to say what it measurably does: it *ensures* a single separator before a dimensions suffix that opens at `@`, whether or not the source had one.
2. **Name the name↔annotation spacing axis in the Axes section** (varied, once item 1 lands), and record the **plain multi-dimension** shape (`int md[][];`, and/or `int b1[] @NN [];` where the annotated dimension is not the first) as either a row or an explicitly-held-constant axis — right now the false branch is pinned only at dimension count 1 and only with no annotation anywhere in the text.

Everything else — the pair audit, the M3 confound declaration and its arithmetic, the digest explanation, the exact-equality assertions, and the corrected over-claims — **verified and standing**.

PR #7182, opened 2026-09-18
test(#7174): acceptLoop's backoff clamp was ungraded — deleting it left the package green at 0 FAIL; #7126 pinned the SUPERVISOR's clamp and was never transferred to this twin

### LINKED ISSUE #7174: internal/daemon: acceptLoop's backoff CLAMP is ungraded — deleting 'if backoff > acceptBackoffMax' leaves all 487 tests green; #7126 pinned the SUPERVISOR's clamp, never this twin
## What

`acceptLoop`'s backoff **clamp** is ungraded. Deleting it outright leaves the whole `./internal/daemon/` suite green at **487 passing, 0 `--- FAIL`**.

```go
// internal/daemon/server.go
if isTransientAcceptErr(err) {
    if backoff == 0 {
        backoff = acceptBackoffStart      // :984   role 1, the START VALUE
    } else {
        backoff *= 2                      // :986   role 2, the FACTOR   <- #7167, pinned by #7172
    }
    if backoff > acceptBackoffMax {       // :988   role 3, the CLAMP    <- THIS, ungraded
        backoff = acceptBackoffMax
    }
    logger.Warn("accept: transient error, backing off", "err", err, "retry_in", backoff)
    time.Sleep(backoff)
```

Line numbers are from PR #7172's head (`64e743bc2`), where `acceptBackoffStart`/`acceptBackoffMax` were hoisted from function-local to package scope (values byte-identical: `5ms` / `1s`). On `main` before that merge they are function-local; the code shape is otherwise the same.

Measured by the independent adversarial reviewer on #7172, full package per row, `--- FAIL` lines counted.

## Why the nearest pin cannot see it

**#7126 (`45f9745d0`) graded the SUPERVISOR's clamp, not this one.** That is the whole content of this issue: the same three-role structure exists at two sites, and pinning a role at one site says nothing about the other. This repo has now measured that four times — #7118 (five whole-span collapse sites, a verdict at one saying nothing about the others), #7162/#7167 (the growth factor at three sites), #7135/#7164 (the modifier prefix across three arms), and now this.

The roles were each found only after the previous one was pinned well enough for the next gap to show. At the supervisor that sequence ran threshold (#7111/#7123) → clamp (#7126) → factor (#7162). Here it is running factor (#7167/#7172) → clamp (this). **Expect a fourth: `acceptBackoffStart`'s VALUE is graded by nothing outside #7172's new file, and there only by a vacuity floor** — the reviewer measured `acceptBackoffStart` 5ms → 900ms as exactly 1 `--- FAIL`, the floor alone, with `0 of 5` ratios graded. A floor is a liveness check, not a pin on the value.

## NOT claimed

- **Not claimed that site A has the same hole.** The sibling factor site fixed alongside this one — `rebuildBackoff` in `internal/daemon/requests_drain.go` — has an in-loop clamp whose deletion is also ALIVE, but there a trailing `if d > max` still clamps, so it is plausibly **equivalent** rather than a hole. That was argued **by algebra and NOT enumerated**, which by this repo's own standard is not yet an equivalence verdict — a false equivalence retires a hole permanently, so it is worse than a false kill. Someone taking this issue should settle site A by enumeration over the distinguishing inputs, or leave it explicitly open. Do not assume it is fine because the argument sounds right.
- **No production defect is claimed.** The clamp is present and correct today. This is a grading gap: nothing would catch its removal or its weakening.
- **Not claimed to be reachable through a wall-clock assertion.** `-race` shifts this package's timings roughly 7× and the automatic CI legs run WITHOUT it, so a duration-based pin here would be scheduling luck rather than evidence.

## The seam, already established

Do not re-derive this — #7172 measured it. `acceptLoop` **announces** `retry_in=<d>` one line before it sleeps, and the test reads the log stream, so the sequence of intended waits is observable with no clock. Its known limit, disclosed in #7172 and inherited here: the test observes the **announced** value, so a defect that announced X and slept 4X would pass. That is an inherent limit of the bound-not-duration design and remains the right trade.

The clamp holds every wait ≤ `1s`, which is why even a 100× factor costs only ~4.5s at this site — so a clamp pin is cheap to run.

## Grading requirement

- **A bound assertion, not a sequence.** Assert every announced wait is `<= acceptBackoffMax`. A literal expected sequence would encode the start value, the factor and the ceiling together and break on any legitimate change to the other two — #7126 established this deliberately and #7172 preserved it.
- **Both directions of the guard's own input set**, as #7162's ratio pin did: force the graded set to exclude everything and to exclude nothing, and show each gives a distinct, attributable failure. A guard that can silently empty its own input set grades nothing.
- **A live vacuity floor with a POSITIVE control** — assert a minimum number of waits actually observed, and name the mutant that kills by the floor alone. Without that, "every wait is bounded" passes vacuously on zero waits.
- **Do not let one test kill roles at two sites.** Verify the new pin does NOT react to the supervisor's clamp mutant, and that #7126's does not react to this one. If a single test covers both, the "separate sites" framing is not earned and the next site's gap will hide behind it.
- **Assert against the constant, then separately against behaviour.** `acceptBackoffMax` is now package-scoped, so a guard can name it — and a guard that names a constant can be satisfied by the constant rather than by what production does with it. #7172's reviewer checked this explicitly for the factor pin; do it again here.

Related: #7126 / `45f9745d0` (the supervisor's clamp — the twin this one was never transferred to), #7167 / #7172 (the factor at this exact site), #7162 / `f9e7838e9`, #7111 / #7123 (the supervisor's threshold), #7160.

cajasmota:
## What

`acceptLoop`'s backoff **clamp** was ungraded: deleting
`if backoff > acceptBackoffMax { backoff = acceptBackoffMax }` outright left the whole of
`./internal/daemon/` green. #7126 (`45f9745d0`) pinned the SUPERVISOR's clamp and was never
transferred to this twin. This adds `internal/daemon/acceptloop_backoff_clamp_test.go`.

**No production change.** The clamp is present and correct today; this is a grading gap.
The diff is one new test file inside `internal/daemon/`. **`./cmd/grafel/` was NOT run and is
not required** — the change touches no extractor, so the graph digest cannot move.

## The assertion, and why it is not a sequence

A **BOUND**: every announced wait must be `<= acceptBackoffMax`. Not the literal
`[5ms 10ms 20ms 40ms 80ms 160ms 320ms 640ms 1s 1s]`, which would encode the start value, the
factor and the ceiling at once and break on any legitimate retune of the other two. #7126
established that shape and #7172 preserved it; this does not regress it.

The seam is inherited, not rebuilt: `acceptLoop` announces `retry_in=<d>` one line before it
sleeps, and this file reuses `announcedAcceptBackoffs` from the factor file rather than building
a second harness (a twinned helper is graded at whichever copy a mutant lands on). No assertion
reads a clock, so `-race` (~7x on this package, and the automatic CI legs run without it) cannot
move a verdict.

`acceptBackoffStart` is a `const`; there is no injection seam and none was added. The longer
sequence is driven by **scripting ten transient errors**, keeping the shipped start value.

## Mutant table

Full `./internal/daemon/` per row, `-count=1`, one mutant at a time, byte-level `cp` backup with
an asserted exact match count and the mutated line read back, `go vet` exit captured to its own
file (0 on every row), verdicts are **anchored** `--- FAIL` **line** counts (`grep -cE '^\s*--- FAIL'`),
never an exit code. Unmutated tree with this file: **0 `--- FAIL`, 490 tests listed.**

| # | Mutant | `--- FAIL` | Verdict | Decomposition |
|---|---|---|---|---|
| M1 | clamp block **deleted** | 1 | DEAD | this test: bound (waits 9,10 = 1.28s, 2.56s) **and** the at-ceiling floor |
| H1 | clamp block deleted, **this file held out** | 0 | **ALIVE** (positive control) | — |
| M2 | `> acceptBackoffMax` → `> 2*acceptBackoffMax` (weakened threshold) | 1 | DEAD | this test: bound (1.28s) |
| M3 | `backoff = acceptBackoffMax` → `= 2 * acceptBackoffMax` (weakened value, up) | 1 | DEAD | this test: bound (2s) |
| M4b | `backoff = acceptBackoffMax` → `= acceptBackoffMax / 3` (weakened value, down) | 2 | DEAD | 1 this test (**floor alone**, 0 of 10 at the ceiling) + 1 the ratio pin |
| M4a | `backoff = acceptBackoffMax` → `= acceptBackoffMax / 2` | 1 | **ALIVE here**, DEAD cross-file | 0 this test + 1 `TestAcceptLoopBackoffClampSkipExcludesOnlyCeilingPairs`, its **ratio assertion** at `acceptloop_backoff_factor_test.go:415` (640ms→500ms = 0.781x), not either of its floors |
| M5 | `>` → `>=` | 0 | **ALIVE, EQUIVALENT** | — |
| M6 | `acceptBackoffMax` 1s → 1h | 2 | DEAD | 1 this test (**floor ALONE**, zero bound errors) + 1 the ratio pin's excluded-floor |
| M7 | `acceptBackoffStart` = `acceptBackoffMax` | 4 | DEAD | 1 this test (**upper floor**) + 1 the start pin + 2 both ratio-pin tests |
| M9 | `acceptBackoffStart` 5ms → 7ms | 1 | DEAD | the start pin **alone** |
| M9ref | M9 **plus** the start pin rewritten to name `acceptBackoffStart` | 0 | **ALIVE** | — |
| G1 | test guard `if w == acceptBackoffMax` → `if true \|\| ...` | 1 | DEAD | this test, upper floor (`:228` on `5164e259a`) |
| G2 | test guard `if w == acceptBackoffMax` → `if false && ...` | 1 | DEAD | this test, vacuity floor (`:220` on `5164e259a`) |
| C1 | **supervisor's** clamp deleted (`waitBackoff`, supervise.go) | 1 | DEAD | #7126's test only; **this test GREEN** |

Every decomposition sums to its row's count.

**Positive proof of the hold-out (H1), carried by the `-list` diff alone:** `go test -list` shows
**490** tests with this file and **488** without it, and the difference between the two sorted
listings is **exactly** `TestAcceptLoopAnnouncedBackoffIsBoundedByTheCeiling` and
`TestAcceptLoopFirstAnnouncedBackoffIsTheShippedStartValue`; `grep -c` for either name in the 488
is 0. So the kills are earned by this file.

An earlier draft of this body also cited "zero `=== RUN` lines" in the held-out run. That is **not
evidence** and has been withdrawn: `=== RUN` is emitted only under `-v`, so on these non-verbose
legs the count is zero for every test in the package whether or not it ran. The `-list` diff stands
on its own.

**One row DISCARDED rather than reported.** `acceptBackoffStart` 5ms → 5µs yields 5 `--- FAIL`
lines, but they are a **harness artefact**: the shared `retry_in` regex does not match the `µ` of
`5µs`, so the helper fatals on a parse error before any assertion in this file runs. It grades
nothing, so it is excluded and 7ms is the row that actually exercises the start value.

**Declared confounds.** Both constants are also read by the ratio pin next door, so every row that
retunes one reacts there too (M6, M7, M9-adjacent). Each such row's contribution is named above.
No row is claimed as isolated except M9 and the guard rows, which are measured isolated.

## The vacuity floor and its named killer

"Every wait is bounded" passes vacuously on zero waits, and — worse — the bound names
`acceptBackoffMax`, which is package-scoped, so it can be satisfied **by the constant**. Both are
closed by the at-ceiling partition, whose two directions are the test's only guard:

- **at least one wait exactly at the ceiling.** Named killer: **M6** (`acceptBackoffMax` 1s → 1h).
  Measured, it fires this floor **alone** — zero bound errors, the start pin still green — on the
  unchanged sequence `[5ms … 1.28s 2.56s]`. That row *is* the proof the bound is not
  self-satisfying: with the constant enlarged, the bound goes quiet and only the floor speaks.
- **not every wait at the ceiling.** Named killer: **M7** (`acceptBackoffStart = acceptBackoffMax`),
  sequence `[1s × 10]`.

Both directions are also scored **on the guard itself** (G1/G2), each giving a distinct,
attributable failure at a different assertion for a different reason: set-**swallowed** vs
set-**emptied**.

**And the fixture actually drives the guard through both branches** — the difference between a
guard that is graded and one that merely looks graded. The shipped ten-error sequence puts **8
waits below the ceiling and 2 at it**, so `w == acceptBackoffMax` is taken in both directions by
the fixture itself. That is not free: the six-error probe next door never reaches the ceiling, so a
guard written against it would have been dead code in its own fixture — exactly the miss that
shipped earlier today. The input set also cannot be empty upstream: `announcedAcceptBackoffs`
`t.Fatalf`s unless `len(waits) == n`.

**A claim I wrote and then measured false, corrected in the file.** I had written that the floor
also catches the shrink direction via `acceptBackoffMax / 2`. It does not (M4a): capping at 500ms
makes the *next* doubling land on exactly 1s, which satisfies the floor. `/3` (M4b) is the shrink
row the floor does kill; `/2` is killed cross-file instead — solely by the **ratio assertion** of
`TestAcceptLoopBackoffClampSkipExcludesOnlyCeilingPairs` at `acceptloop_backoff_factor_test.go:415`,
not by either of its floors. The file header now **names that dependency by file, test and
assertion**: it is load-bearing (the shrink direction is ALIVE in the new file), and an unnamed
cross-file dependency is exactly what gets retuned away by someone who has no idea another file
leans on it.

## Cross-site disjointness, both directions

| direction | result |
|---|---|
| `acceptLoop`'s clamp deleted (M1) | 1 `--- FAIL`, **this test only**; `supervise_backoff_clamp_test.go` GREEN |
| `waitBackoff`'s clamp deleted (C1) | 1 `--- FAIL`, **#7126's test only**; this test GREEN |

So the "separate sites" framing is earned rather than assumed.

## `>` → `>=` is EQUIVALENT, left untested on purpose

Measured ALIVE at 0 `--- FAIL`, and recorded as an equivalence rather than a hole — and the review
established it is **stronger than this PR first claimed**, so the header has been upgraded to match.
The equivalence does **not** rest on the reachable alphabet: it holds for **every value of
`backoff`**, i.e. over the whole `int64` domain. The two forms differ only at
`backoff == acceptBackoffMax`, where the body is an **idempotent write of the value already held**,
and nothing observes the difference — the `if` and its body are adjacent with **no read between the
comparison and the assignment**, `backoff` is a plain local `time.Duration` with no pointer
escaping, and there is no counter, no logging and no early return, so `logger.Warn` then
`time.Sleep` see the same post-state either way. The reachable alphabet
(`{5ms … 640ms, 1s, 2s}`) agrees on every element too, but that is now belt-and-braces rather than
load-bearing — which makes the verdict cheaper to re-check than an enumeration would. The same
equivalence is already recorded at the twin in `supervise.go`. A test pinning `>` here would be
vacuous by construction.

## `acceptBackoffStart`'s value: PINNED

#7174's body predicted a fourth role, and it is in scope here: `acceptBackoffStart`'s VALUE was
graded by nothing outside #7172's vacuity floor, and a floor is a liveness check, not a pin.
It is pinned by a **separate test** (so a `--- FAIL` line attributes to the role that moved),
reading only `waits[0]` from a ONE-error script, so it encodes neither the factor nor the ceiling.

It uses a **literal**, not the constant, and that choice is measured rather than argued: M9 vs
M9ref above is the same production mutant with only the test's expectation form changed —
literal DEAD 1, reference ALIVE 0. A guard that names the constant it checks is satisfied by the
constant.

## Not in scope, and deliberately left open

Site A, `rebuildBackoff` in `internal/daemon/requests_drain.go`. Its in-loop clamp deletion is
ALIVE and plausibly equivalent (a trailing `if d > max` still clamps), but that was argued **by
algebra and not enumerated**, so it is **not** an equivalence verdict by this repo's own standard
and a false equivalence retires a hole permanently. It is neither "fixed" nor declared equivalent
here; it stays unsettled and wants an enumeration over the distinguishing inputs.

Also not graded here, on purpose: the SLEPT duration (this seam observes the ANNOUNCED wait, so a
defect announcing X and sleeping 4X passes — inherent to the design and the right trade), the
growth factor (#7172's arm; this file asserts no ratio), the `backoff == 0` first-wait branch, and
the reset-on-success behaviour.

## Axes

**VARIED:** the clamp's presence, its threshold, its clamped-to value in both directions, the
number of waits observed at and below the ceiling, and the expectation form of the start-value pin
(literal vs reference).
**HELD CONSTANT:** the growth factor, the transient-error classification, the reset-on-success
behaviour, the harness (reused, not twinned), and the shipped values of both constants in the
unmutated tree.

Closes #7174


cajasmota:
## Independent adversarial review — #7182

Own detached worktree at `origin/wt-7174-a1` (`5164e259a`), **not** the author's. Every row below was
re-derived from scratch: byte-level `cp` backup, edit applied with an **asserted exact match count**,
the mutated line **read back**, `go vet ./internal/daemon/` exit captured to its **own file** (0 on
every scored row), full `go test -count=1 -timeout 1800s ./internal/daemon/`, verdicts are **anchored**
`grep -cE '^\s*--- FAIL'` line counts, never an exit code. Tree restored by `cp` and verified with
`git status --porcelain` → empty after every row. Diff confirmed: **1 file, +267, test-only**, so I
agree `./cmd/grafel/` is not required and I did not run it.

**My baseline on this tree: 0 `--- FAIL`, `go test -list` 490, 53.09s.**

### Rows I re-measured (all agree with the PR table)

| row | mine | claimed | attributed to |
|---|---|---|---|
| M1 clamp deleted | **1** | 1 | `acceptloop_backoff_clamp_test.go:207` twice (waits 9,10 = 1.28s, 2.56s) **and** `:220` floor — one test, so one line |
| H1 same, file held out | **0** | 0 | ALIVE — positive control |
| M4a clamp-to `max/2` | **1** | 1 | **0 in this file**; sole killer `acceptloop_backoff_factor_test.go:415` in `TestAcceptLoopBackoffClampSkipExcludesOnlyCeilingPairs` (640ms→500ms = 0.781x) |
| M5 `>` → `>=` | **0** | 0 | ALIVE |
| M6 `acceptBackoffMax` 1s→1h | **2** | 2 | `clamp_test.go:220` **floor ALONE, zero bound errors**, seq `[5ms … 1.28s 2.56s]` + `factor_test.go:421` excluded-floor. Sums. |
| M7 `acceptBackoffStart = max` | **4** | 4 | bound test, start pin, `…GrowsByExactlyTwoPerStep`, `…ClampSkipExcludesOnlyCeilingPairs`. Sums. |
| M9 start 5ms→7ms | **1** | 1 | start pin **alone** (0.01s) |
| M9ref same + expectation → `acceptBackoffStart` | **0** | 0 | **ALIVE** |
| G1 `if true \|\| w == acceptBackoffMax` | **1** | 1 | `:228` — "every one of the 10 … sat at the ceiling" |
| G2 `if false && …` | **1** | 1 | `:220` — "0 of 10 waits sat at it" |
| C1 supervisor clamp deleted (`waitBackoff`) | **1** | 1 | `TestEngineSupervisor_BackoffWaitsAreBoundedByTheCeiling` **only**; this file GREEN |

Positive control re-derived independently, not reused: `-list` **490 with / 488 without**, and the
`diff` of the two sorted listings is **exactly** `TestAcceptLoopAnnouncedBackoffIsBoundedByTheCeiling`
and `TestAcceptLoopFirstAnnouncedBackoffIsTheShippedStartValue`; `grep -c` for either name in the 488 is **0**.

### Attacks that did not land

**`>` → `>=` is equivalent, and by a stronger argument than the PR gives.** I tried to build a
distinguishing input and could not. The two forms differ only at `backoff == acceptBackoffMax`; there
the body is `backoff = acceptBackoffMax`, an idempotent write of the value already held. I checked the
specific hole the brief asked about — a read between the comparison and the assignment — and there is
none: `if` and body are adjacent, `backoff` is a plain local `time.Duration`, no pointer escapes, no
counter, no logging, no early return, and the next statement (`logger.Warn … time.Sleep`) reads the same
post-state either way. So the equivalence does **not** depend on the alphabet at all — it holds for every
`int64`, and the `{5ms … 2s}` enumeration is belt-and-braces rather than load-bearing. Measured ALIVE at 0.
I endorse the EQUIVALENT verdict, and would go further: the header could say "for every value of
`backoff`, not only the reachable ones", which makes the claim cheaper to re-check than an enumeration does.

**The graded-set guard is not dead code in its own fixture.** The shipped 10-error sequence is
`[5ms 10ms 20ms 40ms 80ms 160ms 320ms 640ms 1s 1s]` — **8 below the ceiling, 2 at it** — so
`w == acceptBackoffMax` is taken in both directions by the fixture, which is exactly the miss that
shipped earlier today. G1/G2 fire at **different lines for different reasons** (`:228` set-swallowed,
`:220` set-empty), so the guard cannot silently empty its own input set. I also checked the set cannot be
empty upstream: `announcedAcceptBackoffs` `t.Fatalf`s unless `len(waits) == n`.

**Cross-site disjointness is earned, both directions, re-derived by me.** M1 → clamp test only,
`supervise_backoff_clamp_test.go` green. C1 (I deleted `*backoff > s.backoffMax` in `supervise.go`,
match count asserted 1, read back, vet 0) → `TestEngineSupervisor_BackoffWaitsAreBoundedByTheCeiling`
only, this file green.

**"The literal is the pin, the reference is not" holds, and is documented.** M9 vs M9ref reproduce
exactly (1 vs 0). The brief's counter-attack — that an undocumented magic number gets "fixed" into a
reference and silently un-grades itself — does not land here: the trade is spelled out twice, on
`acceptClampStartValue` and in the `…IsTheShippedStartValue` doc comment, with the measured row pair
attached. That is the right trade: a value pin that names the value is the only kind that survives a
retune of the value.

**The anti-self-satisfaction proof holds.** M6 fires `:220` **alone** inside this file — zero bound
errors, start pin green — on a sequence the bound is entirely happy with. That is the row that shows the
bound is not satisfied by naming the package-scoped constant, and it is the strongest thing in this PR.

**The discarded 5µs row really is a harness artefact.** `retryInRe = retry_in=([0-9a-zA-Z.]+)`
(`acceptloop_backoff_factor_test.go:228`) excludes U+00B5, so `5µs` matches `5`, `time.ParseDuration("5")`
errors, and the helper `t.Fatalf`s before any assertion here runs. Correctly excluded.

### What I found that is not in the PR

1. **The shared `retry_in` regex has two latent holes, one of them fail-dangerous** (pre-existing, not
   introduced here — the 5µs row is its visible symptom). It rejects `µ`, so the harness cannot grade any
   sub-millisecond retune and reports it as an unrelated parse fatal; and it excludes `-`, so an announced
   **negative** duration would be captured as its positive magnitude and would **satisfy the bound**.
   Not reachable with today's fixtures (a negative backoff needs `int64` overflow, ~63 doublings against a
   10-error script), so latent, not a hole in this PR. Worth its own issue against the helper, with the
   5µs measurement as the evidence.
2. **The PR body's G1/G2 line numbers are stale.** It reports `:181` and `:173`; in the shipped file those
   are comment lines. The measured assertions are **`:228`** (`true ||`) and **`:220`** (`false &&`). The
   verdicts and the "distinct lines, distinct reasons" claim are correct — only the citations are from an
   earlier draft.
3. **"the held-out run contains zero `=== RUN` lines"** is only evidence under `-v`; a non-verbose
   `go test` emits none for any test. The `-list` 490/488 diff is the real proof and it stands on its own.
4. **The file header discloses the `max/2` cross-file dependency but does not name its killer.** It says
   "killed instead by the RATIO pin next door". Measured, the sole killer is
   `TestAcceptLoopBackoffClampSkipExcludesOnlyCeilingPairs` in `acceptloop_backoff_factor_test.go` — and
   specifically its ratio assertion, not its floors. Naming file+test in the header would make the
   dependency survivable: the thing that could silently un-grade the shrink direction is a future retune of
   *that* test, and nothing in this file points at it.
5. **Cost note, non-blocking.** The package now runs two independent ten-error probes (~3.3s of real sleeps
   each) — one in each file — because this file declines to share the factor file's. Baseline is 53.09s;
   I consider that the correct trade for attribution, but it is a cost the body does not state.

### What I could NOT verify

- Site A (`rebuildBackoff`, `requests_drain.go`). I did not enumerate it. Leaving it explicitly unsettled
  rather than declaring equivalence is the right call by this repo's standard, and I am not contradicting it.
- The SLEPT duration. Out of the seam by construction, as the header says; I did not attempt to grade it.
- Windows/macOS legs. Every row above is macOS/arm64, no `-race` (matching the automatic CI legs, #7062);
  no assertion in either new test reads a clock, so I do not expect a platform to move a verdict, but I
  did not measure one.

### Verdict

Twelve of twelve claimed rows reproduced, the two decompositions I checked sum exactly, the positive
control re-derived independently, both cross-site directions earned, and the one equivalence verdict
survives a direct attempt to break it — with a stronger justification than the PR claims for itself. The
self-corrections (the `max/2` header claim measured false and fixed; the 5µs row discarded rather than
banked) are the two places I expected to find an overstatement and did not.

**APPROVE.**

Non-blocking, worth folding in if the branch is touched again: correct the `:181`/`:173` citations to
`:228`/`:220`, drop or qualify the `=== RUN` sentence, name the cross-file killer in the header, and file
the `retryInRe` finding as its own issue.

PR #7184, opened 2026-09-18
fix(#7176): extractIndentBody silently dropped every line at baseIndentLen+1 — a dead band that moved EndLine, lost calls, flipped a subtype and dropped `inherit`

### LINKED ISSUE #7176: fsharp: extractIndentBody has a DEAD BAND — a line at exactly baseIndentLen+1 is neither appended nor a terminator, so a sibling falls inside the previous declaration's body; the source comment promised an issue and none was ever filed
## What

`extractIndentBody` has a **dead band**: a line indented at exactly `baseIndentLen+1` is neither appended to the body nor treated as a terminator, so the scan silently skips it **and keeps going**. A sibling declaration that happens to sit one column off therefore falls inside the PREVIOUS declaration's body.

`internal/extractors/fsharp/extractor.go:1034`, and the hazard is described in-source at `:1170-1176`.

## Why this is being filed now

The source comment that names it says, verbatim, that it and one sibling vector *"each needs its own issue and its own tests"*. **Neither issue was ever created** — I searched open and closed. So the hazard has been carried in a comment, unticketed and ungraded, while the helper's blast radius grew.

A documented hazard is not a test. In this repo the sibling with the comment is reliably the ungraded one.

## Blast radius — larger than the comment says

The comment frames the consequence narrowly, as an `inherit` clause attributed to the wrong owner in `collectHierarchyEdges`. But `extractIndentBody` is called by **three** declaration passes, and each uses its result for something different:

| caller | line | what the body is used for |
|---|---|---|
| let pass | `:617` | `endLine`, `collectCalls`, CE-usage edges, match-site edges |
| member pass | `:665` | same shape |
| type pass | `:720` | `endLine`, `classifyTypeSubtype`, CONTAINS edges, hierarchy edges |

So a swallowed line can move an **EndLine**, change a **subtype classification**, attach a **CONTAINS** edge to the wrong parent, or mis-attribute a **call** — not only an `inherit`. **A verdict at one of these three call sites says nothing about the other two**; that is the lesson #7118 established by measurement across five collapse sites and #7174 just repeated for a clamp. Whoever takes this must score all three.

## Not claimed

- **No corpus incidence.** The comment says "low frequency" and that is plausible — it needs a sibling indented exactly one column deeper than its predecessor's base — but it is **uncounted**, and "low frequency" is an adjective nobody has measured. Adjectives have been measured false in this repo repeatedly; treat it as unknown.
- **Not claimed to be a one-line fix.** The band exists because the loop distinguishes "deeper than base" from "at or shallower than base" and the boundary case falls between the two tests. Whether the correct behaviour is *append* or *terminate* is a real question about F#'s offside rule, and there is **no F# toolchain here** (`dotnet`, `fsc`, `fsharpc`, `fsi`, `mono` all absent; `javac` is the only compiler) — cite the spec's lightweight-syntax rules rather than picking the one that makes a test pass.
- **Not claimed the suite can see it today.** The in-source comment says "invisible to this suite", which is the whole reason this is a ticket rather than a bug report.

## Grading requirement

- **A fixture per call site** (let, member, type), not one representative — the three uses diverge, and a fix validated on the hierarchy path alone would leave the other two open.
- **Both directions of the boundary**: a line at `baseIndentLen+1`, and controls at `baseIndentLen` and `baseIndentLen+2`. A fixture that only exercises the band cannot show the neighbours still behave.
- **Assert the emitted artefact, not a counter** — the wrong `EndLine`, the wrong CONTAINS parent, the wrong subtype. A diagnostic signature is not a regression pin.
- **Score the widening direction.** Changing the boundary to *append* enlarges every body, which enlarges what `collectCalls` and the edge scanners see. `activePatternRE` and the declaration scanners read **raw `src`** (#7152), so a larger body can mint more from comments and strings. Forbidden rows required.
- If the fix makes any currently-unreachable line reachable, **assert that line's effect** — an unreachable line becoming reachable is a behaviour change and needs its own row.

Related: #7144 (the dedup keys — an extent-based fix there would REST on this helper, so settle this first), the sibling vector filed alongside this one, #7152, #7118, #7174.

cajasmota:
## The defect

`extractIndentBody` (`internal/extractors/fsharp/extractor.go:1034` on `b31fe2c67`) set
`minBodyIndent := baseIndentLen + 2`, so a line indented at exactly `baseIndentLen+1`
satisfied **neither** `indent >= minBodyIndent` **nor** `indent <= baseIndentLen`. The loop
fell through both branches and **kept scanning**: the line vanished from the body while
everything after it stayed in.

## The offside decision — sourced, not chosen to make a test pass

**Decision: the `baseIndentLen+1` line is APPENDED.** The terminator test stays at
`indent <= baseIndentLen`.

**Citation — F# 4.1 Language Specification, §15.1.4 "Offside Lines"**, whose own worked
example is indented by exactly one column:

```
let x = 1
 let y = 2 <-- unmatched 'let'
let z = 3   <-- warning FS0058: possible
                incorrect indentation: this token is offside of
                context at position (2:1)
```

The one-column-deeper `let y` is *"unmatched"* — it is **not** read as a sibling of `let x`;
it is absorbed as a continuation, and it is the **return to column 0** (`let z`) that goes
offside. The same section lists what *introduces* an offside line, including **"The start of a
`let`, `if` or `module` token"** — so the governing offside line for a declaration **is its own
column**, i.e. `baseIndentLen`. §15.1.8 states the boundary from the closing side: *"When a
token occurs **on or before** the offside limit for the current offside stack, and a permitted
undentation does not apply, enclosing contexts are closed."* On-or-before (`indent <= base`)
closes; **strictly greater continues**. That is `baseIndentLen+1`.

**The sharpest form of the rule, also §15.1.8** (added on review — it closes the falsifier more
directly than the FS0058 example): *"When a token other than `and` appears **directly on the
offside line** of Let context, and the next surrounding context is a SeqBlock, the `$in` token
is inserted."* A sibling sits **directly on** the offside line — at exactly `base`, not `base+1`
— and `and` is the single token the spec exempts. So `base+1` cannot be a sibling; it is body.

Spec text extracted locally from `FSharpSpec-4.1-latest.pdf` (fsharp.org/specs/language-spec/4.1/)
via `pdftotext -layout`; the quotes above are verbatim from §15.1.4 and §15.1.8. An independent
reviewer re-fetched the PDF, ran their own `pdftotext`, and confirmed all quotes verbatim.

**DERIVED-NOT-EXECUTED.** There is no F# toolchain on this machine (`dotnet`, `fsc`, `fsharpc`,
`fsi`, `mono` all absent; `javac` is the only compiler here). **What would falsify it:** an F#
program in which a construct at column `base+1`, following a declaration at column `base`,
parses as a **sibling** of that declaration rather than as part of its body. An independent
reviewer attempted to construct one — `and`-continued `let rec`, shallow `|` match arms,
`type … with` members, CE bodies — and **could not**; the `and` exemption above is exactly why.

Corroborating in-repo precedent (not the basis of the decision): the type pass's own CONTAINS
scan at `:733` already uses `pmIndentLen > typeIndentLen` — strictly-greater — for the same
"is this inside?" question.

## Per-call-site verdicts — all FIVE scored, each with its own fixture

The issue body named three call sites. There are **five**: `compexpr_active_patterns.go:517`
and `validators.go:344` also call `extractIndentBody`, and an earlier revision of this PR wrote
them off as *"covered transitively by the package suite"*. **That was false** — M2 measures the
162 pre-existing tests as blind to this boundary, so transitive coverage here is coverage by a
set already measured not to see it. Both are now pinned with their own rows.

Every row is an **emitted artefact**, measured on `b31fe2c67` before the fix and re-measured
after. No counters, no diagnostic signatures.

| call site (`b31fe2c67`) | artefact asserted | before | after |
|---|---|---|---|
| **let pass** `extractor.go:617` | `alpha` `EndLine` | **5** | **6** |
| | `alpha` CALLS | `[stepTwo]` | `[stepOne stepTwo]` |
| **member pass** `extractor.go:665` | `Alpha` `EndLine` (base indent **4**, deliberately non-zero) | **5** | **6** |
| | `Alpha` CALLS | `[deeperStep]` | `[memberStep deeperStep]` |
| **type pass** `extractor.go:720` | `Shape` `Subtype` | **`type`** | **`discriminated_union`** |
| | `Shape` `EndLine` | **4** | **6** |
| | `Shape` CONTAINS | `[]` | `[Shape.Circle Shape.Square]` |
| | `SCOPE.Schema/du_case/Shape.Circle` | **absent** | present |
| | `Derived` EXTENDS | `[]` | `[Base]` |
| | `Derived` `EndLine` | **7** | **8** |
| **CE-builder pass** `compexpr_active_patterns.go:517` | `Bind` `Subtype` | **`member`** | **`ce_member`** |
| | `Return` `Subtype` | **`member`** | **`ce_member`** |
| | `Bind` `ce_member` property | **absent** | `"true"` |
| | *(companion, from the type pass)* `OptBuilder` `Subtype` | `type` | `computation_builder` |
| **record-name pass** `validators.go:344` | `Person` VALIDATES | **absent** | **`-> Addr`** |
| | *(companion, from the type pass)* `Addr` `Subtype` | `type` | `record` |

Each of the two new passes threads its result through a consumer **only it** produces, and that
is the artefact asserted: `collectCEBuilderTypes` is the sole producer of `ceMemberNames`, which
drives `ce_member` re-typing in the member pass; `collectRecordTypeNames` is the sole producer
of the `recordTypes` set that lets a field typed as another in-file record mint the nested-model
`VALIDATES` edge. The companion rows (`OptBuilder` / `Addr` subtype) come from the *type* pass's
own `detectCEBuilder` / `classifyTypeSubtype` calls and are labelled as such in-test so the
attribution is not confounded.

All five passes fail **independently** under the reverted fix (M1 below) — a verdict at one
would not have covered the others.

| fixture | VARIED | HELD CONSTANT |
|---|---|---|
| let pass | first body line's indent: **1 / 2 / 0** columns | declaration column (0), keyword (`let`), second body line's indent (4), callee names, file path |
| member pass | first body line's indent: **5 / 6 / 4** | member's own column (**4 — a non-zero base, deliberately different from the let fixture's 0**), enclosing type, second body line's indent (8), callee names |
| type / subtype | DU cases' indent: **1 / 2 / 0** | type's column (0), case names, trailing sibling type, file path |
| type / hierarchy | `inherit` clause's indent: **1 / 2 / 0** | derived type's column (0), base type name, preceding declaration |
| CE builder | builder members' indent: **1 / 2 / 0** | builder type's column (0), member names, the `let optional = OptBuilder()` binding |
| record names | record brace line's indent: **1 / 2 / 0** | `Addr`'s column (0), field names, the consuming record `Person` (fields stay at their original columns) |
| widening | the **kind of text** on the base+1 line: comment / string literal / genuine code | the indent (always base+1), the declaration column, the token spelled inside each carrier |

The base-indent axis is the one most easily left constant by accident: the let, type, CE and
record fixtures sit at column 0 and the member fixture at column 4, so `baseIndentLen` is varied
across call sites rather than silently pinned at zero.

## Both sides of the boundary

Every fixture carries controls at `baseIndentLen` **and** `baseIndentLen+2`, not only the band:

- **base+2 (deeper neighbour)** — must behave exactly as base+1 now does. Asserted for the let
  `EndLine`+CALLS, the member `EndLine`+CALLS, the `Shape` subtype+`EndLine`, `Derived` EXTENDS,
  `Bind` subtype, and `Person VALIDATES -> Addr`.
- **base+0 (shallower neighbour)** — must still **terminate**, and this direction is asserted
  with **forbidden rows**, not just must-haves:
  - let: `alpha EndLine == 5`, and `alpha` CALLS **must not** contain `siblingOnly` (which
    belongs to `beta`), while `beta` CALLS must.
  - member: `Alpha EndLine == 4`, and `Alpha` CALLS **must not** contain `deeperStep`.
  - type/subtype: `Shape EndLine == 3`, `Shape.Circle` **must not** be a sub-entity, and the
    sibling `type Other` at column 0 must remain its own entity at `StartLine 7`.
  - type/hierarchy: an `inherit` at column 0 **must not** produce an EXTENDS edge.
  - CE builder: `Bind`/`Return` **must not** be `ce_member`, and `OptBuilder` **must not** be a
    `computation_builder`.
  - record names: `Person` VALIDATES **must not** contain `Addr`, and `Addr` **must not** be a
    `record`.

## The widening direction — scored with forbidden rows

Appending the base+1 line makes every affected body **larger**, so `collectCalls` and the edge
scanners see more text, and per #7152 those scanners read raw `src`.
`TestFSharp_IndentBand7176_WideningMintsNothingFromCommentsOrStrings` is forbidden-only on that:

- let body: base+1 comment `// ghostOne ()` and base+1 string `printfn "ghostStr ()"` — `alpha`
  CALLS must contain `realTwo` and **must not** contain `ghostOne` or `ghostStr`.
- type body: base+1 `inherit RealBase()` plus a base+1 comment `// inherit GhostA()` and a
  base+1 string `member _.Show () = "inherit GhostB()"` — `Derived` EXTENDS must be **exactly**
  `[RealBase]`.

Measured, not assumed: the same ghost carriers at **base+2** already mint nothing today
(`collectCalls` and `collectHierarchyEdges` both scrub), so the widening does not open a new
minting class — it only admits more lines to an already-scrubbing scanner. Both are pinned
anyway, because the fix is what puts those lines in front of the scanner for the first time. An
independent reviewer re-attacked this with their own carriers (a base+1 `// { Fake : string }`
comment and a base+1 `"{ Faker : string }"` string in a type body) and minted nothing.

**Bounded blast radius, measured:** because the old code *skipped and continued*, lines after
the swallowed one were already in the body. The only text newly admitted is the base+1 lines
themselves. `EndLine` therefore moves by exactly the number of swallowed lines (+1 in the let
and member fixtures, +2 in the DU fixture).

## Reachability

The fix makes **no** currently-unreachable line reachable. It goes the other way: with
`minBodyIndent = baseIndentLen + 1` the two branches `indent >= minBodyIndent` and
`indent <= baseIndentLen` are **exhaustive over the integers**, so the implicit fall-through —
the dead band itself — becomes unreachable. That removal is what the whole PR asserts.

## The hazard bullet I deleted, and why it is back

An earlier revision of this PR removed the `collectHierarchyEdges` bullet that said an
off-by-one-indented sibling type *"can therefore fall inside the PREVIOUS type's body and have
its `inherit` clause attributed to the wrong owner"*, on the stated grounds that it was now
fixed and graded. **It was neither.** Measured by me on this fixture:

```fsharp
type Base () =
    member _.M () = ()

type Alpha () =
    member _.A () = ()
 type Beta () =
     inherit Base ()
```

| artefact | `+2` (pre-fix) | `+1` (this PR) |
|---|---|---|
| `Alpha` EXTENDS | **`[Base]`** ← spurious, belongs to `Beta` | **`[Base]`** ← unchanged |
| `Beta` EXTENDS | `[Base]` | `[Base]` |
| `Alpha` EndLine | 7 | **8** |

The mis-attribution is **byte-identical before and after**. The dead band was never the cause:
`inherit` sits at column 5 and was already inside `Alpha`'s body under the old `+2` threshold.
The real cause is that `collectHierarchyEdges` regex-scans the whole body with **no notion of a
nested declaration**. The PR's `TypePass_Hierarchy` row pins a genuinely different case (an
`inherit` clause itself at base+1 inside its **own** type's body).

So the old note was right about the **symptom** and wrong about the **mechanism**. The bullet is
**restored**, re-scoped to name the flat scan as the cause and to point at **#7187**, which
tracks it. Deleting a true warning because its stated cause turned out to be wrong is worse than
leaving it — the next reader loses the hazard entirely. **#7187 is not fixed here.**

## Mutants — re-scored from scratch at `75184ddd0`, not carried over

Harness: byte-level `cp` backup, exact match-count assertion **before** running, the mutated
line **read back** verbatim, `go vet` exit captured to its **own file** (never through a pipe),
`--- FAIL` counted with an **anchored** `grep -cE '^\s*--- FAIL'`, restore by `cp` + `shasum`
verification + marker-count-0 grep. Never `git checkout --`.

|---|---|---|---|---|
| M1 | the fix reverted (`baseIndentLen + 2`) | 0 | **7** | **DEAD** |
| M2 | **positive control** — M1 with `indent_band_7176_test.go` held out | 0 | **0** | **ALIVE 0** |
| M3 | off-by-one the **other** way (`baseIndentLen + 0`) | 0 | **17** | **DEAD** |
| M4 | terminator arm: `indent <= baseIndentLen` → `indent < baseIndentLen` | 0 | **14** | **DEAD** |
| M5 | terminator arm: drop `&& strings.TrimSpace(line) != ""` | 0 | **0** | **ALIVE — equivalent, see below** |

**M1 decomposition (7 = 7, all mine):** `LetPass`, `MemberPass`, `TypePass_SubtypeAndContains`,
`TypePass_Hierarchy`, `WideningMintsNothingFromCommentsOrStrings`, `CEBuilderPass`,
`RecordTypeNamesPass`. All **five** declaration passes fail separately — that is the
per-call-site evidence.

**M2 — positive proof the kill is mine, not a pre-existing test's.** `go test -list '.*'`:
**169 tests with the file, 162 without**; the `-list` diff is exactly my 7 names and nothing
else. The M2 log contains **0** occurrences of the string `IndentBand7176`. (`=== RUN` lines
only appear under `-v`, so a zero `=== RUN` count would not have been evidence; the `-list`
diff is.) The 162 pre-existing tests are structurally blind to the dead band — which is exactly
why "covered transitively by the package suite" was not a coverage claim.

**M3 decomposition (17 = 6 mine + 11 pre-existing):** mine — `LetPass`, `MemberPass`,
`TypePass_SubtypeAndContains`, `TypePass_Hierarchy`, `CEBuilderPass`, `RecordTypeNamesPass`.
Pre-existing — 7 × `TestFSharpFieldTypeRefs_*`, `TestFSharp_TypeAlias`,
`TestFSharp_InheritEmitsExtends`, `TestFSharp_InheritVariants`,
`TestFSharp_HierarchyAnchoredOnTypeNotFile`. My **base+0 controls** are what kill it on my side;
this is the row proving the fixtures pin the **exact** boundary, not merely "some boundary".
**Declared: `Widening` does not fire under M3** — its fixture has no base+0 line.

**M4 decomposition (14 = 4 mine + 10 pre-existing):** mine — `LetPass`, `MemberPass`,
`TypePass_SubtypeAndContains`, `RecordTypeNamesPass`. Pre-existing — the same 7 `FieldTypeRefs`
plus the 3 `Inherit`/`Hierarchy` tests. **Declared: `TypePass_Hierarchy`, `CEBuilderPass` and
`Widening` do NOT fire under M4** — stated symmetrically with M3, per review item 4.
*Confound declared:* M3 and M4 share most of their kill list because both shift where a line at
column `base` lands. They are nonetheless distinct expressions and were scored separately, per
the compound-guard rule.

**M5 — ALIVE and genuinely EQUIVALENT, left untested on purpose.** The argument is **syntactic
dominance**, which is stronger than the fuzz and is what the verdict now rests on: reaching the
`else if` requires the immediately preceding `if strings.TrimSpace(line) == "" { …; continue }`
to have been false, and the `i == 0` arm also `continue`s, so **every** line reaching that
branch has already been shown non-blank. `strings.TrimSpace(line) != ""` is therefore a
tautology at that program point and the conjunct can never be false. Belt-and-braces only: a
side-by-side harness ran both variants over **200,000** randomly generated inputs from a
14-element whitespace/token alphabet (1–6 lines, `baseIndentLen` 0–3, every `afterPos`) —
**0 differing outputs**. An independent reviewer reached the same dominance argument and also
found no distinguishing input. This is a **pre-existing** dead conjunct, not something this PR
introduces, and it is deliberately left in place and untested rather than removed here.

## Gates

- `go test -count=1 ./internal/extractors/fsharp/` — **ok**, 0 `--- FAIL` (169 tests).
- **`go test -count=1 ./cmd/grafel/` with fresh `GRAFEL_HOME` + `GRAFEL_DAEMON_ROOT`** — **ok**,
  165.4s, exit 0, 0 `--- FAIL` (re-run at `75184ddd0`). **Digest UNMOVED**, as expected:
  `writeSpanFixture6118` has 18 members and **zero** `.fs`/`.fsx`/`.fsi`, so an F#-only
  extractor change cannot reach it. Nothing bumped.
- `go test -count=1` on the dependent packages — `./internal/extractor/`, `./internal/resolve/`,
  `./internal/engine/`, `./internal/extractors/cross/...` — **ok**, 0 `--- FAIL`.
- `gofmt -l internal/extractors/fsharp/` — clean.

## What I did not do

- Did not fix **#7187** (the `collectHierarchyEdges` nesting-blind scan). Out of scope here; the
  restored bullet points at it.
- Did not touch **#7144**. Its four arms rest on extent-based containment through this helper,
  which is why this went first.
- Did not widen to `nim` / `rescript` — that is **#7185**, and the `+1` boundary is an
  F#-offside argument that does not transfer.
- Did not measure **corpus incidence**. The old comment's "low frequency" remains an unmeasured
  adjective; I did not verify or refute it.
- Did not remove the dead `&& strings.TrimSpace(line) != ""` conjunct (M5) — out of scope here.

Closes #7176


cajasmota:
## Independent adversarial review — #7184

Own worktree at `origin/wt-7176-a1` (`9ffa555bc`), detached, restored by `cp` + shasum after every
mutant (`ba86b4f719dc7c66be812107b6726a93d89d5066`, `git status` empty at the end). I did not write
this code.

### What I re-derived and CONFIRMS

**The spec citation — verified verbatim, independently.** I fetched
`fsharp.org/specs/language-spec/4.1/FSharpSpec-4.1-latest.pdf` myself and ran my own
`pdftotext -layout`. Both quotes are verbatim and correctly read:

- §15.1.4 lists what introduces an offside line, including **"The start of a `let`, `if` or `module`
  token."** So the governing offside line for a declaration *is* its own column = `baseIndentLen`.
  That is exactly the question I was asked to break, and it comes out the author's way.
- §15.1.8 verbatim: *"When a token occurs on or before the offside limit for the current offside
  stack, and a permitted undentation does not apply, enclosing contexts are closed."* On-or-before
  closes, strictly greater continues.
- The FS0058 worked example is reproduced accurately, warning on `let z = 3` at column 0, offside of
  *"context at position (2:1)"* — i.e. the context the one-column-deeper `let y` pushed.

**A stronger citation the PR did not use, from the same §15.1.8:** *"When a token other than `and`
appears **directly on the offside line** of Let context … the `$in` token is inserted."* A sibling
binding sits *directly on* the offside line — at exactly `base`, not `base+1` — and `and` is the one
token explicitly exempted. Worth adding to the comment; it closes the falsifier directly.

**Falsifier — I tried and could NOT construct one.** Candidates: `and`-continued `let rec` (§15.1.8
names `and` as the exception, and `let … and …` is one declaration, not a sibling); `|` match arms
(§15.1.4's second example makes a shallower `|` a *syntax error*, not a sibling); `type … with`
members and CE bodies (both push contexts strictly deeper). The permitted-undentation rules run the
*opposite* direction (a token left of the offside line that still continues) — that would make this
scanner terminate **early**, which is a separate pre-existing limitation, not something this PR
introduces. **No falsifier found.**

**M5 is genuinely EQUIVALENT — by dominance, not fuzzing.** I read the loop. Reaching the `else if`
requires the immediately preceding `if strings.TrimSpace(line) == "" { …; continue }` to have been
false, so `strings.TrimSpace(line) != ""` is a *syntactic* tautology at that point; the `i == 0`
arm also `continue`s. Every blank line — including a blank at `i == 0` — is appended and `continue`d
before that branch. This is airtight without the 200,000 inputs. I ran the mutant anyway: vet 0,
**0 `--- FAIL`**, ALIVE. **Equivalence verdict upheld.**

**Mutant decompositions re-derived, and every number is exact:**

| mutant | my anchored `grep -cE '^\s*--- FAIL'` | names |
|---|---|---|
| M1 (`+2`) | **5** | exactly the author's 5 |
| M2 (M1, file held out) | **0** | `-list` 167 → 162; diff is **exactly** the 5 `IndentBand7176` names and nothing else; `grep -c IndentBand7176` on the M2 log = **0** |
| M3 (`+0`) | **15** | 4 mine + 11 pre-existing — names match the body line for line; sub-counts add up |
| M4 (`<`) | **13** | 3 mine + 10 pre-existing — `TypePass_Hierarchy` does **not** fire, exactly as declared |

Digest claim spot-checked without the 170s run: `writeSpanFixture6118` contains **zero**
`.fs`/`.fsx`/`.fsi` members, so an F#-only extractor change cannot reach it. I did **not**
independently run `./cmd/grafel/`.

`go test -count=1 ./internal/extractors/fsharp/` at HEAD: exit 0, 0 `--- FAIL`. `gofmt -l` clean.

---

### FINDING 1 (blocking) — the deleted hazard bullet describes a case this PR does NOT fix

The removed bullet said, verbatim:

> an off-by-one-indented sibling type can therefore fall inside the PREVIOUS type's body and have its
> `inherit` clause attributed to the wrong owner

It was removed on the stated grounds that *"it is now fixed and graded"*. **It is neither.** Measured
on a fixture nobody in this PR has:

```fsharp
type Base () =
    member _.M () = ()

type Alpha () =
    member _.A () = ()
 type Beta () =
     inherit Base ()
```

| artefact | `+2` (pre-fix) | `+1` (this PR) |
|---|---|---|
| `Alpha` EXTENDS | **`[Base]`** ← spurious, belongs to `Beta` | **`[Base]`** ← unchanged |
| `Beta` EXTENDS | `[Base]` | `[Base]` |
| `Alpha` EndLine | 7 | 8 |

The mis-attribution is **identical before and after**. It survives because `collectHierarchyEdges`
takes the whole body and regex-scans it for any `inherit`, with no notion of a nested type
declaration — the `inherit` at column 5 was *already* inside `Alpha`'s body under `+2` (5 ≥ 2), so
the dead band was never what caused it. What the fix changes here is only `EndLine`.

The PR's `TypePass_Hierarchy` fixture pins a *different* case (an `inherit` clause itself at base+1
inside its own type's body), which is genuinely fixed. That row does not cover the bullet's case.

**Asked:** either restore the bullet (re-scoped to name `collectHierarchyEdges`' flat scan as the
cause rather than the dead band), or file it with this fixture and reference the issue from the
comment. Deleting a documented hazard that is still live is the exact failure mode #7176 was opened
about.

### FINDING 2 (blocking) — both "transitively covered" call sites move, with artefacts nothing asserts

I probed `compexpr_active_patterns.go:517` and `validators.go:344` directly by diffing emitted
records under `+1` vs `+2`. Both move, and the artefacts are **distinct in kind** from every row in
the PR:

**`collectCEBuilderTypes` (`:517`)** — builder members at base+1:

| artefact | `+2` | `+1` |
|---|---|---|
| `OptBuilder` Subtype | `type` | **`computation_builder`** |
| `Bind` Subtype | `member` | **`ce_member`** |
| `Return` Subtype | `member` | **`ce_member`** |

**`collectRecordTypeNames` (`:344`)** — record braces at base+1:

| artefact | `+2` | `+1` |
|---|---|---|
| `Addr` Subtype | `type` | **`record`** |
| `Person` relationships | *(no `VALIDATES`)* | **`VALIDATES -> Addr`** |

M2 already proves the 162 pre-existing tests are blind to this boundary, so "covered transitively by
the package suite" is covered by **nothing**. `computation_builder` / `ce_member` and the
nested-model `VALIDATES` edge are ungraded at base+1 today. Two rows would do it.

Note the PR is *right* on both — this is a coverage gap, not a defect. But "a verdict at three call
sites says nothing about the other two" is the lesson this repo has measured five times, and it is
cheap to close here.

### What I could NOT verify

- No F# toolchain, so the offside reading stays **derived-not-executed** — I confirmed the *text*,
  not a compiler's behaviour. `dotnet`/`fsc`/`fsi` absent, as stated.
- I did not run `./cmd/grafel/` (one build at a time on this host); the digest claim rests on my
  fixture-membership grep plus the author's green run.
- Corpus incidence remains unmeasured, as the PR says.

### Things I tried that did NOT land

- **Sibling-declaration loss.** I expected a wider body to swallow a sibling entity. It cannot: all
  five call sites enumerate declarations with whole-file `FindAllStringSubmatchIndex` over `src`, not
  by advancing past a body. No entity can be lost.
- **Widening via raw `src`.** My own carriers — a base+1 `// { Fake : string }` comment and a base+1
  `"{ Faker : string }"` string in a type body — minted nothing (`Ghost` stayed `type`). The PR's
  forbidden rows hold up under an independent fixture.
- **M5 counter-example hunt.** The dominance argument is unbreakable; no input distinguishes them.

---

## REQUEST-CHANGES

1. **Blocking — FINDING 1.** The deleted hazard bullet's own scenario is unfixed and unchanged by
   this PR (measured, table above). Restore it re-scoped to `collectHierarchyEdges`' flat scan, or
   file it with the fixture and cite the issue in the comment.
2. **Blocking — FINDING 2.** Add one row each for `:517` (`computation_builder` / `ce_member`
   subtype at base+1) and `:344` (`Person VALIDATES -> Addr` at base+1). Both demonstrably move and
   nothing asserts them; "transitively covered" is false, per M2.
3. **Non-blocking — strengthen the citation.** §15.1.8's *"a token other than `and` appears directly
   on the offside line of Let context"* says a sibling sits at **exactly** `base` and names `and` as
   the sole exception. That is a sharper source for the decision than the FS0058 example, and it
   closes the stated falsifier directly. Worth one sentence in the in-source comment.
4. **Non-blocking.** M4's decomposition does not say in the body that `TypePass_Hierarchy` survives
   it; it does (13 = 3 + 10, verified). Declaring which of your own rows *don't* fire, as you did for
   `Widening` under M3, would make it symmetric.

Everything else — the spec reading, all five mutant verdicts, the M2 positive control, both
decompositions, the widening forbidden rows, the digest argument — I attacked and could not break.

PR #7186, opened 2026-09-18
test(#7183): retryInRe silently DROPPED a negative retry_in; the harness leaned on a count control in another function

### LINKED ISSUE #7183: internal/daemon tests: retryInRe is anchored on 'retry_in=', so a NEGATIVE announced duration matches NOTHING and is silently DROPPED from the wait slice — caught only by a count control in another function (title corrected: it does NOT parse as the positive magnitude)
## What

`retryInRe` — the helper that parses `retry_in=<d>` out of the daemon's log stream for the `acceptLoop` backoff pins — has a character class that excludes `-`. A **negative** announced duration therefore parses as its **positive magnitude**, and would *satisfy* a bound assertion instead of failing it.

`internal/daemon/acceptloop_backoff_factor_test.go` (line ~228 on `5164e259a`; the class is `[0-9a-zA-Z.]`, which admits neither `-` nor `µ` (U+00B5)).

Found by the independent adversarial reviewer on PR #7182 while attacking that PR's mutant table.

## Why this is the direction that matters

Every backoff pin in this package is a **bound** assertion — "every announced wait is `<= acceptBackoffMax`". A parser that strips the sign converts a value that should blow the bound into one that clears it. The failure mode is **a green test over broken behaviour**, which is strictly worse than a noisy one.

Compare the sibling hole, which is benign in the opposite way: the same class excludes `µ`, so a microsecond-scale duration fails to parse and the helper `t.Fatalf`s. That is **fail-loud** — the author of #7182 hit it while scoring a `5ms -> 5µs` mutant and correctly discarded the row as a harness artefact rather than counting it. The `-` case has no such protection.

## Not claimed

- **Not reachable today.** Producing a negative `time.Duration` here needs int64 overflow of the doubling — `backoff *= 2` past `math.MaxInt64` — and the clamp at `acceptBackoffMax` (1s) prevents the sequence ever getting near it. This is **latent**, and it is filed for the fail-dangerous *direction*, not for a live defect.
- **Not caused by #7182 or #7172.** `retryInRe` predates both; those PRs consume it. Nothing in either introduced or worsened this.
- **Not claimed that the `µ` exclusion is a defect.** It is fail-loud and arguably fine. Fixing the sign should not be an excuse to silently widen the class — see below.
- **Incidence uncounted.** No corpus or CI evidence that a negative duration has ever been announced.

## Why it is still worth fixing rather than closing as unreachable

The guard that makes it unreachable is the clamp — and the clamp is *the thing these tests exist to grade*. A test harness whose correctness depends on the production behaviour it is asserting is circular: mutate the clamp away (which is exactly what #7182's mutants do) and the harness's safety margin goes with it. The cost here is one character class.

## Grading requirement

- **A must-have row** asserting the helper REJECTS (or correctly signs) a negative `retry_in`, rather than returning its magnitude. Assert the emitted artefact — the parsed value or the failure — not a counter about the parse.
- **Preserve the fail-loud property for unparseable units.** If the fix widens the class, a `µs` duration must still be caught rather than silently mis-parsed; that is the `-` defect reintroduced under a new name. Score BOTH: a rejected `µ` and a rejected `-`.
- **Score the widening direction.** Broadening a character class is where this repo's serious defects come from. A forbidden row proving the parser does not accept something that is not a duration at all.
- **Do not let the fix depend on the clamp.** The point of the issue is that the harness should be sound independently of the behaviour it grades.

Related: #7182 / #7174 (the `acceptLoop` clamp pin that consumes this helper), #7172 / #7167 (the factor pins, same helper), #7126, #6421.

cajasmota:
## What

`retryInRe` — the helper that turns the daemon's announced `retry_in=<d>` into a
`time.Duration` for every backoff pin in `./internal/daemon/` — used the character class
`[0-9a-zA-Z.]`, which admits neither `-` nor `µ` (U+00B5). This PR replaces the class with
`\S+` (the whole whitespace-delimited token, `time.ParseDuration` as the sole arbiter), adds
an explicit rejection of non-positive announcements, and grades all of it with a 19-row table
test.

**Test-harness change only. No production file is touched.** The diff is two `_test.go` files.
Because it stays inside `internal/daemon/` and touches no extractor, entity kind or digest input,
`go test ./cmd/grafel/` is not required and was not run.

## The real justification: the old class MIS-VALUED, it did not merely drop

The issue's framing (and my own first one) was "a negative announcement is lost". The adversarial
review on this PR measured something strictly stronger over a 400,000-case fuzz plus a 49-token
hand table:

| | old `[0-9a-zA-Z.]+` | new `\S+` |
|---|---|---|
| capture ≠ whole token | 6,077 / 400,000 | **0 / 400,000** |
| **mis-valued** (accepted as a plausible-but-wrong duration) | **6,077** — 100% of its accept-where-new-rejects cases | **0** |

Worked examples of the old behaviour: `1h-2m` → `1h0m0s`, `5m-s` → `5m0s`, `5ms ` → `5ms`. Each is
a silent truncation into a *different, plausible* duration — the fail-dangerous direction, at scale
and reachable by ordinary text. `\S+` mis-values nothing **by construction**: the capture is always
the entire whitespace-delimited token, so shaving is impossible and every widening resolves to
*rejection*, never to a wrong value.

That retires the issue's framing. For completeness, measured on `b31fe2c67`: `retry_in=-5ms` gave
the old regex **zero matches** (the literal `retry_in=` never completes, so the announcement was
dropped from the slice) — not, as #7183's body states, a match on the positive magnitude. The only
thing catching it was `announcedAcceptBackoffs`' `len(waits) != n` count control, one assertion
away in another function.

## Rejection vs signed parsing — what was chosen and why

**Both, in sequence.** The regex captures the token verbatim, `ParseDuration` reads it with its
sign intact, and the parser then **rejects** any wait `<= 0`, returning a typed
`*nonPositiveRetryInError` carrying the value it read. The rows assert that carried value
(`np.Parsed == -5*time.Millisecond`), not an error string and not a counter — so "correctly signed,
then rejected" is distinguishable from "rejected having read the magnitude" (mutant M4).

**Rejecting a zero wait is provably correct, not a judgement call.** In `server.go:983`
`if backoff == 0 { backoff = acceptBackoffStart }` runs **before** the log line, and `backoff == 0`
is the first-error sentinel — so the announced value is structurally always `>= acceptBackoffStart`
(5ms) and production **cannot** legitimately announce `retry_in=0s`. `d <= 0` therefore rejects
nothing valid. Scope note for anyone tempted to relax it later: the only other production
`retry_in` emitter is `internal/daemon/sched/stagegate.go:458`, in a different package that this
helper never reads.

## Independence from the clamp

#7183's central point is that the only reason a negative wait is unreachable today is
`acceptBackoffMax`, and the clamp is what these tests mutate away. The rule added here is a
**lower** bound (`> 0`); every pin in this package asserts an **upper** bound
(`<= acceptBackoffMax`) or a ratio. The review measured this rather than arguing it: with the
three-line clamp deleted from `server.go`, `TestParseAnnouncedRetryIns` still passes, and with the
clamp **and** the `d <= 0` block both deleted, four rows fail. `parseAnnouncedRetryIns` references
only `retryInRe`, `time.ParseDuration` and its own error type.

## The µ decision, and the full alphabet of what `\S+` newly admits

`5µs` is **parsed at its exact value**, not rejected. #7183 asks that the fail-loud property be
preserved and names the mode to avoid as *silent mis-parsing*; faithful parsing satisfies that more
strongly than the old path, which silently shortened the token to `"5"` and only then happened to
be unparseable.

`\S+` newly admits exactly two classes, and **both are graded**:

- **leading sign** — `+5ms` → `5ms` (row), `-5ms` → rejected with `Parsed: -5ms` (row)
- **micro sign** — U+00B5 `5µs` → `5µs` (row) and U+03BC `5μs` → `5µs` (row). `ParseDuration`
  accepts both spellings, so pinning only U+00B5 would leave half the admitted alphabet ungraded.

Newly-admitted-and-ungraded is the accept-direction hole the forbidden rows cannot see, which is
where this repo's serious defects come from; the two new `wantOK` rows close it, and `+5ms` and
`5μs` are both extra killers for a revert of the regex arm (M2 went 8 → 10).

**Varied**
- sign of the announced duration: positive, **explicit `+`**, negative, zero
- unit spelling: `ms`, U+00B5 `µs`, **U+03BC `μs`**, composite `1h2m3s`
- position of the offending value in the stream: only value, and middle of three
- token well-formedness: valid duration, no unit, no magnitude, unknown unit, wrong case,
  scientific notation, lone sign, doubled sign, trailing junk
- number of announcements in one stream: 0, 1, 3

**Held constant**
- the log format (logfmt `key=value`, space-delimited, produced by `slog`'s `TextHandler`)
- the production code — `server.go`, the clamp, the factor and the constants are untouched
- `announcedAcceptBackoffs`' own count control (`len(waits) != n`), left exactly as it was, so
  these rows do not share a killer with it
- the existing factor and clamp pins (#7182 / #7172) — their mutant tables stand; the scoring scope
  below is the whole package, so any collateral would have shown

## Mutants

Every row **re-scored from scratch** after the two new rows were added — no count is carried over
from the first round. Scope: `go test -count=1 -timeout 1800s ./internal/daemon/`, the whole
package. Each mutant applied with an asserted exact match count, mutated line read back, `go vet`
exit captured to its own file (never through a pipe), `--- FAIL` counted with
`grep -cE '^\s*--- FAIL'`. Restore by `cp` + `shasum` equality + a marker grep asserting count 0.

| # | mutant (exact edit, match count asserted) | vet | `--- FAIL` | verdict |
|---|---|---|---|---|
| M1 | whole fix reverted: regex → `[0-9a-zA-Z.]+` (1) **and** the `if d <= 0` block deleted (1) | 0 | 11 | DEAD |
| M1-pc | **positive control**: M1 exactly, with `acceptloop_retry_in_parse_test.go` moved out of the package | 0 | **0**, `test_exit=0` | **ALIVE** |
| M2 | regex arm only: `\S+` → `[0-9a-zA-Z.]+` (1), positivity kept | 0 | 10 | DEAD |
| M3 | positivity arm only: the `if d <= 0` block deleted (1), regex kept | 0 | 5 | DEAD |
| M4 | rejects, but carrying the **magnitude**: `Parsed: d` → `Parsed: -d` (1) | 0 | 4 | DEAD |
| M5 | compound arm split: `d <= 0` → `d < 0` (1) | 0 | 2 | DEAD |
| M6 | widening: `\S+` → `.+` (1) | 0 | 2 | DEAD |

**Positive control, proved by a `-list` diff, not a `=== RUN` count.** `go test -list '.*'
./internal/daemon/` captured with and without the new file differs by exactly one line
(`5d4 < TestParseAnnouncedRetryIns`; 491 → 490 `Test*`). With the fix reverted and only that test
held out, the package is green: `test_exit=0`, 0 anchored `--- FAIL`. A zero `=== RUN` count would
have proved nothing — `=== RUN` is only emitted under `-v` and these legs were not verbose.

### Decompositions (each sums)

- **M1 = 11** = 1 parent + 10 subtests: `µs`-exact, **`+5ms`**, **U+03BC `μs`**, negative-signed,
  negative-in-the-middle, negative-µs, zero-rejected, lone-sign, doubled-sign, trailing-junk.
- **M2 = 10** = 1 parent + 9 subtests — M1's ten **minus** zero-rejected, which survives because M2
  keeps the positivity check. **11 = 10 + 1.** The two new rows are two of M2's killers, as the
  review predicted.
- **M3 = 5** = 1 parent + 4 subtests: negative-signed, negative-in-the-middle, negative-µs,
  zero-rejected.
- **M2 and M3 are overlapping arms with distinct unique killers — not mirror images.** They
  **overlap on the three negative rows**. M2-unique = **6** (µs-exact, `+5ms`, U+03BC `μs`,
  lone-sign, doubled-sign, trailing-junk); M3-unique = **1** (zero-rejected). 6 + 1 + 3 overlap = 10
  = M2's subtest count; 1 + 3 = 4 = M3's. Each arm therefore has killers the other does not, so
  neither is masked by the other — that conclusion stands; the earlier "mirror image" wording
  claimed a disjointness that does not hold and is corrected here.
- **M4 = 4** = 1 parent + the 3 rows asserting a non-zero `np.Parsed`. Zero-rejected survives
  (`-0 == 0`), which is the isolation: M4 probes the **sign**, not the rejection. The mutant still
  rejects, still returns `*nonPositiveRetryInError`, still returns no waits, and is killed **only**
  by the carried value's sign.
- **M5 = 2** = 1 parent + zero-rejected alone. The `<` / `<=` arms are graded independently.
- **M6 = 2** = 1 parent + the single row whose log line carries a trailing field
  (`… retry_in=5ms err=EMFILE`), which `.+` swallows whole into an unparseable token.
  **Declared confound:** Go's `.` does not match `\n`, so rows whose fixture is a bare
  `retry_in=…\n` cannot distinguish `.+` from `\S+`. The one row that can, does — deliberately the
  realistic one, since a real `slog` line always carries neighbouring fields.

### Proof the fail-loud `µ` path survives

- The U+00B5 row (`retry_in=5µs` → exactly `5 * time.Microsecond`) is killed by M1 and M2, so it is
  genuinely graded; pre-fix the same input captured `"5"` and errored. The U+03BC row is killed by
  the same two.
- Loudness is unchanged in kind and strictly wider in coverage: anything `ParseDuration` refuses
  still fails the test, and it now sees the **whole** token, so no byte can shorten a value into a
  different plausible one — 0 mis-valuations over 400,000 fuzzed inputs, against 6,077 for the old
  class. Eight forbidden rows pin that direction: `5`, `banana`, `ms`, `1e3ms`, `5MS`, `-`, `--5ms`,
  and `5ms,` (trailing junk makes the token unparseable rather than being shaved to a clean-looking
  `5ms` — which the old class did).
- The `-5µs` row exercises both holes in one input and is killed by M1, M2, M3 and M4.

## What I did not do

- Did not widen further; `\S+` defers the whole decision to `time.ParseDuration` rather than
  guessing which bytes look duration-ish.
- Did not touch the clamp, the factor pins, `acceptBackoffMax`, or `server.go`.
- Did not re-litigate #7182's or #7172's rows.
- Did not run `./cmd/grafel/` — two `_test.go` files, no extractor.
- Did not claim reachability: a negative announcement is still unreachable in production today.
  This is a harness-soundness fix, filed and fixed for the direction of the failure.
- Not pinned, and recorded as a known gap rather than fixed: `-0s` → `Parsed: 0` (rejected, but by
  the zero arm rather than the sign arm) has no row of its own.
- All legs ran on darwin/arm64; no other platform was measured locally.

Closes #7183


cajasmota:
## Adversarial review — #7186

Independent checkout at `/Users/jorgecajas/Projects/archigraph-wt-7186-rev` (detached `origin/wt-7183-a1`, `ece2de303`). Every mutant applied by asserted exact match count, mutated line read back, `go vet` exit captured to its own file, `--- FAIL` counted with `grep -cE '^[[:space:]]*--- FAIL'`, restore by `cp` + `shasum` equality + clean `git status --porcelain`. Scope for every leg: `go test -count=1 -timeout 1800s ./internal/daemon/` (whole package). Baseline: `ok ... 52.2s`, `base_exit=0`, **0** anchored FAIL lines.

### 1. Every mutant verdict reproduces, and the rows are the ones claimed

| # | claimed | measured | killing subtests (measured) |
|---|---|---|---|
| M1 full revert | 9 | **9** | µs-exact, neg-signed, neg-middle, neg-µs, zero, lone-sign, doubled-sign, trailing-junk (+parent) |
| M1-pc hold-out | ALIVE 0 | **0**, `test_exit=0` | — |
| M2 regex arm | 8 | **8** | M1's eight **minus** zero (+parent) |
| M3 positivity arm | 5 | **5** | neg-signed, neg-middle, neg-µs, zero (+parent) |
| M4 `Parsed: d`→`-d` | 4 | **4** | **exactly** the three non-zero `np.Parsed` rows (+parent) |
| M5 `<=`→`<` | 2 | **2** | zero alone (+parent) |

All `vet_exit=0`. **M4's kills are exactly the three rows that assert a non-zero `np.Parsed`, and zero survives** — "the rows pin the SIGN" is earned, not asserted. Decomposition sums: **9 = 8 + 1** confirmed by row identity, not just by count.

**Hold-out, independently derived.** `go test -list '.*'` with vs without the file differs by exactly one line, `5d4 < TestParseAnnouncedRetryIns`. My raw counts are 494→493 lines; 3 of those are `Fuzz`/`Benchmark`/`Example`, so **491→490 `Test*`** — the PR's figure is right.

### 2. `\S+` — what it newly admits (attack 1). Nothing is mis-valued.

400,000-case fuzz over an alphabet of digits, unit letters, `µ`/`μ`/fullwidth digits, `. - + , ; " ' ( ) [ ] { } \ / * _ =`, space and tab, plus a 49-token hand table:

- **Capture mismatch: 0/400,000.** The capture is *always* the whole whitespace-delimited token. Shaving is impossible by construction, so `ParseDuration` really is the sole arbiter.
- **NEW-MISVALUED: 0.** No input is parsed into a plausible-but-wrong duration. Every widening lands on *reject*, never on *wrong value*. `5-ms`, `5ms-`, `5ms,`, `5ms;`, `5ms/2`, `1h-2m`, `5m-s`, `5ms"`, `5ms ` all → error.
- **Old regex mis-valued 6,077/400,000** (100% of its "accept where new rejects"). E.g. `1h-2m` → old `1h0m0s`; `5m-s` → old `5m0s`; `5ms ` → old `5ms`. That is the defect, measured at scale.
- **The new-admission set is exactly two classes:** tokens with a leading sign (`+`/`-`) and tokens containing a micro sign. 122 new accepts in the fuzz, all correctly valued. `+5ms`→5ms, `-5ms`→-5ms, `5µs`(U+00B5)→5µs, `5μs`(U+03BC)→5µs, `-0s`→0.

### 3. Findings (neither blocks)

**F1 — two newly-admitted, correctly-valued inputs are ungraded.** `+5ms` and `5μs` (U+03BC GREEK SMALL LETTER MU, which `ParseDuration` accepts alongside U+00B5) are both *new accepts* under `\S+` with no row. Under the old class both yield 0 matches, so a `+5ms` `wantOK` row would be an **additional M2 killer** — cheap, and it closes the accept-direction half of the alphabet the eight forbidden rows leave open (the `sealed`-ALIVE pattern). `-0s` → `Parsed: 0` rejected is likewise unpinned.

**F2 — "mirror image" overstates disjointness.** M2 and M3 **overlap on the three negative rows**. M2-unique = 4 (µs, lone sign, doubled sign, trailing junk); M3-unique = 1 (zero). Each arm still has unique killers, so neither is masked — the conclusion stands, the word does not.

### 4. Zero-wait rejection is correct (attack 3)

`server.go:983` — `if backoff == 0 { backoff = acceptBackoffStart }` runs **before** the log line, and `backoff == 0` is the first-error sentinel. The announced value is structurally always `>= acceptBackoffStart` (5ms). **Production cannot legitimately announce `retry_in=0s`**, so `d <= 0` rejects nothing valid. Scope note: the only other production `retry_in` emitter is `internal/daemon/sched/stagegate.go:458` (`s.cfg.StageGateRetry`), a different package this helper never reads.

### 5. Clamp independence — deleted it and measured (attack 4)

Removed the three-line clamp from `server.go` (match count asserted 1, block read back absent, `vet_exit=0`):
- **A** clamp gone, fix intact → `TestParseAnnouncedRetryIns` **passes**, `exit=0`.
- **B** clamp gone **and** the `d <= 0` block gone → **4 rows fail** (three negatives + zero).

So the rule still holds and still fires with the clamp absent. `parseAnnouncedRetryIns` references only `retryInRe`, `time.ParseDuration` and its own error type — no upper bound, no `acceptBackoffMax`. `server.go` restored, shasum-equal, tree clean.

### 6. Count-control interaction — not weakened (attack 5)

Under `\S+` the **only** surviving drop mode is an empty value (`retry_in=` followed by whitespace/EOL) — measured NOMATCH under *both* regexes, and `slog` never renders a `time.Duration` empty. So `len(waits) != n` guards exactly what it did. Nor can a previously-passing test now fail: in `server.go` `retry_in` is the **last** attribute on the line, so `\S+` terminates at the `\n` — confirmed by the green whole-package baseline.

One recorded behaviour change, unreachable here: a `retry_in=` appearing inside a quoted `msg` field. Old captured `5ms` (a spurious extra wait, silently); new captures `5ms"` → hard failure. `announcedAcceptBackoffs` uses a fixed msg literal containing no such text, and the new direction is the loud one.

### What I could NOT verify

- Reachability of a negative announcement in production — the PR explicitly does not claim it, and I did not find a path.
- Behaviour on a non-macOS platform; all legs ran on darwin/arm64.
- `./cmd/grafel/` was **correctly not run**: the diff is two `_test.go` files inside `internal/daemon/` and no extractor, entity kind or digest input is touched. I confirmed this from the diff rather than by running it.

## APPROVE

1. All six mutants reproduce at the exact counts **and the exact rows** claimed; M4 isolates the sign, M5 isolates the zero arm, hold-out is 0 with a one-line `-list` diff.
2. The `\S+` widening is safe in the strong sense: 0/400,000 capture mismatches and **0 mis-valuations** — every widening resolves to rejection, and the old class mis-valued 6,077 of the same inputs.
3. Zero-rejection rejects nothing production can emit (`backoff == 0` sentinel precedes the log).
4. Clamp independence measured by deleting the clamp, not argued.
5. **F1 (non-blocking):** add a `+5ms` `wantOK` row (and optionally U+03BC `5μs`) — both are new accepts with no pin, and `+5ms` would be a free extra M2 killer.
6. **F2 (wording):** M2/M3 overlap on the three negative rows; "mirror image" should read "each arm has unique killers".

PR #7188, opened 2026-09-18
fix(#7187): a nested type's inherit clause was attributed to the OUTER type too — the hierarchy scan had no notion of nesting

### LINKED ISSUE #7187: fsharp: collectHierarchyEdges regex-scans a type's whole body with no notion of nesting, so a nested sibling's 'inherit' is attributed to the OUTER type — the dead band was never the cause, and the hazard comment blaming it was wrong about the mechanism
## What

`collectHierarchyEdges` regex-scans a type's whole body text for `inherit` clauses **with no notion of a nested declaration**, so a sibling type that falls inside another type's body has its `inherit` attributed to the wrong owner.

Measured by the independent adversarial reviewer on PR #7184, against both `b31fe2c67` and that PR's head:

```fsharp
type Alpha =
 type Beta =
     inherit Base
```

→ `Alpha EXTENDS [Base]` — **identically before and after** #7184's boundary fix.

## Why this is filed now, and why it is NOT the dead band

`extractIndentBody`'s dead band (#7176) carried an in-source comment claiming exactly this consequence: *"An off-by-one-indented sibling type can therefore fall inside the PREVIOUS type's body and have its `inherit` clause attributed to the wrong owner."* PR #7184 fixes the dead band and **deleted that bullet as now-fixed**.

The reviewer measured it: **the bullet describes a real, still-present defect, and the dead band was never its cause.** With the old `+2` boundary the `inherit` at column 5 was *already* inside `Alpha`'s body; the mis-attribution comes from the hierarchy scan not recognising the nested `type Beta` header, not from which lines the body-extractor collects. Changing the boundary moves *which* lines are in the body and does not give the scan any notion of ownership.

So the hazard comment was **correct about the symptom and wrong about the mechanism**, and removing it removed a true warning. #7184 is restoring it, corrected, and pointing here.

## Not claimed

- **Not claimed to be caused or worsened by #7184.** The verdict is byte-identical on both sides of that change.
- **Not claimed to be limited to `inherit`.** `collectHierarchyEdges` is the site measured, but the same "regex-scan the whole body, no nesting" shape may apply to the other body consumers — `collectCalls`, the CE-usage scan, match-site edges, `classifyTypeSubtype`. **Unmeasured. Check each rather than assuming**; a verdict at one body consumer says nothing about the others, which this repo has measured six times this week.
- **No corpus incidence.** Whether real F# nests a `type` inside another type's textual body at a shallower-than-expected indent is uncounted. The construct in the probe is unusual; whether the *general* case (any nested declaration whose body is scanned by an outer owner) is common is the question that decides priority, and it is open.
- **Legality derived-not-executed.** No F# toolchain here (`dotnet`, `fsc`, `fsharpc`, `fsi`, `mono` all absent; `javac` is the only compiler). Cite the spec for any nesting shape a fixture relies on.

## Grading requirement

- **A must-have row** asserting `Beta EXTENDS [Base]` and a **forbidden row** asserting `Alpha` does NOT. Both halves, separately — they can pass for opposite reasons, and a count assertion sees neither.
- **The forbidden name must be unique in the fixture** — F# dedup keys are name-keyed (#7144, #7152), so an entity sharing a name merges silently and is invisible to both a presence check and a count check.
- **Score the other body consumers** in the same PR or say explicitly which were left: `collectCalls`, CE-usage, match-site edges, `classifyTypeSubtype`. If the fix teaches only the hierarchy scan about nesting, the others keep the defect and the PR should say so rather than implying a general fix.
- **Do not regress #7184's boundary rows.** That PR's five fixtures pin the `baseIndentLen+1` behaviour sourced from F# 4.1 spec §15.1.4/§15.1.8; a nesting-aware scan must leave them green.

Related: #7176 / #7184 (the dead band whose comment mis-attributed this symptom), #7185 (the same helper duplicated in `nim`/`rescript`), #7144, #7152.

cajasmota:
## What

`collectHierarchyEdges` regex-scanned a type's whole body for `inherit` / `interface ... with`
with **no notion of a nested declaration**, so a nested type's clause was attributed to the OUTER
type as well. Measured on `4d88f6372` with the issue's reproducer:

| entity | EXTENDS before | EXTENDS after |
|---|---|---|
| `ShallowHost` (outer, inherits nothing) | `[SeventySevenBase]` ← spurious | `[]` |
| `ShallowNested` (nested) | `[SeventySevenBase]` | `[SeventySevenBase]` |

The nested owner's edge was **already correct**, so the defect is a spurious edge, not a missing
one. The fix had to *move* the clause, not suppress it.

New `maskNestedTypeBodies` (`internal/extractors/fsharp/extractor.go`) blanks each nested type
declaration — header line plus the offside block it owns — with spaces, leaving newlines in place
so byte offsets and the stamped `line` Property are unchanged. `collectHierarchyEdges` matches
over the masked text; the object-expression brace guard (`insideBraces`) deliberately keeps
consulting the UNMASKED `scrubbed`, so masking cannot perturb any depth count.

## The design decision: header scan, not recorded spans

Both spans are emitted (outer 6–9, nested 7–9 on the reproducer), so subtracting the nested span
from the outer one is a genuine alternative. Rejected, third reason decisive:

1. Those spans live on **other `EntityRecord`s the assembly loop has not built yet** when
   `collectHierarchyEdges` runs; using them means a second pass and cross-entity coupling in a
   function whose contract is "given one body, return its edges".
2. `typeSeen` in that loop **dedups by NAME**, so a nested type whose name is already taken has no
   record and no span. A span-subtraction fix silently leaves that clause on the outer type.
3. Wherever the excision predicate and the re-attribution predicate diverge, the edge is **dropped
   rather than moved**. Sharing `typeRE` — the regex the file-level pass uses to decide what becomes
   a type — is the closest available coupling; a span-based fix couples them through a third
   artefact that can diverge independently. Mutant M3 below is that divergence, written out and
   scored.

   **CORRECTED after review (#7188 review item 1): the regex is shared, the COMPOSED predicate is
   not.** Two things sit between excision and re-attribution:

   - **`typeSeen` dedups by NAME.** A nested type colliding with an earlier one is masked out of the
     outer body and then produces no entity, so its clause lands nowhere. See "The recall loss this
     PR introduces" below.
   - **Excision reads the SCRUBBED body, re-attribution reads raw `src`.** Probed by the reviewer
     (fixture P10, `    (* c *)type P10Nested() =`): neither predicate matched, so nothing was
     excised without an owner. That half is **unexercised, not refuted**.

The block extent is **not** a re-statement of `extractIndentBody`'s rule — it is that function's own
answer, called from the END of the matched header with the header's own indent, exactly as the
file-level pass calls it (#7176's `baseIndentLen+1` partition, F# 4.1 spec §15.1.4 / §15.1.8). Its
output is a contiguous prefix, so its LENGTH is the block's extent. That is what makes the masked
region equal the nested type's own body EXACTLY, so a clause back at the outer's member column
*after* the nested block still belongs to the outer type
(`TestFSharp_NestedType7187_SiblingAfterNestedResumes` pins that direction).

**CORRECTED after review (item 2).** The first version walked from the newline after the match
START, which is only the same thing for a SINGLE-LINE header: `typeRE`'s `\s+`, `(?:<[^>]*>)?` and
`(?:\([^)]*\))?` all cross newlines, so a wrapped parameter list puts the header's *continuation* on
the next line, and a continuation at or shallower than the `type` column terminated the block early
— leaving #7187 alive verbatim (reviewer's fixture P4a). Now fixed by scanning from `m[1]`, and
pinned by `TestFSharp_NestedType7187_MultiLineNestedHeader` (three arms: shallow continuation, deep
continuation control, and a THREE-line header — see mutant M4 for why two lines was not enough).

## The recall loss this PR introduces — pinned, not just disclosed

`TestFSharp_NestedType7187_NameCollisionDropsTheEdge`. Measured both sides of the same fixture:

| | nested `Twin` colliding with a top-level `Twin` |
|---|---|
| **before this PR** | `CollideHost EXTENDS [CollideBase]` — **wrong owner, edge present** |
| **after this PR** | **no entity in the file carries it** — `CollideHost []`, top-level `Twin []`, no nested `Twin` entity |

This is a **new** wrong-edge → missing-edge conversion, narrow but real, and it is the same failure
mode M3 is rejected for. It is **not** "pre-existing behaviour, unchanged" — the earlier wording was
wrong about the consequence and is retracted. The trade is taken deliberately: attributing an edge
to a type that does not declare it is the error #6326 exists to avoid, and unwinding `typeSeen` is a
change to entity identity across the whole extractor rather than to this scan. Both the loss and a
renamed-nested control are asserted, so the loss is attributable to the collision and nothing else,
and a later change that restores the edge has to come here and say so.

Follow-up filed as **#7191**.

**#7184's boundary rows are untouched**: the whole `./internal/extractors/fsharp/` suite,
`indent_band_7176_test.go`'s five call sites included, is green — 0 `--- FAIL`.

## Tests

`internal/extractors/fsharp/nested_type_hierarchy_7187_test.go`, **six** tests. Every target name
is unique within its fixture (#7144 / #7152: F# dedup keys are name-keyed, so a shared name merges
silently and is invisible to both a presence and a count check). `fsHierTargets` fails if the owner
entity is missing, so no "does not have the edge" row can pass vacuously by the entity having been
deleted.

RED first: 3 `--- FAIL` on the fix reverted (now 5 / 11 assertion lines with the review's two new
tests); the multi-line-header row went RED on its own before the item-2 fix.

**CORRECTED after review (item 3): must-have / forbidden is NOT a second graded axis.** Where a type
declares a clause of its own, its row is `eqStrs(got, []string{own})`, and the paired
"does not contain T" row is **dominated by it on every input**, not merely on the mutants scored: a
forbidden row fires iff `T ∈ got`, and `T != own`, so `got != [own]` and the `eqStrs` row fires too.
Confirmed under M1 at lines 80&84, 127&145, 130&150. Those forbidden rows are **diagnostics** — they
name the leaked target in the message — and are not counted as coverage. The forbidden-direction
rows that genuinely carry weight are the `len(got) != 0` rows on an owner that declares NOTHING
(`ShallowHost`, `WrapHost`, `DeepHost`, `ThreeHost`, `CollideHost`), which have no `eqStrs` row to
dominate them. The brace-guard witness is a single `eqStrs` row and claims no axis at all.

**Axes VARIED:** owner (outer vs nested); clause kind (`inherit`→EXTENDS and
`interface ... with`→IMPLEMENTS); nested header indent (1, the issue's shallow form, and 4, the
conventional member column); position of the outer's own clause (before vs after the nested block);
header line count (1, 2, 3) and continuation indent (shallower vs deeper than the `type` column);
whether the nested type's name collides with an earlier type.
**Axes HELD CONSTANT:** nesting depth (one level); the outer declaration's column (0); file path;
one nested type per fixture; `.fs` extension; no generics on any nested header.

**DERIVED-NOT-EXECUTED:** no F# toolchain on this machine (`dotnet`, `fsc`, `fsharpc`, `fsi`,
`mono` all absent), so no fixture was compiled. The shallow form is the exact text measured on the
issue; the conventional form is F#'s ordinary nested-type layout.

## Mutants

Each applied with an asserted exact match count, the mutated line **read back**, `go vet` into its
own file (never through a pipe), `--- FAIL` counted with an anchored `grep -cE '^[[:space:]]*--- FAIL'`,
restored by `cp` and verified by `shasum` + a 0-count on the mutant marker.

|---|---|---|---|---|
| M1 | fix reverted — both scans read `scrubbed` again (2 occurrences, both mutated; `_ = masked` added so it compiles) | 0 | **5** | **DEAD** |
| M2 | M1 **with the new test file held out** | 0 | **0** | **ALIVE 0 — positive control** |
| M3 | the plausible-wrong fix: drop hierarchy edges from any type whose span **overlaps** another type's span (naive span subtraction) — suppresses BOTH edges | 0 | **5** | **DEAD** |
| M4 | item-2 fix half-reverted: delegate to `extractIndentBody`, but from the newline after the match **START** | 0 | **1** | **DEAD** — *was ALIVE 0 until the three-line-header arm was added; see below* |
| M4b | item-2 fix fully reverted: the original hand-rolled walk, verbatim | 0 | **1** | **DEAD** |
| EC-1 | brace guard swapped to consult `masked` at **both** sites | 0 | **1** | **DEAD** — *was ALIVE 0 until the brace-guard witness was added; see below* |

All mutants re-scored from scratch against the post-review test file — no kill count carried over.

Decomposition of M1's 5 failures: `ShallowNesting` 2 assertion lines + `ConventionalNesting` 4 +
`SiblingAfterNestedResumes` 1 + `MultiLineNestedHeader` 2 + `NameCollisionDropsTheEdge` 2 = **11**
lines across **5** tests. 2+4+1+2+2 = 11, 5 = 5. All five are mine; no other test in the package
failed — no confound.

Decomposition of M3's 5 failures: **10 assertion lines, all MUST-HAVE rows reporting `= []`, and 0
forbidden rows** — `ShallowNested` 1, `ConvHost`×2 + `ConvNested`×2 = 4, `ResumeHost`/`ResumeNested`
2, `WrapNested`/`DeepNested` 2, `Triplet` 1. 1+4+2+2+1 = 10. Only the must-have half catches the
suppress-both fix; the forbidden half passes under M3 because the spurious edge IS gone. M3's reach
is wider than the one behaviour probed — it nulls hierarchy edges for *any* overlapping pair — but
no pre-existing fixture in the package has overlapping type spans, so nothing else failed; declared
rather than assumed.

M4 and M4b are **isolated**: each produced exactly 1 `--- FAIL`, on `MultiLineNestedHeader` only,
with every other test in the package green — so the item-2 row grades the scan-start independently
of the masking mechanism.

**M4 is the finding worth recording.** Scored first with only the two-line-header arms, it was
**ALIVE at 0 `--- FAIL`**: `extractIndentBody` appends its FIRST line unconditionally when non-blank
(the same-line-body case), so starting one line early still absorbs exactly ONE shallow
continuation. It takes a SECOND one to terminate early. That is an ungraded hole, not an
equivalence — it was fixed by adding the three-line-header arm, after which M4 is DEAD. Recorded
because "the two-line case passes" would otherwise have read as proof the scan start was pinned.

**EC-1 — the brace guard's input was an ungraded causal claim.** The first version of this PR wrote
*"the BRACE guard deliberately keeps consulting the unmasked `scrubbed`, so masking cannot change
any object-expression depth count."* Swapping **both** guard sites to `masked` was **ALIVE at 0
`--- FAIL`**, so that sentence asserted something the suite never observed — in a comment explaining
why the code is written as it is, where the next reader takes it as settled. **Resolved by
constructing the distinguishing input and pinning it**, not by softening alone.

Masking preserves length, so offsets are identical and only content differs, and `insideBraces` asks
`count("{") > count("}")`. A verdict flip therefore requires a masked region holding **more `}` than
`{`** — a brace region that OPENS outside a nested block and CLOSES inside it. The opposite
imbalance cannot flip it, because `>` reads a negative depth the same as zero. The witness
(`TestFSharp_NestedType7187_BraceGuardReadsTheUnmaskedBody`): an object expression opens at line 11
column 8, the nested header is line 13 column 4, its closing `}` is line 14 inside the nested block,
and `BraceHost`'s own `inherit` is line 15 — genuinely its own clause, since the expression has
closed. **Measured: emitted with the guard on `scrubbed`, silently dropped with the guard on
`masked`.** EC-1 is now DEAD at exactly 1 `--- FAIL`, isolated to that row.

The doc comment no longer makes the causal claim. It now states the exact shape in which the two
inputs differ, that `scrubbed` is the conservative side, that the shape is **almost certainly not
legal F#** (DERIVED-NOT-EXECUTED — no toolchain here), and that **no corpus incidence is counted**.
This is a different axis from the scrubbed-vs-raw-`src` one still marked unexercised: that one is
about the MASKER's input, this one about the GUARD's.

M2's hold-out is proved by a `go test -list '.*'` diff, not by a zero `=== RUN` count:
exactly `TestFSharp_NestedType7187_ShallowNesting`, `_ConventionalNesting`,
`_SiblingAfterNestedResumes` disappear (172 → 169 tests).

## Other body consumers — MEASURED, and they KEEP the defect

This fix teaches **only the hierarchy scan** about nesting. Measured with a throwaway probe on the
fixed tree (since deleted), one nested type per case:

| consumer | verdict |
|---|---|
| `collectHierarchyEdges` | **FIXED** here |
| CE-builder detection (`compexpr_active_patterns.go:517`) | **STILL LEAKS** — a nested type with `Bind`/`Return` stamps the OUTER type `subtype=computation_builder`, `ce_builder=true`, `ce_builder_members=Bind,Return`. Measured. |
| type→member `CONTAINS` (the `memberRE` indent loop, `extractor.go:~730`) | **STILL LEAKS** — outer *and* nested both claim the nested type's `Bind`/`Return`. Measured. This loop scans the whole file by indent, a different mechanism from the body scan. |
| `classifyTypeSubtype` | **no leak observed** on a nested DU and a nested record (`DuHost`/`RecHost` stayed `subtype=type`); it branches on the declaration text first. Not proven leak-free in general — two shapes only. |
| `extractTypeMembers` / `parseRecordFields` | **no leak observed** on the same two shapes — it is gated on the owner's own subtype, which did not leak. Conditional on the row above. |
| `collectCalls` | **NOT MEASURED for nesting.** It runs on `let`/`member` bodies, not type bodies; a nested *type* inside a function body was not probed. |
| match-site edges, `validators.go:344` | **NOT MEASURED.** |

So: this is **not** a general nesting fix, and the two rows above are live defects after this PR.

## Not claimed

- **No corpus incidence counted.** Whether real F# nests declarations this way is open; it decides
  priority, not correctness.
- The `type X() as this =` self-identifier form is the one shape masking cannot help: `typeRE` does
  not match it, so such a nested type is neither masked out of the outer body nor given an owner —
  its clause still lands on the enclosing type. Noted in the doc comment, not fixed.
- ~~A nested type whose name collides ... Pre-existing dedup behaviour, unchanged.~~ **RETRACTED**
  after review: the dedup is pre-existing, the **edge loss is new and caused by this PR**. Moved up
  to "The recall loss this PR introduces", where it is measured and pinned.
- The scrubbed-body vs raw-`src` divergence (design note (b)) is **unexercised**: the reviewer's
  probe for a drop through that gap did not land. Not claimed to be impossible.

## Gate

`go test -count=1 ./cmd/grafel/` with FRESH `GRAFEL_HOME` + `GRAFEL_DAEMON_ROOT`: PASS (re-run after
the review changes), **digest UNMOVED** (`writeSpanFixture6118` has 18 members and zero `.fs`/`.fsx`/`.fsi`, so no F# body ever
reaches this scan). No constant was touched.
`go test -count=1 ./internal/extractors/fsharp/`: PASS, 0 `--- FAIL`. All **seven**
`IndentBand7176_*` rows PASS under the fix (re-checked after the item-2 change and again after
the brace-guard witness).

Closes #7187


cajasmota:
## Adversarial review — independent worktree `archigraph-wt-7188-rev` @ `e2038d31c`

Everything below was run in my own detached checkout of `origin/wt-7187-a1`, restored by `cp` and
verified by `shasum` (`d9a0be95…`, 0 mutant markers, `git status` clean) before the green run.
My probes live in an untracked `zzrev7188_probe_test.go`; they are diagnostic dumps (entity,
`StartLine`-`EndLine`, every EXTENDS/IMPLEMENTS with its `line` Property), not assertions.

### What I could not break

| attack | result |
|---|---|
| **Outer type with its OWN `inherit` AND a nested type** (the brief's #1 — "the single most likely way this fix is wrong") | **Does not land.** P1: outer clause BEFORE the nested block *and* an `interface … with` AFTER it → `P1Host [EXTENDS OBase@L11, IMPLEMENTS IOuter@L16]`, `P1Nested [EXTENDS NBase@L13, IMPLEMENTS INest@L14]`. Exactly right. `ConventionalNesting` already covers the before-position; P1 adds the after-position on the IMPLEMENTS arm. |
| **Line/offset exactness across a masked region** | **Claim holds.** P5 (nested block containing a blank line and a member, outer clause on line 11): `P5Host EXTENDS E2@L11`, `P5Nested EXTENDS E1@L8` — true file lines. `EndLine` is computed from the pre-mask `body`, so masking cannot reach it. |
| **Nested-in-nested** | **Clean at 3 levels.** P2: `P2A→B1@L9`, `P2B→B2@L11`, `P2C→B3@L13`, no cross-contamination. |
| **Sibling resumption / two sibling nested blocks** | **Clean.** P3: `P3N1→C1`, `P3N2→C2`, `P3Host→C3` only. |
| Access-modifier nested (`type private …`), object expression inside the nested block | **Clean** (P8, P7). The `insideBraces`-on-unmasked decision costs nothing observable. |
| **M1 (fix reverted)** | **Reproduced exactly:** 3 `--- FAIL`, 7 assertion lines, and *only* those 3 tests failed in the whole package — the PR's no-confound claim is correct. |
| **#7176's boundary rows** | All 7 `IndentBand7176_*` PASS under the fix **and** under M1 — untouched, and they grade nothing here. Full package: **0 `--- FAIL`**, 724 `=== RUN`. |
| Scrub-vs-raw predicate divergence in the DROP direction (P10, `    (* c *)type P10Nested() =`) | **Attempted, did not land.** Neither predicate matched, so nothing was excised without re-attribution. Recorded as a failed attack, not as proof of identity. |

### 1. The fix turns a mis-attributed edge into a MISSING one — the exact failure its own M3 rationale forbids

Measured both sides of the same fixture (P9):

```
type J1() =
    class end
type Dup() =        // top-level
    class end
type P9Host() =
    type Dup() =    // nested, name collides -> typeSeen drops it
        inherit J1()
```

* **pre-fix (M1 applied):** `P9Host EXTENDS->J1@L8` — wrong owner, edge present.
* **post-fix:** **no entity in the file carries `EXTENDS J1`.** `P9Host []`, top-level `Dup []`, no nested `Dup` entity.

The PR's own standard for rejecting span subtraction is "a fix that suppresses both converts a wrong
edge into a missing one". That is precisely what happens here, and the excision predicate
(`typeRE` over the body) genuinely is *not* the re-attribution predicate (`typeRE` over `src`
**plus `typeSeen`**) — the `typeSeen` filter sits between them. "Identical by construction" is
therefore true of the regex and false of the composed predicate.

"Not claimed" discloses the shape but labels it *"Pre-existing dedup behaviour, unchanged."* The
dedup is pre-existing; **the edge loss is new and is caused by this PR**. Please reword, and pin it —
a test here is the one place a forbidden-direction row would earn its keep, because the must-have
row has no owner to hang on.

### 2. "Byte-for-byte `extractIndentBody`'s closing rule" is false for a multi-line header

`extractIndentBody` starts its scan at `m[1]` (after the `=`). `maskNestedTypeBodies` starts after
the **first newline following `m[0]`**. For a single-line header these coincide. But `typeRE`'s
`\s+`, `(?:<[^>]*>)?` and `(?:\([^)]*\))?` all cross newlines, so a header spanning lines diverges:
the masker treats the header's *continuation* line as the first body line and terminates on it if
its indent `<= headerIndent`.

Controlled pair, only the continuation indent varied:

| fixture | continuation indent vs `type` column | outer |
|---|---|---|
| P4a | 2 (shallower) | **`P4aHost EXTENDS->D1@L7` — spurious; #7187 survives verbatim** (byte-identical under M1) |
| P4b | 8 (deeper) | `[]` — clean |
| P4c (realistic wrapped `(x: int,\n  y: int)`) | deeper | `[]` — clean |

Incidence is confined to a continuation line at or shallower than the `type` keyword, which is
probably not even legal F#, so this is a doc-accuracy defect more than a live bug. But the fix is one
line — walk from the end of the **matched header** rather than from the first newline after `m[0]` —
and it would make the stated equivalence actually true.

### 3. Re-derived M3's asymmetry, and the sharper answer: the forbidden rows catch nothing

M3's reported shape is right. The converse is stronger than the PR says. Each forbidden row fires iff
`T ∈ got`; its paired must-have row is `eqStrs(got, []string{own})` with `T != own`, so
`T ∈ got` ⇒ the must-have row fires too. **Domination is total over every input, not just over M1 and
M3** — no mutant can exist that a forbidden row catches and the must-have row does not. Confirmed on
M1, where they co-fire in pairs: lines 70 & 74, 117 & 135, 120 & 140. Three of M1's seven assertion
lines are redundant diagnostics.

The genuinely load-bearing forbidden-direction row is `ShallowNesting`'s `len(got) != 0`, because
`ShallowHost` has no own clause and so has no `eqStrs` row to dominate it. The PR body presents
must-have/forbidden as a varied *axis*; it is one graded axis plus three error messages. Not a
blocker — but don't count it as coverage.

### Accurately disclosed, re-measured and confirmed

`type X() as this =` nested: `P6Host EXTENDS->F1@L6`, no `P6Nested` entity. Leak present exactly as
documented.

### Not verified

* `./cmd/grafel/` PASS / digest-unmoved — not re-run (machine limits); I did not independently check
  that `writeSpanFixture6118` has no `.fs`/`.fsx`/`.fsi` member.
* The residual-leak table (CE-builder, `memberRE` CONTAINS, `classifyTypeSubtype`,
  `extractTypeMembers`) — I did not re-measure the author's leak/no-leak claims.
* Corpus incidence of any of this. CRLF input. `.fsi` paths.

---

## REQUEST-CHANGES

1. **Reword the `typeSeen` disclosure and pin it.** "Pre-existing dedup behaviour, unchanged" is
   wrong about the consequence: pre-fix the clause landed on the outer type, post-fix it lands
   nowhere (measured, P9 above). This PR creates a recall regression, narrow but real, and it is the
   same failure mode the PR rejects M3 for. Add a test.
2. **Drop or correct "byte-for-byte `extractIndentBody`'s closing rule".** It holds only for
   single-line headers; P4a shows the original defect surviving for a multi-line one. Either walk
   from the end of the matched header (one line) or state the restriction in the doc comment.
3. **Soften the "identical predicate" claim** in both the PR body and the `maskNestedTypeBodies`
   doc comment: the regex is shared, the *composed* predicate is not — `typeSeen` (item 1) and
   scrubbed-body vs raw-`src` both sit between excision and re-attribution. I probed the scrub/raw
   asymmetry for a drop (P10) and it did not land, so that half is unexercised rather than refuted.
4. **Do not present must-have/forbidden as a second graded axis.** Three of the four forbidden rows
   are provably dominated (item 3 above); say they are diagnostics.

None of 1–4 questions the core mechanism. On the eight shapes I could attack it — outer-own-clause
both sides, three-level nesting, sibling nested pairs, resumption, modifiers, object expressions,
line stamping, and the #7176 boundary rows — the masking is correct and the `line` Properties are
exact.

PR #7189, opened 2026-09-18
fix(#7185): the same dead band in nim and rescript — two languages, two separately sourced boundaries, not one answer copied from F#

### LINKED ISSUE #7185: extractIndentBody's dead band is COPY-PASTED: the same base+1 gap exists byte-identically in nim and rescript, not just fsharp — and the +1 fix is an F#-offside argument that does NOT transfer to either
## What

`extractIndentBody`'s dead band (#7176) is **not one function**. The same helper name exists in seven extractor packages, and the **3-argument variant carries a byte-identical dead band in three of them**:

| package | file | dead band? |
|---|---|---|
| **fsharp** | `extractor.go` | **YES** — being fixed by #7176 / PR #7184 |
| **nim** | `nim.go` | **YES** — unfixed |
| **rescript** | `extractor.go` | **YES** — unfixed |
| elm | `extractor.go` | no — different design |
| haskell | `extractor.go` | no — different design |
| idris | `extractor.go` | no — different design |
| pony | `extractor.go` | no — different design |

Measured on `b31fe2c67`.

## The shape, identical in all three

```go
minBodyIndent := baseIndentLen + 2
...
if indent >= minBodyIndent {
    bodyLines = append(bodyLines, line)
} else if indent <= baseIndentLen && strings.TrimSpace(line) != "" {
    break
}
```

A line at exactly `baseIndentLen+1` satisfies **neither** branch: it is not appended, and it does not terminate. The loop silently skips it and keeps going, so a sibling declaration one column off falls inside the PREVIOUS declaration's body. Only the comment beside each copy differs — *"F# typically uses 4-space indent, but 2 is minimum"*, *"Nim typically uses 2-space indent"*, *"ReScript uses 2-space indent conventionally"* — which is the tell that the logic was copied and only the prose localised.

**The 2-argument variants are a different design, not a fixed one.** They test `line[0] == ' ' || line[0] == '\t'` — binary, indented-or-not — so there is no band to fall into. They are listed above so nobody re-derives them.

## Why this is filed separately from #7176

#7176 is scoped to F#, and its fix is **correctly** scoped: PR #7184 sources the `+1` boundary from the **F# 4.1 Language Specification §15.1.4 / §15.1.8**, which is an argument about *F#'s offside rule*. That reasoning does not transfer, and that is the whole point:

- **Nim is indentation-significant**, so a base+1 line plausibly has the same defect — but it needs **Nim's** indentation rules cited, not F#'s.
- **ReScript is brace-delimited**; indentation is *not* semantic there. The `+2` heuristic is a convention-following guess rather than a language rule, so "what should a base+1 line do" may not even have a right answer in the same sense. It may be correct to leave ReScript alone and say so.

So this is three separate decisions wearing one shape. **Transferring the ANSWER from #7176 would be exactly the error this repo keeps measuring** — the fix is per-language; only the *question* transfers.

## Provenance

Found while grounding PR #7184's report. That PR disclosed two `extractIndentBody` call sites my issue body had missed (`compexpr_active_patterns.go`, `validators.go` — the F# helper has **five** callers, not the three I wrote). Checking the caller list repo-wide surfaced the cross-package duplication.

**#7176's body is wrong on the caller count and is corrected by this.** The blast-radius argument there still holds and in fact widens: within F# alone the helper feeds `EndLine`, `classifyTypeSubtype`, CONTAINS edges, hierarchy edges and `collectCalls`.

## Not claimed

- **Not claimed to be a live defect in nim or rescript.** The *shape* is verified byte-identical; whether a base+1 line occurs in real Nim or ReScript, and what it should do, is **unmeasured**. No corpus incidence counted for any of the three — #7176's "low frequency" adjective was inherited from a comment and has never been measured either.
- **Not claimed the fix is `+1` everywhere.** That is #7184's F#-specific, spec-sourced conclusion. Copying it without sourcing each language would be transferring an answer across twins.
- **Not claimed the 2-arg variants are correct** — only that they have no dead band. Their binary rule has its own properties, unexamined here.
- **No toolchain for any of the three.** No F#, no Nim, no ReScript compiler on this machine (`javac` is the only compiler present), so every legality claim is derived-not-executed and must cite a spec.

## Grading requirement, per package taken

- **Cite that language's indentation rules** for the chosen boundary, and label it derived-not-executed with a stated falsifier.
- **A fixture per call site**, asserting the emitted artefact — the wrong `EndLine`, the wrong parent, the wrong subtype — not an internal counter. A verdict at one call site says nothing about the others; PR #7184 demonstrated this by finding the F# type pass flips `Subtype` from `type` to `discriminated_union`, a consequence the issue never predicted.
- **Both sides of the boundary**: base+1, plus controls at base and base+2.
- **Score the widening direction.** Appending the boundary line grows every body, and these scanners read **raw `src`** (#7152), so a larger body can mint more from comments and string literals.
- **Do not "fix" all three in one PR** unless each language's boundary is separately sourced in that PR's body.

Related: #7176 / #7184 (the F# arm and its spec citation), #7144 (blocked on #7176; its arms rest on this helper's containment), #7152, #7118.

cajasmota:
`extractIndentBody` is copy-pasted byte-identically into `internal/extractors/nim/nim.go`
and `internal/extractors/rescript/extractor.go`:

    minBodyIndent := baseIndentLen + 2
    ...
    if indent >= minBodyIndent { append } else if indent <= baseIndentLen { break }

A line at exactly `baseIndentLen+1` satisfies NEITHER branch. It is not appended
and it does not terminate, so the loop steps over it AND KEEPS SCANNING. The
emitted body then has a HOLE — the base+1 line gone while deeper lines below it
are still collected.

Both files change by one line, `+2` -> `+1`. THE TWO DECISIONS ARE SOURCED
SEPARATELY AND THE ARGUMENTS ARE NOT THE SAME ARGUMENT. PR #7184 chose `+1` for
fsharp from the F# 4.1 offside rule; copying that here because it worked there
would be transferring an ANSWER across twins. Only the question transfers.

NIM — sourced to Nim's own indentation rules.
Nim manual, Lexical Analysis -> Indentation:
  "Nim's standard grammar describes an indentation sensitive language. This means
   that all the control structures are recognized by indentation. Indentation
   consists only of spaces; tabulators are not allowed."
and the grammar pseudo-terminals that section defines:
  IND{>}  "denotes an indentation that consists of MORE SPACES than the entry at
           the top of the stack"
  IND{=}  "an indentation that has the SAME number of spaces"
An indented statement list is IND{>} — strictly more spaces than the enclosing
entry — while a SIBLING is IND{=}, exactly the enclosing column. The manual names
no minimum step, so "more spaces" is satisfied by one. base+1 is body; only
base-or-less can be a sibling, which is exactly the existing `<=` terminator.
DERIVED-NOT-EXECUTED: no `nim`/`nimble` on this machine (`javac` is the only
compiler present). FALSIFIER: a Nim program in which a statement indented exactly
one space deeper than its enclosing declaration is rejected, or parses as that
declaration's sibling.

RESCRIPT — the boundary is NOT sourced to a ReScript language rule, BECAUSE THERE
IS NONE, and the PR says so rather than borrowing one.
ReScript Language Manual v11, "Let Binding" -> Block Scope:
  "Bindings can be scoped through `{}`."
  "The value of the last line of a scope is implicitly returned."
Scope is delimited by BRACES. The manual assigns no meaning to a line's column, so
ANY column > 0 is a legal body indentation and ANY column is a legal top-level
declaration indentation. No threshold can be CORRECT for ReScript; this helper is
an approximation of brace matching the extractor does not do. "What should a
base+1 line do" genuinely has no language-level answer here, as #7185 anticipated.

The dead band is a defect anyway, independent of that answer, because it did not
pick a different threshold — IT MADE THE LOOP NON-TOTAL. No threshold justifies
"skip this line and keep going".

THE APPEND-OVER-TERMINATE TIEBREAK IS CARRIED BY TABS, AND IT IS MEASURED RATHER
THAN SUPPOSED. `countIndent` counts a TAB as ONE column, and ReScript has no column
rule at all, so tab-indented source is legal and entirely ordinary. Every line of a
tab-indented top-level body therefore sits at exactly base+1 — THE WHOLE BODY IS
INSIDE THE DEAD BAND. Measured on this branch, both sides run by me, artefacts
printed rather than inferred:

    fixture: "let alpha = () => {\n\tstepOne()\n\t<Comp />\n}"
      +2 (parent):  alpha 1-1  CALLS:[]         RENDERS:[]     <- entire body invisible
      +1 (head):    alpha 1-3  CALLS:[stepOne]  RENDERS:[Comp]

TERMINATE would have dropped those bodies too: it converts the silent skip into a
silent truncation at the first tab, so the whole tab-indented class loses its CALLS
and RENDERS either way. That is the ground for APPEND. The weaker form of the same
argument — "a body line MAY legally sit at base+1" — does follow from the brace
rule, but it is a hypothetical; the tab class is its measured instance, and it is
what the decision actually rests on. (An earlier draft of this body had these the
wrong way round and demoted the tab question to "separate, untouched"; it is not
separate, it is the load-bearing case.) Same shape measured in nim —
`1-2 CALLS:[]` -> `1-3 CALLS:[stepOne]` — but academic there, since the Nim manual
forbids tabs outright; `countIndent`'s tab handling is a live defect in the nim
copy and is NOT fixed here.

The cost of APPEND — a top-level declaration hand-indented by one column is then
absorbed into the previous body — is NOT a regression: MEASURED in `NoHoleInTheBody`,
under `+2` that declaration's deeper body lines are ALREADY absorbed (`alpha` CALLS
`inner` before and after, identically), with only its header line silently missing.
The fix stops hiding the mis-attribution; it does not create it.

A THIRD OPTION EXISTS AND IS DELIBERATELY OUT OF SCOPE: BRACE MATCHING, which is
what ReScript actually needs, braces being the real delimiter. "Two ways to make it
total" above is exhaustive only for a loop that classifies lines by column;
replacing the column heuristic with a brace scanner is the correct fix and is a
different change from this one.
DERIVED-NOT-EXECUTED: no `rescript`/`bsc`/`bsb`/`res` on this machine. FALSIFIER,
and it is a different KIND of falsifier: not "a ReScript program where base+1 is a
sibling" — those exist and are legal, indentation being free — but a CORPUS claim,
that real ReScript has more one-column-indented TOP-LEVEL declarations than
one-column-indented BLOCK BODIES. No ReScript corpus was measured; the claim rests
on `rescript format` being the single canonical style, which empties the band in
conforming source either way.

FOUR CALL SITES, FOUR SEPARATE VERDICTS, each asserting the EMITTED ARTEFACT. A
verdict at one site says nothing about another — #7184 proved that by finding two
sites it had called "covered transitively" both moved:

  nim.go:149        proc  `alpha` EndLine 3->4; CALLS [stepTwo] -> [stepOne stepTwo]
  nim.go:207        type  `Alpha` EndLine 3->4  (base is hard-coded 0 here, a
                          different base and a different Kind from :149)
                          SCOPE LIMIT: this row grades the boundary at base 0,
                          which is right only because the fixture puts `type Alpha
                          = object` at column 0 — idiomatic Nim writes an indented
                          `type` SECTION, where the hard-coded 0 makes every type
                          absorb the ones after it (#7190, reproducing identically
                          at +2 and +1, so neither caused nor cured here).
  rescript:182      let   `alpha` EndLine 2->3; CALLS [stepTwo] -> [stepOne stepTwo]
  rescript:220      type  `alpha` EndLine 2->3  (base = the decl's own indent)

Both sides of the boundary are pinned at every site: base+1 (the band), a control
at base (the sibling must terminate) and a control at base+2 (body under every
candidate threshold).

THE WIDENING DIRECTION IS PINNED, AND ITS PREMISE DIFFERS FROM F#'s. #7152
recorded that the fsharp scanners read RAW `src`. MEASURED HERE, these do NOT:
nim's `collectCalls` (nim.go:469) and rescript's `collectCalls` (extractor.go:427)
and `collectRenders` (extractor.go:480) each run `stripStringsAndComments(body)`
first. The `…_WideningMintsNothingFromCommentsOrStrings` rows hold the forbidden
half anyway — a comment and a string literal at base+1 in both languages, plus a
JSX tag inside a ReScript comment — each behind a must-have EndLine assertion so
the forbidden half cannot go vacuous.

MUTANTS. Full package suite per row, `go vet` into its own file before scoring,
exact match count asserted before applying, mutated line READ BACK, restored by
`cp` and verified by shasum with the marker grepped to 0. `--- FAIL` counted with
an anchored `grep -cE '^\s*--- FAIL'`; `=== RUN` counted too, so a broken build
could not read as a kill.

  nim       M2  revert, new file HELD OUT   ALIVE 0   -list 44->37, diff = exactly my 7
  nim       M1  revert to `+2`              DEAD 3    both call sites fail independently
  nim       M3  `+0` (other direction)      DEAD 8    7 mine + 1 pre-existing, sums
  nim       EA-1 first-line `i == 0` branch DEAD 1    was ALIVE 0 before the row below
  rescript  M2  revert, new file HELD OUT   ALIVE 0   -list 26->18, diff = exactly my 8
  rescript  M1  revert to `+2`              DEAD 4    both call sites fail independently
  rescript  M3  `+0` (other direction)      DEAD 7    7 mine + 0 pre-existing, sums

COORDINATOR MUTANT EA-1, run by nobody else, AND THE ROW ADDED BECAUSE OF IT.
Both copies open with a first-line special case that BYPASSES the threshold this
PR moves:

  if i == 0 && strings.TrimSpace(line) != "" { append; continue }

nim's comment names the exact construct it serves — `Same-line body: "proc foo()
= result"`. Disabling it in both packages killed exactly ONE test across the two,
a rescript synthetic fixture, and NOTHING in nim: prose asserting what no test
observes, in the copy whose own comment names the construct.

  nim  EA-1  first-line branch disabled, BEFORE the new row   ALIVE 0
  nim  EA-1  first-line branch disabled, AFTER  the new row   DEAD 1

The killing assertions, both artefacts moving:

  inner CALLS = "", want "compute"
  inner EndLine = 2, want 3

It is not a generic gap: it interacts DIRECTLY with the boundary moved here. The
remainder after the regex match starts at the space before the body, so its indent
is 1. At base 0 that clears minBodyIndent under +1 anyway and the branch is
invisible — which is why every other row in this PR is blind to it. At base 2, an
INDENTED declaration, 1 fails `>= 3`, satisfies `<= 2`, and the loop terminates on
the first line with an EMPTY body. The first-line branch is the only thing that
carries a same-line body past the threshold, so it is the one case where the
first line bypasses the value this PR changed.

THE INDENTED SAME-LINE FORM IS LEGAL NIM, derived from the manual's own grammar
("The Nim grammar"), fetched and grepped rather than recalled:

  routine = optInd identVis pattern? genericParamList?
    paramListColon pragma? ('=' COMMENT? stmt)? indAndComment

The body is `('=' COMMENT? stmt)?` — a `stmt` directly after `=` with NO IND{>}
interposed, which IS the same-line form — and the production is led by `optInd`,
so a routine may itself be indented, i.e. nested. `proc inner(): int = compute()`
at column 2 inside another proc is therefore grammatical. DERIVED-NOT-EXECUTED;
FALSIFIER: a Nim program in which a nested routine with its body on the
declaration line is rejected.

M1 and M3 bracket the boundary from both sides, so the rows pin `+1` exactly
rather than "some boundary". CONFOUND DECLARED on M3: `+0` does not merely shift
the threshold, it makes the terminator unreachable (`indent >= 0` is always true),
so the body runs to EOF and it breaks more than the one behaviour probed — that is
why the base controls fail under it, and why `NoHoleInTheBody` (whose fixture has
no base-level terminator to lose) is the one row of mine that survives it.
M2 is the positive control: ALIVE at 0 with a non-zero `=== RUN` count (nim 52,
rescript 18) — a real pass of the pre-existing suites, not a build break. The 37
and 18 pre-existing tests are BLIND to this boundary, so "covered transitively"
was worth exactly zero here too.

GATES. `go test -count=1 ./cmd/grafel/` with FRESH `GRAFEL_HOME` and
`GRAFEL_DAEMON_ROOT`: exit 0, 0 `--- FAIL`, digest UNMOVED. Checked rather than
assumed: `writeSpanFixture6118`'s 18 members are 7 `.cs`, 1 `.java`, 1 `.ts`,
1 `.js`, 6 `.py`, 1 `.go`, 1 `.rs` — ZERO `.nim`/`.res`/`.resi`. So the digest
CANNOT move, and an unmoved digest here is evidence of nothing about the change;
the reason is stated instead of the result alone. Full suites green for both
packages changed, plus every other package that names a `.nim`/`.res`/`.resi`
file: `internal/custom/nim`, `internal/extractors/cross/testmap`,
`internal/daemon/extract`, `internal/quality`, `internal/classifier`,
`internal/substrate`, `internal/feedback`, `internal/extractors/swift`,
`internal/engine`. All exit 0 with 0 anchored `--- FAIL`.

NOT CLAIMED. No corpus incidence was measured for either language — whether a
base+1 line actually occurs in real Nim or ReScript is still unknown, exactly as
#7185 said. Neither legality claim was compiled. `countIndent`'s treatment of a
TAB as one column is load-bearing for the ReScript decision above and is discussed
there; in the nim copy it contradicts the manual's "tabulators are not allowed"
outright and is left untouched here, unfiled.

Closes #7185


cajasmota:
## Adversarial review — independent worktree at `origin/wt-7185-a1` (10123c3fb), parent 4d88f6372

I did not write this code. Everything below I re-derived myself; where I could not, I say so.
The one-line-per-file claim is true (`git diff 4d88f6372 HEAD` ignoring comments = exactly two
lines, `+2` -> `+1`), so reverting that line reproduces main's behaviour for this helper exactly —
that is how I built my "before" side.

### MEASURED — every quantitative claim in the body reproduces

| claim | my measurement |
|---|---|
| nim M2 held-out ALIVE 0 | 0 anchored `--- FAIL`, **70 `=== RUN`** (52 nim + 18 rescript — sums to the body's split) |
| nim `-list` 44 -> 37 | 44 / 37; `diff` of the sorted name sets = **exactly the 7 new names**, nothing else moved |
| rescript `-list` 26 -> 18 | 26 / 18; diff = **exactly the 8 new names** |
| M1 (`+2`) DEAD 3 nim / 4 rescript | 7 total, exactly those 7 names, both call sites failing independently |
| M3 (`+0`) DEAD 8 nim / 7 rescript | 15 total = 7 mine + 7 mine + 1 pre-existing (`TestNimImportPlaceholderDoesNotDropCrossFileCalls_6481`). `NoHoleInTheBody` survives M3, exactly as the confound paragraph predicts |
| scrub at 3 sites | confirmed: nim `collectCalls` and rescript `collectCalls` + `collectRenders` all call `stripStringsAndComments(body)` |

Method: `cp` backup, exact match count asserted before each `sed`, mutated line **read back**,
`go vet` into its own file (exit 0 each time, so no mutant read as a kill via a broken build),
restore by `cp` verified by `shasum` + a marker grep to 0. Final tree: `git diff HEAD` = 0 lines.

### `NoHoleInTheBody` — re-derived independently, and it holds

I did not run the author's test; I built my own fixture and printed the emitted artefacts on both
sides.

```
fixture: "let alpha = 1\n let gamma = () => {\n   inner()\n }"
  +2 (main):  alpha 1-2  CALLS:[inner]      gamma 2-2  CALLS:[]
  +1 (head):  alpha 1-4  CALLS:[inner]      gamma 2-3  CALLS:[inner]
```

`alpha CALLS inner` on **both** sides. The claim "the fix stops hiding the mis-attribution, it does
not create it" is **true as stated**; only `EndLine` moves. A variant with gamma's body at indent 2
instead of 3 behaves identically. Attack failed to land.

### Both language sources, read independently against the primary docs

- **Nim manual, Lexical Analysis -> Indentation.** The quoted sentence is verbatim, and so are
  `IND{>}` = "more spaces than the entry at the top of the stack" and `IND{=}` = "the same number".
  I specifically looked for a minimum step: **the manual names none**, so one space satisfies
  `IND{>}`. The `+1` reading holds and the old comment ("Nim typically uses 2-space indent") was a
  style observation masquerading as a rule. Derived-not-executed (no `nim` here) — same as the PR.
- **ReScript Manual v11, Let Binding -> Block Scope.** "Bindings can be scoped through `{}`";
  indentation carries no semantics. The PR's refusal to source a threshold here is correct, and the
  non-totality argument stands on its own: with `>= base+1` / `<= base` the two arms are now
  complementary, so no band can exist.

### Three things I'd change, none blocking

1. **The ReScript tiebreak rests on a weaker argument than the one available — and I measured the
   stronger one.** `countIndent` counts a TAB as **1 column**, and ReScript has no column rule, so
   tab-indented source is perfectly legal. Under `+2` a tab-indented body is *entirely* in the dead
   band:
   ```
   fixture: "let alpha = () => {\n\tstepOne()\n\t<Comp />\n}"
     +2 (main):  alpha 1-1  CALLS:[]          RENDERS:[]     <- whole body invisible
     +1 (head):  alpha 1-3  CALLS:[stepOne]   RENDERS:[Comp]
   ```
   (Same shape measured in nim: `1-2 []` -> `1-3 [stepOne]`; only academic there, since Nim forbids
   tabs.) So "a body line may legally sit at base+1" is not a hypothetical — it is the **entire class
   of tab-indented ReScript**, and TERMINATE would drop it too. That is a far better ground for
   choosing APPEND than the un-measured corpus claim the body currently leans on. The "NOT CLAIMED"
   section demotes the tab question to "a separate question... untouched here"; it is in fact the
   load-bearing case for this very decision. Please promote it.
2. **Say in the body what the code comment already says**: that there is a third option — brace
   matching, which is what ReScript actually needs — and that it is deliberately out of scope. The
   body's "of the two ways to make it total" is exhaustive only for a fixed line-classification
   loop; a reader of the body alone is not told that.
3. **Nit:** the scrub citations (`nim.go:435`, `extractor.go:390` / `:443`) match neither main
   (431 / 386 / 439) nor head (465 / 423 / 476) — off by exactly +4 from main.

### A separate defect found while attacking site :207 — NOT caused by this PR, deserves its own issue

`nim.go:207` passes `baseIndentLen` hard-coded to `0` instead of the declaration's own indent. On
idiomatic Nim — a `type` section with the declarations indented 2 and fields indented 4 — every type
swallows the ones after it:

```
type
  Alpha* = object
    a: int
  Beta* = object
    b: string
  -> Alpha 1-5, Beta 4-6      (Alpha.EndLine 5 >= Beta.StartLine 4)
```

Identical under `+2` and `+1`, so this PR neither causes nor fixes it. It does mean the `:207` row
grades the boundary at a base value (0) that happens to be right only because the fixture puts
`type Alpha = object` at column 0 — the row is genuine (DEAD under both M1 and M3, I confirmed), but
it is structurally unable to see the wrong-base defect. Worth a follow-up issue; I have not filed one.

Also noted and out of scope: trailing blank lines are appended unconditionally, so `EndLine`
over-reports by one whenever a blank line precedes the sibling — which is why the nim ProcSite row
asserts `EndLine 4` on a fixture whose last real line is 3. Pre-existing; the row is still correct
as a pin.

### What I could NOT verify

- **Neither legality claim was compiled.** No `nim`/`nimble`, no `rescript`/`bsc`/`res` on this
  machine. Both readings are from the primary manuals only, as the PR states.
- **No corpus incidence**, for either language — the PR says this too. The only `.nim`/`.res`/`.resi`
  files on disk anywhere in the repo are the three in `internal/quality/golden/nim-objects-mini/src/`,
  and I checked them: **zero** single-space-indented lines and **zero** tab-indented lines. So the
  "all other packages green" list is honest but says nothing about this change — those fixtures
  cannot move. I re-ran `internal/classifier`, `internal/custom/nim` and
  `internal/extractors/cross/manifest` (the last one the body does not list; it reads `.nimble`
  manifests, not the extractor) — all exit 0, 0 anchored `--- FAIL`.
- **I did not run `./cmd/grafel/`.** I enumerated `writeSpanFixture6118`'s members myself: 7 `.cs`,
  1 `.java`, 1 `.ts`, 1 `.js`, 6 `.py`, 1 `.go`, 1 `.rs` = 18, **zero** `.nim`/`.res`/`.resi` — and
  `cmd/grafel/` contains no reference to any of those three extensions at all. The author is right
  that an unmoved digest is evidence of nothing here, so reproducing it would have been a 170s
  vacuous result. Recording the reason instead, which is what the PR does.

---

## APPROVE

1. Every mutant count, every `-list` delta, and the `=== RUN` split reproduce **exactly** under an
   independent run.
2. `NoHoleInTheBody`'s load-bearing measurement re-derived on my own fixture: `alpha CALLS inner`
   identically at `+2` and `+1`. The "reveals, does not create" claim survives.
3. Both language sources check out against the primary manuals, including the specific thing I went
   looking for — Nim names **no** minimum indentation step.
4. The digest claim is right and the gate run would have been vacuous; verified rather than repeated.
5. Non-blocking: promote the **tab-indentation** measurement above (it is the ReScript-specific
   ground the decision actually needs), name brace-matching as the excluded third option, fix the
   three off-by-4 line citations.
6. Follow-up issue warranted for `nim.go:207`'s hard-coded base `0`, which absorbs sibling types in
   idiomatic Nim `type` sections. Pre-existing and untouched by this PR.

Issue #7181, opened 2026-09-18
fsharp: moduleRE, memberRE and typeRE modifier allowlists are graded in the NARROWING direction only — adding 'sealed' to any of the three is ALIVE at 0, while dropping 'private' is DEAD at 14/13/26


cajasmota:
## What

All three of the F# extractor's remaining modifier allowlists are graded in the **narrowing** direction only. Adding a bogus word to any of them is invisible; removing a real one is loudly caught.

Measured on `2fae3a81c`, full `./internal/extractors/fsharp/`, `-count=1`, anchored `grep -cE '^\s*--- FAIL'`, `go vet` 0 on every row, each mutated line read back, restored by `cp` with shasum verified and `git diff HEAD` empty:

| site | allowlist | **widening** — add `sealed` | **narrowing** — drop `private` |
|---|---|---|---|
| `moduleRE` (`extractor.go:133`) | `rec\|public\|private\|internal` | **ALIVE, 0** | DEAD, **14** — `TestModuleModifiers_ModuleIsNeverMissed` |
| `memberRE` (`:275`) | `inline\|val\|public\|private\|internal` | **ALIVE, 0** | DEAD, **13** — `TestMemberModifiers_NameIsNeverTheModifier` |
| `typeRE` (`:337`) | `public\|private\|internal` | **ALIVE, 0** | DEAD, **26** — `TestTypeModifiers_TypeIsNeverMissed` |

The narrowing rows are decisive proof the tests exist, run, and bite — this is **not** "these patterns are untested". It is that all of their coverage points one way.

## Why this direction is the one that matters

Every serious defect in this repo has come from a predicate made **more permissive**, and recall-style must-have rows are structurally blind to it. The three tests above are all "…IsNeverMissed" / "…NameIsNeverTheModifier" — recall tests by construction. They assert that a legitimate declaration still produces its entity, which is exactly the property a *widened* allowlist preserves.

`sealed` is the right probe rather than an arbitrary one: it was measured **ALIVE at 0** against `activePatternRE` in `3bb6d28b8` while six hand-picked modifier words were all DEAD. That is the same word, the same direction and the same result, now at three more sites. Hand-picked forbidden rows leave the rest of the alphabet open; one structural assertion closed it for the `letRE`/`activePatternRE` pair, and nothing equivalent exists for these three.

## Why the existing parity test does not reach them

[...]

PR #7192, opened 2026-09-18
fix(#7151): the LOCAL module form `module Foo =` produced no entity at all — moduleRE was anchored `\s*$`

### LINKED ISSUE #7151: fsharp: moduleRE is anchored \s*$, so LOCAL modules ('module Foo =') are never extracted at all — a missing form, not a missing modifier

cajasmota:
## What

`moduleRE` was anchored `\s*$` (`internal/extractors/fsharp/extractor.go:132` on `a352a6a29`):

```
(?m)^([ \t]*)module(?:\s+(?:rec|public|private|internal)\b)*\s+([\w.]+)\s*$
```

so it matched only a declaration with **nothing after the name**. F# spells a module two ways:

```fsharp
module Foo                  // TOP-LEVEL: scope is the remainder of the FILE
module Foo =                // LOCAL:     scope is the offside block below
    let helper x = x + 1
```

The trailing ` =` made the anchor fail and the match was abandoned, so the local form produced **no entity at all** — with or without a modifier, at any indent. Measured on `a352a6a29` with `runFSharp` over `namespace N\n\nmodule Target =\n    let x = 1\n`:

| declaration | module entities before | after |
|---|---|---|
| `module Target =` | 0 | 1 |
| `module private Target =` | 0 | 1 |
| `module rec Target =` | 0 | 1 |
| `    module Target =` (indented) | 0 | 1 |
| `module Target` (top-level) | 1 | 1 (unchanged) |
| `module M = A.B.C` (abbreviation) | 0 | 0 (unchanged, now #7194) |

## The change

One optional group after the NAME capture: `([ \t]*=)?`.

- It is `[ \t]*=`, **not** `\s*=`. `\s` matches `\n` in Go, so a newline-crossing group could reach an `=` on a later line and attach it to a top-level declaration. **See the section below — a match-count comparison is blind to this, and the row that grades it asserts the signature instead.**
- The final `\s*$` is **left alone** for the opposite reason: `\s` includes `\r` and Go's `(?m)` `$` matches before the `\n` only, so `\s*$` is what absorbs the `\r` of a CRLF source. Narrowing it to `[ \t]*$` cannot consume the `\r` and drops **every module, of both forms, in every CRLF file**. **Now graded** — see the CRLF section below; it was ungraded prose in the previous round.
- It is anchored at end-of-line, so `module-abbrev` (`module M = A.B.C`, spec § 10) still does not match — unchanged, and pinned so it is not decided by accident. Whether an alias *should* mint is **#7194**, filed, not decided here.

## ⚠ STATED SCOPE LIMIT — the local arm's recall is conditional on `stripStringsAndComments` (#7193)

The new local arm is gated on `stripStringsAndComments`. That function **runs away** on two ordinary, legal F# constructs, blanking the remainder of the file — so a **real** `module Target =` below one is silently dropped and **this fix does not apply for the rest of that file**:

| file | modules emitted, today, on this branch |
|---|---|
| `let p = 1` … `module Target =` | `[Target]` (baseline) |
| `let p = @"C:\"` … `module Target =` | **`[]`** ← verbatim string, trailing backslash |
| `let q = '"'` … `module Target =` | **`[]`** ← char literal holding a quote |
| `let p = @"say ""hi"""` … `module Target =` | `[Target]` (well-formed verbatim string is fine) |
| `let p = @"C:\"` … `module Target` (top-level) | `[Target]` ← **POSITIVE CONTROL** |

**Mechanism:** the `case '"'` arm of `stripStringsAndComments` carries the comment `// Check for verbatim string @"..."` and performs **no such check**, so `inStr` treats `\` as a C-style escape and eats the closing quote of a verbatim string; separately there is **no general char-literal state**.

**READ THIS BEFORE FIXING #7193 (F4) — the package already has a recorded decision that a GENERAL char-literal scrub is wrong,** and the previous round's hand-off omitted it. `charBraceRE` (#6326) deliberately matches **only** `'{'` and `'}'`, because F# identifiers may end in an apostrophe, so a general `'.'` scrub misreads `c' '}'` — a primed identifier next to a char literal, ordinary F# — as the span `' '`. That reasoning and its counter-example are pinned by `TestFSharp_PrimedIdentifierBeforeCharLiteralBrace` in `hierarchy_test.go`. **A naive `'.'` state added to `stripStringsAndComments` turns that test RED at the same moment it turns the two recording subtests GREEN** — the reviewer's minimal simulated fix did exactly that. Any #7193 fix has to satisfy both. Cross-referenced from the extractor comment and the test header so the next author does not repeat it.

**This is PRE-EXISTING and NOT caused by this PR.** Before #7151 the local form minted 0 *unconditionally*, so the count in those two rows was 0 there too. Nine non-test call sites in the package are affected, not only this one. Filed as **#7193**.

**The attribution is clean** because of the last row: a TOP-LEVEL `module Target` in the **byte-identical** file still mints, because the top-level arm is ungated. So the loss belongs to the scrubber and to nothing else in the extractor.

**Recorded, not fixed**, by `TestLocalModule7151_ScrubRunawayHidesLocalModule_7193`. Its comment says so explicitly: a fix for #7193 **SHOULD turn its first two subtests red**, that is the **signal that it worked**, and they must be **updated deliberately with a note rather than deleted** — their failure must not be read as a regression in #7151. Fixing the scrubber **here** was deliberately declined: it needs a per-construct enumeration (verbatim strings, char literals, interpolated strings, `@` + triple-quote combinations) and mixing it in would make this PR's axes unreviewable.

Both counter-examples are legal F# — **derived, not executed** (no toolchain here): MS Learn "Strings" states a verbatim `@"…"` treats `\` literally and escapes a quote as `""`, and `'"'` is an ordinary character literal. `@"C:\"` (a trailing-backslash Windows path) and `'"'` (quote-aware parsing / CSV / JSON code) are both everyday idioms.

## CRLF — the other half of that bullet, previously ungraded prose (F1)

The previous round said of this bullet *"this was previously justified by a claim no row graded. Corrected."* **Only the first half was.** The `\s*$` / `\r` half stayed unhedged prose, and the review measured the narrowing mutant `([ \t]*=)?\s*$` → `([ \t]*=)?[ \t]*$` as **ALIVE at 0**. The package had **zero** CRLF coverage: no `\r\n` escape and no literal CR byte in any of its 19 `*_test.go` files, and the string `CRLF` appeared nowhere in the package — the previous round's comment was its first occurrence.

`TestLocalModule7151_CRLFSourceMintsBothForms` now grades it. Measured on this branch:

| CRLF fixture | HEAD | with `\s*$` narrowed to `[ \t]*$` |
|---|---|---|
| `module Target =\r\n` | `Target`, `"module Target ="` | **none** |
| `module Target\r\n` | `Target`, `"module Target"` | **none** |
| `module private Ns.Target =\r\n` | `Ns.Target`, `"module Ns.Target ="` | **none** |
| `module private Ns.Target\r\n` | `Ns.Target`, `"module Ns.Target"` | **none** |

The mutant is **M8**, now **DEAD at 5** (1 parent + all 4 subtests) where it was ALIVE at 0. The forbidden direction is a module abbreviation in CRLF still minting nothing — it fails alone under any widening that lets the anchor swallow the rest of the line, which would leave all four must-have rows passing.

**The blast radius is larger than what the newline-crossing row grades**, and that is worth stating: this is every module of both forms in a whole class of real files, versus one shape this PR itself calls not-legal-F#.

**Spelling, chosen deliberately because the package had no precedent either way:** the fixtures use the **escape `\r\n`** inside an interpreted string literal, **not** a literal CR byte in the file. A literal CR is invisible in review and easy for an editor to eat; the escape is explicit at the call site. The test says so.

**SCOPE — moduleRE only.** `namespaceRE` carries the **identical** `\s*$` exposure and is deliberately **not** covered here; that, the package-wide CRLF gap and the remaining unaudited `MustCompile` patterns are **#7198**. Both the test and the extractor comment say so, so the twin is not left looking audited.

## `[ \t]*` vs `\s*` — a match count is BLIND here, so the row asserts the artefact

This was previously justified by a claim **no row graded**. Corrected.

On `"module Foo\n=\n"` (and `"module Foo\n\n   =\n"`) **both** patterns match **exactly once**. They differ only in whether **group 3 captures**:

| pattern | matches | group 3 | classification | Signature |
|---|---|---|---|---|
| `([ \t]*=)?` (this PR) | 1 | unset | TOP-LEVEL | `module Foo` |
| `(\s*=)?` (the naive form) | 1 | `"\n="` | LOCAL | `module Foo =` ← wrong |

A bare match-count comparison, or a row that merely asserted "one module entity exists", would pass under **both** patterns and grade nothing. `TestLocalModule7151_NewlineBeforeEqualsStaysTopLevel` therefore asserts the **emitted signature**: a must-have on `Signature == "module Foo"`, graded by mutant **M7** (`[ \t]*=` → `\s*=`).

### ⚠ RETRACTED (F2): the previous round's "the forbidden row can fail alone" was FALSE

The previous body, the test header and the in-test comment all claimed the forbidden half "can fail **alone** (no module record in the file may carry a local ` =` signature, however many records there are)". **It cannot, and the claim is retracted rather than patched over.** On a single-declaration fixture the predicate fires only if some module record's `Signature` ends `" ="`, the only module record is `Foo` (the namespace record is filtered out by `fs7151Modules`), and the must-have pins `Foo.Signature` to exactly `"module Foo"` — so **forbidden is a subset of must-have for every possible extractor behaviour on those inputs.** Analytically impossible, and measured:

| probe | `--- FAIL` |
|---|---|
| M7 alone | 4 |
| **M7a** — M7 **plus** that forbidden loop neutralised (`if false && …`) | **4** ← contributes **zero** incremental grading |

It is now labelled a **dominated diagnostic** in all three places.

**The claim is EARNED by a new row that needs a SECOND module record.** `TestLocalModule7151_StrayEqualsBelowLaterDeclIsForbiddenAlone` puts the stray `=` below the **later** declaration:

```fsharp
namespace Outer
                    // module Foo -> TOP "module Foo"   (unaffected by the mutant)
module Foo
module Bar          // under `\s*=`: mis-signed "module Bar ="
=
```

`\s*` matches only whitespace, so it cannot reach from after `Foo` across the intervening `module Bar` text. Under M7 the must-have (scoped to `Foo` only, deliberately) **passes** while the forbidden **fails, alone, on `Bar`**. Measured incrementally:

| probe | `--- FAIL` |
|---|---|
| M7 alone | 4 |
| **M7b** — M7 **plus** the NEW forbidden loop neutralised | **3** ← it *is* incrementally grading |

**The arrangement first suggested — `module Foo\n=\nmodule Bar\n` — does NOT earn it**, and I checked rather than adopting it: there `\s*=` mis-signs **Foo**, so the must-have fails too and the two halves mask each other exactly as before. Reported back rather than shipped.

The shape this guards against — `module Foo` followed by a line whose first non-whitespace byte is `=` — is **not itself legal F#**, so the hazard is narrow. Stated so, rather than overclaimed; the mechanism is nonetheless the reason the pattern is written as it is, and it is now falsifiable.

## Subtype decision: both forms keep `module`; the distinction is in the SIGNATURE

```
module Foo    -> Signature "module Foo"
module Foo =  -> Signature "module Foo ="
```

**The earlier revision of this enumeration asserted eight rows and one of them was FALSE**, so all eight were **re-derived from source one at a time** rather than the wrong one being repaired in place. Exactly **TWO** consumers carry behaviour off this subtype, not three:

### Behaviour-carrying (2)

| consumer | verified how | effect of a NEW subtype |
|---|---|---|
| `internal/resolve/imports.go:285` | read: `if e.Kind == "SCOPE.Component" && e.Subtype == "module" { continue }` in pass 2 of the module→entity reverse index | a local module would start registering as a resolvable call target |
| `internal/mcp/denoise.go:137` | read: `bareKind == "component" && (subtype == "file" \|\| subtype == "module")` → `noiseContainer`, and again at `:141` off the property fallback | a local module would stop being denoised |

### Indifferent (6) — verified, not assumed

| consumer | verified how | why indifferent |
|---|---|---|
| `internal/extractor/file_carrier.go` clause 3 | `grep -n module internal/extractor/file_carrier.go` → hits at `:48 :71 :72 :75 :207`, **all inside `//` prose**; clause 3 is `records[i].Name == path` at `:217` | **THE PREVIOUSLY FALSE ROW.** It keys on `Name`, never on `Subtype`; a split subtype would have changed nothing there. What *does* change is a **Name** effect, not a subtype one: a local `module Core.fs =` in a root `Core.fs` can now reach the same path-named route a top-level one already reached |
| `internal/docgen/llm_bundle.go:1615-1617` | read `isModuleKind`: `strings.Contains(strings.ToLower(strings.TrimPrefix(kind,"SCOPE.")),"module")` | reads the **KIND**, not the subtype |
| `cmd/grafel/nestjs_shadow_fold_test.go:37,89,162` | `grep -n '"module"'` → `n.Subtype == "file" \|\| "import" \|\| "module"` | reads the subtype, but it is a **test** census with no product behaviour; a new subtype would merely start appearing in it |
| `internal/dashboard/handlers_iac.go:270,312` | read `iacToolForEntity`: gates on `language == "terraform" \|\| language == "hcl"` **before** testing `subtype == "resource" \|\| subtype == "module"`; the `:703` `"module"` is a Terraform ref-prefix string | never sees an fsharp record |
| `internal/vbnet/symbols.go:177` | read: `s.Kind == KindType && s.TypeName == "module"` over the VB.NET symbol table's own `Symbol` type | never sees an `EntityRecord` |
| `internal/mcp/mmapview.go:125,180` / `mmapview_test.go:467` | read: `if key == "module"`, `out["module"] = string(mod)` | `"module"` is a **PROPERTY KEY** (the module-rollup label), unrelated |

**Corrected in all three places the false claim lived:** this body, the `extractor.go` comment, and the `local_module_7151_test.go` header. The extractor comment now names the true key and says plainly that an earlier revision claimed otherwise and was wrong.

Both behaviour-carrying consumers read the subtype as *"container scope, not a callable"*, and a local module is a container scope by the same reading — so the subtype is the wrong place to carry the top-level/local distinction. The Signature is: it echoes the declaration head, it is an emitted artefact, and the rows assert it. (`entityTupleKey` does not hash `Properties`, so a Property would additionally have been invisible to the digest; a second reason not to put it there.)

**Consequence, stated plainly (review item 3):** because `mcp/denoise` classifies the new entity as `noiseContainer`, the payoff of this fix is **graph-level** — resolution, file carriers, docgen — and the new record does **not** surface in ordinary MCP output.

**Enlarged surface, stated rather than left to be inferred (review item 5):** the dedup key is `"module:"+name` and carries neither indent nor scope, so a file declaring **both** a top-level `module Foo` and a local `module Foo =` now has the two forms competing for the same key. The key stays #7144's arm 3 and no row here depends on it (one declaration of interest per source file), but this widening does make the collision **reachable**, and that is a fact about this commit rather than about #7144.

## The modifier allowlist is NOT touched (requirement 4)

`(?:\s+(?:rec|public|private|internal)\b)*` and the mandatory `\s+` separator are **byte-for-byte unchanged**; the review confirmed this by extracting both pattern literals and comparing programmatically — the prefix through `([\w.]+)` is byte-identical and the only delta is `\s*$` → `([ \t]*=)?\s*$`. #7181 measured that the allowlist is graded in the narrowing direction only (dropping `private` is DEAD at 14; adding `sealed` is ALIVE at 0). This commit neither improves nor worsens that. The new rows *do* vary the modifier across the local form. **Corrected nit:** the previous round said this makes #7135's `TestModuleModifiers_LocalModuleEqualsGapIsSeparate` "grade the fix instead of the gap". **Overstated** — that test asserts only `a != b`, so it passes at `0 == 0` and would keep passing if this fix were reverted; it is not a grader of this fix. What `TestLocalModule7151_ModifierRelationHoldsForLocalForm` adds is the artefact: it compares Name, Subtype, Signature **and** StartLine across the modifier phrases and `Fatalf`s if the base count is not exactly 1.

## Requirement 3 — the widening direction is scored, and the guard is deliberately narrow

These declaration scanners read the **raw `src`**, not the scrubbed copy (#7152), so a looser `moduleRE` enlarges what commented-out and string-embedded text can mint. The new **LOCAL arm only** is gated on `stripStringsAndComments` (byte-offset preserving, so the name capture's offsets index the scrubbed copy directly), computed lazily.

- `TestLocalModule7151_MaskedLocalFormMintsNothing` — `module Ghost =` inside `(* … *)`, inside a NESTED `(* (* … *) *)`, and inside `"""…"""`, with and without a modifier: 5 forbidden rows, 0 module entities.
  **FIXED (F3): the `nested block comment` row did not vary the axis it was named for.** With `module Ghost =` placed **before** the first `*)`, a non-nesting scrubber blanks it identically to the plain block-comment row — so it was a duplicate, and the Axes section nonetheless listed nested `(* (* … *) *)` as a masking context varied here. Measured: deleting the `depth++` branch from `stripStringsAndComments` left that row **GREEN**. The declaration now sits **between the inner and the outer closer**, and the same deletion mints `Ghost` with signature `"module Ghost ="` — so the row grades the depth tracking. Scored as **M9** below (a fixture-validity mutant of `stripStringsAndComments`, not of this PR's change).
- `TestLocalModule7151_MaskedTopLevelFormIsUNCHANGED` — the top-level `module Ghost` inside `(* … *)` **still mints**, exactly as before. That is #7152's pre-existing phantom, asserted here **not as an endorsement** but as proof the guard is narrow: had it been applied to both arms this row goes red and the commit would have been silently doing #7152's work. It should go red when #7152 lands and be rewritten to the forbidden direction.

## Legality — DERIVED, NOT EXECUTED

No F# toolchain here (`dotnet`, `fsc`, `fsharpc`, `fsi`, `mono` all absent; `javac` is the only compiler present).

1. **F# Language Specification § 10 "Namespaces and Modules"** (fsharp.github.io/fslang-spec/namespaces-and-modules/): `module-defn := attributes? module access? ident = module-defn-body` — the `=` and its body are part of the production, and the body is an offside block. The file-scoped spelling is the separate `named-module` production (`module long-ident module-elems`), which has **no** `=` and whose elements run to the end of the file. § 10 also gives `module-abbrev := module ident = long-ident`, the alias this change deliberately still does not match (#7194).
2. **MS Learn "Modules"** (learn.microsoft.com/dotnet/fsharp/language-reference/modules) prints both syntax blocks: top-level `module [accessibility-modifier] [qualified-namespace.]module-name` with declarations following and no `=`, and local `module [accessibility-modifier] module-name = declarations`, with the rule that a local module's contents must be indented under it while a top-level module's extend to the end of the file.
3. **MS Learn "Strings"** — the source for the two #7193 counter-examples above: a verbatim `@"…"` treats `\` literally and escapes a quote as `""`.

**Falsifier:** if `module Foo =` followed by an indented `let` were *not* a legal F# local module — i.e. if a compiler rejected `TestLocalModule7151_LocalFormIsExtracted`'s fixture — this change is a widening onto an illegal shape and should be reverted, not re-tuned. `dotnet fsi --use:<fixture>.fsx` on a machine with the toolchain settles it; it was **not run here**.

## Mutants — RE-SCORED after this round (the old table is superseded)

A fix round invalidates the old table, so **every** mutant was re-scored on the post-fix tree, not carried over. `go vet` into its own file first (a non-compiling mutant is VOID, not a kill); each substitution asserted its **exact pre-count** in Python (a multiline `grep -cF` counts per-line and reported 89 for a 1-occurrence block — caught and replaced) and printed its post-count; restored by `cp` from a byte-for-byte backup and verified with `shasum -c` **plus** an empty `git diff HEAD` after every round; `--- FAIL` counted with `grep -cE '^[[:space:]]*--- FAIL'`; `-count=1`; the **full** `./internal/extractors/fsharp/` suite every time.

| # | mutant | vet | `--- FAIL` | verdict | prev. round | decomposition |
|---|---|---|---|---|---|---|
| M0 | baseline, no mutant | 0 | 0 | — | 0 | — |
| M1 | **the fix reverted** — `extractor.go` restored wholesale to `a352a6a29` | 0 | **21** | DEAD | 18 | 1 `LocalFormIsExtracted` + its 13 subtests + 1 `BothFormsInOneFileStayDistinct` + 1 `ModifierRelationHoldsForLocalForm` parent (its 5 subtests never run — the parent `Fatalf`s) + 1 `ScrubRunaway…_7193` parent + its `well-formed verbatim string` subtest + **1 `CRLFSourceMintsBothForms` parent + its 2 LOCAL subtests** (the 2 top-level CRLF subtests pass — the old pattern matched the top-level form in CRLF already) = **21** |
| M2 | **positive control** — M1 with `local_module_7151_test.go` HELD OUT | 0 | **0** | ALIVE 0 | 0 | `go test -list` diff: exactly the **11** `TestLocalModule7151_*` funcs disappear and nothing else |
| M3 | **mis-scope A** — local form with the **top-level SIGNATURE** (`signature += " ="` → `_ = isLocal`) | 0 | **18** | DEAD | 15 | M1's set minus the revert-only rows: 1 parent + 13 subtests + `BothForms…` + **CRLF parent + its 2 LOCAL subtests** (they assert `"module Target ="`). `ModifierRelationHolds…` stays GREEN — the relation is symmetric under this mutant, which is why it is not the assertion that pins the distinction |
| M4 | **mis-scope B** — local form with the **top-level extent start** (`if isLocal { startLine = 1 }`) | 0 | **18** | DEAD | 15 | same decomposition as M3, same green relation, for the same reason |
| M5 | scrub gate on the local arm dropped (**both** the block and its `var moduleScrubbed string`; dropping only the block is **VOID** — `declared and not used`) | 0 | **9** | DEAD | 9 | 1 `MaskedLocalFormMintsNothing` + its 5 subtests + 1 `ScrubRunaway…_7193` parent + its 2 runaway subtests. Also what proves the two #7193 recording rows are **not vacuous** — they are sensitive to the gate in the opposite direction |
| M6 | scrub gate **widened to both arms** (`if isLocal {` → `if true {`) | 0 | **3** | DEAD | 3 | 1 `MaskedTopLevelFormIsUNCHANGED` + 1 `ScrubRunaway…_7193` parent + its `POSITIVE CONTROL` subtest. Independent second confirmation that one arm only is gated |
| M7 | `([ \t]*=)?` → `(\s*=)?`, the naive newline-crossing group | 0 | **4** | DEAD | 3 | 1 `NewlineBeforeEqualsStaysTopLevel` parent + its 2 subtests + **1 `StrayEqualsBelowLaterDeclIsForbiddenAlone`** (the new F2 earning row) |
| M7a | **F2 probe** — M7 **plus** the OLD forbidden loop neutralised | 0 | **4** | *no change* | *n/a* | proves that loop contributes **zero** incremental grading: it is dominated by its must-have sibling. This is why the "can fail alone" claim is retracted rather than defended |
| M7b | **F2 probe** — M7 **plus** the NEW (earning) forbidden loop neutralised | 0 | **3** | *−1* | *n/a* | proves the new forbidden row **is** incrementally grading |
| **M8** | **NEW (F1)** — `([ \t]*=)?\s*$` → `([ \t]*=)?[ \t]*$`, the CRLF narrowing | 0 | **5** | DEAD | **ALIVE 0** | 1 `CRLFSourceMintsBothForms` parent + **all 4** subtests. The claim the comment makes is now graded in exactly the direction it claims |
| **M9** | **NEW (F3)** — the `depth++` branch deleted from `stripStringsAndComments` (a **fixture-validity** mutant of an existing function, *not* of this PR's change) | 0 | **5** | DEAD | *n/a* | 3 pre-existing nesting tests + **1 `MaskedLocalFormMintsNothing` parent + its `nested block comment, declaration between the two closers` subtest**. Before F3 repositioned that fixture the subtest stayed **GREEN** under this mutant — the proof it was a duplicate of the plain block-comment row |

**Confound, declared:** M1 is a whole-file revert, so it perturbs the regex, the local gate and the signature simultaneously. M3–M8 are the decomposition: each isolates one arm of the change and each has its own distinct killer set. M7a/M7b are incremental-grading probes rather than mutants of the change, and M9 grades a FIXTURE rather than the change — all three are labelled as such rather than counted as coverage of this PR.

## Axes

**Varied:** modifier phrase (none, `private`, `internal`, `public`, `rec`, `rec private`); indentation (0, 4 spaces, a tab); name qualification (`Target` vs `Ns.Target`); whitespace between name and `=` (one space, three spaces, a tab, none); trailing whitespace after `=`; **whether the `=` is on the declaration's own line at all** (new); declaration form (local / top-level / module abbreviation); masking context (bare source, `(* … *)`, nested `(* (* … *) *)` **with the declaration between the two closers, so the row actually varies nesting** — F3, `"""…"""`); **whether a real declaration is preceded by an ordinary string or char literal** (new — the axis the first review found open, and where #7193 lives); **line terminator (LF and CRLF)** (new — F1; the package had zero CRLF coverage before this round); **the number of module declarations in the file** (one, or two with a stray `=` below the second — new, F2); and the two forms co-occurring in one file.

**Held constant:** entity Kind (always `SCOPE.Component`); the declared name (always `Target`/`Foo`, or `Ns.Target` where qualification is the axis, so a wrong capture is unambiguous); **one declaration of interest per source file except in the two rows that deliberately need a second** (`BothFormsInOneFileStayDistinct`, `StrayEqualsBelowLaterDeclIsForbiddenAlone` — both use DISTINCT names, so neither depends on the `"module:"+name` dedup key), so no row depends on that key — that key carries neither indent nor scope, is #7144's arm 3, and is deliberately not exercised (the enlarged collision surface is declared above); and the file path within each table.

**Not varied, so not claimed:** attributes (`[<AutoOpen>] module Foo =`); `module type`; a local module nested inside another local module; the module body beyond a single `let`; interpolated strings (`$"…"`) and `@` + triple-quote combinations as #7193 triggers — enumerating the scrubber's trigger set belongs to #7193, not here.

## Gate

- `go test -count=1 ./internal/extractors/fsharp/` — **PASS**, 0 `--- FAIL`, 0.66s. Every pre-existing test in the package stays green, including #7135's `TestModuleTypeModifiers_OnlyRealDeclarations` and `TestModuleModifiers_LocalModuleEqualsGapIsSeparate`.
- `gofmt -l internal/extractors/fsharp/` — clean. `go vet ./internal/extractors/fsharp/` — clean.
- `go test -count=1 ./cmd/grafel/` — **NOT re-run in this round, deliberately**, and the reason is that it could not be evidence: `writeSpanFixture6118` (`cmd/grafel/custom_extractor_spans_6118_test.go:32`) writes 18 file members — 7 `.cs`, 6 `.py`, 1 `.java`, 1 `.rs`, 1 `.go`, 1 `.ts`, 1 `.js` — and **zero** `.fs`/`.fsx`/`.fsi`, and `find internal/quality/golden cmd/grafel -iname '*.fs*'` returns **0**. The digest could not move. It passed (`ok … 172.418s`, digest unmoved) on the pre-review commit `1f35d993e`, and this round changes **no extractor behaviour at all** — only comments and new test rows.

## What I left out, deliberately

- **#7198 — `namespaceRE`'s identical `\s*$` / CRLF exposure**, the package-wide CRLF gap, and the remaining unaudited `MustCompile` patterns. F1's coverage is scoped to `moduleRE`; fixing the twin here would make this PR's axes unreviewable, and the comment cites #7198 so the neighbour is not left looking audited.
- **#7197 — the unreachable `if len(m) < 8` arity guard** at the head of the module loop (`moduleRE` has 3 capture groups, so `FindAllStringSubmatchIndex` always returns 8). Dead code, not a defect, filed separately and deliberately untouched here.
- **#7193 itself.** Disclosed with a positive control and a pinning row, not fixed. Fixing it needs a per-construct enumeration of the scrubber's trigger set and would make this PR's axes unreviewable.
- **#7144 entirely.** No extent work (`EndLine` still equals `StartLine` for both forms — arm 1), and no change to the scope-less `"module:"+name` dedup key (arm 3).
- **#7152.** The top-level arm still reads raw source; only the new local arm is gated, and the asymmetry is asserted in both directions rather than left to be inferred.
- **#7194 — the module abbreviation `module M = A.B.C`.** Whether an alias should mint is not decided here; what is pinned is that this commit did not decide it by accident.
- **Incidence.** No corpus frequency for any shape here — neither the fix's payoff nor #7193's reach is frequency-weighted. The case for the fix is that a documented declaration form was absent from the graph.

Closes #7151
Refs #7193, #7194, #7197, #7198


cajasmota:
## Adversarial review — #7192

Independent worktree at `origin/wt-7151-a1` (`1f35d993e`), not the author's. Everything below is MEASURED unless marked otherwise; probe sources and logs are in my scratchpad, the worktree is left clean (`git status --porcelain` empty, `shasum 71cbdff2cc…` restored after every round).

### What the pattern newly admits — enumerated, not hand-picked

Both patterns compiled side by side and run over 31 inputs, each also in CRLF, reporting the group-3 capture (`probe1.log`):

| newly matching (old=0 → new=1) | still not matching (old=0, new=0) |
|---|---|
| `module Foo =`, `module Foo=`, `module Foo  =`, `module Foo\t=`, `module Foo =␠`, `module Foo =\t`, `module Foo =␠␠␠`, `module rec Foo =`, `module private Foo =`, `module Foo.Bar =`, `  module Foo =` — every one classified LOCAL, in both LF and CRLF | `module Foo ==`, `module Foo =x`, `module Foo = A.B.C`, `module Foo = begin`, `module Foo =//c`, `module Foo = (* c *)`, `module Foo = =`, `module type Foo =`, `let module Foo =`, `x module Foo =`, `module =` |

Top-level rows (`module Foo`, `module Foo␠`, CRLF variants) are unchanged at 1 and still classified TOP. **The newly-admitted set is exactly the intended one**; I could not find an input that newly matches and should not. Attack 1 lands nothing.

### `[ \t]*` vs `\s*` — load-bearing, confirmed, with a caveat

The naive `(\s*=)?` is **not** neutralised by the trailing `\s*$`, contra the obvious counter-hypothesis. On `"module Foo\n=\n"` and `"module Foo\n\n   =\n"`, both patterns match **once**, but `\s*` captures group 3 (→ LOCAL, `Signature "module Foo ="`) while `[ \t]*` leaves it unset (→ TOP). So a bare match-count comparison would have shown no difference and the distinction is real only in the capture — worth saying, because that is the measurement the justification needs. Caveat: the shape it protects against (`module Foo` followed by a line whose first non-whitespace byte is `=`) is not legal F#, so the hazard is narrow; the mechanism claim is nonetheless correct as written.

### Consumer spot-checks — one of the eight is wrong

- `internal/resolve/imports.go:285` — **confirmed**, `e.Kind == "SCOPE.Component" && e.Subtype == "module"` → `continue`, so a local module does not register as a call target. Same treatment as a top-level module.
- `internal/mcp/denoise.go:137` — **confirmed**, `bareKind == "component" && (subtype == "file" || subtype == "module")` → `noiseContainer`. A newly-minted local module is denoised identically to a top-level one. Consistent, but worth stating plainly in the PR: the fix adds the entity to the graph and MCP still denoises it, so the payoff is graph-level, not surface-level.
- `internal/extractor/file_carrier.go` clause 3 — **the table row is FALSE**. `grep` for `module` outside comments in that file returns **zero** lines; clause 3 keys on `Name == path`, never on `Subtype`. A new subtype would have changed nothing there. The same claim is repeated in the new `extractor.go` comment ("extractor/file_carrier's clause 3 keys on it") and in the test file's header, so it will outlive the PR body. The correct count is **two** behaviour-carrying consumers, not three — the decision is still right, the stated reason is over-counted by one.

### FINDING — the new scrub gate silently reverts the fix for a whole file (measured)

The local arm is gated on `stripStringsAndComments`, which tracks `"` strings with a C-style `\` escape and has no `'…'` char-literal state. Two legal F# shapes make it run away and blank the rest of the file, so the **real** local module below it is dropped (`probe2.log`):

| file | modules emitted |
|---|---|
| `let p = 1` … `module Target =` | `[Target]` (baseline) |
| `let p = @"C:\"` … `module Target =` | **`[]`** |
| `let q = '"'` … `module Target =` | **`[]`** |
| `let p = @"C:\"` … `module Target` (top-level) | `[Target]` ← positive control |
| `let p = @"say ""hi"""` … `module Target =` | `[Target]` |
| `let x' = 1` … `module Target =` | `[Target]` |

The top-level control in the *identical* file still mints, so the loss is attributable to the new gate and to nothing else in the extractor. Both inputs are legal F# — *derived, not executed* (no F# toolchain here): MS Learn "Strings" states a verbatim `@"…"` treats backslash literally and escapes a quote as `""`, and `'"'` is an ordinary character literal. `@"C:\"` (trailing-backslash path) and `'"'` (quote-aware parsing/CSV/JSON code) are both ordinary F# idioms.

This is **not a regression** — before the PR the local form minted 0 unconditionally — but it is an undisclosed hole in exactly the recall the PR claims, and the "AXES / NOT VARIED" section does not cover it: masking context is varied only for text that *should* be suppressed, never for a real declaration preceded by an ordinary string. This is the "fixture pins one axis and leaves its neighbour open" shape.

### Mutants re-derived

- **M1** (`extractor.go` restored wholesale to `a352a6a29`; `git log a352a6a29..HEAD -- extractor.go` = exactly one commit, so the revert is the PR and nothing else): vet 0, **16** anchored `--- FAIL`. Decomposition re-derived from the failure list: 1 `LocalFormIsExtracted` parent + 13 subtests (the table has exactly 13 rows) + 1 `BothFormsInOneFileStayDistinct` + 1 `ModifierRelationHoldsForLocalForm` parent = **16**. Exact.
- **M3** (`signature += " ="` → `_ = isLocal`; match count asserted 1 before, 0 after, line read back): vet 0, **15**. Exact.
- **M6** (`if isLocal {` → `if true {`, line read back): vet 0, **1** — `MaskedTopLevelFormIsUNCHANGED`, the claimed sole killer. This settles attack 4 in **both** directions: the top-level phantom really exists on HEAD (the row passes at baseline) and the local arm really is the only gated one (widening the gate kills exactly that row). The control is not vacuous.
- Baseline: 0 `--- FAIL`, 0.63s. I did not re-run M2/M4/M5.

### Digest claim — author is right, so I did not spend the 172s

`writeSpanFixture6118` has **18** members: 7 `.cs`, 6 `.py`, 1 `.java`, 1 `.rs`, 1 `.go`, 1 `.ts`, 1 `.js`, and **zero** `.fs*` (my first count said 17 and no `.rs` — a `func ` inside a Go fixture string truncated my extraction; re-derived correctly). `find internal/quality/golden cmd/grafel -iname '*.fs*'` → **0**. The unmoved digest is evidence of nothing and re-running `./cmd/grafel/` would have bought nothing.

### #7181 claim — verified by diffing the pattern string, not the hunk

Extracted both literals from `git show a352a6a29:…` and HEAD and compared programmatically: the prefix through `([\w.]+)` is **byte-identical**; the only delta is `\s*$` → `([ \t]*=)?\s*$`. The modifier allowlist and the mandatory `\s+` separator are untouched.

### Module abbreviation

`module M = A.B.C`, `module private M = A.B.C` and the indented form all match **0** under both patterns — genuinely unchanged, not newly broken, and the anchor is what makes it so. I agree it deserves its **own issue**: it is a documented § 10 declaration form absent from the graph, i.e. the same defect class this PR fixes, and the alias question (mint an entity? an alias edge? nothing?) is a design call, not a regex tweak.

### What I could NOT verify

- Any F# legality statement (no toolchain; `javac` only). Both my new counter-examples are derived-not-executed and cited.
- Incidence of any of these shapes in a real corpus — neither the PR's claim nor my counter-examples are frequency-weighted.
- M2, M4, M5 (not re-run).
- Whether the new dedup collision surface matters: `"module:"+name` is shared across both forms, so a file with top-level `module Foo` **and** a local `module Foo =` now emits one entity where it previously emitted one for a different reason. The PR declares the key out of scope (#7144 arm 3) and holds one declaration per file, so no row depends on it — but the widening does enlarge that surface, and that is not stated.

---

**REQUEST-CHANGES**

1. **Blocking.** The scrub gate drops a real local module for the remainder of any file containing `@"…\"` or `'"'`. Either narrow the gate or, at minimum, add these two as measured rows with the top-level positive control, and say in the axes section that the local arm's recall is conditional on `stripStringsAndComments`. Silently, the fix does not apply to such files. A follow-up issue on the scrubber itself (verbatim-string backslash, char-literal state) would be the durable fix.
2. **Blocking, cheap.** Correct the `file_carrier.go` clause 3 row: it keys on `Name == path`, not on the subtype (zero non-comment occurrences of `module` in that file). Fix it in the PR body **and** in the `extractor.go` comment and the test-file header, and restate the count as two behaviour-carrying consumers. The decision itself stands.
3. Non-blocking: state that `mcp/denoise` classifies the new entity as `noiseContainer`, so the fix is graph-level and not visible in MCP output.
4. Non-blocking: file the module-abbreviation issue.
5. Non-blocking: note the enlarged `"module:"+name` collision surface (top-level and local with the same name now compete), even while the key stays #7144's.

Everything else I attacked held: the widening admits exactly the intended set in LF and CRLF, `[ \t]*` is genuinely load-bearing in the capture, M1/M3/M6 reproduce to the line, the narrow-guard control is honest, the digest argument is correct, and the #7181 no-touch claim is byte-exact.

cajasmota:
## Fix round — `c41caa06f`

Thanks for the review; the `file_carrier` catch and the scrub-gate measurement were both right, and the `[ \t]*` caveat ("a match count is blind here") was the most useful thing in it. Body rewritten; **no behaviour change** in this round — the regex, the local gate and the signature are byte-for-byte as reviewed. Three documentation defects and two ungraded axes.

**1 — the scrub-gate hole: DISCLOSED, not fixed.** Confirmed pre-existing, filed as **#7193** (`case '"'` carries a comment claiming a verbatim check that is not there; no char-literal state at all; nine non-test call sites). The PR body now carries a **⚠ STATED SCOPE LIMIT** section with your table including the top-level positive control, and says plainly it is not caused here (the local form minted 0 unconditionally before). `TestLocalModule7151_ScrubRunawayHidesLocalModule_7193` records today's behaviour with an in-file comment stating that a fix for #7193 **SHOULD turn its first two subtests red**, that this is the signal it worked, and that they must be updated deliberately with a note rather than deleted. Not fixed here: it needs a per-construct enumeration and would make this PR's axes unreviewable.

**2 — the false consumer row, and the other seven.** Corrected in **all three** places (body, `extractor.go` comment, test-file header), restated as **two** behaviour-carrying consumers, and the extractor comment now names the true key (`records[i].Name == path`, `file_carrier.go:217`) and says an earlier revision claimed otherwise and was wrong. The other **seven were re-derived from source one at a time** — all seven hold; the body shows the grep or the read for each. Two refinements worth flagging: `nestjs_shadow_fold_test.go` does read the subtype but is a **test** census with no product behaviour, and `file_carrier` **is** touched by this PR — just via **Name**, not Subtype (a local `module Core.fs =` in a root `Core.fs` now reaches the path-named route a top-level one already reached). Both are now stated.

**3 — `[ \t]*` graded by the artefact.** New `TestLocalModule7151_NewlineBeforeEqualsStaysTopLevel` asserts `Signature`, must-have plus a forbidden row that can fail alone, and the body states that a match-count comparison is blind here (your measurement: both patterns match once, they differ only in the capture). New mutant **M7** (`[ \t]*=` → `\s*=`) is DEAD at 3 and kills exactly that row.

**4/5 — non-blocking, both in.** #7194 named on the abbreviation row and in the extractor comment; `mcp/denoise`'s `noiseContainer` consequence stated (the payoff is graph-level, not MCP-visible); the enlarged `"module:"+name` collision surface declared in both the body and the test header.

**Mutants fully re-scored** (old table superseded, not carried over) — M1 **16→18**, M2 0, M3 15, M4 15, M5 **6→9**, M6 **1→3**, M7 **3** (new). Two harness defects caught in the process and stated in the body: a multiline `grep -cF` pre-count reported **89** for a 1-occurrence block (replaced with a Python exact count), and M5 as previously written is **VOID** rather than DEAD — dropping the gate block alone leaves `declared and not used: moduleScrubbed`, so the `var` decl must go too. M5's new killers also prove the two #7193 recording rows are **not vacuous**: they are sensitive to the gate in the opposite direction, and M6's new killer is an independent second confirmation that only one arm is gated.

`./internal/extractors/fsharp/` green, 0 `--- FAIL`; `gofmt -l` and `go vet` clean. `./cmd/grafel/` deliberately not re-run, for the reason you gave: zero `.fs*` in the digest fixture, so it could not be evidence — and this round changes no extractor behaviour.

cajasmota:
## Fix round 2 — `a8461c8ae`

All four addressed, plus the nits. **No behaviour change again** — the regex, the local gate and the signature are still byte-for-byte as first reviewed. Everything below is measured; the full `./internal/extractors/fsharp/` suite was run per mutant row with `-count=1`, `vet` gated before any count was read, exact match counts asserted in Python, `cp` restore verified by `shasum -c` + empty `git diff HEAD` + a zero marker grep.

**F1 — CRLF, now graded.** You were right that only the first half of that bullet was corrected. `TestLocalModule7151_CRLFSourceMintsBothForms`: both forms, with and without a modifier and a qualified name, asserting Name/Signature/StartLine plus a no-CR-leak check, and a forbidden row (an abbreviation in CRLF still mints nothing) that fails alone under any widening that lets the anchor swallow the line. The narrowing mutant `\s*$` → `[ \t]*$` is **M8, DEAD at 5** (parent + all 4 subtests) where it was ALIVE at 0. Scoped to `moduleRE` as instructed; `namespaceRE` cited as **#7198** in both the test and the extractor comment so the twin is not left looking audited. I used the **escape `\r\n`**, not a literal CR byte, and the test says why (a literal CR is invisible in review and an editor can eat it; the escape is explicit at the call site).

**F2 — the claim was false; retracted, then earned.** Removed from all three places and replaced with the dominance argument plus your M7a measurement. Earned by `TestLocalModule7151_StrayEqualsBelowLaterDeclIsForbiddenAlone`. **One correction to the suggested fixture** — `"namespace Outer\nmodule Foo\n=\nmodule Bar\n"` does **not** earn it: under `\s*=` it is **Foo** that gets mis-signed, so the must-have fails too and the halves mask each other exactly as before (measured: Foo → `"module Foo ="`, Bar → `"module Bar"`). The stray `=` has to sit below the **later** declaration — `"module Foo\nmodule Bar\n=\n"` — because `\s*` cannot reach from after `Foo` across the intervening `module Bar` text. Then Foo stays `"module Foo"` (must-have passes, scoped to Foo deliberately) and **Bar** becomes `"module Bar ="` (forbidden fails alone). Measured: M7 alone **4**; M7 + old loop neutralised **4** (zero incremental); M7 + new loop neutralised **3** (−1, incrementally grading).

**F3 — fixture repositioned.** Declaration now between the inner and outer closers. Scored as **M9** (`depth++` deleted — a fixture-validity mutant of `stripStringsAndComments`, labelled as such, not counted as coverage of this PR): **DEAD at 5**, and the `nested block comment` subtest is now among the killers, minting `Ghost` with `"module Ghost ="`. It was GREEN under that mutant before.

**F4 — cross-referenced.** `charBraceRE`/#6326 and `TestFSharp_PrimedIdentifierBeforeCharLiteralBrace`, with the apostrophe-terminated-identifier reasoning and the explicit warning that a naive `'.'` state turns that test red at the same moment it turns the two recording subtests green. In both the extractor comment and the test header.

**Nits:** subtest labels now read "RECORDS 0, a #7193 fix makes it 1"; the #7135 sentence is retracted in place (it asserts only `a != b`, passes at `0 == 0`, and is not a grader of this fix); consumer citations now name `resolve.BuildImportTable` / `mcp.classifyNoise` / the file-carrier clause by symbol rather than by line, with the reason stated.

**Re-scored table** (new rows invalidated the old one):

| # | mutant | vet | `--- FAIL` | prev | verdict |
|---|---|---|---|---|---|
| M0 | baseline | 0 | 0 | 0 | — |
| M1 | fix reverted wholesale | 0 | **21** | 18 | DEAD |
| M2 | M1, test file held out | 0 | 0 | 0 | ALIVE 0 (exactly 11 `TestLocalModule7151_*` removed) |
| M3 | local form, top-level signature | 0 | **18** | 15 | DEAD |
| M4 | local form, top-level extent | 0 | **18** | 15 | DEAD |
| M5 | scrub gate dropped (+ its `var`) | 0 | 9 | 9 | DEAD |
| M6 | gate widened to both arms | 0 | 3 | 3 | DEAD |
| M7 | `[ \t]*=` → `\s*=` | 0 | **4** | 3 | DEAD |
| M7a | M7 + old forbidden loop neutralised | 0 | 4 | n/a | *probe: zero incremental* |
| M7b | M7 + new forbidden loop neutralised | 0 | 3 | n/a | *probe: −1, grading* |
| **M8** | `\s*$` → `[ \t]*$` (CRLF) | 0 | **5** | **ALIVE 0** | DEAD |
| **M9** | `depth++` deleted (fixture-validity) | 0 | 5 | n/a | DEAD |

Every decomposition is in the body. M1/M3/M4 moved by exactly the CRLF parent + its 2 **local** subtests (the 2 top-level CRLF subtests pass under M1 — the old pattern already matched the top-level form in CRLF).

**#7197 and #7198 deliberately untouched**, both listed in "what I left out". `gofmt -l` and `go vet` clean; `./cmd/grafel/` still not re-run, for the reason already agreed (zero `.fs*` in the digest fixture, and no extractor behaviour changed in either round).

